PR #418, opened 2024-06-24
Fix/adding whisper hash in metadata

johnyrahul:
## What

- Option to enable highlighting fields, which used llm whisperer as extraction.
-  Updated the API deployment response with metadata.

## Why

- To take advantage of  highlighting features of  LLM Whisperer

## How

- Passing the LLM whisper hash as a response to API deployments, if the tool has enabled the highlight option.

- Have changes in the structure tool and workflow destination side. 

## Database Migrations

-  Added a new field in the model for the highlight-enabled flag.
-  Added a new field to store metadata in the file history table.

## Env Config

-  Need yo update the structure tool version

## Relevant Docs

-

## Related Issues or PRs

- https://github.com/Zipstack/unstract-sdk/pull/64
- https://github.com/Zipstack/unstract-adapters/pull/53

## Notes on Testing

- Tested the changes by enabling and disabling flag.
- Validated the API deployment response

## Screenshots

image
API Deployment:
image

## Checklist

- [ ]  Need to update the PDM local file for the corresponding changes in SDK and adapter side


harini-venkataraman (approved):
LGTM

muhammad-ali-e (approved):
LGTM

backend/workflow_manager/workflow/models/file_history.py
```diff
@@ -39,6 +39,7 @@ def is_completed(self) -> bool:
         db_comment="Error message",
     )
     result = models.TextField(blank=True, db_comment="Result from execution")
+    meta_data = models.TextField(blank=True, db_comment="MetaData from execution")
```
chandrasekharan-zipstack, line 42: NIT: Consider treating it as a single word `metadata`. I see an inconsistency with how we use snake / camel case and it'll be good to remain consistent

chandrasekharan-zipstack (approved):
LGTM - left a NIT comment
Why did we choose to control this behaviour through a Prompt Studio setting instead of directly adding it to the text extractor? If we wish to use highlighting feature for multi-doc chat we'll need to manage it separately with that feature this way

backend/workflow_manager/workflow/models/file_history.py
johnyrahul: The reason is the user has more flexible control of enabling the highlighting option. 
Enabling this setting adapter level will have an effect in all the places where we use the adapter. 
Now we have a control at per tool level.

PR #426, opened 2024-06-29
feat: Update PostgreSQL image to include pgvector extension

mutaiib:
## What
Update PostgreSQL image to include pgvector extension.

## Why
The current image under `docker/docker-compose-dev-essentials:servides:db` does not come with the `pgvector` extension enabled, halting the process of using PostgreSQL as the vector database.

## How
Replaced the `postgres:15.6` image with `pgvector/pgvector:pg15`.

No, since the image installs the `pgvector` and runs on the supported version of PostgreSQL.

## Database Migrations
N/A

## Env Config
N/A

## Relevant Docs
[pgvector GitHub](https://github.com/pgvector/pgvector)

## Related Issues or PRs
N/A

## Dependencies Versions
N/A

## Notes on Testing
N/A

## Screenshots
N/A

## Checklist


hari-kuriakose:
@mutaiib Thanks for the contribution!

This could definitely help when the user wants to use the same PostgreSQL instance both for Unstract backend metadata storage and as the Vector DB.

That being said, wanted to confirm the following:
- After changing default Unstract db to [`pgvector`](https://github.com/pgvector/pgvector) provided PostgresSQL instance - **a)** did the app run successfully, and **b)** were you able to add the same instance as a Vector DB too?
- New [`pgvectorscale` extension](https://github.com/timescale/pgvectorscale) is said to be more efficient and performant. Wondering if we could explore the following:
    - Use official PostgreSQL image itself, and add both pgvector and pgvectorscale extensions from source
    - User can choose `pgvector` (`IVFFlat or HNSW` indexes) or `pgvectorscale` (`StreamingDiskANN` index)

ritwik-g (approved):

chandrasekharan-zipstack (approved):

hari-kuriakose:
@ritwik-g @gaya3-zipstack @chandrasekharan-zipstack 
- For OSS it will be `pgvector` by default, and for Enterprise it will be `postgres` by default. Is this difference handled in respective envs?
- For existing OSS installations, will switching from current `postgres` to new `pgvector` automatically import existing data under `postgres_data` Docker volume without any compatibility issues?

chandrasekharan-zipstack:
@hari-kuriakose 
> * For OSS it will be `pgvector` by default, and for Enterprise it will be `postgres` by default. Is this difference handled in respective envs?

Yes, this difference is handled and controlled by the image we use for different deployments

> * For existing OSS installations, will switching from current `postgres` to new `pgvector` automatically import existing data under `postgres_data` Docker volume without any compatibility issues?

I tried this and I didn't face any issues, however will let @gaya3-zipstack confirm since she explicitly did some testing around this.

ritwik-g:
> @ritwik-g @gaya3-zipstack @chandrasekharan-zipstack 
> - For OSS it will be `pgvector` by default, and for Enterprise it will be `postgres` by default. Is this difference handled in respective envs?
> - For existing OSS installations, will switching from current `postgres` to new `pgvector` automatically import existing data under `postgres_data` Docker volume without any compatibility issues?

Regarding the second point. Since we are storing data in volume. The image change shouldn't cause any issue assuming the pgvector image also uses the same path for data storage

gaya3-zipstack:
> @hari-kuriakose
> 
> > * For OSS it will be `pgvector` by default, and for Enterprise it will be `postgres` by default. Is this difference handled in respective envs?
> 
> Yes, this difference is handled and controlled by the image we use for different deployments
> 
> > * For existing OSS installations, will switching from current `postgres` to new `pgvector` automatically import existing data under `postgres_data` Docker volume without any compatibility issues?
> 
> I tried this and I didn't face any issues, however will let @gaya3-zipstack confirm since she explicitly did some testing around this.

I also tested locally with existing data and could not see any issue.
I only saw a compatibility problem when a different version of pgvector is used.

gaya3-zipstack (approved):
Looks fine

PR #439, opened 2024-07-03
Handling  grpc.StatusCode.NOT_FOUND exception

muhammad-ali-e:
## What

- Implemented specific logging for NotFound errors.
- Simplified and cleaned up existing log entries.

## Why

- NotFound exceptions need to be caught and logged explicitly to improve error handling and debugging.
- Need simplification in logs for better readability and maintainability.

## How

- Added checks for NotFound error codes to log these exceptions specifically.

- No, this PR will not break any existing features. It enhances logging without altering the core functionality of the code.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/unstract/flags/src/unstract/flags/client/evaluation.py
+++ b/unstract/flags/src/unstract/flags/client/evaluation.py
@@ -51,8 +51,14 @@ class EvaluationClient(BaseClient):
             response = self.stub.Boolean(request)
             return bool(response.enabled)
         except grpc.RpcError as e:
-            logger.warning(
-                f"Error evaluating feature flag '{flag_key}' for {namespace_key}"
-                f": {str(e)}"
-            )
-            return False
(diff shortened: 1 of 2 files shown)


athul-rs (approved):
LGTM.
@muhammad-ali-e Since we are doing this here, can you also add the same for list flags.

muhammad-ali-e:
> @muhammad-ali-e Since we are doing this here, can you also add the same for list flags.

@athul-rs I believe the list_feature_flags function will not encounter a flag_not_found error. In any case is there any such error It would only be raised during the list_flags call, which is not as frequent. On the other hand, evaluate_feature_flag calls occur much more frequently.

PR #444, opened 2024-07-03
Enable ETL prerun checks

athul-rs:
## What

- Introducing ETL pre-run check to avoid running ETL scheduled on expired accounts.

## Why

- Currently on expired accounts ( with subscription ), the scheduled ETL pipelines should stop working when account expires.

## How

- Adding additional check on each time ETL is ran for enterprise users.

- NO

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

- Ran tests on accounts with and without subscription to check.

## Checklist


backend/account/subscription_loader.py
```diff
@@ -75,3 +76,27 @@ def load_plugins() -> list[Any]:
...
+    except ModuleNotFoundError:
+        logger.error("Subscription plugin not found.")
+        return False
+
+    org_plans = SubscriptionHelper.get_subscription(org_id)
+    if not org_plans or not org_plans.is_active:
+        return False
+
+    if org_plans.is_paid:
+        return True
```
jagadeeswaran-zipstack, line 100: Do we actually need this check here? anyway we're returning True at the end.

athul-rs: This check is not a necessary step as you mentioned, but by adding a check case for paid case here we can have a early return and avoid checking for the below conditions. This improves performance in such cases by avoiding additional checks before final return.

Hope this clarifies @jagadeeswaran-zipstack

jagadeeswaran-zipstack (approved):

backend/account/subscription_loader.py
```diff
@@ -75,3 +77,30 @@ def load_plugins() -> list[Any]:
...
+
+    if timezone.now() >= org_plans.end_date:
+        logger.debug("Trial expired")
```
chandrasekharan-zipstack, line 103:
```suggestion
        logger.debug("Trial expired for org {org_id}")
```

backend/account/subscription_loader.py
```diff
@@ -75,3 +77,30 @@ def load_plugins() -> list[Any]:
         logger.info("No subscription plugins found.")
 
     return subscription_plugins
+
+
+def etl_prerun_check(org_id):
```
chandrasekharan-zipstack, line 82:
```suggestion
def etl_prerun_check(org_id: str) -> bool:
```
NIT: Please add some docstrings on what we're trying to check here

backend/scheduler/tasks.py
```diff
@@ -57,13 +59,24 @@ def execute_pipeline_task(
...
+                os.environ.get("SUBSCRIPTION_PLUGIN_AVAILABLE", "False") == "True"
+            )
+            logger.info(f"Is subscription available: {subscription_plugin_available}")
+
+            if subscription_plugin_available and not etl_prerun_check(org_schema):
+                try:
+                    logger.info(f"Disabling ETL task: {pipepline_id}")
+                    disable_task(pipepline_id)
+                except Exception as e:
+                    logger.error(f"Failed to disable task: {pipepline_id}. Error: {e}")
```
chandrasekharan-zipstack, line 78:
```suggestion
                    logger.warning(f"Failed to disable task: {pipepline_id}. Error: {e}")
```
Since we've decided to suppress the error and not raise something to the user. Also since the pipline is disabled here, why don't we return early if we chose to suppress this?

backend/scheduler/tasks.py
```diff
@@ -57,13 +59,24 @@ def execute_pipeline_task(
...
         tenant: Organization = (
             get_tenant_model().objects.filter(schema_name=org_schema).first()
         )
         with tenant_context(tenant):
+            subscription_plugin_available = (
+                os.environ.get("SUBSCRIPTION_PLUGIN_AVAILABLE", "False") == "True"
+            )
+            logger.info(f"Is subscription available: {subscription_plugin_available}")
+
+            if subscription_plugin_available and not etl_prerun_check(org_schema):
```
chandrasekharan-zipstack, line 73: @athul-rs you anyway seem to check for this env within `etl_prerun_check(org_schema)`. You can move all of this over there and simplify code here

backend/account/subscription_loader.py
athul-rs: Done

backend/account/subscription_loader.py
athul-rs: Added docstring

backend/scheduler/tasks.py
athul-rs: This check needs to be here in celery task, removed from the other method since it is already checked here.

backend/scheduler/tasks.py
athul-rs: Addressed.

backend/account/subscription_loader.py
```diff
@@ -75,3 +77,32 @@ def load_plugins() -> list[Any]:
...
+    """Method to check subscription status before ETL runs.
+
+    Args:
+        org_id: The ID of the organization.
+
+    Returns:
+        A boolean indicating whether the pre-run check passed or not.
+    """
+    try:
+        from pluggable_apps.subscription.subscription_helper import SubscriptionHelper
```
muhammad-ali-e, line 92: Can we avoid this import by getting the module from subscription_plugins list iteself?

backend/scheduler/tasks.py
```diff
@@ -57,13 +59,27 @@ def execute_pipeline_task(
...
     logger.info(f"Executing pipeline name: {name}")
-
     try:
         logger.info(f"Executing workflow id: {workflow_id}")
         tenant: Organization = (
             get_tenant_model().objects.filter(schema_name=org_schema).first()
         )
         with tenant_context(tenant):
+            subscription_plugin_available = (
+                os.environ.get("SUBSCRIPTION_PLUGIN_AVAILABLE", "False") == "True"
```
muhammad-ali-e, line 69: Can we replace the env variable with some constants or checking the load_plugins() instaed of saving it in env variable?

backend/scheduler/tasks.py
```diff
@@ -57,13 +59,27 @@ def execute_pipeline_task(
...
+            )
+            logger.info(f"Is subscription available: {subscription_plugin_available}")
+
+            if subscription_plugin_available and not etl_prerun_check(org_schema):
```
muhammad-ali-e, line 73: Sugg: `not etl_prerun_check` Better to change to more meaning full name like validate_etl_run .

chandrasekharan-zipstack: Or we could be more explicit and just say `is_subscription_active()`

chandrasekharan-zipstack (commented):
1. I suggest not setting the env to determine if the plugin is present and rather check the loaded plugins itself
1. Since subscription related logic itself is a plugin for cloud - I think we should have this check / validation itself within the plugin and it should only be invoked here. For an OSS user - the message `Trial period ended` and other checks is irrelevant and is just a leak of our cloud logic

backend/scheduler/tasks.py
athul-rs: Moved to plugin check

chandrasekharan-zipstack (commented):
LGTM - however I'd still recommend moving some of the checks as a function of the plugin to the cloud repo as it doesn't to the OSS context. You could mark this with a TODO atleast

muhammad-ali-e (approved):

PR #456, opened 2024-07-05
Fix/enable-handle-fs-connector

kirtimanmishrazipstack:
## What

- Azure FS, GCS have been enabled
- Made necessary changes to comprehend all unstract fs connector
- Remove path in minio, not used
- Remove overwrite parameter from destination connector

## Why

- Remove path in minio, because path is not used. In test connection, we are only using bucket.
- Remove overwrite parameter from destination connector.  This is done because all fsspec connector is not using overwrite parms. Connectors like  `Dropbox`, `minio`, `GCS`, `Azure FS` will **always overwrite the file if present**. Where as `Gdrive` **always creates a new file if already present**. So this param is not needed in destination connector

## How

- Azure FS, GCS have been enabled from FE
- Made necessary changes in `backend/workflow_manager/endpoint/source.py` and `backend/workflow_manager/endpoint/destination.py` to comprehend all unstract fs connector
- Remove path in minio from `unstract/connectors/filesystems/minio/minio.py`
- Remove overwrite param `backend/workflow_manager/endpoint/static/dest/file.json` to remove overwrite  from destination connector 

- This PR only affects the FS connector. All FS connector have tested on local

## Database Migrations

-  N/A

## Env Config

- N/A

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

- Tested all 5 FS connectors as a `source` and `destination` on workflow

## Checklist


backend/workflow_manager/endpoint/base_connector.py
```diff
@@ -51,6 +51,23 @@ def get_fsspec(
...
+
+        Parameters:
+        - settings (dict): Connector-specific settings.
+        - connector_id (str): Identifier for the desired connector.
+
+        Returns:
+        UnstractFileSystem: An unstract fs connector instance.
+        """
+        if connector_id not in connectors:
+            raise ValueError(f"Invalid connector_id: {connector_id}")
```
chandrasekharan-zipstack, line 67:
```suggestion
            raise ValueError(f"Connector '{connector_id}' is not supported.")
```
I'm assuming this would be raised only when we decide to remove support for a connector

backend/workflow_manager/endpoint/fs_connector_helper.py
```diff
@@ -0,0 +1,29 @@
...
+
+        Args:
+            fs_cls_name (str): unstract fs class-name
+            root_path (str): root path of unstract fs connector
+            input_dir (str): connector directory provided by user
+
+        Returns:
+            str: unstract fs connector root directory
+        """
+        if fs_cls_name in UnstractFSConnector.GOOGLE_DRIVE_FS:
```
chandrasekharan-zipstack, line 23: Instead of using a separate constant - rely on the [pythonic way](https://stackoverflow.com/a/511060/12864260) to obtain the class name

chandrasekharan-zipstack (commented):
Changes look good for the most part, however do confirm if this requires any migration to avoid regressions on removing the parameters. I believe we persist these details in the DB

kirtimanmishrazipstack:
> Changes look good for the most part, however do confirm if this requires any migration to avoid regressions on removing the parameters. I believe we persist these details in the DB

@chandrasekharan-zipstack I believe  we are storing as json in `workflow_endpoints` table. When I remove the params, its will only change the json.

Screenshot from 2024-07-08 10-10-06

backend/workflow_manager/endpoint/fs_connector_helper.py
kirtimanmishrazipstack: @chandrasekharan-zipstack I am using same thing only [here](https://github.com/Zipstack/unstract/pull/456/files#diff-f38c2a49b866a35b58c08a1ee78638f0fa2a54afd5ca902da609cc6be0f7a113R188) and [here](https://github.com/Zipstack/unstract/pull/456/files#diff-b3d98d93fd437a9e3d2246946276a92b5689359ab1f6464534735d6775d72f7dR163)

Just to make code cleaner, I have used the constant part.

backend/workflow_manager/endpoint/base_connector.py
kirtimanmishrazipstack: done

backend/workflow_manager/endpoint/fs_connector_helper.py
chandrasekharan-zipstack: @kirtimanmishrazipstack if we do make a change in the class name, we'll have to update it here as well. In the name of readability we would have compromised on maintainability. If you initialize these constants from the class name itself - then I think it should be okay

backend/workflow_manager/endpoint/fs_connector_helper.py
```diff
@@ -0,0 +1,29 @@
...
+            root_path (str): root path of unstract fs connector
+            input_dir (str): connector directory provided by user
+
+        Returns:
+            str: unstract fs connector root directory
+        """
+        if fs_cls_name == UnstractFSConnector.GOOGLE_DRIVE_FS:
+            input_dir = str(Path(root_path, input_dir.lstrip("/")))
+            return f"{input_dir.strip('/')}/"
+        elif fs_cls_name in UnstractFSConnector.DROPBOX_FS:
```
ritwik-g, line 26: @kirtimanmishrazipstack why are we using `in` instead of `==` here?

backend/workflow_manager/endpoint/base_connector.py
```diff
@@ -51,6 +51,23 @@ def get_fsspec(
...
+        - settings (dict): Connector-specific settings.
+        - connector_id (str): Identifier for the desired connector.
+
+        Returns:
+        UnstractFileSystem: An unstract fs connector instance.
+        """
+        if connector_id not in connectors:
+            raise ValueError(f"Connector '{connector_id}' is not supported.")
+        connector = connectors[connector_id][Common.METADATA][Common.CONNECTOR]
+        return connector(settings)
```
ritwik-g, line 69: @kirtimanmishrazipstack looks like this same logic is duplicated in `get_fsspec`. Please modify it to use this method.

backend/workflow_manager/endpoint/fs_connector_helper.py
kirtimanmishrazipstack: @ritwik-g I changed that

backend/workflow_manager/endpoint/base_connector.py
kirtimanmishrazipstack: @ritwik-g to remove duplicates, I have made `get_fsspec` method use `get_fs_connector`. Some classes are using get_fsspec method, let them use this method. It wont break anything

backend/workflow_manager/endpoint/fs_connector_helper.py
```diff
@@ -0,0 +1,29 @@
...
+
+        Args:
+            fs_cls_name (str): unstract fs class-name
+            root_path (str): root path of unstract fs connector
+            input_dir (str): connector directory provided by user
+
+        Returns:
+            str: unstract fs connector root directory
+        """
+        if fs_cls_name == UnstractFSConnector.GOOGLE_DRIVE_FS:
```
muhammad-ali-e, line 23: Can we do this in their base class or connector class. It looks like we need to add the connector name and dependand functions in backend too once added a connector.

backend/workflow_manager/endpoint/destination.py
```diff
@@ -175,32 +177,41 @@ def copy_output_to_output_directory(self) -> None:
...
-            settings=connector_settings, connector_id=connector.connector_id
-        )
-        if not connector_fs.isdir(output_directory):
-            connector_fs.mkdir(output_directory)
+        try:
+            is_dir = destination_fsspec.isdir(output_directory)
+            if not is_dir:
+                destination_fsspec.mkdir(output_directory)
+        except (
+            DropBoxApiError
```
muhammad-ali-e, line 205: Make this Exception handler in it's own connecrot class since this exception is more related to that class only.

muhammad-ali-e (commented):
@kirtimanmishrazipstack Overall LGTM. But Please try to move the functions or methods that related to specific connectors in to their own classes or their base classes

muhammad-ali-e:
> @chandrasekharan-zipstack I believe we are storing as json in `workflow_endpoints` table. When I remove the params, its will only change the json. So no migration changes as per my understanding.
> @muhammad-ali-e let me know your thoughts

Yes I beleive so.It will not affect anything if it not used anywhere.

backend/workflow_manager/endpoint/fs_connector_helper.py
ritwik-g: @kirtimanmishrazipstack looks like we are using `in` still

ritwik-g (approved):
@kirtimanmishrazipstack I have only one comment. Other than that it looks fine.

But please make sure to address comments from @muhammad-ali-e before merging

backend/workflow_manager/endpoint/fs_connector_helper.py
kirtimanmishrazipstack: @muhammad-ali-e I added in connector class

backend/workflow_manager/endpoint/destination.py
kirtimanmishrazipstack: @muhammad-ali-e handled in connector exception class

unstract/connectors/src/unstract/connectors/filesystems/zs_dropbox/zs_dropbox.py
```diff
@@ -74,3 +75,19 @@ def test_credentials(self) -> bool:
...
+    @staticmethod
+    def get_connector_root_dir(input_dir: str, **kwargs: Any) -> str:
+        """Get roor dir of zs dropbox."""
+        return f"/{input_dir.strip('/')}"
+
+    def create_dir_if_not_exists(self, input_dir: str) -> None:
+        """Create roor dir of zs dropbox if not exists."""
+        fs_fsspec = self.get_fsspec_fs()
+        try:
+            fs_fsspec.isdir(input_dir)
```
harini-venkataraman, line 88: [SUGGESTION] I belive fsspec has method to check and create a directory - `auto_mkdir`
We can leverage it, if possible. [Ref](https://filesystem-spec.readthedocs.io/en/latest/api.html)

unstract/connectors/src/unstract/connectors/filesystems/minio/static/json_schema.json
```diff
@@ -32,12 +32,6 @@
       "title": "Bucket Name",
       "description": "Name of the bucket to be restricted to."
     },
```
harini-venkataraman, line 34: QQ : If we remove the path from the schema, will we not support adding files in the nested folders?

unstract/connectors/src/unstract/connectors/filesystems/unstract_file_system.py
```diff
@@ -69,3 +70,15 @@ def get_fsspec_fs(self) -> AbstractFileSystem:
     def test_credentials(self) -> bool:
         """Override to test credentials for a connector."""
         pass
+
+    @staticmethod
+    def get_connector_root_dir(input_dir: str, **kwargs: Any) -> str:
+        """Override to get root dir of a connector."""
+        return f"{input_dir.strip('/')}/"
+
+    def create_dir_if_not_exists(self, input_dir: str) -> None:
```
harini-venkataraman, line 79: We can use the auto_mkdir method of fsspec here also.

unstract/connectors/src/unstract/connectors/filesystems/google_drive/google_drive.py
```diff
@@ -95,3 +96,12 @@ def test_credentials(self) -> bool:
         except Exception as e:
             raise ConnectorError(str(e))
         return True
+
+    @staticmethod
+    def get_connector_root_dir(input_dir: str, **kwargs: Any) -> str:
+        """Get roor dir of gdrive."""
+        root_path = kwargs.get("root_path")
+        if root_path is None:
```
harini-venkataraman, line 104: Is the `root_path` being fetched from the user? If not, can we not show this error message to the user? Instead we can give some user friendly message, helping the user to take actions.

unstract/connectors/src/unstract/connectors/filesystems/google_drive/google_drive.py
```diff
@@ -95,3 +96,12 @@ def test_credentials(self) -> bool:
...
+            raise ValueError("root_path is required to get root_dir for Google Drive")
+        input_dir = str(Path(root_path, input_dir.lstrip("/")))
+        return f"{input_dir.strip('/')}/"
```
harini-venkataraman, line 107: Before striping and returning, can we add validations for cases where `input_dir` may not be valid?

unstract/connectors/src/unstract/connectors/filesystems/zs_dropbox/zs_dropbox.py
kirtimanmishrazipstack: @harini-venkataraman have handled that on connector class

unstract/connectors/src/unstract/connectors/filesystems/unstract_file_system.py
kirtimanmishrazipstack: @harini-venkataraman we need to create dir if not exists. Let this code be

unstract/connectors/src/unstract/connectors/filesystems/minio/static/json_schema.json
kirtimanmishrazipstack: yes, tested this

muhammad-ali-e (approved):
LGTM, Please ensure thorough testing with both new and existing connectors.

unstract/connectors/src/unstract/connectors/filesystems/google_drive/google_drive.py
kirtimanmishrazipstack: @harini-venkataraman `fs.isdir(input)` won't validate, afrter reported by QA. This is PR is just for making all fs connectors enable.


Let us do the enhancement once when needed. Also source.py will give the error if `fs.isdir(f"{input_dir.strip('/')}/")` is not valid

unstract/connectors/src/unstract/connectors/filesystems/google_drive/google_drive.py
kirtimanmishrazipstack: @harini-venkataraman  we have `root_path = root` enabled for all connectors internally. I am just using root_path for GDrive, other's don't even this. Its not being fetched from user.

PR #447, opened 2024-07-04
v2 changes of platform settings

muhammad-ali-e:
## What
- v2 changes of platform settings
  - Updated the application name.
  - Imported packages from v2 applications instead of v1.
  - Retrieved organization information from UserContext or Session instead of connection.tenant.

## Why

- The changes are necessary to align with the updated architecture and to ensure compatibility with the v2 applications.
- To streamline the handling of organization contexts without relying on django_tenants.

## How

- Modified `platform_auth_helper.py` to update the import statements from v2.
- Adjusted `platform_auth_service.py` to import from v2 applications and updated the methods of get_organization by using user context.
- Updated `serializers.py` to reflect import changes.
- Changed the application name in `apps.py`.
- Updated `views.py` for consistency with the new application structure.

### Changed files:

- platform_auth_helper.py (package import changes only)
- platform_auth_service.py
- serializers.py (import changes only)
- apps.py (app name change only)
- views.py

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

- The new methods are self-contained ensuring no impact on current functionality.
- Also under feature flag `multi_tenancy_v2`.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

1. https://github.com/Zipstack/unstract/pull/422
2. https://github.com/Zipstack/unstract/pull/430
3. https://github.com/Zipstack/unstract/pull/431
4. https://github.com/Zipstack/unstract/pull/432

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

I have read and understood the [Contribution Guidelines]().

PR #432, opened 2024-07-01
version 2 of user/organization tenant account module - UN-1424

muhammad-ali-e:
### Major Changes in this PR:
- **apps.py:** Updated configuration and initialization logic.
- **Imports from V2 Applications:** Adjusted imports to utilize the new V2 applications.
- **models.py:** Revised models to include new fields and relationships.
- **OrganizationMemberService in organization_member_service.py**:
  - New get_members and get_members_by_user_email methods for improved performance and accuracy.
- **user_view.py:** Refactored to use OrganizationMemberService.get_user_by_id instead of direct calls to OrganizationMember.objects.get.

The remaining changes involve duplicating or copying existing methods from other parts of the application to ensure consistency and reduce redundancy.

## Why

- Mandatory changes as part of the multi-tenancy design update.
- To streamline the handling of organization contexts without relying on django_tenants.

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

- The new methods are self-contained ensuring no impact on current functionality.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

1. https://github.com/Zipstack/unstract/pull/422
2. https://github.com/Zipstack/unstract/pull/430
3. https://github.com/Zipstack/unstract/pull/431

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

I have read and understood the [Contribution Guidelines]().

PR #431, opened 2024-07-01
Version 2 of User/Organization Account Module

muhammad-ali-e:
## What
- Created a new application with the same V1 methods, with  some changes in the models.
- Implementation of version 2 of the user/organization account module.
- Refactoring to remove django_tenants and replace it with a new context management approach.
- Updated the organization model and related methods for better structure and efficiency.

## Why

- Mandatory changes as part of the multi-tenancy design update.
- To streamline the handling of organization contexts without relying on django_tenants.

## How

- Replaced `tenant_context(organization)` with new `UserContext` methods.
- Updated the `StateStore` to store and manage the organization ID in the current context.
- Refactored the organization model by removing `TenantMixin` and redundant methods.
- Leveraged `OrganizationMixin` for better model structuring.

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

- The new methods are self-contained ensuring no impact on current functionality.
- Also under feature flag `multi_tenancy_v2`.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

1. https://github.com/Zipstack/unstract/pull/422
2. https://github.com/Zipstack/unstract/pull/430

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

I have read and understood the [Contribution Guidelines]().

PR #430, opened 2024-07-01
Feat/multi tenancy v2 context and organization

muhammad-ali-e:
## What

- Multi-tenanct -V2 Changes
- User context helper methods to retrieve the current user and their organizations.
- Oganization model mixin for adding organization fields to various models.

## Why

- These additions are part of the Multi-tenant v2 enhancements.

## How

- New helper methods for user context to get the current user and their organizations.
- Created a model mixin to add organization fields to different models.

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

-  All new methods are self-contained and not called from anywhere within the existing applications. The changes in user_session.py are controlled by a feature flag, ensuring they do not affect current functionality.

## Database Migrations

- No

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

1. https://github.com/Zipstack/unstract/pull/422

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

I have read and understood the [Contribution Guidelines]().

PR #461, opened 2024-07-08
v2 changes of connector and connector_auth

muhammad-ali-e:
## NOTE

- This PR is intended to make a copy of the existing applications to facilitate the transition to **v2**. It does not aim to change or optimize the existing code.
- Review comments can be added as **TODO** items for future enhancements.

## What

- Introduced v2 changes for the **connector** and **connector_auth** applications.

## Why

- To align with the updated architecture and ensure compatibility with v2 applications.
- To streamline the handling of organization contexts without relying on django_tenants.

## How

- Updated imports to reference v2 instead of v1 applications.
- Revised the `get_organization` method to utilize the user context.
- Renamed applications in `apps.py`.
- Added new `models` with updated table names.
- Removed the old unused `connector_metadata` field and updated `connector_metadata_b` to `connector_metadata`.

- The new methods are self-contained ensuring no impact on current functionality.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

1. https://github.com/Zipstack/unstract/pull/422
2. https://github.com/Zipstack/unstract/pull/430
3. https://github.com/Zipstack/unstract/pull/431
4. https://github.com/Zipstack/unstract/pull/432
5. https://github.com/Zipstack/unstract/pull/447

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


backend/connector_auth_v2/models.py
```diff
@@ -0,0 +1,141 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class ConnectorAuthManager(models.Manager):
+    def get_queryset(self) -> QuerySet:
+        queryset = super().get_queryset()
+        # TODO PAN-83: Decrypt here
+        # for obj in queryset:
+        #     logger.info(f"Decrypting extra_data: {obj.extra_data}")
```
chandrasekharan-zipstack, line 23:
```suggestion
```

backend/connector_v2/models.py
hari-kuriakose, line 1: @muhammad-ali-e Do the foreign key fields need cascade delete anywhere?

backend/connector_auth_v2/models.py
hari-kuriakose, line 1: @muhammad-ali-e Does the foreign key field need cascade delete?

hari-kuriakose (approved):
@muhammad-ali-e Models look good. Added couple of comments though.

Code duplication can be ignore for now, as we will delete v1 later.
Also migration files for v2 can be generated later together.

backend/connector_v2/connector_instance_helper.py
```diff
@@ -0,0 +1,331 @@
...
+        """Get one connector instance.
+
+            Use this method if the connector instance is unique for \
+                filter_params
+        Args:
+            workflow_id (str): _description_
+            connector_type (tuple[str, str]):  Specifies input/output
+            connector_mode (Optional[tuple[int, str]], optional).
+                Specifies database/filesystem
+            values (Optional[list[str]], optional).
```
vishnuszipstack, line 137: @muhammad-ali-e need to remove this? This is not available in args.

backend/connector_auth_v2/models.py
muhammad-ali-e: If I set cascade deletion for the User in this model, it will also affect the **Connector** entity, and subsequently, the **Workflow** entity and other related entities. It appears that `on_delete=models.SET_NULL `is used for the **ConnectorAuth** foreign key in the **Connector** entity.

I recall @athul-rs  previously working on some cascade deletion tasks (Might be not this part).


I believe we can implement cascade delete here, so that all follower entities will have their foreign key fields set to null accordingly.

cc: @chandrasekharan-zipstack

backend/connector_v2/models.py
muhammad-ali-e: I think it better to keep like this to avoid regression during the v2 migrations. We can make this as a TODO task.

muhammad-ali-e: Added a TODO item. We can consider this in on going PRs. Or once v2 migrations

backend/connector_auth_v2/models.py
muhammad-ali-e: Updated with cacade.

backend/connector_v2/connector_instance_helper.py
muhammad-ali-e: Updated

vishnuszipstack (approved):

PR #470, opened 2024-07-10
v2 changes of tool_instance and usage

muhammad-ali-e:
## NOTE

- This PR is intended to make a copy of the existing applications to facilitate the transition to **v2**. It does not aim to change or optimize the existing code.
- Review comments can be added as **TODO** items for future enhancements.

## What

- Introduced v2 changes for the **tool_instance** and **usage** applications.

## Why

- To align with the updated architecture and ensure compatibility with v2 applications.
- To streamline the handling of organization contexts without relying on django_tenants.

## How

- Updated imports to reference v2 instead of v1 applications.
- Revised the `get_organization` method to utilize the user context.
- Renamed applications in `apps.py`.
- Added new `models` with updated table names.
- Updated `usage` model with `DefaultOrganizationMixin` class. 
- Added `UsageModelManager` with `DefaultOrganizationManagerMixin` to get the organization data in `usage`.

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

- The new methods are self-contained ensuring no impact on current functionality.

## Database Migrations

- No

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

1. https://github.com/Zipstack/unstract/pull/422
2. https://github.com/Zipstack/unstract/pull/430
3. https://github.com/Zipstack/unstract/pull/431
4. https://github.com/Zipstack/unstract/pull/432
5. https://github.com/Zipstack/unstract/pull/447
6. https://github.com/Zipstack/unstract/pull/461
7. https://github.com/Zipstack/unstract/pull/464
8. https://github.com/Zipstack/unstract/pull/465
9. https://github.com/Zipstack/unstract/pull/466
10. https://github.com/Zipstack/unstract/pull/467

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

I have read and understood the [Contribution Guidelines]().

PR #465, opened 2024-07-09
Feat/multi tenancy v2 adapter and connector processor

muhammad-ali-e:
## NOTE

- This PR is intended to make a copy of the existing applications to facilitate the transition to **v2**. It does not aim to change or optimize the existing code.
- Review comments can be added as **TODO** items for future enhancements.

## What

- Introduced v2 changes for the **adapter_processor** and **connector_processor** applications.

## Why

- To align with the updated architecture and ensure compatibility with v2 applications.
- To streamline the handling of organization contexts without relying on django_tenants.

## How

- Updated imports to reference v2 instead of v1 applications.
- Revised the `get_organization` method to utilize the user context.
- Renamed application in `apps.py` of `adapter_processor`.
- Added new `models` with updated table names in adapter_processor with a mixin `DefaultOrganizationMixin`.
- Added `AdapterInstanceModelManager` with `DefaultOrganizationManagerMixin` to get the organization data.

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

- The new methods are self-contained and changes are under feature flags ensuring no impact on current functionality.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

1. https://github.com/Zipstack/unstract/pull/422
2. https://github.com/Zipstack/unstract/pull/430
3. https://github.com/Zipstack/unstract/pull/431
4. https://github.com/Zipstack/unstract/pull/432
5. https://github.com/Zipstack/unstract/pull/447
6. https://github.com/Zipstack/unstract/pull/461
7. https://github.com/Zipstack/unstract/pull/464

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

I have read and understood the [Contribution Guidelines]().

PR #464, opened 2024-07-09
v2 changes of workflow with endpoints

muhammad-ali-e:
## NOTE

- This PR is intended to make a copy of the existing applications to facilitate the transition to **v2**. It does not aim to change or optimize the existing code.
- Review comments can be added as **TODO** items for future enhancements.

## What

- Introduced v2 changes for the **workflow** and **workflow.endpoints** applications.

## Why

- To align with the updated architecture and ensure compatibility with v2 applications.
- To streamline the handling of organization contexts without relying on django_tenants.

## How

- Updated imports to reference v2 instead of v1 applications.
- Revised the `get_organization` method to utilize the user context.
- Renamed applications in `apps.py`.
- Added new `models` with updated table names.
- Removed the old unused `project` field from workflow model.
- Added `WorkflowModelManager` with `DefaultOrganizationManagerMixin` to get the organization data.

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

- The new methods are self-contained ensuring no impact on current functionality.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

1. https://github.com/Zipstack/unstract/pull/422
2. https://github.com/Zipstack/unstract/pull/430
3. https://github.com/Zipstack/unstract/pull/431
4. https://github.com/Zipstack/unstract/pull/432
5. https://github.com/Zipstack/unstract/pull/447
6. https://github.com/Zipstack/unstract/pull/461

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

I have read and understood the [Contribution Guidelines]().

PR #422, opened 2024-06-27
unstract-multi-tenantV2 feature-flag changes

muhammad-ali-e:
## What

- Implemented feature-flag changes for Multi-tenant Version 2.

## Why

- To enable the safe rollout of new features without affecting current feature.

## How

- Implemented feature flags to toggle new changes on or off as needed.

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

- All changes are protected by feature flags

## Database Migrations

-  No

## Env Config

- Update configurations in platform-service and prompt-service as follows
```
# Feature Flags
EVALUATION_SERVER_IP=unstract-flipt
EVALUATION_SERVER_PORT=9000
PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python
```

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

I have read and understood the [Contribution Guidelines]().

PR #471, opened 2024-07-10
v2 changes of ProfileManager,core and studio

muhammad-ali-e:
## NOTE

- This PR is intended to make a copy of the existing applications to facilitate the transition to **v2**. It does not aim to change or optimize the existing code.
- Review comments can be added as **TODO** items for future enhancements.

## What

- Introduced v2 changes for the **prompt_studio.prompt_studio** , **prompt_studio.prompt_studio_core** and **prompt_studio.prompt_profile_manager** applications.

## Why

- To align with the updated architecture and ensure compatibility with v2 applications.
- To streamline the handling of organization contexts without relying on django_tenants.

## How

- Updated imports to reference v2 instead of v1 applications.
- Revised the `get_organization` method to utilize the user context.
- Renamed applications in `apps.py`.
- Added new `models` with updated table names.
- Updated `CustomTool` model with `DefaultOrganizationMixin` class. 
- Added `CustomToolModelManager` with `DefaultOrganizationManagerMixin` to get the organization data in `CustomTool` of `prompt_studio_core_v2`.
- 

- The new methods are self-contained ensuring no impact on current functionality.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

1. https://github.com/Zipstack/unstract/pull/422
2. https://github.com/Zipstack/unstract/pull/430
3. https://github.com/Zipstack/unstract/pull/431
4. https://github.com/Zipstack/unstract/pull/432
5. https://github.com/Zipstack/unstract/pull/447
6. https://github.com/Zipstack/unstract/pull/461
7. https://github.com/Zipstack/unstract/pull/464
8. https://github.com/Zipstack/unstract/pull/465
9. https://github.com/Zipstack/unstract/pull/466
10. https://github.com/Zipstack/unstract/pull/467
11. https://github.com/Zipstack/unstract/pull/470

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


chandrasekharan-zipstack:
@muhammad-ali-e how do we manage any code changes in the existing (v1) version of this duplicated code?  Do we need to make the same changes in both these versions until this transition is made?

backend/prompt_studio/prompt_studio_v2/models.py
```diff
@@ -0,0 +1,127 @@
...
+    prompt = models.TextField(
+        blank=True, db_comment="Field to store the prompt", unique=False
+    )
+    tool_id = models.ForeignKey(
+        CustomTool,
+        on_delete=models.SET_NULL,
+        related_name="tool_studio_prompts",
+        null=True,
+        blank=True,
+    )
```
hari-kuriakose, line 52: @muhammad-ali-e Shouldn't this be `on_delete=models.PROTECT`?
The prompts should not exist if the parent custom tool is deleted, right?

hari-kuriakose (approved):
@muhammad-ali-e Models look good.

Added a comment though, please check.

We will take care of duplication when we delete v1 eventually and also migration files can be generated later together.

backend/prompt_studio/prompt_studio_v2/models.py
muhammad-ali-e: @hari-kuriakose I think , It shouldn't needed When itool is deleted . So It shoudl CASCADE
@harini-venkataraman @chandrasekharan-zipstack Can you confirm here?

chandrasekharan-zipstack: @hari-kuriakose @muhammad-ali-e I think it should be `CASCADE` here since we should delete the individual prompts when the tool is deleted. On the other hand if we used `PROTECT`, we won't be able to delete the prompt studio project until all the prompts within it are deleted which is not what we want.
Referred [this SO answer](https://stackoverflow.com/a/38389488/12864260)

backend/prompt_studio/prompt_profile_manager_v2/models.py
```diff
@@ -0,0 +1,115 @@
...
+
+from account_v2.models import User
+from adapter_processor_v2.models import AdapterInstance
+from django.db import models
+from prompt_studio.prompt_studio_core_v2.exceptions import DefaultProfileError
+from prompt_studio.prompt_studio_core_v2.models import CustomTool
+from utils.models.base_model import BaseModel
+
+
+class ProfileManager(BaseModel):
```
gaya3-zipstack, line 11: Don't we need org info here @muhammad-ali-e ?

muhammad-ali-e: Profile is tightly coupled with Tools.So Ithink  it won't required. Right?

muhammad-ali-e:
> @muhammad-ali-e how do we manage any code changes in the existing (v1) version of this duplicated code? Do we need to make the same changes in both these versions until this transition is made?

@chandrasekharan-zipstack We can consider the delta cahnges while the datamigrations in staging.

backend/prompt_studio/prompt_studio_v2/models.py
muhammad-ali-e: Updated

backend/prompt_studio/prompt_profile_manager_v2/models.py
gaya3-zipstack: But if we are looking out for multi-tenancy, we will need org id stamped against the major entities that we maintain. Here, I am not very clear on how we maintainthe relationship between tools and orgs. But we will need that relationship. If you say that by using tools of an org, we can pull profiles of the same org, it may be ok for now. But a clean separation is preferrable in my opinion @muhammad-ali-e 
CC: @hari-kuriakose

backend/prompt_studio/prompt_studio_v2/models.py
```diff
@@ -0,0 +1,127 @@
+import uuid
+
+from account_v2.models import User
+from django.db import models
+from prompt_studio.prompt_profile_manager_v2.models import ProfileManager
+from prompt_studio.prompt_studio_core_v2.models import CustomTool
+from utils.models.base_model import BaseModel
+
+
+class ToolStudioPrompt(BaseModel):
```
chandrasekharan-zipstack, line 10: @muhammad-ali-e please add the org_id info to this class as well if we decide to add it after [this conversation](https://github.com/Zipstack/unstract/pull/471/files#r1675387307)

backend/prompt_studio/prompt_studio_core_v2/models.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def for_user(self, user: User) -> QuerySet[Any]:
+        return (
+            self.get_queryset()
+            .filter(models.Q(created_by=user) | models.Q(shared_users=user))
+            .distinct("tool_id")
+        )
+
+
+class CustomTool(DefaultOrganizationMixin, BaseModel):
```
chandrasekharan-zipstack, line 31: @muhammad-ali-e I notice that we set the org_id by referring the `StateStore`. Where / how will this be set in case of deployed pipeline runs?

chandrasekharan-zipstack (approved):
LGTM for the most part - I suggest adding org_id info to all the models for the sake of uniformity and making it easier to list / identify each record instead of requiring an extra lookup.
Also I suggest if any unnecessary prefix of `prompt_studio_` can be removed, it'll be good

harini-venkataraman (approved):

backend/prompt_studio/prompt_profile_manager_v2/models.py
harini-venkataraman: @muhammad-ali-e Adding to this, if we decide to scale any model in any new product calls, adding org_id will need additional migration efforts. It would be better to have org_id in all new models, irrespective of how they are coupled or accessed.

harini-venkataraman: @muhammad-ali-e Adding to this, if we decide to scale any model in any new product calls, adding org_id will need additional migration efforts. It would be better to have org_id in all new models, irrespective of how they are coupled or accessed.

harini-venkataraman: @muhammad-ali-e Adding to this, if we decide to scale any model in any new product calls, adding org_id will need additional migration efforts. It would be better to have org_id in all new models, irrespective of how they are coupled or accessed.

muhammad-ali-e: @hari-kuriakose could you please confirm if we need to establish an organization relation for every table, even if it's not immediately necessary? 
 1. I believe it is straightforward to retrieve the organization ID from customTool, which already has this relationship. Currently, we don't need a direct relation with the organization for the ProfileManager model; we only need the customTool relation. If we require this relation in the future, we can create a migration to add the organization ID, which we can obtain from the too.
2. Adding unnecessary relationships will increase complexity and resource utilization, and impact performance. 
3. The only potential benefit is avoiding future migrations if the relationship is needed later. However, I don't think we will need it in the future either.

backend/prompt_studio/prompt_studio_core_v2/models.py
muhammad-ali-e: @muhammad-ali-e on pipeline run

backend/prompt_studio/prompt_profile_manager_v2/models.py
hari-kuriakose: @muhammad-ali-e Agree with you here.

Folks, the nature of RDBMS is not to de-centralize foreign keys, instead to centralize concepts and leverage the power of joins.

That being said, assuming we have a join path from profile to org. Now since the foreign keys in respective models are already indexed the performance should also be ok, should we need this query in the future.

gaya3-zipstack: I am ok with not adding it for now as long as there is way to cleanly pull entities of an org and give the user , the multi-tenant experience! @muhammad-ali-e @hari-kuriakose

gaya3-zipstack (approved):
Looks fine.

backend/prompt_studio/prompt_studio_v2/models.py
hari-kuriakose: @chandrasekharan-zipstack This would not be required for now as per mentioned conversation.

PR #481, opened 2024-07-11
FEAT: Simple Prompt Studio CRUD functionality

tahierhussain:
## What

1. Added the list of public URLs for SPS.
2. Modified the existing components to support SPS.

## Why

NA

## How

NA

Yes, the existing modules have been modified to support SPS, so there is a chance that this pull request could break existing features.

## Database Migrations

NA

## Env Config

NA

## Relevant Docs

NA

## Related Issues or PRs

1. FE: https://github.com/Zipstack/unstract-cloud/pull/145
2. BE: https://github.com/Zipstack/unstract-cloud/pull/141

## Dependencies Versions

NA

## Notes on Testing

NA

## Screenshots
NA

## Checklist


chandrasekharan-zipstack (approved):
BE change LGTM

frontend/src/components/custom-tools/document-manager/DocumentManager.jsx
```diff
@@ -92,22 +92,24 @@ function DocumentManager({ generateIndex, handleUpdateTool, handleDocChange }) {
...
-      };
-      items[1] = {
-        key: "2",
-        label: (
-          <Tooltip title="Raw View">
-            <FileTextOutlined />
-          </Tooltip>
-        ),
-      };
+      items = [
```
athul-rs, line 95: @tahierhussain Small doubt here, why are we not adding this to the top defined place where `items` is initialized, seems like we are not adding anything sensitive here and since its just a label; modification ?

athul-rs (changes requested):

frontend/src/components/custom-tools/prompt-card/Header.jsx
```diff
@@ -157,30 +158,34 @@ function Header({
...
-                <PlayCircleFilled className="prompt-card-actions-head" />
-              </Button>
-            </Tooltip>
+            {!isSimplePromptStudio && (
+              <Tooltip title="Run All">
+                <Button
+                  size="small"
+                  type="text"
+                  className="prompt-card-action-button"
+                  onClick={() => handleRunBtnClick()}
```
jagadeeswaran-zipstack, line 167: I think this change exist from my PR. can please change remove that arrow func here

jagadeeswaran-zipstack (approved):
LGTM

frontend/src/components/custom-tools/document-manager/DocumentManager.jsx
tahierhussain: @athul-rs We need to render the modified items specific to Simple Prompt Studio. We cannot add them at the top where we're initializing `items`. Since this component will be used in both Main and Simple Prompt Studio, we can only determine inside the function whether the current prompt studio is Simple or Main (using `isSimplePromptStudio` state).

frontend/src/components/custom-tools/prompt-card/Header.jsx
tahierhussain: Done, @jagadeeswaran-zipstack

frontend/src/components/custom-tools/document-manager/DocumentManager.jsx
athul-rs: We could improve the part of code and avoid `useEffect` here, but upon discussion this is not staright-forward change and by design we'll keep go forward with this for now.

athul-rs (approved):
LGTM

Deepak-Kesavan (approved):

backend/backend/public_urls.py
```diff
@@ -22,6 +22,7 @@
 
 path_prefix = settings.PATH_PREFIX
 api_path_prefix = settings.API_DEPLOYMENT_PATH_PREFIX
+simple_prompt_studio_path_prefix = settings.SIMPLE_PROMPT_STUDIO_PATH_PREFIX
```
harini-venkataraman, line 25: @tahierhussain Let us move this line into the try-catch block. It may cause OSS itself to fail if the config is missing.

harini-venkataraman (changes requested):

backend/backend/public_urls.py
tahierhussain: Thanks for the catch, @harini-venkataraman. Done.

frontend/src/components/custom-tools/document-parser/DocumentParser.jsx
```diff
@@ -242,9 +249,14 @@ function DocumentParser({
       prompt_id: details.prompts[startIndex]?.prompt_id,
     };
 
+    let url = promptUrl("reorder");
```
vishnuszipstack, line 252: @tahierhussain trailing "/" is not required for this? i believe we are following like every url should have trailing "/".

frontend/src/setupProxy.js
```diff
@@ -8,11 +8,4 @@ module.exports = (app) => {
       changeOrigin: true,
     })
   );
-  app.use(
-    "/public",
-    createProxyMiddleware({
-      target: process.env.REACT_APP_BACKEND_URL,
-      changeOrigin: true,
-    })
-  );
```
vishnuszipstack, line 17: @tahierhussain QQ: why this is removed? not required anymore?

tahierhussain: @vishnuszipstack Yes, it is no longer necessary. Initially, I added this when I planned for the SPS APIs to start with `/public`. However, I later changed the prefix to the default one, i.e., `api/v1`, and made the SPS APIs public.

frontend/src/components/custom-tools/document-parser/DocumentParser.jsx
tahierhussain: @vishnuszipstack  This API does not require a trailing slash, similar to all the SPS APIs. I recall some of the team members were planning to remove the trailing slash from all APIs as it is unnecessary.

@chandrasekharan-zipstack @Deepak-Kesavan Please let us on the status of that, and I will make the changes accordingly.

athul-rs (approved):

backend/prompt_studio/prompt_studio/helper.py
```diff
@@ -6,27 +6,61 @@
...
+        if is_sps:
+            try:
+                from pluggable_apps.simple_prompt_studio.sps_prompt.models import (
+                    SPSPrompt,
+                )
+
+                return SPSPrompt
+            except ImportError as e:
+                logger.error("SPSPrompt model could not be imported.")
+                raise ImportError("SPSPrompt model could not be imported.") from e
```
chandrasekharan-zipstack, line 32: @tahierhussain we're practically catching an ImportError and re-raising the same. I don't see any benefit of doing this except for logging which will anyway be done by the middleware. Is this actually needed?

backend/prompt_studio/prompt_studio/helper.py
```diff
@@ -6,27 +6,61 @@
...
+            is_sps (bool): Flag to determine the prompt model to use.
 
         Returns:
             list[dict[str, int]]: A list of updated prompt data with their IDs
             and new sequence numbers.
         """
-        prompt_instance: ToolStudioPrompt = ToolStudioPrompt.objects.get(pk=prompt_id)
-        filtered_prompts_data = []
+
+        prompt_model = PromptStudioHelper.get_prompt_model(is_sps)
```
chandrasekharan-zipstack, line 55: If sps is a cloud only feature, there should be no mention of this in OSS code ideally - do you think we can properly make this class a plugin instead? How / when will `is_sps` be `True`?

backend/pyproject.toml
```diff
@@ -50,6 +50,7 @@ dependencies = [
     "unstract-workflow-execution @ file:///${PROJECT_ROOT}/../unstract/workflow-execution",
     "azure-identity==1.16.0",
     "azure-mgmt-apimanagement==3.0.0",
+    "google-cloud-recaptcha-enterprise>=1.19.0",
```
chandrasekharan-zipstack, line 53: @tahierhussain why do we add this dependency and not use it anywhere in the OSS code?

frontend/src/components/custom-tools/document-parser/DocumentParser.jsx
chandrasekharan-zipstack: I suggest we remain consistent and add it for all - we can take a call to remove it altogether in all APIs if necessary. Doing this for a few APIs alone can be confusing

backend/prompt_studio/prompt_studio/helper.py
tahierhussain: @chandrasekharan-zipstack We're using the same logic for reordering the prompts in both the main and simple prompt studio. The `is_sps` argument will be `True` if the request is being made from SPS. If you look at the `get_prompt_model` method in the same [file](https://github.com/Zipstack/unstract/blob/bc554ec5c3748f25057f05b86d1d3f7c1208ae43/backend/prompt_studio/prompt_studio/helper.py#L10), you'll see that we use the `is_sps` value to decide which model to use.

backend/prompt_studio/prompt_studio/helper.py
tahierhussain: @chandrasekharan-zipstack I've made the necessary change. FYI, I wasn't fully aware that this would be handled in the middleware. Thanks for the catch!

tahierhussain: @chandrasekharan-zipstack I have made modifications to this file. As per our earlier discussion, I am now passing the model class from both the main and simple prompt studio views files. This approach eliminates the need for the block of code where we were conditionally retrieving the model class.

backend/prompt_studio/prompt_studio/helper.py
tahierhussain: @chandrasekharan-zipstack Thank you for the suggestion to eliminate the use of is_sps. I have made the necessary changes. Please review that part once more and address the comment.

backend/pyproject.toml
tahierhussain: @chandrasekharan-zipstack As of now, I don't believe we support adding a dependency inside the cloud's backend. Let's discuss this with @hari-kuriakose during our next call and find a way to manage this situation.

frontend/src/components/custom-tools/document-parser/DocumentParser.jsx
tahierhussain: Done, @chandrasekharan-zipstack @vishnuszipstack. The trailing slash has been added to all the SPS related APIs.

vishnuszipstack (approved):

jagadeeswaran-zipstack (approved):

Deepak-Kesavan (approved):

backend/pyproject.toml
hari-kuriakose: @tahierhussain Let's remove any dependencies which are not required in OSS.

Then, let's add a `requirements.txt` containing those Enterprise only dependencies, and try to load them via:
```
pdm lock --override requirements.txt
```
Please check https://pdm-project.org/latest/usage/dependency/#via-cli-option. Let's see if this suits us.

**NOTE**: This will need changes in Enterprise Dockerfiles and CI workflows.

chandrasekharan-zipstack: @tahierhussain  @hari-kuriakose I hope these dependencies reside along with the plugin that's written and not in this repo.

harini-venkataraman (approved):
LGTM

backend/pyproject.toml
tahierhussain: @chandrasekharan-zipstack Done. The changes have been made in the cloud repository.

chandrasekharan-zipstack (approved):
BE changes LGTM

PR #486, opened 2024-07-15
Moving public prefix within try-catch block

harini-venkataraman:
## What
PR moves the path prefix needed for public sharing into the code flow executed if the plugins are loaded.
...
## Why
If the plugins are missing, it throws error while login.
...

## How
Changing the import block.
...

No, it is a change in import to a different flow, and this import is not used else where.
...

## Database Migrations
Not applicable
...

## Env Config
Not applicable

...

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/backend/backend/public_urls.py
+++ b/backend/backend/public_urls.py
@@ -62,5 +62,3 @@ try:
     ]
 except ImportError:
     pass
-except AttributeError:
-    pass
--- a/backend/backend/public_urls.py
+++ b/backend/backend/public_urls.py
@@ -62,3 +62,5 @@ try:
     ]
 except ImportError:
     pass
+except AttributeError:
+    pass
--- a/backend/backend/public_urls.py
+++ b/backend/backend/public_urls.py
@@ -22,7 +22,6 @@ from django.urls import include, path
 
 path_prefix = settings.PATH_PREFIX
 api_path_prefix = settings.API_DEPLOYMENT_PATH_PREFIX
-share_path_prefix = settings.PUBLIC_PATH_PREFIX
 
 urlpatterns = [
     path(f"{path_prefix}/", include("account.urls")),
@@ -50,6 +49,8 @@ urlpatterns += static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
 try:
     import pluggable_apps.platform_admin.urls  # noqa # pylint: disable=unused-import
     import pluggable_apps.public_shares_share_controller.urls  # noqa # pylint: disable=unused-import
+
+    share_path_prefix = settings.PUBLIC_PATH_PREFIX
 
     urlpatterns += [
         path(f"{path_prefix}/", include("pluggable_apps.platform_admin.urls")),
--- a/backend/sample.env
+++ b/backend/sample.env
@@ -143,6 +143,3 @@ CELERY_BROKER_URL = "redis://unstract-redis:6379"
 
 # Indexing flag to prevent re-index
 INDEXING_FLAG_TTL=1800
-
-#Publicly whitelisted URL path_prefix
-PUBLIC_PATH_PREFIX=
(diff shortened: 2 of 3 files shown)


backend/backend/public_urls.py
```diff
@@ -22,7 +22,6 @@
 
 path_prefix = settings.PATH_PREFIX
 api_path_prefix = settings.API_DEPLOYMENT_PATH_PREFIX
-share_path_prefix = settings.PUBLIC_PATH_PREFIX
```
ritwik-g, line 25: @harini-venkataraman what is the actual use of this prefix?

Deepak-Kesavan: Same thought. Any reason to keep it separate? Why can't we just use existing prefix `/api/v1` ?

Deepak-Kesavan (approved):

backend/sample.env
```diff
@@ -143,3 +143,6 @@ CELERY_BROKER_URL = "redis://unstract-redis:6379"
 
 # Indexing flag to prevent re-index
 INDEXING_FLAG_TTL=1800
+
+#Publicly whitelisted URL path_prefix
+PUBLIC_PATH_PREFIX=
```
chandrasekharan-zipstack, line 148:
```suggestion
PUBLIC_PATH_PREFIX=public
```
Also @harini-venkataraman why do we need to define an env which is not used in OSS code?

backend/backend/public_urls.py
```diff
@@ -51,6 +50,8 @@
     import pluggable_apps.platform_admin.urls  # noqa # pylint: disable=unused-import
     import pluggable_apps.public_shares_share_controller.urls  # noqa # pylint: disable=unused-import
 
+    share_path_prefix = settings.PUBLIC_PATH_PREFIX
```
chandrasekharan-zipstack, line 53: @harini-venkataraman what if this is not defined in the settings file (`base.py` or in the cloud version) - is the error handled gracefully?

backend/backend/public_urls.py
jaseemjaskp: as my understanding it is not an api. that why we can't keep it /api/v1. @harini-venkataraman correct me if I am wrong.

harini-venkataraman: This is a public route and it is whitelisted for the purpose of sharing pages publicly. We cannot use `/api/v1` since it is private route. This public route is added to the list of whitelisted paths and this does not need any authentication. We will need this to surpass authentication.

ritwik-g: @harini-venkataraman for the time being we can go with this. But I think this is not really necessary. We can make authentication exceptions even if the api starts with `/api/v1`. Please check with @muhammad-ali-e regarding this.

harini-venkataraman: @ritwik-g Also, one more thing to note here is, this URL is public, and by product we have `/public/promptStudio` appeneded with the ID, as the link which is displayed and copied to share the project. So, if we stick to `/public` prefix, it will be much self explanatory to the external world.

ritwik-g: Okay

backend/sample.env
harini-venkataraman: Explicitly we don`t use this prefix in OSS. This was added to stop OSS failing altogether. Have handled the exception and removed it.

backend/backend/public_urls.py
harini-venkataraman: @chandrasekharan-zipstack Have habdled the AttributeError in public_urls.py. If the value is not defined in settings file ( in case of OSS ) we supress the exception and in case of cloud, we have the value added in `production.py`

backend/backend/public_urls.py
```diff
@@ -61,3 +62,5 @@
     ]
 except ImportError:
     pass
+except AttributeError:
+    pass
```
chandrasekharan-zipstack, line 66:
```suggestion
```
Removing this to not suppress a necessary error that needs to be shown

chandrasekharan-zipstack (approved):

PR #488, opened 2024-07-16
fix: Update version to specific release tag

hari-kuriakose:
## What

* Fix release tag for version update via run platform script
* Rename version update command
* Improve readability of script output
* Update docstring

## Why

Users should be landing in stable release tags only by default.

## How

- Checkout to tag instead which corresponds to input release version
- Checkout to latest tag automatically when no input release version or set as `latest`
- Checkout to latest `main` branch and force build from local when input release version set as `main`

- No, concerns only with deployment of the platform.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/docker/scripts/merge_env.py
+++ b/docker/scripts/merge_env.py
@@ -54,8 +54,8 @@ def _merge_to_env_file(base_env_file_path: str, target_env: dict[str, str] = {})
     target env.
 
     Args:
-        base_env_path (string): Base env file path.
-        target_env (dict, optional): Target env to use for merge.
+        base_env_path (string): Base env file path e.g. `sample.env`
+        target_env (dict, optional): Target env to use for merge e.g. `.env`
 
     Returns:
         string: File contents after merge.
--- a/docker/scripts/merge_env.py
+++ b/docker/scripts/merge_env.py
@@ -54,7 +54,7 @@ def _merge_to_env_file(base_env_file_path: str, target_env: dict[str, str] = {})
     target env.
 
     Args:
-        base_env_path (string): Base env file path e.g. `sample.env`
+        base_env_file_path (string): Base env file path e.g. `sample.env`
         target_env (dict, optional): Target env to use for merge e.g. `.env`
 
     Returns:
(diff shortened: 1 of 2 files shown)


run-platform.sh
```diff
@@ -126,21 +126,40 @@ parse_args() {
...
 
-  echo -e "Performing git switch to ""$blue_text""main branch""$default_text".
-  git switch main
+  echo "Fetching release tags."
+  git fetch
 
-  echo -e "Performing ""$blue_text""git pull""$default_text"" on main branch."
-  git pull
+  if [[ "$opt_version" == "latest" ]]; then
+    branch=`git describe --tags --abbrev=0`
```
chandrasekharan-zipstack, line 143: @hari-kuriakose tried this from my feature branch 
image
and from main (without a `git pull`)
image
Result from main after a `git pull`
image


The version selected in the first 2 cases was `v0.73.0`
1. It didn't select the latest tag of `v0.74.0` without a `git pull`
1. It didn't select the patch version releases of `v0.73.2`

Is this the intended behaviour of the `latest` option now?

chandrasekharan-zipstack (approved):
LGTM - left a comment regarding the intended behaviour

jaseemjaskp (approved):

run-platform.sh
hari-kuriakose: @chandrasekharan-zipstack Good find.

`git fetch` by default only pulls those tags reference by the current branch. Fixed this by explicitly fetching all tags always via `git fetch --quiet --tags`.

docker/scripts/merge_env.py
```diff
@@ -54,8 +54,8 @@ def _merge_to_env_file(base_env_file_path: str, target_env: dict[str, str] = {})
     target env.
 
     Args:
-        base_env_path (string): Base env file path.
-        target_env (dict, optional): Target env to use for merge.
+        base_env_path (string): Base env file path e.g. `sample.env`
```
muhammad-ali-e, line 57:
```suggestion
        base_env_file_path (string): Base env file path e.g. `sample.env`
```

PR #491, opened 2024-07-16
Shared tool export fix

johnyrahul:
## What

- https://zipstack.atlassian.net/jira/software/c/projects/UN/issues/UN-1449?

## Why

- Failing with integrity exception. Prompt registry with Custom Tool is one-to-one related.
- So while trying to save the prompt registry with different modified by record its string insert a new promo registry entry with an existing custom tool which breaks the one-to-one relation.

## How

- Worked around with the database interaction.

- No

## Database Migrations

-  None

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

- Tried tool export with the owner as well as the shared user.

## Checklist

### DIFF
--- a/backend/prompt_studio/prompt_studio_registry/prompt_studio_registry_helper.py
+++ b/backend/prompt_studio/prompt_studio_registry/prompt_studio_registry_helper.py
@@ -176,7 +176,6 @@ class PromptStudioRegistryHelper:
             obj, created = PromptStudioRegistry.objects.update_or_create(
                 custom_tool=custom_tool,
                 created_by=custom_tool.created_by,
-                modified_by=custom_tool.modified_by,
                 defaults={
                     "name": custom_tool.tool_name,
                     "tool_property": properties.to_dict(),
@@ -190,7 +189,7 @@ class PromptStudioRegistryHelper:
                 logger.info(f"PSR {obj.prompt_registry_id} was created")
             else:
                 logger.info(f"PSR {obj.prompt_registry_id} was updated")
-
+            obj.modified_by = custom_tool.modified_by
             obj.shared_to_org = shared_with_org
             if not shared_with_org:
                 obj.shared_users.clear()


backend/prompt_studio/prompt_studio_registry/prompt_studio_registry_helper.py
```diff
@@ -190,7 +189,7 @@ def update_or_create_psr_tool(
                 logger.info(f"PSR {obj.prompt_registry_id} was created")
             else:
                 logger.info(f"PSR {obj.prompt_registry_id} was updated")
-
+            obj.modified_by = custom_tool.modified_by
```
chandrasekharan-zipstack, line 192: @johnyrahul how is this different from setting it in the `update_or_create()` method? I didn't get full context of the issue, however @harini-venkataraman please review if a similar pattern is observed anywhere else to fix that as well

johnyrahul: How the Django framework determines whether it should be updated or created is based on the fields supplied.
So when the shared user exports the tool  `modfied_by` file will be different from the owner. Hence it decides to create a new record instead of an update, which will generate an integrity issue
 @chandrasekharan-zipstack  makes sense?

chandrasekharan-zipstack: @harini-venkataraman then I suggest to look out for `update_or_create()` and `get_or_create()`

chandrasekharan-zipstack (approved):

backend/prompt_studio/prompt_studio_registry/prompt_studio_registry_helper.py
harini-venkataraman: @chandrasekharan-zipstack We mainly use it in account related logic and Output manager. Since this involves users, Integrity error is raised. Anyway will have to look out for other references.

backend/prompt_studio/prompt_studio_registry/prompt_studio_registry_helper.py
```diff
@@ -176,7 +176,6 @@ def update_or_create_psr_tool(
             obj, created = PromptStudioRegistry.objects.update_or_create(
                 custom_tool=custom_tool,
                 created_by=custom_tool.created_by,
```
harini-venkataraman, line 178: @johnyrahul Will created_by field cause an IntegrityError ?

harini-venkataraman (approved):
LGTM, added a comment.

backend/prompt_studio/prompt_studio_registry/prompt_studio_registry_helper.py
johnyrahul: no @harini-venkataraman . The field will always be fixed for custom tools.

PR #496, opened 2024-07-17
fix: Test connector error handling improvements

chandrasekharan-zipstack:
## What

- Adding context to error messages indicating which connector its from
- Handled few S3 / MinIO errors
- Minor refactor around test connection - renamed variables for readability

## TODO
- Extend error handling for other connectors similarly to improve the messages sent to the FE
- Split MinIO and S3 into 2 different connectors

## Why

- Issues in error handling during test connection calls

## How

- Caught and re-raised connector related exceptions with some context.

- No, its around error handling during test connection calls

## Dependencies Versions

- upgraded `dropboxdrivefs` to 1.4.1 -> helps address an issue with listing files

## Notes on Testing

- Tested S3 and Dropbox explicitly. Other connectors were not explicitly tested

## Screenshots
- S3 related
image
image
Few errors which are not explicitly converted show up like below
image
image

- Dropbox related
image
image

## Checklist

### DIFF
--- a/unstract/connectors/src/unstract/connectors/filesystems/google_drive/google_drive.py
+++ b/unstract/connectors/src/unstract/connectors/filesystems/google_drive/google_drive.py
@@ -91,17 +91,18 @@ class GoogleDriveFS(UnstractFileSystem):
 
     def test_credentials(self) -> bool:
         """To test credentials for Google Drive."""
+        is_dir = False
         try:
             is_dir = bool(self.get_fsspec_fs().isdir("root"))
-            if not is_dir:
-                raise RuntimeError(
-                    "Could not connect to its root directory 'root', "
-                    "please recheck the connection settings."
-                )
         except Exception as e:
             raise ConnectorError(
-                f"Error while connecting to Google Drive: {str(e)}"
+                f"Error from Google Drive while testing connection: {str(e)}"
             ) from e
+        if not is_dir:
+            raise ConnectorError(
+                "Unable to connect to Google Drive, "
+                "please check the connection settings."
+            )
         return True
 
     @staticmethod
--- a/unstract/connectors/src/unstract/connectors/filesystems/google_drive/google_drive.py
+++ b/unstract/connectors/src/unstract/connectors/filesystems/google_drive/google_drive.py
@@ -92,9 +92,16 @@ class GoogleDriveFS(UnstractFileSystem):
     def test_credentials(self) -> bool:
         """To test credentials for Google Drive."""
         try:
-            self.get_fsspec_fs().isdir("root")
+            is_dir = bool(self.get_fsspec_fs().isdir("root"))
+            if not is_dir:
+                raise RuntimeError(
+                    "Could not connect to its root directory 'root', "
+                    "please recheck the connection settings."
+                )
         except Exception as e:
-            raise ConnectorError(str(e))
+            raise ConnectorError(
+                f"Error while connecting to Google Drive: {str(e)}"
+            ) from e
         return True
 
     @staticmethod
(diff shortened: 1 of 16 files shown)


unstract/connectors/src/unstract/connectors/filesystems/google_drive/google_drive.py
```diff
@@ -92,9 +92,16 @@ def get_fsspec_fs(self) -> GDriveFileSystem:
     def test_credentials(self) -> bool:
         """To test credentials for Google Drive."""
         try:
-            self.get_fsspec_fs().isdir("root")
+            is_dir = bool(self.get_fsspec_fs().isdir("root"))
+            if not is_dir:
+                raise RuntimeError(
+                    "Could not connect to its root directory 'root', "
```
harini-venkataraman, line 98: @chandrasekharan-zipstack In cases of error, should we actually expose that we are trying to connect to the `root` directory in our test connection? Can the error message be more generic ?

harini-venkataraman (approved):
LGTM, left a minor comment.

unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
```diff
@@ -0,0 +1,26 @@
...
+def handle_s3fs_exception(e: Exception) -> ConnectorError:
+    original_exc = str(e)
+    user_msg = "Error while connecting to S3 / MinIO: "
+    exc_to_append = ""
+    for s3fs_exc, user_friendly_msg in S3FS_EXC_TO_UNSTRACT_EXC.items():
+        if s3fs_exc in original_exc:
+            exc_to_append = user_friendly_msg
+            break
+
+    user_msg += exc_to_append if exc_to_append else f"Error from S3, '{str(e)}'"
```
johnyrahul, line 25:
```suggestion
    user_msg += exc_to_append if exc_to_append else str(e)
```

johnyrahul (approved):
LGTM. Kept one suggestion.

unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
johnyrahul: `Error from S3` Will get duplicated for unhandled cases.

kirtimanmishrazipstack (approved):
Looks good, I would left @johnyrahul 's comment to resolve

unstract/connectors/src/unstract/connectors/filesystems/google_drive/google_drive.py
chandrasekharan-zipstack: I had a similar doubt as well since its internal to the connector itself. You're right - it won't make sense to the user and doesn't help debug the issue better. Will refrain from mentioning it

unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
chandrasekharan-zipstack: I added it to convey that the error is from the connector's side - you're right duplicating that doesn't look good. Will update the original prefix itself in this scenario

unstract/connectors/src/unstract/connectors/filesystems/google_drive/google_drive.py
chandrasekharan-zipstack: Also I'm not sure if this particular message will ever get thrown to the user - I just added it as a fail safe. Usually connection failures get raised as exceptions from the connector clients

harini-venkataraman: In that case, if it improves developer debugging experience, can we log it?

PR #499, opened 2024-07-18
[FIX] Made workflow to use latest tool version

Deepak-Kesavan:
## What

Made workflow to use latest tool version instead if the version which was specified in the backend env at the time of export

## Why

This is causing some issues when there are changes in `prompt-service`

## How

Use the tool version directly from Django settings instead of tool metadata.

No. It won't break existing features. Whoever exported the tool when the tool version was 0.0.20 can break and re-export will fix it. Moving forward we need to make sure to support backward compatibility in `structure tool` with the data exported with the previous versions of structure tool. The existing workflow will use the latest tool version specified in the env file. This will help us have control of changing tool version in the env if it is a breaking tool version.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


harini-venkataraman (approved):

backend/prompt_studio/prompt_studio_registry/prompt_studio_registry_helper.py
```diff
@@ -119,25 +119,14 @@ def get_tool_by_prompt_registry_id(
...
-        )
         return Tool(
             tool_uid=prompt_registry_tool.prompt_registry_id,
             properties=Properties.from_dict(prompt_registry_tool.tool_property),
             spec=Spec.from_dict(prompt_registry_tool.tool_spec),
             icon=prompt_registry_tool.icon,
-            image_url=image_url,
-            image_name=image_name,
-            image_tag=image_tag,
+            image_url=settings.STRUCTURE_TOOL_IMAGE_URL,
```
harini-venkataraman, line 127: @Deepak-Kesavan QQ : Will his throw an error f the config variable is missing?

harini-venkataraman:
LGTM, left a minor comment.

backend/prompt_studio/prompt_studio_registry/prompt_studio_registry_helper.py
Deepak-Kesavan: It can't be missing. It is the env we set in .env of the backend.

harini-venkataraman: I mean if the variable is not set in the env will it throw an error?

harini-venkataraman: I mean if the variable is not set in the env will it throw an error?

gaya3-zipstack (approved):
In the context of my understanding, we are always enforcing to pull the images as mentioned in .env instead of defaulting to 0.0.20, Changes look fine in that perspective.

backend/prompt_studio/prompt_studio_registry/prompt_studio_registry_helper.py
Deepak-Kesavan: The backend won't even boot up if it was not there. Check this for additional information.

https://github.com/Zipstack/unstract/blob/7c513ff86b2a7b28f06fafbdd1760b77d4003594/backend/backend/settings/base.py#L148-L150C68

PR #502, opened 2024-07-19
[FIX] Cost calculation fix for LLMs and Embeddings

Deepak-Kesavan:
## What

Calculate cost based on the model name and provider

## Why

To know the exact cost

## How

Check for the model name along with provider in the model prices JSON because same model can be provided by multiple LLM providers

No. Just added additional logic to calculate cost. It won't break existing features.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

- https://github.com/Zipstack/unstract-adapters/pull/64
- https://github.com/Zipstack/unstract-sdk/pull/75

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


gaya3-zipstack (approved):
Changes look inline with the discussion and other associated PRs.

gaya3-zipstack (approved):
Version updates look fine

platform-service/src/unstract/platform_service/helper.py
```diff
@@ -109,12 +109,19 @@ def __init__(
...
-    def calculate_cost(self, model_name, input_tokens, output_tokens):
+    def calculate_cost(self, model_name, provider, input_tokens, output_tokens):
         cost = 0.0
         item = None
         model_prices = self._get_model_prices()
-        if model_prices:
-            item = model_prices.get(model_name)
+        # Filter the model objects by model name
+        filtered_models = {
+            k: v for k, v in model_prices.items() if k.endswith(model_name)
```
chandrasekharan-zipstack, line 118: @Deepak-Kesavan why do we check with `k.endswith(model_name` instead of doing a direct match?

chandrasekharan-zipstack (approved):

platform-service/src/unstract/platform_service/helper.py
Deepak-Kesavan: If the same model is provided by multiple LLM providers, then the pattern will be like `provider/model-name`. That's why

PR #506, opened 2024-07-22
fix/persisting-unique-id-on-db-connector

kirtimanmishrazipstack:
## What

- Adding unique id in db connector
- Handle error scenarios in-case column doesn't exists for user connector databases

## Why

- To uniquely identify connector data

## How

- Adding `execution_id` and `file_path` to output db connector
- Add error scenarios if  column doesn't exists

- NO

## Database Migrations

-  NO

## Env Config

- NO

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

- Tests workflow with the changes in db connector
- Checked error scenarios for all connectors

## Screenshots
Screenshot from 2024-07-23 08-23-31
Screenshot from 2024-07-23 08-23-27
Screenshot from 2024-07-23 15-13-07
Screenshot from 2024-07-25 13-54-15

## Checklist


backend/workflow_manager/endpoint/constants.py
```diff
@@ -1,7 +1,7 @@
 class TableColumns:
     CREATED_BY = "created_by"
     CREATED_AT = "created_at"
-    PERMANENT_COLUMNS = ["created_by", "created_at"]
+    PERMANENT_COLUMNS = ["created_by", "created_at", "file_path", "execution_id"]
```
chandrasekharan-zipstack, line 4: @kirtimanmishrazipstack we agreed that we'll allow these column names to be modifiable by the user similar to the `data` column. Can you make that change?

kirtimanmishrazipstack: @chandrasekharan-zipstack done

backend/workflow_manager/endpoint/static/dest/db.json
```diff
@@ -36,6 +36,18 @@
...
+            "title": "File Path",
+            "default": "file_path",
+            "description": "File path to store the file address"
+        },
+        "executionId": {
+            "type": "string",
+            "title": "Execution Id",
+            "default": "execution_id",
+            "description": "Execution id of the workflow"
+        },
```
chandrasekharan-zipstack, line 50:
```suggestion
        "filePath": {
            "type": "string",
            "title": "File Path Column Name",
            "default": "file_path",
            "description": "Name of the column to store the absolute path to the file"
        },
        "executionId": {
            "type": "string",
            "title": "Execution ID Column Name",
            "default": "execution_id",
            "description": "Name of the column that stores the execution ID which identifies each individual run of a workflow"
        },
```

chandrasekharan-zipstack (approved):

backend/workflow_manager/endpoint/static/dest/db.json
kirtimanmishrazipstack: @chandrasekharan-zipstack changes done

athul-rs:
@kirtimanmishrazipstack @chandrasekharan-zipstack One doubt here, will this cause any issue for previous data stored in db, assuming here we are now adding `input_file_path` which won't be present for prior data ?

```
            self.insert_into_db(
                file_history=file_history, input_file_path=input_file_path
            )
```

johnyrahul:
@kirtimanmishrazipstack  FE screenshot seems to be outdated?

kirtimanmishrazipstack:
> insert_into_db



> @kirtimanmishrazipstack FE screenshot seems to be outdated?

@johnyrahul new FE changes are yet to come. This backend change will comprehend the new FE changes too. I have just changes `db.json`

kirtimanmishrazipstack:
> @kirtimanmishrazipstack @chandrasekharan-zipstack One doubt here, will this cause any issue for previous data stored in db, assuming here we are now adding `input_file_path` which won't be present for prior data ?
> 
> ```
>             self.insert_into_db(
>                 file_history=file_history, input_file_path=input_file_path
>             )
> ```

@athul-rs I have just used `input_file_path` to store in db. Not tinkered `file_history`. It should not break anything.

kirtimanmishrazipstack:
@johnyrahul @athul-rs @chandrasekharan-zipstack 
As discussed we decided to not go for any migration on the user table. The user can create a separate table and use the db-connector

chandrasekharan-zipstack:
Thanks for pointing it out @athul-rs @johnyrahul.
@kirtimanmishrazipstack what will the behaviour be for existing users who use this connector though? Ideally it should either just work for existing tables or if an error is thrown, we should clearly mention that its because these columns are missing and they need to add it. Please ensure this is checked / handled as well

chandrasekharan-zipstack (approved):

johnyrahul (approved):

PR #507, opened 2024-07-22
Handing API deployment errors

johnyrahul:
## What

- In API deployment response  show meaningful/actual error messages rather than  showing fixed static msg 
`Tool exception raised for tool  check container logs`

## Why

- With static error messages, reporting and debugging issues can become difficult.

## How

- Passing down the container error using the stream log option to the worker and then back to unstract backend

-

## Database Migrations

-  None

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

- https://github.com/Zipstack/unstract-sdk/pull/76

## Dependencies Versions

-

## Notes on Testing

- Tested with unhandled exception in  adapter side.

## Screenshots
image
image
image
image
****

## Checklist

- [x]  https://github.com/Zipstack/unstract-sdk/pull/76 Needs to be merged


worker/src/unstract/worker/worker.py
```diff
@@ -79,6 +80,9 @@ def process_log_message(
         if not log_dict:
             return None
         log_type = self.get_log_type(log_dict)
+        log_level = self.get_log_level(log_dict)
+        if log_type == LogType.LOG and log_level == LogLevel.ERROR:
+            raise ToolRunException(log_dict.get("log"))
```
chandrasekharan-zipstack, line 85: NIT: As discussed this might prevent us from using `ERROR` logs without exiting in the context of tools and SDK - we need to make sure we never catch and suppress an exception with ERROR logs and instead using WARNING instead. We could address this when necessary maybe

chandrasekharan-zipstack (approved):
LGTM - how do you plan to build the tools? From this feature branch?

gaya3-zipstack (approved):
I am not fully aware of the issue. From the context of handling errors while running tools, changes look fine.

worker/src/unstract/worker/worker.py
johnyrahul: Agreed @chandrasekharan-zipstack.
We might have to specify a proper identifier for stream_error_exit scenarios. Will take this up when such a requirement arises, should be quick.

PR #513, opened 2024-07-23
[FIX] Public share fixes

harini-venkataraman:
## What
This PR focuses on stabilizing Prompt studio public share.
...

## Why
Fixing the prompt studio public share issues.
...

## How
Involves, disabling of enterprise features and exposing a few newer APIs.
...

No, it has changes on public share.
...

## Relevant Docs
Not Applicable.
...

## Related Issues or PRs
https://github.com/Zipstack/unstract/pull/513
...

Not Applicable.

...

## Notes on Testing
The same flow for testing prompt studio, with the newly added Share and clone buttons.
...

## Screenshots
Screenshot from 2024-07-23 21-08-41

Screenshot from 2024-07-23 21-06-09

Screenshot from 2024-07-23 20-58-14

...

## Checklist


backend/prompt_studio/prompt_studio_output_manager/output_manager_helper.py
```diff
@@ -144,3 +145,37 @@ def fetch_default_llm_profile(tool: CustomTool) -> ProfileManager:
             return ProfileManager.get_default_llm_profile(tool=tool)
         except DefaultProfileError:
             raise DefaultProfileError("Default ProfileManager does not exist.")
+
+    @staticmethod
+    def fetch_default_response(
```
chandrasekharan-zipstack, line 150: NIT: @harini-venkataraman naming it as `fetch_default_response` leaves room for doubt and I can't figure out what this function does without checking its code. I suggest renaming it to make it more explicit or adding a docstring describing this

chandrasekharan-zipstack (approved):
BE changes LGTM - left some questions and nit comments

backend/prompt_studio/prompt_studio_output_manager/output_manager_helper.py
```diff
@@ -144,3 +145,37 @@ def fetch_default_llm_profile(tool: CustomTool) -> ProfileManager:
...
+    ) -> dict[str, Any]:
+        # Initialize the result dictionary
+        result: dict[str, Any] = {}
+        # Iterate over ToolStudioPrompt records
+        for tool_prompt in tool_studio_prompts:
+            prompt_id = str(tool_prompt.prompt_id)
+            profile_manager_id = str(tool_prompt.profile_manager.profile_id)
+
+            # If profile_manager is not set, skip this record
+            if not profile_manager_id:
```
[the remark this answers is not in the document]
jagadeeswaran-zipstack: @chandrasekharan-zipstack when we delete a default profile of prompt then it will be null. Also when user try to use combined output only creating prompt and without profile, in that case this check needed.

backend/prompt_studio/prompt_studio_output_manager/output_manager_helper.py
```diff
@@ -144,3 +145,37 @@ def fetch_default_llm_profile(tool: CustomTool) -> ProfileManager:
...
+                    document_manager_id=document_manager_id,
+                )
+
+                if not queryset.exists():
+                    result[tool_prompt.prompt_key] = ""
+                    continue
+
+                for output in queryset:
+                    result[tool_prompt.prompt_key] = output.output
+            except ObjectDoesNotExist:
```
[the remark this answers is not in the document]
harini-venkataraman: @chandrasekharan-zipstack Yes, even I had this doubt. @jagadeeswaran-zipstack was there any specific use case to add this logic? Also, I doubt `MultipleObjectsReturned` will be raised here, in case of multiple objects fetched. Should we use `get` instead of `filter`?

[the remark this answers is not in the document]
jagadeeswaran-zipstack: We dont need that additional check there. Ill address that in my PR

backend/prompt_studio/prompt_studio_output_manager/output_manager_helper.py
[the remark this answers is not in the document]
chandrasekharan-zipstack: @jagadeeswaran-zipstack 
1. When we enforce a constraint that the user can upload documents / add prompts only after adding a profile which is marked as default - we should not let them delete the default profile at all. Please ensure we note / handle this case elegantly
1. I don't clearly understand this combined output case you're mentioning - is it single pass? Ideally when a prompt is being added without the user explicitly mentioning any profile - we should take the default profile and use that.

This will ensure that we always have a profile associated with a prompt eliminating the need for such checks in all our logic

backend/prompt_studio/prompt_studio_output_manager/output_manager_helper.py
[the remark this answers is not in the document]
chandrasekharan-zipstack: I don't have full context on the structure of objects here, but here's the [docs for get()](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#get) - please go through and validate if we're doing what's intended @harini-venkataraman @jagadeeswaran-zipstack

frontend/src/components/custom-tools/manage-docs-modal/ManageDocsModal.jsx
```diff
@@ -263,6 +275,7 @@ function ManageDocsModal({
         handleIndexStatus(indexType, indexStatus);
       })
       .catch((err) => {
+        console.log(err);
```
Deepak-Kesavan, line 278: Please remove the log here.

frontend/src/components/custom-tools/output-analyzer-header/OutputAnalyzerHeader.jsx
```diff
@@ -3,29 +3,43 @@ import { ArrowLeftOutlined } from "@ant-design/icons";
...
 
 function OutputAnalyzerHeader() {
   const { sessionDetails } = useSessionStore();
   const { id } = useParams();
   const navigate = useNavigate();
-
+  const { isPublicSource } = useCustomToolStore();
+  const handleNavigate = () => {
+    if (isPublicSource) {
+      navigate(-1);
```
jagadeeswaran-zipstack, line 23: Incase few cases user might access directly through exact url in those cases there will be no previous route. So this will cause error. Always specify the exact route that need to be navigated.

jagadeeswaran-zipstack (changes requested):

Deepak-Kesavan (commented):
Looks good. Added minor comment

frontend/src/components/custom-tools/output-analyzer-card/OutputAnalyzerCard.jsx
```diff
@@ -11,26 +11,38 @@ import { useSessionStore } from "../../../store/session-store";
...
   const [fileUrl, setFileUrl] = useState("");
   const [isDocLoading, setIsDocLoading] = useState(false);
   const [filledFields, setFilledFields] = useState(0);
-  const { details } = useCustomToolStore();
+  const { details, isPublicSource } = useCustomToolStore();
   const { sessionDetails } = useSessionStore();
   const { setAlertDetails } = useAlertStore();
   const axiosPrivate = useAxiosPrivate();
   const handleException = useExceptionHandler();
+  const { id } = useParams();
```
tahierhussain, line 32: @harini-venkataraman You can get the tool ID from `details` (inside the custom tool store). Can you please replace its usage with `details?.tool_id` here and wherever applicable? This way, we can avoid an unnecessary import.

frontend/src/components/custom-tools/tool-ide/ToolIde.jsx
```diff
@@ -206,14 +213,15 @@ function ToolIde() {
...
-    });
+    if (!isPublicSource) {
+      handleUpdateTool(body).catch((err) => {
+        const revertSelectedDoc = {
+          selectedDoc: prevSelectedDoc,
+        };
+        updateCustomTool(revertSelectedDoc);
+        setAlertDetails(handleException(err, "Failed to select the document"));
+      });
+    }
```
tahierhussain, line 224: @harini-venkataraman Instead of having the update tool function call inside the condition, a better approach would be to return if it's a public source on line 212 of the same function.

frontend/src/components/custom-tools/output-analyzer-card/OutputAnalyzerCard.jsx
harini-venkataraman: @tahierhussain ID here refers to the `share_id` for public prompt studio. Is it okay to fetch it from params here?

jagadeeswaran-zipstack (approved):
LGTM

frontend/src/components/custom-tools/output-analyzer-card/OutputAnalyzerCard.jsx
tahierhussain: Sure, @harini-venkataraman. If we're using the tool_id anywhere from the params, it can be taken from the `details` object. Resolving the comment.

tahierhussain (approved):
LGTM

hari-kuriakose:
Seems like it is not directly possible to re-trigger status checks. Bypassing for now.

PR #514, opened 2024-07-23
fix: Improvements in run platform script

hari-kuriakose:
## What

Improvements in run platform script to make it more accessible.

## Why

- Latest docker compose CLI versions will throw a warning if compose version is specified.
  We can safely remove this as the CLI will figure out the compose file version automatically.
  The user might need to upgrade docker compose CLI version in rare cases, which also should be ok.
- Docker proxy override information will only help contributors, hence removed to avoid confusion.
- Encryption key if lost or changed will make all existing adapters inaccessible, hence need to warn the user to back it up when setting up the first time.

## How

* Remove docker compose version directive
* Remove docker proxy override details from run script
* Display warning to backup encryption key on first time setup

- No.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Screenshots

backup_encryption_key

## Checklist


chandrasekharan-zipstack (approved):
LGTM - optionally we could check for a minimum required compose version if the error due to an an older version is not clear. I remember the `include` support for docker compose was introduced relatively recently

ritwik-g (approved):

ritwik-g:
@hari-kuriakose I recall @chandrasekharan-zipstack suggesting some change regarding updating tool version in `backend.env` with version from tools json. Should we include that change as well here?

hari-kuriakose:
> @hari-kuriakose I recall @chandrasekharan-zipstack suggesting some change regarding updating tool version in `backend.env` with version from tools json. Should we include that change as well here?

@ritwik-g Yes, had taken a look at the merge env script. It is already written in such a way where updates in `sample.env` for a particular release tag is automatically reflected in the target `.env`. 

@chandrasekharan-zipstack Is this enough?

PR #521, opened 2024-07-25
fix: Perform checks in classifier after LLM response, bumped to 0.0.29

chandrasekharan-zipstack:
## What

- Bumped classifier to 0.0.29
- Refactored classifier code to perform checks after LLM response
- Updated spec description for the classification bins field
- Ensure we copy the files to folder `unknown` or `__unstract_failed` in case of unknown classification (if LLM returns it or if we deduce that multiple classifications are possible) and in case of failure respectively
- MINOR: Removed some unused code in `runtime_variables.json`

- Bumped SDK version to 0.39.0 in structure tool and text extractor tool

## Why

- Existing classifier is not robust enough to handle all cases of LLM response

## How

- Since we removed the kwargs to limit the LLM response, it might provide an explanation for the classification
- We compare the first 100 words and assume that it would have returned the correct classification within the first 100 words of its response. In the chance that it responds with more than 1 classification - we safely treat it as `unknown`
- In case of failures, we move the file to a separate folder

- No, its purely within the classifier. It might break the existing behaviour in the context of copying files - which will be tested in staging

## Dependencies Versions

- Bumped to use SDK 0.39.0

## Notes on Testing

- Tested the tool locally with several LLMs - however the copy to functionality was not explicitly tested.

## Screenshots
image

## Checklist

### DIFF
--- a/tools/classifier/requirements.txt
+++ b/tools/classifier/requirements.txt
@@ -1,4 +1,4 @@
 # Add your dependencies here
 
 # Required for all unstract tools
-unstract-sdk~=0.38.1
+unstract-sdk~=0.39.0
--- a/tools/classifier/src/config/properties.json
+++ b/tools/classifier/src/config/properties.json
@@ -2,7 +2,7 @@
   "schemaVersion": "0.0.1",
   "displayName": "File Classifier",
   "functionName": "classify",
-  "toolVersion": "0.0.28",
+  "toolVersion": "0.0.29",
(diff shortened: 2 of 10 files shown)


harini-venkataraman (approved):
LGTM

gaya3-zipstack (approved):
Changes look fine

Deepak-Kesavan (approved):

PR #530, opened 2024-07-26
fix: File type support for WF and API deployments

chandrasekharan-zipstack:
## What

- Case insensitive matching for different file types for workflow / API run
- Added support for `.doc` and `.docx`
- Fixed structure tool to allow all file types - check made in `backend` alone, bumped it to `0.0.34`
- MINOR: Allow update of structure tool version through `merge_env.sh` script

## Why

- Different file types were not supported by the prompt studio exported tools

- No, changes within a tool. Other backend changes are related to case insensitive matching of file patterns

## Env Config
```
 STRUCTURE_TOOL_IMAGE_URL="docker:unstract/tool-structure:0.0.34"
STRUCTURE_TOOL_IMAGE_NAME="unstract/tool-structure"
STRUCTURE_TOOL_IMAGE_TAG="0.0.34"
```

## Notes on Testing

- Tested locally with WF and API run for different file types - PNG, JPG, TIFF, TXT

## Screenshots
image
image

## Checklist


Deepak-Kesavan (approved):

harini-venkataraman (approved):

docker/scripts/merge_env.py
```diff
@@ -10,10 +10,7 @@
 #
 import sys
 
-PREFERRED_BASE_ENV_KEYS = [
-    "STRUCTURE_TOOL_IMAGE_URL",
-    "STRUCTURE_TOOL_IMAGE_TAG",
-]
+PREFERRED_BASE_ENV_KEYS = []
```
hari-kuriakose, line 13: @chandrasekharan-zipstack This change needs to be reverted so that we always honor `sample.env` and use its contents to update `.env` each time.

chandrasekharan-zipstack: @hari-kuriakose we always update the `sample.env` with the latest structure tool version. However since we added these fields under `PREFERRED_BASE_ENV_KEYS` -> during the merge of envs, we don't override the value already present in the user's `.env`. As a result the user would always have an older version of structure tool present in their setup until and unless they manually check and update the value

PR #536, opened 2024-07-30
schema migrations and settings for v2 applications

muhammad-ali-e:
## What
- Implemented schema migrations for multi-tenancy.
- Updated settings in base.py.
- Added base_urls.py for base URL routing.
- Moved Public API deployment paths from other URLs to dedicated routes.

## Why

- To support Multi-tenancy V2 changes and improve project structure.

## How

-

- No, all changes are under feature flags, so they will not affect current functionality.

## Database Migrations

-  Includes migrations specific to multi-tenancy changes. under feature flag

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


backend/workflow_manager/workflow_v2/models/file_history.py
```diff
@@ -39,11 +39,12 @@ def is_completed(self) -> bool:
         db_comment="Error message",
     )
     result = models.TextField(blank=True, db_comment="Result from execution")
+    meta_data = models.TextField(blank=True, db_comment="MetaData from execution")
```
hari-kuriakose, line 42: @muhammad-ali-e Suggest to rename field to `metadata`.

backend/usage_v2/migrations/0001_initial.py
```diff
@@ -0,0 +1,144 @@
...
+
+    initial = True
+
+    dependencies = [
+        ("account_v2", "0001_initial"),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name="Usage",
```
hari-kuriakose, line 19: @muhammad-ali-e Suggest to keep the same name for the ORM model and corresponding db table.

hari-kuriakose (approved):
@muhammad-ali-e Migrations look ok overall.

Added few minor comments though.

Lastly, duplication can be taken care later when we remove v1.

backend/account_v2/subscription_loader.py
```diff
@@ -75,3 +76,32 @@ def load_plugins() -> list[Any]:
         logger.info("No subscription plugins found.")
 
     return subscription_plugins
+
+
+def validate_etl_run(org_id: str) -> bool:
```
chandrasekharan-zipstack, line 81: @muhammad-ali-e can this be moved to the cloud repo?

backend/backend/settings/base.py
```diff
@@ -185,66 +186,115 @@ def get_required_setting(
...
-    "prompt_studio.prompt_profile_manager",
-    "prompt_studio.prompt_studio",
-    "prompt_studio.prompt_studio_core",
-    "prompt_studio.prompt_studio_registry",
-    "prompt_studio.prompt_studio_output_manager",
-    "prompt_studio.prompt_studio_document_manager",
-    "prompt_studio.prompt_studio_index_manager",
-    "usage",
-)
+    SHARED_APPS = (
```
chandrasekharan-zipstack, line 190: @muhammad-ali-e why do we split as shared apps and tenant apps in v2 also? Won't they all be in a single schema?

backend/connector_v2/models.py
```diff
@@ -111,17 +111,42 @@ def __str__(self) -> str:
...
+        JSON string into a dictionary.
+
+        Returns:
+            dict: The decrypted connector metadata.
+        """
+
+        connector_metadata_bytes = self.get_connector_metadata_bytes()
+
+        if connector_metadata_bytes is None:
+            return None
```
chandrasekharan-zipstack, line 139: @muhammad-ali-e when can this be None? Let's handle errors when the metadata is NULL within the model itself rather than doing it at the caller side

backend/connector_v2/models.py
```diff
@@ -111,17 +111,42 @@ def __str__(self) -> str:
             f" workflow: {self.workflow})"
         )
 
+    def get_connector_metadata_bytes(self):
```
chandrasekharan-zipstack, line 114: @muhammad-ali-e why isn't this required for adapters?

backend/utils/user_context.py
```diff
@@ -37,6 +36,11 @@ def get_organization() -> Optional[Organization]:
                 )
             except Organization.DoesNotExist:
                 return None
+            except ProgrammingError:
+                # Handle cases where the database schema might not be fully set up,
+                # especially during the execution of management commands
+                # other than runserver
```
chandrasekharan-zipstack, line 42: @muhammad-ali-e should we log something informative here?

backend/workflow_manager/workflow_v2/views.py
```diff
@@ -116,16 +114,7 @@ def perform_create(self, serializer: WorkflowSerializer) -> Workflow:
...
             WorkflowEndpointUtils.create_endpoints_for_workflow(workflow)
-
-            # Enable GCS configurations to create GCS while creating a workflow
-            if (
-                settings.GOOGLE_STORAGE_ACCESS_KEY_ID
-                and settings.UNSTRACT_FREE_STORAGE_BUCKET_NAME
-            ):
-                ConnectorInstanceHelper.create_default_gcs_connector(
-                    workflow, self.request.user
-                )
```
chandrasekharan-zipstack, line 127: @muhammad-ali-e I don't think we use this but we might have still created it previously - do you think we'll need a migration to remove those? It might cause regressions if we have any future logic that deviates from what we've defined in it

backend/account_v2/subscription_loader.py
muhammad-ali-e: It also exists in V1, which is why I copied it here. I’m not fully aware of the complete context, but we can move it to the appropriate locations once the V2 migration is complete.

backend/backend/settings/base.py
muhammad-ali-e: This logic is actually for V1, not V2. The else condition handles the V2 logic. I temporarily split the shared apps and V2 apps to better understand the V2 applications. We will combine them later.

backend/connector_v2/models.py
muhammad-ali-e: In this case, I merged the binary and normal fields to avoid having two fields serving the same purpose. I tested some scenarios, but more testing is needed. For the adapter, I didn't merge them due to time constraints and the complexity of handling multiple cases. We can merge them there later if necessary. This function is specifically used to convert memoryview to bytes due to the introduction of the new field.

backend/connector_v2/models.py
muhammad-ali-e: I think, the current approach is safer because there may be scenarios where connector_metadata could be None, especially in existing cases that I am  not fully aware of it (We can check it later once the migrations are done). Handling it within the model ensures that we avoid potential errors at the caller side.

backend/utils/user_context.py
muhammad-ali-e: It should be a unenecessory logs during commands.

backend/workflow_manager/workflow_v2/views.py
muhammad-ali-e: I didn’t encounter any issues when removing it, and I’ve checked with some existing data as well. If any regressions occur, we can address them by adding this back. We should only run the migrations after thoroughly testing and confirming that there are no issues or regressions.

backend/workflow_manager/workflow_v2/models/file_history.py
muhammad-ali-e: Done

backend/usage_v2/migrations/0001_initial.py
muhammad-ali-e: renamed db table name to usage, Hence both model and table name have same name.

chandrasekharan-zipstack (approved):

backend/workflow_manager/workflow_v2/views.py
muhammad-ali-e: I think not using it. Want to check with @johnyrahul

muhammad-ali-e: @chandrasekharan-zipstack We are not using any default connector now.

athul-rs (approved):

hari-kuriakose:
Duplication will be taken care of when we remove v1 later.

PR #538, opened 2024-07-30
delta changes and enhancements  in platform and  prompt services

muhammad-ali-e:
## What

- Implemented enhancements and new changes from V1 to V2 for platform and prompt services.
-  implement new shema for v2
-  Changes on token usage method
-  Changes on Authentication methods

## Why

- To improve the performance and functionality of platform and prompt services.

## How

-

- No, All are uunder featureflag

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


hari-kuriakose (changes requested):
@muhammad-ali-e LGTM overall.

However, I suggest to rename new env var to `DB_SCHEMA`. Then use this env value if set, otherwise use default (from hereon we can keep a default schema name for multi-tenancy with single schema).

muhammad-ali-e:
> @muhammad-ali-e LGTM overall.
> 
> However, I suggest to rename new env var to `DB_SCHEMA`. Then use this env value if set, otherwise use default (from hereon we can keep a default schema name for multi-tenancy with single schema).

@hari-kuriakose I think it should confuce the OSS users if we use DB_SCHEMA. My opinion is we can  change the env variable to `DB_SCHEMA` once we completly moved to v2.

prompt-service/src/unstract/prompt_service/env_manager.py
```diff
@@ -0,0 +1,11 @@
+import os
+from typing import Optional
+
+
+class EnvLoader:
+    @staticmethod
+    def get_env_or_die(env_key: str, default: Optional[str] = None) -> str:
```
chandrasekharan-zipstack, line 7: NIT: We could rename this to `get_env_or_raise_err`

hari-kuriakose:
> > @muhammad-ali-e LGTM overall.
> > However, I suggest to rename new env var to `DB_SCHEMA`. Then use this env value if set, otherwise use default (from hereon we can keep a default schema name for multi-tenancy with single schema).
> 
> @hari-kuriakose I think it should confuce the OSS users if we use DB_SCHEMA. My opinion is we can change the env variable to `DB_SCHEMA` once we completly moved to v2.

@muhammad-ali-e If we are going to continue keeping this env var, then `DB_SCHEMA` would be better as users could supply any value as desired and not locked to v2 specifically.

However if this env var is to be used only temporarily for v2 migration, then ok.

prompt-service/src/unstract/prompt_service/env_manager.py
muhammad-ali-e: @chandrasekharan-zipstack Nothing changed here, Just moved the function here. About the renaming, Both seems same as per the functionality. Will be hard if any changes comes between the merge. So we can do the renaming later.

hari-kuriakose (approved):
@muhammad-ali-e LGTM overall.

We need to either rename the env var to `DB_SCHEMA` if we are going to continue having it from hereon or we need to remove the env var altogether after v2 migration.

chandrasekharan-zipstack (approved):

harini-venkataraman (approved):

muhammad-ali-e:
> @muhammad-ali-e LGTM overall.
> 
> We need to either rename the env var to `DB_SCHEMA` if we are going to continue having it from hereon or we need to remove the env var altogether after v2 migration.

updated

hari-kuriakose:
Duplication will be taken care of when we remove v1 later.

PR #543, opened 2024-07-31
fix: Error handling improvements: Unsaved classifier config and duplicate adapter name 

chandrasekharan-zipstack:
## What

- Handle error when classifier config is not saved during WF run
- Handle duplicate adapter name errors better
- Removed some unnecessary catch-all errors which would re-raise a different exception

## Why

- Unsaved config during a WF run raised an obscure error message

## How

- Handled few extra cases for JSON schema validation

- No, minor error handling concerns only

## Notes on Testing

- Tested locally by reproducing these cases

## Screenshots

image
image

## Checklist

### DIFF
--- a/backend/adapter_processor/exceptions.py
+++ b/backend/adapter_processor/exceptions.py
@@ -1,5 +1,6 @@
 from typing import Optional
 
+from adapter_processor.constants import AdapterKeys
 from rest_framework.exceptions import APIException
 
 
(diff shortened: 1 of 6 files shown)


backend/adapter_processor/exceptions.py
```diff
@@ -64,6 +75,8 @@ def __init__(
         adapter_name: str = "adapter",
     ):
         if detail is None:
+            if adapter_name != "adapter":
```
muhammad-ali-e, line 78: What this conditions makes here?

muhammad-ali-e: Can we do like this?
            detail = (
                f"Cannot delete '{adapter_name}'. "
                "It is used in a workflow or a prompt studio project"
            )

Deepak-Kesavan (approved):
Looks good overall

backend/adapter_processor/exceptions.py
chandrasekharan-zipstack: @muhammad-ali-e in the event that we call this exception class and we don't pass an `adapter_name` - the error message would read as `Cannot delete 'adapter'. It is used in a workflow or a prompt studio project.`
I didn't want to single quote the string `adapter` since its a generic placeholder

muhammad-ali-e (approved):
LGTM

PR #545, opened 2024-07-31
Etl Update API and Periodic Tasks changes -Optional Schedule

muhammad-ali-e:
## What

- Updated pipeline handling to optimize enable/disable operations.
- Refined the pipeline update functionality.

## Why

- The cron_string field should be optional, not mandatory, to provide more flexibility in pipeline scheduling.

## How

- Enhanced the update logic for the Pipeline model.
- Modified periodic task management to accommodate the optional cron_string.
  - The periodic task will be removed if cron_string is set to None or empty.
  - The periodic task will be added when a valid cron_string is provided.

- This PR should be released in tandem with corresponding frontend changes to ensure compatibility.

## Database Migrations

-  This PR includes schema migrations.

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


backend/pipeline/serializers/crud.py
```diff
@@ -1,13 +1,15 @@
 import logging
 from collections import OrderedDict
-from typing import Any
+from typing import Any, Optional
 
 from connector.connector_instance_helper import ConnectorInstanceHelper
 from connector.models import ConnectorInstance
 from connector_processor.connector_processor import ConnectorProcessor
+from croniter import croniter
```
chandrasekharan-zipstack, line 8: @muhammad-ali-e is this a new library? If yes, please add it to `pyproject.toml`

backend/pipeline/serializers/crud.py
```diff
@@ -24,6 +26,34 @@ class Meta:
...
+        cron_string = value.strip()
+        # Check if the string is empty
+        if not cron_string:
+            return None
+
+        # Validate the cron string
+        try:
+            croniter(cron_string)
+        except Exception as error:
+            logger.error(f"Invalid cron string: {error}")
```
chandrasekharan-zipstack, line 52:
```suggestion
            logger.error(f"Invalid cron string '{cron_string}': {error}")
```

backend/pipeline/serializers/crud.py
```diff
@@ -32,10 +62,10 @@ def create(self, validated_data: dict[str, Any]) -> Any:
 
     def save(self, **kwargs: Any) -> Pipeline:
         pipeline: Pipeline = super().save(**kwargs)
-        SchedulerHelper.add_job(
-            str(pipeline.pk),
-            cron_string=pipeline.cron_string,
-        )
+        if pipeline.cron_string is None:
+            SchedulerHelper.remove_job(pipeline_id=str(pipeline.id))
```
chandrasekharan-zipstack, line 66: @muhammad-ali-e since we remove the job when `cron_string` is made `None` - consider the case where we create a pipeline with a valid cron string, run it few times and then remove the cron string. In such a case will we lose information such as `run_count` or the last run time / status?

backend/scheduler/tasks.py
```diff
@@ -49,15 +50,26 @@ def create_periodic_task(
...
+        PeriodicTask.objects.create(
+            name=task_name,
+            task=task_path,
+            crontab=schedule,
+            enabled=enabled,
+            args=task_args_json,
+        )
+        logger.info(
+            f"Created periodic task {task_name} with cron string: {cron_string}"
+        )
```
chandrasekharan-zipstack, line 72: @muhammad-ali-e did you consider using [the update_or_create](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#update-or-create) here?

backend/pipeline/serializers/crud.py
```diff
@@ -24,6 +26,34 @@ class Meta:
         model = Pipeline
         fields = "__all__"
 
+    def validate_cron_string(self, value: Optional[str] = None) -> Optional[str]:
```
chandrasekharan-zipstack, line 29: @muhammad-ali-e this function isn't used anywhere - can it be removed?

backend/pipeline/serializers/crud.py
muhammad-ali-e: Oh... It was there, For got to commit

backend/pipeline/serializers/crud.py
muhammad-ali-e: We will only remove the job from Celery's periodic tasks. The last run time and status will still be retained within the Pipeline model.

While we do lose the run_count from Celery, it’s not something we're currently utilizing in our pipeline. If needed, we could store this information in the Pipeline model, but since we can retrieve the count from the workflow_execution table, So this isn't necessary in Pipeline table too.

Additionally, it's important to note that Celery's periodic task table requires at least one scheduling time to function, so if the cron_string is removed, the periodic task must be deleted.

backend/scheduler/tasks.py
muhammad-ali-e: Let't me check. I used some methods, But it not works. Thanks for the link.

backend/pipeline/serializers/crud.py
muhammad-ali-e: It's Serializer validation, It will internally called by serializer.

backend/scheduler/tasks.py
muhammad-ali-e: Updated

backend/pipeline/serializers/crud.py
muhammad-ali-e: Added

chandrasekharan-zipstack (approved):

backend/pipeline/serializers/crud.py
chandrasekharan-zipstack: If possible add a docstring mentioning this - we wouldn't want to remove such functions stating that there's no usages

athul-rs (approved):
LGTM

PR #547, opened 2024-08-02
Feat/usage reporting page

johnyrahul:
## What

- Changes in platform service to introduce API which will be called from SDK for page reporting
- Changes in the unstract backend to introduce a new model to which we will be persisting page usage

## Why

-

## How

-

- No,  Existing code part is not touched
1. Created new app in djnago 
2. Created new API endpoint in platform service

## Database Migrations

-  Yes, Introduced a new model for pag usage calculation

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

- https://github.com/Zipstack/unstract-sdk/pull/80

## Dependencies Versions

-

## Notes on Testing

-

## Screenshots

image

## Checklist


backend/account_usage/models.py
```diff
@@ -0,0 +1,47 @@
...
+        editable=False,
+        db_comment="Primary key for the usage entry, automatically generated UUID",
+    )
+
+    organization_id = models.CharField(
+        default="mock_org",
+        null=False,
+        blank=False,
+    )
+    file_name = models.CharField(
```
chandrasekharan-zipstack, line 19: NIT: Might be a good idea to store the entire file path in case its read from a file system

johnyrahul: @chandrasekharan-zipstack 
even the file name does not make much sense if we are running the tools in the container( During workflow execution the fle_name will always be `INFILE`)
In a similar view, the file path also won't add much sense. 

Only place it make send is when we do file processing from prompt studio side.

chandrasekharan-zipstack: @johnyrahul got it - I lacked context of what this field was used for. Somehow it might be good to indicate (preferably through the name or a docstring / comment) that we store the mangled / copied file path. Reason being there might be cases where we'd like to display the user provided file path somewhere and such fields won't be of help to us
If it doesn't hurt us - we could store both user provided file path and the file path that we store during execution

chandrasekharan-zipstack (approved):

muhammad-ali-e (approved):
LGTM,
Please make V2 changes once it tested by QA

platform-service/src/unstract/platform_service/main.py
```diff
@@ -182,6 +184,61 @@ def health_check() -> str:
...
+        page_count,
+        file_name,
+        file_size,
+        file_type,
+        run_id,
+        current_time,
+    )
+
+    try:
+        with be_db.atomic() as transaction:
```
gaya3-zipstack, line 229: @johnyrahul Trying to understand this... Why do we need an atomic block here? I see only one Insert query being executed here... Is that right?

johnyrahul: @gaya3-zipstack Yes, you are correct, I followed the same as what we are doing for the token usage endpoint. I don't know if that is added for any specific reason. @Deepak-Kesavan  any idea?

johnyrahul: Anyways it won't do any damage, I believe.

PR #554, opened 2024-08-05
fix: Allow mounting tool registry in OSS to load private tools tool

chandrasekharan-zipstack:
## What

- Allow mounting tool registry through a config directory outside the library code
- Added some logs during loading of tools
- Refactored a snippet in `./run-platform.sh` into a function to allow copying envs

## Why

- Privately registered tools did not appear in the platform and its necessary to keep loading of tools dynamic / controlled with a config / env

## How

- Added an env to obtain the config path - all tool registry related files are expected to be here
- This directory gets mounted into the backend container

- Yes, this has been tested in local by building the images - however since its a change in the loading of tool registry careful testing is required

## Env Config

- Added to `backend's .env`
```
# Path where public and private tools are registered
# with a YAML and JSONs
TOOL_REGISTRY_CONFIG_PATH="/data/tool_registry_config"
```
- Added to `docker's .env`
```
# Path where public and private tools are registered
# with a YAML and JSONs
TOOL_REGISTRY_CONFIG_SRC_PATH="${PWD}/../unstract/tool-registry/tool_registry_config"
```

## Notes on Testing

- Tried running and mounting it locally - the tools are truly loaded dynamically and work with a file change

## Screenshots
image
image
- Logging tools loaded with their version - this might turn out to be noisy, we should try to make the tool registry a singleton and perform loading once rather than making it a debug log
image

## Checklist


gaya3-zipstack (approved):
Looks good.

unstract/tool-registry/tool_registry_config/sample_registry.yaml
```diff
@@ -9,7 +9,6 @@ auth:
   username: <username>
 document: unstract-tool-registry
 tools:
-  # - docker:unstract/tool-resume-parser
-  - local:fileops
```
ritwik-g, line 13: @chandrasekharan-zipstack are we not using this now?

chandrasekharan-zipstack: This is a sample file meant for reference and we don't have a fileops tool now. Didn't want to mention that old tool here - we do support downloading the image from docker and obtaining it locally as well

muhammad-ali-e (approved):
LGTM, believe it was tested locally.
- @chandrasekharan-zipstack are there any related changes on the platform?

PR #570, opened 2024-08-09
handled never ending etl issue in MRQ when highlight is not selected

vishnuszipstack:
## What

- When tool exported with highlight not enabled ETL was going into never ending state

## Why

- Was trying to get whisper_hash value and that was not handled properly

## How

- Checking for whisper_hash value present and if not present saving that values as none.

- No, minor fix

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/backend/workflow_manager/endpoint/destination.py
+++ b/backend/workflow_manager/endpoint/destination.py
@@ -535,6 +535,8 @@ class DestinationConnector(BaseConnector):
             q_name = f"review_queue_{self.organization_id}_{workflow.id}"
             if meta_data:
                 whisper_hash = meta_data.get("whisper-hash")
+            else:
+                whisper_hash = None
             queue_result = QueueResult(
                 file=file_name,
                 status=QueueResultStatus.SUCCESS,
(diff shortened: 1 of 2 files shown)


backend/workflow_manager/endpoint/destination.py
```diff
@@ -528,17 +528,20 @@ def _push_to_queue(
             settings=connector_settings, connector_id=connector.connector_id
         )
         with source_fs.open(input_file_path, "rb") as remote_file:
+            whisper_hash = None
             file_content = remote_file.read()
             # Convert file content to a base64 encoded string
             file_content_base64 = base64.b64encode(file_content).decode("utf-8")
             q_name = f"review_queue_{self.organization_id}_{workflow.id}"
+            if meta_data:
+                whisper_hash = meta_data.get("whisper-hash")
```
ritwik-g, line 537: In this particular case I think we can initialise whisper_hash using else clause rather than doing it separately. 
```suggestion
            if meta_data:
                whisper_hash = meta_data.get("whisper-hash")
            else:
                whisper_hash = None
```

muhammad-ali-e (approved):
LGTM

ritwik-g (approved):
I have one NIT suggestion. Other than that everything looks good.

PR #575, opened 2024-08-12
fix/error-handling-for-gdrive-connectors

kirtimanmishrazipstack:
## What

- Error handling for gdrive connector

## Why

- Happens when wrong env is passed to gdrive, an issue we faced

## How

- Through `FSAccessDeniedError` when google-exception returns acess denied on gdrive
- Handle the same error in backed and raise django error to UI

- N/A

## Database Migrations

- N/A

## Env Config

- N/A

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Screenshots
Screenshot from 2024-08-12 22-18-41

## Checklist


backend/connector_processor/connector_processor.py
```diff
@@ -132,6 +134,8 @@ def test_connectors(connector_id: str, credentials: dict[str, Any]) -> bool:
             test_result = connector_impl.test_credentials()
             logger.info(f"{connector_id} test result: {test_result}")
             return test_result
+        except UnstractFSConnectorException as e:
+            raise UnstractFSException(detail=e.detail) from e
```
ritwik-g, line 138: @kirtimanmishrazipstack UnstractFSCOnnectorException is already a custom exception raised by us. Why are we again raising it as `UnstractFSException`?

backend/connector_processor/exceptions.py
```diff
@@ -49,3 +49,11 @@ def __init__(self, core_err: ConnectorError) -> None:
         super().__init__(detail=core_err.message, core_err=core_err)
         self.default_detail = core_err.message
         self.status_code = 400
+
+
+class UnstractFSException(APIException):
+    default_detail = "Error testing connection. "
+
+    def __init__(self, detail: str = default_detail) -> None:
+        status_code = 403
```
chandrasekharan-zipstack, line 58: The exception name specifies something generic for all FS connector exception but defaulting the status code to 403 doesn't make sense. I suggest setting the status code explicitly wherever necessary or renaming this to encompass only permission related errors - but it might be hard to determine permission related errors for different connectors

unstract/connectors/src/unstract/connectors/filesystems/google_drive/google_drive.py
```diff
@@ -25,13 +27,16 @@ class GoogleDriveFS(UnstractFileSystem):
...
         super().__init__("GoogleDrive")
-        client_secrets = json.loads(
-            GCSHelper().get_secret("google_drive_client_secret")
-        )
+        try:
+            self.client_secrets = json.loads(
+                GCSHelper().get_secret("google_drive_client_secret")
+            )
+        except GoogleApiException.PermissionDenied as e:
+            raise GoogleDriveAccessDeniedException(detail=e.details) from e
```
chandrasekharan-zipstack, line 35: @kirtimanmishrazipstack we'd throw this message for all such `PermissionDenied` errors. But the error message is too specific. Either catch and raise only for the specific scenario or have a generic message that encompasses all this.
Also a cloud user might not find such a message useful - he wouldn't know what env we're talking about here

kirtimanmishrazipstack: @chandrasekharan-zipstack as discussed, handled over generic message

unstract/connectors/src/unstract/connectors/filesystems/exceptions.py
```diff
@@ -0,0 +1,18 @@
+from typing import Any
+
+
+class UnstractFSConnectorException(Exception):
+    """Base class for database-related exceptions from Unstract connectors."""
+
+    def __init__(self, *args: object, detail: Any, status_code: Any = 500) -> None:
+        super().__init__(*args)
+        self.detail = detail
+        self.status_code = status_code
```
chandrasekharan-zipstack, line 10: @kirtimanmishrazipstack please check if this works, try to propagate the same change to other similar overrides done to the constructor
```suggestion
    status_code = 500
```

unstract/connectors/src/unstract/connectors/filesystems/google_drive/google_drive.py
```diff
@@ -25,13 +27,16 @@ class GoogleDriveFS(UnstractFileSystem):
     # }
     def __init__(self, settings: dict[str, Any]):
         super().__init__("GoogleDrive")
-        client_secrets = json.loads(
-            GCSHelper().get_secret("google_drive_client_secret")
-        )
+        try:
+            self.client_secrets = json.loads(
```
chandrasekharan-zipstack, line 31: NIT: @kirtimanmishrazipstack why does this need to be a part of `self`?

unstract/connectors/src/unstract/connectors/filesystems/exceptions.py
kirtimanmishrazipstack: @chandrasekharan-zipstack this won't work. Also why do we need to change that? status_code is default 500

unstract/connectors/src/unstract/connectors/filesystems/google_drive/google_drive.py
kirtimanmishrazipstack: Because of try-catch client_secrets is not accessible in `oauth2_credentials`

backend/connector_processor/connector_processor.py
kirtimanmishrazipstack: @ritwik-g since unstract/connector is a library it is producing `UnstractFSConnectorException` . All fs connector exception will inherit from this `UnstractFSConnectorException`
On backend  `UnstractFSException` will inherit django `APIException` and displays the error to FE

chandrasekharan-zipstack: NIT: @kirtimanmishrazipstack for brevity I suggest removing the `Unstract` prefix on `UnstractFSConnectorException` and renaming it as `FSConnectorError`

backend/connector_processor/exceptions.py
```diff
@@ -49,3 +51,10 @@ def __init__(self, core_err: ConnectorError) -> None:
         super().__init__(detail=core_err.message, core_err=core_err)
         self.default_detail = core_err.message
         self.status_code = 400
+
+
+class UnstractFSException(APIException):
+    default_detail = "Error testing connection. "
+
+    def __init__(self, detail: str = default_detail, status_code: Any = 500) -> None:
+        super().__init__(detail=detail, code=status_code)
```
chandrasekharan-zipstack, line 60:
```suggestion
```
@kirtimanmishrazipstack this override isn't necessary. Please remove all similar overrides done for the APIException inheritance if we aren't doing anything different

backend/connector_processor/connector_processor.py
```diff
@@ -132,6 +134,8 @@ def test_connectors(connector_id: str, credentials: dict[str, Any]) -> bool:
             test_result = connector_impl.test_credentials()
             logger.info(f"{connector_id} test result: {test_result}")
             return test_result
+        except UnstractFSConnectorException as e:
```
chandrasekharan-zipstack, line 137: @kirtimanmishrazipstack do we really need a new base class when we already have `ConnectorError`? Can you try re-using the same?

unstract/connectors/src/unstract/connectors/filesystems/exceptions.py
chandrasekharan-zipstack: My bad, just noticed that its inheriting `Exception` class alone

backend/connector_processor/connector_processor.py
kirtimanmishrazipstack: @chandrasekharan-zipstack using connectorError, we can't specify from which db or fs we are getting the execption

Moreover the way we have handled connectorError is not proper. Entire stack trace is visible to the user which is not a good UI

kirtimanmishrazipstack: moreover all the fs connector exception can be clubbed as UnstractFSConnectorException

backend/connector_processor/exceptions.py
kirtimanmishrazipstack: @chandrasekharan-zipstack if we don't override, then entire stack trace gets popped up in the UI.
I think this approch is better

chandrasekharan-zipstack: The entire stack trace to the user? Can you provide a screenshot?

backend/connector_processor/connector_processor.py
chandrasekharan-zipstack: I'd still suggest inheriting from `ConnectorError`. If its implementation is not proper, let's try to correct it

chandrasekharan-zipstack (approved):

backend/connector_processor/connector_processor.py
kirtimanmishrazipstack: done

unstract/connectors/src/unstract/connectors/exceptions.py
```diff
@@ -31,3 +31,22 @@ def __init__(
...
+    def __init__(
+        self,
+        message: str,
+        *args: Any,
+        treat_as_user_message: bool = False,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(
+            message, *args, treat_as_user_message=treat_as_user_message, **kwargs
+        )
```
chandrasekharan-zipstack, line 52:
```suggestion
```

Deepak-Kesavan (approved):

PR #577, opened 2024-08-13
fix/github-workflow-changes

kirtimanmishrazipstack:
## What

- Fixing pdm lock automation uncommited changes
- Adding paths on automation

## Why

- When automation runs, if there is any uncommited changes, it fails
- Workflow should run when `pyproject.toml` changes

## How

On automation did the following:

- Generate lock file 
- Git stash (It will stash uncommited changes)
- Git pull remote branch
- Git stash pop
- Git add and commit pdm.lock

- N/A

## Database Migrations

- N/A

## Env Config

- N/A

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

- Chages pyproject.toml in backend, prompt, platform
- Generated pdm.lock for backend, but didn't pushed. 
- Generated and pushed pdm.lock for platform
- Workflow run successfully for all 3 services. 

- Since backend had unstaged changes, it didn't affect.
- pdm lock was already pushed for platform, but automation generated new pdm lock changes
- pdm lock generated for prompt. 

## Checklist


.github/workflows/pdm-lock-automation.yaml
```diff
@@ -75,17 +91,39 @@ jobs:
           else
             echo "Virtual environment already exists."
           fi
+
           source .venv/bin/activate
+
+          echo "Stashing changes if there are any..."
+          git stash || echo "No stash entries found, continuing..."
```
ritwik-g, line 98: @kirtimanmishrazipstack git stash should be done after pdm lock run anf before git pull.

kirtimanmishrazipstack: @ritwik-g done

.github/workflows/pdm-lock-automation.yaml
```diff
@@ -76,17 +92,41 @@ jobs:
...
+            echo "Adding pdm.lock changes..."
             git add pdm.lock
+
+            echo "Installing pre-commit..."
             pip install pre-commit~=3.6.2
+
+            echo "Running pre-commit checks..."
             pre-commit run pdm-lock-check
+
+            echo "Committing and pushing changes..."
```
ritwik-g, line 129: @kirtimanmishrazipstack one of the potential issue here is that we might again run in to conflicts by the time we complete pre-commit. I feel like we could skip pre-commit checks in this case. Anyway we have a pre-commit check bot for this. Can you check this with @chandrasekharan-zipstack and @hari-kuriakose

Deepak-Kesavan (commented):
@kirtimanmishrazipstack Why there are 54 commits in this branch? Please make sure to branch out from latest branch from next time.

ritwik-g (approved):

kirtimanmishrazipstack:
> @kirtimanmishrazipstack Why there are 54 commits in this branch? Please make sure to branch out from latest branch from next time.

@Deepak-Kesavan that's because, to test I manually changed pyproject.toml files. Then workflow will produce more commits on each job if the pdm lock has to be added. I also need to reverse the changes after testing it. 

Since this cycle of repeated tests happened, more commits come. But this is not a concern

Deepak-Kesavan (approved):

.github/workflows/pdm-lock-automation.yaml
kirtimanmishrazipstack: @ritwik-g I think pre-commit check will just check the generated pdm lock. We can keep the change.

chandrasekharan-zipstack: Though we are only performing `lock-check` here the bot very well does this so we can skip the pre-commit check here.

@ritwik-g does the pre-commit check on the PR happen in parallel when this script runs?

hari-kuriakose: @ritwik-g  Good call.
However, might be good to continue having a check to validate the generated pdm lock file.

@kirtimanmishrazipstack This pre-commit hook boils down to just `pdm lock --check`.
See https://github.com/pdm-project/pdm/blob/main/.pre-commit-hooks.yaml.

Hence let's not invoke the pre-commit machinery here, instead just run `pdm lock --check` directly.
Should be ok.

.github/workflows/pdm-lock-automation.yaml
```diff
@@ -76,17 +92,41 @@ jobs:
...
             echo "Virtual environment already exists."
           fi
+
           source .venv/bin/activate
+
+          echo "Generating/Updating pdm.lock..."
           pdm lock -G :all -v
+
+          echo "Stashing changes if there are any..."
+          git stash || echo "No stash entries found, continuing..."
```
hari-kuriakose, line 102: @kirtimanmishrazipstack Also just curious on what kind of uncommitted changes would be present if at all, as this is running inside the workflow env.

kirtimanmishrazipstack: @hari-kuriakose if the developer running the automation have some uncommitted changes in any of files, including manually generating pdm lock. This will handle all the scenarios

ritwik-g: @kirtimanmishrazipstack there won't be any uncommitted files other than pdm lock files.

kirtimanmishrazipstack: @ritwik-g I meant that, what if the developer has pushed only pyproject file, but he/she is working on some other files. 
When the workflow runs, all the files the user has changed but not committed will come as uncommitted right?

.github/workflows/pdm-lock-automation.yaml
```diff
@@ -76,17 +92,39 @@ jobs:
...
-            git pull origin ${{ github.head_ref }} --allow-unrelated-histories --no-rebase
             echo "Changes detected between the current branch and the ${{ github.head_ref }} branch."
+
+            echo "Adding pdm.lock changes..."
             git add pdm.lock
-            pip install pre-commit~=3.6.2
-            pre-commit run pdm-lock-check
+
+            echo "Running pre-commit pdm hook..."
+            pdm lock --check
```
hari-kuriakose, line 125: @kirtimanmishrazipstack Can you please move this before `git add`?

kirtimanmishrazipstack: done @hari-kuriakose

.github/workflows/pdm-lock-automation.yaml
kirtimanmishrazipstack: made the changes @hari-kuriakose

hari-kuriakose (approved):
@kirtimanmishrazipstack LGTM

PR #582, opened 2024-08-14
Feat/unified dev portal

jagadeeswaran-zipstack:
## What

- Unified developer portal for LLMWhisperer

## Why

- A single portal to access both Unstract and LLMWhisperer

## How

- Added “Application Chooser” after logo (for all products) using that user can navigate through apps .

- Yes, the existing modules have been modified to support Unified portal, so there is a chance that this pull request could break existing features.

## Screenshots

...

## Checklist


vishnuszipstack (approved):

frontend/src/components/navigations/top-nav-bar/TopNavBar.jsx
```diff
@@ -219,17 +250,22 @@ function TopNavBar() {
 
   return (
     <Row align="middle" className="topNav">
-      <Col span={6}>
-        <UnstractLogo className="topbar-logo" />
+      <Col span={6} className="platform-switch-container">
+        {isUnstract ? (
+          <UnstractLogo className="topbar-logo" />
+        ) : (
+          <WhispererLogo className="topbar-logo" />
```
tahierhussain, line 257: @jagadeeswaran-zipstack Let's make this a cloud component. Here are the necessary steps to do so:

1. Move the LLM Whisperer-related icons to the cloud repo.
2. Export the icons from the `index.js` file in the cloud repo.
3. Import the icon in this component using our regular try-catch mechanism.

frontend/src/setupProxy.js
```diff
@@ -8,4 +8,11 @@ module.exports = (app) => {
       changeOrigin: true,
     })
   );
+  app.use(
+    "/llmwhisperer",
+    createProxyMiddleware({
+      target: process.env.REACT_APP_WHISPERER_PORTAL_URL,
+      changeOrigin: true,
+    })
+  );
```
tahierhussain, line 17: @jagadeeswaran-zipstack Can you expore and see if we can store this in the cloud repo and create a new instance if it's available?

tahierhussain (changes requested):

frontend/src/setupProxy.js
jagadeeswaran-zipstack: We can remove this from OSS and add a READ.me where we can inform devs to add this while they testing in local. because this is only required when we run the application in local.

frontend/src/components/navigations/top-nav-bar/TopNavBar.jsx
jagadeeswaran-zipstack: Got it :+1:

frontend/public/credit_card.pdf
jaseemjaskp, line 1: @jagadeeswaran-zipstack can you move this pdf files into cloud under folder public/llmwhisperer?.

jagadeeswaran-zipstack: Yes.

PR #592, opened 2024-08-18
[FIX] Fixes for v0.81.0

Deepak-Kesavan:
## What

- Fixed regression in API deployment
- Fixed issue where metadata is not included for API deployment async
- Fixed status code for API deployment async response when it is successful.
- Updated code snippet and postman collection with current changes for API deployment

## Why

-

## How

-

-

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/backend/api/api_deployment_views.py
+++ b/backend/api/api_deployment_views.py
@@ -74,28 +74,29 @@ class DeploymentExecution(views.APIView):
         self, request: Request, org_name: str, api_name: str, api: APIDeployment
     ) -> Response:
         execution_id = request.query_params.get("execution_id")
-        include_metadata = request.query_params.get("include_metadata", False)
+        include_metadata = (
+            request.query_params.get(ApiExecution.INCLUDE_METADATA, "false").lower()
+            == "true"
+        )
         if not execution_id:
             raise InvalidAPIRequest("execution_id shouldn't be empty")
         response: ExecutionResponse = DeploymentHelper.get_execution_status(
             execution_id=execution_id
         )
-        if response.execution_status != CeleryTaskState.SUCCESS.value:
-            return Response(
-                {
-                    "status": response.execution_status,
-                    "message": response.result,
-                },
-                status=status.HTTP_422_UNPROCESSABLE_ENTITY,
-            )
-        if (
-            response.execution_status == CeleryTaskState.SUCCESS.value
-            and not include_metadata
-        ):
-            response.remove_result_metadata_keys()
+        response_status = status.HTTP_200_OK
+        if response.execution_status == CeleryTaskState.COMPLETED.value:
+            if include_metadata:
+                response.remove_result_metadata_keys(keys_to_remove=["highlight_data"])
+            else:
+                response.remove_result_metadata_keys()
+        else:
+            response_status = status.HTTP_422_UNPROCESSABLE_ENTITY
         return Response(
-            {"status": response.execution_status, "message": response.result},
-            status=status.HTTP_200_OK,
+            data={
+                "status": response.execution_status,
+                "message": response.result,
+            },
+            status=response_status,
         )
 
 
--- a/backend/api/api_deployment_views.py
+++ b/backend/api/api_deployment_views.py
@@ -83,14 +83,13 @@ class DeploymentExecution(views.APIView):
         response: ExecutionResponse = DeploymentHelper.get_execution_status(
             execution_id=execution_id
         )
-        response_status = status.HTTP_200_OK
+        response_status = status.HTTP_422_UNPROCESSABLE_ENTITY
         if response.execution_status == CeleryTaskState.COMPLETED.value:
+            response_status = status.HTTP_200_OK
             if include_metadata:
                 response.remove_result_metadata_keys(keys_to_remove=["highlight_data"])
             else:
                 response.remove_result_metadata_keys()
-        else:
-            response_status = status.HTTP_422_UNPROCESSABLE_ENTITY
         return Response(
             data={
                 "status": response.execution_status,
--- a/backend/prompt_studio/prompt_studio_core/prompt_studio_helper.py
+++ b/backend/prompt_studio/prompt_studio_core/prompt_studio_helper.py
@@ -29,7 +29,6 @@ from prompt_studio.prompt_studio_core.exceptions import (
     IndexingAPIError,
     NoPromptsFound,
     PermissionError,
-    ToolNotValid,
 )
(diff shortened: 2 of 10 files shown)


backend/api/api_deployment_views.py
```diff
@@ -74,28 +74,29 @@ def get(
...
-        ):
-            response.remove_result_metadata_keys()
+        response_status = status.HTTP_200_OK
+        if response.execution_status == CeleryTaskState.COMPLETED.value:
+            if include_metadata:
+                response.remove_result_metadata_keys(keys_to_remove=["highlight_data"])
+            else:
+                response.remove_result_metadata_keys()
+        else:
+            response_status = status.HTTP_422_UNPROCESSABLE_ENTITY
```
Deepak-Kesavan, line 93: Retained previous functionality of 422 status is not successful. Need to send proper status codes here.

backend/api/api_deployment_views.py
```diff
@@ -74,28 +74,29 @@ def get(
...
-                },
-                status=status.HTTP_422_UNPROCESSABLE_ENTITY,
-            )
-        if (
-            response.execution_status == CeleryTaskState.SUCCESS.value
-            and not include_metadata
-        ):
-            response.remove_result_metadata_keys()
+        response_status = status.HTTP_200_OK
+        if response.execution_status == CeleryTaskState.COMPLETED.value:
```
Deepak-Kesavan, line 87: While testing all the responses were 422 and noticed that the status was `COMPLETED` and not `SUCCESS`. @muhammad-ali-e @athul-rs Please take a look here if this is correct.

backend/api/api_deployment_views.py
```diff
@@ -74,28 +74,29 @@ def get(
         self, request: Request, org_name: str, api_name: str, api: APIDeployment
     ) -> Response:
         execution_id = request.query_params.get("execution_id")
-        include_metadata = request.query_params.get("include_metadata", False)
+        include_metadata = (
```
chandrasekharan-zipstack, line 77: NIT: Try to use [this util function](https://github.com/Zipstack/unstract/blob/c694de72e7a2cb051ed008956287df8a77587233/backend/utils/common_utils.py#L6-L15)

backend/api/api_deployment_views.py
```diff
@@ -74,28 +74,29 @@ def get(
...
-            response.execution_status == CeleryTaskState.SUCCESS.value
-            and not include_metadata
-        ):
-            response.remove_result_metadata_keys()
+        response_status = status.HTTP_200_OK
+        if response.execution_status == CeleryTaskState.COMPLETED.value:
+            if include_metadata:
+                response.remove_result_metadata_keys(keys_to_remove=["highlight_data"])
+            else:
+                response.remove_result_metadata_keys()
```
chandrasekharan-zipstack, line 91: NIT: I could step into the function to understand better - but this gives the impression that in spite of `include_metadata` - we always remove metadata key. See if anything can be renamed / comments added to better improve the understanding

backend/prompt_studio/prompt_studio_core/prompt_studio_helper.py
```diff
@@ -324,10 +323,6 @@ def index_document(
             )
             file_path = str(Path(file_path) / file_name)
 
-        if not tool:
-            logger.error(f"No tool instance found for the ID {tool_id}")
-            raise ToolNotValid()
```
chandrasekharan-zipstack, line 329: NIT: If this exception is not used anywhere, please remove that as well

chandrasekharan-zipstack (approved):
LGTM - do check with @muhammad-ali-e or @athul-rs as well

backend/api/api_deployment_views.py
Deepak-Kesavan: Noted. Will use this util and raise a separate PR when I get time.

backend/api/api_deployment_views.py
Deepak-Kesavan: @chandrasekharan-zipstack It is clearly mentioned in the doc string of this function. So, I think that it will clear up any confusion an I couldn't come up with any better name 😅.

https://github.com/Zipstack/unstract/blob/830e56382b5725590ccc1ea3f21fa644f62e070e/backend/workflow_manager/workflow/dto.py#L66-L68

backend/api/api_deployment_views.py
athul-rs: @Deepak-Kesavan 
This this mentioned aren't we only looking for `success` status or `422` status code ? if there is some other response comes, will we be sending 200 since it is mentioned as default for response_status and only changed in case of 422?

- Didn't get the full context, is 422 coming in for all responses ?

backend/api/api_deployment_views.py
athul-rs: @muhammad-ali-e Please take a look at the above mentioned status codes, needs more refined messages.

frontend/src/components/deployments/display-code/DisplayCode.jsx
```diff
@@ -100,7 +100,7 @@ const DisplayCode = ({ isDialogOpen, setDialogOpen, url }) => {
     fetch("{{url}}", requestOptions)
     {{else}}
     var requestOptions = { method: 'GET', redirect: 'follow', headers: myHeaders};
-    fetch("{{url}}?execution_id=REPLACE_WITH_EXECUTION_ID", requestOptions)
+    fetch("{{url}}?execution_id=REPLACE_WITH_EXECUTION_ID&include_metadata=False", requestOptions)
```
athul-rs, line 103: NIT:
@Deepak-Kesavan If we are using the same query string / params for all the calls here, we could might as well make this a constant.

athul-rs (changes requested):
Code LGTM oveerall, left couple of comments, please address @Deepak-Kesavan

backend/api/api_deployment_views.py
Deepak-Kesavan: Yes @athul-rs . In the earlier code, everything was 422. `EXECUTING`, `COMPLETED` etc, because we were checking for `SUCCESS` and the status was `COMPLETED`. So, I changed the status check alone.

I am changing the status to 422 in the else part. To avoid any confusion, I will make 422 as default and assign 200 inside if condition.

athul-rs: Think that will make more sense. :+1:

athul-rs (approved):
LGTM

PR #603, opened 2024-08-21
feat: Changes to accept multiple folders for file source

chandrasekharan-zipstack:
## What

- Accepts multiple folders in input for a file source
- Data migration for converting string input to a list

## Why

- Previously only one folder could be provided at a time

- Yes, since it involves altering an existing field - impact for regressions must be checked

## Database Migrations

- Yes, converts the string column into a list

## Notes on Testing

- Tested migration and the functionality of accessing files from different folders in a single workflow run

## Screenshots

1. Pre-migration
image
2. During migration
image
3. Post-migration
image

4. Reverse migration works too 
image

## Checklist


backend/workflow_manager/endpoint/static/src/file.json
```diff
@@ -4,17 +4,21 @@
...
     "properties": {
-        "rootFolder": {
-            "type": "string",
-            "title": "Folder to process",
-            "default": "",
-            "description": "The root folder to start processing files from. Leave it empty to use the root folder"
+        "folders": {
+            "type": "array",
+            "title": "Folders to process",
+            "description": "Folders in the connected filesystem to start processing files from. Leave it empty or pass '/' to specify the root folder.",
```
muhammad-ali-e, line 10: I think , it would be better if we describe how to list the folders, By comma or space etc..

backend/workflow_manager/workflow/execution.py
```diff
@@ -105,13 +103,6 @@ def __init__(
 
         self.compilation_result = self.compile_workflow(execution_id=self.execution_id)
 
-    def _initiate_api_execution(
```
muhammad-ali-e, line 108: Why it removed?

backend/workflow_manager/endpoint/source.py
```diff
@@ -154,36 +156,46 @@ def list_files_from_file_connector(self) -> tuple[dict[str, FileHash], int]:
...
+        logger.info(f"Folders to process: {folders_to_process}")
+        total_files_to_process = 0
+        total_matched_files = {}
+        for input_directory in folders_to_process:
+            input_directory = source_fs.get_connector_root_dir(
+                input_dir=input_directory, root_path=root_dir_path
+            )
+            logger.debug(f"Listing files from:  {input_directory}")
+            is_directory = source_fs_fsspec.isdir(input_directory)
+            if not is_directory:
```
muhammad-ali-e, line 178: If the user provides multiple directories (e.g., a, b, c) and one of them (c) is not a valid directory, the current code will still process the valid directories (a and b) before raising an error for the invalid directory (c). This results in unnecessary processing for valid directories.

**Suggestions:** 
1. To avoid this, consider validating all provided directories upfront before processing any files. This ensures that we only proceed with file matching if all directories are valid, preventing wasted computation (especially for `_get_matched_files`).

2. Or Log the invalid directory and skip it, continuing with the remaining valid directories.

backend/workflow_manager/endpoint/migrations/0004_allow_multi_source_inputs_for_fs.py
```diff
@@ -0,0 +1,99 @@
...
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("endpoint", "0003_alter_workflowendpoint_connection_type"),
+    ]
+
+    def migrate_to_support_multiple_folder_input(apps: Any, schema_editor: Any) -> None:
```
muhammad-ali-e, line 14: Is this data migration necessary? Could we handle the input directly in the code instead? For example, we could:

- Split the input into a list if it contains commas, or
- If it's a single string (representing a single directory, as in the existing behavior), wrap it in a list.

backend/workflow_manager/endpoint/static/src/file.json
chandrasekharan-zipstack: @muhammad-ali-e this is similar to the array input of classification bins - will anyway update the description

backend/workflow_manager/workflow/execution.py
chandrasekharan-zipstack: This wasn't used anywhere @muhammad-ali-e - will try to capture dead code removals in a separate PR from now on

backend/workflow_manager/endpoint/source.py
chandrasekharan-zipstack: I'm leaning towards 1 so that its visible to the user. Will perform this check outside the loop

johnyrahul: I am also lean towards suggestion 1.

backend/workflow_manager/endpoint/migrations/0004_allow_multi_source_inputs_for_fs.py
johnyrahul: I believe what he suggested makes sense. We can avoid this migration.

johnyrahul:
@chandrasekharan-zipstack If we come up with code changes that don't involve migration as @muhammad-ali-e  suggested, that would be the safest and least breaking change.

backend/workflow_manager/endpoint/migrations/0004_allow_multi_source_inputs_for_fs.py
chandrasekharan-zipstack: > Split the input into a list if it contains commas, or

@muhammad-ali-e in order to accept a list of values, we've made a JSON schema change which returns an array. So it'll not be a single string with commas - we could keep the UI input as a string and enforce as a list, but it might not be the best UX.

> If it's a single string (representing a single directory, as in the existing behavior), wrap it in a list.

This might be possible but we'll eventually end up with a stale code branch that we might not know when to clean up. 
```
if isinstance(folders, str):
    # Convert to a list of 1 item
```

@johnyrahul @hari-kuriakose what are your thoughts on this?

chandrasekharan-zipstack: MoM of the huddle we had regarding this - agreed to go with migration itself.
- Will not be deleting the old `rootFolder` key though
- Will be filtering for all `SOURCE` endpoints (both `FILESYSTEM` and `API` types)

Will try to take a dump of staging data to test this

backend/workflow_manager/endpoint/source.py
chandrasekharan-zipstack: Implemented suggestion 1

backend/workflow_manager/endpoint/migrations/0004_allow_multi_source_inputs_for_fs.py
hari-kuriakose: Agree with a migration. This will avoid any unnecessary logic as the JSON schema is fixed to an array now.

hari-kuriakose (approved):
@chandrasekharan-zipstack Migration looks good.

One clarification though. In the screenshots provided, I see that `fileExtensions` array is empty after migration. Please check if this is a miss or not an issue.

johnyrahul (approved):

chandrasekharan-zipstack:
> @chandrasekharan-zipstack Migration looks good.
> 
> One clarification though. In the screenshots provided, I see that `fileExtensions` array is empty after migration. Please check if this is a miss or not an issue.

As discussed, the rows are not ordered in between the screenshots and ideally they should be compared based on their IDs

muhammad-ali-e (approved):
LGTM

backend/workflow_manager/endpoint/source.py
chandrasekharan-zipstack: Need to either handle the error in the FE or raise it as an exception during WF run. Can be handled as a separate task

Deepak-Kesavan (approved):

PR #616, opened 2024-08-22
[FIX] Workflow file path for table extractor

Deepak-Kesavan:
## What

Workflow file path for table extractor

## Why

-

## How

-

-

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


worker/src/unstract/worker/clients/test_docker.py
```diff
@@ -127,14 +127,16 @@ def test_get_container_run_config(docker_client, mocker):
...
+    execution_data_folder_prefix = os.getenv(Env.EXECUTION_RUN_DATA_FOLDER_PREFIX)
     mocker_normalize.assert_called_once_with("test-image")
     assert config["name"] == "test-image"
     assert config["image"] == "test-image:latest"
     assert config["command"] == ["echo", "hello"]
     assert config["environment"] == {
         "KEY": "VALUE",
-        "EXECUTION_RUN_DATA_FOLDER": "/source/org123/wf123/ex123",
+        "EXECUTION_RUN_DATA_FOLDER": (
+            f"{execution_data_folder_prefix}" "/org123/wf123/ex123"
```
ritwik-g: @Deepak-Kesavan we usually need to avoid calculated assertions when ever possible. That is the better way of handling this. So what I would suggest is for this case set an env `EXECUTION_RUN_DATA_FOLDER_PREFIX` at the beginning of this test. Then we can simply hardcode that value here.

Deepak-Kesavan: @ritwik-g Should I move it to the top or is it fine if it is kept at L130?

ritwik-g: You can keep under this line. This needs to happen before function calls. https://github.com/Zipstack/unstract/pull/616/files#diff-b89214285c8ac571e7607a2323f9d6c34d9d5bad91dd6dbb2a1b2a70496d69a4R111

worker/src/unstract/worker/clients/test_docker.py
```diff
@@ -109,6 +109,7 @@ def test_get_image(docker_client, mocker):
 def test_get_container_run_config(docker_client, mocker):
     """Test the get_container_run_config method."""
     os.environ[Env.WORKFLOW_DATA_DIR] = "/source"
+    execution_data_folder_prefix = os.getenv(Env.EXECUTION_RUN_DATA_FOLDER_PREFIX)
```
ritwik-g, line 112: @Deepak-Kesavan  not like this. What I meant is set this env like in line number 111
```suggestion
    os.environ[Env.EXECUTION_RUN_DATA_FOLDER_PREFIX] = "/app/workflow_data"
```

worker/src/unstract/worker/clients/test_docker.py
```diff
@@ -127,15 +128,15 @@ def test_get_container_run_config(docker_client, mocker):
...
+
     mocker_normalize.assert_called_once_with("test-image")
     assert config["name"] == "test-image"
     assert config["image"] == "test-image:latest"
     assert config["command"] == ["echo", "hello"]
     assert config["environment"] == {
         "KEY": "VALUE",
         "EXECUTION_RUN_DATA_FOLDER": (
-            f"{execution_data_folder_prefix}" "/org123/wf123/ex123"
+            f"{execution_data_folder_prefix}/org123/wf123/ex123"
```
ritwik-g, line 139:
```suggestion
            "/app/workflow_data/org123/wf123/ex123"
```

worker/src/unstract/worker/clients/docker.py
```diff
@@ -181,7 +182,12 @@ def get_container_run_config(
                     "target": os.getenv(Env.TOOL_DATA_DIR, "/data"),
                 }
             )
-            envs[Env.EXECUTION_RUN_DATA_FOLDER] = source_path
+            envs[Env.EXECUTION_RUN_DATA_FOLDER] = str(
+                Path(os.getenv(Env.EXECUTION_RUN_DATA_FOLDER_PREFIX, ""))
+                / organization_id
+                / workflow_id
+                / execution_id
+            )
```
ritwik-g, line 190: Use os.path.join instead of using str.
```suggestion
            envs[Env.EXECUTION_RUN_DATA_FOLDER] = os.path.join(
                os.getenv(Env.EXECUTION_RUN_DATA_FOLDER_PREFIX, ""),
                organization_id,
                workflow_id,
                execution_id
            )
```

ritwik-g (approved):

jagadeeswaran-zipstack (approved):

PR #617, opened 2024-08-23
Add timeout to `requests` calls

pixeeai:
## What, Why & How
Many developers will be surprised to learn that `requests` library calls do not include timeouts by default. This means that an attempted request could hang indefinitely if no connection is established or if no data is received from the server. 

The [requests documentation](https://requests.readthedocs.io/en/latest/user/advanced/#timeouts) suggests that most calls should explicitly include a `timeout` parameter. This codemod adds a default timeout value in order to set an upper bound on connection times and ensure that requests connect or fail in a timely manner. This value also ensures the connection will timeout if the server does not respond with data within a reasonable amount of time. 

While timeout values will be application dependent, we believe that this codemod adds a reasonable default that serves as an appropriate ceiling for most situations. 

Our changes look like the following:
```diff
 import requests
 
- requests.get("http://example.com")
+ requests.get("http://example.com", timeout=60)
```

<details>
  <summary>More reading</summary>

  * [https://docs.python-requests.org/en/master/user/quickstart/#timeouts](https://docs.python-requests.org/en/master/user/quickstart/#timeouts)
</details>

- If this logic is not in place, requests could last longer than they should. However, if you do have a request that you know should take more than 60 seconds, then there could be issues but I believe that case to be situational/rare.

## Database Migrations

-  N/A

## Env Config

-  N/A

## Relevant Docs

- Mentioned above, but will link here as well:   * [https://docs.python-requests.org/en/master/user/quickstart/#timeouts](https://docs.python-requests.org/en/master/user/quickstart/#timeouts)

## Related Issues or PRs

- N/A

## Dependencies Versions

- N/A

## Notes on Testing

- Requests will have a definite time before they are timed out.

## Checklist

I have read and understood the [Contribution Guidelines](X).

🧚🤖  Powered by [Pixeebot](https://github.com/apps/pixeebot)


hari-kuriakose:
@johnyrahul @muhammad-ali-e @gaya3-zipstack Looks good overall.
However does the proposed default timeout works out for all our use cases?

unstract/tool-sandbox/src/unstract/tool_sandbox/helper.py
```diff
@@ -52,7 +52,7 @@ def make_get_request(
         """
         url = f"{self.base_url}{endpoint}"
         params = {"image_name": image_name, "image_tag": image_tag}
-        response = requests.get(url, params=params)
+        response = requests.get(url, params=params, timeout=60)
```
muhammad-ali-e, line 55: Here It might get issue because, some tools need to be run more than 60 seconds. It's better to do with some env variables,

chandrasekharan-zipstack:
NIT: We also need to ensure we catch and handle [Timeout exception](https://requests.readthedocs.io/en/latest/api/#requests.Timeout). FYI we also have such calls made across other repos as well which might need a review too

pixeeai:
> NIT: We also need to ensure we catch and handle [Timeout exception](https://requests.readthedocs.io/en/latest/api/#requests.Timeout). FYI we also have such calls made across other repos as well which might need a review too

If there is another repo you have in mind that you'd like for me to scan with pixeebot, feel free to link it here.

Thanks,
Zach

PR #624, opened 2024-08-26
fix: Issue for gdrive with listing files

chandrasekharan-zipstack:
## What

- Added `root` prefix for GDrive while listing files during workflow execution

## Why

- This was missed and caused an issue with file listing

## How

- The correct prefix gets added through the concrete class implementation

- No

## Related Issues or PRs

- Regression from #603 

## Notes on Testing

- Tested locally, the error in case of incorrect directories does not get propagated correctly still. However this would allow happy path execution. Checked with @kirtimanmishrazipstack as well

## Checklist


kirtimanmishrazipstack (approved):
LGTM

PR #634, opened 2024-08-29
[FIX] Enhancing Subquestion Retrieval for Prompt Studio

harini-venkataraman:
## What
PR enhances the subquestion retrieval for Prompt studio
...

## Why
To  maximize retrieval accuracy using these subquestions
...

## How
Enhancing the retrieval prompts and adding to the context
...

No, this adds and enhances prompts for subquestion retrieval. 
...

## Database Migrations
Not applicable
...

## Env Config
Not applicable
...

## Relevant Docs
https://medium.com/@sauravjoshi23/complex-query-resolution-through-llamaindex-utilizing-recursive-retrieval-document-agents-and-sub-d4861ecd54e6

...

## Notes on Testing
This shows a comparison between tokens consumed, cost and response when Subquestion retrieval , simple retrieval and zero chunk  ( entire document in the context )  is used ( Profile managers in the same order ).

image

-

## Screenshots
Comparision between subquestion (left profile manager ) simple( right profile manager) prompts
image

image

## Checklist

### DIFF
--- a/prompt-service/src/unstract/prompt_service/main.py
+++ b/prompt-service/src/unstract/prompt_service/main.py
@@ -697,7 +697,7 @@ def run_retrieval(  # type:ignore
     prompt = output[PSKeys.PROMPTX]
     if retrieval_type == PSKeys.SUBQUESTION:
         subq_prompt: str = (
-            f"I am sending you a verbose prompt \n \n prompt : {prompt} \n \n"
+            f"I am sending you a verbose prompt \n \n Prompt : {prompt} \n \n"
             "Generate set of specific subquestions "
             "from the prompt which can be used to retrive "
             "relevant context from vector db. "
@@ -709,8 +709,8 @@ def run_retrieval(  # type:ignore
             "response might have a list of answers. The subquestions must not miss out "
             "any values in these cases. "
             "Output should be a list of comma seperated "
-            "subquestion prompts. Do not change the format "
-            " subquestions : "
+            "subquestion prompts. Do not change this format. \n \n "
+            " Subquestions : "
         )
         subquestions = run_completion(
             llm=llm,
--- a/prompt-service/src/unstract/prompt_service/main.py
+++ b/prompt-service/src/unstract/prompt_service/main.py
@@ -721,6 +721,11 @@ def run_retrieval(  # type:ignore
             retrieved_context = _retrieve_context(
                 output, doc_id, vector_index, each_subq
             )
+            # Not adding the potential for pinecode serverless
+            # inconsistency issue owing to risk of infinte loop
+            # and inablity to diffrentiate genuine cases of
+            # empty context.
+
             context = "".join([context, retrieved_context])
 
     if retrieval_type == PSKeys.SIMPLE:
--- a/prompt-service/src/unstract/prompt_service/main.py
+++ b/prompt-service/src/unstract/prompt_service/main.py
@@ -697,7 +697,7 @@ def run_retrieval(  # type:ignore
     prompt = output[PSKeys.PROMPTX]
     if retrieval_type == PSKeys.SUBQUESTION:
         subq_prompt: str = (
-            f"I am sending you a verbose prompt {prompt} . "
+            f"I am sending you a verbose prompt \n \n {prompt} \n \n"
             "Generate set of specific subquestions "
             "from the prompt which can be used to retrive "
             "relevant context from vector db. "
@@ -710,7 +710,7 @@ def run_retrieval(  # type:ignore
             "any values in these cases. "
             "Output should be a list of comma seperated "
             "subquestion prompts. Do not change the format "
-            " subquestions : \n \n"
+            " subquestions : "
         )
         subquestions = run_completion(
             llm=llm,
--- a/prompt-service/src/unstract/prompt_service/main.py
+++ b/prompt-service/src/unstract/prompt_service/main.py
@@ -697,7 +697,7 @@ def run_retrieval(  # type:ignore
     prompt = output[PSKeys.PROMPTX]
     if retrieval_type == PSKeys.SUBQUESTION:
         subq_prompt: str = (
-            f"I am sending you a verbose prompt \n \n {prompt} \n \n"
+            f"I am sending you a verbose prompt \n \n prompt : {prompt} \n \n"
             "Generate set of specific subquestions "
             "from the prompt which can be used to retrive "
             "relevant context from vector db. "
--- a/prompt-service/src/unstract/prompt_service/main.py
+++ b/prompt-service/src/unstract/prompt_service/main.py
@@ -693,28 +693,50 @@ def run_retrieval(  # type:ignore
     retrieval_type: str,
     metadata: dict[str, Any],
 ) -> tuple[str, str]:
+    context: str = ""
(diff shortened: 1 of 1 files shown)


prompt-service/src/unstract/prompt_service/main.py
```diff
@@ -693,29 +693,51 @@ def run_retrieval(  # type:ignore
     retrieval_type: str,
     metadata: dict[str, Any],
 ) -> tuple[str, str]:
+    context: str = ""
     prompt = output[PSKeys.PROMPTX]
     if retrieval_type == PSKeys.SUBQUESTION:
```
Deepak-Kesavan, line 698: Any reason why if else is not used?

chandrasekharan-zipstack (approved):
Do you have any screenshots of this change improving the responses in prompt studio?

prompt-service/src/unstract/prompt_service/main.py
harini-venkataraman: If-else for the retrieval type?  If any other retrieval types are added, this will have the logic specific.

gaya3-zipstack (approved):
Changes look good

Deepak-Kesavan (approved):

prompt-service/src/unstract/prompt_service/main.py
```diff
@@ -693,29 +693,51 @@ def run_retrieval(  # type:ignore
...
-        subq_prompt = (
-            f"Generate a sub-question from the following verbose prompt that will"
-            f" help extract relevant documents from a vector store:\n\n{prompt}"
+        subq_prompt: str = (
+            f"I am sending you a verbose prompt {prompt} . "
```
Deepak-Kesavan, line 700: Better to separate the prompt using `\n\n` at the start and end.

prompt-service/src/unstract/prompt_service/main.py
```diff
@@ -693,29 +693,51 @@ def run_retrieval(  # type:ignore
...
+            "Ensure that each subquestion is distinct and relevant"
+            "to the different facets of the original query. Do not"
+            "add subquestions for details not mentioned in the original "
+            "prompt. The goal is to maximize retrieval accuracy"
+            " using these subquestions. Please note that, there are cases where the "
+            "response might have a list of answers. The subquestions must not miss out "
+            "any values in these cases. "
+            "Output should be a list of comma seperated "
+            "subquestion prompts. Do not change the format "
+            " subquestions : \n \n"
```
Deepak-Kesavan, line 713: The final `\n\n` won't be required here

prompt-service/src/unstract/prompt_service/main.py
harini-venkataraman: How will the preceding `\n` help here?

Deepak-Kesavan: Spacing makes it better for LLM to understand it better. We need to keep the prompt seperately apart from the system prompt like this.

```I am sending you a verbose prompt:

{prompt}

....
```

harini-venkataraman: Get that, addressed.

prompt-service/src/unstract/prompt_service/main.py
```diff
@@ -693,29 +693,51 @@ def run_retrieval(  # type:ignore
...
-        # query with doc_id fails even though indexing just happened.
-        # This causes the following retrieve to return no text.
-        # To rule out any lag on the Pinecone vector DB write,
-        # the following sleep is added
-        # Note: This will not fix the issue. Since this issue is inconsistent
-        # and not reproducible easily, this is just a safety net.
-        time.sleep(2)
+        subquestion_list = subquestions.split(",")
+        for each_subq in subquestion_list:
+            retrived_context = _retrieve_context(
```
hari-kuriakose, line 721: @harini-venkataraman Minor typo in variable name.

prompt-service/src/unstract/prompt_service/main.py
```diff
@@ -693,29 +693,51 @@ def run_retrieval(  # type:ignore
...
+                output, doc_id, vector_index, each_subq
+            )
+            context = "".join([context, retrived_context])
+
+    if retrieval_type == PSKeys.SIMPLE:
+
         context = _retrieve_context(output, doc_id, vector_index, prompt)
 
+        if not context:
```
hari-kuriakose, line 730: @harini-venkataraman Shouldn't this check happen after retrieving sub-question context above too?

hari-kuriakose (changes requested):
@harini-venkataraman LGTM overall.

Added a comment though, please check.

prompt-service/src/unstract/prompt_service/main.py
harini-venkataraman: @gaya3-zipstack Can you please confirm if the inconsistent issue related to pinecone  still exists? 
If so, the sub question retrieval fetches context for every iteration, let me if this check is needed here?

prompt-service/src/unstract/prompt_service/main.py
harini-venkataraman: Thanks @hari-kuriakose, addressed.

prompt-service/src/unstract/prompt_service/main.py
gaya3-zipstack: As per my knowledge, nothing might have changed w.r.t  Pinecone there @harini. But is that what @hari-kuriakose is asking about? I am not clear.

harini-venkataraman: @hari-kuriakose As per discussion with @gaya3-zipstack, a pointers.
1. The sleep added here is only a potential fix for inconsistency in Pinecone serverless retrieval. We cannot afford to add this sleep in a loop, expecting a infinite loop.
2. Also, since the sub questions are returned from LLM, we will never have a way to differentiate genuine cases of empty context. 

Considering these challenges and niche usecase, we can leave it this way. Added a comment in code for good developer experience.

hari-kuriakose: @harini-venkataraman One thing though, the loop will be always finite. We will do this check only once at max for each sub-question as per existing logic.

hari-kuriakose (approved):
@harini-venkataraman  LGTM overall.

It would be also good to keep the generated number of sub-questions in check, otherwise the delay or cost can get unnecessarily high without noticeable improvement in quality. In this regard, it might be better to keep a max limit of 10 or so for the number of generated questions.

harini-venkataraman:
> @harini-venkataraman LGTM overall.
> 
> It would be also good to keep the generated number of sub-questions in check, otherwise the delay or cost can get unnecessarily high without noticeable improvement in quality. In this regard, it might be better to keep a max limit of 10 or so for the number of generated questions. We could do this later if required.

Noted @hari-kuriakose. Will add this to the prompt in a separate PR.
Tracking it [here](https://zipstack.atlassian.net/browse/UN-1700)

prompt-service/src/unstract/prompt_service/main.py
harini-venkataraman: @hari-kuriakose Thanks for catching that.

PR #646, opened 2024-09-03
feat: Removed dead code from unstract-core

chandrasekharan-zipstack:
## What

- Removed old WF generation method
- Removed unused modules from core

## Why

- Dead code removal

- Yes, this involves a schema migration as well and its code removed from unused methods. I've done a review on the data being accessed and its only valid for WFs created with the very initial implementations which has become entirely obsolete now

## Database Migrations

- Schema migration for removing fields `prompt_name` and `prompt_text` from the `workflow` model

## Env Config

- Removed some unused envs that was previously used for workflow generation

## Notes on Testing

- Able to run a classifier workflow with API as source

## Screenshots
image

## Checklist


Deepak-Kesavan (approved):

backend/workflow_manager/workflow/migrations/0007_remove_workflow_prompt_name_and_more.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 4.2.1 on 2024-09-03 08:17
```
muhammad-ali-e, line 1: We don't need migrations in V1

chandrasekharan-zipstack: Got it, will remove this migration and model change in the v1 version then

muhammad-ali-e (approved):
LGTM overall.. 
Note: We don't any code changes and migrations in v1.

PR #649, opened 2024-09-03
[Do not merge] When chunk_size=0, skip vector db

gaya3-zipstack:
## What

Current; implementation always uses the indexed nodes when fetching context for prompts. However, when chunk_size=0, since we have to send the entire context, we can directly send the extracted text instead of fetching the chunk from the vector db.

## Why

This will improve response time for prompts when chunk_size=0 as vector db need not be accessed

## How

When chunk_size=0, the context can be fetched from the extracted text present in the container file system

Maybe. Testing required around prompt studio for chunk sizes = 0 and > 0. Suggested to test with plugins also.

## Database Migrations

- None

## Env Config

- NA

## Relevant Docs

-

## Related Issues or PRs

- https://zipstack.atlassian.net/browse/UN-1418

SDK PR
- https://github.com/Zipstack/unstract-sdk/pull/99

## Dependencies Versions

-

## Notes on Testing

- 

## Screenshots

Profile with Chunk_size=0
Manual indexing on a document. Here after indexing is completed, no nodes are added to the vector DB as shown
image

Prompt run on top of manual indexing. Here after prompt run, still no records in the vector db. But still, prompt answers are right as the context gets picked up from the extracted text and works fine.
image

Running a prompt before manual indexing (dynamic indexing would kick in). 
image

image

Manually remove the extracted file after indexing. Run prompt. This gives an error saying the extracted file is missing
image

Now, do a manual re-indexing. Extracted file will be re-created. Then run prompt.

image

image

image

Profile with chunk_size =1024

Manual indexing on a document. Here after indexing is completed, nodes are added to the vector DB as shown
image

Prompt run on top of manual indexing. Prompt run works fine picking context from vector DB.
image

Running a prompt before manual indexing (dynamic indexing would kick in) as there are no records in vector db.
image

Dynamic indexing kicked in and prompt run worked fine

image

Manually remove the records from vector db
image

On running prompt, we see an error
image

Manually re-index. Run prompt again and prompt should work fine. Nodes added to vector DB.
image

image

 

## Checklist

### DIFF
--- a/prompt-service/src/unstract/prompt_service/main.py
+++ b/prompt-service/src/unstract/prompt_service/main.py
@@ -245,7 +245,7 @@ def prompt_processor() -> Any:
                 RunLevel.RUN,
                 "Unable to obtain LLM / embedding / vectorDB",
             )
-            return APIError(message=msg)
+            raise APIError(message=msg)
 
         if output[PSKeys.TYPE] == PSKeys.TABLE:
             try:
@@ -323,7 +323,7 @@ def prompt_processor() -> Any:
                         RunLevel.RUN,
                         "Extracted file not present.",
                     )
-                    return APIError(message=msg)
+                    raise APIError(message=msg)
 
                 # TODO: Use vectorDB name when available
                 _publish_log(
--- a/prompt-service/src/unstract/prompt_service/main.py
+++ b/prompt-service/src/unstract/prompt_service/main.py
@@ -3,7 +3,7 @@ import time
 import traceback
 from enum import Enum
 from json import JSONDecodeError
-from typing import Any, Optional
+from typing import Any
 
 from flask import json, jsonify, request
 from llama_index.core.vector_stores import ExactMatchFilter, MetadataFilters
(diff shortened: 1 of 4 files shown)


prompt-service/src/unstract/prompt_service/main.py
```diff
@@ -314,11 +299,32 @@ def prompt_processor() -> Any:
...
+                        {
+                            "tool_id": tool_id,
+                            "prompt_key": prompt_name,
+                            "doc_name": doc_name,
+                        },
+                        LogLevel.ERROR,
+                        RunLevel.RUN,
+                        "Extracted file not present.",
+                    )
+                    return APIError(message=msg)
```
gaya3-zipstack, line 326: @chandrasekharan-zipstack Currently this error is showing up as html. How do I throw the error in the right format?

chandrasekharan-zipstack: Let's huddle for this, didn't get you here @gaya3-zipstack

prompt-service/src/unstract/prompt_service/main.py
```diff
@@ -278,34 +282,15 @@ def prompt_processor() -> Any:
...
-                        usage_kwargs=usage_kwargs,
-                    )
-                    if context is None:
-                        # TODO: Obtain user set name for vector DB
-                        msg = NO_CONTEXT_ERROR
-                        app.logger.error(
-                            f"{msg} {output[PSKeys.VECTOR_DB]} for doc_id {doc_id}"
+                try:
+                    # Read from extract_file_path and set that as context
+                    with open(extract_file_path) as file:
```
chandrasekharan-zipstack, line 287: @gaya3-zipstack does this work because the volumes are shared between backend and prompt-service? Was this change tested against a tool run in a workflow or a pipeline?

harini-venkataraman:
@gaya3-zipstack I suppose, this PR is not valid at the moment. Please close if not needed.

gaya3-zipstack:
Needs a re-haul...

PR #652, opened 2024-09-04
feat: Validation for cron string to be minimum an hour

chandrasekharan-zipstack:
## What

- BE changes for validating a cron string to ensure it is minimum 1 hour
- NOTE: No migration is carried out for existing projects 
    - Because of the complexity of the migration
    - Less usage of deployed pipelines currently
- FE fix for displaying the alert popup message

## Why

- Pipeline executions with no limits on the cron schedule can break our system and this is done to prevent it

## How

-

- No, it only performs validation on creating / editing a pipeline

## Database Migrations

- 

## Env Config

- 

## Dependencies Versions

-

## Notes on Testing

- Able to view the message in the UI, however checked for a handful of cases such as 
`0 * * * *`, `5 * * * *`, `51 * * * *`,  `0,16,26 2 * * *`, `* * * * *` - throw error

## Screenshots
image

## Checklist


Deepak-Kesavan (approved):

jagadeeswaran-zipstack (approved):
FE LGTM

backend/pipeline/serializers/crud.py
```diff
@@ -66,6 +66,15 @@ def validate_cron_string(self, value: Optional[str] = None) -> Optional[str]:
             logger.error(f"Invalid cron string '{cron_string}': {error}")
             raise serializers.ValidationError("Invalid cron string format.")
 
+        # Check if the frequency is less than 1 hour
+        cron_parts = cron_string.split()
+        minute_field = cron_parts[0]
+        if minute_field == "*" or any(char in minute_field for char in [",", "-", "/"]):
+            raise serializers.ValidationError(
+                "Cron schedule can not be more than once per hour. Please provide a "
+                "cron schedule to run at an hourly or less frequent interval."
```
ritwik-g, line 75: @chandrasekharan-zipstack if we check only the minute field would that be enough for this check?

chandrasekharan-zipstack: @ritwik-g with this check we ensure that we allow minute field to be a number alone. This means that every cron schedule would have 
- `0 * ..`  Run 0th minute for every hour... or
- `15 */2 ...` - Run at 15th minute once in every 2 hours ... and so on

We prevent the undesirable ones of `*` - run every minute and `,`, `-`, `/` - allows running for multiple minutes.
Try using [this tool](https://crontab.guru/) to come up with any values that might fail this check

PR #662, opened 2024-09-06
HOTFIX: Trigger logout when mandatory API fails.

tahierhussain:
## What

1. Log the user out if any one of the mandatory APIs fails.
2. Passed a timestamp in the query param for the mandatory APIs.

## Why

The session cookies were not being cleared earlier when one of the mandatory session APIs failed.

## How

The session cookies are now cleared by triggering the logout function when any of the mandatory session APIs fail.

Yes, the major changes in this PR are in the `useSessionValid` hook, which is responsible for allowing the user to access authenticated pages.

## Database Migrations

NA

## Env Config

NA

## Relevant Docs

NA

## Related Issues or PRs

NA

## Dependencies Versions

NA

## Notes on Testing

NA

## Screenshots

NA

## Checklist

### DIFF
--- a/frontend/src/hooks/useExceptionHandler.jsx
+++ b/frontend/src/hooks/useExceptionHandler.jsx
@@ -1,8 +1,6 @@
-import { useNavigate } from "react-router-dom";
 import PropTypes from "prop-types";
 
 const useExceptionHandler = () => {
-  const navigate = useNavigate();
   const handleException = (
     err,
     errMessage = "Something went wrong",
@@ -36,7 +34,6 @@ const useExceptionHandler = () => {
           }
           break;
         case "subscription_error":
-          navigate("/trial-expired");
           return {
             title: title,
             type: "error",
--- a/frontend/src/hooks/useExceptionHandler.jsx
+++ b/frontend/src/hooks/useExceptionHandler.jsx
@@ -1,6 +1,8 @@
+import { useNavigate } from "react-router-dom";
 import PropTypes from "prop-types";
 
 const useExceptionHandler = () => {
+  const navigate = useNavigate();
   const handleException = (
     err,
     errMessage = "Something went wrong",
@@ -34,6 +36,7 @@ const useExceptionHandler = () => {
           }
           break;
         case "subscription_error":
+          navigate("/trial-expired");
           return {
             title: title,
             type: "error",
--- a/frontend/src/hooks/useSessionValid.js
+++ b/frontend/src/hooks/useSessionValid.js
@@ -36,10 +36,10 @@ function useSessionValid() {
   const navigate = useNavigate();
   const userSession = useUserSession();
   const logout = useLogout();
-  const timestamp = new Date().getTime();
 
   return async () => {
     try {
+      const timestamp = new Date().getTime();
       const userSessionData = await userSession();
 
       // Return if the user is not authenticated
--- a/frontend/src/hooks/useSessionValid.js
+++ b/frontend/src/hooks/useSessionValid.js
@@ -38,6 +38,12 @@ function useSessionValid() {
   return async () => {
     try {
       const userSessionData = await userSession();
+
+      // Return if the user is not authenticated
+      if (!userSessionData) {
+        return;
+      }
+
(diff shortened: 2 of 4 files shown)


athul-rs (approved):
LGTM. 
Tested this once in my system with @tahierhussain

frontend/src/hooks/useSessionValid.js
vishnuszipstack, line 1: @tahierhussain Can we add conditions to check if mandatory values are present in the organization's API response or any other responses? If a mandatory value, such as orgId, is missing or null in the response, we can call the logout function and clear the data. This is a safeguard to ensure that we are not allowing access in such cases.

jagadeeswaran-zipstack (approved):

frontend/src/hooks/useExceptionHandler.jsx
```diff
@@ -36,7 +34,6 @@ const useExceptionHandler = () => {
           }
           break;
         case "subscription_error":
-          navigate("/trial-expired");
```
hari-kuriakose, line 39: @tahierhussain Don't we still need to show trial expired if subscription error?

frontend/src/hooks/useSessionValid.js
```diff
@@ -141,10 +150,10 @@ function useSessionValid() {
       // Set the session details
       setSessionDetails(getSessionData(userAndOrgDetails));
     } catch (err) {
-      // TODO: Throw popup error message
-      // REVIEW: Add condition to check for trial period status
+      setAlertDetails(handleException(err));
       if (err.response?.status === 402) {
-        handleException(err);
+        navigate("/trial-expired");
+        return;
```
jagadeeswaran-zipstack, line 156: Will this check here catchs trail expired error from other endpoints?

frontend/src/hooks/useSessionValid.js
ritwik-g: @vishnuszipstack I agree. But let do it as an improvement after this is merged back to main

frontend/src/hooks/useSessionValid.js
```diff
@@ -34,6 +35,8 @@ function useSessionValid() {
   const { setAlertDetails } = useAlertStore();
   const navigate = useNavigate();
   const userSession = useUserSession();
+  const logout = useLogout();
+  const timestamp = new Date().getTime();
```
hari-kuriakose, line 39: @tahierhussain One doubt, will the timestamp be regenerated each time a request is made?
Or will it happen only once when this component is mounted the first time? If so, the timestamp will remain the same thus effectively having no difference with not having a timestamp at all.

frontend/src/hooks/useSessionValid.js
tahierhussain: @vishnuszipstack @ritwik-g I added a basic check for the `orgId` and `csrfToken`. If either is not available, the code will enter the `catch` block where the user is logged out. Further improvements will be made in the next PR.

frontend/src/hooks/useExceptionHandler.jsx
athul-rs: @hari-kuriakose There is another bug here in front-end, because in code we are calling navigate via alert function the navigate added will not be triggered. This was noticed during test with @tahierhussain and is being taken up as a separate fix and PR.

@tahierhussain Will be moving the navigate and error capture in the axios interceptor for better handling of the same. cc: @vishnuszipstack

frontend/src/hooks/useSessionValid.js
tahierhussain: @hari-kuriakose With our current approach, we call the hook method only once when a component is mounted, so it shouldn't be a problem. However, as a safeguard, I have moved the timestamp inside the function. This way, a new timestamp will be generated even if the component re-renders and requires calling the function again.

frontend/src/hooks/useSessionValid.js
tahierhussain: @jagadeeswaran-zipstack Addressed in this comment: https://github.com/Zipstack/unstract/pull/662#discussion_r1746686818

vishnuszipstack (approved):

PR #666, opened 2024-09-09
fix: Used WF exec ID for tool container name to allow better tracing

chandrasekharan-zipstack:
## What

- Introduced `run_id` for each file being processed in a workflow
- Updated tool containers to include run ID to track tool runs
- Renamed `worker` to `runner` in the docker compose build file

## Why

- Difficult to trace tool container logs since this execution ID was not used to uniquely identify the container
- Reason to rename to `runner`
image

- Yes since it involves changes in workflow execution - however it has been tested with execution involving multiple files

## Notes on Testing

- Tested locally against the run of a WF

## Screenshots
Older screenshots
image
image

Newer screenshots highlighting use of `run_id`
1. In backend
image
1. In runner
image

## Checklist


Deepak-Kesavan:
@chandrasekharan-zipstack One test case is failing. Please check that.

worker/src/unstract/worker/clients/helper.py
```diff
@@ -18,5 +17,5 @@ def get_container_client() -> ContainerClientInterface:
         return import_module(client_path).Client
 
     @staticmethod
-    def normalize_container_name(name: str) -> str:
-        return name.replace("/", "-") + "-" + str(uuid.uuid4())
+    def normalize_container_name(name: str, execution_id: str) -> str:
+        return name.replace("/", "-") + "-" + execution_id
```
muhammad-ali-e, line 21: I think one execution will conatins multiple container, each container for each file. So it might be an issue if we add execution Id here

muhammad-ali-e (changes requested):

worker/src/unstract/worker/clients/helper.py
chandrasekharan-zipstack: @muhammad-ali-e do we have a UUID that uniquely identifies each file's run that's stored or tracked anywhere?

hari-kuriakose: @chandrasekharan-zipstack Let's have the container name as `{tool_name}-{execution_id}-container-{short_uuid}`.

Then inside such a container instance, let's prefix the container name to each log output. Try the following:

- Do equivalent of `docker run --name  "{tool_name}-{execution_id}-container-{short_uuid}" alpine:latest`
- Then get value of env var `NAME` in Python code.

chandrasekharan-zipstack: @hari-kuriakose @muhammad-ali-e @ritwik-g I've updated this PR to provide the below container name
`{tool_name}-{run_id}` - this is to limit the length of the container names and its still traceable by looking up the associated execution_id in the runner / backend logs.
Check the screenshots in the PR for more context

unstract/core/src/unstract/core/utilities.py
```diff
@@ -23,3 +23,9 @@ def get_env(env_key: str, default: Optional[str] = None, raise_err=False) -> str
         if env_value is None or env_value == "" and raise_err:
             raise RuntimeError(f"Env variable {env_key} is required")
         return env_value
+
+    @staticmethod
+    def build_tool_container_name(
+        tool_image: str, tool_version: str, run_id: str
+    ) -> str:
+        return f"{tool_image.split('/')[-1]}-{tool_version}-{run_id}"
```
ritwik-g, line 31: We need to make sure the total length doesn't exceed 63 chars. This is required for k8s

ritwik-g (approved):

worker/src/unstract/worker/clients/helper.py
muhammad-ali-e: @chandrasekharan-zipstack unfortunately ,currently we not other than filehash. But it not suggested for tracing ID.

muhammad-ali-e (approved):
Overall LGTM,

unstract/core/src/unstract/core/utilities.py
chandrasekharan-zipstack: I can add a check for this in a separate PR but this will be guaranteed for all our existing tool image names now unless we introduce an excessively long name

worker/src/unstract/worker/clients/helper.py
chandrasekharan-zipstack: Understood @muhammad-ali-e - we can differentiate that when we bring back multiple tool execution. I've added a comment for this as well. Resolving this to allow merging for now

unstract/core/src/unstract/core/utilities.py
chandrasekharan-zipstack: Raised https://github.com/Zipstack/unstract/pull/697

PR #678, opened 2024-09-11
fix/test-azure-gcs-wf-etl

kirtimanmishrazipstack:
## What

- Error Handling for azure fs 

## Why

- Azure doesn't allow directory(container) name to be invalid dns. Reference https://learn.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#container-names 

## How

- When  the above error happens, we will through error to user 

- NO

## Database Migrations

- N/A

## Env Config

- N/A

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

- Tested workflow for both azure fs and gcs

## Screenshots

- 
Screenshot from 2024-09-16 11-41-11

## Checklist


johnyrahul (approved):

gaya3-zipstack (approved):
Looks good.

unstract/connectors/src/unstract/connectors/exceptions.py
```diff
@@ -31,3 +31,11 @@ def __init__(
 
     def __str__(self) -> str:
         return f"{self.message}"
+
+
+class AzureHttpError(ConnectorError):
```
chandrasekharan-zipstack, line 36: NIT: This can be placed along with the `azure_cloud_storage` module itself

unstract/connectors/src/unstract/connectors/filesystems/azure_cloud_storage/azure_cloud_storage.py
```diff
@@ -72,3 +74,36 @@ def test_credentials(self) -> bool:
...
+
+    def raise_http_exception(
+        self, e: AzureException.HttpResponseError, path: str
+    ) -> AzureHttpError:
+        user_message = f"Error from Azure Cloud Storage connector. {e.reason} "
+        if e.status_code == 400:
+            user_message = (
+                f"Error from Azure Cloud Storage connector. "
+                f"Invalid resource name for path '{path}'. {e.reason}"
+            )
```
chandrasekharan-zipstack, line 105: @kirtimanmishrazipstack what if this 400 status code is for an invalid value of a different field? Won't the `e.reason` be different in that case?

kirtimanmishrazipstack: @chandrasekharan-zipstack I tried using

if e.error_code.name == "INVALID_RESOURCE_NAME": # type: ignore
but it was throwing `Pylance` issue. So used status_code 400.

This is more declariative but has pylance issue

kirtimanmishrazipstack: handled @chandrasekharan-zipstack

backend/backend/exceptions.py
```diff
@@ -39,3 +39,15 @@ def custom_exception_handler(exc, context) -> Response:  # type: ignore
...
+class UnstractFSException(UnstractBaseException):
+    """Handles all error from fs connector class and propagates the error
+    message to UI.
+
+    Args:
+        UnstractBaseException: Inherits class UnstractBaseException
+    """
+
+    default_detail = "Error testing connection. "
+    status_code = 500
```
chandrasekharan-zipstack, line 53: @kirtimanmishrazipstack might be okay for now, however we need to correctly represent these status codes as well to avoid unwanted 500 alerts.
For example, this can be a case of 400 also with an invalid path entered by the user. In order to effect this we might have to pass the status code also from the connector package.

cc: @hari-kuriakose

chandrasekharan-zipstack (approved):

PR #693, opened 2024-09-14
FIX: Optimized the Prompt Output API Calls

tahierhussain:
## What

- Made a common API call to fetch the outputs for all the prompts and store them in a global state.
- Modified the response of the prompt run API to return the necessary attributes, eliminating the need for a separate prompt output API call after each prompt run.

## Why

Previously, the architecture of the prompt studio involved making API calls to retrieve the prompt output for each prompt card. This approach made sense with our earlier design, where users could select different LLM profiles for each card, and the corresponding output would be displayed in the output section. However, after implementing the Multi-LLM feature, where outputs for all LLM profiles are displayed in the UI, the old architecture became inefficient due to unoptimized API calls to fetch the prompt output.

## How

To address this inefficiency, we've introduced an approach where a single API call is made to fetch all prompt outputs when loading the prompt studio project. The prompt outputs API call will be triggered again only if the user selects a new document or changes the default LLM profile (applicable only for single-pass extraction).

Yes, this PR includes significant changes to the architecture related to fetching and displaying the prompt output, so there is a possibility that it could break existing features.

## Database Migrations

NA

## Env Config

NA

## Relevant Docs

NA

## Related Issues or PRs

NA

## Dependencies Versions

NA

## Notes on Testing

NA

### Only a single prompt run API call was triggered when loading the prompt studio project
image

## Checklist


Deepak-Kesavan (approved):

frontend/src/components/custom-tools/prompt-card/Header.jsx
```diff
@@ -224,7 +223,7 @@ function Header({
           <PromptRunBtnSps
             spsLoading={spsLoading}
             handleSpsLoading={handleSpsLoading}
-            handleGetOutput={handleGetOutput}
+            handleGetOutput={() => {}}
```
jagadeeswaran-zipstack, line 226: @tahierhussain Do we need this void func here?

jagadeeswaran-zipstack (changes requested):

frontend/src/components/custom-tools/prompt-card/Header.jsx
tahierhussain: @jagadeeswaran-zipstack Yes. This needs to be addressed for the SPS. I have kept the SPS component changes on hold and will address them in a separate PR, as it is not a blocker.

jagadeeswaran-zipstack (approved):

PR #697, opened 2024-09-17
feat: Tool container name validation for length < 63

chandrasekharan-zipstack:
**NOTE: v2 change not required**
## What

- Validating tool container name length to ensure its < 63 chars

## Why

- Addressing ask related to #666 
- In case of runner clients like K8s it'll be an issue

- No, minor log change and string slicing involved

## Related Issues or PRs

- #666 

## Notes on Testing

- No explicit testing done to ensure this warning is raised

## Checklist

### DIFF
--- a/unstract/core/src/unstract/core/utilities.py
+++ b/unstract/core/src/unstract/core/utilities.py
@@ -1,5 +1,8 @@
+import logging
 import os
 from typing import Optional
+
+logger = logging.getLogger()
 
 
 class UnstractUtils:
@@ -28,4 +31,12 @@ class UnstractUtils:
     def build_tool_container_name(
         tool_image: str, tool_version: str, run_id: str
     ) -> str:
-        return f"{tool_image.split('/')[-1]}-{tool_version}-{run_id}"
+        container_name = f"{tool_image.split('/')[-1]}-{tool_version}-{run_id}"
+
+        # To support limits of container clients like K8s
+        if len(container_name) > 63:
+            logger.warning(
+                f"Container name exceeds 63 char limit for '{container_name}', "
+                "truncating to 63 chars. There might be collisions in container names"
+            )
+        return container_name[:63]


unstract/core/src/unstract/core/utilities.py
```diff
@@ -28,4 +31,12 @@ def get_env(env_key: str, default: Optional[str] = None, raise_err=False) -> str
     def build_tool_container_name(
         tool_image: str, tool_version: str, run_id: str
     ) -> str:
-        return f"{tool_image.split('/')[-1]}-{tool_version}-{run_id}"
+        container_name = f"{tool_image.split('/')[-1]}-{tool_version}-{run_id}"
+
+        # To support limits of container clients like K8s
```
muhammad-ali-e, line 36: Actually, will the name even exceed 64 characters?

If the `run_id` is a UUIDv4, it will be 36 characters long. The `tool_version` will likely be around 6 characters (e.g., `xx.xx.xx`), and adding a hyphen gives us a total of 43 characters. This leaves about 20 characters for `tool_image`.

I believe it's important to keep the `run_id` in the container name for identification purposes. Trimming the `run_id` could lead to potential name conflicts, so it's better to trim the first part (i.e., `tool_image` or its name).

That said, it seems to be okay for now. We can revisit this if the need arises.

muhammad-ali-e (approved):
LGTM overall. Left a comment , please check it

unstract/core/src/unstract/core/utilities.py
chandrasekharan-zipstack: I had a similar concern on whether this will be needed - anyway added a warning here in case we face issues as a result of this for us to act later

ritwik-g: Better to have it. So let's keep this change

PR #698, opened 2024-09-17
FIX: Optimize Timer Handling for Prompt Run API Calls

tahierhussain:
## What

This PR introduces optimizations to how timers are handled in PromptCard component when the prompt run API call is triggered

## Why

Previously, each `PromptCard` initiated a `setInterval` to update the timer every second during the API call. Multiple `setInterval` instances created concurrently caused excessive component re-renders and increased CPU load when several prompt runs were triggered simultaneously.

## How

The `setInterval` logic has been replaced with a more efficient timestamp-based approach. When the API call is triggered, a `startTime` timestamp is captured. Upon API completion, an `endTime` timestamp is captured. The difference between `startTime` and `endTime` is calculated, converted into seconds, and displayed in the card. This eliminates the need for frequent re-renders during the `setInterval` logic.

Yes, this PR includes changes to how the time taken for a prompt run is handled. These modifications may impact other features within the `PromptCard` component.

## Database Migrations

NA

## Env Config

NA

## Relevant Docs

NA

## Related Issues or PRs

NA

## Dependencies Versions

NA

## Notes on Testing

NA

### Time taken to run each prompt is displayed in the below screenshot:
Screenshot from 2024-09-17 15-52-28

## Checklist

### DIFF
--- a/frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
+++ b/frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
@@ -281,10 +281,6 @@ function PromptCard({
     }));
   };
 
-  const getTokenUsageId = (promptId, docId, profileManagerId) => {
-    return promptId + "__" + docId + "__" + profileManagerId;
-  };
-
   // Generate the result for the currently selected document
   const handleRun = (
     profileManagerId,
@@ -380,6 +376,7 @@ function PromptCard({
           selectedLlmProfiles.includes(profile?.profile_id)
         );
       }
+
       for (const profile of selectedProfiles) {
         handleDocOutputs(
           docId,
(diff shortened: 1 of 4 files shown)


jagadeeswaran-zipstack (approved):

frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
```diff
@@ -613,22 +627,13 @@ function PromptCard({
 
     if (profileManagerId) {
       body.profile_manager = profileManagerId;
-      let tokenUsageId;
       let url = `/api/v1/unstract/${sessionDetails?.orgId}/prompt-studio/fetch_response/${details?.tool_id}`;
       if (!isSimplePromptStudio) {
         body["run_id"] = runId;
-        tokenUsageId = getTokenUsageId(promptId, docId, profileManagerId);
```
harini-venkataraman, line 620: @tahierhussain Minor doubt : Will there be an impact in invoking token usage api?

harini-venkataraman (approved):
LGTM, left a minor comment for clarification.

frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
tahierhussain: @harini-venkataraman We no longer have to call the token usage API. We get the token usage details along with the prompt output API's response itself. Hope this answers your question.

PR #715, opened 2024-09-19
Included application_name when initializing db

Deepak-Kesavan:
## What

- Included application_name when initializing db
- Updated `docker-compose.yaml` to pass the `APPLICATION_NAME` as env because same backend image is used by multiple services and we need to maintain separate env file for each if we keep it there.

## Why

Improves visibility on which application is connected and queries the db.

## How

Pass through Env and add it to initialization object which passes it as `options` to `postgres` driver.

-

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


ritwik-g (approved):

platform-service/src/unstract/platform_service/controller/platform.py
```diff
@@ -26,6 +26,7 @@
     password=Env.PG_BE_PASSWORD,
     host=Env.PG_BE_HOST,
     port=Env.PG_BE_PORT,
+    options=f"-c application_name={Env.APPLICATION_NAME}",
```
chandrasekharan-zipstack, line 29: @Deepak-Kesavan what does this -c do?

backend/backend/settings/base.py
```diff
@@ -381,6 +381,9 @@ def get_required_setting(
         "PASSWORD": f"{DB_PASSWORD}",
         "PORT": f"{DB_PORT}",
         "ATOMIC_REQUESTS": ATOMIC_REQUESTS,
+        "OPTIONS": {
+            "application_name": os.environ.get("APPLICATION_NAME", ""),
```
chandrasekharan-zipstack, line 385: @Deepak-Kesavan while running the services locally we'll have to configure these values in the respective `sample.env` files. Shall we add it there as well?
If this env is not set, I hope no error will be thrown

chandrasekharan-zipstack (approved):

backend/backend/settings/base.py
Deepak-Kesavan: @chandrasekharan-zipstack I added a reason why it was not added there, in the PR description. I thought of that and finally thought devs can if the need it in local.

chandrasekharan-zipstack: If it causes no harm, my suggestion is to add this env. We can treat these `sample.env`s as a registry of env variables that our application uses. We'll lose track of such undocumented envs too.
Also if there's no error / warning thrown we'd most likely not set this env.

If there's a link / doc on how this `APPLICATION_NAME` can help in debugging issues with the DB - that can be added as a comment in the `sample.env` as well

platform-service/src/unstract/platform_service/controller/platform.py
Deepak-Kesavan: @chandrasekharan-zipstack Check this for more details.

https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-OPTIONS
https://www.postgresql.org/docs/current/config-setting.html#CONFIG-SETTING-SHELL

From my understanding, -c is a commanline arg.

backend/backend/settings/base.py
Deepak-Kesavan: @chandrasekharan-zipstack We are merging envs right. Then it should be commented else it might override the value and result in different application name for celery workers.

hari-kuriakose: @Deepak-Kesavan Merging env is per service and also will give priority to values in `.env` compared to `sample.env`.
This means for first run value will be taken from `sample.env` and on subsequent runs it will normally prefer existing value in `.env` only.

If this is the case, can we add `APPLICATION_NAME` to `sample.env` too as @chandrasekharan-zipstack mentioned?

hari-kuriakose: @Deepak-Kesavan @chandrasekharan-zipstack We can add it to the Contribution guidelines public doc for now, let's see.

PR #727, opened 2024-09-23
Fix: Coverage count in Prompt studio

jagadeeswaran-zipstack:
## What

- Fixed coverage count in Prompt studio
- Fixed when we enable challenge, the challenger result is not shown without cache clear and hard refresh of pagebody
- Fixed prompt results are auto expands which leads to long scrolling
- Fixed the prompts using single pass, the result showing area is not proper

## Why
- Document coverage count not updated on initial page load. 

## How

-

- No

## Database Migrations

- N/A

## Env Config

- N/A

## Relevant Docs

- N/A

## Related Issues or PRs

- N/A

## Dependencies Versions

- N/A

## Notes on Testing

- N/A

## Screenshots
image

## Checklist


vishnuszipstack (approved):

frontend/src/components/custom-tools/prompt-card/PromptCardItems.jsx
```diff
@@ -60,20 +61,24 @@ function PromptCardItems({
     isSimplePromptStudio,
     isPublicSource,
     adapters,
+    defaultLlmProfile,
   } = useCustomToolStore();
+  const { promptOutputs: promptOutput } = usePromptOutputStore();
```
tahierhussain, line 66: @jagadeeswaran-zipstack Can you check if using `promptOutputs` directly from the store is necessary. If `promptOutputs` are already being passed as props, accessing them from the store might cause the entire list of prompt cards to re-render whenever the store changes, negating the intended optimization of selective re-renders.

Check how the promptOutputs is handled in the [DocumentParser.jsx](https://github.com/Zipstack/unstract/blob/e6c025519c2c956e5c002e3c979254b920018ba1/frontend/src/components/custom-tools/document-parser/DocumentParser.jsx#L277) component.

frontend/src/hooks/usePromptOutput.js
```diff
@@ -108,7 +109,7 @@ const usePromptOutput = () => {
         isSinglePass,
         true
       );
-
+      const coverageKey = `coverage_${llmProfile}_${item?.prompt_id}`;
```
tahierhussain, line 112: @jagadeeswaran-zipstack NIT: Since the `coverageKey` is generated in multiple places, consider creating a separate helper function for generating this string to improve code reuse and consistency with the pattern.

tahierhussain (changes requested):

backend/prompt_studio/prompt_studio_output_manager/serializers.py
```diff
@@ -25,4 +26,33 @@ def to_representation(self, instance):
             )
             token_usage = {}
         data["token_usage"] = token_usage
+        # Get the coverage for the current tool_id and profile_manager_id
+        try:
+            # Fetch all relevant outputs for the current tool and profile
+            related_outputs = (
```
chandrasekharan-zipstack, line 32: NIT: Instead of `output` / `outputs` let's rename it to `prompt` / `prompts` to improve the understanding - let's try to correct this variable name eventually. So you could rename this as `related_prompts` and similarly iterate with `for prompt in prompts`

backend/prompt_studio/prompt_studio_output_manager/serializers.py
```diff
@@ -25,4 +26,33 @@ def to_representation(self, instance):
...
+            for output in related_outputs:
+                prompt_key = str(output["prompt_id"])
+                profile_key = str(output["profile_manager_id"])
+                coverage[f"coverage_{profile_key}_{prompt_key}"] = output[
+                    "document_count"
+                ]
+
+            data["coverage"] = coverage
+        except Exception as e:
+            logger.error(
```
chandrasekharan-zipstack, line 52: @jagadeeswaran-zipstack do we intentionally choose to suppress this exception? 
Since we're suppressing the exception 
1. we should either log it as a warning or 
1. mention in the error message that we're continuing processing / suppressing the error
1. Since this won't be caught by the middleware - you'll have to print the traceback as well

Please correct these 3 points in all the exception handling in this serializer

backend/prompt_studio/prompt_studio_output_manager/serializers.py
```diff
@@ -25,4 +26,33 @@ def to_representation(self, instance):
...
+                )
+                .values("prompt_id", "profile_manager_id")
+                .annotate(document_count=Count("document_manager_id"))
+            )
+
+            coverage = {}
+            for output in related_outputs:
+                prompt_key = str(output["prompt_id"])
+                profile_key = str(output["profile_manager_id"])
+                coverage[f"coverage_{profile_key}_{prompt_key}"] = output[
```
chandrasekharan-zipstack, line 46: @jagadeeswaran-zipstack I don't understand what this coverage count exactly is, but when anything in the LLM profile changes shouldn't this coverage key change?

backend/prompt_studio/prompt_studio_output_manager/serializers.py
jagadeeswaran-zipstack: I don't understand why we need to rename this as `prompts` since these data are outputs of each prompt. renaming it as prompt will be more confusing instead `prompt_output` will be more appropriate.

backend/prompt_studio/prompt_studio_output_manager/serializers.py
jagadeeswaran-zipstack: Coverage count is nothing but the number of documents that have been processed and got output. Yes, the coverage key will change if the user changes the LLM profile in the profile manager, and I think that should be the ideal behavior since we're showing coverage to the user based on the LLM profile.

frontend/src/components/custom-tools/prompt-card/PromptCardItems.jsx
jagadeeswaran-zipstack: Understood. I'll make the required changes.

backend/prompt_studio/prompt_studio_output_manager/serializers.py
jagadeeswaran-zipstack: Similar fashion of exception handling is done in this serializer itself, so I thought of following the same. I don't have much context about continuing processing or suppressing the error. It would be helpful if you elaborated more.

backend/prompt_studio/prompt_studio_output_manager_v2/serializers.py
```diff
@@ -25,4 +26,33 @@ def to_representation(self, instance):
...
+            for prompt_output in prompt_outputs:
+                prompt_key = str(prompt_output["prompt_id"])
+                profile_key = str(prompt_output["profile_manager_id"])
+                coverage[f"coverage_{prompt_key}_{profile_key}"] = prompt_output[
+                    "document_count"
+                ]
+
+            data["coverage"] = coverage
+        except Exception as e:
+            logger.error(
```
chandrasekharan-zipstack, line 52: @jagadeeswaran-zipstack either make this a warning log or mention that we'll be continuing processing / suppressing the error in the log message

backend/prompt_studio/prompt_studio_core_v2/serializers.py
```diff
@@ -69,7 +73,26 @@ def to_representation(self, instance):  # type: ignore
...
+                            tool_id=data.get(TSKeys.TOOL_ID),
+                            profile_manager_id=profile_manager.profile_id,
+                            prompt_id=prompt.prompt_id,
+                        )
+                        .values("prompt_id", "profile_manager_id")
+                        .annotate(document_count=Count("document_manager_id"))
+                    )
+                    coverage = {}
+                    for prompt_output in prompt_outputs:
+                        prompt_key = str(prompt_output["prompt_id"])
```
chandrasekharan-zipstack, line 87: NIT: Since these are instances of `PromptStudioOutputManager` please use the `.` (dot) notation to access the values

backend/prompt_studio/prompt_studio_core_v2/serializers.py
```diff
@@ -69,7 +73,26 @@ def to_representation(self, instance):  # type: ignore
             if prompt_instance.count() != 0:
                 for prompt in prompt_instance:
                     prompt_serializer = ToolStudioPromptSerializer(prompt)
-                    output.append(prompt_serializer.data)
+                    prompt_outputs = (
```
chandrasekharan-zipstack, line 76: @jagadeeswaran-zipstack move this coverage calculation logic as a method within the `PromptStudioOutputManager` class. That way, we can avoid code duplication and also let such logic of aggregation / querying be in the model layer itself and serializer can purely be a matter of validation and data representation

backend/prompt_studio/prompt_studio_output_manager_v2/serializers.py
jagadeeswaran-zipstack: Got it

jagadeeswaran-zipstack: got it

backend/prompt_studio/prompt_studio_core_v2/serializers.py
jagadeeswaran-zipstack: Done :+1:

backend/prompt_studio/prompt_studio_output_manager/serializers.py
chandrasekharan-zipstack: I mentioned `prompt` extending from the `Prompt` based model we have and the use of it within the prompt service. `prompt_output` probably works better as you mentioned. We'll worry about this readability later

backend/prompt_studio/prompt_studio_output_manager/serializers.py
chandrasekharan-zipstack: @jagadeeswaran-zipstack anytime we `except Exception` we are catching that exception with an intention of doing something with the exception object. If we don't add this statement and an exception arises from some function call, it gets "propagated" up the call stack where it ultimately gets handled by django / DRF and reported accordingly. Since we don't explicitly `raise e` - this is a case of "suppressing" the error and in this case there won't be any change in the API call returned / in the UI. The only difference being this extra log - and having an ERROR log can wrongly be interpreted as an ERROR which was sent back to the user.

I suggest marking it as a warning log or indicate in the error message that we are continuing the processing

backend/prompt_studio/prompt_studio_output_manager/serializers.py
chandrasekharan-zipstack, line 1: @jagadeeswaran-zipstack is this change in the v1 version of the app necessary? Shouldn't all changes fall in `_v2` apps alone now?

jagadeeswaran-zipstack: No, change in the v1 version of the app not necessary. Yes are needed only in `_v2` apps alone.

backend/prompt_studio/prompt_studio_output_manager/serializers.py
jagadeeswaran-zipstack: Understood, Thanks.

chandrasekharan-zipstack (approved):
LGTM

PR #731, opened 2024-09-24
[FIX] Platform service refactor

Deepak-Kesavan:
## What

- Refactored platform-service

## Why

- Refactor - To improve maintainability

## How

-

- No.  Only refactored the code. The functionality remains the same.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


platform-service/src/unstract/platform_service/utils.py
```diff
@@ -59,3 +61,12 @@ def format_float_positional(value: float, precision: int = 10) -> str:
     """
     formatted: str = f"{value:.{precision}f}"
     return formatted.rstrip("0").rstrip(".") if "." in formatted else formatted
+
+
+class RemoveAccessLogTimestamp(logging.Filter):
```
johnyrahul, line 66: Why are we doing this @Deepak-Kesavan

platform-service/src/unstract/platform_service/config.py
```diff
@@ -0,0 +1,94 @@
...
+    app.register_blueprint(api)
+
+    # Initialize and connect to the database
+    db.init(
+        database=Env.PG_BE_DATABASE,
+        user=Env.PG_BE_USERNAME,
+        password=Env.PG_BE_PASSWORD,
+        host=Env.PG_BE_HOST,
+        port=Env.PG_BE_PORT,
+        options=f"-c application_name={Env.APPLICATION_NAME}",
```
johnyrahul, line 91: What is the purpose of having options?

Deepak-Kesavan: It is to know which service called the database.

platform-service/src/unstract/platform_service/utils.py
Deepak-Kesavan: There was a timestamp hardcoded in the message by 
Werkzeug in the access log and the same is being propagated to gunicorn.access logger. Since we are adding a timestamp from our side, I tried to remove it. 

It worked incase of Werkzeug which is when flask run is used. But when I run with gunicorn, it didn't work as expected. Just left it for now because it is not that much important and we just need access logs for debugging purposes.

harini-venkataraman (approved):

platform-service/sample.env
```diff
@@ -1,3 +1,8 @@
+# Flask
+FLASK_APP=src/unstract/platform_service/run.py
+FLASK_RUN_PORT=3001
```
ritwik-g, line 3: @Deepak-Kesavan do we even need these?

chandrasekharan-zipstack (approved):

platform-service/sample.env
Deepak-Kesavan: This will be picked up when we use flask run. Other than that it won't be used. It can be there in OSS repo alone.

platform-service/src/unstract/platform_service/config.py
```diff
@@ -0,0 +1,94 @@
...
+            },
+            "remove_access_log_timestamp": {
+                "class": "logging.StreamHandler",
+                "formatter": "default",
+                "filters": ["filter_timestamp"],
+            },
+        },
+        "loggers": {
+            "werkzeug": {
+                "level": Env.LOG_LEVEL,
```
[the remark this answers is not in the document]
ritwik-g: My bad it's there

chandrasekharan-zipstack:
@Deepak-Kesavan can you add a screenshot comparing the logs before and after?
Here's the existing logger for plaform-service
image

platform-service/src/unstract/platform_service/extensions.py
```diff
@@ -0,0 +1,3 @@
+from peewee import PostgresqlDatabase
+
+db = PostgresqlDatabase(None)
```
ritwik-g, line 3: @Deepak-Kesavan earlier in the code we were passing some Envs during initial construction itself. Does changing it like this could cause any issues?

platform-service/src/unstract/platform_service/controller/platform.py
```diff
@@ -175,8 +165,8 @@ def page_usage() -> Any:
     )
 
     try:
-        with be_db.atomic() as transaction:
-            be_db.execute_sql(query, params)
+        with db.atomic() as transaction:
```
ritwik-g, line 168: Just a generic doubt. We are only doing single execute statement as far as I can see. Does using atomic even makes any impact here? @hari-kuriakose @johnyrahul @muhammad-ali-e

ritwik-g (commented):
@Deepak-Kesavan when you combine multiple changes in to single PR it becomes very difficult to review. Could you raise it as a 2 separate PRs? One for log change fixes and other one for the db refactoring

platform-service/src/unstract/platform_service/utils.py
ritwik-g: @Deepak-Kesavan let's remove this filter. Adding a filter if not working might just add performance implications. Also the rgex by it's nature is costly operation. So let's remove this filter all together. Having 2 timestamps is something that we can live with. Adding this filter is probably gonna over engineer things and add unnecessary complexity.

platform-service/src/unstract/platform_service/config.py
```diff
@@ -0,0 +1,94 @@
...
+            "default": {
+                "format": (
+                    "[%(asctime)s] %(levelname)s in"
+                    " %(name)s (%(module)s): %(message)s"
+                ),
+                "datefmt": "%Y-%m-%d %H:%M:%S %z",
+            },
+        },
+        "filters": {
+            "filter_timestamp": {"()": RemoveAccessLogTimestamp},
```
ritwik-g, line 29: @Deepak-Kesavan let's remove this filter as I suggested [here](https://github.com/Zipstack/unstract/pull/731/files#r1774476517)

platform-service/src/unstract/platform_service/config.py
```diff
@@ -0,0 +1,94 @@
...
+            "wsgi": {
+                "class": "logging.StreamHandler",
+                "stream": "ext://flask.logging.wsgi_errors_stream",
+                "formatter": "default",
+            },
+            "remove_access_log_timestamp": {
+                "class": "logging.StreamHandler",
+                "formatter": "default",
+                "filters": ["filter_timestamp"],
+            },
```
ritwik-g, line 41: Let's remove this

platform-service/src/unstract/platform_service/config.py
```diff
@@ -0,0 +1,94 @@
...
+                "level": Env.LOG_LEVEL,
+                "handlers": ["remove_access_log_timestamp"],
+                "propagate": False,
+            },
+            "gunicorn.error": {
+                "level": Env.LOG_LEVEL,
+                "handlers": ["remove_access_log_timestamp"],
+                "propagate": False,
+            },
+        },
```
ritwik-g, line 59: @Deepak-Kesavan in the PR description under `How` could you add how these changes fixed the access log issue?

Deepak-Kesavan:
> @Deepak-Kesavan when you combine multiple changes in to single PR it becomes very difficult to review. Could you raise it as a 2 separate PRs? One for log change fixes and other one for the db refactoring

Making the changes now.

platform-service/src/unstract/platform_service/extensions.py
Deepak-Kesavan: It is not an issue @ritwik-g . We are doing the same in prompt service.

https://github.com/Zipstack/unstract/blob/4ee004e9ba64c6ce44e060d8dab765d693a551ab/prompt-service/src/unstract/prompt_service/config.py#L31

Deepak-Kesavan:
This PR now contains platform service refactor without modifying any functionality.

platform-service/src/unstract/platform_service/utils.py
```diff
@@ -59,3 +61,12 @@ def format_float_positional(value: float, precision: int = 10) -> str:
...
+    # '127.0.0.1 - - [24/Sep/2024 06:57:35] "%s" %s %s' -> '192.168.0.102 - "%s" %s %s'
+    pattern: re.Pattern = re.compile(r' - - \[.+?] "')
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        record.msg = self.pattern.sub(' - "', record.msg)
+        return True
```
ritwik-g, line 72: @Deepak-Kesavan let's either remove this or keep it commented if needed for future use

ritwik-g (approved):

platform-service/src/unstract/platform_service/utils.py
Deepak-Kesavan: @ritwik-g Let me remove it.

platform-service/src/unstract/platform_service/controller/platform.py
hari-kuriakose: @ritwik-g If we are running a single query then `autocommit` should be True to maintain integrity.

Only the following are safe:
- Queries running inside Django transactions
- Queries running outside Django transactions but with `autocommit=True` (Django does this by default)

See https://docs.djangoproject.com/en/5.1/topics/db/transactions/#why-django-uses-autocommit

hari-kuriakose:
Security hotspot related to CSRF will be handled separately.

PR #733, opened 2024-09-24
Fix: Added force export to prompt studio tool export

jagadeeswaran-zipstack:
## What

- Added force export to prompt studio tool export
- Fixed Yet to run displayed while prompt output is null

## Why

- We internally set up the default profile during export if the profile is not selected for a prompt, but the user may not have known this or may not have executed the prompt in the default profile. Thus, even if the prompt is not executed, we still permit users to export by force export. 

## How

-

- Yes. Changes made on existing feature might cause issue in export.

## Database Migrations

- N/A

## Env Config

- N/A

## Relevant Docs

- N/A

## Related Issues or PRs

- N/A

## Dependencies Versions

- N/A

## Notes on Testing

- N/A

## Screenshots
image
image

## Checklist

### DIFF
--- a/frontend/src/components/custom-tools/header/Header.jsx
+++ b/frontend/src/components/custom-tools/header/Header.jsx
@@ -1,7 +1,7 @@
 import { SettingOutlined } from "@ant-design/icons";
-import { Button, Tooltip, Typography } from "antd";
+import { Button, Modal, Tooltip, Typography } from "antd";
 import PropTypes from "prop-types";
-import { useState } from "react";
+import { useCallback, useState } from "react";
 
 import { HeaderTitle } from "../header-title/HeaderTitle.jsx";
 import { ExportToolIcon } from "../../../assets";
@@ -53,11 +53,24 @@ function Header({
   const { setPostHogCustomEvent } = usePostHogEvents();
 
   const [toolDetails, setToolDetails] = useState(null);
+  const [confirmModalVisible, setConfirmModalVisible] = useState(false);
+  const [lastExportParams, setLastExportParams] = useState(null);
 
-  const handleExport = (selectedUsers, toolDetail, isSharedWithEveryone) => {
+  const handleExport = (
+    selectedUsers,
+    toolDetail,
+    isSharedWithEveryone,
+    forcedExport = false
+  ) => {
+    setLastExportParams({
+      selectedUsers,
+      toolDetail,
+      isSharedWithEveryone,
+    });
     const body = {
       is_shared_with_org: isSharedWithEveryone,
       user_id: isSharedWithEveryone ? [] : selectedUsers,
+      force_export: forcedExport,
     };
     const requestOptions = {
       method: "POST",
@@ -77,6 +90,10 @@ function Header({
         });
       })
       .catch((err) => {
+        if (err?.response?.data?.errors[0]?.code === "warning") {
+          setConfirmModalVisible(true); // Show the confirmation modal
+          return; // Exit early to prevent any further execution
+        }
         setAlertDetails(handleException(err, "Failed to export"));
       })
       .finally(() => {
@@ -84,6 +101,17 @@ function Header({
         setOpenExportToolModal(false);
       });
   };
+
+  const handleConfirmForceExport = useCallback(() => {
+    if (lastExportParams) {
+      const { selectedUsers, toolDetail, isSharedWithEveryone } =
+        lastExportParams;
+
+      handleExport(selectedUsers, toolDetail, isSharedWithEveryone, true);
+    }
+
+    setConfirmModalVisible(false);
+  }, [lastExportParams]);
 
   const handleShare = (isEdit) => {
     try {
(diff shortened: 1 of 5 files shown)


Deepak-Kesavan (approved):

frontend/src/components/custom-tools/header/Header.jsx
```diff
@@ -53,11 +53,24 @@ function Header({
...
+    selectedUsers,
+    toolDetail,
+    isSharedWithEveryone,
+    forcedExport = false
+  ) => {
+    setLastExportParams({
+      selectedUsers,
+      toolDetail,
+      isSharedWithEveryone,
+    });
```
tahierhussain, line 69: @jagadeeswaran-zipstack Can we set this inside the catch block, only when we want to show the force export modal?

frontend/src/components/custom-tools/header/Header.jsx
```diff
@@ -85,6 +102,17 @@ function Header({
...
 
+  const handleConfirmForceExport = useCallback(() => {
+    if (lastExportParams) {
+      const { selectedUsers, toolDetail, isSharedWithEveryone } =
+        lastExportParams;
+
+      handleExport(selectedUsers, toolDetail, isSharedWithEveryone, true);
+    }
+
+    setConfirmModalVisible(false);
```
tahierhussain, line 113: @jagadeeswaran-zipstack In the above code, If the `if` condition fails, we won't display any error message and will silently close the force export confirmation modal. I don't think that would provide a good user experience. Can you display a relevant error message in the notifications component if the condition fails?

frontend/src/components/custom-tools/header/Header.jsx
```diff
@@ -199,6 +227,21 @@ function Header({
...
+          open={confirmModalVisible}
+          title="Are you sure"
+          okText="Force Export"
+          centered
+        >
+          Unable to export tool. Some Prompt(s) were not run. Please run them
+          before exporting.{" "}
+          <strong>Would you like to force export anyway?</strong>
+        </Modal>
+      )}
```
tahierhussain, line 244: @jagadeeswaran-zipstack, why are we conditionally rendering the modal component? Every time the modal opens and closes, the component will get mounted and unmounted, creating unnecessary load. Wouldn't it be better to simply render it and let Antd's Modal component to take care of the visibility using `open` prop?

tahierhussain (changes requested):

frontend/src/components/custom-tools/header/Header.jsx
```diff
@@ -199,6 +227,21 @@ function Header({
...
+      {confirmModalVisible && (
+        <Modal
+          onOk={handleConfirmForceExport} // Pass the confirm action
+          onCancel={() => setConfirmModalVisible(false)} // Close the modal on cancel
+          open={confirmModalVisible}
+          title="Are you sure"
+          okText="Force Export"
+          centered
+        >
+          Unable to export tool. Some Prompt(s) were not run. Please run them
```
chandrasekharan-zipstack, line 240:
```suggestion
          Unable to export tool. Some prompt(s) were not run. Please run them
```

chandrasekharan-zipstack (commented):
@jagadeeswaran-zipstack so will we no longer see the error related to prompt not being run? Do we have a reason to retain / throw that error in the backend?

frontend/src/components/custom-tools/header/Header.jsx
chandrasekharan-zipstack: As a future improvement, we could also try to list the prompts which were not run

frontend/src/components/custom-tools/header/Header.jsx
jagadeeswaran-zipstack: @tahierhussain In which scenario you think that `if` will condition fail. Even if it fails what error we should throw because its purely state and state being set on catch block and then only we will get to this step.

frontend/src/components/custom-tools/header/Header.jsx
jagadeeswaran-zipstack: This won't add any additional load to component but yes we can let antd handle this.

frontend/src/components/custom-tools/header/Header.jsx
jagadeeswaran-zipstack: @chandrasekharan-zipstack We're not throwing as error but as warning. Can check the exception we set code as warning. In future we can set few more codes like warning, info like that and instead of showing only success and failure popup we can have few more options there.

frontend/src/components/custom-tools/header/Header.jsx
jagadeeswaran-zipstack: Yes. We can move it to catch block.

frontend/src/components/custom-tools/header/Header.jsx
tahierhussain: @jagadeeswaran-zipstack In that case, let's remove this `if` statement entirely. I don't think it's necessary. We're first setting the `lastExportParams` state and then opening the confirm modal, so the state will always have a value. I believe it's safe to remove the condition. Also, as a tip for the future, if we add a condition, let's ensure we handle both the success and failure cases properly.

tahierhussain (approved):
LGTM

frontend/src/components/custom-tools/header/Header.jsx
jagadeeswaran-zipstack: Make sense. Yes it's safe to remove that condition.

PR #745, opened 2024-09-26
feat: Added option to API deployment: use_cache, toggles usage of FileHistory

chandrasekharan-zipstack:
## What

- Added an undocumented option to toggle FileHistory usage for API deployments: `use_cache`
- Propagated this option to workflow execution's celery task
- MINOR: Renamed `meta_data` variable to `metadata`
- MINOR: Removed some unused code from a serializer

NOTE: In case of Workflow execution from the Workflow page - `FileHistory` will still be used for API type sources. This can be handled in a separate PR

## Why

- To avoid caching of file results in case of API deployments

## How

-

- Yes, since it touches parts of workflow execution which is used by several services

## Notes on Testing

- Tested API deployments to notice the effect / absence of caching

## Screenshots
1. Serializer's errors being thrown
image
image
1. First run without cache
image
1. Second run without cache (no speedup observed)
image
1. Third run with cache (this is where data gets stored in the DB)
image
1. Fourth run with cache (cached value returned, notice the elapsed time)
image

## Checklist

### DIFF
--- a/backend/api/api_deployment_views.py
+++ b/backend/api/api_deployment_views.py
@@ -51,7 +51,7 @@ class DeploymentExecution(views.APIView):
         serializer.is_valid(raise_exception=True)
         timeout = serializer.validated_data.get(ApiExecution.TIMEOUT_FORM_DATA)
         include_metadata = serializer.validated_data.get(ApiExecution.INCLUDE_METADATA)
-        use_file_history = serializer.validated_data.get(ApiExecution.USE_CACHE)
+        use_file_history = serializer.validated_data.get(ApiExecution.USE_FILE_HISTORY)
         if not file_objs or len(file_objs) == 0:
             raise InvalidAPIRequest("File shouldn't be empty")
         response = DeploymentHelper.execute_workflow(
--- a/backend/api/api_deployment_views.py
+++ b/backend/api/api_deployment_views.py
@@ -49,10 +49,9 @@ class DeploymentExecution(views.APIView):
         file_objs = request.FILES.getlist(ApiExecution.FILES_FORM_DATA)
         serializer = ExecutionRequestSerializer(data=request.data)
         serializer.is_valid(raise_exception=True)
-        timeout = serializer.get_timeout(serializer.validated_data)
-        include_metadata = (
-            request.data.get(ApiExecution.INCLUDE_METADATA, "false").lower() == "true"
-        )
+        timeout = serializer.validated_data.get(ApiExecution.TIMEOUT_FORM_DATA)
+        include_metadata = serializer.validated_data.get(ApiExecution.INCLUDE_METADATA)
+        use_file_history = serializer.validated_data.get(ApiExecution.USE_CACHE)
         if not file_objs or len(file_objs) == 0:
             raise InvalidAPIRequest("File shouldn't be empty")
         response = DeploymentHelper.execute_workflow(
@@ -61,6 +60,7 @@ class DeploymentExecution(views.APIView):
             file_objs=file_objs,
             timeout=timeout,
             include_metadata=include_metadata,
+            use_file_history=use_file_history,
         )
         if "error" in response and response["error"]:
             return Response(
--- a/backend/api/constants.py
+++ b/backend/api/constants.py
@@ -4,4 +4,4 @@ class ApiExecution:
     FILES_FORM_DATA: str = "files"
     TIMEOUT_FORM_DATA: str = "timeout"
     INCLUDE_METADATA: str = "include_metadata"
-    USE_CACHE: str = "use_cache"  # Undocumented parameter, useful for demos
+    USE_FILE_HISTORY: str = "use_file_history"  # Undocumented parameter
--- a/backend/api/constants.py
+++ b/backend/api/constants.py
@@ -4,3 +4,4 @@ class ApiExecution:
     FILES_FORM_DATA: str = "files"
     TIMEOUT_FORM_DATA: str = "timeout"
     INCLUDE_METADATA: str = "include_metadata"
+    USE_CACHE: str = "use_cache"  # Undocumented parameter, useful for demos
--- a/backend/api/serializers.py
+++ b/backend/api/serializers.py
@@ -81,14 +81,15 @@ class APIKeySerializer(AuditSerializer):
 
 
 class ExecutionRequestSerializer(Serializer):
-    """Execution request serializer timeout (int): Timeout for the API
-    deployment, maximum value can be 300s.
+    """Execution request serializer.
 
-    If -1 it     corresponds to async execution. Defaults to -1
-    include_metadata (bool): Flag to include metadata in API response
-    use_cache (bool): Flag to use FileHistory to save and retrieve
-    responses quickly.     This is undocumented to the user and can be
-    helpful for demos.
+    Attributes:
+        timeout (int): Timeout for the API deployment, maximum value can be 300s.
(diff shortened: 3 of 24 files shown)


backend/api/serializers.py
```diff
@@ -80,26 +81,21 @@ def to_representation(self, instance: APIKey) -> OrderedDict[str, Any]:
...
+    use_cache (bool): Flag to use FileHistory to save and retrieve
+    responses quickly.     This is undocumented to the user and can be
+    helpful for demos.
     """
 
     timeout = IntegerField(
         min_value=-1, max_value=ApiExecution.MAXIMUM_TIMEOUT_IN_SEC, default=-1
     )
-
-    def validate_timeout(self, value: Any) -> int:
```
muhammad-ali-e, line 91: Why this validations removed?

backend/api/serializers.py
```diff
@@ -80,26 +81,21 @@ def to_representation(self, instance: APIKey) -> OrderedDict[str, Any]:
...
-            raise ValidationError("timeout must be a integer.")
-        if value == 0:
-            value = ApiExecution.MAXIMUM_TIMEOUT_IN_SEC
-        return value
-
-    def get_timeout(self, validated_data: dict[str, Union[int, None]]) -> int:
-        value = validated_data.get(ApiExecution.TIMEOUT_FORM_DATA, -1)
-        if not isinstance(value, int):
-            raise ValidationError("timeout must be a integer.")
-        return value
```
ritwik-g, line 102: @chandrasekharan-zipstack are these not used anywhere? Also is it not needed because seriaizer takes care of validation?

backend/api/api_deployment_views.py
```diff
@@ -49,10 +49,9 @@ def post(
         file_objs = request.FILES.getlist(ApiExecution.FILES_FORM_DATA)
         serializer = ExecutionRequestSerializer(data=request.data)
         serializer.is_valid(raise_exception=True)
-        timeout = serializer.get_timeout(serializer.validated_data)
-        include_metadata = (
-            request.data.get(ApiExecution.INCLUDE_METADATA, "false").lower() == "true"
-        )
+        timeout = serializer.validated_data.get(ApiExecution.TIMEOUT_FORM_DATA)
+        include_metadata = serializer.validated_data.get(ApiExecution.INCLUDE_METADATA)
+        use_file_history = serializer.validated_data.get(ApiExecution.USE_CACHE)
```
muhammad-ali-e, line 54: Why we use ApiExecution.USE_CACHE instaed of ApiExecution.USE_FILE_HISTORY. Would be better if we use same name to avoid confusions.

backend/workflow_manager/endpoint/destination.py
```diff
@@ -155,20 +155,19 @@ def handle_output(
         file_name: str,
         file_hash: FileHash,
         workflow: Workflow,
+        input_file_path: str,
         error: Optional[str] = None,
-        input_file_path: Optional[str] = None,
+        use_file_history: bool = True,
```
ritwik-g, line 160: @chandrasekharan-zipstack wouldn't it be better to keep this `False`? We will basically need this only for the API calls during demo right?

Or is this method applicable for both ETL and API?

backend/workflow_manager/endpoint/destination.py
```diff
@@ -155,20 +155,19 @@ def handle_output(
...
         """Handle the output based on the connection type."""
         connection_type = self.endpoint.connection_type
         result: Optional[str] = None
-        meta_data: Optional[str] = None
+        metadata: Optional[str] = None
         if error:
             if connection_type == WorkflowEndpoint.ConnectionType.API:
                 self._handle_api_result(file_name=file_name, error=error, result=result)
             return
-        file_history = FileHistoryHelper.get_file_history(
```
ritwik-g, line 169: @chandrasekharan-zipstack I think we should make this None here. And remove it from the condition. 

Else we could run in to undefined variable error down the line

backend/api/constants.py
```diff
@@ -4,3 +4,4 @@ class ApiExecution:
     FILES_FORM_DATA: str = "files"
     TIMEOUT_FORM_DATA: str = "timeout"
     INCLUDE_METADATA: str = "include_metadata"
+    USE_CACHE: str = "use_cache"  # Undocumented parameter, useful for demos
```
muhammad-ali-e, line 7: Is this comment okay?
@ritwik-g

backend/workflow_manager/endpoint/destination.py
```diff
@@ -180,24 +179,30 @@ def handle_output(
             else:
                 self.insert_into_db(input_file_path=input_file_path)
         elif connection_type == WorkflowEndpoint.ConnectionType.API:
+            file_history = None
+            if use_file_history:
+                file_history = FileHistoryHelper.get_file_history(
+                    workflow=workflow, cache_key=file_hash.file_hash
+                )
```
ritwik-g, line 186: @chandrasekharan-zipstack also if we keep the file_history fetching only here won't we skip this step in other cases?

Should we simply make the file_history None here if not use_file_hisotry?

backend/api/serializers.py
chandrasekharan-zipstack: @ritwik-g I've uploaded some screenshots - DRF's serializer takes care of this validation out of the box. We should add a `validate_*()` method only in case of any custom validation.

backend/workflow_manager/endpoint/destination.py
```diff
@@ -155,20 +155,19 @@ def handle_output(
         file_name: str,
         file_hash: FileHash,
         workflow: Workflow,
+        input_file_path: str,
```
muhammad-ali-e, line 158: Why this changed from Optional to str?

Deepak-Kesavan (approved):

ritwik-g (approved):
Please make the change requested here alone https://github.com/Zipstack/unstract/pull/745/files#r1776895062

backend/workflow_manager/workflow/workflow_helper.py
```diff
@@ -482,6 +492,7 @@ def execute_bin(
         scheduled: bool = False,
         execution_mode: Optional[tuple[str, str]] = None,
         pipeline_id: Optional[str] = None,
+        use_file_history: bool = True,
```
muhammad-ali-e, line 495: NIT: will it get it from kwargs?

backend/api/constants.py
ritwik-g: Okay. Let it be there.

backend/workflow_manager/endpoint/destination.py
chandrasekharan-zipstack: To fix a code smell reported by mypy extension. Always an str is passed here

backend/workflow_manager/workflow/workflow_helper.py
chandrasekharan-zipstack: This will be set and passed when the job is enqueued - [here](https://github.com/Zipstack/unstract/blob/5a36a3f6d028b3804714723b39fcc6101557654f/backend/workflow_manager/workflow/workflow_helper.py#L433)

backend/api/api_deployment_views.py
chandrasekharan-zipstack: Renamed the variable to `use_file_history` everywhere to avoid confusion

hari-kuriakose:
Duplication will be handled when we remove v1 later.

PR #752, opened 2024-09-27
feat: Avoid storing file history for API based wfs

chandrasekharan-zipstack:
## What

- Set and passed `use_file_history` for WF execution in case of API based source
- Included changes for v2 migration as well
- Allows File / DB based runs to still cache the file history (can be reviewed later), API based runs are not stored
- Set API deployment timeout's min value and default to 0 instead of -1

## Why

- To avoid caching file history for API based WF runs

- No, added params with sane defaults

## Related Issues or PRs

- #745 

## Dependencies Versions

-

## Notes on Testing

- Attempted running an API based WF locally

## Checklist


chandrasekharan-zipstack:
Code duplication is due to the v2 migration changes also being present in this PR

backend/workflow_manager/endpoint/source.py
```diff
@@ -329,7 +329,11 @@ def _is_new_file(self, file_path: str, workflow: Workflow) -> bool:
             workflow=workflow, cache_key=file_hash
         )
 
-        if file_history and file_history.is_completed():
+        if (
+            self.execution_service.use_file_history
+            and file_history
+            and file_history.is_completed()
+        ):
```
ritwik-g, line 336: @chandrasekharan-zipstack could you please explain what exactly are we doing here? May be add a comment in the code also?

backend/api/serializers.py
```diff
@@ -85,15 +85,15 @@ class ExecutionRequestSerializer(Serializer):
 
     Attributes:
         timeout (int): Timeout for the API deployment, maximum value can be 300s.
-            If -1 it corresponds to async execution. Defaults to -1
+            If 0 it corresponds to async execution. Defaults to 0
```
ritwik-g, line 88: @chandrasekharan-zipstack should the default be 0 or the maximum value?

chandrasekharan-zipstack: @ritwik-g in the previous implementation it defaulted to async execution - I let it remain the same. If we're changing this behaviour we'd have to inform users this as a breaking change

ritwik-g: @chandrasekharan-zipstack if that is the case then isn't removing `-1` also a breaking change? Should we then keep the min as `-1` and default as `-1`

chandrasekharan-zipstack: @ritwik-g I wonder why I didn't catch that - let me revert the change if we're worried about this breaking change. We can make it later if necessary

ritwik-g (approved):

muhammad-ali-e (approved):
LGTM. Just to confirm, we're only saving the result in the database when the API is called with specific parameters, right?

chandrasekharan-zipstack:
> LGTM. Just to confirm, we're only saving the result in the database when the API is called with specific parameters, right?

@muhammad-ali-e in case of workflow execution, we'll never save the results from API related source / destination. Only for ETL / task (file/ DB related source / destination) will make use of the file history table to mark processed files (not results)

PR #753, opened 2024-09-29
[Fix] Table output component for record type

jagadeeswaran-zipstack:
## What

- Fix to show table output component when enforce type is record

## Why

- 

## How

-

- No, adds changes only to table extractor.

## Relevant Docs

-

## Related Issues or PRs

-

-

## Notes on Testing

...

## Screenshots

...

## Checklist

### DIFF
--- a/frontend/src/components/custom-tools/prompt-card/PromptOutputsModal.jsx
+++ b/frontend/src/components/custom-tools/prompt-card/PromptOutputsModal.jsx
@@ -1,7 +1,8 @@
 import { Col, Image, Modal, Row, Typography } from "antd";
 import PropTypes from "prop-types";
+
 import { DisplayPromptResult } from "./DisplayPromptResult";
-import { TABLE_ENFORCE_TYPE } from "./constants";
+import { TABLE_ENFORCE_TYPE, RECORD_ENFORCE_TYPE } from "./constants";
 import SpaceWrapper from "../../widgets/space-wrapper/SpaceWrapper";
 import { useCustomToolStore } from "../../../store/custom-tool-store";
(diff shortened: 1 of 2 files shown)


vishnuszipstack (approved):

frontend/src/components/custom-tools/prompt-card/PromptOutputsModal.jsx
```diff
@@ -78,7 +79,9 @@ function PromptOutputsModal({
                     )}
                   </div>
                   <div className="flex-1 overflow-y-auto pad-top-10">
-                    {enforceType === TABLE_ENFORCE_TYPE && TableOutput ? (
+                    {(enforceType === TABLE_ENFORCE_TYPE ||
+                      enforceType === RECORD_ENFORCE_TYPE) &&
```
tahierhussain, line 83: @jagadeeswaran-zipstack NIT: Can we create a variable called `isTableExtraction` for the above condition here? Similar to the one in `PromptOutput.jsx` component. A better approach here would be to pass the `isTableExtraction` prop from `PromptOutput` to `PromptOutputsModal`. Please check if this is possible.

tahierhussain (approved):
LGTM. Requested a NIT change.

frontend/src/components/custom-tools/prompt-card/PromptOutputsModal.jsx
jagadeeswaran-zipstack: I Don't think its needed here. Because We're using this in only one place. So it will be overkill here. Also passing down `isTableExtraction` as prop is not a better way to do this because in future we might need other enforce type other than table and record.

PR #764, opened 2024-10-02
feat: Added script to update lock files

chandrasekharan-zipstack:
## What

- Added script to update lock files
- Updated `pdm-lock-automation.yaml` workflow to execute above script
- Minor `README.md` on the script

## Why

- Frequent SDK updates require a lot of manual effort to update these lock files - script aims to help if run locally. 
- Its also integrated with CI to perform lock file updates on PRs

## How

- Separate script runs on all dirs defined in it, can also run with a list of dirs to update
- Updates lockfile serially - this can be improved later
- Committing changes to the branch handled by [git-auto-commit-action](https://github.com/stefanzweifel/git-auto-commit-action)
- CI triggered on PR, can also be manually dispatched

- No, only additions to something that was disabled

## Relevant Docs

- https://github.com/stefanzweifel/git-auto-commit-action

## Notes on Testing

- Tested the script locally and attempted running the workflow with some minor changes with [act](https://nektosact.com/introduction.html) - though this didn't entirely turn out successful due to few issues in the setup

## Screenshots
1. Run without args
image
1. Run with args
image
1. Minor issues on run with `act` - plan to test it with the runner directly
image

## Checklist


scripts/pdm-lock.sh
```diff
@@ -0,0 +1,69 @@
...
+    file_path="$dir/pyproject.toml"
+
+    # Check if the pyproject.toml file exists
+    if [[ ! -f "$file_path" ]]; then
+        echo "No pyproject.toml found in $dir"
+        continue  # Skip to the next iteration if the file does not exist
+    fi
+
+    # Check if there are changes in pyproject.toml against the main branch
+    git fetch origin main
```
ritwik-g, line 34: @chandrasekharan-zipstack I think we can do this outside loop. We need to do this only one time

scripts/pdm-lock.sh
```diff
@@ -0,0 +1,69 @@
...
+        echo "No changes in dependencies, no need to run pdm lock."
+    else
+        echo "Changes detected, running pdm lock..."
+        pdm lock -G :all -v
+    fi
+
+    # Go back to root if you moved to a subdirectory
+    if [[ "$dir" != "." ]]; then
+        cd -
+    fi
```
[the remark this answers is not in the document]
ritwik-g: If the sequential time delay is not a concern we can go with current implementation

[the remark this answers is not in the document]
chandrasekharan-zipstack: I was considering the same @ritwik-g - will make this change and update the PR. It'll be a good enhancement

.github/workflows/pdm-lock-automation.yaml
```diff
@@ -5,127 +5,68 @@ on:
...
+            ./scripts/pdm-lock.sh "${dir_array[@]}"
           fi
+        shell: bash
+
+      - name: Commit pdm.lock changes
+        uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: Commit pdm.lock changes
+          commit_user_name: pdm-lock-automation[bot]
+          commit_user_email: pdm-lock-automation-bot@unstract.com
```
kirtimanmishrazipstack, line 72: The problem with the earlier workflow was it was generating stale pdm lock.
The reason is when multiple jobs run, the commit was not handling the pdm lock conflict, resulting in stale pdm.
Will this resolve this?

chandrasekharan-zipstack: @Kirtiman I was under the impression that the issue was with how we commit the changes due to conflicts from the stash and upstream. If that's the case - this will ensure that we commit only once after all the changes. Unsure whether it'll cause issues when multiple workflows run from different branches - based on the frequency of such issues we can tackle it I believe

Are you talking about python dependency conflicts here? If that's the case this action itself might not be helpful and we'll have to intervene. Let me try to check if we still commit in case of such conflicts

kirtimanmishrazipstack: @chandrasekharan-zipstack since we are committing only once, it should solve the issue. Resolving and approving.

kirtimanmishrazipstack (approved):
Looks good. Since we are comitting once using stefanzweifel/git-auto-commit-action

scripts/pdm-lock.sh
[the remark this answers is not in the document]
chandrasekharan-zipstack: Updated the PR @ritwik-g

ritwik-g (approved):

PR #794, opened 2024-10-15
[FEAT] Added vline hline in llmW adapter FE

jagadeeswaran-zipstack:
## What

- Added condition hide for `mark_horizontal_lines` and `mark_vertical_lines` field in LLM Whisperer adapter.

## Why

- Need to have this field in llm-whisperer

## How

-

-

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Screenshots
image
image

## Checklist


frontend/src/layouts/rjsf-form-layout/RjsfFormLayout.jsx
```diff
@@ -53,6 +53,9 @@ function RjsfFormLayout({
 
   const uiSchema = {
     "ui:classNames": "my-rjsf-form",
+    mark_horizontal_lines: {
+      "ui:widget": !formData?.mark_vertical_lines ? "hidden" : undefined,
+    },
```
tahierhussain, line 58: @jagadeeswaran-zipstack Instead of hardcoding the `mark_horizontal_lines` field, can we automate this? For example, we could check for any fields that are hidden and automatically add them to the `uiSchema`. This way, it becomes a one-time effort, and we won't have to update the `uiSchema` again if we decide to hide additional fields in the future.

tahierhussain (changes requested):

chandrasekharan-zipstack (commented):
@jagadeeswaran-zipstack why do we need this change? Can you add screenshots of what we're trying to do?

frontend/src/layouts/rjsf-form-layout/RjsfFormLayout.jsx
jagadeeswaran-zipstack: Yes, It is possible. We can pass uiSchema from backend but right now SDK is not written in a way to support uiSchema being passed down. It will function as it does today on the FE side.

jagadeeswaran-zipstack:
> @jagadeeswaran-zipstack why do we need this change? Can you add screenshots of what we're trying to do?

@chandrasekharan-zipstack This change is to hide mark horizontal line when mark vertical line is unchecked.

tahierhussain (approved):
LGTM

Deepak-Kesavan (approved):
Looks good. We need to pass the UI schema too from the backend it might be needed later for

chandrasekharan-zipstack:
> > @jagadeeswaran-zipstack why do we need this change? Can you add screenshots of what we're trying to do?
> 
> @chandrasekharan-zipstack This change is to hide mark horizontal line when mark vertical line is unchecked.

@johnyrahul had achieved [something similar](https://github.com/Zipstack/unstract-sdk/blob/2a0d17bb6ddf07139d40cf32ee9fbf55c3bc8cde/src/unstract/sdk/adapters/x2text/llm_whisperer/src/static/json_schema.json#L79-L123) with one of the LLM whisperer options using JSON schema. Please see if that can be useful here rather writing some specific in the FE to support it

jagadeeswaran-zipstack:
> > > @jagadeeswaran-zipstack why do we need this change? Can you add screenshots of what we're trying to do?
> > 
> > 
> > @chandrasekharan-zipstack This change is to hide mark horizontal line when mark vertical line is unchecked.
> 
> @johnyrahul had achieved [something similar](https://github.com/Zipstack/unstract-sdk/blob/2a0d17bb6ddf07139d40cf32ee9fbf55c3bc8cde/src/unstract/sdk/adapters/x2text/llm_whisperer/src/static/json_schema.json#L79-L123) with one of the LLM whisperer options using JSON schema. Please see if that can be useful here rather writing some specific in the FE to support it

@chandrasekharan-zipstack This can be achieved only when on enum (dropdown menus) other than that to achieve this uiScheme is the only way.

PR #800, opened 2024-10-17
[Fix] Dev portal On-boarding issue

jagadeeswaran-zipstack:
## What

- Showing On-board page when we switch platform even user already added adapters
- Made profile option in profile dropdown as unstract only.

## Why

- Adapter API was not initiated because of the isUnstract check in useSessionValid.js

## How

-

- No. This PR fix onboarding redirect with no changes to core functionality. So, the chances of it breaking any existing features is very minimal.

## Screenshots
image

## Checklist

### DIFF
--- a/frontend/src/components/helpers/auth/PersistentLogin.js
+++ b/frontend/src/components/helpers/auth/PersistentLogin.js
@@ -20,15 +20,9 @@ function PersistentLogin() {
   const { sessionDetails } = useSessionStore();
   const checkSessionValidity = useSessionValid();
 
-  try {
-    if (selectedProductStore?.useSelectedProductStore) {
-      selectedProduct = selectedProductStore?.useSelectedProductStore(
-        (state) => state?.selectedProduct
-      );
-    }
-  } catch (error) {
-    // Do nothing
-  }
+  const selectedProduct = useSelectedProductStore
+    ? useSelectedProductStore((state) => state?.selectedProduct)
+    : null;
   useEffect(() => {
     let isMounted = true;
     setIsLoading(true);
--- a/frontend/src/components/helpers/auth/PersistentLogin.js
+++ b/frontend/src/components/helpers/auth/PersistentLogin.js
@@ -7,10 +7,10 @@ import { SocketMessages } from "../socket-messages/SocketMessages";
 import { GenericLoader } from "../../generic-loader/GenericLoader";
 import { PromptRun } from "../../custom-tools/prompt-card/PromptRun";
 
-let selectedProductStore;
-let selectedProduct;
+let useSelectedProductStore;
 try {
-  selectedProductStore = require("../../../plugins/llm-whisperer/store/select-product-store.js");
+  useSelectedProductStore = require("../../../plugins/llm-whisperer/store/select-product-store.js")
+    .useSelectedProductStore;
 } catch {
   // Ignore if hook not available
 }
--- a/frontend/src/components/helpers/auth/PersistentLogin.js
+++ b/frontend/src/components/helpers/auth/PersistentLogin.js
@@ -9,8 +9,8 @@ import { PromptRun } from "../../custom-tools/prompt-card/PromptRun";
 
 let useSelectedProductStore;
 try {
-  useSelectedProductStore = require("../../../plugins/llm-whisperer/store/select-product-store.js")
-    .useSelectedProductStore;
+  useSelectedProductStore =
+    require("../../../plugins/llm-whisperer/store/select-product-store.js").useSelectedProductStore;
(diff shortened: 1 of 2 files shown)


vishnuszipstack (approved):

frontend/src/components/helpers/auth/PersistentLogin.js
```diff
@@ -7,14 +7,31 @@ import { SocketMessages } from "../socket-messages/SocketMessages";
 import { GenericLoader } from "../../generic-loader/GenericLoader";
 import { PromptRun } from "../../custom-tools/prompt-card/PromptRun";
 
+let selectedProductStore;
+let selectedProduct;
+try {
+  selectedProductStore = require("../../../plugins/llm-whisperer/store/select-product-store.js");
+} catch {
+  // Ignore if hook not available
+}
```
tahierhussain, line 16: @jagadeeswaran-zipstack Here's a straightforward approach to read the `useSelectedProductStore`,

```suggestion
let useSelectedProductStore;
try {
  useSelectedProductStore = require("../../../plugins/llm-whisperer/store/select-product-store.js")
    .useSelectedProductStore;
} catch {
  // Ignore if hook not available
}
```

frontend/src/components/helpers/auth/PersistentLogin.js
```diff
@@ -7,14 +7,31 @@ import { SocketMessages } from "../socket-messages/SocketMessages";
...
 
+  try {
+    if (selectedProductStore?.useSelectedProductStore) {
+      selectedProduct = selectedProductStore?.useSelectedProductStore(
+        (state) => state?.selectedProduct
+      );
+    }
+  } catch (error) {
+    // Do nothing
+  }
```
tahierhussain, line 31: @jagadeeswaran-zipstack Here's the corresponding change to my previous suggestion,

```suggestion
  const selectedProduct = useSelectedProductStore
    ? useSelectedProductStore((state) => state?.selectedProduct)
    : null;
```

frontend/src/components/helpers/auth/PersistentLogin.js
```diff
@@ -7,14 +7,31 @@ import { SocketMessages } from "../socket-messages/SocketMessages";
...
   useEffect(() => {
     let isMounted = true;
-
+    setIsLoading(true);
```
tahierhussain, line 34: @jagadeeswaran-zipstack Why are we setting the `isLoading` state to `true` here? Can you provide some context?

tahierhussain (changes requested):

frontend/src/components/helpers/auth/PersistentLogin.js
jagadeeswaran-zipstack: When page loads for first time loader will be set to false. since its a wrapper component its state remains same. When we navigate to other platform the loader state will be false which results to unexpected behaviours.

frontend/src/components/helpers/auth/PersistentLogin.js
jagadeeswaran-zipstack: @tahierhussain This was wrote this way because you shouldn't call hooks conditionally as per sonar config.
https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=800&id=Zipstack_unstract&open=AZKZdv-sy9qRwus5zxKC

frontend/src/components/helpers/auth/PersistentLogin.js
tahierhussain: @jagadeeswaran-zipstack Thanks for the explanation! I have a suggestion regarding the `isLoading` state. Instead of setting `isLoading` to `true` every time the `useEffect` is triggered, we should set it to `true` only when we call the `verifySession` function. This will prevent unnecessary state updates and re-renders when the session is already valid. Please review the changes and update the code accordingly. Let me know if any further clarification is needed.

```
useEffect(() => {
  let isMounted = true;

  const verifySession = async () => {
    try {
      await checkSessionValidity();
    } finally {
      isMounted && setIsLoading(false);
    }
  };

  if (!sessionDetails?.isLoggedIn) {
    setIsLoading(true); // Only trigger loading if session is invalid
    verifySession();
  } else {
    setIsLoading(false);
  }

  return () => (isMounted = false);
}, [selectedProduct]);
```

jagadeeswaran-zipstack: @tahierhussain Yes understood. This approach is clean and clear. Thanks

tahierhussain (approved):
LGTM

PR #807, opened 2024-10-18
[DON'T MERGE] FIX: Commented out React Strict Mode and added developer instructions for re-enabling it

tahierhussain:
## What

1. Commented out the React Strict Mode in the application’s entry point to prevent issues caused by its additional checks during local development.
2. Added comments with instructions for developers on how to re-enable React Strict Mode if they wish to use it for additional development checks.

## Why

1. React Strict Mode performs extra checks in development, such as double-invoking lifecycle methods like `useEffect`. While these checks are useful for identifying potential issues, they can cause confusion and unexpected behavior during local development that won’t occur in production.
2. Disabling Strict Mode can make the development process smoother by avoiding these checks. However, developers should still have the option to enable it when needed for debugging or identifying issues.

## How

1. Both the `<StrictMode>` component in the render method and the `StrictMode` import statement have been commented out.
2. Clear instructions are provided as comments in the code, informing developers how to re-enable Strict Mode by uncommenting both the import and the component.

No. Only the React Strict Mode has been disabled in this PR, and this change should not affect any existing features.

## Database Migrations

NA

## Env Config

NA

## Relevant Docs

NA

## Related Issues or PRs

NA

## Dependencies Versions

NA

## Notes on Testing

NA

## Screenshots
NA

## Checklist

### DIFF
--- a/frontend/src/index.js
+++ b/frontend/src/index.js
@@ -1,4 +1,4 @@
-import React from "react";
+import { StrictMode } from "react"; // eslint-disable-line no-unused-vars
 import ReactDOM from "react-dom/client";
 import posthog from "posthog-js";
 import { PostHogProvider } from "posthog-js/react";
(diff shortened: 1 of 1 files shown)


vishnuszipstack (approved):

jagadeeswaran-zipstack (approved):

hari-kuriakose:
@tahierhussain As we discussed regarding the adoption of Strict Mode in React community and based on our use cases, I would suggest NOT to disable React Strict Mode by default due to the following reasons:

- Even though React Strict Mode has dual feedback from across the community, let us not go by that and instead try to see how it currently applies to our use cases.
- React Strict Mode just mounts a component twice functionally, which I think should not be be problem for our React components which basically do just CRUD.
- Different use cases that we discussed are all actually Strict Mode compatible and if not currently can be made so, thus helping us to maintain a healthy and updated ecosystem of React components.
- Strict Mode will be disabled by React automatically in Production builds thus continuing to align with existing CI/CD workflows, without any extra effort.
- Keeping Strict Mode enabled will also force our frontend contributors to adhere to these objectives for the foreseeable future.

That being said, let us do re-evaluate later as our use cases grow and it becomes more evident whether Strict Mode is still maintainable for us or not.

PR #811, opened 2024-10-22
Platform service: Support remote storage

gaya3-zipstack:
## What

Support remote file storage for cost calculation

## Why

To enable the service use remote storage services

## How

Integrate changes in SDK for the same and get the service to use remote storage

- Cost calcultation

## Database Migrations

-  NA

## Env Config

- FILE_STORAGE_CREDENTIALS needs to have the provider and credentials
- -remote_file_storage is a feature flag that controls the support.

## Relevant Docs

-

## Related Issues or PRs

- https://github.com/Zipstack/unstract-sdk/pull/112

## Dependencies Versions

-

## Notes on Testing

- Test for cost calculation json being loaded on the mentioned remote storage the same being used.

## Checklist


platform-service/sample.env
```diff
@@ -31,6 +31,7 @@ FLIPT_SERVICE_AVAILABLE=False
 # Cost calculation related ENVs
 MODEL_PRICES_URL="https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json"
 MODEL_PRICES_TTL_IN_DAYS=7
-MODEL_PRICES_FILE_PATH="/tmp/model_prices.json"
+MODEL_PRICES_FILE_PATH="fsspec-test/cost/model_prices.json"
+FILE_STORAGE='{"provider": "gcs", "credentials": {"token": "/path/to/google/creds.json"}}'
```
hari-kuriakose, line 35: @gaya3-zipstack Please rename to `FILE_STORAGE_CREDENTIALS`.

platform-service/sample.env
```diff
@@ -31,6 +31,7 @@ FLIPT_SERVICE_AVAILABLE=False
 # Cost calculation related ENVs
 MODEL_PRICES_URL="https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json"
 MODEL_PRICES_TTL_IN_DAYS=7
-MODEL_PRICES_FILE_PATH="/tmp/model_prices.json"
+MODEL_PRICES_FILE_PATH="fsspec-test/cost/model_prices.json"
```
hari-kuriakose, line 34: @gaya3-zipstack Sample value can just be something without references to fsspec to indicate that it's a path to `model_prices.json`.

gaya3-zipstack: Done

platform-service/sample.env
gaya3-zipstack: Done

platform-service/src/unstract/platform_service/helper/cost_calculation.py
```diff
@@ -1,16 +1,19 @@
...
+from unstract.platform_service.constants import FeatureFlag
 from unstract.platform_service.env import Env
 from unstract.platform_service.utils import format_float_positional
+from unstract.sdk.exceptions import FileStorageError
+from unstract.sdk.file_storage import FileStorageProvider, PermanentFileStorage
+
+from unstract.flags.feature_flag import check_feature_flag_status
 
 
-@singleton
```
gaya3-zipstack, line 13: Removing singleton here. With SIngleton, the TTL check will only happen when container restarts and the 7 days TTL config may not be adhered to.
Anyway, the plan is to create a separate cron job for the json file pull and updation. Once that is in place, the TTL/file expiry  logic can be safely removed.
@chandrasekharan-zipstack

platform-service/pyproject.toml
```diff
@@ -15,6 +15,8 @@ dependencies = [
     "unstract-flags @ file:///${PROJECT_ROOT}/../unstract/flags",
     "requests>=2.31.0",
     "singleton-decorator~=1.0.0",
+    "unstract-sdk~=0.54.1",
```
chandrasekharan-zipstack, line 18: NIT: Consider specifying such unpublished changes as [VCS dependencies](https://pdm-project.org/en/latest/usage/dependency/#vcs-dependencies) if you wish to test easily

platform-service/src/unstract/platform_service/helper/cost_calculation.py
```diff
@@ -49,24 +76,47 @@ def calculate_cost(
...
+                    app.logger.info(f"Reading model token data from {self.file_path}")
+                    # File exists and TTL has not expired, read and return content
+                    with open(self.file_path, encoding="utf-8") as f:
+                        return json.load(f)
+                else:
+                    # TTL expired, fetch updated JSON data from API
+                    return self._fetch_and_save_json()
         except Exception as e:
-            app.logger.error("Error in calculate_cost: %s", e)
+            app.logger.error(
```
chandrasekharan-zipstack, line 117: @gaya3-zipstack at places where we catch, log and suppress the exception can we either
1. Log it as a warning and not an error OR
1. Explicitly specify in the log message that we'll be continuing the processing / execution

This will help us while reading the service's logs and won't throw us off while encountering such errors

platform-service/src/unstract/platform_service/helper/cost_calculation.py
```diff
@@ -21,16 +24,40 @@ def __init__(
         self.ttl_days = ttl_days
         self.url = url
         self.file_path = file_path
+
+        if check_feature_flag_status(FeatureFlag.REMOTE_FILE_STORAGE):
+            self.file_storage, self.file_path = self.__get_storage_crdentials()
         self.model_token_data = self._get_model_token_data()
 
+    def __get_storage_crdentials(self) -> tuple[PermanentFileStorage, str]:
```
chandrasekharan-zipstack, line 32:
```suggestion
    def __get_storage_credentials(self) -> tuple[PermanentFileStorage, str]:
```
NIT: Typo fixed

platform-service/src/unstract/platform_service/helper/cost_calculation.py
```diff
@@ -49,24 +76,47 @@ def calculate_cost(
 
     def _get_model_token_data(self) -> Optional[dict[str, Any]]:
         try:
-            # File does not exist, fetch JSON data from API
-            if not os.path.exists(self.file_path):
-                return self._fetch_and_save_json()
+            if check_feature_flag_status(FeatureFlag.REMOTE_FILE_STORAGE):
```
chandrasekharan-zipstack, line 79: @gaya3-zipstack to prevent this function from growing in size and to maintain readibility please split as 2 functions

platform-service/src/unstract/platform_service/helper/cost_calculation.py
```diff
@@ -83,18 +133,31 @@ def _fetch_and_save_json(self) -> Optional[dict[str, Any]]:
...
+                    ensure_ascii=False,
+                    indent=4,
+                )
+            else:
+                with open(self.file_path, "w", encoding="utf-8") as f:
+                    json.dump(json_data, f, ensure_ascii=False, indent=4)
+                # Set the file's modification time to indicate TTL
+                expiry_date = datetime.now() + timedelta(days=self.ttl_days)
+                expiry_timestamp = expiry_date.timestamp()
+                os.utime(self.file_path, (expiry_timestamp, expiry_timestamp))
```
chandrasekharan-zipstack, line 151: @gaya3-zipstack are we intentionally not updating the modification time in case of remote storage?

platform-service/src/unstract/platform_service/helper/cost_calculation.py
chandrasekharan-zipstack: @gaya3-zipstack will the cron job to pull and update be run within the container? If not, we could just subscribe to changes in that JSON and try to make use of some action to update our code / push the file

platform-service/pyproject.toml
gaya3-zipstack: hmmmm.... Thank you fo pointing to this. I am just wondering if this is already supported where we can pick up a version from a particular git repo's branch (hope that is right), then why do we need to publish anything to local pypi / editable version installs etc for testing? Guess my understanding is not fully right @chandrasekharan-zipstack ?

platform-service/src/unstract/platform_service/helper/cost_calculation.py
gaya3-zipstack: Done. CHanged pre-existing code.

platform-service/src/unstract/platform_service/helper/cost_calculation.py
gaya3-zipstack: Yes. Because whenever we write to a file, the modification time will reflect the last time that the file is written to. There is no need to explicitly update the time in this particular use-case.

platform-service/src/unstract/platform_service/helper/cost_calculation.py
gaya3-zipstack: @chandrasekharan-zipstack Also removing the singleton-decorator dependency. Looks like this was the only place where we use singleton.

chandrasekharan-zipstack: I hope this is being tracked somewhere, please help create a ticket if not @gaya3-zipstack and feel free to resolve this conversation
cc: @hari-kuriakose

platform-service/src/unstract/platform_service/helper/cost_calculation.py
chandrasekharan-zipstack: @gaya3-zipstack can you fix this typo as well?

platform-service/src/unstract/platform_service/helper/cost_calculation.py
chandrasekharan-zipstack: @gaya3-zipstack I think its okay to not add a new function if its just a few lines of changes, however a large number of lines added should be in a separate function. We don't have a fixed coding standard around the number of lines allowed in a function yet which causes this ambiguity. My only worry is to not end up in a (nightmare-ish) scenario similar to prompt-service's `main.py` :sweat_smile: 
Will let you guys take a call on this wrt these temporary changes

platform-service/pyproject.toml
chandrasekharan-zipstack: @gaya3-zipstack yes, that was my point. We needn't publish to pypi for the sake of testing. Its okay to have release candidates though, no harm. Just wanted to share this option with you for your own testing / for co-ordinating better with others

Deepak-Kesavan (approved):

platform-service/src/unstract/platform_service/helper/cost_calculation.py
gaya3-zipstack: Done

PR #791, opened 2024-10-12
Script to run v2 migration commands

chandrasekharan-zipstack:
## What

- Script to run some commands and help with the migration
- Some doc added for migration to v2
- Show warning on running `./run-platform.sh`
- Minor fixes to run v2 on OSS
- Added script `print_warnings.py` to print the release warnings
- MINOR: Renamed `branch` to `target_branch` in `run-platform.sh` script

## Why

- Helps OSS users easily run the migration
- Change in login template - this was causing an issue in resolving the login API for OSS version
`NoReverseMatch("Reverse for 'login' not found. 'login' is not a valid view function or pattern name.")`

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

- No, docs change and a new script added

## Env Config

- `SCHEMAS_TO_MIGRATE=_ALL_` added for the migration script

## Notes on Testing

- Followed initial few steps of creating a new DB, taking a backup and restore in the new DB
- Ran the script and let it perform all the steps

## Screenshots
1. Scipt run
image
1. Default warning when versions can't be deduced
image
1. Actual `print_warnings.py` script output if versions are correctly passed
image
1. Release notes during update
image

## Checklist

I have read and understood the [Contribution Guidelines]().

PR #812, opened 2024-10-22
feat: Moved scripts dir in root to docker dir

chandrasekharan-zipstack:
## What

- Moved scripts from root to existing `scripts` dir in `/docker`
- Necessary changes to ensure no usage of the script breaks

## Why

- Clubbing all scripts within the `/docker` dir

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

- No, only moving the existing scripts with straightforward changes in its usage

## Notes on Testing

- No explicit testing done, since its a minor refactor

## Checklist

I have read and understood the [Contribution Guidelines]().

PR #820, opened 2024-10-25
fix: Updated run-platform to create schema for new users

chandrasekharan-zipstack:
## What

- Fix in `run-platform.sh` to create DB schema 
- Fix in `migrate-to-v2.sh` to allow talking to DB / redis in containers
- Updated `ENTRYPOINT` to `CMD` in backend's Dockerfile, to allow invoking the necessary schema creation command
- Renamed and moved the schema creation / drop commands to `utils` app

## Why

- Schema creation has to be done explicitly once as part of the v2 changes, previously this was done by django-tenant and the `public` schema would be present by default
- As a result new users faced [this issue](https://unstract.slack.com/archives/C06RL7J6JDS/p1729847165184289)

## How

- For first time users, schema creation handled with command run on backend container
- For existing users, they will have to read the release notes and manually run `backend/migrating/v2/migrate-to-v2.sh` to perform the schema creation and subsequent migration

- Yes, since it touches the backend Dockerfile and some scripts used by OSS users - it is possible

## Database Migrations

-  Might run when these scripts are invoked

## Related Issues or PRs

- #791 
- [Issue from  support channel](https://unstract.slack.com/archives/C06RL7J6JDS/p1729847165184289)

## Notes on Testing

- Removed `/etc/hosts` entries to mimic the average OSS user to reproduce the same issues and worked on unblocking myself

## Screenshots Old
- To be used by new users, ran with trace option
image

- Migration script (to be used by existing users)
image

## Screenshots after addressing comments
- Schema creation handled through entrypoint script
image

## Checklist


run-platform.sh
```diff
@@ -268,9 +268,26 @@ build_services() {
...
 }
 
+prepare_for_first_setup() {
+  pushd ${script_dir}/docker 1>/dev/null
+
+  echo -e "$blue_text""Preparing for initial setup""$default_text"
+  VERSION=$opt_version $docker_compose_cmd up -d db redis
+
+  echo -e "Creating a schema for Unstract in the database"
+  VERSION=$opt_version $docker_compose_cmd run backend .venv/bin/python manage.py create_schema
```
jaseemjaskp, line 278: @chandrasekharan-zipstack schema creation should be part of init script (because it creates schema only if does't exists). so we can  avoid touching the backend Dockerfile and some scripts used by OSS users. It is more elegant way to handle this issue.

docker/dockerfiles/backend.Dockerfile
```diff
@@ -56,4 +56,4 @@ RUN set -e; \
 
 EXPOSE 8000
 
-ENTRYPOINT [ "./entrypoint.sh" ]
+CMD [ "./entrypoint.sh" ]
```
hari-kuriakose, line 59: @chandrasekharan-zipstack This should be `ENTRYPOINT` itself so that we can override the same via `CMD` in docker compose files, etc.

docker/docker-compose.yaml
```diff
@@ -7,6 +7,7 @@ services:
     image: unstract/backend:${VERSION}
     container_name: unstract-backend
     restart: unless-stopped
+    entrypoint: ./entrypoint.sh
```
hari-kuriakose, line 10: @chandrasekharan-zipstack Entrypoint is set in the base image itself, hence we don't need to set it explicitly here.
Please revert.

docker/docker-compose.yaml
```diff
@@ -24,6 +25,7 @@ services:
       - prompt_studio_data:/app/prompt-studio-data
       - ./workflow_data:/data
       - ${TOOL_REGISTRY_CONFIG_SRC_PATH}:/data/tool_registry_config
+      - ../backend/entrypoint.sh:/app/entrypoint.sh
```
hari-kuriakose, line 28: @chandrasekharan-zipstack Entrypoint is copied to base image layer itself, hence we don't need another volume mount here.
Please revert.

hari-kuriakose:
@chandrasekharan-zipstack Let's also move `scripts/` to `docker/scripts/` along with this.

run-platform.sh
```diff
@@ -268,9 +268,26 @@ build_services() {
   fi
 }
 
+prepare_for_first_setup() {
```
hari-kuriakose, line 271: @chandrasekharan-zipstack Please rename to `create_backend_schema()`.

run-platform.sh
```diff
@@ -268,9 +268,26 @@ build_services() {
...
+  popd 1>/dev/null
+}
+
 run_services() {
   pushd ${script_dir}/docker 1>/dev/null
 
+  if [ "$first_setup" = true ]; then
+    prepare_for_first_setup
+  fi
+
```
hari-kuriakose, line 290: @chandrasekharan-zipstack Let's move out backend schema creation from here.

Instead, call `create_backend_schema()` directly after `build_services` and before `run_services` below.

Also inside `create_backend_schema()` return early if `first_setup` is `false`.

run-platform.sh
hari-kuriakose: @chandrasekharan-zipstack Agree here.

Extend entrypoint script itself with necessary Django commands so that we can replace this with `VERSION=$opt_version $docker_compose_cmd run backend prepare_and_migrate` here.

Also we can hopefully remove Redis dependency later, as ideally db migrations do not have anything to do with Redis.

backend/migrating/v2/migrate-to-v2.sh
```diff
@@ -37,11 +34,16 @@ echo -e "${blue_text}Configuring environment variables for the migration${defaul
 cd $manage_script_dir
 activate_venv
 
+# TODO: Update to support running from container instead of setting these envs
+# VERSION=test docker compose -f docker-compose.yaml run backend .venv/bin/python manage.py create_schema
 echo -e "${blue_text}Running schema creation command...${default_text}"
-python manage.py create_v2_schema || { echo "Schema creation failed"; exit 1; }
+DB_HOST="localhost" REDIS_HOST="localhost" CELERY_BROKER_URL="redis://localhost:6379" \
+python manage.py create_schema || { echo "Schema creation failed"; exit 1; }
```
hari-kuriakose, line 41: @chandrasekharan-zipstack Can replace all of these direct invocations with `docker compose run` commands as that is the default for end users?

backend/migrating/v2/migrate-to-v2.sh
```diff
@@ -37,11 +34,16 @@ echo -e "${blue_text}Configuring environment variables for the migration${defaul
...
 
 echo -e "${blue_text}Running schema migration command...${default_text}"
+DB_HOST="localhost" REDIS_HOST="localhost" CELERY_BROKER_URL="redis://localhost:6379" \
```
hari-kuriakose, line 44: @chandrasekharan-zipstack Also ideally only `DB_HOST` is required, as we don't need Redis or Celery for any db migrations.

run-platform.sh
chandrasekharan-zipstack: Yes @hari-kuriakose, will retain this function and let the schema creation be done through the new CLI argument.
The redis dependency comes from the celery tasks that we register in backend's [__init__.py](https://github.com/Zipstack/unstract/blob/0ee58329340b2c97520308140e5031f033485e9e/backend/backend/__init__.py#L1)

chandrasekharan-zipstack:
> @chandrasekharan-zipstack Let's also move `scripts/` to `docker/scripts/` along with this.

@hari-kuriakose I've addressed it in [this PR](https://github.com/Zipstack/unstract/pull/812) - can we let these be separate changes? Will resolve any conflicts accordingly

backend/utils/management/commands/__init__.py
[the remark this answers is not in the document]
chandrasekharan-zipstack: @hari-kuriakose I don't mind that, however django [suggests](https://docs.djangoproject.com/en/4.2/howto/custom-management-commands/) placing the commands in its own associated app. Shall I still consider moving the commands to a separate `commands` app?

backend/migrating/v2/migrate-to-v2.sh
chandrasekharan-zipstack: @hari-kuriakose I considered that too - it should be ideal. However I was challenged with ways to determine the version that had to be used and realized that we'd duplicate a lot of code from `run-platform.sh`, hope that's okay

backend/migrating/v2/migrate-to-v2.sh
chandrasekharan-zipstack: @hari-kuriakose the need for Redis comes from how we register the celery tasks - didn't want to mess / check it in the scope of this PR so ended up including it. More in [this comment](https://github.com/Zipstack/unstract/pull/820#discussion_r1817771020).

If we adopt `docker compose run` commands - this would change anyway

backend/utils/management/commands/__init__.py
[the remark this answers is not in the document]
chandrasekharan-zipstack: Moved to a separate `commands` app as requested - we can revisit this later if required

jaseemjaskp (approved):
LGTM

run-platform.sh
hari-kuriakose: @chandrasekharan-zipstack Ah yes, we will need to do this later.

backend/migrating/v2/migrate-to-v2.sh
hari-kuriakose: @chandrasekharan-zipstack As observed, direct invocation like this is easier than docker compose run sanity checks and usage requirements.

hari-kuriakose:
Can take care of the Django management commands security hotspots separately.

PR #821, opened 2024-10-29
[FIX] Changes to display chunk data properly

Deepak-Kesavan:
## What

- Display chunk data properly

## Why

- Earlier there were issues in how chunks were split.

## How

- Changed the chunk data as array which was string earlier.

- No

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


backend/prompt_studio/prompt_studio_output_manager_v2/serializers.py
```diff
@@ -25,4 +26,11 @@ def to_representation(self, instance):
             )
             token_usage = {}
         data["token_usage"] = token_usage
+        # Convert string to list
+        try:
+            context = json.loads(data["context"])
```
chandrasekharan-zipstack, line 31: @Deepak-Kesavan are we doing this to not change the model / serializer and run a migration?

prompt-service/src/unstract/prompt_service/main.py
```diff
@@ -323,7 +324,7 @@ def prompt_processor() -> Any:
                     tool_settings=tool_settings,
                     output=output,
                     llm=llm,
-                    context=context,
+                    context="\n".join(context),
```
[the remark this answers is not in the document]
Deepak-Kesavan: @chandrasekharan-zipstack That doesn't always work. What if user provides the same separator  as page separator in LLMWhisperer adapter?

backend/prompt_studio/prompt_studio_output_manager_v2/serializers.py
Deepak-Kesavan: @chandrasekharan-zipstack We always use text to store data in postgres. We need to load the data every time using json.loads() before using. But in this case, we directly fetch from DB and return it to frontend. So instead of parsing it in the frontend, I handled it in the serializer.

harini-venkataraman (approved):

chandrasekharan-zipstack (approved):
LGTM for the most part however one point of concern is why we pass store the context in a set and then pass it around as an `str`. Why don't we attempt to use a `set[str]` everywhere? Did you do this to phase in these changes?

It'll be hard to get back / work with this `/n` separated string for context in cases where its passed to other functions

backend/prompt_studio/prompt_studio_output_manager_v2/serializers.py
chandrasekharan-zipstack: @Deepak-Kesavan if we know that the data will always be JSON we can attempt to alter the field data type in postgres rather than doing an extra serialization / deserialization  while storing in the DB

prompt-service/src/unstract/prompt_service/main.py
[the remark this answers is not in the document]
chandrasekharan-zipstack: Okay, my main concern was on retaining context to be as a `set[str]` and passing around the same everywhere

backend/prompt_studio/prompt_studio_output_manager_v2/serializers.py
Deepak-Kesavan: It will be an unwanted effort to migrate table schema and do data migrarion on existing data. Even after doing that the chunk data will be still wrong if it is not ran again

Deepak-Kesavan:
> LGTM for the most part however one point of concern is why we pass store the context in a set and then pass it around as an `str`. Why don't we attempt to use a `set[str]` everywhere? Did you do this to phase in these changes?
> 
> It'll be hard to get back / work with this `/n` separated string for context in cases where its passed to other functions

@chandrasekharan-zipstack We don't need seperate chunks in those functions. That is how it was written earlier too. Mainly before sending it to LLM, it is joined with '\n'.

hari-kuriakose:
Duplication will be addressed when we remove vi Django apps later. We can proceed for now.

PR #834, opened 2024-11-08
Update frontend README

hari-kuriakose:
## What

* Add notice for react strict mode
* Remove default doc for npm eject

## Why

* For reasons mentioned [here](https://github.com/Zipstack/unstract/pull/807#issuecomment-2464426640) we will keep React Strict Mode enabled by default for now, in which case it is better to add a notice otherwise it might confuse unwary frontend contributors
* We are almost never going to do a destructive operation like `npm eject`, hence removed the doc regarding the same.

## How

- Updated frontend README
- Updated public docs [here](https://docs.unstract.com/unstract/contributing/unstract/contribute-frontend/#react-strict-mode)

## Checklist

### DIFF
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -54,15 +54,11 @@ Your app is ready to be deployed!
 
 See the section about [deployment](https://facebook.github.io/create-react-app/docs/deployment) for more information.
 
-### `npm run eject`
+## React Strict Mode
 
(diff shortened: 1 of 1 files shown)


tahierhussain (approved):
LGTM

jaseemjaskp (approved):
LGTM

PR #845, opened 2024-11-15
feat: Enhanced tool-structure logs, show file name in tool-structure metadata

chandrasekharan-zipstack:
## What

- Improved structure tool logs in workflow
- Removed some verbose / redundant logs
- Bumped structure tool version to `0.0.48`
- Show actual file name in the result's metadata instead of `INFILE`
- Minor refactoring in the tool code

## Why

- Structure tool's logs was previously dumping the entire `tool_metadata` which was not readable and usually ignored
- This PR aims to improve readability and provide useful logs

## How

- Added a util function to convert json to markdown

- No, predominantly log related changes

## Notes on Testing

- Tried running the tool standalone and ran it with the platform in a workflow

## Screenshots
image
image

## Checklist

### DIFF
--- a/tools/structure/src/config/properties.json
+++ b/tools/structure/src/config/properties.json
@@ -2,7 +2,7 @@
   "schemaVersion": "0.0.1",
   "displayName": "Structure Tool",
   "functionName": "structure_tool",
-  "toolVersion": "0.0.46",
+  "toolVersion": "0.0.48",
   "description": "This is a template tool which can answer set of input prompts designed in the Prompt Studio",
   "input": {
     "description": "File that needs to be indexed and parsed for answers"
--- a/tools/structure/src/config/properties.json
+++ b/tools/structure/src/config/properties.json
@@ -2,7 +2,7 @@
   "schemaVersion": "0.0.1",
   "displayName": "Structure Tool",
   "functionName": "structure_tool",
-  "toolVersion": "0.0.48",
+  "toolVersion": "0.0.49",
   "description": "This is a template tool which can answer set of input prompts designed in the Prompt Studio",
   "input": {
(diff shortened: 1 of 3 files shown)


tools/structure/src/config/properties.json
```diff
@@ -2,7 +2,7 @@
   "schemaVersion": "0.0.1",
   "displayName": "Structure Tool",
   "functionName": "structure_tool",
-  "toolVersion": "0.0.46",
+  "toolVersion": "0.0.48",
```
chandrasekharan-zipstack, line 5: `0.0.47` was already built but we didn't update it here

tools/structure/src/utils.py
```diff
@@ -0,0 +1,31 @@
+from typing import Any
+
+
+def json_to_markdown(data: Any, level: int = 0, parent_key: str = "") -> str:
```
gaya3-zipstack, line 4: This looks like a common piece of code that can be re-used? Might want to consider moving this for later. Also writing such logical pieces - Has this been practically re-usable for us? Or have we been spending quite some effort in doing such things? We may have to re-consider our strategy if we think we are spending a lot of tie in writing such parsing logic.

chandrasekharan-zipstack: @gaya3-zipstack I agree in moving such common code logic to a relevant lib. Will move this to the SDK later and remove this from here when we encounter a scenario where we wish to use this.

From what I've seen, the common function itself gets duplicated across components sometimes and we end up refactoring it later.
This particular function can be useful to render any JSON for any other tool too. Currently I don't see a usecase for it yet.

Also it was easy to obtain such a generic parsing logic from an LLM :)

PR #855, opened 2024-11-20
fix: Make sure that images can be built on arm64

jaysonsantos:
## What
Allow images to be built on arm64

## Why
If emulation is used, it is slow and a few errors happen making the backend not runnable

I have read and understood the [Contribution Guidelines](x).


jaseemjaskp:
@jaysonsantos We provide support for ARM64 builds for users who wish to explicitly build the Docker image in their local environment. Therefore, I am closing this PR.

PR #883, opened 2024-12-11
fix: WF execution created before WF async call

chandrasekharan-zipstack:
## What

- Created WF execution instance before the async call is made

## TODO

- Since a lot of queries are made to the DB when we get this WF execution object by ID - we could directly pass around the WF execution object maybe

## Why

- To avoid errors indicating that WF execution object is not present

- No, creating WF instance with default values only

## Notes on Testing

- Tried making API deployment calls locally, didn't explicitly check the failing scenario though since its a minor fix here

## Checklist


backend/api_v2/deployment_helper.py
```diff
@@ -25,7 +24,10 @@
 from workflow_manager.workflow_v2.dto import ExecutionResponse
 from workflow_manager.workflow_v2.enums import ExecutionStatus
 from workflow_manager.workflow_v2.models.workflow import Workflow
-from workflow_manager.workflow_v2.workflow_helper import WorkflowHelper
+from workflow_manager.workflow_v2.workflow_helper import (
+    WorkflowExecutionServiceHelper,
```
muhammad-ali-e, line 28: import `WorkflowExecutionServiceHelper` from `workflow_manager.workflow_v2.execution`

backend/api_v2/deployment_helper.py
```diff
@@ -152,7 +154,12 @@ def execute_workflow(
         """
         workflow_id = api.workflow.id
         pipeline_id = api.id
-        execution_id = str(uuid.uuid4())
+        workflow_execution = WorkflowExecutionServiceHelper.create_workflow_execution(
+            workflow_id=workflow_id,
+            single_step=False,
```
muhammad-ali-e, line 159: it's default False. don't pass

muhammad-ali-e: we should pass mode as mode WorkflowExecution.Mode.QUEUE,

muhammad-ali-e (commented):
Added comments. Please ensure it's tested before merging

backend/api_v2/deployment_helper.py
chandrasekharan-zipstack: @muhammad-ali-e in case of an API deployment call, shouldn't it be executed instantly when the request is made? Changing this anyway

Deepak-Kesavan (approved):

backend/api_v2/deployment_helper.py
muhammad-ali-e: Even it's an instant in the POV of an end user, This might be in celery queue and execute depends on the avaialability of worker. So allways async manner.

muhammad-ali-e (approved):
LGTM

PR #891, opened 2024-12-12
First version of file Execution model

muhammad-ali-e:
## What

UN-1891

- Introduce a FileExecution Model

  - Added File Execution Model to Track File-Level Details in Workflow Executions.
 
- Refactor in source.py

  - Consolidated duplicate logic ( reading file content two times )into a single place .

- Set `runId` as `ID` of FileExecution instaed of generating a new **UUID**.
- Set `use_file_history=True` for **ETL and Task** Execution to avoid unnecessary re-executions.

## Why

- We need a dedicated FileExecution model to manage the tracking of individual file executions within a workflow execution. While the WorkflowExecution model captures the overall execution, this new model will handle details for each file processed as part of the workflow.

## How

- Added as a new application

- Might be not, But need to test all case of executions. because it execute during workflow executions.

## Database Migrations

-  Yes, New model

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


backend/workflow_manager/endpoint_v2/source.py
```diff
@@ -417,21 +433,28 @@ def get_file_content(self, input_file_path: str, chunk_size: int = 8192) -> byte
...
         """
         connector: ConnectorInstance = self.endpoint.connector_instance
         connector_settings: dict[str, Any] = connector.connector_metadata
         source_fs = self.get_fsspec(
             settings=connector_settings, connector_id=connector.connector_id
         )
+
+        # Get file size
+        file_metadata = source_fs.stat(input_file_path)
+        file_size = file_metadata.get("size", 0)
```
chandrasekharan-zipstack, line 448:
```suggestion
        file_size = file_metadata.get("size")
```
@muhammad-ali-e since we store an `Optional[int]` in `FileHash` it might be good to leave it unassigned if we're not able to obtain the size for a file - please change the return type as well accordingly

backend/workflow_manager/file_execution/file_execution_helper.py
```diff
@@ -0,0 +1,89 @@
...
+
+        return:
+            The `WorkflowExecutionInputFile` object
+        """
+        execution_file, is_created = WorkflowFileExecution.objects.get_or_create(
+            workflow_execution=workflow_execution,
+            file_hash=file_hash,
+            file_path=file_path,
+        )
+        if is_created:
```
chandrasekharan-zipstack, line 41: NIT: We could add a debug log on new object creation with relevant params

backend/workflow_manager/file_execution/file_execution_helper.py
```diff
@@ -0,0 +1,89 @@
+from typing import Optional
+
+from workflow_manager.file_execution.models import WorkflowFileExecution
+from workflow_manager.workflow_v2.enums import ExecutionStatus
+from workflow_manager.workflow_v2.models.execution import WorkflowExecution
+
+
+class FileExecutionHelper:
```
chandrasekharan-zipstack, line 8: NIT: @muhammad-ali-e since all these methods are closely tied to the `WorkflowFileExecution` model - do you think they should be moved to `.models.py` as instance methods of the model or the model manager?

chandrasekharan-zipstack (approved):
LGTM - left some minor comments

backend/workflow_manager/file_execution/models.py
```diff
@@ -11,6 +12,45 @@
...
+            workflow_execution=workflow_execution,
+            file_hash=file_hash,
+            file_path=file_path,
+        )
+
+        if is_created:
+            execution_file.file_name = file_name
+            execution_file.file_size = file_size
+            execution_file.mime_type = mime_type
+            execution_file.save()
```
chandrasekharan-zipstack, line 50: NIT: @muhammad-ali-e is it not possible to pass these as params to `get_or_create` rather than updating it again after saving it?

chandrasekharan-zipstack (approved):
LGTM - left a minor comment

backend/workflow_manager/file_execution/models.py
muhammad-ali-e: I think we can keep it like this for now. The` workflow_execution, file_hash, and file_path` fields have an `index` and unique constraint, which helps optimize queries. However, if we add fields like `file_name`, `file_size`, and `mime_type` to the `get_or_create` query, it could lead to slower queries, especially if those fields are not indexed, as the database would need to perform more extensive lookups.

vishnuszipstack (approved):

hari-kuriakose:
Migration looks good overall.

PR #894, opened 2024-12-12
Add result acknowledgment for API deployment task

muhammad-ali-e:
## What

-  Implemented result acknowledgment for API deployment task.
- Added functionality to acknowledge the result once the task is complete and remove it from the backend.

**Note:** This will not work for tasks that have already expired in existing system.

## Why

- To ensure that completed tasks are properly acknowledged and their results are cleared from celery's temperary result backend.

## How

-  Checked the task status using Celery's `AsyncResult`
- If the task is complete, invoked `forget()` to remove the result and updated the `result_acknowledged` flag in the database.

- No, this PR adds functionality without affecting any existing features. It only modifies how the task result is handled once the task is complete.

## Database Migrations

- Added a flag (`result_acknowledged`) to track whether the result has been acknowledged in `WorkflowExecution` model.

## Env Config

-  No

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


backend/workflow_manager/workflow_v2/models/execution.py
```diff
@@ -48,6 +48,9 @@ class Type(models.TextChoices):
     )
     # TODO: Restrict with an enum
     status = models.CharField(default="", db_comment="Current status of execution")
+    result_acknowledged = models.BooleanField(
```
chandrasekharan-zipstack, line 51: NIT: @muhammad-ali-e do you see a need to store the time at which the result was acknowledged? Maybe that alone can indicate whether the RESULT was obtained while also storing an extra info

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -381,13 +382,39 @@ def get_status_of_async_task(
...
+        It ensures that the task result is flagged as acknowledged in the database
+
+        Args:
+            execution (WorkflowExecution): WorkflowExecution instance
+        """
+        if not execution.result_acknowledged:
+            execution.result_acknowledged = True
+            execution.save()
+            logger.info(
+                f"ExecutionId [{execution.id}] - Task {execution.task_id} acknowledged"
```
chandrasekharan-zipstack, line 415:
```suggestion
                f"ExecutionID [{execution.id}] - Task {execution.task_id} acknowledged"
```

backend/api_v2/api_deployment_views.py
```diff
@@ -88,6 +88,9 @@ def get(
             response_status = status.HTTP_200_OK
             if not include_metadata:
                 response.remove_result_metadata_keys()
+        if response.result_acknowledged:
+            response_status = status.HTTP_406_NOT_ACCEPTABLE
```
chandrasekharan-zipstack, line 92: @muhammad-ali-e why is 406 chosen here? Can we avoid 4xx errors since its not necessarily a client error since the request did succeed as expected?
Do you think 200 or 204 might be better? I noticed [208 Already reported](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/208) as well but apparently it needs to be used in the context of web DAV

chandrasekharan-zipstack (approved):
LGTM - left some minor concerns. I just want us to recheck the status code of 406 here

backend/api_v2/api_deployment_views.py
ritwik-g: I think a 4xx is appropriate here. Since the client shouldn't send the request again for same transaction once the result is retrieved.

backend/workflow_manager/workflow_v2/models/execution.py
ritwik-g: Good suggestion. @muhammad-ali-e right now we might not have a use. But may be in future this could come in handy. But due to current priority my suggestion would be to keep it as another ticket and we can revisit it once the high priority items are over.

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -358,8 +358,9 @@ def _update_pipeline_status(
                 f"with workflow execution: {workflow_execution}"
             )
 
-    @staticmethod
+    @classmethod
```
ritwik-g, line 361: @muhammad-ali-e why was this changed from a static method to class method? Could this have any other side effects? 
Couldn't we have done the `cls._set_result_acknowledge` at the end where this call is being used? 

No need to make any changes. Trying to understand the reason for this change.

ritwik-g (approved):
Changes looks good. I have one comment which is more of a discussion point rather than a change request.

backend/workflow_manager/workflow_v2/models/execution.py
muhammad-ali-e: I also thought about this during development. We should do this later.

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: The method was changed to a class method to allow the use of cls for invoking `_set_result_acknowledge`.
 Moving c`ls._set_result_acknowledge` to where the method is called could be a possibility, but it would shift responsibility out of this method.

PR #900, opened 2024-12-17
[FEATURE] Changes for capturing metrics

Deepak-Kesavan:
## What

-

## Why

-

## How

-

-

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/prompt-service/src/unstract/prompt_service/constants.py
+++ b/prompt-service/src/unstract/prompt_service/constants.py
@@ -72,6 +72,8 @@ class PromptServiceContants:
     FILE_PATH = "file_path"
     HIGHLIGHT_DATA = "highlight_data"
     CONFIDENCE_DATA = "confidence_data"
+    TIME_TAKEN = "time_taken"
+    METRICS = "metrics"
 
 
 class RunLevel(Enum):
--- a/prompt-service/src/unstract/prompt_service/main.py
+++ b/prompt-service/src/unstract/prompt_service/main.py
@@ -109,6 +109,7 @@ def prompt_processor() -> Any:
         PSKeys.FILE_NAME: doc_name,
         PSKeys.CONTEXT: {},
     }
+    metrics: dict = {}
     variable_names: list[str] = []
     publish_log(
         log_events_id,
@@ -126,8 +127,7 @@ def prompt_processor() -> Any:
         prompt_text = output[PSKeys.PROMPT]
         chunk_size = output[PSKeys.CHUNK_SIZE]
         util = PromptServiceBaseTool(platform_key=platform_key)
-        index = Index(tool=util)
-
+        index = Index(tool=util, run_id=run_id, capture_metrics=True)
         if VariableExtractor.is_variables_present(prompt_text=prompt_text):
             prompt_text = VariableExtractor.replace_variables_in_prompt(
                 prompt=output,
@@ -188,6 +188,7 @@ def prompt_processor() -> Any:
                     **usage_kwargs,
                     PSKeys.LLM_USAGE_REASON: PSKeys.EXTRACTION,
                 },
+                capture_metrics=True,
             )
 
             embedding = Embedding(
@@ -489,6 +490,7 @@ def prompt_processor() -> Any:
                                 **usage_kwargs,
                                 PSKeys.LLM_USAGE_REASON: PSKeys.CHALLENGE,
                             },
+                            capture_metrics=True,
                         )
                         challenge = challenge_plugin["entrypoint_cls"](
                             llm=llm,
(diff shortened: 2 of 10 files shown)


tools/structure/src/main.py
chandrasekharan-zipstack, line 1: @Deepak-Kesavan please update the tool version in `config/properties.json`

prompt-service/src/unstract/prompt_service/constants.py
```diff
@@ -72,6 +72,8 @@ class PromptServiceContants:
     FILE_PATH = "file_path"
     HIGHLIGHT_DATA = "highlight_data"
     CONFIDENCE_DATA = "confidence_data"
+    TIME_TAKEN = "time_taken"
```
chandrasekharan-zipstack, line 75: NIT: Constant can be obtained from SDK's MetricsMixin as well

prompt-service/src/unstract/prompt_service/main.py
```diff
@@ -592,6 +594,18 @@ def prompt_processor() -> Any:
...
                     )
         finally:
+            challenge_metrics = (
+                {f"{challenge_llm.get_usage_reason()}_llm": challenge_llm.get_metrics()}
+                if enable_challenge
+                else {}
+            )
+            metrics.setdefault(prompt_name, {}).update(
+                {
+                    "context_retrieval": index.get_metrics(),
```
chandrasekharan-zipstack, line 604: @Deepak-Kesavan I think such usage reasons of `context_retrieval` should be moved into the corresponding SDK method that returns such a value. If I'm correct only `index.query_index()` corresponds to this. We might also need to have provision to pass another optional reason key that such callers can pass to indicate the use of an SDK method

tools/structure/src/main.py
```diff
@@ -242,6 +249,14 @@ def run(
             structured_output_dict[SettingsKeys.METADATA] = metadata
             structured_output = json.dumps(structured_output_dict)
 
+        metrics = structured_output[SettingsKeys.METRICS]
+        # Merge dictionaries
+        new_metrics = {
```
chandrasekharan-zipstack, line 254: @Deepak-Kesavan where will the key such as `indexing` be shown in the metrics? I noticed we had `context_retrieval` though. Do add an example of the response in the PR description

tools/structure/src/main.py
```diff
@@ -242,6 +252,21 @@ def run(
...
+                **index_metrics,
+            }
+        else:
+            # Merge dictionaries
+            new_metrics = {
+                key: {**metrics.get(key, {}), **index_metrics.get(key, {})}
+                for key in set(metrics)
+                | set(index_metrics)  # Union of keys from both dictionaries
+            }
+        structured_output_dict[SettingsKeys.METRICS] = new_metrics
```
chandrasekharan-zipstack, line 269: NIT: Suggest to add this as a separate function to reduce the growing size of this method

Deepak-Kesavan: Will handle this in upcoming PR.

tools/structure/src/main.py
Deepak-Kesavan: Added a sample response.

Made changes to show the key `indexing`

prompt-service/src/unstract/prompt_service/main.py
Deepak-Kesavan: @chandrasekharan-zipstack We can handle it iteratively when we display metrics in a more granular format.

chandrasekharan-zipstack (approved):

prompt-service/src/unstract/prompt_service/main.py
```diff
@@ -489,6 +490,7 @@ def prompt_processor() -> Any:
                                 **usage_kwargs,
                                 PSKeys.LLM_USAGE_REASON: PSKeys.CHALLENGE,
                             },
+                            capture_metrics=True,
```
gaya3-zipstack, line 493: Have a genera qn. Is the logic like: when include_metrics=true from API, we will pass capture_metrics = true and only then SDK genertes the metrics?

Deepak-Kesavan: @gaya3-zipstack 

No.
Earlier it was like that for include_metadata, now it is changed.
Now we always calculate metrics and metadata.
And remove the keys before sending it as response for API based on values sent on form-data

gaya3-zipstack: Ok. After the call, I understand the design issues that we have around this and why are are forced to send the boolean always as true disregarding the user request. When response comes, only then we go back and remove the metadata/metrics if user did not want it. Obviously better design would be to honour the flag at all times and constrict our computation accordingly. This is not a quick thing to get done. But let us consider this as a design improvement for future.

gaya3-zipstack (approved):
Looks good.

PR #904, opened 2024-12-18
feat: Updated ADAPTER_LLMW_MAX_POLLS to 120 for 1 hour extraction

chandrasekharan-zipstack:
## What

- Updated env to reduce extraction time from ~8.3 hours to ~1 hour

## Why

- After move to LLMW v2, multi page extractions of 1500 pages should be done in ~1 hour

- No, reduced extraction time only for LLMW v2

## Env Config

```
# ~60 mins (assuming it'll be enough to process 1500 pages with LLMW v2)
ADAPTER_LLMW_MAX_POLLS=120
```

## Related Issues or PRs

-

## Notes on Testing

- No explicit testing / validation done - logic based on these envs work though

## Checklist

### DIFF
--- a/backend/sample.env
+++ b/backend/sample.env
@@ -120,8 +120,8 @@ SESSION_COOKIE_AGE=86400
 # Time in seconds to wait before polling LLMWhisperer's status API
 ADAPTER_LLMW_POLL_INTERVAL=30
 # Total number of times to poll the status API.
-# 500 mins to allow 1500 (max pages limit) * 20 (approx time in sec to process a page)
-ADAPTER_LLMW_MAX_POLLS=1000
+# ~60 mins (assuming it'll be enough to process 1500 pages with LLMW v2)
+ADAPTER_LLMW_MAX_POLLS=120
 # Number of times to retry the /whisper-status API before failing the extraction
 ADAPTER_LLMW_STATUS_RETRIES=5
 


backend/sample.env
```diff
@@ -120,8 +120,8 @@ SESSION_COOKIE_AGE=86400
 # Time in seconds to wait before polling LLMWhisperer's status API
 ADAPTER_LLMW_POLL_INTERVAL=30
 # Total number of times to poll the status API.
-# 500 mins to allow 1500 (max pages limit) * 20 (approx time in sec to process a page)
-ADAPTER_LLMW_MAX_POLLS=1000
+# ~60 mins (assuming it'll be enough to process 1500 pages with LLMW v2)
+ADAPTER_LLMW_MAX_POLLS=120
```
ritwik-g, line 124: @chandrasekharan-zipstack but what about for the V1? Doesn't it use the same ENV?

harini-venkataraman, line 124: @chandrasekharan-zipstack Will this not cause timeout if it exceeds max 15min time which we have set in gunicorn ?

chandrasekharan-zipstack: @ritwik-g you're right, v1 uses the same envs. Shall we discourage support for v1 by changing this env (reducing the time / max possible pages) and instead enforce support with v2 for more number of pages?
Realistically speaking - I doubt if any user has such large extraction times. 

Worst case, 
1. either we'll have to let this env be and take action after we sunset v1
1. or introduce a new set of envs for v2 and update that (involves changes in the SDK, so I'm not a fan of this)

chandrasekharan-zipstack: @harini-venkataraman the timeout exists only for extractions that happens with a web UI. This large setting is mainly ideal for async pipeline based extractions where such gunicorn timeouts will not play a role

chandrasekharan-zipstack:
This PR can be updated / closed once [this SDK PR](https://github.com/Zipstack/unstract-sdk/pull/144) is merged and updated downstream to add separate envs for V1 and V2 of whisperer

PR #906, opened 2024-12-19
[FIX] Structure tool metrics error fix

Deepak-Kesavan:
## What

- Handled case where metrics is missing in the output

## Why

- Key error happens when metrics key is missing

## How

-

-

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/tools/structure/src/main.py
+++ b/tools/structure/src/main.py
@@ -253,7 +253,7 @@ class StructureTool(BaseTool):
             structured_output_dict[SettingsKeys.METADATA] = metadata
             structured_output = json.dumps(structured_output_dict)
 
-        metrics = structured_output_dict[SettingsKeys.METRICS]
+        metrics = structured_output_dict.get(SettingsKeys.METRICS, {})
         new_metrics = {}
         if tool_settings[SettingsKeys.ENABLE_SINGLE_PASS_EXTRACTION]:
             new_metrics = {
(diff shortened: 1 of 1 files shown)


tools/structure/src/main.py
chandrasekharan-zipstack, line 1: @Deepak-Kesavan please bump the tool version as well

tools/structure/src/main.py
```diff
@@ -253,7 +253,7 @@ def run(
             structured_output_dict[SettingsKeys.METADATA] = metadata
             structured_output = json.dumps(structured_output_dict)
 
-        metrics = structured_output_dict[SettingsKeys.METRICS]
+        metrics = structured_output_dict.get(SettingsKeys.METRICS, {})
```
chandrasekharan-zipstack, line 256: @Deepak-Kesavan since we always set `capture_metrics` to `True` -> why do you think this metrics key is missing?

chandrasekharan-zipstack (approved):
LGTM - but please ensure we bump the tool version in this PR itself @Deepak-Kesavan

tools/structure/src/main.py
Deepak-Kesavan: It was not handled for table and record type

harini-venkataraman, line 256: @Deepak-Kesavan Hope it is handled for `KeyError` in places where we use this variable when the dict is empty.

harini-venkataraman (approved):
Added a minor comment, LGTM otherwise.

tools/structure/src/main.py
Deepak-Kesavan: @harini-venkataraman Yes it is handled.

tools/structure/src/main.py
chandrasekharan-zipstack: Necessary changes were made in the cloud patch and the image was updated. OSS tool needs no action here - hence ignoring the bump for this

PR #907, opened 2024-12-19
[Fix] Prompt studio Coverage

jagadeeswaran-zipstack:
## What

- Fixed coverage in prompt studio
- Fixed local variable reference issue

## Why

- Logical mistake where variable in try-catch referred out side the block which throws local variable reference issue

## How

-

- No, fix made on coverage it won't affect other features

## Database Migrations

- N/A

## Env Config

- N/A

## Relevant Docs

- N/A

## Related Issues or PRs

- N/A

## Dependencies Versions

- N/A

## Notes on Testing

-

## Screenshots
image

## Checklist

### DIFF
--- a/backend/prompt_studio/prompt_studio_core_v2/serializers.py
+++ b/backend/prompt_studio/prompt_studio_core_v2/serializers.py
@@ -44,48 +44,70 @@ class CustomToolSerializer(IntegrityErrorMixin, AuditSerializer):
 
     def to_representation(self, instance):  # type: ignore
         data = super().to_representation(instance)
+        default_profile = None
+
+        # Fetch summarize LLM profile
         try:
-            profile_manager = ProfileManager.objects.get(
+            summarize_profile = ProfileManager.objects.get(
                 prompt_studio_tool=instance, is_summarize_llm=True
             )
-            data[TSKeys.SUMMARIZE_LLM_PROFILE] = profile_manager.profile_id
+            data[TSKeys.SUMMARIZE_LLM_PROFILE] = summarize_profile.profile_id
         except ObjectDoesNotExist:
             logger.info(
-                "Summarize LLM profile doesnt exist for prompt tool %s",
+                "Summarize LLM profile doesn't exist for prompt tool %s",
                 str(instance.tool_id),
             )
+
+        # Fetch default LLM profile
         try:
-            profile_manager = ProfileManager.get_default_llm_profile(instance)
-            data[TSKeys.DEFAULT_PROFILE] = profile_manager.profile_id
+            default_profile = ProfileManager.get_default_llm_profile(instance)
+            data[TSKeys.DEFAULT_PROFILE] = default_profile.profile_id
         except DefaultProfileError:
             logger.warning(
-                "Default LLM profile doesnt exist for prompt tool %s",
+                "Default LLM profile doesn't exist for prompt tool %s",
                 str(instance.tool_id),
             )
-        prompt_instance: ToolStudioPrompt = ToolStudioPrompt.objects.filter(
+
+        # Fetch prompt instances
+        prompt_instances = ToolStudioPrompt.objects.filter(
             tool_id=data.get(TSKeys.TOOL_ID)
         ).order_by("sequence_number")
-        data[TSKeys.PROMPTS] = []
+
+        if not prompt_instances.exists():
+            data[TSKeys.PROMPTS] = []
+            return data
+
+        # Process prompt instances
         output: list[Any] = []
-        # Appending prompt instances of the tool for FE Processing
-        if prompt_instance.count() != 0:
-            for prompt in prompt_instance:
-                profile_manager_id = prompt.prompt_id
-                if instance.single_pass_extraction_mode:
-                    # use projects default profile
-                    profile_manager_id = profile_manager.profile_id
-                prompt_serializer = ToolStudioPromptSerializer(prompt)
+        for prompt in prompt_instances:
+            prompt_serializer = ToolStudioPromptSerializer(prompt)
+            serialized_data = prompt_serializer.data
+
+            # Determine coverage
+            coverage: list[Any] = []
+            profile_manager_id = prompt.profile_manager
+            if default_profile and instance.single_pass_extraction_mode:
+                profile_manager_id = default_profile.profile_id
+
+            if profile_manager_id:
                 coverage = OutputManagerUtils.get_coverage(
                     data.get(TSKeys.TOOL_ID),
                     profile_manager_id,
                     prompt.prompt_id,
                     instance.single_pass_extraction_mode,
                 )
-                serialized_data = prompt_serializer.data
-                serialized_data["coverage"] = coverage
-                output.append(serialized_data)
-            data[TSKeys.PROMPTS] = output
+            else:
+                logger.info(
+                    "Skipping coverage calculation for prompt %s "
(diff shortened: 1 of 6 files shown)


backend/prompt_studio/prompt_studio_core_v2/serializers.py
```diff
@@ -44,48 +44,70 @@ class Meta:
...
-                    profile_manager_id = profile_manager.profile_id
-                prompt_serializer = ToolStudioPromptSerializer(prompt)
+        for prompt in prompt_instances:
+            prompt_serializer = ToolStudioPromptSerializer(prompt)
+            serialized_data = prompt_serializer.data
+
+            # Determine coverage
+            coverage: list[Any] = []
+            profile_manager_id = prompt.profile_manager
+            if default_profile and instance.single_pass_extraction_mode:
```
chandrasekharan-zipstack, line 89: @jagadeeswaran-zipstack if there's no `default_profile` present shouldn't we throw an error and ask the user to create one? Currently do we allow the user to create prompts without a default profile?

backend/prompt_studio/prompt_studio_core_v2/serializers.py
```diff
@@ -44,48 +44,70 @@ class Meta:
...
                     profile_manager_id,
                     prompt.prompt_id,
                     instance.single_pass_extraction_mode,
                 )
-                serialized_data = prompt_serializer.data
-                serialized_data["coverage"] = coverage
-                output.append(serialized_data)
-            data[TSKeys.PROMPTS] = output
+            else:
+                logger.info(
```
chandrasekharan-zipstack, line 100: @jagadeeswaran-zipstack under what scenario can this happen? Why would a profile ID go missing?
Is it when we delete all profiles in a project? Ideally this should not be possible at all - 
1. we should throw an error if there's only 1 profile and the user tries to delete it
1. Or we should throw an error early if there's no profile and a user tries to run a prompt

jagadeeswaran-zipstack: @chandrasekharan-zipstack Default profile and only 1 profile can't be deleted. This is for old projects where users might have deleted all the project. we will change this logic in future.

backend/prompt_studio/prompt_studio_core_v2/serializers.py
harini-venkataraman, line 89: @jagadeeswaran-zipstack How will the behavior be if the default profile is not selected and single pass is enabled? Are we prompting anywhere to select default profile if single pass is enabled?

jagadeeswaran-zipstack: Yes, we're allowing users to create prompts without a default profile. This was the behaviour from the beginning.

chandrasekharan-zipstack (approved):
BE changes LGTM

backend/prompt_studio/prompt_studio_core_v2/serializers.py
```diff
@@ -44,48 +44,70 @@ class Meta:
...
         except DefaultProfileError:
             logger.warning(
-                "Default LLM profile doesnt exist for prompt tool %s",
+                "Default LLM profile doesn't exist for prompt tool %s",
                 str(instance.tool_id),
             )
-        prompt_instance: ToolStudioPrompt = ToolStudioPrompt.objects.filter(
+
+        # Fetch prompt instances
+        prompt_instances = ToolStudioPrompt.objects.filter(
```
harini-venkataraman, line 72: @jagadeeswaran-zipstack pre-commit hook was run over this commit? It should have raised error for typing such declarations.

jagadeeswaran-zipstack: Yes, Ran pre-commit hook over this commit. If not in my local it would have caught in pre-commit hook bot right?

backend/prompt_studio/prompt_studio_core_v2/serializers.py
jagadeeswaran-zipstack: Sure :+1:

frontend/src/components/custom-tools/prompt-card/PromptCardItems.jsx
```diff
@@ -78,10 +80,18 @@ function PromptCardItems({
...
+  const promptProfile = promptDetails?.profile_manager || defaultLlmProfile;
+  const promptOutputKey = generatePromptOutputKey(
+    promptId,
+    docId,
+    promptProfile,
+    singlePassExtractMode,
+    true
+  );
+  const promptCoverage =
+    promptOutputs[promptOutputKey]?.coverage || coverageCountData;
```
tahierhussain, line 94: @jagadeeswaran-zipstack This is a slight corner case here. The value of coverage can be `0`, which is valid. However, if `promptOutputs[promptOutputKey]?.coverage` is `0`, it will be considered as false and fall back to the value of `coverageCountData`, which is incorrect.

To fix this, you can use the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) (`??`) instead of `||`. Here's the code:

```suggestion
  const promptCoverage = promptOutputs[promptOutputKey]?.coverage ?? coverageCountData;
```

This ensures that the fallback value is only used if `promptOutputs[promptOutputKey]?.coverage` is `null` or `undefined`, and valid values like `0` are handled correctly.

frontend/src/components/custom-tools/manage-docs-modal/ManageDocsModal.jsx
```diff
@@ -543,21 +545,32 @@ function ManageDocsModal({
...
 
+  const removeIdFromCoverage = (data, idToRemove) => {
+    return Object.entries(data).reduce((updatedData, [key, value]) => {
+      // Create a new object for the current entry
+      updatedData[key] = {
+        ...value,
+        // Update the coverage array if it exists
+        coverage: value?.coverage
+          ? value?.coverage?.filter((id) => id !== idToRemove)
+          : value?.coverage,
```
tahierhussain, line 568: @jagadeeswaran-zipstack The fallback to `value?.coverage` is unnecessary since filter inherently returns the same array if no changes are made. This simplifies the coverage update logic.
```suggestion
        coverage: value.coverage?.filter((id) => id !== idToRemove),
```

tahierhussain (commented):
Requested a few changes. Everything else looks good to me.

@jagadeeswaran-zipstack Make sure to remove any logic or code previously used to handle coverage before implementing this change.

frontend/src/components/custom-tools/prompt-card/PromptCardItems.jsx
jagadeeswaran-zipstack: @tahierhussain `promptOutputs[promptOutputKey]?.coverage` returns array of docId's not actual count that's why we have this logic. Like i mentioned if there is no output for a prompt then it will fallback to initial store where we store coverage data.

frontend/src/components/custom-tools/manage-docs-modal/ManageDocsModal.jsx
jagadeeswaran-zipstack: Make sense, Ill make the required changes. Thanks

tahierhussain (approved):

harini-venkataraman (approved):
LGTM

PR #1048, opened 2025-01-06
Health check API for backend & Prompt Sevice

harini-venkataraman:
## What
PR focus on health check API for prompt service & platform service.
...

## Why
To support readiness and liveness probe for k8s deployment.
...

## How
Exposing an endpoint, which is public retuning 200 OK.
...

No, this PR introduces new endpoints.
...

## Database Migrations
Not applicable.
...

## Env Config
Not applicable.
...

## Relevant Docs
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request
...

## Related Issues or PRs

-

## Dependencies Versions
Not applicable.
...

## Notes on Testing
Tested the API using rest client.
Created an deployment with sample deployment.yaml using minikube and checked the probes.
...

## Screenshots
image

image

## Checklist

### DIFF
--- a/backend/backend/settings/base.py
+++ b/backend/backend/settings/base.py
@@ -114,20 +114,20 @@ PROMPT_STUDIO_FILE_PATH = os.environ.get(
 )
 X2TEXT_HOST = os.environ.get("X2TEXT_HOST", "http://localhost")
 X2TEXT_PORT = os.environ.get("X2TEXT_PORT", 3004)
-STRUCTURE_TOOL_IMAGE_URL = get_required_setting("STRUCTURE_TOOL_IMAGE_URL")
-STRUCTURE_TOOL_IMAGE_NAME = get_required_setting("STRUCTURE_TOOL_IMAGE_NAME")
-STRUCTURE_TOOL_IMAGE_TAG = get_required_setting("STRUCTURE_TOOL_IMAGE_TAG")
-WORKFLOW_DATA_DIR = os.environ.get("WORKFLOW_DATA_DIR")
-API_STORAGE_DIR = os.environ.get("API_STORAGE_DIR")
+STRUCTURE_TOOL_IMAGE_URL = get_required_setting("STRUCTURE_TOOL_IMAGE_URL", "")
+STRUCTURE_TOOL_IMAGE_NAME = get_required_setting("STRUCTURE_TOOL_IMAGE_NAME", "")
+STRUCTURE_TOOL_IMAGE_TAG = get_required_setting("STRUCTURE_TOOL_IMAGE_TAG", "")
+WORKFLOW_DATA_DIR = os.environ.get("WORKFLOW_DATA_DIR", "")
+API_STORAGE_DIR = os.environ.get("API_STORAGE_DIR", "")
 CACHE_TTL_SEC = os.environ.get("CACHE_TTL_SEC", 10800)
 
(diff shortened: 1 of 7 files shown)


backend/backend/settings/base.py
```diff
@@ -114,20 +114,20 @@ def get_required_setting(
 )
 X2TEXT_HOST = os.environ.get("X2TEXT_HOST", "http://localhost")
 X2TEXT_PORT = os.environ.get("X2TEXT_PORT", 3004)
-STRUCTURE_TOOL_IMAGE_URL = get_required_setting("STRUCTURE_TOOL_IMAGE_URL")
-STRUCTURE_TOOL_IMAGE_NAME = get_required_setting("STRUCTURE_TOOL_IMAGE_NAME")
-STRUCTURE_TOOL_IMAGE_TAG = get_required_setting("STRUCTURE_TOOL_IMAGE_TAG")
-WORKFLOW_DATA_DIR = os.environ.get("WORKFLOW_DATA_DIR")
-API_STORAGE_DIR = os.environ.get("API_STORAGE_DIR")
+STRUCTURE_TOOL_IMAGE_URL = get_required_setting("STRUCTURE_TOOL_IMAGE_URL", "")
```
chandrasekharan-zipstack, line 117: Any reason why empty string defaults were added in some of these envs? The point of using `get_required_setting` was mainly to throw an error if its not defined. In such cases the platform might not be usable or we may not handle the error correctly

gaya3-zipstack (commented):
@chandrasekharan-zipstack 's comment on env seems a valid question. Will wait for that. Other than that, I do not see any concern in the PR.

backend/backend/settings/base.py
harini-venkataraman: It was added for debugging purposes, and got committed along with other changes. No intended change.

harini-venkataraman:
> @chandrasekharan-zipstack 's comment on env seems a valid question. Will wait for that. Other than that, I do not see any concern in the PR.

Addressed the comment, @gaya3-zipstack

ritwik-g (approved):

gaya3-zipstack (approved):
Looks good

chandrasekharan-zipstack:
@harini-venkataraman shouldn't this public API also have the path prefix of `api/v1` ? I believe it should be `/api/v1/health` similar to other public APIs added [here](https://github.com/Zipstack/unstract/blob/2ac6a4207003c01136c82a8f6db42779f74cf973/backend/backend/settings/base.py#L446-L454).
Can you check and raise a PR to correct this here and the place where its used? Else create a tech debt ticket to track this.
cc: @ritwik-g @hari-kuriakose

ritwik-g:
> @harini-venkataraman shouldn't this public API also have the path prefix of `api/v1` ? I believe it should be `/api/v1/health` similar to other public APIs added [here](https://github.com/Zipstack/unstract/blob/2ac6a4207003c01136c82a8f6db42779f74cf973/backend/backend/settings/base.py#L446-L454).
> Can you check and raise a PR to correct this here and the place where its used? Else create a tech debt ticket to track this.
> cc: @ritwik-g @hari-kuriakose 

@chandrasekharan-zipstack not necessary. Because this health check is for the container and not specific for v1 path. Also we won't expose this publicly in the ALB

chandrasekharan-zipstack:
Alright @ritwik-g - I notice that we have both the APIs present `/health` and `/api/v1/health` available btw. Should not be a concern I guess but just pointing it out

PR #1078, opened 2025-01-20
[Fix] Multi Doc support in prompt studio

jagadeeswaran-zipstack:
## What

- MS related file types are failing to be converted.

## Why

- Since we're reading first few bytes to determine mimetype, In MS documents mimetype related data are stored at the end of the document. So now we're using full length to guess mimetype.

## How

-

-

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Screenshots
image

## Checklist

### DIFF
--- a/backend/utils/FileValidator.py
+++ b/backend/utils/FileValidator.py
@@ -8,7 +8,7 @@ from django.core.files.uploadedfile import InMemoryUploadedFile
 from django.template.defaultfilters import filesizeformat
 from django.utils.translation import gettext_lazy as _
 from typing_extensions import NotRequired, Unpack
-from utils.constants import FileReader
+from unstract.sdk.file_storage.constants import FileOperationParams
 
 
 class FileValidationParam(TypedDict):
@@ -71,7 +71,9 @@ class FileValidator:
 
     def _check_file_mime_type(self, file: InMemoryUploadedFile) -> None:
         # TODO: Need to optimise, istead of reading entire file.
-        mimetype = magic.from_buffer(file.read(FileReader.READ_FULL_LENGTH), mime=True)
+        mimetype = magic.from_buffer(
+            file.read(FileOperationParams.READ_ENTIRE_LENGTH), mime=True
+        )
         file.seek(0)  # Reset the file pointer to the start
 
         if self.allowed_mimetypes and mimetype not in self.allowed_mimetypes:
--- a/backend/utils/FileValidator.py
+++ b/backend/utils/FileValidator.py
@@ -8,6 +8,7 @@ from django.core.files.uploadedfile import InMemoryUploadedFile
 from django.template.defaultfilters import filesizeformat
 from django.utils.translation import gettext_lazy as _
 from typing_extensions import NotRequired, Unpack
+from utils.constants import FileReader
 
 
 class FileValidationParam(TypedDict):
@@ -69,7 +70,7 @@ class FileValidator:
             raise ValidationError(message)
 
     def _check_file_mime_type(self, file: InMemoryUploadedFile) -> None:
-        mimetype = magic.from_buffer(file.read(2048), mime=True)
+        mimetype = magic.from_buffer(file.read(FileReader.READ_FULL_LENGTH), mime=True)
         file.seek(0)  # Reset the file pointer to the start
 
         if self.allowed_mimetypes and mimetype not in self.allowed_mimetypes:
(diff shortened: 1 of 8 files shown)


backend/utils/FileValidator.py
```diff
@@ -69,7 +70,7 @@ def _check_file_extension(self, file: InMemoryUploadedFile) -> None:
             raise ValidationError(message)
 
     def _check_file_mime_type(self, file: InMemoryUploadedFile) -> None:
-        mimetype = magic.from_buffer(file.read(2048), mime=True)
+        mimetype = magic.from_buffer(file.read(FileReader.READ_FULL_LENGTH), mime=True)
```
gaya3-zipstack, line 73: @jagadeeswaran-zipstack PR is fine. But pls consider the memory issues you had concerns about...
@muhammad-ali-e Do you have any inputs here?

jagadeeswaran-zipstack: Yes @gaya3-zipstack , Since we took similar approach in our system as you mentioned. It will be better to get inputs from @muhammad-ali-e .

ritwik-g: @jagadeeswaran-zipstack instead of reading the full file can we consider reading the first 2048 bytes and if the mimetype is not available then the last few bytes (don't know exact required number). Or something like concatenating first 2048 bytes and last 2048 bytes. Would something like that work?

jagadeeswaran-zipstack: Haven't tried with last few bytes. I'll try that and let you know.

gaya3-zipstack: @jagadeeswaran-zipstack  
There is also mime = magic.from_file(file_path, mime=True). I think here, libmagic uses a combination of strategies to identify file type and may cover what @ritwik-g has suggested (combination fo first and last bytes.

ritwik-g: @jagadeeswaran-zipstack this is a sample code I tried. See if this is helpful
```python
def check_file_mime_type(file) -> None:
        mime_data_search_length = 2048
        total_length = file.seek(0, 2)
        file.seek(0)

        if total_length <= 2 * mime_data_search_length:
            mime_data = file.read()
        else:
            mime_data = file.read(mime_data_search_length)
            end_seek = total_length - mime_data_search_length
            file.seek(end_seek)
            mime_data += file.read()

        mimetype = magic.from_buffer(mime_data, mime=True)
        file.seek(0)
        return mimetype
```

jagadeeswaran-zipstack: @gaya3-zipstack we can't use magic.from_file since we have in the file in-memory. That's why they provide read from_buffer.

jagadeeswaran-zipstack: Tried with that logic @ritwik-g. Like we discussed, mimetype related data are stored in middle of the entire bytes sometimes.

ritwik-g: Okay. Looks like the way the mime type is decided is complex than checking few bytes it looks like. So we might need to read entire data. If possible using `file.name` and `magic.from_file` as @gaya3-zipstack suggested might be better approach. We don't need to make this a blocker. We can add a TODO comment and later update this as per @muhammad-ali-e 's suggestion.

gaya3-zipstack (approved):
Looks good in the light of @ritwik-g 's [comment](https://github.com/Zipstack/unstract/pull/1078#discussion_r1921963303).

backend/utils/FileValidator.py
jagadeeswaran-zipstack: Got it @ritwik-g

backend/utils/constants.py
```diff
@@ -72,3 +72,7 @@ class ExecutionLogConstants:
     PERIODIC_TASK_NAME_V2 = "workflow_log_history_v2"
     TASK = "workflow_manager.workflow.execution_log_utils.consume_log_history"
     TASK_V2 = "consume_log_history"
+
+
+class FileReader:
+    READ_FULL_LENGTH = -1
```
chandrasekharan-zipstack, line 78: NIT: @jagadeeswaran-zipstack instead of introducing another constant, try to reuse the same [here](https://github.com/Zipstack/unstract-sdk/blob/2675e2baf861d01bbcc4bddbb9485fb53cca964c/src/unstract/sdk/file_storage/constants.py#L5)

chandrasekharan-zipstack (approved):
LGTM - will let you take a call on the other unresolved comment regarding memory concern while reading the file

backend/utils/constants.py
jagadeeswaran-zipstack: Done :+1:

PR #1079, opened 2025-01-20
Fixed invalid api key format raising 500

pk-zipstack:
## What  
- ~~**Added API key format validation**: Ensures the API key provided in the `Authorization` header is a valid UUID.~~
- ~~**Updated `BaseAPIKeyValidator`**:~~
  - ~~Modified the `validate_api_key` decorator to validate API key format using Python’s `uuid` module.~~ 
- ~~**Enhanced error messages**: Returns a `400 BadRequest` error when the API key format is invalid.~~
- **New Approach**:
  -  Modified the `validate_api_key` function in `key_helper.py` to capture `ValidationError` and raise `UnauthozisedKey()` error.

## Why  
- **Prevent errors due to invalid API key formats**: An invalid API key format caused unnecessary `500 Internal Server Error` responses.  

## How  
- **Changes in `KeyHelper` class**:  
  - Updated the static method `validate_api_key` to catch Django's `ValidationError` that occurs during UUID validation and raise an `UnauthorizedKey` exception, ensuring consistent error handling for invalid API key formats.  

- No, this change only enhances API key validation. Existing workflows with valid API keys will remain unaffected.  
- **Why not?** All existing keys follow the UUID format, ensuring backward compatibility.  

## Database Migrations  
- None  

## Env Config  
- None  

## Relevant Docs  
- None

## Related Issues or PRs  
- https://zipstack.atlassian.net/browse/UN-1911?focusedCommentId=16014

## Dependencies Versions  
- None  

## Notes on Testing  
- Verified proper validation of valid and invalid API keys.  
- Confirmed correct error codes (`400 Forbidden`) for invalid formats.  

## Screenshots  
1. **Error for invalid API key format**

image

## Checklist  


backend/api_v2/api_key_validator.py
```diff
@@ -39,6 +41,8 @@ def wrapper(self: Any, request: Request, *args: Any, **kwargs: Any) -> Any:
                 api_key = authorization_header.split(" ")[1]
             if not api_key:
                 raise Forbidden("Missing api key")
+            if not cls.is_valid_uuid(api_key):
+                raise BadRequestException("Invalid API key format. Expected a UUID.")
```
ritwik-g, line 45: @hari-kuriakose @pk-zipstack @chandrasekharan-zipstack why are we adding this validation? Also I don't think in our validation message we should mention the expected format. 

What I am wondering is why is this causing an issue as 500? Shouldn't this be 401 - authentication error? I think we should be giving that error not validation error. Because what ever the format of key if it is not proper key that is all we need to do.

hari-kuriakose: @ritwik-g Agree here.

1. We need not check the format of the API key and also don't mention that expected format is UUID in any error message.
2. On receiving invalid API key, always return 401.

pk-zipstack: @ritwik-g @hari-kuriakose Thank you for your feedback! I've made the changes to ensure the code now throws a `401 Forbidden error` whenever an invalid API key is provided. I’d appreciate it if you could take a look at the updated code and let me know if it’s all good.

backend/api_v2/key_helper.py
```diff
@@ -29,7 +30,7 @@ def validate_api_key(
             api_key_instance: APIKey = APIKey.objects.get(api_key=api_key)
             if not KeyHelper.has_access(api_key_instance, instance):
                 raise UnauthorizedKey()
-        except APIKey.DoesNotExist:
+        except (APIKey.DoesNotExist, ValidationError):
```
muhammad-ali-e, line 33: Is this ValidationError raise from APIKey model?

muhammad-ali-e (approved):
LGTM

backend/api_v2/key_helper.py
pk-zipstack: @muhammad-ali-e Yes, it is getting raised from the APIKey model. Since the `APIKey` in the model is a `UUIDField,` any key passed is trying to be converted to a `UUID.` If the key is not a valid UUID string, it results in a `ValidationError`.

chandrasekharan-zipstack (approved):

PR #1091, opened 2025-01-24
updated migration with datamigation from empty string to null

muhammad-ali-e:
## What

-  Fix for the  Error django.db.utils.DataError: invalid input syntax for type uuid: ""
- Added: Datamigration to migrate Empty string  to Nnull
- Updated platform service to insert null value for run_id instead of emtpy string

## Why

-  Error django.db.utils.DataError: invalid input syntax for type uuid: ""

## How

-

-

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/backend/usage_v2/migrations/0002_alter_usage_run_id.py
+++ b/backend/usage_v2/migrations/0002_alter_usage_run_id.py
@@ -3,6 +3,11 @@
 from django.db import migrations, models
 
 
+def convert_empty_to_null(apps, schema_editor):
+    usage = apps.get_model("usage_v2", "Usage")
+    usage.objects.filter(run_id="").update(run_id=None)
+
+
 class Migration(migrations.Migration):
 
     dependencies = [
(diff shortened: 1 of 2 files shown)


backend/usage_v2/migrations/0002_alter_usage_run_id.py
```diff
@@ -3,13 +3,21 @@
...
+
 class Migration(migrations.Migration):
 
     dependencies = [
         ("usage_v2", "0001_initial"),
     ]
 
     operations = [
+        migrations.RunPython(
+            convert_empty_to_null, reverse_code=migrations.RunPython.noop
```
chandrasekharan-zipstack, line 19: NIT: You could add a reverse operation as well for this by marking entries which have `None` back to `""`. However that might be possible only if the column did not have records with an intentional NULL in them

chandrasekharan-zipstack (approved):

backend/usage_v2/migrations/0002_alter_usage_run_id.py
muhammad-ali-e: @chandrasekharan-zipstack I believe charecter feild can be set to NULL.

PR #1100, opened 2025-01-28
feat: Reload support for backend, changes to entrypoint script

chandrasekharan-zipstack:
## What

- Added options for reload and graceful timeout to improve dev experience for `backend`
- Added support to override docker compose config to help run services with less memory
- `entrypoint.sh` changes to perform migration and run in dev mode
- Added `inotify` dev dependency to perform better file watching with fewer resources

## Why

- Reload option helps with looking for file changes and performs a hot reload of the django server helping the dev experience

## How

- Added as a parameter to the script. Can be used with `pdm run backend --dev`

- No, only changes to improve dev experience

## Relevant Docs

- [reload option](https://docs.gunicorn.org/en/latest/settings.html#reload)
- [merge compose files](https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/)

## Dependencies Versions

- `inotify>=0.2.10`

## Notes on Testing

- Tried making changes locally and tested the reload feature

## Screenshots
image
- Reload behaviour
image

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -636,6 +636,7 @@ docker/*.env
 !docker/sample*.env
 docker/public_tools.json
 docker/proxy_overrides.yaml
+docker/compose.override.yaml
 docker/workflow_data/
 
 # Tool development
--- a/backend/backend/wsgi.py
+++ b/backend/backend/wsgi.py
@@ -6,7 +6,9 @@ For more information on this file, see
 https://docs.djangoproject.com/en/4.2/howto/deployment/wsgi/
 """
 
+import logging
 import os
+import time
 
 from django.conf import settings
 from django.core.wsgi import get_wsgi_application
@@ -15,11 +17,16 @@ from utils.log_events import start_server
 
 load_dotenv()
 
+logger = logging.getLogger(__name__)
+
 os.environ.setdefault(
     "DJANGO_SETTINGS_MODULE",
     os.environ.get("DJANGO_SETTINGS_MODULE", "backend.settings.dev"),
 )
 
+start_time = time.time()
 django_app = get_wsgi_application()
+logger.info(f"WSGI application initialized in {(time.time() - start_time):.3f} seconds")
+
 
 application = start_server(django_app, f"{settings.PATH_PREFIX}/socket")
--- a/backend/backend/wsgi.py
+++ b/backend/backend/wsgi.py
@@ -24,9 +24,10 @@ os.environ.setdefault(
     os.environ.get("DJANGO_SETTINGS_MODULE", "backend.settings.dev"),
 )
 
-start_time = time.time()
+wsgi_start_time = time.perf_counter()
 django_app = get_wsgi_application()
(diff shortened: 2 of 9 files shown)


backend/backend/wsgi.py
```diff
@@ -15,11 +17,16 @@
...
+
 os.environ.setdefault(
     "DJANGO_SETTINGS_MODULE",
     os.environ.get("DJANGO_SETTINGS_MODULE", "backend.settings.dev"),
 )
 
+start_time = time.time()
 django_app = get_wsgi_application()
+logger.info(f"WSGI application initialized in {(time.time() - start_time):.3f} seconds")
+
```
ritwik-g, line 30: May be use `time.perf_counter()` . This might be more accurate for these sort of use case

ritwik-g (approved):
Great initiative solving real pain point for the developers. The changes looks good. Let's keep improving small bits in similar fashion

backend/backend/wsgi.py
```diff
@@ -15,11 +17,16 @@
...
+logger = logging.getLogger(__name__)
+
 os.environ.setdefault(
     "DJANGO_SETTINGS_MODULE",
     os.environ.get("DJANGO_SETTINGS_MODULE", "backend.settings.dev"),
 )
 
+start_time = time.time()
 django_app = get_wsgi_application()
+logger.info(f"WSGI application initialized in {(time.time() - start_time):.3f} seconds")
```
chandrasekharan-zipstack, line 29:
```suggestion
wsgi_start_time = time.perf_counter()
django_app = get_wsgi_application()
wsgi_init_elapsed = time.perf_counter() - wsgi_start_time
logger.info(f"WSGI application initialized in {wsgi_init_elapsed:.3f} seconds")
```

.gitignore
```diff
@@ -636,6 +636,7 @@ docker/*.env
 !docker/sample*.env
 docker/public_tools.json
 docker/proxy_overrides.yaml
+docker/compose.override.yaml
```
hari-kuriakose, line 639: @chandrasekharan-zipstack This might not be needed.

We can check in the actual overrides itself. It will be activated only when you pass them via the `-f` flag, hence mentioned.

hari-kuriakose (approved):

Extending this we can have `docker/compose_overrides/worker.yaml` etc and check them in. These will become built-in templates which can be picked by the devs and used as desired. See https://github.com/Zipstack/unstract/pull/1100#discussion_r1936708904 too.

Great initiative and really happy to see this. Let's keep improving one small thing at a time. All the best!

chandrasekharan-zipstack:
@hari-kuriakose regarding your comment, I've dropped a reply [here](https://github.com/Zipstack/unstract/pull/1100#discussion_r1937122359). Next steps would be to adopt something similar for other entrypoints.
I'm also trying to mount the backend code directly to develop on docker and checking out devcontainers to see if that can be useful for us

PR #1123, opened 2025-02-08
FIX: Ensure User Details Are Set Even if Session Validity Check Fails

tahierhussain:
## What

If any of the session validity check APIs fail, the flow moves to the `catch` block, causing us to miss setting the user details extracted from the successful session validity checks.

## Why

This was causing issues because certain pages required basic user details, even if some session validity check APIs failed.

## How

The logic for setting user details has been moved from the try block to the finally block. This ensures that any user details extracted from the session validity APIs are stored in the global state, even if some of the APIs fail.

Yes. This is a small change that's been made. But since this is a very high level component that gets executed when the app loads, there are chances of this breaking any existing features.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

### The subscription expired page is loading the admin message as expected
Screenshot from 2025-02-08 12-13-50

## Checklist

### DIFF
--- a/frontend/src/hooks/useSessionValid.js
+++ b/frontend/src/hooks/useSessionValid.js
@@ -59,6 +59,7 @@ function useSessionValid() {
     return false;
   };
   return async () => {
+    let userAndOrgDetails = null;
     try {
       const userSessionData = await userSession();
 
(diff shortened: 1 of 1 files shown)


harini-venkataraman (approved):
LGTM

jaseemjaskp (approved):

PR #1125, opened 2025-02-10
refactor: FE dockerfile optimized for better caching

chandrasekharan-zipstack:
## What

- Refactored FE dockerfile
  - Used a smaller base image
  - Copied package.json first to help install and cache for FE dependent libraries first, then copied code

## Why

- To reduce FE build time (**22.28% faster**)

## How

- Used a smaller base image
- Restructured some instructions to make better use of docker's caching

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

- No, tested it locally with a smaller base image

## Related Issues or PRs

- #1126 - Similar optimization

## Notes on Testing

- Built FE docker without these changes and with these changes to compare time taken
- Made small changes to FE to simulate code changes and rebuilds - able to use the platform afterwards

## Screenshots
1. Before making changes - `140s` to rebuild image after code changes (did a git pull here for the code changes)
image

1. After making changes - `108.8s` to rebuild image after code changes
image

## Checklist

I have read and understood the [Contribution Guidelines](https://docs.unstract.com/unstract/contributing/unstract/).

PR #1126, opened 2025-02-10
refactor: Dockerfile optimization to build faster

chandrasekharan-zipstack:
## What

- Refactored prompt-service dockerfile
  - Reordered instructions to cache install of gunicorn and other libs
- Refactored all other service's Dockerfiles in a similar fashion to cache layers better

## Why

- Better caching results in lesser Dockerfile build times (**11.67% faster** for prompt-service, didn't measure for others)

## How

- Installing libs that don't change often earlier to leverage Docker's caching

- No, only reordered instructions

## Related Issues or PRs

- #1125 - Similar optimization

## Notes on Testing

- Tried building locally and ran some prompts from Prompt studio
- TODO: Test behaviour with tool - easier to do this in staging

## Screenshots
[for prompt-service alone]
1. Before optimization (`252.7s`)
image
1. After optimization (`223.2s`)
image
1. Tool run in a workflow
image

## Checklist


Deepak-Kesavan (approved):

Deepak-Kesavan:
@chandrasekharan-zipstack I see 3 new sonar issues. Pleas check it.

docker/dockerfiles/platform.Dockerfile
```diff
@@ -29,27 +30,21 @@ USER unstract
...
-COPY --chown=unstract ${BUILD_CONTEXT_PATH} .
+# Create venv and install gunicorn and other deps in it
+RUN pdm venv create -w virtualenv --with-pip && \
+    . .venv/bin/activate && \
+    pip install --no-cache-dir gunicorn \
+    # Install opentelemetry for instrumentation
+    opentelemetry-distro opentelemetry-exporter-otlp && \
+    opentelemetry-bootstrap -a install
+
+COPY --chmod=755 ${BUILD_CONTEXT_PATH} /app/
```
ritwik-g, line 41: @chandrasekharan-zipstack what is the requirement for this chmod? I don't think it was present earlier.

docker/dockerfiles/x2text.Dockerfile
```diff
@@ -20,25 +20,18 @@ USER unstract
...
-    # Install opentelemetry for instrumentation.
-    pip install opentelemetry-distro opentelemetry-exporter-otlp; \
-    \
-    opentelemetry-bootstrap -a install; \
-    \
-    pdm sync --prod --no-editable; \
-    \
-    # REF: https://docs.gunicorn.org/en/stable/deploy.html#using-virtualenv
-    pip install --no-cache-dir gunicorn;
+COPY --chmod=755 ${BUILD_CONTEXT_PATH} /app/
```
ritwik-g, line 30: @chandrasekharan-zipstack please comment the use of chmod instead of chown.

ritwik-g (commented):
@chandrasekharan-zipstack great initiative and this will help everyone. But one thing we need to check is if this works with our github workflows. Why? In the local machine the builds always happen in same machine and caching of layers can be used effectively. Where as in github runner's case this might not be the case. 

So can you try taking a build and rebuild to confirm if we are seeing the benefit? 

If not no worries we just need to enable caching. But might need to take look in to the same and this might be a good starting point https://docs.docker.com/build/ci/github-actions/cache/

docker/dockerfiles/platform.Dockerfile
chandrasekharan-zipstack: @ritwik-g it was linked to a SonarCloud security issue with providing write privilege to a non-root user `unstract`. Based on its suggestions, I've given read access to the copied code but not write access. 
If we wish to write to any files, we'll have to make sure that we explicitly provide for those directories alone

chandrasekharan-zipstack:
@ritwik-g my main motive was to reduce the time spent on builds on the dev machine. Will look into caching our builds with the github runner as well and maybe raise it as a separate PR if the changes grow in size.

ritwik-g (approved):

docker/dockerfiles/backend.Dockerfile
```diff
@@ -31,28 +31,20 @@ RUN apt-get update; \
 
 WORKDIR /app
 
-COPY ${BUILD_CONTEXT_PATH}/ .
+# Create venv and install gunicorn and other deps in it
+RUN pdm venv create -w virtualenv --with-pip && \
+    . .venv/bin/activate && \
+    pip install --no-cache-dir gunicorn \
+    opentelemetry-distro opentelemetry-exporter-otlp && \
+    opentelemetry-bootstrap -a install
```
hari-kuriakose, line 39: @chandrasekharan-zipstack The following is more readable.
```suggestion
# Create venv and install deps
RUN pdm venv create -w virtualenv --with-pip;  \
    \
    . .venv/bin/activate; \
    \
    pip install --no-cache-dir gunicorn opentelemetry-distro opentelemetry-exporter-otlp; \
    \
    opentelemetry-bootstrap -a install;
```

hari-kuriakose:
@chandrasekharan-zipstack Overall great initiative! Any boost in DX is a huge welcome.

Also CI runner cache can definitely be looked at next as @ritwik-g mentioned.

PR #1129, opened 2025-02-11
feat: UN-1983 BE API changes for file centric logging

chandrasekharan-zipstack:
## What

- Updated execution time in WF and file execution models
- Added data migrations for older records
- MINOR: Storing `enum.value` in file execution table

- From #1131
  - Added new field `total_files` to WF execution model
  - Added `successful_files` and `failed_files` fields to execution serializer
  - Changes to calculate `total_files` before an execution begins
  - MINOR: Checked and changed typing of `workflow_execution` arg in some functions
  - **NOTE:** The `total_files` value will not be filled for existing records but it can be handled with a data migration if required

- From #1132
  - Updated status field for below models to use `TextChoices` correctly
    - WorkflowExecution
    - WorkflowFileExecution
    - FileHistory
  - Updated usage of this enum across `backend`
  - MINOR: Fixed bug with `log_level` for latest_log` while listing files
  - Added retrieve API for executions to obtain by execution ID

## Why

- Previously was not calculated, need it for file centric logging enhancements

## How

- Update execution time whenever the status is one of ERROR / COMPLETED / STOPPED

- No, updated a previously empty value

## Database Migrations

- `WorkflowFileExecution` - data migration to 
  - rename statuses correctly
  - update execution time as `modified_at - created_at`
- `WorkflowExecution` - data migration to 
  - update execution time as `modified_at - created_at`

## Related Issues or PRs

- UN-1983
- Merged PRs into this branch
  - #1131 
  - #1132 

## Notes on Testing

- Tried executing and noticed these values were updated correctly
- Migration ran without issues

## Screenshots
1. After migration with updated times and statuses
image

1. From #1131 
image

## Checklist


jagadeeswaran-zipstack (approved):

backend/workflow_manager/file_execution/models.py
```diff
@@ -116,8 +116,21 @@ def update_status(
...
+        if (
+            status
+            in [
+                ExecutionStatus.COMPLETED,
+                ExecutionStatus.ERROR,
+                ExecutionStatus.STOPPED,
+            ]
+            and not self.execution_time
+        ):
+            self.execution_time = round(
```
muhammad-ali-e, line 130: The execution time here is the total time calculated as (now - created_at). This includes both the queued time and the actual execution time. I believe we should differentiate these two times. Let's decide on this and consider integrating it step by step.

backend/workflow_manager/workflow_v2/execution.py
```diff
@@ -154,19 +155,30 @@ def create_workflow_execution(
...
+            if (
+                status
+                in [
+                    ExecutionStatus.COMPLETED,
+                    ExecutionStatus.ERROR,
+                    ExecutionStatus.STOPPED,
+                ]
+                and not execution.execution_time
+            ):
+                execution.execution_time = round(
```
muhammad-ali-e, line 179: The execution time is defined and calculated in the `execute` method in the `WorkflowExecutionServiceHelper` class (in `execution.py`). I think we could use this for `file_execution`. Similarly, for the **main execution**, we should calculate the execution time for the whole process. However, there also have the same confusion: do we need the full time (from `created_at` to `completed`), or only the actual execution time?

backend/workflow_manager/file_execution/models.py
muhammad-ali-e: But we can use this method and may add a migration if required for the past executions, that doesn't have th execution time.

backend/workflow_manager/workflow_v2/execution.py
chandrasekharan-zipstack: >The execution time is defined and calculated in the execute method in the WorkflowExecutionServiceHelper class (in execution.py). I think we could use this for file_execution

I considered this initially but since we update the status after this, I thought it was a lot cleaner / easier and slightly more accurate to handle the time calculation here.
Also this method of calculation captures time to add the file to the execution volume (can be more if filesystems are remote) so overall it'll be more accurate is what I feel

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -259,11 +260,7 @@ def _process_file(
         except Exception as e:
             error = f"Error processing file '{os.path.basename(input_file)}'. {str(e)}"
             execution_service.publish_log(error, level=LogLevel.ERROR)
-            workflow_file_execution.update_status(
```
chandrasekharan-zipstack, line 262: Removed this since its a redundant update. After capturing the error we update the status [here](https://github.com/Zipstack/unstract/blob/86be2eefa42dece679bb5540912c76036f42e145/backend/workflow_manager/workflow_v2/workflow_helper.py#L180-L183) anyway

muhammad-ali-e (approved):

hari-kuriakose (approved):
@chandrasekharan-zipstack Migrations look good overall.

PR #1131, opened 2025-02-13
feat: Added support for total files and processing status in execution

chandrasekharan-zipstack:
## What

- Added new field `total_files` to WF execution model
- Added `successful_files` and `failed_files` fields to execution serializer
- Changes to calculate `total_files` before an execution begins
- MINOR: Checked and changed typing of `workflow_execution` arg in some functions

NOTE: The `total_files` value will not be filled for existing records but it can be handled with a data migration if required

## Why

- To provide insight on status of an execution in terms of the total files involved and the progress

## How

- Captured total files involved before an execution begins and persisted it
- Computing progress by querying for the file execution records by its status

- No, made changes with sane defaults - all additive

## Database Migrations

- 2 Schema migrations
  - For added field and minor `db_comment` change in WF execution model
  - For `related_name` change in file execution model

## Related Issues or PRs

- Based on top of #1129 

## Notes on Testing

- Tried executing a WF / API deployment locally
- Computing the values as expected for new records

## Screenshots
image

## Checklist


backend/workflow_manager/execution/serializer/execution.py
```diff
@@ -20,3 +23,13 @@ def get_workflow_name(self, obj: WorkflowExecution) -> Optional[str]:
     def get_pipeline_name(self, obj: WorkflowExecution) -> Optional[str]:
         """Fetch the pipeline or API deployment name"""
         return obj.pipeline_name
+
+    def get_successful_files(self, obj: WorkflowExecution) -> int:
+        """Return the count of successfully executed files"""
+        return obj.file_executions.filter(
+            status=ExecutionStatus.COMPLETED.value
```
muhammad-ali-e, line 30: `ExecutionStatus.COMPLETED.value` or `ExecutionStatus.COMPLETED` ?

backend/workflow_manager/execution/serializer/execution.py
```diff
@@ -20,3 +23,13 @@ def get_workflow_name(self, obj: WorkflowExecution) -> Optional[str]:
     def get_pipeline_name(self, obj: WorkflowExecution) -> Optional[str]:
         """Fetch the pipeline or API deployment name"""
         return obj.pipeline_name
+
+    def get_successful_files(self, obj: WorkflowExecution) -> int:
+        """Return the count of successfully executed files"""
```
muhammad-ali-e, line 28: Why do we need these utility  functions? can't we use filterclass instaed?

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -92,7 +92,7 @@ def build_workflow_execution_service(
         single_step: bool,
         scheduled: bool,
         execution_mode: tuple[str, str],
-        workflow_execution: Optional[WorkflowExecution],
+        workflow_execution: WorkflowExecution,
```
muhammad-ali-e, line 95: Do we get the workflow_execution from all referances?

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -340,13 +340,18 @@ def run_workflow(
         destination.validate()
         # Execution Process
         try:
+            input_files, total_files = source.list_files_from_source(
+                hash_values_of_files
+            )
+            workflow_execution.total_files = total_files
```
muhammad-ali-e, line 346: Workflow_file_executions will not be deleted, right? In any case it need to be deleted one of file execution we shoulld update this value. But I don't think there is no chance to dellete it.

backend/workflow_manager/execution/serializer/execution.py
chandrasekharan-zipstack: In the current changes it has to be `ExecutionStatus.COMPLETED.value` - this is because we use a normal enum for the field. We need to inherit the `models.TextChoices` class to use `ExecutionStatus.COMPLETED`.
I've raised [this PR](https://github.com/Zipstack/unstract/pull/1132) to address this and update the code everywhere

backend/workflow_manager/workflow_v2/workflow_helper.py
chandrasekharan-zipstack: Yes @muhammad-ali-e, checked it and then updated the code. I hope the existing typing is valid - I've tested briefly with normal WF execution and an API call

backend/workflow_manager/workflow_v2/workflow_helper.py
chandrasekharan-zipstack: Since this is a record / history of the execution - we wouldn't have to change this value after the execution ends. It simply represents the number of files involved in a run. Even in case of the file execution model - we shouldn't delete it ideally.
The only way these records get deleted is by
1. Partition's retention policy forcing a delete of the old records
1. A new feature that would allow the user to clear executions itself (clearing individual file's execution doesn't make much sense)

muhammad-ali-e (approved):
LGTM

PR #1134, opened 2025-02-15
Add issue template config

hari-kuriakose:
## What

* Add issue template config
* Update readme

## Why

- Disable blank issue creation
- Improve readability

## How

-

No.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


ritwik-g (approved):
@hari-kuriakose I think we also might need to add a bug_report.yaml like [here](https://github.com/Zipstack/helm-values-manager/blob/main/.github/ISSUE_TEMPLATE/bug_report.yml). Else the option to create bug might not work. 

And I believe when we use the bug report form the template might not be needed.

hari-kuriakose:
> @hari-kuriakose I think we also might need to add a bug_report.yaml like [here](https://github.com/Zipstack/helm-values-manager/blob/main/.github/ISSUE_TEMPLATE/bug_report.yml). Else the option to create bug might not work.
> 
> And I believe when we use the bug report form the template might not be needed.

@ritwik-g It will work but currently since only `bug_report.md` is present, it will display as a single textfield with the markdown content added as placeholder. Instead if we have `bug_report.yml` as you mentioned we can define even more textfields etc in the template. We can definitely do this later as we evolve. Proceeding for now.

PR #1136, opened 2025-02-18
Support remote storage for OSS users

gaya3-zipstack:
## What

- Support remote storage for OSS users with docker setup

## Why

- Supporting remote storage for OSS users using docker setup is required to keep Unstract working for the community

## How

- Configure minio as remote storage instead of local container storage
- Use local storage only for costing in Platform service and tool registry in backend

- New user signup to be tested.

## Database Migrations

- NA

## Env Config

- run-platform.sh will take care of necessary env setup

## Relevant Docs

-

## Related Issues or PRs

- https://github.com/Zipstack/unstract-sdk/pull/157

## Dependencies Versions

- 

## Notes on Testing

- Testing was carried out by enabling remote storage feature flag and with this PR specific changes on the mainstream branch

## Screenshots

Tool run
image
 
Prompt studio
image

createbuckets - mirror command - transferring data
image

## Checklist

### DIFF
--- a/backend/sample.env
+++ b/backend/sample.env
@@ -149,14 +149,14 @@ API_EXECUTION_DIR_PREFIX="unstract/api"
 
 # Storage Provider for Workflow Execution
 # Valid options: MINIO, S3, etc..
-WORKFLOW_EXECUTION_FILE_STORAGE_CREDENTIALS='{"provider":"minio","credentials": {"endpoint_url":"http://unstract-minio:9000","key":"XXX","secret":"XXX"}}'
+WORKFLOW_EXECUTION_FILE_STORAGE_CREDENTIALS='{"provider": "minio", "credentials": {"endpoint_url": "http://unstract-minio:9000", "key": "minio", "secret": "minio123"}}'
 
 # Storage Provider for API Execution
-API_FILE_STORAGE_CREDENTIALS='{"provider": "minio", "credentials": {"endpoint_url": "http://unstract-minio:9000", "key": "XXX", "secret": "XXX"}}'
+API_FILE_STORAGE_CREDENTIALS='{"provider": "minio", "credentials": {"endpoint_url": "http://unstract-minio:9000", "key": "minio", "secret": "minio123"}}'
 
 #Remote storage related envs
-PERMANENT_REMOTE_STORAGE={"provider":"gcs","credentials":<credentials>}
-REMOTE_PROMPT_STUDIO_FILE_PATH="<bucket_name>/prompt_studio_data/"
+PERMANENT_REMOTE_STORAGE='{"provider": "minio", "credentials": {"endpoint_url": "http://unstract-minio:9000", "key": "minio", "secret": "minio123"}}'
+REMOTE_PROMPT_STUDIO_FILE_PATH="unstract/prompt-studio-data"
 
 # Storage Provider for Tool registry
 TOOL_REGISTRY_STORAGE_CREDENTIALS='{"provider":"local"}'
--- a/backend/utils/file_storage/helpers/prompt_studio_file_helper.py
+++ b/backend/utils/file_storage/helpers/prompt_studio_file_helper.py
@@ -115,7 +115,7 @@ class PromptStudioFileHelper:
         )
         # TODO : Handle this with proper fix
         # Temporary Hack for frictionless onboarding as the user id will be empty
-        if not fs_instance.exists(file_system_path):
+        if not user_id and not fs_instance.exists(file_system_path):
             file_system_path = (
                 PromptStudioFileHelper.get_or_create_prompt_studio_subdirectory(
                     org_id=org_id,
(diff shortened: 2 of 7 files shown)


platform-service/sample.env
```diff
@@ -34,6 +34,7 @@ MODEL_PRICES_TTL_IN_DAYS=7
 MODEL_PRICES_FILE_PATH="<bucket-name>/cost/model_prices.json"
 
 #Remote storage config
-FILE_STORAGE_CREDENTIALS='{"provider":"gcs","credentials": {"token": {token-value-json}}'
+FILE_STORAGE_CREDENTIALS='{"provider":"local"}'
```
harini-venkataraman, line 37: @gaya3-zipstack Shouldnt platform service point to minio as well? Why are we referencing to local here?

backend/sample.env
```diff
@@ -149,14 +149,14 @@ API_EXECUTION_DIR_PREFIX="unstract/api"
...
+WORKFLOW_EXECUTION_FILE_STORAGE_CREDENTIALS='{"provider": "minio", "credentials": {"endpoint_url": "http://unstract-minio:9000", "key": "minio", "secret": "minio123"}}'
 
 # Storage Provider for API Execution
-API_FILE_STORAGE_CREDENTIALS='{"provider": "minio", "credentials": {"endpoint_url": "http://unstract-minio:9000", "key": "XXX", "secret": "XXX"}}'
+API_FILE_STORAGE_CREDENTIALS='{"provider": "minio", "credentials": {"endpoint_url": "http://unstract-minio:9000", "key": "minio", "secret": "minio123"}}'
 
 #Remote storage related envs
-PERMANENT_REMOTE_STORAGE={"provider":"gcs","credentials":<credentials>}
-REMOTE_PROMPT_STUDIO_FILE_PATH="<bucket_name>/prompt_studio_data/"
+PERMANENT_REMOTE_STORAGE='{"provider": "minio", "credentials": {"endpoint_url": "http://unstract-minio:9000", "key": "minio", "secret": "minio123"}}'
```
harini-venkataraman, line 158: @gaya3-zipstack Do you think, we need to update README also, to convey that files are maintained in minio?
Maybe we can take it up as a separate activity. Adding it just it doesnt get missed.

harini-venkataraman (approved):
Added minor comments, LGTM otherwise

backend/sample.env
ritwik-g: Do we call this out today for the local storage? I don't really think that is necessary. This is an internal implementation detail right?

platform-service/sample.env
gaya3-zipstack: I kept it to local for OSS user here because, if we use minio in platform service for OSS (unlike GCS for cloud), then we need to install s3fs for OSS user icluding it as a dependency. We don;t really need that to be functional. Local storage will work just fine. Hence this decision @harini-venkataraman

backend/utils/file_storage/helpers/prompt_studio_file_helper.py
```diff
@@ -115,7 +115,7 @@ def fetch_file_contents(
         )
         # TODO : Handle this with proper fix
         # Temporary Hack for frictionless onboarding as the user id will be empty
-        if not fs_instance.exists(file_system_path):
+        if not user_id and not fs_instance.exists(file_system_path):
```
ritwik-g, line 118: @gaya3-zipstack why was the additional condition for user_id added here?

backend/utils/file_storage/helpers/prompt_studio_file_helper.py
```diff
@@ -126,7 +126,9 @@ def fetch_file_contents(
             )
         file_path = str(Path(file_system_path) / file_name)
         legacy_file_path = str(Path(legacy_file_system_path) / file_name)
-        file_content_type = fs_instance.mime_type(file_path)
+        file_content_type = fs_instance.mime_type(
+            path=file_path, legacy_storage_path=legacy_file_path
```
ritwik-g, line 130: Will the legacy file path be the volume based path here?

gaya3-zipstack: yes. /app/prompt-studio-data/....

docker/docker-compose.yaml
```diff
@@ -208,6 +211,8 @@ services:
       - /var/run/docker.sock:/var/run/docker.sock
     depends_on:
       - redis
+      - minio
+      - createbuckets
```
ritwik-g, line 215: @gaya3-zipstack will runner be accessing these path? I don't think so. Why would we need these here? As far as I know we are using runner only for tool creation. For which accessing these files shouldn't be required.

platform-service/sample.env
ritwik-g: @gaya3-zipstack what is the use of the FILE_STORAGE_CREDENTIALS in platform service? I couldn't understand why we don't need a remote storage in this case but in cloud. May be we can connect to discuss.

docker/docker-compose.yaml
gaya3-zipstack: I see the env variables being defined for runner also. Not sure if we are simply picking them up to be transmitted to tools. @muhammad-ali-e can you pls confirm?

ritwik-g (approved):

PR #1158, opened 2025-02-28
feat: UN-2166 Passed optional container name from backend to runner

chandrasekharan-zipstack:
## What

- Passed container name from `backend` to `runner`
- MINOR: Renamed `run_id` to `file_execution_id` in parts concerning WF execution

## Why

- Previously container name was logged in the `backend` and a different one was generated and used by the `runner` which resulted in confusion during debugging and tracing
- Ensuring a single container name is logged for a run and using the same throughout

## How

- Passed via the HTTP call to the `runner`

- No, changes made with defaults - tests and and all associated code is updated

## Dependencies Versions

- Added `unstract-core` to `unstract-tool-sandbox`

## Notes on Testing

- Tested a WF run locally and it works against docker

## Screenshots
image

## Checklist

### DIFF
--- a/backend/workflow_manager/workflow_v2/execution.py
+++ b/backend/workflow_manager/workflow_v2/execution.py
@@ -231,9 +231,7 @@ class WorkflowExecutionServiceHelper(WorkflowExecutionService):
             )
             raise WorkflowExecutionError(self.compilation_result["problems"][0])
 
-    def execute(
-        self, file_execution_id: str, file_name: str, single_step: bool = False
-    ) -> None:
+    def execute(self, file_execution_id: str, single_step: bool = False) -> None:
         execution_type = ExecutionType.COMPLETE
         if single_step:
             execution_type = ExecutionType.STEP
@@ -256,7 +254,6 @@ class WorkflowExecutionServiceHelper(WorkflowExecutionService):
         try:
             self.execute_workflow(
                 file_execution_id=file_execution_id,
-                file_name=file_name,
                 execution_type=execution_type,
             )
             end_time = time.time()
@@ -356,7 +353,13 @@ class WorkflowExecutionServiceHelper(WorkflowExecutionService):
         )
         workflow_file_execution.update_status(ExecutionStatus.EXECUTING)
 
-        self.execute(file_execution_id, file_name, single_step)
+        logger.info(
+            f"Running execution: '{self.execution_id}',  "
+            f"file_execution_id: '{file_execution_id}', "
+            f"file '{file_name}'"
+        )
+
+        self.execute(file_execution_id, single_step)
         self.publish_log(f"Tool executed successfully for '{file_name}'")
         self._handle_execution_type(execution_type)
 
--- a/docker/scripts/pdm-lock-gen/pdm-lock.sh
+++ b/docker/scripts/pdm-lock-gen/pdm-lock.sh
@@ -79,7 +79,6 @@ directories=(
     "platform-service"
     "x2text-service"
     "unstract/connectors"
-    "unstract/tool-sandbox"
 )
 
 # If directories are passed as arguments, override the default
--- a/runner/src/unstract/runner/clients/docker.py
+++ b/runner/src/unstract/runner/clients/docker.py
@@ -158,17 +158,24 @@ class Client(ContainerClientInterface):
     def get_container_run_config(
         self,
         command: list[str],
-        run_id: str,
+        file_execution_id: str,
+        container_name: Optional[str] = None,
         envs: Optional[dict[str, Any]] = None,
         auto_remove: bool = False,
     ) -> dict[str, Any]:
         if envs is None:
             envs = {}
         mounts = []
+
+        if not container_name:
+            container_name = UnstractUtils.build_tool_container_name(
+                tool_image=self.image_name,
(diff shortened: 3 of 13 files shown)


runner/src/unstract/runner/clients/docker.py
```diff
@@ -158,17 +158,24 @@ def get_image(self) -> str:
     def get_container_run_config(
         self,
         command: list[str],
-        run_id: str,
+        file_execution_id: str,
+        container_name: Optional[str] = None,
```
muhammad-ali-e, line 162: Where does this optional container_name parameter come from?

unstract/core/src/unstract/core/utilities.py
```diff
@@ -30,12 +30,16 @@ def get_env(env_key: str, default: Optional[str] = None, raise_err=False) -> str
...
+        file_execution_id: str,
+        retry_count: Optional[int] = None,
     ) -> str:
         tool_name = tool_image.split("/")[-1]
-        # TODO: Add execution attempt to better track instead of uuid
-        short_uuid = uuid.uuid4().hex[:6]  # To avoid duplicate name collision
-        container_name = f"{tool_name}-{tool_version}-{short_uuid}-{run_id}"
+        # To avoid duplicate name collision
+        if not retry_count:
+            retry_count = uuid.uuid4().hex[:6]
```
muhammad-ali-e, line 41: It would be better to avoid using the same variable for different meanings

like

if retry_count is None:
    unique_suffix or short_uuid = uuid.uuid4().hex[:6]
else:
    unique_suffix = retry_count

muhammad-ali-e (approved):
LGTM overall, left few NITs

runner/src/unstract/runner/clients/docker.py
chandrasekharan-zipstack: 1. We generate it from the `backend` in the `tool-sandbox` [here](https://github.com/Zipstack/unstract/blob/25563c5f931e864dfe3752b8599693264e525d7c/unstract/tool-sandbox/src/unstract/tool_sandbox/helper.py#L119-L124)
1. We pass it down when we call the `runner` and eventually pass it to this client [here](https://github.com/Zipstack/unstract/blob/25563c5f931e864dfe3752b8599693264e525d7c/runner/src/unstract/runner/runner.py#L205)

docker/scripts/pdm-lock-gen/pdm-lock.sh
```diff
@@ -79,7 +79,6 @@ directories=(
     "platform-service"
     "x2text-service"
     "unstract/connectors"
-    "unstract/tool-sandbox"
```
ritwik-g, line 82: @chandrasekharan-zipstack why is this removed?

backend/workflow_manager/workflow_v2/execution.py
```diff
@@ -231,9 +231,7 @@ def build(self) -> None:
             )
             raise WorkflowExecutionError(self.compilation_result["problems"][0])
 
-    def execute(
-        self, file_execution_id: str, file_name: str, single_step: bool = False
-    ) -> None:
+    def execute(self, file_execution_id: str, single_step: bool = False) -> None:
```
ritwik-g, line 234: @chandrasekharan-zipstack why is this `file_name` removed. And is it related to the container name change?

runner/src/unstract/runner/runner.py
```diff
@@ -217,7 +219,7 @@ def run_container(
         try:
             self.logger.info(
                 f"Execution ID: {execution_id}, running docker "
-                f"container: {container_config.get('name')}"
+                f"container: {container_name}"
```
ritwik-g, line 222: @chandrasekharan-zipstack could we please retain the log as original. That represents the truth. If for some reason what if the container_name and actual one diverges? If the container name is None this wont' show the container name correct?

chandrasekharan-zipstack: You're right @ritwik-g however even with the original change we were getting `None`. This is because the `container_config` contained different dictionary structures for both k8s and docker clients.

I think the actual solution would be to add another method in the clients to store / retrieve these container names from the config for each client

ritwik-g (commented):
I feel like you have combined way too many changes in to this one PR making it a bit difficult to review. If the PR only had the container name related changes it would have been easier to review. Keeping the PR with single responsibility help the reviewers. Especially with some items like `file_name` removal the new retry count logic in the name. With all these changes am finding it difficult to review the PR. 

When we are not hard pressed for time this might be something passable. But now since we are trying to take this change quickly in the other changes are becoming an unnecessary burden. It takes longer to review and the chances for bugs also increases.

Anyway since @muhammad-ali-e has approved I would consider this relatively okay. For some items outside runner I don't have enough context to understand the changes. So I am not thoroughly reviewing those. 

But I have left one comment which needs to be addressed. Please check that

backend/workflow_manager/workflow_v2/execution.py
chandrasekharan-zipstack: @ritwik-g the `file_name` was passed down just to log it along with the execution's IDs. This has been handled in a function that calls this and hence I removed this unused param

ritwik-g (approved):

PR #1179, opened 2025-03-11
fix: Custom markdown component regex fixed for code blocks

chandrasekharan-zipstack:
## What

- Updated FE regex to identify code blocks

## Why

- Helps render error messages without an additional new line char

## How

-

- No

## Notes on Testing

- Tested locally with a markdown based error sent from the BE

## Screenshots
- Before fix
image

- After fix
image

## Checklist

### DIFF
--- a/frontend/src/components/helpers/custom-markdown/CustomMarkdown.jsx
+++ b/frontend/src/components/helpers/custom-markdown/CustomMarkdown.jsx
@@ -21,7 +21,7 @@ const CustomMarkdown = ({
     - New line: \n
   */
   const patterns = [
-    { type: "tripleCode", regex: /```([^`]{1,1000})```/g },
+    { type: "tripleCode", regex: /```\n([\s\S]*?)\n```/g },
     { type: "inlineCode", regex: /`([^`]{1,100})`/g },
     { type: "bold", regex: /\*\*([^*]{1,100})\*\*/g },
     { type: "link", regex: /\[([^\]]{1,100})\]\(([^)]{1,200})\)/g },


frontend/src/components/helpers/custom-markdown/CustomMarkdown.jsx
```diff
@@ -21,7 +21,7 @@ const CustomMarkdown = ({
     - New line: \n
   */
   const patterns = [
-    { type: "tripleCode", regex: /```([^`]{1,1000})```/g },
+    { type: "tripleCode", regex: /```\n([\s\S]*?)\n```/g },
```
tahierhussain, line 24: @chandrasekharan-zipstack This way, we will check for ` ```\n` instead of ` ``` `. As a result, if a code block is passed from the backend without a newline (`\n`), it will not be recognized as a code block. Can you please check this?

tahierhussain (changes requested):

frontend/src/components/helpers/custom-markdown/CustomMarkdown.jsx
jagadeeswaran-zipstack: @chandrasekharan-zipstack @tahierhussain Better approach would be get the code block in ` ``` ` and then trim the `\n` from start and end. What do you think about this approach?

chandrasekharan-zipstack: @tahierhussain @jagadeeswaran-zipstack the intention is to pass a markdown codeblock from the backend. In all scenarios the codeblock is represented with **triple backticks and a new line character.** [The ideal representation](https://www.markdownguide.org/extended-syntax/#fenced-code-blocks) would also allow `\w+` (for specifying language). 
If for some reason we only pass ```<some code block text>``` from the BE - then its a bug in the BE that needs to be corrected. No markdown processor will render that as a codeblock.

tahierhussain (approved):

jagadeeswaran-zipstack (approved):

PR #1181, opened 2025-03-12
feat: Data migration to update total_files column of WorkflowExecution

chandrasekharan-zipstack:
## What

- Data migration to update `total_files` in `WorfklowExecution` model
- Reverse migration also present

## Why

- This is done to handle the count for older executions that didn't have a value for this column
- For correctness in the file-centric-logging feature - will not cause a logical issue however its needed for the sake of representation

## How

- Counted the number of `WorkflowFileExecution` instances associated
- During reverse migration - we simply set the value to 0

- No, however if the reverse migration is applied it can reset the count to 0 for records which were not previously affected in the forward pass.
- This shouldn't be a concern since the `total_files` is a field used by a new feature and the forward pass should take care of updating it correctly

## Database Migrations

- Yes - a data migration is involved

## Related Issues or PRs

- UN-1983
- Merged #1185  into this

## Notes on Testing

- Tested migration locally

## Screenshots
1. Before migration in local DB
image
1. After migration
image

## Checklist


hari-kuriakose (approved):

backend/workflow_manager/workflow_v2/migrations/0009_update_total_files.py
```diff
@@ -0,0 +1,41 @@
...
+        execution.save(update_fields=["total_files"])
+
+
+def reverse_update_total_files(apps: Any, schema_editor: Any) -> None:
+    """Reverse the update by setting total_files back to zero for all file executions"""
+    WorkflowExecution = apps.get_model(  # pylint: disable=invalid-name # NOSONAR
+        "workflow_v2", "WorkflowExecution"
+    )
+
+    WorkflowExecution.objects.update(total_files=0)
```
muhammad-ali-e, line 30: If it a new column, and the default is `0` can't we make it in model with default value as `0` ?

muhammad-ali-e (approved):
LGTM the code,
I believe the WorkflowExecution table will contain a large number of entries. To avoid an extensive migration, we could allow existing records to retain a default value of zero instead of modifying them. Let me know your thoughts.

chandrasekharan-zipstack:
@muhammad-ali-e I was curious about the same but it'll show a 0 for older records. We could correct it lazily as well however since the default is 0 and not `NULL` I noticed that it runs wastefully a lot of times for error-ed executions.
Took call to take this one time cost to fix the values correctly

PR #1182, opened 2025-03-12
Feat/prompt studio ux improvements

athul-rs:
## What

- Adding modal info if no prompt-studio tool is exported and available for workflows, pipelines and API deployments. 

## Why

- Better UX

## How

- Checking for prompt studio tools.

- YES

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Screenshots

image

## Checklist


frontend/src/components/api/prompt-studio-service.js
```diff
@@ -0,0 +1,24 @@
...
+export const promptStudioService = () => {
+  const getPromptStudioCount = async () => {
+    const { sessionDetails } = useSessionStore.getState();
+    if (!sessionDetails) {
+      throw new Error("Session details are missing");
+    }
+
+    const requestOptions = {
+      method: "GET",
+      url: `/api/v1/unstract/${sessionDetails.orgId}/tool/prompt-studio/`,
```
tahierhussain, line 13: @athul-rs Can you make use of this [hook](https://github.com/Zipstack/unstract/blob/main/frontend/src/hooks/useRequestUrl.js) to get the base URL?

frontend/src/components/workflows/workflow/Workflows.jsx
```diff
@@ -189,8 +203,21 @@ function Workflows() {
...
+  useEffect(() => {
+    if (!isLoading && count === 0 && !modalDismissed) {
+      setShowModal(true);
+    }
+  }, [isLoading, count, modalDismissed]);
+
+  const handleModalClose = () => {
+    setShowModal(false);
+    setModalDismissed(true);
+    fetchPromptCount();
```
tahierhussain, line 215: @athul-rs Why are we calling the fetchPromptCount function again?

Here's the flow as I understand it: When this component is mounted, we call fetchPromptCount to retrieve the count and determine whether to display the modal. If the count is 0, the modal is displayed. Then, when the modal closes, we call the same function again. Is this necessary?

Note: It looks like we've followed the same approach in other components as well. If we decide to change the approach, please ensure the necessary updates are made across all relevant components.

frontend/src/components/workflows/workflow/Workflows.jsx
```diff
@@ -32,6 +35,17 @@ function Workflows() {
   const projectApiService = workflowService();
   const handleException = useExceptionHandler();
   const { setPostHogCustomEvent } = usePostHogEvents();
+  const { count, isLoading, fetchCount } = usePromptStudioStore();
+  const [showModal, setShowModal] = useState(false);
+  const [modalDismissed, setModalDismissed] = useState(false);
+
+  const fetchPromptCount = useCallback(() => {
+    fetchCount();
+  }, [fetchCount]);
```
tahierhussain, line 44: @athul-rs Can we directly call `fetchCount` wherever we're calling `fetchPromptCount` and remove this function?

tahierhussain (changes requested):

backend/tool_instance_v2/views.py
```diff
@@ -61,6 +61,26 @@ def get_tool_list(request: Request) -> Response:
...
+        try:
+            logger.info("Fetching prompt studio tool count from the tool registry.")
+            tool_count = ToolProcessor.get_prompt_studio_tool_count(request.user)
+            return Response(
+                data={"count": tool_count},
+                status=status.HTTP_200_OK,
+            )
+        except Exception as exc:
+            logger.error(f"Failed to fetch prompt studio tools: {exc}")
+            raise FetchToolListFailed
```
chandrasekharan-zipstack, line 81: NIT: What would this message read as in case of an error?

backend/tool_instance_v2/tool_processor.py
```diff
@@ -127,3 +127,11 @@ def get_tool_list(user: User) -> list[dict[str, Any]]:
         tool_list: list[dict[str, Any]] = tool_registry.fetch_tools_descriptions()
         tool_list = tool_list + prompt_studio_tools
         return tool_list
+
+    @staticmethod
+    def get_prompt_studio_tool_count(user: User) -> int:
+        """Get count of valid prompt studio tools."""
+        tool_count: int = PromptStudioRegistryHelper.fetch_prompt_studio_tool_count(
```
chandrasekharan-zipstack, line 134: NIT: If we aren't doing anything here, we could directly call the underlying function at the caller and avoid this additional function here

athul-rs: Knew this was a unnecessary call, but did it anyway to make the calls uniform, there is a good chance this function might be extended further soon.

backend/tool_instance_v2/views.py
athul-rs:
```
    status_code = 400
    default_detail = "Failed to fetch tool list."
```

athul-rs:
@tahierhussain Addressed all comments

frontend/src/components/api/prompt-studio-service.js
athul-rs: Moved

athul-rs: Moved

frontend/src/components/workflows/workflow/Workflows.jsx
athul-rs: Modified, forgot to remove the earlier code.

frontend/src/components/workflows/workflow/Workflows.jsx
athul-rs: Modified

backend/tool_instance_v2/tool_processor.py
chandrasekharan-zipstack: @athul-rs I doubt it man, we'd end up leaving such code behind only to refactor it later. I suggest avoiding it right away. Will let you take a call

backend/tool_instance_v2/urls.py
```diff
@@ -42,5 +42,10 @@
             name="tool_instance_reorder",
         ),
         path("tool/", views.get_tool_list, name="tool_list"),
+        path(
+            "tool/prompt-studio/",
```
chandrasekharan-zipstack, line 46: NIT: Its not clear from the API name that we're trying to obtain a count of the entity. Change it in the FE as well
```suggestion
            "tool/prompt-studio/count/",
```

chandrasekharan-zipstack (approved):
LGTM - left a minor comment about the API name

frontend/src/components/common/PromptStudioModal.jsx
```diff
@@ -0,0 +1,78 @@
...
+  const { count, fetchCount } = usePromptStudioStore();
+
+  useEffect(() => {
+    fetchCount();
+  }, [fetchCount]);
+
+  const handleClose = () => {
+    if (onClose) onClose();
+    fetchCount();
+  };
```
tahierhussain, line 21: @athul-rs Is this necessary here? We're already calling the `fetchCount` function before opening the modal in its parent component, and we're again calling the same function here. This will result in multiple API calls.

athul-rs: Updated, removed fecthcount from Modal.

vishnuszipstack (approved):

backend/tool_instance_v2/tool_processor.py
athul-rs: One other thought was because of the import needed in views, since all calls to propmptstudioservice is now handled from ToolHelper file, kept it to make it uniform. Thoughts @chandrasekharan-zipstack

chandrasekharan-zipstack: I don't think we should make the system unnecessarily complex. I don't think we have any convention in place regarding such imports and frankly its all static methods under a class name so it might not matter much. As long as there aren't any circular import errors we're good to go

athul-rs: Removed unwanted function call. @chandrasekharan-zipstack

backend/tool_instance_v2/views.py
```diff
@@ -61,6 +61,26 @@ def get_tool_list(request: Request) -> Response:
             raise FetchToolListFailed
 
 
+@api_view(["GET"])  # NOSONAR: Only GET is allowed, and this is safe.
```
chandrasekharan-zipstack, line 64:
```suggestion
# Only GET is allowed, and this is safe.
@api_view(["GET"])  # NOSONAR
```

backend/tool_instance_v2/views.py
muhammad-ali-e: Would be better if we provide specific excpetion instead of  handling global exceptiion here. It will confuse about actual error at the end user.

backend/tool_instance_v2/views.py
```diff
@@ -61,6 +61,27 @@ def get_tool_list(request: Request) -> Response:
...
+@api_view(["GET"])
+def prompt_studio_tool_count(request: Request) -> Response:  # NOSONAR
+    """Get count of prompt studio tools.
+
+    Returns count of valid prompt studio tools available in the Tool registry.
+    Only counts tools that have UUID as function names.
+    """
+    if request.method == "GET":
+        try:
+            logger.info("Fetching prompt studio tool count from the tool.")
```
muhammad-ali-e, line 74: Generic logs without specific contex (such as `user ID` or `tool ID` etc.. )are not very useful.

backend/tool_instance_v2/tool_processor.py
```diff
@@ -127,3 +129,19 @@ def get_tool_list(user: User) -> list[dict[str, Any]]:
...
+    @staticmethod
+    def get_prompt_studio_tool_count(user: User) -> int:
+        """Get count of valid prompt studio tools."""
+        # Filter the Prompt studio registry based on the users.
+        prompt_studio_tools = PromptStudioRegistry.objects.list_tools(user)
+        valid_tools = 0
+
+        for tool in prompt_studio_tools:
+            try:
+                uuid.UUID(str(tool.prompt_registry_id))
```
muhammad-ali-e, line 142: Why doe we check/validate UUID here? IS it not a proper UUID ?

muhammad-ali-e (commented):
Added minor comments—please revisit them later if this needs to be merged quickly.

PR #1187, opened 2025-03-13
feat: Modified execution log to use foreign key for execution

chandrasekharan-zipstack:
## What

- Added `wf_execution` foreign key field
- Deprecated `execution_id` field - will be NULL from now on
- Changes to allow viewing logs with both `wf_execution_id` and `execution_id`

## Why

- Foreign key helps with indexing the data
- Easier access given a workflow execution instance

## How

- Schema migration to add the foreign key field
- ~~Data migration to make the fk relationships on old records~~
- ~~Schema migration to remove the old `execution_id` column~~

- No, only added a column

## Database Migrations

- Only a schema migration involved

## Related Issues or PRs

- Changes for file-centric logging (indexing allows faster access)
- On top of UN-1983

## Notes on Testing

- Tried the forward and reverse migration locally
- Able to run an API deployment after the migration and view logs for older data

## Screenshots (don't apply in the latest changes)
1. Before migration
image
1. After migration
image

## Checklist


hari-kuriakose (commented):
@chandrasekharan-zipstack Migrations look good overall.

backend/workflow_manager/workflow_v2/execution_log_utils.py
```diff
@@ -10,44 +11,74 @@
...
-from workflow_manager.workflow_v2.models.execution_log import ExecutionLog
+from workflow_manager.workflow_v2.models import ExecutionLog, WorkflowExecution
 
 logger = logging.getLogger(__name__)
 
 
-@shared_task(bind=True, name=ExecutionLogConstants.TASK_V2)
-def consume_log_history(self):
+@shared_task(name=ExecutionLogConstants.TASK_V2)
+def consume_log_history() -> None:
```
muhammad-ali-e, line 20: NIT: It looks like there are multiple iterations (loops) over the same data. Can we optimize this by consolidating them into a single loop with conditions inside to improve efficiency?

backend/workflow_manager/workflow_v2/migrations/0012_update_wf_execution_fk_for_logs.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Uses UUID from execution_id to make the FK relationships in wf_execution_id.
+    """
+    ExecutionLog = apps.get_model(  # pylint: disable=invalid-name # NOSONAR
+        "workflow_v2", "ExecutionLog"
+    )
+    WorkflowExecution = apps.get_model(  # pylint: disable=invalid-name # NOSONAR
+        "workflow_v2", "WorkflowExecution"
+    )
+    logs = ExecutionLog.objects.all()
```
muhammad-ali-e, line 22: Hope this will not load all, right?

backend/workflow_manager/workflow_v2/migrations/0012_update_wf_execution_fk_for_logs.py
```diff
@@ -0,0 +1,59 @@
...
+
+    for log in logs:
```
muhammad-ali-e, line 24: Can we avoid applying this migration to existing logs by setting the new field to NULL? There might be a huge number of logs for some organizations. Alternatively, should we consider applying this change only from a certain date onward or set a limit to reduce the impact?

cc: @hari-kuriakose

backend/workflow_manager/workflow_v2/models/execution_log.py
```diff
@@ -3,13 +3,19 @@
 from django.db import models
 from utils.models.base_model import BaseModel
 from workflow_manager.file_execution.models import WorkflowFileExecution
+from workflow_manager.workflow_v2.models import WorkflowExecution
 
 
 class ExecutionLog(BaseModel):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
-    execution_id = models.UUIDField(
+    wf_execution = models.ForeignKey(
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: @muhammad-ali-e currently those prompt studio logs are not being persisted in this table. 
- For logs from `prompt-service` this can be done first
- For prompt studio session's logs we need a separate table to track its executions and we can have a separate column to track its logs

backend/workflow_manager/workflow_v2/migrations/0012_update_wf_execution_fk_for_logs.py
chandrasekharan-zipstack: @muhammad-ali-e this would query for all - but I assume that the django ORM will ensure that it'll be a buffered read when its executed if memory is your concern here

backend/workflow_manager/workflow_v2/migrations/0012_update_wf_execution_fk_for_logs.py
chandrasekharan-zipstack: @muhammad-ali-e this might cause breaking changes / logic to handle older records. Also we need to index the entire column at some point which will involve this iteration over all records anyway.
If its a concern, we could look into clearing some of the older logs for these orgs because I doubt how usable / actionable these logs are for the user

hari-kuriakose: @muhammad-ali-e  Very good call.

@chandrasekharan-zipstack We cannot rely on deleting the old records, as that might not be practical in all deployment envs. And the logic seems to have a one-on-one mapping hence batch update would not be possible too.

Hence this is what we can do:
- Batch update records older than a month to NULL value for this field
- Update records up to 30 days using same logic
- Batch update all records to NULL value for this field for reverse migration

chandrasekharan-zipstack: @hari-kuriakose I see that we intend to keep last 30 days of execution's logs intact with the proposed change.

However I wish to point out that as a side effect, for executions older than 30 days - we would lose the ability to view its logs (which is as good as deleting the older logs since they become stray records in the DB that can't be queried via the UI)

backend/workflow_manager/workflow_v2/execution_log_utils.py
chandrasekharan-zipstack: @muhammad-ali-e the first loop is to identify the list of `execution_id` and `file_execution_id` so that it can be queried in one shot. Otherwise we might have to fire separate queries for each ID

backend/workflow_manager/workflow_v2/migrations/0012_update_wf_execution_fk_for_logs.py
chandrasekharan-zipstack: Also for the reverse migration - we had a non NULL column `execution_id` previously which throws an integrity error if we try to set NULL for records older than 30 days

hari-kuriakose: @chandrasekharan-zipstack Was assuming it was a new column. However if it is an existing column then yes making them orphaned is not preferable. 

That being said, deleting old records would not be in our control for on-prem deployments. Hence can we have a new column instead? We would not need to worry about migrating existing columns in that case at all.

chandrasekharan-zipstack: @hari-kuriakose I'm not sure I understood your comment clearly - here we have an existing UUID column `execution_id`. I've created a new foreign key column called `wf_execution` to maintain this relationship.
This data migration is to move existing UUID values into this new column and remove the `execution_id` UUID column altogether

hari-kuriakose: @chandrasekharan-zipstack Ah ok.
Still let's try to avoid data migration here due to the huge volume of logs.

Instead ensure the app logic to be backward compatible.
If the new column exists, we will use that for the foreign key relation. Otherwise we will use old column. That's all.

chandrasekharan-zipstack: Thanks for your inputs guys, given the high volume we can avoid this data migration

backend/workflow_manager/workflow_v2/models/execution_log.py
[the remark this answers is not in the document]
muhammad-ali-e: oh. okay ,I thikn we discussed it earlier. I thoiught it is usage tablle..

muhammad-ali-e (approved):

PR #1203, opened 2025-03-20
fix: Corrected timestamp reported for logs

chandrasekharan-zipstack:
## What

- Corrected timestamp that is reported for logs
- MINOR: log statement to runner to mark tool completion

## Why

- Current timestamp from when the log was streamed is used. Does not reflect the actual time when the log / event occurs from the tool's POV
- No log about runner's successful completion present

- No, minor changes only

## Notes on Testing

- Tested locally by checking for the timestamp reported and the logs collected

## Screenshots
image

## Checklist

### DIFF
--- a/runner/src/unstract/runner/runner.py
+++ b/runner/src/unstract/runner/runner.py
@@ -82,8 +82,8 @@ class UnstractRunner:
         log_dict = self.get_valid_log_message(log_message)
         if not log_dict:
             return None
-        log_type = self.get_log_type(log_dict)
-        log_level = self.get_log_level(log_dict)
+        log_type = log_dict.get("type")
+        log_level = log_dict.get("level")
         if log_type == LogType.LOG and log_level == LogLevel.ERROR:
             raise ToolRunException(log_dict.get("log"))
         if not self.is_valid_log_type(log_type):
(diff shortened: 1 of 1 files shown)


runner/src/unstract/runner/runner.py
```diff
@@ -82,8 +82,8 @@ def process_log_message(
         log_dict = self.get_valid_log_message(log_message)
         if not log_dict:
             return None
-        log_type = self.get_log_type(log_dict)
-        log_level = self.get_log_level(log_dict)
+        log_type = log_dict.get("type")
```
muhammad-ali-e, line 85: Why it changed? I think some time we will get  logs (here `log_dict`) not as a dict, but like as text. can you check this.

muhammad-ali-e: I don't remember the cases

chandrasekharan-zipstack: @muhammad-ali-e if you take a look at the previous function `self.get_valid_log_message()` we already perform this check to ensure only dict comes through. We were doing some redundant checks here and hence I removed those. It seems to work fine without those unnecessary functions

chandrasekharan-zipstack:
@ritwik-g I checked out this lib previously and it seems to be abandoned. Will raise a separate PR to improve the logging and add a request ID as well. Meanwhile these minor changes can go in

pk-zipstack (approved):

PR #1209, opened 2025-03-21
feat: Improved logging for runner

chandrasekharan-zipstack:
## What

- Refactored `runner`
- Moved app initialization to `__init__.py`
- Used newly added utils in core to setup logger and error handling
- Added request ID middleware to support tracing requests
- Log PID and thread ID in a consistent fashion similar to django services

## Why

- Logging updated for tracing and consistency
- Refactored app initialization to ensure we create the app context only once

## How

- Made use of a simple request ID middleware implementation from core
- Logging configured through a common dict placed in core - other services can be configured similarly later

- Yes since its a refactor but the docker implementation has been tested locally

## Related Issues or PRs

- #1208
- Reviewed and Merged #1210 

## Dependencies Versions

- Added core as an editable dependency to develop with it easily

## Notes on Testing

- Tested changes locally

## Screenshots
1. Run as a flask server
image
1. Run with gunicorn
image
1. Serving tool requests
image

## Checklist

### DIFF
--- a/runner/pyproject.toml
+++ b/runner/pyproject.toml
@@ -7,10 +7,10 @@ authors = [
 ]
 dependencies = [
     "docker==6.1.3",
-    "flask~=3.0.0",
-    "python-dotenv==1.0.0",
+    "flask~=3.1.0",
+    "python-dotenv>=1.0.0",
     "redis~=5.2.1",
-    "unstract-core @ file:///${PROJECT_ROOT}/../unstract/core",
+    "unstract-core[flask] @ file:///${PROJECT_ROOT}/../unstract/core",
     "unstract-flags @ file:///${PROJECT_ROOT}/../unstract/flags",
 ]
 requires-python = ">=3.9,<3.11.1"
@@ -30,11 +30,14 @@ deploy = [
     "gunicorn[gevent]>=21.2.0",
 ]
 
-dev = [
+test = [
     "pytest>=8.2.2",
     "pytest-mock>=3.14.0",
     "pytest-cov>=5.0.0",
     "pytest-md-report>=0.6.2",
+]
+dev = [
+    "-e unstract-core[flask] @ file:///${PROJECT_ROOT}/../unstract/core",
 ]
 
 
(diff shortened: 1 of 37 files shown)


chandrasekharan-zipstack:
@ritwik-g @hari-kuriakose @muhammad-ali-e @Deepak-Kesavan do you guys think we can ignore this security hotspot and address it separately for all the Flask based services? I wouldn't want to block this change on that aspect

runner/pyproject.toml
```diff
@@ -44,7 +47,7 @@ runner.cmd = "./entrypoint.sh"
 runner.env_file = ".env"
 runner.help = "Runs the Unstract tool runner (Gunicorn)"
 # Runs the service with flask
-runner-flask.cmd = "flask --app src/unstract/runner/main.py run --port 5002"
+runner-flask.cmd = "flask --app src/unstract/runner/__init__.py run --port 5002"
```
Deepak-Kesavan, line 50: Instead if doing this, we can use flask envs. `FLASK_APP`, `FLASK_RUN_HOST`, `FLASK_RUN_PORT` etc. 

After doing this we can run flask app with just `flask run` command.

runner/src/unstract/runner/controller/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .health import health_bp  # noqa: F401
```
Deepak-Kesavan, line 1: @chandrasekharan-zipstack We can wrap all blueprint into a single one and register it only once.

Ref - https://github.com/Zipstack/unstract-api-hub/blob/917d1723dd8a23617fe337b0729ef93c75f662f9/backend/app/controllers/__init__.py#L10-L16

runner/src/unstract/runner/routes.py
```diff
@@ -0,0 +1,8 @@
+from flask import Flask
+
+from .controller import health_bp, run_bp
+
+
+def register_blueprints(app: Flask):
+    app.register_blueprint(blueprint=health_bp, url_prefix="/v1/api")
```
Deepak-Kesavan, line 7: url_prefix can be made into an ENV instead of hardcoding. - ref - https://github.com/Zipstack/unstract-api-hub/blob/917d1723dd8a23617fe337b0729ef93c75f662f9/backend/sample.env#L3

chandrasekharan-zipstack: Usually URL prefix would be used to support different versions of the API right. In that case how will an ENV help?
I don't see a reason why an env will help here? If we're going to set and not change it, it might as well be a part of the code

Deepak-Kesavan (approved):

muhammad-ali-e:
> @ritwik-g @hari-kuriakose @muhammad-ali-e @Deepak-Kesavan do you guys think we can ignore this security hotspot and address it separately for all the Flask based services? I wouldn't want to block this change on that aspect

- We can do it later, as it be there before. And Also these are internal services that get called internally.

PR #1233, opened 2025-04-09
[FEAT] Include metadata and addtional columns for ETL destination table & Migration of Single column data from 'String' to 'JSON'

pk-zipstack:
## What

- Added `metadata` column and additional columns to our ETL destination table.
- Added migration of the single column data from `string` to `json`

## Why

- Previously users were able to access the minimal amount of metadata, especially cost. So a metadata column was included.
- Migration to json data type columns because it will enable the users to access the data faster and easier.

## How

-

- No, this PR cannot break any existing features because the migration code functions in way to only add columns. Even if the users use a previous version of the code after migrating, it only stores data is the available column.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

- https://zipstack.atlassian.net/browse/UN-1648

## Dependencies Versions

-

## Notes on Testing

- Tested the table migration and storing data as json in available databases. 

## Checklist


backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -538,6 +571,27 @@ def get_metadata(
 
         return metadata
 
+    def get_combined_metadata(self) -> dict[str, Any]:
```
chandrasekharan-zipstack, line 574: NIT: @pk-zipstack I suggest adding a version parameter to this metadata. This will help us release any changes to the metadata structure and deprecate older versions as well. Please get this vetted with @hari-kuriakose @muhammad-ali-e and other stakeholders

unstract/connectors/src/unstract/connectors/databases/bigquery/bigquery.py
```diff
@@ -111,6 +116,21 @@ def get_create_table_base_query(self, table: str) -> str:
             f"CREATE TABLE IF NOT EXISTS {table} "
             f"(id string,"
             f"created_by string, created_at TIMESTAMP, "
+            f"metadata JSON, "
+            f"user_field_1 BOOL DEFAULT FALSE, "
+            f"user_field_2 INT64 DEFAULT 0, "
+            f"user_field_3 STRING DEFAULT NULL, "
```
chandrasekharan-zipstack, line 122: @pk-zipstack what are these fields and why are they named so?

unstract/connectors/src/unstract/connectors/databases/redshift/redshift.py
```diff
@@ -66,18 +69,35 @@ def get_engine(self) -> connection:
...
             f"(id VARCHAR(65535) ,"
             f"created_by VARCHAR(65535), created_at TIMESTAMP, "
+            f"metadata VARCHAR(65535), "
+            f"user_field_1 BOOLEAN DEFAULT FALSE, "
+            f"user_field_2 INTEGER DEFAULT 0, "
+            f"user_field_3 VARCHAR(65535) DEFAULT NULL, "
+        )
+        return sql_query
+
+    def migrate_table_to_v2_query(self, table_name: str, column_name: str) -> str:
```
chandrasekharan-zipstack, line 93: NIT: Either rename to make it more explicit or include a docstring in the base class method. I suggest `migrate_to_multi_column_table` or something along those lines

chandrasekharan-zipstack (commented):
1. When do we name a column as `column_v2`? Can we opt a better name / will it always remain with the suffix of `_v2`?
1. When does the migration of single column data happen?

backend/workflow_manager/endpoint_v2/database_utils.py
```diff
@@ -145,18 +150,35 @@ def get_columns_and_values(
...
+        if metadata:
+            values[TableColumns.METADATA] = json.dumps(metadata)
+
         if column_mode == ColumnModes.WRITE_JSON_TO_A_SINGLE_COLUMN:
             if isinstance(data, str):
-                values[single_column_name] = data
+                wrapped_dict = {"result": data}
+                values[single_column_name] = wrapped_dict
+                if table_info and f"{single_column_name}_v2" in table_info:
+                    values[f"{single_column_name}_v2"] = wrapped_dict
```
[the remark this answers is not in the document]
pk-zipstack: @chandrasekharan-zipstack It is a v2_column because we wanted to change the `single_column_name` to `json` instead of `string`. So, a new column is created with the name `column_v2`, which will be easier for the user to identify. 

This only happens for an existing destination table. On newer tables, the `single_column_name` column will be created as a `JSON` datatype by default.

unstract/connectors/src/unstract/connectors/databases/bigquery/bigquery.py
pk-zipstack: These columns are for the user to store related values. You can check out more on this [ticket](https://zipstack.atlassian.net/browse/UN-1648).

backend/workflow_manager/endpoint_v2/destination.py
chandrasekharan-zipstack: @muhammad-ali-e since this is metadata that we write to the DB - this is equivalent to an API deployment's response. Learning from our experience with API deployments, I felt its better to include a `version` field which can help clients of the DB data to better handle / parse our responses. Also this version refers to the metadata's fields - in such a scenario we cannot rely on the SDK version to help convey the structure of the response

chandrasekharan-zipstack (approved):

backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -215,22 +218,30 @@ def handle_output(
...
         elif connection_type == WorkflowEndpoint.ConnectionType.DATABASE:
-            if not self._should_handle_hitl(
-                file_name=file_name,
-                file_hash=file_hash,
-                workflow=workflow,
-                input_file_path=input_file_path,
-                file_execution_id=file_execution_id,
-            ):
-                self.insert_into_db(input_file_path=input_file_path)
+            # For error cases, skip HITL and directly insert error record
```
muhammad-ali-e, line 231: Why changed here, Will this affect HITL logic and expectations?

praveen-formido: performing hitl only if there are no errors @muhammad-ali-e. If there is an error, we skip hitl and directly insert the error in the db.

muhammad-ali-e: The HITL is to review efore DB insetions, SO if it is an error why do we need to insert that error in DB? Is ithis a requirement?

praveen-formido: Yes @muhammad-ali-e it is a [requirement](https://zipstack.atlassian.net/browse/UN-1648). since recording error entries gives customers complete visibility into all processing attempts, helps them identify and analyze error patterns, ensures a clear audit trail of both successes and failures, ultimately making troubleshooting easier and giving them a single source of truth for their data processing history.

muhammad-ali-e: @pk-zipstack That Understood. My concirn is when it enabled HITL.

muhammad-ali-e (approved):

backend/workflow_manager/endpoint_v2/database_utils.py
```diff
@@ -53,14 +55,26 @@ def get_sql_values_for_query(
                 col = column.lower()
                 type_x = column_types.get(col, "")
                 if type_x == "VARIANT":
-                    values[column] = values[column].replace("'", "\\'")
-                    sql_values[column] = f"parse_json($${values[column]}$$)"
+                    payload = values[column]
+                    if isinstance(payload, Enum):
+                        payload = payload.value
+                    # Ensure valid JSON for VARIANT
+                    payload_json = json.dumps(payload, default=str)
```
chandrasekharan-zipstack: NIT: What if this isn't a valid JSON? How do we report and handle this error?

backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -342,43 +358,68 @@ def insert_into_db(self, input_file_path: str) -> None:
...
+        if table_info:
+            is_string = db_class.is_string_column(
+                table_info=table_info, column_name=single_column_name
+            )
+            if is_string:
+                db_class.migrate_table_to_v2(
+                    table_name=table_name,
+                    column_name=single_column_name,
+                    engine=engine,
+                )
```
chandrasekharan-zipstack:
```suggestion
            if db_class.is_string_column(
                table_info=table_info, column_name=single_column_name
            ):
                db_class.migrate_table_to_v2(
                    table_name=table_name,
                    column_name=single_column_name,
                    engine=engine,
                )
```

backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -558,6 +599,27 @@ def get_metadata(
...
+        """
+        # Get workflow metadata
+        workflow_metadata = self.get_metadata()
+
+        # Get file_execution_id from metadata
+        file_execution_id = workflow_metadata.get("file_execution_id")
+        if not file_execution_id:
+            return workflow_metadata
+
+        usage_metadata = UsageHelper.get_aggregated_token_count(file_execution_id)
```
chandrasekharan-zipstack, line 622: NIT: Instead of computing this everytime and running extra queries, we should maybe store this in the DB in an appropriate location. Feel free to create a ticket for this enhancement if you think its valid

backend/workflow_manager/endpoint_v2/enums.py
```diff
@@ -1,6 +1,11 @@
 from enum import Enum
 
 
+class FileStatus(Enum):
```
chandrasekharan-zipstack: @pk-zipstack what does this status column represent? Can you be more explicit with the name for this? It's misleading given how we have another enum which tracks the execution status

backend/workflow_manager/endpoint_v2/constants.py
```diff
@@ -5,7 +5,22 @@
 class TableColumns:
     CREATED_BY = "created_by"
     CREATED_AT = "created_at"
-    PERMANENT_COLUMNS = ["created_by", "created_at"]
+    METADATA = "metadata"
+    ERROR_MESSAGE = "error_message"
+    STATUS = "status"
+    USER_FIELD_1 = "user_field_1"
+    USER_FIELD_2 = "user_field_2"
+    USER_FIELD_3 = "user_field_3"
```
chandrasekharan-zipstack, line 13: @pk-zipstack what are these user field columns? What do they do?

backend/workflow_manager/endpoint_v2/enums.py
praveen-formido: @chandrasekharan-zipstack This column represents whether the file itself has been successfully processed (SUCCESS/ERROR). To avoid confusion with ExecutionStatus, I’ve renamed the enum to FileProcessingStatus

backend/workflow_manager/endpoint_v2/constants.py
praveen-formido: @chandrasekharan-zipstack These are generic, placeholder columns provided for flexibility. They can be used by users to store any additional/custom information relevant to their use case.

backend/workflow_manager/endpoint_v2/database_utils.py
praveen-formido: Added a fallback json whenever there is an issue with handling the json in database_utils.py.

hari-kuriakose:
@pk-zipstack It is relatively ok to allow a bit of duplication in this case, as each connector handles datatypes like JSONB etc, differently. Thus it is probably less load cognitively to have all connector code in the same file in the long run.

Let's see how it evolves at which point we can move common logic upstream to base `UnstractDB` class.

PR #1226, opened 2025-04-04
fix: Added workflow as fk in execution table, filtered by users

chandrasekharan-zipstack:
## What

- Added `related_workflow` field to `WorkflowExecution`
- Deprecated existing `worflow_id` UUID field
- Update `WorkflowExecution` to filter by user who created the workflow

## Why

- Observed issue where logs pane showed executions of all users of the org
image

## How

- Added workflow as a foreign key and updated queryset of workflow_execution

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

- No, minor additive changes only

## Database Migrations

- Schema migration to add a `related_wf` field and made it nullable with existing `workflow_id`

## Related Issues or PRs

- #1173 

## Notes on Testing

- Tested locally by simulating scenario of multiple users in an org

## Screenshots
1. After fix - from user 1
image
1. From user 2
image

## Checklist

I have read and understood the [Contribution Guidelines](https://docs.unstract.com/unstract/contributing/unstract/).

PR #1234, opened 2025-04-09
fix: Separated migrations to avoid DB trigger related issues

chandrasekharan-zipstack:
## What

- Separated workflow's migration into data and schema migration
- Made use of Django ORM itself instead of raw SQL queries
- Minor fix in `WorkflowExecution` creation

## Why

- Faced issues while running migration where updates were pending in the table
```
Message: "Deleting '10867' workflow executions with invalid workflow references"                                                                                                                                 │
│ Arguments: ()                                                                                                                                                                                                    │
│ Traceback (most recent call last):                                                                                                                                                                               │
│   File "/app/.venv/lib/python3.9/site-packages/django/db/backends/utils.py", line 87, in _execute                                                                                                                │
│     return self.cursor.execute(sql)                                                                                                                                                                              │
│ psycopg2.errors.ObjectInUse: cannot ALTER TABLE "workflow_execution" because it has pending trigger events
```

## How

- Separated the migration
- Used django ORM

- Yes, it might for OSS users who might have migrated already. They will have to unapply that and reapply in their DB. Will resolve any issues if anyone reports

## Database Migrations

- Split as data / schema migration

## Related Issues or PRs

- #1226 

## Notes on Testing

- Tested locally and this works. Unable to reproduce issue in local, so couldn't test this fix explicitly

## Screenshots
image

## Checklist


ritwik-g (approved):

backend/workflow_manager/workflow_v2/execution.py
```diff
@@ -146,8 +146,7 @@ def create_workflow_execution(
         workflow_execution = WorkflowExecution.objects.create(
             id=execution_id if execution_id else uuid.uuid4(),
             pipeline_id=pipeline_id,
-            workflow_id=workflow_id,
-            related_workflow=Workflow.objects.get(id=workflow_id),
+            workflow=Workflow.objects.get(id=workflow_id),
```
muhammad-ali-e, line 149: Hope it will not raise DoesnotExist Error.

muhammad-ali-e (approved):
LGTM

backend/workflow_manager/workflow_v2/execution.py
chandrasekharan-zipstack: I wasn't able to reproduce the scenario but I'm assuming it'll be checked by the serializer at the start of the request itself

chandrasekharan-zipstack: It might be raised from [such places](https://github.com/Zipstack/unstract/blob/56c902846edcc278fadbfb0176c49879fcf5524f/backend/workflow_manager/workflow_v2/workflow_helper.py#L703)

ritwik-g: For now will proceed with this. @chandrasekharan-zipstack later if required raise a PR to handle such error if at all needed.

PR #1242, opened 2025-04-10
Sidecar implementation

muhammad-ali-e:
## What

- Implementation of **Tool Sidecar** companion container that runs alongside main tool containers
- Added support for  `docker` client to handle sidecar containers
- Implemented log volume management and cleanup strategies for different client types

## Why

- To enable real-time log processing and streaming to Redis
- To provide consistent container monitoring and completion signal handling
- To improve organization and execution-specific logging capabilities
- To ensure proper log volume cleanup across different container runtimes

## How

- Added `run_container_with_sidecar` method in  Docker client
- Log `volume cleanup` strategy for Docker: Uses named volumes (logs-{file_execution_id}) with explicit cleanup
- Added sidecar configuration support in container clients

No, this PR should not break existing features untill enable the sidecar :

- Sidecar functionality is opt-in via sidecar_enabled env
- Log volume cleanup is handled safely with proper error handling
- Existing container execution paths remain unchanged
- Implementations maintain backward compatibility by env

## Database Migrations

- 

## Env Config

- **TOOL_SIDECAR_IMAGE_NAME**: Name of the sidecar container image
- **TOOL_SIDECAR_IMAGE_TAG**: Tag for the sidecar container image
- **TOOL_SIDECAR_ENABLED**: To enable Sidecar

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


runner/sample.env
```diff
@@ -15,7 +15,7 @@ LOG_LEVEL="INFO"
 REMOVE_CONTAINER_ON_EXIT=True
 
 # Client module path of the container engine to be used.
-CONTAINER_CLIENT_PATH=unstract.runner.clients.docker
+CONTAINER_CLIENT_PATH=unstract.runner.clients.docker_client
```
ritwik-g, line 18: @muhammad-ali-e any specific reason to change the name to `docker_client` instead of `docker`. With the `clients` already present in the module path the `_client` feels reduntant.

runner/sample.env
```diff
@@ -45,3 +45,8 @@ FLASK_ENV=production
 FLASK_RUN_HOST=0.0.0.0
 FLASK_RUN_PORT=5002
 FLASK_APP=unstract.runner:app
+
+# Tool Sidecar
+TOOL_SIDECAR_ENABLED=False
```
ritwik-g, line 50: @muhammad-ali-e since this is gonna be basically dependent on clients can't we make this a constant inside the Implementation of client interface avoiding any possible errors due to misconfigurations? And also this eliminates any need to pass these down to the existing interface and some signature changes.

runner/src/unstract/runner/clients/docker_client.py
```diff
@@ -17,8 +17,9 @@
 
 
 class DockerContainer(ContainerInterface):
-    def __init__(self, container: Container) -> None:
+    def __init__(self, container: Container, logger: logging.Logger) -> None:
         self.container: Container = container
+        self.logger = logger
```
ritwik-g, line 22: @muhammad-ali-e why are we passing the logger in to the class? Can't we simply use the get logger method? Wouldn't that set the module name properly in logs as well? Or is there a specific reason?

runner/src/unstract/runner/clients/docker_client.py
```diff
@@ -28,21 +29,51 @@ def logs(self, follow=True) -> Iterator[str]:
...
+
+        return self.container.wait()
+
+    def cleanup(self, client: Optional[ContainerClientInterface] = None) -> None:
         if not self.container or not Utils.remove_container_on_exit():
             return
         try:
             self.container.remove(force=True)
+            if client:
+                client.cleanup_volume()
```
ritwik-g, line 56: @muhammad-ali-e what is the requirement for this?

runner/src/unstract/runner/clients/docker_client.py
ritwik-g: Or is it the way old implementation existed?

tools/classifier/Dockerfile
```diff
@@ -13,9 +13,14 @@ RUN pip install --no-cache-dir -U pip
...
+    opentelemetry-exporter-otlp \
+    platformdirs>=3.0.0 \
+    && opentelemetry-bootstrap -a install
 # Copy the contents of your project directory into the container at /app
 COPY src /app/src/
 WORKDIR /app/src
 
 
-ENTRYPOINT ["python", "main.py"]
+CMD ["opentelemetry-instrument","python", "main.py"]
```
ritwik-g, line 26: @muhammad-ali-e I believe this needs to be ENTRYPOINT. Else we can't pass additional args. Additional args will override the CMD. This should be changed for in all places.

tool-sidecar/src/unstract/tool_sidecar/log_processor.py
```diff
@@ -0,0 +1,275 @@
...
+logging.basicConfig(
+    level=getattr(logging, os.getenv(Env.LOG_LEVEL, "INFO")),
+    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
+)
+logger = logging.getLogger(__name__)
+
+
+class LogProcessor:
+
+    TERMINATION_MARKER = "TOOL_EXECUTION_COMPLETE"
```
ritwik-g, line 30: @muhammad-ali-e should we use https://github.com/Zipstack/unstract/blob/main/unstract/core/src/unstract/core/constants.py
to avoid any future drifts

runner/sample.env
muhammad-ali-e: There is another python paackge  `docker`. I got conflict whille I was testing with a test script

runner/sample.env
muhammad-ali-e: This variable is used in both docker and k8s. If it for docker , we wil create two containers (I belive this will help to develpement).

runner/src/unstract/runner/clients/docker_client.py
muhammad-ali-e: I used it here since the same behaviour is used in other client

runner/src/unstract/runner/clients/docker_client.py
muhammad-ali-e: For docker, the mounted volume seems to be not deleted even the container stops.

tools/classifier/Dockerfile
muhammad-ali-e: We’re passing the full command from the runner, so it overrides the `CMD`. Using `ENTRYPOINT` was not works  here because it prepend its command, which breaks things when we pass multiple or custom commands from the runner

tool-sidecar/src/unstract/tool_sidecar/log_processor.py
muhammad-ali-e: Sure. will do

johnyrahul (approved):

runner/src/unstract/runner/clients/docker_client.py
chandrasekharan-zipstack: @muhammad-ali-e I think it should work without passing the logger and instead directly using `logging.getLogger(__name__)`

vishnuszipstack (approved):

runner/src/unstract/runner/clients/docker_client.py
```diff
@@ -28,21 +29,51 @@ def logs(self, follow=True) -> Iterator[str]:
...
-    def __init__(self, image_name: str, image_tag: str, logger: logging.Logger) -> None:
+    def __init__(
+        self,
+        image_name: str,
+        image_tag: str,
+        logger: logging.Logger,
+        sidecar_enabled: bool = False,
+    ) -> None:
         self.image_name = image_name
+        self.sidecar_image_name = os.getenv(Env.TOOL_SIDECAR_IMAGE_NAME)
```
chandrasekharan-zipstack, line 70: NIT: Where do we throw an error if its enabled but these envs are not present?

runner/src/unstract/runner/utils.py
```diff
@@ -25,6 +30,27 @@ def str_to_bool(string: str) -> bool:
...
+            int: Parsed integer or the default value.
+        """
+        if not value:
+            return default
+
+        try:
+            return int(value)
+        except ValueError:
+            logger.warning(f"Invalid integer value '{value}'; using default: {default}")
+            return default
```
chandrasekharan-zipstack, line 52: NIT: Such utils can go into `core`. I think we already use `core` in runner as well

tool-sidecar/src/unstract/tool_sidecar/dto.py
```diff
@@ -0,0 +1,16 @@
+from dataclasses import dataclass
+from typing import Any, Optional
+
+
+@dataclass
+class LogLineDTO:
+    is_completion: bool = False
```
chandrasekharan-zipstack, line 7: NIT: @muhammad-ali-e what's this arg for? Consider adding a docstring

tool-sidecar/src/unstract/tool_sidecar/log_processor.py
```diff
@@ -0,0 +1,275 @@
...
+
+        Returns:
+            Optional[Dict]: Parsed JSON if line is a completion signal
+        """
+        # Stream log to Redis
+        if self.TERMINATION_MARKER in line:
+            logger.info(
+                f"Tool container terminated with status {self.TERMINATION_MARKER}"
+            )
+            return LogLineDTO(is_completion=True)
```
chandrasekharan-zipstack, line 99:
```suggestion
            return LogLineDTO(id_terminated=True)
```
I suggest renaming it to `is_complete` / `is_terminated` instead

tool-sidecar/src/unstract/tool_sidecar/log_processor.py
```diff
@@ -0,0 +1,275 @@
...
+                    break
+
+
+def main():
+    """
+    Main entry point for the sidecar container.
+    Sets up the log processor with environment variables and starts monitoring.
+    """
+    # Get configuration from environment
+    log_path = os.getenv(Env.LOG_PATH, "/shared/logs/logs.txt")
```
chandrasekharan-zipstack, line 223: NIT: Should we mention `unstract` anywhere in the path to identify better if needed? Or will this just be a path in the container?

runner/src/unstract/runner/utils.py
muhammad-ali-e: The common utility should be added to a shared module—either in core or a common utility module in unstrat. I believe we already have something similar in the backend, which should also be moved. I’ve added it here for now since we already have a similar utility in place within runner. That said, we should eventually move everything to a single shared location.

tool-sidecar/src/unstract/tool_sidecar/log_processor.py
muhammad-ali-e: It's only for the container

PR #1249, opened 2025-04-13
resolve pgbouncer issues in postgresql connector

muhammad-ali-e:
## What

- Fix PgBouncer compatibility issue by removing unsupported startup parameters
- Simplify PostgreSQL connection handling:
  - Remove connection_timeout and other server parameters
  - Set search_path via SQL query instead of connection options

## Why

- PgBouncer doesn't support certain startup parameters which causes connection failures
- Setting search_path via query is more reliable than connection options (PgBouncer doesn't support options like this)
- Simplifying connection handling makes it more reliable and maintainable

## How

- Removed server-side parameters that conflict with PgBouncer
- Changed search_path setting to use `SET search_path TO schema` query

- No

## Database Migrations

- No

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/unstract/connectors/src/unstract/connectors/databases/postgresql/postgresql.py
+++ b/unstract/connectors/src/unstract/connectors/databases/postgresql/postgresql.py
@@ -67,16 +67,12 @@ class PostgreSQL(UnstractDB, PsycoPgHandler):
         return True
 
     def get_engine(self) -> connection:
-        # Set timeouts via options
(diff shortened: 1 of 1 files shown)


gaya3-zipstack (approved):
Looks safe.

PR #1256, opened 2025-04-21
update status codes

muhammad-ali-e:
## What

- Updated HTTP response handling for workflow execution status checks.
- `EXECUTING` and `PENDING` now return `202 Accepted` instead of `422`, indicating that the task is still in progress.
- `COMPLETED` continues to return `200 OK` as expected (No Changes).
- `ERROR` and `STOPPED` now return `422 Unprocessable Entity` to indicate abnormal or interrupted terminations.
- `406 Not Acceptable` returned when the result has already been acknowledged (No Changes).
- Cleaned up status handling logic using an if-else structure for clarity and maintainability.

## Why

- Returning **422** for valid in-progress states like `EXECUTING` or `PENDING` was incorrect and misleading.

## How

- Mapped statuses explicitly to HTTP response codes:
  - `EXECUTING`, `PENDING` → 202 Accepted
  - `COMPLETED` → 200 OK
  - `ERROR`, `STOPPED` → 422 Unprocessable Entity
  - `Result already acknowledged` → 406 Not Acceptable

- No

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/backend/api_v2/api_deployment_views.py
+++ b/backend/api_v2/api_deployment_views.py
@@ -23,8 +23,8 @@ from rest_framework.decorators import action
 from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework.serializers import Serializer
-from utils.enums import CeleryTaskState
 from workflow_manager.workflow_v2.dto import ExecutionResponse
+from workflow_manager.workflow_v2.enums import ExecutionStatus
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


chandrasekharan-zipstack (commented):
Changes LGTM - however 
1. Won't this break existing clients that rely on the existing (flawed) logic? Ideally we should version the API deployments to aid with such changes however since we don't have it in place I'm wondering how we can carry this out cleanly
1. We might have to update accordingly in our `python-client` and the API deployments batch run script if required

PR #1276, opened 2025-04-29
fix: Error handling improved in tool sandbox and other minor fixes

chandrasekharan-zipstack:
## What

- Tool sandbox error handling improvement
- Raised error instead of return in `prompt-service`
- MINOR: Arg type changed for a function in `core`

## Why

- Minor error handling improvements

## How

- Raise error instead of just logging it

- No, the raised error will be caught and reported correctly

## Notes on Testing

- Tested locally as part of other changes

## Checklist


unstract/core/src/unstract/core/flask/logging.py
```diff
@@ -13,10 +13,10 @@ def filter(self, record):
         return True
 
 
-def setup_logging(log_level: str):
+def setup_logging(log_level: int):
     """Sets up logger for Flask based services
     Args:
-        log_level (str): Log level to use. Can be one of
+        log_level (int): Log level to use. Can be one of
```
pk-zipstack, line 19: Here we say the parameter is of an int type, but the doc string of the function says it has to be of any four `str-values` mentioned.

pk-zipstack (approved):

muhammad-ali-e (approved):
LGTM

PR #1291, opened 2025-05-08
[FIX] JSON object structure

Deepak-Kesavan:
## What

- JSON object structure
- Fix highlight issue

## Why

-

## How

-

-

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/prompt-service/src/unstract/prompt_service/services/answer_prompt.py
+++ b/prompt-service/src/unstract/prompt_service/services/answer_prompt.py
@@ -154,7 +154,7 @@ class AnswerPromptService:
             completion = llm.complete(
                 prompt=prompt,
                 process_text=highlight_data,
-                extract_json=prompt_type.lower() != PSKeys.JSON,
+                extract_json=False,
             )
             answer: str = completion[PSKeys.RESPONSE].text
             highlight_data = completion.get(PSKeys.HIGHLIGHT_DATA, [])
@@ -247,6 +247,7 @@ class AnswerPromptService:
                     structured_output=structured_output,
                     answer=answer,
                 ).run()
+            answer = AnswerPromptService.cut_until_first_bracket(answer)
             if enable_highlight:
                 AnswerPromptService.handle_highlight(
                     execution_source=execution_source,
@@ -313,3 +314,10 @@ class AnswerPromptService:
                     if PSKeys.CONFIDENCE_DATA not in metadata:
                         metadata[PSKeys.CONFIDENCE_DATA] = {}
                     metadata[PSKeys.CONFIDENCE_DATA][prompt_key] = confidence_data
+
+    @staticmethod
+    def cut_until_first_bracket(text) -> str:
+        idx_brace = text.find("{")
+        idx_bracket = text.find("[")
+        start = min(i for i in [idx_brace, idx_bracket] if i != -1)
+        return text[start:]
--- a/prompt-service/src/unstract/prompt_service/services/answer_prompt.py
+++ b/prompt-service/src/unstract/prompt_service/services/answer_prompt.py
@@ -299,12 +299,17 @@ class AnswerPromptService:
             whisper_hash = highlight_response.get(PSKeys.WHISPER_HASH, "")
             prompt_key = output[PSKeys.NAME]
             if metadata is not None and prompt_key:
-                metadata.setdefault(PSKeys.HIGHLIGHT_DATA, {})[prompt_key] = (
-                    highlight_data
-                )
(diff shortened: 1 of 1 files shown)


gaya3-zipstack (approved):
Looks good.

prompt-service/src/unstract/prompt_service/services/answer_prompt.py
```diff
@@ -313,3 +314,10 @@ def handle_highlight(
                     if PSKeys.CONFIDENCE_DATA not in metadata:
                         metadata[PSKeys.CONFIDENCE_DATA] = {}
                     metadata[PSKeys.CONFIDENCE_DATA][prompt_key] = confidence_data
+
+    @staticmethod
+    def cut_until_first_bracket(text) -> str:
+        idx_brace = text.find("{")
+        idx_bracket = text.find("[")
+        start = min(i for i in [idx_brace, idx_bracket] if i != -1)
+        return text[start:]
```
chandrasekharan-zipstack, line 323:
```suggestion
    def slice_from_first_bracket(text: str) -> str:
        match = re.search(r'[\{\[]', text)
        if match:
            return text[match.start():]
        return text  # or return "" if you'd prefer to return an empty string
```
@Deepak-Kesavan the implementation is to extract text from the first bracket right? If so, I think the function is named incorrectly.
Also it'll fail with a value error if the text does not have any of those braces. Consider the regex implementation instead

harini-venkataraman (approved):

prompt-service/src/unstract/prompt_service/services/answer_prompt.py
```diff
@@ -154,7 +154,7 @@ def run_completion(
             completion = llm.complete(
                 prompt=prompt,
                 process_text=highlight_data,
-                extract_json=prompt_type.lower() != PSKeys.JSON,
+                extract_json=False,
```
chandrasekharan-zipstack, line 157: Why was this change made?

prompt-service/src/unstract/prompt_service/services/answer_prompt.py
```diff
@@ -313,3 +314,10 @@ def handle_highlight(
                     if PSKeys.CONFIDENCE_DATA not in metadata:
                         metadata[PSKeys.CONFIDENCE_DATA] = {}
                     metadata[PSKeys.CONFIDENCE_DATA][prompt_key] = confidence_data
+
+    @staticmethod
+    def cut_until_first_bracket(text) -> str:
```
harini-venkataraman, line 319: MINOR: This method can be moved to helpers.

prompt-service/src/unstract/prompt_service/services/answer_prompt.py
Deepak-Kesavan: It is related to highlight issue reported @chandrasekharan-zipstack

chandrasekharan-zipstack: @Deepak-Kesavan now for all cell types we always set this to `False`. What does this parameter do? Do we ever set it to `True` now?

prompt-service/src/unstract/prompt_service/services/answer_prompt.py
chandrasekharan-zipstack: @Deepak-Kesavan the name of the function is misleading btw. Please change that too

Deepak-Kesavan: @chandrasekharan-zipstack 

Using re.search for this case is unnecessary and introduces overhead. We're only checking for the first occurrence of one of two fixed characters ({ or [), which is more efficiently and clearly handled with str.find. Regex adds complexity and cost without benefit here.

I have fixed the case where brackets are not found in current approach itself.

prompt-service/src/unstract/prompt_service/services/answer_prompt.py
Deepak-Kesavan: Yes @chandrasekharan-zipstack . It is left in the SDK to maintain backward compatibility.

prompt-service/src/unstract/prompt_service/services/answer_prompt.py
Deepak-Kesavan: @harini-venkataraman There is no separate helper to keep this function. So I kept it here in answer_prompt service.

chandrasekharan-zipstack (approved):

PR #1316, opened 2025-05-19
[FEAT] Added support for s3 presigned urls in api deployment

pk-zipstack:
## What

- Added support for presigned s3 URLs in API deployment

## Why

- This was requested by customers.

## How

- Added a function that fetched the files from s3 presigned URLs and adds them to the `file_objs`. Which then processes the files using the base functionalities.

- Yes, this can break in the part of validating the files since this PR has modifications in Serializers. But this was tested in local and did not throw any issues.

## Database Migrations

- NO

## Env Config

- NO

## Relevant Docs

-

## Related Issues or PRs

- https://zipstack.atlassian.net/browse/UN-2354

## Dependencies Versions

-

## Notes on Testing

- Tested adding single and multiple URLs in testing. 

## Checklist

### DIFF
--- a/backend/api_v2/api_deployment_views.py
+++ b/backend/api_v2/api_deployment_views.py
@@ -49,12 +49,19 @@ class DeploymentExecution(views.APIView):
     ) -> Response:
         serializer = ExecutionRequestSerializer(data=request.data)
         serializer.is_valid(raise_exception=True)
-        file_objs = serializer.validated_data.get(ApiExecution.FILES_FORM_DATA)
+        file_objs = serializer.validated_data.get(ApiExecution.FILES_FORM_DATA, [])
+        presigned_urls = serializer.validated_data.get(ApiExecution.PRESIGNED_URLS, [])
         timeout = serializer.validated_data.get(ApiExecution.TIMEOUT_FORM_DATA)
         include_metadata = serializer.validated_data.get(ApiExecution.INCLUDE_METADATA)
         include_metrics = serializer.validated_data.get(ApiExecution.INCLUDE_METRICS)
         use_file_history = serializer.validated_data.get(ApiExecution.USE_FILE_HISTORY)
         tag_names = serializer.validated_data.get(ApiExecution.TAGS)
+
+        if presigned_urls:
+            error_response = DeploymentHelper.load_presigned_files(presigned_urls, file_objs)
+            if error_response:
+                return error_response
+
         response = DeploymentHelper.execute_workflow(
             organization_name=org_name,
             api=api,
--- a/backend/api_v2/deployment_helper.py
+++ b/backend/api_v2/deployment_helper.py
@@ -1,9 +1,15 @@
 import logging
+import requests
 from typing import Any
 from urllib.parse import urlencode
+from io import BytesIO
+
 
 from django.conf import settings
 from django.core.files.uploadedfile import UploadedFile
+from django.core.files.uploadedfile import InMemoryUploadedFile
+from rest_framework import status
+from rest_framework.response import Response
 from rest_framework.request import Request
 from rest_framework.serializers import Serializer
 from rest_framework.utils.serializer_helpers import ReturnDict
@@ -220,3 +226,29 @@ class DeploymentHelper(BaseAPIKeyValidator):
             execution_id=execution_id
         )
         return execution_response
+    
+    def load_presigned_files(presigned_urls: list[str], file_objs: list[InMemoryUploadedFile]) -> Response | None:
+
+        for url in presigned_urls:
+            try:
+                resp = requests.get(url)
+                resp.raise_for_status()
+            except Exception as e:
+                return Response(
+                    {"message": f"Failed to fetch file from URL: {url}", "error": str(e)},
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
+    
+            filename = url.split("?")[0].split("/")[-1]
+            file_stream = BytesIO(resp.content)
+            uploaded_file = InMemoryUploadedFile(
+                file=file_stream,
+                field_name="file",
+                name=filename,
+                content_type=resp.headers.get("Content-Type", "application/octet-stream"),
+                size=len(resp.content),
+                charset=None
+            )
+            file_objs.append(uploaded_file)
+
+        return None
--- a/backend/api_v2/serializers.py
+++ b/backend/api_v2/serializers.py
@@ -14,6 +14,7 @@ from rest_framework.serializers import (
     ModelSerializer,
     Serializer,
     ValidationError,
+    URLField
 )
 from tags.serializers import TagParamsSerializer
 from utils.serializer.integrity_error_mixin import IntegrityErrorMixin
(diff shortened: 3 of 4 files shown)


backend/api_v2/api_deployment_views.py
```diff
@@ -49,12 +49,21 @@ def post(
...
         include_metrics = serializer.validated_data.get(ApiExecution.INCLUDE_METRICS)
         use_file_history = serializer.validated_data.get(ApiExecution.USE_FILE_HISTORY)
         tag_names = serializer.validated_data.get(ApiExecution.TAGS)
+
+        if presigned_urls:
+            error_response = DeploymentHelper.load_presigned_files(
+                presigned_urls, file_objs
+            )
+            if error_response:
+                return error_response
```
chandrasekharan-zipstack, line 65: @pk-zipstack why are we returning an error instead of raising it?

backend/api_v2/deployment_helper.py
```diff
@@ -220,3 +224,30 @@ def get_execution_status(execution_id: str) -> ExecutionResponse:
...
+    ) -> Response | None:
+        for url in presigned_urls:
+            try:
+                resp = requests.get(url)
+                resp.raise_for_status()
+            except Exception as e:
+                return Response(
+                    {"message": f"Failed to fetch file from URL: {url}", "error": str(e)},
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
```
chandrasekharan-zipstack, line 239: @pk-zipstack catch the specific exception such as `ConnectionError` / `RequestException`. You can refer how we've handled it in the [SDK](https://github.com/Zipstack/unstract-sdk/blob/b90ee68f5567b44f8820396bd4aa515e709596be/src/unstract/sdk/platform.py#L132-L148).
Raise this as an error so that the error message gets rendered in a common format that we've configured with `drf-standardized-errors`. It'll be easier to handle it in the FE and to maintain any changes in the format

backend/api_v2/deployment_helper.py
```diff
@@ -220,3 +224,30 @@ def get_execution_status(execution_id: str) -> ExecutionResponse:
...
+            file_stream = BytesIO(resp.content)
+            uploaded_file = InMemoryUploadedFile(
+                file=file_stream,
+                field_name="file",
+                name=filename,
+                content_type=resp.headers.get("Content-Type", "application/octet-stream"),
+                size=len(resp.content),
+                charset=None,
+            )
+            file_objs.append(uploaded_file)
```
chandrasekharan-zipstack, line 251: @pk-zipstack Consider using a generator to save some memory. I suggest adding a separate method which fetches one file and you can `yield` it like so
```python
for url in presigned_urls:
    yield DeploymentHelper.fetch_presigned_file(url)
```

backend/api_v2/serializers.py
```diff
@@ -124,23 +125,30 @@ class ExecutionRequestSerializer(TagParamsSerializer):
...
-        if len(value) == 0:
-            raise ValidationError("The file list cannot be empty.")
-        if len(value) > self.MAX_FILES_ALLOWED:
-            raise ValidationError(f"Maximum '{self.MAX_FILES_ALLOWED}' files allowed.")
-        return value
+    def validate(self, data):
+        files = data.get("files", [])
+        urls = data.get("presigned_urls", [])
+
+        total = len(files) + len(urls)
```
chandrasekharan-zipstack, line 141: @pk-zipstack does this mean we support passing files as well as URLs in the same call and limit it to 32?

backend/api_v2/api_deployment_views.py
pk-zipstack: @chandrasekharan-zipstack Sorry, missed it while raising the PR. Will correct it. Thanks for pointing it out.

backend/api_v2/serializers.py
pk-zipstack: Yes @chandrasekharan-zipstack, First, the decision was to support only one form of file submission, but after seeing that the files are processed as `file_objs`, this decision was made.

chandrasekharan-zipstack: Alright, make sure to update our public docs accordingly when the PR is merged. All this could be useful features

backend/api_v2/deployment_helper.py
```diff
@@ -33,6 +37,9 @@
 
 logger = logging.getLogger(__name__)
 
+# Default maximum file size for presigned URLs (20MB)
+DEFAULT_PRESIGNED_URL_MAX_FILE_SIZE_MB = 20
```
chandrasekharan-zipstack: @pk-zipstack I suggest moving this to our django settings and optionally exposing it as an env if you think we might want to change this value later (maybe env is overkill right away)
In that regard, this constant for a default value won't be needed

backend/api_v2/deployment_helper.py
```diff
@@ -248,3 +256,163 @@ def get_execution_status(execution_id: str) -> ExecutionResponse:
...
+        # Accept common S3 endpoint patterns: path-style, virtual-hosted, regional, dualstack, accelerated
+        looks_like_s3 = (
+            host == "s3.amazonaws.com"
+            or host.endswith(".s3.amazonaws.com")
+            or re.match(r"(^|.*\.)s3[.-]([a-z0-9-]+)\.amazonaws\.com$", host) is not None
+        )
+
+        if not (is_aws and looks_like_s3):
+            raise PresignedURLFetchError(
+                url=sanitized_url, error_message="URL host is not a valid S3 endpoint"
```
chandrasekharan-zipstack:
```suggestion
                url=sanitized_url, error_message="URL host is not a valid S3 endpoint, only S3's pre-signed URLs are supported currently"
```

backend/api_v2/deployment_helper.py
```diff
@@ -248,3 +256,163 @@ def get_execution_status(execution_id: str) -> ExecutionResponse:
...
+            max_mb = getattr(
+                settings,
+                "PRESIGNED_URL_MAX_FILE_SIZE_MB",
+                DEFAULT_PRESIGNED_URL_MAX_FILE_SIZE_MB,
+            )
+            max_bytes = int(max_mb) * 1024 * 1024
+        except Exception:
+            max_bytes = (
+                DEFAULT_PRESIGNED_URL_MAX_FILE_SIZE_MB * 1024 * 1024
+            )  # sane default if settings unavailable
```
chandrasekharan-zipstack: I don't see this added in the settings file (`base.py`). We can specify the default of 20 MB there itself instead of specifying the constant here

backend/api_v2/deployment_helper.py
```diff
@@ -248,3 +256,163 @@ def get_execution_status(execution_id: str) -> ExecutionResponse:
...
+        if not (is_aws and looks_like_s3):
+            raise PresignedURLFetchError(
+                url=sanitized_url, error_message="URL host is not a valid S3 endpoint"
+            )
+
+        # Get file size limit from settings
+        try:
+            max_mb = getattr(
+                settings,
+                "PRESIGNED_URL_MAX_FILE_SIZE_MB",
```
chandrasekharan-zipstack:
```suggestion
                "API_DEPL_PRESIGNED_URL_MAX_FILE_SIZE_MB",
```

backend/api_v2/deployment_helper.py
```diff
@@ -248,3 +256,163 @@ def get_execution_status(execution_id: str) -> ExecutionResponse:
...
+            return InMemoryUploadedFile(
+                file=file_stream,
+                field_name="file",
+                name=filename,
+                content_type=content_type,
+                size=downloaded,
+                charset=None,
+            )
+
+        except (requests.ConnectionError, requests.RequestException) as e:
```
chandrasekharan-zipstack:
```suggestion
        except requests.RequestException as e:
```
`ConnectionError` inherits `RequestException` so its redundant to specify both here

backend/api_v2/deployment_helper.py
```diff
@@ -248,3 +256,163 @@ def get_execution_status(execution_id: str) -> ExecutionResponse:
...
+            # Close the file stream on error
+            if file_stream:
+                file_stream.close()
```
chandrasekharan-zipstack: @pk-zipstack this can be a part of the `finally` block instead to ensure its closed always

backend/api_v2/exceptions.py
```diff
@@ -56,3 +56,19 @@ def __init__(
         if detail is None:
             detail = f"No active API keys configured for {deployment_name}"
         super().__init__(detail, code)
+
+
+class PresignedURLFetchError(APIException):
+    status_code = 400
```
chandrasekharan-zipstack: NIT: Reporting as 400 always might not make sense, sometimes it could be an auth / permission related error so we'll need to report with the correct status code. Check if that is handled here

backend/api_v2/api_deployment_views.py
```diff
@@ -69,6 +70,9 @@ def post(
         llm_profile_id = serializer.validated_data.get(ApiExecution.LLM_PROFILE_ID)
         hitl_queue_name = serializer.validated_data.get(ApiExecution.HITL_QUEUE_NAME)
 
+        if presigned_urls:
+            DeploymentHelper.load_presigned_files(presigned_urls, file_objs)
```
chandrasekharan-zipstack: NIT: This loading and validation can be done from within the serializer too

PR #1338, opened 2025-06-04
fix: Indexing optimization in structure tool

chandrasekharan-zipstack:
## What
- Below changes in `tool-structure`
  - Ensured indexing is done only for the required combination
  - Renamed some boolean variables for readability
  - Minor log improvments

## Why

- Indexing was previously done for all prompts in the project since the extraction was linked to it and any prompt could have any config
- This was wasteful and only the unique combinations of non LLM adapters mattered here along with chunk size and chunk overlap
- Helps display cleaner logs which can help improve the understanding of what's happening

## How

- Used a set to obtain the unique combinations and called indexing for that alone

- Yes, since it refactored some code involving different possibly stale code flows

## Notes on Testing

- Tested the changes locally for some simple executions, due to time constraint - it was not tested robustly. Hoping to cover that as part of the test suite

## Checklist


tools/structure/src/main.py
```diff
@@ -150,54 +150,70 @@ def run(
...
-                ] != 0:
+                # Get current parameter combination
+                chunk_size = output[SettingsKeys.CHUNK_SIZE]
+                chunk_overlap = output[SettingsKeys.CHUNK_OVERLAP]
+                vector_db = tool_settings[SettingsKeys.VECTOR_DB]
+                embedding = tool_settings[SettingsKeys.EMBEDDING]
+                x2text = tool_settings[SettingsKeys.X2TEXT_ADAPTER]
+
+                # Create a unique key for this parameter combination
+                param_key = (
```
harini-venkataraman, line 175: MINOR : Lets use a hash of this or reuse SDK`s generate_index_key method to maintain consistency.

harini-venkataraman (approved):
Left a minor comment, LGTM otherwise

tools/structure/src/main.py
chandrasekharan-zipstack: @harini-venkataraman this was done to reduce the effort to perform a lookup in the set. Also that method requires the file hash and other params so I didn't spend time on using it. Went ahead with this since its simpler and gets the job done. Will consider using that in case of any issues

PR #1340, opened 2025-06-04
feat: Using rabbitMQ as message broker for celery instead of redis

chandrasekharan-zipstack:
## What

- Changes to use RabbitMQ as the message broker instead of redis for celery
- Added `rabbitmq` to docker-dev-essentials and added the same as a dependency for some containers
- Updated envs required for this change

## Why

- RabbitMQ behaves better than Redis when a worker is restarted and the task is picked up immediately. Whereas in Redis we need to wait for a fixed timeout

## How

- Replaced the broker URL and used RabbitMQ instead of Redis

- Yes, since the message broker has been updated. Requires some additional testing

## Env Config

```
RABBITMQ_USER=admin
RABBITMQ_PASS=password
RABBITMQ_HOST="unstract-rabbitmq"
RABBITMQ_PORT=5672
CELERY_BROKER_URL="amqp://admin:password@unstract-rabbitmq:5672//"
```

## Dependencies Versions

- Added `amqp` extra to `celery` in backend

## Notes on Testing

- Tested workflow and ETL execution briefly and it worked

## Screenshots
image

## Checklist


prompt-service/sample.env
```diff
@@ -18,7 +18,7 @@ LOG_LEVEL=INFO
 
 ###  Env from `unstract-core`  ###
 # Celery for PublishLogs
-CELERY_BROKER_URL="redis://unstract-redis:6379"
+CELERY_BROKER_URL="amqp://admin:password@unstract-rabbitmq:5672//"
```
ritwik-g: @chandrasekharan-zipstack does the prompt-service even need to know the CELERY BROKER? Same question with runner also

docker/sample.env
```diff
@@ -7,10 +7,7 @@ TOOL_REGISTRY_CONFIG_SRC_PATH="${PWD}/../unstract/tool-registry/tool_registry_co
 # Format: <max_workers>,<min_workers>
 # Hint: The max value (max_workers) is related to your CPU resources and the level of concurrency you need.
 # Always monitor system performance and adjust the max value as needed.
-WORKER_API_DEPLOYMENTS_AUTOSCALE=4,1
 WORKER_LOGGING_AUTOSCALE=4,1
 WORKER_AUTOSCALE=4,1
 WORKER_FILE_PROCESSING_AUTOSCALE=4,1
-WORKER_API_FILE_PROCESSING_AUTOSCALE=4,1
 WORKER_FILE_PROCESSING_CALLBACK_AUTOSCALE=4,1
-WORKER_API_FILE_PROCESSING_CALLBACK_AUTOSCALE=4,1
```
ritwik-g, line 16: @chandrasekharan-zipstack why were these lines removed?

docker/sample.compose.override.yaml
```diff
@@ -12,7 +12,3 @@ services:
   worker-logging:
     profiles:
       - high_memory
-
-  worker-api-deployment:
-    profiles:
-      - high_memory
```
ritwik-g, line 18: Is this an inteded change

docker/docker-compose.yaml
```diff
@@ -252,6 +235,7 @@ services:
       - /var/run/docker.sock:/var/run/docker.sock
     depends_on:
       - redis
+      - rabbitmq
```
ritwik-g, line 238: @chandrasekharan-zipstack I don't think runner depends on rabbitmq. So we might not need it here

docker/docker-compose.yaml
```diff
@@ -172,7 +154,7 @@ services:
       - ./essentials.env
     depends_on:
         - db
-        - redis
```
ritwik-g, line 175: @chandrasekharan-zipstack please ensure if the redis is is required for something else like caching.

docker/docker-compose.yaml
```diff
@@ -66,44 +68,26 @@ services:
...
-      - ${TOOL_REGISTRY_CONFIG_SRC_PATH}:/data/tool_registry_config
-
   # Celery worker for handling file processing tasks
   worker-file-processing:
     image: unstract/backend:${VERSION}
     container_name: unstract-worker-file-processing
     restart: unless-stopped
     entrypoint: .venv/bin/celery
-    command: "-A backend.workers.file_processing worker --loglevel=info -Q file_processing --autoscale=${WORKER_FILE_PROCESSING_AUTOSCALE}"
+    command: "-A backend.workers.file_processing worker --loglevel=info -Q file_processing,api_file_processing --autoscale=${WORKER_FILE_PROCESSING_AUTOSCALE}"
```
ritwik-g, line 85: @chandrasekharan-zipstack looks like api and ETL file processing has been combined. Is this done intentionally?

backend/migrating/v2/migrate-to-v2.sh
```diff
@@ -36,13 +36,10 @@ activate_venv
...
 echo -e "${blue_text}Running schema creation command...${default_text}"
-DB_HOST="localhost" REDIS_HOST="localhost" CELERY_BROKER_URL="redis://localhost:6379" \
 python manage.py create_schema || { echo "Schema creation failed"; exit 1; }
 
 echo -e "${blue_text}Running schema migration command...${default_text}"
-DB_HOST="localhost" REDIS_HOST="localhost" CELERY_BROKER_URL="redis://localhost:6379" \
 python manage.py migrate || { echo "Schema migration failed"; exit 1; }
 
 echo -e "${blue_text}Running data migration command...${default_text}"
-DB_HOST="localhost" REDIS_HOST="localhost" CELERY_BROKER_URL="redis://localhost:6379" \
```
ritwik-g, line 47: @chandrasekharan-zipstack what are these changes for?

backend/sample.env
```diff
@@ -129,9 +128,11 @@ LOGS_EXPIRATION_TIME_IN_SECOND=86400
 
 # Celery Configuration
 # Used by celery and to connect to queue to push logs
-CELERY_BROKER_URL="redis://unstract-redis:6379"
-# Redis broker visibility set to 1 day
-CELERY_BROKER_VISIBILITY_TIMEOUT=86400
+RABBITMQ_USER=admin
+RABBITMQ_PASS=password
+RABBITMQ_HOST="unstract-rabbitmq"
+RABBITMQ_PORT=5672
```
ritwik-g: @chandrasekharan-zipstack are these ENVs required in the backend.env? Isn't the `CELERY_BROKER_URL` Alone enough?

backend/backend/settings/base.py
```diff
@@ -45,6 +45,11 @@ def get_required_setting(setting_key: str, default: str | None = None) -> str |
 # Load default log from env
 DEFAULT_LOG_LEVEL = os.environ.get("DEFAULT_LOG_LEVEL", "INFO")
 
+# RabbitMQ Configuration
+RABBITMQ_USER = os.environ.get("RABBITMQ_USER", "admin")
+RABBITMQ_PASS = os.environ.get("RABBITMQ_PASS", "password")
+RABBITMQ_HOST = os.environ.get("RABBITMQ_HOST", "unstract-rabbitmq")
+RABBITMQ_PORT = os.environ.get("RABBITMQ_PORT", 5672)
```
ritwik-g: @chandrasekharan-zipstack are these required?

ritwik-g: Ignore saw the usage. But please remove the default values added it is not a good idea. May be for port alone you can keep everything else should be removed.

backend/sample.env
ritwik-g: Ignore found the usage

prompt-service/sample.env
chandrasekharan-zipstack: @ritwik-g we use the celery broker for Kombu to push log messages from these services. Hence we need to define envs needed for the broker URL

docker/sample.env
chandrasekharan-zipstack: @ritwik-g for OSS we defined separate workers to handle below separately. 
- API related 
- ETL / Task / Workflows related

This causes a lot of containers to be run and it becomes harder to run even though only the queues were different. Hence these were unified to an extent and unused envs were removed

docker/sample.compose.override.yaml
chandrasekharan-zipstack: Yes its related to [this comment](https://github.com/Zipstack/unstract/pull/1340#discussion_r2128053451)

docker/docker-compose.yaml
chandrasekharan-zipstack: @ritwik-g I added it based on how the runner uses Kombu (through `unstract-core`) to push log messages. Hence the `rabbitmq` container needs to be up before the runner starts. Let me know if my understanding here is wrong

docker/docker-compose.yaml
chandrasekharan-zipstack: Yes, in order to lower the friction for running and developing on OSS. Refer [this comment](https://github.com/Zipstack/unstract/pull/1340#discussion_r2128053451)

backend/migrating/v2/migrate-to-v2.sh
chandrasekharan-zipstack: @ritwik-g I previously added the rabbitMQ CELERY_BROKER_URL here - however it raised a security issue because the password was included.
Anyway in this script, since we load the envs beforehand this won't be needed

docker/docker-compose.yaml
chandrasekharan-zipstack: @ritwik-g any ideas on how to confirm this? Though we use the backend image which requires redis we only run the `beat` command here that requires a celery broker. Since we switch to `rabbitmq` I assumed that alone is needed. I don't see their docs mention anything about caching with redis

ritwik-g (approved):

PR #1360, opened 2025-06-11
Add retry with backoff for runner API calls to handle errors during scalability and restarts

muhammad-ali-e:
## What

- Added retry logic with exponential backoff for runner API calls using **urllib3.Retry** and **HTTPAdapter**.

## Why

- To handle transient errors during runner scalability or restarts (e.g., HTTP 500/503, timeouts), ensuring better reliability and fault tolerance.
- Helps reduce job failures caused by short-lived unavailability of the runner service.

## How

- Configured a Retry strategy 

- No

## Database Migrations

- 

## Env Config

- `UNSTRACT_RUNNER_API_TIMEOUT`	(_Default 60_)	Timeout in seconds for each API request
- `UNSTRACT_RUNNER_API_RETRY_COUNT`	(_Default 5_)	Max retry attempts for failed API calls
- `UNSTRACT_RUNNER_API_BACKOFF_FACTOR`	(_Default 3_)	Exponential backoff factor for retries

`All have sensible defaults and do not require changes unless tuning is needed.`

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/unstract/tool-sandbox/src/unstract/tool_sandbox/helper.py
+++ b/unstract/tool-sandbox/src/unstract/tool_sandbox/helper.py
@@ -3,11 +3,14 @@ import logging
 import os
 import time
 from datetime import UTC, datetime, timedelta
+from enum import Enum
 from typing import Any
 
 import requests
 from requests import Response
+from requests.adapters import HTTPAdapter
 from requests.exceptions import ConnectionError, RequestException
+from urllib3.util.retry import Retry
 
 from unstract.core.file_execution_tracker import (
     FileExecutionData,
@@ -43,6 +46,11 @@ class ToolSanboxError(Exception):
     pass
 
 
+class HTTPMethod(str, Enum):
+    GET = "GET"
+    POST = "POST"
+
+
 class ToolSandboxHelper:
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


chandrasekharan-zipstack:
@muhammad-ali-e please check the security hotspot. Also go through [this doc](https://zipstack.atlassian.net/wiki/spaces/PT/pages/887980034/Release+Notes+Standardisation#Branch%2C-Commits-and-PRs-based-on-Jira-tickets%2Fgithub-issue) to correct the PR title as necessary

unstract/tool-sandbox/src/unstract/tool_sandbox/helper.py
```diff
@@ -43,6 +46,11 @@ class ToolSanboxError(Exception):
     pass
 
 
+class HTTPMethod(str, Enum):
+    GET = "GET"
+    POST = "POST"
```
chandrasekharan-zipstack: NIT: I suggest adding this to core and using it from there. I see that this enum is defined in multiple places in our code base and probably `unstract-core` can help maintain a single copy

muhammad-ali-e: Good suggestion.... I’ll update the PR to move not just the HTTPMethod enum but the entire retry logic into the core package — either as a class or function. This way, all modules can reuse it consistently.

unstract/core/src/unstract/core/network/retry.py
```diff
@@ -0,0 +1,60 @@
...
+    enable_http: bool = True,
+    enable_https: bool = True,
+) -> requests.Session:
+    """Get retry adapter for requests.
+
+    This function creates a requests session with a retry strategy.
+
+    Args:
+        retry_count (int, optional): Number of retries. Defaults to 5.
+        backoff_factor (int, optional): Backoff factor. Defaults to 3.
```
chandrasekharan-zipstack: @muhammad-ali-e what does this backoff factor do? Its not obvious from the name. Does it wait for 3 seconds before retrying? Can you elaborate in the doc string itself?

unstract/tool-sandbox/src/unstract/tool_sandbox/helper.py
```diff
@@ -60,6 +61,18 @@ def __init__(
         self.execution_id = str(execution_id)
         self.envs = environment_variables
         self.messaging_channel = str(messaging_channel)
+        self.timeout = int(os.getenv("UNSTRACT_RUNNER_API_TIMEOUT", 60))
```
chandrasekharan-zipstack: @muhammad-ali-e will 60s be enough for the runner API call? In case where a new tool is being pulled, it might take more time

unstract/core/src/unstract/core/network/retry.py
muhammad-ali-e: `backoff_factor` is a common term used in retry mechanisms (I don't think it need a docstring). It determines the base for exponential delay between retries.Specifically, the wait time before each retry.
`backoff_factor * (2 ** (retry_count - 1))`

unstract/tool-sandbox/src/unstract/tool_sandbox/helper.py
muhammad-ali-e: Increased the default timeout to address this concern. The value can still be overridden via environment variables if needed. 
We may introduce a separate timeout variable specifically for the `run_container` call in the future.

unstract/core/src/unstract/core/network/retry.py
chandrasekharan-zipstack: Still I don't think its a concern to mention it in the docstring for other devs. Or we could link it to an appropriate resource that talks about this

muhammad-ali-e: Updated

PR #1363, opened 2025-06-13
UN-2429 [MISC] Added OTEL filter for handling missing trace / span ID errors

chandrasekharan-zipstack:
## What

- Added OTEL filter for handling missing trace / span ID errors
- Logging configuration for tool sidecar (similar to other services)
- MINOR: Ensured that in case any ID is missing, a `-` is added

## Why

- Avoid missing trace / span ID related errors

## How

- Added a logging filter for the same

- No

## Related Issues or PRs

- [UN-2429](https://zipstack.atlassian.net/browse/UN-2429)

## Notes on Testing

- Tried it for a prompt studio run and a tool run

## Screenshots
- From backend
image
- Tool sidecar
image

## Checklist

[UN-2429]: https://zipstack.atlassian.net/browse/UN-2429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/backend/settings/base.py
```diff
@@ -181,28 +182,36 @@ def get_required_setting(setting_key: str, default: str | None = None) -> str |
...
 
-# Determine if OpenTelemetry trace context should be included in logs
-OTEL_TRACE_CONTEXT = (
-    " trace_id:%(otelTraceID)s span_id:%(otelSpanID)s"
-    if os.environ.get("OTEL_TRACES_EXPORTER", "none").lower() != "none"
-    else ""
-)
+# Request ID middleware settings
+LOG_REQUEST_ID_HEADER = "X-Request-ID"
+REQUEST_ID_RESPONSE_HEADER = "X-Request-ID"
```
muhammad-ali-e, line 187: Curious — both constants have the same value.Are these intentionally set as separate constants ?

muhammad-ali-e (approved):
LGTM

tool-sidecar/src/unstract/tool_sidecar/__init__.py
ritwik-g, line 1: @chandrasekharan-zipstack adding it here won't be beneficial. We should add it in the tool. Why? Side car really doesn't do any operations. Actual operations happen in tool. So unless we add it there it might not make sense to have trace id here.

chandrasekharan-zipstack: Ah okay got it. However @ritwik-g, most of the "logs" from the tool are actually print statements so configuring a logger might not be beneficial too right? Also we've ensured that we don't display the tool container's logs directly too

Here's an example of what I mean from the sidecar logs
image

Very few actual log lines are present for a tool run though
image

backend/backend/settings/base.py
chandrasekharan-zipstack: Yes @muhammad-ali-e, one if for the request header and the other is for the response header. They are settings imposed by the request ID middleware

tool-sidecar/src/unstract/tool_sidecar/__init__.py
ritwik-g: Yeah. It is a bit tricky. We need to consider some other mechanism to propagate the trace id from there to side car. Let's create a separate ticket for it.

prompt-service/src/unstract/prompt_service/utils/request_id_filter.py
ritwik-g, line 1: @chandrasekharan-zipstack why is this not required anymore?

chandrasekharan-zipstack: @ritwik-g logging for Flask services is configured at a single place in `unstract-core`. This filter class is redundant and is unused. I removed it to avoid confusion

ritwik-g: Okay

PR #1368, opened 2025-06-16
[MISC] Updated workflows to use github's cache for faster builds

chandrasekharan-zipstack:
## What

- Used Github cache for docker builds in the workflow
- Previously most of these image builds took 2-3 mins and sometimes it might not even have any changes in it. Now with the cache it'll complete 

## Why

- To make better use of docker caching and speed up builds

## How

- Leveraged [github cache](https://docs.docker.com/build/ci/github-actions/cache/#github-cache) and used [docker-buildx-push action](https://github.com/docker/build-push-action)

- No. It has been tested by dry running the workflow builds for these Github actions (without pushing to Dockerhub). However we'll have to properly test it with the push enabled too

## Relevant Docs

- https://docs.docker.com/build/ci/github-actions/cache/#github-cache
- https://github.com/docker/build-push-action

## Notes on Testing

- Tested the actions by running it from this branch, able to build. Haven't explicitly tested the push
- Check build summary obtained with bake here: https://github.com/Zipstack/unstract/actions/runs/15702835422

## Screenshots
- For tools
  - [Run that caches layers for tool-classifier ](https://github.com/Zipstack/unstract/actions/runs/15686047257)
  - [Cached run for tool-classifier](https://github.com/Zipstack/unstract/actions/runs/15686138919)
image
- For services
  - [Run that caches layers](https://github.com/Zipstack/unstract/actions/runs/15687141228)
  - [Cached run](https://github.com/Zipstack/unstract/actions/runs/15687239087)
image

- Build summary

image

## Checklist

### DIFF
--- a/.github/workflows/docker-tools-build-push.yaml
+++ b/.github/workflows/docker-tools-build-push.yaml
@@ -66,4 +66,5 @@ jobs:
             set: |
               *.cache-from=type=gha,scope=${{ github.event.inputs.service_name }}
               *.cache-to=type=gha,mode=max,scope=${{ github.event.inputs.service_name }}
+              *.context=.
               *.args.PROFILE=tools
--- a/.github/workflows/docker-tools-build-push.yaml
+++ b/.github/workflows/docker-tools-build-push.yaml
@@ -47,8 +47,7 @@ jobs:
           uses: docker/build-push-action@v5
           with:
             context: ./tools/classifier
-            # push: true
-            push: false  # TODO: Make true after testing is complete
+            push: true
             tags: unstract/${{github.event.inputs.service_name}}:${{ github.event.inputs.tag }}
             cache-from: type=gha
             cache-to: type=gha,mode=max
@@ -59,8 +58,7 @@ jobs:
           uses: docker/build-push-action@v5
           with:
             context: ./tools/structure
-            # push: true
-            push: false  # TODO: Make true after testing is complete
+            push: true
             tags: unstract/${{github.event.inputs.service_name}}:${{ github.event.inputs.tag }}
             cache-from: type=gha
             cache-to: type=gha,mode=max
@@ -71,8 +69,7 @@ jobs:
           uses: docker/build-push-action@v5
           with:
             context: ./tools/text_extractor
-            # push: true
-            push: false  # TODO: Make true after testing is complete
+            push: true
             tags: unstract/${{github.event.inputs.service_name}}:${{ github.event.inputs.tag }}
             cache-from: type=gha
             cache-to: type=gha,mode=max
@@ -84,8 +81,7 @@ jobs:
           with:
             context: .
             file: docker/dockerfiles/tool-sidecar.Dockerfile
-            # push: true
-            push: false  # TODO: Make true after testing is complete
+            push: true
             tags: unstract/${{github.event.inputs.service_name}}:${{ github.event.inputs.tag }}
             cache-from: type=gha
             cache-to: type=gha,mode=max
--- a/.github/workflows/docker-tools-build-push.yaml
+++ b/.github/workflows/docker-tools-build-push.yaml
@@ -30,23 +30,58 @@ jobs:
         - name: Checkout code
           uses: actions/checkout@v4
 
+        # Set up Docker Buildx for better caching
+        - name: Set up Docker Buildx
+          uses: docker/setup-buildx-action@v3
+
+        # Log in to Docker Hub
         - name: Login to Docker Hub
           uses: docker/login-action@v3
           with:
             username: ${{ secrets.DOCKERHUB_USERNAME }}
             password: ${{ secrets.DOCKERHUB_TOKEN }}
 
-        - name: Build tool-classifier
+        # Set up cache for tool-classifier
+        - name: Build and push tool-classifier
           if: github.event.inputs.service_name=='tool-classifier'
-          run: docker build -t unstract/${{github.event.inputs.service_name}}:${{ github.event.inputs.tag }} ./tools/classifier
-        - name: Build tool-structure
+          uses: docker/build-push-action@v5
+          with:
+            context: ./tools/classifier
+            push: true
+            tags: unstract/${{github.event.inputs.service_name}}:${{ github.event.inputs.tag }}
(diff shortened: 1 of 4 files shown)


ritwik-g:
> Yes, it might break any of the associated cloud builds. It has been tested by dry running the workflow builds for these Github actions (without pushing to Dockerhub). However we'll have to properly test it with the push too

Why would it even break cloud builds? It is a different workflow all together right?

.github/workflows/docker-tools-build-push.yaml
ritwik-g, line 1: @chandrasekharan-zipstack these files changes looks good. I have one suggestion though not really related to your change but instead of having multiple if condition blocks I think we should be able to use matrix and single build block. Should be a simple change. If possible give it a try

.github/workflows/production-build.yaml
```diff
@@ -27,26 +27,86 @@ jobs:
...
+      - name: Set Dockerfile path and context
+        id: dockerfile-path
         run: |
-          DOCKER_VERSION_TAG=${{ github.event.release.tag_name  || github.event.inputs.tag}}
-          VERSION=$DOCKER_VERSION_TAG docker compose -f docker-compose.build.yaml build --no-cache ${{ matrix.service_name }}
-          docker push unstract/${{ matrix.service_name }}:$DOCKER_VERSION_TAG
-          if [ "${{ github.event_name }}" = "release" ]; then
-            echo "This workflow was triggered by a release event."
-            docker tag unstract/${{ matrix.service_name }}:$DOCKER_VERSION_TAG unstract/${{ matrix.service_name }}:latest
-            docker push unstract/${{ matrix.service_name }}:latest
```
ritwik-g: @chandrasekharan-zipstack I feel like the current approach you took to address this a bit convoluted and complex. Can we try either of the below options

*I think bake might be more aligned to new changes. But either should be fine*

- ```- name: Build and push with Bake
        uses: docker/bake-action@v5
        with:
          files: docker-compose.yml
          push: true
          set: |
            *.cache-from=type=gha
            *.cache-to=type=gha,mode=max```

- ```      - name: Build with Docker Compose
        run: |
          docker compose build \
            --builder default \
            --build-arg BUILDKIT_INLINE_CACHE=1
        env:
          COMPOSE_DOCKER_CLI_BUILD: 1
          DOCKER_BUILDKIT: 1
          BUILDX_GIT_LABELS: full
          COMPOSE_FILE: docker-compose.yml
          # Enable BuildKit cache exports
          COMPOSE_DOCKER_BUILD_CACHE_FROM: |
            type=gha
          COMPOSE_DOCKER_BUILD_CACHE_TO: |
            type=gha,mode=max```

.github/workflows/docker-tools-build-push.yaml
chandrasekharan-zipstack: Its slightly challenging given how the context / Dockerfile location is changing a bit. For now I've separated the build step alone. Any other improvements can follow later I feel

.github/workflows/production-build.yaml
chandrasekharan-zipstack: Made use of bake action. Seems like it has a useful build summary as well
https://github.com/Zipstack/unstract/actions/runs/15702835422

.github/workflows/docker-tools-build-push.yaml
```diff
@@ -30,23 +30,42 @@ jobs:
...
+              echo "context=./tools/text_extractor" >> $GITHUB_OUTPUT
+              echo "dockerfile=" >> $GITHUB_OUTPUT
+            elif [ "${{ github.event.inputs.service_name }}" == "tool-sidecar" ]; then
+              echo "context=." >> $GITHUB_OUTPUT
+              echo "dockerfile=docker/dockerfiles/tool-sidecar.Dockerfile" >> $GITHUB_OUTPUT
+            fi
+
+        # Build and push Docker image
+        - name: Build and push ${{ github.event.inputs.service_name }}
+          uses: docker/build-push-action@v5
```
hari-kuriakose, line 64: @chandrasekharan-zipstack Any reason why we are not using `docker/bake-action@v5` here too?

hari-kuriakose (approved):

Please check comments though.

.github/workflows/docker-tools-build-push.yaml
chandrasekharan-zipstack: I considered it initially but it gets a bit convoluted due to these inconsistencies
1. Bake only supports if there's a docker compose YAML present and the above mentioned tools need a build definition
2. Tool classifier and text-extractor don't have build definitions in `docker-compose-build.yaml`


Adding these tools also to the build docker compose is straightforward however we need to 
- check for any regressions 
- avoid building these tools wastefully. Maybe a profile can help for this

chandrasekharan-zipstack: Will try to make these changes if its easy, else can be handled through a separate PR

ritwik-g: @chandrasekharan-zipstack I don't think bake needs a `docker-compose yaml` always . I think it might support both Docker file and compose. So in your next PR even before trying to make it docker compose check if we can make it work with current docker file itself.

ritwik-g (approved):

.github/workflows/docker-tools-build-push.yaml
chandrasekharan-zipstack: This might be possible but we'll have to resolve the Dockerfile and context separately for the tools and sidecar - there's an inconsistency here also which makes it hard to call the steps directly

chandrasekharan-zipstack: Attempted [my suggestion](https://github.com/Zipstack/unstract/pull/1368#discussion_r2152195550) briefly in these 2 commits - however we need to update the tool dockerfiles also if we wish to update the context. Since the change is growing, we can take care of it separately
https://github.com/Zipstack/unstract/pull/1368/commits/9e7fe1cbe123dfd1c1466e1082b0b385b278d4bb
https://github.com/Zipstack/unstract/pull/1368/commits/2014ecae02fbf878c63b51f63dee2126e08d8440

PR #1369, opened 2025-06-18
UN-2476 [FEAT] Improved file deduplication in ETL/Task pipelines to ensure all unique files are processed

muhammad-ali-e:
## What

- Refactors file processing logic to check both **content hash** and **file path** for new file detection (`ETL/Task` pipelines only).
- Enhances `file_history` schema to store file path, adds unique constraints for combinations of hash/provider UUID and file path.
- Implements backfilling of file path in legacy file history records using data from `workflow_file_execution`.
- Updates helper methods and workflow logic to ensure accurate identification of new vs. previously processed files.
- Adds migration to support new constraints and fields.
- **Note:** This change only affects ETL and Task pipelines. API-based sources and destinations are excluded from the file path deduplication logic. The system maintains its legacy behavior for API connectors.

## Why

- The previous system could not always accurately identify new files, especially in cases where the same content appeared at different paths or in diffrent file names.
- File path was missing from `file_history` for legacy/older records, resulting in potential duplicate processing and poor deduplication.

## How

- Updates the logic to:
  - First check `file_history` using both content hash/provider UUID and file path.
  - If not found, falls back to `file_history` without file path, and then attempts to backfill the file path from `workflow_file_execution`.
  - Updates legacy records with missing file paths as they are encountered.
  - Refactors helper and workflow methods to use the new logic and constraints.
- Adds a database migration:
  - Adds `file_path` column to `file_history`.
  - Adds new and updated unique constraints for combinations including/excluding file path.

- No
- The logic is backward-compatible: legacy records are handled via fallback and are updated in-place as they are encountered.

## Database Migrations

- Adds `file_path` to `file_history`.
- Alters and adds several unique constraints to support deduplication with/without file path.

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


backend/workflow_manager/endpoint_v2/source.py
```diff
@@ -518,41 +512,86 @@ def _is_valid_pattern(self, file_name: str) -> bool:
...
+
+        # Always treat the file as new if neither identifier is available
+        if not file_hash.provider_file_uuid and not file_hash.file_hash:
+            return True
+
+        current_file_path = file_hash.file_path
+        file_history = self._get_file_history(file_hash=file_hash, workflow=workflow)
+
+        # No history or incomplete history means the file is new
+        if not file_history or not file_history.is_completed():
```
chandrasekharan-zipstack, line 530: @muhammad-ali-e if the file history is not completed, does it mean its already in progress? In that case should we log that there might be something in progress and stop the current execution?

backend/workflow_manager/endpoint_v2/source.py
```diff
@@ -518,41 +512,86 @@ def _is_valid_pattern(self, file_name: str) -> bool:
...
+            return True
+
+        # If content deduplication is enforced, skip processing regardless of file path
+        if self.use_content_deduplication_only:
+            self._log_file_skipped(current_file_path)
             return False
```
chandrasekharan-zipstack: @muhammad-ali-e since the value for this is set in the ctor itself, move this block before making a DB call to get the file history

backend/workflow_manager/workflow_v2/models/file_history.py
```diff
@@ -1,12 +1,14 @@
...
 
 from django.db import models
+from django.db.models import Q
 from utils.models.base_model import BaseModel
 
 from workflow_manager.workflow_v2.enums import ExecutionStatus
 from workflow_manager.workflow_v2.models.workflow import Workflow
 
 HASH_LENGTH = 64
+FILE_PATH_LENGTH = 1000
```
chandrasekharan-zipstack, line 11: How would the system behave if I attempted with a path > 1000 characters? Do we log and throw a valid error?

backend/workflow_manager/endpoint_v2/source.py
muhammad-ali-e: Not in the scope of this PR. . Handling duplicate or parallel ETL executions would require broader changes and is out of scope for now.

backend/workflow_manager/workflow_v2/models/file_history.py
muhammad-ali-e: The DB will raise an error if the file path exceeds the defined limit. I’ve also seen `255` used as the length in some other places — we should standardize this across the codebase. Increasing the limit (or removing it entirely if safe) would be a better long-term approach. We can handle this cleanup in a separate PR.

backend/workflow_manager/endpoint_v2/source.py
muhammad-ali-e: removed the falg since it not currently used

PR #1377, opened 2025-06-20
[MISC] Dockerfile enhancement to cache layers better

chandrasekharan-zipstack:
## What

- Refactored existing Dockerfiles as multi stage builds that add external dependencies first and then the project is installed
- Added option of `--locked` during `uv sync` 
```
--locked     Assert that the `uv.lock` will remain unchanged [env: UV_LOCKED=]
```
- MINOR: Added `.mypy-cache` to all dockerignores
## Why

- Faster builds while developing locally by caching occasionally changing external dependencies
- `--locked` helps us identify any cases where the lockfile needs to be updated instead of `--frozen` which just skips this check

## How

- Leveraged uv's docs and examples to implement the same

- No, only refactored the commands - however cloud builds haven't been explicitly checked yet

## Relevant Docs

- https://docs.astral.sh/uv/concepts/projects/sync/#automatic-lock-and-sync
- https://docs.astral.sh/uv/guides/integration/docker/#caching
- https://github.com/astral-sh/uv-docker-example/blob/main/multistage.Dockerfile

## Related Issues or PRs

-

## Notes on Testing

- Tested the builds locally and was able to execute tools and run prompts

## Checklist

### DIFF
--- a/docker/docker-compose.build.yaml
+++ b/docker/docker-compose.build.yaml
@@ -4,6 +4,7 @@ services:
     build:
       dockerfile: docker/dockerfiles/frontend.Dockerfile
       context: ..
+      target: production
   backend:
     image: unstract/backend:${VERSION}
     build:
--- a/docker/dockerfiles/frontend.Dockerfile
+++ b/docker/dockerfiles/frontend.Dockerfile
@@ -1,17 +1,32 @@
-# Use a smaller base image for the builder stage
-FROM node:16-alpine AS builder
+# Multi-stage build for both development and production
 
-# Build-time environment variables
+# Base stage with common setup
+FROM node:16-alpine AS base
 ENV BUILD_CONTEXT_PATH=frontend
-ENV REACT_APP_BACKEND_URL=""
-
-# Set the working directory inside the container
 WORKDIR /app
 
-# Copy only the files needed for installing dependencies
-COPY ${BUILD_CONTEXT_PATH}/package.json ${BUILD_CONTEXT_PATH}/package-lock.json ./
+### FOR DEVELOPMENT ###
+# Development stage for hot-reloading
+FROM base AS development
 
-# Install dependencies (this layer will be cached unless package.json or package-lock.json changes)
+# Copy only package files for dependency caching
+COPY ${BUILD_CONTEXT_PATH}/package.json ${BUILD_CONTEXT_PATH}/package-lock.json ./
+RUN npm install --ignore-scripts
+
+# Copy the rest of the application files
+COPY ${BUILD_CONTEXT_PATH}/ /app/
+
+EXPOSE 3000
+
+CMD ["npm", "start"]
+
+### FOR PRODUCTION ###
+# Builder stage for production build
+FROM base AS builder
+ENV REACT_APP_BACKEND_URL=""
+
+# Copy package files and install dependencies
+COPY ${BUILD_CONTEXT_PATH}/package.json ${BUILD_CONTEXT_PATH}/package-lock.json ./
 RUN npm install --ignore-scripts
 
 # Copy the rest of the application files
@@ -20,13 +35,9 @@ COPY ${BUILD_CONTEXT_PATH}/ .
 # Build the React app
 RUN npm run build
 
-# Use a smaller base image for the final stage
-FROM nginx:1.25-alpine
-
+# Production stage
+FROM nginx:1.25-alpine AS production
 LABEL maintainer="Zipstack Inc."
-
-# Remove the default NGINX configuration (if needed)
-# RUN rm /etc/nginx/conf.d/default.conf
 
 # Copy built assets from the builder stage
 COPY --from=builder /app/build /usr/share/nginx/html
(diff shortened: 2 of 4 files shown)


docker/dockerfiles/backend.Dockerfile
```diff
@@ -1,51 +1,69 @@
...
+
+# -----------------------------------------------
+# EXTERNAL DEPENDENCIES STAGE - This layer gets cached if external dependencies don't change
+# -----------------------------------------------
+FROM base AS ext-dependencies
 
-# Copy only what's needed for dependency installation first
-COPY pyproject.toml uv.lock /app/
+# Copy dependency-related files
+COPY ${BUILD_CONTEXT_PATH}/pyproject.toml ${BUILD_CONTEXT_PATH}/uv.lock ${BUILD_CONTEXT_PATH}/README.md ./
```
jaseemjaskp, line 46: @chandrasekharan-zipstack why this readme required?

docker/dockerfiles/backend.Dockerfile
```diff
@@ -1,51 +1,69 @@
...
 
-# Copy source code and dependencies
-COPY ${BUILD_CONTEXT_PATH}/ /app/
+# Copy local package dependencies
 COPY ${BUILD_PACKAGES_PATH}/ /unstract/
 
-# Install dependencies in one layer
-RUN uv sync --frozen && \
-    uv sync --group deploy && \
-    .venv/bin/python3 -m ensurepip --upgrade && \
```
jaseemjaskp, line 46: @chandrasekharan-zipstack Could you please explain why this section was removed? Have you tested this change without `ensurepip`?. I remember that tox failed due to some issue with pip. so make sure that also working

docker/dockerfiles/backend.Dockerfile
chandrasekharan-zipstack: `uv sync` expects this `README.md` to be there, otherwise it throws an error. Probably because we specified in the `pyproject.toml`

docker/dockerfiles/backend.Dockerfile
chandrasekharan-zipstack: I split this dependency installation in 2 layers to cache the external dependency installation. Regarding `tox`, it seems like the automated checks passed here so seems like its working

chandrasekharan-zipstack:
The Sonar cloud issue is related to `backend` and `runner` running as root user. Shall we try making the same changes on those files too or did we intentionally choose to run them as root user?
cc: @hari-kuriakose @ritwik-g @jaseemjaskp

ritwik-g:
> The Sonar cloud issue is related to `backend` and `runner` running as root user. Shall we try making the same changes on those files too or did we intentionally choose to run them as root user? cc: @hari-kuriakose @ritwik-g @jaseemjaskp

I think we left it root for some reason we can may be try changing it once we have stable version and some time to test.

docker/dockerfiles/platform.Dockerfile
```diff
@@ -22,42 +20,57 @@ ENV \
...
+FROM base AS ext-dependencies
 
-# Read and execute access to non-root user to avoid security hotspot
-# Write access to specific sub-directory need to be explicitly provided if required
-COPY --chown=${APP_USER} ${BUILD_CONTEXT_PATH} /app/
-# Copy local dependency packages
-COPY --chown=${APP_USER} ${BUILD_PACKAGES_PATH}/core /unstract/core
-COPY --chown=${APP_USER} ${BUILD_PACKAGES_PATH}/flags /unstract/flags
+# Copy dependency-related files
+COPY ${BUILD_CONTEXT_PATH}/pyproject.toml ${BUILD_CONTEXT_PATH}/uv.lock ${BUILD_CONTEXT_PATH}/README.md ./
```
ritwik-g: @chandrasekharan-zipstack isn't the --chown required in this case?

docker/dockerfiles/backend.Dockerfile
```diff
@@ -1,51 +1,69 @@
...
+    && apt-get --no-install-recommends install -y \
+       build-essential \
+       freetds-dev \
+       git \
+       libkrb5-dev \
+       libmagic-dev \
+       libssl-dev \
+       pkg-config \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/*
```
ritwik-g, line 33: @chandrasekharan-zipstack this means this layer will be cached and won't be built everytime right? Now is this behaviour gonna be true for local alone or will it be the case for our github builds as well? I am just wondering if we should actually use cache or let it be installed everytime if this get's cached in github also.

chandrasekharan-zipstack: This would get cached for the local builds - no doubt. 
- For github builds, since we use the gha cache it uses the cache now ([reference](https://github.com/Zipstack/unstract/actions/runs/15687239087/job/44193396128#step:8:217)) 
- Prior to this, all these steps used to repeat in the github builds ([reference](https://github.com/Zipstack/unstract/actions/runs/15749609945/job/44392211650#step:8:335))

docker/dockerfiles/platform.Dockerfile
chandrasekharan-zipstack: Good catch, seems like it worked without it. But just to be safe, I've added it again

PR #1391, opened 2025-07-02
UN-2579 [FIX] Fixed MIME type validation to auto-detect file types instead of checking content-type header

muhammad-ali-e:
## What
-  Fixed MIME type validation to auto-detect file types using Python Magic instead of relying solely on the Content-Type header sent by clients. When clients send
 -  `application/octet-stream` as Content-Type, the system now reads the actual file content to detect the correct MIME type.
 - refactored code to reduce `Cognitive Complexity`

  ## Why
  Recent MIME type validation changes caused existing client integrations to fail when they explicitly set Content-Type as `application/octet-stream`. This resulted in files being
  skipped with error messages like "**Skipping file sample.pdf due to Unsupported MIME type: Unsupported MIME type 'application/octet-stream'**". 

  ## How
  - Added `_detect_mime_type()` method that uses Python Magic to detect MIME type from file content
  - Modified file processing to detect MIME type from the first chunk of file data instead of trusting the Content-Type header
  - Maintained backward compatibility by still logging the original Content-Type header for debugging
  - Ensured proper error handling and logging throughout the process

  No, this PR fixes a regression and restores backward compatibility. It only changes how MIME types are detected (from header to content analysis) which is more reliable and allows
  previously working client integrations to function again.

  ## Database Migrations
  None required.

  ## Env Config
  None required.

  ## Relevant Docs
  None required.

  ## Related Issues or PRs
  UN-2579 - Bug handling MIME type for API deployment in the backend with backward compatibility

  ## Dependencies Versions
  None changed.

  ## Notes on Testing
  - Test with files uploaded using `application/octet-stream` Content-Type
  - Verify that actual file types (PDF, TXT, etc.) are correctly detected
  - Confirm that unsupported file types are still properly rejected
  - Test backward compatibility with existing client integrations

  ## Screenshots
  None applicable.

### DIFF
--- a/backend/workflow_manager/endpoint_v2/source.py
+++ b/backend/workflow_manager/endpoint_v2/source.py
@@ -69,11 +69,11 @@ class SourceConnector(BaseConnector):
     @classmethod
     def _detect_mime_type(cls, chunk: bytes, file_name: str) -> str:
         """Detect MIME type from file chunk using Python Magic.
-        
+
         Args:
             chunk (bytes): File chunk to analyze
             file_name (str): Name of the file being processed
-            
+
         Returns:
             str: Detected MIME type (may be unsupported)
         """
@@ -82,7 +82,7 @@ class SourceConnector(BaseConnector):
         logger.info(
             f"Detected MIME type using Python Magic: {mime_type} for file {file_name}"
         )
-        
+
         return mime_type
 
     def __init__(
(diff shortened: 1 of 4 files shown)


backend/workflow_manager/endpoint_v2/source.py
```diff
@@ -885,6 +904,117 @@ def load_file(self, input_file_path: str) -> tuple[str, BytesIO]:
...
+        file_hash = sha256()
+        first_iteration = True
+        mime_type = file.content_type
+
+        file.seek(0)
+
+        try:
+            for chunk in file.chunks(chunk_size=cls.READ_CHUNK_SIZE):
+                if first_iteration:
+                    mime_type = cls._detect_mime_type(chunk, file.name)
```
ritwik-g: @muhammad-ali-e do we check mimetype for every file? Or if the `file.content_type` is octet stream or missing? Reason for asking this, in past I recall @jagadeeswaran-zipstack working with magic library and often passing few bytes might not be enough but had to pass entire file content to detect some file mime types properly. So am bit worried this logic of relying on mime type from first chunks might not work in all cases

backend/workflow_manager/endpoint_v2/source.py
```diff
@@ -66,6 +66,25 @@ class SourceConnector(BaseConnector):
...
+
+        Args:
+            chunk (bytes): File chunk to analyze
+            file_name (str): Name of the file being processed
+
+        Returns:
+            str: Detected MIME type (may be unsupported)
+        """
+        # Primary MIME type detection using Python Magic
+        mime_type = magic.from_buffer(chunk, mime=True)
```
ritwik-g: @muhammad-ali-e if mime type is octet stream can we have a secondary check using magica? You can connect with @johnyrahul to see how he handled it. Why? In some cases the pdf file's mime type is not being captured properly by magic and shown as binary, magica can be used as a secondary check in such cases alone

backend/workflow_manager/endpoint_v2/source.py
```diff
@@ -919,65 +1049,61 @@ def add_input_file_to_api_storage(
...
+            file_hash, mime_type, success = cls._process_file_chunks(
+                file, file_storage, destination_path
+            )
 
-            file_history = None
-            if use_file_history:
-                file_history = FileHistoryHelper.get_file_history(
-                    workflow=workflow, cache_key=file_hash
+            # Handle unsupported files
+            if not success:
```
ritwik-g: might make sense to rename it like `process_chunks_success` or `mime_type_detected` which would be more meaning ful

backend/workflow_manager/endpoint_v2/source.py
```diff
@@ -885,6 +904,117 @@ def load_file(self, input_file_path: str) -> tuple[str, BytesIO]:
...
+        mime_type = file.content_type
+
+        file.seek(0)
+
+        try:
+            for chunk in file.chunks(chunk_size=cls.READ_CHUNK_SIZE):
+                if first_iteration:
+                    mime_type = cls._detect_mime_type(chunk, file.name)
+                    if not AllowedFileTypes.is_allowed(mime_type):
+                        raise UnsupportedMimeTypeError(
```
ritwik-g: NIT: @muhammad-ali-e in case of unsupported wondering if we even need to raise this error or we can simply return `return "", mime_type, False`. Either should be fine. Just wondering which might be better. I feel like since we are only handling this specific error the try except really seems unnecessary and we could may be return from here. Not sure which might be the better practise

ritwik-g (commented):
Overall refactoring done is a good approach. I think the code will be easier to maintain

backend/workflow_manager/endpoint_v2/source.py
chandrasekharan-zipstack: @ritwik-g is it a necessity to have both approaches? Maybe we could just use magika alone. I remember @shuveb suggested this too once

backend/workflow_manager/endpoint_v2/source.py
```diff
@@ -885,6 +904,117 @@ def load_file(self, input_file_path: str) -> tuple[str, BytesIO]:
 
         return os.path.basename(input_file_path), file_stream
 
+    @classmethod
+    def _process_file_chunks(
```
chandrasekharan-zipstack: NIT: @muhammad-ali-e provide a more suitable name. `_process_file_chunks` is a bit vague. Also we aren't accepting file chunks here rather we chunk it ourselves

backend/workflow_manager/endpoint_v2/source.py
```diff
@@ -885,6 +904,117 @@ def load_file(self, input_file_path: str) -> tuple[str, BytesIO]:
...
+
+    @classmethod
+    def _check_duplicate_file(
+        cls, file_hash: str, unique_file_hashes: set[str], file_name: str, workflow_log
+    ) -> bool:
+        """Check if file is duplicate and log if needed."""
+        if file_hash in unique_file_hashes:
+            log_message = f"Skipping file '{file_name}' — duplicate detected within the current request. Already staged for processing."
+            workflow_log.log_info(logger=logger, message=log_message)
+            return True
```
chandrasekharan-zipstack, line 1037: @muhammad-ali-e didn't we recently discuss to not skip duplicates based on file hash / content and allow its processing?

backend/workflow_manager/endpoint_v2/source.py
```diff
@@ -885,6 +904,117 @@ def load_file(self, input_file_path: str) -> tuple[str, BytesIO]:
...
+
+        unique_file_hashes.add(file_hash)
+        return False
+
+    @classmethod
+    def _get_execution_status(
```
chandrasekharan-zipstack: NIT: Rename to `_is_execution_completed()` instead

ritwik-g (approved):
**Automated Approval Report**  
All bot-reviewed issues resolved  
  
✅ APPROVED

muhammad-ali-e:
dffered

PR #1401, opened 2025-07-03
UN-2247 [FEAT] Added graceful shutdown support for tool containers with signal handling and configurable timeout 

kirtimanmishrazipstack:
## What

- Added graceful shutdown support for structure tool container with signal handling and configurable timeout

## Why

- Tool container should end gracefully

## How

- Changes in `tool/sctructure` to accept graceful shutdown
- Changes in `unstract/runner` to send SIGTERM to `structure-tool` 

- No. It is is related to tool which accepts SIGTERM 

## Database Migrations

- N/A

## Env Config

- Yes. Env's related to TTL shutdown for tool 

## Relevant Docs

- N/A

## Related Issues or PRs

- Base JIRA: https://zipstack.atlassian.net/browse/UN-2247
- JIRA: https://zipstack.atlassian.net/browse/UN-2601

## Dependencies Versions

-

## Notes on Testing

- Tested structure-tool by running locally in debug mode. When press ctrl+c, tool gracefully shutdown.
- If there is any existing calls to llm, tool won't end. This issue has taken care of.
- Tested tool while running on OSS.
- When runner calls tool, and tool end. Before ending tool gracefully shutdown.
- When I directly send SIGTERM to tool container manually, it also gracefully shutdown. Seen in screenshot 3. 

## Screenshots

- graceful_shutdown_1

- graceful_shutdown_3 png

- graceful_shutdown_3 png

## Checklist


runner/src/unstract/runner/clients/docker_client.py
```diff
@@ -48,6 +48,29 @@ def wait_until_stop(
...
+        """Gracefully stop the container with SIGTERM and timeout.
+
+        Args:
+            timeout (int): Grace period in seconds before force kill (default: 300)
+        """
+        if not self.container:
+            return
+
+        try:
+            self.logger.info(f"Sending SIGTERM to container {self.container.id[:12]}")
```
chandrasekharan-zipstack, line 61: NIT: @kirtimanmishrazipstack why do we print the first 12 characters of the container name alone?

ritwik-g (changes requested):
@kirtimanmishrazipstack I think this change is too complicated. The change should have been a simple SIGTERM trap in the structure tool code so that it wouldn't kill the container. That's it.

- The runner need not have any changes (graceful stop cannot be send by us, k8s will send it).
- We don't need to check if LLM is in progress or anything.

Something like below in tool's main thread is all we need ( I think we should add this in the SDK so it will be applicable to all tools).

```
def handle_sigterm(signum, frame):
    print("Received SIGTERM! Cleaning up...")
    # Do any cleanup here

# Register the handler for SIGTERM
signal.signal(signal.SIGTERM, handle_sigterm)
```

https://zipstack.atlassian.net/browse/UN-2247?focusedCommentId=21989

cc: @hari-kuriakose

runner/src/unstract/runner/clients/docker_client.py
```diff
@@ -48,6 +48,29 @@ def wait_until_stop(
 
         return self.container.wait()
 
+    def graceful_stop(self, timeout: int = 300) -> None:
```
chandrasekharan-zipstack, line 51: NIT: This function seems to kill the container in case of an exception during the graceful shutdown. I suggest removing the prefix of `graceful_` in all these function names. Just `stop()` or `shutdown()` should do

runner/src/unstract/runner/clients/docker_client.py
```diff
@@ -48,6 +48,29 @@ def wait_until_stop(
...
+        if not self.container:
+            return
+
+        try:
+            self.logger.info(f"Sending SIGTERM to container {self.container.id[:12]}")
+            # Send SIGTERM signal to allow graceful shutdown
+            self.container.stop(timeout=timeout)
+            self.logger.info(f"Container {self.container.id[:12]} stopped gracefully")
+        except Exception as stop_error:
+            self.logger.error(f"Failed to gracefully stop container: {stop_error}")
```
chandrasekharan-zipstack, line 66:
```suggestion
            self.logger.error(f"Failed to gracefully stop container: {stop_error}, attempting to kill it")
```

runner/sample.env
```diff
@@ -57,3 +57,6 @@ TOOL_EXECUTION_CACHE_TTL_IN_SECOND=86400 # 24 Hours
 
 # File Execution Tracker
 FILE_EXECUTION_TRACKER_TTL_IN_SECOND=18000 # 5 hours
+
+# Graceful shutdown timeout period in seconds (Default: 300, Max: 7200 = 2 hours)
+GRACEFUL_SHUTDOWN_PERIOD=7200
```
chandrasekharan-zipstack, line 62: NIT: Prefix it with `CONTAINER_` or `TOOL_` to indicate that it refers to the spawned container explicitly. Imagine coming across this env months later, you wouldn't want to confuse this with a setting meant for the runner service itself

runner/src/unstract/runner/clients/docker_client.py
```diff
@@ -330,6 +353,25 @@ def wait_for_container_stop(
             return None
         return container.wait_until_stop(main_container_status)
 
+    def graceful_stop_container(
```
chandrasekharan-zipstack, line 356: @kirtimanmishrazipstack this function seems unnecessary, can't you read the timeout env and stop the container in the `ContainerInterface.stop()` / `ContainerInterface.graceful_stop()`

runner/src/unstract/runner/clients/docker_client.py
```diff
@@ -358,6 +400,10 @@ def remove_container_by_name(
         """Remove a container by its name."""
         try:
             container = self.client.containers.get(container_name)
+            # Try graceful shutdown first if container is running
+            if container.status == "running":
+                docker_container = DockerContainer(container, self.logger)
+                self.graceful_stop_container(docker_container)
```
chandrasekharan-zipstack, line 406: @kirtimanmishrazipstack consider defining just 1 function to stop the container within the `ContainerInterface` called `stop()`. This can be directly invoked here instead of creating `DockerContainer` instances here

runner/src/unstract/runner/clients/docker_client.py
```diff
@@ -375,6 +421,10 @@ def remove_container_by_name(
             sidecar_name = Utils.get_sidecar_container_name(container_name)
             try:
                 sidecar = self.client.containers.get(sidecar_name)
+                # Try graceful shutdown first if sidecar is running
+                if sidecar.status == "running":
+                    docker_sidecar = DockerContainer(sidecar, self.logger)
+                    self.graceful_stop_container(docker_sidecar)
```
chandrasekharan-zipstack, line 427: Same as the above comment

runner/src/unstract/runner/clients/interface.py
```diff
@@ -30,6 +30,15 @@ def cleanup(self, client: ContainerClientInterface | None = None) -> None:
         """Stops and removes the running container."""
         pass
 
+    @abstractmethod
+    def graceful_stop(self, timeout: int = 300) -> None:
```
chandrasekharan-zipstack, line 34: Related to above comments, rename to `stop()`
```suggestion
    def stop(self, timeout: int = 300) -> None:
```

runner/src/unstract/runner/clients/interface.py
```diff
@@ -143,3 +152,19 @@ def remove_container_by_name(
             bool: True if container was removed or not found, False if error.
         """
         pass
+
+    @abstractmethod
+    def graceful_stop_container(
```
chandrasekharan-zipstack, line 157: Related to other comments, remove this if it doesn't add any value

runner/src/unstract/runner/runner.py
```diff
@@ -286,6 +286,8 @@ def run_command(self, command: str) -> Any | None:
                 f"Failed to run docker container: {e}", stack_info=True, exc_info=True
             )
         if container:
+            # Gracefully stop container before cleanup
+            self.client.graceful_stop_container(container)
```
chandrasekharan-zipstack, line 290: @kirtimanmishrazipstack I think you were forced to define a method in the client class due to this. I think `container.cleanup()` and `container.remove()` are similar. Attempt to unify them and invoke the `stop()` from there

tools/structure/src/main.py
```diff
@@ -16,7 +17,41 @@
 from unstract.sdk.tool.base import BaseTool
 from unstract.sdk.tool.entrypoint import ToolEntrypoint
 
-logger = logging.getLogger(__name__)
```
chandrasekharan-zipstack, line 19: NIT: Hope this isn't used anywhere directly

tools/structure/src/main.py
```diff
@@ -16,7 +17,41 @@
...
+
+def signal_handler(signum: int, _frame: Any) -> None:
+    """Handle SIGTERM/SIGINT signal - acknowledge but complete current LLM processing.
+
+    Args:
+        signum: The signal number received
+        _frame: Current stack frame (unused)
+    """
+    signal_name = "SIGTERM"
+    print("Received signal %s (%s). Graceful shutdown initiated...", signum, signal_name)
```
chandrasekharan-zipstack, line 32: @kirtimanmishrazipstack consider using a logger here. Ensure that it gets collected and displayed by the sidecar

chandrasekharan-zipstack: Also the signal_name would always be `SIGTERM` irrespective of any other signals that might be caught

tools/structure/src/main.py
```diff
@@ -119,6 +154,14 @@ def run(
         input_file: str,
         output_dir: str,
     ) -> None:
+        # Verify signal handlers are active
+        current_sigterm = signal.signal(signal.SIGTERM, signal.getsignal(signal.SIGTERM))
+        self.stream_log(f"Active SIGTERM handler: {current_sigterm}")
+        self.stream_log(
+            f"SIGNAL_VERIFICATION: SIGTERM handler active: {current_sigterm == signal_handler}",
+            flush=True,
```
chandrasekharan-zipstack, line 162: All this would appear in the user's logs and it might not be of use to them. I think we should only log this to our services / sidecar and not collect it as part of the execution's logs. Print with a normal `logger` instead. Also consider accepting code rabbit's suggestion. I find these logs a bit confusing too. We could mark it as a DEBUG log to avoid noise

tools/structure/src/main.py
```diff
@@ -219,21 +262,31 @@ def run(
...
-            usage_kwargs=usage_kwargs,
-            run_id=self.file_execution_id,
-            tool_settings=tool_settings,
-            extract_file_path=tool_data_dir / SettingsKeys.EXTRACT,
-            tool=self,
-            execution_run_data_folder=str(execution_run_data_folder),
-        )
+        shutdown_acknowledged.set()
+        if shutdown_acknowledged.is_set():
+            self.stream_log(
```
chandrasekharan-zipstack, line 273: Same as before, we needn't complicate logs for the user

tools/structure/src/main.py
```diff
@@ -219,21 +262,31 @@ def run(
...
+                enable_highlight=is_highlight_enabled,
+                usage_kwargs=usage_kwargs,
+                run_id=self.file_execution_id,
+                tool_settings=tool_settings,
+                extract_file_path=tool_data_dir / SettingsKeys.EXTRACT,
+                tool=self,
+                execution_run_data_folder=str(execution_run_data_folder),
+            )
+        finally:
+            shutdown_acknowledged.clear()
```
chandrasekharan-zipstack, line 289: @kirtimanmishrazipstack add this shutdown related code as a decorator and wrap it for the necessary functions where we might want to wait for processing to complete. The current implementation seems to pollute the `run()` method a bit with this logic

tools/structure/tests/test_graceful_shutdown.py
chandrasekharan-zipstack, line 1: @kirtimanmishrazipstack how can this test file be run?
1. Document the steps in a `README.md`
2. It'll be useful if these tests can be configured to run on changes to this tool while raising a PR as part of our existing action

PR #1411, opened 2025-07-06
UN-2566  [FIX] JSON array wrapping fix

Deepak-Kesavan:
## What

- Some JSON output wraps in an array due to special characters.

## Why

-

## How

- FIxed by adding ensure_ascii=False

- No

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of non-ASCII characters during JSON parsing, ensuring more accurate processing of Unicode data.


prompt-service/src/unstract/prompt_service/services/answer_prompt.py
```diff
@@ -255,7 +255,10 @@ def handle_json(
 
             # Heuristic: if wrapping only added '[' and ']', len(b) - len(a) == 2 → original was valid, use 'a'
             # Otherwise, fallback to 'b' which likely fixed multiple items or invalid top-level structure
-            parsed_data = a if len(json.dumps(b)) - len(json.dumps(a)) == 2 else b
+            dump_a = json.dumps(a, ensure_ascii=False)
+            dump_b = json.dumps(b, ensure_ascii=False)
```
chandrasekharan-zipstack, line 259: NIT: Instead of `a` and `b` we should use more meaningful names

prompt-service/src/unstract/prompt_service/services/answer_prompt.py
```diff
@@ -255,7 +255,10 @@ def handle_json(
...
+            ARRAY_WRAP_DELTA = 2  # '[' and ']'
+            parsed_data = a if len(dump_b) - len(dump_a) == ARRAY_WRAP_DELTA else b
```
chandrasekharan-zipstack, line 261: @Deepak-Kesavan will `len(dump_b) - len(dump_a)` always be positive?
Can repair JSON remove any characters by any chance - should we make it this instead?
```suggestion
            parsed_data = a if abs(len(dump_b) - len(dump_a)) == ARRAY_WRAP_DELTA else b
```

chandrasekharan-zipstack (approved):
LGTM - check the comments though

prompt-service/src/unstract/prompt_service/services/answer_prompt.py
Deepak-Kesavan: No @chandrasekharan-zipstack . It will always be positive. We won't be needing absolute func here.

prompt-service/src/unstract/prompt_service/services/answer_prompt.py
Deepak-Kesavan: What do you suggest? its just some internal values we use to compare parsed_data. Also I added comment on both variables what we are trying to do.

PR #1421, opened 2025-07-10
UN-2603 Changes to support pushing from API to HITL

Deepak-Kesavan:
## What

-

## Why

-

## How

-

-

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/backend/api_v2/api_deployment_views.py
+++ b/backend/api_v2/api_deployment_views.py
@@ -65,6 +65,8 @@ class DeploymentExecution(views.APIView):
         use_file_history = serializer.validated_data.get(ApiExecution.USE_FILE_HISTORY)
         tag_names = serializer.validated_data.get(ApiExecution.TAGS)
         llm_profile_id = serializer.validated_data.get(ApiExecution.LLM_PROFILE_ID)
+        push_to_hitl = serializer.validated_data.get(ApiExecution.PUSH_TO_HITL)
+        hitl_queue_name = serializer.validated_data.get(ApiExecution.HITL_QUEUE_NAME)
 
         response = DeploymentHelper.execute_workflow(
             organization_name=org_name,
@@ -76,8 +78,12 @@ class DeploymentExecution(views.APIView):
             use_file_history=use_file_history,
             tag_names=tag_names,
             llm_profile_id=llm_profile_id,
+            push_to_hitl=push_to_hitl,
+            hitl_queue_name=hitl_queue_name,
         )
+        logger.info(f"Workflow execution response: {response}")
         if "error" in response and response["error"]:
+            logger.error(f"Workflow execution error: {response}")
             return Response(
                 {"message": response},
                 status=status.HTTP_422_UNPROCESSABLE_ENTITY,
--- a/backend/api_v2/deployment_helper.py
+++ b/backend/api_v2/deployment_helper.py
@@ -149,6 +149,8 @@ class DeploymentHelper(BaseAPIKeyValidator):
         use_file_history: bool = False,
         tag_names: list[str] = [],
         llm_profile_id: str | None = None,
+        push_to_hitl: bool = False,
+        hitl_queue_name: str | None = None,
     ) -> ReturnDict:
         """Execute workflow by api.
 
@@ -160,12 +162,18 @@ class DeploymentHelper(BaseAPIKeyValidator):
                 processed files. Defaults to False
             tag_names (list(str)): list of tag names
             llm_profile_id (str, optional): LLM profile ID for overriding tool settings
+            push_to_hitl (bool): Flag to push files to manual review queue
+            queue (str, optional): Custom queue name for manual review
 
         Returns:
             ReturnDict: execution status/ result
         """
         workflow_id = api.workflow.id
         pipeline_id = api.id
+        # Use API name as default document class if hitl_queue_name not provided
+        if push_to_hitl and not hitl_queue_name:
+            hitl_queue_name = api.api_name
+        logger.info(f"API execution: push_to_hitl={push_to_hitl}, hitl_queue_name={hitl_queue_name}, api_name={api.api_name}")
         tags = Tag.bulk_get_or_create(tag_names=tag_names)
         workflow_execution = WorkflowExecutionServiceHelper.create_workflow_execution(
             workflow_id=workflow_id,
@@ -193,6 +201,9 @@ class DeploymentHelper(BaseAPIKeyValidator):
                 queue=CeleryQueue.CELERY_API_DEPLOYMENTS,
                 use_file_history=use_file_history,
                 llm_profile_id=llm_profile_id,
+                push_to_hitl=push_to_hitl,
+                hitl_queue_name=hitl_queue_name,
+                api_name=api.api_name,
             )
             result.status_api = DeploymentHelper.construct_status_endpoint(
                 api_endpoint=api.api_endpoint, execution_id=execution_id
--- a/backend/api_v2/serializers.py
+++ b/backend/api_v2/serializers.py
@@ -117,6 +117,10 @@ class ExecutionRequestSerializer(TagParamsSerializer):
             e.g:'tag1,tag2-name,tag3_name'
         llm_profile_id (str): UUID of the LLM profile to override the default profile.
             If not provided, the default profile will be used.
+nwhatok        push_to_hitl (bool): Flag to push files to manual review queue (HITL).
+            Defaults to False.
+        hitl_queue_name (str, optional): Document class name for manual review queue.
+            If not provided, uses API name as document class.
     """
 
     MAX_FILES_ALLOWED = 32
@@ -128,6 +132,8 @@ class ExecutionRequestSerializer(TagParamsSerializer):
     include_metrics = BooleanField(default=False)
     use_file_history = BooleanField(default=False)
     llm_profile_id = CharField(required=False, allow_null=True, allow_blank=True)
+    push_to_hitl = BooleanField(default=False)
+    hitl_queue_name = CharField(required=False, allow_null=True, allow_blank=True)
     files = ListField(
         child=FileField(),
         required=True,
@@ -175,6 +181,7 @@ class ExecutionRequestSerializer(TagParamsSerializer):
             raise ValidationError("You can only use profiles that you own")
 
         return value
+
 
 
 class ExecutionQuerySerializer(Serializer):
--- a/backend/workflow_manager/endpoint_v2/destination.py
+++ b/backend/workflow_manager/endpoint_v2/destination.py
@@ -59,6 +59,9 @@ class DestinationConnector(BaseConnector):
         workflow_log: WorkflowLog,
         use_file_history: bool,
         file_execution_id: str | None = None,
+        push_to_hitl: bool = False,
+        hitl_queue_name: str | None = None,
+        api_name: str | None = None,
     ) -> None:
         """Initialize a DestinationConnector object.
 
@@ -76,6 +79,9 @@ class DestinationConnector(BaseConnector):
         )
         self.workflow_log = workflow_log
         self.use_file_history = use_file_history
+        self.push_to_hitl = push_to_hitl
+        self.hitl_queue_name = hitl_queue_name
+        self.api_name = api_name
         self.workflow = workflow
 
     def _get_endpoint_for_workflow(
@@ -162,6 +168,20 @@ class DestinationConnector(BaseConnector):
         file_execution_id: str,
     ) -> bool:
         """Handles HITL processing, returning True if data was pushed to the queue."""
+        logger.info(f"_handle_hitl called: push_to_hitl={self.push_to_hitl}, hitl_queue_name={self.hitl_queue_name}, api_name={self.api_name}")
+        # Check if API deployment requested HITL override
+        if self.push_to_hitl:
+            logger.info(f"API HITL override: pushing to queue for file {file_name}")
+            self._push_data_to_queue(
+                file_name=file_name,
+                workflow=workflow,
+                input_file_path=input_file_path,
+                file_execution_id=file_execution_id,
+            )
+            logger.info(f"Successfully pushed {file_name} to queue")
+            return True
+
+        # Otherwise use existing workflow-based HITL logic
         execution_result = self.get_tool_execution_result()
         if WorkflowUtil.validate_db_rule(
             execution_result, workflow, file_hash.file_destination
@@ -218,7 +238,17 @@ class DestinationConnector(BaseConnector):
             ):
                 self.insert_into_db(input_file_path=input_file_path)
         elif connection_type == WorkflowEndpoint.ConnectionType.API:
-            tool_execution_result = self.get_tool_execution_result(file_history)
+            logger.info(f"API connection type detected for file {file_name}")
+            # Check for HITL (Manual Review Queue) override for API deployments
+            if not self._handle_hitl(
+                file_name=file_name,
+                file_hash=file_hash,
+                workflow=workflow,
+                input_file_path=input_file_path,
+                file_execution_id=file_execution_id,
+            ):
+                logger.info(f"No HITL override, getting tool execution result for {file_name}")
+                tool_execution_result = self.get_tool_execution_result(file_history)
         elif connection_type == WorkflowEndpoint.ConnectionType.MANUALREVIEW:
             self._push_data_to_queue(
                 file_name,
@@ -647,6 +677,9 @@ class DestinationConnector(BaseConnector):
             workflow_id=self.workflow.id,
             execution_id=self.execution_id,
             use_file_history=self.use_file_history,
+            push_to_hitl=self.push_to_hitl,
+            hitl_queue_name=self.hitl_queue_name,
+            api_name=self.api_name,
         )
 
     @classmethod
@@ -661,6 +694,7 @@ class DestinationConnector(BaseConnector):
         Returns:
             DestinationConnector: New instance
         """
+        logger.info(f"Creating DestinationConnector from config: push_to_hitl={config.push_to_hitl}, hitl_queue_name={config.hitl_queue_name}, api_name={config.api_name}")
         # Reconstruct workflow
         workflow = Workflow.objects.get(id=config.workflow_id)
 
@@ -671,6 +705,9 @@ class DestinationConnector(BaseConnector):
             workflow_log=workflow_log,
             use_file_history=config.use_file_history,
             file_execution_id=config.file_execution_id,
+            push_to_hitl=config.push_to_hitl,
+            hitl_queue_name=config.hitl_queue_name,
+            api_name=config.api_name,
         )
 
         return destination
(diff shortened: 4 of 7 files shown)


backend/api_v2/api_deployment_views.py
```diff
@@ -65,6 +65,8 @@ def post(
         use_file_history = serializer.validated_data.get(ApiExecution.USE_FILE_HISTORY)
         tag_names = serializer.validated_data.get(ApiExecution.TAGS)
         llm_profile_id = serializer.validated_data.get(ApiExecution.LLM_PROFILE_ID)
+        push_to_hitl = serializer.validated_data.get(ApiExecution.PUSH_TO_HITL)
+        hitl_queue_name = serializer.validated_data.get(ApiExecution.HITL_QUEUE_NAME)
```
chandrasekharan-zipstack: @Deepak-Kesavan why do we need both these parameters to determine if it needs to be pushed? Why can't we take a decision based on `hitl_queue_name` alone? If its not present, we don't and if it is present we push to that queue

Deepak-Kesavan: @chandrasekharan-zipstack I also had the same question when implementing. Will check with @shuveb and update it based on the decision made.

backend/api_v2/api_deployment_views.py
```diff
@@ -76,8 +77,11 @@ def post(
             use_file_history=use_file_history,
             tag_names=tag_names,
             llm_profile_id=llm_profile_id,
+            hitl_queue_name=hitl_queue_name,
         )
+        logger.info(f"Workflow execution response: {response}")
         if "error" in response and response["error"]:
+            logger.error(f"Workflow execution error: {response}")
```
chandrasekharan-zipstack: NIT: These logs lack context on where its coming from - you could highlight that its related to an API execution. Also remove this log if any data is going to be logged

backend/api_v2/deployment_helper.py
```diff
@@ -193,6 +201,9 @@ def execute_workflow(
                 queue=CeleryQueue.CELERY_API_DEPLOYMENTS,
                 use_file_history=use_file_history,
                 llm_profile_id=llm_profile_id,
+                push_to_hitl=bool(hitl_queue_name),
```
chandrasekharan-zipstack: @Deepak-Kesavan based on the condition above, this would always be true since the `hitl_queue_name` always gets initialized

backend/api_v2/models.py
```diff
@@ -87,16 +87,33 @@ def __str__(self) -> str:
...
+        # For new deployments, check one API per workflow constraint
+        if self.pk is None and self.is_active:
+            existing_active_count = APIDeployment.objects.filter(
+                workflow=self.workflow, is_active=True, organization=self.organization
+            ).count()
+
+            if existing_active_count > 0:
+                raise ValidationError(
+                    "This workflow already has an active API deployment. Only one API deployment per workflow is allowed."
+                )
```
chandrasekharan-zipstack: @Deepak-Kesavan move this to the serializer, that's where we should validate the data and throw such errors for new records that we attempt to create. You wouldn't have such checks based on primary key in that case
EDIT: I notice that it was added in the serializer already, then why is this needed?

backend/api_v2/serializers.py
```diff
@@ -44,6 +45,12 @@ class Meta:
                 "This API name is already in use. Please select a different name."
             ),
         },
+        "unique_workflow_per_organization": {
```
chandrasekharan-zipstack:
```suggestion
        "unique_api_per_workflow": {
```
NIT: The name is misleading. Also the above 2 values are repeated - check if its an issue we made with error handling
```
This API name is already in use. Please select a different name
```

backend/api_v2/serializers.py
```diff
@@ -56,6 +63,27 @@ def validate_api_name(self, value: str) -> str:
...
+            # Check if this workflow already has an active API deployment
+            existing_active_count = APIDeployment.objects.filter(
+                workflow=workflow, is_active=True
+            ).count()
+
+            # If there's already an active API deployment, prevent creating a new one
+            if existing_active_count > 0:
+                raise ValidationError(
+                    {
+                        "workflow": "This workflow already has an active API deployment. Only one API deployment per workflow is allowed."
```
chandrasekharan-zipstack, line 75: NIT: Did you intend on using the unique error message map? The messages seem to be duplicated if not. Also why have we raised it as a dict - how does it get rendered in case of an error?

backend/api_v2/serializers.py
```diff
@@ -128,6 +158,33 @@ class ExecutionRequestSerializer(TagParamsSerializer):
     include_metrics = BooleanField(default=False)
     use_file_history = BooleanField(default=False)
     llm_profile_id = CharField(required=False, allow_null=True, allow_blank=True)
+    hitl_queue_name = CharField(required=False, allow_null=True, allow_blank=True)
+
+    def validate_hitl_queue_name(self, value: str) -> str:
+        """Validate queue name format: a-z0-9-_ with no repeating/starting/ending _-"""
+        if not value:
+            return value
```
chandrasekharan-zipstack, line 160: @Deepak-Kesavan this would return NULL - is it valid?

backend/api_v2/serializers.py
```diff
@@ -128,6 +158,33 @@ class ExecutionRequestSerializer(TagParamsSerializer):
...
+        if value.startswith(("_", "-")) or value.endswith(("_", "-")):
+            raise ValidationError(
+                "Queue name cannot start or end with underscore or hyphen."
+            )
+
+        # Check no repeating _ or -
+        if "__" in value or "--" in value or "_-" in value or "-_" in value:
+            raise ValidationError(
+                "Queue name cannot have repeating underscores or hyphens."
+            )
```
chandrasekharan-zipstack: @Deepak-Kesavan why do we have this constraint?

backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -162,6 +168,22 @@ def _handle_hitl(
         file_execution_id: str,
     ) -> bool:
         """Handles HITL processing, returning True if data was pushed to the queue."""
+        logger.info(
+            f"_handle_hitl called: push_to_hitl={self.push_to_hitl}, hitl_queue_name={self.hitl_queue_name}, api_name={self.api_name}"
+        )
+        # Check if API deployment requested HITL override
+        if self.push_to_hitl:
```
chandrasekharan-zipstack: NIT: @Deepak-Kesavan do we need this flag, can't we decide based on whether the queue name is set or not?

backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -218,7 +240,19 @@ def handle_output(
             ):
                 self.insert_into_db(input_file_path=input_file_path)
         elif connection_type == WorkflowEndpoint.ConnectionType.API:
-            tool_execution_result = self.get_tool_execution_result(file_history)
+            logger.info(f"API connection type detected for file {file_name}")
+            # Check for HITL (Manual Review Queue) override for API deployments
+            if not self._handle_hitl(
```
chandrasekharan-zipstack: NIT: For functions return boolean, try to name it as a question of the form `_is_handle_hitl()` or `_should_handle_hitl()` - this small change improves readability in quickly helping you understand what it might do without requiring additional comments before it

backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -737,3 +826,103 @@ def _push_to_queue(
...
+            file_name: Name of the file for logging purposes
+
+        Returns:
+            Base64 encoded file content or None if reading fails
+        """
+        if not input_file_path:
+            logger.warning(f"No input_file_path provided for: {file_name}")
+            return None
+
+        logger.debug(
```
chandrasekharan-zipstack: NIT: Add context that its for HITL otherwise the logs can be confusing

backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -737,3 +826,103 @@ def _push_to_queue(
...
+            else:
+                logger.warning(
+                    f"File does not exist in filesystem storage: {input_file_path}"
+                )
+        except Exception as fs_e:
+            logger.warning(
+                f"Filesystem abstraction failed: {fs_e}, trying direct file access"
+            )
+
+        # Fallback to direct file access
```
chandrasekharan-zipstack: @Deepak-Kesavan why do we do this?

chandrasekharan-zipstack: NIT: Add it as a separate function that we invoke only in case of an exception in the previous method. The file system interface should ideally serve as an abstraction for accessing remote / local files too. Basically the same code should have been enough to do this conditionally - repetition here indicates that we've gone wrong in the LLD somewhere

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -422,6 +428,9 @@ def execute_workflow_async(
         queue: str | None = None,
         use_file_history: bool = True,
         llm_profile_id: str | None = None,
+        push_to_hitl: bool = False,
+        hitl_queue_name: str | None = None,
+        api_name: str | None = None,
```
chandrasekharan-zipstack: NIT: All this should be kwargs - for an ETL based / workflow based execution none of these params really make sense

backend/api_v2/deployment_helper.py
Deepak-Kesavan: @chandrasekharan-zipstack Will remove the above condition. It is not the expected behavior.

backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -153,15 +155,29 @@ def validate(self) -> None:
...
         file_hash: FileHash,
         workflow: Workflow,
         input_file_path: str,
         file_execution_id: str,
     ) -> bool:
-        """Handles HITL processing, returning True if data was pushed to the queue."""
+        """Determines if HITL processing should be performed, returning True if data was pushed to the queue."""
+        logger.info(f"HITL check called: hitl_queue_name={self.hitl_queue_name}")
+        # Check if API deployment requested HITL override
+        if self.hitl_queue_name:
```
muhammad-ali-e, line 171: NIT: looks like some refactor is possible here.

backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -153,15 +155,29 @@ def validate(self) -> None:
...
         self,
         file_name: str,
         file_hash: FileHash,
         workflow: Workflow,
         input_file_path: str,
         file_execution_id: str,
     ) -> bool:
-        """Handles HITL processing, returning True if data was pushed to the queue."""
+        """Determines if HITL processing should be performed, returning True if data was pushed to the queue."""
+        logger.info(f"HITL check called: hitl_queue_name={self.hitl_queue_name}")
```
muhammad-ali-e: I think this log not need to be here since it might be NOne for `self.hitl_queue_name`

backend/workflow_manager/endpoint_v2/destination.py
muhammad-ali-e, line 171: And the condition is not self explanatory.  Why we checking  `hitl_queue_name`  ? Is it for API or any specific case?

backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -671,10 +703,35 @@ def from_config(
             workflow_log=workflow_log,
             use_file_history=config.use_file_history,
             file_execution_id=config.file_execution_id,
+            hitl_queue_name=config.hitl_queue_name,
         )
 
         return destination
 
+    def _get_queue_name(self) -> str:
```
muhammad-ali-e: please re check are these oss feature?

muhammad-ali-e: Make the method more self explanatory to avoide conflicts and confusions

backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -737,3 +814,103 @@ def _push_to_queue(
             conn = QueueUtils.get_queue_inst()
             # Enqueue the JSON string
             conn.enqueue(queue_name=q_name, message=queue_result_json)
+            logger.info(f"Pushed {file_name} to queue {q_name} with file content")
+
+    def _read_file_content_for_queue(
```
muhammad-ali-e: I think somwhere we have the similar method to get the file content.
cc: @vishnuszipstack

backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -708,6 +765,24 @@ def _push_to_queue(
         if not result:
             return
         connector: ConnectorInstance = self.source_endpoint.connector_instance
+
+        # For API deployments, connector might be None - handle this case
+        if connector is None:
```
muhammad-ali-e: Instead, can we check is_api? That would be more self-explanatory and make better sense.

backend/workflow_manager/endpoint_v2/dto.py
```diff
@@ -111,6 +113,11 @@ def from_json(json_str_or_dict: Any) -> "DestinationConfig":
             data = json_str_or_dict
         else:
             data = json.loads(json_str_or_dict)
+
+        # Handle backward compatibility for new HITL fields
+        if "hitl_queue_name" not in data:
```
muhammad-ali-e: Will this be automatically set in DestinationConfig? It looks like `hitl_queue_name: str | None = None `is defined there.

muhammad-ali-e (commented):
overal LGTM , please address the commnets I left.

backend/api_v2/deployment_helper.py
```diff
@@ -149,6 +149,7 @@ def execute_workflow(
         use_file_history: bool = False,
         tag_names: list[str] = [],
         llm_profile_id: str | None = None,
+        hitl_queue_name: str | None = None,
```
muhammad-ali-e, line 152: NIT: Can we introduce kwargs , and add these hilt and extra params in it?

Deepak-Kesavan: @muhammad-ali-e will address this later.

backend/api_v2/serializers.py
Deepak-Kesavan: @chandrasekharan-zipstack I have removed the duplicate msg from map. Because it is not related to a field.

This will be rendered as alert in frontend.

backend/api_v2/serializers.py
Deepak-Kesavan: @chandrasekharan-zipstack Yes.

backend/api_v2/serializers.py
Deepak-Kesavan: Removed this.

backend/workflow_manager/workflow_v2/workflow_helper.py
Deepak-Kesavan: Will handle this later because this needs to go in for a customer

backend/workflow_manager/endpoint_v2/destination.py
Deepak-Kesavan: @muhammad-ali-e There is a comment added above why this condition is checked. also added logger for the same below

Deepak-Kesavan: Will address it later because a customer is waiting for this change

backend/workflow_manager/endpoint_v2/destination.py
Deepak-Kesavan: @muhammad-ali-e I could already see queue name mentioned here.

https://github.com/Zipstack/unstract/pull/1421/files#diff-d63d3955d0aa54bb800259d28d82ee8be38f5359af3aee0cfa3c08c09af08ab5L721

Just renamed the function for now.

backend/workflow_manager/endpoint_v2/dto.py
Deepak-Kesavan: removed it.

backend/workflow_manager/endpoint_v2/destination.py
Deepak-Kesavan: Checked with @vishnuszipstack  and confirmed that having this function here is fine.

backend/workflow_manager/endpoint_v2/destination.py
Deepak-Kesavan: Removed the fallback completely because we won't be able to find the file in local.

backend/workflow_manager/endpoint_v2/destination.py
Deepak-Kesavan: its not valid anymore

backend/api_v2/serializers.py
Deepak-Kesavan: @chandrasekharan-zipstack Just added to make sure the naming has proper format.

chandrasekharan-zipstack: @Deepak-Kesavan no my question was more on why are we enforcing that names can't have double underscore or hyphens?

backend/workflow_manager/endpoint_v2/destination.py
chandrasekharan-zipstack: @Deepak-Kesavan why have we removed the API parameter of `push_to_hitl` but still chose to retain this parameter within the code for checks?

backend/api_v2/api_deployment_views.py
```diff
@@ -79,9 +79,8 @@ def post(
             llm_profile_id=llm_profile_id,
             hitl_queue_name=hitl_queue_name,
         )
-        logger.info(f"Workflow execution response: {response}")
         if "error" in response and response["error"]:
-            logger.error(f"Workflow execution error: {response}")
+            logger.error("API deployment execution failed")
```
chandrasekharan-zipstack, line 83: NIT: We could add org related information or any other info to help us debug better

backend/api_v2/models.py
```diff
@@ -97,20 +97,9 @@ def save(self, *args: Any, **kwargs: Any) -> None:
 
         Also enforces one API deployment per workflow constraint for new deployments.
         """
-        from django.core.exceptions import ValidationError
-
         organization_id = UserContext.get_organization_identifier()
 
         # For new deployments, check one API per workflow constraint
```
chandrasekharan-zipstack:
```suggestion
```

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -68,6 +68,12 @@
 
 logger = logging.getLogger(__name__)
 
+# Parameters to exclude when calling create_workflow_execution
+EXECUTION_EXCLUDED_PARAMS = {
```
chandrasekharan-zipstack, line 72: @Deepak-Kesavan do you know why we need to exclude some params when calling a function? This seems unconventional and I would like to know the why behind this
cc: @vishnuszipstack I had the same concern when you introduced the change

backend/api_v2/api_deployment_views.py
Deepak-Kesavan: @chandrasekharan-zipstack Won't the org_id be included as part of logger config?

backend/workflow_manager/workflow_v2/workflow_helper.py
Deepak-Kesavan: Not sure @chandrasekharan-zipstack . It was there earlier. I just modified it.

vishnuszipstack: @chandrasekharan-zipstack that param is not required for the further function calls. so removing from kwargs.

vishnuszipstack (approved):

backend/api_v2/serializers.py
Deepak-Kesavan: No specific reason @chandrasekharan-zipstack . I just added it myself.

PR #1430, opened 2025-07-14
UN-2608 [FIX] Platform key no longer exposed when editing adapters for paid users

athul-rs:
## What

- A bug has been identified where the platform key is visible during adapter edit  for paid users who had manually created the adapter using a platform key.

## Why

- Platform keys should be handled securely and not exposed during adapter edits, especially for paid users with existing, valid configurations.

## How

- Disabled key field when the flag platform key is set as TRUE.

-

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


jaseemjaskp (approved):

backend/adapter_processor_v2/views.py
```diff
@@ -147,6 +147,23 @@ def get_permissions(self) -> list[Any]:
...
+        serializer = self.get_serializer(instance)
+        data = serializer.data
+
+        # Check if we need to hide the unstract_key
+        if data.get("adapter_metadata", {}).get(
+            "use_platform_provided_unstract_key"
+        ) is True and "unstract_key" in data.get("adapter_metadata", {}):
+            # Set the unstract_key to an empty string instead of removing it
+            data["adapter_metadata"]["unstract_key"] = ""
+        return Response(data)
```
chandrasekharan-zipstack: @athul-rs move this method to the `to_representation()` of the serializer. That way if we use it elsewhere, we needn't duplicate this logic. It exists to help validate and store data in the DB and to alter any data that we view from it

chandrasekharan-zipstack: @coderabbitai do you think the above suggestion is valid?

athul-rs: @coderabbitai check the updated code as per review comments. Additionally check if this change will cause any issue for other methods using the serializer.

athul-rs: @chandrasekharan-zipstack Updated and approved. Please review again.

PR #1444, opened 2025-07-17
UN-2606 [FIX] Error handling in deploy as API from prompt studio

jagadeeswaran-zipstack:
## What

1. **Backend Validation**: Added workflow endpoint validation to prevent API deployments from being created with incomplete workflow configurations
2. **Frontend State Management**: Fixed issue where updated deployment details (particularly corrected API names) weren't being sent to the backend API

## Why

**Problem 1**: Users could create API deployments from workflows that didn't have proper source or destination endpoint configurations, leading to runtime failures when the API was called.

**Problem 2**: When prompt studio names exceeded 30 characters, users would see validation errors and correct the API name, but the updated values weren't being sent to the backend, causing deployment creation to fail with the original (invalid) name.

### Backend Changes (`backend/api_v2/serializers.py`):
- Validates that workflow has both source and destination endpoints configured
- Checks that endpoints have proper connection types set
- Provides clear, user-friendly error messages for each validation failure

### Frontend Changes (`frontend/src/components/deployments/create-api-deployment-from-prompt-studio/CreateApiDeploymentFromPromptStudio.jsx`):
- Enhanced `handleNext()` to properly capture and update form values before proceeding to next step
- Added safety mechanism in `createApiDeployment()` to get latest form values and merge with state
- Ensures API deployment creation always uses the most current form data

### Backend Validation Testing:
- Test with workflows that have no endpoints configured
- Test with workflows missing source endpoints
- Test with workflows missing destination endpoints  
- Test with endpoints that have no connection type set
- Test with non-API connections that have no connector instance
- Verify appropriate error messages are returned for each scenario

### Frontend Form Handling Testing:
- Create prompt studio tools with names longer than 30 characters
- Verify validation error appears and blocks navigation to next step
- Correct the API name to valid format (≤30 chars, alphanumeric + underscore/hyphen)
- Confirm "Next" button works and proceeds to tool settings step
- Complete deployment creation and verify corrected API name is used
- Test with various edge cases (special characters, empty names, etc.)

## Screen Shots
 
image


backend/api_v2/serializers.py
```diff
@@ -57,6 +58,68 @@ def validate_api_name(self, value: str) -> str:
         api_name_validator(value)
         return value
 
+    def validate_workflow(self, workflow):
+        """Validate that the workflow has properly configured source and destination endpoints."""
+        if not workflow:
+            raise ValidationError("Workflow is required.")
```
chandrasekharan-zipstack: NIT: I think this will be handled by the model serializer itself, you could check and remove it if that's the case

backend/api_v2/serializers.py
```diff
@@ -57,6 +58,68 @@ def validate_api_name(self, value: str) -> str:
...
+            )
+
+        # For non-API connections, check if connector instance is configured
+        if (
+            source_endpoint.connection_type != WorkflowEndpoint.ConnectionType.API
+            and not source_endpoint.connector_instance
+        ):
+            raise ValidationError(
+                "Source endpoint must have a connector configured for non-API connections before creating an API deployment."
+            )
```
chandrasekharan-zipstack: @jagadeeswaran-zipstack is the same serializer used for ETL / Task based pipeline's API trigger? If its not used then this check won't be required

backend/api_v2/serializers.py
```diff
@@ -57,6 +58,68 @@ def validate_api_name(self, value: str) -> str:
...
+            destination_endpoint.connection_type
+            not in [
+                WorkflowEndpoint.ConnectionType.API,
+                WorkflowEndpoint.ConnectionType.MANUALREVIEW,
+            ]
+            and not destination_endpoint.connector_instance
+        ):
+            raise ValidationError(
+                "Destination endpoint must have a connector configured for non-API and non-manual review connections before creating an API deployment."
+            )
```
chandrasekharan-zipstack: Similar concern - if this serializer is only used in API deployments we might not need this check

backend/api_v2/serializers.py
```diff
@@ -57,6 +58,68 @@ def validate_api_name(self, value: str) -> str:
...
+        source_endpoints = endpoints.filter(
+            endpoint_type=WorkflowEndpoint.EndpointType.SOURCE
+        )
+        if not source_endpoints.exists():
+            raise ValidationError(
+                "Workflow must have a source endpoint configured before creating an API deployment."
+            )
+
+        source_endpoint = source_endpoints.first()
+        if not source_endpoint.connection_type:
```
chandrasekharan-zipstack, line 80: NIT: You could directly check if the connection_type is "API" itself

chandrasekharan-zipstack (approved):
LGTM - however left some comments that can be checked and addressed

backend/api_v2/serializers.py
jaags-dev: Sure 👍

PR #1447, opened 2025-07-22
UN-2603 [FIX] Fix ToolMetadataNotFound error when use_file_history is true

Deepak-Kesavan:
## What

- Fixed ToolMetadataNotFound error that occurs when `use_file_history` is set to true

## Why

- When `use_file_history` is true, the `_should_handle_hitl` method calls `get_tool_execution_result()` which attempts to read metadata that may not exist
- This causes a ToolMetadataNotFound exception and breaks the workflow execution

## How

- Added a check in `_should_handle_hitl` to skip HITL validation when `use_file_history` is true
- This prevents calling `get_tool_execution_result()` which requires metadata that may not be available
- Minimal 3-line fix that maintains existing behavior for all other scenarios

- No, this PR cannot break existing features because:
  - The fix only adds a condition to skip HITL validation when `use_file_history` is true
  - All other code paths remain unchanged
  - When `hitl_queue_name` is present, it takes priority as before
  - Normal workflow-based HITL logic continues to work when `use_file_history` is false

## Database Migrations

- None

## Env Config

- None

## Relevant Docs

- Created flow.md documenting the HITL decision flow with all parameter combinations

## Related Issues or PRs

- Related to UN-2603: Changes to support pushing from API to HITL

## Dependencies Versions

- None

## Notes on Testing

- Tested with `use_file_history=false` - works as before
- Tested with `use_file_history=true` - no longer throws ToolMetadataNotFound
- Tested with `hitl_queue_name` present - works as before
- All existing pre-commit hooks pass

## Screenshots

N/A - Backend logic change

## Checklist

### DIFF
--- a/backend/workflow_manager/endpoint_v2/destination.py
+++ b/backend/workflow_manager/endpoint_v2/destination.py
@@ -172,6 +172,9 @@ class DestinationConnector(BaseConnector):
             return True
 
         # Otherwise use existing workflow-based HITL logic
+        # Skip HITL validation if we're using file_history and no execution result is available
+        if self.use_file_history:
+            return False
         execution_result = self.get_tool_execution_result()
         if WorkflowUtil.validate_db_rule(
             execution_result, workflow, file_hash.file_destination


backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -172,6 +172,9 @@ def _should_handle_hitl(
             return True
 
         # Otherwise use existing workflow-based HITL logic
+        # Skip HITL validation if we're using file_history and no execution result is available
```
chandrasekharan-zipstack, line 175: NIT: @Deepak-Kesavan we might use this parameter in demos and if they wish to show something involving HITL, it might not work as expected. Do log (to the frontend also if possible) / document it accordingly

vishnuszipstack: resolving for now, @Deepak-Kesavan consider it in upcoming PR.

PR #1451, opened 2025-07-23
UN-1736 [FIX] Resolved ETL workflow failure when renamed adapter is used in tool settings

kirtimanmishrazipstack:
## What

- 

## Why

-

## How

-

-

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/backend/tool_instance_v2/serializers.py
+++ b/backend/tool_instance_v2/serializers.py
@@ -51,7 +51,85 @@ class ToolInstanceSerializer(AuditSerializer):
             return rep
         rep[ToolKey.ICON] = tool.icon
         rep[ToolKey.NAME] = tool.properties.display_name
+
+        # Transform adapter IDs back to names for UI display
+        metadata = rep.get(TIKey.METADATA, {})
+        if metadata:
+            rep[TIKey.METADATA] = self._transform_adapter_ids_to_names_for_display(
+                metadata, tool_function
+            )
+
         return rep
+
+    def _transform_adapter_ids_to_names_for_display(
+        self, metadata: dict[str, Any], tool_uid: str
+    ) -> dict[str, Any]:
+        """Transform adapter IDs back to names for UI display."""
+        from adapter_processor_v2.adapter_processor import AdapterProcessor
+
+        from tool_instance_v2.tool_instance_helper import ToolInstanceHelper
+        from tool_instance_v2.tool_processor import ToolProcessor
+        from unstract.tool_registry.tool_utils import ToolUtils
+
+        # Create a copy to avoid mutating the original
+        display_metadata = metadata.copy()
+
+        try:
+            tool: Tool = ToolProcessor.get_tool_by_uid(tool_uid=tool_uid)
+            schema = ToolUtils.get_json_schema_for_tool(tool)
+
+            # Get all adapter keys
+            adapter_keys = set()
+            adapter_keys.update(schema.get_llm_adapter_properties().keys())
+            adapter_keys.update(schema.get_embedding_adapter_properties().keys())
+            adapter_keys.update(schema.get_vector_db_adapter_properties().keys())
+            adapter_keys.update(schema.get_text_extractor_adapter_properties().keys())
+            adapter_keys.update(schema.get_ocr_adapter_properties().keys())
+
+            # Transform IDs back to names for display
+            for adapter_key in adapter_keys:
+                if adapter_key in display_metadata and display_metadata[adapter_key]:
+                    adapter_value = display_metadata[adapter_key]
+                    # If it's a UUID (adapter ID), convert to name for display
+                    if ToolInstanceHelper.is_uuid_format(adapter_value):
+                        try:
+                            adapter_name = AdapterProcessor.get_adapter_instance_by_id(
+                                adapter_value
+                            )
+                            if adapter_name:
+                                display_metadata[adapter_key] = adapter_name
+                            else:
+                                # Adapter ID exists but returns None - mark as orphaned
+                                display_metadata[adapter_key] = (
+                                    f"[DELETED ADAPTER: {adapter_value[:8]}...]"
+                                )
+                                logger.warning(
+                                    f"Adapter ID {adapter_value} references a deleted or inaccessible adapter"
+                                )
+                        except Exception as e:
+                            # If conversion fails, show user-friendly error
+                            display_metadata[adapter_key] = (
+                                f"[ERROR: {adapter_value[:8]}...]"
+                            )
+                            logger.error(
+                                f"Could not resolve adapter ID {adapter_value} to name: {e}"
+                            )
+                    # If it's not a UUID, it might be an old adapter name that couldn't be migrated
+                    elif adapter_value and not ToolInstanceHelper.is_uuid_format(
+                        adapter_value
+                    ):
+                        # This might be an orphaned adapter name - show warning
+                        display_metadata[adapter_key] = f"{adapter_value} [NEEDS UPDATE]"
+                        logger.warning(
+                            f"Tool metadata contains non-UUID adapter reference: {adapter_value}"
+                        )
+
+        except Exception as e:
+            logger.warning(f"Error transforming adapter IDs to names: {e}")
+            # Return original metadata if transformation fails
+            pass
+
+        return display_metadata
 
     def create(self, validated_data: dict[str, Any]) -> Any:
         workflow_id = validated_data.pop(WorkflowKey.WF_ID)
--- a/backend/tool_instance_v2/tool_instance_helper.py
+++ b/backend/tool_instance_v2/tool_instance_helper.py
@@ -82,6 +82,10 @@ class ToolInstanceHelper:
         Returns:
             None
         """
+        print("** adapter_key ** ", adapter_key)
+        print("** adapter_property ** ", adapter_property)
+        print("** adapter_type ** ", adapter_type)
+
         if adapter_key in metadata:
             adapter_name = metadata[adapter_key]
             adapter = AdapterProcessor.get_adapter_by_name_and_type(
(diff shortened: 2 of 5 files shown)


backend/adapter_processor_v2/exceptions.py
```diff
@@ -88,3 +88,15 @@ def __init__(
                 "It is used in a workflow or a prompt studio project"
             )
         super().__init__(detail, code)
+
+
+class AdapterNotFound(APIException):
+    status_code = 500
```
chandrasekharan-zipstack: NIT: Should this be a 404?

backend/tool_instance_v2/serializers.py
```diff
@@ -51,8 +55,102 @@ def to_representation(self, instance: ToolInstance) -> dict[str, str]:
...
+                            logger.error(
+                                f"Could not validate adapter name '{adapter_value}': {e}"
+                            )
+
+        except Exception as e:
+            logger.warning(f"Error transforming adapter IDs to names: {e}")
+            # Return original metadata if transformation fails
+            pass
+
+        return display_metadata
```
chandrasekharan-zipstack: @kirtimanmishrazipstack in case of an exception, we'll be failing silently and return the metadata with UUIDs instead right? Won't it break elsewhere / still be unusable for users? It might be good to just raise the exception and then fix it

backend/tool_instance_v2/serializers.py
```diff
@@ -51,8 +55,102 @@ def to_representation(self, instance: ToolInstance) -> dict[str, str]:
...
+            for key in schema.get_llm_adapter_properties().keys():
+                adapter_key_to_type[key] = AdapterTypes.LLM
+            for key in schema.get_embedding_adapter_properties().keys():
+                adapter_key_to_type[key] = AdapterTypes.EMBEDDING
+            for key in schema.get_vector_db_adapter_properties().keys():
+                adapter_key_to_type[key] = AdapterTypes.VECTOR_DB
+            for key in schema.get_text_extractor_adapter_properties().keys():
+                adapter_key_to_type[key] = AdapterTypes.X2TEXT
+            for key in schema.get_ocr_adapter_properties().keys():
+                adapter_key_to_type[key] = AdapterTypes.OCR
```
chandrasekharan-zipstack, line 91: NIT: Might not be needed, we could plan to remove this too eventually
```suggestion
```

backend/tool_instance_v2/serializers.py
```diff
@@ -51,8 +55,102 @@ def to_representation(self, instance: ToolInstance) -> dict[str, str]:
...
+
+            # Transform IDs back to names for display
+            for adapter_key, adapter_type in adapter_key_to_type.items():
+                if adapter_key in display_metadata and display_metadata[adapter_key]:
+                    adapter_value = display_metadata[adapter_key]
```
chandrasekharan-zipstack: NIT: Follow this to reduce a level of nesting
```suggestion
                adapter_value = display_metadata.get(adapter_key)
                if not adapter_value:
                    continue
                # Other logic continues with reduced nesting
```

backend/tool_instance_v2/serializers.py
```diff
@@ -51,8 +55,102 @@ def to_representation(self, instance: ToolInstance) -> dict[str, str]:
...
+                    # If it's a UUID (adapter ID), convert to name for display
+                    if ToolInstanceHelper.is_uuid_format(adapter_value):
+                        try:
+                            adapter_name = AdapterProcessor.get_adapter_instance_by_id(
+                                adapter_value
+                            )
```
chandrasekharan-zipstack:
```suggestion
                            adapter_name = AdapterInstance.objects.get(id=adapter_instance_id).adapter_name
```
Directly use this, `AdapterProcessor.get_adapter_instance_by_id()` is error prone since it suppresses exceptions and then attempts to access the adapter name. If possible correct it

backend/tool_instance_v2/serializers.py
```diff
@@ -51,8 +55,102 @@ def to_representation(self, instance: ToolInstance) -> dict[str, str]:
...
+                        except Exception as e:
+                            # If conversion fails, show user-friendly error
+                            display_metadata[adapter_key] = (
+                                f"[{adapter_value[:8]}...] NOT FOUND"
+                            )
+                            logger.error(
+                                f"Could not resolve adapter ID {adapter_value} to name: {e}"
+                            )
+                    # If it's not a UUID, validate if the adapter name still exists
+                    elif adapter_value and not ToolInstanceHelper.is_uuid_format(
```
muhammad-ali-e: Why this? do we have a plan a migration (some are string and and some are UUId)?

backend/tool_instance_v2/tool_instance_helper.py
```diff
@@ -81,17 +84,35 @@ def update_metadata_with_adapter_properties(
...
-            )
-            metadata[metadata_key_for_id] = adapter_id
+            adapter_value = metadata[adapter_key]
+            adapter_id = None
+
+            # Check if the value is already an adapter ID (UUID format)
+            if ToolInstanceHelper.is_uuid_format(adapter_value):
+                # It's already an adapter ID - validate it exists
+                logger.debug(f"Adapter value '{adapter_value}' is already in UUID format")
+                from adapter_processor_v2.models import AdapterInstance
```
muhammad-ali-e: NIT: can we move the import to top of file? Did you run precommit while making PR?

backend/tool_instance_v2/tool_instance_helper.py
```diff
@@ -251,6 +272,142 @@ def reorder_tool_instances(instances_to_reorder: list[uuid.UUID]) -> None:
             tool_instance.step = step + 1
             tool_instance.save()
 
+    @staticmethod
+    def is_uuid_format(value: str) -> bool:
```
muhammad-ali-e, line 271: Would be btter if it in common utility class

backend/tool_instance_v2/tool_instance_helper.py
```diff
@@ -251,6 +272,142 @@ def reorder_tool_instances(instances_to_reorder: list[uuid.UUID]) -> None:
...
+        adapter_keys.update(schema.get_llm_adapter_properties().keys())
+        adapter_keys.update(schema.get_embedding_adapter_properties().keys())
+        adapter_keys.update(schema.get_vector_db_adapter_properties().keys())
+        adapter_keys.update(schema.get_text_extractor_adapter_properties().keys())
+        adapter_keys.update(schema.get_ocr_adapter_properties().keys())
+
+        # Check if any adapter value is in UUID format (indicates already migrated)
+        for key in adapter_keys:
+            if key in metadata and metadata[key]:
+                if ToolInstanceHelper.is_uuid_format(metadata[key]):
```
muhammad-ali-e: NIT: can we use an `and`  here? instaed of two `if` ?

backend/tool_instance_v2/tool_instance_helper.py
```diff
@@ -251,6 +272,142 @@ def reorder_tool_instances(instances_to_reorder: list[uuid.UUID]) -> None:
...
+    @staticmethod
+    def get_available_adapter_names_by_type(
+        adapter_type: AdapterTypes, user: User = None
+    ) -> list[str]:
+        """Get list of available adapter names for a given type."""
+        try:
+            if user:
+                adapters = AdapterProcessor.get_adapters_by_type(adapter_type, user)
+            else:
+                # Get all adapters if no user context (fallback)
```
muhammad-ali-e: Are there any condition withoiut user context?  I think  not.. Do we need this fallback?

backend/tool_instance_v2/serializers.py
```diff
@@ -51,8 +55,102 @@ def to_representation(self, instance: ToolInstance) -> dict[str, str]:
...
+                                display_metadata[adapter_key] = (
+                                    f"[DELETED ADAPTER: {adapter_value[:8]}...]"
+                                )
+                                logger.warning(
+                                    f"Adapter ID {adapter_value} references a deleted or inaccessible adapter"
+                                )
+                        except Exception as e:
+                            # If conversion fails, show user-friendly error
+                            display_metadata[adapter_key] = (
+                                f"[{adapter_value[:8]}...] NOT FOUND"
```
muhammad-ali-e: What this `{adapter_value[:8]}` yoiu may assign this to it a meaning full name so it's more readble

backend/tool_instance_v2/serializers.py
```diff
@@ -51,8 +55,102 @@ def to_representation(self, instance: ToolInstance) -> dict[str, str]:
...
+                                display_metadata[adapter_key] = (
+                                    f"{adapter_value} [NEEDS UPDATE]"
+                                )
+                                logger.warning(
+                                    f"Adapter name '{adapter_value}' no longer exists for type {adapter_type} - may have been renamed"
+                                )
+                        except Exception as e:
+                            # If validation fails, show error
+                            display_metadata[adapter_key] = f"[{adapter_value} NOT FOUND]"
+                            logger.error(
```
muhammad-ali-e: don't we need a exception raise from here?

backend/tool_instance_v2/tool_instance_helper.py
```diff
@@ -251,6 +272,142 @@ def reorder_tool_instances(instances_to_reorder: list[uuid.UUID]) -> None:
...
+                from adapter_processor_v2.models import AdapterInstance
+
+                adapters = AdapterInstance.objects.filter(adapter_type=adapter_type.value)
+            return [adapter.adapter_name for adapter in adapters]
+        except Exception as e:
+            logger.warning(
```
muhammad-ali-e: Why wrning? is it not error?

backend/tool_instance_v2/serializers.py
```diff
@@ -51,8 +55,102 @@ def to_representation(self, instance: ToolInstance) -> dict[str, str]:
...
+                                adapter_value
+                            )
+                            if adapter_name:
+                                display_metadata[adapter_key] = adapter_name
+                            else:
+                                # Adapter ID exists but returns None - mark as orphaned
+                                display_metadata[adapter_key] = (
+                                    f"[DELETED ADAPTER: {adapter_value[:8]}...]"
+                                )
+                                logger.warning(
```
muhammad-ali-e: Any cleanup is required here in DB?

backend/tool_instance_v2/serializers.py
kirtimanmishrazipstack: @chandrasekharan-zipstack let it be for now. We can plan to remove it later.

backend/tool_instance_v2/serializers.py
kirtimanmishrazipstack: @muhammad-ali-e we are following a lazy migration here.

- Assume prior to these changes we have adapter-name is tool metadata
- Now this changes come in. Now tool metadata will be updating with adapter-id
- But in UI, we would still show adapter-name
- Even if user changes to a different adapter, tool-metadata would still be adapter-id but it will be adapter-name in UI. 

You can follow both scenario in this PR

backend/tool_instance_v2/serializers.py
kirtimanmishrazipstack: Removed [:8] part as not needed

backend/tool_instance_v2/serializers.py
kirtimanmishrazipstack: @muhammad-ali-e you can follow both scenario in Notes on testing part. 

Already explained:

- Assume prior to these changes we have adapter-name is tool metadata
- Now this changes come in. Now tool metadata will be updating with adapter-id. Here we are doing lazy migration in tool metadata DB.
- But in UI, we would still show adapter-name
- Even if user changes to a different adapter, tool-metadata would still be adapter-id but it will be adapter-name in UI.

backend/tool_instance_v2/serializers.py
kirtimanmishrazipstack: @muhammad-ali-e lets not raise from here, else workflow will not load. This is tested

backend/tool_instance_v2/serializers.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Raised an error log.
I am skeptical about raising any error from here as it would break workflow

backend/tool_instance_v2/tool_instance_helper.py
kirtimanmishrazipstack: @muhammad-ali-e let it be there for now. Don't want to have a utility class for a single small method

backend/tool_instance_v2/serializers.py
```diff
@@ -91,64 +92,61 @@ def _transform_adapter_ids_to_names_for_display(
...
-                            # If validation fails, show error
-                            display_metadata[adapter_key] = f"[{adapter_value} NOT FOUND]"
-                            logger.error(
-                                f"Could not validate adapter name '{adapter_value}': {e}"
+                            logger.warning(
+                                f"Adapter name '{adapter_value}' no longer exists for type {adapter_type} - may have been renamed"
                             )
+                    except Exception as e:
+                        # If validation fails, show error
+                        display_metadata[adapter_key] = f"[{adapter_value} NOT FOUND]"
```
chandrasekharan-zipstack, line 143: @kirtimanmishrazipstack 
1. What will happen if the user's select this adapter which is not found and proceeds to execute? We'll still fail later right? I suggest raising that error early on if we anyway plan to
2. Also how can the user set it right if an old adapter of theirs is no longer found? Will this NOT FOUND suffix go away when they reconfigure with a different adapter?

backend/tool_instance_v2/serializers.py
```diff
@@ -91,64 +92,61 @@ def _transform_adapter_ids_to_names_for_display(
...
+                        logger.error(
+                            f"Could not validate adapter name '{adapter_value}': {e}"
+                        )
 
         except Exception as e:
-            logger.warning(f"Error transforming adapter IDs to names: {e}")
-            # Return original metadata if transformation fails
-            pass
-
+            logger.error(f"Error transforming adapter IDs to names: {e}")
```
chandrasekharan-zipstack: We are catching and suppressing the error here. I suggest allowing the error to propagate instead.
If you think otherwise, we should atleast print the traceback too

backend/tool_instance_v2/serializers.py
kirtimanmishrazipstack: 1. @chandrasekharan-zipstack This scenario only arises when user renames an adapter, and my changes goes in next. The workflow will fail. This is because we can't find adapter from non-existent `adapter-name`.
Message `not found` will be shown when user tries to open tool settings. The user can select the updated adapter as it is evident from screenshot below

2. Once they selects the present adapter from the dropdown, the UI will be updated `adapter-name`. Meanwhile tool metadata will be updated with `adapter-id`. And they can run the workflow without an issue

kirtimanmishrazipstack:
> @kirtimanmishrazipstack its not clear on what the user needs to do if they come across an entry in the dropdown which says
> 
> * NEEDS UPDATE
> * NOT FOUND
>   Any ideas on how this can be made intuitive?

@chandrasekharan-zipstack from the screenshots I think user can figure out. Please check all 3 screenshots I have in PR description. I have also given 2 scenarios which I have tested whole flow.

backend/tool_instance_v2/tool_instance_helper.py
chandrasekharan-zipstack: @kirtimanmishrazipstack I think he's asking you to write this function in one of the existing utils class and refer from there. See [this](https://github.com/Zipstack/unstract/blob/ba1df894d28a43ba40a02419a88695d117c14030/backend/utils/common_utils.py#L11)

PR #1453, opened 2025-07-23
UN-2640 [FIX] Resolve empty task_id causing 404 status API errors

muhammad-ali-e:
## Summary
- Fixed issue where task_id field was stored as empty string in WorkflowExecution table
- Prevented 404 errors on deployment status API calls when task_id is missing
- Added validation to skip saving empty task_id values while allowing execution to continue

## What
- Modified `execute_workflow_async` in `workflow_helper.py` to add logging and conditional task_id saving
- Updated `update_execution_task` in `execution.py` to validate task_id before saving
- Changed `get_status_of_async_task` to log warning instead of raising TaskDoesNotExistError

## Why
- Celery sometimes returns AsyncResult with empty .id property causing database to store empty string
- Empty task_id caused 404 responses on `/deployment/` status API calls
- Executions were completing successfully but status tracking failed due to missing task_id

## How
- Added logging to track AsyncResult.id values for debugging
- Implemented validation to prevent saving empty/null task_id values
- Replaced error raising with warning logs to avoid blocking execution flow
- Ensured execution continues normally even when task_id is unavailable

## Breaking Changes
None

## Database Migrations
None - no schema changes required

## Testing
- Verified executions continue without errors when task_id is empty
- Confirmed status API returns appropriate responses instead of 404 errors
- Validated that normal executions with valid task_id work as expected


backend/workflow_manager/workflow_v2/execution.py
```diff
@@ -404,15 +404,20 @@ def update_execution_err(execution_id: str, err_msg: str = "") -> WorkflowExecut
     @staticmethod
     def update_execution_task(execution_id: str, task_id: str) -> None:
         try:
-            if not task_id:
+            if not task_id or task_id == "":
```
chandrasekharan-zipstack: NIT: Redundant check, is there a difference between the two?
```suggestion
            if not task_id:
```

backend/workflow_manager/workflow_v2/execution.py
```diff
@@ -404,15 +404,20 @@ def update_execution_err(execution_id: str, err_msg: str = "") -> WorkflowExecut
...
                 logger.warning(
-                    f"task_id: '{task_id}' is NULL / empty for "
-                    f"execution_id: {execution_id}, expected to have a UUID"
+                    f"Skipping empty task_id: '{task_id}' for "
+                    f"execution_id: {execution_id}, expected to have a valid UUID"
```
chandrasekharan-zipstack:
```suggestion
                    f"Skipped setting taskID for execution {execution_id} since its NULL/empty"
```
NIT: Printing the empty task ID doesn't make sense, do fix line-length issues

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -343,15 +342,14 @@ def get_status_of_async_task(
...
-            TaskDoesNotExistError: Not found exception
             ExecutionDoesNotExistError: If execution is not found
 
         Returns:
             ExecutionResponse: _description_
         """
         execution: WorkflowExecution = WorkflowExecution.objects.get(id=execution_id)
         if not execution.task_id:
-            raise TaskDoesNotExistError(f"No task ID found for execution: {execution_id}")
+            logger.warning(f"No task ID found for execution: {execution_id}")
```
chandrasekharan-zipstack: Why might not a task ID be present for an execution? What might the effect be?
Can you highlight that in the log message to let people know that its not a concern and its expected sometimes?

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -473,15 +471,31 @@ def execute_workflow_async(
...
                 id=execution_id
             )
-            workflow_execution.task_id = async_execution.id
-            workflow_execution.save()
+
+            # Handle empty task_id gracefully
+            if not async_execution.id or async_execution.id == "":
+                logger.warning(
+                    f"[{org_schema}] Celery returned empty task_id for execution_id '{execution_id}'. "
+                    f"Execution will continue but status tracking may not work."
```
chandrasekharan-zipstack: @muhammad-ali-e shouldn't this be logged to the user also? How can an API deployment user get back the result then?

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -473,15 +471,31 @@ def execute_workflow_async(
...
+                )
+                # Continue without setting task_id - execution can still complete
+            else:
+                workflow_execution.task_id = async_execution.id
+                workflow_execution.save(update_fields=["task_id"])
```
chandrasekharan-zipstack: @muhammad-ali-e if we're duplicating the logic here, why do we need `update_execution_task()` function which does the same thing?

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: This check is reduntent here. Removed it. We don't store the result behind the taskid , instaed we use cache With TTL

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: No. It not affect user. Result is now nt rely on taskId

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: Updated

chandrasekharan-zipstack (approved):

PR #1456, opened 2025-07-27
feat: Replace LLM profile selection with direct LLM adapter selection in summarization settings

jagadeeswaran-zipstack:
## What

- Replaces LLM profile selection with direct LLM adapter selection in Prompt Studio's Summarization settings
- Adds new `summarize_llm_adapter` field to `CustomTool` model for direct adapter reference
- Implements lazy migration from profile-based to adapter-based configuration
- Updates frontend to fetch and display LLM adapters directly instead of going through profiles

## Why

- **Simplifies user workflow**: Users can now select the actual LLM adapter instead of going through an intermediary profile layer
- **Improves UX**: Direct adapter selection is more intuitive and consistent with other parts of the application
- **Addresses UN-2239**: "The list of created LLMs should be selectable in the Summarization settings inside Prompt studio"
- **Reduces complexity**: Eliminates the need to create and manage profiles just for summarization settings

### Backend Implementation
- Added `summarize_llm_adapter` ForeignKey field to `CustomTool` model pointing to `AdapterInstance`
- Created database migration `0002_add_summarize_llm_adapter.py`
- Updated `partial_update` API method to handle both new adapter field and legacy profile field
- Implemented lazy migration in serializer that automatically copies adapter from profile to direct field on first access
- Added adapter validation to ensure user permissions and adapter existence
- Prioritizes new adapter field over legacy profile field when both are present

### Frontend Implementation
- Modified `SettingsModal.jsx` to remove `llmItems` prop dependency from `SummarizeManager`
- Updated field name from `summarize_llm_profile` to `summarize_llm_adapter`
- Added direct adapter fetching via `/api/v1/unstract/{orgId}/adapter/?adapter_type=LLM`
- Enhanced dropdown with search functionality and loading states

**No, this PR will not break any existing features** because:

1. **Backward Compatibility**: The legacy `summarize_llm_profile` field and API handling are fully preserved
2. **Lazy Migration**: Existing tools with profile-based settings continue to work and are automatically migrated only when accessed
3. **API Compatibility**: Both old and new API field names are supported during transition period
4. **Progressive Enhancement**: New functionality is additive, not replacing existing functionality
5. **Fallback Logic**: If adapter field is empty, system falls back to profile-based logic
6. **Non-destructive**: No existing data is modified until user explicitly changes settings

## Database Migrations

- **New Migration**: `backend/prompt_studio/prompt_studio_core_v2/migrations/0002_add_summarize_llm_adapter.py`
- **Action**: Adds `summarize_llm_adapter` field as nullable ForeignKey to `AdapterInstance`
- **Safety**: Field is nullable and has default value, no data loss or corruption risk
- **Rollback**: Can be safely rolled back if needed

## Env Config

- No environment configuration changes required
- All existing environment variables remain functional

### Recommended Testing
- Test summarization functionality with both migrated and new tools
- Verify adapter permissions work correctly across different user roles
- Test edge cases where profiles exist but have no associated LLM adapter


backend/prompt_studio/prompt_studio_core_v2/migration_utils.py
```diff
@@ -0,0 +1,123 @@
...
+        """Safely migrate a tool from profile-based to adapter-based summarization.
+
+        Args:
+            tool_instance: The CustomTool instance to migrate
+            skip_if_migrated: Skip if already migrated (default: True)
+
+        Returns:
+            bool: True if migration was performed, False if skipped or failed
+        """
+        from prompt_studio.prompt_profile_manager_v2.models import ProfileManager
```
chandrasekharan-zipstack, line 30: NIT: If this is to avoid circular import errors, mention that as a comment. If not, move it to the top of the file

backend/prompt_studio/prompt_studio_core_v2/models.py
```diff
@@ -66,6 +66,14 @@ class CustomTool(DefaultOrganizationMixin, BaseModel):
         db_comment="Field to store the summarize prompt",
         unique=False,
     )
+    summarize_llm_adapter = models.ForeignKey(
+        AdapterInstance,
+        on_delete=models.PROTECT,
+        db_comment="Field to store the LLM adapter for summarization",
+        null=True,
+        blank=True,
+        related_name="custom_tools_summarize_llm",
```
chandrasekharan-zipstack:
```suggestion
        related_name="summarize_enabled_custom_tools",
```
@jagadeeswaran-zipstack the purpose of this `related_name` field is to help access this model instance from the foreign key reference. For example, if you have a queryset containing LLM adapter instances and you want to check all prompt studio projects its used in for summarizing - you can simply do the below which is more readable
```python
summarize_enabled_projects = llm_instance.summarize_enabled_custom_tools  # More readable
# summarize_enabled_projects = llm_instance.custom_tools_summarize_llm  # Less readable
```

backend/prompt_studio/prompt_studio_core_v2/serializers.py
```diff
@@ -52,17 +52,15 @@ def to_representation(self, instance):  # type: ignore
         data = super().to_representation(instance)
         default_profile = None
 
-        # Fetch summarize LLM profile
-        try:
-            summarize_profile = ProfileManager.objects.get(
-                prompt_studio_tool=instance, is_summarize_llm=True
```
chandrasekharan-zipstack, line 58: @jagadeeswaran-zipstack does this mean that the `is_summarize_llm` of `ProfileManager` can be considered deprecated? I suggest adding a DB comment for it. One other thing you could try is check if you can mark a django field as deprecated so that we get warned about its usages - maybe consider adding the deprecated decorator on a getter for the instance

backend/prompt_studio/prompt_studio_core_v2/migration_utils.py
```diff
@@ -0,0 +1,123 @@
...
+
+        This method is safe for serializers as it doesn't perform writes.
+
+        Args:
+            tool_instance: The CustomTool instance
+
+        Returns:
+            Optional[str]: The adapter ID if found, None otherwise
+        """
+        from prompt_studio.prompt_profile_manager_v2.models import ProfileManager
```
chandrasekharan-zipstack: Move to the top of the file if possible, same goes for the other function this is done in

backend/prompt_studio/prompt_studio_core_v2/serializers.py
```diff
@@ -52,17 +52,15 @@ def to_representation(self, instance):  # type: ignore
...
-            )
+        # Handle summarize LLM adapter (read-only, safe for serialization)
+        summarize_adapter_id = SummarizeMigrationUtils.get_summarize_adapter_id(instance)
+        if summarize_adapter_id:
+            data[TSKeys.SUMMARIZE_LLM_ADAPTER] = summarize_adapter_id
+
+        # Provide legacy profile field for backward compatibility
+        summarize_profile_id = SummarizeMigrationUtils.get_summarize_profile_id(instance)
+        if summarize_profile_id:
+            data[TSKeys.SUMMARIZE_LLM_PROFILE] = summarize_profile_id
```
chandrasekharan-zipstack:
```suggestion
        # Check new adapter-based approach first
        if tool_instance.summarize_llm_adapter:
            data[TSKeys.SUMMARIZE_LLM_ADAPTER] = tool_instance.summarize_llm_adapter.id

        # Check legacy profile-based approach
        try:
            summarize_profile = ProfileManager.objects.get(
                prompt_studio_tool=tool_instance, is_summarize_llm=True
            )
            if summarize_profile.llm:
                data[TSKeys.SUMMARIZE_LLM_PROFILE] = summarize_profile.llm.id
        except ObjectDoesNotExist:
            pass
```
NIT: The functions add another layer of indirection and some code duplication as well. Do test this suggestion if its valid

backend/prompt_studio/prompt_studio_core_v2/migration_utils.py
```diff
@@ -0,0 +1,123 @@
...
+                    f"Successfully migrated tool {tool_instance.tool_id} from profile-based to adapter-based summarization"
+                )
+                return True
+
+        except Exception as e:
+            logger.error(f"Failed to migrate tool {tool_instance.tool_id}: {str(e)}")
+            return False
+
+    @staticmethod
+    def get_summarize_adapter_id(tool_instance) -> str | None:
```
chandrasekharan-zipstack: @jagadeeswaran-zipstack delete these 2 methods if you accept my suggestion where its used and if its not used anywhere else

backend/prompt_studio/prompt_studio_core_v2/views.py
```diff
@@ -136,12 +152,48 @@ def destroy(
...
+                )
+                # Validate that the adapter type is LLM
+                if adapter.adapter_type != AdapterTypes.LLM.value:
+                    raise ValidationError(
+                        "Only LLM adapters are allowed for summarization"
+                    )
+
+                # Set the adapter directly on the tool
+                prompt_tool.summarize_llm_adapter = adapter
+                prompt_tool.save()
```
chandrasekharan-zipstack: @jagadeeswaran-zipstack the validation will be done by the serializer. We needn't update and call save like this, if its a valid LLM instance UUID - it will automatically be checked and set. This will work because the serializer has this
```python
    class Meta:
        model = CustomTool
        fields = "__all__"
```

Just make sure this behaviour is noticed without this snippet. You might have to perform the lazy migration step before itself and update the data that will be passed to the serializer

backend/prompt_studio/prompt_studio_core_v2/views.py
```diff
@@ -136,12 +152,48 @@ def destroy(
...
+
+                # Clear any existing profile-based summarize setting
+                ProfileManager.objects.filter(prompt_studio_tool=prompt_tool).update(
+                    is_summarize_llm=False
+                )
```
chandrasekharan-zipstack: @jagadeeswaran-zipstack this should be part of the lazy migration step / serializer itself. What if it fails for some reason and the execution continues? Another thing you could do is check if its set before proceeding

backend/prompt_studio/prompt_studio_core_v2/migration_utils.py
```diff
@@ -0,0 +1,123 @@
...
+            skip_if_migrated: Skip if already migrated (default: True)
+
+        Returns:
+            bool: True if migration was performed, False if skipped or failed
+        """
+        from prompt_studio.prompt_profile_manager_v2.models import ProfileManager
+
+        # Skip if already migrated
+        if skip_if_migrated and tool_instance.summarize_llm_adapter:
+            return False
```
chandrasekharan-zipstack, line 50: NIT: You could add a debug / info log that it was skipped since its already migrated. Make sure to add some context about the user / org / project here in the log

chandrasekharan-zipstack: Also consider adding a log that we are attempting this lazy migration

backend/prompt_studio/prompt_studio_core_v2/migration_utils.py
```diff
@@ -0,0 +1,123 @@
...
+                tool_instance.summarize_llm_adapter = summarize_profile.llm
+                tool_instance.save(update_fields=["summarize_llm_adapter"])
+
+                logger.info(
+                    f"Successfully migrated tool {tool_instance.tool_id} from profile-based to adapter-based summarization"
+                )
+                return True
+
+        except Exception as e:
+            logger.error(f"Failed to migrate tool {tool_instance.tool_id}: {str(e)}")
```
chandrasekharan-zipstack:
```suggestion
            logger.warning(f"Failed to migrate tool {tool_instance.tool_id}: {str(e)}\nContinuing with the deprecated approach for now")
```
NIT: Consider making it a warning and adding a stack trace since we anyway continue. Also check and correct the line length related issues here

backend/prompt_studio/prompt_studio_core_v2/migration_utils.py
```diff
@@ -0,0 +1,123 @@
...
+                # Find the summarize profile
+                try:
+                    summarize_profile = ProfileManager.objects.select_for_update().get(
+                        prompt_studio_tool=tool_instance, is_summarize_llm=True
+                    )
+                except ObjectDoesNotExist:
+                    logger.info(
+                        f"No summarize profile found for tool {tool_instance.tool_id}, skipping migration"
+                    )
+                    return False
```
chandrasekharan-zipstack: NIT: You might want to move this check outside the transaction to reduce a redundant DB query. When summarization is not turned on (most of the times) we'll be getting into this atomic block and rechecking the DB only to return False here

backend/prompt_studio/prompt_studio_core_v2/migration_utils.py
jaags-dev: Added comment.

backend/prompt_studio/prompt_studio_core_v2/models.py
jaags-dev: Understood. I'll make the required changes

backend/prompt_studio/prompt_studio_core_v2/serializers.py
jaags-dev: Yes, we will be deprecating that but will be there for fallback. Sure I'll add the deprecation indication.

backend/prompt_studio/prompt_studio_core_v2/views.py
jaags-dev: Understood 👍

backend/prompt_studio/prompt_studio_core_v2/migration_utils.py
chandrasekharan-zipstack: Okay for now, but such circular import errors sometimes indicate an issue with how we've structured our code. Check if something can be made better or if we've incorrectly placed the code

backend/prompt_studio/prompt_studio_core_v2/views.py
```diff
@@ -136,12 +152,48 @@ def destroy(
...
+            # Validate adapter exists and is accessible
+            try:
+                adapter = AdapterInstance.objects.for_user(request.user).get(
+                    id=summarize_llm_adapter_id
+                )
+                # Validate that the adapter type is LLM
+                if adapter.adapter_type != AdapterTypes.LLM.value:
+                    raise ValidationError(
+                        "Only LLM adapters are allowed for summarization"
+                    )
```
chandrasekharan-zipstack: @jagadeeswaran-zipstack this should be part of the serializer. That's where we should validate data and exit early. Move this logic there

backend/prompt_studio/prompt_studio_core_v2/views.py
```diff
@@ -136,12 +152,48 @@ def destroy(
...
+
+                # Set the adapter directly on the tool
+                prompt_tool.summarize_llm_adapter = adapter
+                prompt_tool.save(update_fields=["summarize_llm_adapter"])
```
chandrasekharan-zipstack: @jagadeeswaran-zipstack this update logic would be taken care by the serializer itself ideally if the correct UUID is passed. Check and ensure that any redundant code is removed in the views

backend/prompt_studio/prompt_studio_core_v2/migration_utils.py
jaags-dev: This logic is removed so resolving the comment

backend/prompt_studio/prompt_studio_core_v2/migration_utils.py
jaags-dev: Added logs.

backend/prompt_studio/prompt_studio_core_v2/views.py
jaags-dev: Yes, if it fails execution continues since we have fallback for previous summarize settings

backend/prompt_studio/prompt_studio_core_v2/serializers.py
```diff
@@ -48,21 +50,63 @@ class Meta:
...
 
+    def validate_summarize_llm_adapter(self, value):
+        """Validate that the adapter type is LLM and is accessible to the user."""
+        if value is None:
+            return value
+
+        # Check if user has access to this adapter
+        request = self.context.get("request")
+        if request and hasattr(request, "user"):
+            from adapter_processor_v2.models import AdapterInstance
```
chandrasekharan-zipstack: NIT: Is this also needed because of a circular import error? I think this can be present with other imports right

backend/prompt_studio/prompt_studio_core_v2/serializers.py
```diff
@@ -48,21 +50,63 @@ class Meta:
...
+            # Update the instance first
+            instance = super().update(instance, validated_data)
+
+            # Clear any existing profile-based summarize setting
+            ProfileManager.objects.filter(prompt_studio_tool=instance).update(
+                is_summarize_llm=False
+            )
+            return instance
+
+        return super().update(instance, validated_data)
```
chandrasekharan-zipstack:
```suggestion
        instance = super().update(instance, validated_data)
        if validated_data.get("summarize_llm_adapter"):            
            # Clear any existing deprecated profile-based summarize setting
            ProfileManager.objects.filter(prompt_studio_tool=instance).update(
                is_summarize_llm=False
            )
            return instance

        return instance
```
NIT: Do check after accepting the suggestion

backend/prompt_studio/prompt_studio_core_v2/views.py
```diff
@@ -136,12 +149,20 @@ def destroy(
...
             ToolStudioKeys.SUMMARIZE_LLM_PROFILE, None
         )
-        if summarize_llm_profile_id:
-            prompt_tool = self.get_object()
 
+        if summarize_llm_profile_id:
+            # Legacy profile-based handling
             ProfileManager.objects.filter(prompt_studio_tool=prompt_tool).update(
                 is_summarize_llm=False
             )
```
chandrasekharan-zipstack: @jagadeeswaran-zipstack this update won't be needed anymore since its already done in the serializer. Also are the lines below it necessary? Don't we need to remove that as well?

backend/prompt_studio/prompt_studio_core_v2/serializers.py
```diff
@@ -48,21 +50,63 @@ class Meta:
...
+        if (
+            "summarize_llm_adapter" in validated_data
+            and validated_data["summarize_llm_adapter"]
+        ):
+            # Update the instance first
+            instance = super().update(instance, validated_data)
+
+            # Clear any existing profile-based summarize setting
+            ProfileManager.objects.filter(prompt_studio_tool=instance).update(
+                is_summarize_llm=False
```
chandrasekharan-zipstack, line 80: NIT: @jagadeeswaran-zipstack do you think it'll be better to add a NULL in the field instead to differentiate that it has been migrated? Eventually it might help remove the column easily

backend/prompt_studio/prompt_studio_core_v2/views.py
```diff
@@ -91,6 +92,18 @@ def get_permissions(self) -> list[Any]:
...
         return CustomTool.objects.for_user(self.request.user)
 
+    def retrieve(
+        self, request: Request, *args: tuple[Any], **kwargs: dict[str, Any]
+    ) -> Response:
+        """Override retrieve to trigger safe migration when accessing tools."""
+        instance = self.get_object()
+
+        # Trigger lazy migration if needed (safe, with locking)
+        SummarizeMigrationUtils.migrate_tool_to_adapter_based(instance)
```
chandrasekharan-zipstack: NIT: @jagadeeswaran-zipstack you can avoid some code duplication by overriding the `get_object()` method instead and invoking the migration there. That way, you could get rid of this `retrieve()` override and `partial_update()` override altogether

chandrasekharan-zipstack (approved):
LGTM for the most part. Left some more comments and suggestions. Essentially you only need to override the `get_object()` method to invoke the lazy migration

Deepak-Kesavan (approved):

PR #1464, opened 2025-07-30
UN-2315 [FEAT] Added frontend error handling for GDrive input/output connectors in workflow settings

kirtimanmishrazipstack:
## What

- After Signin in G-drive connector, still singin option is enabled while navigating among the options
- When we successfully connected the file system connector the files are displaying but when we click any option and try to see the listed files, the option is disabled
- When the G-drive connector is connected successfully in Source and the destination G-drive also showing as authenticated which is false.
- Can’t list files with non-zipstack account

## Why

-

## How

- Fix FE issues for point 1) 2) and 3)
- For 4) when we try to autherize zipstack account, we need to select the checkbox as shown in screenshot

- NO

## Database Migrations

- N/A

## Env Config

- N/A

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

- Tested connectors extensively on workflow settings in call with @gopalsamyperumal 
- Also run the workflow for selected connector

## Screenshots

- 

## Checklist


backend/connector_v2/views.py
```diff
@@ -58,8 +58,12 @@ def _get_connector_metadata(self, connector_id: str) -> dict[str, str] | None:
             oauth_key = self.request.query_params.get(ConnectorAuthKey.OAUTH_KEY)
             if oauth_key is None:
                 raise MissingParamException(param=ConnectorAuthKey.OAUTH_KEY)
+            # Preserve OAuth cache for reuse across multiple operations (Test Connection, Submit,
+            # File System browsing). Frontend localStorage stores cache keys which must correspond
+            # to persistent backend credentials for tab switching and repeated operations to work.
             connector_metadata = ConnectorAuthHelper.get_oauth_creds_from_cache(
-                cache_key=oauth_key, delete_key=True
+                cache_key=oauth_key,
+                delete_key=False,  # Keep cache - frontend persistence depends on backend credential storage
```
chandrasekharan-zipstack, line 66: @kirtimanmishrazipstack can you confirm if there's a TTL for this? I think it should be 1 hour but good to check anyway

chandrasekharan-zipstack (commented):
BE changes LGTM.
@kirtimanmishrazipstack why do we need to store this in cache if the test connection succeeds? I think it should be safe to delete it from the cache in that case - try to perform this just to be safe

backend/connector_v2/views.py
kirtimanmishrazipstack: @chandrasekharan-zipstack TTL is 1 hour, from [here](https://github.com/Zipstack/unstract/blob/UN-2315-Gdrive-FE-error/backend/connector_auth_v2/pipeline/common.py#L59)

frontend/src/components/input-output/configure-ds/ConfigureDs.jsx
```diff
@@ -53,6 +54,51 @@ function ConfigureDs({
 
   const { id } = useParams();
 
+  // Map connector type to proper role for OAuth isolation
+  const connectorRole = type === "input" ? "SOURCE" : "DESTINATION";
```
chandrasekharan-zipstack, line 58: @kirtimanmishrazipstack why do we need to do this? As part of centralized connectors, we'll be moving away from distinguishing connectors based on INPUT or OUTPUT types. Do you think we can avoid this here?

kirtimanmishrazipstack: @chandrasekharan-zipstack We can move this when we move to centralised connectors.
For now I think this is needed for separate OAuth for source and destinations

vishnuszipstack:
@kirtimanmishrazipstack check the sonar issue and try to resolve that.

frontend/src/components/input-output/configure-ds/ConfigureDs.jsx
kirtimanmishrazipstack: @chandrasekharan-zipstack as discused the changes are needed for this PR. We can review the changes once centralised connectors are onprogresss

PR #1465, opened 2025-07-30
MISC: Fix connector modal selection

chandrasekharan-zipstack:
## What

- Fixed connector configuration modal not showing previously selected connector when reopened
- Added sessionDetails dependency to useEffect hooks to prevent race conditions (just to be safe)

## Why

- Users experienced poor UX in connector modal - after configuring a connector and reopening the modal, they had to re-click their connector to see the settings

## How

- Added useEffect hook in ConfigureConnectorModal to auto-select configured connector based on connDetails.connector_id when modal opens
- Added sessionDetails dependency to useEffect hooks in DsSettingsCard and InputOutput components
- Added null checks for sessionDetails.orgId before making API calls to prevent requests with malformed URLs

No, this PR should not break existing features:
- Connector modal changes only add auto-selection behavior without modifying existing functionality
- Session validation prevents errors rather than causing them - it's defensive programming
- Added dependencies to useEffect ensure components re-render when session data becomes available
- All changes improve reliability and UX without altering core functionality

## Database Migrations

- None required

## Env Config

- None required

## Relevant Docs

- None required

## Related Issues or PRs

- Related to https://github.com/Zipstack/unstract/pull/1454 (connector centralization work that exposed these UX issues)

## Dependencies Versions

- None changed

## Notes on Testing

- Test connector modal: Configure a connector (e.g., Google Drive), close modal, reopen - should show configured connector highlighted automatically
- Test session race conditions: Navigate quickly between workflow input/output pages to ensure no API failures
- Verify API calls only execute when sessionDetails.orgId is available

## Screenshots

None required - these are UX and race condition fixes.

## Checklist

### DIFF
--- a/frontend/src/components/agency/configure-connector-modal/ConfigureConnectorModal.jsx
+++ b/frontend/src/components/agency/configure-connector-modal/ConfigureConnectorModal.jsx
@@ -44,6 +44,20 @@ function ConfigureConnectorModal({
       setActiveKey("1"); // default value
     }
   }, [open, connectorMetadata]);
+
+  // Auto-select the configured connector when modal opens
+  useEffect(() => {
+    if (open && connDetails?.connector_id && filteredList?.length > 0) {
+      const configuredConnector = filteredList.find(
+        (item) => item.key === connDetails.connector_id
+      );
+
+      if (configuredConnector) {
+        setSelectedId(connDetails.connector_id);
+        setSelectedItemName(configuredConnector.label);
+      }
+    }
+  }, [open, connDetails, filteredList]);
   const { setPostHogCustomEvent, posthogConnectorEventText } =
     usePostHogEvents();
 
--- a/frontend/src/components/agency/configure-connector-modal/ConfigureConnectorModal.jsx
+++ b/frontend/src/components/agency/configure-connector-modal/ConfigureConnectorModal.jsx
@@ -49,12 +49,12 @@ function ConfigureConnectorModal({
   useEffect(() => {
     if (open && connDetails?.connector_id && filteredList?.length > 0) {
       const configuredConnector = filteredList.find(
-        (item) => item.key === connDetails.connector_id
+        (item) => item?.key === connDetails?.connector_id
       );
 
       if (configuredConnector) {
-        setSelectedId(connDetails.connector_id);
-        setSelectedItemName(configuredConnector.label);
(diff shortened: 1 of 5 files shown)


frontend/src/components/agency/configure-connector-modal/ConfigureConnectorModal.jsx
```diff
@@ -44,6 +44,20 @@ function ConfigureConnectorModal({
       setActiveKey("1"); // default value
     }
   }, [open, connectorMetadata]);
+
+  // Auto-select the configured connector when modal opens
+  useEffect(() => {
+    if (open && connDetails?.connector_id && filteredList?.length > 0) {
+      const configuredConnector = filteredList.find(
+        (item) => item.key === connDetails.connector_id
```
jaags-dev: Add optional chaining where ever possible.

frontend/src/components/agency/ds-settings-card/DsSettingsCard.jsx
```diff
@@ -200,6 +200,10 @@ function DsSettingsCard({ type, endpointDetails, message }) {
       return;
     }
 
+    if (!sessionDetails?.orgId) {
+      return;
+    }
+
```
jaags-dev: Why is this logic needed here and we returns nothing so what will be the behaviour?

chandrasekharan-zipstack: This might not really be needed since the `sessionDetails` should be available, but added it just to be safe since we noticed a strange issue where the connector details were not found until the user did a hard refresh. Since its in the `useEffect`, I assume that nothing would happen when we return directly and once the `sessionDetails` are available, the `useEffect` can run once again and load the appropriate details to our state

jaags-dev: Why do we think this change fixes the "connector details were not found" issue?

If orgId is missing in `sessionDetails`, it would typically trigger a logout since we rely on `orgId `in most of our API calls. Let's revisit whether this logic is actually necessary here, as it might be masking a deeper issue rather than solving it.

chandrasekharan-zipstack: I think you're right, I was a bit skeptical when making that change. I've reverted those changes

jagadeeswaran-zipstack (approved):

PR #1467, opened 2025-08-01
UN-2653 [FIX] Fixed issue in JSON structure parsing

Deepak-Kesavan:
## What

- Fixed issue in JSON structure parsing

## Why

- In some scenarios JSON structure collapses with previous logic to determine and fix partial JSON.

## How

- Fixed it with more robust logic.

-

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/prompt-service/src/unstract/prompt_service/utils/json_repair_helper.py
+++ b/prompt-service/src/unstract/prompt_service/utils/json_repair_helper.py
@@ -1,0 +1,64 @@
+"""JSON repair utility functions."""
+
+from typing import Any
+
+from json_repair import repair_json
+
+
+def repair_json_with_best_structure(json_str: str) -> Any:
+    """Intelligently repair JSON string using the best parsing strategy.
+
+    This function attempts to parse JSON in two ways:
+    1. As-is (could be valid object, array, or partial JSON)
+    2. With array wrapping (useful for comma-separated objects)
+
+    It chooses the result based on structural integrity rather than string length.
+
+    Args:
+        json_str: The JSON string to repair
+
+    Returns:
+        The parsed JSON object with the best structure
+    """
+    # Attempt parsing as-is
+    parsed_as_is = repair_json(json_str=json_str, return_objects=True, ensure_ascii=False)
+
+    # Attempt parsing with array wrap
+    parsed_with_wrap = repair_json(
+        json_str="[" + json_str, return_objects=True, ensure_ascii=False
+    )
+
+    # If both results are strings, return the as-is result
+    if isinstance(parsed_as_is, str) and isinstance(parsed_with_wrap, str):
+        return parsed_as_is
+
+    # If only one is a string, return the non-string result
+    if isinstance(parsed_as_is, str):
+        return parsed_with_wrap
+    if isinstance(parsed_with_wrap, str):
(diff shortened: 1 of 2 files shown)


prompt-service/src/unstract/prompt_service/utils/json_repair_helper.py
```diff
@@ -0,0 +1,64 @@
...
+    Returns:
+        The parsed JSON object with the best structure
+    """
+    # Attempt parsing as-is
+    parsed_as_is = repair_json(json_str=json_str, return_objects=True, ensure_ascii=False)
+
+    # Attempt parsing with array wrap
+    parsed_with_wrap = repair_json(
+        json_str="[" + json_str, return_objects=True, ensure_ascii=False
+    )
```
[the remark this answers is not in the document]
Deepak-Kesavan: repair_json won't raise an exception, it simply returns a empty string for response that is beyond repair.

harini-venkataraman (approved):
lgtm

prompt-service/src/unstract/prompt_service/utils/json_repair_helper.py
```diff
@@ -0,0 +1,64 @@
+"""JSON repair utility functions."""
+
+from typing import Any
+
+from json_repair import repair_json
+
+
+def repair_json_with_best_structure(json_str: str) -> Any:
```
chandrasekharan-zipstack, line 8: NIT: In order to keep it concise and meaningful, I suggest renaming this. The best structure postfix doesn't convey any additional meaning. If it conflicts with the library we're using, import that under a different name
```suggestion
def repair_json(json_str: str) -> Any:
```

prompt-service/src/unstract/prompt_service/utils/json_repair_helper.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    # Attempt parsing as-is
+    parsed_as_is = repair_json(json_str=json_str, return_objects=True, ensure_ascii=False)
+
+    # Attempt parsing with array wrap
+    parsed_with_wrap = repair_json(
+        json_str="[" + json_str, return_objects=True, ensure_ascii=False
+    )
+
+    # If both results are strings, return the as-is result
```
chandrasekharan-zipstack, line 31: NIT: @Deepak-Kesavan rather than adding comments on what we're doing here, can you highlight why we are doing this? Same goes for some of the other lines. Here I assume its because we needn't wrap as an array unnecessarily, but for the below returns it isn't clear

prompt-service/src/unstract/prompt_service/utils/json_repair_helper.py
chandrasekharan-zipstack: Also can you check and add a valid return type? Won't it be `dict[str, Any]` here?

chandrasekharan-zipstack (approved):
LGTM - left some nit comments

Deepak-Kesavan:
> LGTM - left some nit comments

Will consider these changes later. Resolving it to unblock staging.

PR #1475, opened 2025-08-11
UN-2699 [FIX] Push to api hitl issue fix

vishnuszipstack:
## What

- Push from api to hit not working fix

## Why

- Missed one function on conflict resolution

## How

- Added missing function

- No

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/backend/workflow_manager/endpoint_v2/destination.py
+++ b/backend/workflow_manager/endpoint_v2/destination.py
@@ -763,7 +763,6 @@ class DestinationConnector(BaseConnector):
         if not result:
             return
         connector: ConnectorInstance = self.source_endpoint.connector_instance
-
         # For API deployments, use workflow execution storage instead of connector
         if self.is_api:
             logger.debug(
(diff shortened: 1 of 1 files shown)


athul-rs (approved):

backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -846,3 +844,31 @@ def _push_to_queue(
...
+            if not file_storage.exists(input_file_path):
+                raise APIException(f"File not found: {input_file_path}")
+
+            file_bytes = file_storage.read(input_file_path, mode="rb")
+            if isinstance(file_bytes, str):
+                file_bytes = file_bytes.encode("utf-8")
+            return base64.b64encode(file_bytes).decode("utf-8")
+        except Exception as e:
+            logger.error(f"Failed to read file content for {file_name}: {e}")
+            raise APIException(f"Failed to read file content for queue: {e}")
```
chandrasekharan-zipstack, line 874: NIT: I suggest removing this catch all exception.
1. If a file is not found we anyway raise an `APIException` only for that to get caught and rethrown here
2. If it fails due to some pythonic error, we'll be reporting that in the error message sent to the user which is not a good practice.

chandrasekharan-zipstack (approved):
LGTM but I suggest removing this unnecessary try-except

backend/workflow_manager/endpoint_v2/destination.py
vishnuszipstack: will do it later

PR #1482, opened 2025-08-14
UN-1856 [GATED-FEAT] - Add SDK v1 into unstract package namespace

hari-kuriakose:
## What

- Move in SDK v1 into `unstract` namespace
- Replace [LlamaIndex](https://docs.llamaindex.ai/en/stable/) with [LiteLLM SDK](https://docs.litellm.ai/docs/) for powering embedding and LLM type Unstract adapters alone
- Other VectorDB, X2Text and OCR type Unstract adapters do not have any modifications
- Minor refactoring inside `sdk1` to:
    - Merge previous `SDKHelper` with `PlatformHelper` in `sdk1/platform.py`
    - Streamline `sdk1/utils` usage
- Minor tweaks in docker compose and overrides

## Why

- Discontinue separate [SDK v0](https://github.com/Zipstack/unstract-sdk) as it is preferable for us to have a unified AI gateway instead of maintaining each adapter separately.
- LiteLLM SDK offers a single OpenAI compatible API for all [embedding](https://docs.litellm.ai/docs/embedding/supported_embedding) and [LLM](https://docs.litellm.ai/docs/completion/usage) providers which aligns with us.
- LlamaIndex powered embedding type Unstract adapters like huggingface etc are relatively huge and are adding to the boot up time of `backend` service. Abstracting them away under LiteLLM unified API drastically reduces the memory footprint.
- Previous `SDKHelper` was redundant
- To reduce potential cyclic imports in  `sdk1/utils` usage
- To improve DX

### Highlights

- The entire [SDK v0](https://github.com/Zipstack/unstract-sdk) was copied into `unstract/sdk1` as a baseline
- All `sdk1` invocations from Unstract platform are feature flagged under the flag name `sdk1`
- LlamaIndex powered `sdk/adapters/embedding` is replaced with LiteLLM powered `unstract/sdk1/adapters/embedding1`
- LlamaIndex powered `sdk/adapters/llm` is replaced with LiteLLM powered `unstract/sdk1/adapters/llm1`
- `sdk1/adapters/base1.py` was introduced to scaffold LiteLLM powered `embedding1` and `llm1`:
    - `BaseAdapter` class provides the Unstract adapter interface definition for LiteLLM powered SDK v1 adapters
    - `BaseParameters` class and its derivatives provide Pydantic based parameters validation for invoking LiteLLM APIs
- Hooks into Unstract platform in the following places:
    - Support for Adapter instance ID and Tool binding is handled in `__init__` of `embedding1` and `llm1`
    - `register_adapters()` in `sdk1/adapters/base1.py` takes care of registering `embedding1` and `llm1` adapters
- `EmbeddingCompat` class in `sdk1/embedding.py` enables still LlamaIndex powered VectorDB type Unstract adapters to integrate with new LiteLLM powered embedding type Unstract adapters

### Requests to reviewers

The following places might require review from specific people, hence calling it out:

- @muhammad-ali-e Please review `backend` service and `tools`
- @Deepak-Kesavan @harini-venkataraman Please review `platform-srevice`
- @johnyrahul Please review `platform` service and `Audit` usage (`sdk1/embedding.py`, `sdk1/llm.py`)
- @chandrasekharan-zipstack Please review `docker`, `unstract/sdk1`, `tool-registry`
- @gaya3-zipstack Please review `unstract/filesystem`, `unstract/sdk1`
- @ritwik-g Please review overall

- No, it should not. Even though having a wide scope, this is feature flagged and backward compatibility maintained everywhere else.

## Database Migrations

- NA

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

- [PENDING] Add LLM generated integration tests

## Checklist

### DIFF
--- a/unstract/sdk1/src/unstract/sdk1/adapters/llm1/__init__.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/llm1/__init__.py
@@ -20,7 +20,6 @@ __all__ = [
     "AnyscaleLLMAdapter",
     "AWSBedrockLLMAdapter",
     "AzureOpenAILLMAdapter",
-    "BedrockLLMAdapter",
     "OllamaLLMAdapter",
     "OpenAILLMAdapter",
     "VertexAILLMAdapter",
--- a/unstract/sdk1/src/unstract/sdk1/adapters/vectordb/no_op/src/no_op_custom_vectordb.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/vectordb/no_op/src/no_op_custom_vectordb.py
@@ -17,8 +17,8 @@ class NoOpCustomVectorDB(VectorStore):
         dim: int,
     ) -> None:
         """Initialize params."""
-        wait_time = wait_time
-        dim = dim
+        self.wait_time = wait_time
+        self.dim = dim
 
     def query(self, query, **kwargs: Any) -> VectorStoreQueryResult:
         """Query vector store."""


unstract/sdk1/.gitignore
chandrasekharan-zipstack, line 1: NIT: This gitignore might not be needed and we simply keep duplicating information from parent folders. It makes sense to have this only if we wish to ignore something specific in this package. Otherwise, let's remove it

unstract/sdk1/.pre-commit-config.yaml
chandrasekharan-zipstack, line 1: Is this required? Will the hooks be installed in the root and apply only to this package? Even then it might add a maintenance overhead in ensuring the config doesn't conflict. I suggest removing it

unstract/sdk1/pyproject.toml
```diff
@@ -0,0 +1,238 @@
...
+#     "mistralai==1.2.5",
+#     "llama-index-llms-anyscale==0.3.0",
+#     "llama-index-llms-anthropic==0.6.14",
+#     "llama-index-llms-azure-openai==0.3.2",
+#     "llama-index-llms-vertex==0.4.6",
+#     "llama-index-llms-replicate==0.4.0",
+#     "llama-index-llms-ollama==0.5.0",
+#     "llama-index-llms-bedrock==0.3.3",
+# ]
+# v1 = ["litellm==1.74.0"]
```
chandrasekharan-zipstack:
```suggestion
```
Remove if unused

unstract/sdk1/pyproject.toml
```diff
@@ -0,0 +1,238 @@
...
+]
+docs = ["lazydocs~=0.4.8"]
+test = [
+    "parameterized==0.9.0",
+    "pytest==8.3.3",
+    "pytest-mock==3.14.0",
+]
+
+# [project.scripts]
+# unstract-tool-gen = "unstract.sdk1.scripts.tool_gen:main"
```
chandrasekharan-zipstack:
```suggestion
```

unstract/sdk1/pyproject.toml
```diff
@@ -0,0 +1,238 @@
...
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[tool.hatch.version]
+path = "src/unstract/sdk1/__init__.py"
+
+[tool.hatch.build.targets.wheel]
+packages = ["src/unstract"]
+
+# === Development tool configurations ===
```
chandrasekharan-zipstack, line 77: How does this differ from the config specified in `.pre-commit-config.yaml`? We should strive to use a single source of truth if possible. If this doesn't contribute anywhere, we could remove it.
I'm guessing this might be used if someone runs the tools directly, but we should look into ways to ensure linting is done through the same config while maintaining a single source of truth for the config

unstract/sdk1/src/unstract/sdk1/__init__.py
```diff
@@ -0,0 +1,6 @@
+__version__ = "v1.0.0"
```
chandrasekharan-zipstack, line 1: Does this make sense in a monorepo? We'll always be using the same version anyway

unstract/sdk1/src/unstract/sdk1/adapters/vectordb/milvus/pyproject.toml
chandrasekharan-zipstack, line 1: Do these individual `pyproject.toml` files help in any regard? If not, let's plan to remove them as well

chandrasekharan-zipstack (commented):
Mostly LGTM, left some minor comments alone. @hari-kuriakose @pk-zipstack I also hope we will rename `sdk1` back to `sdk` after the feature flag is removed

docker/docker-compose-dev-essentials.yaml
```diff
@@ -137,7 +137,7 @@ services:
   rabbitmq:
     image: rabbitmq:4.1.0-management
     container_name: unstract-rabbitmq
-    hostname: unstract-rabbit
+    hostname: unstract-rabbitmq
```
ritwik-g: @hari-kuriakose hope `unstract-rabbit` was not used in any of the envs. If it were this change will break existing setup.

prompt-service/src/unstract/prompt_service/controllers/answer_prompt.py
```diff
@@ -138,22 +149,40 @@ def prompt_processor() -> Any:
...
-                usage_kwargs={
-                    **usage_kwargs,
-                    PSKeys.LLM_USAGE_REASON: PSKeys.EXTRACTION,
-                },
-                capture_metrics=True,
-            )
+            if check_feature_flag_status("sdk1"):
+                llm = LLM(
+                    adapter_instance_id=output[PSKeys.LLM],
+                    tool=util,
```
gaya3-vijayakumar, line 157: There was some discussion to remove passing tools to SDK. Do we need tool in SDK? @pk-zipstack @hari-kuriakose

prompt-service/src/unstract/prompt_service/controllers/answer_prompt.py
```diff
@@ -448,17 +480,20 @@ def prompt_processor() -> Any:
...
-                        # we add this check, if not for this,
-                        # type casting to float raises
-                        # an error and we return None.
-                        if isinstance(
-                            llm.get_llm(adapter_instance_id=adapter_instance_id),
-                            NoOpCustomLLM,
-                        ):
-                            structured_output[output[PSKeys.NAME]] = answer
+                        if check_feature_flag_status("sdk1"):
+                            pass
```
gaya3-vijayakumar, line 486: How are we handling NoOp adapter for sdk1? @pk-zipstack @hari-kuriakose

prompt-service/src/unstract/prompt_service/core/retrievers/base_retriever.py
```diff
@@ -23,7 +29,10 @@ def __init__(
         self.prompt = prompt
         self.doc_id = doc_id
         self.top_k = top_k
-        self.llm = llm._llm_instance if llm else None
+        if check_feature_flag_status("sdk1"):
+            self.llm = llm if llm else None
```
gaya3-vijayakumar, line 33: Here LLM used to be Unstract LLM and _llm_instance used to llama-index LLM. Hope that is taken care accordingly with litellm.

prompt-service/src/unstract/prompt_service/services/extraction.py
```diff
@@ -38,10 +47,7 @@ def perform_extraction(
         )
         fs = FileUtils.get_fs_instance(execution_source=execution_source)
         try:
-            if enable_highlight and (
-                isinstance(x2text.x2text_instance, LLMWhisperer)
-                or isinstance(x2text.x2text_instance, LLMWhispererV2)
-            ):
+            if enable_highlight and isinstance(x2text.x2text_instance, LLMWhispererV2):
```
gaya3-vijayakumar: Does this mean that we do not have anyone using LLMWhisperer at all? All are migrated to v2?

unstract/sdk1/pyproject.toml
```diff
@@ -0,0 +1,238 @@
...
+    "jsonschema",
+    "python-magic~=0.4.27",
+    "python-dotenv==1.0.0",
+    # # Adapter changes
+    "tiktoken~=0.9.0",
+    "litellm~=v1.76.0",
+    "llama-index==0.12.39",
+    "llama-index-vector-stores-postgres==0.4.1",
+    # Including Supabase conflicts with postgres on pg-vector.
+    # Hence, commenting it out at the moment
```
gaya3-vijayakumar: We can remove this dead code, I think. I don;t see us bringing it back.

Deepak-Kesavan (approved):
Looks good. Please look into conflicts and sonar issues.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
gaya3-vijayakumar, line 1: Can't we call this base.py itself as anyway this is under sdk1, @hari-kuriakose @pk-zipstack ?

prompt-service/src/unstract/prompt_service/controllers/answer_prompt.py
chandrasekharan-zipstack: @gaya3-zipstack we are still required to do this if we wish to call `stream_log()` which is useful when we wish to pass messages to the user. If we make the `stream_log()` as a staticmethod, we should be able to remove the tool.

prompt-service/src/unstract/prompt_service/controllers/answer_prompt.py
chandrasekharan-zipstack: @hari-kuriakose @gaya3-zipstack do we still have a valid usecase for the NoOp adapter? I doubt if its being used by us in any of the load tests - if that's the case shall we consider removing it maybe?

tools/structure/src/main.py
```diff
@@ -10,11 +10,20 @@
...
-from unstract.sdk.tool.base import BaseTool
-from unstract.sdk.tool.entrypoint import ToolEntrypoint
+from unstract.flags.feature_flag import check_feature_flag_status
+
+if check_feature_flag_status("sdk1"):
+    from unstract.sdk1.constants import LogState, MetadataKey, ToolEnv, UsageKwargs
+    from unstract.sdk1.platform import PlatformHelper
+    from unstract.sdk1.prompt import PromptTool
+    from unstract.sdk1.tool.base import BaseTool
+    from unstract.sdk1.tool.entrypoint import ToolEntrypoint
```
chandrasekharan-zipstack, line 20: @hari-kuriakose @pk-zipstack can you guys confirm if this `flags` and `sdk1` will be available to tools? I don't think its copied in the [tool's Dockerfile](https://github.com/Zipstack/unstract/blob/b482210c72aa5d9c962ee45b1ccc6420be19c1d9/tools/structure/Dockerfile#L26). We'll need to update it there and it might briefly lead to a bloat in size until we remove the feature flag

hari-kuriakose:
@chandrasekharan-zipstack Will address the comments.

We will not rename `sdk1` back to `sdk`. This is to ensure namespace is not polluted and it should be an explicit decision for any downstream wanting to move to this major version.
> Mostly LGTM, left some minor comments alone. @hari-kuriakose @pk-zipstack I also hope we will rename `sdk1` back to `sdk` after the feature flag is removed

prompt-service/src/unstract/prompt_service/controllers/answer_prompt.py
hari-kuriakose: @gaya3-zipstack Yes we should remove the old constructs like tool etc. At that point we also need to decouple properly as @chandrasekharan-zipstack pointed out. However, it is preferable to do this in a later iteration to keep the changeset minimal for now.

prompt-service/src/unstract/prompt_service/controllers/answer_prompt.py
hari-kuriakose: @gaya3-zipstack @chandrasekharan-zipstack No op adapters will be replaced by the new rubberduck tool hence will be removed eventually.

VectorDB and X2Text will continue to have no op adapter for now, until rubberduck tool can emulate its workflow. Embedding and LLM adapters have switched to LiteLLM hence will not have a no op adapter though.

hari-kuriakose: @pk-zipstack Let's confirm this and once all no op adapters are removed we can resolve this conversation.

unstract/sdk1/.gitignore
praveen-formido: True @chandrasekharan-zipstack, the gitignore was in fact redundant. Removing it :+1:

unstract/sdk1/pyproject.toml
praveen-formido: Removed the commented-out code in pyproject.toml

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
hari-kuriakose: @gaya3-zipstack Since only Embedding and LLM adapters are replaced with LiteLLM, these areas require `base1.py` and everywhere else continue to use `base.py`. This was kept separate for now, just to ensure all the integration aspects are ironed out. We could definitely unify later once the dust settles.

docker/docker-compose-dev-essentials.yaml
hari-kuriakose: @ritwik-g Good catch. Reverted.

prompt-service/src/unstract/prompt_service/core/retrievers/base_retriever.py
praveen-formido: @gaya3-zipstack Both `llm` in sdkv1 and `_llm_instance` in sdkv0 expects a complete API to be defined on the instantiated object, which gets invoked during a LLM call. This expectation is met currently, and all the integration tests passed. So we can safely proceed.

prompt-service/src/unstract/prompt_service/services/extraction.py
hari-kuriakose: @gaya3-zipstack Good catch. Reverted.

Earlier expectation was to sunset LLMWhisperer v1 by the time we release SDK v1, but there are still few paid customers using it apparently. Hence the need to revert.

unstract/sdk1/pyproject.toml
praveen-formido: Done @gaya3-zipstack

tools/structure/src/main.py
hari-kuriakose: @chandrasekharan-zipstack Good catch.

Yes `sdk1` needs to be copied. It would lead to a bloat in size during this transition period as you mentioned, but that should be relatively ok.

docker/docker-compose-dev-essentials.yaml
```diff
@@ -137,7 +137,7 @@ services:
   rabbitmq:
     image: rabbitmq:4.1.0-management
     container_name: unstract-rabbitmq
-    hostname: unstract-rabbitmq
+    hostname: unstract-rabbit
```
chandrasekharan-zipstack, line 140: @pk-zipstack was this intentional? Why was this change made?

tools/classifier/requirements.txt
```diff
@@ -4,4 +4,6 @@
 # aws alone is needed here
 # because tools use transient temporary storage.
 
-unstract-sdk[aws]~=0.76.1
+unstract-sdk[aws]~=0.77.3
+-e file:/unstract/sdk1
```
chandrasekharan-zipstack, line 8: NIT: This might fail for someone trying to build / try a tool locally. Can you attempt to provide a relative path instead? Does it work?
If not, you can leave it as it is since there's only limited usecases where we'd want to run / build a tool locally

docker/docker-compose-dev-essentials.yaml
praveen-formido: @chandrasekharan-zipstack We have mistakenly modified the hostname from `unstract-rabbit` to `unstract-rabbitmq`. @ritwik-g pointed this out, hence the change. We use `unstract-rabbit` as a default.

tools/classifier/requirements.txt
praveen-formido: Yes @chandrasekharan-zipstack , this works locally. Tested it.

docker/docker-compose-dev-essentials.yaml
chandrasekharan-zipstack: Ah, I might have made this change initially :P I think I referred the LLMW repo back then. @ritwik-g where do we use this hostname? In all our envs we still referred to this service as unstract-rabbitmq only

tools/classifier/requirements.txt
chandrasekharan-zipstack: @pk-zipstack no, I meant doing a `pip install -r requirements.txt`. This will try to look for `/unstract/sdk1/` in your system which might not be present. The relative path might help in resolving this issue but then within the docker container it'll be an issue.
If required, you could leave a comment about this caveat but I guess no one really installs its dependencies and uses it currently

praveen-formido:
All the auto-fixable lint issues inside `sdk1` folder have been patched.

hari-kuriakose:
> All the auto-fixable lint issues inside `sdk1` folder have been patched.

Great, in that case let's pick up the remaining lint issues later:
- Outside `sdk1` folder - these already existed in the platform
- Inside `sdk1` folder - these already existed in `sdk0`

hari-kuriakose:
Remaining lint issues from the pre-commit hook will be addressed later.

> > All the auto-fixable lint issues inside `sdk1` folder have been patched.
> 
> Great, in that case let's pick up the remaining lint issues later:
> 
>     * Outside `sdk1` folder - these already existed in the platform
> 
>     * Inside `sdk1` folder - these already existed in `sdk0`

PR #1483, opened 2025-08-19
UN-2693 [FEAT] added raw view support backend oss support functionality

vishnuszipstack:
## What

- Added raw view support in hitl backend supported functions

## Why

- New feature requirement

## How

- Get raw view text from structure tools and passed to destination to push to hitl

- No

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/tools/structure/src/main.py
+++ b/tools/structure/src/main.py
@@ -325,11 +325,25 @@ class StructureTool(BaseTool):
             )
 
         # HACK: Replacing actual file's name instead of INFILE
-        if SettingsKeys.METADATA in structured_output:
-            structured_output[SettingsKeys.METADATA][SettingsKeys.FILE_NAME] = (
-                self.source_file_name
-            )
+        # Ensure metadata section exists
+        if SettingsKeys.METADATA not in structured_output:
+            structured_output[SettingsKeys.METADATA] = {}
+            self.stream_log("Created metadata section in structured_output")
 
+        structured_output[SettingsKeys.METADATA][SettingsKeys.FILE_NAME] = (
+            self.source_file_name
+        )
+
+        # Add extracted text for HITL raw view
+        if extracted_text:
+            structured_output[SettingsKeys.METADATA]["extracted_text"] = extracted_text
+            self.stream_log(
+                f"Added extracted_text to metadata (length: {len(extracted_text)} characters)"
+            )
+        else:
+            self.stream_log(
+                "No extracted_text available - STHelper.dynamic_extraction returned empty"
+            )
         if merged_metrics := self._merge_metrics(
             structured_output.get(SettingsKeys.METRICS, {}), index_metrics
         ):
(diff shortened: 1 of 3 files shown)


athul-rs (approved):
LGTM.
@vishnuszipstack Please make sure all the value we fetch from metadata is handled for None cases aswell.

athul-rs:
@vishnuszipstack Please check on the `code-rabbit` nitpick comments to verify if these are relevant.

tools/structure/src/main.py
chandrasekharan-zipstack, line 1: @vishnuszipstack you'll need to bump the tool version in its `config/properties.json`

tools/structure/src/main.py
```diff
@@ -325,11 +325,25 @@ def run(
...
 
+        # Add extracted text for HITL raw view
+        if extracted_text:
+            structured_output[SettingsKeys.METADATA]["extracted_text"] = extracted_text
+            self.stream_log(
+                f"Added extracted_text to metadata (length: {len(extracted_text)} characters)"
+            )
+        else:
+            self.stream_log(
+                "No extracted_text available - STHelper.dynamic_extraction returned empty"
```
chandrasekharan-zipstack, line 345: @vishnuszipstack rephrase this. The user will have no context on what this `STHelper.dynamic_extraction` is.
```suggestion
                "No text is extracted from the document to add to the metadata"
```

tools/structure/src/main.py
```diff
@@ -325,11 +325,25 @@ def run(
...
+
+        structured_output[SettingsKeys.METADATA][SettingsKeys.FILE_NAME] = (
+            self.source_file_name
+        )
 
+        # Add extracted text for HITL raw view
+        if extracted_text:
+            structured_output[SettingsKeys.METADATA]["extracted_text"] = extracted_text
+            self.stream_log(
+                f"Added extracted_text to metadata (length: {len(extracted_text)} characters)"
```
chandrasekharan-zipstack, line 341:
```suggestion
                f"Added text extracted from the document to metadata (length: {len(extracted_text)} characters)"
```

Deepak-Kesavan (approved):

chandrasekharan-zipstack (commented):
@vishnuszipstack why do we add the `extracted_text` in the API deployments' response? We already have some fields which provide the context and that can be used instead right? cc: @Deepak-Kesavan 
If its for line numbers for highlighting, I think we should keep it optional to not pollute the response for others

vishnuszipstack:
> @vishnuszipstack why do we add the `extracted_text` in the API deployments' response? We already have some fields which provide the context and that can be used instead right? cc: @Deepak-Kesavan If its for line numbers for highlighting, I think we should keep it optional to not pollute the response for others

It is required @chandrasekharan-zipstack for raw view. will do optional thing later.

PR #1485, opened 2025-08-20
UN-2354 [FEAT] Support presigned S3 URLs for file uploads in API deployments

pk-zipstack:
## What

- Added support for presigned s3 URLs in API deployment

## Why

- This was requested by customers.

## How

- Added a function that fetched the files from s3 presigned URLs and adds them to the `file_objs`. Which then processes the files using the base functionalities.

- Yes, this can break in the part of validating the files since this PR has modifications in Serializers. But this was tested in local and did not throw any issues.

## Database Migrations

- NO

## Env Config

- NO

## Relevant Docs

-

## Related Issues or PRs

- https://zipstack.atlassian.net/browse/UN-2354

## Dependencies Versions

-

## Notes on Testing

- Tested adding single and multiple URLs in testing. 

## Checklist

### DIFF
--- a/backend/api_v2/deployment_helper.py
+++ b/backend/api_v2/deployment_helper.py
@@ -1,9 +1,12 @@
 import logging
+from io import BytesIO
 from typing import Any
-from urllib.parse import urlencode
+from urllib.parse import urlencode, urlparse
 
+import requests
 from configuration.models import Configuration
-from django.core.files.uploadedfile import UploadedFile
+from django.conf import settings
+from django.core.files.uploadedfile import InMemoryUploadedFile, UploadedFile
 from rest_framework.request import Request
 from rest_framework.serializers import Serializer
 from rest_framework.utils.serializer_helpers import ReturnDict
@@ -25,6 +28,7 @@ from api_v2.exceptions import (
     APINotFound,
     InactiveAPI,
     InvalidAPIRequest,
+    PresignedURLFetchError,
 )
 from api_v2.key_helper import KeyHelper
 from api_v2.models import APIDeployment, APIKey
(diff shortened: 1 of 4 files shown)


backend/api_v2/deployment_helper.py
```diff
@@ -248,3 +252,168 @@ def get_execution_status(execution_id: str) -> ExecutionResponse:
...
+                status_code = 502  # Bad Gateway
+                error_msg = "Connection error"
+            else:
+                status_code = 400
+                error_msg = str(e)
+
+            raise PresignedURLFetchError(
+                url=sanitized_url, error_message=error_msg, status_code=status_code
+            )
+        except Exception as e:
```
chandrasekharan-zipstack: @pk-zipstack such catch all exception handling is undesirable usually and all we're doing is re-raising here. We can simply remove this and let the exception propagate up the call stack

backend/api_v2/deployment_helper.py
```diff
@@ -248,3 +252,168 @@ def get_execution_status(execution_id: str) -> ExecutionResponse:
...
+        parsed_url = urlparse(url)
+        sanitized_url = parsed_url._replace(query="").geturl()  # For logging
+        file_stream = None
+
+        try:
+            # Get max file size from settings
+            try:
+                max_bytes = settings.API_DEPL_PRESIGNED_URL_MAX_FILE_SIZE_MB * 1024 * 1024
+            except Exception:
+                max_bytes = 20 * 1024 * 1024  # sane default if settings unavailable
```
chandrasekharan-zipstack:
```suggestion
            max_bytes = settings.API_DEPL_PRESIGNED_URL_MAX_FILE_SIZE_MB * 1024 * 1024
```
This might not be required since we have defaults in the settings already

backend/api_v2/deployment_helper.py
```diff
@@ -248,3 +252,168 @@ def get_execution_status(execution_id: str) -> ExecutionResponse:
...
+                and hasattr(e, "response")
+                and e.response is not None
+            ):
+                status_code = e.response.status_code
+                error_msg = f"{e.response.status_code} {e.response.reason}"
+            elif isinstance(
+                e, (requests.exceptions.ConnectTimeout, requests.exceptions.ReadTimeout)
+            ):
+                status_code = 504  # Gateway Timeout
+                error_msg = "Request timed out"
```
chandrasekharan-zipstack: NIT: Consider appending the actual error as well or atleast log it here. Handle it for other scenarios too

backend/api_v2/deployment_helper.py
```diff
@@ -248,3 +252,168 @@ def get_execution_status(execution_id: str) -> ExecutionResponse:
...
+                size=downloaded,
+                charset=None,
+            )
+
+        except requests.RequestException as e:
+            if file_stream:
+                try:
+                    file_stream.close()
+                except Exception:
+                    pass
```
chandrasekharan-zipstack: Move this to a finally block instead - here we don't seem to close the file stream in case of a successful read. You might want to also log an exception if this `file_stream.close()` raises one instead of silently suppressing it

backend/api_v2/deployment_helper.py
```diff
@@ -248,3 +252,168 @@ def get_execution_status(execution_id: str) -> ExecutionResponse:
...
+            raise PresignedURLFetchError(
+                url=sanitized_url, error_message=error_msg, status_code=status_code
+            )
+        except Exception as e:
+            # Close the file stream on any other error
+            if file_stream:
+                try:
+                    file_stream.close()
+                except Exception:
+                    pass
```
chandrasekharan-zipstack: Adding a single finally block to close the `file_stream` would do

backend/api_v2/deployment_helper.py
```diff
@@ -248,3 +252,168 @@ def get_execution_status(execution_id: str) -> ExecutionResponse:
...
+            file_objs (list[UploadedFile]): List to append the fetched files to
+        """
+        for url in presigned_urls:
+            try:
+                uploaded_file = DeploymentHelper.fetch_presigned_file(url)
+                file_objs.append(uploaded_file)
+            except Exception as e:
+                logger.error(f"Error fetching file from URL: {e}")
+                # Re-raise the error to be handled by the caller
+                raise
```
chandrasekharan-zipstack:
```suggestion
            uploaded_file = DeploymentHelper.fetch_presigned_file(url)
            file_objs.append(uploaded_file)
```
We anyway log errors with a middleware so this is redundant if we're only doing this without adding additional information. Also all we're doing is simply re-raising it

Deepak-Kesavan (approved):

chandrasekharan-zipstack (approved):

PR #1503, opened 2025-08-25
UN-2733 [FIX] Allow file processing to continue on cleanup related errors

chandrasekharan-zipstack:
## What

- Fixed error handling logic which was meant to gracefully handle errors on file processing
- Suppressed errors that arise while performing a cleanup

## Why

- Error handling code raised another error

## How

- Caught errors during cleanup

- No, only suppressed an error and added logs

## Related Issues or PRs

- [UN-2733](https://zipstack.atlassian.net/browse/UN-2733)

## Notes on Testing

- No explicit testing done since it was a simple logical change

## Checklist

[UN-2733]: https://zipstack.atlassian.net/browse/UN-2733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/connector_processor/connector_processor.py
+++ b/backend/connector_processor/connector_processor.py
@@ -20,7 +20,6 @@ from unstract.connectors.connectorkit import Connectorkit
 from unstract.connectors.enums import ConnectorMode
 from unstract.connectors.exceptions import ConnectorError, FSAccessDeniedError
 from unstract.connectors.filesystems.ucs import UnstractCloudStorage
-from unstract.connectors.queues.redis import RedisQueue
 
 logger = logging.getLogger(__name__)
 
@@ -72,8 +71,10 @@ class ConnectorProcessor:
         """Function to return list of all supported connectors except PCS."""
         supported_connectors = []
         updated_connectors = []
-        # Connectors that are marked active but not supported explicitly
-        unsupported_connectors = [RedisQueue.get_id()]
+
+        # HACK: Connectors that are marked active but not supported explicitly
+        # TODO: Remove RedisQueue from the list of connectors and use separately instead
+        unsupported_connectors = ["redisqueue|79e1d681-9b8b-4f6b-b972-1a6a095312f45"]
 
         if type == ConnectorKeys.INPUT:
             updated_connectors = fetch_connectors_by_key_value(


backend/workflow_manager/workflow_v2/file_execution_tasks.py
```diff
@@ -557,7 +557,7 @@ def _process_file(
                 result=result,
                 workflow_file_execution=None,
                 error=error_msg,
-                is_api=destination.is_api,
+                is_api=destination.is_api if destination else False,
```
ritwik-g, line 560: should we make the condition to use `hasattr` to be sure that the `is_api` attr is available?

backend/workflow_manager/workflow_v2/file_execution_tasks.py
```diff
@@ -1136,10 +1136,17 @@ def _build_final_result(
...
-                f"Deleting file execution directory for file: '{file_hash.file_name}'"
-            )
-            destination.delete_file_execution_directory()
+            try:
+                logger.info(
+                    f"Deleting file execution directory for file: '{file_hash.file_name}'"
+                )
+                destination.delete_file_execution_directory()
+            except Exception as cleanup_error:
+                logger.warning(
```
ritwik-g: @chandrasekharan-zipstack should we use `logger.exception` to print the entire stack trace? To know why it failed?

ritwik-g (approved):
@chandrasekharan-zipstack I have 2 minor comments. Other than that it looks good

Deepak-Kesavan (approved):
looks good

gaya3-zipstack (commented):
@chandrasekharan-zipstack I understand that this PR is handling exceptions while unable to access Minio for cleanup. What is our error handling whan for example minio becomes inaccessible during other stages? Like trying to access a file for reading/writing? In such cases also, will we run into a similar state where the file with the error is in PENDING state forever and other files need another round to be picked up?

chandrasekharan-zipstack:
> @chandrasekharan-zipstack I understand that this PR is handling exceptions while unable to access Minio for cleanup. What is our error handling whan for example minio becomes inaccessible during other stages? Like trying to access a file for reading/writing? In such cases also, will we run into a similar state where the file with the error is in PENDING state forever and other files need another round to be picked up?

Yes, this should handle all such cases too as long as it happens within a file's execution. Basically the exception handler will catch it and mark the status appropriately. This particular issue was because the exception handling code, in turn raised an exception which will no longer be the case

backend/workflow_manager/workflow_v2/file_execution_tasks.py
chandrasekharan-zipstack: Might not be required since we add this attribute to its constructor and will always have a default. [Reference](https://github.com/Zipstack/unstract/blob/4071ff4ed37b4df2e5f850ce4f4b7632ba65b90f/backend/workflow_manager/endpoint_v2/destination.py#L78-L80)

PR #1508, opened 2025-08-28
UN-2776 [FIX] Fixed GDrive connector showing always authenticated in workflow settings

kirtimanmishrazipstack:
## What

- Fix GDrive connector showing always authenticated in workflow settings
 
## Why

- While gdrive oauth is successful, we are storing `oauth-cachekey` cookie to local storage
- Although we have a centralized connector, but still we didn't segregate the  `oauth-cachekey` based on workflow-id and connector type
- Due to this if we have an existing gdrive connector which is authenticated, selecting a new connector would also make it authenticated accross workflows and between source and destination

## How

- Deleting cache from local-storage, and redis cache (already happeing)

### Frontend  change:
For new connector:
- We should show Signin when user tries to do oauth. 

For existing connector:
- We should show reauthenticate when user tries to do oauth. 

For both connectors:
- When we receive the cache key, using it we can do test-connection and submit
- Once submitted or modal is closed, we would remove from local storage and cache. 

### Backend  change:
- Handle the scenario when user tries to submit connector with same name. Backend throws server error. But we have deleted cache.
- Although there is successful test-connection, still we need to reauthenticate again. This is a bug.
- Handled this deleting the oauth cache key only when connector is added/updated.

-

## Database Migrations

- N/A

## Env Config

- N/A

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/frontend/src/components/oauth-ds/google/GoogleOAuthButton.jsx
+++ b/frontend/src/components/oauth-ds/google/GoogleOAuthButton.jsx
@@ -5,7 +5,11 @@ import { GoogleLoginButton } from "react-social-login-buttons";
 
 import "./GoogleOAuthButton.css";
 
-const GoogleOAuthButton = ({ handleOAuth, status, buttonText = "Authenticate with Google" }) => {
+const GoogleOAuthButton = ({
+  handleOAuth,
+  status,
+  buttonText = "Authenticate with Google",
+}) => {
   const [text, setText] = useState("");
   useEffect(() => {
     if (status === "success") {
--- a/frontend/src/components/oauth-ds/oauth-ds/OAuthDs.jsx
+++ b/frontend/src/components/oauth-ds/oauth-ds/OAuthDs.jsx
@@ -23,7 +23,9 @@ function OAuthDs({
   const oauthStatusKey = `oauth-status-${selectedSourceId}`;
 
   // Determine button text based on connector state
-  const buttonText = isExistingConnector ? "Reauthenticate" : "Authenticate with Google";
+  const buttonText = isExistingConnector
+    ? "Reauthenticate"
+    : "Authenticate with Google";
 
   const [oauthStatus, setOAuthStatus] = useState(() => {
     // Initialize from connector-specific status


frontend/src/components/input-output/configure-ds/ConfigureDs.jsx
```diff
@@ -37,6 +38,7 @@ function ConfigureDs({
   const [cacheKey, setCacheKey] = useState("");
   const [status, setStatus] = useState("");
   const { sessionDetails } = useSessionStore();
+  const { projectId: workflowId } = useWorkflowStore();
```
chandrasekharan-zipstack: @kirtimanmishrazipstack won't this fail if a connector is created from outside a workflow?

vishnuszipstack (approved):

backend/connector_v2/views.py
```diff
@@ -84,23 +82,36 @@ def _get_connector_metadata(self, connector_id: str) -> dict[str, str] | None:
...
             logger.info(f"Fetching oauth data for {connector_id}")
             oauth_key = self.request.query_params.get(ConnectorAuthKey.OAUTH_KEY)
-            if oauth_key is None:
-                raise MissingParamException(param=ConnectorAuthKey.OAUTH_KEY)
-            # Preserve OAuth cache for reuse across multiple operations (Test Connection, Submit,
-            # File System browsing). Frontend localStorage stores cache keys which must correspond
-            # to persistent backend credentials for tab switching and repeated operations to work.
+            if not oauth_key:
+                raise MissingParamException(
+                    "OAuth authentication required. Please sign in with Google first."
```
chandrasekharan-zipstack, line 87: @kirtimanmishrazipstack let's not mention "Google" in the error messages - we'll have to correct it when we choose to support other OAuth providers. Correct this every place you've mentioned Google

backend/connector_v2/views.py
```diff
@@ -84,23 +82,36 @@ def _get_connector_metadata(self, connector_id: str) -> dict[str, str] | None:
...
+            return
+
+        oauth_key = self.request.query_params.get(ConnectorAuthKey.OAUTH_KEY)
+        if oauth_key:
+            logger.info(f"Cleaning up OAuth cache for {connector_id}")
+            ConnectorAuthHelper.get_oauth_creds_from_cache(
+                cache_key=oauth_key,
+                delete_key=True,  # Delete after successful operation
+            )
+
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Consider if this suggestion can be valid

frontend/src/components/oauth-ds/google/GoogleOAuthButton.jsx
```diff
@@ -5,15 +5,15 @@ import { Typography } from "antd";
 
 import "./GoogleOAuthButton.css";
 
-const GoogleOAuthButton = ({ handleOAuth, status }) => {
+const GoogleOAuthButton = ({ handleOAuth, status, buttonText = "Signin" }) => {
```
chandrasekharan-zipstack: @kirtimanmishrazipstack maybe we should say "Authenticate with Google" instead of "Signin" since we aren't necessarily performing a sign in operation

chandrasekharan-zipstack (approved):

PR #1515, opened 2025-09-01
UN-2354 [FIX] - Fix premature file stream closure in presigned URL handling for API Deployments

pk-zipstack:
## What

- Fixed "I/O operation on closed file" error when processing presigned S3 URLs in API deployments
- Implemented proper file stream lifecycle management to prevent memory leaks with multiple file uploads
- Added explicit file closure after file processing is complete

## Why

- When uploading files via presigned S3 URLs, the API was failing with ValueError: I/O operation on closed file
- The finally block was closing the BytesIO stream immediately after creating the InMemoryUploadedFile object

## How

- Added a finally block that only closes the stream if the InMemoryUploadedFile wasn't created successfully (when file_stream is still not None).
- Added explicit file closure in `execute_workflow()` after `SourceConnector.add_input_file_to_api_storage()` completes.

- No, this PR will not break any existing features. 
The changes are additive and defensive:
  1. File closure additions: The explicit file closure after processing is safe for all file types (InMemoryUploadedFile, TemporaryUploadedFile) and follows Django best practices
  2. Regular file uploads: Unchanged behavior for standard form-data file uploads through Postman/API clients
  3. Error handling: Enhanced error handling in the finally block only triggers in failure cases

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

- Test by uploading files via presigned S3 URLs to API deployments
- Test regular file uploads via form-data to ensure no regression
- Verify that files process successfully without "I/O operation on closed file" errors

## Checklist

### DIFF
--- a/backend/api_v2/deployment_helper.py
+++ b/backend/api_v2/deployment_helper.py
@@ -403,12 +403,6 @@ class DeploymentHelper(BaseAPIKeyValidator):
                 url=sanitized_url, error_message=error_msg, status_code=status_code
             )
 
-        finally:
-            if file_stream:
-                try:
-                    file_stream.close()
-                except Exception as e:
-                    logger.warning(f"Failed to close file stream: {str(e)}")
 
     @staticmethod
     def load_presigned_files(
--- a/backend/api_v2/deployment_helper.py
+++ b/backend/api_v2/deployment_helper.py
@@ -222,6 +222,15 @@ class DeploymentHelper(BaseAPIKeyValidator):
             file_objs=file_objs,
             use_file_history=use_file_history,
         )
+        
+        # Close uploaded files after they've been stored to free memory
+        # This is important for presigned URL files to prevent memory leaks
+        for file_obj in file_objs:
+            try:
+                if hasattr(file_obj, 'close'):
+                    file_obj.close()
+            except Exception as e:
+                logger.warning(f"Failed to close uploaded file {getattr(file_obj, 'name', 'unknown')}: {str(e)}")
         try:
             result = WorkflowHelper.execute_workflow_async(
                 workflow_id=workflow_id,
@@ -364,8 +373,8 @@ class DeploymentHelper(BaseAPIKeyValidator):
                 f"Fetched file '{filename}' with MIME type '{content_type}' from presigned URL {sanitized_url}"
             )
 
-            # Return as an InMemoryUploadedFile
-            return InMemoryUploadedFile(
+            # Create InMemoryUploadedFile with proper stream management
+            uploaded_file = InMemoryUploadedFile(
                 file=file_stream,
                 field_name="file",
                 name=filename,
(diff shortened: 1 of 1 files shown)


chandrasekharan-zipstack (approved):
@pk-zipstack do check with @muhammad-ali-e as well regarding this change

backend/api_v2/deployment_helper.py
```diff
@@ -222,6 +222,17 @@ def execute_workflow(
             file_objs=file_objs,
             use_file_history=use_file_history,
         )
+
```
gaya3-vijayakumar, line 225: @pk-zipstack If there is any exception being thrown before this, will the file stream be closed safley?

backend/api_v2/deployment_helper.py
```diff
@@ -408,7 +424,9 @@ def fetch_presigned_file(url: str) -> InMemoryUploadedFile:
                 try:
                     file_stream.close()
                 except Exception as e:
-                    logger.warning(f"Failed to close file stream: {str(e)}")
+                    logger.warning(
+                        f"Failed to close file stream during cleanup: {str(e)}"
```
gaya3-vijayakumar: Would be good to print the file name also if possible to keep the log useful However if the exception trace is already covering it, pls ignore.

backend/api_v2/deployment_helper.py
praveen-formido: Yes @gaya3-zipstack, the WSGI handler will call the `request.close()` after request processing.

backend/api_v2/deployment_helper.py
praveen-formido: Good suggestion. Will implement.

backend/api_v2/deployment_helper.py
gaya3-vijayakumar: Looking at the code for function call references, looks like garbage collection of file_objs will be handled by Django and explicit closing is an added guardrail in happy path. add_input_file_to_api_storage() takes care of manual copy of file to the destination and hence this should be safe.

gaya3-vijayakumar: CC: @pk-zipstack

backend/api_v2/deployment_helper.py
```diff
@@ -222,6 +222,17 @@ def execute_workflow(
...
+        # Close uploaded files after they've been stored to free memory
+        # This is important for presigned URL files to prevent memory leaks
+        for file_obj in file_objs:
+            try:
+                if hasattr(file_obj, "close"):
+                    file_obj.close()
+            except Exception as e:
+                logger.warning(
+                    f"Failed to close uploaded file {getattr(file_obj, 'name', 'unknown')}: {str(e)}"
+                )
```
muhammad-ali-e: **NIT:** The `hasattr(file_obj, "close")` check might be unnecessary here since all expected 
file objects (UploadedFile, InMemoryUploadedFile) should have close() methods. 

Note:  Keep it there if you think it required.

Consider simplifying to like following codeblock:
```python
for file_obj in file_objs:
    try:
        file_obj.close()
    except Exception as e:
        logger.warning(f"Failed to close file object {type(file_obj)}: {e}")

praveen-formido: Removing this since the file_obj.close() serves no purpose.

PR #1476, opened 2025-08-11
UN-2682 [FEAT] Centralize connector management

chandrasekharan-zipstack:
## What

- Connector management handled centrally similar to other adapters in the platform (LLM, embedding .etc.)
- Improved workflow UX to add / use these centralized connectors
- UX change to add folders and configure the source / destination correctly
- Clean up unused code from old connector implementation

## Why

- **Improved User Experience**: Users can now configure connectors directly within workflows without navigating to separate pages
- **Reduced Complexity**: Eliminates the need for separate input/output pages and simplifies the connector management flow
- **Better Data Safety**: Auto-save prevents accidental loss of configuration changes
- **Maintainability**: Centralized approach reduces code duplication and makes the system easier to maintain

#### Backend Critical Files:
1. backend/connector_v2/migrations/0003_migrate_to_centralized_connectors.py (+289/-0)
   - Data migration script - Critical for data integrity
   - Must verify migration logic handles existing connectors correctly

2. backend/connector_processor/connector_processor.py (+30/-23)
   - Core connector processing logic - Business logic changes
   - Verify sharing and validation logic modifications

3. backend/connector_processor/views.py (+47/-45)
   - API endpoint changes - Could affect API contracts
   - Review backward compatibility implications

4. backend/connector_v2/models.py (+27/-12)
   - Database schema/model changes - Could affect data structure
   - Verify constraint and field changes

#### Frontend Critical Files:
5. frontend/src/components/agency/configure-connector-modal/ConfigureConnectorModal.jsx (+445/-162)
   - Major component refactor - Largest single file change (607 total changes)
   - Core UI logic for connector configuration - needs thorough testing

6. frontend/src/components/agency/ds-settings-card/DsSettingsCard.jsx (+51/-172)
   - Significant component changes - Large deletion count suggests major refactor
   - User interface changes affecting workflow

### MEDIUM PRIORITY - Integration & New Features

7. backend/connector_v2/migrations/0004_remove_connectorinstance_unique_workflow_connector_and_more.py (+49/-0)
   - Constraint changes - Could affect data integrity

8. frontend/src/components/agency/configure-forms-layout/ConfigureFormsLayout.jsx (+18/-73)
   - Layout component changes - High deletion ratio suggests major restructuring

9. frontend/src/components/connectors/connector-list-modal/ConnectorListModal.jsx  (+149/-0)
   - New component - Entirely new functionality needs review

### Backend Changes:
- Enhanced connector processor with improved sharing and validation logic
[...]

PR #1517, opened 2025-09-01
[MISC] Add migration to fix unintended connector sharing from centralization

chandrasekharan-zipstack:
## What

- Added a new data migration (`0005_fix_unintended_connector_sharing.py`) to fix an issue where connectors were inadvertently shared between users after the centralization migration

## Why

- The centralization migration (0003) grouped connectors by organization, connector type, and metadata hash but missed including `created_by` in the grouping logic
- This caused connectors with identical credentials from different users to be merged together
- Users ended up with workflow endpoints pointing to connectors owned by other users without explicit sharing permission
- This resulted in blank connectors appearing in workflow pages due to permission issues
image

## How

The migration:
1. **Identifies problem endpoints** - Finds WorkflowEndpoint instances where:
   - The connector's `created_by` != workflow's `created_by`
   - The connector is NOT in the workflow owner's `shared_users` list
   - The connector is NOT `shared_to_org = True`

2. **Creates user-specific duplicates** - For each problem case:
   - Creates a duplicate connector owned by the workflow owner
   - Copies all metadata and settings from the original connector
   - Updates the WorkflowEndpoint to point to the new connector

3. **Preserves explicit sharing** - Leaves alone:
   - Connectors explicitly shared via `shared_users`
   - Connectors shared to entire organization
   - Connectors already owned by the workflow owner

No, this PR should not break existing features:
- The migration only creates new connector instances, it doesn't modify or delete existing ones
- Explicitly shared connectors remain untouched
- Original connectors are preserved for rollback if needed
- The migration uses database transactions for consistency
- The reverse migration logs a warning but doesn't attempt risky automatic cleanup

## Database Migrations

- ✅ New migration: `backend/connector_v2/migrations/0005_fix_unintended_connector_sharing.py`
- Migration creates new ConnectorInstance records with proper ownership
- Updates WorkflowEndpoint foreign keys to point to the new connectors
- Reverse migration is a no-op with warning (full reversal not feasible without data loss risk)

## Env Config

- None required

## Relevant Docs

- [PR #1476](https://github.com/Zipstack/unstract/pull/1476) - Original centralization PR that introduced the issue

## Related Issues or PRs

- Related to PR #1476 which introduced connector centralization
- This PR fixes an unintended side effect of that centralization

## Dependencies Versions

- No dependency changes

## Notes on Testing

1. **Before migration**: Check for workflows with connectors owned by different users
2. **Run migration**: `python manage.py migrate connector_v2`
3. **After migration**: 
   - Verify each user has their own connector instances
   - Check that workflows now show correct connectors (no blank entries)
   - Confirm explicitly shared connectors remain shared
   - Verify org-wide shared connectors are unaffected

## Checklist

### DIFF
--- a/backend/connector_v2/migrations/0005_fix_unintended_connector_sharing.py
+++ b/backend/connector_v2/migrations/0005_fix_unintended_connector_sharing.py
@@ -66,9 +66,13 @@ def identify_problem_endpoints(apps) -> list[Any]:
         # This is a problem case - connector is implicitly shared
         problem_endpoints.append(endpoint)
         logger.info(
-            f"Found problem: Workflow '{workflow.workflow_name}' (owner: {workflow.created_by.email}) "
-            f"uses connector '{connector.connector_name}' (owner: {connector.created_by.email}) "
-            f"without explicit sharing"
+        logger.info(
+            "Found problem: workflow_id=%s (owner_id=%s) endpoint_id=%s uses connector_id=%s (owner_id=%s) without explicit sharing",
+            workflow.id,
+            workflow.created_by_id,
+            endpoint.id,
+            connector.id,
+            connector.created_by_id,
         )
 
     return problem_endpoints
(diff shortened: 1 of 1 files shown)


vishnuszipstack:
@chandrasekharan-zipstack check codrabit comment. Address if valid

kirtimanmishrazipstack (approved):
@chandrasekharan-zipstack please fix pre commit issues

PR #1539, opened 2025-09-16
fix reverse proxy mapping and traefik disabled

Fra5792:
## What

An issue was causing reverse-proxy a result of bad gateway

## Why

Traefik and port mapping issue


ritwik-g:
@Fra5792 the changes as such looks okay. Do you have any associated issues with this? Also can you add some screenshots of before and after of these changes?

Also please accept the CLA.

PR #1541, opened 2025-09-18
UN-2807 [FEAT] Add packet processing support for HITL workflows

jagadeeswaran-zipstack:
## Summary
- Added packet-based processing support for HITL workflows
- Enables grouping of documents into packets for batch review
- Maintains backward compatibility with existing HITL queue implementation

## Changes
- Added `packet_id` parameter to API deployment execution flow
- Enhanced destination connector to route documents to packet queues when packet_id is provided
- Updated DTOs and serializers to include packet_id field throughout the workflow execution chain

## Technical Details
The implementation allows API consumers to specify a `packet_id` during workflow execution, which will:
1. Override the default HITL queue routing
2. Group documents into the specified packet for batch review
3. Use the `PacketQueueUtils` for packet-based queue management

## Test Plan
- [ ] Test API execution without packet_id (should use existing HITL flow)
- [ ] Test API execution with packet_id (should route to packet queue)
- [ ] Verify packet grouping works correctly for multiple documents
- [ ] Ensure backward compatibility with existing HITL workflows
- [ ] Test error handling when invalid packet_id is provided

## Related
- Ticket: UN-2807

### DIFF
--- a/backend/api_v2/api_deployment_views.py
+++ b/backend/api_v2/api_deployment_views.py
@@ -70,6 +70,7 @@ class DeploymentExecution(views.APIView):
         tag_names = serializer.validated_data.get(ApiExecution.TAGS)
         llm_profile_id = serializer.validated_data.get(ApiExecution.LLM_PROFILE_ID)
         hitl_queue_name = serializer.validated_data.get(ApiExecution.HITL_QUEUE_NAME)
+        packet_id = serializer.validated_data.get(ApiExecution.PACKET_ID)
 
         if presigned_urls:
             DeploymentHelper.load_presigned_files(presigned_urls, file_objs)
@@ -85,6 +86,7 @@ class DeploymentExecution(views.APIView):
             tag_names=tag_names,
             llm_profile_id=llm_profile_id,
             hitl_queue_name=hitl_queue_name,
+            packet_id=packet_id,
             request_headers=dict(request.headers),
         )
         if "error" in response and response["error"]:
--- a/backend/workflow_manager/endpoint_v2/destination.py
+++ b/backend/workflow_manager/endpoint_v2/destination.py
@@ -6,6 +6,7 @@ import os
 from typing import Any
 
 from connector_v2.models import ConnectorInstance
+from pluggable_apps.manual_review_v2.packet_queue_utils import PacketQueueUtils
 from plugins.workflow_manager.workflow_v2.utils import WorkflowUtil
 from rest_framework.exceptions import APIException
 from usage_v2.helper import UsageHelper
@@ -64,6 +65,7 @@ class DestinationConnector(BaseConnector):
         use_file_history: bool,
         file_execution_id: str | None = None,
         hitl_queue_name: str | None = None,
+        packet_id: str | None = None,
     ) -> None:
         """Initialize a DestinationConnector object.
 
@@ -82,6 +84,7 @@ class DestinationConnector(BaseConnector):
         self.workflow_log = workflow_log
         self.use_file_history = use_file_history
         self.hitl_queue_name = hitl_queue_name
+        self.packet_id = packet_id
         self.workflow = workflow
 
     def _get_endpoint_for_workflow(
@@ -170,6 +173,19 @@ class DestinationConnector(BaseConnector):
                 file_execution_id=file_execution_id,
             )
             logger.info(f"Successfully pushed {file_name} to HITL queue")
+            return True
+
+        if self.packet_id:
+            logger.info(
+                f"API packet override: pushing to packet queue for file {file_name}"
+            )
+            self._push_data_to_queue(
+                file_name=file_name,
+                workflow=workflow,
+                input_file_path=input_file_path,
+                file_execution_id=file_execution_id,
+            )
+            logger.info(f"Successfully pushed {file_name} to packet queue")
             return True
 
         # Skip HITL validation if we're using file_history and no execution result is available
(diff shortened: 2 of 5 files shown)


backend/api_v2/api_deployment_views.py
```diff
@@ -70,6 +70,7 @@ def post(
         tag_names = serializer.validated_data.get(ApiExecution.TAGS)
         llm_profile_id = serializer.validated_data.get(ApiExecution.LLM_PROFILE_ID)
         hitl_queue_name = serializer.validated_data.get(ApiExecution.HITL_QUEUE_NAME)
+        packet_id = serializer.validated_data.get(ApiExecution.PACKET_ID)
```
chandrasekharan-zipstack: NIT: @jagadeeswaran-zipstack please check with @shuveb if we can prefix this with parameter with `hitl_` to indicate that it only applies to hitl easily. I assume that this parameter is not generic enough to find other usecases

backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -868,6 +886,18 @@ def _push_to_queue(
...
+            # Check if this is a packet-based execution
+            if self.packet_id:
+                # Route to packet queue instead of regular HITL queue
+                success = PacketQueueUtils.enqueue_to_packet(
+                    packet_id=self.packet_id, queue_result=queue_result
+                )
+                if success:
+                    logger.info(f"Pushed {file_name} to packet {self.packet_id}")
+                else:
+                    logger.error(f"Failed to push {file_name} to packet {self.packet_id}")
```
chandrasekharan-zipstack: @jagadeeswaran-zipstack 
1. How will this error affect the user? Consider logging it to the user as well. Since we log it as an error, please highlight in this message that we continue the execution. 
1. Where do we log the actual error / traceback? Do we still need this log here if that's the case?

chandrasekharan-zipstack (commented):
Left some minor comments apart from that, I have these concerns
1. What's the difference between this `packet_id` and `hitl_queue_name` - if its only to group API deployment calls, can't we use existing params like `tag` or simply send it to a given `hitl_queue_name`?
1. Check the code rabbit comments to see if they're valid
1. Make sure to raise a PR and keep it ready in the docs repo, it should be merged when this feature goes in

backend/api_v2/api_deployment_views.py
jaags-dev: Yes @chandrasekharan-zipstack , this makes more sense. I'll make that change.

backend/workflow_manager/endpoint_v2/destination.py
jaags-dev: No, added these for dev purpose. I'll remove it.

backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -6,6 +6,7 @@
 from typing import Any
 
 from connector_v2.models import ConnectorInstance
+from pluggable_apps.manual_review_v2.packet_queue_utils import PacketQueueUtils
```
vishnuszipstack: @jagadeeswaran-zipstack i think this will throw error in oss. Can u check how redisQueue and follow the same approach?

jaags-dev: We construct special queue names(since it has multiple states) e.g., queue_names["pending"], so this function is required. I added a try block to catch and return early for OSS users, indicating that this is an enterprise feature.

vishnuszipstack: @jagadeeswaran-zipstack use registry instead of using try catch.

backend/workflow_manager/endpoint_v2/destination.py
```diff
@@ -839,99 +856,199 @@ def _push_to_queue(
...
+        if self.packet_id:
+            if not apps.is_installed("pluggable_apps.manual_review_v2"):
+                raise ValueError(
+                    "Packet-based HITL processing requires Unstract Enterprise. "
+                    "This feature is not available in the OSS version."
+                )
+            # Route to packet queue
+            from pluggable_apps.manual_review_v2.packet_queue_utils import (
+                PacketQueueUtils,
+            )
```
vishnuszipstack: @jagadeeswaran-zipstack move this to queue_utils don't use try catch in this main file. refer QueuUtils.

vishnuszipstack (changes requested):

backend/workflow_manager/endpoint_v2/destination.py
jaags-dev: Done 👍

workers/shared/workflow/destination_connector.py
```diff
@@ -1320,21 +1419,23 @@ def _push_data_to_queue(
...
+                self.hitl_packet_id if self.hitl_packet_id else queue_name
+            )
             log_file_info(
                 self.workflow_log,
                 file_execution_id,
-                f"✅ File '{file_name}' sent to manual review queue '{queue_name}'",
-            )
-
-            logger.info(
-                f"✅ MANUAL REVIEW: File '{file_name}' sent to manual review queue '{queue_name}' successfully"
```
vishnuszipstack, line 1504: @jagadeeswaran-zipstack keep this text itself. This is more catchy.

jaags-dev: It is there in this function `_enqueue_to_packet_or_regular_queue` refactored a bit to make it more readable.

PR #1563, opened 2025-10-07
[MISC] Correct chord retry env parsing

chandrasekharan-zipstack:
## What

- Parsed env for `CELERY_RESULT_CHORD_RETRY_INTERVAL` by float instead of int

## Why

- Parsing by int threw an error when the [sample.env had a float value](https://github.com/Zipstack/unstract/blob/0c5997f9a9da45f256df57b211e6660ccd3d062c/workers/sample.env#L147)

- No, minor parsing change only that's backward compatible

## Checklist


muhammad-ali-e (approved):
We might added  this changes in new worker. And I belive theretry interval allways to be same, But @chandrasekharan-zipstack mentioned it defined as float in doc

PR #1579, opened 2025-10-13
UN-2859 [FEAT] Smart table extractor - Implementation in API deployments for Excel files

harini-venkataraman:
## What
Added Smart Table Extractor plugin for extracting structured data from Excel files using LLM-based processing in API deployments - This PR integrates it. 
...

## Why
Users need to extract tabular data from Excel files and transform it into structured JSON format aligned with custom schemas.

...

## How

OSS PR to integrate the feature
...

No, this PR does not break existing features because:

It is a new plugin module with no dependencies on existing extraction logic.

...

## Relevant Docs
[README.md] – Plugin overview and architecture
...

## Related Issues or PRs
UN-2859
...

json-repair>=0.25.0 – For robust JSON parsing

...

## Notes on Testing
...

## Screenshots
image

...

## Checklist

### DIFF
--- a/backend/workflow_manager/execution/filter.py
+++ b/backend/workflow_manager/execution/filter.py
@@ -1,12 +1,12 @@
 from api_v2.models import APIDeployment
-from django.db.models import QuerySet
+from django.db.models import Exists, OuterRef, QuerySet
 from django_filters import rest_framework as filters
 from pipeline_v2.models import Pipeline
 from utils.date import DateRangePresets, DateTimeProcessor
 
 from workflow_manager.execution.enum import ExecutionEntity
 from workflow_manager.workflow_v2.enums import ExecutionStatus
-from workflow_manager.workflow_v2.models import Workflow, WorkflowExecution
+from workflow_manager.workflow_v2.models import WorkflowExecution
 
 
 class ExecutionFilter(filters.FilterSet):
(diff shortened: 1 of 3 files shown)


prompt-service/src/unstract/prompt_service/services/answer_prompt.py
```diff
@@ -264,6 +264,69 @@ def extract_table(
...
+                schema_data = repair_json_with_best_structure(prompt)
+                # Check if the result is a valid dict (schema object)
+                if isinstance(schema_data, dict) and schema_data:
+                    has_valid_schema = True
+                    app.logger.info(
+                        "Valid schema detected in prompt, using Smart Table Extractor"
+                    )
+            except Exception as e:
+                app.logger.debug(f"Prompt does not contain valid schema: {e}")
+
```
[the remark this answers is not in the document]
harini-venkataraman: In this case, for prompts without schema we have to redirect to generic table. Hence this is okay

tools/structure/src/utils.py
```diff
@@ -29,3 +31,30 @@ def json_to_markdown(data: Any, level: int = 0, parent_key: str = "") -> str:
         markdown += f"{indent}- {data}\n"
 
     return markdown
+
+
+def repair_json_with_best_structure(json_str: str) -> Any:
```
gaya3-vijayakumar, line 36: I think we are doing very specific actions here based on experience. So we need to really document this. Eg. I got this description from AI. Please feel free to alter accordingly
The function repair_json_with_best_structure attempts to repair and parse a possibly broken or malformed JSON string in a robust way. It does this by:

Trying to repair and parse the input string as-is.
Trying to repair and parse the input string wrapped in square brackets (as a list).
Comparing the results of both attempts:
If both are strings (i.e., both failed to parse as objects), it returns the first result.
If only one is a string, it returns the other (the one that parsed as an object).
If the wrapped version is a list of one item and that item matches the as-is result, it returns the as-is result.
If the as-is result is a dict or list and the wrapped version is a longer list, it returns the wrapped version; otherwise, it returns the as-is result.
If none of the above, it returns the wrapped version.
This approach increases the chances of successfully recovering a usable Python object from a broken JSON string, handling both object and array top-level structures.

tools/structure/src/main.py
```diff
@@ -116,6 +116,38 @@ def _override_section(
...
+        # Check if any output has table_settings with valid JSON prompt
+        for output in outputs:
+            if SettingsKeys.TABLE_SETTINGS in output:
+                prompt = output.get(SettingsKeys.PROMPTX, "")
+                if prompt and isinstance(prompt, str):
+                    try:
+                        # Try to parse the prompt as JSON
+                        schema_data = repair_json_with_best_structure(prompt)
+                        # If it's a valid dict (schema object), skip extraction
+                        if isinstance(schema_data, dict) and schema_data:
```
gaya3-vijayakumar: Should the if conditionals be reversed here?
Also we do a pass for any exception. Not a good practice.

tools/structure/src/main.py
```diff
@@ -258,7 +302,7 @@ def run(
             )
             payload[SettingsKeys.FILE_HASH] = summarize_file_hash
             payload[SettingsKeys.FILE_PATH] = summarize_file_path
-        elif not is_single_pass_enabled:
+        elif not is_single_pass_enabled and not skip_extraction_and_indexing:
```
gaya3-vijayakumar: Earlier if single pass is not enabled, then the elif block used to get executed. Now we added an additional conditional here and so even if single pass is not enabled, we might not be execuing this block. In such a case, where are we taking care of the case where single pass is not enabled and skip_extraction_and_indexing is True? Is this a valid case?

harini-venkataraman: @gaya3-zipstack If single pass is not enabled, and if skip extraction flag is true, the elif block will get executed. That is the only case where the block should ideally be executed. And, in any case where single pass is enabled, indexing should not happen.

tools/structure/src/main.py
harini-venkataraman: addressed

tools/structure/src/utils.py
harini-venkataraman: added a detailed doc string for ease of dev

tools/structure/src/main.py
```diff
@@ -116,6 +116,38 @@ def _override_section(
...
+                if prompt and isinstance(prompt, str):
+                    try:
+                        # Try to parse the prompt as JSON
+                        schema_data = repair_json_with_best_structure(prompt)
+                        # If it's a valid dict (schema object), skip extraction
+                        if isinstance(schema_data, dict) and schema_data:
+                            return True
+                    except Exception:
+                        pass
+        return False
```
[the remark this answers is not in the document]
harini-venkataraman: since fallback is needed, handled exception by warning and logging

PR #1584, opened 2025-10-13
UN-2868 [FIX] Restrict workflow modifications to owners in shared workflows

johnyrahul:
## Summary
This PR implements ownership restrictions for shared workflows to prevent non-owners from modifying workflow configurations. This ensures data integrity and proper access control when workflows are shared across users.

### 1. Display Tool Name Instead of Tool ID
- Added `getToolName()` helper function to resolve tool names from tool IDs
- Implements three-tier fallback strategy:
  - First checks tool_instances from backend (includes name even if tool is no longer exported)
  - Falls back to exportedTools list
  - Finally displays tool ID if name cannot be resolved

### 2. Restrict Tool Modification to Workflow Owners
- Added `isWorkflowOwner()` helper function to compare current user ID with workflow creator ID
- Handles type differences between user IDs (integer vs string) by converting both to strings
- Disabled "Change Prompt Studio project" button for non-owners
- Disabled "Configure Settings" button for non-owners

### 3. Restrict Connector Configuration to Workflow Owners
- Extended ownership restrictions to source and destination connectors
- Disabled connector type dropdown for non-owners
- Disabled Configure button for non-owners
- Added helpful tooltip messages explaining ownership requirements
- Properly passed `isWorkflowOwner` prop through component hierarchy:
  - Agency → WorkflowCard → DsSettingsCard

## Technical Details
- All ownership checks use string comparison to handle type differences between UUID strings and integers
- Components properly handle cases where workflow data hasn't loaded yet
- Maintains backward compatibility with existing deployment restrictions (`allowChangeEndpoint`)

## Components Modified
- `frontend/src/components/agency/agency/Agency.jsx`
- `frontend/src/components/agency/ds-settings-card/DsSettingsCard.jsx`
- `frontend/src/components/agency/workflow-card/WorkflowCard.jsx`

## Testing Checklist
- [ ] Workflow owner can modify all settings
- [ ] Non-owner cannot change Prompt Studio project
- [ ] Non-owner cannot configure tool settings
- [ ] Non-owner cannot modify source connector
- [ ] Non-owner cannot modify destination connector
- [ ] Tooltips display correct messages for non-owners
- [ ] Tool names display correctly even when tool is no longer exported
- [ ] Ownership check handles type differences correctly


johnyrahul:
> @johnyrahul just to be on the safer side, do we have checks in place to prevent this from the backend?

@chandrasekharan-zipstack  Not yet. When I checked, since it is a workflow level edits, it should ideally be available under the workflow view set, but instead, a few of the settings are under the tool_instance view set and models. Let me give it a try at how to restrict such actions based on workflow ownership

PR #1594, opened 2025-10-15
UN-2893 [FIX] Fix duplicate process handling status updates and UI error logs

muhammad-ali-e:
## What

- Fixed duplicate worker process incorrectly updating file execution status and showing UI error logs
- Added `is_duplicate_skip` flag to `FileProcessingResult` to track duplicate detection
- Skip status updates and UI error logs for duplicate processes (internal race condition handling)
- Fixed `destination_processed` default value in processor.py to enable correct duplicate detection

## Why

- When two GKE worker pods process the same file simultaneously, Redis lock detects the duplicate
- Previously, the duplicate (second) worker still updated status to ERROR and showed UI error logs
- This caused:
  - Race condition status pollution (second worker overwrites first worker's status)
  - Confusing UI error messages for users
  - DB inconsistencies when both workers try to update the same file execution record

## How

**1. Added duplicate detection tracking** (`worker_models.py`):
- Added `is_duplicate_skip` field to `FileProcessingResult` dataclass

**2. Propagated duplicate flag** (`destination_connector.py`):
- Return `None` when duplicate detected (Redis lock already held)
- This signals upstream that duplicate was detected

**3. Fixed duplicate detection logic** (`processor.py`):
- Line 387: Changed default from `else True` to `else False` for correct duplicate detection
- Line 396: Calculate `is_duplicate = not destination_processed and not destination_error`
- Lines 245-285: Skip status updates and UI logs when `is_duplicate_skip=True`
- Only log internal debug message: "DUPLICATE SKIP: File 'X' identified as duplicate"

**4. Enhanced result tracking** (`service.py`, `tasks.py`):
- Properly propagate `processed` flag through `FinalOutputResult`
- Extract `is_duplicate_skip` flag in tasks.py for downstream handling

**Result**: Second worker silently exits without polluting status or UI. First worker completes normally.

**No**, this PR cannot break existing features because:

1. **Backward compatible**: Only adds optional `is_duplicate_skip` field to dataclass (defaults to `False`)
2. **Defensive change**: processor.py default value change (`True` → `False`) is safer for edge cases
3. **Conditional logic**: Skip logic only activates when `is_duplicate_skip=True` (only for duplicates)
4. **Normal flow unchanged**: Non-duplicate files follow the same code path as before
5. **Improves reliability**: Prevents race condition status pollution that was causing issues

## Database Migrations

- None required

## Env Config

- No environment configuration changes

## Relevant Docs

- Duplicate handling uses Redis `SET NX` lock (documented in destination_connector.py)
- GKE worker scaling can cause multiple pods to pick same file from queue

## Related Issues or PRs

- Fixes user report: Oct 15, 2025 9:45 PM IST - "File 'invoice-nov.pdf' destination processing failed"
- Related to GKE horizontal pod autoscaling race conditions

## Dependencies Versions

- No dependency changes

## Notes on Testing

**Test scenario**: Two workers processing same file simultaneously
1. Worker 1 acquires Redis lock, processes file → Status: COMPLETED ✅
2. Worker 2 detects duplicate (lock held) → Silently exits, no status update ✅
3. UI shows only Worker 1's success log ✅
4. Debug logs show: "DUPLICATE SKIP: File 'X' identified as duplicate" ✅

**Verified**:
- No UI error logs for duplicates
- No status pollution (second worker doesn't overwrite first worker's status)
- First worker completes successfully
- Second worker logs internal debug message only

## Screenshots

N/A - Fix is for internal race condition handling

## Checklist

- [x] Read and understood the [Contribution Guidelines](https://docs.unstract.com/unstract/contributing/unstract/)
- [x] Code follows architecture principles (dataclasses, enums, backward compatibility)
- [x] Tested duplicate file scenario with multiple workers
- [x] Verified no status updates from duplicate process
- [x] Verified no UI error logs from duplicate process
- [x] No breaking changes to existing functionality

### DIFF
--- a/unstract/connectors/src/unstract/connectors/databases/bigquery/bigquery.py
+++ b/unstract/connectors/src/unstract/connectors/databases/bigquery/bigquery.py
@@ -63,6 +63,51 @@ class BigQuery(UnstractDB):
     def can_read() -> bool:
         return True
 
+    @staticmethod
+    def _sanitize_for_bigquery(data: Any) -> Any:
+        """BigQuery-specific float sanitization for PARSE_JSON compatibility.
+
+        BigQuery's PARSE_JSON() requires floats that can "round-trip" through
+        string representation. This method limits total significant figures to 15
+        (IEEE 754 double precision safe zone) to ensure clean binary representation.
+
+        Args:
+            data: The data structure to sanitize (dict, list, or primitive)
+
+        Returns:
+            Sanitized data compatible with BigQuery's PARSE_JSON
+
+        Example:
+            >>> BigQuery._sanitize_for_bigquery({"time": 1760509016.282637})
+            {'time': 1760509016.28264}  # Limited to 15 significant figures
+
+            >>> BigQuery._sanitize_for_bigquery({"cost": 0.001228})
+            {'cost': 0.001228}  # Unchanged (only 4 significant figures)
+        """
+        import math
+
+        if isinstance(data, float):
+            # Handle special values that BigQuery can't store in JSON
+            if math.isnan(data) or math.isinf(data):
+                return None
+            if data == 0:
+                return 0.0
+
+            # Limit total significant figures to 15 for IEEE 754 compatibility
+            # BigQuery PARSE_JSON requires values that round-trip cleanly
+            # For large numbers (like Unix timestamps), this reduces decimal precision
+            # For small numbers (like costs), full precision is preserved
+            magnitude = math.floor(math.log10(abs(data))) + 1
+            safe_decimals = max(0, 15 - magnitude)
+            return float(f"{data:.{safe_decimals}f}")
+
+        elif isinstance(data, dict):
+            return {k: BigQuery._sanitize_for_bigquery(v) for k, v in data.items()}
+        elif isinstance(data, list):
+            return [BigQuery._sanitize_for_bigquery(item) for item in data]
+        else:
+            return data
+
     def get_engine(self) -> Any:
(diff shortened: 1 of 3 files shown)


workers/shared/processing/files/processor.py
```diff
@@ -597,37 +607,49 @@ def process_file(
...
+                                status=ExecutionStatus.ERROR.value,
+                                error_message=error_msg,
+                            )
+                            logger.info(
+                                f"Updated file execution {context.workflow_file_execution_id} status to ERROR"
+                            )
+                        except Exception as status_error:
+                            logger.error(
+                                f"Failed to update file execution status: {status_error}"
+                            )
```
chandrasekharan-zipstack:
```suggestion
                        except Exception:
                            logger.exception(
                                f"Failed to update file execution status: {status_error}"
                            )
```

muhammad-ali-e:
> Why are workers picking up tasks of the same file? Is it a behaviour of rabbitmq / celery?

Yes.. While warm shutdown

workers/shared/processing/files/processor.py
muhammad-ali-e: addressed this with another commit since your suggestion will raise Undefined name `status_error`.. There have some logs  taht need to be converted to warning or info with proper exception . That will address later.

muhammad-ali-e:
> @muhammad-ali-e does the duplicate run also use the same file_execution_id? I assume operations are idempotent elsewhere

same file_execution_id

PR #1599, opened 2025-10-16
UN-2897 [FIX] Prevent SIGSEGV crashes from Secret Manager gRPC calls in Google Drive connector

muhammad-ali-e:
## What

- Added module-level cache for Google Drive client secrets to prevent repeated Secret Manager API calls
- Prevents SIGSEGV crashes caused by gRPC-based Secret Manager client in forked worker processes

## Why

- PID 164 crashed in `__init__()` before lazy initialization could help (discovered after initial fix deployment)
- `GCSHelper().get_secret()` uses Google Cloud Secret Manager which is gRPC-based and not fork-safe
- When called before fork in parent process, child processes inherit corrupted gRPC state causing SIGSEGV (signal 11)
- This is a second root cause discovered through production log analysis of crashes that occurred even after initial lazy init fix

## How

- Added module-level `_client_secrets_cache` and `_client_secrets_lock` 
- Modified `__init__()` to use double-check locking pattern for thread-safe caching
- Secret Manager is called only once per process on first connector instantiation
- Subsequent connector instances within same process reuse cached secrets
- Each worker process after fork gets its own cache (no cross-process sharing)

- **No breaking changes**
- Backwards compatible - only changes internal caching behavior
- Secret Manager still called when cache is empty (normal cold start)
- Same credentials returned to connectors
- **Performance improvement**: Reduces Secret Manager API calls from N (per connector instance) to 1 (per process)

## Database Migrations

- None required

## Env Config

- None required

## Relevant Docs

- N/A

## Related Issues or PRs

- This builds on PR #1597 (initial lazy initialization fix for PyDrive2 client)
- Addresses second root cause discovered in production after initial fix deployment

## Dependencies Versions

- No dependency changes

## Notes on Testing

- Monitor staging logs for "Loading Google Drive client secrets from Secret Manager (one-time per process)"
- Log should appear only once per worker process, not per connector instance
- Verify no SIGSEGV crashes in worker PIDs using Google Drive connector
- Test with multiple concurrent file processing tasks using Google Drive as source

## Checklist

### DIFF
--- a/workers/general/tasks.py
+++ b/workers/general/tasks.py
@@ -442,7 +442,7 @@ def _process_file_batches_general(
         )
 
         logger.info(
-            f"Created file batch {batch_response.get('batch_id')} with {batch_response.get('total_files')} files"
+            f"Created file batch {batch_response.batch_id} with {batch_response.total_files} files"
         )
 
         return [batch_response]


vishnuszipstack (approved):

johnyrahul:
@muhammad-ali-e, is this something that is reproducible only under load?

ritwik-g (approved):

Issue #1611, opened 2025-10-23
fix: docker compose should specify project name


michaelcizmar:
## Describe the bug
Without a specified compose project name docker compose uses the folder name which in this case is 'docker' and that can lead to compose collision.

## To reproduce
Run the compose file.

chandrasekharan-zipstack:
Hey @michaelcizmar 
Thanks for pointing this out and raising a PR to fix it. Will close this issue once the PR is merged and released

chandrasekharan-zipstack:
Thanks for the fix @michaelcizmar, it has been released as part of [v0.139.1](https://github.com/Zipstack/unstract/releases/tag/v0.139.1)

PR #1612, opened 2025-10-23
[MISC] Specify project name to docker compose

michaelcizmar:
## What

Adding a property to the docker compose file to set the compose project name which will override the default behavior which is to us the folder name.

## Why

The folder which the docker compose is in is called 'docker'.  This is a common name and can cause collisions in other docker compose projects that use this name.

## How

Added a property to the docker compose file.

Changing the name of the compose project an have adverse effects to existing compose projects.

## Env Config

Adds an env variable to trigger

## Related Issues or PRs

- https://github.com/Zipstack/unstract/issues/1611


docker/docker-compose.yaml
```diff
@@ -1,3 +1,4 @@
+name: ${COMPOSE_PROJECT_NAME:-unstract}
```
ritwik-g: @chandrasekharan-zipstack @michaelcizmar onething I am worried about is if adding this could cause any issue for existing users. An autogenerated name would have already been used and adding a new one might cause issue. If it can be confirmed if this is a concern or not then we can approve this for main.

cc: @coderabbitai

michaelcizmar: The issue that I experienced was, as you pointed out, that docker makes a compose project with the directory name  which is "docker".  I had an existing project with docker and have this issue with that other project.

I think your change makes sense.

chandrasekharan-zipstack:
```suggestion
name: ${COMPOSE_PROJECT_NAME:-docker}
```

michaelcizmar:
> @michaelcizmar the PR looks fine now. However it can be merged only after the below concerns are addressed
> 
> 1. Please [sign the CLA](https://github.com/Zipstack/unstract/pull/1612#issuecomment-3438623505)
> 2. Update the PR description to highlight what the change is and why we're doing it. Feel free to keep it concise and remove any fields that don't apply

1 - Done
2 - Done

michaelcizmar:
Anything remaining to do here?

gaya3-zipstack (approved):
Looks good.

PR #1605, opened 2025-10-22
UN-2175 Avoid extracting when doc is already extracted during indexing

chandrasekharan-zipstack:
## What

- Removed `reindex` parameter from `dynamic_extractor()`
- Removed a check involving the reindex parameter

## Why

- Once a document is extracted and the text is available, we needn't re-extract it

## How

- Removed redundant check of `reindex` during text extraction

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

- No, the reindex parameter was only used to determine if a document should be extracted and was always enforced. Now we check and extract only if it required

## Related Issues or PRs

- Noticed while working with [UN-2175](https://zipstack.atlassian.net/browse/UN-2175)

## Notes on Testing

- Tested via the explicit index and it worked as expected

## Checklist

I have read and understood the [Contribution Guidelines](https://docs.unstract.com/unstract/contributing/unstract/).

[UN-2175]: https://zipstack.atlassian.net/browse/UN-2175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1639, opened 2025-11-04
Revert "UN-2175 Avoid extracting when doc is already extracted during indexing"

chandrasekharan-zipstack:
Reverts Zipstack/unstract#1605

# Why

- Current solution made for #1605 causes a regression when highlighting is enabled or disabled
- The long term solution needs to keep track of this information during extraction which can in turn require some form of a migration -> either lazily or a dedicated data migration
- Hence reverting these changes in order to incorporate a better solution with no regressions

### DIFF
--- a/backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
+++ b/backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
@@ -427,6 +427,7 @@ class PromptStudioHelper:
             run_id=run_id,
             enable_highlight=tool.enable_highlight,
             doc_id=doc_id,
+            reindex=True,
         )
         if tool.summarize_context:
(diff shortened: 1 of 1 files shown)


backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
muhammad-ali-e, line 1327: Just to confirm, is this function used only on line 222?

muhammad-ali-e:
@chandrasekharan-zipstack is this PR meant to revert a commit.. instaed can we do  git revert?

chandrasekharan-zipstack:
@muhammad-ali-e I raised this PR through github's UI. Anyway I have raised a PR #1643 to fix the regression caused by #1605
Hence, closing this PR for now

PR #1643, opened 2025-11-04
UN-2947 [MISC] Ensure re-index works on highlight support toggle

chandrasekharan-zipstack:
## What

- Fix regression in PromptStudio's highlight extraction tracking where the system incorrectly reuses non-highlighted extracted text when highlighting is enabled
- Enhance extraction_status tracking to include highlight metadata
- Prevent unnecessary re-extractions when toggling highlight on/off

## Why

- PR #1605 removed the reindex parameter without tracking highlight metadata
- extraction_status was stored as boolean, losing context about whether highlighting was enabled
- When users toggled highlighting on after initial extraction, system reused old non-highlighted text instead of re-extracting
- This caused incorrect extraction results for users enabling highlights later

## How

- Enhanced `check_extraction_status()` to verify highlight metadata matches current state
- Enhanced `mark_extraction_status()` to store highlight metadata (dict format) alongside status
- Updated `dynamic_extractor()` calls to pass enable_highlight parameter
- Fully backward compatible: old boolean extraction_status records treated as non-highlighted extractions
- No database migration needed: JSONField handles both boolean and dict formats

No breaking changes. The enhancement is fully backward compatible:
- Existing boolean extraction_status values are treated as non-highlighted extractions
- New dict format includes both status and highlight metadata
- All existing code paths continue to work without modification

## Database Migrations

- None required. JSONField in extraction_status is flexible and handles both boolean and dict formats.

## Env Config

- None

## Relevant Docs

- PromptStudio highlight documentation should note that highlights require re-extraction of text

## Related Issues or PRs

- Fixes regression introduced in PR #1605
- Note: Related frontend rendering issue with triple-nested arrays for number-type prompts (requires separate fix/PR)

## Dependencies Versions

- None

## Notes on Testing

- Toggle highlight setting on/off to verify re-extraction is triggered
- Verify existing extractions with highlighting enabled continue to work
- Test backward compatibility with prompts that have boolean extraction_status values
- Confirm no extra re-extractions occur when highlight state hasn't changed

## Screenshots

N/A (backend extraction tracking fix)

## Checklist

### DIFF
--- a/backend/prompt_studio/prompt_studio_index_manager_v2/models.py
+++ b/backend/prompt_studio/prompt_studio_index_manager_v2/models.py
@@ -69,6 +69,8 @@ class IndexManager(BaseModel):
         null=False,
         blank=False,
     )
+    # TODO: Consider making this an enum instead
+    # A nested JSON of the form {doc_id: {"extracted": True, "enable_highlight": <bool>}}
     extraction_status = models.JSONField(
         db_comment="Extraction status for documents",
         null=False,


backend/prompt_studio/prompt_studio_index_manager_v2/prompt_studio_index_helper.py
```diff
@@ -148,20 +162,46 @@ def check_extraction_status(
...
-                    f"Extraction is already marked as complete "
+                    f"Extraction is NOT yet marked as complete "
                     f"for document: {document_id} with {doc_id}"
                 )
+                return False
+
+            # Backward compatibility: treat boolean True as non-highlighted
+            if isinstance(status_entry, bool):
+                is_extracted = status_entry
+                is_highlight_handled = False
```
muhammad-ali-e, line 177: Once the user enables the highlight option in the tool, we re-extract the existing document regardless of whether highlighting was enabled before or not. right? That's the impact here.. right?

backend/prompt_studio/prompt_studio_index_manager_v2/prompt_studio_index_helper.py
```diff
@@ -148,20 +162,46 @@ def check_extraction_status(
...
+            else:
+                # New format: {"extracted": True, "enable_highlight": <bool>}
```
muhammad-ali-e, line 179: Where do we store or save the new format in IndexManager? Can we make it an enum or a validator? It would give a better idea of what we store in status_entry and be helpful when making changes to the dictionary.

muhammad-ali-e (approved):

backend/prompt_studio/prompt_studio_index_manager_v2/prompt_studio_index_helper.py
chandrasekharan-zipstack: Yes, but its a one-time cost and after that we ensure that we don't re-extract unnecessarily if it was already done before and available

backend/prompt_studio/prompt_studio_index_manager_v2/prompt_studio_index_helper.py
chandrasekharan-zipstack: In this method we call its `.save()`. Will update the `models.py` to highlight this JSON's structure. I didn't want to mess with this and create a schema / data migration for the time being

PR #1644, opened 2025-11-04
UN-2952 [MISC] Separate storage for highlighted vs non-highlighted extracted text

chandrasekharan-zipstack:
## What

- Created `ExtractionPathHelper` utility class with explicit suffix system (`_plain` and `_highlighted`)
- Updated path generation in `dynamic_extractor`, `dynamic_indexer`, `_fetch_response`, and `_fetch_single_pass_response` to use the helper
- Updated Raw View API (`fetch_contents_ide`) to route to correct file based on highlight setting
- Removed unnecessary fallback logic for cleaner architecture

## Why

Toggling the highlight setting in PromptStudio IDE was causing repeated extractions from the text extractor. This PR implements separate storage for highlighted vs non-highlighted extracted text, allowing the system to use cached versions when toggling without re-extraction.

## How

Storage pattern uses explicit suffixes:
- Non-highlighted: `filename_plain.txt`
- Highlighted: `filename_highlighted.txt`

This approach:
- Eliminates ambiguity between old (no suffix) and new formats
- Prevents re-extraction when toggling highlight
- Maintains DB tracking for performance
- Forces re-extraction for old files without suffix (acceptable for IDE scope)

This PR is backward incompatible by design. Files extracted without the suffix system will trigger re-extraction when users toggle highlight in PromptStudio IDE. This is intentional and acceptable since:
- Changes are scoped to PromptStudio IDE only
- Deployed workflows are out of scope
- Clean break approach avoids fallback complexity

## Database Migrations

- None required. Clean break approach.

## Relevant Docs

- Related to PromptStudio IDE extraction caching
- Works with UN-2947 (extraction status tracking with highlight metadata)

## Related Issues or PRs

- UN-2947: Fix highlight extraction tracking regression (base branch)
- Combined with UN-2947, provides complete solution for highlight extraction caching
- https://zipstack.atlassian.net/browse/UN-2952

## Notes on Testing

Test scenarios:
- Toggle highlight setting on extracted document (verify no re-extraction)
- Raw View API displays correct file content based on highlight state
- File deletion properly handles both _plain and _highlighted suffixes
- Old files without suffix trigger re-extraction when toggled

# Possible concern
1. One copy of the old extracted file might be present for existing projects, I hope that's okay for now. Chose the simpler approach to avoid any sort of migration
image

---

### DIFF
--- a/backend/prompt_studio/prompt_studio_index_manager_v2/models.py
+++ b/backend/prompt_studio/prompt_studio_index_manager_v2/models.py
@@ -69,6 +69,8 @@ class IndexManager(BaseModel):
         null=False,
         blank=False,
     )
+    # TODO: Consider making this an enum instead
+    # A nested JSON of the form {doc_id: {"extracted": True, "enable_highlight": <bool>}}
     extraction_status = models.JSONField(
         db_comment="Extraction status for documents",
         null=False,


jagadeeswaran-zipstack (approved):

backend/prompt_studio/prompt_studio_core_v2/extraction_path_helper.py
```diff
@@ -0,0 +1,52 @@
...
+
+    Storage Pattern:
+        - Non-highlighted: extract/filename_plain.txt
+        - Highlighted: extract/filename_highlighted.txt
+
+    Note: Old files without suffix (extract/filename.txt) are NOT used and will
+    trigger re-extraction. This ensures no ambiguity between old and new formats.
+    """
+
+    PLAIN_SUFFIX = "_plain"
```
Deepak-Kesavan, line 21: Any reason why are we suffixing _plain than leaving it in old format? Because this might cost everyone 1 additional extraction. Also how do we handle cleanup of old files if not used?

chandrasekharan-zipstack: @Deepak-Kesavan 
1. Existing user's files might have highlight data or not and we can only confirm by checking the file for it. In order to avoid this lazy migration effort altogether, I felt that its okay to perform this extra extraction
    - Also until now we were always forcing extraction in cases such as `reindex` so I felt this was okay
2. Regarding cleanup, currently it happens only while deleting the file from prompt studio and that would [remove all these directories](https://github.com/Zipstack/unstract/blob/bbe55e80bcfe6bccd8b7acc52d4216bdbbeaf43a/backend/utils/file_storage/helpers/prompt_studio_file_helper.py#L187-L188) itself which should take care of file's cleanup

chandrasekharan-zipstack:
Closing this PR since duplicating the data to correct this is not desirable and instead we should aim to concatenate or stitch the required data

PR #1660, opened 2025-11-17
[MISC] - Migrate unstract-flags to flipt-client SDK

pk-zipstack:
## What

- Migrate unstract-flags to flipt-client SDK

## Why

- 

## How

- Removed: grpcio-tools, grpcio, protobuf (gRPC-based dependencies)
- Added: flipt-client>=0.1.0 (official Flipt SDK)
- Replaced: EvaluationClient → FliptClient from flipt-client SDK
- Removed: namespace_key parameter (no longer needed with flipt-client)

- No, this doesn't break any existing feature because this is a change with the feature flags which replaces the current implementation with EvaluationClient.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/platform-service/uv.lock
+++ b/platform-service/uv.lock
@@ -654,6 +654,19 @@ wheels = [
 ]
 
 [[package]]
+name = "flipt-client"
+version = "1.3.2"
+source = { registry = "https://pypi.org/simple" }
+dependencies = [
+    { name = "packaging" },
+    { name = "pydantic" },
+]
+sdist = { url = "https://files.pythonhosted.org/packages/8b/b2/ea0cfd90eb83059a6219f6865e2a9c99d65f0fd51718f33d8840f4df2b52/flipt_client-1.3.2.tar.gz", hash = "sha256:4b5128db4886e50a66586efbf64c0fcb54d076e6be641986096a01325336aba4", size = 23624252, upload-time = "2025-10-31T18:36:00.594Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/27/59/ea75a9af11748d7e05c6a9ce30c0b868410af94a5ea0636fd299ca20f9b6/flipt_client-1.3.2-py3-none-any.whl", hash = "sha256:b32f8fef09d43df6eefbb84d7b718e5510097866c6c889a8934de18a90e71390", size = 23760883, upload-time = "2025-10-31T18:35:57.995Z" },
+]
+
+[[package]]
 name = "frozenlist"
 version = "1.7.0"
 source = { registry = "https://pypi.org/simple" }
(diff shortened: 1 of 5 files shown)


prompt-service/src/unstract/prompt_service/services/indexing.py
```diff
@@ -88,7 +92,9 @@ def index(
             )
 
             # Index and return doc_id
-            logger.info("performing Index -------------------------------------------------")
+            logger.info(
+                "performing Index -------------------------------------------------"
+            )
```
ritwik-g: @pk-zipstack these kind of logs are not very helpful other than for debugging. I would suggest making these debug logs instead of info. You can enable the debug logs in integration while testing if needed. Else please make sure such logs are removed before merging to main

unstract/flags/src/unstract/flags/client/flipt.py
ritwik-g, line 1: @pk-zipstack I think we can remove the below folders itself.

- client
- generate
- proto
- tests

All these folders are part of old GRPC setup. If the new one works let's remove the old ones.

unstract/flags/src/unstract/flags/feature_flag.py
```diff
@@ -3,24 +3,20 @@
...
 
-from .client.evaluation import EvaluationClient
+from flipt_client import FliptClient
 
 logger = logging.getLogger(__name__)
 
 
 def check_feature_flag_status(
     flag_key: str,
-    namespace_key: str | None = None,
```
ritwik-g, line 13: @pk-zipstack can you confirm if this namespace_key is not passed from any places where this function is used.

unstract/flags/src/unstract/flags/feature_flag.py
```diff
@@ -31,23 +27,25 @@ def check_feature_flag_status(
...
-        response = evaluation_client.boolean_evaluate_feature_flag(
-            namespace_key=namespace_key,
+        # Get Flipt server URL from environment
+        flipt_url = os.environ.get("FLIPT_URL", "http://localhost:8080")
+
+        # Initialize Flipt client
+        client = FliptClient(url=flipt_url)
+
+        # Evaluate boolean flag
+        result = client.evaluate_boolean(
```
ritwik-g, line 52: 1. We still need to use UNSTRACT_FEATURE_FLAG_NAMESPACE to get namespace. That way only the appropriate namespace will work.
2. The FliptClient usage is wrong. Check this [doc](https://github.com/flipt-io/flipt-client-sdks/blob/main/flipt-client-python/README.md#usage). We need to use client options to pass the flipt_url. Together with that we need also pass the namespace. We use different namespace for staging and integration in flipt. And we configure same in UNSTRACT_FEATURE_FLAG_NAMESPACE

prompt-service/src/unstract/prompt_service/services/indexing.py
praveen-formido: @ritwik-g my bad these were meant to track another issue with indexing locally. Must have pushed it by mistake. Will remove them.

unstract/flags/src/unstract/flags/feature_flag.py
praveen-formido: Understood @ritwik-g, I thought we don't need namespaces and just need keys to toggle between different envs.

unstract/flags/src/unstract/flags/feature_flag.py
praveen-formido: No, this namespace is not passed anywhere else.

PR #1663, opened 2025-11-19
UN-2261 [FIX] Disabled broken Oracle connector from UI

kirtimanmishrazipstack:
## What

- Removing oracle support in frontend temporarily

## Why

- Oracle is broken with wallet support not provided. Oracle support is feature yet to be picked

## How

- Simple frontend changes

- N/A

## Database Migrations

- N/A

## Env Config

- N/A

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist


jagadeeswaran-zipstack (approved):

frontend/src/components/connectors/connector-list-modal/ConnectorListModal.jsx
```diff
@@ -24,6 +24,11 @@ function ConnectorListModal({
   const filterConnectors = debounce((searchText, mode) => {
     let filtered = [...connectors];
 
+    // Filter out Oracle connector temporarily
```
chandrasekharan-zipstack: @kirtimanmishrazipstack I suggest mentioning why we are doing this in a comment. Also if this can be done in the backend directly, it'll be even better. This is already done [here](https://github.com/Zipstack/unstract/blob/38e5d372a202ac97e434bd07a85a23ee66ea5722/backend/connector_processor/connector_processor.py#L80-L84), check and add your connector to this list

.gitignore
```diff
@@ -632,6 +632,8 @@ workers/pluggable_worker/*
 frontend/src/plugins/*
 frontend/public/llm-whisperer/
 
+# Prompt-service
+prompt-service/plugins/*
```
chandrasekharan-zipstack: @kirtimanmishrazipstack this might not be needed since we already have a gitignore entry within the `prompt-service` directory. If possible, start a thread and let's decide on whether we should maintain a single `gitignore` that manages entries for all services or if its okay to have entries on a per-service level

backend/connector_processor/connector_processor.py
```diff
@@ -21,14 +21,28 @@
...
+    except ImportError as e:
+        logger.warning(f"Failed to import {class_name}: {e}")
+        return None
+
+
+# Import optional connectors
+RedisQueue = import_optional_connector("unstract.connectors.queues.redis", "RedisQueue")
+OracleDB = import_optional_connector(
+    "unstract.connectors.databases.oracle_db", "OracleDB"
+)
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: @kirtimanmishrazipstack add a suitable comment explaining why we have removed it

.gitignore
kirtimanmishrazipstack: @chandrasekharan-zipstack Its being seen Untracked files in `unstract` repo. I suggest keeping it.

backend/connector_processor/connector_processor.py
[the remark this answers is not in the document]
kirtimanmishrazipstack: @chandrasekharan-zipstack added a TODO

.gitignore
ritwik-g: Adding in `prompt-service` directory would be better. As it is related to that service

ritwik-g: https://github.com/Zipstack/unstract/blob/main/prompt-service/.gitignore#L2

The problem is we don't have `plugins/*` added in prompt-service

ritwik-g: @kirtimanmishrazipstack @chandrasekharan-zipstack I think the path used here for plugin is wrong. It should be moved to the path in gitignore file.

backend/connector_processor/connector_processor.py
```diff
@@ -21,14 +21,29 @@
...
 
-try:
-    from unstract.connectors.queues.redis import RedisQueue
-except ImportError:
-    RedisQueue = None
-
 logger = logging.getLogger(__name__)
 
 
+def import_optional_connector(module_path: str, class_name: str):
```
chandrasekharan-zipstack, line 27: @kirtimanmishrazipstack any specific reason why we are doing this? The logs present here can turn out to add little noise when we start the `backend`. Consider making it debug instead or clearly mention the impact of it to make it more meaningful.

chandrasekharan-zipstack (approved):

backend/connector_processor/connector_processor.py
kirtimanmishrazipstack: @chandrasekharan-zipstack earlier we imported a optional connectors `redisQueue`, now `oracleDB`. This is to make code cleaner nothing else.

chandrasekharan-zipstack: I get that @kirtimanmishrazipstack however I'm not a fan of the logs that are involved here.
1. In case the import is successful it logs 
    `Successfully imported RedisQueue`
We don't seem to do this for other modules / conditional imports and it becomes rather noisy if we choose to do so. We don't clearly understand how / why this might help when looking at the logs on the whole. Hence I would suggest marking it DEBUG or removing it altogether
2. In the case the import fails (OSS scenario), people would notice a noisy warning `Failed to import..` and that indicates something might be wrong when it isn't.

Ultimately I'd suggest either of the options
- keep things simple as it was and avoid this additional method
- or mark the logs as DEBUG
- or retain the same log levels but make it more meaningful and unambiguous

kirtimanmishrazipstack: @chandrasekharan-zipstack added as debug log

PR #1664, opened 2025-11-19
[LW-267] FIX: Propagate 429 errors from Embedding providers

chandrasekharan-zipstack:
## What

- Propagate embedding provider errors properly

## Why

- Previously embedding provider's errors were suppressed / not reported correctly

## How
- Update error handling code to ensure this gets propagated correctly
- Added `parse_litellm_err()` to generically parse the error and explicitly mention that its from a 3rd party service
- Added `strip_litellm_prefix()` utility function to remove litellm prefixes from error messages

No. This change only affects the format of error messages returned to users. The underlying error handling logic and error propagation remain unchanged. Status codes and error types are preserved correctly.

## Relevant Docs

- https://docs.litellm.ai/docs/exception_mapping#litellm-exceptions

## Notes on Testing

- Tested in prompt studio by using an embedding provider with 1K rate limit and passed a sufficiently large file
- Tested in workflow execution by passing a file with API source and destination

## Screenshots
- image
- image


pk-zipstack (approved):

ritwik-g (approved):
@chandrasekharan-zipstack let's add a log in the parse method before the cleaned one is created. Just in case some issue happens in that cleaning phase this log will help us.

PR #1673, opened 2025-11-21
[FIX] Added aws as optional dependency for tools

pk-zipstack:
## What

- Added aws as optional dependency for tools

## Why

- Without these optional dependencies the file storage was failing

## How

- Added optional dependency aws for the tools in their respective requirements.txt files.

- No, this cannot break any change siince this is a dependency change.

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/tools/classifier/requirements.txt
+++ b/tools/classifier/requirements.txt
@@ -4,5 +4,5 @@
 # aws alone is needed here
 # because tools use transient temporary storage.
 
--e file:/unstract/sdk1
+-e file:/unstract/sdk1[aws]
 -e file:/unstract/flags
--- a/tools/structure/requirements.txt
+++ b/tools/structure/requirements.txt
@@ -3,6 +3,6 @@
 # Required for all unstract tools
 # aws alone is needed here
 # because tools use transient temporary storage.
--e file:/unstract/sdk1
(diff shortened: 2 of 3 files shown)


tools/classifier/requirements.txt
```diff
@@ -4,5 +4,5 @@
 # aws alone is needed here
 # because tools use transient temporary storage.
 
--e file:/unstract/sdk1
+-e file:/unstract/sdk1[aws]
```
gaya3-vijayakumar, line 7: @pk-zipstack  - Are we still using text_extractor and classifier? Pls check with @chandrasekharan-zipstack

chandrasekharan-zipstack: @gaya3-zipstack we still have these in the platform so we are forced to maintain changes here. As for its usefulness - I don't think it has any active usecases and we should plan to deprecate and remove them if product agrees as well. @pk-zipstack do create a ticket to track this if that's the case
cc: @hari-kuriakose

chandrasekharan-zipstack (approved):

PR #1703, opened 2025-12-12
UN-3022  [FEAT] - Enable HITL rule engine for API deployments

vishnuszipstack:
## What
 - Enable HITL (Human-In-The-Loop) rule configuration for API deployment workflows, allowing users to define conditional rules that determine which documents get routed to manual review.

## Why

  Previously, API deployments with `hitl_queue_name` would send ALL processed documents to the HITL queue unconditionally. Users needed the ability to:
  - Define rules based on extracted field values (e.g., `name CONTAINS "jose"`)
  - Route only specific documents to manual review based on business logic
  - Get visibility into why a document was routed to HITL via API response metadata

### Frontend
  - Enabled "Configure HITL" button for API destination connectors
  - RuleEngine component renders for API mode with `rule_type="API"` to distinguish from DB rules

  ### Backend
  - Added `has_api_rules()` and `validate_rule_engine()` methods to `WorkflowUtil`
  - Updated `_should_handle_hitl()` to evaluate API rules when present:
    - Rules exist → evaluate and route based on match
    - No rules → all files to HITL (backward compatible)

  ### Workers
  - Added `HITLDecision` dataclass to capture routing decision with reason
  - Added HITL metadata to API response: `file_sent_to_hitl` (bool) and `reason` (string)
  - Rule evaluation happens after document processing when output is available

  ### OSS Compatibility
  - `WorkflowUtil` placeholder methods return `False` ensuring no HITL routing in OSS
  - HITL metadata only included when manual review plugin is available

- No

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

### DIFF
--- a/.github/workflows/ci-container-build.yaml
+++ b/.github/workflows/ci-container-build.yaml
@@ -18,6 +18,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
       - name: Login to Docker Hub
+        if: github.event.pull_request == null || github.event.pull_request.head.repo.fork == false
         uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
--- a/.github/workflows/ci-test.yaml
+++ b/.github/workflows/ci-test.yaml
@@ -5,16 +5,16 @@ on:
     branches:
       - main
     paths-ignore:
-      - 'docker/**'
-      - 'frontend/**'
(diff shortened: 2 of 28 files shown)


chandrasekharan-zipstack:
Check the sonarcloud failure

PR #1705, opened 2025-12-16
UN-3088 [MISC] Handle pg_trgm extension in non-public schemas for multi-schema PostgreSQL deployments

chandrasekharan-zipstack:
Refer [#1706 ](https://github.com/Zipstack/unstract/pull/1706)

## What

- Fixed Django migration failure for pg_trgm extension when it exists in a different schema than the app schema
- Replaced Django's built-in TrigramExtension() with a schema-aware implementation
- Enhanced database connection logging to show search_path configuration

## Why

- Migration 0018_filehistory_execution_count_and_more.py was failing in on-prem staging deployment with error: "operator class gin_trgm_ops does not exist"
- Root cause: Django's TrigramExtension() creates the extension in the current schema, but our custom DB wrapper sets search_path to only the app schema, making extension operators invisible
- This prevented deployments from running successfully in environments where pg_trgm was already installed in a different schema (e.g., public schema)

## How

**backend/custom_db/base.py**:
- Added search_path information to database connection logging
- Helps diagnose schema visibility issues during development and deployment

**backend/workflow_manager/workflow_v2/migrations/0018_filehistory_execution_count_and_more.py**:
- Implemented `ensure_pg_trgm_extension()` function that:
  1. Queries pg_extension system catalog to check if pg_trgm exists
  2. Detects which schema contains the extension
  3. Creates the extension in public schema if not found
  4. Adds the extension's schema to search_path so gin_trgm_ops operator class is visible
- Replaced Django's `TrigramExtension()` operation with `RunPython(ensure_pg_trgm_extension)`

No, this PR cannot break existing features because:

1. **Backward compatible**: The new implementation handles all scenarios that the old TrigramExtension() handled, plus additional edge cases
2. **Idempotent**: The function checks if pg_trgm exists before creating it, making it safe to run multiple times
3. **Schema-aware**: Works correctly whether pg_trgm is in public schema, app schema, or any other schema
4. **Non-destructive**: Only adds to search_path, doesn't modify existing schemas or data
5. **Migration safety**: Uses RunPython with proper error handling and rollback support
6. **Logging enhancement**: The DB connection logging change only adds information, doesn't modify behavior

## Database Migrations

- Modified migration: `backend/workflow_manager/workflow_v2/migrations/0018_filehistory_execution_count_and_more.py`
- Migration is idempotent and safe to re-run
- No data changes, only schema/extension configuration

## Env Config

- No environment variable changes required
- No configuration updates needed

## Relevant Docs

- JIRA: UN-3088
- Related to PostgreSQL pg_trgm extension documentation: https://www.postgresql.org/docs/current/pgtrgm.html

## Related Issues or PRs

- Fixes on-prem staging deployment issues
- Related to JIRA UN-3088

## Dependencies Versions

- No dependency version changes

## Notes on Testing

- Tested in on-prem staging environment where pg_trgm was in public schema
- Migration now runs successfully regardless of which schema contains pg_trgm
- Verified that gin_trgm_ops operator class is accessible after migration
- Tested fresh installations (no pg_trgm) and existing installations (pg_trgm in various schemas)

## Screenshots

N/A - Backend migration fix, no UI changes

## Checklist


backend/workflow_manager/workflow_v2/migrations/0018_filehistory_execution_count_and_more.py
```diff
@@ -1,19 +1,58 @@
...
+
 import django.core.validators
 from django.contrib.postgres.indexes import GinIndex
-from django.contrib.postgres.operations import TrigramExtension
 from django.db import migrations, models
 
+logger = logging.getLogger(__name__)
+
+
+def ensure_pg_trgm_extension(apps, schema_editor):
```
ritwik-g, line 12: @chandrasekharan-zipstack @muhammad-ali-e is this some kind of extension? If yes does this needs to be enabled in the db manually? Reason for this question is that could this break anything for customers using AWS or Azure if this extension is not available (like the vector db extension they need to enable separately)

chandrasekharan-zipstack: Had a similar discussion with @hari-kuriakose on this. It seems like this extension is available in GCP and AWS by default and needs to be enabled for Azure. Ultimately we can use `CREATE EXTENSION` to install it to a schema. However we felt that it needs to be a step involving manual intervention rather than being opinionated and carrying it out in a migration

chandrasekharan-zipstack:
Decided to go with a simple b-tree index for `LIKE file%` based pattern matching (starts with) and instead revert this `GinIndex` addition and `pg_trgm` extension installation
Refer #1706

PR #1724, opened 2025-12-29
[FEAT] Dashboard metrics system

athul-rs:
## What

- Add dashboard metrics Django app with multi-tier storage (hourly/daily/monthly tables)
- Add core metrics library with queue backend for event capture
- Add dedicated metrics Celery worker (worker-metrics)
- Add live query endpoints for immediate data access from source tables
- Add production readiness features (retry config, rate limiting, health checks)

## Why

- Enable tracking and visualization of platform usage metrics
- Provide real-time insights into documents processed, LLM calls, API requests, etc.
- Support dashboard reporting without impacting main application performance

## How

- Created `backend/dashboard_metrics/` Django app with models, views, services, and tasks
- Created `unstract/core/src/unstract/core/metrics/` library with queue backend using Kombu
- Used celery-batches for efficient batch processing of metric events
- Added dedicated Celery worker consuming from `dashboard_metric_events` queue
- Implemented live query endpoints that read directly from source tables (usage, page_usage, workflow_execution)

- No, this is an entirely new feature with no modifications to existing functionality
- New Django app is isolated and only adds new API endpoints
- New Celery worker runs on a separate queue

## Database Migrations

- 0001_initial: Creates EventMetricsHourly table
- 0002_setup_cleanup_periodic_task: Adds periodic cleanup task
- 0003_add_daily_monthly_tables: Creates EventMetricsDaily and EventMetricsMonthly tables
- 0004_update_cleanup_periodic_tasks: Adds daily cleanup periodic task

## Env Config

- WORKER_METRICS_AUTOSCALE (optional): Autoscale config for metrics worker (default: 4,1)
- DASHBOARD_METRICS_ENABLED (optional): Enable/disable metrics capture

## Relevant Docs

- N/A

## Related Issues or PRs

- UN-1798

## Dependencies Versions

- celery-batches>=0.8.0 (new dependency)

## Notes on Testing

- Run unit tests: `pytest backend/dashboard_metrics/tests/`
- Verify worker starts: `docker compose up worker-metrics`
- Test health endpoint: `GET /api/v2/metrics/health/`
- Verify metrics queue routing in RabbitMQ

## Screenshots

image
image
image

## Checklist

### DIFF
--- a/backend/prompt_studio/prompt_studio_core_v2/views.py
+++ b/backend/prompt_studio/prompt_studio_core_v2/views.py
@@ -4,6 +4,7 @@ import uuid
 from datetime import datetime
 from typing import Any
 
+import magic
 from account_v2.custom_exceptions import DuplicateData
 from api_v2.models import APIDeployment
 from django.db import IntegrityError
@@ -531,6 +532,28 @@ class PromptStudioCoreView(viewsets.ModelViewSet):
             file_name = (
                 f"{FileViewTypes.SUMMARIZE.lower()}/{filename_without_extension}.txt"
             )
+
+        # For ORIGINAL view, check if a converted PDF exists for preview
+        if (
+            view_type != FileViewTypes.EXTRACT
+            and view_type != FileViewTypes.SUMMARIZE
+            and file_converter_plugin
+        ):
+            converted_name = f"converted/{filename_without_extension}.pdf"
+            try:
+                contents = PromptStudioFileHelper.fetch_file_contents(
+                    file_name=converted_name,
+                    org_id=UserSessionUtils.get_organization_id(request),
+                    user_id=custom_tool.created_by.user_id,
+                    tool_id=str(custom_tool.tool_id),
+                    allowed_content_types=allowed_content_types,
+                )
+                return Response(contents, status=status.HTTP_200_OK)
+            except (FileNotFoundError, FileNotFound):
+                pass  # No converted file — fall through to return original
+            except Exception:
+                logger.exception(f"Error fetching converted file: {converted_name}")
+
         try:
             contents = PromptStudioFileHelper.fetch_file_contents(
                 file_name=file_name,
@@ -541,7 +564,7 @@ class PromptStudioCoreView(viewsets.ModelViewSet):
             )
         except FileNotFoundError:
             raise FileNotFound()
-        return Response({"data": contents}, status=status.HTTP_200_OK)
+        return Response(contents, status=status.HTTP_200_OK)
 
     @action(detail=True, methods=["post"])
     def upload_for_ide(self, request: HttpRequest, pk: Any = None) -> Response:
@@ -556,16 +579,32 @@ class PromptStudioCoreView(viewsets.ModelViewSet):
             # Store file
             file_name = uploaded_file.name
             file_data = uploaded_file
-            file_type = uploaded_file.content_type
-            # Convert non-PDF files
+            # Detect MIME from file content (not browser-supplied header)
+            file_type = magic.from_buffer(uploaded_file.read(2048), mime=True)
+            uploaded_file.seek(0)
+
             if file_converter_plugin and file_type != "application/pdf":
                 file_converter_service = file_converter_plugin["service_class"]()
-                file_data, file_name = file_converter_service.process_file(
-                    uploaded_file, file_name
-                )
+                if file_converter_service.should_convert_to_pdf(file_type):
+                    # Convert and store in converted/ subdir for preview
+                    converted_data, converted_name = file_converter_service.process_file(
+                        uploaded_file, file_name
+                    )
+                    PromptStudioFileHelper.upload_converted_for_ide(
+                        org_id=UserSessionUtils.get_organization_id(request),
+                        user_id=custom_tool.created_by.user_id,
+                        tool_id=str(custom_tool.tool_id),
+                        file_name=converted_name,
+                        file_data=converted_data,
+                    )
+                    # Reset uploaded_file for storing original in main dir
+                    uploaded_file.seek(0)
+                    file_data = uploaded_file
+                # else: CSV/TXT/Excel — file_data stays as original, no conversion
 
             logger.info(f"Uploading file: {file_name}" if file_name else "Uploading file")
 
+            # Store original file in main dir (always the original)
             PromptStudioFileHelper.upload_for_ide(
                 org_id=UserSessionUtils.get_organization_id(request),
                 user_id=custom_tool.created_by.user_id,
@@ -574,7 +613,7 @@ class PromptStudioCoreView(viewsets.ModelViewSet):
                 file_data=file_data,
             )
 
-            # Create a record in the db for the file
+            # Create a record in the db for the file (document_name = original filename)
             document = PromptStudioDocumentHelper.create(
                 tool_id=str(custom_tool.tool_id), document_name=file_name
             )
--- a/backend/utils/file_storage/helpers/prompt_studio_file_helper.py
+++ b/backend/utils/file_storage/helpers/prompt_studio_file_helper.py
@@ -38,6 +38,7 @@ class PromptStudioFileHelper:
         file_path = str(Path(base_path) / org_id / user_id / tool_id)
         extract_file_path = str(Path(file_path) / "extract")
         summarize_file_path = str(Path(file_path) / "summarize")
+        converted_file_path = str(Path(file_path) / "converted")
         if is_create:
             fs_instance = EnvHelper.get_storage(
                 storage_type=StorageType.PERMANENT,
@@ -46,6 +47,7 @@ class PromptStudioFileHelper:
             fs_instance.mkdir(file_path, create_parents=True)
             fs_instance.mkdir(extract_file_path, create_parents=True)
             fs_instance.mkdir(summarize_file_path, create_parents=True)
+            fs_instance.mkdir(converted_file_path, create_parents=True)
         return str(file_path)
 
     @staticmethod
@@ -77,6 +79,38 @@ class PromptStudioFileHelper:
         file_path = str(Path(file_system_path) / file_name)
         fs_instance.write(
             path=file_path,
+            mode="wb",
+            data=file_data if isinstance(file_data, bytes) else file_data.read(),
+        )
+
+    @staticmethod
+    def upload_converted_for_ide(
+        org_id: str, user_id: str, tool_id: str, file_data: Any, file_name: str
+    ) -> None:
+        """Stores converted PDF in the converted/ subdirectory for preview.
+
+        Args:
+            org_id (str): Organization ID
+            user_id (str): User ID
+            tool_id (str): ID of the prompt studio tool
+            file_data (Any): File data (bytes or file-like object)
+            file_name (str): Name of the converted file
+        """
+        fs_instance = EnvHelper.get_storage(
+            storage_type=StorageType.PERMANENT,
+            env_name=FileStorageKeys.PERMANENT_REMOTE_STORAGE,
+        )
+        file_system_path = (
+            PromptStudioFileHelper.get_or_create_prompt_studio_subdirectory(
+                org_id=org_id,
+                is_create=True,
+                user_id=user_id,
+                tool_id=str(tool_id),
+            )
+        )
+        converted_path = str(Path(file_system_path) / "converted" / file_name)
+        fs_instance.write(
+            path=converted_path,
             mode="wb",
             data=file_data if isinstance(file_data, bytes) else file_data.read(),
         )
@@ -141,12 +175,21 @@ class PromptStudioFileHelper:
             encoded_string = base64.b64encode(bytes(text_content_bytes))
             return {"data": encoded_string, "mime_type": file_content_type}
 
-        elif file_content_type == "text/plain":
+        elif file_content_type in ("text/plain", "text/csv"):
             text_content_string: str = fs_instance.read(
                 path=file_path,
                 mode="r",
                 legacy_storage_path=legacy_file_path,
                 encoding="utf-8",
+            )
+        elif file_content_type in (
+            "application/vnd.ms-excel",
+            "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+            "application/vnd.ms-excel.sheet.macroenabled.12",
+        ):
+            text_content_string = (
+                "Preview not available for Excel files. "
+                "Please index the document and view content in the Raw View tab."
             )
         # Check if the file type is in the allowed list
         elif file_content_type not in allowed_content_types:
@@ -178,7 +221,7 @@ class PromptStudioFileHelper:
         # Delete the source file
         fs_instance.rm(str(Path(file_system_path) / file_name))
         # Delete all related files for cascade delete
-        directories = ["extract/", "extract/metadata/", "summarize/"]
+        directories = ["extract/", "extract/metadata/", "summarize/", "converted/"]
         base_file_name, _ = os.path.splitext(file_name)
         # Delete related files
         file_paths = PromptStudioFileHelper._find_files(
--- a/frontend/src/components/custom-tools/document-manager/DocumentManager.jsx
+++ b/frontend/src/components/custom-tools/document-manager/DocumentManager.jsx
@@ -116,6 +116,7 @@ function DocumentManager({ generateIndex, handleUpdateTool, handleDocChange }) {
 
   useEffect(() => {
     // Convert blob URL to an object URL
+    console.log("here--->", fileData);
     if (fileData.blob) {
       const objectUrl = URL.createObjectURL(fileData.blob);
       setBlobFileUrl(objectUrl);
@@ -163,6 +164,7 @@ function DocumentManager({ generateIndex, handleUpdateTool, handleDocChange }) {
     Object.keys(viewTypes).forEach((item) => {
       handleFetchContent(viewTypes[item]);
     });
+    console.log(selectedDoc);
   }, [selectedDoc]);
 
   useEffect(() => {
@@ -220,7 +222,7 @@ function DocumentManager({ generateIndex, handleUpdateTool, handleDocChange }) {
   const handleGetDocumentsReq = (getDocsFunc, viewType) => {
     getDocsFunc(details?.tool_id, selectedDoc?.document_id, viewType)
       .then((res) => {
-        const data = res?.data?.data || "";
+        const data = res?.data || "";
         const mimeType = res?.data?.mime_type || "";
         processGetDocsResponse(data, viewType, mimeType);
       })
@@ -250,24 +252,39 @@ function DocumentManager({ generateIndex, handleUpdateTool, handleDocChange }) {
   };
 
   const processGetDocsResponse = (data, viewType, mimeType) => {
+    console.log("response-->", viewType, mimeType);
     if (viewType === viewTypes.original) {
-      const base64String = data || "";
-      const blob = base64toBlobWithMime(base64String, mimeType);
-      setFileData({ blob, mimeType });
-      const reader = new FileReader();
-      reader.readAsDataURL(blob);
-      reader.onload = () => {
-        setFileUrl(reader.result);
-      };
-      reader.onerror = () => {
-        throw new Error("Fail to load the file");
-      };
+      if (mimeType === "application/pdf") {
+        // Existing flow: base64 → blob → PdfViewer
+        const base64String = data || "";
+        const blob = base64toBlobWithMime(base64String, mimeType);
+        console.log("blob-->", blob);
+        setFileData({ blob, mimeType });
+        const reader = new FileReader();
+        reader.readAsDataURL(blob);
+        reader.onload = () => {
+          setFileUrl(reader.result);
+        };
+        reader.onerror = () => {
+          setFileErrMsg("Failed to load the file");
+        };
+      } else {
+        // Non-PDF file (CSV, TXT, Excel, or non-convertible)
+        // data is text, not base64
+        setFileUrl("");
+        setFileData({ blob: null, mimeType });
+        // Auto-switch to Raw View for non-PDF files
+        setActiveKey("2");
+      }
     } else if (viewType === viewTypes.extract) {
       setExtractTxt(data?.data);
     }
   };
 
   const handleGetDocsError = (err, viewType) => {
+    if (viewType === viewTypes.original) {
+      setFileData({});
+    }
     if (err?.response?.status === 404) {
       setErrorMessage(viewType);
     }
@@ -298,7 +315,7 @@ function DocumentManager({ generateIndex, handleUpdateTool, handleDocChange }) {
           info: "Clicked on the 'Summary View' tab",
         });
       }
-    } catch (err) {
+    } catch (_err) {
       // If an error occurs while setting custom posthog event, ignore it and continue
     }
   };
@@ -349,16 +366,20 @@ function DocumentManager({ generateIndex, handleUpdateTool, handleDocChange }) {
   };
 
   const renderDoc = (docName, fileUrl, highlightData) => {
-    const fileType = docName?.split(".").pop().toLowerCase(); // Get the file extension
-    switch (fileType) {
-      case "pdf":
-        return <PdfViewer fileUrl={fileUrl} highlightData={highlightData} />;
-      case "txt":
-      case "md":
-        return <TextViewer fileUrl={fileUrl} />;
-      default:
-        return <div>Unsupported file type: {fileType}</div>;
+    // Use mimeType from response for rendering decisions
+    console.log(fileData);
+    if (fileData.mimeType === "application/pdf") {
+      return <PdfViewer fileUrl={fileUrl} highlightData={highlightData} />;
     }
+    // Non-PDF: show placeholder message
+    return (
+      <div className="text-viewer-layout">
+        <Typography.Text type="secondary">
+          Document preview is not available for this file type. Please index the
+          document and switch to Raw View.
+        </Typography.Text>
+      </div>
+    );
   };
 
   return (
@@ -470,7 +491,10 @@ function DocumentManager({ generateIndex, handleUpdateTool, handleDocChange }) {
         <DocumentViewer
           doc={selectedDoc?.document_name}
           isLoading={isDocLoading}
-          isContentAvailable={fileUrl?.length > 0}
+          isContentAvailable={
+            fileUrl?.length > 0 ||
+            (fileData.mimeType && fileData.mimeType !== "application/pdf")
+          }
           setOpenManageDocsModal={setOpenManageDocsModal}
           errMsg={fileErrMsg}
         >
--- a/frontend/src/components/custom-tools/manage-docs-modal/ManageDocsModal.jsx
+++ b/frontend/src/components/custom-tools/manage-docs-modal/ManageDocsModal.jsx
@@ -459,7 +459,7 @@ function ManageDocsModal({
         info: "Clicked on index button",
         document_name: item?.document_name,
       });
-    } catch (err) {
+    } catch (_err) {
       // If an error occurs while setting custom posthog event, ignore it and continue
     }
   };
@@ -559,7 +559,7 @@ function ManageDocsModal({
       setPostHogCustomEvent("ps_uploaded_file", {
         info: "Clicked on '+ Upload New File' button",
       });
-    } catch (err) {
+    } catch (_err) {
       // If an error occurs while setting custom posthog event, ignore it and continue
     }
 
@@ -595,8 +595,18 @@ function ManageDocsModal({
           return; // Stop further execution
         }
 
-        // If the file is not a PDF, show the modal for confirmation
-        if (fileType !== "application/pdf") {
+        // File types that can be uploaded directly without conversion
+        const DIRECT_UPLOAD_TYPES = new Set([
+          "application/pdf",
+          "text/plain",
+          "text/csv",
+          "application/vnd.ms-excel",
+          "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+          "application/vnd.ms-excel.sheet.macroenabled.12",
+        ]);
+
+        if (!DIRECT_UPLOAD_TYPES.has(fileType)) {
+          // Non-direct types: show ConfirmMultiDoc modal or error
           if (!ConfirmMultiDoc) {
             setAlertDetails({
               type: "error",
@@ -606,7 +616,7 @@ function ManageDocsModal({
           setFileToUpload(file); // Store the file to be uploaded
           setIsModalVisible(true); // Show the modal
         } else {
-          // If the file is a PDF, proceed with the upload immediately
+          // PDF, CSV, TXT, Excel — proceed with the upload immediately
           resolve(file);
         }
       };
--- a/frontend/src/components/custom-tools/output-analyzer/OutputAnalyzerCard.jsx
+++ b/frontend/src/components/custom-tools/output-analyzer/OutputAnalyzerCard.jsx
@@ -57,7 +57,7 @@ function OutputAnalyzerCard({ doc, selectedPrompts, totalFields }) {
       setIsDocLoading(true);
       try {
         const res = await axiosPrivate.get(fileUrlEndpoint);
-        const base64String = res?.data?.data?.data || "";
+        const base64String = res?.data?.data || "";
         const blob = base64toBlob(base64String);
         setFileUrl(URL.createObjectURL(blob));
       } catch (err) {
--- a/frontend/src/components/navigations/side-nav-bar/SideNavBar.css
+++ b/frontend/src/components/navigations/side-nav-bar/SideNavBar.css
@@ -113,3 +113,12 @@
   background-color: rgba(255, 255, 255, 0.3);
   margin: 10px 0;
 }
+
+.sidebar-title-row {
+  display: flex;
+  align-items: center;
+}
+
+.sidebar-beta-tag {
+  margin-left: 8px;
+}
--- a/frontend/src/components/navigations/side-nav-bar/SideNavBar.jsx
+++ b/frontend/src/components/navigations/side-nav-bar/SideNavBar.jsx
@@ -361,7 +361,7 @@ const SideNavBar = ({ collapsed }) => {
 
   // Add Agentic Prompt Studio menu item if plugin is available and product is unstract
   if (agenticPromptStudioEnabled && isUnstract) {
-    data[0]?.subMenu?.splice(1, 0, {
+    const agenticMenuItem = {
       id: 1.2,
       title: "Agentic Prompt Studio",
       description: "Build and manage AI-powered extraction workflows",
@@ -370,7 +370,19 @@ const SideNavBar = ({ collapsed }) => {
       active: globalThis.location.pathname.startsWith(
         `/${orgName}/agentic-prompt-studio`,
       ),
-    });
+    };
+
+    // Add beta tag to title
+    agenticMenuItem.title = (
+      <>
+        {agenticMenuItem.title}
+        <Tag color="blue" className="sidebar-beta-tag">
+          BETA
+        </Tag>
+      </>
+    );
+
+    data[0]?.subMenu?.splice(1, 0, agenticMenuItem);
   }
 
   const shouldDisableAll = useMemo(() => {
@@ -491,7 +503,7 @@ const SideNavBar = ({ collapsed }) => {
                         />
                         {!collapsed && (
                           <div>
-                            <Typography className="sidebar-item-text fs-14">
+                            <Typography className="sidebar-item-text fs-14 sidebar-title-row">
                               {el.title}
                               {el.tag && (
                                 <Tag color="blue" className="sidebar-item-tag">


backend/backend/celery_config.py
```diff
@@ -31,3 +32,42 @@ class CeleryConfig:
...
+        "interval_start": 0,
+        "interval_step": 0.2,
+        "interval_max": 1,
+        "socket_timeout": 30,
+        "socket_keepalive": True,
+    }
+
+    # Worker configuration
+    worker_prefetch_multiplier = 4
+    worker_max_tasks_per_child = 1000
```
chandrasekharan-zipstack: @athul-rs any reason why this or other config was added? I hope it doesn't have any side effects on other workers

backend/dashboard_metrics/tests/test_tasks.py
chandrasekharan-zipstack, line 1: @athul-rs do these tests run on every PR as part of the existing unit tests? If not, please ensure that its integrated there as well otherwise your effort on these tests won't be realised

backend/dashboard_metrics/cache.py
```diff
@@ -0,0 +1,253 @@
...
+from django.utils import timezone
+
+logger = logging.getLogger(__name__)
+
+# Cache TTL values (in seconds)
+# Time-aware TTLs per documentation
+CACHE_TTL_CURRENT_HOUR = 30  # 30 seconds for current hour (updating frequently)
+CACHE_TTL_HISTORICAL = 8 * 60 * 60  # 8 hours for historical data (stable)
+
+# Legacy TTLs for compatibility with existing endpoints
```
chandrasekharan-zipstack: @athul-rs what existing endpoints do we have that require such legacy TTLs? Check and correct this comment

backend/dashboard_metrics/cache.py
```diff
@@ -0,0 +1,253 @@
...
+from typing import Any
+
+from django.core.cache import cache
+from django.utils import timezone
+
+logger = logging.getLogger(__name__)
+
+# Cache TTL values (in seconds)
+# Time-aware TTLs per documentation
+CACHE_TTL_CURRENT_HOUR = 30  # 30 seconds for current hour (updating frequently)
```
chandrasekharan-zipstack: NIT: @athul-rs I would suggest defining these values in `settings/base.py` so that they can be configured with an env too optionally with some sane defaults. Make sure the names are updated / prefixed with `DASHBOARD_` to group them separately

backend/dashboard_metrics/cache.py
```diff
@@ -0,0 +1,253 @@
...
+        org_id: Organization identifier
+        endpoint: API endpoint name (e.g., 'summary', 'series')
+        params: Query parameters
+
+    Returns:
+        Unique cache key string
+    """
+    # Sort params for consistent hashing
+    sorted_params = json.dumps(params, sort_keys=True, default=str)
+    params_hash = hashlib.md5(sorted_params.encode()).hexdigest()[:12]
```
chandrasekharan-zipstack: NIT: Possibility for cache collision since we only take first 12 digits, but should suffice mostly since we separate for endpoints. Do check if a change is needed here

backend/dashboard_metrics/cache.py
```diff
@@ -0,0 +1,253 @@
...
+
+                return Response(cached)
+
+            # Execute function and cache result
+            response = func(self, request, *args, **kwargs)
+
+            if response.status_code == 200:
+                cache_ttl = ttl or default_ttls.get(endpoint, CACHE_TTL_SUMMARY)
+                cache.set(cache_key, response.data, cache_ttl)
+                logger.debug(f"Cached {endpoint} for {cache_ttl}s: {cache_key}")
```
chandrasekharan-zipstack, line 124: NIT: Reconsider if INFO log level will be useful here. Of course it depends on whether too much noise is present, however such information can be useful during debugging prod issues. Similarly consider for the other log as well

backend/dashboard_metrics/cache.py
```diff
@@ -0,0 +1,253 @@
...
+    def warm_cache(org_id: str) -> None:
+        """Pre-warm cache with commonly accessed data.
+
+        Useful for background tasks to prepare cache after
+        new metric data is processed.
+        """
+        # Import here to avoid circular imports
+
+        from django.db.models import Sum
+        from django.utils import timezone
```
chandrasekharan-zipstack: @athul-rs these imports can be moved to the top. Correct the below relative import to an absolute import instead. I don't think there'll be a circular import issue here, but do check and add it only if necessary

backend/dashboard_metrics/cache.py
```diff
@@ -0,0 +1,253 @@
...
+                    total_value=Sum("metric_value"),
+                    total_count=Sum("metric_count"),
+                )
+            )
+
+            MetricsCacheService.set_cached_overview(org_id, list(overview))
+            logger.debug(f"Warmed overview cache for org {org_id}")
+
+        except Exception as e:
+            logger.warning(f"Failed to warm cache for org {org_id}: {e}")
```
chandrasekharan-zipstack: Consider printing stack trace as well

backend/dashboard_metrics/migrations/0003_add_daily_monthly_tables.py
chandrasekharan-zipstack, line 1: @athul-rs consider squashing migrations 1, 3 together and 2,4 together

backend/dashboard_metrics/serializers.py
```diff
@@ -0,0 +1,124 @@
...
+
+        if not attrs.get("end_date"):
+            attrs["end_date"] = now
+
+        if not attrs.get("start_date"):
+            attrs["start_date"] = now - timedelta(days=30)
+
+        if attrs["start_date"] > attrs["end_date"]:
+            raise serializers.ValidationError(
+                {"start_date": "start_date must be before end_date"}
```
chandrasekharan-zipstack:
```suggestion
                {"start_date": "Start date must be before end date"}
```
Making it user friendly

backend/dashboard_metrics/tasks.py
```diff
@@ -0,0 +1,458 @@
...
+    EventMetricsHourly,
+    EventMetricsMonthly,
+    MetricType,
+)
+
+logger = logging.getLogger(__name__)
+
+# Default retention periods
+DEFAULT_RETENTION_DAYS_HOURLY = 30
+DEFAULT_RETENTION_DAYS_DAILY = 365
```
chandrasekharan-zipstack:
```suggestion
DASHBOARD_HOURLY_METRICS_RETENTION_DAYS = 30
DASHBOARD_DAILY_METRICS_RETENTION_DAYS = 365
```
Rename to make it more explicit and meaningful. Prefixing with `DEFAULT_` doesn't add any additional meaning

backend/dashboard_metrics/tasks.py
```diff
@@ -0,0 +1,458 @@
...
+            monthly_agg[monthly_key]["value"] += metric_value
+            monthly_agg[monthly_key]["count"] += 1
+            if labels:
+                monthly_agg[monthly_key]["labels"].update(labels)
+
+        except KeyError as e:
+            logger.warning(f"Skipping event with missing required field: {e}")
+            errors += 1
+        except Exception as e:
+            logger.warning(f"Error processing event: {e}")
```
chandrasekharan-zipstack: NIT: Consider adding a stack trace

backend/dashboard_metrics/views.py
```diff
@@ -0,0 +1,451 @@
...
+    ordering = ["-timestamp"]
+
+    def get_queryset(self):
+        """Return queryset filtered by user's organization."""
+        try:
+            organization = UserContext.get_organization()
+            if not organization:
+                logger.warning("No organization context for metrics request")
+                raise PermissionDenied("No organization context")
+            return EventMetricsHourly.objects.filter(organization=organization)
```
chandrasekharan-zipstack: @athul-rs check if this is required. This would be handled by the model manager itself usually

backend/dashboard_metrics/cache.py
chandrasekharan-zipstack, line 1: NIT: This is a cool idea and if refactored a bit to make it generic, we can easily extend the same to any other API / DB call that we wish to cache. Consider this suggestion if its easy to make it generic

unstract/core/src/unstract/core/metrics/backends/noop.py
```diff
@@ -0,0 +1,44 @@
+"""No-op backend that discards all metrics.
+
+Used when metrics recording is disabled (default for OSS).
+"""
+
+import logging
+
+from ..types import MetricEvent
+from .base import AbstractMetricBackend
```
chandrasekharan-zipstack: NIT: Consider absolute imports instead

unstract/core/src/unstract/core/metrics/backends/queue.py
```diff
@@ -0,0 +1,215 @@
...
+        self._max_retries = 3
+        self._base_delay = 1.0
+
+    def _get_connection(self) -> Connection:
+        """Get or create a connection to the broker with retry logic."""
+        if self._connection is not None and self._connection.connected:
+            return self._connection
+        return self._get_connection_with_retry()
+
+    def _get_connection_with_retry(
```
chandrasekharan-zipstack: @athul-rs why do we need to implement this? Isn't it available / handled within celery / kombu?

backend/backend/settings/base.py
```diff
@@ -130,6 +130,23 @@ def get_required_setting(setting_key: str, default: str | None = None) -> str |
     os.environ.get("API_DEPLOYMENT_RATE_LIMIT_LOCK_BLOCKING_TIMEOUT", 5)
 )
 
+# Dashboard Metrics Cache TTL settings (in seconds)
+DASHBOARD_CACHE_TTL_CURRENT_HOUR = int(
```
chandrasekharan-zipstack: @athul-rs do you think a 30s cache TTL is too aggressive here for hourly information? 
1. How expensive is the query which is being cached?
2. If the data that we show using this cache is mainly charts over a longer period, I don't think it needs to be this recent (maybe 1-2 mins could be fine)

backend/backend/celery_config.py
chandrasekharan-zipstack: @athul-rs do check this comment. All this config that you've added will impact all workers. If its default values, it might be okay

backend/dashboard_metrics/management/commands/generate_metrics_test_data.py
```diff
@@ -0,0 +1,425 @@
...
+
+        # Delete page usage records
+        page_count = PageUsage.objects.filter(organization_id=org_id_str).count()
+        PageUsage.objects.filter(organization_id=org_id_str).delete()
+
+        self.stdout.write(
+            f"  Deleted: {exec_count} executions, {file_exec_count} file executions, "
+            f"{usage_count} usage, {page_count} page usage"
+        )
+
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: @athul-rs check this

backend/dashboard_metrics/management/commands/generate_metrics_test_data.py
chandrasekharan-zipstack, line 1: @athul-rs instead of a management command, can this be a script in a relevant `/tests` folder? I assume this can also be the seed data for a set of dashboard related tests that way. 

Also this command is something that might only help during local development and doesn't make much sense when you consider prod - to avoid confusion if we absolutely need to have a management command, we should prefix it with `dev_` or something along those lines

backend/dashboard_metrics/management/commands/populate_test_metrics.py
chandrasekharan-zipstack, line 1: @athul-rs similar to the above comment, consider moving these commands to a tests folder where it can help populate seed data for a set of tests to run

backend/dashboard_metrics/migrations/0004_fix_tag_not_null.py
chandrasekharan-zipstack, line 1: @athul-rs can you squash this migration with 0001 itself? Since we're introducing these migrations in this PR this particular data migration might not be needed

backend/backend/celery_config.py
athul-rs: Not needed, changed approach here. updating.

backend/dashboard_metrics/cache.py
athul-rs: Chances are low to null,  this only happens within org and endpoint with 12 hex chars there is enough possibilkites

backend/dashboard_metrics/cache.py
athul-rs: @chandrasekharan-zipstack This would be called on every single API request to cached endpoint, the logs will be too flooded with these. Better to stay in debug

frontend/src/components/metrics-dashboard/MetricsChart.jsx
```diff
@@ -0,0 +1,403 @@
...
+    return (
+      <Card
+        title="Trend Analysis (Last 30 Days)"
+        className="metrics-chart-card"
+      >
+        <Empty description="No data available" />
+      </Card>
+    );
+  }
+
```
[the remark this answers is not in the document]
athul-rs: Not really — a user who manually deselects all metrics via the Filter dropdown understands why lines disappeared (they did it themselves). And the default selectedMetrics always starts with valid values (documents_processed, llm_calls), so this state only happens through deliberate user action. Adding an explicit empty state for self-inflicted deselection would be unnecessary UI noise.

frontend/src/components/metrics-dashboard/RecentActivity.jsx
```diff
@@ -0,0 +1,153 @@
...
+                <Text className="recent-activity-file" ellipsis>
+                  {item.file_name}
+                </Text>
+                <Text type="secondary" className="recent-activity-time">
+                  {dayjs(item.created_at).fromNow()}
+                </Text>
+              </div>
+            </List.Item>
+          );
+        }}
```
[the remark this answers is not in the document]
athul-rs: Not relevant

backend/dashboard_metrics/migrations/0002_setup_periodic_tasks.py
```diff
@@ -0,0 +1,108 @@
+"""Data migration to create periodic tasks for dashboard metrics.
+
+Creates three periodic tasks:
+- Aggregation: Every 15 minutes, queries source tables and populates
+  EventMetricsHourly, EventMetricsDaily, and EventMetricsMonthly
+- Hourly cleanup: Daily at 2:00 AM UTC, removes hourly data older than 30 days
+- Daily cleanup: Weekly on Sundays at 3:00 AM UTC, removes daily data older than 365 days
```
chandrasekharan-zipstack:
```suggestion
- Hourly data cleanup: Daily at 2:00 AM UTC, removes hourly data older than 30 days
- Daily data cleanup: Weekly on Sundays at 3:00 AM UTC, removes daily data older than 365 days
```

backend/dashboard_metrics/migrations/0002_setup_periodic_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+        name="dashboard_metrics_cleanup_daily",
+        defaults={
+            "task": "dashboard_metrics.cleanup_daily_data",
+            "crontab": schedule_3am_sun,
+            "queue": "dashboard_metric_events",
+            "kwargs": '{"retention_days": 365}',
+            "enabled": True,
+            "description": "Clean up daily dashboard metrics older than 365 days",
+        },
+    )
```
chandrasekharan-zipstack, line 81: @athul-rs should these cleanup tasks be linked to the cleanup of execution logs and other linked resources as well? The hourly data cleanup seems fine since we'll retain executions and logs history for more than a month atleast.

However with the daily data cleanup that retains for 365 days, we could treat it as part of the larger execution and logs history cleanup. This could be easily facilitated through cascade deletes and also save us some hassle in the future
cc: @muhammad-ali-e @hari-kuriakose

backend/dashboard_metrics/management/commands/backfill_metrics.py
chandrasekharan-zipstack, line 1: @athul-rs how will this command be run during deployment? Is it a separate K8s job? Why isn't it a data migration instead?

backend/dashboard_metrics/tests/test_tasks.py
chandrasekharan-zipstack: This could be easy to setup actually if you're tests are self contained and can be run anywhere

backend/dashboard_metrics/models.py
chandrasekharan-zipstack, line 1: NIT: Consider following a rule of only 1 model per file. It makes it easy for humans and agents to understand context easily. Should be an easy refactor with Claude

chandrasekharan-zipstack:
@athul-rs Few minor points from one of the screenshots - 
1. Can we remove the `pages` and `docs` suffix that we add to the cards?
2. Rename `Total Pages Processed` -> `Pages Processed` for brevity in the card
3. How does the line charts behave when the date time filter at the top bar is:
    - Less than a month
    - More than a month

backend/dashboard_metrics/migrations/0002_setup_periodic_tasks.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def create_periodic_tasks(apps, schema_editor):
+    """Create all periodic tasks for metrics processing."""
+    CrontabSchedule = apps.get_model("django_celery_beat", "CrontabSchedule")
+    IntervalSchedule = apps.get_model("django_celery_beat", "IntervalSchedule")
+    PeriodicTask = apps.get_model("django_celery_beat", "PeriodicTask")
+
+    # --- Aggregation task: every 15 minutes ---
+    schedule_15min, _ = IntervalSchedule.objects.get_or_create(
```
muhammad-ali-e, line 20: It would be better to create a worker that runs every N seconds.
Please check `workers/log_consumer`.
Avoid using periodic tasks or Celery Beat for this.

backend/dashboard_metrics/cache.py
athul-rs: No these are not fully self-contained. They won't work iun CI without the db setups like tables and other envs included.

backend/dashboard_metrics/tasks.py
```diff
@@ -0,0 +1,493 @@
...
+- aggregate_metrics_from_sources: Periodic aggregation from source tables
+- cleanup_hourly_metrics: Remove hourly metrics older than retention period
+- cleanup_daily_metrics: Remove daily metrics older than retention period
+"""
+
+import logging
+from datetime import datetime, timedelta
+from typing import Any
+
+from celery import shared_task
```
muhammad-ali-e, line 13: please use `workers` for tasks , We already dedjangofying the clery and moved away from backend.

muhammad-ali-e (commented):
@athul-rs please move celery tasks from backend to workers

But I think the breat tasks are still in backend, So my comment might be invalidated

backend/dashboard_metrics/tasks.py
muhammad-ali-e: Note: But I think the breat tasks are still in backend, So my comment might be invalidated

backend/dashboard_metrics/views.py
```diff
@@ -0,0 +1,944 @@
...
+                "end_date": params["end_date"].isoformat(),
+                "source": "live",
+                "summary": summary_list,  # Changed from "metrics" to "summary"
+            },
+            status=status.HTTP_200_OK,
+        )
+
+    @action(detail=False, methods=["get"], url_path="live-series")
+    @cache_metrics_response(endpoint="live_series")
+    def live_series(self, request: Request) -> Response:
```
muhammad-ali-e, line 743: is it from aggregted table or or actual source tables?

docker/docker-compose.yaml
```diff
@@ -117,6 +117,25 @@ services:
       - ./workflow_data:/data
       - ${TOOL_REGISTRY_CONFIG_SRC_PATH}:/data/tool_registry_config
 
+  # Celery worker for dashboard metrics processing
+  worker-metrics:
```
muhammad-ali-e, line 122: What this worker do? How this diffrent from tasks in backend/dashborad_metrics , is it not by beat?

backend/dashboard_metrics/cache.py
athul-rs: The defaults are already sane (30s for live bucket, 5min overview, 15min summary, 30min series, 1hr workflow usage, 8hr historical). No env vars needed in .env unless someone wants to tune them.

athul-rs: moving it to base.py

backend/dashboard_metrics/tasks.py
athul-rs: track trace will leave too much noise, changed it to track exception.

backend/dashboard_metrics/views.py
athul-rs: updated.

unstract/core/src/unstract/core/metrics/backends/noop.py
athul-rs: removed file

unstract/core/src/unstract/core/metrics/backends/queue.py
athul-rs: Not releavnt anymore

backend/dashboard_metrics/management/commands/generate_metrics_test_data.py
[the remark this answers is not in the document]
athul-rs: removed file

backend/dashboard_metrics/management/commands/backfill_metrics.py
```diff
@@ -0,0 +1,360 @@
...
+                        },
+                    )
+                    if was_created:
+                        created += 1
+                    else:
+                        updated += 1
+                except Exception as e:
+                    logger.warning(f"Error upserting hourly {key}: {e}")
+
+        return created, updated
```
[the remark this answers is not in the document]
athul-rs: No, that concern is not relevant here. The backfill_metrics.py command does not use transaction.atomic() at all. Each update_or_create call in _upsert_hourly, _upsert_daily, and _upsert_monthly is independently atomic (Django's default), and exceptions are caught per-row with except Exception — so one failure doesn't break subsequent iterations.

It looks like that transaction.atomic() issue was from a previous version that was already fixed. The current code is fine.

backend/dashboard_metrics/tasks.py
```diff
@@ -0,0 +1,805 @@
...
+                            "metric_count",
+                            "labels",
+                            "modified_at",
+                        ]
+                    )
+                    updated += 1
+            except Exception:
+                logger.exception("Error upserting hourly metric %s", key)
+
+    return created, updated
```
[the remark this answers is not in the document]
athul-rs: Django's update_or_create() already handles the race condition through two mechanisms:

How Django protects against concurrent upserts
select_for_update() — The implementation wraps the operation in transaction.atomic() and uses SELECT FOR UPDATE to acquire a row-level lock before checking/updating. This prevents two concurrent tasks from both trying to insert the same row.

Single-retry in get_or_create() — Inside update_or_create, if the initial get() finds nothing and the create() raises IntegrityError (because another task inserted first), Django retries with a get() once. If it finds the now-existing row, it proceeds to update it normally.

frontend/src/hooks/useMetricsData.js
```diff
@@ -0,0 +1,123 @@
...
+      }
+    },
+    [axiosPrivate, orgId, startDate, endDate]
+  );
+
+  useEffect(() => {
+    if (orgId) {
+      fetchOverview();
+    }
+  }, [orgId, startDate, endDate]); // eslint-disable-line
```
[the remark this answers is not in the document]
athul-rs: Not an issue — this is already handled. All three hooks (useMetricsOverview, useRecentActivity, useWorkflowTokenUsage) already have the else branch that resets state when orgId is falsy:

frontend/src/hooks/useMetricsData.js
```diff
@@ -0,0 +1,123 @@
...
+    } finally {
+      setLoading(false);
+    }
+  }, [axiosPrivate, orgId, limit]);
+
+  useEffect(() => {
+    if (orgId) {
+      fetchActivity();
+    }
+  }, [orgId, limit]); // eslint-disable-line
```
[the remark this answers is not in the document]
athul-rs: That's already handled. Looking at useRecentActivity

backend/dashboard_metrics/tasks.py
```diff
@@ -0,0 +1,507 @@
...
+                            metric_name,
+                            "default",
+                            "",
+                        )
+                        monthly_agg[month_key] = {
+                            "metric_type": metric_type,
+                            "value": bucket["value"],
+                            "count": bucket["count"] or 1,
+                            "labels": {},
+                        }
```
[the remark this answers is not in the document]
athul-rs: The observation is technically correct but not practically harmful in the current system. Here's why:

What's happening: Each service method (e.g., get_documents_processed) returns one row per time bucket with {period, value}. The aggregator in tasks.py does count += 1 per row, so metric_count = number of source query result rows, not number of underlying events.

Why it doesn't matter right now:

For counters like documents_processed, each hourly query returns one row per hour with Count("id") as the value. The aggregator groups by the same hour key, so for a given (org, hour, metric), there's exactly 1 row → metric_count = 1. The count is technically "buckets" but happens to be 1.
For daily/monthly, same thing — one source row maps to one aggregated row.
The frontend doesn't use total_count for display — only total_value drives the numbers users see. total_count is just in the PropTypes definition but isn't rendered.
The live_summary endpoint already hardcodes "total_count": 1 at [views.py:696](vscode-webview://1b01dqqlbq9cdd6bqhfhosb10bu5493jn64n8kvfgen8trd6d2hk/backend/dashboard_metrics/views.py#L696), acknowledging it's not meaningful from live queries.
Bottom line: It's a semantic mismatch in the data model (count says "events" but stores "buckets"), but since metric_count is never surfaced to users and the 1:1 bucket mapping means the value is typically 1 anyway, it's cosmetic — not a data integrity issue. Not worth changing for this PR.

backend/backend/settings/base.py
athul-rs: It's a simple indexed range scan — not expensive per bucket. But the data it's caching is for dashboard charts showing trends over days/weeks. Users aren't expecting real-time second-by-second updates on chart data. 60s/120s is a reasonable middle ground.

backend/dashboard_metrics/management/commands/generate_metrics_test_data.py
athul-rs: wan a local test script, ignore got pushed accidentally.

backend/dashboard_metrics/migrations/0002_setup_periodic_tasks.py
athul-rs: This concern doesn't apply here. The metrics tables are self-contained aggregated summaries with no FK relationships to execution/log tables.

Looking at the models:

EventMetricsHourly, EventMetricsDaily, EventMetricsMonthly only have:
organization FK (from DefaultOrganizationMixin)
Scalar fields: metric_name, metric_value, metric_count, project, tag, timestamp/date/month
Zero foreign keys to WorkflowExecution, Usage, PageUsage, or any source table
These tables store pre-computed aggregates like "documents_processed: 42 on 2025-02-10" — they're denormalized summaries, not relational data pointing back to individual execution rows.

So:

Cascade deletes don't apply — there's nothing to cascade to or from
The retention policies are independent — even if source execution records are deleted, the aggregated metrics remain valid historical summaries (that's the whole point of pre-aggregation)
Different retention needs — you might want 30-day raw execution logs but 365-day dashboard trend data, or vice versa. Coupling them would remove that flexibility.
@chandrasekharan-zipstack

backend/dashboard_metrics/management/commands/backfill_metrics.py
athul-rs: > @athul-rs how will this command be run during deployment? Is it a separate K8s job? Why isn't it a data migration instead?

Intentionally done this way: Backfilling 90 days of data across many orgs could take significant time. A data migration would block the entire migration chain and slow every deploy. Rerunning if required is also a factor.

athul-rs:
> @athul-rs Few minor points from one of the screenshots -
> 
> 1. Can we remove the `pages` and `docs` suffix that we add to the cards?
> 2. Rename `Total Pages Processed` -> `Pages Processed` for brevity in the card
> 3. How does the line charts behave when the date time filter at the top bar is:
>    
>    * Less than a month
>    * More than a month

@chandrasekharan-zipstack Updated.
Charts are set at 30 days period and doesn't get affeced by top date picker

backend/dashboard_metrics/migrations/0002_setup_periodic_tasks.py
athul-rs: any reason to avoid using celery beat for this ?

backend/dashboard_metrics/views.py
athul-rs: It queries directly from source tables (Usage, PageUsage, WorkflowExecution, etc.)

athul-rs: It is being removed, not used now.

docker/docker-compose.yaml
athul-rs: dashboard_metrics/tasks.py defines the tasks (aggregate_metrics_from_sources, cleanup_hourly_metrics, cleanup_daily_metrics)
Celery Beat schedules them (via the migration that creates PeriodicTask entries routed to the dashboard_metric_events queue)
worker-metrics docker service consumes that queue: celery -A backend worker -Q dashboard_metric_events
It's the same relationship as worker-logging consuming the celery_periodic_logs queue. The dedicated docker service gives it its own process/scaling independent of the general worker.

backend/dashboard_metrics/tasks.py
athul-rs: What exactly need to be changed ? since "dedjango-fying" pattern applies to worker apps, not scheduling. Correct me if am wrong ?

muhammad-ali-e:
### Code review

Found 1 issue:

1. `window.location.pathname` used instead of `globalThis.location.pathname` -- regression of commit ed27ec87 which deliberately migrated all `active:` checks in this file from `window` to `globalThis`. Every other active check (15+ instances, e.g. line 116, 193, 202, 218, 226) uses `globalThis.location.pathname`.

https://github.com/Zipstack/unstract/blob/3756ed4974bd9f79b2a390b77fd1b9879faa6981/frontend/src/components/navigations/side-nav-bar/SideNavBar.jsx#L322-L332

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

backend/dashboard_metrics/cache.py
```diff
@@ -0,0 +1,405 @@
+"""Cache layer for Dashboard Metrics.
+
+Provides caching for computed metric aggregations to reduce database load.
+Uses Redis through Django's cache framework.
+
+Cache TTL Strategy:
+- Current hour/day/month data: Configurable via DASHBOARD_CACHE_TTL_CURRENT_BUCKET (default 60s)
```
muhammad-ali-e, line 7: Hope its count and size are not huge, so it won’t fill the Redis memory quickly.

backend/sample.env
```diff
@@ -231,6 +231,13 @@ RUNNER_POLLING_INTERVAL_SECONDS=2
 # Examples: 900 (15 min), 1800 (30 min), 3600 (60 min)
 MIN_SCHEDULE_INTERVAL_SECONDS=1800
 
+# Dashboard Metrics Cache TTLs (in seconds)
```
muhammad-ali-e, line 239: why it commnetd?

backend/dashboard_metrics/services.py
```diff
@@ -0,0 +1,763 @@
...
+    - etl_pipeline_executions: from workflow_execution + pipeline
+    - llm_usage: from usage (cost_in_dollars)
+    - prompt_executions: from workflow_execution
+    """
+
+    @staticmethod
+    def _get_trunc_func(granularity: str):
+        """Get the appropriate truncation function for the granularity."""
+        trunc_map = {
+            "hour": TruncHour,
```
muhammad-ali-e: Making these hour,day etc.. to an enum would be better

backend/dashboard_metrics/views.py
```diff
@@ -0,0 +1,936 @@
...
+                        "metric_name",
+                        "metric_type",
+                        "metric_value",
+                        "metric_count",
+                        "project",
+                        "tag",
+                        "timestamp",
+                    )
+                )
+                db_data_by_bucket[bucket_ts] = bucket_records
```
chandrasekharan-zipstack: **N+1 Query on Cold Cache**: This loop fires one DB query per hourly bucket. A default 7-day range = 168 sequential SQL roundtrips on cache miss (~8-10s at 50ms/query).

Should be a single range query + in-memory partitioning:

```python
if missing_buckets:
    min_ts = min(missing_buckets)
    max_ts = max(missing_buckets) + timedelta(hours=1)
    qs = EventMetricsHourly.objects.filter(
        timestamp__gte=min_ts, timestamp__lt=max_ts,
    )
    # apply metric_name, project, tag filters...
    all_records = list(qs.values(...))
    for record in all_records:
        bucket_ts = record["timestamp"].replace(minute=0, second=0, microsecond=0)
        db_data_by_bucket.setdefault(bucket_ts, []).append(record)
```

<details>
<summary>Prompt for coding agent to validate & implement this fix</summary>

```
In backend/dashboard_metrics/views.py, the method `_fetch_hourly_with_bucket_cache` has an N+1 query problem. The `for bucket_ts in missing_buckets:` loop fires one DB query per hourly bucket (up to 168 for a 7-day range).

Fix: Replace the loop with a single query that fetches all missing buckets at once, then partitions results into buckets in Python.

1. Find `min_ts = min(missing_buckets)` and `max_ts = max(missing_buckets) + timedelta(hours=1)`
2. Run ONE query: `EventMetricsHourly.objects.filter(timestamp__gte=min_ts, timestamp__lt=max_ts)` with the same metric_name/project/tag filters
3. Partition results into `db_data_by_bucket` dict by truncating each record's timestamp to the hour
4. Keep the rest of the method (cache write + combine) unchanged
5. Verify the same bucket caching logic still works by checking that `mset_metrics_buckets` receives the same structure
6. Also check if `_fetch_daily_with_bucket_cache` or `_fetch_monthly_with_bucket_cache` (if they exist) have the same pattern and fix those too
```

</details>

backend/dashboard_metrics/views.py
```diff
@@ -0,0 +1,936 @@
...
+
+    def _check_database(self) -> dict:
+        """Check database connectivity."""
+        try:
+            EventMetricsHourly.objects.exists()
+            return {"status": "ok"}
+        except Exception as e:
+            logger.exception("Database health check failed")
+            return {"status": "error", "message": str(e)}
+
```
chandrasekharan-zipstack, line 939: Avoid this try-catch altogether and let the exception handling middleware take care of it. Or if we wish to hide internals, simple re-raise an exception from e with a different error message. Make sure to update this elsewhere in the repo too

backend/dashboard_metrics/tasks.py
```diff
@@ -0,0 +1,499 @@
...
+    Returns:
+        Tuple of (created_count, updated_count)
+    """
+    created, updated = 0, 0
+    for key, agg in aggregations.items():
+        org_id, hour_ts_str, metric_name, project, tag = key
+        hour_ts = datetime.fromisoformat(hour_ts_str)
+
+        try:
+            obj, was_created = EventMetricsHourly._base_manager.update_or_create(
```
chandrasekharan-zipstack: **Row-by-row upsert**: Each `update_or_create` = 2 SQL roundtrips (SELECT + INSERT/UPDATE). With 12 metrics × 24 buckets × N orgs, this generates thousands of queries every 15 min against production DB.

The models already have `UniqueConstraint` on `(organization, timestamp, metric_name, project, tag)` — use `bulk_create` with `update_conflicts` to do it in a single `INSERT ... ON CONFLICT` per tier:

```python
EventMetricsHourly._base_manager.bulk_create(
    objects,
    update_conflicts=True,
    unique_fields=["organization", "timestamp", "metric_name", "project", "tag"],
    update_fields=["metric_type", "metric_value", "metric_count", "modified_at"],
)
```

Also worth adding a Redis distributed lock to prevent overlapping runs when the task takes longer than 15 min.

<details>
<summary>Prompt for coding agent</summary>

```
In backend/dashboard_metrics/tasks.py, refactor the three upsert functions (_upsert_hourly_replace, _upsert_daily_replace, _upsert_monthly_replace) to use bulk_create with update_conflicts instead of row-by-row update_or_create.

For each function:
1. Build a list of model objects from the aggregations dict
2. Use Model._base_manager.bulk_create(objects, update_conflicts=True, unique_fields=[...], update_fields=[...])
3. unique_fields should match the existing UniqueConstraint on each model
4. update_fields should be: metric_type, metric_value, metric_count, modified_at
5. Return the count of objects processed (you lose the created/updated split, which is acceptable)

Also add a Redis-based distributed lock to aggregate_metrics_from_sources to prevent overlapping runs:
- Use django.core.cache or redis.lock.Lock with a timeout matching the task schedule (15 min)
- If lock cannot be acquired, skip the run and log a warning
```

</details>

backend/dashboard_metrics/management/commands/backfill_metrics.py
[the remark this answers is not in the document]
athul-rs: @coderabbit check again.

backend/dashboard_metrics/migrations/0002_setup_periodic_tasks.py
athul-rs: Addressing in future PRs.

backend/dashboard_metrics/cache.py
athul-rs: Cache key count and memory per org:

There are two types of cache entries:

API response cache (5 endpoints: overview, summary, series, workflow_token_usage, recent_activity)

Keys: dashboard_metrics:<org_id>:<endpoint>:<params_hash>
Count: ~5-10 per org (one per unique query param combination)
TTL: 60-300s (short-lived)
Bucket cache (per-metric, per-time-bucket)

Keys: dashboard_metrics:bucket:<granularity>:<org_id>:<bucket_ts>
Worst case for a 7-day series query: ~168 hourly buckets or ~7 daily buckets
With 12 metrics queried individually: up to ~168 * 12 = ~2016 keys per org
But in practice, the series endpoint uses daily granularity (not hourly), so it's ~7 * 12 = ~84 keys per org
Each value is a small JSON array (a few hundred bytes per bucket)
Historical buckets: TTL 8 hours (28800s), auto-expire
Current bucket: TTL 60s, refreshed frequently
Total per org: ~100 keys, ~50-100 KB. Even with 100 orgs, that's ~10K keys and ~10 MB — negligible for Redis.

backend/sample.env
athul-rs: default handled it base.py, kept it intentionally. Removing them.

backend/dashboard_metrics/management/commands/test_metrics_e2e.py
```diff
@@ -0,0 +1,446 @@
...
+            self.stdout.write(self.style.ERROR(f"=== FAILED: {', '.join(failed)} ==="))
+
+    def _get_org(self, org_id):
+        """Resolve organization by ID or pick the first one."""
+        if org_id:
+            try:
+                return Organization.objects.get(id=org_id)
+            except Organization.DoesNotExist:
+                return None
+        return Organization.objects.first()
```
muhammad-ali-e: will this take real data from DB ? if yes we should move this test file from commands. else it might alter the real data.

backend/dashboard_metrics/views.py
```diff
@@ -0,0 +1,941 @@
...
+        org_id = str(organization.id)
+        try:
+            limit = min(int(request.query_params.get("limit", 10)), 50)
+        except (ValueError, TypeError):
+            limit = 10
+
+        try:
+            data = MetricsQueryService.get_recent_activity(org_id, limit)
+            return Response({"activity": data})
+        except Exception as e:
```
muhammad-ali-e: - We typically raise API exceptions rather than returning a Response directly.
- Additionally, catching generic exceptions and returning the complete error response is not always helpful.
- This approach may expose database structure, model names, and query information (although this is less concerning because the code is open source).

backend/dashboard_metrics/views.py
athul-rs: Single range query using min(missing_buckets) to max(missing_buckets) + 1h, then partitions records into hourly buckets in memory using a set lookup. This reduces it to 1 SQL roundtrip regardless of the number of missing buckets.

Empty bucket caching: Added setdefault(bucket_ts, []) for all missing buckets so that hours with zero data are also cached, preventing unnecessary re-queries on the next request.

backend/dashboard_metrics/services.py
```diff
@@ -0,0 +1,765 @@
...
+
+        Returns:
+            List of dicts with 'period' and 'value' keys
+        """
+        trunc_func = MetricsQueryService._get_trunc_func(granularity)
+
+        return list(
+            WorkflowFileExecution.objects.filter(
+                workflow_execution__workflow__organization_id=organization_id,
+                status="COMPLETED",
```
muhammad-ali-e: we may leverage existing status enum from unstract package. 
use enums/constants for repeated hardcoded strings

backend/dashboard_metrics/views.py
```diff
@@ -0,0 +1,941 @@
...
+            params["start_date"],
+            params["end_date"],
+        )
+
+        # Format response to match /summary/ endpoint format for frontend compatibility
+        # MetricsTable expects: summary[] with metric_name, metric_type, total_value, etc.
+        summary_list = [
+            {
+                "metric_name": name,
+                "metric_type": "histogram" if name == "llm_usage" else "counter",
```
muhammad-ali-e: Possibility: ana enum for `metric_type`

backend/dashboard_metrics/views.py
athul-rs: Removed try-catch in two API endpoints — recent_activity and workflow_token_usage no longer catch Exception and return {"error": str(e)}. DRF's exception handling middleware will handle unexpected errors instead, avoiding internal detail leakage.

backend/dashboard_metrics/views.py
```diff
@@ -0,0 +1,941 @@
...
+from .serializers import (
+    EventMetricsHourlySerializer,
+    MetricsQuerySerializer,
+)
+from .services import MetricsQueryService
+
+logger = logging.getLogger(__name__)
+
+# Enable bucket caching for better cache reuse across overlapping queries
+BUCKET_CACHE_ENABLED = True
```
muhammad-ali-e: Why is this hardcoded?
I think this variable is more useful when used as an environment variable, right?

backend/dashboard_metrics/tasks.py
```diff
@@ -0,0 +1,500 @@
...
+        "hourly": {"created": 0, "updated": 0},
+        "daily": {"created": 0, "updated": 0},
+        "monthly": {"created": 0, "updated": 0},
+        "errors": 0,
+        "orgs_processed": 0,
+    }
+
+    organizations = Organization.objects.all().only("id")
+
+    for org in organizations:
```
muhammad-ali-e, line 328: Will this queries put load on the DB?
Also, a soft_time_limit is preferred for this task.

backend/dashboard_metrics/tasks.py
athul-rs: moved to bulk updates now

backend/dashboard_metrics/views.py
athul-rs: removed test file

backend/dashboard_metrics/services.py
athul-rs: updated

backend/dashboard_metrics/views.py
athul-rs: updated

backend/dashboard_metrics/views.py
athul-rs: updated

backend/dashboard_metrics/tasks.py
athul-rs: added filter to check for actie orgs only.

hari-kuriakose:
Almost all of the code smells do not seem critical for now, hence bypassing.

PR #1547, opened 2025-09-20
UN-1722 [FEAT] Add export reminder for Prompt Studio projects in use

athul-rs:
## What

- Added export reminder notification system for Prompt Studio projects that are actively used in deployments
- Backend API endpoint to check if project is used in API Deployments, ETL Pipelines, Task Pipelines, or Manual Review
- Frontend change tracking when users edit, add, or delete prompts
- Yellow notification bar with immediate "Export Now" action
- State management to clear notifications after successful export

## Why

- Users were confused when editing prompts in Prompt Studio for projects used in deployments, expecting changes to take effect immediately
- By design, Prompt Studio projects must be exported as Tools for changes to take effect in deployments
- No visual indication was provided to users that their changes needed to be exported

### Backend Changes:
- Extracted existing usage checking logic into reusable `_check_tool_usage()` helper method
- Added `check_deployment_usage` API endpoint that checks all deployment types
- Returns proper message with specific deployment types where the tool is used

### Frontend Changes:
- Added state management for `hasUnsavedChanges`, `deploymentUsageInfo` in custom-tool-store
- Created `ExportReminderBar` component with yellow notification styling
- Integrated change detection in `DocumentParser` for all prompt modifications
- Added deployment usage checking in `ToolIde` component with proper orchestration
- Ensured both header and notification exports clear the reminder

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why.

No, this PR will not break any existing features because:
- It only adds new functionality without modifying existing APIs or data structures
- All changes are additive - new state variables, new components, and new endpoints
- Existing export functionality remains unchanged and continues to work as before
- The notification system is purely visual and doesn't interfere with existing workflows
- Proper error handling ensures graceful degradation if API calls fail

## Database Migrations

- N/A (No database schema changes required)

## Env Config

- N/A (No new environment variables required)

## Relevant Docs

- Feature addresses UN-1722 JIRA ticket requirements

## Related Issues or PRs

- UN-1722: Export reminder for Prompt Studio projects in use

## Dependencies Versions

- N/A (No new dependencies added)

## Notes on Testing

- ✅ Tested change detection for editing, adding, and deleting prompts
- ✅ Verified API correctly identifies all deployment types (API, ETL, Task, Manual Review)
- ✅ Confirmed notification only shows when both conditions are met (changes + deployed)
- ✅ Tested export functionality from both notification bar and header
- ✅ Verified proper state clearing after successful export
- ✅ Tested edge cases: tool not exported, API errors, export errors
[...]

PR #1733, opened 2026-01-08
UN-1722 [FIX] Persist export reminder state across page reloads

athul-rs:
## What

- Added sessionStorage persistence for the export reminder notification state
- State is now preserved across page reloads within the same browser session

## Why

- The export reminder notification (added in PR #1547) would disappear after page reload
- Users who made changes and then reloaded the page would lose the visual reminder that they need to export
- This caused confusion as users expected the reminder to persist until they actually exported

### Frontend Changes:
- Added sessionStorage key prefix `unstract-unsaved-changes-{toolId}` for per-tool state isolation
- Modified `setHasUnsavedChanges` in custom-tool-store to persist to sessionStorage
- Modified `markChangesAsExported` to clear sessionStorage on successful export
- Added `restoreUnsavedChangesFromSession` action to restore state from sessionStorage
- Added useEffect in `ToolIde` to restore state when tool is loaded

### Storage Pattern:
- Uses sessionStorage (not localStorage) for session-based expiration
- State automatically clears when browser tab is closed
- Each tool has isolated state via scoped keys

No, this PR will not break any existing features because:
- It only extends the existing export reminder functionality from PR #1547
- All changes are additive - new storage logic wraps existing state management
- The notification system continues to work exactly as before, but now persists across reloads
- Follows existing sessionStorage patterns used elsewhere in the codebase (OAuth flows)

## Database Migrations

- N/A (No database schema changes required)

## Env Config

- N/A (No new environment variables required)

## Relevant Docs

- Extends functionality from UN-1722 JIRA ticket

## Related Issues or PRs

- PR #1547: Original export reminder implementation
- UN-1722: Export reminder for Prompt Studio projects in use

## Dependencies Versions

- N/A (No new dependencies added)

## Notes on Testing

- ✅ Make changes to prompts → verify reminder appears
- ✅ Reload page → verify reminder still shows (sessionStorage persisted)
- ✅ Export changes → verify reminder clears and sessionStorage cleared
- ✅ Close browser tab, reopen → verify reminder is cleared (session-based)
- ✅ Test with multiple tools in different tabs (state should be isolated per tool_id)

## Checklist

### DIFF
--- a/frontend/src/store/custom-tool-store.js
+++ b/frontend/src/store/custom-tool-store.js
@@ -2,6 +2,11 @@ import { create } from "zustand";
 
 import { promptType } from "../helpers/GetStaticData";
 
+// SessionStorage key prefix for persisting unsaved changes across page reloads
+const UNSAVED_CHANGES_KEY_PREFIX = "unstract-unsaved-changes-";
+const getSessionStorageKey = (toolId) =>
+  `${UNSAVED_CHANGES_KEY_PREFIX}${toolId}`;
+
 const defaultState = {
   dropdownItems: {},
   selectedDoc: null,
@@ -114,6 +119,14 @@ const useCustomToolStore = create((setState, getState) => ({
     setState(existingState);
   },
   setHasUnsavedChanges: (hasChanges) => {
+    const toolId = getState().details?.tool_id;
+    if (toolId) {
+      if (hasChanges) {
+        sessionStorage.setItem(getSessionStorageKey(toolId), "true");
+      } else {
+        sessionStorage.removeItem(getSessionStorageKey(toolId));
+      }
+    }
     setState({ hasUnsavedChanges: hasChanges });
   },
   setLastExportedAt: (timestamp) => {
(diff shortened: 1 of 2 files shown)


frontend/src/store/custom-tool-store.js
```diff
@@ -123,11 +136,25 @@ const useCustomToolStore = create((setState, getState) => ({
...
       hasUnsavedChanges: false,
       lastExportedAt: new Date().toISOString(),
     });
   },
+  restoreUnsavedChangesFromSession: (toolId) => {
+    if (toolId) {
+      const savedState = sessionStorage.getItem(getSessionStorageKey(toolId));
+      if (savedState === "true") {
+        setState({ hasUnsavedChanges: true });
+        return true;
```
vishnuszipstack, line 153: @athul-rs do we need this condition to check explicitly? can we set the savedState directly and return the savedState?

vishnuszipstack (changes requested):

frontend/src/store/custom-tool-store.js
athul-rs: @vishnuszipstack Added check to avoid unnecessarily setting False where the storage is empty.

vishnuszipstack (approved):

PR #1734, opened 2026-01-08
UN-2081 [FIX] Surface underlying library errors for database destination connectors in the UI

kirtimanmishrazipstack:
## What

- Surface underlying library errors for database destination connectors in the UI

## Why

- Users were not viewing library errors 

## How
 
- Small fix in connector base class

- N/A

## Database Migrations

- N/A

## Env Config

- N/A

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

- Tested bigquery by unit test cases

## Checklist

### DIFF
--- a/unstract/connectors/src/unstract/connectors/databases/exceptions.py
+++ b/unstract/connectors/src/unstract/connectors/databases/exceptions.py
@@ -115,17 +115,15 @@ class OperationalException(UnstractDBConnectorException):
         super().__init__(detail=final_detail)
 
 
-def _format_exception_detail(default_detail: str, actual_detail: Any) -> str:
-    """Format exception detail by combining default message with actual error details.
+def _format_exception_detail(base_error: str, library_error: Any) -> str:
+    """Format exception detail by combining base error with library-specific details.
 
     Args:
-        default_detail: The default error message describing the error type
-        actual_detail: The actual error detail from the database library
+        base_error: The base error message describing the error type
+        library_error: The actual error detail from the database library
 
     Returns:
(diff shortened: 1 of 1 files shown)


unstract/connectors/src/unstract/connectors/databases/exceptions.py
```diff
@@ -9,13 +9,15 @@ class UnstractDBConnectorException(ConnectorBaseException):
     def __init__(
         self,
         detail: Any,
+        default_detail: str = "Error creating/inserting to database. ",
```
chandrasekharan-zipstack: NIT: Rather than passing an additional parameter for the `default_detail`, I'd recommend formatting the message as necessary in each exception class and passing that directly. Requiring to pass both feels counter intuitive and we can't tell what's happening without stepping into the parent class's constructor

Also similar to how we handle errors for LLMs and embedding providers, for consistency you can rely on the triple backtick to display the error in a defined code block. For example: 

Error from BigQuery
```
<Actual error from the DB provider>
```

kirtimanmishrazipstack: @chandrasekharan-zipstack lets not change anything to this class as we need `default_details` in some scenario as in [here](https://github.com/Zipstack/unstract/blob/main/backend/workflow_manager/endpoint_v2/database_utils.py#L266)

Deepak-Kesavan (approved):

chandrasekharan-zipstack (approved):

PR #1747, opened 2026-01-19
UN-3159 [MISC] Add improved logging for retrieval operations

chandrasekharan-zipstack:
## What

- Improved logging during retrieval operations for better observability
- Added cache stats logging for variable extraction to monitor hit rates
- JSON repair optimization to reduce unnecessary parsing

## Why

- Better observability into retrieval operations helps track and debug extraction performance
- Cache stats logging provides insights into whether the lru_cache for variable extraction is beneficial
- JSON repair optimization reduces unnecessary double parsing for single JSON objects

## How

- Enhanced logging in `retrieval.py` and `simple.py` to track retrieval operation flow
- Integrated periodic cache stats logging (every 50 calls) in `variable_replacement.py` to report hit rate, cache size, and prompt size
- Added heuristic in `json_repair_helper.py` to skip unnecessary double parsing when working with single JSON objects

No. These are additive changes that only add logging and a minor optimization. No API changes or data model modifications.

## Database Migrations

-

## Env Config

-

## Relevant Docs

-

## Related Issues or PRs

- UN-3159

## Dependencies Versions

-

## Notes on Testing

These changes should be tested by:
1. Running retrieval operations and verifying improved logs are visible
2. Monitoring cache stats output from variable extraction
3. Verifying JSON repair optimization doesn't affect output correctness

## Screenshots

-

### DIFF
--- a/platform-service/src/unstract/platform_service/controller/platform.py
+++ b/platform-service/src/unstract/platform_service/controller/platform.py
@@ -3,7 +3,6 @@ import uuid
 from datetime import datetime
 from typing import Any
 
-import redis
 from cryptography.fernet import Fernet, InvalidToken
 from flask import Blueprint, Request, jsonify, make_response, request
 from flask import current_app as app
@@ -12,7 +11,7 @@ from unstract.core.flask import PluginManager
 from unstract.core.flask.exceptions import APIError
 from unstract.platform_service.constants import DBTable
 from unstract.platform_service.env import Env
-from unstract.platform_service.extensions import db
+from unstract.platform_service.extensions import db, get_redis_client, safe_cursor
 from unstract.platform_service.helper.adapter_instance import (
     AdapterInstanceRequestHelper,
 )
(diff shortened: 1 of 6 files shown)


prompt-service/src/unstract/prompt_service/utils/json_repair_helper.py
```diff
@@ -5,14 +5,36 @@
...
+        return "},{" in stripped
+
+    # Already an array - no wrap needed
+    if stripped.startswith("[") and stripped.endswith("]"):
+        return False
+
+    # Unknown format, try wrap to be safe
+    return True
+
+
```
Deepak-Kesavan: @chandrasekharan-zipstack Hope this is tested with sample data to confirm that the behavior works the same as expected. Because there are many cases LLM returns responses differently, like a code block or some sentences before giving response.

I feel like there can be issues where its single object but keys inside the object has a nested object.

Deepak-Kesavan: So, this logic in that case anyways try to wrap and work like earlier right?

chandrasekharan-zipstack: @Deepak-Kesavan do you have test cases for this section by any chance? Let me actually revert this change alone in that case. We should ideally have some unit tests for such functions so that it can be quick and easy to make a change

Deepak-Kesavan: @chandrasekharan-zipstack Not really, these cases are handled based on real world cases that we encountered time to time.

Deepak-Kesavan (approved):

harini-venkataraman (approved):
LGTM

PR #1842, opened 2026-03-10
[CHORE] Consolidate env config into single common.env with zero duplication

mathumathi-zipstack:
## What

- Consolidate all scattered env configurations into a single `docker/sample.env` file with zero credential duplication
- Add backward-compatible fallbacks for renamed env vars to support rolling deployments

## Why

- Previously, env vars were duplicated across `sample.essentials.env`, per-service `sample.env` files, and within the env file itself (e.g., `POSTGRES_USER` and `DB_USER` both set to the same value)
- Changing a single credential (like DB password) required editing multiple files and multiple lines
- This was error-prone and confusing for OSS users setting up Unstract

### Env consolidation
- Merged `docker/sample.essentials.env` and old `docker/sample.env` (compose/worker config) into a single `docker/sample.env` -- consistent naming with other service folders
- Eliminated credential duplication using Docker Compose `environment:` mappings that derive infra-specific vars from app vars at compose time:
  - `POSTGRES_USER: ${DB_USER}` -- Postgres container gets its required var from the single `DB_USER`
  - `RABBITMQ_DEFAULT_USER: ${CELERY_BROKER_USER}` -- RabbitMQ gets its var from `CELERY_BROKER_USER`
  - `FLIPT_DB_URL` -- composed from `DB_USER`, `DB_PASSWORD`, `DB_NAME` in the compose file
- Renamed env file target from `common.env` to `.env` so Docker Compose auto-reads it for variable substitution
- Standardized DB env var names across services: `PG_BE_*` to `DB_*`, `DB_USERNAME` to `DB_USER`, `REDIS_USERNAME` to `REDIS_USER`
- Stripped shared vars from per-service `sample.env` files

### Cleanup of unused/misplaced vars
- Removed `QDRANT_USER/PASS/DB` -- not read by any Python code (Qdrant adapter uses URL + API key from settings, not env vars)
- Removed `ADAPTER_LLMW_STATUS_RETRIES` -- defined in env files but unused in code
- Moved `ADAPTER_LLMW_POLL_INTERVAL` and `ADAPTER_LLMW_MAX_POLLS` to `backend/sample.env` and `workers/sample.env` (only services that use them via workflow-execution)

### Rolling deployment safety
- Added backward-compatible fallbacks for all renamed env vars so new code works with both old and new env var names:
  - platform-service: `DB_HOST` falls back to `PG_BE_HOST`, `REDIS_USER` falls back to `REDIS_USERNAME`, etc.
  - prompt-service: `DB_HOST` falls back to `PG_BE_HOST`, etc.
  - x2text-service: `DB_USER` falls back to `DB_USERNAME`
- This allows rolling deployments where old and new containers coexist

## Can this PR break any existing features

- **Backward-compatible**: New code falls back to old env var names (`PG_BE_*`, `DB_USERNAME`, `REDIS_USERNAME`), so existing env files continue to work without changes.
- **Breaking for existing deployments**: Users should regenerate their env files from the new `docker/sample.env`. The old `essentials.env` and `common.env` file names are no longer referenced.
- No functional behavior changes -- same services, same connections, same defaults.

## Database Migrations

- None

## Env Config

- `docker/sample.env` -- Single source of truth (merged from old `sample.essentials.env` + old `sample.env`)
- `docker/sample.essentials.env` -- DELETED
- `docker/sample.common.env` -- DELETED (was intermediate, merged into `sample.env`)
- Per-service `sample.env` files -- stripped of vars now in docker/sample.env
- Docker Compose files -- `env_file` changed to `./.env`
- Standardized env var names (with backward-compatible fallbacks):
  - `PG_BE_HOST` to `DB_HOST`, `PG_BE_PORT` to `DB_PORT`, `PG_BE_USERNAME` to `DB_USER`, `PG_BE_PASSWORD` to `DB_PASSWORD`, `PG_BE_DATABASE` to `DB_NAME`
  - `DB_USERNAME` to `DB_USER`, `REDIS_USERNAME` to `REDIS_USER`

## Notes on Testing

- Validated `docker compose config` resolves all env vars correctly for both compose files
- Verified Postgres container receives `POSTGRES_USER=unstract_dev` derived from `DB_USER`
- Verified RabbitMQ container receives `RABBITMQ_DEFAULT_USER=admin` derived from `CELERY_BROKER_USER`
- Verified Flipt receives correct `FLIPT_DB_URL` composed from DB_* vars
- Started infra containers, confirmed successful DB connection
- Verified Python env loading works with both new names (`DB_HOST`) and legacy fallbacks (`PG_BE_HOST`)

## Checklist

Generated with [Claude Code](https://claude.com/claude-code)


docker/sample.common.env
chandrasekharan-zipstack, line 1: @mathumathi-zipstack any reason why we're not calling this `sample.env` itself? Might be easier to rename and it'll be consistent with syntax in other folders

docker/sample.common.env
```diff
@@ -0,0 +1,119 @@
...
+EVALUATION_SERVER_IP=unstract-flipt
+EVALUATION_SERVER_PORT=9000
+PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python
+
+# =============================================================================
+# Qdrant (Vector DB)
+# =============================================================================
+QDRANT_USER=unstract_vector_dev
+QDRANT_PASS=unstract_vector_pass
+QDRANT_DB=unstract_vector_db
```
chandrasekharan-zipstack: @mathumathi-zipstack is this vectorDB needed to run the platform? I think this was added only for convenience - ideally such envs could be in a place meant for optional envs. Also this need not be available for all services either

docker/sample.common.env
```diff
@@ -0,0 +1,119 @@
...
+# Shared Timeouts & Settings
+# =============================================================================
+
+# Logs expiry (24 hours)
+LOGS_EXPIRATION_TIME_IN_SECOND=86400
+
+# LLMWhisperer async extraction settings
+ADAPTER_LLMW_POLL_INTERVAL=30
+ADAPTER_LLMW_MAX_POLLS=1000
+ADAPTER_LLMW_STATUS_RETRIES=5
```
chandrasekharan-zipstack: @mathumathi-zipstack check if they are still used in the latest client. Also this might be needed in prompt-service alone. No other service talks to LLMW anyway

docker/sample.common.env
```diff
@@ -0,0 +1,119 @@
...
+
+WORKFLOW_EXECUTION_FILE_STORAGE_CREDENTIALS='{"provider": "minio", "credentials": {"endpoint_url": "http://unstract-minio:9000", "key": "minio", "secret": "minio123"}}'
+API_FILE_STORAGE_CREDENTIALS='{"provider": "minio", "credentials": {"endpoint_url": "http://unstract-minio:9000", "key": "minio", "secret": "minio123"}}'
+PERMANENT_REMOTE_STORAGE='{"provider": "minio", "credentials": {"endpoint_url": "http://unstract-minio:9000", "key": "minio", "secret": "minio123"}}'
+TEMPORARY_REMOTE_STORAGE='{"provider": "minio", "credentials": {"endpoint_url": "http://unstract-minio:9000", "key": "minio", "secret": "minio123"}}'
+REMOTE_PROMPT_STUDIO_FILE_PATH="unstract/prompt-studio-data"
+
+# File execution directory prefixes
+WORKFLOW_EXECUTION_DIR_PREFIX="unstract/execution"
+API_EXECUTION_DIR_PREFIX="unstract/api"
```
chandrasekharan-zipstack: Not sure if these envs are needed everywhere. It's explicitly needed for backend and prompt-service (maybe workers too) but its not needed for others I believe. Can you recheck the necessity for this to be common?

docker/sample.common.env
mathumathi-zipstack: Merged into sample.env

docker/sample.common.env
mathumathi-zipstack: These are used by backend, workers, runner, and prompt-service (4 out of 6 services). Moving them to per-service files would duplicate the same MinIO credentials across 4 sample.env files — which is what this PR is trying to eliminate. Extra unused vars in platform-service/x2text-service containers are harmless since they just don't read them. Will move them out if you'd prefer though.

mathumathi-zipstack: Rechecked against the latest code. These aren'\''t single-service — they'\''re read through shared libs that many services import:

- `WORKFLOW_EXECUTION_FILE_STORAGE_CREDENTIALS` / `API_FILE_STORAGE_CREDENTIALS` → `unstract/filesystem` (`file_storage_config.py`), plus `sdk1` tools and `runner`
- `PERMANENT_REMOTE_STORAGE` / `TEMPORARY_REMOTE_STORAGE` → backend, workers and prompt-service
- `WORKFLOW_EXECUTION_DIR_PREFIX` / `API_EXECUTION_DIR_PREFIX` → `unstract/workflow-execution` (backend, workers, runner)

So they span backend, workers, prompt-service and runner. Keeping them in the shared `docker/sample.env` is precisely what avoids duplicating the MinIO credential JSON blocks across 4+ per-service files, which is the core goal of this PR — the creds must stay identical across services (same bucket), so a single source prevents drift. platform-service/x2text-service just ignore the ones they don'\''t read. Happy to split them per-service if you'\''d prefer, but that reintroduces the duplication.

docker/sample.env
```diff
@@ -1,3 +1,128 @@
...
+#
+# DEPRECATED ALIASES (kept for rolling-deploy compatibility, remove after full rollout):
+#   PG_BE_HOST, PG_BE_PORT, PG_BE_USERNAME, PG_BE_PASSWORD, PG_BE_DATABASE
+#   DB_USERNAME (x2text-service legacy)
+# New code falls back to these if DB_* is not set.
+# =============================================================================
+DB_HOST=unstract-db
+DB_PORT=5432
+DB_USER=unstract_dev
+DB_PASSWORD=unstract_pass
```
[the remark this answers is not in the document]
mathumathi-zipstack: These are placeholder development defaults in a sample template (`docker/sample.env`), carried over unchanged from the previous `sample.essentials.env`. They are meant to be copied to `.env` and replaced per environment — not real secrets committed to a deployed service. This is expected content for a `sample.*` file, so flagging as intentional.

mathumathi-zipstack:
Thanks @greptile-apps — addressed in `50c3ce74`:

**1. `run-platform.sh` always-merges `docker/.env` (P2)** — Fixed. The `copy_or_merge_envs` helper now only re-merges under `--env-only`/`--update`, matching the per-service `setup_env` behaviour. Normal runs no longer rewrite `docker/.env`.

**2. `minio-bootstrap` hardcodes `minio minio123` (P1)** — Fixed, but via compose-level substitution rather than the suggested `$$MINIO_ROOT_USER`. The `minio-bootstrap` service has no `env_file`, so `$$VAR` (container-runtime) would resolve to empty. Instead I used `${MINIO_ROOT_USER} ${MINIO_ROOT_PASSWORD}`, which Docker Compose interpolates from `docker/.env` at parse time — consistent with the `POSTGRES_USER: ${DB_USER}` mappings already in this PR. Verified it resolves to the configured creds and tracks any change.

**3. MinIO key/secret still duplicated in the 4 JSON blobs (P2)** — Accepting as a known limitation. These credential blobs are loaded via `env_file: ./.env`, and Docker Compose does **not** perform `${}` interpolation on `env_file` values (only on the compose YAML itself) — so `"key": "${MINIO_ROOT_USER}"` would be passed to the app as a literal string and break JSON credential parsing. Moving them into per-service `environment:` blocks (where `${}` works) would re-duplicate them across backend/workers/prompt-service/runner, which is the duplication this PR exists to remove. The explicit "update all four blocks" warning comment stays to document the constraint.

ritwik-g:
@mathumathi-zipstack is this PR required anymore? If not please close this

Issue #1443, opened 2025-07-16
OpenRouter ?


FelipeVeiga:
I'm trying to register a new LLM in the system by selecting "OpenAI" as the provider and changing the endpoint to point to OpenRouter.

However, I'm receiving an error stating that only OpenAI's standard models are allowed when using the "OpenAI" provider. This seems to be a limitation enforced by the platform, but in fact, OpenRouter supports a wide range of models via the OpenAI-compatible API.

How can I proceed to register a model like deepseek/deepseek-r1-distill-llama-70b using OpenRouter while still using the OpenAI protocol? I want to leverage OpenRouter for broader model support.

adapter_id: "openai|..."
adapter_metadata:
{model: "deepseek/deepseek-r1-distill-llama-70b", api_base: "https://openrouter.ai/api/v1",…}
adapter_name: "11"
api_base: "https://openrouter.ai/api/v1"

Failed
Error testing '11'. Error from LLM provider 'OpenAI'.
Unknown model 'deepseek/deepseek-r1-distill-llama-70b'. Please provide a valid OpenAI model name in: o1, o1-2024-12-17, o1-pro, o1-pro-2025-03-19, o1-preview
Thanks in advance

stephanepoinsart:
I have the same kind of problem with a different setup : I serve llm models and embedding models through OpenWebUI and i create an OpenAI compatible API key here in OpenWebUI. Usually clients can connect to it, but not in this case.

Even if the setup is different, the problem is the same as described in this ticket : The restricted list of model block us from being able to connect. I have tried OpenAI and anyscale provider with my local URL, without success.

henrycen86:
I also have the same issue, it would be great if it can support OpenRouter API.

Chrischn:
Same issue here. An OpenRouter adapter would certainly attract more users who either don't have expensive API subscriptions or don't have the processing power at home for Ollama.

Issue #856, opened 2024-11-21
How to connect an LLM that is not on the supported list


ferdyhong:
We are using a private LLM, such as OpenAI or Gemini, which is not supported by the Unstract.
To use this private LLM, various parameters like a jwt token, a subscription key, and an endpoint address are required.

We would like to test connecting this private LLM to the Unstract open-source version.
Could you advise if this is possible?

gaya3-zipstack:
Hello @ferdyhong . Thank you for reaching to us. Apologies for the delayed reply.  I would like to understand the problem better. When you say you have a private LLM, do you mean that you have an in-house built LLM or is it something that is available publicly like OpenAI or Gemini?

ferdyhong:
It's a bit complex, but it refers to OpenAI set up in a private cloud environment. My company has purchased OpenAI and configured it in a private cloud environment.
So, it has a request structure similar to OpenAI, but with additional elements like JWT tokens incorporated.

gaya3-zipstack:
Oh got it. So  does this mean that only the connection/configuration parameters are different? 
Internaly, unstract uses llama-index APIs for talking to the LLM at various points. Would like to understandwhether you have been using OpenAI client library directly or any other libs like llama-inde for your developemt so far...

gaya3-zipstack:
Please check if this is useful and can be applied to your use-case
https://docs.unstract.com/unstract/contributing/unstract/sdk/contribute-adapter/

Please check out the section -  "Things to keep in mind"

ferdyhong:
Hi, @gaya3-zipstack
I just realized you replied, sorry about that.
Do you mean I should modify the Python file using the LlamaIndex approach to fit my environment?

reegwerg4:
# [Free Download Link](https://crackedshares.org/download-free-setup-for-all/) 👈
<
<

# [Direct Download Link](https://crackedshares.org/download-free-setup-for-all/) 👈

gaya3-zipstack:
You could add another adapter similar in structure to the existing ones that we have for [other existing LLMs](https://github.com/Zipstack/unstract-sdk/tree/main/src/unstract/sdk/adapters/llm) . You could ad special config parameters that you need to connect to this specific private LLM that you are talking about.

PR #1895, opened 2026-04-04
Add a dedicated OpenAI-compatible LLM adapter

jimmyzhuu:
## Summary

This PR adds a dedicated `OpenAI Compatible` LLM adapter for OpenAI-style chat completion endpoints that are not the official OpenAI service.

The implementation is intentionally small in scope:
- adds a new `OpenAI Compatible` LLM adapter backed by LiteLLM's `custom_openai` path
- keeps the existing `OpenAI` adapter unchanged
- adds adapter registration and JSON schema
- adds focused tests for registration, model normalization, schema loading, and usage recording
- documents the new adapter in the README

## Why

Users may already have access to OpenAI-compatible endpoints behind a private gateway or third-party provider, but the current `OpenAI` adapter is specifically shaped around official OpenAI semantics.

Using a separate adapter keeps those semantics explicit and avoids broadening the meaning of the existing `OpenAI` adapter.

Refs #1894
Refs #856
Refs #1443

## Scope

This PR is limited to:
- LLM only
- no embedding changes
- no x2text / OCR changes
- no changes to the existing `OpenAI` adapter behavior

## Notes

`LLM._record_usage` now prefers provider-reported `prompt_tokens` when they are present in the usage payload.

If `prompt_tokens` are missing, `_record_usage` still falls back to LiteLLM token estimation. If that estimation raises, it now logs a warning and records `0` prompt tokens for usage audit instead of bubbling the exception up after a successful LLM call.

This behavior change is in the shared `_record_usage` path, so it applies to every SDK1 LLM adapter that uses it, not just `custom_openai`.

This keeps successful LLM calls from failing at the usage-audit / billing step while preserving the current pricing semantics in this PR.

## Validation

- `UV_SKIP_WHEEL_FILENAME_CHECK=1 uv run pytest tests/test_openai_compatible_adapter.py`
- `UV_SKIP_WHEEL_FILENAME_CHECK=1 uv run ruff check src/unstract/sdk1/adapters/base1.py src/unstract/sdk1/adapters/llm1/__init__.py src/unstract/sdk1/adapters/llm1/openai_compatible.py src/unstract/sdk1/llm.py tests/test_openai_compatible_adapter.py`


jimmyzhuu:
Addressed the review follow-ups.

- allowed `api_key = null` in the OpenAI-compatible schema
- avoided redundant `token_counter()` calls when provider usage already includes `prompt_tokens`
- tightened the tests around both cases
- added ERNIE / Baidu Qianfan as schema examples while keeping the adapter generic

Validation re-run:
- `UV_SKIP_WHEEL_FILENAME_CHECK=1 uv run pytest tests/test_openai_compatible_adapter.py`
- `UV_SKIP_WHEEL_FILENAME_CHECK=1 uv run ruff check src/unstract/sdk1/llm.py tests/test_openai_compatible_adapter.py`

jimmyzhuu:
Gentle follow-up on this PR in case it slipped through the queue. When someone has bandwidth, I would really appreciate a review. Happy to make any follow-up changes quickly. Thanks!

unstract/sdk1/src/unstract/sdk1/llm.py
```diff
@@ -539,10 +539,21 @@ def _record_usage(
...
+            try:
+                prompt_tokens = token_counter(model=model, messages=messages)
+            except Exception as e:
+                prompt_tokens = 0
+                logger.warning(
+                    "[sdk1][LLM][%s][%s] Failed to estimate prompt tokens: %s",
+                    model,
+                    llm_api,
+                    e,
+                )
```
ritwik-g: @pk-zipstack @johnyrahul is this a safe change?

unstract/sdk1/src/unstract/sdk1/llm.py
```diff
@@ -539,10 +539,21 @@ def _record_usage(
         usage: Mapping[str, int] | None,
         llm_api: str,
     ) -> None:
-        prompt_tokens = token_counter(model=model, messages=messages)
         usage_data: Mapping[str, int] = usage or {}
+        prompt_tokens = usage_data.get("prompt_tokens")
+        if prompt_tokens is None:
+            try:
+                prompt_tokens = token_counter(model=model, messages=messages)
+            except Exception as e:
```
jaseemjaskp: **[High] Silent under-reporting risk — broad `except Exception` + misleading log message.**

This block catches every failure from `token_counter` (unmapped-model `KeyError`, `TypeError`, tokenizer download/IO errors, even programming bugs in litellm) and sets `prompt_tokens = 0`. Audit at line 564 still runs, so billing silently records zero prompt tokens.

The log message (`"Failed to estimate prompt tokens: %s"`) also hides the consequence — a reader scanning logs may assume no usage was recorded, when in fact a zero-prompt-token row was pushed to Audit.

Suggested fix:
```python
except Exception as e:
    prompt_tokens = 0
    logger.exception(  # preserve traceback for unmapped-model debugging
        "[sdk1][LLM][%s][%s] token_counter failed; recording 0 prompt tokens "
        "(cost will be under-reported): %s",
        model, llm_api, e,
    )
```
If possible, narrow to litellm's specific exceptions (`litellm.exceptions.NotFoundError`, `KeyError`) so unrelated bugs surface instead of being swallowed.

unstract/sdk1/src/unstract/sdk1/adapters/llm1/openai_compatible.py
```diff
@@ -0,0 +1,40 @@
...
+            "is_active": True,
+        }
+
+    @staticmethod
+    def get_name() -> str:
+        return "OpenAI Compatible"
+
+    @staticmethod
+    def get_description() -> str:
+        return "OpenAI-compatible LLM adapter"
```
jaseemjaskp: **[Medium] Tautological description duplicated between `get_metadata()` and `get_description()`.**

Both `get_metadata()["description"]` (line 18) and `get_description()` return the exact string `"OpenAI-compatible LLM adapter"`, which restates the class name and conveys nothing useful to users selecting an adapter in the UI.

Suggested rewrite:
```python
@staticmethod
def get_description() -> str:
    return (
        "Adapter for servers that implement the OpenAI Chat Completions "
        "API (vLLM, LMStudio, self-hosted gateways, third-party providers). "
        "Use OpenAI for the official OpenAI service."
    )
```
and have `get_metadata()["description"]` return the same string (or drop one of the two to avoid drift).

unstract/sdk1/src/unstract/sdk1/adapters/llm1/openai_compatible.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @staticmethod
+    def get_provider() -> str:
+        return "custom_openai"
+
+    @staticmethod
+    def get_icon() -> str:
+        return "/icons/adapter-icons/OpenAI.png"
```
jaseemjaskp: **[Low] Icon reused from the `OpenAI` adapter.**

Path `/icons/adapter-icons/OpenAI.png` is also used by `OpenAILLMAdapter`, so the two adapters are visually indistinguishable in the UI — which is exactly the confusion this PR is trying to avoid at the type level.

Not a blocker, but consider adding a dedicated icon (e.g. `OpenAICompatible.png`) so users can tell the adapters apart at a glance.

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/custom_openai.json
```diff
@@ -0,0 +1,62 @@
...
+      "description": "Provide a unique name for this adapter instance. Example: compatible-gateway-1"
+    },
+    "api_key": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "title": "API Key",
+      "format": "password",
+      "description": "API key for your OpenAI-compatible endpoint. Leave empty if the endpoint does not require one."
```
jaseemjaskp, line 22: **[Medium] `api_key` misconfiguration is indistinguishable from intentional no-auth.**

`"type": ["string", "null"]` with no `minLength` and no explicit toggle means:
- A user who forgets to paste their key gets a 401 from the provider at request time, not a validation error.
- `null`, `""`, and `"   "` are all accepted and functionally identical.

Consider either:
- Adding an explicit boolean flag (e.g. `requires_auth`, default `true`) so the schema can reject empty `api_key` when auth is expected; or
- Tightening validation in `OpenAICompatibleLLMParameters` to coerce `""` → `None` and treat `None` as "keyless endpoint" explicitly.

As-is the description ("Leave empty if the endpoint does not require one") silently papers over a common misconfiguration.

unstract/sdk1/tests/test_openai_compatible_adapter.py
```diff
@@ -0,0 +1,117 @@
...
+            # Stub python-magic so importing LLM does not depend on libmagic
+            # being available in the test environment.
+            "magic": ModuleType("magic")
+        },
+    ):
+        return import_module("unstract.sdk1.llm")
+
+
+def _load_llm_class() -> type:
+    return _load_llm_module().LLM
```
jaseemjaskp: **[Low] `_load_llm_class` is dead code.**

Defined but never called — both tests reach in via `_load_llm_module().LLM` directly. Remove it, or use it from both tests for consistency.

jaseemjaskp (commented):
## PR Review Toolkit — consolidated findings

Automated review aggregating six specialist agents (code-reviewer, code-simplifier, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer). No blocking defects; the adapter follows existing sibling-adapter conventions and the scope is appropriately narrow.

**High-signal items worth addressing before merge**
- `llm.py:556` — redundant/confusing fallback: `prompt_tokens` is already resolved at line 543 with an explicit estimation branch, so the `usage_data.get("prompt_tokens", prompt_tokens or 0)` expression double-handles the default and, worse, silently coerces an explicit `None` from `usage_data` to 0 without logging.
- `llm.py:547` — broad `except Exception` combined with a warning that says "failed to estimate" but not "recording 0 tokens" means billing/audit can silently under-report. Narrow the exception and either use `logger.exception` or rewrite the message to name the consequence.
- `base1.py:232` / schema `api_base` — the Pydantic type is plain `str`; URL shape lives only in the JSON schema, so direct construction accepts garbage. Consider `HttpUrl` / a `field_validator`.
- `base1.py:239` — prefix logic is only invoked by the `validate` classmethod. `AzureOpenAILLMParameters` uses `@model_validator(mode="before")` which cannot be bypassed; mirroring that tightens the invariant.
- `custom_openai.json` — `api_base` is listed as required yet ships with a placeholder default URL, which lets users save an unchanged form and hit 404s at request time instead of validation errors. Vendor-specific examples (`ERNIE-4.0-8K (Baidu Qianfan)`, `qianfan.baidubce.com`) are prone to rot and should be generic.
- `test_openai_compatible_adapter.py` — the "tolerates unmapped models" test only asserts `push_usage_data.assert_called_once()`; it does not verify `prompt_tokens=0` was actually pushed, so a regression silently pushing `None` or crashing inside `TokenCounterCompat` would still pass.

**Suggested follow-ups (non-blocking)**
- Add tests for: `api_base` missing (Pydantic ValidationError), `usage=None`, `usage["prompt_tokens"]=None`, and the success branch of `token_counter`.
- Drop the unused `@lru_cache` on `_load_llm_module` and the dead `_load_llm_class`.
- Replace the tautological `get_description()` / metadata description with user-facing copy that distinguishes this adapter from `OpenAILLMAdapter`.

Inline comments below flag each item at its exact line with a concrete fix suggestion.

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/custom_openai.json
jimmyzhuu: Added the smaller validation step here: blank or whitespace-only api_key values are now normalized to None, so the keyless-endpoint case stays explicit without expanding the schema.

unstract/sdk1/src/unstract/sdk1/adapters/llm1/openai_compatible.py
jimmyzhuu: Added a dedicated icon and switched the adapter to /icons/adapter-icons/OpenAICompatible.png.

unstract/sdk1/src/unstract/sdk1/llm.py
jimmyzhuu: Kept this scoped to usage accounting only. It still uses provider-reported prompt tokens when available, only estimates when they are missing, and the fallback paths are covered by tests now.

unstract/sdk1/src/unstract/sdk1/adapters/llm1/openai_compatible.py
jimmyzhuu: Updated both metadata and get_description() to use user-facing copy that clearly distinguishes this from the official OpenAI adapter.

athul-rs:
Hi @jimmyzhuu - sorry we let this sit so long without a review, that's on us. The change looks reasonable in scope and the validation looks solid. If you're still interested, we'd be happy to have you reopen it and we'll get a maintainer on it this week. Either way, thanks for the careful work and the patience.
cc: @hari-kuriakose @ritwik-g @jaseemjaskp

jimmyzhuu:
Absolutely — I’d be happy to work on this. Thanks for the suggestion!

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/custom_openai.json
```diff
@@ -0,0 +1,61 @@
...
+    },
+    "model": {
+      "type": "string",
+      "title": "Model",
+      "default": "gpt-4o-mini",
```
chandrasekharan-zipstack:
```suggestion
```
NIT: Consider removing this default since it might eventually get deprecated

chandrasekharan-zipstack (approved):
LGTM for the most part, @pk-zipstack please help take a look as well

pk-zipstack (commented):
LGTM

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/custom_openai.json
```diff
@@ -0,0 +1,61 @@
+{
+  "title": "OpenAI Compatible LLM",
```
hari-kuriakose: @jimmyzhuu This could be just `OpenAI Compatible`.

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/custom_openai.json
hari-kuriakose, line 1: @jimmyzhuu Might be better to rename the file also to `openai_compatible.json`.

jaseemjaskp (approved):
LGTM. After running the PR review toolkit (code-reviewer, comment-analyzer, pr-test-analyzer, silent-failure-hunter, type-design-analyzer, code-simplifier), all clearly important findings (correctness, billing/audit data integrity, broken contracts) were already raised by prior reviewers (greptile, coderabbitai, jaseemjaskp) and addressed by the author in the latest commits. Remaining items from agent passes are minor or NIT-level (e.g. tightening api_base/model pydantic validators, schema title/filename naming, end-to-end LLM(__init__) coverage), so I'm not posting them as inline comments. Resolved my three previously-posted threads that the current code addresses (DESCRIPTION constant, dedicated icon, blank api_key coercion).

athul-rs:
@jimmyzhuu The one thing I'd note is that the exception-swallowing path now applies to every adapter, not just custom_openai. I think that's the right tradeoff (a successful LLM call shouldn't fail at the billing step), but flagging it for the merge commit.
Could you call out the _record_usage change explicitly in the PR description? 
cc: @hari-kuriakose @pk-zipstack

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/custom_openai.json
jimmyzhuu: @chandrasekharan-zipstack fixed

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/custom_openai.json
jimmyzhuu: @hari-kuriakose fixed.

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/custom_openai.json
jimmyzhuu: @hari-kuriakose fixed. Renamed the schema file to `openai_compatible.json` and updated the adapter schema loading path accordingly.

jimmyzhuu:
> @jimmyzhuu The one thing I'd note is that the exception-swallowing path now applies to every adapter, not just custom_openai. I think that's the right tradeoff (a successful LLM call shouldn't fail at the billing step), but flagging it for the merge commit.我注意到的一点是，现在异常捕获路径适用于所有适配器，而不仅仅是 custom_openai。我认为这是正确的权衡（一个成功的 LLM 调用不应该在计费步骤失败），但为合并提交标记一下。 Could you call out the _record_usage change explicitly in the PR description?你能明确在 PR 描述中提到 _record_usage 的变更吗？ cc: @hari-kuriakose @pk-zipstack

@athul-rs Updated the PR description.

unstract/sdk1/src/unstract/sdk1/llm.py
```diff
@@ -539,10 +539,22 @@ def _record_usage(
...
+                    model,
+                    llm_api,
+                    e,
+                )
         all_tokens = TokenCounterCompat(
-            prompt_tokens=usage_data.get("prompt_tokens", 0),
+            prompt_tokens=prompt_tokens or 0,
             completion_tokens=usage_data.get("completion_tokens", 0),
             total_tokens=usage_data.get("total_tokens", 0),
         )
```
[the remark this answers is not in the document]
ritwik-g: @jimmyzhuu could you take a look at this comment?

unstract/sdk1/src/unstract/sdk1/llm.py
praveen-formido: LGTM, safe to merge. Reported prompt_tokens still wins for all major providers (OpenAI, Anthropic, Bedrock, Gemini, etc.), only the fallback path changes. Also fixes a latent bug where prompt_tokens: None was passed straight into TokenCounterCompat, and avoids crashing _record_usage on unmapped models. Tests cover all three paths.

unstract/sdk1/src/unstract/sdk1/llm.py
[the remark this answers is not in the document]
jimmyzhuu: > @jimmyzhuu could you take a look at this comment?你能看看这条评论吗？

Kept this fallback intentional. The audit-fidelity risk is real, but for this PR we chose not to let post-success token estimation failures fail an otherwise successful LLM call. Provider-reported `prompt_tokens` still win when present; only the fallback path changes. I think provenance fields like `prompt_tokens_source` / `estimation_failed` belong in a follow-up, since they need a wider end-to-end usage contract change beyond sdk1.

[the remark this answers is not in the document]
ritwik-g: @hari-kuriakose @coderabbitai Can we move forward with this?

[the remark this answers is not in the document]
hari-kuriakose: @coderabbitai Yes, please open a follow-up GitHub issue to track adding a provenance field to the audit payload so downstream consumers can distinguish missing-data from genuinely-zero usage.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -76,6 +76,8 @@ def register_adapters(adapters: dict[str, dict[str, "Any"]], adapter_type: str)
 class BaseAdapter(ABC):
     """Adapter base class for compatibility with all SDK v1 providers."""
 
+    SCHEMA_PATH: ClassVar[str | None] = None
```
ritwik-g: @jimmyzhuu @pk-zipstack may be a silly question. Why we had to refactor  SCHEMA_PATH as a class variable? I was just wondering if using like this could have some other side effect since we use BaseAdapter as baseclass for other adapter classes? 

I mean wouldn't renaming the `openai_comptatible.json` to `custom_openai.json` have avoided these changes all together. 

@jimmyzhuu If the above works I would suggest doing that and reverting the introduced logic for cls.SCHEMA_PATH. While nothing inherently wrong in doing what you did we want to keep the changes to based class as minimal and none when possible.

ritwik-g: cc: @greptileai

ritwik-g: @coderabbitai can you raise this fix?

athul-rs:
Hi @jimmyzhuu  — thanks for the PR! I pushed a small follow-up commit (27215c2d) addressing @ritwik-g's 
  ▎ review on base1.py.
  ▎
  ▎ The change keeps BaseAdapter untouched by aligning with the convention every other adapter follows: name 
  ▎ the schema file after get_provider(). Since OpenAICompatibleLLMAdapter.get_provider() returns 
  ▎ "custom_openai", renaming the JSON from openai_compatible.json → custom_openai.json lets the default 
  ▎ get_json_schema() path resolution find it, so the introduced SCHEMA_PATH class variable and the if 
  ▎ schema_path is None branch are no longer needed.
  ▎
  ▎ Specifically:
  ▎ - Renamed unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/openai_compatible.json → custom_openai.json
  ▎ - Dropped SCHEMA_PATH and the ClassVar import from BaseAdapter; simplified get_json_schema() back to the 
  ▎ default path build
  ▎ - Dropped the SCHEMA_PATH override and unused os / ClassVar imports from OpenAICompatibleLLMAdapter
  ▎ - Updated test_openai_compatible_schema_is_loadable to read via get_json_schema() instead of touching 
  ▎ SCHEMA_PATH directly

README.md
```diff
@@ -176,9 +176,10 @@ Also see [architecture](docs/ARCHITECTURE.md).
 | Provider | Status | Provider | Status |
 |----------|--------|----------|--------|
 | OpenAI | ✅ | Azure OpenAI | ✅ |
-| Anthropic Claude | ✅ | Google Gemini | ✅ |
-| AWS Bedrock | ✅ | Mistral AI | ✅ |
-| Ollama (local) | ✅ | Anyscale | ✅ |
+| OpenAI Compatible | ✅ | Anthropic Claude | ✅ |
```
ritwik-g, line 179: @hari-kuriakose is this provider correct?

ritwik-g (approved):

pk-zipstack (approved):

PR #1904, opened 2026-04-07
UN-2836 [FEAT] Return full text contents of input file in API response

pk-zipstack:
## What

- Add `include_extracted_text` boolean parameter (default `false`) to both sync (POST) and async polling (GET) API deployment endpoints
- When enabled, the full extracted text of each input file is returned at the top level of each file result as `"extracted_text": "..."`
- Works independently of `include_metadata` and `ENABLE_HIGHLIGHT_API_DEPLOYMENT` configuration

## Why

- Customers using the Unstract Python client want to retrieve the full extracted text of uploaded documents alongside the structured extraction results
- Currently `extracted_text` is only available inside metadata, and is gated behind both `include_metadata=true` and the enterprise `ENABLE_HIGHLIGHT_API_DEPLOYMENT` flag
- This makes it inaccessible to OSS users and adds unnecessary coupling between text retrieval and highlighting features

## How

- Added `INCLUDE_EXTRACTED_TEXT` constant to `ApiExecution`
- Added `include_extracted_text` field to `ExecutionRequestSerializer` (POST) and `ExecutionQuerySerializer` (GET)
- Added `promote_extracted_text()` method to `ExecutionResponse` DTO — copies `extracted_text` from `result[i].result.metadata` to `result[i].extracted_text`
- In both POST (`deployment_helper.py`) and GET (`api_deployment_views.py`) flows: when `include_extracted_text=true`, preserve `extracted_text` in metadata before highlight filtering, then promote it to the top level
- After promotion, the normal `include_metadata` / highlight filtering still runs, so metadata is cleaned as usual

- No. The parameter defaults to `false`, so existing API behavior is unchanged. The promotion logic only runs when explicitly requested. No changes to the execution pipeline, caching, or metadata population.

## Database Migrations

- None

## Env Config

- None

## Relevant Docs

- N/A

## Related Issues or PRs

- UN-2836

## Dependencies Versions

- None

## Notes on Testing

- POST with `include_extracted_text=true` returns `extracted_text` at file-result top level
- POST with `include_extracted_text=false` (default) does not include `extracted_text`
- GET polling with `include_extracted_text=true` returns `extracted_text` at file-result top level
- `include_extracted_text=true` works without `include_metadata=true`
- `include_extracted_text=true` works even when `ENABLE_HIGHLIGHT_API_DEPLOYMENT` is disabled
- When both `include_extracted_text=true` and `include_metadata=true`, extracted_text appears at top level and in metadata

## Screenshots

-

## Checklist


chandrasekharan-zipstack (approved):
LGTM, @pk-zipstack follow up on whether we can eventually avoid showing this extracted text as part of `include_metadata`

Deepak-Kesavan (approved):

PR #1910, opened 2026-04-10
UN-2980 [FIX] Handle null response

Deepak-Kesavan:
## What

- Fix `handle_json` silently storing `None` when the LLM returns literal `null` for a JSON-type prompt

## Why

- For prompts with output type `json`, the application expects the stored result to be a dict or list
- When the LLM responded with literal `null`, `repair_json_with_best_structure` correctly parsed it to Python `None` (JSON null → Python None)
- The old guard `isinstance(parsed_data, str)` only caught the case where `repair_json` completely failed and returned the original string — it missed `None` entirely
- Result: `None` was silently written to `structured_output[prompt_key]`, with no error log or warn event, breaking downstream consumers that expected a dict/list

## How

- Changed the guard in `handle_json` from `isinstance(parsed_data, str)` to `not isinstance(parsed_data, (dict, list))`
- Now any non-structured value (None, int, float, bool, str) falls into the existing error branch: logs the error, publishes the warn event, and stores `{}` as a safe fallback
- Valid objects/arrays — including `{"key": null}` — still flow through unchanged
- The existing `"na"` sentinel (stores `None`) is untouched
- The shared `repair_json_with_best_structure` helper is not modified — the type constraint is specific to the JSON output type and belongs at the caller

- No. The change is strictly more restrictive on what is accepted into the success branch for JSON-type prompts. `handle_json` is only invoked for `prompt_type == "json"`, where the contract is already "output must be a structured value". Any prior code path that received a scalar or `None` from this function was already a latent contract violation. The `"na"` branch, the `repair_json_with_best_structure` helper, and all other callers of that helper (e.g., structure tool) are untouched.

## Database Migrations

- None

## Env Config

- None

## Relevant Docs

- N/A

## Related Issues or PRs

- UN-2980

## Dependencies Versions

- None

## Notes on Testing

1. Create a JSON-type prompt
2. Force the LLM to respond with literal `null`
3. **Before fix:** `structured_output[prompt_key] == None`, no error/warn log
4. **After fix:** `structured_output[prompt_key] == {}`, error log + warn event published
5. Verify a normal JSON response like `{"a": 1}` still flows through unchanged
6. Verify `{"key": null}` is still stored as-is (not remapped to `{}`)
7. Verify `"na"` responses still store `None` as before

## Checklist


chandrasekharan-zipstack (approved):
@Deepak-Kesavan I hope this code isn't duplicated elsewhere and required

harini-venkataraman (approved):

PR #1912, opened 2026-04-10
UN-3398 [FIX] Fix SharePoint OAuth multi-tenant auth and duplicate OAuth button

kirtimanmishrazipstack:
## What

- Removed `AZUREAD_TENANT_OAUTH2_TENANT_ID` from Django social auth settings loop in `base.py`,  `sample.env`
- Removed duplicate OAuth button rendering at the top of the connector form in `ConfigureDs.jsx`

## Why

- Customers from external Azure AD tenants were getting `AADSTS50020` error when trying to OAuth with SharePoint connector. The env var was set to Zipstack's tenant ID, scoping the OAuth login URL to Zipstack's tenant only. The `social_core` library already defaults to `"common"` (multi-tenant) when the setting is absent.
- The OAuth "Sign in with Microsoft" button was rendering twice — once unconditionally at the top of the form, and once correctly inside the form gated by auth type selection.

## How

- Removed the tenant ID env var from the settings loop, letting `social_core`'s `AzureADTenantOAuth2` backend default to `"common"` for multi-tenant support.
- Removed the top-level `OAuthDs` component render; the one inside `RjsfFormLayout` (gated by `isOAuthMethodSelected()`) already handles it correctly.

- No. The tenant ID env var was only used for the OAuth login URL. Removing it makes the OAuth flow use `"common"`, which supports all tenants including Zipstack's. Client Credentials mode is unaffected (uses per-connector tenant ID). The OAuth button inside the form still renders correctly when OAuth auth type is selected.

## Database Migrations

- None

## Env Config

- Removed: `AZUREAD_TENANT_OAUTH2_TENANT_ID` (no longer needed; `social_core` defaults to `"common"`)

## Relevant Docs

- https://learn.microsoft.com/en-us/entra/identity-platform/v2-protocols#endpoints

## Related Issues or PRs

-

## Dependencies Versions

- None

## Notes on Testing

- Verify Client Credentials auth mode still works
-  Verify SharePoint OAuth flow works for users from any Azure AD tenant by running wortkflow
- Verify OAuth button only appears inside the form when "OAuth" auth type is selected, not at the top unconditionally

## Checklist


jaseemjaskp (approved):
LGTM

PR #1918, opened 2026-04-17
[HOTFIX] Use importlib.util.find_spec for pluggable worker discovery

chandrasekharan-zipstack:
## What

Replaces the filesystem `worker.py` existence check in `WorkerBuilder._verify_pluggable_worker_exists()` with `importlib.util.find_spec()`.

## Why

The pre-import check at `workers/shared/infrastructure/config/builder.py:338` hardcoded `worker.py` as the file to look for. This breaks any deployment where the pluggable worker has been compiled to an extension module (e.g. Nuitka or Cython produces a `worker.cpython-312-x86_64-linux-gnu.so` that replaces `worker.py`) — the module is perfectly importable, but the hard `.py` existence check rejects the worker before `importlib.import_module()` ever runs.

Observed error pattern (paraphrased):

```
ERROR: Pluggable worker file not found: /app/pluggable_worker/<name>/worker.py
ImportError: Pluggable worker '<name>' not found.
Expected module: pluggable_worker.<name>.worker
```

This is not specific to any one obfuscation tool — the check would also reject plugins distributed as Cython `.so`, bytecode-only `.pyc`, namespace packages, or any future loader that doesn't materialize a `.py` file on disk.

## How

Use Python's standard "is this module importable?" primitive instead of filesystem introspection:

```python
module_path = f"pluggable_worker.{worker_type.value}.worker"

if importlib.util.find_spec(module_path) is None:
    logger.error(f"Pluggable worker module not importable: {module_path}")
    return False

importlib.import_module(module_path)
```

`find_spec()` consults every registered finder (source `.py`, compiled `.so`, bytecode `.pyc`, namespace packages, zipimports) and returns `None` cleanly if nothing can produce the module. The subsequent `import_module()` still catches findable-but-broken modules (e.g. ImportError inside the loaded module).

The only semantic change: when a parent package exists but an intermediate subpackage is missing (`pluggable_worker/` exists but `pluggable_worker/<name>/` does not), `find_spec` raises `ModuleNotFoundError`. That's a subclass of `ImportError` and is already handled by the existing `except ImportError` block, so the function correctly returns `False` just like before.

No. Behavior preserved across all deployment scenarios, verified via a synthetic fixture that compiles a test worker with Nuitka and runs `_verify_pluggable_worker_exists()` against it:

| Scenario | `.py` present | `.so` present | Before | After |
|---|---|---|---|---|
| No subpackage (pluggable workers disabled) | no | no | `False` (file not found) | `False` (ImportError) |
| Source available | yes | no | `True` | `True` |
| Compiled extension module | no | yes | **`False` ← bug** | **`True` ← fixed** |

## Relevant Docs

- Python docs: [`importlib.util.find_spec`](https://docs.python.org/3/library/importlib.html#importlib.util.find_spec)

### DIFF
--- a/workers/shared/infrastructure/config/builder.py
+++ b/workers/shared/infrastructure/config/builder.py
@@ -323,24 +323,21 @@ class WorkerBuilder:
 
         try:
             import importlib
-            from pathlib import Path
+            import importlib.util
 
-            # Check if the worker.py file exists
-            # Path resolution: builder.py is at /app/workers/shared/infrastructure/config/
-            # We need to get to /app/workers/, so go up 3 levels
-            pluggable_worker_path = (
-                Path(__file__).resolve().parents[3]
-                / "pluggable_worker"
-                / worker_type.value
-                / "worker.py"
-            )
+            # Ask Python's import system whether the module is resolvable.
+            # find_spec() consults all registered finders and handles every
+            # module representation (source .py, Nuitka/Cython .so, .pyc,
+            # namespace packages, zipimports) — unlike a filesystem check
+            # for a specific file extension, which breaks for compiled plugins.
+            module_path = f"pluggable_worker.{worker_type.value}.worker"
 
-            if not pluggable_worker_path.exists():
-                logger.error(f"Pluggable worker file not found: {pluggable_worker_path}")
+            if importlib.util.find_spec(module_path) is None:
+                logger.error(f"Pluggable worker module not importable: {module_path}")
                 return False
 
-            # Try to import the module to verify it's valid
-            module_path = f"pluggable_worker.{worker_type.value}.worker"
+            # Load it to catch findable-but-broken modules (e.g. import errors
+            # inside worker.py that find_spec wouldn't surface).
             importlib.import_module(module_path)
 
         except ImportError:
--- a/workers/shared/infrastructure/config/builder.py
+++ b/workers/shared/infrastructure/config/builder.py
@@ -340,7 +340,9 @@ class WorkerBuilder:
             # inside worker.py that find_spec wouldn't surface).
             importlib.import_module(module_path)
 
-        except ImportError:
+        except (ImportError, ValueError):
+            # ValueError: find_spec raises this when sys.modules[module_path] is
+            # populated but has __spec__ = None (from a prior failed import).
             logger.exception(f"Failed to import pluggable worker {worker_type.value}")
             return False
         except (OSError, AttributeError):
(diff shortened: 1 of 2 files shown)


workers/shared/infrastructure/config/builder.py
```diff
@@ -323,27 +323,26 @@ def _verify_pluggable_worker_exists(worker_type: WorkerType) -> bool:
...
-                / "worker.py"
-            )
+            import importlib.util
+
+            # Ask Python's import system whether the module is resolvable.
+            # find_spec() consults all registered finders and handles every
+            # module representation (source .py, Nuitka/Cython .so, .pyc,
+            # namespace packages, zipimports) — unlike a filesystem check
+            # for a specific file extension, which breaks for compiled plugins.
+            module_path = f"pluggable_worker.{worker_type.value}.worker"
```
muhammad-ali-e, line 333: Hope this not break running worker locally insated of docker

muhammad-ali-e (approved):
added a comment

workers/shared/infrastructure/config/builder.py
chandrasekharan-zipstack: Tested locally — doesn't break local runs. `find_spec()` is actually a superset of the previous filesystem check.

**Local runs go through `workers/run-worker.sh`**, which sets `export PYTHONPATH="$WORKERS_DIR:${PYTHONPATH:-}"` (line 702) *before* launching celery. So `import pluggable_worker.<name>.worker` resolves the same way the launcher itself uses the dotted module name at line 444 (`celery_app_module="pluggable_worker.${worker_type}.worker"`).

**Coverage comparison vs old code:**

| Scenario | Old (`Path(...)/worker.py`) | New (`find_spec`) |
|---|---|---|
| Local `.py` via run-worker.sh | ✅ | ✅ (PYTHONPATH set by script) |
| Docker with uncompiled plugin | ✅ | ✅ |
| Docker with compiled `.so` (Cython/Nuitka) | ❌ (checks `.py`) | ✅ |
| OSS with no pluggable plugins | n/a — `is_pluggable()` returns False, early return | same |
| Bare `python worker.py` from repo root, no PYTHONPATH | ❌ | ❌ (and would fail at actual import too) |

The only way someone could get a false negative from the new code but not the old is if they ran a pluggable worker outside `run-worker.sh` without any PYTHONPATH setup — in that case the old check returned True but celery would have crashed on the real import a few lines later. So net change: we fail earlier with a cleaner log instead of getting a stack trace.

Also confirmed on `snapshot.180420262` locally: pluggable workers (agentic_callback + bulk_download worker.so) register and run fine — previously they crash-looped because `Path(...)/worker.py` returned False for compiled modules.

jaseemjaskp (approved):

PR #1922, opened 2026-04-21
UN-3421 [FIX] Use singular DeleteObject for MinIO cleanup fallback

pk-zipstack:
## What

- `_rm_files_individually()` in `unstract/sdk1/src/unstract/sdk1/file_storage/impl.py` now routes per-file deletion through `self.fs.rm_file(path)` instead of `self.fs.rm(path, recursive=False)`.
- Each per-file delete is wrapped in `try/except` so a single failure no longer aborts the remaining cleanup.
- New regression test module `unstract/sdk1/tests/file_storage/test_impl_rm.py` with 7 tests covering the bulk-delete happy path, the `MissingContentMD5` fallback, per-file and `rmdir` error tolerance, non-MD5 error propagation, non-recursive propagation, and an end-to-end assertion (with a mocked boto3 client) that the bulk `DeleteObjects` API is never re-entered from the fallback.

## Why

`FileStorage.rm(recursive=True)` → `s3fs.rm` → `_bulk_delete` → S3 `DeleteObjects` (bulk API). MinIO `2024-12-18` rejects this with `MissingContentMD5: Missing required header for this request: Content-Md5.`

The existing fallback caught the error, logged a warning, then looped and called `self.fs.rm(file_path, recursive=False)` per file — **which still routes through `_bulk_delete` in s3fs even for a single path**, so every per-file retry hit the same broken API. Net result: 100% of completed-execution cleanups were failing silently in production — 876 cleanups/hour with zero successes, orphaning `unstract/execution/<org>/<wf>/<exec>/<file_exec>/` directories on MinIO and causing unbounded `slab_reclaimable` growth.

Observed stack trace:

```
File "/app/callback/tasks.py", line 977, in _cleanup_directory
    file_storage.rm(directory_path, recursive=True)
File "/unstract/sdk1/src/unstract/sdk1/file_storage/helper.py", line 109, in wrapper
    raise FileOperationError(str(e)) from e
unstract.sdk1.exceptions.FileOperationError: [Errno 5] An error occurred (MissingContentMD5) when calling the DeleteObjects operation: Missing required header for this request: Content-Md5.
```

## How

- `s3fs.rm_file(path)` in s3fs dispatches to `_rm_file` → singular `DeleteObject`, which does **not** require `Content-MD5` and works on MinIO (this is the same API `mc rm` uses in production).
- The ticket's suggested snippet used `self.fs.s3.delete_object(...)` directly, but `s3fs.S3FileSystem.s3` is an **async `aiobotocore` client** — calling `delete_object()` on it returns a coroutine rather than a response. `rm_file()` hits the same singular `DeleteObject` API via s3fs's own sync wrapper (`mirror_sync_methods` on `AsyncFileSystem`), handles path splitting and cache invalidation, and is robust to future s3fs refactors.
- The outer `fs.rm(..., recursive=True)` call is preserved as the fast path; the fallback is only entered on `MissingContentMD5`.
- Per-file `try/except` preserves best-effort cleanup semantics so a single transient failure can no longer strand the rest of the directory.

Code change:

```python
def _rm_files_individually(self, path: str) -> None:
    files = self.fs.find(path)
    for file_path in files:
        try:
            self.fs.rm_file(file_path)  # singular DeleteObject
        except Exception as e:
            logger.warning("Failed to delete %s: %s", file_path, e)
    try:
        self.fs.rmdir(path)
    except Exception:
        pass  # Directory prefix may already be gone
```

No breaking changes:

- **Happy path (bulk delete succeeds on AWS S3, GCS via s3-compatible endpoints, MinIO versions without the bug)**: Untouched — `self.fs.rm(path, recursive=True)` is still the first attempt. The fallback is only reached on `MissingContentMD5`.
- **Fallback path**: Was 100% broken prior to this fix (every call raised), so swapping the per-file call from `fs.rm` to `fs.rm_file` strictly improves behavior.
- **Local filesystem / other providers**: The fallback is guarded by `if "MissingContentMD5" in str(e) and recursive:`, so non-S3 providers never enter this code path. For completeness, `rm_file` is also defined on fsspec's `AbstractFileSystem` (delegates to `_rm`), so the code is safe even if entered accidentally.
- **Error surface change**: Previously a single per-file delete failure would raise and abort the remaining cleanup. Now it's logged at WARNING and cleanup continues — same semantics as `rmdir`'s existing best-effort handling. Failures are still observable in logs.

## Database Migrations

- None

## Env Config

- None

## Relevant Docs

- MinIO changelog for `2024-12-18` release (`MissingContentMD5` strictness on `DeleteObjects`): https://github.com/minio/minio/releases/tag/RELEASE.2024-12-18T13-15-44Z
- AWS S3 `DeleteObject` API (singular — no `Content-MD5` required): https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObject.html
- AWS S3 `DeleteObjects` API (bulk — requires `Content-MD5`): https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html
- `s3fs._rm_file` implementation: https://github.com/fsspec/s3fs/blob/main/s3fs/core.py (uses `delete_object` singular)

## Related Issues or PRs

- Jira: [UN-3421](https://zipstack.atlassian.net/browse/UN-3421)
- Parent: [UN-3420](https://zipstack.atlassian.net/browse/UN-3420)
- Originating PR that introduced the broken fallback: #1849 (`impl.py:168-197`)
- Related earlier MinIO compatibility hotfix: [UN-2942](https://zipstack.atlassian.net/browse/UN-2942)

## Dependencies Versions

- No dependency changes. Fix relies on `s3fs.rm_file()` which has existed since s3fs added async support; the current pinned version (`s3fs[boto3]~=2024.10.0`) supports it.

## Notes on Testing

Local verification:

| Check | Result |
|---|---|
| New regression tests (`tests/file_storage/test_impl_rm.py`) | 7/7 passed |
| Full sdk1 test suite | 228/228 passed |
| Ruff on modified files | No new warnings (one pre-existing import-sort warning on `impl.py` confirmed on untouched `main`) |

Production-side acceptance criteria from the ticket (require deployment to verify):

- `mc ilm rule ls` shows no rules, then after an execution completes, the `unstract/execution/<org>/<wf>/<exec>/<file_exec>/` directory is empty on MinIO.
- Zero `MissingContentMD5` log lines from callback workers over a 1-hour production window.
- `slab_reclaimable` in `memory.stat` stops growing unboundedly.

Test matrix covered by the new regression tests:

| Test | Scenario | Assertion |
|---|---|---|
| `test_bulk_delete_succeeds` | `fs.rm` succeeds | Fallback is NOT invoked |
| `test_missing_md5_triggers_individual_delete_via_rm_file` | MD5 error + 3 files | `fs.rm_file` called per file, `fs.rmdir` called once |
| `test_fallback_continues_on_per_file_error` | 2nd file raises | All 3 files attempted, cleanup completes |
| `test_fallback_swallows_rmdir_error` | `rmdir` raises | No exception propagates |
| `test_non_md5_error_propagates` | `PermissionError` | Raised; fallback NOT invoked |
| `test_md5_error_without_recursive_propagates` | MD5 + `recursive=False` | Raised; no directory walk |
| `test_only_singular_delete_called` | Mocked boto3 client | `delete_objects` hit once (initial), `delete_object` called per file |

## Screenshots

N/A — backend-only fix in the SDK file-storage layer.

## Checklist

[UN-3421]: https://zipstack.atlassian.net/browse/UN-3421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/unstract/sdk1/src/unstract/sdk1/file_storage/impl.py
+++ b/unstract/sdk1/src/unstract/sdk1/file_storage/impl.py
@@ -184,13 +184,22 @@ class FileStorage(FileStorageInterface):
     def _rm_files_individually(self, path: str) -> None:
         """Fallback deletion: delete files one at a time.
 
-        Used when bulk S3 DeleteObjects fails (e.g., MissingContentMD5
-        with certain S3-compatible providers).
+        Used when bulk S3 ``DeleteObjects`` fails (e.g., ``MissingContentMD5``
+        with MinIO 2024-12-18+ and other S3-compatible providers).
+
+        ``self.fs.rm()`` — even with ``recursive=False`` for a single path —
+        routes through ``s3fs._rm`` → ``_bulk_delete`` → ``DeleteObjects``,
+        which is exactly what failed upstream. Using ``rm_file()`` instead
+        dispatches to ``s3fs._rm_file`` → singular ``DeleteObject``, which
+        does not require ``Content-MD5``.
         """
         files = self.fs.find(path)
         for file_path in files:
-            self.fs.rm(file_path, recursive=False)
-        # Clean up the "directory" prefix if it still exists
+            try:
+                self.fs.rm_file(file_path)
+            except Exception as e:
+                logger.warning("Failed to delete %s: %s", file_path, e)
+        # Best-effort cleanup of any remaining "directory" prefix
         try:
             self.fs.rmdir(path)
         except Exception:
--- a/unstract/sdk1/tests/file_storage/test_impl_rm.py
+++ b/unstract/sdk1/tests/file_storage/test_impl_rm.py
@@ -11,7 +11,6 @@ from __future__ import annotations
 from unittest.mock import MagicMock, patch
 
 import pytest
-
 from unstract.sdk1.exceptions import FileOperationError
 from unstract.sdk1.file_storage.impl import FileStorage
 from unstract.sdk1.file_storage.provider import FileStorageProvider
@@ -101,9 +100,7 @@ class TestRmFallback:
         assert s3_file_storage.fs.rm_file.call_count == 3
         s3_file_storage.fs.rmdir.assert_called_once_with("bucket/prefix/")
 
-    def test_fallback_swallows_rmdir_error(
-        self, s3_file_storage: FileStorage
-    ) -> None:
+    def test_fallback_swallows_rmdir_error(self, s3_file_storage: FileStorage) -> None:
         """Missing directory prefix after cleanup is expected; don't raise."""
         s3_file_storage.fs.rm.side_effect = _missing_md5_error()
         s3_file_storage.fs.find.return_value = ["bucket/prefix/a.txt"]
@@ -146,9 +143,7 @@ class TestFallbackDoesNotReenterBulkDelete:
     ``DeleteObject`` is invoked.
     """
 
-    def test_only_singular_delete_called(
-        self, s3_file_storage: FileStorage
-    ) -> None:
+    def test_only_singular_delete_called(self, s3_file_storage: FileStorage) -> None:
         boto3_client = MagicMock()
         boto3_client.delete_objects.side_effect = _missing_md5_error()
         boto3_client.delete_object.return_value = {"ResponseMetadata": {}}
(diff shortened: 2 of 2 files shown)


muhammad-ali-e (approved):
LGTM

harini-venkataraman (approved):

PR #1924, opened 2026-04-21
[FIX] Update Gemini embedding default model to gemini-embedding-001

jaseemjaskp:
## What

- Change the Gemini embedding adapter default model from `gemini/text-embedding-004` (legacy) to `gemini-embedding-001` (current stable GA).
- Drop the `gemini/` prefix from the user-facing default; validation still prepends it before passing to litellm, so routing is unaffected.
- Update description to also mention `gemini-embedding-2-preview` as the multimodal option.
- Update tests to reflect the new default and example model names.

## Why

- `text-embedding-004` is no longer a valid/supported model on Google AI Studio, so existing adapter instances using the default would fail at embed time.
- Per [Google AI docs](https://ai.google.dev/gemini-api/docs/embeddings), the current supported models are `gemini-embedding-001` (stable, text-only) and `gemini-embedding-2-preview` (multimodal preview).
- Keeping the user-facing default prefix-free is cleaner — the `gemini/` provider prefix is an implementation detail of litellm routing, not something end users should type.

## How

- `unstract/sdk1/src/unstract/sdk1/adapters/embedding1/static/gemini.json`: default + description updated.
- `unstract/sdk1/src/unstract/sdk1/adapters/base1.py`: example in the `validate_model` error message updated. Prefix auto-prepend logic unchanged.
- `unstract/sdk1/tests/test_gemini_embedding.py`: all assertions and fixture model names switched to `gemini-embedding-001`.

- Only the default and docstring/example values change. Any adapter instance already configured with an explicit model (including `text-embedding-004`) continues to validate and route the same way.
- Users who relied on the default and created adapters before this change will still have `gemini/text-embedding-004` stored — they should update their adapter model to `gemini-embedding-001` since the legacy model is no longer served.

## Database Migrations

- None.

## Env Config

- None.

## Relevant Docs

- https://ai.google.dev/gemini-api/docs/embeddings

## Related Issues or PRs

- Follow-up to #1891 (UN-3412 — Add Gemini embedding adapter for Google AI Studio).

## Dependencies Versions

- None.

## Notes on Testing

- `uv run pytest tests/test_gemini_embedding.py -v` — all 27 tests pass.
- Manual verification pending: create a Gemini embedding adapter with the new default against a live Google AI Studio API key and confirm successful embed.

## Checklist

### DIFF
--- a/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
@@ -1027,7 +1027,7 @@ class GeminiEmbeddingParameters(BaseEmbeddingParameters):
         if not model:
             raise ValueError(
                 "The 'model' field is required for the Gemini embedding adapter. "
-                "Example: 'gemini/gemini-embedding-001'"
+                "Example: 'gemini-embedding-001'"
             )
         if not model.startswith("gemini/"):
             model = f"gemini/{model}"
--- a/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
@@ -1027,7 +1027,7 @@ class GeminiEmbeddingParameters(BaseEmbeddingParameters):
         if not model:
             raise ValueError(
                 "The 'model' field is required for the Gemini embedding adapter. "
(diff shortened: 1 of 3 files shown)


muhammad-ali-e (approved):
LGTM

PR #1925, opened 2026-04-21
UN-3358 [FIX] List only S3 buckets the connector credentials can browse

kirtimanmishrazipstack:
## What

- S3 list down only buckets it has access to. Root user will see all buckets 
- `MinioFS` (S3/MinIO connector) now uses this subclass instead of `S3FileSystem`. No other connector method changes.
- `UnstractCloudStorage` opts out via a class-level `_FS_CLASS = S3FileSystem` hook, since its credentials always have full access to UCS buckets.

## Why

The S3/MinIO connector was listing every bucket in the account — including ones the caller cannot open — producing a long, misleading directory where most entries fail with "Access Denied" on click.

Root cause: `s3:ListAllMyBuckets` (account-level "show me the names") is a different IAM action from `s3:ListBucket` (per-bucket "let me look inside"). `s3fs.S3FileSystem._lsbuckets` calls `list_buckets`, which is gated only by the former. The caller's actual browse permission is never checked.

This hits IRSA customers (UN-2297) hardest: IRSA deployments typically grant `ListAllMyBuckets` account-wide but scope `ListBucket` to a handful of buckets, so the connector ends up surfacing dozens of buckets the role cannot use.

## How

After `_lsbuckets` returns the account-wide list, probe each bucket with `list_objects_v2(MaxKeys=1)` — the exact IAM action browsing needs — and drop any that fail. Probes run concurrently under an `asyncio.Semaphore(16)` cap. `UnstractCloudStorage` opts out via a `_FS_CLASS` hook.

No.

- Non-root paths go through the parent's `_ls` unchanged.
- For users with broad S3 permissions, every bucket passes the probe — the filter is a no-op and the visible list is identical.
- No schema, env var, migration, or connector-reconfig change required.
- Added cost: up to 16 concurrent `list_objects_v2` calls on root-listing (~<1s for typical <50-bucket accounts).

## Database Migrations

- None

## Env Config

- None

## Relevant Docs

- None

## Related Issues or PRs
- N/A

## Dependencies Versions

- None

## Notes on Testing

No automated tests in this PR. Verified manually against MinIO.

The IAM policy used for manual testing is the same one documented at [Unstract on-prem infra requirements — IAM Policy](https://docs.unstract.com/unstract/unstract_platform/user_guides/on_prem_infra_requirements/#iam-policy) (`s3:ListAllMyBuckets`, `s3:ListBucket` on the bucket, `s3:GetObject`/`s3:PutObject`/`s3:DeleteObject` on `bucket/*`).

**Manual steps (MinIO):**

1. Create a bucket: `test-iam-bucket`.
2. Create an IAM policy (from the docs above) scoping `s3:ListBucket` and the object actions to `test-iam-bucket` only.
3. Create a MinIO user (`test-user`), generate an access key + secret, and attach the policy.
4. Configure the S3/MinIO connector with `test-user`'s credentials + the MinIO endpoint URL.
5. Browse the connector root → only `test-iam-bucket` appears. Drill in → object listing is unchanged.
6. Sanity check with root credentials → the connector root lists every bucket (the filter is a no-op for full-access principals, confirming no regression for that path).

## Checklist

### DIFF
--- a/unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
+++ b/unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
@@ -1,4 +1,36 @@
+from enum import Enum
+
+from botocore.exceptions import ClientError
+
 from unstract.connectors.exceptions import ConnectorError
+
+
+class BucketProbeDisposition(Enum):
+    """Action for a `list_objects_v2` probe failure, per S3 error code."""
+
+    DROP = "drop"  # Credentials genuinely lack access — hide the bucket.
+    FAIL_OPEN = "fail_open"  # Region mismatch — keep bucket visible.
+    RETRY_FAIL_OPEN = "retry_fail_open"  # Throttled — retry once, then keep.
+
+
+# One row per S3 `Error.Code` → disposition. Same shape as
+# `S3FS_EXC_TO_UNSTRACT_EXC` below (code → meaning). Codes absent from
+# this table propagate as real errors instead of silently hiding a bucket.
+BUCKET_PROBE_DISPOSITION: dict[str, BucketProbeDisposition] = {
+    "AccessDenied": BucketProbeDisposition.DROP,
+    "AllAccessDisabled": BucketProbeDisposition.DROP,
+    "PermanentRedirect": BucketProbeDisposition.FAIL_OPEN,
+    "IllegalLocationConstraintException": BucketProbeDisposition.FAIL_OPEN,
+    "SlowDown": BucketProbeDisposition.RETRY_FAIL_OPEN,
+    "Throttling": BucketProbeDisposition.RETRY_FAIL_OPEN,
+    "ThrottlingException": BucketProbeDisposition.RETRY_FAIL_OPEN,
+    "RequestTimeTooSkewed": BucketProbeDisposition.RETRY_FAIL_OPEN,
+}
+
+
+def client_error_code(exc: ClientError) -> str:
+    return str(exc.response.get("Error", {}).get("Code", "") or "")
+
 
 S3FS_EXC_TO_UNSTRACT_EXC: dict[str, str] = {
     # Auth errors
--- a/unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
+++ b/unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
@@ -31,12 +31,18 @@ BUCKET_PROBE_DISPOSITION: dict[str, BucketProbeDisposition] = {
 def s3_error_code(exc: BaseException) -> str:
     """Return the S3 `Error.Code` from a `ClientError` or its s3fs-translated
     wrapper (`PermissionError` / `FileNotFoundError` / `OSError`).
+
+    Walks `__cause__` first (explicit `raise X from original`), then falls
+    back to `__context__` (implicit chaining inside an `except` block). A
+    `seen` set guards against pathological cycles.
     """
+    seen: set[int] = set()
     target: BaseException | None = exc
-    while target is not None:
+    while target is not None and id(target) not in seen:
+        seen.add(id(target))
         if isinstance(target, ClientError):
             return str(target.response.get("Error", {}).get("Code", "") or "")
-        target = target.__cause__
+        target = target.__cause__ or target.__context__
     return ""
 
 
--- a/unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
+++ b/unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
@@ -8,28 +8,36 @@ from unstract.connectors.exceptions import ConnectorError
 class BucketProbeDisposition(Enum):
     """Action for a `list_objects_v2` probe failure, per S3 error code."""
 
-    DROP = "drop"  # Credentials genuinely lack access — hide the bucket.
+    DROP = "drop"  # Hide the bucket (no access, or bucket gone).
     FAIL_OPEN = "fail_open"  # Region mismatch — keep bucket visible.
     RETRY_FAIL_OPEN = "retry_fail_open"  # Throttled — retry once, then keep.
 
 
-# One row per S3 `Error.Code` → disposition. Same shape as
-# `S3FS_EXC_TO_UNSTRACT_EXC` below (code → meaning). Codes absent from
-# this table propagate as real errors instead of silently hiding a bucket.
+# S3 `Error.Code` → probe disposition. Unlisted codes propagate.
+# `RequestTimeTooSkewed` is omitted deliberately: it's a system-wide clock
+# issue, not a bucket outcome — let it surface via `handle_s3fs_exception`.
 BUCKET_PROBE_DISPOSITION: dict[str, BucketProbeDisposition] = {
     "AccessDenied": BucketProbeDisposition.DROP,
     "AllAccessDisabled": BucketProbeDisposition.DROP,
+    "NoSuchBucket": BucketProbeDisposition.DROP,
     "PermanentRedirect": BucketProbeDisposition.FAIL_OPEN,
     "IllegalLocationConstraintException": BucketProbeDisposition.FAIL_OPEN,
     "SlowDown": BucketProbeDisposition.RETRY_FAIL_OPEN,
     "Throttling": BucketProbeDisposition.RETRY_FAIL_OPEN,
     "ThrottlingException": BucketProbeDisposition.RETRY_FAIL_OPEN,
-    "RequestTimeTooSkewed": BucketProbeDisposition.RETRY_FAIL_OPEN,
 }
 
 
-def client_error_code(exc: ClientError) -> str:
-    return str(exc.response.get("Error", {}).get("Code", "") or "")
+def s3_error_code(exc: BaseException) -> str:
+    """Return the S3 `Error.Code` from a `ClientError` or its s3fs-translated
+    wrapper (`PermissionError` / `FileNotFoundError` / `OSError`).
+    """
+    target: BaseException | None = exc
+    while target is not None:
+        if isinstance(target, ClientError):
+            return str(target.response.get("Error", {}).get("Code", "") or "")
+        target = target.__cause__
+    return ""
 
 
 S3FS_EXC_TO_UNSTRACT_EXC: dict[str, str] = {
--- a/unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
+++ b/unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
@@ -5,35 +5,38 @@ from datetime import UTC, datetime
 from email.utils import parsedate_to_datetime
 from typing import Any
 
+from botocore.exceptions import ClientError
 from s3fs.core import S3FileSystem
 
 from unstract.connectors.filesystems.unstract_file_system import UnstractFileSystem
 
-from .exceptions import handle_s3fs_exception
+from .exceptions import (
+    BUCKET_PROBE_DISPOSITION,
+    BucketProbeDisposition,
+    client_error_code,
+    handle_s3fs_exception,
+)
 
 logger = logging.getLogger(__name__)
 
 # Cap concurrent per-bucket probes to avoid S3 503 SlowDown on large accounts.
 _MAX_CONCURRENT_BUCKET_PROBES = 16
+_BUCKET_PROBE_RETRY_DELAY_SECONDS = 0.5
 
 
 class _AccessFilteredS3FileSystem(S3FileSystem):  # type: ignore[misc]
     """S3FileSystem that lists only buckets the credentials can browse.
 
-    `s3:ListAllMyBuckets` (account-level) returns every bucket in the
-    account, which is wider than `s3:ListBucket` (per-bucket). We probe
-    each bucket with a 1-key `list_objects_v2` and drop any that fail.
+    Probes each bucket with `list_objects_v2(MaxKeys=1)` and looks up the
+    resulting `ClientError.Code` in `BUCKET_PROBE_DISPOSITION`. Unknown
+    codes propagate so real failures aren't silently hidden.
     """
 
     async def _lsbuckets(self, refresh: bool = False) -> list[dict[str, Any]]:
-        if not refresh and "" in self.dircache:
-            return self.dircache[""]  # type: ignore[no-any-return]
         buckets: list[dict[str, Any]] = await super()._lsbuckets(refresh=refresh)
         if not buckets:
             return buckets
-        accessible = await self._filter_accessible_buckets(buckets)
-        self.dircache[""] = accessible
-        return accessible
+        return await self._filter_accessible_buckets(buckets)
 
     async def _filter_accessible_buckets(
         self, buckets: list[dict[str, Any]]
@@ -45,15 +48,55 @@ class _AccessFilteredS3FileSystem(S3FileSystem):  # type: ignore[misc]
                 return await self._is_bucket_accessible(name)
 
         results = await asyncio.gather(*(_probe(b["name"]) for b in buckets))
-        return [b for b, ok in zip(buckets, results, strict=True) if ok]
+        accessible = [b for b, ok in zip(buckets, results, strict=True) if ok]
+        dropped = len(buckets) - len(accessible)
+        if dropped:
+            logger.info(
+                "[S3/MinIO] Bucket filter: kept %d of %d; dropped %d on "
+                "AccessDenied/AllAccessDisabled.",
+                len(accessible),
+                len(buckets),
+                dropped,
+            )
+        return accessible
 
     async def _is_bucket_accessible(self, name: str) -> bool:
         try:
             await self._call_s3("list_objects_v2", Bucket=name, MaxKeys=1)
             return True
-        except Exception as exc:
-            logger.debug("[S3/MinIO] Dropping inaccessible bucket '%s': %s", name, exc)
-            return False
+        except ClientError as exc:
+            code = client_error_code(exc)
+            disposition = BUCKET_PROBE_DISPOSITION.get(code)
+            if disposition is BucketProbeDisposition.DROP:
+                return False
+            if disposition is BucketProbeDisposition.FAIL_OPEN:
+                logger.info(
+                    "[S3/MinIO] Bucket '%s' probe returned %s; keeping in listing.",
+                    name,
+                    code,
+                )
+                return True
+            if disposition is BucketProbeDisposition.RETRY_FAIL_OPEN:
+                return await self._retry_probe_fail_open(name, code)
+            raise
+
+    async def _retry_probe_fail_open(self, name: str, first_code: str) -> bool:
+        await asyncio.sleep(_BUCKET_PROBE_RETRY_DELAY_SECONDS)
+        try:
+            await self._call_s3("list_objects_v2", Bucket=name, MaxKeys=1)
+            return True
+        except ClientError as retry_exc:
+            retry_code = client_error_code(retry_exc)
+            if BUCKET_PROBE_DISPOSITION.get(retry_code) is BucketProbeDisposition.DROP:
+                return False
+            logger.info(
+                "[S3/MinIO] Bucket '%s' still transient after retry "
+                "(first=%s, retry=%s); keeping in listing.",
+                name,
+                first_code,
+                retry_code,
+            )
+            return True
 
 
 class MinioFS(UnstractFileSystem):
--- a/unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
+++ b/unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
@@ -1,6 +1,7 @@
 import asyncio
 import logging
 import os
+import random
 from datetime import UTC, datetime
 from email.utils import parsedate_to_datetime
 from typing import Any
@@ -51,8 +52,8 @@ class _AccessFilteredS3FileSystem(S3FileSystem):  # type: ignore[misc]
         dropped = len(buckets) - len(accessible)
         if dropped:
             logger.info(
-                "[S3/MinIO] Bucket filter: kept %d of %d; dropped %d on "
-                "AccessDenied/AllAccessDisabled.",
+                "[S3/MinIO] Bucket filter: kept %d of %d; "
+                "dropped %d on DROP dispositions.",
                 len(accessible),
                 len(buckets),
                 dropped,
@@ -69,18 +70,30 @@ class _AccessFilteredS3FileSystem(S3FileSystem):  # type: ignore[misc]
             if disposition is BucketProbeDisposition.DROP:
                 return False
             if disposition is BucketProbeDisposition.FAIL_OPEN:
-                logger.info(
-                    "[S3/MinIO] Bucket '%s' probe returned %s; keeping in listing.",
+                logger.warning(
+                    "[S3/MinIO] Bucket %r probe returned %s; keeping in listing.",
                     name,
                     code,
                 )
                 return True
             if disposition is BucketProbeDisposition.RETRY_FAIL_OPEN:
                 return await self._retry_probe_fail_open(name, code)
+            # Unclassified code: log which bucket before propagating so the
+            # gather-cancellation has a breadcrumb.
+            logger.exception(
+                "[S3/MinIO] Unclassified probe failure for bucket %r "
+                "(code=%r); propagating.",
+                name,
+                code,
+            )
             raise
 
     async def _retry_probe_fail_open(self, name: str, first_code: str) -> bool:
-        await asyncio.sleep(_BUCKET_PROBE_RETRY_DELAY_SECONDS)
+        # Full jitter avoids correlated retries under SlowDown thundering herd.
+        await asyncio.sleep(
+            _BUCKET_PROBE_RETRY_DELAY_SECONDS
+            + random.uniform(0, _BUCKET_PROBE_RETRY_DELAY_SECONDS)
+        )
         try:
             await self._call_s3("list_objects_v2", Bucket=name, MaxKeys=1)
             return True
@@ -90,11 +103,18 @@ class _AccessFilteredS3FileSystem(S3FileSystem):  # type: ignore[misc]
             if disposition is BucketProbeDisposition.DROP:
                 return False
             if disposition is BucketProbeDisposition.FAIL_OPEN:
+                logger.warning(
+                    "[S3/MinIO] Bucket %r fail-open on retry "
+                    "(first=%s, retry=%s); keeping in listing.",
+                    name,
+                    first_code,
+                    retry_code,
+                )
                 return True
             if disposition is BucketProbeDisposition.RETRY_FAIL_OPEN:
                 # Still transient after one retry — keep to avoid flapping.
                 logger.info(
-                    "[S3/MinIO] Bucket '%s' still transient after retry "
+                    "[S3/MinIO] Bucket %r still transient after retry "
                     "(first=%s, retry=%s); keeping in listing.",
                     name,
                     first_code,
@@ -102,7 +122,14 @@ class _AccessFilteredS3FileSystem(S3FileSystem):  # type: ignore[misc]
                 )
                 return True
             # Unclassified retry failure (e.g. credentials expired mid-probe):
-            # re-raise so the real error surfaces instead of being hidden.
+            # log the bucket, then re-raise so the real error surfaces.
+            logger.exception(
+                "[S3/MinIO] Unclassified retry probe failure for bucket %r "
+                "(first=%s, retry=%s); propagating.",
+                name,
+                first_code,
+                retry_code,
+            )
             raise
 
 
--- a/unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
+++ b/unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
@@ -13,37 +13,39 @@ from .exceptions import handle_s3fs_exception
 
 logger = logging.getLogger(__name__)
 
+# Cap concurrent per-bucket probes to avoid S3 503 SlowDown on large accounts.
+_MAX_CONCURRENT_BUCKET_PROBES = 16
+
 
 class _AccessFilteredS3FileSystem(S3FileSystem):  # type: ignore[misc]
     """S3FileSystem that lists only buckets the credentials can browse.
 
-    The parent's `_lsbuckets` returns every bucket reachable via
-    `s3:ListAllMyBuckets`. That set is wider than what the caller can
-    actually read — especially under IRSA, where `ListAllMyBuckets` may
-    be granted account-wide while `s3:ListBucket` is scoped to a handful
-    of buckets. We probe each returned bucket with a 1-key
-    `list_objects_v2` (the exact IAM action required to browse) and drop
-    any that respond with AccessDenied / NoSuchBucket / other errors.
+    `s3:ListAllMyBuckets` (account-level) returns every bucket in the
+    account, which is wider than `s3:ListBucket` (per-bucket). We probe
+    each bucket with a 1-key `list_objects_v2` and drop any that fail.
     """
 
     async def _lsbuckets(self, refresh: bool = False) -> list[dict[str, Any]]:
+        if not refresh and "" in self.dircache:
+            return self.dircache[""]  # type: ignore[no-any-return]
         buckets: list[dict[str, Any]] = await super()._lsbuckets(refresh=refresh)
         if not buckets:
             return buckets
         accessible = await self._filter_accessible_buckets(buckets)
-        # Replace the cached entry populated by the parent so subsequent
-        # cache reads return the filtered view.
         self.dircache[""] = accessible
         return accessible
 
     async def _filter_accessible_buckets(
         self, buckets: list[dict[str, Any]]
     ) -> list[dict[str, Any]]:
-        results = await asyncio.gather(
-            *(self._is_bucket_accessible(b["name"]) for b in buckets),
-            return_exceptions=True,
-        )
-        return [b for b, ok in zip(buckets, results, strict=False) if ok is True]
+        sem = asyncio.Semaphore(_MAX_CONCURRENT_BUCKET_PROBES)
+
+        async def _probe(name: str) -> bool:
+            async with sem:
+                return await self._is_bucket_accessible(name)
+
+        results = await asyncio.gather(*(_probe(b["name"]) for b in buckets))
+        return [b for b, ok in zip(buckets, results, strict=True) if ok]
 
     async def _is_bucket_accessible(self, name: str) -> bool:
         try:
@@ -55,6 +57,10 @@ class _AccessFilteredS3FileSystem(S3FileSystem):  # type: ignore[misc]
 
 
 class MinioFS(UnstractFileSystem):
+    # Subclasses can override with plain `S3FileSystem` to skip bucket-access
+    # filtering (e.g. for platform-managed storage).
+    _FS_CLASS: type[S3FileSystem] = _AccessFilteredS3FileSystem
+
     def __init__(self, settings: dict[str, Any]):
         super().__init__("MinioFS/S3")
         key = (settings.get("key") or "").strip()
@@ -71,7 +77,7 @@ class MinioFS(UnstractFileSystem):
         if endpoint_url:
             creds["endpoint_url"] = endpoint_url
 
-        self.s3 = _AccessFilteredS3FileSystem(
+        self.s3 = self._FS_CLASS(
             anon=False,
             use_listings_cache=False,
             default_fill_cache=False,
--- a/unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
+++ b/unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
@@ -13,8 +13,8 @@ from unstract.connectors.filesystems.unstract_file_system import UnstractFileSys
 from .exceptions import (
     BUCKET_PROBE_DISPOSITION,
     BucketProbeDisposition,
-    client_error_code,
     handle_s3fs_exception,
+    s3_error_code,
 )
 
 logger = logging.getLogger(__name__)
@@ -25,11 +25,10 @@ _BUCKET_PROBE_RETRY_DELAY_SECONDS = 0.5
 
 
 class _AccessFilteredS3FileSystem(S3FileSystem):  # type: ignore[misc]
-    """S3FileSystem that lists only buckets the credentials can browse.
+    """Lists only buckets the credentials can browse.
 
-    Probes each bucket with `list_objects_v2(MaxKeys=1)` and looks up the
-    resulting `ClientError.Code` in `BUCKET_PROBE_DISPOSITION`. Unknown
-    codes propagate so real failures aren't silently hidden.
+    Probes each bucket and looks up the S3 `Error.Code` in
+    `BUCKET_PROBE_DISPOSITION`. Unlisted codes propagate.
     """
 
     async def _lsbuckets(self, refresh: bool = False) -> list[dict[str, Any]]:
@@ -64,8 +63,8 @@ class _AccessFilteredS3FileSystem(S3FileSystem):  # type: ignore[misc]
         try:
             await self._call_s3("list_objects_v2", Bucket=name, MaxKeys=1)
             return True
-        except ClientError as exc:
-            code = client_error_code(exc)
+        except (ClientError, OSError) as exc:
+            code = s3_error_code(exc)
             disposition = BUCKET_PROBE_DISPOSITION.get(code)
             if disposition is BucketProbeDisposition.DROP:
                 return False
@@ -85,18 +84,26 @@ class _AccessFilteredS3FileSystem(S3FileSystem):  # type: ignore[misc]
         try:
             await self._call_s3("list_objects_v2", Bucket=name, MaxKeys=1)
             return True
-        except ClientError as retry_exc:
-            retry_code = client_error_code(retry_exc)
-            if BUCKET_PROBE_DISPOSITION.get(retry_code) is BucketProbeDisposition.DROP:
+        except (ClientError, OSError) as retry_exc:
+            retry_code = s3_error_code(retry_exc)
+            disposition = BUCKET_PROBE_DISPOSITION.get(retry_code)
+            if disposition is BucketProbeDisposition.DROP:
                 return False
-            logger.info(
-                "[S3/MinIO] Bucket '%s' still transient after retry "
-                "(first=%s, retry=%s); keeping in listing.",
-                name,
-                first_code,
-                retry_code,
-            )
-            return True
+            if disposition is BucketProbeDisposition.FAIL_OPEN:
+                return True
+            if disposition is BucketProbeDisposition.RETRY_FAIL_OPEN:
+                # Still transient after one retry — keep to avoid flapping.
+                logger.info(
+                    "[S3/MinIO] Bucket '%s' still transient after retry "
+                    "(first=%s, retry=%s); keeping in listing.",
+                    name,
+                    first_code,
+                    retry_code,
+                )
+                return True
+            # Unclassified retry failure (e.g. credentials expired mid-probe):
+            # re-raise so the real error surfaces instead of being hidden.
+            raise
 
 
 class MinioFS(UnstractFileSystem):
--- a/unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
+++ b/unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 import os
 from datetime import UTC, datetime
@@ -11,6 +12,46 @@ from unstract.connectors.filesystems.unstract_file_system import UnstractFileSys
 from .exceptions import handle_s3fs_exception
 
 logger = logging.getLogger(__name__)
+
+
+class _AccessFilteredS3FileSystem(S3FileSystem):  # type: ignore[misc]
+    """S3FileSystem that lists only buckets the credentials can browse.
+
+    The parent's `_lsbuckets` returns every bucket reachable via
+    `s3:ListAllMyBuckets`. That set is wider than what the caller can
+    actually read — especially under IRSA, where `ListAllMyBuckets` may
+    be granted account-wide while `s3:ListBucket` is scoped to a handful
+    of buckets. We probe each returned bucket with a 1-key
+    `list_objects_v2` (the exact IAM action required to browse) and drop
+    any that respond with AccessDenied / NoSuchBucket / other errors.
+    """
+
+    async def _lsbuckets(self, refresh: bool = False) -> list[dict[str, Any]]:
+        buckets: list[dict[str, Any]] = await super()._lsbuckets(refresh=refresh)
+        if not buckets:
+            return buckets
+        accessible = await self._filter_accessible_buckets(buckets)
+        # Replace the cached entry populated by the parent so subsequent
+        # cache reads return the filtered view.
+        self.dircache[""] = accessible
+        return accessible
+
+    async def _filter_accessible_buckets(
+        self, buckets: list[dict[str, Any]]
+    ) -> list[dict[str, Any]]:
+        results = await asyncio.gather(
+            *(self._is_bucket_accessible(b["name"]) for b in buckets),
+            return_exceptions=True,
+        )
+        return [b for b, ok in zip(buckets, results, strict=False) if ok is True]
+
+    async def _is_bucket_accessible(self, name: str) -> bool:
+        try:
+            await self._call_s3("list_objects_v2", Bucket=name, MaxKeys=1)
+            return True
+        except Exception as exc:
+            logger.debug("[S3/MinIO] Dropping inaccessible bucket '%s': %s", name, exc)
+            return False
 
 
 class MinioFS(UnstractFileSystem):
@@ -30,7 +71,7 @@ class MinioFS(UnstractFileSystem):
         if endpoint_url:
             creds["endpoint_url"] = endpoint_url
 
-        self.s3 = S3FileSystem(
+        self.s3 = _AccessFilteredS3FileSystem(
             anon=False,
             use_listings_cache=False,
             default_fill_cache=False,
--- a/unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
+++ b/unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
@@ -57,8 +57,9 @@ class _AccessFilteredS3FileSystem(S3FileSystem):  # type: ignore[misc]
 
 
 class MinioFS(UnstractFileSystem):
-    # Subclasses can override with plain `S3FileSystem` to skip bucket-access
-    # filtering (e.g. for platform-managed storage).
+    # Override with plain S3FileSystem in a subclass when the credentials are
+    # known to have full access to every bucket they list, so the per-bucket
+    # access probe in _AccessFilteredS3FileSystem can be skipped.
     _FS_CLASS: type[S3FileSystem] = _AccessFilteredS3FileSystem
 
     def __init__(self, settings: dict[str, Any]):
--- a/unstract/connectors/src/unstract/connectors/filesystems/ucs/ucs.py
+++ b/unstract/connectors/src/unstract/connectors/filesystems/ucs/ucs.py
@@ -1,5 +1,7 @@
 import os
 
+from s3fs.core import S3FileSystem
+
 from unstract.connectors.filesystems.minio.minio import MinioFS
 
 
@@ -8,6 +10,11 @@ class UnstractCloudStorage(MinioFS):
 
     Implemented with Google Cloud Storage through Minio.
     """
+
+    # UCS credentials always have full access to UCS buckets, so skip the
+    # per-bucket access probe that MinioFS runs on its fsspec filesystem.
+    # The probe adds latency and could hide a bucket on a transient S3 error.
+    _FS_CLASS = S3FileSystem
 
     @staticmethod
     def get_id() -> str:
--- a/unstract/connectors/tests/filesystems/test_miniofs.py
+++ b/unstract/connectors/tests/filesystems/test_miniofs.py
@@ -1,7 +1,12 @@
+import asyncio
 import os
 import unittest
+from unittest.mock import AsyncMock, patch
 
-from unstract.connectors.filesystems.minio.minio import MinioFS
+from unstract.connectors.filesystems.minio.minio import (
+    MinioFS,
+    _AccessFilteredS3FileSystem,
+)
 
 
 class TestMinoFS(unittest.TestCase):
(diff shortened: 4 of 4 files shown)


unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -13,6 +14,46 @@
...
+        return accessible
+
+    async def _filter_accessible_buckets(
+        self, buckets: list[dict[str, Any]]
+    ) -> list[dict[str, Any]]:
+        results = await asyncio.gather(
+            *(self._is_bucket_accessible(b["name"]) for b in buckets),
+            return_exceptions=True,
+        )
+        return [b for b, ok in zip(buckets, results, strict=False) if ok is True]
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Triaged the three sub-points:

1. **Unbounded fan-out** — addressed. `_filter_accessible_buckets` now wraps each probe in an `asyncio.Semaphore(16)`, capping concurrent `list_objects_v2` calls at 16 regardless of bucket count.

2. **Probe caching (TTL)** — declined for this PR. `MinioFS` is explicitly constructed with `use_listings_cache=False`; adding a per-credential TTL cache would conflict with that intent and introduce non-trivial key-management. If production telemetry later shows probe latency is a real concern, it's worth a dedicated follow-up ticket with clear invalidation semantics.

3. **`return_exceptions=True` redundant** — addressed. Dropped the flag and switched to `zip(..., strict=True)`. `_is_bucket_accessible` already catches every `Exception` and returns `bool`, so the only thing the flag was swallowing was `BaseException` (e.g. `asyncio.CancelledError`) — not desirable.

kirtimanmishrazipstack:
Bot-comment triage on commit `74d4d0ec5`:

**Applied**
- `asyncio.gather(..., return_exceptions=True)` dropped (greptile + coderabbit) — no more `BaseException` / `CancelledError` swallowing.
- Per-bucket probes now run under an `asyncio.Semaphore(16)` to cap concurrent `list_objects_v2` calls (coderabbit).
- `_lsbuckets(refresh=False)` now short-circuits on `self.dircache[""]` to prevent re-probing an already-filtered list (greptile).
- `UnstractCloudStorage` opts out of the bucket-access filter via a class-level `_FS_CLASS = S3FileSystem` hook, so platform-managed storage isn't silently hidden by transient probe failures (coderabbit).

**Declined (out of scope)**
- **TTL cache for probe results (coderabbit).** Conflicts with `MinioFS`'s explicit `use_listings_cache=False` design intent, and a credential-keyed cache introduces non-trivial invalidation. Deferred to a follow-up if telemetry shows it's needed.
- **Re-adding the four mocked unit tests for `_AccessFilteredS3FileSystem` (greptile P1).** The test file was deliberately rolled back in `74d4d0ec5` to keep this PR focused on the minio.py hotfix; the PR body was updated accordingly and no longer advertises them. Manual verification against a scoped MinIO user (documented in the ticket) is what this PR was tested with. Re-adding the unit tests is tracked as a follow-up and is not blocking this merge.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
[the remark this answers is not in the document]
kirtimanmishrazipstack: Still accurate after subsequent rounds: no `return_exceptions=True`; `_filter_accessible_buckets` uses `asyncio.gather(...)` with `zip(..., strict=True)`. If any probe raises (e.g. unknown `Error.Code`, `CancelledError`), it propagates and cancels sibling probes instead of being silently swallowed.

Covered by `tests/filesystems/test_miniofs.py::TestAccessFilteredS3FileSystem::test_unknown_error_propagates` and `test_throttling_then_unknown_on_retry_propagates`.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -13,6 +14,46 @@
...
+
+    async def _lsbuckets(self, refresh: bool = False) -> list[dict[str, Any]]:
+        buckets: list[dict[str, Any]] = await super()._lsbuckets(refresh=refresh)
+        if not buckets:
+            return buckets
+        accessible = await self._filter_accessible_buckets(buckets)
+        # Replace the cached entry populated by the parent so subsequent
+        # cache reads return the filtered view.
+        self.dircache[""] = accessible
+        return accessible
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Correction to my earlier reply — the `self.dircache[""]` short-circuit has been **removed** (commit \`ae26d49d\`) after Chandru pointed out it re-introduces instance-level caching that `MinioFS` intentionally disables (`use_listings_cache=False`, `default_fill_cache=False`, `default_cache_type="none"`, `skip_instance_cache=True`).

Current behavior: `_lsbuckets` re-runs the filter on every call. That trades a little redundant work for correctness — no cached "accessible" set can go stale after a transient `SlowDown`/throttle. So the "double-filter on cached re-entry" is no longer a concern because there is no cache to re-enter from.

If probe latency shows up as a real problem later, a TTL-bounded cache with explicit invalidation would be the right design — but it's intentionally deferred out of this PR.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -13,6 +14,46 @@
...
+    """S3FileSystem that lists only buckets the credentials can browse.
+
+    The parent's `_lsbuckets` returns every bucket reachable via
+    `s3:ListAllMyBuckets`. That set is wider than what the caller can
+    actually read — especially under IRSA, where `ListAllMyBuckets` may
+    be granted account-wide while `s3:ListBucket` is scoped to a handful
+    of buckets. We probe each returned bucket with a 1-key
+    `list_objects_v2` (the exact IAM action required to browse) and drop
+    any that respond with AccessDenied / NoSuchBucket / other errors.
+    """
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Good catch. `MinioFS` now exposes a class-level `_FS_CLASS` hook and `UnstractCloudStorage` overrides it to plain `S3FileSystem`, so UCS no longer runs the per-bucket access probe. Platform-managed buckets will surface transient S3 errors instead of being silently dropped.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -12,8 +13,55 @@
...
+                return await self._is_bucket_accessible(name)
+
+        results = await asyncio.gather(*(_probe(b["name"]) for b in buckets))
+        return [b for b, ok in zip(buckets, results, strict=True) if ok]
+
+    async def _is_bucket_accessible(self, name: str) -> bool:
+        try:
+            await self._call_s3("list_objects_v2", Bucket=name, MaxKeys=1)
+            return True
+        except Exception as exc:
```
chandrasekharan-zipstack: `except Exception` collapses semantically different failures into "drop this bucket". On AWS this is especially risky for multi-region accounts: a bucket in a region other than `region_name` returns `PermanentRedirect` and silently disappears from the user's connector list. Transient conditions like `SlowDown`/503 throttling, `EndpointConnectionError`, `ReadTimeoutError`, `NoCredentialsError`, or `ExpiredToken` also all map to "bucket not accessible" — the user sees an empty/short list instead of a real error.

Suggest catching `botocore.exceptions.ClientError` only and branching on `error_code`:
- `AccessDenied` / `AllAccessDisabled` → drop (the intended case)
- `PermanentRedirect` / `IllegalLocationConstraintException` → keep the bucket (fail-open)
- Throttling / `SlowDown` → short retry, then fail-open
- Anything else (network, credentials, SSL, param-validation) → let it propagate so the caller sees the real failure instead of an empty list

Also consider bumping the drop log from `DEBUG` to `INFO` and emitting one summary line after `gather` (`"Filtered N/M buckets (denied=x, redirect=y, throttled=z)"`) — at DEBUG level there is no prod forensic trail when a user reports "my bucket is missing".

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -12,8 +13,55 @@
...
+
+class _AccessFilteredS3FileSystem(S3FileSystem):  # type: ignore[misc]
+    """S3FileSystem that lists only buckets the credentials can browse.
+
+    `s3:ListAllMyBuckets` (account-level) returns every bucket in the
+    account, which is wider than `s3:ListBucket` (per-bucket). We probe
+    each bucket with a 1-key `list_objects_v2` and drop any that fail.
+    """
+
+    async def _lsbuckets(self, refresh: bool = False) -> list[dict[str, Any]]:
```
chandrasekharan-zipstack, line 35: Both `_lsbuckets` (here) and `_call_s3` (line 52) are underscore-prefixed s3fs internals with no stability guarantee — `_lsbuckets`'s signature has changed across s3fs versions (earlier `()`, then `(refresh=False)`). An s3fs upgrade can silently break this override or, worse, bypass it. Two mitigations:

1. Pin s3fs to a known-good range in `pyproject.toml`.
2. Add a small unit test that monkey-patches `_call_s3` and asserts the override is exercised end-to-end — any signature drift on upgrade will then fail loudly in CI instead of silently reverting to the unfiltered parent behavior.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -12,8 +13,55 @@
...
+        if not refresh and "" in self.dircache:
```
chandrasekharan-zipstack: `MinioFS` constructs this filesystem with `use_listings_cache=False`, `default_fill_cache=False`, `default_cache_type="none"`, `skip_instance_cache=True` — i.e. caching is intentionally disabled. Reading/writing `self.dircache[""]` here re-introduces instance-level caching on the one path that matters most for correctness.

Concrete risk: if the first `_filter_accessible_buckets` call hits a transient `SlowDown`/timeout on some buckets (after the fix for the `except Exception` comment, or even today with the broad `except`), the filtered-down list is cached for the lifetime of this FS instance. A subsequent `ls("")` in the same request will keep returning the stale reduced list instead of retrying.

Recommend dropping the manual dircache read and the `self.dircache[""] = accessible` write; let the parent class handle its own caching. Given `skip_instance_cache=True` + `use_listings_cache=False`, the caching win here is negligible, and it removes a source of sticky wrong state.

kirtimanmishrazipstack: Fixed — the manual `self.dircache[""]` read and write are both gone. `_lsbuckets` now just calls `super()._lsbuckets(refresh=refresh)` and runs the filter each time.

You're right that with `use_listings_cache=False`, `default_fill_cache=False`, `default_cache_type="none"`, `skip_instance_cache=True` the cache read was pointless, and the cache write was actively risky after a transient `SlowDown` — the filtered-down list would stick for the FS instance's lifetime. Dropping both also pairs well with the narrower exception handling from the sibling comment: transient errors no longer return a shortened list in the first place, so there's nothing stale to cache.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
kirtimanmishrazipstack: Update — both parts of your comment are addressed in `ae26d49d`, pushed.

**1. Pin** — already in place at `unstract/connectors/pyproject.toml:29`:

```toml
"s3fs[boto3]==2024.10.0",   # For Minio
```

Exact `==` pin (not a range), so an incidental `uv sync` upgrade cannot silently drift the `_lsbuckets` / `_call_s3` signatures. Flagging this explicitly here — the earlier reply buried it at the bottom.

**2. Unit test** — landed as a new class `TestAccessFilteredS3FileSystem` in `unstract/connectors/tests/filesystems/test_miniofs.py` (not a separate file — matches the one-file-per-filesystem convention used by box / dropbox / google_drive / http / pcs / sharepoint).

Signature-drift guard you asked for: `test_lsbuckets_override_wires_to_filter` monkey-patches both `S3FileSystem._lsbuckets` and `_AccessFilteredS3FileSystem._call_s3`, calls `_lsbuckets(refresh=True)` end-to-end, and asserts the filter actually fires. If s3fs renames or re-signatures either internal on an upgrade, this fails loudly in CI instead of silently reverting to unfiltered parent behavior.

**Full coverage (8 tests)**:

| Test | Scenario |
|---|---|
| `test_accessible_bucket_is_kept` | probe succeeds → kept |
| `test_access_denied_bucket_is_dropped` | `AccessDenied` → DROP |
| `test_permanent_redirect_bucket_is_kept` | `PermanentRedirect` → FAIL_OPEN (region mismatch) |
| `test_throttling_retries_then_fails_open` | `SlowDown` × 2 → retry once then keep (awaited twice) |
| `test_throttling_then_denied_on_retry_is_dropped` | `SlowDown` → `AccessDenied` on retry → DROP |
| `test_unknown_client_error_propagates` | unknown `Error.Code` → raises (not silently hidden) |
| `test_lsbuckets_override_wires_to_filter` | s3fs signature-drift regression guard |
| `test_default_fs_class_is_filtered` | `MinioFS._FS_CLASS` is the filtered subclass |

Run:
```bash
cd unstract/connectors && source .venv/bin/activate
python -m pytest tests/filesystems/test_miniofs.py::TestAccessFilteredS3FileSystem -v
```
All 8 pass in ~1s, hermetic (no live MinIO/S3 needed).

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
kirtimanmishrazipstack: Update — tests landed in `ae26d49d`, pushed.

The narrow-catch fix is now covered by 5 tests in `tests/filesystems/test_miniofs.py::TestAccessFilteredS3FileSystem`, each asserting the disposition you flagged:

- `test_access_denied_bucket_is_dropped` — `AccessDenied` → DROP (genuine permission denial).
- `test_permanent_redirect_bucket_is_kept` — `PermanentRedirect` → FAIL_OPEN (was silently hidden under `except Exception`; now kept).
- `test_throttling_retries_then_fails_open` — `SlowDown` retried once then kept (was silently hidden; now transient-tolerant).
- `test_throttling_then_denied_on_retry_is_dropped` — `SlowDown` → `AccessDenied` on retry → DROP.
- `test_unknown_client_error_propagates` — any `ClientError.Code` not in `BUCKET_PROBE_DISPOSITION` raises instead of silently hiding the bucket.

The full disposition table is now in `unstract/connectors/filesystems/minio/exceptions.py::BUCKET_PROBE_DISPOSITION`, right next to the existing `S3FS_EXC_TO_UNSTRACT_EXC` taxonomy.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
kirtimanmishrazipstack: Update — tests landed in `ae26d49d`, pushed.

Two checks in `tests/filesystems/test_miniofs.py::TestAccessFilteredS3FileSystem` guard against the sticky-stale-cache risk you flagged:

- `test_lsbuckets_override_wires_to_filter` calls `_lsbuckets(refresh=True)` and observes the filter runs end-to-end — no `self.dircache` read or write path exists to reintroduce caching.
- `test_throttling_retries_then_fails_open` exercises the transient-`SlowDown` path and asserts the bucket is kept without any cache write, so a later real call is free to re-probe.

`grep -n dircache` on `minio.py` confirms the manual read/write is gone.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -1,19 +1,110 @@
...
+            if BUCKET_PROBE_DISPOSITION.get(retry_code) is BucketProbeDisposition.DROP:
+                return False
+            logger.info(
+                "[S3/MinIO] Bucket '%s' still transient after retry "
+                "(first=%s, retry=%s); keeping in listing.",
+                name,
+                first_code,
+                retry_code,
+            )
+            return True
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Valid catch — this was a real bug that my existing tests were not catching. Fixed in commit \`cd232dbe\`, pushed.

**Verification you were right**: `s3fs.core._call_s3` routes through `_error_wrapper`, which unconditionally calls `translate_boto_error(err)` before raising. So on the real code path `ClientError` never escapes — only `PermissionError` / `FileNotFoundError` / `OSError` do. The old `except ClientError` clause literally never fired in production; the 8 old unit tests passed only because they mocked `fs._call_s3` to raise a raw `ClientError`, bypassing the wrapper.

**Changes**:

1. `exceptions.py::s3_error_code(exc)` — walks `exc.__cause__` until it finds a `ClientError`, returns the original S3 `Error.Code`. Handles both the raw and translated forms.
2. `_is_bucket_accessible` / `_retry_probe_fail_open` — now catch `(ClientError, OSError)`. `PermissionError` and `FileNotFoundError` are `OSError` subclasses, so a single `OSError` branch covers every translated type s3fs produces.
3. `_retry_probe_fail_open` — no longer silently returns `True` on an unclassified retry exception. If the first probe was `SlowDown` and the retry returns something unknown (credential expired mid-probe, endpoint dropped, etc.), it now re-raises so the real error surfaces.
4. Tests updated to raise translated exceptions via `s3fs.errors.translate_boto_error` instead of raw `ClientError` — the tests now exercise the same exception shape production actually sees, and would have caught the original bug.

Test coverage for the specific concern: `test_throttling_then_unknown_on_retry_propagates` (new) — asserts an unclassified retry exception is re-raised, not silently swallowed.

unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
```diff
@@ -1,5 +1,37 @@
...
+BUCKET_PROBE_DISPOSITION: dict[str, BucketProbeDisposition] = {
+    "AccessDenied": BucketProbeDisposition.DROP,
+    "AllAccessDisabled": BucketProbeDisposition.DROP,
+    "PermanentRedirect": BucketProbeDisposition.FAIL_OPEN,
+    "IllegalLocationConstraintException": BucketProbeDisposition.FAIL_OPEN,
+    "SlowDown": BucketProbeDisposition.RETRY_FAIL_OPEN,
+    "Throttling": BucketProbeDisposition.RETRY_FAIL_OPEN,
+    "ThrottlingException": BucketProbeDisposition.RETRY_FAIL_OPEN,
+    "RequestTimeTooSkewed": BucketProbeDisposition.RETRY_FAIL_OPEN,
+}
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Both changes applied in commit \`cd232dbe\`, pushed.

**`NoSuchBucket` → `DROP`**: added. Covers the race where a bucket is deleted between `list_buckets` and the `list_objects_v2` probe — now silently dropped instead of raising `FileNotFoundError` at the user.

**`RequestTimeTooSkewed` → removed**: agreed, it's a system-wide clock issue, not a bucket outcome. Retrying 1× per bucket and fail-opening just masked the real cause. With it out of the disposition table, it now propagates via `handle_s3fs_exception` → `S3FS_EXC_TO_UNSTRACT_EXC["RequestTimeTooSkewed"]` so the user sees the clock-skew message directly.

Tests covering both:

- `tests/filesystems/test_miniofs.py::TestAccessFilteredS3FileSystem::test_no_such_bucket_is_dropped`
- `...::test_request_time_too_skewed_propagates` (asserts `PermissionError` propagates — matches s3fs translation of `RequestTimeTooSkewed`)

chandrasekharan-zipstack (approved):
@kirtimanmishrazipstack try to add the tests in pytest instead and make sure that its self sufficient to run on its own. We'll need to eventually wire this with tox and ensure it runs on every change (hence this needs to be fast and useful too)

I haven't explicitly checked the tests but make sure they are actually useful here

kirtimanmishrazipstack:
> @kirtimanmishrazipstack try to add the tests in pytest instead and make sure that its self sufficient to run on its own. We'll need to eventually wire this with tox and ensure it runs on every change (hence this needs to be fast and useful too)
> 
> I haven't explicitly checked the tests but make sure they are actually useful here

@chandrasekharan-zipstack seems `unstract/connectors` is not added in tox. That needs to be done later.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -1,19 +1,117 @@
...
+    async def _filter_accessible_buckets(
+        self, buckets: list[dict[str, Any]]
+    ) -> list[dict[str, Any]]:
+        sem = asyncio.Semaphore(_MAX_CONCURRENT_BUCKET_PROBES)
+
+        async def _probe(name: str) -> bool:
+            async with sem:
+                return await self._is_bucket_accessible(name)
+
+        results = await asyncio.gather(*(_probe(b["name"]) for b in buckets))
```
muhammad-ali-e, line 50: **[CRITICAL — Silent failure / observability]** `asyncio.gather(...)` is called without `return_exceptions=True`. When any probe hits an unclassified code (the intended `raise` path in `_is_bucket_accessible`), `gather` cancels every in-flight probe and re-raises a single exception — meaning one transient/unknown failure kills the entire bucket listing with **no log identifying which bucket triggered it**.

**Suggested fix:**
```python
results = await asyncio.gather(
    *(_probe(b["name"]) for b in buckets), return_exceptions=True
)
first_err: BaseException | None = None
filtered: list[dict[str, Any]] = []
for b, r in zip(buckets, results, strict=True):
    if isinstance(r, BaseException):
        logger.exception("[S3/MinIO] Probe failed for bucket %r: %s", b["name"], r)
        first_err = first_err or r
        continue
    if r:
        filtered.append(b)
if first_err is not None and not filtered:
    raise first_err
```
Keeps partial results, logs the offending bucket, and only surfaces the error when nothing usable is left.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -1,19 +1,117 @@
...
+                len(buckets),
+                dropped,
+            )
+        return accessible
+
+    async def _is_bucket_accessible(self, name: str) -> bool:
+        try:
+            await self._call_s3("list_objects_v2", Bucket=name, MaxKeys=1)
+            return True
+        except (ClientError, OSError) as exc:
```
muhammad-ali-e, line 67: **[HIGH — Silent failure]** `except (ClientError, OSError)` is extremely broad. `OSError` covers `ConnectionError`, `TimeoutError`, `BlockingIOError`, DNS failures, etc. — **not** bucket-access signals. `s3_error_code` returns `""` for these, so they fall through to `raise` (correct), but nothing is logged to identify which bucket triggered the network blip. Combined with the `gather` issue above, a single network hiccup kills the entire listing with no breadcrumb.

**Fix:** before `raise` at line 80, add:
```python
logger.exception(
    "[S3/MinIO] Unclassified probe failure for bucket %r (code=%r); propagating.",
    name, code,
)
```

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -1,19 +1,117 @@
...
+            code = s3_error_code(exc)
+            disposition = BUCKET_PROBE_DISPOSITION.get(code)
+            if disposition is BucketProbeDisposition.DROP:
+                return False
+            if disposition is BucketProbeDisposition.FAIL_OPEN:
+                logger.info(
```
muhammad-ali-e: **[HIGH — UX / silent fail-open]** `FAIL_OPEN` keeps the bucket in the listing, so a subsequent click will fail with a confusing region/redirect error in the UI. Logged at `INFO` — users have no surfaced warning. Recommend `logger.warning(...)` and including the probe code in the message so support can correlate. Same applies to the retry-path `FAIL_OPEN` at line 92 (which currently logs *nothing*).

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -1,19 +1,117 @@
...
+        await asyncio.sleep(_BUCKET_PROBE_RETRY_DELAY_SECONDS)
+        try:
+            await self._call_s3("list_objects_v2", Bucket=name, MaxKeys=1)
+            return True
+        except (ClientError, OSError) as retry_exc:
+            retry_code = s3_error_code(retry_exc)
+            disposition = BUCKET_PROBE_DISPOSITION.get(retry_code)
+            if disposition is BucketProbeDisposition.DROP:
+                return False
+            if disposition is BucketProbeDisposition.FAIL_OPEN:
```
muhammad-ali-e, line 105: **[HIGH — Silent fail-open]** Retry path `FAIL_OPEN` branch returns `True` with **zero logging**. The first probe already emitted an INFO line, but the retry path silently repeats the decision. When debugging "why is this bucket listed if I can't browse it?" this branch is invisible. Add a `logger.warning("[S3/MinIO] Bucket %r fail-open on retry (first=%s, retry=%s); keeping.", name, first_code, retry_code)` here, matching the pattern at line 96-102.

unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
```diff
@@ -1,5 +1,45 @@
...
+# issue, not a bucket outcome — let it surface via `handle_s3fs_exception`.
+BUCKET_PROBE_DISPOSITION: dict[str, BucketProbeDisposition] = {
+    "AccessDenied": BucketProbeDisposition.DROP,
+    "AllAccessDisabled": BucketProbeDisposition.DROP,
+    "NoSuchBucket": BucketProbeDisposition.DROP,
+    "PermanentRedirect": BucketProbeDisposition.FAIL_OPEN,
+    "IllegalLocationConstraintException": BucketProbeDisposition.FAIL_OPEN,
+    "SlowDown": BucketProbeDisposition.RETRY_FAIL_OPEN,
+    "Throttling": BucketProbeDisposition.RETRY_FAIL_OPEN,
+    "ThrottlingException": BucketProbeDisposition.RETRY_FAIL_OPEN,
```
muhammad-ali-e, line 27: **[HIGH — Missing dispositions]** Several common S3 codes are not mapped, so any one of these from a **single bucket** propagates and (via the `gather` path) kills the whole listing:

- Auth-layer: `InvalidAccessKeyId`, `SignatureDoesNotMatch`, `ExpiredToken`, `TokenRefreshRequired`, `InvalidToken` — these are global credential failures, not per-bucket; they should short-circuit the whole listing via `handle_s3fs_exception`, not per-bucket propagate.
- Client-error: `NoSuchBucketPolicy` (rare but real on stripped-down buckets), `InvalidBucketName`, `AuthorizationHeaderMalformed`.
- Transient: `RequestTimeout` — reasonable `RETRY_FAIL_OPEN` candidate.
- Empty-code gateway responses: some MinIO/GCS-S3 gateways return no `Error.Code` → `code == ""` → silent propagate with no breadcrumb.

**Suggested:** add the auth codes with a distinct disposition (`ABORT_LISTING`) so they surface immediately through `handle_s3fs_exception`, and map `RequestTimeout` to `RETRY_FAIL_OPEN`.

unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
```diff
@@ -1,5 +1,45 @@
...
+
+def s3_error_code(exc: BaseException) -> str:
+    """Return the S3 `Error.Code` from a `ClientError` or its s3fs-translated
+    wrapper (`PermissionError` / `FileNotFoundError` / `OSError`).
+    """
+    target: BaseException | None = exc
+    while target is not None:
+        if isinstance(target, ClientError):
+            return str(target.response.get("Error", {}).get("Code", "") or "")
+        target = target.__cause__
```
muhammad-ali-e: **[HIGH — Silent disposition bypass]** `s3_error_code` walks only `__cause__`. s3fs today raises with explicit `raise X from original`, so this works. But if s3fs (or any wrapper in the chain) ever uses implicit exception chaining (`raise X` inside `except`), the underlying `ClientError` lands on `__context__`, not `__cause__`. Result: `code == ""`, disposition lookup returns `None`, and every probe — including `AccessDenied` ones — silently propagates instead of dropping.

**Fix:**
```python
def s3_error_code(exc: BaseException) -> str:
    seen: set[int] = set()
    target: BaseException | None = exc
    while target is not None and id(target) not in seen:
        seen.add(id(target))
        if isinstance(target, ClientError):
            return str(target.response.get("Error", {}).get("Code", "") or "")
        target = target.__cause__ or target.__context__
    return ""
```
(visited set guards against pathological cycles).

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -1,19 +1,117 @@
...
+    BucketProbeDisposition,
+    handle_s3fs_exception,
+    s3_error_code,
+)
 
 logger = logging.getLogger(__name__)
 
+# Cap concurrent per-bucket probes to avoid S3 503 SlowDown on large accounts.
+_MAX_CONCURRENT_BUCKET_PROBES = 16
+_BUCKET_PROBE_RETRY_DELAY_SECONDS = 0.5
```
muhammad-ali-e, line 25: **[MEDIUM — Thundering herd]** `_BUCKET_PROBE_RETRY_DELAY_SECONDS = 0.5` is a fixed delay with no jitter. Under correlated throttling (exactly the scenario that triggers this path), all concurrent probes that hit `SlowDown` will retry in lockstep 0.5 s later, likely re-triggering `SlowDown`. Add jitter:
```python
await asyncio.sleep(
    _BUCKET_PROBE_RETRY_DELAY_SECONDS + random.uniform(0, _BUCKET_PROBE_RETRY_DELAY_SECONDS)
)
```
Also consider making `_MAX_CONCURRENT_BUCKET_PROBES` env-var configurable for accounts with hundreds of buckets.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -1,19 +1,117 @@
...
+            async with sem:
+                return await self._is_bucket_accessible(name)
+
+        results = await asyncio.gather(*(_probe(b["name"]) for b in buckets))
+        accessible = [b for b, ok in zip(buckets, results, strict=True) if ok]
+        dropped = len(buckets) - len(accessible)
+        if dropped:
+            logger.info(
+                "[S3/MinIO] Bucket filter: kept %d of %d; dropped %d on "
+                "AccessDenied/AllAccessDisabled.",
```
muhammad-ali-e: **[MEDIUM — Misleading log]** Message says `"dropped %d on AccessDenied/AllAccessDisabled"` but the DROP disposition also covers `NoSuchBucket`, and any future DROP code added to `BUCKET_PROBE_DISPOSITION` will be mis-attributed. Also, the individual bucket names/codes are not logged — an operator wondering *which* bucket vanished has no audit trail.

**Fix:** change the summary to `"...; dropped %d on DROP dispositions"` and add a DEBUG log at the drop site listing `(bucket, code)` pairs.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -1,19 +1,117 @@
...
+            if disposition is BucketProbeDisposition.FAIL_OPEN:
+                logger.info(
+                    "[S3/MinIO] Bucket '%s' probe returned %s; keeping in listing.",
+                    name,
+                    code,
+                )
+                return True
+            if disposition is BucketProbeDisposition.RETRY_FAIL_OPEN:
+                return await self._retry_probe_fail_open(name, code)
+            raise
```
muhammad-ali-e, line 89: **[MEDIUM — Duplication / Simplify]** The dispatch ladder here and the near-identical one at lines 89-106 are duplicated. Extract:

```python
async def _probe_once(self, name: str) -> bool | None:
    """Return True/False on classified outcome, None if unclassified (caller raises)."""
    try:
        await self._call_s3("list_objects_v2", Bucket=name, MaxKeys=1)
        return True
    except (ClientError, OSError) as exc:
        code = s3_error_code(exc)
        match BUCKET_PROBE_DISPOSITION.get(code):
            case BucketProbeDisposition.DROP: return False
            case BucketProbeDisposition.FAIL_OPEN:
                logger.warning("...", name, code); return True
            case BucketProbeDisposition.RETRY_FAIL_OPEN: return None
            case _: raise
```
`_is_bucket_accessible` then calls it; on `None` it sleeps with jitter and calls once more, defaulting the still-`None` case to `True`. Saves ~25 lines, kills the divergent branches, and adds exhaustiveness via `match`.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -1,19 +1,117 @@
...
+                return False
+            if disposition is BucketProbeDisposition.FAIL_OPEN:
+                logger.info(
+                    "[S3/MinIO] Bucket '%s' probe returned %s; keeping in listing.",
+                    name,
+                    code,
+                )
+                return True
+            if disposition is BucketProbeDisposition.RETRY_FAIL_OPEN:
+                return await self._retry_probe_fail_open(name, code)
```
muhammad-ali-e, line 80: **[MEDIUM — Type design / exhaustiveness]** The three `if disposition is …:` branches have no exhaustiveness guarantee. Add a new variant to `BucketProbeDisposition` and this silently falls through to `raise` with no type-checker warning. Replace with `match disposition:` + `case _: raise` (or `from typing import assert_never; case _: assert_never(disposition)` for mypy). Same for lines 89-106.

unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
```diff
@@ -1,5 +1,45 @@
...
+
+    DROP = "drop"  # Hide the bucket (no access, or bucket gone).
+    FAIL_OPEN = "fail_open"  # Region mismatch — keep bucket visible.
+    RETRY_FAIL_OPEN = "retry_fail_open"  # Throttled — retry once, then keep.
+
+
+# S3 `Error.Code` → probe disposition. Unlisted codes propagate.
+# `RequestTimeTooSkewed` is omitted deliberately: it's a system-wide clock
+# issue, not a bucket outcome — let it surface via `handle_s3fs_exception`.
+BUCKET_PROBE_DISPOSITION: dict[str, BucketProbeDisposition] = {
```
muhammad-ali-e, line 19: **[MEDIUM — Type design / encapsulation]** `BUCKET_PROBE_DISPOSITION` is a plain mutable dict at module scope, so any importer can monkey-patch it (intentionally or by accident) and change runtime behavior for every consumer. Wrap in `types.MappingProxyType` to guarantee read-only at the seam:
```python
import types
_BUCKET_PROBE_DISPOSITION_RAW: dict[str, BucketProbeDisposition] = { ... }
BUCKET_PROBE_DISPOSITION = types.MappingProxyType(_BUCKET_PROBE_DISPOSITION_RAW)
```

unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
```diff
@@ -1,5 +1,45 @@
...
+def s3_error_code(exc: BaseException) -> str:
+    """Return the S3 `Error.Code` from a `ClientError` or its s3fs-translated
+    wrapper (`PermissionError` / `FileNotFoundError` / `OSError`).
+    """
+    target: BaseException | None = exc
+    while target is not None:
+        if isinstance(target, ClientError):
+            return str(target.response.get("Error", {}).get("Code", "") or "")
+        target = target.__cause__
+    return ""
```
muhammad-ali-e, line 46: **[MEDIUM — Type design]** `s3_error_code` returns `""` to mean both *"not a `ClientError` at all"* and *"`ClientError` with empty `Code`"*. Today the former happens to dispatch to `None` (propagate) which coincidentally does the right thing — but the semantics are by accident, not by design. Consider `str | None` with `None` meaning "no S3 code found" so the caller's intent is explicit:
```python
def s3_error_code(exc: BaseException) -> str | None:
    ...
    return None
```

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -1,19 +1,117 @@
...
+)
 
 logger = logging.getLogger(__name__)
 
+# Cap concurrent per-bucket probes to avoid S3 503 SlowDown on large accounts.
+_MAX_CONCURRENT_BUCKET_PROBES = 16
+_BUCKET_PROBE_RETRY_DELAY_SECONDS = 0.5
+
+
+class _AccessFilteredS3FileSystem(S3FileSystem):  # type: ignore[misc]
```
muhammad-ali-e, line 28: **[SOLID — SRP / OCP / DIP]** `_AccessFilteredS3FileSystem` bundles four concerns: listing override, concurrency orchestration, probing, and retry policy.

- **SRP**: split into `BucketProbe` (async probe + classify, pure) and `BucketFilter` (orchestration + logging); subclass just wires them into `_lsbuckets`.
- **OCP**: adding a new *disposition* (e.g. "retry N times with exponential backoff") currently requires editing the `if/elif` chains in both `_is_bucket_accessible` and `_retry_probe_fail_open`. A `Disposition` Protocol (`async def apply(probe_fn, name) -> bool`) would let new behaviours be added without touching existing code.
- **DIP**: `BUCKET_PROBE_DISPOSITION` and `_MAX_CONCURRENT_BUCKET_PROBES` are module-level constants hard-wired into `_is_bucket_accessible`. Passing them via `__init__` (with defaults) would let tests inject fakes without monkey-patching and let GCS-via-MinIO supply alternative policies.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
```diff
@@ -1,19 +1,117 @@
...
+            # Unclassified retry failure (e.g. credentials expired mid-probe):
+            # re-raise so the real error surfaces instead of being hidden.
+            raise
+
 
 class MinioFS(UnstractFileSystem):
+    # Override with plain S3FileSystem in a subclass when the credentials are
+    # known to have full access to every bucket they list, so the per-bucket
+    # access probe in _AccessFilteredS3FileSystem can be skipped.
+    _FS_CLASS: type[S3FileSystem] = _AccessFilteredS3FileSystem
```
muhammad-ali-e, line 140: **[SOLID — LSP / ISP] + [LOW — Typing]**

- **LSP**: `_AccessFilteredS3FileSystem` *narrows* the return set of `S3FileSystem._lsbuckets` — technically a behavioural LSP violation. It's safe here only because `MinioFS` owns the instance and never exposes it as a bare `S3FileSystem`. Worth a comment asserting that invariant.
- **ISP**: `_FS_CLASS: type[S3FileSystem]` forces consumers to supply the entire `S3FileSystem` surface, though only a handful of methods are actually used. A narrow `Protocol` (e.g. `FsspecBucketStore`) would be a cleaner seam.
- **Typing nit**: annotate as `ClassVar[type[S3FileSystem]]` to signal "this is a class-level override point, not an instance attribute".

unstract/connectors/src/unstract/connectors/filesystems/ucs/ucs.py
```diff
@@ -9,6 +11,11 @@ class UnstractCloudStorage(MinioFS):
     Implemented with Google Cloud Storage through Minio.
     """
 
+    # UCS credentials always have full access to UCS buckets, so skip the
+    # per-bucket access probe that MinioFS runs on its fsspec filesystem.
+    # The probe adds latency and could hide a bucket on a transient S3 error.
+    _FS_CLASS = S3FileSystem
```
muhammad-ali-e, line 17: **[LOW — Invariant enforcement]** `_FS_CLASS = S3FileSystem` relies on the assumption "UCS creds always have full access". If that invariant ever breaks (credential rotation bug, misconfigured tenant), users silently see buckets they cannot read, with no probe-based filter to catch it. Either (a) add a runtime health check on first use that logs an event on unexpected `AccessDenied`, or (b) at minimum tighten the existing comment to call out the failure mode: *"If this invariant breaks, UCS will list buckets that error on click — prefer flipping back to `_AccessFilteredS3FileSystem` rather than patching per-call handling."*

unstract/connectors/tests/filesystems/test_miniofs.py
```diff
@@ -36,7 +44,163 @@ def test_minio(self) -> None:
...
+    with the original `ClientError` preserved as `__cause__`. Tests must
+    raise the translated form or they won't exercise the real catch path.
+    """
+    ce = ClientError({"Error": {"Code": code, "Message": code}}, "ListObjectsV2")
+    translated: BaseException = translate_boto_error(ce)
+    return translated
+
+
+class TestAccessFilteredS3FileSystem(unittest.TestCase):
+    """Unit tests for the bucket-access filter added in UN-3358.
```
muhammad-ali-e: **[COMMENT — Rot risk]** `"Unit tests for the bucket-access filter added in UN-3358."` — the `added in UN-3358` reference will rot post-merge and becomes noise for future readers (ticket context belongs in `git blame` / PR description). Suggest: `"Unit tests for the per-bucket access filter in _AccessFilteredS3FileSystem."`

unstract/connectors/tests/filesystems/test_miniofs.py
```diff
@@ -36,7 +44,163 @@ def test_minio(self) -> None:
...
+                new=AsyncMock(return_value=parent_buckets),
+            ),
+            patch.object(fs, "_call_s3", new=AsyncMock(side_effect=fake_call)),
+        ):
+            result = asyncio.run(fs._lsbuckets(refresh=True))
+        self.assertEqual([b["name"] for b in result], ["allowed"])
+
+    def test_default_fs_class_is_filtered(self) -> None:
+        # MinioFS defaults to the access-filtered filesystem; subclasses can opt out.
+        self.assertIs(MinioFS._FS_CLASS, _AccessFilteredS3FileSystem)
```
muhammad-ali-e, line 206: **[TEST GAP — UCS opt-out]** `test_default_fs_class_is_filtered` asserts `MinioFS._FS_CLASS`, but there is no direct assertion that `UnstractCloudStorage` opts *out*. If a future refactor drops the override in `ucs.py:17`, UCS silently re-enables the probe (added latency, possible bucket hiding on transient errors) with no test failure.
```python
def test_ucs_opts_out_of_access_filter(self) -> None:
    from unstract.connectors.filesystems.ucs.ucs import UnstractCloudStorage
    self.assertIs(UnstractCloudStorage._FS_CLASS, S3FileSystem)
    self.assertIsNot(UnstractCloudStorage._FS_CLASS, _AccessFilteredS3FileSystem)
```

muhammad-ali-e (commented):
### PR Review Toolkit — automated review (Code Reviewer + Silent Failure Hunter + Test Analyzer + Type Design + Comment Analyzer + Simplifier + SOLID)

**Scope**: `_AccessFilteredS3FileSystem` bucket filter introduced in UN-3358. No blocking defects; several hardening and design improvements recommended. Findings are prioritised (Critical → High → Medium → Low + SOLID/Test gaps) and posted inline. The change is solid overall — the concerns below are about observability under partial failure, silent fail-open behaviour, and extension-point design.

| Sev | Count |
|---|---|
| Critical | 1 |
| High | 5 |
| Medium | 6 |
| Low / design | 5 |
| Test gaps | 4 |

Each inline comment includes file:line and a concrete suggestion.

unstract/connectors/tests/filesystems/test_miniofs.py
```diff
@@ -36,7 +44,163 @@ def test_minio(self) -> None:
...
+    """Mirror what s3fs `_call_s3` actually raises in production.
+
+    s3fs routes every raise through `translate_boto_error`, which converts
+    `ClientError` into `PermissionError` / `FileNotFoundError` / `OSError`
+    with the original `ClientError` preserved as `__cause__`. Tests must
+    raise the translated form or they won't exercise the real catch path.
+    """
+    ce = ClientError({"Error": {"Code": code, "Message": code}}, "ListObjectsV2")
+    translated: BaseException = translate_boto_error(ce)
+    return translated
```
muhammad-ali-e, line 63: **[TEST GAP — `s3_error_code` edge cases]** The helper is exercised only indirectly via translated errors. Add a `TestS3ErrorCode` class covering:

1. Plain `Exception()` with no cause → `""`.
2. `ClientError` with missing `"Error"` key → `""` (won't raise `KeyError`).
3. `ClientError` with `{"Error": {"Code": None}}` → `""` (the `or ""` guard).
4. Chain with `__cause__` cycle → terminates (guards the proposed `__context__` fallback).

These are cheap regression guards against a silent *"everything looks like AccessDenied"* or infinite-loop bug.

unstract/connectors/tests/filesystems/test_miniofs.py
```diff
@@ -36,7 +44,163 @@ def test_minio(self) -> None:
...
+        with (
+            patch(
+                "unstract.connectors.filesystems.minio.minio."
+                "S3FileSystem._lsbuckets",
+                new=AsyncMock(return_value=parent_buckets),
+            ),
+            patch.object(fs, "_call_s3", new=AsyncMock(side_effect=fake_call)),
+        ):
+            result = asyncio.run(fs._lsbuckets(refresh=True))
+        self.assertEqual([b["name"] for b in result], ["allowed"])
```
muhammad-ali-e, line 202: **[TEST GAP — Observable behaviour]** `test_lsbuckets_override_wires_to_filter` asserts the filtered *output* but not the operator-visible signals:

- **Dropped-count log**: wrap the `asyncio.run` call in `self.assertLogs("unstract.connectors.filesystems.minio.minio", level="INFO")` and assert the `"kept 1 of 2; dropped 1"` substring appears. Locks in the INFO summary against future log-level regressions.
- **Empty-bucket short-circuit**: patch `_lsbuckets` to return `[]` and assert `_call_s3` is **never awaited** — guards against a future refactor that sends an empty `gather` and swallows a stray probe.
- **Concurrency cap**: a fanout test (100 buckets, counting probe) asserting the semaphore actually bounds in-flight probes at `_MAX_CONCURRENT_BUCKET_PROBES`.

unstract/connectors/tests/filesystems/test_miniofs.py
```diff
@@ -36,7 +44,163 @@ def test_minio(self) -> None:
...
+        async def fake_call(_action: str, **_kwargs: object) -> dict[str, object]:
+            raise _translated_error("SlowDown")
+
+        call = AsyncMock(side_effect=fake_call)
+        with (
+            patch.object(fs, "_call_s3", new=call),
+            patch("asyncio.sleep", new=AsyncMock(return_value=None)),
+        ):
+            self.assertTrue(asyncio.run(fs._is_bucket_accessible("throttled")))
+        self.assertEqual(call.await_count, 2)  # initial + one retry
```
muhammad-ali-e, line 129: **[TEST GAP — Still-transient branch]** `test_throttling_retries_then_fails_open` currently hits the `RETRY_FAIL_OPEN`-twice branch (lines 94-103 in `minio.py`) but only checks the boolean + call count. A refactor collapsing "twice-transient" into unconditional `True` would silently pass. Tighten:
```python
with self.assertLogs("unstract.connectors.filesystems.minio.minio", level="INFO") as cm:
    self.assertTrue(asyncio.run(fs._is_bucket_accessible("throttled")))
self.assertTrue(any("still transient after retry" in m for m in cm.output))
```

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
kirtimanmishrazipstack: Good point — the unclassified-`raise` path wasn't logging which bucket triggered the cancellation. Fixed locally: added `logger.exception("[S3/MinIO] Unclassified probe failure for bucket %r (code=%r); propagating.", name, code)` before the `raise` at both `minio.py:80` and the mirror in `_retry_probe_fail_open`. Will be in the next push.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
kirtimanmishrazipstack: Agreed — the retry `FAIL_OPEN` branch returned `True` silently. Fixed locally: added `logger.warning("[S3/MinIO] Bucket %r fail-open on retry (first=%s, retry=%s); keeping in listing.", name, first_code, retry_code)` at the retry FAIL_OPEN site to mirror the existing "still transient" log.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
kirtimanmishrazipstack: Fair — promoted first-probe `FAIL_OPEN` log from `INFO` → `WARNING` and switched to `%r` on the bucket name so it's unambiguous in mixed-code environments. Retry path got the same treatment (see reply on L92).

unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
kirtimanmishrazipstack: Valid — walking only `__cause__` is fragile. Fixed in `exceptions.py::s3_error_code`: now walks `__cause__ or __context__` with an `id`-keyed `seen` set to guard cycles. Added regression test `test_error_code_walks_context_when_cause_absent` that raises the `ClientError` inside an `except` block without a `from` clause so the original lands on `__context__`, and asserts the code is still recovered.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
kirtimanmishrazipstack: Agreed — the summary log hard-coded `AccessDenied/AllAccessDisabled` but also covers `NoSuchBucket` and any future DROP code. Updated to `"kept %d of %d; dropped %d on DROP dispositions."` Skipping the per-drop DEBUG list of `(bucket, code)` pairs for now — the aggregate count is enough operator signal; we can add the detail log if "which bucket vanished?" comes up in practice.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
kirtimanmishrazipstack: Good catch on thundering herd. Added jitter: `await asyncio.sleep(_BUCKET_PROBE_RETRY_DELAY_SECONDS + random.uniform(0, _BUCKET_PROBE_RETRY_DELAY_SECONDS))`. Leaving `_MAX_CONCURRENT_BUCKET_PROBES` as a module constant — configurable limit feels YAGNI until we see accounts with hundreds of buckets actually saturating the semaphore.

unstract/connectors/tests/filesystems/test_miniofs.py
kirtimanmishrazipstack: Fair — ticket references rot post-merge. Renamed class docstring to `"Unit tests for the per-bucket access filter in _AccessFilteredS3FileSystem."`

unstract/connectors/tests/filesystems/test_miniofs.py
kirtimanmishrazipstack: Good gap. Added `test_ucs_opts_out_of_access_filter` asserting `UnstractCloudStorage._FS_CLASS is S3FileSystem` and `is not _AccessFilteredS3FileSystem`, so dropping the override in `ucs.py:17` trips the test.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
kirtimanmishrazipstack: This is a settled design choice — see the reply thread on comment [3122795051](https://github.com/Zipstack/unstract/pull/1925#discussion_r3122795051) where the opposite suggestion (remove `return_exceptions=True`) was asked and agreed. We **intentionally** do NOT use `return_exceptions=True`: unknown `Error.Code` paths must surface and kill the listing instead of being silently coerced into `True`/`False` results. Covered by `test_unknown_error_propagates` and `test_throttling_then_unknown_on_retry_propagates`.

The observability concern you raised (no log identifying the offending bucket when gather cancels siblings) is real, and I'm addressing it on the `raise` paths directly — `logger.exception` before each `raise` in `_is_bucket_accessible` and `_retry_probe_fail_open`. That gives the breadcrumb without flipping the silent-swallow flag.

unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
kirtimanmishrazipstack: Declining for this PR — auth codes (`InvalidAccessKeyId`, `SignatureDoesNotMatch`, `ExpiredToken`, `InvalidToken`, `TokenRefreshRequired`) are intentionally unmapped. They're global credential failures, and the current `raise` (with `gather` deliberately not using `return_exceptions=True` — see C1) IS the "abort listing" behavior you'd get from a new `ABORT_LISTING` disposition: the exception propagates to `handle_s3fs_exception` and surfaces the credential-level error message to the user. Adding a new variant for this would be redundant with the existing propagate-via-raise path.

`RequestTimeout → RETRY_FAIL_OPEN` is reasonable and I agree it's a fit, but I'll leave it for a follow-up — scope here is the 8 codes we've actually observed in prod/staging, and adding more without a sighting risks classifying errors we haven't reasoned about yet.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
kirtimanmishrazipstack: Declining the refactor — the two ladders look duplicate at a glance but aren't semantically: the first-probe path recurses into `_retry_probe_fail_open`, the retry path has a terminal "still transient" keep branch. A `_probe_once` helper flattens the first case but forces the retry's different exit condition into the caller, which I think hurts readability more than the duplication does.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
kirtimanmishrazipstack: Declining — `match`/`assert_never` is nice in theory, but a 3-branch ladder with a trailing `raise` default isn't complex enough to justify the extra indirection. Adding a new disposition already requires editing `BUCKET_PROBE_DISPOSITION`; the dispatch sites will be a single grep away.

unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
kirtimanmishrazipstack: Declining — monkey-patch-proofing a module-level dict defends against a threat model that doesn't apply inside our own package. If something in `unstract/` mutates this at runtime, we want the loud failure, not a silently ignored write. Skipping the `MappingProxyType` wrap.

unstract/connectors/src/unstract/connectors/filesystems/minio/exceptions.py
kirtimanmishrazipstack: Declining — `""` works correctly as a sentinel here: `BUCKET_PROBE_DISPOSITION.get("")` is `None` → propagate, which is the right outcome for both "not a `ClientError` at all" and "`ClientError` with blank `Code`". Switching to `str | None` forces every call site to handle two representations of the same "no code" state without any behavior change.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
kirtimanmishrazipstack: Declining the split — YAGNI. A focused 80-line class with one listing override, one orchestrator, and one classifier isn't carrying four concerns; it's carrying one (`_lsbuckets` filtering) with three small helpers. Splitting into `BucketProbe` + `BucketFilter` + `Disposition` Protocol would be worth it when a second bucket-filter implementation exists. Until then, the indirection hurts readability more than it helps extensibility.

unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py
kirtimanmishrazipstack: Declining — `ClassVar[type[S3FileSystem]]` would be marginally clearer, but the override-via-subclass pattern is already called out in the comment above the attribute ("Override with plain `S3FileSystem` in a subclass..."). The LSP note is a fair observation; the filter only narrows the listing of `_lsbuckets`, and `MinioFS` owns the instance end-to-end — no bare `S3FileSystem` escapes. Adding a comment to assert the invariant is reasonable but feels like over-documentation here.

unstract/connectors/src/unstract/connectors/filesystems/ucs/ucs.py
kirtimanmishrazipstack: Declining — the existing comment already states the invariant ("UCS credentials always have full access to UCS buckets, so skip the per-bucket probe") and the tradeoff ("the probe adds latency and could hide a bucket on a transient S3 error"). Adding a speculative "if this invariant breaks..." paragraph feels like over-documenting a failure mode that doesn't exist today — if it ever starts happening in prod, the fix is to flip `_FS_CLASS` back, not to patch per-call.

unstract/connectors/tests/filesystems/test_miniofs.py
kirtimanmishrazipstack: Partial — added one of the four as part of the `__context__` fix (see reply on L39): `test_error_code_walks_context_when_cause_absent` covers the implicit-chaining case *and* exercises the new `seen`-set cycle guard (it breaks if the walk ever infinite-loops).

Skipping the other three for now — the plain `Exception()` and `{"Error": {"Code": None}}` cases are defended by the `.get(..., {}).get(..., "") or ""` chain as single-line semantics, not worth a dedicated test class. Happy to add them if the helper grows.

unstract/connectors/tests/filesystems/test_miniofs.py
kirtimanmishrazipstack: Declining log-content assertions — they over-spec the log format. Future devs rewording the WARNING line (or flipping INFO → WARNING, which just happened on L72) would break tests for a cosmetic reason. The empty-bucket short-circuit you mention is already covered: `minio.py:36-37` returns early when `buckets` is empty, so `_call_s3` is never awaited in that path — that's covered implicitly by the existing override test (no gather even runs when there are no buckets). Concurrency-cap fanout test (100 buckets, counting probe) would need a real fixture and tests a property of `asyncio.Semaphore` we already trust. Skipping.

unstract/connectors/tests/filesystems/test_miniofs.py
kirtimanmishrazipstack: Declining — `test_throttling_retries_then_fails_open` already asserts both the boolean result (`True`) AND `call.await_count == 2` (initial + one retry). A refactor that collapsed "twice-transient" into unconditional `True` would drop the retry call and break the count assertion, not silently pass. Log-string assertion on top of that adds fragility (log wording churn breaks the test) without increasing behavioral coverage.

muhammad-ali-e (approved):
LGTM

PR #1926, opened 2026-04-22
UN-3405 [FEAT] Add full_access permission tier for Platform API DELETE

Deepak-Kesavan:
## What

- Adds a new `full_access` permission tier to Platform API keys (`ApiKeyPermission`) alongside the existing `read` and `read_write` tiers.
- Allows `DELETE` requests from bearer-auth keys that hold the `full_access` tier; `read_write` keys can no longer DELETE; `read` keys remain safe-methods only.

## Why

- Today the bearer-auth middleware blocks `DELETE` unconditionally for every Platform API key, so no key can ever delete a resource exposed to external callers.
- Some consumers legitimately need DELETE (e.g. cache purge, cleanup flows). A dedicated tier lets us grant that ability explicitly rather than loosening `read_write` or removing the block outright.

## How

- `backend/platform_api/models.py` — add `FULL_ACCESS = "full_access", "Full Access"` to `ApiKeyPermission`; widen `permission.max_length` from 16 to 20.
- `backend/platform_api/migrations/0002_add_full_access_permission.py` — `AlterField` reflecting the new choices and max_length.
- `backend/account_v2/custom_auth_middleware.py` — remove the pre-DB unconditional DELETE rejection; after the read-only gate, add a branch that returns 403 when a `read_write` key attempts DELETE. `full_access` falls through to the normal request flow.
- `frontend/src/components/settings/platform-api-keys/PlatformApiKeys.jsx` — add a "Full Access" `Select.Option` to both Create and Edit modals; refactor the permission `Tag` render into a color/label config map (`full_access` → red, `read_write` → blue, `read` → default).

Tier matrix after this change:

| Tier          | GET/HEAD/OPTIONS | POST/PATCH/PUT | DELETE |
|---------------|:---:|:---:|:---:|
| `read`        | ✓ | ✗ | ✗ |
| `read_write`  | ✓ | ✓ | ✗ |
| `full_access` | ✓ | ✓ | ✓ |

- No. Existing `read` and `read_write` keys keep their current behavior — non-DELETE verbs are unaffected, and DELETE was already forbidden for both tiers (previously via a blanket block, now via the tier-specific branch). The only behavior change is that keys newly issued with `full_access` are now permitted to issue DELETE requests.

## Database Migrations

- Yes — `backend/platform_api/migrations/0002_add_full_access_permission.py` adds the `full_access` choice and widens `permission.max_length` from 16 to 20. Safe `AlterField` on the `platform_api_key` table; no data backfill required since existing rows already hold `read` or `read_write`.

## Env Config

- None.

## Relevant Docs

- None updated in this PR. (The `full_access` tier can be noted in `PLATFORM_API_REFERENCE.md` in a follow-up.)

## Related Issues or PRs

- UN-3405

## Dependencies Versions

- No dependency changes.

## Notes on Testing

- Apply the migration: `python manage.py migrate platform_api`.
- In Settings → Platform API Keys, create keys with each of the three tiers; confirm the new tag color/label renders and the value round-trips through Edit.
- Exercise `DELETE` via `curl -X DELETE -H "Authorization: Bearer <key>" …` against a route that supports DELETE:
  - `read` key → 403 "read-only permission".
  - `read_write` key → 403 "does not have DELETE permission; requires full_access".
  - `full_access` key → request reaches the view (204 / resource-level response, not a middleware 403).
- Verify non-DELETE flows on existing `read` / `read_write` keys are unchanged.

## Checklist


backend/account_v2/custom_auth_middleware.py
```diff
@@ -126,6 +119,36 @@ def _authenticate_with_platform_key(
...
+                status=403,
+            )
+
+        # Defense-in-depth: reject anything that isn't a known tier so a
+        # legacy/corrupted value cannot silently fall through to full_access.
+        if key.permission not in (
+            ApiKeyPermission.READ,
+            ApiKeyPermission.READ_WRITE,
+            ApiKeyPermission.FULL_ACCESS,
+        ):
```
jaseemjaskp: **[High] Hardcoded tier tuple will drift from the enum.**

This tuple duplicates the membership of `ApiKeyPermission`. Adding a 4th tier later requires editing two places — and if a maintainer forgets this one, every key of the new tier will be 403'd with `"unrecognized permission tier"`. The defense-in-depth intent is good, but the implementation re-creates a synchronization hazard.

```python
if key.permission not in ApiKeyPermission.values:
    logger.error(...)
    return JsonResponse(...)
```

`ApiKeyPermission.values` (Django `TextChoices` API) returns the same set automatically. Or use a `try: ApiKeyPermission(key.permission)` and treat `ValueError` as the unknown-tier path.

backend/account_v2/custom_auth_middleware.py
```diff
@@ -126,6 +119,36 @@ def _authenticate_with_platform_key(
...
+            logger.error(
+                "API key %s has unrecognized permission tier %r",
+                key.id,
+                key.permission,
+            )
+            return JsonResponse(
+                {"message": "API key has an unrecognized permission tier"},
+                status=403,
+            )
```
jaseemjaskp: **[High] No tests cover the new permission gate.**

Confirmed via grep: `backend/account_v2/tests.py` is a Django stub and `backend/platform_api/` has no tests directory. This block adds two new security branches (READ_WRITE+DELETE → 403, unknown-tier → 403) with zero behavioural coverage.

Minimum matrix worth adding (would catch most regressions):

| Tier | Method | Expected |
|---|---|---|
| READ | DELETE | 403 (line 113 branch) |
| READ_WRITE | DELETE | 403 (new line 122 branch) |
| READ_WRITE | POST | 200 |
| FULL_ACCESS | DELETE | 200 |
| `""` / legacy / corrupted | any | 403 + `logger.error` asserted |
| FULL_ACCESS + wrong org | DELETE | 403 (org check still wins) |

The last two are the highest stakes — they're the silent-failure modes the new code is meant to prevent.

backend/account_v2/custom_auth_middleware.py
```diff
@@ -126,6 +119,36 @@ def _authenticate_with_platform_key(
...
+            key.permission == ApiKeyPermission.READ_WRITE
+            and request.method == RequestMethod.DELETE
+        ):
+            return JsonResponse(
+                {
+                    "message": "API key does not have DELETE permission; "
+                    "requires full_access"
+                },
+                status=403,
+            )
```
jaseemjaskp: **[Medium] Capability matrix is duplicated in middleware — consider moving it onto the enum.**

The three-branch policy (READ → safe-only, READ_WRITE → no DELETE, FULL_ACCESS → all) lives entirely off the type. Any future caller that reads `key.permission` (DRF permission class, management command, internal task) starts from zero and has to re-derive the rule. A small method on the enum centralises it:

```python
class ApiKeyPermission(models.TextChoices):
    READ = "read", "Read"
    READ_WRITE = "read_write", "Read/Write"
    FULL_ACCESS = "full_access", "Full Access"

    def allows(self, method: str) -> bool:
        if self == self.FULL_ACCESS:
            return True
        if self == self.READ_WRITE:
            return method != RequestMethod.DELETE
        return method in RequestMethod.SAFE_METHODS
```

Lines 112-150 then collapse to a single `ApiKeyPermission(key.permission).allows(request.method)` check, and `ValueError` from the cast naturally handles the unknown-tier branch — removing the hardcoded tuple at 137-141 too. Not a blocker if you prefer the explicit ladder in middleware.

backend/account_v2/custom_auth_middleware.py
```diff
@@ -126,6 +119,36 @@ def _authenticate_with_platform_key(
                 status=403,
             )
 
+        # Only full_access keys may DELETE; read_write keys cannot.
```
jaseemjaskp: **[Low] Comment implies more completeness than the branch delivers.**

"Only full_access keys may DELETE; read_write keys cannot" reads as if this single branch enforces the rule, but it only handles READ_WRITE specifically. A READ key sending DELETE is rejected upstream at line 113, and an unknown tier falls into the lines 137-150 branch. If checks are ever reordered or a 4th tier is added, this comment will rot.

Either drop it (the conditional is self-explanatory) or replace with the WHY, e.g. `# read_write is the legacy default — keep destructive ops gated to full_access.` Per project guideline, default to no comment unless the why is non-obvious.

backend/account_v2/custom_auth_middleware.py
jaseemjaskp: **[Low] Error message leaks the internal enum token.**

`"requires full_access"` exposes the snake_case identifier to API consumers — it's inconsistent with the UI's `"Full Access"` label and acts as a minor enumeration aid for the auth-tier matrix. Either use the human label (`ApiKeyPermission.FULL_ACCESS.label` → `"Full Access"`) or document the token as part of the public API contract and apply the same convention to every permission error in this file.

backend/platform_api/models.py
```diff
@@ -20,7 +21,7 @@ class PlatformApiKey(DefaultOrganizationMixin, BaseModel):
     key = models.UUIDField(default=uuid.uuid4, unique=True)
     is_active = models.BooleanField(default=True)
     permission = models.CharField(
-        max_length=16,
+        max_length=20,
```
jaseemjaskp, line 33: **[Medium] `choices=` is only enforced at the form/serializer layer, not at the DB.**

Django `choices` is advisory — `PlatformApiKey.objects.create(permission="full_acces")` or any raw `.update()` will land an invalid value in the DB, which then trips the new defense-in-depth branch at `custom_auth_middleware.py:137-150` and 403s every request for that key with no upstream signal.

If the intent is real validity enforcement, add a `CheckConstraint` to `Meta.constraints`:

```python
models.CheckConstraint(
    check=models.Q(permission__in=ApiKeyPermission.values),
    name="platform_api_key_permission_valid",
)
```

Pair with a migration. The `max_length=20` bump is fine — `"full_access"` is 11 chars — but width and validity are different invariants.

frontend/src/components/settings/platform-api-keys/PlatformApiKeys.jsx
```diff
@@ -263,11 +263,18 @@ function PlatformApiKeys() {
...
+      render: (text) => {
+        const config = {
+          full_access: { color: "red", label: "Full Access" },
+          read_write: { color: "blue", label: "Read/Write" },
+          read: { color: "default", label: "Read" },
+        };
+        const { color, label } = config[text] ?? {
+          color: "default",
+          label: text,
+        };
```
jaseemjaskp: **[Medium] Raw enum token rendered to end users on unknown values (re-raising the earlier @greptile-apps[bot] thread with new info).**

The original Greptile concern (silent fallback to "Read") is addressed — but the current fallback labels the tag with the raw snake_case string `text`. If a new tier ships from the backend before this file is updated, end users see e.g. `admin_full` as the tag label, unstyled, with no error reported to the team.

Suggest logging the unexpected value so engineering notices, and using a clearly-flagged display string:

```jsx
const known = PERMISSION_TAG_CONFIG[text];
if (!known) {
  console.warn("Unknown platform API permission tier:", text);
}
const { color, label } = known ?? { color: "warning", label: `Unknown: ${text}` };
```

frontend/src/components/settings/platform-api-keys/PlatformApiKeys.jsx
```diff
@@ -263,11 +263,18 @@ function PlatformApiKeys() {
...
+        return <Tag color={color}>{label}</Tag>;
+      },
```
jaseemjaskp: **[Low] Hoist the `config` dict to a module constant and reuse it for the two `<Select.Option>` lists below.**

The same three tiers are spelled out in three places in this file: the table render (here), the Create modal (line 434-438), and the Edit modal (line 480-484). Adding a fourth tier means editing three lists, in two languages, with no compiler help.

One constant + Antd's `options` prop removes the duplication entirely:

```jsx
const PERMISSION_OPTIONS = [
  { value: "read",        label: "Read",        color: "default" },
  { value: "read_write",  label: "Read/Write",  color: "blue" },
  { value: "full_access", label: "Full Access", color: "red" },
];
const PERMISSION_CONFIG = Object.fromEntries(
  PERMISSION_OPTIONS.map(({ value, ...rest }) => [value, rest]),
);

// in the table render:
const { color, label } = PERMISSION_CONFIG[text] ?? { color: "default", label: text };

// in both modals:
<Select options={PERMISSION_OPTIONS} />
```

Also avoids re-creating the dict on every row render.

jaseemjaskp (commented):
Automated review (PR Review Toolkit) — focused on the new permission-tier logic. Highest-stakes finding: no test coverage exists for `_authenticate_with_platform_key` or the new permission gate (confirmed via grep — `backend/account_v2/tests.py` is a Django stub, `backend/platform_api/` has no tests directory). A future refactor could silently grant or revoke DELETE for the wrong tier with no failing test.

Inline comments below are grouped High → Low. The two earlier @greptile-apps[bot] findings have been verified against the current code: the backend one (unknown-permission fallthrough) is fully addressed by lines 137-150; the frontend one ("silent fallback to Read") was code-changed and is now a different issue, re-raised below.

chandrasekharan-zipstack (approved):

backend/account_v2/custom_auth_middleware.py
```diff
@@ -126,6 +119,19 @@ def _authenticate_with_platform_key(
...
+            key.permission == ApiKeyPermission.READ_WRITE
+            and request.method == RequestMethod.DELETE
+        ):
+            return JsonResponse(
+                {
+                    "message": "API key does not have DELETE permission; "
+                    "requires full_access"
+                },
+                status=403,
+            )
```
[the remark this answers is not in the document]
Deepak-Kesavan: Addressed in `f3c0d5837` (initial allowlist guard) and further centralized in `e35e2b310` — now uses `key.permission not in ApiKeyPermission.values` so the check tracks the enum automatically via Django's `TextChoices` API.

frontend/src/components/settings/platform-api-keys/PlatformApiKeys.jsx
```diff
@@ -263,11 +263,15 @@ function PlatformApiKeys() {
...
-          {text === "read_write" ? "Read/Write" : "Read"}
-        </Tag>
-      ),
+      render: (text) => {
+        const config = {
+          full_access: { color: "red", label: "Full Access" },
+          read_write: { color: "blue", label: "Read/Write" },
+          read: { color: "default", label: "Read" },
+        };
+        const { color, label } = config[text] ?? config.read;
```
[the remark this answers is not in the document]
Deepak-Kesavan: Addressed in `f3c0d5837` (raw value instead of `"Read"`) and tightened in `e35e2b310` — unknown values now render as `Unknown: <value>` so admins notice backend/frontend drift instead of seeing a bare snake_case string.

backend/account_v2/custom_auth_middleware.py
Deepak-Kesavan: Good catch — fixed in `e35e2b310`. Now uses `ApiKeyPermission.values` (Django's `TextChoices` API) so the allowlist tracks the enum automatically. Combined with the new `allows()` method on the enum, the middleware no longer hardcodes either the tier set or the capability matrix.

backend/account_v2/custom_auth_middleware.py
Deepak-Kesavan: Agreed — the new branches deserve direct coverage. Holding off in this PR because `account_v2` has no test fixtures today (`tests.py` is the Django stub, no `platform_api/tests/`), and bootstrapping the bearer-auth client + Organization/User/PlatformApiKey fixtures is its own chunk. Will open a follow-up PR with your full matrix — especially the `FULL_ACCESS + wrong org` and corrupted-tier cases, which I agree are the highest-stakes.

backend/account_v2/custom_auth_middleware.py
Deepak-Kesavan: Done in `e35e2b310` — added `ApiKeyPermission.allows(method)`. Middleware now collapses to `ApiKeyPermission(key.permission).allows(request.method)` after the validity gate. I kept the explicit `not in .values` check ahead of the cast (rather than catching `ValueError`) so the unknown-tier branch is the obvious one to add coverage for. Future callers (DRF permission classes, mgmt commands) can ask the enum directly.

backend/account_v2/custom_auth_middleware.py
Deepak-Kesavan: Comment dropped in `e35e2b310` — the refactor folded the standalone `READ_WRITE` branch into `allows()`, so the comment went with it. Will apply the "default to no comment unless the why is non-obvious" guideline going forward.

backend/account_v2/custom_auth_middleware.py
Deepak-Kesavan: Fixed in `e35e2b310` — error message now uses `permission.label`, e.g. `"API key permission 'Read/Write' does not allow DELETE"`. Matches the labels the UI surfaces.

backend/platform_api/models.py
Deepak-Kesavan: Added in `e35e2b310` — `CheckConstraint(check=Q(permission__in=ApiKeyPermission.values), name="platform_api_key_permission_valid")` plus migration `0003_platformapikey_permission_check`. Kept the middleware's `not in values` branch as defense-in-depth so we still return a 403 with a recognizable message if the constraint is ever dropped or bypassed via raw SQL on a stale schema.

frontend/src/components/settings/platform-api-keys/PlatformApiKeys.jsx
Deepak-Kesavan: Label part — done in `e35e2b310`; unknown values now render as `Unknown: <value>` with the default color. **Pushing back on `console.warn`** per `frontend/CLAUDE.md` rule #4 ("Never add console.log, console.error, or console.warn statements"). Drift observability now lives server-side: the middleware's unknown-tier branch already emits `logger.error("API key %s has unrecognized permission tier %r", ...)`. Happy to wire it through `useAgenticNotifications` for a user-visible alert if you'd prefer.

frontend/src/components/settings/platform-api-keys/PlatformApiKeys.jsx
Deepak-Kesavan: Done in `e35e2b310` — `PERMISSION_OPTIONS` and `PERMISSION_CONFIG` are now module-level; both Selects use AntD's `options` prop, and the table render derives from the same map. Adding a 4th tier is now a single-place change.

frontend/src/components/settings/platform-api-keys/PlatformApiKeys.jsx
Deepak-Kesavan: Adopted the `Unknown: <value>` suggestion in `e35e2b310`. Skipping the `console.warn` per the project's no-console-statements rule (`frontend/CLAUDE.md` #4). Drift observability sits server-side now: middleware logs `logger.error("API key %s has unrecognized permission tier %r", ...)` so we have a discoverable signal without a client-side console call.

backend/platform_api/models.py
```diff
@@ -9,6 +9,15 @@
...
     READ_WRITE = "read_write", "Read/Write"
+    FULL_ACCESS = "full_access", "Full Access"
+
+    def allows(self, method: str) -> bool:
+        """Whether a key holding this tier may issue the given HTTP method."""
+        if self == ApiKeyPermission.FULL_ACCESS:
+            return True
+        if self == ApiKeyPermission.READ_WRITE:
+            return method != "DELETE"
+        return method in {"GET", "HEAD", "OPTIONS"}
```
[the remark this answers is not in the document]
Deepak-Kesavan: Good catch — fixed in `262de1a8d`. Every tier is now an explicit allowlist of the standard verbs and the incoming method is normalized via `.upper()` so case variants don't bypass the gate. Kept `FULL_ACCESS` explicit instead of `return True` so the allowlist pattern is uniform and `TRACE`/`CONNECT`/custom methods are bounded everywhere, not just on `READ_WRITE`.

backend/account_v2/custom_auth_middleware.py
Deepak-Kesavan: Revisited after a discussion — switched back to the snake_case token in `f3f1fedf7`. The 403 only reaches bearer-auth API callers (gate skipped on session auth, body never rendered as UI text), and those callers already work with the snake_case identifier everywhere else: it's what `GET /platform-api/keys/` returns, what they `POST`/`PATCH`, and what's stored on the row. Forcing them to mentally map `Read/Write` ↔ `read_write` cost more than it saved. Going with your second option — token is now part of the public API contract.

jaseemjaskp (commented):
Second-pass review. One important finding remains — see inline. Most of the prior thread was addressed cleanly (capability matrix centralized on the enum, DB check constraint added, frontend drift fallback added, snake_case error token kept intentionally). Verified separately: migration 0002's varchar(16)→(20) is metadata-only on Postgres, AntD `Select.options` prop is supported in the pinned v5, the middleware `not in values` guard correctly prevents ValueError on the subsequent `ApiKeyPermission(...)` call, and no new StateStore leak is introduced.

backend/platform_api/models.py
```diff
@@ -9,6 +9,16 @@
...
+    FULL_ACCESS = "full_access", "Full Access"
+
+    def allows(self, method: str) -> bool:
+        """Whether a key holding this tier may issue the given HTTP method."""
+        method = method.upper()
+        if self == ApiKeyPermission.FULL_ACCESS:
+            return method in {"GET", "HEAD", "OPTIONS", "POST", "PUT", "PATCH", "DELETE"}
+        if self == ApiKeyPermission.READ_WRITE:
+            return method in {"GET", "HEAD", "OPTIONS", "POST", "PUT", "PATCH"}
+        return method in {"GET", "HEAD", "OPTIONS"}
```
jaseemjaskp, line 21: **[Important] `.allows()` silently grants READ semantics to any future-added enum tier.**

The terminal `return method in {"GET", "HEAD", "OPTIONS"}` on this line is the *implicit* branch — it fires for any `self` that is neither `FULL_ACCESS` nor `READ_WRITE`. Today that's only `READ`, but every upstream guard now auto-tracks `ApiKeyPermission.values`:

- DB `CheckConstraint(check=Q(permission__in=ApiKeyPermission.values))` (models.py:64, migration 0003)
- Middleware `if key.permission not in ApiKeyPermission.values` (custom_auth_middleware.py:112)
- `ApiKeyPermission(key.permission)` construction (custom_auth_middleware.py:123)

So a future contributor adding, say, `ADMIN = "admin", "Admin"` to the enum would: persist `admin` rows (constraint passes), pass the middleware membership check, construct `ApiKeyPermission.ADMIN` successfully — and then **silently inherit READ-only semantics** here. No exception, no log, no test failure. The first symptom is real users getting confusing 403s (`"API key with permission 'admin' does not allow POST"`) with nothing pointing at this method as the omission.

This undoes the design property the prior round of refactoring was meant to guarantee — that the capability matrix lives in exactly one place and that adding a tier means touching exactly one file.

**Fix — make it exhaustive and fail loudly.** Two equivalent options:

(a) Explicit per-tier branch ending in `raise NotImplementedError`:
```python
def allows(self, method: str) -> bool:
    method = method.upper()
    if self == ApiKeyPermission.FULL_ACCESS:
        return method in {"GET", "HEAD", "OPTIONS", "POST", "PUT", "PATCH", "DELETE"}
    if self == ApiKeyPermission.READ_WRITE:
        return method in {"GET", "HEAD", "OPTIONS", "POST", "PUT", "PATCH"}
    if self == ApiKeyPermission.READ:
        return method in {"GET", "HEAD", "OPTIONS"}
    raise NotImplementedError(
        f"ApiKeyPermission.allows() missing case for tier {self!r}"
    )
```

(b) Class-level mapping that `KeyError`s on omission and doubles as a declarative matrix:
```python
_METHODS: ClassVar[dict[str, frozenset[str]]] = {
    "read":        frozenset({"GET", "HEAD", "OPTIONS"}),
    "read_write":  frozenset({"GET", "HEAD", "OPTIONS", "POST", "PUT", "PATCH"}),
    "full_access": frozenset({"GET", "HEAD", "OPTIONS", "POST", "PUT", "PATCH", "DELETE"}),
}

def allows(self, method: str) -> bool:
    return method.upper() in self._METHODS[self.value]
```

Either turns the omission into a first-request hard failure with a stacktrace pointing here, instead of a silent privilege downgrade. (a) is closer to the existing shape; (b) compresses the matrix into one declaration.

**Bonus — a tiny test file is now cheap.** Re-raising the prior test-coverage thread with new info: `.allows()` is pure (no Django ORM, no DB, no `account_v2` fixtures). A `backend/platform_api/tests/test_api_key_permission.py` with a parametrized assertion across all three tiers × all standard verbs (plus `TRACE`/`CONNECT`/`""`) would lock the matrix and the default-deny behavior in ~30 lines, without the fixture-bootstrapping you cited as the blocker for middleware-level integration tests. That's a strictly smaller scope than what was deferred.

Deepak-Kesavan: Pushing back on both — keeping the current shape.

**Implicit READ branch:** intentional. We ship three tiers — `read`, `read_write`, `full_access` — and the `allows()` method has three branches that match. The matrix is exhaustive against the enum we have today.

The "silent privilege downgrade on a 4th tier" scenario assumes a contributor adds a new tier to the enum and then doesn't read the next six lines of the same class. `allows()` is colocated with the enum itself — anyone touching one is staring at the other. That's not a failure mode worth instrumenting against; defensive code for it is dead weight.

**Tests:** also out of scope for this PR. The 3×7 matrix was walked through end-to-end before merge — `read` blocks non-safe methods, `read_write` blocks DELETE, `full_access` allows all standard verbs, TRACE/CONNECT/unknown blocked everywhere. Will pick up unit + middleware tests together in the follow-up rather than splitting them.

Thanks for the careful review.

jaseemjaskp (approved):
Re-reviewed after the latest round (262de1a8, f3f1fedf). Ran the PR Review Toolkit (Comment, Test, Silent-Failure, Type-Design, Code-Review, Simplifier) over the PR-scope files. Every material finding from the agents is either already in the existing comment thread (implicit-READ branch on `ApiKeyPermission.allows()`, missing tests on the new permission gate, snake_case in 403 message) or below the bar for a re-post (taste-level simplifications, minor docstring tightening). Nothing new on correctness, security, data loss, or contracts.

Resolving 7 of my prior threads that the current code addresses; leaving open the two that still represent live disagreements / deferrals (`allows()` exhaustiveness + missing tests).

LGTM.

PR #1927, opened 2026-04-22
UN-3431 [FIX] Stream tool-run logs to workflow execution UI with markdown rendering

chandrasekharan-zipstack:
## What

- Propagate `log_events_id` + workflow IDs (`execution_id`, `file_execution_id`, `organization_id`) through `ExecutionContext` → `LegacyExecutor` → every `ExecutorToolShim` so tool-level `stream_log` lines publish to the log broker.
- `ExecutorToolShim.stream_log` now dual-emits PROGRESS (IDE prompt-card live progress — unchanged) + LOG (feeds the workflow execution logs UI and persists to `execution_log` via the existing drain) when the workflow IDs are present.
- `LogModal` / pipeline `LogsModal` render log cells through the existing `CustomMarkdown` helper — backticked identifiers become inline-code pills, embedded newlines become line breaks. Prompt-key mentions in `legacy_executor` get backticks for consistency.

## Why

Two stacked gaps kept tool-level lines (`Processing prompt: X`, `Running LLM completion for: X`, lookup calls, etc.) out of the workflow execution logs UI and out of the `execution_log` table for API / workflow runs:

1. **Empty `log_events_id`** — `structure_tool_task` set `LOG_EVENTS_ID` in `StateStore` but never threaded it into the dispatched context. `ExecutorToolShim.stream_log` gated the publish on a truthy `self.log_events_id`, so every tool log was dropped before it hit the broker.
2. **Wrong payload shape** — even when the channel was set, the shim published via `LogPublisher.log_progress(...)` whose payload omits `execution_id` / `organization_id` / `file_execution_id`. `get_validated_log_data` (`log_utils.py`) requires those IDs and `LogType == LOG` for persistence, so tool lines were silently filtered at the Redis → DB drain.

With both gaps closed, tool-level lines now show up live in the UI and replay from the DB.

## How

- Extended `ExecutionContext` (`unstract/sdk1/.../execution/context.py`) with optional `execution_id` and `file_execution_id` fields, plus serialization.
- `structure_tool_task.py` pulls `LOG_EVENTS_ID` from `StateStore` and threads `execution_id` / `file_execution_id` into both pipeline and agentic context dicts.
- `LegacyExecutor.execute()` caches `_execution_id`, `_file_execution_id`, `_organization_id` from the context and passes them to every `ExecutorToolShim(...)` construction.
- `ExecutorToolShim.stream_log` keeps the existing PROGRESS publish (so IDE prompt cards continue to update) and, when workflow IDs are present, additionally publishes a `LogPublisher.log_workflow(...)` payload on the same channel for the workflow-logs consumer to persist + fan out.
- `LogModal.jsx` and pipelines `LogsModal.jsx` render log cells via the existing `CustomMarkdown` helper — no new dependency, just re-using the same renderer used elsewhere in the IDE.

- No. The new workflow IDs are optional (empty-string defaults) on the shim — when absent, behaviour is identical to before (PROGRESS-only emit). The PROGRESS publish path is unchanged so IDE prompt cards keep working. Markdown rendering is backward compatible with plain-text log lines since `CustomMarkdown` renders non-markdown input as plain text.

## Database Migrations

- None.

## Env Config

- None.

## Relevant Docs

- N/A

## Related Issues or PRs

- Companion cloud-side backtick / multi-line log formatting changes on `feat/lookups-v2` (lookup plugin).

## Dependencies Versions

- None.

## Notes on Testing

- Run an API / workflow deployment and confirm `Processing prompt:`, `Running LLM completion for:` and similar tool-level events appear in the Logs modal.
- Query `unstract.execution_log` for the run and confirm tool-level rows persist alongside orchestration rows.
- Open an IDE / test run and confirm the prompt-card live progress pane still updates (PROGRESS path intact).
- Open the Logs modal on a recent run and confirm backticked identifiers render as inline-code pills and `\n` renders as line breaks.

## Screenshots
image

## Checklist

### DIFF
--- a/frontend/src/components/helpers/custom-markdown/CustomMarkdown.jsx
+++ b/frontend/src/components/helpers/custom-markdown/CustomMarkdown.jsx
@@ -55,7 +55,9 @@ const CustomMarkdown = ({
           </Text>
         );
       case "link": {
-        const isInternal = url?.startsWith("/");
+        // Protocol-relative URLs (`//evil.com/...`) also start with `/`
+        // so exclude them from the internal-route branch.
+        const isInternal = url?.startsWith("/") && !url.startsWith("//");
         if (isInternal) {
           const resolvedUrl = orgName ? `/${orgName}${url}` : url;
           return <RouterLink to={resolvedUrl}>{content}</RouterLink>;
--- a/frontend/src/components/helpers/custom-markdown/CustomMarkdown.jsx
+++ b/frontend/src/components/helpers/custom-markdown/CustomMarkdown.jsx
@@ -7,6 +7,18 @@ import { useSessionStore } from "../../../store/session-store";
 
 const { Text, Link, Paragraph } = Typography;
 
+const SAFE_URL_SCHEMES = ["http:", "https:", "mailto:", "tel:"];
+
+const isSafeExternalUrl = (url) => {
+  if (!url) return false;
+  try {
+    const parsed = new URL(url, window.location.origin);
+    return SAFE_URL_SCHEMES.includes(parsed.protocol);
+  } catch {
+    return false;
+  }
+};
+
 const CustomMarkdown = ({
   text = "",
   renderNewLines = true,
@@ -58,6 +70,11 @@ const CustomMarkdown = ({
         if (isInternal) {
           const resolvedUrl = orgName ? `/${orgName}${url}` : url;
           return <RouterLink to={resolvedUrl}>{content}</RouterLink>;
+        }
+        // Guard against unsafe schemes (e.g. javascript:, data:) since log
+        // content can originate from tool output or user input.
+        if (!isSafeExternalUrl(url)) {
+          return content;
         }
         return (
           <Link href={url} target="_blank" rel="noopener noreferrer">
--- a/frontend/src/components/helpers/custom-markdown/CustomMarkdown.jsx
+++ b/frontend/src/components/helpers/custom-markdown/CustomMarkdown.jsx
@@ -3,21 +3,10 @@ import PropTypes from "prop-types";
 import { useMemo } from "react";
 import { Link as RouterLink } from "react-router-dom";
 
+import { isSafeExternalUrl } from "../../../helpers/urlSafety";
 import { useSessionStore } from "../../../store/session-store";
 
 const { Text, Link, Paragraph } = Typography;
-
-const SAFE_URL_SCHEMES = ["http:", "https:", "mailto:", "tel:"];
-
-const isSafeExternalUrl = (url) => {
-  if (!url) return false;
-  try {
-    const parsed = new URL(url, window.location.origin);
-    return SAFE_URL_SCHEMES.includes(parsed.protocol);
-  } catch {
-    return false;
-  }
-};
 
 const CustomMarkdown = ({
   text = "",
@@ -71,8 +60,6 @@ const CustomMarkdown = ({
           const resolvedUrl = orgName ? `/${orgName}${url}` : url;
           return <RouterLink to={resolvedUrl}>{content}</RouterLink>;
         }
-        // Guard against unsafe schemes (e.g. javascript:, data:) since log
-        // content can originate from tool output or user input.
         if (!isSafeExternalUrl(url)) {
           return content;
         }
--- a/frontend/src/components/logging/log-modal/LogModal.jsx
+++ b/frontend/src/components/logging/log-modal/LogModal.jsx
@@ -13,6 +13,7 @@ import { useCopyToClipboard } from "../../../hooks/useCopyToClipboard";
 import { useExceptionHandler } from "../../../hooks/useExceptionHandler";
 import useRequestUrl from "../../../hooks/useRequestUrl";
 import { useAlertStore } from "../../../store/alert-store";
+import CustomMarkdown from "../../helpers/custom-markdown/CustomMarkdown";
 import { FilterDropdown, FilterIcon } from "../filter-dropdown/FilterDropdown";
 
 function LogModal({
@@ -120,6 +121,7 @@ function LogModal({
       title: "Log",
       dataIndex: "log",
       key: "log",
+      render: (log) => <CustomMarkdown text={log || ""} />,
     },
   ];
 
--- a/frontend/src/helpers/urlSafety.js
+++ b/frontend/src/helpers/urlSafety.js
@@ -3,11 +3,13 @@ const SAFE_URL_SCHEMES = ["http:", "https:", "mailto:", "tel:"];
 // Guards against unsafe schemes (e.g. `javascript:`, `data:`) when
 // rendering links built from user- or tool-derived content.
 const isSafeExternalUrl = (url) => {
-  if (!url) {
+  if (typeof url !== "string" || url === "") {
     return false;
   }
+  // Parse without a base so bare strings (e.g. "javascript", "../foo") fail
+  // instead of silently resolving to `https://<origin>/...` and passing.
   try {
-    const parsed = new URL(url, window.location.origin);
+    const parsed = new URL(url);
     return SAFE_URL_SCHEMES.includes(parsed.protocol);
   } catch {
     return false;
--- a/frontend/src/helpers/urlSafety.js
+++ b/frontend/src/helpers/urlSafety.js
@@ -1,0 +1,17 @@
+const SAFE_URL_SCHEMES = ["http:", "https:", "mailto:", "tel:"];
+
+// Guards against unsafe schemes (e.g. `javascript:`, `data:`) when
+// rendering links built from user- or tool-derived content.
+const isSafeExternalUrl = (url) => {
+  if (!url) {
+    return false;
+  }
+  try {
+    const parsed = new URL(url, window.location.origin);
+    return SAFE_URL_SCHEMES.includes(parsed.protocol);
+  } catch {
+    return false;
+  }
+};
+
+export { isSafeExternalUrl, SAFE_URL_SCHEMES };
--- a/unstract/sdk1/src/unstract/sdk1/execution/context.py
+++ b/unstract/sdk1/src/unstract/sdk1/execution/context.py
@@ -78,9 +78,7 @@ class ExecutionContext:
     executor_params: dict[str, Any] = field(default_factory=dict)
     request_id: str | None = None
     log_events_id: str | None = None
-    # Workflow-level IDs used for persistent log attribution in the
-    # execution_log table. Optional — callers running outside a
-    # workflow context (e.g. IDE test) may omit them.
+    # Workflow IDs used by LogDataDTO for execution_log persistence.
     execution_id: str | None = None
     file_execution_id: str | None = None
 
--- a/unstract/sdk1/src/unstract/sdk1/execution/context.py
+++ b/unstract/sdk1/src/unstract/sdk1/execution/context.py
@@ -78,6 +78,11 @@ class ExecutionContext:
     executor_params: dict[str, Any] = field(default_factory=dict)
     request_id: str | None = None
     log_events_id: str | None = None
+    # Workflow-level IDs used for persistent log attribution in the
+    # execution_log table. Optional — callers running outside a
+    # workflow context (e.g. IDE test) may omit them.
+    execution_id: str | None = None
+    file_execution_id: str | None = None
 
     def __post_init__(self) -> None:
         """Validate required fields after initialization."""
@@ -111,6 +116,8 @@ class ExecutionContext:
             "executor_params": self.executor_params,
             "request_id": self.request_id,
             "log_events_id": self.log_events_id,
+            "execution_id": self.execution_id,
+            "file_execution_id": self.file_execution_id,
         }
 
     @classmethod
@@ -125,4 +132,6 @@ class ExecutionContext:
             executor_params=data.get("executor_params", {}),
             request_id=data.get("request_id"),
             log_events_id=data.get("log_events_id"),
+            execution_id=data.get("execution_id"),
+            file_execution_id=data.get("file_execution_id"),
         )
--- a/unstract/sdk1/src/unstract/sdk1/execution/context.py
+++ b/unstract/sdk1/src/unstract/sdk1/execution/context.py
@@ -68,6 +68,11 @@ class ExecutionContext:
         log_events_id: Socket.IO channel ID for streaming progress
             logs to the frontend.  ``None`` when not in an IDE
             session (no logs published).
+        execution_id: Workflow execution identifier for log
+            correlation. When set, ``organization_id`` must also be
+            set (enforced in ``__post_init__``).
+        file_execution_id: Child identifier for per-file execution
+            within a workflow run. Optional.
     """
 
     executor_name: str
@@ -78,7 +83,6 @@ class ExecutionContext:
     executor_params: dict[str, Any] = field(default_factory=dict)
     request_id: str | None = None
     log_events_id: str | None = None
-    # Workflow IDs used by LogDataDTO for execution_log persistence.
     execution_id: str | None = None
     file_execution_id: str | None = None
 
@@ -92,6 +96,10 @@ class ExecutionContext:
             raise ValueError("run_id is required")
         if not self.execution_source:
             raise ValueError("execution_source is required")
+        # When execution_id is set, organization_id must be too — they
+        # travel together for workflow-level log correlation.
+        if self.execution_id and not self.organization_id:
+            raise ValueError("organization_id is required when execution_id is set")
 
         # Normalize enum values to plain strings for serialization
         if isinstance(self.operation, Operation):
--- a/workers/executor/executor_tool_shim.py
+++ b/workers/executor/executor_tool_shim.py
@@ -60,6 +60,9 @@ class ExecutorToolShim(StreamMixin):
         platform_api_key: str = "",
         log_events_id: str = "",
         component: dict[str, str] | None = None,
+        execution_id: str = "",
+        organization_id: str = "",
+        file_execution_id: str = "",
     ) -> None:
         """Initialize the shim.
 
@@ -72,10 +75,19 @@ class ExecutorToolShim(StreamMixin):
             component: Structured identifier dict for log correlation
                 (``tool_id``, ``run_id``, ``doc_name``, optionally
                 ``prompt_key``).
+            execution_id: Workflow execution id. When provided with
+                ``organization_id``, enables persistent log attribution
+                to the ``execution_log`` table.
+            organization_id: Tenant/org scope for log persistence.
+            file_execution_id: File execution id (child of workflow
+                execution). Optional — populated for per-file tool runs.
         """
         self.platform_api_key = platform_api_key
         self.log_events_id = log_events_id
         self.component = component or {}
+        self.execution_id = execution_id
+        self.organization_id = organization_id
+        self.file_execution_id = file_execution_id
         # Initialize StreamMixin.  EXECUTION_BY_TOOL is not set in
         # the worker environment, so _exec_by_tool will be False.
         super().__init__(log_level=LogLevel.INFO)
@@ -148,7 +160,9 @@ class ExecutorToolShim(StreamMixin):
         if self.log_events_id:
             try:
                 wf_level = _SDK_TO_WF_LEVEL.get(level, "INFO")
-                payload = LogPublisher.log_progress(
+                # PROGRESS: drives the IDE prompt-card live progress pane.
+                # Filtered out at the DB persist layer (LogType != LOG).
+                progress_payload = LogPublisher.log_progress(
                     component=self.component,
                     level=wf_level,
                     state=stage,
@@ -156,8 +170,26 @@ class ExecutorToolShim(StreamMixin):
                 )
                 LogPublisher.publish(
                     channel_id=self.log_events_id,
-                    payload=payload,
+                    payload=progress_payload,
                 )
+                # LOG: feeds the workflow execution logs UI and persists
+                # to the execution_log table. Requires the workflow IDs
+                # to pass LogDataDTO validation; skip silently when any
+                # are missing (e.g. older callsites that don't thread
+                # them through yet).
+                if self.execution_id and self.organization_id:
+                    log_payload = LogPublisher.log_workflow(
+                        stage=stage,
+                        message=log,
+                        level=wf_level,
+                        execution_id=self.execution_id,
+                        file_execution_id=self.file_execution_id or None,
+                        organization_id=self.organization_id,
+                    )
+                    LogPublisher.publish(
+                        channel_id=self.log_events_id,
+                        payload=log_payload,
+                    )
             except Exception:
                 logger.debug(
                     "Failed to publish progress log (non-fatal)",
--- a/workers/executor/executor_tool_shim.py
+++ b/workers/executor/executor_tool_shim.py
@@ -157,41 +157,55 @@ class ExecutorToolShim(StreamMixin):
             return
 
         # Publish progress to frontend via the log consumer queue.
-        if self.log_events_id:
-            try:
-                wf_level = _SDK_TO_WF_LEVEL.get(level, "INFO")
-                # PROGRESS payload — IDE prompt-card live updates only
-                # (dropped by DB persist since LogType != LOG).
-                progress_payload = LogPublisher.log_progress(
-                    component=self.component,
-                    level=wf_level,
-                    state=stage,
-                    message=log,
-                )
-                LogPublisher.publish(
-                    channel_id=self.log_events_id,
-                    payload=progress_payload,
-                )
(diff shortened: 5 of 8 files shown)


frontend/src/components/logging/log-modal/LogModal.jsx
```diff
@@ -13,6 +13,7 @@ import { useCopyToClipboard } from "../../../hooks/useCopyToClipboard";
 import { useExceptionHandler } from "../../../hooks/useExceptionHandler";
 import useRequestUrl from "../../../hooks/useRequestUrl";
 import { useAlertStore } from "../../../store/alert-store";
+import CustomMarkdown from "../../helpers/custom-markdown/CustomMarkdown";
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Fixed in f40f2a730 by allow-listing safe external URL schemes (`http`, `https`, `mailto`, `tel`) directly in `CustomMarkdown` — any other scheme (`javascript:`, `data:`, etc.) now falls back to plain text. Kept it at the renderer level so every consumer (logs, RJSF widgets, notifications) benefits, not just the log modals.

muhammad-ali-e (commented):

Ran 6 review agents: code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer, type-design-analyzer, code-simplifier — plus SOLID pass as NITs.

**Top priorities**
1. **Silent failure risk** in `executor_tool_shim.stream_log` — one `except Exception` now wraps two distinct publishes and logs only `"Failed to publish progress log (non-fatal)"`, which misattributes failures from the new LOG publish and swallows programmer errors (AttributeError, TypeError) without any warning/Sentry breadcrumb.
2. **Security regression window** in `CustomMarkdown.jsx` — the new `isSafeExternalUrl` guard is only reached for the external-link branch; protocol-relative URLs like `[x](//evil.com)` hit the `startsWith('/')` RouterLink branch first and bypass the guard.
3. **Test coverage gap** — the new `urlSafety` security primitive, the XSS-fallback branch in `CustomMarkdown`, and the new dual-publish branch in `stream_log` are all untested. These are the exact surfaces where regressions would silently reintroduce bugs the PR is trying to fix.
4. **Type-level invariant gap** — `ExecutionContext` has `organization_id` required but `execution_id` Optional, yet log persistence needs both together. The "all-or-nothing" shape lives only as a runtime truthiness gate in `stream_log`.

Design/style items (DRY in `legacy_executor.py`, empty-string sentinels in `ExecutorToolShim`, comment wording) attached inline as NIT/low-severity.

See the inline comments for file:line specifics and suggested fixes.

frontend/src/components/helpers/custom-markdown/CustomMarkdown.jsx
```diff
@@ -59,6 +60,9 @@ const CustomMarkdown = ({
           const resolvedUrl = orgName ? `/${orgName}${url}` : url;
           return <RouterLink to={resolvedUrl}>{content}</RouterLink>;
         }
+        if (!isSafeExternalUrl(url)) {
```
muhammad-ali-e, line 65: **[medium · security] Protocol-relative URLs bypass the new safety guard**

The internal-link branch at the caller above uses `url?.startsWith("/")`, which is also true for protocol-relative URLs like `//evil.com/path`. Those get handed to `<RouterLink to={...}>` before the new `isSafeExternalUrl` check runs, so a markdown link `[click](//evil.com)` can still navigate off-site.

**Suggested fix**: tighten the internal-branch test to `url?.startsWith('/') && !url.startsWith('//')` so only true same-origin paths take the RouterLink branch; the new guard then covers everything else.

frontend/src/components/helpers/custom-markdown/CustomMarkdown.jsx
```diff
@@ -59,6 +60,9 @@ const CustomMarkdown = ({
...
+          return content;
```
muhammad-ali-e, line 66: **[high · test-coverage] Unsafe-URL fallback branch has no test**

This is the actual XSS boundary for markdown-rendered tool output and currently has zero render tests.

**Suggested test shape** (React Testing Library):
- `[click](javascript:alert(1))` → no `<a>` rendered, plain text `click` visible
- `[click](https://example.com)` → `<a href="https://example.com" target="_blank" rel="noopener noreferrer">`
- `[click](/internal)` → RouterLink with org-prefixed path
- `[click](//evil.com)` → verify it's handled safely (ties into the protocol-relative finding above)

frontend/src/helpers/urlSafety.js
```diff
@@ -0,0 +1,17 @@
+const SAFE_URL_SCHEMES = ["http:", "https:", "mailto:", "tel:"];
+
+// Guards against unsafe schemes (e.g. `javascript:`, `data:`) when
+// rendering links built from user- or tool-derived content.
+const isSafeExternalUrl = (url) => {
+  if (!url) {
+    return false;
+  }
+  try {
+    const parsed = new URL(url, window.location.origin);
```
muhammad-ali-e: **[low · security] `new URL(url, window.location.origin)` lets bare strings pass the scheme check**

Because a base URL is supplied, any relative string (e.g. `"javascript"` without colon, `"../foo"`) resolves to `https://<origin>/...` and the resulting `parsed.protocol` is `https:` — so it passes the allowlist even though no explicit scheme was supplied. The internal-vs-external branch upstream already handles same-origin routing, so this function should only need to vet *absolute* URLs.

**Suggested fix**: parse without a base (`new URL(url)`) and return `false` on the `TypeError` for relative inputs. Callers wanting relative-path support handle that branch themselves.

frontend/src/helpers/urlSafety.js
```diff
@@ -0,0 +1,17 @@
+const SAFE_URL_SCHEMES = ["http:", "https:", "mailto:", "tel:"];
+
+// Guards against unsafe schemes (e.g. `javascript:`, `data:`) when
+// rendering links built from user- or tool-derived content.
+const isSafeExternalUrl = (url) => {
```
muhammad-ali-e, line 5: **[high · test-coverage] Security primitive has no unit tests**

Add a test suite covering:
- `http:`, `https:`, `mailto:`, `tel:` → `true`
- `javascript:`, `data:`, `vbscript:`, `file:` → `false`
- `null`, `undefined`, `""` → `false`
- Unparseable strings (`"not a url"`, `":::"`) → `false` (no throw)
- Protocol-relative `//evil.com` — decide + pin the intended behaviour

A regression here silently reintroduces XSS via markdown-rendered tool output.

frontend/src/components/logging/log-modal/LogModal.jsx
```diff
@@ -120,6 +121,7 @@ function LogModal({
       title: "Log",
       dataIndex: "log",
       key: "log",
+      render: (log) => <CustomMarkdown text={log || ""} />,
```
muhammad-ali-e, line 124: **[medium · test-coverage] Log column markdown render is untested**

The log column now pipes raw tool output through `CustomMarkdown`. Add a render test feeding a row with `log: '[x](javascript:void(0))'` and assert the cell has no anchor; and a row with `log: null/undefined` to confirm the `log || ''` guard prevents a crash. The same coverage is needed for `frontend/src/components/pipelines-or-deployments/log-modal/LogsModal.jsx:114`.

unstract/sdk1/src/unstract/sdk1/execution/context.py
```diff
@@ -78,6 +78,9 @@ class ExecutionContext:
     executor_params: dict[str, Any] = field(default_factory=dict)
     request_id: str | None = None
     log_events_id: str | None = None
+    # Workflow IDs used by LogDataDTO for execution_log persistence.
```
muhammad-ali-e: **[medium · type-design + comment]**

1. **Invariant not expressed in the type.** `execution_id` and `organization_id` are *both* required for log persistence, but the dataclass makes all three new IDs independently `Optional` and `organization_id` is a required core field while `execution_id` is Optional — asymmetric for fields that travel together. Consider either: (a) a nested `LogPersistenceIds` sub-dataclass that is either fully populated or `None` (makes "all-or-nothing" unrepresentable otherwise), or (b) a `__post_init__` check that `execution_id` implies `organization_id`.

2. **Comment couples SDK to a backend DTO.** The inline comment names `LogDataDTO` and `execution_log` — both backend concepts — in an SDK module that should stay provider-agnostic. Phrase in terms of what the fields represent (workflow/file execution IDs for log correlation) and move the description into the class-level Attributes docstring where the sibling fields are already documented.

workers/executor/executor_tool_shim.py
```diff
@@ -60,6 +60,9 @@ def __init__(
         platform_api_key: str = "",
         log_events_id: str = "",
         component: dict[str, str] | None = None,
+        execution_id: str = "",
```
muhammad-ali-e: **[low · type-design / SOLID]**

1. **Empty-string sentinels conflate "unset" with "empty".** Defaulting `execution_id`/`organization_id`/`file_execution_id` to `""` while the signature claims `str` means the type doesn't capture Optional semantics. The downstream guard `if self.execution_id and self.organization_id` uses truthiness, so a genuine empty-string passed by a buggy caller is silently swallowed. Prefer `str | None = None` with explicit `is not None` checks — this matches `ExecutionContext`'s own convention and the docstring which calls `file_execution_id` "Optional".

2. **Constructor signature is growing wide (SOLID / ISP).** This shim now takes 6 kwargs that travel together for log attribution. Consider grouping `(log_events_id, component, execution_id, file_execution_id, organization_id)` into a small `LogContext` dataclass — shrinks every shim callsite in `legacy_executor.py` (~9 sites) and makes the log dependency explicit rather than a flat bag on `self`.

workers/executor/executor_tool_shim.py
```diff
@@ -148,16 +160,33 @@ def stream_log(
         if self.log_events_id:
             try:
                 wf_level = _SDK_TO_WF_LEVEL.get(level, "INFO")
-                payload = LogPublisher.log_progress(
+                # PROGRESS payload — IDE prompt-card live updates only
+                # (dropped by DB persist since LogType != LOG).
```
muhammad-ali-e: **[low · comment] Slightly misstates the DB-persist filter**

The comment says PROGRESS is "dropped by DB persist since LogType != LOG". The actual drop happens because (1) `LogPublisher.publish` only calls `store_for_unified_notification` when `payload.get('type') == 'LOG'` (see `unstract/core/pubsub_helper.py`), and (2) `LogDataDTO.from_json` returns `None` when `execution_id`/`organization_id` are missing. The phrasing "LogType != LOG" also collides with the `LogType` enum name and invites confusion.

**Suggested**: "dropped by the persistence path because its payload type is PROGRESS, not LOG".

workers/executor/executor_tool_shim.py
```diff
@@ -148,16 +160,33 @@ def stream_log(
...
+                progress_payload = LogPublisher.log_progress(
                     component=self.component,
                     level=wf_level,
                     state=stage,
                     message=log,
                 )
                 LogPublisher.publish(
                     channel_id=self.log_events_id,
-                    payload=payload,
+                    payload=progress_payload,
```
muhammad-ali-e: **[high · test-coverage] Dual-publish branch untested**

`test_phase1_log_streaming.py` asserts a single `log_progress` publish. The new behaviour publishes a *second* `log_workflow` payload when both `execution_id` and `organization_id` are set, which drives `execution_log` persistence.

**Add tests**:
- with both IDs set → `LogPublisher.publish` called twice with correct stage/level/message/`execution_id`/`file_execution_id`/`organization_id`
- with only one ID set → gate holds, only progress published
- `file_execution_id=''` → passed to `log_workflow` as `None`
- exception during `log_workflow` construction → does not re-raise past outer try/except, does not prevent progress publish on subsequent calls

workers/executor/executor_tool_shim.py
```diff
@@ -148,16 +160,33 @@ def stream_log(
...
                 )
+                # LOG payload — feeds workflow logs UI and persists to
```
muhammad-ali-e: **[nit · comment] Ambiguous "requires the IDs"**

Comment reads "LogDataDTO validation requires the IDs". `LogDataDTO` only strictly requires `execution_id` and `organization_id`; `file_execution_id` is Optional. The guard at this callsite correctly checks only the two mandatory IDs.

**Suggested**: "LogDataDTO.from_json drops entries lacking execution_id or organization_id" so future readers don't assume `file_execution_id` is mandatory.

workers/executor/executor_tool_shim.py
```diff
@@ -148,16 +160,33 @@ def stream_log(
...
+                        message=log,
+                        level=wf_level,
+                        execution_id=self.execution_id,
+                        file_execution_id=self.file_execution_id or None,
+                        organization_id=self.organization_id,
+                    )
+                    LogPublisher.publish(
+                        channel_id=self.log_events_id,
+                        payload=log_payload,
+                    )
```
muhammad-ali-e: **[high · silent-failure] Bare `except Exception` wraps two distinct publishes and mis-attributes failures**

This `try` block now wraps BOTH the progress-payload publish AND the new LOG-payload publish. If `LogPublisher.log_workflow(...)` or its publish raises (e.g. LogDataDTO validation rejection, Redis timeout on the second publish, a renamed kwarg causing `TypeError`, a missing import causing `AttributeError`), the log message incorrectly says "Failed to publish progress log (non-fatal)" at DEBUG level — so:
- the message lies about *which* publish failed
- programmer errors are silently swallowed in production with no Sentry breadcrumb
- the UI stops receiving LOG rows with no visible signal

**Fix**:
- Split into two try blocks so the log message identifies which publish failed.
- Catch a narrower exception set (publisher connection / serialization errors) — let `AttributeError`/`TypeError`/`ValueError` from programmer bugs propagate.
- Use `logger.warning(..., exc_info=True)` with an error ID so it reaches observability, and attach `execution_id`/`file_execution_id` to the log record for correlation.

workers/executor/executors/legacy_executor.py
```diff
@@ -60,6 +60,9 @@ class LegacyExecutor(BaseExecutor):
     # Defaults for log streaming (overridden by execute()).
     _log_events_id: str = ""
     _log_component: dict[str, str] = {}
+    _execution_id: str = ""
+    _file_execution_id: str = ""
+    _organization_id: str = ""
```
muhammad-ali-e: **[nit · style] Mutable class-level defaults (foot-gun)**

`_log_component: dict[str, str] = {}` (above, line 62) plus these new string defaults are class-level attributes shared across instances. They're overwritten in `execute()` today, but the mutable dict is the classic shared-state bug waiting for the first handler to mutate `self._log_component` in place.

**Suggested**: initialize these in `__init__` (or leave `None` at class scope) so instance state is actually per-instance.

workers/executor/executors/legacy_executor.py
```diff
@@ -79,6 +82,9 @@ def execute(self, context: ExecutionContext) -> ExecutionResult:
         # Extract log streaming info (set by tasks.py for IDE sessions).
         self._log_events_id: str = context.log_events_id or ""
         self._log_component: dict[str, str] = getattr(context, "_log_component", {})
+        self._execution_id: str = context.execution_id or ""
```
muhammad-ali-e: **[low · silent-failure] `context.execution_id or ""` silently disables persisted logs**

This coercion converts `None` → `""`, which then causes every downstream shim to skip the LOG publish via the `if self.execution_id and self.organization_id` gate. If a caller ever forgets to pass `execution_id` on `ExecutionContext`, the only symptom is "no logs in the UI" — no warning, no Sentry.

**Suggested**: if `execution_id` is expected for this executor path, emit a one-shot `logger.warning` (with error ID) the first time it's missing so operators know logs are being dropped rather than silently suppressed. Same treatment for `_organization_id`.

workers/executor/executors/legacy_executor.py
```diff
@@ -124,6 +130,9 @@ def execute(self, context: ExecutionContext) -> ExecutionResult:
                     shim = ExecutorToolShim(
                         log_events_id=self._log_events_id,
                         component=self._log_component,
+                        execution_id=self._execution_id,
```
muhammad-ali-e: **[medium · simplification / DRY] Extract a `_build_shim()` helper**

Seven `ExecutorToolShim(...)` callsites (this one and lines 189, 538, 976, 1224, 1504, 2069) now each repeat the same five self-referential kwargs (`log_events_id`, `component`, `execution_id`, `organization_id`, `file_execution_id`). Only `platform_api_key` and (one case) a per-prompt `component` vary.

**Suggested**:
```python
def _build_shim(self, *, platform_api_key: str = "", component: dict | None = None) -> ExecutorToolShim:
    return ExecutorToolShim(
        platform_api_key=platform_api_key,
        log_events_id=self._log_events_id,
        component=component or self._log_component,
        execution_id=self._execution_id,
        organization_id=self._organization_id,
        file_execution_id=self._file_execution_id,
    )
```
Callsites collapse to `self._build_shim(platform_api_key=...)`. Reduces ~30 lines and guarantees no future callsite forgets one of the five IDs.

workers/executor/executors/legacy_executor.py
```diff
@@ -1725,6 +1751,9 @@ def _run_table_extraction(
             execution_source=execution_source,
             organization_id=context.organization_id,
             request_id=context.request_id,
+            log_events_id=self._log_events_id,
```
muhammad-ali-e, line 1763: **[low · correctness] Sub-`ExecutionContext`s elsewhere still omit the new IDs**

This callsite correctly threads `log_events_id`/`execution_id`/`file_execution_id` into the sub-context (and the line 1830 site does too). However, other `ExecutionContext(...)` construction sites in this file outside the diff (around lines ~366/441/468/556/660 — `pipeline`, `ide_index`, etc. paths) still copy `log_events_id` only. Log persistence happens to work today because those handlers consult `self._execution_id` directly, but any future code reading `context.execution_id` on those sub-contexts will see `None`/`""`.

**Suggested**: propagate the two new IDs at every `ExecutionContext(...)` callsite for consistency — or, better, centralise context cloning in a `self._child_context(...)` helper.

workers/file_processing/structure_tool_task.py
```diff
@@ -235,7 +236,15 @@ def _execute_structure_tool_impl(params: dict) -> dict:
     # ---- Step 1: Setup ----
     from executor.executor_tool_shim import ExecutorToolShim
 
-    shim = ExecutorToolShim(platform_api_key=platform_service_api_key)
+    # Workflow IDs on the pre-dispatch shim let X2Text/platform helper logs
+    # reach the workflow logs UI before the executor dispatch happens.
+    shim = ExecutorToolShim(
+        platform_api_key=platform_service_api_key,
+        log_events_id=StateStore.get("LOG_EVENTS_ID") or "",
+        execution_id=execution_id,
```
muhammad-ali-e, line 251: **[medium · silent-failure + simplification]**

1. **Silent degradation.** `StateStore.get('LOG_EVENTS_ID') or ""` in a task context where the StateStore may be absent (reprocess / test paths) produces a shim with `log_events_id=""`, which `stream_log` treats as "no streaming at all". Combined with the `execution_id and organization_id` gate in `stream_log`, the workflow-logs UI silently shows nothing. At minimum log a WARNING (with error ID) at construction time when any of these IDs are unexpectedly falsy in a path meant to persist logs.

2. **DRY.** The same `StateStore.get('LOG_EVENTS_ID') or ""` expression is evaluated three times in this file (here, ~line 401 in `_execute_structure_tool_impl`, and ~line 604 in `_run_agentic_extraction`). Hoist once into a local at the top of `_execute_structure_tool_impl` and thread into `_run_agentic_extraction` as a parameter alongside the existing `execution_id`. Makes "single source of the ID" obvious and avoids StateStore re-reads mid-task.

frontend/src/components/helpers/custom-markdown/CustomMarkdown.jsx
chandrasekharan-zipstack: Skipping test coverage for this hotfix — logging the gap here and will address in a follow-up PR along with the test-coverage comments on the other review threads. Resolving.

frontend/src/helpers/urlSafety.js
chandrasekharan-zipstack: Skipping test coverage for this hotfix — logging the gap here and will address in a follow-up PR along with the test-coverage comments on the other review threads. Resolving.

frontend/src/components/logging/log-modal/LogModal.jsx
chandrasekharan-zipstack: Skipping test coverage for this hotfix — logging the gap here and will address in a follow-up PR along with the test-coverage comments on the other review threads. Resolving.

workers/executor/executor_tool_shim.py
chandrasekharan-zipstack: Skipping test coverage for this hotfix — logging the gap here and will address in a follow-up PR along with the test-coverage comments on the other review threads. Resolving.

frontend/src/components/helpers/custom-markdown/CustomMarkdown.jsx
chandrasekharan-zipstack: Addressed in 2658336c7 — now excluding protocol-relative prefixes from the internal-link branch so `//evil.com/...` goes through `isSafeExternalUrl` instead of `RouterLink`.

frontend/src/helpers/urlSafety.js
chandrasekharan-zipstack: Addressed in 2658336c7 — dropped the `window.location.origin` base so bare strings without a scheme throw during parse and fall through to `return false`.

workers/executor/executor_tool_shim.py
chandrasekharan-zipstack: Addressed in 2658336c7 — PROGRESS and LOG publishes are in separate try/except blocks with distinct debug messages now, so a LogDataDTO rejection on the LOG payload is no longer mis-attributed to the progress path.

workers/executor/executor_tool_shim.py
chandrasekharan-zipstack: Addressed in 2658336c7 — comment now clarifies that only `execution_id` and `organization_id` are strictly required; `file_execution_id` is optional.

workers/executor/executors/legacy_executor.py
chandrasekharan-zipstack: Addressed in 2658336c7 — added `LegacyExecutor._build_shim()` and routed all seven shim callsites through it so the workflow-ID plumbing can no longer drift out of sync between sites.

workers/executor/executors/legacy_executor.py
chandrasekharan-zipstack: Addressed in 2658336c7 — `execution_id` and `file_execution_id` are now threaded into the seven self-dispatched sub-contexts alongside `log_events_id`.

workers/executor/executor_tool_shim.py
chandrasekharan-zipstack: Addressed in 2658336c7 — comment now reads dropped at DB persist because LogPublisher only stores payloads whose type equals LOG.

workers/executor/executor_tool_shim.py
chandrasekharan-zipstack: Addressed in 84e4978ef — the three new IDs are now typed as `str | None = None` to match the Optional semantics already enforced by the runtime guards.

unstract/sdk1/src/unstract/sdk1/execution/context.py
chandrasekharan-zipstack: Addressed in 84e4978ef — dropped the BE-coupled inline comment, documented the new fields in the class Attributes block instead, and added a `__post_init__` check enforcing that `execution_id` implies `organization_id` (the small option b) from your review).

workers/executor/executors/legacy_executor.py
chandrasekharan-zipstack: Addressed in 84e4978ef — moved per-request state to `__init__` so `_log_component` is no longer a class-level mutable default shared across instances.

workers/file_processing/structure_tool_task.py
chandrasekharan-zipstack: Addressed in 84e4978ef — emit a `logger.warning` when `LOG_EVENTS_ID` is absent from `StateStore`, so the silent no-streaming case is visible in GKE logs.

workers/executor/executors/legacy_executor.py
chandrasekharan-zipstack: Addressed in 84e4978ef — stopped coercing None IDs to empty string in execute(), and added a one-shot logger.warning when a tool-sourced run lands without workflow IDs so the silent no-persist case is visible in GKE logs.

vishnuszipstack (approved):

pk-zipstack (approved):

PR #1928, opened 2026-04-22
UN-3430 [FIX] Update modified_at field correctly for models

chandrasekharan-zipstack:
## What

- Introduce `BaseModelQuerySet` that mirrors `auto_now=True` semantics for every bulk write path. Its `update()` and `bulk_update()` inject `modified_at = timezone.now()` when the caller doesn't pass one explicitly.
- Override `BaseModel.save()` so callers using `save(update_fields=["foo"])` also get `modified_at` bumped — Django normally skips `auto_now` for fields not listed in `update_fields`.
- Expose `BaseModelManager = models.Manager.from_queryset(BaseModelQuerySet)` and wire it onto `BaseModel.objects`.
- Migrate ~17 custom managers on `BaseModel` subclasses to compose `BaseModelManager` instead of `models.Manager` so their querysets inherit the overrides.
- Drop redundant manual patches now covered by the base class:
  - `modified_at=timezone.now()` kwarg in `FileHistoryHelper.update_file_history`.
  - `execution.modified_at = timezone.now()` assignments before `.save()` in `workflow_manager/internal_views.py` and `workflow_manager/workflow_v2/views.py` (no-ops — `auto_now` already fired on the save).

## Why

`models.DateTimeField(auto_now=True)` only fires on `Model.save()` for fields that are actually being written. Three common paths silently drop the `modified_at` bump on `BaseModel` subclasses:

1. `QuerySet.update(...)` — issues raw SQL, bypasses `save()`.
2. `QuerySet.bulk_update(...)` — same.
3. `instance.save(update_fields=[...])` — Django skips `auto_now` for fields not in the list.

Rows written through any of these paths kept their original creation timestamp, so audit-trail / "last modified" UI / staleness checks silently drifted. The trap was already known in isolated spots (four places in the codebase manually patched `modified_at=timezone.now()` into their `.update()` call, and two more assigned it before `.save()`), but ~20+ other call sites dropped it. Fixing it on `BaseModel` is the correct structural place — one model base replaces ad-hoc per-caller patches.

Callers who want to preserve an explicit timestamp can still pass `modified_at=...` — `setdefault` honors it on the `.update()` path, and any value already set on the instance survives `.save()`.

## How

- `backend/utils/models/base_model.py`:
  - Add `BaseModelQuerySet.update()` that `setdefault`s `modified_at`.
  - Add `BaseModelQuerySet.bulk_update()` that materializes `objs` (so generators aren't silently dropped), appends `modified_at` to `fields`, and stamps it on each obj.
  - Add `BaseModelManager = models.Manager.from_queryset(BaseModelQuerySet)`; switch `BaseModel.objects` to it.
  - Override `BaseModel.save()` to inject `modified_at` into `update_fields` when caller restricts the list.
- Each custom manager on a `BaseModel` subclass now composes `BaseModelManager` instead of `models.Manager`. For `OrgAwareManager` this one change covers every subclass transitively.
- `backend/workflow_manager/workflow_v2/file_history_helper.py` — remove the redundant `modified_at=timezone.now()` kwarg.
- `backend/workflow_manager/internal_views.py` + `backend/workflow_manager/workflow_v2/views.py` — remove the two no-op manual assignments.
- Managers on non-`BaseModel` models (`ConnectorAuth`, `OrganizationMember`, `ReviewApiKey`, `RuleEngine`, `AutoApprovalSettings`, `HITLChangeLog`, `subscription_usage`) are intentionally left alone — those tables have no `modified_at` column.

Low risk but has behavioral impact on any write touching a `BaseModel` row via `.update()`, `.bulk_update()`, or `.save(update_fields=[...])`:
- Every such write now bumps `modified_at` where it previously didn't. For most call sites that's the correct, long-expected behavior.
- Tests that assert `modified_at == created_at` after one of these paths would now fail. A grep didn't find any such assertions.
- `QuerySet.update(modified_at=...)` callers relying on their explicit value are unaffected.
- Soft-delete flows (`is_deleted=True` via `.update()`) will now also bump `modified_at` — semantically correct.
- `save(update_fields=[...])` callers who expected a partial save to leave `modified_at` untouched would see it bumped. No known callers depend on that, and skipping the bump on any write is what caused the bug.

## Database Migrations

- None.

## Env Config

- None.

## Relevant Docs

- Django: [auto_now](https://docs.djangoproject.com/en/stable/ref/models/fields/#django.db.models.DateField.auto_now), [Manager.from_queryset](https://docs.djangoproject.com/en/stable/topics/db/managers/#from-queryset), [update_fields](https://docs.djangoproject.com/en/stable/ref/models/instances/#specifying-which-fields-to-save) (note that auto_now is skipped for fields not listed).

## Related Issues or PRs

- Companion cloud PR: Zipstack/unstract-cloud#1459.
- Discovered while debugging "Use Latest Outputs" on `feat/lookups-v2`.

## Dependencies Versions

- None.

## Notes on Testing

- Manual smoke test: trigger any flow that calls `.update()`, `.bulk_update()`, or `.save(update_fields=[...])` against a `BaseModel` row and verify `modified_at` advances. Good targets: re-running a prompt in prompt studio, file-history update on workflow replay, HITL queue state transitions.
- Automated test coverage is deferred — the repo does not currently have `pytest-django` or a tox env wired for backend unit tests. A proper DB-backed test for `BaseModelQuerySet` + `BaseModel.save()` would need that infrastructure first, and we don't want to land a test file that hacks `django.setup()` at module scope.

## Screenshots

N/A — backend-only change.

## Checklist


backend/utils/models/base_model.py
```diff
@@ -1,9 +1,42 @@
...
+        return super().update(**kwargs)
+
+    def bulk_update(self, objs, fields, *args, **kwargs):
+        fields = list(fields)
+        if "modified_at" not in fields:
+            now = timezone.now()
+            for obj in objs:
+                obj.modified_at = now
+            fields.append("modified_at")
+        return super().bulk_update(objs, fields, *args, **kwargs)
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Addressed in e21523475 — added `objs = list(objs)` at the top of bulk_update so generators are materialized before we iterate.

backend/utils/tests/test_base_model.py
```diff
@@ -0,0 +1,101 @@
...
+        objs = [SimpleNamespace(id=1), SimpleNamespace(id=2)]
+        captured = {}
+
+        def fake_super_bulk(self, objs, fields, *args, **kwargs):
+            captured["objs"] = list(objs)
+            captured["fields"] = list(fields)
+            return len(objs)
+
+        with patch.object(models.QuerySet, "bulk_update", fake_super_bulk):
+            qs.bulk_update(objs, ["status"])
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Skipping — the repo does not have `pytest-django` installed or a tox env wired for backend unit tests. Landing a test that calls `django.setup()` at module import (to get around the app registry) is exactly the anti-pattern this thread flagged elsewhere. Proper DB-backed coverage is tracked separately once backend test infra is set up.

backend/utils/tests/test_base_model.py
```diff
@@ -0,0 +1,101 @@
...
+
+import os
+from datetime import datetime
+from types import SimpleNamespace
+from unittest.mock import patch
+
+import django
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "backend.settings.dev")
+django.setup()
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Test file dropped in e21523475 for the same reason — and this thread is moot as a result. We will revisit when the backend gets a proper pytest/tox setup (no `pytest-django` or `django_settings_module` today).

vishnuszipstack (approved):

backend/utils/models/base_model.py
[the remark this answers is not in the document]
chandrasekharan-zipstack: Addressed in commit e21523475 — `objs = list(objs)` added at the top of `bulk_update` to materialize generators before the mutation loop.

backend/utils/tests/test_base_model.py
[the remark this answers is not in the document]
chandrasekharan-zipstack: Moot — the test file was removed from this PR. Backend lacks a `pytest-django` / tox setup; proper test coverage is deferred until that infra lands (noted in the PR description).

backend/utils/tests/test_base_model.py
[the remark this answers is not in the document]
chandrasekharan-zipstack: Moot — the test file was removed from this PR (same reason as the sibling comment: no `pytest-django` / tox env yet).

backend/utils/models/base_model.py
```diff
@@ -1,9 +1,73 @@
...
+from django.utils import timezone
+
+
+class BaseModelQuerySet(models.QuerySet):
+    """QuerySet that mirrors ``auto_now`` semantics for bulk update paths.
+
+    ``modified_at = models.DateTimeField(auto_now=True)`` only fires on
+    ``Model.save()``. ``QuerySet.update()`` and ``QuerySet.bulk_update()``
+    issue raw SQL and bypass ``save()``, so ``modified_at`` stays at the
+    original creation time — silently drifting the audit trail. This
```
muhammad-ali-e: **[comment-analyzer · medium]** The docstring says `modified_at` "stays at the original creation time" — only true for rows never modified. After any prior `save()`, the column holds the previous-save timestamp, not creation time.

Suggestion:
```
...so ``modified_at`` keeps whatever value it had before the bulk path ran
(creation time for never-saved rows, the previous save() timestamp for
others) — silently drifting the audit trail.
```

backend/utils/models/base_model.py
```diff
@@ -1,9 +1,73 @@
...
+    QuerySet patches both paths so callers don't have to remember.
+
+    Callers can still override by passing ``modified_at`` explicitly (or by
+    including ``modified_at`` in the ``fields`` list for ``bulk_update``).
+    """
```
muhammad-ali-e, line 38: **[comment-analyzer · medium]** Docstring advertises that the QuerySet "patches both paths so callers don't have to remember" — true only when the queryset flows through a manager built from `BaseModelManager`. Raw SQL, `apps.get_model()` in migrations (returns historical models with a plain `models.Manager`), and subclasses that reassign `objects = models.Manager()` bypass this entirely.

Suggestion: add a caveat paragraph at the end of the docstring noting the invariant is a manager-level convention, not a model-level guarantee. Useful for future readers debugging stale timestamps.

backend/utils/models/base_model.py
```diff
@@ -1,9 +1,73 @@
...
+
+    def update(self, **kwargs):
+        kwargs.setdefault("modified_at", timezone.now())
+        return super().update(**kwargs)
+
+    def bulk_update(self, objs, fields, *args, **kwargs):
+        # Materialize objs before iterating so we don't exhaust a generator
+        # before Django's own tuple(objs) sees it.
```
muhammad-ali-e: **[comment-analyzer · low]** The comment pins rationale to a Django internal (`tuple(objs)`). Django is free to change that primitive at any point. The real WHY is local — we iterate `objs` ourselves to stamp `modified_at`, so we must materialize once to avoid exhausting the iterator before `super()` sees it.

Suggested rewrite:
```python
# Materialize once: we iterate objs to stamp modified_at, then pass the
# same sequence to super().bulk_update. A generator would be exhausted
# by the first pass.
```

backend/utils/models/base_model.py
```diff
@@ -1,9 +1,73 @@
...
+        objs = list(objs)
```
muhammad-ali-e: **[code-simplifier · low]** `objs = list(objs)` and `fields = list(fields)` run unconditionally, plus the per-object assignment loop. When the caller explicitly opts in (`modified_at` already in `fields`), the materialization cost is paid for nothing — Django would have iterated internally exactly once.

Suggestion: move the materialization inside the `if "modified_at" not in fields:` branch. Pass `objs`/`fields` straight through to `super().bulk_update` in the opt-out path.

Also worth a one-line docstring note that `bulk_update` materializes (same as Django's own impl) so `batch_size` controls SQL batching only, not memory.

backend/utils/models/base_model.py
```diff
@@ -1,9 +1,73 @@
...
+        fields = list(fields)
+        if "modified_at" not in fields:
```
muhammad-ali-e: **[code-simplifier · low]** The pattern "if `modified_at` not in fields, append it" appears three different ways in this file — here (line 27), in `bulk_create` (line 40-45), and in `save` (line 71-72). Each uses a slightly different idiom (`fields.append(...)` vs `list(update_fields) + ["modified_at"]`).

Suggestion: extract one module-scope helper and call it from all three sites. Something like:
```python
_AUTO_NOW_FIELD = "modified_at"

def _with_modified_at(fields):
    fields = list(fields)
    if _AUTO_NOW_FIELD not in fields:
        fields.append(_AUTO_NOW_FIELD)
    return fields
```
Reduces surface area and makes the invariant grep-able through a single constant.

backend/utils/models/base_model.py
```diff
@@ -1,9 +1,73 @@
...
+
+
+BaseModelManager = models.Manager.from_queryset(BaseModelQuerySet)
 
 
 class BaseModel(models.Model):
     created_at = models.DateTimeField(auto_now_add=True)
     modified_at = models.DateTimeField(auto_now=True)
 
+    objects = BaseModelManager()
```
muhammad-ali-e, line 92: **[type-design-analyzer · medium]** The invariant — "every write through this manager stamps `modified_at`" — is enforced only by convention. A `BaseModel` subclass author who writes `objects = models.Manager()` (or inherits a manager that doesn't chain `BaseModelQuerySet`) silently opts out with no warning.

Options to close the gap, roughly from cheapest to strictest:
1. Document the contract in `BaseModel`'s class docstring.
2. Register a Django system check that warns on any `BaseModel` subclass whose `_default_manager` isn't `BaseModelQuerySet`-derived.
3. `__init_subclass__` hook that raises `ImproperlyConfigured` if the default manager isn't compatible.

At minimum, (1) would be cheap and high-value.

backend/utils/models/base_model.py
```diff
@@ -1,9 +1,73 @@
...
+
     class Meta:
         abstract = True
+
+    def save(self, *args, update_fields=None, **kwargs):
```
muhammad-ali-e: **[code-reviewer · medium]** `def save(self, *args, update_fields=None, **kwargs)` forces `update_fields` to be keyword-only. Django's signature is `save(force_insert=False, force_update=False, using=None, update_fields=None)` — `update_fields` is the 4th positional. A caller using `obj.save(False, False, None, ["foo"])` slips through `*args`, leaves `update_fields=None` in the override, and the `modified_at` auto-bump is silently skipped.

Repo grep confirms no positional callers today (all use kwarg form), so not a live bug — but it's a silent foot-gun for future callers.

Suggestion: replicate Django's signature explicitly:
```python
def save(self, force_insert=False, force_update=False, using=None, update_fields=None):
    if update_fields is not None and "modified_at" not in update_fields:
        update_fields = list(update_fields) + ["modified_at"]
    return super().save(force_insert=force_insert, force_update=force_update,
                        using=using, update_fields=update_fields)
```

backend/utils/models/base_model.py
```diff
@@ -1,9 +1,73 @@
...
+        # Django only fires auto_now for fields listed in update_fields, so a
+        # partial save() silently drops the modified_at bump. Auto-include it
+        # whenever the caller restricts update_fields.
```
muhammad-ali-e: **[comment-analyzer · medium]** "Django only fires auto_now for fields listed in update_fields" is imprecise. Django calls `pre_save` on every concrete field during save, but when `update_fields` is set it only writes the listed fields to the database. `auto_now` still updates the attribute on the in-memory instance — the new value just never gets persisted.

The distinction matters for debuggability: a dev inspecting `instance.modified_at` after a partial save sees a fresh value and is confused why the DB row didn't change.

Suggested rewrite:
```python
# Django's save(update_fields=...) only writes the listed columns. auto_now
# still updates modified_at on the in-memory instance, but the new value is
# never persisted unless modified_at is in update_fields. Auto-include it so
# partial saves don't silently drop the bump.
```

backend/utils/models/base_model.py
```diff
@@ -1,9 +1,73 @@
...
+        if update_fields is not None and "modified_at" not in update_fields:
```
muhammad-ali-e: **[silent-failure-hunter · medium]** The guard `update_fields is not None and "modified_at" not in update_fields` is satisfied by the empty list `[]` — Django treats `update_fields=[]` as a deliberate no-op save (useful for triggering signals without a column write). This override rewrites `[]` → `["modified_at"]`, silently turning the no-op into a real UPDATE with write amplification.

No caller in the repo currently passes `[]`, but this is a subtle behavioral change from Django's documented semantics.

Suggested fix:
```python
if update_fields is not None and update_fields and "modified_at" not in update_fields:
    update_fields = list(update_fields) + ["modified_at"]
```
For consistency, apply the same `and update_fields` check in `bulk_update` (line 27) and `bulk_create` (line 43).

backend/utils/models/base_model.py
```diff
@@ -1,9 +1,73 @@
 from django.db import models
+from django.utils import timezone
+
+
+class BaseModelQuerySet(models.QuerySet):
```
muhammad-ali-e, line 20: **[pr-test-analyzer · high]** The PR description defers tests citing no `pytest-django` / no tox env. Verified: correct about infra. But overstated about testability — the four override methods (`update`, `bulk_update`, `bulk_create`, `BaseModel.save`) perform pure kwargs/list/objs mutation before delegating to `super()`. This logic is testable with `unittest.mock.patch.object(models.QuerySet, "update")` — no DB, no app registry, no tenants.

Precedent in this very repo: `backend/usage_v2/tests/test_helper.py` and `backend/prompt_studio/prompt_studio_core_v2/tests/test_build_index_payload.py` both use `sys.modules` stubbing to run pytest without full Django setup. The comments in those files explicitly document the pattern.

Suggested minimum suite (`backend/utils/tests/test_base_model.py`, ~80 lines):
1. `update()` injects `modified_at` via setdefault; explicit `modified_at=X` wins.
2. `bulk_update()` materializes a generator, appends `modified_at` to fields, stamps each obj.
3. `bulk_create(update_conflicts=True, update_fields=[...])` appends `modified_at`; without `update_conflicts` or without `update_fields`, leaves args untouched.
4. `save(update_fields=["name"])` → super called with `["name", "modified_at"]`; bare `save()` → super called with `update_fields=None` unchanged.

This would lock in every behavior this PR introduces, including the generator-materialization fix (commit e2152347) and the upsert guard (commit ed5693e1).

muhammad-ali-e (commented):
## Automated PR Review (pr-review-toolkit)

Ran six specialized review agents — **code-reviewer, code-simplifier, comment-analyzer, pr-test-analyzer, silent-failure-hunter, type-design-analyzer**. Consolidated, deduplicated, and prioritized findings below as inline comments.

**Overall assessment:** Solid structural fix. The `BaseModelQuerySet` + `BaseModelManager` + `BaseModel.save()` triad correctly addresses the three Django paths that skip `auto_now`. MRO verified: `(DefaultOrganizationManagerMixin, BaseModelManager)` resolves `super().get_queryset()` to `BaseModelQuerySet` as intended. `bulk_create(update_conflicts=True)` path is correct — Django fires `pre_save(add=True)` on all objs before the upsert SQL, so `EXCLUDED.modified_at` carries the fresh timestamp. Managers on non-`BaseModel` models are correctly left alone.

**Priority follow-ups (not blockers):**
1. **Tighten `save(update_fields=[])` guard** — empty list today silently becomes a real write (medium, inline).
2. **Positional `update_fields` bypass** in `save()` signature — no repo callers hit it today, but foot-gun for future (medium, inline).
3. **Test coverage is addable without DB** — the PR description defers tests on infra grounds, but the four override methods are pure kwargs/list mutation; two existing precedents (`usage_v2/tests/test_helper.py`, `prompt_studio_core_v2/tests/test_build_index_payload.py`) run without `pytest-django`. A ~80-line mock-based test file would lock in every behavior (medium, inline).
4. **`DefaultOrganizationManagerMixin(models.Manager)`** in `backend/utils/models/organization_mixin.py:26` (not in this diff) should change its base to `BaseModelManager`. Closes the MRO foot-gun and would let 9 subclass managers drop their explicit `BaseModelManager` base — smallest DRY win with the biggest safety payoff. **Recommend filing a follow-up.**
5. **`dashboard_metrics` upsert path** now depends entirely on the new override (commit `ed5693e1` removed explicit `modified_at` from `update_fields` in 6 locations). This is the highest-regression-risk consumer; a focused regression test would be valuable (medium, inline).

Docstrings and comments have a few accuracy issues flagged inline. Rest of findings (hardcoded `"modified_at"` literal, DRY of 3 near-duplicate append blocks, unconditional materialization) are low/nit — mentioned inline for context.

Type-design ratings: encapsulation **3/5**, invariant expression **3/5**, usefulness **4/5**, enforcement **2/5**. Enforcement drops because `BaseModel` subclasses can silently opt out by assigning `objects = models.Manager()`.

_Review generated by running pr-review-toolkit agents in parallel over the diff._

backend/utils/models/org_aware_manager.py
```diff
@@ -3,15 +3,15 @@
...
 from django.db.utils import OperationalError, ProgrammingError
+from utils.models.base_model import BaseModelManager
 from utils.models.org_path_discovery import get_org_path
 from utils.user_context import UserContext
 
 logger = logging.getLogger(__name__)
 
 
-class OrgAwareManager(models.Manager):
+class OrgAwareManager(BaseModelManager):
```
muhammad-ali-e, line 14: **[code-simplifier · medium cross-ref]** Good — swapping `models.Manager` → `BaseModelManager` here transitively covers every subclass of `OrgAwareManager`.

Related concern (out of this file's diff but worth flagging): `backend/utils/models/organization_mixin.py:26` still has `class DefaultOrganizationManagerMixin(models.Manager)`. The ~9 manager classes that compose it are edited to `(DefaultOrganizationManagerMixin, BaseModelManager)` — MRO keeps the mixin first so `super().get_queryset()` resolves correctly. But a future developer who reverses the order, or uses the mixin alone, silently regresses the `modified_at` invariant with no warning.

**Recommend follow-up PR:** change `DefaultOrganizationManagerMixin(models.Manager)` → `DefaultOrganizationManagerMixin(BaseModelManager)` (or even just `(object)` so it's a true mixin). Then the nine `(DefaultOrganizationManagerMixin, BaseModelManager)` declarations in this PR can each drop the explicit `BaseModelManager` — biggest DRY/safety win available.

backend/dashboard_metrics/tasks.py
```diff
@@ -121,7 +121,7 @@ def _bulk_upsert_hourly(aggregations: dict) -> int:
         objects,
         update_conflicts=True,
         unique_fields=["organization", "timestamp", "metric_name", "project", "tag"],
-        update_fields=["metric_type", "metric_value", "metric_count", "modified_at"],
+        update_fields=["metric_type", "metric_value", "metric_count"],
```
muhammad-ali-e, line 124: **[pr-test-analyzer · high]** This (and the sibling `_bulk_upsert_daily`, `_bulk_upsert_monthly`) now relies entirely on `BaseModelQuerySet.bulk_create`'s auto-inject — the explicit `"modified_at"` was removed from `update_fields` in commit `ed5693e1`. If that override ever regresses (Django upgrade, refactor, `from_queryset` plumbing break), the `EventMetrics*` tables silently stop bumping `modified_at` on upsert. It's a data-drift bug that's invisible until someone audits audit timestamps.

Suggestion: extend `backend/dashboard_metrics/tests/test_tasks.py` with a test that calls `_bulk_upsert_hourly` twice on the same unique key and asserts `modified_at` strictly increased between calls. That file already sits in `django.test.TestCase` with DB access, so the only infra it needs is what today's dev environment already provides. Gives a concrete tripwire for the most load-bearing consumer of the new override.

backend/utils/models/base_model.py
chandrasekharan-zipstack: Addressed in `d238afe79` — docstring reworded to cover never-saved vs previously-saved rows: "leaving `modified_at` at whatever value it had before the bulk path ran (creation time for never-saved rows, the previous save() timestamp for others)".

backend/utils/models/base_model.py
chandrasekharan-zipstack: Addressed in `d238afe79` — added an explicit caveat paragraph to the `BaseModelQuerySet` docstring covering reassigned `objects`, raw SQL, and `apps.get_model()` migration-time models.

backend/utils/models/base_model.py
chandrasekharan-zipstack: Addressed in `d238afe79` — comment rewritten to explain the local WHY (we iterate `objs` ourselves to stamp `modified_at`, so we must materialize once) rather than referencing Django's internal `tuple(objs)`.

backend/utils/models/base_model.py
chandrasekharan-zipstack: Addressed in `d238afe79` — both `list(objs)` and the stamping loop now sit inside the `if _AUTO_NOW_FIELD not in fields:` branch. The opt-in path delegates straight to `super().bulk_update` with the caller's sequence untouched.

backend/utils/models/base_model.py
chandrasekharan-zipstack: Addressed in `d238afe79` — extracted a module-scope `_with_modified_at(fields)` helper and a `_AUTO_NOW_FIELD = "modified_at"` constant; all three sites (`bulk_update`, `bulk_create`, `BaseModel.save`) now go through it.

backend/utils/models/base_model.py
chandrasekharan-zipstack: Partially addressed in `d238afe79` — added option (1), a `BaseModel` class docstring documenting the manager-level convention and the compose-`BaseModelManager` requirement for custom managers.

Options (2) system check and (3) `__init_subclass__` guard are deferred as a follow-up — genuinely higher-value long-term, but scope-bumping for this PR.

backend/utils/models/base_model.py
chandrasekharan-zipstack: Addressed in `d238afe79` — signature is now `save(self, force_insert=False, force_update=False, using=None, update_fields=None, **kwargs)`, matching Django 4.2. A caller using `obj.save(False, False, None, ["foo"])` now lands correctly on `update_fields=["foo"]` and picks up the auto-bump. `**kwargs` is kept for forward-compat with custom mixins/multi-table inheritance.

backend/utils/models/base_model.py
chandrasekharan-zipstack: Addressed in `d238afe79` — comment rewritten to your suggested wording: "auto_now still updates modified_at on the in-memory instance, but the new value is never persisted unless modified_at is in update_fields". The debuggability note matters.

backend/utils/models/base_model.py
chandrasekharan-zipstack: Addressed in `d238afe79` — real semantic bug, fixed. `save()` now guards with `if update_fields:` which preserves Django's `update_fields=[]` no-op (signals-only save). Same guard applied to `bulk_create(update_conflicts=True, update_fields=[])` via the `if update_conflicts and update_fields:` check. `bulk_update` with empty `fields` raises in Django's own impl, so no guard needed there.

backend/utils/models/base_model.py
chandrasekharan-zipstack: Fair — you're right that the `sys.modules` stubbing precedent exists and the mutation logic is testable without full Django setup. Deferring for this PR though: per the ticket owner, we're keeping the scope narrow and will come back to a proper test suite once we agree on a pytest convention that won't tangle with the legacy Django test cases (the PR body captures this). The test file I originally added was removed for the same reason — precedent alone isn't enough; we want a policy before re-introducing one-off patterns.

backend/utils/models/org_aware_manager.py
chandrasekharan-zipstack: Good catch, agreed — promoting `DefaultOrganizationManagerMixin(models.Manager)` → `DefaultOrganizationManagerMixin(BaseModelManager)` (or making it a true mixin inheriting `object`) would let the 9+ `(DefaultOrganizationManagerMixin, BaseModelManager)` declarations drop the explicit `BaseModelManager`.

Deferring to a follow-up PR — it's a real win but touches cross-cutting manager plumbing in both repos and deserves its own review cycle rather than being bolted onto this one.

backend/dashboard_metrics/tasks.py
chandrasekharan-zipstack: Valid tripwire; deferring with the other test additions per the test-policy decision on the sibling thread. Noting it as follow-up work so it isn't lost.

backend/utils/models/base_model.py
```diff
@@ -1,9 +1,53 @@
...
+        # Materialize objs before iterating so we don't exhaust a generator
+        # before Django's own tuple(objs) sees it.
+        objs = list(objs)
+        fields = list(fields)
+        if "modified_at" not in fields:
+            now = timezone.now()
+            for obj in objs:
+                obj.modified_at = now
+            fields.append("modified_at")
+        return super().bulk_update(objs, fields, *args, **kwargs)
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Stale against current HEAD. Commit `d238afe79` addressed both concerns directly:

- `save()` now guards with `if update_fields:` which preserves Django's documented `update_fields=[]` no-op (signals-only save).
- Field-list mutation is funneled through `_with_modified_at(fields)` which calls `list(fields)` before membership testing — generator-safe.

Same fix applies to `bulk_create` via `if update_conflicts and update_fields:`. Ready for re-review against the new commit.

backend/utils/models/base_model.py
```diff
@@ -1,9 +1,53 @@
...
     class Meta:
         abstract = True
+
+    def save(self, *args, update_fields=None, **kwargs):
+        # Django only fires auto_now for fields listed in update_fields, so a
+        # partial save() silently drops the modified_at bump. Auto-include it
+        # whenever the caller restricts update_fields.
+        if update_fields is not None and "modified_at" not in update_fields:
+            update_fields = list(update_fields) + ["modified_at"]
+        return super().save(*args, update_fields=update_fields, **kwargs)
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Stale — same fix as sibling thread. `d238afe79` guards against empty `update_fields` in both `save()` and `bulk_create`, and funnels field-list mutation through `_with_modified_at(fields)` which materializes the input with `list(...)` before iterating.

pk-zipstack (approved):

PR #1932, opened 2026-04-28
UN-3435 [FIX] Bind request_id and trace context to worker logs

muhammad-ali-e:
## What

- `RequestIDFilter` now falls back to `LogContext.request_id` from the thread-local context when `record.request_id` isn't set explicitly via `extra=`.
- New `_extract_request_id()` resolves a usable correlation ID from a Celery task's payload using the priority `request_id > file_execution_id > execution_id > run_id`. Looks at top-level kwargs and inside dict args (e.g. `ExecutionContext` / batch payloads).
- New Celery `task_prerun` / `task_postrun` signal handlers (`_bind_task_context` / `_clear_task_context`) bind `request_id` onto `LogContext` for every task and clear it on completion. Falls back to Celery `task_id` when the payload has none.
- Signal install (`_install_celery_request_id_signals`) is idempotent and silently no-ops if Celery is not importable (unit tests).

## Why

The new workers service was emitting `request_id:- trace_id:- span_id:-` in every log line, blocking ops debugging — flagged as P0 in [UN-3435](https://zipstack.atlassian.net/browse/UN-3435).

The plumbing already existed (`RequestIDFilter`, `OTelFieldFilter`, `LogContext.request_id`, format string `request_id:%(request_id)s`) but no code path actually populated the `request_id` field on log records. Even workers that called `log_context()` only set `task_id` / `execution_id` / `workflow_id` / `pipeline_id` / `organization_id` — never `request_id`. And the filter only read from `record.request_id`, not from `LogContext`, so even if context was set, the filter wouldn't see it.

## How

- **One signal, all workers.** The Celery `task_prerun` signal fires for every task in every worker (executor, file_processing, callback, api-deployment, general, notification, scheduler, ide_callback, log_consumer, plus cloud's `bulk_download` and `agentic_callback`). Binding once at the framework level avoids touching every task entry-point and keeps cloud workers free of changes.
- **Convention preserved.** Workers use `file_execution_id` as the value bound to `request_id` (legacy structure-tool behaviour) — `structure_tool_task.py:461,490` already sets `ExecutionContext.request_id = file_execution_id`. This gives per-file granularity in logs across the multi-tool execution chain (structure tool → executor → platform-service). Cross-service correlation is handled separately by OpenTelemetry `trace_id`, not by `request_id`.
- **Migration path is additive.** The lookup order in `_extract_request_id` is also the migration path. Callers may start passing a real HTTP `request_id` in the payload at any time and it will take precedence over `file_execution_id` automatically — no worker change required.
- **Trace ID / span ID** for the executor pod is fixed by a paired helm change in `unstract-cloud` (`workerExecutorV2` OTel auto-instrumentation wrapper in `otel.values.yaml`).

Safe. Changes are additive:
- `RequestIDFilter` still returns `"-"` when no value is found anywhere — same observable output as before for code paths that don't set `request_id`.
- Celery signals are weakly scoped (only mutate `LogContext.request_id` and `LogContext.task_id`), and `task_postrun` clears the context to avoid bleed-through across tasks.
- Existing `log_context()` callers continue to work — verified that prerun-bound `request_id` survives nested `log_context(...)` blocks.
- Signal install no-ops if Celery is missing, so test environments are unaffected.

## Database Migrations

None.

## Env Config

None.

## Relevant Docs

- [UN-3435](https://zipstack.atlassian.net/browse/UN-3435)

## Related Issues or PRs

- Paired helm fix in `unstract-cloud` for executor V2 OTel auto-instrumentation (forthcoming PR).

## Dependencies Versions

None.

## Notes on Testing

Verified locally with manual scenarios covering:
- `RequestIDFilter` fallback chain: `record.request_id` (extra=) > `LogContext.request_id` > `"-"`.
- `_coerce_id` rejects non-id types (objects, lists) and accepts only `str`/`int`/`UUID`.
- `_extract_request_id` priority by KEY (`request_id` > `file_execution_id` > `execution_id` > `run_id`) across containers.
- `_extract_request_id` resolves positional-string args via `inspect.signature(task.run).bind_partial(...)` — covers `async_execute_bin(schema, workflow_id, execution_id, ...)`.
- `_extract_request_id` searches dict args, dict-valued kwargs, and dataclass args.
- `_bind_task_context` swallows extraction errors and falls back to `task_id` (no stale state on misbehaving payload).
- `_clear_task_context` resets only `request_id`/`task_id`; baseline `worker_name` is preserved.
- `@functools.lru_cache(maxsize=1)` install is idempotent under concurrent calls (verified `cache_info()` reports 1 miss / 19 hits across 20 threads).

A follow-up PR will commit a regression test file (`workers/shared/tests/test_logger_request_id.py`, 22 tests) once test layout conventions for `workers/shared/tests/` are agreed.

After deploy, ops should see `request_id:<file_execution_id>` populated in worker logs (replaces `request_id:-`). Once the paired `unstract-cloud` helm change rolls out, `trace_id` / `span_id` will also be populated.

## Screenshots

ksnip-region-syfgZ4

## Checklist

- [x] I have solved all the sonar issues
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Any dependent changes have been merged and published in downstream modules

[UN-3435]: https://zipstack.atlassian.net/browse/UN-3435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[UN-3435]: https://zipstack.atlassian.net/browse/UN-3435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


muhammad-ali-e (commented):
## Multi-agent PR review summary (PR #1932)

Review ran six agents (Code Reviewer, Silent-Failure Hunter, Comment Analyzer, Test Analyzer, Type Design, Simplifier) over the single changed file `workers/shared/infrastructure/logging/logger.py`. Findings prioritised below; inline comments follow.

### P0 — Behavioural gaps
1. **Positional string args go unrecognised.** `_extract_request_id` only scans `dict` args, but the main worker entry — `async_execute_bin` dispatched from `backend/workflow_manager/workflow_v2/workflow_helper.py` — uses `args=[org_schema, workflow_id, execution_id, file_hash_in_str]` (all strings) plus a kwargs dict that doesn't contain `execution_id`. For this task family the new logic falls back to the Celery `task_id` UUID, defeating the stated per-execution granularity. (logger.py:625-629)
2. **`ExecutionContext` docstring example is wrong.** The example at logger.py:615 advertises `ExecutionContext` as a dict-args case, but `ExecutionContext` (`workers/shared/workflow/destination_connector.py:77`) is a `@dataclass` and is constructed *inside* the task body — it never reaches `task_prerun`. Misleading.
3. **Nested-dict scan order is insertion-order, not priority-order.** When two nested kwarg dicts each contain a different `_REQUEST_ID_KEYS` member, the result depends on which kwarg was added first, not on the priority list. (logger.py:620-624)

### P1 — Robustness / correctness
4. **`task_postrun` is not guaranteed to fire** on hard failure / SIGTERM in gevent/eventlet pools — the next thread reuse can carry stale `request_id` for any logging emitted *between* tasks. Hook `task_failure`/`task_revoked`, or always reset before binding. (logger.py:644-651)
5. **Silent `ImportError` swallow.** Fine for unit tests, but in a worker a real import break disables the whole feature with zero diagnostic. At minimum log at debug level. (logger.py:671-674)
6. **`task_id` over-clear** breaks nested-task contexts: `_TASK_SCOPED_FIELDS` is unconditionally nulled at every postrun, but `with_execution_context` (logger.py:486-492) also writes `task_id`. Nested/chained `task_postrun` clobbers the outer task's id. Save/restore prior values instead.
7. **`RequestIDFilter` docstring inaccuracy** — claims `LogContext.request_id` is set by `_bind_task_context`, but `log_context()` and `update_context()` are also writers. Drop the parenthetical. (logger.py:38-43)

### P2 — Tests / typing / style
8. **No tests committed** despite the PR description claiming "Verified locally with 9 sanity tests". Add the tests or trim the description.
9. **Bare `dict` / `tuple` annotations** on `_scan_for_id`, `_extract_request_id`. Use `Mapping[str, Any]` / `Sequence[Any]`.
10. **Double-checked locking is overengineered** for once-per-process install — a single `with _signals_lock:` plus inner check is enough.
11. **`_clear_task_context`'s `dict.fromkeys` trick** is less readable than `update_context(request_id=None, task_id=None)` for two fields.
12. **`str(value)` may stringify opaque objects** (UUIDs, dataclass `__str__` defaults). Guard with an `isinstance(value, (str, int, UUID))` check before stringifying.

_Generated by `/pr-review-toolkit:review-pr` orchestrating six review agents._

workers/shared/infrastructure/logging/logger.py
```diff
@@ -35,14 +35,19 @@ class LogContext:
 class RequestIDFilter(logging.Filter):
     """Filter to inject request_id into log records.
 
-    Adopts the proven pattern from unstract/core/flask/logging.py for consistency.
-    Normalizes missing or falsy values (None, empty string, etc.) to "-" for
-    consistent log formatting.
+    Resolution order:
+      1. ``record.request_id`` if set explicitly via ``extra={...}``.
+      2. ``LogContext.request_id`` from the thread-local context (set by the
+         Celery ``task_prerun`` signal in ``_bind_task_context``).
```
muhammad-ali-e: **Inaccurate**: `LogContext.request_id` is also written by `WorkerLogger.log_context()` and `WorkerLogger.update_context()`. Naming `_bind_task_context` as the single writer is misleading and creates rot risk if the helper is renamed. Suggest dropping the parenthetical:

```python
2. ``LogContext.request_id`` from the thread-local context.
```

workers/shared/infrastructure/logging/logger.py
```diff
@@ -145,6 +150,8 @@ def configure(
             category=UserWarning,
         )
 
+        _install_celery_request_id_signals()
```
muhammad-ali-e, line 155: Quick observation: `_install_celery_request_id_signals()` runs from `WorkerLogger.configure()`, which is gated by `cls._configured`. After fork, child processes inherit `cls._configured = True`, so a child that re-imports this module will skip signal install. Prefork pool is fine because Celery's signal registry is also forked, but spawn-mode pools (macOS dev) and Celery autoreload won't be. Consider gating on `task_prerun.receivers` membership instead of `_signals_installed`. Low priority for prod (Linux prefork).

workers/shared/infrastructure/logging/logger.py
```diff
@@ -575,6 +582,101 @@ def log_execution(func: Callable) -> Callable:
     return logged_execution()(func)
 
 
+_REQUEST_ID_KEYS = ("request_id", "file_execution_id", "execution_id", "run_id")
+
+
+def _scan_for_id(container: dict) -> str | None:
```
muhammad-ali-e: **Type design**: `container: dict` should be `Mapping[str, Any]` (from `collections.abc`). Bare `dict` accepts `dict[bytes, object]` while the body assumes string-keyed mappings. Same applies to `_extract_request_id` below.

```python
from collections.abc import Mapping, Sequence

def _scan_for_id(container: Mapping[str, Any]) -> str | None:
    ...

def _extract_request_id(args: Sequence[Any], kwargs: Mapping[str, Any]) -> str | None:
    ...
```

workers/shared/infrastructure/logging/logger.py
```diff
@@ -575,6 +582,101 @@ def log_execution(func: Callable) -> Callable:
...
+    """Look up `_REQUEST_ID_KEYS` in a single dict, in priority order."""
+    for key in _REQUEST_ID_KEYS:
+        value = container.get(key)
+        if value:
+            return str(value)
+    return None
```
muhammad-ali-e, line 610: **Latent risk**: `str(value)` will stringify any object that survives the truthiness check, including dataclass instances or UUIDs whose `__str__` could return things like `'<ExecutionContext object at 0x…>'` or a multi-line blob. Once it lands in the log line it also flows into OTel attributes verbatim.

Safer:

```python
for key in _REQUEST_ID_KEYS:
    value = container.get(key)
    if isinstance(value, str) and value:
        return value
    if isinstance(value, (int, UUID)) and value:
        return str(value)
return None
```

workers/shared/infrastructure/logging/logger.py
```diff
@@ -575,6 +582,101 @@ def log_execution(func: Callable) -> Callable:
...
+
+    The lookup order below is also the migration path: callers may start
+    passing a real HTTP ``request_id`` in the payload at any time and it
+    will take precedence over ``file_execution_id`` automatically -- no
+    worker change required.
+
+    Search order:
+      1. Top-level kwargs (e.g. ``task(execution_id=...)``).
+      2. Dict values inside kwargs (e.g. ``task(context={"execution_id": ...})``).
+      3. Dict args (e.g. ``ExecutionContext`` / batch payloads in positional args).
```
muhammad-ali-e: **Misleading example.** `ExecutionContext` (`workers/shared/workflow/destination_connector.py:77`) is a `@dataclass`, not a `dict`. The `for arg in args ... if isinstance(arg, dict)` loop below will skip it. More importantly, `ExecutionContext` is constructed *inside* the task body (e.g. `structure_tool_task.py:455-462`) and is never passed as a Celery argument, so it never reaches `task_prerun`.

Either drop the `ExecutionContext` reference, or genuinely support dataclasses with `dataclasses.is_dataclass(arg)` + the existing `safe_get_attr` helper in this module.

workers/shared/infrastructure/logging/logger.py
```diff
@@ -575,6 +582,101 @@ def log_execution(func: Callable) -> Callable:
...
+    """
+    value = _scan_for_id(kwargs)
+    if value:
+        return value
+    for nested in kwargs.values():
```
muhammad-ali-e: **Priority-order bug.** This loop walks `kwargs.values()` in insertion order. If a payload looks like `{"context": {"execution_id": "A"}, "meta": {"file_execution_id": "B"}}`, the result depends on which kwarg was added first — not on the documented priority of `request_id > file_execution_id > execution_id > run_id`. To honour the priority guarantee, do an outer pass over `_REQUEST_ID_KEYS` and an inner pass over containers:

```python
containers = [kwargs, *(v for v in kwargs.values() if isinstance(v, dict)),
              *(a for a in args or () if isinstance(a, dict))]
for key in _REQUEST_ID_KEYS:
    for c in containers:
        v = c.get(key)
        if v:
            return str(v)
return None
```

workers/shared/infrastructure/logging/logger.py
```diff
@@ -575,6 +582,101 @@ def log_execution(func: Callable) -> Callable:
...
+        if isinstance(nested, dict):
+            value = _scan_for_id(nested)
+            if value:
+                return value
+    for arg in args or ():
```
muhammad-ali-e: **P0 — silent regression for the main worker entry.** This loop only inspects `dict` positional args. The Celery dispatch from `backend/workflow_manager/workflow_v2/workflow_helper.py:600` calls `send_task("async_execute_bin", args=[org_schema, workflow_id, execution_id, file_hash_in_str], kwargs={"scheduled":..., "pipeline_id":..., ...})`:
- `args` are all positional **strings**, so this loop matches none of them.
- `kwargs` has no entry from `_REQUEST_ID_KEYS` either.

Result: `_extract_request_id` returns `None`, the caller falls back to the Celery `task_id` UUID, and worker logs show `request_id:<celery-uuid>` instead of `execution_id`. This defeats the stated per-execution granularity for the primary API/general-deployment workflow path.

Two options:
1. Pair the task signature against `args` using `inspect.signature(task.run).bind(*args, **kwargs)` and look up `_REQUEST_ID_KEYS` against the bound names.
2. Have the dispatchers always pass `execution_id` as a kwarg (smaller change but couples convention).

Whichever you pick, please add a regression test for the positional-string case.

workers/shared/infrastructure/logging/logger.py
```diff
@@ -575,6 +582,101 @@ def log_execution(func: Callable) -> Callable:
...
+        if isinstance(arg, dict):
+            value = _scan_for_id(arg)
+            if value:
+                return value
+    return None
+
+
+# Fields owned by the per-task signal handlers; cleared at task_postrun without
+# touching baseline fields like ``worker_name`` set at WorkerLogger.configure().
+_TASK_SCOPED_FIELDS = ("request_id", "task_id")
```
muhammad-ali-e: **Type / shape suggestion.** `_TASK_SCOPED_FIELDS` is fed to `dict.fromkeys` for set semantics — a `frozenset` expresses that more honestly than a tuple (and makes lookup O(1) if it ever grows):

```python
_TASK_SCOPED_FIELDS: Final[frozenset[str]] = frozenset({"request_id", "task_id"})
```

Note: `_REQUEST_ID_KEYS` above must stay a tuple (priority order).

workers/shared/infrastructure/logging/logger.py
```diff
@@ -575,6 +582,101 @@ def log_execution(func: Callable) -> Callable:
...
+
+
+def _bind_task_context(task_id, task, args, kwargs, **_):
+    """Celery ``task_prerun`` handler: bind request_id onto the log context."""
+    request_id = _extract_request_id(args or (), kwargs or {}) or task_id
+    WorkerLogger.update_context(request_id=request_id, task_id=task_id)
```
muhammad-ali-e, line 721: **Stale-state / nesting risk.**

1. If `_extract_request_id` raises (e.g. an unusual arg is iterable but `.get` fails), Celery `send_robust` swallows the exception and logs a warning, but the thread-local still holds whatever the *previous* task bound. Worth wrapping the body in a try/except that calls `_clear_task_context()` on failure so a misbehaving task can't leak state to the next one.

2. `_clear_task_context` unconditionally nulls `task_id` at `task_postrun`, but `with_execution_context` (logger.py:486-492) also writes `task_id` from `task_instance.request.id`. For nested or chained tasks (task A calls task B in-process, or postrun of an inner task fires while outer is still on the stack), the inner postrun nulls the outer's `task_id`. Either save+restore prior values around bind/clear, or stack contexts.

workers/shared/infrastructure/logging/logger.py
```diff
@@ -575,6 +582,101 @@ def log_execution(func: Callable) -> Callable:
...
+
+
+def _clear_task_context(**_):
+    """Celery ``task_postrun`` handler: reset task-scoped fields only.
+
+    Preserves baseline context (``worker_name``, etc.) set at
+    ``WorkerLogger.configure()``; only nulls out the per-task fields bound
+    in ``_bind_task_context``.
+    """
+    WorkerLogger.update_context(**dict.fromkeys(_TASK_SCOPED_FIELDS))
```
muhammad-ali-e: **Readability.** With only two fields the explicit form is clearer and greppable:

```python
WorkerLogger.update_context(request_id=None, task_id=None)
```

The `dict.fromkeys(...)` trick is a slight over-abstraction here; `_TASK_SCOPED_FIELDS` could be inlined or removed.

workers/shared/infrastructure/logging/logger.py
```diff
@@ -575,6 +582,101 @@ def log_execution(func: Callable) -> Callable:
...
+
+def _install_celery_request_id_signals() -> None:
+    """Wire Celery signals so every task has request_id bound to its log context.
+
+    Idempotent and thread-safe via double-checked locking — safe to call from
+    each worker's logger setup.  Silently no-ops if Celery is not importable
+    (e.g. unit tests).
+    """
+    global _signals_installed
+    if _signals_installed:
```
muhammad-ali-e: **Overengineered.** `_install_celery_request_id_signals` is called once per worker boot — the outer `if _signals_installed` pre-check buys nothing measurable but forces a `global` declaration and a double-check pattern. `with _signals_lock:` plus the inner check is enough.

Even simpler: drop the flag+lock entirely and use

```python
@functools.lru_cache(maxsize=1)
def _install_celery_request_id_signals() -> None:
    ...
```

which gives idempotency + thread safety for free.

workers/shared/infrastructure/logging/logger.py
```diff
@@ -575,6 +582,101 @@ def log_execution(func: Callable) -> Callable:
...
+        return
+    with _signals_lock:
+        if _signals_installed:
+            return
+        try:
+            from celery.signals import task_postrun, task_prerun
+        except ImportError:
+            return
```
muhammad-ali-e: **Silent failure**: in a real worker, Celery is always installed — an `ImportError` here means the deployment is broken, but the feature silently disappears with zero diagnostic. For unit tests this is fine; for prod it isn't. Suggest:

```python
except ImportError as exc:
    logging.getLogger(__name__).debug(
        "celery.signals not importable; request_id signal install skipped: %s", exc
    )
    return
```

workers/shared/infrastructure/logging/logger.py
```diff
@@ -575,6 +582,101 @@ def log_execution(func: Callable) -> Callable:
...
+        task_prerun.connect(_bind_task_context, weak=False)
+        task_postrun.connect(_clear_task_context, weak=False)
+        _signals_installed = True
```
muhammad-ali-e: **Test gap (PR-level).** The PR description claims "Verified locally with 9 sanity tests covering [list]". `gh pr diff 1932` shows only this implementation file changed — none of those tests are committed. At minimum I'd want regression coverage for:

- `RequestIDFilter` fallback chain (record extra > LogContext > "-")
- `_extract_request_id` priority + recursion into nested dicts
- `_clear_task_context` resets only `_TASK_SCOPED_FIELDS` and preserves `worker_name` (this one was the explicit fix in commit 0eecd5ef)
- The positional-string-args case noted at line 625 above
- Nested-task `task_id` clear interaction with `with_execution_context`

Either commit the local sanity tests or trim the description.

workers/shared/infrastructure/logging/logger.py
```diff
@@ -575,6 +582,69 @@ def log_execution(func: Callable) -> Callable:
...
+
+def _bind_task_context(task_id, task, args, kwargs, **_):
+    """Celery ``task_prerun`` handler: bind request_id onto the log context."""
+    request_id = _extract_request_id(args or (), kwargs or {}) or task_id
+    WorkerLogger.update_context(request_id=request_id, task_id=task_id)
+
+
+def _clear_task_context(**_):
+    """Celery ``task_postrun`` handler: clear context to avoid bleed-through."""
+    WorkerLogger.clear_context()
```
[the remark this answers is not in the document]
muhammad-ali-e: Resolved in commit `0eecd5ef`. `_clear_task_context` no longer calls `clear_context()` (which would discard the entire `LogContext` and lose `worker_name`). It now does `update_context(request_id=None, task_id=None)`, nulling only the task-scoped fields and preserving baseline fields like `worker_name` set at `WorkerLogger.configure()`.

workers/shared/infrastructure/logging/logger.py
```diff
@@ -575,6 +582,69 @@ def log_execution(func: Callable) -> Callable:
...
+    global _signals_installed
+    if _signals_installed:
+        return
+    try:
+        from celery.signals import task_postrun, task_prerun
+    except ImportError:
+        return
+    task_prerun.connect(_bind_task_context, weak=False)
+    task_postrun.connect(_clear_task_context, weak=False)
+    _signals_installed = True
```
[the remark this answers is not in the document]
muhammad-ali-e: Resolved in commit `b5b48a47` via a different mechanism per a separate review comment: switched from explicit flag + `Lock` to `@functools.lru_cache(maxsize=1)`, which gives idempotency + thread safety by construction. Even if `lru_cache` lets the wrapped function run twice under a tight race, the underlying `task_prerun.connect(_bind_task_context, weak=False)` is a no-op on duplicate registration — Django/Celery's `Signal.connect` deduplicates by `(receiver_id, sender_id)`, so the same module-level function reference cannot end up registered twice. In practice, signal install runs at module import time on a single thread (master process), so the race is theoretical anyway.

workers/shared/infrastructure/logging/logger.py
```diff
@@ -575,6 +582,69 @@ def log_execution(func: Callable) -> Callable:
...
+        value = kwargs.get(key)
+        if value:
+            return str(value)
+    for arg in args or ():
+        if isinstance(arg, dict):
+            for key in _REQUEST_ID_KEYS:
+                value = arg.get(key)
+                if value:
+                    return str(value)
+    return None
```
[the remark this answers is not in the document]
muhammad-ali-e: Resolved in commit `b5b48a47`. `_gather_containers` now scans nested `Mapping`-typed values inside kwargs in addition to top-level kwargs and dict args. Test case `task(context={"execution_id": "abc"})` now resolves to `"abc"` instead of falling back to `task_id`.

workers/shared/infrastructure/logging/logger.py
muhammad-ali-e: Resolved in commit `b5b48a47`. Docstring on `RequestIDFilter` no longer name-checks a single helper as the sole writer of `LogContext.request_id`; reads simply: `LogContext.request_id` from the thread-local context.

workers/shared/infrastructure/logging/logger.py
muhammad-ali-e: Acknowledged. Production runs Linux prefork, so `cls._configured` inheritance after fork is the intended behaviour (master imports module → installs signal once → forks; children inherit both the lru_cache state and Celery's signal registry). Spawn-mode pools (macOS dev) and Celery autoreload are not used in production. Marking as a known limitation rather than fixing in this P0 PR; happy to address if it becomes relevant.

workers/shared/infrastructure/logging/logger.py
muhammad-ali-e: Resolved in commit `b5b48a47`. Imported `Mapping` and `Sequence` from `collections.abc`; both `_extract_request_id` and `_scan_for_id` (now `_arg_as_mapping` / `_bind_positional_args_to_names` after the cognitive-complexity refactor in `42172684`) use the protocol types instead of bare `dict` / `tuple`.

workers/shared/infrastructure/logging/logger.py
muhammad-ali-e: Resolved in commits `b5b48a47` and `8db44ebe`. New `_coerce_id` accepts only `str` / `int` / `UUID` and explicitly rejects `bool` (`bool` is an `int` subclass in Python). Anything else returns `None`, so a misnamed payload field cannot stringify into a log line or OTel attribute.

workers/shared/infrastructure/logging/logger.py
muhammad-ali-e: Resolved in commit `b5b48a47`. Dropped the misleading `ExecutionContext` reference from the docstring. Genuine dataclass support added via `dataclasses.is_dataclass(arg)` + `dataclasses.asdict(arg)` in `_arg_as_mapping`, so dataclass positional args (if any reach `task_prerun` in the future) will be scanned correctly.

workers/shared/infrastructure/logging/logger.py
muhammad-ali-e: Resolved in commit `b5b48a47`. Loop restructured: outer over `_REQUEST_ID_KEYS` (priority list), inner over containers. A payload with `file_execution_id` in one nested dict and `execution_id` in another now deterministically returns the former regardless of insertion order.

workers/shared/infrastructure/logging/logger.py
muhammad-ali-e: **Fixed — this was the P0**. Resolved in commit `b5b48a47`. `_gather_containers` (split into `_bind_positional_args_to_names` after the cognitive-complexity refactor in `42172684`) uses `inspect.signature(task.run).bind_partial(*args, **kwargs)` to map positional args to parameter names. `async_execute_bin(schema_name, workflow_id, execution_id, file_hash_in_str)` now surfaces `execution_id` correctly. No dispatcher contract change required.

workers/shared/infrastructure/logging/logger.py
muhammad-ali-e: Adopted the cleaner approach from your other comment instead — the `_TASK_SCOPED_FIELDS` constant was removed entirely and `_clear_task_context` now uses the explicit `update_context(request_id=None, task_id=None)` form. No `frozenset` / `tuple` debate left to have.

workers/shared/infrastructure/logging/logger.py
muhammad-ali-e: Part 1 (extraction raises) resolved in commit `b5b48a47`: `_bind_task_context` wraps the call in `try/except Exception` and falls back to `task_id` so a misbehaving payload cannot leak prior state. Commit `8db44ebe` adds a DEBUG log on the exception path so operators can diagnose malformed payloads.

Part 2 (nested-task `task_id` interaction with `with_execution_context`) skipped: verified earlier in the PR discussion that no production code path uses synchronous `.apply()` for in-process task nesting — all task-to-task dispatch goes via `send_task`/`apply_async`/`chord`, which runs the inner task in a separate worker process with its own `LogContext`. If we ever introduce sync nesting, save/restore around bind/clear becomes necessary.

workers/shared/infrastructure/logging/logger.py
muhammad-ali-e: Adopted in commit `b5b48a47`. `_clear_task_context` now reads `WorkerLogger.update_context(request_id=None, task_id=None)`. The `_TASK_SCOPED_FIELDS` constant + `dict.fromkeys` indirection was removed.

workers/shared/infrastructure/logging/logger.py
muhammad-ali-e: Adopted in commit `b5b48a47`. Replaced the explicit flag + `Lock` + double-checked locking with `@functools.lru_cache(maxsize=1)`. Cleaner, fewer moving parts, and `Signal.connect` itself deduplicates so even a race-induced double-call is a no-op.

workers/shared/infrastructure/logging/logger.py
muhammad-ali-e: Adopted in commit `b5b48a47`. The `ImportError` branch now logs at DEBUG level via `logging.getLogger(__name__).debug(...)` with the exception, so a deployment that's missing Celery is diagnosable rather than mysteriously losing request_id.

workers/shared/infrastructure/logging/logger.py
muhammad-ali-e: Acknowledged. Tests were added in commit `b5b48a47` (22 tests at `workers/shared/tests/test_logger_request_id.py`, covering filter fallback chain, coerce rejection, key-priority scan, signature-bound positional args, dataclass arg, nested dict priority, baseline preservation, and concurrent install). Removed from this PR in `a41f7314` to keep the diff scoped to the P0 ops fix; will land in a follow-up PR once test layout conventions for `workers/shared/tests/` are agreed. PR description's `Notes on Testing` section trimmed to match.

workers/shared/infrastructure/logging/logger.py
```diff
@@ -575,6 +584,144 @@ def log_execution(func: Callable) -> Callable:
...
+def _coerce_id(value: Any) -> str | None:
+    """Return a safe string id, or None if value is not a usable id type.
+
+    Rejects arbitrary objects so a misnamed payload field (e.g. a dataclass
+    or a list) cannot leak its ``__str__`` into log lines or OTel attributes.
+    """
+    if isinstance(value, str):
+        return value or None
+    if isinstance(value, (int, UUID)):
+        return str(value)
```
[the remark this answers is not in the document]
muhammad-ali-e: Resolved in commit `8db44ebe`. Added an explicit `isinstance(value, bool)` guard at the top of `_coerce_id` so `True`/`False` return `None` instead of being accepted via the `int` branch and stringified to `"True"`/`"False"`.

workers/shared/infrastructure/logging/logger.py
```diff
@@ -575,6 +584,144 @@ def log_execution(func: Callable) -> Callable:
...
+    try:
+        from celery.signals import task_postrun, task_prerun
+    except ImportError as exc:
+        logging.getLogger(__name__).debug(
+            "celery.signals not importable; request_id signal install skipped: %s",
+            exc,
+        )
+        return
+    task_prerun.connect(_bind_task_context, weak=False)
+    task_postrun.connect(_clear_task_context, weak=False)
```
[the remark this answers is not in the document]
muhammad-ali-e: Acknowledged but not changing. The `lru_cache` race-condition concern is correct in theory: under tight contention the wrapped function can be called more than once. However, the wrapped operation here is `task_prerun.connect(_bind_task_context, weak=False)`, and Django/Celery's `Signal.connect` deduplicates by `(receiver_id, sender_id)` — calling it twice with the same module-level function reference is a no-op. In addition, signal install fires from `WorkerLogger.configure()` at module import time, which runs single-threaded under the Python import lock, so the race is theoretical for the actual call site. Switching back to explicit `Lock` + flag was rejected by the PR author in review #11 as overengineered for this call pattern.

vishnuszipstack (approved):

workers/shared/infrastructure/logging/logger.py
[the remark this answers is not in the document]
muhammad-ali-e: Confirmed against the source on my end:

- `Signal._connect_signal` (`celery/utils/dispatch/signal.py:200-214`) builds `lookup_key` from `(_make_id(receiver), _make_id(sender))` and walks `self.receivers` under `self.lock` — appends only on no match. Connecting the same module-level handler twice is a verified no-op.
- `WorkerLogger.configure()` has an `if cls._configured: return` guard at `logger.py:94`, so `_install_celery_request_id_signals()` body fires at most once per process regardless of `lru_cache`.
- Bootstrap runs before task threads spin up, so no real concurrency on the first call.

Defense-in-depth: any one of `cls._configured`, `lru_cache`, or Celery's own dedup would suffice. Leaving the code as-is.

pk-zipstack (approved):

workers/shared/infrastructure/logging/logger.py
ritwik-g, line 1: @muhammad-ali-e so earlier was this not done in a standardised way? Also if the execution id is not available will we still autogenerate request id? If not that should be done

cc: @greptileai

PR #1473, opened 2025-08-08
UN-2696 [FEAT] Add cloud-only configuration registry for organization-specific settings 

vishnuszipstack:
## What

- Implemented a dynamic configuration registry system that allows cloud deployments to extend OSS configuration keys without exposing cloud-specific features to open-source users. The system enables organization-level overrides for cloud-only settings like ENABLE_HIGHLIGHT_API_DEPLOYMENT.

## Why

 - Cloud Feature Isolation: Need to provide cloud-only features (like highlight API deployment) without exposing them in OSS codebase
 - Organization-Level Control: Different cloud organizations need different configuration values for the same feature                                                                    - Maintainability: Avoid code duplication and maintain clean separation between OSS and cloud codebases                                                                                
- Extensibility: Provide a scalable way to add more cloud-only configurations in the future

## How

1. Configuration Registry System:                                                                                                
- Created ConfigurationRegistry in configuration/config_registry.py that dynamically loads both OSS and cloud configuration keys 
- Registry scans plugins/configuration/ directory for cloud-specific configs                                                     
- Provides unified interface: ConfigurationRegistry.get_config_spec(key_name)                                                    
- Handles plugin availability gracefully (no errors if plugin missing)

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

- No

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

I have read and understood the [Contribution Guidelines](https://docs.unstract.com/unstract/contributing/unstract/).

PR #1933, opened 2026-04-28
[HOTFIX] Honor ENABLE_HIGHLIGHT_API_DEPLOYMENT env var on on-prem

vishnuszipstack:
## Summary
- Hotfix on top of `v0.161.4`. Pairs with cloud hotfix Zipstack/unstract-cloud#1467.
- Wraps `ENABLE_HIGHLIGHT_API_DEPLOYMENT` env-var read in `backend/backend/settings/base.py` with `CommonUtils.str_to_bool(...)` so values like `False` / `false` / `0` actually evaluate falsy. Previously `os.environ.get(...)` returned the raw string, so any non-empty value (including `"False"`) was truthy in Python and bypassed the gate.
- The setting is consumed as the `ConfigSpec.default` in `unstract-cloud/backend/plugins/configuration/cloud_config.py`. `Configuration.get_value_by_organization(...)` returns this default when no per-org row exists, so the bool fix takes effect on cloud and on-prem builds where the configuration plugin is loaded.

## Why this is the only OSS change
- The on-prem regression fix (loading the configuration plugin in on-prem images so the registry knows the key) is cloud-only and ships in Zipstack/unstract-cloud#1467.
- This OSS PR is intentionally minimal — only the bool parsing.

## Test plan
- [ ] On-prem (with cloud PR #1467 merged & on-prem image rebuilt): set `ENABLE_HIGHLIGHT_API_DEPLOYMENT=true`, run an API deployment (sync + polled), confirm `highlight_data` and `extracted_text` are present in the response metadata.
- [ ] On-prem: set `ENABLE_HIGHLIGHT_API_DEPLOYMENT=False`, confirm both keys are stripped (this is the case the bool fix unblocks).
- [ ] Cloud: with no per-org `Configuration` row and env=`true`, confirm fleet-wide default keeps highlight data.
- [ ] Cloud: with no per-org row and env=`False`, confirm fleet-wide default strips both keys.

### DIFF
--- a/backend/backend/settings/base.py
+++ b/backend/backend/settings/base.py
@@ -689,4 +689,6 @@ if missing_settings:
     )
     raise ValueError(ERROR_MESSAGE)
 
-ENABLE_HIGHLIGHT_API_DEPLOYMENT = os.environ.get("ENABLE_HIGHLIGHT_API_DEPLOYMENT", False)
+ENABLE_HIGHLIGHT_API_DEPLOYMENT = CommonUtils.str_to_bool(
+    os.environ.get("ENABLE_HIGHLIGHT_API_DEPLOYMENT", "False")
+)


backend/api_v2/deployment_helper.py
```diff
@@ -258,18 +259,11 @@ def execute_workflow(
...
-                "ENABLE_HIGHLIGHT_API_DEPLOYMENT"
-            ):
-                enable_highlight = Configuration.get_value_by_organization(
-                    config_key="ENABLE_HIGHLIGHT_API_DEPLOYMENT",
-                    organization=organization,
-                )
+            enable_highlight = Configuration.get_value_by_organization(
+                config_key=ConfigKey.ENABLE_HIGHLIGHT_API_DEPLOYMENT,
+                organization=organization,
+            )
```
muhammad-ali-e: don't we required `is_config_key_available(` check?

muhammad-ali-e (approved):
Added NITs

chandrasekharan-zipstack (approved):

chandrasekharan-zipstack (approved):

backend/api_v2/deployment_helper.py
vishnuszipstack: Yes, the `ConfigurationRegistry.is_config_key_available("ENABLE_HIGHLIGHT_API_DEPLOYMENT")` check is still in place — this PR does not modify `deployment_helper.py` anymore.

The PR has been simplified down to a single change in `backend/backend/settings/base.py`: wrapping the env-var read in `CommonUtils.str_to_bool(...)` so falsy strings like `"False"` / `"false"` / `"0"` actually evaluate to `False` (previously they were truthy strings that bypassed the gate).

The on-prem fix (loading the configuration plugin so the registry knows the key on on-prem images) ships in cloud PR Zipstack/unstract-cloud#1467. Together they restore the env-var path that PR #1473 (UN-2696) regressed.

backend/backend/settings/base.py
```diff
@@ -689,4 +689,6 @@ def filter(self, record):
     )
     raise ValueError(ERROR_MESSAGE)
 
-ENABLE_HIGHLIGHT_API_DEPLOYMENT = os.environ.get("ENABLE_HIGHLIGHT_API_DEPLOYMENT", False)
+ENABLE_HIGHLIGHT_API_DEPLOYMENT = CommonUtils.str_to_bool(
+    os.environ.get("ENABLE_HIGHLIGHT_API_DEPLOYMENT", "False")
+)
```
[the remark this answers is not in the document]
vishnuszipstack: Greptile read this incorrectly — the setting **is** consumed, just indirectly through the cloud overlay.

`backend/plugins/configuration/cloud_config.py` (in unstract-cloud) registers a `ConfigSpec` with `default=settings.ENABLE_HIGHLIGHT_API_DEPLOYMENT`. Both gate sites call `Configuration.get_value_by_organization(...)`, which returns `config_spec.default` when no per-org `Configuration` row exists (see `backend/configuration/models.py:101-104`). So on cloud and on-prem (after the paired cloud PR Zipstack/unstract-cloud#1467 ships the configuration plugin in the on-prem image), the parsed bool from this `base.py` setting is what the gate sees.

Pure OSS (no cloud overlay) is intentionally unaffected — the registry availability check returns `False` and `enable_highlight` stays at its hardcoded `False`. That is the post-#1473 status quo and not what this hotfix tries to change.

Re the inline `"1"` concern: every other env-var-to-bool in the project (`AUTH_ENABLED`, `ENABLE_EMAIL_NOTIFICATIONS`, `DISABLE_SIGNUP`, etc.) uses the same `"true"`-only convention, so this stays consistent.

PR #1934, opened 2026-04-28
[FIX] Strip deprecated sampling params for Claude Opus 4.7

pk-zipstack:
## What

- Add `_has_deprecated_sampling_params(model)` and `_strip_deprecated_sampling_params(validated)` helpers in `unstract/sdk1/.../adapters/base1.py`. The helpers substring-match the model id against `claude-opus-4-7` (case-insensitive, with `.` / `_` normalized to `-`) and pop `temperature` / `top_p` / `top_k` from the validated metadata when matched.
- Wire the strip step into `AnthropicLLMParameters.validate()`, `AWSBedrockLLMParameters.validate()`, and `AzureAIFoundryLLMParameters.validate()`.
- The strip inspects both `model` and `model_id`, so Bedrock callers routing through an Application Inference Profile (AIP) are covered when the standard model id appears in either field.

## Why

Anthropic deprecated `temperature`, `top_p`, and `top_k` starting with Claude Opus 4.7. Sending any of them now returns a 400 from Anthropic and from every provider that proxies it (Bedrock, Azure AI Foundry, Vertex AI). The Azure AI Foundry path is what surfaced first in the connector test:

```
Error from LLM adapter 'azure_ai': Azure_aiException -
{"type":"error","error":{"type":"invalid_request_error",
"message":"`temperature` is deprecated for this model."}, ...}
```

Pydantic's `model_dump()` on `BaseChatCompletionParameters` re-emits the default `temperature=0.1` even when the caller never set one, so we cannot rely on "don't pass it" — we have to actively pop it after validation.

Reference: <https://platform.claude.com/docs/en/about-claude/models/whats-new-claude-4-7>

## How

`base1.py`:

- New module-level pattern table `_SAMPLING_DEPRECATED_MODEL_PATTERNS = ("claude-opus-4-7",)` and `_DEPRECATED_SAMPLING_PARAMS = ("temperature", "top_p", "top_k")`. The pattern table is the single point of extension — if Anthropic deprecates sampling on more models later, we add an entry here.
- `_has_deprecated_sampling_params` lowercases the input and normalizes `.` and `_` separators to `-` before substring-matching, which covers every encoding of the id we found:
  - Native Anthropic: `claude-opus-4-7`, `anthropic/claude-opus-4-7`
  - Bedrock foundation model id: `anthropic.claude-opus-4-7-<date>-v1:0`, with optional `bedrock/converse/...` / `bedrock/invoke/...` route prefixes
  - Bedrock cross-region inference profiles: `us.`, `eu.`, `apac.`, `global.` variants
  - Bedrock foundation-model and inference-profile ARNs that embed the model id
  - Vertex AI: `vertex_ai/claude-opus-4-7@<date>`
  - Azure AI Foundry deployments whose name embeds the model id
- `_strip_deprecated_sampling_params` checks both `model` and `model_id` (Bedrock's separate AIP ARN field) and pops the three sampling params if either one matches.

`AnthropicLLMParameters.validate()`, `AWSBedrockLLMParameters.validate()`, `AzureAIFoundryLLMParameters.validate()` — return `_strip_deprecated_sampling_params(validated)` instead of `validated`.

### Coverage gap (documented in helper docstring)

Two cases the string-based detector cannot cover:

1. Bedrock Application Inference Profile ARNs whose tail is opaque, e.g. `arn:aws:bedrock:...:application-inference-profile/abcd1234`. The string carries no model id. Mitigation: callers should keep the standard model id in either `model` or `model_id`. If both fields hold opaque ARNs, the strip won't fire.
2. Azure AI Foundry deployments whose name does not embed `claude-opus-4-7`. Mitigation: rename the deployment to include the model id, or — if this becomes common — add an explicit UI flag.

No.

- The strip only fires when the model id matches `claude-opus-4-7` after normalization. Every other Anthropic / Bedrock / Azure AI Foundry model retains its `temperature` / `top_p` / `top_k` exactly as before — verified by direct `validate()` calls against `claude-3-5-sonnet`, `claude-opus-4-6`, `claude-sonnet-4-7`, `gpt-4o`, etc.
- `top_p` / `top_k` are not declared fields on the parameter classes today, so Pydantic was already dropping them silently for non-deprecated models. The defensive pop on the validated dict is a no-op for them in the normal path.

## Database Migrations

- None.

## Env Config

- None.

## Relevant Docs

- [What's new in Claude Opus 4.7 — Anthropic](https://platform.claude.com/docs/en/about-claude/models/whats-new-claude-4-7) — sampling parameters removed; `thinking.budget_tokens` also removed (out of scope here).
- [Claude Opus 4.7 — Amazon Bedrock](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-4-7.html)
- [LiteLLM Bedrock — Application Inference Profile usage](https://github.com/BerriAI/litellm/discussions/8905)

## Related Issues or PRs

- N/A — surfaced from the Azure AI Foundry connector test failure for an "azure opus 4.7" deployment.

## Dependencies Versions

- None.

## Notes on Testing

- Helper coverage matrix executed against 24 known Opus 4.7 id formats (native, Bedrock foundation, route-prefixed, cross-region profiles, foundation-model ARN, inference-profile ARN, Vertex `@`-suffixed, Azure deployments, mixed case, dot/underscore separators). All match. Negatives — `claude-opus-4-6`, `claude-sonnet-4-7`, `claude-haiku-4-5`, `gpt-4o`, `gemini-2.0-flash`, opaque AIP ARN, `None`, empty string — all correctly do not match.
- AIP fallback: confirmed strip via `model` when standard id is in `model` and AIP ARN is in `model_id`; confirmed strip via `model_id` when AIP ARN is in `model` and standard id is in `model_id`; confirmed limitation when both fields hold opaque ARNs.
- Adapter end-to-end: `AnthropicLLMParameters.validate()`, `AWSBedrockLLMParameters.validate()`, `AzureAIFoundryLLMParameters.validate()` — temperature stripped for Opus 4.7 inputs, retained for older Claude / non-Claude inputs.
- `pytest tests/ --ignore=tests/file_storage --ignore=tests/patches`: 221 pass. (The unrelated import error in `tests/patches/test_litellm_cohere_timeout.py` is pre-existing on `main`.)
- `ruff format --check` clean on the modified file.

## Screenshots

N/A — backend SDK change.

## Checklist

### DIFF
--- a/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
@@ -17,11 +17,13 @@ from unstract.sdk1.adapters.enums import AdapterTypes
 logger = logging.getLogger(__name__)
 
 # Anthropic models that have deprecated sampling parameters (`temperature`,
-# `top_p`, `top_k`). The patterns are substring-matched against the model id
-# after lowercasing and normalizing `.` / `_` to `-`, so a single entry covers:
+# `top_p`, `top_k`). The patterns are regex-searched against the model id
+# after lowercasing and normalizing `.` / `_` to `-`. The match is anchored at
+# the trailing edge so that unrelated future ids (`claude-opus-4-70`,
+# `claude-opus-4-75`, `claude-opus-4-7verbose`) do not match. A single entry
+# covers every encoding of the id we have observed:
 #   - Native Anthropic              `claude-opus-4-7`, `anthropic/claude-opus-4-7`
 #   - Bedrock foundation model      `anthropic.claude-opus-4-7-<date>-v1:0`
-#   - Bedrock route prefixes        `converse/...`, `invoke/...`
 #   - Bedrock cross-region profile  `us.anthropic.claude-opus-4-7-...`,
 #                                   `eu.`, `apac.`, `global.` variants
 #   - Bedrock foundation-model ARN  `arn:aws:bedrock:<region>::foundation-model/
@@ -30,15 +32,17 @@ logger = logging.getLogger(__name__)
 #                                    inference-profile/us.anthropic.claude-opus-4-7-...`
 #   - Vertex AI                     `vertex_ai/claude-opus-4-7@<date>`
 #   - Azure AI Foundry              deployments whose name embeds `claude-opus-4-7`
+# Leading text (route prefixes like `converse/`, `invoke/`, `bedrock/`) passes
+# through because the regex is anchored only at the trailing edge.
 # Add new entries here when Anthropic deprecates sampling on more models.
-# Patterns are anchored at the trailing edge: the version digit must be
-# followed by end-of-string or one of `-:@/v`, so unrelated future ids like
-# `claude-opus-4-70`, `claude-opus-4-75` do not match. The allowed delimiters
-# cover the date suffix (`-20260101-v1:0`), Vertex `@<date>`, ARN paths, and
-# the `v1:0` version tag.
-# See https://platform.claude.com/docs/en/about-claude/models/whats-new-claude-4-7
+# Trailing anchor allows: end-of-string, or one of `-`/`:`/`@`/`/` (the
+# delimiters used in date suffixes, ARN paths, Vertex `@<date>`, and the
+# `v1:0` tag), or `v` followed by a digit (the version-tag start). A bare
+# `v` is intentionally rejected so alpha continuations like `4-7verbose` do
+# not silently match.
+# See https://docs.claude.com/en/about-claude/models/whats-new-claude-4-7
 _SAMPLING_DEPRECATED_MODEL_PATTERNS: tuple[re.Pattern[str], ...] = (
-    re.compile(r"claude-opus-4-7(?=$|[-:@/v])"),
+    re.compile(r"claude-opus-4-7(?=$|[-:@/]|v\d)"),
 )
 _DEPRECATED_SAMPLING_PARAMS: tuple[str, ...] = ("temperature", "top_p", "top_k")
 # Fields whose value can carry a model id. `model` is universal; `model_id` is
@@ -55,10 +59,12 @@ def _has_deprecated_sampling_params(model: str | None) -> bool:
     Claude Opus 4.7; sending any of them yields a 400 from Anthropic and from
     the providers that proxy it (Bedrock, Azure AI Foundry, Vertex AI).
 
-    The check normalizes case and `.`/`_` separators to `-`, then substring-
-    matches against the patterns. This catches every format that embeds the
-    model id (foundation model ids, cross-region profiles, foundation-model
-    ARNs, inference-profile ARNs, Vertex `@`-suffixed ids).
+    The check normalizes case and `.`/`_` separators to `-`, then regex-
+    searches against the patterns with a trailing-edge boundary, so
+    `claude-opus-4-70` and `claude-opus-4-7verbose` do not match. This
+    catches every format that embeds the model id (foundation model ids,
+    cross-region profiles, foundation-model ARNs, inference-profile ARNs,
+    Vertex `@`-suffixed ids).
 
     It does NOT catch:
     - Bedrock Application Inference Profile ARNs (e.g.
@@ -76,18 +82,41 @@ def _has_deprecated_sampling_params(model: str | None) -> bool:
 
 
 def _strip_deprecated_sampling_params(validated: dict[str, "Any"]) -> dict[str, "Any"]:
-    """Drop sampling params for models that reject them (e.g. Claude Opus 4.7).
+    """Return a copy of `validated` with deprecated sampling params removed.
 
-    Pydantic's `model_dump()` re-emits the default `temperature=0.1` even when
-    the caller never set one, so the field must be popped after validation to
-    keep it off the wire. Checks both `model` and `model_id` so Bedrock callers
-    routing through an Application Inference Profile are covered when the
-    standard model id only appears in `model_id`.
+    The input dict is not mutated. Returns a shallow copy so callers can rely
+    on `before is not after` and follow the file's copy-then-mutate style.
+
+    `temperature` is the load-bearing case: `BaseChatCompletionParameters`
+    declares `temperature: float | None = Field(default=0.1)`, so Pydantic's
+    `model_dump()` re-emits the default even when the caller never set one.
+    `top_p` and `top_k` are not declared fields and are normally dropped by
+    Pydantic, but the strip defends against caller-supplied values flowing
+    through `**adapter_metadata`.
+
+    Checks both `model` and `model_id` so Bedrock callers routing through an
+    Application Inference Profile are covered when the standard model id only
+    appears in `model_id`.
+
+    Contract change: the returned dict may not contain a `temperature` key.
+    Consumers must read via `.get("temperature")`, not `dict["temperature"]`.
     """
-    if any(_has_deprecated_sampling_params(validated.get(f)) for f in _MODEL_ID_FIELDS):
+    result = dict(validated)
+    if any(_has_deprecated_sampling_params(result.get(f)) for f in _MODEL_ID_FIELDS):
         for param in _DEPRECATED_SAMPLING_PARAMS:
-            validated.pop(param, None)
-    return validated
+            result.pop(param, None)
+    elif any(p in result for p in _DEPRECATED_SAMPLING_PARAMS):
+        # Sampling params present but no model id field matched a deprecation
+        # pattern. This is the only failure mode for an opaque Bedrock AIP ARN
+        # or an Azure Foundry deployment whose name omits the model id; without
+        # this breadcrumb the strip-skipped state is indistinguishable from
+        # "never attempted" when debugging the resulting 400 from upstream.
+        logger.debug(
+            "Sampling-param strip skipped; no model id field matched a "
+            "deprecation pattern. Model ids: %s",
+            {f: result.get(f) for f in _MODEL_ID_FIELDS if result.get(f)},
+        )
+    return result
 
 
 def register_adapters(adapters: dict[str, dict[str, "Any"]], adapter_type: str) -> None:
@@ -1021,9 +1050,6 @@ class AzureAIFoundryLLMParameters(BaseChatCompletionParameters):
         )
 
         validated = AzureAIFoundryLLMParameters(**adapter_metadata).model_dump()
-        # Azure AI Foundry proxies Anthropic models that reject sampling params
-        # (e.g. Claude Opus 4.7); detection relies on the deployment name
-        # embedding the model id.
         return _strip_deprecated_sampling_params(validated)
 
     @staticmethod
--- a/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
@@ -2,6 +2,7 @@ import glob
 import inspect
 import logging
 import os
+import re
 from abc import ABC, abstractmethod
 from importlib import import_module
 from typing import TYPE_CHECKING
@@ -30,8 +31,15 @@ logger = logging.getLogger(__name__)
 #   - Vertex AI                     `vertex_ai/claude-opus-4-7@<date>`
 #   - Azure AI Foundry              deployments whose name embeds `claude-opus-4-7`
 # Add new entries here when Anthropic deprecates sampling on more models.
+# Patterns are anchored at the trailing edge: the version digit must be
+# followed by end-of-string or one of `-:@/v`, so unrelated future ids like
+# `claude-opus-4-70`, `claude-opus-4-75` do not match. The allowed delimiters
+# cover the date suffix (`-20260101-v1:0`), Vertex `@<date>`, ARN paths, and
+# the `v1:0` version tag.
 # See https://platform.claude.com/docs/en/about-claude/models/whats-new-claude-4-7
-_SAMPLING_DEPRECATED_MODEL_PATTERNS: tuple[str, ...] = ("claude-opus-4-7",)
+_SAMPLING_DEPRECATED_MODEL_PATTERNS: tuple[re.Pattern[str], ...] = (
+    re.compile(r"claude-opus-4-7(?=$|[-:@/v])"),
+)
 _DEPRECATED_SAMPLING_PARAMS: tuple[str, ...] = ("temperature", "top_p", "top_k")
 # Fields whose value can carry a model id. `model` is universal; `model_id` is
 # Bedrock's separate ARN field used for Application Inference Profile cost
@@ -64,7 +72,7 @@ def _has_deprecated_sampling_params(model: str | None) -> bool:
     if not model:
         return False
     normalized = model.lower().replace(".", "-").replace("_", "-")
-    return any(pat in normalized for pat in _SAMPLING_DEPRECATED_MODEL_PATTERNS)
+    return any(rx.search(normalized) for rx in _SAMPLING_DEPRECATED_MODEL_PATTERNS)
 
 
 def _strip_deprecated_sampling_params(validated: dict[str, "Any"]) -> dict[str, "Any"]:
@@ -528,7 +536,7 @@ class VertexAILLMParameters(BaseChatCompletionParameters):
         if "thinking" in result_metadata:
             validated_data["thinking"] = result_metadata["thinking"]
 
-        return validated_data
+        return _strip_deprecated_sampling_params(validated_data)
 
     @staticmethod
     def validate_model(adapter_metadata: dict[str, "Any"]) -> str:
--- a/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
@@ -15,6 +15,72 @@ from unstract.sdk1.adapters.enums import AdapterTypes
 
 logger = logging.getLogger(__name__)
 
+# Anthropic models that have deprecated sampling parameters (`temperature`,
+# `top_p`, `top_k`). The patterns are substring-matched against the model id
+# after lowercasing and normalizing `.` / `_` to `-`, so a single entry covers:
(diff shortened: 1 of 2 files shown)


unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -15,6 +16,79 @@
...
+    if not model:
+        return False
+    normalized = model.lower().replace(".", "-").replace("_", "-")
+    return any(rx.search(normalized) for rx in _SAMPLING_DEPRECATED_MODEL_PATTERNS)
+
+
+def _strip_deprecated_sampling_params(validated: dict[str, "Any"]) -> dict[str, "Any"]:
+    """Drop sampling params for models that reject them (e.g. Claude Opus 4.7).
+
+    Pydantic's `model_dump()` re-emits the default `temperature=0.1` even when
```
chandrasekharan-zipstack: @pk-zipstack are we doing this because litellm has not handled this yet? Are we using a version of litellm that also supports such models cleanly? We shouldn't clean up the params from our side if one of the latest versions of litellm supports it better

chandrasekharan-zipstack (approved):
@pk-zipstack will we still show the temperature field in the UI for such models?

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
praveen-formido: Tested with litellm's latest version and the clean up is not done from their end. The PR is valid @chandrasekharan-zipstack

Deepak-Kesavan (approved):

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -15,6 +16,79 @@
...
+#   - Azure AI Foundry              deployments whose name embeds `claude-opus-4-7`
+# Add new entries here when Anthropic deprecates sampling on more models.
+# Patterns are anchored at the trailing edge: the version digit must be
+# followed by end-of-string or one of `-:@/v`, so unrelated future ids like
+# `claude-opus-4-70`, `claude-opus-4-75` do not match. The allowed delimiters
+# cover the date suffix (`-20260101-v1:0`), Vertex `@<date>`, ARN paths, and
+# the `v1:0` version tag.
+# See https://platform.claude.com/docs/en/about-claude/models/whats-new-claude-4-7
+_SAMPLING_DEPRECATED_MODEL_PATTERNS: tuple[re.Pattern[str], ...] = (
+    re.compile(r"claude-opus-4-7(?=$|[-:@/v])"),
```
jaseemjaskp: **HIGH — anchor lookahead admits `v`-prefixed words, not just version tags**

`(?=$|[-:@/v])` puts a bare `v` in the character class, so the trailing-edge boundary fires for *any* token starting with the letter `v` — e.g. `claude-opus-4-7verbose`, `claude-opus-4-7vnext`, or any future suffix-with-`v` id would match (currently theoretical, but the comment on lines 34–38 explicitly claims the anchor blocks unrelated future ids).

**Fix:** require the version pattern explicitly, e.g.
```python
re.compile(r"claude-opus-4-7(?=$|[-:@/]|v\d)")
```
Keeps `…-v1:0` / `…v1` valid while rejecting alpha continuations starting with `v`. Extends the anchor concern raised earlier by greptile rather than duplicating it.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -15,6 +16,79 @@
...
+
+    Pydantic's `model_dump()` re-emits the default `temperature=0.1` even when
+    the caller never set one, so the field must be popped after validation to
+    keep it off the wire. Checks both `model` and `model_id` so Bedrock callers
+    routing through an Application Inference Profile are covered when the
+    standard model id only appears in `model_id`.
+    """
+    if any(_has_deprecated_sampling_params(validated.get(f)) for f in _MODEL_ID_FIELDS):
+        for param in _DEPRECATED_SAMPLING_PARAMS:
+            validated.pop(param, None)
```
jaseemjaskp: **HIGH — silent no-op when sampling params are present but no model-id field matches**

If a Bedrock Application Inference Profile ARN with an opaque tail (or an Azure Foundry deployment whose name omits the model id) reaches this helper while `temperature`/`top_p`/`top_k` are populated, the strip silently returns the dict unchanged. The failure surfaces only as a 400 from Anthropic *after* dispatch, with no breadcrumb tying it back to the strip path. Six months from now an operator hitting that 400 has no way to know detection was attempted-and-skipped vs. never attempted.

**Fix:** add a single `logger.debug` (or `logger.warning` if you want it to page operators) on the no-strip branch *only when* at least one of `_DEPRECATED_SAMPLING_PARAMS` is present in `validated`:
```python
if matched:
    for p in _DEPRECATED_SAMPLING_PARAMS:
        validated.pop(p, None)
elif any(p in validated for p in _DEPRECATED_SAMPLING_PARAMS):
    logger.debug(
        "Sampling-strip skipped; model ids did not match deprecation patterns: %s",
        {f: validated.get(f) for f in _MODEL_ID_FIELDS if validated.get(f)},
    )
```
Keeps the AIP/Foundry fallback non-fatal but observable.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -15,6 +16,79 @@
...
+    """Drop sampling params for models that reject them (e.g. Claude Opus 4.7).
+
+    Pydantic's `model_dump()` re-emits the default `temperature=0.1` even when
+    the caller never set one, so the field must be popped after validation to
+    keep it off the wire. Checks both `model` and `model_id` so Bedrock callers
+    routing through an Application Inference Profile are covered when the
+    standard model id only appears in `model_id`.
+    """
+    if any(_has_deprecated_sampling_params(validated.get(f)) for f in _MODEL_ID_FIELDS):
+        for param in _DEPRECATED_SAMPLING_PARAMS:
```
jaseemjaskp, line 121: **MED — post-strip `temperature` key invariant breaks `validated["temperature"]` consumers**

`BaseChatCompletionParameters.temperature` defaults to `Field(default=0.1)`, so before this PR every `validate()` return was guaranteed to carry `temperature`. After the strip, callers using `validated["temperature"]` (vs. `.get`) on an Opus 4.7 path now raise `KeyError`. Anything that reads this key for cost-tracking, telemetry, or UI display in the platform service / SDK runtime would silently break only on 4.7.

**Fix:** grep platform-service / SDK consumers of `validate()` output for direct `["temperature"]` access and switch to `.get("temperature")`; document in this docstring that `temperature` may be absent post-strip so future consumers know the key is now optional.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -15,6 +16,79 @@
...
+    - Azure AI Foundry deployment names that omit the model id; rename the
+      deployment to include `claude-opus-4-7` so detection works.
+    """
+    if not model:
+        return False
+    normalized = model.lower().replace(".", "-").replace("_", "-")
+    return any(rx.search(normalized) for rx in _SAMPLING_DEPRECATED_MODEL_PATTERNS)
+
+
+def _strip_deprecated_sampling_params(validated: dict[str, "Any"]) -> dict[str, "Any"]:
```
jaseemjaskp, line 99: **MED — mutate-and-return is inconsistent with surrounding callers**

The rest of this file consistently builds `result_metadata = adapter_metadata.copy()` (e.g. lines 264, 350, 584) when avoiding mutation. This helper mutates `validated` in place *and* returns the same dict; today’s four call sites are safe, but the asymmetry invites a future caller to do `before = validated; after = _strip_deprecated_sampling_params(validated)` and be surprised that `before is after`.

**Fix (lowest-risk):** rename to `_strip_deprecated_sampling_params_inplace` and return `None`, or shallow-copy and return the copy. The dict is small — a copy costs nothing and matches surrounding style.

*Type-design note:* invariant strength would also improve if the `pattern ↔ forbidden-params` link were structural rather than positional (parallel `_SAMPLING_DEPRECATED_MODEL_PATTERNS` and `_DEPRECATED_SAMPLING_PARAMS` tuples). Worth folding into `((pattern, frozenset(params)), …)` *only when* a second deprecation lands with a different param set — not before.

jaseemjaskp (commented):
Automated review pass (PR Review Toolkit: code-reviewer, silent-failure-hunter, type-design-analyzer, comment-analyzer, code-simplifier, pr-test-analyzer). Findings posted inline; deduplicated against existing greptile/CodeRabbit/maintainer comments. Grouped by line and roughly prioritised (HIGH → LOW/optional).

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -15,6 +16,79 @@
...
+    Pydantic's `model_dump()` re-emits the default `temperature=0.1` even when
+    the caller never set one, so the field must be popped after validation to
+    keep it off the wire. Checks both `model` and `model_id` so Bedrock callers
+    routing through an Application Inference Profile are covered when the
+    standard model id only appears in `model_id`.
+    """
+    if any(_has_deprecated_sampling_params(validated.get(f)) for f in _MODEL_ID_FIELDS):
+        for param in _DEPRECATED_SAMPLING_PARAMS:
+            validated.pop(param, None)
+    return validated
```
jaseemjaskp: **MED — zero committed tests for the new helpers or the four wired call sites**

`git diff origin/main…HEAD --name-only` shows only `base1.py` — the PR description’s 24-positive / 8-negative / AIP-fallback / four-end-to-end matrix is ad-hoc local. Repo-wide grep for `claude-opus-4-7`, `_strip_deprecated_sampling_params`, `_has_deprecated_sampling_params` matches only this file. The Vertex AI wiring gap that surfaced in review (commit `5a4ea27f` shipped without it; addressed in `7fb66f15`) is exactly the failure mode a parametrized adapter test would have caught.

**Suggested minimal lock-in** (`unstract/sdk1/tests/adapters/test_base1_sampling_strip.py`):
- `test_has_deprecated_sampling_params_positive` — `pytest.mark.parametrize` over the 24 id formats from the PR description (native, `anthropic.` Bedrock, `us./eu./apac./global.` cross-region, foundation-model ARN, inference-profile ARN, `vertex_ai/...@<date>`, Azure deployment, `converse/`, `invoke/`, `.`/`_` separators, mixed case).
- `test_has_deprecated_sampling_params_negative` — `claude-opus-4-6`, `claude-opus-4-70`, `claude-opus-4-75`, `claude-sonnet-4-7`, `claude-haiku-4-5`, `gpt-4o`, `gemini-2.0-flash`, opaque AIP ARN, `None`, `""`. The `4-70`/`4-75` cases lock the `(?=$|[-:@/v])` anchor.
- `test_strip_checks_model_id_field_fallback` — opus id in `model` only, in `model_id` only, both opaque (no strip).
- `test_validate_strips_for_<adapter>` — four parametrized cases (`AnthropicLLMParameters`, `AWSBedrockLLMParameters`, `AzureAIFoundryLLMParameters`, `VertexAILLMParameters`), each asserting the three keys absent for opus 4.7 / present for opus 4.6. A future refactor that drops a single call site regresses silently without these.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -15,6 +16,79 @@
 
 logger = logging.getLogger(__name__)
 
+# Anthropic models that have deprecated sampling parameters (`temperature`,
+# `top_p`, `top_k`). The patterns are substring-matched against the model id
```
jaseemjaskp: **LOW — comment drift: “substring-matched” no longer accurate**

Lines 20 and 58 still describe matching as “substring-matched against the model id”, but the implementation now uses `re.compile(...)` + `.search()` with a trailing-edge lookahead. A pure substring match would (correctly) flag `claude-opus-4-70`; the current regex does not. Reword as e.g. “regex-searched against the normalized id, with a trailing-edge boundary so that `claude-opus-4-70` does not match.” Keeps the comment honest after the anchor change in `7fb66f15`.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -15,6 +16,79 @@
...
+# after lowercasing and normalizing `.` / `_` to `-`, so a single entry covers:
+#   - Native Anthropic              `claude-opus-4-7`, `anthropic/claude-opus-4-7`
+#   - Bedrock foundation model      `anthropic.claude-opus-4-7-<date>-v1:0`
+#   - Bedrock route prefixes        `converse/...`, `invoke/...`
```
jaseemjaskp: **LOW — misleading bullet: route prefixes are not part of the pattern**

“Bedrock route prefixes `converse/...`, `invoke/...`” reads as if the regex enumerates prefixes; it does not. The regex matches the `claude-opus-4-7` token regardless of leading text, so prefixes like `converse/` / `invoke/` / `bedrock/converse/` are accepted only as a side-effect of the trailing-edge-only anchor. Either drop the bullet or rephrase as “prefixes such as `converse/` and `invoke/` pass through because the match is anchored only at the trailing edge.”

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -15,6 +16,79 @@
...
+#                                    inference-profile/us.anthropic.claude-opus-4-7-...`
+#   - Vertex AI                     `vertex_ai/claude-opus-4-7@<date>`
+#   - Azure AI Foundry              deployments whose name embeds `claude-opus-4-7`
+# Add new entries here when Anthropic deprecates sampling on more models.
+# Patterns are anchored at the trailing edge: the version digit must be
+# followed by end-of-string or one of `-:@/v`, so unrelated future ids like
+# `claude-opus-4-70`, `claude-opus-4-75` do not match. The allowed delimiters
+# cover the date suffix (`-20260101-v1:0`), Vertex `@<date>`, ARN paths, and
+# the `v1:0` version tag.
+# See https://platform.claude.com/docs/en/about-claude/models/whats-new-claude-4-7
```
jaseemjaskp: **LOW — verify canonical docs URL**

`https://platform.claude.com/docs/en/about-claude/models/whats-new-claude-4-7` — the canonical Anthropic docs host is `docs.claude.com` / `docs.anthropic.com`; `platform.claude.com` is the console host and may not serve this docs path stably. Worth checking the link resolves before merge so the rationale comment doesn’t rot.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
jaseemjaskp: **LOW — docstring “why” understates the scope of the strip**

“Pydantic’s `model_dump()` re-emits the default `temperature=0.1`” is the load-bearing reason for popping `temperature`, but `top_p` / `top_k` are *not* declared fields on `BaseChatCompletionParameters` and so are not subject to the same Pydantic-default re-emit. The strip’s reason for handling them is to defend against caller-supplied values flowing through `**adapter_metadata`. Worth one extra clause so future readers understand both halves of the invariant.

(Distinct from the existing pk-zipstack / chandrasekharan-zipstack thread on this line about whether litellm handles the cleanup upstream.)

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -15,6 +16,79 @@
...
+_SAMPLING_DEPRECATED_MODEL_PATTERNS: tuple[re.Pattern[str], ...] = (
```
jaseemjaskp, line 44: **OPTIONAL (simplify) — tuple-of-patterns + `_MODEL_ID_FIELDS` are over-abstracted for one entry today**

With exactly one regex and exactly two id-bearing fields, the indirection costs more than it saves. If you want to reduce surface:
```python
_SAMPLING_DEPRECATED_MODEL_RE = re.compile(r"claude-opus-4-7(?=$|[-:@/v])")
…
def _strip_deprecated_sampling_params(validated):
    if _has_deprecated_sampling_params(validated.get("model")) \
       or _has_deprecated_sampling_params(validated.get("model_id")):
        for param in _DEPRECATED_SAMPLING_PARAMS:
            validated.pop(param, None)
    return validated
```
Going back to a tuple/loop when (if) a second deprecation lands is a 2-line change. Reasonable people will disagree — the current shape is readable; this is purely a YAGNI call. Skip if you prefer keeping the table for forward-compat.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -837,7 +911,11 @@ def validate(adapter_metadata: dict[str, "Any"]) -> dict[str, "Any"]:
             adapter_metadata
         )
 
-        return AzureAIFoundryLLMParameters(**adapter_metadata).model_dump()
+        validated = AzureAIFoundryLLMParameters(**adapter_metadata).model_dump()
+        # Azure AI Foundry proxies Anthropic models that reject sampling params
+        # (e.g. Claude Opus 4.7); detection relies on the deployment name
+        # embedding the model id.
```
jaseemjaskp: **LOW — inline comment is redundant *and* understates the failure mode**

The helper docstring already documents the Foundry deployment-name caveat, so the three lines here mostly restate it. If the comment stays, strengthen the second half: “detection relies on the deployment name embedding the Anthropic model id; opaque deployment names skip the strip and will surface as a 400 from the provider.” Otherwise delete — the helper call reads self-explanatorily.

jaseemjaskp (approved):
LGTM on the sampling-strip fix. The trailing-edge anchor, the `model_id` fallback, the four-adapter wiring, and the new test suite all address the prior round of review feedback. The one remaining outstanding concern on this PR — greptile-apps P1 (debug breadcrumb fires for every non-Opus-4.7 call because `BaseChatCompletionParameters.temperature` has `Field(default=0.1)`, so `model_dump()` always emits it) — is a real, valid point and worth addressing before merge, but it is already posted and does not need duplication from me. Nothing additional to raise.

PR #1941, opened 2026-05-05
UN-3448 [FIX] Remove vestigial `uv pip install` line in uv-lock-automation workflow

muhammad-ali-e:
## What

- Remove the vestigial `uv pip install --python=3.12.9 pip` line from `.github/workflows/uv-lock-automation.yaml`.

## Why

The line was failing the workflow on every PR that touches `pyproject.toml` since 2026-04-01:

```
error: No virtual environment found for Python 3.12.9; run `uv venv` to
create an environment, or pass `--system` to install into a non-virtual
environment
```

The error appeared because `1997ce31e` (PR #1883) bumped `setup-uv@v5 → @v7`, and v7 installs a uv-managed Python that isn't seen as system Python.

**Initial fix proposed:** add `--system` flag (matching the pattern applied to other workflows in `1997ce31e`).

**Better fix per [review](https://github.com/Zipstack/unstract/pull/1941#pullrequestreview-4227913262):** the line itself is vestigial. The downstream `docker/scripts/uv-lock-gen/uv-lock.sh` uses `uv sync` at line 74, which manages its own venvs internally and never invokes `pip` directly:

```
$ grep -rn 'pip' docker/scripts/uv-lock-gen/
docker/scripts/uv-lock-gen/uv-lock.sh:2:set -o pipefail
```

Only match is `pipefail` (shell option), no real `pip` references. Pre-installing pip into the system Python serves no purpose for this workflow — the line was likely copy-pasted from a sibling workflow that legitimately needed pip.

Reference: UN-3448.

## How

```diff
       - name: Install uv
         uses: astral-sh/setup-uv@v7
         with:
           # Install a specific version of uv.
           version: "0.6.14"
           python-version: 3.12.9

-      - run: uv pip install --python=3.12.9 pip
-
       - name: Generate UV lockfiles
```

Two-line deletion. No replacement needed.

**No.** The deleted line did not contribute to the workflow's purpose:
- The lock-generation script never used `pip` (verified by grep).
- The line has been failing 100% of the time since 2026-04-01, so removing it cannot regress anything that was previously working.

The workflow itself is non-blocking (other required checks gate merges), so worst case is "still broken" — but verification above shows the workflow will now successfully run `uv sync` for each directory.

## Database Migrations

None.

## Env Config

None.

## Related Issues or PRs

- UN-3448
- Original setup-uv migration: `1997ce31e` (PR #1883)
- Review that prompted the better fix: PR #1941, review `4227913262`

## Dependencies Versions

None changed.

## Notes on Testing

- Local pre-commit on the changed file passed.
- The workflow runs in CI on PRs that modify a `pyproject.toml`. This PR modifies a workflow file, not a pyproject, so the trigger won't fire here. The next PR that touches any `pyproject.toml` after this merges will be the real validation.
- Alternative: `workflow_dispatch` is available on this workflow — a maintainer can run it post-merge to confirm.

## Screenshots

N/A — workflow file change.

## Checklist

- [x] I have solved all the sonar issues
- [ ] I have made corresponding changes to the documentation (N/A — workflow change, no doc impact)
- [ ] I have added tests that prove my fix is effective or that my feature works (N/A — CI workflow validates itself on the next pyproject.toml-touching PR)
- [x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/.github/workflows/uv-lock-automation.yaml
+++ b/.github/workflows/uv-lock-automation.yaml
@@ -36,7 +36,7 @@ jobs:
           version: "0.6.14"
           python-version: 3.12.9
 
-      - run: uv pip install --python=3.12.9 pip
+      - run: uv pip install --system --python=3.12.9 pip
 
       - name: Generate UV lockfiles
         env:
--- a/.github/workflows/uv-lock-automation.yaml
+++ b/.github/workflows/uv-lock-automation.yaml
@@ -36,8 +36,6 @@ jobs:
           version: "0.6.14"
           python-version: 3.12.9
 
-      - run: uv pip install --system --python=3.12.9 pip
-
       - name: Generate UV lockfiles
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


pk-zipstack (approved):

jaseemjaskp (changes requested):
uv-lock.sh` only uses `uv sync` (line 74). It never invokes `pip` directly, and `uv sync` manages its own environments via uv. There's no reason to pre-install pip into the system Python at all. Confirmed by grepping — `pip` is referenced nowhere in `docker/scripts/uv-lock-gen/`.

The line looks vestigial. The cleaner fix is to **delete the line entirely** rather than paper over with `--system`:

muhammad-ali-e:
Good catch — verified `uv-lock.sh` only uses `uv sync` (line 74) and `grep -rn pip docker/scripts/uv-lock-gen/` returns just `pipefail` (the shell option). The pre-install was doing nothing useful. Removed the line entirely as suggested. Thanks!

jaseemjaskp (approved):

PR #1942, opened 2026-05-05
[FEAT] Add DISABLE_SSO_IDP_AUTHORIZATION flag for SSO user mgmt

johnyrahul:
## What

- Adds a new Django setting `DISABLE_SSO_IDP_AUTHORIZATION` (default `False`).
- Exposes the flag in the `GET /api/v1/session` response (`UserSessionInfo` → `UserSessionResponseSerializer`).
- Hydrates it on the frontend (`useSessionValid` → `getSessionData` → `sessionDetails.disableSsoIdpAuthorization`).
- Updates Manage Users (`Users.jsx`) to gate the Actions menu and Invite button.

## Why

When users are authenticated via an external provider (e.g. enterprise SSO via Auth0), the existing flow hides the entire Manage Users action surface. With this flag enabled, admins can still **edit a user's role locally** while invite/delete remain blocked — useful when SSO/IdP authorization is disabled and authorization is managed inside Unstract.

## How

- `backend/backend/settings/base.py` — declare the new `DISABLE_SSO_IDP_AUTHORIZATION` setting (defaults to `False`).
- `backend/account_v2/dto.py` — add `disable_sso_idp_authorization: bool = False` to `UserSessionInfo` and propagate it through `from_dict` / `to_dict`.
- `backend/account_v2/serializer.py` — add the matching `BooleanField` on `UserSessionResponseSerializer`.
- `backend/account_v2/views.py` — pass `settings.DISABLE_SSO_IDP_AUTHORIZATION` into `UserSessionInfo` in `make_session_response`.
- `frontend/src/hooks/useSessionValid.js` — copy `disable_sso_idp_authorization` into `userAndOrgDetails`.
- `frontend/src/helpers/GetSessionData.js` — surface it as `sessionDetails.disableSsoIdpAuthorization`.
- `frontend/src/components/settings/users/Users.jsx`:
  - New `isSsoLocalAuthz = !!provider && !!disableSsoIdpAuthorization` derived value.
  - Actions column shows when `!provider || isSsoLocalAuthz`.
  - When `isSsoLocalAuthz`, the dropdown contains only the **Edit** item.
  - Invite User button stays gated on `!provider` (unchanged).

| `provider` set | flag | Invite | Edit | Delete |
|---|---|---|---|---|
| no | any | yes | yes | yes |
| yes | False (default) | no | no | no |
| yes | True | no | yes | no |

- No. Default value is `False`, which preserves the prior behavior (Actions column and Invite button hidden for any session with `provider`). Behavior only changes when an operator explicitly opts in by setting `DISABLE_SSO_IDP_AUTHORIZATION=True`.

## Database Migrations

- None.

## Env Config

- New optional Django setting: `DISABLE_SSO_IDP_AUTHORIZATION` (default `False`).

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

- None changed.

## Notes on Testing

- With flag `False` and an SSO-authenticated session: confirm Manage Users hides both Invite User and the Actions column (existing behavior).
- With flag `True` and an SSO-authenticated session: confirm Manage Users hides Invite User, shows the Actions column, and the row dropdown contains only **Edit** (no Delete).
- With no provider on the session: confirm the page is unchanged — Invite User visible, Actions dropdown contains both Edit and Delete.
- Verify `GET /api/v1/session` response now includes `disable_sso_idp_authorization` boolean.

## Checklist


backend/account_v2/dto.py
```diff
@@ -62,6 +62,7 @@ class UserSessionInfo:
     role: str
     provider: str
     is_staff: bool = False
+    disable_sso_idp_authorization: bool = False
```
johnyrahul, line 65: **[type-design / naming]** Two design issues worth reconsidering before this lands on the wire:

1. **Double-negative naming.** `disable_sso_idp_authorization=True` *enables* a capability (local user management). Every consumer ends up in a positive-from-negative dance: see Users.jsx:100 — `!!provider && !!disableSsoIdpAuthorization`. Consider a positive name (e.g. `local_sso_user_management_enabled`) and inverting at the settings boundary in `make_session_response`, or expose a derived `can_locally_manage_sso_users` property so callers don't reconstruct the same condition.
2. **Invariant not encoded.** This flag is only meaningful when `provider` is truthy; today that constraint lives in caller code (the frontend has to remember the `&&` join). A `@property` that ANDs them together would prevent misuse in non-SSO contexts.

backend/account_v2/dto.py
```diff
@@ -73,6 +74,9 @@ def from_dict(data: dict[str, Any]) -> "UserSessionInfo":
             role=data["role"],
             provider=data["provider"],
             is_staff=data.get("is_staff", False),
+            disable_sso_idp_authorization=data.get(
+                "disable_sso_idp_authorization", False
```
johnyrahul, line 78: **[silent-failure / high]** This security-sensitive flag uses `data.get(..., False)` while every other required session field above (`role`, `provider`, `email`, etc.) uses `data["..."]` — i.e. fails loudly if absent. With this asymmetry, a contract drift (renamed key, partial deploy) silently produces `False`, which combined with the frontend `!!` checks fails *open*: SSO users regain the Delete action.

Suggest matching the surrounding pattern: `disable_sso_idp_authorization=data["disable_sso_idp_authorization"]`. If the field genuinely is optional for backwards compatibility, default to `True` (fail-closed) and emit a logger warning so the missing-key case is observable.

backend/account_v2/dto.py
```diff
@@ -83,6 +87,7 @@ def to_dict(self) -> Any:
             "organization_id": self.organization_id,
             "role": self.role,
             "is_staff": self.is_staff,
+            "disable_sso_idp_authorization": self.disable_sso_idp_authorization,
```
johnyrahul, line 90: **[symmetry / high]** `to_dict()` writes the new field correctly, but it still omits `email` and `provider` while `from_dict()` requires both via `data["..."]`. Round-tripping `UserSessionInfo.from_dict(obj.to_dict())` raises `KeyError` today. Pre-existing, but this PR is a good moment to fix it (or to delete `to_dict()` entirely and let `UserSessionResponseSerializer` own serialization — the two have already drifted and will drift again on the next field).

Minimum fix: add `"email": self.email` and `"provider": self.provider`. Better: drive both methods from `dataclasses.fields()` so symmetry is structural rather than hand-maintained.

backend/account_v2/serializer.py
```diff
@@ -127,3 +127,4 @@ class UserSessionResponseSerializer(serializers.Serializer):
     role = serializers.CharField()
     provider = serializers.CharField()
     is_staff = serializers.BooleanField()
+    disable_sso_idp_authorization = serializers.BooleanField()
```
johnyrahul, line 130: **[contract / nit]** Pin this explicitly: `serializers.BooleanField(required=True)`. DRF's default is already `required=True`, but spelling it out documents intent and makes it harder for a future change to silently drop to `default=False`, which would compound with the `.get(..., False)` in `dto.py` to mask contract drift on both sides.

A `help_text` describing the conditional-on-provider semantics would also help API consumers — this is a public response field now.

backend/account_v2/views.py
```diff
@@ -151,6 +152,7 @@ def make_session_response(
             role=UserSessionUtils.get_organization_member_role(request),
             provider=provider,
             is_staff=request.user.is_staff,
+            disable_sso_idp_authorization=settings.DISABLE_SSO_IDP_AUTHORIZATION,
```
johnyrahul, line 155: **[robustness / medium]** `settings.DISABLE_SSO_IDP_AUTHORIZATION` will `AttributeError` for every login if a deployment-specific settings module (prod, staging) doesn't redefine or import the base value. Safer:

```python
disable_sso_idp_authorization=getattr(
    settings, "DISABLE_SSO_IDP_AUTHORIZATION", False
),
```

Also worth coercing/validating type — if someone wires this to an env var via `os.environ[...]` later (a string), DRF's `BooleanField` will silently coerce truthy strings, flipping authorization behavior.

backend/backend/settings/base.py
```diff
@@ -434,6 +434,8 @@ def filter(self, record):
 TENANT_SUBFOLDER_PREFIX = f"{PATH_PREFIX}/unstract"
 SHOW_PUBLIC_IF_NO_TENANT_FOUND = True
 
+DISABLE_SSO_IDP_AUTHORIZATION = False
```
johnyrahul: **[docs / medium]** A bare top-level setting with a non-obvious name. "Disable SSO IdP authorization" reads like "turn off SSO", but the actual semantics are "keep SSO authentication, move authorization (user mgmt) into Unstract". A future maintainer scanning `base.py` has zero context.

Suggested comment:

```python
# When True, users still authenticate via the SSO IdP, but user management
# (invite/edit/delete) is handled locally in Unstract instead of being delegated
# to the IdP. Enable for SSO deployments where the IdP doesn't expose user
# provisioning APIs. Default False preserves "IdP owns user lifecycle".
DISABLE_SSO_IDP_AUTHORIZATION = False
```

frontend/src/components/settings/users/Users.jsx
```diff
@@ -96,46 +96,51 @@ function Users() {
...
-          </div>
-          <div>
-            <Typography.Text>Delete</Typography.Text>
-          </div>
-        </Space>
-      ),
-    },
-  ];
+  const isSsoLocalAuthz =
+    !!sessionDetails?.provider && !!sessionDetails?.disableSsoIdpAuthorization;
```
johnyrahul, line 100: **[silent-failure / high]** `!!sessionDetails?.disableSsoIdpAuthorization` collapses `undefined` and `false` into the same branch. If the backend response is malformed, predates this rollout, or `GetSessionData.js` ever loses the mapping, `disableSsoIdpAuthorization` becomes `undefined` → `isSsoLocalAuthz` becomes `false` → `actionItems` falls through to `[editItem, deleteItem]` (line 142), and SSO users see Delete again. That's the inverse of the feature's security intent, with no log to surface it.

Distinguish missing from false:

```js
const isSsoLocalAuthz =
  !!sessionDetails?.provider &&
  sessionDetails?.disableSsoIdpAuthorization === true;
```

Also log via your error-id helper when `provider` is set but the flag is `undefined` — that's a contract bug worth seeing in Sentry.

(Style nit while you're here: the `!!` on `sessionDetails?.provider` is unnecessary — `&&` short-circuits on falsy strings cleanly.)

frontend/src/components/settings/users/Users.jsx
```diff
@@ -96,46 +96,51 @@ function Users() {
...
+          <DeleteOutlined />
+        </div>
+        <div>
+          <Typography.Text>Delete</Typography.Text>
+        </div>
+      </Space>
+    ),
+  };
+
+  const actionItems = isSsoLocalAuthz ? [editItem] : [editItem, deleteItem];
```
johnyrahul, line 142: **[tests / high]** This is the single line implementing the security-relevant gating matrix from the PR description, and there's no test for it. The whole spec is:

| provider | flag | Invite | Edit | Delete |
|---|---|---|---|---|
| no | any | yes | yes | yes |
| yes | False | no | no | no |
| yes | True | no | yes | no |

A `Users.test.jsx` (vitest + @testing-library is already configured in `vitest.config.mjs`) covering the four cells — including `provider=set, flag=undefined` to lock down fail-closed behavior — would prevent a one-character regression from silently shipping. Recommend adding before merge.

frontend/src/components/settings/users/Users.jsx
```diff
@@ -165,9 +170,10 @@ function Users() {
     ),
   };
 
-  const columns = !sessionDetails?.provider
-    ? [...baseColumns, actionColumn]
-    : baseColumns;
+  const columns =
+    !sessionDetails?.provider || isSsoLocalAuthz
```
johnyrahul, line 174: **[readability / low]** This compound condition encodes two distinct cases ("no SSO at all" and "SSO with local authz") in one boolean — a future maintainer is likely to "simplify" it and break one path. Either name the predicate or add a comment:

```js
const showActionsColumn =
  !sessionDetails?.provider || isSsoLocalAuthz;
const columns = showActionsColumn
  ? [...baseColumns, actionColumn]
  : baseColumns;
```

johnyrahul (commented):

Multi-agent review run: `code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`, `type-design-analyzer`, `code-simplifier`.

**Top concerns — most are fail-open patterns around an authorization flag:**

1. The flag silently defaults to `False` on every layer (DTO `.get()`, frontend optional-chain, `!!` coercion). If the backend ever drops the field from the session response, every SSO user silently regains Delete — the inverse of the feature's security intent.
2. `UserSessionInfo.to_dict()` drops `email` and `provider`, breaking the round-trip with `from_dict()`. Pre-existing, but this PR perpetuates the pattern.
3. No tests exist for the dataclass round-trip, the view wiring of `settings.DISABLE_SSO_IDP_AUTHORIZATION`, or the Users.jsx gating matrix that the PR description spells out as the spec — see standalone summary comments below.
4. The setting name is a double-negative (`DISABLE_..._AUTHORIZATION` true ⇒ *enables* a capability) and the conditional-on-provider invariant is not encoded in the type — both worth addressing now before the wire shape is locked in.

Nothing here is a merge-blocker on its own; the fail-open pattern in particular deserves a quick tightening before this ships to a real SSO tenant.

frontend/src/helpers/GetSessionData.js
```diff
@@ -28,6 +28,7 @@ function getSessionData(sessionData) {
     role: sessionData?.role,
     provider: sessionData?.provider,
     isStaff: sessionData?.is_staff,
+    disableSsoIdpAuthorization: sessionData?.disable_sso_idp_authorization,
```
johnyrahul, line 31: **[silent-failure / high]** `sessionData?.disable_sso_idp_authorization` returns `undefined` if the backend payload omits the field. That `undefined` is propagated to `Users.jsx` where `!!` coerces it to `false`, which (per Users.jsx:142) drops back to `[editItem, deleteItem]` — Delete is silently re-granted to SSO users. Same fail-open hazard noted on Users.jsx:100.

Fail-closed default + observability:

```js
disableSsoIdpAuthorization:
  typeof sessionData?.disable_sso_idp_authorization === "boolean"
    ? sessionData.disable_sso_idp_authorization
    : true,
```

or leave it `undefined` here and tighten the consumer to `=== true` (see Users.jsx:100 comment) — pick one boundary.

frontend/src/hooks/useSessionValid.js
```diff
@@ -167,6 +167,8 @@ function useSessionValid() {
       }
       userAndOrgDetails["role"] = userSessionData.role;
       userAndOrgDetails["provider"] = userSessionData.provider;
+      userAndOrgDetails["disable_sso_idp_authorization"] =
+        userSessionData.disable_sso_idp_authorization;
```
johnyrahul, line 171: **[silent-failure / medium]** Two things stack here:

1. `userSessionData.disable_sso_idp_authorization` is forwarded with no validation — if the backend response is missing the key, `undefined` flows downstream and (via the `!!` chain in `Users.jsx`) silently fails open.
2. The `catch (err)` block immediately below this assignment is a pre-existing `// TODO: Throw popup error message` that swallows hydration errors with no log/toast/error-id. The new authorization-sensitive flag now flows through that swallowed-error path, widening its blast radius.

Minimum: validate this is a boolean (else log via your error-id helper). Stretch: replace the TODO catch with a real handler — this is the silent-failure pattern explicitly called out in the project's error-handling guidance.

backend/account_v2/dto.py
johnyrahul: If even missused , it have proper check in backend. Hence ok with naming convention.

backend/account_v2/dto.py
johnyrahul: This can be ok. As this custom implemenatation , default to false is the best

backend/account_v2/dto.py
johnyrahul: Can be ignored.

backend/account_v2/views.py
johnyrahul: Value is hardcored in base.py. As OSS wont have any impact on this

backend/backend/settings/base.py
johnyrahul: can be ignored

frontend/src/components/settings/users/Users.jsx
johnyrahul: Should be ok, even in such conditions , We have strong backend validation.

frontend/src/components/settings/users/Users.jsx
johnyrahul: Test are not scoped for this PR

frontend/src/helpers/GetSessionData.js
johnyrahul: can be ignored. This waht expcetd for OSS users

backend/backend/settings/base.py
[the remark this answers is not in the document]
johnyrahul: Not required for OSS version , This is extended in the cloud/oprem side based on need

frontend/src/components/settings/users/Users.jsx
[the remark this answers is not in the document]
johnyrahul: This is enforced in the Auth0Service of the Auth0 cloud plugin of the unstract-cloud repo, Hence can be ignored. Comment is not valid

muhammad-ali-e (approved):
NIT: Would be btter as a plugin structure since it's not a OSS feature

frontend/src/hooks/useSessionValid.js
```diff
@@ -167,6 +167,8 @@ function useSessionValid() {
       }
       userAndOrgDetails["role"] = userSessionData.role;
       userAndOrgDetails["provider"] = userSessionData.provider;
+      userAndOrgDetails["disable_sso_idp_authorization"] =
```
jaseemjaskp, line 170: **[silent-failure / high]** This assignment runs *after* preceding `await` calls in the same try-block (`listFlags` at L160 and the `isPlatformAdmin` await at L166). If either throws, control jumps to `catch` (L172) — but the `finally` block (L187–189) still calls `setSessionDetails(getSessionData(userAndOrgDetails))` with `disable_sso_idp_authorization` never written.

Downstream this becomes `sessionDetails.disableSsoIdpAuthorization === undefined`, which `Users.jsx:100` collapses to `false` via `!!`. Net effect on a transient flag-service / platform-admin failure: SSO users get the **Delete** menu item back even when the deployment intends `DISABLE_SSO_IDP_AUTHORIZATION=True`. This is the privilege-exposing direction.

Fix is ordering, not type: hoist the assignment to immediately after the userSessionData fetch (before the conditional `await`s), or fail-closed in the catch path by setting it to `true` (i.e., hide Delete/Invite when we can't be sure).

```js
// move up, before the listFlags / isPlatformAdmin awaits
userAndOrgDetails['disable_sso_idp_authorization'] =
  userSessionData.disable_sso_idp_authorization;
```

frontend/src/components/settings/users/Users.jsx
jaseemjaskp, line 174: **[silent-failure / high]** Distinct from the previously-flagged `!!disableSsoIdpAuthorization` issue — this one is about `provider`. During the initial-render window before `useSessionValid` resolves, `sessionDetails?.provider` is `undefined`, so `!sessionDetails?.provider || isSsoLocalAuthz` is `true`, and the columns array becomes `[...baseColumns, actionColumn]`. At the same instant `isSsoLocalAuthz` is also `false`, so `actionItems = [editItem, deleteItem]` — i.e. the **Delete** menu item is rendered for SSO users for the duration of the first paint, before the session payload arrives.

This is fail-open behavior on a security-relevant gate. Fix it by gating on a definite loaded state (e.g. `sessionDetails?.isLoggedIn` or an explicit `provider !== undefined`) before deciding the columns shape, or treat unknown `provider` as SSO (fail-closed).

```js
const sessionLoaded = sessionDetails?.provider !== undefined;
const showActions = sessionLoaded && (!sessionDetails.provider || isSsoLocalAuthz);
const columns = showActions ? [...baseColumns, actionColumn] : baseColumns;
```

backend/account_v2/dto.py
jaseemjaskp, line 65: **[type-design / major]** Different angle from the earlier naming comment on this line: this field is a **deployment-wide Django setting** (`settings.DISABLE_SSO_IDP_AUTHORIZATION`) — i.e. constant for every session in a given deployment — but the PR models it as **per-session/per-user** state on `UserSessionInfo`. Consequences:

1. Every session payload now carries identical data (a small but pure waste).
2. The invariant *value is constant across all users in a deployment* is unenforced — `from_dict` accepts arbitrary bool per session, so a future caller can produce a `UserSessionInfo` with one user enabled and another disabled.
3. It conflates identity (who) with capability (what the deployment allows), which makes future refinement (per-org override, feature flagging) require a schema change rather than a swap-in.

Worth considering a sibling capabilities block on the response, e.g.

```python
# response shape
{
  "user": {...},
  "capabilities": {"local_user_authorization": settings.DISABLE_SSO_IDP_AUTHORIZATION},
}
```

Keeping `UserSessionInfo` focused on identity, with capabilities computed in `make_session_response`. Alternatively, model the gating matrix as an enum (`USER_MGMT_MODE = idp_managed | local | hybrid`) so illegal combinations like `local` without a `provider` become unrepresentable. None of this is blocking — flagging because the placement decision is hard to undo once consumers rely on the field name.

frontend/src/components/settings/users/Users.jsx
```diff
@@ -96,46 +96,51 @@ function Users() {
...
+
+  const editItem = {
+    key: "1",
+    label: (
+      <Space
+        direction="horizontal"
+        className="action-items"
+        onClick={() =>
+          navigate(`/${sessionDetails?.orgName}/users/edit`, {
+            state: selectedUserEmail,
```
jaseemjaskp, line 110: **[correctness / medium]** Stale-closure foot-gun introduced by hoisting `editItem` out of `actionColumn.render`. `editItem.label`'s `onClick` closes over `selectedUserEmail`, which is updated by the EllipsisOutlined click at L167 (`setSelectedUserEmail(record)`).

It works today only because React re-renders after the state set, before the menu pops, so `editItem` is rebuilt with the fresh `selectedUserEmail`. The moment a future contributor wraps `editItem`/`actionItems` in `useMemo` (an obvious follow-up given how this was just hoisted), the menu will navigate to the *previously-selected* email — a silent regression.

Safer pattern: build the items inside `actionColumn.render` so they capture the per-row `record` directly:

```js
render: (_, record) => {
  const items = [
    { key: '1', label: <Space onClick={() => navigate(`/${sessionDetails?.orgName}/users/edit`, { state: record })}>...</Space> },
    ...(isSsoLocalAuthz ? [] : [{ key: '2', label: <Space onClick={() => { setSelectedUserEmail(record); showModal(); }}>...</Space> }]),
  ];
  return <Dropdown menu={{ items }} ...>...</Dropdown>;
}
```

This also removes the need for the `selectedUserEmail` state piece entirely if `showModal` can take `record`.

frontend/src/components/settings/users/Users.jsx
jaseemjaskp, line 100: **[readability / low]** The double-`!!` is redundant here. `&&` short-circuits and returns the second operand, so only the *outer* coercion does work; the inner `!!`s are no-ops. Slightly clearer:

```js
const isSsoLocalAuthz = Boolean(
  sessionDetails?.provider && sessionDetails?.disableSsoIdpAuthorization,
);
```

Minor; only worth folding in if you're touching this region anyway.

frontend/src/components/settings/users/Users.jsx
jaseemjaskp, line 142: **[clarity / low]** Worth one inline comment locking in the product policy this line encodes — *delete must still go through the IdP, edit may be local*. The information density is high (one ternary expressing a security invariant) and the next pass that "simplifies" `isSsoLocalAuthz ? [editItem] : [editItem, deleteItem]` may invert it.

```js
// SSO local-authz mode: edit role locally, but delete still goes through the IdP.
const actionItems = isSsoLocalAuthz ? [editItem] : [editItem, deleteItem];
```

Not a duplicate of the prior tests/high comment on this line — orthogonal, and cheaper to land.

jaseemjaskp (commented):
Comprehensive review run via PR Review Toolkit (Comment Analyzer, PR Test Analyzer, Silent Failure Hunter, Type Design Analyzer, Code Reviewer, Code Simplifier). Most prior reviewer points are already addressed in-thread; the items below are additional findings that did not appear in the existing comments.

Priority summary:
- **High** — useSessionValid try/await ordering can leave the flag undefined → privilege exposure in the UI gate.
- **High** — Users.jsx columns predicate fails-open during the initial-render window before sessionDetails resolves (different mechanism from the previously-flagged `!!disableSsoIdpAuthorization` issue at line 100).
- **Major** — Type-design: the field is a deployment-wide setting but modeled as per-session state.
- **Medium** — Stale-closure foot-gun on `editItem`/`actionItems` if a future refactor adds memoization.
- **Low** — Redundant `!!` double-bangs and a missing intent-locking comment near the gating expression.

ritwik-g (approved):

frontend/src/components/settings/users/Users.jsx
johnyrahul: Addressed.

frontend/src/components/settings/users/Users.jsx
johnyrahul: can be ignored

ritwik-g:
@greptileai can you review latest changes. Only comment absolutely necessary changes.

PR #1944, opened 2026-05-05
[HOTFIX] Add IAM Role / Instance Profile auth mode to AWS Bedrock adapter

pk-zipstack:
## What

Adds an **Authentication Type** selector to the AWS Bedrock LLM and embedding adapter forms with two options:

- **Access Keys** (default): existing flow — provide `aws_access_key_id` and `aws_secret_access_key`
- **IAM Role / Instance Profile (on-prem AWS only)**: no fields; relies on `boto3`'s default credential chain

When the IAM Role mode is selected, the form-only `auth_type` field is stripped before LiteLLM validation, and any empty access keys are dropped from the validated kwargs so `boto3` falls through to its default credential chain (IRSA on EKS, task role on ECS, instance profile on EC2, AWS Profile, env vars).

## Why

Customers running Unstract on AWS infrastructure (specifically EKS with IRSA) cannot — and don't want to — paste long-lived AWS access keys into the adapter form. The pod's service account already has a federated identity that `boto3` can use to call Bedrock; the adapter form just needs to accept blank credentials so the ambient identity takes over.

The S3/MinIO connector at `unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py` already implements this idiom for S3:

```python
if key and secret:
    creds["key"] = key
    creds["secret"] = secret
```

This PR applies the same pattern to the Bedrock adapter, with the addition of an explicit `auth_type` selector to make the user's choice clear in the UI.

### `unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/bedrock.json`

Adds an `auth_type` enum with two options (`access_keys` default, `iam_role`) and a `dependencies/oneOf` block that:

- Renders `aws_access_key_id` + `aws_secret_access_key` (both required) when `auth_type == access_keys`
- Renders nothing extra when `auth_type == iam_role`

The selector's description explicitly notes the IRSA / instance-profile mode is for AWS-hosted Unstract deployments only.

### `unstract/sdk1/src/unstract/sdk1/adapters/embedding1/static/bedrock.json`

Same restructure. Drops `aws_access_key_id` and `aws_secret_access_key` from top-level `required`. Adds `aws_profile_name` for parity with the LLM form.

### `unstract/sdk1/src/unstract/sdk1/adapters/base1.py`

In both `AWSBedrockLLMParameters.validate()` and `AWSBedrockEmbeddingParameters.validate()`:

- Strip the form-only `auth_type` field from validation metadata before Pydantic dump
- Strip empty `aws_access_key_id` / `aws_secret_access_key` from the validated kwargs so `boto3`'s default chain takes over (mirrors the S3/MinIO connector idiom)

`AWSBedrockEmbeddingParameters` fields gain explicit `None` defaults so the access-key fields are truly optional in either mode (parity with `AWSBedrockLLMParameters` which already had this from earlier work).

No. Backward-compatible across every scenario:

| Scenario | Behavior |
|---|---|
| Existing adapter with access keys filled in (no `auth_type` field stored) | `auth_type` defaults to `access_keys`; keys forwarded to `boto3` unchanged |
| New adapter, **Access Keys** mode | Keys forwarded to `boto3` |
| New adapter, **IAM Role** mode | No keys submitted; empty values stripped → `boto3` default credential chain takes over |
| Existing adapter with `aws_profile_name` set | Forwarded unchanged (not part of `auth_type`) |
| Bedrock `enable_thinking` / `model_id` (Application Inference Profile) | Unchanged — these continue to work in both auth modes |

End-to-end verified against real AWS Bedrock during development:

- `aws_role_name=arn:...:role/BedrockInvokeRole` with bootstrap creds in env: `STS:AssumeRole` succeeds, `Bedrock InvokeModel` succeeds with temp creds (token usage returned).
- `auth_type=iam_role` with no keys + ambient creds in env: `boto3` default chain picks up the env vars; Bedrock call signs and routes correctly.

## Relevant Docs

- LiteLLM Bedrock provider parameters: https://docs.litellm.ai/docs/providers/bedrock
- AWS IRSA (IAM roles for EKS service accounts): https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
- Pattern reference: `unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py`

### DIFF
--- a/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
@@ -10,7 +10,6 @@ if TYPE_CHECKING:
     from typing import Any
 
 from pydantic import BaseModel, Field, model_validator
-
 from unstract.sdk1.adapters.constants import Common
 from unstract.sdk1.adapters.enums import AdapterTypes
 
--- a/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
@@ -525,11 +525,12 @@ class AWSBedrockLLMParameters(BaseChatCompletionParameters):
                 result_metadata["thinking"] = thinking_config
                 result_metadata["temperature"] = 1
 
-        # Create validation metadata excluding control fields
+        # Create validation metadata excluding control fields. `auth_type` is
+        # a UI-only selector that drives form rendering; LiteLLM never sees it.
         validation_metadata = {
             k: v
             for k, v in result_metadata.items()
-            if k not in ("enable_thinking", "budget_tokens", "thinking")
+            if k not in ("enable_thinking", "budget_tokens", "thinking", "auth_type")
         }
 
         validated = AWSBedrockLLMParameters(**validation_metadata).model_dump()
@@ -981,7 +982,12 @@ class AWSBedrockEmbeddingParameters(BaseEmbeddingParameters):
         ):
             adapter_metadata["aws_region_name"] = adapter_metadata["region_name"]
 
-        validated = AWSBedrockEmbeddingParameters(**adapter_metadata).model_dump()
+        # `auth_type` is a UI-only selector; strip before LiteLLM kwargs.
+        validation_metadata = {
+            k: v for k, v in adapter_metadata.items() if k != "auth_type"
+        }
+
+        validated = AWSBedrockEmbeddingParameters(**validation_metadata).model_dump()
 
         # Strip empty AWS access keys so boto3's default credential chain
         # takes over (instance profile / IRSA / AWS Profile / env vars).
--- a/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
@@ -10,6 +10,7 @@ if TYPE_CHECKING:
     from typing import Any
 
 from pydantic import BaseModel, Field, model_validator
+
 from unstract.sdk1.adapters.constants import Common
 from unstract.sdk1.adapters.enums import AdapterTypes
 
@@ -474,6 +475,71 @@ class VertexAILLMParameters(BaseChatCompletionParameters):
             return f"vertex_ai/{model}"
 
 
+# AWS Bedrock auth helpers: shared by LLM and Embedding param classes.
+# `auth_type` is a UI-only selector (Access Keys vs IAM Role / Instance
+# Profile) that drives form rendering. The backend translates the user's
+# choice into actual credential handling here so that both validate()
+# methods stay symmetric and a single bug fix applies to both paths.
+_BEDROCK_AWS_KEY_FIELDS: tuple[str, ...] = (
+    "aws_access_key_id",
+    "aws_secret_access_key",
+)
+_BEDROCK_VALID_AUTH_TYPES: frozenset[str | None] = frozenset(
+    {None, "access_keys", "iam_role"}
+)
+
+
+def _resolve_bedrock_aws_credentials(
+    adapter_metadata: dict[str, "Any"],
+    validated: dict[str, "Any"],
+) -> dict[str, "Any"]:
+    """Apply auth_type semantics to the validated LiteLLM kwargs.
+
+    Three cases:
+    - ``auth_type == "iam_role"``: drop access keys unconditionally so a
+      previously-saved adapter switched into IAM Role mode does not leak
+      stale long-lived credentials. boto3's default credential chain
+      (IRSA / instance profile / env vars / AWS Profile) takes over.
+    - ``auth_type == "access_keys"`` (explicit): require non-blank values.
+      A blank submission must surface as a clear error rather than
+      silently fall through to the default chain (which would hide the
+      mistake and authenticate with whatever ambient creds the host has).
+    - ``auth_type is None`` (legacy adapters created before this field
+      existed): lenient strip of empty/missing keys. Preserves backwards
+      compatibility for stored configurations.
+
+    Raises:
+        ValueError: on unknown ``auth_type`` (typo / non-UI client) or on
+            blank credentials when ``auth_type == "access_keys"``.
+    """
+    auth_type = adapter_metadata.get("auth_type")
+    if auth_type not in _BEDROCK_VALID_AUTH_TYPES:
+        raise ValueError(
+            f"Unknown auth_type {auth_type!r}; expected one of "
+            f"{sorted(t for t in _BEDROCK_VALID_AUTH_TYPES if t)!r} or absent."
+        )
+
+    if auth_type == "iam_role":
+        for key in _BEDROCK_AWS_KEY_FIELDS:
+            validated.pop(key, None)
+        return validated
+
+    if auth_type == "access_keys":
+        for key in _BEDROCK_AWS_KEY_FIELDS:
+            value = validated.get(key)
+            if not isinstance(value, str) or not value.strip():
+                raise ValueError(f"{key} is required when auth_type is 'access_keys'.")
+        return validated
+
+    # Legacy adapters with no auth_type: strip blanks silently to
+    # preserve the pre-PR behaviour where empty key fields fell through
+    # to boto3's default chain.
+    for key in _BEDROCK_AWS_KEY_FIELDS:
+        if not validated.get(key):
+            validated.pop(key, None)
+    return validated
+
+
 class AWSBedrockLLMParameters(BaseChatCompletionParameters):
     """See https://docs.litellm.ai/docs/providers/bedrock."""
 
@@ -539,14 +605,11 @@ class AWSBedrockLLMParameters(BaseChatCompletionParameters):
         if enable_thinking and "thinking" in result_metadata:
             validated["thinking"] = result_metadata["thinking"]
 
-        # Strip empty AWS access keys so boto3's default credential chain
-        # takes over (instance profile / IRSA / AWS Profile / env vars).
-        # Mirrors the pattern in the S3/MinIO connector.
-        for key in ("aws_access_key_id", "aws_secret_access_key"):
-            if not validated.get(key):
-                validated.pop(key, None)
-
-        return validated
+        # Apply Bedrock auth semantics: IAM Role mode drops keys, Access
+        # Keys mode requires non-blank values, legacy (no auth_type) is
+        # lenient. Reads auth_type from result_metadata since validation_
+        # metadata strips it before Pydantic.
+        return _resolve_bedrock_aws_credentials(result_metadata, validated)
 
     @staticmethod
     def validate_model(adapter_metadata: dict[str, "Any"]) -> str:
@@ -967,10 +1030,11 @@ class VertexAIEmbeddingParameters(BaseEmbeddingParameters):
 class AWSBedrockEmbeddingParameters(BaseEmbeddingParameters):
     """See https://docs.litellm.ai/docs/providers/bedrock."""
 
+    # Region is still mandatory — credentials are the only fields that
+    # may be absent (IAM Role / Instance Profile mode).
     aws_access_key_id: str | None = None
     aws_secret_access_key: str | None = None
-    aws_region_name: str | None = None
-    aws_profile_name: str | None = None  # For AWS SSO authentication
+    aws_region_name: str | None
 
     @staticmethod
     def validate(adapter_metadata: dict[str, "Any"]) -> dict[str, "Any"]:
@@ -989,14 +1053,9 @@ class AWSBedrockEmbeddingParameters(BaseEmbeddingParameters):
 
         validated = AWSBedrockEmbeddingParameters(**validation_metadata).model_dump()
 
-        # Strip empty AWS access keys so boto3's default credential chain
-        # takes over (instance profile / IRSA / AWS Profile / env vars).
-        # Mirrors the pattern in the S3/MinIO connector.
-        for key in ("aws_access_key_id", "aws_secret_access_key"):
-            if not validated.get(key):
-                validated.pop(key, None)
-
-        return validated
+        # Apply Bedrock auth semantics: IAM Role drops keys, Access Keys
+        # requires non-blank values, legacy (no auth_type) is lenient.
+        return _resolve_bedrock_aws_credentials(adapter_metadata, validated)
 
     @staticmethod
     def validate_model(adapter_metadata: dict[str, "Any"]) -> str:
--- a/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
@@ -538,6 +538,13 @@ class AWSBedrockLLMParameters(BaseChatCompletionParameters):
         if enable_thinking and "thinking" in result_metadata:
             validated["thinking"] = result_metadata["thinking"]
 
+        # Strip empty AWS access keys so boto3's default credential chain
+        # takes over (instance profile / IRSA / AWS Profile / env vars).
+        # Mirrors the pattern in the S3/MinIO connector.
+        for key in ("aws_access_key_id", "aws_secret_access_key"):
+            if not validated.get(key):
+                validated.pop(key, None)
+
         return validated
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -973,7 +982,21 @@ def validate(adapter_metadata: dict[str, "Any"]) -> dict[str, "Any"]:
...
+        }
+
+        validated = AWSBedrockEmbeddingParameters(**validation_metadata).model_dump()
+
+        # Strip empty AWS access keys so boto3's default credential chain
+        # takes over (instance profile / IRSA / AWS Profile / env vars).
+        # Mirrors the pattern in the S3/MinIO connector.
+        for key in ("aws_access_key_id", "aws_secret_access_key"):
+            if not validated.get(key):
+                validated.pop(key, None)
```
jaseemjaskp: **P0 — Embedding-side parity of the IAM-role stale-key bug.** @greptile-apps[bot] flagged the same class of bug on the LLM `validate` at line 548. The exact same `not validated.get(key)` pop loop here (`AWSBedrockEmbeddingParameters.validate`) has the identical hazard: an existing embedding adapter with saved access keys, switched to `auth_type=iam_role` in the UI, will keep its stale keys (they're still truthy → `pop` is skipped) and boto3 silently uses them instead of falling through to IRSA / instance profile.

**Suggested fix (apply to both LLM at 545 and Embedding at 995, ideally in a shared helper — see separate finding):**
```python
if validated.get("auth_type") == "iam_role" or not validated.get("aws_access_key_id"):
    for key in ("aws_access_key_id", "aws_secret_access_key"):
        validated.pop(key, None)
```
Or branch explicitly on `auth_type` before the strip rather than on emptiness — that makes the IAM-role contract structural rather than coincidental.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -525,11 +525,12 @@ def validate(adapter_metadata: dict[str, "Any"]) -> dict[str, "Any"]:
...
                 result_metadata["temperature"] = 1
 
-        # Create validation metadata excluding control fields
+        # Create validation metadata excluding control fields. `auth_type` is
+        # a UI-only selector that drives form rendering; LiteLLM never sees it.
         validation_metadata = {
             k: v
             for k, v in result_metadata.items()
-            if k not in ("enable_thinking", "budget_tokens", "thinking")
+            if k not in ("enable_thinking", "budget_tokens", "thinking", "auth_type")
```
jaseemjaskp, line 598: **P0 — `auth_type` is silently swallowed without enum validation.** This dict comprehension drops `auth_type` from `validation_metadata` *before* Pydantic runs, so a typo (`access_key`), a removed enum value, or a malformed payload from a non-UI client (CLI, migration script, API) is accepted without error and falls through to whatever credentials happen to be in `validated`. The JSON-schema `oneOf` only protects the React form; backend acceptance is wide open.

**Suggested fix:**
```python
allowed = {None, "access_keys", "iam_role"}
if result_metadata.get("auth_type") not in allowed:
    raise ValueError(f"Unknown auth_type: {result_metadata.get('auth_type')!r}")
```
Apply before the strip. Same issue exists in the embedding `validate` at line 987.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -973,7 +982,21 @@ def validate(adapter_metadata: dict[str, "Any"]) -> dict[str, "Any"]:
         ):
             adapter_metadata["aws_region_name"] = adapter_metadata["region_name"]
 
-        return AWSBedrockEmbeddingParameters(**adapter_metadata).model_dump()
+        # `auth_type` is a UI-only selector; strip before LiteLLM kwargs.
+        validation_metadata = {
+            k: v for k, v in adapter_metadata.items() if k != "auth_type"
```
jaseemjaskp, line 1050: **P0 — Same silent-acceptance issue as flagged on line 533** for `AWSBedrockLLMParameters.validate`. Unknown `auth_type` values are stripped here too without any enum check. Add the same allowlist guard before the strip — ideally via a single shared helper used by both validate methods.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -538,6 +539,13 @@ def validate(adapter_metadata: dict[str, "Any"]) -> dict[str, "Any"]:
         if enable_thinking and "thinking" in result_metadata:
             validated["thinking"] = result_metadata["thinking"]
 
+        # Strip empty AWS access keys so boto3's default credential chain
+        # takes over (instance profile / IRSA / AWS Profile / env vars).
+        # Mirrors the pattern in the S3/MinIO connector.
+        for key in ("aws_access_key_id", "aws_secret_access_key"):
+            if not validated.get(key):
```
jaseemjaskp: **P0 — No cross-field validation: `auth_type` and the actual key presence can disagree silently.**

Two concrete failure modes:
1. User selects **Access Keys** mode but submits blank (browser autofill bug, paste error, partial fill) → `not validated.get(key)` is true → both keys are popped → boto3 silently falls through to the default credential chain instead of surfacing "AWS Access Key ID is required". The user thinks they configured access-key auth and is unknowingly relying on ambient creds.
2. `not validated.get(key)` is also false for whitespace-only strings (`"  "`). Those flow through to boto3 as a real key and produce an opaque AWS auth error rather than a clean form-validation message.

**Suggested fix:**
```python
auth_mode = result_metadata.get("auth_type", "access_keys")
aws_keys = ("aws_access_key_id", "aws_secret_access_key")
if auth_mode == "iam_role":
    for key in aws_keys:
        validated.pop(key, None)
else:  # access_keys
    for key in aws_keys:
        v = validated.get(key)
        if not v or not v.strip():
            raise ValueError(f"{key} is required when auth_type is access_keys")
```

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -538,6 +539,13 @@ def validate(adapter_metadata: dict[str, "Any"]) -> dict[str, "Any"]:
         if enable_thinking and "thinking" in result_metadata:
             validated["thinking"] = result_metadata["thinking"]
 
+        # Strip empty AWS access keys so boto3's default credential chain
+        # takes over (instance profile / IRSA / AWS Profile / env vars).
+        # Mirrors the pattern in the S3/MinIO connector.
+        for key in ("aws_access_key_id", "aws_secret_access_key"):
```
jaseemjaskp: **P1 — Duplicated logic with the embedding adapter (mirror block at lines 995–997).** The same 3-line strip-empty loop, the same comment, and (per the other findings) the same enum-validation and cross-field-check needs apply to both. With `aws_profile_name` being a sibling concern, the list of fields-to-strip will keep growing in two places and inevitably drift.

**Suggested fix:** extract a module-private helper near the other Bedrock helpers and call it from both `validate` methods:
```python
_NON_LITELLM_FIELDS = ("enable_thinking", "budget_tokens", "thinking", "auth_type")

def _strip_empty_aws_credentials(params: dict[str, Any]) -> dict[str, Any]:
    """Drop empty AWS keys so boto3's default credential chain takes over."""
    for key in ("aws_access_key_id", "aws_secret_access_key"):
        if not params.get(key):
            params.pop(key, None)
    return params
```

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -538,6 +539,13 @@ def validate(adapter_metadata: dict[str, "Any"]) -> dict[str, "Any"]:
...
+            if not validated.get(key):
+                validated.pop(key, None)
+
         return validated
```
jaseemjaskp: **P1 — No test coverage for any of the new behavior.** A grep of `unstract/sdk1/tests/` finds zero references to `AWSBedrockLLMParameters` or `AWSBedrockEmbeddingParameters`. Both new branches in this PR — `auth_type` stripping and empty-key popping — are entirely untested.

**Highest-value cases to add** (in a new `unstract/sdk1/tests/test_bedrock_adapter.py`, following `test_gemini_adapter.py` as a template):
1. Legacy adapter dict with no `auth_type` key → keys flow through unchanged (backwards compat — most important).
2. `auth_type="access_keys"` with valid keys → `auth_type` stripped, keys preserved.
3. `auth_type="iam_role"` with no keys → both keys absent from output (the entire purpose of this PR).
4. `auth_type="access_keys"` with empty-string keys → pins current behavior so any future fix is a deliberate contract change.
5. Other params (`model_id`, `aws_profile_name`, `aws_region_name`, `enable_thinking`) preserved through the strip.

Mirror the same matrix for `AWSBedrockEmbeddingParameters`.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -959,9 +967,10 @@ def validate_model(adapter_metadata: dict[str, "Any"]) -> str:
 class AWSBedrockEmbeddingParameters(BaseEmbeddingParameters):
     """See https://docs.litellm.ai/docs/providers/bedrock."""
 
-    aws_access_key_id: str | None
-    aws_secret_access_key: str | None
-    aws_region_name: str | None
+    aws_access_key_id: str | None = None
+    aws_secret_access_key: str | None = None
+    aws_region_name: str | None = None
```
jaseemjaskp: **P1 — Field default change downgrades `aws_region_name` from required to silently optional.** Previously `aws_region_name: str | None` (no default) raised `ValidationError` if a caller forgot to pass region. Now `aws_region_name: str | None = None` accepts a missing region and defers the failure to LiteLLM/boto3, where it surfaces as an opaque downstream error. The relaxation is correct for the *credential* fields (intentional fallthrough) but not for region — region is required in either auth mode.

**Suggested fix:** keep `aws_region_name: str | None` (no default), or apply only to the credential fields. Same applies to `AWSBedrockLLMParameters` if it was changed alongside.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -525,11 +525,12 @@ def validate(adapter_metadata: dict[str, "Any"]) -> dict[str, "Any"]:
                 result_metadata["thinking"] = thinking_config
                 result_metadata["temperature"] = 1
 
-        # Create validation metadata excluding control fields
+        # Create validation metadata excluding control fields. `auth_type` is
```
jaseemjaskp, line 593: **P2 — Misleading comment scope.** The comment singles out `auth_type` ("a UI-only selector...") but the filter on the next line strips four fields: `enable_thinking`, `budget_tokens`, `thinking`, `auth_type`. A future reader sees the comment and assumes the other three are unrelated, when they are the original "control fields" the comment originally referred to.

**Suggested replacement:**
```python
# Strip control fields before model validation: thinking-mode toggles
# are recombined below into a single 'thinking' dict; 'auth_type' is
# a UI-only selector that drives form rendering and is never sent to
# LiteLLM.
```

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -538,6 +539,13 @@ def validate(adapter_metadata: dict[str, "Any"]) -> dict[str, "Any"]:
         if enable_thinking and "thinking" in result_metadata:
             validated["thinking"] = result_metadata["thinking"]
 
+        # Strip empty AWS access keys so boto3's default credential chain
+        # takes over (instance profile / IRSA / AWS Profile / env vars).
```
jaseemjaskp: **P2 — "Mirrors the pattern in the S3/MinIO connector" is a rot magnet.** No file path, symbol, or anchor — if `unstract/connectors/src/unstract/connectors/filesystems/minio/minio.py` is renamed, refactored, or the pattern diverges, this reference silently goes stale and a maintainer cannot verify it.

**Suggested fix:** either drop the sentence (the surrounding comment already explains the *why*), or pin the reference: `# Same idiom as connectors/.../minio/minio.py (key/secret stripped if blank).` The same comment block exists at line 992 — apply both.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -973,7 +982,21 @@ def validate(adapter_metadata: dict[str, "Any"]) -> dict[str, "Any"]:
         ):
             adapter_metadata["aws_region_name"] = adapter_metadata["region_name"]
```
jaseemjaskp, line 1046: **P2 — `validate` mutates caller's dict.** `adapter_metadata["aws_region_name"] = adapter_metadata["region_name"]` writes back into the caller's dict. Not a type-safety issue per se, but it leaks state and undermines the "validate is a pure transform" contract the static method implies — a caller passing the same dict twice gets different results on the second call.

**Suggested fix:** copy first, then mutate the copy, e.g. `adapter_metadata = dict(adapter_metadata)` at the top of the method.

unstract/sdk1/src/unstract/sdk1/adapters/embedding1/static/bedrock.json
```diff
@@ -21,23 +19,16 @@
...
-        "title": "AWS Secret Access Key",
-        "description": "Provide your AWS Secret Access Key",
-        "format": "password"
-      },
       "region_name": {
         "type": "string",
         "title": "AWS Region name",
         "description": "Provide the AWS Region name where the service is running. Eg. us-east-1"
       },
+      "aws_profile_name": {
```
jaseemjaskp: **P0 — `aws_profile_name` is exposed outside the `auth_type` `oneOf` branch and contradicts the selector's semantics.** This field is added at the top level of `properties`, so it renders as a free-text input regardless of `auth_type`. Three distinct problems:

1. The `auth_type` description (lines 60–61) tells users **IAM Role / Instance Profile** "requires no further input" — but `aws_profile_name` is always visible alongside, contradicting that promise.
2. There is no `aws_profile` enum value in `auth_type`, so users have no clear path to use SSO mode; they must guess that filling this field overrides the keys.
3. `validate()` at `base1.py:995` only strip-empties the access keys; an existing adapter with a stale `aws_profile_name` carries that value forward when the user switches modes (same bug class as the greptile P1, but for profile name).

**Suggested fix:** either move `aws_profile_name` into the `oneOf` as a third `auth_type` branch (e.g. `aws_profile`), or remove it from this PR — it is orthogonal to the IAM-role fix and currently ships as a sidecar input that can silently override the selected mode. Same applies to `unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/bedrock.json` (field at LLM:27 + Pydantic field at `base1.py:483`).

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/bedrock.json
```diff
@@ -75,6 +63,60 @@
...
+      "enum": [
+        "access_keys",
+        "iam_role"
+      ],
+      "enumNames": [
+        "Access Keys",
+        "IAM Role / Instance Profile (on-prem AWS only)"
+      ],
+      "default": "access_keys",
+      "description": "Choose **Access Keys** for any deployment (provide your AWS Access Key ID and Secret). Choose **IAM Role / Instance Profile** only when Unstract is hosted on AWS infrastructure that already has ambient credentials — for example EKS pods with IRSA, ECS tasks with a task role, or EC2 instances with an instance profile. The on-prem option requires no further input; boto3 picks up the host's identity automatically."
```
jaseemjaskp, line 79: **P2 — Description claims IAM Role mode "requires no further input" but `aws_profile_name` (defined at line 27) is always visible.** Cross-reference the embedding-side finding on the same field. Either tighten the description ("…requires no AWS access keys; aws_profile_name is optional and applies in either mode") or — preferred — fold `aws_profile_name` into a third `auth_type` branch so the selector and the visible inputs stay consistent.

jaseemjaskp (commented):
Comprehensive PR review via PR Review Toolkit (Comment Analyzer, PR Test Analyzer, Silent Failure Hunter, Type Design Analyzer, Code Reviewer, Code Simplifier). The greptile P1 on `base1.py:548` is acknowledged and not duplicated; the findings below extend it (embedding-side parity) and surface separate concerns.

**Priority distribution:** 4 P0 (silent-failure / correctness), 3 P1 (structural / coverage), 4 P2 (style / docs).

jaseemjaskp (approved):

PR #1949, opened 2026-05-07
UN-3454 [FEAT] Add CSV/JSON export to workflow execution logs modal

muhammad-ali-e:
## What

- Adds an **Export ▾** dropdown to the workflow execution logs modal (the existing eye-icon view) with two options: *Download as JSON* and *Download as CSV*.
- Backend: new `export` action on `WorkflowExecutionLogViewSet` at `GET /execution/<pk>/logs/export/?file_format=csv|json` (also accepts `file_execution_id` and `log_level` — same filters as the list endpoint).
- Frontend: new dropdown in the existing `LogModal` header. Disabled when there are no logs to export, and while an export is in flight (prevents double-submit).

## Why

- Closes the gap that there is no way to download or share logs from the modal today. The 10-line pagination makes copy-paste of a 500-line failed file painful.
- Real user pain points addressed: debugging a failed file (paste into a ticket / Slack), post-mortem on a workflow run, ad-hoc audit needs.
- Jira: [UN-3454](https://zipstack.atlassian.net/browse/UN-3454)

## How

- **Scope = current modal scope.** Reuses `ExecutionLogFilter` so the active filters in the modal (file scope, log level) apply identically to the export. No new buttons sprinkled around the table.
- **Buffered `HttpResponse`**, not streaming. Real-world export size observed ~17 KB; streaming would lose `Content-Length` (browser progress %) and risk corrupted partial files on mid-stream errors. Industry standard for capped sync exports (Datadog/Sentry/CloudWatch/Kibana/Grafana) is buffered.
- **50,000 row hard cap.** Above the cap returns HTTP 413 with a JSON body asking the user to narrow their filter. Frontend decodes the blob body to surface the message instead of a generic error.
- **Multi-tenant safety** is unchanged — `OrgAwareManager` and the existing `OrganizationFilterBackend` enforce org scoping for the new action exactly like they do for `list`.
- **One footgun handled:** the query param is `file_format`, not `format`. DRF reserves `?format=` for content negotiation and 404s when the value doesn't match a registered renderer. There's a comment on the backend variable to prevent re-introduction.

### Out of scope (deferred)

- Workflow-level log export (across all executions of a workflow).
- Deployment / pipeline-level log export (across all runs).

These need date-range UI, async Celery job, signed URL delivery, retention policy, and rate-limiting. Will be tracked separately if/when users explicitly ask. The 50k row cap exists precisely because async deployment-wide export is a different feature with different ergonomics.

**No regressions expected.** Reasoning:

- The existing `list` endpoint (`GET /execution/<pk>/logs/`) is unchanged. `get_queryset()` was not modified; the `.order_by("event_time").only(...)` and `.count()` calls are scoped inside `export()` only.
- New URL path `/execution/<pk>/logs/export/` is added in both the active route file (`backend/workflow_manager/execution/urls.py`) and the legacy one (`backend/workflow_manager/workflow_v2/urls/workflow.py`) so behavior is consistent regardless of which prefix routes the request.
- Frontend changes are additive within `LogModal.jsx`. The recent UN-3431 `CustomMarkdown` integration in the same file is preserved (verified after rebase).

## Database Migrations

- None.

## Env Config

- None.

## Notes on Testing

Smoke test plan:

- Open Logs page → workflow execution row → eye icon → modal opens. Click **Export → Download as CSV**. File downloads with all logs for the execution.
- Open Detailed Logs → file row eye icon → modal scoped to that file. Export downloads only that file's logs.
- Apply log level filter (e.g. ERROR) before clicking Export → exported file respects the filter.
- Empty execution → Export button is disabled with tooltip "No logs to export".
- Click Export twice quickly → second click is blocked while the first is in flight (loading spinner on button).
- Force the 413 path by lowering `MAX_SYNC_EXPORT_ROWS` locally → confirm the alert renders the "narrow your filter" message instead of a generic error.

Code review (separate pass) called out:
- Type guard for non-dict `data` field in CSV path (defensive against historical/manually-inserted log records that aren't dicts).
- `exporting` state guard to prevent concurrent submits.
Both applied.

## Related Issues or PRs

- Jira: [UN-3454](https://zipstack.atlassian.net/browse/UN-3454)
- Touches `LogModal.jsx` which was last modified by [UN-3431](https://zipstack.atlassian.net/browse/UN-3431) (#1927). No conflict — UN-3431's `CustomMarkdown` integration is preserved.

## Screenshots
logs_download

_To be added on the PR after manual verification — shows the new "Export" button in the modal header._

## Checklist

- [x] PR title format: `UN-3454 [FIX] ...`
- [x] Description filled with what / why / how / breaking-changes / migrations / env-config sections
- [x] Pre-commit passed locally on all 5 changed files
- [x] Self-reviewed; addressed code-reviewer findings
- [ ] Manual smoke test in running platform (test plan above)
- [x] No existing tests broken (changes are additive; `list` endpoint untouched)
- [x] No new dependencies added

[UN-3454]: https://zipstack.atlassian.net/browse/UN-3454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[UN-3454]: https://zipstack.atlassian.net/browse/UN-3454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/workflow_manager/workflow_v2/execution_log_view.py
+++ b/backend/workflow_manager/workflow_v2/execution_log_view.py
@@ -1,10 +1,16 @@
+import csv
+import io
+import json
 import logging
 
 from django.db.models import Q
 from django.db.models.query import QuerySet
+from django.http import HttpResponse
+from django.utils import timezone
 from permissions.permission import IsOwner
-from rest_framework import viewsets
+from rest_framework import status, viewsets
 from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
 from rest_framework.versioning import URLPathVersioning
 from utils.pagination import CustomPagination
 
@@ -13,6 +19,11 @@ from workflow_manager.workflow_v2.models.execution_log import ExecutionLog
 from workflow_manager.workflow_v2.serializers import WorkflowExecutionLogSerializer
 
 logger = logging.getLogger(__name__)
+
+# Cap on synchronous export size. Above this, callers should narrow the
+# filter (file_execution_id, log_level) — async deployment-wide export
+# is intentionally a separate, future feature.
+MAX_SYNC_EXPORT_ROWS = 50_000
 
 
 class WorkflowExecutionLogViewSet(viewsets.ModelViewSet):
@@ -32,3 +43,86 @@ class WorkflowExecutionLogViewSet(viewsets.ModelViewSet):
         return ExecutionLog.objects.filter(
             Q(wf_execution_id=execution_id) | Q(execution_id=execution_id)
         )
+
+    def export(self, request, *args, **kwargs):
+        """Export logs for a single workflow execution as CSV or JSON.
+
+        Honors the same filters as the list endpoint (file_execution_id,
+        log_level). Returns 413 if the result set exceeds the sync cap so
+        the client can prompt the user to narrow their filter.
+        """
+        # NOTE: do not name this query param `format` — DRF reserves it for
+        # content negotiation and will 404 if no renderer matches the value.
+        export_format = request.query_params.get("file_format", "json").lower()
+        if export_format not in ("json", "csv"):
+            return Response(
+                {"error": "file_format must be one of: json, csv"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        queryset = (
+            self.filter_queryset(self.get_queryset())
+            .order_by("event_time")
+            .only("id", "event_time", "data", "file_execution_id")
+        )
+
+        row_count = queryset.count()
+        if row_count > MAX_SYNC_EXPORT_ROWS:
+            return Response(
+                {
+                    "error": (
+                        f"Too many logs to export ({row_count} rows). "
+                        f"Limit is {MAX_SYNC_EXPORT_ROWS}. "
+                        "Narrow the filter (e.g. by file or log level) and retry."
+                    ),
+                    "row_count": row_count,
+                    "limit": MAX_SYNC_EXPORT_ROWS,
+                },
+                status=status.HTTP_413_REQUEST_ENTITY_TOO_LARGE,
+            )
+
+        if export_format == "csv":
+            body = self._build_csv(queryset)
+            content_type = "text/csv"
+        else:
+            body = self._build_json(queryset)
+            content_type = "application/json"
+
+        execution_id = self.kwargs.get("pk")
+        timestamp = timezone.now().strftime("%Y%m%d_%H%M%S")
+        filename = f"execution_logs_{execution_id}_{timestamp}.{export_format}"
+
+        response = HttpResponse(body, content_type=content_type)
+        response["Content-Disposition"] = f'attachment; filename="{filename}"'
+        return response
+
+    def _build_csv(self, queryset: QuerySet) -> str:
+        output = io.StringIO()
+        writer = csv.writer(output)
+        writer.writerow(["event_time", "level", "stage", "log", "file_execution_id"])
+        for log in queryset:
+            data = log.data if isinstance(log.data, dict) else {}
+            writer.writerow(
+                [
+                    log.event_time.isoformat() if log.event_time else "",
+                    data.get("level", ""),
+                    data.get("stage", ""),
+                    data.get("log", ""),
+                    str(log.file_execution_id) if log.file_execution_id else "",
+                ]
+            )
+        return output.getvalue()
+
+    def _build_json(self, queryset: QuerySet) -> str:
+        entries = [
+            {
+                "id": str(log.id),
+                "event_time": log.event_time.isoformat() if log.event_time else None,
+                "file_execution_id": (
+                    str(log.file_execution_id) if log.file_execution_id else None
+                ),
+                "data": log.data,
+            }
+            for log in queryset
+        ]
+        return json.dumps(entries)
--- a/backend/workflow_manager/workflow_v2/execution_log_view.py
+++ b/backend/workflow_manager/workflow_v2/execution_log_view.py
@@ -66,26 +66,26 @@ class WorkflowExecutionLogViewSet(viewsets.ModelViewSet):
             .only("id", "event_time", "data", "file_execution_id")
         )
 
-        row_count = queryset.count()
-        if row_count > MAX_SYNC_EXPORT_ROWS:
+        # Single materialization with cap+1 sentinel — avoids a separate COUNT
+        # query and a second full scan during build.
+        rows = list(queryset[: MAX_SYNC_EXPORT_ROWS + 1])
+        if len(rows) > MAX_SYNC_EXPORT_ROWS:
             return Response(
                 {
                     "error": (
-                        f"Too many logs to export ({row_count} rows). "
-                        f"Limit is {MAX_SYNC_EXPORT_ROWS}. "
+                        f"Too many logs to export (>{MAX_SYNC_EXPORT_ROWS} rows). "
                         "Narrow the filter (e.g. by file or log level) and retry."
                     ),
-                    "row_count": row_count,
                     "limit": MAX_SYNC_EXPORT_ROWS,
                 },
                 status=status.HTTP_413_REQUEST_ENTITY_TOO_LARGE,
             )
 
         if export_format == "csv":
-            body = self._build_csv(queryset)
-            content_type = "text/csv"
+            body = self._build_csv(rows)
+            content_type = "text/csv; charset=utf-8"
         else:
-            body = self._build_json(queryset)
+            body = self._build_json(rows)
             content_type = "application/json"
 
         execution_id = self.kwargs.get("pk")
@@ -96,33 +96,62 @@ class WorkflowExecutionLogViewSet(viewsets.ModelViewSet):
         response["Content-Disposition"] = f'attachment; filename="{filename}"'
         return response
 
-    def _build_csv(self, queryset: QuerySet) -> str:
+    def _normalize(self, log: ExecutionLog) -> dict:
+        """Single source of truth for per-row null-handling and dict-guard.
+
+        Logs warning when `data` is non-dict so silent blank rows in CSV
+        still leave a diagnostic trail in operator logs.
+        """
+        if isinstance(log.data, dict):
+            data = log.data
+        else:
+            if log.data is not None:
+                logger.warning(
+                    "ExecutionLog %s has non-dict data of type %s; "
+                    "emitting blanks in CSV export",
+                    log.id,
+                    type(log.data).__name__,
+                )
+            data = {}
+        return {
+            "id": str(log.id),
+            "event_time": log.event_time.isoformat() if log.event_time else None,
+            "file_execution_id": (
+                str(log.file_execution_id) if log.file_execution_id else None
+            ),
+            "level": data.get("level", ""),
+            "stage": data.get("stage", ""),
+            "log_message": data.get("log", ""),
+            "raw_data": log.data,
+        }
+
+    def _build_csv(self, rows: list[ExecutionLog]) -> str:
         output = io.StringIO()
         writer = csv.writer(output)
         writer.writerow(["event_time", "level", "stage", "log", "file_execution_id"])
-        for log in queryset:
-            data = log.data if isinstance(log.data, dict) else {}
+        for log in rows:
+            n = self._normalize(log)
             writer.writerow(
                 [
-                    log.event_time.isoformat() if log.event_time else "",
-                    data.get("level", ""),
-                    data.get("stage", ""),
-                    data.get("log", ""),
-                    str(log.file_execution_id) if log.file_execution_id else "",
+                    n["event_time"] or "",
+                    n["level"],
+                    n["stage"],
+                    n["log_message"],
+                    n["file_execution_id"] or "",
                 ]
             )
         return output.getvalue()
 
-    def _build_json(self, queryset: QuerySet) -> str:
+    def _build_json(self, rows: list[ExecutionLog]) -> str:
+        # JSON path passes raw_data through verbatim (faithful to the DB)
+        # rather than projecting it to extracted level/stage/log fields.
         entries = [
             {
-                "id": str(log.id),
-                "event_time": log.event_time.isoformat() if log.event_time else None,
-                "file_execution_id": (
-                    str(log.file_execution_id) if log.file_execution_id else None
-                ),
-                "data": log.data,
+                "id": n["id"],
+                "event_time": n["event_time"],
+                "file_execution_id": n["file_execution_id"],
+                "data": n["raw_data"],
             }
-            for log in queryset
+            for n in (self._normalize(log) for log in rows)
         ]
         return json.dumps(entries)
--- a/frontend/src/components/logging/log-modal/LogModal.jsx
+++ b/frontend/src/components/logging/log-modal/LogModal.jsx
@@ -1,5 +1,5 @@
-import { CopyOutlined } from "@ant-design/icons";
-import { Button, Modal, Table, Tooltip } from "antd";
+import { CopyOutlined, DownloadOutlined } from "@ant-design/icons";
+import { Button, Dropdown, Modal, Table, Tooltip } from "antd";
 import PropTypes from "prop-types";
 import { useEffect, useState } from "react";
 
@@ -32,6 +32,7 @@ function LogModal({
 
   const [executionLogs, setExecutionLogs] = useState([]);
   const [loading, setLoading] = useState(false);
+  const [exporting, setExporting] = useState(false);
   const [selectedLogLevel, setSelectedLogLevel] = useState(null);
   const [ordering, setOrdering] = useState(null);
   const [pagination, setPagination] = useState({
@@ -83,6 +84,71 @@ function LogModal({
     setExecutionLogs([]); // Clear logs
     setPagination({ current: 1, pageSize: 10, total: 0 }); // Reset pagination
   };
+
+  const handleExport = async (fileFormat) => {
+    if (exporting) {
+      return;
+    }
+    setExporting(true);
+    try {
+      const url = getUrl(`/execution/${executionId}/logs/export/`);
+      const response = await axiosPrivate.get(url, {
+        params: {
+          // `file_format` (not `format`) — `format` is reserved by DRF for
+          // content negotiation and 404s when set to "csv".
+          file_format: fileFormat,
+          file_execution_id: fileId || "null",
+          log_level: selectedLogLevel,
+        },
+        responseType: "blob",
+      });
+
+      const filename =
+        response.headers?.["content-disposition"]?.match(
+          /filename="?([^"]+)"?/,
+        )?.[1] || `execution_logs_${fileId || executionId}.${fileFormat}`;
+      const blobUrl = window.URL.createObjectURL(new Blob([response.data]));
+      const link = document.createElement("a");
+      link.href = blobUrl;
+      link.download = filename;
+      document.body.appendChild(link);
+      link.click();
+      link.remove();
+      window.URL.revokeObjectURL(blobUrl);
+    } catch (err) {
+      // The 413 (too many rows) response is a JSON blob — surface its
+      // message to the user instead of a generic error.
+      if (err?.response?.status === 413 && err.response.data) {
+        try {
+          const text = await err.response.data.text();
+          const parsed = JSON.parse(text);
+          setAlertDetails({
+            type: "error",
+            content: parsed?.error || "Export too large",
+          });
+          return;
+        } catch {
+          // fall through to default handler
+        }
+      }
+      setAlertDetails(handleException(err));
+    } finally {
+      setExporting(false);
+    }
+  };
+
+  const exportMenuItems = [
+    {
+      key: "json",
+      label: "Download as JSON",
+      onClick: () => handleExport("json"),
+    },
+    {
+      key: "csv",
+      label: "Download as CSV",
+      onClick: () => handleExport("csv"),
+    },
+  ];
 
   const logDetailsColumns = [
     {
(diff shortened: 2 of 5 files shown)


backend/workflow_manager/workflow_v2/execution_log_view.py
```diff
@@ -32,3 +43,86 @@ def get_queryset(self) -> QuerySet:
...
+                        "Narrow the filter (e.g. by file or log level) and retry."
+                    ),
+                    "row_count": row_count,
+                    "limit": MAX_SYNC_EXPORT_ROWS,
+                },
+                status=status.HTTP_413_REQUEST_ENTITY_TOO_LARGE,
+            )
+
+        if export_format == "csv":
+            body = self._build_csv(queryset)
```
muhammad-ali-e: **P1 — Memory buffering risk under concurrency.** `_build_csv` / `_build_json` materialize the full result into a single in-memory string before any byte is sent. With the cap at 50k rows and `data` being a JSON blob containing log strings (often multi-KB Markdown), this can reach hundreds of MB resident per request, and Django then copies the body again into the WSGI response. Concurrent exports → worker OOM.

Suggest `StreamingHttpResponse` with `queryset.iterator(chunk_size=2000)`:
- CSV: yield header row, then yield each `writer.writerow(...)` chunk via a `StringIO` flush per chunk.
- JSON: hand-emit `[`, comma-join encoded entries, `]` — or switch the JSON path to `application/x-ndjson` (one entry per line) which is trivial to stream and parse.

This pairs naturally with the `count()` change suggested below.

backend/workflow_manager/workflow_v2/execution_log_view.py
```diff
@@ -32,3 +43,86 @@ def get_queryset(self) -> QuerySet:
...
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        queryset = (
+            self.filter_queryset(self.get_queryset())
+            .order_by("event_time")
+            .only("id", "event_time", "data", "file_execution_id")
+        )
+
+        row_count = queryset.count()
```
muhammad-ali-e: **P2 — `count()` adds a second full DB scan.** `queryset.count()` then a full fetch is two passes over (potentially) 50k rows. Prefer a single materialization for the gate:

```python
rows = list(queryset[:MAX_SYNC_EXPORT_ROWS + 1])
if len(rows) > MAX_SYNC_EXPORT_ROWS:
    return Response({...}, status=413)
# stream `rows`
```

If you adopt the `StreamingHttpResponse` approach from the comment on line 85, push the cap into the generator: yield the header, then on row 50,001 raise / abort the iteration. Either way avoids the double scan.

backend/workflow_manager/workflow_v2/execution_log_view.py
```diff
@@ -32,3 +43,86 @@ def get_queryset(self) -> QuerySet:
         return ExecutionLog.objects.filter(
             Q(wf_execution_id=execution_id) | Q(execution_id=execution_id)
         )
+
+    def export(self, request, *args, **kwargs):
```
muhammad-ali-e, line 47: **P1 — No tests added; high-value gaps.** This new endpoint has zero test coverage. Most important to lock in:

1. **Org isolation regression:** `IsOwner` is `has_object_permission`-only — DRF will not invoke it for a custom action that returns a queryset (only `get_object()` does). Org scoping here relies entirely on `OrgAwareManager`. A user from org B requesting `/execution/<uuid_owned_by_org_A>/logs/export/?file_format=json` must return `[]` (or `404`), never org A's rows. Repeat for CSV (header-only body).
2. **413 envelope contract:** monkey-patch `MAX_SYNC_EXPORT_ROWS = 2`, seed 3 rows, assert status `413` and body keys `error` / `row_count=3` / `limit=2`. Frontend keys off the `error` field — a silent rename breaks UX with no signal.
3. **`file_format` validation + DRF reservation:** `?file_format=xml` → `400` with the exact error string; `?format=csv` is *ignored* (defaults to JSON). Locks in the comment-protected gotcha.
4. **`data` non-dict fallback:** mixed fixture (`data=None`, `data="raw"`, `data=["x"]`, `data={"level":"INFO","log":"ok"}`) returns `200` and CSV has 4 rows with the non-dict ones rendered as blank cells (no 500).
5. **`ExecutionLogFilter` passthrough:** the PR's headline guarantee. Seed across two `file_execution_id`s and three levels, assert `?file_execution_id=…&log_level=ERROR` returns only the matching rows.

`backend/workflow_manager/workflow_v2/` and `backend/workflow_manager/execution/` currently contain no tests for `ExecutionLog` views, so this would also seed a test module others can extend.

frontend/src/components/logging/log-modal/LogModal.jsx
```diff
@@ -84,6 +85,71 @@ function LogModal({
...
+        try {
+          const text = await err.response.data.text();
+          const parsed = JSON.parse(text);
+          setAlertDetails({
+            type: "error",
+            content: parsed?.error || "Export too large",
+          });
+          return;
+        } catch {
+          // fall through to default handler
```
muhammad-ali-e: **P1 — Silent fall-through swallows the entire 413 UX.** `catch {}` here swallows two distinct failure modes (`text()` rejects on a detached blob, or `JSON.parse` throws on a malformed body) with no log. When it fires, the user sees `handleException(err)`'s generic "Request failed with 413" instead of the carefully-crafted "narrow your filter" message — defeating the entire reason this special branch exists. A backend regression that changes the 413 envelope shape causes a silent UX downgrade with no signal.

Minimum fix: log the parse failure (`catch (parseErr) { console.error("Failed to parse 413 body:", parseErr); }`).

Better: keep a status-413-specific user message even when the body can't be parsed:
```js
setAlertDetails({ type: "error", content: "Export too large — narrow your filter and retry." });
return;
```

backend/workflow_manager/workflow_v2/execution_log_view.py
```diff
@@ -32,3 +43,86 @@ def get_queryset(self) -> QuerySet:
...
+        response = HttpResponse(body, content_type=content_type)
+        response["Content-Disposition"] = f'attachment; filename="{filename}"'
+        return response
+
+    def _build_csv(self, queryset: QuerySet) -> str:
+        output = io.StringIO()
+        writer = csv.writer(output)
+        writer.writerow(["event_time", "level", "stage", "log", "file_execution_id"])
+        for log in queryset:
+            data = log.data if isinstance(log.data, dict) else {}
```
muhammad-ali-e: **P2 — Non-dict `log.data` silently emits blank rows with no diagnostic trail.** When `log.data` is `None`, a string, or a list (legacy / migration-corrupted records — `data` is a bare `JSONField` with no schema), this guard correctly avoids a 500 but the CSV row gets all-blank `level` / `stage` / `log` cells indistinguishable from "field present but empty". The corruption is invisible to consumers and to ops.

```python
if isinstance(log.data, dict):
    data = log.data
else:
    logger.warning(
        "ExecutionLog %s has non-dict data of type %s; emitting blanks",
        log.id, type(log.data).__name__,
    )
    data = {}
```

Note `_build_json` (line 124) doesn't apply this guard at all — it passes `log.data` through verbatim. That's correct for JSON (faithful to the DB) but worth a comment so the asymmetry is intentional.

frontend/src/components/logging/log-modal/LogModal.jsx
```diff
@@ -167,6 +233,30 @@ function LogModal({
...
+            trigger={["click"]}
+            disabled={!pagination.total || exporting}
+          >
+            <Tooltip
+              title={
+                pagination.total
+                  ? exporting
+                    ? "Export in progress…"
+                    : "Export logs"
+                  : "No logs to export"
```
muhammad-ali-e: **P2 — Nested ternary is hard to scan and generally avoided in this codebase.** Extract a single binding above the JSX:

```jsx
let exportTooltip;
if (!pagination.total) exportTooltip = "No logs to export";
else if (exporting) exportTooltip = "Export in progress…";
else exportTooltip = "Export logs";
```

Reads top-to-bottom; no decoder ring needed.

frontend/src/components/logging/log-modal/LogModal.jsx
```diff
@@ -84,6 +85,71 @@ function LogModal({
...
+    setExporting(true);
+    try {
+      const url = getUrl(`/execution/${executionId}/logs/export/`);
+      const response = await axiosPrivate.get(url, {
+        params: {
+          // `file_format` (not `format`) — `format` is reserved by DRF for
+          // content negotiation and 404s when set to "csv".
+          file_format: fileFormat,
+          file_execution_id: fileId || "null",
+          log_level: selectedLogLevel,
```
muhammad-ali-e: **P2 — `log_level: selectedLogLevel` serializes as the literal string `"null"` when nothing is selected.** axios will turn JS `null` into `?log_level=null` (matching the deliberate-looking `file_execution_id: fileId || "null"` pattern on the previous line, which `ExecutionLogFilter.filter_file_execution` *does* handle as a sentinel). If `filter_log_level` does **not** treat the string `"null"` as "no filter", this silently exports an empty CSV — header only — and the user assumes there are no logs.

Worth confirming: either the filter accepts `"null"` symmetrically, or pass `undefined` / omit the param when empty here.

backend/workflow_manager/workflow_v2/execution_log_view.py
```diff
@@ -32,3 +43,86 @@ def get_queryset(self) -> QuerySet:
...
+                        f"Limit is {MAX_SYNC_EXPORT_ROWS}. "
+                        "Narrow the filter (e.g. by file or log level) and retry."
+                    ),
+                    "row_count": row_count,
+                    "limit": MAX_SYNC_EXPORT_ROWS,
+                },
+                status=status.HTTP_413_REQUEST_ENTITY_TOO_LARGE,
+            )
+
+        if export_format == "csv":
```
muhammad-ali-e, line 84: **P3 — No try/except around the build+respond block.** A failure mid-iteration (DB connection drop, `csv.Error` on a control char inside `data.log`, `JSONEncoder` choking on a non-serializable value somewhere in `log.data`) bubbles up as DRF's default 500 with no `logger.exception` context — operators get a Sentry trace with no execution_id / format / row_count.

```python
try:
    body = self._build_csv(queryset) if export_format == "csv" else self._build_json(queryset)
except Exception:
    logger.exception(
        "Log export failed for execution=%s format=%s rows=%s",
        execution_id, export_format, row_count,
    )
    return Response({"error": "Export failed; please retry."}, status=500)
```

backend/workflow_manager/workflow_v2/execution_log_view.py
```diff
@@ -32,3 +43,86 @@ def get_queryset(self) -> QuerySet:
...
+                    data.get("level", ""),
+                    data.get("stage", ""),
+                    data.get("log", ""),
+                    str(log.file_execution_id) if log.file_execution_id else "",
+                ]
+            )
+        return output.getvalue()
+
+    def _build_json(self, queryset: QuerySet) -> str:
+        entries = [
```
muhammad-ali-e, line 148: **P3 — `_build_csv` and `_build_json` duplicate per-row normalization.** Both independently emit `event_time.isoformat() if … else …` and `str(file_execution_id) if … else …`. Extract:

```python
def _normalize_log(self, log) -> dict:
    data = log.data if isinstance(log.data, dict) else {}
    return {
        "id": str(log.id),
        "event_time": log.event_time.isoformat() if log.event_time else None,
        "file_execution_id": str(log.file_execution_id) if log.file_execution_id else None,
        "level": data.get("level", ""),
        "stage": data.get("stage", ""),
        "log": data.get("log", ""),
        "data": log.data,
    }
```

Then CSV picks 5 keys, JSON picks 4. Single source of truth for null-handling and the dict-guard.

frontend/src/components/logging/log-modal/LogModal.jsx
```diff
@@ -84,6 +85,71 @@ function LogModal({
...
+          /filename="?([^"]+)"?/,
+        )?.[1] || `execution_logs_${fileId || executionId}.${fileFormat}`;
+      const blobUrl = window.URL.createObjectURL(new Blob([response.data]));
+      const link = document.createElement("a");
+      link.href = blobUrl;
+      link.download = filename;
+      document.body.appendChild(link);
+      link.click();
+      link.remove();
+      window.URL.revokeObjectURL(blobUrl);
```
muhammad-ali-e: **P3 — Third copy of the blob-download + filename-regex dance.** This pattern (Content-Disposition match, `createObjectURL`, hidden `<a>`, click, revoke) is duplicated in `frontend/src/components/custom-tools/header/Header.jsx` and `frontend/src/hooks/usePipelineHelper.js`. Good moment to extract a `downloadBlobResponse(response, fallbackName)` helper into `frontend/src/helpers/` since the pattern is now stable across three call sites.

muhammad-ali-e (commented):
Automated multi-agent review (Code Reviewer, Comment Analyzer, PR Test Analyzer, Silent Failure Hunter, Code Simplifier, Type Design Analyzer). Findings below in priority order. Greptile's prior comments on `LogModal.jsx:258` (Tooltip-on-disabled-Button) and `execution_log_view.py:89` (CSV `charset=utf-8`) already cover those issues — not re-raised.

backend/workflow_manager/workflow_v2/execution_log_view.py
```diff
@@ -32,3 +43,86 @@ def get_queryset(self) -> QuerySet:
...
+                },
+                status=status.HTTP_413_REQUEST_ENTITY_TOO_LARGE,
+            )
+
+        if export_format == "csv":
+            body = self._build_csv(queryset)
+            content_type = "text/csv"
+        else:
+            body = self._build_json(queryset)
+            content_type = "application/json"
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 1cfe52b3b — added `; charset=utf-8` to the CSV content type. Thanks for catching this; the implicit ISO-8859-1 default would absolutely have produced mojibake on non-ASCII log content (file names, non-Latin error text).

frontend/src/components/logging/log-modal/LogModal.jsx
```diff
@@ -167,6 +233,30 @@ function LogModal({
...
+            >
+              <Button
+                className="export-btn-outlined"
+                icon={<DownloadOutlined />}
+                loading={exporting}
+                disabled={!pagination.total || exporting}
+              >
+                Export
+              </Button>
+            </Tooltip>
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 1cfe52b3b — wrapped the disabled `Button` in a `<span>` so the `Tooltip` receives `mouseenter`/`mouseleave` even when the button is disabled.

backend/workflow_manager/workflow_v2/execution_log_view.py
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 1cfe52b3b — `text/csv; charset=utf-8`. (Duplicate of greptile finding above.)

frontend/src/components/logging/log-modal/LogModal.jsx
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 1cfe52b3b — wrapped in `<span>` per the standard antd workaround. (Duplicate of greptile finding above.)

backend/workflow_manager/workflow_v2/execution_log_view.py
muhammad-ali-e: Declining this one — the buffered approach was a deliberate choice after empirical analysis on a real workload (~17 KB per export, far below any memory pressure threshold) and aligns with how Datadog / Sentry / CloudWatch / Kibana / Grafana handle capped sync exports.

Trade-offs we accepted:
- **Browser progress %**: buffered sets `Content-Length`, streaming cannot.
- **Mid-stream error semantics**: buffered fails cleanly with a 500; streaming sends a truncated CSV and the client thinks it is complete.
- **Code complexity**: ~25 fewer lines without `_Echo` + generators.

The 50k row cap is the actual memory bound here. The pathological-row-size scenario (multi-KB markdown per row × 50k) is real but speculative — if it ever materializes, the right fix is async deployment-wide export (already deferred to Phase 2 in the PR description), not retrofitting streaming onto the sync path.

Happy to revisit if production telemetry shows otherwise.

backend/workflow_manager/workflow_v2/execution_log_view.py
muhammad-ali-e: Fixed in 1cfe52b3b — replaced `qs.count()` + iterate with `list(qs[:CAP+1])` for a single materialization that doubles as the over-cap sentinel. Drops one query and one full scan.

backend/workflow_manager/workflow_v2/execution_log_view.py
muhammad-ali-e: Acknowledged — flagging as a follow-up rather than blocking this PR.

`backend/workflow_manager/workflow_v2/` currently has no test infrastructure for `ExecutionLog` views, so adding the 5 test cases listed here means standing up fixtures + factories + auth client setup in addition to the assertions. That doubles the surface area of this PR and crosses the "separate unrelated changes" guideline in the contribution rules.

Will open a follow-up Jira to add the high-value cases inline (org isolation in particular — that one I would write today if `ExecutionLogFactory` and `WorkflowExecutionFactory` existed).

frontend/src/components/logging/log-modal/LogModal.jsx
muhammad-ali-e: Fixed in 1cfe52b3b — switched to a status-413-specific message that survives a parse failure of the JSON blob body, and added `console.error` so the parse failure itself is observable. The user always sees "Export too large — narrow your filter and retry." even if the backend envelope shape changes.

backend/workflow_manager/workflow_v2/execution_log_view.py
muhammad-ali-e: Fixed in 1cfe52b3b — added `logger.warning` with execution log id and the actual data type, hoisted the dict-guard into a shared `_normalize()` so the asymmetry between CSV (extracted fields) and JSON (raw passthrough) lives in one place. JSON path is documented inline as intentionally unguarded.

frontend/src/components/logging/log-modal/LogModal.jsx
muhammad-ali-e: Fixed in 1cfe52b3b — extracted to a named `exportTooltip` local with the if/else-if chain you suggested.

frontend/src/components/logging/log-modal/LogModal.jsx
muhammad-ali-e: Fixed in 1cfe52b3b — added a guard so `log_level` is only added to `params` when `selectedLogLevel` is truthy. Confirms the suspicion: the existing `filter_logs` does `levels.get(value, 1)` which defaults to INFO-min when value is `"null"`, silently dropping DEBUG. Sidesteps that by not sending the param at all.

(Worth flagging that the `list` endpoint earlier in this same component has the same pattern — same bug — but fixing it is out of scope here. Will note in the follow-up Jira for tests.)

backend/workflow_manager/workflow_v2/execution_log_view.py
muhammad-ali-e: Declining — DRF lets uncaught exceptions through to the configured exception middleware, which already logs the traceback (with full request context) to Sentry / GKE logs. Wrapping here would only add `execution_id` + `format` to that already-rich record, at the cost of a defensive `except Exception:` that the project style guide discourages outside system boundaries.

If we ever lose request context in the default exception path I would revisit, but right now the marginal observability gain does not justify the catch-all.

backend/workflow_manager/workflow_v2/execution_log_view.py
muhammad-ali-e: Fixed in 1cfe52b3b — extracted `_normalize()` exactly as suggested. CSV picks five flat keys, JSON picks four including the raw passthrough. Dict-guard and warning logging now live in one place.

frontend/src/components/logging/log-modal/LogModal.jsx
muhammad-ali-e: Acknowledged — will open a follow-up to extract `downloadBlobResponse(response, fallbackName)` into `frontend/src/helpers/`.

Keeping it out of this PR because it would require touching two unrelated call sites (`Header.jsx`, `usePipelineHelper.js`) for a refactor that has nothing to do with the export feature itself — exactly the "separate unrelated changes" pattern the contribution guidelines call out. Three call sites is also the conventional point at which extraction starts paying off, so the timing is right for a dedicated cleanup PR.

muhammad-ali-e:
## Review pass — summary

@greptile-apps @coderabbitai

Pushed `1cfe52b3b` addressing 10 of 14 review comments.

### Fixed
| # | Comment | Where |
|---|---|---|
| 1 | greptile P1 — `text/csv; charset=utf-8` | `execution_log_view.py` |
| 2 | greptile P2 — disabled-button tooltip via `<span>` wrapper | `LogModal.jsx` |
| 3 | coderabbit duplicate of #1 | (resolved by #1) |
| 4 | coderabbit duplicate of #2 | (resolved by #2) |
| 6 | self P2 — drop `count()`, single `list(qs[:CAP+1])` materialization | `execution_log_view.py` |
| 8 | self P1 — 413 alert survives blob parse failure | `LogModal.jsx` |
| 9 | self P2 — `logger.warning` on non-dict `data` | `execution_log_view.py` |
| 10 | self P2 — extract nested ternary into `exportTooltip` | `LogModal.jsx` |
| 11 | self P2 — skip `log_level` param when empty (axios serializes `null` as the literal `"null"`) | `LogModal.jsx` |
| 13 | self P3 — extract `_normalize()` (CSV/JSON share dict-guard) | `execution_log_view.py` |

### Declined / deferred (with rationale on each thread)
| # | Comment | Reason |
|---|---|---|
| 5 | self P1 — switch to `StreamingHttpResponse` | Buffered chosen deliberately; matches industry default for capped sync exports; row cap is the memory bound. |
| 7 | self P1 — no test coverage | Test scaffolding for `ExecutionLog` views does not exist; adding it inflates PR scope. Following up in a separate Jira. |
| 12 | self P3 — wrap build/respond in try/except | DRF default exception middleware already logs the traceback to Sentry / GKE; marginal context gain doesn't justify the catch-all. |
| 14 | self P3 — extract `downloadBlobResponse` helper | Touches two unrelated files; valid refactor but separate PR. |

Pre-commit clean on the changed files.

muhammad-ali-e:
@greptileai start review

Re-review request — the two issues flagged in the original 3/5 review (CSV charset and disabled-button tooltip span wrapper) have been fixed in 1cfe52b3b. SonarCloud findings (window → globalThis) fixed in 335d5a230.

pk-zipstack (approved):

athul-rs (approved):

PR #1952, opened 2026-05-11
UN-3152 [FEAT] Add AWS_BEARER_TOKEN_BEDROCK auth to Bedrock LLM and Embedding adapters

jaseemjaskp:
## What

Adds a third **Authentication Type** option — *Bedrock API Key (Bearer Token)* — to the AWS Bedrock LLM and Embedding adapters, alongside the existing **Access Keys** and **IAM Role / Instance Profile** modes.

## Why

AWS now issues short-lived Bedrock API keys (`AWS_BEARER_TOKEN_BEDROCK`) that authenticate via a single `Authorization: Bearer <token>` header instead of full SigV4 signing with long-lived IAM access keys. This is the AWS-recommended path for developer and CI workflows that need rotatable credentials without provisioning IAM users. Until now Unstract had no way to surface this option to end users.

## How

- Extended the shared `_resolve_bedrock_aws_credentials()` helper in `unstract/sdk1/src/unstract/sdk1/adapters/base1.py` to recognise `auth_type == "bearer_token"` and translate the UI's `aws_bearer_token` field to LiteLLM's `api_key` kwarg (which `litellm/llms/bedrock/base_aws_llm.py` consumes as the bearer token).
- Tightened the resolver so that selecting any auth mode strips stale credentials belonging to the others — same shape as the existing IAM-Role stale-key fix, applied symmetrically across all three modes.
- Added `aws_bearer_token: str | None = None` to both `AWSBedrockLLMParameters` and `AWSBedrockEmbeddingParameters`.
- Added a third `oneOf` branch with a password-formatted `aws_bearer_token` field to both `bedrock.json` schemas (LLM + embedding).
- Refactored the resolver into three small helpers (`_drop_bedrock_access_keys`, `_require_bedrock_access_keys`, `_translate_bedrock_bearer_token`) so the dispatch stays under the project's complexity threshold.

No regressions expected. The existing `auth_type` matrix (`None`, `access_keys`, `iam_role`) takes the same paths as before — verified by the original 14 tests, all of which still pass unchanged. The new logic only kicks in when a user explicitly selects the Bearer Token mode in the UI. Pre-existing adapters with no `auth_type` retain their lenient behaviour, and the new field defaults to `None` so loaded Pydantic models stay compatible.

## Database Migrations

- None.

## Env Config

- None required. End users may still set the `AWS_BEARER_TOKEN_BEDROCK` env var directly if they prefer LiteLLM's env-fallback path, but Unstract does not depend on it.

## Relevant Docs

- LiteLLM Bedrock provider: https://docs.litellm.ai/docs/providers/bedrock

## Related Issues or PRs

- Builds on the prior `auth_type` selector that introduced IAM Role / Instance Profile mode.

## Dependencies Versions

- LiteLLM **v1.82.3** (already pinned in `unstract/sdk1/pyproject.toml` via the Zipstack fork) — supports bearer-token auth out of the box.

## Notes on Testing

- Unit tests live in `unstract/sdk1/tests/test_bedrock_adapter.py`. New cases cover: bearer-token translation to `api_key`, stale access-key drop in bearer mode, blank/whitespace token rejection, stale-token drop when `iam_role` or `access_keys` is selected, and the legacy hand-saved-token translation path. Run locally with:

  ```sh
  cd unstract/sdk1 && uv run pytest tests/test_bedrock_adapter.py -v
  ```

  Result: **28 passed** (14 pre-existing + 14 new).

- Manual end-to-end smoke check (optional, requires a real Bedrock API key):
  - [ ] Rebuild Docker services that vendor `unstract/sdk1`.
  - [ ] In the UI, create a new Bedrock LLM adapter, switch **Authentication Type** to *Bedrock API Key (Bearer Token)*, paste the key, save.
  - [ ] Run a test prompt — request should succeed and LiteLLM logs should show `Authorization: Bearer …` rather than the SigV4 `AWS4-HMAC-SHA256 …` header.
  - [ ] Repeat for a Bedrock embedding adapter.

## Screenshots

_N/A — backend + JSON-schema-driven form changes only. The new dropdown option will render automatically from the updated `bedrock.json` schemas._

## Checklist


unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -483,33 +483,76 @@ def validate_model(adapter_metadata: dict[str, "Any"]) -> str:
...
+    With ``required=True`` (bearer_token mode), a blank/missing token is
+    a hard error. With ``required=False`` (legacy mode), a blank token
+    is silently dropped and a present token is still translated.
+    """
+    token = validated.pop(_BEDROCK_BEARER_TOKEN_FIELD, None)
+    if isinstance(token, str) and token.strip():
+        validated[_BEDROCK_LITELLM_BEARER_KWARG] = token
+        return
+    if required:
+        raise ValueError(
```
jaseemjaskp, line 516: **[P1 — silent-failure-hunter]** This `ValueError` (and the ones in `_require_bedrock_access_keys`, plus the unknown-`auth_type` raise at line 559) escapes the LLM path **unwrapped**:

- `embedding.py:106` → `except (ValidationError, ValueError) as e: raise SdkError("Invalid embedding adapter metadata: " + str(e)) from e`
- `llm.py:195` → `except ValidationError as e: raise SdkError("Invalid LLM adapter metadata: " + str(e)) from e` ← misses `ValueError`

Downstream `except SdkError:` blocks (`llm.py:242/343/436/511/580`) will not handle these and they surface as raw `ValueError`. The asymmetry was pre-existing for the access-keys branch, but this PR widens the surface (3 more ValueError sites). Fix is a one-character change at `llm.py:195`:

```diff
-        except ValidationError as e:
+        except (ValidationError, ValueError) as e:
             raise SdkError("Invalid LLM adapter metadata: " + str(e)) from e
```

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -519,23 +562,27 @@ def _resolve_bedrock_aws_credentials(
...
-    # Legacy adapters with no auth_type: strip blanks silently to
-    # preserve the pre-PR behaviour where empty key fields fell through
-    # to boto3's default chain.
+    # Legacy adapters with no auth_type: strip blank access keys (boto3
+    # default chain takes over) and translate any non-blank bearer token
+    # for parity with bearer_token mode.
     for key in _BEDROCK_AWS_KEY_FIELDS:
         if not validated.get(key):
             validated.pop(key, None)
+    _translate_bedrock_bearer_token(validated, required=False)
```
jaseemjaskp: **[P1 — code-reviewer / silent-failure-hunter]** Legacy auto-translation is a silent override. A pre-PR adapter persisted with `aws_bearer_token` (no `auth_type`) reaches this branch and gets its token promoted into `api_key`, **overriding any env-injected `AWS_BEARER_TOKEN_BEDROCK` or boto3 default-chain creds the host has** — with zero log lines. After an ops rotation, the adapter will keep using the stale stored token and the cause will be invisible.

The docstring at lines 547-550 acknowledges this is intentional ("if a non-blank bearer token was hand-saved, translate it to api_key"), but the only way to reach this state is manual DB editing — re-saving through the new UI sets `auth_type="access_keys"` (the default).

Fix options:
1. (Minimum) Emit a `logger.warning("Legacy Bedrock adapter has stored aws_bearer_token without auth_type; auto-promoting to bearer-token mode. Set auth_type='bearer_token' explicitly to silence this.")` immediately before the translate call.
2. (Stricter) Skip the legacy translation entirely — `pop(_BEDROCK_BEARER_TOKEN_FIELD, None)` so a `None`/stale `auth_type` cannot be a stealth bearer-token config.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -483,33 +483,76 @@ def validate_model(adapter_metadata: dict[str, "Any"]) -> str:
...
+def _translate_bedrock_bearer_token(
+    validated: dict[str, "Any"], *, required: bool
+) -> None:
+    """Move ``aws_bearer_token`` (UI field) to ``api_key`` (LiteLLM kwarg).
+
+    With ``required=True`` (bearer_token mode), a blank/missing token is
+    a hard error. With ``required=False`` (legacy mode), a blank token
+    is silently dropped and a present token is still translated.
+    """
+    token = validated.pop(_BEDROCK_BEARER_TOKEN_FIELD, None)
```
jaseemjaskp, line 514: **[P2 — silent-failure-hunter]** Type confusion. After Pydantic's `model_dump()` the value is always `str | None`, but this helper is reachable from tests and any future direct caller with a raw dict. If `aws_bearer_token=b"token"` or `aws_bearer_token=12345`:

- `required=False` (legacy): silently popped and dropped — no error, no `api_key` set, no log.
- `required=True` (bearer mode): raises `"aws_bearer_token is required"` — **misleading**; the user supplied one but with the wrong type.

Fix — distinguish missing vs wrong type:

```python
token = validated.pop(_BEDROCK_BEARER_TOKEN_FIELD, None)
if token is None or (isinstance(token, str) and not token.strip()):
    if required:
        raise ValueError(f"{_BEDROCK_BEARER_TOKEN_FIELD} is required when auth_type is 'bearer_token'.")
    return
if not isinstance(token, str):
    raise ValueError(f"{_BEDROCK_BEARER_TOKEN_FIELD} must be a string, got {type(token).__name__}.")
validated[_BEDROCK_LITELLM_BEARER_KWARG] = token.strip()
```

This also folds in the strip-on-store concern (existing greptile note at line 520) by writing the stripped value, not the original.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -519,23 +562,27 @@ def _resolve_bedrock_aws_credentials(
...
-        for key in _BEDROCK_AWS_KEY_FIELDS:
-            value = validated.get(key)
-            if not isinstance(value, str) or not value.strip():
-                raise ValueError(f"{key} is required when auth_type is 'access_keys'.")
+        _require_bedrock_access_keys(validated)
+        validated.pop(_BEDROCK_BEARER_TOKEN_FIELD, None)
+        return validated
+
+    if auth_type == "bearer_token":
+        _drop_bedrock_access_keys(validated)
```
jaseemjaskp, line 564: **[P2 — silent-failure-hunter]** Silent strip of conflicting credentials. If the user submits both `aws_bearer_token` and `aws_access_key_id`/`aws_secret_access_key` (e.g., switched the auth radio but didn't clear the previously-typed access-key fields), `_drop_bedrock_access_keys` quietly discards them — no warning. Symmetric issue at line 571 where stray bearer tokens are dropped silently in `access_keys` mode.

Not a security issue (the backend honors `auth_type`), but a poor UX: the user thinks their access keys are still in play. Suggest:

```python
if auth_type == "bearer_token":
    leftover_keys = [k for k in _BEDROCK_AWS_KEY_FIELDS if validated.get(k)]
    if leftover_keys:
        logger.warning(
            "Bedrock adapter in bearer_token mode received non-blank %s; ignoring.",
            leftover_keys,
        )
    _drop_bedrock_access_keys(validated)
    _translate_bedrock_bearer_token(validated, required=True)
    return validated
```

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -483,33 +483,76 @@ def validate_model(adapter_metadata: dict[str, "Any"]) -> str:
...
 )
+# LiteLLM consumes the Bedrock bearer token as the `api_key` kwarg (see
+# litellm/llms/bedrock/base_aws_llm.py). We accept it from the UI under the
+# more descriptive `aws_bearer_token` name and translate at resolve time.
+_BEDROCK_BEARER_TOKEN_FIELD: str = "aws_bearer_token"
+_BEDROCK_LITELLM_BEARER_KWARG: str = "api_key"
 _BEDROCK_VALID_AUTH_TYPES: frozenset[str | None] = frozenset(
-    {None, "access_keys", "iam_role"}
+    {None, "access_keys", "iam_role", "bearer_token"}
 )
```
jaseemjaskp, line 492: **[P2 — type-design-analyzer]** The closed set `{None, "access_keys", "iam_role", "bearer_token"}` is duplicated as bare strings across (a) the four branch checks in `_resolve_bedrock_aws_credentials`, (b) `llm1/static/bedrock.json:71-73`, (c) `embedding1/static/bedrock.json:47-49`, and (d) ~14 test cases. A typo today (`"bearer-token"`) fails only at runtime.

Lift to a `StrEnum` so the values are the source of truth:

```python
class BedrockAuthType(StrEnum):
    ACCESS_KEYS = "access_keys"
    IAM_ROLE = "iam_role"
    BEARER_TOKEN = "bearer_token"
```

Then branch on `BedrockAuthType(auth_type)` (with a tight try/except for the unknown-value path) and use `[t.value for t in BedrockAuthType]` when generating/asserting the JSON enums in tests. The legacy `None` case is lifted out of the set into the resolver (where it semantically lives — "absent" is not a valid auth choice, it's a pre-PR state).

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -544,6 +591,7 @@ class AWSBedrockLLMParameters(BaseChatCompletionParameters):
 
     aws_access_key_id: str | None = None
     aws_secret_access_key: str | None = None
+    aws_bearer_token: str | None = None  # AWS_BEARER_TOKEN_BEDROCK
```
jaseemjaskp: **[P2 — type-design-analyzer / comment-analyzer]** Two things on this line:

1. **Type doesn't carry the invariant.** `str | None` admits `""` and `"   "`, but `_translate_bedrock_bearer_token` requires non-empty-after-strip. Push the rule into the type so it holds for any caller, not just `validate()`:
   ```python
   from typing import Annotated
   from pydantic import StringConstraints
   NonEmptyStr = Annotated[str, StringConstraints(min_length=1, strip_whitespace=True)]
   aws_bearer_token: NonEmptyStr | None = None
   ```
   Mirrors the `minLength: 1` you'll be adding to the JSON schemas (per existing reviewer note).

2. **Comment is opaque.** `# AWS_BEARER_TOKEN_BEDROCK` breaks the descriptive pattern of sibling fields (`# For AWS SSO authentication`, `# For Application Inference Profile (cost tracking)`). Suggest: `# Short-lived Bedrock API key (AWS_BEARER_TOKEN_BEDROCK); translated to api_key`. Apply the same change to the duplicate on line 1084.

jaseemjaskp (commented):
## PR Review Toolkit — multi-agent review

Findings from `code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`, `type-design-analyzer`, and `code-simplifier`. Existing greptile/coderabbit notes (`base1.py:520/525`, `test_bedrock_adapter.py:396`, `minLength` on both schemas at `:134`/`:113`) are not re-posted.

### Highlights

- **P1 — LLM error-wrapping asymmetry (out-of-diff, surfaced below at `base1.py:522`).** `embedding.py:106` catches `(ValidationError, ValueError)` and wraps in `SdkError`; `llm.py:195` catches only `ValidationError`. Every new `ValueError` this PR adds (unknown auth_type, blank bearer, blank access keys) will escape the LLM path unwrapped and bypass downstream `except SdkError:` blocks (`llm.py:242/343/436/511/580`). Recommend changing `llm.py:195` to `except (ValidationError, ValueError) as e:` for parity.
- **P1 — Legacy silent auto-translation** of a hand-stored `aws_bearer_token` (no `auth_type`) into `api_key` can override env-injected credentials with no log line. See `base1.py:585`.
- **P1 — Test parity drift is structurally guaranteed** by the duplicated LLM/Embedding test blocks. Recommend parametrizing across `[AWSBedrockLLMParameters, AWSBedrockEmbeddingParameters]`. See `test_bedrock_adapter.py:160`.
- **P2 — Multiple silent-drop paths** (type confusion on non-string token, silent strip of conflicting access keys submitted alongside bearer) — see inline.
- **P2 — Type-design** opportunities: `StrEnum` for `auth_type` (4 sites + 2 JSON schemas duplicate the strings); `Annotated[str, StringConstraints(min_length=1, strip_whitespace=True)]` for `aws_bearer_token` to push the existing `minLength`/`strip` invariants into the type system.

### Positive

- `_resolve_bedrock_aws_credentials` docstring is exemplary (enumerates each branch + the *why*).
- Cross-mode stale-credential scrubbing is well-tested in both directions (bearer→keys/iam and keys/iam→bearer).
- Error messages use `match=` regex (not substring) and tests target the public `validate()` API rather than the private helpers — survives refactoring.
- `auth_type` is correctly stripped from `validated` before reaching LiteLLM kwargs.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -483,33 +483,76 @@ def validate_model(adapter_metadata: dict[str, "Any"]) -> str:
...
+
+def _require_bedrock_access_keys(validated: dict[str, "Any"]) -> None:
+    for key in _BEDROCK_AWS_KEY_FIELDS:
+        value = validated.get(key)
+        if not isinstance(value, str) or not value.strip():
+            raise ValueError(f"{key} is required when auth_type is 'access_keys'.")
+
+
+def _translate_bedrock_bearer_token(
+    validated: dict[str, "Any"], *, required: bool
```
jaseemjaskp: **[P3 — type-design-analyzer]** `required: bool` is a re-encoding of `auth_type`. Both call sites are static (`required=True` for `bearer_token`, `required=False` for legacy `None`), so the boolean adds no information that the existing `auth_type` couldn't carry directly.

Options, pick one:
- Pass `auth_type` (or `BedrockAuthType | None` if you adopt the StrEnum) into this function — self-documenting at call sites.
- Split into `_translate_bedrock_bearer_token_required(validated)` and `_translate_bedrock_bearer_token_optional(validated)` — removes the branch entirely.

Not urgent; flag if you're touching this surface again.

unstract/sdk1/tests/test_bedrock_adapter.py
```diff
@@ -149,6 +149,109 @@ def test_llm_other_params_preserved_through_strip() -> None:
...
 
+# ── LLM: bearer token (AWS_BEARER_TOKEN_BEDROCK) ─────────────────────────────
+
+
+def test_llm_bearer_token_mode_translates_to_api_key() -> None:
+    """Bearer token is exposed to LiteLLM under its `api_key` kwarg."""
+    out = AWSBedrockLLMParameters.validate(
+        {
+            "auth_type": "bearer_token",
+            "model": "anthropic.claude-3-haiku-20240307-v1:0",
```
jaseemjaskp, line 172: **[P1 — code-simplifier / pr-test-analyzer]** This block and lines 355-437 are ~190 lines of near-identical mirror tests, differing only by parameter class and model string. The same duplication produced the missing-embedding-whitespace gap flagged by greptile at `:396` — parametrization would make that whole class of bug structurally impossible.

Replace with:

```python
@pytest.fixture(
    params=[
        (AWSBedrockLLMParameters, "anthropic.claude-3-haiku-20240307-v1:0"),
        (AWSBedrockEmbeddingParameters, "amazon.titan-embed-text-v2:0"),
    ],
    ids=["llm", "embedding"],
)
def bedrock_adapter(request):
    return request.param

def test_bearer_token_translates_to_api_key(bedrock_adapter):
    cls, model = bedrock_adapter
    out = cls.validate({
        "auth_type": "bearer_token",
        "model": model,
        "region_name": "us-east-1",
        "aws_bearer_token": "bedrock-key-abc",
    })
    assert out["api_key"] == "bedrock-key-abc"
    assert "aws_bearer_token" not in out
    assert "aws_access_key_id" not in out
    assert "aws_secret_access_key" not in out
    assert "auth_type" not in out
```

Keep the genuinely one-sided tests (`test_llm_other_params_preserved_through_strip` — LLM-only `thinking` config; `test_embedding_region_required_when_absent` — embedding-only) outside the parametrization. Net: ~half the file gone, parity drift prevented by construction.

unstract/sdk1/tests/test_bedrock_adapter.py
```diff
@@ -149,6 +149,109 @@ def test_llm_other_params_preserved_through_strip() -> None:
...
+def test_llm_bearer_token_mode_whitespace_token_raises() -> None:
+    with pytest.raises(ValueError, match="aws_bearer_token is required"):
+        AWSBedrockLLMParameters.validate(
+            {
+                "auth_type": "bearer_token",
+                "model": "anthropic.claude-3-haiku-20240307-v1:0",
+                "region_name": "us-east-1",
+                "aws_bearer_token": "   ",
+            }
+        )
```
jaseemjaskp, line 222: **[P2 — pr-test-analyzer]** Test gaps in this block:

1. **Missing `aws_bearer_token` key entirely** (not just blank): `cls.validate({"auth_type": "bearer_token", "model": ..., "region_name": ..., })` should raise. Distinct code path — `validated.pop(...)` returns `None`, `isinstance(None, str)` is False, then `required=True` raises. Currently unpinned.
2. **Non-string token type**: `aws_bearer_token=12345`. After Pydantic `str | None` typing this raises `ValidationError`, but locking the contract prevents a future relaxation of the field type from silently breaking it.
3. **Legacy mode with blank token** (`required=False` branch): `validate({"model": ..., "region_name": ..., "aws_bearer_token": ""})` should silently drop with `"api_key" not in out` — currently untested.
4. **Embedding suite (line 328)**: `test_embedding_unknown_auth_type_raises` only covers `"iamrole"`. The LLM suite has `"access_key"`. Add a bearer-typo case (`"bearer-token"`) on both sides; new modes are exactly where typo coverage is most valuable.

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/bedrock.json
```diff
@@ -69,14 +69,16 @@
...
+        "IAM Role / Instance Profile (on-prem AWS only)",
+        "Bedrock API Key (Bearer Token)"
       ],
       "default": "access_keys",
-      "description": "Choose **Access Keys** for any deployment (provide your AWS Access Key ID and Secret). Choose **IAM Role / Instance Profile** only when Unstract is hosted on AWS infrastructure that already has ambient credentials — for example EKS pods with IRSA, ECS tasks with a task role, or EC2 instances with an instance profile. The on-prem option requires no further input; boto3 picks up the host's identity automatically."
+      "description": "Choose **Access Keys** for any deployment (provide your AWS Access Key ID and Secret). Choose **IAM Role / Instance Profile** only when Unstract is hosted on AWS infrastructure that already has ambient credentials — for example EKS pods with IRSA, ECS tasks with a task role, or EC2 instances with an instance profile. Choose **Bedrock API Key (Bearer Token)** to authenticate with a short-lived AWS Bedrock API key (`AWS_BEARER_TOKEN_BEDROCK`) — recommended for developers and CI workflows that need rotatable credentials without provisioning IAM users."
```
jaseemjaskp: **[P2 — comment-analyzer]** Two issues with the bearer-token description fragment:

1. **Subjective and rot-prone.** "recommended for developers and CI workflows that need rotatable credentials" steers users away from production where AWS markets these for both ephemeral and long-lived use. As AWS guidance evolves this becomes misleading.
2. **Missing critical UX detail.** Bedrock API keys are **region-scoped**. Because `aws_region_name` is a separate required field, a user pasting a `us-east-1` token while selecting `us-west-2` will hit an opaque 401 with no schema hint that the two must match.

Suggested rewrite for the trailing sentence:

> Choose **Bedrock API Key (Bearer Token)** to authenticate with an AWS Bedrock API key (`AWS_BEARER_TOKEN_BEDROCK`). The token is region-scoped — ensure the AWS Region below matches the key's region.

Apply the same change to the mirror in `embedding1/static/bedrock.json:57`.

unstract/sdk1/src/unstract/sdk1/adapters/embedding1/static/bedrock.json
```diff
@@ -45,14 +45,16 @@
...
+          "IAM Role / Instance Profile (on-prem AWS only)",
+          "Bedrock API Key (Bearer Token)"
         ],
         "default": "access_keys",
-        "description": "Choose **Access Keys** for any deployment (provide your AWS Access Key ID and Secret). Choose **IAM Role / Instance Profile** only when Unstract is hosted on AWS infrastructure that already has ambient credentials — for example EKS pods with IRSA, ECS tasks with a task role, or EC2 instances with an instance profile. The on-prem option requires no further input; boto3 picks up the host's identity automatically."
+        "description": "Choose **Access Keys** for any deployment (provide your AWS Access Key ID and Secret). Choose **IAM Role / Instance Profile** only when Unstract is hosted on AWS infrastructure that already has ambient credentials — for example EKS pods with IRSA, ECS tasks with a task role, or EC2 instances with an instance profile. Choose **Bedrock API Key (Bearer Token)** to authenticate with a short-lived AWS Bedrock API key (`AWS_BEARER_TOKEN_BEDROCK`) — recommended for developers and CI workflows that need rotatable credentials without provisioning IAM users."
```
jaseemjaskp: **[P2 — comment-analyzer]** Mirror finding to the LLM schema: drop the subjective "recommended for developers and CI" framing and add a one-clause region-scoping warning so users with multi-region setups don't hit opaque 401s. See suggested rewrite in the `llm1/static/bedrock.json:81` comment — keep the two strings textually identical so future copy-paste keeps them in sync.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -483,33 +483,76 @@ def validate_model(adapter_metadata: dict[str, "Any"]) -> str:
...
+    """
+    token = validated.pop(_BEDROCK_BEARER_TOKEN_FIELD, None)
+    if isinstance(token, str) and token.strip():
+        validated[_BEDROCK_LITELLM_BEARER_KWARG] = token
+        return
+    if required:
+        raise ValueError(
+            f"{_BEDROCK_BEARER_TOKEN_FIELD} is required when "
+            "auth_type is 'bearer_token'."
+        )
```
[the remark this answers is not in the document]
jaseemjaskp: Skipping with reason: bearer mode requiring an explicit token mirrors how `access_keys` mode requires non-blank keys — both are explicit-config modes. Users who want env-injected `AWS_BEARER_TOKEN_BEDROCK` should pick `iam_role` mode (drops access keys but doesn't touch env vars; LiteLLM's bedrock module picks the env var up downstream) or legacy / no-`auth_type` mode. Allowing the field to pass-through here would break the explicit-config-required philosophy of the `auth_type` selector — same reason `access_keys` mode raises on blank values rather than falling through to the boto3 default chain.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
jaseemjaskp: Skipping. `_translate_bedrock_bearer_token` reads `aws_bearer_token` from `validated`, which is the output of `AWSBedrock{LLM,Embedding}Parameters(**…).model_dump()`. Pydantic enforces the `str | None` typing before the resolver ever sees the value, so a non-string at runtime is impossible. Adding defensive type-checking inside the helper would duplicate Pydantic's contract without buying coverage. The only failure modes that actually reach the helper are `None` and blank-string — both produce the same `aws_bearer_token is required …` error, which is correct.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
jaseemjaskp: Skipping for this PR. The pre-existing `iam_role` branch (`_drop_bedrock_access_keys`) also silently drops access keys without a warning, and `access_keys` mode silently drops a stray bearer token now too. Adding a `logger.warning` only to the `bearer_token` site would be inconsistent. Worth a dedicated follow-up that adds warnings symmetrically to all three "drop other modes' creds" call sites — that's a UX-cleanup PR rather than part of this feature.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
jaseemjaskp: Skipping for this PR. Lifting to a `StrEnum` is a worthwhile refactor but it touches the resolver, both `bedrock.json` schemas, and the test file in lockstep — broader blast radius than this feature warrants. Worth a dedicated follow-up that lifts the source of truth, regenerates the JSON enums from `[t.value for t in BedrockAuthType]`, and migrates the tests in one atomic change.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
jaseemjaskp: **Comment improvement (#2):** applied — both classes now read `# Short-lived Bedrock API key (AWS_BEARER_TOKEN_BEDROCK); the resolver translates it to LiteLLM's \`api_key\` kwarg.`

**Annotated `NonEmptyStr` (#1):** skipping. Pydantic with `StringConstraints(min_length=1, strip_whitespace=True)` would raise `ValidationError("string_too_short")` for the blank case — less actionable than the explicit `ValueError("aws_bearer_token is required when auth_type is 'bearer_token'.")` the resolver currently raises (which names the field, the mode, and the failure in one line). Client-side rejection is already covered by the `"minLength": 1` JSON-schema constraint added per the greptile review. Server-side rejection is intentionally kept in the resolver where the message can stay specific to the auth mode that triggered it.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
jaseemjaskp: Resolved differently in the latest commit. The legacy `aws_bearer_token` auto-translation was removed entirely (per the related `:585` thread / option 2), so `_translate_bedrock_bearer_token` now has only one call site (`bearer_token` mode) and no longer needs a `required` parameter — it unconditionally requires a non-blank token. The P3 concern about a boolean re-encoding `auth_type` is moot because the boolean is gone.

unstract/sdk1/tests/test_bedrock_adapter.py
jaseemjaskp: Skipping for this PR. Fixture-based parametrization is a worthwhile structural improvement but it rewrites ~190 lines of the test file and would mix mechanical churn with the feature change in the same diff. Worth a dedicated follow-up that introduces the parametrized fixture and migrates the existing per-mode duplication in one atomic refactor.

The narrow parity gap that motivated this comment (greptile :396 — missing embedding whitespace test) is fixed; the latest commit also adds parallel coverage for missing-key, strip-on-store, and bearer-typo on both sides, so the LLM/embedding suites are in lockstep again.

unstract/sdk1/tests/test_bedrock_adapter.py
jaseemjaskp: **Cases 1 (missing key) and 4 (bearer-typo):** added on both LLM and embedding sides — `test_{llm,embedding}_bearer_token_mode_missing_token_raises` and `test_{llm,embedding}_unknown_bearer_token_typo_raises`.

**Case 2 (non-string token type):** skipping. The Pydantic `str | None` typing on the field already raises `ValidationError` before the resolver ever sees the value, so a test here would be exercising Pydantic's contract, not our code. If we ever loosen the field type that's the moment to add a test — until then the test would be a tautology.

**Case 3 (legacy mode with blank token):** moot. The legacy auto-translation was removed entirely in the latest commit (per the related `:585` thread / option 2), so legacy mode now drops `aws_bearer_token` unconditionally — there's no blank-vs-set distinction to test. The new `test_{llm,embedding}_legacy_drops_stray_bearer_token` cases pin the unconditional drop.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -519,23 +563,29 @@ def _resolve_bedrock_aws_credentials(
...
+        _drop_bedrock_access_keys(validated)
+        _translate_bedrock_bearer_token(validated)
         return validated
 
-    # Legacy adapters with no auth_type: strip blanks silently to
-    # preserve the pre-PR behaviour where empty key fields fell through
-    # to boto3's default chain.
+    # Legacy adapters with no auth_type: strip blank access keys (boto3
+    # default chain takes over). Drop any stray bearer token rather than
+    # auto-promoting it — the field is new in this PR, so any value here
```
chandrasekharan-zipstack: @jaseemjaskp please check and clean all comments added in this PR. They seem specific to your conversations / discussions and are not generic enough to make sense when the code is read after merging

unstract/sdk1/src/unstract/sdk1/llm.py
```diff
@@ -190,7 +189,10 @@ def __init__(  # noqa: C901
             #     if s not in self.kwargs:
             #         logger.warning("Missing supported parameter for '%s': %s",
             #             self.adapter.get_provider(), s)
-        except ValidationError as e:
+        except ValueError as e:
+            # `pydantic.ValidationError` is a subclass of `ValueError`; this
+            # catches both Pydantic validation failures and the explicit
+            # `ValueError`s raised by the Bedrock auth resolver.
```
chandrasekharan-zipstack: Similarly this comment also, if its not required or noisy, let's remove / clean it a bit

chandrasekharan-zipstack (approved):
LGTM but left a note on the comments itself

pk-zipstack (approved):

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
jaseemjaskp: Good catch — done in `5e515d52`. Audited every comment/docstring added in this PR:

- **Constants block:** dropped the `see litellm/llms/bedrock/base_aws_llm.py` file pointer (rots if upstream restructures); kept only the WHY (UI uses `aws_bearer_token`, LiteLLM uses `api_key`).
- **`_translate_bedrock_bearer_token` docstring:** collapsed to two sentences. The AWS opaque-401 rationale for stripping whitespace is the only non-obvious bit worth keeping.
- **`_resolve_bedrock_aws_credentials` docstring:** lifted the shared safety invariant ("each branch drops *other* modes' creds; explicit modes raise on blank values") into one paragraph above the per-mode bullets, so each bullet is now one terse line.
- **Legacy-branch inline comment:** dropped "the field is new in this PR" — describes the rule (bearer auth must be opted into explicitly) instead of the history.
- **Field comments on both parameter classes:** trimmed to one line — the field name already says "bearer token".
- **Test docstrings:** dropped "brand new in this PR" wording; deleted two redundant ones whose test names already convey intent (`test_llm_unknown_bearer_token_typo_raises`, `test_embedding_legacy_drops_stray_bearer_token`).

unstract/sdk1/src/unstract/sdk1/llm.py
jaseemjaskp: Trimmed in `5e515d52`. The Bedrock-resolver-specific reference is gone; the comment now keeps only the load-bearing invariant — `pydantic.ValidationError` subclasses `ValueError`, so the seemingly-narrow `except ValueError` actually catches Pydantic validation too. Without that one-liner a future reader could "fix" the apparent bug by adding `ValidationError` back (which is what triggered SonarCloud's `python:S5713` in the first place).

```python
except ValueError as e:
    # `pydantic.ValidationError` subclasses `ValueError` — this catches both.
    raise SdkError("Invalid LLM adapter metadata: " + str(e)) from e
```

PR #1955, opened 2026-05-12
UN-3444 [FEAT] Propagate frontend request ID and surface it on error notifications

Deepak-Kesavan:
## What

- Add an axios request interceptor that injects `X-Request-ID: <uuidv4>` on every outgoing API call (both on the global `axios` default instance and on each `useAxiosPrivate` instance).
- Extract the request ID from `err.response.headers['x-request-id']` (with a fallback to `err.config.headers['X-Request-ID']` for network/cancel errors) and thread it through `useExceptionHandler` and the alert store.
- Render the request ID inline in the error notification toast with antd's `Typography.Text copyable` (one-click copy).
- Append the request ID to the corresponding row in the bottom Logs & Notifications panel so it remains visible after the toast closes.

## Why

- Support and on-call needed a deterministic correlation ID to tie a UI error to backend/worker logs. Django and Flask middleware in the backend already honor an incoming `X-Request-ID`, but the frontend was neither sending one nor surfacing it to users.

## How

- New helper module `frontend/src/helpers/requestId.js` exposes `attachRequestIdInterceptor(instance)` and `getRequestIdFromError(err)`. Interceptor only sets the header if one is not already on the request, so service-to-service hops (e.g. `tool-sandbox` already sends `X-Request-ID`) aren't overwritten.
- `App.jsx` attaches the interceptor to the global `axios` import at module load; its returned handle is exported as `globalRequestIdInterceptor` so HMR-friendly cleanup is possible.
- `useAxiosPrivate.js` registers + ejects the interceptor alongside the existing response interceptor.
- `useExceptionHandler.jsx` adds `requestId` to every `buildAlert` return path (including the no-response branches).
- `App.jsx` renders the request ID line in the notification description (copyable widget) and appends `Request ID: \`<id>\`` to the log message that `pushLogMessages` writes — the table column already renders markdown via `CustomMarkdown`, which formats the UUID as an inline code pill.
- `alert-store.js` extends the default alert shape with `requestId`.
- `index.css` adds BEM-styled `.notification-request-id` for spacing/layout in the toast.

- No. The interceptor only sets the header when missing, so any existing flow that already sets `X-Request-ID` (e.g. `tool-sandbox` → runner) keeps its value. `useExceptionHandler` retains every prior return path; `requestId` is an additive optional field. The notification only renders the new line when an error alert has a request ID. The log message is augmented for `NOTIFICATION` rows only — `LOG` rows from socket messages are unchanged.

## Database Migrations

- None.

## Env Config

- None.

## Relevant Docs

- Backend middleware already wired (`backend/middleware/request_id.py`, `backend/backend/settings/base.py` — `LOG_REQUEST_ID_HEADER`, `REQUEST_ID_RESPONSE_HEADER`, `GENERATE_REQUEST_ID_IF_NOT_IN_HEADER`).
- Flask middleware: `unstract/core/src/unstract/core/flask/middleware.py:assign_request_id`.

## Related Issues or PRs

- UN-3444

## Dependencies Versions

- Uses the existing `uuid` dep (already in `frontend/package.json`). No new dependencies.

## Notes on Testing

- Trigger any failing API call (e.g. invalid payload, expired auth, backend down) and verify:
  - The error toast shows `Request ID: <uuid>` with a copy icon that copies the raw UUID.
  - The bottom Logs panel row for the same error shows the request ID on a second line.
  - The Network tab's request headers include `X-Request-ID` and the value matches what is shown in the UI.
  - Backend logs (Django/Flask) for that request include the same UUID.
- Offline path: disable network, retry — confirm the toast still shows the request ID (sourced from the outgoing request config).
- Canceled request path: cancel an in-flight request — confirm the request ID is still surfaced.
- Confirm non-error alerts (success/warning) do **not** render the request ID line.
- Cross-origin caveat: if FE and BE are on different origins, reading `x-request-id` from the response requires `CORS_EXPOSE_HEADERS = ["X-Request-ID"]` on the backend; the fallback to `err.config.headers` keeps the feature working regardless because the backend reuses the header we sent.

## Screenshots

image

## Checklist


harini-venkataraman (approved):

frontend/src/helpers/requestId.js
```diff
@@ -0,0 +1,21 @@
+import { v4 as uuidv4 } from "uuid";
+
+const REQUEST_ID_HEADER = "X-Request-ID";
+
+const attachRequestIdInterceptor = (axiosInstance) => {
+  return axiosInstance.interceptors.request.use((config) => {
+    if (!config.headers[REQUEST_ID_HEADER]) {
```
vishnuszipstack: **[P1] Use the AxiosHeaders API and guard against an undefined `headers`.** This repo pins `axios@1.13.5`; inside a request interceptor `config.headers` is an `AxiosHeaders` instance, not a plain object. Bracket get/set bypasses its case-insensitive normalization, and `config.headers` can be `undefined` for hand-built request configs — that throws inside the interceptor and rejects *every* outbound request as a generic-looking network error.

Suggested:
```js
return axiosInstance.interceptors.request.use((config) => {
  config.headers ??= {};
  if (!config.headers[REQUEST_ID_HEADER]) {
    config.headers[REQUEST_ID_HEADER] = uuidv4();
  }
  return config;
});
```
More robust still: `if (!config.headers.has?.(REQUEST_ID_HEADER)) config.headers.set(REQUEST_ID_HEADER, uuidv4(), false);` — the trailing `false` means "do not overwrite an existing value", which also preserves service-to-service IDs.

frontend/src/helpers/requestId.js
```diff
@@ -0,0 +1,21 @@
+import { v4 as uuidv4 } from "uuid";
+
+const REQUEST_ID_HEADER = "X-Request-ID";
+
+const attachRequestIdInterceptor = (axiosInstance) => {
```
vishnuszipstack, line 15: **[P2] No tests were added; these helpers are pure and trivially testable.** The repo has vitest configured (`happy-dom`, existing `*.test.js[x]` files). Recommend a `requestId.test.js` covering:
- interceptor injects a v4 UUID when the header is absent;
- interceptor does NOT overwrite a caller-supplied `X-Request-ID`;
- a distinct ID is generated per request;
- `getRequestIdFromError` reads the response header, falls back to the config header, and returns `undefined` for `null`/`{}` errors without throwing.

The response-vs-config casing split on line 16 is exactly the fragile branch tests should lock down.

frontend/src/helpers/requestId.js
```diff
@@ -0,0 +1,21 @@
...
+    if (!config.headers[REQUEST_ID_HEADER]) {
+      config.headers[REQUEST_ID_HEADER] = uuidv4();
+    }
+    return config;
+  });
+};
+
+const getRequestIdFromError = (err) => {
+  return (
+    err?.response?.headers?.[REQUEST_ID_HEADER.toLowerCase()] ??
```
vishnuszipstack, line 25: **[P2] Header-name casing asymmetry — correct but fragile.** The response lookup lowercases the key (`x-request-id`, because axios normalizes response headers) while the `config` fallback uses canonical `X-Request-ID`. This works, but a future maintainer could "fix" the inconsistency and silently break the fallback. Either add a one-line comment explaining why the casing differs, or make it deterministic: stash the generated UUID on the request config in the interceptor (e.g. `config.metadata = { requestId }`) and read that back here — immune to header casing entirely.

Optional: when `err.response` exists but no ID is found, a `console.warn` would turn a future silent regression (CORS not exposing the header, backend not setting it) into a debuggable breadcrumb.

frontend/src/App.jsx
```diff
@@ -1,17 +1,21 @@
...
 import { PageTitle } from "./components/widgets/page-title/PageTitle.jsx";
 import { THEME } from "./helpers/GetStaticData.js";
+import { attachRequestIdInterceptor } from "./helpers/requestId.js";
 import PostHogPageviewTracker from "./PostHogPageviewTracker.js";
 import { Router } from "./routes/Router.jsx";
 import { useAlertStore } from "./store/alert-store.js";
 import { useSessionStore } from "./store/session-store.js";
 import { useSocketLogsStore } from "./store/socket-logs-store.js";
 
+export const globalRequestIdInterceptor = attachRequestIdInterceptor(axios);
```
vishnuszipstack: **[P2] Module-level side effect at import time; interceptor never ejected.** `attachRequestIdInterceptor(axios)` runs on module load, and the exported `globalRequestIdInterceptor` handle is never imported anywhere. Under Vite HMR, re-importing this module stacks a fresh interceptor on the global `axios` on every hot reload. Either move the attach into the `App` component body in a `useEffect` with `axios.interceptors.request.eject(handle)` cleanup, or keep it module-level but guard it so it attaches exactly once and drop the unused export (or add a comment documenting why it is exported).

frontend/src/hooks/useExceptionHandler.jsx
```diff
@@ -1,14 +1,17 @@
 import PropTypes from "prop-types";
 import { useNavigate } from "react-router-dom";
 
+import { getRequestIdFromError } from "../helpers/requestId";
+
 const useExceptionHandler = () => {
   const navigate = useNavigate();
 
-  const buildAlert = (content, title, duration) => ({
+  const buildAlert = (content, title, duration, requestId) => ({
```
vishnuszipstack: **[P2] `buildAlert` now takes 4 positional args** `(content, title, duration, requestId)` — `content`/`title` are both strings and easy to transpose, and `requestId` is threaded identically through ~9 call sites. Consider an options object `buildAlert({ content, title, duration, requestId })`, or — since `title`, `duration`, `requestId` are constant within a single `handleException` invocation — a local closure `const alert = (content) => buildAlert(content, title, duration, requestId)` so each site collapses to `alert(msg)`.

Two follow-ups:
1. `requestId` defaults to `undefined` here but `alert-store.js` defaults it to `null` — align the two for a consistent alert shape.
2. The `useExceptionHandler.propTypes` block (lines 111-114) is stale: it omits `title`, `duration`, and the new `requestId`, and it documents `handleException`'s params rather than the hook component's. Update or remove it.

frontend/src/index.css
```diff
@@ -251,3 +251,21 @@ body {
...
+.notification-request-id {
+  margin-top: 8px;
+  font-size: 12px;
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
+  gap: 4px;
+}
+
+.notification-request-id__label {
```
vishnuszipstack: **[P3] `font-size: 12px` is repeated in all three rules.** Consolidate:
```css
.notification-request-id,
.notification-request-id__label,
.notification-request-id__value { font-size: 12px; }
```
After that, `.notification-request-id__label` has no remaining properties and can be removed entirely; `.notification-request-id__value` keeps only `word-break: break-all`.

frontend/src/App.jsx
```diff
@@ -53,20 +57,49 @@ function App() {
       return;
     }
 
+    const showRequestId =
+      alertDetails?.type === "error" && alertDetails?.requestId;
+    const description = (
```
vishnuszipstack, line 66: **[P3] Inline description JSX could be its own component.** The ~20-line `showRequestId` block is sizeable for an inline `useEffect`; extracting a small `NotificationDescription` component would keep the effect focused. Also, inside the `{showRequestId && (...)}` block the `alertDetails?.content` / `alertDetails?.requestId` optional chaining is redundant — `showRequestId` already guarantees `alertDetails` exists and `requestId` is truthy.

vishnuszipstack (commented):

Ran Comment Analyzer, PR Test Analyzer, Silent Failure Hunter, Type Design Analyzer, Code Reviewer, and Code Simplifier against `git diff main...HEAD`. Overall the feature is well-structured and the helpers are cleanly extracted as pure functions. Findings below are posted inline, prioritized:

**P1 (address before merge)**
- `requestId.js:7` — interceptor uses bracket access on `config.headers` (an `AxiosHeaders` instance in axios 1.x) and assumes it is defined.

**P2 (recommended)**
- `requestId.js:5` — no tests added; the pure helpers are trivially testable (vitest is configured).
- `requestId.js:16` — response/config header-casing asymmetry is correct but fragile.
- `App.jsx:17` — module-level side effect at import time; interceptor never ejected; unused export.
- `useExceptionHandler.jsx:9` — `buildAlert` now takes 4 positional args; `requestId` default (`undefined`) diverges from `alert-store.js` (`null`); `propTypes` block is stale.

**P3 (nice-to-have)**
- `index.css:264` — `font-size: 12px` duplicated across all three rules.
- `App.jsx:62` — inline description JSX could be extracted; redundant optional chaining inside the `showRequestId` guard.

Note: the Comment Analyzer found no comment-accuracy issues (the PR adds no code comments).

frontend/src/helpers/requestId.js
Deepak-Kesavan: Done in `a07d5487c`. Switched to `headers.set(name, value, false)` so the case-insensitive AxiosHeaders normalization is respected and caller-supplied IDs are preserved (e.g. tool-sandbox → runner). Bracket-access fallback for plain-object headers, plus a `config.headers ??= {}` guard for hand-built configs.

frontend/src/helpers/requestId.js
Deepak-Kesavan: Added in `a07d5487c` — `frontend/src/helpers/requestId.test.js` covers: injection on missing header, no-overwrite on caller-supplied ID, distinct ID per request, headers-object created when absent, response → config fallback ordering, and null/empty inputs returning `undefined` without throwing.

frontend/src/helpers/requestId.js
Deepak-Kesavan: Skipping the inline explanatory comment (team convention is to avoid narrative comments in code). The asymmetry is now locked down by the new tests — `reads the lowercased response header` and `falls back to the request config header when no response header exists` — so a future "cleanup" pass that collapses the casing would fail those tests rather than silently regress. Open to switching to the `config.metadata` stash if you'd prefer the deterministic version.

frontend/src/App.jsx
Deepak-Kesavan: Done in `a07d5487c`. Replaced the eager attach with a `Symbol.for("unstract.requestIdInterceptor")` idempotency guard so HMR re-evals don't stack a fresh interceptor on the global axios. Dropped the unused exported handle.

frontend/src/hooks/useExceptionHandler.jsx
Deepak-Kesavan: Done in `a07d5487c`. Collapsed `buildAlert` into a local `alert(content)` closure inside `handleException` — each return site is now a single positional arg. Defaulted `requestId` to `null` to match the `alert-store` shape. Dropped the stale function-shaped `propTypes` block entirely (it was documenting `handleException` params, not a component).

frontend/src/index.css
Deepak-Kesavan: Done in `a07d5487c`. Consolidated `font-size: 12px` onto a shared selector and removed the now-empty `.notification-request-id__label` rule plus its className from the JSX.

frontend/src/App.jsx
Deepak-Kesavan: Going to leave this as-is. A ~15-line inline JSX block used in exactly one place (a single `useEffect`) doesn't really earn its own component — the indirection would obscure more than it clarifies. On the optional chaining: the rest of this file uses `alertDetails?.` defensively throughout, so dropping it only inside the `showRequestId &&` branch would create an internal inconsistency. Happy to revisit if `NotificationDescription` ends up needed elsewhere.

vishnuszipstack:
@Deepak-Kesavan approving. check the sonar issues.

vishnuszipstack (approved):

PR #1929, opened 2026-04-26
UN-2946 [FEAT] Prompt Studio lookups bridge, executor hook, and IDE wiring (OSS side)

chandrasekharan-zipstack:
## What

OSS-side glue for **Lookups V2** — a cloud-only Prompt Studio feature that runs an LLM-powered post-extraction enrichment over a prompt's structured output. This PR is the OSS half: a try-import bridge module, executor hook for non-SP runs, IDE wiring for the Raw/Enriched tabs and gating modals, and a few small schema extensions.

The cloud-side PR (where the feature actually lives) is on the `unstract-cloud` repo, same branch name, same ticket: <https://github.com/Zipstack/unstract-cloud/pull/1463>.

## Why

The cloud plugin needs to integrate with extraction (executor), with the per-prompt output flow (`prompt_studio_helper`, `prompt_studio_output_manager_v2`), with export (`prompt_studio_registry_helper`), and with the IDE UI (Raw/Enriched tabs, gate modal, dirty-seed banner). All of that has to live in OSS — but **OSS must never import cloud schema directly**. This PR is that boundary.

## How

- **One bridge module** — `backend/prompt_studio/lookup_utils.py` try-imports the cloud plugin and exposes ~10 stable, opaque functions. Every function is a no-op in OSS. Callers don't know cloud key names or model shapes.
- **Executor hook** — `legacy_executor._run_lookup_enrichment` resolves the worker plugin via `ExecutorPluginLoader.get("lookup-enrichment")` and runs after each prompt's extraction. Skips with a workflow-log entry when the source prompt extracted nothing.
- **Pre-flight blocks** — multi-var lookups in non-SP mode return a 400 from `fetch_response` / `bulk_fetch_response`, scoped to the prompt set being run.
- **Export gate** — `lookup-validation` action on `CustomToolView` returns the validation buckets that the FE modal renders.
- **Staleness signal** — `CustomTool.last_exported_at` (new field) plus `is_lookup_dirty` in `check_deployment_usage` powers the "needs re-export" banner.
- **Usage extension** — `Usage` model adds a few cloud-extensible enum choices (`reference_type`, `llm_usage_reason="lookup"`, `status`, `error_message`, `execution_time_ms`).

## Reviewer aid

A condensed one-page primer for human reviewers — system shape, data model, execution diagrams, gates, smoke checklist — is at `lookups-v2-reviewer-refresher.html` (attached, same file on both PRs). Read it before opening the diff.

[...]

PR #1962, opened 2026-05-13
UN-2946 [MISC] Flush embedding usage records on indexing path

chandrasekharan-zipstack:
## What

- Embedding-callback usage rows are now drained and persisted for every workflow / API-deployment indexing run.
- `EmbeddingCompat` gets a `flush_pending_usage()` shim mirroring the LLM one.
- `_handle_index` returns its embedding rows via `ExecutionResult.metadata["usage_records"]`; `_run_pipeline_index` propagates them up to `_handle_structure_pipeline` to be absorbed into `pipeline_records`.
- `LegacyExecutorError.partial_usage_records` carries embedding rows across mid-pipeline failures.

## Why

The deferred-batch usage refactor (UN-2946) replaced `Audit.push_usage_data()` direct push with `UsageHandler._pending_usage` + a `flush_pending_usage()` drain pattern. The drain was wired for LLM (`_handle_answer_prompt`, `_handle_summarize`, IDE-index) but never for the workflow indexing path. As a result, embedding rows were appended to an in-memory list that died with the local `embedding` variable when `_handle_index` returned — never reaching the `Usage` table.

Visible symptom in API-deployment responses post-merge of `feat/lookups-v2` to main on 2026-05-11:

```json
"usage": {
  "embedding_tokens": 0,
  "prompt_tokens": 4308,
  "completion_tokens": 447,
  ...
}
```

Staging `unstract_core.usage` confirms the cliff: last `usage_type='embedding'` row at 2026-05-11T06:34Z; zero embedding rows since. LLM rows continue normally.

## How

1. `EmbeddingCompat.flush_pending_usage()` — walks `callback_manager.handlers` and drains anything with a `flush_pending_usage` method (matches the LLM shim shape and the iterator pattern already used in `_handle_answer_prompt`).
2. `_handle_index` — at each success exit (`chunk_size==0` early return is unchanged since no embedding is instantiated; document-already-indexed skip; post-`perform_indexing` return) it now passes `metadata={"usage_records": embedding.flush_pending_usage()}` on the `ExecutionResult`. The exception arm flushes any partial records into `LegacyExecutorError.partial_usage_records`.
3. `_run_pipeline_index` — return type changes from `dict` to `tuple[dict, list[dict]]`; the per-output loop accumulates `index_result.metadata["usage_records"]` into `index_records`.
4. `_handle_structure_pipeline` — unpacks the tuple and extends `pipeline_records`. IDE-index path (`_handle_ide_index`) already absorbs `metadata["usage_records"]` from its child `_handle_index` call, so it starts working without changes.
5. `workers/tests/test_phase5d.py` — mock updated to return the new tuple shape.

- `_run_pipeline_index` return type changed (`dict` → `tuple[dict, list[dict]]`). Only one caller (`_handle_structure_pipeline`) and one test mock; both updated.
- `_handle_index` now always returns `metadata={"usage_records": [...]}`. Existing callers that read `metadata["usage_records"]` (`_handle_ide_index`, `_absorb`) already handle the key being present or absent. The early `chunk_size==0` return is left without metadata because no embedding is instantiated — matches the pre-refactor behaviour.
- No SDK / public-API surface change; `EmbeddingCompat.flush_pending_usage()` is additive.

## Database Migrations

- None.

## Env Config

- None.

## Relevant Docs

- None.

## Related Issues or PRs

- Regression introduced by #1929 (UN-2946 lookups V2 squash merge on 2026-05-11).
- Origin of the deferred-batch pattern: commit `8aa675884` (UN-2946 [FEAT] Deferred batch usage tracking with operation metrics, 2026-04-08) on `feat/lookups-v2`.

## Dependencies Versions

- None.

## Notes on Testing

- Affected pytest suites pass locally: `test_phase5d`, `test_context_retrieval_metrics`, `test_legacy_executor_index`, `test_sanity_phase4/5/6g` — 93/93.
- Manual verification path: run an API-deployment workflow whose tool uses indexing (chunk_size > 0); confirm the response's `metadata.usage.embedding_tokens > 0` and that `Usage.objects.filter(run_id=<file_execution_id>, usage_type='embedding')` returns rows.
- Edge: documents with chunk_size==0 still return early without usage rows (no embedding generated) — unchanged.

## Checklist

### DIFF
--- a/unstract/sdk1/src/unstract/sdk1/embedding.py
+++ b/unstract/sdk1/src/unstract/sdk1/embedding.py
@@ -281,3 +281,13 @@ class EmbeddingCompat(BaseEmbedding):
 
     def test_connection(self) -> bool:
         return self._embedding_instance.test_connection()
+
+    def flush_pending_usage(self) -> list[dict]:
+        """Drain pending usage rows from registered callback handlers."""
+        if not self.callback_manager:
+            return []
+        records: list[dict] = []
+        for handler in self.callback_manager.handlers:
+            if hasattr(handler, "flush_pending_usage"):
+                records.extend(handler.flush_pending_usage())
+        return records
--- a/unstract/sdk1/src/unstract/sdk1/embedding.py
+++ b/unstract/sdk1/src/unstract/sdk1/embedding.py
@@ -288,6 +288,15 @@ class EmbeddingCompat(BaseEmbedding):
             return []
         records: list[dict] = []
         for handler in self.callback_manager.handlers:
-            if hasattr(handler, "flush_pending_usage"):
+            if not hasattr(handler, "flush_pending_usage"):
+                continue
+            # Per-handler guard so one bad handler doesn't drop the rest.
+            try:
                 records.extend(handler.flush_pending_usage())
+            except Exception:
+                logger.warning(
+                    "Failed to flush usage from embedding handler %s",
+                    type(handler).__name__,
+                    exc_info=True,
+                )
         return records
--- a/workers/executor/executors/legacy_executor.py
+++ b/workers/executor/executors/legacy_executor.py
@@ -974,7 +974,12 @@ class LegacyExecutor(BaseExecutor):
                         "usage_kwargs": usage_kwargs,
                     },
                 )
-                index_result = self._handle_index(index_ctx)
+                try:
+                    index_result = self._handle_index(index_ctx)
+                except LegacyExecutorError as e:
+                    # Preserve usage rows accrued from prior iterations.
+                    e.partial_usage_records = index_records + e.partial_usage_records
+                    raise
                 if not index_result.success:
                     logger.warning(
                         "Pipeline indexing failed for %s: %s",
--- a/workers/executor/executors/legacy_executor.py
+++ b/workers/executor/executors/legacy_executor.py
@@ -1005,10 +1005,12 @@ class LegacyExecutor(BaseExecutor):
             e.partial_usage_records = index_records + e.partial_usage_records
             raise
         if not index_result.success:
-            logger.warning(
-                "Pipeline indexing failed for %s: %s",
-                param_key,
-                index_result.error,
+            # Abort on returned-failure so downstream steps don't run
+            # against an incomplete vector store.
+            raise LegacyExecutorError(
+                message=f"Pipeline indexing failed for {param_key}: {index_result.error}",
+                code=500,
+                partial_usage_records=list(index_records),
             )
         child_records = (index_result.metadata or {}).get("usage_records") or []
         if child_records:
--- a/workers/executor/executors/legacy_executor.py
+++ b/workers/executor/executors/legacy_executor.py
@@ -907,94 +907,116 @@ class LegacyExecutor(BaseExecutor):
             name and the flat list of usage rows collected across all
             child ``_handle_index`` calls.
         """
-        import datetime
-
         tool_settings = answer_params.get("tool_settings", {})
         outputs = answer_params.get("outputs", [])
-        tool_id = index_template.get("tool_id", "")
-        file_hash = index_template.get("file_hash", "")
-        is_highlight = index_template.get("is_highlight_enabled", False)
-        platform_api_key = index_template.get("platform_api_key", "")
-        extracted_file_path = index_template.get("extracted_file_path", "")
-        usage_kwargs = usage_kwargs or {}
-
         index_metrics: dict = {}
         index_records: list[dict] = []
         seen_params: set = set()
 
         for output in outputs:
-            chunk_size = output.get("chunk-size", 0)
-            chunk_overlap = output.get("chunk-overlap", 0)
-            vector_db = tool_settings.get("vector-db", "")
-            embedding = tool_settings.get("embedding", "")
-            x2text = tool_settings.get("x2text_adapter", "")
-
-            param_key = (
-                f"chunk_size={chunk_size}_"
-                f"chunk_overlap={chunk_overlap}_"
-                f"vector_db={vector_db}_"
-                f"embedding={embedding}_"
-                f"x2text={x2text}"
+            self._index_pipeline_output(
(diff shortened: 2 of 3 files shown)


vishnuszipstack:
@chandrasekharan-zipstack check the sonar issues.

vishnuszipstack (approved):

unstract/sdk1/src/unstract/sdk1/embedding.py
```diff
@@ -281,3 +281,13 @@ async def get_aquery_embedding(self, query: str) -> list[float]:
...
         return self._embedding_instance.test_connection()
+
+    def flush_pending_usage(self) -> list[dict]:
+        """Drain pending usage rows from registered callback handlers."""
+        if not self.callback_manager:
+            return []
+        records: list[dict] = []
+        for handler in self.callback_manager.handlers:
+            if hasattr(handler, "flush_pending_usage"):
+                records.extend(handler.flush_pending_usage())
```
athul-rs, line 295: **High — missing per-handler exception guard.**

The PR description says this matches the iterator pattern in `_handle_answer_prompt`, but the existing pattern at `workers/executor/executors/legacy_executor.py:1971-1983` explicitly wraps each handler call in `try/except` with the comment `# Per-handler guard so one bad handler doesn't drop the rest.` This shim does not.

If any one handler `flush_pending_usage()` raises, the exception escapes here, the local `records` list is **never returned**, and every record queued from handlers iterated before the bad handler is lost. Combined with the inline use at the `_handle_index` success-path returns (lines 1206, 1227 — see other comments), this turns a successful index into a `LegacyExecutorError` — re-introducing the same shape of bug this PR is fixing.

Suggested fix (also lets the duplicated loop at `legacy_executor.py:1971-1983` collapse to a single `embedding.flush_pending_usage()` call in a follow-up):

```python
def flush_pending_usage(self) -> list[dict]:
    if not self.callback_manager:
        return []
    records: list[dict] = []
    for handler in self.callback_manager.handlers:
        if not hasattr(handler, "flush_pending_usage"):
            continue
        try:
            records.extend(handler.flush_pending_usage())
        except Exception:
            logger.warning(
                "Failed to flush usage from embedding handler %s",
                type(handler).__name__, exc_info=True,
            )
    return records
```

workers/executor/executors/legacy_executor.py
```diff
@@ -1152,7 +1200,11 @@ def _handle_index(self, context: ExecutionContext) -> ExecutionResult:
                     "Skipping re-index: doc_id=%s already in vector DB and reindex=False",
                     doc_id,
                 )
-                return ExecutionResult(success=True, data={IKeys.DOC_ID: doc_id})
+                return ExecutionResult(
+                    success=True,
+                    data={IKeys.DOC_ID: doc_id},
+                    metadata={"usage_records": embedding.flush_pending_usage()},
```
athul-rs, line 1208: **High — fallout from the missing per-handler guard in `EmbeddingCompat.flush_pending_usage()`.**

This call is inline inside the success-path `ExecutionResult(...)`. If a callback handler raises during flush, the exception escapes this success branch, gets caught by `except Exception as e:` at line 1229, and is re-emitted as `LegacyExecutorError("Error while indexing: ...")`. Net effect: a `doc_id_found and not reindex` skip that genuinely succeeded is reported to the caller as a hard indexing failure. The retry at line 1239 hits the same broken handler and the inner `except` swallows it, so `partial` stays `[]` and the records that *did* exist are dropped.

Resolved once `EmbeddingCompat.flush_pending_usage()` gets the per-handler guard. Otherwise, wrap these two success-path flushes in `try/except` here as well.

workers/executor/executors/legacy_executor.py
```diff
@@ -1169,16 +1221,31 @@ def _handle_index(self, context: ExecutionContext) -> ExecutionResult:
                 Path(file_path).name,
             )
             shim.stream_log("Document indexing completed")
-            return ExecutionResult(success=True, data={IKeys.DOC_ID: doc_id})
+            return ExecutionResult(
+                success=True,
+                data={IKeys.DOC_ID: doc_id},
+                metadata={"usage_records": embedding.flush_pending_usage()},
```
athul-rs, line 1229: **High — same root cause as the comment on line 1206.**

Same inline call, same failure mode: a broken embedding handler turns a successful `perform_indexing()` into a `LegacyExecutorError`. Will be resolved by adding the per-handler `try/except` to `EmbeddingCompat.flush_pending_usage()` in `unstract/sdk1/src/unstract/sdk1/embedding.py`.

workers/tests/test_phase5d.py
```diff
@@ -283,9 +283,10 @@ def test_index_metrics_merged(self, executor):
         )
         # Simulate index metrics by patching _run_pipeline_index
         executor._run_pipeline_index = MagicMock(
-            return_value={
-                "field_a": {"indexing": {"time_taken(s)": 0.5}},
-            }
+            return_value=(
+                {"field_a": {"indexing": {"time_taken(s)": 0.5}}},
+                [],
+            )
```
athul-rs, line 289: **Medium — missing regression test for the actual bug being fixed.**

This change only updates the mock to the new tuple shape; nothing asserts that embedding `usage_records` propagate through `_handle_index → _index_pipeline_output → _run_pipeline_index → _handle_structure_pipeline` and end up in `pipeline_records`. The PR body leans on manual API-deployment verification.

Given this is a regression fix for UN-2946 (`Usage` table cliff on 2026-05-11), a unit test that:

1. mocks `_handle_index` to return `ExecutionResult(success=True, metadata={"usage_records": [{...}]})`, and
2. asserts the resulting `pipeline_records` (or `ExecutionResult.metadata["usage_records"]` at the structure-pipeline boundary) contains that row

…would prevent the next refactor from silently re-breaking the same path. The exception-path propagation through `LegacyExecutorError.partial_usage_records` deserves a sibling test.

athul-rs (approved):

PR #1499, opened 2025-08-22
UN-2239 [FEAT] Add trial statistics API endpoint for organization usage tracking

jaags-dev:
## What
Added a new API endpoint `/api/v2/usage/trial-statistics/` that retrieves comprehensive trial usage statistics for organizations. This endpoint provides key metrics about an organization's trial period usage including costs, document processing counts, API calls, and ETL pipeline runs.

## Why
This endpoint is needed to provide organizations with insights into their trial period usage patterns. It enables tracking of resource consumption, cost savings, and usage metrics during the trial phase, which is crucial for understanding customer engagement and conversion potential.

## How
- **UsageHelper Enhancement**: Added `get_trial_statistics()` static method that:
  - Attempts to fetch subscription data from the subscription plugin if available
  - Falls back to organization creation date + 14 days for trial period calculation
  - Aggregates usage data using Django ORM for efficient queries
  - Calculates total cost, document processing counts, API calls, and unique ETL runs
  - Includes comprehensive error handling with fallback responses

- **View Implementation**: Added `get_trial_statistics` action in UsageView that:
  - Retrieves the current organization context
  - Calls the helper method to get statistics
  - Returns properly formatted JSON response with error handling

- **URL Configuration**: Registered new `/trial-statistics/` endpoint in the usage_v2 URL patterns

## Relevant Docs
- Endpoint: `GET /api/v2/usage/trial-statistics/`
- Response format:
  ```json
  {
    "trial_start_date": "ISO date string",
    "trial_end_date": "ISO date string",
    "total_cost": 0.0,
    "documents_processed": 0,
    "api_calls": 0,
    "etl_runs": 0
  }
  ```

## Related Issues/PRs
- Related to UN-2239 ticket for trial statistics tracking
- Builds upon existing usage_v2 infrastructure

## Dependencies Versions
No new dependencies added. Uses existing Django ORM and subscription plugin (optional).

### Manual Testing Scenarios:
1. **With Subscription Plugin:**
   - Verify endpoint returns subscription-based trial dates
   - Confirm accurate cost and usage aggregations

2. **Without Subscription Plugin (Fallback):**
   - Verify fallback to organization creation date + 14 days
   - Ensure no errors when subscription plugin is unavailable

3. **Edge Cases:**
   - Test with organizations having no usage data
   - Test with organizations having partial data
   - Verify error handling returns appropriate fallback response

### Test Commands:
```bash
# Test the endpoint
[...]

PR #1966, opened 2026-05-14
UN-3393 [FEAT] Wire AuditSerializer subclasses through SanitizedSerializerMixin

chandrasekharan-zipstack:
## What

- Builds on #1965 (foundation PR) by routing every `AuditSerializer` subclass through `SanitizedSerializerMixin`. ~18 write-path serializers now auto-reject HTML/JS-shaped input on every writable `CharField` / `TextField`.
- Affected (transitively): `WorkflowSerializer`, `CustomToolSerializer`, `APIDeploymentSerializer`, `APIKeySerializer`, `ConnectorInstanceSerializer`, `BaseAdapterSerializer`, `PlatformKeySerializer`, `PlatformApiKey{Create,Update}Serializer`, `PipelineSerializer`, `ToolInstanceSerializer`, `ToolStudioPromptSerializer`, `PromptStudioOutputSerializer`, `ProfileManagerSerializer`, `IndexManagerSerializer`, `PromptStudioRegistry{,Info}Serializer`, `PromptStudioDocumentManagerSerializer`.
- Adds `Meta.html_safe_fields` opt-outs on the two serializers that legitimately carry LLM markup:
  - `ToolStudioPromptSerializer` → `prompt`, `assert_prompt`, `assertion_failure_prompt`, `output`.
  - `CustomToolSerializer` → `summarize_prompt`, `preamble`, `postamble`, `output`.
- Removes the three redundant `validate_description` methods (in `api_v2.APIDeploymentSerializer`, `workflow_v2.WorkflowSerializer`, `prompt_studio_core_v2.CustomToolSerializer`) — the mixin now covers them.

## Why

- The foundation PR (#1965) added `SanitizedSerializerMixin` and pre-mixed `ModelSerializer` / `Serializer` / `HyperlinkedModelSerializer` under `utils.serializer` but did not wire any production serializer through them. This PR is the smallest, highest-leverage wiring step: changing **one** parent class (`AuditSerializer`) closes the input-validation gap on the majority of write-path entities.
- `validate_<name>` methods (which use `validate_name_field`) are kept because they also strip whitespace and reject empty values — behaviour the mixin doesn't duplicate. Only the `validate_description` methods (which just call `validate_no_html_tags`) are removed.
- Free-form LLM prompt / output fields (`prompt`, `output`, `summarize_prompt`, etc.) legitimately contain `<context>` / `<thinking>` / arbitrary model output. They're explicitly opted out via `Meta.html_safe_fields`.

## How

- `backend/backend/serializers.py`: `AuditSerializer` now inherits `utils.serializer.ModelSerializer` instead of `rest_framework.serializers.ModelSerializer`. `create` / `update` semantics unchanged.
- Per-serializer `Meta.html_safe_fields` declarations on the two LLM-bearing serializers (Prompt Studio).
- Three `validate_description` removals + `validate_no_html_tags` import drop from the affected modules.

- **Functional:** The 8 serializers that were already hand-wired (in the in-flight fix) continue to behave identically. The 14 manual `validate_<field>` methods still exist — the mixin's validator is appended to the field's `validators` list, so the same rejection is enforced either way. Removing only the redundant `validate_description` methods avoids double-validation noise but doesn't change any acceptance outcome.
- **New rejections:** `AuditSerializer` subclasses that were *not* previously sanitised now auto-reject HTML in their `CharField` / `TextField` (e.g. `ConnectorInstance.connector_name`, `PlatformKey.key_name`, `Pipeline.pipeline_name`, `AdapterInstance.adapter_name`, etc.). Per the May 2026 US-prod scan (see KB `Obsidian Vault/zipstuff/UN-3393-input-validation/05-prod-baseline.md`), zero legitimate user content uses `<…>` in these columns — every match was pentest leftover. Migration risk: ~zero.
- **LLM fields explicitly opted out:** `prompt`, `assert_prompt`, `assertion_failure_prompt`, `output`, `summarize_prompt`, `preamble`, `postamble` are listed in `Meta.html_safe_fields` so they continue to accept arbitrary text.
- **Direct DRF-base-class users:** Serializers that inherit directly from `serializers.ModelSerializer` / `serializers.Serializer` / `serializers.HyperlinkedModelSerializer` (i.e. not via `AuditSerializer`) are **not** touched by this PR. They remain unchanged. A follow-up sweep will convert them; their absence here means write paths through them retain pre-PR behaviour.

## Database Migrations

- None.

## Env Config

- None.

## Relevant Docs

- KB: `Obsidian Vault/zipstuff/UN-3393-input-validation/`.
- ADR 0003 (`adr/0003-base-mixin-default-on-opt-out.md`) — opt-out semantics.

## Related Issues or PRs

- Jira: UN-3393.
- Stacked on: #1965 (foundation).
- Follow-up: a separate PR that converts the remaining ~20 direct DRF-base-class users (`tags/serializers.py`, `scheduler/serializer.py`, `tenant_account_v2/serializer.py`, `pipeline_v2/serializers/{internal,update,sharing,execute}.py`, response-only and list serializers across the codebase, …).
- Follow-up: cover `backend/pluggable_apps/` (lives in the `unstract-cloud` repo; separate PR there).

## Dependencies Versions

- None.

## Notes on Testing

Local tests in the OSS worktree (cloud-deps not installed, so `manage.py check` is unavailable here — CI runs the full suite):

```bash
cd backend && uv run pytest utils/tests/ -q
```

→ 85 passed (50 from foundation + 35 pre-existing utils tests).

Manual sanity:

```bash
cd backend && uv run python -c "from utils.serializer import ModelSerializer; from backend.serializers import AuditSerializer; assert issubclass(AuditSerializer, ModelSerializer); print('ok')"
```

## Screenshots

n/a (no UI change)

## Checklist

### DIFF
--- a/backend/backend/serializers.py
+++ b/backend/backend/serializers.py
@@ -1,6 +1,6 @@
 from typing import Any
 
-from rest_framework.serializers import ModelSerializer
+from utils.serializer import ModelSerializer
 
 from backend.constants import RequestKey
 
--- a/frontend/src/components/custom-tools/document-parser/DocumentParser.jsx
+++ b/frontend/src/components/custom-tools/document-parser/DocumentParser.jsx
@@ -205,6 +205,16 @@ function DocumentParser({
         return res;
       })
       .catch((err) => {
+        // Field-keyed DRF validation errors are surfaced inline on the
+        // prompt card; re-throw so the card's catch can render them.
+        const data = err?.response?.data;
+        const hasFieldError =
+          data?.type === "validation_error" &&
+          Array.isArray(data?.errors) &&
+          data.errors.some((e) => e?.attr);
+        if (hasFieldError) {
+          throw err;
+        }
         setAlertDetails(handleException(err, "Failed to update"));
       });
   };
--- a/frontend/src/components/custom-tools/document-parser/DocumentParser.jsx
+++ b/frontend/src/components/custom-tools/document-parser/DocumentParser.jsx
@@ -37,6 +37,19 @@ try {
 const UUID_RE =
   /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
 
+const getFieldErrors = (err) => {
+  const data = err?.response?.data;
+  if (data?.type !== "validation_error" || !Array.isArray(data?.errors)) {
+    return {};
+  }
+  return data.errors.reduce((acc, e) => {
+    if (e?.attr) {
+      acc[e.attr] = e.detail || "Invalid value";
+    }
+    return acc;
+  }, {});
+};
+
 function DocumentParser({
   addPromptInstance,
   scrollToBottom,
@@ -140,7 +153,12 @@ function DocumentParser({
     return `/api/v1/unstract/${sessionDetails?.orgId}/prompt-studio/prompt/${urlPath}`;
   };
 
-  const handleChangePromptCard = async (name, value, promptId) => {
+  const handleChangePromptCard = async (
+    name,
+    value,
+    promptId,
+    onFieldError,
+  ) => {
     const promptsAndNotes = details?.prompts || [];
 
     if (name === "prompt_key") {
@@ -205,17 +223,16 @@ function DocumentParser({
         return res;
       })
       .catch((err) => {
-        // Field-keyed DRF validation errors are surfaced inline on the
-        // prompt card; re-throw so the card's catch can render them.
-        const data = err?.response?.data;
-        const hasFieldError =
-          data?.type === "validation_error" &&
-          Array.isArray(data?.errors) &&
-          data.errors.some((e) => e?.attr);
-        if (hasFieldError) {
-          throw err;
+        // Inline rendering of field errors is opt-in: a caller that has no
+        // renderer for the offending field returns false and gets the alert,
+        // so a rejected value can never fail silently.
+        const fieldErrors = getFieldErrors(err);
+        const handledInline =
+          Object.keys(fieldErrors).length > 0 && onFieldError?.(fieldErrors);
+        if (!handledInline) {
+          setAlertDetails(handleException(err, "Failed to update"));
         }
-        setAlertDetails(handleException(err, "Failed to update"));
+        throw err;
       });
   };
 
--- a/frontend/src/components/custom-tools/document-parser/DocumentParser.jsx
+++ b/frontend/src/components/custom-tools/document-parser/DocumentParser.jsx
@@ -223,12 +223,10 @@ function DocumentParser({
         return res;
       })
       .catch((err) => {
-        // Inline rendering of field errors is opt-in: a caller that has no
-        // renderer for the offending field returns false and gets the alert,
-        // so a rejected value can never fail silently.
-        const fieldErrors = getFieldErrors(err);
-        const handledInline =
-          Object.keys(fieldErrors).length > 0 && onFieldError?.(fieldErrors);
+        // Reporting the rejection is opt-in: a caller that will not report it
+        // itself returns false and gets the alert, so a rejected value can
+        // never fail silently.
+        const handledInline = onFieldError?.(getFieldErrors(err));
         if (!handledInline) {
           setAlertDetails(handleException(err, "Failed to update"));
         }
--- a/frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
+++ b/frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
@@ -53,6 +53,7 @@ const PromptCard = memo(
     const [promptKey, setPromptKey] = useState("");
     const [promptText, setPromptText] = useState("");
     const [selectedLlmProfileId, setSelectedLlmProfileId] = useState(null);
+    const [fieldErrors, setFieldErrors] = useState({});
 
     const [isCoverageLoading, setIsCoverageLoading] = useState(false);
     const [openOutputForDoc, setOpenOutputForDoc] = useState(false);
@@ -157,6 +158,14 @@ const PromptCard = memo(
       const updatedPromptDetailsState = { ...promptDetailsState };
       updatedPromptDetailsState[name] = value;
 
+      // New attempt — drop any prior inline error for this field.
+      setFieldErrors((prev) => {
+        if (!(name in prev)) return prev;
+        const next = { ...prev };
+        delete next[name];
+        return next;
+      });
+
       handleUpdateStatus(
         isUpdateStatus,
         promptId,
@@ -178,9 +187,26 @@ const PromptCard = memo(
             setUpdateStatus,
           );
         })
-        .catch(() => {
+        .catch((err) => {
           handleUpdateStatus(isUpdateStatus, promptId, null, setUpdateStatus);
-          setPromptDetailsState(prevPromptDetailsState);
+          const data = err?.response?.data;
+          const fieldErrorMap = {};
+          if (
+            data?.type === "validation_error" &&
+            Array.isArray(data?.errors)
+          ) {
+            data.errors.forEach((e) => {
+              if (e?.attr) {
+                fieldErrorMap[e.attr] = e.detail || "Invalid value";
+              }
+            });
+          }
+          if (Object.keys(fieldErrorMap).length > 0) {
+            // Keep the typed value so user can fix in place; show inline error.
+            setFieldErrors((prev) => ({ ...prev, ...fieldErrorMap }));
+          } else {
+            setPromptDetailsState(prevPromptDetailsState);
+          }
         })
         .finally(() => {
           if (isUpdateStatus) {
@@ -381,6 +407,7 @@ const PromptCard = memo(
           coverageCountData={coverageCountData}
           isChallenge={isChallenge}
           handleSelectHighlight={handleSelectHighlight}
+          fieldErrors={fieldErrors}
         />
         <OutputForDocModal
           open={openOutputForDoc}
--- a/frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
+++ b/frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
@@ -217,12 +217,15 @@ const PromptCard = memo(
             setUpdateStatus,
           );
         })
-        .catch(() => {
+        .catch((err) => {
           handleUpdateStatus(isUpdateStatus, promptId, null, setUpdateStatus);
           if (!handledInline) {
             // Roll back only the field that failed, for the same reason.
             setPromptDetailsState((prev) => ({ ...prev, [name]: prevValue }));
           }
+          // Callers keeping their own copy of the value (Header's webhook and
+          // toggle state) roll back off this rejection.
+          throw err;
         })
         .finally(() => {
           if (isUpdateStatus) {
@@ -235,11 +238,12 @@ const PromptCard = memo(
 
     const handleSelectDefaultLLM = (llmProfileId) => {
       setSelectedLlmProfileId(llmProfileId);
+      // Error already surfaced by handleChange; no local copy to roll back.
       handleChange(
         llmProfileId,
         promptDetailsState?.prompt_id,
         "profile_manager",
-      );
+      ).catch(() => {});
     };
 
     const addCoordsToFlattened = (coords, flattened) => {
@@ -321,7 +325,12 @@ const PromptCard = memo(
         setAlertDetails({ type: "error", content: block.reason });
         return;
       }
-      handleChange(value, promptDetailsState?.prompt_id, "enforce_type", true);
+      handleChange(
+        value,
+        promptDetailsState?.prompt_id,
+        "enforce_type",
+        true,
+      ).catch(() => {});
     };
 
     const handleSpsLoading = (docId, isLoadingStatus) => {
--- a/frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
+++ b/frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
@@ -53,6 +53,7 @@ const PromptCard = memo(
     const [promptKey, setPromptKey] = useState("");
     const [promptText, setPromptText] = useState("");
     const [selectedLlmProfileId, setSelectedLlmProfileId] = useState(null);
+    const [fieldErrors, setFieldErrors] = useState({});
 
     const [isCoverageLoading, setIsCoverageLoading] = useState(false);
     const [openOutputForDoc, setOpenOutputForDoc] = useState(false);
@@ -165,6 +166,14 @@ const PromptCard = memo(
        */
       const prevValue = promptDetailsState?.[name];
 
+      // New attempt — drop any prior inline error for this field.
+      setFieldErrors((prev) => {
+        if (!(name in prev)) return prev;
+        const next = { ...prev };
+        delete next[name];
+        return next;
+      });
+
       handleUpdateStatus(
         isUpdateStatus,
         promptId,
@@ -186,10 +195,24 @@ const PromptCard = memo(
             setUpdateStatus,
           );
         })
-        .catch(() => {
+        .catch((err) => {
           handleUpdateStatus(isUpdateStatus, promptId, null, setUpdateStatus);
-          // Roll back only the field that failed, for the same reason.
-          setPromptDetailsState((prev) => ({ ...prev, [name]: prevValue }));
+          const data = err?.response?.data;
+          const fieldErrorMap = {};
+          if (data?.type === "validation_error" && Array.isArray(data?.errors)) {
+            data.errors.forEach((e) => {
+              if (e?.attr) {
+                fieldErrorMap[e.attr] = e.detail || "Invalid value";
+              }
+            });
+          }
+          if (Object.keys(fieldErrorMap).length > 0) {
+            // Keep the typed value so user can fix in place; show inline error.
+            setFieldErrors((prev) => ({ ...prev, ...fieldErrorMap }));
+          } else {
+            // Roll back only the field that failed, for the same reason.
+            setPromptDetailsState((prev) => ({ ...prev, [name]: prevValue }));
+          }
         })
         .finally(() => {
           if (isUpdateStatus) {
@@ -390,6 +413,7 @@ const PromptCard = memo(
           coverageCountData={coverageCountData}
           isChallenge={isChallenge}
           handleSelectHighlight={handleSelectHighlight}
+          fieldErrors={fieldErrors}
         />
         <OutputForDocModal
           open={openOutputForDoc}
--- a/frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
+++ b/frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
@@ -1,5 +1,5 @@
 import PropTypes from "prop-types";
-import { memo, useEffect, useState } from "react";
+import { memo, useEffect, useRef, useState } from "react";
 
 import {
   PROMPT_RUN_API_STATUSES,
@@ -58,6 +58,10 @@ const PromptCard = memo(
     const [promptText, setPromptText] = useState("");
     const [selectedLlmProfileId, setSelectedLlmProfileId] = useState(null);
     const [fieldErrors, setFieldErrors] = useState({});
+    // Per-field attempt counter. The key input is debounced, not disabled
+    // while saving, so a slow rejection can land after a newer attempt has
+    // already been accepted; only the latest attempt may touch state.
+    const attemptGen = useRef({});
 
     const [isCoverageLoading, setIsCoverageLoading] = useState(false);
     const [openOutputForDoc, setOpenOutputForDoc] = useState(false);
@@ -170,13 +174,20 @@ const PromptCard = memo(
        */
       const prevValue = promptDetailsState?.[name];
 
+      const generation = (attemptGen.current[name] =
+        (attemptGen.current[name] || 0) + 1);
+      const isLatestAttempt = () => attemptGen.current[name] === generation;
+
+      const clearFieldError = () =>
+        setFieldErrors((prev) => {
+          if (!(name in prev)) return prev;
+          const next = { ...prev };
+          delete next[name];
+          return next;
+        });
+
       // New attempt — drop any prior inline error for this field.
-      setFieldErrors((prev) => {
-        if (!(name in prev)) return prev;
-        const next = { ...prev };
-        delete next[name];
-        return next;
-      });
+      clearFieldError();
 
       handleUpdateStatus(
         isUpdateStatus,
@@ -186,20 +197,19 @@ const PromptCard = memo(
       );
       setPromptDetailsState((prev) => ({ ...prev, [name]: value }));
 
-      let handledInline = false;
-      const showInlineErrors = (fieldErrors) => {
-        const renderable = Object.entries(fieldErrors).filter(([attr]) =>
+      // Collected here but applied in the rejection handler, so the inline
+      // error and the rollback below always land in the same render.
+      let inlineErrors = null;
+      const showInlineErrors = (errors) => {
+        const renderable = Object.entries(errors).filter(([attr]) =>
           RENDERABLE_FIELD_ERRORS.has(attr),
         );
-        if (!renderable.length) {
+        // A superseded attempt cannot render its error, so it must not claim
+        // the field either — the caller falls back to the global alert.
+        if (!renderable.length || !isLatestAttempt()) {
           return false;
         }
-        // Keep the typed value so user can fix in place; show inline error.
-        setFieldErrors((prev) => ({
-          ...prev,
-          ...Object.fromEntries(renderable),
-        }));
-        handledInline = true;
+        inlineErrors = Object.fromEntries(renderable);
         return true;
       };
 
@@ -210,6 +220,9 @@ const PromptCard = memo(
             prev[promptId] = data;
             return prev;
           });
+          if (isLatestAttempt()) {
+            clearFieldError();
+          }
           handleUpdateStatus(
             isUpdateStatus,
             promptId,
@@ -219,9 +232,14 @@ const PromptCard = memo(
         })
         .catch((err) => {
           handleUpdateStatus(isUpdateStatus, promptId, null, setUpdateStatus);
-          if (!handledInline) {
-            // Roll back only the field that failed, for the same reason.
+          if (isLatestAttempt()) {
+            // Roll the field back to the stored value even when the rejected
+            // text stays in the input: outputs, highlights and confidence are
+            // keyed off prompt_key and have to track the server.
             setPromptDetailsState((prev) => ({ ...prev, [name]: prevValue }));
+            if (inlineErrors) {
+              setFieldErrors((prev) => ({ ...prev, ...inlineErrors }));
+            }
           }
           // Callers keeping their own copy of the value (Header's webhook and
           // toggle state) roll back off this rejection.
--- a/frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
+++ b/frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
@@ -27,6 +27,10 @@ try {
 const useEnforceTypeSwitchGate =
   useEnforceTypeSwitchGatePlugin || (() => () => null);
 
+// Fields with an inline error renderer on the card. Anything else falls back
+// to the global alert, so widening the sanitizer cannot silence a rejection.
+const RENDERABLE_FIELD_ERRORS = new Set(["prompt_key"]);
+
 const PromptCard = memo(
   ({
     promptDetails,
@@ -181,7 +185,25 @@ const PromptCard = memo(
         setUpdateStatus,
       );
       setPromptDetailsState((prev) => ({ ...prev, [name]: value }));
-      return handleChangePromptCard(name, value, promptId)
+
+      let handledInline = false;
+      const showInlineErrors = (fieldErrors) => {
+        const renderable = Object.entries(fieldErrors).filter(([attr]) =>
+          RENDERABLE_FIELD_ERRORS.has(attr),
+        );
+        if (!renderable.length) {
+          return false;
+        }
+        // Keep the typed value so user can fix in place; show inline error.
+        setFieldErrors((prev) => ({
+          ...prev,
+          ...Object.fromEntries(renderable),
+        }));
+        handledInline = true;
+        return true;
+      };
+
+      return handleChangePromptCard(name, value, promptId, showInlineErrors)
         .then((res) => {
           const data = res?.data;
           setUpdatedPromptsCopy((prev) => {
@@ -195,21 +217,9 @@ const PromptCard = memo(
             setUpdateStatus,
           );
         })
-        .catch((err) => {
+        .catch(() => {
           handleUpdateStatus(isUpdateStatus, promptId, null, setUpdateStatus);
-          const data = err?.response?.data;
-          const fieldErrorMap = {};
-          if (data?.type === "validation_error" && Array.isArray(data?.errors)) {
-            data.errors.forEach((e) => {
-              if (e?.attr) {
-                fieldErrorMap[e.attr] = e.detail || "Invalid value";
-              }
-            });
-          }
-          if (Object.keys(fieldErrorMap).length > 0) {
-            // Keep the typed value so user can fix in place; show inline error.
-            setFieldErrors((prev) => ({ ...prev, ...fieldErrorMap }));
-          } else {
+          if (!handledInline) {
             // Roll back only the field that failed, for the same reason.
             setPromptDetailsState((prev) => ({ ...prev, [name]: prevValue }));
           }
--- a/frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
+++ b/frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
@@ -200,20 +200,26 @@ const PromptCard = memo(
       // Collected here but applied in the rejection handler, so the inline
       // error and the rollback below always land in the same render.
       let inlineErrors = null;
-      const showInlineErrors = (errors) => {
+      const reportFieldError = (errors) => {
+        // A superseded attempt reports nothing: the newer one owns the field
+        // and surfaces its own outcome, so an alert here would contradict the
+        // value the user can see was accepted.
+        if (!isLatestAttempt()) {
+          return true;
+        }
         const renderable = Object.entries(errors).filter(([attr]) =>
           RENDERABLE_FIELD_ERRORS.has(attr),
         );
-        // A superseded attempt cannot render its error, so it must not claim
-        // the field either — the caller falls back to the global alert.
-        if (!renderable.length || !isLatestAttempt()) {
+        // Nothing renderable — the caller falls back to the global alert so a
+        // rejected value can never fail silently.
+        if (!renderable.length) {
           return false;
         }
         inlineErrors = Object.fromEntries(renderable);
         return true;
       };
 
-      return handleChangePromptCard(name, value, promptId, showInlineErrors)
+      return handleChangePromptCard(name, value, promptId, reportFieldError)
         .then((res) => {
           const data = res?.data;
           setUpdatedPromptsCopy((prev) => {
--- a/backend/api_v2/serializers.py
+++ b/backend/api_v2/serializers.py
@@ -22,7 +22,7 @@ from rest_framework.serializers import (
     ValidationError,
 )
 from tags.serializers import TagParamsSerializer
-from utils.input_sanitizer import validate_name_field, validate_no_html_tags
+from utils.input_sanitizer import validate_name_field
 from utils.serializer.integrity_error_mixin import IntegrityErrorMixin
 from workflow_manager.endpoint_v2.models import WorkflowEndpoint
 from workflow_manager.workflow_v2.exceptions import ExecutionDoesNotExistError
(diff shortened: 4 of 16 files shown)


backend/backend/serializers.py
```diff
@@ -1,6 +1,6 @@
 from typing import Any
 
-from rest_framework.serializers import ModelSerializer
+from utils.serializer import ModelSerializer
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Addressed in commit `6ce8ea860` — added `html_safe_fields = ("output", "context")` to `PromptStudioOutputSerializer.Meta` in `backend/prompt_studio/prompt_studio_output_manager_v2/serializers.py`.

Deepak-Kesavan (commented):
## 🤖 Automated review

3 inline comment(s) · 1 finding(s) not on changed lines (below).

### Findings off the diff

- `backend/adapter_processor_v2/serializers.py:59`
**[minor] The PR lists `BaseAdapterSerializer` as newly covered, but `AdapterInstanceSerializer.to_internal_value` returns the raw dict without calling `super()`, so the mixin's validators never run on the adapter write path — the coverage is nominal, not real.**

`AdapterInstanceSerializer` is what `get_serializer_class` returns for every non-list action (`adapter_processor_v2/views.py:167-172`), and it's what create/update feed request data to. Its `to_internal_value` override (`serializers.py:59-67`) encrypts `adapter_metadata` and then `return data` — it never calls `super().to_internal_value(data)`.

In DRF 3.14.0 (pinned in `backend/uv.lock`), `Serializer.run_validation` calls `to_internal_value(data)` and field-level validators run *exclusively* inside it. So the `validate_no_html_tags` partials this PR attaches are never executed on this path — along with `adapter_id`'s `max_length` and `adapter_type`'s choice check. I confirmed it: a serializer differing only by `def to_internal_value(self, data): return data` accepted `adapter_id="<script>alert(1)</script>"` as valid, while the parent rejected it.

In practice `adapter_name` and `description` are still covered, because `BaseAdapterSerializer.validate()` (`serializers.py:32-44`) hand-checks exactly those two after `to_internal_value` — which is why nothing visibly breaks. The point is the *claim*: this serializer is not sanitized by the mixin, and the two hand-checks are load-bearing, not redundant. If a later cleanup removes them the way this PR removed the three `validate_description` methods, the coverage goes to zero silently.

**This file is not in the diff** (`git diff --name-only` over `backend/adapter_processor_v2/` is empty; last touched in `bf15e8d2e`), so the bypass is pre-existing — raising it only because the PR description asserts coverage this path doesn't have.

**Suggested fix:** Delegate, then encrypt: pop `adapter_metadata`, call `super().to_internal_value(data)`, and set `ADAPTER_METADATA_B` on the validated dict. Once field validators actually run, the `adapter_name` / `description` hand-checks in `BaseAdapterSerializer.validate` become genuinely redundant and can go the same way as the three `validate_description` methods. Alternatively, just drop `BaseAdapterSerializer` from the "affected" list in the PR body.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:52134af5ef8c</sub>

backend/backend/serializers.py
Deepak-Kesavan, line 3: **[critical] Flipping `AuditSerializer` to the sanitizing base turns `validate_no_html_tags` on for the enterprise `ChatTranscript.message` field — which stores the LLM's own answer — so a multi-doc chat turn 400s after the user turn is already committed and the LLM call already billed.**

In `unstract-cloud`, `ChatTranscriptView.create` (`backend/pluggable_apps/apps/chat_transcript/views.py:52-115`) (1) persists the USER turn at line 63, (2) calls `MultiDocService.chat(...)` at line 79 — a paid LLM round-trip — then (3) feeds the assistant reply straight back through `ChatTranscriptSerializer(data={"message": chat_response, ...})` and `is_valid(raise_exception=True)` at line 103.

`ChatTranscriptSerializer` is an `AuditSerializer` with `fields = "__all__"` (`pluggable_apps/apps/chat_transcript/serializer.py:26-39`) and `ChatTranscript.message` is a `TextField` (`chat_transcript_v2/models.py:28`) → DRF `CharField`, writable, with no `Meta.html_safe_fields`. After this PR the mixin appends `validate_no_html_tags` to it.

Run against the PR checkout's own `utils/input_sanitizer.py` (Django 4.2.1 / DRF 3.14.0), ordinary LLM answers reject:
- `"The invoice total is < 500 USD and the tax rate is > 5%."` → **REJECT** (`<[^>]*>` spans `< 500 USD ... >`)
- `"Sure! if x<y then return 1."` → **REJECT** (`<[a-zA-Z/!]` matches `<y`)
- `"Use <br> to break lines."` / a markdown table → **REJECT**
- `"See data:image/png;base64,..."` → **REJECT** (`JS_PROTOCOL_PATTERN`)

Net result: HTTP 400, the assistant reply is never stored, `chat_history.session_id = api_response[...]` at line 106 never runs so the session id is lost for the next turn, and the conversation is left with an orphaned user turn. `ATOMIC_REQUESTS` defaults to `False` (`backend/backend/settings/base.py:222-223`, no enterprise override), so the user turn committed at line 63 is not rolled back. The LLM spend is wasted.

Same class, same repo: `CannedQuestionSerializer` (`canned_question/serializers.py:21`, `question` TextField, written at `canned_question/views.py:78`) and `ChatHistorySerializer.label` (`chat_history/serializer.py:20`, written at `chat_history/views.py:69` from user-supplied text).

This is not deferrable to the follow-up cloud PR: `unstract-cloud` has no `backend/backend/serializers.py` of its own (`settings/cloud.py` does `from backend.settings.base import *`), so the OSS file changed here is the one that ships there. The default flips the moment the enterprise repo picks up this version — before any opt-out can land.

**Suggested fix:** Don't flip the `AuditSerializer` default before the enterprise opt-outs exist. Either (a) land `Meta.html_safe_fields = ("message",)` on `ChatTranscriptSerializer` (plus `question` / `label` on `CannedQuestionSerializer` / `ChatHistorySerializer`) in lockstep so the two repos flip together, or (b) keep `AuditSerializer` on `rest_framework.ModelSerializer` and opt each serializer in explicitly, so a serializer nobody audited cannot silently acquire the validator.

Worth noting regardless of which path you pick: any field that stores **model output** should be exempt by construction rather than by remembering — the three opt-outs in this PR were found the same way Greptile found the `PromptStudioOutputSerializer` one, and the enterprise repo has no reviewer watching this default.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:56eba291a0a1</sub>

frontend/src/components/custom-tools/document-parser/DocumentParser.jsx
```diff
@@ -205,6 +205,16 @@ function DocumentParser({
...
       .catch((err) => {
+        // Field-keyed DRF validation errors are surfaced inline on the
+        // prompt card; re-throw so the card's catch can render them.
+        const data = err?.response?.data;
+        const hasFieldError =
+          data?.type === "validation_error" &&
+          Array.isArray(data?.errors) &&
+          data.errors.some((e) => e?.attr);
+        if (hasFieldError) {
+          throw err;
```
Deepak-Kesavan: **[major] The new re-throw suppresses the global alert for *every* consumer of `handleChangePromptCard`, but `NotesCard` — the other consumer — catches it with an empty handler and has no inline renderer, so a rejected note title now reverts with zero feedback where it previously showed a toast.**

Rename a Note card's title to anything with angle brackets — `Section <A>`, `Line items <qty>`, `amount<total`.

`EditableText` (name = `prompt_key`, `EditableText.jsx:22,62`) calls `NotesCard.handleChange`, which calls `handleChangePromptCard` (`NotesCard.jsx:73`). `ToolStudioPrompt.prompt_key` is a writable `TextField` and is **not** in `ToolStudioPromptSerializer.Meta.html_safe_fields` (only `prompt`, `assert_prompt`, `assertion_failure_prompt`, `output` are), so the PATCH returns `{type:'validation_error', errors:[{attr:'prompt_key', detail:'Prompt key must not contain HTML or script tags.'}]}` — reproduced against the PR checkout.

`hasFieldError` is therefore true, so `DocumentParser.jsx:216` re-throws and **skips** `setAlertDetails(handleException(err, "Failed to update"))` on line 218. The rejection lands in `NotesCard.jsx:87-90`, whose catch only calls `handleUpdateStatus(...)` and `setPromptDetailsState(prevPromptDetailsState)` — no alert. `NotesCard` never received the `fieldErrors` state or the `error` prop that `PromptCard`/`Header` got in this PR, and its `EditableText` is rendered without one.

The revert then flows back through `EditableText`'s `useEffect(() => setText(defaultText), [defaultText])` (`EditableText.jsx:28-30`) and rewrites the input. **Before this PR the user saw a "Failed to update" toast; now the title just snaps back with no explanation at all.** One verifier simulated the prop chain end-to-end: `NOTES CARD -> alerts shown: 0`.

**Suggested fix:** Keep the global alert as the fallback for callers that can't render inline. Cheapest correct fix is in `NotesCard`'s catch: if the error is field-keyed, call `setAlertDetails(handleException(err, "Failed to update"))` before reverting. Cleaner: have `DocumentParser` take an `onFieldError` callback and only skip the alert when one is supplied, so the "who renders this?" decision lives at the call site instead of being implicit.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:f484f3389fc1</sub>

frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
```diff
@@ -178,9 +187,26 @@ const PromptCard = memo(
...
+            data?.type === "validation_error" &&
+            Array.isArray(data?.errors)
+          ) {
+            data.errors.forEach((e) => {
+              if (e?.attr) {
+                fieldErrorMap[e.attr] = e.detail || "Invalid value";
+              }
+            });
+          }
+          if (Object.keys(fieldErrorMap).length > 0) {
```
Deepak-Kesavan: **[major] `PromptCard` populates `fieldErrors` for *every* rejected attr and skips the rollback on that basis, but only `prompt_key` is ever rendered — so any other attr produces no inline error, no rollback, and no alert: a completely silent failed save.**

Set a post-processing webhook URL that trips the sanitizer, e.g. `https://hooks.example.com/x?data:application/json` (matches `JS_PROTOCOL_PATTERN`). `ToolStudioPrompt.postprocessing_webhook_url` is a writable `TextField` not in the opt-out tuple.

The PATCH returns 400 with `attr='postprocessing_webhook_url'`. `DocumentParser.jsx:216` re-throws (field-keyed). `PromptCard.jsx:198-206` writes `fieldErrors['postprocessing_webhook_url']`, so `Object.keys(fieldErrorMap).length > 0` is true and the `else` branch at `:208` — `setPromptDetailsState(prevPromptDetailsState)` — is **skipped**. `PromptCardItems.jsx:235` passes only `fieldErrors?.prompt_key` down, so nothing renders it. And `Header`'s own rollback `.catch(() => setStateFn(...))` (`Header.jsx:137`) never fires, because `PromptCard.handleChange` resolves after swallowing the error.

Net user-visible result: the field keeps showing a value the server rejected, no message appears anywhere, and the card's state silently diverges from the server. The same holds for `enforce_type`, `required`, and any non-field error — `drf-standardized-errors` sets `attr='non_field_errors'` rather than null, so those take the same silent path.

**Suggested fix:** Only take the inline-error path for attrs the UI can actually render, and fall back to the old behaviour otherwise — e.g. gate on a `RENDERABLE_FIELD_ERRORS` set (currently just `prompt_key`); for anything left over keep `setPromptDetailsState(prevPromptDetailsState)` and `setAlertDetails(handleException(err, "Failed to update"))`. As written, adding one field to the sanitizer's reach silently degrades that field's error handling to nothing, which is the opposite of what the inline-error work is for.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:95daaa8949fa</sub>

backend/backend/serializers.py
chandrasekharan-zipstack: Fixed at the root rather than by delaying the default flip.

`HTML_TAG_PATTERN` now matches only script-capable tags — `script|iframe|object|embed|svg|math|link|meta|base|form|style|applet|frame|frameset|template` — and `JS_PROTOCOL_PATTERN` only the `data:` MIME types that can carry markup (`text/html`, `image/svg+xml`, the script types). All four of your reject cases pass now: `The invoice total is < 500 USD and the tax rate is > 5%.`, `if x<y then return 1.`, `Use <br> to break lines.`, `See data:image/png;base64,...`. Each is a regression test in `test_input_sanitizer.py`.

`` is still rejected, via the event-handler rule.

On the two enterprise serializers you named: `CannedQuestionSerializer.question` and `ChatHistorySerializer.label` are user-typed text, so the validator is what we want there — they stay covered. `ChatTranscript.message` stores model output and keeps `html_safe_fields = ("message",)` in cloud #1499 for an answer that quotes a literal `<script>`. `Chat.jsx` there now has a catch that resets the composer and surfaces the error.

Keeping the opt-out default: the VAPT finding is specifically that opt-in coverage misses serializers nobody remembers to wire up. Narrowing what counts as dangerous addresses the false positives without giving that up.

Commit `5ffaa0af1`.

frontend/src/components/custom-tools/document-parser/DocumentParser.jsx
chandrasekharan-zipstack: Took the cleaner option you suggested.

`handleChangePromptCard(name, value, promptId, onFieldError)` — the caller says whether it rendered the error. `PromptCard` returns `true` only for attrs it can actually show; anything else returns `false` and `DocumentParser` falls back to `setAlertDetails(handleException(err, "Failed to update"))`. `NotesCard` passes no callback at all, so it gets the toast back with no change to that file.

The re-throw stays unconditional, so the existing local rollbacks still run.

Commit `ee6216501`.

frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
chandrasekharan-zipstack: Gated on `RENDERABLE_FIELD_ERRORS` as suggested (currently just `prompt_key`). Anything else — `postprocessing_webhook_url`, `enforce_type`, `required`, `non_field_errors` — takes the old path: rollback plus alert. The rollback now keys on whether an inline error was actually rendered, not on the shape of the error, so widening the sanitizer's reach can't silently degrade a field to nothing.

Also fixed the dead `Header` rollbacks you spotted while you were in there. `PromptCard.handleChange` re-throws after handling, so `Header.jsx`'s `.catch(() => setStateFn(...))` fires for the webhook URL, webhook toggle, disable and required controls. The three call sites with no local copy to restore (`handleSelectDefaultLLM`, `handleTypeChange`, `EditableText`) opt out with an explicit no-op catch rather than leaving floating rejections.

Commits `ee6216501`, `be37d40d1`.

frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
```diff
@@ -157,14 +162,40 @@ const PromptCard = memo(
       const updatedPromptDetailsState = { ...promptDetailsState };
       updatedPromptDetailsState[name] = value;
 
+      // New attempt — drop any prior inline error for this field.
+      setFieldErrors((prev) => {
```
Deepak-Kesavan: **[minor] [unverified] `fieldErrors` is cleared when an attempt *starts* but never when one *succeeds*, so an out-of-order rejection paints a permanent inline error over a `prompt_key` the server accepted.**

In the prompt-key input: user types `bad<form>` — `EditableText`'s 1000 ms debounce (EditableText.jsx:52-65) fires attempt A, which clears `fieldErrors.prompt_key` and PATCHes. User keeps typing `good_key` -> attempt B clears again and PATCHes. B returns 200 first; A's 400 lands afterwards, `showInlineErrors` runs and sets `fieldErrors.prompt_key`. Nothing ever clears it: the success branch (PromptCard.jsx:199-211) calls only `setUpdatedPromptsCopy` and `handleUpdateStatus`, no `useEffect` resets it, and the card is keyed by `item.prompt_id` inside a memo'd `PromptCardWrapper` (DocumentParser.jsx:300) so a successful save does not remount it either. Result: the key is persisted correctly but the card shows `status="error"` plus a red "Prompt key must not contain HTML or script tags." indefinitely, until the user happens to edit `prompt_key` again.

`grep -n setFieldErrors -r frontend/src/` returns exactly three hits: the `useState` (:60), the clear at :166, and the set inside `showInlineErrors` (:190). `handledInline` is a per-invocation `let`, so a stale rejection has no way to know a later attempt superseded it, and the input is not disabled while a PATCH is in flight (`EditableText` `disabled` is only `isCoverageLoading || isSinglePassExtractLoading || isPublicSource`), so two overlapping PATCHes for the same field are genuinely possible.

Marked unverified because I did not reproduce the ordering at runtime: the 1000 ms trailing debounce means A must be more than a second slower than an identical B, so this is a latency-tail case rather than a routine flow. The missing clear-on-success is unconditional, though.

To be clear about what is *not* wrong here: the skipped rollback is correct in this scenario. `setPromptDetailsState(updatedPromptDetailsState)` runs synchronously at invocation (:176), so B's value wins regardless of resolution order — had `handledInline` been false, A's rollback to its own snapshot would have clobbered the accepted `good_key`.

**Suggested fix:** Clear `fieldErrors[name]` in the `.then` success branch so a later success always beats an earlier failure. If you want the ordering airtight, give each attempt a monotonic per-field generation (a `useRef({})` counter incremented in `handleChange` and captured in the closure) and have `showInlineErrors` set the error only when its generation is still the latest.

<sub>Not reproduced — flagged for your judgement rather than asserted.</sub>

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:a4bc1eaf0ffe</sub>

frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
```diff
@@ -178,9 +209,14 @@ const PromptCard = memo(
             setUpdateStatus,
           );
         })
-        .catch(() => {
+        .catch((err) => {
           handleUpdateStatus(isUpdateStatus, promptId, null, setUpdateStatus);
-          setPromptDetailsState(prevPromptDetailsState);
+          if (!handledInline) {
```
Deepak-Kesavan: **[minor] [unverified] In single-pass extract / Simple Prompt Studio, skipping the `promptDetailsState` rollback leaves the card's `prompt_key` on the rejected value while outputs stay keyed by the stored key, so highlight and confidence data silently stop resolving for that prompt.**

A follow-on from the rollback split you added in `ee6216501` — a different consequence from the one raised last round, and narrower than it first looks, so here is the scoped version.

User renames a prompt key to something the sanitizer rejects, e.g. `total <form> amount` (`form` is in the narrowed `HTML_TAG_PATTERN`). The PATCH 400s, `showInlineErrors` returns true, `handledInline` is true, and `setPromptDetailsState(prevPromptDetailsState)` is skipped (PromptCard.jsx:214-216). `promptDetailsState.prompt_key` now holds `total <form> amount`; the DB holds the old key. `PromptCard.jsx:79-88` re-syncs from props only once (`isPromptDetailsStateUpdated` one-shot guard), so a later successful PATCH of a different field — which does refresh the zustand `details` store via DocumentParser.jsx:210-221 — never repairs it. The divergence survives until the card unmounts.

**Where it actually bites:** `PromptOutput.jsx:242/246/249` index `promptOutputData.highlightData` / `.confidenceData` / `.wordConfidenceData` by `promptDetails.prompt_key`, so all three resolve to `undefined` — no highlight boxes, no confidence, no error. But that block sits inside the `(singlePassExtractMode || isSimplePromptStudio)` branch (PromptOutput.jsx:189-192). The **ordinary** run path (:289-290) passes those props unkeyed (:494-496), because `output_manager_helper.py:160-171` unwraps `highlight_data.get(prompt.prompt_key)` server-side for non-single-pass. So this reproduces in single-pass extract mode, and on the Simple Prompt Studio route (`promptPatchApiSps`, DocumentParser.jsx:192-194) — not on the default path. `OutputForDocModal` (PromptCard.jsx:432) also shows the rejected key as its heading, but that is cosmetic.

Marked unverified: reproduced by reading, not by running, and it needs single-pass mode plus a user working past a visible red inline error.

**Suggested fix:** Keep the rejected text in the input — that is the point of the feature — but keep it out of `promptDetailsState`, which is the card's model of the *persisted* row. Roll `promptDetailsState` back unconditionally and drive the input from a separate draft: `promptKey` is already its own `useState`, so leaving it on the typed text and suppressing `EditableText`'s `defaultText` effect while `fieldErrors.prompt_key` is set would keep every consumer keyed on the server's `prompt_key` working.

<sub>Not reproduced — flagged for your judgement rather than asserted.</sub>

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:604a83e3a39c</sub>

Deepak-Kesavan (commented):
## 🤖 Automated review (re-review)

2 inline comment(s) · 2 finding(s) not on changed lines (below).

### Findings off the diff

- `backend/utils/input_sanitizer.py:17`
**[critical] `HTML_TAG_PATTERN`'s ambiguous `<\s*/?\s*` prefix backtracks quadratically, and this PR is what attaches it to every writable CharField/TextField on ~18 OSS serializers plus every enterprise `AuditSerializer` subclass — so one small authenticated request pins a gunicorn worker for minutes of pure CPU.**

Any authenticated org member PATCHes a covered field with `"<" + " "*N + "x"` (no leading/trailing whitespace, so DRF's `trim_whitespace` does not shorten it). Measured end-to-end through the real serializer field in the PR checkout, using the checkout's own venv (DRF 3.17.1 per `backend/uv.lock`):

```
  2 KB body ->  0.16 s   8 KB ->  2.51 s   16 KB -> 10.37 s   32 KB -> 40.16 s
```

Clean O(n^2) — and `is_valid()` returns **True**. The value is not even rejected; it is pure wasted CPU.

Django's default `DATA_UPLOAD_MAX_MEMORY_SIZE` is 2.5 MB (no override in `backend/backend/settings/`), so a single 2.5 MB `description` extrapolates to ~65 CPU-hours in one request. CPython's regex engine does not release the GIL, and `backend/entrypoint.sh:81-85` runs gunicorn `--worker-class gthread --threads 512 --workers 2 --timeout 600` — so one such request freezes an entire worker process (all 512 threads, every tenant) and is not reaped for 10 minutes. Two concurrent requests take the shared backend down.

Reachable fields include `CustomTool.description` and `ToolStudioPrompt.prompt_key` (both `models.TextField` with no `max_length`, so no `MaxLengthValidator` in front — and DRF's `run_validators` collects all errors rather than short-circuiting, so the sanitizer runs even on bounded fields), plus `Tag.description`, `Notification.*`, and every enterprise `AuditSerializer` CharField (`CannedQuestion.question`, `ChatHistory.label`, `AppDeployment.description`) — the enterprise repo declares no `html_safe_fields` anywhere.

`origin/main` was **not** vulnerable: its `HTML_TAG_PATTERN = r"<[^>]*>|<[a-zA-Z/!]"` is linear (0.036 ms at 32 KB).

**Suggested fix:** Collapse the ambiguous prefix into one unambiguous character class:
```python
HTML_TAG_PATTERN = re.compile(rf"<[\s/]*({_SCRIPTABLE_TAGS})\b", re.IGNORECASE)
```
Verified in the checkout: 200 000 spaces -> 16 ms (linear), and it still matches `<script>x`, `</ script >`, `< / SCRIPT`, `<svg onload=1` while still allowing `qty < 500` and `the <invoice_no> field`. Also cap input length before running any regex, and add a regression test asserting the validator completes in bounded time on `"<" + " "*50000 + "x"`.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:fd36ebbebd19</sub>

- `backend/utils/input_sanitizer.py:38`
**[major] Follow-up to your reply on commit `5ffaa0af1`: narrowing the tag rule to 15 script-capable tags leaves `EVENT_HANDLER_PATTERN`'s hand-maintained `_DOM_EVENTS` allow-list as the only guard on inert carrier tags — and it omits every modern no-interaction event, so payloads `origin/main` rejected are now accepted on fields main already protected.**

Your reply says "`` is still rejected, via the event-handler rule" — true for `onerror`, but the rule is a fixed ~60-name list and the tag rule no longer backstops it. Executed against the shipped `validate_no_html_tags` / `validate_name_field` in the PR checkout, compared with `git show origin/main:backend/utils/input_sanitizer.py`:

```
  main=rejected  PR=ACCEPTED  <video src=1 onloadstart=alert(1)>      media autoload, zero interaction
  main=rejected  PR=ACCEPTED  <audio src=x onloadedmetadata=alert(1)> zero interaction
  main=rejected  PR=ACCEPTED  <b onbeforetoggle=alert(1)>             popover API
  main=rejected  PR=ACCEPTED  <b onanimationend=alert(1)>             CSS animation
  main=rejected  PR=ACCEPTED  <body onpageshow=alert(1)>              bfcache
  main=rejected  PR=ACCEPTED  <x onauxclick=alert(1)>
  main=rejected  PR=ACCEPTED  <img src=x onerror/**/=alert(1)>        comment between name and '='
```

Two independent lenses reproduced this by importing the module from the checkout, not by reading it. Also missing from the list: `oncanplay`, `onended`, `ontimeupdate`, `ontransitionend`, `onbeforeinput`, `onslotchange`, `onhashchange`, `onpopstate`, `onbeforeunload`.

The sharp part is that this is a **regression on fields main already guarded**, not just an incomplete new guard: `validate_name_field("<b onbeforetoggle=alert(1)>")` now returns ACCEPTED, and it fronts `APIDeployment.display_name` (api_v2/serializers.py:83), `Workflow.workflow_name` (workflow_v2:65), `CustomTool.tool_name` (prompt_studio_core_v2:127), `AdapterInstance.adapter_name` (adapter_processor_v2:51), `connector_v2:57`, `notification_v2:126`, `account_v2/serializer.py:15,18`. The three `description` fields whose `validate_description` this PR deletes are in the same position: the mixin calls the identical function, but that function is now weaker than the one main ran.

**Not currently exploitable, and I want to be straight about that:** I found no `dangerouslySetInnerHTML` or `rehype-raw` in OSS `frontend/src` (`CustomMarkdown.jsx` is token-based, `MarkdownRenderer.jsx` uses `remarkGfm` only), no `mark_safe`/`format_html` on these fields in either backend, and the one `unstract-cloud` hit (`plugins/verticals/.../ValidatedTextArea.jsx:198`) escapes at line 87 and is not fed by these fields. This is defence-in-depth that got thinner, under a PR titled as stored-XSS prevention — not a live XSS path.

**Suggested fix:** Make the event rule structural instead of an allow-list, so it cannot rot as new DOM events ship:
```python
EVENT_HANDLER_PATTERN = re.compile(r"<[a-zA-Z][^>]*\bon[a-z]{3,}\s*=", re.IGNORECASE | re.DOTALL)
```
Benign prose (`onboarding=`, `connection=`) still passes because it has no leading `<tag`, which is what the allow-list was built to protect. Add `onloadstart`, `onloadedmetadata`, `onbeforetoggle`, `onanimationend`, `onpageshow` and `onerror/**/=` as regression tests alongside the four false-positive cases from `5ffaa0af1`.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:ec64e2a0c2b4</sub>

frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
chandrasekharan-zipstack: Fixed in `06fad5445`.

You are right that the clear was one-sided. `handleChange` now takes a per-field generation from a `useRef` counter, and every state write in both the success and the failure path is gated on that attempt still being the latest:

- success clears `fieldErrors[name]`, so a rejection followed by a correction no longer leaves the red text behind;
- a superseded rejection sets nothing at all, so the latency-tail ordering you traced (B accepted, A's 400 lands after) can no longer paint an error over a value the server took.

The generation also gates the rollback, which matters now that the rollback is unconditional (see the sibling thread) - a stale snapshot must not clobber an accepted value.

frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
chandrasekharan-zipstack: Fixed in `06fad5445`, taking the split you suggested.

`setPromptDetailsState(prevPromptDetailsState)` is now unconditional, so the card's copy always tracks the stored row and the single-pass / Simple Prompt Studio consumers keyed by `prompt_key` keep resolving. The rejected text stays visible because `EditableText` skips its `defaultText` sync while an inline error is showing, which is the only reason the rollback was being skipped in the first place.

One implementation note: `showInlineErrors` no longer calls `setFieldErrors` itself - it records the renderable errors and the rejection handler applies them together with the rollback. That keeps both in one synchronous block, so they are guaranteed to land in the same render and the `error` prop is set by the time `EditableText`'s effect runs on the rolled-back `defaultText`.

Deepak-Kesavan (commented):
## 🤖 Automated review (re-review)

1 inline comment(s) · 1 finding(s) not on changed lines (below).

### Findings off the diff

- `frontend/src/components/custom-tools/prompt-card/PromptCard.jsx:164`
**[major] Making the rollback unconditional is the right call, but its target is wrong: `prevPromptDetailsState` is a whole-object snapshot of the card's own *optimistic* state, not a read of the stored row — so the rollback reverts sibling fields the server accepted in the meantime, and across overlapping rejections it settles on a value the server rejected.**

Both halves were reproduced independently by two lenses, one of them by rendering the real `PromptCard` under vitest with `handleChangePromptCard` returning deferred promises.

**(a) A prompt_key rejection reverts a sibling field the server just accepted — this half is new in `06fad5445`.**
Stored row: `prompt_key="A"`, `active=true`. User edits the key; after the 1000 ms debounce (EditableText.jsx:57-62) the PATCH goes out and `prevPromptDetailsState` is snapshotted as the whole object. While it is in flight the user flips the disable toggle — `Header.jsx:158-176` calls `handleChange` with no debounce at all — and the server accepts it, so `active=false` locally and remotely. The key PATCH then 400s. `attemptGen` is keyed **per field name**, so an accepted PATCH on `active` never bumps the `prompt_key` generation: `isLatestAttempt()` is true and PromptCard.jsx:229 writes the whole stale snapshot back, flipping `active` to `true`. `Header.jsx:201-206` (`useEffect(..., [promptDetails, details])`) re-derives `isDisablePrompt`/`required`/`webhookEnabled`/`webhookUrl` off the new object identity, so the toggle visibly reverts while the server holds the new value. Same shape with the prompt body: `PromptCardItems.jsx:218-229` feeds `defaultText={promptDetails?.prompt}` to the textarea `EditableText`, which is never passed `error`, so the new early return at EditableText.jsx:29-33 does not protect it — its `[defaultText]` effect calls `setText` and the body the user just saved disappears from the screen.
Probe output at `06fad5445`: `after toggle accepted: {..."active":false}` -> `after key 400 rollback: {..."active":true}`; and `body accepted, card.prompt = P1-new body` -> `after key 400 rollback, card.prompt = P0`. At `e6e048a61` neither happened, because `if (!handledInline)` skipped the rollback entirely for a renderable prompt_key error. This path is created by making the rollback unconditional.

**(b) Across two overlapping rejections the card settles on a rejected value — the invariant your commit message states does not hold.**
Stored `prompt_key="good"`. Attempt 1 PATCHes `bad1` and writes it optimistically at :190. A render happens, so attempt 2 — fired ~1 s later by the same debounce, with the input **not** disabled while saving (EditableText.jsx:118-120 disables only on `isCoverageLoading`/`isSinglePassExtractLoading`/`isPublicSource`) — captures `prevPromptDetailsState = {prompt_key: "bad1"}`. Both 400. Attempt 1's rejection is correctly ignored. Attempt 2 rolls back to *its* snapshot, so the card settles on `bad1`, which the server rejected and never stored.
Rendered-component probe: `06fad5445` -> `bad1`; `e6e048a61` -> `bad2`; correct answer `good`. The generation counter suppresses the superseded attempt's *write* but never corrects the surviving attempt's *target*. Nothing repairs it afterwards: the success path only writes `updatedPromptsCopy` (:206-209), and the resync effect at :83-92 is one-shot behind `isPromptDetailsStateUpdated`.
Downstream this is the same consequence as the finding this commit closes — `PromptOutput.jsx:242/246/249` key `highlightData`/`confidenceData`/`wordConfidenceData` on `promptDetails.prompt_key` inside the `(singlePassExtractMode || isSimplePromptStudio)` branch (:189-192), and `PromptCard.jsx:449` passes it as `OutputForDocModal`'s `promptKey`.

Both halves need a PATCH round-trip to outlast the next user action — ordinary on a loaded backend, and exactly the race your new comment at PromptCard.jsx:60-63 acknowledges. Neither is a deterministic failure on every rejection.

**Suggested fix:** Roll back to the last server-confirmed value, and only for the field this attempt owns:

```js
// seeded from `promptDetails` in the :83-92 effect, refreshed in .then from res.data
const lastSaved = useRef({});
// ...
setPromptDetailsState((prev) => ({ ...prev, [name]: lastSaved.current[name] }));
```

That fixes both halves at once: sibling fields are never touched, and the rollback becomes idempotent across any number of overlapping rejections. Reading the stored value out of `useCustomToolStore.getState().details.prompts` by `prompt_id` at rejection time works equally well — `DocumentParser.jsx:207-224` keeps that copy current on every success. `frontend/` has vitest 3.2.6 wired up with a `test` script, and the deferred-promise harness both lenses used to reproduce this is about 30 lines, so the overlapping-rejection case is cheap to lock down.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:bc01ad124064</sub>

frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
```diff
@@ -157,30 +166,74 @@ const PromptCard = memo(
...
-        .catch(() => {
+        .catch((err) => {
           handleUpdateStatus(isUpdateStatus, promptId, null, setUpdateStatus);
-          setPromptDetailsState(prevPromptDetailsState);
+          if (isLatestAttempt()) {
+            // This state mirrors the stored row — outputs, highlights and
+            // confidence are keyed off prompt_key — so it tracks the server
+            // even when the rejected text stays in the input.
+            setPromptDetailsState(prevPromptDetailsState);
+            if (inlineErrors) {
```
Deepak-Kesavan, line 240: **[minor] A superseded rejection now reports nothing at all: `showInlineErrors` still returns `true` (so `DocumentParser` suppresses the global alert) while the recorded error is dropped by the `isLatestAttempt()` gate — so a rejected key produces no alert and no inline error, which is the silent failure `RENDERABLE_FIELD_ERRORS` was added to prevent.**

Stored `prompt_key` is `good`. User types `bad<script>`; the debounced PATCH (gen 1) goes out and is slow. The user selects-all and deletes the field to retype; `EditableText` fires a second `handleChange` with `""` (gen 2). `DocumentParser.jsx:165-168` short-circuits an empty `prompt_key` and resolves **without a request**, so gen 2 becomes the latest generation without ever contacting the server. Gen 1's 400 then lands: `showInlineErrors` records into `inlineErrors` and returns `true`, so `DocumentParser.jsx:230-233` suppresses the alert — but `isLatestAttempt()` is false (1 !== 2), so `setFieldErrors` at :231 never runs. The user gets zero feedback that their key was rejected and never saved.

Reproduced by extracting both handlers verbatim and running them against a controllable fake axios: at `06fad5445` the probe printed `fieldErrors.prompt_key = undefined, alerts = 0`; the identical probe against `e6e048a61` printed `Must not contain HTML`, and `origin/main` calls `setAlertDetails` unconditionally. So this commit removes the only signal in this race.

Filed as minor, not major: it needs PATCH latency beyond ~1.2 s plus a clear-the-field gesture, and the stored row is left intact — the cost is the missing feedback, not lost data.

**Suggested fix:** Make the suppression decision and the render decision the same decision — one line:

```js
const showInlineErrors = (errors) => {
  const renderable = Object.entries(errors).filter(([attr]) => RENDERABLE_FIELD_ERRORS.has(attr));
  if (!renderable.length || !isLatestAttempt()) return false;   // <- fall back to the alert
  inlineErrors = Object.fromEntries(renderable);
  return true;
};
```

Alternatively, bump the generation only for attempts that actually reach the network (move the increment past `DocumentParser`'s short-circuit, or have `handleChangePromptCard` signal a no-op). Either way, a superseded rejection that leaves the field un-rendered must fall back to the global alert rather than vanish.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:881886066052</sub>

chandrasekharan-zipstack: Valid, fixed in 998e0719f.

`showInlineErrors` now returns `false` for a superseded attempt, so the suppression decision and the render decision are the same decision:

```js
if (!renderable.length || !isLatestAttempt()) {
  return false;
}
```

A rejection that cannot render its inline error falls back to the global alert instead of vanishing. Took this over moving the generation bump past `DocumentParser`'s short-circuit, since the no-op path would then need to signal back through `handleChangePromptCard` for the same result.

Deepak-Kesavan (approved):

frontend/src/components/custom-tools/prompt-card/PromptCard.jsx
```diff
@@ -165,31 +174,76 @@ const PromptCard = memo(
...
+      let inlineErrors = null;
+      const showInlineErrors = (errors) => {
+        const renderable = Object.entries(errors).filter(([attr]) =>
+          RENDERABLE_FIELD_ERRORS.has(attr),
+        );
+        // A superseded attempt cannot render its error, so it must not claim
+        // the field either — the caller falls back to the global alert.
+        if (!renderable.length || !isLatestAttempt()) {
+          return false;
+        }
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Valid — fixed in 7ebdc3d71.

The callback conflated two different reasons for returning `false`: "this field has no inline renderer" (the caller *should* alert) and "a newer attempt has superseded this one" (it should not). Only the first deserves the fallback.

`reportFieldError` now returns early for a superseded attempt, so the rejection is claimed and stays silent — the newer attempt owns the field and reports its own outcome, whether it succeeds or fails.

While there, `DocumentParser` now consults the callback for every rejection instead of only ones carrying field errors. Otherwise a stale attempt that failed with a non-field error (a 500, a dropped connection) would still have alerted over a value the user can see was accepted.

PR #1971, opened 2026-05-19
[MISC] Add tox-driven test rig foundation (unit + integration + e2e)

hari-kuriakose:
## Summary

- Adds a single tox-driven entry point that brings every existing test suite under one rig, with consistent reporting.
- Introduces an e2e tier on top of the existing docker-compose stack; testcontainers + local subprocesses available for dev.
- Surfaces **critical-path gaps** and **regressions** explicitly in CI reports — the team target is zero regressions.

This is the **foundation only** — no new tests beyond a single e2e smoke. Existing co-located unit tests stay where they are; the rig just discovers and runs them.

## What's new

- **`tests/groups.yaml`** — single source of truth for 15 groups (9 unit, 1 integration placeholder, 5 e2e) with a `depends_on` graph.
- **`tests/critical_paths.yaml`** — 10 critical flows registered, each pointing at the group(s) that cover it.
- **`tests/rig/`** — the dispatcher (`python -m tests.rig run|list-groups|expand|validate|platform|report`).
- **`tests/e2e/`** — `platform` session fixture + one smoke test exercising the e2e plumbing end-to-end.
- **`tox.ini`** — rewritten with tier-shaped envs (`unit`, `integration`, `e2e`, `groups`, `rig`); legacy `runner|sdk1|prompt-service` envs preserved as thin aliases.
- **CI** — `ci-test.yaml` rewritten to call the rig; new `ci-test-e2e.yaml` slow lane (main + `run-e2e`-labeled PRs + nightly).
- **Pre-commit hook** — `unstract-tests` fires only when developers opt in by creating `.test-selection`.

## Usage

```bash
tox -e unit                                    # all unit groups, parallel, with coverage
tox -e e2e -- e2e-smoke                        # smoke against docker compose
tox -e groups -- unit-backend e2e-smoke        # explicit groups (deps auto-pulled)
tox -e groups -- --from-file .test-selection   # newline-delimited group list
tox -e rig -- list-groups                      # discovery
python -m tests.rig validate                   # CI-friendly manifest check
```

Full docs: [`tests/README.md`](tests/README.md).

## Reports

- Per-group: `reports/<group>/{junit.xml, report.md, exit.txt}`
- Aggregate: `reports/{summary.md, summary.json, combined-test-report.md}` — same `combined-test-report.md` path keeps the existing sticky-PR-comment workflow working.
- Coverage (when `--coverage`): `reports/{coverage.xml, htmlcov/}` — `coverage.xml` is consumable by the existing SonarCloud integration.

## Critical paths

CI on `main` runs with `--fail-on-critical-gap --update-baseline`, so:
- ⚠️ **Gaps** — currently 6 (pipeline-etl-execute, usage-token-tracking, workflow-execution-fan-out, callback-result-delivery, plus the 4 covered by e2e groups that haven't shipped tests yet). Visible in PR comments; not blocking until covered.
- ❌ **Regressions** — paths that were covered on the cached main baseline but aren't now. Always blocking.

## Test plan

- [x] `python -m tests.rig validate` succeeds (`OK — 15 groups, 10 critical paths`).
- [x] `python -m tests.rig list-groups` and `expand <group>` print correct dep-expanded ordering.
- [x] `validate` correctly fails on an injected unknown-group reference in `critical_paths.yaml`.
- [x] `python -m tests.rig run --no-coverage --no-parallel unit-core` end-to-end: creates per-group venv, installs rig pytest plugins, runs pytest, writes `junit.xml` + `report.md` + `exit.txt`, and aggregates into `summary.md`/`summary.json` with critical-path gaps section.
- [x] `bash .github/scripts/combine-test-reports.sh` shim works (delegates to `python -m tests.rig report combine` and preserves the legacy `combined-test-report.md` path).
- [ ] Reviewer: trigger `ci-test.yaml` on this PR and confirm the sticky comment renders.
- [ ] Reviewer: label this PR `run-e2e` to smoke-test `ci-test-e2e.yaml` against docker compose.

## Notes for reviewers

- The unit-core run surfaced two pre-existing import errors in `unstract/core/tests/` (`LogHelper` not exported; `account_services` submodule missing). Those are **not** rig issues — the rig correctly captured them as errors. Fixing them is out of scope here.
- `unstract/connectors/uv.lock` had unrelated churn (litellm source switch); intentionally left out of this PR.
- No existing tests were moved. Per-service unit tests stay co-located with their service code; only e2e and cross-service integration tests live under `tests/`.


tests/rig/cli.py
```diff
@@ -0,0 +1,689 @@
...
+    "pytest>=8.0.1",
+    "pytest-md-report>=0.6.2",
+    "pytest-timeout>=2.3.1",
+    "pytest-xdist>=3.5.0",
+    "pytest-cov>=4.1.0",
+    "pytest-mock>=3.12.0",
+)
+
+
+def _prepare_group_env(group: GroupDefinition, *, env: dict[str, str]) -> None:
```
jaseemjaskp, line 617: **Silent `uv sync` / `uv pip install` failures (HIGH)**

Three `subprocess.run(..., check=False)` calls swallow returncode and stderr unconditionally (lines 556, 562, 569). A network blip during `uv sync`, a broken lock-file, a missing dep group, or registry auth failures all leave the group with a half-prepared venv. `pytest` then runs against incomplete deps and either fails loudly with a *different* error or — worse — passes because the affected import path was not exercised.

The comment at cli.py:500-503 documents this as intentional, but the cost is high: a silent install failure is indistinguishable from a real test pass in CI, and the upstream uv warnings are easy to miss in xdist-interleaved output.

**Suggested fix:** capture `result.returncode` and stderr; on nonzero, log a marker line that includes the group name and abort the group with a synthetic junit (errors=1) so the failure shows up in the markdown summary.

tests/rig/cli.py
```diff
@@ -0,0 +1,689 @@
...
+    Exit 5 ("no tests collected") must produce failures=0; otherwise an empty
+    hurl group would show ⚪ via :class:`GroupResult` while also being counted
+    as a failure in totals + critical-path evaluation.
+    """
+    is_failure = exit_code != 0 and exit_code != 5
+    failures = 1 if is_failure else 0
+    failure_tag = f'<failure message="hurl exit {exit_code}"/>' if is_failure else ""
+    path.write_text(
+        '<?xml version="1.0" encoding="UTF-8"?>\n'
+        f'<testsuite name="{group_name}" tests="1" failures="{failures}" '
```
jaseemjaskp: **XML injection in synthetic JUnit when group name contains `"` / `&` / `<` (CRITICAL)**

`group_name` is interpolated raw into `name="{group_name}"` and `classname="{group_name}"` (lines 668 & 670). `_build_group` performs no character validation on the YAML key, so a group renamed to e.g. `hurl-api & docs` (or anything with a quote) produces malformed XML. `parse_junit` then takes the malformed-XML branch in `reporting.py` and the build records `errors=1` even when hurl actually passed — a phantom failure on a green run.

**Suggested fix:** Build the XML with `xml.etree.ElementTree` (or `xml.sax.saxutils.escape(name, {'\"': '&quot;'})`) instead of f-string interpolation. Cheap one-liner that closes the class of bug entirely.

tests/rig/cli.py
```diff
@@ -0,0 +1,689 @@
...
+    # If you're debugging "ModuleNotFoundError" in a group, scroll up for the
+    # uv warnings — they're the smoking gun.
+    _prepare_group_env(group, env=env)
+
+    workdir = group.absolute_workdir()
+
+    if group.runner == "hurl":
+        cmd = _hurl_command(group, workdir)
+        exit_code = _spawn(cmd, env=env, cwd=workdir, timeout=timeout)
+        _write_synthetic_junit(junit, group.name, exit_code)
```
jaseemjaskp: **Inconsistent error handling between synthetic-junit write and exit.txt write (MAJOR)**

`_write_synthetic_junit(junit, group.name, exit_code)` (line 511) calls `path.write_text(...)` with no error handling, while the very next block (lines 527-533) wraps the equally trivial exit.txt write in a defensive `try/except OSError`. If `reports/<group>/` is read-only or the disk fills, the synthetic-junit write raises, aborts the `cmd_run` loop, and orphans data from previously-completed groups before any summary is rendered.

**Suggested fix:** Wrap the call in the same `try/except OSError` pattern used for exit.txt, log the failure, and let the loop proceed so other groups still report.

tests/rig/cli.py
```diff
@@ -0,0 +1,689 @@
...
+                workers=args.workers,
+                timeout=args.timeout or group.timeout_seconds,
+                endpoints=endpoints,
+            )
+            if result is not None:
+                group_results.append(result)
+            # Always fold the exit code into overall_exit, even when junit.xml
+            # was never written (segfault/OOM/missing binary). Otherwise the
+            # rig silently returns 0 for catastrophic group failures.
+            if exit_code not in _NON_FAILING_PYTEST_EXIT_CODES and overall_exit == 0:
```
jaseemjaskp: **`errors>0`+`exit_code==0` lets rig exit 0 with a red report (MAJOR)**

If `parse_junit` returns a `GroupResult` with `errors=1` (truncated junit, non-numeric counters per reporting.py:75–128) and pytest's own exit code was 0, the markdown will render `❌` but `overall_exit` is never bumped — only `exit_code not in _NON_FAILING_PYTEST_EXIT_CODES` flips it. Net effect: report says fail, CI exits green.

**Suggested fix:** After appending each result, also do `if (result.errors or result.failed) and overall_exit == 0: overall_exit = 1`. Cheap belt-and-braces for the case where the junit attests to a problem the exit code didn't.

tests/rig/cli.py
```diff
@@ -0,0 +1,689 @@
...
+        f'  <testcase classname="{group_name}" name="hurl-suite">{failure_tag}'
+        f"</testcase>\n"
+        f"</testsuite>\n"
+    )
+
+
+def _spawn(cmd: list[str], *, env: dict[str, str], cwd: Path, timeout: int) -> int:
+    start = time.monotonic()
+    try:
+        result = subprocess.run(cmd, cwd=cwd, env=env, timeout=timeout + 30)
```
jaseemjaskp, line 753: **`_spawn` timeout doesn't kill the process group → zombies + port conflicts (MAJOR)**

`subprocess.run(..., timeout=...)` only kills the *direct* child. Under `uv run` + `pytest-xdist`, the actual workers are grandchildren. On timeout they orphan, holding xdist sockets and MinIO/Redis clients; subsequent groups in the same invocation can then hit "address already in use" with no diagnostic.

**Suggested fix:** Switch to `subprocess.Popen(..., start_new_session=True)` and on `TimeoutExpired` call `os.killpg(p.pid, SIGTERM)` followed by `SIGKILL` after a short grace. Drain stdout/stderr to avoid hangs caused by a full pipe buffer (chatty tests can deadlock before the timeout fires).

tests/rig/coverage.py
```diff
@@ -0,0 +1,100 @@
...
+    files = sorted(reports_dir.glob(".coverage.*"))
+    if not files:
+        return
+
+    target = reports_dir / ".coverage"
+    if target.exists():
+        target.unlink()
+
+    base = _coverage_base()
+    result = subprocess.run(
```
jaseemjaskp: **No timeout on coverage subprocesses → indefinite hang (MAJOR)**

`subprocess.run([*base, "combine", ...], capture_output=True)` (and the `xml` / `html` invocations at lines 84-93) have no `timeout=`. If `uv run --with coverage[toml]` resolves over a slow index (the same fragility `_prepare_group_env` already inherited), the call hangs the whole job up to the 60-minute GHA cap — and the rig's per-group timeouts don't apply because this runs *after* the loop.

**Suggested fix:** Add `timeout=300` to each `subprocess.run`; on `TimeoutExpired`, log and return. The docstring already says "errors are logged not raised", so this is a natural extension of the existing policy.

tests/rig/runtime.py
```diff
@@ -0,0 +1,293 @@
...
+    def up(self) -> PlatformEndpoints: ...
+    def down(self) -> None: ...
+
+
+class LocalRuntime:
+    """Assume a developer-managed stack; just collect endpoints from env."""
+
+    name: ClassVar[str] = "local"
+
+    def up(self) -> PlatformEndpoints:
```
jaseemjaskp, line 113: **`LocalRuntime.up()` never validates the env it returns (MAJOR)**

`LocalRuntime.up()` returns `PlatformEndpoints.from_env()` with no readiness probe. If the developer forgot to start `run-platform.sh`, every URL defaults to `http://localhost:8000` etc. and the rig hands those to pytest. The first test that touches the platform fails with a connection-refused buried 10 layers deep in fixture setup, instead of a clear "platform isn't running" message from the rig.

**Suggested fix:** Call `_wait_ready(endpoints, timeout_seconds=5)` (or a quick `_responds` sweep with a low budget) so missing infra surfaces at the rig layer with the URL list intact and an actionable error.

tests/rig/runtime.py
```diff
@@ -0,0 +1,293 @@
...
+                "docker",
+                "compose",
+                "-p",
+                self.project_name,
+                *files,
+                "down",
+                "-v",
+                "--remove-orphans",
+            ],
+            check=False,
```
jaseemjaskp, line 156: **`ComposeRuntime.down()` silently discards nonzero returncode (MAJOR)**

`_run(..., check=False)` returns `None` and the returncode is never inspected (line 156). When `docker compose down -v --remove-orphans` fails to remove a volume (locked file, dangling network, daemon error), the leftover Postgres data persists and the next `up()` either fails with a port conflict or — more insidiously — reuses stale schema, causing what looks like a phantom regression in tests/integration groups.

**Suggested fix:** Inspect the result and print stderr on nonzero so the developer / CI log sees the cleanup failure. The rig should not silently lose state-leak warnings.

tests/rig/groups.py
```diff
@@ -0,0 +1,237 @@
...
+@dataclass(frozen=True)
+class GroupDefinition:
+    name: str
+    tier: Tier
+    paths: tuple[str, ...]
+    runner: Runner = "pytest"
+    workdir: str = "."
+    markers: str | None = None
+    pytest_extra: tuple[str, ...] = ()
+    env: dict[str, str] = field(default_factory=dict)
```
jaseemjaskp: **Mutable `env: dict` inside a frozen dataclass leaks shared state (MAJOR)**

`GroupDefinition` is `frozen=True` but `env: dict[str, str] = field(default_factory=dict)` is a mutable container. Any caller can do `group.env["KEY"] = "value"` and silently mutate the manifest record — and because `_execute_group` builds env on top of `group.env` via `update`, a bug that swaps the order (or someone refactoring `_prepare_group_env`) could permanently scribble onto the shared record across groups in one rig invocation.

**Suggested fix:** Use `env: Mapping[str, str] = field(default_factory=dict)` plus `__post_init__` coercion to `types.MappingProxyType(dict(env))`. Adding `slots=True` to all frozen dataclasses in this PR is a free win and catches typo-assignment in future `__post_init__` extensions.

tests/rig/critical_paths.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+@dataclass(frozen=True)
+class CriticalPathRegistry:
+    paths: tuple[CriticalPath, ...]
+    _by_id: dict[str, CriticalPath] = field(default_factory=dict, init=False, repr=False)
+
+    def __post_init__(self) -> None:
+        # `frozen=True` blocks direct assignment; route through object.__setattr__.
+        object.__setattr__(self, "_by_id", {p.id: p for p in self.paths})
```
jaseemjaskp, line 54: **Duplicate `CriticalPath.id` silently last-wins (MAJOR)**

`__post_init__` builds `_by_id = {p.id: p for p in self.paths}` (line 48). Two paths sharing an `id` silently collapse — `by_id()` returns whichever happens to come last, while `paths` still iterates both. `evaluate()` would then compute one CriticalPathStatus per `paths` entry (yielding two near-duplicate rows in the summary, both pointing at the second entry's covered_by). No load-time error.

**Suggested fix:** Before assigning `_by_id`, count ids and raise `ValueError(f"duplicate critical-path ids: {dupes}")`. One extra line of code, eliminates a class of silent manifest bug.

tests/rig/critical_paths.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def by_id(self, path_id: str) -> CriticalPath:
+        if path_id not in self._by_id:
+            raise KeyError(path_id)
+        return self._by_id[path_id]
+
+
+@dataclass(frozen=True)
+class CriticalPathStatus:
```
jaseemjaskp, line 63: **`CriticalPathStatus` allows contradictory state (MAJOR)**

The dataclass permits `state="covered"` with `covering_groups_run=()` — exactly the contradiction `evaluate()` is supposed to prevent. The type carries no invariant; future refactors to `evaluate()` (or a new producer of `CriticalPathStatus`) can silently build the contradictory record and `_render_markdown` will happily emit "✅ covered (no groups)".

**Suggested fix:** Add `__post_init__` enforcing `state == "covered"` ⇒ `covering_groups_run` non-empty, and `state == "regression"` ⇒ `covering_groups_run` empty. Makes the contradictory state unrepresentable instead of relying on reviewer vigilance.

.pre-commit-config.yaml
```diff
@@ -86,6 +86,17 @@ repos:
         files: ^frontend/src/
         types_or: [javascript, jsx, ts, tsx, json]
 
+      - id: unstract-tests
+        name: Run selected unstract tests (.test-selection)
+        # Only fires when the developer has opted in by creating .test-selection
+        # (newline-delimited group names, gitignored). Runs serially without
+        # coverage to stay fast in a commit hook.
+        entry: bash -c 'if [ -f .test-selection ]; then python -m tests.rig run --from-file .test-selection --no-coverage --no-parallel; fi'
```
jaseemjaskp: **Pre-commit hook invokes bare `python` despite combine script's `python3` fix (CRITICAL)**

Line 94 runs `python -m tests.rig run --from-file .test-selection ...`. The companion `combine-test-reports.sh` change in this same PR was explicitly hardened to use `python3` first because stock Ubuntu (and many container images) ship `python3` only — no `python` symlink. A developer on such a machine who creates `.test-selection` to opt in will hit `command not found` (exit 127) from inside the `bash -c '...; fi'` and see no diagnostic about why the hook failed.

**Suggested fix:** mirror the combine-script logic, e.g. `bash -c 'if [ -f .test-selection ]; then $(command -v python3 || command -v python) -m tests.rig run --from-file .test-selection --no-coverage --no-parallel; fi'`.

tests/rig/selection.py
```diff
@@ -0,0 +1,121 @@
...
+        selected.update(_groups_for_changed_paths(manifest, base=changed_base))
+
+    return manifest.expand(sorted(selected)) if selected else []
+
+
+def _groups_for_changed_paths(manifest: GroupManifest, *, base: str) -> set[str]:
+    """Pick groups whose ``paths`` overlap any file in ``git diff base...HEAD``."""
+    try:
+        result = subprocess.run(
+            ["git", "diff", "--name-only", f"{base}...HEAD"],
```
jaseemjaskp: **`--changed-only` is a footgun on push-to-main (MAJOR)**

The `git diff origin/main...HEAD` invocation is correct on PR branches, but on a `push: branches: [main]` event the checked-out commit *is* `origin/main`, so the diff is empty. `_groups_for_changed_paths` returns `set()`, no other selectors are passed → `cmd_run` exits 2 ("no groups selected") and CI goes red for the wrong reason. There's also no auto-fallback when `HEAD == base`.

**Suggested fix:** Detect `git rev-parse HEAD == git rev-parse <base>` and fall back to `HEAD~1...HEAD` (or print a clear actionable error so the user can pick a different selector). At minimum, document that `--changed-only` is PR-only in the README/help text.

jaseemjaskp (commented):
## Multi-agent review summary

Ran six review passes (Code Reviewer, Code Simplifier, Comment Analyzer, PR Test Analyzer, Silent Failure Hunter, Type Design Analyzer) against the rig. Existing coderabbitai/greptile threads cover most of the headline bugs; the inline comments below are net-new findings.

**Top themes uncovered:**
1. **Silent prep failures** — `_prepare_group_env` and synthetic-junit writes both fail-soft in ways that let `pytest` produce a misleading red/green that does not reflect the actual state of the build.
2. **Subprocess hygiene** — `_spawn` timeout does not kill the process group, and `coverage` invocations have no timeout. Both can hang or leak processes between groups.
3. **Type invariants** — `GroupDefinition.env` is mutable inside a frozen dataclass; `CriticalPathRegistry` silently last-wins on duplicate ids; `CriticalPathStatus` allows contradictory `state="covered"` with empty `covering_groups_run`.
4. **Pre-commit hook regression** — uses bare `python` despite `combine-test-reports.sh` being explicitly fixed for `python3`-only environments.
5. **Docstring drift** — three docstrings claim guarantees the surrounding code does not actually provide (URL sentinel, x2text wiring, scope-aware re-aggregation in `cmd_report`).

Findings overlapping with existing bot threads (e.g. `_responds()` laxity, smoke `< 500`, `previous-summary.json` duplication, `cmd_run` complexity, JUnit defusedxml hardening) were deliberately dropped to avoid noise.

Low-impact items not posted inline:
- `runtime.py:67-73` x2text_url is declared on `PlatformEndpoints` but never exported in `cli.py:486-490`. Either delete the field or wire `env.setdefault("UNSTRACT_X2TEXT_URL", ...)`.
- `tests/README.md:144-149` testcontainers row should warn that `down()` is best-effort and leaked containers can block the next run on port conflict.
- `tests/critical_paths.yaml:22-24` perpetual gap TODO has no tracking ticket; consider linking an issue.
- `tox.ini` `passenv = UNSTRACT_*` glob is wide enough to leak adapter API keys into CI logs via compose `_run` (no `capture_output=True`).

No files were modified.

tests/rig/cli.py
```diff
@@ -0,0 +1,689 @@
...
+# Pytest exit codes that the rig treats as non-failure for aggregation:
+#   0 — all tests passed
+#   5 — no tests collected (optional placeholders, empty hurl group, etc.)
+_NON_FAILING_PYTEST_EXIT_CODES = (0, 5)
+
+
+@lru_cache(maxsize=1)
+def _rig_session_id() -> str:
+    """Stable per-invocation sentinel, computed once.
+
```
jaseemjaskp, line 47: **Docstring overstates what the session sentinel proves (MINOR but high-rot risk)**

The docstring says the session ID proves "THIS rig invocation set the platform URLs — not a stale shell value the developer forgot to unset." But lines 487-490 export the URLs via `env.setdefault(...)`, so a pre-set developer/shell value wins over the rig's value. The sentinel only proves the rig ran, not that the rig owns the URLs. This also contradicts the README claim at tests/README.md:152 that `setdefault` is intentional.

**Suggested fix:** Reword to "…so e2e tests can prove the rig ran (URL ownership is intentionally cooperative via `setdefault` so a developer's pre-set `UNSTRACT_*_URL` wins; the session ID is the rig's signature)." Future readers will otherwise misread this guarantee.

tests/rig/cli.py
```diff
@@ -0,0 +1,689 @@
...
+    reports_dir: Path = args.reports_dir
+    combine_and_report(reports_dir)
+    manifest = load_groups()
+    registry = cp.load_critical_paths()
+    group_results: list[GroupResult] = []
+    for name in manifest.names():
+        tier = manifest.get(name).tier
+        result = parse_junit(name, tier, reports_dir)
+        if result is not None:
+            group_results.append(result)
```
jaseemjaskp, line 291: **`cmd_report` has zero self-test coverage (HIGH testing gap)**

`cmd_report` (lines 261-288) is the public entrypoint for re-running `report combine` after a build — used by CI retries and any manual `python -m tests.rig report` invocation. It re-parses every group's junit, re-loads the baseline, re-writes summary.md/json/combined-test-report.md, and handles `baseline_corrupt` propagation. None of `tests/rig/tests/test_cli.py` exercises any of this.

Also, `cmd_report` calls `cp.evaluate(...)` *without* passing `scope_groups`, so any path missing from this build's results is treated as out-of-scope rather than a regression — the asymmetry vs `cmd_run` (which does pass `scope_groups`) is not documented near either call site and is exactly the kind of invariant that rots when one path is refactored.

**Suggested fix:** Add a `test_cmd_report_re_aggregates_existing_junit` that populates two groups' `junit.xml` + `exit.txt` and asserts all three summary artifacts are written. Add a second test asserting `baseline_corrupt` banner appears in the re-aggregated output. Document the `scope_groups=None` choice next to the call.

tests/e2e/smoke/test_smoke.py
```diff
@@ -0,0 +1,18 @@
...
+
+import pytest
+import requests
+
+pytestmark = [pytest.mark.e2e, pytest.mark.critical]
+
+
+def test_backend_health(platform) -> None:
+    response = requests.get(f"{platform.backend_url.rstrip('/')}/health/", timeout=10)
+    assert response.status_code < 500, f"backend /health returned {response.status_code}"
```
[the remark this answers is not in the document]
hari-kuriakose: Intentional. The smoke test now asserts the `UNSTRACT_RIG_SESSION_ID` sentinel is present (proving the rig wired the env) plus `/health` < 500. A stricter status-code/JSON-contract assertion is deferred until the smoke suite grows real endpoint coverage — `< 500` is the right floor for a liveness probe that must tolerate 401/404 on an unseeded stack.

tests/README.md
```diff
@@ -0,0 +1,248 @@
...
+│   ├── smoke/               # Login → /health smoke
+│   ├── workflows/           # (future) workflow execution e2e
+│   ├── api_deployment/      # (future) API deployment e2e
+│   ├── prompt_studio/       # (future) Prompt Studio e2e
+│   └── hurl/                # (future) hurl-based HTTP suites
+├── integration/             # Cross-service tests needing infra but not full platform
+├── fixtures/                # Sample PDFs, JSON, adapter configs
+└── compose/
+    └── docker-compose.test.yaml   # Test overlay on docker/docker-compose.yaml
+```
```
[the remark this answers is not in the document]
hari-kuriakose: Acknowledged — markdownlint MD040 (fenced-code languages) in tests/README.md is a docs-lint nit; will fold into a docs pass. Not blocking the rig.

tests/rig/cli.py
```diff
@@ -0,0 +1,485 @@
...
+    pr.add_argument("--no-parallel", dest="parallel", action="store_false")
+    pr.add_argument("--workers", default="auto", help="pytest-xdist worker count (default: auto).")
+    pr.add_argument("--timeout", type=int, help="Per-group timeout override in seconds.")
+    pr.add_argument("--reports-dir", type=Path, default=REPO_ROOT / "reports")
+    pr.add_argument("--fail-on-critical-gap", action="store_true",
+                    help="Treat uncovered critical paths as a build failure.")
+    pr.add_argument("--changed-only", action="store_true",
+                    help="Auto-select groups overlapping `git diff origin/main...HEAD`.")
+    pr.add_argument("--changed-base", default="origin/main")
+    pr.add_argument("--baseline", type=Path, default=REPO_ROOT / "reports" / "previous-summary.json")
```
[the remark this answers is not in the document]
hari-kuriakose: Fixed in 52e22210 — `previous-summary.json` is referenced via the argparse default; the literal isn't duplicated across call sites (cmd_run/cmd_report/list-critical-paths all take it as an arg). Leaving as-is; extracting a constant for a default-value string adds indirection without removing duplication.

tests/rig/critical_paths.py
```diff
@@ -0,0 +1,132 @@
...
+    if not isinstance(raw, dict) or "paths" not in raw:
+        raise ValueError(f"{path or DEFAULT_REGISTRY}: expected top-level `paths:` list")
+    return CriticalPathRegistry(
+        paths=tuple(
+            CriticalPath(
+                id=str(p["id"]),
+                description=str(p.get("description", "")),
+                entry=str(p.get("entry", "")),
+                covered_by=tuple(p.get("covered_by") or ()),
+            )
```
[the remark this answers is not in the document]
hari-kuriakose: Fixed in 52e22210 (partial). `covered_by` is normalized via `tuple(p.get('covered_by') or ())`; a scalar string would indeed splat into chars. Added duplicate-id detection at load; a scalar-vs-list type check on `covered_by` is a reasonable follow-up but low-risk given the manifest is repo-owned.

tests/rig/critical_paths.py
```diff
@@ -0,0 +1,132 @@
...
+        groups_run_green: names of groups that ran AND passed in this build.
+        baseline: parsed previous-summary.json from the main-branch cache, or None.
+                  Expected shape: ``{"covered_paths": ["auth-login", ...]}``.
+
+    Returns:
+        Statuses in the original registry order.
+    """
+    previously_covered: set[str] = set(
+        (baseline or {}).get("covered_paths", []) if baseline else []
+    )
```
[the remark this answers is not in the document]
hari-kuriakose: Fixed in c492099c / 52e22210 — `load_baseline` and `merge_into_baseline` raise `BaselineCorruptError` on malformed JSON, and `evaluate` reads `(baseline or {}).get('covered_paths') or []` so a non-dict/None baseline degrades safely.

tests/rig/groups.py
```diff
@@ -0,0 +1,165 @@
...
+        # group at runtime too.
+        if g.optional:
+            continue
+        wd = g.absolute_workdir()
+        if not wd.exists():
+            raise ValueError(f"group {name!r}: workdir does not exist: {wd}")
+        for p in g.absolute_paths():
+            # Path may be a directory or a file; just verify it resolves under repo root.
+            if not p.exists():
+                raise ValueError(f"group {name!r}: test path does not exist: {p}")
```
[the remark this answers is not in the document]
hari-kuriakose: Acknowledged — `_validate_paths` checks existence; the comment about resolving under repo root was corrected in an earlier round. A strict repo-boundary containment check is a reasonable hardening follow-up (manifest is repo-owned, so low risk today).

tests/rig/reporting.py
```diff
@@ -0,0 +1,180 @@
...
+  reports/summary.json        — machine-readable
+  reports/summary.md          — human-readable, PR-comment friendly
+  reports/combined-test-report.md — backward-compatible alias for the
+                                    existing sticky-comment workflow.
+"""
+
+from __future__ import annotations
+
+import json
+import xml.etree.ElementTree as ET
```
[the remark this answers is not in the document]
hari-kuriakose: Already uses stdlib `xml.etree.ElementTree` for parsing and the synthetic-junit writer now XML-escapes the group name (52e22210), closing the malformed-XML class.

tests/rig/reporting.py
```diff
@@ -0,0 +1,180 @@
...
+            return "⚪"
+        return "❌"
+
+
+def parse_junit(group_name: str, tier: str, reports_dir: Path) -> GroupResult | None:
+    junit_path = reports_dir / group_name / "junit.xml"
+    exit_path = reports_dir / group_name / "exit.txt"
+    if not junit_path.exists():
+        return None
+    exit_code = int(exit_path.read_text().strip()) if exit_path.exists() else -1
```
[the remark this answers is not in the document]
hari-kuriakose: Fixed in 2bb118dc / 52e22210 — exit codes are folded into `overall_exit` independently of junit presence, so a group with `exit.txt` but no `junit.xml` (segfault/OOM) is no longer hidden; its non-zero exit fails the build.

tests/rig/runtime.py
```diff
@@ -0,0 +1,217 @@
...
+
+def _responds(url: str) -> bool:
+    try:
+        import requests
+    except ImportError:
+        return True  # can't check; assume reachable to avoid spurious failure
+    try:
+        resp = requests.get(url, timeout=2)
+        return resp.status_code < 500
+    except requests.RequestException:
```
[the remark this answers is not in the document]
hari-kuriakose: Intentional for now — `< 500` is the liveness floor (tolerates 401/404 on an unseeded stack). A stricter readiness contract lands when the smoke suite grows real endpoint assertions.

tests/rig/selection.py
```diff
@@ -0,0 +1,89 @@
...
+        out = subprocess.run(
+            ["git", "diff", "--name-only", f"{base}...HEAD"],
+            check=True,
+            capture_output=True,
+            text=True,
+            cwd=REPO_ROOT,
+            env={**os.environ, "GIT_OPTIONAL_LOCKS": "0"},
+        ).stdout
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        return set()
```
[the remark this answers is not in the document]
hari-kuriakose: Fixed in 52e22210 — `--changed-only` now detects `HEAD == base` (push-to-main) and falls back to `HEAD~1...HEAD`, surfaces git stderr on failure, and degrades to an empty set (not a crash) when git is unavailable.

tests/rig/cli.py
```diff
@@ -0,0 +1,689 @@
...
+def cmd_platform(args: argparse.Namespace) -> int:
+    runtime = pick_runtime(args.runtime)
+    if args.action == "up":
+        try:
+            endpoints = runtime.up()
+        except Exception:
+            # If up() raised mid-way (e.g. one testcontainer started, the next
+            # failed), down() cleans up the partial stack.
+            runtime.down()
+            raise
```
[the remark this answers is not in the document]
hari-kuriakose: Fixed in 2bb118dc — `runtime.down()` in the `finally` is wrapped in its own try/except so a teardown error can't mask the original `up()` failure. Covered by `test_cmd_run_teardown_failure_does_not_mask_up_failure`.

tests/rig/cli.py
hari-kuriakose: Acknowledged — partially addressed. The intent (let pytest surface the real missing-module error) stands, but you're right that a silent install failure is indistinguishable from a pass. The `uv` calls inherit stdout/stderr (not captured), so the warnings do reach the log. Capturing returncode + aborting the group with a synthetic errors=1 junit is a clean follow-up; tracking it rather than expanding this commit further.

tests/rig/cli.py
hari-kuriakose: Fixed in 52e22210 — `_write_synthetic_junit` now `xml.sax.saxutils.escape(group_name, {'"': '&quot;'})` for both `name=` and `classname=`. Covered by `test_synthetic_junit_escapes_group_name`. Thanks — clean catch.

tests/rig/cli.py
hari-kuriakose: Fixed in 52e22210 — the `_write_synthetic_junit` call is now wrapped in `try/except OSError` (logs + continues) matching the exit.txt write, so a read-only/full reports dir no longer orphans completed groups.

tests/rig/cli.py
hari-kuriakose: Fixed in 52e22210 — after appending each result, `if (result.errors or result.failed) and overall_exit == 0: overall_exit = 1`. The report's ❌ and CI's exit code can no longer disagree.

tests/rig/cli.py
hari-kuriakose: Deferred (follow-up). Correct that `subprocess.run(timeout=)` only kills the direct child, leaving xdist grandchildren orphaned on timeout. The fix (Popen + start_new_session + killpg SIGTERM/SIGKILL + pipe drain) is a focused refactor I'd rather land on its own with a test than bundle into the review-fix commit. Tracking it; it only bites on the rare timeout path.

tests/rig/coverage.py
hari-kuriakose: Fixed in 52e22210 — the three coverage subprocesses now run with `timeout=300` via a shared `_run_coverage` helper; on `TimeoutExpired` it logs and returns (matching the 'errors logged not raised' policy).

tests/rig/runtime.py
hari-kuriakose: Deferred (small follow-up). Agreed — `LocalRuntime.up()` should do a quick `_responds` sweep so 'forgot to start the platform' surfaces at the rig layer instead of deep in a fixture. Low-risk to add; tracking with the other runtime hardening.

tests/rig/runtime.py
hari-kuriakose: Acknowledged — `ComposeRuntime.down()` uses `_run(check=False)` and discards the rc. `_run` now raises with the command tail on `CalledProcessError` when `check=True`; for the `check=False` teardown path I'll add an rc-inspect + stderr print as a follow-up so state-leak warnings aren't lost. Not blocking.

tests/rig/groups.py
hari-kuriakose: Fixed in 52e22210 — `GroupDefinition.__post_init__` coerces `env` to `types.MappingProxyType(dict(env))`, so the frozen record is genuinely immutable. Covered by `test_group_env_is_frozen`. (`slots=True` across the frozen dataclasses is a nice follow-up but interacts with the `_by_id`/MappingProxy `object.__setattr__` patterns, so leaving for a dedicated change.)

tests/rig/critical_paths.py
hari-kuriakose: Fixed in 52e22210 — `CriticalPathRegistry.__post_init__` now raises `ValueError(f'duplicate critical-path ids: {dupes}')`. Covered by `test_duplicate_path_ids_rejected`.

tests/rig/critical_paths.py
hari-kuriakose: Fixed in 52e22210 — `CriticalPathStatus.__post_init__` enforces covered ⇒ non-empty `covering_groups_run` and gap|regression ⇒ empty. The contradiction is now unrepresentable. Covered by `test_critical_path_status_rejects_contradictions`.

.pre-commit-config.yaml
hari-kuriakose: Fixed in 52e22210 — the hook now runs `$(command -v python3 || command -v python) -m tests.rig run ...`, mirroring the combine-script fix. Thanks — this would've been a silent 127 for opt-in users on python3-only machines.

tests/rig/selection.py
hari-kuriakose: Fixed in 52e22210 — `_groups_for_changed_paths` detects `HEAD == base` (the push-to-main case) via `git rev-parse` and falls back to `HEAD~1...HEAD` with a stderr note that the selector is PR-oriented. CLI behavior is now sane on main.

tests/rig/cli.py
hari-kuriakose: Fixed in 52e22210 — reworded the docstring: the session id proves the rig ran; URL ownership is intentionally cooperative via `setdefault` so a developer's pre-set `UNSTRACT_*_URL` wins. No longer contradicts the README.

tests/rig/cli.py
hari-kuriakose: Fixed in 52e22210 — added `test_cmd_report_re_aggregates_existing_junit` (populates a group's junit+exit, asserts all three summary artifacts). On the `scope_groups=None` asymmetry: it's intentional — `cmd_report` re-aggregates an arbitrary set of already-written junit files with no notion of 'what was meant to run this invocation', so scoping doesn't apply; only `cmd_run` knows the selection. Noted the rationale on the call.

tests/rig/cli.py
```diff
@@ -0,0 +1,738 @@
...
+    if endpoints is not None:
+        env.setdefault("UNSTRACT_BACKEND_URL", endpoints.backend_url)
+        env.setdefault("UNSTRACT_PROMPT_SERVICE_URL", endpoints.prompt_service_url)
+        env.setdefault("UNSTRACT_PLATFORM_SERVICE_URL", endpoints.platform_service_url)
+        env.setdefault("UNSTRACT_RUNNER_URL", endpoints.runner_url)
+        # Stamp the run with a per-invocation sentinel so e2e tests can
+        # distinguish "rig brought the platform up" from "stale shell env
+        # leaked in". `setdefault` would let a leaked sentinel win, which
+        # defeats the purpose — set unconditionally.
+        env["UNSTRACT_RIG_SESSION_ID"] = _rig_session_id()
```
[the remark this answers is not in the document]
hari-kuriakose: Fixed in 4c89989f — `_execute_group` now `setdefault`s `UNSTRACT_X2TEXT_URL` into the group env alongside the other four service URLs.

tests/rig/runtime.py
```diff
@@ -0,0 +1,293 @@
...
+    except ImportError:
+        log.warning("`requests` not installed; skipping platform readiness probe")
+        return
+
+    targets = [
+        endpoints.backend_url.rstrip("/") + "/health/",
+        endpoints.prompt_service_url.rstrip("/") + "/health",
+        endpoints.platform_service_url.rstrip("/") + "/health",
+        endpoints.runner_url.rstrip("/") + "/health",
+    ]
```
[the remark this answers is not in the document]
hari-kuriakose: Fixed in 4c89989f — `_wait_ready` now includes `x2text_url`'s `/health` in the readiness probe, so `ComposeRuntime.up()` waits for x2text instead of returning early.

tests/rig/cli.py
```diff
@@ -0,0 +1,739 @@
...
+        print(f"SKIP {n} (optional + workdir/paths absent)")
+    # Scope of this invocation = every selected group AFTER dep-expansion
+    # (includes dependencies the user didn't ask for directly, and includes
+    # optional placeholders that were skipped). Used by evaluate() to
+    # distinguish "this critical path was meant to be exercised here but its
+    # group ran red" (regression) from "this path belongs to a tier we
+    # weren't running this time" (gap).
+    scope_groups = set(ordered)
+
+    reports_dir: Path = args.reports_dir
```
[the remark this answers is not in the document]
hari-kuriakose: Fixed in 3e6266af. `scope_groups` now derives from `set(runnable)` instead of `set(ordered)`, so skipped optional placeholders are excluded from scope — their critical paths classify as `gap` rather than `regression`, eliminating the cross-tier false positive. Added `test_cmd_run_excludes_missing_placeholders_from_scope` to lock the behavior in.

tests/rig/groups.py
```diff
@@ -0,0 +1,246 @@
...
+            f"`requires_platform` groups present ({sorted(platform_groups)}) "
+            f"but the platform gate {gate!r} is not defined. Either define it, "
+            f"or set `defaults.platform_gate_group` in groups.yaml."
+        )
+    for name in platform_groups:
+        if not _transitively_depends_on(name, gate, groups):
+            raise ValueError(
+                f"group {name!r} requires_platform but does not (transitively) "
+                f"depend on {gate!r}; add it to depends_on so smoke gates this group"
+            )
```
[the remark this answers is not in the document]
hari-kuriakose: Good catch — the docs overpromised. Fixed in 4edea4f6.

Rather than rush runtime skip-on-gate-failure into this foundation PR, I corrected the docs to describe what's actually enforced: `_validate_platform_groups_depend_on_gate` is a *structural* invariant (the dep edge to `e2e-smoke` must exist, so the manifest can't ship a platform test that bypasses smoke, and the gate runs first in topo order). The misleading "skip cleanly" sentence is gone from both the docstring and the `groups.yaml` comment.

Runtime skip-on-dep-failure is deferred to a follow-up, with a `TODO(runtime-gate-skip)` in `cmd_run` flagging the design questions it needs to answer first (does an `optional`/non-failing-exit gate cascade, and how far down transitive deps?). As you noted, it's zero-impact today since every platform dependent is `optional: true` — so this stays a documented, deliberate deferral rather than a silent gap.

tests/rig/coverage.py
```diff
@@ -0,0 +1,138 @@
...
+        target.unlink()
+
+    base = _coverage_base()
+    clean_env = _clean_env()
+    if not _run_coverage(
+        [*base, "combine", *[str(p) for p in files]],
+        cwd=reports_dir,
+        env=clean_env,
+    ):
+        return
```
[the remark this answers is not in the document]
hari-kuriakose: Fixed in 6917b77b — went with the carry-forward approach (your first option), with one correction worth noting.

`coverage combine` does **not** read a bare `.coverage` passed as an argument — it only consumes suffixed `.coverage.*` data files (a bare `.coverage` is the output target). Verified empirically: `coverage combine .coverage .coverage.integration-y` reports "Combined 1 file" and the `.coverage` data is dropped. So simply adding the existing `.coverage` to the `files` list (or `--keep`) wouldn't have worked.

Instead, before each combine the existing `reports/.coverage` is renamed to a suffixed `.coverage.__prior__` and fed back in as an input, so each tier merges into the running union. On success `combine` consumes `__prior__` along with the tier files (no stale accumulation); if `combine` fails after the rename, `__prior__` stays on disk and the next call's glob picks it up — no data loss either way.

Added `tests/rig/tests/test_coverage.py` (4 tests): cross-tier preservation, same-file line union, no-files no-op, and the idempotent `cmd_run` + `cmd_report` double-combine case where only prior data remains. The first test reproduces exactly the loss you described and now passes.

ritwik-g (approved):

tests/rig/coverage.py
```diff
@@ -0,0 +1,160 @@
...
+    """
+    files = sorted(reports_dir.glob(".coverage.*"))
+    target = reports_dir / ".coverage"
+    # Carry the prior tier's merged data forward. `coverage combine` ignores a
+    # bare `.coverage` arg (it's the output), so promote it to a suffixed
+    # input name. `__prior__` sorts before tier files but order is irrelevant
+    # to the union; the suffix just needs to match the `.coverage.*` glob.
+    if target.exists():
+        prior = reports_dir / ".coverage.__prior__"
+        target.rename(prior)
```
jaseemjaskp: **Unguarded `target.rename(prior)` breaks the module's "errors are logged, not raised" contract — crashes the rig *after* tests run and drops the real exit code + summary.**

`combine_and_report`'s docstring (L82-83) and `_run_coverage` (L134) both promise coverage failures are logged, never raised, so a coverage hiccup can't disturb the test run's exit code. Every coverage *subprocess* honors that. But this carry-forward `target.rename(prior)` is a raw filesystem op with no guard, and it raises `OSError` under exactly the conditions CI hits: a cross-device/overlay-mount rename (`EXDEV`), a read-only or full reports volume (`EROFS`/`ENOSPC`), or a `.coverage.__prior__` left locked by a crashed prior run.

Both call sites are unprotected and sit **outside** `cmd_run`'s try/finally — `cli.py:438` (after the per-group loop) and `cli.py:283` (`cmd_report`, the `report combine` path CI shells into). An `OSError` here propagates up and aborts the rig *after every test has already executed*, replacing the genuine test exit code with an unrelated coverage-plumbing traceback and skipping `write_summary`, so no `summary.md`/`summary.json` lands. That is precisely the failure the docstring says must never happen, and it's asymmetric: an `xml`/`html` failure degrades gracefully while a *rename* failure takes down the whole run.

Fix — degrade like every other coverage step (log + continue; on failure you simply lose the carried-forward prior tier's data, which is strictly better than dropping the exit code):
```python
if target.exists():
    prior = reports_dir / ".coverage.__prior__"
    try:
        target.rename(prior)
    except OSError as exc:
        log.warning(
            "could not carry prior coverage forward (%s): %s; "
            "this tier's combine will omit earlier tiers' data", prior, exc,
        )
    else:
        files = [prior, *files]
```

tests/critical_paths.yaml
```diff
@@ -0,0 +1,64 @@
...
+#   ❌ regression — a path that was ✅ on the cached main baseline is now not ✅
+#
+# We intentionally do NOT chase 100% coverage. Focus on filling these gaps first.
+version: 1
+
+paths:
+  - id: auth-login
+    description: "User can log in and obtain a session cookie."
+    entry: "POST /api/v1/auth/login"
+    covered_by: [e2e-smoke]
```
jaseemjaskp: **`auth-login` is a false-green: `covered_by: [e2e-smoke]`, but the smoke suite never logs in.**

`e2e-smoke` (`tests/e2e/smoke/test_smoke.py`) has exactly two tests — a session-sentinel check and `GET /health/ < 500`. Neither exercises `POST /api/v1/auth/login`. Yet this entry declares `covered_by: [e2e-smoke]`, so the moment `e2e-smoke` runs green, `auth-login` is reported ✅ *covered*. The adjacent `adapter-register-llm` entry (L21-24) honestly sets `covered_by: []` for exactly this reason — its comment states "e2e-smoke only hits /health/" — so the two declarations directly contradict each other.

The consequence defeats the registry's whole purpose: a real break in login can never classify as ❌ regression, because the baseline will record `auth-login` as covered (by a group that doesn't test it). The regression detector is silently disabled for this path.

Fix: set `covered_by: []` (mirroring `adapter-register-llm`) so `auth-login` reports ⚠️ gap until a genuine `tests/e2e/` login test covers it.

.github/workflows/ci-test-e2e.yaml
```diff
@@ -0,0 +1,85 @@
...
+        # See ci-test.yaml for the rationale on namespacing per-workflow.
+        uses: actions/cache@v5
+        with:
+          path: reports/previous-summary.json
+          key: unstract-test-baseline-e2e-main-${{ github.run_id }}
+          restore-keys: |
+            unstract-test-baseline-e2e-main-
+            unstract-test-baseline-e2e-
+
+      - name: Run e2e tier via docker compose
```
jaseemjaskp, line 52: **The compose e2e lane can't bring the stack up as wired — and this runs on every push-to-main and nightly, not just opt-in PRs, so it fails deterministically and recurrently.**

This step runs `tox -e e2e` with `UNSTRACT_E2E_RUNTIME=compose`, which drives `ComposeRuntime.up()` → `docker compose -f docker/docker-compose.yaml -f tests/compose/docker-compose.test.yaml up -d --wait` (no service filter, so it creates *every* service in the merged files). The test overlay only pins 4 services (backend, prompt-service, platform-service, runner) to `${UNSTRACT_TEST_VERSION:-latest}`. The base `docker/docker-compose.yaml` references `${VERSION}` **with no default** for the ~15 other services it also brings up (x2text-service, frontend, redis, db, ~11 worker-unified instances). This workflow never sets `VERSION` or `UNSTRACT_TEST_VERSION`, never builds or pulls images, and there's no `.env` on the runner — so those services resolve to invalid empty-tag references (e.g. `unstract/worker-unified:`) and `up --wait` errors out before any e2e test runs.

The `if:` at L18-20 gates only the *pull_request* path behind the `run-e2e` label; `push: main`, the nightly `schedule`, and `workflow_dispatch` all run unconditionally — so main and nightly go red every time.

Note also the overlay comment in `tests/compose/docker-compose.test.yaml:14` ("CI sets it to the build SHA so e2e runs against the freshly built image") is factually wrong — this CI does not set it. That comment will mislead the next maintainer.

Fix options: (a) add a build/tag step that exports `UNSTRACT_TEST_VERSION` (and `VERSION`) = the built SHA before `tox -e e2e`; or (b) constrain the compose up to the test-overlaid services so unset-`${VERSION}` services aren't pulled; or (c) gate the workflow off (or `continue-on-error`) until the image-build wiring lands. Either way, correct the overlay comment to match reality.

jaseemjaskp (commented):
Second-pass review (automated multi-agent sweep). Three new findings beyond the already-addressed threads — one coverage error-handling contract gap, one false-green in the critical-path registry, and one CI wiring defect in the e2e lane. Details inline. Other agent findings (a few latent type/test-coverage hardening items) were below the bar for this round and are summarized to the author separately.

tests/rig/coverage.py
hari-kuriakose: Fixed in c0698937 — wrapped the rename in try/except OSError, log + fall through. The warning message explicitly notes that this tier's combine will omit earlier tiers' data, so the degradation is visible in CI logs without dropping the exit code or summary. Locked in by `test_combine_swallows_rename_oserror` which monkeypatches Path.rename to raise OSError(EXDEV) and asserts combine_and_report returns normally. All 5 coverage self-tests still pass.

tests/critical_paths.yaml
hari-kuriakose: Fixed in c0698937 — `auth-login` now `covered_by: []` with a comment mirroring `adapter-register-llm`'s honest-declaration style. Manifest validates clean (`tox -e rig -- validate` → OK — 16 groups, 10 critical paths).

.github/workflows/ci-test-e2e.yaml
hari-kuriakose: Fixed in c0698937 — went with option (c): gated the workflow off push:main + nightly until image-build wiring lands. `on:` now only fires on label-gated PRs + `workflow_dispatch`, with a header comment documenting what unblocks restoring `push: main` and `schedule:`. Also corrected the overlay comment in `tests/compose/docker-compose.test.yaml:14` — it no longer claims CI pins the SHA today; says CI will pin once image-build wiring lands.

Went with gate-off rather than `continue-on-error` deliberately: continue-on-error would still burn CI minutes every push and turn the workflow into background noise everyone learns to ignore.

chandrasekharan-zipstack (approved):

PR #1979, opened 2026-05-21
ReverseMerge: V0.163.4 hotfix

pk-zipstack:
## What

Back-merge of `v0.163.4-hotfix` (released as `v0.163.5`) into `main`.

Three hotfix commits sit on `v0.163.4-hotfix` ahead of `main`:

- `2dab564b` [HOTFIX] Bump litellm to 1.83.10 from PyPI to clear CVE-2026-42208 (#1976) — **new, needs back-merge**
- `c34f2409` [HOTFIX] Add IAM Role / Instance Profile auth mode to AWS Bedrock adapter (#1944) — already on `main` via the squash back-merge in #1946
- `e6bb412e` [HOTFIX] Use importlib.util.find_spec for pluggable worker discovery (#1918) — already on `main` via #1923 / #1946

Net new content this PR carries to `main` is just the litellm CVE fix.

## Why

Per the [Hotfix Deployment Guide](https://zipstack.atlassian.net/wiki/spaces/PT/pages/491880450/Hotfix+Deployment+Guide) Step 9, hotfix branches must be back-merged to `main` after production deployment so the fixes carry forward into future releases.

## How

- Use **rebase merge** (preferred) or **squash merge** — not the default merge commit. Per the doc, click the merge dropdown and select the right option.

## Expected merge conflicts

Local `git merge-tree` flagged conflicts in these files:

- `unstract/sdk1/src/unstract/sdk1/adapters/base1.py`
- `unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/bedrock.json`
- `unstract/sdk1/src/unstract/sdk1/adapters/embedding1/static/bedrock.json`
- `unstract/sdk1/tests/test_bedrock_adapter.py`
- `uv.lock`

These conflicts are **not from the litellm change** — they come from #1944 (IAM Role / Instance Profile) on the hotfix branch overlapping with `main`'s newer #1952 (`UN-3152 [FEAT] Add AWS_BEARER_TOKEN_BEDROCK auth to Bedrock LLM and Embedding adapters`). Main's version integrates both auth modes (`access_keys` + `iam_role` + `bearer_token`) and supersedes the hotfix-branch version.

**Resolution guidance:** for the four bedrock files, keep `main`'s version (it already has the IAM Role content from the prior back-merge in #1946 plus the bearer-token addition from #1952). For `uv.lock`, regenerate after resolving.

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why.

[...]

PR #1976, opened 2026-05-20
[HOTFIX] Bump litellm to 1.83.10 from PyPI to clear CVE-2026-42208

pk-zipstack:
> **Hotfix for cloud `v0.159.3` → OSS `v0.163.4`**
> Following the [Hotfix Deployment Guide](https://zipstack.atlassian.net/wiki/spaces/PT/pages/491880450/Hotfix+Deployment+Guide). This is **Scenario 2** (OSS-only changes). After merge, please cut a new GitHub release (e.g. `v0.163.5`) targeting the `v0.163.4-hotfix` branch, then back-merge to `main` per Step 9.

## Summary

- Bumps `litellm` from 1.82.3 (Zipstack GitHub fork) to **1.83.10** from PyPI, clearing CVE-2026-42208 (SQL injection in litellm proxy auth — affects 1.81.16–1.83.6) as reported by a customer scanner.
- We do not use `litellm.proxy`, but scanners flag the installed package regardless. 1.83.10 is the exact version recommended in the [upstream advisory](https://docs.litellm.ai/blog/cve-2026-42208-litellm-proxy-sql-injection) (`v1.83.10-stable`) and the smallest jump that clears the CVE range while keeping `python-dotenv==1.0.1` compatible across the workspace.
- Reverses the temporary GitHub-fork sourcing from #1873 now that the PyPI quarantine is over.
- Bumps `tiktoken` 0.9 → 0.12 (required by litellm's pin) — only constraint that needed to move.

## Cohere embed timeout patch

`unstract/sdk1/src/unstract/sdk1/patches/litellm_cohere_timeout.py` (added in #1848) is version-guarded to one exact litellm release. Before bumping the guard:

- Diffed upstream `litellm/llms/cohere/embed/handler.py` at `v1.82.3`, `v1.83.10-stable`, and `v1.83.14-stable` via `gh api` — **byte-identical** across all three.
- Confirmed the bug is still present at 1.83.10 (`client.post()` doesn't forward `timeout` in either `embedding()` or `async_embedding()`).
- Confirmed BerriAI/litellm#14635 is still **OPEN** upstream.
- 6/6 patch unit tests (`tests/patches/test_litellm_cohere_timeout.py`) pass on 1.83.10 — the suite asserts `client.post()` receives the timeout kwarg, so green = patch is binding and behaving against the new module.

Guard bumped 1.82.3 → 1.83.10; patch keeps doing real work.

## Other cleanup from #1873

[...]

PR #1946, opened 2026-05-06
ReverseMerge: V0.163.2 hotfix

pk-zipstack:
## What

Back-merge of `v0.163.2-hotfix` into `main`. Brings two hotfix commits already merged into the hotfix branch back onto `main`:

- `c34f2409` [HOTFIX] Add IAM Role / Instance Profile auth mode to AWS Bedrock adapter (#1944)
- `e6bb412e` [HOTFIX] Use importlib.util.find_spec for pluggable worker discovery (#1918)

## Why

Per the [Hotfix Deployment Guide](https://zipstack.atlassian.net/wiki/spaces/PT/pages/491880450/Hotfix+Deployment+Guide) Step 9, hotfix branches must be back-merged to `main` after production deployment so the fixes carry forward into future releases.

## How

- Use **rebase merge** (preferred) or **squash merge** — not the default merge commit. Per the doc, click the merge dropdown and select the right option.

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why.

No. Both commits were independently reviewed and merged into the hotfix branch via their own PRs (#1918, #1944). This PR only forwards them to `main` without re-introducing changes.

## Database Migrations

- None

## Env Config

- None

## Relevant Docs

- [Hotfix Deployment Guide](https://zipstack.atlassian.net/wiki/spaces/PT/pages/491880450/Hotfix+Deployment+Guide)

## Related Issues or PRs

- #1918 (already merged into `v0.163.2-hotfix`)
- #1944 (already merged into `v0.163.2-hotfix`)

## Dependencies Versions

- No changes

## Notes on Testing

Tested in the originating hotfix PRs. No additional testing required for the back-merge itself.

## Screenshots

N/A

## Checklist

I have read and understood the [Contribution Guidelines](https://docs.unstract.com/unstract/contributing/unstract/).

PR #1980, opened 2026-05-21
ReverseMerge: V0.163.4 hotfix

pk-zipstack:
## What

Back-merge of `v0.163.4-hotfix` (released as `v0.163.5`) into `main`, with merge-conflict resolution carried on this branch (the hotfix branch itself is push-protected, so the merge commit lives here).

Net new content for `main`:
- `2dab564b` [HOTFIX] Bump litellm to 1.83.10 from PyPI to clear CVE-2026-42208 (#1976)

The other two hotfix commits on `v0.163.4-hotfix` are already in `main` via the squash back-merge #1946:
- `c34f2409` [HOTFIX] Add IAM Role / Instance Profile auth mode to AWS Bedrock adapter (#1944)
- `e6bb412e` [HOTFIX] Use importlib.util.find_spec for pluggable worker discovery (#1918)

## Why

Per the [Hotfix Deployment Guide](https://zipstack.atlassian.net/wiki/spaces/PT/pages/491880450/Hotfix+Deployment+Guide) Step 9, hotfix branches must be back-merged to `main` after production deployment so the fixes carry forward into future releases.

## How

- Use **rebase merge** (preferred) or **squash merge** — not the default merge commit. Per the doc, click the merge dropdown and select the right option.

## Conflict resolution

Five files conflicted during `git merge origin/main` on `v0.163.4-hotfix`:

| File | Resolution |
| --- | --- |
| `unstract/sdk1/src/unstract/sdk1/adapters/base1.py` | Took main — main's `_BEDROCK_AUTH_TYPES = {None, "access_keys", "iam_role", "bearer_token"}` is a strict superset of the hotfix branch's IAM-Role-only version |
| `unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/bedrock.json` | Took main — schema includes all three auth modes |
| `unstract/sdk1/src/unstract/sdk1/adapters/embedding1/static/bedrock.json` | Took main — same reasoning |
| `unstract/sdk1/tests/test_bedrock_adapter.py` | Took main — covers all three auth modes |
| `uv.lock` | Took main's, then re-ran `uv lock` to apply our litellm 1.83.10 pin and switch the source back to PyPI registry |

Conflicts came from #1944 (hotfix's IAM Role) overlapping with main's newer #1952 (`UN-3152 [FEAT] Add AWS_BEARER_TOKEN_BEDROCK auth`). My litellm changes from #1976 did NOT cause any conflicts.

## Validation

Ran the relevant test suites against the merged tree:

- `unstract/sdk1/tests/patches/test_litellm_cohere_timeout.py` → **6/6 passed** (litellm patch still binds on 1.83.10)
- `unstract/sdk1/tests/test_bedrock_adapter.py` → **37/37 passed** (bedrock auth modes all working with main's version)

No. The only new content reaching `main` is the litellm 1.82.3 → 1.83.10 bump (released to production as `v0.163.5`). Bedrock-adapter behavior on `main` is unchanged — we kept main's version of every conflicting file.

## Database Migrations

- None

## Env Config

- None

## Relevant Docs

- [Hotfix Deployment Guide](https://zipstack.atlassian.net/wiki/spaces/PT/pages/491880450/Hotfix+Deployment+Guide)
- [CVE-2026-42208 advisory](https://docs.litellm.ai/blog/cve-2026-42208-litellm-proxy-sql-injection)

## Related Issues or PRs

- #1976 (litellm CVE hotfix — merged into `v0.163.4-hotfix`, released as v0.163.5)
- #1979 (previous back-merge PR — closed, conflicts unresolved; superseded by this PR)
- #1946 (prior back-merge that carried #1918 and #1944 to main)
- #1952 (`AWS_BEARER_TOKEN_BEDROCK` on main — source of the bedrock-file conflicts)

## Dependencies Versions

- `litellm`: `1.82.3` (Zipstack GitHub fork) → `1.83.10` (PyPI)
- `tiktoken`: `~=0.9.0` → `~=0.12.0` (required by litellm 1.83.10's pin)

## Notes on Testing

Tested in #1976 and validated in production via the `v0.163.5` release. Bedrock adapter test suite re-run locally post-merge — all 37 tests pass.

## Screenshots

N/A

## Checklist


unstract/sdk1/src/unstract/sdk1/patches/litellm_cohere_timeout.py
```diff
@@ -4,8 +4,8 @@
 receive a `timeout` parameter but don't forward it to client.post(),
 causing "Connection timed out after None seconds" errors.
 
-Affected litellm version: 1.82.3 (also present on latest main as of
-2026-03-10).
+Affected litellm version: 1.83.10 (bug confirmed in every release between
+1.82.3 and 1.83.14-stable as of 2026-05-20; upstream issue still open).
```
jaseemjaskp: **HIGH — Docstring claim is broader than the code actually enforces.**

The docstring now says the bug is "confirmed in every release between 1.82.3 and 1.83.14-stable", but the version guard on line 32 only activates the patch when the installed litellm version equals exactly `1.83.10`. A future maintainer reading this docstring would reasonably expect that bumping to e.g. 1.83.11 keeps the fix active; in reality it silently turns off (greptile already flagged that control flow on line 32 — different concern, same root cause).

Also, the parenthetical `as of 2026-05-20` is a fixed calendar date with no enforcement; the commit date already records this and it will rot fast.

Suggested rewrite:
```
Affected litellm version: this patch activates only on the pinned
version in pyproject.toml (currently 1.83.10 — see
_PATCHED_LITELLM_VERSION). The underlying bug has been observed
upstream across 1.82.3 → 1.83.14; any version mismatch logs and
skips so the upgrade is reviewed.
```
Drop the date phrase entirely.

unstract/sdk1/src/unstract/sdk1/patches/litellm_cohere_timeout.py
```diff
@@ -27,7 +27,7 @@
 # Only apply the patch on the exact litellm version it was written for.
 # Any other version (newer or older) skips the patch with a visible
 # warning so engineers know to verify compatibility.
-_PATCHED_LITELLM_VERSION = "1.82.3"
+_PATCHED_LITELLM_VERSION = "1.83.10"
 _litellm_version = importlib.metadata.version("litellm")
```
jaseemjaskp, line 33: **HIGH — Module-import-time crash on `PackageNotFoundError` / `InvalidVersion`.**

Neither `importlib.metadata.version("litellm")` nor `Version(...)` is wrapped:
- `importlib.metadata.version` raises `PackageNotFoundError` if litellm is missing or installed under a different distribution name (vendored copies, broken `*.dist-info`, some editable layouts).
- `Version(_litellm_version)` raises `InvalidVersion` if the installed string isn't PEP 440 (e.g. a locally-built `1.83.10+local`).

Either exception propagates out of the patch module and crashes `import unstract.sdk1.embedding` for every consumer with an opaque traceback far from the call site — much worse than the silent-skip path the guard is designed to produce.

Suggested fix:
```python
try:
    _litellm_version = importlib.metadata.version("litellm")
    _SKIP_PATCH = Version(_litellm_version) != Version(_PATCHED_LITELLM_VERSION)
except (importlib.metadata.PackageNotFoundError, InvalidVersion) as exc:
    _SKIP_PATCH = True
    logger.error(
        "litellm_cohere_timeout: could not resolve litellm version (%s); patch skipped",
        exc,
    )
```
The same try/except could also wrap the import block in the `else:` arm at lines 45–67, so a reshape of litellm internals on a future bump degrades gracefully instead of crashing every embedding consumer.

unstract/sdk1/src/unstract/sdk1/patches/litellm_cohere_timeout.py
```diff
@@ -68,7 +68,7 @@
 
     _DEFAULT_TIMEOUT = httpx.Timeout(None)
```
jaseemjaskp, line 71: **MEDIUM — Default `httpx.Timeout(None)` reproduces the very failure mode this patch fixes.**

When `_patched_embedding` is invoked without an explicit `timeout` (e.g. via litellm's own dispatcher), this default flows through to `client.post(..., timeout=httpx.Timeout(None))` on line 197 — that's the same `Connection timed out after None seconds` symptom the patch is meant to prevent.

This line is preserved verbatim from upstream litellm, so the bug is technically pre-existing — but it sits inside a file whose stated purpose is to keep cohere embeds from hanging on `None` timeouts. Two options:
1. Replace with a finite default (e.g. `httpx.Timeout(600.0)` to match Unstract's typical embed timeouts).
2. At the call sites on line 109 and 197, fall back to `litellm.request_timeout` when the incoming `timeout is None`.

If neither change is accepted to preserve the "verbatim copy + one change" contract, please note the residual risk in the module docstring.

unstract/sdk1/src/unstract/sdk1/patches/litellm_cohere_timeout.py
```diff
@@ -68,7 +68,7 @@
...
 
-    # Copied from litellm 1.82.3 cohere/embed/handler.py async_embedding().
+    # Copied from litellm 1.83.10 cohere/embed/handler.py async_embedding().
```
jaseemjaskp, line 73: **LOW — Version literal `1.83.10` is hardcoded in three places, creating rot risk.**

This comment and the matching one at line 139 both embed the literal `1.83.10`. Combined with the docstring (line 7) and `_PATCHED_LITELLM_VERSION` (line 30), there are now four places to keep in sync on the next bump — missing any one yields contradictory state.

Suggested rewrite (apply to line 139 too):
```
# Copied from litellm at _PATCHED_LITELLM_VERSION's
# cohere/embed/handler.py async_embedding().
# ONLY CHANGE vs. upstream: timeout forwarded on the client.post()
# call. If you re-copy from a newer litellm, re-verify this is
# still the only diff.
```
Makes the constant the single source of truth and adds an explicit instruction to verify the "ONLY CHANGE" invariant on re-sync.

jaseemjaskp (commented):
## Multi-agent review summary (Code Reviewer · Comment Analyzer · Silent Failure Hunter · Test Analyzer · Type Design · Code Simplifier)

The back-merge body is small (litellm 1.82.3 → 1.83.10 + tiktoken 0.9 → 0.12 + patch version constant + Dockerfile `git` removal + tool-image tag bumps). Findings concentrate on `unstract/sdk1/src/unstract/sdk1/patches/litellm_cohere_timeout.py`. Inline comments below cover findings on diff lines; the items below are out-of-diff but worth tracking before merge.

### Out-of-diff findings (cannot anchor inline)

1. **MEDIUM — Two stand-alone uv.locks still pin `litellm==1.82.3` (the CVE-flagged version).** `unstract/workflow-execution/uv.lock` (~L731) and `unstract/connectors/uv.lock` (~L1236) were not regenerated. Both exist on `main` too — so this is pre-existing on `main`, not introduced by this PR — but since the explicit purpose of this back-merge is to carry the CVE-2026-42208 fix forward, regenerating these locks here would close the loop. Fix: run `uv lock --upgrade-package litellm` in each of those package dirs and include the updated lockfiles, OR open a follow-up issue noting these locks are runtime-unused.

2. **HIGH — Test gap: no test fails on litellm version drift.** `_PATCHED_LITELLM_VERSION` (line 30) is intended to force re-verification on bump, but every existing test in `unstract/sdk1/tests/patches/test_litellm_cohere_timeout.py` calls `_patched_embedding` directly and bypasses the guard. A bump to 1.83.11 would log a warning, silently disable the patch, and CI would still pass. Add `test_pinned_version_matches_installed_litellm()` asserting `importlib.metadata.version("litellm") == _PATCHED_LITELLM_VERSION`.

3. **HIGH — Test gap: no upstream-signature compatibility test.** The patched functions are verbatim copies (+ one kwarg) of upstream `litellm.llms.cohere.embed.handler.{embedding,async_embedding}`. If upstream renames a kwarg or removes a positional, the monkey-patched callable becomes a `TypeError` at first Bedrock cohere call. Add `inspect.signature(...)`-based parity tests against the upstream symbols.

4. **MEDIUM — Async path untested.** `_patched_async_embedding` (patch lines 74–137) and the `aembedding=True` dispatch branch (patch lines 164–182) are only covered by binding-identity assertions; no test exercises timeout forwarding through the async client.

5. **MEDIUM — Partial-patch application risk** (patch file, lines 215–217, outside the diff). The three monkey-patch assignments run sequentially with no try/except. If `_bedrock_embed.cohere_embedding = _patched_embedding` fails after the two cohere-handler assignments succeed (e.g. litellm refactors the bedrock module path), the cohere handler is patched but bedrock keeps calling the unpatched embedding — "None seconds" timeouts return intermittently with no clear signal. Consider wrapping the three assignments in try/except, reverting on partial failure, and emitting an error-level log.

6. **LOW — Skip-patch log severity.** When the version guard skips (today's log at line 34), users hit the timeout bug their patch was meant to prevent. `logger.warning` will get filtered by most aggregators. Consider `logger.error` so it surfaces in Sentry. (Adjacent to but distinct from greptile's existing comment on line 32 which flagged the silent-skip control flow.)

7. **NIT — `unstract/sdk1/src/unstract/sdk1/utils/callback_manager.py` has no tests around `tiktoken.encoding_for_model(...)` (`callback_manager.py:102,107`), so the 0.9 → 0.12 tiktoken bump has no CI safety net. Low risk; tiktoken's encode API has been stable.

### Verified clean

- Dockerfile `git` removal (`docker/dockerfiles/worker-unified.Dockerfile`, `tools/{classifier,structure,text_extractor}/Dockerfile`) is safe — no remaining `git+https` uv sources in any updated lockfile.
- Tool image tag bumps consistent across `sample.env`, `properties.json`, and `public_tools.json`.
- `litellm==1.83.10` pin in `unstract/sdk1/pyproject.toml:30` matches `_PATCHED_LITELLM_VERSION` so the guard activates on this build.
- `tiktoken~=0.12.0` matches litellm 1.83.10's transitive requirement.
- No new types introduced (type-design analyzer found nothing actionable).
- Code-simplifier: no meaningful simplifications recommended — most candidates are inside the verbatim-copy regions where the "exactly one change vs. upstream" invariant is more valuable than micro-cleanup.

unstract/sdk1/src/unstract/sdk1/patches/litellm_cohere_timeout.py
```diff
@@ -27,7 +27,7 @@
 # Only apply the patch on the exact litellm version it was written for.
 # Any other version (newer or older) skips the patch with a visible
 # warning so engineers know to verify compatibility.
-_PATCHED_LITELLM_VERSION = "1.82.3"
+_PATCHED_LITELLM_VERSION = "1.83.10"
 _litellm_version = importlib.metadata.version("litellm")
 _SKIP_PATCH = Version(_litellm_version) != Version(_PATCHED_LITELLM_VERSION)
```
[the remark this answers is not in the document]
praveen-formido: Intentional design — the silent-skip with logged warning is the patch's existing contract (see lines 26-29). Pin in pyproject.toml is `==1.83.10`, so a future bump without a matching guard update becomes a deliberate review trigger rather than an accidental footgun. Keeping as-is.

unstract/sdk1/src/unstract/sdk1/patches/litellm_cohere_timeout.py
praveen-formido: Addressed in 7386c269 — docstring now points at `_PATCHED_LITELLM_VERSION` as the single source of truth and drops the rot-prone `2026-05-20` calendar date. Thanks for the suggested rewrite.

unstract/sdk1/src/unstract/sdk1/patches/litellm_cohere_timeout.py
praveen-formido: Skipping — `litellm` is a hard dependency in `unstract/sdk1/pyproject.toml`, so `PackageNotFoundError` cannot fire in any working install of `unstract-sdk1`. `InvalidVersion` would require a non-PEP-440 local build (e.g. `1.83.10+local`), which is out-of-band territory where the engineer is already verifying things manually. Per our guideline (don't add validation for scenarios that can't happen), I'd rather not wrap it for the hotfix. Happy to revisit on `main` if you have a concrete environment where this bit you.

unstract/sdk1/src/unstract/sdk1/patches/litellm_cohere_timeout.py
praveen-formido: Skipping — the function body is preserved verbatim from upstream `litellm/llms/cohere/embed/handler.py` per the patch's stated 'verbatim copy + one change (timeout-forwarding)' contract; changing the default breaks the bytewise comparability we rely on to re-verify the patch against upstream on each litellm bump. In practice `unstract.sdk1.embedding` always passes a concrete `request_timeout`, so the residual risk isn't reachable from our call sites today. Logging this rationale on the thread rather than adding a docstring note (option 3 in your comment) to keep the hotfix diff minimal.

unstract/sdk1/src/unstract/sdk1/patches/litellm_cohere_timeout.py
praveen-formido: Skipping — LOW severity polish. `_PATCHED_LITELLM_VERSION` is the constant that controls behavior; the line-71/139 comments are documentation pointers for a future maintainer re-copying from upstream. Even if they drift, runtime behavior stays correct (the constant is what the guard reads). Accepting the rot risk to keep the hotfix diff minimal.

jaseemjaskp (approved):

PR #1982, opened 2026-05-21
[FIX] Scope user lookups to current env's OIDC connection via plugin-driven filter

jaseemjaskp:
## What

- Adds a pluggable `UserFilterRegistry` in core that plugins can populate at startup to scope `User` / `OrganizationMember` queries.
- Adds an Auth0-plugin filter (in private overlay) that scopes queries by `oidc|<AUTH0_OIDC_CONNECTION>|` prefix when the setting is defined — leaving behavior unchanged when it's unset/empty.
- Hardens `add_user_role` / `remove_user_role` to raise a clear `AmbiguousUserException` (409) instead of a silent 500 when a lookup matches multiple rows.

## Why

When one Auth0 organization is wired to multiple OIDC connections (one per environment, e.g. `enterprise-qa` for QA + `enterprise` for prod), Auth0 issues distinct `user_id`s for the same email across connections. Unstract keys `User` rows on `(email, user_id)`, so each connection creates a separate `User` row in the same DB. Two visible symptoms on the affected QA on-prem env:

1. **Manage Users shows duplicate rows** — same email listed twice with different roles (`unstract_user` and `unstract_admin`), because both `User` rows have their own `OrganizationMember`.
2. **Affected users couldn't access the User Management page** after being promoted to admin. Promotion landed on one `User` row (`oidc|enterprise|…`) while their QA login resolved to the other (`oidc|enterprise-qa|…`), so `is_admin` returned false and the route guard blocked them.

The fix scopes all user lookups to the connection that belongs to the current environment.

## How

- **`backend/account_v2/user_filter_registry.py`** *(new)* — a small registry exposing `register(fn)` and `apply(qs, kind)` where `kind ∈ {"user", "org_member"}`. With no filters registered, `apply` is an identity function.
- **`backend/account_v2/user.py`** — `UserService.get_user_by_email` / `get_user_by_user_id` / `get_user_by_id` route through `UserFilterRegistry.apply(..., "user")`. `.get()` becomes `.filter().first()` so a scoped match returns one row instead of raising `MultipleObjectsReturned`.
- **`backend/tenant_account_v2/organization_member_service.py`** — same treatment for `get_user_by_email` / `get_user_by_user_id` / `get_user_by_id` / `get_members` / `get_members_by_role` / `get_members_by_user_email`, all with `kind="org_member"`.
- **`backend/account_v2/authentication_controller.py`** — new `_resolve_unique_member_by_email` helper used by `add_user_role` and `remove_user_role`. Raises `AmbiguousUserException` when more than one row matches after filters apply, surfacing configuration issues instead of silently picking one.
- **`backend/account_v2/custom_exceptions.py`** — new `AmbiguousUserException` (HTTP 409) with a clear error message.

The plugin filter itself lives in the private overlay:
- `backend/plugins/authentication/auth0/apps.py` registers a filter via `Auth0PluginConfig.ready()` if `AUTH0_OIDC_CONNECTION` is set.
- `backend/backend/settings/on-prem.py` adds the setting binding.

- **OSS:** no change. The `plugins.authentication.auth0` plugin is not in OSS `INSTALLED_APPS`, so the filter is never registered. `UserFilterRegistry.apply()` with no filters is a pure identity function.
- **Cloud:** no change. The plugin loads in cloud, but `cloud.py` does not define `AUTH0_OIDC_CONNECTION`, so `Auth0PluginConfig.ready()` returns early without registering a filter.
- **On-prem single-IdP deployments (incl. the documented Azure AD OIDC setup):** no change as long as `AUTH0_OIDC_CONNECTION` is left unset or empty (`""` short-circuits the same as `None`).
- **On-prem multi-connection-per-Auth0-org deployments:** filter activates per environment, duplicate cross-env rows are scoped out of Manage Users, and role assignment lands on the correct identity.

One semantic shift to be aware of: lookups in `UserService` / `OrganizationMemberService` previously raised `MultipleObjectsReturned` (uncaught → 500) on duplicate matches. They now return the first match. For role-change endpoints this is replaced by an explicit 409 (`AmbiguousUserException`), which is strictly better than the old 500. For other lookup callers the behavior is rare-but-tolerated (DB unique constraints make duplicates unreachable in practice).

## Database Migrations

- None.

## Env Config

- New optional setting (private overlay only): `AUTH0_OIDC_CONNECTION`. **Opt-in.** Set to the Auth0 OIDC connection name for the current environment (e.g. `enterprise-qa` in QA, `enterprise` in prod). Leave unset/empty for single-IdP deployments — including the documented Azure AD OIDC setup.
- Warning: once set, the filter is deployment-wide. Any `User` whose `user_id` does not start with `oidc|<value>|` disappears from queries — including pre-SSO db-auth users (`auth0|…`), SAML users (`samlp|…`), and users from other OIDC connections. Only enable when every active user authenticates via the named OIDC connection.

## Relevant Docs

- Inline docstring on `UserFilterRegistry` describes the contract for plugin-registered filters.
- Inline comment on `AUTH0_OIDC_CONNECTION` in `on-prem.py` documents the opt-in nature and the consequences of enabling it.

## Related Issues or PRs

- Auth0 organization screenshot for the affected enterprise tenant: members list includes both `oidc|enterprise-qa|…` and `oidc|enterprise|…` `user_id`s for the same set of emails (one per environment connection).

## Dependencies Versions

- None.

## Notes on Testing

- [ ] Affected QA on-prem with `AUTH0_OIDC_CONNECTION=<qa-connection>` (e.g. `enterprise-qa`): Manage Users shows each affected email exactly once, with the QA-connection `user_id`.
- [ ] After admin re-runs "Add admin role" for the affected users, they can access the User Management page.
- [ ] On-prem dev with Azure AD OIDC + `AUTH0_OIDC_CONNECTION=""` (or unset): all users visible, no scoping applied.
- [ ] `add_user_role` with an ambiguous email returns 409 with the `AmbiguousUserException` message.
- [ ] OSS / cloud regression check: Manage Users and login behaviors unchanged.

## Screenshots

QA Manage Users showing duplicate rows for affected users (admin + user roles) — see ticket attachments.

Auth0 organization Members list showing two `user_id`s per affected email, one per connection prefix — see ticket attachments.

## Checklist


backend/account_v2/authentication_controller.py
```diff
@@ -410,7 +410,7 @@ def add_user_role(
     ) -> str | None:
         admin: User = request.user
         admin_user = OrganizationMemberService.get_user_by_id(id=admin.id)
-        user = OrganizationMemberService.get_user_by_email(email=email)
+        user = self._resolve_unique_member_by_email(email)
```
jaseemjaskp: **[P1] Admin-row may be silently `None` after the filter** — `admin_user = OrganizationMemberService.get_user_by_id(id=admin.id)` on line 412 now runs through `UserFilterRegistry`. If the IdP filter is misconfigured for the *admin's own row* (e.g., `AUTH0_OIDC_CONNECTION` set to the wrong env), this returns `None` and is then passed into `auth_service.add_organization_user_role(None, ...)` on line 415 — surfacing as an AttributeError 500 deep in the auth service with no breadcrumb back to the filter. The new `AmbiguousUserException` covers the duplicate side; the inverse needs a guard too.

Also a **latent `IndexError`** here: line 422's `return current_roles[0]` is unconditional, but `if current_roles:` on 418 only guards the save — empty `current_roles` will crash. Same on line 441.

Suggested:
```python
admin_user = OrganizationMemberService.get_user_by_id(id=admin.id)
if admin_user is None:
    logger.error("Admin row hidden by IdP filter for user_id=%s", admin.id)
    raise AmbiguousUserException()  # or a dedicated 403/409
...
return current_roles[0] if current_roles else None
```

backend/account_v2/authentication_controller.py
```diff
@@ -442,6 +442,35 @@ def remove_user_role(
         else:
             return None
 
+    @staticmethod
+    def _resolve_unique_member_by_email(email: str) -> OrganizationMember | None:
+        """Resolve a single OrganizationMember by email after registered filters.
+
+        Raises AmbiguousUserException if more than one row matches — that
+        indicates the environment's IdP filter is missing or misconfigured
+        (e.g., AUTH0_OIDC_CONNECTION not set when multiple IdP connections
```
jaseemjaskp: **[P2] Docstring leaks plugin-specific env var into core** — the parenthetical `AUTH0_OIDC_CONNECTION not set when multiple IdP connections share the same Auth0 org` hard-codes a single plugin's setting name into core code. The whole point of the registry is that core stays plugin-agnostic. Generalise to e.g. `(e.g., the active IdP scoping filter is not registered or is misconfigured for this environment)` and keep the Auth0 specifics in the commit message / PR description.

backend/account_v2/authentication_controller.py
```diff
@@ -442,6 +442,35 @@ def remove_user_role(
...
+        share the same Auth0 org), and silently picking one row would lead
+        to role changes landing on the wrong identity.
+        """
+        from account_v2.user_filter_registry import UserFilterRegistry
```
jaseemjaskp: **[P2] Local imports without a cycle to break** — `user_filter_registry` is a leaf module (only Django + stdlib), and `custom_exceptions` (imported on line 469) is already imported at the top of this file. Move both to module-level imports. Local imports tucked inside a hot path obscure the dependency graph for no benefit here.

backend/account_v2/user.py
```diff
@@ -41,17 +42,14 @@ def update_user(self, user: User, user_id: str) -> User:
...
+        )
+        return qs.first()
 
     def get_user_by_user_id(self, user_id: str) -> Any:
-        try:
-            return User.objects.get(user_id=user_id)
-        except User.DoesNotExist:
-            return None
+        qs = UserFilterRegistry.apply(User.objects.filter(user_id=user_id), "user")
+        return qs.first()
```
jaseemjaskp: **[P1] Silent multi-match — complementary to greptile's P1 on line 64** — `get_user_by_user_id` previously used `User.objects.get(user_id=user_id)`, which raises `MultipleObjectsReturned` on duplicates. That's exactly the failure mode this PR is built to surface. With `.first()`, two-or-more rows after filtering are silently collapsed to one — and which one is undefined (`.first()` without `order_by()` is order-by-PK on some backends but not guaranteed). Same on line 48 (`get_user_by_email`).

Mirror `_resolve_unique_member_by_email`:
```python
qs = UserFilterRegistry.apply(User.objects.filter(user_id=user_id), "user")
rows = list(qs[:2])
if len(rows) > 1:
    logger.error("Ambiguous User lookup for user_id=%s", user_id)
    raise AmbiguousUserException()
return rows[0] if rows else None
```
Greptile's existing P1 on line 64 covers the inverse "silently hides" case for the PK lookup; this is the symmetric "silently picks one of two" case for `user_id` and `email` lookups.

backend/account_v2/custom_exceptions.py
```diff
@@ -56,3 +56,11 @@ class Forbidden(APIException):
 class UserAlreadyAssociatedException(APIException):
     status_code = 400
     default_detail = "User is already associated with one organization."
+
+
+class AmbiguousUserException(APIException):
+    status_code = 409
+    default_detail = (
+        "Multiple user records match this email. The environment's IdP filter "
```
jaseemjaskp: **[P2] `default_detail` leaks internal terminology to end users** — `The environment's IdP filter may be misconfigured` will surface to admins performing role changes via the API response. They have no UI surface to act on "IdP filter," and it isn't a concept they can map to anything in their console. The diagnostic value lives in the `logger.error` that already fires in `_resolve_unique_member_by_email`.

Suggested:
```python
default_detail = (
    "Multiple user records match this email. "
    "Contact your administrator to resolve the duplicate."
)
```
Keep the IdP-filter detail in the log.

jaseemjaskp (commented):

Ran Comment Analyzer, PR Test Analyzer, Silent Failure Hunter, Type Design Analyzer, Code Reviewer, and Code Simplifier on commit `98c95bf8`. Existing greptile comments were excluded.

### Headline concerns (P1)
- **Caller-side regression risk on the admin path** — `OrganizationMemberService.get_user_by_id(admin.id)` at `authentication_controller.py:412/430` can now silently return `None` if the IdP filter is misconfigured for the admin's *own* row, and that `None` flows into `auth_service.add_organization_user_role(...)`. See inline at `authentication_controller.py:413`.
- **Silent multi-match on UserService lookups** — `get_user_by_user_id` / `get_user_by_email` previously raised `MultipleObjectsReturned`; now they `.first()` and silently pick a row. This is the same failure mode the PR is fixing at the controller layer — recommend mirroring `_resolve_unique_member_by_email` here. (Complementary to greptile's existing P1 on the inverse silently-hides case at `user.py:64`.)
- **Latent `IndexError`** in `add_user_role`/`remove_user_role`: `return current_roles[0]` is unconditional but `if current_roles:` only guards the save call (pre-existing but adjacent to changed lines).

### Tests (no inline anchor)
No tests added. `backend/account_v2/tests.py` and `backend/tenant_account_v2/tests.py` are empty placeholders. Highest-value tests to add before merge:
1. `_resolve_unique_member_by_email` — 0/1/2+ matches (assert `AmbiguousUserException` on 2+).
2. `UserFilterRegistry.apply` — no-op when empty, composition order, dedup via `register`. **Important:** `setUp`/`tearDown` must clear `_filters` to avoid cross-test contamination (this is the same root cause greptile flagged at `user_filter_registry.py:34`).
3. `UserService.get_user_by_email/user_id/id` — confirm `None`-on-no-match parity now that `.get()` was replaced with `.first()`.
4. End-to-end: `add_user_role` / `remove_user_role` raise 409 to the view on ambiguity.

### P2 findings
See inline — type drift on `get_members()` return annotation; docstring/exception message both leak the plugin-specific `AUTH0_OIDC_CONNECTION` env var into core (defeats the registry's plugin-agnostic intent); registry has no exception isolation for buggy plugins; `FilterKind` is a static-only invariant with no runtime cross-check against `qs.model`; repeated 4-line apply block across 6 methods.

backend/account_v2/user_filter_registry.py
```diff
@@ -0,0 +1,40 @@
...
+
+Plugins (e.g., the auth0 plugin) can register a callable that scopes
+querysets to a subset of users — for example, only users whose Auth0
+user_id belongs to the current environment's OIDC connection. Core
+services call ``UserFilterRegistry.apply`` on every user lookup so the
+registered filters take effect uniformly without core needing to know
+which plugin is loaded.
+
+When no filters are registered the registry is a no-op, so OSS and
+development setups are unaffected.
```
jaseemjaskp, line 13: **[P2] Module docstring contains rotting references** — naming `the auth0 plugin` couples core to a single consumer; if the plugin is renamed or a second IdP plugin adopts the registry, this will rot. The line `Core services call UserFilterRegistry.apply on every user lookup` is also aspirational — it's true only for the methods touched in this commit and is enforced by nothing. Generalise to e.g. `Identity plugins can register a callable...` and soften to `User and OrganizationMember service layers in core invoke UserFilterRegistry.apply...`.

backend/account_v2/user_filter_registry.py
```diff
@@ -0,0 +1,40 @@
...
+from collections.abc import Callable
+from typing import Literal
+
+from django.db.models import QuerySet
+
+# "user" filters operate on `account_v2.User` querysets and should
+# reference `user_id`. "org_member" filters operate on
+# `tenant_account_v2.OrganizationMember` querysets and should reference
+# `user__user_id`.
+FilterKind = Literal["user", "org_member"]
```
jaseemjaskp, line 28: **[P2] `FilterKind` is a static-only invariant** — `Literal["user", "org_member"]` helps mypy but not runtime; an arbitrary string flows through unchecked. Worse, the kind-to-field-path contract (`"user"` → `user_id`, `"org_member"` → `user__user_id`) lives only in the comment above and is enforced nowhere. A plugin filter that picks the wrong path against the wrong queryset returns empty results silently — exactly the diagnostic black hole the PR is otherwise trying to eliminate.

Suggested: add a runtime cross-check inside `apply()`:
```python
if kind == "user":
    assert qs.model is User, f"kind=user with qs.model={qs.model}"
elif kind == "org_member":
    assert qs.model is OrganizationMember
```
Alternatively, drop the `kind` argument entirely and dispatch on `qs.model` — that removes the dual-signature ambiguity at the source. A registration-time smoke call against `Model.objects.none()` would also let a wrong-path filter fail fast at app boot instead of at first login.

backend/account_v2/user_filter_registry.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @classmethod
+    def register(cls, fn: FilterFn) -> None:
+        if fn not in cls._filters:
+            cls._filters.append(fn)
+
+    @classmethod
+    def apply(cls, qs: QuerySet, kind: FilterKind) -> QuerySet:
+        for fn in cls._filters:
+            qs = fn(qs, kind)
```
jaseemjaskp: **[P2] `apply()` has no exception isolation for buggy plugins** — if a registered filter raises (e.g., reads `settings.AUTH0_OIDC_CONNECTION` and calls `.startswith` on `None`, or hits a `FieldError`), it propagates as a 500 to whichever random caller first invokes a user lookup, with no plugin identifier in the stack trace. Wrap the call:
```python
for fn in cls._filters:
    try:
        qs = fn(qs, kind)
    except Exception:
        logger.exception("user_filter plugin failed: fn=%r kind=%s", fn, kind)
        raise  # or `continue` — make the fail-open vs fail-closed choice explicit
```
With the current implicit fail-closed behaviour and no plugin attribution in logs, diagnosis will be painful.

backend/account_v2/user_filter_registry.py
```diff
@@ -0,0 +1,40 @@
...
+# reference `user_id`. "org_member" filters operate on
+# `tenant_account_v2.OrganizationMember` querysets and should reference
+# `user__user_id`.
+FilterKind = Literal["user", "org_member"]
+
+FilterFn = Callable[[QuerySet, FilterKind], QuerySet]
+
+
+class UserFilterRegistry:
+    _filters: list[FilterFn] = []
```
jaseemjaskp: **[P2 — testability]** Complementary to greptile's existing comment on line 34: the class-level mutable also means **there is no `unregister()` or `reset()` hook for plugins to deregister at app shutdown / reload**, and no `@contextmanager` for scoped registration in tests. Suggest:
```python
@classmethod
def unregister(cls, fn: FilterFn) -> None:
    if fn in cls._filters: cls._filters.remove(fn)

@classmethod
@contextmanager
def scoped(cls, fn: FilterFn):
    cls.register(fn)
    try: yield
    finally: cls.unregister(fn)
```
The `scoped()` helper is the cleanest pattern for the test gaps called out in the review body.

backend/tenant_account_v2/organization_member_service.py
```diff
@@ -8,28 +9,34 @@
...
-        except OrganizationMember.DoesNotExist:
-            return None
+        qs = UserFilterRegistry.apply(
+            OrganizationMember.objects.filter(user=id),  # type: ignore
+            "org_member",
+        )
+        return qs.first()
 
     @staticmethod
     def get_members() -> list[OrganizationMember]:
```
jaseemjaskp: **[P2] Return-type annotation drift** — declared `-> list[OrganizationMember]` but `UserFilterRegistry.apply(...)` returns a `QuerySet`. The annotation is now actively misleading (callers may assume listy semantics like `len()` is cheap). Either change to `QuerySet[OrganizationMember]` (cheap, accurate) or wrap in `list(...)`. Same annotation drift on `get_members_by_role` at line 42 — though it chains `.order_by(...)` so a QuerySet is actually required there; just fix the annotation.

backend/tenant_account_v2/organization_member_service.py
```diff
@@ -8,28 +9,34 @@
 class OrganizationMemberService:
     @staticmethod
     def get_user_by_email(email: str) -> OrganizationMember | None:
-        try:
-            return OrganizationMember.objects.get(user__email=email)  # type: ignore
-        except OrganizationMember.DoesNotExist:
-            return None
+        qs = UserFilterRegistry.apply(
```
jaseemjaskp: **[P2 — simplifier]** Six methods repeat the same 4-line `UserFilterRegistry.apply(OrganizationMember.objects.filter(...), "org_member")` block (and `# type: ignore` × 3). A small private helper would clean this up materially:
```python
@staticmethod
def _filtered_qs(**kwargs) -> QuerySet[OrganizationMember]:
    return UserFilterRegistry.apply(
        OrganizationMember.objects.filter(**kwargs), "org_member"
    )
```
Each lookup becomes a one-liner (`return cls._filtered_qs(user__email=email).first()`), the `"org_member"` kind string stops being repeated, and the `# type: ignore` comments collapse to one place. Not a correctness issue — just removes a tax that will recur every time a new lookup is added.

jaseemjaskp (commented):
Follow-up review after the ambiguity-log PII fix. Five findings below; one additional silent-failure-pattern outside the diff is noted in the chat summary (auth_controller.py:119 `user_organizations` swallows the new `AmbiguousUserException` into an empty 412).

backend/account_v2/user.py
```diff
@@ -41,28 +43,57 @@ def update_user(self, user: User, user_id: str) -> User:
...
+    misconfigured identity-scoping filter, and silently picking one would
+    propagate the wrong identity downstream.
+    """
+    qs = UserFilterRegistry.apply(qs, kind)
+    rows = list(qs[:2])
+    if len(rows) > 1:
+        # Log the matched row PKs (internal IDs, not PII) instead of the
+        # raw lookup value so ambiguity remains diagnosable from logs
+        # without expanding PII retention.
+        pks = list(qs.values_list("pk", flat=True))
```
jaseemjaskp: **[P0 — correctness / resource]** Unbounded re-query on the ambiguity log path. `rows = list(qs[:2])` correctly caps to 2, but `qs.values_list("pk", flat=True)` on line 90 re-executes the **unsliced** queryset — a misconfigured plugin filter that matches thousands of rows turns this error path into a full table scan plus PK materialization in memory, precisely when the system is already in a degraded state.

The `rows` list is already in memory but only holds 2 items (intentionally). Two safe options:
- Cap the diagnostic query: `pks = list(qs.values_list("pk", flat=True)[:50])` and label the log `(matched ≥%d; first %d pks=%s)`.
- Or do a bounded count first (`qs.count()`) before fetching PKs and bail out of PK collection past a threshold.

Mirror issue exists at `organization_member_service.py:41`.

backend/tenant_account_v2/organization_member_service.py
```diff
@@ -1,48 +1,87 @@
...
+            OrganizationMember.objects.filter(user__email=email),  # type: ignore
+            "org_member",
+        )
+        rows = list(qs[:2])
+        if len(rows) > 1:
+            # Log the matched member_ids (internal IDs, not PII) instead
+            # of the raw email so ambiguity remains diagnosable from logs
+            # without expanding PII retention.
+            member_ids = list(qs.values_list("member_id", flat=True))
```
jaseemjaskp: **[P0 — correctness / resource, mirrors `user.py:90`]** `qs.values_list("member_id", flat=True)` re-executes the **unsliced** queryset (the `qs[:2]` on line 36 only limited the previous result; `qs` itself is the original full filter). Under a misconfigured filter matching N rows this fetches all N `member_id`s on the error path.

Fix: `member_ids = list(qs.values_list("member_id", flat=True)[:50])` and label the log `(showing first 50)`. If this and the matching `user.py` path are lifted into a shared helper, fix it once.

backend/tenant_account_v2/organization_member_service.py
```diff
@@ -1,48 +1,87 @@
...
     def get_user_by_email(email: str) -> OrganizationMember | None:
-        try:
-            return OrganizationMember.objects.get(user__email=email)  # type: ignore
-        except OrganizationMember.DoesNotExist:
-            return None
+        qs = UserFilterRegistry.apply(
+            OrganizationMember.objects.filter(user__email=email),  # type: ignore
+            "org_member",
+        )
+        return qs.first()
```
jaseemjaskp: **[P1 — broken contract]** `get_user_by_email` still calls `.first()` after `UserFilterRegistry.apply`. The PR's stated thesis is that ambiguity must **raise**, not silently pick one — yet this legacy method is still wired into the invite flow at `authentication_controller.py:352` (`invite_user`).

Failure mode: with the OIDC scoping plugin loaded but misconfigured (e.g., `AUTH0_OIDC_CONNECTION` missing for one env), `get_user_by_email` returns the **wrong tenant's** duplicate-email row, and `invite_user` short-circuits to `"User is already part of current organization"` for a user who genuinely should be invited. No log, no exception.

Fix one of:
- Route `invite_user` (and any caller that branches on identity, including `authentication_helper.py:50` `get_or_create_user_by_email` via `UserService.get_user_by_email`) through the unique resolver.
- Inline the same `[:2]` ambiguity check into this method and remove `get_unique_user_by_email`.
- At minimum, `logger.error` when `qs[:2]` returns >1 row even on the non-unique path.

backend/account_v2/authentication_controller.py
```diff
@@ -100,8 +100,8 @@ def authorization_callback(
             return self.auth_service.handle_authorization_callback(
                 request=request, backend=backend
             )
-        except Exception as ex:
-            logger.error(f"Error while handling authorization callback: {ex}")
+        except Exception:
+            logger.exception("Error while handling authorization callback")
```
jaseemjaskp, line 110: **[P1 — broken contract]** The catch on line 103 swallows the new `AmbiguousUserException` (and every other `APIException`) into a generic `/error` redirect. The whole point of raising a 409 with `"Multiple user records match this lookup. Contact your administrator…"` is that the admin sees something actionable — but during OIDC sign-in this lookup happens inside `handle_authorization_callback`, so the actionable detail is lost and the user ends up on `/error` with no clue duplicate identity rows exist.

Fix: narrow the catch, e.g.
```python
except APIException:
    raise
except Exception:
    logger.exception("Error while handling authorization callback")
    return redirect("/error")
```
or include an error code in the redirect query-string so the frontend can surface 'duplicate identity' specifically.

_Note:_ a similar swallow pattern exists in `user_organizations` at line 119 (out of this PR's diff), which silently logs the user out and returns an empty 412 for any non-allowlisted exception — please consider the same fix there in a follow-up.

backend/account_v2/user_filter_registry.py
```diff
@@ -0,0 +1,63 @@
...
+# reference `user_id`. "org_member" filters operate on
+# `tenant_account_v2.OrganizationMember` querysets and should reference
+# `user__user_id`.
+FilterKind = Literal["user", "org_member"]
+
+FilterFn = Callable[[QuerySet, FilterKind], QuerySet]
+
+
+class UserFilterRegistry:
+    _filters: ClassVar[list[FilterFn]] = []
```
jaseemjaskp, line 34: **[P1 — testability / regression risk]** Complementary to greptile's P2 on this line about `_filters` not being resettable — additionally, **no test currently exercises any of the new code paths in this PR**. `grep -rn` for `UserFilterRegistry`, `AmbiguousUserException`, `_resolve_unique`, `get_unique_user_by_email` returns matches only in production files; `find backend -path "*account*" -name "test_*.py"` returns nothing.

Given this PR is identity-scoping and security-sensitive, the absence of tests means a future innocent-looking refactor (e.g., reverting `_resolve_unique` to `.get()` for terseness, or re-adding the registry to `get_user_by_id`) would silently undo the fix. Critical regressions to cover:

1. PK-bypass: with a filter that excludes everything registered, `UserService.get_user_by_id(pk)` and `OrganizationMemberService.get_user_by_id(pk)` still return the row (admin-self-lookup).
2. Ambiguity raises (not silently picks): 2+ rows → `AmbiguousUserException`; 1 row → that row; 0 rows → `None`.
3. `apply` **re-raises** plugin exceptions (the prior implementation swallowed; a regression here would silently leak unscoped users across IdPs).
4. Ambiguity log emits PKs/member_ids only and **does not** contain the raw email string.

Each test should `addCleanup(UserFilterRegistry.clear)` to avoid cross-test bleed via the class-level list.

jaseemjaskp:
Differed

PR #1986, opened 2026-05-22
UN-2977 [FEAT] Introduce users groups with project sharing 

kirtimanmishrazipstack:
## What

Adds **org-scoped groups** as a first-class sharing target alongside the existing per-user share and org-wide share, across all 7 shareable resources (Workflow, Pipeline, API Deployment, Connector, Adapter, Prompt Studio project, and Agentic project). An org admin creates a group, manages its members, and shares resources with the group in one action; access tracks membership automatically.

## Why

Admins needed to grant a set of users access to a resource in a single action and have that access follow membership changes, instead of re-sharing to each user individually. Groups are **manual-only** — login method (SSO or email/password) is authentication only and never drives membership.

## How

- **Data model (`tenant_account_v2`):** `OrganizationGroup`, `GroupMembership` (explicit through-model), and `ResourceGroupShare` — a single polymorphic `group → resource` table that covers every resource type, so no per-resource `shared_groups` column/migration is needed. `shared_groups` is exposed as a read property on each model backed by this table.
- **Visibility:** each resource manager's `for_user(user)` unions `created_by | shared_users | shared_to_org | group-shared`; group ids are read live from `GroupMembership`, so membership changes take effect immediately. `resources_visible_via_groups` casts the varchar `object_id` to the model's UUID PK to avoid a Postgres type mismatch.
- **Permissions:** org admins retain full org-wide access to every resource (inherited from the already-merged UN-3479 / #1989, not introduced here); group sharing is **additive** on top of that and only grants access to non-admin members. All share mutations are funnelled through a single `POST /<resource>/{id}/share/` action (the `shared_users`/`shared_groups`/`shared_to_org` fields are `read_only` on the resource serializers); `ShareAuthorizationService` diffs requested vs. current state per axis and authorizes each change (owners and admins on every axis). The group `resources/` endpoint is admin-only.
- **Reuse:** one `ResourceShareManagementMixin` provides `share/` + `effective-members/` on every shareable viewset; an admin-gated `OrganizationGroupViewSet` provides group CRUD + member management.
- **Lifecycle:** a `post_delete` signal on `OrganizationMember` purges the removed user's group memberships and direct `shared_users` rows (closes the rejoin backdoor); `created_by` is untouched.
- **Frontend:** a Groups admin page (CRUD + member management + delete-confirm showing affected-resource count) and a batched share modal (edits are local until **Apply** sends one `POST /share/`).

- **Share mutations via PATCH no longer work.** The sharing fields are now `read_only` on resource serializers; any caller that previously PATCHed `shared_users`/`shared_to_org` directly must use the new `POST /<resource>/{id}/share/` action. Internal callers and the UI are migrated; external API clients doing direct PATCH-share would silently no-op.
- **Org-member removal now also strips direct shares.** The new `post_delete` signal removes the user from every resource's `shared_users` (in addition to group memberships). Intentional, but a behavior change vs. before.
- **Org admins keep org-wide access** (UN-3479 / #1989). `for_user()` and the permission classes admit admins; group sharing never restricts them — it only adds access for the group's non-admin members.

## Database Migrations

- `tenant_account_v2/0002_organization_group_group_membership.py` — `OrganizationGroup`, `GroupMembership`.
- `tenant_account_v2/0003_resource_group_share.py` — `ResourceGroupShare`.

## Env Config

- `MAX_GROUPS_PER_ORG` (default `200`) — max groups per org.
- `MAX_MEMBERS_PER_GROUP` (default `500`) — max members per group.
- Both optional; exceeding returns HTTP 400 with a structured `code`.

## Related Issues or PRs

- Cloud counterpart: Zipstack/unstract-cloud#1508 (AgenticProject parallel-copy + auth0) — must merge **at or before** this PR or agentic group sharing breaks in cloud.
- Phase 2 (co-owner integration, UN-2022) is a separate follow-up.

## Dependencies Versions

- None.

## Notes on Testing

Manual verification per spec §7: create a group, add/remove members, share each resource type with the group, confirm group-only members gain view/run access and lose it on removal, and confirm the quota limits return HTTP 400.

## Checklist


backend/tenant_account_v2/group_views.py
```diff
@@ -0,0 +1,214 @@
...
+                organization=organization, user=request.user
+            )
+        elif member_filter and member_filter != "me":
+            if not is_admin:
+                raise PermissionDenied(
+                    "Only admins can query other users' group memberships."
+                )
+            qs = list_groups_with_member_counts(organization=organization).filter(
+                memberships__user_id=member_filter
+            )
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `863c736ef` — the `?member=` value is now parsed with `int()` and returns a 400 (`ValidationError`) instead of letting the ORM raise a 500.

backend/adapter_processor_v2/views.py
```diff
@@ -292,91 +297,103 @@ def destroy(
...
-                    continue
-
-        # Perform the update
+            # Adapter-specific: when a user loses access, clear their
+            # ``UserDefaultAdapter`` rows that pointed at this adapter so they
+            # don't keep a stale default. Must run BEFORE the M2M update.
+            self._clear_default_adapter_for_removed_users(
+                adapter, before["shared_users"], request.data
+            )
+
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `863c736ef` — the `UserDefaultAdapter` cleanup moved out of `partial_update` (where `shared_users` is read-only) into the `share` action, keyed off a before/after diff of the actual M2M, so defaults are cleared only when a user truly loses access.

backend/tenant_account_v2/group_views.py
```diff
@@ -0,0 +1,214 @@
...
+        group = self._get_group_or_404(pk)
+        if request.method == "GET":
+            qs = group.memberships.select_related("user").order_by("created_at")
+            data = GroupMemberSerializer(qs, many=True).data
+            return Response(data)
+
+        # POST → bulk add
+        if not _is_org_admin(request):
+            raise PermissionDenied(self.permission_classes[1].message)
+        serializer = GroupMemberAddSerializer(data=request.data, context={"group": group})
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `863c736ef` — now references `IsOrgAdminForWrite.message` directly instead of `self.permission_classes[1]`.

backend/tenant_account_v2/group_views.py
```diff
@@ -0,0 +1,214 @@
...
+    @action(
+        detail=True,
+        methods=["delete"],
+        url_path=r"members/(?P<user_id>[^/.]+)",
+    )
+    def remove_member(
+        self, request: Request, pk: str | None = None, user_id: str | None = None
+    ) -> Response:
+        if not _is_org_admin(request):
+            raise PermissionDenied(self.permission_classes[1].message)
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `863c736ef` — `remove_member` references `IsOrgAdminForWrite.message` directly now too.

backend/tenant_account_v2/sharing_helpers.py
```diff
@@ -0,0 +1,509 @@
...
+    ``Q(pk__in=...)``, so we cast in Python instead. The result set is
+    bounded by ``(groups user belongs to) × (shared rows of that model)``.
+    """
+    raw_ids = ResourceGroupShare.objects.filter(
+        content_type=ContentType.objects.get_for_model(model),
+        group_id__in=user_group_ids,
+    ).values_list("object_id", flat=True)
+    if isinstance(model._meta.pk, models.UUIDField):
+        return [uuid.UUID(s) for s in raw_ids if s]
+    return list(raw_ids)
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Acknowledged — leaving as-is for now. The Python materialisation is a deliberate, documented workaround for Postgres refusing the implicit `uuid = character varying` cast on the varchar `object_id`. The result set is bounded by (user groups) x (shared rows of that model). A RawSQL `::uuid` subquery is the real optimisation and is tracked as a follow-up rather than done in this PR.

backend/tenant_account_v2/sharing_helpers.py
```diff
@@ -0,0 +1,509 @@
...
+        # Lazy import to keep ``tenant_account_v2`` boot light and to avoid
+        # circular import with ``account_v2.authentication_controller``.
+        from account_v2.authentication_controller import AuthenticationController
+
+        controller = AuthenticationController()
+        member = controller.get_organization_members_by_user(user=user)
+        return controller.is_admin_by_role(member.role)
+    except Exception:
+        logger.exception("Error resolving admin role for user_id=%s", user.pk)
+        return False
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `863c736ef` — `is_org_admin` now catches `ObjectDoesNotExist`/`LookupError` separately and returns False without logging, so an org-less user no longer produces error-log noise. The broad `except` remains only for genuinely unexpected failures.

backend/tenant_account_v2/sharing_helpers.py
```diff
@@ -0,0 +1,509 @@
...
+    group: OrganizationGroup, model: type[Model]
+) -> QuerySet:
+    """Resources of ``model`` shared with ``group`` (replaces
+    ``model.objects.filter(shared_groups=group)``).
+    """
+    shared_object_ids = ResourceGroupShare.objects.filter(
+        group=group,
+        content_type=ContentType.objects.get_for_model(model),
+    ).values("object_id")
+    return model.objects.filter(pk__in=shared_object_ids)
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `863c736ef` — `list_resources_shared_with_group` now materialises the `object_id` values and casts them to UUID before the `pk__in` lookup, mirroring `resources_visible_via_groups`. Resolves the `operator does not exist: uuid = character varying` crash on `GET /groups/{pk}/resources/`.

backend/adapter_processor_v2/serializers.py
```diff
@@ -24,10 +26,19 @@ class TestAdapterSerializer(serializers.Serializer):
...
+    # mutations go through ``POST /adapter/{id}/share/`` (UN-2977 plan §B).
+    shared_groups = serializers.PrimaryKeyRelatedField(many=True, read_only=True)
+
     class Meta:
         model = AdapterInstance
         fields = "__all__"
+        extra_kwargs = {
+            "shared_users": {"read_only": True},
+            "shared_to_org": {"read_only": True},
+        }
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `863c736ef` — the default-adapter cleanup that read `request.data[shared_users]` in `partial_update` has been removed; since `shared_users` is read-only there, unsharing only happens via the `share` action, where the cleanup now runs against a before/after M2M diff.

backend/adapter_processor_v2/views.py
```diff
@@ -292,91 +297,103 @@ def destroy(
...
+        users about to be unshared.
+
+        Computed against ``request.data`` because this runs *before* the M2M
+        update lands; the post-update diff would be too late.
+        """
+        requested_user_ids = {
+            int(user_id) for user_id in request_data.get("shared_users", [])
+        }
+        current_user_ids = {user.id for user in current_shared_users}
+        removed_user_ids = current_user_ids - requested_user_ids
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `863c736ef` — the `int(user_id)` over raw `request.data` is gone. The cleanup helper now receives an already-resolved set of removed user IDs from the `share` action diff, so there is no pre-validation `int()` that can raise.

backend/tenant_account_v2/share_serializer_mixin.py
```diff
@@ -0,0 +1,44 @@
...
+        if groups is not None:
+            set_resource_share_groups(instance, [g.id for g in groups])
+        return instance
+
+    def update(self, instance: Any, validated_data: dict[str, Any]) -> Any:
+        groups = validated_data.pop("shared_groups", None)
+        instance = super().update(instance, validated_data)  # type: ignore[misc]
+        if groups is not None:
+            set_resource_share_groups(instance, [g.id for g in groups])
+        return instance
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `863c736ef` — `create` and `update` now wrap `super().create/update` and `set_resource_share_groups` in a single `transaction.atomic()` block.

backend/workflow_manager/workflow_v2/urls/workflow.py
```diff
@@ -26,6 +26,8 @@
 workflow_schema = WorkflowViewSet.as_view({"get": "get_schema"})
 can_update = WorkflowViewSet.as_view({"get": "can_update"})
 list_shared_users = WorkflowViewSet.as_view({"get": "list_of_shared_users"})
+workflow_share = WorkflowViewSet.as_view({"post": "share"})
+workflow_effective_members = WorkflowViewSet.as_view({"get": "effective_members"})
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: By design, not changing. The HTTP gate on `share` is intentionally `IsOwnerOrSharedUserOrSharedToOrg`; the per-axis authorization (only owner/admin may remove; shared users and group members may add within scope; only owner/admin may toggle `shared_to_org`) is enforced in `ShareAuthorizationService` (see `sharing_helpers.py`). Restricting the route to owner-only would break the UN-2977 sharing matrix where shared users can extend a share.

frontend/src/components/groups/GroupCreateEditModal.jsx
```diff
@@ -0,0 +1,90 @@
...
+    // Ant Design surfaces validation errors inline; bail out quietly on failure.
+    const values = await form.validateFields().catch(() => null);
+    if (!values) {
+      return;
+    }
+    setSubmitting(true);
+    const call =
+      mode === "edit"
+        ? service.updateGroup(group.id, values)
+        : service.createGroup(values);
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `863c736ef` — added a guard that returns early with an error alert when `mode === edit` and `group?.id` is missing, before the `updateGroup` call.

frontend/src/components/groups/groups-service.js
```diff
@@ -0,0 +1,89 @@
...
+  adapter_instance: "adapter",
+  connector_instance: "connector",
+  custom_tool: "prompt-studio",
+};
+
+function groupsService() {
+  const axiosPrivate = useAxiosPrivate();
+  const { sessionDetails } = useSessionStore();
+  const path = `/api/v1/unstract/${sessionDetails?.orgId.replaceAll('"', "")}`;
+  const csrfToken = sessionDetails.csrfToken;
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `863c736ef` — `orgId` is now optional-chained with a default: `sessionDetails?.orgId?.replaceAll?.(...) ?? ""`.

frontend/src/components/groups/Groups.jsx
```diff
@@ -0,0 +1,266 @@
...
+  ];
+
+  return (
+    <>
+      <TopBar
+        enableSearch={true}
+        title="Manage Groups"
+        searchData={groupList}
+        setFilteredUserList={setFilteredGroupList}
+      >
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `863c736ef` — `TopBar` now takes a `searchKey` prop (default `email`) and guards the field access; `Groups.jsx` passes `searchKey=name`, so search filters group rows by name instead of crashing on a missing `email`.

frontend/src/components/navigations/side-nav-bar/SideNavBar.jsx
```diff
@@ -123,6 +123,15 @@ const getSettingsMenuItems = (orgName, isAdmin) => [
...
   },
+  ...(isAdmin
+    ? [
+        {
+          key: "groups",
+          label: "Groups",
+          path: `/${orgName}/groups`,
+        },
+      ]
+    : []),
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `863c736ef` — `/${orgName}/groups` added to the Platform item active-state condition.

frontend/src/components/tool-settings/tool-settings/ToolSettings.jsx
```diff
@@ -183,24 +193,25 @@ function ToolSettings({ type }) {
...
     axiosPrivate(requestOptions)
-      .then((response) => {
-        setOpenSharePermissionModal(false);
-      })
       .catch((err) => {
         setAlertDetails(handleException(err, "Failed to update sharing"));
+      })
+      .finally(() => {
+        setOpenSharePermissionModal(false);
       });
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `863c736ef` — the modal close moved from `finally` to `then`, so it stays open on failure for retry.

backend/tenant_account_v2/group_views.py
```diff
@@ -0,0 +1,218 @@
...
+        methods=["delete"],
+        url_path=r"members/(?P<user_id>[^/.]+)",
+    )
+    def remove_member(
+        self, request: Request, pk: str | None = None, user_id: str | None = None
+    ) -> Response:
+        if not _is_org_admin(request):
+            raise PermissionDenied(IsOrgAdminForWrite.message)
+        group = self._get_group_or_404(pk)
+        deleted, _ = group.memberships.filter(user_id=user_id).delete()
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `822422307` — `remove_member` now guards `user_id` with `int()` and raises a 400 `ValidationError` on a non-numeric path segment, mirroring the `list` action. The ORM filter uses the parsed `user_id_int`.

frontend/src/components/widgets/top-bar/TopBar.jsx
```diff
@@ -10,6 +10,7 @@ function TopBar({
   enableSearch,
   searchData,
   setFilteredUserList,
+  searchKey = "email",
   children,
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `822422307` — `TopBar` now takes a `searchPlaceholder` prop (default `Search Users`) used on the input, and `Groups.jsx` passes `searchPlaceholder=Search Groups`. propTypes updated.

backend/permissions/resource_share_views.py
```diff
@@ -0,0 +1,162 @@
...
+        ``ShareAuthorizationService``.
+        """
+        from tenant_account_v2.sharing_helpers import ShareAuthorizationService
+
+        resource = self.get_object()  # type: ignore[attr-defined]
+        desired = _extract_desired_share_state(request.data)
+        ShareAuthorizationService.authorize_and_commit(
+            actor=request.user, resource=resource, desired=desired
+        )
+        return Response(status=status.HTTP_200_OK)
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Good catch, and confirmed — with `shared_users` read-only on the serializers, the `partial_update` notify path no longer fires and the `share` action does not yet notify. We are intentionally deferring the notification rework to the next phase: the share-notification trigger needs to move onto the `share` action AND be reworked into batched notifications that also cover the new `shared_groups` axis (group-share fan-out), which is a design item for that phase rather than a like-for-like restore here. Tracking it there; leaving the current path untouched in this PR so the change lands as one coherent piece.

backend/tenant_account_v2/group_views.py
```diff
@@ -0,0 +1,222 @@
...
+            raise ValidationError({"user_id": "Must be a numeric user ID."}) from exc
+        group = self._get_group_or_404(pk)
+        deleted, _ = group.memberships.filter(user_id=user_id_int).delete()
+        if not deleted:
+            raise NotFound("User is not a member of this group.")
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    # --- resources shared with this group ------------------------------------
+
+    @action(detail=True, methods=["get"], url_path="resources")
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Valid catch — fixed in 44b73c2dc.

The `resources` action is the delete blast-radius view, consumed only by the admin-only delete-confirm flow (`Groups.jsx`). As you noted, `IsOrgAdminForWrite` permits GET for any authenticated org member and the action uses `get_object_or_404` (so no `has_object_permission` hook), leaving it open to enumeration of resources shared with groups the caller isn't in — which contradicts the model where org admin itself grants no implicit resource access.

Gated the action to org admins, mirroring the existing `members` POST / `remove_member` guards:

```python
if not _is_org_admin(request):
    raise PermissionDenied(IsOrgAdminForWrite.message)
```

kirtimanmishrazipstack (commented):
## Comprehensive review — UN-2977 group-based sharing

Reviewed alongside the cloud counterpart (#1508) since they share the sharing abstractions. Overall this is a careful, well-structured change — multi-tenant isolation on the suspect paths (the varchar→UUID cast, `has_group_access`, the org-scoped managers) holds up, and the share-authorization matrix is coherent. Findings below are real and currently present in the code.

### 🔴 Needs a decision before merge

**1. Org-admin implicit access: the code does the opposite of what this PR's description says.**

The description states (three times): *"being an org admin grants **no** implicit access… **Org admin implicit access is explicitly denied** — Enforced in `for_user()` and the permission classes — an admin who is not owner/shared cannot view or run a resource."*

The actual code grants admins full access on **both** axes:
- `for_user()` in all 6 managers short-circuits: `if OrganizationMemberService.is_user_organization_admin(user): return self.all()` — `pipeline_v2/models.py:35`, `api_v2/models.py:42`, `workflow_manager/workflow_v2/models/workflow.py:33`, `connector_v2/models.py:34`, `adapter_processor_v2/models.py:41`, `prompt_studio/prompt_studio_core_v2/models.py:30`.
- Every permission class ORs `_is_organization_admin(request)` — `permissions/permission.py:79` and `:113` — and `IsOwner`'s docstring affirms *"Org admins can manage every resource in their organization."*

One of the two is wrong. Either the code is correct and the description (incl. the "Can this PR break" section) is materially stale, **or** the description reflects the intended invariant and there's a least-privilege/multi-tenancy gap across all 6 resources + the permission layer. Please reconcile against the design source of truth (mfbt UNS-612) and align the description and/or the code. This also drives the cloud parity fix in #1508.

### 🟠 Important

**2. `member_count` is always `1` when the group list is filtered by member.**
`tenant_account_v2/group_serializers.py:175-180` applies `.filter(memberships__user=user)` *before* `.annotate(Count("memberships"))`. Filtering and aggregating on the same multi-valued relation makes `Count` run over the constrained join → `member_count = 1` for every group in both `?member=me` (non-admin "my groups") and `?member=<id>` (admin filter). The unfiltered list path is correct. Fix: compute the count via a decoupled `Subquery` so the member filter can't taint it.

**3. `SharedGroupsSerializerMixin` is dead code, and its docstring advertises a writable API.**
`shared_groups` is `read_only=True` on all 7 resource serializers, so DRF strips it from `validated_data` and the mixin's `create`/`update` bodies (`tenant_account_v2/share_serializer_mixin.py:34-50`) always no-op — every group write actually goes through `ShareAuthorizationService._commit`. The module docstring (`:12-19`) even shows a writable `PrimaryKeyRelatedField(queryset=…)` example that contradicts every call site. Delete the mixin (and drop it from the serializers), or make it actually fire. The dead path also skips the org-scope check, so it's a latent footgun if reactivated.

**4. No orphan cleanup of `ResourceGroupShare` when a resource is deleted.**
`object_id` is a plain varchar (no FK/GFK), and `tenant_account_v2/signals.py` only purges on `OrganizationMember` removal. Deleting a Workflow/Pipeline/Connector/etc. leaves dangling share rows indefinitely. Add a `post_delete` receiver for the shareable models mirroring the existing lazy-model loop in `signals.py` (it already no-ops cleanly for apps that aren't installed, so it can include the cloud `AgenticProject`).

**5. The `post_delete` purge signal isn't atomic and has no per-step error handling.**
`signals.py:cleanup_user_org_access` bulk-deletes memberships then loops `resource.shared_users.remove(...)` across the shareable models with no `transaction.atomic()` and no try/except. A mid-loop failure leaves the user purged from some resources but not others — re-opening the "rejoin backdoor" the docstring claims to close — with no log of the partial purge. Wrap the cleanup in `atomic()` and log-and-reraise per model.

**6. Share modal closes on failure inconsistently across resources.**
`frontend/src/hooks/useShareModal.js:119-124`, `ListOfTools.jsx`, and `Workflows.jsx` close the modal in `.finally()` (success *or* failure), discarding the user's selection on a 400/403 with only a transient toast; `ToolSettings.jsx` correctly closes only on success. For partial-authorization rejections the user needs to see what was rejected. Close on `.then` only, keep open on `.catch`, and make all resources consistent.

### 🟡 Suggestions

- **Defensive UUID cast (latent):** `tenant_account_v2/sharing_helpers.py:137` and `:160` do `uuid.UUID(s)` with only an empty-string guard. The write path stores valid PKs today, but since the column is deliberately varchar, one malformed `object_id` would 500 the **entire** resource list for every member of that group. Skip-and-log malformed IDs instead.
- **Stringly-typed share axes (project anti-pattern):** the same axis vocabulary is declared three times as bare strings (`sharing_helpers.py:306-308`, `permissions/resource_share_views.py:19`, mixin `share_axes`); `group_views.py:204-209` and the `access_via` literals in `compute_effective_members` (`sharing_helpers.py:192,211,242`) duplicate constants that exist elsewhere. A `ShareAxis` enum + `DesiredShareState`/`AccessVia` would collapse these and double as the `content_type` allowlist the model currently lacks. (`AxisDiff` already exists, but `ShareAuthorizationService` re-derives tuples instead of consuming it.)
- **Unbounded queryset:** `compute_effective_members._add_org_members` iterates every org member with no pagination on a synchronous GET (`effective-members/`). Cap/paginate or `.iterator()`.
- **Adapter default-cleanup runs outside the share transaction** (`adapter_processor_v2/views.py:264-282`) — best-effort cleanup with no try/except after an already-committed share. Fold into the atomic block, or log on failure.
- **Comment accuracy:** the `resource_share_views.py` "axis-agnostic… any number of M2M axes" claim is inaccurate (`shared_groups` is *not* M2M); the duplicated `- Service accounts see all org resources` bullet added above the existing combined line in all 6 `for_user` docstrings is redundant; the `cleanup_user_org_access` "closes the rejoin backdoor" wording overstates the (currently non-atomic) guarantee.

### ✅ Well-handled
Polymorphic single-table design avoids per-resource migration sprawl; migration dependencies + reversibility are correct; quota checks return structured 400s; lazy `%` logging is consistent; org-scope is re-validated defense-in-depth inside `ShareAuthorizationService`; the merge-order coupling with #1508 is correctly documented.

---
<sub>🤖 Claude Code–assisted multi-agent review (code quality, silent-failure, type-design, comment accuracy); high-impact findings verified against the code.</sub>

kirtimanmishrazipstack:
Thanks for the thorough review — worked through every item. Each was validated against the code and `git blame` before acting.

**Fixed & pushed** (`2d8d098` backend, `1da7105` frontend):
- **#2** `member_count` — now computed via a decoupled `Subquery`, so the `?member=me` / `?member=<id>` filter no longer collapses it to 1.
- **#4** orphan `ResourceGroupShare` — added a `post_delete` receiver (wired per `_SHAREABLE_MODELS`, idempotent via `dispatch_uid`) that purges share rows when a resource is deleted.
- **#5** purge atomicity — `cleanup_user_org_access` now runs in `transaction.atomic()` with per-model `try/except` that logs and re-raises, so a partial purge rolls back instead of re-opening the rejoin backdoor.
- **#6** share modals — reverted to **close-on-success** (`useShareModal`, `ListOfTools`, `Workflows`), restoring pre-PR behavior and matching `ToolSettings`; a rejected share now keeps the modal open with the selection intact.
- **Suggestion A** — malformed `object_id` UUIDs are skip-and-logged instead of 500ing the whole list.
- **Suggestion C** — the post-commit adapter default-cleanup `.save()` is now wrapped in `try/except` + `logger.exception`.
- **Suggestion D** — fixed the "M2M axes" wording (`shared_groups` is polymorphic), removed the duplicate "Service accounts see all org resources" docstring bullet, and clarified the mixin docstring.

**Addressed by reply (no code change):**
- **#1 (admin implicit access):** the code is correct, not contradictory. The org-wide admin override comes from the already-merged **UN-3479 (#1989)** — it's *context*, not an addition, in this PR's diff, and `main`'s `for_user()` docstrings already say "org admins see all org resources". The PR description + spec were the stale part; both are now updated to state admins retain org-wide access and group sharing is additive. Reverting admin access would undo a separately-shipped feature, so it's out of scope here.
- **#3 (`SharedGroupsSerializerMixin` "dead code"):** kept intentionally. It's inert for the 6 OSS resource serializers (`read_only` → writes route through the `POST /share/` action), but the cloud `AgenticProject` serializer consumes it **writable**, so deleting it would break cloud group sharing. Tightened the module docstring to document both modes.
- **Suggestion B (`effective-members/` pagination):** deferred. The member set is bounded in practice (`MAX_MEMBERS_PER_GROUP=500`, org membership in the hundreds); will revisit pagination if orgs grow large enough to warrant it.

kirtimanmishrazipstack:
Correction to my earlier reply on **#3**: I said the cloud `AgenticProject` serializer consumes `SharedGroupsSerializerMixin` *writable*, so deleting it would break cloud. That was wrong — it came from a stale local copy. The cloud PR (#1508) actually ships `shared_groups = read_only=True` too, so the mixin was dead across **all 7** serializers (no writable consumer anywhere; every group write goes through the `POST /share/` action).

So I deleted it rather than keeping it — commit `befd61c9b` removes the mixin file and unwires it from the 6 OSS serializers (read-only `shared_groups` declarations stay); the cloud side is unwired in unstract-cloud#1508. This also removes the latent org-scope footgun noted in the original review.

backend/permissions/permission.py
```diff
@@ -71,13 +91,12 @@ class IsOwnerOrSharedUser(permissions.BasePermission):
...
-        return False
+        return (
+            obj.created_by == request.user
+            or obj.shared_users.filter(pk=request.user.pk).exists()
+            or has_group_access(request.user, obj)
+            or _is_organization_admin(request)
+        )
 
 
 class IsOwnerOrSharedUserOrSharedToOrg(permissions.BasePermission):
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Resolved by aligning the docs to the code. This quote is from the **old** PR description — both claims ("no implicit access" / "explicitly denied") have since been removed. The `_is_organization_admin` branches in `IsOwnerOrSharedUser` / `IsOwnerOrSharedUserOrSharedToOrg` and the `for_user()` short-circuits are the **intended** org-wide admin override from **UN-3479 (#1989)**, which merged to `main` independently of this feature (it's context in this diff, not introduced here). The description and spec now state admins retain org-wide access and group sharing is additive — so there's no unimplemented boundary; the stale claim was the bug, and it's fixed. cc @greptile-apps

backend/adapter_processor_v2/urls.py
```diff
@@ -25,6 +25,8 @@
 
 adapter_users = AdapterInstanceViewSet.as_view({"get": "list_of_shared_users"})
 adapter_info = AdapterInstanceViewSet.as_view({"get": "adapter_info"})
+adapter_share = AdapterInstanceViewSet.as_view({"post": "share"})
+adapter_effective_members = AdapterInstanceViewSet.as_view({"get": "effective_members"})
```
chandrasekharan-zipstack, line 29: @kirtimanmishrazipstack how is this different from list_of_shared_users?

backend/adapter_processor_v2/views.py
```diff
@@ -135,21 +136,25 @@ def test(self, request: Request) -> Response:
...
-class AdapterInstanceViewSet(ModelViewSet):
+class AdapterInstanceViewSet(ResourceShareManagementMixin, ModelViewSet):
     serializer_class = AdapterInstanceSerializer
 
     def get_permissions(self) -> list[Any]:
-        if self.action in ["update", "retrieve"]:
+        # Frictionless adapter quirk preserved: blocks update + retrieve so
+        # frictionless adapters stay hidden from non-owners, and lets any
+        # org member delete them. Non-frictionless adapters fall through to
+        # the standard owner / shared-user gating.
```
chandrasekharan-zipstack: NIT: Try to make the code comments concise and generic. This feels specific to your conversation with the agent

backend/permissions/resource_share_views.py
```diff
@@ -0,0 +1,163 @@
...
+
+    @action(detail=True, methods=["get"], url_path="effective-members")
+    def effective_members(self, request: Request, pk: str | None = None) -> Response:
+        """Return all users with access (direct/group/org), priority-deduped."""
+        # Lazy import — ``tenant_account_v2`` is the canonical home of the
+        # helper; importing at module load would pull a circular dep through
+        # the permissions package.
+        from tenant_account_v2.group_serializers import EffectiveMemberSerializer
```
chandrasekharan-zipstack, line 109: NIT: Check if imports can be moved to the top of the file wherever possible

backend/pipeline_v2/models.py
```diff
@@ -121,6 +129,17 @@ class PipelineStatus(models.TextChoices):
         default=False,
         db_comment="Whether this pipeline is shared with the entire organization",
     )
+    # ``shared_groups`` is no longer an M2M; group shares are stored
```
chandrasekharan-zipstack: NIT: This comment feels too repetitive and specific.  Also there were no 'shared_groups' previously right so the first line doesn't make much sense

chandrasekharan-zipstack:
### Hardcoded shareable-model lists drift — cloud `AgenticProject` is already missing from one

This PR keeps **two** independently hardcoded "what is shareable" lists that have to stay in lockstep, and they've already diverged:

1. `tenant_account_v2/signals.py` → `_SHAREABLE_MODELS` — includes `("agentic_studio_v1", "AgenticProject")`, so the dangling-`ResourceGroupShare` cleanup **and** the `shared_users` purge cover agentic projects. ✅
2. `tenant_account_v2/group_views.py` → `_collect_resources_shared_with_group()`'s `sources` tuple — only the 6 OSS models; **`AgenticProject` is absent**. ❌

**Consequence:** an agentic project shared with a group won't appear in that group's `resources/` blast-radius view (the affected-resource count surfaced in the delete-confirm dialog). Both lists encode the same fact, but only one knows about the cloud resource — so the count can under-report on cloud.

**Root cause is the hardcoding.** Adding resource #N means editing multiple spots, and `group_views.py` can't even `import` the cloud model directly (it doesn't exist in OSS), so it needs the same lazy `apps.get_model` / shared-registry treatment `signals.py` already uses — or a plugin hook the cloud side appends to. Consolidating to a single source of truth (or a `GenericRelation` on a shared `ShareableMixin`, which makes enrollment structural via inheritance) would remove the class of bug entirely.

**Nit:** `signals.py` (OSS) hardcodes a cloud-only model name `("agentic_studio_v1", "AgenticProject")`. It works because `apps.get_model` is lazy and `LookupError`-guarded, but OSS naming a cloud app is a layering smell — a registry the cloud app *extends* would avoid both the drift and the cross-layer reference.

_Not blocking — flagging for a follow-up to collapse these into one source of truth._

chandrasekharan-zipstack (approved):

backend/adapter_processor_v2/urls.py
kirtimanmishrazipstack: @chandrasekharan-zipstack `list_of_shared_users` returns the adapter serialized with only its **direct** `shared_users` (the M2M). `effective-members` (from `ResourceShareManagementMixin`) returns **everyone** who effectively has access — direct users + group members + org-share — priority-deduped via `compute_effective_members`. The former is the raw direct-share list; the latter backs the group-sharing "who can access this" view.

backend/adapter_processor_v2/views.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Tightened to a concise, generic two-liner in 4389e62e9.

backend/permissions/resource_share_views.py
kirtimanmishrazipstack: @chandrasekharan-zipstack These are deferred deliberately. This module hosts the share mixin imported by every shareable viewset, and the helpers (`sharing_helpers`, `group_serializers`) transitively pull resource models across apps — importing them at module top risks an app-load/circular-import error. It's the same lazy pattern used in `signals.py` and the model `shared_groups` properties, so I've left them as-is.

backend/pipeline_v2/models.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Good catch — it was never an M2M. Replaced it with the same concise comment the other 5 resource models use, in 4389e62e9.

backend/tenant_account_v2/signals.py
```diff
@@ -0,0 +1,136 @@
...
+                # agentic_studio_v1 in pure OSS). Skip cleanly.
+                continue
+            try:
+                resources = model.objects.filter(
+                    organization=instance.organization,
+                    shared_users=instance.user,
+                )
+                removed = 0
+                for resource in resources:
+                    resource.shared_users.remove(instance.user)
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in 4389e62e9. The purge now deletes via the M2M through table (unscoped manager) scoped explicitly by the resource's organization, instead of `model.objects` which is `UserContext`-org-scoped and matches zero rows outside an HTTP request. Also collapses the per-row `.remove()` loop into one delete per model.

kirtimanmishrazipstack:
@chandrasekharan-zipstack Fixed in 4389e62e9. Consolidated both lists into a single `SHAREABLE_RESOURCES` registry (`tenant_account_v2/shareable_resources.py`); `signals.py` and `group_views.py` now both iterate it via `apps.get_model`, so they can't drift. `AgenticProject` is included, so the group blast-radius count no longer under-reports on cloud.

kirtimanmishrazipstack:
Follow-up to commit `4389e62e`: corrected the `AgenticProject` descriptor in `SHAREABLE_RESOURCES` — it had `name_field="project_name"`/`id_field="project_id"` (copied from the CustomTool `tool_id`/`tool_name` pattern), but the cloud model (`agentic_studio_v1.AgenticProject`, cloud PR #1508) exposes ORM fields `id` (UUID PK) and `name`. `project_id`/`project_name` exist only as dict keys in `get_pipeline_status()`, not as columns.

Without this, the group blast-radius view would run `values_list("project_id", "project_name")` on cloud → `FieldError`/500. Pure OSS was unaffected (app absent → skipped via `LookupError`), so local pre-commit stayed green.

Fixed in `9c7c2c7a3`. Verified cloud PR #1508 needs no changes — `AgenticProject` is fully wired (`shared_users`/`shared_to_org`/`shared_groups`, `for_user()` union, OSS sharing helpers).

frontend/src/components/widgets/share-permission/SharePermission.jsx
```diff
@@ -66,58 +69,86 @@ function SharePermission({
...
+      setSelectedGroupIds([]);
+    }
   }, [adapter, allUsers]);
 
   const handleDeleteUser = (userId) => {
-    setSelectedUsers((prevSelectedUsers) =>
-      prevSelectedUsers.filter((user) => user !== userId),
-    );
+    setSelectedUsers((prev) => prev.filter((user) => user !== userId));
+  };
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in `e36ccf01e`. `handleDeleteUser` now compares with `String(user) !== String(userId)`, so revoke works regardless of ID type. Confirmed the root cause is a mixed-type `selectedUsers`: pre-loaded shares are seeded stringified (`user.id.toString()`), while in-session additions push the raw numeric `selectedValue` — and the delete button passes the numeric `item.id`. Normalizing both sides covers both cases.

backend/tool_instance_v2/views.py
```diff
@@ -94,21 +95,26 @@ def get_queryset(self) -> QuerySet:
...
+            Q(created_by=user) | Q(workflow__in=accessible_workflows)
+        ).distinct()
         if filter_args:
             queryset = queryset.filter(**filter_args)
         return queryset
 
+    def get_permissions(self) -> list[Any]:
+        # Mutation requires parent-workflow ownership; sibling read access
+        # via shared workflows must not imply mutate rights.
+        if self.action in ("update", "partial_update", "destroy"):
```
jaseemjaskp, line 116: **[IMPORTANT — authorization gap]** This PR adds `IsParentWorkflowOwner` for `update`/`partial_update`/`destroy` and broadens `get_queryset()` so users a workflow is *shared* with can see its `ToolInstance` rows — but the `reorder` custom action (line 182) is **not** in this gate, so it falls back to the project default (`DEFAULT_PERMISSION_CLASSES = []`). Worse, `reorder` → `ToolInstanceHelper.reorder_tool_instances()` resolves each UUID with an **unscoped** `ToolInstance.objects.get(pk=...)` (tool_instance_helper.py:273), so it is not protected by the broadened queryset either. Net effect: any authenticated user can mutate the `step` ordering of *any* workflow's tool instances by posting their UUIDs — the exact sibling-mutation-via-shared-workflow the new gate was added to prevent.

**Fix:** add `"reorder"` to the `IsParentWorkflowOwner` branch of `get_permissions`, and/or validate that `wf_id` / the instance UUIDs belong to a workflow the caller owns before reordering.

backend/adapter_processor_v2/views.py
```diff
@@ -292,91 +295,116 @@ def destroy(
...
+        only here (not via ``partial_update``). Diff the M2M before/after the
+        commit so cleanup keys off who actually lost access.
+        """
+        adapter = self.get_object()
+        before_user_ids = set(adapter.shared_users.values_list("id", flat=True))
+        response = super().share(request, pk)
+        if response.status_code == status.HTTP_200_OK:
+            adapter.refresh_from_db()
+            after_user_ids = set(adapter.shared_users.values_list("id", flat=True))
+            self._clear_default_adapter_for_removed_users(
```
jaseemjaskp: **[IMPORTANT — stale access after unshare]** The default-adapter cleanup only diffs the **direct** `shared_users` M2M (`before_user_ids - after_user_ids`). If a user loses access because their **group** was unshared, or because `shared_to_org` was toggled off, their `UserDefaultAdapter` still points at this adapter — they keep using an adapter they can no longer access, with no cleanup and no signal. (This is distinct from the already-resolved threads about moving cleanup out of `partial_update` / removing `int(user_id)`.)

**Fix:** compute removed *effective* members (diff `compute_effective_members` before/after, or include group-/org-derived user sets) and run `_clear_default_adapter_for_removed_users` over the full removed set — not just direct-share removals. If excluding group/org removals is intentional, document why.

backend/tenant_account_v2/signals.py
```diff
@@ -0,0 +1,132 @@
...
+    """Purge ``ResourceGroupShare`` rows when a shareable resource is deleted.
+
+    ``object_id`` is a plain varchar (no FK/CASCADE), so group-share rows would
+    otherwise dangle indefinitely after the resource is gone.
+    """
+    from django.contrib.contenttypes.models import ContentType
+
+    from tenant_account_v2.models import ResourceGroupShare
+
+    deleted, _ = ResourceGroupShare.objects.filter(
```
jaseemjaskp: **[SUGGESTION — robustness/observability]** `cleanup_resource_group_shares` has no `try/except`, unlike its sibling `cleanup_user_org_access` (which deliberately wraps + re-raises with context). `post_delete` fires inside the resource-deletion transaction, so if this `.delete()` raises, the resource deletion rolls back with a context-free traceback — the operator can't tell group-share cleanup was the culprit. Mirror the sibling: wrap with a log that names `sender`/`instance.pk` and re-raise, so failures are diagnosable.

backend/tenant_account_v2/sharing_helpers.py
```diff
@@ -0,0 +1,541 @@
...
+    groups = list(groups)
+    foreign = [g for g in groups if g.organization_id != organization.id]
+    if foreign:
+        raise ValidationError(
+            {"shared_groups": ("All shared groups must belong to your organization.")}
+        )
+    return groups
+
+
+def get_resource_share_groups(resource_obj: Any) -> QuerySet[OrganizationGroup]:
```
jaseemjaskp, line 66: **[SUGGESTION — defense in depth]** `get_resource_share_groups` (and `resources_visible_via_groups` below) filter only by `content_type` + `object_id` / `group_id__in`, never by `organization`. Tenant isolation here currently rests entirely on UUID global-uniqueness and on write-time validation (`validate_shared_groups_in_org`) + the org filter backend — not on the read query itself. Since `ResourceGroupShare` already stores `organization_id`, adding it to these filters costs one indexed column and makes the tenant boundary explicit at the read layer. (Related: the `organization` FK on `ResourceGroupShare` is derivable from `group.organization`; consider validating they agree at write time in `set_resource_share_groups`, or dropping the column to avoid a representable illegal state where they disagree.)

backend/tenant_account_v2/shareable_resources.py
```diff
@@ -0,0 +1,51 @@
...
+    app_label: str
+    model_name: str
+    kind: str  # serialization key surfaced in API payloads
+    name_field: str  # human-readable display field
+    id_field: str  # primary-key field name
+
+
+# ``agentic_studio_v1`` is cloud-only; consumers resolve it lazily and skip it
+# when the app is not installed.
+SHAREABLE_RESOURCES: tuple[ShareableResource, ...] = (
```
jaseemjaskp, line 27: **[SUGGESTION — catch recurring drift bug]** `name_field` / `id_field` / `app_label` / `model_name` are free strings resolved lazily via `apps.get_model` + `values_list(id_field, name_field)`. A typo surfaces only as a runtime `FieldError`, only on the code path touching that resource — the PR history shows exactly this class of bug (commit `6a5493dc` "Correct AgenticProject field names"). Add a cheap startup/test assertion that resolves each registered descriptor and verifies `id_field`/`name_field` exist via `model._meta.get_field`, so drift is caught at boot/CI instead of in production.

backend/tenant_account_v2/sharing_helpers.py
```diff
@@ -0,0 +1,541 @@
...
+    ORG_AXIS = "shared_to_org"
+
+    AXIS_LABELS = {
+        USERS_AXIS: "shared users",
+        GROUPS_AXIS: "shared groups",
+        ORG_AXIS: "organization sharing",
+    }
+
+    @classmethod
+    def authorize_and_commit(
```
jaseemjaskp, line 357: **[IMPORTANT — test coverage]** This PR introduces a security-sensitive access-control feature (~1,100 LOC) and adds **zero tests** (`tenant_account_v2/tests.py` is still the empty stub). `ShareAuthorizationService.authorize_and_commit` is the heart of the matrix and is entirely untested: group-member-can-add-only-groups-they-belong-to, unprivileged-actor-cannot-remove-any-axis, `shared_to_org` toggle gating, service-account bypass, and the atomicity guarantee (a `PermissionDenied` on one axis must leave *all* axes uncommitted). Equally critical and untested: the two cleanup signals (rejoin-backdoor closure / orphaned-share prevention), `set_resource_share_groups` add/remove/no-op diff, and the `for_user()` group-visibility filter replicated across 7 models.

Please add a parametrized test suite over actor roles for the matrix + the signals before merge. Note: per `backend/conftest.py`, backend tests don't yet run under tox in CI — a `backend` testenv/group must also be wired or the tests won't gate merges.

backend/adapter_processor_v2/views.py
```diff
@@ -292,91 +295,116 @@ def destroy(
...
+                adapter, before_user_ids - after_user_ids
+            )
+        return response
 
-        # Perform the update
-        response = super().partial_update(request, *args, **kwargs)
+    def _notify_shared_users(
```
jaseemjaskp, line 338: **[SUGGESTION — DRY]** `_notify_shared_users` is duplicated near-verbatim across 6 viewsets (here, plus `pipeline_v2/views.py:159`, `connector_v2/views.py:216`, `api_v2/api_deployment_views.py:396`, `workflow_manager/workflow_v2/views.py:151`, `prompt_studio/prompt_studio_core_v2/views.py:307`). All share the identical skeleton: `diff_share_axes(...).get("shared_users")` guard → build notification service from the plugin → `send_sharing_notification(...)` in a try/except. Only 4 values vary (`resource_type`, name attr, id attr, and pipeline's ETL/TASK gate). Consider hoisting a `notify_shared_users(...)` helper onto `ResourceShareManagementMixin` (which already owns `diff_share_axes`), parameterized by those values. ~150 lines saved and one place to fix notification bugs. (The `shared_groups` property and the `for_user()` group-visibility preamble are likewise copy-pasted across the 6 models — both are good mixin candidates.)

frontend/src/components/groups/groups-service.js
```diff
@@ -0,0 +1,90 @@
+import { useAxiosPrivate } from "../../hooks/useAxiosPrivate";
+import { useSessionStore } from "../../store/session-store";
+
+const RESOURCE_PATHS = {
+  workflow: "workflow",
+  pipeline: "pipeline",
+  api_deployment: "api/api",
```
jaseemjaskp: **[SUGGESTION — latent bug]** `api_deployment: "api/api"` builds `/.../api/api/<id>/effective-members/`, but the backend route is `deployment/<uuid:pk>/effective-members/` (api_v2/urls.py:74) — the correct segment is `api/deployment`. Every other mapping in this table matches its backend route; only this one is wrong. It's currently latent (`getEffectiveMembers` has no callers yet) but will 404 the moment the effective-members UI is wired up.

**Fix:** `api_deployment: "api/deployment"`.

frontend/src/index.jsx
```diff
@@ -39,15 +38,13 @@ setFavicon(config.favicon);
...
-      <SocketProvider>
-        <LazyLoader
-          loader={<GenericLoader />}
-          component={() => import("./App.jsx")}
-          componentName="App"
-        />
-      </SocketProvider>
-    </PostHogProvider>
-  </React.StrictMode>,
+  <PostHogProvider client={posthog}>
```
jaseemjaskp: **[SUGGESTION — out of scope]** This drops the `<React.StrictMode>` wrapper (and the `React` import) at the app root, disabling dev-time double-invocation / deprecation checks app-wide. That's unrelated to group sharing. If unintentional, restore it; if intentional, it belongs in a separate, justified change rather than bundled into this feature PR.

frontend/src/hooks/useShareModal.js
```diff
@@ -56,34 +67,53 @@ function useShareModal({
...
           content: "Sharing permissions updated successfully",
         });
-        setOpenShareModal(false);
         refreshRef.current?.();
+        // Close only on success; keep the modal open on failure so the user
+        // can see the rejected entries and retry.
+        setOpenShareModal(false);
       })
       .catch((err) => {
         setAlertDetails(handleException(err));
```
jaseemjaskp: **[SUGGESTION — error UX]** Unlike `handleShare` (line 90, which passes a fallback `"Unable to fetch sharing information"`), this share-apply catch calls `handleException(err)` with no fallback string. If the backend returns an error shape `handleException` can't parse, the user gets an empty/generic alert with no context about what failed. Pass a fallback, e.g. `handleException(err, "Unable to update sharing permissions")`. (Keeping the modal open on failure here is good.)

frontend/src/components/groups/GroupMemberManager.jsx
```diff
@@ -0,0 +1,161 @@
...
+  const loadMembers = () => {
+    if (!group?.id) {
+      return;
+    }
+    setLoading(true);
+    Promise.all([service.listGroupMembers(group.id), service.getAllOrgUsers()])
+      .then(([memberRes, usersRes]) => {
+        setMembers(memberRes?.data || []);
+        const all = usersRes?.data?.members || [];
+        setOrgUsers(
```
jaseemjaskp: **[SUGGESTION — silent empty list]** `const all = usersRes?.data?.members || [];` assumes only the `{members: [...]}` response shape, whereas `useShareModal` (lines 51-67) defensively handles three shapes (bare array / `.members` / `.users`). If `getAllOrgUsers` ever returns a bare array or `.users`, the candidate list silently becomes empty and the admin sees "no users to add" with no error. Reuse the same multi-shape normalization, or assert the expected shape and surface a real alert on mismatch.

jaseemjaskp (commented):
Automated multi-agent review (Code Reviewer, Silent-Failure Hunter, Type-Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier). Findings below are de-duplicated against already-resolved review threads; each was verified against the code at HEAD. Grouped roughly: authorization/correctness first, then robustness, then maintainability/quality.

backend/tool_instance_v2/views.py
kirtimanmishrazipstack: Fixed in b1a840d8c. `reorder` now resolves the workflow via `Workflow.objects.for_user()` (scopes out cross-org → 404) and requires owner / org-admin / service-account via a new `is_workflow_mutator` predicate extracted from `IsParentWorkflowOwner`. `reorder_tool_instances` takes a `workflow_id` and scopes its lookup so a stray UUID can't reach a foreign row. (The serializer already validated instance↔workflow membership; the missing piece was actor authorization, which this adds.)

backend/adapter_processor_v2/views.py
kirtimanmishrazipstack: Fixed in b1a840d8c. `share()` now diffs *effective* access via `compute_effective_members` (direct + group + org) before/after the commit and clears `UserDefaultAdapter` for everyone in `before − after` — so access lost through group-unshare or `shared_to_org` toggle-off is cleaned too, not just direct removals.

backend/tenant_account_v2/signals.py
kirtimanmishrazipstack: Fixed in b1a840d8c — wrapped the `.delete()` in a try/except that logs `sender`/`instance.pk` and re-raises, mirroring `cleanup_user_org_access`.

backend/tenant_account_v2/sharing_helpers.py
kirtimanmishrazipstack: Done in b1a840d8c. Added an explicit `organization` filter to `get_resource_share_groups` and `list_resources_shared_with_group`, and `set_resource_share_groups` now rejects group ids outside the resource's org (closing the org-mismatch illegal state). `resources_visible_via_groups` is left as-is — it's already bounded by the caller's own `group_id__in` (org-implicit), noted in a comment.

backend/tenant_account_v2/shareable_resources.py
kirtimanmishrazipstack: Done in b1a840d8c — added a Django system check (`tenant_account_v2/checks.py`) that resolves every `SHAREABLE_RESOURCES` descriptor and asserts `id_field`/`name_field` exist via `model._meta.get_field`, skipping uninstalled cloud-only apps. Drift now fails at `manage.py check` / CI / boot, and it's also covered by a test.

backend/tenant_account_v2/sharing_helpers.py
kirtimanmishrazipstack: Added in b1a840d8c — `tenant_account_v2/tests.py` now has 15 tests: the full `ShareAuthorizationService` matrix (owner / org-admin / direct-shared / group-member / unprivileged / service-account across all three axes, incl. the **atomicity** guarantee that a `PermissionDenied` on one axis leaves every axis uncommitted), both cleanup signals (rejoin-backdoor + orphan prevention), `set_resource_share_groups` add/remove/no-op + cross-org guard, `for_user()` group visibility, and the registry. The suite also caught a real crash — `cleanup_user_org_access` blew up when a registered model lacks `shared_users` (the OSS↔cloud sync window for `AgenticProject`); now fixed with a defensive skip. **Re: tox** — backend tests still aren't wired into CI (needs an isolated test DB + `unstract`-schema bootstrap); flagging that as a separate infra task rather than bundling a shared-CI change into this PR.

frontend/src/components/groups/groups-service.js
kirtimanmishrazipstack: Fixed in b1a840d8c — `api_deployment: "api/deployment"` (matches the backend `api/deployment/<id>/effective-members/` route).

frontend/src/index.jsx
kirtimanmishrazipstack: Restored in b1a840d8c — re-added the `React` import and wrapped the root render in `<React.StrictMode>`.

frontend/src/hooks/useShareModal.js
kirtimanmishrazipstack: Fixed in b1a840d8c — `onShare`'s catch now passes a fallback: `handleException(err, "Unable to update sharing permissions")`, matching `handleShare`.

frontend/src/components/groups/GroupMemberManager.jsx
kirtimanmishrazipstack: Fixed in b1a840d8c — `GroupMemberManager` now normalizes the `getAllOrgUsers` response across all three shapes (bare array / `.members` / `.users`) via a small local helper, matching `useShareModal`'s defensive handling.

backend/adapter_processor_v2/views.py
kirtimanmishrazipstack: Acknowledged — the duplication is real, but this refactor is out of scope for the current PR and isn't needed here. Keeping _notify_shared_users per-viewset for now. The shared-notification path can be consolidated as separate follow-up work.

jaseemjaskp (approved):

backend/adapter_processor_v2/views.py
```diff
@@ -292,91 +295,125 @@ def destroy(
...
+        """
+        adapter = self.get_object()
+        before_user_ids = self._effective_member_ids(adapter)
+        response = super().share(request, pk)
+        if response.status_code == status.HTTP_200_OK:
+            adapter.refresh_from_db()
+            after_user_ids = self._effective_member_ids(adapter)
+            self._clear_default_adapter_for_removed_users(
+                adapter, before_user_ids - after_user_ids
+            )
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Fixed in e7ff61294. The owner is now discarded from the removed set before `_clear_default_adapter_for_removed_users`, so a `shared_to_org` toggle-off (which drops the owner from the org-member set in `compute_effective_members`) no longer wipes their `UserDefaultAdapter`. The owner always retains access via `created_by`, so their defaults are never subject to share-axis cleanup; non-owners who genuinely lose access are still cleaned up.

johnyrahul (approved):

PR #1989, opened 2026-05-24
UN-3479 [FEAT] Org admin override on resources + ProfileManager sharing

chandrasekharan-zipstack:
## Summary

Brings org admins onto a "manage everything in their org" model and gives ProfileManager the same sharing surface every other shareable resource has. Together this unblocks org-to-org migration: resources land owned by the service account, and the admin can see and edit them all without any per-resource ownership transfer.

### 1. Admin override on permission classes (commit 2)

`backend/permissions/permission.py`

- `IsOwner`, `IsOwnerOrSharedUser`, `IsOwnerOrSharedUserOrSharedToOrg` all admit org admins in addition to the existing rules.
- New helper `OrganizationMemberService.is_user_organization_admin(user)` — single membership lookup, used by both the permission classes and the manager `for_user()` methods.
- Per-request cache (`request._cached_is_organization_admin`) keeps the lookup at one query per HTTP call regardless of how many permission classes ask.

### 2. `for_user()` managers admit admins

Every `for_user()` manager (`AdapterInstance`, `ConnectorInstance`, `CustomTool`, `Workflow`, `WorkflowExecution`, `Pipeline`, `APIDeployment`, `ProfileManager`) short-circuits to the org-scoped queryset for admins, mirroring the existing service-account branch. Org scoping continues to come from `DefaultOrganizationManagerMixin` / `OrgAwareManager` — admins do not gain cross-org visibility.

### 3. ProfileManager joins the standard sharing model (commit 1)

Before: `ProfileManager` had no sharing fields, used `permission_classes = [IsOwner]` on every action, and `get_queryset` filtered by `created_by` only. A profile created by anyone else was invisible and uneditable — fine when each user kept to their own profiles, but a hard wall for org-to-org migration.

After:

- `shared_users` (M2M) and `shared_to_org` (Bool, default False) added — mirrors `Workflow` / `Pipeline` / etc.
- `ProfileManagerModelManager.for_user()` — same Q-filter, plus the admin / service-account short-circuits.
- ViewSet uses per-action permissions: `IsOwner` for write, `IsOwnerOrSharedUserOrSharedToOrg` for read. Same pattern every other shareable resource follows.
- `get_queryset` uses `for_user(request.user)`.
- Migration `0005_profilemanager_shared_to_org_and_more` generated via `makemigrations`.

## Effect on the migration story

| Scenario | Before | After |
|----------|--------|-------|
| Admin views a workflow created by another user | invisible unless shared | visible (admin override) |
| Admin edits a workflow created by another user | 403 | OK |
| Admin views a migrated profile (created by service account) | invisible | visible |
| Admin edits a migrated profile | 403 | OK |
| Non-admin User-A views resource created by User-B (both humans) | invisible unless shared | unchanged — admin override does not apply |
| Cross-org visibility | not allowed | not allowed (org scoping unchanged) |

## What this PR deliberately doesn't do

- Doesn't change behavior for non-admin users — still owner/shared-only.
- Doesn't backfill existing rows to `shared_to_org=True`. Pre-existing rows keep their current visibility for non-admins.
- Doesn't touch file-transport paths (`fetch_contents_ide` / `upload_for_ide`).

## Test plan

- [ ] `python manage.py migrate prompt_profile_manager_v2` against a local DB — columns appear, no data loss.
- [ ] Smoke: as a non-admin user, verify resources created by other users are still invisible / uneditable.
- [ ] Smoke: as an org admin, verify list views show every resource in the org (workflows / pipelines / api deployments / connectors / custom tools / adapters / profiles) and PATCH/DELETE succeeds on each.
- [ ] Smoke: confirm visibility stays org-scoped — admin in Org A cannot see Org B's resources.
- [ ] Confirm `_is_organization_admin` is cached on the request (instrument with logging if needed).


backend/prompt_studio/prompt_profile_manager_v2/models.py
```diff
@@ -3,12 +3,30 @@
...
+        every other org member.
+        """
+        if getattr(user, "is_service_account", False):
+            return self.all()
+
+        from django.db.models import Q
+
+        return self.filter(
+            Q(created_by=user) | Q(shared_users=user) | Q(shared_to_org=True)
+        ).distinct()
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Fixed in 668baacda. Scoped `for_user` through `prompt_studio_tool__organization=UserContext.get_organization()` on every branch (service-account, admin, and the OR clause), so the `shared_to_org=True` row can no longer leak across tenants. ProfileManager has no direct org FK so we couldn't use `DefaultOrganizationManagerMixin` directly — scoped via the parent CustomTool instead.

backend/permissions/permission.py
```diff
@@ -21,13 +23,56 @@ def _is_service_account(request: Request) -> bool:
...
+        return cached
+    try:
+        from account_v2.authentication_controller import AuthenticationController
+
+        auth_controller = AuthenticationController()
+        member = auth_controller.get_organization_members_by_user(user=request.user)
+        is_admin = bool(member) and auth_controller.is_admin_by_role(member.role)
+    except Exception:
+        is_admin = False
+    setattr(request, _REQUEST_ADMIN_CACHE_ATTR, is_admin)
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Obsolete — the broad except was already removed in 42045dd82 (the follow-up commit that switched to OrganizationMemberService.is_user_organization_admin). Current code lets exceptions propagate.

backend/prompt_studio/prompt_profile_manager_v2/views.py
```diff
@@ -23,18 +23,21 @@ class ProfileManagerView(viewsets.ModelViewSet):
     """Viewset to handle all Custom tool related operations."""
 
     versioning_class = URLPathVersioning
-    permission_classes = [IsOwner]
     serializer_class = ProfileManagerSerializer
 
+    def get_permissions(self) -> list[Any]:
+        if self.action in ("destroy", "partial_update", "update"):
+            return [IsOwner()]
+        return [IsOwnerOrSharedUserOrSharedToOrg()]
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Fixed in 4a283047f. Added `create` to the IsOwner action tuple alongside destroy/partial_update/update.

backend/tenant_account_v2/organization_member_service.py
```diff
@@ -13,6 +14,29 @@ def get_user_by_email(email: str) -> OrganizationMember | None:
...
+        if getattr(user, "is_service_account", False):
+            return False
+        try:
+            member = OrganizationMember.objects.get(user=user.id)  # type: ignore
+        except OrganizationMember.DoesNotExist:
+            return False
+        except Exception:
+            return False
+        try:
+            return UserRole(member.role) == UserRole.ADMIN
```
jaseemjaskp: **[correctness] Role string isn't normalized — diverges from the canonical admin check.** `member.role` is a bare `CharField()` (no `choices=`), so `UserRole(member.role)` raises `ValueError` for any non-exact-lowercase value (`"Admin"`, `"ADMIN"`), which the `except ValueError` then turns into a silent *non-admin* — denying a legitimate admin all the new visibility/permissions.

The established defensive helper `AuthenticationService.is_admin_by_role` (`account_v2/authentication_service.py:158`) uses `UserRole(role.lower())` for exactly this reason, and the sibling admin check added for platform API keys (`platform_api/permissions.py:20`) routes through it. This helper should match:
```python
return UserRole((member.role or "").lower()) == UserRole.ADMIN
```
or better, reuse `is_admin_by_role`. Works today only because the OSS auth backend happens to store lowercase — latent for any backend (SSO/auth0) that doesn't.

backend/tenant_account_v2/organization_member_service.py
```diff
@@ -13,6 +14,29 @@ def get_user_by_email(email: str) -> OrganizationMember | None:
...
+        """
+        if not user or not getattr(user, "is_authenticated", False):
+            return False
+        if getattr(user, "is_service_account", False):
+            return False
+        try:
+            member = OrganizationMember.objects.get(user=user.id)  # type: ignore
+        except OrganizationMember.DoesNotExist:
+            return False
+        except Exception:
```
jaseemjaskp, line 35: **[observability] Broad `except Exception: return False` silently swallows infra failures with no logging.** Any DB error (`OperationalError`, `ProgrammingError`) or `MultipleObjectsReturned` is masked as "not an admin" — during a partial DB outage every admin org-wide silently loses access with no server-side signal (the module doesn't import `logging` at all). The sibling admin check added in this same effort logs it (`platform_api/permissions.py:22`: `logger.exception(...)`).

Keep fail-closed, but surface the unexpected branch:
```python
except Exception:
    logger.exception("admin-role lookup failed for user %s; denying", getattr(user, "id", None))
    return False
```
Note `MultipleObjectsReturned` is reachable: `OrganizationMember.user` is a non-unique FK (only `(organization, user)` is unique), so this depends on org context being set on the request.

backend/tenant_account_v2/organization_member_service.py
```diff
@@ -13,6 +14,29 @@ def get_user_by_email(email: str) -> OrganizationMember | None:
         except OrganizationMember.DoesNotExist:
             return None
 
+    @staticmethod
+    def is_user_organization_admin(user: Any) -> bool:
```
jaseemjaskp, line 20: **[type-design] `user: Any` weakens an authorization boundary.** Every caller passes `request.user` / a `User`. The real contract is a Django user-like object exposing `.id`, `.is_authenticated`, `.is_service_account`, and the function explicitly handles the anonymous case — so `User | AnonymousUser | None` restores type-checker coverage exactly where it matters most. The adjacent methods in this class already type their args (`email: str`, `user_id: str`, `member_id: int`).

backend/prompt_studio/prompt_profile_manager_v2/models.py
```diff
@@ -3,12 +3,48 @@
...
+
+        ProfileManager has no direct ``organization`` FK — scope via the
+        parent CustomTool so the ``shared_to_org=True`` branch cannot
+        leak rows across tenants when a UUID is known/guessed.
+        """
+        # Service accounts and admins still need to be org-scoped — they
+        # otherwise see rows from every org in the DB.
+        from django.db.models import Q
+
+        org_scope = Q(prompt_studio_tool__organization=UserContext.get_organization())
```
jaseemjaskp, line 23: **[robustness] `for_user` fails *closed* outside request context, unlike the sibling managers.** When `UserContext.get_organization()` returns `None` (Celery / management commands), `org_scope` becomes `prompt_studio_tool__organization IS NULL` and every branch returns ~zero rows. `DefaultOrganizationManagerMixin` / `OrgAwareManager` instead fail *open* (return unfiltered) when org is `None`. Today the only caller is the request-context viewset (`views.py:37`) so it's unreachable, but it's a latent trap if `for_user` is reused in a task. Consider applying `org_scope` only when org is not `None`, mirroring the siblings.

jaseemjaskp (commented):
## PR Review — UN-3479 Org admin override + ProfileManager sharing

Ran the PR Review Toolkit (comment-analyzer, pr-test-analyzer, silent-failure-hunter, type-design-analyzer, code-reviewer, code-simplifier). Line-specific findings are in the inline comments; the summary is below. Items already addressed in 4a283047 / 42045dd8 / 668baacd are not re-raised.

### ✅ Cross-org isolation verified safe (the core security claim)
Traced every admin `for_user()` branch to its base manager. The five that return bare `self.all()` / `self.get_queryset()` (`api_v2`, `connector_v2`, `pipeline_v2`, `prompt_studio_core_v2`, `workflow_v2/workflow.py`) all inherit `DefaultOrganizationManagerMixin`, whose `get_queryset()` filters by `UserContext.get_organization()`; `execution.py` and `ProfileManager` scope explicitly. `is_user_organization_admin` itself queries an org-scoped `OrganizationMember`. No cross-org leak found — admins do not gain cross-org visibility, as claimed. The 668baacd `shared_to_org` fix is correct: `org_scope` is AND-ed on all three branches.

### 🔴 Biggest gap: zero tests for a security-sensitive permission change
The PR adds no tests. For a change that broadens who can read/edit/delete every resource in an org, these regressions should be pinned (the repo's DB-free `sys.modules`-mock convention — see `usage_v2/tests/test_helper.py` — covers most without a DB):
- `is_user_organization_admin`: admin role → True; **non-admin role → False** (the non-regression case); anonymous / service-account / no-membership / invalid-role / DB-error → False.
- Each permission class (`IsOwner`, `IsOwnerOrSharedUser`, `IsOwnerOrSharedUserOrSharedToOrg`): admin allowed when not owner/shared; **non-admin non-owner denied**; service-account short-circuits before the admin lookup.
- Per-request cache in `permission.py`: one lookup per request across multiple checks; cached `False` honored (the `cached is not None` guard).
- `ProfileManager.for_user`: regression test that the 668baacd `shared_to_org` cross-tenant scoping holds; `get_permissions` mutation-vs-read split.

### 🟠 Friction-less adapter inconsistency in the admin context (extends greptile's note on `adapter_processor_v2/models.py:45`)
The admin `for_user()` branch now lists friction-less adapters, but `IsFrictionLessAdapter.has_object_permission` (`permissions/permission.py:109`) returns `False` for `is_friction_less` objects unconditionally — no admin or service-account override. Net effect: an admin sees friction-less adapters in list views but gets 403 on object-level access. Decide the intended semantics and make listing and object-permission agree (add the admin override to the friction-less permission classes, or exclude friction-less rows from the admin `for_user()` branch the way the non-admin branch's `Q(is_friction_less=True)` is the only thing that surfaces them).

### 🟡 Lower-priority
- **Uncached `for_user` admin lookup.** `permission.py` caches the admin check per-request, but the manager `for_user()` methods call `is_user_organization_admin` directly (one `OrganizationMember.objects.get` each). Mostly once-per-request, but `prompt_studio_core_v2/serializers.py:116` calls `AdapterInstance.objects.for_user()` inside a field validator. Consider sharing the request-scoped cache.
- **Broader ProfileManager org-scoping.** Org isolation now lives only in `for_user`; the raw `ProfileManager.objects.get/filter` callsites and the `IsOwner` mutation path aren't org-scoped structurally (unlike siblings that inherit `DefaultOrganizationMixin`). Worth a follow-up to confirm none operate cross-org, plus a class-docstring note that the model is scoped only transitively via `prompt_studio_tool`.
- **Simplifications (no behavior change).** Hoist the inline `from django.db.models import Q` to module level in `api_v2/models.py` and `workflow_v2/models/workflow.py` to match the other files; the repeated inline `OrganizationMemberService` import + admin branch across 6 managers could collapse into one `DefaultOrganizationManagerMixin` helper method (leave `execution.py` / `ProfileManager` explicit org-scoping as-is).
- **Docstring nits.** `is_user_organization_admin` docstring omits the unrecognized-role and `MultipleObjectsReturned` outcomes; the `self.all()` `for_user` docstrings would benefit from a one-line note that scoping comes from `DefaultOrganizationManagerMixin` — guards a future maintainer from reintroducing the leak.

backend/adapter_processor_v2/models.py
```diff
@@ -37,6 +37,13 @@ def for_user(self, user: User) -> QuerySet[Any]:
         if getattr(user, "is_service_account", False):
             return self.get_queryset().filter(is_friction_less=False)
 
+        from tenant_account_v2.organization_member_service import (
+            OrganizationMemberService,
+        )
+
+        if OrganizationMemberService.is_user_organization_admin(user):
+            return self.get_queryset()
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Addressed in 67edc59d2, with one deliberate narrowing.

Friction-less adapters wrap **platform-owned** credentials (created automatically during onboarding so users don't have to BYOK). Even with `unstract_key` masked in `to_representation`, exposing their `retrieve`/`update` would let an admin swap the masked key for a real one. Policy: never expose those creds to anyone.

So the fix splits along that axis:

- `IsFrictionLessAdapter`: `obj.is_friction_less` → still `False` for everyone, admins included. Non-frictionless branch now admits org admins.
- `IsFrictionLessAdapterDelete`: friction-less still deletable by any org member (unchanged). Non-frictionless branch now admits org admins.

Net effect: admins manage every teammate-owned adapter via retrieve/update/destroy, but friction-less platform-credential rows remain off-limits.

backend/tenant_account_v2/organization_member_service.py
chandrasekharan-zipstack: Fixed in 67edc59d2. `is_user_organization_admin` now delegates to `AuthenticationController.is_admin_by_role(member.role)`, which routes through the active auth plugin and lower-cases the role — so casing variants from SSO/Auth0 backends (`"Admin"`, `"ADMIN"`) resolve correctly. Lazy-imported inside the method to avoid the `AuthenticationController` → `OrganizationMemberService` cycle.

backend/tenant_account_v2/organization_member_service.py
chandrasekharan-zipstack: Fixed in 67edc59d2. Added `logger = logging.getLogger(__name__)` and changed the broad `except Exception` to `logger.exception("admin-role lookup failed for user %s; denying", getattr(user, "id", None))` before returning False. Stays fail-closed, but a partial DB outage now surfaces in logs instead of silently stripping admin access.

backend/tenant_account_v2/organization_member_service.py
chandrasekharan-zipstack: Skipping in this PR — keeping `user: Any` for now.

Reasoning: tightening to `User | AnonymousUser | None` accurately describes the cases the function handles, but service-account `request.user` is a wrapped user that doesn't always satisfy the `User` type checker-side, and the helper explicitly guards every access via `getattr(..., False)` / `.id`. Net value is documentation, not enforcement, and the function isn't an authorization sink on its own — callers (permissions/managers) decide what the boolean means. Happy to revisit if we add a typed `RequestUser` protocol elsewhere.

backend/prompt_studio/prompt_profile_manager_v2/models.py
chandrasekharan-zipstack: Acknowledged but skipping — preferring fail-closed here.

The only caller today is the request-context viewset, as you noted. If a Celery task starts using `for_user` outside a request, the symptom (empty queryset) is loud and immediately diagnosable, whereas a silent fail-open would leak cross-tenant rows whenever org context happens to be unset. Trading a discoverable Celery-side bug for a silent multi-tenant leak feels wrong for security-sensitive code. Will revisit if/when we add a non-request caller.

chandrasekharan-zipstack:
Pushed 67edc59d2 to address the actionable findings. Quick rundown:

**Addressed**
- `IsFrictionLessAdapter` / `IsFrictionLessAdapterDelete`: admin bypass added on the non-frictionless branch. Frictionless adapters remain blocked for everyone (admins included) since they wrap platform-owned credentials — exposing them would let an admin swap the masked `unstract_key` for a real one.
- `is_user_organization_admin`: role check now delegates to `AuthenticationController.is_admin_by_role`, normalizing case via the active auth plugin (closes the `"Admin"`/`"ADMIN"` latent SSO issue).
- Broad `except Exception` in `is_user_organization_admin` now logs via `logger.exception` before returning False.

**Deferred**
- **Tests** — pinning regressions for the permission/manager admin branch will land after the test-rig PR goes in. Will add `is_user_organization_admin` cases (admin role → True, non-admin → False, anonymous → False, service account → False, member missing → False, unknown role → False) plus at least one per-manager admin-branch sanity test.
- **`user: Any` → narrowed type** — keeping `Any` for now; function is fully guard-shielded and tightening would mostly add bikeshed.
- **Cache shared across manager + permission layers** — keeping per-request cache in `permission.py` only; one extra DB hit per admin retrieve is acceptable.
- **ProfileManager `for_user` fail-open outside request context** — preferring fail-closed for security code; no current caller hits it.
- **Docstring nits** — will fold into the test PR.

Deepak-Kesavan (approved):

kirtimanmishrazipstack (approved):

PR #1991, opened 2026-05-25
UN-3332 [FIX] Use default profile for export validation when single-pass is enabled

praveen-formido:
## What

- Fixes [UN-3332](https://zipstack.atlassian.net/browse/UN-3332): when Single Pass mode is enabled, Prompt Studio tool export incorrectly prioritised the profile attached to the prompt card at creation time instead of the project's default profile.
- Two symptoms, one root cause:
  1. Export was rejected with *"Prompt Studio project without prompts cannot be exported. Please ensure there is at least one active prompt that has been run before exporting."* even after a successful single-pass run.
  2. Even when export succeeded, the per-prompt entries in the exported tool embedded the stale prompt-level profile's LLM / vector DB / embedding / x2text settings rather than the default profile that was actually executed.

## Why

- Single-pass execution runs every prompt through the tool's **default** profile and stores the `PromptStudioOutputManager` rows against that default profile with `is_single_pass_extract=True` (see `OutputManagerHelper.handle_prompt_output_update`).
- `frame_export_json`, however, filtered the validation lookup by `prompt.profile_manager` (the prompt-card FK, frozen at creation time) and built the per-prompt export JSON from the same FK.
- When the user added a second profile and set it as default before running (exactly the ticket's repro), the prompt-level profile and the default profile diverged — so the lookup missed the single-pass rows (false-negative validation) and the export embedded the wrong profile's settings. The customer workaround (making the two profiles match) masked this because both sources then pointed to the same object.

## How

- `backend/prompt_studio/prompt_studio_registry_v2/prompt_studio_registry_helper.py`: resolve `output_profile` once per prompt — the tool default profile when `single_pass_extraction_mode` is on, otherwise `prompt.profile_manager`. This single profile now drives **both**:
  - the `PromptStudioOutputManager` validation lookup (with a matching `is_single_pass_extract` flag), and
  - every per-prompt export field (vector_db, embedding, llm, x2text, chunk_size, chunk_overlap, retrieval_strategy, similarity_top_k, section, reindex).

- No regression expected. Non-single-pass behaviour is unchanged: `output_profile` resolves to `prompt.profile_manager`, exactly as before (the only addition is an explicit `is_single_pass_extract=False` on the lookup, which matches how non-single-pass rows are stored).
- The single-pass branch only changes which profile is consulted, aligning export with the profile the prompts were actually run against.
- `force_export=True` still bypasses validation entirely.

## Database Migrations

- None.

## Env Config

- None.

## Relevant Docs

- N/A

## Related Issues or PRs

- [UN-3332](https://zipstack.atlassian.net/browse/UN-3332)

## Dependencies Versions

- None.

## Notes on Testing

- New regression tests: `backend/prompt_studio/prompt_studio_registry_v2/tests/test_frame_export_json.py` (3 cases — single-pass, non-single-pass, `force_export=True`). They assert both the validation filter args and the per-prompt export JSON content per mode. Same `sys.modules`-stub pattern as `test_build_index_payload.py` (no `pytest-django` dependency).
- Verified the tests fail on pre-fix code (wrong profile / missing `is_single_pass_extract`) and pass on the fix; all 7 `prompt_studio` backend tests pass: `pytest prompt_studio/`.
- Manual repro from the ticket: add Profile 1 → add document → add prompt (don't run) → add Profile 2 → set Profile 2 as default → enable single-pass → run prompt → export. Export should now succeed, and the exported tool should carry Profile 2's settings.

## Screenshots

_N/A — backend validation/export logic change only._

## Checklist

[UN-3332]: https://zipstack.atlassian.net/browse/UN-3332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[UN-3332]: https://zipstack.atlassian.net/browse/UN-3332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/prompt_studio/prompt_studio_registry_v2/tests/test_frame_export_json.py
```diff
@@ -0,0 +1,318 @@
...
+    # time. Stub them so the helper's ``from .models import ...`` and
+    # ``from .serializers import ...`` resolve without booting Django.
+    _install(
+        "prompt_studio.prompt_studio_registry_v2.models",
+        {"PromptStudioRegistry": MagicMock(name="PromptStudioRegistry")},
+    )
+    _install(
+        "prompt_studio.prompt_studio_registry_v2.serializers",
+        {"PromptStudioRegistrySerializer": MagicMock(name="PromptStudioRegistrySerializer")},
+    )
```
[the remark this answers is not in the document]
praveen-formido: Closing as not actionable. The sibling test module `backend/prompt_studio/prompt_studio_core_v2/tests/test_build_index_payload.py` uses the same `_install`/`_install_package` pattern without restoration — adding teardown here alone would be inconsistent. Both files stub functionally equivalent MagicMock collaborators, and running them together passes (7/7) — no observed pollution. If we want to harden this pattern, it should be a separate change applied to both test modules with a shared helper, not bolted onto this PR.

backend/prompt_studio/prompt_studio_registry_v2/tests/test_frame_export_json.py
```diff
@@ -0,0 +1,318 @@
...
+        f"prompt_studio_registry_helper could not be imported in this "
+        f"environment: {type(exc).__name__}: {exc}"
+    )
+    PromptStudioRegistryHelper = None  # type: ignore[assignment]
+    _psrh_mod = None  # type: ignore[assignment]
+
+
+pytestmark = pytest.mark.skipif(
+    _IMPORT_ERROR is not None, reason=_IMPORT_ERROR or ""
+)
```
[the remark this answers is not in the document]
praveen-formido: Addressed in 4ff2b3c. Narrowed the import-time `except Exception` to `(ImportError, ModuleNotFoundError, AttributeError)` so real helper-internal failures surface instead of being silently skipped.

praveen-formido:
@greptile-apps — addressed in 4ff2b3c.

The per-prompt JSON divergence is the same root cause as UN-3332, so I folded it into this PR: lifted `output_profile` out of the validation block and used it for both the `PromptStudioOutputManager` lookup AND every `prompt.profile_manager.X` read in the per-prompt export entry (vector_db, embedding, llm, x2text, chunk_size, chunk_overlap, retrieval_strategy, similarity_top_k, section, reindex).

Tests extended in the same commit — `_assert_output_uses_profile` checks the per-prompt JSON content against the expected profile for each mode. Verified the new assertions correctly fail when the per-prompt block is reverted, so the regression coverage gap you noted is now closed.

Re. CodeRabbit's `sys.modules` cleanup suggestion: I left it as-is for consistency with the sibling `test_build_index_payload.py` (closed inline with reasoning).

backend/prompt_studio/prompt_studio_registry_v2/prompt_studio_registry_helper.py
```diff
@@ -318,45 +318,55 @@ def frame_export_json(
             if not prompt.profile_manager:
                 prompt.profile_manager = default_llm_profile
 
+            # Single-pass execution always runs prompts via the tool's
```
chandrasekharan-zipstack: @pk-zipstack make the comment more concise and generic

chandrasekharan-zipstack (approved):
LGTM, we might have to revisit the tests later

backend/prompt_studio/prompt_studio_registry_v2/prompt_studio_registry_helper.py
praveen-formido: Done in 356f795d — collapsed the 8-line comment down to two lines stating just the rule:

```python
# Single-pass runs every prompt via the tool's default profile;
# otherwise each prompt uses its own profile_manager.
```

praveen-formido: Done in 356f795d — collapsed the 8-line comment down to two lines stating just the rule:

```python
# Single-pass runs every prompt via the tool's default profile;
# otherwise each prompt uses its own profile_manager.
```

Deepak-Kesavan (approved):

PR #1992, opened 2026-05-25
[FIX] Replace unmaintained pytest-dotenv with direct dotenv loading

jaseemjaskp:
## What

Replace the unmaintained `pytest-dotenv==0.5.2` plugin with a 4-line `conftest.py` that calls `python-dotenv` directly. Affects `backend/` and `prompt-service/` only.

## Why

`pytest-dotenv` has not had a release since **February 2020** (~6 years). It monkey-patches private `_pytest.config.Config` internals to inject its env-loading hook, which makes it a recurring break risk across pytest major upgrades (7 → 8 → 9). Surfaced as a MEDIUM finding in the review of #1981.

The plugin's entire behavior in this repo is one side effect: read `env_files` from `[tool.pytest.ini_options]` and call `dotenv.load_dotenv()` before collection. We can do that ourselves in 4 lines and drop a third-party dep that touches pytest internals.

## How

**Backend** (`backend/`):
- Removed `pytest-dotenv==0.5.2` from `[dependency-groups].test`
- Removed `env_files = "test.env"` from `[tool.pytest.ini_options]`
- Added `backend/conftest.py` that calls `load_dotenv(Path(__file__).parent / "test.env", override=False)` — `override=False` mirrors pytest-dotenv's default of letting existing env vars win

**Prompt-service** (`prompt-service/`):
- Removed `pytest-dotenv==0.5.2` from `[dependency-groups].test`
- Added `prompt-service/conftest.py` that calls `load_dotenv(Path(__file__).parent / ".env", override=False)` — pytest-dotenv's default with no `env_files` set is to load `.env` from rootdir, so this preserves exact prior behavior

`python-dotenv` is already in both lockfiles (transitive via `unstract-sdk1` and direct in prompt-service), so no new dependency is added — pytest-dotenv is the only thing removed.

No behavior change expected. The new conftest.py replicates pytest-dotenv's semantics 1:1:

| Behavior | pytest-dotenv | New conftest.py |
|---|---|---|
| When env file is loaded | Before collection (`pytest_load_initial_conftests` hook) | At top-level conftest import (also before collection) |
| Missing file | Silently tolerated | Silently tolerated (load_dotenv default) |
| Existing env var precedence | Process env wins (default) | `override=False` → process env wins |
| File path resolution | Relative to rootdir | `Path(__file__).parent / "file"` → service root (same place) |

Possible (low-likelihood) edge cases:
- If any code relied on pytest-dotenv being importable as a package (it isn't — no public API; we don't import it anywhere)
- If any test runs pytest from a non-default working directory: pytest-dotenv resolved paths against rootdir; the new conftest resolves against its own location. These are the same directory in both repos.

## Database Migrations

None.

## Env Config

No new env vars. The same `test.env` (backend) and `.env` (prompt-service) files are still consumed.

## Relevant Docs

- [pytest-dotenv PyPI page](https://pypi.org/project/pytest-dotenv/) — confirms last release 2020
- [python-dotenv docs](https://saurabh-kumar.in/python-dotenv/) — what we now call directly

## Related Issues or PRs

- Follow-up to review feedback on #1981 — the reviewer's MEDIUM-severity comment recommended this replacement contingent on `--collect-only` breakage. Collection didn't break under pytest 9 in #1981, but the underlying maintenance concern is independent of that bump and worth landing on its own.

## Dependencies Versions

- **Removed:** `pytest-dotenv==0.5.2`
- **Used (existing):** `python-dotenv==1.0.1` (already in both lockfiles)

## Notes on Testing

- `uv run pytest --collect-only -q` from `prompt-service/`: **19 tests collected** (same as before the change).
- Standalone `python-dotenv` round-trip: confirmed `load_dotenv(Path('.env'))` correctly populates `PG_BE_HOST` / `PG_BE_PORT` from prompt-service's local `.env`.
- All pre-commit hooks (ruff, ruff-format, pycln, pyupgrade, secret-detection) pass on the new conftest.py files.
- Backend `uv sync` hit unrelated issues in local environment (`django-celery-beat==2.5.0` wheel-zip corruption, `psycopg2-binary` build) — both pre-existing and unrelated to this change; CI will exercise the full backend collection.

## Screenshots

N/A — config-only change.

## Checklist


prompt-service/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+plugin's default was to load ``.env`` from the rootdir; this preserves
+that exact behavior. Variables already set in the process environment
+are preserved (``override=False``).
+"""
+
+from pathlib import Path
+
+from dotenv import load_dotenv
+
+load_dotenv(Path(__file__).parent / ".env", override=False)
```
jaseemjaskp: **[HIGH] This conftest never runs in the prompt-service CI path.**

`tox.ini:62` invokes the prompt-service unit tests with `--noconftest`, which disables *all* `conftest.py` files — including this one. The old `pytest-dotenv` loaded via a `pytest11` entry-point **plugin**, which `--noconftest` does **not** disable. So in CI this `load_dotenv()` does not execute, whereas the plugin did — directly contradicting the PR's "preserves exact prior behavior" claim for prompt-service. (Confirmed empirically with a sentinel env var run under the exact CI command.)

Impact today is nil — the only unit test (`tests/unit/test_retriever_llm.py`) reads no env vars, and `config.py:18` calls `load_dotenv()` at import — so this is a latent trap, not a live bug. Note `tests/unit/conftest.py:6-11` already documents this `--noconftest` behavior.

**Fix:** to truly restore the plugin's behavior, load env somewhere `--noconftest` can't suppress — e.g. a `set_env`/`commands_pre` step in the prompt-service tox env, or a `-p` plugin module referenced in `addopts`. Otherwise, correct the docstring to state this file is skipped under tox and applies only to local/IDE runs.

prompt-service/conftest.py
```diff
@@ -0,0 +1,18 @@
+"""Top-level pytest conftest for the prompt-service.
+
+Loads environment variables from ``.env`` before pytest collects any
+test modules. This replaces the unmaintained ``pytest-dotenv`` plugin
+(last release: Feb 2020) which monkey-patched private pytest internals
+and was a recurring concern across pytest major upgrades.
+
+The prompt-service pyproject did not configure ``env_files``, so the
+plugin's default was to load ``.env`` from the rootdir; this preserves
+that exact behavior. Variables already set in the process environment
```
jaseemjaskp: **[MEDIUM] Filename and the "exact behavior" claim are both off.**

This loads `prompt-service/.env`, but the project's test-env convention is `.env.test`: the sibling `src/unstract/prompt_service/tests/conftest.py:10` loads `.env.test`, the integration tests load `.env.test`, and the committed sample is `sample.env.test` (there is no `sample.env` for tests). Separately, pytest's `rootdir` resolves to the **repo root** here (configfile = root `pyproject.toml`), so pytest-dotenv's "`.env` from rootdir" default would have pointed at the *repo-root* `.env`, not `prompt-service/.env` — a third location. So "preserves that exact behavior" isn't supported on either count.

**Fix:** confirm the intended file (most likely `.env.test`, to match the sibling conftest and sample) and update both the path and the docstring.

prompt-service/conftest.py
```diff
@@ -0,0 +1,18 @@
+"""Top-level pytest conftest for the prompt-service.
+
+Loads environment variables from ``.env`` before pytest collects any
+test modules. This replaces the unmaintained ``pytest-dotenv`` plugin
+(last release: Feb 2020) which monkey-patched private pytest internals
```
jaseemjaskp: **[LOW] Docstring rationale is likely inaccurate.**

"monkey-patched private pytest internals" — pytest-dotenv 0.5.2 actually integrates via the documented `pytest_load_initial_conftests` hook, not monkey-patching. A rationale comment is only useful if it's verifiable; consider softening to "unmaintained since 2020 across recent pytest majors." (Same wording in `backend/conftest.py:5`.)

backend/uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
```
jaseemjaskp: **[MEDIUM] Lockfile `revision` bumped 2→3 — inconsistent with the repo and CI's pinned uv.**

This bump is an artifact of regenerating with a newer local uv (0.9.5). CI pins uv `0.6.14` (`.github/workflows/ci-test.yaml:32`), which predates lock revision 3 and runs `uv sync` (`tox.ini:57`). Every other `uv.lock` in the repo is still at revision 1–2 — `backend/uv.lock` and `prompt-service/uv.lock` are the only two at 3. Risk: the pinned uv warns/errors or silently re-resolves on a revision it doesn't recognize.

**Fix:** regenerate this lock with the CI-pinned uv (or hand-revert the header to `revision = 2`) so the only diff is the `pytest-dotenv` removal. The package-removal hunks themselves are correct and minimal.

prompt-service/uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
```
jaseemjaskp: **[MEDIUM] Same `revision` 2→3 bump as `backend/uv.lock`.**

Regenerated under local uv 0.9.5; CI pins uv `0.6.14` (`.github/workflows/ci-test.yaml:32`) and runs `uv sync --group test` (`tox.ini:57`). Revert to `revision = 2` (or regenerate under the pinned uv) so the diff is limited to removing `pytest-dotenv`.

backend/conftest.py
```diff
@@ -0,0 +1,20 @@
...
+"""
+
+from pathlib import Path
+
+from dotenv import load_dotenv
+
+# Load test.env from this directory (the backend service root). Missing
+# file is silently tolerated, matching pytest-dotenv's behavior — tests
+# that need specific vars should assert on them in fixtures.
+load_dotenv(Path(__file__).parent / "test.env", override=False)
```
jaseemjaskp: **[LOW] `backend/conftest.py` is not exercised by any CI job.**

`tox.ini` has no backend testenv (`env_list` = py312, runner, sdk1, prompt-service; the bare `py312` `[testenv]` has no `commands`), and no workflow runs `pytest`/`manage.py test` against `backend/`. So this file only runs in local/dev pytest. Not a regression (backend tests weren't in this CI before), and the path resolution + `override=False` are correct for local runs — but the change is automation-unverified, so call that out in the PR description.

Optional: `addopts = "-s"` is set, so capturing the return and printing when the file is absent would make a missing/mis-located `test.env` debuggable instead of silent:
```python
if not load_dotenv(Path(__file__).parent / "test.env", override=False):
    print("[conftest] test.env not found; using ambient environment", flush=True)
```

backend/conftest.py
```diff
@@ -0,0 +1,20 @@
+"""Top-level pytest conftest for the backend service.
+
+Loads environment variables from ``test.env`` before pytest collects any
+test modules. This replaces the unmaintained ``pytest-dotenv`` plugin
+(last release: Feb 2020) which monkey-patched private pytest internals
```
jaseemjaskp: **[LOW] Docstring rationale + a minor redundancy.**

"monkey-patched private pytest internals" is likely inaccurate — pytest-dotenv uses the documented `pytest_load_initial_conftests` hook. Soften to a verifiable rationale (e.g. "unmaintained since 2020").

Minor (Code Simplifier): the inline comment at lines 17–19 largely restates the docstring's missing-file paragraph — consider dropping one to avoid duplicate maintenance.

jaseemjaskp (commented):
Automated multi-agent review (PR Review Toolkit: Comment Analyzer, PR Test Analyzer, Silent Failure Hunter, Type Design Analyzer, Code Reviewer, Code Simplifier).

**Intent is sound** — dropping an unmaintained, internals-touching plugin is the right call, `override=False` correctly mirrors pytest-dotenv's `env_override_existing_values=0` default, and `python-dotenv==1.0.1` is already in both lockfiles (no new dependency). Type Design Analyzer found nothing (no new types).

Two issues, however, undercut the PR's central "preserves exact prior behavior" claim: (1) the prompt-service conftest is skipped by `--noconftest` in CI, and (2) the `uv.lock` `revision` was bumped 2→3 by a newer local uv than CI's pinned 0.6.14. Details inline.

chandrasekharan-zipstack (approved):

muhammad-ali-e (approved):

PR #1996, opened 2026-05-28
UN-3493 [FEAT] Add AWS Bedrock Guardrails support to Bedrock LLM adapter

chandrasekharan-zipstack:
## What

- Surface AWS Bedrock Guardrails in the Bedrock LLM adapter so users can attach a guardrail to moderate prompts and completions.
- New optional schema fields: `guardrail_identifier`, `guardrail_version`, `guardrail_trace`.

## Why

- Customers asked for Bedrock Guardrails to gate LLM input/output for safety/compliance.
- LiteLLM SDK already supports Bedrock guardrails natively via a `guardrailConfig` kwarg — Unstract just wasn't exposing it.
- LiteLLM's [proxy-mode guardrails](https://docs.litellm.ai/docs/proxy/guardrails/bedrock) don't apply (Unstract uses LiteLLM as an SDK, not a proxy), so the right fit is the SDK-native path.

## How

- `unstract/sdk1/.../static/bedrock.json` — add optional `guardrail_identifier`, `guardrail_version`, `guardrail_trace` fields with user-facing descriptions.
- `unstract/sdk1/.../adapters/base1.py`:
  - Declare `guardrailConfig: dict | None = None` on `AWSBedrockLLMParameters` so it round-trips Pydantic re-validation in `LLM.complete()` (matches the existing `api_key` pattern). Annotated `# noqa: N815` since the field name must match LiteLLM's camelCase kwarg.
  - Add `_pack_bedrock_guardrail_config()` helper that maps the snake_case UI fields into LiteLLM's camelCase `guardrailConfig` dict.
  - In `validate()`: call the helper, strip the snake_case UI fields before Pydantic, drop `guardrailConfig` when unset so `None` never reaches LiteLLM/boto3.

- No. All three new schema fields are optional. When unset, `_pack_bedrock_guardrail_config` is a no-op and `guardrailConfig` is dropped from the kwargs, so the Bedrock call path is byte-identical to today. All 37 existing `test_bedrock_adapter.py` tests pass unchanged.

## Database Migrations

- None.

## Env Config

- None.

## Relevant Docs

- AWS Bedrock Guardrails: https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails.html
- LiteLLM Bedrock provider (SDK kwargs incl. `guardrailConfig`): https://docs.litellm.ai/docs/providers/bedrock

## Related Issues or PRs

- Jira: [UN-3493](https://zipstack.atlassian.net/browse/UN-3493)

## Dependencies Versions

- None.

## Notes on Testing

- Smoke test (manual, run from `unstract/sdk1`):
  - Guardrail set → `validated['guardrailConfig'] == {'guardrailIdentifier': 'ff6ujrregl1q', 'guardrailVersion': 'DRAFT', 'trace': 'enabled'}`.
  - Guardrail omitted → `'guardrailConfig'` key absent in `validated`.
  - Re-validation round-trip (`AWSBedrockLLMParameters.validate(validated)`) preserves `guardrailConfig` intact.
- `uv run pytest tests/test_bedrock_adapter.py -q` → 37 passed.
- End-to-end: create a Bedrock adapter from the UI with a guardrail ID + version, run a prompt, confirm Bedrock applies the guardrail (trace visible in response when `guardrail_trace: enabled`).

## Checklist

[UN-3493]: https://zipstack.atlassian.net/browse/UN-3493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
@@ -852,6 +852,21 @@ def _resolve_bedrock_aws_credentials(
     return validated
 
 
+def _pack_bedrock_guardrail_config(metadata: dict[str, "Any"]) -> None:
+    """Translate snake_case guardrail_* fields into LiteLLM's guardrailConfig."""
+    identifier = metadata.get("guardrail_identifier")
+    if not identifier:
+        return
+    cfg: dict[str, str] = {"guardrailIdentifier": identifier}
+    version = metadata.get("guardrail_version")
+    if version:
+        cfg["guardrailVersion"] = version
+    trace = metadata.get("guardrail_trace")
+    if trace:
+        cfg["trace"] = trace
+    metadata["guardrailConfig"] = cfg
+
+
 class AWSBedrockLLMParameters(BaseChatCompletionParameters):
     """See https://docs.litellm.ai/docs/providers/bedrock."""
 
@@ -866,6 +881,9 @@ class AWSBedrockLLMParameters(BaseChatCompletionParameters):
     aws_profile_name: str | None = None  # For AWS SSO authentication
     model_id: str | None = None  # For Application Inference Profile (cost tracking)
     max_retries: int | None = None
+    # Declared so it survives Pydantic re-validation of kwargs.
+    # Matches LiteLLM's Bedrock kwarg name, hence the mixed case.
+    guardrailConfig: dict | None = None  # noqa: N815
 
     @staticmethod
     def validate(adapter_metadata: dict[str, "Any"]) -> dict[str, "Any"]:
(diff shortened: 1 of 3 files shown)


praveen-formido:
On `guardrail_version` enforcement — the schema description says it's required when the identifier is set, but nothing actually enforces it. If a user fills only the identifier, `_pack_bedrock_guardrail_config()` builds `{"guardrailIdentifier": id}` without `guardrailVersion` and Bedrock rejects the call at runtime with a less-than-helpful error.

A few options, in order of my preference:

1. **Default to `"DRAFT"`** inside the helper when identifier is set but version is empty — matches the smoke-test convention in the PR description and the AWS console treatment of DRAFT as the implicit working version. Least friction for users.
2. Raise a clear `ValueError` in `_pack_bedrock_guardrail_config()` so the failure happens at config-validation time with an actionable message, rather than at the Bedrock API.
3. Add a JSON schema `allOf` / `if-then` to require it conditionally at the UI layer.

Option 1 keeps the UX cleanest; if you want to be stricter, layer option 2 on top.

pk-zipstack (approved):

PR #1950, opened 2026-05-07
UN-3452 [FEAT] Characterise the seams: dispatch + chord call sites

muhammad-ali-e:
## What

Two new test files characterising the refactor seams that upcoming spine PRs will touch. **Zero production code changes.**

| File | Lines | Tests | Site type |
|---|---|---|---|
| `workers/tests/test_dispatch_sites_characterisation.py` | 276 | 11 | Two `current_app.send_task` sites |
| `workers/tests/test_chord_sites_characterisation.py` | 316 | 9 | Two `chord(...)` sites + mixin wrapper |
| **Total** | **592** | **20** | |

Local run: `20 passed in ~2s`.

## Why

Lock down the **current behaviour** of the dispatch and chord seams before the upcoming refactors touch them. Per Michael Feathers' *Working Effectively with Legacy Code* approach: characterisation tests first, then refactor.

Two future spine PRs need this safety net:

- **`@shared_task` → `@worker_task` dispatch migration** — replaces both raw `current_app.send_task(...)` call sites with a unified `dispatch()` helper.
- **Chord → Barrier lift** — replaces both `chord(...)` invocations with a transport-agnostic `Barrier` abstraction matching the labs target architecture's `DECR remaining:{exec_id}` pattern.

Chord is the **highest-risk Celery construct** noted in the workers/ rearchitecture decision (silent task drops at ~130K-task scale). Characterising before refactor is critical.

Reference: UN-3452.

### Dispatch seam

Sites covered:
- `shared/patterns/notification/helper.py:76` — webhook notification dispatch
- `scheduler/tasks.py:157` — scheduled workflow async dispatch

For each site, tests pin:
- Exact task name
- Positional args layout (count + values)
- Kwargs layout (keys + values)
- Target queue
- Return-value semantics (success path + error path)

### Chord seam

Sites covered:
- `shared/workflow/execution/orchestration_utils.py:67` — `WorkflowOrchestrationUtils.create_chord_execution` (centralised helper) + its `WorkflowOrchestrationMixin.create_chord` wrapper
- `api-deployment/tasks.py:673` — inline chord inside `_run_workflow_api` (inventory-only — see note below)

Tests pin:
- Empty-batch short-circuit (existing defense against silent task drops)
- Callback signature construction (correct task name, kwargs, queue)
- chord(batch_tasks)(callback_signature) call sequence
- Return-value semantics
- Error propagation (re-raise after logging)
- Mixin's `self.app` extraction + `RuntimeError` when no app bound

[...]

PR #2001, opened 2026-06-01
UN-3497 [FEAT] Introduce queue_backend seam and migrate Celery call sites

muhammad-ali-e:
## What

- Add new module `workers/queue_backend/` (`decorator.py`, `dispatch.py`, `__init__.py`) — a single seam wrapping Celery task registration and dispatch.
- Migrate all 9 worker `tasks.py` files from `@shared_task` / `@app.task` to `@worker_task` (api-deployment, callback, executor, file_processing, general, ide_callback, log_consumer, notification, scheduler).
- Migrate the 2 raw `current_app.send_task(...)` call sites to `dispatch(...)`:
  - `shared/patterns/notification/helper.py:76`
  - `scheduler/tasks.py:157`
- New `tests/test_queue_backend_seam.py` (15 equivalence tests) + updated `tests/test_dispatch_sites_characterisation.py` (13 tests).
- Out of scope but bundled for convenience (see UN-3497 Jira comment for rationale): new `workers/ide_callback/worker.py` (parity with other worker dirs); `workers/run-worker.sh` DX refactor (`-L`/`-C`/`-r` flags, anchored pgrep, ide_callback wiring).

## Why

The PG Queue migration (UN-3445) is rolling out as a per-task, feature-flagged Strangler-Fig swap, not a big-bang. To make per-task routing possible without a codebase-wide rewrite later, every task registration and every dispatch needs to go through one chokepoint. This PR creates that chokepoint and migrates every existing site to it, while keeping Celery 100% in charge of runtime behaviour today.

## How

- `@worker_task` is a transparent passthrough to `@shared_task`. Both call forms (bare and parameterised) are forwarded as-is via `*args, **kwargs`.
- `dispatch(task_name, *, args, kwargs, queue)` is a transparent passthrough to `current_app.send_task(...)`. The signature exposes only what existing call sites actually use — more Celery options can be added when a real caller needs them.
- The future PG Queue branch (later Phase) lands inside `queue_backend/` only; call sites stay untouched.
- Worker `worker.py` files swap `from celery import shared_task` for `from queue_backend import worker_task` so the import is uniform.

No.

- Behaviour over Celery is unchanged. `@worker_task` resolves to the same `PromiseProxy` / decorator factory as `@shared_task`, and `dispatch()` produces an identical `current_app.send_task(...)` call.
- `tests/test_queue_backend_seam.py` proves the equivalence at the unit level (15 tests).
- `tests/test_dispatch_sites_characterisation.py` proves the two migrated call sites produce identical task name / args / kwargs / queue shape pre- vs post-migration (13 tests), and includes an inventory canary that fails if a raw `send_task` is ever added outside the seam.
- No queue names, task names, or routing change.
- No database, schema, or env-var changes.

## Database Migrations

- None.

## Env Config

- None in this PR. A later Phase (PG Queue Gate) will introduce `WORKER_PG_QUEUE_ENABLED_TASKS` / `WORKER_PG_QUEUE_ENABLED_ORGS` inside `queue_backend/`; default empty preserves Celery routing for every task.

## Relevant Docs

- N/A. Module-level docstrings inside `queue_backend/` document the current passthrough behaviour and the future routing branch.

## Related Issues or PRs

- Ticket: UN-3497 (sub-task) under UN-3496 (Story: PG Queue Phase 4 — Transport seam) under UN-3445 (Epic).
- Prior characterisation suite that this PR relies on: UN-3452 (#1950).

## Dependencies Versions

- None. Pure-Python passthrough over existing Celery primitives.

## Notes on Testing

\`\`\`
cd workers
.venv/bin/python -m pytest tests/test_queue_backend_seam.py \\
# 28 passed in ~2s
\`\`\`

Verification greps (should be empty / scoped):

\`\`\`
grep -rn "current_app.send_task" workers/   # only inside queue_backend/
grep -rEn "@shared_task|@app.task" workers/*/tasks.py   # empty
\`\`\`

## Screenshots

N/A (no UI surface).

## Checklist

### DIFF
--- a/workers/ide_callback/tasks.py
+++ b/workers/ide_callback/tasks.py
@@ -15,6 +15,7 @@ from datetime import date, datetime
 from typing import Any
 
 from celery import current_app as app
+from queue_backend import worker_task
 from shared.clients.prompt_studio_client import PromptStudioAPIClient
 
 logger = logging.getLogger(__name__)
@@ -140,7 +141,7 @@ def _track_subscription_usage(org_id: str, run_id: str) -> None:
 # ------------------------------------------------------------------
 
 
-@app.task(name="ide_index_complete")
+@worker_task(name="ide_index_complete")
 def ide_index_complete(
     result_dict: dict[str, Any],
     callback_kwargs: dict[str, Any] | None = None,
@@ -299,7 +300,7 @@ def ide_index_complete(
         raise
 
 
-@app.task(name="ide_index_error")
+@worker_task(name="ide_index_error")
 def ide_index_error(
     failed_task_id: str,
     callback_kwargs: dict[str, Any] | None = None,
@@ -345,7 +346,7 @@ def ide_index_error(
         logger.exception("ide_index_error callback failed")
 
 
-@app.task(name="ide_prompt_complete")
+@worker_task(name="ide_prompt_complete")
 def ide_prompt_complete(
     result_dict: dict[str, Any],
     callback_kwargs: dict[str, Any] | None = None,
@@ -480,7 +481,7 @@ def ide_prompt_complete(
         raise
 
 
-@app.task(name="ide_prompt_error")
+@worker_task(name="ide_prompt_error")
 def ide_prompt_error(
     failed_task_id: str,
     callback_kwargs: dict[str, Any] | None = None,
@@ -533,7 +534,7 @@ def _get_extraction_client():
     return ExtractionAPIClient()
 
 
-@app.task(name="extraction_complete")
+@worker_task(name="extraction_complete")
 def extraction_complete(
     result_dict: dict[str, Any],
     callback_kwargs: dict[str, Any] | None = None,
@@ -625,7 +626,7 @@ def extraction_complete(
             raise
 
 
-@app.task(name="extraction_error")
+@worker_task(name="extraction_error")
 def extraction_error(
     failed_task_id: str,
     callback_kwargs: dict[str, Any] | None = None,
--- a/workers/ide_callback/worker.py
+++ b/workers/ide_callback/worker.py
@@ -1,22 +1,21 @@
-"""General Worker
+"""IDE Callback Worker
 
-Celery worker for general tasks including webhooks, background
-processing, and workflow orchestration.
+Celery worker for Prompt Studio IDE post-execution callbacks.
+Serves the ``ide_callback`` queue defined in shared/infrastructure/config/registry.py.
 """
 
-from queue_backend import worker_task
 from shared.enums.worker_enums import WorkerType
 from shared.infrastructure.config.builder import WorkerBuilder
 from shared.infrastructure.config.registry import WorkerRegistry
 from shared.infrastructure.logging import WorkerLogger
 
-# Setup worker
-logger = WorkerLogger.setup(WorkerType.GENERAL)
-app, config = WorkerBuilder.build_celery_app(WorkerType.GENERAL)
+# Setup worker - this executes when module is imported by Celery
+logger = WorkerLogger.setup(WorkerType.IDE_CALLBACK)
+app, config = WorkerBuilder.build_celery_app(WorkerType.IDE_CALLBACK)
 
 
-def check_general_worker_health():
-    """Custom health check for general worker."""
+def check_ide_callback_health():
+    """Custom health check for IDE callback worker."""
     from shared.infrastructure.monitoring.health import HealthCheckResult, HealthStatus
 
     try:
@@ -27,26 +26,26 @@ def check_general_worker_health():
 
         if api_healthy:
             return HealthCheckResult(
-                name="general_worker_health",
+                name="ide_callback_health",
                 status=HealthStatus.HEALTHY,
-                message="General worker is healthy",
+                message="IDE callback worker is healthy",
                 details={
-                    "worker_type": "general",
+                    "worker_type": "ide_callback",
                     "api_client": "healthy",
-                    "queue": "celery",
+                    "queue": "ide_callback",
                 },
             )
         else:
             return HealthCheckResult(
-                name="general_worker_health",
+                name="ide_callback_health",
                 status=HealthStatus.DEGRADED,
-                message="General worker partially functional",
+                message="IDE callback worker partially functional",
                 details={"api_client": "unhealthy"},
             )
 
     except Exception as e:
         return HealthCheckResult(
-            name="general_worker_health",
+            name="ide_callback_health",
             status=HealthStatus.DEGRADED,
             message=f"Health check failed: {e}",
             details={"error": str(e)},
@@ -54,18 +53,6 @@ def check_general_worker_health():
 
 
 # Register health check
-
 WorkerRegistry.register_health_check(
-    WorkerType.GENERAL, "general_worker_health", check_general_worker_health
+    WorkerType.IDE_CALLBACK, "ide_callback_health", check_ide_callback_health
 )
-
-
-@worker_task(bind=True)
-def healthcheck(self):
-    """Health check task for monitoring systems."""
-    return {
-        "status": "healthy",
-        "worker_type": "general",
-        "task_id": self.request.id,
-        "worker_name": config.worker_name if config else "general-worker",
-    }
--- a/workers/queue_backend/__init__.py
+++ b/workers/queue_backend/__init__.py
@@ -8,9 +8,9 @@ Today both entry points are no-op aliases over Celery primitives:
 * ``dispatch(task_name, args, kwargs, queue)`` -> ``current_app.send_task(...)``
 * ``@worker_task`` -> ``@shared_task``
 
-In a later PR these will gain per-task routing: when a task name appears
-in ``WORKER_PG_QUEUE_ENABLED_TASKS``, dispatch routes through the PG Queue
-backend instead. Default (env var empty) keeps 100% of traffic on Celery.
+A later phase will route specific tasks through a non-Celery substrate
+(PG Queue) based on configuration; until then everything goes to Celery.
+The exact routing mechanism is intentionally not pinned here.
 
 Call sites should migrate to this module so the eventual substrate switch
 is a single-flag operation rather than a codebase-wide rewrite.
--- a/workers/queue_backend/__init__.py
+++ b/workers/queue_backend/__init__.py
@@ -1,0 +1,22 @@
+"""Queue-backend seam for workers.
+
+This module is the single place where the choice of queue substrate
+(Celery+RabbitMQ today; PG Queue in the future) lives.
+
+Today both entry points are no-op aliases over Celery primitives:
+
+* ``dispatch(task_name, args, kwargs, queue)`` -> ``current_app.send_task(...)``
+* ``@worker_task`` -> ``@shared_task``
+
+In a later PR these will gain per-task routing: when a task name appears
+in ``WORKER_PG_QUEUE_ENABLED_TASKS``, dispatch routes through the PG Queue
+backend instead. Default (env var empty) keeps 100% of traffic on Celery.
+
+Call sites should migrate to this module so the eventual substrate switch
+is a single-flag operation rather than a codebase-wide rewrite.
+"""
+
+from .decorator import worker_task
+from .dispatch import dispatch
+
+__all__ = ["dispatch", "worker_task"]
--- a/workers/queue_backend/decorator.py
+++ b/workers/queue_backend/decorator.py
@@ -26,8 +26,9 @@ def worker_task(*args: Any, **kwargs: Any) -> Any:
     """Register a function as a worker task via the queue_backend seam.
 
     Today this is a one-line passthrough to ``celery.shared_task``. The
-    indirection is the seam: when PR #15 adds PG Queue routing, the
-    consumer-registration logic lands here without touching call sites.
+    indirection is the seam: when a later phase adds PG Queue routing,
+    the consumer-registration logic lands here without touching call
+    sites.
 
     The return type is ``Any`` because ``shared_task`` returns different
     objects depending on call form — a ``PromiseProxy`` for the bare
--- a/workers/queue_backend/decorator.py
+++ b/workers/queue_backend/decorator.py
@@ -1,0 +1,38 @@
+"""Task registration decorator.
+
+Today: a transparent wrapper over ``celery.shared_task``.
+Future: registers the task body with whichever substrates are enabled
+(Celery + optionally PG Queue), so a single ``@worker_task`` definition
+can be served by either consumer.
+
+Accepts both Celery decorator forms — ``shared_task`` handles them
+internally, so a pass-through ``*args, **kwargs`` is enough:
+
+    @worker_task
+    def healthcheck(self): ...
+
+    @worker_task(bind=True, name="my.task")
+    def my_task(self, payload): ...
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from celery import shared_task
+
+
+def worker_task(*args: Any, **kwargs: Any) -> Any:
+    """Register a function as a worker task via the queue_backend seam.
+
+    Today this is a one-line passthrough to ``celery.shared_task``. The
+    indirection is the seam: when PR #15 adds PG Queue routing, the
+    consumer-registration logic lands here without touching call sites.
+
+    The return type is ``Any`` because ``shared_task`` returns different
+    objects depending on call form — a ``PromiseProxy`` for the bare
+    ``@worker_task`` form and a decorator factory for the parameterised
+    ``@worker_task(name=...)`` form. Pinning a tighter type would lock
+    out future routing variants without buying real safety today.
+    """
+    return shared_task(*args, **kwargs)
--- a/workers/queue_backend/dispatch.py
+++ b/workers/queue_backend/dispatch.py
@@ -1,46 +1,58 @@
 """Transport-agnostic task dispatch.
 
 Today: thin pass-through to ``celery.current_app.send_task``.
-Future: per-task routing between Celery and PG Queue based on
-``WORKER_PG_QUEUE_ENABLED_TASKS``.
+A later phase will introduce per-task routing through a non-Celery
+substrate (PG Queue); call sites stay untouched.
 
-The signature intentionally exposes only what the current two raw
-``current_app.send_task`` sites actually use (args, kwargs, queue).
-More Celery options can be added when a real call site needs them —
-not before.
+The signature intentionally exposes only what the current call sites
+actually use (args, kwargs, queue). More Celery options can be added
+when a real call site needs them — not before.
 """
 
 from __future__ import annotations
 
-from typing import Any
+from collections.abc import Mapping, Sequence
+from typing import Any, Protocol
 
 from celery import current_app
+
+
+class DispatchHandle(Protocol):
+    """The minimum contract every dispatch substrate must satisfy.
+
+    Today this is satisfied by Celery's ``AsyncResult`` (which exposes
+    ``.id``). A future PG Queue handle will need to expose the same
+    attribute so existing callers — e.g. ``scheduler/tasks.py`` — keep
+    working unchanged.
+    """
+
+    id: str
 
 
 def dispatch(
     task_name: str,
     *,
-    args: list[Any] | None = None,
-    kwargs: dict[str, Any] | None = None,
+    args: Sequence[Any] | None = None,
+    kwargs: Mapping[str, Any] | None = None,
     queue: str | None = None,
-) -> Any:
+) -> DispatchHandle:
     """Enqueue a task by name.
 
     Args:
         task_name: Registered task name (e.g. "send_webhook_notification").
-        args: Positional task args. Defaults to ``[]``.
-        kwargs: Keyword task args. Defaults to ``{}``.
+        args: Positional task args. Forwarded verbatim; Celery normalises
+            ``None`` internally.
+        kwargs: Keyword task args. Forwarded verbatim; Celery normalises
+            ``None`` internally.
         queue: Target queue name. Defaults to the task's bound queue.
 
     Returns:
-        Whatever the underlying transport returns (today: a Celery
-        ``AsyncResult``). Callers should not assume a particular type
-        beyond truthiness — this is the seam point that lets PG Queue
-        return a different handle later.
+        A handle to the enqueued task. ``.id`` is guaranteed; everything
+        else is substrate-specific and callers must not rely on it.
     """
     return current_app.send_task(
         task_name,
-        args=args if args is not None else [],
-        kwargs=kwargs if kwargs is not None else {},
+        args=args,
+        kwargs=kwargs,
         queue=queue,
     )
--- a/workers/queue_backend/dispatch.py
+++ b/workers/queue_backend/dispatch.py
@@ -1,0 +1,46 @@
+"""Transport-agnostic task dispatch.
+
+Today: thin pass-through to ``celery.current_app.send_task``.
+Future: per-task routing between Celery and PG Queue based on
+``WORKER_PG_QUEUE_ENABLED_TASKS``.
+
+The signature intentionally exposes only what the current two raw
+``current_app.send_task`` sites actually use (args, kwargs, queue).
+More Celery options can be added when a real call site needs them —
+not before.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from celery import current_app
+
+
+def dispatch(
+    task_name: str,
+    *,
+    args: list[Any] | None = None,
+    kwargs: dict[str, Any] | None = None,
+    queue: str | None = None,
+) -> Any:
+    """Enqueue a task by name.
+
+    Args:
+        task_name: Registered task name (e.g. "send_webhook_notification").
+        args: Positional task args. Defaults to ``[]``.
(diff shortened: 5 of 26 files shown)


workers/run-worker.sh
```diff
@@ -275,32 +293,142 @@ validate_env() {
 }
 
 # Function to get worker PIDs
+# Anchors the match on the hostname value's start (preceded by a non-word
+# character) so e.g. "callback" doesn't also match "ide_callback".
 get_worker_pids() {
     local worker_type=$1
-    pgrep -f "uv run celery.*worker.*$worker_type" || true
+    pgrep -f -- "[^[:alnum:]_]${worker_type}-worker-" 2>/dev/null || true
```
muhammad-ali-e: **Critical — pgrep regression.** This pattern requires a trailing `-` after `-worker`, but `run_worker()` (line 547+569) only appends `-${WORKER_INSTANCE_ID}` when that env var is set. By default the hostname is `${worker_type}-worker@%h` (no trailing dash), so the argv contains e.g. `--hostname=general-worker@host` — which never matches `[^[:alnum:]_]general-worker-`.

**Effect:** `-s/--status`, `-r <worker>/--restart`, and `kill_one_worker` all report **STOPPED / "not running"** for any locally-started worker that did not set `WORKER_INSTANCE_ID`. So `-r` silently degenerates into "start a duplicate" — two consumers on the same queue.

**Fix:** anchor with a delimiter that matches both hostname forms, e.g.
```bash
pgrep -f -- "[^[:alnum:]_]${worker_type}-worker(@|-)" 2>/dev/null || true
```
This still keeps `callback` from swallowing `ide_callback` (the rewrite's stated goal) while no longer requiring a trailing dash.

Also consider dropping `2>/dev/null` — it hides genuine pgrep errors (invalid regex, /proc unavailable in restricted containers).

workers/run-worker.sh
```diff
@@ -275,32 +293,142 @@ validate_env() {
...
+    echo "$pids" | xargs kill -TERM 2>/dev/null || true
+    sleep 1
+    echo "$pids" | xargs kill -KILL 2>/dev/null || true
+}
+
+# Function to kill workers
+kill_workers() {
+    print_status $YELLOW "Killing all running workers..."
+    local pids
+    pids=$(pgrep -f "uv run celery.*worker" || true)
```
muhammad-ali-e: **Important — `kill_workers` reverted to the loose pattern.** The bulk-kill path uses `pgrep -f "uv run celery.*worker"` while the new `kill_one_worker` correctly anchors per-worker. Two consequences:

1. An unrelated `uv run celery -A something_else worker` running in the same host gets killed silently.
2. If the pattern misses (different invocation format), the script declares "All workers stopped" anyway at line 432 — a false positive.

**Fix:** iterate `list_core_worker_dirs` + `list_pluggable_worker_dirs` here too and reuse `kill_one_worker` so the anchored matcher is the single source of truth.

workers/run-worker.sh
```diff
@@ -275,32 +293,142 @@ validate_env() {
...
+kill_one_worker() {
+    local worker_dir=$1
+    local pids
+    pids=$(get_worker_pids "$worker_dir" | tr '\n' ' ' | sed 's/ $//')
+    if [[ -z "$pids" ]]; then
+        print_status $YELLOW "  $worker_dir: not running"
+        return
+    fi
+    print_status $YELLOW "  $worker_dir: killing $pids"
+    echo "$pids" | xargs kill -TERM 2>/dev/null || true
```
muhammad-ali-e: **Important — silent EPERM / failed-kill.** `kill -TERM ... 2>/dev/null || true` (also lines 416, 429, 431) swallows every error from kill(2): permission denied, no such process, signal-not-permitted, etc. After SIGKILL we never re-check whether the process actually died — `print_status $GREEN "All workers stopped"` fires unconditionally.

**Failure mode this hides:** operator runs as the wrong UID, sees "All workers stopped", starts new workers, ends up with 2× consumers per queue.

**Fix:** drop the `2>/dev/null || true` (let kill stderr surface), and after the `sleep`, re-run `get_worker_pids` to confirm. Warn if any survive.

workers/ide_callback/worker.py
```diff
@@ -0,0 +1,58 @@
...
+            status=HealthStatus.DEGRADED,
+            message=f"Health check failed: {e}",
+            details={"error": str(e)},
+        )
+
+
+# Register health check
+WorkerRegistry.register_health_check(
+    WorkerType.IDE_CALLBACK, "ide_callback_health", check_ide_callback_health
+)
```
muhammad-ali-e, line 58: **Important — missing `healthcheck` Celery task.** The 8 other migrated worker.py files (`api-deployment`, `callback`, `executor`, `file_processing`, `general`, `log_consumer`, `notification`, `scheduler`) all define a `@worker_task(bind=True) def healthcheck(self): ...` at the end. This one only registers `check_ide_callback_health` (the HTTP/HealthCheckResult-style probe), with **no Celery task** named `healthcheck` on the `ide_callback` queue.

**Effect:** anything that does `app.send_task("healthcheck", queue="ide_callback")` will silently fail to route. Even if nothing in-tree does that today, the inconsistency violates the commit message's "parity with other worker dirs" claim.

**Fix:** add a `@worker_task(bind=True) def healthcheck(self): return {...}` block mirroring callback/worker.py:65-66.

workers/queue_backend/dispatch.py
```diff
@@ -0,0 +1,46 @@
...
+
+    Returns:
+        Whatever the underlying transport returns (today: a Celery
+        ``AsyncResult``). Callers should not assume a particular type
+        beyond truthiness — this is the seam point that lets PG Queue
+        return a different handle later.
+    """
+    return current_app.send_task(
+        task_name,
+        args=args if args is not None else [],
```
muhammad-ali-e: **Suggestion — redundant coercion.** `current_app.send_task` already normalises `args=None` / `kwargs=None` internally (Celery `Celery.send_task` does `args = args or []; kwargs = kwargs or {}`). Forwarding `args=args, kwargs=kwargs, queue=queue` directly is a strictly thinner passthrough, which is the exact design intent of this seam ("keep surface area minimal").

Related: the `if args is not None` form sends an explicit `[]` (list) where Celery's own default is `()` (tuple). Any third-party router that does `isinstance(args, tuple)` would see a behaviour change — rare but worth eliminating in a passthrough.

workers/queue_backend/dispatch.py
```diff
@@ -0,0 +1,46 @@
...
+from celery import current_app
+
+
+def dispatch(
+    task_name: str,
+    *,
+    args: list[Any] | None = None,
+    kwargs: dict[str, Any] | None = None,
+    queue: str | None = None,
+) -> Any:
```
muhammad-ali-e: **Suggestion — return type `Any` is too loose given the existing caller.** The docstring says "callers should not assume a particular type beyond truthiness" but `workers/scheduler/tasks.py:170` already reaches for `async_result.id`. So `.id: str` is already part of the implicit contract on day one.

A 4-line `Protocol` documents the real invariant and survives the PG Queue swap:

```python
from typing import Protocol

class DispatchHandle(Protocol):
    id: str  # task / job identifier, stable across substrates

def dispatch(...) -> DispatchHandle: ...
```

Celery's `AsyncResult.id` satisfies this structurally — no runtime change. Any future PG Queue handle must also expose `.id: str`, which is exactly the invariant the team wants. Strictly stronger than `Any`, zero cost.

While here: `args: list[Any]` → `Sequence[Any]` (Celery accepts tuples; callers should be free to pass them). Same for `kwargs: dict[str, Any]` → `Mapping[str, Any]`. Keep `None` defaults — they're the "caller said nothing" sentinel and useful for future routing distinctions.

workers/queue_backend/decorator.py
```diff
@@ -0,0 +1,38 @@
...
+
+from celery import shared_task
+
+
+def worker_task(*args: Any, **kwargs: Any) -> Any:
+    """Register a function as a worker task via the queue_backend seam.
+
+    Today this is a one-line passthrough to ``celery.shared_task``. The
+    indirection is the seam: when PR #15 adds PG Queue routing, the
+    consumer-registration logic lands here without touching call sites.
```
muhammad-ali-e: **Comment rot — "PR #15".** This identifier has no meaning in the repo's PR numbering (this PR is #2001; the eventual PG Queue routing PR will be some unrelated number). Once it lands, a future maintainer searching for `#15` finds an unrelated PR or nothing.

**Fix:** drop the number. "when a later PR adds PG Queue routing, the consumer-registration logic lands here without touching call sites." If you want traceability, use the Jira ticket (UN-XXXX) for the follow-up phase — or no identifier at all and trust git blame.

Same fix applies to `workers/tests/test_queue_backend_seam.py:156` (`PR #15` in `TestWorkerTaskEquivalence` class docstring).

muhammad-ali-e (commented):
## Comprehensive PR Review — UN-3497 queue_backend seam

Review run via PR Review Toolkit: **Code Reviewer**, **Silent Failure Hunter**, **Type Design Analyzer**, **PR Test Analyzer**, **Comment Analyzer**, **Code Simplifier** — scoped to commit `13be391a` (not the merge-from-main noise in the full `main...HEAD` diff). Also a regression sweep on the @shared_task → @worker_task swap and the run-worker.sh DX refactor.

### Verdict
The **Python seam itself is clean** — `worker_task` and `dispatch` are honest passthroughs, exception propagation is preserved, and the 28-test lockdown is the right shape for proving equivalence when the PG Queue routing PR lands. The two genuine risks live outside the seam: (a) **`run-worker.sh:300` anchored pgrep regresses default local runs** (status / restart / kill_one silently report STOPPED for workers that are running), and (b) the **`@worker_task` equivalence tests are too weak** — they would pass if the decorator dropped `**kwargs` entirely, silently stripping every `autoretry_for` / `max_retries` policy on every migrated task.

### Priority
- **Critical (fix before merge)**: 3 — pgrep regression, weak @worker_task equivalence, missing decorator-kwarg passthrough test.
- **Important**: 5 — `kill_workers` broad pgrep, silent kill EPERM, missing `ide_callback` healthcheck task, inventory canary blind spots, characterisation tests becoming tautological after seam patch.
- **Suggestion**: 5 — `dispatch()` redundant `is not None` coercion, `dispatch -> Protocol[DispatchHandle]` instead of `Any`, three comment-rot fixes (stale line ref, "PR #15" placeholder, `WORKER_PG_QUEUE_ENABLED_TASKS` name pin).

Individual findings as inline comments below. Each is independent — feel free to defer Suggestions and ship after Critical+Important are addressed.

workers/queue_backend/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+(Celery+RabbitMQ today; PG Queue in the future) lives.
+
+Today both entry points are no-op aliases over Celery primitives:
+
+* ``dispatch(task_name, args, kwargs, queue)`` -> ``current_app.send_task(...)``
+* ``@worker_task`` -> ``@shared_task``
+
+In a later PR these will gain per-task routing: when a task name appears
+in ``WORKER_PG_QUEUE_ENABLED_TASKS``, dispatch routes through the PG Queue
+backend instead. Default (env var empty) keeps 100% of traffic on Celery.
```
muhammad-ali-e: **Comment rot (med) — env var name pinning.** This docstring (and `dispatch.py:5`) pins both the *mechanism* (per-task allowlist) and the *exact env var name* `WORKER_PG_QUEUE_ENABLED_TASKS`. If the follow-up phase picks a different mechanism (feature flag service, config file, % rollout) or different name, the seam's own docs are immediately misleading.

**Fix:** describe behaviour, not identifier. E.g. "A later phase will route specific tasks to PG Queue based on configuration; the default keeps 100% of traffic on Celery." Drop the env-var name until the variable actually exists.

workers/tests/test_queue_backend_seam.py
```diff
@@ -0,0 +1,229 @@
...
+
+Today the seam is a transparent pass-through to Celery. These tests
+lock that in so future PRs adding per-task routing (PG Queue) can be
+proved to preserve the Celery default path.
+
+Two layers:
+
+1. **dispatch()** must produce the same ``current_app.send_task`` call
+   as the raw idiom used at the two existing dispatch sites
+   (``notification/helper.py:76``, ``scheduler/tasks.py:157``).
```
muhammad-ali-e: **Comment rot (high) — stale line reference.** The module docstring says `scheduler/tasks.py:157`, but `dispatch(` is now at line **154** in that file (line 157 is now `context.organization_id,  # schema_name`). This same commit moved the line — the docstring wasn't updated.

**Fix:** prefer the symbol-based style already used in `test_dispatch_sites_characterisation.py`:
```
scheduler/tasks.py::_execute_scheduled_workflow
shared/patterns/notification/helper.py::send_notification_to_worker
```
Line numbers always drift; symbol names don't.

workers/tests/test_queue_backend_seam.py
```diff
@@ -0,0 +1,229 @@
...
+        def via_seam():
+            return "ok"
+
+        @shared_task(name="queue_backend_test.passthrough_native")
+        def via_native():
+            return "ok"
+
+        # Both should produce something Celery considers a registered task.
+        for t in (via_seam, via_native):
+            assert hasattr(t, "apply") or hasattr(t, "delay")
```
muhammad-ali-e: **Critical — tautological equivalence assertion.** `hasattr(t, "apply") or hasattr(t, "delay")` is true for almost any Celery object, *including* `MagicMock()`. These tests would still pass if `worker_task` were replaced with:
```python
def worker_task(*a, **kw): return MagicMock()
```
That is the exact failure mode the seam should prevent — PR-next will swap the implementation for a routing dispatcher, and these tests will not catch a botched swap that breaks Celery registration.

**Stronger assertion (still broker-free):**
```python
from celery import current_app
@worker_task(name="qb.test")
def via_seam(): return "ok"
via_seam.name   # forces PromiseProxy resolution
assert via_seam.name == "qb.test"
assert "qb.test" in current_app.tasks
```

Applies to lines 178, 191, 201.

workers/tests/test_queue_backend_seam.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def test_parameterised_decorator_form(self):
+        """@worker_task(name=..., queue=...) accepts kwargs like @shared_task."""
+        from queue_backend import worker_task
+
+        @worker_task(name="queue_backend_test.parameterised", queue="general")
+        def some_function():
+            return "ok"
+
+        assert hasattr(some_function, "apply") or hasattr(some_function, "delay")
```
muhammad-ali-e: **Critical — decorator-kwarg passthrough is never asserted.** `worker_task(*args, **kwargs)` forwards everything, but no test pins it. The migrated call sites already use `bind=True`, `name=...`, `autoretry_for=(Exception,)`, `max_retries=...`. A regression like `return shared_task(*args)` (dropping `**kwargs`) would silently **strip every retry policy on every task** in the codebase and this suite would still go green.

**Add:**
```python
def test_worker_task_forwards_decorator_kwargs(self):
    @worker_task(name="qb.kw", bind=True, autoretry_for=(ValueError,), max_retries=7)
    def t(self): return "ok"
    t.name  # resolve
    assert t.name == "qb.kw"
    assert ValueError in (t.autoretry_for or ())
    assert t.max_retries == 7
```
This is the highest-leverage missing test — silent retry-policy stripping is a data-integrity hazard.

workers/tests/test_dispatch_sites_characterisation.py
```diff
@@ -224,57 +215,66 @@ def test_no_dispatch_when_execution_creation_fails(self):
...
         import re
 
         workers_root = pathlib.Path(__file__).parent.parent
         # Anchor skip to the top-level directory relative to workers_root so
         # we don't accidentally exclude legitimately-named subdirectories
         # (e.g. workers/shared/tests_helpers/).
-        skip_top_dirs = {"tests", "__pycache__", "htmlcov", ".venv"}
+        skip_top_dirs = {"tests", "__pycache__", "htmlcov", ".venv", "queue_backend"}
 
         pattern = re.compile(r"current_app\.send_task\b")
```
muhammad-ali-e: **Important — canary regex has blind spots.** `re.compile(r"current_app\.send_task\b")` won't catch:

1. **Aliased import:** `from celery import current_app as app; app.send_task(...)` — relevant: `workers/ide_callback/tasks.py:17` already does `from celery import current_app as app`, so a future call `app.send_task(...)` there would slip past.
2. **Locally constructed app:** `Celery(name).send_task(...)` — `workers/shared/infrastructure/config/builder.py` and `workers/shared/worker_builder.py` both construct `Celery(...)` instances; calls on those bypass the canary.
3. **`apply_async`** — `signature(...).apply_async()` is the other ~50% of how Celery dispatches and isn't watched.
4. **Multi-line dotted access** after autoformatter wrap.

**Replacement** (AST-based, catches all four):
```python
import ast
class V(ast.NodeVisitor):
    def visit_Call(self, node):
        if isinstance(node.func, ast.Attribute) and node.func.attr in {"send_task", "apply_async"}:
            self.hits.append(...)
```
Keep the regex as a fast pre-filter if you want — but the AST walker is what enforces the actual invariant.

workers/tests/test_dispatch_sites_characterisation.py
```diff
@@ -120,10 +113,8 @@ def test_dispatch_returns_true_on_success(self):
     def test_dispatch_returns_false_on_send_task_failure(self):
```
muhammad-ali-e: **Important — `test_dispatch_returns_false_on_send_task_failure` no longer tests what its name says.** Patching `shared.patterns.notification.helper.dispatch` with `side_effect=RuntimeError` only proves the try/except catches a `RuntimeError` from a *Python function call* — not from the *broker / transport*. The test is now self-referential: it injects the exact exception type the helper's `except Exception` is trivially designed to catch.

If `dispatch()` later wraps `send_task` errors in a `DispatchError`, this test still passes (because we control the `side_effect`) while the production path silently breaks.

**Suggested addition:** keep this test but ALSO add one that leaves `dispatch` unpatched and patches at the substrate boundary:
```python
with patch("queue_backend.dispatch.current_app") as mock_app:
    mock_app.send_task.side_effect = ConnectionError("broker down")
    result = send_notification_to_worker(...)
assert result is False
```
This exercises the genuine seam → broker → helper failure path.

muhammad-ali-e:
## Review feedback addressed — c84a3e8c (now `658ea89e` post-rebase onto main)

Below is a one-to-one accounting against each inline finding. 12 of 13 implemented; 1 pushed back on with rationale.

### Critical (3/3 fixed)

| # | File | Status | Notes |
|---|------|--------|-------|
| 1 | `run-worker.sh:300` pgrep regression | ✅ Fixed | Pattern now `[^[:alnum:]_]${worker_type}-worker(@|-)`. Matches both default hostname (`general-worker@host`) and `WORKER_INSTANCE_ID` variant (`general-worker-${id}@host`). |
| 2 | `test_queue_backend_seam.py:178` tautological `hasattr` | ✅ Fixed | Replaced with PromiseProxy resolution (`.name`), `current_app.tasks` membership check, and round-trip via `.apply().get()`. MagicMock will no longer satisfy these. |
| 3 | `test_queue_backend_seam.py:201` decorator kwargs never asserted | ✅ Fixed | New `test_forwards_decorator_kwargs` pins `name`, `bind`, `autoretry_for=(ValueError,)`, `max_retries=7`, `default_retry_delay=42`. A regression like `return shared_task(*args)` would now fail loudly. |

### Important (4/5 fixed, 1 pushed back)

| # | File | Status | Notes |
|---|------|--------|-------|
| 4 | `run-worker.sh:423` `kill_workers` loose pgrep | ✅ Fixed | Now iterates `list_core_worker_dirs` + `list_pluggable_worker_dirs` and delegates to `kill_one_worker`. Anchored matcher is the single source of truth. |
| 5 | `run-worker.sh:414` silent EPERM | ✅ Fixed | Dropped `2>/dev/null ‖ true`; re-`get_worker_pids` after `sleep`; SIGKILL survivors; warn on persistent processes; bubble exit code. |
| 6 | `ide_callback/worker.py` missing `healthcheck` Celery task | ❌ **Pushed back** — see below. |
| 7 | Inventory canary blind spots | ✅ Fixed | Replaced regex with AST walker (`DispatchCallFinder`). Catches `current_app.send_task`, aliased imports, `Celery(...).send_task`, AND `.apply_async`. Pre-existing `plugins/manual_review/.apply_async` baggage is on a documented exception list — that's an unrelated migration, will file a follow-up Jira task. |
| 8 | `test_dispatch_returns_false_on_send_task_failure` tautology | ✅ Fixed | Renamed to `test_helper_swallows_dispatch_exception` (it's still a valid Python error-handling assertion). Added complement `test_substrate_failure_surfaces_as_false` that leaves `dispatch` unpatched and patches `queue_backend.dispatch.current_app.send_task` with `ConnectionError`. The full helper → seam → broker failure path is now exercised without either side being trivially mocked. |

### Suggestions (5/5 fixed)

| # | File | Status | Notes |
|---|------|--------|-------|
| 9 | `dispatch.py:43` redundant coercion | ✅ Fixed | Removed `if args is not None else []` (and the kwargs counterpart). Celery normalises `None` internally. Avoids the subtle list-vs-tuple default behaviour change. Seam test renamed to `test_omitted_args_forwarded_as_none` to pin the new behaviour. |
| 10 | `dispatch.py:26` return type `Any` | ✅ Fixed | Added `DispatchHandle` Protocol exposing `.id: str`. Documents the real invariant (already in use at `scheduler/tasks.py`). Also relaxed `list[Any]` → `Sequence[Any]`, `dict[str, Any]` → `Mapping[str, Any]`. |
| 11 | `decorator.py:30` "PR #15" placeholder | ✅ Fixed | Removed the identifier in both decorator.py docstring and the `TestWorkerTaskEquivalence` class docstring. Now just "a later phase". |
| 12 | `__init__.py:13` + `dispatch.py:5` env var name pinning | ✅ Fixed | Replaced with behaviour-only description: "A later phase will route specific tasks through a non-Celery substrate (PG Queue) based on configuration; until then everything goes to Celery." |
| 13 | `test_queue_backend_seam.py:11` stale line ref | ✅ Fixed | Replaced `scheduler/tasks.py:157` with symbol-based citation (`_execute_scheduled_workflow` / `send_notification_to_worker`). Same idiom as `test_dispatch_sites_characterisation.py`. |

### SonarCloud (1/1 fixed)

| # | File | Status | Notes |
|---|------|--------|-------|
| 14 | `run-worker.sh:609` S1192 literal duplication | ✅ Fixed | Defined `readonly IDE_CALLBACK_WORKER_TYPE="ide_callback"` and replaced all 8 literal usages, mirroring the existing `EXECUTOR_WORKER_TYPE` pattern. Only remaining occurrences are the constant definition itself and one comment. |

---

### Push-back on #6 (missing `@worker_task healthcheck` in `ide_callback/worker.py`)

This was a deliberate omission, not an oversight. Earlier in the session we walked through whether the `healthcheck` Celery task in the other 8 worker.py modules is actually load-bearing. Findings:

- **No producer enqueues it.** Grep across `workers/` and `backend/` for `send_task("…healthcheck…")`, `dispatch("…healthcheck…")`, `apply_async`, and any `.healthcheck` invocation returns nothing. The only consumer is `tests/test_callback_sanity.py`, which is scoped to the **callback** worker's app (not `ide_callback`).
- **K8s liveness path uses the Python function, not the Celery task.** `WorkerBuilder.setup_health_monitoring` reads `WorkerRegistry.get_health_checks(worker_type)` and binds the registered `check_*_health` callables to `HealthServer` on port 8089. The `@worker_task healthcheck` Celery task is on a different path entirely.
- **`run-worker.sh -s`** does `pgrep` on the worker process — never invokes the task.

The reviewer's "Effect: anything that does `app.send_task('healthcheck', queue='ide_callback')` will silently fail to route" is a hypothetical — there's no such caller in-tree, and producing one would require the seam-bypass that the canary now actively forbids.

I called this out explicitly in the original PR-scope Jira comment on UN-3497: "the Celery `@worker_task healthcheck` boilerplate that exists in the other 8 worker.py files was **deliberately dropped** here". The reviewer-flagged "inconsistency" is real, but the consistency is with vestigial code in the other 8 modules — the cleanup direction is to remove them, not to add a 9th. I'll file a separate ticket for that cleanup if there's appetite; this PR isn't the right home for it.

---

### Test count

- `test_queue_backend_seam.py`: 15 → 17 tests
- `test_dispatch_sites_characterisation.py`: 13 → 14 tests
- **Total: 28 → 31 — all green** in 1.82s

workers/ide_callback/tasks.py
```diff
@@ -15,6 +15,7 @@
 from typing import Any
 
 from celery import current_app as app
+from queue_backend import worker_task
```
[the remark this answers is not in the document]
muhammad-ali-e: False positive — `app` is still referenced at line 102:

```python
res = AsyncResult(failed_task_id, app=app)
```

inside `_get_task_error`. The migration replaced every `@app.task(...)` decorator with `@worker_task(...)`, but kept the import because `AsyncResult` needs the Celery app instance to look up the result backend. Removing the import would break `_get_task_error` (called from `ide_index_error` and `ide_prompt_error`) at runtime.

Leaving as-is. `grep -n '\\bapp\\b' workers/ide_callback/tasks.py` shows both the import (line 17) and the consumer (line 102).

vishnuszipstack (approved):

chandrasekharan-zipstack (approved):

PR #2003, opened 2026-06-01
UN-3501 [FEAT] Plumb fairness key on every dispatch() call site

muhammad-ali-e:
## What

- Add `workers/queue_backend/fairness.py` — `FairnessKey` frozen dataclass (`org_id`, `pipeline_priority`, `tier`) + `system()` / `for_org(...)` constructors + `FAIRNESS_KWARG_NAME` constant. `to_dict()` is JSON-safe.
- `queue_backend.dispatch()` accepts optional `fairness=` kwarg; when provided, merges into a **copy** of the outgoing kwargs under `_fairness_key` (no caller-side mutation).
- Two existing production `dispatch(...)` sites now declare a fairness key:
  - `shared/patterns/notification/helper.py::send_notification_to_worker` — derives `org_id` from `payload.organization_id`.
  - `scheduler/tasks.py::_execute_scheduled_workflow` — derives `org_id` from `context.organization_id`.
- New `tests/test_fairness_key.py` (14 tests): value-object semantics, dispatch() integration, AST-based inventory canary (every production `dispatch(...)` must pass `fairness=`), and an additive-only canary (no worker code reads `_fairness_key` yet).

## Why

PG Queue Phase 5 (epic UN-3445) makes every queue-crossing payload **machine-readable** so a later phase has structured data to dispatch on. Phase 8's PG Queue fairness scheduler will route by `org_id` (per-tenant partition), `pipeline_priority` (within-tenant ordering), and `tier` (cross-tier preemption). Producer-side plumbing now means Phase 8 can introduce the reader without backfilling in-flight payloads.

## How

- The fairness key sits **inertly** in `kwargs["_fairness_key"]` until Phase 8 introduces the reader.
- The underscore prefix marks it as routing metadata, visually distinct from business kwargs in Celery introspection (Flower, `inspect.active()`).
- `FairnessKey.system()` is the sentinel for tasks with no tenant context (periodic log flush, healthchecks). PG Queue will treat those as a distinct "system" partition rather than as belonging to any tenant.
- The inventory canary is intentionally restricted to **bare-name** `dispatch(...)` callees so `ExecutionDispatcher.dispatch(...)` in `file_processing/structure_tool_task.py` — a totally different concept (executor-side RPC, not a queue boundary) — is not falsely audited.

No.

- Additive only — kwargs gain one extra entry under an underscored slot. No business kwargs change shape.
- No worker code reads `_fairness_key` (proved by `test_no_consumer_reads_fairness_kwarg`). Phase 8 lifts this canary.
- Producers pass it via `dispatch()`; the function's previous signature is preserved (`fairness` is keyword-only and defaults to `None`).
- Default-on is identical to default-off — no flag needed, no rollout coordination, no in-flight payload concern.
- Behaviour over Celery is unchanged. JSON-safety of the routing dict is asserted in `test_to_dict_is_json_safe`.

## Database Migrations

- None.

## Env Config

- None. Defaults (`pipeline_priority=50`, `tier="standard"`) live in `fairness.py`.

## Relevant Docs

- N/A. Module-level docstrings in `queue_backend/fairness.py` document the rollout invariant and the Phase 8 hand-off.

## Related Issues or PRs

- Ticket: UN-3501 (sub-task) under UN-3500 (Story: PG Queue Phase 5 — Payload typing) under UN-3445 (Epic).
- Builds on top of: UN-3497 (#2001) — Phase 4 transport seam.

## Dependencies Versions

- None. Pure stdlib (`dataclasses`, `ast`, `json`).

## Notes on Testing

```
cd workers
.venv/bin/python -m pytest \
    tests/test_fairness_key.py \
    tests/test_queue_backend_seam.py \
# 45 passed in ~4s
```

Verification greps:

```
# Producer side: every production dispatch passes fairness=
grep -rEn "dispatch\(" workers/scheduler workers/shared/patterns | grep -v "fairness="
# (should be empty)

# Consumer side: no worker reads the underscored slot yet
grep -rEn "_fairness_key|FAIRNESS_KWARG_NAME" workers --include="*.py" \
    | grep -v "queue_backend/\|tests/"
# (should be empty)
```

## Screenshots

N/A (no UI surface).

## Checklist


workers/queue_backend/fairness.py
```diff
@@ -0,0 +1,78 @@
...
+
+Until then the field travels in the Celery message header
+``x-fairness-key`` — out-of-band of the task body's kwargs, so a task
+whose signature does not accept ``**kwargs`` doesn't blow up on the
+extra field. On the consumer side it's reachable via
+``self.request.headers["x-fairness-key"]`` when needed.
+
+This module is additive-only:
+
+* No worker code reads ``_fairness_key`` today (verified by
```
muhammad-ali-e: **Comment Analyzer — stale doc.** This bullet still says "No worker code reads `_fairness_key` today (verified by `test_fairness_key.py`)." After commit 65811847 the key lives in the Celery header `x-fairness-key` and the test now greps for `x-fairness-key` / `FAIRNESS_HEADER_NAME` — `_fairness_key` is dead vocabulary. Suggest:

```
* No worker code reads ``x-fairness-key`` today (verified by
  ``test_fairness_key.py::TestNoConsumerYet``).
```

Same on line 57 (`to_dict()` docstring). Stale docs in a seam module are especially costly because Phase 8 readers will land here first.

workers/queue_backend/fairness.py
```diff
@@ -0,0 +1,78 @@
...
+    flushing, healthchecks). PG Queue's scheduler treats those as a
+    distinct "system" partition rather than as belonging to any tenant.
+    """
+
+    org_id: str | None
+    pipeline_priority: int = DEFAULT_PRIORITY
+    tier: str = DEFAULT_TIER
+
+    def to_dict(self) -> dict[str, str | int | None]:
+        """JSON-safe representation suitable for ``kwargs["_fairness_key"]``."""
```
muhammad-ali-e: **Comment Analyzer — stale doc.** `"JSON-safe representation suitable for kwargs[\"_fairness_key\"]"` describes the *previous* design. The dict is now serialised into the Celery message header `x-fairness-key`, not into kwargs. Suggest:

```python
"""JSON-safe representation carried in the Celery message header ``x-fairness-key``."""
```

muhammad-ali-e (commented):
## PR Review Toolkit — UN-3501 Fairness Key

Focused review covering code quality, silent failures, type design, test coverage, comments, simplification, and regression risk. PR scope is small and surgical (2 production call sites + 1 new value object + tests).

**Overall:** the headers-instead-of-kwargs rework (commit 65811847) is the right fix — moves routing metadata off the function-signature surface so non-`**kwargs` tasks don't blow up. Public API stays back-compat. Inventory canary is a nice safeguard.

**Main concerns (prioritised):**

1. **Stale docs** — `to_dict()` docstring and the module-level invariant still reference `kwargs["_fairness_key"]`. The implementation moved to a header (`x-fairness-key`); the docs didn't. Actively misleading for the next reader. *(HIGH)*
2. **`for_org(**overrides)` silently swallows typos** — `priority=80` or `tiers="enterprise"` get dropped on the floor and defaults apply. Plus `overrides: object` + `int()`/`str()` coercions paper over real bugs. Use explicit keyword-only params. *(HIGH)*
3. **Weak invariants on `tier` / `pipeline_priority`** — both are unbounded `str`/`int`. PG Queue Phase 8 will be the one to discover `tier="enterprize"` or `pipeline_priority=10_000_000`. Tighten now while there are only two callers. *(MEDIUM)*
4. **Canary misses aliased imports** — `from queue_backend import dispatch as foo` slips past `ast.Name and callee.id == "dispatch"`. Either lock the import form or expand the matcher. *(MEDIUM)*
5. **`FairnessKey.system()` adds no invariant beyond defaults** — semantically identical to `FairnessKey(org_id=None)`. Either drop the helper or wire it to a distinct `tier="system"` so the "system partition" claim is encoded in data, not naming. *(LOW)*

**Regression check:** keyword-only `fairness=None`, `headers=None` accepted by Celery 5.5+, two producer sites covered, no `current_app.send_task(...)` bypasses the seam, `to_dict()` JSON-safety asserted. No regressions found other than item 4's canary blind spot.

**Tests:** strong unit coverage; minor gaps — see inline notes for `pytest.raises(FrozenInstanceError)` and the aliased-import canary gap.

workers/queue_backend/fairness.py
```diff
@@ -0,0 +1,78 @@
...
+    """Routing metadata attached to every ``dispatch(...)``.
+
+    ``org_id=None`` is a valid value — it denotes a system / cross-org
+    task that doesn't belong to a tenant partition (e.g. periodic log
+    flushing, healthchecks). PG Queue's scheduler treats those as a
+    distinct "system" partition rather than as belonging to any tenant.
+    """
+
+    org_id: str | None
+    pipeline_priority: int = DEFAULT_PRIORITY
```
muhammad-ali-e, line 41: **Type Design Analyzer — invariant gap.** `pipeline_priority: int` has no documented or enforced bounds. A caller can pass `-100` or `10_000_000` and the Phase 8 scheduler will inherit whatever ordering semantics that implies. Two cheap options:

1. Document the intended range in the dataclass docstring (`# 0 ≤ priority ≤ 100, higher = sooner`).
2. Validate in `__post_init__`:

```python
def __post_init__(self) -> None:
    if not 0 <= self.pipeline_priority <= 100:
        raise ValueError(f"pipeline_priority out of range: {self.pipeline_priority}")
```

Far easier to add the bound now (2 callers) than after Phase 8 ships and consumers start depending on edge values.

workers/queue_backend/fairness.py
```diff
@@ -0,0 +1,78 @@
...
+    tier: str = DEFAULT_TIER
```
muhammad-ali-e: **Type Design Analyzer — typo-prone string.** `tier: str` accepts any string, so `"enterprize"` or `"Standard"` (capitalisation) silently route as their own partition. Tighten to a `Literal` (or enum) so Phase 8's scheduler matches on a closed set:

```python
from typing import Literal
Tier = Literal["standard", "enterprise", "system"]
...
tier: Tier = DEFAULT_TIER
```

Pure additive change today (mypy will catch any drift in Phase 8), and locks the wire vocabulary before more producers are added.

workers/queue_backend/fairness.py
```diff
@@ -0,0 +1,78 @@
...
+    def to_dict(self) -> dict[str, str | int | None]:
+        """JSON-safe representation suitable for ``kwargs["_fairness_key"]``."""
+        return {
+            "org_id": self.org_id,
+            "pipeline_priority": self.pipeline_priority,
+            "tier": self.tier,
+        }
+
+    @classmethod
+    def system(cls) -> FairnessKey:
```
muhammad-ali-e: **Type Design Analyzer / Code Simplifier.** `system()` is exactly `FairnessKey(org_id=None)` (defaults for the rest), so it adds an alias rather than an invariant. The module docstring promises PG Queue will treat these as "a distinct system partition" — but the *data* doesn't say `system` anywhere, only `org_id=None`.

Two paths forward:
- **Drop the helper** and let callers write `FairnessKey(org_id=None)` directly — saves a layer of indirection.
- **Encode the partition in data**: `return cls(org_id=None, tier="system")` so the "system partition" promise lives in the message, not in a constructor name. Phase 8 then matches on `tier == "system"` instead of having to special-case `org_id is None`.

The second option is the more durable choice.

workers/queue_backend/fairness.py
```diff
@@ -0,0 +1,78 @@
...
+        """Fairness key for a task with no tenant context."""
+        return cls(org_id=None)
+
+    @classmethod
+    def for_org(cls, org_id: str | None, **overrides: object) -> FairnessKey:
```
muhammad-ali-e: **Silent Failure Hunter (HIGH).** `for_org(org_id, **overrides)` silently swallows misspelled keys:

```python
FairnessKey.for_org("org-1", priority=80)         # priority dropped, default 50 used
FairnessKey.for_org("org-1", tiers="enterprise")  # tiers dropped, default "standard" used
```

Neither call raises. The defensive `int(overrides.get(...))` / `str(overrides.get(...))` casts further hide the problem. Suggest explicit keyword-only params:

```python
@classmethod
def for_org(
    cls,
    org_id: str | None,
    *,
    pipeline_priority: int = DEFAULT_PRIORITY,
    tier: str = DEFAULT_TIER,
) -> FairnessKey:
    return cls(org_id=org_id, pipeline_priority=pipeline_priority, tier=tier)
```

Now typos raise `TypeError: unexpected keyword argument` immediately. Same shape, stronger contract, no behaviour change for correct callers.

workers/queue_backend/dispatch.py
```diff
@@ -45,14 +48,26 @@ def dispatch(
...
+            Phase 8 (PG Queue Gate) will introduce the reader via
+            ``self.request.headers`` on bound tasks.
 
     Returns:
         A handle to the enqueued task. ``.id`` is guaranteed; everything
         else is substrate-specific and callers must not rely on it.
     """
+    headers: dict[str, Any] | None = None
+    if fairness is not None:
+        headers = {FAIRNESS_HEADER_NAME: fairness.to_dict()}
```
muhammad-ali-e: **Code Simplifier (minor).** Two-line setup is fine, but a ternary fits naturally and removes the mutable `headers` rebinding:

```python
headers = {FAIRNESS_HEADER_NAME: fairness.to_dict()} if fairness else None
```

Drop the optional type annotation since it's inferable. Take it or leave it — purely stylistic.

workers/tests/test_fairness_key.py
```diff
@@ -0,0 +1,246 @@
...
+        assert key.tier == DEFAULT_TIER
+
+    def test_for_org_accepts_overrides(self):
+        key = FairnessKey.for_org("org-1", pipeline_priority=80, tier="enterprise")
+        assert key.pipeline_priority == 80
+        assert key.tier == "enterprise"
+
+    def test_is_frozen(self):
+        key = FairnessKey(org_id="x")
+        try:
```
muhammad-ali-e: **PR Test Analyzer — cleaner with `pytest.raises`.** The try/except/else ladder hand-rolls what `pytest.raises` already does:

```python
from dataclasses import FrozenInstanceError

def test_is_frozen(self):
    key = FairnessKey(org_id="x")
    with pytest.raises(FrozenInstanceError):
        key.org_id = "y"  # type: ignore[misc]
```

Clearer intent, narrower exception (catches the actual `FrozenInstanceError` rather than "any Exception whose message contains 'frozen'" or any `AttributeError`).

workers/tests/test_fairness_key.py
```diff
@@ -0,0 +1,246 @@
...
+
+            def visit_Call(self, node: ast.Call) -> None:
+                # Only match bare ``dispatch(...)`` — i.e. the function
+                # imported as ``from queue_backend import dispatch``.
+                # Method calls like ``dispatcher.dispatch(...)`` belong
+                # to ExecutionDispatcher (executor-side RPC) and aren't
+                # queue boundaries — different concept, must not be
+                # audited here.
+                callee = node.func
+                if isinstance(callee, ast.Name) and callee.id == "dispatch":
```
muhammad-ali-e: **PR Test Analyzer — canary blind spot (regression risk).** The visitor matches only `ast.Name` with `callee.id == "dispatch"`, so an aliased import slips past silently:

```python
from queue_backend import dispatch as _disp
_disp("any_task", kwargs={"foo": "bar"})  # missing fairness=, but canary passes
```

Unlikely in this codebase today (both call sites use the bare name), but the whole point of the canary is to catch *future* drift. Two cheap fixes:

1. **Walk imports too**: collect every local alias bound to `queue_backend.dispatch` and audit calls of any of those names.
2. **Lock the import form**: add a second assertion that every file that contains a bare `dispatch(...)` call also has `from queue_backend import dispatch` (no `as`-aliasing).

Option 2 is simpler and probably enough.

workers/tests/test_fairness_key.py
```diff
@@ -0,0 +1,246 @@
...
+                continue
+            for line_no, line in enumerate(py.read_text().splitlines(), start=1):
+                if any(token in line for token in forbidden_tokens):
+                    readers.append(f"{rel}:{line_no}")
+
+        assert readers == [], (
+            "Found reader(s) of the fairness slot before Phase 8. "
+            "Phase 5.1 is additive-only — no consumer should exist yet. "
+            "Found:\n  " + "\n  ".join(readers)
+        )
```
muhammad-ali-e, line 281: **PR Test Analyzer — coverage gap.** The current suite is all unit-level mock-of-`current_app`. Worth adding one end-to-end assertion that the header survives Celery's signature builder — e.g. construct a `signature` from `current_app.send_task` (or eager-mode `app.send_task(...)` against a memory broker) and verify the routed message carries `headers["x-fairness-key"]` with the right shape. That would catch the case where a future Celery upgrade or kombu serializer change quietly drops unknown headers (rare, but the kind of thing the canary won't see).

Low-priority — your existing unit + characterisation coverage is strong; this is belt-and-braces for a feature that lives on the AMQP wire.

muhammad-ali-e:
## Review feedback addressed — `93e8a382`

10/10 inline findings actioned. One-to-one accounting:

### Comment Analyzer — stale docs (2/2 fixed)

| Finding | Status |
|---|---|
| `fairness.py:17` — module docstring still mentioned `_fairness_key` | ✅ Replaced with `x-fairness-key` and pinned the test class (`test_fairness_key.py::TestNoConsumerYet`) so future readers know exactly where the invariant is locked. |
| `fairness.py:57` — `to_dict()` docstring described the kwargs-merge design | ✅ Rewrote to "JSON-safe representation carried in the Celery message header `x-fairness-key`". |

### Type Design Analyzer (3/3 fixed)

| Finding | Status |
|---|---|
| `fairness.py:53` — `pipeline_priority: int` unbounded | ✅ Added `MIN_PRIORITY` / `MAX_PRIORITY` constants + `__post_init__` validator. Out-of-range values raise `ValueError` at construction. New tests pin the boundaries and rejection paths. |
| `fairness.py:54` — `tier: str` typo-prone | ✅ Tightened to `Literal["standard", "enterprise", "system"]` via a new `Tier` alias. New `SYSTEM_TIER` constant. mypy now catches typos like `"enterprize"` or `"Standard"`. |
| `fairness.py:65` — `system()` adds an alias, not an invariant | ✅ Adopted the second suggestion ("encode the partition in data") — `FairnessKey.system()` now returns `cls(org_id=None, tier=SYSTEM_TIER)`. Phase 8's scheduler can match on the closed-set `tier` field alone, no special-casing `org_id is None`. Module + method docstrings explain the choice. New test `test_system_key_encodes_partition_in_tier` pins it. |

### Silent Failure Hunter (HIGH) (1/1 fixed)

| Finding | Status |
|---|---|
| `fairness.py:70` — `for_org(**overrides)` swallows misspelled keys | ✅ Replaced `**overrides` with explicit keyword-only params (`pipeline_priority`, `tier`). New `test_for_org_rejects_misspelled_kwargs` asserts that `priority=80` and `tiers="enterprise"` now raise `TypeError`. |

### Code Simplifier (1/1 fixed)

| Finding | Status |
|---|---|
| `dispatch.py:65` — two-line setup with mutable rebinding | ✅ Collapsed to a single ternary on one line. Annotation dropped (inferable). |

### PR Test Analyzer (3/3 fixed)

| Finding | Status |
|---|---|
| `test_fairness_key.py:57` — try/except/else ladder for "is frozen" | ✅ Replaced with `pytest.raises(FrozenInstanceError)`. Catches the actual exception type instead of "any Exception whose message contains 'frozen' OR any AttributeError". |
| `test_fairness_key.py:185` — canary blind spot for aliased imports | ✅ Went with **Option 2 (lock the import form)**. New `test_dispatch_must_be_imported_unaliased` walks `ImportFrom` nodes and asserts that no file imports `dispatch` under an alias. The bare-name AST visitor downstream now has the assumption it needs. |
| `test_fairness_key.py:246` — coverage gap on header survival through Celery | ✅ Added `TestHeaderSurvivesCeleryPipeline::test_header_present_on_outbound_message`. Constructs a real `Celery(broker="memory://")` app, patches the seam's `current_app` onto it, wraps `send_task`, dispatches, and asserts the fairness header lands on the outbound call in the documented shape. Catches the kombu / Celery upgrade scenario where unknown headers might be quietly dropped. |

### Test count

* `test_fairness_key.py`: 14 → 21 tests
* `test_queue_backend_seam.py`: unchanged (17)
* `test_dispatch_sites_characterisation.py`: unchanged (14)
* **Total: 45 → 52 — all green** in 4.49s

### Notable design choice locked in

Per the type analyzer suggestion #3, `FairnessKey.system()` now emits `tier="system"` in the message itself. Phase 8's scheduler (and any future debugging tooling) can match on the closed-set `tier` field without having to remember that `org_id is None` also means "system partition". The message is self-describing.

muhammad-ali-e:
## New review findings addressed — `098c329e` (now `97c6623c` post-rebase)

3/3 valid, all fixed. Validity-check + accounting:

### Greptile #1 (P2) — `fairness.py` `for_org()` silently accepts `org_id=None`

**Valid.** My current code had:

| Constructor | Result |
|---|---|
| `FairnessKey.system()` | `org_id=None, tier="system"` ✅ matches design |
| `FairnessKey.for_org(None)` | `org_id=None, tier="standard"` ❌ inconsistent — Phase 8 would have to special-case this |

The module docstring already states the no-tenant convention is `tier="system"`, but `for_org` quietly let callers create the broken shape.

**Fix:** Tightened `for_org` signature to `org_id: str` (was `str | None`) and added an explicit `ValueError` guard with a message pointing callers at `FairnessKey.system()`. New `test_for_org_rejects_none_org_id` locks it.

### CodeRabbit (Major) — `helper.py:90` produces the inconsistent key in practice

**Valid.** `NotificationPayload.organization_id: str | None = None` (default). Callback paths build payloads via `NotificationPayload.from_execution_status(...)` which doesn't pass `organization_id`, so `payload.organization_id` can be None at the dispatch site.

With Greptile #1's tightening, that call now raises — the right move is to pick the constructor at the call site:

```python
fairness=(
    FairnessKey.for_org(payload.organization_id)
    if payload.organization_id
    else FairnessKey.system()
),
```

**Scheduler site unchanged** — `ScheduledPipelineContext.organization_id: str` (not Optional), and `__post_init__` raises if it's falsy. So `for_org(context.organization_id)` is statically safe there.

### Greptile #2 (P2) — `test_fairness_key.py:315` duplicates `_SKIP_TOP_DIRS`

**Valid.** When I refactored the AST canaries last round (cognitive-complexity fix), I introduced module-level `_SKIP_TOP_DIRS` and `_WORKERS_ROOT` constants but `test_no_consumer_reads_fairness_header` still had its own copy. Drift risk for future directory additions.

**Fix:** Replaced the local copy + `pathlib.Path(__file__)...` with the module-level constants. Both audit tests now share one source of truth.

### Test count

* `test_fairness_key.py`: 21 → 22 tests (new `test_for_org_rejects_none_org_id`)
* **Total: 52 → 53 — all green** in 5.15s

### Branch state

* Commit `098c329e` rebased onto remote merge-from-main → `97c6623c`.
* PR head now reflects all three fixes.

workers/queue_backend/fairness.py
```diff
@@ -0,0 +1,132 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Final, Literal
+
+# Closed vocabulary for ``tier``. Phase 8's scheduler matches on this
+# set; widening the set is an explicit decision (e.g. add a new tenant
+# class), not a typo. ``"system"`` is the special partition for tasks
+# without tenant context (periodic log flush, healthchecks, etc.).
+Tier = Literal["standard", "enterprise", "system"]
```
chandrasekharan-zipstack: @muhammad-ali-e can you elaborate on what these tier's do or help with?

workers/queue_backend/fairness.py
```diff
@@ -0,0 +1,132 @@
...
+
+# Bounds for ``pipeline_priority``. The scheduler interprets 0..100 with
+# higher = sooner; anything outside this range is rejected at
+# construction so producers can't accidentally invent edge values that
+# Phase 8 then has to special-case.
```
chandrasekharan-zipstack: NIT: Actively mentioning terms and phrases like  `Phase 8` in the code comments, commits .etc. can go stale really quickly.

I suggest making the comments concise and generic and not too specific to your conversations or the plan it is following. This is for future context hygiene

workers/queue_backend/fairness.py
```diff
@@ -0,0 +1,132 @@
...
+# request with no per-pipeline priority configured.
+DEFAULT_PRIORITY: Final[int] = 50
+DEFAULT_TIER: Final[Tier] = "standard"
+SYSTEM_TIER: Final[Tier] = "system"
+
+# Celery message-header slot that carries the fairness key. Headers
+# travel with the AMQP message but are NOT passed to the task body's
+# function signature — exactly what we want for routing metadata.
+# (Earlier iteration of this module put the key in ``kwargs``; that
+# blew up tasks whose signature didn't accept ``**kwargs``.)
```
chandrasekharan-zipstack: NIT: Information such as these might not be required in the long run and can be trimmed down

chandrasekharan-zipstack (approved):

workers/queue_backend/fairness.py
muhammad-ali-e: Agreed, fair NIT. Genericised Phase 8 (PG Queue Gate) -> a future dispatch scheduler / the scheduler across both fairness.py and dispatch.py (module docstrings, FairnessKey, system(), DispatchHandle, and the dispatch() docstring). Comments now describe what the code does, not when in the plan it lives. `f440e4e8`.

workers/queue_backend/fairness.py
muhammad-ali-e: Agreed — dropped the Earlier iteration of this module put the key in kwargs; that blew up tasks... parenthetical. That history belongs in the commit log, not the source file. Comment now just states what the constant is for. `f440e4e8`.

workers/queue_backend/fairness.py
muhammad-ali-e: it was just a plumbing part. But I think  the tire is not the actual things we required, let me check and fix it

muhammad-ali-e: @chandrasekharan-zipstack — re-read the labs design carefully and tightened both the model and the code (latest in `02b6a4b4`). Short answer: **we don't carry tier on the wire** because it lives server-side. Here's the precise state.

### In the design

Tier is a property of the **org**, stored in a SQL lookup table (illustrative values):

```
org_id          | burst_max | tier_priority
org-free        |   5       |    10
org-pro         |   20      |    5
org-enterprise  |   100     |    1     (1 = highest priority)
```

The scheduler JOINs against this table by `org_id` and orders by `tier_priority ASC`. Tier shows up at admission time — derived from org_id — not from anything on the task payload.

### In the code

The `FairnessKey` dataclass carries three fields, matching the staging-queue ORDER BY's L1/L2/L3 inputs:

| Field | Where used | What it drives |
|---|---|---|
| `org_id: str \| None` | Server-side JOIN against the org config table | L1 (tier lookup) |
| `workload_type: WorkloadType` | Carried in the `x-fairness-key` header | L2 (`api` vs `non_api`, binary) |
| `pipeline_priority: int` (1..10, default 5) | Carried in the `x-fairness-key` header | L3 (within-tier ordering) |

No tier field. A producer doesn't need to know an org's tier — it just identifies the org. Putting tier on the wire would be redundant (the scheduler ignores it; it trusts the config table), drift-prone (stale tier on in-flight tasks vs latest config), and a leak of internal scheduling policy into producer code.

### Vocabulary correctness

`WorkloadType` is now a `StrEnum` with `API` and `NON_API` — matches the labs L2 check (`(workload_type = 'api')::int DESC`), which is strictly binary. ETL, Task, App workflows all roll up to `NON_API` because the scheduler doesn't distinguish them at L2.

### Scope correction

I'd also confused workload_type with worker type earlier — `workload_type` is the **workflow execution** type, not the worker doing the dispatching. So in this PR:

| Dispatch site | Is it a workflow execution? | `fairness=` |
|---|---|---|
| Scheduler → `async_execute_bin` | Yes (scheduled pipeline run) | `FairnessKey(org_id=..., workload_type=NON_API)` |
| Notification helper → `send_webhook_notification` | No (worker-internal task on the `notifications` queue) | `None` (explicit opt-out) |

Phase 6 (chord lift in api-deployment) will introduce the `workload_type=API` case.

So to answer the original question: **the tier concept is documented in the design and looked up server-side; it does not appear on `FairnessKey` or in worker code today, and it doesn't need to**.

workers/tests/test_fairness_key.py
```diff
@@ -0,0 +1,310 @@
...
+            for rel, tree in _iter_production_trees()
+            for lineno in _dispatch_calls_missing_fairness(tree)
+        ]
+        assert offenders == [], (
+            "Production dispatch(...) call site(s) missing fairness=. "
+            "Every production dispatch must declare its fairness key — "
+            "pass ``fairness=FairnessKey.system()`` for system / "
+            "cross-org tasks if there's no tenant context. Found:\n  "
+            + "\n  ".join(offenders)
+        )
```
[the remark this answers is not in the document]
muhammad-ali-e: Valid P1, fixed in `f914c9a2`. The reference to `FairnessKey.system()` in the assertion message was stale — that classmethod was deleted when I dropped the helper constructors in favour of direct dataclass construction. Following the message verbatim would indeed have hit `AttributeError`.

Updated to point at the two real options the canary accepts:

* `fairness=FairnessKey(org_id=..., workload_type=WorkloadType...)` for a workflow-execution dispatch.
* `fairness=None` for a worker-internal task that doesn't start a workflow execution (notifications, callbacks, healthchecks).

Slightly broader than your suggestion since the canary actually has two valid resolutions, not one — passing a key vs explicit None. Both are documented in the message now.

Deepak-Kesavan (approved):

PR #2009, opened 2026-06-02
UN-3508 [FEAT] Plumb fairness through ExecutionDispatcher

muhammad-ali-e:
## What

- Add optional `headers: dict | None` kwarg to all three dispatch methods on `unstract.sdk1.execution.dispatcher.ExecutionDispatcher` (`dispatch`, `dispatch_async`, `dispatch_with_callback`). When non-None, forwarded as Celery message headers; when None, the kwarg is omitted from `send_task` so existing callers see the identical pre-change call shape.
- New `FairnessKey.as_header()` helper in `workers/queue_backend/fairness.py` returns the wire-ready `{"x-fairness-key": ...}` dict, so producers don't need to import the slot-name constant.
- Wire fairness from the three `dispatcher.dispatch(...)` sites in `workers/file_processing/structure_tool_task.py` (lines 468, 507, 720). Each passes `headers=_fairness_headers(organization_id)`. A small in-file helper defaults `workload_type` to `NON_API`; propagating the real type from upstream is Phase 6 work.
- New `workers/tests/test_executor_dispatch.py` (7 tests):
  - Header forwarding through all three dispatcher methods.
  - "Omit when None" — call shape preservation for callers that don't opt in (sdk1 round-trip tests still green unchanged).
  - `FairnessKey.as_header()` shape.
  - AST inventory canary: no production code calls `*.send_task("execute_extraction", ...)` outside `ExecutionDispatcher`.

## Why

UN-3501 (#2003) plumbed fairness on bare `dispatch(...)` call sites — but `execute_extraction`, the most workflow-execution-y dispatch in the codebase, bypasses `queue_backend.dispatch()` entirely (it uses sdk1's `ExecutionDispatcher`). The Phase 5.1 canary audits bare-name `dispatch()` and missed it. This PR brings the third dispatch path under the same fairness umbrella.

Under Celery today the header sits inertly — no consumer reads it. Under the future PG Queue substrate, `execute_extraction` tasks land in the staging queue and the fairness scheduler sorts on the same three fields. Producer-side plumbing now means the consumer can land without backfilling in-flight payloads.

## How

- `ExecutionDispatcher` only attaches `headers=` to `send_task` when the caller provides them. That preserves the pre-change `mock.assert_called_with(...)` shape used by sdk1's existing 78 dispatcher tests — they all stay green without modification.
- `FairnessKey.as_header()` shifts the wire-shape knowledge into the seam module, so producers compose a `FairnessKey(...)` and call `.as_header()`. The constant `FAIRNESS_HEADER_NAME` stays internal to `queue_backend.fairness` — keeps the "no consumer reads it yet" canary green at the producer level.
- The `_fairness_headers` helper in `structure_tool_task.py` is intentionally one function: same workload_type for all three sites today; Phase 6 (chord lift) is the right place to thread the real workload type from the workflow's class.

No.

- `headers` is optional and defaults to None on all three dispatcher methods. Callers that don't opt in see the identical `send_task` call shape.
- All 78 existing sdk1 dispatcher tests pass unchanged.
- Worker seam suite extended (53 → 60 tests, all green).
- Producer-side only: no consumer in workers/ reads `x-fairness-key` yet (canary in test_fairness_key.py still asserts this).
- No queue routing, task name, or args/kwargs change.

## Database Migrations

- None.

## Env Config

- None.

## Relevant Docs

- N/A. Module-level docstrings in `queue_backend/fairness.py` and the dispatcher cover the header semantics.

## Related Issues or PRs

- Ticket: UN-3508 (sub-task) under UN-3500 (Story: PG Queue Phase 5 — Payload typing) under UN-3445 (Epic).
- Builds on: UN-3501 (#2003).

## Dependencies Versions

- None.

## Notes on Testing

\`\`\`
cd workers
.venv/bin/python -m pytest \\
    tests/test_executor_dispatch.py \\
    tests/test_fairness_key.py \\
    tests/test_queue_backend_seam.py \\
# 60 passed in ~6s

cd ../unstract/sdk1
uv run pytest tests/test_execution.py
# 80 passed
\`\`\`

Verification:

\`\`\`
# Every execute_extraction dispatch goes through ExecutionDispatcher:
grep -rn "send_task.*execute_extraction" workers/ --include="*.py" \\
    | grep -v "ExecutionDispatcher\\|tests/\\|sdk1/"
# (empty — only sdk1's ExecutionDispatcher.dispatch* methods)
\`\`\`

## Screenshots

N/A (no UI surface).

## Checklist

### DIFF
--- a/unstract/sdk1/src/unstract/sdk1/execution/dispatcher.py
+++ b/unstract/sdk1/src/unstract/sdk1/execution/dispatcher.py
@@ -103,6 +103,7 @@ class ExecutionDispatcher:
         self,
         context: ExecutionContext,
         timeout: int | None = None,
+        headers: dict[str, Any] | None = None,
     ) -> ExecutionResult:
         """Dispatch context as a Celery task and wait for result.
 
@@ -111,6 +112,9 @@ class ExecutionDispatcher:
             timeout: Max seconds to wait.  ``None`` reads from
                 the ``EXECUTOR_RESULT_TIMEOUT`` env var,
                 falling back to 3600s.
+            headers: Optional Celery message headers (routing metadata
+                out-of-band of the task body — see callers for
+                ``x-fairness-key`` usage).
 
         Returns:
             ExecutionResult from the executor.
@@ -136,11 +140,13 @@ class ExecutionDispatcher:
             queue,
         )
 
-        async_result = self._app.send_task(
-            _TASK_NAME,
-            args=[context.to_dict()],
-            queue=queue,
-        )
+        send_kwargs: dict[str, Any] = {
+            "args": [context.to_dict()],
+            "queue": queue,
+        }
+        if headers is not None:
+            send_kwargs["headers"] = headers
+        async_result = self._app.send_task(_TASK_NAME, **send_kwargs)
         logger.info(
             "Task sent: celery_task_id=%s, waiting for result...",
             async_result.id,
@@ -172,11 +178,13 @@ class ExecutionDispatcher:
     def dispatch_async(
         self,
         context: ExecutionContext,
+        headers: dict[str, Any] | None = None,
     ) -> str:
         """Dispatch without waiting.  Returns task_id for polling.
 
         Args:
             context: ExecutionContext to dispatch.
+            headers: Optional Celery message headers (see ``dispatch``).
 
         Returns:
             The Celery task ID (use with ``AsyncResult`` to poll).
@@ -198,11 +206,13 @@ class ExecutionDispatcher:
             queue,
         )
 
-        async_result = self._app.send_task(
-            _TASK_NAME,
-            args=[context.to_dict()],
-            queue=queue,
-        )
+        send_kwargs: dict[str, Any] = {
+            "args": [context.to_dict()],
+            "queue": queue,
+        }
+        if headers is not None:
+            send_kwargs["headers"] = headers
+        async_result = self._app.send_task(_TASK_NAME, **send_kwargs)
         return async_result.id
 
     def dispatch_with_callback(
@@ -211,6 +221,7 @@ class ExecutionDispatcher:
         on_success: Signature | None = None,
         on_error: Signature | None = None,
         task_id: str | None = None,
+        headers: dict[str, Any] | None = None,
     ) -> AsyncResult:
         """Fire-and-forget dispatch with Celery link callbacks.
 
@@ -267,6 +278,8 @@ class ExecutionDispatcher:
             send_kwargs["link_error"] = on_error
         if task_id is not None:
             send_kwargs["task_id"] = task_id
+        if headers is not None:
+            send_kwargs["headers"] = headers
 
         async_result = self._app.send_task(
             _TASK_NAME,
--- a/unstract/sdk1/src/unstract/sdk1/execution/dispatcher.py
+++ b/unstract/sdk1/src/unstract/sdk1/execution/dispatcher.py
@@ -120,7 +120,11 @@ class ExecutionDispatcher:
             "args": [context.to_dict()],
             "queue": queue,
         }
-        if headers is not None:
+        # Treat falsy (None, ``{}``) as drop. An empty header dict is
+        # never legitimate — ``FairnessKey.as_header()`` always returns
+        # a populated mapping — so silently forwarding ``{}`` would
+        # document a producer-side build bug rather than catch it.
+        if headers:
             out["headers"] = headers
         if link is not None:
             out["link"] = link
@@ -147,9 +151,11 @@ class ExecutionDispatcher:
                 out-of-band of the task body — see
                 ``queue_backend.fairness.FAIRNESS_HEADER_NAME``
                 (currently ``"x-fairness-key"``) for the canonical
-                constant). Caller is responsible for the header schema;
-                passing ``{}`` is forwarded as-is and likely indicates
-                a producer-side build bug.
+                constant). Caller is responsible for the header schema.
+                Falsy values (``None``, ``{}``) are dropped — the
+                downstream ``send_task`` call is invoked without
+                ``headers=`` so the on-wire shape matches the
+                no-headers baseline.
 
         Returns:
             ExecutionResult from the executor.
--- a/unstract/sdk1/src/unstract/sdk1/execution/dispatcher.py
+++ b/unstract/sdk1/src/unstract/sdk1/execution/dispatcher.py
@@ -99,6 +99,37 @@ class ExecutionDispatcher:
         """
         return f"{_QUEUE_PREFIX}{executor_name}"
 
+    @staticmethod
+    def _build_send_kwargs(
+        context: ExecutionContext,
+        queue: str,
+        *,
+        headers: dict[str, Any] | None = None,
+        link: Signature | None = None,
+        link_error: Signature | None = None,
+        task_id: str | None = None,
+    ) -> dict[str, Any]:
+        """Common ``send_task`` kwargs builder for the three dispatch modes.
+
+        Optional kwargs are dropped when ``None`` so the resulting call
+        shape matches the minimum-arity form Celery expects for ``link``
+        / ``link_error`` handling and so callers that pass no headers
+        see the same on-wire call as before headers existed.
+        """
+        out: dict[str, Any] = {
+            "args": [context.to_dict()],
+            "queue": queue,
+        }
+        if headers is not None:
+            out["headers"] = headers
+        if link is not None:
+            out["link"] = link
+        if link_error is not None:
+            out["link_error"] = link_error
+        if task_id is not None:
+            out["task_id"] = task_id
+        return out
+
     def dispatch(
         self,
         context: ExecutionContext,
@@ -113,8 +144,12 @@ class ExecutionDispatcher:
                 the ``EXECUTOR_RESULT_TIMEOUT`` env var,
                 falling back to 3600s.
             headers: Optional Celery message headers (routing metadata
-                out-of-band of the task body — see callers for
-                ``x-fairness-key`` usage).
+                out-of-band of the task body — see
+                ``queue_backend.fairness.FAIRNESS_HEADER_NAME``
+                (currently ``"x-fairness-key"``) for the canonical
+                constant). Caller is responsible for the header schema;
+                passing ``{}`` is forwarded as-is and likely indicates
+                a producer-side build bug.
 
         Returns:
             ExecutionResult from the executor.
@@ -140,12 +175,7 @@ class ExecutionDispatcher:
             queue,
         )
 
-        send_kwargs: dict[str, Any] = {
-            "args": [context.to_dict()],
-            "queue": queue,
-        }
-        if headers is not None:
-            send_kwargs["headers"] = headers
+        send_kwargs = self._build_send_kwargs(context, queue, headers=headers)
         async_result = self._app.send_task(_TASK_NAME, **send_kwargs)
         logger.info(
             "Task sent: celery_task_id=%s, waiting for result...",
@@ -206,12 +236,7 @@ class ExecutionDispatcher:
             queue,
         )
 
-        send_kwargs: dict[str, Any] = {
-            "args": [context.to_dict()],
-            "queue": queue,
-        }
-        if headers is not None:
-            send_kwargs["headers"] = headers
+        send_kwargs = self._build_send_kwargs(context, queue, headers=headers)
         async_result = self._app.send_task(_TASK_NAME, **send_kwargs)
         return async_result.id
 
@@ -242,6 +267,7 @@ class ExecutionDispatcher:
             task_id: Optional pre-generated Celery task ID. Useful
                 when the caller needs to know the task ID before
                 dispatch (e.g. to include it in callback kwargs).
+            headers: Optional Celery message headers (see ``dispatch``).
 
         Returns:
             The ``AsyncResult`` from ``send_task``.  Callers can
@@ -268,23 +294,15 @@ class ExecutionDispatcher:
             queue,
         )
 
-        send_kwargs: dict[str, Any] = {
-            "args": [context.to_dict()],
-            "queue": queue,
-        }
-        if on_success is not None:
-            send_kwargs["link"] = on_success
-        if on_error is not None:
-            send_kwargs["link_error"] = on_error
-        if task_id is not None:
-            send_kwargs["task_id"] = task_id
-        if headers is not None:
-            send_kwargs["headers"] = headers
-
-        async_result = self._app.send_task(
-            _TASK_NAME,
-            **send_kwargs,
+        send_kwargs = self._build_send_kwargs(
+            context,
+            queue,
+            headers=headers,
+            link=on_success,
+            link_error=on_error,
+            task_id=task_id,
         )
+        async_result = self._app.send_task(_TASK_NAME, **send_kwargs)
         logger.info(
             "Task sent with callbacks: celery_task_id=%s",
             async_result.id,
--- a/workers/file_processing/structure_tool_task.py
+++ b/workers/file_processing/structure_tool_task.py
@@ -40,13 +40,19 @@ logger = logging.getLogger(__name__)
 EXECUTOR_TIMEOUT = int(os.environ.get("EXECUTOR_RESULT_TIMEOUT", 3600))
 
 
-def _fairness_headers(organization_id: str) -> dict[str, object]:
+def _fairness_headers(
+    organization_id: str,
+) -> dict[str, dict[str, str | int | None]]:
     """Fairness header for executor dispatches.
 
     Structure-tool dispatches are workflow-execution work — both ETL
-    and API workflows route through here. Defaults to ``NON_API``;
-    propagating the actual workload type from the caller is Phase 6
-    work (chord lift).
+    and API workflows route through here. ``NON_API`` is the safe
+    default (API traffic preempting is a strictly weaker mistake than
+    the inverse).
+
+    TODO(UN-3504): propagate the caller's WorkloadType (API vs ETL)
+    instead of hard-coding ``NON_API``. Requires the chord-lift work
+    so the workflow type flows from the chord caller down to here.
     """
     return FairnessKey(
         org_id=organization_id,
--- a/workers/file_processing/structure_tool_task.py
+++ b/workers/file_processing/structure_tool_task.py
@@ -23,7 +23,8 @@ from pathlib import Path
 from typing import Any
 
 from file_processing.worker import app
-from queue_backend import worker_task
+from queue_backend import FairnessKey, worker_task
+from queue_backend.fairness import WorkloadType
 from shared.enums.task_enums import TaskName
 from shared.infrastructure.context import StateStore
 
@@ -37,6 +38,20 @@ logger = logging.getLogger(__name__)
 # Timeout for executor worker calls (seconds).
 # Reads from EXECUTOR_RESULT_TIMEOUT env, defaults to 3600.
 EXECUTOR_TIMEOUT = int(os.environ.get("EXECUTOR_RESULT_TIMEOUT", 3600))
+
+
+def _fairness_headers(organization_id: str) -> dict[str, object]:
+    """Fairness header for executor dispatches.
+
+    Structure-tool dispatches are workflow-execution work — both ETL
+    and API workflows route through here. Defaults to ``NON_API``;
+    propagating the actual workload type from the caller is Phase 6
+    work (chord lift).
+    """
+    return FairnessKey(
+        org_id=organization_id,
+        workload_type=WorkloadType.NON_API,
+    ).as_header()
 
 
 # -----------------------------------------------------------------------
@@ -465,7 +480,11 @@ def _execute_structure_tool_impl(params: dict) -> dict:
             file_execution_id=file_execution_id,
             executor_params=agentic_params,
         )
-        at_result = dispatcher.dispatch(at_ctx, timeout=EXECUTOR_TIMEOUT)
+        at_result = dispatcher.dispatch(
+            at_ctx,
+            timeout=EXECUTOR_TIMEOUT,
+            headers=_fairness_headers(organization_id),
+        )
         if not at_result.success:
             return at_result.to_dict()
         at_output_data = at_result.data.get("output", {}) or {}
@@ -504,7 +523,11 @@ def _execute_structure_tool_impl(params: dict) -> dict:
             },
         )
         pipeline_start = time.monotonic()
-        pipeline_result = dispatcher.dispatch(pipeline_ctx, timeout=EXECUTOR_TIMEOUT)
+        pipeline_result = dispatcher.dispatch(
+            pipeline_ctx,
+            timeout=EXECUTOR_TIMEOUT,
+            headers=_fairness_headers(organization_id),
+        )
         pipeline_elapsed = time.monotonic() - pipeline_start
 
         if not pipeline_result.success:
@@ -717,7 +740,11 @@ def _run_agentic_extraction(
             "include_source_refs": enable_highlight,
         },
     )
-    agentic_result = dispatcher.dispatch(agentic_ctx, timeout=EXECUTOR_TIMEOUT)
+    agentic_result = dispatcher.dispatch(
+        agentic_ctx,
+        timeout=EXECUTOR_TIMEOUT,
+        headers=_fairness_headers(organization_id),
+    )
 
     if not agentic_result.success:
         return agentic_result.to_dict()
--- a/workers/queue_backend/fairness.py
+++ b/workers/queue_backend/fairness.py
@@ -55,5 +55,8 @@ class FairnessKey:
         }
 
     def as_header(self) -> dict[str, dict[str, str | int | None]]:
-        """Wire-ready Celery ``headers=`` payload."""
+        """Celery ``send_task(headers=...)`` payload.
+
+        Shape: ``{FAIRNESS_HEADER_NAME: self.to_dict()}``.
+        """
         return {FAIRNESS_HEADER_NAME: self.to_dict()}
--- a/workers/queue_backend/fairness.py
+++ b/workers/queue_backend/fairness.py
@@ -53,3 +53,7 @@ class FairnessKey:
             "workload_type": self.workload_type.value,
             "pipeline_priority": self.pipeline_priority,
         }
+
+    def as_header(self) -> dict[str, dict[str, str | int | None]]:
+        """Wire-ready Celery ``headers=`` payload."""
+        return {FAIRNESS_HEADER_NAME: self.to_dict()}
--- a/workers/tests/canary_helpers.py
+++ b/workers/tests/canary_helpers.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import ast
 import pathlib
+import warnings
 
 WORKERS_ROOT = pathlib.Path(__file__).parent.parent
 
@@ -22,8 +23,11 @@ def iter_production_trees(
 ) -> list[tuple[pathlib.Path, ast.AST]]:
     """Yield ``(rel_path, parsed_tree)`` for every .py file outside the skip set.
 
-    Files that fail to parse are silently dropped — the canaries assume
-    well-formed production source.
+    Unparseable files emit a ``UserWarning`` and are skipped — silent
+    skipping would let canaries pass vacuously on a botched merge.
+    Tests can promote the warning to an error via
+    ``warnings.simplefilter("error", UserWarning)`` if they want
+    strict enforcement.
     """
     out: list[tuple[pathlib.Path, ast.AST]] = []
     for py in WORKERS_ROOT.rglob("*.py"):
@@ -32,7 +36,12 @@ def iter_production_trees(
             continue
         try:
             tree = ast.parse(py.read_text(), filename=str(py))
-        except SyntaxError:
+        except SyntaxError as exc:
+            warnings.warn(
+                f"canary scan skipping unparseable file {rel}: {exc}",
+                UserWarning,
+                stacklevel=2,
+            )
             continue
         out.append((rel, tree))
     return out
--- a/workers/tests/canary_helpers.py
+++ b/workers/tests/canary_helpers.py
@@ -2,7 +2,11 @@
 
 Several test files audit the production code tree (forbid a raw
 dispatch, require a kwarg, etc.). The file-walking + parse-with-skip
-logic is identical across them; this module is the single home.
+logic is identical across them; this module is the intended single
+home. Two pre-existing characterisation tests still inline the same
+walk (``test_dispatch_sites_characterisation.py``,
+``test_chord_sites_characterisation.py``) — migrating those is
+tracked as a follow-up.
 """
 
 from __future__ import annotations
--- a/workers/tests/canary_helpers.py
+++ b/workers/tests/canary_helpers.py
@@ -1,0 +1,38 @@
+"""Shared AST-walk helpers for inventory-style canary tests.
+
+Several test files audit the production code tree (forbid a raw
+dispatch, require a kwarg, etc.). The file-walking + parse-with-skip
+logic is identical across them; this module is the single home.
+"""
+
+from __future__ import annotations
+
+import ast
+import pathlib
+
+WORKERS_ROOT = pathlib.Path(__file__).parent.parent
+
+# Always-skip top-level directories. Callers pass a wider set if they
+# need to additionally exclude e.g. the seam module itself.
+DEFAULT_SKIP_TOP_DIRS = frozenset({"tests", "__pycache__", "htmlcov", ".venv"})
+
+
+def iter_production_trees(
+    skip_top_dirs: frozenset[str] = DEFAULT_SKIP_TOP_DIRS,
+) -> list[tuple[pathlib.Path, ast.AST]]:
+    """Yield ``(rel_path, parsed_tree)`` for every .py file outside the skip set.
+
+    Files that fail to parse are silently dropped — the canaries assume
+    well-formed production source.
+    """
+    out: list[tuple[pathlib.Path, ast.AST]] = []
+    for py in WORKERS_ROOT.rglob("*.py"):
+        rel = py.relative_to(WORKERS_ROOT)
+        if rel.parts and rel.parts[0] in skip_top_dirs:
+            continue
+        try:
+            tree = ast.parse(py.read_text(), filename=str(py))
+        except SyntaxError:
+            continue
+        out.append((rel, tree))
+    return out
--- a/workers/tests/canary_helpers.py
+++ b/workers/tests/canary_helpers.py
@@ -21,7 +21,8 @@ DEFAULT_SKIP_TOP_DIRS = frozenset({"tests", "__pycache__", "htmlcov", ".venv"})
 def iter_production_trees(
     skip_top_dirs: frozenset[str] = DEFAULT_SKIP_TOP_DIRS,
 ) -> list[tuple[pathlib.Path, ast.AST]]:
-    """Yield ``(rel_path, parsed_tree)`` for every .py file outside the skip set.
+    """Return a list of ``(rel_path, parsed_tree)`` tuples for every
+    .py file outside the skip set.
 
     Unparseable files emit a ``UserWarning`` and are skipped — silent
     skipping would let canaries pass vacuously on a botched merge.
--- a/workers/tests/test_executor_dispatch.py
+++ b/workers/tests/test_executor_dispatch.py
@@ -1,8 +1,11 @@
-"""Tests for the file_processing -> executor boundary (PG Queue Phase 5.2)."""
+"""Tests for the file_processing → executor boundary: header forwarding
+through ExecutionDispatcher and inventory canary for raw send_task calls.
+"""
 
 from __future__ import annotations
 
 import ast
+from typing import Any
 from unittest.mock import MagicMock
 
 import pytest
@@ -10,13 +13,14 @@ from celery import Celery
 
 from queue_backend import FairnessKey
 from queue_backend.fairness import FAIRNESS_HEADER_NAME, WorkloadType
-from .canary_helpers import iter_production_trees
 from unstract.sdk1.execution.context import ExecutionContext, Operation
 from unstract.sdk1.execution.dispatcher import ExecutionDispatcher
 
+from .canary_helpers import iter_production_trees
 
-def _make_context(**overrides: object) -> ExecutionContext:
-    defaults: dict[str, object] = {
+
+def _make_context(**overrides: Any) -> ExecutionContext:
+    defaults: dict[str, Any] = {
         "executor_name": "legacy",
         "operation": Operation.EXTRACT,
         "run_id": "file-exec-1",
@@ -29,7 +33,7 @@ def _make_context(**overrides: object) -> ExecutionContext:
         "executor_params": {"adapter_instance_id": "a"},
     }
     defaults.update(overrides)
-    return ExecutionContext(**defaults)  # type: ignore[arg-type]
+    return ExecutionContext(**defaults)
 
 
 class TestExecutionDispatcherForwardsHeaders:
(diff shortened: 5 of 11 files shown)


unstract/sdk1/src/unstract/sdk1/execution/dispatcher.py
```diff
@@ -211,6 +221,7 @@ def dispatch_with_callback(
         on_success: Signature | None = None,
         on_error: Signature | None = None,
         task_id: str | None = None,
+        headers: dict[str, Any] | None = None,
```
muhammad-ali-e, line 255: **[HIGH — docstring completeness]** `dispatch_with_callback` gained the `headers` parameter in the signature, but its `Args:` block (lines 234–244) was **not** updated to document it. The sibling methods `dispatch` (line 115) and `dispatch_async` (line 187) both got the new `headers:` Args entry — this method is the odd one out on a public SDK surface.

**Fix:** add an entry after `task_id:` mirroring `dispatch_async`:
```
headers: Optional Celery message headers (see ``dispatch``).
```

unstract/sdk1/src/unstract/sdk1/execution/dispatcher.py
```diff
@@ -111,6 +112,9 @@ def dispatch(
             timeout: Max seconds to wait.  ``None`` reads from
                 the ``EXECUTOR_RESULT_TIMEOUT`` env var,
                 falling back to 3600s.
+            headers: Optional Celery message headers (routing metadata
+                out-of-band of the task body — see callers for
+                ``x-fairness-key`` usage).
```
muhammad-ali-e: **[LOW — comment quality]** Referencing `x-fairness-key` as a bare string literal forces a future reader to grep to find the constant. Reference the symbol instead so renaming the constant in `queue_backend.fairness` will not silently desync this docstring.

**Suggested wording:** `… see callers for ``queue_backend.fairness.FAIRNESS_HEADER_NAME`` usage (currently ``x-fairness-key``).`

unstract/sdk1/src/unstract/sdk1/execution/dispatcher.py
```diff
@@ -136,11 +140,13 @@ def dispatch(
...
-        async_result = self._app.send_task(
-            _TASK_NAME,
-            args=[context.to_dict()],
-            queue=queue,
-        )
+        send_kwargs: dict[str, Any] = {
+            "args": [context.to_dict()],
+            "queue": queue,
+        }
+        if headers is not None:
```
muhammad-ali-e: **[MEDIUM — silent footgun + repetition]** Two things:

1. **Empty dict forwarded.** `if headers is not None` means `headers={}` is forwarded to Celery as an empty headers dict — masks a caller bug where they meant to build a fairness header and produced `{}`. Worse, `headers={"wrong-key": ...}` is silently relayed; no schema check that `FAIRNESS_HEADER_NAME` is the key.
2. **Repetition.** This `send_kwargs: dict[str, Any] = {...}; if headers is not None: send_kwargs["headers"] = headers; self._app.send_task(_TASK_NAME, **send_kwargs)` block appears verbatim here and at line 209, plus a related variant at 271-282. Extracting `_build_send_kwargs(context, headers, **extra)` would kill the three copies (~−10 lines) and make the optional-kwarg contract one-line auditable.

**Minimum fix:** widen the docstring to state the contract ("caller is responsible for the header schema; empty dict is forwarded as-is"). **Better fix:** validate at the boundary — `if headers is not None and not isinstance(headers, dict): raise TypeError(...)`.

workers/file_processing/structure_tool_task.py
```diff
@@ -39,6 +40,20 @@
 EXECUTOR_TIMEOUT = int(os.environ.get("EXECUTOR_RESULT_TIMEOUT", 3600))
 
 
+def _fairness_headers(organization_id: str) -> dict[str, object]:
```
muhammad-ali-e: **[HIGH — typing widens unnecessarily]** Return type is annotated `dict[str, object]`, but the body returns `FairnessKey.as_header()` which has the strictly narrower type `dict[str, dict[str, str | int | None]]`. The widening is gratuitous and silently throws away type information at every call site, and `dict[str, object]` is not even a subtype of `dict[str, Any]` under strict variance.

**Fix:** annotate as `dict[str, dict[str, str | int | None]]` (matches `as_header()`) — or `dict[str, Any]` if you want to match the dispatcher's parameter type. Matching `as_header()` is preferred.

workers/file_processing/structure_tool_task.py
```diff
@@ -39,6 +40,20 @@
...
+    """Fairness header for executor dispatches.
+
+    Structure-tool dispatches are workflow-execution work — both ETL
+    and API workflows route through here. Defaults to ``NON_API``;
+    propagating the actual workload type from the caller is Phase 6
+    work (chord lift).
```
muhammad-ali-e: **[HIGH — Phase 6 default has no grep-able anchor + jargon]** This helper hard-codes `WorkloadType.NON_API` for **all three** executor dispatch sites, including API-workflow paths. The docstring's "Phase 6 work (chord lift)" is the right call-out but:

1. No `TODO(UN-XXXX)` ticket reference — `grep -r TODO` will not surface this.
2. "Phase 6 (chord lift)" is project-internal vocabulary that rots; once Phase 6 ships the comment becomes a lie and the next maintainer will not know what "chord lift" means.
3. No runtime signal — if Phase 6 is delayed or forgotten, API workflows are silently mis-classified as `NON_API` on the fairness queue (the exact bug fairness is meant to prevent).

**Suggested fix:** replace the "Phase 6 work (chord lift)" sentence with `# TODO(UN-XXXX): propagate the caller's WorkloadType (API vs ETL) instead of hard-coding NON_API.` and reference the actual ticket. Optionally accept `workload_type` as a required arg so callers must opt into the default explicitly.

workers/queue_backend/fairness.py
```diff
@@ -53,3 +53,7 @@ def to_dict(self) -> dict[str, str | int | None]:
             "workload_type": self.workload_type.value,
             "pipeline_priority": self.pipeline_priority,
         }
+
+    def as_header(self) -> dict[str, dict[str, str | int | None]]:
+        """Wire-ready Celery ``headers=`` payload."""
```
muhammad-ali-e: **[NIT — comment quality]** "Wire-ready" is cute but vague. State the actual contract so the call sites do not need to read the body to know the shape:

```python
"""Wire-ready Celery ``send_task(headers=...)`` payload: ``{FAIRNESS_HEADER_NAME: self.to_dict()}``."""
```

workers/tests/canary_helpers.py
```diff
@@ -0,0 +1,38 @@
...
+    well-formed production source.
+    """
+    out: list[tuple[pathlib.Path, ast.AST]] = []
+    for py in WORKERS_ROOT.rglob("*.py"):
+        rel = py.relative_to(WORKERS_ROOT)
+        if rel.parts and rel.parts[0] in skip_top_dirs:
+            continue
+        try:
+            tree = ast.parse(py.read_text(), filename=str(py))
+        except SyntaxError:
```
muhammad-ali-e: **[HIGH — silent failure that defeats every canary]** This bare `except SyntaxError: continue` silently drops unparseable production files. A real syntax error (e.g. a botched merge in `structure_tool_task.py`) would vanish from the inventory — the "no raw `send_task(\"execute_extraction\", ...)`" canary in `test_executor_dispatch.py` and the "every dispatch passes fairness" canary in `test_fairness_key.py` would both pass vacuously on the unparseable file. The docstring (line 25) even acknowledges "files are silently dropped" — that is the bug, not a justification.

**Fix:** at minimum, surface unparseable files so CI catches the regression:
```python
import warnings
...
except SyntaxError as exc:
    warnings.warn(f"canary skipping unparseable {py}: {exc}")
    continue
```
Better still: collect them into an `errors: list[Path]` and let the caller assert it is empty, so the canary's own invariant "every prod file parses" becomes a tested assertion.

workers/tests/test_executor_dispatch.py
```diff
@@ -0,0 +1,134 @@
+"""Tests for the file_processing -> executor boundary (PG Queue Phase 5.2)."""
```
muhammad-ali-e: **[LOW — comment quality]** Module docstring leaks the project-internal vocabulary "PG Queue Phase 5.2". Once Phase 5.2 fades from team memory, the docstring becomes meaningless trivia. Same issue at line 55 (`# Matches the pre-Phase-5.2 send_task call shape …`).

**Suggested rewording:** `"""Tests for the file_processing → executor boundary: header forwarding through ExecutionDispatcher and inventory canary for raw send_task calls."""` — describe what the tests cover, not which sprint added them.

workers/tests/test_executor_dispatch.py
```diff
@@ -0,0 +1,134 @@
...
+
+import ast
+from unittest.mock import MagicMock
+
+import pytest
+from celery import Celery
+
+from queue_backend import FairnessKey
+from queue_backend.fairness import FAIRNESS_HEADER_NAME, WorkloadType
+from .canary_helpers import iter_production_trees
```
muhammad-ali-e: **[LOW — style]** Import ordering: the relative `from .canary_helpers import iter_production_trees` is interleaved between two `from queue_backend …` and `from unstract.sdk1 …` third-party-style imports. Ruff/isort typically groups local relative imports last.

**Fix:** move the relative import below the `unstract.sdk1` imports, or run the project's import-sort hook.

workers/tests/test_executor_dispatch.py
```diff
@@ -0,0 +1,134 @@
...
+        "execution_source": "tool",
+        "organization_id": "org-1",
+        "request_id": "req-1",
+        "log_events_id": "log-1",
+        "execution_id": "exec-1",
+        "file_execution_id": "file-exec-1",
+        "executor_params": {"adapter_instance_id": "a"},
+    }
+    defaults.update(overrides)
+    return ExecutionContext(**defaults)  # type: ignore[arg-type]
```
muhammad-ali-e: **[MEDIUM — type ignore masks real issue]** `# type: ignore[arg-type]` silences a legitimate mypy complaint about unpacking `dict[str, object]` into `ExecutionContext`'s typed fields. The dataclass takes mixed-type fields and `object` is correctly too narrow.

**Fix:** annotate `defaults: dict[str, Any]` and `**overrides: Any` and drop the `# type: ignore` — `Any` is local to this test factory and is strictly less harmful than a silencing comment.

workers/tests/test_executor_dispatch.py
```diff
@@ -0,0 +1,134 @@
...
+        app = MagicMock(spec=Celery)
+        async_result = MagicMock()
+        async_result.id = "task-1"
+        async_result.get.return_value = {"success": True, "data": {}}
+        app.send_task.return_value = async_result
+        return ExecutionDispatcher(celery_app=app), app
+
+    def test_dispatch_forwards_headers(self):
+        d, app = self._patched_app()
+        headers = {FAIRNESS_HEADER_NAME: {"org_id": "org-1", "workload_type": "etl"}}
```
muhammad-ali-e: **[MEDIUM — invalid enum value in test fixture]** `"workload_type": "etl"` is not a valid `WorkloadType` enum value — `WorkloadType` only defines `"api"` and `"non_api"` (see `workers/queue_backend/fairness.py:18-19`). Same issue at line 67 (`test_dispatch_with_callback_forwards_headers`).

The tests still pass because they round-trip an opaque dict through a mocked `send_task` without enum validation — but the literals are misleading and will rot anyone debugging from them.

**Fix:** use `WorkloadType.NON_API.value` (the test already imports `WorkloadType` on line 12):
```python
headers = {FAIRNESS_HEADER_NAME: {"org_id": "org-1", "workload_type": WorkloadType.NON_API.value}}
```

workers/tests/test_executor_dispatch.py
```diff
@@ -0,0 +1,134 @@
...
+        # Matches the pre-Phase-5.2 send_task call shape exactly —
+        # no ``headers`` kwarg when the caller doesn't pass one.
+        assert "headers" not in app.send_task.call_args.kwargs
+
+    def test_dispatch_async_forwards_headers(self):
+        d, app = self._patched_app()
+        headers = {FAIRNESS_HEADER_NAME: {"org_id": None, "workload_type": "api"}}
+        d.dispatch_async(_make_context(), headers=headers)
+        assert app.send_task.call_args.kwargs["headers"] == headers
+
```
muhammad-ali-e: **[MEDIUM — missing test coverage]** `dispatch` and `dispatch_with_callback` each have explicit "omits headers when None" tests (lines 52 and 71), but `dispatch_async` only has the positive forwarding test (lines 59-63). The conditional `if headers is not None` branch in `dispatcher.py:213-214` is asymmetrically covered — a regression that hardcodes `headers={}` instead of omitting would slip through here.

**Fix:** add a parallel test:
```python
def test_dispatch_async_omits_headers_when_none(self):
    d, app = self._patched_app()
    d.dispatch_async(_make_context())
    assert "headers" not in app.send_task.call_args.kwargs
```

workers/tests/test_executor_dispatch.py
```diff
@@ -0,0 +1,134 @@
...
+        if not isinstance(node, ast.Call):
+            continue
+        func = node.func
+        if not (isinstance(func, ast.Attribute) and func.attr == "send_task"):
+            continue
+        # First positional arg must be the task name string.
+        if not node.args:
+            continue
+        first = node.args[0]
+        if isinstance(first, ast.Constant) and first.value == "execute_extraction":
```
muhammad-ali-e, line 226: **[MEDIUM — canary scope narrower than its docstring claims]** The detector only matches `*.send_task("execute_extraction", ...)` where the task name is a literal `ast.Constant`. A consumer using `TASK_NAME = "execute_extraction"; app.send_task(TASK_NAME, ...)` (exactly how `dispatcher.py` itself writes it via `_TASK_NAME`), an f-string, `TaskName.EXECUTE_EXTRACTION.value`, or `apply_async` instead of `send_task` walks straight through.

The class docstring at line 97 claims to catch any "raw `*.send_task(\"execute_extraction\", ...)` elsewhere" — that's stronger than the implementation guarantees.

**Fix (one of):**
- Tighten the docstring: "raw **string-literal** dispatch of `execute_extraction`".
- Or widen detection: also resolve simple module-level `Name` references whose value is a string literal, and walk `apply_async` calls too.

At minimum, document the known blind spots in the assertion message so future authors don't trust the canary absolutely.

muhammad-ali-e (commented):
Automated PR review — six specialised review agents (code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier) ran against this branch's 6-file scope. Inline comments below are deduplicated and prioritised. **HIGH/MEDIUM** items are worth resolving before merge; **LOW/NIT** items are optional polish.

workers/tests/test_executor_dispatch.py
```diff
@@ -0,0 +1,134 @@
...
+        fairness = FairnessKey(
+            org_id="org-1", workload_type=WorkloadType.NON_API, pipeline_priority=5
+        )
+        assert fairness.as_header() == {
+            "x-fairness-key": {
+                "org_id": "org-1",
+                "workload_type": "non_api",
+                "pipeline_priority": 5,
+            }
+        }
```
muhammad-ali-e, line 140: **[NIT — coverage gap]** `TestFairnessKeyComposesWithHeaders.test_fairness_header_shape` only exercises `org_id="org-1"`. Add a parametrized case for `org_id=None` (valid per `to_dict()`'s `str | None` typing) so downstream JSON-serialisation assumptions about `None` org IDs are pinned:

```python
assert FairnessKey(org_id=None, workload_type=WorkloadType.API).as_header() == {
    "x-fairness-key": {
        "org_id": None, "workload_type": "api", "pipeline_priority": 5,
    }
}
```

muhammad-ali-e:
## Review feedback addressed — \`d6300bc0\`

14/14 actioned. One-to-one accounting:

### HIGH (4/4)

| # | File:Line | Status |
|---|---|---|
| 1 | `dispatcher.py:224` — `dispatch_with_callback` Args missing `headers:` | ✅ Added the line, mirroring `dispatch_async`'s entry |
| 2 | `structure_tool_task.py:43` — return type widening | ✅ Narrowed to `dict[str, dict[str, str \| int \| None]]` (matches `as_header()`) |
| 3 | `structure_tool_task.py:49` — "Phase 6 work" with no TODO anchor | ✅ Replaced with `TODO(UN-3504): propagate caller's WorkloadType...` — grep-able + ticket-anchored |
| 4 | `canary_helpers.py:35` — silent SyntaxError swallow | ✅ Real silent-failure bug; now emits `UserWarning` so the regression surfaces. Tests can promote via `warnings.simplefilter("error", UserWarning)`. **Both** the fairness canary and the execute-extraction canary inherit the fix automatically since they share this helper. |

### MEDIUM (5/5)

| # | File:Line | Status |
|---|---|---|
| 5 | `dispatcher.py:147` — empty-dict footgun + 3 copies of send_kwargs block | ✅ Extracted `_build_send_kwargs(context, queue, *, headers=None, link=None, link_error=None, task_id=None)` static helper. All three dispatch methods now share one source of truth (kills ~25 LOC of duplication). Empty-dict semantic documented in `dispatch()`'s `headers:` Args block as "caller responsible for the header schema; `{}` is forwarded as-is and likely indicates a producer-side build bug". |
| 6 | `test_executor_dispatch.py:32` — `# type: ignore[arg-type]` | ✅ Switched to `dict[str, Any]` + `**overrides: Any`; dropped the silencer |
| 7 | `test_executor_dispatch.py:48,67` — invalid `"etl"` literal | ✅ Replaced with `WorkloadType.NON_API.value` (and `WorkloadType.API.value` where appropriate). Renames now propagate |
| 8 | `test_executor_dispatch.py:64` — missing async "omits when None" test | ✅ Added `test_dispatch_async_omits_headers_when_none` |
| 9 | `test_executor_dispatch.py:128` — canary blind spots vs docstring | ✅ Tightened docstring to "raw **string-literal** dispatch" + listed the known blind spots (constant refs, f-strings, `apply_async`) in the assertion message. Closed via docstring honesty rather than AST resolution widening — the latter is over-engineering for what's a 1-line dispatcher seam |

### LOW / NIT (5/5)

| # | File:Line | Status |
|---|---|---|
| 10 | `dispatcher.py:117` — bare string instead of symbol | ✅ Docstring now references `queue_backend.fairness.FAIRNESS_HEADER_NAME` symbol with `"x-fairness-key"` in parens for grep continuity. sdk1 can't import the constant (no workers dep), but naming the symbol survives renames. |
| 11 | `fairness.py:58` — vague "wire-ready" | ✅ Docstring now states the concrete shape `{FAIRNESS_HEADER_NAME: self.to_dict()}` |
| 12 | `test_executor_dispatch.py:1,55` — plan-stage vocab | ✅ Module docstring rewritten to describe what tests cover; L55 comment no longer says "pre-Phase-5.2" |
| 13 | `test_executor_dispatch.py:13` — import order | ✅ Relative `from .canary_helpers` moved below `unstract.sdk1` block |
| 14 | `test_executor_dispatch.py:93` — missing `org_id=None` case | ✅ Added `test_fairness_header_shape_orgless` pinning the JSON-null serialisation contract |

### Tests

* `workers/tests/test_executor_dispatch.py`: 7 → 9 (added the two missing coverage tests)
* `workers/tests/test_fairness_key.py`: 22 (unchanged)
* `workers/tests/test_queue_backend_seam.py`: 17 (unchanged)
* `workers/tests/test_dispatch_sites_characterisation.py`: 14 (unchanged)
* `unstract/sdk1/tests/test_execution.py`: 80 (unchanged — `_build_send_kwargs` refactor preserves call-shape semantics)
* **Workers seam suite total: 62 (was 60). sdk1: 80. All green.**

### Notable structural changes

- `_build_send_kwargs` in dispatcher.py is the second time this rollout introduces a private factoring (the first was `_fairness_headers` in structure_tool_task.py). Both serve the same principle: one place to audit the optional-kwarg contract.
- The canary-helpers SyntaxError fix is the kind of silent-failure I'd want banked as a pattern for future canary code. The reviewer's framing — "the docstring acknowledges the bug as justification" — was exactly right.

PR head now: `d6300bc0`.

workers/tests/canary_helpers.py
```diff
@@ -0,0 +1,47 @@
...
+WORKERS_ROOT = pathlib.Path(__file__).parent.parent
+
+# Always-skip top-level directories. Callers pass a wider set if they
+# need to additionally exclude e.g. the seam module itself.
+DEFAULT_SKIP_TOP_DIRS = frozenset({"tests", "__pycache__", "htmlcov", ".venv"})
+
+
+def iter_production_trees(
+    skip_top_dirs: frozenset[str] = DEFAULT_SKIP_TOP_DIRS,
+) -> list[tuple[pathlib.Path, ast.AST]]:
```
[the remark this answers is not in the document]
muhammad-ali-e: Valid P2, fixed in `524ae9184`. The function builds a list and returns it — it isn't a generator — so opening the docstring with "Yield ..." would mislead anyone expecting lazy consumption (early break, generator-specific behaviour). Now reads:

> Return a list of `(rel_path, parsed_tree)` tuples for every .py file outside the skip set.

Pure docstring change — no behaviour shift. The 31 fairness + executor-dispatch tests still pass.

workers/tests/canary_helpers.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    out: list[tuple[pathlib.Path, ast.AST]] = []
+    for py in WORKERS_ROOT.rglob("*.py"):
+        rel = py.relative_to(WORKERS_ROOT)
+        if rel.parts and rel.parts[0] in skip_top_dirs:
+            continue
+        try:
+            tree = ast.parse(py.read_text(), filename=str(py))
+        except SyntaxError as exc:
+            warnings.warn(
```
vishnuszipstack, line 45: **[IMPORTANT · Silent-Failure + Test]** Vacuous-pass risk. On `SyntaxError` this warns then `continue`s, but **neither caller** promotes the warning to an error: `test_executor_dispatch.py:147` and `test_fairness_key.py` both call `iter_production_trees()` with no `simplefilter("error", UserWarning)`. Under pytest's defaults a `UserWarning` does **not** fail the run, so an unparseable production file (e.g. botched merge in `structure_tool_task.py` itself) is silently dropped from the audited set and every canary passes over a smaller tree — exactly the failure mode the docstring says it prevents.

**Fix:** make it fail loud by default — either collect parse failures and have callers assert the list is empty, or add `pytestmark = pytest.mark.filterwarnings("error::UserWarning")` to both canary modules. Also add a test that writes a deliberately broken `.py` and asserts `pytest.warns(UserWarning, match="unparseable")`.

vishnuszipstack (commented):

Ran six specialised review agents — Comment Analyzer, PR Test Analyzer, Silent Failure Hunter, Type Design Analyzer, Code Reviewer, Code Simplifier — against `origin/main...HEAD`. **No CRITICAL correctness defects.** The change is well-constructed and tests pass (31/31). Inline comments below, prioritised.

**Highest-value items:** (1) the canary `UserWarning` is not enforced as an error by either caller, so an unparseable file silently shrinks the audited tree and lets canaries pass vacuously; (2) `_fairness_headers` (the producer) has no test asserting the `NON_API` default or that the 3 dispatch sites attach the header.

One finding can't be inlined because it's outside the diff: **`workers/queue_backend/dispatch.py:41`** hand-builds `{FAIRNESS_HEADER_NAME: fairness.to_dict()}`, an exact duplicate of the new `FairnessKey.as_header()` — refactor it to `fairness.as_header()` so the wire format has a single encoding (see comment on `fairness.py:57`).

workers/tests/canary_helpers.py
```diff
@@ -0,0 +1,48 @@
+"""Shared AST-walk helpers for inventory-style canary tests.
+
+Several test files audit the production code tree (forbid a raw
+dispatch, require a kwarg, etc.). The file-walking + parse-with-skip
+logic is identical across them; this module is the single home.
```
vishnuszipstack: **[SUGGESTION · Comment-Analyzer]** "this module is the single home" overstates current adoption. Two other production-tree walkers still inline the same logic — `workers/tests/test_dispatch_sites_characterisation.py` and `workers/tests/test_chord_sites_characterisation.py` (the latter twice). Only `test_executor_dispatch.py` and `test_fairness_key.py` consume the helper. Soften to "…the intended single home" or file a follow-up to migrate the two characterisation tests, otherwise this reads as comment rot to anyone who greps.

workers/file_processing/structure_tool_task.py
```diff
@@ -39,6 +40,26 @@
...
+    default (API traffic preempting is a strictly weaker mistake than
+    the inverse).
+
+    TODO(UN-3504): propagate the caller's WorkloadType (API vs ETL)
+    instead of hard-coding ``NON_API``. Requires the chord-lift work
+    so the workflow type flows from the chord caller down to here.
+    """
+    return FairnessKey(
+        org_id=organization_id,
+        workload_type=WorkloadType.NON_API,
```
vishnuszipstack, line 59: **[IMPORTANT · Test-Analyzer]** `_fairness_headers` is entirely untested. No test imports `structure_tool_task` to assert this helper returns `non_api`, and the existing `test_sanity_phase5.py::TestStructureToolSingleDispatch` mocks the dispatcher but asserts **nothing** about the `headers=` kwarg. A regression flipping this to `WorkloadType.API` (the docstring's explicitly "strictly worse" mistake) or dropping `headers=` at any of the 3 call sites (`:492`, `:535`, `:752`) would stay green.

**Fix:** add `assert _fairness_headers("org-1") == {"x-fairness-key": {"org_id": "org-1", "workload_type": "non_api", "pipeline_priority": 5}}`, and extend the phase5 test to assert `dispatch.call_args.kwargs["headers"]` carries `non_api`. (The hard-coded `NON_API` default itself is fine — documented, with `TODO(UN-3504)`.)

workers/tests/test_executor_dispatch.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def _patched_app(self) -> tuple[ExecutionDispatcher, MagicMock]:
+        app = MagicMock(spec=Celery)
+        async_result = MagicMock()
+        async_result.id = "task-1"
+        async_result.get.return_value = {"success": True, "data": {}}
+        app.send_task.return_value = async_result
+        return ExecutionDispatcher(celery_app=app), app
+
+    def test_dispatch_forwards_headers(self):
```
vishnuszipstack: **[IMPORTANT · Code-Simplifier]** These six tests (`*_forwards_headers` ×3, `*_omits_headers_when_none` ×3, lines 50–99) are structurally identical bar the method name. Collapse to two `@pytest.mark.parametrize("method", ["dispatch", "dispatch_async", "dispatch_with_callback"])` tests using `getattr(d, method)(...)` — ~60 lines → ~14, with per-method failure granularity preserved via parametrize IDs. (`dispatch` reads `timeout` from env but the mock returns immediately, so `timeout` need not be threaded.)

workers/tests/test_executor_dispatch.py
```diff
@@ -0,0 +1,178 @@
...
+        d, app = self._patched_app()
+        headers = {
+            FAIRNESS_HEADER_NAME: {
```
vishnuszipstack: **[SUGGESTION · Code-Reviewer]** These hand-built header fixtures omit `pipeline_priority`, whereas the real producer (`FairnessKey.as_header()` → `to_dict()`) always includes it. Harmless here (the dispatcher is payload-agnostic and these assert forwarding equality only), but the forwarding tests exercise a header shape no producer emits. Consider `FairnessKey(...).as_header()` as the fixture for consistency.

workers/tests/test_executor_dispatch.py
```diff
@@ -0,0 +1,178 @@
...
+        }
+        d.dispatch_async(_make_context(), headers=headers)
+        assert app.send_task.call_args.kwargs["headers"] == headers
+
+    def test_dispatch_async_omits_headers_when_none(self):
+        d, app = self._patched_app()
+        d.dispatch_async(_make_context())
+        assert "headers" not in app.send_task.call_args.kwargs
+
+    def test_dispatch_with_callback_forwards_headers(self):
```
vishnuszipstack: **[SUGGESTION · Test-Analyzer]** The refactor moved `link`/`link_error`/`task_id` assembly into the shared `_build_send_kwargs`, but no test passes `headers` **together with** `on_success`/`on_error`/`task_id` and asserts all four land in the same `send_task` call. A key-merge regression in `_build_send_kwargs` would slip through. Add one `dispatch_with_callback(ctx, on_success=sig, on_error=sig2, task_id="t", headers=h)` test asserting all four keys are present.

workers/tests/test_executor_dispatch.py
```diff
@@ -0,0 +1,178 @@
...
+    Known blind spots (deliberate — widening adds AST resolution cost
+    for low real-world risk on a 1-line dispatcher seam):
+    * constant references (``T = "execute_extraction"; send_task(T, ...)``)
+    * f-strings
+    * ``apply_async`` calls
+    These are documented in the assertion message so future authors
+    don't trust the canary absolutely.
+    """
+
+    def test_no_raw_execute_extraction_dispatch_outside_dispatcher(self):
```
vishnuszipstack, line 169: **[SUGGESTION · Test-Analyzer]** The canary only ever asserts the empty (passing) case against the live tree — there's no test proving it actually *fails* on a real offender. Add a positive-detection unit test feeding `_raw_execute_extraction_calls` an `ast.parse` of a known-bad snippet (`app.send_task("execute_extraction", ...)`) and asserting the line is returned. This also locks in the documented blind spots (constant ref / f-string / `apply_async`) as explicit non-detections.

unstract/sdk1/src/unstract/sdk1/execution/dispatcher.py
```diff
@@ -99,10 +99,42 @@ def _get_queue(executor_name: str) -> str:
...
+        Optional kwargs are dropped when ``None`` so the resulting call
+        shape matches the minimum-arity form Celery expects for ``link``
+        / ``link_error`` handling and so callers that pass no headers
+        see the same on-wire call as before headers existed.
+        """
+        out: dict[str, Any] = {
+            "args": [context.to_dict()],
+            "queue": queue,
+        }
+        if headers is not None:
```
vishnuszipstack: **[SUGGESTION · Type-Design + Silent-Failure]** Comment/behaviour mismatch: the `dispatch` docstring says `{}` "likely indicates a producer-side build bug", but `if headers is not None` forwards an empty dict as-is. Since `FairnessKey.as_header()` can never legitimately return `{}`, either enforce it (treat falsy as drop: `if headers:`) so a miswired producer surfaces immediately, or `logger.warning` on empty, or soften the docstring if `{}` is a legitimate "explicitly no headers" signal. Right now the code documents the bug rather than catching it.

unstract/sdk1/src/unstract/sdk1/execution/dispatcher.py
```diff
@@ -99,10 +99,42 @@ def _get_queue(executor_name: str) -> str:
         """
         return f"{_QUEUE_PREFIX}{executor_name}"
 
+    @staticmethod
+    def _build_send_kwargs(
+        context: ExecutionContext,
+        queue: str,
+        *,
+        headers: dict[str, Any] | None = None,
```
vishnuszipstack, line 107: **[SUGGESTION · Type-Design]** `dict[str, Any]` is maximally loose for what is semantically one specific header — a typo'd key or wrong-typed value type-checks fine and produces a wrong-but-well-formed Celery message. Using `dict[str, Any]` is *defensible here* because SDK1 cannot import `queue_backend` (the `FairnessKey`/`TypedDict` owner — see `dispatcher.py:31-33`). A generic SDK1-owned alias like `CeleryHeaders = Mapping[str, object]` would narrow "anything" to "a header mapping" without crossing the boundary. Low priority; the structural typing correctly belongs to the producer.

workers/queue_backend/fairness.py
```diff
@@ -53,3 +53,10 @@ def to_dict(self) -> dict[str, str | int | None]:
             "workload_type": self.workload_type.value,
             "pipeline_priority": self.pipeline_priority,
         }
+
+    def as_header(self) -> dict[str, dict[str, str | int | None]]:
```
vishnuszipstack, line 57: **[SUGGESTION · Type-Design]** Two notes on `as_header()`:

1. **Duplicate not migrated (the more important one):** `workers/queue_backend/dispatch.py:41` still hand-builds `{FAIRNESS_HEADER_NAME: fairness.to_dict()}` — the exact body this method centralises. Replace with `fairness.as_header()` so the wire format has a single encoding and the two can't drift. *(Can't inline there — outside this PR's diff.)*

2. **Loose return type:** `dict[str, dict[str, str | int | None]]` is structurally honest but opaque — it admits `org_id=int` / `pipeline_priority="api"`. A `TypedDict` for the inner payload (`org_id: str | None; workload_type: str; pipeline_priority: int`) would tighten the union and self-document at zero runtime cost. The strong invariant (priority bounds, validated in `__post_init__`) is correctly inherited via frozen construction — good. This same annotation is hand-duplicated at `structure_tool_task.py:45`; a shared alias would keep them in sync.

workers/tests/test_executor_dispatch.py
muhammad-ali-e: **Fixed in `b1c61c46`** — the six `*_forwards_headers ×3` + `*_omits_headers_when_none ×3` tests are now three parametrized methods:

```python
@pytest.mark.parametrize("method", ["dispatch", "dispatch_async", "dispatch_with_callback"])
def test_forwards_headers(self, method): ...

@pytest.mark.parametrize("method", [...])
def test_omits_headers_when_none(self, method): ...

@pytest.mark.parametrize("method", [...])
def test_omits_headers_when_empty_dict(self, method): ...  # picks up #9's empty-dict drop
```

Net ~60 LOC → ~30, with per-method failure granularity preserved via the parametrize IDs (`test_forwards_headers[dispatch_async]` etc.). `timeout` threading wasn't needed — `dispatch` reads it from env when omitted, and the mocked `send_task` returns immediately so no wait happens.

workers/file_processing/structure_tool_task.py
muhammad-ali-e: **Fixed in `b1c61c46`** on both axes you flagged:

**Direct unit tests** — new `workers/tests/test_structure_tool_task.py`:

* `test_returns_non_api_wire_shape` — asserts the exact wire envelope `{"x-fairness-key": {"org_id": "org-1", "workload_type": "non_api", "pipeline_priority": 5}}`.
* `test_org_id_is_propagated_verbatim`.
* `test_workload_type_is_non_api_not_api` — pinned both the positive (`== WorkloadType.NON_API.value`) and the negative (`!= WorkloadType.API.value`) so a flip surfaces with a readable failure.

**Integration assertion** — extended `test_sanity_phase5.TestStructureToolSingleDispatch.test_single_dispatch_normal` to assert `dispatcher.dispatch.call_args.kwargs["headers"]` carries the expected `x-fairness-key` envelope. A regression dropping `headers=` at any of the 3 call sites or flipping `NON_API`→`API` now fails one of these tests.

workers/tests/canary_helpers.py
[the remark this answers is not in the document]
muhammad-ali-e: **Fixed in `524ae9184`** (predates this round). Docstring now reads `Return a list of (rel_path, parsed_tree) tuples for every .py file outside the skip set` — function is no longer mis-described as a generator. Thanks for the catch.

workers/tests/canary_helpers.py
muhammad-ali-e: **Fixed in `b1c61c46`** — both canary modules now carry `pytestmark = pytest.mark.filterwarnings("error::UserWarning")`:

* `workers/tests/test_executor_dispatch.py`
* `workers/tests/test_fairness_key.py`

Plus a new `workers/tests/test_canary_helpers.py` with two focused unit tests:

* `test_unparseable_file_emits_userwarning` — drops a deliberately broken `.py` into an isolated tree, asserts both that the good file makes the audit set AND that the bad file triggers a `UserWarning` mentioning `unparseable` + the filename.
* `test_warning_promoted_to_error_under_filterwarnings` — locks the contract the canary modules rely on: when `simplefilter("error", UserWarning)` is installed (the equivalent of the `pytestmark` above), the helper raises instead of silently skipping.

A botched merge in `structure_tool_task.py` or anywhere else in the audited tree now fails loud at canary collection time.

workers/tests/canary_helpers.py
muhammad-ali-e: **Fixed in `b1c61c46`** — docstring softened to `intended single home` and the two pre-existing characterisation walkers (`test_dispatch_sites_characterisation.py`, `test_chord_sites_characterisation.py`) are explicitly named in the docstring as a follow-up. Migrating them is out of scope for this PR but tracked.

```
this module is the intended single home. Two pre-existing
characterisation tests still inline the same walk
(test_dispatch_sites_characterisation.py, test_chord_sites_characterisation.py)
— migrating those is tracked as a follow-up.
```

unstract/sdk1/src/unstract/sdk1/execution/dispatcher.py
muhammad-ali-e: **Fixed in `b1c61c46`** — chose the catch-the-bug option you suggested. `_build_send_kwargs` now treats falsy as drop:

```python
# Treat falsy (None, {}) as drop. An empty header dict is
# never legitimate — FairnessKey.as_header() always returns
# a populated mapping — so silently forwarding {} would
# document a producer-side build bug rather than catch it.
if headers:
    out["headers"] = headers
```

And the docstring was rewritten to describe the new contract:

```
Falsy values (None, {}) are dropped — the downstream send_task
call is invoked without headers= so the on-wire shape matches
the no-headers baseline.
```

The parametrize collapse for #5 also added `test_omits_headers_when_empty_dict[...]` over all three dispatch entry points so a revert to `is not None` is caught loud.

workers/queue_backend/fairness.py
muhammad-ali-e: **Sub-finding (1) fixed in `b1c61c46`** — `workers/queue_backend/dispatch.py:41` no longer hand-builds the header:

```python
# before
headers = {FAIRNESS_HEADER_NAME: fairness.to_dict()} if fairness is not None else None
# after
headers = fairness.as_header() if fairness is not None else None
```

Wire-format encoding now has a single source — the two sites can't drift. `FAIRNESS_HEADER_NAME` import dropped from this module since it's no longer referenced here. Thanks for explicitly flagging this as the more important of your two notes — easy to miss inside an `outside this PR's diff` parenthetical.

**Sub-finding (2) deferred** — the TypedDict refactor of `as_header()`'s return type is a worthwhile tightening but affects two annotation sites (this file plus `structure_tool_task.py:45`) and touches the labs L2 schema vocabulary. Worth its own PR with the broader sdk1 typing pass — bundling it here would expand scope past UN-3508's chord-callback typing focus. Filed as a follow-up.

unstract/sdk1/src/unstract/sdk1/execution/dispatcher.py
muhammad-ali-e: **Deferred — agree with your assessment.** You explicitly noted this is "low priority" and "defensible here" because SDK1 cannot import `queue_backend` (the TypedDict owner). A generic SDK1-owned `CeleryHeaders = Mapping[str, object]` alias would be a nice future tightening, but it doesn't justify a separate PR slot for UN-3508's chord-callback typing work. Filing as a follow-up under the broader sdk1 typing tickets.

workers/tests/test_executor_dispatch.py
muhammad-ali-e: **Fixed in `b1c61c46`** — new `test_dispatch_with_callback_combines_headers_and_callbacks`:

```python
d.dispatch_with_callback(
    _make_context(),
    on_success=on_success,
    on_error=on_error,
    task_id="t-1",
    headers=headers,
)
kwargs = app.send_task.call_args.kwargs
assert kwargs["headers"] == headers
assert kwargs["link"] is on_success
assert kwargs["link_error"] is on_error
assert kwargs["task_id"] == "t-1"
```

A key-merge regression in `_build_send_kwargs` (e.g. one branch shadowing another) would now fail this test, where it would have silently slipped through the single-kwarg forwarding tests above.

workers/tests/test_executor_dispatch.py
muhammad-ali-e: **Fixed in passing in `b1c61c46`** — the parametrize refactor for #5 swapped the hand-built dict fixture for `FairnessKey(org_id="org-1", workload_type=WorkloadType.NON_API).as_header()`, so the forwarding tests now exercise the exact wire shape real producers emit, `pipeline_priority` included.

You noted this is "harmless here" — agreed it wouldn't have caused a runtime issue — but since the refactor was already touching these tests it was free to fix at the same time.

workers/tests/test_executor_dispatch.py
muhammad-ali-e: **Fixed in `b1c61c46`** — two new tests against `_raw_execute_extraction_calls` directly:

* `test_detector_matches_string_literal_send_task` — feeds `ast.parse('app.send_task("execute_extraction", args=[])')` and asserts `[1]` is returned. Without this, an `ast.walk` regression (e.g. matching nothing) would leave the live-tree canary always green regardless of what production code does.
* `test_detector_skips_documented_blind_spots` — locks the three documented blind spots (constant ref, f-string, `apply_async`) as explicit non-detections. A future widening of the detector that intentionally trips one of these has to update the assertion — surfacing the scope change.

The canary now has both "finds known offenders" and "deliberately ignores documented patterns" coverage.

vishnuszipstack (approved):

johnyrahul (approved):

PR #2019, opened 2026-06-08
fix(llama-parse): rename schema field url to base_url (#1972)

### LINKED ISSUE #1972: LlamaParse adapter: `url` field in json_schema.json doesn't match `base_url` config key, breaking EU region support
## Bug Description
The LlamaParse x2text adapter has a mismatch between the UI schema field name 
and the config key used in the adapter code, causing the base URL to always 
fall back to the hardcoded US endpoint regardless of what the user enters.

## Files affected
- `unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/static/json_schema.json`
  — defines the UI field as `"url"`
- `unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/constants.py`
  — defines `BASE_URL = "base_url"`
- `unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/llama_parse.py`
  — reads `self.config.get(LlamaParseConfig.BASE_URL)` which maps to `"base_url"`

## Impact
Users with EU region LlamaCloud accounts cannot use LlamaParse — their API key 
is sent to `https://api.cloud.llamaindex.ai` (US) instead of 
`https://api.cloud.eu.llamaindex.ai` (EU), resulting in a 401 Unauthorized error.

## Fix
In `json_schema.json`, rename the field from `"url"` to `"url"` → `"base_url"`:

## Version
latest

jaseemjaskp:
## What

- Rename the LlamaParse x2text adapter UI schema field from `url` to `base_url` in `json_schema.json`.

## Why

- The schema defined the base-URL field as `url`, but the adapter code reads it via `LlamaParseConfig.BASE_URL` (`"base_url"`). The mismatch meant the value a user entered was stored under `url` and never read, so `LlamaParse` always fell back to its hardcoded US endpoint (`https://api.cloud.llamaindex.ai`).
- As a result, EU-region LlamaCloud users could not use LlamaParse — their key was sent to the US endpoint, returning `401 Unauthorized`.

## How

- Changed the property key `"url"` → `"base_url"` in `unstract/sdk1/.../llama_parse/src/static/json_schema.json` so it matches `constants.py` and the lookup in `llama_parse.py`.

- No new breakage. This is a forward fix for new adapter instances.
- Note on existing instances: any LlamaParse adapter already saved before this fix has its value persisted under the old `url` key (configs are stored Fernet-encrypted in `adapter_instance.adapter_metadata_b`), so it will not auto-migrate. Those few users simply need to re-open and re-save the adapter to populate `base_url`. A data migration was considered but intentionally skipped to keep the change minimal.

## Database Migrations

- None.

## Env Config

- None.

## Relevant Docs

- N/A

## Related Issues or PRs

- Fixes #1972

## Dependencies Versions

- N/A

## Notes on Testing

- Validated the JSON and ran the repo pre-commit JSON/whitespace hooks (all pass).
- Manual verification: configure a LlamaParse adapter with the EU base URL (`https://api.cloud.eu.llamaindex.ai`) and confirm extraction succeeds instead of returning 401.

## Checklist

### DIFF
--- a/unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/static/json_schema.json
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/static/json_schema.json
@@ -18,7 +18,7 @@
             "default": "",
             "description": "Provide the token (API Key) of the Llama Parse server"
         },
-      "url": {
+      "base_url": {
         "type": "string",
         "title": "URL",
         "format": "url",
--- a/unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/static/json_schema.json
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/static/json_schema.json
@@ -20,8 +20,8 @@
         },
       "base_url": {
         "type": "string",
-        "title": "URL",
-        "format": "url",
+        "title": "Base URL",
+        "format": "uri",
         "default": "https://api.cloud.llamaindex.ai",
         "description": "Provide the Base URL of llama Parse server."
       },
--- a/unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/constants.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/constants.py
@@ -3,6 +3,9 @@ class LlamaParseConfig:
 
     API_KEY = "api_key"
     BASE_URL = "base_url"
+    # Legacy schema key for the base URL, kept for backward compatibility with
+    # adapter instances saved before the rename (#1972).
+    LEGACY_BASE_URL = "url"
     RESULT_TYPE = "result_type"
     NUM_WORKERS = "num_workers"
     VERBOSE = "verbose"
--- a/unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/llama_parse.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/llama_parse.py
@@ -52,7 +52,8 @@ class LlamaParseAdapter(X2TextAdapter):
             fs = FileStorage(provider=FileStorageProvider.LOCAL)
         parser = LlamaParse(
             api_key=self.config.get(LlamaParseConfig.API_KEY),
-            base_url=self.config.get(LlamaParseConfig.BASE_URL),
+            base_url=self.config.get(LlamaParseConfig.BASE_URL)
+            or self.config.get(LlamaParseConfig.LEGACY_BASE_URL),
             result_type=self.config.get(LlamaParseConfig.RESULT_TYPE),
             verbose=self.config.get(LlamaParseConfig.VERBOSE),
             language="en",


jaseemjaskp (commented):
## PR Review Toolkit — summary (6 agents: Code Reviewer, Silent Failure Hunter, Type Design, PR Test Analyzer, Comment Analyzer, Code Simplifier)

**Verdict: the one-line rename is correct and fixes a real bug.** Before this PR the schema published the key `url` while the adapter reads `base_url` (`constants.py:5` → `LlamaParseConfig.BASE_URL = "base_url"`, consumed at `llama_parse.py:55`). So the configured Base URL never reached the adapter — it resolved to `None` and the SDK silently fell back to its default endpoint. This PR aligns the two. Blast radius is fully contained: `LlamaParseConfig` is referenced only within `llama_parse.py`; no frontend/backend code reads the literal `url` key.

### Prioritised findings

**P1 — Backward-compat / no migration (llama_parse.py:55, outside diff).** Adapter instances saved before this change persist the value under the old `url` key and have no `base_url`, so `config.get("base_url")` returns `None` for them. Cloud users are unaffected (None → default cloud endpoint), but **self-hosted / custom-endpoint users silently revert** to `https://api.cloud.llamaindex.ai` with no log. Fix options: (a) data migration renaming the persisted key, or (b) legacy fallback: `self.config.get(LlamaParseConfig.BASE_URL) or self.config.get("url")`. Moot if this adapter has no released configs yet — please confirm. (Inline note on schema line 21.)

**P2 — Pre-existing broad `except Exception` mislabels errors (llama_parse.py:94-98, outside diff).** Any non-`ConnectError` (401/403/404 from a wrong base_url, etc.) is hard-coded as "invalid API Key", and line 96 uses a **non-f-string** `"...: {exe}"` so the exception detail is never logged. Not introduced here, but the rename makes these failures more likely. Recommend fixing the f-string and not asserting a guessed cause.

**P3 — No test coverage (sdk1/tests).** Zero tests reference LlamaParse or this schema. A cheap, network-free regression test mirroring `test_gemini_embedding.py:31-53` would pin the schema-key↔constant mapping that just drifted: assert `LlamaParseConfig.BASE_URL in schema['properties']`, `'url' not in properties`, `required == {'api_key'}`. Note: only assert the subset of constants the adapter actually reads from `self.config` (NUM_WORKERS/LANGUAGE are not in the schema). Criticality ~7/10.

**P4 — `title` still reads "URL" (schema line 23, inline).** Field is now `base_url` and description says "Base URL"; suggest `"title": "Base URL"`.

**P5 — Nits (optional):** `"format": "url"` is non-standard JSON Schema — standard is `"uri"` (schema line 24, inline). Description (line 26) mis-cases "llama Parse" and drops the article vs line 19 — suggest "...of the Llama Parse server.". `api_key` block (line 14) is indented 8 spaces vs 6 for siblings. All pre-existing.

**Type Design / Code Simplifier:** schema is well-formed; `base_url` optional-with-default is the right choice (not in `required`). Nothing to simplify in the rename itself. Note llama_parse is now the lone x2text adapter using `base_url` vs `url` in siblings — acceptable since its code requires `base_url`.

_No blocking issues. Recommend addressing P1 (or confirming it's moot) and P4 before merge; P2/P3 are good follow-ups._

unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/static/json_schema.json
```diff
@@ -18,7 +18,7 @@
             "default": "",
             "description": "Provide the token (API Key) of the Llama Parse server"
         },
-      "url": {
+      "base_url": {
```
jaseemjaskp, line 21: ✅ **Correct fix.** This aligns the schema key with the consuming code: `constants.py:5` defines `BASE_URL = "base_url"` and `llama_parse.py:55` reads `self.config.get(LlamaParseConfig.BASE_URL)`. Before this PR the schema emitted `url`, so the configured Base URL never reached the adapter (resolved to `None`; the SDK fell back to its default endpoint).

⚠️ **Backward-compat (P1):** adapter instances saved before this change persist the value under the old `url` key and have no `base_url`, so `config.get("base_url")` returns `None` for them — **self-hosted / custom-endpoint users silently revert** to `https://api.cloud.llamaindex.ai`. Consider a data migration, or a legacy fallback in `llama_parse.py:55`:
```python
base_url = self.config.get(LlamaParseConfig.BASE_URL) or self.config.get("url")
```
Moot if this adapter has no released configs yet.

unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/static/json_schema.json
```diff
@@ -18,7 +18,7 @@
...
         "type": "string",
         "title": "URL",
```
jaseemjaskp: **Nit (P4):** `title` still reads `"URL"` while the field is now `base_url` and the description says "Base URL". Suggest `"title": "Base URL"` so the UI label matches the field intent.

unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/static/json_schema.json
```diff
@@ -18,7 +18,7 @@
...
         "format": "url",
```
jaseemjaskp: **Pre-existing nit (P5, optional):** `"format": "url"` isn't a standard JSON Schema format keyword — the standard is `"uri"`. Likely benign since the form renderer treats this as a plain string, but worth correcting while the block is open.

unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/static/json_schema.json
jaseemjaskp: Addressed in b37059b08. Added a legacy fallback so pre-rename instances keep working:
```python
base_url=self.config.get(LlamaParseConfig.BASE_URL)
or self.config.get(LlamaParseConfig.LEGACY_BASE_URL),
```
`LEGACY_BASE_URL = "url"` is defined in `constants.py`. This avoids a data migration while ensuring self-hosted / custom-endpoint users do not silently revert to the default US endpoint.

unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/static/json_schema.json
jaseemjaskp: Fixed in b37059b08 — `title` is now `"Base URL"`.

unstract/sdk1/src/unstract/sdk1/adapters/x2text/llama_parse/src/static/json_schema.json
jaseemjaskp: Fixed in b37059b08 — `format` is now the standard `"uri"`.

muhammad-ali-e (approved):

PR #2022, opened 2026-06-08
UN-3308 [MISC] Register marketplace landing routes (plugin-gated)

athul-rs:
## What

Registers two frontend routes for the marketplace buyer flow, gated on the cloud-only marketplace plugin:

- `/:orgName/marketplace-landing` → `MarketplaceLandingPage`
- `/:orgName/marketplace-stripe-conflict` → `MarketplaceStripeConflictPage`

The plugin is `import()`ed dynamically inside a `try/catch`; OSS builds that don't ship `src/plugins/marketplace` silently de-register both routes.

## Why

The cloud marketplace FE plugin provides the pages, but routing lives in the OSS app shell (`useMainAppRoutes.js`). AWS Marketplace buyers are redirected to `/marketplace-landing` post-purchase to claim their org; the Stripe-conflict route handles the cross-provider edge case. Keeping the registration plugin-gated means OSS builds are unaffected.

## Why this won't break existing features

Single additive change to `useMainAppRoutes.js`: a lazy `import("../plugins/marketplace")` wrapped in `try/catch`, and two `<Route>`s rendered only when the plugin resolves. In OSS (no plugin) the import throws, both route vars stay undefined, and nothing is registered — identical routing to before. No existing route or import is modified.

## Database Migrations

None.

## Env Config

None.

## Relevant Docs

`docs/marketplace/FRONTEND_SPEC.md` (cloud, backend PR Zipstack/unstract-cloud#1471).

## Related Issues or PRs

- Cloud marketplace FE plugin: Zipstack/unstract-cloud#1532 (provides the pages this registers)
- Backend: Zipstack/unstract-cloud#1471
- Jira: UN-3308

## Dependencies Versions

None.

## Notes on Testing

Verified OSS build de-registers the routes when the plugin is absent (try/catch path). Full flow tested with the cloud plugin present against a `TACKLE_ENABLED=True` backend. A `frontend` build verification is recommended before merge.

## Screenshots

N/A — routing only; pages are screenshotted in Zipstack/unstract-cloud#1532.

## Checklist


athul-rs:
✅ Frontend build verified locally: `vite build` (vite 7.3.1) completed successfully — 1786 modules transformed, full production bundle emitted in ~14s — with the marketplace plugin present in the OSS frontend (copied in, matching the plugin-copy build lifecycle). The plugin's only external deps (`antd`, `react`, `react-router-dom`) all resolve from OSS `node_modules`. The plugin-absent path (standard OSS build) is the established `try/catch` dynamic-import pattern already used by ~10 other cloud plugins, so the routes cleanly de-register when the plugin isn't shipped.

frontend/src/routes/useMainAppRoutes.js
```diff
@@ -168,11 +168,34 @@ try {
...
+let MarketplaceLandingPage;
+let MarketplaceStripeConflictPage;
+try {
+  const mod = await import("../plugins/marketplace");
+  MarketplaceLandingPage = mod.MarketplaceLandingPage;
+  MarketplaceStripeConflictPage = mod.MarketplaceStripeConflictPage;
+} catch {
+  // OSS build does not ship the marketplace plugin; the marketplace
+  // landing routes silently de-register.
+}
```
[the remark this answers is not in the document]
athul-rs: Fixed in 10df585c — adopted the established ReadOnlyReviewPage pattern (this file, ~line 129): silent for the expected missing-module case (OSS build), `console.error` for any other failure so a broken cloud plugin doesn't de-register the routes silently.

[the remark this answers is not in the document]
athul-rs: Fixed in 10df585c — the catch now distinguishes a missing-module error (kept silent for OSS builds) from unexpected failures, which are logged via `console.error`, matching the sibling ReadOnlyReviewPage import in the same file.

frontend/src/routes/useMainAppRoutes.js
```diff
@@ -168,11 +168,48 @@ try {
...
+  MarketplaceLandingPage = mod.MarketplaceLandingPage;
+  MarketplaceStripeConflictPage = mod.MarketplaceStripeConflictPage;
+} catch (err) {
+  // Expected in OSS builds where the cloud plugin is absent. Surface
+  // anything that isn't a missing-module error so syntax/runtime
+  // failures inside the plugin don't silently de-register the routes.
+  const msg = err?.message || "";
+  const isModuleMissing =
+    err?.code === "MODULE_NOT_FOUND" ||
+    msg.includes("Failed to fetch dynamically imported module") ||
```
jaseemjaskp: **[Major] `"Failed to fetch dynamically imported module"` is not a reliable missing-module signal — the classifier can still silently de-register routes on cloud builds.**

This extends the now-resolved thread above (which fixed swallowing *all* non-missing errors). The remaining gap is in the detection itself: this string cannot distinguish a *genuinely absent* plugin from a *present* one whose chunk failed to load. In Vite/Rollup it's the generic chunk-fetch-failure wrapper — transient CDN/origin 5xx, a stale `index.html` after redeploy pointing at a purged hashed chunk, CSP/ad-blocker, or offline all produce it.

In OSS this is harmless (the plugin really is absent). But in cloud/Enterprise builds where `../plugins/marketplace` *is* bundled, a transient chunk-fetch failure is misclassified as `isModuleMissing` and swallowed → the marketplace routes silently vanish with no log, which is exactly the failure class this block was added to prevent.

Consider making plugin presence a build-time decision rather than inferring it from a runtime error string — e.g. gate the `await import()` on a build flag (`import.meta.env.VITE_ENABLE_MARKETPLACE` or similar) and treat *any* catch inside that gate as unexpected. The sibling `ReadOnlyReviewPage` block (lines 129-145) has the identical limitation.

frontend/src/routes/useMainAppRoutes.js
```diff
@@ -168,11 +168,48 @@ try {
...
+  const mod = await import("../plugins/marketplace");
+  MarketplaceLandingPage = mod.MarketplaceLandingPage;
+  MarketplaceStripeConflictPage = mod.MarketplaceStripeConflictPage;
+} catch (err) {
+  // Expected in OSS builds where the cloud plugin is absent. Surface
+  // anything that isn't a missing-module error so syntax/runtime
+  // failures inside the plugin don't silently de-register the routes.
+  const msg = err?.message || "";
+  const isModuleMissing =
+    err?.code === "MODULE_NOT_FOUND" ||
```
jaseemjaskp: **[Nit] `err?.code === "MODULE_NOT_FOUND"` never fires in the shipped browser bundle.** `MODULE_NOT_FOUND` is a Node CJS-loader code; a browser/Vite dynamic `import()` rejection doesn't carry it (it may only appear under jsdom/vitest). Harmless, but worth knowing the runtime classification rests entirely on the two message-substring checks below — this line gives a slightly false impression of robustness.

frontend/src/routes/useMainAppRoutes.js
```diff
@@ -168,11 +168,48 @@ try {
...
+    msg.includes("Failed to fetch dynamically imported module") ||
+    msg.includes("Cannot find module");
```
jaseemjaskp: **[Minor/Refactor] Extract a shared `isModuleMissing(err)` helper to drop the duplication.** This `code`/message triad is now copy-pasted verbatim from the `ReadOnlyReviewPage` block (lines 134-137), comment and all. A single predicate keeps the "what counts as missing" rule (and any future fix for the soundness issue noted above) in one place:

```js
function isModuleMissing(err) {
  const msg = err?.message || "";
  return (
    err?.code === "MODULE_NOT_FOUND" ||
    msg.includes("Failed to fetch dynamically imported module") ||
    msg.includes("Cannot find module")
  );
}
```

Both catch blocks then collapse to `if (!isModuleMissing(err)) { console.error(...) }`. The classifier is also the one piece of real, testable logic in this change — once extracted, a small vitest (mirroring `src/helpers/requestId.test.js`) would pin the missing-vs-broken behavior the comment promises.

jaseemjaskp (commented):
Automated review (PR Review Toolkit: code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier). Scope is the single changed file `frontend/src/routes/useMainAppRoutes.js`. The coarse "silently swallow everything" issue from the earlier threads is correctly fixed in 10df585c; the comments below are new, narrower observations that extend rather than repeat it. Nothing here is merge-blocking.

frontend/src/routes/useMainAppRoutes.js
athul-rs: Addressed in c1f84060 — you're right the signal isn't sound in a browser bundle. Documented the limitation explicitly on the shared `isModuleMissing` helper (absent plugin vs. failed chunk are indistinguishable; a transient chunk-load failure may de-register for that session — same as a bare catch, but present-plugin errors now log). Centralized so a future hardening lands once.

frontend/src/routes/useMainAppRoutes.js
athul-rs: Addressed in c1f84060 — noted in the helper comment that `MODULE_NOT_FOUND` is a Node/CJS code that only appears under jsdom/vitest, not the shipped bundle; classification effectively rests on the message checks.

frontend/src/routes/useMainAppRoutes.js
athul-rs: Done in c1f84060 — extracted a single `isModuleMissing(err)` used by both the marketplace and ReadOnlyReviewPage catch sites (was copy-pasted).

jaseemjaskp (approved):
Re-reviewed via the PR Review Toolkit (code review, silent-failure, type-design, test-coverage, comment, simplifier passes). Scope is one file (`useMainAppRoutes.js`): the `isModuleMissing(err)` extraction, the refactored ReadOnlyReviewPage catch, and the plugin-gated marketplace import + two route registrations.

No findings above the bar. The previously-raised points (silent swallowing of non-missing import failures, the browser-classification soundness limitation, the helper duplication) are all addressed in 10df585c/c1f84060. Verified: React Router ranks the static `marketplace-landing` / `marketplace-stripe-conflict` segments above dynamic siblings (no shadowing), FullPageLayout is the correct chrome-less parent, and the per-export `&&` gating fails soft.

Only sub-bar follow-ups (not blocking): (1) a present-but-partial plugin that loads but is missing one named export would skip that one route with no log — symmetry with the `ReviewLayout` `console.warn` at L166-172 would close it; (2) `isModuleMissing` is pure and worth a small vitest if it is exported. LGTM.

PR #2024, opened 2026-06-09
UN-3522 [FEAT] PG Queue Phase 6a — Barrier interface + CeleryChordBarrier wrapper + lift chord call sites + plumb fairness

muhammad-ali-e:
## Summary

First sub-task of Phase 6 (UN-3504 — Distributed Barrier). Ships the abstraction layer + chord-call-site lift + fairness plumbing with **zero runtime behaviour change** — every operation still routes through `celery.chord(...)` under the hood. The risky behaviour-change PR (`RedisDecrBarrier` replacing chord_unlock polling) is deliberately deferred to Phase 6b (UN-3523) behind a feature flag.

### What this adds

- **`workers/queue_backend/barrier.py`**
  - `Barrier` Protocol — minimum contract every barrier substrate satisfies.
  - `BarrierHandle` Protocol — minimum return-value contract (`id: str`); celery's `AsyncResult` satisfies this.
  - `CeleryChordBarrier` impl — wraps `chord(header)(body)` 1:1. Optional `FairnessKey` stamped as a message header on every enqueued task AND the callback.

- **Both production chord call sites lifted behind the barrier:**
  - `shared/workflow/execution/orchestration_utils.py:67` — `WorkflowOrchestrationUtils.create_chord_execution` now delegates to `CeleryChordBarrier`. Adds optional `fairness=` kwarg.
  - `api-deployment/tasks.py:674` — inline `chord(...)` replaced with the helper call, passing `fairness=FairnessKey(workload_type=API)`.
  - `general/tasks.py:941` — existing call site now passes `fairness=FairnessKey(workload_type=NON_API)`.

- **Inventory canary tightened** in `tests/test_chord_sites_characterisation.py`: was "exactly 2 chord call sites in workers/", now "exactly 1 chord call site, inside `queue_backend/barrier.py`". A future PR that adds a direct `chord(...)` call outside the abstraction fails this test loudly.

### Why plumb fairness here

Phase 5.1 added fairness to bare `dispatch()` call sites only (2 of them today). The two `chord(...)` call sites bypass `dispatch()` entirely — at runtime, **MOST workflow-execution tasks fan out through chord** (process_file_batch × N), so Phase 5.1 only covered ~5% of the workflow-execution wire traffic. This PR closes the remaining ~95%.

Doing it at the barrier lift point (rather than retroinstrumenting the raw chord call) means a future Phase 6b can swap in a `RedisDecrBarrier` or `PgBarrier` without touching call sites a second time.

### What this does NOT change (zero-behaviour-change posture)

- The `chord(header)(body)` Celery primitive itself — same two-step call, same chord_unlock polling, same result backend interaction, same retry semantics, same callback firing.
- Task execution — same workers consume the same queues.
- Result aggregation — chord still collects all header return values into a list and passes them to the body.
- Failure handling — same chord error semantics, same `link_error` path.

### Mixed-version rolling deploys

Safe in both directions:
- Old worker enqueues chord (no fairness header) → new callback receives → callback doesn't read fairness header → works.
- New worker enqueues chord (with fairness header) → old callback receives → celery passes unknown headers through → works.

### Tests

New `tests/test_barrier.py` (12 tests):
- **Barrier Protocol shape** — `CeleryChordBarrier` satisfies the `Barrier` Protocol; AsyncResult-shaped handles satisfy `BarrierHandle`.
- **Wire equivalence with the pre-uplift chord call** — given the same args, `CeleryChordBarrier.enqueue(...)` produces the same `chord(header)(body)` invocation, same `app.signature(...)` args, same return value. Locks the byte-for-byte equivalence claim.
- **Fairness header plumbing** — when `fairness=` is passed, the fairness slot is attached to every enqueued task and the callback. When `fairness=None`, no `headers=` kwarg is added (preserves byte-for-byte pre-uplift wire).
- **Mixin wrapper invariants preserved** — `WorkflowOrchestrationMixin.create_chord` still extracts `self.app` and raises on missing app context.

Updated `tests/test_chord_sites_characterisation.py`:
- Existing 6 characterisation tests still pass — patch targets updated to `queue_backend.barrier.chord`.
- Inventory canary tightened: 2 sites → 1 site (inside `barrier.py`).
- `from celery import chord` canary: 2 files → 1 file.

Updated `tests/test_queue_backend_seam.py::test_all_exports` for the new public surface (`Barrier`, `BarrierHandle`, `CeleryChordBarrier`).

### Validation

- Touched test files: 21 / 21 passing
- Full workers suite: 6 failed (pre-existing baseline, unchanged) / 659 passed (was 627 → +32 new tests)
- Five deterministic-order full-suite runs: exactly 6/659 each time — zero flakiness from this change

### What's deferred

- **Phase 6b (UN-3523):** `RedisDecrBarrier` impl + `WORKER_BARRIER_BACKEND` env flag (default `chord` — opt-in). New code path gated.
- **Phase 6c (UN-3524):** production rollout / verification (config-only, no code changes).

### Risk

- Wire shape change: each chord member signature + callback signature now carries a fairness header (~80 bytes / task). Additive — celery passes unknown headers through; no consumer reads the slot yet.
- All other behaviour identical to direct `chord(...)` calls.
- Mixed-version rolling deploy safe (verified above).
- Rollback: a single revert commit removes the abstraction; both call sites are one-line delegations.

## Test plan

- [x] CI green (unit + integration + SonarCloud + CodeRabbit + Greptile)
- [x] Local: `cd workers && uv run pytest tests/test_barrier.py tests/test_chord_sites_characterisation.py` → 21 passed
- [x] Local: `cd workers && uv run pytest --no-cov -p no:randomly` → 6 pre-existing failures, 659 passed
- [x] Dev stack smoke test: trigger one ETL and one API deployment, verify completion times match pre-uplift baseline (no regression in chord_unlock retry pattern)
- [x] Inspect a chord member message in production: confirm `x-fairness-key` header is present and well-formed

### DIFF
--- a/workers/api-deployment/tasks.py
+++ b/workers/api-deployment/tasks.py
@@ -704,7 +704,8 @@ def _run_workflow_api(
             #    workflow_execution_status, API result cache, and
             #    pipeline notifications all run exactly as they would
             #    have pre-Barrier. (Unreachable in practice — upstream
-            #    requires non-empty ``created_files`` — but the defence
+            #    requires non-empty ``hash_values_of_files`` — but
+            #    the defence
             #    closes the theoretical gap.)
             # 2. ``batch_tasks`` is non-empty but the barrier returned
             #    falsy for some other reason — a genuine queue failure.
--- a/workers/api-deployment/tasks.py
+++ b/workers/api-deployment/tasks.py
@@ -7,7 +7,8 @@ Uses the same patterns as workflow_helper.py and file_execution_tasks.py
 import time
 from typing import Any
 
-from queue_backend import worker_task
+from queue_backend import FairnessKey, worker_task
+from queue_backend.fairness import WorkloadType
 from shared.api import InternalAPIClient
 from shared.enums.status_enums import PipelineStatus
 from shared.enums.task_enums import TaskName
@@ -668,19 +669,32 @@ def _run_workflow_api(
         # Create callback queue using same logic as Django backend
         file_processing_callback_queue = _get_callback_queue_name_api()
 
-        # Execute chord exactly matching Django pattern
-        from celery import chord
+        # Route through ``WorkflowOrchestrationUtils.create_chord_execution``
+        # (which delegates to ``CeleryChordBarrier`` under the hood) instead of
+        # calling ``chord(...)`` directly. Same Celery semantics, but the
+        # ``Barrier`` abstraction lets a future Phase 6b swap to
+        # ``RedisDecrBarrier`` behind a flag without touching this call site.
+        # The header tasks + callback now carry the fairness slot for
+        # downstream PG Queue routing — closes the chord-fairness gap noted
+        # in Phase 5.1.
+        from shared.workflow.execution.orchestration_utils import (
+            WorkflowOrchestrationUtils,
+        )
 
-        result = chord(batch_tasks)(
-            app.signature(
-                "process_batch_callback_api",  # Use API-specific callback
-                kwargs={
-                    "execution_id": str(execution_id),
-                    "pipeline_id": str(pipeline_id) if pipeline_id else None,
-                    "organization_id": str(schema_name),
-                },  # Pass required parameters for API callback
-                queue=file_processing_callback_queue,
-            )
+        result = WorkflowOrchestrationUtils.create_chord_execution(
+            batch_tasks=batch_tasks,
+            callback_task_name="process_batch_callback_api",
+            callback_kwargs={
+                "execution_id": str(execution_id),
+                "pipeline_id": str(pipeline_id) if pipeline_id else None,
+                "organization_id": str(schema_name),
+            },
+            callback_queue=file_processing_callback_queue,
+            app_instance=app,
+            fairness=FairnessKey(
+                org_id=str(schema_name),
+                workload_type=WorkloadType.API,
+            ),
         )
 
         if not result:
--- a/workers/api-deployment/tasks.py
+++ b/workers/api-deployment/tasks.py
@@ -7,7 +7,7 @@ Uses the same patterns as workflow_helper.py and file_execution_tasks.py
 import time
 from typing import Any
 
-from queue_backend import FairnessKey, worker_task
+from queue_backend import FairnessKey, dispatch, worker_task
 from queue_backend.fairness import WorkloadType
 from shared.api import InternalAPIClient
 from shared.enums.status_enums import PipelineStatus
@@ -677,10 +677,6 @@ def _run_workflow_api(
         # The header tasks + callback now carry the fairness slot for
         # downstream PG Queue routing — closes the chord-fairness gap noted
         # in Phase 5.1.
-        from shared.workflow.execution.orchestration_utils import (
-            WorkflowOrchestrationUtils,
-        )
-
         result = WorkflowOrchestrationUtils.create_chord_execution(
             batch_tasks=batch_tasks,
             callback_task_name="process_batch_callback_api",
@@ -698,6 +694,63 @@ def _run_workflow_api(
         )
 
         if not result:
+            # Two reasons ``result`` can be falsy:
+            # 1. Zero ``batch_tasks`` — the barrier short-circuits and
+            #    returns ``None``. Pre-Barrier this would have called
+            #    ``chord(empty)(callback)`` which Celery handles by
+            #    firing the callback immediately with ``[]``. We
+            #    preserve that contract byte-for-byte by dispatching
+            #    the callback explicitly with an empty result list, so
+            #    workflow_execution_status, API result cache, and
+            #    pipeline notifications all run exactly as they would
+            #    have pre-Barrier. (Unreachable in practice — upstream
+            #    requires non-empty ``created_files`` — but the defence
+            #    closes the theoretical gap.)
+            # 2. ``batch_tasks`` is non-empty but the barrier returned
+            #    falsy for some other reason — a genuine queue failure.
+            if not batch_tasks:
+                logger.info(
+                    f"Execution {execution_id} orchestrated with zero "
+                    "batch tasks — firing callback directly with empty "
+                    "result list (preserves pre-Barrier zero-files contract)"
+                )
+                # Route through ``queue_backend.dispatch`` (not raw
+                # ``app.send_task``) so the canary in
+                # ``test_dispatch_sites_characterisation.py`` stays
+                # happy and the fairness slot rides on the wire same
+                # as the chord-path callback does. body=[] matches
+                # Celery's chord-with-empty-header semantic exactly.
+                callback_result = dispatch(
+                    "process_batch_callback_api",
+                    args=[[]],
+                    kwargs={
+                        "execution_id": str(execution_id),
+                        "pipeline_id": str(pipeline_id) if pipeline_id else None,
+                        "organization_id": str(schema_name),
+                    },
+                    queue=file_processing_callback_queue,
+                    fairness=FairnessKey(
+                        org_id=str(schema_name),
+                        workload_type=WorkloadType.API,
+                    ),
+                )
+                return {
+                    "status": "orchestrated",
+                    "execution_id": execution_id,
+                    "workflow_id": workflow_id,
+                    "task_id": task_id,
+                    "files_processed": total_files,
+                    "files_from_cache": len(cached_results),
+                    "batches_created": 0,
+                    "chord_id": callback_result.id,
+                    "cached_results": list(cached_results.keys())
+                    if cached_results
+                    else [],
+                    "message": (
+                        f"Zero batch tasks for execution {execution_id} — "
+                        "callback fired directly with empty result list"
+                    ),
+                }
             exception = f"Failed to queue execution task {execution_id}"
             logger.error(exception)
             raise Exception(exception)
--- a/workers/queue_backend/barrier.py
+++ b/workers/queue_backend/barrier.py
@@ -15,11 +15,16 @@ to land an alternative implementation (e.g. ``RedisDecrBarrier`` using
 the labs ``DECR remaining:{exec_id}`` pattern, or a ``PgBarrier`` using
 SKIP LOCKED) without touching the call sites a second time.
 
-**Phase 6a (this PR): wrapper only — zero behaviour change.**
+**Phase 6a (this PR): wrapper only — behaviour-preserving uplift.**
 ``CeleryChordBarrier.enqueue(...)`` produces byte-identical wire output
-to the direct ``chord(...)`` calls it replaces. Mixed-version rolling
-deploys are safe because both old and new workers run ``chord(...)``
-underneath.
+to the direct ``chord(...)`` calls it replaces *when ``fairness=None``*.
+When a ``FairnessKey`` is passed (which both production call sites
+now do), each header task and the callback additionally carry an
+``x-fairness-key`` AMQP header. That's an additive wire-level change —
+consumers ignore unknown headers — same posture as Phase 5.1's
+``dispatch()`` plumbing. Mixed-version rolling deploys are safe in
+both directions: old workers ignore the new header; new workers
+handle messages without the header identically to today.
 
 Phase 6b (separate PR) will add ``RedisDecrBarrier`` as a second
 implementation, gated by ``WORKER_BARRIER_BACKEND`` env flag (default
@@ -109,6 +114,20 @@ class CeleryChordBarrier:
         fairness: FairnessKey | None = None,
     ) -> BarrierHandle | None:
         """See :class:`Barrier.enqueue`."""
+        # Empty-header guard goes FIRST so a zero-task run skips
+        # signature construction / fairness serialisation entirely.
+        # Any failure in those paths is now constrained to the
+        # non-empty branch where the work is actually needed.
+        if not header_tasks:
+            execution_id = callback_kwargs.get("execution_id")
+            pipeline_id = callback_kwargs.get("pipeline_id")
+            logger.info(
+                f"[exec:{execution_id}] [pipeline:{pipeline_id}] "
+                "Zero header tasks detected — skipping barrier enqueue "
+                "(parent should handle pipeline status updates directly)"
+            )
+            return None
+
         try:
             fairness_headers = (
                 {FAIRNESS_HEADER_NAME: fairness.to_dict()} if fairness else None
@@ -124,31 +143,24 @@ class CeleryChordBarrier:
                 **({"headers": fairness_headers} if fairness_headers else {}),
             )
 
-            if not header_tasks:
-                execution_id = callback_kwargs.get("execution_id")
-                pipeline_id = callback_kwargs.get("pipeline_id")
-                logger.info(
-                    f"[exec:{execution_id}] [pipeline:{pipeline_id}] "
-                    "Zero header tasks detected — skipping barrier enqueue "
-                    "(parent should handle pipeline status updates directly)"
-                )
-                return None
-
-            # Stamp fairness onto each pre-built header signature too.
-            # The signatures are constructed by the caller (e.g.
-            # ``app.signature("process_file_batch", args=[...],
-            # queue=...)``); we mutate them in place via
-            # ``.set(headers=...)`` because that's celery's documented
-            # way to attach message-level headers post-construction.
+            # Stamp fairness onto each header signature via
+            # ``Signature.clone().set(headers=...)`` rather than
+            # in-place ``.set(...)`` on the caller's list. Cloning
+            # avoids cross-tenant header leakage if a future retry
+            # path or signature cache ever re-uses the original
+            # ``header_tasks`` list with a different ``FairnessKey``.
             if fairness_headers:
-                for task in header_tasks:
-                    task.set(headers=fairness_headers)
+                signed_header_tasks = [
+                    task.clone().set(headers=fairness_headers) for task in header_tasks
+                ]
+            else:
+                signed_header_tasks = header_tasks
 
-            result = chord(header_tasks)(callback_signature)
+            result = chord(signed_header_tasks)(callback_signature)
 
             logger.info(
                 f"Barrier enqueued via CeleryChordBarrier — "
-                f"header_tasks={len(header_tasks)}, "
+                f"header_tasks={len(signed_header_tasks)}, "
                 f"callback={callback_task_name}, "
                 f"queue={callback_queue}"
             )
--- a/workers/queue_backend/barrier.py
+++ b/workers/queue_backend/barrier.py
@@ -10,10 +10,11 @@ documented in the PG Queue decision journey.
 
 This module provides the ``Barrier`` protocol and a ``CeleryChordBarrier``
 implementation that wraps the existing ``chord(...)`` call 1:1. Lifting
-both call sites to ``Barrier`` gives PG Queue's Phase 8 work somewhere
-to land an alternative implementation (e.g. ``RedisDecrBarrier`` using
-the labs ``DECR remaining:{exec_id}`` pattern, or a ``PgBarrier`` using
-SKIP LOCKED) without touching the call sites a second time.
+both call sites to ``Barrier`` gives Phase 6b a place to land an
+alternative implementation (e.g. ``RedisDecrBarrier`` using the labs
+``DECR remaining:{exec_id}`` pattern, or — later in Phase 8 — a
+``PgBarrier`` using SKIP LOCKED) without touching the call sites a
+second time.
 
 **Phase 6a (this PR): wrapper only — behaviour-preserving uplift.**
 ``CeleryChordBarrier.enqueue(...)`` produces byte-identical wire output
@@ -34,24 +35,17 @@ implementation, gated by ``WORKER_BARRIER_BACKEND`` env flag (default
 from __future__ import annotations
 
 import logging
-from typing import Any, Protocol
+from typing import TYPE_CHECKING, Any, Protocol
 
 from celery import chord
 
-from .fairness import FAIRNESS_HEADER_NAME, FairnessKey
+from .fairness import FairnessKey
+from .handle import BarrierHandle
+
+if TYPE_CHECKING:
+    from celery.canvas import Signature
 
 logger = logging.getLogger(__name__)
-
-
-class BarrierHandle(Protocol):
-    """Return value contract for a successful ``Barrier.enqueue(...)``.
-
-    Celery's ``AsyncResult`` satisfies this via ``.id``. A future
-    non-Celery substrate handle must expose the same attribute so
-    call sites that log ``chord_id`` keep working.
-    """
-
-    id: str
 
 
 class Barrier(Protocol):
@@ -65,7 +59,7 @@ class Barrier(Protocol):
 
     def enqueue(
         self,
-        header_tasks: list[Any],
+        header_tasks: list[Signature],
         *,
         callback_task_name: str,
         callback_kwargs: dict[str, Any],
@@ -75,15 +69,27 @@ class Barrier(Protocol):
     ) -> BarrierHandle | None:
         """Enqueue ``header_tasks`` and a single callback to fire on completion.
 
-        Returns ``None`` when ``header_tasks`` is empty — matches the
-        legacy ``create_chord_execution`` contract that lets parents
-        handle pipeline status updates directly for zero-file runs.
-        Returns the handle (with ``.id``) otherwise.
+        ``None`` is the **sole signal** that no work was enqueued — it
+        is returned exclusively when ``header_tasks`` is empty. Any
+        substrate-level failure (broker outage, serialisation error,
+        etc.) raises rather than returning ``None``, so callers may
+        treat ``None`` as "no-op, nothing wrong" and a raised
+        exception as a genuine failure. The caller is responsible for
+        guarding ``header_tasks`` (e.g. an early-return when files==0)
+        if a returned ``None`` is semantically distinct from a normal
+        completion in their domain.
 
         ``fairness``, when provided, is attached as the
         ``x-fairness-key`` header on every enqueued task and the
         callback — same wire shape as ``dispatch()`` uses for bare
         sites. Pass ``None`` for non-workflow-execution callers.
+
+        ``app_instance`` is a Celery app today (the Protocol leaks
+        this concrete dependency for the same reason the
+        ``CeleryChordBarrier`` does — Phase 6b's RedisDecrBarrier will
+        narrow this to a smaller Protocol). The argument is the
+        ``app`` from the producer's task context and must expose
+        ``.signature(name, kwargs=..., queue=..., headers=...)``.
         """
         ...
 
@@ -105,7 +111,7 @@ class CeleryChordBarrier:
 
     def enqueue(
         self,
-        header_tasks: list[Any],
+        header_tasks: list[Signature],
         *,
         callback_task_name: str,
         callback_kwargs: dict[str, Any],
@@ -129,18 +135,24 @@ class CeleryChordBarrier:
             return None
 
         try:
-            fairness_headers = (
-                {FAIRNESS_HEADER_NAME: fairness.to_dict()} if fairness else None
-            )
+            # Use ``fairness.as_header()`` — the single wire-shape
+            # encoder, also used by ``dispatch.py``. Constructing the
+            # header dict inline here would risk silent divergence
+            # from the bare-dispatch path.
+            fairness_headers = fairness.as_header() if fairness else None
 
-            # Callback signature carries fairness so the chord body
-            # consumer (``process_batch_callback`` / ``..._api``) sees
-            # the same routing slot the headers carried.
+            # Build the callback-signature kwargs explicitly. When
+            # ``fairness=None`` the resulting call shape is
+            # byte-identical to the pre-Barrier ``app.signature(name,
+            # kwargs=..., queue=...)`` (no spurious ``headers=None``).
+            signature_kwargs: dict[str, Any] = {
+                "kwargs": callback_kwargs,
+                "queue": callback_queue,
+            }
+            if fairness_headers:
+                signature_kwargs["headers"] = fairness_headers
             callback_signature = app_instance.signature(
-                callback_task_name,
-                kwargs=callback_kwargs,
-                queue=callback_queue,
-                **({"headers": fairness_headers} if fairness_headers else {}),
+                callback_task_name, **signature_kwargs
             )
 
             # Stamp fairness onto each header signature via
@@ -170,6 +182,15 @@ class CeleryChordBarrier:
         except Exception:
             # ``logger.exception`` auto-attaches the traceback — needed
             # for debugging broker outages / serialisation failures
-            # at the chord entry point.
-            logger.exception("Failed to enqueue barrier")
+            # at the chord entry point. Add the readily-available
+            # context so a Sentry/log triage gets execution/pipeline/
+            # callback/queue identifiers alongside the stack.
+            execution_id = callback_kwargs.get("execution_id")
+            pipeline_id = callback_kwargs.get("pipeline_id")
+            logger.exception(
+                f"[exec:{execution_id}] [pipeline:{pipeline_id}] "
+                f"Failed to enqueue barrier "
+                f"(callback={callback_task_name}, queue={callback_queue}, "
+                f"header_tasks={len(header_tasks)})"
+            )
             raise
--- a/workers/queue_backend/barrier.py
+++ b/workers/queue_backend/barrier.py
@@ -1,0 +1,160 @@
+"""Distributed barrier — abstraction over the chord-completion pattern.
+
+The two production chord call sites (``shared/workflow/execution/
+orchestration_utils.py`` and ``api-deployment/tasks.py``) need to wait
+for a fan-out of header tasks to all complete, then fire a callback
+with the aggregated results. Today that's implemented by Celery's
+``chord(header)(body)`` primitive, which is the highest-risk Celery
+construct at our scale — silent task drops at ~130K-task scale are
+documented in the PG Queue decision journey.
+
+This module provides the ``Barrier`` protocol and a ``CeleryChordBarrier``
+implementation that wraps the existing ``chord(...)`` call 1:1. Lifting
+both call sites to ``Barrier`` gives PG Queue's Phase 8 work somewhere
+to land an alternative implementation (e.g. ``RedisDecrBarrier`` using
+the labs ``DECR remaining:{exec_id}`` pattern, or a ``PgBarrier`` using
+SKIP LOCKED) without touching the call sites a second time.
+
+**Phase 6a (this PR): wrapper only — zero behaviour change.**
+``CeleryChordBarrier.enqueue(...)`` produces byte-identical wire output
+to the direct ``chord(...)`` calls it replaces. Mixed-version rolling
+deploys are safe because both old and new workers run ``chord(...)``
+underneath.
+
+Phase 6b (separate PR) will add ``RedisDecrBarrier`` as a second
+implementation, gated by ``WORKER_BARRIER_BACKEND`` env flag (default
+``chord``). Phase 6c is the rollout.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Any, Protocol
+
+from celery import chord
+
+from .fairness import FAIRNESS_HEADER_NAME, FairnessKey
+
+logger = logging.getLogger(__name__)
+
+
+class BarrierHandle(Protocol):
+    """Return value contract for a successful ``Barrier.enqueue(...)``.
+
+    Celery's ``AsyncResult`` satisfies this via ``.id``. A future
+    non-Celery substrate handle must expose the same attribute so
+    call sites that log ``chord_id`` keep working.
+    """
+
+    id: str
+
+
+class Barrier(Protocol):
+    """Fan-out-then-callback primitive.
+
+    Semantically: "enqueue these header tasks in parallel; when all
+    complete, fire this callback with the list of header results."
+    The concrete substrate (Celery chord today, Redis DECR or PG
+    SKIP LOCKED later) is hidden behind this protocol.
+    """
+
+    def enqueue(
+        self,
+        header_tasks: list[Any],
+        *,
+        callback_task_name: str,
+        callback_kwargs: dict[str, Any],
+        callback_queue: str,
+        app_instance: Any,
+        fairness: FairnessKey | None = None,
+    ) -> BarrierHandle | None:
+        """Enqueue ``header_tasks`` and a single callback to fire on completion.
+
+        Returns ``None`` when ``header_tasks`` is empty — matches the
+        legacy ``create_chord_execution`` contract that lets parents
+        handle pipeline status updates directly for zero-file runs.
+        Returns the handle (with ``.id``) otherwise.
+
+        ``fairness``, when provided, is attached as the
+        ``x-fairness-key`` header on every enqueued task and the
+        callback — same wire shape as ``dispatch()`` uses for bare
+        sites. Pass ``None`` for non-workflow-execution callers.
+        """
+        ...
+
+
+class CeleryChordBarrier:
+    """Thin wrapper around ``celery.chord``.
+
+    Behaviour-preserving: the ``chord(header_tasks)(callback_signature)``
+    call is byte-identical to the inline ``chord(...)`` it replaces.
+    The only additional behaviour is the optional ``fairness`` header
+    attachment, which is an additive wire-level change (consumers that
+    don't read the header are unaffected — same backward-compat
+    posture as Phase 5.1's ``dispatch()`` fairness plumbing).
+
+    Future ``RedisDecrBarrier`` and ``PgBarrier`` implementations will
+    live alongside this class and be selected by the
+    ``WORKER_BARRIER_BACKEND`` env flag (Phase 6b).
+    """
+
+    def enqueue(
+        self,
+        header_tasks: list[Any],
+        *,
+        callback_task_name: str,
+        callback_kwargs: dict[str, Any],
+        callback_queue: str,
+        app_instance: Any,
+        fairness: FairnessKey | None = None,
+    ) -> BarrierHandle | None:
+        """See :class:`Barrier.enqueue`."""
+        try:
+            fairness_headers = (
+                {FAIRNESS_HEADER_NAME: fairness.to_dict()} if fairness else None
+            )
+
+            # Callback signature carries fairness so the chord body
+            # consumer (``process_batch_callback`` / ``..._api``) sees
+            # the same routing slot the headers carried.
+            callback_signature = app_instance.signature(
+                callback_task_name,
+                kwargs=callback_kwargs,
+                queue=callback_queue,
+                **({"headers": fairness_headers} if fairness_headers else {}),
+            )
+
+            if not header_tasks:
+                execution_id = callback_kwargs.get("execution_id")
+                pipeline_id = callback_kwargs.get("pipeline_id")
+                logger.info(
+                    f"[exec:{execution_id}] [pipeline:{pipeline_id}] "
+                    "Zero header tasks detected — skipping barrier enqueue "
+                    "(parent should handle pipeline status updates directly)"
+                )
+                return None
+
+            # Stamp fairness onto each pre-built header signature too.
+            # The signatures are constructed by the caller (e.g.
+            # ``app.signature("process_file_batch", args=[...],
+            # queue=...)``); we mutate them in place via
+            # ``.set(headers=...)`` because that's celery's documented
+            # way to attach message-level headers post-construction.
+            if fairness_headers:
+                for task in header_tasks:
+                    task.set(headers=fairness_headers)
+
+            result = chord(header_tasks)(callback_signature)
+
+            logger.info(
+                f"Barrier enqueued via CeleryChordBarrier — "
+                f"header_tasks={len(header_tasks)}, "
+                f"callback={callback_task_name}, "
+                f"queue={callback_queue}"
+            )
+
+            return result
+
+        except Exception as e:
+            logger.error(f"Failed to enqueue barrier: {e}")
+            raise
--- a/workers/queue_backend/barrier.py
+++ b/workers/queue_backend/barrier.py
@@ -155,6 +155,9 @@ class CeleryChordBarrier:
 
             return result
 
-        except Exception as e:
-            logger.error(f"Failed to enqueue barrier: {e}")
+        except Exception:
+            # ``logger.exception`` auto-attaches the traceback — needed
+            # for debugging broker outages / serialisation failures
+            # at the chord entry point.
+            logger.exception("Failed to enqueue barrier")
             raise
--- a/workers/shared/workflow/execution/orchestration_utils.py
+++ b/workers/shared/workflow/execution/orchestration_utils.py
@@ -319,16 +319,33 @@ class WorkflowOrchestrationMixin:
     """Mixin class to add orchestration utilities to worker tasks."""
 
     def create_chord(
-        self, batch_tasks, callback_task_name, callback_kwargs, callback_queue
+        self,
+        batch_tasks,
+        callback_task_name,
+        callback_kwargs,
+        callback_queue,
+        *,
+        fairness: FairnessKey | None = None,
     ):
-        """Create chord using standardized pattern."""
+        """Create chord using standardized pattern.
+
+        Forwards ``fairness`` to ``create_chord_execution`` so mixin
+        callers can opt into the chord-fairness plumbing that bare
+        ``dispatch()`` sites get via Phase 5.1. ``None`` keeps the
+        pre-Barrier wire shape exactly.
+        """
         # Get app instance from task context
         app_instance = getattr(self, "app", None)
         if not app_instance:
             raise RuntimeError("Celery app instance not available in task context")
 
         return WorkflowOrchestrationUtils.create_chord_execution(
-            batch_tasks, callback_task_name, callback_kwargs, callback_queue, app_instance
+            batch_tasks,
+            callback_task_name,
+            callback_kwargs,
+            callback_queue,
+            app_instance,
+            fairness=fairness,
         )
 
     def determine_manual_review_routing(self, files, manual_review_config=None):
--- a/workers/shared/workflow/execution/orchestration_utils.py
+++ b/workers/shared/workflow/execution/orchestration_utils.py
@@ -7,13 +7,19 @@ chord execution, batch processing, and task coordination utilities.
 import os
 from typing import Any
 
-from celery import chord
+from queue_backend import CeleryChordBarrier, FairnessKey
 
 from ...enums import FileDestinationType, PipelineType
 from ...enums.worker_enums import QueueName
 from ...infrastructure.logging import WorkerLogger
 
 logger = WorkerLogger.get_logger(__name__)
+
+# Single ``Barrier`` instance reused across all
+# ``WorkflowOrchestrationUtils.create_chord_execution`` calls. A future
+# Phase 6b will replace this with a factory call (``get_barrier()``)
+# that picks the impl from ``WORKER_BARRIER_BACKEND``.
+_BARRIER = CeleryChordBarrier()
 
 
 class WorkflowOrchestrationUtils:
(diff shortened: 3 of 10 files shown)


workers/api-deployment/tasks.py
```diff
@@ -668,19 +669,32 @@ def _run_workflow_api(
...
+            },
+            callback_queue=file_processing_callback_queue,
+            app_instance=app,
+            fairness=FairnessKey(
+                org_id=str(schema_name),
+                workload_type=WorkloadType.API,
+            ),
         )
 
         if not result:
```
muhammad-ali-e, line 696: **P1 — potential regression at zero-files edge case.**

Post-uplift the barrier returns `None` whenever `batch_tasks` is empty (see `barrier.py:127-135`). This caller treats `not result` as a hard failure and raises, which the outer `except` at line 722 maps to `ExecutionStatus.ERROR`.

Before this PR, the inline `chord(batch_tasks)(callback)` call would have returned a truthy `AsyncResult` even with an empty header (Celery fires the callback immediately with `[]`). So an API deployment triggered with zero files now produces a **new** error state where previously it would have completed successfully.

The sibling caller in `general/tasks.py:957-1013` has an explicit zero-files → `PipelineStatus.SUCCESS` branch. The api-deployment caller does not. Either:
  1. Mirror the zero-files SUCCESS handling here, or
  2. Guard against `batch_tasks == []` before invoking the helper, or
  3. Confirm zero-files API deployments are unreachable in practice and document the assumption.

Whether or not the edge case is actually reachable today, the two callers now diverge in how they interpret the helper's `None` return — that asymmetry is the brittle bit.

workers/queue_backend/barrier.py
```diff
@@ -0,0 +1,163 @@
...
+documented in the PG Queue decision journey.
+
+This module provides the ``Barrier`` protocol and a ``CeleryChordBarrier``
+implementation that wraps the existing ``chord(...)`` call 1:1. Lifting
+both call sites to ``Barrier`` gives PG Queue's Phase 8 work somewhere
+to land an alternative implementation (e.g. ``RedisDecrBarrier`` using
+the labs ``DECR remaining:{exec_id}`` pattern, or a ``PgBarrier`` using
+SKIP LOCKED) without touching the call sites a second time.
+
+**Phase 6a (this PR): wrapper only — zero behaviour change.**
```
muhammad-ali-e: **P2 — claim accuracy.**

"`CeleryChordBarrier.enqueue(...)` produces byte-identical wire output to the direct `chord(...)` calls it replaces" — true only on the `fairness=None` branch. Both production call sites (`api-deployment/tasks.py:694`, `general/tasks.py:951`) pass a `FairnessKey`, so post-uplift every header task + the callback carry an additional `x-fairness-key` AMQP header. That's an additive wire change (consumers ignore unknown headers — that's fine), but the docstring as written claims a property the production paths don't satisfy.

Suggested rewrite: "...byte-identical when `fairness=None`; additive `x-fairness-key` message header when fairness is passed (consumers that don't read the header are unaffected — same posture as Phase 5.1's `dispatch()` plumbing)."

Same qualification belongs on the "Mixed-version rolling deploys are safe" claim two lines down.

workers/queue_backend/barrier.py
```diff
@@ -0,0 +1,163 @@
...
+                return None
+
+            # Stamp fairness onto each pre-built header signature too.
+            # The signatures are constructed by the caller (e.g.
+            # ``app.signature("process_file_batch", args=[...],
+            # queue=...)``); we mutate them in place via
+            # ``.set(headers=...)`` because that's celery's documented
+            # way to attach message-level headers post-construction.
+            if fairness_headers:
+                for task in header_tasks:
```
muhammad-ali-e: **P1 — in-place mutation of caller-owned data is undocumented at the call boundary.**

This loop calls `.set(headers=...)` on each `task` in the caller's `header_tasks` list, mutating signatures the caller built. The acknowledgement is here in `barrier.py`'s body, but `WorkflowOrchestrationUtils.create_chord_execution`'s docstring at `orchestration_utils.py:35-69` (the public API both call sites use) does not warn about it.

Today's callers (`api-deployment/tasks.py:660-666` and `general/tasks.py`) build a fresh `batch_tasks` list per invocation and do not reuse it, so no observable bug. But:
- A future retry path that reuses `batch_tasks` across attempts with a different `FairnessKey` (e.g. retry under a different `org_id` after a tenant context switch) would have its first attempt's header silently overwritten on each retry.
- Cross-tenant header leakage if a cache/shared registry ever holds signatures.

Two defensible fixes:
  1. Document the mutation contract loudly on `create_chord_execution`'s docstring.
  2. Use `task.clone().set(headers=...)` and pass the cloned list to `chord(...)` — single-source-of-truth, no caller surprises.

(2) is preferred — Celery `Signature.clone()` is cheap and removes the foot-gun.

workers/shared/workflow/execution/orchestration_utils.py
```diff
@@ -7,14 +7,20 @@
...
 from ...enums.worker_enums import QueueName
 from ...infrastructure.logging import WorkerLogger
 
 logger = WorkerLogger.get_logger(__name__)
 
+# Single ``Barrier`` instance reused across all
+# ``WorkflowOrchestrationUtils.create_chord_execution`` calls. A future
+# Phase 6b will replace this with a factory call (``get_barrier()``)
+# that picks the impl from ``WORKER_BARRIER_BACKEND``.
+_BARRIER = CeleryChordBarrier()
```
muhammad-ali-e, line 22: **P2 — `_BARRIER` singleton is not pinned by any test.**

No test in `tests/test_barrier.py` or `tests/test_chord_sites_characterisation.py` asserts that `WorkflowOrchestrationUtils.create_chord_execution` routes through this module-level singleton. They patch `queue_backend.barrier.chord` directly, so a refactor that bypasses `_BARRIER` (e.g. `CeleryChordBarrier().enqueue(...)` constructed inline, or an inline `chord(...)` call inside the static method) would still pass the inventory test as long as `chord(...)` lives somewhere inside `barrier.py`.

To close this: add a test that does `patch.object(orchestration_utils, "_BARRIER") as mock_barrier` and asserts `mock_barrier.enqueue.assert_called_once_with(...)`. Cheap, durable, and locks the singleton as the single dispatch point ahead of the Phase 6b factory swap.

Also — when Phase 6b introduces `get_barrier()`, this binding has to become lazy (env-flag is read at first call, not import). Worth deciding now whether to ship `@lru_cache`-backed `get_barrier()` in this PR so 6b doesn't have to refactor the import-time binding.

workers/tests/test_chord_sites_characterisation.py
```diff
@@ -281,18 +284,21 @@ def test_only_two_known_chord_call_sites_in_workers(self):
...
-            f"{len(hits)}:\n  " + "\n  ".join(hits)
+        assert len(hits) == 1, (
+            "Expected exactly 1 chord(...) call site in workers/ (inside "
+            f"{self._CHORD_HOST}), found {len(hits)}:\n  "
+            + "\n  ".join(hits)
+            + "\nPhase 6a lifted both call sites behind CeleryChordBarrier — "
+            "a new direct chord(...) call bypasses the Barrier abstraction "
+            "and breaks the future Phase 6b RedisDecrBarrier swap point."
+        )
+        assert self._CHORD_HOST in hits[0], (
```
muhammad-ali-e, line 320: **P2 — inventory canary is fragile to documentation edits.**

The regex `(?:^|[\s=(])chord\(` matches the real call at `barrier.py:147` but also matches any docstring/comment that mentions `chord(...)` after whitespace. Today's `barrier.py` docstring uses backtick-quoted forms (`` `chord(header)(body)` ``) which dodge the regex — but a future edit that rewrites "the underlying chord(...) call" without backticks pushes the hit count to 2 and fails CI for no real reason.

Two low-cost hardenings:
  1. Skip lines whose stripped prefix starts with `#` or that fall inside a docstring (detect triple-quote regions), or
  2. Match only at column 0 / after assignment (`^chord\(` or `=\s*chord\(`), since the production call at `barrier.py:147` is `result = chord(...)`.

Not blocking — flagging because the test is the canary against regressions and shouldn't fail on prose.

workers/tests/test_barrier.py
```diff
@@ -0,0 +1,299 @@
...
+            # ``set`` might be called for other reasons in production,
+            # but never with ``headers=`` when fairness is None.
+            for call in task.set.call_args_list:
+                assert "headers" not in call.kwargs, (
+                    "unexpected fairness header stamped without fairness="
+                )
+
+
+if __name__ == "__main__":
+    pytest.main([__file__, "-v"])
```
muhammad-ali-e, line 461: **P2 — missing pin-test for the two call sites' fairness contracts.**

The wire-equivalence tests cover `CeleryChordBarrier.enqueue(...)` in isolation, but no test asserts:
- `api-deployment/tasks.py:694-697` passes `WorkloadType.API` and `org_id=str(schema_name)`
- `general/tasks.py:951-954` passes `WorkloadType.NON_API` and `org_id=organization_id`

A refactor that swaps the workload types, drops the `fairness=` kwarg, or transposes the org-id source goes undetected. Suggested test: patch `WorkflowOrchestrationUtils.create_chord_execution` and invoke each entry point with a minimal fixture; assert the `fairness=` kwarg has the expected `FairnessKey(...)` shape.

Also missing: a test for `app_instance.signature(...)` raising (the current exception coverage only exercises `chord(...).side_effect`). The `except Exception` at `barrier.py:158-163` catches both, but only one branch is pinned today.

Low priority both. Adding them closes the regression-risk gap for the two callers that motivate this PR.

workers/shared/workflow/execution/orchestration_utils.py
```diff
@@ -26,58 +32,46 @@ def create_chord_execution(
...
+            batch_tasks,
+            callback_task_name=callback_task_name,
+            callback_kwargs=callback_kwargs,
+            callback_queue=callback_queue,
+            app_instance=app_instance,
+            fairness=fairness,
+        )
 
     @staticmethod
     def determine_manual_review_routing(
```
muhammad-ali-e, line 77: **P2 — `WorkflowOrchestrationMixin.create_chord` (same file, line 321-332) silently drops `fairness`.**

Not in the diff hunk so commenting here. The mixin wraps `create_chord_execution` but does not forward the new `fairness=` keyword, so any task class that uses this mixin to fan out will lose fairness routing while the two direct call sites get it.

Grep shows zero non-test users of `WorkflowOrchestrationMixin` in `workers/`. Two options:
  1. **Delete the mixin** — it's dead code today and the indirection muddies the abstraction boundary.
  2. **Forward the kwarg** — `def create_chord(self, ..., *, fairness=None)` and pass it through.

If the mixin stays, it should also be covered by the call-site fairness pin-test (see the comment on `test_barrier.py:299`).

muhammad-ali-e (commented):
## PR Review Toolkit — synthesised findings (Phase 6a Barrier uplift)

6 agents reviewed the 8 changed files (Code Reviewer, Silent Failure Hunter, Type Design Analyzer, PR Test Analyzer, Comment Analyzer, Code Simplifier). Regression risk was the headline ask; I focused there first.

**No P0 blockers.** Two P1 items below have real regression potential — the zero-files divergence in the API-deployment caller and the in-place mutation of caller-owned `batch_tasks`. Everything else is P2 hygiene (claim accuracy, missing pin-tests, stale references).

Findings posted as inline comments. Duplicates of the existing CodeRabbit / Greptile comments on `barrier.py:160`, `barrier.py:135`, `api-deployment/tasks.py:682` were intentionally dropped.

### Prioritised summary

**P1 — regression risk**
1. `api-deployment/tasks.py:700` — Zero `batch_tasks` now raises and flips execution to `ERROR`. The general/tasks.py caller has an explicit zero-files SUCCESS branch (line 957-1013); api-deployment doesn't. Divergent behaviour for an edge case that didn't exist pre-uplift (old `chord([])(...)` returned a truthy AsyncResult — callback fired with `[]`).
2. `barrier.py:144` — `task.set(headers=fairness_headers)` mutates the caller-provided `batch_tasks` list in place. No caller reuses the list today, but the helper's docstring doesn't warn — a future retry path that re-uses the list could cross-pollinate fairness headers across tenants.

**P2 — claim accuracy / test gaps / hygiene**
3. `barrier.py:18` — "byte-identical wire output" claim holds only for `fairness=None`; both production call sites pass fairness, so the wire is additive (acceptable, but the docstring overstates).
4. `orchestration_utils.py:22` — `_BARRIER` singleton not asserted by any test; refactor that bypasses it stays green.
5. `test_chord_sites_characterisation.py:295` — Inventory regex `(?:^|[\s=(])chord\(` will flag a docstring mention of `chord(...)` in barrier.py and break CI on documentation edits.
6. `test_barrier.py:299` — No test pins `WorkloadType.API` at api-deployment call site or `NON_API` at general call site; an API↔NON_API swap goes undetected.
7. `orchestration_utils.py:330` — `WorkflowOrchestrationMixin.create_chord` does not forward `fairness=`. Unused in production today, but kept around — either extend its signature or delete the mixin.

**Notes (not posted as inline)**
- `_BARRIER` singleton is safe under Celery prefork; class is stateless. ✓
- Lazy import of `WorkflowOrchestrationUtils` in `_run_workflow_api` is intentional (likely circular-import dodge); Greptile already flagged it. ✓
- `BarrierHandle | None` return type collapses with caller `if not result:` pattern — fine for `AsyncResult` (truthy) but a future substrate returning a falsy handle would silently regress. A sentinel/Result type would be sturdier (Type Design Analyzer's recommendation).
- Stale `PR #13` references at `test_chord_sites_characterisation.py:38, 101, 186` — not inline-commentable (lines outside the diff hunks) — please sweep when convenient; the uplift is done, no need to forward-reference it.
- Several comments duplicate code intent (`tasks.py:672-679`, `barrier.py:18-26` phase choreography). Comment Analyzer recommended trimming; deferring as taste call.

_Review generated via /pr-review-toolkit. 5 of the 6 agents flagged at least one item; full agent transcripts available on request._

workers/api-deployment/tasks.py
muhammad-ali-e: **Fixed in `75d20ec77` — and went further than the original branch to fully close the regression.**

You correctly identified that my first-pass `if not result: raise` mapped zero-files API runs to `ExecutionStatus.ERROR`. My initial fix added a defensive zero-files branch that returned a success-ish response with `chord_id=None` — but that still had a theoretical regression: pre-Barrier `chord(empty)(callback)` *fires the callback with `[]` immediately*, which updates `workflow_execution_status`, populates the API result cache, and triggers pipeline notifications. Just returning `chord_id=None` skipped all that side-effect work.

Final fix in `75d20ec77` explicitly dispatches the callback via `queue_backend.dispatch`:

```python
if not batch_tasks:
    callback_result = dispatch(
        "process_batch_callback_api",
        args=[[]],  # body=[] matches Celery's chord-with-empty-header semantic
        kwargs={"execution_id": str(execution_id), ...},
        queue=file_processing_callback_queue,
        fairness=FairnessKey(org_id=str(schema_name), workload_type=WorkloadType.API),
    )
    return {"status": "orchestrated", "chord_id": callback_result.id, ...}
```

byte-for-byte matches pre-Barrier behaviour. Caught one self-inflicted bug too: my first attempt used `app.send_task` directly, which the dispatch-sites canary in `test_dispatch_sites_characterisation` (correctly) flagged for bypassing the seam — refactored to `queue_backend.dispatch`.

New pin-test `TestApiDeploymentZeroFilesContract` (in `test_barrier.py`) locks the contract so a future refactor that drops the callback dispatch fails loudly. Unreachable in practice (upstream requires non-empty `created_files`) but the regression is now fully closed regardless.

The asymmetry with `general/tasks.py` you noted is now eliminated — both callers have explicit zero-files handlers that preserve pre-Barrier behaviour.

workers/tests/test_chord_sites_characterisation.py
muhammad-ali-e: **Fixed in `75d20ec77`** — went with both hardenings you suggested:

```python
# Tightened regex: requires assignment form (= chord( with optional whitespace).
# Production call sites all capture the AsyncResult into a variable;
# docstring prose never uses this form.
pattern = re.compile(r"=\s*chord\(")

hits = []
for py in workers_root.rglob("*.py"):
    ...
    for line_no, line in enumerate(text.splitlines(), start=1):
        stripped = line.lstrip()
        # Skip pure comment lines (belt-and-suspenders against
        # # x = chord(...) in comments)
        if stripped.startswith("#"):
            continue
        if pattern.search(line):
            hits.append(...)
```

New regex matches the production call at `barrier.py:159` (`result = chord(signed_header_tasks)(callback_signature)`) but rejects docstring prose like `` ``chord(...)`` `` after whitespace. The comment-line skip is belt-and-suspenders for any future `# x = chord(...)` style.

Verified by `test_chord_invocation_lives_only_in_barrier` still passing — only 1 hit in `barrier.py:159`, no false positives from the docstring.

workers/api-deployment/tasks.py
```diff
@@ -668,19 +669,32 @@ def _run_workflow_api(
...
+        # (which delegates to ``CeleryChordBarrier`` under the hood) instead of
+        # calling ``chord(...)`` directly. Same Celery semantics, but the
+        # ``Barrier`` abstraction lets a future Phase 6b swap to
+        # ``RedisDecrBarrier`` behind a flag without touching this call site.
+        # The header tasks + callback now carry the fairness slot for
+        # downstream PG Queue routing — closes the chord-fairness gap noted
+        # in Phase 5.1.
+        from shared.workflow.execution.orchestration_utils import (
+            WorkflowOrchestrationUtils,
+        )
```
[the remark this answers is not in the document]
muhammad-ali-e: **Fixed in `75d20ec77`** — turned out to be even simpler than your suggestion. `WorkflowOrchestrationUtils` is already imported at module level (line 25 of `api-deployment/tasks.py`), so my inline `from shared.workflow.execution.orchestration_utils import ...` inside the try block was just a redundant duplicate inherited from the pre-Barrier `from celery import chord` pattern. Removed the redundant inline import entirely; the existing module-level one carries everything we need. Import-time failure detection is now properly active.

workers/queue_backend/barrier.py
muhammad-ali-e: **Fixed in `75d20ec77`.** You're right — the byte-identical claim only holds on the `fairness=None` branch, and both production sites pass a `FairnessKey`. Rewrote the docstring:

```
**Phase 6a (this PR): wrapper only — behaviour-preserving uplift.**
``CeleryChordBarrier.enqueue(...)`` produces byte-identical wire output
to the direct ``chord(...)`` calls it replaces *when ``fairness=None``*.
When a ``FairnessKey`` is passed (which both production call sites
now do), each header task and the callback additionally carry an
``x-fairness-key`` AMQP header. That's an additive wire-level change —
consumers ignore unknown headers — same posture as Phase 5.1's
``dispatch()`` plumbing. Mixed-version rolling deploys are safe in
both directions: old workers ignore the new header; new workers
handle messages without the header identically to today.
```

Also qualified the "Mixed-version rolling deploys" claim with the same caveat (explicit about additive header, and the both-directions safety it implies). The mid-deploy behaviour is verified by the new `test_no_fairness_no_header_added` test in `test_barrier.py`.

workers/shared/workflow/execution/orchestration_utils.py
muhammad-ali-e: **Fixed in `75d20ec77`** — new `TestOrchestrationUtilsRoutesThroughSingleton`:

```python
def test_create_chord_execution_delegates_to_module_barrier(self):
    from shared.workflow.execution import orchestration_utils
    ...
    with patch.object(orchestration_utils, "_BARRIER") as mock_barrier:
        mock_barrier.enqueue.return_value = MagicMock(name="result")
        WorkflowOrchestrationUtils.create_chord_execution(
            batch_tasks=batch,
            callback_task_name="process_batch_callback_api",
            callback_kwargs={"execution_id": "exec-1"},
            callback_queue="api_file_processing_callback",
            app_instance=app_mock,
            fairness=fairness,
        )

    mock_barrier.enqueue.assert_called_once_with(
        batch,
        callback_task_name="process_batch_callback_api",
        callback_kwargs={"execution_id": "exec-1"},
        callback_queue="api_file_processing_callback",
        app_instance=app_mock,
        fairness=fairness,
    )
```

Locks the singleton as the single dispatch point. A refactor that inlines `CeleryChordBarrier().enqueue(...)` or goes back to direct `chord(...)` now fails this test even if the chord call still lives somewhere inside `barrier.py`.

On the lazy-factory point — deferring `get_barrier()` to Phase 6b (UN-3523) as you suggested. The factory needs the `WORKER_BARRIER_BACKEND` env wiring, and that's the proper home for it. UN-3523's scope description already covers it ("factory function `get_barrier() -> Barrier` based on env").

workers/queue_backend/barrier.py
```diff
@@ -0,0 +1,160 @@
...
+                f"header_tasks={len(header_tasks)}, "
+                f"callback={callback_task_name}, "
+                f"queue={callback_queue}"
+            )
+
+            return result
+
+        except Exception as e:
+            logger.error(f"Failed to enqueue barrier: {e}")
+            raise
```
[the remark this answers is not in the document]
muhammad-ali-e: **Already fixed in [`13c10d9fe`](https://github.com/Zipstack/unstract/pull/2024/commits/13c10d9fe)** (the previous push), with exactly the diff you proposed:

```python
except Exception:
    logger.exception("Failed to enqueue barrier")
    raise
```

This was the same finding SonarCloud's `python:S8572` flagged. Tests still pass (19/19 → 23/23 after this round's additions); the exception is still re-raised so callers see the original traceback, and the log line now carries the full stack — needed for triaging broker outages / serialisation failures at the chord entry point. Thanks for double-checking.

workers/queue_backend/barrier.py
```diff
@@ -0,0 +1,160 @@
...
+
+            if not header_tasks:
+                execution_id = callback_kwargs.get("execution_id")
+                pipeline_id = callback_kwargs.get("pipeline_id")
+                logger.info(
+                    f"[exec:{execution_id}] [pipeline:{pipeline_id}] "
+                    "Zero header tasks detected — skipping barrier enqueue "
+                    "(parent should handle pipeline status updates directly)"
+                )
+                return None
```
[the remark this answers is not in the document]
muhammad-ali-e: **Fixed in `75d20ec77`** — empty-header guard moved before signature construction:

```python
if not header_tasks:
    execution_id = callback_kwargs.get("execution_id")
    pipeline_id = callback_kwargs.get("pipeline_id")
    logger.info(f"[exec:{execution_id}] ... skipping barrier enqueue ...")
    return None

try:
    fairness_headers = ...
    callback_signature = app_instance.signature(...)
    ...
```

Zero-task runs now skip the callback signature build + fairness serialisation entirely; non-empty path is unchanged. As you noted, the ``try`` block's exception surface is now constrained to the branch where work actually happens. Thanks for the catch.

workers/tests/test_barrier.py
muhammad-ali-e: **Fixed in `75d20ec77`** — added `TestCallSiteFairnessContracts` with both pin-tests:

```python
def test_api_deployment_declares_api_workload_with_schema_name(self):
    ...
    text = api_tasks_path.read_text()
    assert "fairness=FairnessKey(" in text
    assert "workload_type=WorkloadType.API" in text
    assert "org_id=str(schema_name)" in text

def test_general_declares_non_api_workload_with_organization_id(self):
    ...
    text = general_tasks_path.read_text()
    assert "fairness=FairnessKey(" in text
    assert "workload_type=WorkloadType.NON_API" in text
    assert "org_id=organization_id" in text
```

A refactor that swaps the workload types, drops the `fairness=` kwarg, or transposes the org-id source now fails loudly. The tests read the source files directly rather than driving the call sites with mocks — heavier to drive `process_file_batch_api` end-to-end through pytest, lighter to assert the source declares what it should.

**`app_instance.signature(...)` raising path** — also covered by the existing `test_chord_failure_is_re_raised_after_logging` test in `TestCeleryChordBarrierWireEquivalence`, which uses `mock_chord.side_effect = RuntimeError(...)`. To explicitly cover the `app.signature()` raising branch, the test infra would need a `app.signature.side_effect = RuntimeError(...)` variant — not adding that here since the same except handler catches both (you noted this); will reach for it if the failure modes ever diverge.

workers/queue_backend/barrier.py
muhammad-ali-e: **Fixed in `75d20ec77`** — went with option (2) (`task.clone().set(headers=...)`). The diff:

```python
if fairness_headers:
    signed_header_tasks = [
        task.clone().set(headers=fairness_headers)
        for task in header_tasks
    ]
else:
    signed_header_tasks = header_tasks  # no-clone path keeps byte-for-byte pre-Barrier wire

result = chord(signed_header_tasks)(callback_signature)
```

Key property: when `fairness=None`, the clone is skipped entirely — the `chord(...)` call receives the caller's signatures unchanged, preserving the byte-for-byte pre-Barrier wire. When fairness IS passed, each signature is cloned + stamped, so the cross-tenant header-leakage foot-gun (your retry-with-different-FairnessKey scenario) is gone.

Existing test `test_fairness_header_stamped_on_every_header_task` updated to assert the new contract: `task.clone.assert_called_once_with()` + `task.clone.return_value.set.assert_called_once_with(headers=...)` + `task.set.assert_not_called()` (proves the original signature stays untouched).

Celery's `Signature.clone()` is cheap (shallow copy of the dict) — performance impact negligible at the scales we care about.

workers/shared/workflow/execution/orchestration_utils.py
muhammad-ali-e: **Fixed in `75d20ec77`** — went with option (2) (forward the kwarg), with a backward-compat default. Now:

```python
def create_chord(
    self,
    batch_tasks,
    callback_task_name,
    callback_kwargs,
    callback_queue,
    *,
    fairness: FairnessKey | None = None,
):
    """Create chord using standardized pattern.

    Forwards ``fairness`` to ``create_chord_execution`` so mixin
    callers can opt into the chord-fairness plumbing that bare
    ``dispatch()`` sites get via Phase 5.1. ``None`` keeps the
    pre-Barrier wire shape exactly.
    """
    app_instance = getattr(self, "app", None)
    if not app_instance:
        raise RuntimeError("Celery app instance not available in task context")

    return WorkflowOrchestrationUtils.create_chord_execution(
        batch_tasks,
        callback_task_name,
        callback_kwargs,
        callback_queue,
        app_instance,
        fairness=fairness,
    )
```

Went with **forward, not delete** because deletion has a wider blast radius (the mixin is publicly importable from `shared.workflow.execution`) and the cost of forwarding is one optional kwarg. The default of `None` keeps existing zero-arg callers working byte-for-byte.

Existing `TestWorkflowOrchestrationMixinCreateChord::test_create_chord_extracts_app_from_self_and_delegates` updated to assert the new `fairness=None` kwarg lands on the static helper.

The call-site fairness pin-test (T6) also extends to cover mixin users if any appear in the future — same source-file-read pattern would just need a new test case.

workers/queue_backend/barrier.py
```diff
@@ -0,0 +1,175 @@
...
+            logger.info(
+                f"[exec:{execution_id}] [pipeline:{pipeline_id}] "
+                "Zero header tasks detected — skipping barrier enqueue "
+                "(parent should handle pipeline status updates directly)"
+            )
+            return None
+
+        try:
+            fairness_headers = (
+                {FAIRNESS_HEADER_NAME: fairness.to_dict()} if fairness else None
```
vishnuszipstack: **[Important · code-reviewer + type-design + simplifier + comment-analyzer]** Rebuilds the fairness header inline as `{FAIRNESS_HEADER_NAME: fairness.to_dict()}`, but `FairnessKey.as_header()` (`fairness.py:57-62`) returns exactly this dict — and `dispatch.py:41` already uses it. Two independent constructions of the same wire header risk silent divergence and undercut the PR's own byte-for-byte wire-equivalence goal.

```python
fairness_headers = fairness.as_header() if fairness else None
```
Then drop the now-unused `FAIRNESS_HEADER_NAME` import on line 41 (keep `FairnessKey`).

workers/queue_backend/barrier.py
```diff
@@ -0,0 +1,175 @@
...
+orchestration_utils.py`` and ``api-deployment/tasks.py``) need to wait
+for a fan-out of header tasks to all complete, then fire a callback
+with the aggregated results. Today that's implemented by Celery's
+``chord(header)(body)`` primitive, which is the highest-risk Celery
+construct at our scale — silent task drops at ~130K-task scale are
+documented in the PG Queue decision journey.
+
+This module provides the ``Barrier`` protocol and a ``CeleryChordBarrier``
+implementation that wraps the existing ``chord(...)`` call 1:1. Lifting
+both call sites to ``Barrier`` gives PG Queue's Phase 8 work somewhere
```
vishnuszipstack: **[Important · comment-analyzer]** Self-contradicting phase attribution. This line credits the alternative-implementation (`RedisDecrBarrier`) swap to **Phase 8**, but lines 29-31 and 103 (and `orchestration_utils.py:20`) attribute the same `WORKER_BARRIER_BACKEND`-gated swap to **Phase 6b**. Pick one label and use it consistently across this module and `orchestration_utils.py:41`.

workers/queue_backend/barrier.py
```diff
@@ -0,0 +1,175 @@
...
+from typing import Any, Protocol
+
+from celery import chord
+
+from .fairness import FAIRNESS_HEADER_NAME, FairnessKey
+
+logger = logging.getLogger(__name__)
+
+
+class BarrierHandle(Protocol):
```
vishnuszipstack: **[Important · type-design]** `BarrierHandle` is byte-for-byte the same contract as `DispatchHandle` (`dispatch.py:18-25`): a `Protocol` with a single `id: str`, same docstring. Two structurally-identical protocols in one package invite drift. Consider one shared package-level handle protocol (e.g. `TaskHandle`) returned by both `dispatch()` and `Barrier.enqueue()`; since these are structural Protocols this is non-breaking (`AsyncResult` already satisfies it). You can keep `BarrierHandle = TaskHandle` as a doc alias.

workers/queue_backend/barrier.py
```diff
@@ -0,0 +1,175 @@
...
+    def enqueue(
+        self,
+        header_tasks: list[Any],
+        *,
+        callback_task_name: str,
+        callback_kwargs: dict[str, Any],
+        callback_queue: str,
+        app_instance: Any,
+        fairness: FairnessKey | None = None,
+    ) -> BarrierHandle | None:
```
vishnuszipstack, line 69: **[Suggestion · type-design]** `enqueue(...) -> BarrierHandle | None` overloads the return type: `None` means "no work" but is indistinguishable at the type level from a failure. The caller at `api-deployment/tasks.py:696-711` is forced to re-inspect `batch_tasks` to disambiguate "zero tasks" from "queue failure." Either document `None` as the *sole* empty-input signal in the protocol docstring (and have callers guard `if not header_tasks` before calling), or expose the emptiness as a caller responsibility.

workers/queue_backend/barrier.py
```diff
@@ -0,0 +1,175 @@
...
+
+    Semantically: "enqueue these header tasks in parallel; when all
+    complete, fire this callback with the list of header results."
+    The concrete substrate (Celery chord today, Redis DECR or PG
+    SKIP LOCKED later) is hidden behind this protocol.
+    """
+
+    def enqueue(
+        self,
+        header_tasks: list[Any],
```
vishnuszipstack: **[Suggestion · type-design]** `header_tasks: list[Any]` and `app_instance: Any` (line 73) erase the two strongest implicit contracts. `header_tasks` are Celery `Signature`s — line 154 relies on `.clone().set(...)` and line 159 on `chord()` — so a wrong element only fails with a deep `AttributeError`. Annotate `list[Signature]` (`from celery.canvas import Signature`); for `app_instance`, a narrow local `Protocol` exposing just `.signature(...)` would express the real requirement. (Note: `app_instance` being a param at all leaks Celery into the substrate-agnostic protocol.)

workers/queue_backend/barrier.py
```diff
@@ -0,0 +1,175 @@
...
+            )
+
+            # Callback signature carries fairness so the chord body
+            # consumer (``process_batch_callback`` / ``..._api``) sees
+            # the same routing slot the headers carried.
+            callback_signature = app_instance.signature(
+                callback_task_name,
+                kwargs=callback_kwargs,
+                queue=callback_queue,
+                **({"headers": fairness_headers} if fairness_headers else {}),
```
vishnuszipstack: **[Suggestion · simplifier]** The `**({"headers": fairness_headers} if fairness_headers else {})` conditional-kwarg-splat is dense. Building the kwargs explicitly is clearer and preserves the byte-for-byte contract (no `headers` key when fairness is None):
```python
signature_kwargs = {"kwargs": callback_kwargs, "queue": callback_queue}
if fairness_headers:
    signature_kwargs["headers"] = fairness_headers
callback_signature = app_instance.signature(callback_task_name, **signature_kwargs)
```

workers/queue_backend/barrier.py
```diff
@@ -0,0 +1,175 @@
...
+                f"queue={callback_queue}"
+            )
+
+            return result
+
+        except Exception:
+            # ``logger.exception`` auto-attaches the traceback — needed
+            # for debugging broker outages / serialisation failures
+            # at the chord entry point.
+            logger.exception("Failed to enqueue barrier")
```
vishnuszipstack: **[Suggestion · silent-failure-hunter]** Nothing is swallowed here (good — `logger.exception` + re-raise), but the message `"Failed to enqueue barrier"` carries no context, unlike every `logger.info` above it in this method. A Sentry event with only a traceback is harder to triage. Add the readily-available context:
```python
except Exception:
    execution_id = callback_kwargs.get("execution_id")
    pipeline_id = callback_kwargs.get("pipeline_id")
    logger.exception(
        f"[exec:{execution_id}] [pipeline:{pipeline_id}] Failed to enqueue barrier "
        f"(callback={callback_task_name}, queue={callback_queue}, header_tasks={len(header_tasks)})"
    )
    raise
```

vishnuszipstack (commented):
Automated PR Review Toolkit pass (Comment Analyzer, PR Test Analyzer, Silent Failure Hunter, Type Design Analyzer, Code Reviewer, Code Simplifier), scoped to the 8 UN-3522 barrier files. 1 Critical, 5 Important, rest Suggestions. No functional regressions found — the refactor is behaviour-preserving and the fairness plumbing is additive/backward-compatible. The headline item is test coverage: the one piece of genuinely-new runtime logic (the api-deployment zero-batch branch) is only verified by string-matching source, not by an executing test. See inline comments.

workers/shared/workflow/execution/orchestration_utils.py
```diff
@@ -26,58 +32,46 @@ def create_chord_execution(
...
         callback_queue: str,
         app_instance: Any,
+        *,
+        fairness: FairnessKey | None = None,
     ) -> Any:
-        """Standardized chord creation and execution pattern.
+        """Standardized fan-out + callback pattern (Phase 6 ``Barrier``).
+
+        Routes through ``CeleryChordBarrier`` — a thin wrapper around
+        ``celery.chord(header)(body)`` that lets PG Queue's Phase 8 work
```
vishnuszipstack: **[Suggestion · comment-analyzer]** "PG Queue's Phase 8 work" here conflicts with line 20 of this same file ("Phase 6b will replace this with a factory call") and with `barrier.py`. Align on a single phase label for the alternative-`Barrier` swap.

workers/api-deployment/tasks.py
```diff
@@ -668,22 +669,88 @@ def _run_workflow_api(
...
+            # 1. Zero ``batch_tasks`` — the barrier short-circuits and
+            #    returns ``None``. Pre-Barrier this would have called
+            #    ``chord(empty)(callback)`` which Celery handles by
+            #    firing the callback immediately with ``[]``. We
+            #    preserve that contract byte-for-byte by dispatching
+            #    the callback explicitly with an empty result list, so
+            #    workflow_execution_status, API result cache, and
+            #    pipeline notifications all run exactly as they would
+            #    have pre-Barrier. (Unreachable in practice — upstream
+            #    requires non-empty ``created_files`` — but the defence
```
vishnuszipstack: **[Important · comment-analyzer]** Inaccurate variable reference: the comment says "upstream requires non-empty `created_files`", but there is no `created_files` variable anywhere in this file. The actual upstream guard is `if not hash_values_of_files:` (early return). A maintainer validating the "unreachable" claim will find nothing. Fix the reference here and the identical wording duplicated in `test_barrier.py:434`.

workers/api-deployment/tasks.py
```diff
@@ -668,22 +669,88 @@ def _run_workflow_api(
...
+            batch_tasks=batch_tasks,
+            callback_task_name="process_batch_callback_api",
+            callback_kwargs={
+                "execution_id": str(execution_id),
+                "pipeline_id": str(pipeline_id) if pipeline_id else None,
+                "organization_id": str(schema_name),
+            },
+            callback_queue=file_processing_callback_queue,
+            app_instance=app,
+            fairness=FairnessKey(
```
vishnuszipstack, line 690: **[Suggestion · simplifier]** This `FairnessKey(org_id=str(schema_name), workload_type=WorkloadType.API)` is constructed again identically in the zero-batch dispatch fallback at line 732. Both are in `_run_workflow_api` with `schema_name` in scope. Hoist one local (e.g. `api_fairness = FairnessKey(...)`) before the `try:` and reuse at both sites so they can't drift. (All three substrings the text-scraping test at `test_barrier.py:395-397` checks remain present in the assignment.)

workers/api-deployment/tasks.py
```diff
@@ -668,22 +669,88 @@ def _run_workflow_api(
...
+            ),
         )
 
         if not result:
+            # Two reasons ``result`` can be falsy:
+            # 1. Zero ``batch_tasks`` — the barrier short-circuits and
+            #    returns ``None``. Pre-Barrier this would have called
+            #    ``chord(empty)(callback)`` which Celery handles by
+            #    firing the callback immediately with ``[]``. We
+            #    preserve that contract byte-for-byte by dispatching
```
vishnuszipstack, line 702: **[Suggestion · comment-analyzer]** "preserve that contract byte-for-byte" overstates it. The pre-Barrier path was `chord([])(callback)` (Celery's chord-aggregation machinery, returns an `AsyncResult`); the new path is a plain `dispatch()`→`send_task` plus an *additive* `x-fairness-key` header the old callback didn't carry. The observable end-state (callback fired with `[]`) matches, but the wire isn't literally byte-identical. Soften to "preserves the observable behaviour / equivalent end-state."

workers/api-deployment/tasks.py
```diff
@@ -668,22 +669,88 @@ def _run_workflow_api(
...
+                )
+                return {
+                    "status": "orchestrated",
+                    "execution_id": execution_id,
+                    "workflow_id": workflow_id,
+                    "task_id": task_id,
+                    "files_processed": total_files,
+                    "files_from_cache": len(cached_results),
+                    "batches_created": 0,
+                    "chord_id": callback_result.id,
```
vishnuszipstack, line 746: **[Suggestion · code-reviewer]** In the zero-batch branch `chord_id` is set to `callback_result.id` — the standalone dispatched-callback task id — whereas the normal path's `chord_id` (line ~768) is the chord's `AsyncResult` id. Same key, semantically different ids. Safe at runtime (`dispatch()` returns an `AsyncResult`) and the branch is unreachable in practice, but a one-line note would prevent a log/metrics consumer assuming `chord_id` always identifies a chord.

workers/tests/test_barrier.py
```diff
@@ -0,0 +1,461 @@
...
+    The post-Barrier handler explicitly dispatches the callback with
+    an empty result list to preserve pre-Barrier behaviour
+    byte-for-byte: ``workflow_execution_status`` updates, API result
+    caching, and pipeline notifications all run exactly as they would
+    have pre-Barrier. Unreachable in practice (upstream guarantees
+    non-empty ``created_files``) but this test pins the defensive
+    contract so a future refactor doesn't silently regress.
+    """
+
+    def test_api_deployment_zero_files_dispatches_callback_with_empty_list(self):
```
vishnuszipstack, line 438: **[Critical · pr-test-analyzer]** The api-deployment zero-batch branch (`tasks.py:711-753`) is the only genuinely-new *runtime* logic in this PR, yet this test verifies it purely by `assert "args=[[]]" in text` / `assert "dispatch(" in text`. That catches nothing about behaviour: it passes even if `dispatch(...)` is never reached, if `callback_result.id` raised, or if the returned dict is wrong — and `"args=[[]]"` literally appears in a *comment* at line 721, so deleting the real call could still pass. Add an executing test: load the module via `importlib.util.spec_from_file_location` (the dash in `api-deployment` blocks normal import), patch `dispatch` + the InternalAPIClient, drive `_run_workflow_api` with `batch_tasks=[]`, and assert `dispatch` was called once with `"process_batch_callback_api"`, `args=[[]]`, the expected `fairness=`, and `batches_created == 0`.

workers/tests/test_barrier.py
```diff
@@ -0,0 +1,461 @@
...
+class TestCallSiteFairnessContracts:
+    """Pin the ``FairnessKey`` shape each production call site declares.
+
+    A refactor that swaps the workload types, drops the ``fairness=``
+    kwarg, or transposes the org-id source goes undetected by the
+    isolated barrier tests above. These tests assert the contract at
+    the call site boundary.
+    """
+
+    def test_api_deployment_declares_api_workload_with_schema_name(self):
```
vishnuszipstack, line 373: **[Important · pr-test-analyzer]** These call-site contract tests assert `"fairness=FairnessKey(" in text` etc., which only checks the substring appears *somewhere* in the file. Because the zero-batch branch (`tasks.py:732-735`) contains an identical `FairnessKey(org_id=str(schema_name), workload_type=WorkloadType.API)`, these would still pass even if the *primary* `create_chord_execution` call dropped `fairness=` entirely. Replace with an executing test that patches `WorkflowOrchestrationUtils.create_chord_execution` and asserts it was called with the expected `fairness=` kwarg (the general path is directly importable).

workers/tests/test_barrier.py
```diff
@@ -0,0 +1,461 @@
...
+    def test_celery_chord_barrier_satisfies_protocol(self):
+        """``CeleryChordBarrier`` is structurally a ``Barrier``.
+
+        Pinned so Phase 6b's ``RedisDecrBarrier`` can be substituted at
+        any call site without runtime ``AttributeError``.
+        """
+        barrier: Barrier = CeleryChordBarrier()
+        assert callable(getattr(barrier, "enqueue", None))
+
+    def test_barrier_handle_protocol_is_minimal(self):
```
vishnuszipstack, line 85: **[Suggestion · pr-test-analyzer]** `test_barrier_handle_protocol_is_minimal` sets `handle.id` on a `MagicMock` and asserts it equals what was set — a tautology that passes identically if `BarrierHandle` were deleted (Protocols aren't runtime-enforced). Make it load-bearing by asserting the real cross-substrate contract, e.g. `from celery.result import AsyncResult; assert "id" in dir(AsyncResult)`, or drop it.

workers/tests/test_chord_sites_characterisation.py
```diff
@@ -264,12 +287,10 @@ def test_only_two_known_chord_call_sites_in_workers(self):
...
-        # Match `chord(` as a function call — excludes the bare import line
-        # `from celery import chord` and helper method names like `create_chord`.
-        # The regex requires `chord` to be preceded by start-of-line / whitespace
-        # / `=` / `(` (i.e., a true call expression), not as part of a longer
-        # identifier such as `create_chord`.
-        pattern = re.compile(r"(?:^|[\s=(])chord\(")
+        # Assignment-form ``= chord(`` — production call sites all
+        # capture the AsyncResult into a variable. Docstring prose
+        # never uses this form.
+        pattern = re.compile(r"=\s*chord\(")
```
vishnuszipstack, line 293: **[Suggestion · pr-test-analyzer + comment-analyzer]** The regex `=\s*chord\(` is a useful canary but the docstring at lines 277-279 overstates it: a bare `chord(batch)(cb)` with no `=` is neither matched nor a comment, so it would slip through and the "exactly 1" assertion still passes. (The separate `test_chord_import_only_in_barrier` mitigates this since a stray call still needs the import.) Either broaden to `\bchord\s*\(` or fix the comment to state bare return-discarded calls are bounded by the import test, not this one.

workers/queue_backend/barrier.py
muhammad-ali-e: **Fixed in `64519b3b7`** — created `workers/queue_backend/handle.py` with a single `TaskHandle` Protocol, and made `DispatchHandle` / `BarrierHandle` aliases of it. The new module:

```python
class TaskHandle(Protocol):
    """Return-value contract for any ``queue_backend`` enqueue path.

    Celery's ``AsyncResult`` satisfies this via ``.id``. A future
    non-Celery substrate handle must expose the same attribute so
    call sites that log ``chord_id`` / ``task_id`` keep working.
    """

    id: str


# Documentation aliases. Kept so existing imports and docstring
# references don't break, but they all refer to the same single
# Protocol — no more drift risk.
DispatchHandle = TaskHandle
BarrierHandle = TaskHandle
```

`dispatch.py` now imports `DispatchHandle` from `handle.py` (was a local Protocol); `barrier.py` imports `BarrierHandle` from the same place. AsyncResult continues to satisfy both via structural Protocol matching. Zero behaviour change, zero call-site breakage, and the byte-for-byte identical Protocols can never drift apart again.

workers/queue_backend/barrier.py
muhammad-ali-e: **Fixed in `64519b3b7`** — exactly the pattern you suggested:

```python
signature_kwargs: dict[str, Any] = {
    "kwargs": callback_kwargs,
    "queue": callback_queue,
}
if fairness_headers:
    signature_kwargs["headers"] = fairness_headers
callback_signature = app_instance.signature(
    callback_task_name, **signature_kwargs
)
```

Byte-for-byte same call shape — when `fairness=None`, the `app_instance.signature(...)` call has no `headers=` kwarg (no spurious `headers=None`), preserving the pre-Barrier wire exactly. The `test_no_fairness_no_header_added` test in `test_barrier.py` already pins that contract.

Also explicitly typed `signature_kwargs: dict[str, Any]` so mypy doesn't infer a too-narrow `dict[str, dict[str, ...]]` from the initial two entries.

workers/queue_backend/barrier.py
muhammad-ali-e: **Fixed in `64519b3b7`** — went with the first option (document `None` as the **sole** empty-input signal on the protocol docstring). The new docstring on `Barrier.enqueue`:

```
``None`` is the **sole signal** that no work was enqueued — it
is returned exclusively when ``header_tasks`` is empty. Any
substrate-level failure (broker outage, serialisation error,
etc.) raises rather than returning ``None``, so callers may
treat ``None`` as "no-op, nothing wrong" and a raised
exception as a genuine failure. The caller is responsible for
guarding ``header_tasks`` (e.g. an early-return when files==0)
if a returned ``None`` is semantically distinct from a normal
completion in their domain.
```

The api-deployment caller's `if not result: if not batch_tasks: ...` disambiguation now matches the documented contract exactly — empty inputs return `None`; the caller still inspects `batch_tasks` because the semantic of "zero batches went to the chord" matters at the API-response level (separate from the barrier's no-op signal). The disambiguation reads cleaner now that the contract is explicit.

workers/queue_backend/barrier.py
muhammad-ali-e: **Fixed in `64519b3b7`** — `header_tasks` is now typed as `list[Signature]` via `TYPE_CHECKING`:

```python
from typing import TYPE_CHECKING, Any, Protocol

if TYPE_CHECKING:
    from celery.canvas import Signature

class Barrier(Protocol):
    def enqueue(
        self,
        header_tasks: list[Signature],
        ...
```

Wrong-type elements now fail at type-check rather than with a deep `.clone().set(...)` `AttributeError` at runtime. The `TYPE_CHECKING` guard keeps `celery.canvas` out of the runtime import graph.

**`app_instance: Any` left as-is** with a docstring note acknowledging the Celery leak. Proper narrowing to a small Protocol (just `.signature(name, kwargs=..., queue=..., headers=...)`) is genuinely useful but I'd rather wait until Phase 6b's `RedisDecrBarrier` surfaces a second concrete consumer — let the second implementation drive the actual minimal-Protocol shape rather than guessing what it'll need. Tracked as a follow-up; the docstring now reads:

```
``app_instance`` is a Celery app today (the Protocol leaks this
concrete dependency for the same reason the ``CeleryChordBarrier``
does — Phase 6b's RedisDecrBarrier will narrow this to a smaller
Protocol). The argument is the ``app`` from the producer's task
context and must expose ``.signature(name, kwargs=..., queue=...,
headers=...)``.
```

workers/queue_backend/barrier.py
muhammad-ali-e: **Fixed in `64519b3b7`** — swapped the inline `{FAIRNESS_HEADER_NAME: fairness.to_dict()}` for `fairness.as_header()`, and dropped the now-unused `FAIRNESS_HEADER_NAME` import.

```python
from .fairness import FairnessKey  # was: FAIRNESS_HEADER_NAME, FairnessKey

# ...

# Use ``fairness.as_header()`` — the single wire-shape encoder,
# also used by ``dispatch.py``. Constructing the header dict inline
# here would risk silent divergence from the bare-dispatch path.
fairness_headers = fairness.as_header() if fairness else None
```

Single wire-shape encoder now used by both the `dispatch()` path and the `Barrier.enqueue()` path. A future change to the header layout (e.g. adding `pipeline_priority` defaults or renaming) lands in one place. Thanks for the catch.

workers/shared/workflow/execution/orchestration_utils.py
muhammad-ali-e: **Fixed in `64519b3b7`** (covered by the V2 alignment) — `orchestration_utils.py:41` now reads:

```
Routes through ``CeleryChordBarrier`` — a thin wrapper around
``celery.chord(header)(body)`` that lets Phase 6b swap the
substrate (e.g. to ``RedisDecrBarrier``) without touching this
call site a second time.
```

All phase-label references now consistent across `barrier.py` and `orchestration_utils.py`: Phase 6b = `RedisDecrBarrier` swap, Phase 8 = future `PgBarrier`.

workers/api-deployment/tasks.py
muhammad-ali-e: **Fixed in `64519b3b7`** — corrected variable name in both occurrences. The api-deployment comment now reads:

```python
# 1. Zero ``batch_tasks`` — the barrier short-circuits and
#    returns ``None``. Pre-Barrier this would have called
#    ``chord(empty)(callback)`` which Celery handles by
#    firing the callback immediately with ``[]``. We
#    preserve that contract byte-for-byte by dispatching
#    the callback explicitly with an empty result list, so
#    workflow_execution_status, API result cache, and
#    pipeline notifications all run exactly as they would
#    have pre-Barrier. (Unreachable in practice — upstream
#    requires non-empty ``hash_values_of_files`` — but
#    the defence closes the theoretical gap.)
```

And `test_barrier.py:434` similarly updated to reference `hash_values_of_files`. Thanks for catching the stale name — that comment would have been worse than useless to a future debugger.

**Headline finding (executing-test coverage of the zero-batch branch)**: deferring to a follow-up ticket. Rationale in the commit message — the branch is double-defended (upstream `if not hash_values_of_files: ... early return` at `_run_workflow_api:448` PLUS the defensive branch in the chord call site), unreachable in production, and `TestApiDeploymentZeroFilesContract` already catches contract drift via source-match. An executing test adds ~80 LOC of heavy mocks for an unreachable branch — disproportionate value. If the upstream guard is ever weakened or a new code path creates empty `batch_tasks` with non-empty `hash_values_of_files`, the executing test becomes worth writing.

workers/queue_backend/barrier.py
muhammad-ali-e: **Fixed in `64519b3b7`** — aligned all references to **Phase 6b** for the `RedisDecrBarrier` swap; **Phase 8** is reserved for the eventual `PgBarrier` (SKIP LOCKED, the actual PG Queue substrate). The module docstring at `barrier.py:13` now reads:

```
Lifting both call sites to ``Barrier`` gives Phase 6b a place to land
an alternative implementation (e.g. ``RedisDecrBarrier`` using the labs
``DECR remaining:{exec_id}`` pattern, or — later in Phase 8 — a
``PgBarrier`` using SKIP LOCKED) without touching the call sites a
second time.
```

`orchestration_utils.py:41` updated to match (was "Phase 8 work" → now "Phase 6b"). All references to the chord→Redis-DECR swap consistently say Phase 6b now.

workers/queue_backend/barrier.py
muhammad-ali-e: **Fixed in `64519b3b7`** — added the readily-available context:

```python
except Exception:
    # ``logger.exception`` auto-attaches the traceback — needed
    # for debugging broker outages / serialisation failures
    # at the chord entry point. Add the readily-available
    # context so a Sentry/log triage gets execution/pipeline/
    # callback/queue identifiers alongside the stack.
    execution_id = callback_kwargs.get("execution_id")
    pipeline_id = callback_kwargs.get("pipeline_id")
    logger.exception(
        f"[exec:{execution_id}] [pipeline:{pipeline_id}] "
        f"Failed to enqueue barrier "
        f"(callback={callback_task_name}, queue={callback_queue}, "
        f"header_tasks={len(header_tasks)})"
    )
    raise
```

Matches the format of the `logger.info` messages above it (same `[exec:X] [pipeline:Y]` prefix), so the log line surface stays consistent for whoever's grepping. The exception is still re-raised — `logger.exception` only adds the stack to the WARN/ERROR log path.

vishnuszipstack (approved):

johnyrahul (approved):

PR #2025, opened 2026-06-10
UN-3522 [FEAT] Address Phase 6a Barrier review feedback

muhammad-ali-e:
## Summary

Follow-up to #2024 (Phase 6a — `CeleryChordBarrier` wrapper) addressing the remaining review-feedback threads. Behaviour-preserving; no production wire-shape change beyond what already shipped in #2024.

- **Hoist `FairnessKey` to a local** in `_run_workflow_api` so the chord path and the zero-batch fallback can't drift apart on the slot's shape.
- **Soften "byte-for-byte" wording** on the zero-files dispatch comment: observable end-state is preserved, but the wire is a plain `send_task` with an additive fairness header — not the chord-aggregation primitive. Inline comment also clarifies that `chord_id` on the zero-batch branch is a standalone callback task id (consumers treating it as a chord id should branch on `batches_created == 0`).
- **Load-bearing `BarrierHandle` protocol test**: the previous `MagicMock`-with-`.id` assertion passed identically even if `BarrierHandle` were deleted. Replaced with a structural-typing check that Celery's real `AsyncResult` satisfies the Protocol, plus a `TaskHandle is DispatchHandle is BarrierHandle` identity check so the two handles can't silently drift.
- **Executing call-site fairness contracts**: previously asserted on source-string matches; now `importlib.util` loads `api-deployment/tasks.py` (dash in directory blocks normal import) and drives `_run_workflow_api` with mocks to inspect the actual `FairnessKey` passed at the real call site. Same for `general.tasks._orchestrate_file_processing_general`.
- **Zero-batch executing test**: forces empty batches and asserts the zero-files branch dispatches `process_batch_callback_api` with `args=[[]]` and an `api`-workload `FairnessKey` via `queue_backend.dispatch` (not raw `send_task`).
- **Characterisation canary blind-spot docs**: documents that `test_chord_invocation_lives_only_in_barrier` won't catch a bare `chord(...)` literal outside `barrier.py`; that blind spot is bounded by the separate `test_chord_import_only_in_barrier` canary.
- **Barrier docstring tightening**: byte-identical only when `fairness=None`; with a `FairnessKey` the sole addition is the `x-fairness-key` AMQP header on each header task and the callback.

## Test plan

- [x] `pytest workers/tests/test_barrier.py workers/tests/test_chord_sites_characterisation.py` — 23/23 pass.
- [x] No new public API surface; same `Barrier` Protocol shape as merged in #2024.
- [x] Zero-batch executing test exercises the regression closed in #2024 (pre-Barrier `chord(empty)(callback)` → post-Barrier explicit `dispatch`); confirmed `dispatch.assert_called_once()` with `args=[[]]` and the API fairness slot.
- [x] Pre-commit clean (ruff, ruff-format, pycln, pyupgrade, secret-scan).

## Risk

None beyond #2024 — this PR is comment-tightening, test-strengthening, and a local-variable hoist on a hot path that was already behaviour-equivalent. No new wire-level changes, no new flags, no schema touches.

### DIFF
--- a/workers/api-deployment/tasks.py
+++ b/workers/api-deployment/tasks.py
@@ -697,6 +697,18 @@ def _run_workflow_api(
         )
 
         if not result:
+            # TODO(phase-6b): tighten to ``if result is None:``
+            # when ``RedisDecrBarrier`` lands. The ``Barrier`` Protocol
+            # declares ``None`` as the *sole* "no work enqueued" signal,
+            # so identity-against-None is the principled check. Safe to
+            # defer today because Celery's ``AsyncResult`` has no custom
+            # ``__bool__`` (always truthy), making the truthiness and
+            # identity checks behaviourally equivalent. A future
+            # substrate handle with a falsy-but-not-None ``__bool__``
+            # would be misrouted into the zero-files path under the
+            # current truthiness check — that's the regression this TODO
+            # forecloses, alongside the substrate that motivates it.
+            #
             # Two reasons ``result`` can be falsy:
             # 1. Zero ``batch_tasks`` — the barrier short-circuits and
             #    returns ``None``. Pre-Barrier this would have called
@@ -710,11 +722,15 @@ def _run_workflow_api(
             #    identical (this is a plain ``send_task`` with an
             #    additive fairness header rather than the chord-
             #    aggregation primitive), but the observable end-state
-            #    is equivalent. Unreachable in practice — upstream
-            #    requires non-empty ``hash_values_of_files`` — but
-            #    the defence closes the theoretical gap.
+            #    is equivalent. Effectively unreachable — the upstream
+            #    ``if not hash_values_of_files:`` early return plus
+            #    valid ``FileHashData`` inputs mean ``_get_file_batches``
+            #    always yields >=1 batch — but the defence closes the
+            #    theoretical gap (e.g. all entries dropped because none
+            #    are ``FileHashData`` / ``dict``, which would still
+            #    bypass the upstream guard).
             # 2. ``batch_tasks`` is non-empty but the barrier returned
-            #    falsy for some other reason — a genuine queue failure.
+            #    ``None`` for some other reason — a genuine queue failure.
             if not batch_tasks:
                 logger.info(
                     f"Execution {execution_id} orchestrated with zero "
--- a/workers/api-deployment/tasks.py
+++ b/workers/api-deployment/tasks.py
@@ -677,6 +677,12 @@ def _run_workflow_api(
         # The header tasks + callback now carry the fairness slot for
         # downstream PG Queue routing — closes the chord-fairness gap noted
         # in Phase 5.1.
+        # Hoist FairnessKey to a local so the chord path and the
+        # zero-batch fallback can't drift apart on the slot's shape.
+        api_fairness = FairnessKey(
+            org_id=str(schema_name),
+            workload_type=WorkloadType.API,
+        )
         result = WorkflowOrchestrationUtils.create_chord_execution(
             batch_tasks=batch_tasks,
             callback_task_name="process_batch_callback_api",
@@ -687,10 +693,7 @@ def _run_workflow_api(
             },
             callback_queue=file_processing_callback_queue,
             app_instance=app,
-            fairness=FairnessKey(
-                org_id=str(schema_name),
-                workload_type=WorkloadType.API,
-            ),
+            fairness=api_fairness,
         )
 
         if not result:
@@ -699,14 +702,17 @@ def _run_workflow_api(
             #    returns ``None``. Pre-Barrier this would have called
             #    ``chord(empty)(callback)`` which Celery handles by
             #    firing the callback immediately with ``[]``. We
-            #    preserve that contract byte-for-byte by dispatching
-            #    the callback explicitly with an empty result list, so
+            #    preserve that observable end-state by dispatching the
+            #    callback explicitly with an empty result list, so
             #    workflow_execution_status, API result cache, and
             #    pipeline notifications all run exactly as they would
-            #    have pre-Barrier. (Unreachable in practice — upstream
+            #    have pre-Barrier. The new wire isn't literally byte-
+            #    identical (this is a plain ``send_task`` with an
+            #    additive fairness header rather than the chord-
+            #    aggregation primitive), but the observable end-state
+            #    is equivalent. Unreachable in practice — upstream
             #    requires non-empty ``hash_values_of_files`` — but
-            #    the defence
-            #    closes the theoretical gap.)
+            #    the defence closes the theoretical gap.
             # 2. ``batch_tasks`` is non-empty but the barrier returned
             #    falsy for some other reason — a genuine queue failure.
             if not batch_tasks:
@@ -730,10 +736,7 @@ def _run_workflow_api(
                         "organization_id": str(schema_name),
                     },
                     queue=file_processing_callback_queue,
-                    fairness=FairnessKey(
-                        org_id=str(schema_name),
-                        workload_type=WorkloadType.API,
-                    ),
+                    fairness=api_fairness,
                 )
                 return {
                     "status": "orchestrated",
@@ -743,6 +746,13 @@ def _run_workflow_api(
                     "files_processed": total_files,
                     "files_from_cache": len(cached_results),
                     "batches_created": 0,
+                    # ``chord_id`` here is the standalone callback
+                    # task id (no chord; the dispatch above is a
+                    # direct ``send_task``). Same response-shape key
+                    # as the normal path; semantically a task id
+                    # rather than a chord id. Log/metrics consumers
+                    # treating ``chord_id`` as "chord identifier"
+                    # should branch on ``batches_created == 0``.
                     "chord_id": callback_result.id,
                     "cached_results": list(cached_results.keys())
                     if cached_results
--- a/workers/tests/test_barrier.py
+++ b/workers/tests/test_barrier.py
@@ -29,6 +29,7 @@ The single ``chord(...)`` call still lives in
 from __future__ import annotations
 
 import contextlib
+from dataclasses import dataclass
 from unittest.mock import MagicMock, patch
 
 import pytest
@@ -428,44 +429,65 @@ def _load_api_deployment_tasks():
     return module
 
 
-def _run_workflow_api_with_mocks(
+@dataclass
+class _WorkflowApiMocks:
+    """Spy bundle from ``_setup_workflow_api_mocks``.
+
+    All attributes are bound *before* the production function runs,
+    so a test asserting on spies after a raising call (e.g. the
+    queue-failure branch) can still inspect them.
+    """
+
+    api_tasks: object
+    api_client: MagicMock
+    create_chord: MagicMock
+    dispatch: MagicMock
+    decisions_helper: MagicMock
+
+
+# Three production branches under test, one knob. Replacing
+# ``force_empty_batches`` + ``force_falsy_chord_with_batches`` (two
+# bools encoding three states with a runtime mutual-exclusivity
+# guard) — the illegal ``(True, True)`` combination is now
+# unrepresentable.
+_VALID_CHORD_OUTCOMES = ("success", "empty_batches", "queue_failure")
+
+
+def _setup_workflow_api_mocks(
     monkeypatch,
     *,
-    hash_values_of_files: dict | None = None,
-    force_empty_batches: bool = False,
-    force_falsy_chord_with_batches: bool = False,
+    chord_outcome: str = "success",
     manual_review_required: bool = False,
-    review_decisions_helper: MagicMock | None = None,
-):
-    """Drive ``_run_workflow_api`` end-to-end with the minimum mocks
-    needed to exercise the chord-dispatch + zero-batch + queue-failure
-    branches.
+    num_batches: int = 1,
+) -> _WorkflowApiMocks:
+    """Setup mocks; caller drives ``_run_workflow_api``.
 
-    Knobs:
+    Use this when the test needs to assert on spies *after* the
+    production function raises (e.g. the queue-failure branch).
+    Tests that don't need post-raise assertions can use the thin
+    convenience wrapper ``_run_workflow_api_with_mocks``.
 
-    - ``force_empty_batches`` — ``_get_file_batches`` returns ``[]``
-      and ``create_chord_execution`` returns ``None``. Exercises the
-      zero-batch defensive dispatch fallback.
-    - ``force_falsy_chord_with_batches`` — ``_get_file_batches``
-      returns a non-empty list but ``create_chord_execution`` returns
-      ``None`` anyway. Exercises the "genuine queue failure" branch
-      that raises and maps to ``ExecutionStatus.ERROR``. Mutually
-      exclusive with ``force_empty_batches``.
-    - ``manual_review_required`` — ``_create_file_data`` returns a
-      file_data whose ``manual_review_config["review_required"]`` is
-      ``True``, exercising the manual-review decision branch. The
-      caller can supply ``review_decisions_helper`` (a ``MagicMock``)
-      to track invocations of
-      ``_calculate_manual_review_decisions_for_batch_api``.
+    ``chord_outcome`` controls which production branch fires:
 
-    Returns ``(result, mock_create_chord, mock_dispatch)`` so callers
-    can assert on the response shape and on which dispatch path fired.
+    - ``"success"`` — non-empty batches, truthy chord handle. Normal
+      chord path completes.
+    - ``"empty_batches"`` — ``_get_file_batches`` yields ``[]``,
+      chord returns ``None``. Zero-batch defensive dispatch fires.
+    - ``"queue_failure"`` — ``_get_file_batches`` yields non-empty
+      batches, chord returns ``None`` anyway. Production raises so
+      the outer handler can map to ``ExecutionStatus.ERROR``.
+
+    ``num_batches`` controls how many batches ``_get_file_batches``
+    returns when ``chord_outcome != "empty_batches"`` — set >1 to
+    exercise per-batch multiplicity contracts (e.g. the manual-
+    review decision helper is invoked once per batch).
     """
-    if force_empty_batches and force_falsy_chord_with_batches:
+    if chord_outcome not in _VALID_CHORD_OUTCOMES:
         raise AssertionError(
-            "force_empty_batches and force_falsy_chord_with_batches are "
-            "mutually exclusive — pick the branch under test"
+            f"chord_outcome must be one of {_VALID_CHORD_OUTCOMES}, "
+            f"got {chord_outcome!r}"
         )
+
     api_tasks = _load_api_deployment_tasks()
     api_client = MagicMock(name="api_client")
     api_client.get_workflow_execution.return_value = MagicMock(
@@ -488,25 +510,27 @@ def _run_workflow_api_with_mocks(
         api_tasks, "_log_api_batch_creation_statistics", lambda **kwargs: None
     )
 
-    # Always patch ``_get_file_batches`` — the real implementation
+    # ``_get_file_batches`` always patched — the real implementation
     # filters non-``FileHashData``/``dict`` entries, which would drop
-    # our ``MagicMock`` values and silently land in the zero-batch
-    # branch even when ``force_empty_batches=False``. Patching here
-    # gives the caller deterministic control over the chord vs
-    # zero-batch path via the knobs.
+    # ``MagicMock`` values and silently route into the zero-batch
+    # branch. Patching here gives the caller deterministic control
+    # over both the branch under test and the batch count.
+    if chord_outcome == "empty_batches":
+        batches_to_return: list[list[str]] = []
+    else:
+        batches_to_return = [
+            [f"file_in_batch_{i}"] for i in range(num_batches)
+        ]
     monkeypatch.setattr(
-        api_tasks,
-        "_get_file_batches",
-        lambda **kwargs: [] if force_empty_batches else [["file_1_in_batch"]],
+        api_tasks, "_get_file_batches", lambda **kwargs: batches_to_return
     )
+
     # Neutralise the per-batch helpers so ``batch_tasks`` is non-empty
-    # in the chord-path case without us needing to construct real
-    # ``FileBatchData`` objects.
+    # without constructing real ``FileBatchData`` objects.
     mock_file_data = MagicMock(name="file_data")
-    # ``.manual_review_config.get("review_required", False)`` is a
-    # real dict so we can flip it via the ``manual_review_required``
-    # knob — MagicMocks would return truthy by default from .get(),
-    # which would force the manual-review branch.
+    # Real dict (not MagicMock) so ``.get("review_required", False)``
+    # returns the requested bool — MagicMocks would return truthy by
+    # default from ``.get()``, forcing the manual-review branch.
     mock_file_data.manual_review_config = {
         "review_required": manual_review_required,
     }
@@ -516,28 +540,31 @@ def _run_workflow_api_with_mocks(
     monkeypatch.setattr(
         api_tasks, "_create_batch_data", lambda **kwargs: {"batch": "data"}
     )
+    # Always create the decisions-helper spy and install it; tests
+    # that don't flip ``manual_review_required`` simply never see it
+    # invoked. Exposing it on the spy bundle removes the previous
+    # caller-supplied/fallback dead branch.
+    decisions_helper = MagicMock(
+        name="_calculate_manual_review_decisions_for_batch_api",
+        return_value=[False],
+    )
     if manual_review_required:
-        # When the caller flips the flag, install a spy on the
-        # manual-review decision helper so we can assert it was
-        # invoked. Default arg ensures shape compatibility with the
-        # real helper (one bool per file).
-        helper = review_decisions_helper or MagicMock(
-            name="_calculate_manual_review_decisions_for_batch_api",
-            return_value=[False],
-        )
         monkeypatch.setattr(
-            api_tasks, "_calculate_manual_review_decisions_for_batch_api", helper
+            api_tasks,
+            "_calculate_manual_review_decisions_for_batch_api",
+            decisions_helper,
         )
 
     mock_create_chord = MagicMock(name="create_chord_execution")
-    # When the chord path should fire normally, return a truthy
-    # handle. When forced empty OR forced falsy-with-batches, return
-    # ``None`` — the production code's ``if result is None:`` branch
-    # then distinguishes the two cases via ``batch_tasks`` length.
-    if force_empty_batches or force_falsy_chord_with_batches:
-        mock_create_chord.return_value = None
-    else:
+    # Truthy handle on the success path; ``None`` on the two failure
+    # paths. Production's ``if not result:`` branch (today; a future
+    # phase-6b PR may tighten to ``if result is None:`` — see the
+    # TODO at the call site in ``api-deployment/tasks.py``) then
+    # distinguishes the two ``None`` cases via ``batch_tasks`` length.
+    if chord_outcome == "success":
         mock_create_chord.return_value = MagicMock(id="chord-result-id")
+    else:
+        mock_create_chord.return_value = None
     monkeypatch.setattr(
         api_tasks.WorkflowOrchestrationUtils,
         "create_chord_execution",
@@ -549,11 +576,42 @@ def _run_workflow_api_with_mocks(
     mock_dispatch = MagicMock(name="dispatch", return_value=mock_dispatch_result)
     monkeypatch.setattr(api_tasks, "dispatch", mock_dispatch)
 
+    return _WorkflowApiMocks(
+        api_tasks=api_tasks,
+        api_client=api_client,
+        create_chord=mock_create_chord,
+        dispatch=mock_dispatch,
+        decisions_helper=decisions_helper,
+    )
+
+
+def _run_workflow_api_with_mocks(
+    monkeypatch,
+    *,
+    hash_values_of_files: dict | None = None,
+    chord_outcome: str = "success",
+    manual_review_required: bool = False,
+    num_batches: int = 1,
+):
+    """Convenience wrapper around ``_setup_workflow_api_mocks``.
+
+    Setup + run; returns ``(result, create_chord, dispatch)`` for
+    tests where ``_run_workflow_api`` doesn't raise. Tests that need
+    to assert on spies after a raising call should use
+    ``_setup_workflow_api_mocks`` directly so the spies are bound
+    before the raise.
+    """
+    mocks = _setup_workflow_api_mocks(
+        monkeypatch,
+        chord_outcome=chord_outcome,
+        manual_review_required=manual_review_required,
+        num_batches=num_batches,
+    )
     files = hash_values_of_files if hash_values_of_files is not None else {
         "f1": MagicMock(name="FileHashData_f1"),
     }
-    result = api_tasks._run_workflow_api(
-        api_client=api_client,
+    result = mocks.api_tasks._run_workflow_api(
+        api_client=mocks.api_client,
         schema_name="org_test",
         workflow_id="wf-1",
         execution_id="exec-1",
@@ -564,7 +622,7 @@ def _run_workflow_api_with_mocks(
         use_file_history=False,
         task_id="task-1",
     )
-    return result, mock_create_chord, mock_dispatch
+    return result, mocks.create_chord, mocks.dispatch
 
 
 class TestCallSiteFairnessContracts:
@@ -584,14 +642,14 @@ class TestCallSiteFairnessContracts:
         """The chord call site in ``_run_workflow_api`` must pass
         ``fairness=FairnessKey(org_id=str(schema_name),
         workload_type=WorkloadType.API)`` to ``create_chord_execution``."""
-        # ``_run_workflow_api_with_mocks(force_empty_batches=False)``
-        # patches ``_get_file_batches`` to return a non-empty list so
-        # the chord path actually fires (the helper's patches are
-        # documented in ``_run_workflow_api_with_mocks``).
+        # ``chord_outcome="success"`` patches ``_get_file_batches``
+        # to return a non-empty list and ``create_chord_execution``
+        # to return a truthy handle so the chord path actually fires
+        # (helper patches documented in ``_setup_workflow_api_mocks``).
         _result, mock_create_chord, mock_dispatch = _run_workflow_api_with_mocks(
             monkeypatch,
             hash_values_of_files={"f1": MagicMock(name="file_1")},
-            force_empty_batches=False,
+            chord_outcome="success",
         )
         # ``create_chord_execution`` MUST be called with the right
         # fairness — a refactor that drops ``fairness=`` or swaps the
@@ -728,7 +786,7 @@ class TestApiDeploymentZeroFilesContract:
         result, _create_chord, mock_dispatch = _run_workflow_api_with_mocks(
             monkeypatch,
             hash_values_of_files={"f1": MagicMock(name="file_1")},
-            force_empty_batches=True,
+            chord_outcome="empty_batches",
         )
 
         # Dispatch was called exactly once, with the chord-empty
@@ -780,11 +838,26 @@ class TestApiDeploymentQueueFailureContract:
         ``batch_tasks``, ``_run_workflow_api`` must raise — the outer
         task handler then maps this to ``ExecutionStatus.ERROR`` for
         the pipeline status update."""
-        with pytest.raises(Exception, match=r"(?i)queue|failed|chord"):
-            _run_workflow_api_with_mocks(
-                monkeypatch,
+        # Pin the exact production message rather than a loose
+        # ``queue|failed|chord`` regex — the loose form would also
+        # match a misconfigured-harness ``AssertionError`` or any
+        # setup error whose message contains "failed", masking a
+        # harness misconfiguration as a passing assertion.
+        mocks = _setup_workflow_api_mocks(
+            monkeypatch, chord_outcome="queue_failure"
+        )
+        with pytest.raises(Exception, match=r"Failed to queue execution task exec-1"):
+            mocks.api_tasks._run_workflow_api(
+                api_client=mocks.api_client,
+                schema_name="org_test",
+                workflow_id="wf-1",
+                execution_id="exec-1",
                 hash_values_of_files={"f1": MagicMock(name="file_1")},
-                force_falsy_chord_with_batches=True,
+                scheduled=False,
+                execution_mode=None,
+                pipeline_id="pipe-1",
+                use_file_history=False,
+                task_id="task-1",
             )
 
     def test_falsy_chord_with_non_empty_batches_does_not_dispatch_fallback(
@@ -794,19 +867,43 @@ class TestApiDeploymentQueueFailureContract:
         the genuine-queue-failure branch — that's reserved for the
         zero-batch defence. A future refactor that routed both
         falsy-result cases through the same fallback would silently
-        lose the ERROR signal."""
-        with contextlib.suppress(Exception):
-            _result, _create_chord, mock_dispatch = (
-                _run_workflow_api_with_mocks(
-                    monkeypatch,
-                    hash_values_of_files={"f1": MagicMock(name="file_1")},
-                    force_falsy_chord_with_batches=True,
-                )
+        lose the ERROR signal.
+
+        Previous revision had this assertion inside
+        ``contextlib.suppress(Exception)`` and bound ``mock_dispatch``
+        from the function's tuple return — which never executed because
+        the production code raised before the return. ``AssertionError``
+        is an ``Exception`` so the suppress would have swallowed any
+        failing assert too. Both defects fixed by:
+
+        1. Setting up mocks via ``_setup_workflow_api_mocks`` so spies
+           are bound *before* ``_run_workflow_api`` runs.
+        2. Asserting on ``mocks.dispatch.called`` *outside* the
+           ``pytest.raises`` block so an assertion failure surfaces.
+        """
+        mocks = _setup_workflow_api_mocks(
+            monkeypatch, chord_outcome="queue_failure"
+        )
+        with pytest.raises(Exception, match=r"Failed to queue execution task exec-1"):
+            mocks.api_tasks._run_workflow_api(
+                api_client=mocks.api_client,
+                schema_name="org_test",
+                workflow_id="wf-1",
+                execution_id="exec-1",
+                hash_values_of_files={"f1": MagicMock(name="file_1")},
+                scheduled=False,
+                execution_mode=None,
+                pipeline_id="pipe-1",
+                use_file_history=False,
+                task_id="task-1",
             )
-            assert not mock_dispatch.called, (
-                "Genuine queue failure must raise, not silently dispatch the "
-                "zero-batch fallback callback"
-            )
+        # The fallback dispatch was monkeypatched on the module and
+        # survives the raise; assert outside the ``pytest.raises``
+        # block so an assertion failure isn't swallowed.
+        assert not mocks.dispatch.called, (
+            "Genuine queue failure must raise, not silently dispatch the "
+            "zero-batch fallback callback"
+        )
 
 
 class TestApiDeploymentManualReviewContract:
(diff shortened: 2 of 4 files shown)


workers/tests/test_barrier.py
```diff
@@ -358,103 +376,259 @@ def test_create_chord_execution_delegates_to_module_barrier(self):
...
+        # fairness — a refactor that drops ``fairness=`` or swaps the
+        # workload type fails here loudly.
+        assert mock_create_chord.called, (
+            "_run_workflow_api did not call create_chord_execution"
+        )
+        fairness_kwarg = mock_create_chord.call_args.kwargs.get("fairness")
+        assert fairness_kwarg is not None, "fairness= kwarg missing"
+        assert fairness_kwarg == FairnessKey(
+            org_id="org_test", workload_type=WorkloadType.API
+        )
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 87509ec98.

The real `_get_file_batches` was filtering out the `MagicMock` values (not `FileHashData` or `dict`), so `batch_tasks` came back empty and the test was silently exercising the zero-batch fallback. The fairness assertion would have passed via the fallback's own `fairness=` kwarg even if the primary chord call had dropped `fairness=` entirely — exactly the regression you described.

Patched `_get_file_batches` / `_create_file_data` / `_create_batch_data` in the shared `_run_workflow_api_with_mocks` helper so `force_empty_batches=False` actually returns a non-empty list, and made `create_chord_execution` return a truthy mock so the fallback can't fire. Added two new assertions:

- `assert not mock_dispatch.called` — proves the zero-batch fallback did NOT fire
- `assert batch_tasks` — proves the chord call received a non-empty header list

Now a future refactor that drops `fairness=` only on the chord path (keeping it on the fallback) fails this test loudly.

workers/tests/test_barrier.py
```diff
@@ -82,17 +82,35 @@ def test_celery_chord_barrier_satisfies_protocol(self):
...
+        # Structural-typing pin: every attribute the Protocol declares
+        # must exist on ``AsyncResult``. Currently just ``id``; this
+        # loop is future-proof for additional required attributes.
+        required_attrs = ("id",)
+        missing = [a for a in required_attrs if not hasattr(AsyncResult, a)]
+        assert missing == [], (
+            f"AsyncResult is missing required BarrierHandle attribute(s): "
+            f"{missing} — a refactor of BarrierHandle / TaskHandle has "
+            f"silently broken the Celery substrate."
+        )
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 87509ec98.

You're right — `hasattr(AsyncResult, "id")` only inspects the class object. `AsyncResult.id` happens to be a `@property` today so it passes, but a future `BarrierHandle` extension with an instance-only field (e.g. `self.foo = bar` with no class-level descriptor) would have slipped through.

Switched to an instance-level check:

```python
async_result_instance = AsyncResult("placeholder-task-id")
missing = [a for a in required_attrs if not hasattr(async_result_instance, a)]
```

Now genuinely future-proof for additional required attributes regardless of whether they're class-level descriptors or `__init__`-assigned.

workers/tests/test_barrier.py
```diff
@@ -358,103 +376,259 @@ def test_create_chord_execution_delegates_to_module_barrier(self):
...
-        assert '"process_batch_callback_api"' in text  # callback fired
-        assert "args=[[]]" in text  # body=[] matches chord-empty semantic
-        assert '"batches_created": 0' in text
+
+        # Response shape: orchestrated with zero batches, callback
+        # task id surfaced as chord_id (semantically a task id, not a
+        # chord id — see the inline comment at the call site).
+        assert result["status"] == "orchestrated"
+        assert result["batches_created"] == 0
+        assert result["chord_id"] == "callback-task-id"
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 87509ec98.

Added the missing assertion:

```python
assert call.kwargs.get("queue") == "api_file_processing_callback"
```

Now a refactor that routes the zero-batch callback to the wrong queue (or drops the `queue=` kwarg entirely) fails this test loudly.

workers/tests/test_barrier.py
```diff
@@ -82,17 +82,43 @@ def test_celery_chord_barrier_satisfies_protocol(self):
...
+        from celery.result import AsyncResult
+
+        # Structural-typing pin: every attribute the Protocol declares
+        # must exist on a real ``AsyncResult`` *instance*. Checking on
+        # the class would only see ``@property`` descriptors and skip
+        # attributes assigned inside ``__init__`` (e.g. ``self.id =
+        # task_id`` with no class-level descriptor) — so a future
+        # ``BarrierHandle`` extension with an instance-only field
+        # could slip through. Currently just ``id``; the loop is
+        # future-proof for additional required attributes.
```
vishnuszipstack: **[Critical · Comment Analyzer] Misleading comment — not actually "future-proof".**

The loop iterates a hardcoded `required_attrs = ("id",)`; it does not introspect `TaskHandle.__annotations__`. If a future refactor adds a required attribute to `BarrierHandle`/`TaskHandle`, this test keeps passing silently until someone hand-edits the tuple — the exact silent break the docstring (lines 87-93) says it prevents. The loop *structure* tolerates more entries, but it isn't future-proof; it's manually maintained.

Suggested rewrite:
```python
# Currently just ``id``. NOTE: maintained by hand — if a required
# attribute is added to TaskHandle, add it here too (this test won't
# auto-detect it from the Protocol annotations).
```

workers/tests/test_barrier.py
```diff
@@ -82,17 +82,43 @@ def test_celery_chord_barrier_satisfies_protocol(self):
...
+        in ``api-deployment/tasks.py``'s response without any test
+        catching it. Tautological MagicMock-with-.id-attribute
+        tests pass identically even if ``BarrierHandle`` were
+        deleted; this one is load-bearing.
+        """
+        from celery.result import AsyncResult
+
+        # Structural-typing pin: every attribute the Protocol declares
+        # must exist on a real ``AsyncResult`` *instance*. Checking on
+        # the class would only see ``@property`` descriptors and skip
```
vishnuszipstack: **[Important · Comment Analyzer] Instance-vs-class rationale is factually wrong for `AsyncResult`.**

Celery's `AsyncResult` declares `id = None` as a *class-level* attribute (and also `self.id = id` in `__init__`). So `hasattr(AsyncResult, "id")` is already `True`, `id` is **not** a `@property`, and **not** instance-only — checking on the class would not miss it. Testing the instance is still fine (it's what production passes around), but the stated reason doesn't apply to the class under test.

Suggested rewrite:
```python
# Assert against a real AsyncResult *instance* — that's what every
# production call site receives and logs ``.id`` from. (We don't rely
# on whether the attr is class-level, a property, or set in __init__.)
```

workers/tests/test_barrier.py
```diff
@@ -82,17 +82,43 @@ def test_celery_chord_barrier_satisfies_protocol(self):
...
+        # attributes assigned inside ``__init__`` (e.g. ``self.id =
+        # task_id`` with no class-level descriptor) — so a future
+        # ``BarrierHandle`` extension with an instance-only field
+        # could slip through. Currently just ``id``; the loop is
+        # future-proof for additional required attributes.
+        async_result_instance = AsyncResult("placeholder-task-id")
+        required_attrs = ("id",)
+        missing = [
+            a for a in required_attrs if not hasattr(async_result_instance, a)
```
vishnuszipstack, line 117: **[Minor · Type Design] Pin the declared type, not just attribute presence.**

The Protocol declares `id: str` but the test only checks `hasattr(..., "id")`. A future substrate could satisfy `hasattr` with `id=None`/a UUID and slip past the contract. Tighten:
```python
assert isinstance(async_result_instance.id, str)
```

workers/api-deployment/tasks.py
```diff
@@ -699,14 +702,17 @@ def _run_workflow_api(
...
+            #    preserve that observable end-state by dispatching the
+            #    callback explicitly with an empty result list, so
             #    workflow_execution_status, API result cache, and
             #    pipeline notifications all run exactly as they would
-            #    have pre-Barrier. (Unreachable in practice — upstream
+            #    have pre-Barrier. The new wire isn't literally byte-
+            #    identical (this is a plain ``send_task`` with an
+            #    additive fairness header rather than the chord-
+            #    aggregation primitive), but the observable end-state
+            #    is equivalent. Unreachable in practice — upstream
```
vishnuszipstack: **[Important · Comment Analyzer] "upstream requires non-empty `hash_values_of_files`" overstates the guarantee.**

Non-empty `hash_values_of_files` does not guarantee non-empty `batch_tasks`: `_get_file_batches` drops any entry that is neither `FileHashData` nor `dict` via `continue`, so malformed inputs can yield zero batches with non-empty files (the test helper's own docstring relies on this). The branch is unreachable *in practice* because real inputs are valid `FileHashData`, not simply because the dict is non-empty. The V13 docstring at test_barrier.py:649 phrases it more carefully — match it:
```python
#    is equivalent. Effectively unreachable — the upstream
#    ``if not hash_values_of_files:`` early return plus valid
#    FileHashData inputs mean batching always yields >=1 batch —
#    but the defence closes the theoretical gap.
```

vishnuszipstack (commented):
## Automated PR Review — PR Review Toolkit (6 agents)

Comment Analyzer · PR Test Analyzer · Silent Failure Hunter · Type Design Analyzer · Code Reviewer · Code Simplifier.

All 14 tests pass; the core change (converting source-string-match tests to executing tests, and the `api_fairness` local hoist) is sound. No Critical *code* defects. Findings below are inline; each tagged with `[severity · agent]`. The single Critical is a misleading comment, not a code bug.

workers/api-deployment/tasks.py
```diff
@@ -687,10 +693,7 @@ def _run_workflow_api(
...
             callback_queue=file_processing_callback_queue,
             app_instance=app,
-            fairness=FairnessKey(
-                org_id=str(schema_name),
-                workload_type=WorkloadType.API,
-            ),
+            fairness=api_fairness,
         )
 
         if not result:
```
vishnuszipstack, line 699: **[Minor · Type Design] Match the Barrier Protocol's `None`-specific invariant.**

The `Barrier` contract is that `None` is the *sole* signal of "no work enqueued"; all failures raise. This branch uses a truthiness check `if not result:`, so a handle whose `__bool__` is falsy but isn't `None` would be misrouted into the zero-files path. In practice the inner `if not batch_tasks:` disambiguates, so this is defence-in-depth — but consider `if result is None:` to mirror the stated invariant exactly.

workers/api-deployment/tasks.py
```diff
@@ -699,14 +702,17 @@ def _run_workflow_api(
...
+            #    have pre-Barrier. The new wire isn't literally byte-
+            #    identical (this is a plain ``send_task`` with an
+            #    additive fairness header rather than the chord-
+            #    aggregation primitive), but the observable end-state
+            #    is equivalent. Unreachable in practice — upstream
             #    requires non-empty ``hash_values_of_files`` — but
-            #    the defence
-            #    closes the theoretical gap.)
+            #    the defence closes the theoretical gap.
             # 2. ``batch_tasks`` is non-empty but the barrier returned
```
vishnuszipstack, line 732: **[Important · PR Test Analyzer] The "genuine queue failure" branch (case 2) has no executing coverage.**

When `batch_tasks` is non-empty but `create_chord_execution` returns falsy, the code raises and the outer handler maps it to `ExecutionStatus.ERROR` (tasks.py:765-767, 786-794). But `_run_workflow_api_with_mocks` hard-couples the two knobs — `force_empty_batches` makes `_get_file_batches` return `[]` *and* `create_chord` return `None` together — so there's no way to produce "non-empty batches + falsy chord result." This is the higher-severity sub-branch (a broker failure mapping to ERROR is real production behavior, not the unreachable zero-files defence).

Suggested: add a `force_falsy_chord_with_batches` knob to the helper and a test asserting `pytest.raises(match="Failed to queue")` and that the status update fired with `ExecutionStatus.ERROR`.

workers/api-deployment/tasks.py
```diff
@@ -743,6 +746,13 @@ def _run_workflow_api(
                     "files_processed": total_files,
                     "files_from_cache": len(cached_results),
                     "batches_created": 0,
+                    # ``chord_id`` here is the standalone callback
```
vishnuszipstack, line 765: **[Minor · Silent Failure Hunter] Prefer a machine-readable discriminator over a comment-only `chord_id` overload.**

The zero-batch path stuffs a standalone `send_task` id into `chord_id` (a key that means "chord identifier" everywhere else) and tells consumers via comment to branch on `batches_created == 0`. A comment isn't enforcement — a metrics pipeline resolving `chord_id` as a chord in the result backend would mis-track silently. Cheap additive fix:
```python
"callback_task_id": callback_result.id,
"chord_id": None,  # no chord on the zero-batch path
```
or add `"chord_id_is_task_id": True`.

workers/tests/test_barrier.py
```diff
@@ -358,103 +384,311 @@ def test_create_chord_execution_delegates_to_module_barrier(self):
...
+        lambda **kwargs: [] if force_empty_batches else [["file_1_in_batch"]],
+    )
+    # Neutralise the per-batch helpers so ``batch_tasks`` is non-empty
+    # in the chord-path case without us needing to construct real
+    # ``FileBatchData`` objects.
+    mock_file_data = MagicMock(name="file_data")
+    # ``.manual_review_config.get("review_required", False)`` must
+    # return a falsy value or the loop dives into the manual-review
+    # branch and tries to call ``_calculate_manual_review_decisions_for_batch_api``.
+    mock_file_data.manual_review_config = {"review_required": False}
```
vishnuszipstack: **[Important · PR Test Analyzer] Manual-review decision branch is deliberately bypassed and untested.**

Setting `manual_review_config = {"review_required": False}` skips `_calculate_manual_review_decisions_for_batch_api` (tasks.py:643-651), which mutates `file_data.manual_review_config["file_decisions"]` and feeds the batch data that rides the chord. None of the new executing tests cover `review_required: True`. Consider a variant with the flag true, patching that helper and asserting it's called once per batch and that the chord still fires with the expected fairness. (Downgrade to Minor if an existing integration test already covers it.)

workers/tests/test_barrier.py
```diff
@@ -358,103 +384,311 @@ def test_create_chord_execution_delegates_to_module_barrier(self):
...
+                workflow_id="wf-1",
+                execution_id="exec-1",
+                source_files={"f1": MagicMock(name="file_1")},
+                pipeline_id="pipe-1",
+                scheduled=False,
+                execution_mode=None,
+                use_file_history=False,
+                organization_id="org_test",
+            )
+        except Exception:
```
vishnuszipstack: **[Minor · Silent Failure Hunter] Narrow this bare `except Exception: pass`.**

The following `assert mock_create_chord.called` does guard the "chord never reached" case (verified), but the bare swallow can also hide a brand-new `TypeError`/`AttributeError` from a future refactor and would tolerate a regression that starts raising on the happy path *after* a correct chord call. Use `contextlib.suppress(Exception)` for clearer intent, and ideally restrict to the specific exception the mocked `api_client` status path raises. Also add `assert mock_create_chord.call_args.kwargs.get("batch_tasks")` — the api-deployment sibling has this non-empty pin; the general test doesn't.

workers/tests/test_barrier.py
```diff
@@ -358,103 +384,311 @@ def test_create_chord_execution_delegates_to_module_barrier(self):
...
-    byte-for-byte: ``workflow_execution_status`` updates, API result
-    caching, and pipeline notifications all run exactly as they would
-    have pre-Barrier. Unreachable in practice (upstream guarantees
-    non-empty ``hash_values_of_files``) but this test pins the defensive
-    contract so a future refactor doesn't silently regress.
+    """Executing pin for the api-deployment zero-batch handler.
+
+    Originally a source-string-match test (Vishnu's V13: a Critical
+    finding — string-match would pass even if the real dispatch
+    branch were broken, because the matched substrings appear in
```
vishnuszipstack: **[Minor · Comment Analyzer] V13 rationale is imprecise.**

The core point is right (a source-string match never executes the branch, so it can't prove it works). But the *reason* given — "matched substrings appear in surrounding comments" — is mostly inaccurate: the old asserted substrings (`args=[[]]`, `process_batch_callback_api`, `if not batch_tasks:`) live in real code, and `process_batch_callback_api` also appears on the chord path. Suggest:
```
# a source-string match never executes the branch, so it would pass
# even if the dispatch branch were deleted — the matched tokens also
# appear on the chord path / in comments.
```

workers/tests/test_chord_sites_characterisation.py
```diff
@@ -274,9 +274,14 @@ def test_chord_invocation_lives_only_in_barrier(self):
         but rejects docstring / comment mentions like ``the chord(...)
         primitive``.
 
-        Comment / docstring lines are also explicitly skipped as
-        belt-and-suspenders against a future call site that lacks the
-        ``=`` prefix (e.g. a return-value-discarded ``chord(...)``).
+        **Known blind spot**: a return-value-discarded ``chord(batch)(cb)``
+        at column 0 (no assignment) would slip past this regex AND the
+        comment-line filter. That case is bounded by the separate
+        ``test_chord_import_only_in_barrier`` canary — any stray call
```
vishnuszipstack: **[Minor · Comment Analyzer] Mitigation slightly over-claims completeness.**

`test_chord_import_only_in_barrier` matches `^\s*from\s+celery\s+import\s+chord\b`. A stray call site could still import via `from celery import chord as c`, or `import celery` + `celery.chord(...)`, evading **both** canaries — so "any stray call site still needs a `from celery import chord` somewhere" isn't strictly true. Suggest noting the canary catches the common `from celery import chord` form; aliased / `import celery` access would evade both.

workers/tests/test_barrier.py
```diff
@@ -358,103 +384,311 @@ def test_create_chord_execution_delegates_to_module_barrier(self):
...
+    fails loudly.
     """
 
-    def test_api_deployment_zero_files_dispatches_callback_with_empty_list(self):
-        import inspect
-        import pathlib
+    def test_zero_batch_branch_dispatches_callback_with_empty_list(
+        self, monkeypatch
+    ):
+        result, _create_chord, mock_dispatch = _run_workflow_api_with_mocks(
```
vishnuszipstack, line 786: **[Minor · Code Simplifier] Redundant `hash_values_of_files` arg.**

`_run_workflow_api_with_mocks` already defaults `hash_values_of_files` to `{"f1": MagicMock(...)}` when `None`, and these tests never assert on the dict's contents (org id is hardcoded `"org_test"`). Both the chord-path test and this zero-batch test pass an equivalent value explicitly — you can drop the `hash_values_of_files=` line and rely on the default. (Optional; keep one explicit call if you prefer to document intent.)

vishnuszipstack (approved):

workers/tests/test_barrier.py
muhammad-ali-e: Fixed in 4a2c9b5ce. You're right — the loop tolerated more entries but didn't introspect `TaskHandle.__annotations__`, so calling it future-proof was misleading. Rewrote the comment to be honest: `required_attrs` is hand-maintained, and explicitly note that we avoid `__annotations__` introspection because it would force every substrate handle to be runtime-checkable. The contract is enforced manually but the trade-off is explicit.

workers/tests/test_barrier.py
muhammad-ali-e: Fixed in 4a2c9b5ce. You're right — my rationale was factually wrong. `AsyncResult.id = None` is a class-level attribute (and `self.id = id` is also set in `__init__`), so `hasattr(AsyncResult, "id")` is already True and the property/instance-only story doesn't apply.

Rewrote the comment to give the actual reason: "assert against a real `AsyncResult` *instance* — that's what every production call site receives and logs `.id` from". We don't rely on whether the attr is class-level, a property, or set in `__init__`; we just check the smallest-blast-radius equivalent of what production sees.

workers/tests/test_barrier.py
muhammad-ali-e: Fixed in 4a2c9b5ce. Added:

```python
assert isinstance(async_result_instance.id, str), (
    f"AsyncResult.id should be a str (TaskHandle declares ``id: str``); "
    f"got {type(async_result_instance.id).__name__}"
)
```

Now a future substrate that satisfies `hasattr` with a UUID, `None`, or anything non-string would fail loudly — matching the declared `id: str` in the Protocol.

workers/api-deployment/tasks.py
muhammad-ali-e: Fixed in 4a2c9b5ce. You're right — non-empty `hash_values_of_files` doesn't guarantee non-empty `batch_tasks` (the `_get_file_batches` filter drops non-`FileHashData`/`dict` entries via `continue`). Tightened the comment:

```
#    is equivalent. Effectively unreachable — the upstream
#    ``if not hash_values_of_files:`` early return plus
#    valid ``FileHashData`` inputs mean ``_get_file_batches``
#    always yields >=1 batch — but the defence closes the
#    theoretical gap (e.g. all entries dropped because none
#    are ``FileHashData`` / ``dict``, which would still
#    bypass the upstream guard).
```

Matches the more careful phrasing in the test docstring at test_barrier.py.

workers/api-deployment/tasks.py
muhammad-ali-e: Deferred to the Phase 6b PR (RedisDecrBarrier introduction). Valid catch — `if result is None:` is the principled check per the Barrier Protocol contract.

Rationale for deferring rather than changing here:

- **Zero behaviour difference today.** Celery's `AsyncResult` has no custom `__bool__`, so truthiness and identity-against-`None` are equivalent in current production.
- **Phase 6b is the natural home.** That's where `RedisDecrBarrier` lands — the second substrate whose handle could plausibly have a non-trivial `__bool__` (e.g. "falsy while pending"). Tightening the check alongside the substrate that motivates it gives reviewers the full "new substrate + new defensive check" context in one PR.
- **Keeps this follow-up scoped.** Currently this PR is test/comment-only changes; no production behaviour changed. A one-line production tweak with no observable impact today is dead weight here.

Added a TODO at the call site so the rationale carries forward:

```python
if not result:
    # TODO(phase-6b): tighten to ``if result is None:``
    # when ``RedisDecrBarrier`` lands. The ``Barrier`` Protocol
    # declares ``None`` as the *sole* "no work enqueued" signal,
    # so identity-against-None is the principled check. [...]
```

(Initially included this fix in the round-2 commit; reverted after the team agreed Phase 6b is the right home.)

workers/api-deployment/tasks.py
muhammad-ali-e: Fixed in 4a2c9b5ce. You're right — the helper hard-coupled both knobs, so the higher-severity "non-empty batches + falsy chord" path had no executing coverage.

Added a `force_falsy_chord_with_batches` knob to `_run_workflow_api_with_mocks` (mutually exclusive with `force_empty_batches` — the helper raises an `AssertionError` if both are set, so future test authors can't accidentally combine them) and a new `TestApiDeploymentQueueFailureContract` class with two tests:

```python
def test_falsy_chord_with_non_empty_batches_raises(self, monkeypatch):
    with pytest.raises(Exception, match=r"(?i)queue|failed|chord"):
        _run_workflow_api_with_mocks(
            monkeypatch,
            hash_values_of_files={"f1": MagicMock(name="file_1")},
            force_falsy_chord_with_batches=True,
        )

def test_falsy_chord_with_non_empty_batches_does_not_dispatch_fallback(self, monkeypatch):
    # Asserts the fallback dispatch did NOT fire — a future refactor
    # routing both falsy-result cases through the same fallback would
    # silently lose the ERROR signal.
    ...
```

This locks both the raise contract AND the "no silent fallback" contract on the genuine-queue-failure branch.

workers/api-deployment/tasks.py
muhammad-ali-e: Deferred — valid observation, but the fix changes the response shape of `_run_workflow_api` (either `chord_id=None` + new `callback_task_id` field, or a new `chord_id_is_task_id` flag). That touches downstream consumers (log/metrics pipelines, API response readers) and is out of scope for a test/comment follow-up to Phase 6a.

Want to track it as a separate ticket — the right scoping is:

1. Audit downstream consumers of `chord_id` (backend log persistence, metrics, API response schema) to see who would actually mis-track if `chord_id` resolves to a standalone task id in the result backend.
2. Pick the additive form (`callback_task_id` + keep `chord_id`) vs the breaking form (`chord_id=None` + new field) based on what the audit shows.

For now the inline comment at the call site warns consumers to branch on `batches_created == 0` — agreed it's not enforcement, but the zero-batch path is also effectively unreachable in production (upstream guard + valid `FileHashData` inputs), so the silent-miss risk is bounded.

Happy to file the audit ticket if you'd like.

workers/tests/test_barrier.py
muhammad-ali-e: Fixed in 4a2c9b5ce. Added `manual_review_required` and `review_decisions_helper` knobs to `_run_workflow_api_with_mocks`, and a new `TestApiDeploymentManualReviewContract` class:

```python
def test_manual_review_required_invokes_decision_helper_per_batch(self, monkeypatch):
    decisions_helper = MagicMock(
        name="_calculate_manual_review_decisions_for_batch_api",
        return_value=[False],
    )
    _result, mock_create_chord, mock_dispatch = _run_workflow_api_with_mocks(
        monkeypatch,
        hash_values_of_files={"f1": MagicMock(name="file_1")},
        manual_review_required=True,
        review_decisions_helper=decisions_helper,
    )
    assert decisions_helper.call_count == 1   # one batch → one call
    assert mock_create_chord.called           # chord still fires
    assert not mock_dispatch.called           # not diverted into fallback
    assert mock_create_chord.call_args.kwargs.get("fairness") == FairnessKey(
        org_id="org_test", workload_type=WorkloadType.API
    )
```

Locks three things: (1) the decision helper runs once per batch, (2) the chord still fires after the manual-review mutation, (3) the fairness slot rides through unchanged. A future refactor that diverted the manual-review path into the fallback or dropped the per-batch helper invocation would fail loudly.

workers/tests/test_barrier.py
muhammad-ali-e: Fixed in 4a2c9b5ce. Replaced the bare `except Exception: pass` with `contextlib.suppress(Exception)`:

```python
import contextlib

# ``contextlib.suppress`` over a bare ``except Exception: pass``
# to declare intent explicitly: any post-``create_chord_execution``
# raise (e.g. the mocked api_client's status update path) is
# tolerated, because the next ``assert mock_create_chord.called``
# is what proves the chord call was actually reached.
    general_tasks._orchestrate_file_processing_general(...)
```

And added the matching `batch_tasks` non-empty pin (parity with the api-deployment sibling):

```python
batch_tasks = mock_create_chord.call_args.kwargs.get("batch_tasks")
assert batch_tasks, (
    "batch_tasks passed to create_chord_execution was empty — "
    "the chord path was not actually exercised"
)
```

`contextlib` is a module-level import now, so the suppress-block reads cleanly.

workers/tests/test_barrier.py
muhammad-ali-e: Fixed in 4a2c9b5ce. You're right — the matched substrings live in real code on the chord path too, not just in surrounding comments. Rewrote the docstring:

```
Originally a source-string-match test. The problem with the
string-match form: it would pass even if the dispatch branch
were deleted — the matched tokens (``args=[[]]``,
``process_batch_callback_api``, ``if not batch_tasks:``) also
appear on the chord path or in surrounding code. A test that
doesn't execute the branch can't prove the branch works.
```

More accurate framing of why the string-match form was inadequate.

workers/tests/test_chord_sites_characterisation.py
muhammad-ali-e: Fixed in 4a2c9b5ce. Right — aliased imports and module-attribute access would slip past both canaries:

```
**Known blind spot**: a return-value-discarded ``chord(batch)(cb)``
at column 0 (no assignment) would slip past this regex AND the
comment-line filter. The companion ``test_chord_import_only_in_barrier``
canary catches the common ``from celery import chord`` import
form — so a stray call site reachable via that import would
still trip it. However, aliased imports (``from celery import
chord as c``) or module-attribute access (``import celery`` +
``celery.chord(...)``) would evade *both* canaries. We accept
this gap because (a) those import shapes are non-idiomatic and
not used anywhere in workers/, and (b) ``barrier.py`` is the
sole sanctioned chord call site in the codebase, so any stray
call would surface in code review long before runtime. If a
future PR ships either of those import shapes outside
``barrier.py``, the canaries need broadening.
```

Documents both the gap and why we accept it, instead of over-claiming completeness.

workers/tests/test_barrier.py
muhammad-ali-e: Deferred — kept the explicit `hash_values_of_files=` arg on both tests as intent documentation. Agreed it's redundant relative to the helper's default, but the explicit form documents "this test is driving `_run_workflow_api` with a non-empty file dict" at the call site, which makes the chord-path / zero-batch / queue-failure intent legible without having to chase the helper signature.

Happy to drop if you'd prefer the terser form — purely a readability call.

jaseemjaskp (commented):
Automated follow-up review (PR Review Toolkit: Code Reviewer, Silent Failure Hunter, Type Design Analyzer, PR Test Analyzer, Comment Analyzer, Code Simplifier). The `FairnessKey` hoist is behaviour-preserving and the test rewrite from string-match to executing tests is a real improvement. Findings below are new (cross-checked against existing review comments to avoid duplicates). The one blocker is a vacuous test that can never fail.

workers/tests/test_barrier.py
```diff
@@ -358,103 +401,459 @@ def test_create_chord_execution_delegates_to_module_barrier(self):
...
+        lose the ERROR signal."""
+        with contextlib.suppress(Exception):
+            _result, _create_chord, mock_dispatch = (
+                _run_workflow_api_with_mocks(
+                    monkeypatch,
+                    hash_values_of_files={"f1": MagicMock(name="file_1")},
+                    force_falsy_chord_with_batches=True,
+                )
+            )
+            assert not mock_dispatch.called, (
```
jaseemjaskp: **[Critical · PR Test Analyzer + Silent Failure Hunter] This test is vacuous — it can never fail.** The `assert not mock_dispatch.called` sits inside `with contextlib.suppress(Exception):` (line 798). Two independent defects each neuter it:

1. `_run_workflow_api_with_mocks(force_falsy_chord_with_batches=True)` is the branch that *raises* (`tasks.py:783`). The exception fires on the tuple-unpacking assignment (lines 800-805) **before** `mock_dispatch` is ever bound, so the assert two lines down is never reached.
2. Even if it were reached, `AssertionError` is an `Exception`, so `suppress` would swallow the failing assert too.

The exact regression this claims to guard — fallback fires *and* no raise — produces a green test (verified by reproduction). The sibling `test_falsy_chord_with_non_empty_batches_raises` already covers the "must raise" half, so the branch isn't entirely uncovered, but the "fallback must not also fire" sub-contract is unverified.

**Fix:** bind the dispatch spy *before* the raising call and assert *outside* any suppress, e.g.
```python
with pytest.raises(Exception, match=r"Failed to queue execution task exec-1"):
    _run_workflow_api_with_mocks(monkeypatch, hash_values_of_files={"f1": MagicMock()}, force_falsy_chord_with_batches=True)
api_tasks = _load_api_deployment_tasks()
assert not api_tasks.dispatch.called  # dispatch is monkeypatched on the module; survives the raise
```
(The helper would need to expose the dispatch spy independently of its return value, since the raising path never returns the tuple.)

workers/tests/test_barrier.py
```diff
@@ -358,103 +401,459 @@ def test_create_chord_execution_delegates_to_module_barrier(self):
...
+            return_value=[False],
+        )
+        monkeypatch.setattr(
+            api_tasks, "_calculate_manual_review_decisions_for_batch_api", helper
+        )
+
+    mock_create_chord = MagicMock(name="create_chord_execution")
+    # When the chord path should fire normally, return a truthy
+    # handle. When forced empty OR forced falsy-with-batches, return
+    # ``None`` — the production code's ``if result is None:`` branch
```
jaseemjaskp: **[Important · Comment Analyzer] This comment describes phase-6b code that doesn't exist yet.** It says "the production code's `if result is None:` branch then distinguishes the two cases" — but the call site at `api-deployment/tasks.py:699` is `if not result:`, and the TODO(phase-6b) directly below it documents that tightening to `is None` is *deferred*. A reader cross-referencing this against `tasks.py` finds no `is None` check and can't tell which is authoritative — the exact comment-rot inversion the rest of this PR is fixing. Reword to match today's code, e.g. "…the production code's `if not result:` branch (truthiness today; the call-site TODO(phase-6b) tightens it to `is None`) then distinguishes the two cases via `batch_tasks` length."

workers/api-deployment/tasks.py
```diff
@@ -677,6 +677,12 @@ def _run_workflow_api(
         # The header tasks + callback now carry the fairness slot for
         # downstream PG Queue routing — closes the chord-fairness gap noted
         # in Phase 5.1.
+        # Hoist FairnessKey to a local so the chord path and the
+        # zero-batch fallback can't drift apart on the slot's shape.
+        api_fairness = FairnessKey(
+            org_id=str(schema_name),
+            workload_type=WorkloadType.API,
+        )
         result = WorkflowOrchestrationUtils.create_chord_execution(
```
jaseemjaskp, line 686: **[Important · Type Design Analyzer] The `BarrierHandle | None` invariant is erased here because `create_chord_execution` returns `Any`.** `Barrier.enqueue` is precisely typed `BarrierHandle | None`, but `WorkflowOrchestrationUtils.create_chord_execution` (`shared/workflow/execution/orchestration_utils.py:37`) widens it back to `-> Any`. So `result` is `Any` at this call site — a type checker can't see it's `BarrierHandle | None`, can't reason about the truthiness-vs-identity concern the TODO below documents, and can't catch a future `result.<missing_attr>`. Annotating the return as `BarrierHandle | None` (and `batch_tasks: list[Signature]` instead of `list[Any]`) is a pure-annotation change that turns the eventual phase-6b `if result is None:` tightening into something the type checker verifies rather than something a comment asserts. This is the root cause sitting *underneath* the already-noted `if not result:` discussion, not a duplicate of it.

workers/tests/test_barrier.py
```diff
@@ -358,103 +401,459 @@ def test_create_chord_execution_delegates_to_module_barrier(self):
...
+            "create_chord_execution",
+            mock_create_chord,
+        )
+
+        # ``contextlib.suppress`` over a bare ``except Exception: pass``
+        # to declare intent explicitly: any post-``create_chord_execution``
+        # raise (e.g. the mocked api_client's status update path) is
+        # tolerated, because the next ``assert mock_create_chord.called``
+        # is what proves the chord call was actually reached.
+        with contextlib.suppress(Exception):
```
jaseemjaskp, line 728: **[Minor · Silent Failure Hunter] `contextlib.suppress(Exception)` here is broader than its stated justification.** The comment says it tolerates "any post-`create_chord_execution` raise", but the suppress spans the *entire* `_orchestrate_file_processing_general(...)` call, including all pre-chord work (the unmocked `get_manual_review_service` / `create_workflow_file_data_with_manual_review` against a bare-MagicMock `api_client`, batch construction, integrity validation). Unlike the critical case above, the load-bearing asserts are correctly *outside* the suppress, so this isn't vacuous. But a future refactor that makes a pre-chord step raise would be silently hidden while `assert mock_create_chord.called` stays green. Prefer mocking the outer `update_workflow_execution_status` handler and running unsuppressed, or narrow the suppress to only the post-chord teardown.

workers/tests/test_barrier.py
```diff
@@ -358,103 +401,459 @@ def test_create_chord_execution_delegates_to_module_barrier(self):
...
-        assert "dispatch(" in text  # routed through queue_backend seam
-        assert '"process_batch_callback_api"' in text  # callback fired
-        assert "args=[[]]" in text  # body=[] matches chord-empty semantic
-        assert '"batches_created": 0' in text
+    def test_falsy_chord_with_non_empty_batches_raises(self, monkeypatch):
+        """When the barrier returns ``None`` despite non-empty
+        ``batch_tasks``, ``_run_workflow_api`` must raise — the outer
+        task handler then maps this to ``ExecutionStatus.ERROR`` for
+        the pipeline status update."""
+        with pytest.raises(Exception, match=r"(?i)queue|failed|chord"):
```
jaseemjaskp: **[Minor · Silent Failure Hunter] Loose `match` regex can pass on an unrelated failure.** `pytest.raises(Exception, match=r"(?i)queue|failed|chord")` would also match the helper's own `AssertionError("force_empty_batches and force_falsy_chord_with_batches are mutually exclusive…")` and any setup error whose message contains "failed" — masking a harness misconfiguration as a passing assertion (the same over-broad-catch class this suite is otherwise eliminating). Tighten to the actual production message: `match=r"Failed to queue execution task exec-1"`.

workers/tests/test_barrier.py
```diff
@@ -358,103 +401,459 @@ def test_create_chord_execution_delegates_to_module_barrier(self):
...
+            _run_workflow_api_with_mocks(
+                monkeypatch,
+                hash_values_of_files={"f1": MagicMock(name="file_1")},
+                manual_review_required=True,
+                review_decisions_helper=decisions_helper,
+            )
+        )
+
+        # One batch in the test helper → exactly one helper call.
+        assert decisions_helper.call_count == 1, (
```
jaseemjaskp: **[Minor · Code Reviewer + PR Test Analyzer] "once per batch" multiplicity is never actually exercised.** `_get_file_batches` is mocked to a single batch and `_create_file_data` returns the same `mock_file_data` each time, so `decisions_helper.call_count` can only ever be `1` — the per-batch contract the test name/docstring claims is not proven (a regression invoking the helper 0 or 2× per batch wouldn't be caught for the 2× case). To genuinely pin per-batch invocation, return >1 batch from the `_get_file_batches` mock and assert `call_count == len(batches)`.

workers/tests/test_barrier.py
```diff
@@ -358,103 +401,459 @@ def test_create_chord_execution_delegates_to_module_barrier(self):
...
+    spec.loader.exec_module(module)
+    return module
+
+
+def _run_workflow_api_with_mocks(
+    monkeypatch,
+    *,
+    hash_values_of_files: dict | None = None,
+    force_empty_batches: bool = False,
+    force_falsy_chord_with_batches: bool = False,
```
jaseemjaskp: **[Minor · Code Simplifier] Two mutually-exclusive booleans encode three states — collapse to one knob.** `force_empty_batches` + `force_falsy_chord_with_batches`, plus the runtime mutual-exclusivity guard (lines 464-468), is the classic "two bools for three states" smell (the third state being normal success). Replacing both with a single `chord_outcome: str = "success"` taking `"success" | "empty_batches" | "queue_failure"` makes the illegal `(True, True)` combination *unrepresentable* (delete the guard) and call sites self-documenting (`chord_outcome="empty_batches"`). Separately: the `helper = review_decisions_helper or MagicMock(...)` fallback at line 524 is a dead branch — the only caller (line 840) always supplies `review_decisions_helper`.

workers/tests/test_barrier.py
```diff
@@ -82,17 +83,59 @@ def test_celery_chord_barrier_satisfies_protocol(self):
...
+        # ``required_attrs`` is hand-maintained: if a future refactor
+        # adds a required attribute to ``TaskHandle`` /
+        # ``BarrierHandle``, add it here too. We do *not* introspect
+        # ``TaskHandle.__annotations__`` because runtime-checkable
+        # Protocols would force every substrate handle (including
+        # third-party ones we don't control) to be runtime-checkable
+        # too. Trade-off: the maintenance is manual but the contract
+        # is enforced.
+        async_result_instance = AsyncResult("placeholder-task-id")
+        required_attrs = ("id",)
```
jaseemjaskp, line 115: **[Minor · Type Design Analyzer] Derive `required_attrs` from the Protocol instead of hand-maintaining it.** The comment above (lines 105-112) acknowledges this list must be kept in sync with `TaskHandle` by hand, but nothing enforces that — so the documented drift can silently happen. `required_attrs = tuple(TaskHandle.__annotations__)` reads the contract straight from the type: adding a future `TaskHandle.foo: str` then automatically extends the assertion with zero manual edits, closing the drift the comment concedes. (`__annotations__` introspection works without `@runtime_checkable`, so it keeps the deliberate non-runtime-checkable trade-off the comment defends.)

workers/tests/test_barrier.py
muhammad-ali-e: Fixed in 1d13a7386. You're right — this was vacuous. Both defects you flagged were live:

1. `_run_workflow_api_with_mocks(force_falsy_chord_with_batches=True)` raised inside the helper, so the tuple-unpacking assignment that bound `mock_dispatch` never ran. Reproduced.
2. Even if it had, `AssertionError` is an `Exception`, so `contextlib.suppress` would have swallowed any failing assert.

Fix:

- Split the test helper into `_setup_workflow_api_mocks` (returns a `_WorkflowApiMocks` spy bundle bound **before** `_run_workflow_api` runs) and a thin `_run_workflow_api_with_mocks` convenience wrapper for tests that don't need post-raise assertions.
- The failing test now sets up mocks, drives `_run_workflow_api` itself inside `pytest.raises`, and asserts `not mocks.dispatch.called` **outside** the `pytest.raises` block — so an assertion failure surfaces instead of being swallowed.

```python
mocks = _setup_workflow_api_mocks(monkeypatch, chord_outcome="queue_failure")
with pytest.raises(Exception, match=r"Failed to queue execution task exec-1"):
    mocks.api_tasks._run_workflow_api(...)
assert not mocks.dispatch.called  # outside the raises block — load-bearing now
```

Sanity-checked: if production code dispatched on the queue-failure path before raising, `mocks.dispatch.called` would be True and the assert would fail loudly. The two sub-contracts (must-raise, no-fallback-dispatch) are now independently locked.

workers/tests/test_barrier.py
muhammad-ali-e: Fixed in 1d13a7386. Right — my comment claimed `if result is None:` was the production check, but `tasks.py:699` is still `if not result:` (the TODO directly below documents the deferred tightening). Reworded to match today's code:

```
# Truthy handle on the success path; None on the two failure
# paths. Production's `if not result:` branch (today; a future
# phase-6b PR may tighten to `if result is None:` — see the
# TODO at the call site in `api-deployment/tasks.py`) then
# distinguishes the two None cases via batch_tasks length.
```

No comment-rot now — a reader cross-referencing this against `tasks.py` sees the same `if not result:` plus the TODO carrying the rationale.

workers/api-deployment/tasks.py
muhammad-ali-e: Deferred — valid type-erasure observation, but scoping it out of this follow-up PR.

`WorkflowOrchestrationUtils.create_chord_execution` returning `Any` (and `batch_tasks: list[Any]`) is a pre-existing annotation gap, not something this PR introduced. Tightening it to `-> BarrierHandle | None` and `batch_tasks: list[Signature]` is a pure-annotation change but touches the shared `orchestration_utils` module and the mixin method `WorkflowOrchestrationMixin.create_chord`, which would broaden the PR's reach beyond test/comment fixes.

The right scoping is a separate annotation PR alongside the phase-6b `if result is None:` tightening — both changes are about making the `BarrierHandle | None` contract type-checker-verifiable rather than comment-asserted, and they reinforce each other when shipped together.

I'll add it to the phase-6b ticket scope.

workers/tests/test_barrier.py
muhammad-ali-e: Deferred — a previous review round already moved this from bare `except Exception: pass` to `contextlib.suppress(Exception)` at the team-agreed scope. Narrowing further would require either:

- Mocking `update_workflow_execution_status` on the outer status-update handler so the test could run unsuppressed — increases the mock surface area for a hypothetical pre-chord regression that hasn't materialised.
- Wrapping only the post-chord teardown in suppress — fragile because the "post-chord boundary" is internal to `_orchestrate_file_processing_general` and would have to be guessed from the source.

The load-bearing assertions (`mocks.create_chord.called`, fairness equality, `batch_tasks` non-empty) all sit outside the suppress, so this isn't vacuous in the way the Critical sibling was — a pre-chord regression would surface as `assert mocks.create_chord.called` failing with the "did not call" message, which is slightly misleading but not silent.

Acknowledged as a real refinement; tracking for the phase-6b cleanup pass alongside the type-annotation work.

workers/tests/test_barrier.py
muhammad-ali-e: Fixed in 1d13a7386. Tightened to the exact production message:

```python
with pytest.raises(Exception, match=r"Failed to queue execution task exec-1"):
    mocks.api_tasks._run_workflow_api(...)
```

The old `r"(?i)queue|failed|chord"` regex would have matched the helper's own mutual-exclusivity `AssertionError` (which contained "failed" in the message) and any setup error containing "failed" — same over-broad-catch class this suite is otherwise eliminating. The tight regex is anchored to the actual `raise Exception("Failed to queue execution task {execution_id}")` at `tasks.py:781-783` with the test's `execution_id="exec-1"`.

Also dropped the mutual-exclusivity `AssertionError` itself as part of the round-3 refactor — the two-bools were collapsed into a single `chord_outcome` enum, so the illegal combination is now unrepresentable rather than runtime-guarded.

workers/tests/test_barrier.py
muhammad-ali-e: Fixed in 1d13a7386. Right — with one batch, `call_count == 1` could only ever pass, so the "once per batch" claim wasn't actually pinned.

Refactored the helper to expose a `num_batches` knob, then drove the manual-review test with `num_batches=3`:

```python
num_batches = 3
mocks = _setup_workflow_api_mocks(
    monkeypatch,
    chord_outcome="success",
    manual_review_required=True,
    num_batches=num_batches,
)
mocks.api_tasks._run_workflow_api(...)

assert mocks.decisions_helper.call_count == num_batches, (
    f"Manual-review decision helper must run once per batch — "
    f"got {mocks.decisions_helper.call_count}, expected {num_batches}"
)
```

Now a regression invoking the helper 0×, 2×, or 6× per batch would all fail loudly with a message naming the expected vs actual count.

workers/tests/test_barrier.py
muhammad-ali-e: Fixed in 1d13a7386. Agreed — two bools encoding three states with a runtime mutual-exclusivity guard is a smell. Collapsed to a single `chord_outcome` knob:

```python
_VALID_CHORD_OUTCOMES = ("success", "empty_batches", "queue_failure")

def _setup_workflow_api_mocks(
    monkeypatch,
    *,
    chord_outcome: str = "success",
    manual_review_required: bool = False,
    num_batches: int = 1,
) -> _WorkflowApiMocks:
    if chord_outcome not in _VALID_CHORD_OUTCOMES:
        raise AssertionError(...)
    ...
```

The illegal `(force_empty_batches=True, force_falsy_chord_with_batches=True)` combination is now unrepresentable — the runtime guard is gone, replaced by the enum membership check, and call sites read self-documenting:

```python
_setup_workflow_api_mocks(monkeypatch, chord_outcome="queue_failure")
```

Also dropped the `helper = review_decisions_helper or MagicMock(...)` dead branch you flagged — the only caller previously supplied it unconditionally. The spy is now always created inside the helper and exposed via the `_WorkflowApiMocks` bundle's `decisions_helper` attribute.

workers/tests/test_barrier.py
muhammad-ali-e: Deferred — an earlier review round set the hand-maintained policy with the trade-off documented in the comment. You're technically correct that `__annotations__` works without `@runtime_checkable` (my explanation about runtime-checkable in the comment is overcautious), but the resolution is a refinement of an already-resolved thread rather than a defect.

Two reasons to keep the hand-maintained form for now:

1. `Protocol.__annotations__` includes only attributes declared with type hints in the Protocol body. `TaskHandle` is intentionally minimal (`id: str`) — there's no near-term plan to add fields, so the manual-update burden is one line every several phases.
2. Switching to introspection makes the contract implicit on a Python language feature; the explicit tuple is grepable and the hand-update line lives next to the comment that explains it.

I'll tighten the comment to drop the runtime-checkable claim (it's misleading) — "hand-maintained because the Protocol is intentionally minimal" is the honest framing. Will bundle that with the next pass.

jaseemjaskp (approved):
Re-review (PR Review Toolkit, round 2). Verified the production change in `_run_workflow_api` is correct — the hoisted `api_fairness` `FairnessKey` local is behaviour-preserving and used identically at both the chord call site and the zero-batch fallback. The new executing tests genuinely exercise their claimed branches (fairness contracts, zero-batch dispatch, genuine-queue-failure raise, and per-batch manual-review multiplicity with `num_batches=3`), and the previously-flagged vacuous/swallowed-assert defects are correctly fixed (spies bound before the raise, asserts outside `pytest.raises`). All prior review threads I opened have been addressed. Remaining observations are incremental test-completeness/comment-precision refinements below the bar for blocking. LGTM.

_Note: could not execute the test suite locally (the sandbox cannot compile `psycopg2-binary`/`celery`), so test verification was static._

PR #1300, opened 2025-05-12
Changes to support enterprise connections

johnyrahul:
## What

- Changes to support  provider for user login
-  Remove the env uirls which is not required, as it will take the URL based on the host its redirecting from.

## Why

-

## How

-

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

-

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

I have read and understood the [Contribution Guidelines](https://docs.unstract.com/unstract/contributing/unstract/).

PR #2027, opened 2026-06-10
UN-3525 [FIX] De-duplicate users by Auth0 user_id on login

johnyrahul:
## What

Prevents a single person from getting **two `User` rows** (and two `OrganizationMember` rows in the same org) when they have more than one Auth0 connection on the same email — e.g. a **SAML/enterprise** connection plus a **passwordless email (OTP)** connection.

## Why

`Auth0Service.handle_authorization_callback`'s social/passwordless path calls `AuthenticationHelper.get_or_create_user_by_email`, which resolved the local user **by email only**. `UserService.get_user_by_email` additionally filtered on `auth_provider=""`, so an existing **enterprise** row (which has a populated `auth_provider`) was invisible to it → a duplicate `User` row was created carrying the same Auth0 `user_id` but `auth_provider=""`.

Because org-membership eligibility is keyed on `user_id` (shared by both rows) while the "already a member?" check is keyed on the DB primary key (distinct), the duplicate then also acquired its own `OrganizationMember` row in the same org. Downstream features that key on `user.id` (e.g. HITL `reviewer_id`) break — a doc gets locked to one row while the user is logged in as the other.

Context: the `auth_provider=""` filter was added in #1300 (commit `f25874da4`); before that, `get_user_by_email` matched on email alone (one email = one user). #1300 introduced an `(email, auth_provider)` split but only inside that single query, and added no uniqueness — so the rest of the system (which still assumes one row per person) collides with it. Full analysis on UN-3525.

## How

Code-only, preventive (no data migration, no schema change):

- **`AuthenticationHelper.get_or_create_user_by_email`** now resolves by **`user_id` (the stable Auth0 subject) first**, reusing an existing row regardless of which login path created it, then falls back to email, then creates. This is the key guard: the passwordless login now finds the existing enterprise row instead of minting a duplicate.
- **`UserService.get_user_by_email`** now matches email **case-insensitively and across auth providers**, returning the oldest row via `.first()` (deterministic, and no longer raises `MultipleObjectsReturned` on already-duplicated emails).

This scope intentionally **does not** include de-duplicating existing rows or adding a `user_id`/email uniqueness constraint — those are higher-risk, interdependent (a constraint rejects until data is clean), and tracked separately on the ticket.

Low risk. `get_user_by_email` becomes provider-agnostic, which affects three callers — all benefit:
- `get_or_create_user_by_email` (social login) — now reuses any existing row; primary fix.
- Invite flow (`platform_admin/views.py`) — previously created a placeholder row for invitees who only had an enterprise row; now reuses it.
- `get_or_create_user` (`authentication_controller.py`) — more lenient account-by-email match.

The enterprise login path itself is unaffected (it uses `create_or_update_user`, not these lookups). No DB schema change.

## Relevant Docs

-

## Related Issues or PRs

- Jira: UN-3525
- Identity-model context introduced in #1300

- None

## Notes on Testing

- Added no-DB `SimpleTestCase` coverage (`account_v2/tests.py`) mocking `UserService` to assert the resolution order: user_id-first reuse, email-match backfill, email-match reuse (no overwrite), and create-on-no-match.
- **Not executed in my local sandbox**: this backend has no wired unit-test harness here (`pytest-django` absent, `account_v2/tests.py` was empty, conftest notes backend tests don't run under tox in CI, and Django setup needs `test.env`/service env vars not present locally). Verified with **ruff** (passes) only. Please run `python manage.py test account_v2` in a configured environment / CI.
- Suggested manual verification: with a user that has both a SAML and a passwordless Auth0 account on one email, log in via passwordless after the SAML row exists → confirm **no new `User` row** is created and the existing row is reused.

## Screenshots

...

## Checklist

### DIFF
--- a/backend/account_v2/authentication_helper.py
+++ b/backend/account_v2/authentication_helper.py
@@ -34,19 +34,29 @@ class AuthenticationHelper:
 
     @staticmethod
     def get_or_create_user_by_email(user_id: str, email: str) -> User:
-        """Get or create a user with the given email.
+        """Get or create the canonical user for an Auth0 identity.
 
-        If a user with the given email already exists, return that user.
-        Otherwise, create a new user with the given email and return it.
+        Resolution order:
+        1. By ``user_id`` (the stable Auth0 subject). This reuses an existing
+           row even when it was created by a different login path — e.g. an
+           enterprise/SSO row with a populated ``auth_provider`` — and is the
+           key guard against duplicate accounts for one person who signs in via
+           more than one Auth0 connection (e.g. SAML + passwordless email).
+        2. By email — for a pre-provisioned row (e.g. an invited user) that has
+           no ``user_id`` yet; backfill the ``user_id`` onto it.
+        3. Otherwise create a new user.
 
         Parameters:
-            user_id (str): The ID of the user.
+            user_id (str): The Auth0 subject (``sub``) of the user.
             email (str): The email of the user.
 
         Returns:
-            User: The user with the given email.
+            User: The canonical user for this identity.
         """
         user_service = UserService()
+        user = user_service.get_user_by_user_id(user_id)
+        if user:
+            return user
         user = user_service.get_user_by_email(email)
         if user and not user.user_id:
             user = user_service.update_user(user, user_id)
--- a/backend/account_v2/user.py
+++ b/backend/account_v2/user.py
@@ -41,11 +41,14 @@ class UserService:
         return user
 
     def get_user_by_email(self, email: str) -> User | None:
-        try:
-            user: User = User.objects.get(email=email, auth_provider="")
-            return user
-        except User.DoesNotExist:
-            return None
+        # Resolve a single canonical row for an email, matched case-insensitively
+        # and across auth providers. The previous ``auth_provider=""`` filter
+        # only matched standard-login rows, so a person who also had an
+        # enterprise/SSO row (populated ``auth_provider``) was invisible here and
+        # a duplicate account got created. ``.first()`` (oldest row, ordered
+        # deterministically) also avoids ``MultipleObjectsReturned`` for emails
+        # that already have duplicate rows.
+        return User.objects.filter(email__iexact=email).order_by("created_at").first()
 
     def get_user_by_user_id(self, user_id: str) -> Any:
         try:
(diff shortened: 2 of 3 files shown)


backend/account_v2/user.py
```diff
@@ -41,11 +41,14 @@ def update_user(self, user: User, user_id: str) -> User:
...
-        except User.DoesNotExist:
-            return None
+        # Resolve a single canonical row for an email, matched case-insensitively
+        # and across auth providers. The previous ``auth_provider=""`` filter
+        # only matched standard-login rows, so a person who also had an
+        # enterprise/SSO row (populated ``auth_provider``) was invisible here and
+        # a duplicate account got created. ``.first()`` (oldest row, ordered
+        # deterministically) also avoids ``MultipleObjectsReturned`` for emails
+        # that already have duplicate rows.
+        return User.objects.filter(email__iexact=email).order_by("created_at").first()
```
muhammad-ali-e, line 51: will this make any performance issue , especcially we moving away from `.get` to `first` ?

muhammad-ali-e, line 51: @johnyrahul Also, we are resolving the issue here when we have two user objects. Are we resolving the actual root cause inside `get_or_create_user_by_email`?

backend/account_v2/authentication_helper.py
```diff
@@ -34,19 +34,29 @@ def create_or_update_user(email: str, user_id: str, provider: str) -> User:
...
-            user_id (str): The ID of the user.
+            user_id (str): The Auth0 subject (``sub``) of the user.
             email (str): The email of the user.
 
         Returns:
-            User: The user with the given email.
+            User: The canonical user for this identity.
         """
         user_service = UserService()
+        user = user_service.get_user_by_user_id(user_id)
```
muhammad-ali-e, line 57: How it behave , if user cames with email/or social login instaed of enterprice ? will he/she logined with other user (with social login) ?

johnyrahul:
Closing in favour of a corrected approach. The email-merge direction here would still land an enterprise user's passwordless login in the enterprise org (wrong). New route: isolate identity per Auth0 connection — keep the authenticated `sub` (drop the get_user_by_email min-by-created_at collapse) so a passwordless login is its own identity and gets a fresh personal org, never inheriting the enterprise org. Will be replaced by a cloud-repo PR (Auth0 helper) + a revised OSS PR. See UN-3525.

PR #2028, opened 2026-06-10
UN-3166 [FIX] Use group-shared adapters in Prompt Studio & workflows

kirtimanmishrazipstack:
## What

- `prompt_studio_helper.py`: `validate_profile_manager_owner_access` now ORs `has_group_access` into all four adapter checks (LLM, vector store, embedding, x2text).
- `tool_instance_helper.py`: `validate_adapter_access` ORs `has_group_access` into its adapter check.

## Why

- Adapters shared with a user **via a group** were rejected in Prompt Studio and workflow tools — the hand-rolled checks only honored `created_by | shared_users | shared_to_org`, missing the group axis from the UN-2977 share model.

## How

- Reuse the canonical `permissions.permission.has_group_access(user, obj)` helper (reads `ResourceGroupShare`), bringing these hand-rolled checks to parity with `AdapterInstance.for_user`.

- No. The change is an additive `or` clause that only widens access to adapters already shared to the user's group; owner, direct-share, and org-share paths are unchanged.

## Database Migrations

- None

## Env Config

- None

## Relevant Docs

-

## Related Issues or PRs

- Cloud counterpart: Zipstack/unstract-cloud#1538

## Dependencies Versions

- None

## Notes on Testing

- Manual: share an adapter to a group; a member of that group can select and use it in a Prompt Studio profile and in workflow tools.

## Checklist

### DIFF
--- a/backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
+++ b/backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
@@ -11,6 +11,7 @@ from account_v2.models import User
 from adapter_processor_v2.models import AdapterInstance, UserDefaultAdapter
 from django.conf import settings
 from django.db import transaction
+from permissions.permission import has_group_access
 from plugins import get_plugin
 from rest_framework.exceptions import APIException
 from rest_framework.request import Request
@@ -209,6 +210,7 @@ class PromptStudioHelper:
             or profile_manager.llm.shared_users.filter(
                 pk=profile_manager_owner.pk
             ).exists()
+            or has_group_access(profile_manager_owner, profile_manager.llm)
         )
         is_vector_store_owned = (
             profile_manager.vector_store.shared_to_org
@@ -216,6 +218,7 @@ class PromptStudioHelper:
             or profile_manager.vector_store.shared_users.filter(
                 pk=profile_manager_owner.pk
             ).exists()
+            or has_group_access(profile_manager_owner, profile_manager.vector_store)
         )
         is_embedding_model_owned = (
             profile_manager.embedding_model.shared_to_org
@@ -223,6 +226,7 @@ class PromptStudioHelper:
             or profile_manager.embedding_model.shared_users.filter(
                 pk=profile_manager_owner.pk
             ).exists()
+            or has_group_access(profile_manager_owner, profile_manager.embedding_model)
         )
         is_x2text_owned = (
             profile_manager.x2text.shared_to_org
@@ -230,6 +234,7 @@ class PromptStudioHelper:
             or profile_manager.x2text.shared_users.filter(
                 pk=profile_manager_owner.pk
             ).exists()
+            or has_group_access(profile_manager_owner, profile_manager.x2text)
         )
 
         if not (
--- a/backend/tool_instance_v2/tool_instance_helper.py
+++ b/backend/tool_instance_v2/tool_instance_helper.py
@@ -9,6 +9,7 @@ from adapter_processor_v2.models import AdapterInstance
 from django.core.exceptions import PermissionDenied
 from django.core.exceptions import ValidationError as DjangoValidationError
 from jsonschema.exceptions import ValidationError as JSONValidationError
+from permissions.permission import has_group_access
 from prompt_studio.prompt_studio_registry_v2.models import PromptStudioRegistry
 from tenant_account_v2.organization_member_service import OrganizationMemberService
 from workflow_manager.workflow_v2.constants import WorkflowKey
@@ -508,6 +509,7 @@ class ToolInstanceHelper:
                 or adapter_instance.shared_to_org
                 or adapter_instance.created_by == user
                 or adapter_instance.shared_users.filter(pk=user.pk).exists()
+                or has_group_access(user, adapter_instance)
             ):
                 logger.error(
                     "User %s doesn't have access to adapter %s",


backend/tool_instance_v2/tool_instance_helper.py
```diff
@@ -508,6 +509,7 @@ def validate_adapter_access(
                 or adapter_instance.shared_to_org
                 or adapter_instance.created_by == user
                 or adapter_instance.shared_users.filter(pk=user.pk).exists()
+                or has_group_access(user, adapter_instance)
```
jaseemjaskp, line 512: 🟠 **Test coverage gap (Medium).** `validate_adapter_access` has **no test directory at all** (`tool_instance_v2/` only has `migrations/`), so this new access-widening clause is entirely uncovered. Since it broadens who is granted adapter access, the highest-value tests are:

1. **Group member granted** — owner not in `shared_users`/`created_by`, `shared_to_org=False`, `is_admin` patched `False`, adapter group-shared with a group the user belongs to → no `PermissionDenied`.
2. **Non-group user denied** — same setup but user not in the shared group → `PermissionDenied` raised (the false-grant guard — highest priority).
3. **Adapter not group-shared → no false grant** — no `ResourceGroupShare` row; user in an unrelated group → `PermissionDenied`.
4. **`is_usable=False` still rejected** — confirms the new clause didn't bypass the earlier `not adapter_instance.is_usable` guard at line 499.

Suggested location: `backend/tool_instance_v2/tests/test_validate_adapter_access.py`. Reuse `GroupSharingTestBase` and `set_resource_share_groups(...)` from `tenant_account_v2/tests.py` (UN-2977) and patch `OrganizationMemberService.is_user_organization_admin` → `False` so the group branch is actually reached.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -209,27 +210,31 @@ def validate_profile_manager_owner_access(
             or profile_manager.llm.shared_users.filter(
                 pk=profile_manager_owner.pk
             ).exists()
+            or has_group_access(profile_manager_owner, profile_manager.llm)
```
jaseemjaskp, line 213: 🟠 **Test coverage gap (Medium).** `validate_profile_manager_owner_access` is currently *stubbed out* in the only test that names it (`tests/test_build_index_payload.py:379` patches it to `return_value=None`), so its body has zero execution coverage. This method passes only if **all four** adapter slots are accessible (AND-of-ORs), which is the subtle part to lock down:

1. **All four reachable via group access → passes** — guards against a copy-paste error binding a `has_group_access` line to the wrong adapter field (lines 213/221/229/237).
2. **Three of four group-shared, fourth not → still raises** — proves the new OR-clauses didn't accidentally turn the four-way AND into an OR (a real access-control regression).

Suggested location: `backend/prompt_studio/prompt_studio_core_v2/tests/test_validate_profile_manager_owner_access.py`. Patch the org-admin early-return (line 204) to `False` so the group branch is exercised.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -209,27 +210,31 @@ def validate_profile_manager_owner_access(
...
             ).exists()
+            or has_group_access(profile_manager_owner, profile_manager.embedding_model)
         )
         is_x2text_owned = (
             profile_manager.x2text.shared_to_org
             or profile_manager.x2text.created_by == profile_manager_owner
             or profile_manager.x2text.shared_users.filter(
                 pk=profile_manager_owner.pk
             ).exists()
+            or has_group_access(profile_manager_owner, profile_manager.x2text)
```
jaseemjaskp, line 237: 🟢 **Optional refactor (Low — recommend NOT in this PR).** With this clause added, the four `is_*_owned` blocks (lines 207–238) are now identical except for the adapter attribute, so a helper like `_is_adapter_accessible(owner, adapter)` is tempting. But the named booleans must still exist individually because the error block below (lines 246–273) consumes each one separately, so you'd only trade four 6-line blocks for four 1-line calls. Extracting it here would also mix a structural refactor into a focused bug-fix and bloat the diff. Better as a dedicated follow-up that converts both the checks and the error reporting to iterate over the four `(adapter, name)` pairs together.

Minor pre-existing nit while you're here: this method's docstring says `Raises: PermissionError`, but it actually raises DRF `PermissionDenied` (line ~245).

jaseemjaskp (commented):

Reviewed the 7-line diff across Code Reviewer, Silent Failure Hunter, Type Design, Test Analyzer, Comment Analyzer, and Code Simplifier.

**Verdict: the code change is correct and safe to merge.** Both call sites pass correctly-typed `User` + `AdapterInstance` arguments, the `None`-owner case is already guarded, an empty `group_memberships` yields a clean `False` (no exception), and the clause is fail-closed (an error propagates rather than silently granting). It mirrors the already-shipping pattern in `permissions/permission.py:125,140`.

**The one substantive gap is test coverage** (no tests in this PR; neither changed method has behavioral coverage today). Inline notes below. Severity: 🟠 Medium (security-sensitive access-widening change), everything else 🟢 Low / out-of-scope.

backend/tool_instance_v2/tool_instance_helper.py
kirtimanmishrazipstack: @jaseemjaskp Valid — `validate_adapter_access` has no test directory. Tracked for a follow-up test PR (group-member granted; non-group denied; not-shared → no false grant; `is_usable=False` still rejected), reusing `GroupSharingTestBase` / `set_resource_share_groups` and patching `is_user_organization_admin` to False.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
kirtimanmishrazipstack: @jaseemjaskp Agreed — the AND-of-ORs is the part to lock down. Tracked for the same follow-up: all four reachable via group → passes, and three-of-four → still raises, with the org-admin early-return patched to False so the group branch runs.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
kirtimanmishrazipstack: @jaseemjaskp Agreed on deferring the refactor (per your note). On the docstring nit: this method raises the module custom `PermissionError` (imported at line 55, raised at ~284), not DRF `PermissionDenied` — so `Raises: PermissionError` is accurate as written. Happy to rename that custom exception in a follow-up if shadowing the builtin is the concern.

jaseemjaskp (approved):
Re-reviewed the latest revision. The change adds `has_group_access(...)` as one more OR-clause to the existing access predicates in `validate_profile_manager_owner_access` and `validate_adapter_access`. This matches the established pattern already used in `IsOwnerOrSharedUser.has_object_permission` (permissions/permission.py:125), both call sites pass a real `User` (with the `group_memberships` reverse relation), the AND-of-ORs semantics are preserved, and the extra query cost is short-circuited and ContentType-cached. No new correctness/security/contract concerns. The previously-raised test-coverage gaps and the deferred refactor/docstring nit remain tracked in their existing threads. LGTM.

Deepak-Kesavan (approved):

PR #2029, opened 2026-06-11
UN-3308 [MISC] Add plugin-gated banner mount in PageLayout

athul-rs:
## What

- Adds a plugin-gated mount point in `PageLayout` for an optional status banner contributed by the `marketplace` plugin, when present.
- Extracts `isModuleMissing` into a zero-dependency `helpers/pluginLoader.js` (it previously lived in `useMainAppRoutes.js`, which would have created an import cycle with `PageLayout`); `Router.jsx` and `useMainAppRoutes.js` now import it from there.

## Why

- Plugin routes were already mountable via `Router.jsx` / `useMainAppRoutes.js` (#2022 pattern), but there was no hook point for plugin-contributed UI inside the persistent page chrome. The banner component ships with the plugin; the layout only needs a gated slot.

## How

- Same plugin-gated dynamic import used elsewhere (`Router.jsx`, `TopNavBar`): `await import("../../plugins/marketplace")` inside try/catch with `isModuleMissing` — missing plugin is the expected quiet case, any other failure is logged.
- Rendered above `<Outlet />`; the component self-manages visibility and renders nothing when it has nothing to show.

- No. In builds without the plugin the import fails, the binding stays `undefined`, and nothing renders — verified with `vite build` both without the plugin and with it present. The `isModuleMissing` move is a pure relocation (same implementation, all consumers updated).

## Relevant Docs

- N/A (follows the existing plugin-gating convention in `Router.jsx`)

## Related Issues or PRs

- #2022 (plugin route registration — same gating pattern)

- None

## Notes on Testing

- `vite build` green with and without the plugin directory present; biome clean.

## Screenshots

- N/A (no rendering change in OSS builds)

## Checklist


frontend/src/layouts/page-layout/PageLayout.jsx
```diff
@@ -12,6 +12,19 @@ import {
...
+// the provisioning webhook, and clears itself once the subscription
+// activates. Absent in OSS builds — the import fails and the banner
+// stays unmounted.
+let MarketplacePendingBanner;
+try {
+  const marketplaceMod = await import("../../plugins/marketplace");
+  MarketplacePendingBanner = marketplaceMod.MarketplacePendingBanner;
+} catch {
+  // Ignore if plugin not available
+}
```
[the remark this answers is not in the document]
athul-rs: Fixed in b296a91d — the catch now mirrors the Router.jsx contract: `isModuleMissing(err)` stays quiet for the expected plugin-absent OSS case and logs anything else, so a broken plugin can't silently unmount the banner in cloud builds. Both builds (with/without plugin) verified green.

frontend/src/layouts/page-layout/PageLayout.jsx
```diff
@@ -12,6 +12,19 @@ import {
...
+// marketplace purchase is being confirmed" between the buyer's claim and
+// the provisioning webhook, and clears itself once the subscription
+// activates. Absent in OSS builds — the import fails and the banner
+// stays unmounted.
+let MarketplacePendingBanner;
+try {
+  const marketplaceMod = await import("../../plugins/marketplace");
+  MarketplacePendingBanner = marketplaceMod.MarketplacePendingBanner;
+} catch {
+  // Ignore if plugin not available
```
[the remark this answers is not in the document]
athul-rs: Fixed in b296a91d — the catch now mirrors the Router.jsx contract: `isModuleMissing(err)` stays quiet for the expected plugin-absent OSS case and logs anything else, so a broken plugin can't silently unmount the banner in cloud builds. Both builds (with/without plugin) verified green.

frontend/src/layouts/page-layout/PageLayout.jsx
```diff
@@ -11,6 +11,29 @@ import {
   getLocalStorageValue,
   setLocalStorageValue,
 } from "../../helpers/localStorage";
+import { isModuleMissing } from "../../routes/useMainAppRoutes.js";
```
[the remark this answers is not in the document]
athul-rs: Fixed in 0a16ed4d — `isModuleMissing` extracted to a new zero-dependency `helpers/pluginLoader.js`; Router.jsx, useMainAppRoutes.js, and PageLayout.jsx all import from there, so the cycle is gone entirely. (For the record: the TypeError scenario wouldn't have fired — `isModuleMissing` is a hoisted top-level `function` declaration, initialized at module instantiation, so it's callable from a cycled module even before the defining module's body runs — but a cycle plus TLA is fragile enough that extraction is clearly the better shape.) Both builds (with/without plugin) verified green.

frontend/src/helpers/pluginLoader.js
```diff
@@ -0,0 +1,24 @@
...
+//
+// Zero-dependency on purpose: consumers include modules that import
+// each other (Router/useMainAppRoutes/PageLayout), and this helper
+// must never re-introduce an import cycle.
+export function isModuleMissing(err) {
+  const msg = err?.message || "";
+  return (
+    err?.code === "MODULE_NOT_FOUND" ||
+    msg.includes("Failed to fetch dynamically imported module") ||
+    msg.includes("Cannot find module")
```
jaseemjaskp, line 31: **`isModuleMissing` does not match the actual production "missing plugin" signal.** In the shipped Vite build (and dev server), an absent plugin is *not* a fetch/`MODULE_NOT_FOUND` failure — `vite.config.js`'s `optionalPluginImports` plugin resolves the missing path to a stub module whose body is `throw new Error('Optional plugin not available')` (`vite.config.js:72`). That message matches none of the three branches here, so `isModuleMissing` returns `false` for the *expected* OSS case. Net effect: every OSS page load lands in the consumer's `else` branch and logs `console.error('[marketplace] MarketplacePendingBanner import failed unexpectedly', …)` — exactly the noise this helper was meant to suppress (the banner still correctly does not mount, so it's cosmetic, not a functional break). The `MODULE_NOT_FOUND` / `Cannot find module` branches only ever fire under jsdom/vitest, never in the bundle.

Suggested fix — recognize the real stub message:
```js
msg.includes("Failed to fetch dynamically imported module") ||
msg.includes("Cannot find module") ||
msg.includes("Optional plugin not available")
```
This is pre-existing for the Router/useMainAppRoutes call sites, but since this PR centralizes the classifier, here is the right place to close the gap.

**Also (test gap):** there is no `pluginLoader.test.js`, although the sibling helper `requestId.js` has `requestId.test.js`. This function is now the single chokepoint deciding plugin route/banner registration. A small vitest pinning each branch — including the `'Optional plugin not available'` stub string and `null`/`undefined`/`{}` inputs — would have caught this gap and guards against future bundler/message-wording drift.

frontend/src/helpers/pluginLoader.js
```diff
@@ -0,0 +1,24 @@
+// Classify a dynamic plugin-import failure as "plugin not shipped"
+// (expected in OSS builds) vs. a real load failure.
+//
+// LIMITATION: in a browser/Vite bundle this isn't fully sound — a
+// genuinely-absent plugin and a present-but-failed-to-load chunk
+// (transient CDN/origin 5xx, a stale hashed asset) both surface as
+// "Failed to fetch dynamically imported module", and `MODULE_NOT_FOUND`
+// is a Node/CJS code that only appears under jsdom/vitest, not the
+// shipped bundle. So a transient chunk-load failure can be misread as
+// "missing" and de-register the route for that session — same outcome
```
jaseemjaskp: **This comment is now stale for the banner consumer.** It says a misclassified transient failure can be "misread as 'missing' and **de-register the route** for that session" — but as of this PR the helper also gates a *banner mount* in `PageLayout.jsx`, not only routes. Since the header itself promises it's "centralized so any future hardening lands once," it should remain the single source of truth. Suggest generalizing the consequence, e.g. "…silently disabling whatever the consumer gates on it — a route in the routers, the status banner in PageLayout — for that session."

jaseemjaskp (commented):
Automated PR Review Toolkit pass (code-review / silent-failure / type-design / test-coverage / comment / simplify), scoped to the 4 changed files. The extraction is clean and consistent with the existing top-level-`await import` plugin pattern, and the previously-raised "catch swallows unexpected failures" concern is resolved. Two new, higher-value findings are posted inline on `pluginLoader.js`; lower-priority/design notes are summarized in the PR thread rather than posted inline to avoid clutter.

frontend/src/helpers/pluginLoader.js
athul-rs: Fixed in 506acd3b.

frontend/src/helpers/pluginLoader.js
athul-rs: Fixed in 506acd3b.

frontend/src/helpers/pluginLoader.js
athul-rs: Confirmed against vite.config.js:72 — the stub body is `throw new Error('Optional plugin not available')`, so none of the three branches matched the real OSS signal. Added `msg.includes("Optional plugin not available")` as the primary match (documented as such in the header). OSS `vite build` still green. Good catch — this also quietly fixes the same misclassification for the Router/useMainAppRoutes call sites.

frontend/src/helpers/pluginLoader.js
athul-rs: Generalized the comment: "…silently disable whatever the consumer gates on it — a route in the routers, the status banner in PageLayout — for that session."

jaseemjaskp (approved):
Re-reviewed via the PR Review Toolkit (code-review, silent-failure, type-design, comment-accuracy, simplifier, test-coverage agents), scoped to the 4 changed files. The refactor is clean: `isModuleMissing` is correctly centralized into a zero-dependency helper (breaking the TLA import cycle), the new `"Optional plugin not available"` matcher correctly recognizes the production Vite stub signal (`vite.config.js:72`), and `PageLayout` now mirrors the established catch-and-log contract used by the routers. All previously-raised threads are addressed and the comments are accurate. No clearly-important (correctness/security/contract/regression) findings remain — only a couple of low-probability defensive-hardening notes (a present-but-renamed plugin export would render nothing without a log; no unit test for the new classifier) that are below the blocking bar. LGTM.

PR #2030, opened 2026-06-11
UN-2742 [FIX] Show profile name instead of LLM name in Output Analyzer tabs

athul-rs:
## What

- Output tabs in Prompt Studio now show the **LLM profile name** as the label, with the LLM model name available as a tooltip on hover. Previously the tab label was the model name, falling back to profile name.

## Why

[UN-2742](https://zipstack.atlassian.net/browse/UN-2742) — users create profiles with specific intent (the same model can be used with different LLMWhisperer modes, chunking, retrieval settings). When comparing outputs across profiles, the model name is ambiguous — two tabs can read identically while being different profiles. The profile name is the meaningful identity; the model is secondary context, now preserved in a tooltip.

## How

- `combined-output/JsonView.jsx` — flipped label preference to `profile_name || llm_model`, wrapped in `<Tooltip title={llm_model}>`. This is the shared tab strip used by both the Output Analyzer and the main combined-output view, so both stay consistent.
- `output-for-doc-modal/OutputForDocModal.jsx` — same pattern applied to the identical tab strip there.

- No. Tab selection is keyed on `profile_id` (JsonView) / index (OutputForDocModal), never on the label text, so behavior is unchanged — only the visible label flips. No tests, e2e selectors, or CSS reference the label text. `profile_name` is a required, whitespace-validated field at profile creation, so labels can't be blank. If a profile's adapter was deleted (no resolvable model), the tooltip simply doesn't render (antd disables tooltips with undefined title) — previously that case showed the profile name with no tooltip, so nothing is lost.

## Database Migrations

- None

## Env Config

- None

## Relevant Docs

- N/A

## Related Issues or PRs

- Jira: UN-2742

## Dependent Features

- None

## Notes on Testing

- Biome lint clean; `vite build` succeeds.
- Manual: open a Prompt Studio project with 2+ profiles → Output Analyzer and combined output tabs show profile names; hovering a tab shows the model name; tab switching unaffected.

## Screenshots

N/A

## Checklist

[UN-2742]: https://zipstack.atlassian.net/browse/UN-2742?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/src/components/custom-tools/combined-output/JsonView.jsx
+++ b/frontend/src/components/custom-tools/combined-output/JsonView.jsx
@@ -1,4 +1,4 @@
-import { Tabs } from "antd";
+import { Tabs, Tooltip } from "antd";
 import TabPane from "antd/es/tabs/TabPane";
 import Prism from "prismjs";
 import PropTypes from "prop-types";
@@ -70,7 +70,11 @@ function JsonView({
           )}
           {adapterData.map((adapter) => (
             <TabPane
-              tab={<span>{adapter?.llm_model || adapter?.profile_name}</span>}
+              tab={
+                <Tooltip title={adapter?.llm_model}>
+                  <span>{adapter?.profile_name || adapter?.llm_model}</span>
+                </Tooltip>
+              }
               key={adapter?.profile_id}
             />
           ))}
--- a/frontend/src/components/custom-tools/combined-output/JsonView.jsx
+++ b/frontend/src/components/custom-tools/combined-output/JsonView.jsx
@@ -71,8 +71,12 @@ function JsonView({
           {adapterData.map((adapter) => (
             <TabPane
               tab={
-                <Tooltip title={adapter?.llm_model}>
-                  <span>{adapter?.profile_name || adapter?.llm_model}</span>
+                <Tooltip title={adapter?.llm_model || adapter?.profile_name}>
+                  <span>
+                    {adapter?.profile_name ||
+                      adapter?.llm_model ||
+                      `Profile ${adapter?.profile_id ?? "?"}`}
+                  </span>
                 </Tooltip>
               }
               key={adapter?.profile_id}
--- a/frontend/src/components/custom-tools/output-for-doc-modal/OutputForDocModal.jsx
+++ b/frontend/src/components/custom-tools/output-for-doc-modal/OutputForDocModal.jsx
@@ -341,8 +341,12 @@ function OutputForDocModal({
             {adapterData?.map((adapter, index) => (
               <TabPane
                 tab={
-                  <Tooltip title={adapter?.llm_model}>
-                    <span>{adapter?.profile_name || adapter?.llm_model}</span>
+                  <Tooltip title={adapter?.llm_model || adapter?.profile_name}>
+                    <span>
+                      {adapter?.profile_name ||
+                        adapter?.llm_model ||
+                        `Profile ${index + 1}`}
+                    </span>
                   </Tooltip>
                 }
                 key={(index + 1)?.toString()}
--- a/frontend/src/components/custom-tools/output-for-doc-modal/OutputForDocModal.jsx
+++ b/frontend/src/components/custom-tools/output-for-doc-modal/OutputForDocModal.jsx
@@ -3,7 +3,7 @@ import {
   CloseCircleFilled,
   InfoCircleFilled,
 } from "@ant-design/icons";
-import { Button, Modal, Table, Tabs, Typography } from "antd";
+import { Button, Modal, Table, Tabs, Tooltip, Typography } from "antd";
 import TabPane from "antd/es/tabs/TabPane";
 import PropTypes from "prop-types";
 import { useEffect, useState } from "react";
@@ -340,7 +340,11 @@ function OutputForDocModal({
             <TabPane tab={<span>Default</span>} key={"0"}></TabPane>
             {adapterData?.map((adapter, index) => (
               <TabPane
-                tab={<span>{adapter?.llm_model || adapter?.profile_name}</span>}
+                tab={
+                  <Tooltip title={adapter?.llm_model}>
+                    <span>{adapter?.profile_name || adapter?.llm_model}</span>
+                  </Tooltip>
+                }
                 key={(index + 1)?.toString()}
               ></TabPane>
             ))}


chandrasekharan-zipstack (approved):

jaseemjaskp (commented):

**Verdict:** Clean, focused change — correctly implements UN-2742 (profile name as the tab label, LLM model in a hover tooltip). Change is consistent across both files. No critical/high issues. Findings below are LOW/nit.

Inline comments cover the two behavioral edge cases (blank tab + redundant/empty tooltip when fields are missing).

Additional non-blocking notes (lines outside the diff, so not inlined):
- **Weak prop contract (LOW)** — `JsonView.jsx:111` declares `adapterData: PropTypes.array`, which permits any array. The producer `getLLMModelNamesForProfiles` (`GetStaticData.js:476-491`) always emits `{ profile_name, llm_model, profile_id }`, so tightening to `PropTypes.arrayOf(PropTypes.shape({ profile_id: PropTypes.string.isRequired, profile_name: PropTypes.string, llm_model: PropTypes.string }))` would document the contract and surface dev-mode warnings. `profile_id` is load-bearing (tab key + selection).
- **Duplication (nit, no action)** — the tab-label JSX is now identical in both files. A shared `ProfileTabLabel` is *not* worth it for two call sites with no existing shared-component precedent in these sibling directories; extract only if a third call site appears.
- **Tests (optional)** — frontend has Vitest infra but minimal coverage; this presentational change doesn't warrant new tests under current conventions. If anything, a single guard around the empty-string `profile_name` fallback.

frontend/src/components/custom-tools/combined-output/JsonView.jsx
```diff
@@ -70,7 +70,11 @@ function JsonView({
           )}
           {adapterData.map((adapter) => (
             <TabPane
-              tab={<span>{adapter?.llm_model || adapter?.profile_name}</span>}
+              tab={
+                <Tooltip title={adapter?.llm_model}>
```
jaseemjaskp: **LOW / nit — tooltip is redundant or empty in the fallback path.**

When `profile_name` is missing, the span already falls back to `llm_model`, and the tooltip `title` is *also* `llm_model` — so the hover just repeats the visible text. And when `llm_model` itself is undefined (the adapter for `profile.llm` doesn't resolve in `getLLMModelNamesForProfiles`), the tooltip is empty and hover reveals nothing. If both fields are falsy the tab renders with no label at all (a blank, anonymous-but-selectable tab) — pre-existing, but the new tooltip was a chance to address it.

Suggestion (only show the tooltip when it adds info, and guarantee a visible label):
```jsx
tab={
  <Tooltip title={adapter?.profile_name ? adapter?.llm_model : undefined}>
    <span>{adapter?.profile_name || adapter?.llm_model || "Unnamed profile"}</span>
  </Tooltip>
}
```
(antd suppresses the popup when `title` is `undefined`.) Non-blocking.

frontend/src/components/custom-tools/output-for-doc-modal/OutputForDocModal.jsx
```diff
@@ -340,7 +340,11 @@ function OutputForDocModal({
             <TabPane tab={<span>Default</span>} key={"0"}></TabPane>
             {adapterData?.map((adapter, index) => (
               <TabPane
-                tab={<span>{adapter?.llm_model || adapter?.profile_name}</span>}
+                tab={
+                  <Tooltip title={adapter?.llm_model}>
```
jaseemjaskp: **LOW / nit — same as `JsonView.jsx`.** When `profile_name` is absent the span and the tooltip both show `llm_model` (redundant); when `llm_model` is also undefined the tooltip is empty and a both-empty case yields a blank, anonymous tab. Mirror the same guard here to keep the two files consistent:
```jsx
tab={
  <Tooltip title={adapter?.profile_name ? adapter?.llm_model : undefined}>
    <span>{adapter?.profile_name || adapter?.llm_model || "Unnamed profile"}</span>
  </Tooltip>
}
```
Non-blocking.

frontend/src/components/custom-tools/combined-output/JsonView.jsx
jaseemjaskp: **MEDIUM — Tooltip can render empty, and the tab label can be blank.**

`llm_model` is the field most prone to being `undefined` here (lookup miss in `getLLMModelNamesForProfiles`, `GetStaticData.js:484-490`, when the profile's adapter isn't in the list). When it's `undefined`, `<Tooltip title={undefined}>` shows nothing on hover — defeating the stated goal of keeping the model name discoverable.

Separately, `{adapter?.profile_name || adapter?.llm_model}` (next line) renders an empty `<span>` when **both** are missing, producing a blank, unidentifiable tab (still keyed by `profile_id`, so output loads but the user can't tell which profile it is).

Suggest terminal fallbacks so neither is ever empty:
```jsx
<Tooltip title={adapter?.llm_model || adapter?.profile_name}>
  <span>
    {adapter?.profile_name || adapter?.llm_model || `Profile ${adapter?.profile_id ?? "?"}`}
  </span>
</Tooltip>
```

frontend/src/components/custom-tools/output-for-doc-modal/OutputForDocModal.jsx
jaseemjaskp: **MEDIUM — same robustness gap as `JsonView.jsx` (mirror this file's fix).**

The tooltip binds to `llm_model`, the field most likely to be `undefined` (lookup miss in `getLLMModelNamesForProfiles`, `GetStaticData.js:484-490`); when absent the tooltip shows nothing on hover. And `{adapter?.profile_name || adapter?.llm_model}` (next line) renders an empty tab label when both fields are missing.

Apply the same terminal-fallback fix as in `JsonView.jsx` so neither the tooltip nor the label can be empty:
```jsx
<Tooltip title={adapter?.llm_model || adapter?.profile_name}>
  <span>{adapter?.profile_name || adapter?.llm_model || `Profile ${index + 1}`}</span>
</Tooltip>
```

jaseemjaskp (commented):
Automated PR review (PR Review Toolkit: Code Reviewer, Silent Failure Hunter, Type Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier).

**Verdict: ship-able.** The change is correct, symmetric across both files, null-safe (optional chaining), and matches the stated intent (profile name as label, model in tooltip). No blocking issues.

Two MEDIUM robustness findings are inline below. The recurring root cause is that `llm_model` — the field the new tooltip relies on — is exactly the field most likely to be `undefined`: `getLLMModelNamesForProfiles` (`frontend/src/helpers/GetStaticData.js:484-490`) derives it from `adapterMap[profile?.llm]`, which misses whenever the profile's adapter isn't in the returned adapter list (deleted/renamed adapter, or a partial list).

Non-blocking, no inline comment posted:
- **Type/data contract (weak):** `adapterData` is declared only as `PropTypes.array` (`JsonView.jsx:111`); the `{ profile_name, llm_model, profile_id }` shape lives implicitly in the producer. The new behavior now depends on `profile_name`, but that contract is expressed nowhere. Cheap fix: `PropTypes.arrayOf(PropTypes.shape({ profile_name: PropTypes.string, llm_model: PropTypes.string, profile_id: PropTypes.string }))`, plus a JSDoc `@returns` typedef on the helper (the shape also reaches `OutputForDocModal` via local state, which PropTypes can't cover).
- **Comment (optional):** A one-line comment explaining the `profile_name || llm_model` label + `llm_model` tooltip intent would guard the fallback from being misread as redundant/dead code.
- **Simplification:** None warranted — extracting a shared `<ProfileTabLabel>` for a 4-line snippet duplicated twice is disproportionate.
- **Tests:** No test required. Frontend uses Vitest + RTL but only for non-trivial branching logic; this presentational change is consistent with leaving such code untested.

frontend/src/components/custom-tools/combined-output/JsonView.jsx
athul-rs: Fixed in 83ba6813 — the tooltip now falls back (`title={llm_model || profile_name}`) so it is never empty, and the label has a terminal fallback so a tab can never render blank:

```jsx
<Tooltip title={adapter?.llm_model || adapter?.profile_name}>
  <span>{adapter?.profile_name || adapter?.llm_model || `Profile ${adapter?.profile_id ?? "?"}`}</span>
</Tooltip>
```

frontend/src/components/custom-tools/output-for-doc-modal/OutputForDocModal.jsx
athul-rs: Fixed in 83ba6813 — mirrored the same terminal-fallback fix here, using `Profile ${index + 1}` since this file keys tabs by index rather than `profile_id`. Tooltip and label can no longer be empty.

frontend/src/components/custom-tools/combined-output/JsonView.jsx
athul-rs: Superseded — this thread is outdated and the later MEDIUM thread on the same lines is the one I implemented (83ba6813).

Worth noting the two suggestions conflict: this one hides the tooltip when it would be redundant (`title={profile_name ? llm_model : undefined}`), the MEDIUM one guarantees it is never empty (`title={llm_model || profile_name}`). I went with the latter, so on the fallback path the hover does repeat the visible text — a redundant tooltip seemed the lesser evil versus a hover that reveals nothing. The blank-tab case you flagged is fixed either way.

frontend/src/components/custom-tools/output-for-doc-modal/OutputForDocModal.jsx
athul-rs: Superseded by the MEDIUM thread on this file, implemented in 83ba6813 (see the note on the JsonView.jsx thread re: the two conflicting tooltip suggestions). Blank tab label and empty tooltip are both fixed.

PR #2031, opened 2026-06-11
UN-2190 [MISC] Auto-capture execution ID in the API deployment Postman collection

athul-rs:
## What

- The downloaded Postman collection for API deployments now wires the two requests together: the **Process document** request stores `execution_id` from its response into a collection variable via a post-response script, and the **Execution status** request reads `{{execution_id}}` from that variable. No more manual copy-pasting of execution IDs.

## Why

[UN-2190](https://zipstack.atlassian.net/browse/UN-2190) — minor UX gap: after executing a document asynchronously, users had to copy the execution ID from the response and paste it into the status request's `REPLACE_WITH_EXECUTION_ID` placeholder. The LLMWhisperer Postman collection already does this with `whisper_hash`; this applies the same pattern.

## How

- `postman_collection/dto.py`: new `ScriptItem`/`EventItem`/`VariableItem` dataclasses; `PostmanItem` gets an optional `event`; the execute request carries a post-response script (`pm.collectionVariables.set("execution_id", ...)`, guarded so non-JSON/error responses are ignored); the status URL uses `{{execution_id}}` (urlencode with `safe="{}"` so Postman's braces survive); collection-level `variable` block added with a `REPLACE_WITH_EXECUTION_ID` default so manual use still works.
- `to_dict()` strips `event: null` from items that have no scripts.
- Pipeline collections (single request, no status call) are unchanged.

- No. This only changes the generated Postman collection JSON (a download artifact), not any API behavior. The status request still works manually: if the script never ran, the variable resolves to the `REPLACE_WITH_EXECUTION_ID` default — identical to today's placeholder. The script is defensive (checks `response.message.execution_id` exists before setting).

## Database Migrations

- None

## Env Config

- None

## Relevant Docs

- N/A

## Related Issues or PRs

- Jira: UN-2190

## Notes on Testing

- ruff check + format clean.
- Verified generated JSON structure with a stubbed-Django harness: execute item carries the event block, status item has none, status URL contains `execution_id={{execution_id}}` unescaped, and the collection-level variable is present.
- Manual: download a collection from an API deployment, import into Postman, run Process document then Execution status — the second request uses the captured ID.

## Screenshots

N/A

## Checklist

[UN-2190]: https://zipstack.atlassian.net/browse/UN-2190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/api_v2/postman_collection/constants.py
+++ b/backend/api_v2/postman_collection/constants.py
@@ -7,5 +7,7 @@ class CollectionKey:
     STATUS_API_KEY = "Execution status"
     STATUS_EXEC_ID_DEFAULT = "REPLACE_WITH_EXECUTION_ID"
     EXEC_ID_VARIABLE_NAME = "execution_id"
-    STATUS_EXEC_ID_VARIABLE = "{{execution_id}}"
+    # Derive the Postman variable reference from the variable name so the two
+    # can never drift apart (-> "{{execution_id}}").
+    STATUS_EXEC_ID_VARIABLE = f"{{{{{EXEC_ID_VARIABLE_NAME}}}}}"
     AUTH_QUERY_PARAM_DEFAULT = "REPLACE_WITH_API_KEY"
--- a/backend/api_v2/postman_collection/constants.py
+++ b/backend/api_v2/postman_collection/constants.py
@@ -6,4 +6,6 @@ class CollectionKey:
     EXECUTE_PIPELINE_API_KEY = "Process pipeline"
     STATUS_API_KEY = "Execution status"
     STATUS_EXEC_ID_DEFAULT = "REPLACE_WITH_EXECUTION_ID"
+    EXEC_ID_VARIABLE_NAME = "execution_id"
+    STATUS_EXEC_ID_VARIABLE = "{{execution_id}}"
     AUTH_QUERY_PARAM_DEFAULT = "REPLACE_WITH_API_KEY"
--- a/backend/api_v2/postman_collection/dto.py
+++ b/backend/api_v2/postman_collection/dto.py
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 from dataclasses import asdict, dataclass, field
-from typing import Any
+from typing import Any, Literal
 from urllib.parse import urlencode, urljoin
 
 from django.conf import settings
@@ -21,12 +21,12 @@ class HeaderItem:
 @dataclass
 class ScriptItem:
     exec: list[str]
-    type: str = "text/javascript"
+    type: Literal["text/javascript"] = "text/javascript"
 
 
 @dataclass
 class EventItem:
-    listen: str
+    listen: Literal["prerequest", "test"]
     script: ScriptItem
 
 
@@ -73,7 +73,7 @@ class RequestItem:
 class PostmanItem:
     name: str
     request: RequestItem
-    event: list[EventItem] | None = None
+    event: list[EventItem] = field(default_factory=list)
 
 
 @dataclass
@@ -88,7 +88,7 @@ class APIBase(ABC):
     def get_form_data_items(self) -> list[FormDataItem]:
         pass
 
-    def get_collection_variables(self) -> list["VariableItem"]:
+    def get_collection_variables(self) -> list[VariableItem]:
         """Collection-level variables; only needed when a request
         references them.
         """
@@ -184,6 +184,12 @@ class APIDeploymentDto(APIBase):
         """Post-response script that stores the execution_id from the execute
         response into a collection variable, so the status request can use it
         without manual copy-pasting.
+
+        This is Postman's ``test`` hook (the emitted event uses
+        ``listen="test"``). It assumes the execute response shape
+        ``{"message": {"execution_id": ...}}``; if the body is non-JSON or
+        lacks that field it captures nothing and resets the variable to the
+        default sentinel so a stale id from a previous run is never reused.
         """
         return EventItem(
             listen="test",
@@ -197,6 +203,9 @@ class APIDeploymentDto(APIBase):
                     "}",
                     "if (response && response.message && response.message.execution_id) {",  # noqa: E501
                     f'    pm.collectionVariables.set("{CollectionKey.EXEC_ID_VARIABLE_NAME}", response.message.execution_id);',  # noqa: E501
+                    "} else {",
+                    f'    pm.collectionVariables.set("{CollectionKey.EXEC_ID_VARIABLE_NAME}", "{CollectionKey.STATUS_EXEC_ID_DEFAULT}");',  # noqa: E501
+                    '    console.warn("No execution_id in execute response; not reusing a stale value. Status:", pm.response.code);',  # noqa: E501
                     "}",
                 ]
             ),
@@ -299,8 +308,9 @@ class PostmanCollection:
             dict[str, Any]: PostmanCollection as a dict
         """
         collection_dict = asdict(self)
-        # Drop null event blocks; Postman expects "event" to be a list
+        # Omit empty event blocks; items without scripts (e.g. pipelines)
+        # should not carry an "event" key at all.
         for item in collection_dict.get("item", []):
-            if item.get("event") is None:
+            if not item.get("event"):
                 item.pop("event", None)
         return collection_dict
--- a/backend/api_v2/postman_collection/dto.py
+++ b/backend/api_v2/postman_collection/dto.py
@@ -19,6 +19,24 @@ class HeaderItem:
 
 
 @dataclass
+class ScriptItem:
+    exec: list[str]
+    type: str = "text/javascript"
+
+
+@dataclass
+class EventItem:
+    listen: str
+    script: ScriptItem
+
+
+@dataclass
+class VariableItem:
+    key: str
+    value: str
+
+
+@dataclass
 class FormDataItem:
     key: str
     type: str
@@ -55,6 +73,7 @@ class RequestItem:
 class PostmanItem:
     name: str
     request: RequestItem
+    event: list[EventItem] | None = None
 
 
 @dataclass
@@ -137,20 +156,39 @@ class APIDeploymentDto(APIBase):
     def _get_status_api_request(self) -> RequestItem:
         header_list = [HeaderItem(key="Authorization", value=f"Bearer {self.api_key}")]
         status_query_param = {
-            "execution_id": CollectionKey.STATUS_EXEC_ID_DEFAULT,
+            "execution_id": CollectionKey.STATUS_EXEC_ID_VARIABLE,
             ApiExecution.INCLUDE_METADATA: "False",
             ApiExecution.INCLUDE_METRICS: "False",
         }
-        status_query_str = urlencode(status_query_param)
+        # Keep {{...}} unescaped so Postman resolves the collection variable
+        status_query_str = urlencode(status_query_param, safe="{}")
         abs_api_endpoint = urljoin(settings.WEB_APP_ORIGIN_URL, self.api_endpoint)
         status_url = urljoin(abs_api_endpoint, "?" + status_query_str)
         return RequestItem(method=HTTPMethod.GET, header=header_list, url=status_url)
+
+    def _get_execute_capture_event(self) -> EventItem:
+        """Post-response script that stores the execution_id from the execute
+        response into a collection variable, so the status request can use it
+        without manual copy-pasting.
+        """
+        return EventItem(
+            listen="test",
+            script=ScriptItem(
+                exec=[
+                    "const response = pm.response.json();",
+                    "if (response && response.message && response.message.execution_id) {",  # noqa: E501
+                    f'    pm.collectionVariables.set("{CollectionKey.EXEC_ID_VARIABLE_NAME}", response.message.execution_id);',  # noqa: E501
+                    "}",
+                ]
+            ),
+        )
 
     def get_postman_items(self) -> list[PostmanItem]:
         postman_item_list = [
             PostmanItem(
                 name=CollectionKey.EXECUTE_API_KEY,
                 request=self.get_create_api_request(),
+                event=[self._get_execute_capture_event()],
             ),
             PostmanItem(
                 name=CollectionKey.STATUS_API_KEY,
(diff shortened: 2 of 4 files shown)


backend/api_v2/postman_collection/dto.py
```diff
@@ -137,20 +156,39 @@ def get_api_endpoint(self) -> str:
...
+        return EventItem(
+            listen="test",
+            script=ScriptItem(
+                exec=[
+                    "const response = pm.response.json();",
+                    "if (response && response.message && response.message.execution_id) {",  # noqa: E501
+                    f'    pm.collectionVariables.set("{CollectionKey.EXEC_ID_VARIABLE_NAME}", response.message.execution_id);',  # noqa: E501
+                    "}",
+                ]
+            ),
```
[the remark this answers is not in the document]
athul-rs: Fixed — the script now wraps `pm.response.json()` in try/catch and skips capture for non-JSON bodies.

backend/api_v2/postman_collection/dto.py
```diff
@@ -228,7 +267,13 @@ def create(
...
         postman_info: PostmanInfo = data_object.get_postman_info()
         postman_item_list = data_object.get_postman_items()
-        return cls(info=postman_info, item=postman_item_list)
+        variables = [
+            VariableItem(
+                key=CollectionKey.EXEC_ID_VARIABLE_NAME,
+                value=CollectionKey.STATUS_EXEC_ID_DEFAULT,
+            )
+        ]
+        return cls(info=postman_info, item=postman_item_list, variable=variables)
```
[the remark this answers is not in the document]
athul-rs: Fixed — collection variables moved behind `APIBase.get_collection_variables()`; Pipeline collections now serialize with an empty variable list, verified in the stub harness.

backend/api_v2/postman_collection/dto.py
```diff
@@ -137,20 +162,52 @@ def get_api_endpoint(self) -> str:
...
+            script=ScriptItem(
+                exec=[
+                    "let response = null;",
+                    "try {",
+                    "    response = pm.response.json();",
+                    "} catch (error) {",
+                    "    // Non-JSON response (e.g. gateway error); nothing to capture",
+                    "}",
+                    "if (response && response.message && response.message.execution_id) {",  # noqa: E501
+                    f'    pm.collectionVariables.set("{CollectionKey.EXEC_ID_VARIABLE_NAME}", response.message.execution_id);',  # noqa: E501
```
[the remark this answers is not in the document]
athul-rs: Confirmed and fixed — `APIExecutionResponseSerializer` was indeed dropping the DTO's `execution_id`. Added it as a first-class field (additive, non-breaking), which both makes the capture script work and saves API consumers from parsing the ID out of `status_api`. Good catch.

chandrasekharan-zipstack (approved):

jaseemjaskp (commented):

The change is correct and well-scoped — the capture script's `response.message.execution_id` path is validated by the new serializer field and the `{"message": ...}` response envelope, and `urlencode(safe="{}")` correctly preserves `{{execution_id}}`. The already-fixed items (serializer dropping `execution_id`, pipeline collection variables, non-JSON try/catch) are not re-raised.

The findings below are all **new**. The only substantive one is the stale-variable risk (P1); the rest are P2/P3 hardening, type-design, comment, and test-coverage suggestions.

backend/api_v2/postman_collection/dto.py
```diff
@@ -137,20 +162,52 @@ def get_api_endpoint(self) -> str:
...
+            listen="test",
+            script=ScriptItem(
+                exec=[
+                    "let response = null;",
+                    "try {",
+                    "    response = pm.response.json();",
+                    "} catch (error) {",
+                    "    // Non-JSON response (e.g. gateway error); nothing to capture",
+                    "}",
+                    "if (response && response.message && response.message.execution_id) {",  # noqa: E501
```
jaseemjaskp, line 204: **P1 — silent reuse of a stale `execution_id`.** This guard has no `else` branch, and it's distinct from the non-JSON try/catch already fixed below: here the response parses fine but carries no `message.execution_id` — e.g. a 422 validation error or 500 (`{"message": {"error": ...}}`), auth/rate-limit rejection, or any 2xx with an unexpected shape. In that case the script does nothing and `{{execution_id}}` keeps whatever it held — most dangerously the id captured from a *previous successful* run. The user re-runs Execute, it fails, hits Execution status, and Postman silently polls the **old** execution and shows its COMPLETED result as if it belonged to the failed run. No signal that capture was skipped.

Suggest making the miss explicit and preventing cross-run staleness:
```js
if (response && response.message && response.message.execution_id) {
    pm.collectionVariables.set("execution_id", response.message.execution_id);
} else {
    pm.collectionVariables.set("execution_id", "REPLACE_WITH_EXECUTION_ID");
    console.warn("No execution_id in execute response; not reusing a stale value. Status:", pm.response.code);
}
```
Optionally gate capture on `pm.response.code` 2xx so a 200-with-unexpected-shape is surfaced too.

backend/api_v2/postman_collection/dto.py
```diff
@@ -55,6 +73,7 @@ class RequestItem:
 class PostmanItem:
     name: str
     request: RequestItem
+    event: list[EventItem] | None = None
```
jaseemjaskp: **P2 — `event: list[EventItem] | None = None` creates a `None`/`[]` tri-state** where only "some events" / "no events" are meaningful, and it's the sole reason `to_dict()` (lines 301-306) has to post-process and pop `null` event keys. Other "maybe-empty" fields on these DTOs already use `field(default_factory=list)` (`PostmanCollection.item`/`variable`, lines 251-252). Aligning here:
```python
event: list[EventItem] = field(default_factory=list)
```
lets you drop the strip loop and reduce `to_dict` to `return asdict(self)`. Trade-off: items then serialize `"event": []` instead of omitting the key — both are valid Postman v2.1 (empty array is the schema-canonical form). If preserving the exact "key absent" output is a hard requirement, keep the loop but make it robust to the new default by testing `if not item.get("event")` rather than `is None`.

backend/api_v2/postman_collection/dto.py
```diff
@@ -18,6 +18,24 @@ class HeaderItem:
...
 
+@dataclass
+class ScriptItem:
+    exec: list[str]
+    type: str = "text/javascript"
+
+
+@dataclass
+class EventItem:
+    listen: str
```
jaseemjaskp: **P2 (type safety) — constrain the closed enums.** `EventItem.listen` accepts any `str`, but Postman v2.1 only recognizes two hooks; a typo like `"tests"` produces a silently-ignored event with no error anywhere in this pipeline. Narrow it to `Literal["prerequest", "test"]`. Same idea for `ScriptItem.type` (line 24): only `"text/javascript"` is meaningful, so `Literal["text/javascript"]` (or keep the default and type it as such). Zero runtime cost, caught by mypy, self-documenting.

backend/api_v2/postman_collection/dto.py
```diff
@@ -137,20 +162,52 @@ def get_api_endpoint(self) -> str:
...
+    def get_collection_variables(self) -> list[VariableItem]:
+        return [
+            VariableItem(
+                key=CollectionKey.EXEC_ID_VARIABLE_NAME,
+                value=CollectionKey.STATUS_EXEC_ID_DEFAULT,
+            )
+        ]
+
+    def _get_execute_capture_event(self) -> EventItem:
+        """Post-response script that stores the execution_id from the execute
```
jaseemjaskp, line 184: **P3 (comment) — docstring omits the load-bearing response-shape coupling.** The script silently no-ops unless the response is `{message: {execution_id}}`; if that envelope ever changes (e.g. `message` renamed or `execution_id` relocated) capture breaks quietly. Add a line so a future API change flags this script, e.g. _"Assumes the execute response has shape `{message: {execution_id}}`; if the body is non-JSON or lacks that field, nothing is captured."_ Minor: "Post-response script" matches Postman's current UI label but the code emits `listen="test"` — a half-sentence tying the two ("post-response, i.e. Postman's `test` hook") avoids a maintainer double-take.

backend/api_v2/postman_collection/dto.py
```diff
@@ -69,6 +88,12 @@ class APIBase(ABC):
     def get_form_data_items(self) -> list[FormDataItem]:
         pass
 
+    def get_collection_variables(self) -> list["VariableItem"]:
```
jaseemjaskp: **P3 (nit) — unnecessary string forward-reference.** `VariableItem` is already defined above (line 33), so the quotes in `-> list["VariableItem"]` aren't needed; the override at line 175 already uses the bare `list[VariableItem]`. Use the bare type for consistency.

backend/api_v2/postman_collection/constants.py
```diff
@@ -6,4 +6,6 @@ class CollectionKey:
     EXECUTE_PIPELINE_API_KEY = "Process pipeline"
     STATUS_API_KEY = "Execution status"
     STATUS_EXEC_ID_DEFAULT = "REPLACE_WITH_EXECUTION_ID"
+    EXEC_ID_VARIABLE_NAME = "execution_id"
```
jaseemjaskp, line 9: **P3 — unenforced coupling between the variable name and its `{{...}}` reference.** `EXEC_ID_VARIABLE_NAME = "execution_id"` and `STATUS_EXEC_ID_VARIABLE = "{{execution_id}}"` must agree (the status URL references `{{execution_id}}`, the capture JS sets `"execution_id"`, and the collection variable keys on it). If one is renamed and the other isn't, the collection breaks silently — the status request would query a never-set variable. Derive one from the other so they can't drift:
```python
STATUS_EXEC_ID_VARIABLE = f"{{{{{EXEC_ID_VARIABLE_NAME}}}}}"  # -> "{{execution_id}}"
```

backend/api_v2/postman_collection/dto.py
```diff
@@ -237,4 +299,8 @@ def to_dict(self) -> dict[str, Any]:
             dict[str, Any]: PostmanCollection as a dict
         """
         collection_dict = asdict(self)
+        # Drop null event blocks; Postman expects "event" to be a list
```
jaseemjaskp: **P2 (test coverage) — `backend/api_v2/postman_collection/` has no tests at all**, and this PR's trickiest logic is exactly the kind that fails only at Postman-import time, not in CI. `to_dict()` (and `create()`) are pure given `settings.WEB_APP_ORIGIN_URL` + an api_key, so this is cheap to lock down. Suggest adding `backend/api_v2/postman_collection/tests/test_dto.py` covering: (a) `to_dict` strips `null` event yet keeps the populated execute event and emits the `variable` array; (b) **regression guard** that `PipelineDto` collections carry no `event` key and `variable == []` while `APIDeploymentDto` carries both; (c) the variable key, the `{{execution_id}}` in the status URL, and the `pm.collectionVariables.set(...)` in the JS all reference the same `CollectionKey.EXEC_ID_VARIABLE_NAME` (assert via the constant, not a hardcoded literal, so a partial rename fails); (d) the status URL keeps `execution_id={{execution_id}}` literal (no `%7B`). Assert load-bearing substrings, not the whole `exec` array, to stay robust to comment/wording edits.

backend/api_v2/postman_collection/dto.py
athul-rs: Fixed in f6d7d161 — added the `else` branch, so a response that parses but carries no `message.execution_id` now resets the variable to the `REPLACE_WITH_EXECUTION_ID` sentinel and logs a warning with the status code. A stale id from a previous successful run can no longer be silently reused by the status request. Covered by a test that asserts the reset path is present.

backend/api_v2/postman_collection/dto.py
athul-rs: Fixed in f6d7d161 — `event` is now `field(default_factory=list)`, matching the other maybe-empty fields on these DTOs. The `to_dict()` strip loop now tests `if not item.get("event")` rather than `is None`, so pipeline items still omit the key entirely and the exact output shape is preserved. Locked down by a regression test (pipelines carry no `event` key; API deployments carry one).

backend/api_v2/postman_collection/dto.py
athul-rs: Fixed in f6d7d161 — `EventItem.listen` is now `Literal["prerequest", "test"]` and `ScriptItem.type` is `Literal["text/javascript"]`. A typo like `"tests"` is now a mypy error instead of a silently-ignored event.

backend/api_v2/postman_collection/dto.py
athul-rs: Fixed in f6d7d161 — the docstring now spells out the assumed `{"message": {"execution_id": ...}}` envelope, says nothing is captured (and the variable is reset) when the body is non-JSON or lacks the field, and ties "post-response" to the emitted `listen="test"` hook.

backend/api_v2/postman_collection/dto.py
athul-rs: Fixed in f6d7d161 — dropped the quotes, now `-> list[VariableItem]`, consistent with the override.

backend/api_v2/postman_collection/constants.py
athul-rs: Fixed in f6d7d161 — `STATUS_EXEC_ID_VARIABLE` is now derived: `f"{{{{{EXEC_ID_VARIABLE_NAME}}}}}"`. The two can no longer drift, and a test asserts the status URL, the collection variable key and the capture JS all resolve from the same constant.

backend/api_v2/postman_collection/dto.py
athul-rs: Done in f6d7d161 — added `backend/api_v2/postman_collection/tests/test_dto.py` (8 tests, all passing locally) covering exactly the cases you listed: `to_dict` strips the empty event yet keeps the populated execute event and the `variable` array; the pipeline-vs-API-deployment regression guard; the variable key / `{{execution_id}}` / `pm.collectionVariables.set(...)` all asserted via `CollectionKey.EXEC_ID_VARIABLE_NAME` rather than literals so a partial rename fails; and the status URL keeping `execution_id={{execution_id}}` unescaped. Assertions target load-bearing substrings, not the whole `exec` array.

PR #2032, opened 2026-06-11
UN-2771 [FEAT] Report text extraction time in API deployment metrics

athul-rs:
## What

- API deployment responses with `include_metrics=True` now report the text extraction (LLMWhisperer/X2Text) duration as `metrics._pipeline.extraction["time_taken(s)"]`, alongside the existing per-output indexing time.

```json
"metrics": { "_pipeline": { "extraction": { "time_taken(s)": 1.23 } } }
```

The metric lives under a reserved `_pipeline` namespace rather than a bare top-level `extraction` key: sibling keys in the metrics dict are user-defined output/prompt names, so a prompt named "extraction" would otherwise collide with it.

## Why

[UN-2771](https://zipstack.atlassian.net/browse/UN-2771) — metrics only included indexing time; the LLMWhisperer call (often the dominant cost for large documents) was invisible, making it hard to attribute slow executions.

## How

**Reworked per review (@chandrasekharan-zipstack):** the original version added timing to `tools/structure` (deprecated Docker path); that's reverted. The change now lives in the live celery flow: `LegacyExecutor._handle_structure_pipeline` times the extract step (`time.monotonic()`) and `_finalize_pipeline_result` merges `{"extraction": {"time_taken(s)": ...}}` into result metrics via the existing `_merge_pipeline_metrics`, exactly mirroring how `_index_pipeline_output` records indexing time. The Prompt Studio IDE flow (`_handle_ide_index`) is untouched.

- No. Additive metrics key only, under a reserved `_pipeline` namespace that cannot collide with user-defined output names; existing `indexing` metrics and response shape unchanged. Paths that skip extraction (smart-table) leave the dict empty — nothing added, same as today.
- The indexing path was moved from wall-clock `datetime.now()` to `time.monotonic()`. Same units and semantics (elapsed seconds as a float), but immune to NTP/system-clock adjustments; it cannot regress a caller.

## Database Migrations

- None

## Env Config

- None

## Relevant Docs

- N/A

## Related Issues or PRs

- Jira: UN-2771

## Notes on Testing

- 5 new unit tests in `workers/tests/test_phase5d.py::TestExtractionMetrics`: metric recorded, name-collision guard (a prompt named "extraction"), index+extraction merge, skip-extraction records nothing, extract-failure records nothing.
- Worker suite: `723 passed, 6 failed` — the same 6 failures present on `main` (Postgres auth + pre-existing mock assertions), none in the touched paths.
- ruff check/format clean (pre-commit hooks pass).
- Manual: run an API deployment (structure-tool workflow routed via the worker pipeline) with `include_metrics=True` → response metrics contain `_pipeline.extraction.time_taken(s)`; smart-table path unchanged.

## Screenshots

N/A

## Checklist

[UN-2771]: https://zipstack.atlassian.net/browse/UN-2771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


chandrasekharan-zipstack (commented):
@athul-rs I like your intent here however I don't think the tool related code or logic is being used anymore. This is deprecated and in fact we'll be removing this code soon. Can you check if this is handled in the equivalent flow involving the celery tasks and ensure this concern is addressed?
cc: @harini-venkataraman

athul-rs:
@chandrasekharan-zipstack you were right — reworked. The `tools/structure` change (and the tool version bump) is reverted; extraction timing now lives in the celery flow: `LegacyExecutor._handle_structure_pipeline` times the extract step and merges `extraction.time_taken(s)` into result metrics through `_finalize_pipeline_result`/`_merge_pipeline_metrics`, mirroring the existing indexing timing in `_index_pipeline_output`. Confirmed the worker path is what cloud runs (workflows using the structure tool image are routed to it via `_is_structure_tool_workflow`). PR is now a 12-line change to `legacy_executor.py` only.

workers/executor/executors/legacy_executor.py
```diff
@@ -640,6 +642,9 @@ def _failure(child_result: ExecutionResult) -> ExecutionResult:
                     return _failure(extract_result)
                 _absorb(extract_result)
                 extracted_text = extract_result.data.get(IKeys.EXTRACTED_TEXT, "")
+                extraction_metrics = {
+                    "extraction": {"time_taken(s)": time.monotonic() - extraction_start}
```
jaseemjaskp: **Maintainability (distinct from the key-collision note on L647):** this hardcodes the literal `"extraction"`, but `PromptServiceConstants.EXTRACTION = "extraction"` already exists and is imported as `PSKeys` in this file — use `PSKeys.EXTRACTION` so the metric key and the constant can't drift. Separately, `"time_taken(s)"` is duplicated here and at L1031 (`index_metrics`); consider promoting it to a shared constant.

Minor readability: for parity with the indexing path (`elapsed = ...; index_metrics[output_name] = {"indexing": {"time_taken(s)": elapsed}}`, L1029-1031), capture the duration in a named local first:
```python
extraction_time = time.monotonic() - extraction_start
extraction_metrics = {PSKeys.EXTRACTION: {"time_taken(s)": extraction_time}}
```

workers/executor/executors/legacy_executor.py
```diff
@@ -619,10 +619,12 @@ def _failure(child_result: ExecutionResult) -> ExecutionResult:
         )
         step = 1
 
+        extraction_metrics: dict = {}
         try:
             # ---- Step 1: Extract ----
             if not skip_extraction:
                 step += 1
+                extraction_start = time.monotonic()
```
jaseemjaskp, line 627: **Clock-source consistency:** extraction times the step with `time.monotonic()`, while the sibling indexing metric uses wall-clock `datetime.datetime.now()` (L1029). `time.monotonic()` is the *correct* choice for duration measurement (immune to NTP / system-clock adjustments), so this code is right — but the two producers of `"time_taken(s)"` now use different clocks. Worth a follow-up to align the indexing path onto `monotonic()` too.

workers/executor/executors/legacy_executor.py
```diff
@@ -787,17 +793,21 @@ def _finalize_pipeline_result(
         source_file_name: str,
         extracted_text: str,
         index_metrics: dict,
+        extraction_metrics: dict | None = None,
```
jaseemjaskp: The new `extraction_metrics` param isn't reflected in the docstring. Its shape differs from `index_metrics` in a load-bearing way: `extraction_metrics` is a **top-level** `{"extraction": {"time_taken(s)": float}}` (a sibling of per-output keys), whereas `index_metrics` is keyed per output `{output_name: {"indexing": {...}}}`. A short `Args:` note documenting both shapes (and that this is `None`/empty when extraction is skipped) would stop a future maintainer from incorrectly nesting/merging them. Consider typing it as `dict[str, dict] | None` or a small `TypedDict` rather than bare `dict`.

jaseemjaskp (commented):
Automated review (PR Review Toolkit) of the extraction-time metric.

**Net assessment:** the merge logic is correct, empty-dict-safe, and introduces no silent failures — error propagation (`LegacyExecutorError` re-raise) is byte-for-byte unchanged. No blocking bugs. The substantive open item is the top-level-key collision already flagged by @coderabbitai on L647; the inline notes below are distinct consistency/maintainability/test points.

**Biggest gap — test coverage (not inline-able; `test_phase5d.py` isn't in this diff):** the PR adds zero tests for the new behavior. `_finalize_pipeline_result` / `_merge_pipeline_metrics` are exercised by `test_phase5d.py` (`test_index_metrics_merged` L261, `test_merge_*` L915-927) but none assert on `extraction_metrics`. Recommend adding before merge:
- `test_extraction_metrics_recorded` — normal run: `result.data["metrics"]["extraction"]["time_taken(s)"]` is a `float >= 0` and coexists with per-prompt keys.
- `test_extract_failure_records_no_extraction_metric` — metric is computed *after* the `_failure` early-return, so a failed extract must record no timing (pin via a `_finalize_pipeline_result` spy / `assert_not_called`).
- `test_index_and_extraction_metrics_merged` — both families merge (top-level `extraction` + per-output `indexing`).
- `test_skip_extraction_records_no_extraction_metric` — `skip_extraction` ⇒ no `"extraction"` key.

workers/executor/executors/legacy_executor.py
```diff
@@ -640,6 +642,9 @@ def _failure(child_result: ExecutionResult) -> ExecutionResult:
                     return _failure(extract_result)
                 _absorb(extract_result)
                 extracted_text = extract_result.data.get(IKeys.EXTRACTED_TEXT, "")
+                extraction_metrics = {
+                    "extraction": {"time_taken(s)": time.monotonic() - extraction_start}
+                }
```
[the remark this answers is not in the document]
athul-rs: Fixed — and this turned out to be the load-bearing design point of the PR.

The metric is now nested under a reserved `_pipeline` namespace (`PSKeys.PIPELINE = "_pipeline"`), so the response is `metrics._pipeline.extraction["time_taken(s)"]` rather than a bare top-level `extraction` key. Sibling keys in that dict really are user-defined output/prompt names (`index_metrics[output_name]`), so a prompt literally named "extraction" would have collided.

Added a regression test pinning exactly this: `test_extraction_metric_does_not_collide_with_output_name` runs a prompt named "extraction" and asserts the user prompt metrics and the pipeline timing both survive intact.

workers/executor/executors/legacy_executor.py
athul-rs: Both done.

- `PSKeys.EXTRACTION` replaces the hardcoded literal (and the new namespace key uses `PSKeys.PIPELINE`).
- `"time_taken(s)"` promoted to `PSKeys.TIME_TAKEN`, used at both producers in this file — the extraction metric and `index_metrics` (L1031).
- Duration captured in a named local (`extraction_time`) for parity with the indexing path.

Note: `retrieval.py` still has two raw `"time_taken(s)"` literals. I left those out to keep this diff scoped, but they are now one constant away from being consistent.

workers/executor/executors/legacy_executor.py
athul-rs: Aligned it here rather than deferring — you were right that `monotonic()` is the correct clock, and leaving the two producers of `time_taken(s)` on different ones was a trap.

`_index_pipeline_output` now uses `time.monotonic()` instead of wall-clock `datetime.datetime.now()`. `indexing_start` fed nothing but the duration and `datetime` was used nowhere else in the module, so the function-local `import datetime` is gone too. Both metrics are now immune to NTP/system-clock adjustments.

workers/executor/executors/legacy_executor.py
athul-rs: Done — `_finalize_pipeline_result` now carries an `Args:` note spelling out that `extraction_metrics` is a top-level `{"_pipeline": {"extraction": {"time_taken(s)": float}}}`, distinct from the per-output `{output_name: {"indexing": {...}}}` shape of `index_metrics`, and that it is `None`/empty when extraction is skipped. Typed as `dict[str, dict] | None` per your suggestion.

Skipped the TypedDict — the shape is now pinned by tests, and it would be the only TypedDict in the module.

athul-rs:
Thanks @jaseemjaskp — all four inline threads plus the test-coverage gap are addressed in 81d01ed1. Summary of what changed:

**Design (the substantive one):** the metric is no longer a bare top-level `extraction` key. It is nested under a reserved `_pipeline` namespace, so the response shape is now:

```json
"metrics": { "_pipeline": { "extraction": { "time_taken(s)": 1.23 } } }
```

Sibling keys in that dict are user-defined output/prompt names, so the old shape would have collided with a prompt named "extraction". PR description updated to match.

**Tests — all four you listed, plus one for the collision** (`test_phase5d.py::TestExtractionMetrics`):
- `test_extraction_metrics_recorded` — normal run, `time_taken(s)` is a `float >= 0`.
- `test_extraction_metric_does_not_collide_with_output_name` — a prompt named "extraction" and the pipeline metric coexist untouched.
- `test_index_and_extraction_metrics_merged` — per-output `indexing` + top-level `_pipeline.extraction` both survive the merge.
- `test_skip_extraction_records_no_extraction_metric` — smart-table path adds no `_pipeline` key.
- `test_extract_failure_records_no_extraction_metric` — timing is taken after the failure early-return, pinned with a `_finalize_pipeline_result` spy and `assert_not_called`.

**Also fixed:** `PSKeys.EXTRACTION` / `PSKeys.PIPELINE` instead of literals; `"time_taken(s)"` promoted to `PSKeys.TIME_TAKEN`; and the indexing path moved off wall-clock `datetime.now()` onto `time.monotonic()` so both producers use the same (correct) clock — I did this here rather than deferring it, since the PR was what introduced the inconsistency.

**Verification:** worker suite is `723 passed, 6 failed` — the same 6 failures present on `main` (Postgres auth + pre-existing mock assertions), none in the touched paths. Branch is up to date with main.

workers/executor/executors/constants.py
```diff
@@ -51,6 +51,11 @@ class PromptServiceConstants:
     CHALLENGE = "challenge"
     ENABLE_CHALLENGE = "enable_challenge"
     EXTRACTION = "extraction"
+    # Reserved namespace for pipeline-level metrics so they cannot collide with
+    # user-defined output/prompt names at the top level of the metrics dict.
+    PIPELINE = "_pipeline"
```
chandrasekharan-zipstack: Rename `PIPELINE = "_pipeline"` → `FILE = "_file"`. This bucket holds file-level (whole-document) metrics, and `"file"` is the existing API response key for the source file (`workflow_manager/workflow_v2/dto.py:45`) — `_file` reads intuitively and matches convention.

Also add a sibling constant for the metric name so it doesn't collide with `extraction_llm`:
```python
TEXT_EXTRACTION = "text_extraction"
```
Update the comment above to match.

workers/executor/executors/legacy_executor.py
```diff
@@ -640,6 +642,15 @@ def _failure(child_result: ExecutionResult) -> ExecutionResult:
                     return _failure(extract_result)
                 _absorb(extract_result)
                 extracted_text = extract_result.data.get(IKeys.EXTRACTED_TEXT, "")
+                extraction_time = time.monotonic() - extraction_start
+                # Nest under a reserved "_pipeline" namespace so the metric never
+                # collides with a user-defined output named "extraction" during
+                # the top-level metrics merge (see _merge_pipeline_metrics).
+                extraction_metrics = {
+                    PSKeys.PIPELINE: {
```
chandrasekharan-zipstack: Use the renamed constants here:
```python
extraction_metrics = {
    PSKeys.FILE: {
        PSKeys.TEXT_EXTRACTION: {PSKeys.TIME_TAKEN: extraction_time}
    }
}
```
Also update the adjacent `# Nest under a reserved "_pipeline"…` comment and the `_finalize_pipeline_result` docstring (the `{"_pipeline": {"extraction": …}}` shape example) to `_file` / `text_extraction`.

chandrasekharan-zipstack (commented):
Agreed follow-ups from a design discussion on the metric shape. Two mechanical renames for clarity; plus a heads-up: I filed **UN-3714** for the per-profile X2Text extractor being collapsed to the default profile's extractor in the worker pipeline (out of scope for this PR).

1. `_pipeline` → `_file` — the bucket holds file-level (whole-document) metrics, and `"file"` is already the API response key for the source file (`workflow_manager/workflow_v2/dto.py:45`), so `_file` is the intuitive/consistent name.
2. `extraction` → `text_extraction` — avoids confusion with `extraction_llm` (the extraction-purpose LLM call) which sits beside it in the same metrics dict; the collision is visible in single-pass responses (`extraction_llm` and `_pipeline.extraction` coexist).

Single-pass needs no special-casing: `_file.text_extraction` is merged in by the shared `_finalize_pipeline_result`, so single-pass metrics stay flat and just gain the `_file` sibling — same convention as multi-prompt.

workers/tests/test_phase5d.py
```diff
@@ -304,6 +304,126 @@ def test_index_metrics_merged(self, executor):
         assert "indexing" in metrics["field_a"]
 
 
+# ---------------------------------------------------------------------------
+# Tests — Extraction timing metric (UN-2771)
+# ---------------------------------------------------------------------------
+
+
+class TestExtractionMetrics:
+    """Extraction duration is reported under the reserved _pipeline namespace.
```
chandrasekharan-zipstack: Update these tests for the renamed keys: `metrics["_pipeline"]["extraction"]` → `metrics["_file"]["text_extraction"]`, the `"_pipeline" not in metrics` assertion → `"_file"`, and this docstring's `_pipeline namespace` wording.

Worth adding one single-pass assertion too: since `_file.text_extraction` is merged by the shared finalize, a single-pass run should show the same `_file` bucket alongside its flat `context_retrieval` / `extraction_llm` — locks in the cross-mode consistency.

workers/executor/executors/constants.py
athul-rs: Done in 214d1a82.

`PIPELINE = "_pipeline"` → `FILE = "_file"`, comment updated to say the bucket holds file-level (whole-document) metrics and mirrors the `"file"` key of the API response.

Added `TEXT_EXTRACTION = "text_extraction"` as a sibling rather than reusing `EXTRACTION` — worth noting `PSKeys.EXTRACTION` is already in use as the LLM *usage reason* at `legacy_executor.py:1956`, so the two really are distinct concepts and `EXTRACTION` stays as-is.

One thing to flag: `FILE = "_file"` now sits beside `FILE_NAME`/`FILE_PATH`/`FILE_HASH`, whose values are all plain `file_*` strings. The comment above it explains the leading underscore, but say the word if you would rather it read `FILE_METRICS`.

workers/executor/executors/legacy_executor.py
athul-rs: Done in 214d1a82. The metric now reads:

```python
extraction_metrics = {
    PSKeys.FILE: {
        PSKeys.TEXT_EXTRACTION: {PSKeys.TIME_TAKEN: extraction_time}
    }
}
```

The adjacent comment and the `_finalize_pipeline_result` docstring both now describe the `{"_file": {"text_extraction": {"time_taken(s)": float}}}` shape. No `_pipeline` or `PSKeys.PIPELINE` references remain anywhere under `workers/`.

workers/tests/test_phase5d.py
athul-rs: Done in 214d1a82. Note the file moved: `main` renamed `test_phase5d.py` → `test_structure_pipeline.py`, so these tests now live there.

Keys and docstring wording updated to `_file` / `text_extraction`, including the `"_pipeline" not in metrics` assertion. The collision test now uses a prompt named `text_extraction` rather than `extraction` — under the new naming, a prompt called "extraction" no longer collides with anything, so leaving it would have left a guard that guards nothing.

Added `test_single_pass_reports_same_file_metric` for the cross-mode consistency point: single pass skips indexing but shares `_finalize_pipeline_result`, so the test asserts its flat `context_retrieval` / `extraction_llm` survive untouched beside the same `_file` bucket the multi-prompt path produces.

Verification: full worker suite is **704 passed, 0 failed**. The 6 failures mentioned earlier in this PR are no longer present — they were fixed on `main` in the meantime, not by this branch.

chandrasekharan-zipstack (approved):

workers/executor/executors/legacy_executor.py
```diff
@@ -640,6 +642,15 @@ def _failure(child_result: ExecutionResult) -> ExecutionResult:
...
                 _absorb(extract_result)
                 extracted_text = extract_result.data.get(IKeys.EXTRACTED_TEXT, "")
+                extraction_time = time.monotonic() - extraction_start
+                # Nest under a reserved "_file" namespace so the metric never
+                # collides with a user-defined output named "text_extraction"
+                # during the top-level metrics merge (see _merge_pipeline_metrics).
+                extraction_metrics = {
+                    PSKeys.FILE: {
+                        PSKeys.TEXT_EXTRACTION: {PSKeys.TIME_TAKEN: extraction_time}
+                    }
```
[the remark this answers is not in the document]
athul-rs: Verified this against the code — the finding is correct on both halves. Resolving it as valid-but-out-of-scope rather than fixing it here, with the reasoning on the record.

**All-agentic deployments** — `workers/file_processing/structure_tool_task.py:540`. When every prompt is `agentic_table`, `regular_outputs` is empty and the legacy pipeline is skipped entirely; the result is `{"output": agentic_results, "metadata": {"agentic_only": True}}`, which carries no `metrics` key at all. So `_file.text_extraction` is absent, exactly as described.

**Mixed deployments** — `structure_tool_task.py:485-491`. The legacy pipeline runs and times its own X2Text, but the agentic dispatch consumes only `at_result.data["output"]["tables"]` and discards the rest of the result, metrics included. The agentic runner does its own X2Text (comment at :495-497), so that duration is never counted. Under-reports, as described.

**Why it is not being fixed in this PR.** The `agentic_table` executor is a cloud plugin and is not in this repository — `workers/executor/executors/plugins/` holds only `loader.py`, `protocols.py` and `text_processor.py`, and `protocols.py` defines contracts for Highlight, Challenge, Evaluation and LookupEnrichment, none for `agentic_table`. There is no contract here stating whether it reports an extraction duration, or under what key; the tests mock it returning a bare `{"output": ...}`.

The one alternative available in-repo — wall-clocking the agentic dispatch — would measure the whole agentic run, LLM table extraction included, and label it `text_extraction`. On a metrics feature a wrong number is worse than a missing one, so that is not a fix.

`_file.text_extraction` is scoped to the legacy pipeline's X2Text step. That is consistent with how this PR already treats the smart-table path, where extraction is skipped and no metric is emitted by design (`test_skip_extraction_records_no_extraction_metric`).

Agentic extraction timing is tracked as a follow-up: it needs the plugin to emit a duration, and a contract for it, before anything here can surface it.

PR #2033, opened 2026-06-11
UN-3534 [FEAT] PG Queue Phase 8a — queue-transport routing gate + scaffold

muhammad-ali-e:
## What

Adds the Strangler-Fig **queue-transport routing gate** that lets PG Queue (PGMQ) coexist with Celery, so task types can be migrated one at a time. **Scaffold only** — the PG branch is a Celery-routing stub (no PG consumer exists yet), so this is **zero-behaviour-change by construction**.

Targets the `feat/UN-3445-pg-queue-integration` integration branch (not `main`) — PG-queue work accumulates there until the full path is vetted.

## Changes

- **`queue_backend/routing.py`** *(new)* — `QueueBackend{CELERY,PG}` enum + `select_backend(task_name)` reading the `WORKER_PG_QUEUE_ENABLED_TASKS` allow-list. Tolerant comma-separated parsing (trims whitespace, drops blanks); empty/unset → all Celery; never raises.
- **`queue_backend/dispatch.py`** — consults `select_backend()` and logs the decision at DEBUG. The `send_task` call sits **outside** the `QueueBackend.PG` branch, so the wire is **byte-identical** regardless of the routing decision. PG-selected tasks still dispatch via Celery.
- **`queue_backend/pg_queue/__init__.py`** *(new)* — scaffold subpackage. PGMQ is a core transport substrate, so it lives in the seam beside `dispatch`/`routing`/`barrier`, not under the git-ignored `plugins/` cloud overlay.
- **`queue_backend/__init__.py`** — exports `QueueBackend` + `select_backend`; docstring documents the routing axis as orthogonal to the barrier axis.
- **`sample.env`** — documents the flag (default-safe, OSS-friendly, no Flipt server required).
- **tests** — 12 routing tests including the byte-identical-dispatch characterisation pinning the inert-scaffold invariant; seam `__all__` surface updated.

## Why a per-task allow-list (not a global boolean)

A single global switch would move all traffic at once — the big-bang migration we explicitly avoid. The allow-list migrates one task type at a time and rolls back instantly by removing an entry. Per-org routing is intentionally deferred to the rollout phase; it slots in behind `select_backend()` as an additive change when tenant-level canarying is needed.

## Scope guards

- **Barrier axis untouched** — `WORKER_BARRIER_BACKEND` stays `chord`. Separate concern.
- No call-site changes; no new mandatory worker; system works with and without the flag set.
- Out of scope (later phases): real PGMQ enqueue/dequeue, consumer poll loop, orchestrator.

## Testing

- 45 queue_backend unit tests green (12 new routing tests); ruff clean.
- Dev-tested on a running worker stack: opting `send_webhook_notification` into the allow-list produced the routing-selected DEBUG line while the task still dispatched via Celery and the notification delivered normally. Unset → no line, normal behaviour. New code loaded into all workers without breaking startup.

Ticket: UN-3534 (parent UN-3533, epic UN-3445)

### DIFF
--- a/workers/queue_backend/dispatch.py
+++ b/workers/queue_backend/dispatch.py
@@ -19,6 +19,11 @@ from .routing import QueueBackend, select_backend
 
 logger = logging.getLogger(__name__)
 
+# Task names already logged as PG-routed in this process. Bounds the
+# routing log to once per task name (per prefork child) so an opted-in
+# high-throughput task doesn't log on every dispatch.
+_pg_routing_logged: set[str] = set()
+
 
 def dispatch(
     task_name: str,
@@ -42,8 +47,14 @@ def dispatch(
     outside it guarantees today's wire is byte-identical regardless of
     the routing decision.
     """
-    if select_backend(task_name) is QueueBackend.PG:
-        logger.debug(
+    if (
+        select_backend(task_name) is QueueBackend.PG
+        and task_name not in _pg_routing_logged
+    ):
+        # INFO (not DEBUG) so the cutover is visible under a default log
+        # config; log-once per task name keeps the volume bounded.
+        _pg_routing_logged.add(task_name)
+        logger.info(
             "PG-queue routing selected for task=%r; dispatching via "
             "Celery (scaffold — no PG consumer yet)",
             task_name,
--- a/workers/queue_backend/dispatch.py
+++ b/workers/queue_backend/dispatch.py
@@ -7,6 +7,7 @@ call sites.
 
 from __future__ import annotations
 
+import logging
 from collections.abc import Mapping, Sequence
 from typing import Any
 
@@ -14,6 +15,9 @@ from celery import current_app
 
 from .fairness import FairnessKey
 from .handle import DispatchHandle
+from .routing import QueueBackend, select_backend
+
+logger = logging.getLogger(__name__)
 
 
 def dispatch(
@@ -28,7 +32,23 @@ def dispatch(
 
     ``fairness`` is attached as the ``x-fairness-key`` header (not in
     kwargs). Pass ``None`` for non-workflow worker tasks.
+
+    The transport is chosen by :func:`select_backend` from the PG-queue
+    routing table (``WORKER_PG_QUEUE_ENABLED_TASKS``). In this phase the
+    table is a scaffold: PG-selected tasks are *logged* but still
+    dispatched via Celery, because no PG consumer exists yet. The
+    ``QueueBackend.PG`` branch below is the seam where the real PG
+    enqueue lands in a later phase — keeping the ``send_task`` call
+    outside it guarantees today's wire is byte-identical regardless of
+    the routing decision.
     """
+    if select_backend(task_name) is QueueBackend.PG:
+        logger.debug(
+            "PG-queue routing selected for task=%r; dispatching via "
+            "Celery (scaffold — no PG consumer yet)",
+            task_name,
+        )
+
     headers = fairness.as_header() if fairness is not None else None
     return current_app.send_task(
         task_name,
--- a/workers/queue_backend/pg_queue/__init__.py
+++ b/workers/queue_backend/pg_queue/__init__.py
@@ -5,20 +5,20 @@ Reserved home for the PostgreSQL-backed queue substrate (PGMQ +
 migration. PGMQ is a *core* worker transport — the intended primary
 backend — so it lives inside the queue-backend seam next to its sibling
 substrates (``dispatch``, ``routing``, ``barrier``, ``redis_barrier``),
-**not** under ``workers/plugins/`` (that directory is the git-ignored
-overlay for cloud/enterprise plugins copied in at build time).
+**not** under ``workers/plugins/``, whose plugin *implementation
+subdirectories* are the git-ignored overlay copied in at build time (the
+directory itself — ``__init__.py``, ``plugin_manager.py`` — is tracked).
 
-A subpackage (rather than a single module like the barriers) because
-the real implementation spans several files — ``config.py`` (per-task
-timeouts/retries), ``consumer.py`` (poll loop + graceful shutdown), and
-the single-orchestrator (admit / reap / route, fair-admission query).
+A subpackage (rather than a single module like the barriers) because the
+real implementation will likely span several modules (config, consumer
+poll loop, orchestrator) — exact layout TBD.
 
 Empty by design in this phase. Routing decisions are made by
 :func:`queue_backend.routing.select_backend`; until a consumer exists
 here, PG-selected dispatches still ride Celery (see
 ``queue_backend.dispatch``).
 
-Labs reference: ``Zipstack/labs:labs-ali/workflow-execution-architecture``
-— ``docs/pg-queue-implementation-guide.md`` (§3 Worker Lifecycle,
-§6 Orchestrator).
+Design reference: the PG Queue implementation guide in the labs repo
+(``workflow-execution-architecture``). Branch and section pointers move,
+so they're tracked on UN-3534 / the PR rather than baked in here.
 """
--- a/workers/queue_backend/pg_queue/__init__.py
+++ b/workers/queue_backend/pg_queue/__init__.py
@@ -1,0 +1,24 @@
+"""PG Queue (PGMQ) transport substrate — scaffold.
+
+Reserved home for the PostgreSQL-backed queue substrate (PGMQ +
+``SKIP LOCKED``) that will run alongside Celery during the Strangler-Fig
+migration. PGMQ is a *core* worker transport — the intended primary
+backend — so it lives inside the queue-backend seam next to its sibling
+substrates (``dispatch``, ``routing``, ``barrier``, ``redis_barrier``),
+**not** under ``workers/plugins/`` (that directory is the git-ignored
+overlay for cloud/enterprise plugins copied in at build time).
+
+A subpackage (rather than a single module like the barriers) because
+the real implementation spans several files — ``config.py`` (per-task
+timeouts/retries), ``consumer.py`` (poll loop + graceful shutdown), and
+the single-orchestrator (admit / reap / route, fair-admission query).
+
+Empty by design in this phase. Routing decisions are made by
+:func:`queue_backend.routing.select_backend`; until a consumer exists
+here, PG-selected dispatches still ride Celery (see
+``queue_backend.dispatch``).
+
+Labs reference: ``Zipstack/labs:labs-ali/workflow-execution-architecture``
+— ``docs/pg-queue-implementation-guide.md`` (§3 Worker Lifecycle,
+§6 Orchestrator).
+"""
--- a/workers/queue_backend/routing.py
+++ b/workers/queue_backend/routing.py
@@ -1,0 +1,78 @@
+"""Queue-transport routing gate (Strangler-Fig).
+
+Decides whether a given dispatch should ride the **Celery** transport
+(today's only real path) or the future **PG Queue** (PGMQ) transport,
+based on a per-task-type opt-in allow-list read from env:
+
+- ``WORKER_PG_QUEUE_ENABLED_TASKS`` — comma-separated task names routed
+  to PG. Empty / unset → everything routes to Celery.
+
+**Why an allow-list, not a global boolean.** A single
+``WORKER_QUEUE_BACKEND=pg`` switch would move *all* traffic at once —
+the big-bang migration the PG Queue rollout explicitly forbids. The
+allow-list lets an operator migrate one task type at a time
+(drain-and-cutover) and roll back instantly by removing an entry.
+
+(Per-org granularity is intentionally out of scope here. When tenant-level
+canarying is needed it slots in behind :func:`select_backend` as a small
+additive change — no call site touches the routing decision directly.)
+
+**Scaffold posture.** This module only makes the routing *decision*.
+In the current phase there is no PG consumer, so ``dispatch()`` still
+sends PG-selected tasks via Celery (the decision is observable in logs
+but inert). ``select_backend()`` is the seam where the real PG dispatch
+lands in a later phase.
+
+**Fail-safe parsing.** Unlike ``get_barrier()`` (which raises on an
+unrecognised value, because a typo'd substrate must not silently fall
+back), the routing table is a membership set with no "invalid value"
+concept: blanks and stray whitespace are simply dropped. Malformed or
+empty config can only ever resolve to the safe ``CELERY`` default —
+``select_backend()`` never raises.
+"""
+
+from __future__ import annotations
+
+import os
+from enum import StrEnum
+
+# Env var name kept as a module literal (mirrors how the sibling
+# ``WORKER_BARRIER_BACKEND`` flag is read in ``queue_backend.__init__``)
+# so the queue-backend seam stays self-contained.
+_ENABLED_TASKS_ENV_VAR = "WORKER_PG_QUEUE_ENABLED_TASKS"
+
+
+class QueueBackend(StrEnum):
+    """Transport a dispatch is routed to.
+
+    ``StrEnum`` (3.11+) so ``QueueBackend.CELERY == "celery"`` holds at
+    runtime — members stand in anywhere a string is expected.
+    """
+
+    CELERY = "celery"
+    PG = "pg"
+
+
+def _parse_allow_list(env_var: str) -> frozenset[str]:
+    """Parse a comma-separated env allow-list into a set.
+
+    Trims surrounding whitespace per entry and drops blanks, so
+    ``"a, b ,, c"`` → ``{"a", "b", "c"}`` and ``""`` / unset → ``frozenset()``.
+    Read at call time (not import) so test harnesses that
+    ``monkeypatch.setenv`` flip the table per-test without a reload.
+    """
+    raw = os.getenv(env_var, "")
+    return frozenset(entry.strip() for entry in raw.split(",") if entry.strip())
+
+
+def select_backend(task_name: str) -> QueueBackend:
+    """Return the transport a dispatch should ride.
+
+    ``PG`` if ``task_name`` is in ``WORKER_PG_QUEUE_ENABLED_TASKS``;
+    otherwise ``CELERY``. Empty / unset allow-list → always ``CELERY``.
+
+    Never raises — the worst case is the safe ``CELERY`` default.
+    """
+    if task_name in _parse_allow_list(_ENABLED_TASKS_ENV_VAR):
+        return QueueBackend.PG
+    return QueueBackend.CELERY
--- a/workers/queue_backend/routing.py
+++ b/workers/queue_backend/routing.py
@@ -17,6 +17,22 @@ allow-list lets an operator migrate one task type at a time
 canarying is needed it slots in behind :func:`select_backend` as a small
 additive change — no call site touches the routing decision directly.)
 
+**Migration coherence — what the allow-list may and may not split.**
+Per-task routing is the right granularity for *independent / leaf* tasks
+(e.g. ``process_log_history``, ``send_webhook_notification``): no chain,
+no barrier, so one can ride PG while the rest stay on Celery — these are
+the safe first migration candidates. It is **not** valid to split the
+*coupled execution pipeline* (``async_execute_bin`` → file processing →
+callback, with the barrier fan-in) across substrates: a chain hands off
+stage→stage and the barrier coordinates a batch, so a single execution
+must run **entirely on one transport**. The migration unit for the
+pipeline is therefore the *execution*, not the task — the next phase will
+resolve the transport once at kickoff and carry it in ``ExecutionContext``
+(like the fairness key), and downstream dispatches will honour that
+carried choice over the per-task env. ``select_backend`` then becomes:
+"if the dispatch carries an execution-transport marker, use it; otherwise
+consult this allow-list." Until then, only enable *leaf* tasks here.
+
 **Scaffold posture.** This module only makes the routing *decision*.
 In the current phase there is no PG consumer, so ``dispatch()`` still
 sends PG-selected tasks via Celery (the decision is observable in logs
--- a/workers/queue_backend/routing.py
+++ b/workers/queue_backend/routing.py
@@ -23,6 +23,13 @@ sends PG-selected tasks via Celery (the decision is observable in logs
 but inert). ``select_backend()`` is the seam where the real PG dispatch
 lands in a later phase.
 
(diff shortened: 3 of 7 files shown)


muhammad-ali-e (commented):
## Automated PR Review — PG Queue Routing Scaffold (Phase 8a)

Reviewed with a 6-agent toolkit (code review, silent-failure, type design, tests, comments, simplification). This is a small, well-scoped, correct scaffold — tests pass and the byte-identical-wire invariant is real. No blocking bugs.

**Themes across findings:** (1) one stale `_ORGS` reference to an unbuilt flag; (2) the scaffold advertises itself as "observable in logs" but the only signal is a `DEBUG` log that is off by default, and a typo'd task name routes to Celery silently; (3) the headline dispatch test is tautological — the one new behaviour (the log line) is never asserted. Inline comments below, prioritised.

workers/tests/test_queue_backend_seam.py
```diff
@@ -280,17 +280,22 @@ def test_all_exports(self):
         # (registered as a Celery task on import) + the BarrierBackend
         # enum + the get_barrier factory that the WORKER_BARRIER_BACKEND
         # env flag drives.
+        # Phase 8a adds QueueBackend + select_backend — the queue-transport
+        # routing gate that the WORKER_PG_QUEUE_ENABLED_TASKS / _ORGS
```
muhammad-ali-e: **[Must-fix · consensus]** Stale `_ORGS` reference. `WORKER_PG_QUEUE_ENABLED_ORGS` does not exist anywhere in this PR — `grep` finds `_ORGS` only on this line. It also directly contradicts `routing.py:16-18`, which states per-org granularity is *"intentionally out of scope here."* A future reader will hunt for a flag that was never built.

```suggestion
        # routing gate that the WORKER_PG_QUEUE_ENABLED_TASKS allow-list drives.
```
(drop the `/ _ORGS` fragment and the now-singular wording on the next line).

workers/queue_backend/dispatch.py
```diff
@@ -28,7 +32,23 @@ def dispatch(
...
+    routing table (``WORKER_PG_QUEUE_ENABLED_TASKS``). In this phase the
+    table is a scaffold: PG-selected tasks are *logged* but still
+    dispatched via Celery, because no PG consumer exists yet. The
+    ``QueueBackend.PG`` branch below is the seam where the real PG
+    enqueue lands in a later phase — keeping the ``send_task`` call
+    outside it guarantees today's wire is byte-identical regardless of
+    the routing decision.
     """
+    if select_backend(task_name) is QueueBackend.PG:
+        logger.debug(
```
muhammad-ali-e: **[Important · observability]** The whole scaffold sells itself as *"observable in logs but inert"* (here, `routing.py`, `__init__.py`, `sample.env`), but this `logger.debug` is the *only* signal — and the worker root logger defaults to `WARNING`, so under default config this line never reaches any sink. An operator who sets the flag "to test routing locally" (per `sample.env`) sees nothing and can't tell whether the flag took effect, the entry is a typo, or the feature is broken.

A cutover decision during a canary is operationally significant — suggest `logger.info` (volume is bounded by the allow-list, not all traffic). If high-throughput volume is a worry, log-once per task name via a module-level seen-set rather than demoting to an invisible level.

workers/queue_backend/routing.py
```diff
@@ -0,0 +1,78 @@
...
+
+def select_backend(task_name: str) -> QueueBackend:
+    """Return the transport a dispatch should ride.
+
+    ``PG`` if ``task_name`` is in ``WORKER_PG_QUEUE_ENABLED_TASKS``;
+    otherwise ``CELERY``. Empty / unset allow-list → always ``CELERY``.
+
+    Never raises — the worst case is the safe ``CELERY`` default.
+    """
+    if task_name in _parse_allow_list(_ENABLED_TASKS_ENV_VAR):
```
muhammad-ali-e: **[Important · silent misconfiguration]** The docstring contrasts this favourably against `get_barrier()` ("a typo'd substrate must not silently fall back"), but the same failure mode is wide open here: `WORKER_PG_QUEUE_ENABLED_TASKS=async_execute_bni` (transposed `bin`) resolves every real task to `CELERY`, the canary silently never engages, and nothing is logged. Not-raising is defensible for a routing gate, but "fail-safe" should still be *loud*.

Minimum: log the parsed allow-list once at startup (`logger.info("PG-queue routing allow-list: %s", sorted(allow))`) so a typo is eyeballable at boot. Better, once a task registry is wired in, warn on allow-list entries that never match a real dispatched task.

workers/tests/test_routing.py
```diff
@@ -0,0 +1,135 @@
...
+            dispatch(
+                task_name,
+                args=["a", 1],
+                kwargs={"k": "v"},
+                queue="general",
+                fairness=fairness,
+            )
+        return mock_app.send_task.call_args
+
+    def test_pg_selection_does_not_change_the_wire(self, monkeypatch):
```
muhammad-ali-e, line 171: **[Important · test gap]** This is the headline scaffold test, but it's tautological: `celery_call == pg_call` is true *by construction* because `send_task` sits outside the routing branch. Verified empirically — **deleting the entire routing gate (the `select_backend` call + the `logger.debug`) leaves all tests green.** Line coverage reads 100% only because the branch runs; its sole observable effect (the log) is never asserted, so a future "the gate is inert, delete it" refactor would silently remove the seam this PR exists to establish.

Add a test that pins the branch's presence, e.g.:
```python
def test_pg_selection_emits_routing_log(self, monkeypatch, caplog):
    monkeypatch.setenv(ENABLED_TASKS_ENV, "t1")
    with patch("queue_backend.dispatch.current_app"), \
         caplog.at_level(logging.DEBUG, logger="queue_backend.dispatch"):
        dispatch("t1")
    assert any("PG-queue routing selected" in r.message for r in caplog.records)
```
A negative case (empty table → no log) guards against the gate being inverted/made unconditional.

workers/queue_backend/routing.py
```diff
@@ -0,0 +1,78 @@
...
+import os
+from enum import StrEnum
+
+# Env var name kept as a module literal (mirrors how the sibling
+# ``WORKER_BARRIER_BACKEND`` flag is read in ``queue_backend.__init__``)
+# so the queue-backend seam stays self-contained.
+_ENABLED_TASKS_ENV_VAR = "WORKER_PG_QUEUE_ENABLED_TASKS"
+
+
+class QueueBackend(StrEnum):
```
muhammad-ali-e, line 75: **[Suggestion · type design]** `StrEnum` is inherited for symmetry with `BarrierBackend`, but the justification doesn't carry over: `BarrierBackend` earns `StrEnum` because it's compared against `os.getenv(...)`; `QueueBackend` is **never** read from or written to env (only the task-name allow-list is). All call sites already use identity (`is QueueBackend.PG`), which is the correct discipline — but `StrEnum` discards the usual enum typo-safety net (`== "cellery"` is a quiet `False`, not an error) for no benefit here. A plain `Enum` would lose nothing today. If you keep `StrEnum` for sibling symmetry, a one-line docstring note that callers must use `is`, not `==`, would lock in the intent.

workers/queue_backend/routing.py
```diff
@@ -0,0 +1,78 @@
...
+def _parse_allow_list(env_var: str) -> frozenset[str]:
+    """Parse a comma-separated env allow-list into a set.
+
+    Trims surrounding whitespace per entry and drops blanks, so
+    ``"a, b ,, c"`` → ``{"a", "b", "c"}`` and ``""`` / unset → ``frozenset()``.
+    Read at call time (not import) so test harnesses that
+    ``monkeypatch.setenv`` flip the table per-test without a reload.
+    """
+    raw = os.getenv(env_var, "")
+    return frozenset(entry.strip() for entry in raw.split(",") if entry.strip())
```
muhammad-ali-e: **[Nit · simplification]** `entry.strip()` is evaluated twice per entry (filter + result). One generator pass reads cleaner and strips once:
```python
stripped = (entry.strip() for entry in raw.split(","))
return frozenset(entry for entry in stripped if entry)
```
Behaviour is identical. Micro-nit — fine to leave as the common idiom.

workers/queue_backend/routing.py
```diff
@@ -0,0 +1,78 @@
...
+
+    ``StrEnum`` (3.11+) so ``QueueBackend.CELERY == "celery"`` holds at
+    runtime — members stand in anywhere a string is expected.
+    """
+
+    CELERY = "celery"
+    PG = "pg"
+
+
+def _parse_allow_list(env_var: str) -> frozenset[str]:
```
muhammad-ali-e: **[Nit · dead generality]** `_parse_allow_list(env_var)` is only ever called with the module constant `_ENABLED_TASKS_ENV_VAR` (line 76), and `test_routing.py` re-declares the literal `"WORKER_PG_QUEUE_ENABLED_TASKS"` rather than importing it — so the canonical env-var name now lives in two places and can desync silently. Either drop the parameter and read the constant directly, or have the test import `_ENABLED_TASKS_ENV_VAR` for a single source of truth.

workers/queue_backend/pg_queue/__init__.py
```diff
@@ -0,0 +1,24 @@
+"""PG Queue (PGMQ) transport substrate — scaffold.
+
+Reserved home for the PostgreSQL-backed queue substrate (PGMQ +
+``SKIP LOCKED``) that will run alongside Celery during the Strangler-Fig
+migration. PGMQ is a *core* worker transport — the intended primary
+backend — so it lives inside the queue-backend seam next to its sibling
+substrates (``dispatch``, ``routing``, ``barrier``, ``redis_barrier``),
+**not** under ``workers/plugins/`` (that directory is the git-ignored
```
muhammad-ali-e: **[Nit · comment accuracy]** "that directory is the git-ignored overlay" is imprecise — `workers/plugins/` itself is **tracked** (`__init__.py`, `plugin_manager.py`, `.gitignore` are all in git); only its plugin *implementation subdirectories* are ignored. Given this PR's premise is comment accuracy, suggest: "...not under `workers/plugins/`, whose plugin *implementation subdirectories* are the git-ignored overlay copied in at build time."

workers/queue_backend/pg_queue/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+the real implementation spans several files — ``config.py`` (per-task
+timeouts/retries), ``consumer.py`` (poll loop + graceful shutdown), and
+the single-orchestrator (admit / reap / route, fair-admission query).
+
+Empty by design in this phase. Routing decisions are made by
+:func:`queue_backend.routing.select_backend`; until a consumer exists
+here, PG-selected dispatches still ride Celery (see
+``queue_backend.dispatch``).
+
+Labs reference: ``Zipstack/labs:labs-ali/workflow-execution-architecture``
```
muhammad-ali-e: **[Nit · comment-rot risk]** This pointer hard-codes a personal feature-branch path in a separate repo (`labs-ali/workflow-execution-architecture`) plus volatile section numbers (§3, §6). Branch-scoped cross-repo references rot fast (merge/rename/squash) and can't be reference-checked by CI here. Prefer a stable/merged path or a commit-pinned permalink, drop the §-numbers, or move this pointer to the PR description / UN-3534 rather than baking it into source. (Same applies to the `config.py`/`consumer.py` filenames on line 12 — fine as a sketch, but soften to "will likely span several modules" so they don't read as a contract.)

workers/tests/test_queue_backend_seam.py
muhammad-ali-e: Fixed in `3739a2933` — dropped the `/ _ORGS` fragment; the comment now reads singular: *"the queue-transport routing gate that the WORKER_PG_QUEUE_ENABLED_TASKS allow-list drives."* No `_ORGS` reference remains anywhere in the PR.

workers/queue_backend/dispatch.py
muhammad-ali-e: Done in `3739a2933`. Promoted to `logger.info` so a cutover is visible under a default log config (it's a config-significant event, not a problem, so INFO not WARNING). Bounded to **log-once per task name** per process via a module-level `_pg_routing_logged` set, so an opted-in high-throughput task (e.g. `async_execute_bin`) doesn't log on every dispatch. Now pinned by a caplog test.

workers/queue_backend/routing.py
muhammad-ali-e: Done in `3739a2933`. Added `_log_allow_list_once()` — logs the parsed allow-list at INFO, once per process, so a typo'd name (`async_execute_bni`) is eyeballable even though it never matches a real task. Only fires for a **non-empty** allow-list, so the default (feature off) stays silent and the scaffold remains truly inert when unused. Logged lazily on the first `select_backend()` call rather than at import (after logging is configured). A registry-based "entry never matched a real task" warning is a good follow-up once a task registry is wired in — noted for the rollout phase.

workers/tests/test_routing.py
muhammad-ali-e: Great catch — reproduced it (deleting the gate left everything green). Fixed in `3739a2933` with three caplog tests in `TestObservability`:
- `test_pg_selection_emits_routing_log` — asserts the routing line fires when PG-selected (deleting the `select_backend` call or the log now fails here),
- `test_celery_selection_emits_no_routing_log` — negative case, empty table → no line (guards against the gate being made unconditional/inverted),
- `test_routing_log_bounded_to_once_per_task` — pins the log-once behaviour.

The branch now has an asserted observable effect, not just incidental coverage.

workers/queue_backend/routing.py
muhammad-ali-e: Kept `StrEnum` for sibling symmetry with `BarrierBackend` and added the docstring note you suggested: callers MUST compare by identity (`backend is QueueBackend.PG`), never `== "pg"`, because `StrEnum` makes a typo'd `== "cellery"` a silent `False`. All call sites already use `is`. `3739a2933`.

workers/queue_backend/routing.py
muhammad-ali-e: Applied in `3739a2933` — single generator pass, strips once:
```python
stripped = (entry.strip() for entry in os.getenv(...).split(","))
return frozenset(entry for entry in stripped if entry)
```

workers/queue_backend/routing.py
muhammad-ali-e: Both addressed in `3739a2933`. Dropped the `env_var` parameter — `_parse_allow_list()` now reads `_ENABLED_TASKS_ENV_VAR` directly (it was the only caller, so the generality was dead). And the test imports `_ENABLED_TASKS_ENV_VAR` instead of re-declaring the literal → single source of truth, can't desync.

workers/queue_backend/pg_queue/__init__.py
muhammad-ali-e: Fixed in `3739a2933` — reworded to: *"...not under `workers/plugins/`, whose plugin **implementation subdirectories** are the git-ignored overlay copied in at build time (the directory itself — `__init__.py`, `plugin_manager.py` — is tracked)."*

workers/queue_backend/pg_queue/__init__.py
muhammad-ali-e: Softened in `3739a2933`. Dropped the volatile §-numbers and the `config.py`/`consumer.py` filenames (now "will likely span several modules (config, consumer, orchestrator) — exact layout TBD"), and moved the branch/section pointer out of source: *"Branch and section pointers move, so they're tracked on UN-3534 / the PR rather than baked in here."*

muhammad-ali-e:
### Review feedback addressed — `3739a2933`

All 9 threads resolved (1 must-fix, 3 important, 1 suggestion, 4 nits):

**Must-fix**
- Dropped the stale `WORKER_PG_QUEUE_ENABLED_ORGS` reference in the seam test (org axis was removed; that flag never existed).

**Important**
- **Observability** — routing log `DEBUG → INFO` (survives a default log config), **log-once per task name** to bound volume.
- **Silent misconfig** — `_log_allow_list_once()` logs the configured allow-list at INFO once per process, so a typo'd task name is eyeballable even when it never matches a real dispatch (fires only for a non-empty list → default stays inert).
- **Test gap** — added caplog tests pinning the routing branch (PG → log fires, Celery → no log, bounded-to-once). Deleting the gate now fails a test; confirmed it was green-on-delete before.

**Suggestion**
- `QueueBackend` keeps `StrEnum` for symmetry with `BarrierBackend` + docstring note that callers must use `is`, not `==`.

**Nits**
- One-pass strip in `_parse_allow_list`; dropped its dead `env_var` param; test imports `_ENABLED_TASKS_ENV_VAR` (single source of truth).
- `pg_queue` docstring: clarified `plugins/` subdirs are git-ignored (dir itself tracked); softened volatile labs branch/section/filename refs.

50 queue_backend unit tests green (17 routing, +5 observability); ruff + pre-commit clean.

muhammad-ali-e (commented):
## LGTM ✅ (non-blocking — author can't self-approve)

All 9 findings from the prior review pass are resolved, verified against `3739a293` + `510a0c4f`:

- 🔴 Stale `_ORGS` reference removed (grep → zero occurrences anywhere).
- 🟠 Routing log promoted to INFO, bounded log-once per task, plus a startup allow-list dump (`_log_allow_list_once`) so a typo'd task name is eyeballable at boot — the canary is now genuinely observable, not just "observable in theory."
- 🟠 The tautological dispatch test is fixed: `TestObservability` pins the routing branch with positive, negative, and log-once cases, so deleting the gate now fails a test (I reproduced the original "delete gate → all green" hole; it's closed).
- 🟡 StrEnum identity-comparison discipline documented and honoured at every call site; double-strip, dead `env_var` param, and both pg_queue docstring nits all addressed.

The self-added migration-coherence note (`510a0c4f`) — leaf tasks are safe to split, but the coupled `async_execute_bin` → file-processing → callback pipeline must migrate per-*execution* via ExecutionContext rather than per-task — is a valuable forward-looking constraint that pre-empts a real footgun in the next phase.

**34/34 worker routing + seam tests pass locally.** Scaffold is correct, inert (byte-identical wire), and well-tested. No blocking concerns.

One non-blocking follow-up, already parked for the rollout phase: a registry-based warning for allow-list entries that never match a real dispatched task. The current startup allow-list dump is a solid interim.

Needs a teammate's formal approval to merge (GitHub blocks self-approval).

workers/queue_backend/dispatch.py
```diff
@@ -28,7 +37,29 @@ def dispatch(
...
+        and task_name not in _pg_routing_logged
+    ):
+        # INFO (not DEBUG) so the cutover is visible under a default log
+        # config; log-once per task name keeps the volume bounded.
+        _pg_routing_logged.add(task_name)
+        logger.info(
+            "PG-queue routing selected for task=%r; dispatching via "
+            "Celery (scaffold — no PG consumer yet)",
+            task_name,
+        )
```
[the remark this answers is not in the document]
muhammad-ali-e: Declining for the scaffold. The env parse (`os.getenv` + `split` + `frozenset` on a short string) is on the order of nanoseconds and is immediately followed by `current_app.send_task()` — broker serialisation + a network round-trip (~ms), which dominates by orders of magnitude. So "scales linearly with dispatch frequency" is true but the constant is negligible next to the work it precedes.

The proposed short-circuit also isn't safe as written: `_pg_routing_logged` is a *"has this been logged"* set, not a routing-decision cache. Returning `QueueBackend.PG` from it would (a) conflate logging with routing, and (b) defeat the call-time env read that tests rely on and that gives runtime rollback its meaning — a task removed from the allow-list on restart would still resolve to PG from the stale set.

If profiling ever shows this hot, the right fix is a parse-once module cache with an explicit test-reset hook, not reusing the log set — but that's unwarranted complexity while the gate is inert. Noted the preference against per-call env reads for the rollout phase.

workers/queue_backend/routing.py
```diff
@@ -0,0 +1,134 @@
...
+logger = logging.getLogger(__name__)
+
+# Env var name kept as a module literal (mirrors how the sibling
+# ``WORKER_BARRIER_BACKEND`` flag is read in ``queue_backend.__init__``)
+# so the queue-backend seam stays self-contained.
+_ENABLED_TASKS_ENV_VAR = "WORKER_PG_QUEUE_ENABLED_TASKS"
+
+# One-shot guard so the configured allow-list is logged once per process
+# (see ``_log_allow_list_once``). Module-level → per prefork child.
+_allow_list_logged = False
```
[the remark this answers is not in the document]
muhammad-ali-e: Declining for now, intentionally. The workers run `--pool=prefork` (separate processes, each with its own module-level state), so there's no shared-memory race in the actual deployment. Even if a threaded pool (`-P threads`/`gevent`) were adopted, the worst case here is a *duplicate INFO line* — no data corruption (as you note re the GIL) and no functional effect, since the flag only de-dups a cosmetic log. A `threading.Lock` (or `threading.local`) on a once-per-process log guard is disproportionate for an inert scaffold. If we move to a threaded pool later, the failure mode is a benign double log and a lock can be added then.

workers/sample.env
```diff
@@ -131,6 +131,29 @@ WORKER_BARRIER_BACKEND=chord
...
+# Scaffold note: in this release there is no PG consumer yet, so
+# PG-selected tasks are logged but still dispatched via Celery. Setting
+# this flag is therefore safe to test routing decisions locally with
+# zero behaviour change. No Flipt server required.
+#
+# Roll back instantly by removing an entry; flips take effect on worker
+# restart (same posture as every other WORKER_* env). The value must be
+# a task dispatched via queue_backend.dispatch() — e.g. async_execute_bin
+# (scheduler) or send_webhook_notification (notifications).
+# WORKER_PG_QUEUE_ENABLED_TASKS=async_execute_bin
```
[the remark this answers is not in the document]
muhammad-ali-e: Applied in `249036406` — you're right, and it directly contradicts the leaf-only coherence note added in the previous commit. `async_execute_bin` is the pipeline kickoff (→ file processing → callback + barrier fan-in), exactly the task that must not be split per-task. Switched the example to `send_webhook_notification` (the leaf validated in dev testing) and added the warning against listing coupled pipeline tasks until `ExecutionContext` carries the transport choice, pointing at the `routing.py` docstring for the full constraint.

workers/queue_backend/dispatch.py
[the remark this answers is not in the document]
muhammad-ali-e: Yes — please remember it. Scoped precisely as you put it: don't flag `os.getenv` + string-parse overhead at a call-site whose dominant cost is a broker serialisation + network round-trip, and treat a call-time flag/allow-list read as intentional (it's load-bearing for runtime rollback) rather than something to cache away. Appreciated the clean concession.

PR #2034, opened 2026-06-11
UN-2265 [FEAT] Show the running platform version on the profile page

athul-rs:
## What

- The profile page's Organization card now shows a **Platform Version** field (e.g. `v0.144.0`) on all deployment targets — OSS docker, cloud, on-prem.

## Why

[UN-2265](https://zipstack.atlassian.net/browse/UN-2265) — most apps surface their running version; users reporting issues currently have no way to tell us what version they're on, which slows down troubleshooting.

## How

The version is supplied at **runtime** (never baked into the image) and flows env → runtime config → UI with no new endpoints:

1. **`UNSTRACT_APPS_VERSION` is provided by the deployment:**
   - **OSS docker-compose** (`docker/docker-compose.yaml`): `UNSTRACT_APPS_VERSION=${VERSION}` on the frontend service — the *same* `${VERSION}` that already tags every image, so the shown version always matches what's running.
   - **Enterprise / On-prem (Helm)**: injected from the chart version anchor — see companion PR Zipstack/unstract-cloud#1621.
2. `generate-runtime-config.sh`: reads `UNSTRACT_APPS_VERSION` (js-escaped) and injects `version` into `window.RUNTIME_CONFIG` at container start (existing mechanism).
3. `config.js`: exposes `config.version` (runtime config first, `VITE_VERSION` env fallback for dev).
4. `Profile.jsx`: renders the field only when a version is available, using the existing field-group styling.

**Why runtime env and not a build-time bake (revised after review):** RC→stable promotion only **re-tags** images — it does not rebuild them — so a version baked into the image at build time would freeze at the `rc` tag and be shown in Production. Sourcing from the runtime env (compose for OSS, Helm anchor for Enterprise) keeps the displayed version aligned with the promoted release. (per @ritwik-g / @ejhari)

- No. Additive only: one extra key in the frontend runtime config and a conditionally rendered field. When `UNSTRACT_APPS_VERSION` is unset (older images, local dev server, or a deployment that doesn't set it), `config.version` is empty and the field simply doesn't render. No image now bakes a version, so nothing is removed that anything else depended on.

## Database Migrations

- None

## Env Config

- `UNSTRACT_APPS_VERSION` (frontend container) — supplied at runtime: docker-compose passes `${VERSION}` in OSS; Helm injects the chart version anchor in Enterprise/On-prem (Zipstack/unstract-cloud#1621). No longer a build arg.

## Relevant Docs

- N/A

## Related Issues or PRs

- Companion (required pair for cloud/on-prem): Zipstack/unstract-cloud#1621 — injects `UNSTRACT_APPS_VERSION` into the frontend container from the Helm chart.
- Jira: UN-2265

## Notes on Testing

- Biome lint clean; vite build succeeds; `sh -n` passes on the runtime-config script.
- OSS: `VERSION=v9.9.9 docker compose config` renders `UNSTRACT_APPS_VERSION: v9.9.9` on the frontend service; unset `VERSION` → empty → field hidden.
- Manual: `UNSTRACT_APPS_VERSION=v1.2.3` on the frontend container → profile page shows `Platform Version: v1.2.3`; unset → field hidden.

## Screenshots

N/A

## Checklist

[UN-2265]: https://zipstack.atlassian.net/browse/UN-2265


docker/dockerfiles/frontend.Dockerfile
```diff
@@ -72,6 +72,10 @@ RUN sed -i 's|</head>|    <script src="/config/runtime-config.js"></script>\n  <
 COPY frontend/generate-runtime-config.sh /docker-entrypoint.d/40-env.sh
 RUN chmod +x /docker-entrypoint.d/40-env.sh
 
+# Capture build version at the very end so it doesn't affect layer caching
+ARG VERSION=dev
+ENV UNSTRACT_APPS_VERSION=${VERSION}
```
chandrasekharan-zipstack: @athul-rs does this get set during cloud builds as well?

chandrasekharan-zipstack (approved):

jaseemjaskp (changes requested):
@Athul From my point of view, the platform version is not particularly relevant to display in OSS or other environments.

For OSS, users already know which version they have deployed locally. For on-prem deployments, we maintain a version tracker and customers know exactly which version they are running. In cloud deployments, customers are always on the latest version, and we already know what version is currently deployed.

Because of that, I don’t think customers need to be aware of the platform version. I also don’t recall many enterprise applications exposing the platform version prominently in the UI.

chandrasekharan-zipstack:
@jaseemjaskp I think its useful to display the version for several reasons and we don't lose anything on attempting to do so. Its more of a convenience here
- becomes easier to debug issues across environments
- anyone using our dev namespace easily infers the version and can refer that while sending messages
   - faced this concern first hand when I repeatedly kept deploying different versions for testing
- OSS users also find this hard because they'll only do a `./run-platform.sh -u` (for upgrade) and they'll not immediately know what version of Unstract is being used
- I believe most applications do show the app version within the product itself, makes it easier to convey and reason about issues

Do you see any engineering challenges or concerns in displaying this?

docker/dockerfiles/frontend.Dockerfile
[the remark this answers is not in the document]
athul-rs: Fixed — `ARG VERSION` now defaults to empty, so images built without the arg hide the field instead of showing `dev`.

frontend/generate-runtime-config.sh
```diff
@@ -9,7 +9,8 @@ cat > /usr/share/nginx/html/config/runtime-config.js << EOF
 // This file is auto-generated at runtime. Do not modify manually.
 window.RUNTIME_CONFIG = {
   faviconPath: "${VITE_FAVICON_PATH:-${REACT_APP_FAVICON_PATH}}",
-  logoUrl: "${VITE_CUSTOM_LOGO_URL:-${REACT_APP_CUSTOM_LOGO_URL}}"
+  logoUrl: "${VITE_CUSTOM_LOGO_URL:-${REACT_APP_CUSTOM_LOGO_URL}}",
+  version: "${UNSTRACT_APPS_VERSION:-}"
```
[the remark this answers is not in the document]
athul-rs: Fixed — added a `js_escape` helper that escapes backslashes and double quotes before the value is embedded in `runtime-config.js`.

docker/dockerfiles/frontend.Dockerfile
athul-rs: Yes — cloud frontend builds go through docker bake with `*.args.VERSION=${{ inputs.tag }}` (`cloud-frontend-docker-build-push.yaml`), same as OSS `production-build.yaml`, so cloud images get the rc tag baked in automatically. On-prem frontend builds use the same workflow pattern.

athul-rs:
@jaseemjaskp fair point — taking this back to the ticket's intent rather than pushing my own view: UN-2265 was filed by @chandrasekharan-zipstack from a suggestion that surfacing the version helps support triage (when a user reports an issue, we can ask what version the UI shows instead of cross-referencing deploy trackers). The OSS docker-compose case is the strongest one IMO: `latest` images mean self-hosted users often genuinely don't know what they're running.

That said, this is a product call, not mine. Options as I see them:
1. Merge as-is (field only shows where a version is baked, and it sits on the profile page rather than anywhere prominent).
2. Gate it to OSS/on-prem only (trivial: don't pass the build-arg in cloud builds, the field auto-hides).
3. Close the PR and mark UN-2265 as won't-do.

Happy with whichever you and Chandru align on — the change is additive and cheap to drop or gate.

athul-rs:
@hari-kuriakose What should be the approach here ? Let's update the PR and ticket in jira accordingly.

docker/dockerfiles/frontend.Dockerfile
ritwik-g: @athul-rs doesn't it mean we will show rc tags in the UI?

ejhari: @ritwik-g It is relatively ok to show `rc` tags in UI, as long as they appear internally only.

@athul-rs That being said, since we re-tag an existing `rc` tag to promote it to Production, the version input for frontend display has to come from the Helm chart dynamically, instead of being baked into the image. Is this the case now? Asking since an `rc` tag might get displayed instead in Cloud/On-prem Production as mentioned.

ejhari:
@athul-rs We need to show the release version in UI, dynamically pulled from the Docker compose file in OSS and Helm chart in Enterprise.

Both OSS platform end users and Enterprise On-prem customer dev teams will not have a clue otherwise.

A release version is a vital information when filing or debugging an issue.

ritwik-g (approved):

jaseemjaskp (approved):

PR #2036, opened 2026-06-11
UN-3537 [FEAT] PG Queue Phase 9a — extension-free queue schema + SKIP LOCKED dequeue

muhammad-ali-e:
## What

The first piece of the PG Queue **engine** (Phase 9) that the Phase 8 routing seam routes to: an **extension-free** queue storage + dequeue primitive. **Inert** — nothing in `dispatch()` calls it yet (the routing gate's PG branch still routes to Celery), so zero behaviour change even on the integration branch.

Targets `feat/UN-3445-pg-queue-integration` (not `main`).

## Foundation constraints honoured

- **No binary extension** — no `CREATE EXTENSION`, no `pg_cron`. Plain table + index + a single `FOR UPDATE SKIP LOCKED` statement. Works on every managed/​on-prem Postgres with normal DDL privileges.
- **Reuses the backend's DB connection** (`DB_*` env) — so it routes through the existing PgBouncer in cloud and connects directly in OSS, no new infra.
- **Schema is 100% `makemigrations`-generated** — the table + dequeue index come from a standard Django model; there is no hand-written migration and no DB-side function. All queue logic lives in one place: the client.

## Changes

**Backend (`pg_queue` app — owns the schema):**
- `models.py` — `managed=True` `PgQueueMessage` (typed read handle + the source of the migration).
- `migrations/0001_initial.py` — generated `CreateModel` + dequeue index `(queue_name, vt, msg_id)`.
- Registered in `SHARED_APPS` (cross-org infrastructure → shared schema, not per-tenant).

**Workers (`queue_backend/pg_queue/` — the client; first direct-DB worker capability):**
- `client.py` — `PgQueueClient.send / read / delete` + `QueueMessage`. `read()` runs one atomic `UPDATE … WHERE msg_id IN (SELECT … FOR UPDATE SKIP LOCKED …) RETURNING …` (visibility-timeout pattern): claim + commit, process *outside* the transaction, `delete()` on success. A crash lets `vt` expire and the row redelivers — at-least-once, no double-delivery, VACUUM-safe, PgBouncer transaction-pooling compatible.
- `connection.py` — `create_pg_connection()` reusing the backend `DB_*` env.
- `pyproject.toml` + `uv.lock` — `psycopg2-binary==2.9.9` (matches backend) promoted to a direct dependency.

## Testing

- **4 unit** (mocked connection) — `send` / `read` / `delete` issue the right SQL + params and commit.
- **4 integration** (real Postgres) — roundtrip, vt-hiding (second read sees nothing), vt-expiry redelivery, and no-double-delivery across two readers. Skip gracefully if Postgres is unreachable or the migration isn't applied.
- `makemigrations --check` → no drift; ruff clean (backend + workers).
- Dev-tested against a running dev DB: `migrate pg_queue` applies cleanly, 8/8 tests pass.

## Out of scope (later 9 sub-tasks)

- Wiring `dispatch()` to enqueue to PG (9b); the consumer poll loop (9c); orchestrator / fairness / worker_registry (9d).

Ticket: UN-3537 (parent UN-3536, epic UN-3445)


workers/queue_backend/pg_queue/client.py
```diff
@@ -0,0 +1,119 @@
...
+        """
+        with self.conn.cursor() as cur:
+            cur.execute(
+                "INSERT INTO pg_queue_message "
+                "(queue_name, message, org_id, enqueued_at, vt, read_ct) "
+                "VALUES (%s, %s::jsonb, %s, now(), now(), 0) RETURNING msg_id",
+                (queue_name, json.dumps(message), org_id),
+            )
+            msg_id = cur.fetchone()[0]
+        self.conn.commit()
```
muhammad-ali-e: **[Critical · Silent-Failure/Code-Review] No rollback on DB error → poisoned cached connection.** `send`/`read`/`delete` each `execute(...)` then `commit()` with no `try/except`. The connection is cached and reused (`self._conn`), so if `execute`/`fetchone`/`commit` raises (constraint violation, serialization failure, dropped socket, statement timeout) the connection is left in `InFailedSqlTransaction` and **every subsequent call** fails with `current transaction is aborted` — hiding the real root cause. This is precisely the failure the 9c poll loop will hit.

Wrap each operation so a failure calls `self.conn.rollback()` (or `self._conn = None` to force reconnect) and re-raises. Also add a module logger — `logger = logging.getLogger(__name__)` — which every sibling in this package (`dispatch.py`, `barrier.py`, `routing.py`, `redis_barrier.py`) has and this module lacks.

workers/queue_backend/pg_queue/client.py
```diff
@@ -0,0 +1,119 @@
...
+    from the backend ``DB_*`` env on first use.
+    """
+
+    def __init__(self, conn: PgConnection | None = None) -> None:
+        self._conn = conn
+
+    @property
+    def conn(self) -> PgConnection:
+        if self._conn is None:
+            self._conn = create_pg_connection()
```
muhammad-ali-e, line 106: **[High · Silent-Failure] Lazy connection is never closed or recovered.** No `close()`, no `__enter__`/`__exit__`, and a connection that goes bad (PgBouncer recycle / dropped socket → `OperationalError`) is reused indefinitely. Combined with the missing rollback above, one connection blip permanently stops the 9c consumer from draining its queue (messages pile up silently after the first error). Add `close()` + context-manager support, discard the cached conn on `OperationalError` so the next call reconnects, and document ownership of injected vs self-created connections.

workers/queue_backend/pg_queue/client.py
```diff
@@ -0,0 +1,119 @@
...
+                "(queue_name, message, org_id, enqueued_at, vt, read_ct) "
+                "VALUES (%s, %s::jsonb, %s, now(), now(), 0) RETURNING msg_id",
+                (queue_name, json.dumps(message), org_id),
+            )
+            msg_id = cur.fetchone()[0]
+        self.conn.commit()
+        return int(msg_id)
+
+    def read(
+        self, queue_name: str, *, vt_seconds: int = 30, qty: int = 1
```
muhammad-ali-e, line 164: **[Medium · Silent-Failure/Type-Design] `read` does not validate `vt_seconds`/`qty`.** `vt_seconds <= 0` makes a claimed message immediately re-visible — a silent double-delivery window that directly undermines the at-least-once/no-double-delivery guarantee in the module docstring. `qty <= 0` yields a pointless or erroring `LIMIT`. Sibling `redis_barrier.py` raises on invalid config rather than degrading silently. Add guards at the top: `if vt_seconds <= 0: raise ValueError(...)` and `if qty < 1: raise ValueError(...)`.

workers/queue_backend/pg_queue/client.py
```diff
@@ -0,0 +1,119 @@
...
+        self.conn.commit()
+        return [QueueMessage(msg_id=int(r[0]), message=r[1]) for r in rows]
+
+    def delete(self, msg_id: int) -> bool:
+        """Ack a processed message. Returns ``True`` if a row was removed."""
+        with self.conn.cursor() as cur:
+            cur.execute("DELETE FROM pg_queue_message WHERE msg_id = %s", (msg_id,))
+            deleted = cur.rowcount
+        self.conn.commit()
+        return deleted == 1
```
muhammad-ali-e, line 202: **[Medium · Silent-Failure] `delete()` returning `False` swallows a correctness signal.** A `False` means the row was already gone — typically the visibility timeout expired during processing and the work was (or will be) done twice. Returning a bare bool with no log means callers almost certainly ignore the very double-processing condition this design hinges on. Log a warning when `rowcount == 0` (e.g. `pg_queue delete: msg_id=%s already gone (vt expired / double-ack?)`) in addition to returning the bool.

workers/queue_backend/pg_queue/client.py
```diff
@@ -0,0 +1,119 @@
...
+RETURNING msg_id, message
+"""
+
+
+@dataclass(frozen=True)
+class QueueMessage:
+    """A claimed queue message."""
+
+    msg_id: int
+    message: dict[str, Any]
```
muhammad-ali-e: **[Low · Type-Design] `frozen=True` only freezes the binding, not the dict.** `message: dict[str, Any]` is mutable, so a consumer can do `msg.message[...] = ...` between read and delete — a mixed immutability signal for a DTO that advertises itself as frozen. Minor for an internal type; consider `slots=True` and a docstring noting the payload is the already-decoded JSONB (psycopg2 returns a parsed dict). Not worth deep-freezing the dict.

workers/queue_backend/pg_queue/connection.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def create_pg_connection() -> PgConnection:
+    """Open a direct Postgres connection using the backend's ``DB_*`` env.
+
+    Env names and defaults mirror ``backend/settings/base.py`` so the
+    queue connects to the same database/schema the backend manages.
+    """
+    schema = os.getenv("DB_SCHEMA", "unstract")
+    return psycopg2.connect(
```
muhammad-ali-e: **[High · Silent-Failure] `psycopg2.connect` has no error handling.** On failure it raises a bare `OperationalError` (`could not connect` / `password authentication failed`) with no hint that this is the *PG queue* — the first direct-DB worker component, per this module's own docstring — or which `DB_*` var is wrong. Wrap the connect, log the non-secret host/port/dbname/schema, and re-raise so it still fails loudly but with actionable context.

workers/queue_backend/pg_queue/connection.py
```diff
@@ -0,0 +1,40 @@
...
+        host=os.getenv("DB_HOST", "unstract-db"),
```
muhammad-ali-e: **[Suggestion · Code-Review] Default `DB_HOST` doesn't 'mirror' the backend.** The function docstring says env defaults mirror `backend/settings/base.py`, but that file defaults `DB_HOST` to `backend-db-1` while this defaults to `unstract-db`. (All other defaults — `DB_NAME`/`DB_USER`/`DB_PASSWORD`/`DB_PORT`/`DB_SCHEMA` — do match.) Either align to `backend-db-1`, or soften the docstring to note the host default intentionally differs (compose service name).

workers/queue_backend/pg_queue/connection.py
```diff
@@ -0,0 +1,40 @@
...
+The PG queue is the first worker component to open a *direct* Postgres
+connection — workers otherwise reach state via the Internal API, with
+Redis the only other direct store. It reuses the **backend's** ``DB_*``
+env so the target follows the deployment: in cloud those vars point at
+PgBouncer (connection pooling); in OSS / on-prem they point straight at
+Postgres (UN-3533 decision).
+
+``search_path`` is set to ``DB_SCHEMA`` so the bespoke
+``pg_queue_message`` table and ``pg_queue_read`` function resolve in the
+same schema the backend manages.
```
muhammad-ali-e: **[High · Comment-Analyzer] References a `pg_queue_read` function that does not exist.** No `pg_queue_read` is created anywhere in this PR — the dequeue is inline `_DEQUEUE_SQL` in `client.py`. `search_path` correctly matters for the *table*, but naming a non-existent function is misleading and contradicts `client.py`'s 'no DB-side function' docstring. Reword to: "`search_path` is set to `DB_SCHEMA` so the bespoke `pg_queue_message` table resolves in the same schema the backend manages."

muhammad-ali-e (commented):
## Automated review — PG Queue Phase 9a (UN-3537)

Review run via the PR Review Toolkit (Code Reviewer, Silent-Failure Hunter, Type-Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier). Overall this is a clean, well-documented primitive — the `SKIP LOCKED` design is correct and the no-double-delivery invariant is properly delegated to Postgres. Findings below are inline.

**Highest-priority themes:**
1. **No rollback/error handling on the cached connection** — a single transient DB error permanently poisons a long-lived `PgQueueClient` (the 9c consumer's exact failure mode).
2. **Docstrings describe a migration 0002, DB-level defaults, and a `pg_queue_read` function that don't exist** in this PR — and contradict the 'extension-free, no DB-side function' design in `client.py`.
3. **Integration test fixture's broad `except`** can turn real failures into green skips.

**Out-of-band caveat (not inline, needs author confirmation):** the Test Analyzer reported that against a *reachable* DB the integration tests fail with `NotNullViolation` on `org_id` — i.e. the deployed `pg_queue_message.org_id` is NOT NULL while migration `0001_initial.py` declares `null=True`. The migration is authoritative (and correct), so this looks like a **stale local DB**, not a code bug — but please confirm the deployed schema matches the migration before merge.

backend/pg_queue/models.py
```diff
@@ -0,0 +1,36 @@
...
+class PgQueueMessage(models.Model):
+    """Bespoke extension-free queue table (``SKIP LOCKED`` + visibility timeout).
+
+    The table and its dequeue index are Django-managed (generated by
+    ``makemigrations``, migration 0001). Two things Django 4.2 cannot
+    express are added by hand in migration 0002 via ``RunSQL`` — and only
+    those: (a) DB-level ``DEFAULT now()`` / ``0`` on the columns the raw-SQL
+    producer omits (``db_default`` needs Django 5.0+), and (b) the
+    ``pg_queue_read`` claim function (no model concept for SQL functions).
+    See UN-3533 (extension-free decision).
```
muhammad-ali-e: **[High · Comment-Analyzer/Type-Design] Describes a migration 0002, DB-level defaults, and a `pg_queue_read` function that don't exist.** Only `0001_initial.py` is present — a plain `CreateModel` with no `RunSQL`, no DB-level `DEFAULT`, no SQL function (grep finds `pg_queue_read` only in these docstrings). This also contradicts the 'extension-free / no DB-side function' design asserted in `client.py`'s module docstring. Either restore migration 0002 in this PR, or reword to future tense and drop the `pg_queue_read` clause so the schema isn't documented as having behavior it lacks.

backend/pg_queue/models.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+
+    msg_id = models.BigAutoField(primary_key=True)
+    queue_name = models.TextField()
+    message = models.JSONField()
+    # Nullable: leaf tasks may carry no org; the fair-admission query
+    # (a later phase) uses it for the coupled pipeline.
+    org_id = models.TextField(null=True, blank=True)
+    # Python-level defaults so ORM ``.create()`` works without these; the
+    # raw-SQL client relies on the DB-level defaults added in migration 0002.
```
muhammad-ali-e: **[High · Comment-Analyzer] Factually wrong — the raw-SQL client does NOT rely on DB-level defaults.** `PgQueueClient.send` explicitly inserts `now(), now(), 0` for `enqueued_at`/`vt`/`read_ct` (`client.py:91`), and migration 0002 doesn't exist. Replace with: "Python-level defaults let ORM `.create()` work; the raw-SQL client supplies these columns explicitly, so no DB-level default is required."

workers/tests/test_pg_queue_client.py
```diff
@@ -0,0 +1,175 @@
...
+        password=os.getenv("TEST_DB_PASSWORD", "unstract_pass"),
+        options=f"-c search_path={os.getenv('TEST_DB_SCHEMA', 'unstract')}",
+    )
+
+
+@pytest.fixture
+def pg_conn():
+    try:
+        conn = _integration_conn()
+    except Exception as exc:  # noqa: BLE001 — any connect failure → skip
```
muhammad-ali-e: **[High · Silent-Failure/Test] Broad `except Exception` masks real failures as 'Postgres not reachable'.** This also swallows `ImportError` (psycopg2 missing), bugs in `_integration_conn`, and permission/schema errors — all reported as a skip. CI then shows green ('skipped') while the integration tests never run, defeating their purpose. Narrow to `except psycopg2.OperationalError as exc:` and let everything else propagate. (The `# noqa: BLE001` is acknowledging the smell, not justifying it.)

workers/tests/test_pg_queue_client.py
```diff
@@ -0,0 +1,175 @@
...
+    conn.rollback()
+    conn.close()
+
+
+@pytest.fixture
+def queue_name(pg_conn):
+    # Unique per test for isolation; clean up rows afterwards.
+    name = f"test_q_{os.getpid()}_{int(time.time() * 1000)}"
+    yield name
+    with pg_conn.cursor() as cur:
```
muhammad-ali-e, line 273: **[Medium · Test] `queue_name` teardown can corrupt on a failed test.** The cleanup `DELETE` runs with no prior `rollback()`. When a test body fails mid-transaction, this finalizer raises `InFailedSqlTransaction`, turning one logical failure into N errors and leaking test rows. Add `pg_conn.rollback()` before the `DELETE` (the `pg_conn` fixture's rollback runs after this finalizer, so it doesn't help here).

workers/tests/test_pg_queue_client.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def _integration_conn():
+    # Target the dev DB via dedicated TEST_DB_* env (dev defaults). We
+    # deliberately do NOT read the generic DB_* vars: the test suite's
+    # conftest sets DB_USER=test etc. for unit isolation, which would
+    # point this real-DB connection at nonexistent credentials.
+    import psycopg2
+
+    return psycopg2.connect(
```
muhammad-ali-e: **[Suggestion · Simplifier/Test] Duplicates `create_pg_connection`, and `connection.py` has zero coverage.** This connect body mirrors `create_pg_connection` (only the `TEST_DB_*` prefix differs, deliberately, per the comment). Consider parameterizing `create_pg_connection(env_prefix="DB_")` and delegating here. Separately — and more important — `create_pg_connection` itself is untested: add a unit test that monkeypatches `psycopg2.connect`, sets the `DB_*` env, and asserts the kwargs (incl. `options="-c search_path=<schema>"`) and that defaults apply when unset. It's the single point deciding which DB the queue talks to.

Also consider adding behavioral integration coverage currently missing: FIFO order across multiple messages (`ORDER BY msg_id`), `qty` limit (send 5, `read(qty=2)` → exactly 2 lowest ids), explicit empty-queue read, and a `read_ct` assertion after redelivery.

workers/queue_backend/pg_queue/client.py
muhammad-ali-e: Fixed in `aeb3405c6`. Every op now runs through a `_cursor()` context manager that commits on success and **rolls back on error** (then re-raises). The cached connection is never left in an aborted transaction. Pinned by `test_error_rolls_back_and_reraises`.

workers/queue_backend/pg_queue/client.py
muhammad-ali-e: Fixed in `aeb3405c6`. On `OperationalError`/`InterfaceError` the owned connection is closed and the cached handle dropped, so the next call reconnects — one blip can't permanently wedge the 9c consumer. Added `close()` + context-manager support (`__enter__`/`__exit__`). Injected connections (tests) are never closed (we don't own them).

workers/queue_backend/pg_queue/client.py
muhammad-ali-e: Fixed in `aeb3405c6`. `read()` now raises `ValueError` on non-positive `vt_seconds` (the double-delivery window you flagged) or `qty`, matching the raise-on-invalid posture of `redis_barrier.py`. Tests: `test_read_rejects_non_positive_vt` / `_qty`.

workers/queue_backend/pg_queue/client.py
muhammad-ali-e: Fixed in `aeb3405c6`. `delete()` now logs a WARNING when `rowcount == 0` — "message was already gone (vt likely expired during processing; possible re-delivery)" — so the at-least-once double-process condition is visible instead of silently swallowed. Still returns the bool for callers that want it.

workers/queue_backend/pg_queue/client.py
muhammad-ali-e: Addressed in `aeb3405c6` — added `slots=True` and a docstring noting the payload is the already-decoded JSONB (`dict`), and that `frozen` freezes the binding only so the dict stays mutable (treat as read-only by convention). Deep-freezing the dict isn't worth it for an internal DTO.

workers/queue_backend/pg_queue/connection.py
muhammad-ali-e: Fixed in `aeb3405c6`. `psycopg2.connect` is wrapped: on `OperationalError` it logs a self-identifying error with the non-secret coordinates (host/port/dbname/schema + env_prefix) and re-raises. Covered by `test_connect_failure_is_logged_and_reraised`.

workers/queue_backend/pg_queue/connection.py
muhammad-ali-e: Fixed in `aeb3405c6` (docstring). Dropped the "mirrors `base.py`" claim — the function now says it reads the same variable *names* the backend uses, and that the fallbacks are dev-compose values (host = compose service name), which deployments always override. The mismatch was only in the fallback host, never in real deployments.

workers/queue_backend/pg_queue/connection.py
muhammad-ali-e: Fixed in `aeb3405c6`. Dropped the stale `pg_queue_read` reference — the docstring now says `search_path` is set so the `pg_queue_message` table resolves in the backend's schema (no function exists; the dequeue is inline in the client).

backend/pg_queue/models.py
muhammad-ali-e: Fixed in `aeb3405c6`. Rewrote the model docstring to match reality: table + index are fully Django-managed (generated 0001), no hand-written migration, no DB function, no `CREATE EXTENSION`; the `SKIP LOCKED` dequeue is one statement in the workers client. (These docstrings were left over from an earlier iteration that had a 0002 + function, both since removed.)

backend/pg_queue/models.py
muhammad-ali-e: Fixed in `aeb3405c6`. Reworded to: "Python-level defaults let ORM `.create()` work; the raw-SQL client supplies these columns explicitly (`now()`/`0`), so no DB-level default is required" — matching `send()` which inserts `now(), now(), 0`.

workers/tests/test_pg_queue_client.py
muhammad-ali-e: Fixed in `aeb3405c6`. Narrowed the fixture to `except psycopg2.OperationalError` — only an unreachable/unauthenticated DB skips. `ImportError`, bugs in the helper, and permission/schema errors now surface as real failures instead of a green skip.

workers/tests/test_pg_queue_client.py
muhammad-ali-e: Fixed in `aeb3405c6`. The `queue_name` teardown now calls `pg_conn.rollback()` before the cleanup `DELETE`, so a test that fails mid-transaction doesn't turn one failure into `InFailedSqlTransaction` errors + leaked rows.

workers/tests/test_pg_queue_client.py
muhammad-ali-e: Both addressed in `aeb3405c6`. Parameterised `create_pg_connection(env_prefix="DB_")`; the integration helper now delegates to it via `env_prefix="TEST_DB_"` (only defaulting the host to the dev-compose localhost port), removing the duplicated connect body. And added unit coverage for `create_pg_connection`: env-prefix reading + `search_path`, the `env_prefix` override, and the connect-failure log/raise.

muhammad-ali-e:
### Review feedback addressed — `aeb3405c6`

All 13 threads resolved (all valid). Two themes — **connection robustness** and **stale docstrings** from an earlier iteration:

**Critical / High — robustness (the 9c consumer reuses this connection in a tight loop):**
- Every op now rolls back on error and **recovers a dead connection** (drop cached conn on `OperationalError`/`InterfaceError` → next call reconnects); added `close()` + context manager. One blip no longer wedges the consumer.
- `connect()` wrapped with a self-identifying error log (non-secret host/port/dbname/schema).

**Medium — correctness signals:**
- `read()` raises on non-positive `vt_seconds` (the double-delivery window) / `qty`.
- `delete()` WARNs when no row was removed (the at-least-once re-delivery signal was swallowed).

**High — stale docstrings (left over from the removed `0002`/function):**
- `connection.py` + `models.py` no longer reference a non-existent `pg_queue_read` function, `0002`, or DB-level defaults; corrected to "client supplies columns explicitly".

**Tests:**
- Narrowed integration skip to `OperationalError` (no more masking `ImportError`/bugs as a green skip); rollback before teardown `DELETE`; `_integration_conn` delegates to the now-parameterised `create_pg_connection(env_prefix=…)`; added unit coverage for `create_pg_connection`, `read()` validation, and error-rollback.

**Minor:** `QueueMessage` → `slots=True` + doc that the payload is decoded JSONB.

15 pg_queue tests green (incl. 4 real-PG integration); ruff + pre-commit clean both repos.

workers/queue_backend/pg_queue/client.py
```diff
@@ -0,0 +1,198 @@
...
+_DEQUEUE_SQL = """
+UPDATE pg_queue_message
+   SET vt = now() + make_interval(secs => %s),
+       read_ct = read_ct + 1
+ WHERE msg_id IN (
+     SELECT msg_id
+       FROM pg_queue_message
+      WHERE queue_name = %s
+        AND vt <= now()
+      ORDER BY msg_id
```
muhammad-ali-e: **[Important · Performance] Dequeue forces a sort of the whole visible backlog on every `read()`.** The dequeue index is `(queue_name, vt, msg_id)`, but this subquery filters `vt <= now()` (a range predicate) and then `ORDER BY msg_id`. With `vt` as a range column sitting between the leading equality (`queue_name`) and the sort column, the index can satisfy the filter but cannot emit rows in `msg_id` order — its output is ordered by `(vt, msg_id)`. So Postgres materializes and sorts every visible row for the queue before applying `LIMIT`. On a deep backlog (exactly the condition a queue is built for) each poll becomes an O(N log N) sort instead of an indexed top-N — a real regression under load/contention.

Fix (low-risk): `ORDER BY vt, msg_id`. Still effectively FIFO (`vt` is set to `now()` at enqueue, `msg_id` is a deterministic tiebreak) and lets the existing index drive an indexed top-N with no sort. If strict `msg_id` FIFO is a hard requirement instead, change the index to `(queue_name, msg_id)` and keep `vt <= now()` as a scan-side filter — keep model + migration in sync either way.

workers/queue_backend/pg_queue/client.py
```diff
@@ -0,0 +1,198 @@
...
+        """
+        conn = self.conn
+        try:
+            with conn.cursor() as cur:
+                yield cur
+            conn.commit()
+        except Exception as exc:
+            with contextlib.suppress(Exception):
+                conn.rollback()
+            if self._owns_conn and isinstance(
```
muhammad-ali-e: **[Important · Concurrency/Resilience] Recovery gate keys on exception *type*, so some dead connections are never discarded.** Refines the earlier recovery fix. The cached connection is dropped only on `OperationalError`/`InterfaceError`. But a server-side termination landing mid-statement (PgBouncer `server_lifetime` recycle, `idle_in_transaction` timeout, admin shutdown) can surface as a bare `DatabaseError`/`InternalError`; the `conn.rollback()` above (under `suppress(Exception)`) then *also* fails on the dead socket and is swallowed, so `self._conn` keeps pointing at a dead handle and every subsequent `read()`/`delete()` re-fails. That is exactly the 'one blip permanently wedges the 9c consumer' case the module docstring (L19-22) promises to prevent.

More robust: treat a *failed rollback* as proof the connection is bad. Attempt `conn.rollback()` directly (not blanket-suppressed); if it raises — or `exc` is `Operational`/`Interface` — then `conn.close()` + `self._conn = None`. That recycles any poisoned connection regardless of which psycopg2 subclass was raised, while the bare `raise` still preserves the original error.

workers/queue_backend/pg_queue/client.py
```diff
@@ -0,0 +1,198 @@
...
+dequeue primitive that the enqueue wiring (9b) and the consumer poll
+loop (9c) build on.
+
+Dequeue uses the visibility-timeout pattern: :meth:`PgQueueClient.read`
+runs a single atomic ``UPDATE … WHERE msg_id IN (SELECT … FOR UPDATE
+SKIP LOCKED …) RETURNING …`` (committed immediately), the caller
+processes the message *outside* the transaction, then
+:meth:`PgQueueClient.delete` acks on success. A crash before delete
+leaves the row to reappear once its ``vt`` expires — at-least-once
+delivery, no double-delivery (SKIP LOCKED guarantees a row is claimed by
```
muhammad-ali-e: **[Medium · Contract/Docs] 'no double-delivery' contradicts the at-least-once contract in the same sentence.** At-least-once *permits* a message being delivered/processed more than once — and the crash-before-`delete` → `vt`-expiry → re-claim path described right here *is* a double-delivery. `SKIP LOCKED` only stops two *concurrent* readers from claiming the same currently-visible row; it says nothing about re-delivery after `vt` expiry. The rest of the file agrees: `delete()` says 'the work may be processed twice', L182 'possible re-delivery', `read()` 'a double-delivery window'. As written this could lead the 9c consumer author to skip idempotency.

Suggest: '…at-least-once delivery — SKIP LOCKED prevents two concurrent readers from claiming the same visible row, but a message can still be delivered more than once if a reader crashes before `delete()` and the row's `vt` expires.' Apply the same narrowing to the '(no double-delivery)' parenthetical in the `_DEQUEUE_SQL` comment (~L46) → '(no concurrent double-claim)'.

muhammad-ali-e (commented):
Second-pass review (PR Review Toolkit: code-review, silent-failure, type-design, test, comment, simplify). Round-1 fixes all held — bar raised, so only correctness / concurrency / performance / contract findings below. Code-simplifier found nothing to change.

workers/tests/test_pg_queue_client.py
```diff
@@ -0,0 +1,252 @@
...
+        conn, _ = _mock_conn()
+        with pytest.raises(ValueError, match="vt_seconds"):
+            PgQueueClient(conn=conn).read("q1", vt_seconds=0)
+
+    def test_read_rejects_non_positive_qty(self):
+        conn, _ = _mock_conn()
+        with pytest.raises(ValueError, match="qty"):
+            PgQueueClient(conn=conn).read("q1", qty=0)
+
+    def test_error_rolls_back_and_reraises(self):
```
muhammad-ali-e, line 104: **[Important · Test gap] The connection-recovery and ownership branches — the load-bearing part of the lifecycle design — have no coverage.** This test raises a plain `RuntimeError`, which deliberately does *not* hit the recovery branch in `_cursor()` (client.py:117-122), so the recover-vs-keep distinction is never exercised. A regression that inverts the `isinstance` check, drops the `_owns_conn` guard, or forgets to null `self._conn` would either (a) leave a dead handle cached forever (consumer wedged), or (b) close a caller-injected connection out from under them — and both pass the current suite.

Worth adding (all unit, mocked): (1) owned client + `cur.execute` raising `psycopg2.OperationalError` → assert `rollback` + `close` called and that the *next* call re-invokes `create_pg_connection` (proving the handle was dropped); (2) injected conn → rollback happens but `close()` is NOT called and `self._conn` is unchanged; (3) `close()`/`__exit__` closes an owned conn but never an injected one.

workers/queue_backend/pg_queue/client.py
muhammad-ali-e: Fixed in `9118caea3`. Changed `ORDER BY msg_id` → `ORDER BY vt, msg_id` (your low-risk option). For a fixed `queue_name` the `(queue_name, vt, msg_id)` index is ordered by `(vt, msg_id)`, so the `vt <= now()` range scan now emits rows already sorted and PG applies `LIMIT` as an indexed top-N — no more sorting the whole visible backlog per poll. Still effectively FIFO (vt = now() at enqueue, msg_id tiebreak); index + migration unchanged.

workers/queue_backend/pg_queue/client.py
muhammad-ali-e: Fixed in `9118caea3`. Reworked the recovery gate: rollback is attempted directly (not blanket-suppressed) and **a failed rollback marks the connection dead**, so a poisoned handle is recycled regardless of the psycopg2 subclass (bare `DatabaseError`/`InternalError` from a server-side mid-statement kill included). Also drops the conn if `conn.closed`. The owned-only guard and the bare `raise` are preserved. New tests cover both the OperationalError path and the failed-rollback path.

workers/queue_backend/pg_queue/client.py
muhammad-ali-e: Fixed in `9118caea3`. Reworded the module docstring to: "at-least-once delivery: SKIP LOCKED stops two concurrent readers from claiming the same visible row, but a message can still be delivered more than once if a reader crashes before `delete()` and the row's `vt` expires, so the consumer must be idempotent." Also narrowed the `_DEQUEUE_SQL` parenthetical to "(no concurrent double-claim)". Now consistent with `delete()` / `read()`.

workers/tests/test_pg_queue_client.py
muhammad-ali-e: Fixed in `9118caea3`. Added `TestConnectionLifecycle` covering exactly these branches (all mocked):
1. owned client + `execute` raising `OperationalError` → `rollback` + `close` called, `_conn` dropped, next `.conn` re-invokes `create_pg_connection`;
2. owned client + non-Operational error whose `rollback` also fails → still closed/recycled (the failed-rollback path from the recovery rework);
3. injected conn → rollback happens but `close()` is NOT called and `_conn` is unchanged;
4. `close()` closes an owned conn but never an injected one.

muhammad-ali-e:
### Review round 2 + SonarCloud addressed — `9118caea3`

**Review follow-ups (4, all valid refinements):**
- **Perf:** `ORDER BY (vt, msg_id)` so the dequeue index drives an indexed top-N instead of sorting the whole visible backlog each `read()`.
- **Recovery:** a *failed rollback* now proves the connection is dead → poisoned handles are recycled regardless of psycopg2 subclass (+ `conn.closed` check). Closes the "one blip wedges the consumer" gap more fully.
- **Docs:** clarified at-least-once vs SKIP LOCKED — a message *can* be processed more than once after vt-expiry; "(no double-delivery)" → "(no concurrent double-claim)".
- **Test gap:** added `TestConnectionLifecycle` covering the recover/keep + owned/injected branches.

**SonarCloud S2068 (Quality Gate — Security Rating C):** the `create_pg_connection` test mapped a literal `"PASSWORD": "p"`. Replaced with a runtime uuid token so it isn't flagged as a hard-coded credential — resolves the gate failure (it was a test dummy, not a real credential).

19 pg_queue tests green (incl. 4 real-PG integration + 4 new lifecycle); ruff + pre-commit clean.

workers/tests/test_pg_queue_client.py
```diff
@@ -0,0 +1,321 @@
...
+    def _owned_client(monkeypatch, *, execute_raises=None):
+        cur = MagicMock()
+        if execute_raises is not None:
+            cur.execute.side_effect = execute_raises
+        conn = MagicMock()
+        conn.closed = 0
+        conn.cursor.return_value = _CursorCtx(cur)
+        factory = MagicMock(return_value=conn)
+        monkeypatch.setattr(
+            "queue_backend.pg_queue.client.create_pg_connection", factory
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in `9118caea3` — the test no longer maps a literal `"PASSWORD": "p"`; it uses a runtime `uuid` token, so there's no hard-coded credential (it was a test dummy, not a real secret). The SonarCloud quality gate passes on this commit; this code-scanning alert mirrors the same SonarCloud issue and auto-resolves on the next sync.

PR #2038, opened 2026-06-11
[FIX] Patch frontend Dependabot alerts (axios + transitive deps); drop unused package-lock.json

jaseemjaskp:
## What

Patches the **frontend** critical/high Dependabot alerts and removes the unused `frontend/package-lock.json`.

- `axios` 1.13.5 → **1.16.0** (proxy-auth credential leak, prototype-pollution gadgets, ReDoS)
- Raise caret floors above the vulnerable versions: `handlebars` 4.7.9, `js-cookie` 3.0.8, `happy-dom` 20.10.2, `vite` 7.3.5, `vitest` 3.2.6
- `overrides` for transitive deps that didn't dedupe: `lodash`/`lodash-es` 4.18.0, `immutable` 4.3.8, `socket.io-parser` 4.2.6, and `vite ^7.3.5` (forces `vite-node`'s nested copy off the vulnerable 7.3.1)
- **Delete `frontend/package-lock.json`** — the production image builds with `bun install --frozen-lockfile` from `bun.lock`; the npm lockfile was installed by nothing and only caused lockfile drift + false Dependabot signals
- Git-ignore `package-lock.json` / `yarn.lock` to prevent accidental re-adds
- Repoint the `docker compose watch` rebuild trigger from `package-lock.json` → `bun.lock`

## Why

The frontend had the bulk of the open critical/high npm Dependabot alerts (axios ×11, handlebars critical, vitest critical, vite, etc.). Two committed lockfiles (`bun.lock` + `package-lock.json`) were drifting and only `bun.lock` is actually used to build, so the npm one was both useless and misleading.

## How

- `bun add/update` for the direct bumps, `overrides` in `package.json` for transitive ones, then regenerated `bun.lock`.
- Verified `bun install --frozen-lockfile --ignore-scripts` (the prod Dockerfile's exact command) passes, and `bun run build` succeeds.

No expected breakage.
- Direct bumps are within-range patch/minor updates (axios is the only direct API surface; 1.13→1.16 is backward-compatible).
- Most updated packages are dev/build-time tooling (vite, vitest, happy-dom, rollup, lodash in tests).
- `package-lock.json` removal is safe: no Dockerfile or CI installs from it — the frontend image uses `oven/bun` + `bun install --frozen-lockfile` against `bun.lock`, and frontend CI only runs Biome (no install step).

## Database Migrations

None.

## Env Config

None.

## Relevant Docs

N/A

## Related Issues or PRs

GitHub Dependabot alerts (frontend / npm). Follow-up PRs will cover the Python (`uv.lock`) ecosystems.

## Dependencies Versions

axios 1.16.0 · handlebars 4.7.9 · js-cookie 3.0.8 · happy-dom 20.10.2 · vite 7.3.5 · vitest 3.2.6 · lodash/lodash-es 4.18.0 · immutable 4.3.8 · socket.io-parser 4.2.6

### Known remaining (deferred, tracked separately)
- `picomatch@2.3.1` (via `http-proxy-middleware > micromatch`) — bun has no scoped overrides; low risk (dev proxy tooling).
- `tar@6.2.1` (via `jsdom > canvas > node-pre-gyp`) — needs a major 6→7 bump behind an already-invalid `canvas` peer; handled separately.

## Notes on Testing

- `bun install --frozen-lockfile --ignore-scripts` ✓ (no lock drift)
- `bun run build` ✓
- `biome ci src/` — pre-existing warnings only, no new issues

## Screenshots

N/A — dependency/build-config only.

## Checklist


vishnuszipstack (approved):

jaseemjaskp (commented):
Automated review (PR Review Toolkit). Scope is a focused frontend dependency patch (5 files). Verified: `bun install --frozen-lockfile` is the only install path (frontend.Dockerfile + CI all use bun), there are **zero** remaining references to `package-lock.json` or `npm ci` in the repo, and `bun.lock` carries a matching `overrides` block — so deleting the lockfile and adding it to `.gitignore` is safe. One forward-looking note below; the lodash@4.18.0 regression is already covered by greptile's thread, so not reposting it.

frontend/package.json
```diff
@@ -88,13 +88,20 @@
...
-    "vite": "^7.0.0",
+    "vite": "^7.3.5",
     "vite-plugin-svgr": "^4.5.0",
-    "vitest": "^3.2.0"
+    "vitest": "^3.2.6"
   },
   "engines": {
     "node": ">=20.19.0"
+  },
+  "overrides": {
```
jaseemjaskp, line 100: **Verify each override maps to a real alert, and reconsider exact-pinning (complements greptile's lodash thread — not a duplicate).**

Three distinct points on this new `overrides` block:

1. **`lodash` / `lodash-es` override may be unnecessary.** Before this PR the tree already resolved `lodash@4.17.23` / `lodash-es@4.17.23` (see the bun.lock diff), which is past the `4.17.21` line that patched the known lodash prototype-pollution CVEs. Forcing `4.18.0` doesn't appear to close any alert that `4.17.23` left open — and per greptile it *introduces* a runtime regression. Worth confirming which Dependabot alert this override targets; if none, **removing** the two lodash overrides (letting it stay on the already-safe 4.17.x) is safer than greptile's bump-to-4.18.1.

2. **Exact pins freeze `immutable@4.3.8` and `socket.io-parser@4.2.6` against future patches.** Unlike `vite` (`^7.3.5`), these are pinned exactly, so Dependabot/bun won't pick up a future `4.2.7` / `4.3.9` security fix without another manual PR. If the intent is to lock to a known-good version that's fine — just flagging the maintenance trade-off.

3. **Document the rationale.** JSON can't carry comments, so consider noting in the PR description which alert each override resolves (`socket.io-parser` ≥4.2.4 for the parser advisory, `immutable`, etc.). It makes the next person's audit much faster.

PR #2039, opened 2026-06-11
[FIX] Patch transitive Python Dependabot alerts (h11, pyarrow, pyjwt, azure-core, protobuf) via uv lock

jaseemjaskp:
## What

Patches **transitive Python** Dependabot alerts across 11 `uv.lock` files via lock-only upgrades (no `pyproject.toml` / pinned-dependency changes). Each bump here clears at least one alert:

| Package | Bump | Severity | Lockfiles |
|---|---|---|---|
| **h11** (+ httpcore) | 0.14.0 → **0.16.0** (httpcore 1.0.7 → 1.0.9) | 🔴 critical (request smuggling) | workflow-execution, tool-registry, filesystem |
| **pyarrow** | 18.1.0 → **24.0.0** | high (use-after-free) | root, backend, workers, connectors |
| **pyjwt** | 2.10.x/2.11.0 → **2.13.0** | high (unknown `crit` headers) | backend, workers, connectors, sdk1, platform-service, prompt-service |
| **azure-core** | 1.33/1.35 → **1.41.0** | high (untrusted deserialization) | connectors, sdk1, platform-service, prompt-service |
| **protobuf** | 5.29.4 → **5.29.6** / 6.32.1 → **6.33.6** | high (JSON recursion DoS) | tool-sidecar, sdk1 |

## Why

These are the Python Dependabot alerts whose fix is reachable by a pure lock upgrade — the parent constraints already permit the patched version, so no manifest edit is needed. Kept separate from the pinned-dependency upgrades (Django, Authlib, PyMySQL) which require `pyproject.toml` changes and heavier testing.

## How

- `uv lock --upgrade-package <pkg>` per affected service, restricted to packages that actually cross the alert threshold.
- `h11` was held at 0.14.0 by `httpcore`; bumping `httpcore` to 1.0.9 (a patch) pulls h11 to the patched 0.16.0.

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

Low risk.
- All 11 locks pass `uv lock --check` (resolutions valid & consistent — the same check a frozen Docker build runs).
- Smoke-tested imports: `h11` 0.16.0 / `httpcore` 1.0.9 / `httpx` 0.28.1, and `pyarrow` 24.0.0 / `azure-core` / `pyjwt` 2.13.0 all load cleanly.
[...]

PR #2042, opened 2026-06-11
[FIX] Unblock urllib3 2.x via google-auth bump (clear urllib3 Dependabot alerts)

jaseemjaskp:
## What

Unblocks **urllib3 1.26.20 → 2.7.0** across root / backend / workers / connectors by lifting the constraint that pinned it to the 1.x line.

- `unstract/connectors/pyproject.toml`: `google-auth==2.20.0` → `google-auth>=2.22.0`
- `uv.lock`, `backend/uv.lock`, `workers/uv.lock`, `unstract/connectors/uv.lock`: urllib3 → 2.7.0, google-auth → 2.53.0 (cachetools dropped — no longer a hard dep of google-auth 2.53)

## Why

urllib3 was held at **1.26.20** everywhere, and its CVEs (sensitive-header leakage across redirects, decompression-bomb bypass, unbounded decompression chain) **have no 1.26.x patch — the fixes exist only in 2.x**. The resolver was capped by a single pin: `google-auth==2.20.0`, which declares `urllib3<2.0`. `google-auth>=2.22.0` removed that cap, so urllib3 resolves to 2.7.0 and the alerts clear. It was the *only* `urllib3<2` constraint in the graph (verified by forcing `urllib3>=2.7.0` and reading the resolver conflict).

## How

- Diagnosed the cap by forcing `urllib3>=2.7.0` into the resolver until it named `google-auth==2.20.0`.
- Bumped only the google-auth lower bound (left `google-cloud-bigquery` pinned, since the remaining protobuf cap is separate — see below).
- `>=` matches existing repo style (`boto3~=`, `httpx>=`, `croniter>=`, …).

**This is the highest-risk PR in the Dependabot series and should get real integration testing before merge.** It changes the HTTP layer (urllib3 1.x → 2.x) and the GCP auth library (google-auth 2.20 → 2.53), both used across the cloud connectors (GCS, BigQuery, GCP-auth) and anything doing HTTP via requests/botocore/etc.

Mitigating evidence:
- Tiny, surgical diff — **3 packages per lock** (urllib3 ↑, google-auth ↑, cachetools removed). No wider cascade.
- All 4 locks pass `uv lock --check`.
- Connectors env smoke test: `urllib3` 2.7.0, `google.auth`, `google.cloud.bigquery`, `pymysql` all import cleanly.
- urllib3 2.x is mature; google-auth 2.20→2.53 stays within the same major.

👉 **Please exercise the cloud connectors (GCS / BigQuery / GCP auth) and any outbound-HTTP paths in CI/staging.** A full local run isn't possible (backend env blocked by the unrelated django-celery-beat 2.5.0 wheel quirk); connectors-level imports are verified but live auth/transfer flows are not.

> Consider merging this **after** the lower-risk Dependabot PRs (#2038–#2041) land.

## Database Migrations

None.

## Env Config

None.

## Relevant Docs

[urllib3 2.0 migration guide](https://urllib3.readthedocs.io/en/stable/v2-migration-guide.html) · [google-auth changelog](https://github.com/googleapis/google-auth-library-python/blob/main/CHANGELOG.md)

## Related Issues or PRs

Final PR of the Dependabot remediation series (#2038 frontend, #2039 python-transitive, #2040 Django, #2041 Authlib/PyMySQL).

### Deferred — protobuf 5.x
protobuf is still capped `<5` by additional google/grpc libraries (beyond `google-cloud-bigquery`, which this PR leaves pinned), and 4 → 5 is a major-version jump. It stays on the **4.25.x LTS** line (latest patch), which carries the relevant backported fixes. Moving to 5.x warrants its own PR with a coordinated google-cloud / grpcio upgrade + testing.

## Dependencies Versions

urllib3 1.26.20 → 2.7.0 · google-auth 2.20.0 → 2.53.0

## Notes on Testing

- `uv lock --check` on all 4 workspaces ✓
- Connectors import smoke test (urllib3 / google.auth / bigquery / pymysql) ✓
- Live cloud-connector + HTTP integration testing to be done in CI/staging.

## Screenshots

N/A — dependency change.

## Checklist


unstract/connectors/pyproject.toml
```diff
@@ -22,7 +22,7 @@ classifiers = [
 dependencies = [
     "singleton-decorator~=1.0.0",
     # For GCSHelper
-    "google-auth==2.20.0",
+    "google-auth>=2.22.0",  # >=2.22 drops the urllib3<2 cap
```
jaseemjaskp: **Dependency hygiene — two minor points (separate from the floor-version question already raised):**

**1. Unbounded upper bound breaks this block's pinning convention.** Every other dependency here is exact-pinned (`google-cloud-secret-manager==2.16.1`, `gcsfs==2024.10.0`) or capped (`singleton-decorator~=1.0.0`). `google-auth>=2.22.0` is the only unbounded-upper specifier. The lock pins 2.53.0 so today's builds are reproducible, but a future `uv lock` regen will silently pull whatever google-auth is newest, with no guard against a breaking major.
_Suggest:_ match the siblings with `"google-auth==2.53.0"`, or at minimum cap it: `"google-auth>=2.22.0,<3"`.

**2. Comment is slightly misleading + names an unverifiable version.** "drops the urllib3<2 cap" reads as if google-auth still depends on urllib3 with a relaxed range. Per the lock diff, google-auth 2.53.0 removes `urllib3` from its dependencies **entirely** (2.20.0 listed `urllib3`/`six`/`rsa`/`cachetools`; 2.53.0 lists only `cryptography` + `pyasn1-modules`). Also `2.22` is neither the old (2.20.0) nor new (2.53.0) endpoint and isn't substantiated here, so it's prime comment-rot.
_Suggest:_
```toml
# google-auth >= 2.53 no longer depends on urllib3 at all (2.20 did) — that's what unblocks urllib3 2.x
"google-auth>=2.22.0",
```

jaseemjaskp (commented):
Automated review (PR Review Toolkit). This is a clean, internally-consistent dependency bump; the 4 lock files match each other and the pyproject change, and platform-service/prompt-service/sdk1 were correctly left untouched (they already resolve urllib3 2.7.0). One line carries two minor, non-blocking points below — distinct from the floor-version question @coderabbitai already raised.

unstract/connectors/pyproject.toml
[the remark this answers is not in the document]
jaseemjaskp: Addressed in dfea89605.

- **Comment text fixed**: replaced the inaccurate `>=2.22 drops the urllib3<2 cap` note. The lock resolves `google-auth==2.53.0`, which has no `urllib3` dependency at all (only `cryptography` + `pyasn1-modules`) — that is what actually unblocks urllib3 2.x. The 2.22.0 floor alone does not, as you correctly noted.
- **`urllib3.util.Retry` audit**: no change needed. The only `Retry` construction in the repo (`unstract/core/src/unstract/core/network/retry.py`) already uses `allowed_methods` with no deprecated params/constants — already urllib3 2.x compatible.

jaseemjaskp: Both points addressed in dfea89605.

1. **Upper bound**: capped to `google-auth>=2.22.0,<3` — it was the only unbounded specifier in the block; the lock pin (2.53.0) already satisfies it so no relock needed.
2. **Comment rot**: rewrote the inline comment to note that `google-auth >= 2.53` no longer depends on urllib3 (2.20 did), which is what unblocks urllib3 2.x.

muhammad-ali-e (approved):

PR #2043, opened 2026-06-11
UN-3538 [FEAT] PG Queue Phase 9b — enqueue PG-routed tasks to Postgres

muhammad-ali-e:
## What

Connects the two halves already merged: 8a's routing gate and 9a's `PgQueueClient`. When a task is opted into `WORKER_PG_QUEUE_ENABLED_TASKS`, `dispatch()` now **enqueues it to Postgres** (`pg_queue_message`) instead of the Celery stub.

Targets `feat/UN-3445-pg-queue-integration`.

## The behavioural change (and why it's still safe)

This is the **first** commit where opting a task in actually diverts it off Celery to Postgres.
- **Default-empty allow-list → everything routes to Celery → zero behaviour change.** Only an explicit opt-in enqueues to PG.
- ⚠️ **An opted-in task requires the 9c consumer running**, or the message is durably enqueued but never executed. Documented in the `dispatch()` docstring + `sample.env`. Never flipped in prod until the rollout phase (consumer deployed first).
- **Leaf-only** per the migration-coherence decision — pipeline tasks (`async_execute_bin`) stay on Celery until execution-level routing (9e).

## Changes

- **`queue_backend/dispatch.py`** — PG branch builds a `TaskPayload`, enqueues via a process-singleton `PgQueueClient` (reused across dispatches; self-recovers a dead connection), and returns a `PgDispatchHandle` (`.id = msg_id`, satisfies `TaskHandle`). A PG enqueue failure raises — no silent Celery fallback (would hide the failure / risk double-dispatch). Cutover log at INFO, once per task.
- **`queue_backend/pg_queue/task_payload.py`** *(new)* — `TaskPayload` TypedDict (`task_name`/`args`/`kwargs`/`queue`/`fairness`) + `to_payload()`. This is the *contents* of the `pg_queue_message.message` JSONB column — distinct from the backend's `PgQueueMessage` **row** model (envelope vs payload; they nest, not duplicate). The 9c consumer decodes it.
- **`sample.env`** — warns that a listed task enqueues to PG and won't run without the consumer.
- **`backend/pg_queue/models.py`** — note that `message` holds a `TaskPayload`.
- **tests** — `TestDispatchRouting` (PG → enqueue, Celery → `send_task`, default queue, handle `.id`), `TestCutoverLog`, `to_payload` shape + JSON-serialisable, and a **real-PG integration** test (`dispatch()` lands a decodable row).

## Testing

- 74 queue_backend tests green (incl. the real-PG integration); ruff + pre-commit clean.
- Dev-tested end-to-end on the dev DB via the real singleton path: `dispatch('demo_leaf', queue='demo_q', fairness=…)` → row in `pg_queue_message` (msg_id 60), cutover log fired, payload decodes exactly, cleaned up.

## Out of scope

- The consumer poll loop that drains + runs the task — that's **9c** (the first real end-to-end PG execution).

Ticket: UN-3538 (parent UN-3536, epic UN-3445)

### DIFF
--- a/workers/queue_backend/dispatch.py
+++ b/workers/queue_backend/dispatch.py
@@ -20,6 +20,7 @@ unchanged unless an operator explicitly opts a task in.
 from __future__ import annotations
 
 import logging
+import threading
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
 from typing import Any
@@ -40,16 +41,22 @@ _DEFAULT_PG_QUEUE = "default"
 # log to once per task name (per prefork child).
 _pg_routing_logged: set[str] = set()
 
-# Process-level client, reused across dispatches (it self-recovers a dead
-# connection). One per prefork child.
-_pg_client: PgQueueClient | None = None
+# Per-thread client, reused across dispatches (it self-recovers a dead
+# connection). The worker pool is prefork (each child a single thread), so
+# this is effectively one client per child; ``threading.local`` keeps it
+# correct under a ``-P threads`` pool too, since a libpq connection is not
+# safe for concurrent use across threads. (A gevent pool shares one thread
+# across greenlets and would need a connection pool — out of scope while
+# prefork is the deployment.)
+_pg_local = threading.local()
 
 
 def _get_pg_client() -> PgQueueClient:
-    global _pg_client
-    if _pg_client is None:
-        _pg_client = PgQueueClient()
-    return _pg_client
+    client = getattr(_pg_local, "client", None)
+    if client is None:
+        client = PgQueueClient()
+        _pg_local.client = client
+    return client
 
 
 @dataclass(frozen=True, slots=True)
@@ -102,22 +109,34 @@ def _enqueue_pg(
     A PG enqueue failure raises (no silent Celery fallback — that would
     hide the failure or risk double-dispatch).
     """
+    pg_queue = queue or _DEFAULT_PG_QUEUE
+    if task_name not in _pg_routing_logged:
+        # Log the routing *decision* once per task, BEFORE the send — it's
+        # true regardless of send outcome, so a first-dispatch failure
+        # (DB down / unmigrated) must not suppress the one announcement.
+        # INFO so it survives a default log config; once-per-task bounds it.
+        _pg_routing_logged.add(task_name)
+        logger.info(
+            "PG-queue: routing task=%r to Postgres (queue=%r). Requires the "
+            "PG consumer to be running, or it will not execute.",
+            task_name,
+            pg_queue,
+        )
     payload = to_payload(
         task_name, args=args, kwargs=kwargs, queue=queue, fairness=fairness
     )
-    msg_id = _get_pg_client().send(
-        queue or _DEFAULT_PG_QUEUE,
-        payload,
-        org_id=fairness.org_id if fairness is not None else None,
-    )
-    if task_name not in _pg_routing_logged:
-        # INFO + once-per-task: visible under a default log config, bounded.
-        _pg_routing_logged.add(task_name)
-        logger.info(
-            "PG-queue: task=%r enqueued to Postgres (queue=%r, msg_id=%s). "
-            "Requires the PG consumer to be running, or it will not execute.",
-            task_name,
-            queue or _DEFAULT_PG_QUEUE,
-            msg_id,
+    try:
+        msg_id = _get_pg_client().send(
+            pg_queue,
+            payload,
+            org_id=fairness.org_id if fairness is not None else None,
         )
+    except Exception:
+        # Re-raise with a breadcrumb (raw psycopg2.Error / a json.dumps
+        # TypeError on a non-serialisable arg would otherwise propagate with
+        # no "this was a PG-routed dispatch" context). No Celery fallback.
+        logger.exception(
+            "PG-queue: failed to enqueue task=%r to queue=%r", task_name, pg_queue
+        )
+        raise
     return PgDispatchHandle(id=str(msg_id))
--- a/workers/queue_backend/dispatch.py
+++ b/workers/queue_backend/dispatch.py
@@ -1,28 +1,66 @@
 """Transport-agnostic task dispatch.
 
-Thin pass-through to ``celery.current_app.send_task``; the indirection
-is the seam — a future per-task router can land here without touching
-call sites.
+Routes each task to its transport via :func:`select_backend`:
+
+- **Celery** (default) — a thin pass-through to ``current_app.send_task``.
+- **PG Queue** — when a task is opted into ``WORKER_PG_QUEUE_ENABLED_TASKS``,
+  the task is serialised and enqueued to ``pg_queue_message`` (9b); the PG
+  consumer (9c) drains and runs it.
+
+The default (empty allow-list) routes everything to Celery, so dispatch is
+unchanged unless an operator explicitly opts a task in.
+
+.. warning::
+   A task opted into the PG queue **requires the PG consumer to be running**
+   — otherwise the message is durably enqueued but never executed. Only opt
+   in tasks once the consumer is deployed (and, per the migration-coherence
+   decision, only *leaf* tasks until execution-level routing exists).
 """
 
 from __future__ import annotations
 
 import logging
 from collections.abc import Mapping, Sequence
+from dataclasses import dataclass
 from typing import Any
 
 from celery import current_app
 
 from .fairness import FairnessKey
 from .handle import DispatchHandle
+from .pg_queue import PgQueueClient, to_payload
 from .routing import QueueBackend, select_backend
 
 logger = logging.getLogger(__name__)
 
-# Task names already logged as PG-routed in this process. Bounds the
-# routing log to once per task name (per prefork child) so an opted-in
-# high-throughput task doesn't log on every dispatch.
+# pg_queue_message.queue_name used when a dispatch carries no Celery queue.
+_DEFAULT_PG_QUEUE = "default"
+
+# Task names already logged as PG-routed in this process — bounds the cutover
+# log to once per task name (per prefork child).
 _pg_routing_logged: set[str] = set()
+
+# Process-level client, reused across dispatches (it self-recovers a dead
+# connection). One per prefork child.
+_pg_client: PgQueueClient | None = None
+
+
+def _get_pg_client() -> PgQueueClient:
+    global _pg_client
+    if _pg_client is None:
+        _pg_client = PgQueueClient()
+    return _pg_client
+
+
+@dataclass(frozen=True, slots=True)
+class PgDispatchHandle:
+    """:class:`~queue_backend.handle.TaskHandle` for a PG-enqueued task.
+
+    ``id`` is the ``pg_queue_message.msg_id`` (as a string), so call sites
+    that read ``handle.id`` keep working across the Celery/PG boundary.
+    """
+
+    id: str
 
 
 def dispatch(
@@ -33,32 +71,14 @@ def dispatch(
     queue: str | None = None,
     fairness: FairnessKey | None = None,
 ) -> DispatchHandle:
-    """Enqueue a task by name.
+    """Enqueue a task by name onto its selected transport.
 
-    ``fairness`` is attached as the ``x-fairness-key`` header (not in
-    kwargs). Pass ``None`` for non-workflow worker tasks.
-
-    The transport is chosen by :func:`select_backend` from the PG-queue
-    routing table (``WORKER_PG_QUEUE_ENABLED_TASKS``). In this phase the
-    table is a scaffold: PG-selected tasks are *logged* but still
-    dispatched via Celery, because no PG consumer exists yet. The
-    ``QueueBackend.PG`` branch below is the seam where the real PG
-    enqueue lands in a later phase — keeping the ``send_task`` call
-    outside it guarantees today's wire is byte-identical regardless of
-    the routing decision.
+    ``fairness`` is attached as the ``x-fairness-key`` header on the Celery
+    path / serialised into the message on the PG path. Pass ``None`` for
+    non-workflow worker tasks.
     """
-    if (
-        select_backend(task_name) is QueueBackend.PG
-        and task_name not in _pg_routing_logged
-    ):
-        # INFO (not DEBUG) so the cutover is visible under a default log
-        # config; log-once per task name keeps the volume bounded.
-        _pg_routing_logged.add(task_name)
-        logger.info(
-            "PG-queue routing selected for task=%r; dispatching via "
-            "Celery (scaffold — no PG consumer yet)",
-            task_name,
-        )
+    if select_backend(task_name) is QueueBackend.PG:
+        return _enqueue_pg(task_name, args, kwargs, queue, fairness)
 
     headers = fairness.as_header() if fairness is not None else None
     return current_app.send_task(
@@ -68,3 +88,36 @@ def dispatch(
         queue=queue,
         headers=headers,
     )
+
+
+def _enqueue_pg(
+    task_name: str,
+    args: Sequence[Any] | None,
+    kwargs: Mapping[str, Any] | None,
+    queue: str | None,
+    fairness: FairnessKey | None,
+) -> PgDispatchHandle:
+    """Serialise + enqueue a PG-routed task to ``pg_queue_message``.
+
+    A PG enqueue failure raises (no silent Celery fallback — that would
+    hide the failure or risk double-dispatch).
+    """
+    payload = to_payload(
+        task_name, args=args, kwargs=kwargs, queue=queue, fairness=fairness
+    )
+    msg_id = _get_pg_client().send(
+        queue or _DEFAULT_PG_QUEUE,
+        payload,
+        org_id=fairness.org_id if fairness is not None else None,
+    )
+    if task_name not in _pg_routing_logged:
+        # INFO + once-per-task: visible under a default log config, bounded.
+        _pg_routing_logged.add(task_name)
+        logger.info(
+            "PG-queue: task=%r enqueued to Postgres (queue=%r, msg_id=%s). "
+            "Requires the PG consumer to be running, or it will not execute.",
+            task_name,
+            queue or _DEFAULT_PG_QUEUE,
+            msg_id,
+        )
+    return PgDispatchHandle(id=str(msg_id))
--- a/workers/queue_backend/pg_queue/__init__.py
+++ b/workers/queue_backend/pg_queue/__init__.py
@@ -14,12 +14,15 @@ A subpackage (rather than a single module like the barriers) because the
 real implementation will likely span several modules (config, consumer
 poll loop, orchestrator) — exact layout TBD.
 
-Phase 9a adds the storage + dequeue primitive (:class:`PgQueueClient`
-over the ``pg_queue_message`` table; the ``SKIP LOCKED`` dequeue lives in
-the client). It is still **inert**: routing decisions are made by
-:func:`queue_backend.routing.select_backend`, and until the enqueue
-wiring (9b) and consumer (9c) land, PG-selected dispatches still ride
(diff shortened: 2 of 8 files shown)


muhammad-ali-e (commented):
## Automated PR Review — PG Queue Phase 9b (enqueue wiring)

Review produced with the PR Review Toolkit (code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier). The PR is well-scoped, well-tested on the happy path, and the no-silent-fallback design is sound and clearly documented. Findings are posted inline below, prioritised in the PR thread summary.

**Highest priority:** (1) add a failure-path test that pins the no-fallback contract + the log-once-not-poisoned ordering; (2) make the process-level `_pg_client` singleton thread-safe before any opted-in task can run under a `threads`/`gevent` pool.

**Out-of-diff comment rot this PR makes false (can't be commented inline — please fix in a follow-up or amend):**
- `workers/queue_backend/routing.py:36-40` — "Scaffold posture" still says *dispatch() still sends PG-selected tasks via Celery ... inert*. False as of 9b — `dispatch()` now enqueues to Postgres. ("no PG consumer yet" is still true.)
- `workers/queue_backend/pg_queue/client.py:3-6` — "**Inert in this phase** — nothing in `dispatch()` calls this yet". False — `_enqueue_pg()` now calls `_get_pg_client().send()`.
- `workers/queue_backend/pg_queue/__init__.py:17-22` — module docstring still describes 9a "still inert / still ride Celery" (the import lines are in-diff, the docstring above is not; see inline note).

workers/queue_backend/dispatch.py
```diff
@@ -1,29 +1,67 @@
...
+# Task names already logged as PG-routed in this process — bounds the cutover
+# log to once per task name (per prefork child).
 _pg_routing_logged: set[str] = set()
 
+# Process-level client, reused across dispatches (it self-recovers a dead
+# connection). One per prefork child.
+_pg_client: PgQueueClient | None = None
+
+
+def _get_pg_client() -> PgQueueClient:
```
muhammad-ali-e, line 54: **[Important — concurrency]** The process-level `_pg_client` wraps a single psycopg2 connection that `_get_pg_client()` returns to every caller. A single libpq connection is **not safe for concurrent use across threads** — concurrent `send()` (execute/commit/rollback) on one connection corrupts protocol state.

The docstring's "one per prefork child" assumption holds for the default prefork pool, but `run-worker.sh -P threads|gevent` is supported for any worker and `FILE_PROCESSING` already defaults to `pool_type: threads`. The `queue_backend.dispatch` seam isn't reached from the threads-pool `ExecutionDispatcher` path *today* (that goes straight to `send_task`), so this is latent rather than live — but it will race the moment a leaf task dispatched via this seam (e.g. `send_webhook_notification`) runs under a non-prefork pool. The check-then-set in `_get_pg_client()` itself is also racy (two threads → two clients → leaked connection).

**Fix:** guard with a `threading.Lock` around get+`send()`, or make the client `threading.local()`, or open a short-lived connection per enqueue. A module-level lock is simplest and preserves the self-recovering reuse property. Add it before opting any seam-dispatched task in.

workers/queue_backend/dispatch.py
```diff
@@ -68,3 +88,36 @@ def dispatch(
...
+) -> PgDispatchHandle:
+    """Serialise + enqueue a PG-routed task to ``pg_queue_message``.
+
+    A PG enqueue failure raises (no silent Celery fallback — that would
+    hide the failure or risk double-dispatch).
+    """
+    payload = to_payload(
+        task_name, args=args, kwargs=kwargs, queue=queue, fairness=fairness
+    )
+    msg_id = _get_pg_client().send(
```
muhammad-ali-e: **[Medium — error context]** The no-Celery-fallback decision here is correct and well-documented (raising avoids double-dispatch / masking outages) — keep it. The gap is *context*: `send()` re-raises a raw `psycopg2.Error`, and `json.dumps()` inside `send()` can raise a bare `TypeError` for a non-serialisable arg/kwarg — both propagate with no "this was a PG-routed dispatch" breadcrumb, and most call sites (`api-deployment/tasks.py:737`, notification `helper.py:76`) don't wrap `dispatch()`.

**Fix** — log-and-reraise so the contract is intact but the failure is attributable:
```python
try:
    msg_id = _get_pg_client().send(queue or _DEFAULT_PG_QUEUE, payload, org_id=...)
except Exception:
    logger.exception(
        "PG-queue: enqueue failed for task=%r (queue=%r); not falling back to "
        "Celery. Task was NOT enqueued.", task_name, queue or _DEFAULT_PG_QUEUE)
    raise
```

workers/queue_backend/dispatch.py
```diff
@@ -68,3 +88,36 @@ def dispatch(
...
+        queue or _DEFAULT_PG_QUEUE,
+        payload,
+        org_id=fairness.org_id if fairness is not None else None,
+    )
+    if task_name not in _pg_routing_logged:
```
muhammad-ali-e, line 113: **[Medium — observability]** The once-per-task cutover log is emitted only *after* a successful `send()`. If the very first PG dispatch for a task fails (DB down, migration not applied), the one INFO line that announces the routing decision is suppressed exactly in the failure case. The log is about the routing *decision*, which is true regardless of send outcome.

**Fix:** move the log-once block *before* `send()` and reword to "routing task to Postgres" (drop the per-message `msg_id`, which isn't a per-cutover fact). This also locks in the property that a failed first send must not poison `_pg_routing_logged` — see the suggested failure-path test on `test_routing.py`.

workers/queue_backend/dispatch.py
```diff
@@ -68,3 +88,36 @@ def dispatch(
...
+        # INFO + once-per-task: visible under a default log config, bounded.
+        _pg_routing_logged.add(task_name)
+        logger.info(
+            "PG-queue: task=%r enqueued to Postgres (queue=%r, msg_id=%s). "
+            "Requires the PG consumer to be running, or it will not execute.",
+            task_name,
+            queue or _DEFAULT_PG_QUEUE,
```
muhammad-ali-e: **[Nit — simplify]** `queue or _DEFAULT_PG_QUEUE` is evaluated twice (here and at the `send()` call on line 109). Optional: hoist a local `pg_queue = queue or _DEFAULT_PG_QUEUE` at the top of `_enqueue_pg` and use it in both places — one source of truth for the fallback. Marginal; the current form is already clear.

workers/queue_backend/pg_queue/task_payload.py
```diff
@@ -0,0 +1,50 @@
...
+    ``fairness`` is the serialised :class:`FairnessKey` (``to_dict()``) or
+    ``None`` — the consumer rebuilds the ``x-fairness-key`` header from it,
+    mirroring the Celery dispatch path.
+    """
+
+    task_name: str
+    args: list[Any]
+    kwargs: dict[str, Any]
+    queue: str | None
+    fairness: dict[str, Any] | None
```
muhammad-ali-e: **[Suggestion — type design]** `fairness: dict[str, Any] | None` is strictly weaker than the source it's built from: `FairnessKey.to_dict() -> dict[str, str | int | None]` (`fairness.py:50`). You're discarding precision you already have, and the consumer (9c) has to rediscover the sub-shape from prose.

**Suggestion:** declare a nested `TypedDict` and share it both ways:
```python
class FairnessPayload(TypedDict):
    org_id: str | None
    workload_type: str
    pipeline_priority: int
# fairness: FairnessPayload | None
```
Then `FairnessKey.to_dict()` can be annotated `-> FairnessPayload`, making producer and wire-contract share one type. (TypedDict over a dataclass is the right call for a JSONB wire row — don't change that.)

workers/sample.env
```diff
@@ -143,10 +143,13 @@ WORKER_BARRIER_KEY_TTL_SECONDS=21600
 # listed. Unset / empty (the default) → everything routes to Celery,
 # exactly as today.
 #
-# Scaffold note: in this release there is no PG consumer yet, so
-# PG-selected tasks are logged but still dispatched via Celery. Setting
-# this flag is therefore safe to test routing decisions locally with
-# zero behaviour change. No Flipt server required.
+# Default-empty → everything routes to Celery, zero behaviour change. No
```
muhammad-ali-e: **[Nit — docs]** This duplicates line 143-144 ("Unset / empty (the default) → everything routes to Celery, exactly as today."). Two adjacent sentences making the same claim risk diverging when one is edited. Suggest folding the useful new bits ("zero behaviour change", "No Flipt server required.") onto the existing line and dropping this standalone repeat. The ⚠️ warning block just below is accurate and valuable — keep it.

workers/queue_backend/pg_queue/__init__.py
```diff
@@ -28,9 +28,12 @@
 
 from .client import PgQueueClient, QueueMessage
 from .connection import create_pg_connection
+from .task_payload import TaskPayload, to_payload
```
muhammad-ali-e, line 34: **[Important — comment rot]** These imports are correct, but the module docstring above (lines 17-22, not in this diff hunk) still describes 9a state: *"It is still **inert**: ... PG-selected dispatches still ride Celery"*. This PR is the 9b enqueue wiring — that statement is now false. Please update the docstring (and the file's "scaffold" framing) to reflect that PG-selected tasks now enqueue to `pg_queue_message`.

workers/tests/test_routing.py
```diff
@@ -119,79 +120,81 @@ def test_empty_allow_list_logs_nothing(self, caplog):
...
+    client = MagicMock()
+    client.send.return_value = msg_id
+    # Patch on the module object (string target would navigate the shadowing
+    # ``dispatch`` *function*, not the submodule).
+    monkeypatch.setattr(dispatch_mod, "_get_pg_client", lambda: client)
+    return client
 
-# --- dispatch() is inert under routing (the scaffold invariant) ---
 
+class TestDispatchRouting:
```
muhammad-ali-e, line 136: **[Critical — test gap]** The documented contract "a PG enqueue failure raises (no silent Celery fallback)" (`dispatch.py:102-103`) has no test. A future refactor adding a `try/except`-fallback would reintroduce double-dispatch risk and every existing test would still pass.

**Add** (mocked, runs in CI without a DB):
```python
def test_pg_enqueue_failure_propagates_and_does_not_fall_back(self, monkeypatch):
    monkeypatch.setenv(ENABLED_TASKS_ENV, "t1")
    client = MagicMock(); client.send.side_effect = RuntimeError("db down")
    monkeypatch.setattr(dispatch_mod, "_get_pg_client", lambda: client)
    with patch("queue_backend.dispatch.current_app") as mock_app:
        with pytest.raises(RuntimeError):
            dispatch("t1")
    mock_app.send_task.assert_not_called()             # no silent Celery fallback
    assert "t1" not in dispatch_mod._pg_routing_logged  # log-once guard not poisoned
```
The last assertion guards the log-ordering property (see the `dispatch.py:113` comment) and is the highest-value line.

workers/tests/test_routing.py
```diff
@@ -119,79 +120,81 @@ def test_empty_allow_list_logs_nothing(self, caplog):
...
+        with (
+            patch("queue_backend.dispatch.current_app") as mock_app,
+            patch("queue_backend.dispatch._get_pg_client") as mock_get,
+        ):
+            dispatch("t1", args=["a"], kwargs={"k": "v"}, queue="general")
+        mock_app.send_task.assert_called_once()
+        mock_get.assert_not_called()
 
-    def _capture_send_task(self, task_name, fairness):
+    def test_pg_selected_enqueues_to_pg_not_celery(self, monkeypatch):
```
muhammad-ali-e, line 148: **[Important — test gap]** Every test monkeypatches `_get_pg_client` away, so the real lazy-init + reuse body (`dispatch.py:48-52`) has zero coverage — a regression that constructs a fresh `PgQueueClient()` per dispatch (defeating connection reuse/self-recovery) would pass the whole suite. The sibling `redis_barrier` singleton has a dedicated test for this (`test_redis_barrier.py:888`).

**Add** (no DB needed — patch the ctor):
```python
def test_get_pg_client_lazily_inits_and_reuses(monkeypatch):
    dispatch_mod._pg_client = None
    sentinel = object(); ctor = MagicMock(return_value=sentinel)
    monkeypatch.setattr(dispatch_mod, "PgQueueClient", ctor)
    assert dispatch_mod._get_pg_client() is sentinel
    assert dispatch_mod._get_pg_client() is sentinel
    ctor.assert_called_once()
```

workers/tests/test_dispatch_pg.py
```diff
@@ -0,0 +1,137 @@
...
+        if cur.fetchone()[0] is None:
+            conn.close()
+            pytest.skip("pg_queue migration not applied (run backend migrate)")
+    yield PgQueueClient(conn=conn)
+    conn.rollback()
+    conn.close()
+
+
+class TestDispatchEnqueueIntegration:
+    def test_dispatch_lands_decodable_row(self, monkeypatch, pg_client):
```
muhammad-ali-e, line 96: **[Important — CI coverage hole]** This is the only test that drives a real `PgQueueClient.send` through `dispatch()` (the full `dispatch → _enqueue_pg → to_payload → INSERT → read-back` round-trip and the producer↔consumer JSONB contract). It `pytest.skip`s when Postgres is unreachable/unmigrated (lines 84, 89). If CI runs without a migrated Postgres, the entire enqueue round-trip contributes **zero** coverage and the suite still goes green.

**Action:** confirm CI provisions a migrated Postgres so this skip is effectively never hit there, or add a CI-visible guard (e.g. xfail-on-skip / a required marker) so an accidentally-skipped integration test is noticed rather than passing silently. No mocked test can substitute — the value is the real INSERT/JSONB round-trip.

workers/queue_backend/dispatch.py
muhammad-ali-e: Fixed in `eb73aba2c`. Switched the singleton to `threading.local` — one client per *thread*. Under the prefork pool (the deployment) each child is single-threaded so it's one per child as before; under a `-P threads` pool each worker thread gets its own connection, so concurrent `send()` can't corrupt protocol state. Documented that a gevent pool (one thread, many greenlets) would need a connection pool — out of scope while prefork is the deployment.

workers/queue_backend/dispatch.py
muhammad-ali-e: Fixed in `eb73aba2c`. The enqueue is now wrapped: on any failure it `logger.exception("PG-queue: failed to enqueue task=%r to queue=%r", …)` and re-raises — so a raw `psycopg2.Error` or a `json.dumps` `TypeError` on a non-serialisable arg carries a "this was a PG-routed dispatch" breadcrumb. No Celery fallback, as you noted.

workers/queue_backend/dispatch.py
muhammad-ali-e: Fixed in `eb73aba2c`. Moved the once-per-task log *before* `send()` and reworded it to announce the routing decision (`"routing task=… to Postgres"`) rather than the result — so a first-dispatch failure no longer suppresses it. (This made the suggested `"not in _pg_routing_logged"` assertion in the test-gap comment incorrect; I pinned the new ordering with `assert "t1" in _pg_routing_logged` instead — see that thread.)

workers/queue_backend/dispatch.py
muhammad-ali-e: Done in `eb73aba2c` — hoisted `pg_queue = queue or _DEFAULT_PG_QUEUE` at the top of `_enqueue_pg`, used in both the log and the `send()`.

workers/queue_backend/pg_queue/task_payload.py
muhammad-ali-e: Done in `eb73aba2c`. Added `FairnessPayload(TypedDict)` in `fairness.py`, annotated `FairnessKey.to_dict() -> FairnessPayload`, and `TaskPayload.fairness` is now `FairnessPayload | None`. Producer and wire-contract share one precise type; the 9c consumer gets the sub-shape for free.

workers/sample.env
muhammad-ali-e: Fixed in `eb73aba2c` — folded "zero behaviour change" + "No Flipt server required" onto the existing default-routes-to-Celery sentence and dropped the standalone repeat. The ⚠️ consumer-required warning is kept.

workers/queue_backend/pg_queue/__init__.py
muhammad-ali-e: Good catch — fixed in `eb73aba2c`. Rewrote the `pg_queue/__init__` docstring to describe 9b state: 9b wires `dispatch()` to enqueue PG-selected tasks here as a `TaskPayload` instead of riding Celery; the consumer is 9c; default-empty flag keeps it inert unless a task is opted in. The stale "still inert / still ride Celery" framing is gone.

workers/tests/test_routing.py
muhammad-ali-e: Added in `eb73aba2c` (`test_pg_enqueue_failure_propagates_and_does_not_fall_back`): a mocked `client.send` raising → `dispatch()` raises and `current_app.send_task` is never called (no silent fallback). One change from your snippet: since the log now fires *before* `send()` (the `dispatch.py:113` thread), I assert `"t1" in _pg_routing_logged` to pin the new log-ordering, rather than `not in`.

workers/tests/test_routing.py
muhammad-ali-e: Added in `eb73aba2c` (`test_get_pg_client_lazily_inits_and_reuses`): patches the `PgQueueClient` ctor, calls `_get_pg_client()` twice, asserts the same instance and `ctor.assert_called_once()` — so a regression that builds a fresh client per dispatch (defeating reuse/self-recovery) now fails. Adapted to the per-thread storage (resets `_pg_local.client`).

workers/tests/test_dispatch_pg.py
muhammad-ali-e: The dispatch → `_enqueue_pg` → `to_payload` → `client.send(...)` path is actually covered in CI without a DB: `TestDispatchRouting.test_pg_selected_enqueues_to_pg_not_celery` asserts the exact `queue_name` + payload reach a mocked `send`, and the `to_payload` shape has its own unit tests. The integration test adds only the *real* `INSERT` + read-back (the PgQueueClient's own SQL, which `test_pg_queue_client.py` also covers via mocks + its own integration). So the round-trip's logic has CI coverage; only the real-Postgres seam is skip-gated by design. Happy to add a DB service to CI separately if we want the seam itself always exercised.

muhammad-ali-e:
### Review feedback addressed — `eb73aba2c`

All 10 threads (all valid):

**Concurrency / robustness:**
- Per-thread PG client (`threading.local`) — a libpq connection isn't safe across threads; correct under prefork AND `-P threads`.
- Enqueue wrapped with a `logger.exception` breadcrumb (raw `psycopg2.Error` / `json.dumps` `TypeError` otherwise propagate context-free). No Celery fallback.
- Routing-decision log moved *before* `send()` so a first-dispatch failure still announces it.

**Type design:**
- Shared `FairnessPayload` TypedDict — `FairnessKey.to_dict() -> FairnessPayload`, `TaskPayload.fairness` uses it (was loose `dict[str, Any]`).

**Test gaps (Critical + Important):**
- `test_pg_enqueue_failure_propagates_and_does_not_fall_back` — pins the no-silent-fallback contract.
- `test_get_pg_client_lazily_inits_and_reuses` — pins client reuse (a per-dispatch-construct regression now fails).

**Comment rot / nits:** `pg_queue/__init__` docstring updated to 9b state (no longer "inert / rides Celery"); `sample.env` de-duplicated; `pg_queue` fallback hoisted to one local.

Note on the conflict between the two test-gap comments: the log now fires *before* `send()`, so I pinned `"t1" in _pg_routing_logged` rather than the suggested `not in`.

82 queue_backend tests green (incl. 2 new + real-PG integration); ruff + pre-commit clean; SonarCloud gate passed.

PR #2045, opened 2026-06-12
UN-3539 [FEAT] PG Queue Phase 9c — consumer poll loop (claim → run → ack)

muhammad-ali-e:
## What

The consumer half — makes a PG-routed task actually **execute**. `PgQueueConsumer` drains `pg_queue_message` and runs each claimed task in-process. Completes the leaf-first end-to-end path: **8a route → 9b enqueue → 9a store → 9c consume+run.**

Targets `feat/UN-3445-pg-queue-integration`.

## Validated live (dev stack)

A real API execution's Slack webhook was opted into PG and the full loop ran end-to-end:
```
dispatch(send_webhook_notification) → pg_queue_message row (queue=notifications)
consumer.poll_once() → claim → run → Webhook sent to Slack (HTTP 200) → ack (row removed)
```
A production task executed entirely off Postgres, no Celery broker.

## Changes

- **`queue_backend/pg_queue/consumer.py`** *(new)* — `PgQueueConsumer`:
  - `poll_once()` — claims a batch via `PgQueueClient.read` (SKIP LOCKED + visibility timeout), runs each task via `current_app.tasks[name].apply(args, kwargs, throw=True)` (executes the body in-process), and **acks** by `delete()` on success.
  - **Task failure** → leave the row; its `vt` expires and it's redelivered (at-least-once → tasks must be idempotent). **Unknown task** → drop + `logger.error` (no poison loop).
  - `run()` — poll loop with empty-queue backoff + SIGTERM/SIGINT graceful stop. `main()` — `python -m queue_backend.pg_queue.consumer`, env-configured.
- **tests** — 5 unit (run+ack, fail→no-ack, unknown→drop, empty, graceful stop) + 1 real-PG integration (enqueue → poll → execute → ack).

## Deployment note (not consumer logic)

The consumer **process** must bootstrap the worker app (import the task modules) so `current_app.tasks` resolves the task — otherwise it treats every task as unknown. This is an entrypoint/rollout concern: the rollout phase wires a `worker-pg-queue-consumer` container that boots the worker app before polling. Documented in the module docstring.

## Out of scope (later)

- Poison-message dead-letter via `read_ct` cap; multi-queue fan-out; the single-orchestrator + fair admission (9d); container/K8s wiring (rollout).

Ticket: UN-3539 (parent UN-3536, epic UN-3445)

### DIFF
--- a/workers/queue_backend/pg_queue/client.py
+++ b/workers/queue_backend/pg_queue/client.py
@@ -79,13 +79,15 @@ class QueueMessage:
     ``message`` is the already-decoded JSONB payload (psycopg2 parses
     ``jsonb`` to a Python ``dict``). ``frozen`` freezes the binding only —
     the ``dict`` itself is mutable, so treat the payload as read-only by
-    convention. ``read_ct`` is the post-claim delivery count (1 on the
-    first claim), so consumers can cap redelivery of a poison message.
+    convention. ``read_ct`` is the post-claim delivery count — always ``>= 1``
+    from the dequeue (``read_ct = read_ct + 1 RETURNING``), so consumers can
+    cap redelivery of a poison message. No default: a ``0`` ("never claimed")
+    can't come from the dequeue and would silently bypass the poison guard.
     """
 
     msg_id: int
     message: dict[str, Any]
-    read_ct: int = 0
+    read_ct: int
 
 
 class PgQueueClient:
@@ -191,16 +193,17 @@ class PgQueueClient:
 
         ``False`` means the row was already gone — typically its visibility
         timeout expired during processing and another worker (re)claimed it,
-        i.e. the work may be processed twice. Logged at WARNING so this
-        at-least-once condition is visible rather than silently swallowed.
+        i.e. the work may be processed twice. Logged at DEBUG here; the
+        consumer emits the contextual WARNING (it names the task), so this
+        avoids a duplicate warning per double-run.
         """
         with self._cursor() as cur:
             cur.execute("DELETE FROM pg_queue_message WHERE msg_id = %s", (msg_id,))
             deleted = cur.rowcount
         if deleted == 0:
-            logger.warning(
-                "PG-queue: delete(msg_id=%s) removed no row — message was already "
-                "gone (vt likely expired during processing; possible re-delivery).",
+            logger.debug(
+                "PG-queue: delete(msg_id=%s) removed no row — already gone "
+                "(vt likely expired; consumer logs the contextual warning).",
                 msg_id,
             )
         return deleted == 1
--- a/workers/queue_backend/pg_queue/client.py
+++ b/workers/queue_backend/pg_queue/client.py
@@ -68,7 +68,7 @@ UPDATE pg_queue_message
         FOR UPDATE SKIP LOCKED
       LIMIT %s
  )
-RETURNING msg_id, message
+RETURNING msg_id, message, read_ct
 """
 
 
@@ -79,11 +79,13 @@ class QueueMessage:
     ``message`` is the already-decoded JSONB payload (psycopg2 parses
     ``jsonb`` to a Python ``dict``). ``frozen`` freezes the binding only —
     the ``dict`` itself is mutable, so treat the payload as read-only by
-    convention.
+    convention. ``read_ct`` is the post-claim delivery count (1 on the
+    first claim), so consumers can cap redelivery of a poison message.
     """
 
     msg_id: int
     message: dict[str, Any]
+    read_ct: int = 0
 
 
 class PgQueueClient:
@@ -180,7 +182,9 @@ class PgQueueClient:
         with self._cursor() as cur:
             cur.execute(_DEQUEUE_SQL, (vt_seconds, queue_name, qty))
             rows = cur.fetchall()
-        return [QueueMessage(msg_id=int(r[0]), message=r[1]) for r in rows]
+        return [
+            QueueMessage(msg_id=int(r[0]), message=r[1], read_ct=int(r[2])) for r in rows
+        ]
 
     def delete(self, msg_id: int) -> bool:
         """Ack a processed message. Returns ``True`` if a row was removed.
--- a/workers/queue_backend/pg_queue/consumer.py
+++ b/workers/queue_backend/pg_queue/consumer.py
@@ -1,0 +1,157 @@
+"""PG queue consumer — claims tasks from ``pg_queue_message`` and runs them.
+
+The producer side (9b) enqueues a :class:`~queue_backend.pg_queue.TaskPayload`
+when a task is routed to PG. This is the other half: it polls the queue with
+``SKIP LOCKED`` + a visibility timeout (via :class:`PgQueueClient`), runs each
+claimed task **in-process** (no Celery broker), and acks by deleting the row.
+
+A task that fails — or a crash before ack — is redelivered once its ``vt``
+expires (at-least-once; tasks must be idempotent). A task name not in the
+registry can never run, so it is dropped (with a loud log) rather than left to
+redeliver forever as a poison message.
+
+Run as ``python -m queue_backend.pg_queue.consumer`` (config via env). The
+worker bootstrap must have imported/registered the Celery tasks so they
+resolve in ``current_app.tasks``.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+import signal
+import time
+from typing import TYPE_CHECKING
+
+from celery import current_app
+
+from .client import PgQueueClient
+
+if TYPE_CHECKING:
+    from celery import Celery
+
+    from .client import QueueMessage
+
+logger = logging.getLogger(__name__)
+
+_DEFAULT_QUEUE = "default"
+_DEFAULT_BATCH = 10
+_DEFAULT_VT_SECONDS = 30
+_DEFAULT_POLL_INTERVAL = 0.1
+_DEFAULT_BACKOFF_MAX = 2.0
+
+
+class PgQueueConsumer:
+    """Polls one PG queue, runs each claimed task in-process, acks on success."""
+
+    def __init__(
+        self,
+        queue_name: str,
+        *,
+        client: PgQueueClient | None = None,
+        app: Celery | None = None,
+        batch_size: int = _DEFAULT_BATCH,
+        vt_seconds: int = _DEFAULT_VT_SECONDS,
+        poll_interval: float = _DEFAULT_POLL_INTERVAL,
+        backoff_max: float = _DEFAULT_BACKOFF_MAX,
+    ) -> None:
+        self.queue_name = queue_name
+        self._client = client if client is not None else PgQueueClient()
+        self._app = app if app is not None else current_app
+        self.batch_size = batch_size
+        self.vt_seconds = vt_seconds
+        self.poll_interval = poll_interval
+        self.backoff_max = backoff_max
+        self._running = False
+
+    def poll_once(self) -> int:
+        """Claim + process one batch; returns the number of messages claimed."""
+        messages = self._client.read(
+            self.queue_name, vt_seconds=self.vt_seconds, qty=self.batch_size
+        )
+        for message in messages:
+            self._handle(message)
+        return len(messages)
+
+    def _handle(self, message: QueueMessage) -> None:
+        payload = message.message
+        task_name = payload.get("task_name")
+        task = self._app.tasks.get(task_name)
+        if task is None:
+            # Can never run → drop (and shout) rather than redeliver forever.
+            logger.error(
+                "PG-queue consumer: unknown task %r (msg_id=%s) — dropping",
+                task_name,
+                message.msg_id,
+            )
+            self._client.delete(message.msg_id)
+            return
+        try:
+            # Run the task body in-process (eager), propagating failures.
+            task.apply(
+                args=payload.get("args") or [],
+                kwargs=payload.get("kwargs") or {},
+                throw=True,
+            )
+        except Exception:
+            # Leave the row: its vt expires and it is redelivered.
+            logger.exception(
+                "PG-queue consumer: task %r (msg_id=%s) failed — leaving for "
+                "vt-expiry redelivery",
+                task_name,
+                message.msg_id,
+            )
+            return
+        self._client.delete(message.msg_id)  # ack
+
+    def run(self, *, install_signals: bool = True) -> None:
+        """Poll loop with empty-queue backoff and graceful shutdown."""
+        self._running = True
+        if install_signals:
+            self._install_signal_handlers()
+        logger.info(
+            "PG-queue consumer started (queue=%r, batch=%s, vt=%ss)",
+            self.queue_name,
+            self.batch_size,
+            self.vt_seconds,
+        )
+        backoff = self.poll_interval
+        while self._running:
+            if self.poll_once():
+                backoff = self.poll_interval
+            else:
+                time.sleep(backoff)
+                backoff = min(backoff * 2, self.backoff_max)
+        logger.info("PG-queue consumer stopped (queue=%r)", self.queue_name)
+
+    def stop(self, *_: object) -> None:
+        """Request a graceful stop after the current batch."""
+        self._running = False
+
+    def _install_signal_handlers(self) -> None:
+        # signal.signal only works in the main thread.
+        try:
+            signal.signal(signal.SIGTERM, self.stop)
+            signal.signal(signal.SIGINT, self.stop)
+        except ValueError:
+            logger.debug(
+                "PG-queue consumer: signal handlers not installed (non-main thread)"
+            )
+
+
+def main() -> None:
+    logging.basicConfig(level=os.getenv("LOG_LEVEL", "INFO"))
+    PgQueueConsumer(
+        queue_name=os.getenv("WORKER_PG_QUEUE_CONSUMER_QUEUE", _DEFAULT_QUEUE),
+        batch_size=int(os.getenv("WORKER_PG_QUEUE_CONSUMER_BATCH", _DEFAULT_BATCH)),
+        vt_seconds=int(
+            os.getenv("WORKER_PG_QUEUE_CONSUMER_VT_SECONDS", _DEFAULT_VT_SECONDS)
+        ),
+        poll_interval=float(
+            os.getenv("WORKER_PG_QUEUE_CONSUMER_POLL_INTERVAL", _DEFAULT_POLL_INTERVAL)
+        ),
+    ).run()
+
+
+if __name__ == "__main__":
+    main()
--- a/workers/queue_backend/pg_queue/consumer.py
+++ b/workers/queue_backend/pg_queue/consumer.py
@@ -6,9 +6,12 @@ when a task is routed to PG. This is the other half: it polls the queue with
 claimed task **in-process** (no Celery broker), and acks by deleting the row.
 
 A task that fails — or a crash before ack — is redelivered once its ``vt``
-expires (at-least-once; tasks must be idempotent). A task name not in the
-registry can never run, so it is dropped (with a loud log) rather than left to
-redeliver forever as a poison message.
+expires (at-least-once; tasks must be idempotent), bounded by ``max_attempts``
+(``read_ct``): a task that keeps failing past the cap is dropped as a poison
+message (logged with its payload) rather than redelivered forever. A message
+with no ``task_name`` (malformed/foreign) or a name not in the registry can
+never run, so it is likewise dropped with a loud log. The fairness header is
+rebuilt from the payload so a PG-routed run mirrors the Celery dispatch path.
 
 Run as ``python -m queue_backend.pg_queue.consumer`` (config via env). The
 worker bootstrap must have imported/registered the Celery tasks so they
@@ -25,6 +28,7 @@ from typing import TYPE_CHECKING
 
 from celery import current_app
 
+from ..fairness import FAIRNESS_HEADER_NAME
 from .client import PgQueueClient
 
 if TYPE_CHECKING:
@@ -39,6 +43,9 @@ _DEFAULT_BATCH = 10
 _DEFAULT_VT_SECONDS = 30
 _DEFAULT_POLL_INTERVAL = 0.1
 _DEFAULT_BACKOFF_MAX = 2.0
+# A task claimed more than this many times keeps failing — drop it (poison)
+# rather than redeliver forever.
+_DEFAULT_MAX_ATTEMPTS = 5
 
 
 class PgQueueConsumer:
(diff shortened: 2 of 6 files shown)


workers/queue_backend/pg_queue/consumer.py
```diff
@@ -0,0 +1,157 @@
...
+            logger.error(
+                "PG-queue consumer: unknown task %r (msg_id=%s) — dropping",
+                task_name,
+                message.msg_id,
+            )
+            self._client.delete(message.msg_id)
+            return
+        try:
+            # Run the task body in-process (eager), propagating failures.
+            task.apply(
```
muhammad-ali-e, line 155: **[Important] The `fairness` (and `queue`) payload field is silently dropped — breaks the documented producer↔consumer contract.**

`task_payload.py` states *"the consumer rebuilds the `x-fairness-key` header from it, mirroring the Celery dispatch path"*, and the producer (`dispatch._enqueue_pg`) serialises both `fairness` and `queue` into every PG payload. But `_handle` calls `task.apply(args=..., kwargs=..., throw=True)` with no `headers=` and never reads `payload["fairness"]`/`payload["queue"]`. A fairness-aware task run via PG therefore executes **without** the `x-fairness-key` header the Celery path guarantees — a real behavioural divergence, not cosmetic. Flagged independently by 5 of 6 review passes.

**Fix (option A — honour the contract):**
```python
fairness = payload.get("fairness")
headers = {FAIRNESS_HEADER_NAME: fairness} if fairness else None
task.apply(
    args=payload.get("args") or [],
    kwargs=payload.get("kwargs") or {},
    headers=headers,
    throw=True,
)
```
**Option B (defer):** if fairness on the PG path is intentionally out of scope for 9c, add a `# TODO` here and correct the `task_payload.py` docstring so it stops claiming behaviour that doesn't exist. Either way, add a test that pins the chosen behaviour.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -0,0 +1,157 @@
...
+                args=payload.get("args") or [],
+                kwargs=payload.get("kwargs") or {},
+                throw=True,
+            )
+        except Exception:
```
muhammad-ali-e, line 161: **[Important] No max-attempts / `read_ct` guard — a deterministically-failing task redelivers forever (poison message).**

Leaving the row for vt-expiry redelivery is correct for *transient* failures, but a task that fails every time (bad payload, code bug, permanently-missing dep) will be re-claimed and re-fail on every `vt` expiry — ~2,880 identical stack traces/day per poison row at the default 30s vt, pinning a worker and flooding logs. The module docstring frames poison-message handling as "drop unknown tasks", but a *known* task that always raises is the more common poison case and has no guard.

The infrastructure already exists: `client.py` increments `read_ct` on every claim, but the dequeue SQL only returns `msg_id, message`, so `QueueMessage` can't see it. Suggest surfacing `read_ct` and dead-lettering (or dropping with a loud ERROR carrying `msg_id`/`task_name`/`read_ct`) past a threshold. Note `test_failed_task_is_not_acked` currently locks in the unbounded-redelivery behaviour, so fixing this is a test change too.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -0,0 +1,157 @@
...
+        messages = self._client.read(
+            self.queue_name, vt_seconds=self.vt_seconds, qty=self.batch_size
+        )
+        for message in messages:
+            self._handle(message)
+        return len(messages)
+
+    def _handle(self, message: QueueMessage) -> None:
+        payload = message.message
+        task_name = payload.get("task_name")
```
muhammad-ali-e, line 109: **[Important] A missing/blank `task_name` is destroyed as "unknown task None" — silent data loss with a misleading log.**

`payload.get("task_name")` returns `None` for a malformed/truncated/foreign payload; `self._app.tasks.get(None)` is `None`, so the row takes the unknown-task branch and is **permanently deleted** with a log that says `unknown task None` — misdirecting the operator toward task *registration* rather than a malformed payload. The message and its `args`/`kwargs` are gone with no dead-letter capture.

Distinguish the two failure modes:
```python
if not task_name:
    logger.error(
        "PG-queue consumer: payload missing task_name (msg_id=%s) — "
        "dropping malformed message: %r", message.msg_id, payload,
    )
    self._client.delete(message.msg_id)
    return
```
and consider dead-lettering rather than hard-deleting so the payload is recoverable. (Also: binding `payload` to the `TaskPayload` TypedDict and using keyed access would let the type checker catch producer↔consumer drift instead of leaving it to prose.)

muhammad-ali-e (commented):
## Automated review — PG Queue Phase 9c consumer

Ran six specialised review passes (code review, silent-failure hunt, type design, test coverage, comment accuracy, simplification) over the 5-file diff. The consumer is clean, well-documented, and the conftest dedup is correct. Findings below are posted inline.

**Headline:** five of six passes independently flagged the same issue — the consumer silently drops the `fairness` (and `queue`) payload field that the producer serialises and that `task_payload.py` documents the consumer as rebuilding into the `x-fairness-key` header. That is the one I'd resolve (implement it, or correct the contract docstring) before merge. Everything else is Important/Suggestion-grade.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -0,0 +1,157 @@
...
+        except Exception:
+            # Leave the row: its vt expires and it is redelivered.
+            logger.exception(
+                "PG-queue consumer: task %r (msg_id=%s) failed — leaving for "
+                "vt-expiry redelivery",
+                task_name,
+                message.msg_id,
+            )
+            return
+        self._client.delete(message.msg_id)  # ack
```
muhammad-ali-e: **[Suggestion] `delete()`'s return value is discarded on the success path — the lost-ack / double-execution signal is dropped.**

`client.delete()` returns `False` (and logs WARNING) when the row was already gone — i.e. the task ran longer than `vt_seconds`, the row became re-visible, and another worker re-claimed it. That's the exact lost-ack case the vt design exists to surface, and the consumer is the layer that knows the *task name and outcome*. The client's WARNING only carries a `msg_id`.

```python
if not self._client.delete(message.msg_id):  # ack
    logger.error(
        "PG-queue consumer: task %r (msg_id=%s) completed but its row was "
        "already gone — vt=%ss likely too short; task was double-delivered.",
        task_name, message.msg_id, self.vt_seconds,
    )
```
This makes the actionable fix (raise `vt_seconds`) reachable. Not currently covered by any test.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -0,0 +1,157 @@
...
+        if install_signals:
+            self._install_signal_handlers()
+        logger.info(
+            "PG-queue consumer started (queue=%r, batch=%s, vt=%ss)",
+            self.queue_name,
+            self.batch_size,
+            self.vt_seconds,
+        )
+        backoff = self.poll_interval
+        while self._running:
```
muhammad-ali-e, line 193: **[Suggestion] `run()` has no guard around `poll_once()` — a single transient `read()` error tears down the whole loop.**

`PgQueueClient` is built to *recover* its connection after a blip (it discards a dead handle so the next call reconnects), but because `run()` doesn't wrap `poll_once()`, an exception from `read()` propagates out of `run()` and kills the consumer — so every transient DB hiccup becomes a full process restart rather than a logged-and-retried blip. If "let the orchestrator restart it" is the deliberate design, say so in `run()`'s docstring; otherwise wrap `poll_once()` in a narrow handler that logs at ERROR and applies the existing backoff.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -0,0 +1,157 @@
...
+        """Request a graceful stop after the current batch."""
+        self._running = False
+
+    def _install_signal_handlers(self) -> None:
+        # signal.signal only works in the main thread.
+        try:
+            signal.signal(signal.SIGTERM, self.stop)
+            signal.signal(signal.SIGINT, self.stop)
+        except ValueError:
+            logger.debug(
```
muhammad-ali-e: **[Suggestion] Non-main-thread signal-install failure is logged at DEBUG — invisible at normal prod log levels.**

Catching `ValueError` here is correctly scoped (good — `signal.signal` raises exactly that off the main thread). But if a consumer is ever launched off-main-thread in production, `SIGTERM`/`SIGINT` won't trigger graceful shutdown and the loop gets hard-killed mid-batch — and at DEBUG nobody will know graceful-shutdown-via-signals is inactive. Bump to INFO/WARNING. Also consider expanding the line-132 comment to name the consequence ("caller must drive stop()"), and note this `ValueError` branch has no test.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -0,0 +1,157 @@
...
+                "PG-queue consumer: signal handlers not installed (non-main thread)"
+            )
+
+
+def main() -> None:
+    logging.basicConfig(level=os.getenv("LOG_LEVEL", "INFO"))
+    PgQueueConsumer(
```
muhammad-ali-e, line 232: **[Suggestion] `main()` env handling: `backoff_max` not tunable, no parse-error handling, repeated prefix.**

Three small things in one place:
1. `backoff_max` is a constructor param but `main()` never wires `WORKER_PG_QUEUE_CONSUMER_BACKOFF_MAX`, so the 2.0s ceiling isn't operator-tunable while the other four knobs are — silent asymmetry. Add it for symmetry, or drop the unused configurability.
2. `int(os.getenv(...))` / `float(os.getenv(...))` raise a bare `ValueError: invalid literal...` on a bad value, before the "consumer started" log — in a container this looks like an instant crash-loop with a cryptic trace. Wrap and raise a named error identifying the offending variable.
3. The `WORKER_PG_QUEUE_CONSUMER_` prefix is repeated four times; a tiny local `_env(suffix, default)` helper removes the repetition and makes the suffix list scannable.

None of this is covered by a test — a `main()` test (monkeypatch env, patch `PgQueueConsumer`, assert parsed kwargs) would lock the parsing and the intentional `backoff_max` omission.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -0,0 +1,157 @@
...
+_DEFAULT_BATCH = 10
+_DEFAULT_VT_SECONDS = 30
+_DEFAULT_POLL_INTERVAL = 0.1
+_DEFAULT_BACKOFF_MAX = 2.0
+
+
+class PgQueueConsumer:
+    """Polls one PG queue, runs each claimed task in-process, acks on success."""
+
+    def __init__(
```
muhammad-ali-e, line 58: **[Suggestion] No construction-time validation; tuning params are public and mutable.**

`batch_size`, `vt_seconds`, `poll_interval`, `backoff_max` are accepted unchecked and stored as public attributes. Invalid `vt_seconds`/`qty` only raise once `read()` runs *inside the loop* (so a misconfigured consumer fails on its first poll, batch after batch, rather than at `__init__`); `poll_interval <= 0` and `backoff_max < poll_interval` (which makes the `min(backoff*2, backoff_max)` cap meaningless) are never checked at all. Nothing stops `consumer.vt_seconds = 0` mid-loop.

Validate in `__init__` (fail fast: all positive, `backoff_max >= poll_interval`); if these are meant to be immutable post-construction, make them private or expose read-only properties.

workers/tests/test_pg_queue_consumer.py
```diff
@@ -0,0 +1,122 @@
...
+        assert "unknown task" in caplog.text
+
+    def test_empty_batch_acks_nothing(self):
+        client = MagicMock()
+        client.read.return_value = []
+        assert PgQueueConsumer("q", client=client).poll_once() == 0
+        client.delete.assert_not_called()
+
+
+class TestRunLoop:
```
muhammad-ali-e, line 164: **[Test coverage] Good foundation (real `@shared_task` bodies via `apply()`, ack/no-ack/unknown-drop, gated integration test). Notable gaps:**

- **`run()` backoff growth/reset (consumer.py:118-124):** `TestRunLoop` stops on the *first* sleep, so the loop body runs once — the exponential growth, the `backoff_max` cap, and the reset-to-`poll_interval`-when-a-batch-is-found branch never execute. Drive `read` with a scripted `[[], [], [msg], []]` and assert captured sleep durations.
- **Multi-message batch (consumer.py:72-74):** every `poll_once` test reads a single-element batch. Add a `[ok, boom, unknown, ok]` batch and assert ok-tasks run in order, ok+unknown msg_ids are acked, boom's is not, and `poll_once` returns the full length.
- **`fairness` propagation:** no test sets `fairness` — this is what let the dropped-header bug (see consumer.py:91) ship undetected.
- **`main()` env parsing (consumer.py:142-153):** untested.
- **Signal `ValueError` path (consumer.py:131-139):** `install_signals=False` everywhere, so `_install_signal_handlers` never runs.
- **Nit:** `test_failed_task_is_not_acked` doesn't assert `logger.exception` fired — pin it with `caplog` like the unknown-task test does, since that log is the only signal a poison message is cycling.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in `902bb42dc` (Option A — honour the contract). `_handle` now rebuilds the header: `headers = {FAIRNESS_HEADER_NAME: fairness} if fairness else None` and passes it to `task.apply(...)`, so a PG-routed run carries `x-fairness-key` like the Celery path. (`queue` isn't passed — `apply()` runs in-process, there's no broker queue to route to; it stays in the payload for the orchestrator/observability.) Pinned by `test_fairness_header_rebuilt_for_run`.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in `902bb42dc`. Surfaced `read_ct` (dequeue `RETURNING msg_id, message, read_ct` → `QueueMessage.read_ct`) and added a poison guard: a task claimed past `max_attempts` (default 5, env `WORKER_PG_QUEUE_CONSUMER_MAX_ATTEMPTS`) is dropped with a loud ERROR carrying `msg_id`/`task_name`/`read_ct`/payload (recoverable from logs), instead of redelivering forever. `test_failed_task_is_not_acked` still pins single-failure redelivery (read_ct=1 < cap); `test_poison_message_dropped_past_max_attempts` pins the cap. A proper dead-letter table is a follow-up (noted out-of-scope).

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in `902bb42dc`. A missing/blank `task_name` now takes a dedicated branch: `logger.error("payload missing task_name (msg_id=%s) — dropping malformed message: %r", ...)` + drop — distinct from the "unknown task" path, so the operator isn't misdirected toward task registration. Pinned by `test_missing_task_name_dropped_as_malformed`. (The strict `TaskPayload` keyed-access binding is a nice follow-up; left `.get()` for now to avoid TypedDict-vs-dict checker friction.)

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Done in `902bb42dc`. `_handle` now checks the ack return: `if not self._client.delete(msg_id): logger.warning("ack found no row for task %r (msg_id=%s) — it likely exceeded vt and was re-claimed (possible double-run)", ...)`. Pinned by `test_ack_finding_no_row_warns`.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in `902bb42dc`. `run()` now wraps `poll_once()` in try/except: a transient read/DB error is logged (`logger.exception`) and the loop backs off and continues, rather than propagating out and killing the consumer — the client self-recovers its connection on the next call. Pinned by `test_poll_error_does_not_kill_loop` (read raises once, loop keeps polling).

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Done in `902bb42dc` — bumped the non-main-thread signal-install failure to `logger.warning(... "SIGTERM/SIGINT will not trigger graceful shutdown")` so it's visible at normal prod levels.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Done in `902bb42dc`. `main()` now uses a small `_env(suffix, default, cast)` helper (drops the repeated `WORKER_PG_QUEUE_CONSUMER_` prefix) and wires both `BACKOFF_MAX` and `MAX_ATTEMPTS`, so all knobs are operator-tunable. Left parse errors to raise loudly at startup (a typo'd numeric env failing fast at boot is preferable to silently coercing) — happy to add a friendlier message if you'd rather.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Done in `902bb42dc`. `__init__` validates `batch_size` / `vt_seconds` / `poll_interval` / `backoff_max` / `max_attempts` are all positive and raises `ValueError` at construction, so a misconfigured consumer fails immediately rather than batch-after-batch on the first poll. Pinned by `TestConstruction.test_rejects_non_positive_params`.

workers/tests/test_pg_queue_consumer.py
muhammad-ali-e: Added in `902bb42dc`:
- **backoff growth/reset** — `test_backoff_grows_then_resets` drives `read` with `[[], [], [msg], []]` and asserts captured sleeps `[0.1, 0.2, 0.1]` (double, then reset on a batch).
- **multi-message batch** — `test_multi_message_batch` runs `[ok, boom, unknown, ok]`, asserts ok bodies ran in order, ok+unknown msg_ids acked, boom's not, and `poll_once` returns 4.
- **fairness propagation** — `test_fairness_header_rebuilt_for_run` (the gap that let the dropped-header bug ship).
- **signal/`run` resilience** — `test_poll_error_does_not_kill_loop`.
- **failed-task log** — `test_failed_task_is_not_acked_and_logs` now asserts the failure is logged (the poison-cycling signal).
`main()` env parsing is exercised via the `_env` helper; a dedicated `main()` test can follow if you want it pinned.

muhammad-ali-e:
### Review feedback addressed — `902bb42dc`

All 9 comments (3 Important):

**Important — correctness/resilience:**
- **Fairness header** now rebuilt from the payload on the PG run path (was silently dropped) → mirrors the Celery dispatch contract.
- **Poison guard:** surfaced `read_ct` (dequeue `RETURNING` + `QueueMessage`) and drop a task that keeps failing past `max_attempts` (default 5, env-tunable) with a loud ERROR + payload, instead of redelivering forever.
- **Malformed payload** (missing `task_name`) → distinct "missing task_name" drop log, not a misleading "unknown task None".

**Suggestions:**
- `run()` wraps `poll_once()` — a transient blip backs off + continues instead of killing the loop.
- `delete()` returning `False` (vt-exceeded → re-claim) now WARNs (possible double-run).
- `__init__` validates positive tuning params (fail at construction); `main()` wires `backoff_max` + `max_attempts` via env (prefix helper); off-main-thread signal failure → WARNING.

**Tests:** poison drop, missing task_name, fairness propagation, multi-message batch (order + selective ack), ack-no-row warn, construction validation, backoff growth/reset, poll-error resilience. Consumer suite now 14 tests.

57 queue_backend tests green; ruff + pre-commit clean; SonarCloud gate already passing.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -0,0 +1,232 @@
...
+        self._running = False
+
+    def poll_once(self) -> int:
+        """Claim + process one batch; returns the number of messages claimed."""
+        messages = self._client.read(
+            self.queue_name, vt_seconds=self.vt_seconds, qty=self.batch_size
+        )
+        for message in messages:
+            self._handle(message)
+        return len(messages)
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in `96ec32656`. Defaulted `batch_size` to **1** so each message gets its own visibility window — eliminating the shared-vt / tail-double-run case for the common config. Batching is now opt-in, and the constant's doc states the constraint: keep `vt_seconds > batch_size × worst-case task duration` if you raise it. (A per-message vt re-extension is a future option if we want safe large batches.)

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -0,0 +1,232 @@
...
+            )
+            return
+
+        if not self._client.delete(message.msg_id):  # ack
+            logger.warning(
+                "PG-queue consumer: ack found no row for task %r (msg_id=%s) — "
+                "it likely exceeded vt and was re-claimed (possible double-run)",
+                task_name,
+                message.msg_id,
+            )
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in `96ec32656` — downgraded `client.delete()`'s no-row log to DEBUG and kept the consumer-level WARNING (it names the task, the more useful signal). One WARNING per double-run now, not two.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -0,0 +1,232 @@
...
+        # rather than batch-after-batch once the loop starts.
+        for name, value in (
+            ("batch_size", batch_size),
+            ("vt_seconds", vt_seconds),
+            ("poll_interval", poll_interval),
+            ("backoff_max", backoff_max),
+            ("max_attempts", max_attempts),
+        ):
+            if value <= 0:
+                raise ValueError(f"{name} must be positive, got {value!r}")
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in `96ec32656` — `__init__` now also raises if `backoff_max < poll_interval` (with the reason: otherwise `min(poll_interval*2, backoff_max)` shrinks the backoff below `poll_interval`, so it decreases instead of grows). Pinned by `test_rejects_backoff_max_below_poll_interval`.

workers/queue_backend/pg_queue/client.py
```diff
@@ -79,11 +79,13 @@ class QueueMessage:
...
     ``jsonb`` to a Python ``dict``). ``frozen`` freezes the binding only —
     the ``dict`` itself is mutable, so treat the payload as read-only by
-    convention.
+    convention. ``read_ct`` is the post-claim delivery count (1 on the
+    first claim), so consumers can cap redelivery of a poison message.
     """
 
     msg_id: int
     message: dict[str, Any]
+    read_ct: int = 0
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in `96ec32656` — removed the `read_ct = 0` default. `read_ct` is now required, so a `QueueMessage` can't be built in a "never claimed" state that silently bypasses the poison guard (`0 > max_attempts` is always False). The dequeue always returns `>= 1`, and all callers (real + tests) already supply it.

PR #2047, opened 2026-06-12
UN-3541 [FEAT] Wire PG-queue consumer into run-worker.sh + bootstrap guard

muhammad-ali-e:
Makes the 9c PG-queue consumer (UN-3539, #2045) runnable through the normal worker flow — and safe to launch. Split out so #2045 stays focused on consumer logic.

Targets the long-lived `feat/UN-3445-pg-queue-integration` branch (not `main`).

## Why
The consumer is task-name routed: it reads a `task_name` from each PG message and runs it from the worker app's task registry. So the launch path must (a) bootstrap the **right** task set and (b) fail loudly if it didn't — otherwise every message is silently dropped as "unknown task".

## What

**Bootstrapping launcher** — `workers/pg_queue_consumer/__main__.py`
- Sets `WORKER_TYPE` to the source worker (default `notification`) **before** `import worker`. The root `worker.py` loads exactly one worker type's `tasks.py`; a bare import would load the `general` worker's tasks and drop every notification. Overridable via `WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE`.

**run-worker.sh wiring**
- New `pg-queue-consumer` worker type → `uv run python -m pg_queue_consumer` (a plain Postgres poll loop, not a Celery/RabbitMQ worker), run from the workers root; queue routed via `WORKER_PG_QUEUE_CONSUMER_QUEUE`.
- **Opt-in**: not part of `./run-worker.sh all` (default behaviour unchanged — the new arch is experimental/flag-gated). Status shows it only when running, so a deliberate non-start isn't reported as `STOPPED`.

**Startup guard** — `PgQueueConsumer.run(require_tasks=True)`
- Refuses to start on an empty task registry (a strong signal the worker app wasn't bootstrapped). Fails loud instead of silently destroying data.

**Health-port fix**
- Dropped the hard-coded `8086` — the consumer runs no health server, so it binds no port (avoids a collision). A liveness endpoint, if added later, reads an env port.

### Integration fixes found during live dev-test
- **Log path** — detached workers now write to an absolute `$worker_dir/$type.log`, so `-L`/`-C` find it (also fixes the same latent bug for pluggable workers, which likewise run from the workers root).
- **PID discovery** — `get_worker_pids` matches the `python -m` invocation, so `--status` / `-k` / `-r` work for the consumer.

## Testing
- **Live end-to-end**: `./run-worker.sh pg-queue-consumer` drained real `send_webhook_notification` messages from the PG queue → Slack **HTTP 200**, tasks succeeded, rows deleted.
- **Unit**: new test asserts the startup guard refuses an empty-registry app; full consumer suite green.
- Verified `--status` (RUNNING/hidden-when-stopped), single-worker `-r`, and log resolution.

## Out of scope (rollout)
- Real liveness/health endpoint + K8s Deployment for the consumer; per-org canary; runbook.

### DIFF
--- a/workers/pg_queue_consumer/__main__.py
+++ b/workers/pg_queue_consumer/__main__.py
@@ -1,0 +1,33 @@
+"""Entry point: bootstrap a worker app's tasks, then run the PG-queue consumer.
+
+The root ``worker`` module loads exactly ONE worker type's tasks, chosen by the
+``WORKER_TYPE`` env (``worker.py`` → ``WorkerBuilder.build_celery_app`` →
+imports that type's ``tasks.py``). The PG consumer drains a specific source
+worker's queue, so it must register THAT worker's tasks: we set ``WORKER_TYPE``
+from ``WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE`` (default ``notification`` — the
+first migrated leaf task) BEFORE importing ``worker``.
+
+This override is required: ``run-worker.sh`` exports its own ``WORKER_TYPE``
+(the consumer pseudo-type ``pg_queue_consumer``, which owns no tasks), so a
+plain import would fall back to the ``general`` worker's tasks and every
+notification message would be dropped as an unknown task. The consumer's
+startup guard only catches an *empty* registry, not a *wrong* one — so the
+right worker type must be selected here.
+
+Launch via ``python -m pg_queue_consumer`` or ``./run-worker.sh pg-queue-consumer``.
+"""
+
+import os
+
+# Select the source worker whose tasks back this consumer's queue. Must run
+# BEFORE `import worker`, which reads WORKER_TYPE at import time. We overwrite
+# (not setdefault) because the launcher's own WORKER_TYPE owns no tasks.
+os.environ["WORKER_TYPE"] = os.environ.get(
+    "WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE", "notification"
+)
+
+import worker  # noqa: E402,F401 — side-effect: registers the source worker's tasks
+from queue_backend.pg_queue.consumer import main  # noqa: E402
+
+if __name__ == "__main__":
+    main()
--- a/workers/pg_queue_consumer/__main__.py
+++ b/workers/pg_queue_consumer/__main__.py
@@ -1,18 +1,21 @@
 """Entry point: bootstrap a worker app's tasks, then run the PG-queue consumer.
 
 The root ``worker`` module loads exactly ONE worker type's tasks, chosen by the
-``WORKER_TYPE`` env (``worker.py`` → ``WorkerBuilder.build_celery_app`` →
-imports that type's ``tasks.py``). The PG consumer drains a specific source
+``WORKER_TYPE`` env: ``worker.py`` builds the Celery app via
+``WorkerBuilder.build_celery_app`` (which configures, but does NOT import
+tasks), then — in separate module-level logic in ``worker.py`` — ``importlib``-
+loads that type's ``tasks.py``. The PG consumer drains a specific source
 worker's queue, so it must register THAT worker's tasks: we set ``WORKER_TYPE``
 from ``WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE`` (default ``notification`` — the
-first migrated leaf task) BEFORE importing ``worker``.
+worker that owns the first migrated leaf task, ``send_webhook_notification``)
+BEFORE importing ``worker``.
 
 This override is required: ``run-worker.sh`` exports its own ``WORKER_TYPE``
 (the consumer pseudo-type ``pg_queue_consumer``, which owns no tasks), so a
 plain import would fall back to the ``general`` worker's tasks and every
-notification message would be dropped as an unknown task. The consumer's
-startup guard only catches an *empty* registry, not a *wrong* one — so the
-right worker type must be selected here.
+message for the intended worker would be dropped as an unknown task. The
+consumer's startup guard only catches an *empty* registry, not a *wrong* one —
+so the right worker type must be selected here.
 
 Launch via ``python -m pg_queue_consumer`` or ``./run-worker.sh pg-queue-consumer``.
 """
--- a/workers/pg_queue_consumer/__main__.py
+++ b/workers/pg_queue_consumer/__main__.py
@@ -22,15 +22,25 @@ Launch via ``python -m pg_queue_consumer`` or ``./run-worker.sh pg-queue-consume
 
 import os
 
-# Select the source worker whose tasks back this consumer's queue. Must run
-# BEFORE `import worker`, which reads WORKER_TYPE at import time. We overwrite
-# (not setdefault) because the launcher's own WORKER_TYPE owns no tasks.
-os.environ["WORKER_TYPE"] = os.environ.get(
-    "WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE", "notification"
-)
 
-import worker  # noqa: E402,F401 — side-effect: registers the source worker's tasks
-from queue_backend.pg_queue.consumer import main  # noqa: E402
+def _bootstrap_and_run() -> None:
+    # Select the source worker whose tasks back this consumer's queue. Must run
+    # BEFORE `import worker`, which reads WORKER_TYPE at import time. We overwrite
+    # (not setdefault) because the launcher's own WORKER_TYPE owns no tasks.
+    #
+    # Kept inside this guarded function (not at module scope) so the env mutation
+    # and the heavy `import worker` bootstrap only happen on the `python -m`
+    # entry — never as a side-effect if this module is imported by a test, IDE,
+    # or type-checker walking `__main__` files.
+    os.environ["WORKER_TYPE"] = os.environ.get(
+        "WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE", "notification"
+    )
+
+    import worker  # noqa: F401 — side-effect: registers the source worker's tasks
+    from queue_backend.pg_queue.consumer import main
+
+    main()
+
 
 if __name__ == "__main__":
-    main()
+    _bootstrap_and_run()
--- a/workers/queue_backend/pg_queue/consumer.py
+++ b/workers/queue_backend/pg_queue/consumer.py
@@ -24,7 +24,8 @@ import logging
 import os
 import signal
 import time
-from typing import TYPE_CHECKING
+from collections.abc import Callable
+from typing import TYPE_CHECKING, TypeVar
 
 from celery import current_app
 
@@ -37,6 +38,8 @@ if TYPE_CHECKING:
     from .client import QueueMessage
 
 logger = logging.getLogger(__name__)
+
+_T = TypeVar("_T")
 
 _DEFAULT_QUEUE = "default"
 # Default 1: the whole batch shares one vt window (set atomically at claim),
@@ -202,11 +205,22 @@ class PgQueueConsumer:
         self._running = True
         if install_signals:
             self._install_signal_handlers()
+        # Log the registered application tasks at startup. The guard above only
+        # catches an *empty* registry; a *wrong* one (e.g. the launcher selected
+        # the wrong source worker type) is non-empty but missing the target
+        # task, so each message would be dropped as "unknown task". Surfacing
+        # the registry here makes a wrong-type boot diagnosable from one line.
+        app_tasks = sorted(
+            name for name in self._app.tasks if not name.startswith("celery.")
+        )
         logger.info(
-            "PG-queue consumer started (queue=%r, batch=%s, vt=%ss)",
+            "PG-queue consumer started (queue=%r, batch=%s, vt=%ss) — "
+            "%d application task(s) registered: %s",
             self.queue_name,
             self.batch_size,
             self.vt_seconds,
+            len(app_tasks),
+            ", ".join(app_tasks) or "(none)",
         )
         backoff = self.poll_interval
         while self._running:
@@ -245,8 +259,18 @@ class PgQueueConsumer:
 def main() -> None:
     logging.basicConfig(level=os.getenv("LOG_LEVEL", "INFO"))
 
-    def _env(suffix: str, default: object, cast: type) -> object:
-        return cast(os.getenv(f"WORKER_PG_QUEUE_CONSUMER_{suffix}", default))
+    def _env(suffix: str, default: _T, cast: Callable[[str], _T]) -> _T:
+        # Preserve the default's type through to PgQueueConsumer's typed
+        # __init__ (a bare `type` would erase it). On a bad value, fail with the
+        # offending var name instead of a context-free `int('abc')` ValueError.
+        var = f"WORKER_PG_QUEUE_CONSUMER_{suffix}"
+        raw = os.getenv(var)
+        if raw is None:
+            return default
+        try:
+            return cast(raw)
+        except (ValueError, TypeError) as exc:
+            raise ValueError(f"Invalid {var}={raw!r}: {exc}") from exc
 
     PgQueueConsumer(
         queue_name=_env("QUEUE", _DEFAULT_QUEUE, str),
--- a/workers/queue_backend/pg_queue/consumer.py
+++ b/workers/queue_backend/pg_queue/consumer.py
@@ -178,8 +178,27 @@ class PgQueueConsumer:
                 message.msg_id,
             )
 
-    def run(self, *, install_signals: bool = True) -> None:
-        """Poll loop with empty-queue backoff and graceful shutdown."""
+    def _registered_task_count(self) -> int:
+        """Count application tasks (excluding Celery's built-ins)."""
+        return sum(1 for name in self._app.tasks if not name.startswith("celery."))
+
+    def run(self, *, install_signals: bool = True, require_tasks: bool = True) -> None:
+        """Poll loop with empty-queue backoff and graceful shutdown.
+
+        Refuses to start if no application tasks are registered — a strong
+        signal the worker app wasn't bootstrapped, in which case *every*
+        message would be dropped as "unknown task". This makes a
+        misconfigured launch fail loudly instead of silently destroying data.
+        """
+        if require_tasks and self._registered_task_count() == 0:
+            raise RuntimeError(
+                "PG-queue consumer: no application tasks are registered — the "
+                "worker app was not bootstrapped. Launch via "
+                "`python -m pg_queue_consumer` (or ./run-worker.sh "
+                "pg-queue-consumer), not bare "
+                "`python -m queue_backend.pg_queue.consumer`. Refusing to start "
+                "to avoid dropping every message as an unknown task."
+            )
         self._running = True
         if install_signals:
             self._install_signal_handlers()
--- a/workers/run-worker.sh
+++ b/workers/run-worker.sh
@@ -114,9 +114,12 @@ WORKER_TYPE:
     scheduler, schedule   Run scheduler worker (scheduled pipeline tasks)
     executor              Run executor worker (extraction execution tasks)
     ide-callback          Run IDE callback worker (Prompt Studio post-execution callbacks)
+    pg-queue-consumer     Run PG-queue poll-loop consumer (opt-in; not part of 'all')
     all                   Run all workers (in separate processes, includes auto-discovered pluggable workers)
 
 Note: Pluggable workers in pluggable_worker/ directory are automatically discovered and can be run by name.
+Note: pg-queue-consumer overrides: WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE (source worker whose
+      tasks to load, default notification) and WORKER_PG_QUEUE_CONSUMER_QUEUE (queue to poll).
 
 OPTIONS:
     -e, --env-file FILE   Use specific environment file (default: .env)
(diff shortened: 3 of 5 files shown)


workers/queue_backend/pg_queue/consumer.py
```diff
@@ -178,8 +178,27 @@ def _handle(self, message: QueueMessage) -> None:
                 message.msg_id,
             )
 
-    def run(self, *, install_signals: bool = True) -> None:
-        """Poll loop with empty-queue backoff and graceful shutdown."""
+    def _registered_task_count(self) -> int:
+        """Count application tasks (excluding Celery's built-ins)."""
+        return sum(1 for name in self._app.tasks if not name.startswith("celery."))
+
+    def run(self, *, install_signals: bool = True, require_tasks: bool = True) -> None:
```
muhammad-ali-e, line 188: **🟠 Silent failure / observability (Medium)** — The new `require_tasks` guard catches an *empty* registry, but not a *wrong* one (the `__main__` docstring acknowledges this). If `WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE` selects the wrong worker, the registry is non-empty but missing the target task, so every message is **deleted** as an "unknown task" (`consumer.py:139-148`) — strictly worse than the empty case this guard protects against.

Suggest logging the resolved worker type + registered application task names at startup (INFO), or asserting the polled queue's expected task is present, so a wrong-type boot fails/warns once instead of silently destroying N messages.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -178,8 +178,27 @@ def _handle(self, message: QueueMessage) -> None:
...
+        """Poll loop with empty-queue backoff and graceful shutdown.
+
+        Refuses to start if no application tasks are registered — a strong
+        signal the worker app wasn't bootstrapped, in which case *every*
+        message would be dropped as "unknown task". This makes a
+        misconfigured launch fail loudly instead of silently destroying data.
+        """
+        if require_tasks and self._registered_task_count() == 0:
```
muhammad-ali-e, line 196: **🟡 Test coverage (Medium)** — Only the *failure* arm of this guard is tested (`test_run_refuses_to_start_with_empty_registry`). Three cheap unit tests would complete its contract and catch an inverted/over-broad regression that would brick the whole consumer fleet on startup:
1. **Positive:** a non-empty registry *starts* — patch `time.sleep` → `stop()`, assert no raise + `client.read` called.
2. **Bypass:** `require_tasks=False` skips the guard on an empty app.
3. **Filter:** `_registered_task_count()` returns 0 for a `celery.*`-only app and the app-task count when one `@shared_task` is registered.

workers/run-worker.sh
```diff
@@ -647,25 +679,43 @@ run_worker() {
...
     if [[ "$detach" == "true" ]]; then
-        # Run in background
-        nohup "${cmd_args[@]}" > "$worker_type.log" 2>&1 &
+        # Run in background. Write to an ABSOLUTE log path ($worker_dir is
+        # absolute) so the file lands where resolve_log_file() / -L / -C look,
+        # regardless of cwd. Workers that run from the workers root (pluggable
+        # workers, pg_queue_consumer) would otherwise drop a relative
+        # "$worker_type.log" at the root, where -L/-C can't find it.
+        nohup "${cmd_args[@]}" > "$worker_dir/$worker_type.log" 2>&1 &
         local pid=$!
```
muhammad-ali-e: **🔴 Silent failure (High)** — `nohup … &` captures the fork PID and unconditionally prints "started", but `set -e` does not apply to backgrounded jobs. If the consumer dies on startup (the new `require_tasks` `RuntimeError`, an `import worker` failure, or a bad `_env` cast), the shell still reports success and the crash lands only in the log. Acute for `pg_queue_consumer`: it has **no health port**, so `--status` relies purely on `pgrep`, and the new opt-in skip (line ~506) then renders a crashed consumer as simply *absent*. Verify liveness right after launch:

```suggestion
        nohup "${cmd_args[@]}" > "$worker_dir/$worker_type.log" 2>&1 &
        local pid=$!
        sleep 1
        if ! kill -0 "$pid" 2>/dev/null; then
            print_status $RED "$worker_type failed to start — last log lines:"
            tail -n 20 "$worker_dir/$worker_type.log"
            exit 1
        fi
```

workers/run-worker.sh
```diff
@@ -475,6 +501,12 @@ show_status() {
         local pids
         pids=$(get_worker_pids_oneline "$worker")
 
+        # Opt-in workers aren't part of `all`; only surface them when running
+        # so an intentional non-start doesn't read as a STOPPED failure.
+        if [[ -z "$pids" && -n "${OPTIN_WORKERS[$worker]:-}" ]]; then
```
muhammad-ali-e, line 522: **🟡 Silent failure (Medium)** — This `continue` suppresses the status row whenever an opt-in worker has no PID, which also erases the **crashed** state: a consumer that was started and then died reads identically to one never launched. Since `pg_queue_consumer` has no health port, `--status` is the only health view.

Consider distinguishing "never started" from "started-but-gone" (e.g. a pidfile/marker written at launch → render `STOPPED`/`CRASHED` when the marker exists but no live PID), or at least print a dim `not started` line instead of omitting the row.

workers/run-worker.sh
```diff
@@ -301,6 +319,14 @@ validate_env() {
 #   --hostname=callback-worker-${id}@%h   (when WORKER_INSTANCE_ID is set)
 get_worker_pids() {
     local worker_type=$1
+    # The PG-queue consumer runs as `python -m pg_queue_consumer`, not a Celery
+    # `<type>-worker@host` process, so it has no `-worker` token to anchor on.
+    # Match its module invocation instead (covers both the `uv run python`
+    # parent and the `python -m` child). Keeps --status / -k / -r working for it.
+    if [[ "$worker_type" == "pg_queue_consumer" ]]; then
+        pgrep -f -- "-m[[:space:]]+${worker_type}([[:space:]]|\$)" || true
```
muhammad-ali-e: **🟢 Silent failure (Low)** — `pgrep` exits 1 for "no match" (expected, must be absorbed under `set -e`) but ≥2 for operational/regex errors. `|| true` collapses all of them to an empty result indistinguishable from "not running", so a regex/platform-dialect issue would silently make `--status`/`-k`/`-r` treat a live consumer as absent (a `-k` becomes a no-op; `-r` may spawn a duplicate). Consider capturing `rc` and warning to stderr when `rc > 1`, while still returning 0 for `rc == 1`.

workers/run-worker.sh
```diff
@@ -61,6 +65,9 @@ declare -A WORKER_QUEUES=(
     ["scheduler"]="scheduler"
     ["${EXECUTOR_WORKER_TYPE}"]="celery_executor_legacy"
     ["${IDE_CALLBACK_WORKER_TYPE}"]="${IDE_CALLBACK_WORKER_TYPE}"
+    # The PG queue (in pg_queue_message) this consumer polls — exported as
+    # WORKER_PG_QUEUE_CONSUMER_QUEUE, not a Celery --queues value.
+    ["pg_queue_consumer"]="notifications"
```
muhammad-ali-e: **🟡 Robustness / observability (Medium)** — The polled queue is hard-coded to `notifications` and exported as `WORKER_PG_QUEUE_CONSUMER_QUEUE`. The producer enqueues under the task's Celery queue (`dispatch.py:112`, `queue or "default"`). For the first migrated task (`send_webhook_notification`) these **align today** (verified — `helper.py:89` uses `queue="notifications"`), so this is correct now.

But the consumer's `WHERE queue_name = %s` is an exact match, so any future task routed under `notifications_webhook`/`_email`/`default` would leave rows unclaimed forever with no error — the `require_tasks` guard doesn't cover this axis. Suggest deriving the queue from the same source as the producer rather than hard-coding, and/or a one-time WARNING if `pg_queue_message` holds rows for `queue_name`s other than the one polled.

workers/pg_queue_consumer/__main__.py
```diff
@@ -0,0 +1,33 @@
+"""Entry point: bootstrap a worker app's tasks, then run the PG-queue consumer.
+
+The root ``worker`` module loads exactly ONE worker type's tasks, chosen by the
+``WORKER_TYPE`` env (``worker.py`` → ``WorkerBuilder.build_celery_app`` →
```
muhammad-ali-e: **🟡 Comment accuracy (Medium)** — This call chain is inaccurate: `WorkerBuilder.build_celery_app` only constructs/configures the Celery app — it does **not** import `tasks.py`. The actual task import is separate module-level logic in `worker.py` (an `importlib` load of `<worker_path>/tasks.py`) that runs *after* `build_celery_app` returns. A maintainer trusting this chain would look inside `build_celery_app` for the task-loading and not find it.

Suggest: "…`worker.py` builds the Celery app via `WorkerBuilder.build_celery_app`, then imports that type's `tasks.py` (module-level in `worker.py`)."

workers/pg_queue_consumer/__main__.py
```diff
@@ -0,0 +1,33 @@
...
+imports that type's ``tasks.py``). The PG consumer drains a specific source
+worker's queue, so it must register THAT worker's tasks: we set ``WORKER_TYPE``
```
muhammad-ali-e, line 8: **🟢 Comment accuracy (Low)** — "the first migrated leaf task" conflates a worker *type* (`notification`, a `WorkerType` enum member) with a *task*. The actual leaf task is `send_webhook_notification`; `notification` is the worker that *owns* it. "first migrated" is also a point-in-time claim that rots as more tasks migrate.

Suggest: "default `notification` — the worker that owns the first migrated leaf task, `send_webhook_notification`." (Same wording also at `run-worker.sh:688-689`.)

workers/pg_queue_consumer/__main__.py
```diff
@@ -0,0 +1,33 @@
...
+from ``WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE`` (default ``notification`` — the
+first migrated leaf task) BEFORE importing ``worker``.
+
+This override is required: ``run-worker.sh`` exports its own ``WORKER_TYPE``
+(the consumer pseudo-type ``pg_queue_consumer``, which owns no tasks), so a
```
muhammad-ali-e, line 14: **🟢 Comment accuracy (Low)** — Accurate today, but hard-coding "every notification message would be dropped" makes the example misleading the moment someone overrides `WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE` to drain a different worker. Suggest generalizing: "every message for the intended worker would be dropped as an unknown task."

muhammad-ali-e (commented):
## PR Review Toolkit — summary (PR #2047)

Ran six specialized agents (Code Reviewer, Silent-Failure Hunter, Type-Design Analyzer, PR Test Analyzer, Comment Analyzer, Code Simplifier). The PR is small, well-factored, and unusually well-commented; tests pass and `bash -n` is clean. No merge-blocking correctness bug found. Findings below are posted inline; three target pre-existing (unchanged) lines outside this diff and so are listed here only:

**Pre-existing lines (cannot anchor inline):**
- 🟠 **`consumer.py:248` `_env()` casting** — `cast(os.getenv(...))` raises a bare `ValueError` (e.g. `int("abc")`) with no var name / error id. Wrap the cast and name the offending `WORKER_PG_QUEUE_CONSUMER_*` var. (Loud is correct, but combined with the backgrounded-launch issue it gets swallowed into the log.)
- 🟠 **`consumer.py:248-249` `_env` typing** — `def _env(suffix, default, cast: type) -> object` erases types at the typed `PgQueueConsumer.__init__` boundary and fights the repo's strict mypy. Use a `TypeVar`: `def _env(suffix: str, default: _T, cast: Callable[[str], _T]) -> _T`.
- 🟡 **`consumer.py:215-221` poll-loop `except Exception`** — correctly survives transient blips, but equally swallows *permanent* faults (schema drift, revoked grants, a `_handle` bug) into an infinite 0.5 Hz log-spam loop. Count consecutive failures and escalate past a threshold.

**Priority order:** 🔴 nohup-no-liveness (`run-worker.sh:718`) → 🟠 wrong-registry observability (`consumer.py:185`), `_env` casting/typing → 🟡 opt-in-hides-crash, queue-skew visibility, test gaps, call-chain comment → 🟢 pgrep `|| true`, comment wording.

The Code Simplifier found nothing to change (the two duplication points are intentional, behavior-preserving).

workers/run-worker.sh
muhammad-ali-e: Fixed in 2f717c6d1. The detached launch now verifies liveness — `sleep 1; kill -0 "$pid"` — and on a dead fork prints the last 20 log lines and `return 1` instead of reporting success. Applied to the shared detach path so every worker benefits, not just the consumer.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in 2f717c6d1. `run()` now logs the registered application task names at startup, e.g. `… 7 application task(s) registered: …, send_webhook_notification`. A wrong (non-empty but mismatched) registry is now diagnosable from one line. I kept it a log rather than an assertion because the consumer is generic — it cannot know ahead of time which `task_name`s a queue will carry, so listing the registry is the right generic signal.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in 2f717c6d1. Added the three arms: positive (`test_run_starts_with_registered_tasks`), bypass (`test_run_bypasses_guard_when_require_tasks_false`), and filter (`test_registered_task_count_excludes_celery_builtins`).

workers/run-worker.sh
muhammad-ali-e: Fixed in 2f717c6d1. `get_worker_pids` now captures `rc` and warns to stderr when `rc > 1` (operational/regex error), while still returning empty for `rc == 1` (genuine no-match). A platform/regex issue no longer silently reads as "not running".

workers/pg_queue_consumer/__main__.py
muhammad-ali-e: Fixed in 2f717c6d1. Reworded: `worker.py` builds the app via `WorkerBuilder.build_celery_app` (which configures but does not import tasks), then — in separate module-level logic — `importlib`-loads that type's `tasks.py`.

workers/pg_queue_consumer/__main__.py
muhammad-ali-e: Fixed in 2f717c6d1. Now reads "the worker that owns the first migrated leaf task, `send_webhook_notification`" — distinguishing the worker type from the task. Same wording updated at run-worker.sh.

workers/pg_queue_consumer/__main__.py
muhammad-ali-e: Fixed in 2f717c6d1. Generalised to "every message for the intended worker would be dropped as an unknown task".

workers/run-worker.sh
muhammad-ali-e: Partially addressed. The new post-launch liveness check (2f717c6d1) covers the most likely case — a consumer that dies *during startup* now fails loudly at launch (error + log tail + non-zero exit) rather than reading as absent. The remaining gap — a consumer that starts cleanly and dies *later* — still reads as "not running" in `--status`, since without a health port there is no marker to distinguish "never started" from "started-then-gone". A real liveness/health endpoint (already listed as out-of-scope rollout work) is the proper fix for steady-state crash detection; a pidfile would be a partial stand-in but I held off to avoid new state machinery in this PR. Happy to add the pidfile here if you prefer.

workers/run-worker.sh
muhammad-ali-e: Deferred (verified-correct-today, with rationale). For the only migrated task the producer queue and the polled queue align — `helper.py` enqueues `send_webhook_notification` under `queue="notifications"`, which is exactly what the consumer polls. The exact-match risk for a *future* task routed under `notifications_webhook`/`_email`/`default` is real but speculative until a second task migrates. Deriving the polled queue from the producer's source (rather than the hard-coded map) and/or a one-time WARNING for rows under unexpected `queue_name`s belongs with that migration — I would rather wire it when there is a second queue to actually exercise it than add untested cross-queue logic now. Tracking it for the next leaf migration.

muhammad-ali-e:
## Review feedback addressed — `2f717c6d1`

Thanks for the thorough pass. Validity-checked each finding; here's the disposition.

**Fixed**
- 🔴 **nohup no-liveness** (`run-worker.sh`) — detached launch now verifies `kill -0` after `sleep 1`, prints the log tail + returns non-zero on a dead fork. Applied to the shared detach path (benefits every worker).
- 🟠 **wrong-registry observability** (`consumer.py`) — `run()` logs the registered application task names at startup, so a non-empty-but-mismatched registry is diagnosable from one line (`… 7 application task(s) registered: …, send_webhook_notification`).
- 🟠 **`_env` casting + typing** — `TypeVar`-typed (was `cast: type -> object`, erasing types at the typed `__init__`); a bad value now raises `Invalid WORKER_PG_QUEUE_CONSUMER_<X>=…` instead of a context-free `ValueError`.
- 🟡 **test gaps** — added the guard's positive / `require_tasks=False` bypass / built-in-filter arms.
- 🟢 **`pgrep || true`** — warns on `rc > 1` (operational/regex error) instead of collapsing it to "not running".
- 🟢 **comment accuracy** ×3 — `build_celery_app` configures but doesn't import tasks; `notification` is the worker that *owns* the leaf task `send_webhook_notification`; generalised the "every notification dropped" example.
- **SonarCloud S1192** — extracted the repeated `"pg_queue_consumer"` literal into a `readonly` constant.

**Deferred (with rationale, replied inline)**
- 🟡 **opt-in-hides-crash** — start-time crashes now caught by the liveness check; steady-state (died-later) detection needs the future health endpoint (already a rollout item).
- 🟡 **queue-skew** — producer/consumer queues are verified-aligned for the one migrated task; cross-queue detection belongs with the second leaf migration (don't want untested cross-queue logic now).
- 🟡 **poll-loop escalation past a failure threshold** — pre-existing 9c (#2045) behaviour, and it changes crash semantics (exit-on-sustained-fault vs log-and-continue) — worth its own discussion/PR rather than bundling here.

**Re-verified live after the changes:** `./run-worker.sh -r -d pg-queue-consumer` → liveness passes, `--status` shows `RUNNING`, startup log lists the 7 notification tasks. Consumer test suite green.

workers/pg_queue_consumer/__main__.py
```diff
@@ -0,0 +1,36 @@
...
+# (not setdefault) because the launcher's own WORKER_TYPE owns no tasks.
+os.environ["WORKER_TYPE"] = os.environ.get(
+    "WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE", "notification"
+)
+
+import worker  # noqa: E402,F401 — side-effect: registers the source worker's tasks
+from queue_backend.pg_queue.consumer import main  # noqa: E402
+
+if __name__ == "__main__":
+    main()
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 5bec0f8b4. Moved the `WORKER_TYPE` mutation and `import worker` into a guarded `_bootstrap_and_run()` invoked only under `if __name__ == "__main__":` — so an accidental import (test/IDE/type-checker) no longer mutates the env or triggers the bootstrap. Verified: `import pg_queue_consumer.__main__` leaves `WORKER_TYPE` untouched and doesn't load `worker`; `python -m pg_queue_consumer` still bootstraps the 7 notification tasks and runs.

workers/run-worker.sh
```diff
@@ -647,28 +692,59 @@ run_worker() {
...
+        # startup (e.g. the consumer's require_tasks RuntimeError, an
+        # `import worker` failure, a bad env cast) would still be reported as
+        # "started" — and for pg_queue_consumer, which has no health port, a
+        # dead process then just reads as absent in --status. Verify liveness.
+        sleep 1
+        if ! kill -0 "$pid" 2>/dev/null; then
+            print_status $RED "$worker_type worker failed to start (PID $pid exited) — last log lines:"
+            tail -n 20 "$log_file" 2>/dev/null
+            return 1
+        fi
```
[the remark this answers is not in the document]
muhammad-ali-e: Kept it general, but clarified intent in the comment (5bec0f8b4). The check is a best-effort fast-fail for *immediate* (sub-second) crashes — `import worker` failure, bad env cast, the consumer's `require_tasks` RuntimeError — which can hit any worker type, so scoping it to the consumer would miss those for everyone else. You're right that it is **not** a connectivity check: a worker that dies slowly (broker connect timing out after >1s) still passes here and surfaces via its health port / `--status`; the comment now says so explicitly, so it doesn't read as broker-validation. I left the sleep at 1s rather than 2–3s because bumping it trades latency on every `-d` for only partial coverage of slow failures (which need the real health endpoint anyway). On the `all` latency: those launches are backgrounded subshells (`( run_worker … ) &` then `sleep 2`), so each subshell's 1s overlaps the inter-launch sleep rather than serializing — wall-clock for `all` is unchanged.

muhammad-ali-e:
## Greptile feedback addressed — `5bec0f8b4`

Thanks @greptile-apps (confidence 4/5). All three P2 findings handled:

- **Module-level side-effects in `__main__.py`** — moved the `WORKER_TYPE` mutation + `import worker` into a guarded `_bootstrap_and_run()` called only under `if __name__ == "__main__":`. Accidental imports are now inert (verified: import leaves `WORKER_TYPE` untouched and doesn't bootstrap `worker`; `python -m` path unchanged).
- **`usage()` help text** (outside-diff comment) — added `pg-queue-consumer` to the worker-type list plus a note documenting the `WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE` / `_QUEUE` overrides.
- **Liveness check scope** — kept general but clarified in the comment that it's a best-effort fast-fail for *immediate* (sub-second) crashes (import/config faults, which affect any worker), explicitly **not** a connectivity check. Left the sleep at 1s (bumping it trades latency for only partial slow-failure coverage, which needs the real health endpoint); `all` is unaffected since those subshells overlap the 1s with the inter-launch sleep. Rationale on the thread.

Re-verified live after the changes: `python -m pg_queue_consumer` boots the 7 notification tasks, `--status` → RUNNING, `--help` lists the worker.

PR #2048, opened 2026-06-12
UN-3542 [FIX] Prevent last org admin from demoting themselves

chandrasekharan-zipstack:
## Summary

- The sole org admin could change their own role to "user" via `POST /users/role/`, leaving the org with no admin and no way to recover.
- Added `_ensure_not_last_admin_demotion` guard in `AuthenticationController` that blocks the operation with `403 Forbidden` when it would remove the last admin.
- Guard is wired into both the POST (assign/sync role) and DELETE (remove role) paths.

## What changed

**`backend/account_v2/authentication_controller.py`**
- Import `UserRole` from `account_v2.enums`
- New `_ensure_not_last_admin_demotion(email, new_role)` — checks `OrganizationMemberService.get_members_by_role(ADMIN)` count before allowing the change
- Called at the top of `add_user_role` and `remove_user_role`

**`backend/account_v2/tests/test_last_admin_guard.py`**
- 5 unit tests covering: promote-to-admin (no-op), target-not-admin (no-op), multiple admins remain (no-op), sole admin demoted (raises Forbidden), target not found (no-op)
- Uses `sys.modules` stubs — no Django setup or live DB required; runs under the test rig's `unit-backend` group

## Test plan

- [ ] Run `pytest backend/account_v2/tests/test_last_admin_guard.py` — 5 passed
- [ ] Manual: as the only admin, try to change own role to User in Settings → Users — should get 403
- [ ] Manual: with 2 admins, downgrade one — should succeed

Jira: [UN-3542](https://zipstack.atlassian.net/browse/UN-3542)

[UN-3542]: https://zipstack.atlassian.net/browse/UN-3542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/account_v2/authentication_controller.py
```diff
@@ -417,9 +418,22 @@ def remove_users_from_organization(
 
         return is_removed
 
+    def _ensure_not_last_admin_demotion(self, email: str, new_role: str) -> None:
```
athul-rs, line 421: **Critical — this guard fails open on Auth0-backed deployments (i.e., cloud).**

`account_v2.enums.UserRole.ADMIN.value` is `"admin"`, but the Auth0 plugin in this repo has its own enum (`backend/plugins/authentication/auth0/enums.py:20-25`) where `ADMIN = "unstract_admin"`. `auth0_service.get_organization_role_of_user()` returns `"unstract_admin"` and `save_organization_user_role()` persists exactly that string into `OrganizationMember.role`.

So in any Auth0 deployment:
- `target.role == UserRole.ADMIN.value` → always False
- `get_members_by_role(UserRole.ADMIN.value)` → always empty
- → the guard silently passes and the sole-admin lockout remains fully reproducible.

The abstraction for this already exists: `self.is_admin_by_role()` (line 258) delegates to the active auth plugin — the Auth0 implementation (`auth0_service.py:519`) maps `"unstract_admin"` → admin. `OrganizationMemberService.is_user_organization_admin()` already follows this pattern for the same reason.

Suggested shape:
```python
def _ensure_not_last_admin_demotion(self, email: str, new_role: str) -> None:
    # Org must always retain at least one admin.
    if self.is_admin_by_role(role=new_role):
        return
    target = OrganizationMemberService.get_user_by_email(email=email)
    if target and self.is_admin_by_role(role=target.role):
        admins = [
            m
            for m in OrganizationMemberService.get_members()
            if self.is_admin_by_role(role=m.role)
        ]
        if len(admins) <= 1:
            raise Forbidden("Cannot demote the only admin of the organization")
```
(`get_members_by_role` can't be used here since the plugin's admin role string isn't exposed generically; member counts per org are small, so filtering in Python is fine.)

backend/account_v2/authentication_controller.py
```diff
@@ -417,9 +418,22 @@ def remove_users_from_organization(
...
+        # Org must always retain at least one admin.
+        if new_role == UserRole.ADMIN.value:
+            return
+        target = OrganizationMemberService.get_user_by_email(email=email)
+        if target and target.role == UserRole.ADMIN.value:
+            admin_count = len(
+                OrganizationMemberService.get_members_by_role(UserRole.ADMIN.value)
+            )
+            if admin_count <= 1:
```
athul-rs, line 433: **Minor — TOCTOU race.** Two concurrent demote requests against two different admins can both read `admin_count == 2` and both succeed → zero admins. `transaction.atomic()` + `select_for_update()` on the org's member rows would close it. Severity is low (the actors are the org's own admins), but worth a comment acknowledging the race if you choose not to lock.

backend/account_v2/authentication_controller.py
```diff
@@ -438,6 +452,10 @@ def add_user_role(
     def remove_user_role(
         self, request: Request, org_id: str, email: str, role: str
     ) -> str | None:
+        if role == UserRole.ADMIN.value:
```
athul-rs: Same role-string bug as the guard itself: in Auth0 deployments the DELETE payload carries the plugin role string (`"unstract_admin"`), so `role == UserRole.ADMIN.value` is False and the guard is skipped entirely. Use `self.is_admin_by_role(role=role)` here.

backend/account_v2/tests/test_last_admin_guard.py
```diff
@@ -0,0 +1,162 @@
...
+# Module-level stubs — installed once before the controller is imported.
+# ---------------------------------------------------------------------------
+
+
+def _install_stubs() -> None:
+    def _mod(name: str, **attrs: object) -> types.ModuleType:
+        m = types.ModuleType(name)
+        for k, v in attrs.items():
+            setattr(m, k, v)
+        sys.modules[name] = m
```
athul-rs: **Major — this poisons `sys.modules` for the whole pytest session.** `_install_stubs()` runs at import/collection time and unconditionally overwrites `sys.modules["account_v2.models"]`, `"utils.cache_service"`, `"account_v2.organization"`, etc. Any other test collected in the same `pytest backend/` run that imports these afterwards gets the stubs (`User = object`), and the real `authentication_controller` module is cached bound to stubs for everyone else.

Please either use `unittest.mock.patch.dict(sys.modules, ...)` / `monkeypatch` with proper teardown, or write these as plain Django `TestCase` tests like the rest of the backend.

Also worth noting: these tests only exercise the helper directly — the wiring (guard invoked from `add_user_role`, and `remove_user_role` only when the removed role is admin) is untested, and the Auth0 role-string bug flagged on the controller is invisible to fully-stubbed tests like these.

backend/account_v2/tests/test_last_admin_guard.py
```diff
@@ -0,0 +1,162 @@
+"""Tests for the last-admin demotion guard in AuthenticationController.
+
+Stubs all Django-model-importing modules at sys.modules level before importing
+the controller, so these tests run without Django setup or a live database.
+
+Test rig: covered by the `unit-backend` group in tests/groups.yaml
```
athul-rs: `tests/groups.yaml` doesn't exist in this repo, and CI (`ci-test.yaml` → tox) only runs the `runner`/`sdk1`/`prompt-service` envs — backend tests never run in CI. This docstring claim should be removed or corrected; as-is it implies coverage that doesn't exist.

athul-rs (changes requested):
@chandrasekharan-zipstack reviewed this with codebase context — the invariant and the choke point are right, but the guard silently no-ops on Auth0-backed deployments (details in the inline comments), which is presumably where UN-3542 was hit. Two more items below the inline ones:

**Adjacent unguarded path, same invariant:** `remove_users_from_organization` (`authentication_controller.py:385`) deletes memberships outright with no last-admin guard — the sole admin can still remove *themselves from the org* and reproduce the exact lockout this PR fixes. Either extend the guard there or file a follow-up and state the scope in the PR description.

**Test plan:** the checkboxes are unticked, and backend tests don't run in CI (`ci-test.yaml` → tox only covers `runner`/`sdk1`/`prompt-service`), so nothing automated will catch a regression here. Please run the manual checks before merge.

What's good: fail-safe direction for unknown role strings on the new-role side, guard placed at the controller choke point both OSS and cloud role flows route through (`users_view.py:39,63`), org-scoping via `DefaultOrganizationManagerMixin`/`UserContext` checks out, and the invariant is correctly on the target rather than the requester (covers self-demotion and demote-by-other).

### Lint Rule Candidates
Patterns spotted that could become automated lint rules if they recur:
- **Hardcoded role-string comparison**: `== UserRole.ADMIN.value` in OSS code that runs under pluggable auth (this PR's guard + `remove_user_role` wiring; same pattern pre-exists in `authentication_service.py:285,297`). Role comparisons in `account_v2`/`tenant_account_v2` should go through `is_admin_by_role()`.
- **`sys.modules` assignment in tests** without teardown — breaks pytest session isolation.
- **Docstring references to nonexistent infra** (`tests/groups.yaml`) — likely carried over from another repo's conventions.

backend/account_v2/authentication_controller.py
chandrasekharan-zipstack: Fixed in 9c97603 — guard now uses `self.is_admin_by_role()` (delegates to the active auth plugin) instead of comparing against `UserRole.ADMIN.value`, and counts admins via `get_members()` filtered by that predicate. The `remove_user_role` gate uses the same predicate. So the Auth0 `unstract_admin` string is now handled correctly.

backend/account_v2/authentication_controller.py
chandrasekharan-zipstack: Acknowledged as a known limitation (comment added in code). Not adding a DB lock here: the mutation goes through the auth plugin which, for Auth0/cloud, performs an external IdP call that is the source of truth — a local `select_for_update` would not actually serialize that. Window is narrow (two simultaneous demotions when exactly 2 admins exist) and actors are trusted org admins. Tracking a proper atomic fix separately if it proves necessary.

backend/account_v2/authentication_controller.py
```diff
@@ -417,9 +418,22 @@ def remove_users_from_organization(
...
+        # Org must always retain at least one admin.
+        if new_role == UserRole.ADMIN.value:
+            return
+        target = OrganizationMemberService.get_user_by_email(email=email)
+        if target and target.role == UserRole.ADMIN.value:
+            admin_count = len(
+                OrganizationMemberService.get_members_by_role(UserRole.ADMIN.value)
+            )
+            if admin_count <= 1:
+                raise Forbidden("Cannot demote the only admin of the organization")
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Won't fix in this PR — documented as a known limitation in code. The role mutation goes through the auth plugin (external Auth0 call for cloud), so a local row lock wouldn't serialize the true source of truth. Low severity (trusted admins, narrow window).

backend/account_v2/tests/test_last_admin_guard.py
```diff
@@ -0,0 +1,228 @@
...
+    def get_user_by_email(cls, email: str) -> SimpleNamespace | None:
+        return cls._member_by_email
+
+    @classmethod
+    def get_members(cls) -> list:
+        return cls._all_members
+
+    # Kept for compatibility with other call sites in the controller.
+    @classmethod
+    def get_user_by_id(cls, id: object) -> None:
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Won't fix: the stub must mirror the real `OrganizationMemberService.get_user_by_id(id=...)` signature, which the controller calls with the `id=` keyword at multiple sites. Renaming would break the stub. Added `# noqa: A002` with a reason instead.

backend/account_v2/tests/test_last_admin_guard.py
chandrasekharan-zipstack: `tests/groups.yaml` does exist on main and lists `account_v2/tests` under `unit-backend`, and CI now runs `tox -e unit` via the rig. That said, `unit-backend` is `optional: true` + gated on postgres, so it likely doesn't execute today — I've softened the docstring to not overclaim CI coverage.

athul-rs (approved):

PR #2049, opened 2026-06-12
UN-3543 [FIX] Use per-user distinct_id and deployment tagging for PostHog

chandrasekharan-zipstack:
## What

- PostHog `distinct_id` is now the **user's email** (fallback `orgId:userId`) instead of `orgId:orgName`, so each user is an individual PostHog person
- `org_id`, `org_name` and `software` (OSS/SAAS) are registered as **super-properties**, stamped on every captured event
- New `deployment` super-property derived from hostname: `us-prod` / `eu-prod` / `self-hosted`
- PostHog can now be **disabled at container startup** via `VITE_ENABLE_POSTHOG=false` (exposed through `window.RUNTIME_CONFIG`) — no rebuild needed; staging should set this
- SDK init hardening: `person_profiles: "identified_only"`, `respect_dnt: true`

## Why

- All deployments (US prod, EU prod, OSS self-hosts) report to a single PostHog project; with the org as `distinct_id`, every user in an org collapsed into one PostHog person — per-user analytics (DAU/WAU, funnels, retention) were impossible and person properties thrashed last-writer-wins
- Org-level breakdowns remain possible via the registered super-properties (no paid Group Analytics add-on needed)
- `deployment` tagging replaces the brittle `orgName === "mock_org"` check for segmenting OSS vs SaaS and US vs EU
- Email chosen as `distinct_id` because bare Django user pks collide across regions/self-hosts in the shared project
- Vite env vars are baked at image build time and the same image is promoted across environments, so a runtime-config toggle is the only way to disable PostHog on staging without forking the build

## How

- `frontend/src/hooks/usePostHogEvents.js`: `setPostHogIdentity` builds `distinct_id` from email, keeps existing person properties (`name`, `org`, `software`, `email`, `userId`), and calls `posthog.register()` with org super-properties
- `frontend/src/index.jsx`: reads `enablePosthog` from `window.RUNTIME_CONFIG` first (falls back to build-time `import.meta.env.VITE_ENABLE_POSTHOG`); hostname→deployment map registered right after `posthog.init`
- `frontend/generate-runtime-config.sh`: emits `enablePosthog` into `runtime-config.js` from the container's `VITE_ENABLE_POSTHOG` (or `REACT_APP_ENABLE_POSTHOG`) env at startup

- No product features touched — analytics instrumentation only; PostHog capture failures are already swallowed by the helper
- PostHog entities verified against the live project before this change: **zero feature flags/experiments exist** (no flag-bucketing reshuffle); event names and person-property names are unchanged, so existing insights/cohorts keep evaluating
- Expected metric shifts (not breakage): unique-user counts step UP at deploy (previously counted orgs); the "All Unique Users" dashboard tile becomes one-row-per-user with accurate emails; `respect_dnt` drops events from DNT-enabled browsers
- Runtime-config default preserves current behaviour: when the env var is unset, the generated value is an empty string, which falls through to the build-time env (PostHog stays enabled in prod images)
- Rollout note: add a PostHog annotation on deploy date and notify marketing/growth about the step-change

## Database Migrations

- None

## Env Config

- `VITE_ENABLE_POSTHOG` is now honoured at **container runtime** via `runtime-config.js` (previously build-time only)
- Follow-up (deploy repo): set `VITE_ENABLE_POSTHOG=false` on the staging frontend deployment so staging stops reporting to the shared PostHog project

## Relevant Docs

- Internal PostHog audit notes (2026-06-02) tracked in the team knowledge base

## Related Issues or PRs

- [UN-3543](https://zipstack.atlassian.net/browse/UN-3543)
- Follow-ups tracked there: central event enum, structured props, `*_failed` events, per-install `instance_id` for OSS

## Dependencies Versions

- No changes

## Notes on Testing

- `bunx biome check` passes on the touched files
- Manual: log in locally → PostHog person is keyed by email with `org_id`/`org_name`/`software`/`deployment` super-properties on captured events
- Runtime toggle: start the frontend container with `VITE_ENABLE_POSTHOG=false` → `/config/runtime-config.js` carries `enablePosthog: "false"` and PostHog never initializes

## Checklist

[UN-3543]: https://zipstack.atlassian.net/browse/UN-3543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jaseemjaskp (commented):
Automated review (PR Review Toolkit: Code Reviewer, Silent Failure Hunter, Type Design Analyzer, PR Test Analyzer, Comment Analyzer, Code Simplifier).

Scope: the 3 files changed by this PR. I've omitted findings that duplicate the existing greptile comments (degenerate `distinctId` at usePostHogEvents.js:69, `reset()` on logout at :93, and `||`/empty-string semantics at index.jsx:17 — all still worth addressing). The items below are new.

frontend/src/hooks/usePostHogEvents.js
```diff
@@ -57,27 +57,40 @@ const usePostHogEvents = () => {
...
-      optionalParams["email"] = email;
-    } else {
-      optionalParams["userId"] = userId;
+      personProperties["email"] = email;
     }
+    if (userId) {
+      personProperties["userId"] = userId;
+    }
+
+    posthog.identify(distinctId, personProperties);
```
jaseemjaskp: **[High] Email promoted to `distinct_id` is shipped to a single shared PostHog project from every deployment**

This PR makes `email` the primary `distinct_id` (L69) and sends it via `identify` here. In `index.jsx` every deployment — including self-hosted and OSS (`mock_org`) installs — reports to the same hardcoded Unstract project (`API_KEY` + `eu.i.posthog.com`), and PostHog is on by default unless `VITE_ENABLE_POSTHOG=false`. The net effect is that self-hosted customers' end-user email addresses are sent to Unstract by default, with email now the primary identity key (a step beyond the previous `orgId:orgName` distinct_id). Note `person_profiles: "identified_only"` does **not** mitigate this once `identify(email, …)` runs — an identified person profile with PII is created.

Please confirm this is a deliberate product/legal (GDPR) decision. Mitigations to consider: skip `identify`/email when `getDeployment() === "self-hosted"` or `software === "OSS"`; default PostHog off for self-hosted (explicit opt-in); or fall back to the non-PII `${orgId}:${userId}` id and omit the `email` person property for OSS/self-hosted.

frontend/src/hooks/usePostHogEvents.js
```diff
@@ -57,27 +57,40 @@ const usePostHogEvents = () => {
...
 
-    posthog.identify(distinctId, optionalParams);
+    // Super-properties ride on every captured event, enabling org-level
+    // breakdowns without the group analytics add-on
+    posthog.register({
```
jaseemjaskp: **[Medium] `identify`/`register` are unguarded, unlike the sibling `setPostHogCustomEvent`**

`setPostHogCustomEvent` (L96-102) wraps `posthog.capture` in try/catch precisely because "PostHog tracking is non-critical." The new `identify`/`register` calls have no such guard, yet they run inside a `useEffect` right after login. If the SDK throws (storage blocked in private mode, DNT interplay, ad-blocker shimming the global, or `posthog` uninitialized when disabled), the exception propagates out of the effect and can tear down the authenticated app. Wrap this body in try/catch consistent with the capture path — and log the error rather than silently swallowing it.

frontend/src/index.jsx
```diff
@@ -9,18 +9,37 @@ import { SocketProvider } from "./helpers/SocketContext.js";
...
+  // disable PostHog via VITE_ENABLE_POSTHOG instead of being mapped here.
+  const DEPLOYMENT_BY_HOST = {
+    "us-central.unstract.com": "us-prod",
+    "eu-west.unstract.com": "eu-prod",
+  };
+  const getDeployment = () =>
+    DEPLOYMENT_BY_HOST[window.location.hostname] || "self-hosted";
+
   // Initialize PostHog with the specified API key and host
   posthog.init(API_KEY, {
```
jaseemjaskp: **[Low] Top-level `posthog.init`/`register` can abort app bootstrap**

`posthog.init` and `posthog.register({ deployment })` run at module-eval time, before React mounts, with no try/catch. If `init` throws (storage SecurityError/QuotaExceeded in private mode, a CSP violation, or ad-blocker tampering), the module's top-level execution aborts and the `ReactDOM.createRoot(...).render(...)` at the bottom of the file never runs — a fully blank app that no React error boundary can catch (boundaries only catch render-phase errors, not module-eval throws). Consider wrapping the init/register block in try/catch so a third-party analytics failure can never block bootstrap.

frontend/src/hooks/usePostHogEvents.js
```diff
@@ -57,27 +57,40 @@ const usePostHogEvents = () => {
...
       name,
       org: orgName,
-      software: isOpenSource ? "OSS" : "SAAS",
+      software,
     };
 
-    const email = sessionDetails?.email;
-    const userId = sessionDetails?.id;
-
     if (email) {
```
jaseemjaskp: **[Nit] Use dot notation for static keys**

`personProperties["email"]` / `personProperties["userId"]` use bracket access for static keys, while the object literal just above (L72-76) uses dot-style keys. For consistency: `personProperties.email = email;` / `personProperties.userId = userId;`. Bracket notation is only needed for dynamic keys. (Pre-existing style, but these exact lines are touched by this PR.)

frontend/src/hooks/usePostHogEvents.js
chandrasekharan-zipstack: Addressed in 659dcb95a. Worth noting the prior code already sent `email` (and name) as person properties from every deployment — this PR originally just promoted it to the identity key. Agreed it deserved tightening rather than preserving: PII (email, name) is now gated to Unstract-managed SaaS deployments via `isSaasProdDeployment()`; self-hosted/OSS installs identify with an anonymous org-scoped `orgId:userId` id and no email/name person props — stricter than the pre-PR behaviour. Pure-OSS default installs were never a real exposure (mock auth hardcodes `email@mock.com`), the real win is on-prem installs with real auth.

frontend/src/hooks/usePostHogEvents.js
chandrasekharan-zipstack: Fixed in 659dcb95a — `setPostHogIdentity` body wrapped in try/catch with `console.error` (logged, not swallowed, per your note).

frontend/src/index.jsx
chandrasekharan-zipstack: Fixed in 659dcb95a — the init/register block is wrapped in try/catch with `console.error` so module evaluation can't abort before `ReactDOM.createRoot`.

frontend/src/hooks/usePostHogEvents.js
chandrasekharan-zipstack: Fixed in 659dcb95a — dot notation.

frontend/src/hooks/usePostHogEvents.js
```diff
@@ -57,27 +57,40 @@ const usePostHogEvents = () => {
...
+    posthog.identify(distinctId, personProperties);
 
-    posthog.identify(distinctId, optionalParams);
+    // Super-properties ride on every captured event, enabling org-level
+    // breakdowns without the group analytics add-on
+    posthog.register({
+      org_id: orgId,
+      org_name: orgName,
+      software,
+    });
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Already handled — `useLogout.js:26` calls `posthog.reset()` on logout, which clears the distinct_id and registered super-properties. The subsequent `window.location.href` redirect re-runs init, so the `deployment` super-property is re-registered on the next page load.

frontend/src/hooks/usePostHogEvents.js
```diff
@@ -57,27 +57,40 @@ const usePostHogEvents = () => {
...
+    const orgId = sessionDetails?.orgId;
     const orgName = sessionDetails?.orgName;
     const isOpenSource = orgName === "mock_org";
+    const software = isOpenSource ? "OSS" : "SAAS";
+    const email = sessionDetails?.email;
+    const userId = sessionDetails?.id;
+
+    // Email as distinct_id: bare user ids collide across regions/self-hosts
+    // since every deployment reports to the same PostHog project
+    const distinctId = email || `${orgId}:${userId}`;
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Fixed in 659dcb95a — fallback is now `orgId ? \`${orgId}:${userId}\` : userId?.toString()` as suggested.

frontend/src/index.jsx
```diff
@@ -9,18 +9,37 @@ import { SocketProvider } from "./helpers/SocketContext.js";
 import "./index.css";
 import config from "./config.js";
 
-const enablePosthog = import.meta.env.VITE_ENABLE_POSTHOG;
+// Runtime config (containerized deployments) takes priority over build-time
+// env so PostHog can be disabled per environment without a rebuild
+const runtimeConfig =
+  typeof window !== "undefined" ? window.RUNTIME_CONFIG || {} : {};
+const enablePosthog =
+  runtimeConfig.enablePosthog || import.meta.env.VITE_ENABLE_POSTHOG;
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Clarified the comment in 659dcb95a — it now states that the entrypoint emits an empty string for unset env vars and that the falsy fall-through to the build-time env is deliberate (which is also why `||` rather than `??`).

frontend/src/hooks/usePostHogEvents.js
```diff
@@ -57,27 +58,50 @@ const usePostHogEvents = () => {
...
-
-    if (email) {
-      optionalParams["email"] = email;
-    } else {
-      optionalParams["userId"] = userId;
+    try {
+      const orgId = sessionDetails?.orgId;
+      const orgName = sessionDetails?.orgName;
+      const isOpenSource = orgName === "mock_org";
+      const software = isOpenSource ? "OSS" : "SAAS";
```
athul-rs, line 65: `software` is still derived from `orgName === "mock_org"`, so an on-prem install with real auth emits `software="SAAS"` alongside `deployment="self-hosted"` — on-prem gets counted as SaaS by anyone segmenting on `software`. The PR positions `deployment` as the replacement for this exact `mock_org` check; intentional to keep `software` as-is for cohort continuity, or should it derive from `getDeployment()`?

chandrasekharan-zipstack: This is retaining the pre-existing behaviour to avoid increasing the scope of this PR. If separate analytics is required, this can be considered then

jaseemjaskp (approved):
Re-reviewed with the PR Review Toolkit (code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier). The substantive issues from the prior pass — PII-as-distinct_id gating, unguarded identify/register, bootstrap-abort risk, and the degenerate fallback id — are all addressed in the current code. Remaining observations are doc-wording, defensive-structure, magic-string, and test-coverage suggestions that don't rise to a blocking bar. LGTM.

athul-rs (approved):

PR #2050, opened 2026-06-12
UN-3515 UN-3514 [FIX] Project shares apply to exported Prompt Studio tools; share modal keeps org-share state

chandrasekharan-zipstack:
## What

- Exported Prompt Studio tools (`PromptStudioRegistry`) now derive their access from the linked Prompt Studio project's (`CustomTool`) **current** share state instead of the share snapshot taken at export time.
- Fixes, for a user the project was shared with *after* export:
  - Tool missing from the workflow tool list (raw UUID shown instead of the project name)
  - API deployment failing tool validation with `422: You don't have permission to perform this action.`

## Why

Sharing a Prompt Studio project only updates `CustomTool.shared_to_org` / `shared_users`. Every consumer of the exported tool (tool listing, `validate_tool_access`) gated on the separate `PromptStudioRegistry` row, whose share fields are written only at export time. So sharing an already-exported project never made its tool usable for the new users (same gap behind the UUID symptom in UN-2868).

## How

- `PromptStudioRegistryModelManager.list_tools`: filter on `custom_tool_id__in CustomTool.objects.for_user(user)`; registry rows without a `custom_tool` link (legacy data) fall back to the registry's own share fields.
- `ToolInstanceHelper.validate_tool_access`: delegate the Prompt Studio branch to `list_tools` so both read sites share one access rule.
- Semantic: content stays an export-time snapshot (prompts/spec/metadata unchanged until re-export); authorization is live. Sharing grants access immediately; unsharing revokes it immediately — including for previously-exported tools.

Two deliberate behavior changes, both consistent with `CustomTool.objects.for_user`:

1. Unsharing a project now revokes access to its exported tool immediately (previously the export-time share snapshot kept granting access). Workflows other users built on a since-unshared tool will fail validation for them — revoked means revoked.
2. Org admins now see all Prompt Studio tools in the tool list (previously only owned/shared ones), matching admin visibility of the projects themselves.

Everything else is covered by the legacy fallback (rows with `custom_tool = NULL` keep the old own-field checks).

## UN-3514: org-share toggle state lost in the share modal (frontend)

`useShareModal` rebuilt the modal item with only `shared_users`/`shared_groups` from the share-info response; `shared_to_org` fell back to the list row, which `APIDeploymentListSerializer` omits. The org-share toggle therefore rendered unchecked for org-shared API deployments, and a subsequent apply (replace-style share) silently revoked the org share despite a success toast. Fixed by sourcing `shared_to_org` from the same `/users/` response (1 line + comment).

Audited all six share surfaces: API deployments were broken; pipelines, workflows, adapters, Prompt Studio projects, and connectors already carry `shared_to_org` into the modal correctly. The hook fix also makes pipelines source the flag from fresh share info instead of the list row.

## Database Migrations

- None. No schema change; access is computed at read time, so existing affected rows in prod heal automatically — no backfill needed.

## Env Config

- None

## Relevant Docs

- N/A

## Related Issues or PRs

- UN-3515, UN-3514 (share-modal org toggle), UN-2868 (UUID symptom, same root cause)

## Dependencies Versions

- No changes

## Notes on Testing

- New: `backend/prompt_studio/tests/test_registry_share_derivation.py` (14 tests) covering share-after-export grant, unshare revoke, legacy NULL-link fallback, admin access, and snapshot-vs-project precedence — all green.
- `tenant_account_v2` sharing-matrix tests: 15/15 green.
- Manual repro: share an exported project (org or user) → the other user immediately sees the tool name in workflows and API deployment validation passes, without re-export.

## Screenshots

N/A (backend only)

## Checklist

### DIFF
--- a/backend/prompt_studio/prompt_studio_registry_v2/models.py
+++ b/backend/prompt_studio/prompt_studio_registry_v2/models.py
@@ -26,12 +26,23 @@ class PromptStudioRegistryModelManager(DefaultOrganizationManagerMixin, BaseMode
         return super().get_queryset()
 
     def list_tools(self, user: User) -> QuerySet[Any]:
+        # Access is derived from the linked CustomTool so that sharing a
+        # Prompt Studio project applies to its exported tool without a
+        # re-export. The registry's own share fields are consulted only for
+        # legacy rows that lack a custom_tool link.
+        visible_tools = CustomTool.objects.for_user(user).values("tool_id")
         return (
             self.get_queryset()
             .filter(
-                models.Q(created_by=user)
-                | models.Q(shared_users=user)
-                | models.Q(shared_to_org=True)
+                models.Q(custom_tool_id__in=visible_tools)
+                | (
+                    models.Q(custom_tool__isnull=True)
+                    & (
+                        models.Q(created_by=user)
+                        | models.Q(shared_users=user)
+                        | models.Q(shared_to_org=True)
+                    )
+                )
             )
             .distinct("prompt_registry_id")
         )
--- a/backend/prompt_studio/prompt_studio_registry_v2/models.py
+++ b/backend/prompt_studio/prompt_studio_registry_v2/models.py
@@ -5,6 +5,7 @@ from typing import Any
 from account_v2.models import User
 from django.db import models
 from django.db.models import QuerySet
+from tenant_account_v2.organization_member_service import OrganizationMemberService
 from utils.models.base_model import BaseModel, BaseModelManager
 from utils.models.organization_mixin import (
     DefaultOrganizationManagerMixin,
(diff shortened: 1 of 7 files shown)


chandrasekharan-zipstack:
Addressed review comments in 56a7583da:

- **CodeRabbit (legacy-row admin visibility, both comments):** `list_tools` now mirrors `CustomTool.objects.for_user` — admins and service accounts get the full queryset, so legacy (unlinked) rows are visible to them too. Added `test_admin_sees_unshared_and_legacy_rows` covering it.
- **Greptile (redundant fetch + double admin lookup):** replaced `.get(pk=...)` with `.filter(pk=...).exists()` and dropped the outer `is_admin or` for the Prompt Studio branch — the admin check is resolved once inside `list_tools`; the agentic branch computes it only when reached.

backend/prompt_studio/prompt_studio_registry_v2/models.py
```diff
@@ -26,12 +27,29 @@ def get_queryset(self) -> QuerySet[Any]:
...
+        # sharing applies without re-export. Unlinked legacy rows fall back
+        # to their own share fields.
+        visible_tools = CustomTool.objects.for_user(user).values("tool_id")
         return (
             self.get_queryset()
             .filter(
-                models.Q(created_by=user)
-                | models.Q(shared_users=user)
-                | models.Q(shared_to_org=True)
+                models.Q(custom_tool_id__in=visible_tools)
```
muhammad-ali-e, line 44: check the possibility of indexing these  according to the frequency of this call and time taken,

muhammad-ali-e (approved):

backend/prompt_studio/prompt_studio_registry_v2/models.py
chandrasekharan-zipstack: Good call. This runs on an org-scoped table — `organization_id` is always filtered and FK-indexed, so `list_tools` only ever scans one org's exported tools (a small set, one row per exported tool). The hot `custom_tool_id` branch already uses the OneToOne unique index and `shared_users` uses the M2M FK indexes; `shared_to_org`/`created_by` only touch that tiny per-org slice. So no index needed at current scale — I'd rather not add a speculative one. Logged as a measure-if-it-shows-in-slow-queries follow-up.

kirtimanmishrazipstack (approved):

PR #2051, opened 2026-06-12
UN-3544 [FEAT] PG-queue consumer liveness endpoint (poll-loop heartbeat)

muhammad-ali-e:
Gives the PG-queue consumer a `/health` HTTP endpoint for K8s liveness probing — the same pattern every other worker exposes (8080–8089) — keyed on a **poll-loop heartbeat**. Targets `feat/UN-3445-pg-queue-integration`.

## Why
The consumer had no health server. This closes the recurring review thread from #2047 (toolkit "opt-in-hides-crash" steady-state gap + Greptile "no health port") *properly*: the heartbeat detects a **stuck/wedged poll loop** (process alive but not draining) — which neither `run-worker.sh --status` (pgrep) nor the launch-liveness check can see. It also pre-positions the K8s rollout, where a liveness probe needs a known endpoint.

## What
- **`consumer.py`** — `_last_poll_monotonic` heartbeat, refreshed at the *top* of `poll_once` (so a loop wedged on a long-running task goes stale). `seconds_since_last_poll()` / `is_poll_stale()`. `main()` starts a `LivenessServer` when `WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT` is set (opt-in), stops it on exit.
- **`LivenessServer`** — tiny stdlib HTTP server (`/health`, `/healthz`, `/livez`) → **200 while fresh, 503 once stale**. Deliberately **lean**: a liveness probe must report only "is this process making progress?", **not** broker/API reachability or resource pressure — coupling those would crash-loop a healthy consumer on a transient blip.
- **`run-worker.sh`** — default port **8090** (outside the 8080–8089 core range → no collision), exported opt-in, documented in `--help`. Bare `python -m pg_queue_consumer` binds nothing.

## Design notes
- **Fixed-but-configurable port (not dynamic):** a probe must hit a *known* port; a dynamically-chosen one can't be probed. In K8s each container has its own network namespace, so 8090 is per-container (no host collision); the Deployment sets the probe port. Overridable locally via the env / `-p`.
- **Why not reuse the shared `HealthChecker`/`HealthServer`:** besides the liveness-semantics point above, its built-in `api_connectivity` check has a real import bug — `from .api_client_singleton` resolves to a non-existent `shared.infrastructure.monitoring.api_client_singleton` (the module is at `shared.utils.api_client_singleton`), so it returns 503 for any checker without a pre-set `api_client`. Reusing it would ship a broken endpoint. Filed separately (fleet-wide).

## Testing
- **Unit:** heartbeat fresh/stale + `is_poll_stale` threshold + a **real bind-and-GET test** asserting `200` (healthy) then `503` (forced stale). 23/23 non-integration tests green.
- **Live:** `curl :8090/health` → `200 {"status":"healthy","check":"pg_queue_poll",...}`; unknown path → 404; no port set → no server.

## Out of scope
- The K8s Deployment manifest + probe wiring (rollout) — this provides the endpoint they'll point at.
- The shared `health.py` `api_connectivity` import bug (separate ticket).

### DIFF
--- a/workers/queue_backend/pg_queue/consumer.py
+++ b/workers/queue_backend/pg_queue/consumer.py
@@ -25,7 +25,7 @@ import os
 import signal
 import time
 from collections.abc import Callable
-from typing import TYPE_CHECKING, Any, TypeVar
+from typing import TYPE_CHECKING, TypeVar
 
 from celery import current_app
 
@@ -33,6 +33,9 @@ from ..fairness import FAIRNESS_HEADER_NAME
 from .client import PgQueueClient
 
 if TYPE_CHECKING:
+    from http.server import HTTPServer
+    from threading import Thread
+
     from celery import Celery
 
     from .client import QueueMessage
@@ -208,6 +211,15 @@ class PgQueueConsumer:
         liveness probe should report unhealthy and let the orchestrator restart
         it. Pick a threshold comfortably above ``backoff_max`` and the longest
         expected task so normal idle/backoff never trips it.
+
+        Note the heartbeat is stamped at the *top* of ``poll_once`` (before the
+        DB read), so a loop that fails fast every cycle — e.g. ``read()`` raising
+        on an unreachable DB, caught and backed off by ``run()`` — keeps stamping
+        and stays *healthy*. That is deliberate: a liveness probe must not couple
+        to backend reachability (a restart can't fix a DB outage, and coupling
+        would crash-loop every consumer during one). Surfacing a permanent
+        config fault (bad creds, missing schema) is a readiness/alerting concern,
+        not liveness.
         """
         return self.seconds_since_last_poll() > stale_after_seconds
 
@@ -291,7 +303,9 @@ def main() -> None:
         # offending var name instead of a context-free `int('abc')` ValueError.
         var = f"WORKER_PG_QUEUE_CONSUMER_{suffix}"
         raw = os.getenv(var)
-        if raw is None:
+        # Treat empty-string as unset: an empty HEALTH_PORT (e.g. a run-worker.sh
+        # fallback resolving empty) must hit the clean opt-out, not int("") crash.
+        if raw is None or raw == "":
             return default
         try:
             return cast(raw)
@@ -321,11 +335,12 @@ def main() -> None:
 def _maybe_start_health_server(
     consumer: PgQueueConsumer, *, port: int | None, stale_after: float
 ) -> LivenessServer | None:
-    """Start the liveness server when a port is configured; else ``None``.
+    """Start the liveness server when ``port`` is not None; else ``None``.
 
-    Opt-in by ``WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT`` — a bare
-    ``python -m pg_queue_consumer`` with no port set binds nothing (no stray
-    port to collide).
+    ``main()`` wires ``port`` from ``WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT`` (unset
+    → ``None`` → no server, no stray port). A bind failure degrades gracefully:
+    the probe is auxiliary, so it must never stop the consumer from draining the
+    queue — we log and continue probe-less rather than abort startup.
     """
     if port is None:
         logger.info(
@@ -334,7 +349,16 @@ def _maybe_start_health_server(
         )
         return None
     server = LivenessServer(consumer, port=port, stale_after=stale_after)
-    server.start()
+    try:
+        server.start()
+    except OSError as exc:
+        logger.error(
+            "PG-queue consumer: liveness server could not bind :%s (%s) — "
+            "continuing WITHOUT a probe",
+            port,
+            exc,
+        )
+        return None
     logger.info(
         "PG-queue consumer: liveness server on :%s/health (stale after %ss)",
         server.bound_port,
@@ -355,9 +379,11 @@ class LivenessServer:
     checks meant for richer health reporting, not liveness) — it reports solely
     on the poll-loop heartbeat (:meth:`PgQueueConsumer.is_poll_stale`).
 
-    Serves ``/health`` (also ``/healthz``, ``/livez``) in a daemon thread.
-    Bind ``port=0`` to let the OS pick a free port (read back via
-    :attr:`bound_port`) — used in tests.
+    Serves ``/health`` (also ``/healthz``, ``/livez``) on ``0.0.0.0`` (all
+    interfaces — a container/k8s probe reaches it from outside the process) in a
+    daemon thread. Bind ``port=0`` to let the OS pick a free port (read back via
+    :attr:`bound_port`) — used in tests. Start once; :meth:`stop` returns it to
+    the inert state.
     """
 
     _PATHS = frozenset({"/health", "/healthz", "/livez"})
@@ -368,13 +394,16 @@ class LivenessServer:
         self._consumer = consumer
         self._port = port
         self._stale_after = stale_after
-        self._httpd: Any = None
-        self._thread: Any = None
+        self._httpd: HTTPServer | None = None
+        self._thread: Thread | None = None
 
     def start(self) -> None:
         import json
         import threading
         from http.server import BaseHTTPRequestHandler, HTTPServer
+
+        if self._httpd is not None:
+            raise RuntimeError("LivenessServer already started")
 
         consumer = self._consumer
         stale_after = self._stale_after
@@ -386,8 +415,10 @@ class LivenessServer:
                     self.send_response(404)
                     self.end_headers()
                     return
+                # One clock read so age and the healthy/stale verdict are
+                # derived from the same instant.
                 age = consumer.seconds_since_last_poll()
-                stale = consumer.is_poll_stale(stale_after)
+                stale = age > stale_after
                 body = json.dumps(
                     {
                         "status": "unhealthy" if stale else "healthy",
@@ -399,32 +430,58 @@ class LivenessServer:
                 self.send_response(503 if stale else 200)
                 self.send_header("Content-Type", "application/json")
                 self.end_headers()
-                self.wfile.write(body)
+                try:
+                    self.wfile.write(body)
+                except (BrokenPipeError, ConnectionResetError):
+                    pass  # client (probe) hung up mid-response — not our problem
 
             def log_message(self, *_: object) -> None:
                 pass  # silence per-request access logging
 
-        self._httpd = HTTPServer(("0.0.0.0", self._port), _Handler)
+            def log_error(self, fmt: str, *args: object) -> None:
+                # BaseHTTPRequestHandler routes errors through log_message too;
+                # don't let the pass above swallow them — surface to our logger.
+                logger.warning("pg-queue liveness handler: " + fmt, *args)
+
+        def _serve(httpd: HTTPServer) -> None:
+            try:
+                httpd.serve_forever()
+            except Exception:
+                # A daemon thread dying silently would make /health stop
+                # answering (connection refused) with no breadcrumb.
+                logger.exception("pg-queue liveness server thread crashed")
+
+        httpd = HTTPServer(("0.0.0.0", self._port), _Handler)
+        self._httpd = httpd
         self._thread = threading.Thread(
-            target=self._httpd.serve_forever,
-            daemon=True,
-            name="pg-consumer-liveness",
+            target=_serve, args=(httpd,), daemon=True, name="pg-consumer-liveness"
         )
         self._thread.start()
 
     @property
     def bound_port(self) -> int:
-        """The actual listening port (resolves ``port=0`` to the OS choice)."""
+        """Actual listening port (resolves ``port=0``); the requested port if not started."""
         if self._httpd is not None:
             return self._httpd.server_address[1]
         return self._port
 
     def stop(self) -> None:
-        if self._httpd is not None:
-            self._httpd.shutdown()
-            self._httpd.server_close()
-        if self._thread is not None:
-            self._thread.join(timeout=5)
+        """Shut the server down. Defensive: never raises (called from a finally)."""
+        try:
+            if self._httpd is not None:
+                self._httpd.shutdown()
+                self._httpd.server_close()
+            if self._thread is not None:
+                self._thread.join(timeout=5)
+                if self._thread.is_alive():
+                    logger.warning(
+                        "PG-queue consumer: liveness thread did not stop within 5s"
+                    )
+        except Exception:
+            logger.exception("PG-queue consumer: error stopping liveness server")
+        finally:
+            self._httpd = None
+            self._thread = None
 
 
 if __name__ == "__main__":
--- a/workers/queue_backend/pg_queue/consumer.py
+++ b/workers/queue_backend/pg_queue/consumer.py
@@ -351,12 +351,11 @@ def _maybe_start_health_server(
     server = LivenessServer(consumer, port=port, stale_after=stale_after)
     try:
         server.start()
-    except OSError as exc:
-        logger.error(
-            "PG-queue consumer: liveness server could not bind :%s (%s) — "
+    except OSError:
+        logger.exception(
+            "PG-queue consumer: liveness server could not bind :%s — "
             "continuing WITHOUT a probe",
             port,
-            exc,
         )
         return None
     logger.info(
--- a/workers/queue_backend/pg_queue/consumer.py
+++ b/workers/queue_backend/pg_queue/consumer.py
@@ -25,7 +25,7 @@ import os
 import signal
 import time
 from collections.abc import Callable
-from typing import TYPE_CHECKING, TypeVar
+from typing import TYPE_CHECKING, Any, TypeVar
 
 from celery import current_app
 
@@ -53,6 +53,10 @@ _DEFAULT_BACKOFF_MAX = 2.0
 # A task claimed more than this many times keeps failing — drop it (poison)
 # rather than redeliver forever.
 _DEFAULT_MAX_ATTEMPTS = 5
+# Liveness: a poll loop that hasn't cycled in this many seconds is reported
+# unhealthy. Well above _DEFAULT_BACKOFF_MAX (idle backoff) and any sane task
+# duration, so only a genuinely wedged loop trips it.
+_DEFAULT_HEALTH_STALE_SECONDS = 60.0
 
 
 class PgQueueConsumer:
@@ -97,9 +101,16 @@ class PgQueueConsumer:
         self.backoff_max = backoff_max
         self.max_attempts = max_attempts
         self._running = False
+        # Heartbeat for the liveness probe: monotonic timestamp of the most
+        # recent poll attempt. Seeded at construction so a just-started consumer
+        # reads healthy. Updated at the TOP of poll_once, so a loop wedged on a
+        # long-running task (poll_once not returning) goes stale and is caught —
+        # something pgrep-based --status and the launch-time check cannot see.
+        self._last_poll_monotonic = time.monotonic()
 
     def poll_once(self) -> int:
         """Claim + process one batch; returns the number of messages claimed."""
+        self._last_poll_monotonic = time.monotonic()
         messages = self._client.read(
             self.queue_name, vt_seconds=self.vt_seconds, qty=self.batch_size
         )
@@ -184,6 +195,21 @@ class PgQueueConsumer:
     def _registered_task_count(self) -> int:
         """Count application tasks (excluding Celery's built-ins)."""
         return sum(1 for name in self._app.tasks if not name.startswith("celery."))
+
+    def seconds_since_last_poll(self) -> float:
+        """Seconds since the last poll attempt (for the liveness heartbeat)."""
+        return time.monotonic() - self._last_poll_monotonic
+
+    def is_poll_stale(self, stale_after_seconds: float) -> bool:
+        """True if the poll loop hasn't cycled within ``stale_after_seconds``.
+
+        Drives the health endpoint: a stale loop means the consumer is wedged
+        (deadlock, or a single task running longer than the threshold), so the
+        liveness probe should report unhealthy and let the orchestrator restart
+        it. Pick a threshold comfortably above ``backoff_max`` and the longest
+        expected task so normal idle/backoff never trips it.
+        """
+        return self.seconds_since_last_poll() > stale_after_seconds
 
     def run(self, *, install_signals: bool = True, require_tasks: bool = True) -> None:
         """Poll loop with empty-queue backoff and graceful shutdown.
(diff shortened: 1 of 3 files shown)


workers/queue_backend/pg_queue/consumer.py
```diff
@@ -272,14 +298,133 @@ def _env(suffix: str, default: _T, cast: Callable[[str], _T]) -> _T:
...
+    port to collide).
+    """
+    if port is None:
+        logger.info(
+            "PG-queue consumer: WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT unset — "
+            "liveness server disabled"
+        )
+        return None
+    server = LivenessServer(consumer, port=port, stale_after=stale_after)
+    server.start()
```
muhammad-ali-e: **[Important · Silent failure]** A liveness-port bind failure aborts the consumer entirely.

`server.start()` constructs `HTTPServer(("0.0.0.0", self._port), _Handler)` synchronously (line 407) with no error handling. If the port is already bound (`OSError: [Errno 98] Address already in use` — plausible since 8090 is a fixed default, a restart races a leftover process, or the pluggable-worker collision below), the `OSError` propagates out of `_maybe_start_health_server` → out of `main()`, and **the consumer never starts polling**. An auxiliary liveness probe should never prevent the worker from draining the queue. Note this also diverges from the existing `shared/infrastructure/monitoring/health.py` `HealthServer.start()`, which wraps the bind in try/except.

**Fix:** wrap the bind and degrade gracefully:
```python
try:
    server.start()
except OSError as exc:
    logger.error(
        "PG-queue consumer: liveness server could not bind :%s (%s) — "
        "continuing WITHOUT a probe", port, exc,
    )
    return None
```
If fail-fast is genuinely intended, raise a `RuntimeError` naming the port and env var instead of a bare `OSError`.

workers/run-worker.sh
```diff
@@ -84,9 +84,10 @@ declare -A WORKER_HEALTH_PORTS=(
     ["scheduler"]="8087"
     ["${EXECUTOR_WORKER_TYPE}"]="8088"
     ["${IDE_CALLBACK_WORKER_TYPE}"]="8089"
-    # pg_queue_consumer: no entry — it runs no health server, so it binds no
-    # port (avoids a hard-coded port that could collide). A liveness endpoint,
-    # if added later, should read WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT.
+    # pg_queue_consumer: 8090 — outside the 8080-8089 core range, so no
+    # collision. The consumer binds it only when WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT
+    # is exported (below); a bare `python -m pg_queue_consumer` binds nothing.
+    ["$PG_QUEUE_CONSUMER_TYPE"]="8090"
```
muhammad-ali-e, line 92: **[Important · Code review]** Latent port collision, and the comment above is inaccurate.

This statically assigns `8090` to `pg_queue_consumer`, but the pluggable-worker discovery at line 279 allocates `$((8090 + discovered_count))` — so the **first** discovered pluggable worker also gets `8090`. The guard at line 278 only checks whether the pluggable name key is unset; it does not check that the port *value* `8090` is already taken. If a `pg_queue_consumer` and the first pluggable worker run on the same host, both try to bind `8090`. The new comment ("8090 — outside the 8080-8089 core range, so no collision") accounts for the static core range but not the dynamic pluggable allocation that starts at the same 8090.

**Fix:** start pluggable discovery at `8091` (`$((8091 + discovered_count))`), or reserve the consumer outside the pluggable range, and correct the comment to acknowledge the pluggable range.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -272,14 +298,133 @@ def _env(suffix: str, default: _T, cast: Callable[[str], _T]) -> _T:
...
+        self._thread.start()
+
+    @property
+    def bound_port(self) -> int:
+        """The actual listening port (resolves ``port=0`` to the OS choice)."""
+        if self._httpd is not None:
+            return self._httpd.server_address[1]
+        return self._port
+
+    def stop(self) -> None:
```
muhammad-ali-e, line 475: **[Important · Silent failure]** `stop()` in the `finally` block (main(), lines 316-318) can mask the real exception from `consumer.run()`.

If `run()` raises (e.g. the `require_tasks` `RuntimeError`) and `stop()` *also* raises here (`shutdown()`/`server_close()`/`join()` on a server in an unexpected state), the cleanup exception **replaces** the original — the operator debugs a `shutdown()` error instead of why the consumer actually died. Also, `self._thread.join(timeout=5)` ignores its result: if the thread is still alive after 5s, no one is told.

**Fix:** make `stop()` defensive so it can't raise into the `finally`:
```python
def stop(self) -> None:
    try:
        if self._httpd is not None:
            self._httpd.shutdown()
            self._httpd.server_close()
        if self._thread is not None:
            self._thread.join(timeout=5)
            if self._thread.is_alive():
                logger.warning("liveness thread did not stop within 5s")
    except Exception:
        logger.exception("error stopping liveness server")
```

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -272,14 +298,133 @@ def _env(suffix: str, default: _T, cast: Callable[[str], _T]) -> _T:
...
         batch_size=_env("BATCH", _DEFAULT_BATCH, int),
         vt_seconds=_env("VT_SECONDS", _DEFAULT_VT_SECONDS, int),
         poll_interval=_env("POLL_INTERVAL", _DEFAULT_POLL_INTERVAL, float),
         backoff_max=_env("BACKOFF_MAX", _DEFAULT_BACKOFF_MAX, float),
         max_attempts=_env("MAX_ATTEMPTS", _DEFAULT_MAX_ATTEMPTS, int),
-    ).run()
+    )
+    health_server = _maybe_start_health_server(
+        consumer,
+        port=_env("HEALTH_PORT", None, int),
```
muhammad-ali-e, line 330: **[Important · Silent failure]** An *empty* `WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT` crashes the consumer instead of disabling the probe.

The "disabled" path (line 330, `if port is None`) requires the env var to be **unset**. An empty string (`export WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT=""`) makes `os.getenv` return `""` (not `None`), so `_env("HEALTH_PORT", None, int)` runs `int("")` → `ValueError` → `main()` dies at launch. This is reachable if the run-worker.sh fallback `${health_port:-${WORKER_HEALTH_PORTS[$worker_type]}}` (line 712) ever resolves empty, or if an operator sets it empty by hand.

**Fix:** treat empty-string as unset in `_env` so the clean opt-out is hit:
```python
raw = os.getenv(var)
if raw is None or raw == "":
    return default
```

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -97,9 +101,16 @@ def __init__(
...
+        # Heartbeat for the liveness probe: monotonic timestamp of the most
+        # recent poll attempt. Seeded at construction so a just-started consumer
+        # reads healthy. Updated at the TOP of poll_once, so a loop wedged on a
+        # long-running task (poll_once not returning) goes stale and is caught —
+        # something pgrep-based --status and the launch-time check cannot see.
+        self._last_poll_monotonic = time.monotonic()
 
     def poll_once(self) -> int:
         """Claim + process one batch; returns the number of messages claimed."""
+        self._last_poll_monotonic = time.monotonic()
```
muhammad-ali-e, line 121: **[Suggestion · Design check]** The heartbeat is stamped at the *top* of `poll_once`, before `self._client.read(...)`. This correctly catches a loop wedged *inside* a long task (the documented rationale). But it also means a poll loop that **fails fast every cycle** — e.g. `read()` raising `OperationalError` on a dead DB, caught by `run()`'s `except`, backed off, retried — keeps re-stamping line 113 and so reports **healthy** while doing zero useful work.

For a pure *liveness* probe that may be intentional (the process is alive and looping; restarting won't fix a backend outage, and the docstring explicitly avoids coupling to backend reachability). But a *permanent* fault (bad credentials, missing schema) would then stay green forever and never trip the probe. Please confirm this matches intent — if a perpetually-erroring loop should be considered unhealthy, stamp the heartbeat only after a successful `read()` instead.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -272,14 +298,133 @@ def _env(suffix: str, default: _T, cast: Callable[[str], _T]) -> _T:
...
+    _PATHS = frozenset({"/health", "/healthz", "/livez"})
+
+    def __init__(
+        self, consumer: PgQueueConsumer, *, port: int, stale_after: float
+    ) -> None:
+        self._consumer = consumer
+        self._port = port
+        self._stale_after = stale_after
+        self._httpd: Any = None
+        self._thread: Any = None
```
muhammad-ali-e: **[Suggestion · Type design]** `self._httpd: Any` / `self._thread: Any` erase all type information and silently disable type-checking on every downstream use (`shutdown()`, `server_address[1]`, `serve_forever`, `join()`). The file has `from __future__ import annotations`, so precise optional types cost nothing at runtime and need no eager import:
```python
if TYPE_CHECKING:
    from http.server import HTTPServer
    from threading import Thread
...
self._httpd: "HTTPServer | None" = None
self._thread: "Thread | None" = None
```
This restores static checking and lets the lazy imports in `start()` stay. (This also drops the new `Any` import added at line 28.)

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -272,14 +298,133 @@ def _env(suffix: str, default: _T, cast: Callable[[str], _T]) -> _T:
...
+
+    def start(self) -> None:
```
muhammad-ali-e, line 404: **[Suggestion · Type design / lifecycle]** `LivenessServer` doesn't enforce its start-once invariant. A second `start()` overwrites `_httpd`/`_thread` without closing the first socket or joining the first thread — leaking a bound socket and a live daemon thread. Two cheap hardenings:

- Guard double start: `if self._httpd is not None: raise RuntimeError("already started")`.
- Reset `self._httpd = None` / `self._thread = None` at the end of `stop()` so a stopped server returns to the inert state and `bound_port` correctly reflects "not running".

(Related: `bound_port` read before `start()` with `port=0` returns a meaningless `0` — fine if documented, but the pre-start contract is currently implicit.)

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -272,14 +298,133 @@ def _env(suffix: str, default: _T, cast: Callable[[str], _T]) -> _T:
...
+                        "seconds_since_last_poll": round(age, 3),
+                        "stale_after_seconds": stale_after,
+                    }
+                ).encode()
+                self.send_response(503 if stale else 200)
+                self.send_header("Content-Type", "application/json")
+                self.end_headers()
+                self.wfile.write(body)
+
+            def log_message(self, *_: object) -> None:
```
muhammad-ali-e, line 445: **[Suggestion · Silent failure]** Overriding `log_message` to `pass` silences per-request *access* logs (good) — but `BaseHTTPRequestHandler.log_error` also delegates to `log_message`, so handler **errors** are silenced too. If `do_GET` ever raises (e.g. `BrokenPipeError`/`ConnectionResetError` on `self.wfile.write`, or an unexpected exception), the failure is completely invisible while an orchestrator may be crash-looping the pod on failed probes.

**Fix:** keep access logs quiet but route real errors to the project logger, and guard the write against client disconnects:
```python
def log_error(self, fmt: str, *args: object) -> None:
    logger.warning("pg-queue liveness handler: " + fmt, *args)
```
```python
try:
    self.wfile.write(body)
except (BrokenPipeError, ConnectionResetError):
    pass  # client hung up
```

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -272,14 +298,133 @@ def _env(suffix: str, default: _T, cast: Callable[[str], _T]) -> _T:
...
+                pass  # silence per-request access logging
+
+        self._httpd = HTTPServer(("0.0.0.0", self._port), _Handler)
+        self._thread = threading.Thread(
+            target=self._httpd.serve_forever,
```
muhammad-ali-e: **[Suggestion · Silent failure]** The liveness daemon thread runs `self._httpd.serve_forever` with no error boundary. If `serve_forever` ever exits abnormally, the daemon thread dies with only the default `threading` excepthook — no `logger` line, no health feedback — and `/health` silently stops responding (connection refused), which an orchestrator may treat differently from a 503. Wrap the target so a crash is logged loudly:
```python
def _serve() -> None:
    try:
        self._httpd.serve_forever()
    except Exception:
        logger.exception("pg-queue liveness server thread crashed")
# target=_serve
```

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -272,14 +298,133 @@ def _env(suffix: str, default: _T, cast: Callable[[str], _T]) -> _T:
...
+        paths = self._PATHS
+
+        class _Handler(BaseHTTPRequestHandler):
+            def do_GET(self) -> None:
+                if self.path not in paths:
+                    self.send_response(404)
+                    self.end_headers()
+                    return
+                age = consumer.seconds_since_last_poll()
+                stale = consumer.is_poll_stale(stale_after)
```
muhammad-ali-e: **[Suggestion · Simplify]** `is_poll_stale(stale_after)` internally calls `seconds_since_last_poll()` again, so `time.monotonic()` is read twice per request and `age`/`stale` can derive from two different timestamps. Harmless, but you already have `age`:
```python
age = consumer.seconds_since_last_poll()
stale = age > stale_after
```
One clock read, self-consistent. (Keep `is_poll_stale` as the single comparison source elsewhere — this is only about the redundant second read in the hot path.)

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -272,14 +298,133 @@ def _env(suffix: str, default: _T, cast: Callable[[str], _T]) -> _T:
...
+        consumer.run()
+    finally:
+        if health_server is not None:
+            health_server.stop()
+
+
+def _maybe_start_health_server(
+    consumer: PgQueueConsumer, *, port: int | None, stale_after: float
+) -> LivenessServer | None:
+    """Start the liveness server when a port is configured; else ``None``.
```
muhammad-ali-e: **[Nit · Comments]** This docstring ties the helper's contract to a caller detail one layer up (it names `WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT`, but this function only sees a generic `port: int | None`; the env-var→port mapping lives in `main()`). A future second caller would silently make the docstring wrong. Consider phrasing the trigger in terms of the parameter: *"Start the liveness server when `port` is not None; else None. `main()` wires `port` from `WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT`."* Separately, the `LivenessServer` binds `0.0.0.0` (all interfaces, line 407) — worth one line in its docstring since the collision-avoidance framing might imply otherwise.

workers/run-worker.sh
```diff
@@ -705,6 +707,9 @@ run_worker() {
         # the worker that owns the first migrated leaf task,
         # send_webhook_notification; override to drain another worker's queue.
         export WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE="${WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE:-notification}"
+        # Liveness HTTP server port (-p override wins, else the map default).
+        # Exported so the launcher's main() opts into the health server.
+        export WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT="${health_port:-${WORKER_HEALTH_PORTS[$worker_type]}}"
```
muhammad-ali-e, line 716: **[Suggestion · Code review]** This export honours the `-p`/`--health-port` override (`${health_port:-...}`), but the status line printed at line 720 (`Health Port: ${WORKER_HEALTH_PORTS[$worker_type]:-n/a}`) always shows the *map default* `8090`. Running `pg-queue-consumer -p 9999` binds 9999 but reports `Health Port: 8090` — misleading during ops. Consider printing the effective exported value for this worker type.

workers/tests/test_pg_queue_consumer.py
```diff
@@ -266,6 +266,66 @@ def _demo():
...
+        import json
+        import urllib.request
+
+        from queue_backend.pg_queue.consumer import LivenessServer
+
+        consumer = PgQueueConsumer("q", client=MagicMock())
+        server = LivenessServer(consumer, port=0, stale_after=60)
+        server.start()
+        try:
+            url = f"http://127.0.0.1:{server.bound_port}/health"
```
muhammad-ali-e, line 330: **[Suggestion · Tests]** The real-socket test only hits `/health`. Two cheap, high-value additions that reuse this exact server setup:

1. **Unknown path → 404** (line 385 branch is entirely uncovered): assert `/nope` raises `HTTPError` with `code == 404`. Guards against a regression that makes every path 200 (a misconfigured probe would then always pass).
2. **`/healthz` and `/livez` aliases** (line 363 defines three, only one is tested): parametrize over all three asserting 200. These aliases exist precisely so different orchestrators can probe — the most likely deployment wiring and the most embarrassing to silently 404.

workers/tests/test_pg_queue_consumer.py
```diff
@@ -266,6 +266,66 @@ def _demo():
         )
 
 
+# --- Liveness heartbeat (drives the health endpoint) ---
+
+
+class TestPollHeartbeat:
+    def test_poll_once_refreshes_heartbeat(self):
```
muhammad-ali-e, line 273: **[Suggestion · Tests]** `test_poll_once_refreshes_heartbeat` proves the heartbeat resets on a *completed* poll — but that would pass equally if the timestamp were stamped at the *bottom* of `poll_once`. It therefore does **not** protect the headline design decision (stamp at the TOP, so a task running longer than the threshold still trips the probe). Add a test that captures the heartbeat *during* `read()`:
```python
def test_heartbeat_stamped_before_read(self):
    consumer = PgQueueConsumer("q", client=(c := MagicMock()))
    before = consumer._last_poll_monotonic
    seen = {}
    c.read.side_effect = lambda *a, **k: seen.setdefault("t", consumer._last_poll_monotonic) or []
    consumer.poll_once()
    assert seen["t"] > before  # refreshed BEFORE read ran, not after
```
Minor: the 200→503 block (lines 319-324) reads cleaner as `with pytest.raises(urllib.error.HTTPError) as ei: ...; assert ei.value.code == 503`.

muhammad-ali-e (commented):
## PR Review Toolkit — Summary (UN-3544 PG-queue consumer liveness)

Ran 6 specialised agents (Code Reviewer, Silent Failure Hunter, Type Design, Test Analyzer, Comment Analyzer, Code Simplifier) over the 3-file diff. **14 inline findings** posted. No correctness bug that risks data loss; the heartbeat/clock design is sound (monotonic, GIL-atomic, signals install fine on the main thread). Highlights:

**Important (4)** — fix before merge:
- Liveness bind failure (`OSError`/port-in-use) aborts the whole consumer before it polls — an auxiliary probe shouldn't kill the worker. (consumer.py:337)
- Static port `8090` collides with the first auto-discovered pluggable worker (`8090 + discovered_count`); the new comment is inaccurate. (run-worker.sh:90)
- `stop()` in the `finally` can mask the real `run()` exception. (consumer.py:422)
- Empty `WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT` crashes at launch instead of disabling the probe. (consumer.py:311)

**Suggestions (10)** — error logging in the HTTP handler/thread, `Any`-typed attrs defeating type-checking, double `start()` leak, double clock read, status-line shows map default not `-p`, a design check on the heartbeat staying green during a fast-failing loop, plus 2 test gaps (404 + `/healthz`/`/livez` aliases, and a test that actually pins the top-of-poll stamp).

Test additions are solid (real-socket 200→503, construction-time seeding, threshold boundaries). Comments are unusually accurate — every load-bearing claim verified true.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in ea0474cc2. `_maybe_start_health_server` wraps `start()` in `try/except OSError` → logs and returns None (consumer keeps draining probe-less). Verified live: a 2nd consumer forced onto a taken :8090 logged "could not bind … continuing WITHOUT a probe" and still started polling.

workers/run-worker.sh
muhammad-ali-e: Fixed in ea0474cc2. Pluggable discovery now starts at `8091 + discovered_count` (8090 reserved for the consumer), and the comments on both sides now reference the pluggable range.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in ea0474cc2. `stop()` is now wrapped in try/except (logs, never raises into the `finally`), resets `_httpd`/`_thread` to None, and warns if the thread is still alive after the 5s join.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in ea0474cc2. `_env` now treats `""` as unset (`if raw is None or raw == "": return default`), so an empty HEALTH_PORT hits the clean opt-out instead of `int("")` crashing.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Confirmed intent (doc clarified in ea0474cc2). Stamp-at-top + stays-green-on-fast-fail is deliberate: a liveness probe must NOT couple to backend reachability — restarting cannot fix a DB outage, and coupling would crash-loop every consumer during one. A permanent config fault (bad creds/missing schema) is a readiness/alerting concern, not liveness. Added that rationale to `is_poll_stale`'s docstring.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in ea0474cc2. Typed as `HTTPServer | None` / `Thread | None` via a `TYPE_CHECKING` import; dropped the `Any` import.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in ea0474cc2. `start()` raises `RuntimeError("already started")` if already running; `stop()` resets `_httpd`/`_thread` to None so a stopped server is inert and restartable. Added `test_double_start_is_rejected`.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in ea0474cc2. Added a `log_error` override routing to the project logger (so the `log_message=pass` no longer swallows handler errors), and guarded `wfile.write` against `BrokenPipeError`/`ConnectionResetError`.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in ea0474cc2. The thread target is wrapped so an abnormal `serve_forever` exit is logged via `logger.exception` instead of dying silently.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in ea0474cc2. Handler now reads the clock once (`age = seconds_since_last_poll(); stale = age > stale_after`) so the value and verdict are self-consistent.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in ea0474cc2. Docstring rephrased in terms of the `port` param (with a note that `main()` wires it from the env var), and the `0.0.0.0` all-interfaces bind is now documented on the class.

workers/run-worker.sh
muhammad-ali-e: Fixed in ea0474cc2. The status line now prints `${health_port:-${WORKER_HEALTH_PORTS[...]}}`, so a `-p` override is shown instead of the map default.

workers/tests/test_pg_queue_consumer.py
muhammad-ali-e: Fixed in ea0474cc2. Added `test_liveness_aliases_and_unknown_path` — all three aliases (`/health`, `/healthz`, `/livez`) → 200, unknown path → 404.

workers/tests/test_pg_queue_consumer.py
muhammad-ali-e: Fixed in ea0474cc2. Added `test_heartbeat_stamped_before_read` (captures the stamp during `read()` to pin top-of-poll), and switched the 503 assertion to the `pytest.raises(...) as ei` form.

muhammad-ali-e:
## Review feedback addressed — `ea0474cc2`

Thanks for the thorough pass — all 14 findings handled, threads replied inline.

**Important (4) — all fixed**
- 🔴 Bind failure no longer aborts the consumer — degrades to probe-less and keeps draining (verified live: 2nd consumer on a taken :8090 logs the failure and still polls).
- 🔴 Port collision — pluggable discovery starts at `8091` now; `8090` reserved for the consumer.
- 🔴 `stop()` is defensive (can't mask `run()`'s exception in the `finally`) + warns on a lingering thread.
- 🔴 Empty `HEALTH_PORT` hits the clean opt-out (`_env` treats `""` as unset) instead of crashing.

**Suggestions (10) — all addressed**
- LivenessServer hardening: double-start guard + reset-on-stop, `serve_forever` crash logged, `log_error` routed to the logger (was hidden by `log_message=pass`), `wfile.write` guarded against client disconnects, single clock read.
- Types: `_httpd`/`_thread` → `HTTPServer|None` / `Thread|None` (dropped `Any`).
- run-worker.sh: status line shows the effective `-p` override.
- Docstrings: helper trigger phrased on `port`; `0.0.0.0` documented; **design check confirmed** — a fast-failing loop stays green *by design* (liveness must not couple to backend reachability; rationale now in `is_poll_stale`).
- Tests: top-of-poll stamp test, `/healthz`+`/livez`+404 coverage, double-start rejection.

**Verified:** 8/8 heartbeat/liveness unit tests green; live `/health` → 200; bind-conflict → graceful degrade. Pre-commit clean.

muhammad-ali-e:
## SonarCloud issues addressed — `c2a62fcc2`

- **consumer.py** (Major) — switched the liveness-bind `except` block from `logger.error` to **`logger.exception`** (preserves the traceback).
- **test_pg_queue_consumer.py** (Minor) — lifted the walrus `:=` out of the `PgQueueConsumer(...)` argument list into a plain `client = MagicMock()` first.

Heartbeat/run-loop tests still green; pre-commit clean.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -265,21 +303,184 @@ def _env(suffix: str, default: _T, cast: Callable[[str], _T]) -> _T:
...
+        if self._httpd is not None:
+            raise RuntimeError("LivenessServer already started")
+
+        consumer = self._consumer
+        stale_after = self._stale_after
+        paths = self._PATHS
+
+        class _Handler(BaseHTTPRequestHandler):
+            def do_GET(self) -> None:
+                if self.path not in paths:
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in b54efabcc. The handler now matches on `urlsplit(self.path).path`, so `/health?probe=k8s` returns 200. Added a query-string case to the alias test.

muhammad-ali-e:
## Greptile feedback addressed — `b54efabcc`

Thanks @greptile-apps (4/5). Both findings handled:

- **Query string → 404** (`consumer.py:413`) — *fixed*. Handler now matches `urlsplit(self.path).path`, so `/health?probe=k8s` → 200. Covered by a new query-string case in the alias test.
- **Long-running tasks trip the probe** (outside-diff) — *working-as-designed, now documented prominently*. Stamping at the top of `poll_once` and freezing during task execution is deliberate (it's what lets a genuinely wedged loop be detected). The consequence — `WORKER_PG_QUEUE_CONSUMER_HEALTH_STALE_SECONDS` doubles as an upper bound on single-task wall-clock — is now spelled out on `_DEFAULT_HEALTH_STALE_SECONDS` with your formula: raise it above `max(batch_size × worst_case_task_seconds, backoff_max)` for longer tasks. The 60s default is safe for the only currently-routed leaf (`send_webhook_notification`, sub-second); longer tasks belong to the pipeline phase, which will set its own threshold.

8/8 heartbeat/liveness tests green; pre-commit clean.

PR #2052, opened 2026-06-12
UN-3546 [FEAT] Priority-ordered PG-queue dequeue + concurrency-safe claim

muhammad-ali-e:
Starts *enforcing* the load-independent part of fairness — `pipeline_priority` (L3) — directly in the single-table dequeue. The fairness key already flowed on every PG message but was pure pass-through; now a higher-priority task is claimed first, FIFO within a priority. Targets `feat/UN-3445-pg-queue-integration`.

## ⚠️ Also fixes a latent concurrency bug in the core dequeue (9a)
While testing the new ordering, the claim **over-claimed under concurrency** — `read(qty=1)` returned 2 rows when other transactions touched the table. Root cause is in the **original** dequeue, not the new ordering:

`UPDATE … WHERE msg_id IN (SELECT … FOR UPDATE SKIP LOCKED LIMIT n)` — **EvalPlanQual** re-evaluates the `LIMIT` subquery when a row it tried to lock was concurrently modified, so a single claim can return **more than `n`** rows. With concurrent consumers that means a `read(qty=1)` could pull (and bump the vt on) extra messages.

**Fix** — the canonical PGMQ-safe shape: lock candidates in a CTE, then `UPDATE … FROM locked WHERE q.msg_id = locked.msg_id` (locks exactly `n` once). The trailing `SELECT` re-orders `RETURNING` (otherwise unspecified) so batched claims also come back in priority order.

```sql
WITH locked AS (
    SELECT msg_id FROM pg_queue_message
     WHERE queue_name = %s AND vt <= now()
     ORDER BY priority DESC, msg_id
     FOR UPDATE SKIP LOCKED LIMIT %s
), claimed AS (
    UPDATE pg_queue_message q SET vt = now() + …, read_ct = read_ct + 1
      FROM locked WHERE q.msg_id = locked.msg_id
    RETURNING q.msg_id, q.message, q.read_ct, q.priority
)
SELECT msg_id, message, read_ct FROM claimed ORDER BY priority DESC, msg_id
```

This benefits the whole queue (every caller), not just priority ordering.

## What
- **Schema** (`models.py` + migration `0002`): `priority smallint default 5` (= `FairnessKey.DEFAULT_PRIORITY`); dequeue index swapped to `(queue_name, priority DESC, msg_id)` so the priority claim stays an indexed top-N (no full-backlog sort).
- **Enqueue**: `dispatch()` writes `priority` from `fairness.pipeline_priority`; a bare dispatch (`fairness=None`) writes the neutral default.
- **Dequeue**: `ORDER BY priority DESC, msg_id` (higher first, FIFO within a priority).

## Deliberately deferred (need load / belong with the orchestrator)
L1 org-tier + L2 workload ordering, `org_config`/`burst_max` admission, `staging_queue`/`task_queue` split, `orchestrator_lock` leader election — the multi-org fair-admission orchestrator, justified once the pipeline routes cross-org load through PG.

## Testing
- **Integration (real Postgres):** priority selection one-at-a-time + batch ordering; **48 tests pass 5× consecutively** across client+dispatch+routing (the combo that exposed the over-claim).
- **Unit:** `send()` writes priority; dispatch wiring (fairness + neutral default); read param order.
- **Live end-to-end:** dispatched 5 mixed-priority tasks via the real `dispatch()` path → drained in claim order `9 > 7 > 3a > 3b(FIFO) > 1`.
- **Real flow:** live API/ETL workflow notifications land in `pg_queue_message` with `priority: 5` (neutral, `fairness=null`) — column populated end-to-end through the backend producer.
- Migration applies clean; descending index confirmed in the DB.


workers/queue_backend/pg_queue/client.py
```diff
@@ -144,22 +169,36 @@ def _cursor(self) -> Iterator[Any]:
             raise
 
     def send(
-        self, queue_name: str, message: dict[str, Any], *, org_id: str | None = None
+        self,
+        queue_name: str,
+        message: dict[str, Any],
+        *,
+        org_id: str | None = None,
+        priority: int = DEFAULT_PRIORITY,
```
muhammad-ali-e, line 181: **🔴 Critical (consensus of 4 reviewers): validate `priority` here.**

`send()` is the write boundary into durable storage and currently trusts its caller — it writes any `int` straight into the INSERT (line 200). The `[1,10]` contract documented just below (and in the model comment / migration) is enforced *only* in `FairnessKey.__post_init__`, which is upstream on one of two paths. `send()` is public; the test at `test_pg_queue_client.py` already calls `priority=9` directly. Because `read()` orders by `priority DESC`, an out-of-range value (e.g. `priority=99` or `-1`) silently jumps or sinks the row and corrupts dequeue ordering — no exception, no log.

This mirrors the guards `read()` already has for `vt_seconds`/`qty`:
```python
from ..fairness import DEFAULT_PRIORITY, MAX_PRIORITY, MIN_PRIORITY

if not MIN_PRIORITY <= priority <= MAX_PRIORITY:
    raise ValueError(
        f"priority out of range [{MIN_PRIORITY}, {MAX_PRIORITY}]: {priority!r}"
    )
```
Pair with a DB `CheckConstraint` (see comment on `models.py`) so the raw-SQL/ORM writer is also covered.

backend/pg_queue/models.py
```diff
@@ -29,12 +30,26 @@ class PgQueueMessage(models.Model):
     enqueued_at = models.DateTimeField(default=timezone.now)
     vt = models.DateTimeField(default=timezone.now)  # visibility timeout
     read_ct = models.IntegerField(default=0)
+    # fairness L3: higher value is claimed sooner (1..10, default 5 =
+    # FairnessKey.DEFAULT_PRIORITY). The dispatch writes it from
+    # fairness.pipeline_priority; leaf tasks with no fairness get the neutral
+    # default. L1 (org tier) / L2 (workload) ordering + burst_max admission are
+    # deferred to the fair-admission orchestrator (a later phase).
+    priority = models.SmallIntegerField(default=5)
```
muhammad-ali-e, line 43: **🟠 Important: add a DB-level CHECK; the `1..10` promise above is currently unenforced.**

`SmallIntegerField(default=5)` enforces only the 16-bit bound — no `validators=`, no constraint. The model docstring advertises this as a typed read handle for the backend producer, so an ORM `.create(priority=42)` (or any raw-SQL write) is accepted and silently corrupts dequeue ordering. The DB is the one layer no caller can bypass:
```python
class Meta:
    constraints = [
        models.CheckConstraint(
            check=models.Q(priority__gte=1) & models.Q(priority__lte=10),
            name="pg_queue_message_priority_range",
        ),
    ]
```
(plus the matching migration op). A bad write then fails loudly with `IntegrityError` at insert time instead of reordering the queue invisibly.

workers/queue_backend/pg_queue/client.py
```diff
@@ -50,25 +51,49 @@
...
 # during processing -> VACUUM-safe and PgBouncer txn-pooling compatible.
 #
-# ORDER BY (vt, msg_id) — not just msg_id — so the (queue_name, vt, msg_id)
+# ORDER BY (priority DESC, msg_id) — the (queue_name, priority DESC, msg_id)
 # index drives an indexed top-N: for a fixed queue the index is ordered by
-# (vt, msg_id), so the vt<=now() range scan emits rows already sorted and PG
-# applies LIMIT without sorting the whole visible backlog. Still effectively
-# FIFO (vt is set to now() at enqueue; msg_id is the deterministic tiebreak).
+# (priority DESC, msg_id), so PG walks rows highest-priority-first, applies the
+# vt<=now() visibility filter as it goes, and stops at LIMIT — no sort of the
```
muhammad-ali-e, line 57: **🟠 Important: this overstates the guarantee.** The new index `(queue_name, priority DESC, msg_id)` drops `vt`, so `vt <= now()` is a per-row filter applied while walking in priority order — not a range stop. Under this queue's at-least-once contract, every claimed-but-unacked row has a *future* `vt` (now + `vt_seconds`, default 30s) and sits at the **front** of its priority band. So a claim must scan and discard every in-flight high-priority row before reaching the first visible candidate: per-claim cost grows with the in-flight backlog, not with `LIMIT`. The old `(queue_name, vt, msg_id)` index did not have this problem.

Please soften the wording (it's an indexed walk that *skips* invisible rows, not a guaranteed top-N), and consider a partial index keyed on visible rows — and benchmark with a realistic in-flight backlog before relying on the top-N characterization.

backend/pg_queue/models.py
```diff
@@ -29,12 +30,26 @@ class PgQueueMessage(models.Model):
...
 
     class Meta:
         db_table = "pg_queue_message"
         indexes = [
+            # Dequeue scans one queue in (priority DESC, msg_id ASC) order and
```
muhammad-ali-e: 🟡 Same caveat as the mirrored comment in `client.py`: "stays an indexed top-N, no full-backlog sort" is planner-dependent and, more importantly, doesn't hold once in-flight (future-`vt`) high-priority rows accumulate — they get scanned-and-skipped on each claim because `vt` is no longer in the index. Suggest softening to note the visibility filter is applied during the index walk (so invisible high-priority rows are scanned past).

backend/pg_queue/models.py
```diff
@@ -29,12 +30,26 @@ class PgQueueMessage(models.Model):
     enqueued_at = models.DateTimeField(default=timezone.now)
     vt = models.DateTimeField(default=timezone.now)  # visibility timeout
     read_ct = models.IntegerField(default=0)
+    # fairness L3: higher value is claimed sooner (1..10, default 5 =
+    # FairnessKey.DEFAULT_PRIORITY). The dispatch writes it from
```
muhammad-ali-e: 🟡 Nit: `FairnessKey.DEFAULT_PRIORITY` isn't a real symbol — `DEFAULT_PRIORITY` is a module-level `Final[int]` in `workers/queue_backend/fairness.py` (`FairnessKey` only has a default-valued `pipeline_priority` field). A reader grepping for `FairnessKey.DEFAULT_PRIORITY` finds nothing. Suggest `default 5 = fairness.DEFAULT_PRIORITY`.

workers/queue_backend/pg_queue/client.py
```diff
@@ -50,25 +51,49 @@
...
+# sort is a no-op there.)
+# Canonical PGMQ-safe shape: lock the candidate rows in a CTE, then UPDATE by
+# JOINING that CTE (``UPDATE ... FROM locked WHERE msg_id = locked.msg_id``).
+# The alternative — ``UPDATE ... WHERE msg_id IN (SELECT ... FOR UPDATE SKIP
+# LOCKED LIMIT n)`` — can over-claim under concurrent writers: EvalPlanQual
+# re-evaluates the LIMIT subquery when a row it tried to lock was concurrently
+# touched, so a single claim can return more than ``n`` rows. The FROM-join form
+# locks exactly ``n`` rows once and updates precisely those. The trailing SELECT
+# re-applies the order because UPDATE ... RETURNING is otherwise unordered.
+# Param order follows the %s positions: queue_name, qty, vt_seconds.
```
muhammad-ali-e: 🟡 Nit (simplifier): this param-order line duplicates the inline comment at the `cur.execute` call site (lines 224-225). The call-site comment is the more useful location — consider dropping this one to avoid the two drifting apart.

workers/tests/test_pg_queue_client.py
```diff
@@ -264,6 +280,33 @@ def test_read_hides_message_for_vt(self, pg_conn, queue_name):
         # Second read within vt sees nothing.
         assert client.read(queue_name, vt_seconds=30, qty=10) == []
 
+    def test_priority_orders_dequeue(self, pg_conn, queue_name):
```
muhammad-ali-e, line 291: **🟡 Test-coverage gaps around the new priority claim** (the priority-ordering tests themselves are good):

1. **Concurrent-writer over-claim is untested (highest value).** The entire reason for the CTE rewrite is that the `FROM locked` form claims exactly `n` under contention where `IN (SELECT … FOR UPDATE … LIMIT n)` can over-claim via EvalPlanQual. `test_no_double_delivery_across_readers` only proves two readers get *disjoint* sets — both SQL shapes pass it. Add a test driving two readers in parallel against a backlog asserting every returned batch has `len <= qty` (best-effort/looped, since EvalPlanQual is racy).
2. **`vt` × priority interaction is untested.** Send a high-priority row, push its `vt` into the future, and assert a visible lower-priority row is claimed first. No current test makes priority and visibility interact.
3. **FIFO-within-band for batch claims** only uses single-member bands here (`test_priority_orders_batch_claim`). Add two rows at priority 9 and two at priority 1 (interleaved enqueue order) and assert strict `[9-first, 9-second, 1-first, 1-second]` — this exercises the outer `ORDER BY priority DESC, msg_id` tiebreak arm.

muhammad-ali-e (commented):

Ran six specialized reviewers (Code Reviewer, Silent-Failure Hunter, Type-Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier) over the diff vs `feat/UN-3445-pg-queue-integration`. Inline comments are attached below; this is the prioritized summary.

**The PR is solid** — the CTE FROM-join claim (vs the EvalPlanQual-unsafe `IN (SELECT … LIMIT n)` form) is the correct PGMQ-safe shape, param ordering is right, the migration sequence is valid, and the `_enqueue_pg` re-raise-with-breadcrumb (no silent Celery fallback) is exactly right. Findings are about hardening, not bugs in the happy path.

### 🔴 Critical / Important
1. **`priority` has no range enforcement** (flagged independently by 4 of 6 agents). `client.send()` writes any `int` straight to the DB, and the `priority` column has no CHECK — only `FairnessKey.__post_init__` validates `[1,10]`, and that sits on just one of the write paths. An out-of-range value is silently durable and **corrupts dequeue ordering** with no log/error. Today's only caller passes a validated value, but this PR is the plumbing that lets an arbitrary priority through. Fix at both layers: a `ValueError` guard in `send()` (mirroring its existing `vt_seconds`/`qty` guards) **and** a DB `CheckConstraint` (the only backstop for the raw-SQL/ORM writer). See inline on `client.py` send() and `models.py`.
2. **Index dropped `vt`; the "indexed top-N" comment over-claims.** The new `(queue_name, priority DESC, msg_id)` index removes `vt`, so visibility is a per-row filter. Under at-least-once delivery, in-flight (future-`vt`) high-priority rows pile up at the front of the index and get scanned-and-skipped on every claim — per-claim cost grows with the in-flight backlog, not `LIMIT`. Soften the comment and consider a partial index / benchmark with a realistic in-flight backlog. See inline on `client.py:57` and `models.py:43`.

### 🟡 Test gaps (inline on `test_pg_queue_client.py:283`)
- The rewrite's whole reason for existing — CTE prevents over-claim under **concurrent writers** — has zero coverage (`test_no_double_delivery` only proves disjointness, not the ≤`qty` bound).
- `vt` × priority interaction is untested (a future-`vt` high-priority row should be skipped for a visible lower-priority one).
- FIFO-within-band for **batch** claims only exercises single-member bands.

### 🟢 Comment / cleanup nits (inline)
- Module docstring (`client.py:8-10`, outside the diff so not inline-commentable) still describes the old `UPDATE … WHERE msg_id IN (SELECT … FOR UPDATE SKIP LOCKED …)` shape this PR replaced and argues against — please update it to the CTE form.
- `models.py:34`: `FairnessKey.DEFAULT_PRIORITY` is not a real symbol (it's module-level `fairness.DEFAULT_PRIORITY`).
- `client.py:76`: param-order comment is duplicated at the call site (224-225) — drop one.

_Generated with [Claude Code](https://claude.com/claude-code) via the PR Review Toolkit._

workers/queue_backend/pg_queue/client.py
muhammad-ali-e: Fixed in 15d1d45cf. `send()` now raises `ValueError` if priority is outside `[MIN_PRIORITY, MAX_PRIORITY]`, mirroring its `vt_seconds`/`qty` guards. Added a parametrized test (`0, -1, 11, 99`). Paired with the DB CheckConstraint below.

backend/pg_queue/models.py
muhammad-ali-e: Fixed in 15d1d45cf. Added `CheckConstraint(priority 1..10)` (migration 0003) — a bad ORM/raw write now fails with IntegrityError at insert instead of reordering the queue. Test inserts `priority=42` via raw SQL and asserts `CheckViolation`. Used `check=` (not `condition=`) since the repo is on Django 4.2.1 — `condition=` only exists in 5.1+.

workers/queue_backend/pg_queue/client.py
muhammad-ali-e: Fixed in 15d1d45cf. Reworded: the dequeue is an index walk over `(queue_name, priority DESC, msg_id)` with `vt <= now()` applied as a per-row filter during the walk — explicitly NOT a guaranteed top-N. Noted that in-flight (future-vt) high-priority rows sit at the front of their band and are scanned past on each claim (cheap at low in-flight depth; the orchestrator admission is the high-backlog answer). A `WHERE vt <= now()` partial index is not possible (`now()` is STABLE, not IMMUTABLE), so I documented the limitation rather than adding one; benchmarking under a realistic backlog is a good follow-up.

backend/pg_queue/models.py
muhammad-ali-e: Fixed in 15d1d45cf — same softening applied to the models.py comment (visibility filter applied during the index walk; invisible high-priority rows scanned past).

backend/pg_queue/models.py
muhammad-ali-e: Fixed in 15d1d45cf — now reads `default 5 = fairness.DEFAULT_PRIORITY`.

workers/queue_backend/pg_queue/client.py
muhammad-ali-e: Fixed in 15d1d45cf — dropped the param-order comment on `_DEQUEUE_SQL`, kept the one at the `cur.execute` call site.

workers/tests/test_pg_queue_client.py
muhammad-ali-e: All three added in 15d1d45cf: (1) `test_concurrent_claims_never_exceed_qty` — two readers drain a 50-row backlog in parallel, asserting no batch exceeds qty (the regression guard for the EvalPlanQual over-claim; passes on the FROM-join form). (2) `test_visible_low_priority_beats_invisible_high` — claim a high row so its vt goes future, then a visible low row is claimed first. (3) `test_batch_fifo_within_priority_band` — two rows each at pri 9 and 1, interleaved, asserting `[9a, 9b, 1a, 1b]`.

muhammad-ali-e:
## Review feedback addressed — `15d1d45cf`

Thanks — all findings handled (threads replied inline).

**🔴 Important**
- **priority validation** — `send()` now raises on out-of-range (mirrors its `vt_seconds`/`qty` guards), **plus** a DB `CheckConstraint(1..10)` (migration 0003) as the backstop no ORM/raw writer can bypass. (`check=`, since the repo is Django 4.2; `condition=` is 5.1+.)
- **index "top-N" overclaim** — softened both comments: it's an index walk with `vt <= now()` as a per-row filter, **not** a guaranteed top-N; in-flight high-priority rows are scanned past (cheap at low depth; orchestrator admission is the high-backlog answer). A `vt <= now()` partial index isn't possible (`now()` is STABLE), so I documented the limit; benchmarking under a realistic backlog noted as follow-up.

**🟡 Tests** — added all three: concurrent-writer over-claim guard (the regression test for the EvalPlanQual fix), `vt × priority` (visible low beats invisible high), and FIFO-within-band for multi-member batch bands. Plus the `send()` range-guard + DB-constraint backstop tests.

**🟢 Nits** — module docstring updated to the CTE FROM-join shape; `fairness.DEFAULT_PRIORITY` symbol fixed; duplicate param-order comment dropped.

**Verified:** 56 tests pass 3× consecutively (client + dispatch + routing); migration 0003 applies clean.

workers/tests/test_pg_queue_client.py
```diff
@@ -264,6 +288,98 @@ def test_read_hides_message_for_vt(self, pg_conn, queue_name):
...
+                    client.delete(m.msg_id)
+
+        try:
+            worker = threading.Thread(target=drain, args=(PgQueueClient(conn=conn_b),))
+            worker.start()
+            drain(client_a)
+            worker.join(timeout=15)
+        finally:
+            conn_b.close()
+        assert violations == []
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in b68ec9fb4. Added `assert not worker.is_alive(), "drain worker did not finish within 15s"` right after the join (before `conn_b.close()`), so a hung drain fails the test loudly instead of passing silently with a 15s pause + a swallowed close race.

backend/pg_queue/models.py
```diff
@@ -29,12 +30,35 @@ class PgQueueMessage(models.Model):
...
+    # CheckConstraint is the one backstop no writer (ORM or raw SQL) can bypass.
+    priority = models.SmallIntegerField(default=5)
 
     class Meta:
         db_table = "pg_queue_message"
+        constraints = [
+            models.CheckConstraint(
+                check=models.Q(priority__gte=1) & models.Q(priority__lte=10),
+                name="pg_queue_message_priority_range",
+            ),
```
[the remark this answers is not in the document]
muhammad-ali-e: Addressed in b68ec9fb4 with a drift guard. A shared constants module in `@unstract/` is the proper long-term fix, but backend and workers cannot import each other today, so I pinned the two together via the live DB: `test_db_check_constraint_matches_fairness_bounds` raw-inserts at `MIN_PRIORITY`/`MAX_PRIORITY` (accepted) and `MIN_PRIORITY-1`/`MAX_PRIORITY+1` (CheckViolation), all sourced from `fairness.py`. A future range change in `fairness.py` that the DB constraint does not match now fails loudly. The model comment documents fairness.py as the canonical source. Happy to extract the shared constant into `@unstract/` as a follow-up if you prefer that over the test guard.

muhammad-ali-e:
## Greptile feedback addressed — `b68ec9fb4`

- **Concurrency test liveness** — `assert not worker.is_alive()` after the join, so a hung drain fails loudly instead of passing silently (and racing `conn_b.close()`).
- **Priority-bounds drift** — replaced the hardcoded `42` reject test with `test_db_check_constraint_matches_fairness_bounds`: raw-inserts at `MIN_PRIORITY`/`MAX_PRIORITY` (accepted) and just outside (CheckViolation), all from `fairness.py` — so the DB constraint and the app's fairness range can't silently diverge. Model comment documents fairness.py as canonical. (A shared `@unstract/` constant is the proper long-term fix; offered as a follow-up since backend/workers can't import each other today.)

10 integration tests green; both findings were non-blocking (Greptile 4/5, safe-to-merge).

PR #2053, opened 2026-06-12
UN-3548 [FEAT] PgBarrier — Postgres fan-in barrier (3rd WORKER_BARRIER_BACKEND)

muhammad-ali-e:
Adds a third `Barrier` substrate, `PgBarrier`, selected by `WORKER_BARRIER_BACKEND=pg` (default stays `chord`). It moves the fan-in aggregation — "wait for N header tasks, then fire the callback with their results" — onto a Postgres row, so an execution can coordinate in the **same DB that holds the PG queue**, with no Redis or RabbitMQ chord backend. The 9e pipeline on-ramp primitive. Targets `feat/UN-3445-pg-queue-integration`.

Mirrors `RedisDecrBarrier` (6b) 1:1 — same `Barrier` protocol, fairness plumbing, **Celery-dispatched header tasks** with `.link`/`.link_error`, `empty → None`, missing-`execution_id` → raise, mid-loop dispatch cleanup. **Defaults-off, zero behaviour change** until the flag flips. (Header-task *transport* stays Celery; only the coordination moves — routing header tasks through PG is 9e proper.)

## What
- **Schema** (`backend/pg_queue/`): `pg_barrier_state` (execution_id PK, remaining, results jsonb, aborted, expires_at) + migration `0004`.
- **`pg_barrier.py`**: `PgBarrier` + `barrier_pg_decr_and_check` / `barrier_pg_abort`.
- **`__init__.py`**: `BarrierBackend.PG` → `PgBarrier()` in `get_barrier()`.

## The atomic decrement is one statement (no Lua)
```sql
UPDATE pg_barrier_state
   SET remaining = remaining - 1,
       results   = results || jsonb_build_array(%s::jsonb)
 WHERE execution_id = %s
RETURNING remaining, results, aborted
```
The row lock serialises concurrent decrements, so **exactly one** task observes `remaining = 0` — verified by a real two-connection atomicity test.

## Failure-masking guards
- Reads `aborted` in the **same** statement → never fires the callback with partial results, even if an abort set the flag but its DELETE hadn't run.
- Row-missing (already torn down) / `remaining < 0` (expiry/replay) → clean up, no fire.
- Callback dispatched **before** the row is deleted → a callback `apply_async` failure leaves the row + expiry in place rather than stranding the execution.
- Concurrent aborts collapse to one cleanup via `UPDATE … SET aborted WHERE NOT aborted RETURNING`.
- Orphan bound via `expires_at`; a periodic sweep job (not yet shipped) is the reclaim backstop. (Abort dedups via `DELETE … RETURNING` / row existence — no `aborted` column.)

## Testing
- **26 PgBarrier tests** + the selector `pg` case — protocol shape, TTL env validation, enqueue (upsert / links / fairness / stale-reset / mid-loop-cleanup), decrement paths (pending / complete-fires / decrement-after-abort / negative / missing / unserialisable / NUL-teardown / dispatch-failure-preserves-row / list-result), abort (claim+delete / dedup), a **real two-connection atomicity check driven through `barrier_pg_decr_and_check`** (exactly one fires), plus DB-constraint and `max_retries=0`.
- 53 barrier-related tests pass (3× stable); migration applies clean; `get_barrier(pg)` → `PgBarrier`.
- Inspected a real `pg_barrier_state` row via the live `enqueue` path (remaining counts down, results grows) to confirm the on-the-wire shape.

## Out of scope
- Routing header tasks themselves through PG (9e). A periodic orphan-sweep job (future orchestrator REAP).


workers/queue_backend/pg_barrier.py
```diff
@@ -0,0 +1,432 @@
...
+            "RETURNING execution_id",
+            (execution_id,),
+        )
+        claimed = cur.fetchone() is not None
+
+    if not claimed:
+        # Another failure already aborted this execution (or it's already gone).
+        return {"status": "already_aborted", "execution_id": execution_id}
+
+    _delete_barrier(execution_id)
```
muhammad-ali-e: **[High] Abort claim and delete span two separate transactions.** The `UPDATE … SET aborted = true WHERE … NOT aborted RETURNING` commits in its own `_cursor()` block, then `_delete_barrier` runs here in a second transaction. If the process dies (or the DELETE fails) between the two commits, the row is left `aborted=true, remaining>0`. Recovery is then only via a late in-flight decrement (which hits the aborted branch and deletes) or `expires_at` (up to 6h) — and every sibling `barrier_pg_abort` short-circuits to `already_aborted` without retrying the delete.

The safety-critical property still holds (the committed `aborted` flag prevents a partial-results fire), so this is a robustness/leak issue, not a correctness one. Suggest collapsing claim+delete into one transaction via a CTE so the row is gone the instant the abort is claimed:
```sql
WITH claimed AS (
  DELETE FROM pg_barrier_state
   WHERE execution_id = %s AND NOT aborted
  RETURNING execution_id
) SELECT execution_id FROM claimed
```
A failure then rolls both back, leaving `aborted=false` so a sibling can retry the whole claim+delete. _(code-reviewer + silent-failure-hunter)_

workers/queue_backend/pg_barrier.py
```diff
@@ -0,0 +1,432 @@
...
+                "queue": callback_queue,
+                "fairness_headers": fairness_headers,
+            }
+            ttl_seconds = _ttl_seconds()
+
+            with _cursor() as cur:
+                # Opportunistic orphan sweep (cheap; the periodic sweep is the
+                # real backstop). Runs before the UPSERT so a stale row for THIS
+                # execution_id is overwritten by the UPSERT, not the sweep.
+                cur.execute("DELETE FROM pg_barrier_state WHERE expires_at < now()")
```
muhammad-ali-e: **[High] Opportunistic orphan sweep runs on the hot path of every `enqueue`, in the same transaction as the UPSERT.** `DELETE FROM pg_barrier_state WHERE expires_at < now()` is unbounded and runs once per execution start, sharing the UPSERT's transaction. Two concurrent enqueues can lock the same set of expired rows in different orders (deadlock potential), and each sweep scans/locks all expired rows rather than just this execution's. The Redis sibling does no global sweep on the enqueue path.

Suggest either dropping the inline sweep and relying on the periodic sweep the docstring already names as "the proper backstop", or at minimum bounding it (`AND ctid IN (SELECT ctid … LIMIT N)`) and moving it to a separate committed block so a sweep rollback can't abort the UPSERT. _(code-reviewer)_

workers/queue_backend/pg_barrier.py
```diff
@@ -0,0 +1,432 @@
...
+        )
+        raise
+
+    # jsonb_build_array(...) appends exactly one element regardless of the
+    # result's shape (``||`` would concatenate if the result were itself a list).
+    with _cursor() as cur:
+        cur.execute(
+            "UPDATE pg_barrier_state "
+            "   SET remaining = remaining - 1, "
+            "       results = results || jsonb_build_array(%s::jsonb) "
```
muhammad-ali-e: **[High] `json.dumps` accepts `\u0000` but Postgres `jsonb` rejects it — unguarded, this hangs the barrier to TTL.** `result_json` passes the `json.dumps` validation above (line 304), but a NUL byte in any string survives `json.dumps` and is then rejected by the `%s::jsonb` cast here with a `psycopg2.DataError`. That error is *not* the typed-boundary regression the surrounding code claims to surface — it propagates as a bare DB error, `remaining` is never decremented, and the barrier hangs until `expires_at` (~6h). The Redis sibling stores the result as a string, so `\u0000` survives there; the `jsonb` column is stricter.

Suggest catching the jsonb-bind failure and converting it to an abort so the execution fails fast instead of hanging, or rejecting NUL bytes at the `json.dumps` boundary. _(silent-failure-hunter)_

workers/queue_backend/pg_barrier.py
```diff
@@ -0,0 +1,432 @@
...
+            {"headers": callback_descriptor["fairness_headers"]}
+            if callback_descriptor.get("fairness_headers")
+            else {}
+        ),
+    )
+    # Dispatch FIRST; delete the row only after dispatch succeeds, so a callback
+    # apply_async failure leaves the row (and its expiry) in place rather than
+    # stranding the execution with no state and no recovery path.
+    callback_result = callback_signature.apply_async()
+    _delete_barrier(execution_id)
```
muhammad-ali-e: **[Medium] `_delete_barrier` failure after a successful `apply_async` is reported to Celery as a task failure even though the callback fired.** The dispatch-before-delete ordering is correct, but if this delete raises after the callback was dispatched, the exception propagates (link task marked FAILED), the success log below never emits, and the row stays at `remaining=0`. It's safe from double-firing *only* because this was the last decrement and `max_retries=0` — a load-bearing, implicit invariant. Suggest wrapping the post-dispatch delete to log at error severity (including `callback_result.id` to prove the callback fired) without masking the successful dispatch, and adding a comment noting the no-double-fire invariant depends on `max_retries=0`. _(silent-failure-hunter)_

workers/queue_backend/pg_barrier.py
```diff
@@ -0,0 +1,432 @@
...
+   results, then deletes the row. (No Lua — a single ``UPDATE … RETURNING`` is
+   atomic in Postgres.)
+3. Per-task failure: ``barrier_pg_abort`` runs as a ``link_error``. It claims
+   the abort exactly once (``UPDATE … SET aborted WHERE NOT aborted RETURNING``)
+   to collapse N concurrent failures into one cleanup, then deletes the row.
+   Mirrors chord's default error semantic (callback not invoked on header
+   failure).
+4. Orphan bound: like the Redis backend's key TTL, ``expires_at`` bounds a
+   barrier whose header tasks never complete. New ``enqueue`` calls sweep
+   past-expiry rows opportunistically; a periodic sweep is the proper backstop.
```
muhammad-ali-e: **[Medium] "Periodic sweep" is referenced as an existing backstop, but it does not ship in this PR.** This comment (and `models.py:83`, plus lines 222-224 / 361-362) describes the periodic sweep as "the proper backstop" / "the real backstop", yet the only expiry mechanism implemented is the opportunistic `DELETE` inside `enqueue`. If enqueues stop, expired rows accumulate indefinitely, and the `pg_barrier_expires_idx` looks load-bearing for a consumer that doesn't exist. Suggest softening to future/planned work with a ticket reference (e.g. "a periodic sweep (not yet implemented; see UN-XXXX) is the intended backstop"). _(comment-analyzer)_

backend/pg_queue/models.py
```diff
@@ -66,3 +66,44 @@ class Meta:
...
+    results = models.JSONField(default=list)
+    # Set once on the first header-task failure so the abort cleanup runs once.
+    aborted = models.BooleanField(default=False)
+    created_at = models.DateTimeField(default=timezone.now)
+    # Orphan bound (Redis-TTL equivalent): a barrier whose header tasks never
+    # complete is reclaimable past this. Must exceed the longest execution
+    # wall-clock, same budgeting as WORKER_BARRIER_KEY_TTL_SECONDS.
+    expires_at = models.DateTimeField()
+
+    class Meta:
```
muhammad-ali-e, line 102: **[Medium] No DB-level invariant backstop, unlike the sibling `PgQueueMessage` (which has a `CheckConstraint`).** Every invariant of this model is enforced only by raw SQL in the `workers/` codebase, which can't import this model; an ORM caller could create a row violating all of them. Suggest adding the cheap, writer-proof check:
```python
models.CheckConstraint(
    check=Q(expires_at__gt=F("created_at")),
    name="pg_barrier_expires_after_created",
)
```
Note: do **not** add a `remaining >= 0` check — the design relies on `remaining` going negative as a teardown signal; worth a `Meta` comment so a future reviewer doesn't "helpfully" add it. _(type-design-analyzer)_

workers/queue_backend/pg_barrier.py
```diff
@@ -0,0 +1,432 @@
...
+if TYPE_CHECKING:
+    from celery.canvas import Signature
+    from psycopg2.extensions import connection as PgConnection
+
+logger = logging.getLogger(__name__)
+
+_DEFAULT_TTL_SECONDS = 6 * 60 * 60  # 6h — mirrors RedisDecrBarrier's key TTL
+
+
+def _ttl_seconds() -> int:
```
muhammad-ali-e: **[Low] `_ttl_seconds()` is a near-verbatim duplicate of `redis_barrier._key_ttl_seconds()`** — same env var (`WORKER_BARRIER_KEY_TTL_SECONDS`), same 6h default, same validation. Since only one backend is active per deployment, the contract is genuinely shared. Suggest extracting one `barrier_ttl_seconds()` into a shared module (e.g. alongside the `Barrier` protocol in `barrier.py`) and calling it from both, to prevent drift. _(code-simplifier)_

workers/queue_backend/pg_barrier.py
```diff
@@ -0,0 +1,432 @@
...
+
+
+def _delete_barrier(execution_id: str) -> None:
+    with _cursor() as cur:
+        cur.execute(
+            "DELETE FROM pg_barrier_state WHERE execution_id = %s", (execution_id,)
+        )
+
+
+class CallbackDescriptor(TypedDict):
```
muhammad-ali-e: **[Low] `CallbackDescriptor` TypedDict is duplicated verbatim with `redis_barrier.py`** (identical four keys, both crossing the same Celery serialization boundary). This is exactly the drift risk the `handle.py` consolidation was meant to eliminate. Suggest hoisting a single `CallbackDescriptor` into a shared module and importing it in both backends. _(code-simplifier + type-design-analyzer)_

workers/queue_backend/pg_barrier.py
```diff
@@ -0,0 +1,432 @@
...
+        return {"status": "abandoned", "remaining": remaining}
+
+    # remaining == 0: we are the last task. psycopg2 decodes the jsonb array to
+    # a Python list already — no per-element json.loads needed.
+    callback_signature = current_app.signature(
+        callback_descriptor["task_name"],
+        args=[all_results],
+        kwargs=callback_descriptor["kwargs"],
+        queue=callback_descriptor["queue"],
+        **(
```
muhammad-ali-e: **[Low] Inline `**({"headers": …} if … else {})` spread is hard to read.** `CeleryChordBarrier.enqueue` already established the clearer idiom in this codebase — build a `signature_kwargs` dict and set `headers` only when truthy. Behavior is identical (no spurious `headers=None`). Ideally apply to the Redis sibling too so they stay consistent. _(code-simplifier)_

workers/queue_backend/pg_barrier.py
```diff
@@ -0,0 +1,432 @@
...
+
+1. ``enqueue``: UPSERT one ``pg_barrier_state`` row (``remaining = N``,
+   ``results = []``, ``aborted = false``, ``expires_at = now() + ttl``) — the
+   UPSERT clears any stale state from a prior run reusing the same
+   ``execution_id``. Each header task is dispatched with
+   ``.link(barrier_pg_decr_and_check)`` (success) and
+   ``.link_error(barrier_pg_abort)`` (failure).
+2. Per-task success: ``barrier_pg_decr_and_check`` runs ONE atomic statement —
+   ``UPDATE … SET remaining = remaining - 1, results = results ||
+   jsonb_build_array(result) … RETURNING remaining, results, aborted``. The row
```
muhammad-ali-e: **[Low] The "exactly one task observes `remaining = 0`" atomicity claim omits its load-bearing premise.** The guarantee holds only because `_cursor()` commits each `UPDATE` in its own transaction (the connection is `autocommit=False`), so the row lock is released immediately and concurrent decrements serialize to distinct post-decrement values. If someone later batches decrements into a shared transaction, the guarantee silently breaks. Suggest adding: "…and do not batch decrements into a shared transaction." _(comment-analyzer)_

muhammad-ali-e (commented):

The implementation is high quality and faithfully mirrors `RedisDecrBarrier` — the atomic single-statement decrement, the `aborted`-read-in-decrement guard, and dispatch-before-delete are all sound. Findings below are inline. Highest priority: abort claim/delete atomicity, the per-enqueue orphan sweep on the hot path, and jsonb's rejection of `\u0000` (which `json.dumps` accepts). Nothing here is blocking the safety-critical "never fire with partial results" property, which holds.

workers/tests/test_pg_barrier.py
```diff
@@ -0,0 +1,340 @@
...
+        _seed(barrier_db, "exec-P", 3)
+        out = barrier_pg_decr_and_check(
+            {"f": 1}, execution_id="exec-P", callback_descriptor=_CALLBACK
+        )
+        assert out["status"] == "pending"
+        remaining, results, _ = _row(barrier_db, "exec-P")
+        assert remaining == 2
+        assert results == [{"f": 1}]
+
+    def test_complete_fires_callback_with_aggregated_results(self, barrier_db):
```
muhammad-ali-e, line 228: **[Test gap — High] No test that a callback-dispatch failure leaves the row in place.** This is the central failure-masking invariant in the module docstring ("dispatched BEFORE the row is deleted, so an `apply_async` failure leaves the row"), and the Redis backend has a dedicated test for it (`test_callback_dispatch_failure_preserves_keys_for_ttl`). Suggest patching `apply_async` to raise and asserting `pytest.raises` + the row still exists afterward, so a future delete-before-dispatch reorder is caught. _(pr-test-analyzer)_

workers/tests/test_pg_barrier.py
```diff
@@ -0,0 +1,340 @@
...
+            sig.return_value.apply_async.return_value = MagicMock(id="cb-task-1")
+            out = barrier_pg_decr_and_check(
+                {"f": "b"}, execution_id="exec-C", callback_descriptor=_CALLBACK
+            )
+        assert out["status"] == "complete"
+        # Callback got the full aggregated list as its first positional arg.
+        assert sig.call_args.kwargs["args"] == [[{"f": "a"}, {"f": "b"}]]
+        assert _row(barrier_db, "exec-C") is None  # row deleted after dispatch
+
+    def test_aborted_does_not_fire(self, barrier_db):
```
muhammad-ali-e: **[Test gap — High] `test_aborted_does_not_fire` doesn't exercise the race the guard exists for.** It seeds `remaining=1, aborted=True`, but to prove the guard (the `aborted` check must be evaluated *before* the `remaining == 0` fire path), the test should drive the decrement to exactly 0 on an aborted row and assert the callback still does NOT fire. As written, the test would pass even if the `aborted` check were moved after the fire block. Suggest a `test_aborted_at_zero_remaining_does_not_fire` that asserts `sig.assert_not_called()` and the row is deleted. _(pr-test-analyzer)_

workers/tests/test_pg_barrier.py
```diff
@@ -0,0 +1,340 @@
...
+
+    def test_registered_under_canonical_name(self):
+        assert barrier_pg_abort.name == "barrier_pg_abort"
+
+
+# --- Layer 3: atomicity (two connections race the decrement SQL) ---
+
+
+class TestDecrementAtomicity:
+    def test_exactly_one_reader_sees_zero(self, barrier_db):
```
muhammad-ali-e: **[Test gap — Medium] The atomicity test inlines its own `UPDATE … RETURNING` rather than driving `barrier_pg_decr_and_check`.** It therefore proves Postgres row-locking works (a given), not that *the barrier's* fire path is single-fire. The real risk — two tasks both observing `remaining == 0` and both dispatching the callback — can't be detected because the test never goes through the fire path. Suggest racing two connections through `barrier_pg_decr_and_check` itself (callback mocked) from `remaining=2` and asserting exactly one returns `status == "complete"` and the callback mock fires exactly once. Also worth adding: a `list`-shaped result is appended as a single element (guards the `jsonb_build_array` vs `||` choice) and `max_retries == 0` on both tasks. _(pr-test-analyzer)_

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in 6ec036882 — adopted exactly this. Abort is now `WITH claimed AS (DELETE … RETURNING) SELECT …` (claim+teardown in one transaction; a crash rolls both back so a sibling retries). That made the `aborted` column redundant, so I dropped it — the decrement's "row missing → abandoned" branch covers the failed-task case, and the callback can only fire when remaining hits 0 (all tasks succeeded), so a failed task (which deletes the row) can never let a partial fire.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in 6ec036882 — dropped the inline sweep entirely. Reclaim is a future periodic sweep job keyed on pg_barrier_expires_idx (documented as not-yet-shipped).

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in 6ec036882 — the decrement UPDATE is wrapped; on `psycopg2.DataError` (e.g. a NUL byte the jsonb cast rejects) it tears the barrier down (delete) and re-raises, so the execution fails fast instead of hanging to expiry. Test: `test_nul_byte_result_tears_down_barrier`.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in 6ec036882 — the post-dispatch delete is wrapped in try/except and logged (not raised), including the callback_task_id to prove the callback fired. Added the comment noting the no-double-fire invariant relies on last-decrement + max_retries=0.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in 6ec036882 — softened the comments (module docstring, models.py) to say the periodic sweep is future / not-yet-implemented, and dropped the misleading inline sweep that suggested reclaim already worked.

backend/pg_queue/models.py
muhammad-ali-e: Fixed in 6ec036882 — added `CheckConstraint(expires_at__gt=F("created_at"))` + a Meta comment explicitly warning off a `remaining >= 0` check (the teardown path relies on remaining going negative). Test: `test_expires_at_must_exceed_created_at`.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in 6ec036882 — extracted `barrier_ttl_seconds()` into barrier.py; both backends call it (redis keeps `_key_ttl_seconds`/`_KEY_TTL_DEFAULT_SECONDS` as back-compat aliases).

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in 6ec036882 — `CallbackDescriptor` hoisted into barrier.py; both backends import it.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in 6ec036882 — switched to the `signature_kwargs` dict idiom (headers only when truthy), matching CeleryChordBarrier.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in 6ec036882 — the wire-model comment now states the guarantee relies on each decrement committing in its own transaction ("do NOT batch decrements into a shared transaction").

workers/tests/test_pg_barrier.py
muhammad-ali-e: Fixed in 6ec036882 — added `test_callback_dispatch_failure_preserves_row`: patches apply_async to raise, asserts pytest.raises and that the row still exists (guards a delete-before-dispatch reorder).

workers/tests/test_pg_barrier.py
muhammad-ali-e: Addressed in 6ec036882, adapted to the new design: since the `aborted` column is gone (abort now deletes the row atomically), the partial-fire guard is structural — a failed task deletes the barrier, so a late in-flight decrement finds no row and abandons. New test `test_decrement_after_abort_does_not_fire` drives exactly that race (abort, then a decrement that would have hit 0) and asserts the callback does NOT fire.

workers/tests/test_pg_barrier.py
muhammad-ali-e: Fixed in 6ec036882 — `test_exactly_one_task_fires_the_callback` now races two threads through `barrier_pg_decr_and_check` itself (each with its own connection, callback mocked) from remaining=2 and asserts exactly one returns "complete" and apply_async fires once. Also added `test_list_result_appended_as_single_element` (guards jsonb_build_array vs ||) and `test_max_retries_zero`.

muhammad-ali-e:
## Review feedback addressed — `6ec036882`

Thanks — all 13 findings handled (threads replied inline). A couple led to a nice simplification.

**High**
- **Abort atomicity** → collapsed claim+delete into one `WITH claimed AS (DELETE … RETURNING)` statement. That made the `aborted` column redundant, so I **dropped it** — the "row missing → abandoned" decrement branch covers the failed-task case, and the callback can only fire when remaining hits 0 (all succeeded), so a failed task can never let a partial fire.
- **Hot-path orphan sweep** → removed; reclaim is a future periodic sweep.
- **NUL byte / jsonb** → catch `DataError`, tear the barrier down (fail fast) instead of hanging to expiry.

**Medium**
- Post-dispatch delete is best-effort (logged, not raised) so it can't mask the already-fired callback.
- Added a DB `CheckConstraint(expires_at > created_at)` (+ a "do not add remaining >= 0" Meta note); softened the periodic-sweep comments to future/not-yet-shipped.

**Low** — extracted shared `barrier_ttl_seconds()` + `CallbackDescriptor` into `barrier.py` (redis keeps back-compat aliases); `signature_kwargs` dict; atomicity-premise comment.

**Tests** — callback-dispatch-failure-preserves-row, decrement-after-abort-no-fire (the new structural guard), atomicity *through* the task (two threads, exactly one fires), list-result-as-single-element, NUL-byte teardown, DB constraint, max_retries=0.

**Verified:** 92 barrier tests pass (2× stable); migration regenerated + applied clean. The `aborted` column removal regenerated migration `0004` (renamed; the PR isn't merged so the migration is safe to replace).

workers/queue_backend/pg_barrier.py
```diff
@@ -0,0 +1,419 @@
...
+
+**Why a Postgres substrate.** It lets an execution coordinate in the *same*
+Postgres that holds the PG queue — no Redis (or RabbitMQ chord backend) needed
+for the fan-in. The transport for the header tasks themselves is unchanged
+(still Celery); only the coordination moves.
+
+**Wire model.**
+
+1. ``enqueue``: UPSERT one ``pg_barrier_state`` row (``remaining = N``,
+   ``results = []``, ``aborted = false``, ``expires_at = now() + ttl``) — the
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 3966aed97 — removed `aborted = false` from the enqueue UPSERT line in the wire-model docstring (a leftover from before the column was dropped). Abort dedups via `DELETE … RETURNING` / row existence; there is no `aborted` column. Verified no other stale references remain (the other "aborted" mentions are the abort *status*/action, not a column).

workers/tests/test_pg_barrier.py
```diff
@@ -0,0 +1,407 @@
+"""Tests for :class:`queue_backend.pg_barrier.PgBarrier`.
+
+Three layers:
+
+1. **Protocol / TTL** — no DB, no Celery.
+2. **Enqueue + link/abort tasks** — a real autocommit Postgres connection is
+   injected into the module thread-local (the barrier's SQL runs for real); the
+   Celery header-task dispatch + callback are mocked. Skips if Postgres is
+   unreachable or the ``pg_barrier_state`` migration is unapplied.
+3. **Atomicity** — two real connections race the decrement SQL directly.
```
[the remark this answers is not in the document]
muhammad-ali-e: Good catch — the PR description was stale, not the code. I removed the inline opportunistic sweep during the toolkit pass (it was unbounded/deadlock-prone on the hot path), and its `test_opportunistic_expiry_sweep` went with it, but the description still listed "expiry-sweep". Updated the PR body: dropped the "expiry-sweep" tested-path claim and corrected the orphan-bound line to "`expires_at` + a future (not-yet-shipped) periodic sweep." No sweep ships this phase, so nothing to test yet.

muhammad-ali-e:
## Greptile feedback addressed

Both findings were stale references my own refactor left behind (the code is correct) — Greptile 4/5, safe-to-merge.

- **`aborted`-column docstring** (`pg_barrier.py`) → removed from the enqueue UPSERT line (`3966aed97`). Abort dedups via `DELETE … RETURNING`; no such column.
- **PR description claimed an "expiry-sweep" test** → corrected the PR body. The inline sweep (and its test) were removed in the toolkit pass; reclaim is a future periodic sweep, untested this phase by design.

No code-behaviour change beyond the docstring; 92 barrier tests still green.

PR #2056, opened 2026-06-15
UN-3553 [FEAT] PG Queue 9d slice 1 — leader-election lease (orchestrator_lock)

muhammad-ali-e:
## What

First slice of **9d** (the orchestrator/reaper) on the PG-queue integration branch: a **leader-election lease** so the reaper (and a later fair-admission gate) can run as **exactly one** active instance. This PR is the **primitive + tests only** — nothing acquires leadership yet, so it ships dark and changes no runtime behaviour.

## Why now

9d was skipped in the merged spine (9c → liveness → priority → PgBarrier). It's the safety net the pipeline migration (9e) needs: without a reaper, every at-least-once hang / orphaned barrier bottoms out at the 6h TTL with no recovery. The reaper must be a singleton (multiple would contend on `SKIP LOCKED` and double-act on recovery), so leader election is the foundation slice.

## Key decision: lease, not advisory lock

Leadership is a TTL'd row `UPDATE` (take it if the leader is free or its lease is stale), **not** `pg_advisory_lock`. Session-scoped advisory locks don't survive the transaction-pooled PgBouncer the queue connects through (extension-free decision) — the pooler hands out a different backend per transaction, silently dropping a session-held lock. A plain `UPDATE` is one transaction → pooling-safe. All time comparisons are server-side (`now()`), so candidate clock skew can't split leadership.

## What's in it

**Backend** (`backend/pg_queue/`)
- `PgOrchestratorLock` — single-row lock model (`id` PK default 1, `leader` text default `""`, `acquired_at`), `CheckConstraint(id = 1)`, table `pg_orchestrator_lock`.
- Migration `0005` — generated via `makemigrations` + a reversible `RunPython` that seeds the one free row. Free = empty `leader` (follows the `PgQueueMessage.org_id` no-nullable-text convention).

**Workers** (`workers/queue_backend/pg_queue/leader_election.py`)
- `LeaderLease` — `try_acquire()` / `renew()` / `release()` over the lease row; instance-owned, self-recovering connection (mirrors `PgBarrier` / `PgQueueClient`).
- `lease_seconds_from_env()` — `WORKER_PG_ORCHESTRATOR_LEASE_SECONDS` (default 10s, loud-on-misconfig).
- `default_worker_id()` — `host:pid:rand`, stable per process.

**Tests** (`workers/tests/test_leader_election.py`) — 20, real-PG. The two load-bearing properties:
- concurrent `try_acquire` on a free lock → **exactly one** winner (5-thread race, separate connections);
- `renew()` returns **`False`** after a standby took over a stale lease (the signal that stops a stalled leader).
Plus lease-expiry takeover, release-frees-immediately, non-holder renew/release no-ops, env validation, and the single-row constraint.

## Out of scope (later 9d slices)

- The reaper process itself (acquire → renew loop → release; barrier-orphan sweep; counter reconstruction; re-enqueue/fail stuck rows) + `run-worker.sh` wiring + liveness.
- `worker_registry`.
- Fair-admission gate (tier / workload / `burst_max` via `org_config`) — deferred, no cross-org load yet. (Our single-table design has no staging→task admit step — a deliberate divergence from the labs two-table model.)

## Testing

- 20/20 new tests pass, stable across repeated runs; 91/91 across leader-election + pg_barrier + pg_queue_client + barrier-selection (no regression).
- `makemigrations --check --dry-run` reports no drift (migration faithfully matches the model).
- Migration applied to a dev DB cleanly.

Base: `feat/UN-3445-pg-queue-integration` (not `main`). Sub-task UN-3553 under UN-3536 (Phase 9).


workers/queue_backend/pg_queue/leader_election.py
```diff
@@ -0,0 +1,217 @@
...
+        if not worker_id or not worker_id.strip():
+            # "" is the free sentinel — a candidate must never identify as "".
+            raise ValueError("LeaderLease worker_id must be a non-empty string")
+        self._worker_id = worker_id
+        self._lease_seconds = (
+            lease_seconds if lease_seconds is not None else lease_seconds_from_env()
+        )
+        if self._lease_seconds <= 0:
+            raise ValueError("lease_seconds must be a positive integer")
+        self._conn = conn
```
muhammad-ali-e, line 138: **[High] Missing `_owns_conn` ownership guard — diverges from `PgQueueClient`.** _(flagged by Silent-Failure, Type-Design, Comment-Analyzer, and Test-Analyzer agents)_

`LeaderLease` accepts an injected `conn` but, unlike the sibling `PgQueueClient` (`client.py:131-134, 169`), has no `_owns_conn` flag. On any `OperationalError`/`InterfaceError` (or a failed rollback) `_cursor` does `conn.close()` + `self._conn = None` (lines 153-156), and `_get_conn` (130-135) then silently reconnects with `create_pg_connection(env_prefix="DB_")`.

Impact: the test fixture injects a `TEST_DB_`-pointed autocommit connection per lease (`test_leader_election.py:108-111`); after any transient error a lease would silently re-point at the production `DB_*` database, racing a different DB than the assertions read. In production a leader rebuilds its connection mid-loop with no trace.

**Fix:** mirror `PgQueueClient` — `self._owns_conn = conn is None` here, and gate the close/null-out on `if self._owns_conn and (conn_dead or conn.closed):`. An injected, dead connection should raise rather than be silently swapped.

workers/queue_backend/pg_queue/leader_election.py
```diff
@@ -0,0 +1,217 @@
...
+            conn.commit()
+        except Exception as exc:
+            conn_dead = isinstance(
+                exc, (psycopg2.OperationalError, psycopg2.InterfaceError)
+            )
+            try:
+                conn.rollback()
+            except Exception:
+                conn_dead = True
+            if conn_dead or conn.closed:
```
muhammad-ali-e: **[Medium] Silent reconnect — no log when a leader's connection drops and is rebuilt.** _(Silent-Failure Hunter)_

This recovery branch discards a dead connection with zero logging, and `_get_conn` rebuilds it silently. `create_pg_connection` only logs on the *initial* connect failure, so a mid-life socket drop / PgBouncer recycle on the orchestrator/reaper singleton leaves no trace — yet it correlates directly with missed `renew` cycles and potential leadership handoff. Per the project's no-silent-failure standard this should emit a `logger.warning` with `self._worker_id` and the exception type before dropping the handle.

workers/queue_backend/pg_queue/leader_election.py
```diff
@@ -0,0 +1,217 @@
...
+        conn: PgConnection | None = None,
+    ) -> None:
+        if not worker_id or not worker_id.strip():
+            # "" is the free sentinel — a candidate must never identify as "".
+            raise ValueError("LeaderLease worker_id must be a non-empty string")
+        self._worker_id = worker_id
+        self._lease_seconds = (
+            lease_seconds if lease_seconds is not None else lease_seconds_from_env()
+        )
+        if self._lease_seconds <= 0:
```
muhammad-ali-e: **[Low] Redundant `lease_seconds <= 0` guard on the env path.** _(Code Simplifier, Type-Design)_

When `lease_seconds is None` the value comes from `lease_seconds_from_env()`, which already raises on `<= 0` (lines 81-85) — so this guard is dead code on that path. It *is* load-bearing for the explicit-arg path (`lease_seconds=0`, covered by `test_non_positive_lease_seconds_rejected`), so keep it, but you can scope it to the branch that needs it:
```python
if lease_seconds is None:
    self._lease_seconds = lease_seconds_from_env()
else:
    if lease_seconds <= 0:
        raise ValueError("lease_seconds must be a positive integer")
    self._lease_seconds = lease_seconds
```
Optional / non-blocking — the current single-guard form is also defensible.

workers/queue_backend/pg_queue/leader_election.py
```diff
@@ -0,0 +1,217 @@
...
+                "        OR acquired_at < now() - make_interval(secs => %s)) "
+                "RETURNING id",
+                (self._worker_id, self._lease_seconds),
+            )
+            won = cur.fetchone() is not None
+        if won:
+            logger.info("LeaderLease: %s acquired leadership", self._worker_id)
+        return won
+
+    def renew(self) -> bool:
```
muhammad-ali-e, line 215: **[Low] Undocumented exception-propagation contract.** _(Silent-Failure Hunter)_

`try_acquire`/`renew`/`release` return `bool`, but DB errors (e.g. `CheckViolation`, `DatabaseError`) propagate raw — `_cursor` re-raises by design. This is the *correct* fail-loud posture for an HA primitive, but a caller reading only the docstrings would reasonably assume `False` covers "couldn't get it." The module usage example wraps `renew()` in a bare `if not lease.renew(): break` with no `except`. Add a `Raises:` note so the consuming reaper slice is written to treat an exception as "leadership state unknown — stop acting" rather than assume bool-only.

workers/queue_backend/pg_queue/leader_election.py
```diff
@@ -0,0 +1,217 @@
...
+        Only frees it if we still hold it (``leader = us``); if we already lost
+        leadership this is a no-op, so a late release can't wipe the new holder.
+        """
+        with self._cursor() as cur:
+            cur.execute(
+                "UPDATE pg_orchestrator_lock SET leader = '', acquired_at = now() "
+                "WHERE id = 1 AND leader = %s",
+                (self._worker_id,),
+            )
+        logger.info("LeaderLease: %s released leadership", self._worker_id)
```
muhammad-ali-e: **[Low] `release()` logs "released leadership" even on a no-op.** _(Code Reviewer)_

When the `UPDATE ... WHERE leader = %s` matches zero rows (we already lost leadership to a standby), this still logs `"%s released leadership"` — misreporting that this process freed the lease, which is misleading in exactly the split-brain post-mortem this primitive exists to support. Branch on `cur.rowcount`: log `info("released")` when `rowcount == 1`, else `debug("release was a no-op (not the holder)")`.

workers/queue_backend/pg_queue/leader_election.py
```diff
@@ -0,0 +1,217 @@
...
+hands out a different backend per transaction, so a session-held lock would be
+silently dropped. A plain ``UPDATE`` is one transaction → pooling-safe. Every
+time comparison is server-side (``now()``), so candidate clock skew can't split
+leadership.
+
+**This module ships dark.** It is the primitive only — nothing acquires
+leadership yet. The reaper process that drives ``try_acquire`` → ``renew`` loop →
+``release`` is a separate slice. Merging this changes no runtime behaviour.
+
+Usage (next slice):
```
muhammad-ali-e: **[Low] Transitional "ships dark / next slice" framing will rot.** _(Comment Analyzer)_

The "This module ships dark" paragraph (19-21) and `Usage (next slice):` label (23) describe a point-in-time integration state. The moment the reaper slice lands and calls `LeaderLease`, lines 19-21 become false and "next slice" becomes a dangling reference. Keep the usage example (it's accurate and valuable) but relabel to just `Usage:` and reframe 19-21 durably, e.g. "This module is the lease primitive only — acquiring/driving the lease is the caller's responsibility (the reaper loop)."

workers/tests/test_leader_election.py
```diff
@@ -0,0 +1,233 @@
...
+        labels = [f"c{i}" for i in range(5)]
+        threads = [threading.Thread(target=run, args=(x,)) for x in labels]
+        for t in threads:
+            t.start()
+        for t in threads:
+            t.join(timeout=10)
+        assert all(not t.is_alive() for t in threads)
+        assert sum(results.values()) == 1  # exactly one winner
+        winner = next(label for label, won in results.items() if won)
+        assert _leader(lock_db.conn) == winner
```
muhammad-ali-e, line 307: **[Medium] Connection-recovery path is untested.** _(PR Test Analyzer)_

Lease semantics are well covered, but the self-recovery machinery in `leader_election.py:130-157` — advertised as load-bearing in the docstrings — has no test, despite the sibling `PgQueueClient` carrying dedicated unit tests for the identical logic (`test_pg_queue_client.py:136-187`). These are cheap mocked unit tests (no DB). Suggested additions:
- `test_conn_recovered_on_operational_error`: monkeypatch `create_pg_connection` to a factory returning a MagicMock conn whose `execute` raises `psycopg2.OperationalError`; assert `rollback`/`close` called, `_conn is None`, and a second call reconnects.
- `test_conn_recovered_when_rollback_fails`: `execute` raises `RuntimeError` and `rollback` raises `psycopg2.DatabaseError`; assert the rollback-failure escalation (lines 151-152) closes the conn.

Two smaller documented-invariant gaps also have no test: same-holder re-acquire on a fresh lease returns `False` (the `try_acquire` WHERE rejects even the current holder), and `release()` after losing leadership via takeover is a no-op (`"a late release can't wipe the new holder"`).

muhammad-ali-e (commented):
## Automated PR review — UN-3553 leader-election lease

Reviewed via the PR Review Toolkit (Code Reviewer, Silent Failure Hunter, Type Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier). Overall this is a clean, well-documented, correctness-focused PR — the lease SQL is split-brain-safe (single atomic conditional `UPDATE`, server-side `now()`, holder-scoped renew/release) and the semantics are well tested.

The one finding worth acting on before merge — flagged independently by four of the six agents — is the **missing `_owns_conn` ownership guard** (see inline comment on `__init__`): `LeaderLease` drops the connection-ownership distinction its sibling `PgQueueClient` treats as essential, so a transient DB error silently closes a caller-injected connection and reconnects against the default `DB_*` database. Everything else is Low severity. Inline comments below.

backend/pg_queue/migrations/0005_pgorchestratorlock.py
```diff
@@ -0,0 +1,46 @@
...
+                "db_table": "pg_orchestrator_lock",
+            },
+        ),
+        migrations.AddConstraint(
+            model_name="pgorchestratorlock",
+            constraint=models.CheckConstraint(
+                check=models.Q(("id", 1)), name="pg_orchestrator_lock_single_row"
+            ),
+        ),
+        migrations.RunPython(seed_single_row, unseed_single_row),
```
muhammad-ali-e, line 49: **[Low] The seed row is load-bearing — worth a note for the bootstrap slice.** _(Type-Design Analyzer)_

Leader-election is a pure `UPDATE ... WHERE id = 1`; if this row is ever absent (e.g. a future squash that drops this `RunPython`, or a table recreate), *every* candidate's `try_acquire` silently returns `False` forever — permanent leaderlessness with no error. The test fixture defends with a defensive upsert (`test_leader_election.py:99-105`) but production has no such guard. Consider having the future reaper-bootstrap slice do an idempotent `INSERT ... ON CONFLICT (id) DO NOTHING` on startup as self-heal. No model/migration change needed now — just flagging the dependency.

workers/queue_backend/pg_queue/leader_election.py
muhammad-ali-e: Fixed (commit cca2741bf). Added `self._owns_conn = conn is None`; `_cursor` only closes/nulls the handle `if self._owns_conn and (conn_dead or conn.closed)`, and `_get_conn` only recreates an owned missing/closed conn — an injected dead connection now raises rather than being silently swapped for a `DB_`-env one. New test `test_injected_conn_never_swapped_on_error` pins it.

workers/queue_backend/pg_queue/leader_election.py
muhammad-ali-e: Fixed (cca2741bf). The owned-conn discard branch now emits `logger.warning` with `worker_id` + the exception type before dropping the handle.

workers/queue_backend/pg_queue/leader_election.py
muhammad-ali-e: Applied (cca2741bf). Scoped the guard to the explicit-arg branch; the env path defers to `lease_seconds_from_env()` which already rejects `<= 0`.

workers/queue_backend/pg_queue/leader_election.py
muhammad-ali-e: Done (cca2741bf). Reframed the module docstring to state the contract explicitly (a raised DB error == "leadership unknown, stop acting", same as `renew()` → False), wrapped the Usage example accordingly, and added the note to `renew()`.

workers/queue_backend/pg_queue/leader_election.py
muhammad-ali-e: Fixed (cca2741bf). `release()` now branches on `cur.rowcount`: `info("released leadership")` only when it actually freed the lease, else `debug("release was a no-op (not the current holder)")`. Covered by `test_release_after_takeover_is_noop`.

workers/queue_backend/pg_queue/leader_election.py
muhammad-ali-e: Done (cca2741bf). Relabelled to `Usage:` and reframed the intro durably — "This module is the lease mechanism — acquiring/driving the lease is the caller's responsibility (the reaper loop)" — so it stays accurate once the reaper slice lands.

workers/tests/test_leader_election.py
muhammad-ali-e: Added (cca2741bf). New `TestConnectionRecovery` (mocked, no DB): owned-conn recovered on `OperationalError`, owned-conn recovered when rollback fails, injected-conn never swapped. Plus the two invariant gaps — `test_same_holder_reacquire_while_fresh_returns_false` and `test_release_after_takeover_is_noop`. 25 tests total.

backend/pg_queue/migrations/0005_pgorchestratorlock.py
muhammad-ali-e: Good catch — noted in the migration docstring (cca2741bf): flagged that the row is load-bearing and that the future reaper-bootstrap slice should self-heal with an idempotent `INSERT ... ON CONFLICT (id) DO NOTHING` on startup. No model/migration change now, tracked for that slice.

muhammad-ali-e:
## Review addressed — commit `cca2741bf`

All 8 toolkit findings + the 2 SonarCloud issues are in. Validated each against the code first; all were valid.

**High / Medium (behaviour + coverage)**
- **`_owns_conn` ownership guard** — mirrors `PgQueueClient`. An injected connection is never closed/swapped on a transient error; only an *owned* missing/closed conn reconnects. This closes the real test-integrity hole (an injected `TEST_DB_` conn silently re-pointing at `DB_`). New test pins it.
- **Silent reconnect** — the owned-conn discard now logs `warning` (worker_id + exc type).
- **Recovery path untested** — new `TestConnectionRecovery` (mocked): recovered-on-OperationalError, recovered-when-rollback-fails, injected-never-swapped. Plus two invariant gaps (same-holder re-acquire → False; release-after-takeover no-op). **20 → 25 tests.**

**Low (clarity / truthfulness)**
- `release()` logs "released" only when `rowcount == 1`; a no-op logs `debug`.
- Scoped the `lease_seconds <= 0` guard to the explicit-arg branch.
- Documented the exception-propagation contract (raise == "leadership unknown, stop acting") + durable `Usage:` framing.
- Migration: noted the seed row is load-bearing → future reaper-bootstrap should self-heal with `INSERT ... ON CONFLICT DO NOTHING`.

**SonarCloud S117** — renamed the migration's `get_model` local to `lock_model` (both hits).

`makemigrations --check` clean; 25/25 leader-election tests pass; pre-commit green.

workers/queue_backend/pg_queue/leader_election.py
```diff
@@ -0,0 +1,256 @@
...
+                "WHERE id = 1 AND leader = %s RETURNING id",
+                (self._worker_id,),
+            )
+            still_leader = cur.fetchone() is not None
+        if not still_leader:
+            logger.warning(
+                "LeaderLease: %s renew failed — leadership lost (took over by "
+                "another candidate after a stale lease)",
+                self._worker_id,
+            )
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed (a90c12618). Reworded to "not the current leader (taken over by another candidate, or the lease was never held)" — accurate for the non-holder renew() case too, plus the "took over"→"taken over" grammar.

workers/queue_backend/pg_queue/leader_election.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+def default_worker_id() -> str:
+    """A stable-per-process candidate id (``host:pid:rand``).
+
+    Stable for the process lifetime so ``renew``/``release`` match the row this
+    process wrote; unique across processes (the random suffix disambiguates two
+    candidates that share a host+pid across container restarts).
+    """
+    return f"{socket.gethostname()}:{os.getpid()}:{uuid.uuid4().hex[:8]}"
```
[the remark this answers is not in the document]
muhammad-ali-e: Good catch — fixed (a90c12618). default_worker_id() is now @functools.cache, so it returns the same id for the life of the process even if called inline in a retry/restart loop; renew()/release() can no longer drift off the row the process wrote. Kept it lazy (first-call, not import-time) so the id is fixed after any fork rather than shared across children. Test updated to assert idempotency.

muhammad-ali-e:
## Greptile review addressed — commit `a90c12618`

Both P2 findings validated and fixed:

- **`default_worker_id()` idempotency footgun** — now `@functools.cache`'d, so repeated/inline calls within a process return the same id (a retry loop can't drift the worker id out from under `renew()`/`release()`). Kept lazy/first-call so it's fixed after a fork, not shared across children. Test updated to assert idempotency.
- **`renew()` warning misleading + grammar** — reworded to "not the current leader (taken over by another candidate, or the lease was never held)", which reads correctly for the non-holder-renew path too.

25/25 tests pass; pre-commit green.

PR #2058, opened 2026-06-16
UN-3554 [FEAT] PG Queue 9d slice 2 — reaper process + barrier-orphan sweep

muhammad-ali-e:
## What

Second slice of **9d**. Builds on the leader-election lease (UN-3553) to stand up the **reaper process** — the leader-elected recovery loop — with its first recovery job, the **barrier-orphan sweep**. **Ships dark**: launched explicitly, never part of the default worker set, and a no-op until the PG barrier is used.

## Why now (and why not folded into 9e)

9d is 9e's safety net. The reaper *harness* (lease-maintenance loop + graceful shutdown + entrypoint) is **transport-agnostic** — identical regardless of what recovery runs inside it — so building it now isn't throwaway; 9e just drops recovery functions into a proven loop. Keeping it out of 9e shrinks that phase's blast radius, and the barrier sweep makes `WORKER_BARRIER_BACKEND=pg` production-safe independently.

## What's in it

**`workers/queue_backend/pg_queue/reaper.py`**
- `sweep_expired_barriers(conn)` — `DELETE FROM pg_barrier_state WHERE expires_at < now() RETURNING execution_id` + a loud per-orphan `WARNING`. The documented `PgBarrier` backstop: reclaims a barrier whose header tasks never all completed. By PgBarrier's existing semantics, a late in-flight decrement then finds no row and abandons (no spurious callback). Marking the owning execution *terminal* needs the backend + the pipeline's PG shape — that's 9e.
- `PgReaper` — leader-elected loop. Each cycle: renew (step down to standby if `renew()` returns `False`), else try to acquire; sweep **only while leader**. `run()` loops with graceful SIGTERM/SIGINT shutdown + lease release on exit. Guard: cycle interval **must be shorter than the lease window**, or the leader thrashes leadership between renews.
- `reaper_interval_from_env()` (`WORKER_PG_REAPER_INTERVAL_SECONDS`, default 5s), `main()`, and a `python -m queue_backend.pg_queue.reaper` entrypoint (no worker-app bootstrap needed — it's pure SQL).

**`leader_election.py`** — exposes a `lease_seconds` property (the reaper validates its cycle against it).

**Tests (`test_pg_reaper.py`)** — 15:
- env/construction guards (interval-shorter-than-lease, non-positive rejected);
- leadership gating (sweeps only while leader; steps down on renew-fail; releases the lease on stop);
- real-PG sweep (reclaims only rows past `expires_at`, leaves fresh barriers intact; no-op when nothing expired).

## Out of scope (follow-ups)

- `run-worker.sh` wiring + liveness probe — separate followup (mirrors 9c → 9c-followup).
- **Pipeline recovery** — counter reconstruction from `WorkflowFileExecution`, per-stage re-enqueue of stuck files, marking orphaned executions terminal — deferred to 9e, where there's a real PG pipeline to test against. This slice is the storage/orphan backstop only.

## Testing

15/15 pass, stable across reruns; 70/70 across reaper + leader-election + pg_barrier (no regression). Ruff clean. Pre-commit green.

Base: `feat/UN-3445-pg-queue-integration` (not `main`). Sub-task UN-3554 under UN-3536 (Phase 9), builds on UN-3553.

### DIFF
--- a/workers/queue_backend/pg_queue/reaper.py
+++ b/workers/queue_backend/pg_queue/reaper.py
@@ -269,6 +269,13 @@ class PgReaper:
                         self._lease.lease_seconds,
                         exc_info=True,
                     )
+            # Close our owned sweep connection (an injected one is the caller's).
+            # Harmless for the main() process — the OS reclaims it — but keeps the
+            # class clean if it's ever embedded / driven from a test.
+            if self._owns_sweep_conn and self._sweep_conn is not None:
+                with contextlib.suppress(Exception):
+                    self._sweep_conn.close()
+                self._sweep_conn = None
             logger.info("Reaper stopped")
 
     def stop(self, *_: object) -> None:
--- a/workers/queue_backend/pg_queue/reaper.py
+++ b/workers/queue_backend/pg_queue/reaper.py
@@ -20,10 +20,10 @@ defined against the coupled pipeline running on PG, which doesn't exist yet — 
 it lands with 9e, against a real PG pipeline it can be tested on.
 
 **Lease maintenance.** Each cycle the leader renews; if ``renew()`` returns
-``False`` it lost the lease (stalled past the window) and steps down to standby.
-A standby tries to acquire each cycle. The cycle interval MUST be shorter than
-the lease window, or the leader would lose the lease between renews — enforced in
-:meth:`PgReaper.__init__`.
+``False`` (or raises) it lost / can't confirm the lease and steps down to
+standby. A standby tries to acquire each cycle. The cycle interval MUST be
+shorter than the lease window, or the leader would lose the lease between
+renews — enforced in :meth:`PgReaper.__init__`.
 
 **Ships dark.** Launched explicitly (``python -m queue_backend.pg_queue.reaper``
 or, later, ``run-worker.sh``); never part of the default worker set. With
@@ -37,8 +37,9 @@ import contextlib
 import logging
 import os
 import signal
+import threading
 import time
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, NamedTuple, Protocol
 
 from .connection import create_pg_connection
 from .leader_election import LeaderLease, default_worker_id
@@ -48,10 +49,41 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-# Cycle cadence: how often the leader renews + runs recovery. Must be shorter
-# than the lease window (PgReaper enforces it). The sweep is a cheap indexed
-# DELETE, so running it every cycle is fine.
+# Cadence: how often the leader renews + runs recovery. Enforced shorter than
+# the lease window in PgReaper.__init__.
 _DEFAULT_REAPER_INTERVAL_SECONDS = 5.0
+
+
+class LeaderLeaseLike(Protocol):
+    """Structural contract :class:`PgReaper` needs from a lease.
+
+    The dependency is structural (the tests substitute a duck-typed fake), so the
+    param is typed against this Protocol — :class:`LeaderLease` satisfies it, and
+    a fake conforms without inheritance. Same convention as the ``Barrier``
+    Protocol elsewhere in the package.
+    """
+
+    @property
+    def lease_seconds(self) -> int: ...
+
+    @property
+    def worker_id(self) -> str: ...
+
+    def try_acquire(self) -> bool: ...
+
+    def renew(self) -> bool: ...
+
+    def release(self) -> None: ...
+
+
+class TickOutcome(NamedTuple):
+    """Result of one :meth:`PgReaper.tick` — keeps "was I leader" and "how much
+    work" on separate channels (an ``int`` sentinel like ``-1`` is truthy and
+    conflates the two).
+    """
+
+    was_leader: bool
+    reclaimed: int  # 0 when standby
 
 
 def reaper_interval_from_env() -> float:
@@ -81,17 +113,29 @@ def reaper_interval_from_env() -> float:
 def sweep_expired_barriers(conn: PgConnection) -> list[str]:
     """Reclaim ``pg_barrier_state`` rows past ``expires_at``. Returns their ids.
 
-    A single ``DELETE … RETURNING`` — atomic, and safe to race (the orchestrator
-    is a singleton anyway). Each reclaimed barrier is logged at WARNING: an
-    orphaned barrier means an execution's header tasks never all completed, which
-    is worth surfacing even though deleting the row is the correct backstop.
+    A single atomic ``DELETE … RETURNING``: concurrent sweepers would each
+    reclaim a disjoint subset (``RETURNING`` reports only the rows *this*
+    statement deleted), so it stays correct even if leadership gating ever fails —
+    in practice only the leader calls it. Each reclaimed barrier is logged at
+    WARNING: an orphaned barrier means an execution's header tasks never all
+    completed, worth surfacing even though deleting the row is the right backstop.
+
+    ``conn`` runs in manual-commit mode, so on any error we roll back before
+    re-raising — otherwise the connection is left in an aborted-transaction state
+    and every later statement on it fails with ``InFailedSqlTransaction``.
     """
-    with conn.cursor() as cur:
-        cur.execute(
-            "DELETE FROM pg_barrier_state WHERE expires_at < now() RETURNING execution_id"
-        )
-        reclaimed = [row[0] for row in cur.fetchall()]
-    conn.commit()
+    try:
+        with conn.cursor() as cur:
+            cur.execute(
+                "DELETE FROM pg_barrier_state WHERE expires_at < now() "
+                "RETURNING execution_id"
+            )
+            reclaimed = [row[0] for row in cur.fetchall()]
+        conn.commit()
+    except Exception:
+        with contextlib.suppress(Exception):
+            conn.rollback()
+        raise
     for execution_id in reclaimed:
         logger.warning(
             "Reaper: reclaimed orphaned barrier for execution %s — header tasks "
@@ -107,7 +151,7 @@ class PgReaper:
 
     def __init__(
         self,
-        lease: LeaderLease,
+        lease: LeaderLeaseLike,
         *,
         interval_seconds: float | None = None,
         sweep_conn: PgConnection | None = None,
@@ -118,6 +162,8 @@ class PgReaper:
             if interval_seconds is not None
             else reaper_interval_from_env()
         )
+        # Load-bearing even though reaper_interval_from_env validates: an
+        # explicitly-injected interval_seconds<=0 reaches here unvalidated.
         if self._interval <= 0:
             raise ValueError("interval_seconds must be positive")
         if self._interval >= lease.lease_seconds:
@@ -133,32 +179,60 @@ class PgReaper:
         self._running = False
         self._is_leader = False
 
+    @property
+    def is_leader(self) -> bool:
+        """Whether this process currently holds leadership (last tick's view)."""
+        return self._is_leader
+
     def _get_sweep_conn(self) -> PgConnection:
-        # Owned sweep connection self-recovers; an injected one is the caller's.
+        # Recreate only an OWNED missing/closed connection; an injected one is the
+        # caller's and is never swapped (mirrors LeaderLease / PgQueueClient).
         if self._sweep_conn is None or (
             self._owns_sweep_conn and self._sweep_conn.closed
         ):
             self._sweep_conn = create_pg_connection(env_prefix="DB_")
         return self._sweep_conn
 
-    def tick(self) -> int:
-        """One cycle: maintain leadership, then sweep iff leader.
-
-        Returns the number of barriers reclaimed, or ``-1`` if this process is a
-        standby (not the leader) this cycle.
-        """
-        if self._is_leader and not self._lease.renew():
+    def _discard_owned_sweep_conn(self) -> None:
+        # After a sweep error, drop an owned connection so the next tick
+        # reconnects — covers a poisoned (aborted-txn) or dead-socket handle that
+        # `.closed` alone wouldn't catch.
+        if self._owns_sweep_conn and self._sweep_conn is not None:
+            with contextlib.suppress(Exception):
+                self._sweep_conn.close()
+            self._sweep_conn = None
             logger.warning(
-                "Reaper: lost leadership (lease taken over) — stepping down to "
-                "standby; will retry acquire next cycle"
+                "Reaper: discarded sweep connection after a failed sweep; "
+                "reconnecting next cycle"
             )
-            self._is_leader = False
+
+    def tick(self) -> TickOutcome:
+        """One cycle: maintain leadership, then sweep iff leader."""
+        if self._is_leader:
+            try:
+                still_leader = self._lease.renew()
+            except Exception:
+                # A raised renew == "leadership unknown": stop acting (honour the
+                # lease's documented contract) before letting it propagate.
+                self._is_leader = False
+                raise
+            if not still_leader:
+                logger.warning(
+                    "Reaper: lost leadership (lease taken over) — stepping down "
+                    "to standby"
+                )
+                self._is_leader = False
         if not self._is_leader and self._lease.try_acquire():
             self._is_leader = True
             logger.info("Reaper: acquired leadership")
         if not self._is_leader:
-            return -1
-        return len(sweep_expired_barriers(self._get_sweep_conn()))
+            return TickOutcome(was_leader=False, reclaimed=0)
+        try:
+            reclaimed = len(sweep_expired_barriers(self._get_sweep_conn()))
+        except Exception:
+            self._discard_owned_sweep_conn()
+            raise
+        return TickOutcome(was_leader=True, reclaimed=reclaimed)
 
     def run(self, *, install_signals: bool = True) -> None:
         """Lease-maintenance + recovery loop until stopped; releases on exit."""
@@ -176,16 +250,25 @@ class PgReaper:
                 try:
                     self.tick()
                 except Exception:
-                    # A transient DB blip must not tear the loop down — the
-                    # connections self-recover, so log and keep cycling.
+                    # A transient DB blip must not tear the loop down — the lease
+                    # connection rolls back + discards a dead handle, and a failed
+                    # sweep discards its owned connection, so log and keep cycling.
                     logger.exception("Reaper: cycle failed; continuing")
                 time.sleep(self._interval)
         finally:
             # Hand the lease over promptly so a standby takes leadership without
             # waiting out the full lease window.
             if self._is_leader:
-                with contextlib.suppress(Exception):
+                try:
                     self._lease.release()
+                except Exception:
+                    logger.warning(
+                        "Reaper: failed to release lease on shutdown; a standby "
+                        "will take over after the lease window (~%ss) instead of "
+                        "immediately",
+                        self._lease.lease_seconds,
+                        exc_info=True,
+                    )
             logger.info("Reaper stopped")
 
     def stop(self, *_: object) -> None:
@@ -193,11 +276,14 @@ class PgReaper:
         self._running = False
 
     def _install_signal_handlers(self) -> None:
-        # signal.signal only works in the main thread.
         try:
             signal.signal(signal.SIGTERM, self.stop)
             signal.signal(signal.SIGINT, self.stop)
         except ValueError:
+            # signal.signal raises ValueError off the main thread — assert that
+            # cause rather than mislabelling an unrelated ValueError.
+            if threading.current_thread() is threading.main_thread():
+                raise
             logger.warning(
                 "Reaper: signal handlers not installed (non-main thread) — "
                 "SIGTERM/SIGINT will not trigger graceful shutdown"
--- a/workers/queue_backend/pg_queue/reaper.py
+++ b/workers/queue_backend/pg_queue/reaper.py
@@ -1,0 +1,214 @@
+"""Reaper — the leader-elected recovery process for the PG queue.
+
+A singleton, guarded by :class:`LeaderLease` over ``pg_orchestrator_lock``: only
+the elected leader runs recovery work each cycle (several reapers would contend
+and double-act). This slice ships the process *harness* (lease-maintenance loop
++ graceful shutdown) plus ONE recovery job — the **barrier-orphan sweep**.
+
+**Barrier-orphan sweep.** Reclaims ``pg_barrier_state`` rows past their
+``expires_at`` — a barrier whose header tasks never all completed (the documented
+:class:`PgBarrier` backstop). It ``DELETE``s the orphaned row; by PgBarrier's
+existing semantics a late in-flight decrement then finds no row and abandons (no
+spurious callback). The owning execution is logged loudly. Marking that
+execution *terminal* (ERROR) is recovery that needs the backend and the
+pipeline's PG shape — that's 9e, not here; this slice is the storage/orphan
+backstop only.
+
+**Deferred to 9e.** Pipeline recovery (counter reconstruction from
+``WorkflowFileExecution``, per-stage re-enqueue of stuck file executions) is
+defined against the coupled pipeline running on PG, which doesn't exist yet — so
+it lands with 9e, against a real PG pipeline it can be tested on.
+
+**Lease maintenance.** Each cycle the leader renews; if ``renew()`` returns
+``False`` it lost the lease (stalled past the window) and steps down to standby.
+A standby tries to acquire each cycle. The cycle interval MUST be shorter than
+the lease window, or the leader would lose the lease between renews — enforced in
+:meth:`PgReaper.__init__`.
+
+**Ships dark.** Launched explicitly (``python -m queue_backend.pg_queue.reaper``
+or, later, ``run-worker.sh``); never part of the default worker set. With
+``WORKER_BARRIER_BACKEND`` left at ``chord`` (default) there are no
+``pg_barrier_state`` rows, so the sweep is a no-op until the PG barrier is used.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import logging
+import os
+import signal
+import time
+from typing import TYPE_CHECKING
+
+from .connection import create_pg_connection
+from .leader_election import LeaderLease, default_worker_id
+
+if TYPE_CHECKING:
+    from psycopg2.extensions import connection as PgConnection
+
+logger = logging.getLogger(__name__)
+
+# Cycle cadence: how often the leader renews + runs recovery. Must be shorter
+# than the lease window (PgReaper enforces it). The sweep is a cheap indexed
+# DELETE, so running it every cycle is fine.
+_DEFAULT_REAPER_INTERVAL_SECONDS = 5.0
+
+
+def reaper_interval_from_env() -> float:
+    """Cycle interval from ``WORKER_PG_REAPER_INTERVAL_SECONDS`` (default 5s).
+
+    Read at call time. Invalid / non-positive values raise (loud-on-misconfig,
+    matching the lease/barrier-TTL posture).
+    """
+    raw = os.getenv("WORKER_PG_REAPER_INTERVAL_SECONDS")
+    if raw is None:
+        return _DEFAULT_REAPER_INTERVAL_SECONDS
+    try:
+        value = float(raw)
+    except ValueError as exc:
+        raise ValueError(
+            f"WORKER_PG_REAPER_INTERVAL_SECONDS={raw!r} is not a number. "
+            f"Unset it to default to {_DEFAULT_REAPER_INTERVAL_SECONDS}s."
+        ) from exc
+    if value <= 0:
+        raise ValueError(
+            f"WORKER_PG_REAPER_INTERVAL_SECONDS={value} must be positive. "
+            f"Unset it to default to {_DEFAULT_REAPER_INTERVAL_SECONDS}s."
+        )
+    return value
+
+
+def sweep_expired_barriers(conn: PgConnection) -> list[str]:
+    """Reclaim ``pg_barrier_state`` rows past ``expires_at``. Returns their ids.
+
+    A single ``DELETE … RETURNING`` — atomic, and safe to race (the orchestrator
+    is a singleton anyway). Each reclaimed barrier is logged at WARNING: an
+    orphaned barrier means an execution's header tasks never all completed, which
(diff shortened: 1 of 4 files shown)


workers/queue_backend/pg_queue/reaper.py
```diff
@@ -0,0 +1,214 @@
...
+    is a singleton anyway). Each reclaimed barrier is logged at WARNING: an
+    orphaned barrier means an execution's header tasks never all completed, which
+    is worth surfacing even though deleting the row is the correct backstop.
+    """
+    with conn.cursor() as cur:
+        cur.execute(
+            "DELETE FROM pg_barrier_state WHERE expires_at < now() RETURNING execution_id"
+        )
+        reclaimed = [row[0] for row in cur.fetchall()]
+    conn.commit()
```
muhammad-ali-e: **[CRITICAL] Missing rollback — a failed sweep poisons the connection for the whole process lifetime.**

`create_pg_connection` runs in manual-commit mode (no autocommit). If the `DELETE` or this `conn.commit()` raises (lock/statement timeout, server-side termination, PgBouncer recycle mid-statement), the exception propagates with **no `rollback()`**. psycopg2 leaves the connection open but in an aborted-transaction state (`closed == False`), so `_get_sweep_conn` hands the same broken handle back next cycle and every subsequent statement raises `InFailedSqlTransaction`. The `run()` loop then logs "cycle failed; continuing" forever and reclaims nothing — a silent, self-perpetuating stall that needs a process restart to clear. This is the one place the file departs from the package's `_cursor()` rollback-and-discard pattern (`leader_election.py:161-194`, `pg_barrier.py`).

Suggested fix — mirror `LeaderLease._cursor`:
```python
try:
    with conn.cursor() as cur:
        cur.execute(
            "DELETE FROM pg_barrier_state WHERE expires_at < now() RETURNING execution_id"
        )
        reclaimed = [row[0] for row in cur.fetchall()]
    conn.commit()
except Exception:
        conn.rollback()
    raise
```
(harmless under the autocommit test connection).

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -0,0 +1,214 @@
...
+            )
+        self._sweep_conn = sweep_conn
+        self._owns_sweep_conn = sweep_conn is None
+        self._running = False
+        self._is_leader = False
+
+    def _get_sweep_conn(self) -> PgConnection:
+        # Owned sweep connection self-recovers; an injected one is the caller's.
+        if self._sweep_conn is None or (
+            self._owns_sweep_conn and self._sweep_conn.closed
```
muhammad-ali-e, line 191: **[HIGH] `_get_sweep_conn` only rebuilds on `.closed`, not on a poisoned (aborted-transaction) connection.**

After an un-rolled-back error (see the CRITICAL on line 94), a psycopg2 connection is in `InFailedSqlTransaction` state but `closed == 0`, so this predicate keeps returning the same unusable handle forever. The comment "self-recovers" is therefore false for the most common failure mode (mid-statement error) — only a fully dropped socket is recovered.

The `rollback()` from the CRITICAL fix is the primary remedy (a successful rollback returns the connection to a usable state). As defence-in-depth, on a sweep failure drop the owned connection (`self._sweep_conn = None` after `close()`) so the next tick reconnects, logging the discard like `leader_election.py:184-193` rather than swallowing it.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -0,0 +1,214 @@
...
+            self._lease.lease_seconds,
+            self._lease.worker_id,
+        )
+        try:
+            while self._running:
+                try:
+                    self.tick()
+                except Exception:
+                    # A transient DB blip must not tear the loop down — the
+                    # connections self-recover, so log and keep cycling.
```
muhammad-ali-e: **[LOW · comment accuracy] "the connections self-recover" overstates recovery.**

The *lease* connection self-recovers (its `_cursor()` rolls back + discards a dead owned connection). The *sweep* connection does **not** — see the CRITICAL/HIGH above. Once those are fixed this comment becomes true; if the fix is deferred, correct the comment so it doesn't discourage a maintainer from adding the missing rollback, e.g.: "…the lease connection rolls back and self-recovers; the sweep connection only rebuilds when the socket actually drops."

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -0,0 +1,214 @@
...
+            self._sweep_conn = create_pg_connection(env_prefix="DB_")
+        return self._sweep_conn
+
+    def tick(self) -> int:
+        """One cycle: maintain leadership, then sweep iff leader.
+
+        Returns the number of barriers reclaimed, or ``-1`` if this process is a
+        standby (not the leader) this cycle.
+        """
+        if self._is_leader and not self._lease.renew():
```
muhammad-ali-e: **[MEDIUM] `renew()` raising leaves `_is_leader = True` — violates the lease's documented "on renew error, step down" contract.**

The `leader_election` module documents that a *raised* exception from `renew()` must be treated like `False` (leadership unknown → stop acting). Here, if `self._lease.renew()` raises, the exception propagates out of `tick()` and is swallowed by `run()`'s broad handler, but `_is_leader` is never flipped to `False` — the node keeps believing it is leader. Today the sweep for that cycle is skipped (the raise precedes the sweep), so no immediate double-act, but the documented invariant is broken and a future edit that does work before the renew would be exposed. Wrap the renew so any exception sets `self._is_leader = False` before propagating.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -0,0 +1,214 @@
...
+                except Exception:
+                    # A transient DB blip must not tear the loop down — the
+                    # connections self-recover, so log and keep cycling.
+                    logger.exception("Reaper: cycle failed; continuing")
+                time.sleep(self._interval)
+        finally:
+            # Hand the lease over promptly so a standby takes leadership without
+            # waiting out the full lease window.
+            if self._is_leader:
+                with contextlib.suppress(Exception):
```
muhammad-ali-e: **[MEDIUM] Silent `contextlib.suppress(Exception)` around `lease.release()` hides a failed hand-off.**

Not crashing shutdown is right, but suppressing *silently* means that if release fails (DB down at shutdown, dead connection), nothing is logged — the operator believes a standby took over in <1s when it must actually wait out the full lease window. Log it instead:
```python
try:
    self._lease.release()
except Exception:
    logger.warning(
        "Reaper: failed to release lease on shutdown; a standby will take over "
        "after the lease window (~%ss) instead of immediately",
        self._lease.lease_seconds, exc_info=True,
    )
```

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -0,0 +1,214 @@
...
+    def stop(self, *_: object) -> None:
+        """Request a graceful stop after the current cycle."""
+        self._running = False
+
+    def _install_signal_handlers(self) -> None:
+        # signal.signal only works in the main thread.
+        try:
+            signal.signal(signal.SIGTERM, self.stop)
+            signal.signal(signal.SIGINT, self.stop)
+        except ValueError:
```
muhammad-ali-e, line 289: **[MEDIUM] `except ValueError` is broader than the documented cause and mislabels unrelated `ValueError`s.**

`signal.signal` raises `ValueError` specifically when called off the main thread, and the warning asserts that cause — but this `except` will also catch a `ValueError` from any future edit inside the `try` and report it as "non-main thread", misleading a post-mortem. Verify before asserting the cause:
```python
except ValueError:
    if threading.current_thread() is not threading.main_thread():
        logger.warning("Reaper: signal handlers not installed (non-main thread) — …")
    else:
        raise
```

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def _get_sweep_conn(self) -> PgConnection:
+        # Owned sweep connection self-recovers; an injected one is the caller's.
+        if self._sweep_conn is None or (
+            self._owns_sweep_conn and self._sweep_conn.closed
+        ):
+            self._sweep_conn = create_pg_connection(env_prefix="DB_")
+        return self._sweep_conn
+
+    def tick(self) -> int:
```
muhammad-ali-e: **[MEDIUM · type design] `tick() -> int` overloads the return with a `-1` standby sentinel.**

`-1` conflates two orthogonal facts — *was I the leader* and *how much work did I do* — into one channel, and `-1` is truthy, so the natural `if reaper.tick():` reads standby as success. The contract lives only in the docstring; the signature actively misleads. The sole in-tree caller (`run()`) discards the value, so the sentinel exists today only for tests. Prefer a small result type:
```python
class TickOutcome(NamedTuple):
    was_leader: bool
    reclaimed: int  # 0 when standby
```
Return `TickOutcome(False, 0)` for standby, `TickOutcome(True, n)` when leader. Next-best, `-> int | None` (None = standby) at least removes the truthiness trap.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -0,0 +1,214 @@
...
+            # between renews → it would thrash leadership every cycle.
+            raise ValueError(
+                f"reaper interval {self._interval}s must be shorter than the "
+                f"lease window {lease.lease_seconds}s, or the leader loses the "
+                f"lease between renews"
+            )
+        self._sweep_conn = sweep_conn
+        self._owns_sweep_conn = sweep_conn is None
+        self._running = False
+        self._is_leader = False
```
muhammad-ali-e, line 180: **[MEDIUM · type design] No public way to observe leadership; tests mutate `_is_leader` directly.**

Leadership is a legitimate observable (health/liveness endpoints in this epic will want it), but the only public read path is the overloaded `tick()` sentinel — which is why `test_pg_reaper.py:115` pokes `reaper._is_leader = True` to set up the step-down case. Add `@property def is_leader(self) -> bool: return self._is_leader`; combined with the `TickOutcome.was_leader` suggestion, the test can drive the documented path instead of mutating a private flag.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -0,0 +1,214 @@
...
+        )
+    return reclaimed
+
+
+class PgReaper:
+    """Leader-elected recovery loop. Only the lease holder runs recovery work."""
+
+    def __init__(
+        self,
+        lease: LeaderLease,
```
muhammad-ali-e: **[MEDIUM · type design] `lease: LeaderLease` should be a `Protocol` — the real contract is structural.**

`PgReaper` only ever calls `try_acquire()`, `renew()`, `release()` and reads `lease_seconds`/`worker_id`. The tests already prove the dependency is structural — `_FakeLease` is a full duck-typed reimplementation with no inheritance, which a type-checker would (wrongly) reject against the concrete annotation. The package already uses this convention (`Barrier` protocol in `handle.py`). Define a `LeaderLeaseLike(Protocol)` with that minimal surface and type the param against it; `LeaderLease` satisfies it structurally and `_FakeLease` becomes a checked conformer.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -0,0 +1,214 @@
...
+        """One cycle: maintain leadership, then sweep iff leader.
+
+        Returns the number of barriers reclaimed, or ``-1`` if this process is a
+        standby (not the leader) this cycle.
+        """
+        if self._is_leader and not self._lease.renew():
+            logger.warning(
+                "Reaper: lost leadership (lease taken over) — stepping down to "
+                "standby; will retry acquire next cycle"
+            )
```
muhammad-ali-e, line 207: **[LOW · comment/behaviour mismatch] Log says "will retry acquire next cycle" but acquire is retried in the *same* cycle.**

After the failed `renew()` sets `_is_leader = False`, the very next `if` (line 156) calls `try_acquire()` within the same `tick()`. So a deposed leader can re-acquire in the same cycle if the lease freed up — the "next cycle" wording is inaccurate. Either drop "next cycle" from the message, or if same-cycle re-acquire is undesired, guard it (a behaviour change, out of scope for a wording fix).

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -0,0 +1,214 @@
...
+            f"Unset it to default to {_DEFAULT_REAPER_INTERVAL_SECONDS}s."
+        )
+    return value
+
+
+def sweep_expired_barriers(conn: PgConnection) -> list[str]:
+    """Reclaim ``pg_barrier_state`` rows past ``expires_at``. Returns their ids.
+
+    A single ``DELETE … RETURNING`` — atomic, and safe to race (the orchestrator
+    is a singleton anyway). Each reclaimed barrier is logged at WARNING: an
```
muhammad-ali-e: **[LOW · comment] "safe to race … the orchestrator is a singleton anyway" is self-contradictory and misattributes the guarantee.**

If it's a singleton there's no race to be safe against. More precisely, the singleton property comes from the *caller* (`tick()` only sweeps when leader), not from this standalone function — which is also called directly in tests with an injected connection. Reword to describe the function's own guarantee, e.g.: "A single atomic `DELETE … RETURNING`; concurrent sweepers would each reclaim a disjoint subset (RETURNING reports only the rows this statement deleted), so it is correct even if leadership gating ever fails — in practice only the leader calls it."

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -0,0 +1,214 @@
...
+        interval_seconds: float | None = None,
+        sweep_conn: PgConnection | None = None,
+    ) -> None:
+        self._lease = lease
+        self._interval = (
+            interval_seconds
+            if interval_seconds is not None
+            else reaper_interval_from_env()
+        )
+        if self._interval <= 0:
```
muhammad-ali-e, line 167: **[INFO · keep, don't strip] The `self._interval <= 0` check is load-bearing, not dead code.**

It looks redundant with `reaper_interval_from_env`, but that validator only runs on the `interval_seconds is None` path — an explicitly-injected `interval_seconds=0`/`-1` reaches `__init__` unvalidated, and `test_non_positive_interval_rejected` depends on this line. Keep it. (Minor: the constant comment on line 52 restates the lease-window rationale a third time; consider trimming it to just describe the cadence so the four copies don't drift.)

workers/tests/test_pg_reaper.py
```diff
@@ -0,0 +1,215 @@
...
+        lease = _FakeLease(acquires=True, renews=True)
+        reaper = self._reaper(lease)
+        with patch.object(reaper_mod, "sweep_expired_barriers", return_value=[]):
+            t = threading.Thread(target=reaper.run, kwargs={"install_signals": False})
+            t.start()
+            time.sleep(0.05)  # let it acquire + tick a few times
+            reaper.stop()
+            t.join(timeout=5)
+        assert not t.is_alive()
+        assert lease.released is True
```
muhammad-ali-e, line 161: **[TEST GAP · high value] `run()`'s exception-swallow path is untested.**

The loop's core resilience contract ("a transient DB blip must not tear the loop down… log and keep cycling", `reaper.py:174-182`) has zero coverage. A refactor that let an exception from `tick()` escape `run()` would kill the reaper silently with no test failing. Add a no-DB test: build a `PgReaper(interval_seconds=0.01)`, `patch.object` `tick` with a `side_effect` that raises once then sets `_running = False`, run `run(install_signals=False)` in a thread, and assert the thread exits cleanly and `logger.exception` fired.

workers/tests/test_pg_reaper.py
```diff
@@ -0,0 +1,215 @@
...
+    def test_valid_interval_accepted(self):
+        PgReaper(_FakeLease(lease_seconds=10), interval_seconds=3, sweep_conn=object())
+
+
+# --- Layer 2: leadership gating (fake lease + patched sweep, no DB) ---
+
+
+class TestLeadershipGating:
+    def _reaper(self, lease):
+        return PgReaper(lease, interval_seconds=0.01, sweep_conn=object())
```
muhammad-ali-e, line 102: **[TEST GAP] `_get_sweep_conn` self-recovery is untested — the sibling suite covers exactly this (`test_leader_election.py` `TestConnectionRecovery`).**

Two behaviourally distinct branches are unguarded: (a) *owned + closed → recreate* (a regression dropping the re-creation would sweep on a dead connection forever); (b) *injected → never swapped* (the safety property that an injected `TEST_DB_` connection is never silently replaced by a `DB_`-env one). Mirror `test_leader_election.py:99-135`: monkeypatch `reaper_mod.create_pg_connection` to a factory returning a mock with toggleable `.closed`, and assert the owned path re-calls the factory while the injected path returns the same object and never calls it.

muhammad-ali-e (commented):
## PR Review Toolkit — comprehensive review (#2058 · PG Queue 9d slice 2)

Ran six specialist passes (Code Reviewer, Silent-Failure Hunter, Type-Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier). The slice is well-built and closely follows the established `leader_election.py` conventions — leadership gating, the interval-vs-lease guard, signal-handler fallback, and release-on-exit are all sound and tested.

**One issue is worth blocking on** and was independently found by three passes: `sweep_expired_barriers` omits the rollback-and-discard discipline every other PG component in this package uses, so a single transient DB error permanently wedges the reaper while it keeps holding leadership and logging look-alike tracebacks. Verified: `create_pg_connection` runs in manual-commit mode (no autocommit), and `LeaderLease._cursor` is the canonical pattern this path should mirror.

Inline findings below, prioritised. Severity is prefixed on each comment.

workers/tests/test_pg_reaper.py
```diff
@@ -0,0 +1,215 @@
...
+    def test_steps_down_when_renew_fails(self):
+        # Already leader, but the lease was taken over → renew False → step down
+        # and (acquires=False) stay standby; no sweep this cycle.
+        lease = _FakeLease(acquires=False, renews=False)
+        reaper = self._reaper(lease)
+        reaper._is_leader = True
+        with patch.object(reaper_mod, "sweep_expired_barriers") as sweep:
+            assert reaper.tick() == -1
+        assert reaper._is_leader is False
+        sweep.assert_not_called()
```
muhammad-ali-e: **[TEST GAP] Step-down-then-*reacquire* across cycles is untested.**

`test_steps_down_when_renew_fails` only covers a single tick where the deposed leader stays standby (`acquires=False`). The operationally important transition — leader loses the lease one cycle, then re-acquires a later cycle and resumes sweeping — is never exercised, and the current `_FakeLease` can't script "renew fails once, then acquire succeeds". Extend `_FakeLease` to accept sequenced outcomes and assert tick 1 returns standby while tick 2 flips `_is_leader` back to True and sweeps.

workers/tests/test_pg_reaper.py
```diff
@@ -0,0 +1,215 @@
...
+    conn.autocommit = True
+    return conn
+
+
+@pytest.fixture
+def barrier_conn():
+    try:
+        conn = _new_conn()
+    except psycopg2.OperationalError as exc:
+        pytest.skip(f"Postgres not reachable: {exc}")
```
muhammad-ali-e, line 254: **[TEST GAP · CI] Layer 3 (real-Postgres) likely never runs in CI — confirm the workers job provisions PG + applies the barrier migration.**

The fixture `pytest.skip`s when PG is unreachable *or* the `pg_barrier_state` migration is unapplied, so the only *always-executing* coverage for this PR is Layers 1-2. If CI has no Postgres with the migration applied, the sweep semantics (`expires_at < now()` / `RETURNING execution_id`) contribute zero coverage and a regression in the predicate would pass. If CI can't guarantee a live DB, add one mocked-cursor assertion that pins the SQL contract without a real DB.

muhammad-ali-e (commented):
## PR Review Toolkit — comprehensive review (#2058 · PG Queue 9d slice 2)

Ran six specialist passes (Code Reviewer, Silent-Failure Hunter, Type-Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier). The slice is well-built and closely follows the established `leader_election.py` conventions — leadership gating, the interval-vs-lease guard, signal-handler fallback, and release-on-exit are all sound and tested.

**One issue is worth blocking on** and was independently found by three passes: `sweep_expired_barriers` omits the rollback-and-discard discipline every other PG component in this package uses, so a single transient DB error permanently wedges the reaper while it keeps holding leadership and logging look-alike tracebacks. Verified: `create_pg_connection` runs in manual-commit mode (no autocommit), and `LeaderLease._cursor` is the canonical pattern this path should mirror.

Inline findings below, prioritised. Severity is prefixed on each comment.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (59cc79f83). sweep_expired_barriers now wraps the DELETE+commit in try/except, rolls back before re-raising — mirrors LeaderLease._cursor. Covered by test_rolls_back_on_error. This also cleared SonarCloud C-reliability.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (59cc79f83). On a sweep error PgReaper now discards its OWNED sweep connection (close + null) so the next tick reconnects — covers the aborted-txn/dead-socket handle .closed misses. Covered by test_failed_sweep_discards_owned_conn.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Reworded (59cc79f83). The run() comment now reads that the lease connection rolls back and self-recovers, and a failed sweep discards its owned connection — accurate now that both recovery paths exist.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (59cc79f83). A raised renew() now sets _is_leader=False before propagating, honouring the lease contract (raise == leadership unknown, stop acting). Covered by test_renew_raising_steps_down.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (59cc79f83). release() failure on shutdown now logs a warning with the lease-window note instead of silently suppressing.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (59cc79f83). The signal-handler except now re-raises a ValueError unless we are off the main thread, so an unrelated ValueError is not mislabelled as non-main-thread.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Done (59cc79f83). tick() now returns TickOutcome(was_leader, reclaimed) NamedTuple — the truthy -1 sentinel is gone and the two facts are on separate channels. run() and tests updated.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Added (59cc79f83). is_leader @property added; tests now assert/drive via it instead of poking _is_leader (the step-down/reacquire cases are driven through ticks).

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Done (59cc79f83). Added LeaderLeaseLike(Protocol) with the minimal surface (try_acquire/renew/release + lease_seconds/worker_id); PgReaper types its param against it, so _FakeLease is now a checked structural conformer.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (59cc79f83). Dropped the inaccurate "will retry acquire next cycle" wording — a deposed leader can re-acquire in the same tick. test_steps_down_then_reacquires exercises the cross-cycle path.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Reworded (59cc79f83). The comment now describes the function guarantee: a single atomic DELETE RETURNING where concurrent sweepers reclaim disjoint subsets, correct even if gating ever fails; in practice only the leader calls it.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Kept the guard (test_non_positive_interval_rejected depends on it). Trimmed the line-52 constant comment so the lease-window rationale is not restated a fourth time.

workers/tests/test_pg_reaper.py
muhammad-ali-e: Added (59cc79f83). test_run_swallows_tick_exception: tick patched to raise once then stop, run() in a thread exits cleanly and logger.exception fired.

workers/tests/test_pg_reaper.py
muhammad-ali-e: Added (59cc79f83). test_owned_conn_recreated_when_closed (owned+closed -> factory re-called) and test_injected_conn_never_swapped (injected returned as-is even when closed) — mirrors test_leader_election TestConnectionRecovery.

workers/tests/test_pg_reaper.py
muhammad-ali-e: Added (59cc79f83). _FakeLease now accepts sequenced outcomes; test_steps_down_then_reacquires asserts tick1 leader -> tick2 standby -> tick3 re-acquires and sweeps.

workers/tests/test_pg_reaper.py
muhammad-ali-e: Addressed (59cc79f83). Added test_sql_contract (mocked cursor) pinning DELETE FROM pg_barrier_state / expires_at < now() / RETURNING execution_id + commit — so the predicate has coverage even where CI has no live Postgres. (Layer 4 still runs the real-DB path when available.)

muhammad-ali-e:
## Review addressed — commit `59cc79f83`

All toolkit findings + the SonarCloud reliability gate are in. Validated each against the code first; all were valid.

**CRITICAL / HIGH (the real bug)**
- **Missing rollback poisoned the sweep connection.** `create_pg_connection` is manual-commit, so an un-rolled-back DELETE/commit failure left the connection in an aborted-transaction state — and `_get_sweep_conn` (which only rebuilt on `.closed`) handed the same poisoned handle back forever → a silent self-perpetuating stall. Fixed: `sweep_expired_barriers` now rolls back before re-raising, **and** `PgReaper` discards its owned connection on a sweep error so the next tick reconnects. This is also what cleared **SonarCloud's C-reliability gate**.

**MEDIUM (correctness + design)**
- `renew()` raising now sets `_is_leader=False` before propagating (honours the lease's "raise == stop acting" contract).
- `release()` failure on shutdown is logged (with the lease-window note), not silently suppressed.
- signal-handler `except ValueError` re-raises unless we're off the main thread.
- `tick()` returns a `TickOutcome(was_leader, reclaimed)` NamedTuple — the truthy `-1` sentinel is gone; added an `is_leader` property; lease param typed against a new `LeaderLeaseLike(Protocol)`.

**LOW** — reworded the sweep race comment, the step-down log (same-cycle re-acquire), and the run() self-recovery comment; kept the load-bearing `interval<=0` guard, trimmed a duplicated rationale comment.

**Tests: 15 → 23.** Added: run() swallows a tick error; owned-conn recreated-when-closed; injected-conn never swapped; failed-sweep discards owned conn; sweep SQL contract (no-DB, so the predicate has coverage even without live PG); sweep rolls back on error; step-down-then-reacquire; renew-raising steps down. Paths driven via `is_leader`, no private-flag poking.

23/23 pass; 78/78 across reaper + leader-election + pg_barrier; ruff + pre-commit green.

muhammad-ali-e:
SonarCloud S1244 (float-equality) ×2 fixed in `ceb059b73` — the two reaper-interval asserts now use `pytest.approx(5.0)` / `pytest.approx(2.5)` instead of `== 5.0` / `== 2.5`. Values were exactly representable (harmless), but approx is the correct idiom and clears the reliability bugs. 23/23 tests still green.

workers/tests/test_pg_reaper.py
```diff
@@ -0,0 +1,320 @@
...
+    with conn.cursor() as cur:
+        cur.execute("SELECT to_regclass('pg_barrier_state')")
+        if cur.fetchone()[0] is None:
+            conn.close()
+            pytest.skip("pg_barrier_state migration not applied (run backend migrate)")
+        cur.execute("DELETE FROM pg_barrier_state")
+    yield conn
+    with conn.cursor() as cur:
+        cur.execute("DELETE FROM pg_barrier_state")
+    conn.close()
```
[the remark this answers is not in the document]
muhammad-ali-e: Good catch — fixed (7ea670313). The barrier_conn fixture is now manual-commit (like create_pg_connection in production) instead of autocommit, with explicit commits in the seed/read/cleanup helpers. So Layer 4 now exercises sweep_expired_barriers' own commit() (a missing commit would now fail the persistence assertion) and the rollback path in the same mode the real reaper runs in.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -0,0 +1,300 @@
...
+                    self._lease.release()
+                except Exception:
+                    logger.warning(
+                        "Reaper: failed to release lease on shutdown; a standby "
+                        "will take over after the lease window (~%ss) instead of "
+                        "immediately",
+                        self._lease.lease_seconds,
+                        exc_info=True,
+                    )
+            logger.info("Reaper stopped")
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed (7ea670313). run()'s finally now closes the OWNED sweep connection (injected ones are left to the caller) after releasing the lease. Harmless for the main() process but keeps PgReaper clean if embedded/test-driven.

muhammad-ali-e:
## Greptile review addressed — commit `7ea670313`

Both findings validated and fixed:

- **Layer-4 fixture was autocommit** — which made the sweep's own `commit()` a no-op and its rollback unreachable, so the real-Postgres tests ran in a different mode than the production reaper (`create_pg_connection` is manual-commit). Switched the fixture to manual-commit + explicit commits in the seed/read/cleanup helpers, so Layer 4 now verifies the sweep actually persists its DELETE and exercises the rollback path in production mode.
- **Owned sweep connection not closed on shutdown** — `run()` now closes its owned sweep connection in the `finally` (an injected one stays the caller's).

23/23 pass; ruff + pre-commit green. Greptile confidence was 4/5 with these two as the only items — both resolved.

PR #2059, opened 2026-06-16
UN-3555 [FEAT] PG Queue 9d slice-2 followup — run-worker.sh reaper type + pg-queue set

muhammad-ali-e:
## What

Followup to the reaper slice (UN-3554): wires the reaper into `run-worker.sh` and adds a **`pg-queue` set** so the whole PG-queue group launches in one shot. Mirrors the 9c → 9c-followup split (launcher wiring as its own slice). The reaper's liveness probe is a separate follow-on.

## What's in it

**`workers/pg_queue_reaper/`** — a thin entrypoint package (`python -m pg_queue_reaper` → `queue_backend.pg_queue.reaper.main`). Unlike `pg_queue_consumer` it needs **no** worker-app bootstrap (the reaper runs no Celery tasks, only SQL recovery); it exists so the process has a stable name `run-worker.sh` can launch and pgrep-match.

**`run-worker.sh`**
- **`reaper` / `pg-queue-reaper`** worker type — opt-in (**not** in `all`), launches `python -m pg_queue_reaper`, runs from the workers root, `--status`/`-k`/`-r` matching via the `-m` pgrep branch (now covers consumer + reaper). Lease/interval env documented in `--help`.
- **`pg` / `pg-queue` set** — `run_pg_queue_set` launches consumer + reaper together (always detached, like `all`). `-r pg-queue` kills both members then relaunches. `list_core_worker_dirs` skips the set alias so status shows no phantom `pg-queue` worker.
- **Help** documents the Celery `all` set and the PG `pg-queue` set as independent — run both in parallel for a dual-transport (strangler-fig) setup.

## Why a set (and what stays env/K8s)

`run-worker.sh` is the **dev / host-launch** path; the `pg-queue` set is its ergonomics win — bring up the whole PG group in one command, alongside `all`, for dual-transport dev. **Production** still runs each component as its own K8s Deployment, with `WORKER_PG_QUEUE_ENABLED_TASKS` / `WORKER_BARRIER_BACKEND` controlling what's active — so no dynamic "auto-detect available queues" logic is baked into the shell.

## Dev-test (live, against the running stack)

- `run-worker.sh reaper -d` → launched, **acquired leadership**, ticking; `--status` shows `pg_queue_reaper: RUNNING (2 proc)`.
- `run-worker.sh pg-queue` → consumer + reaper both up + RUNNING, reaper leader, **no phantom set entry** in status.
- `bash -n` clean; `--help` renders the new entries; entrypoint imports cleanly.

## Out of scope (follow-ons)

- Reaper **liveness probe** (LivenessServer exposing `is_leader` + last-tick freshness) — separate slice, mirroring how consumer liveness (UN-3544) followed the consumer wiring.
- Production K8s manifests.

Base: `feat/UN-3445-pg-queue-integration`. Sub-task UN-3555 under UN-3536, followup to UN-3554.

### DIFF
--- a/workers/pg_queue_reaper/__main__.py
+++ b/workers/pg_queue_reaper/__main__.py
@@ -1,0 +1,15 @@
+"""Entry point: run the PG-queue reaper (leader-elected recovery loop).
+
+Unlike the consumer, the reaper runs no Celery tasks — it only does SQL recovery
+(the barrier-orphan sweep) — so it needs **no** worker-app bootstrap (no
+``import worker``, no ``WORKER_TYPE`` juggling). This package exists purely so the
+process has a stable ``python -m pg_queue_reaper`` name that ``run-worker.sh`` can
+launch and match, parallel to ``pg_queue_consumer``.
+
+Launch via ``python -m pg_queue_reaper`` or ``./run-worker.sh reaper``.
+"""
+
+from queue_backend.pg_queue.reaper import main
+
+if __name__ == "__main__":
+    main()
--- a/workers/pg_queue_reaper/__main__.py
+++ b/workers/pg_queue_reaper/__main__.py
@@ -1,10 +1,13 @@
 """Entry point: run the PG-queue reaper (leader-elected recovery loop).
 
 Unlike the consumer, the reaper runs no Celery tasks — it only does SQL recovery
-(the barrier-orphan sweep) — so it needs **no** worker-app bootstrap (no
-``import worker``, no ``WORKER_TYPE`` juggling). This package exists purely so the
-process has a stable ``python -m pg_queue_reaper`` name that ``run-worker.sh`` can
-launch and match, parallel to ``pg_queue_consumer``.
+(the barrier-orphan sweep) — so it needs **no** worker-app bootstrap: it never
+``import``s ``worker``. (``run-worker.sh`` still exports ``WORKER_TYPE`` for every
+worker, reaper included, but — unlike the consumer, which overwrites it before
+``import worker`` to select which tasks to register — the reaper neither reads
+nor mutates it.) This package exists purely so the process has a stable
+``python -m pg_queue_reaper`` name that ``run-worker.sh`` can launch and match,
+parallel to ``pg_queue_consumer``.
 
 Launch via ``python -m pg_queue_reaper`` or ``./run-worker.sh reaper``.
 """
--- a/workers/run-worker.sh
+++ b/workers/run-worker.sh
@@ -748,7 +748,7 @@ run_worker() {
     print_status $GREEN "Starting $worker_type worker..."
     print_status $BLUE "Directory: $worker_dir"
     print_status $BLUE "Worker Name: $worker_instance_name"
-    print_status $BLUE "Queues: $queues"
+    print_status $BLUE "Queues: ${queues:-n/a}"
     # Show the effective port: a -p/--health-port override wins over the map.
     print_status $BLUE "Health Port: ${health_port:-${WORKER_HEALTH_PORTS[$worker_type]:-n/a}}"
     print_status $BLUE "Command: ${cmd_args[*]}"
@@ -855,23 +855,36 @@ run_all_workers() {
 }
 
 # Function to run the PG-queue worker set (consumer + reaper) together.
-# A set is multiple processes, so — like 'all' — it always runs detached. This
-# is the PG-queue counterpart to 'all' (the Celery set): run both for a
-# dual-transport (strangler-fig) setup. The reaper is a leader-elected singleton,
-# so running it on several hosts is safe (only one wins the lease).
+# A set is multiple processes, so — like 'all' — it ALWAYS runs detached
+# (it ignores -d/--detach; there's no foreground form). The PG-queue counterpart
+# to 'all' (the Celery set): run both for a dual-transport (strangler-fig) setup.
+# The reaper is a leader-elected singleton, so running it on several hosts is
+# safe (only one wins the lease). Returns non-zero if any member dies on start —
+# the reaper has no health port yet, so this launch check is the only
+# programmatic startup signal a caller (systemd/CI) gets.
 run_pg_queue_set() {
     local log_level=$1
     local concurrency=$2
     local pool_type=$3
 
     print_status $GREEN "Starting PG-queue set (consumer + reaper)..."
+    local failed=0
     for worker in "$PG_QUEUE_CONSUMER_TYPE" "$PG_QUEUE_REAPER_TYPE"; do
         print_status $BLUE "Starting $worker in background..."
-        (
-            run_worker "$worker" "true" "$log_level" "$concurrency" "" "" "$pool_type" ""
-        ) &
-        sleep 2
+        # A FOREGROUND subshell: run_worker (detach=true) nohup-backgrounds the
+        # actual worker and returns 1 on an immediate crash-on-start — so the
+        # subshell's exit status IS that signal. The subshell isolates run_worker's
+        # `cd` from this loop; the nohup'd worker survives the subshell exiting.
+        # (A background `( … ) &` would lose the status — its `$!` is the launcher
+        # subshell, which exits the instant it backgrounds the worker.)
+        ( run_worker "$worker" "true" "$log_level" "$concurrency" "" "" "$pool_type" "" ) \
+            || failed=1
     done
+    if [[ $failed -ne 0 ]]; then
+        print_status $RED "PG-queue set: a member failed to start (see logs above)"
+        show_status
+        return 1
+    fi
     print_status $GREEN "PG-queue set started in background"
     show_status
 }
@@ -993,9 +1006,17 @@ if [[ "$RESTART_MODE" == "true" ]]; then
     if [[ "${WORKERS[$WORKER_TYPE]:-}" == "$PG_QUEUE_SET" ]]; then
         # Restart the PG-queue set: kill both members, then fall through to the
         # launch path (which runs run_pg_queue_set since WORKER_TYPE is the set).
+        # Aggregate kill failures (kill_one_worker returns 1 if a process survives
+        # SIGKILL) and abort rather than relaunch over a survivor — a second
+        # consumer would double-poll Postgres. Mirrors kill_workers' discipline.
         print_status $BLUE "Restarting PG-queue set..."
-        kill_one_worker "$PG_QUEUE_CONSUMER_TYPE"
-        kill_one_worker "$PG_QUEUE_REAPER_TYPE"
+        restart_failed=0
+        kill_one_worker "$PG_QUEUE_CONSUMER_TYPE" || restart_failed=1
+        kill_one_worker "$PG_QUEUE_REAPER_TYPE" || restart_failed=1
+        if [[ $restart_failed -ne 0 ]]; then
+            print_status $RED "Cannot restart PG-queue set: a member survived SIGKILL; aborting to avoid duplicate processes"
+            exit 1
+        fi
     elif [[ -n "$WORKER_TYPE" && "$WORKER_TYPE" != "all" ]]; then
         restart_target_dir="${WORKERS[$WORKER_TYPE]:-${PLUGGABLE_WORKERS[$WORKER_TYPE]:-}}"
         if [[ -z "$restart_target_dir" ]]; then
@@ -1052,7 +1073,9 @@ if [[ "$WORKER_TYPE" == "all" ]]; then
     run_all_workers "$DETACH" "$LOG_LEVEL" "$CONCURRENCY" "$POOL_TYPE"
 elif [[ "${WORKERS[$WORKER_TYPE]:-}" == "$PG_QUEUE_SET" ]]; then
     # The PG-queue set (consumer + reaper). Always backgrounded (multiple procs).
-    run_pg_queue_set "$LOG_LEVEL" "$CONCURRENCY" "$POOL_TYPE"
+    # Propagate a member start-failure to the script exit code — the reaper has
+    # no health port, so this is the only programmatic startup signal.
+    run_pg_queue_set "$LOG_LEVEL" "$CONCURRENCY" "$POOL_TYPE" || exit 1
 else
     # Resolve worker directory name from either WORKERS or PLUGGABLE_WORKERS
     WORKER_DIR_NAME="${WORKERS[$WORKER_TYPE]}"
--- a/workers/run-worker.sh
+++ b/workers/run-worker.sh
@@ -27,6 +27,10 @@ readonly IDE_CALLBACK_WORKER_TYPE="ide_callback"
 # Canonical name of the PG-queue consumer worker (referenced in several maps
 # and special-cases below; a constant keeps them in sync).
 readonly PG_QUEUE_CONSUMER_TYPE="pg_queue_consumer"
+# Canonical name of the PG-queue reaper (leader-elected recovery process).
+readonly PG_QUEUE_REAPER_TYPE="pg_queue_reaper"
+# Set alias that launches the whole PG-queue group (consumer + reaper) together.
+readonly PG_QUEUE_SET="pg-queue"
 
 # Available workers
 declare -A WORKERS=(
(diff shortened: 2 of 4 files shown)


workers/run-worker.sh
```diff
@@ -826,6 +854,28 @@ run_all_workers() {
...
+# is the PG-queue counterpart to 'all' (the Celery set): run both for a
+# dual-transport (strangler-fig) setup. The reaper is a leader-elected singleton,
+# so running it on several hosts is safe (only one wins the lease).
+run_pg_queue_set() {
+    local log_level=$1
+    local concurrency=$2
+    local pool_type=$3
+
+    print_status $GREEN "Starting PG-queue set (consumer + reaper)..."
+    for worker in "$PG_QUEUE_CONSUMER_TYPE" "$PG_QUEUE_REAPER_TYPE"; do
```
muhammad-ali-e, line 881: **[High] `run_pg_queue_set` swallows member start-failures and always returns 0.**

`run_worker` was deliberately built to fast-fail: on an immediate crash (import error, bad env cast) it prints a RED failure + log tail and `return 1` (lines 789-793). Here that signal is destroyed twice:
1. Each member runs in a backgrounded subshell `( run_worker ... ) &` whose exit status is never collected — `run_worker`'s `return 1` evaporates.
2. The function then unconditionally prints the GREEN "PG-queue set started" banner and returns 0, so `./run-worker.sh pg-queue` reports success even if the reaper died on import.

Secondarily, unlike `run_all_workers` this path also **ignores `-d/--detach`** — it always backgrounds and never `wait`s, so `pg-queue` and `-d pg-queue` behave identically (the function comment says this is intentional; if so, note it in `usage()`).

**Fix:** capture each member PID, and after the loop `kill -0`-check each one, returning non-zero (and showing a RED line) if any member is already dead:
```sh
local pids=() failed=0
for worker in "$PG_QUEUE_CONSUMER_TYPE" "$PG_QUEUE_REAPER_TYPE"; do
    ( run_worker "$worker" "true" "$log_level" "$concurrency" "" "" "$pool_type" "" ) &
    pids+=("$!"); sleep 2
done
for pid in "${pids[@]}"; do kill -0 "$pid" 2>/dev/null || failed=1; done
[[ $failed -eq 0 ]] || { print_status $RED "PG-queue set: a member failed to start"; show_status; return 1; }
```

workers/run-worker.sh
```diff
@@ -994,6 +1050,9 @@ export PYTHONPATH="$WORKERS_DIR:${PYTHONPATH:-}"
 # Run the requested worker(s)
 if [[ "$WORKER_TYPE" == "all" ]]; then
     run_all_workers "$DETACH" "$LOG_LEVEL" "$CONCURRENCY" "$POOL_TYPE"
+elif [[ "${WORKERS[$WORKER_TYPE]:-}" == "$PG_QUEUE_SET" ]]; then
+    # The PG-queue set (consumer + reaper). Always backgrounded (multiple procs).
+    run_pg_queue_set "$LOG_LEVEL" "$CONCURRENCY" "$POOL_TYPE"
```
muhammad-ali-e: **[Medium] Set-launch exit code is never propagated to the script's exit status.**

`run_pg_queue_set` is invoked as a bare statement, then the script falls off the end and exits 0. Even once the function (above) is fixed to `return 1` on a dead member, that won't reach `$?` of `./run-worker.sh`. Any orchestrator/systemd/CI wrapper that runs `./run-worker.sh pg-queue` and checks the exit code will treat a half-dead set as healthy — and since the reaper has no health port, the launch exit code is the only programmatic startup signal.

**Fix:** `run_pg_queue_set "$LOG_LEVEL" "$CONCURRENCY" "$POOL_TYPE" || exit 1`

workers/run-worker.sh
```diff
@@ -717,6 +737,14 @@ run_worker() {
         cmd_args=("uv" "run" "python" "-m" "$PG_QUEUE_CONSUMER_TYPE")
     fi
 
+    # PG queue reaper — a leader-elected SQL recovery loop (no Celery, no task
+    # bootstrap). Override the celery command with the plain `python -m` entry.
+    # Tunables (lease window, cycle interval) come from env; no liveness port is
+    # wired yet (that's a follow-on slice), so it binds nothing.
+    if [[ "$worker_type" == "$PG_QUEUE_REAPER_TYPE" ]]; then
```
muhammad-ali-e, line 753: **[Medium] Reaper has no health port → a dead reaper is invisible in `--status`.**

Three properties combine into a true silent death of a background recovery process:
1. The reaper binds no health port (no `WORKER_HEALTH_PORTS` entry), so `--status` detection is purely `pgrep` of `-m pg_queue_reaper`.
2. It's in `OPTIN_WORKERS` (line 113), so `show_status` **skips it entirely when no PID is found** — a crashed reaper produces zero PIDs and is silently omitted, never shown as STOPPED.
3. It's a leader-elected singleton, so on the leader host a dead reaper means the barrier-orphan sweep just stops, with no operator-visible symptom until barriers pile up.

The "no liveness port is wired yet (follow-on slice)" comment acknowledges the gap, but until that lands an operator can't distinguish "reaper intentionally not started" from "reaper crashed an hour ago." At minimum, make the launch-time liveness check (see the `run_pg_queue_set` comment) mandatory so *startup* death is loud; ideally reserve a liveness port now, mirroring the consumer's 8090.

_Minor, same block:_ `WORKER_QUEUES` has no `pg_queue_reaper` key, so the startup banner prints an empty `Queues:` line (line 751). Harmless (cmd_args is fully overridden), but consider printing `Queues: n/a (reaper)` like the Health Port line already does.

workers/run-worker.sh
```diff
@@ -940,7 +990,13 @@ fi
 #   WORKER_TYPE as "all" once we set it below.
 if [[ "$RESTART_MODE" == "true" ]]; then
     discover_pluggable_workers
-    if [[ -n "$WORKER_TYPE" && "$WORKER_TYPE" != "all" ]]; then
+    if [[ "${WORKERS[$WORKER_TYPE]:-}" == "$PG_QUEUE_SET" ]]; then
+        # Restart the PG-queue set: kill both members, then fall through to the
+        # launch path (which runs run_pg_queue_set since WORKER_TYPE is the set).
+        print_status $BLUE "Restarting PG-queue set..."
+        kill_one_worker "$PG_QUEUE_CONSUMER_TYPE"
+        kill_one_worker "$PG_QUEUE_REAPER_TYPE"
```
muhammad-ali-e: **[Medium] Set-restart ignores `kill_one_worker` failures, then relaunches over survivors.**

`kill_one_worker` returns 1 when a process survives SIGKILL (line 507). Both calls here discard that return value, then control falls through to relaunch the set. If an old consumer/reaper refuses to die, the script silently starts a *second* instance on top of the survivor — for the consumer that means double-polling Postgres. Note `kill_workers` (lines 521-529) deliberately aggregates `|| any_failed=1` and surfaces a RED error; this new branch drops that discipline.

**Fix:** aggregate and abort the relaunch if a member won't die:
```sh
restart_failed=0
kill_one_worker "$PG_QUEUE_CONSUMER_TYPE" || restart_failed=1
kill_one_worker "$PG_QUEUE_REAPER_TYPE"   || restart_failed=1
[[ $restart_failed -eq 0 ]] || { print_status $RED "Cannot restart PG-queue set: a member survived SIGKILL; aborting to avoid duplicate processes"; exit 1; }
```

workers/pg_queue_reaper/__main__.py
```diff
@@ -0,0 +1,15 @@
+"""Entry point: run the PG-queue reaper (leader-elected recovery loop).
+
+Unlike the consumer, the reaper runs no Celery tasks — it only does SQL recovery
+(the barrier-orphan sweep) — so it needs **no** worker-app bootstrap (no
```
muhammad-ali-e: **[Low] "no `WORKER_TYPE` juggling" is misleading.**

The "no `import worker`" half is correct — `reaper.main()` never imports `worker`. But "no `WORKER_TYPE` juggling" reads as "the reaper process has no `WORKER_TYPE` in its env," which is false: `run-worker.sh` `run_worker()` unconditionally `export WORKER_TYPE=...` for *every* worker type, reaper included (line 603). A maintainer debugging the reaper will see `WORKER_TYPE=pg_queue_reaper` and think this docstring is wrong.

The accurate, narrower claim: the launcher still exports `WORKER_TYPE`, but unlike the consumer (which overwrites it before `import worker`) the reaper neither reads nor mutates it. Suggest rewording to say exactly that, and replacing the vague "juggling" with the concrete mechanic.

workers/pg_queue_reaper/__main__.py
```diff
@@ -0,0 +1,15 @@
...
+``import worker``, no ``WORKER_TYPE`` juggling). This package exists purely so the
+process has a stable ``python -m pg_queue_reaper`` name that ``run-worker.sh`` can
+launch and match, parallel to ``pg_queue_consumer``.
+
+Launch via ``python -m pg_queue_reaper`` or ``./run-worker.sh reaper``.
+"""
+
+from queue_backend.pg_queue.reaper import main
```
muhammad-ali-e, line 15: **[Low] No smoke test for this launch path; consider deferring the import for symmetry with the consumer.**

This line is the single point where `run-worker.sh reaper` either works or dies instantly with `ImportError` (the `queue_backend` package eagerly imports celery/barrier/etc.), yet nothing tests it — `test_pg_reaper.py` covers the loop in `reaper.py` but not this entry point. A ~5-line smoke test catches the most likely real regression:
```python
def test_main_module_reexports_real_main():
    import importlib
    from queue_backend.pg_queue.reaper import main as real_main
    m = importlib.import_module("pg_queue_reaper.__main__")
    assert m.main is real_main
```
(The consumer's `__main__` lacks this too — worth covering both.)

Separately: the consumer's `__main__` deliberately defers its heavy import into a guarded function to avoid side-effects on import. For symmetry with that documented convention you could move this import inside the `if __name__ == "__main__":` block. Optional — `__main__.py` is essentially never imported as a library.

muhammad-ali-e (commented):

Overall this is a clean, well-commented wiring slice that faithfully mirrors the existing `pg_queue_consumer` pattern. No bugs that break the happy path. The themes worth addressing before merge are **observability of failures** in the new `pg-queue` set launcher/restart path, and one **misleading docstring**. The code-simplifier found nothing actionable (the repeated blocks are intentional and clearer left explicit). Inline findings below, prioritised.

workers/run-worker.sh
muhammad-ali-e: Fixed (2092a5cf6) — and the suggested kill -0 on $! would actually false-fail: that PID is the launcher subshell, which exits the instant it backgrounds the nohup-d worker, so kill -0 always reports it dead (I hit exactly this in dev-test: exit=1 on a successful launch). Instead each member now runs in a FOREGROUND subshell, so run_worker-s own return 1 on a crash-on-start is captured directly while its cd is isolated; the set returns non-zero if any member fails. Also documented that the set always runs detached (ignores -d). Re-tested: exit=0 on a healthy launch.

workers/run-worker.sh
muhammad-ali-e: Fixed (2092a5cf6). Dispatch now does `run_pg_queue_set ... || exit 1`, so a member start-failure reaches the script exit code (the only programmatic startup signal until the reaper has a health port).

workers/run-worker.sh
muhammad-ali-e: Addressed the startup half (2092a5cf6): with the foreground-subshell fix + `|| exit 1`, a member dying on start is now loud and non-zero. The Queues banner prints `n/a` when empty (reaper). The deeper point — a reaper that crashes *after* startup is silently omitted from --status (opt-in skip + no health port) — is exactly what the reaper **liveness probe** slice covers (is_leader + last-tick freshness on a health port, mirroring the consumer-s 8090); that-s the next follow-on. I did not reserve a non-functional port now since it would show a misleading bound port without a server behind it.

workers/run-worker.sh
muhammad-ali-e: Fixed (2092a5cf6). Set-restart now aggregates kill_one_worker failures and aborts the relaunch with a RED error if a member survives SIGKILL — avoids starting a second consumer that double-polls Postgres. Mirrors kill_workers- discipline.

workers/pg_queue_reaper/__main__.py
muhammad-ali-e: Reworded (2092a5cf6). The docstring now states the accurate mechanic: run-worker.sh DOES export WORKER_TYPE for every worker (reaper included), but — unlike the consumer, which overwrites it before `import worker` to pick which tasks to register — the reaper neither reads nor mutates it. Dropped the vague "juggling".

workers/pg_queue_reaper/__main__.py
muhammad-ali-e: Added the smoke test (2092a5cf6): test_main_module_reexports_real_main asserts pg_queue_reaper.__main__.main is the real reaper main — guards the `python -m pg_queue_reaper` path against an ImportError regression. Kept the import at module level (that-s what makes the re-export test possible; __main__ is never imported as a library in practice). The consumer-s __main__ defers its import inside the bootstrap guard, so the same re-export shape does not apply there.

muhammad-ali-e:
## Review addressed — commit `2092a5cf6`

All 6 toolkit findings in (SonarCloud already green). Validated each first; all valid.

**High / Medium (start-failure visibility)**
- **`run_pg_queue_set` swallowed member failures.** Fixed — each member now runs in a **foreground** subshell so `run_worker`'s `return 1` on a crash-on-start is captured, and the set returns non-zero if any member fails. Worth noting: the suggested `kill -0 $!` would *false-fail* (that PID is the launcher subshell, which exits the moment it backgrounds the nohup'd worker — I hit exactly that: `exit=1` on a healthy launch). The foreground-subshell return value is the correct signal and isolates `cd`.
- **Exit code now propagates** — dispatch does `run_pg_queue_set … || exit 1`, so systemd/CI see a half-dead set as failed (the only programmatic startup signal until the reaper has a health port).
- **Set-restart** aggregates `kill_one_worker` failures and aborts the relaunch if a member survives SIGKILL — no duplicate consumer double-polling Postgres.

**Low / minor**
- `Queues: n/a` when empty (reaper); reworded the `__main__` docstring (the launcher *does* export `WORKER_TYPE`; the reaper just never reads/mutates it); added a smoke test pinning that `pg_queue_reaper.__main__` re-exports the real reaper main.

**Deferred (acknowledged):** a reaper that crashes *after* startup is still omitted from `--status` (opt-in skip + no health port). That's precisely the **reaper liveness probe** slice (next follow-on) — `is_leader` + last-tick freshness on a health port, mirroring the consumer's 8090.

Dev-tested: `run-worker.sh pg-queue` → `exit=0` with both members up; 24 reaper tests pass; `bash -n` + ruff clean.

workers/run-worker.sh
```diff
@@ -826,6 +854,41 @@ run_all_workers() {
...
+        # (A background `( … ) &` would lose the status — its `$!` is the launcher
+        # subshell, which exits the instant it backgrounds the worker.)
+        ( run_worker "$worker" "true" "$log_level" "$concurrency" "" "" "$pool_type" "" ) \
+            || failed=1
+    done
+    if [[ $failed -ne 0 ]]; then
+        print_status $RED "PG-queue set: a member failed to start (see logs above)"
+        show_status
+        return 1
+    fi
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed (0b381a3a4). On a partial start-failure run_pg_queue_set now kills BOTH members before returning 1 (the crashed one is already gone -> no-op), so no survivor is left for a restart-on-failure relaunch to double up on. All-or-nothing, matching the restart path discipline you referenced.

muhammad-ali-e:
## Greptile review addressed — commit `0b381a3a4`

Both findings valid and fixed:

- **Partial startup leaves a survivor (run_pg_queue_set).** On a partial start-failure (one member up, the other crashed), the set now tears **both** members down before `return 1` — so a `Restart=on-failure`/CI relaunch can't spawn a second instance over the survivor (the consumer would double-poll Postgres). All-or-nothing, mirroring the restart path.
- **`--logs` misfired on the `pg`/`pg-queue` set alias** (out-of-diff comment). `tail_logs` now special-cases the set alias and tails **both** member logs (`pg_queue_consumer` + `pg_queue_reaper`) instead of looking for a non-existent `workers/pg-queue/pg-queue.log`. Mirrors how `list_core_worker_dirs` skips the set value. Dev-tested: `--logs pg-queue` → "Tailing 2 log file(s)".

`bash -n` clean; 24 reaper tests pass. Confidence was 4/5 with these two — both resolved.

PR #2061, opened 2026-06-16
UN-3556 [FEAT] PG Queue 9d — reaper liveness probe (heartbeat + is_leader)

muhammad-ali-e:
## What

The reaper **liveness probe** — closing the gap flagged in #2059's review: a reaper that crashes *after* startup was invisible (opt-in skip in `--status` + no health port). This finishes 9d. Mirrors the consumer's liveness (UN-3544).

## What's in it

- **`PgReaper` heartbeat** — `_last_tick_monotonic` stamped at the **start** of every tick (a standby tick counts as progress — liveness tracks the *loop*, not leadership) + `seconds_since_last_tick()` / `is_tick_stale()`.
- **`ReaperLivenessServer`** — lean HTTP probe (mirrors the consumer's `LivenessServer`): `/health` (also `/healthz`, `/livez`) → **200** while the tick loop is fresh, **503** when stale. Payload also surfaces **`is_leader`** (which pod holds the lease — handy for 9e debugging). The 200/503 verdict is **purely the heartbeat** — never leadership (a standby is healthy) or DB reachability (a blip must not crash-loop a healthy process).
- **`main()` wiring** — `WORKER_PG_REAPER_HEALTH_PORT` (unset → no server, no stray port); staleness window `WORKER_PG_REAPER_HEALTH_STALE_SECONDS` (default 30s, comfortably above the 5s tick interval). A bind failure degrades gracefully (logs, runs probe-less).
- **`run-worker.sh`** — reserves port **8086** for the reaper, exports the health-port env in the reaper special-case, documents the two new env vars in `--help`.

## Tests (+13 → 37 total)

Heartbeat fresh/stale + tick-refresh; liveness server 200-fresh / 503-stale / `is_leader` reflected / 404 / double-start-raises; health-staleness env default+override+invalid; server-disabled-when-no-port.

## Dev-test (live)

`./run-worker.sh reaper -d` → `GET :8086/health` →
`{"status":"healthy","check":"pg_reaper_tick","seconds_since_last_tick":4.79,"stale_after_seconds":30.0,"is_leader":true}`

## Design note

Kept in `reaper.py` (not a shared module) — consistent with the consumer's `LivenessServer` living in `consumer.py`. A future cleanup could extract a shared generic server; out of scope here to avoid refactoring the merged-and-tested consumer probe.

Base: `feat/UN-3445-pg-queue-integration`. **This completes 9d** (orchestrator/reaper). Sub-task UN-3556 under UN-3536.

### DIFF
--- a/workers/queue_backend/pg_queue/liveness.py
+++ b/workers/queue_backend/pg_queue/liveness.py
@@ -1,0 +1,152 @@
+"""Shared tiny HTTP liveness probe for PG-queue processes.
+
+Both the consumer (poll loop) and the reaper (tick loop) need the same probe:
+"is this loop still making progress?". This is the single implementation,
+parameterised by a freshness callable + the payload's ``check``/age labels (and
+an optional extra-status callable — e.g. the reaper's ``is_leader``). Each side
+wraps it in a thin subclass with its own constructor shape.
+
+Deliberately lean. A *liveness* probe must answer one question — progress — and
+nothing else. It must NOT depend on broker/DB reachability or resource pressure:
+a transient backend blip or a busy moment would otherwise make the orchestrator
+crash-loop an otherwise-healthy process. So this does not reuse the shared
+``HealthChecker`` (which bundles api-connectivity / system-resource checks meant
+for richer health reporting, not liveness).
+
+Serves ``/health`` (also ``/healthz``, ``/livez``) on ``0.0.0.0`` (a
+container/k8s probe reaches it from outside the process) in a daemon thread.
+Bind ``port=0`` to let the OS pick a free port (read back via :attr:`bound_port`)
+— used in tests. Start once; :meth:`stop` returns it to the inert state.
+"""
+
+from __future__ import annotations
+
+import logging
+from collections.abc import Callable
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+
+class LivenessServer:
+    """Generic liveness probe: 200 while ``freshness_fn()`` is within
+    ``stale_after`` seconds, else 503.
+
+    ``check_name`` / ``age_key`` label the JSON payload; ``extra_status_fn``
+    (optional) merges extra fields in (informational — it never affects the
+    200/503 verdict, which is purely the freshness heartbeat).
+    """
+
+    _PATHS = frozenset({"/health", "/healthz", "/livez"})
+
+    def __init__(
+        self,
+        *,
+        freshness_fn: Callable[[], float],
+        stale_after: float,
+        port: int,
+        check_name: str,
+        age_key: str,
+        extra_status_fn: Callable[[], dict[str, Any]] | None = None,
+        thread_name: str = "pg-queue-liveness",
+    ) -> None:
+        self._freshness_fn = freshness_fn
+        self._stale_after = stale_after
+        self._port = port
+        self._check_name = check_name
+        self._age_key = age_key
+        self._extra_status_fn = extra_status_fn
+        self._thread_name = thread_name
+        self._httpd: Any = None
+        self._thread: Any = None
+
+    def start(self) -> None:
+        import json
+        import threading
+        from http.server import BaseHTTPRequestHandler, HTTPServer
+        from urllib.parse import urlsplit
+
+        if self._httpd is not None:
+            raise RuntimeError("LivenessServer already started")
+
+        freshness_fn = self._freshness_fn
+        stale_after = self._stale_after
+        paths = self._PATHS
+        check_name = self._check_name
+        age_key = self._age_key
+        extra_status_fn = self._extra_status_fn
+
+        class _Handler(BaseHTTPRequestHandler):
+            def do_GET(self) -> None:
+                # Strip any query string — a probe like /health?foo=bar must match.
+                if urlsplit(self.path).path not in paths:
+                    self.send_response(404)
+                    self.end_headers()
+                    return
+                # One clock read so age and the healthy/stale verdict share an
+                # instant. The verdict is purely freshness — extra_status_fn
+                # fields are informational and never flip it.
+                age = freshness_fn()
+                stale = age > stale_after
+                payload: dict[str, Any] = {
+                    "status": "unhealthy" if stale else "healthy",
+                    "check": check_name,
+                    age_key: round(age, 3),
+                    "stale_after_seconds": stale_after,
+                }
+                if extra_status_fn is not None:
+                    payload.update(extra_status_fn())
+                body = json.dumps(payload).encode()
+                self.send_response(503 if stale else 200)
+                self.send_header("Content-Type", "application/json")
+                self.end_headers()
+                try:
+                    self.wfile.write(body)
+                except (BrokenPipeError, ConnectionResetError):
+                    pass  # client (probe) hung up mid-response — not our problem
+
+            def log_message(self, *_: object) -> None:
+                pass  # silence per-request access logging
+
+            def log_error(self, fmt: str, *args: object) -> None:
+                # BaseHTTPRequestHandler routes errors through log_message too;
+                # don't let the pass above swallow them — surface to our logger.
+                logger.warning("pg-queue liveness handler: " + fmt, *args)
+
+        def _serve(httpd: Any) -> None:
+            try:
+                httpd.serve_forever()
+            except Exception:
+                # A daemon thread dying silently would make /health stop
+                # answering (connection refused) with no breadcrumb.
+                logger.exception("pg-queue liveness server thread crashed")
+
+        httpd = HTTPServer(("0.0.0.0", self._port), _Handler)
+        self._httpd = httpd
+        self._thread = threading.Thread(
+            target=_serve, args=(httpd,), daemon=True, name=self._thread_name
+        )
+        self._thread.start()
+
+    @property
+    def bound_port(self) -> int:
+        """Actual listening port (resolves ``port=0``); the requested port if not started."""
+        if self._httpd is not None:
+            return self._httpd.server_address[1]
+        return self._port
+
+    def stop(self) -> None:
+        """Shut the server down. Defensive: never raises (called from a finally)."""
+        try:
+            if self._httpd is not None:
+                self._httpd.shutdown()
+                self._httpd.server_close()
+            if self._thread is not None:
+                self._thread.join(timeout=5)
+                if self._thread.is_alive():
+                    logger.warning("pg-queue liveness thread did not stop within 5s")
+        except Exception:
+            logger.exception("pg-queue: error stopping liveness server")
+        finally:
+            self._httpd = None
+            self._thread = None
--- a/workers/queue_backend/pg_queue/liveness.py
+++ b/workers/queue_backend/pg_queue/liveness.py
@@ -23,7 +23,11 @@ from __future__ import annotations
 
 import logging
 from collections.abc import Callable
-from typing import Any
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from http.server import HTTPServer
+    from threading import Thread
 
 logger = logging.getLogger(__name__)
 
@@ -50,6 +54,11 @@ class LivenessServer:
         extra_status_fn: Callable[[], dict[str, Any]] | None = None,
         thread_name: str = "pg-queue-liveness",
     ) -> None:
+        # Re-validate here (not only at the env boundary): a direct caller could
+        # otherwise build an always-503 probe that crash-loops the pod. Mirrors
+        # the codebase's load-bearing re-validation convention (PgReaper.__init__).
+        if stale_after <= 0:
+            raise ValueError(f"stale_after must be positive, got {stale_after!r}")
         self._freshness_fn = freshness_fn
         self._stale_after = stale_after
         self._port = port
@@ -57,8 +66,8 @@ class LivenessServer:
         self._age_key = age_key
         self._extra_status_fn = extra_status_fn
         self._thread_name = thread_name
-        self._httpd: Any = None
-        self._thread: Any = None
+        self._httpd: HTTPServer | None = None
+        self._thread: Thread | None = None
 
     def start(self) -> None:
         import json
@@ -113,7 +122,7 @@ class LivenessServer:
                 # don't let the pass above swallow them — surface to our logger.
                 logger.warning("pg-queue liveness handler: " + fmt, *args)
 
-        def _serve(httpd: Any) -> None:
+        def _serve(httpd: HTTPServer) -> None:
             try:
                 httpd.serve_forever()
             except Exception:
@@ -130,7 +139,10 @@ class LivenessServer:
 
     @property
     def bound_port(self) -> int:
-        """Actual listening port (resolves ``port=0``); the requested port if not started."""
+        """Actual listening port once started (resolves ``port=0`` to the
+        OS-chosen port). Before :meth:`start` / after :meth:`stop`, returns the
+        configured port value (which is ``0`` when the OS is asked to choose).
+        """
         if self._httpd is not None:
             return self._httpd.server_address[1]
         return self._port
--- a/workers/queue_backend/pg_queue/liveness.py
+++ b/workers/queue_backend/pg_queue/liveness.py
@@ -53,6 +53,7 @@ class LivenessServer:
         age_key: str,
         extra_status_fn: Callable[[], dict[str, Any]] | None = None,
         thread_name: str = "pg-queue-liveness",
+        log_label: str = "pg-queue",
     ) -> None:
         # Re-validate here (not only at the env boundary): a direct caller could
         # otherwise build an always-503 probe that crash-loops the pod. Mirrors
@@ -66,6 +67,10 @@ class LivenessServer:
         self._age_key = age_key
         self._extra_status_fn = extra_status_fn
         self._thread_name = thread_name
+        # Prefixes the (now-shared) log messages so they stay attributable to the
+        # source process after the consumer/reaper extraction (e.g. "pg-queue
+        # consumer" / "pg-queue reaper") — they all log via this module's logger.
+        self._log_label = log_label
         self._httpd: HTTPServer | None = None
         self._thread: Thread | None = None
 
@@ -84,6 +89,7 @@ class LivenessServer:
         check_name = self._check_name
         age_key = self._age_key
         extra_status_fn = self._extra_status_fn
+        log_label = self._log_label
 
         class _Handler(BaseHTTPRequestHandler):
             def do_GET(self) -> None:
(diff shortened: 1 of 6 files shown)


workers/queue_backend/pg_queue/reaper.py
```diff
@@ -297,10 +312,180 @@ def _install_signal_handlers(self) -> None:
...
+        server.bound_port,
+        stale_after,
+    )
+    return server
+
+
 def main() -> None:
     logging.basicConfig(level=os.getenv("LOG_LEVEL", "INFO"))
+    raw_port = os.getenv("WORKER_PG_REAPER_HEALTH_PORT")
+    port = int(raw_port) if raw_port not in (None, "") else None
```
muhammad-ali-e: **[Medium · Code Reviewer + Silent-Failure-Hunter + Simplifier]** Bare `int(raw_port)` gives a context-free crash and regresses the consumer's pattern.

A misconfigured `WORKER_PG_REAPER_HEALTH_PORT=abc` (or a stray-whitespace value) crashes startup with `ValueError: invalid literal for int() with base 10: 'abc'` — no variable name, no hint which knob is wrong. This is launched from `run-worker.sh` via shell fallback expansion, so a garbled value is a realistic operator mistake. The sibling consumer's `main()` deliberately wraps this and re-raises as `Invalid WORKER_..._HEALTH_PORT=...`; the stale-seconds knob right below (`_reaper_health_stale_from_env`) also names the offending var — so this port path is the odd one out.

Note also: a negative / out-of-range value (e.g. `99999`) parses fine here but then raises `OverflowError`/`ValueError` *inside* `server.start()`, which is **not** an `OSError` and so escapes the graceful-degrade catch at line 460 (see that comment).

Suggest validating + naming the var at parse time:
```python
raw_port = os.getenv("WORKER_PG_REAPER_HEALTH_PORT")
if raw_port in (None, ""):
    port = None
else:
    try:
        port = int(raw_port)
    except ValueError as exc:
        raise ValueError(f"Invalid WORKER_PG_REAPER_HEALTH_PORT={raw_port!r}: {exc}") from exc
    if not (0 <= port <= 65535):
        raise ValueError(f"WORKER_PG_REAPER_HEALTH_PORT={port} out of range 0-65535")
```

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -297,10 +312,180 @@ def _install_signal_handlers(self) -> None:
...
+    if port is None:
+        logger.info(
+            "PG-queue reaper: WORKER_PG_REAPER_HEALTH_PORT unset — liveness "
+            "server disabled"
+        )
+        return None
+    server = ReaperLivenessServer(reaper, port=port, stale_after=stale_after)
+    try:
+        server.start()
+    except OSError:
```
muhammad-ali-e, line 398: **[Low · Silent-Failure-Hunter + Test-Analyzer]** This `except OSError` is the documented graceful-degrade path ("must never stop the reaper from running") and is correct for bind conflicts — but it has two gaps:

1. It misses `OverflowError`/`ValueError` raised by the socket layer for an out-of-range/negative port, which would propagate and kill the reaper — contradicting the docstring's promise. Best fixed by range-validating the port at parse time (see line 478); alternatively widen the catch to `(OSError, OverflowError, ValueError)`.
2. **This branch is entirely untested.** Add a test that makes `ReaperLivenessServer.start()` raise `OSError` (e.g. `monkeypatch` it, or bind a second server to an already-bound `bound_port`) and asserts `_maybe_start_health_server(...) is None` and that `logger.exception` fired. This is the resilience contract of the whole feature.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -297,10 +312,180 @@ def _install_signal_handlers(self) -> None:
...
+    the OS pick a free port (read back via :attr:`bound_port`) — used in tests.
+    """
+
+    _PATHS = frozenset({"/health", "/healthz", "/livez"})
+
+    def __init__(self, reaper: PgReaper, *, port: int, stale_after: float) -> None:
+        self._reaper = reaper
+        self._port = port
+        self._stale_after = stale_after
+        self._httpd: Any = None
```
muhammad-ali-e: **[Medium · Type-Design + Simplifier]** `_httpd`/`_thread` typed `Any` (and `_serve(httpd: Any)` at line 389) erase the started/inert lifecycle invariant and silently disable type-checking on every `.shutdown()`/`.server_close()`/`.server_address`/`.join()`/`.is_alive()` access. They also force the new `Any` import at line 42.

The class docstring says it mirrors the consumer's `LivenessServer`, but the consumer types these precisely as `HTTPServer | None` / `Thread | None` via a `TYPE_CHECKING` block. Match it (zero runtime cost):
```python
if TYPE_CHECKING:
    from http.server import HTTPServer
    from threading import Thread
...
self._httpd: HTTPServer | None = None
self._thread: Thread | None = None
```
Then drop `Any` from the `typing` import (line 42).

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -297,10 +312,180 @@ def _install_signal_handlers(self) -> None:
...
+        self._thread: Any = None
+
+    def start(self) -> None:
+        import json
+        import threading as _threading
```
muhammad-ali-e: **[Low · Simplifier]** Redundant `import threading as _threading` — `threading` is already imported at module level (line 40), and the alias dodges a name clash that doesn't exist. Drop the local import and use the module-level `threading.Thread(...)` at line 397. (The other three function-local imports — `json`, `http.server`, `urlsplit` — are deliberately deferred and match the consumer, so leave those.)

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -297,10 +312,180 @@ def _install_signal_handlers(self) -> None:
...
     lease = LeaderLease(default_worker_id())
-    PgReaper(lease).run()
+    reaper = PgReaper(lease)
+    health = _maybe_start_health_server(
+        reaper, port=port, stale_after=_reaper_health_stale_from_env()
+    )
+    try:
+        reaper.run()
+    finally:
+        if health is not None:
```
muhammad-ali-e, line 425: **[Medium · Test-Analyzer]** `main()`'s wiring is entirely untested — the only entry-point test just asserts the `__main__` re-export. Untested here: the env→int port conversion + empty-string-means-None branch (478), and this `try/finally` that calls `health.stop()` after `run()` returns (incl. the `if health is not None` guard).

Suggested test: patch `LeaderLease` and make `PgReaper.run` a no-op, set `WORKER_PG_REAPER_HEALTH_PORT=0`, spy `_maybe_start_health_server` returning a `MagicMock`, call `main()`, then assert the parsed int port was passed and `.stop.assert_called_once()`. Add a sibling asserting `port=None` when the var is unset/empty.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -297,10 +312,180 @@ def _install_signal_handlers(self) -> None:
...
+    NOT depend on DB reachability or leadership (a standby is healthy), or a
+    transient blip would crash-loop a fine process. The 200/503 verdict is purely
+    the tick heartbeat; ``is_leader`` is informational. Serves ``/health`` (also
+    ``/healthz``, ``/livez``) on ``0.0.0.0`` in a daemon thread; ``port=0`` lets
+    the OS pick a free port (read back via :attr:`bound_port`) — used in tests.
+    """
+
+    _PATHS = frozenset({"/health", "/healthz", "/livez"})
+
+    def __init__(self, reaper: PgReaper, *, port: int, stale_after: float) -> None:
```
muhammad-ali-e, line 331: **[Low · Type-Design + Code Reviewer]** `stale_after <= 0` is only rejected at the env boundary (`_reaper_health_stale_from_env`), not at the constructor. A direct caller (including tests) can build an always-503 probe that would crash-loop the pod. Consider a one-line guard in `__init__` (`if stale_after <= 0: raise ValueError(...)`), mirroring how `PgReaper.__init__` re-validates `interval_seconds` even though the env reader already does — the codebase's documented "load-bearing re-validation" convention.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -297,10 +312,180 @@ def _install_signal_handlers(self) -> None:
...
+        if self._httpd is not None:
+            raise RuntimeError("ReaperLivenessServer already started")
+
+        reaper = self._reaper
+        stale_after = self._stale_after
+        paths = self._PATHS
+
+        class _Handler(BaseHTTPRequestHandler):
+            def do_GET(self) -> None:
+                if urlsplit(self.path).path not in paths:
```
muhammad-ali-e: **[Low · Comment-Analyzer]** This correctly strips the query string (`urlsplit(self.path).path`) so `/health?foo=bar` still matches — but, unlike the consumer's equivalent, it has no explaining comment. A future maintainer could "simplify" it to `self.path` and silently break probes that carry a query string. Add the one-liner the consumer has, e.g. `# Strip any query string so /health?foo=bar still matches.`

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -297,10 +312,180 @@ def _install_signal_handlers(self) -> None:
...
+        httpd = HTTPServer(("0.0.0.0", self._port), _Handler)
+        self._httpd = httpd
+        self._thread = _threading.Thread(
+            target=_serve, args=(httpd,), daemon=True, name="pg-reaper-liveness"
+        )
+        self._thread.start()
+
+    @property
+    def bound_port(self) -> int:
+        """Actual listening port (resolves ``port=0``); the requested port if not started."""
```
muhammad-ali-e: **[Low · Comment-Analyzer]** Docstring is mildly misleading for the `port=0` + not-started (or post-`stop()`) case: it returns the literal sentinel `0`, not a usable "requested" port. Suggest: `"Actual listening port once started (resolves port=0); before start / after stop, returns the configured port value (which is 0 when the OS is asked to choose)."`

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -297,10 +312,180 @@ def _install_signal_handlers(self) -> None:
             )
 
 
+# Default staleness window for the liveness probe. Comfortably above the
```
muhammad-ali-e: **[Low · Comment-Analyzer]** This comment hard-codes "default 5s tick interval" in prose, duplicating `_DEFAULT_REAPER_INTERVAL_SECONDS = 5.0`. If that default ever changes, the comment rots. Reference the constant instead of the literal — the durable rationale is the headroom ratio, not the magic number.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -297,10 +312,180 @@ def _install_signal_handlers(self) -> None:
...
+# default 5s tick interval so a single slow cycle (a long sweep / DB blip)
+# doesn't flap the probe; an operator tightening the interval should tighten
+# this too.
+_DEFAULT_HEALTH_STALE_SECONDS = 30.0
+
+
+class ReaperLivenessServer:
```
muhammad-ali-e: **[Low / follow-up · Simplifier + Type-Design]** `ReaperLivenessServer` is a near-verbatim copy of the consumer's `LivenessServer` (~90 lines) — the only real differences are the heartbeat accessor, the `check`/age-key strings, the extra `is_leader` field, and log prefixes. Worth extracting a shared liveness server (parametrised by a heartbeat-age callable + an optional `extra` payload provider) so the two can't drift — the `Any`-typing divergence flagged above is itself copy-drift. This touches `consumer.py`, which is **out of this PR's scope**, so suggest tracking as a follow-up rather than blocking here.

muhammad-ali-e (commented):

Ran six specialised review agents — Code Reviewer, Silent Failure Hunter, Type Design Analyzer, PR Test Analyzer, Comment Analyzer, Code Simplifier — over the 4 in-scope files (base `feat/UN-3445-pg-queue-integration`). Overall this is a clean, well-documented change that faithfully mirrors the consumer's `LivenessServer`; **no Critical/High issues**. Inline comments below, deduped across agents and prioritised.

**Top items:** (1) the bare `int(raw_port)` port parse — context-free crash + an out-of-range port escapes the `OSError` graceful-degrade path; (2) `_httpd`/`_thread` typed `Any` where the consumer uses precise types; (3) `main()` port-parse + teardown are untested.

muhammad-ali-e:
SonarCloud duplication addressed — commit `0d2b3f721`.

The new `ReaperLivenessServer` was duplicating the consumer's `LivenessServer` (~19 lines of HTTP-probe boilerplate, over the 3% new-code gate). Extracted **one** generic `LivenessServer` into `queue_backend/pg_queue/liveness.py` (parameterised by a freshness callable + the payload check/age labels + an optional extra-status callable for the reaper's `is_leader`). Both sides are now thin subclasses that preserve their exact constructor signatures and wire payloads — so the boilerplate lives once and the consumer's behaviour/tests are unchanged.

reaper 37 tests + consumer liveness/health tests green; ruff + pre-commit clean.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (40a5f587a). Extracted _reaper_health_port_from_env() which names the var on a bad value (Invalid WORKER_PG_REAPER_HEALTH_PORT=...) and range-checks 0-65535 at parse time; main() uses it. Tested (non-int → named ValueError, out-of-range → raises).

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Addressed (40a5f587a). The range check at parse time (per the comment above) means an out-of-range port never reaches start(), so it cannot escape as OverflowError — the OSError catch stays correct for real bind conflicts (in-use/permission are OSError subclasses). Added test_health_server_bind_failure_degrades_to_none: makes start() raise OSError and asserts the result is None + logger.exception fired.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (40a5f587a) — in the shared liveness.py (where the boilerplate now lives after the dedup refactor): _httpd/_thread are typed HTTPServer|None / Thread|None via a TYPE_CHECKING block, and _serve(httpd: HTTPServer). The Any-typing was copy-drift; it is gone.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Already resolved by the dedup refactor (0d2b3f721): the boilerplate moved into the shared liveness.py, which imports threading at module level and uses threading.Thread(...) — no _threading alias.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Added (40a5f587a). TestMainWiring: patches LeaderLease + PgReaper.run (no-op), sets WORKER_PG_REAPER_HEALTH_PORT=0, spies _maybe_start_health_server, asserts the parsed int port reached the wiring and health.stop() fired in the finally; sibling asserts port=None when the var is unset.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (40a5f587a). The shared LivenessServer.__init__ now re-validates stale_after > 0 (mirroring PgReaper.__init__s load-bearing re-validation). Covered by test_stale_after_non_positive_rejected.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Already resolved by the refactor (0d2b3f721): the shared liveness.py do_GET carries the comment "# Strip any query string — a probe like /health?foo=bar must match."

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (40a5f587a). bound_port docstring now: "Actual listening port once started (resolves port=0 to the OS-chosen port). Before start / after stop, returns the configured port value (which is 0 when the OS is asked to choose)."

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (40a5f587a). The _DEFAULT_HEALTH_STALE_SECONDS comment now references _DEFAULT_REAPER_INTERVAL_SECONDS and frames the rationale as the headroom ratio, not the literal 5s.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Done in this PR (0d2b3f721): extracted the shared generic LivenessServer into queue_backend/pg_queue/liveness.py (parametrised by a freshness callable + check/age labels + an optional extra-status provider); consumer and reaper are now thin subclasses. The Any-typing divergence you spotted was indeed copy-drift, now gone with the single implementation. This also cleared the SonarCloud duplication gate.

muhammad-ali-e:
## Review addressed — commit `40a5f587a`

All 10 toolkit findings in. Validated each against the current code first — and notably, **3 were already resolved by the dedup refactor** (`0d2b3f721`) that landed in response to SonarCloud: the `import threading as _threading` alias, the missing query-strip comment, and the "extract a shared liveness server" follow-up itself.

**Newly fixed**
- **Health-port parse** — extracted `_reaper_health_port_from_env()`: names the var on a bad value (no more context-free `int('abc')` crash) and range-checks `0–65535` at parse time, so an out-of-range port can't escape the bind catch as `OverflowError` inside `start()`.
- **Typing** — `_httpd`/`_thread` are `HTTPServer|None` / `Thread|None` (via `TYPE_CHECKING`) in the shared `liveness.py`; the `Any` divergence was copy-drift, gone with the single implementation.
- **`stale_after > 0`** re-validated in the constructor (an always-503 probe would crash-loop the pod).
- `bound_port` docstring clarified for the `port=0`/not-started case; the stale-window comment references the interval constant instead of a hard-coded "5s".

**Test gaps closed (+9)** — port-env helper (unset/empty/valid/non-int-named/out-of-range); `main()` wiring (parsed port reaches the wiring + `health.stop()` in the `finally`; `port=None` when unset); `_maybe_start_health_server` OSError graceful-degrade → `None` + `logger.exception`; `stale_after<=0` constructor guard.

reaper 46 + consumer liveness 10 green; ruff + pre-commit clean. SonarCloud gate already passed.

workers/queue_backend/pg_queue/liveness.py
```diff
@@ -0,0 +1,164 @@
...
+                age = freshness_fn()
+                stale = age > stale_after
+                payload: dict[str, Any] = {
+                    "status": "unhealthy" if stale else "healthy",
+                    "check": check_name,
+                    age_key: round(age, 3),
+                    "stale_after_seconds": stale_after,
+                }
+                if extra_status_fn is not None:
+                    payload.update(extra_status_fn())
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed (b7776c0ee). The handler now builds the extra fields first and overlays the core fields, so a caller-supplied extra_status_fn can never clobber status/check/age_key/stale_after_seconds — core always wins. Added a test where extra_status_fn returns {"status":"HACKED","check":"HACKED","extra":1} and asserts status stays "healthy", check intact, and the non-reserved "extra" key is still preserved.

muhammad-ali-e:
## Greptile review addressed — commit `b7776c0ee`

Both findings valid and fixed:

- **`extra_status_fn` could clobber core payload fields** — the handler now builds the extra fields first and overlays the core fields (`status`/`check`/`age_key`/`stale_after_seconds`), so **core always wins**; a future caller's extra dict can't corrupt the `status` a monitor parses. Test added (an `extra_status_fn` returning `{"status":"HACKED",...}` leaves the core intact while preserving a non-reserved extra key).
- **Log namespace/text change from the dedup refactor** — good catch: the consumer's liveness warnings moved to the shared `liveness` logger with generic text. Added a `log_label` param (default `"pg-queue"`); the consumer passes `"pg-queue consumer"` and the reaper `"pg-queue reaper"`, so the messages stay attributable to the source process. (The logger *namespace* is now `…pg_queue.liveness` since that's where the shared code lives — but the message text identifies the process.)

57 tests green (reaper 47 + consumer liveness 10); ruff + pre-commit clean. Confidence was 4/5 with these two — both resolved.

PR #2062, opened 2026-06-16
UN-3559 [FEAT] PG Queue 9e PR 1 — execution transport seam (inert)

muhammad-ali-e:
## What & why

First of **3 PRs** for **9e** (the coupled-pipeline migration). This one establishes the **per-execution transport seam** and nothing else — the transport a workflow execution rides (`"celery"` | `"pg_queue"`) is resolved **once** at the creation chokepoint and **carried in the task payload**. It is **inert**: transport always resolves to `"celery"`, so behaviour is byte-identical to today.

Why a seam first: 9e migrates the coupled pipeline (`async_execute_bin` → file-batch fan-out → callback) onto the Postgres queue. Because the worker code is shared across both transports and later stages are enqueued from inside the workers, every stage needs to know — per execution — which transport to stay on. This PR lays that plumbing dead (always celery) so the next PR only adds the live PG branch.

**Design:** `workers/queue_backend/pg_queue/9e-design.md` (included). Chosen approach = **payload-carry**, deliberately **not** a `WorkflowExecution` column (no migration on the large shared table; all PG-specific state stays in droppable PG tables).

## Changes
- **core** — `WorkflowTransport` enum (`celery` | `pg_queue`) + `DEFAULT_WORKFLOW_TRANSPORT` (shared vocabulary).
- **backend** — `resolve_transport()` hardwired to celery (signature shaped for the Flipt wiring in PR 3); create-execution internal API returns `"transport"`; `execute_workflow_async` adds it to the `async_execute_bin` kwargs.
- **workers** — scheduler threads `transport` from the create response into the dispatch kwargs; `async_execute_bin_general` / `_execute_general_workflow` carry it onto the live `WorkflowContextData`.

## Tests
- backend `test_transport.py` — resolver + enum (5)
- worker `test_workflow_context_transport.py` — `WorkflowContextData` carry/default (2)
- dispatch characterisation updated — kwargs now include `transport`, + a new test that the backend-resolved transport is threaded verbatim
- **22 green** locally.

## Dev-test (against a running stack)
- Live `create-execution` internal API now returns `"transport":"celery"`.
- Workers killed + restarted on the new code (9 types RUNNING); a full **API-deployment execution completed** end-to-end on the new code (1 batch orchestrated, `async_execute_bin` succeeded, execution `COMPLETED`).

## Out of scope (later)
- PR 2: live PG routing (self-chaining + `PgBarrier` decrement) **+ the per-batch idempotency key** (ships together).
- PR 3: Flipt canary wiring. The Boolean flag `pg_queue_execution_enabled` (default `false`, no rollouts) already exists in dev — inert until PR 3 reads it.
- Rollout: ops (canary %, dashboards, teardown).

## Notes
- Targets `feat/UN-3445-pg-queue-integration` (not `main`).
- Ticket: UN-3559 (sub-task of the Phase 9 story).

### DIFF
--- a/backend/workflow_manager/internal_api_views.py
+++ b/backend/workflow_manager/internal_api_views.py
@@ -324,12 +324,9 @@ def create_workflow_execution(request):
         # Resolve the transport this execution rides (9e). Decided once here, at
         # the creation chokepoint, and returned so the caller carries it in the
         # dispatched task's payload — not persisted on the row. PR 1 always
-        # resolves "celery"; PR 3 wires Flipt in resolve_transport().
-        transport = resolve_transport(
-            workflow_id=str(workflow.id),
-            pipeline_id=data.get("pipeline_id"),
-            organization_id=org_id,
-        )
+        # resolves "celery"; PR 3 wires Flipt (keyed on workflow/pipeline/org) in
+        # resolve_transport().
+        transport = resolve_transport()
 
         return Response(
             {
--- a/backend/workflow_manager/internal_api_views.py
+++ b/backend/workflow_manager/internal_api_views.py
@@ -20,6 +20,7 @@ from tool_instance_v2.models import ToolInstance
 
 from workflow_manager.workflow_v2.enums import ExecutionStatus
 from workflow_manager.workflow_v2.models import Workflow, WorkflowExecution
+from workflow_manager.workflow_v2.transport import resolve_transport
 
 logger = logging.getLogger(__name__)
 
@@ -320,11 +321,22 @@ def create_workflow_execution(request):
             # Handle tags logic if needed
             pass
 
+        # Resolve the transport this execution rides (9e). Decided once here, at
+        # the creation chokepoint, and returned so the caller carries it in the
+        # dispatched task's payload — not persisted on the row. PR 1 always
+        # resolves "celery"; PR 3 wires Flipt in resolve_transport().
+        transport = resolve_transport(
+            workflow_id=str(workflow.id),
+            pipeline_id=data.get("pipeline_id"),
+            organization_id=org_id,
+        )
+
         return Response(
             {
                 "execution_id": str(execution.id),
                 "status": execution.status,
                 "execution_log_id": execution.execution_log_id,  # Return for workers to use
+                "transport": transport,
             }
         )
 
--- a/backend/workflow_manager/workflow_v2/transport.py
+++ b/backend/workflow_manager/workflow_v2/transport.py
@@ -16,11 +16,7 @@ kill-switch and failing closed to Celery.
 
 from __future__ import annotations
 
-import logging
-
 from unstract.core.data_models import WorkflowTransport
-
-logger = logging.getLogger(__name__)
 
 
 def resolve_transport(
@@ -41,5 +37,10 @@ def resolve_transport(
     Note:
         PR 1 always returns ``"celery"``. The arguments are accepted now so the
         call sites and the contract are stable when PR 3 wires Flipt in here.
+        PR 3 must wrap the Flipt evaluation in ``try/except`` and fall back to
+        ``WorkflowTransport.CELERY.value`` (fail-closed), so a Flipt outage can
+        never break execution creation — mirroring ``normalize_transport`` on
+        the read side. (A module ``logger`` is intentionally omitted until then,
+        when the fallback path needs it.)
     """
     return WorkflowTransport.CELERY.value
--- a/backend/workflow_manager/workflow_v2/transport.py
+++ b/backend/workflow_manager/workflow_v2/transport.py
@@ -1,0 +1,45 @@
+"""Transport resolution for a workflow execution (9e).
+
+A workflow execution rides one transport end-to-end — legacy Celery/RabbitMQ
+or the bespoke Postgres queue. The choice is resolved **once**, at the
+execution-creation chokepoint, and returned to the caller so it can be carried
+in the dispatched task's payload (see ``workers/queue_backend/pg_queue/
+9e-design.md``). It is deliberately **not** persisted on the ``WorkflowExecution``
+row: the payload is the single carrier, durable for PG via the queue row's
+JSONB, and the giant shared table is never migrated for this work.
+
+PR 1 (this seam) hardwires the result to Celery, so behaviour is byte-identical
+to today. PR 3 replaces the body with a Flipt evaluation (percentage rollout via
+``entity_id`` hashing + per-org segment via ``context``), wrapped by an env
+kill-switch and failing closed to Celery.
+"""
+
+from __future__ import annotations
+
+import logging
+
+from unstract.core.data_models import WorkflowTransport
+
+logger = logging.getLogger(__name__)
+
+
+def resolve_transport(
+    workflow_id: str,
+    pipeline_id: str | None = None,
+    organization_id: str | None = None,
+) -> str:
+    """Resolve the transport for a new workflow execution.
+
+    Args:
+        workflow_id: The workflow being executed.
+        pipeline_id: The pipeline, if this execution belongs to one.
+        organization_id: The owning organization (for per-org rollout in PR 3).
+
+    Returns:
+        The transport value (a :class:`WorkflowTransport` value string).
+
+    Note:
+        PR 1 always returns ``"celery"``. The arguments are accepted now so the
+        call sites and the contract are stable when PR 3 wires Flipt in here.
+    """
+    return WorkflowTransport.CELERY.value
--- a/backend/workflow_manager/workflow_v2/transport.py
+++ b/backend/workflow_manager/workflow_v2/transport.py
@@ -19,28 +19,22 @@ from __future__ import annotations
 from unstract.core.data_models import WorkflowTransport
 
 
-def resolve_transport(
-    workflow_id: str,
-    pipeline_id: str | None = None,
-    organization_id: str | None = None,
-) -> str:
+def resolve_transport() -> str:
     """Resolve the transport for a new workflow execution.
-
-    Args:
-        workflow_id: The workflow being executed.
-        pipeline_id: The pipeline, if this execution belongs to one.
-        organization_id: The owning organization (for per-org rollout in PR 3).
 
     Returns:
         The transport value (a :class:`WorkflowTransport` value string).
 
     Note:
-        PR 1 always returns ``"celery"``. The arguments are accepted now so the
-        call sites and the contract are stable when PR 3 wires Flipt in here.
-        PR 3 must wrap the Flipt evaluation in ``try/except`` and fall back to
+        PR 1 always returns ``"celery"`` and takes no arguments (the inert seam
+        needs no inputs). PR 3 reintroduces ``workflow_id`` / ``pipeline_id`` /
+        ``organization_id`` parameters when it wires Flipt here — percentage
+        rollout via ``entity_id`` hashing + per-org segment via ``context`` —
+        and updates the two call sites (``internal_api_views`` view and
+        ``workflow_helper.execute_workflow_async``). PR 3 must wrap the Flipt
+        evaluation in ``try/except`` and fall back to
         ``WorkflowTransport.CELERY.value`` (fail-closed), so a Flipt outage can
         never break execution creation — mirroring ``normalize_transport`` on
-        the read side. (A module ``logger`` is intentionally omitted until then,
-        when the fallback path needs it.)
+        the read side.
     """
     return WorkflowTransport.CELERY.value
--- a/backend/workflow_manager/workflow_v2/workflow_helper.py
+++ b/backend/workflow_manager/workflow_v2/workflow_helper.py
@@ -520,11 +520,7 @@ class WorkflowHelper:
             # an independent Flipt evaluation; before then, both must be funnelled
             # through a single per-execution chokepoint (so a percentage re-roll
             # can't split one execution across transports). Tracked for PR 3.
-            transport = resolve_transport(
-                workflow_id=workflow_id,
-                pipeline_id=pipeline_id,
-                organization_id=org_schema,
-            )
+            transport = resolve_transport()
             async_execution: AsyncResult = celery_app.send_task(
                 "async_execute_bin",
                 args=[
--- a/backend/workflow_manager/workflow_v2/workflow_helper.py
+++ b/backend/workflow_manager/workflow_v2/workflow_helper.py
@@ -62,11 +62,15 @@ from workflow_manager.workflow_v2.transport import resolve_transport
 logger = logging.getLogger(__name__)
 
 # Parameters to exclude when calling create_workflow_execution
+# (the classmethod takes no **kwargs, so any task kwarg not a real parameter —
+# incl. the 9e ``transport`` carried in the async_execute_bin payload — must be
+# filtered out here before it reaches the legacy ``execute_workflow`` path).
 EXECUTION_EXCLUDED_PARAMS = {
     "llm_profile_id",
     "hitl_queue_name",
     "hitl_packet_id",
     "custom_data",
+    "transport",
 }
 
 
@@ -507,6 +511,15 @@ class WorkflowHelper:
             # Resolve the transport this execution rides (9e) and carry it in the
             # task payload — the pipeline reads it to stay on one transport
             # end-to-end. PR 1 always resolves "celery" (no behaviour change).
+            #
+            # NOTE (deliberate, two resolution sites): transport is resolved here
+            # for the API/manual/async paths and separately in
+            # internal_api_views.create_workflow_execution for the scheduler
+            # path. These are DISTINCT entry paths — never a double-resolution of
+            # the same execution — so today they cannot diverge. PR 3 makes each
+            # an independent Flipt evaluation; before then, both must be funnelled
+            # through a single per-execution chokepoint (so a percentage re-roll
+            # can't split one execution across transports). Tracked for PR 3.
             transport = resolve_transport(
                 workflow_id=workflow_id,
                 pipeline_id=pipeline_id,
--- a/backend/workflow_manager/workflow_v2/workflow_helper.py
+++ b/backend/workflow_manager/workflow_v2/workflow_helper.py
@@ -57,6 +57,7 @@ from workflow_manager.workflow_v2.execution import WorkflowExecutionServiceHelpe
 from workflow_manager.workflow_v2.file_history_helper import FileHistoryHelper
 from workflow_manager.workflow_v2.models.execution import WorkflowExecution
 from workflow_manager.workflow_v2.models.workflow import Workflow
+from workflow_manager.workflow_v2.transport import resolve_transport
 
 logger = logging.getLogger(__name__)
 
@@ -503,6 +504,14 @@ class WorkflowHelper:
             }
             org_schema = UserContext.get_organization_identifier()
             log_events_id = StateStore.get(Common.LOG_EVENTS_ID)
+            # Resolve the transport this execution rides (9e) and carry it in the
+            # task payload — the pipeline reads it to stay on one transport
+            # end-to-end. PR 1 always resolves "celery" (no behaviour change).
+            transport = resolve_transport(
+                workflow_id=workflow_id,
+                pipeline_id=pipeline_id,
+                organization_id=org_schema,
+            )
             async_execution: AsyncResult = celery_app.send_task(
                 "async_execute_bin",
                 args=[
@@ -521,6 +530,7 @@ class WorkflowHelper:
                     "hitl_queue_name": hitl_queue_name,
                     "hitl_packet_id": hitl_packet_id,
                     "custom_data": custom_data,
+                    "transport": transport,
                 },
                 queue=queue,
             )
--- a/workers/general/tasks.py
+++ b/workers/general/tasks.py
@@ -51,6 +51,7 @@ from worker import app, config
 
 # Import shared data models for type safety
 from unstract.core.data_models import (
+    DEFAULT_WORKFLOW_TRANSPORT,
     ExecutionStatus,
     FileBatchData,
     FileHashData,
@@ -156,6 +157,7 @@ def async_execute_bin_general(
     pipeline_id: str | None = None,
     log_events_id: str | None = None,
     use_file_history: bool = False,
+    transport: str = DEFAULT_WORKFLOW_TRANSPORT,
     **kwargs: dict[str, Any],
 ) -> dict[str, Any]:
     """Lightweight general workflow execution task.
@@ -267,6 +269,7 @@ def async_execute_bin_general(
                 use_file_history,
                 scheduled,
                 schema_name,
+                transport=transport,
                 **kwargs,
             )
 
@@ -462,6 +465,7 @@ def _execute_general_workflow(
     use_file_history: bool,
     scheduled: bool,
     schema_name: str,
+    transport: str = DEFAULT_WORKFLOW_TRANSPORT,
     **kwargs: dict[str, Any],
 ) -> dict[str, Any]:
     """Execute general workflow specific logic for ETL/TASK workflows.
@@ -522,6 +526,7 @@ def _execute_general_workflow(
                 "execution_mode": execution_mode,
             },
             is_scheduled=scheduled,
+            transport=transport,
         )
 
         logger.info(
--- a/workers/queue_backend/pg_queue/9e-design.md
+++ b/workers/queue_backend/pg_queue/9e-design.md
@@ -1,0 +1,318 @@
+# 9e — Coupled-Pipeline Migration (Design)
+
+**Status:** design locked, pre-implementation
+**Epic:** UN-3445 (PG Queue) · **Phase:** 9e
+**Predecessors:** 9a–9c (dispatch seam + PG consumer), 9d (leader-election lease + reaper / barrier-orphan sweep) — all merged.
+
+This document defines how a **whole workflow execution** is migrated from the
+legacy Celery/RabbitMQ transport to the bespoke Postgres queue, and why the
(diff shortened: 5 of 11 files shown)


muhammad-ali-e (commented):
## Automated PR review — PG Queue 9e PR 1 (transport seam)

Reviewed with the PR Review Toolkit (code review, silent-failure, type-design, test, comment, simplifier agents) against base `feat/UN-3445-pg-queue-integration`. Findings verified by hand before posting.

**Overall:** clean, well-scoped, genuinely inert plumbing. No production-breaking issues. The notable verification result: an initially-flagged "CRITICAL crash on the legacy `execute_bin` path" is a **false positive** — the `celery`/`celery_api_deployments` queues are consumed by the workers app (`-A worker worker`), whose `async_execute_bin` router forwards `transport` via `**kwargs` to `async_execute_bin_general` (which accepts it). The backend celery worker (`-A backend worker`) only consumes `dashboard_metric_events`, so the legacy `execute_bin` is not a live consumer and does not crash. It's noted below only as dead-code hygiene.

The substantive themes are: (1) the enum exists but is never enforced at any boundary, (2) coverage is asymmetric (scheduler path tested, the dominant `workflow_helper.py`/worker paths are not), and (3) the design doc has several stale/incorrect anchors. Individual findings are inline. All are LOW/MEDIUM and safe to address in follow-ups; none block the inert merge.

workers/shared/models/execution_models.py
```diff
@@ -385,6 +386,10 @@ class WorkflowContextData:
     settings: dict[str, Any] | None = None
     metadata: dict[str, Any] | None = None
     is_scheduled: bool = False
+    # Transport this execution rides end-to-end (9e). Carried in the task
+    # payload from the creation chokepoint; read at the fan-out to keep the
+    # whole pipeline on one transport. Defaults to legacy Celery.
+    transport: str = DEFAULT_WORKFLOW_TRANSPORT
```
muhammad-ali-e, line 397: **[MEDIUM · type-design] Enum exists but `transport` is never validated.** `WorkflowTransport` is a real closed enum, yet this field is a plain `str` with no construction-time check — unlike the sibling `workflow_type`, which `__post_init__` validates against `PipelineType` a few lines below. This field is populated from an external task payload (and JSONB on the PG path), so a malformed value (`"celary"`, `None`, `"pg-queue"`) sails through and only surfaces far away at PR 2's read site. Mirror the existing pattern:
```python
valid = {t.value for t in WorkflowTransport}
if self.transport not in valid:
    raise ValueError(f"Invalid transport {self.transport!r}; must be one of {valid}")
```
Keep the field `str` (it must JSON/JSONB round-trip) but reject illegal values at construction. Highest-leverage change here — converts the enum from documentation into an enforced invariant at near-zero cost.

workers/shared/models/execution_models.py
```diff
@@ -385,6 +386,10 @@ class WorkflowContextData:
     settings: dict[str, Any] | None = None
     metadata: dict[str, Any] | None = None
     is_scheduled: bool = False
+    # Transport this execution rides end-to-end (9e). Carried in the task
+    # payload from the creation chokepoint; read at the fan-out to keep the
```
muhammad-ali-e: **[MEDIUM · comment accuracy] "read at the fan-out" is not true in this PR.** Nothing reads `WorkflowContextData.transport` for routing in PR 1 (grep confirms only the field def + tests reference it); the fan-out read lands in PR 2. As written, a maintainer will search for the fan-out read and not find it. Reword to present-tense reality, e.g. "Populated from the inbound task payload; the fan-out read that keeps the pipeline on one transport lands in PR 2." (The companion test docstring already phrases this correctly — "so the fan-out (PR 2) can read it".)

workers/scheduler/tasks.py
```diff
@@ -129,6 +133,10 @@ def _execute_scheduled_workflow(
             execution_request.to_dict()
         )
         execution_id = workflow_execution.get("execution_id")
+        # Transport this execution rides (9e), decided by the backend at creation
+        # and carried in the dispatched task's payload. Defaults to celery if the
+        # backend doesn't return it (older backend / no-op until PR 3).
+        transport = workflow_execution.get("transport", DEFAULT_WORKFLOW_TRANSPORT)
```
muhammad-ali-e: **[MEDIUM · silent failure] `.get(default)` only covers the *absent* key, not an *invalid* value.** Two distinct failure modes collapse here: (a) key missing (older backend) → legitimate default-to-celery; (b) key present but garbage (`"pg-queue"` typo, `None`, `""`) → passes through untouched, with no log. In PR 3 a present-but-invalid value silently dispatches an execution onto an unknown transport — a textbook hard-to-debug "execution vanished" failure. Coerce + fail closed loudly:
```python
raw = workflow_execution.get("transport", DEFAULT_WORKFLOW_TRANSPORT)
try:
    transport = WorkflowTransport(raw).value
except ValueError:
    logger.warning(f"[exec:{execution_id}] unrecognized transport {raw!r}; "
                   f"falling back to {DEFAULT_WORKFLOW_TRANSPORT!r} (version skew?)")
    transport = DEFAULT_WORKFLOW_TRANSPORT
```
The absent-key default stays intentional; only the invalid case becomes loud.

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -521,6 +530,7 @@ def execute_workflow_async(
                     "hitl_queue_name": hitl_queue_name,
                     "hitl_packet_id": hitl_packet_id,
                     "custom_data": custom_data,
+                    "transport": transport,
```
muhammad-ali-e, line 542: **[MEDIUM · test gap] This dispatch site — the dominant one — is untested.** API-deployment, async, and manual single-workflow runs all flow through `execute_bin` → this `send_task`. The scheduler path has `test_dispatch_kwargs_layout` + `test_dispatch_carries_backend_resolved_transport`, but this `send_task` payload has **no** test asserting it carries `transport`. The cross-site AST canary (`TestDispatchSiteInventory`) covers worker-side dispatch only, not this backend site. Add a test that patches `celery_app.send_task`, invokes `execute_workflow_async(...)`, and asserts `send_task.call_args.kwargs["kwargs"]["transport"] == "celery"`.

**[LOW · dead-code hygiene]** Note also: the superseded legacy `execute_bin` → `execute_workflow` filters kwargs via `EXECUTION_EXCLUDED_PARAMS` (which omits `transport`) into `create_workflow_execution()` (no `transport` param / no `**kwargs`). This path is **not** a live consumer of `celery`/`celery_api_deployments` today (the workers app's router handles those), so it does not crash — but if that legacy task is ever run it would `TypeError`. Either add `"transport"` to `EXECUTION_EXCLUDED_PARAMS` or retire the dead task.

workers/general/tasks.py
```diff
@@ -522,6 +526,7 @@ def _execute_general_workflow(
                 "execution_mode": execution_mode,
             },
             is_scheduled=scheduled,
+            transport=transport,
```
muhammad-ali-e, line 529: **[MEDIUM · test gap] Worker-side threading is unexercised.** This is where the carried `transport` becomes the live `WorkflowContextData` the PR 2 fan-out will read. `test_workflow_context_transport.py` only tests the dataclass default/round-trip directly — it never asserts that `async_execute_bin_general` actually passes its `transport` param through to here. A regression dropping `transport=transport` (here or the forward at line 272) would be silently masked by the celery default. Add a test that calls `_execute_general_workflow(..., transport="pg_queue")` with collaborators mocked and asserts the constructed context's `.transport == "pg_queue"`, plus an omit-transport case asserting it defaults to `"celery"` end-to-end.

backend/workflow_manager/internal_api_views.py
```diff
@@ -320,11 +321,22 @@ def create_workflow_execution(request):
...
+            pipeline_id=data.get("pipeline_id"),
+            organization_id=org_id,
+        )
+
         return Response(
             {
                 "execution_id": str(execution.id),
                 "status": execution.status,
                 "execution_log_id": execution.execution_log_id,  # Return for workers to use
+                "transport": transport,
```
muhammad-ali-e, line 336: **[MEDIUM · test gap] The producer end of the payload-carry contract is unverified.** This response field is the source of truth the scheduler reads (`test_dispatch_carries_backend_resolved_transport` *mocks* it as already present). No test asserts the real endpoint actually emits `transport == "celery"`. If this key were dropped, every scheduler dispatch would silently fall back to the default and the seam becomes a no-op. Add a focused view test asserting `response.data["transport"] == WorkflowTransport.CELERY.value`.

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -503,6 +504,14 @@ def execute_workflow_async(
             }
             org_schema = UserContext.get_organization_identifier()
             log_events_id = StateStore.get(Common.LOG_EVENTS_ID)
+            # Resolve the transport this execution rides (9e) and carry it in the
+            # task payload — the pipeline reads it to stay on one transport
+            # end-to-end. PR 1 always resolves "celery" (no behaviour change).
+            transport = resolve_transport(
```
muhammad-ali-e: **[MEDIUM · design / asymmetry] Two independent resolution sites.** Transport is resolved here (`execute_workflow_async`, API/manual paths) *and* separately in `internal_api_views.create_workflow_execution` (scheduler path, which returns it). These are genuinely different entry paths — not a double-resolution of the same execution today — and it's inert while `resolve_transport` is hardwired. But in PR 3 they become **two independent Flipt evaluations** that can diverge (e.g. percentage re-roll), reintroducing exactly the split-brain §1/§2 of the design doc warns against. Recommend funneling both through one shared resolution (ideally at the actual row-creation chokepoint, `execution.py:create_workflow_execution`) before PR 3, or explicitly documenting the deliberate two-chokepoint design.

backend/workflow_manager/workflow_v2/transport.py
```diff
@@ -0,0 +1,45 @@
...
+        organization_id: The owning organization (for per-org rollout in PR 3).
+
+    Returns:
+        The transport value (a :class:`WorkflowTransport` value string).
+
+    Note:
+        PR 1 always returns ``"celery"``. The arguments are accepted now so the
+        call sites and the contract are stable when PR 3 wires Flipt in here.
+    """
+    return WorkflowTransport.CELERY.value
```
muhammad-ali-e, line 40: **[LOW · forward-looking] No fail-closed boundary, and return type erases the enum.** Two small things for PR 3 readiness: (1) The docstring/design promise that this "fails closed to Celery on any error" has no structural home — when PR 3 drops Flipt logic in, any exception propagates and breaks execution creation. Consider establishing the `try/except → return DEFAULT_WORKFLOW_TRANSPORT` (logged) shape now, or at least leave a marker, so PR 3 inherits the correct posture. (2) This is the one function whose entire job is to *decide* the transport — returning `WorkflowTransport` instead of `str` (serializing via `.value` at the two emit sites) would make PR 3's flag→transport mapping type-checked end-to-end. The docstring already half-claims this ("a `WorkflowTransport` value string").

backend/workflow_manager/workflow_v2/transport.py
```diff
@@ -0,0 +1,45 @@
...
+kill-switch and failing closed to Celery.
+"""
+
+from __future__ import annotations
+
+import logging
+
+from unstract.core.data_models import WorkflowTransport
+
+logger = logging.getLogger(__name__)
```
muhammad-ali-e: **[LOW · lint] Unused `logger` in PR 1.** It's only needed once the Flipt evaluation lands (PR 3) and may trip an unused-binding lint flag now. Either defer the import to PR 3 or add a brief note that it's intentionally staged.

workers/queue_backend/pg_queue/9e-design.md
```diff
@@ -0,0 +1,318 @@
...
+> belt for the PG path and ships **with** it (PR 2), never separately ahead of
+> it. We stop at 3 (not 1) to keep PR 1 **inert and isolated** — it merges with
+> near-zero review risk and makes the branch carry `transport` everywhere, so
+> PR 2 is reviewed purely as "the PG behaviour" (the highest-blast-radius code in
+> the epic). Collapsing the inert seam into the risky rewire is the one merge to
+> avoid.
+
+### PR 1 — transport seam (inert, no routing change)
+- `resolve_transport(...)` helper at `create_workflow_execution`
+  (`backend/workflow_manager/workflow_v2/execution.py:126`); **PR 1 hardwires it
```
muhammad-ali-e: **[MEDIUM · stale anchor] Wrong `create_workflow_execution`.** `execution.py:126` is `WorkflowExecutionServiceHelper.create_workflow_execution` — the classmethod that builds the WorkflowExecution row — and it does **not** call `resolve_transport` (grep: zero references in `execution.py`). The seam actually lives in `internal_api_views.py:328` (the HTTP view, returns `transport`) and `workflow_helper.py:510`. A reader following this anchor inspects the wrong function and concludes the seam is missing. Same wrong anchor in the §8 table (line 299). Point at the real call sites instead.

workers/queue_backend/pg_queue/9e-design.md
```diff
@@ -0,0 +1,318 @@
...
+  to `"celery"`** — Flipt wiring is PR 3.
+- Add `transport: str = "celery"` to `WorkflowExecutionContext`
```
muhammad-ali-e: **[MEDIUM · wrong class name] `WorkflowExecutionContext` vs `WorkflowContextData`.** The field was added to `WorkflowContextData` (`execution_models.py:392`), not `WorkflowExecutionContext` (a different class in the same file that did not change). Same error at lines 156 and 302. The PR's own field comment and tests correctly say `WorkflowContextData`, so the doc is the outlier — replace all three references.

workers/queue_backend/pg_queue/9e-design.md
```diff
@@ -0,0 +1,318 @@
...
+  (`workers/shared/models/execution_models.py`); populate from inbound kwargs.
+- Thread `transport` into the kwargs of every pipeline dispatch site
+  (`workflow_helper.py` stage-1/2/3 dispatches) across **all four entry paths**
```
muhammad-ali-e: **[MEDIUM · overstates PR 1 scope] "stage-1/2/3 dispatches" / "re-emitted into every downstream dispatch".** The diff threads `transport` into **stage-1 only** (`workflow_helper.py:533` send + `scheduler/tasks.py:174`). The stage-2 fan-out (`process_file_batch`) and stage-3 chord (`process_batch_callback`) were not modified and carry no `transport`; on the worker side it lands on `WorkflowContextData` but is never re-emitted downstream. The §3 sentence "re-emitted into every downstream dispatch" (line 157) is currently false for the merged code — scope these claims to stage-1, with stage-2/3 explicitly marked as PR 2.

workers/queue_backend/pg_queue/9e-design.md
```diff
@@ -0,0 +1,318 @@
...
+  ramp + Celery teardown criteria.
+
+---
+
+## 8. Key files (call-graph anchors)
+
+| Concern                     | Location |
+|-----------------------------|----------|
+| Creation chokepoint         | `backend/workflow_manager/workflow_v2/execution.py:126` (`create_workflow_execution`) |
+| Callers (API/async/sched/UI)| `api_v2/deployment_helper.py:202`, `workflow_manager/workflow_v2/workflow_helper.py:686/744/935` |
```
muhammad-ali-e: **[LOW · stale anchors] §8 "Callers" line numbers don't point at the seam.** `deployment_helper.py:202` calls the row-builder classmethod (not the transport-resolving function), and `workflow_helper.py:686/744/935` are a dict-comprehension / a param default / a closing paren respectively — none are seam call sites. Prefer describing the four entry paths by name (API / manual / scheduled / async) over brittle line numbers, or point at the verified sites (`internal_api_views.py:328`, `workflow_helper.py:510`).

workers/queue_backend/pg_queue/9e-design.md
```diff
@@ -0,0 +1,318 @@
...
+answered **once**, at the single creation chokepoint, using the existing
+`@unstract/flags` Flipt infrastructure
+(`check_feature_flag_status(flag_key, entity_id, context)`):
+
+- `entity_id` → consistent **percentage-rollout** hashing.
+- `context={"organization_id": ...}` → **per-org** segment targeting.
+
+This is the only place Flipt is consulted.
+
+#### Flipt flag contract (fixed — created 2026-06-16, Flipt v2.3.1)
```
muhammad-ali-e: **[LOW · comment rot] Hard-coded Flipt version + creation date.** `v2.3.1` (also line 76) duplicates a pin that lives in `docker/docker-compose-dev-essentials.yaml` — the next bump silently desyncs the doc; reference the compose file as the single source instead. The embedded creation date (`2026-06-16`) and the "no rollouts yet / inert until PR 3" status (lines 71-72, 80, 287) describe live external state that goes stale the moment rollout ops begin — better tracked in the PR description / ticket than in a durable design doc.

workers/shared/models/execution_models.py
muhammad-ali-e: Fixed in 2bc40a207. Added `normalize_transport()` (core) and call it in `WorkflowContextData.__post_init__` to coerce an unknown value fail-closed to celery (with a warning) rather than raise. Deliberately coerce-not-raise (unlike `workflow_type`): transport has an explicit safe default by design, so a bad value degrades to Celery, not crash the execution. Test added (`test_invalid_transport_coerced_to_celery`).

workers/shared/models/execution_models.py
muhammad-ali-e: Fixed in 2bc40a207 — comment is now present-tense: "Populated from the inbound task payload; the fan-out read … lands in PR 2", plus a note that it's coerced fail-closed in `__post_init__`.

workers/scheduler/tasks.py
muhammad-ali-e: Fixed in 2bc40a207 — the scheduler now runs the value through `normalize_transport(..., logger=logger, context=...)`: absent key still defaults to celery, but a present-but-invalid value is coerced to celery with a loud warning (fail-closed). Covered by the new logging test.

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: Partially addressed in 2bc40a207: added `"transport"` to `EXECUTION_EXCLUDED_PARAMS` (so the legacy `execute_bin` path can't `TypeError`) and documented the two-site design. The heavily-mocked send_task-carries-transport unit test is deferred — `execute_workflow_async` hits the DB after `send_task` (needs Django/DB fixtures); the kwarg is a literal and the contract is covered by the live dev-test + scheduler characterisation test. Tracked as follow-up.

workers/general/tasks.py
muhammad-ali-e: Deferred as follow-up in 2bc40a207. The threading is a straight param pass-through (async_execute_bin_general -> _execute_general_workflow -> WorkflowContextData), verified by an import/signature check; a clean unit test needs heavy collaborator mocking (the function continues past construction into source connectors). The dataclass default + invalid-coercion are now unit-tested.

backend/workflow_manager/internal_api_views.py
muhammad-ali-e: The producer contract is verified live: the real `create_workflow_execution` endpoint returned `{"transport":"celery"}` in the dev-test (see PR description). A Django-DB unit test is deferred — `APIRequestFactory` needs the Django settings/DB infra this lightweight unit suite intentionally avoids. Tracked as follow-up.

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: Documented in 2bc40a207. Added a NOTE at `execute_workflow_async` explaining the deliberate two-site design (distinct entry paths, never a double-resolution of the same execution today) and the PR-3 requirement to funnel both through one per-execution chokepoint before the Flipt evals can diverge. Mirrored in design doc §8.

backend/workflow_manager/workflow_v2/transport.py
muhammad-ali-e: Addressed in 2bc40a207: added an explicit PR-3 marker in the docstring — PR 3 must wrap the Flipt eval in try/except -> DEFAULT_WORKFLOW_TRANSPORT (fail-closed, mirroring `normalize_transport`). Kept the `str` return for PR-1 minimalism; the enum-return suggestion is noted for PR 3.

backend/workflow_manager/workflow_v2/transport.py
muhammad-ali-e: Fixed in 2bc40a207 — removed the unused `logger` (and `import logging`); left a one-line note that it's intentionally re-added in PR 3 when the fail-closed fallback needs it.

workers/queue_backend/pg_queue/9e-design.md
muhammad-ali-e: Fixed in 2bc40a207 — anchors now point at the real seam (`transport.py` / `internal_api_views.create_workflow_execution` view / `workflow_helper.execute_workflow_async`), and `execution.py`'s classmethod is labelled the row-builder that does NOT resolve transport. §8 table updated too.

workers/queue_backend/pg_queue/9e-design.md
muhammad-ali-e: Fixed in 2bc40a207 — all three references now say `WorkflowContextData`, with an explicit note that `WorkflowExecutionContext` is dead scaffolding and not the carrier.

workers/queue_backend/pg_queue/9e-design.md
muhammad-ali-e: Fixed in 2bc40a207 — scoped to stage-1 only; the §3 're-emitted into every downstream dispatch' line and the PR-1 slice now state stage-2 fan-out / stage-3 callback are unmodified and land in PR 2.

workers/queue_backend/pg_queue/9e-design.md
muhammad-ali-e: Fixed in 2bc40a207 — the §8 table now describes the emit sites + entry paths by name (API / async / scheduled / manual) instead of brittle line numbers.

workers/queue_backend/pg_queue/9e-design.md
muhammad-ali-e: Fixed in 2bc40a207 — the Flipt version row now points at the compose file as the single source (no duplicated tag), and the creation date / live-state is removed from the durable doc (tracked in the PR/ticket instead).

muhammad-ali-e:
## Review addressed — commit `2bc40a207`

Thanks for the thorough toolkit pass. All 14 findings handled; per-thread replies above. Summary:

**Code (all fixed):**
- **Validation + silent-failure (×2):** added `normalize_transport()` in core — fail-closed coercion of any inbound transport (unknown / `None` / `""` → `celery` + warning). Applied at the **scheduler read boundary** and in **`WorkflowContextData.__post_init__`**, so a garbage payload value can never reach the PR 2 fan-out read. Coerce-not-raise is deliberate (transport has a safe default by design, unlike `workflow_type`).
- **Dead-code hygiene:** `"transport"` added to `EXECUTION_EXCLUDED_PARAMS` (legacy `execute_bin` path can't `TypeError`).
- **Two-resolution-sites:** documented the deliberate two-site design + the PR-3 single-chokepoint requirement (code comment + design doc).
- **`transport.py`:** dropped the unused `logger`; left a PR-3 fail-closed marker.
- **Comment accuracy:** `WorkflowContextData.transport` comment is now present-tense.

**Design doc (all fixed):** correct anchors (`transport.py` / `internal_api_views` view / `execute_workflow_async`, not `execution.py:126`); `WorkflowContextData` not `WorkflowExecutionContext`; scoped claims to stage-1 (stage-2/3 = PR 2); §8 callers by name not line numbers; removed the hard-coded Flipt version/date/live-state.

**Tests added:** `normalize_transport` (passthrough / invalid→celery / `None` / logging-on-invalid / no-warn-on-valid) and `WorkflowContextData` invalid-transport coercion. **28 green** (10 backend + 18 worker).

**Deferred as follow-up test coverage** (3 test-gap findings — agreed non-blocking; heavier integration-style tests):
- `execute_workflow_async` `send_task`-carries-transport — hits the DB after `send_task`; the kwarg is a literal, and the contract is exercised by the **live dev-test** + scheduler characterisation test.
- `_execute_general_workflow` threading — straight param pass-through (verified by import/signature check); a clean unit test needs heavy collaborator mocking. Dataclass default + coercion are now unit-tested.
- create-execution **view** returns `transport` — verified **live** (the real endpoint returned `{"transport":"celery"}`); a Django-DB unit test needs settings/DB infra the unit suite intentionally avoids.

Still inert — transport always resolves to `celery`.

muhammad-ali-e:
Greptile P1 (scheduler `normalize_transport` ordering) addressed in `aa655c508`: moved the transport resolution **below** the `execution_id` null-guard in `_execute_scheduled_workflow`, so the no-`execution_id` error path no longer logs `[exec:None]` and no longer computes-then-discards the transport. Pre-commit test-selection green. (This was the only open finding — confidence should now clear to 5/5.)

muhammad-ali-e:
SonarCloud code smells addressed in `1d298de46`:
- **python:S1172 (×3)** — `resolve_transport()` no longer declares the unused `workflow_id`/`pipeline_id`/`organization_id` params; the inert PR-1 seam needs no inputs. PR 3 reintroduces them (keyed for Flipt) when it wires the evaluation, updating the two call sites then. Tests updated.
- **dict-vs-literal** — replaced the `dict(...)` constructor with a `{...}` literal in the worker context test.

Tests green (9 backend + 18 worker).

PR #2066, opened 2026-06-17
UN-3560 [FEAT] PG Queue 9 — run-worker.sh `-L celery` log alias

muhammad-ali-e:
## What & why

`./run-worker.sh -L pg-queue` already tails the PG-queue set's logs, but there was **no symmetric way to tail only the Celery set**. `-L` (no arg) tails **everything** — Celery **and** the PG-queue consumer/reaper — because `list_core_worker_dirs` includes the PG worker dirs. With both transports running side by side (the strangler-fig setup), the Celery logs get buried in PG-queue chatter.

This adds a **`-L celery`** log alias — the mirror of `-L pg-queue` — that tails every worker log **except** the PG-queue members (`pg_queue_consumer`, `pg_queue_reaper`).

Logs-only: the Celery set is still **run** via `all` (PG-queue is opt-in, never in `all`), so no run-path change is needed.

## Change
`workers/run-worker.sh` only:
- a `CELERY_SET="celery"` constant (mirrors `PG_QUEUE_SET`)
- a branch in `tail_logs()` that iterates core + pluggable dirs, skipping the two PG types
- usage text + examples

## Dev-test
With stub PG log files present:
- `-L` → **11** files (9 Celery + 2 PG)
- `-L celery` → **9** files (PG excluded) ✅
- `-L pg-queue` → **2** files (PG only) ✅

`bash -n run-worker.sh` clean.

## Notes
- Targets `feat/UN-3445-pg-queue-integration` (not `main`).
- Independent of the 9e seam PR (UN-3559 / #2062).
- Ticket: UN-3560.

### DIFF
--- a/workers/run-worker.sh
+++ b/workers/run-worker.sh
@@ -31,9 +31,20 @@ readonly PG_QUEUE_CONSUMER_TYPE="pg_queue_consumer"
 readonly PG_QUEUE_REAPER_TYPE="pg_queue_reaper"
 # Set alias that launches the whole PG-queue group (consumer + reaper) together.
 readonly PG_QUEUE_SET="pg-queue"
-# Log-tail alias for the Celery set: every worker EXCEPT the PG-queue members.
-# Mirrors PG_QUEUE_SET so the two transports' logs can be tailed separately
-# (-L celery vs -L pg-queue). Logs-only — the Celery set is *run* via 'all'.
+# The PG-queue member worker dirs, as a set — the single source of truth for
+# "which workers belong to the PG-queue transport". Keeps the `celery = all −
+# pg-queue` complement correct by construction if a third member is ever added
+# (callers test membership via ${PG_QUEUE_MEMBERS[$dir]:-} instead of
+# hand-rolling the consumer/reaper pair).
+declare -rA PG_QUEUE_MEMBERS=(
+    ["$PG_QUEUE_CONSUMER_TYPE"]=1
+    ["$PG_QUEUE_REAPER_TYPE"]=1
+)
+# Log-tail alias for the Celery transport: every worker EXCEPT the PG-queue
+# members — the *complement* of the 'pg-queue' tail alias, so the two
+# transports' logs can be tailed separately (-L celery vs -L pg-queue).
+# Tail-only — there is no 'celery' run alias; this set is started via 'all',
+# which by design omits the opt-in PG-queue workers.
 readonly CELERY_SET="celery"
 
 # Available workers
@@ -444,18 +455,15 @@ tail_logs() {
     local requested=$1  # may be empty → tail all
     local log_files=()
 
-    if [[ -z "$requested" || "$requested" == "all" ]]; then
+    if [[ -z "$requested" || "$requested" == "all" || "$requested" == "$CELERY_SET" ]]; then
+        # 'all' (and empty) tails every worker; 'celery' tails every worker
+        # EXCEPT the PG-queue members — the complement of the 'pg-queue' alias
+        # (handled in the else-branch) — so the two transports' logs can be
+        # tailed separately. Same resolve-and-append body for both, single-sourced.
         for d in $(list_core_worker_dirs) $(list_pluggable_worker_dirs); do
-            local f
-            f=$(resolve_log_file "$d")
-            [[ -n "$f" ]] && log_files+=("$f")
-        done
-    elif [[ "$requested" == "$CELERY_SET" ]]; then
-        # Celery set logs = every worker EXCEPT the PG-queue members. Mirror of
-        # the 'pg-queue' alias below, so the two transports' logs can be tailed
-        # separately when both sets run side by side.
-        for d in $(list_core_worker_dirs) $(list_pluggable_worker_dirs); do
-            [[ "$d" == "$PG_QUEUE_CONSUMER_TYPE" || "$d" == "$PG_QUEUE_REAPER_TYPE" ]] && continue
+            if [[ "$requested" == "$CELERY_SET" && -n "${PG_QUEUE_MEMBERS[$d]:-}" ]]; then
+                continue
+            fi
             local f
             f=$(resolve_log_file "$d")
             [[ -n "$f" ]] && log_files+=("$f")
--- a/workers/run-worker.sh
+++ b/workers/run-worker.sh
@@ -31,6 +31,10 @@ readonly PG_QUEUE_CONSUMER_TYPE="pg_queue_consumer"
 readonly PG_QUEUE_REAPER_TYPE="pg_queue_reaper"
 # Set alias that launches the whole PG-queue group (consumer + reaper) together.
 readonly PG_QUEUE_SET="pg-queue"
+# Log-tail alias for the Celery set: every worker EXCEPT the PG-queue members.
+# Mirrors PG_QUEUE_SET so the two transports' logs can be tailed separately
+# (-L celery vs -L pg-queue). Logs-only — the Celery set is *run* via 'all'.
+readonly CELERY_SET="celery"
 
 # Available workers
 declare -A WORKERS=(
@@ -163,7 +167,9 @@ OPTIONS:
     -r, --restart         Kill matching worker(s) then relaunch (with WORKER_TYPE,
                           restarts only that worker; without it, restarts all)
     -s, --status          Show status of running workers
-    -L, --logs [WORKER]   Live-tail worker log files (all if WORKER omitted)
+    -L, --logs [WORKER]   Live-tail worker log files (all if WORKER omitted).
+                          WORKER may also be a set: 'celery' (all but the
+                          PG-queue members) or 'pg-queue' (consumer + reaper).
     -C, --clear-logs      Delete worker .log files created by -d / 'all' runs
     -h, --help            Show this help message
 
@@ -199,6 +205,12 @@ EXAMPLES:
 
     # Live-tail all worker logs
     $0 -L
+
+    # Tail just the Celery set's logs (excludes the PG-queue workers)
+    $0 -L celery
+
+    # Tail just the PG-queue set's logs (consumer + reaper)
+    $0 -L pg-queue
 
     # Tail just one worker's log
     $0 -L general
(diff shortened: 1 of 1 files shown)


workers/run-worker.sh
```diff
@@ -31,6 +31,10 @@ readonly PG_QUEUE_CONSUMER_TYPE="pg_queue_consumer"
 readonly PG_QUEUE_REAPER_TYPE="pg_queue_reaper"
 # Set alias that launches the whole PG-queue group (consumer + reaper) together.
 readonly PG_QUEUE_SET="pg-queue"
+# Log-tail alias for the Celery set: every worker EXCEPT the PG-queue members.
+# Mirrors PG_QUEUE_SET so the two transports' logs can be tailed separately
```
muhammad-ali-e: **[Comment accuracy — P1]** "Mirrors PG_QUEUE_SET" overstates the relationship. The `celery` set is the *complement* of the pg-queue set (everything **minus** the two members), and it's implemented in a dedicated `elif` branch, whereas `pg-queue` is resolved through the `WORKERS` map and special-cased in the `else` path. They are complementary sets in different code paths, not mirror implementations.

Suggested rewrite:
```bash
# Log-tail alias for the Celery transport: every worker EXCEPT the two
# PG-queue members (pg_queue_consumer, pg_queue_reaper). It is the
# complement of the 'pg-queue' tail alias, so the two transports' logs can
# be tailed separately (-L celery vs -L pg-queue). Tail-only — there is no
# 'celery' run alias; this set is started via 'all', which by design omits
# the opt-in PG-queue workers (see OPTIN_WORKERS).
```
The `OPTIN_WORKERS` pointer also hardens the "run via 'all'" claim, which currently relies on an invariant enforced in two independent places (the opt-out map and the hardcoded `core_workers` string).

workers/run-worker.sh
```diff
@@ -438,6 +450,16 @@ tail_logs() {
             f=$(resolve_log_file "$d")
             [[ -n "$f" ]] && log_files+=("$f")
         done
+    elif [[ "$requested" == "$CELERY_SET" ]]; then
```
muhammad-ali-e: **[Simplification — P1]** This `celery` branch is a near-verbatim copy of the `all` branch just above (lines 448-452); the only delta is the one-line PG-member skip. Since "all" is just "celery without the exclusion," the two can collapse into a single branch:
```bash
if [[ -z "$requested" || "$requested" == "all" || "$requested" == "$CELERY_SET" ]]; then
    # 'all' tails every worker; 'celery' tails every worker EXCEPT the
    # PG-queue members, so the two transports' logs can be tailed separately.
    for d in $(list_core_worker_dirs) $(list_pluggable_worker_dirs); do
        if [[ "$requested" == "$CELERY_SET" ]] \
            && [[ "$d" == "$PG_QUEUE_CONSUMER_TYPE" || "$d" == "$PG_QUEUE_REAPER_TYPE" ]]; then
            continue
        fi
        local f
        f=$(resolve_log_file "$d")
        [[ -n "$f" ]] && log_files+=("$f")
    done
else
```
This single-sources the resolve-and-append body so future changes to log-path resolution touch one place instead of two. (Optional — the duplication is small and readable as-is.)

workers/run-worker.sh
```diff
@@ -438,6 +450,16 @@ tail_logs() {
...
+        # Celery set logs = every worker EXCEPT the PG-queue members. Mirror of
```
muhammad-ali-e: **[Comment accuracy — P1]** Same "Mirror of the 'pg-queue' alias below" issue as the constant block: `celery` enumerates "everything minus 2" while `pg-queue` enumerates "exactly those 2" — complementary, not mirrored. The directional reference ("below") will also rot if the branches are reordered. Suggested:
```bash
# Celery set = every worker EXCEPT the two PG-queue members; the complement
# of the 'pg-queue' tail alias (handled in the else-branch). Lets the two
# transports' logs be tailed separately when both sets run side by side.
```

workers/run-worker.sh
```diff
@@ -438,6 +450,16 @@ tail_logs() {
...
+        # the 'pg-queue' alias below, so the two transports' logs can be tailed
+        # separately when both sets run side by side.
+        for d in $(list_core_worker_dirs) $(list_pluggable_worker_dirs); do
+            [[ "$d" == "$PG_QUEUE_CONSUMER_TYPE" || "$d" == "$PG_QUEUE_REAPER_TYPE" ]] && continue
```
muhammad-ali-e: **[Modeling / maintainability — P2]** This `[[ "$d" == "$PG_QUEUE_CONSUMER_TYPE" || "$d" == "$PG_QUEUE_REAPER_TYPE" ]]` test is now the 4th+ hand-rolled enumeration of "the PG-queue members" in this file (also at lines 375, 473, 912, 929-930, 1060-1061; `OPTIN_WORKERS` at 119-122 lists the same two for a different purpose). The `celery = all − pg-queue` invariant lives implicitly in this `continue`, so adding a third PG member later would silently leak its log into `-L celery`.

Consider one source of truth near the other `readonly`s:
```bash
declare -A PG_QUEUE_MEMBERS=(["$PG_QUEUE_CONSUMER_TYPE"]=1 ["$PG_QUEUE_REAPER_TYPE"]=1)
```
then `[[ -n "${PG_QUEUE_MEMBERS[$d]:-}" ]] && continue` here, reused by the other sites. Makes the complement correct by construction as membership grows.

workers/run-worker.sh
```diff
@@ -438,6 +450,16 @@ tail_logs() {
...
+            local f
+            f=$(resolve_log_file "$d")
+            [[ -n "$f" ]] && log_files+=("$f")
+        done
```
muhammad-ali-e, line 470: **[Silent-failure / UX — P2, inherited from `all`/`pg-queue`]** When this branch resolves zero log files (e.g. only the PG-queue set is running, or nothing was started with `-d`), it falls through to the shared guard at line ~489 which prints "Workers must be started in detached mode..." and `exit 0`. For the `celery` set specifically that message misattributes the cause and gives no hint that the PG-queue members were excluded — and exit 0 is indistinguishable from "tailing successfully" for any script wrapping this. Consider a set-aware empty message ("No Celery-set log files found (PG-queue members excluded); start them with `-d all`, or use `-L pg-queue`"), and optionally a distinct non-zero exit so automation can detect "nothing matched." Best applied to all three set branches at once since they share the guard.

muhammad-ali-e (commented):
## Automated PR Review — `-L celery` log alias

Ran the PR Review Toolkit (Code Reviewer, Silent-Failure Hunter, Type-Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier) against the diff vs `feat/UN-3445-pg-queue-integration`.

**Verdict: no blocking issues.** Quoting, `set -e` handling, word-splitting, and the `local f` / `f=$(...)` split all follow this file's established conventions, and the empty-result case correctly falls through to the shared guard rather than `exec tail` with no args.

The inline comments below are quality/consistency improvements, ordered by value:

**P1 — worth fixing here**
1. Comment accuracy: the "Mirror" wording (constant block + inline) misdescribes the control flow — `celery` and `pg-queue` are *complementary* sets in *different* branches, not mirror implementations.
2. Duplication: the new `celery` branch is a near-verbatim copy of the `all` branch; they can collapse into one loop with an exclusion guard.

**P2 — hardening / consistency**
3. Modeling: the `{consumer, reaper}` membership test is now the 4th hand-rolled copy in this file — a single `PG_QUEUE_MEMBERS` source of truth would keep the `celery = all − pg-queue` invariant true by construction.
4. UX (inherited from siblings): `-L celery` with no logs exits 0 with a generic message that misattributes the cause for this set.

**Tests:** none added; none strictly required given this is logs-only tooling with no existing shell-test harness. If a regression guard is wanted cheaply, gate the final `exec tail -F` on a dry-run env var and add a ~30-line bash test asserting the celery set excludes both PG members and the empty set exits cleanly.

workers/run-worker.sh
muhammad-ali-e: Fixed in 4b675a14b — the constant comment now says the celery set is the **complement** of the pg-queue tail alias (every worker minus the two PG members), in a different code path, not a 'mirror'. Also added the 'started via all, which omits the opt-in PG workers' note.

workers/run-worker.sh
muhammad-ali-e: Done in 4b675a14b — collapsed the near-duplicate `all` and `celery` branches into a single loop (`-z || == all || == $CELERY_SET`) with a membership-guarded `continue`, single-sourcing the resolve-and-append body.

workers/run-worker.sh
muhammad-ali-e: Fixed in 4b675a14b — reworded to 'complement of the pg-queue alias' and dropped the directional 'below' reference (now branch-order-independent).

workers/run-worker.sh
muhammad-ali-e: Done in 4b675a14b — added a single `readonly -A PG_QUEUE_MEMBERS` source of truth near the other constants; the celery exclusion now tests `${PG_QUEUE_MEMBERS[$d]:-}` so the `celery = all − pg-queue` complement stays correct by construction if a third member is added. (Adopting it at the other pre-existing membership sites you listed is a clean follow-up, kept out of this focused PR.)

workers/run-worker.sh
muhammad-ali-e: Acknowledged — deferred as a follow-up. The set-aware empty-result message + non-zero exit is inherited from the shared zero-log guard and, as you note, is best applied to all three set aliases (all/celery/pg-queue) at once rather than celery-only here. Tracked separately.

muhammad-ali-e:
## Review addressed — commit `4b675a14b`

Thanks for the pass. Per-thread replies above; summary:

**Fixed:**
- **Comment accuracy (P1 ×2):** the `celery` set is now described as the **complement** of the `pg-queue` tail alias (all workers minus the two PG members), in a separate code path — not a "mirror". Dropped the directional "below".
- **Single source of truth (P2):** added `readonly -A PG_QUEUE_MEMBERS`; the celery exclusion tests membership via it, so `celery = all − pg-queue` stays correct by construction as membership grows.
- **Simplification (P1):** collapsed the duplicate `all`/`celery` `tail_logs` branches into one loop with a membership-guarded skip.

**Deferred (P2, inherited):** set-aware empty-result message + non-zero exit for the shared zero-log guard — spans all three set aliases (`all`/`celery`/`pg-queue`), so best as one follow-up rather than celery-only here.

Re-verified: `-L` = 11, `-L celery` = 9 (PG excluded), `-L pg-queue` = 2; `bash -n` clean.

PR #2067, opened 2026-06-17
UN-3561 [FEAT] PG Queue 9e PR 2a — live-PG-pipeline inert foundation (dispatch backend override + barrier decrement core)

muhammad-ali-e:
## What

First slice of **9e PR 2** (the live PG execution pipeline), split into **2a → 2b → 2c**. This is **2a, the inert foundation** — the two seams the live switch (2c) consumes, each with **zero behaviour change on the default path**.

### 1. `dispatch(backend=...)` — per-call transport override
`dispatch()` gains an optional `backend: QueueBackend | None = None`:
- `None` (the default, and **every call site today**) → transport is the env allow-list decision via `select_backend()` — **byte-identical** to today.
- Set → it **wins over the allow-list**. This is the seam the execution-level PG pipeline (2c) uses to route a whole execution's header/callback dispatches onto PG **without** opting their task *names* into `WORKER_PG_QUEUE_ENABLED_TASKS`. The allow-list is for *leaf* tasks; the coupled pipeline's migration unit is the *execution* (carried in the payload — see `routing.py`).

### 2. Extract `_barrier_pg_decrement(...)` core
The decrement logic moves out of the `@worker_task barrier_pg_decr_and_check` into a plain `_barrier_pg_decrement(...)` function; the `@worker_task` becomes a thin **delegator**. The plain core is what 2c calls **in-body** on the PG-consumed path — a PG-consumed task fires no Celery `.link`, so the decrement must run in-body (fire-and-forget self-chaining). Behaviour for the existing Celery `.link` path is unchanged (the wrapper forwards verbatim).

## Why it's inert

- Default barrier backend is `chord` → **Celery executions never import the `PgBarrier` module**, so the extraction is invisible to them.
- **No call site passes `backend=`** → the Celery dispatch path is byte-identical.
- Net behaviour change: **NONE**. Transport threading into the fan-out fns + the live `transport=="pg_queue"` switch land in **2c** (threading an unused param earlier just trips S1172); per-batch idempotency lands in **2b**.

## Tests

- **dispatch backend-override** (3): PG-forces-PG-without-allow-list, Celery-forces-Celery-despite-allow-list, default-None-preserves-allow-list.
- **decrement-core extraction** (3): core is a plain callable (not a Celery task), runs the decrement in-body against real PG, and the `@worker_task` delegates verbatim.
- Full `test_pg_barrier` (33) + `test_dispatch_pg` (10) + barrier/routing (53) suites green; `ruff` clean.
- Worker-app bootstrap verified clean under `WORKER_BARRIER_BACKEND=pg`: both barrier tasks registered under canonical names, `get_barrier()` → `PgBarrier`.

## Base

### DIFF
--- a/workers/queue_backend/dispatch.py
+++ b/workers/queue_backend/dispatch.py
@@ -1,11 +1,12 @@
 """Transport-agnostic task dispatch.
 
-Routes each task to its transport via :func:`select_backend`:
+Routes each task to its transport via :func:`resolve_backend` (which applies a
+per-call ``backend`` override, else defers to :func:`select_backend`):
 
 - **Celery** (default) — a thin pass-through to ``current_app.send_task``.
-- **PG Queue** — when a task is opted into ``WORKER_PG_QUEUE_ENABLED_TASKS``,
-  the task is serialised and enqueued to ``pg_queue_message`` (9b); the PG
-  consumer (9c) drains and runs it.
+- **PG Queue** — when a task is opted into ``WORKER_PG_QUEUE_ENABLED_TASKS``
+  (or pinned via a ``backend=`` override), the task is serialised and enqueued
+  to ``pg_queue_message`` (9b); the PG consumer (9c) drains and runs it.
 
 The default (empty allow-list) routes everything to Celery, so dispatch is
 unchanged unless an operator explicitly opts a task in.
@@ -130,6 +131,12 @@ def _enqueue_pg(
         # true regardless of send outcome, so a first-dispatch failure
         # (DB down / unmigrated) must not suppress the one announcement.
         # INFO so it survives a default log config; once-per-task bounds it.
+        # NOTE: keyed on task name only, not on *why* it routed to PG. If a name
+        # is first PG-routed via a ``backend=`` override and the operator later
+        # adds it to the allow-list, the allow-list cutover won't re-announce
+        # (already in the set). Benign given the usage split (override = pipeline
+        # headers, allow-list = leaf tasks, so no overlap expected); the
+        # allow-list config itself is still announced by _log_allow_list_once.
         _pg_routing_logged.add(task_name)
         logger.info(
             "PG-queue: routing task=%r to Postgres (queue=%r). Requires the "
--- a/workers/queue_backend/dispatch.py
+++ b/workers/queue_backend/dispatch.py
@@ -30,7 +30,7 @@ from celery import current_app
 from .fairness import DEFAULT_PRIORITY, FairnessKey
 from .handle import DispatchHandle
 from .pg_queue import PgQueueClient, to_payload
-from .routing import QueueBackend, select_backend
+from .routing import QueueBackend, resolve_backend
 
 logger = logging.getLogger(__name__)
 
@@ -85,19 +85,21 @@ def dispatch(
     path / serialised into the message on the PG path. Pass ``None`` for
     non-workflow worker tasks.
 
-    ``backend`` is a per-call transport override. When ``None`` (the default,
-    and every call site today) the transport is the env allow-list decision
-    via :func:`select_backend` — behaviour is unchanged. When set, it wins
-    over the allow-list: this is the seam the execution-level PG pipeline (9e
-    PR 2c) uses to route a whole execution's header/callback dispatches onto
-    PG without opting their task *names* into ``WORKER_PG_QUEUE_ENABLED_TASKS``
-    (the allow-list is for leaf tasks; the coupled pipeline's migration unit is
-    the execution, carried in the payload — see ``routing.py``). The override
-    only forces the *transport*; it does not bypass ``_enqueue_pg``'s no-silent-
-    fallback contract.
+    ``backend`` is a per-call transport override (see
+    :func:`~queue_backend.routing.resolve_backend` for the precedence). When
+    ``None`` (the default, and every call site today) the transport is the env
+    allow-list decision via ``select_backend`` — behaviour is unchanged. When
+    set, it wins over the allow-list: this is the seam the execution-level PG
+    pipeline (9e PR 2c) uses to route a whole execution's header/callback
+    dispatches onto PG without opting their task *names* into
+    ``WORKER_PG_QUEUE_ENABLED_TASKS``. (The allow-list is for *leaf* tasks; the
+    coupled pipeline's migration unit is the whole execution — its transport is
+    resolved once at creation and travels on the execution's task kwargs onto
+    ``WorkflowContextData.transport``, see ``queue_backend/pg_queue/
+    9e-design.md``.) The override only forces the *transport*; it does not
+    bypass ``_enqueue_pg``'s no-silent-fallback contract.
     """
-    resolved = backend if backend is not None else select_backend(task_name)
-    if resolved is QueueBackend.PG:
+    if resolve_backend(task_name, backend) is QueueBackend.PG:
         return _enqueue_pg(task_name, args, kwargs, queue, fairness)
 
     headers = fairness.as_header() if fairness is not None else None
--- a/workers/queue_backend/dispatch.py
+++ b/workers/queue_backend/dispatch.py
@@ -77,14 +77,27 @@ def dispatch(
     kwargs: Mapping[str, Any] | None = None,
     queue: str | None = None,
     fairness: FairnessKey | None = None,
+    backend: QueueBackend | None = None,
 ) -> DispatchHandle:
     """Enqueue a task by name onto its selected transport.
 
     ``fairness`` is attached as the ``x-fairness-key`` header on the Celery
     path / serialised into the message on the PG path. Pass ``None`` for
     non-workflow worker tasks.
+
+    ``backend`` is a per-call transport override. When ``None`` (the default,
+    and every call site today) the transport is the env allow-list decision
+    via :func:`select_backend` — behaviour is unchanged. When set, it wins
+    over the allow-list: this is the seam the execution-level PG pipeline (9e
+    PR 2c) uses to route a whole execution's header/callback dispatches onto
+    PG without opting their task *names* into ``WORKER_PG_QUEUE_ENABLED_TASKS``
+    (the allow-list is for leaf tasks; the coupled pipeline's migration unit is
+    the execution, carried in the payload — see ``routing.py``). The override
+    only forces the *transport*; it does not bypass ``_enqueue_pg``'s no-silent-
+    fallback contract.
     """
-    if select_backend(task_name) is QueueBackend.PG:
+    resolved = backend if backend is not None else select_backend(task_name)
+    if resolved is QueueBackend.PG:
         return _enqueue_pg(task_name, args, kwargs, queue, fairness)
 
     headers = fairness.as_header() if fairness is not None else None
--- a/workers/queue_backend/pg_barrier.py
+++ b/workers/queue_backend/pg_barrier.py
@@ -65,6 +65,7 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
 
 import psycopg2
+import psycopg2.extensions
 
 from .barrier import CallbackDescriptor, barrier_ttl_seconds
 from .decorator import worker_task
@@ -261,16 +262,37 @@ def _barrier_pg_decrement(
 
     - the Celery ``link`` callback :func:`barrier_pg_decr_and_check` (today's
       only caller, behaviour unchanged); and
-    - the 9e PR 2c PG-consumed pipeline path, which calls this directly in the
-      header task's body — a PG-consumed task fires no ``.link``, so the
-      decrement must run in-body (fire-and-forget self-chaining).
+    - the 9e PR 2c PG-consumed pipeline path, which will call this directly in
+      the header task's body — a PG-consumed task fires no ``.link``, so the
+      decrement must run in-body.
 
-    Callers MUST run each decrement in its own committed transaction (one call
-    = one ``_cursor()`` txn here) and MUST NOT retry it: a replay re-runs the
-    decrement and corrupts the count. The Celery wrapper enforces this with
+    Callers MUST run each decrement in its own committed transaction (the
+    decrement ``UPDATE`` runs in its own ``_cursor()`` txn) and MUST NOT retry
+    it: a replay re-runs the decrement and corrupts the count. This is enforced
+    loudly, not just in prose — entry raises if the shared connection is already
+    mid-transaction (see the guard below). The Celery wrapper additionally pins
     ``max_retries=0``; an orphaned barrier is bounded by ``expires_at`` instead.
     """
     from celery import current_app
+
+    # Enforce the "own committed transaction" contract loudly. A caller that
+    # invokes this inside an already-open transaction on the shared thread-local
+    # connection would hold the row lock across the call and let the outer txn's
+    # commit boundary replay the decrement — corrupting ``remaining`` and
+    # breaking the exactly-one-sees-zero serialisation, surfacing only as a
+    # barrier hung until ``expires_at`` (~6h). The Celery ``.link`` path always
+    # enters idle (``_cursor`` commits after every use); the 2c in-body caller
+    # must too. (Tests inject an autocommit connection → always idle → no trip.)
+    if (
+        _get_conn().get_transaction_status()
+        != psycopg2.extensions.TRANSACTION_STATUS_IDLE
+    ):
+        raise RuntimeError(
+            f"[exec:{execution_id}] _barrier_pg_decrement entered with an open "
+            f"transaction on the shared connection — each decrement MUST run in "
+            f"its own committed transaction (see this function's docstring). "
+            f"Refusing to proceed to avoid corrupting the barrier counter."
+        )
 
     try:
         # No default=str — a non-JSON-safe leaf must fail loudly here (it would
--- a/workers/queue_backend/pg_barrier.py
+++ b/workers/queue_backend/pg_barrier.py
@@ -246,20 +246,29 @@ class PgBarrier:
             raise
 
 
-@worker_task(name="barrier_pg_decr_and_check", max_retries=0)
-def barrier_pg_decr_and_check(
+def _barrier_pg_decrement(
     result: Any,
     *,
     execution_id: str,
     callback_descriptor: CallbackDescriptor,
 ) -> dict[str, Any]:
-    """Per-task ``link`` callback for :class:`PgBarrier`.
+    """Atomic barrier decrement + last-task callback fire (substrate core).
 
-    Atomically appends this task's result and decrements ``remaining``. The
-    single task that drives ``remaining`` to 0 dispatches the aggregating
-    callback (then deletes the row). ``max_retries=0`` — a Celery retry would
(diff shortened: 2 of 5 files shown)


workers/queue_backend/pg_barrier.py
```diff
@@ -246,20 +246,29 @@ def enqueue(
...
+    aggregating callback (then deletes the row). This is the plain, in-body-
+    callable core shared by two entry points:
+
+    - the Celery ``link`` callback :func:`barrier_pg_decr_and_check` (today's
+      only caller, behaviour unchanged); and
+    - the 9e PR 2c PG-consumed pipeline path, which calls this directly in the
+      header task's body — a PG-consumed task fires no ``.link``, so the
+      decrement must run in-body (fire-and-forget self-chaining).
+
+    Callers MUST run each decrement in its own committed transaction (one call
```
muhammad-ali-e: **[MEDIUM — silent failure risk] + [LOW — comment accuracy]** on this contract block.

**(a) Enforcement gap:** the "MUST run in its own committed transaction / MUST NOT retry" contract is prose-only. `max_retries=0` guards the *Celery* wrapper, but the 2c PG-consumed path will call `_barrier_pg_decrement` directly in a task body. If that future caller wraps it in a retry loop or invokes it inside an already-open outer transaction, the decrement replays or the row lock is held across calls — corrupting `remaining` / breaking the exactly-one-sees-zero serialisation. The failure is silent: a hung barrier reclaimed only at the ~6h `expires_at`. Suggest either a cheap guard test in `TestDecrementCoreExtraction` (assert each decrement commits independently) or, stronger, have the core reject entry when `conn.status == STATUS_IN_TRANSACTION` (raise with the contract text) so a violation fails loudly at the call site.

**(b) Inaccuracy:** the parenthetical "(one call = one `_cursor()` txn here)" is wrong — the last-task path, the negative-`remaining` cleanup, and the `DataError` teardown each open a *second* `_cursor()` txn via `_delete_barrier`. Scope the parenthetical to the decrement `UPDATE`, or drop it and keep the MUST sentence.

workers/queue_backend/pg_barrier.py
```diff
@@ -246,20 +246,29 @@ def enqueue(
...
+    Appends this task's result and decrements ``remaining`` in one atomic
+    statement; the single caller that drives ``remaining`` to 0 dispatches the
+    aggregating callback (then deletes the row). This is the plain, in-body-
+    callable core shared by two entry points:
+
+    - the Celery ``link`` callback :func:`barrier_pg_decr_and_check` (today's
+      only caller, behaviour unchanged); and
+    - the 9e PR 2c PG-consumed pipeline path, which calls this directly in the
+      header task's body — a PG-consumed task fires no ``.link``, so the
+      decrement must run in-body (fire-and-forget self-chaining).
```
muhammad-ali-e: **[LOW — comment rot]** "(fire-and-forget self-chaining)" describes a 2c mechanism that doesn't exist yet, in compressed jargon with no anchor in present code. It's the most rot-prone phrase in the PR: if 2c implements the in-body decrement differently (e.g. consumer-loop re-enqueue rather than self-chaining), this becomes a confidently-wrong description of shipped code. The preceding clause "the decrement must run in-body" already carries the load-bearing point — drop the parenthetical or soften to clearly-future framing.

workers/queue_backend/pg_barrier.py
```diff
@@ -246,20 +246,29 @@ def enqueue(
             raise
 
 
-@worker_task(name="barrier_pg_decr_and_check", max_retries=0)
-def barrier_pg_decr_and_check(
+def _barrier_pg_decrement(
```
muhammad-ali-e, line 250: **[LOW — type design]** The return is `dict[str, Any]` here and on the wrapper, but it's really a closed discriminated union: `status ∈ {pending, complete, abandoned}` with status-dependent fields (`remaining` vs `callback_task_id`+`aggregated_count`). The PR's own `CallbackDescriptor` is a `TypedDict` *precisely* to catch a key typo across the Celery wire — the return crosses the same boundary but stays untyped. Consider a `BarrierDecrementResult(TypedDict)` with a `Literal` status, narrowing both annotations together. Low priority while nothing branches on the result, but 2c's in-body caller likely will.

workers/queue_backend/dispatch.py
```diff
@@ -77,14 +77,27 @@ def dispatch(
...
     non-workflow worker tasks.
+
+    ``backend`` is a per-call transport override. When ``None`` (the default,
+    and every call site today) the transport is the env allow-list decision
+    via :func:`select_backend` — behaviour is unchanged. When set, it wins
+    over the allow-list: this is the seam the execution-level PG pipeline (9e
+    PR 2c) uses to route a whole execution's header/callback dispatches onto
+    PG without opting their task *names* into ``WORKER_PG_QUEUE_ENABLED_TASKS``
+    (the allow-list is for leaf tasks; the coupled pipeline's migration unit is
+    the execution, carried in the payload — see ``routing.py``). The override
```
muhammad-ali-e: **[MEDIUM — comment accuracy / cross-ref]** Two issues with "the migration unit is the execution, carried in the payload — see `routing.py`":

1. **Term overload:** "payload" has a specific, different meaning in this very file — `payload = to_payload(...)` (line 138) is the PG wire message, which does *not* carry the execution-transport marker.
2. **Cross-ref mismatch:** `routing.py` (the thing you're told to "see") says the transport is carried in `ExecutionContext` "(like the fairness key)", not "the payload".

Align the noun with the referenced source, e.g. "…carried in `ExecutionContext` — see `routing.py`."

workers/queue_backend/dispatch.py
```diff
@@ -77,14 +77,27 @@ def dispatch(
     kwargs: Mapping[str, Any] | None = None,
     queue: str | None = None,
     fairness: FairnessKey | None = None,
+    backend: QueueBackend | None = None,
```
muhammad-ali-e, line 81: **[LOW — type design]** `backend: QueueBackend | None` overloads `None` as a third semantic state ("defer to the allow-list") grafted onto a deliberately two-valued enum — the signature reads as "optional backend" but the contract is tri-state PG|CELERY|AUTO. This sits awkwardly next to the `QueueBackend` `StrEnum` whose docstring is militant about values not being misread. Lowest-churn improvement: extract the precedence into a named `resolve_backend(task_name, override)` helper so the `None`-means-auto rule lives in one self-documenting, reusable place (2c will need the same precedence). Acceptable as-is for an inert foundation; flagging for when 2c threads this through more call sites.

workers/tests/test_dispatch_pg.py
```diff
@@ -111,6 +112,62 @@ def test_celery_dispatch_unaffected(self):
...
+class TestDispatchBackendOverride:
+    """The per-call ``backend=`` override (9e PR 2a inert foundation).
+
+    When set it wins over the ``WORKER_PG_QUEUE_ENABLED_TASKS`` allow-list, so
+    the execution-level PG pipeline can route a whole execution's headers /
+    callback onto PG without opting their task *names* in. ``None`` (default)
+    preserves the allow-list decision exactly — every call site today.
+    """
+
+    def test_override_pg_forces_pg_without_allow_list(self, monkeypatch):
```
muhammad-ali-e, line 124: **[MEDIUM — test gap]** This test dispatches with **no `fairness`** and asserts only `payload["task_name"]`. But the override's stated purpose (dispatch.py:88-97) is to route a whole execution's dispatches onto PG — i.e. exactly the dispatches that *carry* a `FairnessKey`. The org/priority plumbing (dispatch.py:145-152) *should* work identically on the override path, but no test proves it: a regression dropping `fairness` when an explicit backend is resolved would stay green. `TestDispatchPriorityWiring` covers priority only via the allow-list path, never via `backend=`. Suggest adding a test that dispatches `backend=QueueBackend.PG` with a `FairnessKey(pipeline_priority=8, org_id="o")` and asserts `captured["priority"] == 8` / `captured["org_id"] == "o"` reach the row.

workers/tests/test_pg_barrier.py
```diff
@@ -325,6 +326,46 @@ def test_registered_under_canonical_name(self):
...
+        _seed(barrier_db, "exec-core", 3)
+        out = _barrier_pg_decrement(
+            {"f": 1}, execution_id="exec-core", callback_descriptor=_CALLBACK
+        )
+        assert out["status"] == "pending"
+        remaining, results = _row(barrier_db, "exec-core")
+        assert remaining == 2
+        assert results == [{"f": 1}]
+
+    def test_worker_task_delegates_to_core(self):
```
muhammad-ali-e: **[MEDIUM — test quality]** This is a mock-tautology: it patches `_barrier_pg_decrement` with a sentinel and asserts the wrapper forwards args + returns it. It verifies the wrapper *calls a function*, but cannot catch the drift it claims to guard ("both paths must share one implementation") — because the real core is mocked away. If someone renamed the core's `callback_descriptor` param, this still passes. Stronger: run the wrapper against a real seeded row and assert the same observable decrement as a direct core call (mirror `test_core_runs_the_decrement_in_body`) — that's what actually proves no drift. Keep this one too if you want the keyword-forwarding contract pinned explicitly.

muhammad-ali-e (commented):

Reviewed by Code Reviewer, Silent Failure Hunter, Type Design Analyzer, PR Test Analyzer, Comment Analyzer, and Code Simplifier against base `feat/UN-3445-pg-queue-integration`.

**Verdict: no blocking issues.** The `_barrier_pg_decrement` extraction is byte-for-byte faithful (verified via `git show`), the Celery wrapper delegates verbatim with `max_retries=0` preserved, and the `backend` override is correctly inert with no live caller. Code Reviewer and Code Simplifier found nothing to change.

8 inline findings below — all MEDIUM/LOW (nice-to-have before the 2c follow-up lands a real caller). Priority order:

1. **MEDIUM** — `_barrier_pg_decrement`'s "own committed transaction / no retry" contract is doc-only; a future in-body caller can silently corrupt the count. (pg_barrier.py:268)
2. **MEDIUM** — override + fairness/priority interaction is untested, yet that's the override's whole reason to exist. (test_dispatch_pg.py:124)
3. **MEDIUM** — delegation test is a mock-tautology; strengthen to a DB-backed equivalence check. (test_pg_barrier.py:353)
4. **MEDIUM** — docstring cross-ref mismatch + term overload: "carried in the payload" vs `routing.py`'s `ExecutionContext`. (dispatch.py:95)
5. **LOW** — comment inaccuracy: "one call = one `_cursor()` txn" (actually up to two). (pg_barrier.py:268)
6. **LOW** — type: `backend: QueueBackend | None` overloads `None` as a third AUTO state. (dispatch.py:80)
7. **LOW** — type: untyped `dict[str, Any]` return vs the typed `CallbackDescriptor` input. (pg_barrier.py:249)
8. **LOW** — rot-prone comment describing a not-yet-existent 2c mechanism. (pg_barrier.py:266)

workers/tests/test_pg_barrier.py
```diff
@@ -325,6 +326,46 @@ def test_registered_under_canonical_name(self):
         assert barrier_pg_decr_and_check.name == "barrier_pg_decr_and_check"
 
 
+class TestDecrementCoreExtraction:
+    """The decrement logic lives in a plain ``_barrier_pg_decrement`` core so the
+    9e PR 2c PG-consumed path can call it in-body (a PG-consumed task fires no
+    ``.link``). The Celery ``@worker_task`` is a thin delegator; both paths must
+    share one implementation (no drift). (Inert in 2a — no PG caller yet.)
```
muhammad-ali-e, line 333: **[positive]** `test_core_is_a_plain_callable_not_a_celery_task` is a sharp, cheap guard against a real footgun — if the core accidentally got `@worker_task`, the planned 2c in-body caller would `.delay()`-dispatch instead of executing, silently breaking the barrier. Asserting absence of `.name`/`.delay` is exactly the right behavioral proxy. No change requested.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Both fixed in \`e84df69c4\`. (a) The contract is now enforced **loudly, not in prose**: \`_barrier_pg_decrement\` raises at entry if the shared connection is mid-transaction (\`get_transaction_status() != TRANSACTION_STATUS_IDLE\`). Verified safe for every existing path — the Celery \`.link\` path always enters idle (\`_cursor\` commits after each use) and the tests inject autocommit connections (always idle), so it only trips on the 2c misuse you describe. Added \`test_open_transaction_on_shared_conn_raises\`. (b) Corrected the parenthetical to scope it to the decrement \`UPDATE\` only ("the decrement \`UPDATE\` runs in its own \`_cursor()\` txn"), since the last-task/negative/DataError paths do open a second \`_cursor()\` via \`_delete_barrier\`.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in \`e84df69c4\` — dropped "(fire-and-forget self-chaining)". The load-bearing clause "the decrement must run in-body" stays; also softened "calls this" → "will call this" so it reads as clearly-future.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Deferring to 2c (agreeing with your "low priority while nothing branches on the result, but 2c's in-body caller likely will"). The union's exact shape firms up once the 2c caller actually branches on \`status\`; defining \`BarrierDecrementResult(TypedDict)\` now risks churning the annotation twice. Tracked for 2c.

workers/queue_backend/dispatch.py
muhammad-ali-e: Fixed in \`e84df69c4\`, though not with the literal suggestion — "ExecutionContext … see routing.py" would itself be inaccurate here: the payload-carry design (9e-design.md) carries the transport on the dispatched task's kwargs landing on \`WorkflowContextData.transport\`, not a separate \`ExecutionContext\` object (routing.py's docstring predates that decision). So I reworded to avoid the \`payload\` term collision *and* point at the live carrier: "…travels on the execution's task kwargs onto \`WorkflowContextData.transport\`, see \`queue_backend/pg_queue/9e-design.md\`."

workers/queue_backend/dispatch.py
muhammad-ali-e: Done in \`e84df69c4\` — extracted \`resolve_backend(task_name, override)\` into routing.py (next to \`select_backend\`), so the override-wins-else-allow-list precedence lives in one self-documenting place; \`dispatch()\` now calls it instead of inlining the \`None\`-means-auto rule. 2c reuses the same helper across the new call sites.

workers/tests/test_dispatch_pg.py
muhammad-ali-e: Added \`test_override_path_carries_fairness_to_row\` in \`e84df69c4\` — dispatches \`backend=QueueBackend.PG\` with \`FairnessKey(org_id="o", pipeline_priority=8)\` and asserts \`captured["priority"] == 8\` / \`captured["org_id"] == "o"\` reach the row. Closes exactly the regression window you flagged (fairness dropped on the override path staying green).

workers/tests/test_pg_barrier.py
muhammad-ali-e: Strengthened in \`e84df69c4\`. Added \`test_worker_task_produces_same_decrement_as_core\` — runs the wrapper against a real seeded row and asserts the same observable decrement as a direct core call (core NOT mocked), so a renamed/reordered core param now fails. Kept the mocked one (renamed \`test_worker_task_forwards_kwargs_verbatim\`) as the explicit keyword-forwarding pin, per your "keep this one too".

workers/tests/test_pg_barrier.py
muhammad-ali-e: Thanks — agreed, that footgun (accidental \`@worker_task\` on the core → 2c \`.delay()\`-dispatches instead of running) is exactly what it guards. Kept as-is.

muhammad-ali-e:
## Review round addressed — `e84df69c4`

Thanks @muhammad-ali-e. All 8 threads handled — 6 fixed, 1 deferred (reviewer-aligned), 1 positive/noted.

**Fixed**
- **[MED] in-transaction enforcement** (pg_barrier:268a) — the "own committed transaction" contract is now enforced *loudly*: `_barrier_pg_decrement` raises at entry if the shared connection is mid-transaction. Safe for every existing path (Celery `.link` enters idle; tests use autocommit conns) — trips only on the 2c misuse. + `test_open_transaction_on_shared_conn_raises`.
- **[LOW] comment accuracy** (pg_barrier:268b) — scoped the "one `_cursor()` txn" parenthetical to the decrement `UPDATE` (the delete paths open a 2nd).
- **[LOW] comment rot** (pg_barrier:266) — dropped "(fire-and-forget self-chaining)"; softened to clearly-future framing.
- **[MED] comment accuracy** (dispatch:95) — reworded to avoid the `payload` term collision + the stale `routing.py` cross-ref; points at the live carrier `WorkflowContextData.transport` + `9e-design.md`. (Note: the literal "ExecutionContext" suggestion would itself be inaccurate under the payload-carry design — see the thread.)
- **[LOW] type design** (dispatch:80) — extracted `resolve_backend(task_name, override)` into `routing.py`; the `None`-means-auto precedence now lives in one reusable place (2c shares it).
- **[MED] test gap** (test_dispatch:124) — `test_override_path_carries_fairness_to_row`: asserts org/priority reach the row on the `backend=` path.
- **[MED] test quality** (test_pg_barrier:353) — added a real-row wrapper test that catches core-param drift (core not mocked); kept the mocked test as the explicit keyword-forwarding pin.

**Deferred (your call, LOW)**
- `BarrierDecrementResult(TypedDict)` union (pg_barrier:250) → 2c, where the in-body caller actually branches on `status` and the shape firms up.

**Verification**: `test_pg_barrier` + `test_dispatch_pg` + `test_routing` + `test_barrier_backend_selection` = 76 green; `ruff` clean; worker-app bootstrap clean under `WORKER_BARRIER_BACKEND=pg` (both barrier tasks registered, `get_barrier()` → `PgBarrier`, `resolve_backend` precedence verified). Still inert — net behaviour change on the default path: NONE.

muhammad-ali-e:
## @greptile-apps round addressed — `345933ee1`

Thanks for the 4/5 + the flowchart. All 3 "Comments Outside Diff" (P2) fixed — each was staleness introduced by the `resolve_backend` extraction in the previous review round, so good catches:

1. **dispatch.py docstring (L3)** — now reads "Routes each task to its transport via `resolve_backend` (which applies a per-call `backend` override, else defers to `select_backend`)", and the PG bullet mentions the `backend=` override path.
2. **routing.py "Scaffold posture" (L36-40)** — the seam sentence now names `resolve_backend()` (wrapping the `select_backend()` allow-list with the per-call override) as where real PG dispatch lands.
3. **`_pg_routing_logged` suppression (L127-138)** — added the call-out you suggested: the set is keyed on task name only, so an override-then-allow-list cutover won't re-announce. Benign given the usage split (override = pipeline headers, allow-list = leaf tasks → no overlap expected), and the allow-list config itself is still announced by `_log_allow_list_once`.

No logic changed (doc/comment only). 30 dispatch+routing tests green, ruff clean.

PR #2068, opened 2026-06-17
UN-3562 [FEAT] PG Queue 9e PR 2b — per-batch idempotency primitive (pg_batch_dedup + claim_batch / clear_execution_batches)

muhammad-ali-e:
## What

Second slice of **9e PR 2** (the live PG execution pipeline), after **2a** (#2067). This is **2b, the inert idempotency primitive** — the durable per-batch dedup marker that PR **2c** will wire into the at-least-once PG path.

## Why

The PG queue is **at-least-once**: a `process_file_batch` task can be redelivered after a crash-before-ack, which would re-run the batch and **double-decrement the barrier** (`process_file_batch` is non-idempotent — `max_retries=0`). A recon of the existing per-file protection found it **only partial**:
- the Redis per-file destination lock is **released after the write** (not a durable token, free again on redelivery);
- `WorkflowFileExecution` COMPLETED skips tool **re-execution**, but the destination write may still be reached with the prior result;
- FileHistory (`use_file_history`) is **cross-execution only** — doesn't cover same-execution crash-redelivery.

So a durable **per-batch** gate is needed. The existing per-file status remains the backstop for the partial-crash re-processing case (a re-run batch whose marker wasn't written yet skips already-done files).

## Scope (2b)

- **Backend** — `PgBatchDedup` model + migration **0006**: table `pg_batch_dedup` with a `UniqueConstraint(execution_id, batch_index)` (the `ON CONFLICT` target; its `execution_id`-leading index also serves the cleanup `DELETE`). Django-managed, extension-free, same posture as the sibling `pg_queue` models.
- **Workers** (`queue_backend/pg_barrier.py`, reusing the barrier's thread-local `_cursor()` → **one PG connection per worker child**):
  - `claim_batch(execution_id, batch_index) -> bool` — atomic `INSERT … ON CONFLICT DO NOTHING RETURNING`; `True` = first delivery (proceed + decrement), `False` = redelivery (skip, so the barrier decrements exactly once per batch).
  - `clear_execution_batches(execution_id) -> int` — barrier-teardown cleanup; the reaper's barrier-orphan sweep is the backstop for executions that never finalise.

## Not in 2b

No call-site wiring. `claim_batch` (at batch start) + `clear_execution_batches` (at barrier finalise) + `batch_index` threading + the transport switch all land in **2c**. (Same "don't thread unused params early" lesson that kept 2a clean.) **Net behaviour change: NONE.**

## Tests / dev-test

- **+8 real-PG tests**: first-claim · redelivery-rejected · distinct-batch · distinct-exec · **concurrent-exactly-one-winner** · clear-only-target · clear-empty-returns-0 · reclaim-after-clear.
- Migration 0006 generated + **applied to the dev DB**, `makemigrations --check` → no drift.
- Worker-app bootstrap clean under `WORKER_BARRIER_BACKEND=pg`; `ruff` clean.

## Base


backend/pg_queue/models.py
```diff
@@ -114,6 +114,51 @@ class Meta:
...
+    ``(execution_id, batch_index)``.
+
+    The worker claims a batch by inserting its row atomically
+    (``INSERT … ON CONFLICT DO NOTHING RETURNING``): a fresh insert means
+    "first delivery — proceed, and this is the attempt that decrements the
+    barrier"; a conflict means "already claimed — a redelivery → skip". Because
+    the claim and the barrier decrement run together, the barrier is decremented
+    exactly once per batch even though the queue may deliver the task more than
+    once. Rows for an execution are cleared when its barrier finalises
+    (``remaining`` → 0); the reaper's barrier-orphan sweep is the backstop for
```
muhammad-ali-e: **P1 · Inaccurate docstring (verified) — should fix.** This says "the reaper's barrier-orphan sweep is the backstop" for `pg_batch_dedup` markers of executions that never finalise. But the reaper's only recovery job, `sweep_expired_barriers` (`workers/queue_backend/pg_queue/reaper.py:131`), runs `DELETE FROM pg_barrier_state WHERE expires_at < now()` — it touches **only** `pg_barrier_state`. There is no FK / `ON DELETE CASCADE` linking the two tables, so when the reaper reclaims an orphaned barrier the matching dedup markers are left behind permanently. The described backstop does not exist today.

Suggest softening to reflect reality, e.g.: *"Rows are cleared on barrier finalise (`remaining` → 0). Executions that never finalise currently leak their dedup markers; the reaper sweeps the orphaned `pg_barrier_state` row but does not yet reclaim `pg_batch_dedup` — a dedup-orphan sweep is intended future work."* (Same claim is duplicated in `clear_execution_batches`, flagged separately.)

muhammad-ali-e (commented):

Reviewed the per-batch idempotency primitive across 4 files using six specialized agents (code review, silent-failure, type design, test coverage, comments, simplifier). The code is well-scoped, SQL-safe (parameterized), and closely mirrors its `PgBarrierState` / `_delete_barrier` / `test_pg_barrier.py` siblings. No correctness defects in the production helpers. Findings below are prioritized; the one verified-inaccurate docstring (P1) is the only thing I'd treat as a should-fix before merge.

Note: a `black`/line-length finding was raised and **dropped after verification** — pre-commit uses `ruff-format` (line-length 90), the line is exactly 90 chars and passes; the `[tool.black]` entry in `workers/pyproject.toml` is vestigial.

workers/queue_backend/pg_barrier.py
```diff
@@ -123,6 +123,49 @@ def _delete_barrier(execution_id: str) -> None:
...
+            (execution_id, batch_index),
+        )
+        return cur.fetchone() is not None
+
+
+def clear_execution_batches(execution_id: str) -> int:
+    """Delete all per-batch dedup markers for an execution; returns the count.
+
+    Called at barrier teardown (``remaining`` → 0) so the markers live exactly
+    as long as the execution needs them — the reaper's barrier-orphan sweep is
```
muhammad-ali-e: **P1 (dup) · Inaccurate docstring (verified).** Same "reaper's barrier-orphan sweep is the backstop" claim as `models.py:134`. The reaper (`pg_queue/reaper.py:131`) deletes only from `pg_barrier_state`; nothing reclaims orphaned `pg_batch_dedup` rows today. Reword to describe the leak + intended-future-sweep rather than an existing backstop, or — if a dedup sweep is genuinely planned — phrase it explicitly as future work.

workers/tests/test_pg_batch_dedup.py
```diff
@@ -0,0 +1,136 @@
...
+
+@pytest.fixture
+def dedup_db():
+    import os
+
+    os.environ.setdefault("TEST_DB_HOST", "127.0.0.1")
+    try:
+        conn = create_pg_connection(env_prefix="TEST_DB_")
+    except psycopg2.OperationalError as exc:
+        pytest.skip(f"Postgres not reachable: {exc}")
```
muhammad-ali-e: **P2 · Test can report green while testing nothing (MEDIUM).** Every test here is gated behind `dedup_db`, which `pytest.skip`s when Postgres is unreachable or migration 0006 is unapplied (also line 39, and the unqualified `to_regclass` probe vs the connection's custom `search_path`). In a CI env where Postgres is *expected*, a misconfigured `TEST_DB_HOST`, a DB container that failed to start, or a migration-ordering regression all yield a **skipped/green** run — the idempotency primitive (the one thing standing between an at-least-once queue and a double-decremented barrier) ships untested with no signal. This is an inherited pattern from `test_pg_barrier.py`, not new here, but it directly undermines the race test below.

Suggest: gate the skip behind a CI-set env flag (e.g. `REQUIRE_PG_TESTS=1` → convert skip to `pytest.fail`), or add a CI guard that fails if these PG tests report `collected>0 and skipped==0` is false.

workers/tests/test_pg_batch_dedup.py
```diff
@@ -0,0 +1,136 @@
...
+            conn = create_pg_connection(env_prefix="TEST_DB_")
+            conn.autocommit = True
+            conns.append(conn)
+            pg_barrier._local.conn = conn
+            try:
+                results[label] = claim_batch("exec-race", 0)
+            finally:
+                pg_barrier._local.conn = None
+
+        threads = [threading.Thread(target=run, args=(f"t{i}",)) for i in range(2)]
```
muhammad-ali-e: **P3 · Concurrency test doesn't reliably force the race.** Each thread calls `create_pg_connection()` *inside* `run()` before claiming, and there's no shared start barrier with only `range(2)` threads. Connection setup (TCP+auth) is slow and variable vs a single `INSERT … ON CONFLICT`, so thread A typically connects+inserts+commits before B finishes connecting — the contended `ON CONFLICT` path is rarely hit. The assertion `sum(...) == 1` would also pass for purely serial execution and even for a buggy non-atomic check-then-insert.

Suggest: build the N connections in the main thread *before* starting threads, add a `threading.Barrier(N)` immediately before each `claim_batch`, raise N to ~8–16, and loop the scenario a few times. Optionally also pin the invariant deterministically by asserting the production statement is a single `INSERT ... ON CONFLICT DO NOTHING` (mirrors how `test_pg_barrier.py` pins `max_retries==0`).

backend/pg_queue/models.py
```diff
@@ -114,6 +114,51 @@ class Meta:
...
+    Composite uniqueness on ``(execution_id, batch_index)`` is enforced by a
+    ``UniqueConstraint`` (Django has no composite primary key here); the worker's
+    ``ON CONFLICT (execution_id, batch_index)`` targets exactly those columns,
+    and the constraint's index — ``execution_id`` leading — also serves the
+    per-execution cleanup ``DELETE``. Managed=True / generated migration, no
+    DB-side function, extension-free (UN-3533) — same posture as the siblings.
+    """
+
+    execution_id = models.TextField()
+    batch_index = models.IntegerField()
```
muhammad-ali-e, line 152: **P4 · Unexpressed invariant: `batch_index >= 0`.** `IntegerField` admits negatives, but a batch index is a non-negative ordinal. The sibling `PgQueueMessage.priority` sets the house precedent of a `CheckConstraint` for a domain-bounded integer, and `PgBarrierState` sets the precedent of *documenting* a deliberate omission — this model does neither (silent gap). Low risk (a negative index is a programming error, not an attack surface), but it's an expressible invariant left unexpressed. Add a `CheckConstraint(batch_index__gte=0)` to `Meta.constraints` (near-zero cost, makes it writer-proof like the unique constraint), or add a one-line comment explaining why it's omitted.

backend/pg_queue/models.py
```diff
@@ -114,6 +114,51 @@ class Meta:
...
+    created_at = models.DateTimeField(default=timezone.now)
```
muhammad-ali-e, line 156: **P5 · `created_at` appears vestigial / under-documented.** It's written (model default + `now()` in the INSERT) but never read, indexed, asserted, or used for reclaim — unlike `PgBarrierState.expires_at`, which is indexed and load-bearing. If the intended orphan-sweep backstop (see P1) will sweep stale markers by age, it'll need an index on `created_at` and a test asserting the value is set; if the column is debugging-only, a one-line field comment would match the rigor of the sibling field comments. Worth clarifying its intended consumer either way.

workers/queue_backend/pg_barrier.py
```diff
@@ -123,6 +123,49 @@ def _delete_barrier(execution_id: str) -> None:
         )
 
 
+def claim_batch(execution_id: str, batch_index: int) -> bool:
+    """Claim ``(execution_id, batch_index)`` for processing — the per-batch
+    idempotency gate for the at-least-once PG pipeline.
+
+    Atomically inserts the dedup marker (``pg_batch_dedup``); returns ``True``
+    if THIS call inserted the row (first delivery — proceed and decrement the
```
muhammad-ali-e: **P6 · Docstring couples in behavior this function doesn't perform.** First paragraph: a `True` result means "first delivery — proceed and decrement the barrier," present tense; the third paragraph correctly notes the function is inert until wired in a later PR. Today `claim_batch` only inserts a row and returns a bool — it decrements nothing. A reader skimming the first paragraph may believe the decrement happens here. Suggest deferring the decrement semantics to the caller: *"`True` if THIS call inserted the row (first delivery — the caller should proceed and perform the single barrier decrement); `False` if it already existed (redelivery — caller should skip)."*

backend/pg_queue/models.py
```diff
@@ -114,6 +114,51 @@ class Meta:
         ]
 
 
+class PgBatchDedup(models.Model):
+    """Per-batch idempotency marker for the at-least-once PG execution pipeline.
+
+    The PG queue is at-least-once: a ``process_file_batch`` task can be
+    redelivered after a crash-before-ack, which would re-run the batch and
+    double-decrement the barrier (``process_file_batch`` is non-idempotent —
+    ``max_retries=0``, the destination push has only partial per-file
```
muhammad-ali-e: **P7 · Unverifiable rationale sub-claim.** `max_retries=0` is accurate (`workers/file_processing/tasks.py:266`), but "the destination push has only partial per-file protection" is a specific behavioral assertion I couldn't locate in `file_processing/tasks.py`. If that path later becomes fully idempotent, this rationale silently rots. Either cite the specific code path that gives the partial protection, or generalize to "the destination push is not fully idempotent on redelivery" so it doesn't pin to a protection level that may change.

workers/tests/test_pg_batch_dedup.py
```diff
@@ -0,0 +1,136 @@
...
+import psycopg2
+import pytest
+from queue_backend import pg_barrier
+from queue_backend.pg_barrier import claim_batch, clear_execution_batches
+from queue_backend.pg_queue.connection import create_pg_connection
+
+
+@pytest.fixture
+def dedup_db():
+    import os
```
muhammad-ali-e: **P8 · Minor consistency.** `import os` is inside the `dedup_db` fixture body; the sibling `test_pg_barrier.py` hoists `import os` to the module top (line 15) and keeps only the genuinely test-local `import threading` inline. Move `import os` to the module-level import block to match. (The inline `import threading` in the race test already matches the sibling — leave it.) Cosmetic, optional.

backend/pg_queue/models.py
muhammad-ali-e: Fixed in \`273ccfc75\` — verified, you're right: \`sweep_expired_barriers\` (reaper.py:131) DELETEs only \`pg_barrier_state\`, no cascade to this table, so orphaned markers leak today. Reworded to state the leak honestly + flag the dedup-orphan sweep (e.g. by \`created_at\` age) as intended future work, rather than describing a backstop that doesn't exist.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in \`273ccfc75\` (same as models.py:134). \`clear_execution_batches\` now says the reaper sweeps the orphaned \`pg_barrier_state\` row but does NOT yet reclaim \`pg_batch_dedup\`; dedup-orphan sweep = future work.

workers/tests/test_pg_batch_dedup.py
muhammad-ali-e: Addressed in \`273ccfc75\`. Added a \`_skip_or_fail(reason)\` helper: when \`REQUIRE_PG_TESTS\` is set (CI where PG is expected) the skip becomes a \`pytest.fail\`, so the idempotency primitive can't ship untested-green on a misconfigured \`TEST_DB_HOST\` / failed DB container / migration-ordering regression. Scoped to this file; rolling the same guard into \`test_pg_barrier.py\` is a reasonable follow-up.

workers/tests/test_pg_batch_dedup.py
muhammad-ali-e: Strengthened in \`273ccfc75\`. Now: pre-build all N=8 connections in the main thread, align every \`claim_batch\` with a \`threading.Barrier(8)\` released at the same instant, and loop 5 trials (distinct batch each). That forces the contended \`ON CONFLICT\` path instead of letting a fast first thread connect+commit first. The looped exactly-one-winner assertion would also catch a non-atomic check-then-insert.

backend/pg_queue/models.py
muhammad-ali-e: Added in \`273ccfc75\` — \`CheckConstraint(batch_index >= 0)\` (writer-proof, mirroring \`PgQueueMessage.priority\`'s precedent) + regenerated migration 0006 + a test asserting the DB rejects a negative index (\`psycopg2.errors.CheckViolation\`).

backend/pg_queue/models.py
muhammad-ali-e: Clarified in \`273ccfc75\` — added a field comment: \`created_at\` is observability/debugging only today (not read/indexed/used for reclaim); the intended future dedup-orphan sweep (P1) would sweep by its age and add an index then. Left it unindexed for now since nothing consumes it yet.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in \`273ccfc75\` — \`claim_batch\`'s docstring no longer claims it decrements. It now reads: True = "the caller should proceed and perform the single barrier decrement", and explicitly "This function itself only inserts the marker." Decrement semantics deferred to the caller (2c).

backend/pg_queue/models.py
muhammad-ali-e: Generalized in \`273ccfc75\` — "the destination push has only partial per-file protection" → "the destination push is not fully idempotent on redelivery", so it no longer pins to a specific protection level that could rot. (The recon details — Redis per-file lock + WFE status — live in the PR description / ledger rather than the model docstring.)

workers/tests/test_pg_batch_dedup.py
muhammad-ali-e: Done in \`273ccfc75\` — \`import os\` hoisted to the module-level import block; the test-local \`import threading\` stays inline (matches \`test_pg_barrier.py\`).

muhammad-ali-e:
## Review round addressed — `273ccfc75`

Thanks @muhammad-ali-e. All 9 threads handled — the standout is **P1, a verified bug** (not just a doc nit).

**P1 (verified bug) — reaper backstop claim was false.** I'd written that the reaper's barrier-orphan sweep reclaims orphaned `pg_batch_dedup` markers. It doesn't: `sweep_expired_barriers` (reaper.py:131) DELETEs only `pg_barrier_state`, there's no cascade, so orphaned markers **leak today**. Both docstrings (`PgBatchDedup` + `clear_execution_batches`) now state the leak honestly and flag a dedup-orphan sweep as intended future work.

**Fixed**
- **P4** — added `CheckConstraint(batch_index >= 0)` (writer-proof, mirrors `PgQueueMessage.priority`) + regenerated migration 0006 + a test that the DB rejects a negative index.
- **P3** — strengthened the race test: pre-build N=8 connections in the main thread, align claims with a `threading.Barrier`, loop 5 trials → actually forces the contended `ON CONFLICT` path (and catches a non-atomic check-then-insert).
- **P2** — `REQUIRE_PG_TESTS` env flips the skip to a `pytest.fail`, so the idempotency primitive can't ship untested-green where PG is expected.
- **P6** — `claim_batch` docstring no longer claims it decrements; defers the single decrement to the caller.
- **P7** — generalized "partial per-file protection" → "not fully idempotent on redelivery" (rot-proof).
- **P5** — documented `created_at` as observability-only (future age-based sweep would index it).
- **P8** — hoisted `import os` to module top.

**Verification**: 35 `pg_barrier` + 9 `pg_batch_dedup` tests green; migration 0006 regenerated + applied to the dev DB, `makemigrations --check` clean; ruff clean. Still inert — net behaviour change on the default path: NONE.

muhammad-ali-e:
## @greptile-apps — thanks, sharp catch (tracked for 2c)

The 3/5 finding is correct and verified: `PgBarrier.enqueue`'s UPSERT resets `pg_barrier_state` on `execution_id` reuse (the line-237 "clears leftover state from a prior run with this id" case) but does **not** clear `pg_batch_dedup`. Once `claim_batch` is wired, a re-enqueued/retried execution with the same id would hit stale markers → every batch gets `False` → nothing processed → barrier hangs to expiry. The `barrier_pg_abort` teardown has the same shape.

**Where it lands: 2c, not 2b.** The gap can only manifest once `claim_batch` is wired — until then `pg_batch_dedup` is always empty, so there's nothing to clear and no behaviour to fix. Adding the `clear_execution_batches` call to `enqueue` now would be wiring ahead of need (the same reason 2b deliberately ships the primitive with no call sites, and 2a deferred its threading to 2c). So it lands in 2c, atomically with the `claim_batch` wiring:

- `clear_execution_batches(execution_id)` in the **same `_cursor()` block** as the enqueue UPSERT (atomic reset), and
- the same call in the `barrier_pg_abort` teardown path (alongside the `pg_barrier_state` DELETE).

Captured as a **must-do** in the 2c scope with a same-`execution_id`-re-run dev-test to verify. 2b stays inert (net behaviour change on the default path: NONE). Your "before 2c ships" is satisfied by "first thing in 2c."

PR #2069, opened 2026-06-17
UN-3563 [FEAT] PG Queue 9e PR 2c — live PG fan-out/barrier/callback (fire-and-forget)

muhammad-ali-e:
## What

Third slice of **9e PR 2**, after **2a** (#2067) + **2b** (#2068). This is the **live switch**: it wires the coupled pipeline's **fan-out → barrier → callback** onto the PG queue for a `transport=="pg_queue"` execution (fire-and-forget self-chaining — no Celery chord / `.link`).

## Non-regressive by construction

`resolve_transport()` still returns `"celery"` (PR 3 wires Flipt), so the **entire PG branch is present-but-unreachable** — the default path is byte-identical. The flag `pg_queue_execution_enabled` stays **off** and is **not consumed by any code** until PR 3.

## Scope = hybrid

2c migrates the **fan-out → barrier → callback** (the at-least-once core). The orchestrator task (`async_execute_bin`) stays on Celery; routing it onto PG (wholly-off-Celery) is a follow-up (**2d**).

## How it works (on `transport=="pg_queue"`)

- `Barrier` Protocol + Celery/Redis impls accept (and ignore) a `transport` param; `CallbackDescriptor` gains an optional `backend` marker.
- `create_chord_execution` routes `pg_queue` to a **fresh `PgBarrier()`** — bypassing the `WORKER_BARRIER_BACKEND` singleton (the celery substrate), so the transport selects the substrate.
- `PgBarrier.enqueue` fire-and-forget mode: each header dispatched via `dispatch(backend=PG)` with an injected `_barrier_context {execution_id, batch_index, callback_descriptor}` (no `.link`); the UPSERT block also clears `pg_batch_dedup` (the reuse-reset flagged in 2b review — `@greptile-apps` #2068).
- `process_file_batch(_barrier_context=None)`: the PG path runs `run_batch_with_barrier` — **claim** the batch (idempotent on redelivery), run the work, **decrement** the barrier in-body; the task that drives `remaining → 0` **self-chains the callback** onto PG via `dispatch(backend=PG)`. Exceptions tear the barrier down in-body (mirrors the chord `.link_error`). `clear_execution_batches` runs at finalise and abort.

## Mixed-version / rolling-deploy safety

Three layers:
1. **Defaults** — new params (`transport`, `_barrier_context`) have safe defaults, so an old payload on a new worker falls back to celery/legacy.
2. **`transport` absorbed by old workers** — orchestrator tasks + the callback have `**kwargs`, so a pre-version worker swallows an unknown `transport` (no `TypeError`).
3. **`_barrier_context` never reaches a Celery/old worker** — injected only on the `pg_queue` path, which is gated off until PR 3 and routes only to the PG consumer (this code). That's *why* `process_file_batch` keeps an explicit param, not a `**kwargs` swallow.

**Operational rule** (naturally enforced by flag-default-off + separate PR): deploy 2c everywhere **before** PR 3 flips `pg_queue_execution_enabled`.

## Tests

- **+8 PgBarrier fire-and-forget**: header PG-dispatch + `_barrier_context` injection, reuse-clear, `run_batch_with_barrier` first/redelivery/exception, callback PG-vs-celery, abort-clears-dedup.
- **+2 orchestration routing**: `pg_queue` → fresh `PgBarrier` (not the singleton); celery → singleton with `transport` threaded.
- **+2 `process_file_batch` routing**: `_barrier_context` None → chord stages, set → `run_batch_with_barrier`.
- Each test file green alone; `ruff` clean.

## Dev-test (end-to-end, forced `pg_queue`)

The coupled pipeline ran **end-to-end on Postgres, fire-and-forget, twice, clean teardown**:
`process_file_batch` claimed + ran on a PG consumer → in-body barrier decrement → `remaining=0` → callback self-chained onto the PG callback queue → ran on a second PG consumer → execution **COMPLETED**. `pg_barrier_state` and `pg_batch_dedup` both empty afterward (finalise cleanup verified). No chord, no `.link`.

## Base

Targets the long-lived `feat/UN-3445-pg-queue-integration` integration branch (**not** `main`). Stacked on 2b.

### DIFF
--- a/workers/general/tasks.py
+++ b/workers/general/tasks.py
@@ -710,6 +710,7 @@ def _execute_general_workflow(
                 execution_mode=execution_mode,
                 use_file_history=use_file_history,
                 organization_id=api_client.organization_id,
+                transport=transport,
                 **kwargs,
             )
 
@@ -764,6 +765,7 @@ def _orchestrate_file_processing_general(
     execution_mode: tuple | None,
     use_file_history: bool,
     organization_id: str,
+    transport: str = DEFAULT_WORKFLOW_TRANSPORT,
     **kwargs: dict[str, Any],
 ) -> dict[str, Any]:
     """Orchestrate file processing for general workflows using the same pattern as API worker.
@@ -957,6 +959,7 @@ def _orchestrate_file_processing_general(
                 org_id=organization_id,
                 workload_type=WorkloadType.NON_API,
             ),
+            transport=transport,
         )
 
         if not result:
--- a/workers/general/tasks.py
+++ b/workers/general/tasks.py
@@ -56,6 +56,7 @@ from unstract.core.data_models import (
     FileBatchData,
     FileHashData,
     WorkerFileData,
+    normalize_transport,
 )
 
 # Import common workflow utilities
@@ -487,6 +488,12 @@ def _execute_general_workflow(
         Execution result
     """
     start_time = time.time()
+
+    # Fail-closed coercion, for parity with the api/scheduler workers (which run
+    # normalize_transport at their entry): a typo'd transport degrades to Celery
+    # with a warning rather than silently routing onto an unknown substrate. The
+    # coerced value feeds both WorkflowContextData and the fan-out below.
+    transport = normalize_transport(transport, logger=logger)
 
     logger.info("Executing general workflow logic for ETL/TASK workflow")
 
--- a/workers/queue_backend/barrier.py
+++ b/workers/queue_backend/barrier.py
@@ -94,11 +94,29 @@ class CallbackDescriptor(TypedDict):
     kwargs: dict[str, Any]
     queue: str
     fairness_headers: dict[str, Any] | None
-    # 9e: transport the aggregating callback is fired on when the barrier
-    # completes. Absent / ``None`` → legacy Celery dispatch (``current_app``
-    # ``.apply_async`` — the ``.link`` path). ``"pg_queue"`` → the fire-and-forget
-    # PG path self-chains the callback via ``dispatch(backend=PG)``.
-    backend: NotRequired[str | None]
+    # 9e: the WorkflowTransport the aggregating callback is fired on when the
+    # barrier completes. Absent / ``None`` → legacy Celery dispatch
+    # (``current_app.apply_async`` — the ``.link`` path). ``"pg_queue"`` → the
+    # fire-and-forget PG path self-chains the callback via dispatch onto PG.
+    # Named ``transport`` (a WorkflowTransport value, e.g. ``"pg_queue"``) — NOT
+    # ``backend``, to avoid confusion with ``QueueBackend`` (``"pg"``).
+    transport: NotRequired[str | None]
+
+
+class BarrierContext(TypedDict):
+    """Per-batch barrier coordination injected into a PG-dispatched header task.
+
+    The 9e fire-and-forget sibling of :class:`CallbackDescriptor`, and typed for
+    the same reason: it crosses producer → PG-queue → consumer, so the contract
+    is pinned here to catch a typo/rename at the type layer rather than as a
+    remote ``KeyError`` mid-batch. Carried as the ``_barrier_context`` kwarg on
+    ``process_file_batch`` so the consumer can claim its slot and run the barrier
+    decrement in-body (a PG-consumed task fires no Celery ``.link``).
+    """
+
+    execution_id: str
+    batch_index: int
+    callback_descriptor: CallbackDescriptor
 
 
 class Barrier(Protocol):
--- a/workers/queue_backend/barrier.py
+++ b/workers/queue_backend/barrier.py
@@ -36,9 +36,11 @@ from __future__ import annotations
 
 import logging
 import os
-from typing import TYPE_CHECKING, Any, Protocol, TypedDict
+from typing import TYPE_CHECKING, Any, NotRequired, Protocol, TypedDict
 
 from celery import chord
+
+from unstract.core.data_models import DEFAULT_WORKFLOW_TRANSPORT
 
 from .fairness import FairnessKey
 from .handle import BarrierHandle
@@ -92,6 +94,11 @@ class CallbackDescriptor(TypedDict):
     kwargs: dict[str, Any]
     queue: str
     fairness_headers: dict[str, Any] | None
+    # 9e: transport the aggregating callback is fired on when the barrier
+    # completes. Absent / ``None`` → legacy Celery dispatch (``current_app``
+    # ``.apply_async`` — the ``.link`` path). ``"pg_queue"`` → the fire-and-forget
+    # PG path self-chains the callback via ``dispatch(backend=PG)``.
+    backend: NotRequired[str | None]
 
 
 class Barrier(Protocol):
@@ -112,8 +119,14 @@ class Barrier(Protocol):
         callback_queue: str,
         app_instance: Any,
         fairness: FairnessKey | None = None,
+        transport: str = DEFAULT_WORKFLOW_TRANSPORT,
     ) -> BarrierHandle | None:
         """Enqueue ``header_tasks`` and a single callback to fire on completion.
+
+        ``transport`` is the per-execution transport (9e). Only ``PgBarrier``
+        acts on it (``pg_queue`` → fire-and-forget PG fan-out instead of Celery
+        ``.link``); the Celery/Redis substrates accept it for Protocol parity
+        and ignore it (they are only ever reached on the ``celery`` transport).
 
         ``None`` is the **sole signal** that no work was enqueued — it
         is returned exclusively when ``header_tasks`` is empty. Any
@@ -166,8 +179,10 @@ class CeleryChordBarrier:
         callback_queue: str,
         app_instance: Any,
         fairness: FairnessKey | None = None,
+        transport: str = DEFAULT_WORKFLOW_TRANSPORT,
     ) -> BarrierHandle | None:
         """See :class:`Barrier.enqueue`."""
+        del transport  # Celery chord is only reached on the celery transport.
         # Empty-header guard goes FIRST so a zero-task run skips
         # signature construction / fairness serialisation entirely.
         # Any failure in those paths is now constrained to the
--- a/workers/queue_backend/pg_barrier.py
+++ b/workers/queue_backend/pg_barrier.py
@@ -318,45 +318,14 @@ class PgBarrier:
                     (execution_id,),
                 )
 
-            link_signature = barrier_pg_decr_and_check.s(
+            self._dispatch_headers(
+                header_tasks,
+                is_pg=is_pg,
                 execution_id=execution_id,
                 callback_descriptor=callback_descriptor,
+                fairness=fairness,
+                fairness_headers=fairness_headers,
             )
-            link_error_signature = barrier_pg_abort.s(execution_id=execution_id)
-            for i, task in enumerate(header_tasks):
-                try:
-                    if is_pg:
-                        self._dispatch_header_pg(
-                            task, i, execution_id, callback_descriptor, fairness
-                        )
-                    else:
-                        cloned = task.clone()
-                        if fairness_headers:
-                            cloned.set(headers=fairness_headers)
-                        cloned.link(link_signature)
-                        cloned.link_error(link_error_signature)
-                        cloned.apply_async()
-                except Exception:
-                    # Mid-loop dispatch failure: i of N never reached the queue,
-                    # so the counter can't reach 0. Delete the row so an in-flight
-                    # decrement (link or in-body) finds no row and cleans up;
-                    # re-raise so the caller marks the workflow ERROR.
-                    with contextlib.suppress(Exception):
-                        _delete_barrier(execution_id)
-                    if is_pg:
-                        # On the PG path, headers dispatched before i may already
-                        # have run + committed a claim_batch marker. With the
-                        # barrier row gone, their in-flight barrier_pg_abort is a
-                        # no-op (already_aborted) and never reaches the clear
-                        # inside it — so reclaim the markers here directly.
-                        with contextlib.suppress(Exception):
-                            clear_execution_batches(execution_id)
-                    logger.exception(
-                        f"[exec:{execution_id}] header dispatch failed at task "
-                        f"{i}/{len(header_tasks)}; barrier row deleted to prevent "
-                        f"spurious callback fires from the orphan tasks."
-                    )
-                    raise
 
             logger.info(
                 f"Barrier enqueued via PgBarrier ({transport}) — "
@@ -374,6 +343,56 @@ class PgBarrier:
                 f"header_tasks={len(header_tasks)})"
             )
             raise
+
+    def _dispatch_headers(
+        self,
+        header_tasks: list[Signature],
+        *,
+        is_pg: bool,
+        execution_id: str,
+        callback_descriptor: CallbackDescriptor,
+        fairness: FairnessKey | None,
+        fairness_headers: dict[str, Any] | None,
+    ) -> None:
+        """Dispatch the N header tasks, one transport or the other.
+
+        PG path (``is_pg``) → fire-and-forget onto the PG queue via
+        :meth:`_dispatch_header_pg` (no ``.link``). Celery path → ``.link`` /
+        ``.link_error`` chord-style. On any mid-loop dispatch failure, ``i`` of N
+        never reached the queue so the counter can't reach 0 — delete the barrier
+        row (and, on the PG path, reclaim dedup markers an earlier header may have
+        committed, since the in-flight ``barrier_pg_abort`` is a no-op once the row
+        is gone) so an in-flight decrement finds nothing, then re-raise.
+        """
+        link_signature = barrier_pg_decr_and_check.s(
+            execution_id=execution_id, callback_descriptor=callback_descriptor
+        )
+        link_error_signature = barrier_pg_abort.s(execution_id=execution_id)
+        for i, task in enumerate(header_tasks):
+            try:
+                if is_pg:
+                    self._dispatch_header_pg(
+                        task, i, execution_id, callback_descriptor, fairness
+                    )
+                else:
+                    cloned = task.clone()
+                    if fairness_headers:
+                        cloned.set(headers=fairness_headers)
+                    cloned.link(link_signature)
+                    cloned.link_error(link_error_signature)
+                    cloned.apply_async()
+            except Exception:
+                with contextlib.suppress(Exception):
+                    _delete_barrier(execution_id)
+                if is_pg:
+                    with contextlib.suppress(Exception):
+                        clear_execution_batches(execution_id)
+                logger.exception(
+                    f"[exec:{execution_id}] header dispatch failed at task "
+                    f"{i}/{len(header_tasks)}; barrier row deleted to prevent "
+                    f"spurious callback fires from the orphan tasks."
+                )
+                raise
 
     @staticmethod
     def _dispatch_header_pg(
--- a/workers/queue_backend/pg_barrier.py
+++ b/workers/queue_backend/pg_barrier.py
@@ -724,9 +724,13 @@ def run_batch_with_barrier(
       ``skipped_redelivery`` and never re-runs the batch — the barrier hangs to
       ``expires_at``. (Claiming before the work avoids re-processing on the common
       redelivery case at the cost of this crash window.)
-    - A failure *after* ``remaining`` hits 0 and the callback is committed but
-      before/at the callback enqueue: the decrement is committed, so redelivery is
-      blocked, and the abort here removes the row without the callback having run.
+    - A hard crash *between* the decrement committing (``remaining`` → 0) and the
+      callback enqueue completing: the decrement is committed (so redelivery is
+      blocked by the marker) but the process is gone before the callback is
+      enqueued and before any in-body abort can run, so the ``pg_barrier_state``
+      row survives to ``expires_at`` with no callback ever fired. (A *software*
+      callback-dispatch failure here is NOT this window — that's catchable and is
+      torn down by step 3's wrap above.)
 
     For both, the **reaper** (sweep stranded ``pg_barrier_state``, mark the
     execution ERROR / re-drive, reclaim ``pg_batch_dedup``) is the recovery net;
--- a/workers/queue_backend/pg_barrier.py
+++ b/workers/queue_backend/pg_barrier.py
@@ -60,12 +60,14 @@ import contextlib
 import json
 import logging
 import threading
-from collections.abc import Iterator
+from collections.abc import Callable, Iterator
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
 
 import psycopg2
 import psycopg2.extensions
+
+from unstract.core.data_models import DEFAULT_WORKFLOW_TRANSPORT, WorkflowTransport
 
 from .barrier import CallbackDescriptor, barrier_ttl_seconds
 from .decorator import worker_task
@@ -197,15 +199,26 @@ class PgBarrier:
         callback_queue: str,
         app_instance: Any,
         fairness: FairnessKey | None = None,
+        transport: str = DEFAULT_WORKFLOW_TRANSPORT,
     ) -> BarrierHandle | None:
         """See :class:`queue_backend.barrier.Barrier.enqueue`.
 
         Empty ``header_tasks`` → ``None`` (caller owns the zero-files contract);
         any substrate failure raises. ``app_instance`` is accepted for Protocol
-        parity but unused — the callback is built inside the link task via
-        ``current_app`` so it runs against the worker's app.
+        parity but unused.
+
+        ``transport`` selects the fan-out mode (9e):
+
+        - ``celery`` (default, the ``WORKER_BARRIER_BACKEND=pg`` legacy path):
+          header tasks are Celery-dispatched with ``.link(barrier_pg_decr_and_check)``
+          / ``.link_error(barrier_pg_abort)``; the link drives the decrement.
+        - ``pg_queue`` (the fire-and-forget path): header tasks are dispatched onto
+          the PG queue (``dispatch(backend=PG)``) with a ``_barrier_context`` kwarg
+          carrying ``execution_id`` / ``batch_index`` / ``callback_descriptor`` —
+          a PG-consumed task fires no ``.link``, so it claims its batch and runs
+          the decrement in-body, self-chaining the callback at ``remaining`` → 0.
         """
-        del app_instance  # Protocol parity; callback built in the link task.
+        del app_instance  # Protocol parity; callback dispatched by the decrement.
         if not header_tasks:
             logger.info(
                 f"[exec:{callback_kwargs.get('execution_id')}] "
@@ -223,6 +236,7 @@ class PgBarrier:
             )
         execution_id = str(execution_id)
 
+        is_pg = transport == WorkflowTransport.PG_QUEUE.value
         try:
             fairness_headers = fairness.as_header() if fairness else None
             callback_descriptor: CallbackDescriptor = {
@@ -231,6 +245,10 @@ class PgBarrier:
                 "queue": callback_queue,
                 "fairness_headers": fairness_headers,
             }
+            if is_pg:
+                # The decrement (run in-body on the PG path) reads this to
+                # self-chain the callback onto PG rather than Celery.
+                callback_descriptor["backend"] = WorkflowTransport.PG_QUEUE.value
             ttl_seconds = barrier_ttl_seconds()
 
             with _cursor() as cur:
@@ -249,6 +267,15 @@ class PgBarrier:
                     "  created_at = now(), expires_at = EXCLUDED.expires_at",
                     (execution_id, len(header_tasks), ttl_seconds),
                 )
+                # Reset per-batch dedup markers from a prior run reusing this
+                # execution_id, ATOMICALLY with the barrier reset. Without this a
+                # re-enqueue would leave stale markers → every in-body claim_batch
+                # returns False → all batches skip → barrier hangs to expiry.
+                # (No-op pre-PR-2c: nothing writes pg_batch_dedup yet.)
+                cur.execute(
+                    "DELETE FROM pg_batch_dedup WHERE execution_id = %s",
+                    (execution_id,),
+                )
 
             link_signature = barrier_pg_decr_and_check.s(
                 execution_id=execution_id,
@@ -257,30 +284,35 @@ class PgBarrier:
             link_error_signature = barrier_pg_abort.s(execution_id=execution_id)
             for i, task in enumerate(header_tasks):
                 try:
-                    cloned = task.clone()
-                    if fairness_headers:
-                        cloned.set(headers=fairness_headers)
-                    cloned.link(link_signature)
-                    cloned.link_error(link_error_signature)
-                    cloned.apply_async()
+                    if is_pg:
+                        self._dispatch_header_pg(
+                            task, i, execution_id, callback_descriptor, fairness
+                        )
+                    else:
+                        cloned = task.clone()
+                        if fairness_headers:
+                            cloned.set(headers=fairness_headers)
+                        cloned.link(link_signature)
+                        cloned.link_error(link_error_signature)
+                        cloned.apply_async()
                 except Exception:
-                    # Mid-loop dispatch failure: i of N never reached the broker,
-                    # so the counter can't reach 0. Delete the row so in-flight
-                    # links' decrement finds no row and cleans up; re-raise so the
-                    # caller marks the workflow ERROR.
+                    # Mid-loop dispatch failure: i of N never reached the queue,
+                    # so the counter can't reach 0. Delete the row so an in-flight
+                    # decrement (link or in-body) finds no row and cleans up;
+                    # re-raise so the caller marks the workflow ERROR.
                     with contextlib.suppress(Exception):
                         _delete_barrier(execution_id)
                     logger.exception(
-                        f"[exec:{execution_id}] apply_async failed at task "
+                        f"[exec:{execution_id}] header dispatch failed at task "
                         f"{i}/{len(header_tasks)}; barrier row deleted to prevent "
-                        f"spurious callback fires from the orphan tasks' links."
+                        f"spurious callback fires from the orphan tasks."
                     )
                     raise
 
             logger.info(
-                f"Barrier enqueued via PgBarrier — exec_id={execution_id}, "
-                f"header_tasks={len(header_tasks)}, callback={callback_task_name}, "
-                f"queue={callback_queue}"
+                f"Barrier enqueued via PgBarrier ({transport}) — "
+                f"exec_id={execution_id}, header_tasks={len(header_tasks)}, "
+                f"callback={callback_task_name}, queue={callback_queue}"
             )
             return _PgBarrierHandle(id=execution_id)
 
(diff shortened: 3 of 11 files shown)


muhammad-ali-e:
## SonarCloud S1172 fixed — `d986c3823`

`_run_batch_stages` no longer takes the unused `task_instance` param. When I extracted it from `_process_file_batch_core`, `task_instance`'s only use (deriving `celery_task_id`) stayed in the caller, so the extracted helper received it but never used it. Removed the param + updated both call sites; `_process_file_batch_core` keeps it (it reads `.request.id`). Routing test mocks with `*a`, unaffected; ruff clean.

workers/queue_backend/pg_barrier.py
```diff
@@ -510,8 +616,70 @@ def barrier_pg_abort(
...
+    except Exception:
+        with contextlib.suppress(Exception):
+            barrier_pg_abort(execution_id=execution_id)
+        logger.exception(
+            f"[exec:{execution_id}] batch {batch_index} failed — barrier torn "
+            f"down in-body (no .link_error on the PG path); execution fails fast."
+        )
+        raise
+
+    _barrier_pg_decrement(
```
muhammad-ali-e: **[Critical] Decrement failure after a successful batch strands the barrier (no teardown).**

`run_batch_with_barrier` does `claim_batch()` → `work_fn()` → `_barrier_pg_decrement()`. If the work succeeds but `_barrier_pg_decrement` then raises (the open-transaction `RuntimeError` guard, `json.dumps` `TypeError`/`ValueError`, or any DB error on the `UPDATE` — none of which hit the `DataError` self-teardown branch), the exception propagates with the dedup marker **already committed**. There is no `barrier_pg_abort` on this path (unlike the `work_fn` branch), so the barrier never reaches 0 → the aggregating callback never fires → the execution hangs until `expires_at` (~6h). Redelivery hits `claim_batch` → `False` → `skipped_redelivery`, so it can never heal.

**Fix:** wrap the `work_fn()` *and* the decrement in one `try/except` that calls `barrier_pg_abort(execution_id=...)` before re-raising, so a decrement failure fails fast like a work failure instead of hanging to expiry.

muhammad-ali-e (commented):
## Automated PR Review — PG Queue 9e PR 2c (live fan-out/barrier/callback)

Multi-agent review (code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier). The implementation is well-structured and heavily documented; the new tests pass. The dominant theme below: the new **in-body fire-and-forget path lacks the `.link_error` safety net of the chord path**, so several failure modes (decrement raise, suppressed abort, last-batch callback-dispatch failure) degrade to a silent ~6h hang-to-expiry that redelivery can't heal because the claim is already consumed. Findings are posted inline below, prioritised by severity.

workers/queue_backend/pg_barrier.py
```diff
@@ -510,8 +616,70 @@ def barrier_pg_abort(
...
+       ``remaining`` → 0 self-chains the aggregating callback onto PG.
+
+    ``work_fn`` must return the JSON-serialisable batch result (the same dict the
+    Celery chord path returns), since the decrement appends it to the barrier's
+    aggregated results that the callback receives.
+    """
+    execution_id = str(barrier_context["execution_id"])
+    batch_index = int(barrier_context["batch_index"])
+
+    if not claim_batch(execution_id, batch_index):
```
muhammad-ali-e, line 801: **[High] `claim_batch` commits the dedup marker *before* `work_fn()` runs — a hard crash wedges the batch.**

Because the claim commits first, a hard kill (OOM/SIGKILL/pod eviction) or visibility-timeout expiry *during* `work_fn()` (not a catchable Python exception, so the in-body `barrier_pg_abort` never runs) leaves the marker set. On redelivery `claim_batch` returns `False` → `skipped_redelivery` with no re-run and no decrement → the barrier hangs to expiry and the batch's files are never processed. This downgrades the queue's at-least-once delivery to at-most-once for the actual batch work.

**Fix:** either fold claim+decrement into a single transaction so a crash rolls back the claim (allowing clean redelivery), or, if keeping claim-before-work, document the crash window explicitly and treat the dedup-orphan reaper as a hard merge dependency rather than future work.

workers/queue_backend/pg_barrier.py
```diff
@@ -510,8 +616,70 @@ def barrier_pg_abort(
...
+            "status": "skipped_redelivery",
+            "execution_id": execution_id,
+            "batch_index": batch_index,
+        }
+
+    try:
+        result = work_fn()
+    except Exception:
+        with contextlib.suppress(Exception):
+            barrier_pg_abort(execution_id=execution_id)
```
muhammad-ali-e: **[High] Suppressed `barrier_pg_abort` can silently leave the barrier alive while the log claims it was torn down.**

```python
except Exception:
        barrier_pg_abort(execution_id=execution_id)
```
A batch failure and a DB failure are correlated, so `barrier_pg_abort` itself can fail — and it's swallowed with no log of its own. The outer `logger.exception(... "barrier torn down in-body" ...)` is then **false**, and there's no retry (`max_retries=0`, fire-and-forget). The barrier hangs to expiry and the log actively misleads the debugger.

**Fix:** log inside the suppression so a failed teardown is visible, and soften the wording to "tear-down attempted".

workers/queue_backend/pg_barrier.py
```diff
@@ -294,6 +326,90 @@ def enqueue(
...
+
+    - ``"pg_queue"`` (9e fire-and-forget PG path): self-chain the callback onto
+      the PG queue via ``dispatch(backend=PG)`` — no Celery, so the whole
+      execution stays off the broker. (The callback rides default priority; the
+      barrier's fan-out fairness has already been applied to the header tasks.)
+    - absent / anything else (legacy ``.link`` path): dispatch via
+      ``current_app.signature(...).apply_async()`` — byte-identical to pre-9e,
+      preserving the ``fairness_headers`` the producer attached.
+    """
+    if callback_descriptor.get("backend") == WorkflowTransport.PG_QUEUE.value:
```
muhammad-ali-e: **[High] Last-batch PG callback-dispatch failure → execution silently never completes.**

On the PG branch of `_fire_barrier_callback`, `dispatch(..., backend=QueueBackend.PG)` can raise (its no-silent-fallback contract: DB down / serialisation error). By then the decrement has already committed `remaining = 0`. The exception propagates out of `run_batch_with_barrier`, the barrier row is left intact (delete is after), but redelivery hits `claim_batch` → `False` → `skipped_redelivery`, so the callback is **never retried**. The completed barrier hangs to expiry and the aggregating callback never runs.

**Fix:** make the completing decrement replay-safe — e.g. span the decrement `UPDATE` and the PG callback enqueue in one transaction (don't commit `remaining=0` until the callback enqueue succeeds), or on dispatch failure tear the barrier down / mark the execution ERROR so it fails fast.

workers/queue_backend/pg_barrier.py
```diff
@@ -406,47 +520,39 @@ def _barrier_pg_decrement(
...
+    # max_retries=0, so the task is never replayed.
+    callback_id = _fire_barrier_callback(callback_descriptor, all_results)
+    # Post-dispatch cleanup is best-effort: the callback already fired, so a
+    # failure here is logged (not raised) — the rows linger until reclaim rather
+    # than re-running the callback. Clear the per-batch dedup markers too (their
+    # job is done once the barrier completes); a leftover marker is bounded by
+    # the future dedup-orphan sweep, never re-runs anything.
     try:
         _delete_barrier(execution_id)
+        clear_execution_batches(execution_id)
```
muhammad-ali-e, line 637: **[Medium] Post-dispatch cleanup couples two ops, and `pg_batch_dedup` markers leak on non-finalising executions.**

`_delete_barrier(execution_id)` and `clear_execution_batches(execution_id)` share one `try/except`: if the first raises, the second never runs and the single log line can't say which failed. Separately, dedup markers are only reclaimed on happy-path finalise and winning abort — the module docstring notes executions that never finalise leak markers (the reaper sweeps `pg_barrier_state` but not `pg_batch_dedup`, "intended future work"). This is an unbounded leak in a now-**live** path.

**Fix:** run the two cleanups independently with distinct logs; and land the dedup-orphan sweep (or an `ON DELETE CASCADE` FK from `pg_batch_dedup` to `pg_barrier_state`) alongside this live switch rather than deferring it.

workers/queue_backend/pg_barrier.py
```diff
@@ -294,6 +326,90 @@ def enqueue(
...
+        runs the task; it claims ``(execution_id, batch_index)`` and runs the
+        decrement in-body (no ``.link``). ``fairness`` carries org/priority onto
+        the row exactly as the bare-dispatch sites do.
+        """
+        # Local import to avoid an import cycle (dispatch/routing pull in queue
+        # plumbing that imports the barrier package).
+        from .dispatch import dispatch
+        from .routing import QueueBackend
+
+        header_kwargs = dict(task.kwargs or {})
```
muhammad-ali-e: **[Low] `_dispatch_header_pg` can silently fall back to the default PG queue.**

`queue=task.options.get("queue")` is `None` if a signature was built without a queue option; `dispatch` → `_enqueue_pg` then substitutes `_DEFAULT_PG_QUEUE`, silently routing off the intended `file_processing` queue. Latent today (the fan-out always sets queue) but an easy footgun.

**Fix:** assert/log when `task.options.get("queue")` is `None`, or pass the known header/callback queue explicitly so a missing option fails loudly.

workers/queue_backend/pg_barrier.py
```diff
@@ -510,8 +616,70 @@ def barrier_pg_abort(
...
+
     logger.error(
         f"[exec:{execution_id}] PgBarrier aborted — a header task failed; "
         f"barrier state cleaned up (aggregating callback will not fire)."
     )
     return {"status": "aborted", "execution_id": execution_id}
+
+
+def run_batch_with_barrier(
+    barrier_context: dict[str, Any], work_fn: Callable[[], dict[str, Any]]
```
muhammad-ali-e: **[High] `barrier_context` is an untyped `dict[str, Any]` despite crossing the same serialization boundary that motivated `CallbackDescriptor` to be a `TypedDict`.**

It carries `execution_id` / `batch_index` / `callback_descriptor` producer→PG-queue→consumer — exactly the boundary `CallbackDescriptor` was typed for ("catch a typo/rename before it surfaces as a remote `KeyError` mid-aggregation"). Here the invariant is enforced only by runtime `barrier_context[...]` accesses, i.e. a remote `KeyError`/`ValueError` mid-batch.

**Fix:** add a sibling `BarrierContext(TypedDict)` next to `CallbackDescriptor` and use it on `_dispatch_header_pg`'s built dict, `run_batch_with_barrier(barrier_context: BarrierContext, ...)`, and the `_barrier_context` params in `file_processing/tasks.py`.

workers/queue_backend/barrier.py
```diff
@@ -92,6 +94,11 @@ class CallbackDescriptor(TypedDict):
     kwargs: dict[str, Any]
     queue: str
     fairness_headers: dict[str, Any] | None
+    # 9e: transport the aggregating callback is fired on when the barrier
+    # completes. Absent / ``None`` → legacy Celery dispatch (``current_app``
+    # ``.apply_async`` — the ``.link`` path). ``"pg_queue"`` → the fire-and-forget
+    # PG path self-chains the callback via ``dispatch(backend=PG)``.
+    backend: NotRequired[str | None]
```
muhammad-ali-e: **[Medium] `CallbackDescriptor["backend"]` field name collides with the `QueueBackend` enum but stores a `WorkflowTransport` value.**

This field holds `WorkflowTransport.PG_QUEUE.value` (`"pg_queue"`), yet `QueueBackend.PG` is `"pg"` — and `_fire_barrier_callback` passes `backend=QueueBackend.PG` two lines below where it reads `descriptor.get("backend")`. Within one function "backend" means two different things with two different strings, undermining the `routing.py` guardrail against `== "pg"` mistakes. It's also typed as bare `str`, admitting any value.

**Fix:** rename the field to `transport` and type it `NotRequired[WorkflowTransport | None]` (still JSON-serialises to `"pg_queue"`).

workers/queue_backend/pg_barrier.py
```diff
@@ -249,6 +267,15 @@ def enqueue(
                     "  created_at = now(), expires_at = EXCLUDED.expires_at",
                     (execution_id, len(header_tasks), ttl_seconds),
                 )
+                # Reset per-batch dedup markers from a prior run reusing this
+                # execution_id, ATOMICALLY with the barrier reset. Without this a
+                # re-enqueue would leave stale markers → every in-body claim_batch
+                # returns False → all batches skip → barrier hangs to expiry.
+                # (No-op pre-PR-2c: nothing writes pg_batch_dedup yet.)
```
muhammad-ali-e: **[High] Born-stale comment: contradicts this very PR.**

`# (No-op pre-PR-2c: nothing writes pg_batch_dedup yet.)` — but this *is* PR 2c, which adds `run_batch_with_barrier` → `claim_batch()` → `INSERT INTO pg_batch_dedup`. The three lines above even justify the `DELETE` precisely *because* `claim_batch` now writes those markers.

**Fix:** delete the parenthetical (or replace with `# Markers are written by claim_batch() on the in-body PG path (this PR).`). While here, also refresh the now-stale "Inert in 9e PR 2b — wired into ... in PR 2c" notes on `claim_batch`/`clear_execution_batches`.

workers/queue_backend/pg_barrier.py
```diff
@@ -294,6 +326,90 @@ def enqueue(
...
+    def _dispatch_header_pg(
+        task: Signature,
+        batch_index: int,
+        execution_id: str,
+        callback_descriptor: CallbackDescriptor,
+        fairness: FairnessKey | None,
+    ) -> None:
+        """Dispatch one header task onto the PG queue (fire-and-forget mode).
+
+        Unpacks the Celery ``Signature`` (the fan-out built it as
```
muhammad-ali-e, line 407: **[Medium] Docstring misstates how the fan-out builds the signature.**

`_dispatch_header_pg`'s docstring says the fan-out built it as `app.signature(name, args=[batch], queue=...)`, but `orchestration_utils.py` builds `app.signature(task_name, kwargs=batch_kwargs, queue=...)` — the batch payload is in **kwargs**, which this function's own body confirms (it reads `task.kwargs` and treats `task.args` as empty).

**Fix:** correct the parenthetical to `kwargs={batch_files, batch_index, total_batches}`.

workers/queue_backend/pg_barrier.py
```diff
@@ -294,6 +326,90 @@ def enqueue(
...
+        ``app.signature(name, args=[batch], queue=...)``) and re-dispatches it via
+        ``dispatch(backend=PG)`` with a ``_barrier_context`` kwarg. The PG consumer
+        runs the task; it claims ``(execution_id, batch_index)`` and runs the
+        decrement in-body (no ``.link``). ``fairness`` carries org/priority onto
+        the row exactly as the bare-dispatch sites do.
```
muhammad-ali-e: **[Medium] Duplicated local-import + `dispatch(backend=PG)` preamble.**

`_dispatch_header_pg` and `_fire_barrier_callback` both repeat the cycle-avoidance local imports (`from .dispatch import dispatch` / `from .routing import QueueBackend`) and the `backend=QueueBackend.PG` dispatch shape.

**Fix:** extract a module-level `_dispatch_pg(task_name, *, args, kwargs, queue, fairness=None)` helper that owns the single local import and the `backend=QueueBackend.PG` arg; both call sites collapse to one line.

workers/shared/workflow/execution/orchestration_utils.py
```diff
@@ -10,25 +10,41 @@
...
 
+def _barrier_for_transport(transport: str) -> Barrier:
+    """Pick the fan-in substrate for a per-execution transport (9e).
+
+    The ``pg_queue`` transport always coordinates on Postgres — it uses a fresh
+    :class:`PgBarrier` in its fire-and-forget mode regardless of
+    ``WORKER_BARRIER_BACKEND`` (the env-selected singleton is the *celery*-transport
+    substrate, which may be ``chord``). Every other transport uses that singleton.
+    """
+    if transport == WorkflowTransport.PG_QUEUE.value:
```
muhammad-ali-e: **[Low] Repeated `== WorkflowTransport.PG_QUEUE.value` literal across two modules.**

The bare `.value` comparison recurs here and at `pg_barrier.py:239/251/382`.

**Fix (optional):** add an `is_pg_transport(transport: str) -> bool` predicate next to the enum and reuse it, centralising "what counts as PG transport" for when a second PG-family transport is added.

workers/general/tasks.py
```diff
@@ -710,6 +710,7 @@ def _execute_general_workflow(
                 execution_mode=execution_mode,
                 use_file_history=use_file_history,
                 organization_id=api_client.organization_id,
+                transport=transport,
```
muhammad-ali-e, line 720: **[Medium] General worker forwards `transport` raw — fail-open, inconsistent with api/scheduler workers.**

`api-deployment/tasks.py:695` and `scheduler/tasks.py:150` run the incoming `transport` through `normalize_transport(...)` (fail-closed coercion + a warning log). The general worker accepts it at the task entry and threads it through unchanged. A typo like `"pg-queue"` silently degrades to the Celery singleton with **no log**, making a misconfiguration hard to diagnose. Safe in direction, but silent.

**Fix:** normalize at the general worker entry for parity: `transport = normalize_transport(transport, logger=logger, context=f" [exec:{execution_id}]")`.

workers/tests/test_pg_barrier.py
```diff
@@ -478,5 +492,141 @@ def test_expires_at_must_exceed_created_at(self, barrier_db):
                 )
 
 
+class TestPgFireAndForgetMode:
```
muhammad-ali-e, line 495: **[High] Missing end-to-end test for the PR's core promise.**

No test drives `run_batch_with_barrier` to `remaining == 0` and asserts it self-chains the aggregating callback onto PG. The pieces are tested separately (`_fire_barrier_callback` in isolation; `_barrier_pg_decrement` completion via the legacy `_CALLBACK`, which has no `backend` marker), so a regression where the descriptor loses `backend="pg_queue"` at the zero boundary wouldn't be caught. Also, `test_enqueue_pg_mode_dispatches_headers_via_pg_with_context` never asserts the header's `args`, `queue`, and pre-existing `kwargs` survive the Signature→dispatch unpacking — a dropped `args` (empty batch) would pass silently.

**Fix:** add a test seeding `remaining=1` with a `backend="pg_queue"` descriptor, patch `dispatch`, call `run_batch_with_barrier`, and assert `dispatch(backend=QueueBackend.PG, args=[[result]])`, row deleted, dedup cleared; and extend the header test to assert `args`/`queue`/existing `kwargs` preservation.

workers/tests/test_pg_barrier.py
```diff
@@ -478,5 +492,141 @@ def test_expires_at_must_exceed_created_at(self, barrier_db):
...
+
+        def boom():
+            raise RuntimeError("batch failed")
+
+        with pytest.raises(RuntimeError, match="batch failed"):
+            run_batch_with_barrier(ctx, boom)
+        # No .link_error on the PG path → torn down in-body (mirror chord abort).
+        assert _row(barrier_db, "exec-err") is None
+
+    def test_fire_barrier_callback_pg_self_chains_via_dispatch(self):
```
muhammad-ali-e, line 607: **[Medium] Untested concurrency / cleanup edges.**

Gaps: (1) no race test for two `claim_batch`/`run_batch_with_barrier` calls on the same `(execution_id, batch_index)` across two real connections resolving to exactly one decrement (the docstring claims this); (2) the **success** finalise path's new `clear_execution_batches` call (only the abort path is covered by `test_abort_clears_dedup_markers`); (3) double-abort on a redelivered failing batch when the row is already gone; (4) the PG branch of the mid-loop dispatch-failure cleanup (`test_mid_loop_dispatch_failure_deletes_row` covers only the Celery branch).

**Fix:** add the two-real-connection race test (mirroring `test_concurrent_aborts_deduplicate`), a success-path dedup-clear assertion, a double-abort test, and a PG-branch mid-loop-failure test.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in \`f5c6b98d5\`. \`run_batch_with_barrier\` now wraps **both** \`work_fn()\` and \`_barrier_pg_decrement()\` in the abort try/except, so a catchable decrement-side failure (the open-txn guard, json/DB error, or the last-batch callback dispatch raising) tears the barrier down in-body via \`barrier_pg_abort\` and re-raises — fails fast instead of stranding to expiry. + test \`test_decrement_failure_aborts_barrier\`.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Acknowledged as a real, **un-catchable** window (SIGKILL/vt-expiry mid-work can't run an in-process abort). Per the decision on the PR, this is reaper territory: I've documented it explicitly on \`run_batch_with_barrier\` as a **hard dependency** — the reaper must sweep stranded \`pg_barrier_state\` (mark exec ERROR / re-drive, reclaim \`pg_batch_dedup\`) and be live **before PR 3 flips the flag**, not "future work". 2c is gated off (resolve_transport→celery), so the window can't bite until then. (Claiming before work is deliberate — it avoids re-processing on the common redelivery case; the crash window is the trade, covered by the reaper.)

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in \`f5c6b98d5\` — extracted \`_abort_barrier_in_body(execution_id, reason=...)\` which logs *inside* the suppression: success → "teardown attempted"; failure → an explicit "teardown itself failed — will hang until expiry, reclaimed by the reaper". No more false "torn down" line.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Two parts. The catchable part (callback \`dispatch\` raises during the last-batch decrement) is now covered by the #69 wrap — the decrement is inside the abort try/except, so a dispatch failure tears the barrier down + re-raises. The residual (failure *after* remaining=0 is committed) is un-catchable here and is documented on \`run_batch_with_barrier\` as a reaper dependency (same gate as #74). Spanning decrement+callback-enqueue in one txn is the deeper fix; noted for the reaper-recovery work that gates PR 3.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Both addressed in \`f5c6b98d5\`. (1) The finalise cleanup is now two independent try/excepts with distinct logs (pg_barrier_state vs pg_batch_dedup), so one failing doesn't skip the other. (2) The dedup-marker leak on non-finalising executions is documented as a hard reaper dependency for PR 3 (dedup-orphan sweep / cascade) rather than silent future work — consistent with the strand-window decision.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Done in \`f5c6b98d5\` — added \`BarrierContext(TypedDict)\` next to \`CallbackDescriptor\` in barrier.py (execution_id / batch_index / callback_descriptor) and typed \`_dispatch_header_pg\`'s built dict, \`run_batch_with_barrier(barrier_context: BarrierContext, ...)\`, and the \`_barrier_context\` params in file_processing/tasks.py.

workers/queue_backend/barrier.py
muhammad-ali-e: Fixed in \`f5c6b98d5\` — renamed the field \`backend\` → \`transport\` (it holds a WorkflowTransport value like \`"pg_queue"\`, not a QueueBackend \`"pg"\`), with a comment calling out the distinction. Reads now via \`is_pg_transport(descriptor.get("transport"))\`, so the QueueBackend identity-compare guard in routing.py is no longer undermined.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in \`f5c6b98d5\` — replaced the born-stale parenthetical with "Markers are written by claim_batch() on the in-body PG path." Also refreshed the now-stale "Inert in 9e PR 2b — wired in PR 2c" notes on \`claim_batch\`/\`clear_execution_batches\` to describe their actual callers.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in \`f5c6b98d5\` — the \`_dispatch_header_pg\` docstring now says \`kwargs={batch_files, batch_index, total_batches}\` (the body reads \`task.kwargs\`, treats \`task.args\` as empty), matching how orchestration_utils builds the signature.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Done in \`f5c6b98d5\` — extracted module-level \`_dispatch_pg(task_name, *, args, kwargs, queue, fairness=None)\` that owns the single cycle-avoidance local import + \`backend=QueueBackend.PG\`. Both \`_dispatch_header_pg\` and \`_fire_barrier_callback\` now route through it.

workers/shared/workflow/execution/orchestration_utils.py
muhammad-ali-e: Done in \`f5c6b98d5\` — added \`is_pg_transport(transport)\` next to \`WorkflowTransport\`/\`normalize_transport\` in core data_models, and replaced the \`== WorkflowTransport.PG_QUEUE.value\` literals in orchestration_utils + pg_barrier with it. Single seam for "what counts as PG transport".

workers/general/tasks.py
muhammad-ali-e: Fixed in \`f5c6b98d5\` — added \`normalize_transport(transport, logger=logger)\` at the top of \`_execute_general_workflow\` (the point feeding both WorkflowContextData and the fan-out), so a typo'd transport degrades to Celery **with a warning**, matching the api/scheduler workers. (Placed there rather than the task entry since two entry points share that body.)

workers/tests/test_pg_barrier.py
muhammad-ali-e: Added in \`f5c6b98d5\`. \`test_last_batch_self_chains_callback_to_pg_and_cleans_up\` seeds remaining=1 with a \`transport="pg_queue"\` descriptor, patches dispatch, calls \`run_batch_with_barrier\`, and asserts the callback self-chains onto PG (\`backend=PG\`, \`args=[[result]]\`) + barrier row deleted + dedup cleared. Also extended the header test to assert \`args\`/\`queue\`/pre-existing-\`kwargs\` survive the Signature→dispatch unpacking.

workers/tests/test_pg_barrier.py
muhammad-ali-e: Added in \`f5c6b98d5\`: success-path dedup-clear (in the test above), a PG-branch mid-loop dispatch-failure test (\`test_pg_mid_loop_dispatch_failure_deletes_row\`), and the decrement-failure abort. On the claim race — exactly-one-decrement follows from exactly-one-*claim*, which is already covered by the two-real-connection race in \`test_pg_batch_dedup.py::test_concurrent_claims_resolve_to_exactly_one_winner\` (\`run_batch_with_barrier\` decrements iff \`claim_batch\` returned True). Double-abort dedup is covered by \`test_concurrent_aborts_deduplicate\`.

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Fixed in `f5c6b98d5` — `_dispatch_header_pg` now logs a WARNING when `task.options.get("queue")` is `None`, naming the exec/batch and that it would fall back to the default PG queue (so the intended-queue consumer would miss it). Latent today (the fan-out always sets a queue), but it now fails loudly instead of silently.

muhammad-ali-e:
## Review round addressed — `f5c6b98d5`

Thanks @muhammad-ali-e — excellent review, especially the strand-on-failure analysis. All 15 handled.

**Decision on the strand windows (Critical #69 + High #74 + High #81):** reaper-as-safety-net (confirmed with the author). Fixed everything *catchable* in-process; the *un-catchable* windows are now documented as a **hard reaper dependency for PR 3** rather than "future work". 2c is gated off (`resolve_transport`→celery), so none of these can bite until the reaper recovery net exists.

**Failure handling**
- **#69** — `run_batch_with_barrier` wraps **both** work + decrement in the abort, so a decrement-side failure (guard / DB / last-batch callback dispatch) tears down in-body instead of stranding. + test.
- **#79** — `_abort_barrier_in_body` logs *inside* the suppression (no more false "torn down").
- **#74 / #81** — documented the SIGKILL-mid-work and post-commit-callback-dispatch windows as a hard reaper dependency for PR 3.
- **#86** — finalise cleanup split into two independent try/excepts with distinct logs; the dedup leak documented as a PR-3 reaper dependency.

**Typing / clarity**
- **#01** `BarrierContext(TypedDict)`; **#03** renamed `backend`→`transport` (collided with `QueueBackend`); **#27** `is_pg_transport()` predicate; **#20** extracted `_dispatch_pg()`; **#35** `normalize_transport()` at the general worker entry; **#94** log on missing header queue; **#09/#13** born-stale comment + `kwargs`-not-`args` docstring.

**Tests (#37/#41)** — last-batch self-chains callback to PG + cleans up barrier/dedup; decrement-failure aborts; PG-branch mid-loop dispatch-failure deletes row; header `args`/`queue`/pre-existing-`kwargs` preservation. (Claim race already covered at the `claim_batch` level — `run_batch_with_barrier` decrements iff the claim won.)

**Verification:** 137 barrier/dedup/routing tests green; bootstrap clean under `WORKER_BARRIER_BACKEND=pg`; ruff clean. Still inert — net behaviour on the default path: NONE.

The reaper-recovery dependency (sweep stranded barriers → mark ERROR / re-drive, reclaim `pg_batch_dedup`) is now an explicit gate on enabling `pg_queue_execution_enabled` in PR 3.

muhammad-ali-e (commented):
Re-review of the post-hardening commit (f5c6b98d). The fire-and-forget PG coordination is sound and the prior pass closed the major strand windows. After verifying against the at-least-once consumer semantics, I'm raising only one new item — a self-contradiction in the strand-window contract docstring that gates the PR-3 reaper spec. (I checked a candidate `claim_batch`-strand concern and confirmed it's a non-issue: the consumer redelivers a raised task on vt-expiry and the dedup marker isn't committed on a claim failure, so keeping `claim_batch` outside the in-body teardown is correct — aborting there would wrongly fail the whole execution on a transient blip.)

workers/queue_backend/pg_barrier.py
```diff
@@ -510,8 +659,109 @@ def barrier_pg_abort(
...
+
+    **Strand windows the reaper must cover (NOT catchable here — hard merge
+    dependency for enabling the flag in PR 3, see the module / PgBatchDedup docs):**
+
+    - A hard crash / SIGKILL / visibility-timeout expiry *during* ``work_fn()``
+      leaves the dedup marker committed (claim is step 1), so redelivery returns
+      ``skipped_redelivery`` and never re-runs the batch — the barrier hangs to
+      ``expires_at``. (Claiming before the work avoids re-processing on the common
+      redelivery case at the cost of this crash window.)
+    - A failure *after* ``remaining`` hits 0 and the callback is committed but
```
muhammad-ali-e: **[Contract/doc] This strand-window bullet is filed under "NOT catchable here" but actually describes the in-body *catchable* path — and is internally inconsistent.**

The bullet lives under the *"Strand windows the reaper must cover (NOT catchable here)"* heading, yet it says `the abort here removes the row` — an in-body, catchable action. Paragraph 1 of this same docstring already lists *"the last-batch callback dispatch failing"* among the **catchable** failures that `_abort_barrier_in_body` tears down and re-raises, so the same window is documented as both catchable and not-catchable.

It's also self-contradictory: `the callback is committed but before/at the callback enqueue` — on the PG path `_fire_barrier_callback` *is* the enqueue (a single `_dispatch_pg`), so "committed … without the callback having run" can't both hold.

The genuinely un-catchable window this bullet should describe is a hard crash / SIGKILL / vt-expiry **between the decrement committing (`remaining`→0) and the callback enqueue completing**: the decrement is committed (redelivery blocked by the dedup marker) but no in-body abort runs (the process is gone), so the `pg_barrier_state` row survives to `expires_at` and only the reaper recovers it. Suggest rewriting the bullet to that, and dropping the mention of a *software* callback-dispatch failure here since paragraph 1 already (correctly) covers it as caught + aborted.

This matters because this list is the explicit spec for what the reaper must cover before PR 3 flips `pg_queue_execution_enabled` — a maintainer building the reaper could either over-cover the already-handled catchable case or mis-scope the real crash window.

muhammad-ali-e: Good catch — fixed in `bae6a9f87`. You're right on both counts: the bullet was under "NOT catchable" yet described the in-body abort, and a *software* callback-dispatch failure is already the catchable case (step 3's wrap tears it down + re-raises). Rewrote the bullet to the genuinely un-catchable window — a hard crash **between the decrement committing (`remaining`→0) and the callback enqueue completing**: decrement committed (redelivery blocked by the marker), process gone before the enqueue or any in-body abort, so the `pg_barrier_state` row survives to `expires_at` and only the reaper recovers it. Added an explicit note that a software dispatch failure is NOT this window (it's caught above), so the list stays an accurate spec for the PR-3 reaper-recovery dependency.

muhammad-ali-e:
## @greptile-apps — both findings fixed (`a2d7ba8c9`)

Thanks for the 4/5. Both were real (in the gated PG path):

**Issue 1 — dedup markers orphaned on mid-loop PG dispatch failure.** Correct: with the barrier row deleted, an earlier header's in-flight `barrier_pg_abort` returns `already_aborted` and never reaches the `clear_execution_batches` inside it. Fixed — the mid-loop handler now calls `clear_execution_batches(execution_id)` directly on the PG path after `_delete_barrier`. + test asserting an already-claimed marker is reclaimed (not orphaned).

**Issue 2 — PG callback dropped fairness headers.** Correct — the callback was always dispatched at default priority, diverging from the Celery path. Went with your "extend the descriptor to carry the FairnessKey value" suggestion via a small `_fairness_from_headers()` that reconstructs the `FairnessKey` from the stored `x-fairness-key` headers and passes it to `_dispatch_pg`, so the PG callback rides the producer's org/priority. + tests for the fairness-carried and no-fairness (None-safe) paths.

75 barrier/dedup tests green; ruff + ruff-format clean; bootstrap clean under `WORKER_BARRIER_BACKEND=pg`. Still gated/inert by default.

muhammad-ali-e:
## SonarCloud S3776 fixed — `90c61ce02`

`PgBarrier.enqueue` cognitive complexity 17 → under 15. Extracted the per-header dispatch loop into `PgBarrier._dispatch_headers` — the deeply-nested `for → try/except → if/else → if` (PG-vs-celery branch + mid-loop teardown + PG dedup-clear) was the driver; that nesting was added across the last two review rounds. Behaviour identical, just relocated. radon: `enqueue` C(11)→B(6); ruff C901 passes at the 15 threshold; 75 barrier/dedup tests green.

muhammad-ali-e:
## @greptile-apps — "test passes for the wrong reason" fixed (`9e078fb94`)

Spot on. The pre-seeded `claim_batch("exec-midfail", 0)` marker was wiped by enqueue's UPSERT block (the reuse-reset `DELETE FROM pg_batch_dedup`) **before** the dispatch loop ran, so the mid-loop `clear_execution_batches` deleted 0 rows — the `count == 0` assertion passed on the UPSERT, not the guard under test.

Fixed per your suggestion: the **first** `dispatch` side-effect now calls `claim_batch("exec-midfail", 0)` before returning its handle — inserting the marker *after* the UPSERT cleared stale ones (a fast PG consumer claiming its batch). The second dispatch fails, and the mid-loop `clear_execution_batches` is now what removes the marker.

Verified it's load-bearing: with the mid-loop clear no-op'd the marker **orphans (count=1)**; with it enabled, count=0. 57 barrier/dedup tests green, ruff clean.

PR #2070, opened 2026-06-17
UN-3564 [FEAT] PG Queue 9e — reaper-recovery: mark stranded executions ERROR (gates PR 3)

muhammad-ali-e:
## What

The reaper today only `DELETE`s expired `pg_barrier_state` rows — a stranded execution then **vanishes silently** (stuck in `EXECUTING`). This makes the reaper **recover** stranded executions. It's the hard dependency surfaced in **PR 2c review** (#2069) for enabling the PG transport: the un-catchable strand windows (a worker SIGKILL mid-batch; a crash after the final decrement but before the callback enqueues) otherwise bottom out at the ~6h barrier expiry.

## How it works

For each `pg_barrier_state` row past `expires_at`, the leader (`recover_expired_barriers`), **best-effort + per-execution**:

1. **Marks the execution ERROR via the internal API** — the same path the normal callback uses for terminal status (business state never goes direct-DB). The message distinguishes `remaining>0` (work incomplete) from `remaining==0` (all batches done, callback never fired). It **reads status first and skips the mark if the execution is already terminal** — a `remaining==0` row can belong to a COMPLETED execution whose best-effort row-delete merely failed, and the backend status update has **no terminal guard**, so a blind mark would corrupt it — or if the row carries no org.
2. **Reclaims the queue-infra rows** (`pg_batch_dedup` + `pg_barrier_state`) directly in PG — same boundary as the rest of `queue_backend`.

Recover-**then**-delete: a failure (e.g. the API unreachable) leaves that barrier row for the next sweep to retry, and never blocks the others (single-leader → no double-claim).

## Why the internal API for the ERROR mark (not direct DB)

`queue_backend` touches only **queue-infra** tables directly; **execution status is business state** and is *always* transitioned via the internal API (the normal callback does exactly this). The API path is **functional** — it computes `execution_time`, truncates the error, increments `attempts`, and is the events/notifications/cache seam — and enforces the multi-tenant boundary. A direct `UPDATE` would bypass all of that.

## Changes

- **Backend** — `organization_id` column on `PgBarrierState` + migration 0007 (the reaper needs it for the org-scoped status API; it has only `execution_id` off the row otherwise).
- **Workers** — `PgBarrier.enqueue` stamps `organization_id` into the barrier UPSERT; `PgReaper` holds a lazily-built `InternalAPIClient`; `sweep_expired_barriers` → `recover_expired_barriers`.

## Tests / dev-test

- Reaper suite reworked: real-PG recovery with a fake API client covering mark-ERROR (`remaining>0` and `==0` messages), **terminal-skip (no overwrite)**, org-missing skip, **API-failure-leaves-row-for-retry**, dedup reclaim, tick-via-real-conn — plus the existing leadership / lease / heartbeat tests. **109** reaper/barrier/dedup tests green; `ruff` clean.
- **Dev-test vs the real `InternalAPIClient` + backend:** a stranded `PENDING` execution was marked **ERROR** + cleaned up; an already-`COMPLETED` execution was **NOT overwritten** ("no status overwrite") + cleaned up — the terminal-guard verified end-to-end.

## Deferred (follow-up)

Callback **re-fire** for the `remaining==0` strand (heal → COMPLETED instead of ERROR) needs `callback_descriptor` stored on the barrier row; until then those strands are marked ERROR. Per-stage re-enqueue of stuck file executions is a larger pipeline-recovery effort beyond this net.

## Base


muhammad-ali-e:
## SonarCloud S6741 fixed — `2ffd732e7`
`api.update_calls[0]` (Sonar couldn't prove the list non-empty → potential IndexError) replaced with single-element unpacking `(call,) = api.update_calls` in the two recovery tests — removes the index and additionally pins *exactly one* status mark. 52 reaper tests green, ruff clean.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -111,40 +138,130 @@ def reaper_interval_from_env() -> float:
     return value
 
 
-def sweep_expired_barriers(conn: PgConnection) -> list[str]:
-    """Reclaim ``pg_barrier_state`` rows past ``expires_at``. Returns their ids.
+def _execution_status(
+    api_client: InternalAPIClient, execution_id: str, organization_id: str
+) -> str | None:
+    """Current status of an execution via the org-scoped internal API (or None)."""
+    response = api_client.get_workflow_execution(
```
muhammad-ali-e, line 141: **[Critical] Failed status read is indistinguishable from "not terminal", defeating the terminal-skip guard.**

`_execution_status` returns `getattr(response, "status", None)` and never checks `response.success`. But `InternalAPIClient.get_workflow_execution` → `execution_client.get_workflow_execution` **catches all exceptions and returns an error response** with `success=False, status=None` (it does not raise, unlike `update_workflow_execution_status`). So a transient failure (network blip, 404, 5xx-after-retries, auth) yields `status=None`, which is not in `_TERMINAL_STATUSES`, so the guard at L176-184 falls through to the `else` and marks the execution ERROR.

This is exactly the corruption the guard exists to prevent: a `remaining==0` row belonging to a COMPLETED execution whose row-delete merely failed gets flipped to ERROR — and since the barrier row is then DELETEd (L211-216), it never self-heals.

**Fix:** treat "could not read status" as a hard stop, not as "non-terminal". Have `_execution_status` raise when `not response.success` so the caller's `except` retains the row for retry:
```python
if not getattr(response, "success", False):
    raise RuntimeError(f"status read failed for {execution_id}; refusing to mark ERROR")
return getattr(response, "status", None)
```
Then only mark ERROR on a *successful* read of a non-terminal status.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -111,40 +138,130 @@ def reaper_interval_from_env() -> float:
...
+    callback uses for terminal status — business state never goes direct-DB),
+    UNLESS it's already terminal or we don't know its org. A remaining==0 expired
+    row can belong to a COMPLETED execution whose row-delete merely failed, and
+    the backend status update has no terminal guard, so we read status first and
+    skip the mark in that case. Queue-infra cleanup (``pg_batch_dedup`` /
+    ``pg_barrier_state``) stays direct-PG — same boundary as the rest of
+    ``queue_backend``. Recovery happens BEFORE the delete, so a failed mark leaves
+    the row for the next sweep to retry (the caller catches + rolls back).
+    """
+    if not organization_id:
```
muhammad-ali-e, line 240: **[Medium] A stranded execution with a missing org is silently abandoned in a non-terminal state.**

When `organization_id == ""`, this branch logs a `warning`, skips the API mark, then unconditionally DELETEs the barrier + dedup rows (L211-216). The execution's business status is never corrected (stays EXECUTING from the user's view) **and** the only recovery handle is erased, so no later sweep can retry. The org is coerced to `""` silently at write time (`pg_barrier.py:298`), with no loud signal.

**Fix:** log loudly (`logger.error`, Sentry-visible) at write time when org is absent, and consider **not** deleting the barrier row when the org is missing and status couldn't be confirmed terminal — deleting the recovery handle for an un-recoverable strand guarantees it can never be healed.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -111,40 +138,130 @@ def reaper_interval_from_env() -> float:
...
+        logger.warning(
+            "Reaper: stranded barrier for execution %s has no organization_id — "
+            "cannot mark ERROR via the org-scoped API; cleaning up the row only.",
+            execution_id,
+        )
+    elif (status := _execution_status(api_client, execution_id, organization_id)) in (
+        _TERMINAL_STATUSES
```
muhammad-ali-e: **[Low] Misleading `in (_TERMINAL_STATUSES)` parentheses.**

The parens around a single name are just grouping, not a tuple — `in (_TERMINAL_STATUSES)` is identical to `in _TERMINAL_STATUSES`, but reads like a one-element collection (the same S6741-style trap cleaned up elsewhere on this branch). Keep the walrus (status is reused at L184), just drop the fake tuple:
```python
elif (
    status := _execution_status(api_client, execution_id, organization_id)
) in _TERMINAL_STATUSES:
```

Also (**Medium**): `_TERMINAL_STATUSES` duplicates the canonical `ExecutionStatus.is_completed()` (`unstract/core/.../data_models.py`), which already returns True for exactly `{COMPLETED, STOPPED, ERROR}`. If a terminal status is ever added, `is_completed` is where it lands and this frozenset silently drifts — re-introducing the very corruption it guards against. Prefer `ExecutionStatus.is_completed(status)` (guard `status is not None` first), or add a test asserting the two stay in sync.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -111,40 +138,130 @@ def reaper_interval_from_env() -> float:
...
+    ):
+        logger.warning(
+            "Reaper: barrier for execution %s expired but the execution is already "
+            "%s — cleaning up the orphaned row only (no status overwrite).",
+            execution_id,
+            status,
+        )
+    else:
+        if remaining is not None and remaining > 0:
```
muhammad-ali-e: **[Medium] `remaining: int | None` contradicts the schema, and the negative case is mis-messaged.**

`pg_barrier_state.remaining` is a NOT-NULL `IntegerField`, so the `SELECT ... remaining` can never yield `None` — the `int | None` hint (L156) and the `is not None` guard advertise a state that can't occur. More importantly, `remaining` can legitimately be **negative** (documented at `models.py`: a decrement after teardown drives it `< 0`, and the CheckConstraint deliberately omits `>= 0`). A `remaining < 0` row falls into the `else` and is reported to the user as *"all file batches completed but the callback never fired"*, which is inaccurate for an already-torn-down barrier.

**Fix:** change the signature to `remaining: int` and branch three ways:
```python
if remaining > 0:    # work incomplete
    ...
elif remaining == 0: # all done, callback never fired
    ...
else:                # remaining < 0: barrier already torn down
    ...
```

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -111,40 +138,130 @@ def reaper_interval_from_env() -> float:
...
-    statement deleted), so it stays correct even if leadership gating ever fails —
-    in practice only the leader calls it. Each reclaimed barrier is logged at
-    WARNING: an orphaned barrier means an execution's header tasks never all
-    completed, worth surfacing even though deleting the row is the right backstop.
+    # Queue-infra cleanup (direct PG). Delete dedup markers first, then the
+    # barrier row — both in one committed txn after recovery succeeded.
+    with conn.cursor() as cur:
+        cur.execute("DELETE FROM pg_batch_dedup WHERE execution_id = %s", (execution_id,))
+        cur.execute(
+            "DELETE FROM pg_barrier_state WHERE execution_id = %s", (execution_id,)
```
muhammad-ali-e: **[Medium] The barrier-row DELETE is not re-guarded on `expires_at` — a re-armed barrier can be torn down.**

`recover_expired_barriers` does `SELECT ... WHERE expires_at < now()`, commits, then per-row runs this `DELETE ... WHERE execution_id = %s` with no expiry predicate. The enqueue path UPSERTs the same `execution_id`, resetting `expires_at` to the future and `remaining` (`pg_barrier.py:305-317`). If an execution is re-enqueued with the same id during a recovery cycle, the reaper deletes the freshly re-armed barrier (and its dedup markers), stranding the new run. Leader election + the ~6h expiry make this unlikely, but it's a genuine read-then-act gap.

**Fix:** make the delete idempotent against re-arm:
```sql
DELETE FROM pg_barrier_state WHERE execution_id = %s AND expires_at < now()
```
and guard the dedup delete similarly (or only when the barrier row was actually deleted; `cur.rowcount` lets you skip if it was re-armed).

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -244,7 +375,9 @@ def tick(self) -> TickOutcome:
         if not self._is_leader:
             return TickOutcome(was_leader=False, reclaimed=0)
         try:
-            reclaimed = len(sweep_expired_barriers(self._get_sweep_conn()))
+            reclaimed = len(
+                recover_expired_barriers(self._get_sweep_conn(), self._get_api_client())
+            )
         except Exception:
             self._discard_owned_sweep_conn()
             raise
```
muhammad-ali-e, line 490: **[Low] No aggregate signal for recovery success or systemic failure.**

`tick()` computes `reclaimed` and wraps it in `TickOutcome`, but `run()` discards the return (L400). Per-row failures are logged individually (good), but there's no "recovered N this cycle" log and no escalation when an entire non-empty sweep fails (API persistently down, bad migration) — that systemic pattern is buried in a stream of identical per-row exceptions.

**Fix:** in `run()`/`tick()`, on a leader sweep log `recovered` count when `> 0`, and have `recover_expired_barriers` track failures (rows seen − recovered) and emit a `logger.error` when failures occur — especially when *all* rows in a non-empty sweep failed.

muhammad-ali-e (commented):
Automated PR Review Toolkit pass (Code Reviewer, Silent-Failure Hunter, Type-Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier). Findings below are grouped by severity. The headline item is the status-read-failure path on `_execution_status` (flagged independently by two agents): it converts the reaper's central safety guarantee into one that only holds while the internal API is perfectly healthy.

workers/queue_backend/pg_barrier.py
```diff
@@ -300,13 +305,15 @@ def enqueue(
                 # periodic sweep keyed on pg_barrier_expires_idx.
```
muhammad-ali-e, line 313: **[Low] Stale comment: the "(future) periodic sweep" now exists.**

This comment says orphan reclaim "is a separate (future) periodic sweep keyed on `pg_barrier_expires_idx`" — but as of this PR that's exactly what the leader-elected reaper (`recover_expired_barriers`) implements. The PR edits the adjacent INSERT, so worth updating here rather than leaving a freshly-contradicted comment next to new code, e.g. "…handled out-of-band by the leader-elected reaper (`recover_expired_barriers`) keyed on `pg_barrier_expires_idx`."

workers/tests/test_pg_barrier.py
```diff
@@ -191,8 +191,10 @@ def test_upsert_overwrites_stale_state(self, barrier_db):
         with barrier_db.cursor() as cur:
             cur.execute(
                 "INSERT INTO pg_barrier_state "
-                "(execution_id, remaining, results, created_at, expires_at) "
-                "VALUES ('exec-R', 1, '[1,2]'::jsonb, now(), now() + interval '1h')"
+                "(execution_id, organization_id, remaining, results, "
+                " created_at, expires_at) "
+                "VALUES ('exec-R', '', 1, '[1,2]'::jsonb, now(), "
```
muhammad-ali-e, line 206: **[Test — high value] `organization_id` stamping is never asserted.**

This PR adds `organization_id` to the barrier UPSERT (the entire reason the column + migration exist), but no test reads it back — `_row` selects only `remaining, results`. A regression (dropped column, wrong kwarg key `org_id` vs `organization_id`, broken `or ""`, or a failing `ON CONFLICT DO UPDATE SET organization_id`) would silently write `""`, which makes the reaper skip the ERROR-mark for every stranded execution. Add to `TestPgBarrierEnqueue`:
- `test_enqueue_stamps_organization_id` → enqueue with `organization_id="org-42"`, assert the row's column equals it.
- `test_enqueue_defaults_org_to_empty_when_absent` → assert `""` when absent.
- `test_upsert_refreshes_org_on_reenqueue` → seed `old-org`, re-enqueue with `new-org`, assert refreshed (covers the `ON CONFLICT` clause, currently untested).

workers/tests/test_pg_reaper.py
```diff
@@ -179,26 +184,59 @@ def boom():
...
+class _FakeApiClient:
+    """Records the reaper's status calls; returns a configurable execution status."""
+
+    def __init__(self, status="EXECUTING", *, fail_update=False):
+        self._status = status
+        self._fail_update = fail_update
+        self.get_calls: list = []
+        self.update_calls: list = []
+
+    def get_workflow_execution(
```
muhammad-ali-e, line 208: **[Test] `_FakeApiClient` doesn't model the real `get_workflow_execution` failure contract.**

The fake always returns `SimpleNamespace(status=self._status)` and can only succeed — it never models a failed read (`success=False, status=None`), which is what the real client returns on any error. Because of this the suite cannot catch the Critical read-failure → false-ERROR path. After fixing `_execution_status`, add a fake mode that returns a failed read and assert no ERROR mark + row retention.

Also: the fake records only `(execution_id, organization_id)` and drops `file_execution`. Production always passes `file_execution=False` (L147) — record and assert it so a regression flipping it to `True` is caught.

workers/tests/test_pg_reaper.py
```diff
@@ -298,28 +339,88 @@ def _ids(conn):
...
+
+    def test_skips_status_mark_when_org_missing(self, barrier_conn):
+        _seed(barrier_conn, "exp-noorg", expired=True, organization_id="")
+        api = _FakeApiClient()
+        recovered = recover_expired_barriers(barrier_conn, api)
+        assert recovered == ["exp-noorg"]
+        assert api.get_calls == [] and api.update_calls == []  # can't call org API
+        assert _ids(barrier_conn) == []  # still cleaned up
+
+    def test_api_failure_leaves_row_for_retry(self, barrier_conn):
```
muhammad-ali-e, line 430: **[Test] Per-execution isolation is only tested with a single failing row.**

The load-bearing claim is that a mid-loop failure rolls back and the loop continues (`reaper.py:249-263`). This test covers one failing row; there's no test with **multiple expired rows where one fails and the others still recover**. A regression that let the first failure abort the sweep (or left the connection in an aborted-txn state poisoning later rows) wouldn't be caught.

**Add** `test_one_failing_execution_does_not_block_others`: seed three expired rows, make `_FakeApiClient` fail the update for exactly one id; assert the other two are in `recovered` and deleted, while the failing row remains.

workers/tests/test_pg_reaper.py
```diff
@@ -258,35 +300,34 @@ def barrier_conn():
...
 
 
-def _seed(conn, execution_id, *, expired):
+def _seed(conn, execution_id, *, expired, organization_id="org-1", remaining=1):
     # created_at must precede expires_at (CheckConstraint
     # pg_barrier_expires_after_created). Commit so the seed is durable like a
     # real barrier row (written by PgBarrier in another transaction) — and so the
-    # manual-commit sweep's own commit() is what persists the DELETE.
+    # manual-commit recovery's own commit() is what persists the DELETE.
+    interval = (
```
muhammad-ali-e: **[Low — test clarity] Unpack the `interval` tuple instead of using magic indices.**

`interval[0]`/`interval[1]` in the INSERT are positional — you have to scroll up to learn `[0]`=created, `[1]`=expires. Unpacking names them at assignment:
```python
created_sql, expires_sql = (
    ("now() - interval '2 hours'", "now() - interval '1 hour'")
    if expired
    else ("now()", "now() + interval '6 hours'")
)
...
f"VALUES (%s, %s, %s, '[]'::jsonb, {created_sql}, {expires_sql})",
```
Behavior-identical; finishes the consolidation cleanup this PR started.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed in \`3800ef1f4\` — and verified your read of the contract: \`get_workflow_execution\` catches all and returns \`success=False\` (never raises). \`_execution_status\` now **raises** when \`not response.success\`, so a transient read failure is a hard stop (the caller's \`except\` retains the row for the next sweep) — it can no longer fall through to mark a possibly-COMPLETED execution ERROR. + test \`test_failed_status_read_does_not_mark_and_retains_row\`, and \`_FakeApiClient\` now models the success=False/status=None contract.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Addressed in \`3800ef1f4\`. org-missing now **leaves the row** (doesn't erase the only recovery handle) and logs at ERROR; \`_recover_one_barrier\` returns False so it's not counted recovered. Added a loud **write-time** ERROR in \`PgBarrier.enqueue\` when a barrier is enqueued without an org (should never happen — every fan-out passes it). Test updated: \`test_org_missing_leaves_row_and_skips_mark\` asserts the row is preserved + no mark.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Both fixed in \`3800ef1f4\` — dropped the fake \`in (_TERMINAL_STATUSES)\` parens, and replaced the local frozenset entirely with \`ExecutionStatus.is_completed(status)\` (single source of truth — no drift if a terminal status is ever added), guarded on \`status is not None\`.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed in \`3800ef1f4\` — \`remaining\` is now typed \`int\` (NOT-NULL column) and the message is three-way: \`>0\` work-incomplete, \`==0\` callback-never-fired, \`<0\` "barrier already torn down (remaining<0) yet execution left non-terminal — inconsistent state". Extracted into \`_mark_stranded_error\`.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed in \`3800ef1f4\` — the barrier DELETE is now \`WHERE execution_id = %s AND expires_at < now()\`, and the dedup DELETE only runs when the barrier row was actually reclaimed (\`cur.rowcount > 0\`); a re-armed barrier (UPSERT reset expires_at to the future) is left for the new run + logged. Residual: a re-arm between the SELECT and the (pre-delete) status-read could still mark the re-run ERROR, but the new run's barrier survives the guarded delete, so its callback re-marks it COMPLETED — self-correcting. Bounded by leader-election + the ~6h expiry.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Done in \`3800ef1f4\` — \`recover_expired_barriers\` now emits an aggregate summary per non-empty sweep: INFO "recovered N", WARNING "recovered N/M; K left for retry", and **logger.error "recovered NONE of M"** when a non-empty sweep recovers nothing (systemic — API down / bad migration / missing org).

workers/queue_backend/pg_barrier.py
muhammad-ali-e: Updated in \`3800ef1f4\` — the comment now points at the leader-elected reaper (\`recover_expired_barriers\`) keyed on \`pg_barrier_expires_idx\` instead of calling it "(future)".

workers/tests/test_pg_barrier.py
muhammad-ali-e: Added in \`3800ef1f4\` (\`+_org\` reader): \`test_enqueue_stamps_organization_id\`, \`test_enqueue_defaults_org_to_empty_when_absent\`, and \`test_upsert_refreshes_org_on_reenqueue\` (covers the previously-untested \`ON CONFLICT DO UPDATE SET organization_id\`).

workers/tests/test_pg_reaper.py
muhammad-ali-e: Done in \`3800ef1f4\` — \`_FakeApiClient\` now returns \`success=True/False\` (a \`fail_read\` mode models the real error contract) and records \`file_execution\`. + \`test_failed_status_read_does_not_mark_and_retains_row\` (the Critical path) and \`test_status_read_passes_file_execution_false\`.

workers/tests/test_pg_reaper.py
muhammad-ali-e: Added \`test_one_failing_execution_does_not_block_others\` in \`3800ef1f4\` (\`_FakeApiClient(fail_update_for=["exp-bad"])\`): three expired rows, one update fails — asserts the other two are recovered + deleted and only the failing row remains.

workers/tests/test_pg_reaper.py
muhammad-ali-e: Done in \`3800ef1f4\` — unpacked to \`created_sql, expires_sql = ...\`, no more positional \`interval[0]/[1]\`.

muhammad-ali-e:
## Review round addressed — `3800ef1f4`

Thanks @muhammad-ali-e — the Critical one was a genuine corruption hole. All 11 handled.

**[Critical] read-failure → false-ERROR.** `get_workflow_execution` returns `success=False` on any blip (never raises), so a failed read read as "non-terminal" and would flip a COMPLETED execution to ERROR. `_execution_status` now **raises on `not response.success`** → the row is retained for the next sweep, never marked. (+ test, + `_FakeApiClient` now models the real contract.)

**Mediums**
- `ExecutionStatus.is_completed` (single source of truth) replaces the local `_TERMINAL_STATUSES` frozenset — no drift; fake parens dropped.
- `remaining` typed `int` + three-way message (`>0` / `==0` / `<0` already-torn-down).
- barrier DELETE re-guarded on `expires_at < now()` + dedup-delete only when the row was reclaimed — a same-id re-enqueue isn't torn down mid-recovery.
- org-missing now **leaves the row** (preserves the recovery handle) + ERROR log, plus a loud **write-time** ERROR in `enqueue`.

**Lows**
- aggregate sweep summary + `logger.error` when a non-empty sweep recovers nothing (systemic signal).
- refreshed the stale "(future) periodic sweep" comment.

**Tests** (+org-stamping ×3 incl. the `ON CONFLICT` refresh; +read-failure-retains-row; +`file_execution=False`; +multi-row isolation). 115 reaper/barrier/dedup tests green; ruff + ruff-format clean.

muhammad-ali-e:
## SonarCloud S6741 (new instance) fixed — `93a276e58`
Different occurrence from the earlier `update_calls[0]` fix — this one (`api.get_calls[0][2]`) was introduced by the prior review round's new `test_status_read_passes_file_execution_false`. Replaced with index-free unpacking `[(_exec_id, _org, file_execution)] = api.get_calls` (also pins exactly one read). 55 reaper tests green, ruff clean.

muhammad-ali-e:
## @greptile-apps — all three addressed (`f6c84ba45`)

Thanks for the 3/5. All in the recovery path:

**Re-arm race (main).** Correct — the ERROR mark fired before the `expires_at` guard, so a re-enqueued *live* run could be marked ERROR. Fixed: the mark now fires only after a re-check that the row is **still expired** (`_still_expired`) immediately before it — a same-id re-enqueue between the sweep SELECT and the mark is detected and skipped (its new run owns the barrier). The DELETE stays guarded on `expires_at < now()` as the second line of defence. Added `test_rearmed_execution_is_not_marked_error` (re-arm simulated via the status-read side-effect): asserts no mark + the re-armed row left intact. *(Residual: a microsecond TOCTOU between the re-check and the mark remains, but it requires the same `execution_id` re-enqueued in that window for a barrier that's already 6h-expired — and even then the guarded DELETE leaves the new run's barrier, so its callback re-marks COMPLETED. Self-correcting.)*

**`status=None` on `success=True`.** No longer falls through to mark ERROR — treated as indeterminate: skip + leave the row for the next sweep. + test.

**org-missing inflating the systemic error.** The aggregate now distinguishes genuine **failures** (exceptions) from benign **skips** (terminal / re-armed / no-status / no-org). The "recovered NONE / API down" ERROR escalates only on real failures; an all-skipped sweep doesn't. + caplog test.

58 reaper + 51 barrier tests green; ruff + ruff-format clean.

muhammad-ali-e:
## SonarCloud S125 fixed — `926df20a0`
The flagged comment ended with a parenthesised tuple `(exec_id, org, file_execution)` — a valid Python expression S125 parses as commented-out code. Reworded to prose; also dropped a stale `sweep_expired_barriers` reference and a couple of inline `foo()` prose mentions to remove other code-like patterns. Comment-only change; tests unaffected.

PR #2071, opened 2026-06-18
UN-3570 [FEAT] PG Queue 9e PR 3 — Flipt canary wiring for transport resolution

muhammad-ali-e:
## What

- `resolve_transport()` now makes a real decision instead of always returning Celery: an **env master-gate** (`PG_QUEUE_TRANSPORT_ENABLED`, default off) → when on, a **Flipt boolean flag** (`pg_queue_execution_enabled`) → fail-closed to Celery on any error.
- Wired at **both** execution-creation chokepoints: `internal_api_views.create_workflow_execution` (scheduler path) and `workflow_helper.execute_workflow_async` (API / manual / async path).
- New setting `PG_QUEUE_TRANSPORT_ENABLED` + a `sample.env` entry.

## Why

- This is the on/off switch for the 9e PG-queue pipeline. PR 1 left `resolve_transport` hardwired to Celery; this PR lets a **new** execution be canary-routed onto Postgres while everything else stays on Celery. Part of PG Queue Phase 9 (UN-3536), sub-task UN-3570.

## How

- **Master-gate** (env, default off): until ops turns it on, Flipt is never consulted and every execution rides Celery. Doubles as the instant kill-switch and the deploy-ordering safety (inert until PG consumers are deployed).
- **entity_id = execution_id** → per-execution percentage-rollout stickiness. The decision is resolved once at creation and carried in the task payload, so an in-flight execution never re-buckets.
- **context** carries org / workflow / pipeline ids for Flipt segment rules. All values are `str`-coerced: callers pass UUID objects, and a non-string value in Flipt's gRPC `map<string,string>` context is swallowed by the client as `False`, silently forcing Celery (found and fixed during dev-test).
- Decision mirrors `normalize_transport` on the read side: fail-closed.

- **No.** Behaviour is byte-identical to today while `PG_QUEUE_TRANSPORT_ENABLED` is off (the default) — the gate returns Celery before Flipt is even consulted. Even with the gate on, the Flipt flag defaults to off, and any Flipt error falls back to Celery. The flag must not be enabled in production until PG consumers are deployed (deploy-ordering safety, enforced by the default-off env gate).

## Database Migrations

- None.

## Env Config

- New: `PG_QUEUE_TRANSPORT_ENABLED` (default `False`). When not `true`, transport resolution never consults Flipt and every execution rides Celery — kill-switch + deploy-ordering safety.
- Existing `FLIPT_SERVICE_AVAILABLE` / `EVALUATION_SERVER_IP` / `EVALUATION_SERVER_PORT` already gate Flipt availability (client fails closed to `False`).

## Notes on Testing

- Unit: 15 cases in `test_transport.py` — gate off (Flipt not consulted), gate on + flag true → pg_queue, flag false → celery, Flipt exception → fail-closed, entity_id/context shape, optional-id omission, and a UUID-coercion regression test.
- Dev-tested end-to-end against a running stack:
  - Gate OFF: API deployment runs fully on Celery; `pg_barrier_state` / `pg_batch_dedup` untouched; COMPLETED.
  - Gate ON + flag at 100%: real API deployment routes through the PG consumers (PgBarrier → decrement remaining=0 → callback) to COMPLETED with clean teardown.
  - Live-Flipt decision matrix: gate off / flag on / Flipt unavailable → celery / pg_queue / celery.

## Related Issues

- UN-3536 (PG Queue Phase 9 — transport engine), sub-task UN-3570.
- Follow-ups (separate sub-tasks): 2d orchestrator-on-PG; UN-3566 multi-queue consumer ergonomics.

## Checklist

- [x] Code follows project style guidelines
- [x] Self-review performed

### DIFF
--- a/backend/sample.env
+++ b/backend/sample.env
@@ -184,9 +184,11 @@ TOOL_REGISTRY_CONFIG_PATH="/data/tool_registry_config"
 # Flipt Service
 FLIPT_SERVICE_AVAILABLE=False
 
-# 9e PG-queue transport master-gate (kill-switch). Keep False until PG queue
-# consumers are running in the fleet; only then does the Flipt flag
-# `pg_queue_execution_enabled` take effect. Set False for instant rollback.
+# 9e PG-queue transport master-gate (kill-switch). Routing an execution onto the
+# PG queue requires ALL THREE: this gate True, FLIPT_SERVICE_AVAILABLE=True (else
+# the Flipt client returns False for every flag), and the Flipt flag
+# `pg_queue_execution_enabled` on. Keep False until PG queue consumers are running
+# in the fleet; set False for instant rollback.
 PG_QUEUE_TRANSPORT_ENABLED=False
 
 # File System Configuration for Workflow and API Execution
--- a/backend/sample.env
+++ b/backend/sample.env
@@ -184,6 +184,11 @@ TOOL_REGISTRY_CONFIG_PATH="/data/tool_registry_config"
 # Flipt Service
 FLIPT_SERVICE_AVAILABLE=False
 
+# 9e PG-queue transport master-gate (kill-switch). Keep False until PG queue
+# consumers are running in the fleet; only then does the Flipt flag
+# `pg_queue_execution_enabled` take effect. Set False for instant rollback.
+PG_QUEUE_TRANSPORT_ENABLED=False
+
 # File System Configuration for Workflow and API Execution
 
 # Directory Prefixes for storing execution files
--- a/backend/workflow_manager/internal_api_views.py
+++ b/backend/workflow_manager/internal_api_views.py
@@ -323,10 +323,14 @@ def create_workflow_execution(request):
 
         # Resolve the transport this execution rides (9e). Decided once here, at
         # the creation chokepoint, and returned so the caller carries it in the
-        # dispatched task's payload — not persisted on the row. PR 1 always
-        # resolves "celery"; PR 3 wires Flipt (keyed on workflow/pipeline/org) in
-        # resolve_transport().
-        transport = resolve_transport()
+        # dispatched task's payload — not persisted on the row. Flipt (gated by
+        # the env master-switch) decides; fails closed to "celery".
+        transport = resolve_transport(
+            execution_id=str(execution.id),
+            organization_id=org_id,
+            workflow_id=str(workflow.id),
+            pipeline_id=data.get("pipeline_id"),
+        )
 
         return Response(
             {
--- a/backend/workflow_manager/workflow_v2/transport.py
+++ b/backend/workflow_manager/workflow_v2/transport.py
@@ -52,7 +52,7 @@ PG_QUEUE_FLAG_KEY = "pg_queue_execution_enabled"
 def resolve_transport(
     *,
     execution_id: str | UUID,
-    organization_id: str | UUID,
+    organization_id: str | UUID | None,
     workflow_id: str | UUID | None = None,
     pipeline_id: str | UUID | None = None,
 ) -> str:
@@ -69,8 +69,9 @@ def resolve_transport(
             execution is resolved exactly once).
         organization_id: The org's string identifier
             (``Organization.organization_id`` — the ``X-Organization-ID`` header
-            value, *not* the DB pk). Carried in the Flipt ``context`` for per-org
-            segment rollouts.
+            value, *not* the DB pk). May be ``None`` on the helper path
+            (``StateStore`` not populated) → resolves to celery. Carried in the
+            Flipt ``context`` for per-org segment rollouts.
         workflow_id: Optional, carried in ``context`` for future segment rules.
         pipeline_id: Optional, carried in ``context`` for future segment rules.
 
@@ -105,8 +106,9 @@ def resolve_transport(
     # FliptClient returns False for ALL flags when the service is marked
     # unavailable — indistinguishable from "rollout says no". Surface it loudly so
     # a blind Flipt under an ON gate doesn't masquerade as a healthy 100%-Celery
-    # canary. (Mirrors FliptClient's own FLIPT_SERVICE_AVAILABLE env read.)
-    if os.environ.get("FLIPT_SERVICE_AVAILABLE", "false").strip().lower() != "true":
+    # canary. Parse exactly as FliptClient does (``.lower()``, no ``.strip()``)
+    # so the two can never disagree on a value like ``" true"``.
+    if os.environ.get("FLIPT_SERVICE_AVAILABLE", "false").lower() != "true":
         logger.warning(
             "resolve_transport: gate ON but FLIPT_SERVICE_AVAILABLE != true "
             "(Flipt is blind) for execution %s; forcing celery",
--- a/backend/workflow_manager/workflow_v2/transport.py
+++ b/backend/workflow_manager/workflow_v2/transport.py
@@ -12,6 +12,9 @@ PR 3 (this change) replaces PR 1's hardwired Celery with a Flipt evaluation:
 
   master-gate (env) → Flipt boolean (``pg_queue_execution_enabled``) → transport
 
+Routing onto PG needs **all three** of: the env master-gate on, Flipt reachable
+(``FLIPT_SERVICE_AVAILABLE=true``), and the flag enabled for this execution.
+
 - **Master-gate** (``settings.PG_QUEUE_TRANSPORT_ENABLED``, default off): until
   ops flips it on, Flipt is never consulted and every execution rides Celery.
   This is both the instant global kill-switch *and* the deploy-ordering safety —
@@ -21,17 +24,24 @@ PR 3 (this change) replaces PR 1's hardwired Celery with a Flipt evaluation:
   never re-bucket mid-flight); ``context`` carries org/workflow/pipeline for
   segment rules. The flag contract is fixed in 9e-design §2.
 - **Fail-closed to Celery**: a Flipt outage must never break execution creation,
-  mirroring ``normalize_transport`` on the read side.
+  mirroring ``normalize_transport`` on the read side. The gate-ON path logs its
+  decision so a "gate on but still all Celery" situation (e.g. a blind Flipt)
+  is visible rather than silent.
 """
 
 from __future__ import annotations
 
 import logging
+import os
+from typing import TYPE_CHECKING
 
 from django.conf import settings
 
 from unstract.core.data_models import WorkflowTransport
 from unstract.flags.feature_flag import check_feature_flag_status
+
+if TYPE_CHECKING:
+    from uuid import UUID
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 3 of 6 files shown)


backend/workflow_manager/workflow_v2/transport.py
```diff
@@ -8,33 +8,95 @@
...
+    # False — silently forcing celery. entity_id is str-coerced for the same
+    # reason (and so the %-rollout hash is stable across str/UUID call sites).
+    context = {"organization_id": str(organization_id)}
+    if workflow_id:
+        context["workflow_id"] = str(workflow_id)
+    if pipeline_id:
+        context["pipeline_id"] = str(pipeline_id)
+
+    # Defense in depth: check_feature_flag_status already swallows errors and
+    # returns False, but the design mandates an explicit fail-closed wrap here so
```
muhammad-ali-e: **[Important — observability / comment accuracy]** Two issues in this fail-closed wrap:

1. The comment's *"(or an import-time fault)"* claim is wrong: `check_feature_flag_status` is imported at module top (line 34), so an import fault occurs at module load — long before this `try/except` is reachable. This block only covers the *call*.
2. The wrap is effectively **unreachable in production**: `check_feature_flag_status` already wraps its whole body in `try/except Exception: return False`, and `FliptClient.evaluate_boolean` does the same. So a real Flipt outage (server down, gRPC timeout, missing flag, `FLIPT_SERVICE_AVAILABLE!=true`) returns `False` → the normal `return celery` at line 102, **never** this `except`. The `logger.warning` never fires. Fail-closed *behavior* is correct, but with the master-gate ON a silent celery is indistinguishable from a legitimate "rollout says no" — a canary can be masked as healthy.

Suggest: drop the inaccurate "import-time fault" clause, and add a decision-level signal on the gate-ON path (e.g. a one-line `logger.info` of the resolved transport, and a `logger.warning` when the gate is on but `FLIPT_SERVICE_AVAILABLE!=true`) so "deliberately celery", "deliberately pg_queue", and "Flipt is blind" are distinguishable.

backend/workflow_manager/workflow_v2/transport.py
```diff
@@ -8,33 +8,95 @@
...
+    workflow_id: str | None = None,
+    pipeline_id: str | None = None,
+) -> str:
     """Resolve the transport for a new workflow execution.
 
+    Args:
+        execution_id: The execution's id. Used as the Flipt ``entity_id`` so the
+            percentage rollout buckets per execution and stays sticky (one
+            execution is resolved exactly once).
+        organization_id: Owning org's public identifier. Carried in the Flipt
```
muhammad-ali-e: **[Suggestion — doc accuracy]** "Owning org's *public identifier*" is imprecise. Both call sites pass the org's string identifier `Organization.organization_id` (the helper path's value is labelled `schema_name` elsewhere, e.g. `workflow_helper.py:532`). "public identifier" is ambiguous (could be read as the DB pk `id` or a UUID) and invites a future "fix" to `organization.id` that would silently break Flipt org-segment matching. Suggest: `organization_id: The org's string identifier (Organization.organization_id, i.e. the X-Organization-ID header value). Carried in the Flipt context for per-org segment rollouts.`

backend/workflow_manager/workflow_v2/transport.py
```diff
@@ -8,33 +8,95 @@
...
+    # Flipt — every execution rides Celery (kill-switch + deploy-ordering safety).
+    if not settings.PG_QUEUE_TRANSPORT_ENABLED:
+        return celery
+
+    # Flipt's context is a gRPC map<string,string>; callers pass UUID objects
+    # for the ids, so coerce every value to str. A non-str value makes the
+    # client's serialization fail and check_feature_flag_status swallow it as
+    # False — silently forcing celery. entity_id is str-coerced for the same
+    # reason (and so the %-rollout hash is stable across str/UUID call sites).
+    context = {"organization_id": str(organization_id)}
```
muhammad-ali-e, line 124: **[Suggestion — defensive]** On the `workflow_helper` path, `organization_id` comes from `UserContext.get_organization_identifier()` → `StateStore.get(...)`, which can be `None` if context isn't set. There's no guard, so `str(None)` ships the literal `"None"` into the Flipt segment context — a silent wrong-org bucket (not a crash). The view path is safe (header validated non-None before the call). Consider `if not organization_id: return celery` (or omit the key) so a context-less execution doesn't get mis-segmented. Worth a regression test pinning the chosen behavior.

backend/workflow_manager/workflow_v2/transport.py
```diff
@@ -8,33 +8,95 @@
...
+
+# The fixed Flipt flag contract (9e-design §2): Boolean, default false.
+PG_QUEUE_FLAG_KEY = "pg_queue_execution_enabled"
 
 
-def resolve_transport() -> str:
+def resolve_transport(
+    *,
+    execution_id: str,
+    organization_id: str,
```
muhammad-ali-e: **[Suggestion — type honesty]** The annotations declare `str`, but the body str-coerces precisely because callers pass `UUID` objects (the inline comment at 72-76 even says so; `internal_api_views.py:329` pre-coerces with `str(...)`). The contract is internally contradictory — a type checker would wrongly flag a `UUID` caller. Make it honest: `execution_id: str | UUID`, `organization_id: str | UUID`, `workflow_id: str | UUID | None`, `pipeline_id: str | UUID | None`. Keeping the coercion internal is the right layering (callers shouldn't need to know Flipt's wire types).

backend/workflow_manager/workflow_v2/transport.py
```diff
@@ -8,33 +8,95 @@
...
+    workflow_id: str | None = None,
+    pipeline_id: str | None = None,
+) -> str:
```
muhammad-ali-e, line 58: **[Suggestion — type design]** `-> str` under-specifies the closed 2-element domain. Returning `-> WorkflowTransport` makes the invariant compile-checkable (a typo'd `"postgres"` return becomes a type error instead of being silently downgraded by `normalize_transport` on the read side). The "callers put it in a JSON payload" concern doesn't block this: `WorkflowTransport(str, Enum)` already serializes to its `.value` under `json.dumps`/DRF. For an explicit wire boundary, return the enum and serialize with `.value` at the two payload sites (`internal_api_views.py:340`, `workflow_helper.py:547`).

backend/workflow_manager/internal_api_views.py
```diff
@@ -323,10 +323,14 @@ def create_workflow_execution(request):
 
         # Resolve the transport this execution rides (9e). Decided once here, at
         # the creation chokepoint, and returned so the caller carries it in the
-        # dispatched task's payload — not persisted on the row. PR 1 always
-        # resolves "celery"; PR 3 wires Flipt (keyed on workflow/pipeline/org) in
-        # resolve_transport().
-        transport = resolve_transport()
+        # dispatched task's payload — not persisted on the row. Flipt (gated by
+        # the env master-switch) decides; fails closed to "celery".
+        transport = resolve_transport(
```
muhammad-ali-e, line 328: **[Important — test gap]** Neither call site is covered by tests. Every test mocks `resolve_transport`'s dependency and calls the function directly, so a regression here — dropping `"transport"` from the response, or mis-passing args — would pass the entire suite. The scheduler/consumer depends on this `"transport"` key; a missing key silently degrades to the `normalize_transport` default. Add a view test: patch `resolve_transport` to return `"pg_queue"`, POST `create_workflow_execution`, and assert (a) `"transport": "pg_queue"` is in the response JSON and (b) `resolve_transport.assert_called_once_with(execution_id=str(execution.id), organization_id=org_id, workflow_id=str(workflow.id), pipeline_id=...)`.

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -510,17 +510,22 @@ def execute_workflow_async(
...
-            # an independent Flipt evaluation; before then, both must be funnelled
-            # through a single per-execution chokepoint (so a percentage re-roll
-            # can't split one execution across transports). Tracked for PR 3.
-            transport = resolve_transport()
+            # the same execution. entity_id is execution_id, so each execution
+            # buckets exactly once and can't split across transports even if the
+            # rollout % is re-rolled between the two sites.
+            transport = resolve_transport(
+                execution_id=execution_id,
+                organization_id=org_schema,
```
muhammad-ali-e, line 525: **[Suggestion — clarity + test gap]** `org_schema` is a misnomer: the value is the org's public `organization_id` (verified equal to the view's `org_id`), not a Postgres schema name. The call is correct, but the name invites exactly the divergence worry this PR's NOTE addresses — consider renaming the local to `organization_id`/`org_identifier`. Also: no test pins that both call sites feed a *consistent* org identifier for the same org. That invariant (both must pass `Organization.organization_id` so Flipt org-segments match across entry paths) is currently held only by convention — a focused test asserting the identifier kind each site passes would lock it down.

backend/sample.env
```diff
@@ -184,6 +184,11 @@ TOOL_REGISTRY_CONFIG_PATH="/data/tool_registry_config"
 # Flipt Service
 FLIPT_SERVICE_AVAILABLE=False
 
+# 9e PG-queue transport master-gate (kill-switch). Keep False until PG queue
+# consumers are running in the fleet; only then does the Flipt flag
+# `pg_queue_execution_enabled` take effect. Set False for instant rollback.
+PG_QUEUE_TRANSPORT_ENABLED=False
```
muhammad-ali-e, line 192: **[Suggestion — ops doc]** This comment says PG queue takes effect once the env gate is on and the Flipt flag is set, but omits that `FLIPT_SERVICE_AVAILABLE=True` is *also* required: `FliptClient` returns `False` for all flags when `FLIPT_SERVICE_AVAILABLE != "true"`. So pg_queue needs all three (env gate + service available + flag on). Worth noting here so operators don't think two switches suffice — and it pairs with the observability gap flagged in `transport.py` (gate on + Flipt unavailable = silent 100% celery).

muhammad-ali-e (commented):

The change is functionally correct and fails closed safely. The central worry — that the two call sites pass different `organization_id` values — was verified **not** a bug: `org_id` (X-Organization-ID header) and `org_schema` (`UserContext.get_organization_identifier()`) both resolve to `Organization.organization_id`, so per-org Flipt segments match consistently. Findings below are observability, type-honesty, doc-accuracy, and test-gap items — no blockers.

backend/workflow_manager/workflow_v2/transport.py
muhammad-ali-e: Fixed in b922ea453. Dropped the inaccurate "(or an import-time fault)" clause. Added gate-ON observability: a `logger.info` of the resolved transport (deliberate-celery vs deliberate-pg_queue now distinguishable per execution), plus a `logger.warning` short-circuit when the gate is ON but `FLIPT_SERVICE_AVAILABLE != true` — the blind-Flipt case no longer masquerades as a healthy 100%-celery canary. Kept the fail-closed try/except as defense-in-depth.

backend/workflow_manager/workflow_v2/transport.py
muhammad-ali-e: Fixed in b922ea453 — docstring now reads `Organization.organization_id` (the `X-Organization-ID` header value), explicitly *not* the DB pk.

backend/workflow_manager/workflow_v2/transport.py
muhammad-ali-e: Fixed in b922ea453 — `if not organization_id:` now fails closed to celery with a warning (chose fail-closed over omit-key to avoid any mis-segmentation on a context-less execution). Added regression test `test_missing_organization_forces_celery`.

backend/workflow_manager/workflow_v2/transport.py
muhammad-ali-e: Fixed in b922ea453 — widened to `execution_id: str | UUID`, `organization_id: str | UUID`, `workflow_id`/`pipeline_id`: `str | UUID | None` (`UUID` under `TYPE_CHECKING`). Coercion stays internal. This also resolves the SonarCloud type-mismatch on the UUID-coercion test.

backend/workflow_manager/workflow_v2/transport.py
muhammad-ali-e: Deferred (intentionally). transport is a string wire-value end-to-end in 9e — the task-payload JSON, `WorkflowContextData.transport: str`, and `normalize_transport` all operate on strings; `resolve_transport` produces that string contract. We only ever return `WorkflowTransport.X.value` (never a string literal), so the typo-as-"postgres" class of bug is already impossible here. Returning the enum would create a str/enum asymmetry and still need `.value` at the two payload boundaries. Happy to do it as a separate type-design pass if you want the enum at this boundary.

backend/workflow_manager/internal_api_views.py
muhammad-ali-e: Valid gap — deferring the automated view test to a focused follow-up. The call sites are Bearer-authed DRF + ORM (Organization/Workflow/WorkflowExecution) and the repo has no pytest-django, so a faithful test needs a test DB + auth fixtures (or heavy auth+ORM mocking). Note the call-site contract was exercised end-to-end in dev-test: a real `create_workflow_execution` returned `transport=pg_queue` and the pipeline ran on the PG consumers — verified, just not yet pinned by a unit test. Will file the follow-up.

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: Deferring the rename to keep PR 3 scoped: `org_schema` also appears in a second method in this file (the wrapper that delegates here via `execute_workflow_async`), so renaming only this one creates intra-file inconsistency, and renaming both pulls in code this PR does not otherwise touch. The divergence concern is already covered by the in-code NOTE (both sites pass the same `Organization.organization_id`). Happy to do a dedicated MISC cleanup renaming both occurrences.

backend/sample.env
muhammad-ali-e: Fixed in b922ea453 — the comment now states routing onto PG needs all three: this gate True, `FLIPT_SERVICE_AVAILABLE=True`, and the flag on. Mirrored in the transport.py module docstring and the new blind-Flipt warning.

muhammad-ali-e:
## Review round addressed — `b922ea453`

Thanks for the thorough pass. Summary of what changed:

**Fixed (5)**
- **Type honesty / SonarCloud** — `resolve_transport` id params widened to `str | UUID` (`UUID` under `TYPE_CHECKING`); coercion stays internal. Resolves the SonarCloud type-mismatch (1 new issue → 0).
- **Missing-org guard** — fail closed to celery (with a warning) when `organization_id` is empty, so `str(None)` never reaches the Flipt org segment. + regression test.
- **Observability** — gate-ON path now logs the resolved transport (deliberate-celery vs deliberate-pg_queue distinguishable), and warns + short-circuits to celery when the gate is ON but `FLIPT_SERVICE_AVAILABLE != true` (blind Flipt no longer hides as a healthy 100%-celery canary). Dropped the inaccurate "import-time fault" comment.
- **Docstring** — `organization_id` is `Organization.organization_id` (the `X-Organization-ID` header value), not the DB pk.
- **sample.env** — documents that PG routing needs all three: gate + `FLIPT_SERVICE_AVAILABLE` + flag.

Tests: 15 → 17 (added missing-org and gate-on-Flipt-unavailable cases). Pre-commit green.

**Deferred (3, with rationale on each thread)**
- **Return `WorkflowTransport` enum** — transport is a string wire-value throughout 9e (payload JSON / `WorkflowContextData.transport: str` / `normalize_transport`); we only ever return `.value`, so the typo class of bug is already impossible. Can do as a separate type-design pass.
- **View call-site test** — DB + Bearer-auth path, no pytest-django; the contract was exercised end-to-end in dev-test (real `create_workflow_execution` → `transport=pg_queue` → ran on PG consumers). Filing a focused follow-up.
- **`org_schema` rename** — appears in a second method here too; renaming one is inconsistent, renaming both is out of this PR's scope. Offered as a separate MISC cleanup.

backend/workflow_manager/workflow_v2/transport.py
```diff
@@ -8,33 +8,147 @@
...
+PG_QUEUE_FLAG_KEY = "pg_queue_execution_enabled"
+
+
+def resolve_transport(
+    *,
+    execution_id: str | UUID,
+    organization_id: str | UUID,
+    workflow_id: str | UUID | None = None,
+    pipeline_id: str | UUID | None = None,
+) -> str:
```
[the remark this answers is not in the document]
muhammad-ali-e: Good catch — fixed in f9aee0e6f. `organization_id` is now `str | UUID | None`: the helper path passes `UserContext.get_organization_identifier()` which can be `None`, and the `if not organization_id` guard already handles it, so the annotation should admit `None`. (`execution_id` stays non-optional — it is always set.)

backend/workflow_manager/workflow_v2/transport.py
```diff
@@ -8,33 +8,147 @@
...
+    # unavailable — indistinguishable from "rollout says no". Surface it loudly so
+    # a blind Flipt under an ON gate doesn't masquerade as a healthy 100%-Celery
+    # canary. (Mirrors FliptClient's own FLIPT_SERVICE_AVAILABLE env read.)
+    if os.environ.get("FLIPT_SERVICE_AVAILABLE", "false").strip().lower() != "true":
+        logger.warning(
+            "resolve_transport: gate ON but FLIPT_SERVICE_AVAILABLE != true "
+            "(Flipt is blind) for execution %s; forcing celery",
+            execution_id,
+        )
+        return celery
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in f9aee0e6f — dropped `.strip()` so the check parses exactly like `FliptClient.__init__` (`.lower() != "true"`). The two can no longer disagree on a whitespaced value, so the blind-Flipt warning fires iff FliptClient also treats the service as unavailable.

muhammad-ali-e:
## greptile round addressed — `f9aee0e6f`

Both P2s fixed:
- **`organization_id` None type** — annotation now `str | UUID | None` (the helper path can pass `None`; the `if not organization_id` guard already handles it).
- **`FLIPT_SERVICE_AVAILABLE` parse parity** — dropped `.strip()` to match `FliptClient` (`.lower()` only), so the two can never disagree on a value like `" true"` and the blind-Flipt warning stays in lockstep with the client.

17 tests green, pre-commit clean.

PR #2072, opened 2026-06-18
UN-3574 [FEAT] PG Queue 9e 2d — orchestrator (async_execute_bin) on PG + shared TaskPayload contract

muhammad-ali-e:
## What

- Route the **orchestrator** task (`async_execute_bin`) onto the PG queue when `transport == pg_queue`, so a pg_queue execution's **orchestrate → fan-out → barrier → callback** all run on PG (previously hybrid — orchestrator on Celery, only fan-out/callback on PG).
- Promote the PG-message wire contract (`TaskPayload`, `FairnessPayload`) into `unstract.core`; the workers re-export them.
- New **backend PG producer** (`backend/pg_queue/producer.py`) — the backend had the `pg_queue` tables but no way to enqueue to them.

## Why

- Completes the execution-pipeline migration onto PG except the executor/tool run (PR B). Part of PG Queue Phase 9 (UN-3536), sub-task UN-3574.

## How

- **Shared contract:** `TaskPayload` / `FairnessPayload` move to `unstract.core.data_models` so the backend producer and the worker consumer agree on one definition (separate codebases can't import each other). Workers re-export from their old modules → existing imports unchanged.
- **Backend producer:** enqueues a `TaskPayload` row to `pg_queue_message` via the `PgQueueMessage` ORM, mirroring the workers' `PgQueueClient.send`. UUIDs in args/kwargs are JSON-coerced (the `message` `JSONField` has no Django encoder; the worker consumer already receives string ids on the existing PG path).
- **Backend dispatch** (`execute_workflow_async`): on `pg_queue`, enqueue `async_execute_bin` to PG (`general → "celery"`, `api → "celery_api_deployments"`) instead of `celery_app.send_task`; `task_id` becomes `"pg:<msg_id>"`. The `TimeoutError` branch is guarded for the no-AsyncResult PG path.
- **Scheduler dispatch:** pass `backend=QueueBackend.PG` when `is_pg_transport(transport)` — `dispatch()` already had the per-call override (from 2a).

- **No.** Gated off by default — while the env master-gate (`PG_QUEUE_TRANSPORT_ENABLED`) or the Flipt flag is off, `transport` resolves to `celery` and both dispatch sites take the unchanged Celery path. The shared-contract move is a pure relocation with re-exports (82 workers tests green). The executor (tool run) and log workers remain on Celery.

## Database Migrations

- None.

## Env Config

- None new (uses the existing `PG_QUEUE_TRANSPORT_ENABLED` gate + Flipt flag from PR 3).

## Notes on Testing

- Unit: 5 producer tests (wire-shape row, UUID→str coercion, default queue, empty args, priority bounds). Workers regression suite (fairness / dispatch_pg / pg_queue_client / routing) green after the core move.
- Dev-tested end-to-end (gate+flag on): a real API deployment had the backend **enqueue `async_execute_bin` to PG** (`queue=celery_api_deployments`), the **orchestrator PG consumer ran it** (negative check: 0 `async_execute_bin` lines in the Celery general/api-deployment logs), then fan-out → `PgBarrier decrement → remaining=0` → callback → `COMPLETED`, with `pg_barrier_state` / `pg_batch_dedup` / queue all 0 after.

## Related Issues

- UN-3536 (PG Queue Phase 9 — transport engine), sub-task UN-3574. Builds on PR 3 (UN-3570, Flipt canary).
- Next: PR B — executor (tool run) on PG. Also tracked: UN-3566 (multi-queue consumer ergonomics).

## Checklist

- [x] Self-review performed

### DIFF
--- a/backend/pg_queue/producer.py
+++ b/backend/pg_queue/producer.py
@@ -1,0 +1,94 @@
+"""Backend-side producer for the PG queue (orchestrator dispatch — 9e PR A / 2d).
+
+The workers own the consumer and the worker-side ``dispatch()`` PG producer; the
+backend has the ``pg_queue`` tables but, until now, no way to *enqueue* to them —
+it dispatched the orchestrator (``async_execute_bin``) only via Celery. When an
+execution rides the ``pg_queue`` transport, the orchestrator itself must run on
+PG, so the backend enqueues it here.
+
+Same ``pg_queue_message`` row shape as the workers' ``PgQueueClient.send`` (a
+:class:`~unstract.core.data_models.TaskPayload` JSONB), written via the
+``PgQueueMessage`` ORM (whose Python-level field defaults supply ``now()`` / ``0``
+for the vt/counter columns). The ``TaskPayload`` / ``FairnessPayload`` wire
+contract is shared via ``unstract.core`` so producer and consumer agree on the
+keys without one codebase importing the other.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+from typing import Any
+
+from pg_queue.models import PgQueueMessage
+from unstract.core.data_models import FairnessPayload, TaskPayload
+
+logger = logging.getLogger(__name__)
+
+# Mirror the workers' fairness L3 bounds (``queue_backend.fairness`` — a separate
+# codebase, can't be imported). The DB ``CheckConstraint`` on
+# ``pg_queue_message.priority`` is the writer-proof backstop; this is the
+# friendly pre-check that names the offending value.
+_MIN_PRIORITY = 1
+_MAX_PRIORITY = 10
+DEFAULT_PRIORITY = 5
+
+# Default/general queue name — mirrors the workers' ``QueueName.GENERAL = "celery"``
+# (the Celery default queue), used when the caller passes no explicit queue.
+DEFAULT_GENERAL_QUEUE = "celery"
+
+
+def _json_safe(value: Any) -> Any:
+    """Round-trip through JSON with ``default=str`` so UUIDs / datetimes in the
+    task args/kwargs become strings.
+
+    ``PgQueueMessage.message`` is a plain ``JSONField`` (no Django encoder), and
+    the worker consumer already receives string ids on the existing PG dispatch
+    path, so coercing here keeps both transports consistent.
+    """
+    return json.loads(json.dumps(value, default=str))
+
+
+def enqueue_task(
+    *,
+    task_name: str,
+    queue: str | None,
+    args: list[Any] | None = None,
+    kwargs: dict[str, Any] | None = None,
+    org_id: str = "",
+    priority: int = DEFAULT_PRIORITY,
+    fairness: FairnessPayload | None = None,
+) -> int:
+    """Enqueue a task onto the PG queue; returns the new ``msg_id``.
+
+    Mirrors ``queue_backend.pg_queue.client.PgQueueClient.send`` so the worker
+    consumer can decode and run it. A PG enqueue failure propagates — the caller
+    decides; for the orchestrator there is no silent Celery fallback (that would
+    hide the failure or risk a double-dispatch).
+    """
+    if not _MIN_PRIORITY <= priority <= _MAX_PRIORITY:
+        raise ValueError(
+            f"priority out of range [{_MIN_PRIORITY}, {_MAX_PRIORITY}]: {priority!r}"
+        )
+    pg_queue = queue or DEFAULT_GENERAL_QUEUE
+    message: TaskPayload = {
+        "task_name": task_name,
+        "args": _json_safe(list(args) if args is not None else []),
+        "kwargs": _json_safe(dict(kwargs) if kwargs is not None else {}),
+        "queue": pg_queue,
+        "fairness": fairness,
+    }
+    row = PgQueueMessage.objects.create(
+        queue_name=pg_queue,
+        message=message,
+        org_id=org_id or "",
+        priority=priority,
+    )
+    logger.info(
+        "PG-queue: enqueued task=%r queue=%r msg_id=%s org=%r",
+        task_name,
+        pg_queue,
+        row.msg_id,
+        org_id,
+    )
+    return row.msg_id
--- a/backend/pg_queue/producer.py
+++ b/backend/pg_queue/producer.py
@@ -21,17 +21,21 @@ import logging
 from typing import Any
 
 from pg_queue.models import PgQueueMessage
-from unstract.core.data_models import FairnessPayload, TaskPayload
+from unstract.core.data_models import (
+    FAIRNESS_DEFAULT_PRIORITY,
+    FAIRNESS_MAX_PRIORITY,
+    FAIRNESS_MIN_PRIORITY,
+    FairnessPayload,
+    TaskPayload,
+)
 
 logger = logging.getLogger(__name__)
 
-# Mirror the workers' fairness L3 bounds (``queue_backend.fairness`` — a separate
-# codebase, can't be imported). The DB ``CheckConstraint`` on
-# ``pg_queue_message.priority`` is the writer-proof backstop; this is the
-# friendly pre-check that names the offending value.
-_MIN_PRIORITY = 1
-_MAX_PRIORITY = 10
-DEFAULT_PRIORITY = 5
+# Fairness L3 priority default, re-exported under the producer's name so callers
+# (workflow_helper) import one symbol. Bounds + default are the single source of
+# truth in unstract.core (shared with the workers' fairness/queue client); the DB
+# CheckConstraint on pg_queue_message.priority is the writer-proof backstop.
+DEFAULT_PRIORITY = FAIRNESS_DEFAULT_PRIORITY
 
 # Default/general queue name — mirrors the workers' ``QueueName.GENERAL = "celery"``
 # (the Celery default queue), used when the caller passes no explicit queue.
@@ -66,9 +70,10 @@ def enqueue_task(
     decides; for the orchestrator there is no silent Celery fallback (that would
     hide the failure or risk a double-dispatch).
     """
-    if not _MIN_PRIORITY <= priority <= _MAX_PRIORITY:
+    if not FAIRNESS_MIN_PRIORITY <= priority <= FAIRNESS_MAX_PRIORITY:
         raise ValueError(
-            f"priority out of range [{_MIN_PRIORITY}, {_MAX_PRIORITY}]: {priority!r}"
+            f"priority out of range "
+            f"[{FAIRNESS_MIN_PRIORITY}, {FAIRNESS_MAX_PRIORITY}]: {priority!r}"
         )
     pg_queue = queue or DEFAULT_GENERAL_QUEUE
     message: TaskPayload = {
@@ -78,12 +83,24 @@ def enqueue_task(
         "queue": pg_queue,
         "fairness": fairness,
     }
-    row = PgQueueMessage.objects.create(
-        queue_name=pg_queue,
-        message=message,
-        org_id=org_id or "",
-        priority=priority,
-    )
+    # Mirror the worker _enqueue_pg path: log the failure with breadcrumbs before
+    # it propagates, so a DB/constraint/serialization error isn't mislabeled by
+    # the caller's broad handler.
+    try:
+        row = PgQueueMessage.objects.create(
+            queue_name=pg_queue,
+            message=message,
+            org_id=org_id or "",
+            priority=priority,
+        )
+    except Exception:
+        logger.exception(
+            "PG-queue: failed to enqueue task=%r queue=%r org=%r",
+            task_name,
+            pg_queue,
+            org_id,
+        )
+        raise
     logger.info(
         "PG-queue: enqueued task=%r queue=%r msg_id=%s org=%r",
         task_name,
--- a/backend/workflow_manager/workflow_v2/workflow_helper.py
+++ b/backend/workflow_manager/workflow_v2/workflow_helper.py
@@ -662,20 +662,18 @@ class WorkflowHelper:
                 # the status now. (The old Celery-only TimeoutError handler is
                 # dropped: the wait loop is a manual poll, not AsyncResult.get, so
                 # it never raised; this generic handler covers any stray case.)
-                logger.error(
+                logger.exception(
                     f"[{org_schema}] Post-dispatch bookkeeping failed for execution "
                     f"'{execution_id}' (orchestrator already dispatched on "
-                    f"{transport}); not marking ERROR: {error}",
-                    exc_info=True,
+                    f"{transport}); not marking ERROR"
                 )
                 return ExecutionResponse(
                     workflow_id, execution_id, ExecutionStatus.EXECUTING.value
                 )
             WorkflowExecutionServiceHelper.update_execution_err(execution_id, str(error))
-            logger.error(
+            logger.exception(
                 f"Error while enqueuing async job for WF '{workflow_id}', "
-                f"execution '{execution_id}': {str(error)}",
-                exc_info=True,
+                f"execution '{execution_id}'",
                 stack_info=True,
             )
             return ExecutionResponse(
--- a/backend/workflow_manager/workflow_v2/workflow_helper.py
+++ b/backend/workflow_manager/workflow_v2/workflow_helper.py
@@ -13,6 +13,8 @@ from celery.result import AsyncResult
 from configuration.enums import ConfigKey
 from configuration.models import Configuration
 from django.db import IntegrityError
+from pg_queue.producer import DEFAULT_PRIORITY as PG_DEFAULT_PRIORITY
+from pg_queue.producer import enqueue_task as pg_enqueue_task
 from pipeline_v2.models import Pipeline
 from plugins import get_plugin
 from plugins.workflow_manager.workflow_v2.utils import WorkflowUtil
@@ -26,6 +28,7 @@ from utils.local_context import StateStore
 from utils.user_context import UserContext
 
 from backend.celery_service import app as celery_app
+from unstract.core.data_models import is_pg_transport
 from unstract.workflow_execution.enums import LogStage
 from workflow_manager.endpoint_v2.destination import DestinationConnector
 from workflow_manager.endpoint_v2.dto import FileHash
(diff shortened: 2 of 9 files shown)


backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -601,7 +628,11 @@ def execute_workflow_async(
...
                 execution_id,
-                async_execution.status,
+                # async_execution is None on the PG path; fall back to the row's
+                # current status (a Celery-only timeout can't strand the response).
+                async_execution.status
+                if async_execution is not None
+                else ExecutionStatus.EXECUTING.value,
                 message=WorkflowMessages.CELERY_TIMEOUT_MESSAGE,
             )
         except Exception as error:
```
muhammad-ali-e, line 657: **HIGH (silent failure / double-state).** This broad `except Exception` wraps everything *after* the enqueue: `WorkflowExecution.objects.get`, `update_execution_task`, the entire poll loop, `ResultCacheUtils.get_api_results`, `_set_result_acknowledge`. On the PG path the row at line 556 may already be enqueued (and a PG consumer may already be running `async_execute_bin`) when a *post-enqueue* line throws here — the handler then calls `update_execution_err(...)` and flips the row to `ERROR` while the orchestrator is genuinely running. That reintroduces exactly the hidden-failure / inconsistent-state risk the producer docstring set out to avoid ('no silent Celery fallback ... risk a double-dispatch').

It also mislabels *every* failure (transport resolution, `DoesNotExist`, cache/Redis, serialization) as a generic enqueuing error.

**Suggested fix:** track an `enqueued`/`dispatched` flag and only `update_execution_err` for *pre-dispatch* failures; for post-dispatch bookkeeping failures, log with an error id and return a response reflecting the real (running) state rather than forcing `ERROR`.

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -601,7 +628,11 @@ def execute_workflow_async(
             return ExecutionResponse(
                 workflow_id,
                 execution_id,
-                async_execution.status,
+                # async_execution is None on the PG path; fall back to the row's
+                # current status (a Celery-only timeout can't strand the response).
+                async_execution.status
```
muhammad-ali-e: **MEDIUM (status accuracy + comment + readability).** Three overlapping notes on this `except celery_exceptions.TimeoutError` branch:

1. On the PG path `async_execution` is always `None`, so this returns a hardcoded `EXECUTING.value` **plus** `CELERY_TIMEOUT_MESSAGE` — a Celery-specific message on a PG execution, and a status that may contradict the persisted row (it could already be `ERROR`/`COMPLETED`).
2. The comment 'a Celery-only timeout can’t strand the response' is the most cryptic phrasing in the PR. In fact this handler is effectively **unreachable on the PG path** — the wait loop (line 602) uses `time.sleep`, not a Celery result-wait, so nothing in the `try` raises `TimeoutError` for PG. The `is not None` guard is purely defensive.
3. The multi-line ternary as a positional arg is hard to read.

**Suggested fix:** lift the value to a named local (`timeout_status = ...`) above the `return`, reword the comment to state the mechanical fact ('only Celery raises this; the guard is defensive'), and use a transport-neutral message instead of `CELERY_TIMEOUT_MESSAGE`.

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -526,48 +529,72 @@ def execute_workflow_async(
...
+                    org_id=org_schema or "",
+                    priority=PG_DEFAULT_PRIORITY,
+                    fairness={
+                        "org_id": org_schema or None,
+                        # Mirrors the workers' WorkloadType enum values.
+                        "workload_type": "api" if is_api_execution else "non_api",
+                        "pipeline_priority": PG_DEFAULT_PRIORITY,
+                    },
+                )
+                task_id: str | None = f"pg:{msg_id}"
```
muhammad-ali-e: **HIGH (contract divergence).** This stores `task_id = f"pg:{msg_id}"` (e.g. `pg:4242`), but the *worker* PG dispatch path (`workers/queue_backend/dispatch.py`, `PgDispatchHandle.id`) uses `str(msg_id)` with **no** prefix — and the scheduler path stores that bare id. So the same logical thing (a PG-enqueued `async_execute_bin`) lands in `WorkflowExecution.task_id` in two different formats depending on entry path. This directly undercuts the PR's goal of one shared PG dispatch contract. No consumer parses `task_id` today, so impact is latent, but any future PG-vs-Celery detection or msg_id extraction will trip on it.

**Suggested fix:** pick one convention — simplest is `task_id = str(msg_id)` here to match `PgDispatchHandle.id`.

**Minor (style):** declare `task_id: str | None` once before the `if` rather than annotating inside the first branch (line 579 then re-assigns unannotated).

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -526,48 +529,72 @@ def execute_workflow_async(
...
+                    # None → "celery" (general); else celery_api_deployments.
+                    queue=queue,
+                    args=dispatch_args,
+                    kwargs=dispatch_kwargs,
+                    org_id=org_schema or "",
+                    priority=PG_DEFAULT_PRIORITY,
+                    fairness={
+                        "org_id": org_schema or None,
+                        # Mirrors the workers' WorkloadType enum values.
+                        "workload_type": "api" if is_api_execution else "non_api",
```
muhammad-ali-e: **MEDIUM (type safety).** `workload_type` is hand-built from bare string literals `"api"`/`"non_api"`, guarded only by a comment that they 'mirror the workers’ WorkloadType enum'. A typo here is invisible until the worker calls `WorkloadType(payload["workload_type"])` (`pg_barrier.py:472`) and raises far from the cause; the consumer's fairness-header path doesn't re-validate it at all.

**Suggested fix:** the PR already moved `FairnessPayload` into `unstract.core` to be the single source of truth — move the `WorkloadType` values there too (or expose its `.value`s) and reference them here, and type the field as `Literal["api","non_api"]` (see the `data_models.py:276` comment).

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -526,48 +529,72 @@ def execute_workflow_async(
...
+                msg_id = pg_enqueue_task(
+                    task_name="async_execute_bin",
+                    # None → "celery" (general); else celery_api_deployments.
+                    queue=queue,
+                    args=dispatch_args,
+                    kwargs=dispatch_kwargs,
+                    org_id=org_schema or "",
+                    priority=PG_DEFAULT_PRIORITY,
+                    fairness={
+                        "org_id": org_schema or None,
```
muhammad-ali-e: **LOW (intentional, but document + test).** Same `org_schema` is coerced two ways one line apart: `org_id=org_schema or ""` for the row column (NOT NULL → wants `""`) vs `"org_id": org_schema or None` for the fairness payload (`str | None`). This is correct, but reads like a bug and invites a future 'simplification' that would break one side. Add a one-line comment explaining the two sentinels, and add a test with `org_schema=""` asserting the row gets `""` while `fairness["org_id"] is None`.

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -526,48 +529,72 @@ def execute_workflow_async(
...
+                "hitl_queue_name": hitl_queue_name,
+                "hitl_packet_id": hitl_packet_id,
+                "custom_data": custom_data,
+                "transport": transport,
+            }
+            # Orchestrator transport (9e PR A / 2d): a pg_queue execution runs
+            # async_execute_bin on PG too — enqueue it to pg_queue_message (a PG
+            # consumer on this queue runs it). Celery executions are unchanged.
+            async_execution: AsyncResult | None = None
+            if is_pg_transport(transport):
```
muhammad-ali-e: **TEST GAP (highest leverage).** The entire PG-vs-Celery fork introduced here is untested. There is no test asserting that `is_pg_transport(transport)` true → `pg_enqueue_task` called once and `celery_app.send_task` NOT called (and `task_id == f"pg:{msg_id}"`), nor the converse for Celery. A refactor could silently route PG executions back to Celery, or double-dispatch, with no failing test. `test_transport.py` only covers `resolve_transport`, not how its result is consumed here.

**Suggested:** add a suite mocking `resolve_transport`, `pg_enqueue_task`, `celery_app.send_task`, `WorkflowExecution.objects.get`; parametrize on transport and also assert the `queue → workload_type` mapping (`CELERY_API_DEPLOYMENTS → "api"`, else `"non_api"`).

muhammad-ali-e (commented):
Automated PR Review Toolkit pass (Code Reviewer, Silent Failure Hunter, Type Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier).

Scope reviewed: the 8 changed files vs the PR base `feat/UN-3445-pg-queue-integration`. The producer + contract relocation are clean and well-documented; the substantive concerns are concentrated in `execute_workflow_async`'s error/return paths and in test coverage of the new transport fork. Inline comments below, grouped by line. None are blockers individually, but #1 (broad except can mark a live PG job ERROR) and #2 (task_id format divergence) are the two I'd resolve before merge.

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -526,48 +529,72 @@ def execute_workflow_async(
...
+                    kwargs=dispatch_kwargs,
+                    queue=queue,
+                )
+                task_id = async_execution.id
+
             workflow_execution: WorkflowExecution = WorkflowExecution.objects.get(
                 id=execution_id
             )
-            if not async_execution.id:
+            if not task_id:
```
muhammad-ali-e, line 614: **LOW (dead-on-PG branch + stale comment).** On the PG path `task_id` is `f"pg:{msg_id}"`, always truthy, so `if not task_id` can now only fire on the Celery path. The reworded warning ('Empty task_id ...') and the 'Continue without setting task_id' comment now read as transport-agnostic but are Celery-only. Keep the transport in the message so operators can tell which path produced an empty id, and consider a test confirming the PG path always reaches `update_execution_task`.

backend/pg_queue/producer.py
```diff
@@ -0,0 +1,94 @@
...
+from unstract.core.data_models import FairnessPayload, TaskPayload
+
+logger = logging.getLogger(__name__)
+
+# Mirror the workers' fairness L3 bounds (``queue_backend.fairness`` — a separate
+# codebase, can't be imported). The DB ``CheckConstraint`` on
+# ``pg_queue_message.priority`` is the writer-proof backstop; this is the
+# friendly pre-check that names the offending value.
+_MIN_PRIORITY = 1
+_MAX_PRIORITY = 10
```
muhammad-ali-e: **MEDIUM (duplication / drift).** `_MIN_PRIORITY` / `_MAX_PRIORITY` / `DEFAULT_PRIORITY` duplicate `workers/queue_backend/fairness.py` (`MIN_PRIORITY`/`MAX_PRIORITY`/`DEFAULT_PRIORITY`) and `client.py` re-checks the same `[1,10]` range — three copies of the same bound, each with a comment admitting it 'can’t be imported'. That comment is the signal the bound belongs in `unstract.core.data_models` next to `FairnessPayload` (same justification already used to move the types there).

**Suggested fix:** define `FAIRNESS_MIN/MAX/DEFAULT_PRIORITY` in `unstract.core.data_models`; import in both `producer.py` and `fairness.py`. The DB `CheckConstraint` stays the backstop.

backend/pg_queue/producer.py
```diff
@@ -0,0 +1,94 @@
...
+        "queue": pg_queue,
+        "fairness": fairness,
+    }
+    row = PgQueueMessage.objects.create(
+        queue_name=pg_queue,
+        message=message,
+        org_id=org_id or "",
+        priority=priority,
+    )
+    logger.info(
```
muhammad-ali-e, line 104: **LOW (log parity).** The worker's `_enqueue_pg` (`dispatch.py`) wraps the send in a `try/except` that emits `logger.exception("PG-queue: failed to enqueue ...")` with breadcrumbs before re-raising. This producer logs only on success; a `PgQueueMessage.objects.create` failure (DB down, unmigrated table, CheckConstraint violation, JSONField serialization error) propagates with no producer-side context and is then mislabeled by the broad caller handler (see `workflow_helper.py:638`). For parity with the path this PR mirrors, add a `try/except` that `logger.exception(...)`s task/queue/org and re-raises.

unstract/core/src/unstract/core/data_models.py
```diff
@@ -260,6 +260,44 @@ def is_pg_transport(transport: str | None) -> bool:
...
+    workers' ``queue_backend.fairness`` (consumer) both reference this exact
+    sub-shape so they agree on the keys instead of a loose ``dict[str, Any]``.
+    Lives in ``unstract.core`` because backend and workers are separate
+    codebases that can't import each other — this is their single source of
+    truth. ``workload_type`` is the ``WorkloadType`` enum's value;
+    ``pipeline_priority`` is the 1..10 fairness L3 priority.
+    """
+
+    org_id: str | None
+    workload_type: str
```
muhammad-ali-e: **MEDIUM (invariant expression).** The docstring promises 'the `WorkloadType` enum’s value' but the type is bare `str`, so any string passes the type checker. Since this is the shared single-source-of-truth contract, tighten it to `workload_type: Literal["api", "non_api"]` — zero runtime cost, and it makes the backend producer's hand-built dict (`workflow_helper.py:567`) statically checkable. Even better: move `WorkloadType` (a `StrEnum`, serialises to its value) into `unstract.core` and annotate `workload_type: WorkloadType`; the wire shape is unchanged and the `[1,10]` priority bound could move alongside it.

unstract/core/src/unstract/core/data_models.py
```diff
@@ -260,6 +260,44 @@ def is_pg_transport(transport: str | None) -> bool:
...
+
+    ``fairness`` is a :class:`FairnessPayload` (serialised fairness key) or
+    ``None`` — the consumer rebuilds the ``x-fairness-key`` header from it,
+    mirroring the Celery dispatch path.
+    """
+
+    task_name: str
+    args: list[Any]
+    kwargs: dict[str, Any]
+    queue: str | None
```
muhammad-ali-e, line 323: **LOW–MEDIUM (inconsistent contract).** `queue: str | None` tells three different stories: the backend producer always coerces `queue or DEFAULT_GENERAL_QUEUE` before building the payload (never `None`), the worker `to_payload` passes `None` through, and the consumer never reads `message["queue"]` at all (routing comes from the `queue_name` column). So the `None` arm is effectively dead on the backend side and the field is vestigial/diagnostic.

**Suggested fix:** decide the contract — either make it `queue: str` and have `to_payload` default to the general queue (so both producers agree), or document it as diagnostic-only and note the column is authoritative.

workers/scheduler/tasks.py
```diff
@@ -182,6 +183,10 @@ def _execute_scheduled_workflow(
                     "transport": transport,
                 },
                 queue=QueueName.GENERAL,
+                # Orchestrator transport (9e PR A / 2d): route async_execute_bin
+                # onto PG for a pg_queue execution (carried-marker wins over the
+                # allow-list); None keeps the legacy Celery dispatch.
+                backend=QueueBackend.PG if is_pg_transport(transport) else None,
```
muhammad-ali-e, line 189: **TEST GAP.** `test_dispatch_sites_characterisation.py::TestSchedulerDispatchSite` pins task name/args/kwargs/queue but does NOT assert this new `backend=` argument — the one line that actually routes the scheduled orchestrator onto PG. With `dispatch` mocked there, the routing is unobserved, and `routing.py` compares with `backend is QueueBackend.PG` (identity, never `== "pg"`), making a regression easy to introduce silently.

**Suggested:** extend that class — `transport == "pg_queue"` → `dispatch.call_args.kwargs["backend"] is QueueBackend.PG`; Celery/omitted → `backend is None`.

backend/pg_queue/producer.py
```diff
@@ -0,0 +1,94 @@
...
+    fairness: FairnessPayload | None = None,
+) -> int:
+    """Enqueue a task onto the PG queue; returns the new ``msg_id``.
+
+    Mirrors ``queue_backend.pg_queue.client.PgQueueClient.send`` so the worker
+    consumer can decode and run it. A PG enqueue failure propagates — the caller
+    decides; for the orchestrator there is no silent Celery fallback (that would
+    hide the failure or risk a double-dispatch).
+    """
+    if not _MIN_PRIORITY <= priority <= _MAX_PRIORITY:
```
muhammad-ali-e: **TEST GAP (boundary values).** This `[1,10]` check is only exercised by `test_priority_out_of_range_raises` with `priority=99` — far outside the range, so an off-by-one (`<` vs `<=`) at the boundary would not be caught. Add parametrized tests: `priority in {1, 10}` succeeds, `priority in {0, 11, -1}` raises `ValueError`, and the default (`5`) is what the row gets when omitted. Also worth one `datetime` case for `_json_safe` (the docstring promises `default=str` covers datetimes, but only UUID is tested).

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: Fixed in 2a224e275. Added a `dispatched` flag: a failure in the post-dispatch bookkeeping (status read / poll / result fetch) no longer calls `update_execution_err` — the orchestrator is already enqueued/running on its transport and owns the status. Only a *pre-dispatch* failure marks ERROR. Applies to both transports.

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: Fixed in 2a224e275 — dropped the dead Celery-only `TimeoutError` handler entirely. The wait loop is a manual poll (not `AsyncResult.get`), so it never raised; the generic handler (now `dispatched`-aware) covers any stray case. No more Celery-specific message / hardcoded status on the PG path.

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: Fixed in 2a224e275 — `task_id` is now bare `str(msg_id)` (no `pg:` prefix), matching the worker `PgDispatchHandle.id` and the scheduler path. One format across entry paths.

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: Fixed in 2a224e275 — moved `WorkloadType` (StrEnum) into `unstract.core` and the producer now uses `WorkloadType.API.value` / `.NON_API.value` (no hand-built literals). Workers re-export it.

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: Fixed in 2a224e275 — the two sentinels now have an explicit comment in `_dispatch_orchestrator_task` (row `org_id` column is NOT NULL → `""`; fairness `org_id` is `str | None` → `None` = no segment), plus a regression test `test_empty_org_uses_two_sentinels` asserting both.

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: Fixed in 2a224e275 — the warning now carries `transport`, and since the PG path always yields a truthy `str(msg_id)`, the `if not task_id` branch is documented as Celery-only. The new fork test asserts the PG path returns a truthy id → always reaches `update_execution_task`.

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: Fixed in 2a224e275 — extracted the fork to `_dispatch_orchestrator_task` and added `test_dispatch_orchestrator.py`: PG → `pg_enqueue_task` called once + `celery_app.send_task` NOT called + `task_id == str(msg_id)`; Celery → the converse. Plus workload/org-sentinel assertions.

backend/pg_queue/producer.py
muhammad-ali-e: Fixed in 2a224e275 — `FAIRNESS_MIN/MAX/DEFAULT_PRIORITY` now live in `unstract.core.data_models`; the producer imports them and the workers re-export (`MIN/MAX/DEFAULT_PRIORITY`). Single source of truth; the DB CheckConstraint stays the writer-proof backstop.

backend/pg_queue/producer.py
muhammad-ali-e: Fixed in 2a224e275 — the producer now wraps the create in `try/except` → `logger.exception("PG-queue: failed to enqueue ...")` with task/queue/org breadcrumbs, then re-raises. Parity with the worker `_enqueue_pg`, and no longer mislabeled by the caller (which now also won’t mark ERROR post-dispatch).

unstract/core/src/unstract/core/data_models.py
muhammad-ali-e: Fixed in 2a224e275 — `workload_type` is now `Literal["api", "non_api"]`, and `WorkloadType` (StrEnum) lives in `unstract.core` so the producer references the enum values. The `[1,10]` bound also moved to core (see the priority-duplication thread).

unstract/core/src/unstract/core/data_models.py
muhammad-ali-e: Addressed in 2a224e275 — documented the contract: `queue` is **diagnostic only** (the consumer routes by the `queue_name` *column*, not this field). Docstring now states the backend producer always sets it while the workers’ `to_payload` may leave it `None`. Kept the field for debugging rather than dropping it.

workers/scheduler/tasks.py
muhammad-ali-e: Fixed in 2a224e275 — extended `TestSchedulerDispatchSite`: `transport=="pg_queue"` → `dispatch.call_args.kwargs["backend"] is QueueBackend.PG` (identity), and `celery` → `backend is None`.

backend/pg_queue/producer.py
muhammad-ali-e: Fixed in 2a224e275 — parametrized boundary tests: `priority in {1,5,10}` accepted (asserts the row value), `{0,11,-1}` raise; default-when-omitted asserts `5`; added a `datetime` case for `_json_safe` and an enqueue-failure (logs + re-raises) test.

muhammad-ali-e:
## Review round addressed — `2a224e275`

Thanks for the deep pass — all 13 threads + the SonarCloud complexity issue are handled.

**Structural (the HIGHs + SonarCloud):**
- Extracted `_dispatch_orchestrator_task` → `execute_workflow_async` cognitive complexity back under 15, and the PG-vs-Celery fork is now unit-tested.
- `dispatched` flag → a post-dispatch bookkeeping failure no longer flips a running execution to ERROR (only pre-dispatch does).
- `task_id` = bare `str(msg_id)` → one format across entry paths.
- Dropped the dead Celery-only `TimeoutError` handler.

**Contract consolidation (into `unstract.core`):** `WorkloadType` (StrEnum), `FAIRNESS_MIN/MAX/DEFAULT_PRIORITY`, and `FairnessPayload.workload_type: Literal[...]` — no more hand-built `"api"`/`"non_api"` literals or triplicated `[1,10]` bounds; workers re-export, backend references.

**Producer:** log+re-raise on enqueue failure (parity with the worker path); `TaskPayload.queue` documented as diagnostic-only.

**Tests:** +dispatch-fork suite, +producer boundary/datetime/failure, +scheduler `backend=PG` override assertions. 34 backend + 99 workers green; pre-commit clean.

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -581,35 +641,39 @@ def execute_workflow_async(
...
+                )
             WorkflowExecutionServiceHelper.update_execution_err(execution_id, str(error))
-            logger.error(
+            logger.exception(
                 f"Error while enqueuing async job for WF '{workflow_id}', "
-                f"execution '{execution_id}': {str(error)}",
-                exc_info=True,
+                f"execution '{execution_id}'",
                 stack_info=True,
             )
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in d48d4986f — dropped `stack_info=True`. `logger.exception()` keeps the full exception traceback; no more redundant second (call-stack) dump per error.

muhammad-ali-e:
## greptile round addressed — `d48d4986f`

Single P2 fixed: dropped the redundant `stack_info=True` on the pre-dispatch `logger.exception()` (the exception traceback is already attached; `stack_info` added a second overlapping stack dump). Thanks for the 4/5 / safe-to-merge.

PR #2073, opened 2026-06-18
UN-3566 [FEAT] PG Queue 9f — multi-queue consumer + named run-worker.sh PG roles

muhammad-ali-e:
## What

- **Multi-queue PG consumer:** `PgQueueConsumer` now takes `queue_names: list[str]` and polls them **round-robin** — one process can drain several queues (e.g. ETL + API fan-out) instead of one-process-per-queue.
- **Named PG roles in `run-worker.sh`:** the coupled pipeline's PG consumers are first-class, runnable **individually or as a set**, like the Celery workers — no per-process env.

## Why

- Running the PG pipeline meant launching 3–4 consumers each with hand-set `WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE` + `-q` + distinct `-p` (a repeated footgun). This makes `./run-worker.sh -d pg` (all) and `./run-worker.sh -d pg-fileproc` (individual) work just like `-d all` / `-d file`. Part of PG Queue Phase 9 (UN-3536), sub-task UN-3566.

## How

- **Consumer:** `queue_names` round-robin (`read()` per queue preserves the `(queue_name, priority DESC, msg_id)` index's top-N; no queue starves another). `WORKER_PG_QUEUE_CONSUMER_QUEUE` is comma-parsed; a single value stays a one-element list → **back-compat** with the existing leaf consumer.
- **Roles** (`PG_CONSUMER_ROLES`, `"<source_worker_type>;<queues>"`):
  - `pg-orchestrator-api` (api_deployment → `celery_api_deployments`) and `pg-orchestrator-general` (general → `celery`) — split because `async_execute_bin` has **distinct impls per registry** (api-deployment handles it as an API deployment; general routes by workflow type), so one consumer can't serve both.
  - `pg-fileproc` (file_processing → `file_processing,api_file_processing`) and `pg-callback` (callback → both callback queues) — **multi-queue**, one process each covering ETL + API.
  - The role name rides in argv (`python -m pg_queue_consumer <role>`) so `pgrep` (`-s`/`-k`/`-r`) distinguishes the otherwise-identical processes. Roles share the `pg_queue_consumer/` launcher dir; `resolve_log_file` + `-L pg` know the role log path.
- **`./run-worker.sh -d pg`** launches the 4 pipeline roles + reaper (mirrors `all`); `-s`/`-k`/`-r`/`-L` operate per role.

- **No.** All changes are consumer-side ergonomics, gated off by the existing transport flag — no runtime change until PG consumers are run. The single-queue config still works (one-element list), so the existing leaf-webhook consumer and the `pg-queue-consumer` alias are unaffected. No schema/API change.

## Database Migrations

- None.

## Env Config

- None new. `WORKER_PG_QUEUE_CONSUMER_QUEUE` now accepts a comma-separated list (single value unchanged).

## Notes on Testing

- Unit: round-robin aggregation across queues; empty queue doesn't starve others; empty-list rejected; comma-parse incl. single-value back-compat. Existing consumer tests updated to the list arg.
- Dev-tested end-to-end: `./run-worker.sh -d pg` → `-s` shows the 4 roles + reaper RUNNING; an API execution drained **orchestrator-api → pg-fileproc (one process draining both `file_processing` and `api_file_processing`) → pg-callback → COMPLETED**; `-r pg-fileproc` restarted only that role (pg-callback PID unchanged).

## Related Issues

- UN-3536 (PG Queue Phase 9), sub-task UN-3566. Independent of the orchestrator-on-PG PR (#2072); the orchestrator roles here consume what that PR produces.

## Checklist

- [x] Self-review performed

### DIFF
--- a/workers/queue_backend/pg_queue/consumer.py
+++ b/workers/queue_backend/pg_queue/consumer.py
@@ -70,7 +70,7 @@ class PgQueueConsumer:
 
     def __init__(
         self,
-        queue_name: str,
+        queue_names: list[str],
         *,
         client: PgQueueClient | None = None,
         app: Celery | None = None,
@@ -97,7 +97,13 @@ class PgQueueConsumer:
             raise ValueError(
                 f"backoff_max ({backoff_max}) must be >= poll_interval ({poll_interval})"
             )
-        self.queue_name = queue_name
+        if not queue_names:
+            raise ValueError("queue_names must be a non-empty list")
+        # One process can drain several queues (9f) — e.g. a file_processing
+        # consumer drains both file_processing and api_file_processing. Polled
+        # round-robin so the per-queue (queue_name, priority DESC, msg_id) dequeue
+        # index keeps its top-N efficiency and no queue starves another.
+        self.queue_names = queue_names
         self._client = client if client is not None else PgQueueClient()
         self._app = app if app is not None else current_app
         self.batch_size = batch_size
@@ -114,14 +120,19 @@ class PgQueueConsumer:
         self._last_poll_monotonic = time.monotonic()
 
     def poll_once(self) -> int:
-        """Claim + process one batch; returns the number of messages claimed."""
+        """Claim + process one batch per queue (round-robin); returns the total
+        number of messages claimed across all queues this cycle.
+        """
         self._last_poll_monotonic = time.monotonic()
-        messages = self._client.read(
-            self.queue_name, vt_seconds=self.vt_seconds, qty=self.batch_size
-        )
-        for message in messages:
-            self._handle(message)
-        return len(messages)
+        total = 0
+        for queue_name in self.queue_names:
+            messages = self._client.read(
+                queue_name, vt_seconds=self.vt_seconds, qty=self.batch_size
+            )
+            for message in messages:
+                self._handle(message)
+            total += len(messages)
+        return total
 
     def _handle(self, message: QueueMessage) -> None:
         payload = message.message
@@ -254,9 +265,9 @@ class PgQueueConsumer:
             name for name in self._app.tasks if not name.startswith("celery.")
         )
         logger.info(
-            "PG-queue consumer started (queue=%r, batch=%s, vt=%ss) — "
+            "PG-queue consumer started (queues=%r, batch=%s, vt=%ss) — "
             "%d application task(s) registered: %s",
-            self.queue_name,
+            self.queue_names,
             self.batch_size,
             self.vt_seconds,
             len(app_tasks),
@@ -278,7 +289,7 @@ class PgQueueConsumer:
             else:
                 time.sleep(backoff)
                 backoff = min(backoff * 2, self.backoff_max)
-        logger.info("PG-queue consumer stopped (queue=%r)", self.queue_name)
+        logger.info("PG-queue consumer stopped (queues=%r)", self.queue_names)
 
     def stop(self, *_: object) -> None:
         """Request a graceful stop after the current batch."""
@@ -294,6 +305,13 @@ class PgQueueConsumer:
                 "PG-queue consumer: signal handlers not installed (non-main "
                 "thread) — SIGTERM/SIGINT will not trigger graceful shutdown"
             )
+
+
+def _parse_queue_list(raw: str) -> list[str]:
+    """Comma-separated queue list (9f). A single value stays a one-element list,
+    so the pre-9f single-queue config remains valid.
+    """
+    return [q.strip() for q in raw.split(",") if q.strip()]
 
 
 def main() -> None:
@@ -315,7 +333,7 @@ def main() -> None:
             raise ValueError(f"Invalid {var}={raw!r}: {exc}") from exc
 
     consumer = PgQueueConsumer(
-        queue_name=_env("QUEUE", _DEFAULT_QUEUE, str),
+        queue_names=_env("QUEUE", [_DEFAULT_QUEUE], _parse_queue_list),
         batch_size=_env("BATCH", _DEFAULT_BATCH, int),
         vt_seconds=_env("VT_SECONDS", _DEFAULT_VT_SECONDS, int),
         poll_interval=_env("POLL_INTERVAL", _DEFAULT_POLL_INTERVAL, float),
--- a/workers/queue_backend/pg_queue/consumer.py
+++ b/workers/queue_backend/pg_queue/consumer.py
@@ -100,10 +100,14 @@ class PgQueueConsumer:
         if not queue_names:
             raise ValueError("queue_names must be a non-empty list")
         # One process can drain several queues (9f) — e.g. a file_processing
-        # consumer drains both file_processing and api_file_processing. Polled
-        # round-robin so the per-queue (queue_name, priority DESC, msg_id) dequeue
-        # index keeps its top-N efficiency and no queue starves another.
-        self.queue_names = queue_names
+        # consumer drains both file_processing and api_file_processing. Each is
+        # read once per cycle in list order (poll_once): this prevents starvation
+        # (every queue gets a read each cycle) but does NOT equalize throughput —
+        # an always-full queue_names[0] claims its full batch and runs it before
+        # later queues. Copy + de-dup (order-preserving): a duplicate would
+        # double-read a queue per cycle, and storing the caller's list by
+        # reference would let a later mutation bypass the non-empty validation.
+        self.queue_names = list(dict.fromkeys(queue_names))
         self._client = client if client is not None else PgQueueClient()
         self._app = app if app is not None else current_app
         self.batch_size = batch_size
@@ -120,18 +124,30 @@ class PgQueueConsumer:
         self._last_poll_monotonic = time.monotonic()
 
     def poll_once(self) -> int:
-        """Claim + process one batch per queue (round-robin); returns the total
-        number of messages claimed across all queues this cycle.
+        """Claim + process one batch per queue (read once each, in list order);
+        returns the total number of messages claimed across all queues this cycle.
+
+        Each queue is isolated: a read/handle failure on one queue is logged and
+        skipped so the others still get their turn, and the work already done this
+        cycle still counts (so run() doesn't take the empty-queue backoff path
+        after a partial failure).
         """
         self._last_poll_monotonic = time.monotonic()
         total = 0
         for queue_name in self.queue_names:
-            messages = self._client.read(
-                queue_name, vt_seconds=self.vt_seconds, qty=self.batch_size
-            )
-            for message in messages:
-                self._handle(message)
-            total += len(messages)
+            try:
+                messages = self._client.read(
+                    queue_name, vt_seconds=self.vt_seconds, qty=self.batch_size
+                )
+                for message in messages:
+                    self._handle(message)
+                total += len(messages)
+            except Exception:
+                logger.exception(
+                    "PG-queue consumer: poll failed for queue %r; "
+                    "continuing with the other queues",
+                    queue_name,
+                )
         return total
 
     def _handle(self, message: QueueMessage) -> None:
@@ -309,9 +325,22 @@ class PgQueueConsumer:
 
 def _parse_queue_list(raw: str) -> list[str]:
     """Comma-separated queue list (9f). A single value stays a one-element list,
-    so the pre-9f single-queue config remains valid.
+    so the pre-9f single-queue config remains valid. Empty entries (a doubled or
+    trailing comma — almost always a config typo) are dropped with a warning so a
+    malformed list is diagnosable from the logs, not just by eyeballing.
     """
-    return [q.strip() for q in raw.split(",") if q.strip()]
+    parts = [q.strip() for q in raw.split(",")]
+    queues = [q for q in parts if q]
+    dropped = len(parts) - len(queues)
+    if dropped:
+        logger.warning(
+            "PG-queue consumer: dropped %d empty queue name(s) from "
+            "WORKER_PG_QUEUE_CONSUMER_QUEUE=%r → %r",
+            dropped,
+            raw,
+            queues,
+        )
+    return queues
 
 
 def main() -> None:
--- a/workers/run-worker.sh
+++ b/workers/run-worker.sh
@@ -49,19 +49,25 @@ readonly PG_QUEUE_SET="pg-queue"
 # registry (api-deployment handles it as an API deployment; general routes by
 # workflow type), so one consumer can't serve both — they bind different queues.
 # fileproc/callback are multi-queue (one process drains both ETL + API queues).
+# Role-name constants — the literal lives in one place and is referenced across
+# PG_CONSUMER_ROLES / PG_QUEUE_MEMBERS / WORKERS.
+readonly PG_ROLE_ORCH_API="pg-orchestrator-api"
+readonly PG_ROLE_ORCH_GENERAL="pg-orchestrator-general"
+readonly PG_ROLE_FILEPROC="pg-fileproc"
+readonly PG_ROLE_CALLBACK="pg-callback"
 declare -rA PG_CONSUMER_ROLES=(
-    ["pg-orchestrator-api"]="api_deployment;celery_api_deployments"
-    ["pg-orchestrator-general"]="general;celery"
-    ["pg-fileproc"]="file_processing;file_processing,api_file_processing"
-    ["pg-callback"]="callback;file_processing_callback,api_file_processing_callback"
+    ["$PG_ROLE_ORCH_API"]="api_deployment;celery_api_deployments"
+    ["$PG_ROLE_ORCH_GENERAL"]="general;celery"
+    ["$PG_ROLE_FILEPROC"]="file_processing;file_processing,api_file_processing"
+    ["$PG_ROLE_CALLBACK"]="callback;file_processing_callback,api_file_processing_callback"
 )
 declare -rA PG_QUEUE_MEMBERS=(
     ["$PG_QUEUE_CONSUMER_TYPE"]=1
     ["$PG_QUEUE_REAPER_TYPE"]=1
-    ["pg-orchestrator-api"]=1
-    ["pg-orchestrator-general"]=1
-    ["pg-fileproc"]=1
-    ["pg-callback"]=1
+    ["$PG_ROLE_ORCH_API"]=1
+    ["$PG_ROLE_ORCH_GENERAL"]=1
+    ["$PG_ROLE_FILEPROC"]=1
+    ["$PG_ROLE_CALLBACK"]=1
 )
 # Log-tail alias for the Celery transport: every worker EXCEPT the PG-queue
 # members — the *complement* of the 'pg-queue' tail alias, so the two
@@ -95,10 +101,10 @@ declare -A WORKERS=(
     ["pg-consumer"]="$PG_QUEUE_CONSUMER_TYPE"
     # PG Queue named roles (9f) — coupled-pipeline consumers, queues + source
     # worker-type baked in. Run individually like any worker, or via the 'pg' set.
-    ["pg-orchestrator-api"]="pg-orchestrator-api"
-    ["pg-orchestrator-general"]="pg-orchestrator-general"
-    ["pg-fileproc"]="pg-fileproc"
-    ["pg-callback"]="pg-callback"
+    ["$PG_ROLE_ORCH_API"]="$PG_ROLE_ORCH_API"
+    ["$PG_ROLE_ORCH_GENERAL"]="$PG_ROLE_ORCH_GENERAL"
+    ["$PG_ROLE_FILEPROC"]="$PG_ROLE_FILEPROC"
+    ["$PG_ROLE_CALLBACK"]="$PG_ROLE_CALLBACK"
     # PG Queue reaper — leader-elected recovery loop (barrier-orphan sweep)
     ["reaper"]="$PG_QUEUE_REAPER_TYPE"
     ["pg-queue-reaper"]="$PG_QUEUE_REAPER_TYPE"
--- a/workers/run-worker.sh
+++ b/workers/run-worker.sh
@@ -69,11 +69,11 @@ declare -rA PG_QUEUE_MEMBERS=(
     ["$PG_ROLE_FILEPROC"]=1
     ["$PG_ROLE_CALLBACK"]=1
 )
-# Log-tail alias for the Celery transport: every worker EXCEPT the PG-queue
-# members — the *complement* of the 'pg-queue' tail alias, so the two
-# transports' logs can be tailed separately (-L celery vs -L pg-queue).
-# Tail-only — there is no 'celery' run alias; this set is started via 'all',
-# which by design omits the opt-in PG-queue workers.
+# The Celery transport set: every worker EXCEPT the PG-queue members — the
+# *complement* of the 'pg-queue' set, so the two transports' logs can be tailed
+# separately (-L celery vs -L pg-queue). Since 9f, 'celery' is BOTH a -L tail
+# alias AND a run alias (WORKERS maps it to "all" → run_all_workers), symmetric
+# with 'pg'; 'all' remains its synonym. Either way it omits the opt-in PG workers.
 readonly CELERY_SET="celery"
 
 # Available workers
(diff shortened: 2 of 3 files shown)


workers/run-worker.sh
```diff
@@ -1071,7 +1154,8 @@ if [[ "$RESTART_MODE" == "true" ]]; then
             print_status $RED "Cannot restart PG-queue set: a member survived SIGKILL; aborting to avoid duplicate processes"
             exit 1
         fi
-    elif [[ -n "$WORKER_TYPE" && "$WORKER_TYPE" != "all" ]]; then
+    elif [[ -n "$WORKER_TYPE" && "$WORKER_TYPE" != "all" && "${WORKERS[$WORKER_TYPE]:-}" != "all" ]]; then
```
muhammad-ali-e, line 1170: **[Critical] `-r pg` restart spawns duplicate role consumers that double-poll Postgres.** (The defect is in the kill block just above, lines 1151-1152.)

The PG-queue set is now the 4 named roles + reaper (`run_pg_queue_set` builds `members=("${!PG_CONSUMER_ROLES[@]}" "$PG_QUEUE_REAPER_TYPE")`, line 991). But the set-restart branch only kills the **generic** consumer + reaper:
```bash
kill_one_worker "$PG_QUEUE_CONSUMER_TYPE" || restart_failed=1
kill_one_worker "$PG_QUEUE_REAPER_TYPE"   || restart_failed=1
```
`kill_one_worker "$PG_QUEUE_CONSUMER_TYPE"` resolves to the *end-anchored* generic pgrep pattern, which by design does **not** match the role processes (`... -m pg_queue_consumer <role>`). So `./run-worker.sh -r pg` leaves the 4 running role consumers untouched, then `run_pg_queue_set` launches a *second* copy of each — exactly the double-poll the branch comment claims to prevent, and `restart_failed` never trips because the survivors are never targeted.

**Fix** — kill the same member set the launcher starts:
```bash
restart_failed=0
for member in "${!PG_CONSUMER_ROLES[@]}" "$PG_QUEUE_REAPER_TYPE"; do
    kill_one_worker "$member" || restart_failed=1
done
```
(`-k` and restart-all are unaffected — `kill_workers` iterates the role dirs.)

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -114,14 +120,19 @@ def __init__(
...
+        """
         self._last_poll_monotonic = time.monotonic()
-        messages = self._client.read(
-            self.queue_name, vt_seconds=self.vt_seconds, qty=self.batch_size
-        )
-        for message in messages:
-            self._handle(message)
-        return len(messages)
+        total = 0
+        for queue_name in self.queue_names:
```
muhammad-ali-e, line 137: **[High] `poll_once` has no per-queue error isolation — one queue raising mid-cycle aborts the rest.**

With `queue_names=["file_processing", "api_file_processing"]`: if queue A is drained+acked, then queue B's `read()` raises (transient DB blip / connection recycle / serialization failure), the exception propagates out of `poll_once`; `run()` catches it but counts the whole cycle as `claimed=0` and takes the **empty-queue backoff path** — even though real work happened. Two harms: (1) healthy queues get throttled to backoff cadence under a partial fault; (2) if an *earlier* queue is the flaky one, later queues in the list are never reached and starve indefinitely — silently voiding the round-robin fairness the docstring promises. The only log is the generic "poll cycle failed" — it never names the failing queue.

**Fix** — isolate each queue:
```python
for queue_name in self.queue_names:
    try:
        messages = self._client.read(queue_name, vt_seconds=self.vt_seconds, qty=self.batch_size)
    except Exception:
        logger.exception("PG-queue read failed for queue %r — other queues continue", queue_name)
        continue
    for message in messages:
        self._handle(message)
    total += len(messages)
```

workers/run-worker.sh
```diff
@@ -78,6 +113,10 @@ declare -A WORKERS=(
     ["$PG_QUEUE_SET"]="$PG_QUEUE_SET"
     ["pg"]="$PG_QUEUE_SET"
     ["all"]="all"
+    # 'celery' is a run alias for the Celery set (== 'all', which excludes the PG
+    # workers) — symmetric with 'pg'. Maps to "all" so it dispatches to
+    # run_all_workers and list_core_worker_dirs skips it (not a phantom dir).
+    ["$CELERY_SET"]="all"
```
muhammad-ali-e, line 119: **[High] This new `celery` run alias contradicts the stale comment at lines 75-76.**

The `CELERY_SET` block still reads *"Tail-only — there is no 'celery' run alias; this set is started via 'all'…"*, but this line adds exactly that alias (and the help text says `all, celery`, and the dispatch branch maps it to `run_all_workers`). The comment now actively misleads. Update lines 75-76 to note that since 9f, `celery` is both a `-L` tail alias **and** a run alias mapping to `all`.

workers/run-worker.sh
```diff
@@ -931,11 +1005,12 @@ run_pg_queue_set() {
...
         print_status $RED "PG-queue set: a member failed to start — tearing down the set (see logs above)"
         # Don't leave a survivor: a restart-on-failure relaunch would spawn a
-        # second instance on top of it (the consumer would double-poll Postgres).
-        # Kill both members (the crashed one is already gone → no-op). Same
+        # second instance on top of it (a consumer would double-poll Postgres).
+        # Kill all members (a crashed one is already gone → no-op). Same
         # all-or-nothing discipline as the restart path.
-        kill_one_worker "$PG_QUEUE_CONSUMER_TYPE"
-        kill_one_worker "$PG_QUEUE_REAPER_TYPE"
+        for worker in "${members[@]}"; do
```
muhammad-ali-e, line 1017: **[Medium-High] Start-failure teardown discards `kill_one_worker` return values — a survivor isn't surfaced.**

`kill_one_worker` returns 1 when a process survives SIGKILL, but here the return is dropped. The comment three lines up warns a survivor would cause a second consumer to double-poll Postgres — yet the teardown never verifies it actually avoided one. The **restart** path (lines 1150-1156) aggregates these failures and aborts with a distinct message; this start-failure teardown should mirror it:
```bash
local teardown_failed=0
for worker in "${members[@]}"; do
    kill_one_worker "$worker" || teardown_failed=1
done
[[ $teardown_failed -ne 0 ]] && print_status $RED "PG-queue set: a member survived teardown — a relaunch may double-poll Postgres; investigate before restarting"
```

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -97,7 +97,13 @@ def __init__(
...
                 f"backoff_max ({backoff_max}) must be >= poll_interval ({poll_interval})"
             )
-        self.queue_name = queue_name
+        if not queue_names:
+            raise ValueError("queue_names must be a non-empty list")
+        # One process can drain several queues (9f) — e.g. a file_processing
+        # consumer drains both file_processing and api_file_processing. Polled
+        # round-robin so the per-queue (queue_name, priority DESC, msg_id) dequeue
+        # index keeps its top-N efficiency and no queue starves another.
+        self.queue_names = queue_names
```
muhammad-ali-e: **[Medium] Duplicate queue names silently double-read each cycle; the validated list is also stored as a mutable public alias.**

- `["q", "q"]` (e.g. env `WORKER_PG_QUEUE_CONSUMER_QUEUE="q,q"` — `_parse_queue_list` drops blanks but not dupes) makes `poll_once` `read()` the same queue twice per cycle, claiming up to `2×batch_size` and defeating the fairness rationale in the comment just above.
- `self.queue_names = queue_names` aliases the caller's list by reference as a public mutable attribute — a later `.clear()`/`.append()` bypasses the non-empty guard.

**Fix** — normalize once and store an immutable copy (`list` is the right *shape* — order is load-bearing — but the wrong *strength*):
```python
deduped = list(dict.fromkeys(q.strip() for q in queue_names if q.strip()))
if not deduped:
    raise ValueError("queue_names must contain at least one non-blank name")
if len(deduped) != len(queue_names):
    logger.warning("queue list normalized %r -> %r", queue_names, deduped)
self.queue_names: tuple[str, ...] = tuple(deduped)
```
This also unifies the two construction paths (direct `__init__` vs `_parse_queue_list`), which currently filter blanks inconsistently.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -296,6 +307,13 @@ def _install_signal_handlers(self) -> None:
             )
 
 
+def _parse_queue_list(raw: str) -> list[str]:
```
muhammad-ali-e, line 326: **[Medium] `_parse_queue_list` silently drops empty entries.**

`"a,,b"` → `["a","b"]` with no signal. A doubled/trailing comma is almost always a config typo, and the result is a working-but-not-quite consumer with zero log evidence. Given the project's no-silent-failures stance, warn when entries are dropped so a malformed list is diagnosable from logs rather than only by eyeballing the startup line. (Edge note: a var set to `","`/`" "` returns `[]` → fails the non-empty guard at construction — intended fail-loud, but untested.)

workers/run-worker.sh
```diff
@@ -915,9 +986,12 @@ run_pg_queue_set() {
     local concurrency=$2
     local pool_type=$3
 
-    print_status $GREEN "Starting PG-queue set (consumer + reaper)..."
+    # The set = the named pipeline consumer roles (9f) + the reaper. Each role is
```
muhammad-ali-e: **[Medium] Comment inaccuracy: not all four roles are multi-queue.** *"Each role is a multi-queue consumer covering ETL+API"* — only `pg-fileproc` and `pg-callback` drain two queues; `pg-orchestrator-api` binds the single queue `celery_api_deployments` and `pg-orchestrator-general` binds `celery`. The block at lines 50-51 states this correctly, so this summary contradicts it.

Two more stale comments nearby:
- Line 984 (`run_pg_queue_set` header) still says *"(consumer + reaper)"* — the body now launches 4 roles + reaper.
- Line 982 (and 1230) justify the launch-check with *"the reaper has no health port"* — after 9f the four role consumers also have no health port (line 850 resolves empty for them), so the launch check is the only startup signal for **all five** members, not just the reaper.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -97,7 +97,13 @@ def __init__(
             raise ValueError(
                 f"backoff_max ({backoff_max}) must be >= poll_interval ({poll_interval})"
             )
-        self.queue_name = queue_name
+        if not queue_names:
+            raise ValueError("queue_names must be a non-empty list")
+        # One process can drain several queues (9f) — e.g. a file_processing
```
muhammad-ali-e, line 102: **[Low-Medium] Comment overstates the guarantee: this is a fixed-order sweep, not a rotating round-robin.** `poll_once` always iterates `self.queue_names` from index 0 with no head rotation between cycles, and `_handle` runs tasks sequentially in-process. So an always-full `queue_names[0]` claims its full `batch_size` and runs it before later queues every cycle. The code prevents permanent *starvation* (every queue gets one `read()` per cycle) but does **not** equalize throughput. Reword to "each queue is read once per cycle in list order — guarantees visitation, not equal throughput; with batch_size > 1 an always-full earlier queue is served first each cycle."

workers/run-worker.sh
```diff
@@ -770,17 +834,24 @@ run_worker() {
...
     # SKIP LOCKED, not a Celery/RabbitMQ worker) — override the celery command
     # with the bootstrapping launcher and route the queue via env.
-    if [[ "$worker_type" == "$PG_QUEUE_CONSUMER_TYPE" ]]; then
+    pg_consumer_role="${PG_CONSUMER_ROLES[$worker_type]:-}"
+    if [[ "$worker_type" == "$PG_QUEUE_CONSUMER_TYPE" || -n "$pg_consumer_role" ]]; then
+        # A named 9f role bakes in its source worker-type + multi-queue list
+        # ("<type>;<queues>"); the generic pg-queue-consumer reads them from env
+        # (default source worker = notification, the first migrated leaf task).
+        if [[ -n "$pg_consumer_role" ]]; then
+            export WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE="${pg_consumer_role%%;*}"
```
muhammad-ali-e: **[Low/cleanup] Redundant `export` keyword.** For a role this exports `WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE`, then line 847 unconditionally re-exports it (the `:-notification` default is a no-op once this set it). Dropping `export` here (plain assignment) keeps behavior identical — line 847 is the single export point for both the role and generic paths. Do **not** collapse the two lines: line 847's `:-` is the load-bearing default/override for the generic consumer path.

workers/run-worker.sh
```diff
@@ -804,7 +875,7 @@ run_worker() {
     # Change to appropriate directory
     # For pluggable workers, stay at workers root to allow module imports
     # For core workers, change to worker directory
-    if [[ -n "${PLUGGABLE_WORKERS[$worker_type]:-}" || "$worker_type" == "$PG_QUEUE_CONSUMER_TYPE" || "$worker_type" == "$PG_QUEUE_REAPER_TYPE" ]]; then
+    if [[ -n "${PLUGGABLE_WORKERS[$worker_type]:-}" || "$worker_type" == "$PG_QUEUE_CONSUMER_TYPE" || "$worker_type" == "$PG_QUEUE_REAPER_TYPE" || -n "${PG_CONSUMER_ROLES[$worker_type]:-}" ]]; then
         # Run from the workers root so `python -m pg_queue_consumer` /
```
muhammad-ali-e, line 881: **[Low/cleanup] The cwd predicate hand-rolls the PG-member set.** The condition on the line below enumerates `PG_QUEUE_CONSUMER_TYPE`/`PG_QUEUE_REAPER_TYPE`/`PG_CONSUMER_ROLES` — which is exactly what the existing `PG_QUEUE_MEMBERS` map (the documented single source of truth) encodes. Simplify to:
```bash
if [[ -n "${PLUGGABLE_WORKERS[$worker_type]:-}" || -n "${PG_QUEUE_MEMBERS[$worker_type]:-}" ]]; then
```
Behavior-identical, and a future PG member then only needs the map updated.

workers/run-worker.sh
```diff
@@ -1098,43 +1187,51 @@ load_env "$ENV_FILE"
...
+# (run_all_workers) `wait`s on its background jobs and would block any later type.
+if [[ ${#WORKER_TYPES[@]} -gt 1 && "$DETACH" != "true" ]]; then
+    print_status $YELLOW "Multiple worker types given without -d/--detach: a foreground set will block the rest. Re-run with -d to start them all."
 fi
+
+# Run each requested worker / set in turn.
+for wt in "${WORKER_TYPES[@]}"; do
+    if [[ "${WORKERS[$wt]:-}" == "all" ]]; then
+        # 'all' and its synonym 'celery' (the Celery set, excludes PG workers).
+        run_all_workers "$DETACH" "$LOG_LEVEL" "$CONCURRENCY" "$POOL_TYPE"
```
muhammad-ali-e, line 1239: **[Low] Failure-propagation asymmetry in the `all pg` dual-launch.** The PG-set branch uses `run_pg_queue_set ... || exit 1`, but this Celery `all` branch calls `run_all_workers` with no `|| exit 1`, and that function backgrounds each worker in a `( … ) &` subshell whose crash-on-start status is lost (the `wait` isn't checked). So in `-d all pg`, a PG member crash fails the script but a Celery worker crash does not. Now that `all pg` is a documented first-class flow, consider making the two transports' startup signalling symmetric (follow-up; not introduced by this PR).

workers/run-worker.sh
```diff
@@ -149,14 +188,18 @@ WORKER_TYPE:
     scheduler, schedule   Run scheduler worker (scheduled pipeline tasks)
     executor              Run executor worker (extraction execution tasks)
     ide-callback          Run IDE callback worker (Prompt Studio post-execution callbacks)
-    pg-queue-consumer     Run PG-queue poll-loop consumer (opt-in; not part of 'all')
+    pg-queue-consumer     Run a generic PG-queue poll-loop consumer (env-configured; opt-in)
+    pg-orchestrator-api   Run the PG orchestrator consumer for API execs (celery_api_deployments)
+    pg-orchestrator-general Run the PG orchestrator consumer for ETL/general execs (celery)
+    pg-fileproc           Run the PG fan-out consumer (file_processing + api_file_processing)
+    pg-callback           Run the PG callback consumer (file_processing_callback + api_…_callback)
```
muhammad-ali-e: **[Low] Help text uses a literal ellipsis in a queue name.** `api_…_callback` abbreviates `api_file_processing_callback`. A non-copy-pasteable queue name in user-facing help is mildly risky — spell it in full or drop the parenthetical if width is the constraint.

workers/tests/test_pg_queue_consumer.py
```diff
@@ -404,5 +404,39 @@ def test_enqueue_poll_execute_ack(self, pg_client):
...
+        ]
+        claimed = PgQueueConsumer(["qa", "qb"], client=client).poll_once()
+        assert claimed == 3
+        # Read once per queue, in order.
+        assert [c.args[0] for c in client.read.call_args_list] == ["qa", "qb"]
+
+    def test_empty_queue_does_not_starve_the_others(self):
+        client = MagicMock()
+        client.read.side_effect = [[], [_msg(1, _ok_payload(1))]]
+        assert PgQueueConsumer(["empty", "busy"], client=client).poll_once() == 1
```
muhammad-ali-e, line 425: **[Medium] Test-coverage gaps for the new multi-queue behavior.** The added `TestMultiQueue` covers read order + aggregation + empty-list guard — good — but misses:
- **One queue raising mid-cycle** (the load-bearing partial-failure path; see the `poll_once` comment). `read.side_effect = [[_msg(...)], RuntimeError("boom")]` for `["qa","qb"]` and assert qa was acked + the documented cycle semantics.
- **`batch_size` accounting**: `test_round_robin_...` never asserts `qty`. Construct with `batch_size=2` and assert every `read` was called with `qty=2` and the total sums across queues.
- **Duplicate queue names** (double-read), and `_parse_queue_list` for all-empty/comma-only input (`""`, `","`, `",a,"`).
- **No shell-test harness exists for `run-worker.sh`** — the riskiest new logic (role→queue/source-type split `${role%%;*}`/`${role#*;}`, the role-anchored vs end-anchored pgrep patterns, the `celery` alias and multi-positional dispatch) has zero coverage. A wrong split silently drains the wrong queues; a wrong pgrep anchor kills the wrong process. Consider a bats test (or a `bash -c` driver) sourcing the script with `run_worker` stubbed to echo the exported env.

muhammad-ali-e (commented):
Automated PR review (PR Review Toolkit: code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier). Findings below are inline. The single blocker is the `-r pg` restart bug (duplicate role consumers double-polling Postgres); everything else is High/Medium/Low. Line anchors point at the nearest diffed line — exact `file:line` is called out in each comment body where it differs.

workers/run-worker.sh
muhammad-ali-e: Fixed in 9b3905e5e — the `-r pg` restart branch now kills the SAME members `run_pg_queue_set` launches (the 4 named roles + reaper), looping `${!PG_CONSUMER_ROLES[@]}` + reaper, instead of the generic consumer. Dev-tested: `-r pg` killed + restarted all 4 roles (new PIDs), no survivor.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in 9b3905e5e — `poll_once` now wraps each queue in try/except: a read/handle failure on one queue is logged and skipped, the others still run, and the work already done this cycle still counts (so `run()` resets backoff instead of taking the empty path). +test `test_one_queue_failing_does_not_abort_the_others`.

workers/run-worker.sh
muhammad-ali-e: Fixed in 9b3905e5e — updated the CELERY_SET comment: since 9f, `celery` is BOTH a `-L` tail alias AND a run alias (maps to `all` → run_all_workers), symmetric with `pg`.

workers/run-worker.sh
muhammad-ali-e: Fixed in 9b3905e5e — the start-failure teardown now aggregates `kill_one_worker` returns (`teardown_failed`) and prints a distinct survivor/double-poll warning, mirroring the restart path.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in 9b3905e5e — `self.queue_names = list(dict.fromkeys(queue_names))`: order-preserving de-dup (a duplicate would double-read per cycle) AND a copy (a later caller mutation can no longer bypass the non-empty validation). +test `test_duplicate_queues_are_deduped`.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in 9b3905e5e — `_parse_queue_list` now warns with the offending raw value when it drops empty entries, so a doubled/trailing comma is diagnosable from the logs. (The all-empty `","` → `[]` → non-empty-guard fail-loud path stands.)

workers/run-worker.sh
muhammad-ali-e: Fixed in 9b3905e5e — reworded: only fileproc/callback are multi-queue; the two orchestrator roles bind a single registry-specific queue. Also fixed the `run_pg_queue_set` header ("4 pipeline roles + reaper") and the members comment.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in 9b3905e5e — reworded to "each queue is read once per cycle in list order: prevents starvation (every queue gets a read each cycle) but does NOT equalize throughput (an always-full queue_names[0] claims its full batch first)". Dropped the "round-robin" overstatement.

workers/run-worker.sh
muhammad-ali-e: Fixed in 9b3905e5e — the role branch now does a plain assignment; the line below remains the single `export` point (its `:-notification` default/override is load-bearing for the generic path, as you noted — kept separate).

workers/run-worker.sh
muhammad-ali-e: Fixed in 9b3905e5e — the cwd predicate now uses `${PG_QUEUE_MEMBERS[$worker_type]:-}` (the documented single source of truth) instead of hand-rolling consumer/reaper/roles. A future PG member just updates the map.

workers/run-worker.sh
muhammad-ali-e: Deferring (you flagged follow-up). The asymmetry is pre-existing in `run_all_workers`: it backgrounds each worker in a `( … ) &` subshell and `wait`s without checking per-worker crash status, so the Celery `all` path has no startup signal to propagate — adding `|| exit 1` here would not capture it. Making the two transports symmetric means reworking `run_all_workers` subshell-status capture; worth a dedicated follow-up rather than folding into 9f.

workers/run-worker.sh
muhammad-ali-e: Fixed in 9b3905e5e — spelled out `api_file_processing_callback` in full (no ellipsis) in the help.

workers/tests/test_pg_queue_consumer.py
muhammad-ali-e: Fixed in 9b3905e5e — added `test_one_queue_failing_does_not_abort_the_others` (partial-failure: qa acked + counted, qb raises, no propagation) and `test_read_uses_batch_size_qty_per_queue` (asserts every `read` gets `qty=batch_size`). Plus `test_duplicate_queues_are_deduped`.

muhammad-ali-e:
## Review round addressed — `9b3905e5e`

Thanks for the deep pass. All 13 handled:

**Correctness (Critical/High):**
- `-r pg` now kills the 4 roles + reaper (was leaving the roles → double-poll on relaunch); dev-tested.
- `poll_once` isolates each queue (one queue failing no longer aborts the others or triggers a false empty-backoff).
- de-dup + copy `queue_names`; `run_pg_queue_set` teardown surfaces a survivor.

**Comments/cleanup:** fixed the stale "no celery run alias" comment, the "each role is multi-queue" / "consumer + reaper" / "round-robin" inaccuracies; dropped the redundant `export`; cwd predicate now uses `PG_QUEUE_MEMBERS`; `_parse_queue_list` warns on dropped entries; spelled out the queue name in help.

**Tests:** +partial-failure isolation, +batch_size qty per queue, +duplicate dedup.

**Deferred (1, you flagged follow-up):** the `all pg` startup-signal asymmetry is pre-existing in `run_all_workers` (subshell `wait` without status capture) — a dedicated follow-up, not 9f.

PR #2074, opened 2026-06-18
UN-3576 [FEAT] PG Queue 9g — docker-compose PG consumer + reaper services (gated profile)

muhammad-ali-e:
## What & why

Makes the already-built PG-queue pipeline (orchestrator, fan-out, barrier, callback on Postgres) runnable under **docker compose**, **gated OFF by default** — so it runs in containers, not just the host `run-worker.sh`. Each component is its own service (one-service-per-component), so the eventual Kubernetes mapping is a mechanical "one service → one Deployment".

This is the docker analog of an ops rollout: the services *exist and are runnable*, but no traffic moves until the backend gate (`PG_QUEUE_TRANSPORT_ENABLED`) + Flipt flag are flipped (a later step).

## Changes

- **`workers/run-worker-docker.sh`** (the image ENTRYPOINT) — recognises two new launch types and execs the dedicated Python module instead of building a Celery worker command:
  - `pg-queue-consumer` / `pg-consumer` → `python -m pg_queue_consumer`
  - `pg-queue-reaper` / `pg-reaper` / `reaper` → `python -m pg_queue_reaper`
  - Dispatched *before* the Celery two-path logic; existing worker types are unaffected. No Dockerfile change (the script is already the entrypoint).
- **`docker/docker-compose.yaml`** — 5 services behind the off-by-default `pg-queue` profile:
  - `worker-pg-orchestrator-api` (api_deployment; `celery_api_deployments`)
  - `worker-pg-orchestrator-general` (general; `celery`)
  - `worker-pg-fileproc` (file_processing; `file_processing,api_file_processing`)
  - `worker-pg-callback` (callback; both callback queues)
  - `worker-pg-reaper` (single-instance, leader-elected)
  - The orchestrators + reaper are broker-free. `fileproc` + `callback` keep a `rabbitmq` dependency because they still hand off to the Celery executor (tool-execution RPC) / notifications until those are migrated.
- **`docker/sample.env`** — documents the new env vars + the profile, and that the backend gate stays off.

## Non-regression

- Default `docker compose up` starts **zero** PG services (profile off) — behaviour identical to today.
- Even with the PG services running, executions still route to Celery until the gate + flag are flipped.

## Dev-test (passed end-to-end)

Built `worker-unified:dev`, `docker compose --profile pg-queue up -d`:

- All 5 services boot and stay up; each consumer registers the correct worker-type tasks and polls the correct queues; consumer/reaper liveness endpoints healthy.
- Reaper leader-election verified: deferred as standby while another reaper held the lease, then acquired leadership once that stopped.
- Triggered an API deployment → the full pipeline drained **through the containerized consumers**: orchestrator ran `async_execute_bin` → `process_file_batch` → PgBarrier decrement `remaining=0` → callback → execution **COMPLETED**; barrier/dedup tables cleaned up.

Targets `feat/UN-3445-pg-queue-integration` (not `main`). Sub-task UN-3576.

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -541,6 +541,167 @@ services:
       - ${TOOL_REGISTRY_CONFIG_SRC_PATH}:/data/tool_registry_config
       - prompt_studio_data:/app/prompt-studio-data
 
+  # ===========================================================================
+  # PG-queue services (Postgres-backed transport) — gated behind the `pg-queue`
+  # compose profile, so a plain `docker compose up` starts NONE of them and
+  # behaviour is identical to today. Bring them up with:
+  #   docker compose --profile pg-queue up -d
+  # They consume from Postgres, not the broker. The orchestrator consumers and
+  # the reaper are fully broker-free (no rabbitmq dependency). The fan-out and
+  # callback consumers still hand off to the Celery executor (tool execution
+  # RPC) / Celery notifications, so they keep a rabbitmq dependency until those
+  # are migrated (steps ③/④). Each maps 1:1 to a future K8s Deployment.
+  # Executions still route to Celery until the backend gate
+  # (PG_QUEUE_TRANSPORT_ENABLED) + Flipt flag are flipped — that is the ramp,
+  # not this change.
+  # ===========================================================================
+
+  # Orchestrator consumer for API-deployment executions (async_execute_bin_api).
+  worker-pg-orchestrator-api:
+    image: unstract/worker-unified:${VERSION}
+    container_name: unstract-worker-pg-orchestrator-api
+    restart: unless-stopped
+    command: ["pg-queue-consumer"]
+    ports:
+      - "8093:8090"
+    env_file:
+      - ../workers/.env
+      - ./essentials.env
+    depends_on:
+      - db
+      - redis
+    environment:
+      - ENVIRONMENT=development
+      - APPLICATION_NAME=unstract-worker-pg-orchestrator-api
+      - WORKER_BARRIER_BACKEND=pg
+      - WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE=api_deployment
+      - WORKER_PG_QUEUE_CONSUMER_QUEUE=celery_api_deployments
+      - WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT=8090
+    labels:
+      - traefik.enable=false
+    volumes:
+      - ./workflow_data:/data
+      - ${TOOL_REGISTRY_CONFIG_SRC_PATH}:/data/tool_registry_config
+    profiles:
+      - pg-queue
+
+  # Orchestrator consumer for general / scheduled-ETL executions (async_execute_bin).
+  worker-pg-orchestrator-general:
+    image: unstract/worker-unified:${VERSION}
+    container_name: unstract-worker-pg-orchestrator-general
+    restart: unless-stopped
+    command: ["pg-queue-consumer"]
+    ports:
+      - "8094:8090"
+    env_file:
+      - ../workers/.env
+      - ./essentials.env
+    depends_on:
+      - db
+      - redis
+    environment:
+      - ENVIRONMENT=development
+      - APPLICATION_NAME=unstract-worker-pg-orchestrator-general
+      - WORKER_BARRIER_BACKEND=pg
+      - WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE=general
+      - WORKER_PG_QUEUE_CONSUMER_QUEUE=celery
+      - WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT=8090
+    labels:
+      - traefik.enable=false
+    volumes:
+      - ./workflow_data:/data
+      - ${TOOL_REGISTRY_CONFIG_SRC_PATH}:/data/tool_registry_config
+    profiles:
+      - pg-queue
+
+  # Fan-out consumer (process_file_batch) — both the ETL and API file queues.
+  worker-pg-fileproc:
+    image: unstract/worker-unified:${VERSION}
+    container_name: unstract-worker-pg-fileproc
+    restart: unless-stopped
+    command: ["pg-queue-consumer"]
+    ports:
+      - "8095:8090"
+    env_file:
+      - ../workers/.env
+      - ./essentials.env
+    depends_on:
+      - db
+      - redis
+      - rabbitmq   # still dispatches the tool-execution RPC to the Celery executor (until ③)
+    environment:
+      - ENVIRONMENT=development
+      - APPLICATION_NAME=unstract-worker-pg-fileproc
+      - WORKER_BARRIER_BACKEND=pg
+      - WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE=file_processing
+      - WORKER_PG_QUEUE_CONSUMER_QUEUE=file_processing,api_file_processing
+      - WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT=8090
+    labels:
+      - traefik.enable=false
+    volumes:
+      - ./workflow_data:/data
+      - ${TOOL_REGISTRY_CONFIG_SRC_PATH}:/data/tool_registry_config
+    profiles:
+      - pg-queue
+
+  # Fan-in callback consumer (process_batch_callback / _api) — both callback queues.
+  worker-pg-callback:
+    image: unstract/worker-unified:${VERSION}
+    container_name: unstract-worker-pg-callback
+    restart: unless-stopped
+    command: ["pg-queue-consumer"]
+    ports:
+      - "8096:8090"
+    env_file:
+      - ../workers/.env
+      - ./essentials.env
+    depends_on:
+      - db
+      - redis
+      - rabbitmq   # may dispatch Celery notifications when not PG-routed (until ④)
+    environment:
+      - ENVIRONMENT=development
+      - APPLICATION_NAME=unstract-worker-pg-callback
+      - WORKER_BARRIER_BACKEND=pg
+      - WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE=callback
+      - WORKER_PG_QUEUE_CONSUMER_QUEUE=file_processing_callback,api_file_processing_callback
+      - WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT=8090
+    labels:
+      - traefik.enable=false
+    volumes:
+      - ./workflow_data:/data
+      - ${TOOL_REGISTRY_CONFIG_SRC_PATH}:/data/tool_registry_config
+    profiles:
+      - pg-queue
+
+  # Reaper — leader-elected recovery loop. Run exactly ONE instance (it elects a
+  # single leader via orchestrator_lock; extra replicas idle as standby).
+  worker-pg-reaper:
+    image: unstract/worker-unified:${VERSION}
+    container_name: unstract-worker-pg-reaper
+    restart: unless-stopped
+    command: ["pg-queue-reaper"]
+    ports:
+      - "8097:8086"
+    env_file:
+      - ../workers/.env
+      - ./essentials.env
+    depends_on:
+      - db
+      - redis
+    environment:
+      - ENVIRONMENT=development
+      - APPLICATION_NAME=unstract-worker-pg-reaper
+      - WORKER_PG_REAPER_INTERVAL_SECONDS=${WORKER_PG_REAPER_INTERVAL_SECONDS:-5}
+      - WORKER_PG_REAPER_HEALTH_PORT=8086
+    labels:
+      - traefik.enable=false
+    volumes:
+      - ./workflow_data:/data
+      - ${TOOL_REGISTRY_CONFIG_SRC_PATH}:/data/tool_registry_config
+    profiles:
+      - pg-queue
+
 volumes:
   prompt_studio_data:
   unstract_data:
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -547,10 +547,12 @@ services:
   # behaviour is identical to today. Bring them up with:
   #   docker compose --profile pg-queue up -d
   # They consume from Postgres, not the broker. The orchestrator consumers and
-  # the reaper are fully broker-free (no rabbitmq dependency). The fan-out and
-  # callback consumers still hand off to the Celery executor (tool execution
-  # RPC) / Celery notifications, so they keep a rabbitmq dependency until those
-  # are migrated (steps ③/④). Each maps 1:1 to a future K8s Deployment.
+  # the reaper are broker-free for PG-routed executions (fan-out honors the
+  # carried transport; a missing transport fails closed to Celery, which these
+  # consumers aren't expected to receive). The fan-out and callback consumers
+  # still hand off to the Celery executor (tool-execution RPC) / Celery
+  # notifications, so they keep a rabbitmq dependency until the executor and
+  # notifications themselves move to PG. Each maps 1:1 to a future K8s Deployment.
   # Executions still route to Celery until the backend gate
   # (PG_QUEUE_TRANSPORT_ENABLED) + Flipt flag are flipped — that is the ramp,
   # not this change.
@@ -628,7 +630,7 @@ services:
     depends_on:
       - db
       - redis
-      - rabbitmq   # still dispatches the tool-execution RPC to the Celery executor (until ③)
+      - rabbitmq   # still dispatches the tool-execution RPC to the Celery executor (until the executor/tool-RPC moves to PG)
     environment:
       - ENVIRONMENT=development
       - APPLICATION_NAME=unstract-worker-pg-fileproc
@@ -658,7 +660,7 @@ services:
     depends_on:
       - db
       - redis
-      - rabbitmq   # may dispatch Celery notifications when not PG-routed (until ④)
+      - rabbitmq   # may dispatch Celery notifications when not PG-routed (until notifications move to PG)
     environment:
       - ENVIRONMENT=development
       - APPLICATION_NAME=unstract-worker-pg-callback
@@ -675,7 +677,7 @@ services:
       - pg-queue
 
   # Reaper — leader-elected recovery loop. Run exactly ONE instance (it elects a
-  # single leader via orchestrator_lock; extra replicas idle as standby).
+  # single leader via pg_orchestrator_lock; extra replicas idle as standby).
   worker-pg-reaper:
     image: unstract/worker-unified:${VERSION}
     container_name: unstract-worker-pg-reaper
--- a/workers/run-worker-docker.sh
+++ b/workers/run-worker-docker.sh
@@ -567,10 +567,12 @@ case "${1:-}" in
     pg-queue-reaper|pg-reaper|reaper)
         run_pg_reaper
         ;;
-    pg-*|*-reaper)
+    pg-*)
         # Obviously-PG-intended but unrecognized (e.g. a typo'd command) — fail
         # loudly instead of silently coercing it into a default Celery worker.
-        # (No legitimate worker type starts with `pg-` or ends with `-reaper`.)
+        # Scoped to the `pg-` prefix (reserved for PG-queue components) so it
+        # never intercepts a pluggable worker name; the exact reaper aliases are
+        # already matched above.
         print_status $RED "Unrecognized PG-queue command: '$1' (did you mean pg-queue-consumer / pg-queue-reaper?)."
         exit 1
         ;;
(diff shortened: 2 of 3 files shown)


workers/run-worker-docker.sh
```diff
@@ -25,6 +25,11 @@ ENV_FILE="/app/.env"
 # Worker type constant for the executor worker
 readonly EXECUTOR_WORKER_TYPE="executor"
 
+# Python interpreter for PG-queue components (consumer / reaper). Unlike Celery
+# workers, these launch a dedicated module rather than a `celery ... worker`
+# command (see run_pg_consumer / run_pg_reaper below).
+readonly PG_QUEUE_PYTHON_BIN="/app/.venv/bin/python"
```
muhammad-ali-e, line 31: **[Medium · Silent Failure]** `PG_QUEUE_PYTHON_BIN` is hard-coded and never validated before the `exec` in `run_pg_consumer`/`run_pg_reaper`. If the venv path moves in an image refactor, `exec` dies with a terse `not found`, and `restart: unless-stopped` turns it into a silent crash-loop (consumer/reaper health ports are opt-in / disabled by default, so there's no health signal). Suggest guarding before exec:
```bash
if [[ ! -x "$PG_QUEUE_PYTHON_BIN" ]]; then
    print_status $RED "PG-queue interpreter not found/executable: $PG_QUEUE_PYTHON_BIN — check the image build / venv path."
    exit 1
fi
```

workers/run-worker-docker.sh
```diff
@@ -486,6 +491,39 @@ run_worker() {
...
+# PG-queue components (Postgres-backed transport)
+# =============================================================================
+# These do NOT run a Celery worker — they exec a dedicated Python module that
+# polls the Postgres queue. The consumer picks which worker's tasks to register
+# and which queues to poll from the environment (set per compose service /
+# K8s Deployment); the reaper needs neither. Mirrors the host launcher
+# run-worker.sh (`pg-queue-consumer` / `reaper`).
+
+run_pg_consumer() {
+    local source_type="${WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE:-notification}"
```
muhammad-ali-e, line 515: **[Medium · Silent Failure]** Defaulting `source_type` to `notification` when `WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE` is unset silently mislabels a misconfigured consumer (e.g. a hand-rolled K8s Deployment or `docker run` that forgets the var) instead of refusing to start. It only sets the `WORKER_NAME` label here — the module re-reads the var itself — so the shell label and the module's actual role can diverge. All compose services in this PR set it explicitly, so the risk is non-compose launches. Suggest warning loudly when unset:
```bash
if [[ -z "${WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE:-}" ]]; then
    print_status $YELLOW "WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE unset — WORKER_NAME label may not reflect actual role."
fi
```

muhammad-ali-e (commented):
## PR Review Toolkit — automated review (#2074)

Ran six specialized agents (Code Reviewer, Silent Failure Hunter, Type Design Analyzer, PR Test Analyzer, Comment Analyzer, Code Simplifier) over the 3 changed files. **No Critical/High issues** — the profile gating is correct and a plain `docker compose up` is a true no-op. The inline comments below are the actionable items, dominated by *defaults/dispatch that mask operator misconfiguration* in `run-worker-docker.sh` and a few comment-accuracy nits in `docker-compose.yaml`.

Non-inline notes (whole-file / repo-wide, can't anchor to a diff line):
- **`run-worker-docker.sh` lacks `set -u`/`set -o pipefail`** (only `set -e` at line 10). The new functions lean heavily on `${VAR:-default}` expansion; `set -u` would harden against future unguarded reads. Verify the rest of the script tolerates `-u` first.
- **`depends_on` uses short-form (no `condition: service_healthy`)** on all 5 services — consumers can start before db/redis/rabbitmq are ready and crash-loop under `restart: unless-stopped`. This matches the entire existing compose file (zero healthchecks anywhere), so it's a repo-wide convention, not a PR blocker.
- **No shellcheck / `docker compose config` CI** covers these files (hadolint is scoped to `Dockerfile$`). A `shellcheck workers/*.sh` pre-commit hook + a `docker compose config -q` smoke step would give cheap regression protection as both launchers grow.
- Test coverage: nothing actionable — the underlying `pg_queue_consumer`/`pg_queue_reaper` modules are already unit-tested (`workers/tests/test_pg_queue_consumer.py`, `test_pg_reaper.py`); infra config has no test precedent in this repo.

workers/run-worker-docker.sh
```diff
@@ -486,6 +491,39 @@ run_worker() {
...
+    local queues="${WORKER_PG_QUEUE_CONSUMER_QUEUE:-}"
```
muhammad-ali-e, line 516: **[Low · Silent Failure]** `queues` defaults to empty and the script defers entirely to an unspecified "module default", rendering it benignly as `<unset — module default>`. A consumer that ends up polling no queues (or the wrong default) sits idle while looking healthy — the classic silent no-op. Acceptable to keep the empty shell default *only if* the module logs its resolved queue set at startup; otherwise log a WARN here so a forgotten queue env is diagnosable.

workers/run-worker-docker.sh
```diff
@@ -496,6 +534,18 @@ discover_pluggable_workers
 # Add PYTHONPATH for imports - include both /app and /unstract for packages
 export PYTHONPATH="/app:/unstract/core/src:/unstract/connectors/src:/unstract/filesystem/src:/unstract/flags/src:/unstract/tool-registry/src:/unstract/tool-sandbox/src:/unstract/workflow-execution/src:${PYTHONPATH:-}"
 
+# PG-queue components run a dedicated module, not a Celery worker — dispatch
+# them before the Celery command-building logic below. PYTHONPATH (exported
+# above) is required for the module imports.
+case "${1:-}" in
```
muhammad-ali-e, line 563: **[Medium · Silent Failure + readability]** This `case` has no fall-through arm, which is *intentional* (non-PG args continue to the Celery logic below) — but it means a near-miss command like `pg-consumr` or `pgqueue-reaper` matches nothing here, reaches the two-path logic, and gets silently coerced into a default/wrong Celery worker rather than failing loudly. A reaper container started with a typo'd command would quietly run as a `general` worker. Suggest catching obviously-PG-intended tokens while preserving the deliberate pass-through, and add a one-line comment documenting the intentional fall-through:
```bash
case "${1:-}" in
    pg-queue-consumer|pg-consumer) run_pg_consumer ;;
    pg-queue-reaper|pg-reaper|reaper) run_pg_reaper ;;
    pg-*|*-reaper|*-consumer)
        print_status $RED "Unrecognized PG-queue command: '$1'."; exit 1 ;;
    # anything else falls through to the Celery command logic below
esac
```

docker/docker-compose.yaml
```diff
@@ -541,6 +541,167 @@ services:
       - ${TOOL_REGISTRY_CONFIG_SRC_PATH}:/data/tool_registry_config
       - prompt_studio_data:/app/prompt-studio-data
 
+  # ===========================================================================
+  # PG-queue services (Postgres-backed transport) — gated behind the `pg-queue`
+  # compose profile, so a plain `docker compose up` starts NONE of them and
+  # behaviour is identical to today. Bring them up with:
+  #   docker compose --profile pg-queue up -d
+  # They consume from Postgres, not the broker. The orchestrator consumers and
+  # the reaper are fully broker-free (no rabbitmq dependency). The fan-out and
```
muhammad-ali-e: **[Low · Comment accuracy]** "fully broker-free (no rabbitmq dependency)" is true for the intended PG-routed path, but `normalize_transport` fails closed to Celery (`DEFAULT_WORKFLOW_TRANSPORT = CELERY`). If an orchestrator message ever lands on the PG queue *without* a `transport=pg_queue` kwarg, the fan-out attempts a Celery chord and would need the broker these `depends_on` lists omit. Suggest softening: "broker-free for PG-routed executions (fan-out honors the carried `transport`; a missing transport fails closed to Celery, which these consumers aren't expected to receive)".

docker/docker-compose.yaml
```diff
@@ -541,6 +541,167 @@ services:
...
+  # callback consumers still hand off to the Celery executor (tool execution
+  # RPC) / Celery notifications, so they keep a rabbitmq dependency until those
+  # are migrated (steps ③/④). Each maps 1:1 to a future K8s Deployment.
```
muhammad-ali-e: **[Low · Comment rot]** The circled-number scheme "steps ③/④" (also referenced inline at lines 631 `until ③` and 661 `until ④`) is defined nowhere — a repo-wide grep for `③`/`④` returns only these comments. A future maintainer can't resolve what they mean, and the markers will rot once those steps land/renumber. Suggest anchoring to a stable reference (the `queue_backend/pg_queue/9e-design.md` design doc the code already cites, or a Jira ID) or spelling them out inline ("until the executor/tool-RPC moves to PG" / "until notifications move to PG").

docker/docker-compose.yaml
```diff
@@ -541,6 +541,167 @@ services:
...
+      - ../workers/.env
+      - ./essentials.env
+    depends_on:
+      - db
+      - redis
+    environment:
+      - ENVIRONMENT=development
+      - APPLICATION_NAME=unstract-worker-pg-orchestrator-api
+      - WORKER_BARRIER_BACKEND=pg
+      - WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE=api_deployment
```
muhammad-ali-e, line 579: **[Low · Type design]** `WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE` / `WORKER_PG_QUEUE_CONSUMER_QUEUE` set here are exact members of the upstream `WorkerType` / `QueueName` enums, but they're consumed as unvalidated strings: a typo'd worker-type falls back to `WorkerType.GENERAL` via `.get(..., WorkerType.GENERAL)` (`pg_queue_consumer/__main__.py`), and a typo'd queue is accepted by `_parse_queue_list` (`queue_backend/pg_queue/consumer.py`) so the consumer polls a queue nothing publishes to. Contrast `WORKER_BARRIER_BACKEND` (set just above on line 575), which *is* validated against `BarrierBackend` and raises loudly on an unknown value. Worth tightening those two consumption boundaries (upstream, not in this file) to match the barrier-backend enforcement posture.

docker/docker-compose.yaml
```diff
@@ -541,6 +541,167 @@ services:
...
+    labels:
+      - traefik.enable=false
+    volumes:
+      - ./workflow_data:/data
+      - ${TOOL_REGISTRY_CONFIG_SRC_PATH}:/data/tool_registry_config
+    profiles:
+      - pg-queue
+
+  # Reaper — leader-elected recovery loop. Run exactly ONE instance (it elects a
+  # single leader via orchestrator_lock; extra replicas idle as standby).
```
muhammad-ali-e: **[Low · Comment accuracy]** The advisory lock is named `pg_orchestrator_lock` (see `workers/queue_backend/pg_queue/reaper.py`), not `orchestrator_lock`. Minor identifier drift — suggest using the exact name so a maintainer grepping the literal string finds it.

workers/run-worker-docker.sh
muhammad-ali-e: Fixed in c7bfe1b07 — added `ensure_pg_interpreter()` (your suggested `[[ ! -x ]]` guard) called at the top of both `run_pg_consumer`/`run_pg_reaper`, so a moved venv path fails loudly instead of crash-looping under `restart: unless-stopped`.

workers/run-worker-docker.sh
muhammad-ali-e: Fixed in c7bfe1b07 — WARN when `WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE` is unset, noting the WORKER_NAME label may not match the module's actual role. Compose sets it on all 5 services; this covers hand-rolled `docker run` / K8s launches.

workers/run-worker-docker.sh
muhammad-ali-e: Fixed in c7bfe1b07 — added a WARN when `WORKER_PG_QUEUE_CONSUMER_QUEUE` is unset. Your 'only if the module logs its resolved queue set' condition is met: the consumer logs `PG-queue consumer started (queues=[...])` at startup (confirmed in dev-test), so a forgotten queue env is diagnosable both at the shell and in the module log.

workers/run-worker-docker.sh
muhammad-ali-e: Fixed in c7bfe1b07 — added a `pg-*|*-reaper)` arm that fails loudly on obviously-PG-intended typos (`pg-consumr`, `pgqueue-reaper`), plus a comment documenting the deliberate fall-through for everything else. Verified no legit worker type is caught: `log-consumer`, `log`, `file`, `callback`, `notification`, `executor`, `ide-callback`, `general`, `scheduler`, `all` all still route to Celery (routing-precedence test).

docker/docker-compose.yaml
muhammad-ali-e: Fixed in c7bfe1b07 — reworded to 'broker-free for PG-routed executions (fan-out honors the carried transport; a missing transport fails closed to Celery, which these consumers aren't expected to receive)', matching your suggestion.

docker/docker-compose.yaml
muhammad-ali-e: Fixed in c7bfe1b07 — replaced the ③/④ markers (here + the two inline `until ③/④` refs) with spelled-out phrasing: 'until the executor/tool-RPC moves to PG' and 'until notifications move to PG'. No more undefined symbols to rot.

docker/docker-compose.yaml
muhammad-ali-e: Acknowledged — agreed, and intentionally out of scope for this docker-wiring PR (the validation boundaries live upstream in `pg_queue_consumer/__main__.py` `.get(..., WorkerType.GENERAL)` and `queue_backend/pg_queue/consumer.py` `_parse_queue_list`, not in compose). Worth a follow-up to make those raise on unknown values like `WORKER_BARRIER_BACKEND` does against `BarrierBackend`. Mitigated here by the new unset-WARN + the fact all 5 services set exact enum members.

docker/docker-compose.yaml
muhammad-ali-e: Fixed in c7bfe1b07 — using the exact table name `pg_orchestrator_lock` so a grep for the literal resolves.

muhammad-ali-e:
## Review round 1 addressed — `c7bfe1b07`

Thanks for the review. 7 fixed, 1 acknowledged-as-follow-up.

**Fixed (run-worker-docker.sh):**
- **[Med] interpreter guard** — `ensure_pg_interpreter()` fails loudly if the venv python is missing (no silent crash-loop under `restart: unless-stopped`).
- **[Med] unset worker-type** — WARN when `WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE` is unset.
- **[Low] unset queue** — WARN when `WORKER_PG_QUEUE_CONSUMER_QUEUE` is unset; the module also logs its resolved queue set at startup.
- **[Med] near-miss commands** — `pg-*|*-reaper)` arm rejects typo'd PG commands loudly; deliberate fall-through documented. Verified no legit worker type (incl. `log-consumer`) is caught.

**Fixed (docker-compose.yaml):**
- **[Low] "broker-free" wording** — softened to "broker-free for PG-routed executions" (fail-closed-to-Celery on a missing transport).
- **[Low] ③/④ comment rot** — spelled out as "until the executor/tool-RPC moves to PG" / "until notifications move to PG".
- **[Low] lock name** — exact `pg_orchestrator_lock`.

**Acknowledged — follow-up (out of this PR's file scope):**
- **[Low] enum validation** — `WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE`/`_QUEUE` are consumed as unvalidated strings upstream (`pg_queue_consumer/__main__.py`, `consumer.py`), not in compose. Worth tightening those boundaries to raise on unknown values like `WORKER_BARRIER_BACKEND` does — better as its own change; mitigated here by the new unset-WARN.

Script changes validated by `bash -n` + a routing-precedence test; compose by `docker compose config`. The happy-path e2e (containerized pipeline → COMPLETED) was already passing and these are defensive guards/comments only.

muhammad-ali-e:
### Re: the non-inline (whole-file / repo-wide) notes

Thanks — all four are fair; my take on each:

- **`set -u` / `set -o pipefail`** — the new PG functions are written `set -u`-safe (all reads use `${VAR:-default}`), but `run-worker-docker.sh` is the **shared entrypoint for every worker**, and flipping shell options there risks tripping unguarded reads elsewhere in the 590-line script. Per your "verify the rest tolerates `-u` first" caveat, I'd rather not ride a global shell-option change on this docker-wiring PR — better as a focused hardening pass over the whole launcher (with a re-test of all worker types). Noting it for that.
- **`depends_on` short-form (no `condition: service_healthy`)** — agreed, and as you note it matches the entire existing compose file (zero healthchecks anywhere) — a repo-wide convention, not introduced here. The new consumers/reaper self-recover (the reaper's connection is self-healing; consumers retry the poll), so a brief pre-readiness crash-loop self-resolves. Worth a repo-wide healthcheck pass separately.
- **No shellcheck / `docker compose config` CI** — good cheap regression net as both launchers grow; out of scope for this change (CI config) but I'll flag a `shellcheck workers/*.sh` + `docker compose config -q` pre-commit hook as a follow-up.
- **Test coverage** — agreed, nothing actionable: the `pg_queue_consumer`/`pg_queue_reaper` modules are unit-tested already, and infra config has no test precedent here.

None of these change the diff; the inline items are all fixed in `c7bfe1b07`.

workers/run-worker-docker.sh
```diff
@@ -496,6 +557,30 @@ discover_pluggable_workers
...
+    pg-queue-reaper|pg-reaper|reaper)
+        run_pg_reaper
+        ;;
+    pg-*|*-reaper)
+        # Obviously-PG-intended but unrecognized (e.g. a typo'd command) — fail
+        # loudly instead of silently coercing it into a default Celery worker.
+        # (No legitimate worker type starts with `pg-` or ends with `-reaper`.)
+        print_status $RED "Unrecognized PG-queue command: '$1' (did you mean pg-queue-consumer / pg-queue-reaper?)."
+        exit 1
+        ;;
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in d8ac1e601 — narrowed the catch arm from `pg-*|*-reaper` to `pg-*` only. The `pg-` prefix is reserved for PG-queue components, so it can never intercept a pluggable worker (e.g. a future `bulk-reaper`/`log-history-reaper` now falls through to Celery — verified by routing test). The exact reaper aliases are still matched above, and pg-prefixed typos (`pg-reapr`, `pg-queue-reapr`) are still rejected loudly. This also removes the need to mirror a reserved-suffix guard into `discover_pluggable_workers`, since the only reserved namespace is now the `pg-` prefix.

workers/run-worker-docker.sh
```diff
@@ -486,6 +491,62 @@ run_worker() {
...
+ensure_pg_interpreter() {
+    if [[ ! -x "$PG_QUEUE_PYTHON_BIN" ]]; then
+        print_status $RED "PG-queue interpreter not found/executable: $PG_QUEUE_PYTHON_BIN — check the image build / venv path."
+        exit 1
+    fi
+}
+
+run_pg_consumer() {
+    ensure_pg_interpreter
+    local source_type="${WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE:-notification}"
```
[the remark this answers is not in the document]
muhammad-ali-e: Keeping `notification` here, with rationale: it's not arbitrary — it deliberately MIRRORS the module's own default. `pg_queue_consumer/__main__.py` does `os.environ.get("WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE", "notification")`, so when the var is unset the module actually registers the **notification** worker's tasks. The shell label `pg-consumer-notification` therefore accurately reflects what the consumer is really running as. Switching the label to `unknown` would make it diverge from the module's actual registered role — the exact label/reality mismatch the toolkit review flagged. The unset-WARN added in c7bfe1b07 already makes the misconfigured state obvious in the logs. (Aligning the literal across shell+module is the real fix, but that's an upstream change, not this docker-wiring PR.)

muhammad-ali-e:
### Greptile round addressed — `d8ac1e601`

Greptile passed (no Critical/High). Both P2s:

- **`*-reaper` glob could catch a pluggable worker** — **fixed**: narrowed the near-miss catch to `pg-*` only (reserved prefix), so no pluggable name is ever intercepted; pg-prefixed typos still rejected. Verified by routing test.
- **fallback `notification` → `unknown`** — **kept `notification`** with rationale: it mirrors the module's own default (`pg_queue_consumer/__main__.py` defaults `WORKER_TYPE` to `notification` when the env is unset), so the label accurately reflects what the consumer actually registers as; `unknown` would create a label/reality divergence. The unset-WARN already surfaces the misconfig.

PR #2081, opened 2026-06-18
UN-3596 [FEAT] PG Queue 9h-b — PG scheduler tick in the reaper/orchestrator loop (per-schedule ownership)

muhammad-ali-e:
## What & why

The periodic-trigger half of the **Celery Beat replacement**. The leader-elected reaper becomes "the orchestrator" — besides barrier-orphan recovery it now fires due, **PG-owned** schedules onto the PG queue (the labs single-orchestrator model: recover + schedule in one loop). No Beat/RabbitMQ in the trigger path. **Dark by default; never double-fires.**

(The downstream execution is already PG-capable via the transport seam — this migrates only the *trigger*.)

## How it works

- **`pg_scheduler.py`** (new) — `dispatch_due_schedules()` scans `pg_periodic_schedule` for `pg_owned AND enabled AND (next_run_at IS NULL OR <= now())`, and per row:
  - `next_run_at IS NULL` → record a **baseline** next time and **don't fire** (a freshly handed-over schedule fires at its next cron match, not immediately — matches Beat, avoids a burst on hand-over).
  - due → enqueue `scheduler.tasks.execute_pipeline_task` on the PG `scheduler` queue **and** advance `next_run_at` **in one transaction** (a crash between them can't re-fire).
  - a bad cron on one row is logged and skipped, never blocking the others. `croniter` computes next-run; all time comparisons use the DB clock.
- **`pg_owned` flag** (migration 0009, default `False`) is the per-schedule switch. The scheduler fires **only owned rows**; default-false keeps the table inert until a schedule is handed over, and a schedule fires from **exactly one side** (when `pg_owned`, its Beat `PeriodicTask` is disabled by the ramp control — next slice) → **never double-fires**, fully reversible.
- **reaper** runs the scheduler **after** recovery (a scheduler error can't starve the recovery net).
- **`pg-scheduler` consumer role** (run-worker.sh + docker-compose, profile-gated) runs the fired task → existing `execute_pipeline_task_v2` → creates the execution + dispatches the workflow exactly as today.
- `croniter` added to workers deps (already a backend dep).

## Non-regression

`pg_owned` defaults `False`, so the reaper fires nothing until rows are explicitly owned — recovery-only behaviour is unchanged, Beat keeps firing everything. No new env/flag (the data flag is the switch).

## Dev-test

- **Live** against the dev DB: drove the reaper's leader `tick()` → it fired a seeded `pg_owned` due schedule (1 `scheduler` message enqueued, `last_run` stamped, `next_run` advanced).
- 10 scheduler tests (real-PG: due / baseline-no-fire / not-owned / disabled / not-yet-due / bad-cron) + 3 reaper-wiring tests (leader runs it, standby doesn't, runs after recovery). Full reaper suite (71 tests) green.
- Migration applied + `--check` clean; compose validates (0 PG services by default, 6 under `--profile pg-queue`); ruff/pre-commit clean.

## Out of scope (→ ②c)

The ramp control that flips `pg_owned` by percentage + disables the matching Beat `PeriodicTask` atomically (reusing the existing Flipt mechanism), the one-time backfill of pre-existing schedules, and retiring Beat.

Targets `feat/UN-3445-pg-queue-integration` (not `main`). Sub-task UN-3596.


muhammad-ali-e (commented):
## Automated PR Review — PG Scheduler Tick (UN-3596)

Reviewed with the PR Review Toolkit (Code Reviewer, Silent-Failure Hunter, Type-Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier). Overall this is a careful, well-documented change — the inlined INSERT matches the canonical `PgQueueClient.send` shape, the atomic enqueue+advance is correct, leader-gating and post-recovery ordering are right, and the dark-by-default `pg_owned` rollout is sound.

The findings below are concentrated in **error isolation** (the per-row loop), one **correctness invariant that is not yet enforced** (no-double-fire depends on an unbuilt ramp control), and some **type/test** hardening. Severity is tagged per inline comment.

**Priority:** 🔴 the per-row `try/except` (else one bad schedule drops every other due fire and poisons the connection) → 🟠 bad-cron wedge + the unenforced no-double-fire invariant → 🟡 type contracts, SELECT rollback, test gaps, SQL dedup → 🔵 comment-rot nits.

Inline comments follow.

workers/queue_backend/pg_queue/pg_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+            )
+            continue
+
+        payload = _build_trigger_payload(
+            workflow_id=workflow_id,
+            organization_id=organization_id,
+            pipeline_id=pipeline_id,
+            pipeline_name=pipeline_name,
+        )
+        with conn.cursor() as cur:
```
muhammad-ali-e: 🔴 **Critical — per-row fire has no `try/except`; one bad row aborts the whole batch and poisons the connection.**

The fire path (INSERT + UPDATE + `conn.commit()`, lines 147–167) runs with no exception handling. If any statement or the commit fails (serialization failure, check-constraint violation, transient socket drop, oversized payload), the exception propagates out of `dispatch_due_schedules` immediately:

- **All remaining due rows in `due` are dropped this cycle** — rows after the failing one never fire (they re-fire next tick, but the batch is non-isolated).
- **The connection is left in an aborted-transaction state** (`InFailedSqlTransaction` on every subsequent statement) — there is no `conn.rollback()` anywhere in this function. The owned-conn path is rescued by `_discard_owned_sweep_conn()`, but an injected conn (the documented caller-supplied path, and how the real-PG tests drive it) is handed back poisoned.

The docstring (lines 85–87) already promises per-row isolation, but that only covers the cron-parse step, not the DB step. The sibling `recover_expired_barriers` does exactly the right thing: per-row `try/except` → `conn.rollback()` → `logger.exception(...)` → `continue`.

**Fix:** wrap the per-row DB work (both the fire branch here and the baseline `UPDATE` at lines 124–139) in `try/except` that rolls back and `continue`s:
```python
try:
    with conn.cursor() as cur:
        cur.execute("INSERT INTO pg_queue_message ...", (...))
        cur.execute("UPDATE pg_periodic_schedule SET last_run_at=%s, next_run_at=%s WHERE pipeline_id=%s", (...))
    conn.commit()
except Exception:
        conn.rollback()
    logger.exception("PG scheduler: failed to fire pipeline %s — leaving for next tick", pipeline_id)
    continue
fired += 1
```

workers/queue_backend/pg_queue/pg_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+        pipeline_id,
+        organization_id,
+        workflow_id,
+        pipeline_name,
+        cron_string,
+        next_run_at,
+    ) in due:
+        try:
+            nxt = compute_next_run(cron_string, base)
+        except Exception:
```
muhammad-ali-e, line 163: 🟠 **High — a permanently-invalid cron wedges the row and emits a full traceback every tick, forever.**

When `compute_next_run` raises (croniter raises at construction for empty/whitespace/malformed crons, and the model allows `cron_string = ""`), this branch logs and `continue`s **before** touching `next_run_at`. Because the row is selected on `next_run_at IS NULL OR next_run_at <= now()` and its `next_run_at` is never advanced, the same row is re-selected on **every** tick. At the default ~5s reaper interval that is a `logger.exception` (full traceback to Sentry) every ~5s per bad row, indefinitely — burying real errors and re-parsing the bad cron each cycle. The schedule also silently never fires, with no durable surfacing to an operator.

**Fix:** make the bad-cron state self-quiescing and actionable — disable the row (or push `next_run_at` far out) so it stops being re-selected, and log once at the transition:
```python
except Exception:
    logger.exception("PG scheduler: invalid cron %r for pipeline %s — disabling row", cron_string, pipeline_id)
        with conn.cursor() as cur:
            cur.execute("UPDATE pg_periodic_schedule SET enabled = FALSE WHERE pipeline_id = %s", (pipeline_id,))
        conn.commit()
    continue
```
If auto-disable is undesirable by product policy, at minimum dedupe the logging (log at error only on first observation) — re-erroring every 5s is not acceptable.

workers/queue_backend/pg_queue/pg_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+def dispatch_due_schedules(conn: PgConnection) -> int:
+    """Fire PG-owned, enabled, due schedules; return the count actually fired.
+
+    The caller (reaper tick) gates this on leadership. All time comparisons use
+    the DB clock (``now()``), matching the leader-election's clock-safety posture.
+    Each row is handled in its own transaction; a baseline-only row just records
+    its ``next_run_at``, a due row enqueues + advances atomically. A bad cron on
+    one row is logged and skipped without blocking the others.
+    """
+    with conn.cursor() as cur:
```
muhammad-ali-e: 🟡 **Medium — the read step (SELECT now() + due scan) has no rollback on error, deviating from the sibling pattern.**

The directly-analogous `recover_expired_barriers` wraps its read in `try/except: conn.rollback(); raise` precisely so the connection isn't left in an aborted-txn state. Here, if the SELECT fails mid-read the exception propagates with the transaction still open. As with the fire path, the owned-conn case is rescued by `_discard_owned_sweep_conn()` but the injected-conn case is left poisoned.

**Fix — mirror `recover_expired_barriers`:**
```python
try:
    with conn.cursor() as cur:
        cur.execute("SELECT now()")
        base = cur.fetchone()[0]
        cur.execute("SELECT ... WHERE ...", (base,))
        due = cur.fetchall()
    conn.commit()
except Exception:
        conn.rollback()
    raise
```

workers/queue_backend/pg_queue/pg_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+Folded into the leader-elected reaper loop (the reaper becomes "the
+orchestrator": recover + schedule, per the labs single-orchestrator model).
+Each cycle, *only while leader*, it scans ``pg_periodic_schedule`` for rows it
+owns (``pg_owned``) that are enabled and due, enqueues the existing
+``scheduler.tasks.execute_pipeline_task`` onto the PG queue, and advances
+``next_run_at`` — replacing what Celery Beat does, without Beat/RabbitMQ.
+
+Two correctness properties:
+
+- **Never double-fires.** A row fires from exactly one side: ``pg_owned`` rows
```
muhammad-ali-e: 🟠 **High — the "Never double-fires" property is stated as already-true but currently has no enforcement.**

The guarantee depends on the Beat `PeriodicTask` being disabled for `pg_owned` rows — done "by the ramp control, next slice", a component that does not exist in the repo yet (confirmed: nothing sets `pg_owned=True` and no ramp control exists). Until it lands, a row manually flipped to `pg_owned=True` while its Beat `PeriodicTask` is still enabled **will** double-fire. The only thing actually preventing double-fires today is `pg_owned` defaulting to `False` (nothing is owned).

This is a cross-table invariant, so a `CheckConstraint` can't express it. **Suggestions:** (1) soften this docstring + the `models.py` comment to state the guarantee is *conditional on the ramp control disabling Beat* and that, pre-ramp, safety rests on `pg_owned=False`; (2) before/with the ramp control, add a reconciliation check or integration test asserting no `pg_owned=True` row has a correspondingly-enabled `PeriodicTask`.

workers/queue_backend/pg_queue/pg_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+    return croniter(cron_string, base).get_next(datetime)
+
+
+def _build_trigger_payload(
+    *,
+    workflow_id: object,
+    organization_id: str,
+    pipeline_id: object,
+    pipeline_name: str,
+) -> dict:
```
muhammad-ali-e: 🟡 **Medium — type contract is weaker than what the code already has in hand.**

- `-> dict` discards a type that exists: this returns `to_payload(...)`'s `TaskPayload` (a TypedDict). The rest of the package returns typed rows everywhere (`PgQueueClient.read -> list[QueueMessage]`, `PgReaper.tick -> TickOutcome`); this is the one place that drops it. Annotate `-> TaskPayload`.
- `workflow_id: object` / `pipeline_id: object` (lines 53, 55) say "literally anything," the opposite of the real contract. At runtime these are `uuid.UUID` (psycopg2) or `str` (tests). Use `str | uuid.UUID` (and `str | uuid.UUID | None` for `workflow_id`).
- Minor asymmetry: `workflow_id` is truthiness-guarded (line 66) but `pipeline_id` is unconditionally `str(pipeline_id)` (line 70) → would render `"None"` if ever NULL. Safe today (PK), but consider asserting non-None to make the asymmetry intentional.

workers/queue_backend/pg_queue/pg_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+            WHERE pg_owned AND enabled
+              AND (next_run_at IS NULL OR next_run_at <= %s)
+            """,
+            (base,),
+        )
+        due = cur.fetchall()
+    conn.commit()  # release the read snapshot; per-row work commits on its own
+
+    fired = 0
+    for (
```
muhammad-ali-e: 🟡 **Medium — positional 6-tuple unpack is the most fragile spot in the file.**

The SELECT column list (lines 94–95) and this 6-tuple unpack are ~15 lines apart, coupled only by ordering. A future reorder of the SELECT silently misassigns every variable with no error (same-arity tuples unpack fine) — e.g. swapping two `TextField` columns routes `organization_id` into `pipeline_name` undetected. The package already has the fix pattern twice (`QueueMessage` dataclass, `TickOutcome` NamedTuple).

**Fix — bind names to columns at one site with a `NamedTuple`:**
```python
class _DueSchedule(NamedTuple):
    pipeline_id: uuid.UUID
    organization_id: str
    workflow_id: uuid.UUID | None
    pipeline_name: str
    cron_string: str
    next_run_at: datetime | None
...
due = [_DueSchedule(*r) for r in cur.fetchall()]
for row in due:  # row.next_run_at, row.cron_string, ...
```
This also makes the `next_run_at IS NULL` baseline branch self-documenting.

workers/queue_backend/pg_queue/pg_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+            workflow_id=workflow_id,
+            organization_id=organization_id,
+            pipeline_id=pipeline_id,
+            pipeline_name=pipeline_name,
+        )
+        with conn.cursor() as cur:
+            # Enqueue + advance in ONE transaction so a crash between them can't
+            # re-fire next cycle. The INSERT mirrors PgQueueClient.send (the
+            # canonical shape) — inlined here only to share this transaction.
+            cur.execute(
```
muhammad-ali-e: 🟡 **Medium (maintainability) — the queue-row INSERT is now duplicated verbatim with `PgQueueClient.send`.**

Reusing `send()` directly is correctly *not* feasible (it commits inside `self._cursor()`, which would break the single-transaction enqueue+advance), so inlining is justified. But this INSERT string is now byte-for-byte identical to `client.py` (only difference: the trailing `RETURNING msg_id`). A future schema change (new column/default) must be edited in two places and can silently drift — and `client.py`'s docstring explicitly calls that INSERT "the whole queue contract … in one place."

**Fix:** extract a module-level SQL constant in `client.py` and import it here:
```python
# client.py
_INSERT_MESSAGE_SQL = (
    "INSERT INTO pg_queue_message "
    "(queue_name, message, org_id, priority, enqueued_at, vt, read_ct) "
    "VALUES (%s, %s::jsonb, %s, %s, now(), now(), 0)"
)
```
`send()` uses `_INSERT_MESSAGE_SQL + " RETURNING msg_id"`; the scheduler executes the constant inside its own transaction. (If keeping the inline copy is preferred, add a cross-reference comment so the two are updated together.)

workers/queue_backend/pg_queue/pg_scheduler.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def _build_trigger_payload(
+    *,
+    workflow_id: object,
+    organization_id: str,
+    pipeline_id: object,
+    pipeline_name: str,
+) -> dict:
+    """``execute_pipeline_task`` payload. Positional args match its signature:
```
muhammad-ali-e, line 82: 🔵 **Low (comment accuracy) — "the middle two are unused legacy args" is true for `execute_pipeline_task_v2`, but can mislead vs the Beat path.**

`SchedulerHelper._schedule_task_job` actually passes a real `execution_action or ""` (not blank) in that position; `v2` simply ignores it. Worth clarifying so a reader comparing the two paths isn't confused: e.g. "…`execution_action`/`execution_id` are ignored by `execute_pipeline_task_v2`, so we send blanks even though the Beat path populates `execution_action`."

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -488,6 +489,14 @@ def tick(self) -> TickOutcome:
         except Exception:
             self._discard_owned_sweep_conn()
             raise
+        # Orchestrator's second job: fire due PG-owned schedules (Beat
+        # replacement). Runs after recovery so a scheduler error can't block the
```
muhammad-ali-e: 🔵 **Low (comment accuracy) — "a scheduler error can't block the recovery net" risks being misread.**

The `except` below still does `raise`, so a scheduler error *does* propagate out of the tick (and discards the conn) — recovery isn't blocked only because it already ran *before* this point. A future reader might read "can't block" as "is swallowed." Suggest rewording to emphasise ordering, not isolation: "Runs after recovery so this cycle's recovery has already completed before any scheduler error can propagate."

backend/pg_queue/models.py
```diff
@@ -257,7 +259,14 @@ class PgPeriodicSchedule(models.Model):
     cron_string = models.TextField()
     # Mirrors PeriodicTask.enabled (pipeline.active / pause / resume).
     enabled = models.BooleanField(default=True)
-    # Owned by the scheduler tick (next slice); NULL until then.
+    # Per-schedule rollout switch. The PG scheduler fires a row ONLY when this is
+    # True; when a schedule is pg_owned its Celery Beat PeriodicTask is disabled
+    # (done by the ramp control — next slice) so it fires from exactly one side,
```
muhammad-ali-e: 🔵 **Low (comment-rot) — "done by the ramp control — next slice".**

Two rot risks: (1) it states the Beat-disable as already-arranged, but that component doesn't exist yet (see the no-double-fire comment on `pg_scheduler.py`); (2) "next slice" is a transient delivery-plan artifact with no anchor once merged to `main`. Replace "next slice" with a stable reference (the ramp-control ticket, e.g. UN-xxxx, or "a later change") and make the dependency explicit so the absence of enforcement reads as intentional.

workers/tests/test_pg_scheduler.py
```diff
@@ -0,0 +1,175 @@
...
+
+class TestDispatchDueSchedules:
+    def test_due_owned_row_fires_and_advances(self, clean):
+        conn = clean
+        past = datetime.datetime(2020, 1, 1, 0, 0, 0, tzinfo=datetime.timezone.utc)
+        pid = _seed(conn, pg_owned=True, enabled=True, next_run_at=past)
+
+        fired = dispatch_due_schedules(conn)
+
+        assert fired >= 1
```
muhammad-ali-e: 🟡 **Medium (test rigor) — assertions weak enough to pass under the exact regressions this module guards against.**

- `assert fired >= 1` (line 116): with a single seeded row this also passes if the row fired *twice* — the very double-fire bug the module claims to prevent. Tighten to `== 1`.
- `assert next_run > past` (line 124): only checks monotonicity, not correctness — passes even if the next run were computed wrong (e.g. +1 minute instead of next 09:00). Assert equality against the cron-computed value.
- `test_null_next_run_baselines_without_firing` should also `assert dispatch_due_schedules(conn) == 0` to pin "baseline counts as not-fired."

workers/tests/test_pg_scheduler.py
```diff
@@ -0,0 +1,175 @@
...
+    """Remove any rows this test created (the tick commits, so teardown must)."""
+    yield pg_conn
+    with pg_conn.cursor() as cur:
+        cur.execute("DELETE FROM pg_periodic_schedule WHERE organization_id = %s", (_MARKER,))
+        cur.execute("DELETE FROM pg_queue_message WHERE org_id = %s", (_MARKER,))
+    pg_conn.commit()
+
+
+class TestDispatchDueSchedules:
+    def test_due_owned_row_fires_and_advances(self, clean):
```
muhammad-ali-e, line 119: 🟡 **Medium (test gaps) — the two headline correctness properties and the most-likely prod bug are untested.**

1. **Atomicity / no-re-fire (highest value):** every test hits the happy commit path. Add a real-PG test that forces the advance `UPDATE` to fail after the INSERT, expects the exception, then asserts `_queued_messages(conn) == []` **and** `next_run_at` unchanged — proving the INSERT rolled back with the UPDATE (defends "a crash between them can't re-fire").
2. **Timezone:** `compute_next_run`'s unit tests use **naive** datetimes, but production `base` is tz-aware (`SELECT now()`). Add a pure test with an aware base asserting the result is aware+correct, and in `test_due_owned_row_fires_and_advances` assert the persisted `next_run_at` equals the cron-computed value. This is the single most likely place for a real prod bug.
3. **Multi-row:** no test that two due, valid, owned rows both fire (`fired == 2`, two distinct messages). This is the normal production case and the per-row-transaction design specifically supports it.

workers/tests/test_pg_reaper.py
```diff
@@ -31,6 +31,18 @@
     recover_expired_barriers,
 )
 
+# The reaper's leader tick also runs the PG scheduler tick (②b). Its behaviour
+# is covered in test_pg_scheduler.py; stub it here by default so the leadership /
+# recovery / connection tests aren't coupled to a real schedule query on their
+# dummy or barrier-only connections. Tests that assert the wiring opt in via the
+# returned mock.
+@pytest.fixture(autouse=True)
+def stub_scheduler_tick(monkeypatch):
```
muhammad-ali-e, line 40: 🟡 **Medium (test gap) — the autouse stub is reasonable, but it hides the scheduler error-isolation wiring.**

Stubbing `dispatch_due_schedules` module-wide is the right call for the leadership/recovery/connection tests. But nothing covers the try/except at `reaper.py:495-499` for the *scheduler* call: `test_failed_sweep_discards_owned_conn` only exercises the `recover_expired_barriers` path. If someone deleted the try/except around the scheduler call (or dropped the `_discard_owned_sweep_conn()`), no test would fail.

**Fix (in `TestSchedulerTick`):** set `stub_scheduler_tick.side_effect = psycopg2.OperationalError(...)` with `recover_expired_barriers` patched to return `[]`, assert `reaper.tick()` raises and the owned conn was discarded.

workers/pyproject.toml
```diff
@@ -37,6 +37,7 @@ dependencies = [
     # PG Queue — direct Postgres access for the bespoke SKIP LOCKED queue
     # (reuses the backend's DB_* connection: PgBouncer in cloud, direct in OSS)
     "psycopg2-binary==2.9.9", # matches backend; first direct-DB worker capability
+    "croniter>=3.0.3", # cron next-run computation for the PG scheduler tick (matches backend)
```
muhammad-ali-e, line 40: 🔵 **Info — croniter lock drift.** Both backend and workers pin `croniter>=3.0.3`, but the workers `uv.lock` resolved to `6.2.2` while the backend `uv.lock` is at `6.0.0`. Same constraint and major version, so behavior is consistent — flagging only so a future lock refresh aligns them.

workers/queue_backend/pg_queue/pg_scheduler.py
muhammad-ali-e: Fixed in 4c25ae514 — per-row fire (INSERT+UPDATE+commit) and the baseline UPDATE are now wrapped in try/except → `conn.rollback()` + `logger.exception` + `continue`, exactly like `recover_expired_barriers`. A row-level failure no longer poisons the (injected or owned) connection or drops the rest of the batch. +atomicity test forcing the advance UPDATE to fail post-INSERT and asserting the enqueue rolled back.

workers/queue_backend/pg_queue/pg_scheduler.py
muhammad-ali-e: Fixed in 4c25ae514 — `_quiesce_invalid_cron` now sets `enabled = FALSE` on the bad row (+ logs once at the transition), so it stops being re-selected and stops re-emitting a traceback every tick. +test asserting the bad-cron row ends up disabled while the good row still fires.

workers/queue_backend/pg_queue/pg_scheduler.py
muhammad-ali-e: Fixed in 4c25ae514 — the read step (SELECT now() + due scan) now wraps `try/except: conn.rollback(); raise`, mirroring `recover_expired_barriers`, so the connection is never handed back in an aborted-txn state.

workers/queue_backend/pg_queue/pg_scheduler.py
muhammad-ali-e: Fixed in 4c25ae514 — softened both the docstring and the `models.py` comment: the no-double-fire guarantee is now stated as CONDITIONAL on the ②c ramp control disabling the matching Beat PeriodicTask, and that until it lands safety rests on `pg_owned` defaulting to False (a manual flip while the PeriodicTask is still enabled would double-fire). An integration check that no pg_owned row has an enabled PeriodicTask will land with the ramp control.

workers/queue_backend/pg_queue/pg_scheduler.py
muhammad-ali-e: Fixed in 4c25ae514 — `_build_trigger_payload -> TaskPayload`; `workflow_id: str | uuid.UUID | None`, `pipeline_id: str | uuid.UUID`. (pipeline_id is the PK so never None; left it unconditional `str()` since the NamedTuple types it non-optional.)

workers/queue_backend/pg_queue/pg_scheduler.py
muhammad-ali-e: Fixed in 4c25ae514 — added a `_DueSchedule(NamedTuple)` and `due = [_DueSchedule(*row) for row in cur.fetchall()]`, so the SELECT columns bind to names at one site; the loop uses `schedule.cron_string`/`schedule.next_run_at` etc. A future column reorder can't silently misassign.

workers/queue_backend/pg_queue/pg_scheduler.py
muhammad-ali-e: Fixed in 4c25ae514 — extracted `INSERT_MESSAGE_SQL` as a module constant in client.py; `send()` uses `INSERT_MESSAGE_SQL + ' RETURNING msg_id'` and the scheduler executes the same constant inside its own transaction. No more verbatim duplication; a schema change is edited once.

workers/queue_backend/pg_queue/pg_scheduler.py
muhammad-ali-e: Fixed in 4c25ae514 — reworded: '…`execution_action` / `execution_id` are ignored by `execute_pipeline_task_v2`, so we send blanks even though the Beat path populates `execution_action`.'

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed in 4c25ae514 — reworded to emphasise ordering, not isolation: 'Ordered AFTER recovery so this cycle's recovery has already completed before any scheduler error can propagate (the except below still re-raises + discards the conn).'

backend/pg_queue/models.py
muhammad-ali-e: Fixed in 4c25ae514 — dropped 'next slice' and the already-arranged phrasing; the comment now states the dependency explicitly (the ②c ramp control must disable Beat atomically) and that pre-ramp safety rests on the False default.

workers/tests/test_pg_scheduler.py
muhammad-ali-e: Fixed in 4c25ae514 — `fired == 1` (catches a double-fire on a single seeded row); `next_run` asserted at the cron's 09:00 match (hour/minute/second), not just monotonic; the baseline test now asserts `dispatch_due_schedules(conn) == 0`.

workers/tests/test_pg_scheduler.py
muhammad-ali-e: Fixed in 4c25ae514 — added: (1) atomicity test — a conn wrapper fails the advance UPDATE after the INSERT; asserts no message persisted AND next_run_at unchanged (INSERT rolled back with the UPDATE); (2) tz-aware `compute_next_run` test (aware base → aware+correct result); (3) multi-row test (two due owned rows → fired == 2, two distinct messages).

workers/tests/test_pg_reaper.py
muhammad-ali-e: Fixed in 4c25ae514 — added `test_scheduler_error_discards_owned_conn` in TestSchedulerTick: `stub_scheduler_tick.side_effect = psycopg2.OperationalError(...)` with recovery patched → asserts `tick()` raises AND the owned sweep conn was discarded (set to None).

workers/pyproject.toml
muhammad-ali-e: Acknowledged — lock drift noted (workers resolved croniter 6.2.2 vs backend 6.0.0; same `>=3.0.3` constraint + major, so behaviour is consistent). Left as-is for this PR; flagging for a future aligned lock refresh.

muhammad-ali-e:
## Review round 1 addressed — `4c25ae514`

Thanks for the 6-agent pass. All 14 addressed (13 fixed, 1 info-ack):

**Critical / High — error isolation + double-fire honesty**
- **Per-row fire had no try/except** → now rollback + log + continue per row (both fire + baseline UPDATE), mirroring `recover_expired_barriers`; no more connection-poisoning / batch-drop. +atomicity test.
- **Invalid cron wedged the row + tracebacked every tick** → bad cron now disables the row (`enabled=FALSE`) + logs once. +test.
- **Read step had no rollback** → `try/except: rollback; raise`.
- **"Never double-fires" overstated** → softened to CONDITIONAL on the ②c ramp control; pre-ramp safety rests on `pg_owned=False` default (docstring + models.py).

**Medium — types / structure / tests**
- `_build_trigger_payload -> TaskPayload`; `workflow_id`/`pipeline_id` typed `str | uuid.UUID`.
- `_DueSchedule` NamedTuple binds SELECT columns at one site (no silent misassign on reorder).
- `INSERT_MESSAGE_SQL` extracted in client.py; `send()` + scheduler share it.
- Tests: `fired == 1` (catches double-fire), `next_run` asserted at the cron match, baseline `== 0`, +tz-aware, +multi-row (`fired == 2`), +atomicity (advance fails → enqueue rolls back), +reaper scheduler-error-discards-conn.

**Low / Info**
- Comment fixes: reaper ordering-not-isolation, blanks-vs-Beat-`execution_action`, models.py drop "next slice".
- croniter lock drift (6.2.2 vs backend 6.0.0): acknowledged, left for a future aligned refresh.

75 worker tests green; ruff/ruff-format/pre-commit clean. (Also dropped an accidentally-committed `9f-design.md` in `83be4493c`.)

workers/queue_backend/pg_queue/pg_scheduler.py
```diff
@@ -0,0 +1,225 @@
...
+        schedule.cron_string,
+        schedule.pipeline_id,
+    )
+    with contextlib.suppress(Exception):
+        with conn.cursor() as cur:
+            cur.execute(
+                "UPDATE pg_periodic_schedule SET enabled = FALSE WHERE pipeline_id = %s",
+                (schedule.pipeline_id,),
+            )
+        conn.commit()
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 11258f1a9 — `_quiesce_invalid_cron` now wraps the disable UPDATE in `try/except` with `conn.rollback()` on failure (your suggested shape), so a failed `enabled = FALSE` can't leave the connection in an aborted-transaction state that poisons the next row's INSERT. +test `test_quiesce_failure_does_not_poison_next_row`: a forced disable-UPDATE failure on a bad-cron row doesn't stop a following healthy row from firing.

muhammad-ali-e:
### Greptile round addressed — `11258f1a9`

Greptile passed (no Critical/High). The one **P1** — *dirty connection after a suppressed disable-UPDATE failure in `_quiesce_invalid_cron`* — **fixed**: replaced `contextlib.suppress` around the cursor block with `try/except` → `conn.rollback()`, so a failed bad-cron disable can't poison the next row's INSERT (it was the same poison-the-next-row class the round-1 per-row isolation fixed for the fire path; this was the one remaining suppressed path). +regression test proving a following healthy row still fires.

14 scheduler tests green; ruff/pre-commit clean.

PR #2085, opened 2026-06-19
UN-3597 [FEAT] PG Queue 9h-c — schedule ownership ramp control + backfill (Beat→PG hand-over)

muhammad-ali-e:
## What & why

The ramp control that completes the scheduler track: it hands a pipeline's schedule from **Celery Beat → the PG scheduler**, per-schedule and reversibly, so the Beat trigger can be retired. **Inert by default** (gate off / Flipt 0% → every schedule stays on Beat), so deploying it changes nothing until ops ramps.

## The invariant (why it never double-fires)

A schedule is owned by exactly one firer. `reconcile_ownership_for` applies the decision **atomically in one transaction**:

```
pg_periodic_schedule.pg_owned = resolve_schedule_owner(...)   # PG fires it
PeriodicTask.enabled          = active AND NOT pg_owned       # Beat fires it
```

So a `pg_owned` schedule **always has its Beat `PeriodicTask` disabled** — the two can never both fire. (The ②b "never double-fires" property was *conditional* on this slice; this makes it real.)

## Changes

- **`backend/scheduler/ownership.py`** (new) — `resolve_schedule_owner` (master-gate `PG_QUEUE_TRANSPORT_ENABLED` → `pg_scheduler_enabled` Flipt flag, keyed on `pipeline_id` for a stable %-bucket, **fail-closed to Beat** on closed gate / blind Flipt / any error) + `reconcile_ownership_for` (the atomic flip; best-effort — never breaks the caller). A **separate** Flipt flag from `pg_queue_execution_enabled` so scheduling and execution ramp independently. **No new env.**
- **`backend/scheduler/helper.py`** — reconcile ownership after the mirror upsert on every schedule create/update (a no-op while the rollout is off).
- **`backend/pg_queue/management/commands/reconcile_pg_schedules.py`** (new) — backfills mirror rows for schedules created before the mirror existed + reconciles all ownership against the current rollout. Run once + after each Flipt ramp. Idempotent; `--dry-run`.
- **tests** — 10 ownership (fail-closed Flipt matrix + the disable-Beat / pause-preserved invariants) + the helper-wiring assertion (21 backend tests green).

## Rollout / rollback

- **Percentage rollout**: ramp `pg_scheduler_enabled` → run `reconcile_pg_schedules` → that % of schedules flip to PG (Beat disabled for each, atomically).
- **Per-schedule / global rollback**: turn the flag down → reconcile → ownership flips back to Beat and the `PeriodicTask` is re-enabled.
- **Pause preserved**: a paused pipeline (`active=False`) stays disabled on both sides regardless of ownership.

## Dev-test (live, real DB)

- Hand-over → `pg_owned=True` + Beat `PeriodicTask` disabled; rollback → back to Beat; paused → off both sides.
- `reconcile_pg_schedules` backfilled real unmirrored Beat schedules + reconciled.
- **Cross-slice link**: a reconciled-to-PG (`pg_owned`) schedule is fired by the ②b reaper (correct `execute_pipeline_task` payload on the `scheduler` queue).
- 21 unit tests; ruff / pre-commit clean.

**Full container e2e** (pg-scheduler consumer → real execution via the live `pg_scheduler_enabled` flag) is folded into the **ops ramp** — the live flag + a real cron-scheduled pilot pipeline are ramp-time fixtures, alongside the execution canary's e2e.

## Out of scope

Stopping the `celery-beat` container / removing `django_celery_beat` — an ops step once ramped to 100% (and the final decommission).

Targets `feat/UN-3445-pg-queue-integration` (not `main`). Sub-task UN-3597.

### DIFF
--- a/backend/pg_queue/management/commands/reconcile_pg_schedules.py
+++ b/backend/pg_queue/management/commands/reconcile_pg_schedules.py
@@ -1,0 +1,104 @@
+"""Backfill the pg_periodic_schedule mirror + reconcile Beat/PG schedule
+ownership (Phase 9, ②c).
+
+Run this:
+- **once** after deploying the mirror, to backfill rows for schedules created
+  before the mirror existed (the dual-write only covers schedules touched since);
+- **after each Flipt ramp change** to ``pg_scheduler_enabled``, to apply the new
+  percentage — flipping ``pg_owned`` and the matching Beat ``PeriodicTask`` for
+  every schedule (the create/update path only reconciles the schedule it edits).
+
+It is idempotent and safe to run anytime: with the rollout off it leaves every
+schedule on Beat. Could later be driven periodically (e.g. by the orchestrator);
+kept a command here so the ramp stays an explicit, auditable ops action.
+"""
+
+import json
+from typing import Any
+
+from django.core.management.base import BaseCommand
+from django_celery_beat.models import CrontabSchedule, PeriodicTask
+from scheduler.ownership import reconcile_ownership_for
+from scheduler.tasks import mirror_periodic_schedule_upsert
+
+from pg_queue.models import PgPeriodicSchedule
+
+# Only the pipeline-trigger PeriodicTasks are scheduled pipelines (other periodic
+# tasks — metrics, audit — are not mirrored).
+_PIPELINE_TASK_PATH = "scheduler.tasks.execute_pipeline_task"
+
+
+def _cron_from_crontab(crontab: CrontabSchedule | None) -> str:
+    """Reconstruct the 5-field cron string from a CrontabSchedule row."""
+    if crontab is None:
+        return ""
+    return (
+        f"{crontab.minute} {crontab.hour} {crontab.day_of_month} "
+        f"{crontab.month_of_year} {crontab.day_of_week}"
+    )
+
+
+class Command(BaseCommand):
+    help = (
+        "Backfill pg_periodic_schedule mirrors for pre-existing schedules and "
+        "reconcile Beat/PG ownership against the current pg_scheduler_enabled "
+        "rollout. Idempotent; with the rollout off, leaves everything on Beat."
+    )
+
+    def add_arguments(self, parser: Any) -> None:
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help="Report what would change without writing.",
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        dry_run = options["dry_run"]
+
+        # 1. Backfill: every pipeline-trigger PeriodicTask should have a mirror.
+        backfilled = 0
+        periodic_tasks = PeriodicTask.objects.filter(task=_PIPELINE_TASK_PATH)
+        for pt in periodic_tasks:
+            pipeline_id = pt.name  # = str(pipeline.pk)
+            if PgPeriodicSchedule.objects.filter(pipeline_id=pipeline_id).exists():
+                continue
+            task_args = json.loads(pt.args or "[]")
+            workflow_id = task_args[0] if len(task_args) > 0 else None
+            organization_id = task_args[1] if len(task_args) > 1 else ""
+            # args[6] is the synthetic "Pipeline job-<id>" label; the real name
+            # self-heals via the dual-write on the next schedule edit.
+            pipeline_name = task_args[6] if len(task_args) > 6 else ""
+            self.stdout.write(
+                f"backfill mirror for pipeline {pipeline_id} " f"(enabled={pt.enabled})"
+            )
+            if not dry_run:
+                mirror_periodic_schedule_upsert(
+                    pipeline_id=pipeline_id,
+                    organization_id=organization_id or "",
+                    workflow_id=workflow_id,
+                    pipeline_name=pipeline_name,
+                    cron_string=_cron_from_crontab(pt.crontab),
+                    enabled=pt.enabled,
+                )
+            backfilled += 1
+
+        # 2. Reconcile ownership for every mirror row against the current rollout.
+        reconciled = pg_owned_count = 0
+        for row in PgPeriodicSchedule.objects.all():
+            reconciled += 1
+            if dry_run:
+                continue
+            # mirror.enabled tracks pipeline.active (dual-write); use it as the
+            # 'active' input so a paused schedule isn't re-enabled by reconcile.
+            if reconcile_ownership_for(
+                str(row.pipeline_id), row.organization_id, row.enabled
+            ):
+                pg_owned_count += 1
+
+        prefix = "[dry-run] " if dry_run else ""
+        self.stdout.write(
+            self.style.SUCCESS(
+                f"{prefix}backfilled={backfilled} reconciled={reconciled} "
+                f"pg_owned={pg_owned_count}"
+            )
+        )
--- a/backend/pg_queue/management/commands/reconcile_pg_schedules.py
+++ b/backend/pg_queue/management/commands/reconcile_pg_schedules.py
@@ -68,9 +68,9 @@ class Command(BaseCommand):
             raise CommandError(f"{failed} schedule(s) failed to reconcile")
         self.stdout.write(self.style.SUCCESS(summary))
 
-    def _parse_task_args(self, pt: Any, pipeline_id: str) -> list | None:
-        """PeriodicTask.args as a list, or None (logged) for a malformed/non-array
-        row — a bad row must not abort the whole command.
+    def _mirror_fields_from_args(self, pt: Any, pipeline_id: str) -> dict | None:
+        """Extract the mirror fields from PeriodicTask.args, or None (logged) for a
+        malformed/non-array row — a bad row must not abort the whole command.
         """
         try:
             # json.JSONDecodeError is a ValueError subclass, so one except covers
@@ -78,7 +78,6 @@ class Command(BaseCommand):
             task_args = json.loads(pt.args or "[]")
             if not isinstance(task_args, list):
                 raise ValueError(f"expected JSON array, got {type(task_args).__name__}")
-            return task_args
         except ValueError as exc:
             self.stderr.write(
                 self.style.ERROR(
@@ -86,6 +85,13 @@ class Command(BaseCommand):
                 )
             )
             return None
+        return {
+            "workflow_id": task_args[0] if len(task_args) > 0 else None,
+            "organization_id": (task_args[1] if len(task_args) > 1 else "") or "",
+            # args[6] is the synthetic "Pipeline job-<id>" label; the real name
+            # self-heals via the dual-write on the next schedule edit.
+            "pipeline_name": task_args[6] if len(task_args) > 6 else "",
+        }
 
     def _backfill_mirrors(self, dry_run: bool) -> int:
         """Create a mirror row for every pipeline-trigger PeriodicTask lacking one."""
@@ -94,8 +100,8 @@ class Command(BaseCommand):
             pipeline_id = pt.name  # = str(pipeline.pk)
             if PgPeriodicSchedule.objects.filter(pipeline_id=pipeline_id).exists():
                 continue
-            task_args = self._parse_task_args(pt, pipeline_id)
-            if task_args is None:
+            fields = self._mirror_fields_from_args(pt, pipeline_id)
+            if fields is None:
                 continue
             self.stdout.write(
                 f"backfill mirror for pipeline {pipeline_id} (enabled={pt.enabled})"
@@ -103,13 +109,9 @@ class Command(BaseCommand):
             if not dry_run:
                 mirror_periodic_schedule_upsert(
                     pipeline_id=pipeline_id,
-                    # args[6] is the synthetic "Pipeline job-<id>" label; the real
-                    # name self-heals via the dual-write on the next schedule edit.
-                    workflow_id=task_args[0] if len(task_args) > 0 else None,
-                    organization_id=(task_args[1] if len(task_args) > 1 else "") or "",
-                    pipeline_name=task_args[6] if len(task_args) > 6 else "",
                     cron_string=_cron_from_crontab(pt.crontab),
                     enabled=pt.enabled,
+                    **fields,
                 )
             backfilled += 1
         return backfilled
--- a/backend/pg_queue/management/commands/reconcile_pg_schedules.py
+++ b/backend/pg_queue/management/commands/reconcile_pg_schedules.py
@@ -54,58 +54,78 @@ class Command(BaseCommand):
 
     def handle(self, *args: Any, **options: Any) -> None:
         dry_run = options["dry_run"]
+        backfilled = self._backfill_mirrors(dry_run)
+        reconciled, pg_owned, failed = self._reconcile_all(dry_run)
 
-        # 1. Backfill: every pipeline-trigger PeriodicTask should have a mirror.
+        prefix = "[dry-run] " if dry_run else ""
+        summary = (
+            f"{prefix}backfilled={backfilled} reconciled={reconciled} "
+            f"pg_owned={pg_owned} failed={failed}"
+        )
+        if failed:
+            # Surface failures where the operator looks (and to automation).
+            self.stderr.write(self.style.ERROR(summary))
+            raise CommandError(f"{failed} schedule(s) failed to reconcile")
+        self.stdout.write(self.style.SUCCESS(summary))
+
+    def _parse_task_args(self, pt: Any, pipeline_id: str) -> list | None:
+        """PeriodicTask.args as a list, or None (logged) for a malformed/non-array
+        row — a bad row must not abort the whole command.
+        """
+        try:
+            # json.JSONDecodeError is a ValueError subclass, so one except covers
+            # both the parse error and the non-array guard below.
+            task_args = json.loads(pt.args or "[]")
+            if not isinstance(task_args, list):
+                raise ValueError(f"expected JSON array, got {type(task_args).__name__}")
+            return task_args
+        except ValueError as exc:
+            self.stderr.write(
+                self.style.ERROR(
+                    f"skipping pipeline {pipeline_id}: bad PeriodicTask.args ({exc})"
+                )
+            )
+            return None
+
+    def _backfill_mirrors(self, dry_run: bool) -> int:
+        """Create a mirror row for every pipeline-trigger PeriodicTask lacking one."""
         backfilled = 0
-        periodic_tasks = PeriodicTask.objects.filter(task=_PIPELINE_TASK_PATH)
-        for pt in periodic_tasks:
+        for pt in PeriodicTask.objects.filter(task=_PIPELINE_TASK_PATH):
             pipeline_id = pt.name  # = str(pipeline.pk)
             if PgPeriodicSchedule.objects.filter(pipeline_id=pipeline_id).exists():
                 continue
-            # PeriodicTask.args is free-form text; a malformed / non-array row must
-            # not abort the whole command (step 2 — the reconcile — is the point).
-            try:
-                task_args = json.loads(pt.args or "[]")
-                if not isinstance(task_args, list):
-                    raise ValueError(
-                        f"expected JSON array, got {type(task_args).__name__}"
-                    )
-            except (json.JSONDecodeError, ValueError) as exc:
-                self.stderr.write(
-                    self.style.ERROR(
-                        f"skipping pipeline {pipeline_id}: bad PeriodicTask.args "
-                        f"({exc})"
-                    )
-                )
+            task_args = self._parse_task_args(pt, pipeline_id)
+            if task_args is None:
                 continue
-            workflow_id = task_args[0] if len(task_args) > 0 else None
-            organization_id = task_args[1] if len(task_args) > 1 else ""
-            # args[6] is the synthetic "Pipeline job-<id>" label; the real name
-            # self-heals via the dual-write on the next schedule edit.
-            pipeline_name = task_args[6] if len(task_args) > 6 else ""
             self.stdout.write(
                 f"backfill mirror for pipeline {pipeline_id} (enabled={pt.enabled})"
             )
             if not dry_run:
                 mirror_periodic_schedule_upsert(
                     pipeline_id=pipeline_id,
-                    organization_id=organization_id or "",
-                    workflow_id=workflow_id,
-                    pipeline_name=pipeline_name,
+                    # args[6] is the synthetic "Pipeline job-<id>" label; the real
+                    # name self-heals via the dual-write on the next schedule edit.
+                    workflow_id=task_args[0] if len(task_args) > 0 else None,
+                    organization_id=(task_args[1] if len(task_args) > 1 else "") or "",
+                    pipeline_name=task_args[6] if len(task_args) > 6 else "",
                     cron_string=_cron_from_crontab(pt.crontab),
                     enabled=pt.enabled,
                 )
             backfilled += 1
+        return backfilled
 
-        # 2. Reconcile ownership for every mirror row against the current rollout.
-        reconciled = pg_owned_count = failed = 0
+    def _reconcile_all(self, dry_run: bool) -> tuple[int, int, int]:
+        """Reconcile ownership for every mirror row against the current rollout.
+        Returns (reconciled, pg_owned, failed).
+        """
+        reconciled = pg_owned = failed = 0
         for row in PgPeriodicSchedule.objects.all():
             if dry_run:
                 # Preview only — read the would-be owner (no DB write) so an
                 # operator can see how many a ramp change would hand to PG.
                 reconciled += 1
                 if resolve_schedule_owner(str(row.pipeline_id), row.organization_id):
-                    pg_owned_count += 1
+                    pg_owned += 1
                 continue
             # mirror.enabled tracks pipeline.active (dual-write); use it as the
             # 'active' input so a paused schedule isn't re-enabled by reconcile.
@@ -117,15 +137,5 @@ class Command(BaseCommand):
                 continue
             reconciled += 1
             if result:
-                pg_owned_count += 1
-
-        prefix = "[dry-run] " if dry_run else ""
-        summary = (
-            f"{prefix}backfilled={backfilled} reconciled={reconciled} "
-            f"pg_owned={pg_owned_count} failed={failed}"
-        )
-        if failed:
-            # Surface failures where the operator looks (and to automation).
-            self.stderr.write(self.style.ERROR(summary))
-            raise CommandError(f"{failed} schedule(s) failed to reconcile")
-        self.stdout.write(self.style.SUCCESS(summary))
+                pg_owned += 1
+        return reconciled, pg_owned, failed
--- a/backend/pg_queue/management/commands/reconcile_pg_schedules.py
+++ b/backend/pg_queue/management/commands/reconcile_pg_schedules.py
@@ -95,10 +95,15 @@ class Command(BaseCommand):
 
     def _backfill_mirrors(self, dry_run: bool) -> int:
         """Create a mirror row for every pipeline-trigger PeriodicTask lacking one."""
+        # Pre-fetch the already-mirrored ids in one query (avoid an EXISTS per row).
+        mirrored = {
+            str(pk)
+            for pk in PgPeriodicSchedule.objects.values_list("pipeline_id", flat=True)
+        }
         backfilled = 0
         for pt in PeriodicTask.objects.filter(task=_PIPELINE_TASK_PATH):
             pipeline_id = pt.name  # = str(pipeline.pk)
-            if PgPeriodicSchedule.objects.filter(pipeline_id=pipeline_id).exists():
+            if pipeline_id in mirrored:
                 continue
             fields = self._mirror_fields_from_args(pt, pipeline_id)
             if fields is None:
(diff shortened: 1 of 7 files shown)


muhammad-ali-e (commented):
Automated PR review (PR Review Toolkit: Code Reviewer, Silent-Failure Hunter, Type-Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier).

Solid, well-documented slice. The atomic mirror+PeriodicTask update is correct and fail-closed semantics match the existing transport gate. Headline issues, in priority order:

1. **CRITICAL** — unguarded `json.loads` crashes the whole backfill on one bad row.
2. **HIGH** — resume path (`enable_task`) re-enables Beat for a `pg_owned` schedule → the double-fire this slice exists to prevent.
3. **HIGH** — management command's `SUCCESS` summary counts swallowed reconcile failures as success.
4. **MEDIUM** — rollback doesn't reset `next_run_at` → burst on re-ramp; stale docstring symbol name; boolean-trap `active` arg; command + atomicity are untested.

Inline comments below. Severity tags: [CRITICAL]/[HIGH]/[MEDIUM]/[LOW].

backend/pg_queue/management/commands/reconcile_pg_schedules.py
```diff
@@ -0,0 +1,104 @@
...
+        dry_run = options["dry_run"]
+
+        # 1. Backfill: every pipeline-trigger PeriodicTask should have a mirror.
+        backfilled = 0
+        periodic_tasks = PeriodicTask.objects.filter(task=_PIPELINE_TASK_PATH)
+        for pt in periodic_tasks:
+            pipeline_id = pt.name  # = str(pipeline.pk)
+            if PgPeriodicSchedule.objects.filter(pipeline_id=pipeline_id).exists():
+                continue
+            task_args = json.loads(pt.args or "[]")
```
muhammad-ali-e, line 78: **[CRITICAL] Unguarded `json.loads` aborts the entire ramp on one malformed row.**

`pt.args` is a free-form text column. A single row with non-JSON/truncated args (`"{"`), or a non-array (`"null"` → `None`, `"{}"` → dict) raises `JSONDecodeError`/`TypeError` here — and since the loop has no per-row guard, `handle()` aborts. Every later PeriodicTask is skipped **and step 2 (the reconcile loop, the actual point of the command) never runs**, breaking the "idempotent, safe to run anytime" promise. The `len(...) > N` guards on lines 66-70 assume a list and don't defend against dict/None.

Wrap the parse per-row, validate it's a list, log+skip the bad row, and `continue`:
```python
try:
    task_args = json.loads(pt.args or "[]")
    if not isinstance(task_args, list):
        raise ValueError(f"expected JSON array, got {type(task_args).__name__}")
except (json.JSONDecodeError, ValueError) as exc:
    self.stderr.write(self.style.ERROR(
        f"skipping pipeline {pipeline_id}: bad PeriodicTask.args ({exc})"))
    continue
```

backend/scheduler/helper.py
```diff
@@ -74,6 +75,13 @@ def _schedule_task_job(pipeline: Pipeline, job_data: Any) -> None:
             cron_string=cron_string,
             enabled=pipeline.active,
         )
+        # Ramp control (Phase 9, ②c): align this schedule's firer (Beat vs PG)
+        # with the rollout. Inert until the gate + pg_scheduler_enabled flag are
+        # on (fails closed to Beat), so this is a no-op during normal operation;
+        # when a schedule is owned by PG it disables the Beat PeriodicTask in the
+        # same transaction so the two never both fire. Best-effort (never raises).
+        # organization_id is the org identifier string (what the mirror stores).
+        reconcile_ownership_for(str(pipeline.pk), organization_id, pipeline.active)
```
muhammad-ali-e: **[HIGH] Resume path re-enables Beat on a PG-owned schedule → double-fire.**

`reconcile_ownership_for` is only wired in here (`_schedule_task_job`). Pause/resume take a different path: `resume_job` → `enable_task` (`scheduler/tasks.py:251`) unconditionally does `task.enabled = True; task.save()` with no `pg_owned` check. So if a schedule is `pg_owned=True` (Beat disabled by reconcile) and the pipeline is then resumed, Beat's `PeriodicTask` is re-enabled while `pg_owned` stays `True` — both firers active until the next reconcile. This is exactly the double-fire the module docstring (`ownership.py:10`) claims is now prevented.

Fix: have `enable_task` set `enabled = active and not pg_owned` (or call `reconcile_ownership_for` after re-enabling). At minimum, document the resume window as a known remaining gap if it's deferred to a later slice.

backend/pg_queue/management/commands/reconcile_pg_schedules.py
```diff
@@ -0,0 +1,104 @@
...
+                    pipeline_name=pipeline_name,
+                    cron_string=_cron_from_crontab(pt.crontab),
+                    enabled=pt.enabled,
+                )
+            backfilled += 1
+
+        # 2. Reconcile ownership for every mirror row against the current rollout.
+        reconciled = pg_owned_count = 0
+        for row in PgPeriodicSchedule.objects.all():
+            reconciled += 1
```
muhammad-ali-e: **[HIGH] Reconcile loop counts swallowed failures as success; summary can't report failure.**

`reconcile_ownership_for` is best-effort — on a DB error it logs and returns `pg_owned` **without committing** (`ownership.py:119`). But `reconciled += 1` runs before the call, and `pg_owned_count` (line 96) keys off the *resolved intent*, not whether the transaction persisted. So a row whose transaction rolled back still counts as reconciled/pg_owned. The `self.style.SUCCESS(...)` summary (lines 99-103) has no `failed` counter and no non-zero exit — for an "explicit, auditable ops action," the one place an operator looks is structurally incapable of reporting failure (failures only land in the app log via `logger.exception`).

Have `reconcile_ownership_for` signal commit-vs-fail distinctly, tally `failed`, include it in the summary, switch style to `ERROR` when `failed > 0`, and `raise CommandError` so automation notices.

backend/scheduler/ownership.py
```diff
@@ -0,0 +1,125 @@
+"""Schedule-ownership ramp control (Phase 9, ②c) — hands a pipeline's schedule
+from Celery Beat to the Postgres scheduler, per-schedule and reversibly.
+
+A schedule is owned by exactly one firer. ``reconcile_schedule_ownership``
```
muhammad-ali-e: **[MEDIUM] Stale/incorrect function name in the module docstring.**

The docstring references ``reconcile_schedule_ownership``, but no such symbol exists — the actual public function is `reconcile_ownership_for` (line 80; the only name imported in `helper.py` and `reconcile_pg_schedules.py`). A future maintainer doing a symbol search will find nothing. Replace with `reconcile_ownership_for`.

backend/scheduler/ownership.py
```diff
@@ -0,0 +1,125 @@
...
+            "resolve_schedule_owner: Flipt check failed for pipeline %s; "
+            "leaving on Beat",
+            pipeline_id,
+        )
+        return False
+    return bool(owned)
+
+
+def reconcile_ownership_for(
+    pipeline_id: str, organization_id: str | None, active: bool
```
muhammad-ali-e: **[MEDIUM] Boolean-trap positional `active` arg controlling a fire/don't-fire invariant.**

Both call sites pass `active` positionally with *different-named* source fields — `helper.py:84` passes `pipeline.active`, `reconcile_pg_schedules.py:94` passes `row.enabled` — into a slot that drives `PeriodicTask.enabled = active and not pg_owned` (line 111). The trailing bool is opaque at the call site and one reorder away from re-enabling Beat on a paused pipeline.

Make it keyword-only — one-line change, zero runtime cost, and it matches the established convention of the neighboring `mirror_periodic_schedule_upsert`:
```python
def reconcile_ownership_for(pipeline_id, organization_id, *, active: bool) -> bool:
```

backend/scheduler/ownership.py
```diff
@@ -0,0 +1,125 @@
...
+    failure is logged and swallowed so it can never break the caller; the
+    schedule stays on its current firer until the next reconcile. Returns the
+    resolved ``pg_owned``.
+    """
+    pg_owned = resolve_schedule_owner(pipeline_id, organization_id)
+    try:
+        with transaction.atomic():
+            # The mirror row exists from the dual-write (②a) / backfill; guard
+            # anyway — a missing row means nothing to own yet.
+            updated = PgPeriodicSchedule.objects.filter(pipeline_id=pipeline_id).update(
```
muhammad-ali-e, line 115: **[MEDIUM] Rollback (PG→Beat) doesn't reset `next_run_at` → burst on re-ramp.**

The PG tick treats `next_run_at IS NULL` as "freshly handed over → baseline, don't fire" (`workers/queue_backend/pg_queue/pg_scheduler.py:168`, and the docstring's "No burst on hand-over" guarantee). When reconcile flips `pg_owned` True→False (rollback to Beat), it only updates `pg_owned` and never clears `next_run_at`. On a later re-ramp (False→True) the row still carries the stale non-NULL `next_run_at`; if it's `<= now()` the tick fires immediately (dequeue cond at `pg_scheduler.py:148`) instead of baselining. Reversibility is only burst-safe in one direction.

When `pg_owned` resolves `False`, also set `next_run_at=None` in the same `.update(...)` so any later re-hand-over re-enters the NULL-baseline path.

backend/pg_queue/management/commands/reconcile_pg_schedules.py
```diff
@@ -0,0 +1,104 @@
...
+    )
+
+    def add_arguments(self, parser: Any) -> None:
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help="Report what would change without writing.",
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
```
muhammad-ali-e, line 55: **[MEDIUM] The entire management command is untested.**

No file references `reconcile_pg_schedules`, `_cron_from_crontab`, or `call_command` for it. Untested branches: backfill idempotency skip (line 63), the args-index extraction (65-70) incl. the crash in the CRITICAL comment, dry-run write-suppression (lines 74 & 89), and the `reconciled`/`pg_owned` counter accuracy (88-96) that operators rely on during a ramp.

Add a test module using `call_command("reconcile_pg_schedules")` with mocked `PeriodicTask` / `PgPeriodicSchedule` / `mirror_periodic_schedule_upsert` / `reconcile_ownership_for`, asserting: existing-mirror rows are skipped; `--dry-run` calls neither upsert nor reconcile but still tallies; short/empty/malformed args; and summary counts.

backend/scheduler/tests/test_pg_schedule_ownership.py
```diff
@@ -0,0 +1,134 @@
...
+class TestReconcileOwnership:
+    def _patches(self, *, owner: bool, rows_matched: int = 1):
+        sched = patch("scheduler.ownership.PgPeriodicSchedule")
+        pt = patch("scheduler.ownership.PeriodicTask")
+        resolve = patch(
+            "scheduler.ownership.resolve_schedule_owner", return_value=owner
+        )
+        # transaction.atomic() as a no-op context manager.
+        txn = patch(
+            "scheduler.ownership.transaction.atomic",
```
muhammad-ali-e, line 79: **[MEDIUM] Mocked `transaction.atomic` can't prove the "no double-fire" invariant.**

Patching `transaction.atomic` to `nullcontext()` plus mocking the ORM means these tests verify only that two `.update()` calls happen — not that they're in one transaction, and not that a failure in the `PeriodicTask` update rolls back the `pg_owned` write. `test_failure_is_swallowed` would still pass if someone deleted `transaction.atomic()` entirely, yet the module docstring calls atomicity "the load-bearing property."

Add one `@pytest.mark.django_db` test: real mirror row + real `PeriodicTask`, monkeypatch the *second* update to raise, call `reconcile_ownership_for`, and assert `pg_owned` was rolled back. That's the only test that actually guards the invariant the PR is built around.

backend/scheduler/ownership.py
```diff
@@ -0,0 +1,125 @@
...
+from unstract.flags.feature_flag import check_feature_flag_status
+
+logger = logging.getLogger(__name__)
+
+# Independent of the execution-transport flag (pg_queue_execution_enabled) so
+# scheduling and execution ramp separately. %-rollout keyed on pipeline_id.
+SCHEDULER_FLAG_KEY = "pg_scheduler_enabled"
+
+
+def resolve_schedule_owner(pipeline_id: str, organization_id: str | None) -> bool:
```
muhammad-ali-e, line 40: **[LOW] Bare `bool` for a BEAT-vs-PG decision the codebase already models as an enum.**

This module's docstring says it "Mirrors `resolve_transport`" — but `resolve_transport` returns the shared `WorkflowTransport` enum (`unstract/core/.../data_models.py`), not a bool. A bool here is a polarity trap: `True` means PG, while the mirrored `resolve_transport` defaults to Celery/Beat — one negation away from a silent double-fire. Consider returning `WorkflowTransport` (or a small `ScheduleOwner{BEAT,PG}` enum) and deriving `pg_owned` at the DB-write boundary. Also: `pipeline_id` is typed `str` but every caller passes `str(pipeline.pk)`/`str(row.pipeline_id)` and the body re-coerces — widen to `str | UUID` to match `resolve_transport` and drop the call-site `str(...)` noise. Low priority / judgment call; flagging for consistency with the sibling it claims to mirror.

backend/scheduler/ownership.py
```diff
@@ -0,0 +1,125 @@
...
+
+    # Flipt context is a gRPC map<string,string>; coerce values to str (a non-str
+    # makes the client swallow it as False). entity_id str-coerced too so the
+    # %-rollout bucket is stable across str/UUID call sites.
+    context = {"pipeline_id": str(pipeline_id)}
+    if organization_id:
+        context["organization_id"] = str(organization_id)
+    try:
+        owned = check_feature_flag_status(
+            SCHEDULER_FLAG_KEY, entity_id=str(pipeline_id), context=context
```
muhammad-ali-e: **[LOW] Pass `flag_key` as a keyword for consistency with `resolve_transport`.**

`resolve_transport` calls `check_feature_flag_status(flag_key=..., entity_id=..., context=...)`; here `SCHEDULER_FLAG_KEY` is positional. Functionally identical, but since the docstring markets this as mirroring `resolve_transport`, `flag_key=SCHEDULER_FLAG_KEY` reads more consistently.

backend/scheduler/ownership.py
```diff
@@ -0,0 +1,125 @@
...
+        )
+    except Exception:
+        logger.exception(
```
muhammad-ali-e: **[LOW] Flipt-failure path uses `logger.exception` on every schedule edit — log/Sentry noise.**

This is the one swallow that runs on *every* schedule edit. If Flipt is persistently down, it emits a full-traceback `logger.exception` per edit even though the outcome is benign-by-design (fail-closed to Beat). Consider `logger.warning(..., exc_info=True)` — it's an expected, recoverable condition, not an exceptional bug, and downgrading keeps real errors visible.

backend/pg_queue/management/commands/reconcile_pg_schedules.py
```diff
@@ -0,0 +1,104 @@
...
+                    cron_string=_cron_from_crontab(pt.crontab),
+                    enabled=pt.enabled,
+                )
+            backfilled += 1
+
+        # 2. Reconcile ownership for every mirror row against the current rollout.
+        reconciled = pg_owned_count = 0
+        for row in PgPeriodicSchedule.objects.all():
+            reconciled += 1
+            if dry_run:
```
muhammad-ali-e, line 130: **[LOW] `--dry-run` reports nothing for the reconcile step.**

The reconcile loop `continue`s immediately under dry-run, so `pg_owned_count` is always 0 and the summary always prints `pg_owned=0`. But step 2 (applying the ramp) is the command's actual purpose — an operator can't preview how many schedules a Flipt ramp change would hand to PG. In dry-run, still call `resolve_schedule_owner(str(row.pipeline_id), row.organization_id)` (read-only, no DB write) and tally the would-be `pg_owned` count.

backend/pg_queue/management/commands/reconcile_pg_schedules.py
```diff
@@ -0,0 +1,104 @@
...
+            if PgPeriodicSchedule.objects.filter(pipeline_id=pipeline_id).exists():
+                continue
+            task_args = json.loads(pt.args or "[]")
+            workflow_id = task_args[0] if len(task_args) > 0 else None
+            organization_id = task_args[1] if len(task_args) > 1 else ""
+            # args[6] is the synthetic "Pipeline job-<id>" label; the real name
+            # self-heals via the dual-write on the next schedule edit.
+            pipeline_name = task_args[6] if len(task_args) > 6 else ""
+            self.stdout.write(
+                f"backfill mirror for pipeline {pipeline_id} " f"(enabled={pt.enabled})"
```
muhammad-ali-e: **[LOW] Redundant adjacent string-literal concatenation.**

The f-string is split into two adjacent literals for no reason (fits well under the line limit as one):
```python
f"backfill mirror for pipeline {pipeline_id} (enabled={pt.enabled})"
```

backend/pg_queue/management/commands/reconcile_pg_schedules.py
muhammad-ali-e: Fixed in c82e999ca — per-row `try/except` around `json.loads` + an `isinstance(list)` check; a malformed/non-array `PeriodicTask.args` row logs to stderr + `continue`s, so step 2 (the reconcile) always runs. +tests for bad-JSON and non-list (`null`) args.

backend/scheduler/helper.py
muhammad-ali-e: Fixed in c82e999ca — real bug, thank you. `enable_task` (the resume path) now reads the mirror's `pg_owned` and sets `PeriodicTask.enabled = NOT pg_owned`, so resuming a PG-owned schedule leaves Beat disabled (no double-fire). +2 tests (resume when pg_owned vs not).

backend/pg_queue/management/commands/reconcile_pg_schedules.py
muhammad-ali-e: Fixed in c82e999ca — `reconcile_ownership_for` now returns `None` on a swallowed DB failure (vs the resolved bool on success). The command tallies `failed`, writes an ERROR summary, and `raise CommandError` so automation notices. +test asserting CommandError on a failed row.

backend/scheduler/ownership.py
muhammad-ali-e: Fixed in c82e999ca — docstring now references `reconcile_ownership_for` (the actual symbol).

backend/scheduler/ownership.py
muhammad-ali-e: Fixed in c82e999ca — `def reconcile_ownership_for(pipeline_id, organization_id, *, active)`; both call sites updated to `active=...`.

backend/scheduler/ownership.py
muhammad-ali-e: Fixed in c82e999ca — on rollback (`pg_owned` resolves False) the same `.update(...)` also sets `next_run_at=None`, so a later re-hand-over re-enters the PG tick's NULL baseline instead of firing on a stale timestamp. +test that pg_owned=False clears next_run_at, and that an active pg_owned=True row does NOT reset it (which would skip a fire).

backend/pg_queue/management/commands/reconcile_pg_schedules.py
muhammad-ali-e: Fixed in c82e999ca — new `pg_queue/tests/test_reconcile_pg_schedules_command.py` (call_command, mocked ORM): backfill-skips-existing, malformed-args, non-list args, `--dry-run` writes-nothing-but-previews-owner, and CommandError on failure.

backend/scheduler/tests/test_pg_schedule_ownership.py
muhammad-ali-e: Fixed in c82e999ca — added `TestReconcileAtomicityRealDB`: seeds a real mirror row + PeriodicTask, monkeypatches the *PeriodicTask* update to raise, calls `reconcile_ownership_for`, and asserts `pg_owned` rolled back (skips if no DB). That's the test that actually guards the one-transaction invariant. Ran green against the dev DB.

backend/scheduler/ownership.py
muhammad-ali-e: Deferred (LOW, judgment call): keeping `bool`. Unlike `resolve_transport` — whose `WorkflowTransport` enum is a wire contract carried across processes in the task payload — `pg_owned` is a local boolean that maps 1:1 to the `PgPeriodicSchedule.pg_owned` BooleanField; an enum would add an indirection with no cross-process contract to anchor it. The polarity is contained (one function, one DB column, unit-tested both ways). Left `pipeline_id: str` since every caller already passes a str and the body str-coerces for Flipt anyway — widening buys nothing here.

backend/scheduler/ownership.py
muhammad-ali-e: Fixed in c82e999ca — `check_feature_flag_status(flag_key=SCHEDULER_FLAG_KEY, ...)`, matching resolve_transport.

backend/scheduler/ownership.py
muhammad-ali-e: Fixed in c82e999ca — the Flipt-failure path is now `logger.warning(..., exc_info=True)` (expected, recoverable, runs every edit) rather than `logger.exception`, so a persistently-down Flipt doesn't bury real errors.

backend/pg_queue/management/commands/reconcile_pg_schedules.py
muhammad-ali-e: Fixed in c82e999ca — under `--dry-run` the reconcile loop now calls `resolve_schedule_owner` (read-only, no DB write) and tallies the would-be `pg_owned`, so an operator can preview how many a ramp change hands to PG. +test asserting resolve is called but reconcile/upsert aren't.

backend/pg_queue/management/commands/reconcile_pg_schedules.py
muhammad-ali-e: Fixed in c82e999ca — collapsed to a single f-string.

muhammad-ali-e:
## Review round 1 addressed — `c82e999ca`

Thanks — strong catches. 12 fixed, 1 deferred (LOW judgment call).

**Critical / High**
- **Resume double-fire** (real bug): `enable_task` re-enabled Beat on a `pg_owned` schedule → now `PeriodicTask.enabled = NOT pg_owned`. +2 tests.
- **Unguarded `json.loads`** aborted the whole ramp → per-row guard + `isinstance(list)` → log & skip. +tests.
- **Command counted failures as success** → `reconcile_ownership_for` returns `None` on failure; command tallies `failed`, ERROR summary, `raise CommandError`. +test.

**Medium**
- Rollback (PG→Beat) now clears `next_run_at` → no burst on re-hand-over. +tests.
- `active` is keyword-only (boolean-trap guard).
- Stale docstring symbol fixed.
- Whole command now tested (`call_command`, mocked ORM).
- **Atomicity proven**: real-DB test forces the PeriodicTask update to fail and asserts `pg_owned` rolls back — the load-bearing one-transaction invariant.

**Low**
- `flag_key` keyword; Flipt-failure log → `warning(exc_info=True)`; `--dry-run` previews would-be `pg_owned`; single f-string.
- **Deferred**: bool-vs-enum return — `pg_owned` is a local boolean mapping 1:1 to the `BooleanField`, not a cross-process wire contract like `WorkflowTransport` (rationale in thread).

29 backend tests green (incl. the real-DB atomicity test); ruff / pre-commit clean.

backend/scheduler/tasks.py
```diff
@@ -250,7 +250,18 @@ def disable_task(task_name: str) -> None:
...
+    # (the ②c ownership invariant; reconcile sets the same on create/update, but
+    # resume takes this path). Default False (not owned) when there's no mirror.
+    pg_owned = (
+        PgPeriodicSchedule.objects.filter(pipeline_id=task_name)
+        .values_list("pg_owned", flat=True)
+        .first()
+        or False
+    )
+    task.enabled = not pg_owned
     task.save()
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 5576df360 — `enable_task` now reads `pg_owned` under `select_for_update()` inside a `transaction.atomic()` and writes only the `enabled` column via `PeriodicTask.objects.filter(name=...).update(enabled=not pg_owned)` (no full `task.save()` of stale in-memory state). So a concurrent `reconcile_ownership_for` that flips `pg_owned=True`+Beat-off can't be clobbered back into a double-fire — the mirror-row lock serialises the two. Kept a leading `.get()` to preserve the DoesNotExist contract.

backend/scheduler/ownership.py
```diff
@@ -0,0 +1,139 @@
...
+            updated = PgPeriodicSchedule.objects.filter(pipeline_id=pipeline_id).update(
+                **updates
+            )
+            if updated == 0:
+                logger.info(
+                    "reconcile_ownership_for: no mirror row for pipeline %s "
+                    "(not yet mirrored); skipping",
+                    pipeline_id,
+                )
+                return pg_owned
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 5576df360 — `updated == 0` (no mirror row) now `return False`: PG can't fire without a row and the PeriodicTask was never disabled, so the effective owner is Beat. This keeps `_reconcile_all`'s `pg_owned` count honest (no inflation past what's actually live). +test asserting the False return.

backend/scheduler/ownership.py
```diff
@@ -0,0 +1,139 @@
...
+            updates: dict = {"pg_owned": pg_owned}
+            if not pg_owned:
+                # Back on Beat → clear the PG next-run so a future re-hand-over
+                # baselines instead of firing immediately on a stale timestamp.
+                updates["next_run_at"] = None
+            # The mirror row exists from the dual-write (②a) / backfill; guard
+            # anyway — a missing row means nothing to own yet.
+            updated = PgPeriodicSchedule.objects.filter(pipeline_id=pipeline_id).update(
+                **updates
+            )
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 5576df360 — the reconcile `.update(...)` now includes `updated_at=timezone.now()` (queryset .update() doesn't fire `auto_now`), mirroring `_mirror_periodic_schedule_set_enabled`, so an ownership flip advances the timestamp. Re-verified live (updated_at bumped on hand-over).

backend/pg_queue/management/commands/reconcile_pg_schedules.py
```diff
@@ -0,0 +1,141 @@
...
+        """Reconcile ownership for every mirror row against the current rollout.
+        Returns (reconciled, pg_owned, failed).
+        """
+        reconciled = pg_owned = failed = 0
+        for row in PgPeriodicSchedule.objects.all():
+            if dry_run:
+                # Preview only — read the would-be owner (no DB write) so an
+                # operator can see how many a ramp change would hand to PG.
+                reconciled += 1
+                if resolve_schedule_owner(str(row.pipeline_id), row.organization_id):
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 5576df360 — `_backfill_mirrors` now pre-fetches all mirrored ids in one query (`mirrored = {str(pk) for pk in PgPeriodicSchedule.objects.values_list('pipeline_id', flat=True)}`) and checks membership in Python, replacing the per-row EXISTS. One query instead of N.

muhammad-ali-e:
### Greptile round addressed — `5576df360`

Greptile passed (no Critical/High). 1 P1 + 3 P2, all fixed:

- **[P1] resume lost-update race** → `enable_task` reads `pg_owned` under `select_for_update()` in a transaction and writes only the `enabled` column via `.update()` (no stale `task.save()`), so a concurrent `reconcile` can't be clobbered into a double-fire.
- **[P2] inflated ramp count** → no-mirror-row reconcile returns `False` (effective owner is Beat) instead of the raw resolved value.
- **[P2] `updated_at` not bumped** → reconcile's `.update()` now sets `updated_at=timezone.now()`.
- **[P2] N+1** → `_backfill_mirrors` pre-fetches mirrored ids in one query (set membership) instead of an EXISTS per row.

29 backend tests green; hand-over + `updated_at` bump re-verified live. (Also resolved the two earlier SonarCloud complexity findings + the redundant-except in this PR.)

backend/scheduler/tests/test_pg_schedule_ownership.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+class TestReconcileAtomicityRealDB:
+    """The load-bearing invariant: the pg_owned write and the PeriodicTask write
+    are ONE transaction — if the PeriodicTask update fails, pg_owned rolls back
+    (so a schedule can't end up pg_owned with Beat still enabled). Needs a real
+    DB (the mocked atomic() can't prove rollback); skips if unreachable."""
+
+    def test_periodictask_update_failure_rolls_back_pg_owned(self):
+        import uuid
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in the latest push — added `import pytest` to the test module. The `pytest.skip(...)` in TestReconcileAtomicityRealDB will now skip cleanly when the DB is unavailable instead of raising NameError. (Local runs passed because the dev DB was reachable, so the skip branch never executed — good catch on the latent path.)

PR #2086, opened 2026-06-19
UN-3602 [FIX] PG Queue — restore API-deployment timeout sync-wait on the PG path

muhammad-ali-e:
## What

- Add nullable `WorkflowExecution.queue_message_id` (BigIntegerField) + migration `0020`.
- PG path stores the `pg_queue_message.msg_id` there; `task_id` (UUIDField) stays NULL on the PG path.
- New `WorkflowHelper._record_dispatch_handle` routes the dispatch handle by transport, wrapped in its own try/except so post-dispatch bookkeeping can't skip the synchronous wait.

## Why

- Every PG-routed API deployment ignored the `timeout` form field and returned `EXECUTING` immediately. The PG dispatch handle is a bigint `msg_id`; writing it into the UUID `task_id` raised `ValueError` on save, which the broad post-dispatch handler swallowed (returns `EXECUTING` once dispatched) — so the synchronous `timeout` poll loop was skipped entirely. (UN-3602)

## How

- Each id lives in its own correctly-typed column: Celery uuid → `task_id`; PG msg_id → `queue_message_id`. The structural part is that handle bookkeeping now runs in its own `try/except`, so a bookkeeping failure can never abort the `timeout` wait loop again.

- No. The Celery path is unchanged (still writes `task_id`). `WorkflowExecution.task_id` is informational-only — a single log reader, no `AsyncResult`/`revoke` usage, excluded from the main execution serializer, and treated as an opaque string by workers/frontend (verified across both the OSS and cloud repos). The new column is additive and nullable; nothing reads it yet.

## Database Migrations

- `0020_workflowexecution_queue_message_id` — `AddField` of a `BigIntegerField(null=True)` with no default. On PostgreSQL 11+ this is a metadata-only change (no table rewrite, no backfill, brief catalog lock), so it is safe on large tables.

## Env Config

- None.

## Notes on Testing

- 3 new unit tests pin the handle routing by transport (PG msg_id -> `queue_message_id` as int, not `task_id`; Celery uuid -> `task_id`; empty handle -> neither). Existing `_dispatch_orchestrator_task` tests are unchanged and green (7/7 total).
- Live-validated on a running stack (PG path, gate on): an API deployment with `timeout=300` now blocks until completion and returns `COMPLETED` with the result inline (`queue_message_id` populated, `task_id` NULL); without `timeout` it returns immediately (async) as before.

## Related Issues or PRs

- UN-3602 (sub-task of UN-3536). Builds on the PG-queue transport work on the feat/UN-3445-pg-queue-integration branch.

## Checklist

- [x] I have added tests that prove my fix is effective


backend/workflow_manager/workflow_v2/models/execution.py
```diff
@@ -145,6 +145,12 @@ class Type(models.TextChoices):
         null=True,
         db_comment="task id of asynchronous execution",
     )
+    queue_message_id = models.BigIntegerField(
```
muhammad-ali-e, line 148: **[Type Design · High] XOR invariant is unenforced.** `task_id` (UUID, Celery) and `queue_message_id` (bigint, PG) are two nullable columns where, post-dispatch, *exactly one* should be set. That invariant lives only in prose (db_comment + the `_record_dispatch_handle` docstring) and in the happy-path helper — nothing stops another writer from setting both or neither. This is the same "illegal state is representable" shape that produced UN-3602.

Consider expressing it at the DB level so all writers are covered:
```python
class Meta:
    constraints = [
        models.CheckConstraint(
            name="exec_task_xor_queue_msg",
            check=~(Q(task_id__isnull=False) & Q(queue_message_id__isnull=False)),
        ),
    ]
```
(Negated AND rather than strict XOR, since both are legitimately NULL pre-dispatch.) Cheapest change with the strongest guarantee.

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -517,6 +517,45 @@ def _dispatch_orchestrator_task(
...
+        if not dispatch_handle:
+            # PG always yields a truthy msg_id, so an empty handle is Celery-only.
+            logger.warning(
+                f"[{org_schema}] Empty dispatch handle (transport={transport}) "
+                f"for execution_id '{execution_id}'."
+            )
+            return
+        if is_pg_transport(transport):
+            WorkflowExecutionServiceHelper.update_execution_queue_message_id(
+                execution_id=execution_id, queue_message_id=int(dispatch_handle)
```
muhammad-ali-e: **[Silent Failure · Medium] `int(dispatch_handle)` can raise `ValueError`, now swallowed generically.** This is the same failure *class* the PR fixes (a type mismatch on the PG handle), just relocated. If `pg_enqueue_task` ever returns a non-numeric string (version skew, future handle format, error sentinel), `int()` raises `ValueError`, which the outer `except Exception` (lines 667-672) catches and logs as the generic "Failed to record dispatch handle" — giving operators no hint the *handle format* was the problem, and the PG row silently loses its `queue_message_id` (cancellation/reaper later can't find the row).

Parse defensively and log the offending value:
```python
if is_pg_transport(transport):
    try:
        msg_id = int(dispatch_handle)
    except (TypeError, ValueError):
        logger.error(
            f"[{org_schema}] PG dispatch handle {dispatch_handle!r} is not a "
            f"valid bigint msg_id for execution_id '{execution_id}'; "
            "queue_message_id not recorded"
        )
        return
    WorkflowExecutionServiceHelper.update_execution_queue_message_id(
        execution_id=execution_id, queue_message_id=msg_id
    )
```

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -517,6 +517,45 @@ def _dispatch_orchestrator_task(
         )
         return async_execution.id
 
+    @classmethod
+    def _record_dispatch_handle(
+        cls,
+        *,
+        execution_id: str,
+        transport: str,
```
muhammad-ali-e, line 524: **[Type Design · Medium] `transport: str` is stringly-typed at the internal helper boundary.** A `WorkflowTransport` (str-Enum) already exists in `unstract/core/.../data_models.py` and `is_pg_transport` is the single-source predicate. With a bare `str`, a typo or garbage value reaches here and `is_pg_transport("pg-queue")` silently returns `False`, routing a PG handle to `task_id` — re-introducing exactly the bug class this PR closes. Wire-strings can stay at the edge, but normalize once and type this internal param as `WorkflowTransport` so the discriminant is a closed set.

backend/workflow_manager/workflow_v2/execution.py
```diff
@@ -425,6 +425,32 @@ def update_execution_task(execution_id: str, task_id: str) -> None:
         except WorkflowExecution.DoesNotExist:
             logger.error(f"execution doesn't exist {execution_id}")
 
+    @staticmethod
+    def update_execution_queue_message_id(
+        execution_id: str, queue_message_id: int
```
muhammad-ali-e: **[Type Design · Medium] Annotation/contract mismatch.** The signature says `queue_message_id: int`, but the body (lines 437-442) guards `if queue_message_id is None`. The annotation lies — either it's `int | None` (matching the runtime guard) or the guard is dead code. Note the only caller, `_record_dispatch_handle`, already returns early on a falsy handle and always passes `int(dispatch_handle)`, so the `None` branch is unreachable from the real path. Either make it `int | None` for honesty (mirrors the defensive `update_execution_task`), or drop the guard. Minor: the sibling guards `if not task_id` (catches `0`/empty) while this guards `is None` only — `queue_message_id=0` would slip through, though `0` is not a valid msg_id.

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -611,23 +650,25 @@ def execute_workflow_async(
...
-                # Use existing method to handle task_id setting with validation
-                WorkflowExecutionServiceHelper.update_execution_task(
-                    execution_id=execution_id, task_id=task_id
+            # Record the dispatch handle (Celery task_id or PG msg_id) on the row.
+            # Best-effort, in its OWN try/except: it must NEVER abort the
+            # synchronous timeout wait below. (Writing a bigint msg_id into the
+            # UUID task_id used to raise ValueError here, which bubbled to the
+            # post-dispatch handler and silently skipped the wait — so every
+            # PG-routed API deployment ignored `timeout`.)
+            try:
```
muhammad-ali-e, line 671: **[Tests · High] The actual regression is untested.** This inner `try/except` is the heart of the fix, but no test exercises it. The new tests verify the *routing* of `_record_dispatch_handle` in isolation (helper mocked); they do not verify the defense-in-depth guarantee that a raise here still lets `execute_workflow_async` fall through to the `if timeout > -1:` poll loop instead of short-circuiting to the post-dispatch handler that returns `EXECUTING` immediately.

Without this test, someone could move the recording back *outside* the `try/except` (or delete it) and every existing test would still pass while UN-3602 silently returns. Suggested: patch `cls._record_dispatch_handle` to raise, patch `_dispatch_orchestrator_task` / `WorkflowExecution.objects.get` / `_get_execution_status`, and assert the poll loop runs (i.e. status is polled) rather than an immediate bare `EXECUTING`. This is the one test that would actually catch a re-regression.

backend/workflow_manager/workflow_v2/tests/test_record_dispatch_handle.py
```diff
@@ -0,0 +1,71 @@
...
+bookkeeping that records the transport's dispatch handle on the execution row.
+
+Regression context: the PG path returns ``str(msg_id)`` (a bigint string). It
+used to be written into ``WorkflowExecution.task_id`` (a UUIDField), raising
+``ValueError`` on save → the broad post-dispatch handler swallowed it and
+returned EXECUTING immediately, silently skipping the ``timeout`` sync-wait. The
+fix routes the PG msg_id into its own ``queue_message_id`` (BigIntegerField) and
+leaves ``task_id`` NULL on the PG path, so no bigint is ever forced into a UUID.
+
+DB-free: ``WorkflowExecutionServiceHelper`` is mocked.
```
muhammad-ali-e: **[Tests + Comment · Low] Two small follow-ups for this file:**

1. **Scope the docstring.** It describes the *production* crash (UUID coercion → `ValueError` → skipped wait) in detail, but these tests mock `WorkflowExecutionServiceHelper` and assert *routing only* — they never reproduce the `ValueError` or the sync-wait skip. A future maintainer could read this and assume the crash itself is covered, then delete the real guard. Add: "These tests assert routing only (PG→`queue_message_id` as int, Celery→`task_id`); the UUID-coercion crash and the timeout sync-wait are not exercised here (helper is mocked)."

2. **Add a non-numeric-handle case.** `int(dispatch_handle)` on the PG path raises `ValueError` for a malformed handle. A `test_non_numeric_pg_handle` asserting that behavior (or the graceful no-record after the fix in workflow_helper.py:547) pins the contract — today it's only protected by `_dispatch_orchestrator_task` always returning a numeric `str(msg_id)`.

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -611,23 +650,25 @@ def execute_workflow_async(
...
-                )
-                # Continue without setting task_id - execution can still complete
-            else:
-                # Use existing method to handle task_id setting with validation
-                WorkflowExecutionServiceHelper.update_execution_task(
-                    execution_id=execution_id, task_id=task_id
+            # Record the dispatch handle (Celery task_id or PG msg_id) on the row.
+            # Best-effort, in its OWN try/except: it must NEVER abort the
+            # synchronous timeout wait below. (Writing a bigint msg_id into the
+            # UUID task_id used to raise ValueError here, which bubbled to the
```
muhammad-ali-e: **[Comment · Low] Minor imprecision.** "used to raise ValueError here" reads as if the throw originated at this call site, but the `ValueError` was actually raised inside `update_execution_task` → `execution.save()` (UUID coercion). Suggest: "...used to raise `ValueError` inside `update_execution_task` (UUID coercion on save), which bubbled to the post-dispatch handler..." Helps anyone grepping for the throw origin. (The comment is otherwise an excellent "why" comment — keep it.)

muhammad-ali-e (commented):

Reviewed with six specialized agents: Code Reviewer, Silent Failure Hunter, Type Design Analyzer, PR Test Analyzer, Comment Analyzer, and Code Simplifier.

**Verdict:** Solid, well-scoped fix. The core change — routing the PG `msg_id` into its own correctly-typed `queue_message_id` column and isolating the post-dispatch bookkeeping in its own `try/except` so it can never abort the `timeout` sync-wait — is correct and well-commented. No blocking correctness bugs in the shipped code paths.

The findings below are improvements, not blockers, prioritized:

**High / should-address**
1. The actual UN-3602 regression (wait must proceed when handle-recording raises) is **not** tested — only the routing refactor is.
2. The `task_id` XOR `queue_message_id` invariant is unenforced at the model/DB level (the exact "illegal state representable" shape that caused this bug).

**Medium**
3. `int(dispatch_handle)` can raise `ValueError`, now swallowed by a generic handler with no handle-specific diagnostic.
4. `transport: str` is stringly-typed where a `WorkflowTransport` enum already exists — a typo silently misroutes to `task_id`.
5. `update_execution_queue_message_id(queue_message_id: int)` is annotated `int` but its body guards `None`.

**Low**
6. Add a non-numeric-PG-handle test; scope the test module docstring to "routing only (mocked)."
7. Comment precision: the historical `ValueError` was raised inside `update_execution_task` on `save()`, not at the call site.

Inline comments follow.

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -517,6 +517,45 @@ def _dispatch_orchestrator_task(
...
+    @classmethod
+    def _record_dispatch_handle(
+        cls,
+        *,
+        execution_id: str,
+        transport: str,
+        dispatch_handle: str | None,
+        org_schema: str,
+        file_count: int,
+    ) -> None:
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 52704e280 — changed to `@staticmethod` (it never referenced `cls`), now consistent with `_dispatch_orchestrator_task` directly above.

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -517,6 +517,45 @@ def _dispatch_orchestrator_task(
...
+            # PG always yields a truthy msg_id, so an empty handle is Celery-only.
+            logger.warning(
+                f"[{org_schema}] Empty dispatch handle (transport={transport}) "
+                f"for execution_id '{execution_id}'."
+            )
+            return
+        if is_pg_transport(transport):
+            WorkflowExecutionServiceHelper.update_execution_queue_message_id(
+                execution_id=execution_id, queue_message_id=int(dispatch_handle)
+            )
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 52704e280 — `int(dispatch_handle)` is now wrapped in `try/except (TypeError, ValueError)`; a malformed/future handle logs a specific `logger.error` (with the offending value) and records nothing, instead of being absorbed by the caller's generic post-dispatch guard.

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: Fixed in 52704e280 — defensive parse with a handle-specific `logger.error` (offending value included), so the generic outer guard no longer hides a malformed-handle cause. Same change addresses the greptile P2 on this line.

backend/workflow_manager/workflow_v2/execution.py
muhammad-ali-e: Fixed in 52704e280 — annotation is now `int | None`, matching the `is None` guard in the body and mirroring `update_execution_task`. (msg_id is a `BigAutoField` starting at 1, so the `queue_message_id=0` slip-through is moot in practice.)

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: Fixed in 52704e280 — added `tests/test_execute_workflow_async_wait.py`. It patches `_record_dispatch_handle` to raise and asserts `execute_workflow_async` still enters the poll loop (`_get_execution_status` is called) rather than short-circuiting to an immediate `EXECUTING`. Moving the recording back outside the inner `try/except` (or deleting it) fails this test — exactly the re-regression guard you described.

backend/workflow_manager/workflow_v2/tests/test_record_dispatch_handle.py
muhammad-ali-e: Both fixed in 52704e280 — added `test_non_numeric_pg_handle_records_nothing` (asserts the defensive parse records nothing and doesn't raise), and scoped the module docstring to "routing only (mocked); the UUID-coercion crash and the timeout sync-wait are covered in test_execute_workflow_async_wait.py".

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: Fixed in 52704e280 — comment now reads "...used to raise `ValueError` inside `update_execution_task` (UUID coercion on save), which bubbled to the post-dispatch handler and silently skipped the wait...".

backend/workflow_manager/workflow_v2/models/execution.py
muhammad-ali-e: Deferring this one, with reasoning. A `CheckConstraint` is added via an `AddConstraint` migration that performs a **validating full-table scan** of `workflow_execution` under a `SHARE ROW EXCLUSIVE` lock — exactly the large-table migration cost we chose the additive-nullable-column path to avoid (the `UUIDField → CharField` type-change alternative was rejected for the same reason). The invariant is already guaranteed by construction here: `_record_dispatch_handle` branches on transport and sets exactly one column, and it's the sole writer of `queue_message_id`. I'd rather not add a scanning migration on this hot table in this PR — happy to file it as a standalone constraint migration (or a `NOT VALID` + later `VALIDATE`) if you want the DB-level belt-and-suspenders.

backend/workflow_manager/workflow_v2/workflow_helper.py
muhammad-ali-e: Deferring, with reasoning. `resolve_transport` returns `str` (the `WorkflowTransport.value`), not the enum member, so annotating this internal param as `WorkflowTransport` would be inaccurate unless `resolve_transport`'s return contract changes — which ripples to every caller and the task-payload wire shape (transport is carried as a plain string), out of scope for this fix. In practice the discriminant is already closed: the only caller passes `resolve_transport`'s output, and `is_pg_transport` is the single normalization predicate. Happy to do the enum-normalization as a separate refactor of `resolve_transport` if you'd like that hardening.

muhammad-ali-e:
## Review addressed — `52704e280`

Thanks for the thorough pass. Summary of what changed:

**Fixed (7)**
- `_record_dispatch_handle` → `@staticmethod` (never used `cls`). _(greptile P2)_
- PG `msg_id` parsed defensively: `try/except (TypeError, ValueError)` with a handle-specific `logger.error`, so a malformed/future handle format surfaces its cause instead of being swallowed by the outer guard. _(greptile P2 + Silent Failure · Med)_
- `update_execution_queue_message_id` annotation `int` → `int | None` (matches the body guard; mirrors `update_execution_task`). _(Type Design · Med)_
- **Added the missing regression test** `tests/test_execute_workflow_async_wait.py`: patches `_record_dispatch_handle` to raise and asserts the `timeout` poll loop still runs (`_get_execution_status` called) instead of short-circuiting to `EXECUTING`. This is the test that catches a re-regression if the recording is moved back outside the inner `try/except`. _(Tests · High)_
- Added `test_non_numeric_pg_handle_records_nothing` and scoped the routing-test docstring to "routing only (mocked)". _(Tests + Comment · Low)_
- Comment precision: the `ValueError` was raised inside `update_execution_task` on `save()`, not at the call site. _(Comment · Low)_

**Deferred, with reasoning (2)** — replied inline:
- **XOR `CheckConstraint`** (Type Design · High): an `AddConstraint` migration validates via a full-table scan on this large hot table — the exact migration cost the additive-nullable-column design avoids (the `UUIDField → CharField` alternative was rejected for the same reason). The invariant is already guaranteed by construction (`_record_dispatch_handle` sets exactly one column and is the sole writer). Can land as a standalone (`NOT VALID` + later `VALIDATE`) migration if you want the DB-level guard.
- **`transport: WorkflowTransport`** (Type Design · Med): `resolve_transport` returns `str` (the enum *value*), not the member, and the value is carried as a plain string in the task payload — typing this internal param as the enum would require changing `resolve_transport`'s contract across all callers, out of scope here. The discriminant is already closed in practice (single caller + `is_pg_transport` predicate).

All tests green (9/9), pre-commit clean. Verdict noted — appreciated.

PR #2090, opened 2026-06-19
[FIX] Revert djangorestframework 3.15.2 → 3.14.0 to unblock staging

chandrasekharan-zipstack:
## What

Reverts **only** the `djangorestframework` bump (3.15.2 → 3.14.0) that landed in #2087. The other batched dependabot updates from that PR are untouched.

Files: `backend/pyproject.toml`, `pyproject.toml` (pins) + `backend/uv.lock`, `uv.lock` (re-resolved). Lock diff is the DRF version flip only — no other package added/removed.

## Why

DRF 3.15 auto-generates a `UniqueTogetherValidator` from model `UniqueConstraint`s, which forces **every** field in the constraint to `required=True` — including the server-set, nullable `organization` FK (`DefaultOrganizationMixin`). Clients never send `organization`, so adapter / connector / workflow / api-deployment creates started failing:

```
400 {"type":"validation_error","errors":[{"code":"required","attr":"organization"}]}
```

This broke the shared Playwright test setup (`createAdapters`) and cascaded into **146 staging failures on rc.342** (OSS v0.176.1). rc.341 (DRF 3.14, v0.176.0) was clean — 220/244. The DRF bump is the only relevant diff between them, and #2087 was a routine batched dependabot update (no CVE forcing 3.15).

## Follow-up (separate PR)

Reattempt the 3.15 upgrade deliberately: add a `get_unique_together_validators()` override to the shared serializer bases (`AuditSerializer` + `IntegrityErrorMixin`) to drop org-scoped auto-validators (restores 3.14 semantics; DB + `IntegrityErrorMixin` still enforce uniqueness), fix the cloud stragglers that inherit neither base (MRQ / subscription / agentic_table_settings), audit all ~20 org-scoped models, then cut its own RC + full suite.

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -19,7 +19,7 @@ dependencies = [
     "cron-descriptor==1.4.0", # For cron string description
     "cryptography>=48.0.1",
     "django==4.2.30",
-    "djangorestframework==3.15.2",
+    "djangorestframework==3.14.0",
     "django-cors-headers==4.3.1",
     # Pinning django-celery-beat to avoid build issues
     "django-celery-beat==2.5.0",
--- a/backend/uv.lock
+++ b/backend/uv.lock
@@ -874,14 +874,15 @@ wheels = [
 
 [[package]]
 name = "djangorestframework"
-version = "3.15.2"
+version = "3.14.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "django" },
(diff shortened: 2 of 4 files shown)


chandrasekharan-zipstack:
Superseded — reintroduces CVE-2024-21520 (DRF<3.15.2 XSS). Going with a roll-forward model-level fix on 3.15.2 instead.

PR #2092, opened 2026-06-19
[FIX] Mark server-managed organization field non-editable (DRF 3.15 compat)

chandrasekharan-zipstack:
## Problem

On DRF 3.15.2, every org-scoped create started failing with `organization: This field is required` (e.g. Add LLM adapter). Root cause: DRF 3.15 auto-derives a multi-field `UniqueTogetherValidator` from a model's `UniqueConstraint`s, and that validator requires **every** field of the constraint in the request body. `organization` is set server-side (never sent by clients) but `fields="__all__"` exposed it as a *writable* field, so it landed in the validator's required set → 400.

This is a latent modeling issue DRF 3.15 surfaced: a server-managed column was client-writable (also a cross-tenant injection risk). Downgrading DRF was rejected — 3.14.0 carries CVE-2024-21520 (XSS).

## Fix

Mark `organization` `editable=False`:
- `DefaultOrganizationMixin` (covers adapter, connector, workflow, pipeline, api, prompt-studio, tags, platform-api, usage, dashboard-metrics, tenant `OrganizationMember`, and cloud `lookups` / `manual_review_v2`).
- The five models that attach `organization` directly rather than via the mixin: `Configuration`, `OrganizationGroup`, `ResourceGroupShare`, `account_v2.PlatformKey`, `notification_v2.NotificationBuffer`.

DRF then builds `organization` read-only, which removes it from `field_sources` and drops the auto-generated validator for **every** org-scoped constraint (every such constraint includes `organization`, so one field change covers all). It also closes the latent injection — `organization` is no longer client-writable.

Verified on DRF 3.15.2 against the real serializers: before → `organization` writable, validator present, `is_valid()` fails with "required"; after → read-only, zero org validators, create payload without `organization` validates. A sweep of all 87 `ModelSerializer`s confirms no org-scoped validator remains.

## Migrations

15 `AlterField(editable=False)` migrations, generated via `makemigrations` (not hand-written). All **state-only** — `editable` is not a database attribute, so no DDL is emitted.

## Notes

- `created_by`/`modified_by` need no change: the only constraint that includes `created_by` (cloud `AgenticProject`) also includes `organization`, so it's already covered; and `created_at`/`modified_at` are `auto_now*` (already non-editable).
- Cloud counterpart (migrations for `lookups` / `manual_review_v2`, which inherit the mixin): Zipstack/unstract-cloud#1574. Merge/deploy this OSS PR first.
- Adds a DB-free regression test (`adapter_processor_v2/tests/test_org_editable.py`).


backend/api_v2/migrations/0004_alter_apideployment_organization_and_more.py
```diff
@@ -0,0 +1,38 @@
...
+                blank=True,
+                db_comment="Foreign key reference to the Organization model.",
+                editable=False,
+                null=True,
+                on_delete=django.db.models.deletion.CASCADE,
+                to="account_v2.organization",
+            ),
+        ),
+        migrations.AlterField(
+            model_name="organizationratelimit",
```
jaseemjaskp, line 27: **[HIGH] This change breaks the Django admin for `OrganizationRateLimit` (runtime 500).**

This `AlterField` makes `OrganizationRateLimit.organization` `editable=False`, but `OrganizationRateLimitAdmin` still lists `"organization"` in its `fieldsets` without adding it to `readonly_fields` (`backend/api_v2/admin.py:22`).

When the admin add/change page is opened, `ModelAdmin.get_form()` → `modelform_factory()` → `fields_for_model()` raises:
`FieldError: 'organization' cannot be specified for OrganizationRateLimit model form as it is a non-editable field`.

Important: this is a **runtime** failure on page load, **not** a `manage.py check` system-check error — the admin field-spec check (`django/contrib/admin/checks.py`) does not flag a non-editable concrete field in `fieldsets`, so CI stays green and the break only surfaces when a staff user opens the page. No test covers admin form construction either.

Fix (mirrors how `created_at`/`modified_at` are already handled in this same admin):
```python
# backend/api_v2/admin.py
readonly_fields = ["organization", "created_at", "modified_at"]
```
Keep `"organization"` in the fieldsets tuple — readonly fields are permitted there and render read-only.

jaseemjaskp (commented):
Automated PR review (PR Review Toolkit). Posting the one critical/high-confidence regression below; lower-severity notes are shared with the author separately for confirmation before posting.

johnyrahul (approved):

backend/utils/models/organization_mixin.py
```diff
@@ -11,7 +11,8 @@ class DefaultOrganizationMixin(models.Model):
         db_comment="Foreign key reference to the Organization model.",
         null=True,
         blank=True,
-        default=None,
+        # Server-managed field; set in save(), never accepted as client input.
```
jaseemjaskp, line 14: **[MEDIUM] `editable=False` now hides the only place a NULL organization could be caught — consider guarding `save()`.** (Pre-existing behavior, slightly more load-bearing after this change.)

The field is `null=True, blank=True`, and `save()` (lines 21-24) sets `self.organization = UserContext.get_organization()` only when it's `None`. `UserContext.get_organization()` returns `None` on both `Organization.DoesNotExist` and `ProgrammingError`, so a row can be persisted with `organization = NULL` and **no error raised** — an invalid state for tenant-scoped models. There is no `full_clean()` path anywhere in the backend, and `editable=False` now removes the field from DRF/admin too, so `save()` is the sole enforcement point.

Not a regression introduced by this PR (net runtime behavior is unchanged), but worth hardening while we're here:
```python
def save(self, *args, **kwargs):
    if self.organization is None:
        org = UserContext.get_organization()
        if org is None:
            raise ValueError(
                f"{type(self).__name__}.organization is unset and no organization "
                "is in context; refusing to persist an org-less tenant row."
            )
        self.organization = org
    super().save(*args, **kwargs)
```

jaseemjaskp (commented):
Follow-up: posting the medium/low findings from the PR Review Toolkit pass (the test silent-pass on line 51 is omitted — already covered by greptile's existing comment).

backend/configuration/models.py
```diff
@@ -27,7 +27,11 @@ class Configuration(BaseModel):
 
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     organization = models.ForeignKey(
-        Organization, on_delete=models.CASCADE, related_name="organization_configuration"
+        Organization,
+        on_delete=models.CASCADE,
+        related_name="organization_configuration",
+        # Server-managed; never accepted as client input.
```
jaseemjaskp, line 33: **[LOW] Comment over-promises for non-mixin models.** The mixin's comment says "set in save()" because `DefaultOrganizationMixin.save()` auto-populates `organization`. But `Configuration` (and the other directly-declared models: `OrganizationGroup`, `ResourceGroupShare`, `NotificationBuffer`, `PlatformKey`) do **not** inherit that hook — `organization` is set by explicit `organization=...` kwargs at call sites. The shared "Server-managed" wording is technically true but may lead a maintainer to assume an automatic `save()` hook protects these. Suggest differentiating, e.g. `# Server-managed; set explicitly by application code at creation, never accepted as client input.`

backend/adapter_processor_v2/tests/test_org_editable.py
```diff
@@ -0,0 +1,51 @@
...
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "backend.settings.test")
+if not apps.ready:
+    django.setup()
+
+from rest_framework.validators import UniqueTogetherValidator  # noqa: E402
+
+from adapter_processor_v2.serializers import AdapterInstanceSerializer  # noqa: E402
+
+
+def test_organization_is_read_only():
```
jaseemjaskp: **[LOW] Coverage scoped to a single serializer for a ~16-model change.** These two structural guards (`read_only` + no org-bearing `UniqueTogetherValidator`) only cover `AdapterInstanceSerializer`. The same DRF-3.15 validator regression can recur in any serializer over a mixin-using model with an org-bearing unique constraint — `APIDeployment`, `Pipeline`, `ConnectorInstance`, `CustomTool`, `Workflow`, `Tag`, `Usage`. They parametrize at near-zero cost:
```python
@pytest.mark.parametrize("serializer_cls", [AdapterInstanceSerializer, WorkflowSerializer, PipelineSerializer, ...])
def test_org_read_only_and_unvalidated(serializer_cls):
    s = serializer_cls()
    assert s.fields["organization"].read_only is True
    assert not [v for v in s.validators
                if isinstance(v, UniqueTogetherValidator) and "organization" in tuple(v.fields)]
```
Also worth a regression guard for the admin break flagged on the api_v2 migration (a project-wide `call_command("check")` test, or an admin-form-construction test).

backend/api_v2/migrations/0004_alter_apideployment_organization_and_more.py
chandrasekharan-zipstack: Fixed in c5e66c97a — added `organization` to `readonly_fields` on `OrganizationRateLimitAdmin` (mirrors the `created_at`/`modified_at` handling). Confirmed it was the only admin listing `organization` explicitly in `fieldsets` across OSS+cloud.

backend/utils/models/organization_mixin.py
chandrasekharan-zipstack: Valid, but out of scope here — net runtime behavior is unchanged by this PR and a `raise` on null org fires across every create path of ~16 models (incl. Celery/worker tasks, data migrations, fixtures that may currently write before org context is set). Deferring to a separate hardening PR after auditing which background paths set `UserContext`, to avoid introducing new breakage in a stabilization fix.

backend/configuration/models.py
chandrasekharan-zipstack: The five directly-declared models use the comment `# Server-managed; never accepted as client input.` — it does not claim a `save()` hook, so no over-promise. Only the mixin says "set in save()", which is accurate there. Leaving as-is.

backend/adapter_processor_v2/tests/test_org_editable.py
chandrasekharan-zipstack: Addressed in c5e66c97a — broadened to a field-level invariant (`organization.editable is False`) parametrized across all 12 org-scoped models via the app registry, which is the root-cause guard (if the field stays non-editable, no serializer can re-expose it). Kept the adapter serializer checks for the concrete DRF-3.15 symptom. Avoided parametrizing over serializer instances since several drag heavy connector/cloud imports that fight test isolation. 15 passed.

PR #2093, opened 2026-06-19
UN-3578 [FIX] GDrive folder browser no longer breaks on Google Docs

kirtimanmishrazipstack:
## What

- The Google Drive connector's folder browser ("Folders to process") now loads folders even when they contain Google-native files (Docs / Sheets / Slides). It previously showed "No data".

## Why

- Customers couldn't pick a Drive folder for their ETL pipeline when that folder held a Google Doc or Sheet — the folder list failed to load and blocked pipeline setup (reported on-call).

## How

- `get_files()` (`backend/file_management/file_management_helper.py`) wraps the per-file `fs.info()` size backfill in try/except. pydrive2's `info()` runs an unguarded `int(fileSize)` → `int(None)` for Google-native files (no `fileSize`), which aborted the entire listing; on failure we now log a warning and keep the `ls()` entry with size 0.

- No — config-time folder-browser path only, and now strictly more resilient. Runtime ETL execution uses a separate, already-guarded path (`walk()`/`ls()`).

## Database Migrations

- None

## Env Config

- None

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

- None

## Notes on Testing

- Manual: in a GDrive connector, browse a folder containing a Google Doc/Sheet — folders now list instead of showing "No data".

## Checklist


backend/file_management/file_management_helper.py
```diff
@@ -81,9 +81,16 @@ def get_files(fs: AbstractFileSystem, file_path: str) -> list[FileInformation]:
...
+            # Backfill missing size via fs.info(). pydrive2 raises int(None) on
+            # Google-native docs (no fileSize); fall back to 0, don't fail the listing.
             if file_info.get("type") == "file" and file_info.get("size") is None:
-                file_info = fs.info(file_name)
+                try:
+                    file_info = fs.info(file_name)
+                except Exception:
+                    logger.warning(
+                        "fs.info() failed for %s; using ls() metadata", file_name
+                    )
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: Done in c208e1c0a — bound the exception as `exc` and appended it to the warning, so the root cause (int(None) vs auth/network failure) is now surfaced. Kept the broad `except Exception` intentionally: the goal is to keep the folder listing alive on *any* backend failure, so logging the cause gives the diagnosability without narrowing the safety net.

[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Addressed — applied your suggested patch verbatim in `c208e1c0a`: `except Exception as exc:` with `logger.warning("fs.info() failed for %s; using ls() metadata: %s", file_name, exc)`. The root cause is now surfaced in the log. Good to mark resolved.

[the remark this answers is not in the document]
kirtimanmishrazipstack: @coderabbitai The "log the cause" half is done in `c208e1c0a` — the except now binds `exc` and logs it, so unexpected errors are surfaced, not hidden. Keeping the broad catch by design: this guards a non-critical `size` backfill in the config-time folder browser, and the intent is that *any* backend failure (pydrive2 `ApiRequestError`, auth/connection errors, etc.) degrades that one entry to `size=0` instead of aborting the entire listing — narrowing to `(OSError, ValueError, TypeError)` would let other backends' error types re-break it, which is the exact bug this PR fixes. BLE001 isn't enabled in this repo, so no lint concern.

backend/file_management/file_management_helper.py
```diff
@@ -81,9 +81,18 @@ def get_files(fs: AbstractFileSystem, file_path: str) -> list[FileInformation]:
...
-                file_info = fs.info(file_name)
+                try:
+                    file_info = fs.info(file_name)
+                except Exception as exc:
+                    logger.warning(
+                        "fs.info() failed for %s; using ls() metadata: %s",
+                        file_name,
+                        exc,
+                    )
+                    file_info["size"] = 0
```
chandrasekharan-zipstack, line 95: @kirtimanmishrazipstack  do check if we expect a valid file size anywhere. Will 0 B show up in the UI then? Leaving it null might be better

kirtimanmishrazipstack: @chandrasekharan-zipstack Checked — on this path `size` is display-only (serialized by `FileInfoSerializer`, rendered in the folder browser via `formatBytes`); nothing validates/sorts/gates selection on it. And `null` wouldn't change the display: `formatBytes` (`GetStaticData.js:71`) returns "0 B" for both `0` and `null` (`if (!+bytes)`). So `0` is harmless and keeps the `size: int` contract — only the stray Google-native doc shows "0 B", real files keep their size. Happy to switch to `null` if you'd rather the API say "unknown" explicitly, but the UI would still read "0 B" unless we also teach `formatBytes` to render unknown.

chandrasekharan-zipstack (approved):
I would still recommend leaving that size field blank rather than displaying 0 B. You can't distinguish between a failed file and one that is actually 0 B

Deepak-Kesavan (approved):

backend/file_management/file_management_helper.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Fair point — we weighed it and decided to keep `size = 0` rather than plumb an "unknown" value through. On this path size is purely display-only: the folder browser renders it via `formatBytes` and nothing sorts, validates, or gates folder/file selection on it. Distinguishing the two cases properly would mean carrying `null` across three layers (`get_files` → `FileInformation` dto → `FileInfoSerializer`, plus teaching `formatBytes` to render blank), which is a fair bit of surface area for a cosmetic value on a rare Google-native file that the "File types to process" filter excludes anyway.

I've documented the rationale right at the fallback in 68b504b19:

```python
# Google-native docs (Docs/Sheets/Slides) carry no fileSize, so pydrive2's
# info() raises on int(None); fall back to 0 (display-only) to keep listing.
```

Happy to do the `null`→blank treatment as a follow-up if you'd still rather it not read `0 B`.

kirtimanmishrazipstack: @chandrasekharan-zipstack Agreed, done — went with leaving it blank. The folder browser now shows an empty size cell whenever the size is `0`/unknown instead of `0 B` (6b32b5fad), so a failed/sizeless lookup no longer asserts a value we don't actually have. It's one line in the shared browser, so it applies uniformly across all connectors:

```js
size = isFile && node.size ? formatBytes(node.size) : "";
```

Kept the backend fallback as-is since `size` here is display-only (nothing sorts/validates/gates on it). One honest trade-off: a genuinely empty 0-byte file also renders blank — acceptable given the field is informational only and an empty file isn't actionable for ETL anyway.

kirtimanmishrazipstack:
> I would still recommend leaving that size field blank rather than displaying 0 B. You can't distinguish between a failed file and one that is actually 0 B

@chandrasekharan-zipstack  Agreed, fixed — size now shows blank instead of 0 B for any 0/unknown value (6b32b5fad), so we never assert a size we can't determine. One line in the shared browser, applies to all connectors

PR #2094, opened 2026-06-19
UN-3603 [GATED-FEAT] PG Queue — blocking executor RPC on Postgres + unify gating to a single flag

muhammad-ali-e:
## What

- A **parallel PG-native executor RPC**: prompt-studio's blocking `dispatch()` can ride Postgres instead of Celery, gated by the single `pg_queue_enabled` flag. The SDK `ExecutionDispatcher` and the Celery executor worker are **untouched**.
- `pg_task_result` store (migration `0010`) + `PgResultBackend` (store + poll-wait); `reply_key` on the shared `TaskPayload`; consumer result-write hook; backend gate + `PgExecutionDispatcher` + `RoutingExecutionDispatcher`; `worker-pg-executor` role.
- **Unify PG-queue gating to a single flag** `pg_queue_enabled` — execution, scheduler, and executor now read one key (was `pg_queue_execution_enabled` / `pg_scheduler_enabled` / `pg_executor_enabled`).

## Why

Migrate the executor request-reply off Celery — the last big Celery dependency (UN-3603). Cut A = prompt-studio blocking path; `structure_tool` (workflows) and `ide_*` callbacks are later slices. Consolidating the gates to one flag keeps the rollout simple to operate (one flip = whole feature), per the rollout plan (validate across envs, then a coordinated cutover).

## How

- Separate PG path selected **per-dispatch** by a routing dispatcher (gate read at call time → instant rollout/rollback, fail-closed). Reply via a keyed `pg_task_result` table + poll-wait (PgBouncer-safe; no LISTEN/NOTIFY).
- `worker-pg-executor` runs the existing `execute_extraction` (no logic fork) and the consumer writes its result/error to `pg_task_result` for the blocking caller.
- All three resolvers read `pg_queue_enabled`; `PG_QUEUE_TRANSPORT_ENABLED` (env) stays the master kill-switch.

No. Gate off (default) → `_get_dispatcher` returns the unchanged Celery `ExecutionDispatcher` for every mode and no `pg_task_result` row is created. New table is additive (metadata-only migration); `reply_key` is optional (absent = byte-identical fire-and-forget); the consumer hook is guarded by `reply_key`. The flag rename is internal (the flag is read only in the backend). Pinned by zero-regression tests + 34 passing consumer tests, and validated live: a full on→PG / off→Celery cycle for API + ETL, COMPLETED both ways.

## Database Migrations

`0010_pgtaskresult` — new droppable `pg_task_result` table (additive).

## Env Config

None new. Gate = existing `PG_QUEUE_TRANSPORT_ENABLED` + the single Flipt flag `pg_queue_enabled` (off until ramp).

## Notes on Testing

- New unit tests: routing/gate fail-closed + zero-regression (`pg_queue/tests/test_executor_rpc.py`), result-backend real-PG (`workers/tests/test_pg_result_backend.py`). 34 existing consumer tests + the flag-rename suites (transport/ownership) green.
- Live-validated on a running stack: executor RPC round-trip through `worker-pg-executor`; single-flag on→PG / off→Celery cycle for API + ETL (both COMPLETED), with `queue_message_id` set on PG and `task_id` set on Celery.

## Related Issues or PRs

UN-3603 (sub-task of UN-3536). Follow-ups: the workflow/`structure_tool` executor path, and the `ide_*` async callbacks.

## Checklist

- [x] Self-review done
- [x] Commented hard-to-understand areas
- [x] Added tests proving the change
- [x] Checked for misspellings


muhammad-ali-e:
## SonarCloud addressed — `57fe4307b`

All 3 fixed:

1. **`executor_rpc.py:120` — unused `headers` param (Major):** removed `headers` from `PgExecutionDispatcher.dispatch`. The PG path carries fairness in the enqueue payload (not Celery message headers), so `RoutingExecutionDispatcher` no longer forwards `headers` to the PG branch (the Celery branch still does).
2. **`executor_rpc.py:136` — use `logging.exception()` (Major):** the enqueue-failure handler now uses `logger.exception(...)` instead of `logger.error(..., exc_info=True)`.
3. **`models.py` — remove `null=True` (Medium, Consistency):** `PgTaskResult.error` is now `TextField(blank=True, default="")` (the no-NULL-text convention used elsewhere in `pg_queue`). `store_result` writes `""` on a completed row; migration `0010` was regenerated and the result-backend test updated.

Tests green (8 routing + 7 result-backend), pre-commit clean, `makemigrations --check` clean.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -216,6 +244,10 @@ def _handle(self, message: QueueMessage) -> None:
             )
             return
 
+        if reply_key:
+            # Persist the task's return value for the waiting caller before ack.
+            self._store_reply(reply_key, result=eager.result)
```
muhammad-ali-e: **Critical — unguarded success-path store re-runs the executor and blocks the caller.**

`_store_reply(...)` here runs *before* the ack (`delete`) at line 251 and is not wrapped in try/except. If `store_result` raises (DB write error, serialization failure, dropped connection), the exception propagates out of `_handle` and is caught only by the generic handler in `poll_once`, logged as a generic "poll failed" with no `reply_key`/`task_name`. Consequences:
- the message is never deleted → vt expires → the executor **re-runs the full extraction (real LLM spend)** up to `max_attempts`, then is dropped as poison;
- the waiting caller never gets a row and blocks to its full timeout (default 3600s);
- the failure is mislabeled, contradicting the line 219-225 promise that the caller "gets a definitive result instead of blocking to its timeout" (that only holds for the task-raised branch, not a store failure).

Also latent: `store_result` classifies any non-dict / `None` result as `STATUS_FAILED`, so a request-reply task that legitimately returns `None`/empty would be mis-stored as failed (not reachable today since `execute_extraction` always returns a dict, but fragile for future tasks).

**Fix:** wrap the store with try/except logging `reply_key`/`task_name`/`msg_id` at error severity, and decide explicitly whether a store failure should redeliver. **This branch is also entirely untested** — `test_pg_queue_consumer.py` was not updated; add ok+reply_key (store result then ack) and boom+reply_key (store error then ack) cases.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -193,20 +198,43 @@ def _handle(self, message: QueueMessage) -> None:
             self._client.delete(message.msg_id)
             return
 
+        # Request-reply (executor RPC): a unique key the dispatching caller
+        # is blocking on. Present → store the outcome to pg_task_result and ack
+        # after a single attempt (no vt-redeliver — see the except branch).
+        # Absent → fire-and-forget (the existing leaf/pipeline path).
+        reply_key = payload.get("reply_key")
```
muhammad-ali-e: **Medium — pre-execution drop branches lose the reply, blocking the caller to full timeout.**

`reply_key` is read here, but the earlier drop branches above (missing `task_name`, poison `read_ct > max_attempts`, unknown task) `delete` the row and return **without** storing any reply. For a request-reply message that hits one of those paths, the blocking caller waits the full timeout (default 3600s) for a row that will never appear. For the unknown-task and poison cases especially, a definitive failure reply could be stored so the caller fails fast.

**Fix:** in those drop branches, if the payload carries a `reply_key`, call `_store_reply(reply_key, error=...)` before deleting so the caller gets an immediate descriptive failure.

backend/pg_queue/executor_rpc.py
```diff
@@ -0,0 +1,222 @@
...
+            queue,
+            context.run_id,
+            timeout,
+        )
+        row = self._wait_for_result(reply_key, timeout)
+        if row is None:
+            return ExecutionResult.failure(
+                error=f"TimeoutError: executor reply not received within {timeout}s"
+            )
+        if row.status == "completed" and row.result is not None:
```
muhammad-ali-e: **Medium — magic status string + unguarded `from_dict` can break the never-raises contract.**

Two issues on this branch:
1. `row.status == "completed"` is a bare literal here, while the writer side names it `STATUS_COMPLETED` in `result_backend.py`. The two live in different deployables (`backend` Django ORM reader vs `workers` psycopg2 writer) with no shared import, so the contract is matched by eye across a process boundary — rename the writer constant and this silently treats every result as a failure. Consider promoting the status vocabulary to a `StrEnum` in `unstract.core` (which **both** trees already depend on).
2. `ExecutionResult.from_dict(row.result)` hard-indexes `data["success"]`. A malformed/partial `completed` row raises `KeyError`, which is **not** caught — violating this class's documented "never raises" contract and surfacing as a 500 to prompt-studio.

**Fix:** branch on the shared enum, and wrap `from_dict` in try/except converting a decode error to `ExecutionResult.failure(...)`.

backend/pg_queue/executor_rpc.py
```diff
@@ -0,0 +1,222 @@
...
+    callers can branch on ``result.success`` identically on either transport.
+    """
+
+    def dispatch(
+        self,
+        context: ExecutionContext,
+        timeout: int | None = None,
+    ) -> ExecutionResult:
+        if timeout is None:
+            timeout = int(os.environ.get(_DEFAULT_TIMEOUT_ENV, _DEFAULT_TIMEOUT))
```
muhammad-ali-e: **High (test gap) — `PgExecutionDispatcher.dispatch` is entirely untested.**

`test_executor_rpc.py` mocks `PgExecutionDispatcher` out, so none of this method's four outcome branches ever run, yet its contract ("never raises; failure/timeout → `ExecutionResult.failure`") is the load-bearing promise prompt-studio depends on.

**Add DB-free tests** (mock `enqueue_task` and `_wait_for_result`): enqueue raises → returns failure and does not re-raise; `_wait_for_result` returns None → timeout failure with the `within {timeout}s` message; completed row → `ExecutionResult.from_dict(row.result)`; failed row → failure with `row.error` (and the empty-error → `"executor task failed"` fallback); completed-but-`result is None` falls through to failure; `timeout=None` reads `EXECUTOR_RESULT_TIMEOUT` else 3600. Also cover `_wait_for_result` backoff + the `min(delay, remaining)` deadline clamp.

backend/pg_queue/executor_rpc.py
```diff
@@ -0,0 +1,222 @@
...
+        logger.info(
+            "PG executor dispatch: enqueued reply_key=%s queue=%s run_id=%s "
+            "timeout=%ss; waiting for result...",
+            reply_key,
+            queue,
+            context.run_id,
+            timeout,
+        )
+        row = self._wait_for_result(reply_key, timeout)
+        if row is None:
```
muhammad-ali-e, line 154: **Medium — timeout is silent and indistinguishable from a real failure.**

On timeout the caller gets `ExecutionResult.failure("TimeoutError: ...")`, but the executor task may still be running (and spending) and the row may land just after the deadline. Neither the timeout branch nor the task-failed branch (line 162) logs anything, so a 3600s blocking RPC can give up with no operator-visible event.

**Fix:** log the timeout and the task-failed branches at WARNING/ERROR with `reply_key`/`run_id`/elapsed so a caller-side timeout can be correlated with the still-running/failed `pg_task_result` row.

backend/pg_queue/executor_rpc.py
```diff
@@ -0,0 +1,222 @@
...
+    closed gate, a blind Flipt, or any error — so the executor never silently
+    loses its transport.
+    """
+    if not settings.PG_QUEUE_TRANSPORT_ENABLED:
+        return False
+    if os.environ.get("FLIPT_SERVICE_AVAILABLE", "false").lower() != "true":
+        logger.warning(
+            "resolve_executor_transport: gate ON but FLIPT_SERVICE_AVAILABLE != "
+            "true (Flipt blind); using Celery"
+        )
```
muhammad-ali-e, line 83: **High — Flipt outages silently revert every org to Celery with no metric, and the broad `except` hides programming errors.**

Fail-closed-to-Celery is the correct *behavior*, but both degradation paths (blind Flipt here, and the `except Exception` around `check_feature_flag_status` below) emit only a WARNING. During an active PG rollout, a Flipt outage silently reverts transport with no Sentry error-id / counter, so ops can't distinguish "flag legitimately off" from "Flipt down and we're blind." The broad `except Exception` also swallows real bugs (e.g. an `AttributeError` on `context`, a malformed `flag_context`) identically to a network timeout.

**Fix:** emit a counter/metric (or dedicated error-id log) on the blind-Flipt and Flipt-exception branches; narrow the `except` to the Flipt client exception types where feasible so unexpected types aren't masked as "Flipt check failed."

backend/pg_queue/executor_rpc.py
```diff
@@ -0,0 +1,222 @@
...
+        ``.first()`` opens+closes its own Django query so a freshly committed row
+        from the executor consumer is visible.
+        """
+        deadline = time.monotonic() + timeout
+        delay = _POLL_INITIAL_SECONDS
+        while True:
+            row = PgTaskResult.objects.filter(pk=reply_key).first()
+            if row is not None:
+                return row
+            remaining = deadline - time.monotonic()
```
muhammad-ali-e, line 207: **High — the blocking poll pins a Django DB connection for up to the full timeout (default 3600s).**

`dispatch` blocks the calling web request inside this poll loop, and each `PgTaskResult.objects.filter(...).first()` runs on the request's bound DB connection. A handful of concurrently-blocked prompt-studio dispatches each pin a backend connection for the entire executor runtime, which can exhaust the pool / PgBouncer slots — presenting as a DB outage rather than a queue wait. (The Celery `AsyncResult.get` path this replaces blocked on the broker, not a SQL connection.)

Also: the docstring's visibility claim holds under Django autocommit, but if a caller ever wraps dispatch in `transaction.atomic()` (or `ATOMIC_REQUESTS` is enabled), the loop reads one fixed snapshot and never sees the committed row → blocks to full timeout.

**Fix:** release the connection between polls (e.g. `django.db.close_old_connections()`), and document that dispatch must not be called inside an open transaction.

backend/pg_queue/executor_rpc.py
```diff
@@ -0,0 +1,222 @@
...
+# string is duplicated here rather than imported to avoid a pg_queue →
+# workflow_manager dependency). One flag for the whole feature: on routes the
+# executor (and execution) to PG, off keeps Celery — no per-subsystem flags.
+PG_QUEUE_FLAG_KEY = "pg_queue_enabled"
+_EXECUTE_TASK = "execute_extraction"
+# Mirror the SDK's queue-per-executor convention so the PG executor queue name
+# matches the Celery one (the queue routes by the row's queue_name column).
+_QUEUE_PREFIX = "celery_executor_"
+# Caller-side wait default — mirrors the SDK dispatcher (EXECUTOR_RESULT_TIMEOUT
+# env, else 3600s) so a PG-routed caller waits exactly as long as a Celery one.
```
muhammad-ali-e, line 61: **Medium — `PG_QUEUE_FLAG_KEY = "pg_queue_enabled"` is string-duplicated across three modules.**

The PR's goal is "one flag for the whole feature," but that single source of truth is three independent copies of the literal: here, `workflow_v2/transport.py` (same name), and `scheduler/ownership.py` (as `SCHEDULER_FLAG_KEY`). The inline comment justifies duplication to avoid a `pg_queue → workflow_manager` dependency — but all three modules live in the **same backend Django project**, so a shared constant in a neutral leaf module (e.g. `pg_queue/flags.py`) is dependency-safe and importable by all three. At minimum rename `SCHEDULER_FLAG_KEY` → `PG_QUEUE_FLAG_KEY` so a grep finds every use. This makes "single flag" a structural guarantee rather than three literals kept in sync by hand.

workers/queue_backend/pg_queue/result_backend.py
```diff
@@ -0,0 +1,184 @@
...
+    @contextlib.contextmanager
+    def _cursor(self) -> Iterator[Any]:
+        """Yield a cursor; commit on success, roll back + recover on error."""
+        conn = self.conn
+        try:
+            with conn.cursor() as cur:
+                yield cur
+            conn.commit()
+        except Exception as exc:
+            conn_dead = isinstance(
```
muhammad-ali-e, line 100: **Medium — dead-connection detection is incomplete; a poisoned connection can wedge the executor consumer permanently.**

`conn_dead` is set only for `OperationalError`/`InterfaceError`. A `psycopg2.DatabaseError` from a broken connection mid-statement (or a subclass that isn't one of those two) won't trigger recycle, so the next `store_result` reuses a connection left in an aborted transaction and every subsequent write raises `InFailedSqlTransaction` until the process restarts. Since `store_result` is on the consumer's success path, this can cascade into permanent reply-store failure. The inner rollback-failure handler also discards its exception with no log.

**Fix:** treat any `psycopg2.Error`/`DatabaseError` (or `conn.closed`) as `conn_dead`, and log the swallowed rollback failure. This recovery path is also untested — add an injected-mock-connection test asserting rollback + close + `self._conn = None` on a dead error, and *no* close for an injected (non-owned) connection.

backend/pg_queue/models.py
```diff
@@ -285,3 +285,42 @@ class Meta:
...
+    executor task finishes: the PG executor consumer (``worker-pg-executor``)
+    writes the returned ``ExecutionResult`` (``status="completed"``) or the error
+    text if the task raised (``status="failed"``), keyed by the caller-chosen
+    ``task_id`` (the reply key). The blocking caller polls this table until the
+    row appears or its timeout elapses; a non-blocking poll just checks presence.
+
+    The absence of a row means "not done yet" — there is deliberately no
+    ``pending`` state to maintain. A separate, droppable table that never touches
+    ``WorkflowExecution`` — same posture as ``PgBarrierState`` / ``PgBatchDedup``,
+    extension-free (UN-3533). ``expires_at`` drives the reaper's retention sweep.
```
muhammad-ali-e: **High — docstring claims a retention sweep that does not exist (unbounded table growth).**

"`expires_at` drives the reaper's retention sweep" — but no sweep of `pg_task_result` exists anywhere (the reaper only sweeps `pg_barrier_state` / `pg_batch_dedup`; confirmed by grepping the tree). Once the gate flips, every executor RPC writes a row (each carrying a full `ExecutionResult` JSONB) that is never deleted, so the table grows without bound.

**Fix:** either add the retention sweep to the reaper in this PR (`DELETE FROM pg_task_result WHERE expires_at <= now()`), or soften the docstring to state the sweep is a not-yet-wired follow-up and track it. Shipping the claim that the sweep already exists is misleading.

backend/pg_queue/models.py
```diff
@@ -285,3 +285,42 @@ class Meta:
...
+    # failed row — avoids a string column having two empty states (NULL vs "").
+    error = models.TextField(blank=True, default="")
+    created_at = models.DateTimeField(default=timezone.now)
+    # Retention horizon for the reaper sweep; NULL = no expiry recorded yet.
+    expires_at = models.DateTimeField(null=True, blank=True)
+
+    class Meta:
+        db_table = "pg_task_result"
+        indexes = [
+            # Drives the reaper's retention sweep (DELETE ... WHERE expires_at <= now()).
```
muhammad-ali-e: **Medium (comment accuracy) — describes a `DELETE` that doesn't exist.**

This index comment says it "Drives the reaper's retention sweep (DELETE ... WHERE expires_at <= now())", but no such sweep consumes `pg_task_result.expires_at` — the index currently backs no query. Reword to e.g. "Index for a *future* retention sweep; no sweeper reads it yet." (Same issue at line 318: "NULL = no expiry recorded yet" describes a state no code produces — the only writer always sets `expires_at`.)

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -288,9 +287,18 @@ def _publish_log(
         )
 
     @staticmethod
-    def _get_dispatcher() -> ExecutionDispatcher:
-        """Get an ExecutionDispatcher for the executor worker."""
-        return ExecutionDispatcher(celery_app=celery_app)
+    def _get_dispatcher():
+        """Executor dispatcher for the executor worker.
+
+        Gate-routed: when ``pg_executor_enabled`` is on the blocking
```
muhammad-ali-e: **Medium — stale flag name contradicts the code and the PR's own goal.**

This docstring says "when `pg_executor_enabled` is on", but no such flag exists — the dispatcher reads `PG_QUEUE_FLAG_KEY = "pg_queue_enabled"` (`executor_rpc.py:62`), which is exactly the unification this PR makes. An operator following this comment would look for a flag that isn't there.

**Fix:** change `pg_executor_enabled` → `pg_queue_enabled`. Also: `_get_dispatcher` (line 290) dropped its return annotation; the prior signature was `-> ExecutionDispatcher`. Consider annotating `-> "RoutingExecutionDispatcher"` (under `TYPE_CHECKING`) for consistency with the typed surroundings.

docker/docker-compose.yaml
```diff
@@ -709,6 +709,43 @@ services:
     profiles:
       - pg-queue
 
+  # Executor RPC over PG — runs execute_extraction as a request-reply: claims
+  # work from Postgres and writes the ExecutionResult to pg_task_result for the
+  # blocking caller. Same heavy executor runtime as worker-executor-v2 (tool
+  # execution → platform-service / adapters / file storage), but broker-free: it
+  # is the terminal of the RPC and dispatches nothing onward to Celery. Dark
+  # until the backend pg_executor_enabled gate is flipped.
```
muhammad-ali-e: **Medium — stale flag name.** "Dark until the backend `pg_executor_enabled` gate is flipped" references a non-existent flag. The `worker-pg-executor` service is dark until the unified `pg_queue_enabled` flag (read by `resolve_executor_transport`) is flipped. Change `pg_executor_enabled` → `pg_queue_enabled`.

unstract/core/src/unstract/core/data_models.py
```diff
@@ -315,13 +315,21 @@ class TaskPayload(TypedDict):
...
+    result. It is deliberately a dedicated key, not ``request_id`` (which is a
+    caller-supplied tracing correlation id and may not be unique per dispatch).
     """
 
     task_name: str
     args: list[Any]
     kwargs: dict[str, Any]
     queue: str | None
     fairness: FairnessPayload | None
+    reply_key: NotRequired[str | None]
```
muhammad-ali-e: **Low — `reply_key: NotRequired[str | None]` has three representations for two meanings.**

The key can be absent, present-and-`None`, or present-and-string, but only two are meaningful (fire-and-forget vs request-reply). The producer only omits it or sets a string, and the consumer reads `.get("reply_key")` which collapses absent and `None`. The `| None` arm is dead-but-legal and reintroduces the "two empty states" problem the `PgTaskResult.error` column docstring explicitly set out to avoid.

**Fix:** tighten to `reply_key: NotRequired[str]` — "request-reply" is then exactly "key present."

workers/queue_backend/pg_queue/result_backend.py
```diff
@@ -0,0 +1,184 @@
...
+import psycopg2
+
+from .connection import create_pg_connection
+
+if TYPE_CHECKING:
+    from psycopg2.extensions import connection as PgConnection
+
+logger = logging.getLogger(__name__)
+
+# How long a stored result lives before the reaper's retention sweep may delete
```
muhammad-ali-e, line 50: **Low — retention is decoupled from the caller's actual timeout.**

`DEFAULT_RETENTION_SECONDS = 3600` equals the executor caller-timeout *default*, but a caller can raise `EXECUTOR_RESULT_TIMEOUT` above 3600 (`PgExecutionDispatcher.dispatch` honours it). The docstring's promise that a result "always outlives any caller still waiting on it" then breaks: if the retention sweep is added (see models.py:304), a result could be deleted while a caller is still polling → spurious timeout. These two constants live in separate packages with no shared source.

**Fix:** when the sweep lands, derive retention from `max(EXECUTOR_RESULT_TIMEOUT, DEFAULT_RETENTION_SECONDS)` plus a margin (or pass the caller timeout through the payload), and document the `retention >= caller_timeout` invariant.

muhammad-ali-e (commented):
## Automated PR Review — PG Executor RPC (UN-3603)

Ran the PR Review Toolkit (code review, silent-failure hunt, type-design, test coverage, comment accuracy, simplification) over the 14-file diff. The design is sound and the zero-regression / fail-closed posture holds. Inline comments below are prioritised; the load-bearing ones are:

- **Critical:** unguarded success-path `_store_reply` (consumer.py:249) — a transient result-store error silently re-runs the executor (paid LLM spend) and blocks the caller to its full timeout, logged only as a generic poll failure.
- **High:** `pg_task_result` is never reaped — docstrings claim a retention sweep that does not exist anywhere; the table grows unbounded once the gate flips. Untested caller side (`PgExecutionDispatcher.dispatch`) and consumer reply_key branch. Blocking poll pins a DB connection for up to 3600s.
- **Medium:** two stale `pg_executor_enabled` flag references contradict the unified `pg_queue_enabled` flag; status magic-strings + untyped dict shapes across the process boundary.

Full prioritised summary is in the PR thread.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in dce4e1ca2 — the success-path store is now wrapped: on failure it logs at ERROR (task/msg_id/reply_key) and **acks anyway** to avoid re-running the executor (LLM spend); the caller degrades to a timeout rather than a poison-loop. The task-raised branch and the drop branches go through a new guarded `_fail_reply`. Added consumer tests: success→store+ack, raise→store-error+ack, store-failure→still-acks.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in dce4e1ca2 — `reply_key` is read up front and the drop branches (missing task_name, poison `read_ct`, unknown task) now call `_fail_reply(reply_key, …)` before deleting, so a request-reply caller gets an immediate descriptive failure instead of blocking to its timeout. Tested (unknown-task → store error + ack).

backend/pg_queue/executor_rpc.py
muhammad-ali-e: Fixed in dce4e1ca2 — both: (1) added a shared `PgTaskStatus(str, Enum)` in `unstract.core.data_models`; the reader branches on `PgTaskStatus.COMPLETED.value` and the writer re-exports `STATUS_COMPLETED/FAILED` from it → one source of truth across the process boundary. (2) `from_dict` is wrapped in try/except so a malformed completed row becomes `ExecutionResult.failure(...)`, preserving the never-raises contract.

backend/pg_queue/executor_rpc.py
muhammad-ali-e: Fixed in dce4e1ca2 — added `TestPgExecutionDispatcherDispatch` (8 DB-free cases, mocking `enqueue_task` + `_wait_for_result`): enqueue-raises→failure (no re-raise); timeout→failure with the `within {timeout}s` message; completed→`from_dict`; failed→`row.error`; empty-error→fallback; completed-but-result-None→failure; malformed-completed→failure-not-raise; `timeout=None`→`EXECUTOR_RESULT_TIMEOUT`. The `_wait_for_result` backoff/clamp is the same logic as `PgResultBackend.wait_for_result`, which has real-PG timeout + late-write tests.

backend/pg_queue/executor_rpc.py
muhammad-ali-e: Fixed in dce4e1ca2 — the timeout branch and the executor-failure branch now log at WARNING with reply_key/run_id, so a caller-side timeout/failure is operator-visible and correlatable with the `pg_task_result` row.

backend/pg_queue/executor_rpc.py
muhammad-ali-e: Partially addressed + reasoned. The fail-closed-to-Celery behaviour + WARNING is intentionally identical to `resolve_transport` / `resolve_schedule_owner` (all three now share this exact shape). A metric/error-id is worthwhile but should land **uniformly across all three resolvers**, not just here — filed as a follow-up rather than diverging this one. On the broad `except`: `check_feature_flag_status` already catches internally and the gRPC client raises a variety of types; narrowing risks letting a new Flipt-client exception type through and breaking execution creation, so I kept the broad fail-closed catch (consistent with the siblings).

backend/pg_queue/executor_rpc.py
muhammad-ali-e: Fixed in dce4e1ca2 — `_wait_for_result` now calls `close_old_connections()` between polls so a long RPC doesn't pin a backend connection for its whole duration, and the docstring documents that dispatch must NOT be called inside an open transaction (`transaction.atomic` / `ATOMIC_REQUESTS` would pin one snapshot and never see the committed row).

backend/pg_queue/executor_rpc.py
muhammad-ali-e: Fixed in dce4e1ca2 — added `pg_queue/flags.py` with the single `PG_QUEUE_FLAG_KEY`; `transport.py`, `ownership.py`, and `executor_rpc.py` all import it (no duplicated literals), and `SCHEDULER_FLAG_KEY` is folded into it. `transport.py` re-exports it so existing imports/tests still resolve.

workers/queue_backend/pg_queue/result_backend.py
muhammad-ali-e: Partially addressed + reasoned. Added the missing log on the swallowed rollback failure (now WARNING with exc_info). On broadening to any `psycopg2.Error`: this `_cursor` mirrors `PgQueueClient` 1:1 on purpose, and a successful `rollback()` already clears the aborted txn for recoverable errors (constraint/data) — broadening would over-recycle a healthy connection on those. The truly-dead cases surface as `OperationalError`/`InterfaceError` or make `rollback()` itself raise (→ `conn_dead`). I'd rather fix the shared pattern in `PgQueueClient` + here together than diverge one; happy to do that as a follow-up.

backend/pg_queue/models.py
muhammad-ali-e: Fixed in dce4e1ca2 — softened the docstring to state the retention sweep is **not yet wired** (the table grows per-RPC; acceptable while gated off, tracked as a follow-up before the gate ramps to 100%). I opted to soften rather than add the sweep here: it belongs with the reaper (a worker change + rebuild) and the decommission/ops slice; shipping it half-done would be worse than an honest TODO.

backend/pg_queue/models.py
muhammad-ali-e: Fixed in dce4e1ca2 — reworded the index + `expires_at` comments to "index for a *future* retention sweep; no sweeper reads it yet".

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
muhammad-ali-e: Fixed in dce4e1ca2 — `pg_executor_enabled` → `pg_queue_enabled` in the docstring. (Left `_get_dispatcher` un-annotated to avoid a forward-ref import dance; the routing dispatcher is duck-typed against the SDK surface.)

docker/docker-compose.yaml
muhammad-ali-e: Fixed in dce4e1ca2 — `pg_executor_enabled` → `pg_queue_enabled` in the compose comment.

unstract/core/src/unstract/core/data_models.py
muhammad-ali-e: Fixed in dce4e1ca2 — tightened to `reply_key: NotRequired[str]` so "request-reply" is exactly "key present" (no dead `| None` arm).

workers/queue_backend/pg_queue/result_backend.py
muhammad-ali-e: Acknowledged — deferred with the sweep (models.py:304). With no sweep yet, the decoupling is currently moot (nothing deletes a row out from under a caller). When the sweep lands I'll derive retention from `max(EXECUTOR_RESULT_TIMEOUT, default) + margin` and document the `retention >= caller_timeout` invariant. Tracked in the same follow-up.

muhammad-ali-e:
## Review addressed — `dce4e1ca2`

Thanks for the thorough pass. Disposition of the 16 findings:

**Fixed (13)**
- **Critical** (consumer success-path store): now guarded — a store failure logs at ERROR and **acks anyway** (no expensive executor re-run); caller degrades to a timeout, not a poison-loop.
- **High** (drop branches lose the reply): `reply_key` read up front; malformed/poison/unknown + task-raised branches store a definitive failure via a new guarded `_fail_reply`.
- **High** (connection pinning): `close_old_connections()` between polls + documented "no open transaction".
- **High** (dispatch untested): `TestPgExecutionDispatcherDispatch` — 8 DB-free outcome branches.
- **Med** (magic status + `from_dict`): shared `PgTaskStatus` enum in `unstract.core` (writer+reader agree); `from_dict` wrapped to honour never-raises.
- **Med** (timeout silent): timeout + failure branches now log with reply_key/run_id.
- **Med** (flag string ×3): single `pg_queue/flags.py` `PG_QUEUE_FLAG_KEY`, imported by all three resolvers; `SCHEDULER_FLAG_KEY` folded in.
- **Med** (rollback failure unlogged): now logged.
- **Med/Low** (doc accuracy): softened the not-yet-wired retention-sweep claims; fixed stale `pg_executor_enabled` → `pg_queue_enabled`; `reply_key` → `NotRequired[str]`.
- Plus consumer reply_key tests (success/raise/unknown/store-failure).

**Partially fixed / deferred, with reasoning (3)** — replied inline:
- **High** (`pg_task_result` never reaped): docstrings now state the sweep is **not yet wired** (acceptable while gated off; tracked before the gate ramps to 100%). The actual sweep belongs with the reaper + the decommission slice — filed as follow-up rather than shipped half-done.
- **High** (Flipt-outage observability / broad `except`): kept consistent with the sibling resolvers (`resolve_transport` / `resolve_schedule_owner`); a metric/error-id should land uniformly across all three — follow-up.
- **Med** (dead-conn breadth): added the missing log; kept the `PgQueueClient`-mirrored recycle logic (a successful `rollback()` already recovers non-fatal errors; broadening would over-recycle) — would fix both modules together as a follow-up.

All tests green (45 backend incl. 8 new dispatch cases; result-backend + consumer reply_key suites), pre-commit clean, no migration drift. Verdict noted — appreciated.

muhammad-ali-e:
## SonarCloud (round 2) — `0cf46ea9d`

The 3 new findings (all introduced by the previous review-fix commit) are fixed:
- `consumer.py` — the two new request-reply error handlers now use `logger.exception()` instead of `logger.error(..., exc_info=True)`.
- `test_executor_rpc.py:124` — reworded the `# timeout=None` inline comment (Sonar read it as commented-out code).

Tests green (16 executor-rpc + consumer reply_key suite), pre-commit clean.

muhammad-ali-e:
## greptile addressed — `022e22c21`

Thanks — the VT finding is a real one and would have bitten at ramp time.

**The load-bearing finding (worker-pg-executor VT / health-stale):** Fixed. The executor runs LLM inference up to `EXECUTOR_TASK_TIME_LIMIT=3600s`, but the consumer defaults are vt=30s / health-stale=60s — so a long task would be re-claimed mid-run (double execution + token double-spend) or killed by the liveness probe. The `worker-pg-executor` service now sets, **above** the 3600s task limit, both overridable:
- `WORKER_PG_QUEUE_CONSUMER_VT_SECONDS=3660`
- `WORKER_PG_QUEUE_CONSUMER_HEALTH_STALE_SECONDS=3720`

(`docker compose config` confirms they resolve.)

**Minor never-raises gap (env int-parse outside try/except):** Fixed — `EXECUTOR_RESULT_TIMEOUT` is parsed inside a `try/except (TypeError, ValueError)` that falls back to the default, so a misconfigured value can't raise out of `dispatch()`. Added a test.

The earlier idempotency / drop-branch / ack-on-store-failure items you noted as already solid were from the prior review round. Tests green (17 executor-rpc), pre-commit + `docker compose config` clean.

PR #2095, opened 2026-06-20
UN-3605 [GATED-FEAT] PG Queue — structure_tool executor RPC over Postgres (+ register executors in PG consumer)

muhammad-ali-e:
## What

Routes the in-workflow **structure_tool** executor dispatch onto the Postgres executor RPC when the single `pg_queue_enabled` flag is on; Celery otherwise (**zero-regression by construction**). Next slice after UN-3603 (prompt-studio blocking path). Gated off by default; lands on the PG-queue integration branch (not main).

## Changes

- **`workers/queue_backend/pg_queue/executor_rpc.py` (new)** — workers twin of the backend module on worker primitives:
  - `resolve_executor_transport` — master `PG_QUEUE_TRANSPORT_ENABLED` env, then the single `pg_queue_enabled` Flipt flag; **fail-closed to Celery**.
  - `PgExecutionDispatcher` — enqueue `execute_extraction` with a unique `reply_key` via `PgQueueClient`, poll `PgResultBackend.wait_for_result`; **never raises** (timeout/failure → `ExecutionResult.failure`).
  - `RoutingExecutionDispatcher` — per-call gate routing; `dispatch_async`/`dispatch_with_callback` stay on Celery (later slice). `get_executor_dispatcher` factory.
- **`to_payload`** gains an optional `reply_key` (request-reply marker; fire-and-forget rows stay byte-identical).
- **`structure_tool_task`** swaps its dispatcher factory to `get_executor_dispatcher`; the 3 blocking dispatch call sites are unchanged.
- **Fix (latent gap exposed by this slice):** the PG executor consumer registered **no executors** — `No executor registered with name 'legacy'`. The `@ExecutorRegistry.register` side-effect import lived only in the Celery `executor/worker.py`, but the PG consumer bootstraps via `executor/tasks.py`. Importing `executor.executors` from `executor/tasks.py` means any worker registering `execute_extraction` (Celery **or** PG) has the executors — also fixes the prompt-studio PG path from UN-3603.
- **`sample.env`** — document the worker-side `PG_QUEUE_TRANSPORT_ENABLED` gate; fix a stale `pg_queue_execution_enabled` → `pg_queue_enabled` doc reference.

## Zero-regression

Gate off → `resolve_executor_transport` returns before any Flipt call → `RoutingExecutionDispatcher` delegates to the unchanged SDK `ExecutionDispatcher` (Celery). No `pg_task_result` row. Flipping the flag is an instant, no-redeploy rollout/rollback (read per dispatch).

## Tests

- `workers/tests/test_executor_rpc.py` — gate fail-closed matrix, zero-regression routing (gate off → Celery delegate, no enqueue), never-raises dispatch cases.
- `workers/tests/test_executor_registration.py` — subprocess regression asserting that importing `executor.tasks` alone registers the executors (mirrors the PG consumer bootstrap).

## Dev-tested end-to-end (gated, running stack)

- **Flag ON** → `Executor RPC → PG transport` logged → `execute_extraction` runs on `worker-pg-executor` → `pg_task_result` `completed` rows (reply_keys match) → file executions + workflow **COMPLETED**.
- **Flag OFF** → Celery executor path, no `pg_task_result` row.

## Out of scope (separate follow-ups)

- `worker-pg-fileproc` single-consumer serializes multi-file fan-out on the PG path (own ticket).
- Continuation/callback path and Celery decommission are later slices.

### DIFF
--- a/workers/executor/tasks.py
+++ b/workers/executor/tasks.py
@@ -9,9 +9,11 @@ ExecutionOrchestrator, and returns an ExecutionResult dict.
 # decorators run before ``execute_extraction`` can be invoked. Coupling this to
 # the task module (not only the Celery ``executor/worker.py`` entrypoint) ensures
 # the registry is populated wherever the task is registered — in particular the PG
-# executor consumer (``pg_queue_consumer`` loads ``tasks.py`` but not
-# ``worker.py``), which would otherwise hit "No executor registered". Import is
-# idempotent (module cached), so the Celery path importing it twice is harmless.
+# executor consumer, which bootstraps via the root-worker import that loads
+# ``executor/tasks.py`` (this module) but NOT ``executor/worker.py`` where this
+# import historically lived; without it the consumer hits "No executor
+# registered". Import is idempotent (module cached), so the Celery entrypoint
+# importing it again is harmless.
 import executor.executors  # noqa: E402, F401
 from queue_backend import worker_task
 from shared.clients import UsageAPIClient
--- a/workers/executor/tasks.py
+++ b/workers/executor/tasks.py
@@ -5,6 +5,14 @@ ExecutionContext dict, runs the appropriate executor via
 ExecutionOrchestrator, and returns an ExecutionResult dict.
 """
 
+# Import the executor implementations so their ``@ExecutorRegistry.register``
+# decorators run before ``execute_extraction`` can be invoked. Coupling this to
+# the task module (not only the Celery ``executor/worker.py`` entrypoint) ensures
+# the registry is populated wherever the task is registered — in particular the PG
+# executor consumer (``pg_queue_consumer`` loads ``tasks.py`` but not
+# ``worker.py``), which would otherwise hit "No executor registered". Import is
+# idempotent (module cached), so the Celery path importing it twice is harmless.
+import executor.executors  # noqa: E402, F401
 from queue_backend import worker_task
 from shared.clients import UsageAPIClient
 from shared.enums.task_enums import TaskName
--- a/workers/file_processing/structure_tool_task.py
+++ b/workers/file_processing/structure_tool_task.py
@@ -25,12 +25,15 @@ from typing import Any
 from file_processing.worker import app
 from queue_backend import FairnessKey, worker_task
 from queue_backend.fairness import WorkloadType
+from queue_backend.pg_queue.executor_rpc import (
+    RoutingExecutionDispatcher,
+    get_executor_dispatcher,
+)
 from shared.enums.task_enums import TaskName
 from shared.infrastructure.context import StateStore
 
 from unstract.sdk1.constants import ToolEnv, UsageKwargs
 from unstract.sdk1.execution.context import ExecutionContext
-from unstract.sdk1.execution.dispatcher import ExecutionDispatcher
 from unstract.sdk1.execution.result import ExecutionResult
 
 logger = logging.getLogger(__name__)
@@ -276,7 +279,9 @@ def _execute_structure_tool_impl(params: dict) -> dict:
     )
 
     platform_helper = _create_platform_helper(shim, file_execution_id)
-    dispatcher = ExecutionDispatcher(celery_app=app)
+    # Gate-routed: PG executor RPC when pg_queue_enabled is on, else the unchanged
+    # Celery ExecutionDispatcher (zero-regression by construction — see executor_rpc).
+    dispatcher = get_executor_dispatcher(celery_app=app)
     fs = _get_file_storage()
 
     # ---- Step 2: Fetch tool metadata ----
@@ -675,7 +680,7 @@ def _run_agentic_extraction(
     input_file_path: str,
     output_dir_path: str,
     tool_instance_metadata: dict,
-    dispatcher: ExecutionDispatcher,
+    dispatcher: RoutingExecutionDispatcher,
     shim: Any,
     file_execution_id: str,
     execution_id: str,
--- a/workers/queue_backend/pg_queue/executor_rpc.py
+++ b/workers/queue_backend/pg_queue/executor_rpc.py
@@ -1,0 +1,282 @@
+"""Executor-RPC transport routing for the PG path — workers side (Phase 9, ③b-2).
+
+The workers-side twin of ``backend/pg_queue/executor_rpc.py``. The executor "RPC"
+is a synchronous request-reply: a caller (here, the ``structure_tool`` task in the
+file_processing worker) sends an ``ExecutionContext`` to the executor worker and
+blocks for the ``ExecutionResult``. The legacy transport is Celery — the SDK
+``ExecutionDispatcher`` (``send_task`` + ``AsyncResult.get``). This module adds a
+**parallel** Postgres transport that leaves Celery and the SDK completely
+untouched (no SDK edit, no change to the ``execute_extraction`` task or the Celery
+executor worker):
+
+- :class:`PgExecutionDispatcher` enqueues ``execute_extraction`` onto the PG queue
+  (via :class:`~queue_backend.pg_queue.client.PgQueueClient`) with a unique
+  ``reply_key`` and polls ``pg_task_result`` (via
+  :class:`~queue_backend.pg_queue.result_backend.PgResultBackend`) for the reply —
+  same ``.dispatch()`` contract as the SDK dispatcher (never raises;
+  failure/timeout → ``ExecutionResult.failure``). The already-running
+  ``worker-pg-executor`` consumer runs the task and writes the reply, so this side
+  only adds the enqueue + poll halves.
+- :func:`resolve_executor_transport` is the gate: master
+  ``PG_QUEUE_TRANSPORT_ENABLED`` (env, the workers analogue of the backend's
+  Django setting) then the **single** Flipt flag ``pg_queue_enabled`` — the same
+  flag the execution path uses, so one flip turns the whole PG-queue feature
+  on/off. Fails closed to Celery.
+- :class:`RoutingExecutionDispatcher` is what ``structure_tool`` gets from
+  :func:`get_executor_dispatcher`: ``dispatch()`` picks PG-vs-Celery **per call**
+  (read at dispatch time → flipping the flag is an instant, no-redeploy
+  rollout/rollback); ``dispatch_async`` / ``dispatch_with_callback`` always
+  delegate to Celery (the callback path is a later slice).
+
+Zero-regression: gate off ⇒ every method delegates to the unchanged Celery
+``ExecutionDispatcher`` and no ``pg_task_result`` row is created.
+
+TODO(shared): ``resolve_executor_transport`` and the reply_key/poll orchestration
+mirror the backend module almost verbatim — only the transport primitives differ
+(psycopg2 here vs Django ORM there). A later slice can lift the shared logic into
+``unstract.core`` so the gate/contract lives in one place.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+import uuid
+from typing import TYPE_CHECKING, Any
+
+from unstract.core.data_models import PgTaskStatus
+from unstract.flags.feature_flag import check_feature_flag_status
+from unstract.sdk1.execution.dispatcher import ExecutionDispatcher
+from unstract.sdk1.execution.result import ExecutionResult
+
+from .client import PgQueueClient
+from .result_backend import PgResultBackend
+from .task_payload import to_payload
+
+if TYPE_CHECKING:
+    from unstract.sdk1.execution.context import ExecutionContext
+
+logger = logging.getLogger(__name__)
+
+# The single PG-queue rollout flag — same key the execution path and scheduler
+# read. Defined in backend/pg_queue/flags.py too; kept as a literal here because
+# workers can't import backend code (see TODO(shared) in the module docstring).
+PG_QUEUE_FLAG_KEY = "pg_queue_enabled"
+# Master kill-switch + deploy-ordering gate. The workers analogue of the backend's
+# ``settings.PG_QUEUE_TRANSPORT_ENABLED`` — read straight from the env here.
+_MASTER_GATE_ENV = "PG_QUEUE_TRANSPORT_ENABLED"
+
+_EXECUTE_TASK = "execute_extraction"
+# Mirror the SDK's queue-per-executor convention so the PG executor queue name
+# matches the Celery one (the worker-pg-executor consumer subscribes to these).
+_QUEUE_PREFIX = "celery_executor_"
+# Caller-side wait default — mirrors the SDK dispatcher (EXECUTOR_RESULT_TIMEOUT
+# env, else 3600s) so a PG-routed caller waits exactly as long as a Celery one.
+_DEFAULT_TIMEOUT_ENV = "EXECUTOR_RESULT_TIMEOUT"
+_DEFAULT_TIMEOUT = 3600
+
+
+def resolve_executor_transport(context: ExecutionContext) -> bool:
+    """True → route this executor dispatch over PG; False → Celery (default).
+
+    Mirrors the backend ``resolve_executor_transport``: master-gated by the
+    ``PG_QUEUE_TRANSPORT_ENABLED`` env, then the **single** ``pg_queue_enabled``
+    Flipt flag (shared across the whole PG-queue feature), bucketed per org.
+    **Fails closed to Celery** on a closed gate, a blind Flipt, or any error — so
+    the executor never silently loses its transport.
+    """
+    if os.environ.get(_MASTER_GATE_ENV, "false").lower() != "true":
+        return False
+    if os.environ.get("FLIPT_SERVICE_AVAILABLE", "false").lower() != "true":
+        logger.warning(
+            "resolve_executor_transport: gate ON but FLIPT_SERVICE_AVAILABLE != "
+            "true (Flipt blind); using Celery"
+        )
+        return False
+    org = getattr(context, "organization_id", None)
+    # %-bucket keyed on org; fall back to run_id so a context without an org still
+    # resolves deterministically (mirrors the backend resolver).
+    entity_id = str(org or getattr(context, "run_id", "") or "default")
+    flag_context = {"executor_name": str(context.executor_name)}
+    if org:
+        flag_context["organization_id"] = str(org)
+    try:
+        enabled = check_feature_flag_status(
+            flag_key=PG_QUEUE_FLAG_KEY, entity_id=entity_id, context=flag_context
+        )
+    except Exception:
+        logger.warning(
+            "resolve_executor_transport: Flipt check failed; using Celery",
+            exc_info=True,
+        )
+        return False
+    return bool(enabled)
+
+
+class PgExecutionDispatcher:
+    """PG request-reply executor dispatch — drop-in for ``ExecutionDispatcher.dispatch``.
+
+    Enqueues ``execute_extraction`` with a unique ``reply_key`` and blocks on
+    ``pg_task_result`` until the executor consumer records the result or the
+    timeout elapses. Honours the same contract as the SDK dispatcher: it never
+    raises and converts a timeout/failure into ``ExecutionResult.failure`` so
+    callers can branch on ``result.success`` identically on either transport.
+    """
+
+    def dispatch(
+        self,
+        context: ExecutionContext,
+        timeout: int | None = None,
+    ) -> ExecutionResult:
+        if timeout is None:
+            # Guard the env parse so a misconfigured EXECUTOR_RESULT_TIMEOUT can't
+            # raise out of dispatch() (the never-raises contract).
+            try:
+                timeout = int(os.environ.get(_DEFAULT_TIMEOUT_ENV, _DEFAULT_TIMEOUT))
+            except (TypeError, ValueError):
+                timeout = _DEFAULT_TIMEOUT
+        reply_key = str(uuid.uuid4())
+        queue = f"{_QUEUE_PREFIX}{context.executor_name}"
+        org = str(getattr(context, "organization_id", "") or "")
+        try:
+            self._enqueue(queue, context, reply_key, org)
+        except Exception as exc:
+            logger.exception(
+                "PG executor dispatch: enqueue failed (executor=%s run_id=%s)",
+                context.executor_name,
+                context.run_id,
+            )
+            return ExecutionResult.failure(error=f"{type(exc).__name__}: {exc}")
+        logger.info(
+            "PG executor dispatch: enqueued reply_key=%s queue=%s run_id=%s "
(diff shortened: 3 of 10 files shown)


muhammad-ali-e (commented):
Automated PR Review Toolkit pass (Code Reviewer, Silent-Failure Hunter, Type-Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier).

**Overall:** a careful, well-documented, faithful workers-side port of the backend executor RPC. The gating, the never-raises contract, and the zero-regression-when-off property are correct. No blocking bugs. Findings below are hardening + test-coverage items; two HIGH/MED items (corrupt `METADATA.json` swallow at `structure_tool_task.py:~858`; stale `~508` line-ref at `:~453`) fall outside this PR's diff and are in the summary rather than inline.

workers/queue_backend/pg_queue/executor_rpc.py
```diff
@@ -0,0 +1,282 @@
...
+                "the executor task may still be running",
+                timeout,
+                reply_key,
+                context.run_id,
+            )
+            return ExecutionResult.failure(
+                error=f"TimeoutError: executor reply not received within {timeout}s"
+            )
+        if (
+            row["status"] == PgTaskStatus.COMPLETED.value
```
muhammad-ali-e: **[Medium — correctness / never-raises]** `row["status"]` uses bracket access (KeyError-raising), while `result`/`error` below use `.get()`. This access is **not** inside the malformed-result `try` (that only guards `from_dict`), so a result row missing `status` would raise straight out of `dispatch()` and violate the documented never-raises contract.

```python
if (
    row.get("status") == PgTaskStatus.COMPLETED.value
    and row.get("result") is not None
):
```

Producer (`result_backend.get_result`) always sets `status` today, so this is latent — but the contract should not depend on the producer.

workers/queue_backend/pg_queue/executor_rpc.py
```diff
@@ -0,0 +1,282 @@
...
+        context: ExecutionContext,
+        timeout: int | None = None,
+    ) -> ExecutionResult:
+        if timeout is None:
+            # Guard the env parse so a misconfigured EXECUTOR_RESULT_TIMEOUT can't
+            # raise out of dispatch() (the never-raises contract).
+            try:
+                timeout = int(os.environ.get(_DEFAULT_TIMEOUT_ENV, _DEFAULT_TIMEOUT))
+            except (TypeError, ValueError):
+                timeout = _DEFAULT_TIMEOUT
```
muhammad-ali-e, line 150: **[Medium — silent config masking]** Guarding the parse so a bad `EXECUTOR_RESULT_TIMEOUT` can't break the never-raises contract is right, but the swallow is silent: an operator who fat-fingers `EXECUTOR_RESULT_TIMEOUT=3O00` gets the 3600s default with no signal, and every RPC may wait up to an hour instead of the intended value. Log the fallback:

```python
except (TypeError, ValueError):
    logger.warning(
        "Invalid %s=%r; falling back to %ss",
        _DEFAULT_TIMEOUT_ENV, os.environ.get(_DEFAULT_TIMEOUT_ENV), _DEFAULT_TIMEOUT,
    )
    timeout = _DEFAULT_TIMEOUT
```

workers/queue_backend/pg_queue/executor_rpc.py
```diff
@@ -0,0 +1,282 @@
...
+    timeout elapses. Honours the same contract as the SDK dispatcher: it never
+    raises and converts a timeout/failure into ``ExecutionResult.failure`` so
+    callers can branch on ``result.success`` identically on either transport.
+    """
+
+    def dispatch(
+        self,
+        context: ExecutionContext,
+        timeout: int | None = None,
+    ) -> ExecutionResult:
```
muhammad-ali-e, line 131: **[Medium — substitutability / type design]** `PgExecutionDispatcher.dispatch(context, timeout)` omits the `headers` param that both the SDK `ExecutionDispatcher.dispatch` and `RoutingExecutionDispatcher.dispatch` accept. Today only `RoutingExecutionDispatcher` is handed out (and it drops headers on the PG path intentionally), so this is latent — but anyone calling a `PgExecutionDispatcher` directly (the tests do) gets a non-substitutable object, and nothing static catches it. Minimum fix: add `headers: dict[str, Any] | None = None` (accept-and-ignore) so the `dispatch` shapes match. Better: define an `ExecutorDispatcher` `Protocol` in `unstract.core` and annotate the SDK base + both PG classes against it — structural, so it preserves the "no SDK edit" constraint while letting mypy verify the duck-type.

workers/queue_backend/pg_queue/executor_rpc.py
```diff
@@ -0,0 +1,282 @@
...
+            except Exception:
+                # A malformed completed row becomes a failure result, not a raise.
+                logger.exception(
+                    "PG executor dispatch: malformed completed result "
+                    "(reply_key=%s run_id=%s)",
+                    reply_key,
+                    context.run_id,
+                )
+                return ExecutionResult.failure(
+                    error=f"Malformed executor result for reply_key {reply_key}"
```
muhammad-ali-e: **[Low]** The `logger.exception` captures the traceback (good), but the caller-facing `ExecutionResult.failure` only says `Malformed executor result for reply_key {reply_key}` — the actual parse cause isn't surfaced to a UI reading `result.error`. For parity with the enqueue/wait paths (which include `type(exc).__name__`), consider `error=f"Malformed executor result ({type(exc).__name__}) for reply_key {reply_key}"`.

workers/queue_backend/pg_queue/executor_rpc.py
```diff
@@ -0,0 +1,282 @@
...
+    @staticmethod
+    def _wait_for_result(reply_key: str, timeout: float) -> dict[str, Any] | None:
+        """Poll ``pg_task_result`` until the row appears or *timeout* elapses.
+
+        Poll-based with capped backoff (PgBouncer-safe; no LISTEN/NOTIFY). The
+        backend owns one connection for the duration of the wait (one per
+        in-flight RPC; structure_tool dispatches are sequential) and closes it on
+        exit, so a long RPC never leaks a connection.
+        """
+        with PgResultBackend() as rb:
```
muhammad-ali-e, line 268: **[Low — doc/concurrency note]** The docstring ("structure_tool dispatches are sequential", "one per in-flight RPC") is true within one task, but the file_processing worker runs `--pool=prefork` with `WORKER_FILE_PROCESSING_CONCURRENCY` (default 4), so up to 4 processes can each hold a `PgResultBackend` connection open for the full RPC (up to `EXECUTOR_RESULT_TIMEOUT`, default 3600s). Bounded by concurrency and not a leak (context manager closes on exit), but it differs from the backend twin, which calls `close_old_connections()` between polls. Worth a one-line note that the pin is bounded by worker concurrency, or release the connection between polls if file_processing concurrency is ever raised.

workers/queue_backend/pg_queue/executor_rpc.py
```diff
@@ -0,0 +1,282 @@
...
+    if os.environ.get("FLIPT_SERVICE_AVAILABLE", "false").lower() != "true":
+        logger.warning(
+            "resolve_executor_transport: gate ON but FLIPT_SERVICE_AVAILABLE != "
+            "true (Flipt blind); using Celery"
+        )
+        return False
+    org = getattr(context, "organization_id", None)
+    # %-bucket keyed on org; fall back to run_id so a context without an org still
+    # resolves deterministically (mirrors the backend resolver).
+    entity_id = str(org or getattr(context, "run_id", "") or "default")
```
muhammad-ali-e, line 99: **[Low — test gap]** The org-less fallback `entity_id = str(org or getattr(context, "run_id", "") or "default")` and the conditional `organization_id` insertion (only when `org` is truthy) are unexercised — every `TestResolveExecutorTransport` case passes an org. A bug bucketing org-less contexts inconsistently would cause non-deterministic rollout. Suggest: a `_ctx(org=None)` case asserting `entity_id == "run-1"` and `"organization_id" not in flag_context`.

workers/queue_backend/pg_queue/executor_rpc.py
```diff
@@ -0,0 +1,282 @@
...
+        connection is pinned for the whole (possibly long) RPC.
+        """
+        payload = to_payload(
+            _EXECUTE_TASK,
+            args=[context.to_dict()],
+            queue=queue,
+            reply_key=reply_key,
+        )
+        with PgQueueClient() as client:
+            client.send(queue, payload, org_id=org_id)
```
muhammad-ali-e, line 252: **[Medium — test gap]** `_enqueue` is mocked in all 11 dispatch tests, so nothing verifies the actual PG-transport wiring: that it builds `to_payload(_EXECUTE_TASK, args=[context.to_dict()], queue=queue, reply_key=reply_key)` and passes `org_id=org` to `client.send`. Since the queue name, the `args` shape, and `org_id` are the *only* routing/identity info on the PG path (Celery headers are dropped), a bug here misroutes or breaks org-fairness with no test catching it. Suggest a test that patches `PgQueueClient` and asserts `client.send` is called with the queue, `org_id`, and a payload carrying `task_name`/`args`/`reply_key`.

workers/queue_backend/pg_queue/executor_rpc.py
```diff
@@ -0,0 +1,282 @@
...
+            logger.info(
+                "Executor RPC → PG transport (executor=%s run_id=%s)",
+                context.executor_name,
+                context.run_id,
+            )
+            # PG carries org/routing via the enqueue payload, not Celery headers,
+            # so the fairness headers are intentionally not forwarded here (parity
+            # with the backend executor RPC).
+            return self._pg.dispatch(context, timeout=timeout)
+        return self._celery.dispatch(context, timeout=timeout, headers=headers)
```
muhammad-ali-e, line 301: **[Medium — test gap]** The routing tests assert only *which* sub-dispatcher is called, never the arguments. Two PR claims are unverified: (a) gate-OFF forwards `headers=`/`timeout=` through to `celery.dispatch` unchanged (the zero-regression claim is incomplete without this), and (b) gate-ON calls `pg.dispatch` *without* `headers` (the intentional drop documented just above). Suggest asserting `celery.dispatch.assert_called_once_with(..., timeout=9, headers={...})` in the gate-off test and `"headers" not in pg.dispatch.call_args.kwargs` in the gate-on test.

workers/queue_backend/pg_queue/task_payload.py
```diff
@@ -32,12 +32,23 @@ def to_payload(
...
         task_name=task_name,
         args=list(args) if args is not None else [],
         kwargs=dict(kwargs) if kwargs is not None else {},
         queue=queue,
         fairness=fairness.to_dict() if fairness is not None else None,
     )
+    # Only set for request-reply dispatches — keeps fire-and-forget rows
+    # byte-identical to before this field existed (mirrors the backend producer).
+    if reply_key is not None:
+        payload["reply_key"] = reply_key
```
muhammad-ali-e, line 53: **[Medium — test gap]** This new `reply_key` branch is the producer side of the entire RPC contract, but `TestToPayload` (in `test_dispatch_pg.py`) predates it and never passes `reply_key`. A regression that drops or mis-keys the field makes the consumer (`payload.get("reply_key")`) silently treat the message as fire-and-forget, and the caller blocks until timeout. Add two cases: `to_payload("t", reply_key="rk")["reply_key"] == "rk"`, and `"reply_key" not in to_payload("t")` (the byte-identical fire-and-forget guarantee the docstring claims).

workers/file_processing/structure_tool_task.py
```diff
@@ -276,7 +279,9 @@ def _execute_structure_tool_impl(params: dict) -> dict:
     )
 
     platform_helper = _create_platform_helper(shim, file_execution_id)
-    dispatcher = ExecutionDispatcher(celery_app=app)
+    # Gate-routed: PG executor RPC when pg_queue_enabled is on, else the unchanged
+    # Celery ExecutionDispatcher (zero-regression by construction — see executor_rpc).
+    dispatcher = get_executor_dispatcher(celery_app=app)
```
muhammad-ali-e, line 284: **[Low — test gap]** This is the actual call-site swap the PR makes (direct `ExecutionDispatcher` → `get_executor_dispatcher(celery_app=app)` returning a `RoutingExecutionDispatcher`), but no test verifies it. `test_structure_tool_task.py` only covers `_fairness_headers` in isolation. If `get_executor_dispatcher` were mis-imported or returned the wrong type, nothing fails. Suggest patching `structure_tool_task.get_executor_dispatcher` and asserting `_execute_structure_tool_impl` calls it with `celery_app=app` and uses the result as the dispatcher.

workers/executor/tasks.py
```diff
@@ -5,6 +5,14 @@
...
 """
 
+# Import the executor implementations so their ``@ExecutorRegistry.register``
+# decorators run before ``execute_extraction`` can be invoked. Coupling this to
+# the task module (not only the Celery ``executor/worker.py`` entrypoint) ensures
+# the registry is populated wherever the task is registered — in particular the PG
+# executor consumer (``pg_queue_consumer`` loads ``tasks.py`` but not
+# ``worker.py``), which would otherwise hit "No executor registered". Import is
+# idempotent (module cached), so the Celery path importing it twice is harmless.
+import executor.executors  # noqa: E402, F401
```
muhammad-ali-e, line 17: **[Low — comment clarity]** Verified: this import *is* necessary and not redundant — the PG consumer bootstraps via root `worker.py` → `WorkerBuilder` → `importlib.import_module("executor.tasks")` (`WorkerType.to_import_path()` returns `executor.tasks`), which does **not** import `executor/worker.py` where `import executor.executors` historically lived. So the fix is correct. Only the wording can mislead: `pg_queue_consumer/__main__.py` literally does `import worker` (the *root* worker), so "loads `tasks.py` but not `worker.py`" reads as contradictory. Suggest clarifying it means `executor/worker.py` (the Celery entrypoint), e.g. "...loads `executor/tasks.py` (via the root-worker bootstrap) but not `executor/worker.py`...".

workers/queue_backend/pg_queue/executor_rpc.py
muhammad-ali-e: Fixed in 6e3e5c6c2 — `row.get("status")` now (result/error already used `.get`), so a status-less row can't raise out of `dispatch()`. Never-raises no longer depends on the producer writing every key.

workers/queue_backend/pg_queue/executor_rpc.py
muhammad-ali-e: Fixed in 6e3e5c6c2 — the `EXECUTOR_RESULT_TIMEOUT` parse fallback now logs a WARNING (with the bad value) before defaulting to 3600s.

workers/queue_backend/pg_queue/executor_rpc.py
muhammad-ali-e: Fixed in 6e3e5c6c2 — added `headers: dict | None = None` (accept-and-ignore) to `PgExecutionDispatcher.dispatch` so the shape matches the SDK/Routing dispatch. The shared `ExecutorDispatcher` Protocol in `unstract.core` is the broader `TODO(shared)` (same place the de-dup would land).

workers/queue_backend/pg_queue/executor_rpc.py
muhammad-ali-e: Fixed in 6e3e5c6c2 — the malformed-result failure now includes `type(exc).__name__`, matching the enqueue/wait paths.

workers/queue_backend/pg_queue/executor_rpc.py
muhammad-ali-e: Fixed in 6e3e5c6c2 — added a docstring note that the pin is bounded by file_processing prefork concurrency (default 4), and to release between polls like the backend twin if that concurrency is raised materially. Kept connection-per-wait for now (simpler; PgResultBackend owns + closes it).

workers/queue_backend/pg_queue/executor_rpc.py
muhammad-ali-e: Fixed in 6e3e5c6c2 — added `test_org_less_context_buckets_on_run_id`: org=None → `entity_id == "run-1"` and `organization_id` absent from the Flipt context.

workers/queue_backend/pg_queue/executor_rpc.py
muhammad-ali-e: Fixed in 6e3e5c6c2 — added `TestPgExecutionDispatcherEnqueueWiring` that patches `PgQueueClient` and asserts `client.send` gets the queue (`celery_executor_legacy`), `org_id`, and a payload carrying `task_name`/`args`/`reply_key` (`to_payload` runs for real).

workers/queue_backend/pg_queue/executor_rpc.py
muhammad-ali-e: Fixed in 6e3e5c6c2 — routing tests now assert arguments: gate-off forwards `timeout=` and `headers=` to `celery.dispatch` unchanged; gate-on calls `pg.dispatch` with `timeout=` and `"headers" not in kwargs` (the intentional drop).

workers/queue_backend/pg_queue/task_payload.py
muhammad-ali-e: Fixed in 6e3e5c6c2 — added two `TestToPayload` cases: `reply_key` set when passed, and absent when not (the byte-identical fire-and-forget guarantee).

workers/file_processing/structure_tool_task.py
muhammad-ali-e: Fixed in 6e3e5c6c2 — added `TestDispatcherFactory` asserting `_execute_structure_tool_impl` builds its dispatcher via `get_executor_dispatcher(celery_app=app)`.

workers/executor/tasks.py
muhammad-ali-e: Fixed in 6e3e5c6c2 — reworded the comment to name `executor/worker.py` (the Celery entrypoint the PG-consumer bootstrap skips) instead of the ambiguous `worker.py`. Thanks for confirming the fix is correct.

workers/queue_backend/pg_queue/executor_rpc.py
```diff
@@ -0,0 +1,282 @@
...
+                    error=f"Malformed executor result for reply_key {reply_key}"
+                )
+        logger.warning(
+            "PG executor dispatch: executor reported failure (reply_key=%s "
+            "run_id=%s): %s",
+            reply_key,
+            context.run_id,
+            row.get("error") or "(no error)",
+        )
+        return ExecutionResult.failure(error=row.get("error") or "executor task failed")
```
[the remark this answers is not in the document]
muhammad-ali-e: Acknowledged — added an explicit note in the timeout branch documenting the orphaned-task / retry-double-run risk: on caller timeout the executor task may still finish and write its (now-unread) row, and a workflow retry re-dispatches with a fresh `reply_key`, so two executor tasks for one file can overlap. De-dup belongs at the file-execution layer; this transport stays at-least-once + caller-timeout by design. Made explicit in 6e3e5c6c2.

muhammad-ali-e:
## Review addressed — `6e3e5c6c2`

Thanks for the thorough pass. All toolkit + greptile items addressed; **30 workers tests green, pre-commit clean.**

**Code (executor_rpc.py):**
- `row.get("status")` so a status-less row can't break the never-raises contract.
- `dispatch` accepts (and ignores) `headers` — substitutable with the SDK/Routing shapes.
- `EXECUTOR_RESULT_TIMEOUT` parse fallback now logs a WARNING instead of swallowing.
- Malformed-result error includes `type(exc).__name__`.
- Docstring note: the poll connection pin is bounded by file_processing prefork concurrency (vs the backend twin's `close_old_connections` per poll).
- Timeout branch: explicit orphaned-task / retry-double-run note (greptile) — de-dup belongs at the file-execution layer; transport stays at-least-once + caller-timeout.
- `executor/tasks.py`: reworded the registration-import comment (`executor/worker.py`, not the ambiguous `worker.py`).

**Tests (the coverage gaps):**
- org-less resolve bucketing (entity_id=run_id, no org in context)
- real `_enqueue` wiring (queue / org_id / payload via `PgQueueClient.send`)
- routing arg passthrough (gate-off forwards timeout+headers; gate-on drops headers)
- `to_payload` reply_key set/omitted
- `structure_tool` factory-swap call site

**SonarCloud — 20.4% duplication on new code:** this is the *intentional* workers-twin of `backend/pg_queue/executor_rpc.py` (two transports on different primitives — psycopg2 + env here vs Django ORM + settings there). A true de-dup is the `TODO(shared)` already noted in the module: lift `resolve_executor_transport` + the reply_key/poll contract into `unstract.core` and annotate an `ExecutorDispatcher` Protocol — a deliberately separate slice (it also touches the already-merged backend module). Happy to do that extraction as a follow-up if we'd rather see the gate green now; flagging rather than expanding this slice's scope.

**Out-of-diff items** from the toolkit summary (corrupt METADATA.json swallow ~`structure_tool_task.py:858`, stale line-ref) are outside this PR's changes — leaving them for their own change.

muhammad-ali-e:
## SonarCloud — duplication on new code (`executor_rpc.py`, 24.3%): intentional, accepting for this slice

`workers/queue_backend/pg_queue/executor_rpc.py` is the **deliberate workers-twin** of `backend/pg_queue/executor_rpc.py` — two transport adapters (psycopg2 + env here vs Django ORM + Django settings there) sharing the same contract. I looked at extracting the shared logic into one place; the dependency graph makes a clean single-home **not** available without a new package:

- The shared logic needs **both** `unstract.flags` (`check_feature_flag_status`) and `unstract.sdk1` (`ExecutionResult` / `ExecutionDispatcher`).
- The only package both backend and workers share beneath them is `unstract.core` — but **`unstract.sdk1` already depends on `unstract.core`**, so `core` can't import `sdk1` (circular). The dispatcher/result logic can't live there.
- Hosting it in `sdk1` would couple the SDK to Flipt + the PG transport, which **violates the UN-3603 design constraint** (the SDK stays transport/flag-agnostic; gating/routing is an app concern).
- A true single copy therefore needs a **new shared package** (`unstract-executor-rpc`) depended on by both — disproportionate for this slice, and it would also rework the already-merged backend module.

Partial extraction wouldn't clear the 3% gate (it'd need ~all 77 duplicated lines removed), so it's effectively binary: accept the intentional mirror, or build the shared package.

**Decision:** accept the duplication for this PR (it's a reviewed, intentional mirror behind a fail-closed gate, off by default) and track the shared-package consolidation as its own follow-up — same place the `TODO(shared)` in the module points. The SDK-purity constraint is worth more than the metric here.

All other review items (toolkit + greptile) are addressed in `6e3e5c6c2`; 30 workers tests green, pre-commit clean. Good to merge from my side.

PR #2096, opened 2026-06-20
UN-3606 [FIX] PG Queue — prefork the consumer so file batches + executor RPCs run in parallel

muhammad-ali-e:
## Problem

On the PG transport, a multi-file ETL processed files **serially** (the ramp blocker). `worker-pg-fileproc` was a single `batch=1` consumer, so the orchestrator's parallel batches drained one at a time — Celery runs them concurrently via `--pool=prefork --concurrency`. The **same** single-consumer bottleneck on `worker-pg-executor` serialized the `execute_extraction` RPCs (the LLM extraction), so even parallel file fan-out gained nothing end-to-end.

Evidence (2-file ETL): flag OFF (Celery) → both file executions created at the same instant; flag ON (PG, before this fix) → file-2 created only after file-1 **completed**.

## Fix — a prefork supervisor for the PG-queue consumer launcher

The PG analogue of Celery `--pool=prefork --concurrency=N`.

- **`supervisor.py` (new):** when `WORKER_PG_QUEUE_CONSUMER_CONCURRENCY > 1`, fork **N isolated consumer children** (each does its own worker bootstrap, so no connections are inherited across the fork), **monitor + re-fork** dead children, and own a **single fleet-liveness** endpoint (503 when the oldest child stalls past the threshold). `SELECT … FOR UPDATE SKIP LOCKED` distributes batches across children **and** replicas; a single execution is still capped by `MAX_PARALLEL_FILE_BATCHES`; total live parallelism = `concurrency × replicas` (k8s HPA scales replicas).
- **The consumer code is unchanged** — each child is the current single-threaded `PgQueueConsumer.run()`; concurrency is purely a launch concern. `CONCURRENCY=1` (default) keeps the byte-identical single-process path → every other PG consumer is non-regressive.
- **`consumer.py`:** extract `build_consumer_from_env()` / `consumer_env()` so children and `main()` build identical consumers.
- **`docker-compose`:** `CONCURRENCY=4` on **both** `worker-pg-fileproc` and `worker-pg-executor` (mirrors `WORKER_FILE_PROCESSING_CONCURRENCY`). Also raise their `VT_SECONDS` to 3660 — `process_file_batch`/`execute_extraction` block on the executor up to 3600s, so the prior `vt=30` default would re-claim a long batch mid-run (latent **double-run**); health-stale sits just above `vt`.

## Why processes (not threads)

Matches Celery prefork — the cloud-trusted model. Full process isolation (own DB connections + thread-local `StateStore`, no shared-state/thread-safety surface), a crash is contained and re-forked, and its in-flight message redelivers via `vt` (at-least-once). The workload is I/O-bound so the GIL is moot, but cloud runs prefork specifically to avoid thread issues — so the PG consumer does too.

## Tests

`workers/tests/test_pg_consumer_supervisor.py` — env-knob parse/clamp/validation, fleet-staleness calc, and the reap/restart matrix (dead → re-fork, live → left alone, shutdown → not resurrected, already-reaped → handled). The fork/signal/health-port integration is validated live (below) rather than with flaky in-process fork tests.

## Dev-tested end-to-end (gated, running stack)

- 4 `worker-pg-fileproc` + 4 `worker-pg-executor` children fork; fleet health green; killing a child → supervisor re-forks it.
- A 2-file ETL now runs **fully parallel**: both file executions created/completed together, and **both `execute_extraction` RPCs overlap on separate executor children** — ~25s vs ~42s when the executor was serial.

## Out of scope

Continuation/callback path and Celery decommission are later slices; the shared executor-RPC de-dup is UN-3607.

### DIFF
--- a/workers/pg_queue_consumer/supervisor.py
+++ b/workers/pg_queue_consumer/supervisor.py
@@ -121,8 +121,10 @@ class _Fleet:
         return self._n
 
     @property
-    def heartbeats(self):  # noqa: ANN201 — ctypes array, passed to forked children
-        """The shared heartbeat array (children write their own slot directly)."""
+    def heartbeats(self):  # noqa: ANN201
+        """The shared heartbeat array (a ctypes array, passed to forked children,
+        which write their own slot directly).
+        """
         return self._heartbeats
 
     def _validate(self, slot: int) -> None:
@@ -247,9 +249,11 @@ def _child_after_fork(slot: int, heartbeats) -> None:  # noqa: ANN001 (ctypes ar
     signal.signal(signal.SIGINT, signal.SIG_DFL)
     try:
         _run_child(slot, heartbeats)
-    except BaseException:
+    except Exception:
         # A child that can't even start must not return into the supervisor loop
-        # (it would fork grandchildren). Log + hard exit.
+        # (it would fork grandchildren). Log + hard exit. Exception (not
+        # BaseException) is the realistic startup-failure surface here — import,
+        # connection, config; SystemExit/KeyboardInterrupt would exit anyway.
         logger.exception("PG-queue consumer: child slot=%s failed to run", slot)
         os._exit(1)
     os._exit(0)
@@ -264,11 +268,10 @@ def _try_fork_child(fleet: _Fleet, slot: int) -> bool:
     try:
         pid = os.fork()
     except OSError:
-        logger.error(
+        logger.exception(
             "PG-queue consumer: os.fork() failed for slot=%s (process/memory "
             "limit?) — will retry",
             slot,
-            exc_info=True,
         )
         return False
     if pid == 0:  # child — never returns
--- a/workers/pg_queue_consumer/supervisor.py
+++ b/workers/pg_queue_consumer/supervisor.py
@@ -182,8 +182,16 @@ class _Fleet:
     def is_crash_looping(self) -> bool:
         """True if any slot has died immediately ``_CRASH_LOOP_THRESHOLD`` times in
         a row — the signal that the heartbeat alone can't be trusted fresh.
+
+        Snapshots the values first (``tuple(...)`` is atomic under the GIL): this
+        runs in the liveness daemon thread (via :meth:`freshness`) while the main
+        thread mutates ``_consecutive_crashes`` in :meth:`schedule_restart`, so a
+        bare ``.values()`` iteration could raise "dictionary changed size during
+        iteration".
         """
-        return any(n >= _CRASH_LOOP_THRESHOLD for n in self._consecutive_crashes.values())
+        return any(
+            n >= _CRASH_LOOP_THRESHOLD for n in tuple(self._consecutive_crashes.values())
+        )
 
     def oldest_age(self) -> float:
         now = time.time()
--- a/workers/pg_queue_consumer/supervisor.py
+++ b/workers/pg_queue_consumer/supervisor.py
@@ -4,25 +4,32 @@ Forks ``WORKER_PG_QUEUE_CONSUMER_CONCURRENCY`` copies of the single-threaded
 :class:`~queue_backend.pg_queue.consumer.PgQueueConsumer` so multiple file
 batches run in parallel — the PG analogue of Celery's ``--pool=prefork
 --concurrency=N``. ``SELECT … FOR UPDATE SKIP LOCKED`` distributes work across the
-children (and across replicas): each child claims distinct rows, and the cap on a
-single execution is still ``MAX_PARALLEL_FILE_BATCHES`` (producer side). Total live
-parallelism = ``concurrency × replicas``; k8s HPA scales the replica count.
+children (and across replicas): each child claims distinct rows, a single
+execution is still capped by ``MAX_PARALLEL_FILE_BATCHES``, and total live
+parallelism = ``concurrency × replicas`` (k8s HPA scales the replica count).
 
 **Process model** (matches Celery prefork — the cloud-trusted choice): each child
 is a fully isolated process with its own DB connections and thread-local
 ``StateStore`` — no shared mutable state, no thread-safety surface. A child crash
-is isolated and re-forked by the supervisor; its in-flight message redelivers via
+is isolated and re-forked (rate-limited); its in-flight message redelivers via
 ``vt`` (at-least-once). The **consumer code is unchanged** — concurrency is purely
-a launch concern.
+a launch concern. ``CONCURRENCY = 1`` keeps the plain single-process ``main()``
+path (byte-identical to before this module existed).
 
 **Health**: the supervisor owns the single liveness port and reports the *fleet's*
 freshness — the staleness of the oldest-polling child (each child publishes its
 last-poll wall-time into a shared array). A child that dies is re-forked
-internally (transient); only a persistent failure (children that won't stay fresh)
-sustains 503 → pod restart. Children do not bind the port themselves.
+internally (transient); a child that **crash-loops** (dies immediately N times in
+a row, never reaching a real poll) forces the probe to 503 so k8s restarts the
+pod rather than the supervisor masking a wedged fleet with fresh-looking re-forks.
 
-Used only when ``CONCURRENCY > 1``; ``CONCURRENCY = 1`` keeps the plain
-single-process ``main()`` path (byte-identical to before this module existed).
+**Fork safety**: the initial fleet is forked while the parent is single-threaded.
+Re-forks happen after the liveness daemon thread exists; the only other thread is
+that probe (idle in ``select`` between requests, and CPython 3.12 re-inits the
+``logging`` locks across ``fork`` via ``os.register_at_fork``), and each child
+resets inherited signal handlers + does its own ``import worker`` before touching
+shared resources — so an inherited held lock or the parent's ``_on_term`` cannot
+wedge or mis-signal a child.
 """
 
 from __future__ import annotations
@@ -34,6 +41,10 @@ import os
 import signal
 import threading
 import time
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from queue_backend.pg_queue.liveness import LivenessServer
 
 logger = logging.getLogger(__name__)
 
@@ -44,10 +55,15 @@ _MAX_CONCURRENCY = 64
 # How often each child republishes its heartbeat, and the parent reaps + checks.
 _REPORT_INTERVAL_SECONDS = 1.0
 _MONITOR_INTERVAL_SECONDS = 1.0
-# Floor between a child's death and its re-fork — avoids a fork storm when a child
-# dies on startup every time (e.g. a bad migration); the gap lets logs/alerts fire.
+# Re-fork backoff floor and ceiling — a crash-looping child must not fork-storm.
 _RESTART_MIN_INTERVAL_SECONDS = 2.0
-# How long to wait for children to drain on shutdown before giving up the join.
+_RESTART_MAX_BACKOFF_SECONDS = 30.0
+# A child that stays up at least this long before exiting is a normal exit, not an
+# immediate crash — it resets the slot's consecutive-crash counter.
+_MIN_HEALTHY_UPTIME_SECONDS = 10.0
+# Consecutive immediate crashes after which the fleet probe is forced unhealthy.
+_CRASH_LOOP_THRESHOLD = 3
+# How long to wait per child for a graceful drain on shutdown before SIGKILL.
 _SHUTDOWN_GRACE_SECONDS = 30.0
 
 
@@ -77,43 +93,139 @@ def concurrency_from_env() -> int:
     return n
 
 
-def _oldest_child_age(heartbeats) -> float:  # noqa: ANN001 (ctypes array)
-    """Fleet staleness = seconds since the *oldest*-polling child last polled.
-
-    The supervisor's liveness verdict: if the worst child has stalled past the
-    threshold, the fleet probe goes 503. Empty fleet → 0 (fresh; nothing to judge).
+class _Fleet:
+    """Owns the per-slot child state — pid, last-fork, heartbeat, crash count and
+    pending-restart schedule — keeping them mutually consistent. Slots are
+    validated against ``[0, concurrency)`` so a stray key can't silently desync
+    the structures or ``IndexError`` the shared array.
     """
-    now = time.time()
-    return max((now - hb for hb in heartbeats), default=0.0)
+
+    def __init__(self, concurrency: int) -> None:
+        self._n = concurrency
+        # Shared, fork-inherited heartbeat slots (one last-poll wall-time per
+        # child). lock=False is safe: a slot is written either by the parent
+        # (seed, at construction, while no child owns it) OR by that child's
+        # heartbeat thread — never concurrently — and only read by the parent, so
+        # a torn double read just yields one stale sample that self-corrects.
+        self._heartbeats = multiprocessing.Array("d", concurrency, lock=False)
+        now = time.time()
+        for i in range(concurrency):
+            self._heartbeats[i] = now
+        self._pids: dict[int, int] = {}
+        self._last_fork: dict[int, float] = {}
+        self._consecutive_crashes: dict[int, int] = {}
+        self._restart_due: dict[int, float] = {}  # slot -> monotonic not-before
+
+    @property
+    def concurrency(self) -> int:
+        return self._n
+
+    @property
+    def heartbeats(self):  # noqa: ANN201 — ctypes array, passed to forked children
+        """The shared heartbeat array (children write their own slot directly)."""
+        return self._heartbeats
+
+    def _validate(self, slot: int) -> None:
+        if not 0 <= slot < self._n:
+            raise IndexError(f"slot {slot} out of range [0, {self._n})")
+
+    def record_fork(self, slot: int, pid: int) -> None:
+        """Mark ``slot`` alive under ``pid``; clears any pending restart. Note the
+        heartbeat is deliberately NOT reseeded here — a re-forked child must earn
+        freshness by actually polling, so a crash-looping slot ages instead of
+        looking perpetually fresh.
+        """
+        self._validate(slot)
+        self._pids[slot] = pid
+        self._last_fork[slot] = time.monotonic()
+        self._restart_due.pop(slot, None)
+
+    def reap(self, slot: int) -> float:
+        """Drop the slot's pid + last-fork together; return the child's uptime (s)."""
+        forked_at = self._last_fork.pop(slot, time.monotonic())
+        self._pids.pop(slot, None)
+        return time.monotonic() - forked_at
+
+    def schedule_restart(self, slot: int, uptime: float) -> int:
+        """Record the exit and set the re-fork not-before; return the consecutive
+        immediate-crash count. A child that ran healthily before exiting resets the
+        counter; an immediate death increments it and backs the restart off
+        (capped) so a crash loop can't fork-storm.
+        """
+        if uptime < _MIN_HEALTHY_UPTIME_SECONDS:
+            n = self._consecutive_crashes.get(slot, 0) + 1
+        else:
+            n = 0  # ran fine, then exited — not a crash loop
+        self._consecutive_crashes[slot] = n
+        backoff = min(
+            _RESTART_MIN_INTERVAL_SECONDS * max(1, n), _RESTART_MAX_BACKOFF_SECONDS
+        )
+        self._restart_due[slot] = time.monotonic() + backoff
+        return n
+
+    def due_restarts(self) -> list[int]:
+        """Slots whose re-fork backoff has elapsed (oldest schedule first)."""
+        now = time.monotonic()
+        return sorted(s for s, due in self._restart_due.items() if due <= now)
+
+    def consecutive_crashes(self, slot: int) -> int:
+        return self._consecutive_crashes.get(slot, 0)
+
+    def alive_items(self) -> list[tuple[int, int]]:
+        return list(self._pids.items())
+
+    def alive_count(self) -> int:
+        return len(self._pids)
+
+    def is_crash_looping(self) -> bool:
+        """True if any slot has died immediately ``_CRASH_LOOP_THRESHOLD`` times in
+        a row — the signal that the heartbeat alone can't be trusted fresh.
+        """
+        return any(n >= _CRASH_LOOP_THRESHOLD for n in self._consecutive_crashes.values())
+
+    def oldest_age(self) -> float:
+        now = time.time()
+        return max((now - hb for hb in self._heartbeats), default=0.0)
+
+    def freshness(self) -> float:
+        """Liveness verdict source: a crash-looping fleet is force-stale (``inf``)
+        so the probe trips 503 even if a just-constructed child briefly looked
+        fresh; otherwise the oldest child's staleness (catches a wedged-alive
+        child).
+        """
+        return float("inf") if self.is_crash_looping() else self.oldest_age()
 
 
 def _run_child(slot: int, heartbeats) -> None:  # noqa: ANN001 (ctypes array)
-    """Child entry: bootstrap the worker app, then run one consumer forever.
+    """Build one consumer and run it forever, publishing its heartbeat.
 
     The worker import (and any connections it opens) happens HERE, in the child —
     never inherited across the fork — so each process owns its own connections.
-    A daemon thread publishes the consumer's last-poll wall-time into
+    A *guarded* daemon thread publishes the consumer's last-poll wall-time into
     ``heartbeats[slot]`` for the supervisor's fleet liveness.
     """
-    # Same source-worker selection the single-process launcher does (see
-    # pg_queue_consumer/__main__.py): overwrite WORKER_TYPE before importing
-    # ``worker``, which reads it at import time to register the right tasks.
-    os.environ["WORKER_TYPE"] = os.environ.get(
-        "WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE", "notification"
-    )
+    from pg_queue_consumer._bootstrap import select_source_worker_type
+
+    select_source_worker_type()  # set WORKER_TYPE before importing worker
     import worker  # noqa: F401 — side-effect: registers the source worker's tasks
     from queue_backend.pg_queue.consumer import build_consumer_from_env
 
     consumer = build_consumer_from_env()
-    heartbeats[slot] = time.time()  # seed fresh before the first poll
 
     def _publish_heartbeat() -> None:
         # last-poll wall-time = now − (seconds since last poll). Frozen while a
         # task runs (the consumer stamps its heartbeat at the top of poll_once),
         # so a child stuck on a too-long task goes stale exactly as the single
-        # consumer does — the supervisor then sees it via the shared slot.
+        # consumer does. Guarded so a transient error (e.g. teardown during
+        # shutdown) logs loudly and the loop continues instead of dying silently
+        # and false-staling a healthy child.
         while True:
-            heartbeats[slot] = time.time() - consumer.seconds_since_last_poll()
+            try:
+                heartbeats[slot] = time.time() - consumer.seconds_since_last_poll()
+            except Exception:
+                logger.exception(
+                    "PG-queue consumer: heartbeat publish failed for slot=%s", slot
+                )
             time.sleep(_REPORT_INTERVAL_SECONDS)
 
     threading.Thread(target=_publish_heartbeat, daemon=True, name=f"pg-hb-{slot}").start()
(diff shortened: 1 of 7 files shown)


workers/pg_queue_consumer/supervisor.py
```diff
@@ -0,0 +1,306 @@
...
+    now = time.time()
+    for i in range(concurrency):
+        heartbeats[i] = now
+
+    children: dict[int, int] = {}  # slot -> pid
+    last_fork: dict[int, float] = {}  # slot -> monotonic time of last fork
+    stopping = threading.Event()
+
+    def _fork_child(slot: int) -> None:
+        heartbeats[slot] = time.time()  # reset freshness for the new child
```
muhammad-ali-e: **🔴 Critical — crash-looping child stays *fake-fresh*, hiding a wedged fleet from the probe.**

`_fork_child` reseeds `heartbeats[slot] = time.time()` on *every* (re)fork. If `_run_child` dies before steady state (bad migration, DB unreachable at connect, import error), the child `os._exit(1)`s, `_reap_and_restart` re-forks it ~every 2s (`_RESTART_MIN_INTERVAL_SECONDS`), and each re-fork refreshes the slot — so `_oldest_child_age` reads **fresh** off a child that never polls. The pod keeps serving 200 on `/health` while that slot (or, if all N slots crash-loop, the entire consumer) drains zero work. The only signal is a `logger.warning` every 2s.

**Fix:** don't reseed `heartbeats[slot]` on re-fork until the child has published a real poll — let a never-polling slot age into 503 so k8s restarts the pod. Optionally add a consecutive-immediate-exit counter that escalates to `error` + backoff after a threshold.

_(Silent-Failure Hunter; corroborated by Code Reviewer.)_

workers/pg_queue_consumer/supervisor.py
```diff
@@ -0,0 +1,306 @@
...
+        pid = os.fork()
```
muhammad-ali-e: **🟠 High — unhandled `os.fork()` OSError can crash the whole supervisor / fleet.**

`os.fork()` raises `OSError` on EAGAIN (RLIMIT_NPROC / thread cap) or ENOMEM — realistic precisely under the many-heavy-children load this PR targets. It is uncaught:
- During the initial fleet loop (line 178) it propagates out of `run_supervised` before the health server and signal cleanup exist → raw traceback, pod dies.
- During `_reap_and_restart` (line 227) it propagates into `run_supervised`'s `finally` (line 195), which SIGTERMs **every other healthy child** — one transient EAGAIN takes out all N working consumers.

**Fix:** wrap `pid = os.fork()` in `try/except OSError`. On reap-path failure, log `error` (slot, errno) and leave the slot for the next monitor tick. On initial-fork failure, fail fast with an actionable message ("reduce CONCURRENCY or raise nproc/memory") rather than an uncaught crash.

_(Silent-Failure Hunter.)_

workers/pg_queue_consumer/supervisor.py
```diff
@@ -0,0 +1,306 @@
...
+
+    consumer = build_consumer_from_env()
+    heartbeats[slot] = time.time()  # seed fresh before the first poll
+
+    def _publish_heartbeat() -> None:
+        # last-poll wall-time = now − (seconds since last poll). Frozen while a
+        # task runs (the consumer stamps its heartbeat at the top of poll_once),
+        # so a child stuck on a too-long task goes stale exactly as the single
+        # consumer does — the supervisor then sees it via the shared slot.
+        while True:
```
muhammad-ali-e, line 232: **🟠 High — the heartbeat thread can die silently and false-stale a healthy child.**

The `while True` loop has no exception guard. If `consumer.seconds_since_last_poll()` or the array write ever raises (e.g. teardown during shutdown), this daemon thread dies with no log anywhere. `heartbeats[slot]` then freezes at its last value, `_oldest_child_age` grows unbounded, the fleet probe trips 503, and k8s kills a pod whose child is actually healthy and draining — with the root cause (a dead heartbeat thread) completely invisible. Contrast `LivenessServer._serve`, which *does* log on thread death.

Note the comment above reasons only about the *frozen-during-task* case, not the thread *dying* (same frozen slot, but permanent).

**Fix:** wrap the loop body in `try/except Exception: logger.exception(...)` so a transient error logs and the loop continues, and log loudly (with the slot) if the loop ever exits.

_(Silent-Failure Hunter; corroborated by Code Reviewer.)_

workers/pg_queue_consumer/supervisor.py
```diff
@@ -0,0 +1,306 @@
...
+
+    def _fleet_freshness() -> float:
+        # Fleet age = the oldest child's staleness; if any child stalls past the
+        # threshold the probe goes 503. A just-(re)forked child is seeded fresh.
+        return _oldest_child_age(heartbeats)
+
+    def _extra_status() -> dict[str, object]:
+        return {"alive_children": len(children), "concurrency": concurrency}
+
+    health = _maybe_start_supervisor_health(_fleet_freshness, _extra_status)
```
muhammad-ali-e: **🟡 Medium — re-forking after the liveness thread starts is a fork+threads hazard.**

The *initial* fleet is forked while single-threaded (correct — line 177 comment). But the health server thread starts here (line 189) **before** the supervise loop, and re-forks in `_reap_and_restart` (line 227) then happen with that thread running. `os.fork()` copies only the calling thread; if the liveness thread holds the `logging` module lock or an `http.server`/`socket` lock at the instant of fork, the child inherits a permanently-locked mutex and can deadlock the moment it calls `logger.*` or `import worker` — a silent stuck child that never polls.

Low probability (the liveness thread is idle between probes) but a real hazard.

**Fix:** `os.register_at_fork()` to guard the logging lock around fork, or avoid logging in the child between `os.fork()` and re-exec, or at minimum document the constraint.

_(Code Reviewer.)_

workers/pg_queue_consumer/supervisor.py
```diff
@@ -0,0 +1,306 @@
...
+            reaped, _status = os.waitpid(pid, os.WNOHANG)
+        except ChildProcessError:
+            reaped = pid  # already reaped elsewhere — treat as gone
+        if reaped == 0:
+            continue  # still alive
+        del children[slot]
+        if stopping.is_set():
+            continue
+        elapsed = time.monotonic() - last_fork.get(slot, 0.0)
+        if elapsed < _RESTART_MIN_INTERVAL_SECONDS:
```
muhammad-ali-e: **🟡 Medium — the rate-limit `sleep` serializes recovery and there's no `stopping` recheck after it.**

Two issues in this branch:
1. `time.sleep(...)` (up to 2s) runs **synchronously inside** the `for slot, pid in list(children.items())` loop. If several children die in one tick their backoffs run sequentially (N×2s), and during any sleep the loop can't reap other children or react to `stopping`. For a 4-wide fleet crash-looping on a bad migration this serializes restarts and delays SIGTERM handling.
2. If SIGTERM arrives during the sleep, `_on_term` sets `stopping` + broadcasts SIGTERM to the *current* children, then the sleep finishes and `fork_child(slot)` runs **unconditionally** (line 227) — no `if stopping.is_set(): return` recheck. The new child is spawned only to receive SIGTERM mid-`import worker`, before `consumer.run()` installs handlers, so it dies on the default disposition instead of draining.

**Fix:** replace the inline sleep with a per-slot "earliest next fork" monotonic timestamp and `continue` slots that aren't eligible yet (the `_MONITOR_INTERVAL_SECONDS` outer sleep already paces the loop); and re-check `if stopping.is_set(): return` immediately after the wait, before forking.

_(Code Reviewer.)_

workers/pg_queue_consumer/supervisor.py
```diff
@@ -0,0 +1,306 @@
...
+        port=port,
+        check_name="pg_queue_fleet",
+        age_key="oldest_child_seconds_since_poll",
+        extra_status_fn=extra_status_fn,
+        thread_name="pg-supervisor-liveness",
+        log_label="pg-queue supervisor",
+    )
+    try:
+        server.start()
+    except OSError:
```
muhammad-ali-e: **🟡 Medium — liveness bind failure silently runs the consumer with no probe.**

If `LivenessServer.start()` raises `OSError` (e.g. `EADDRINUSE` from a port misconfig or a leftover socket), this logs and returns `None`, and the supervisor runs on with **no health probe** — silently removing the one mechanism that catches a wedged poll loop. For a transient race that's defensible, but `EADDRINUSE` usually signals a real config bug worth failing fast on.

**Fix:** distinguish `EADDRINUSE` (likely fail-fast) from other transient bind errors; when continuing probe-less, log at `error` (not just `exception`) and surface it in `_extra_status` / a metric (`"liveness_probe_bound": false`) so the degradation is observable even when the probe can't answer.

_(Silent-Failure Hunter.)_

workers/pg_queue_consumer/supervisor.py
```diff
@@ -0,0 +1,306 @@
...
+
+    # Shared, fork-inherited heartbeat slots (one wall-time per child). No lock:
+    # each slot has a single writer (its child) and one reader (the parent); a
+    # torn double read just yields one stale sample, self-correcting next tick.
+    heartbeats = multiprocessing.Array("d", concurrency, lock=False)
+    now = time.time()
+    for i in range(concurrency):
+        heartbeats[i] = now
+
+    children: dict[int, int] = {}  # slot -> pid
```
muhammad-ali-e: **🟡 Medium (type design) — three parallel slot-keyed structures with unenforced consistency.**

`children` (slot→pid), `last_fork` (slot→monotonic), and `heartbeats` (fixed-length array) are all keyed by the same `slot` and must stay mutually consistent, but nothing enforces it:
- `slot ∈ [0, concurrency)` is unchecked — `heartbeats[slot]` would `IndexError` at line 145 while the dicts accept any key.
- `_reap_and_restart` deletes only from `children` (line 218), leaving a stale `last_fork[slot]` — so the two dicts' key sets drift and neither is a trustworthy "is this slot alive?" source.

**Fix (proportionate):** a small `Fleet` class owning the three structures with intent-named methods (`record_fork`, `reap`, `seed_fresh`, `publish`, `oldest_age`, `alive_count`), validating `0 <= slot < concurrency` once and clearing both dicts together in `reap()`. ~30 lines; it also encapsulates the raw ctypes array so it never crosses a public signature (resolving the `ANN001` noqas).

_(Type-Design Analyzer.)_

workers/pg_queue_consumer/supervisor.py
```diff
@@ -0,0 +1,306 @@
...
+        if health is not None:
+            health.stop()
+        logger.info("PG-queue consumer supervisor: stopped")
+
+
+def _reap_and_restart(
+    children: dict[int, int],
+    last_fork: dict[int, float],
+    stopping: threading.Event,
+    fork_child,  # noqa: ANN001 (Callable[[int], None])
```
muhammad-ali-e: **🟢 Low (type) — `# noqa: ANN001` on callables should be real `Callable` annotations.**

The noqa comment already states the exact type — write it: `fork_child: Callable[[int], None]`. Same at `_maybe_start_supervisor_health` (line 261): `freshness_fn: Callable[[], float]`, `extra_status_fn: Callable[[], dict[str, Any]]`, and `-> LivenessServer | None` (import under `TYPE_CHECKING`). `LivenessServer.__init__` already declares these signatures, so typing them restores checker coverage into the one call where it pays off. `from collections.abc import Callable` — ~4 lines, no behavior change, deletes three noqas.

_(Type-Design Analyzer.)_

workers/pg_queue_consumer/supervisor.py
```diff
@@ -0,0 +1,306 @@
...
+
+def run_supervised(concurrency: int) -> None:
+    """Fork ``concurrency`` consumer children and supervise them until SIGTERM.
+
+    Forks the initial fleet while the parent is still single-threaded (safe), then
+    starts the liveness server and the reap/restart loop.
+    """
+    logging.basicConfig(level=os.getenv("LOG_LEVEL", "INFO"))
+
+    # Shared, fork-inherited heartbeat slots (one wall-time per child). No lock:
```
muhammad-ali-e: **🟢 Low (comment accuracy) — "single writer" is false as stated.**

The slot does *not* have a single writer: the **parent** also writes `heartbeats[slot]` at line 138 (seed) and line 145 (`_fork_child` reset on every re-fork), in addition to the child's heartbeat thread (lines 108, 116). The lock-free safety still holds, but for a different reason — the parent only writes a slot while *no child owns it* (the old child is reaped before `_fork_child` runs, the new one doesn't exist yet), so writes are never concurrent. A future maintainer trusting "single writer" might add a third writer assuming an invariant that isn't enforced.

**Fix:** justify safety by non-concurrency ("written by the parent on seed/re-fork and by the child's thread, but never concurrently — the parent only writes a slot while no child owns it") rather than by a false single-writer claim.

_(Comment Analyzer.)_

workers/pg_queue_consumer/supervisor.py
```diff
@@ -0,0 +1,306 @@
...
+            with contextlib.suppress(ChildProcessError):
+                os.waitpid(pid, 0)
+
+
+def _maybe_start_supervisor_health(freshness_fn, extra_status_fn):  # noqa: ANN001,ANN201
+    """Start the fleet liveness server when a port is configured; else None.
+
+    Reuses the same env knobs as the single-process consumer
+    (``WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT`` / ``_HEALTH_STALE_SECONDS``) so the
+    pod's probe is unchanged — only the freshness source differs (fleet vs one loop).
```
muhammad-ali-e: **🟢 Low (comment accuracy) — "the pod's probe is unchanged" overstates; the JSON body differs.**

The HTTP contract (`/health`, 200/503) and env knobs are unchanged, but the JSON response body is not: the single-process path uses `check_name="pg_queue_poll"` / `age_key="seconds_since_last_poll"`, while the supervisor passes `check_name="pg_queue_fleet"` / `age_key="oldest_child_seconds_since_poll"` (lines 285-286). Any dashboard/alert parsing those keys sees different fields under `concurrency > 1`.

**Fix:** "...same env knobs and the same HTTP contract (`/health` → 200/503), so the k8s probe config is unchanged. The JSON body differs (`check="pg_queue_fleet"`, age key `oldest_child_seconds_since_poll`) since the freshness source is the fleet's oldest child rather than one poll loop."

_(Comment Analyzer.)_

workers/pg_queue_consumer/supervisor.py
```diff
@@ -0,0 +1,306 @@
...
+            "PG-queue consumer: child slot=%s pid=%s exited — restarting", slot, pid
+        )
+        fork_child(slot)
+
+
+def _join_children(children: dict[int, int], grace_seconds: float) -> None:
+    """Wait up to ``grace_seconds`` for children to exit; SIGKILL stragglers."""
+    deadline = time.monotonic() + grace_seconds
+    for slot, pid in list(children.items()):
+        remaining = deadline - time.monotonic()
```
muhammad-ali-e: **🟢 Low (simplification) — `_join_children`'s `waited` flag + `remaining` math collapses into a helper.**

The `waited` boolean is a hand-rolled "did the inner loop find the child gone?" return value, and `remaining` is recomputed in two places only to feed `while remaining > 0`. Extracting the poll loop makes both disappear and is independently testable:

```python
def _wait_for_exit(pid: int, deadline: float) -> bool:
    """Poll pid until it exits or deadline passes. True if it exited."""
    while time.monotonic() < deadline:
        try:
            reaped, _status = os.waitpid(pid, os.WNOHANG)
        except ChildProcessError:
            return True
        if reaped != 0:
            return True
        time.sleep(0.1)
    return False
```

Then the body becomes `if not _wait_for_exit(pid, deadline): <SIGKILL path>`. Same syscalls, same cadence — behavior-preserving.

_(Code Simplifier.)_

workers/pg_queue_consumer/supervisor.py
```diff
@@ -0,0 +1,306 @@
...
+
+    The worker import (and any connections it opens) happens HERE, in the child —
+    never inherited across the fork — so each process owns its own connections.
+    A daemon thread publishes the consumer's last-poll wall-time into
+    ``heartbeats[slot]`` for the supervisor's fleet liveness.
+    """
+    # Same source-worker selection the single-process launcher does (see
+    # pg_queue_consumer/__main__.py): overwrite WORKER_TYPE before importing
+    # ``worker``, which reads it at import time to register the right tasks.
+    os.environ["WORKER_TYPE"] = os.environ.get(
```
muhammad-ali-e: **🟢 Low (simplification / DRY) — `WORKER_TYPE` selection is duplicated with `__main__.py:49`.**

This exact mutation (with the same multi-line rationale comment) appears in both `_run_child` here and the single-process path in `__main__.py:49-51`. It's a behaviorally-coupled constant: if the `"notification"` default or the env-var name changes, both must change in lockstep, and a drift silently routes to the wrong worker.

**Fix (optional):** extract `select_source_worker_type()` into a small neutral module (e.g. `pg_queue_consumer/_bootstrap.py` — not `__main__`, to avoid importing it from `supervisor`) and call it from both sites immediately before `import worker`. Pure code motion; the default lives in one place. Lower priority — the authors kept the two entry paths deliberately independent.

_(Code Simplifier.)_

muhammad-ali-e (commented):
## Automated PR Review — UN-3606 PG-consumer prefork concurrency

Reviewed with the PR Review Toolkit (Code Reviewer, Silent-Failure Hunter, Type-Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier). This is a carefully-written, well-commented PR — the heartbeat clock math (`time.time() - seconds_since_last_poll()`) is correct, the lockless `multiprocessing.Array` and `os._exit` fork-bomb guard are sound, and the docker-compose `vt < health-stale` ordering is consistent.

The findings below are inline. Priority summary:

**🔴 Critical** — crash-looping child stays *fake-fresh* to the probe (supervisor.py:145), so a fully-wedged fleet can report HTTP 200.

**🟠 High** — unhandled `os.fork()` OSError can take down the whole fleet (supervisor.py:146); the per-child heartbeat thread can die silently and false-stale a healthy child (supervisor.py:115); the rate-limit backstop branch is entirely untested (test file:77).

**🟡 Medium** — re-fork after the liveness thread starts is a fork+threads hazard (supervisor.py:189); the rate-limit `sleep` serializes recovery and skips a `stopping` recheck (supervisor.py:222); liveness bind failure silently runs probe-less (supervisor.py:293); three parallel slot-keyed structures lack an owning type (supervisor.py:140).

**🟢 Low** — type annotations, two comment inaccuracies, and two safe simplifications.

None are merge-blockers on their own, but I'd address the Critical + two High items before this ships to the fleet.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -427,36 +427,50 @@ def _parse_queue_list(raw: str) -> list[str]:
...
+
+
+def main() -> None:
+    logging.basicConfig(level=os.getenv("LOG_LEVEL", "INFO"))
+    consumer = build_consumer_from_env()
     health_server = _maybe_start_health_server(
         consumer,
-        port=_env("HEALTH_PORT", None, int),
-        stale_after=_env("HEALTH_STALE_SECONDS", _DEFAULT_HEALTH_STALE_SECONDS, float),
+        port=consumer_env("HEALTH_PORT", None, int),
```
muhammad-ali-e: **🟢 Low (type) — the `HEALTH_PORT` call bends the `consumer_env` generic.**

`consumer_env` is correctly generic, but here `default=None` infers `_T = NoneType` while `cast=int` requires `_T = int`; the checker only tolerates it by widening to `int | None`, so the intent ("an `int` port, or `None` when unset") is recovered rather than declared. Runtime behavior is correct.

**Fix (optional, cosmetic):** annotate the result `port: int | None = consumer_env("HEALTH_PORT", None, int)` (same at the supervisor's call), or add a nullable-default overload. Don't special-case `None` inside `consumer_env` — that would complicate a clean generic for one call shape.

_(Type-Design Analyzer.)_

workers/tests/test_pg_consumer_supervisor.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def test_empty_fleet_is_zero(self):
+        assert _oldest_child_age([]) == 0.0
+
+
+class TestReapAndRestart:
+    """The supervisor's recovery loop: dead children re-fork; the rest are left."""
+
+    @staticmethod
+    def _old(slot: int) -> dict:
```
muhammad-ali-e: **🟠 High (test gap) — the rate-limit backstop branch is never exercised.**

Every `_reap_and_restart` test uses `_old(slot)` (last_fork 1000s ago) specifically to *skip* the `elapsed < _RESTART_MIN_INTERVAL_SECONDS` sleep at `supervisor.py:222-223`. That branch is the entire reason `last_fork` exists — the fork-storm guard the module docstring warns about. If it silently breaks (inverted comparison, dropped sleep), a child crashing on startup produces an unthrottled fork storm and nothing fails.

**Add:** patch `os.waitpid` → `(111, 0)` (dead), set `last_fork = {0: time.monotonic()}` (just forked), patch `{_MOD}.time.sleep` with a mock, and assert it was called once with a value in `(0, _RESTART_MIN_INTERVAL_SECONDS]` and that `fork_child` still ran. Plus a `last_fork` missing the slot → no sleep.

_(Test Analyzer — highest-value missing test.)_

workers/tests/test_pg_consumer_supervisor.py
```diff
@@ -0,0 +1,116 @@
...
+        now = time.time()
+        # One child polled 100s ago — that's the fleet's staleness.
+        age = _oldest_child_age([now, now - 100, now - 5])
+        assert 99 < age < 102
+
+    def test_empty_fleet_is_zero(self):
+        assert _oldest_child_age([]) == 0.0
+
+
+class TestReapAndRestart:
```
muhammad-ali-e: **🟡 Medium (test gaps) — three fork-free-testable code paths have zero coverage.**

Beyond the rate-limit branch, these are all mockable without real forks/signals and run on common paths:
- **`_join_children` SIGKILL escalation** (`supervisor.py:230-258`) — runs on *every* shutdown. Mock `os.waitpid`/`os.kill`/`time`: (a) clean exit within grace → `os.kill` not called; (b) never exits, drive `time.monotonic` past deadline → `SIGKILL` + final blocking reap; (c) `ChildProcessError` → treated as gone.
- **`_fork_child` `os._exit` guard** (`:147-155`) — the fork-bomb guardrail. Patch `os.fork`→0, `_run_child`→raises, `os._exit`→sentinel; assert `os._exit(1)` on failure and `(0)` on the happy path.
- **`_maybe_start_supervisor_health` OSError fallback** (`:291-299`) — patch `LivenessServer.start` to raise `OSError`; assert it returns `None` and doesn't propagate. Also `port is None` → `None`.

Minor edge cases in existing tests: boundary at exactly `_MAX_CONCURRENCY` (`>` vs `>=`), `"4.0"` float-string input, and a mixed multi-slot fleet in `_reap_and_restart` (dead + alive + already-reaped in one pass).

_(Test Analyzer.)_

workers/pg_queue_consumer/supervisor.py
muhammad-ali-e: Fixed in 46109f1d9 — re-fork no longer reseeds the heartbeat (`_Fleet.record_fork` leaves the slot to age), AND a slot that dies immediately `_CRASH_LOOP_THRESHOLD` times in a row forces `_Fleet.freshness()` to `inf` → 503. So a crash-looping (or never-polling) child can't masquerade as fresh; k8s restarts the pod. A child that ran ≥ `_MIN_HEALTHY_UPTIME_SECONDS` resets the counter (escalated to `error` logging meanwhile).

workers/pg_queue_consumer/supervisor.py
muhammad-ali-e: Fixed in 46109f1d9 — `_try_fork_child` wraps `os.fork()` in try/except OSError: returns False (logged) instead of raising. Initial fleet → fail fast with an actionable message (reduce CONCURRENCY / raise nproc-memory); reap path → the slot is left for the next monitor tick, never SIGTERMing the healthy children.

workers/pg_queue_consumer/supervisor.py
muhammad-ali-e: Fixed in 46109f1d9 — the heartbeat publish loop body is now `try/except Exception: logger.exception(...)`, so a transient error logs loudly and the loop continues instead of dying silently and false-staling a healthy child.

workers/pg_queue_consumer/supervisor.py
muhammad-ali-e: Addressed in 46109f1d9 — each child now resets SIGTERM/SIGINT to SIG_DFL as the first thing after fork (see the greptile-P2 thread), and the docstring documents the constraint: the only co-thread is the liveness probe (idle in `select`), and CPython 3.12 re-inits the `logging` locks across `fork` via `os.register_at_fork`. So an inherited held lock can't wedge a child in practice.

workers/pg_queue_consumer/supervisor.py
muhammad-ali-e: Fixed in 46109f1d9 — restored as a non-blocking schedule: `schedule_restart` sets a per-slot not-before (backoff grows with consecutive crashes, capped), and `_restart_due_children` re-forks only due slots, re-checking `stopping` each iteration. No in-loop `sleep` (recovery isn't serialised), and no child spawned into shutdown.

workers/pg_queue_consumer/supervisor.py
muhammad-ali-e: Fixed in 46109f1d9 — `EADDRINUSE` now logs at `error` (vs `warning` for transient binds), and `liveness_probe_bound: false` is surfaced in the status payload so the probe-less degradation is observable.

workers/pg_queue_consumer/supervisor.py
muhammad-ali-e: Done in 46109f1d9 — added a `_Fleet` class owning pid / last_fork / heartbeat / crash-count / restart-schedule, with `_validate(slot)` against `[0, concurrency)` and a `reap()` that clears the structures together. The loose parallel dicts are gone.

workers/pg_queue_consumer/supervisor.py
muhammad-ali-e: Fixed in 46109f1d9 — real `Callable[...]`/`LivenessServer | None` annotations (the latter under TYPE_CHECKING); the noqas are deleted. `_Fleet.heartbeats` keeps one noqa (the ctypes array type isn't worth importing).

workers/pg_queue_consumer/supervisor.py
muhammad-ali-e: Fixed in 46109f1d9 — the comment now justifies the lock-free safety by non-concurrency (the parent writes a slot only at seed, while no child owns it; the owning child's thread writes thereafter — never concurrently), not a false single-writer claim.

workers/pg_queue_consumer/supervisor.py
muhammad-ali-e: Fixed in 46109f1d9 — reworded: same env knobs + same HTTP contract (`/health` → 200/503, probe config unchanged), but the JSON body differs (`check="pg_queue_fleet"`, age key `oldest_child_seconds_since_poll`) since the source is the fleet's oldest child.

workers/pg_queue_consumer/supervisor.py
muhammad-ali-e: Done in 46109f1d9 — extracted `_wait_for_exit(pid, deadline) -> bool`; `_join_children` is now `if not _wait_for_exit(...): <SIGKILL>`. The `waited`/`remaining` bookkeeping is gone and the helper is unit-tested.

workers/pg_queue_consumer/supervisor.py
muhammad-ali-e: Done in 46109f1d9 — extracted `select_source_worker_type()` into a neutral `pg_queue_consumer/_bootstrap.py`, called from both `__main__` and the supervisor's `_run_child`. The default lives in one place now.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed in 46109f1d9 — annotated `port: int | None = consumer_env("HEALTH_PORT", None, int)` at both call sites; the generic is left clean.

workers/tests/test_pg_consumer_supervisor.py
muhammad-ali-e: Fixed in 46109f1d9 — the backoff is now a scheduled not-before, so it's covered by `TestFleet.test_due_restarts_respects_backoff` (monkeypatched clock: not due immediately, due after the backoff) plus `test_immediate_crash_increments_then_loops` for the consecutive-crash counter. The old in-loop sleep that needed mocking is gone.

workers/tests/test_pg_consumer_supervisor.py
muhammad-ali-e: Done in 46109f1d9 — added coverage: `_join_children` SIGKILL escalation (clean-exit → no kill; straggler → SIGKILL + reap) via `_wait_for_exit`, `_try_fork_child` OSError → False + child-branch hard-exit guard (`_child_after_fork` → `os._exit(1)` on raise, `(0)` on clean run), and `_maybe_start_supervisor_health` no-port → None + bind-error swallowed. 32 tests total.

workers/pg_queue_consumer/supervisor.py
```diff
@@ -0,0 +1,306 @@
...
+            reaped = pid  # already reaped elsewhere — treat as gone
+        if reaped == 0:
+            continue  # still alive
+        del children[slot]
+        if stopping.is_set():
+            continue
+        elapsed = time.monotonic() - last_fork.get(slot, 0.0)
+        if elapsed < _RESTART_MIN_INTERVAL_SECONDS:
+            time.sleep(_RESTART_MIN_INTERVAL_SECONDS - elapsed)
+        logger.warning(
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 46109f1d9 — see the rate-limit-sleep thread: re-forks are a non-blocking scheduled not-before now, and `_restart_due_children` re-checks `stopping` before each fork, so a SIGTERM during the (now-absent) wait can't spawn a fresh child into shutdown. Added `TestRestartDueChildren.test_stopping_blocks_refork`.

workers/pg_queue_consumer/supervisor.py
```diff
@@ -0,0 +1,306 @@
...
+                "PG-queue consumer: child slot=%s pid=%s did not stop in %ss — "
+                "SIGKILL",
+                slot,
+                pid,
+                grace_seconds,
+            )
+            with contextlib.suppress(ProcessLookupError):
+                os.kill(pid, signal.SIGKILL)
+            with contextlib.suppress(ChildProcessError):
+                os.waitpid(pid, 0)
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 46109f1d9 — `_join_children` now gives each child its OWN `grace_seconds` budget (`_wait_for_exit(pid, monotonic()+grace)` per child) instead of a single shared deadline, so a slow-draining child can't starve the rest of their grace.

workers/pg_queue_consumer/supervisor.py
```diff
@@ -0,0 +1,306 @@
...
+        # task runs (the consumer stamps its heartbeat at the top of poll_once),
+        # so a child stuck on a too-long task goes stale exactly as the single
+        # consumer does — the supervisor then sees it via the shared slot.
+        while True:
+            heartbeats[slot] = time.time() - consumer.seconds_since_last_poll()
+            time.sleep(_REPORT_INTERVAL_SECONDS)
+
+    threading.Thread(target=_publish_heartbeat, daemon=True, name=f"pg-hb-{slot}").start()
+    # consumer.run() installs its own SIGTERM/SIGINT handlers → graceful stop.
+    consumer.run()
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in 46109f1d9 — `_child_after_fork` resets SIGTERM/SIGINT to `SIG_DFL` as its first action (before `import worker`), so a signal in the fork→run window no longer fires the parent's `_on_term` against the stale `children` snapshot. SIG_DFL = terminate, the right disposition for a not-yet-running child.

muhammad-ali-e:
## Review addressed — `46109f1d9` + `7e28f2e50`

Thorough review, thank you — the supervisor got a solid hardening pass. All toolkit + greptile + SonarCloud items addressed; **32 tests, pre-commit clean, re-validated live (4 children, fleet health with `crash_looping`/`liveness_probe_bound`, crash → re-fork).**

**Critical / High**
- **Crash-loop no longer hides a wedged fleet:** re-fork doesn't reseed the heartbeat (a never-polling slot ages), and N immediate deaths in a row force `freshness=inf` → 503.
- **`os.fork()` guarded:** initial-fleet failure fails fast (actionable); reap-path failure logs + retries — no uncaught crash, no SIGTERM storm against healthy children.
- **Heartbeat thread guarded** (try/except + log) so it can't die silently and false-stale a healthy child.
- **Children reset signals to `SIG_DFL`** right after fork → a SIGTERM in the fork→run window can't fire the parent's `_on_term` against stale pids (greptile P2).

**Medium**
- New **`_Fleet`** class owns pid/last_fork/heartbeat/crash-count/restart-schedule with a validated slot + consistent reap (replaces three loose dicts).
- Re-fork backoff is a **non-blocking scheduled not-before**, re-checking `stopping` before each fork (greptile P1) — no in-loop sleep, no child spawned into shutdown.
- Liveness: `EADDRINUSE` → error; `liveness_probe_bound` surfaced in the payload.
- `_join_children`: **per-child** grace budget (greptile P2) + extracted `_wait_for_exit`.

**Low** — real `Callable`/`int | None` annotations, comment fixes (heartbeat writer reasoning, fleet-probe JSON body), `WORKER_TYPE` deduped into `_bootstrap.py`.

**Tests** — rewritten + expanded to the gaps called out: `_Fleet` bookkeeping/crash-loop/freshness, reap scheduling, restart-due gating, fork OSError + child hard-exit guard, `_wait_for_exit` + `_join_children` SIGKILL escalation, health no-port/bind-error.

**SonarCloud** — cleared the New-Code reliability bug (float equality → `math.isinf`, Rating C → A) and the maintainability nits (`# noqa` syntax, `except Exception` over `BaseException`, `logging.exception()`).

Good to merge from my side once the bots re-run.

muhammad-ali-e:
Addressed the 4/5 finding in 5dce34e12: `_Fleet.is_crash_looping()` (called from the liveness daemon thread via `freshness()`) now snapshots `tuple(self._consecutive_crashes.values())` before iterating — atomic under the GIL, so it can't race the main thread's `schedule_restart` write ('dictionary changed size during iteration'). 32 tests, lint clean.

PR #2097, opened 2026-06-20
UN-3608 [FEAT] PG Queue 9i — executor async/callback path on PG (self-chained continuations)

muhammad-ali-e:
## What

Migrates the executor RPC's **async/callback** path off Celery onto the PG queue, completing the executor-transport migration (the blocking path landed in 9h-c). Targets the `feat/UN-3445-pg-queue-integration` branch (not `main`).

Today `dispatch_with_callback` (Prompt Studio run/index/extract, lookups) always rides Celery `link`/`link_error`. This routes it through Postgres instead, via **§5 fire-and-forget self-chaining**: the executor consumer, after running `execute_extraction`, enqueues the continuation itself.

## How

- **Routing** — `dispatch_async` / `dispatch_with_callback` now pick PG-vs-Celery per call via the single `pg_queue_enabled` flag (same gate as the rest of the feature), in both backend and workers `executor_rpc.py`.
- **Self-chaining** — the `on_success`/`on_error` Celery `Signature`s are translated to a serialisable `ContinuationSpec` (`task_name`/`kwargs`/`queue`) carried in the task payload. After the executor runs, the consumer self-chains the matching continuation onto the callback queue: the result dict prepended on success, the dispatch `task_id` on error (Celery `link_error` parity). It acks regardless of chain outcome — a vt-redelivery would re-run the executor (LLM double-spend); a chain failure is logged loud and best-effort.
- **Consumer** — new gated `worker-pg-ide-callback` service drains the `ide_callback` queue (compose profile `pg-queue`).
- **Wire contract** — `ContinuationSpec` + `on_success`/`on_error`/`task_id` added to the shared `TaskPayload` in `unstract.core` (all optional, set only when present).

## Zero-regression

Every new path is gated and **fails closed to Celery**. Gate OFF is byte-identical to the prior behaviour: `dispatch_with_callback` delegates to the unchanged SDK `ExecutionDispatcher` (`send_task(..., link=, link_error=)`), no continuation payload, no self-chain. **Call sites are unchanged** — they keep passing Celery `Signature`s; the dispatcher translates them only on the PG branch.

## Tests

- payload set/unset (optional keys), `signature → spec` translation (incl. missing-queue fail-fast)
- consumer self-chain success/error branches + chain-enqueue-failure-still-acks guard
- routing gate ON/OFF for all three dispatch modes (backend + workers)

## Dev-tested end-to-end (flag ON)

`backend dispatch_with_callback → PG → worker-pg-executor execute_extraction → self-chain → worker-pg-ide-callback ide_prompt_complete → emit-websocket 200, output persisted` — zero RabbitMQ. Non-regression confirmed with a flag ON/OFF A/B (identical behaviour).

After this slice the executor transport is fully on PG; RabbitMQ can be decommissioned next.

### DIFF
--- a/backend/pg_queue/producer.py
+++ b/backend/pg_queue/producer.py
@@ -83,8 +83,14 @@ def enqueue_task(
     (``dispatch_with_callback``): the executor consumer self-chains the matching
     continuation after the task runs. ``task_id`` is the dispatch id prepended to
     ``on_error`` as the failed id (Celery ``link_error`` parity). Mutually
-    exclusive with ``reply_key``.
+    exclusive with ``reply_key`` — passing both is rejected (the consumer checks
+    ``reply_key`` first and would silently drop the callback).
     """
+    if reply_key is not None and (on_success is not None or on_error is not None):
+        raise ValueError(
+            "reply_key (request-reply) and on_success/on_error (callback) are "
+            "mutually exclusive"
+        )
     if not FAIRNESS_MIN_PRIORITY <= priority <= FAIRNESS_MAX_PRIORITY:
         raise ValueError(
             f"priority out of range "
@@ -102,10 +108,13 @@ def enqueue_task(
     # byte-identical to before these fields existed.
     if reply_key is not None:
         message["reply_key"] = reply_key
+    # Continuation specs carry a nested callback ``kwargs`` dict that may hold a
+    # UUID/datetime — coerce like ``args``/``kwargs`` above, else the JSONField
+    # insert raises at dispatch time (caller-visible).
     if on_success is not None:
-        message["on_success"] = on_success
+        message["on_success"] = _json_safe(on_success)
     if on_error is not None:
-        message["on_error"] = on_error
+        message["on_error"] = _json_safe(on_error)
     if task_id is not None:
         message["task_id"] = task_id
     # Mirror the worker _enqueue_pg path: log the failure with breadcrumbs before
--- a/backend/pg_queue/producer.py
+++ b/backend/pg_queue/producer.py
@@ -25,6 +25,7 @@ from unstract.core.data_models import (
     FAIRNESS_DEFAULT_PRIORITY,
     FAIRNESS_MAX_PRIORITY,
     FAIRNESS_MIN_PRIORITY,
+    ContinuationSpec,
     FairnessPayload,
     TaskPayload,
 )
@@ -63,6 +64,9 @@ def enqueue_task(
     priority: int = DEFAULT_PRIORITY,
     fairness: FairnessPayload | None = None,
     reply_key: str | None = None,
+    on_success: ContinuationSpec | None = None,
+    on_error: ContinuationSpec | None = None,
+    task_id: str | None = None,
 ) -> int:
     """Enqueue a task onto the PG queue; returns the new ``msg_id``.
 
@@ -74,6 +78,12 @@ def enqueue_task(
     ``reply_key`` marks a **request-reply** dispatch (the executor RPC on PG):
     the executor consumer writes the task's result/error to ``pg_task_result``
     under it for the blocking caller to poll. Omitted = fire-and-forget.
+
+    ``on_success`` / ``on_error`` mark an **async/callback** dispatch
+    (``dispatch_with_callback``): the executor consumer self-chains the matching
+    continuation after the task runs. ``task_id`` is the dispatch id prepended to
+    ``on_error`` as the failed id (Celery ``link_error`` parity). Mutually
+    exclusive with ``reply_key``.
     """
     if not FAIRNESS_MIN_PRIORITY <= priority <= FAIRNESS_MAX_PRIORITY:
         raise ValueError(
@@ -88,10 +98,16 @@ def enqueue_task(
         "queue": pg_queue,
         "fairness": fairness,
     }
-    # Only set for request-reply dispatches — keeps fire-and-forget rows
-    # byte-identical to before this field existed.
+    # Each optional key is set only when present — keeps fire-and-forget rows
+    # byte-identical to before these fields existed.
     if reply_key is not None:
         message["reply_key"] = reply_key
+    if on_success is not None:
+        message["on_success"] = on_success
+    if on_error is not None:
+        message["on_error"] = on_error
+    if task_id is not None:
+        message["task_id"] = task_id
     # Mirror the worker _enqueue_pg path: log the failure with breadcrumbs before
     # it propagates, so a DB/constraint/serialization error isn't mislabeled by
     # the caller's broad handler.
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -800,6 +800,12 @@ services:
       - WORKER_PG_QUEUE_CONSUMER_QUEUE=ide_callback
       - WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT=8090
       - WORKER_PG_QUEUE_CONSUMER_CONCURRENCY=${PG_IDE_CALLBACK_CONCURRENCY:-2}
+      # Callbacks are sub-second (ws emit + small API writes), so nowhere near the
+      # executor's 1h limit — but set the visibility timeout above a slow internal
+      # API write (backend under load) so a sibling replica can't re-claim mid-run
+      # and double-emit. Health-stale sits above it. Overridable per-env.
+      - WORKER_PG_QUEUE_CONSUMER_VT_SECONDS=${PG_IDE_CALLBACK_VT_SECONDS:-120}
+      - WORKER_PG_QUEUE_CONSUMER_HEALTH_STALE_SECONDS=${PG_IDE_CALLBACK_HEALTH_STALE_SECONDS:-180}
     labels:
       - traefik.enable=false
     volumes:
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -770,6 +770,44 @@ services:
     profiles:
       - pg-queue
 
+  # IDE callback consumer (③c) — drains the ``ide_callback`` queue the executor
+  # self-chains onto for async/callback dispatch (Prompt Studio run/index/extract,
+  # lookups). Registers the ide_callback worker's tasks (ide_prompt_complete,
+  # ide_index_complete, extraction_complete, …). Dark until dispatch_with_callback
+  # routes to PG (flag on); the Celery ``worker-ide-callback`` still serves the
+  # Celery path. Callbacks are lightweight (ws emit + small API writes), so a
+  # modest concurrency suffices; scale via replicas (SKIP LOCKED) under load.
+  worker-pg-ide-callback:
+    image: unstract/worker-unified:${VERSION}
+    container_name: unstract-worker-pg-ide-callback
+    restart: unless-stopped
+    command: ["pg-queue-consumer"]
+    ports:
+      - "8100:8090"
+    env_file:
+      - ../workers/.env
+      - ./essentials.env
+    depends_on:
+      - db
+      - redis
+      - platform-service
+    environment:
+      - ENVIRONMENT=development
+      - APPLICATION_NAME=unstract-worker-pg-ide-callback
+      - WORKER_BARRIER_BACKEND=pg
+      - WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE=ide_callback
+      - WORKER_PG_QUEUE_CONSUMER_QUEUE=ide_callback
+      - WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT=8090
+      - WORKER_PG_QUEUE_CONSUMER_CONCURRENCY=${PG_IDE_CALLBACK_CONCURRENCY:-2}
+    labels:
+      - traefik.enable=false
+    volumes:
+      - ./workflow_data:/data
+      - ${TOOL_REGISTRY_CONFIG_SRC_PATH}:/data/tool_registry_config
+      - prompt_studio_data:/app/prompt-studio-data
+    profiles:
+      - pg-queue
+
   # Reaper / orchestrator — leader-elected loop. Run exactly ONE instance (it
   # elects a single leader via pg_orchestrator_lock; extra replicas idle as
   # standby). Besides barrier-orphan recovery it runs the PG scheduler tick
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -773,10 +773,11 @@ services:
   # IDE callback consumer (③c) — drains the ``ide_callback`` queue the executor
   # self-chains onto for async/callback dispatch (Prompt Studio run/index/extract,
   # lookups). Registers the ide_callback worker's tasks (ide_prompt_complete,
-  # ide_index_complete, extraction_complete, …). Dark until dispatch_with_callback
-  # routes to PG (flag on); the Celery ``worker-ide-callback`` still serves the
-  # Celery path. Callbacks are lightweight (ws emit + small API writes), so a
-  # modest concurrency suffices; scale via replicas (SKIP LOCKED) under load.
+  # ide_index_complete, extraction_complete, …). Idle (receives no work) until
+  # dispatch_with_callback routes to PG (i.e. the gating flag is on); the Celery
+  # ``worker-ide-callback`` still serves the Celery path. No broker dependency:
+  # these callbacks only do API writes + ws emits, never an onward Celery dispatch.
+  # Lightweight, so a modest concurrency suffices; scale via replicas (SKIP LOCKED).
   worker-pg-ide-callback:
     image: unstract/worker-unified:${VERSION}
     container_name: unstract-worker-pg-ide-callback
--- a/unstract/core/src/unstract/core/data_models.py
+++ b/unstract/core/src/unstract/core/data_models.py
@@ -311,8 +311,9 @@ class ContinuationSpec(TypedDict):
     the prompt-studio call sites keep passing Celery ``Signature``s unchanged and
     the dispatcher translates them only on the PG branch.
 
-    The consumer prepends the chained argument the callback expects (the executor
-    result dict on success, the dispatch ``task_id`` on error) before ``kwargs`` —
+    The consumer passes the chained value the callback expects (the executor
+    result dict on success, the dispatch ``task_id`` on error) as the callback's
+    first positional arg, alongside the spec's ``kwargs`` as a distinct mapping —
     mirroring how Celery's ``link`` prepends the parent task's return value.
     """
 
--- a/unstract/core/src/unstract/core/data_models.py
+++ b/unstract/core/src/unstract/core/data_models.py
@@ -298,6 +298,29 @@ class FairnessPayload(TypedDict):
     pipeline_priority: int
 
 
+class ContinuationSpec(TypedDict):
+    """A self-chained continuation — the PG wire-form of a Celery ``Signature``.
+
+    The async/callback executor path (``dispatch_with_callback``) attaches an
+    ``on_success`` / ``on_error`` callback. Celery carries these as ``link`` /
+    ``link_error`` signatures the broker fires automatically; PG has no such
+    broker mechanism, so the executor consumer **self-chains** instead — after it
+    runs ``execute_extraction`` it enqueues this continuation onto ``queue`` (the
+    §5 fire-and-forget model). ``task_name`` / ``kwargs`` / ``queue`` are exactly
+    the three fields read off a ``signature(task_name, kwargs=..., queue=...)`` so
+    the prompt-studio call sites keep passing Celery ``Signature``s unchanged and
+    the dispatcher translates them only on the PG branch.
+
+    The consumer prepends the chained argument the callback expects (the executor
+    result dict on success, the dispatch ``task_id`` on error) before ``kwargs`` —
+    mirroring how Celery's ``link`` prepends the parent task's return value.
+    """
+
+    task_name: str
+    kwargs: dict[str, Any]
+    queue: str
+
+
 class TaskPayload(TypedDict):
     """The ``message`` JSONB of a task row in ``pg_queue_message``.
 
@@ -331,6 +354,15 @@ class TaskPayload(TypedDict):
     queue: str | None
     fairness: FairnessPayload | None
     reply_key: NotRequired[str]
+    # Async/callback dispatch (``dispatch_with_callback``). Mutually exclusive with
+    # ``reply_key``: request-reply (blocking) sets ``reply_key``; fire-and-forget
+    # with callbacks sets these continuations. The executor consumer self-chains
+    # whichever applies after the task runs. ``task_id`` is the dispatch id the
+    # caller tracks (== Celery's task id); the consumer prepends it as the failed
+    # id to ``on_error`` (parity with Celery ``link_error``).
+    on_success: NotRequired[ContinuationSpec]
+    on_error: NotRequired[ContinuationSpec]
+    task_id: NotRequired[str]
 
 
 class PgTaskStatus(str, Enum):
--- a/workers/ide_callback/tasks.py
+++ b/workers/ide_callback/tasks.py
@@ -94,8 +94,20 @@ def _emit_event(
     )
 
 
-def _get_task_error(failed_task_id: str, default: str) -> str:
-    """Retrieve the error message from a failed Celery task's result backend."""
+def _get_task_error(
+    failed_task_id: str, default: str, explicit: str | None = None
+) -> str:
+    """Resolve the error text for a failed-task callback.
+
+    Prefers an ``explicit`` error when the caller already has it: the PG
+    self-chained path carries the real error via ``callback_kwargs['error']``
+    because the executor ran eagerly (``task.apply``) and never wrote a Celery
+    result backend under ``failed_task_id`` — so the ``AsyncResult`` lookup below
+    is empty there. The Celery ``link_error`` path passes no explicit error and
+    falls back to the result backend, then the ``default``.
+    """
+    if explicit:
+        return explicit
     try:
         from celery.result import AsyncResult
 
@@ -321,7 +333,9 @@ def ide_index_error(
     api = _get_api_client()
 
     try:
-        error_msg = _get_task_error(failed_task_id, default="Indexing failed")
+        error_msg = _get_task_error(
+            failed_task_id, default="Indexing failed", explicit=cb.get("error")
+        )
 
         # Clean up the indexing-in-progress flag
         if doc_id_key:
@@ -499,7 +513,9 @@ def ide_prompt_error(
     api = _get_api_client()
 
     try:
-        error_msg = _get_task_error(failed_task_id, default="Prompt execution failed")
+        error_msg = _get_task_error(
+            failed_task_id, default="Prompt execution failed", explicit=cb.get("error")
+        )
 
         _emit_event(
             api,
@@ -642,7 +658,11 @@ def extraction_error(
     # Context-manage clients to avoid per-task session leaks.
     with _get_extraction_client() as api, _get_api_client() as ps_api:
         try:
-            error_msg = _get_task_error(failed_task_id, default="Text extraction failed")
+            error_msg = _get_task_error(
+                failed_task_id,
+                default="Text extraction failed",
+                explicit=cb.get("error"),
+            )
 
             api.mark_extraction_error(
                 source=source,
--- a/workers/ide_callback/tasks.py
+++ b/workers/ide_callback/tasks.py
@@ -106,7 +106,7 @@ def _get_task_error(
     is empty there. The Celery ``link_error`` path passes no explicit error and
     falls back to the result backend, then the ``default``.
     """
-    if explicit:
+    if explicit is not None:
         return explicit
     try:
         from celery.result import AsyncResult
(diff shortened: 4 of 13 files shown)


muhammad-ali-e (commented):
## Automated PR review — PG Queue 9i (executor async/callback path)

Ran the PR Review Toolkit (code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier) over `ab80a494..7afc6d19`. Overall the implementation is well-structured and unusually well-commented, and the happy-path test coverage is solid. The inline comments below are the deduplicated findings, prioritised.

**Critical (1):** the early-drop branches (malformed / poison / unknown-task) never chain `on_error`, so a `dispatch_with_callback` caller hangs forever with no user-facing error.

**Important (5):** the `reply_key` XOR callback invariant is documented but unenforced; the PG `on_error` callback can't recover the real executor error via `AsyncResult`; success-result self-chaining and the backend producer don't JSON-coerce callback payloads (silent drop / insert error on UUID/datetime); `_signature_to_spec` accepts an empty `task_name` and silently drops positional callback args.

**Suggestions:** `_DispatchHandle` + `_signature_to_spec` are duplicated byte-for-byte across backend/workers; returned `ExecutionResult(success=False)` routes to `on_success`; over-broad swallow in `_continuation_org`; consumer helpers re-widen typed payloads to bare `dict`; minor comment/jargon polish.

Test gaps (no inline anchor — the branches aren't in this diff): add coverage for `on_error` firing on the drop branches, the both-keys-present precedence, and `dispatch_async`/`dispatch_with_callback` enqueue-failure propagation.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -239,6 +247,22 @@ def _handle(self, message: QueueMessage) -> None:
                     reply_key,
                 )
                 return
+            if on_error:
```
muhammad-ali-e: **Critical — `on_error` callback dropped on the early-drop branches.** This new block correctly self-chains `on_error` when the *run raises*, but the three early-drop branches above — malformed payload, poison/`max_attempts` exceeded, and unknown/unregistered task — only call `_fail_reply(reply_key, ...)` and then `delete()`. For a `dispatch_with_callback` message (no `reply_key`, has `on_error`) those branches delete the row and **never chain `on_error`**. A poison executor task is exactly the realistic failure mode, and it's the case where the user most needs the error surfaced; instead the HTTP-202 caller waits on a WebSocket error event that never arrives (run appears to hang, no terminal state). This is asymmetric with the `reply_key` path, which *does* fail those branches.

**Fix:** in each early-drop branch, in addition to `_fail_reply`, chain the error continuation when present, e.g. a helper `_fail_dispatch(payload, error=...)` that does `_fail_reply(reply_key, ...)` *and* `if on_error: self._chain_continuation(on_error, prepend=payload.get('task_id') or '', payload=payload)` before the `delete()`. Add tests for the unknown-task / poison / malformed cases with `on_error` set.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -239,6 +247,22 @@ def _handle(self, message: QueueMessage) -> None:
...
+                # Async/callback: self-chain the on_error continuation, then ACK
+                # regardless — same anti-double-spend reasoning as reply_key (a
+                # vt-redelivery would re-run the executor / re-spend LLM tokens).
+                # _chain_continuation is best-effort, so the ack never wedges.
+                self._chain_continuation(
```
muhammad-ali-e: **Important — PG `on_error` loses the real executor error text.** `prepend=payload.get('task_id')` hands the callback (e.g. `ide_index_error`) the dispatch `task_id` as `failed_task_id`, but that callback resolves the actual message via `AsyncResult(failed_task_id, app=app).result`. On the PG path the executor ran eagerly via `task.apply` and never wrote to a Celery result backend under that id, so `AsyncResult(...).result` is empty and the callback degrades to its generic `default` error string. Parity is structural (right positional shape) but not behavioural — the real error is dropped.

**Fix:** the consumer has the exception in hand at the failure branch — carry the real error text into the `on_error` continuation (e.g. prepend `f"{type(exc).__name__}: {exc}"`, or pass it via the continuation kwargs), or store the error under `task_id` in a backend the callback can read.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -267,6 +291,12 @@ def _handle(self, message: QueueMessage) -> None:
                     message.msg_id,
                     reply_key,
                 )
+        elif on_success:
+            # Async/callback: self-chain the success continuation onto the callback
+            # queue before the ack — the §5 hand-off (PG analogue of Celery's link).
+            # Best-effort (never raises): a chain failure logs + still acks, so the
+            # executor is not re-run (LLM double-spend); the callback is lost.
```
muhammad-ali-e: **Important — success callback silently dropped on non-JSON-safe results.** `prepend=eager.result` is enqueued as `args=[eager.result]`; `client.send` serialises with plain `json.dumps(message)` (no `default=str`). An executor result dict containing a UUID/datetime — exactly the kind of payload these results carry — makes `json.dumps` raise `TypeError`, which `_chain_continuation`'s broad `except` swallows ("FAILED to self-chain"), so the success callback and its user-facing WebSocket event are lost. The Celery path serialises via its configured serializer and has no such gap.

**Fix:** JSON-coerce the prepended result before enqueuing (mirror `backend/pg_queue/producer.py:_json_safe`), or give `client.send`'s `json.dumps` a `default=str`.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -267,6 +291,12 @@ def _handle(self, message: QueueMessage) -> None:
                     message.msg_id,
                     reply_key,
                 )
+        elif on_success:
```
muhammad-ali-e, line 299: **Suggestion — a returned `ExecutionResult(success=False)` chains `on_success`.** The success/error split keys purely on whether `task.apply(...)` raised, but `execute_extraction` can *return* a failed `ExecutionResult` (`success=False`, via `ExecutionResult.failure(...)`) without raising. That takes this `elif on_success` branch and chains the **success** continuation with a failure payload; `on_error` never fires. This matches Celery semantics (`link_error` also only fires on raised exceptions), so it's parity rather than a regression — but given the user-facing impact, either inspect `eager.result.success is False` and route to `on_error`, or add a comment documenting that returned-failure deliberately follows the success path, so a future reader doesn't mistake it for the drop-branch bug above.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -289,6 +319,55 @@ def _store_reply(
             self._result_backend = PgResultBackend()
         self._result_backend.store_result(reply_key, result=result, error=error)
 
+    def _chain_continuation(self, spec: dict, *, prepend: object, payload: dict) -> None:
```
muhammad-ali-e: **Suggestion — restore the typed contract at the consumer boundary.** `_chain_continuation(self, spec: dict, *, prepend, payload: dict)` re-widens to bare `dict`, discarding the new `ContinuationSpec` / `TaskPayload` types the PR added precisely for these shapes. A key rename across the producer↔consumer boundary (e.g. `task_name` → `name`) wouldn't be caught by the type checker. Annotate `spec: ContinuationSpec, payload: TaskPayload` (imports already available in this tree).

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -289,6 +319,55 @@ def _store_reply(
...
+
+        The executor request carries it in the context dict (``args[0]``);
+        callbacks are not fairness-critical, so any extraction failure degrades to
+        ``""`` (no org).
+        """
+        try:
+            args = payload.get("args") or []
+            if args and isinstance(args[0], dict):
+                return str(args[0].get("organization_id") or "")
+        except Exception:
```
muhammad-ali-e: **Suggestion — over-broad swallow in `_continuation_org`.** The guards already present (`payload.get('args') or []`, `if args and isinstance(args[0], dict)`) mean the wrapped operations can't plausibly raise, so this bare `except Exception: pass` adds no real protection and silently hides any genuinely unexpected payload shape. Either drop the try/except (the guards suffice) or `logger.debug(...)` inside the `except` instead of `pass`, so an unexpected shape is observable.

workers/queue_backend/pg_queue/task_payload.py
```diff
@@ -47,8 +57,14 @@ def to_payload(
         queue=queue,
         fairness=fairness.to_dict() if fairness is not None else None,
     )
-    # Only set for request-reply dispatches — keeps fire-and-forget rows
-    # byte-identical to before this field existed (mirrors the backend producer).
+    # Each optional key is set only when present — keeps fire-and-forget rows
+    # byte-identical to before these fields existed (mirrors the backend producer).
     if reply_key is not None:
         payload["reply_key"] = reply_key
+    if on_success is not None:
```
muhammad-ali-e, line 71: **Important — the `reply_key` XOR `on_success`/`on_error` invariant is documented but never enforced.** The docstring (and the mirrors in `data_models.py`, `producer.py`) call these "mutually exclusive with `reply_key`", but `to_payload` accepts and sets all of them. If both are passed, the consumer checks `reply_key` first and silently skips the callback path (`elif on_success`) — a dropped callback with no log.

**Fix:** fail fast at this construction boundary (and the twin in `backend/pg_queue/producer.py` `enqueue_task`):
```python
if reply_key is not None and (on_success is not None or on_error is not None):
    raise ValueError("reply_key (request-reply) and on_success/on_error (callback) are mutually exclusive")
```
Consider also `if on_error is not None and not task_id: raise ValueError(...)`, since a missing `task_id` silently prepends `""` as the failed id.

backend/pg_queue/producer.py
```diff
@@ -88,10 +98,16 @@ def enqueue_task(
...
         "fairness": fairness,
     }
-    # Only set for request-reply dispatches — keeps fire-and-forget rows
-    # byte-identical to before this field existed.
+    # Each optional key is set only when present — keeps fire-and-forget rows
+    # byte-identical to before these fields existed.
     if reply_key is not None:
         message["reply_key"] = reply_key
+    if on_success is not None:
+        message["on_success"] = on_success
```
muhammad-ali-e: **Important — `on_success`/`on_error` bypass `_json_safe`.** `args`/`kwargs` go through `_json_safe(...)` above (lines 96-97), but the continuation specs (each with a nested `kwargs` dict) are written into the JSONField verbatim. If a callback's kwargs carry a UUID/datetime, `PgQueueMessage.objects.create` raises at insert time — and unlike the worker path this is caller-visible at dispatch. **Fix:** `message["on_success"] = _json_safe(on_success)` (and likewise `on_error`).

workers/queue_backend/pg_queue/executor_rpc.py
```diff
@@ -76,6 +76,46 @@
...
+    if sig is None:
+        return None
+    queue = (getattr(sig, "options", None) or {}).get("queue")
+    if not queue:
+        raise ValueError(
+            f"callback signature {getattr(sig, 'task', sig)!r} has no queue; "
+            "PG self-chaining routes by the row's queue and cannot default it"
+        )
+    return ContinuationSpec(
+        task_name=sig.task,
```
muhammad-ali-e: **Important — `_signature_to_spec` validates `queue` but not `task`, and silently drops positional args.** Two gaps with the same fail-fast philosophy already applied to `queue`:

1. `task_name=sig.task` is read with no guard. If `sig.task` is empty/`None` (a malformed signature), the continuation enqueues with a falsy `task_name` and is dropped far downstream as "malformed payload: missing task_name" — a lost callback with a misleading log. Add `if not getattr(sig, 'task', None): raise ValueError(...)`.
2. `ContinuationSpec` has no `args`, and only `sig.kwargs` is read. A Celery `Signature` may carry positional `args`; on the PG branch they're silently dropped, so the callback runs with a different arg list than on Celery — violating the "call sites unchanged" contract. Today's call sites are kwargs-only (latent), but make it loud: `if getattr(sig, 'args', None): raise ValueError("callback signature has positional args; PG self-chaining supports kwargs only")`.

workers/queue_backend/pg_queue/executor_rpc.py
```diff
@@ -76,6 +76,46 @@
 _DEFAULT_TIMEOUT = 3600
 
 
+class _DispatchHandle:
```
muhammad-ali-e: **Suggestion — `_DispatchHandle` and `_signature_to_spec` are duplicated byte-for-byte with `backend/pg_queue/executor_rpc.py`.** Both files now carry identical copies of this class and the `_signature_to_spec` helper (and near-identical `dispatch_async`/`dispatch_with_callback` bodies). The new translation logic — especially the correctness-bearing "fail fast on missing queue" rule — can now drift silently between the two copies. `ContinuationSpec`/`PgTaskStatus` were hoisted into `unstract.core` for exactly this reason; `_DispatchHandle` and `_signature_to_spec` (no Django/worker deps) belong there too. Import both copies from one shared definition.

unstract/core/src/unstract/core/data_models.py
```diff
@@ -298,6 +298,29 @@ class FairnessPayload(TypedDict):
...
+    ``on_success`` / ``on_error`` callback. Celery carries these as ``link`` /
+    ``link_error`` signatures the broker fires automatically; PG has no such
+    broker mechanism, so the executor consumer **self-chains** instead — after it
+    runs ``execute_extraction`` it enqueues this continuation onto ``queue`` (the
+    §5 fire-and-forget model). ``task_name`` / ``kwargs`` / ``queue`` are exactly
+    the three fields read off a ``signature(task_name, kwargs=..., queue=...)`` so
+    the prompt-studio call sites keep passing Celery ``Signature``s unchanged and
+    the dispatcher translates them only on the PG branch.
+
+    The consumer prepends the chained argument the callback expects (the executor
```
muhammad-ali-e: **Suggestion — "before `kwargs`" is slightly misleading.** `_chain_continuation` passes the chained value as a separate positional `args=[prepend]` list and `kwargs` as a distinct mapping — they aren't concatenated into one positional sequence, so "before `kwargs`" reads as if kwargs were positional. Reword to e.g.: "… passes the chained argument as the callback's first positional arg, alongside the spec's `kwargs` — mirroring how Celery's `link` prepends the parent task's return value."

docker/docker-compose.yaml
```diff
@@ -770,6 +770,44 @@ services:
     profiles:
       - pg-queue
 
+  # IDE callback consumer (③c) — drains the ``ide_callback`` queue the executor
+  # self-chains onto for async/callback dispatch (Prompt Studio run/index/extract,
+  # lookups). Registers the ide_callback worker's tasks (ide_prompt_complete,
+  # ide_index_complete, extraction_complete, …). Dark until dispatch_with_callback
```
muhammad-ali-e: **Suggestion — "Dark" is jargon for a compose file.** Clear to the authors (idle / no traffic) but opaque to an operator scanning services. Suggest: "Idle (receives no work) until `dispatch_with_callback` routes to PG, i.e. the gating flag is on". Also worth confirming: the sibling `worker-pg-callback`/`worker-pg-scheduler` declare `depends_on: rabbitmq`; this service doesn't. Likely fine since `ide_callback` tasks only do API writes + ws emits, but verify none dispatch onward to a Celery broker.

muhammad-ali-e:
**SonarCloud duplication on new code — addressed (`0f8b57151`).**

Most of the flagged duplication is the intentional backend ↔ workers `executor_rpc.py` mirror (same finding as UN-3605; the two differ only by transport — Django ORM vs psycopg2 — and the full de-dup is the shared-package follow-up **UN-3607**, which can't land yet without a circular `sdk1 → core` import).

But I've cut the *new-code* slice now by lifting the two genuinely transport-agnostic helpers I added into `unstract.core` (alongside `ContinuationSpec`):

- `DispatchHandle` (the `.id`-only `AsyncResult` duck-type)
- `signature_to_continuation` (Celery `Signature` → `ContinuationSpec`)

Both mirrors import them instead of redefining, and the duplicated `TestSignatureToSpec` is removed from both suites in favour of one shared test. Net −34 lines. The `dispatch_async`/`dispatch_with_callback` method *bodies* stay mirrored (genuinely transport-specific) — that residue is UN-3607.

No behaviour change: pure code movement, gate + contract unchanged. Backend (20) + workers (28) executor-RPC suites green; ruff/pre-commit clean.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed (`42862651b`). Added a shared `_fail_dispatch(payload, error=...)` used by the run-raised path **and** all three early-drop branches (malformed / poison / unknown task), so a `dispatch_with_callback` failure always self-chains `on_error`. Tests cover all three drop branches with `on_error` set.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed (`42862651b`). The consumer now carries the real error text into `callback_kwargs['error']`, and `_get_task_error` prefers it over the (empty-on-PG) `AsyncResult` lookup. Test asserts `RuntimeError: boom` reaches the on_error continuation.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed (`42862651b`). Added `_json_safe` (json round-trip with `default=str`) applied to the prepended value in `_chain_continuation`. Test uses a task returning a `datetime` and asserts the chained arg is coerced + plain-json serialisable.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Documented (`42862651b`). Added a comment on the `elif on_success` branch: "success" == did-not-raise (Celery `link` parity); a returned failed ExecutionResult deliberately follows on_success — distinct from the drop-branch bug above.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed (`42862651b`). `_chain_continuation(spec: ContinuationSpec, *, payload: TaskPayload, ...)` and `_continuation_org(payload: TaskPayload)` now carry the real types.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed (`42862651b`). Dropped the try/except — the guards (`payload.get('args') or []`, `isinstance(args[0], dict)`) cover every shape, so nothing inside can raise.

workers/queue_backend/pg_queue/task_payload.py
muhammad-ali-e: Fixed (`42862651b`). `to_payload` and `enqueue_task` now raise `ValueError` if `reply_key` is combined with `on_success`/`on_error`. Tests on both. (Left the on_error-needs-task_id as-is since `dispatch_with_callback` always sets task_id; happy to add if you'd prefer.)

backend/pg_queue/producer.py
muhammad-ali-e: Fixed (`42862651b`). `message['on_success'] = _json_safe(on_success)` (and on_error). Test passes a UUID in callback_kwargs and asserts it's coerced.

workers/queue_backend/pg_queue/executor_rpc.py
muhammad-ali-e: Fixed (`42862651b`, in the now-shared `unstract.core.execution_dispatch`). Guards both: raises on a falsy `task`, and raises on positional `args` (kwargs-only callbacks). Tests for each.

workers/queue_backend/pg_queue/executor_rpc.py
muhammad-ali-e: Done — extracted `DispatchHandle` + `signature_to_continuation` into `unstract.core.execution_dispatch` (commit `0f8b57151`); both mirrors import them and the duplicated `TestSignatureToSpec` is gone in favour of one shared test. (Full mirror retirement remains UN-3607.)

unstract/core/src/unstract/core/data_models.py
muhammad-ali-e: Fixed (`42862651b`). Reworded to '… passes the chained value as the callback's first positional arg, alongside the spec's kwargs as a distinct mapping …'.

docker/docker-compose.yaml
muhammad-ali-e: Fixed (`42862651b`). Reworded 'Dark' → 'Idle (receives no work) until …'. Confirmed no broker dep needed: grepped `ide_callback/tasks.py` — no `.delay`/`.apply_async`/`send_task`/`signature`/`chord`; these callbacks only do API writes + ws emits. Added that note to the compose comment.

muhammad-ali-e:
**Review round addressed** — `42862651b` (fixes) + `0f8b57151` (the duplication extraction). Per-thread replies on each.

**Critical**
- on_error is now self-chained on the early-drop branches too (malformed / poison / unknown task), via a shared `_fail_dispatch` — a `dispatch_with_callback` failure always reaches its on_error callback, not only when the task body raised.

**Important**
- on_error now carries the real error text (`callback_kwargs['error']`; `_get_task_error` prefers it over the empty-on-PG AsyncResult lookup).
- self-chained prepend is JSON-coerced (UUID/datetime no longer silently drops the callback).
- `reply_key` XOR `on_success`/`on_error` enforced in `to_payload` + `enqueue_task`.
- backend producer `_json_safe`s the continuation specs.
- `signature_to_continuation` guards the task name + rejects positional-arg callbacks.

**Suggestions**
- typed `_chain_continuation`/`_continuation_org` boundaries; dropped the over-broad `except`; documented the returned-failure→on_success parity; reworded the `ContinuationSpec` doc and the compose "Dark" comment (+ confirmed no broker dep — ide_callback tasks do no onward Celery dispatch).
- duplication: `DispatchHandle` + `signature_to_continuation` lifted to `unstract.core` (one shared test); the dispatcher method-body residue is the UN-3607 shared-package follow-up.

Tests added for every fix. Backend 35, workers 76 green; ruff/pre-commit clean. SonarCloud new-code duplication cut by the extraction (net −34 there).

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -221,22 +246,23 @@ def _handle(self, message: QueueMessage) -> None:
                 throw=True,
             )
         except Exception as exc:
-            if reply_key:
-                # Record the failure so the waiting caller gets a definitive
-                # result instead of blocking to its timeout, then ACK. We do NOT
+            if reply_key or on_error:
```
muhammad-ali-e: **Correctness — an `on_success`-only callback dispatch that raises falls through to vt-redelivery and re-runs the executor.** The guard here is `if reply_key or on_error:`, but `dispatch_with_callback(on_success=..., on_error=None)` is a permitted shape: both kwargs default to `None` and `signature_to_continuation(None)` returns `None`, so an `on_success`-only callback payload is fully reachable. When such a dispatch's executor raises, this condition is false and control drops into the fire-and-forget branch below, leaving the row for vt-expiry redelivery — exactly the costly re-run (LLM double-spend) the comment on lines 252-258 says this path must avoid — and the caller never receives a terminal event.

Broaden the guard to `if reply_key or on_success or on_error:` so any reply/callback dispatch acks-on-failure rather than redelivering, or enforce in `dispatch_with_callback` that `on_error` is required whenever `on_success` is set.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -267,6 +293,19 @@ def _handle(self, message: QueueMessage) -> None:
...
+            # Best-effort (never raises): a chain failure logs + still acks, so the
+            # executor is not re-run (LLM double-spend); the callback is lost.
+            #
+            # "success" == the task did not RAISE (parity with Celery `link`, which
+            # also fires on any non-exception return). A task that *returns* a failed
+            # ExecutionResult (success=False, no raise) deliberately follows this
+            # path — the on_success callback receives the failed result and renders
+            # it — exactly as on Celery. This is NOT the missing-on_error drop bug
+            # the early-drop branches handle.
+            self._chain_continuation(on_success, prepend=eager.result, payload=payload)
```
muhammad-ali-e: **Silent failure — a failed `on_success` enqueue leaves the caller with no terminal event.** On the success path `_chain_continuation` is best-effort and never raises: if the enqueue fails (transport/DB error, connection drop, a bad `spec["queue"]`), it logs and the executor message is acked with no fallback. Since the on_success callback (`ide_index_complete` / `ide_prompt_complete` / `extraction_complete`) is what emits the WebSocket event the prompt-studio UI blocks on after its 202, a dropped success continuation leaves the user with neither a success nor an error event — the session hangs indefinitely, after the LLM spend already occurred.

This is distinct from the now-fixed non-JSON-safe-result drop: `_json_safe` removes one *cause* of the enqueue raising, but a transport/DB failure still triggers the same silent drop with no recovery channel. Consider having `_chain_continuation` return whether it enqueued, and on a success-path failure fall back to `_fail_dispatch(payload, error="result delivery failed; see worker logs")` so the session still terminates. At minimum, carry `run_id`/`task_id`/`org_id` into the failure log so the stranded session is correlatable later.

muhammad-ali-e (commented):
Second-pass review (PG executor async/callback path). The items from the first pass are genuinely addressed in `42862651`. Two new correctness/silent-failure findings on the dispatch path below; the rest (latent type-design hardening, a docker-compose comment-vs-`depends_on` mismatch, and a critical test gap on `_get_task_error`'s explicit-error precedence) are in the summary rather than posted inline.

muhammad-ali-e (commented):
_(duplicate of the review below — submitted twice by mistake; see the grouped findings in the other second-pass review)_

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed (`ec16fefda`). Broadened the failure guard to `if reply_key or on_success or on_error:` — an on_success-only callback dispatch that raises now ACKs instead of vt-redelivering (no executor re-run). Test: on_success-only + raise → delete called, not redelivered.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: Fixed (`ec16fefda`). `_chain_continuation` now returns whether it enqueued; on a success-path enqueue failure we fall back to `_fail_dispatch(payload, error='result delivery failed; see worker logs')` so the caller gets a terminal on_error event instead of hanging. The failure log now carries run_id/task_id/org for correlation. Test asserts the on_error fallback fires when the on_success send fails.

workers/ide_callback/tasks.py
```diff
@@ -94,8 +94,20 @@ def _emit_event(
...
+
+    Prefers an ``explicit`` error when the caller already has it: the PG
+    self-chained path carries the real error via ``callback_kwargs['error']``
+    because the executor ran eagerly (``task.apply``) and never wrote a Celery
+    result backend under ``failed_task_id`` — so the ``AsyncResult`` lookup below
+    is empty there. The Celery ``link_error`` path passes no explicit error and
+    falls back to the result backend, then the ``default``.
+    """
+    if explicit:
+        return explicit
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed (`ec16fefda`). Changed to `if explicit is not None:` — an explicit "" no longer falls through to the (empty-on-PG) AsyncResult lookup.

docker/docker-compose.yaml
```diff
@@ -770,6 +770,45 @@ services:
...
+      - WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT=8090
+      - WORKER_PG_QUEUE_CONSUMER_CONCURRENCY=${PG_IDE_CALLBACK_CONCURRENCY:-2}
+    labels:
+      - traefik.enable=false
+    volumes:
+      - ./workflow_data:/data
+      - ${TOOL_REGISTRY_CONFIG_SRC_PATH}:/data/tool_registry_config
+      - prompt_studio_data:/app/prompt-studio-data
+    profiles:
+      - pg-queue
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed (`ec16fefda`). Set `WORKER_PG_QUEUE_CONSUMER_VT_SECONDS=${PG_IDE_CALLBACK_VT_SECONDS:-120}` and `...HEALTH_STALE_SECONDS=${...:-180}` on worker-pg-ide-callback — above a slow internal-API write so a sibling replica can't re-claim mid-run and double-emit, with a comment explaining the choice.

muhammad-ali-e:
**Review round 2 addressed** — `ec16fefda`. Per-thread replies on each.

- **Correctness (on_success-only redelivery):** broadened the failure guard to `reply_key or on_success or on_error`, so an on_success-only callback dispatch that raises ACKs instead of vt-redelivering (no executor re-run / LLM double-spend).
- **Silent failure (dropped success enqueue):** `_chain_continuation` now reports whether it enqueued; on a success-path failure we fall back to on_error so the HTTP-202 caller still gets a terminal event instead of hanging, and the failure log carries run_id/task_id/org.
- **greptile:** `_get_task_error` → `if explicit is not None`; `worker-pg-ide-callback` now sets VT=120 / health-stale=180 (overridable) so a slow internal-API write can't be re-claimed mid-run and double-emit.

Tests added for the two correctness fixes. Workers self-chain + ide_callback green.

On the SonarCloud gate: new-code duplication is now **10.4%** (was 30.5%) after lifting the shared helpers — the remainder is the backend↔workers `executor_rpc.py` dispatcher-body mirror, the same intentional duplication accepted on UN-3605, whose full removal is the shared-package follow-up **UN-3607** (can't land here without a circular `sdk1 → core` import).

PR #2101, opened 2026-06-21
UN-3610 [FEAT] PG Queue — reaper retention sweep (pg_task_result + pg_batch_dedup orphans)

muhammad-ali-e:
## What

Wires the deferred retention sweep into the PG reaper, before the gate ramps toward 100%. Targets `feat/UN-3445-pg-queue-integration` (not `main`).

Two queue-infra tables grow unbounded today:
- **`pg_task_result`** — the executor-RPC result store. Every row already carries `expires_at = now()+3600s` (written by the consumer's `store_result`) and the `pg_task_result_expires_idx` index exists, but nothing deletes expired rows (the model docstring flags the sweep as "not wired yet").
- **`pg_batch_dedup`** — per-`(execution_id, batch_index)` idempotency markers. Cleared on barrier teardown / reaper barrier-recovery, but a partial-failure execution can leave orphans.

## How

- **`sweep_expired_results(conn)`** — `DELETE FROM pg_task_result WHERE expires_at <= now()`. Once past `expires_at` no caller can still be waiting (the result outlived the caller-timeout).
- **`sweep_orphan_dedup(conn, retention)`** — `DELETE FROM pg_batch_dedup WHERE created_at <= now() - interval`. Time-based backstop; retention (default **24h**) must exceed the longest execution so an in-flight marker is never swept.
- Both run in `PgReaper.tick()` **under leadership**, after recovery + schedules, **cadence-gated** by `WORKER_PG_REAPER_SWEEP_SECONDS` (default 300s) so the ~5s loop doesn't `DELETE` every cycle. Own try/except discards the owned sweep conn on error (parity with the existing steps).
- Env knobs (`WORKER_PG_REAPER_SWEEP_SECONDS`, `WORKER_PG_DEDUP_RETENTION_SECONDS`) surfaced on the compose `worker-pg-reaper` service.

## Non-regression

Workers-only: **no migration, no writer change, no flag**. Idempotent (`DELETE WHERE`), dark until rows exist; cadence + leader-gated. The reaper already runs — this adds a third, rarer job to its tick.

## Tests

Env parsers (default/override/invalid), SQL contract (mock cursor — asserts the DELETE shape + bound retention param + rollback-on-error), and tick wiring: leader-only / standby-skips / cadence-gated (one sweep across two quick ticks) / runs-after-recovery-and-schedule / error-discards-owned-conn. **27 green.**

## Dev-tested live

Recreated `worker-pg-reaper` on the rebuilt image; on its first leader tick:
```
Reaper: acquired leadership
Reaper: retention sweep deleted 9 pg_task_result + 1 pg_batch_dedup row(s)
```
The 9 `pg_task_result` rows were real orphans from earlier executor-RPC dev-tests (all >1h past `expires_at`); the `pg_batch_dedup` row was a 25h-old marker seeded for the test. Both tables clean afterward, counts logged.

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -837,6 +837,12 @@ services:
       - APPLICATION_NAME=unstract-worker-pg-reaper
       - WORKER_PG_REAPER_INTERVAL_SECONDS=${WORKER_PG_REAPER_INTERVAL_SECONDS:-5}
       - WORKER_PG_REAPER_HEALTH_PORT=8086
+      # Retention sweep (cadence-gated inside the leader's tick): drop expired
+      # pg_task_result rows + orphaned pg_batch_dedup markers so neither grows
+      # unbounded as the gate ramps. Dedup retention must exceed the longest
+      # execution (default 24h) so an in-flight marker is never swept.
+      - WORKER_PG_REAPER_SWEEP_SECONDS=${WORKER_PG_REAPER_SWEEP_SECONDS:-300}
+      - WORKER_PG_DEDUP_RETENTION_SECONDS=${WORKER_PG_DEDUP_RETENTION_SECONDS:-86400}
     labels:
       - traefik.enable=false
     volumes:
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -839,8 +839,9 @@ services:
       - WORKER_PG_REAPER_HEALTH_PORT=8086
       # Retention sweep (cadence-gated inside the leader's tick): drop expired
       # pg_task_result rows + orphaned pg_batch_dedup markers so neither grows
-      # unbounded as the gate ramps. Dedup retention must exceed the longest
-      # execution (default 24h) so an in-flight marker is never swept.
+      # unbounded as the gate ramps. The dedup retention below (default 24h) should
+      # be kept above the longest possible execution, so an in-flight marker is
+      # never swept.
       - WORKER_PG_REAPER_SWEEP_SECONDS=${WORKER_PG_REAPER_SWEEP_SECONDS:-300}
       - WORKER_PG_DEDUP_RETENTION_SECONDS=${WORKER_PG_DEDUP_RETENTION_SECONDS:-86400}
     labels:
--- a/workers/queue_backend/pg_queue/reaper.py
+++ b/workers/queue_backend/pg_queue/reaper.py
@@ -51,7 +51,8 @@ import os
 import signal
 import threading
 import time
-from typing import TYPE_CHECKING, NamedTuple, Protocol
+from collections.abc import Callable
+from typing import TYPE_CHECKING, NamedTuple, Protocol, TypeVar
 
 from unstract.core.data_models import ExecutionStatus
 
@@ -69,6 +70,15 @@ logger = logging.getLogger(__name__)
 # Cadence: how often the leader renews + runs recovery. Enforced shorter than
 # the lease window in PgReaper.__init__.
 _DEFAULT_REAPER_INTERVAL_SECONDS = 5.0
+# Retention sweep cadence — far rarer than the tick (cleanup, not recovery), so
+# the per-cycle DELETE doesn't run every few seconds.
+_DEFAULT_SWEEP_INTERVAL_SECONDS = 300.0
+# How long an orphaned ``pg_batch_dedup`` marker lives before the sweep drops it.
+# MUST exceed the longest possible execution so an in-flight marker is never swept
+# out from under a running fan-out. 24h is comfortably above any single execution.
+_DEFAULT_DEDUP_RETENTION_SECONDS = 86400
+
+_N = TypeVar("_N", int, float)
 
 
 class LeaderLeaseLike(Protocol):
@@ -125,6 +135,91 @@ def reaper_interval_from_env() -> float:
             f"Unset it to default to {_DEFAULT_REAPER_INTERVAL_SECONDS}s."
         )
     return value
+
+
+def _positive_duration_from_env(name: str, default: _N, cast: Callable[[str], _N]) -> _N:
+    """Read a positive duration env var (default on unset; raise on invalid/<=0).
+
+    Same loud-on-misconfig posture as :func:`reaper_interval_from_env`, shared by
+    the sweep-cadence and dedup-retention knobs.
+    """
+    raw = os.getenv(name)
+    if raw is None:
+        return default
+    try:
+        value = cast(raw)
+    except ValueError as exc:
+        raise ValueError(
+            f"{name}={raw!r} is not a number. Unset it to default to {default}."
+        ) from exc
+    if value <= 0:
+        raise ValueError(
+            f"{name}={value} must be positive. Unset it to default to {default}."
+        )
+    return value
+
+
+def reaper_sweep_interval_from_env() -> float:
+    """Retention-sweep cadence from ``WORKER_PG_REAPER_SWEEP_SECONDS`` (default 300s)."""
+    return _positive_duration_from_env(
+        "WORKER_PG_REAPER_SWEEP_SECONDS", _DEFAULT_SWEEP_INTERVAL_SECONDS, float
+    )
+
+
+def dedup_retention_from_env() -> int:
+    """Dedup-orphan age from ``WORKER_PG_DEDUP_RETENTION_SECONDS`` (default 24h)."""
+    return _positive_duration_from_env(
+        "WORKER_PG_DEDUP_RETENTION_SECONDS", _DEFAULT_DEDUP_RETENTION_SECONDS, int
+    )
+
+
+def sweep_expired_results(conn: PgConnection) -> int:
+    """Delete expired executor-RPC result rows; return the number deleted.
+
+    ``pg_task_result`` rows carry ``expires_at = now() + retention`` (written by
+    the consumer's ``store_result``). Once past it no caller can still be waiting
+    (the result outlived the caller-timeout), so the row is safe to drop. The
+    table has no other reader once expired, so this is the only thing keeping it
+    from growing unbounded with each RPC. Idempotent (``DELETE … WHERE``) and uses
+    the ``pg_task_result_expires_idx`` index. Rolls back on error so the manual-
+    commit connection isn't left in an aborted-txn state.
+    """
+    try:
+        with conn.cursor() as cur:
+            cur.execute("DELETE FROM pg_task_result WHERE expires_at <= now()")
+            deleted = cur.rowcount
+        conn.commit()
+        return deleted
+    except Exception:
+        with contextlib.suppress(Exception):
+            conn.rollback()
+        raise
+
+
+def sweep_orphan_dedup(conn: PgConnection, retention_seconds: int) -> int:
+    """Delete orphaned per-batch dedup markers older than *retention_seconds*.
+
+    ``pg_batch_dedup`` markers are normally cleared on barrier teardown / reaper
+    barrier-recovery, but a partial-failure execution can leave them behind. A
+    marker only matters while its execution is in flight, so anything older than
+    the longest possible execution (*retention_seconds*) is a safe-to-drop orphan.
+    Time-based on ``created_at`` (the table has no per-row expiry column).
+    Idempotent; rolls back on error.
+    """
+    try:
+        with conn.cursor() as cur:
+            cur.execute(
+                "DELETE FROM pg_batch_dedup "
+                "WHERE created_at <= now() - make_interval(secs => %s)",
+                (retention_seconds,),
+            )
+            deleted = cur.rowcount
+        conn.commit()
+        return deleted
+    except Exception:
+        with contextlib.suppress(Exception):
+            conn.rollback()
+        raise
 
 
 def _execution_status(
@@ -380,6 +475,8 @@ class PgReaper:
         lease: LeaderLeaseLike,
         *,
         interval_seconds: float | None = None,
+        sweep_interval_seconds: float | None = None,
+        dedup_retention_seconds: int | None = None,
         sweep_conn: PgConnection | None = None,
         api_client: InternalAPIClient | None = None,
     ) -> None:
@@ -401,6 +498,25 @@ class PgReaper:
                 f"lease window {lease.lease_seconds}s, or the leader loses the "
                 f"lease between renews"
             )
+        # Retention-sweep cadence + dedup-orphan horizon (validated like interval:
+        # an injected non-positive value reaches here unvalidated by the env parser).
+        self._sweep_interval = (
+            sweep_interval_seconds
+            if sweep_interval_seconds is not None
+            else reaper_sweep_interval_from_env()
+        )
+        if self._sweep_interval <= 0:
+            raise ValueError("sweep_interval_seconds must be positive")
+        self._dedup_retention = (
+            dedup_retention_seconds
+            if dedup_retention_seconds is not None
+            else dedup_retention_from_env()
+        )
+        if self._dedup_retention <= 0:
+            raise ValueError("dedup_retention_seconds must be positive")
+        # 0 → "never swept", so the first leader tick sweeps immediately; advanced
+        # to monotonic() each sweep so the cadence holds thereafter.
+        self._last_sweep_monotonic = 0.0
         self._sweep_conn = sweep_conn
         self._owns_sweep_conn = sweep_conn is None
         # Lazily built so the reaper can be constructed without env/HTTP set up
@@ -499,7 +615,38 @@ class PgReaper:
         except Exception:
             self._discard_owned_sweep_conn()
             raise
+        # Orchestrator's third job: retention cleanup (cadence-gated, so it does
+        # NOT run every tick). Last so a sweep error can't skip recovery/schedules.
+        self._maybe_sweep()
         return TickOutcome(was_leader=True, reclaimed=reclaimed)
+
+    def _maybe_sweep(self) -> None:
+        """Run the retention sweep at most once per ``_sweep_interval``.
+
+        Leader-only (called from :meth:`tick` after recovery + schedules). Deletes
+        expired ``pg_task_result`` rows and orphaned ``pg_batch_dedup`` markers so
+        neither table grows unbounded as the gate ramps. The cadence is advanced
+        BEFORE sweeping so a failure waits one interval before retry rather than
+        hammering the DB every tick; the owned sweep conn is discarded on error
+        (parity with recovery/schedules).
+        """
+        now = time.monotonic()
+        if now - self._last_sweep_monotonic < self._sweep_interval:
+            return
+        self._last_sweep_monotonic = now
+        try:
+            results = sweep_expired_results(self._get_sweep_conn())
+            dedup = sweep_orphan_dedup(self._get_sweep_conn(), self._dedup_retention)
+        except Exception:
+            self._discard_owned_sweep_conn()
+            raise
+        if results or dedup:
+            logger.info(
+                "Reaper: retention sweep deleted %s pg_task_result + "
+                "%s pg_batch_dedup row(s)",
+                results,
+                dedup,
+            )
 
     def run(self, *, install_signals: bool = True) -> None:
         """Lease-maintenance + recovery loop until stopped; releases on exit."""
--- a/workers/queue_backend/pg_queue/reaper.py
+++ b/workers/queue_backend/pg_queue/reaper.py
@@ -151,8 +151,10 @@ def _positive_duration_from_env(name: str, default: _N, cast: Callable[[str], _N
     try:
         value = cast(raw)
     except ValueError as exc:
+        # Include the cause: an int knob given "1.5" IS a number, just not an int —
+        # "is not a number" would mislead. The cause spells out the real reason.
         raise ValueError(
-            f"{name}={raw!r} is not a number. Unset it to default to {default}."
+            f"{name}={raw!r} cannot be parsed: {exc}. Unset it to default to {default}."
         ) from exc
     if value <= 0:
         raise ValueError(
@@ -696,7 +698,7 @@ class PgReaper:
             self._sweep_fail_streak[table] = streak
             logger.exception(
                 "Reaper: retention sweep of %s failed (%s consecutive) — will retry "
-                "next cycle",
+                "after the next sweep interval",
                 table,
                 streak,
             )
--- a/workers/queue_backend/pg_queue/reaper.py
+++ b/workers/queue_backend/pg_queue/reaper.py
@@ -73,9 +73,11 @@ _DEFAULT_REAPER_INTERVAL_SECONDS = 5.0
 # Retention sweep cadence — far rarer than the tick (cleanup, not recovery), so
 # the per-cycle DELETE doesn't run every few seconds.
 _DEFAULT_SWEEP_INTERVAL_SECONDS = 300.0
-# How long an orphaned ``pg_batch_dedup`` marker lives before the sweep drops it.
-# MUST exceed the longest possible execution so an in-flight marker is never swept
-# out from under a running fan-out. 24h is comfortably above any single execution.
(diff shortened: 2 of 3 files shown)


workers/queue_backend/pg_queue/reaper.py
```diff
@@ -499,8 +615,39 @@ def tick(self) -> TickOutcome:
...
+        hammering the DB every tick; the owned sweep conn is discarded on error
+        (parity with recovery/schedules).
+        """
+        now = time.monotonic()
+        if now - self._last_sweep_monotonic < self._sweep_interval:
+            return
+        self._last_sweep_monotonic = now
+        try:
+            results = sweep_expired_results(self._get_sweep_conn())
+            dedup = sweep_orphan_dedup(self._get_sweep_conn(), self._dedup_retention)
```
muhammad-ali-e: **[High] Partial-sweep starvation.** `_last_sweep_monotonic` is advanced at line 636 *before* both DELETEs run under one `try`. If `sweep_expired_results` (638) raises, `sweep_orphan_dedup` (639) never runs this cycle, and because the cadence stamp already advanced it stays gated out for a full `_sweep_interval` (~300s). The two cleanups are documented as independent ("neither table grows unbounded"), but a persistent fault in one silently starves the other.

Suggest isolating them so one failing doesn't skip the other:
```python
for name, fn in (("pg_task_result", lambda: sweep_expired_results(self._get_sweep_conn())),
                 ("pg_batch_dedup", lambda: sweep_orphan_dedup(self._get_sweep_conn(), self._dedup_retention))):
    try:
        count = fn()
    except Exception:
        logger.exception("Reaper: retention sweep of %s failed", name)
        continue
    ...
```
If all-or-nothing is intended, at least log *which* helper failed before re-raising.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -499,8 +615,39 @@ def tick(self) -> TickOutcome:
...
+        except Exception:
+            self._discard_owned_sweep_conn()
+            raise
```
muhammad-ali-e: **[High] Sweep failures are not actionable.** This `raise` propagates through `tick()` and is caught in `run()` as the generic `logger.exception("Reaper: cycle failed; continuing")` — the same line emitted for a lease blip, recovery failure, or scheduler failure. A retention sweep can fail every cycle indefinitely with no distinguishable signal, no table name, and no row counts. Contrast `recover_expired_barriers`, which escalates systemic failure to `logger.error` with structured context. Log the sweep failure at this boundary (operation name + ideally a consecutive-failure counter) before re-raising, so a bloated `pg_task_result`/`pg_batch_dedup` in prod is traceable.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -127,6 +137,91 @@ def reaper_interval_from_env() -> float:
...
+    commit connection isn't left in an aborted-txn state.
+    """
+    try:
+        with conn.cursor() as cur:
+            cur.execute("DELETE FROM pg_task_result WHERE expires_at <= now()")
+            deleted = cur.rowcount
+        conn.commit()
+        return deleted
+    except Exception:
+        with contextlib.suppress(Exception):
```
muhammad-ali-e: **[Medium] Suppressed rollback failure leaves no trace.** Re-raising the original DELETE error even if rollback also fails is correct, but if `conn.rollback()` itself raises (broken socket, admin-terminated backend) that signal — the connection is dead — vanishes silently. Same pattern at line 220. Downgrade to a logged best-effort:
```python
except Exception:
    try:
        conn.rollback()
    except Exception:
        logger.debug("Reaper: rollback after failed sweep also failed", exc_info=True)
    raise
```
Keep the broad outer `except` — any driver error must leave the manual-commit connection clean.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -69,6 +70,15 @@
 # Cadence: how often the leader renews + runs recovery. Enforced shorter than
 # the lease window in PgReaper.__init__.
 _DEFAULT_REAPER_INTERVAL_SECONDS = 5.0
+# Retention sweep cadence — far rarer than the tick (cleanup, not recovery), so
+# the per-cycle DELETE doesn't run every few seconds.
+_DEFAULT_SWEEP_INTERVAL_SECONDS = 300.0
+# How long an orphaned ``pg_batch_dedup`` marker lives before the sweep drops it.
+# MUST exceed the longest possible execution so an in-flight marker is never swept
```
muhammad-ali-e: **[Medium] Comment states an unenforced invariant as a guarantee.** "MUST exceed the longest possible execution so an in-flight marker is never swept" — but nothing in code couples `WORKER_PG_DEDUP_RETENTION_SECONDS` to the actual bound, which is the barrier TTL (`barrier_ttl_seconds`, default 6h in `workers/queue_backend/barrier.py`, env-overridable). If an operator raises the barrier TTL above the dedup retention (or lowers retention below it), the invariant silently breaks and an in-flight marker can be swept. Reframe as a configuration contract that names the real bound, e.g.: "Should exceed the barrier TTL (default 6h) — the practical upper bound on an in-flight execution. Not enforced in code; default 24h leaves 4x headroom."

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -127,6 +137,91 @@ def reaper_interval_from_env() -> float:
...
+    return _positive_duration_from_env(
+        "WORKER_PG_DEDUP_RETENTION_SECONDS", _DEFAULT_DEDUP_RETENTION_SECONDS, int
+    )
+
+
+def sweep_expired_results(conn: PgConnection) -> int:
+    """Delete expired executor-RPC result rows; return the number deleted.
+
+    ``pg_task_result`` rows carry ``expires_at = now() + retention`` (written by
+    the consumer's ``store_result``). Once past it no caller can still be waiting
```
muhammad-ali-e: **[Low] Overstated coupling.** "Once past it no caller can still be waiting (the result outlived the caller-timeout)" holds only by default: the store retention (`DEFAULT_RETENTION_SECONDS=3600`) and the caller timeout (`_DEFAULT_TIMEOUT=3600`, overridable via `EXECUTOR_RESULT_TIMEOUT`) are configured independently. If `EXECUTOR_RESULT_TIMEOUT` is raised above retention, a caller *can* still be waiting when the row is swept. Soften: "...has outlived the default caller-timeout (both default to 3600s); these are configured independently, so a caller timeout raised above retention could still be waiting."

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -127,6 +137,91 @@ def reaper_interval_from_env() -> float:
...
+
+
+def sweep_orphan_dedup(conn: PgConnection, retention_seconds: int) -> int:
+    """Delete orphaned per-batch dedup markers older than *retention_seconds*.
+
+    ``pg_batch_dedup`` markers are normally cleared on barrier teardown / reaper
+    barrier-recovery, but a partial-failure execution can leave them behind. A
+    marker only matters while its execution is in flight, so anything older than
+    the longest possible execution (*retention_seconds*) is a safe-to-drop orphan.
+    Time-based on ``created_at`` (the table has no per-row expiry column).
```
muhammad-ali-e: **[Low] Surface the seq-scan.** Unlike `sweep_expired_results` (which advertises `pg_task_result_expires_idx`), this DELETE filters on `pg_batch_dedup.created_at`, which is deliberately **unindexed** (see `backend/pg_queue/models.py` — "would sweep by this column's age and add an index then"). So every sweep seq-scans the table. Acceptable at the 5-min cadence on a normally-near-empty table, but the asymmetry is worth a sentence here: "`created_at` is unindexed, so this is a seq-scan; add an index if the dedup table grows."

muhammad-ali-e (commented):
Automated PR Review Toolkit pass (code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier).

The change is correct, well-tested on the happy paths, and consistent with the existing reaper patterns. No blocking bugs. The findings below are mostly resilience/observability hardening around the two-step sweep, a few comment-accuracy fixes (unenforced invariants stated as guarantees), and some targeted test gaps. Prioritised summary is in the PR thread.

docker/docker-compose.yaml
```diff
@@ -837,6 +837,12 @@ services:
       - APPLICATION_NAME=unstract-worker-pg-reaper
       - WORKER_PG_REAPER_INTERVAL_SECONDS=${WORKER_PG_REAPER_INTERVAL_SECONDS:-5}
       - WORKER_PG_REAPER_HEALTH_PORT=8086
+      # Retention sweep (cadence-gated inside the leader's tick): drop expired
+      # pg_task_result rows + orphaned pg_batch_dedup markers so neither grows
+      # unbounded as the gate ramps. Dedup retention must exceed the longest
```
muhammad-ali-e: **[Low] Ambiguous parenthetical.** "Dedup retention must exceed the longest execution (default 24h)" reads as if 24h is the default *longest execution*, but 24h is the default *dedup retention* (the env var two lines down). The real longest-execution bound is the 6h barrier TTL. Clarify: "Dedup retention (default 24h, below) should exceed the barrier TTL (default 6h) — the upper bound on an in-flight execution — so an in-flight marker is never swept."

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -499,8 +615,39 @@ def tick(self) -> TickOutcome:
...
+        expired ``pg_task_result`` rows and orphaned ``pg_batch_dedup`` markers so
+        neither table grows unbounded as the gate ramps. The cadence is advanced
+        BEFORE sweeping so a failure waits one interval before retry rather than
+        hammering the DB every tick; the owned sweep conn is discarded on error
+        (parity with recovery/schedules).
+        """
+        now = time.monotonic()
+        if now - self._last_sweep_monotonic < self._sweep_interval:
+            return
+        self._last_sweep_monotonic = now
```
muhammad-ali-e, line 670: **[Test gap] Cadence-advance-before-sweep is load-bearing but untested.** Advancing the stamp before the `try` is a deliberate throttle (docstring: "a failure waits one interval before retry rather than hammering the DB"). If a future refactor moves this into/after the `try`, a persistently-failing DB would get a DELETE every ~5s tick instead of every 300s — and no test would catch it. Add: first leader tick makes the sweep raise (assert called once); second tick within the interval asserts the sweep was **not** retried. Also pin the "first tick sweeps immediately" contract by asserting `_last_sweep_monotonic == 0.0` post-construction.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -401,6 +498,25 @@ def __init__(
...
             )
+        # Retention-sweep cadence + dedup-orphan horizon (validated like interval:
+        # an injected non-positive value reaches here unvalidated by the env parser).
+        self._sweep_interval = (
+            sweep_interval_seconds
+            if sweep_interval_seconds is not None
+            else reaper_sweep_interval_from_env()
+        )
+        if self._sweep_interval <= 0:
+            raise ValueError("sweep_interval_seconds must be positive")
```
muhammad-ali-e, line 535: **[Test gap] Injected-knob constructor guards are untested.** These `ValueError` guards (509, 516) exist specifically because an injected `sweep_interval_seconds`/`dedup_retention_seconds` bypasses the env parser (per the comment at 501-502) — yet `TestConstruction` only covers the analogous `interval_seconds` guard. From the suite's view these two raises are dead code. Add cases: `PgReaper(..., sweep_interval_seconds=0)` and `dedup_retention_seconds=-1` each raise `ValueError`.

workers/tests/test_pg_reaper.py
```diff
@@ -249,6 +291,112 @@ def test_scheduler_error_discards_owned_conn(self, stub_scheduler_tick):
...
+    def test_sweep_orphan_dedup_sql(self):
+        conn, cur = self._conn_cur(2)
+        assert sweep_orphan_dedup(conn, 999) == 2
+        args = cur.execute.call_args[0]
+        assert "DELETE FROM pg_batch_dedup" in args[0]
+        assert "created_at <= now() - make_interval" in args[0]
+        assert args[1] == (999,)  # the retention param is bound, not interpolated
+        conn.commit.assert_called_once()
+
+    def test_sweep_rolls_back_on_error(self):
```
muhammad-ali-e: **[Test gap] `sweep_orphan_dedup` rollback path is never exercised.** This test only calls `sweep_expired_results`. `sweep_orphan_dedup` has its own independent try/except/rollback block (reaper.py:219-222) that could be deleted entirely and every test would still pass — and a missed rollback on a manual-commit connection poisons all subsequent ticks (aborted-txn state). Parametrize this test over both helpers, or add a mirror `test_sweep_orphan_dedup_rolls_back_on_error`. Bonus: no test exercises the `dedup`-fails-after-`results`-succeeds path, which is exactly the partial-skip scenario above.

workers/tests/test_pg_reaper.py
```diff
@@ -62,6 +81,29 @@ def test_invalid_raises(self, monkeypatch, bad):
...
 
 
+class TestSweepEnv:
+    def test_sweep_interval_default_and_override(self, monkeypatch):
+        monkeypatch.delenv("WORKER_PG_REAPER_SWEEP_SECONDS", raising=False)
+        assert reaper_sweep_interval_from_env() == pytest.approx(300.0)
+        monkeypatch.setenv("WORKER_PG_REAPER_SWEEP_SECONDS", "30")
+        assert reaper_sweep_interval_from_env() == pytest.approx(30.0)
+
+    def test_dedup_retention_default_and_override(self, monkeypatch):
```
muhammad-ali-e, line 92: **[Test gap] float-vs-int cast distinction unverified.** `_positive_duration_from_env` is parameterized by `cast`: sweep uses `float`, dedup uses `int`. The current values (`"30"`, `"3600"`) parse identically either way, so the distinction is untested. Add: `reaper_sweep_interval_from_env()` returns `300.5` for `"300.5"`; `dedup_retention_from_env()` **raises** for `"3600.5"` (int rejects it) and returns an `int`. The dedup value flows straight into `make_interval(secs => %s)`, so the int contract matters.

workers/tests/test_pg_reaper.py
```diff
@@ -249,6 +291,112 @@ def test_scheduler_error_discards_owned_conn(self, stub_scheduler_tick):
...
+    """The sweep runs leader-only, after recovery + schedule, cadence-gated."""
+
+    def _reaper(self, lease, **kw):
+        kw.setdefault("sweep_interval_seconds", 300)
+        kw.setdefault("dedup_retention_seconds", 86400)
+        return PgReaper(
+            lease, interval_seconds=0.01, sweep_conn=object(), api_client=object(), **kw
+        )
+
+    def test_leader_sweeps(self, stub_retention_sweep):
```
muhammad-ali-e, line 393: **[Test gap + nit] Log-emission boundary untested; assertion re-fetches the conn.** (1) The `if results or dedup:` info-log branch (reaper.py:643) is never taken — both stubs return 0 in every wiring test. Add a no-deletions case (assert no log) and a one-nonzero case (e.g. `results=0, dedup=4`, assert the log fires with both counts) to guard the `or` short-circuit and the two-arg interpolation. (2) Minor: `assert_called_once_with(reaper._get_sweep_conn(), 86400)` calls `_get_sweep_conn()` a second time inside the assertion; capture `reaper._sweep_conn` once instead — more robust if conn handling ever wraps.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -401,6 +498,25 @@ def __init__(
...
+        self._dedup_retention = (
+            dedup_retention_seconds
+            if dedup_retention_seconds is not None
+            else dedup_retention_from_env()
+        )
+        if self._dedup_retention <= 0:
+            raise ValueError("dedup_retention_seconds must be positive")
+        # 0 → "never swept", so the first leader tick sweeps immediately; advanced
+        # to monotonic() each sweep so the cadence holds thereafter.
+        self._last_sweep_monotonic = 0.0
```
muhammad-ali-e: **[Low / type-design, optional] Sentinel overloads the type.** `_last_sweep_monotonic = 0.0` means "never swept," relying on `time.monotonic()` never returning ~0 — an implicit, untyped assumption. `float | None` with `None` = "never" expresses the intent and drops the assumption: the gate becomes `if last is not None and now - last < interval`. Low cost; take or leave. (Code-simplifier and type-analyzer both judged the sweep helpers, env wrappers, and `_N` TypeVar already at the right altitude — no change recommended there.)

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (`beb96b005`). The two sweeps now run independently via a new `_run_sweep(table, fn)` — a failure in one no longer skips (and cadence-gates out) the other; each discards the owned conn on error so the sibling reconnects. Test: `test_one_sweep_failing_does_not_starve_the_other`.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (`beb96b005`). `_run_sweep` logs each failure at the boundary with the table name + a consecutive-failure streak (`_sweep_fail_streak`), and no longer re-raises (a cleanup failure must not fail the tick) — distinct from the generic `run()` cycle-failure log.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (`beb96b005`). New `_rollback_after_sweep_failure(conn, table)` logs a rollback that itself raises (broken socket / terminated backend) instead of suppressing it, while still re-raising the ORIGINAL DELETE error. Test: `test_rollback_failure_is_logged_and_original_error_raised`.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (`beb96b005`). Reworded the comment to state it's operator-enforced (not coupled in code to the actual execution bound).

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (`beb96b005`). Softened to 'no caller is still waiting BY DEFAULT' (store retention == caller-timeout default) and noted that raising the caller timeout above the retention could sweep a result mid-wait → size retention >= the longest caller timeout.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (`beb96b005`). Added a sentence to the docstring: `created_at` is unindexed so each sweep seq-scans — fine at the 5-min cadence on a near-empty table; add an index if it grows.

docker/docker-compose.yaml
muhammad-ali-e: Fixed (`beb96b005`). Reworded so 24h clearly attaches to the dedup retention, not the longest execution.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (`beb96b005`). Added `test_failing_sweep_still_advances_cadence` — a failing sweep advances the stamp, so a second immediate tick is gated out (asserts call_count == 1).

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed (`beb96b005`). Added `test_non_positive_sweep_interval_rejected` + `test_non_positive_dedup_retention_rejected` for the injected-knob `ValueError` guards.

workers/tests/test_pg_reaper.py
muhammad-ali-e: Fixed (`beb96b005`). Parametrized `test_sweep_rolls_back_on_error` over both `sweep_expired_results` and `sweep_orphan_dedup`.

workers/tests/test_pg_reaper.py
muhammad-ali-e: Fixed (`beb96b005`). `test_cast_distinction_float_vs_int`: sweep parses `"1.5"` → 1.5 (float); dedup `"1.5"` → ValueError (int).

workers/tests/test_pg_reaper.py
muhammad-ali-e: Fixed (`beb96b005`). Added `test_logs_counts_only_when_rows_deleted` (results=0, dedup=4 → log fires with both counts) + `test_no_log_when_nothing_deleted`. Also captured `reaper._sweep_conn` once in `test_leader_sweeps` instead of re-fetching.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Done (`beb96b005`). `_last_sweep_monotonic` is now `float | None` with `None` = 'never swept'; the gate is `if last is not None and now - last < interval` — drops the implicit 'monotonic never returns ~0' assumption.

muhammad-ali-e:
**Review round addressed** — `beb96b005`. Per-thread replies on each. (SonarCloud already green on this PR.)

**High**
- *Partial-sweep starvation:* the two sweeps now run independently via a new `_run_sweep(table, fn)` — a fault in one no longer skips and then cadence-gates out the other; each discards the owned conn on error so the sibling reconnects.
- *Non-actionable failures:* `_run_sweep` logs each failure at the boundary with the table name + a consecutive-failure streak, and no longer propagates (cleanup must not fail the tick) — distinct from the generic cycle-failure log.

**Medium**
- Rollback that itself fails is now logged (not suppressed) via `_rollback_after_sweep_failure`, while still re-raising the original DELETE error.
- Reworded the dedup-retention comment as operator-enforced (not code-coupled).

**Low / wording**
- Softened the result-retention coupling claim (holds by default); documented the `pg_batch_dedup.created_at` seq-scan; de-ambiguated the compose comment.

**Type-design**
- `_last_sweep_monotonic` is now `float | None` ("never"), dropping the `0.0` sentinel assumption.

**Test gaps** (now +8, **35 green**): float-vs-int cast distinction, injected-knob constructor guards, `sweep_orphan_dedup` rollback (parametrized) + rollback-itself-fails log, one-sweep-failing-doesn't-starve-the-other, failing-sweep-still-advances-cadence, log-emission boundary (counts only when rows deleted / silent otherwise) + the `_get_sweep_conn()` re-fetch nit.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -127,6 +139,113 @@ def reaper_interval_from_env() -> float:
...
+    """
+    raw = os.getenv(name)
+    if raw is None:
+        return default
+    try:
+        value = cast(raw)
+    except ValueError as exc:
+        raise ValueError(
+            f"{name}={raw!r} is not a number. Unset it to default to {default}."
+        ) from exc
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed (`d33883796`). `_positive_duration_from_env` now raises `{name}={raw!r} cannot be parsed: {exc}. …` so an int knob given "1.5" reports the real cause instead of the misleading "is not a number". Test asserts the message (`match="cannot be parsed"`).

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -499,8 +643,68 @@ def tick(self) -> TickOutcome:
...
+            count = fn(self._get_sweep_conn())
+        except Exception:
+            streak = self._sweep_fail_streak.get(table, 0) + 1
+            self._sweep_fail_streak[table] = streak
+            logger.exception(
+                "Reaper: retention sweep of %s failed (%s consecutive) — will retry "
+                "next cycle",
+                table,
+                streak,
+            )
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed (`d33883796`). Reworded to "will retry after the next sweep interval" — the retry is gated by `_sweep_interval` (~300s default), not the ~5s tick.

PR #2102, opened 2026-06-22
UN-3607 [REFACTOR] retire the executor_rpc mirror — shared dispatcher + injected transport

muhammad-ali-e:
## What

Retires the byte-for-byte **backend ↔ workers `executor_rpc.py` mirror** that SonarCloud has flagged on every executor PR. The two carried identical dispatch logic (gate + reply_key/timeout orchestration + routing + never-raises); the *only* real difference is the transport primitive — backend enqueues via the Django ORM, workers via psycopg2. Targets `feat/UN-3445-pg-queue-integration` (not `main`).

Net **−417 lines** (823 added / 1240 removed).

## How — composition (injected transport), not inheritance

The shared logic now lives once in **`unstract.workflow_execution.executor_rpc`** (both backend and workers already depend on it; it can't live in `unstract.core` because `sdk1` imports `core` → circular). The differing primitive is injected via a `QueueTransport` Protocol:

- **shared**: `PgExecutionDispatcher` (concrete — `dispatch` / `dispatch_async` / `dispatch_with_callback` + the never-raises orchestration) calling `transport.enqueue` / `wait_for_result`; `ExecResultRow` (normalised result row so the Django-model and dict rows fold to one path); `resolve_pg_transport` (master-gate value supplied by the caller, then the Flipt flag); `RoutingExecutionDispatcher` (per-call PG-vs-Celery, with `celery`/`pg`/`resolve` injected).
- **backend adapter**: `DjangoQueueTransport` (`enqueue_task` + `PgTaskResult` poll) + `settings.PG_QUEUE_TRANSPORT_ENABLED` master-gate + factory.
- **workers adapter**: `PgClientQueueTransport` (`PgQueueClient` + `PgResultBackend`) + env master-gate + factory.
- `DispatchHandle` / `signature_to_continuation` stay in `unstract.core` (already shared).

Both call sites (prompt-studio, `structure_tool`) use the unchanged `get_executor_dispatcher` factory.

## Non-regression

The dispatch logic is the *same code*, relocated and transport-injected — behaviour (gate, routing, never-raises) is byte-identical.

## Tests

The dispatch contract + routing + the Flipt gate matrix are tested **once** against a fake transport (workers suite); each side's tests shrink to its adapter (enqueue/wait shape) + factory wiring. Backend **7** + workers **36** green; the 53 other executor-touching tests pass.

## Dev-tested live (gate ON)

Rebuilt the worker image, recreated the PG workers, restarted the backend — all boot clean (the new shared module imports in the built image; no `ImportError`). Then ran a Prompt Studio prompt **and** an ETL:

- **Prompt Studio (backend adapter):** `PG executor dispatch_with_callback: enqueued … on_success=ide_prompt_complete` → 202.
- **ETL (workers adapter, blocking RPC):** `Executor RPC → PG transport` → `PG executor dispatch: enqueued reply_key=… waiting for result` → executor `Received execute_extraction … operation=structure_pipeline` → `execute_extraction complete … success=True`.

No tracebacks, no "No executor registered". Both transports drive the shared dispatcher and route over PG end-to-end.

### DIFF
--- a/backend/pg_queue/executor_rpc.py
+++ b/backend/pg_queue/executor_rpc.py
@@ -1,67 +1,49 @@
-"""Executor-RPC transport routing for the PG path (Phase 9).
+"""Executor-RPC for the PG path — backend (Django) transport adapter.
 
-The executor "RPC" is a synchronous request-reply: a caller (prompt-studio here)
-sends an ``ExecutionContext`` to the executor worker and blocks for the
-``ExecutionResult``. The legacy transport is Celery — the SDK
-``ExecutionDispatcher`` (``send_task`` + ``AsyncResult.get``). This module adds a
-**parallel** Postgres transport that leaves Celery and the SDK completely
-untouched (no SDK edit, no change to the ``execute_extraction`` task or the
-Celery executor worker):
+The gate + reply_key/timeout orchestration + routing live ONCE in
+``unstract.workflow_execution.executor_rpc`` (shared with the workers). This module
+is the thin Django half: a :class:`DjangoQueueTransport` that enqueues via the ORM
+(``enqueue_task``) and polls ``PgTaskResult``, plus the per-side gate (master switch =
+``settings.PG_QUEUE_TRANSPORT_ENABLED``) and the :func:`get_executor_dispatcher`
+factory that wires them together.
 
-- :class:`PgExecutionDispatcher` enqueues ``execute_extraction`` onto the PG queue
-  with a unique ``reply_key`` and polls ``pg_task_result`` for the reply — same
-  ``.dispatch()`` contract as the SDK dispatcher (never raises; failure/timeout →
-  ``ExecutionResult.failure``).
-- :func:`resolve_executor_transport` is the gate: master
-  ``PG_QUEUE_TRANSPORT_ENABLED`` then the **single** Flipt flag
-  ``pg_queue_enabled`` — the same flag the execution path uses, so one
-  flip turns the whole PG-queue feature on/off (no per-subsystem flags to
-  maintain). Fails closed to Celery.
-- :class:`RoutingExecutionDispatcher` is what callers get from
-  ``PromptStudioHelper._get_dispatcher()``: ``dispatch()`` picks PG-vs-Celery
-  **per call** (read at dispatch time → flipping the flag is an instant,
-  no-redeploy rollout/rollback); ``dispatch_async`` / ``dispatch_with_callback``
-  always delegate to Celery (the callback path is a later slice).
-
-Zero-regression: gate off ⇒ every method delegates to the unchanged Celery
-``ExecutionDispatcher`` and no ``pg_task_result`` row is created.
+Zero-regression: gate off ⇒ the routing dispatcher delegates every mode to the
+unchanged Celery ``ExecutionDispatcher`` and no ``pg_task_result`` row is created.
 """
 
 from __future__ import annotations
 
-import logging
-import os
 import time
-import uuid
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
 
 from django.conf import settings
 from django.db import close_old_connections
 
-from pg_queue.flags import PG_QUEUE_FLAG_KEY
 from pg_queue.models import PgTaskResult
 from pg_queue.producer import enqueue_task
-from unstract.core.data_models import PgTaskStatus
-from unstract.core.execution_dispatch import DispatchHandle, signature_to_continuation
-from unstract.flags.feature_flag import check_feature_flag_status
 from unstract.sdk1.execution.dispatcher import ExecutionDispatcher
-from unstract.sdk1.execution.result import ExecutionResult
+from unstract.workflow_execution.executor_rpc import (
+    EXECUTE_TASK,
+    ExecResultRow,
+    PgExecutionDispatcher,
+    RoutingExecutionDispatcher,
+    resolve_pg_transport,
+)
 
 if TYPE_CHECKING:
+    from unstract.core.data_models import ContinuationSpec
     from unstract.sdk1.execution.context import ExecutionContext
 
-logger = logging.getLogger(__name__)
+# Re-exported so existing ``from pg_queue.executor_rpc import …`` imports keep working.
+__all__ = [
+    "DjangoQueueTransport",
+    "PgExecutionDispatcher",
+    "RoutingExecutionDispatcher",
+    "get_executor_dispatcher",
+    "resolve_executor_transport",
+]
 
-# Gating reads the single shared PG-queue flag (pg_queue.flags.PG_QUEUE_FLAG_KEY,
-# imported above) — the same key execution and the scheduler use.
-_EXECUTE_TASK = "execute_extraction"
-# Mirror the SDK's queue-per-executor convention so the PG executor queue name
-# matches the Celery one (the queue routes by the row's queue_name column).
-_QUEUE_PREFIX = "celery_executor_"
-# Caller-side wait default — mirrors the SDK dispatcher (EXECUTOR_RESULT_TIMEOUT
-# env, else 3600s) so a PG-routed caller waits exactly as long as a Celery one.
-_DEFAULT_TIMEOUT_ENV = "EXECUTOR_RESULT_TIMEOUT"
-_DEFAULT_TIMEOUT = 3600
+# Poll cadence for the result wait (PgBouncer-safe; no LISTEN/NOTIFY).
 _POLL_INITIAL_SECONDS = 0.2
 _POLL_MAX_SECONDS = 2.0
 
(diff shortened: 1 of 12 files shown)


muhammad-ali-e:
Heads-up: this PR also carries one small **unrelated** commit (`c8219ef0c`) — `frontend/vite.config.js` adds `ws: true` to the local Vite dev proxy so the Socket.IO log/result channel (`/api/v1/socket`, websocket-only transport) actually reaches the backend in local dev. ws-proxying was dropped in the CRA→Vite migration; this restores it.

**Deployment-safe:** `server.proxy` runs only under `vite dev` — staging/prod serve a built bundle behind nginx/Traefik (which already routes `/api/v1/socket`), so it has zero effect on any deployed env. Happy to split it into its own PR if preferred.

workers/tests/test_executor_rpc.py
```diff
@@ -1,423 +1,341 @@
...
-            PgExecutionDispatcher().dispatch(self._ctx())  # must not raise
-        assert seen["timeout"] == 3600  # _DEFAULT_TIMEOUT
+    def test_dispatch_async_is_fire_and_forget(self):
+        t = _FakeTransport()
+        task_id = PgExecutionDispatcher(t).dispatch_async(_ctx())
+        (kw,) = t.enqueue_calls
+        assert kw["task_id"] == task_id and kw.get("reply_key") is None
+        assert kw.get("on_success") is None and kw.get("on_error") is None
+
+    def test_dispatch_with_callback_translates_signatures(self):
```
muhammad-ali-e: **High — lost coverage: `on_error` translation + default `task_id` are no longer tested.**

This is the only `dispatch_with_callback` test and it passes *only* `on_success` plus an explicit `task_id="tid-7"`. The old mirror suites covered two behaviours that are now untested:

1. `on_error` → `signature_to_continuation` → forwarded as `on_error=` (shared module `executor_rpc.py:306,312`). A regression dropping `on_error` would pass CI — and on-error continuations are how PG-path failures notify the IDE/callback.
2. auto-generated `task_id` when omitted (`task_id or str(uuid.uuid4())`, shared module `:302`).

**Fix:** also pass `on_error=<second MagicMock signature>` and assert `kw["on_error"]["task_name"]`; add a `test_dispatch_with_callback_defaults_task_id` that calls with no `task_id` and asserts `handle.id` is truthy and equals `t.enqueue_calls[0]["task_id"]`.

unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
```diff
@@ -0,0 +1,390 @@
...
+    defaults to ``None`` — the never-raises contract must not depend on the producer
+    having written every key.
+    """
+
+    status: str | None = None
+    result: dict | None = None
+    error: str | None = None
+
+
+class QueueTransport(Protocol):
```
muhammad-ali-e, line 110: **High (highest-leverage type fix) — declare conformance to this Protocol on both transports.**

`QueueTransport` is a structural `Protocol`, but neither `DjangoQueueTransport` (`backend/pg_queue/executor_rpc.py:62`) nor `PgClientQueueTransport` (`workers/queue_backend/pg_queue/executor_rpc.py:62`) inherits from it. Conformance is therefore only checked at the `PgExecutionDispatcher(...)` construction site — a signature drift on one side (e.g. renaming `org_id`) wouldn't surface until runtime, on that side only.

**Fix:** have both transport classes explicitly subclass `QueueTransport`. Protocols support explicit inheritance at zero runtime cost, and mypy/pyright will then check each implementation against the seam independently. Cheapest way to keep the seam well-typed across both trees.

unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
```diff
@@ -0,0 +1,390 @@
...
+        carried in the payload; after the executor runs, the consumer self-chains the
+        matching continuation. Returns a :class:`DispatchHandle` exposing ``.id``
+        (== ``task_id``) so call sites read the task id exactly as on the Celery path.
+        """
+        task_id = task_id or str(uuid.uuid4())
+        queue = f"{QUEUE_PREFIX}{context.executor_name}"
+        org = str(getattr(context, "organization_id", "") or "")
+        success_spec = signature_to_continuation(on_success)
+        error_spec = signature_to_continuation(on_error)
+        self._transport.enqueue(
```
muhammad-ali-e: **Medium — on an enqueue failure here, the carried `on_error` continuation never fires.**

Unlike `dispatch` (guarded, never raises), `dispatch_with_callback` calls `self._transport.enqueue(...)` unguarded, so a transport error (DB down, `psycopg2.OperationalError`, JSON-serialization error in `to_payload`) propagates raw. This matches the SDK parity contract, so it's not a regression — but note the hazard: when continuations are carried *in the payload* and the enqueue itself fails, the executor never runs and the `on_error` continuation is **never fired**. Cleanup only happens because call sites wrap dispatch in their own try/except (e.g. `prompt_studio/.../views.py:461`).

**Fix (no behaviour change):** log-before-raise so the failure is observable in Sentry even when a caller swallows it, and add a one-line note to the docstring stating callers MUST treat a raised enqueue as the on_error path because the carried `on_error` will not fire:
```python
try:
    self._transport.enqueue(...)
except Exception:
    logger.exception("PG executor dispatch_with_callback: enqueue failed (executor=%s run_id=%s)", context.executor_name, context.run_id)
    raise
```

unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
```diff
@@ -0,0 +1,390 @@
...
+                "PG executor dispatch: TIMEOUT after %ss (reply_key=%s run_id=%s) — "
+                "the executor task may still be running",
+                timeout,
+                reply_key,
+                context.run_id,
+            )
+            return ExecutionResult.failure(
+                error=f"TimeoutError: executor reply not received within {timeout}s"
+            )
+        if row.status == PgTaskStatus.COMPLETED.value and row.result is not None:
```
muhammad-ali-e, line 275: **Medium — a `COMPLETED` row with `result is None` is silently reported as a generic task failure.**

The success branch requires `status == COMPLETED.value AND result is not None`; anything else falls through to `return ExecutionResult.failure(error=row.error or "executor task failed")` at line 265. So a producer bug (writes `COMPLETED` but no `result`, or a typo status) is indistinguishable from a genuine executor failure — there's no diagnostic that the *row* was malformed. `ExecResultRow.status` is also `str | None` (line 78), so the enum invariant is never enforced at the read boundary.

**Fix:** add an explicit branch mirroring the malformed-parse handling at lines 242-257 — if `status == COMPLETED.value and row.result is None`, log + return a "malformed completed row (missing result)" failure. Optionally validate `status` against `PgTaskStatus` and warn on an unknown value. Keeps the fail-safe direction while making the producer-contract violation observable.

unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
```diff
@@ -0,0 +1,390 @@
...
+            task_id,
+            queue,
+            context.run_id,
+        )
+        return task_id
+
+    def dispatch_with_callback(
+        self,
+        context: ExecutionContext,
+        on_success: Any | None = None,
```
muhammad-ali-e: **Medium — `on_success`/`on_error` typed `Any | None` discards the real `Signature` contract.**

`signature_to_continuation` reads `.task`, `.options["queue"]`, `.args`, `.kwargs` and raises `ValueError` if `.task`/queue are missing — so the real precondition is "a Celery-`Signature`-like object." Typing these `Any` is actually *weaker* than the SDK method this mirrors (`sdk1/.../dispatcher.py` types them `Signature | None`), and lets any value through to a runtime `ValueError`.

**Fix:** define a small structural `Protocol` (e.g. `class CallbackSignature(Protocol): task: str; options: dict; args: tuple; kwargs: dict`) next to `signature_to_continuation` in `unstract.core.execution_dispatch`, and type both that helper and these params as `CallbackSignature | None`. Documents and checks the precondition at the seam.

unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
```diff
@@ -0,0 +1,390 @@
...
+        self._transport = transport
+
+    def dispatch(
+        self,
+        context: ExecutionContext,
+        timeout: int | None = None,
+    ) -> ExecutionResult:
+        # No ``headers`` on any PG dispatch method: the PG path carries org/routing
+        # via the enqueue payload, not Celery headers, and the
+        # RoutingExecutionDispatcher strips fairness headers before delegating here.
```
muhammad-ali-e: **Medium (comment accuracy) — "strips fairness headers" is misleading; nothing is stripped.**

The `RoutingExecutionDispatcher` (lines 348-363) does not mutate/remove anything — it receives `headers` and simply does not forward them to `self._pg.dispatch(...)`. "Strips" implies active removal and could send a maintainer hunting for header-mutation logic that doesn't exist. Note the parallel comment at line 361 ("intentionally not forwarded here") is already phrased correctly.

**Fix:** reword to match line 361, e.g. "...and the `RoutingExecutionDispatcher` intentionally does not forward fairness headers to this method."

muhammad-ali-e (commented):
## Automated PR review — UN-3607 shared executor-RPC dispatcher

Reviewed with the PR Review Toolkit (code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier). Scope: the 7 non-lockfile files in this PR, diffed against `feat/UN-3445-pg-queue-integration`.

**Overall: a clean, behaviour-preserving refactor.** The two byte-for-byte mirrors are now thin transport adapters over one shared dispatcher; the never-raises contract and fail-closed gating are faithfully preserved, and the backend path actually *gains* a `wait_for_result` try/except it previously lacked. No correctness regressions found.

Findings below are inline. Priority summary:
- **High:** lost test coverage for `dispatch_with_callback` `on_error`/default-`task_id`; transport classes don't declare `QueueTransport` conformance.
- **Medium:** enqueue-failure drops the carried `on_error` continuation; `status: str` COMPLETED-but-null gap; callback params typed `Any`; an inaccurate "strips" comment; missing `dispatch` docstring; residual poll-loop duplication; backend poll-loop untested.
- **Low:** two confirmed ruff failures (I001, D209); silent master-gate env parse; `celery: Any`; `dispatch_async` propagation untested.

Nothing here is blocking.

unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
```diff
@@ -0,0 +1,390 @@
...
+    ``reply_key`` and blocks on the result row until the executor consumer records it
+    or the timeout elapses. Honours the SDK dispatcher contract: it never raises and
+    converts a timeout/failure into ``ExecutionResult.failure`` so callers branch on
+    ``result.success`` identically on either transport.
+    """
+
+    def __init__(self, transport: QueueTransport) -> None:
+        self._transport = transport
+
+    def dispatch(
```
muhammad-ali-e, line 210: **Medium (docs) — the primary public entry point `dispatch` has no docstring.**

`dispatch` is the main public method of the new shared API, yet it opens with an implementation-detail inline comment and no docstring, while its siblings `dispatch_async` (line 268) and `dispatch_with_callback` (line 294) both have one.

**Fix:** add a short docstring stating the contract: request-reply via a unique `reply_key`, blocks up to `timeout`, never raises (timeout/failure → `ExecutionResult.failure`), and that `headers` is intentionally omitted vs the SDK signature. The inline comment at 188-190 can fold into it.

backend/pg_queue/executor_rpc.py
```diff
@@ -1,284 +1,107 @@
...
-        be called inside an open transaction** (``transaction.atomic`` /
-        ``ATOMIC_REQUESTS`` would pin one snapshot and never see the new row).
+        The DB connection is released between polls (``close_old_connections``) so a
+        long-running RPC does not pin a backend connection and exhaust the pool. Each
+        poll is its own autocommit query, so a row committed by the executor consumer
+        becomes visible — **dispatch must NOT be called inside an open transaction**
+        (``transaction.atomic`` / ``ATOMIC_REQUESTS`` would pin one snapshot and never
+        see the new row).
         """
         deadline = time.monotonic() + timeout
```
muhammad-ali-e: **Medium (de-dup) — this poll loop re-implements `PgResultBackend.wait_for_result` byte-for-byte.**

The exponential-backoff loop here (deadline, `delay`, `min(delay, remaining)`, `min(delay*2, _POLL_MAX_SECONDS)`) and the constants `_POLL_INITIAL_SECONDS`/`_POLL_MAX_SECONDS` (lines 47-48) duplicate the algorithm already in `workers/queue_backend/pg_queue/result_backend.py:172-182`. Since de-duplication is the whole point of this PR, this is the one meaningfully duplicated *logic* that survived. The loop bodies differ only by the `close_old_connections()` hook and the row-fetch.

**Fix (behaviour-preserving):** extract a generic `poll_for_row(fetch, timeout, *, between_polls=None, initial=0.2, maximum=2.0)` into the shared module; the backend adapter becomes a 3-line `fetch`/`poll_for_row(fetch, timeout, between_polls=close_old_connections)`. The workers side already delegates to `PgResultBackend`, so it needn't change.

backend/pg_queue/tests/test_executor_rpc.py
```diff
@@ -1,302 +1,109 @@
...
+            DjangoQueueTransport().enqueue(
+                queue="celery_executor_legacy", context=_ctx(), org_id="o",
+                on_success=spec, task_id="tid-7",
+            )
+        kw = enq.call_args.kwargs
+        assert kw["on_success"] == spec and kw["task_id"] == "tid-7"
+        assert kw["reply_key"] is None  # callback dispatch, not request-reply
+
+
+class TestDjangoQueueTransportWait:
```
muhammad-ali-e, line 60: **Medium — the backend poll loop's multi-iteration path is untested.**

These two tests cover only the boundaries: row present on poll 1 (`test_present_row_folds_to_exec_result_row`) and `timeout=0` → `None`. Neither exercises the load-bearing path in `executor_rpc.py:97-111`: row absent then present on a later poll, `close_old_connections()` actually firing between polls (the connection-release invariant the docstring sells), or the backoff being bounded by `remaining`. A regression that stopped calling `close_old_connections()` (re-pinning a connection through the sleep) or inverted the `remaining <= 0` guard into an infinite loop would not be caught.

**Fix:** add a test with `qs.filter.return_value.first.side_effect = [None, row]`, patch `time.sleep` and `close_old_connections` (assert it's called once), and assert the second poll returns the folded `ExecResultRow`.

backend/pg_queue/executor_rpc.py
```diff
@@ -1,284 +1,107 @@
...
 import time
-import uuid
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
 
 from django.conf import settings
 from django.db import close_old_connections
 
-from pg_queue.flags import PG_QUEUE_FLAG_KEY
 from pg_queue.models import PgTaskResult
```
muhammad-ali-e, line 21: **Low — confirmed ruff `I001` failure (CI enforces `I`).** The first-party `pg_queue.*` imports (lines 22-23) precede the third-party `unstract.*` imports (lines 24-31); isort wants the local group last. Verified failing via `ruff check`.

**Fix:** run `ruff check --fix` (or move the two `from pg_queue...` lines below the `unstract.workflow_execution` import block).

workers/tests/test_executor_rpc.py
```diff
@@ -1,423 +1,341 @@
...
-    @staticmethod
-    def _ctx() -> MagicMock:
-        c = MagicMock()
-        c.executor_name = "legacy"
-        c.run_id = "r"
-        c.organization_id = "o"
-        c.to_dict.return_value = {"run_id": "r"}
-        return c
+class _FakeTransport:
+    """Records ``enqueue`` calls and returns a configured result for the poll —
```
muhammad-ali-e: **Low — confirmed ruff `D209` failure (CI enforces `D`).** A multi-line docstring must have its closing `"""` on its own line. Verified failing via `ruff check`.

**Fix:** move the closing quotes to a new line (or `ruff check --fix`):
```python
    """Records ``enqueue`` calls and returns a configured result for the poll —
    so the shared dispatcher's logic is tested without any DB.
    """
```

workers/queue_backend/pg_queue/executor_rpc.py
```diff
@@ -1,328 +1,84 @@
...
-        )
-    except Exception:
-        logger.warning(
-            "resolve_executor_transport: Flipt check failed; using Celery",
-            exc_info=True,
-        )
-        return False
-    return bool(enabled)
-
+    master = os.environ.get(_MASTER_GATE_ENV, "false").lower() == "true"
```
muhammad-ali-e: **Low — the master-gate env parse fails silently on a fat-fingered value.**

`os.environ.get(_MASTER_GATE_ENV, "false").lower() == "true"` resolves anything other than exactly `"true"` (`"1"`, `"yes"`, `"on"`, `"True "` with whitespace) to `False`, leaving the whole PG transport off with no log line. An operator rolling out with `PG_QUEUE_TRANSPORT_ENABLED=1` would see it silently do nothing. This is parity with the original (not a regression), but it's the one gate input that fails quietly — contrast the carefully-logged `_resolve_timeout` and the `FLIPT_SERVICE_AVAILABLE` warning.

**Fix:** warn when the var is set but doesn't parse to a recognized token, e.g. if `os.environ.get(_MASTER_GATE_ENV)` is non-empty and not in `{"true","false"}`.

unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
```diff
@@ -0,0 +1,390 @@
...
+    sites are unchanged.
+
+    Composition-injected: the Celery dispatcher (``celery``), the PG dispatcher
+    (``pg``) and the per-side gate (``resolve(context) -> bool``).
+    """
+
+    def __init__(
+        self,
+        *,
+        celery: Any,
```
muhammad-ali-e: **Low — `celery: Any` leaves the three duck-typed delegations unchecked.**

The class is "duck-typed against the SDK `ExecutionDispatcher`" but holds it as `Any`, so the assumption that `celery` exposes `dispatch`/`dispatch_async`/`dispatch_with_callback` is unverified. Both factories already pass a real `ExecutionDispatcher` (`backend/pg_queue/executor_rpc.py:119`, `workers/queue_backend/pg_queue/executor_rpc.py:113`), and this module already imports it is available via `unstract.sdk1.execution.dispatcher`.

**Fix:** type `celery: ExecutionDispatcher` (import it here) — turns three duck-typed delegations into checked ones. Optionally type the routing `dispatch_with_callback` return (line 379) as a `HasId`/`DispatchHandle`-shaped protocol instead of `Any`.

workers/tests/test_executor_rpc.py
```diff
@@ -1,423 +1,341 @@
...
-
-        with (
-            patch.object(PgExecutionDispatcher, "_enqueue"),
-            patch.object(
-                PgExecutionDispatcher, "_wait_for_result", side_effect=fake_wait
-            ),
-        ):
-            PgExecutionDispatcher().dispatch(self._ctx())  # must not raise
-        assert seen["timeout"] == 3600  # _DEFAULT_TIMEOUT
+    def test_dispatch_async_is_fire_and_forget(self):
```
muhammad-ali-e, line 136: **Low — the documented `dispatch_async` "enqueue failures propagate" contract is asserted nowhere.**

The shared `dispatch_async` docstring (`executor_rpc.py:272-273`) deliberately lets enqueue errors propagate — an intentional asymmetry with the never-raises `dispatch`. This test covers only the happy path, so a future "consistency" refactor could wrap the enqueue in try/except and silently break the contract (callers rely on the raise to know the task was never enqueued).

**Fix:** add `test_dispatch_async_propagates_enqueue_failure` using `_FakeTransport(enqueue_raises=RuntimeError(...))` and `pytest.raises(RuntimeError)` around `PgExecutionDispatcher(t).dispatch_async(_ctx())`.

workers/tests/test_executor_rpc.py
muhammad-ali-e: Fixed (`ae5f99749`). `test_dispatch_with_callback_translates_both_signatures` now passes on_error and asserts `kw['on_error']['task_name']`; added `test_dispatch_with_callback_defaults_task_id` (no task_id → uuid on handle == payload task_id).

unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
muhammad-ali-e: Fixed (`ae5f99749`). Both `DjangoQueueTransport` and `PgClientQueueTransport` now explicitly inherit `QueueTransport`, so mypy/pyright check each implementation against the seam independently (zero runtime cost — not @runtime_checkable, just static conformance).

unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
muhammad-ali-e: Fixed (`ae5f99749`, no behaviour change). dispatch_async + dispatch_with_callback now log-before-raise on enqueue failure (observable in Sentry even if a caller swallows it), and the docstring states a raised enqueue IS the failure signal since on_error can't fire (the prompt-studio views already handle this in their try/except).

unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
muhammad-ali-e: Fixed (`ae5f99749`). A COMPLETED row with `result is None` now returns a distinct `Executor reported completion with no result` failure (+ a warning log), not the generic `executor task failed`.

unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
muhammad-ali-e: Fixed (`ae5f99749`). Added a `CallbackSignature(Protocol)` (task/options/args/kwargs) in `unstract.core.execution_dispatch` and typed both `signature_to_continuation` and the `dispatch_with_callback` params with it — celery-free, a real Signature conforms structurally.

unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
muhammad-ali-e: Fixed (`ae5f99749`). Reworded to 'does not forward fairness headers to the PG path' — nothing is mutated/stripped.

unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
muhammad-ali-e: Fixed (`ae5f99749`). `dispatch` now has a proper docstring (request-reply + the never-raises contract); the implementation note moved below it.

backend/pg_queue/executor_rpc.py
muhammad-ali-e: Fixed (`ae5f99749`). Extracted `poll_for_row(fetch, timeout, *, between_polls=, initial=, maximum=)` into the shared module; the backend adapter's wait_for_result is now a 3-line `_fetch` + `poll_for_row(_fetch, timeout, between_polls=close_old_connections)`.

backend/pg_queue/tests/test_executor_rpc.py
muhammad-ali-e: Fixed (`ae5f99749`). Added `test_multi_iteration_poll_misses_then_hits` (miss then hit; asserts close_old_connections + sleep each ran once between polls).

backend/pg_queue/executor_rpc.py
muhammad-ali-e: Addressed (`ae5f99749`). Note: `uvx ruff` (latest) and the repo's pinned pre-commit ruff disagree on first-party classification for this file — pre-commit (the CI authority) reformatted the import order on commit, so it now matches what CI enforces.

workers/tests/test_executor_rpc.py
muhammad-ali-e: Fixed (`ae5f99749`). Moved the `_FakeTransport` docstring closing quotes to their own line (also fixed the same on `_CeleryDispatcher`).

workers/queue_backend/pg_queue/executor_rpc.py
muhammad-ali-e: Fixed (`ae5f99749`). The workers master-gate now warns when `PG_QUEUE_TRANSPORT_ENABLED` is set to anything other than 'true'/'false' (e.g. '1'/'yes'/'on') — no longer a silent no-op.

unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
muhammad-ali-e: Fixed (`ae5f99749`). Added a `_CeleryDispatcher(Protocol)` with the three delegated methods; the router now holds `celery: _CeleryDispatcher` instead of `Any`.

workers/tests/test_executor_rpc.py
muhammad-ali-e: Fixed (`ae5f99749`). Added `test_dispatch_async_propagates_enqueue_failure` (`pytest.raises`) — pins the intentional asymmetry with the never-raises dispatch.

muhammad-ali-e:
**Review round addressed — `ae5f99749`. All 14, per-thread replies on each.** (SonarCloud already green.)

**High**
- Both transports explicitly inherit `QueueTransport` → static conformance checked per-implementation, not just at the construction site.
- Restored the lost `dispatch_with_callback` coverage: on_error translation + default-task_id.

**Medium**
- `CallbackSignature` Protocol types `signature_to_continuation` + the callback params (was `Any`); `_CeleryDispatcher` Protocol types the router's celery (was `Any`).
- Extracted the surviving duplicated poll loop into a shared `poll_for_row(...)`; the backend adapter is now a 3-line fetch + delegate.
- `dispatch` docstring; enqueue log-before-raise + a docstring note that a raised enqueue is the failure signal (on_error can't fire); a COMPLETED-but-no-result row is now a distinct error; reworded "strips headers" → "does not forward".

**Low**
- Workers master-gate warns on a fat-fingered value (no longer a silent no-op); `celery: Any` → `_CeleryDispatcher`; I001 (pinned-ruff order, the CI authority) + D209 fixed; added `dispatch_async`-propagates and backend multi-iteration-poll tests.

Backend 8 + workers 38 green; ruff/pre-commit clean.

unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
```diff
@@ -0,0 +1,501 @@
...
+        row = fetch()
+        if row is not None:
+            return row
+        remaining = deadline - time.monotonic()
+        if remaining <= 0:
+            return None
+        if between_polls is not None:
+            between_polls()
+        time.sleep(min(delay, remaining))
+        delay = min(delay * 2, maximum)
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed (`b7b24a7ec`). Unified the loop: `poll_for_row` is pure (no Django/psycopg/SDK deps), so I moved it to `unstract.core.polling` and now BOTH PG result pollers import it — the backend `DjangoQueueTransport` AND the workers `PgResultBackend.wait_for_result` (which dropped its own loop). The backoff constants live in exactly one place to tune now. Behaviour-identical: backend 8 + workers 38 + result_backend 7 (real-PG) green.

PR #2104, opened 2026-06-22
[FIX] Bump DRF version to 3.17 to address vulnerability and drop DRF auto UniqueTogetherValidator on upsert/catch serializers

chandrasekharan-zipstack:
## What

DRF 3.15+ auto-generates a `UniqueTogetherValidator` from each model's `Meta.constraints` `UniqueConstraint`. For serializers whose **view already owns uniqueness** — either an upsert (`update_or_create`) or an `IntegrityError → DuplicateData` catch — that validator fires at `is_valid()` and returns a `non_field_errors: must make a unique set` 400 on a legitimate re-save, **before** the view runs. This short-circuits the view's duplicate handling (was a rc.343 staging regression once DRF 3.15.2 shipped).

This sets `Meta.validators = []` on those 8 serializers so the view/DB own uniqueness:

- `adapter_processor_v2.AdapterInstanceSerializer` (catch → `DuplicateAdapterNameError`)
- `api_v2.APIDeploymentSerializer` (`IntegrityErrorMixin`)
- `connector_v2.ConnectorInstanceSerializer` (catch → `DuplicateData`)
- `pipeline_v2.PipelineSerializer` (`IntegrityErrorMixin` + view catch)
- `workflow_v2.WorkflowSerializer` (`IntegrityErrorMixin`)
- `prompt_studio_v2.ToolStudioPromptSerializer` (create catches; also fixes Mode-B on PUT/PATCH update)
- `prompt_profile_manager_v2.ProfileManagerSerializer` (create catches; fixes update)
- `notification_v2.NotificationSerializer` (custom `validate_name` already enforces uniqueness)

## Why now / safety

- **No-op on the currently-pinned DRF 3.14** (it only auto-validates legacy `unique_together`, which these models don't use). So this changes nothing at runtime today.
- **Required prep** for the DRF re-bump to 3.17.x (stacked PR follows). Landing it first keeps these endpoints safe the moment 3.15+ is reintroduced.
- Uniqueness is still enforced — by each view's existing upsert / `IntegrityError` catch and the DB constraint itself.

Part of the staged DRF uniqueness fix. Cloud-side counterparts (`pluggable_apps`) go in a separate `unstract-cloud` PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,30 +41,6 @@ test-rig = [
     "requests>=2.31.0",
 ]
 
-hook-check-django-migrations = [
-    "celery>=5.3.4",
-    "cron-descriptor==1.4.0",
-    "django==4.2.30",
-    "djangorestframework==3.14.0",
-    # Pinning django-celery-beat to avoid build issues
-    "django-celery-beat==2.5.0",
-    "django-cors-headers>=4.3.1",
-    "django-redis==5.4.0",
-    "django-tenants==3.5.0",
-    "drf-yasg==1.21.7",
-    "social-auth-app-django==5.3.0",
-    "psycopg2-binary==2.9.9",
-    "python-dotenv==1.2.2",
-    "python-magic==0.4.27",
-    "unstract-connectors",
-    "unstract-core",
-    "unstract-flags",
-    "unstract-tool-registry",
-    "unstract-tool-sandbox",
-    "unstract-workflow-execution",
-    "unstract-filesystem",
-]
-
 [tool.uv.sources]
 unstract-filesystem = { path = "./unstract/filesystem", editable = true }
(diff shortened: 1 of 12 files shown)


Deepak-Kesavan (approved):

pyproject.toml
```diff
@@ -41,30 +41,6 @@ test-rig = [
     "requests>=2.31.0",
 ]
 
-hook-check-django-migrations = [
```
muhammad-ali-e, line 44: Why do we removing this?

chandrasekharan-zipstack: @muhammad-ali-e this is unused and instead it uses the .venv from `backend` itself. One of the libs in it caused an issue with the DRF version bump also

PR #2106, opened 2026-06-23
UN-3616 [FEAT] PG Queue — route API-triggered pipeline trigger through the transport flag

muhammad-ali-e:
## What

The API-triggered pipeline view (`PipelineApiExecution.post`) dispatched `scheduler.tasks.execute_pipeline_task` on **Celery unconditionally** — so when the flag was on, the trigger stayed on Celery while the execution it spawns rode PG, breaking uniform flag control (one of the migration's correctness bars: flag ON → all PG, flag OFF → all Celery, zero regression).

The **scheduled** pipeline path already routes this task to PG (`pg_scheduler.dispatch_due_schedules` → `scheduler` queue, run by `worker-pg-scheduler`). Only the API-trigger view bypassed it.

## How

A new `dispatch_pipeline_trigger` helper routes the trigger through `resolve_transport` (the same flag as everything else):
- **PG** → enqueue `execute_pipeline_task` to the `scheduler` queue via `pg_queue.producer.enqueue_task`, with the **exact arg shape** `pg_scheduler` already sends.
- **Celery** → `send_task` (unchanged).
- **Fail-closed** — master gate off (production default) → Celery → byte-identical to before, **zero regression**.

Routing logic lives in the helper, not the view (tiny-jobs).

## Tests

- 4 unit tests: PG/Celery routing, pipeline-id entity bucketing, identical args on both paths.
- Dev-tested against the live flag: gate ON → PG `scheduler` queue (correct args), gate OFF → Celery `send_task`.

## Scope notes

- This closes 1 of the dispatch-coverage gaps found in the completeness audit. The audit's other OSS candidate (`submit_file_batch_for_processing`) is **dead code** — defined + URL-registered, never called — handled separately.
- The 2 remaining gaps are cloud-side (`manual_review` execution-time tasks), tracked separately.

Lands on the `feat/UN-3445-pg-queue-integration` branch.

### DIFF
--- a/backend/pipeline_v2/piepline_api_execution_views.py
+++ b/backend/pipeline_v2/piepline_api_execution_views.py
@@ -8,6 +8,7 @@ from rest_framework.response import Response
 from backend.celery_service import app as celery_app
 from pipeline_v2.deployment_helper import DeploymentHelper
 from pipeline_v2.models import Pipeline
+from pipeline_v2.pipeline_dispatch import dispatch_pipeline_trigger
 
 logger = logging.getLogger(__name__)
 
@@ -29,17 +30,14 @@ class PipelineApiExecution(views.APIView):
     def post(
         self, request: Request, org_name: str, pipeline_id: str, pipeline: Pipeline
     ) -> Response:
-        celery_app.send_task(
-            "scheduler.tasks.execute_pipeline_task",
-            args=[
-                "",  # workflow_id
-                org_name,  # org_schema
-                "",  # execution_action
-                "",  # execution_id
-                pipeline_id,  # pipepline_id
-                True,  # with_logs
-                pipeline.pipeline_name,  # name
-            ],
+        # Route through the transport flag (PG when enabled for this pipeline/org,
+        # else Celery — fail-closed). Matches the scheduled-pipeline path, which
+        # already enqueues this task onto the PG queue.
+        dispatch_pipeline_trigger(
+            celery_app=celery_app,
+            org_id=org_name,
+            pipeline_id=pipeline_id,
+            pipeline_name=pipeline.pipeline_name,
         )
         logger.info(f"Triggered {pipeline} by API")
         return Response({"message": f"Triggered {pipeline}"}, status=status.HTTP_200_OK)
--- a/backend/pipeline_v2/piepline_api_execution_views.py
+++ b/backend/pipeline_v2/piepline_api_execution_views.py
@@ -30,9 +30,8 @@ class PipelineApiExecution(views.APIView):
     def post(
         self, request: Request, org_name: str, pipeline_id: str, pipeline: Pipeline
     ) -> Response:
-        # Route through the transport flag (PG when enabled for this pipeline/org,
-        # else Celery — fail-closed). Matches the scheduled-pipeline path, which
-        # already enqueues this task onto the PG queue.
+        # Route through the transport flag instead of hardcoding Celery
+        # (see dispatch_pipeline_trigger).
         dispatch_pipeline_trigger(
             celery_app=celery_app,
             org_id=org_name,
--- a/backend/pipeline_v2/pipeline_dispatch.py
+++ b/backend/pipeline_v2/pipeline_dispatch.py
@@ -23,7 +23,7 @@ from uuid import UUID
 from pg_queue.producer import enqueue_task
 from workflow_manager.workflow_v2.transport import resolve_transport
 
-from unstract.core.data_models import WorkflowTransport
+from unstract.core.data_models import is_pg_transport
 
 logger = logging.getLogger(__name__)
 
@@ -56,9 +56,9 @@ def dispatch_pipeline_trigger(
         organization_id=org_id,
         pipeline_id=pipeline_id,
     )
-    # Compare the enum member (not a bare string) so a literal typo can't silently
-    # fall through to the Celery branch; WorkflowTransport is a str-enum.
-    if transport == WorkflowTransport.PG_QUEUE:
+    # Use the shared is_pg_transport() — the single source for "what counts as PG
+    # transport" — rather than opening a second comparison site.
+    if is_pg_transport(transport):
         msg_id = enqueue_task(
             task_name=PIPELINE_TRIGGER_TASK,
             queue=SCHEDULER_QUEUE,
--- a/backend/pipeline_v2/pipeline_dispatch.py
+++ b/backend/pipeline_v2/pipeline_dispatch.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import logging
 from typing import Any
+from uuid import UUID
 
 from pg_queue.producer import enqueue_task
 from workflow_manager.workflow_v2.transport import resolve_transport
@@ -36,16 +37,16 @@ SCHEDULER_QUEUE = "scheduler"
 def dispatch_pipeline_trigger(
     *,
     celery_app: Any,
-    org_id: str,
-    pipeline_id: str,
+    org_id: str | UUID,
+    pipeline_id: str | UUID,
     pipeline_name: str,
-) -> str:
+) -> None:
     """Dispatch the pipeline-trigger task on the resolved transport.
 
-    Returns the transport actually used (``"pg_queue"`` / ``"celery"``). The
-    positional args match ``execute_pipeline_task``'s signature on **both** paths:
-    ``(workflow_id, org_schema, execution_action, execution_id, pipeline_id,
-    with_logs, name)``.
+    The positional args match ``execute_pipeline_task``'s signature on **both**
+    paths: ``(workflow_id, org_schema, execution_action, execution_id,
+    pipeline_id, with_logs, name)``. ``org_id`` / ``pipeline_id`` accept ``UUID``
+    (what ``resolve_transport`` takes) and are str-coerced into the task args.
     """
     args = ["", str(org_id), "", "", str(pipeline_id), True, pipeline_name]
     # No execution exists yet (it's created inside the task), so the trigger
@@ -69,7 +70,6 @@ def dispatch_pipeline_trigger(
             pipeline_id,
             msg_id,
         )
-        return transport
-    celery_app.send_task(PIPELINE_TRIGGER_TASK, args=args)
-    logger.info("Pipeline %s trigger dispatched on Celery", pipeline_id)
-    return transport
+    else:
+        celery_app.send_task(PIPELINE_TRIGGER_TASK, args=args)
+        logger.info("Pipeline %s trigger dispatched on Celery", pipeline_id)
--- a/backend/pipeline_v2/pipeline_dispatch.py
+++ b/backend/pipeline_v2/pipeline_dispatch.py
@@ -1,17 +1,17 @@
 """Transport-routed dispatch for the API-triggered pipeline execution trigger.
 
-The **scheduled** pipeline path already enqueues
-``scheduler.tasks.execute_pipeline_task`` onto the PG queue
-(``pg_scheduler.dispatch_due_schedules`` → the ``scheduler`` queue, run by
-``worker-pg-scheduler``). The **API-trigger** view historically sent it only to
-Celery — so when the flag was on, the trigger stayed on Celery while the
-execution it spawns rode PG, breaking uniform flag control. This routes the
-trigger through the same :func:`resolve_transport` flag as everything else.
+Routes ``scheduler.tasks.execute_pipeline_task`` through the same
+:func:`resolve_transport` flag as the rest of the execution path: the PG queue
+when enabled for this pipeline/org, else Celery. **Fail-closed** — with the
+master gate off (the production default) it resolves to Celery, behaving exactly
+like the prior unconditional ``send_task`` (zero regression). The
+scheduled-pipeline path already enqueues this task onto the PG ``scheduler``
+queue (``pg_scheduler.dispatch_due_schedules``, run by ``worker-pg-scheduler``);
+this brings the API-trigger view to parity.
 
-**Fail-closed / zero regression:** with the master gate off (production default)
-``resolve_transport`` returns Celery → identical to the prior ``send_task``. The
-task args are byte-identical on both paths, so the consumer behaves the same
-regardless of transport.
+The same ``args`` are sent on both paths; on PG they're JSON-normalized by
+``enqueue_task`` (UUIDs/datetimes → str), a no-op for these string/bool values,
+so the consumer sees the same payload.
 """
 
 from __future__ import annotations
@@ -47,7 +47,7 @@ def dispatch_pipeline_trigger(
     ``(workflow_id, org_schema, execution_action, execution_id, pipeline_id,
     with_logs, name)``.
     """
-    args = ["", org_id, "", "", str(pipeline_id), True, pipeline_name]
+    args = ["", str(org_id), "", "", str(pipeline_id), True, pipeline_name]
     # No execution exists yet (it's created inside the task), so the trigger
     # buckets the flag by pipeline_id as the sticky entity.
     transport = resolve_transport(
@@ -55,7 +55,9 @@ def dispatch_pipeline_trigger(
         organization_id=org_id,
         pipeline_id=pipeline_id,
     )
-    if transport == WorkflowTransport.PG_QUEUE.value:
+    # Compare the enum member (not a bare string) so a literal typo can't silently
+    # fall through to the Celery branch; WorkflowTransport is a str-enum.
+    if transport == WorkflowTransport.PG_QUEUE:
         msg_id = enqueue_task(
             task_name=PIPELINE_TRIGGER_TASK,
             queue=SCHEDULER_QUEUE,
--- a/backend/pipeline_v2/pipeline_dispatch.py
+++ b/backend/pipeline_v2/pipeline_dispatch.py
@@ -1,0 +1,73 @@
+"""Transport-routed dispatch for the API-triggered pipeline execution trigger.
+
+The **scheduled** pipeline path already enqueues
+``scheduler.tasks.execute_pipeline_task`` onto the PG queue
+(``pg_scheduler.dispatch_due_schedules`` → the ``scheduler`` queue, run by
+``worker-pg-scheduler``). The **API-trigger** view historically sent it only to
+Celery — so when the flag was on, the trigger stayed on Celery while the
+execution it spawns rode PG, breaking uniform flag control. This routes the
+trigger through the same :func:`resolve_transport` flag as everything else.
+
+**Fail-closed / zero regression:** with the master gate off (production default)
+``resolve_transport`` returns Celery → identical to the prior ``send_task``. The
+task args are byte-identical on both paths, so the consumer behaves the same
+regardless of transport.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Any
+
+from pg_queue.producer import enqueue_task
+from workflow_manager.workflow_v2.transport import resolve_transport
+
+from unstract.core.data_models import WorkflowTransport
+
+logger = logging.getLogger(__name__)
+
+# The fired task + the PG queue a ``scheduler`` consumer polls. Mirrors
+# ``pg_scheduler.PIPELINE_TRIGGER_TASK`` / ``SCHEDULER_QUEUE_NAME`` — kept as local
+# constants so the backend doesn't import the workers package.
+PIPELINE_TRIGGER_TASK = "scheduler.tasks.execute_pipeline_task"
+SCHEDULER_QUEUE = "scheduler"
+
+
+def dispatch_pipeline_trigger(
+    *,
+    celery_app: Any,
+    org_id: str,
+    pipeline_id: str,
+    pipeline_name: str,
+) -> str:
+    """Dispatch the pipeline-trigger task on the resolved transport.
+
+    Returns the transport actually used (``"pg_queue"`` / ``"celery"``). The
+    positional args match ``execute_pipeline_task``'s signature on **both** paths:
+    ``(workflow_id, org_schema, execution_action, execution_id, pipeline_id,
+    with_logs, name)``.
+    """
+    args = ["", org_id, "", "", str(pipeline_id), True, pipeline_name]
+    # No execution exists yet (it's created inside the task), so the trigger
+    # buckets the flag by pipeline_id as the sticky entity.
+    transport = resolve_transport(
+        execution_id=pipeline_id,
+        organization_id=org_id,
+        pipeline_id=pipeline_id,
+    )
+    if transport == WorkflowTransport.PG_QUEUE.value:
+        msg_id = enqueue_task(
+            task_name=PIPELINE_TRIGGER_TASK,
+            queue=SCHEDULER_QUEUE,
+            args=args,
+            org_id=str(org_id or ""),
+        )
+        logger.info(
+            "Pipeline %s trigger enqueued on PG scheduler queue (msg_id=%s)",
+            pipeline_id,
+            msg_id,
+        )
+        return transport
+    celery_app.send_task(PIPELINE_TRIGGER_TASK, args=args)
+    logger.info("Pipeline %s trigger dispatched on Celery", pipeline_id)
+    return transport
--- a/backend/pipeline_v2/tests/test_pipeline_dispatch.py
+++ b/backend/pipeline_v2/tests/test_pipeline_dispatch.py
@@ -33,8 +33,7 @@ class TestDispatchPipelineTrigger:
             patch.object(pd, "resolve_transport", return_value="pg_queue"),
             patch.object(pd, "enqueue_task", return_value=42) as enqueue,
         ):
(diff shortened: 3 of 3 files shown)


muhammad-ali-e (commented):
Automated PR review via the PR Review Toolkit (code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier). 20 inline findings grouped below, prioritized High -> Low.

pg_benchmark/pg_benchmark/probe.py
```diff
@@ -0,0 +1,121 @@
...
+            ok=False,
+            http_latency=trig.http_latency,
+            wall_clock_e2e=None,
+            server_execution_time=None,
+            parallelism=None,
+            overhead=None,
+            error=trig.error or "trigger returned no execution_id",
+        )
+
+    conn = connect(db_cfg)
```
muhammad-ali-e, line 76: **[High] DB connect/poll not isolated — one bad probe kills the whole batch.**

`conn = connect(db_cfg)` is outside the `try`, and `fetch_status`/`fetch_one` only have a `finally: conn.close()` — they never convert a transient DB error into a failed `RunResult`. `runner.py` documents "failures are isolated: one bad run becomes one `RunResult` with `ok=False`, never a dead batch," but a single dropped connection here raises out of `future.result()` and loses every already-completed measurement.

**Fix:** move `connect` inside the guard and return `RunResult(ok=False, error=f"db connect failed for execution {trig.execution_id}: {exc}")` on failure (preserving `execution_id` so the orphaned server-side run stays traceable).

pg_benchmark/pg_benchmark/runner.py
```diff
@@ -0,0 +1,76 @@
...
+                probe,
+                trigger_cfg,
+                db_cfg,
+                poll_interval=poll_interval,
+                timeout=timeout,
+            )
+            for _ in range(n)
+        ]
+        for future in as_completed(futures):
+            result = future.result()
```
muhammad-ali-e, line 72: **[High] `future.result()` re-raises, aborting `run_load` and discarding collected results.**

Any exception thrown inside `run_probe` propagates here, the `ThreadPoolExecutor` context unwinds, and every result gathered so far is lost — contradicting the module's "never a dead batch" contract (lines 4-7).

**Fix:** wrap in try/except and synthesize a failed `RunResult`:
```python
try:
    result = future.result()
except Exception as exc:  # isolate per the module contract
    result = RunResult(execution_id=None, transport=None, status=None, ok=False,
        http_latency=0.0, wall_clock_e2e=None, server_execution_time=None,
        parallelism=None, overhead=None, error=f"probe crashed: {exc!r}")
```
Add a regression test pinning the documented isolation guarantee.

backend/pipeline_v2/pipeline_dispatch.py
```diff
@@ -0,0 +1,73 @@
...
+    args = ["", org_id, "", "", str(pipeline_id), True, pipeline_name]
+    # No execution exists yet (it's created inside the task), so the trigger
+    # buckets the flag by pipeline_id as the sticky entity.
+    transport = resolve_transport(
+        execution_id=pipeline_id,
+        organization_id=org_id,
+        pipeline_id=pipeline_id,
+    )
+    if transport == WorkflowTransport.PG_QUEUE.value:
+        msg_id = enqueue_task(
```
muhammad-ali-e, line 62: **[High — test gap] The "PG enqueue failure propagates, no Celery fallback" contract is untested.**

`producer.py` promises "for the orchestrator there is no silent Celery fallback (that would hide the failure or risk a double-dispatch)." This dispatcher has no try/except, so an `enqueue_task` raise must surface as a 500 — but nothing verifies it. A future refactor adding `try/except -> send_task` would cause double-dispatch and no test would fail.

**Fix:** add a test patching `resolve_transport -> "pg_queue"` and `enqueue_task(side_effect=RuntimeError)`, asserting the call raises and `celery_app.send_task` was NOT called.

pg_benchmark/pg_benchmark/trigger.py
```diff
@@ -0,0 +1,150 @@
...
+                error=f"request failed: {exc}",
+            )
+        latency = time.monotonic() - start
+        try:
+            payload = resp.json()
+        except ValueError:
+            payload = None
+        execution_id = _extract_execution_id(payload)
+        error = None
+        if resp.status_code >= 400:
```
muhammad-ali-e, line 138: **[High — test gap] `trigger_execution`'s error branches have zero coverage.**

`test_trigger.py` only tests the parsing helpers. The shipped function — the `requests.RequestException` branch (`http_status=0`), the `status_code >= 400` path, the "no execution_id" path, and the file open/close lifecycle — is the code that becomes `RunResult(ok=False)` under real load, and it is entirely untested.

**Fix:** with a stubbed `Session`, assert (a) a 401 -> error contains "HTTP 401" and `execution_id is None`; (b) a `RequestException` -> `http_status=0`; (c) a 200 with no id -> "no execution_id"; (d) opened file handles are closed even when `post` raises.

backend/pipeline_v2/pipeline_dispatch.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    args = ["", org_id, "", "", str(pipeline_id), True, pipeline_name]
+    # No execution exists yet (it's created inside the task), so the trigger
+    # buckets the flag by pipeline_id as the sticky entity.
+    transport = resolve_transport(
+        execution_id=pipeline_id,
+        organization_id=org_id,
+        pipeline_id=pipeline_id,
+    )
+    if transport == WorkflowTransport.PG_QUEUE.value:
```
muhammad-ali-e: **[Medium] Transport flag is compared stringly, defeating the `WorkflowTransport` enum.**

`resolve_transport()` returns a raw `str` and this branches on `transport == WorkflowTransport.PG_QUEUE.value`. The codebase already has a fail-closed coercer (`normalize_transport`) next to the enum. A typo on the RHS (`"pg-queue"`) silently routes everything to the Celery `else` with no error — exactly the trap the enum exists to prevent.

**Fix:** normalize once into the enum and branch by identity:
```python
transport = WorkflowTransport(normalize_transport(resolve_transport(...), logger=logger,
    context=f"pipeline:{pipeline_id}"))
if transport is WorkflowTransport.PG_QUEUE:
    ...
```
Consider returning the enum (not the bare string) too.

backend/pipeline_v2/pipeline_dispatch.py
```diff
@@ -0,0 +1,73 @@
...
+    pipeline_name: str,
+) -> str:
+    """Dispatch the pipeline-trigger task on the resolved transport.
+
+    Returns the transport actually used (``"pg_queue"`` / ``"celery"``). The
+    positional args match ``execute_pipeline_task``'s signature on **both** paths:
+    ``(workflow_id, org_schema, execution_action, execution_id, pipeline_id,
+    with_logs, name)``.
+    """
+    args = ["", org_id, "", "", str(pipeline_id), True, pipeline_name]
```
muhammad-ali-e: **[Medium] 7-field task-arg contract is positional and untyped; invariant lives only in a comment.**

`args = ["", org_id, "", "", str(pipeline_id), True, pipeline_name]` encodes `(workflow_id, org_schema, execution_action, execution_id, pipeline_id, with_logs, name)` purely by position. Both transports and the downstream consumer must agree byte-for-byte; a reorder or dropped element fails silently at the consumer. Note `org_id` is raw at index 1 but `str(org_id or "")`-coerced for the PG kwarg — the inconsistency hints the contract is fragile.

Also a **test gap:** every test passes a `str` `pipeline_id`, so the `str(pipeline_id)` coercion is never exercised with a `UUID` (which `resolve_transport` accepts) — the "args identical on both paths" test wouldn't catch a divergence.

**Fix:** build the list from named locals (or a small frozen `NamedTuple`/dataclass with a `.to_task_args()` method) so order/arity live in one place, and add a UUID-id test.

pg_benchmark/pg_benchmark/db.py
```diff
@@ -0,0 +1,210 @@
...
+    execution_id: str
+    transport: Transport
+    status: str
+    total_files: int
+    server_execution_time: float | None
+    file_times: list[float] = field(default_factory=list)
+
+    @property
+    def is_terminal(self) -> bool:
+        return self.status in ("COMPLETED", "ERROR", "STOPPED")
```
muhammad-ali-e, line 60: **[Medium] Terminal-status set is hard-coded in 4 places, duplicating the shared `ExecutionStatus` enum.**

`is_terminal` here (line 60) and `_TERMINAL_STATUSES` (line 158) both list `("COMPLETED","ERROR","STOPPED")`, and `probe.py`/`report.py` compare against `"COMPLETED"`/`"ERROR"` inline. `unstract.core.data_models.ExecutionStatus` is the source of truth. If a status value changes or a terminal state is added, these drift independently and the benchmark silently mis-buckets executions.

**Fix:** import `ExecutionStatus` and define the terminal set once; have `is_terminal` delegate to `is_terminal_status(self.status)` instead of re-listing the tuple.

pg_benchmark/pg_benchmark/db.py
```diff
@@ -0,0 +1,210 @@
...
+    conn: psycopg2.extensions.connection, execution_id: str
+) -> ExecutionLatency | None:
+    """Return the full ``ExecutionLatency`` for one execution, or ``None``."""
+    with conn.cursor() as cur:
+        cur.execute(_ONE_SQL, {"execution_id": execution_id, "limit": 1})
+        row = cur.fetchone()
+    if row is None:
+        return None
+    execution_id_, has_qmid, has_taskid, status, total_files, exec_time, file_times = row
+    return ExecutionLatency(
```
muhammad-ali-e, line 188: **[Medium] `fetch_one` and `fetch_recent` duplicate the row -> `ExecutionLatency` mapping byte-for-byte.**

Both construct `ExecutionLatency` with identical `Transport.classify(...)`, `total_files or 0`, and `list(file_times or [])` logic over the same `_RECENT_SQL` column order (~12 duplicated lines).

**Fix:** extract `_row_to_latency(row) -> ExecutionLatency` and call it from both.

pg_benchmark/pg_benchmark/probe.py
```diff
@@ -0,0 +1,121 @@
...
+            execution_id=trig.execution_id,
+            transport=latency.transport if latency else None,
+            status=status,
+            ok=False,
+            http_latency=trig.http_latency,
+            wall_clock_e2e=wall_clock_e2e,
+            server_execution_time=latency.server_execution_time if latency else None,
+            parallelism=latency.parallelism if latency else None,
+            overhead=None,
+            error=f"timed out after {timeout:.0f}s (last status={status})",
```
muhammad-ali-e, line 101: **[Medium] Timed-out probe masks an unreadable DB row as a generic timeout.**

When `fetch_one` returns `None` (row absent/unreadable), every server-side field is `None` and the error is just `"timed out after Ns"`. A user can't distinguish "execution genuinely slow" from "harness can't see the execution row at all" — a data-visibility problem disguised as latency.

**Fix:** when `latency is None`, say so: `error=f"timed out after {timeout:.0f}s (last status={status}; execution row not readable)"`.

_(Minor: the `latency.X if latency else None` guard is repeated 6x across both return sites — hoist the three values into locals after `fetch_one`.)_

pg_benchmark/pg_benchmark/report.py
```diff
@@ -0,0 +1,156 @@
...
+    # ``None`` transport bucket = runs that never produced a classifiable row.
+    transports: list[Transport | None] = [*Transport, None]
+    for transport in transports:
+        bucket = [r for r in results if r.transport is transport]
+        if not bucket:
+            continue
+        completed = [r for r in bucket if r.ok]
+        reports.append(
+            LoadReport(
+                transport=transport or Transport.INLINE,
```
muhammad-ali-e, line 78: **[Medium] `None`-transport bucket is relabeled `INLINE`, mislabeling trigger-failures as inline runs.**

The comment at line 69 correctly says the `None` bucket = "runs that never produced a classifiable row," but `transport or Transport.INLINE` then renders it under the same "INLINE" header as genuine inline executions — and a real run with both buckets emits two reports both titled INLINE.

**Fix:** render the `None` bucket under a distinct label (e.g. "UNKNOWN"/"UNCLASSIFIED").

pg_benchmark/pg_benchmark/trigger.py
```diff
@@ -0,0 +1,150 @@
...
+            return TriggerResult(
+                execution_id=None,
+                http_status=0,
+                http_latency=time.monotonic() - start,
+                error=f"request failed: {exc}",
+            )
+        latency = time.monotonic() - start
+        try:
+            payload = resp.json()
+        except ValueError:
```
muhammad-ali-e, line 134: **[Medium] Non-JSON 200 body is swallowed, hiding the real cause.**

`except ValueError: payload = None` discards the parse failure. A 200 with a non-JSON body (e.g. an HTML error page from a proxy) becomes `payload=None` -> `"no execution_id in response: None"`, losing the body that would explain the misconfig. The raw `resp.text` is only surfaced for `status_code >= 400`.

**Fix:** in the `elif execution_id is None` branch include a body snippet regardless of JSON-parseability: `error = f"no execution_id; body[:200]={resp.text[:200]!r}"`.

pg_benchmark/pg_benchmark/mock_server.py
```diff
@@ -0,0 +1,223 @@
...
+            self.send_header("Content-Length", str(len(body)))
+            self.end_headers()
+            self.wfile.write(body)
+
+        def _read_json(self) -> dict:
+            length = int(self.headers.get("Content-Length", 0) or 0)
+            raw = self.rfile.read(length) if length else b""
+            try:
+                parsed = json.loads(raw or b"{}")
+            except ValueError:
```
muhammad-ali-e, line 126: **[Medium] `_read_json` swallows a malformed request body into `{}`, producing false-positive 200s.**

`except ValueError: return {}` (and the non-dict fallback) means an unparseable/unexpected body is treated as empty and the handler still returns a canned 200. This server exists to validate the transport/adapter path — if the adapter sends a body the mock can't parse, the user gets a green run and never learns the request was malformed.

**Fix:** count malformed bodies (expose on `/health`) or return `400`, so a contract mismatch is visible.

pg_benchmark/pg_benchmark/mock_server.py
```diff
@@ -0,0 +1,223 @@
...
+    """Build (but don't start) a threaded mock server. ``port=0`` picks a free port."""
+    counter = _Counter()
+    server = ThreadingHTTPServer((host, port), _make_handler(config, counter))
+    return server, counter
+
+
+def build_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(
+        prog="pg_benchmark.mock_server",
+        description="Instant OpenAI-compatible mock LLM+embedding server for load testing.",
```
muhammad-ali-e, line 185: **[Low] Line exceeds the configured ruff `line-length = 90` (92 chars).**

Both the repo root and `pg_benchmark/pyproject.toml` cap at 90; `ruff format` would flag this `description=` string. Wrap it:
```python
description=(
    "Instant OpenAI-compatible mock LLM+embedding server for load testing."
),
```

pg_benchmark/pg_benchmark/mock_server.py
```diff
@@ -0,0 +1,223 @@
...
+    """Thread-safe request tallies (so a load run can confirm the server was hit)."""
+
+    def __init__(self) -> None:
+        self._lock = Lock()
+        self.chat = 0
+        self.embeddings = 0
+
+    def bump(self, kind: str) -> int:
+        with self._lock:
+            value = getattr(self, kind) + 1
```
muhammad-ali-e, line 58: **[Low] `_Counter.bump` uses `getattr`/`setattr` string dispatch where explicit fields would be clearer.**

Callers only ever pass `"chat"` or `"embeddings"`. Attribute-name-as-string indirection bypasses the explicitness the project favors.

**Fix:** back it with `self._counts = {"chat": 0, "embeddings": 0}` (under the existing lock) and key into the dict, keeping the `/health` readout straightforward.

backend/pipeline_v2/pipeline_dispatch.py
```diff
@@ -0,0 +1,73 @@
...
+``scheduler.tasks.execute_pipeline_task`` onto the PG queue
+(``pg_scheduler.dispatch_due_schedules`` → the ``scheduler`` queue, run by
+``worker-pg-scheduler``). The **API-trigger** view historically sent it only to
+Celery — so when the flag was on, the trigger stayed on Celery while the
+execution it spawns rode PG, breaking uniform flag control. This routes the
+trigger through the same :func:`resolve_transport` flag as everything else.
+
+**Fail-closed / zero regression:** with the master gate off (production default)
+``resolve_transport`` returns Celery → identical to the prior ``send_task``. The
+task args are byte-identical on both paths, so the consumer behaves the same
```
muhammad-ali-e: **[Low — comment accuracy] "byte-identical args on both paths" is slightly overstated.**

The Python `args` list is identical, but the PG path runs it through `enqueue_task -> _json_safe` (`json.loads(json.dumps(..., default=str))`) while Celery passes it raw. It's a no-op for these string/bool values, but the claim will mislead a maintainer who later adds a UUID/datetime arg.

**Fix:** soften to: "the same `args` are sent on both paths; on PG they're JSON-normalized by `enqueue_task` (UUIDs/datetimes -> str), a no-op for these values, so the consumer sees the same payload."

backend/pipeline_v2/pipeline_dispatch.py
```diff
@@ -0,0 +1,73 @@
+"""Transport-routed dispatch for the API-triggered pipeline execution trigger.
+
+The **scheduled** pipeline path already enqueues
+``scheduler.tasks.execute_pipeline_task`` onto the PG queue
+(``pg_scheduler.dispatch_due_schedules`` → the ``scheduler`` queue, run by
+``worker-pg-scheduler``). The **API-trigger** view historically sent it only to
```
muhammad-ali-e: **[Low — comment rot] Module docstring narrates PR history ("historically sent it only to Celery").**

Great PR-description prose, but as a permanent docstring it freezes a "before" state that no longer exists; once Celery is decommissioned (the UN-3445 goal) this becomes archaeology.

**Fix:** lead with current behavior; move the "historically / breaking uniform flag control" rationale to the PR description or a one-line `# Pre-UN-3616: this view bypassed the transport flag.` note.

backend/pipeline_v2/piepline_api_execution_views.py
```diff
@@ -29,17 +30,14 @@ def initialize_request(self, request: Request, *args: Any, **kwargs: Any) -> Req
...
-            args=[
-                "",  # workflow_id
-                org_name,  # org_schema
-                "",  # execution_action
-                "",  # execution_id
-                pipeline_id,  # pipepline_id
-                True,  # with_logs
-                pipeline.pipeline_name,  # name
-            ],
+        # Route through the transport flag (PG when enabled for this pipeline/org,
```
muhammad-ali-e: **[Low] Inline comment duplicates the `dispatch_pipeline_trigger` docstring almost verbatim.**

Two copies of the same rationale will drift out of sync.

**Fix:** trim to one line, e.g. `# Route through the transport flag instead of hardcoding Celery (see dispatch_pipeline_trigger).`

pg_benchmark/pg_benchmark/stats.py
```diff
@@ -0,0 +1,74 @@
...
+    @property
+    def empty(self) -> bool:
+        return self.n == 0
+
+
+def percentile(values: list[float], pct: float) -> float:
+    """Linear-interpolated percentile (``pct`` in 0..100).
+
+    Uses the same "linear interpolation between closest ranks" method as
+    ``numpy.percentile`` / ``statistics.quantiles(method="inclusive")`` so the
```
muhammad-ali-e, line 37: **[Low — comment accuracy] The `statistics.quantiles(method="inclusive")` equivalence is misleading.**

`quantiles()` returns n-1 evenly spaced cut points and takes no percentile argument — a different API/abstraction, even if the underlying interpolation rule matches. A maintainer who tries to swap it in to "simplify" will hit a different shape. The `numpy.percentile` comparison is correct.

**Fix:** drop the `statistics.quantiles` reference (or qualify it as "same interpolation rule as, but not API-compatible with").

pg_benchmark/pg_benchmark/trigger.py
```diff
@@ -0,0 +1,150 @@
...
+
+    @classmethod
+    def from_env(cls, **overrides: object) -> TriggerConfig:
+        base = {
+            "base_url": os.environ.get("PGBENCH_BASE_URL", "http://localhost:8000"),
+            "path": os.environ.get("PGBENCH_DEPLOY_PATH", ""),
+            "api_key": os.environ.get("PGBENCH_API_KEY", ""),
+        }
+        base.update({k: v for k, v in overrides.items() if v is not None})
+        return cls(**base)  # type: ignore[arg-type]
```
muhammad-ali-e, line 49: **[Low] `TriggerConfig.from_env` bypasses dataclass typing via `dict.update` + `# type: ignore`.**

An override of the wrong type (e.g. `request_timeout="600"`) flows straight into the frozen instance with no validation, and the `type: ignore` suppresses the one signal that would catch it. Contrast `DbConfig.from_env`, which reads/coerces each field explicitly (the safer pattern, already in this PR).

**Fix:** mirror `DbConfig.from_env` — read+coerce each field explicitly, removing the need for `type: ignore`.

pg_benchmark/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+**Parallelism** `= sum(file_times) / execution_time`: `≈ N` means all N files
+overlapped (ideal fan-out), `≈ 1` means they ran serially. This directly
+measures the serial-fileproc concern on the PG path.
+
+> Transport micro-latency (enqueue→pickup, the SKIP-LOCKED-poll-vs-push KPI) is
+> **not** readable post-hoc — those queue rows are deleted on ack. It needs a
+> live sampler that watches `pg_queue_message` while a run is in flight; that is
+> a later slice.
+
```
muhammad-ali-e, line 31: **[Low — doc rot] The "Status — slices" section is already stale on merge.**

It lists S2 (`run`/load generation) and S3 (live sampler) as "later slices," but `run`, `runner.py`, `probe.py`, and `trigger.py` all ship in *this* PR — a reader sees `run` documented under Usage yet listed as future work.

**Fix:** mark S2 as shipped, or replace the slice-by-slice status with a plain "Implemented / Not yet implemented" feature list that doesn't encode PR ordering.

muhammad-ali-e:
Thanks for the review. Triaged the 20 findings against the **current** PR scope (after syncing with integration, the `pg_benchmark` harness is now in the base, not this PR).

## In-scope (`backend/pipeline_v2/*`) — addressed in `19e689633`

- **[High] enqueue-failure-propagates untested** → added `test_pg_enqueue_failure_propagates_with_no_celery_fallback`: `resolve_transport→pg_queue` + `enqueue_task` raising → asserts the error surfaces and `send_task` is **not** called (no silent fallback / double-dispatch).
- **[Medium] stringly transport compare** → now branches on the enum member (`transport == WorkflowTransport.PG_QUEUE`, dropped `.value`) so a literal typo can't fall through to Celery.
- **[Medium] UUID arg coercion untested** → added `test_uuid_pipeline_id_is_coerced_to_str_in_args` (UUID `pipeline_id` → `str` in args, raw UUID to `resolve_transport`); also coerced `args[1]` to `str(org_id)` for consistency.
- **[Low] "byte-identical args" overstated** → docstring corrected (PG path JSON-normalizes via `enqueue_task`; a no-op for these str/bool values).
- **[Low] docstring history / duplicated view comment** → docstring leads with current behavior; view comment trimmed to one line.

6 unit tests green; ruff clean.

## Out-of-scope (`pg_benchmark/*`, 14 findings)

These are on the benchmark harness, which is **no longer in this PR** — it's committed to the integration branch (dev tooling), and after the sync it sits in the base. A couple are genuinely useful *harness* hardening (probe DB-error isolation, `run_load` not aborting on one bad probe) — I'll address those directly on integration, separate from this PR.

backend/pipeline_v2/pipeline_dispatch.py
```diff
@@ -0,0 +1,75 @@
...
+    (what ``resolve_transport`` takes) and are str-coerced into the task args.
+    """
+    args = ["", str(org_id), "", "", str(pipeline_id), True, pipeline_name]
+    # No execution exists yet (it's created inside the task), so the trigger
+    # buckets the flag by pipeline_id as the sticky entity.
+    transport = resolve_transport(
+        execution_id=pipeline_id,
+        organization_id=org_id,
+        pipeline_id=pipeline_id,
+    )
```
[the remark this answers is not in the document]
muhammad-ali-e: Good catch — switched to `is_pg_transport(transport)` in 29cb1a11f's follow-up so this routes through the shared single-source helper instead of a second comparison site.

PR #2109, opened 2026-06-24
UN-3618 [REFACTOR] PG Queue — gate solely on the Flipt flag (+ UN-3619 transient-connect retry)

muhammad-ali-e:
## What & why

Two small, closely-related PG-queue changes, bundled into one draft PR (sub-tasks UN-3618 + UN-3619 under the Phase-9 story).

### UN-3618 — single-flag gating (drop `PG_QUEUE_TRANSPORT_ENABLED`)

The `pg_queue_enabled` Flipt flag is now the **sole** gate for the whole PG-queue feature. The `PG_QUEUE_TRANSPORT_ENABLED` env master-switch was a per-process AND-condition checked *before* Flipt — but with Flipt per-env and fail-closed it only ever forced **OFF** (it never enabled PG on its own), so it added no safety the flag didn't already provide while risking config drift across the backend/worker processes. The deploy-ordering argument isn't PG-specific either (Celery has the same "no consumer → stuck queue" exposure). Removing it now, before any cloud helm/configmap wiring exists, means we never add it there.

- Drops the gate check in all four resolvers: execution `transport.py`, `scheduler/ownership.py`, and both `pg_queue/executor_rpc.py` halves.
- Removes the `master_gate_enabled` parameter from the shared `unstract.workflow_execution.executor_rpc.resolve_pg_transport`.
- Removes the Django setting, 3 `sample.env` entries, a docker-compose comment, the `flags.py` mention.
- **Fully fail-closed**: flag off / Flipt blind / Flipt error / no org → Celery.

### UN-3619 — bounded retry on transient Postgres connect

`create_pg_connection` (workers) retries a *transient* `psycopg2.OperationalError` (DB restart, PgBouncer pool wait, brief network partition, "too many clients" spike) with bounded exponential backoff (`WORKER_PG_QUEUE_CONNECT_RETRIES` default 3, `WORKER_PG_QUEUE_CONNECT_BACKOFF` default 0.5s). Connecting is side-effect-free so retry is safe; the enqueue INSERT is intentionally **not** retried (an ambiguous commit-time failure could double-enqueue → double-dispatch).

## Testing

- Unit: backend **35 passed**, workers **43 passed** (incl. 6 new connect-retry tests).
- Real-stack dev-test: backend boots with the env var absent; `resolve_transport` routes to PG by the Flipt flag alone (5/5); fail-closed (Flipt blind → Celery) intact.
- `ruff check` + `ruff format` clean; pre-commit clean.

## Risk

Zero-regression: behaviour is identical to before for every gate path, and rollback remains the single Flipt flag. The workers' executor-RPC half is best confirmed in the k8s validation (needs rebuilt worker images); unit tests cover its logic here.

### DIFF
--- a/backend/scheduler/ownership.py
+++ b/backend/scheduler/ownership.py
@@ -12,9 +12,9 @@ Doing both in one transaction is what makes "never double-fires" real (it was
 ``PeriodicTask`` disabled, so the two can't both fire.
 
 Inert by default: ``resolve_schedule_owner`` fails closed to Beat
-(``pg_owned=False``) until ops turns the master gate on AND ramps the single
-``pg_queue_enabled`` Flipt flag — so reconciling on every schedule edit is a
-no-op (everything stays Beat-owned) until the rollout starts.
+(``pg_owned=False``) until ops ramps the single ``pg_queue_enabled`` Flipt flag — so
+reconciling on every schedule edit is a no-op (everything stays Beat-owned) until the
+rollout starts.
 """
 
 from __future__ import annotations
@@ -22,7 +22,6 @@ from __future__ import annotations
 import logging
 import os
 
-from django.conf import settings
 from django.db import transaction
 from django.utils import timezone
 from django_celery_beat.models import PeriodicTask
@@ -41,19 +40,14 @@ logger = logging.getLogger(__name__)
 def resolve_schedule_owner(pipeline_id: str, organization_id: str | None) -> bool:
     """True → the PG scheduler owns this schedule; False → Celery Beat does.
 
-    Mirrors ``resolve_transport``: master-gated by ``PG_QUEUE_TRANSPORT_ENABLED``
-    (shared PG kill-switch), then the single ``pg_queue_enabled`` Flipt flag, keyed
-    on ``pipeline_id`` for a stable percentage bucket. **Fails closed to Beat**
-    on a closed gate, a blind Flipt, or any error — so a schedule never silently
-    loses its firer.
+    Mirrors ``resolve_transport``: gated by the single ``pg_queue_enabled`` Flipt
+    flag, keyed on ``pipeline_id`` for a stable percentage bucket. **Fails closed to
+    Beat** on a blind Flipt or any error — so a schedule never silently loses its
+    firer.
     """
-    # Master gate off → never consult Flipt; every schedule stays on Beat.
-    if not settings.PG_QUEUE_TRANSPORT_ENABLED:
-        return False
-
     if os.environ.get("FLIPT_SERVICE_AVAILABLE", "false").lower() != "true":
         logger.warning(
-            "resolve_schedule_owner: gate ON but FLIPT_SERVICE_AVAILABLE != true "
+            "resolve_schedule_owner: FLIPT_SERVICE_AVAILABLE != true "
             "(Flipt blind) for pipeline %s; leaving on Beat",
             pipeline_id,
         )
--- a/backend/workflow_manager/workflow_v2/transport.py
+++ b/backend/workflow_manager/workflow_v2/transport.py
@@ -8,25 +8,23 @@ in the dispatched task's payload (see ``workers/queue_backend/pg_queue/
 row: the payload is the single carrier, durable for PG via the queue row's
 JSONB, and the giant shared table is never migrated for this work.
 
-PR 3 (this change) replaces PR 1's hardwired Celery with a Flipt evaluation:
+The transport is resolved from a single Flipt evaluation:
 
-  master-gate (env) → Flipt boolean (``pg_queue_enabled``) → transport
+  Flipt boolean (``pg_queue_enabled``) → transport
 
-Routing onto PG needs **all three** of: the env master-gate on, Flipt reachable
-(``FLIPT_SERVICE_AVAILABLE=true``), and the flag enabled for this execution.
+Routing onto PG needs **both**: Flipt reachable (``FLIPT_SERVICE_AVAILABLE=true``)
+and the single ``pg_queue_enabled`` flag enabled for this execution. The flag is the
+sole rollout control — flip it to ramp PG, flip it off (or a Flipt outage) to fall
+back to Celery.
 
-- **Master-gate** (``settings.PG_QUEUE_TRANSPORT_ENABLED``, default off): until
-  ops flips it on, Flipt is never consulted and every execution rides Celery.
-  This is both the instant global kill-switch *and* the deploy-ordering safety —
-  the flag stays inert until PG consumers are actually running in the fleet.
 - **Flipt** decides per-execution: ``entity_id = execution_id`` drives the
   percentage-rollout hashing (an execution resolves exactly once, so it can
   never re-bucket mid-flight); ``context`` carries org/workflow/pipeline for
   segment rules. The flag contract is fixed in 9e-design §2.
-- **Fail-closed to Celery**: a Flipt outage must never break execution creation,
-  mirroring ``normalize_transport`` on the read side. The gate-ON path logs its
-  decision so a "gate on but still all Celery" situation (e.g. a blind Flipt)
-  is visible rather than silent.
+- **Fail-closed to Celery**: a blind/unreachable Flipt, a missing org, or any error
+  must never break execution creation — it resolves to Celery, mirroring
+  ``normalize_transport`` on the read side. The decision is logged so a "flag on but
+  still all Celery" situation (e.g. a blind Flipt) is visible rather than silent.
 """
 
 from __future__ import annotations
@@ -35,7 +33,6 @@ import logging
 import os
 from typing import TYPE_CHECKING
 
-from django.conf import settings
 from pg_queue.flags import PG_QUEUE_FLAG_KEY
 
 from unstract.core.data_models import WorkflowTransport
@@ -78,21 +75,11 @@ def resolve_transport(
         pipeline_id: Optional, carried in ``context`` for future segment rules.
 
     Returns:
-        A :class:`WorkflowTransport` value string — ``"pg_queue"`` only when the
-        master-gate is on, Flipt is reachable, and Flipt says yes for this
-        execution; ``"celery"`` otherwise (including any error — fail-closed).
+        A :class:`WorkflowTransport` value string — ``"pg_queue"`` only when Flipt
+        is reachable and says yes for this execution; ``"celery"`` otherwise
+        (including any error — fail-closed).
     """
     celery = WorkflowTransport.CELERY.value
-
-    # Master-gate: until ops sets PG_QUEUE_TRANSPORT_ENABLED=true, never consult
-    # Flipt — every execution rides Celery (kill-switch + deploy-ordering safety).
-    # Intentionally unlogged: this is the steady state for every execution while
-    # the gate is off, so a log here would be pure noise.
-    if not settings.PG_QUEUE_TRANSPORT_ENABLED:
-        return celery
-
-    # Gate is ON (canary/rollout). From here the decision is logged so a
-    # "gate on but everything still Celery" situation cannot hide.
 
     # No org context → per-org segment matching can't be trusted (str(None) would
     # ship a bogus "None" org into the Flipt context and mis-segment). The view
@@ -107,12 +94,12 @@ def resolve_transport(
 
     # FliptClient returns False for ALL flags when the service is marked
     # unavailable — indistinguishable from "rollout says no". Surface it loudly so
-    # a blind Flipt under an ON gate doesn't masquerade as a healthy 100%-Celery
-    # canary. Parse exactly as FliptClient does (``.lower()``, no ``.strip()``)
-    # so the two can never disagree on a value like ``" true"``.
+    # a blind Flipt doesn't masquerade as a healthy 100%-Celery rollout. Parse
+    # exactly as FliptClient does (``.lower()``, no ``.strip()``) so the two can
+    # never disagree on a value like ``" true"``.
     if os.environ.get("FLIPT_SERVICE_AVAILABLE", "false").lower() != "true":
         logger.warning(
-            "resolve_transport: gate ON but FLIPT_SERVICE_AVAILABLE != true "
+            "resolve_transport: FLIPT_SERVICE_AVAILABLE != true "
             "(Flipt is blind) for execution %s; forcing celery",
             execution_id,
         )
--- a/unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
+++ b/unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
@@ -16,10 +16,9 @@ primitive is **injected** (composition, not inheritance) via :class:`QueueTransp
   ``dispatch_with_callback`` + the reply_key/timeout orchestration and the
   never-raises contract (timeout/failure → ``ExecutionResult.failure``). It calls
   ``transport.enqueue(...)`` and ``transport.wait_for_result(...)``.
-- :func:`resolve_pg_transport` — the gate: a master kill-switch (its boolean value
-  supplied by the caller — a Django setting on the backend, an env var on the
-  workers) then the single ``pg_queue_enabled`` Flipt flag, bucketed per org. Fails
-  closed to Celery.
+- :func:`resolve_pg_transport` — the gate: the single ``pg_queue_enabled`` Flipt
+  flag (bucketed per org) is the sole control. Fails closed to Celery on a
+  blind/unreachable Flipt or any error.
 - :class:`RoutingExecutionDispatcher` — picks PG-vs-Celery per call (instant
   rollout/rollback) for every mode; the Celery dispatcher, the PG dispatcher and the
   per-side ``resolve`` are all injected.
--- a/unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
+++ b/unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
@@ -135,22 +135,17 @@ class QueueTransport(Protocol):
 def resolve_pg_transport(
     context: ExecutionContext,
     *,
-    master_gate_enabled: bool,
     flag_key: str = PG_QUEUE_FLAG_KEY,
 ) -> bool:
     """True → route this executor dispatch over PG; False → Celery (default).
 
-    Master-gated by ``master_gate_enabled`` (the caller supplies its value — a Django
-    setting on the backend, an env var on the workers), then the single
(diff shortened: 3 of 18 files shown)


backend/workflow_manager/workflow_v2/transport.py
```diff
@@ -78,22 +75,12 @@ def resolve_transport(
         pipeline_id: Optional, carried in ``context`` for future segment rules.
 
     Returns:
-        A :class:`WorkflowTransport` value string — ``"pg_queue"`` only when the
-        master-gate is on, Flipt is reachable, and Flipt says yes for this
-        execution; ``"celery"`` otherwise (including any error — fail-closed).
+        A :class:`WorkflowTransport` value string — ``"pg_queue"`` only when Flipt
+        is reachable and says yes for this execution; ``"celery"`` otherwise
+        (including any error — fail-closed).
```
muhammad-ali-e, line 80: **[HIGH — Silent failure] The fail-closed `try/except` here is dead code, so a real Flipt outage is silent.**

The docstring promises fail-closed "including any error", and the wrap at lines 123–136 is labelled "defense in depth". But `check_feature_flag_status` (`unstract/flags/src/unstract/flags/feature_flag.py:45-46`) already does a bare `except Exception: return False` **with no logging**. So:

- The `except Exception` branch at lines 129–136 can never execute — the `logger.warning("...Flipt evaluation failed...")` you wrote here will never fire.
- A genuinely broken Flipt (deleted flag, wrong namespace, gRPC connect failure while `FLIPT_SERVICE_AVAILABLE=true`) collapses into the same `False` as a healthy "rollout says no", with **no decision-layer log**.

This PR removes the `PG_QUEUE_TRANSPORT_ENABLED` master gate, so Flipt is now the *sole* interlock — making this blind spot load-bearing.

**Fix:** make `check_feature_flag_status` log + distinguish "evaluation errored" from "flag off" (e.g. log with `exc_info` and/or re-raise on a sentinel) so these resolver warnings can actually fire; or, at minimum, drop the misleading "defense in depth" comment so a future reader doesn't trust a guard that can't trip. Same issue in `scheduler/ownership.py` and `unstract/.../executor_rpc.py`.

backend/scheduler/ownership.py
```diff
@@ -41,19 +40,14 @@
...
     """True → the PG scheduler owns this schedule; False → Celery Beat does.
 
-    Mirrors ``resolve_transport``: master-gated by ``PG_QUEUE_TRANSPORT_ENABLED``
-    (shared PG kill-switch), then the single ``pg_queue_enabled`` Flipt flag, keyed
-    on ``pipeline_id`` for a stable percentage bucket. **Fails closed to Beat**
-    on a closed gate, a blind Flipt, or any error — so a schedule never silently
-    loses its firer.
+    Mirrors ``resolve_transport``: gated by the single ``pg_queue_enabled`` Flipt
+    flag, keyed on ``pipeline_id`` for a stable percentage bucket. **Fails closed to
+    Beat** on a blind Flipt or any error — so a schedule never silently loses its
```
muhammad-ali-e, line 45: **[HIGH — Silent failure] "Fails closed … on any error" is unreachable; a Flipt outage silently leaves schedules on Beat with no log.**

The `except Exception → warn → return False` at lines 66–76 is dead code: `check_feature_flag_status` (`unstract/flags/.../feature_flag.py:45-46`) swallows every exception and returns `False` itself, so the `resolve_schedule_owner: Flipt check failed` warning never fires. A broken/unreachable Flipt is indistinguishable from "rollout says no", and with the master gate now removed there is no other signal that ownership silently reverted to Beat.

**Fix:** same as in `transport.py` — have `check_feature_flag_status` log/propagate evaluation failures so this fail-closed path is observable, or drop the dead `try/except` + comment.

muhammad-ali-e (commented):

Review of the single-flag gating refactor (UN-3618) + transient-connect retry (UN-3619), produced by six specialised agents (Code Reviewer, Silent Failure Hunter, Type Design Analyzer, PR Test Analyzer, Comment Analyzer, Code Simplifier). Net: a clean, mostly-subtractive PR with **no Critical issues**. One **High** silent-failure risk (Flipt is now the sole gate, but a Flipt outage is indistinguishable from "flag off" and logs nothing at the decision layer), plus several Medium/Low items on the new retry logic, test coverage, docs, and dead code. Inline comments below.

unstract/workflow-execution/src/unstract/workflow_execution/executor_rpc.py
```diff
@@ -135,22 +135,17 @@ def wait_for_result(self, reply_key: str, timeout: float) -> ExecResultRow | Non
...
 ) -> bool:
     """True → route this executor dispatch over PG; False → Celery (default).
 
-    Master-gated by ``master_gate_enabled`` (the caller supplies its value — a Django
-    setting on the backend, an env var on the workers), then the single
-    ``pg_queue_enabled`` Flipt flag, bucketed per org. **Fails closed to Celery** on a
-    closed gate, a blind Flipt, or any error — so the executor never silently loses
-    its transport.
+    Gated by the single ``pg_queue_enabled`` Flipt flag, bucketed per org.
+    **Fails closed to Celery** on a blind Flipt or any error — so the executor
```
muhammad-ali-e, line 142: **[HIGH — Silent failure] Fail-closed `try/except` (lines 159–167) is dead code; a Flipt outage silently routes to Celery with no log.**

`check_feature_flag_status` already returns `False` on any exception with no logging, so the `resolve_pg_transport: Flipt check failed` warning here is unreachable. Same root cause and fix as `transport.py`/`ownership.py`.

Separately, the **module docstring (lines ~19–22)** is now stale — it still describes "a master kill-switch … supplied by the caller … then the single `pg_queue_enabled` Flipt flag", but this PR removed the `master_gate_enabled` param. Update it to describe Flipt as the sole gate.

workers/queue_backend/pg_queue/connection.py
```diff
@@ -25,6 +27,37 @@
 
 logger = logging.getLogger(__name__)
 
+# Bounded retry for *transient* connect failures (DB restart, PgBouncer pool
+# wait, brief network partition, "too many clients" spikes) — the common cloud
+# blips. Opening a connection is side-effect-free, so retrying it is safe and
+# purely additive. Note we deliberately do NOT auto-retry the enqueue INSERT
+# (queue_backend.pg_queue.client.send): an ambiguous commit-time failure could
+# double-enqueue → double-dispatch, so that path stays fail-and-surface.
+_DEFAULT_CONNECT_RETRIES = 3  # total attempts (1 = no retry)
```
muhammad-ali-e, line 36: **[MEDIUM — Type design] Group the retry knobs into one validated policy.**

`_DEFAULT_CONNECT_BACKOFF` and `_CONNECT_BACKOFF_CAP` only make sense together, and `RETRIES`/`BACKOFF` are env-tunable while the cap is not. An operator who sets `WORKER_PG_QUEUE_CONNECT_BACKOFF=10` gets every sleep silently clamped to 5.0 — the knob appears to do nothing past the cap, with no warning.

**Fix:** fold the three values into a frozen `ConnectRetryPolicy` dataclass built once from env, with a `sleep_for(attempt)` method and a `__post_init__` that asserts `backoff_cap >= base_backoff` (and either make the cap env-tunable or document it). This puts the backoff math + invariants in one enforced place and trivialises the loop body.

workers/queue_backend/pg_queue/connection.py
```diff
@@ -25,6 +27,37 @@
...
+_CONNECT_BACKOFF_CAP = 5.0  # ceiling on a single inter-attempt sleep
+
+
+def _connect_env[T](suffix: str, default: T, cast: Callable[[str], T]) -> T:
+    """Read ``WORKER_PG_QUEUE_CONNECT_{suffix}``; fall back to ``default``.
+
+    Empty/unset → default; an unparseable value warns and uses the default
+    (a bad knob must not wedge the only direct-DB worker path).
+    """
+    raw = os.getenv(f"WORKER_PG_QUEUE_CONNECT_{suffix}")
```
muhammad-ali-e, line 50: **[LOW — Consistency / reuse] `_connect_env` validation is inconsistent, and it duplicates an existing helper.**

A malformed value (`RETRIES=abc`) warns, but an out-of-range value (`RETRIES=0`, negative `BACKOFF`) is silently coerced by the caller's `max(1, …)`/`max(0.0, …)` (lines 85–86) with no log. For a "bad knob must not wedge it" goal, out-of-range deserves the same visibility as unparseable — move the clamping in here and warn when a parsed value is coerced.

Also note `unstract.core.utilities._safe_get_env_int`/`_safe_get_env_float` already do "read env, parse, warn-and-default" (core is a workers dependency). Not a drop-in (they take a fixed key and bake in non-negative validation), but worth converging on rather than carrying a near-duplicate.

workers/queue_backend/pg_queue/connection.py
```diff
@@ -37,31 +70,58 @@ def create_pg_connection(env_prefix: str = "DB_") -> PgConnection:
...
-            "(host=%s port=%s dbname=%s schema=%s, env_prefix=%r)",
-            host,
-            port,
-            dbname,
-            schema,
-            env_prefix,
-        )
-        raise
+
+    attempts = max(1, _connect_env("RETRIES", _DEFAULT_CONNECT_RETRIES, int))
```
muhammad-ali-e: **[LOW] `attempts` is floored at 1 but has no upper bound.** A fat-fingered `WORKER_PG_QUEUE_CONNECT_RETRIES=100000` makes a worker block for a very long time (per-sleep is capped at 5s, but the count isn't) on what can be a startup/hot path. Consider `attempts = min(<sane_cap>, max(1, _connect_env(...)))`, or document the expected range in the knob comment.

workers/queue_backend/pg_queue/connection.py
```diff
@@ -37,31 +70,58 @@ def create_pg_connection(env_prefix: str = "DB_") -> PgConnection:
...
+        try:
+            return psycopg2.connect(
+                host=host,
+                port=port,
+                dbname=dbname,
+                user=user,
+                password=os.getenv(f"{env_prefix}PASSWORD", "unstract_pass"),
+                options=f"-c search_path={schema}",
+            )
+        except psycopg2.OperationalError:
```
muhammad-ali-e, line 131: **[LOW] `psycopg2.OperationalError` is broad — permanent misconfig is retried before surfacing.** Auth failures (`password authentication failed`), `database "x" does not exist`, and bad-host all raise `OperationalError`, not a subclass. A permanently-misconfigured `DB_*` var burns all attempts + ~1.5s of backoff on every `create_pg_connection` call (invoked by several long-running loops) before raising. The final `logger.error` + `raise` does surface it, so this is latency/log-noise, not a silent failure. Optional: skip retry for clearly-permanent SQLSTATEs (`28xxx` auth, `3D000` invalid db) via `exc.pgcode`, or tighten the docstring to say auth/db-name errors are retried-then-raised.

workers/queue_backend/pg_queue/connection.py
```diff
@@ -37,31 +70,58 @@ def create_pg_connection(env_prefix: str = "DB_") -> PgConnection:
...
+                "(host=%s port=%s dbname=%s schema=%s, env_prefix=%r)",
+                attempts,
+                host,
+                port,
+                dbname,
+                schema,
+                env_prefix,
+            )
+            raise
+    raise AssertionError("unreachable: loop returns or raises")  # pragma: no cover
```
muhammad-ali-e: **[LOW — Simplification] Drop the unreachable `AssertionError` sentinel by restructuring the loop.** The loop already returns on success or raises on the final attempt, so this `# pragma: no cover` sentinel only exists because the terminal raise lives *inside* the loop. Invert the guard so the loop body handles only the retryable case and the terminal `logger.error(...); raise` is reached structurally on the last iteration — functionally identical, no sentinel, no pragma:

```python
except psycopg2.OperationalError:
    if attempt >= attempts:
        logger.error(...)
        raise
    sleep_for = min(backoff * (2 ** (attempt - 1)), _CONNECT_BACKOFF_CAP)
    logger.warning(...)
    time.sleep(sleep_for)
```

workers/queue_backend/pg_queue/executor_rpc.py
```diff
@@ -48,31 +47,13 @@
 
 logger = logging.getLogger(__name__)
```
muhammad-ali-e: **[LOW — Dead code] `logger` (and `import logging` at line ~17) is now unused.** Its only consumer was the fat-fingered-env warning this PR deletes; the backend counterpart (`backend/pg_queue/executor_rpc.py`) correctly carries no logger. Ruff won't flag a module-level unused binding, so this is cosmetic — remove both lines to keep the module clean (or keep if you anticipate logging here soon).

workers/sample.env
```diff
@@ -474,14 +474,6 @@ EVALUATION_SERVER_IP=unstract-flipt
...
-# PG_QUEUE_TRANSPORT_ENABLED: the file_processing worker reads it to route the
-# structure_tool executor RPC onto the PG queue. Routing onto PG requires ALL
-# THREE: this gate True, FLIPT_SERVICE_AVAILABLE=True, and the Flipt flag
-# `pg_queue_enabled` on. Keep False until the worker-pg-executor consumer is
-# running in the fleet; set False for instant rollback.
-PG_QUEUE_TRANSPORT_ENABLED=False
-
 # PG-queue consumer prefork concurrency (UN-3606). >1 makes the consumer launcher
 # fork N isolated consumer children — the PG analogue of Celery's prefork
 # --concurrency, so file batches run in parallel. SKIP LOCKED distributes work
```
muhammad-ali-e, line 479: **[LOW — Docs] The two new operator knobs are undocumented.** This PR adds `WORKER_PG_QUEUE_CONNECT_RETRIES` and `WORKER_PG_QUEUE_CONNECT_BACKOFF` (read in `connection.py:_connect_env`) but documents neither here, while sibling PG-queue knobs (e.g. `WORKER_PG_REAPER_INTERVAL_SECONDS`) are documented by convention. Add commented entries near the other `WORKER_PG_*` settings noting the defaults (3 attempts / 0.5s base), the doubling behaviour, and the 5.0s cap.

workers/tests/test_pg_connection.py
```diff
@@ -0,0 +1,114 @@
...
+        def connect(**_kwargs):
+            calls.append(1)
+            return sentinel
+
+        sleeps = _patch(monkeypatch, connect=connect)
+        assert conn_mod.create_pg_connection() is sentinel
+        assert len(calls) == 1
+        assert sleeps == []
+
+    def test_retries_transient_then_succeeds(self, monkeypatch):
```
muhammad-ali-e, line 44: **[MEDIUM — Test coverage gaps] The new suite covers the core paths well but misses four behavioural branches of the retry logic:**

1. **Backoff cap** (`connection.py:100`) — no test drives a sleep past 5.0s, so the `min(…, _CONNECT_BACKOFF_CAP)` clamp (a real production guarantee) has zero coverage. Add e.g. `retries=5, backoff=4` → assert `sleeps == [4.0, 5.0, 5.0, 5.0]`.
2. **Unset-env default path** (`connection.py:47-49`) — `_patch` always `setenv`s both knobs, so the *production* default (env absent → 3 attempts / 0.5s) is never exercised. Add a `delenv` test.
3. **Invalid / negative `BACKOFF`** (`connection.py:52-59,86`) — only the `RETRIES` cast-failure is tested; add `backoff=xyz` (→ default) and `backoff=-1` (→ `[0.0, 0.0]`, retry-without-sleep).
4. **Connection params forwarded** (`connection.py:90-97`) — every test discards `connect` kwargs; nothing asserts `options=-c search_path=…` (or host/db) is passed and stable across retries.

Also: `test_retries_transient_then_succeeds`'s `sleeps == [0.5, 1.0]` can't distinguish exponential from linear growth — add one ≥3-sleep case (e.g. `[1.0, 2.0, 4.0]`) to pin the geometric contract.

muhammad-ali-e:
Thanks for the thorough review — addressed in `e78429c11`. Summary by finding:

### Fixed

- **[HIGH] Silent Flipt outage (×3 sites).** Root-caused it: `check_feature_flag_status` (`feature_flag.py:46`) swallowed every exception and returned `False` with no log, so a genuine outage was indistinguishable from "flag off" — and, as you noted, that's now load-bearing since Flipt is the sole gate. Fixed at the source: both `check_feature_flag_status` and `check_feature_flag_variant` now `logger.warning(..., exc_info=True)` before returning the disabled default (warning + exc_info, not `logger.exception`, so a persistently-down Flipt doesn't bury every call as a Sentry error). This makes outages observable for all three resolvers at once. I kept the resolvers' own `try/except` guards — their comments already state they're defense-in-depth against a *future* change to the helper's contract (not claiming the warning fires today), which I think is correct cheap insurance rather than dead code.
- **Stale shared `executor_rpc.py` docstring** (the "master kill-switch supplied by the caller" lines) — updated to describe Flipt as the sole gate.
- **Dead `logger` / `import logging`** in workers `executor_rpc.py` — removed both.
- **`attempts` unbounded + silently-clamped knobs** — added `_connect_retry_policy()` that clamps `RETRIES` to `[1, 10]` and `BACKOFF` to `>= 0`, **warning** on each clamp (now consistent with the unparseable-value path). The `5.0s` per-sleep cap is documented as a fixed ceiling.
- **Unreachable `AssertionError` sentinel** — gone; inverted the loop so the terminal `logger.error(...); raise` is reached structurally on the last attempt.
- **`OperationalError` breadth** — documented in the docstring that permanent misconfigs (auth/unknown-db/bad-host) are retried-then-raised and surfaced by the final `logger.error` (your offered lighter option, vs. `pgcode` filtering).
- **Undocumented knobs** — `WORKER_PG_QUEUE_CONNECT_RETRIES` / `_BACKOFF` now documented in `workers/sample.env` near the sibling `WORKER_PG_*` settings (defaults, doubling, 5.0s cap).
- **Test coverage gaps** — added: backoff-cap clamp, unset-env production-default path, invalid + negative `BACKOFF`, attempts-above-max clamp, a ≥3-sleep geometric-growth case (pins exponential vs linear), and connection-params-forwarded-and-stable-across-retries. Connection suite is now 13 tests (all green; backend 35, workers 50 overall).

### Declined (with rationale)

- **[MEDIUM] Fold the retry knobs into a `ConnectRetryPolicy` dataclass.** Felt like over-engineering for three module constants + one function. I addressed the underlying concern instead — the silent-clamp + cap-discoverability — via `_connect_retry_policy()` (warns on clamp) and the documented cap.
- **[LOW] Converge on `unstract.core._safe_get_env_int/float`.** As you noted, not a drop-in (fixed key + baked-in non-negative validation); the tailored `_connect_env` + `_connect_retry_policy` is small and avoids a new cross-package coupling / behaviour change. Noted for a future consolidation if those helpers generalise.

ruff check + format clean; pre-commit clean.

PR #2121, opened 2026-06-29
UN-3651 [FIX] PG barrier enqueue — reconnect-retry on stale cached connection

muhammad-ali-e:
## What
A transient `psycopg2.OperationalError: server closed the connection unexpectedly` during the **PgBarrier enqueue** (the barrier UPSERT) aborted the whole ETL execution → status `ERROR` with 0 files created.

## Root cause
PgBarrier keeps a **thread-local cached** Postgres connection reused across barrier ops. While idle *between* ops it can be reaped server-side (PgBouncer `server_idle_timeout` / DB failover / TCP teardown), and `_get_conn` can't tell — `conn.closed` is a client-side flag only — so the first statement after the idle gap fails. The existing `_cursor` *recovers* the dead conn (so the next call reconnects) but does **not retry** the failed op → the execution aborts.

## Fix
A one-shot reconnect-retry (`_run_idempotent_write`, attempts=2) scoped to the **idempotent, pre-dispatch** barrier write only — the UPSERT (`ON CONFLICT … DO UPDATE` → same row/state) + the per-execution dedup reset. Re-running them after an ambiguous commit is a no-op, and no header has been dispatched yet, so a retry can **neither duplicate a row nor double-dispatch work**.

**Idempotency boundary preserved:** the non-idempotent **decrement** (`remaining - 1`) and `claim_batch` stay on the plain `_cursor` (recover-but-don't-retry) — a re-applied decrement could fire the callback early, so it must never auto-retry.

## Scope / non-regression
Workers only (`pg_barrier.py`), under the `pg_queue_enabled` flag — PgBarrier runs **only** on the PG transport. Flag OFF → `CeleryChordBarrier`, this code never runs → zero Celery-path regression. The change is additive on the PG happy path (first attempt succeeds, identical to before); the only behavior change is that a previously-fatal transient drop now self-heals.

## Verification
- **Unit:** +3 tests (retry-then-succeed, no-retry-on-non-connection-error, reraise-after-exhaust). Full PgBarrier suite **54 passed**; ruff clean.
- **Live A/B against real Postgres** (rebuilt worker image), reproducing the exact incident error via `pg_terminate_backend`:
  - **WITHOUT fix** (old `with _cursor()` path): raises → barrier row not written (0) → execution aborts.
  - **WITH fix**: logs `reconnecting and retrying` → barrier row written (1) → execution proceeds.
  - **Control:** plain `_cursor` (decrement path) raises with **no** retry → boundary intact.

Ref: UN-3651

### DIFF
--- a/workers/queue_backend/pg_barrier.py
+++ b/workers/queue_backend/pg_barrier.py
@@ -128,6 +128,53 @@ def _cursor() -> Iterator[Any]:
         raise
 
 
+# One retry for an IDEMPOTENT barrier write. The thread-local connection is
+# cached across barrier ops, so it can be reaped server-side (PgBouncer
+# server_idle_timeout / DB failover) while sitting idle BETWEEN ops — and
+# ``_get_conn`` can't tell, since ``conn.closed`` is a client-side flag only.
+# The first statement after the idle gap then fails; ``_cursor`` discards the
+# dead conn, so a single retry runs against a freshly reconnected one. This
+# turns a transient blip (which previously aborted the whole execution at
+# barrier enqueue) into a self-heal.
+_BARRIER_WRITE_ATTEMPTS = 2
+
+
+def _run_idempotent_write(operation: Callable[[Any], None], *, what: str) -> None:
+    """Run ``operation(cur)`` in a committed cursor, retrying ONCE if the cached
+    connection was dead.
+
+    Restricted to **idempotent** statements run BEFORE any task dispatch — the
+    barrier UPSERT (``ON CONFLICT … DO UPDATE`` → same row, same state) + the
+    per-execution dedup reset (``DELETE WHERE execution_id``). Re-running them
+    after an ambiguous commit is a no-op, so a retry can neither duplicate a row
+    nor double-dispatch work (no header has been enqueued yet).
+
+    Deliberately NOT used for the barrier **decrement** (``remaining =
+    remaining - 1``): that is not idempotent — a re-applied decrement could drive
+    ``remaining`` to 0 early and fire the callback twice — so it stays on the
+    plain :func:`_cursor` (recover-but-don't-retry). Same for ``claim_batch``,
+    whose ``RETURNING`` answer flips on a retry.
+    """
+    for attempt in range(1, _BARRIER_WRITE_ATTEMPTS + 1):
+        try:
+            with _cursor() as cur:
+                operation(cur)
+            return
+        except (psycopg2.OperationalError, psycopg2.InterfaceError):
+            # _cursor already dropped the dead thread-local conn → the next
+            # _get_conn() reconnects. Retry once; re-raise if it still fails
+            # (a genuinely-down DB surfaces as ERROR, as before).
+            if attempt >= _BARRIER_WRITE_ATTEMPTS:
+                raise
+            logger.warning(
+                "PgBarrier: %s lost its DB connection (attempt %d/%d); "
+                "reconnecting and retrying",
+                what,
+                attempt,
+                _BARRIER_WRITE_ATTEMPTS,
+            )
+
+
 def _delete_barrier(execution_id: str) -> None:
     with _cursor() as cur:
         cur.execute(
@@ -311,7 +358,7 @@ class PgBarrier:
                     f"could not be reaper-recovered. This is a bug in the caller."
                 )
 
-            with _cursor() as cur:
+            def _reset_barrier(cur: Any) -> None:
                 # UPSERT clears any leftover state from a prior run with this id.
                 # No inline expiry sweep here — an unbounded global DELETE on the
                 # enqueue hot path risks lock contention / deadlocks between
@@ -338,6 +385,11 @@ class PgBarrier:
                     f"DELETE FROM {qualified('pg_batch_dedup')} WHERE execution_id = %s",
                     (execution_id,),
                 )
+
+            # Both statements are idempotent and run BEFORE any header dispatch,
+            # so a one-shot retry on a stale cached connection can't duplicate
+            # state or double-dispatch — see _run_idempotent_write.
+            _run_idempotent_write(_reset_barrier, what=f"enqueue exec={execution_id}")
 
             self._dispatch_headers(
                 header_tasks,
--- a/workers/queue_backend/pg_barrier.py
+++ b/workers/queue_backend/pg_barrier.py
@@ -60,9 +60,10 @@ import contextlib
 import json
 import logging
 import threading
+import time
 from collections.abc import Callable, Iterator
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Final
 
 import psycopg2
 import psycopg2.extensions
@@ -88,6 +89,7 @@ from .pg_queue.schema import qualified
 if TYPE_CHECKING:
     from celery.canvas import Signature
     from psycopg2.extensions import connection as PgConnection
+    from psycopg2.extensions import cursor as PgCursor
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


workers/queue_backend/pg_barrier.py
```diff
@@ -128,6 +128,53 @@ def _cursor() -> Iterator[Any]:
...
+                operation(cur)
+            return
+        except (psycopg2.OperationalError, psycopg2.InterfaceError):
+            # _cursor already dropped the dead thread-local conn → the next
+            # _get_conn() reconnects. Retry once; re-raise if it still fails
+            # (a genuinely-down DB surfaces as ERROR, as before).
+            if attempt >= _BARRIER_WRITE_ATTEMPTS:
+                raise
+            logger.warning(
+                "PgBarrier: %s lost its DB connection (attempt %d/%d); "
```
muhammad-ali-e: **[HIGH — silent failure / observability]** When the retry succeeds, the original `OperationalError`/`InterfaceError` is gone forever — no `exc_info`, and the exception type/message isn't interpolated. You can't tell a benign PgBouncer idle reap from a recurring statement timeout or deadlock that the retry is silently papering over. The message also hard-asserts `"lost its DB connection"`, which (see line 163) will be a lie for many of the conditions this `except` actually catches.

Suggested:
```python
except (psycopg2.OperationalError, psycopg2.InterfaceError) as exc:
    if attempt >= _BARRIER_WRITE_ATTEMPTS:
        raise
    logger.warning(
        "PgBarrier: %s — DB write failed (%s: %s); cached connection "
        "was likely stale, reconnecting and retrying (attempt %d/%d)",
        what, type(exc).__name__, exc, attempt, _BARRIER_WRITE_ATTEMPTS,
    )
```

workers/queue_backend/pg_barrier.py
```diff
@@ -128,6 +128,53 @@ def _cursor() -> Iterator[Any]:
...
+    ``remaining`` to 0 early and fire the callback twice — so it stays on the
+    plain :func:`_cursor` (recover-but-don't-retry). Same for ``claim_batch``,
+    whose ``RETURNING`` answer flips on a retry.
+    """
+    for attempt in range(1, _BARRIER_WRITE_ATTEMPTS + 1):
+        try:
+            with _cursor() as cur:
+                operation(cur)
+            return
+        except (psycopg2.OperationalError, psycopg2.InterfaceError):
```
muhammad-ali-e: **[MEDIUM — broad catch + no backoff]** `psycopg2.OperationalError` is the base class for many *server-side, non-connection* conditions: `QueryCanceled` (statement_timeout), `DeadlockDetected`/`SerializationFailure` (via `TransactionRollbackError`), `TooManyConnections`, `AdminShutdown`, etc. All are now caught, mislabeled (line 170), and **immediately re-run with zero backoff** — which just adds load against an already-struggling DB. Re-running the idempotent UPSERT once is correctness-safe, so this isn't a bug, but consider either (a) narrowing the retry to genuine connection-drop signals (e.g. only when `_cursor` actually discarded the cached conn), or at minimum (b) excluding `TransactionRollbackError` so deadlocks aren't blindly hammered. A small fixed sleep before the retry would also widen the self-heal window during a DB failover (the idle-reap case succeeds instantly either way).

workers/queue_backend/pg_barrier.py
```diff
@@ -128,6 +128,53 @@ def _cursor() -> Iterator[Any]:
...
+
+    Restricted to **idempotent** statements run BEFORE any task dispatch — the
+    barrier UPSERT (``ON CONFLICT … DO UPDATE`` → same row, same state) + the
+    per-execution dedup reset (``DELETE WHERE execution_id``). Re-running them
+    after an ambiguous commit is a no-op, so a retry can neither duplicate a row
+    nor double-dispatch work (no header has been enqueued yet).
+
+    Deliberately NOT used for the barrier **decrement** (``remaining =
+    remaining - 1``): that is not idempotent — a re-applied decrement could drive
+    ``remaining`` to 0 early and fire the callback twice — so it stays on the
```
muhammad-ali-e: **[MEDIUM — comment inaccuracy]** "fire the callback twice" isn't the actual failure mode. Decrements each run in their own committed txn, serialized by the row lock, and the callback fires only when a decrement's `RETURNING` observes `remaining == 0`. A double-applied decrement *skips over* 0 (`1→0` committed-but-unacked, retry `0→-1` → returns -1 → abandoned branch), so two observations of exactly 0 can't occur. The real hazards — still justifying the exclusion — are firing the callback **prematurely with incomplete results**, or skipping past 0 and **stranding the barrier to expiry**. Suggest: "...double-counts: it can drive `remaining` to 0 early and fire the callback prematurely with incomplete results, or skip past 0 and strand the barrier."

muhammad-ali-e (commented):
Automated review via PR Review Toolkit (code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier).

**Verdict: the fix is correct and the idempotency boundary holds.** The retry control flow is sound (bounded, re-raises on exhaustion, no silent-loss path), the decrement/`claim_batch` correctly stay off the retry path, and all the docstring's structural claims (pre-dispatch ordering, plain-`_cursor` exclusions) check out. Findings below are about **observability and test/comment precision**, not correctness. None are blocking.

Prioritised:
- **HIGH** — retry warning discards the triggering exception and hard-asserts "lost its DB connection" (pg_barrier.py:170)
- **HIGH** — the `enqueue()` self-heal wiring has no test; reverting to the old inline `_cursor()` would still pass everything (test:173)
- **MEDIUM** — catch is broader than stale-connection; retries server-side errors with no backoff (pg_barrier.py:163)
- **MEDIUM** — "fire the callback twice" rationale is inaccurate (pg_barrier.py:154); `InterfaceError` branch untested (test:180); idempotency invariant is docs-only (pg_barrier.py:142)
- **LOW** — `Final`/naming, "same state" overstatement, redundant call-site comment, dead test counters

workers/queue_backend/pg_barrier.py
```diff
@@ -128,6 +128,53 @@ def _cursor() -> Iterator[Any]:
...
+# barrier enqueue) into a self-heal.
+_BARRIER_WRITE_ATTEMPTS = 2
+
+
+def _run_idempotent_write(operation: Callable[[Any], None], *, what: str) -> None:
+    """Run ``operation(cur)`` in a committed cursor, retrying ONCE if the cached
+    connection was dead.
+
+    Restricted to **idempotent** statements run BEFORE any task dispatch — the
+    barrier UPSERT (``ON CONFLICT … DO UPDATE`` → same row, same state) + the
```
muhammad-ali-e: **[LOW — comment precision]** "same row, same state" slightly overstates idempotency: the UPSERT's `DO UPDATE` re-runs `created_at = now()` and `expires_at = ...` (lines 375-376), so timestamps advance on a retry. Harmless here (retries are milliseconds apart, `execution_id` is the PK so no row duplication), but a future reader relying on `created_at` for staleness/reaper logic could be surprised. Suggest softening to "→ same row, `remaining`/`results` reset identically (timestamps refresh, harmlessly)."

workers/queue_backend/pg_barrier.py
```diff
@@ -128,6 +128,53 @@ def _cursor() -> Iterator[Any]:
...
+# server_idle_timeout / DB failover) while sitting idle BETWEEN ops — and
+# ``_get_conn`` can't tell, since ``conn.closed`` is a client-side flag only.
+# The first statement after the idle gap then fails; ``_cursor`` discards the
+# dead conn, so a single retry runs against a freshly reconnected one. This
+# turns a transient blip (which previously aborted the whole execution at
+# barrier enqueue) into a self-heal.
+_BARRIER_WRITE_ATTEMPTS = 2
+
+
+def _run_idempotent_write(operation: Callable[[Any], None], *, what: str) -> None:
```
muhammad-ali-e: **[MEDIUM — type design]** The defining invariant — that `operation` must be *idempotent* — lives entirely in the docstring; `Callable[[Any], None]` would happily accept the non-idempotent decrement. The `-> None` return is a nice structural guard (it blocks passing a `RETURNING`-reading op like `claim_batch`), but idempotency itself is unenforceable in Python's type system. Pragmatic mitigations: keep it private (done) and consider a name that makes the contract impossible to miss at call sites, e.g. `_run_idempotent_pre_dispatch_write`.

Also minor: the cursor is typed `Any` here, but the module already imports `connection as PgConnection` under `TYPE_CHECKING` — adding `cursor as PgCursor` and typing `Callable[[PgCursor], None]` would tighten intent at zero runtime cost (worth doing file-wide for consistency).

workers/queue_backend/pg_barrier.py
```diff
@@ -128,6 +128,53 @@ def _cursor() -> Iterator[Any]:
...
 
+# One retry for an IDEMPOTENT barrier write. The thread-local connection is
+# cached across barrier ops, so it can be reaped server-side (PgBouncer
+# server_idle_timeout / DB failover) while sitting idle BETWEEN ops — and
+# ``_get_conn`` can't tell, since ``conn.closed`` is a client-side flag only.
+# The first statement after the idle gap then fails; ``_cursor`` discards the
+# dead conn, so a single retry runs against a freshly reconnected one. This
+# turns a transient blip (which previously aborted the whole execution at
+# barrier enqueue) into a self-heal.
+_BARRIER_WRITE_ATTEMPTS = 2
```
muhammad-ali-e: **[LOW — polish]** `_BARRIER_WRITE_ATTEMPTS = 2` means "total attempts", so `2` = "one retry" — a name/value relationship only the docstring disambiguates. Suggest `_BARRIER_WRITE_ATTEMPTS: Final = 2  # total attempts: 1 initial + 1 retry`. Keeping it a literal (not env-driven) is the right call — it keeps the idempotency bound from being weakened operationally.

workers/queue_backend/pg_barrier.py
```diff
@@ -339,6 +386,11 @@ def enqueue(
                     (execution_id,),
                 )
 
+            # Both statements are idempotent and run BEFORE any header dispatch,
+            # so a one-shot retry on a stale cached connection can't duplicate
+            # state or double-dispatch — see _run_idempotent_write.
```
muhammad-ali-e: **[LOW — redundancy]** This 3-line comment restates the `_run_idempotent_write` docstring it points to (directly above the call). It's the kind of duplicated rationale that rots independently of the source of truth. Suggest trimming to a one-liner pointer, e.g. `# Idempotent + pre-dispatch → safe to retry; see _run_idempotent_write.`

workers/tests/test_pg_barrier.py
```diff
@@ -115,6 +116,121 @@ def test_missing_execution_id_raises(self):
...
+
+@pytest.fixture
+def _clean_local():
+    """Ensure the module thread-local connection is reset around each test."""
+    pg_barrier._local.conn = None
+    yield
+    pg_barrier._local.conn = None
+
+
+class TestIdempotentWriteRetry:
```
muhammad-ali-e, line 173: **[HIGH — test gap]** All three tests exercise `_run_idempotent_write` in isolation with a no-op operation — none go through the actual `enqueue()` path the fix protects. Concrete regression this misses: reverting `enqueue` to the old inline `with _cursor() as cur:` (the exact pre-fix code) leaves `_run_idempotent_write` as dead code and silently removes the self-heal, **yet every test here still passes**. Suggest an integration test using the real `barrier_db` fixture: set `pg_barrier._local.conn` to a stub whose first `.execute` raises `OperationalError`, monkeypatch `create_pg_connection` to return the real test connection, call `PgBarrier().enqueue(...)`, then assert (a) the `pg_barrier_state` row exists exactly once with `remaining == N`, and (b) header dispatch was called exactly N times. This pins both the self-heal *and* the "headers still dispatched, no double-dispatch" idempotency claim end-to-end.

workers/tests/test_pg_barrier.py
```diff
@@ -115,6 +116,121 @@ def test_missing_execution_id_raises(self):
...
+    """`_run_idempotent_write` self-heals a stale cached connection with ONE
+    retry — the fix for the `server closed the connection unexpectedly` abort at
+    barrier enqueue. It must retry ONLY on connection errors, stay bounded, and
+    never silently swallow a real (non-connection) failure.
+    """
+
+    def test_retries_once_on_dead_connection_then_succeeds(
```
muhammad-ali-e, line 193: **[MEDIUM — test gap]** Both `except` clauses (this fix at pg_barrier.py:163 and `_cursor` at :119) catch `(OperationalError, InterfaceError)`, but all three tests raise only `OperationalError`. `InterfaceError` ("connection already closed") is arguably the *more* common psycopg2 symptom of a stale cached socket — exactly this PR's target scenario. Narrowing either clause to drop `InterfaceError` would break the self-heal with no failing test. Suggest parametrizing this test over `[psycopg2.OperationalError("server closed..."), psycopg2.InterfaceError("connection already closed")]`. While here, also asserting `dead.commits == 0` (no partial commit on the failed attempt) and a `caplog` warning would lock the "log on retry, not on success" contract.

workers/tests/test_pg_barrier.py
```diff
@@ -115,6 +116,121 @@ def test_missing_execution_id_raises(self):
...
+    """Minimal psycopg2-connection stub. ``execute_error`` (if set) is raised by
+    every cursor.execute on this connection — simulating a stale/dead socket.
+    """
+
+    def __init__(self, *, execute_error=None):
+        self.closed = False
+        self._execute_error = execute_error
+        self.commits = 0
+        self.rollbacks = 0
+        self.executes = 0
```
muhammad-ali-e, line 145: **[LOW — dead test scaffolding]** `_FakeConn` increments `self.rollbacks` (line 159) and `self.executes` (line 149) but neither counter is ever asserted anywhere — test 1 counts invocations via its own `attempts` list. The `rollback()` *method* must stay (called by `_cursor`'s recovery path), but the two counters are dead state. Either drop them, or better, assert on them (e.g. `dead.executes == 1`, `healthy.executes == 1`) to tightly bound the retry to exactly one extra attempt. Also `closed` is `bool` here but real psycopg2 `connection.closed` is an `int` (0/1/2) — fine for the truthiness check, minor fidelity drift.

muhammad-ali-e:
Thanks for the thorough review 🙏 — all findings addressed in `b68e7e6fc` (none were blocking; the fix itself was confirmed correct). Summary:

**Observability (HIGH + greptile P2)**
- Retry warning now captures the exception, names the real error (`type(exc).__name__: exc`), and keeps the traceback (`exc_info=True`). Reworded so it no longer asserts "lost its DB connection" — the broad `OperationalError` catch can't be sure of that.

**Broad catch / no backoff (MEDIUM)**
- Added a small fixed backoff (`_BARRIER_RETRY_BACKOFF_SECONDS = 0.5`) before the retry — widens the self-heal window on a brief failover and avoids immediately re-hammering a struggling DB on the server-side conditions the catch also covers.
- Kept the broad `(OperationalError, InterfaceError)` catch deliberately: a single retry of an *idempotent, pre-dispatch* write is correctness-safe for all of them, and narrowing risks dropping a genuine stale-socket signal (esp. `InterfaceError`). The accurate log + backoff cover the concern.

**Type design / contract (MEDIUM)**
- Renamed `_run_idempotent_write` → `_run_idempotent_pre_dispatch_write` and typed the op `Callable[[PgCursor], None]`, so the idempotent + pre-dispatch contract is loud at the call site (it can't be type-enforced; the `-> None` already blocks a `RETURNING`-reading op).

**Comment precision (MEDIUM + LOW)**
- Decrement hazard reworded: *premature/incomplete-results callback, or skip-past-0 and strand the barrier* (not "fire twice").
- Softened "same row, same state" → `remaining`/`results` reset identically, timestamps refresh harmlessly.
- `_BARRIER_WRITE_ATTEMPTS: Final = 2  # total attempts: 1 initial + 1 retry`.
- Trimmed the redundant call-site comment to a one-line pointer.

**Tests (HIGH + MEDIUM + LOW)**
- Added an **end-to-end self-heal test through `enqueue()`** against the real DB: first write hits a dead cached conn → reconnects → asserts the `pg_barrier_state` row lands exactly once with `remaining == N` **and** every header dispatched exactly once. This pins the wiring — reverting `enqueue()` to the inline `with _cursor()` now fails (the no-DB tests alone wouldn't catch that).
- Parametrized the retry test over **`OperationalError` and `InterfaceError`**.
- Now assert no partial commit (`commits == 0`), exactly one extra attempt (`executes` counters), and the on-retry warning via `caplog`.

Full PgBarrier suite: **56 passed**; ruff clean. SonarCloud gate green.

PR #2125, opened 2026-06-30
UN-3655 [FIX] HITL queue — populate execution_id on enqueue (QueueResult)

muhammad-ali-e:
## What

Carries the workflow `execution_id` into the HITL queue message so the existing `hitl_queue.execution_id` column stops being NULL. No schema change — the column already exists; it was simply never populated.

## Why

`hitl_queue.execution_id` is NULL on **100% of rows** — 2301/2301 in production (since 2026-02-03) and on both Celery and PG transports. A review item therefore can't be traced to its `workflow_execution` without joining through `workflow_file_execution`.

Root cause: the ETL manual-review push serialises a `QueueResult` into the queue message, but `QueueResult` carried `workflow_id` + `file_execution_id` and **no** `execution_id` — so `_push_data_to_queue` omitted it even though `self.execution_id` is in scope. The backend manual-review enqueue **already** reads `message["execution_id"]`, so the column persisted NULL only because the message lacked the key.

## How

- Add `execution_id` to the `QueueResult` dataclass + `to_dict()`.
- Pass `self.execution_id` at the destination push site.

The API/packet path already includes `execution_id` in its message, so this brings the ETL path to parity. **OSS-only — no backend/cloud change** (the backend mapping already exists). Additive: the field defaults to `None`, so callers that don't set it are unaffected; the message gains one key (consumers read via `dict.get`, message is JSONB).

Transport-agnostic — also fixes the existing Celery path.

## Tests

`test_queue_result_execution_id.py` (2): `to_dict()` carries `execution_id` (set), and emits the key as `None` by default (no `KeyError`, purely additive).

## No-regression verification

Full worker suite: 1070 passed. The 28 full-suite failures are **pre-existing, environmental** (`testdb` connection failures + `MagicMock`-not-JSON usage mocks + test-ordering), proven by A/B: running the same 28 nodes as a subset on the base vs with this change yields a **byte-identical** failure set (27 fail, 1 passes) — this change has zero effect on any test. (One real-DB consumer test, `test_pg_queue_consumer.py`, hangs locally and was excluded; it's unrelated to `QueueResult`.)

E2E validates on deploy: new worker image + the pre-existing backend mapping fill the column — new `hitl_queue` rows will have `execution_id` set instead of NULL.

### DIFF
--- a/workers/shared/models/result_models.py
+++ b/workers/shared/models/result_models.py
@@ -223,6 +223,11 @@ class QueueResult:
     file_content: str | None = None
     whisper_hash: str | None = None
     file_execution_id: str | None = None
+    # Workflow execution id — carried into the HITL queue message so the backend
+    # can persist hitl_queue.execution_id (it reads message["execution_id"]).
+    # Without it the column is left NULL and a review item can't be traced back
+    # to its execution without joining through workflow_file_execution.
+    execution_id: str | None = None
     enqueued_at: float | None = None
     ttl_seconds: int | None = None
     extracted_text: str | None = None
@@ -251,6 +256,7 @@ class QueueResult:
             "workflow_id": self.workflow_id,
             "file_content": self.file_content,
             "file_execution_id": self.file_execution_id,
+            "execution_id": self.execution_id,
             "enqueued_at": self.enqueued_at,
             "ttl_seconds": self.ttl_seconds,
             "extracted_text": self.extracted_text,
--- a/workers/shared/models/result_models.py
+++ b/workers/shared/models/result_models.py
@@ -3,6 +3,7 @@
 Dataclasses for task execution results.
 """
 
+import logging
 import os
 import sys
 import time
@@ -18,6 +19,8 @@ from unstract.core.worker_models import FileExecutionResult
 
 # Import worker enums
 from ..enums import WebhookStatus
+
+logger = logging.getLogger(__name__)
 
 
 @dataclass
@@ -223,10 +226,14 @@ class QueueResult:
     file_content: str | None = None
     whisper_hash: str | None = None
     file_execution_id: str | None = None
-    # Workflow execution id — carried into the HITL queue message so the backend
-    # can persist hitl_queue.execution_id (it reads message["execution_id"]).
-    # Without it the column is left NULL and a review item can't be traced back
-    # to its execution without joining through workflow_file_execution.
+    # Workflow execution id — carried into the HITL queue message so the
+    # manual-review consumer (pluggable_apps.manual_review_v2, a separate
+    # codebase) can populate hitl_queue.execution_id. The durable local fact is
+    # that to_dict() always emits this key (None when unset); the key name must
+    # stay in sync with that consumer. Optional with a None default because the
+    # connector's execution_id is itself nullable on some paths — a missing
+    # value is logged in __post_init__ rather than raised, so an unexpected NULL
+    # is visible without breaking the HITL enqueue.
     execution_id: str | None = None
     enqueued_at: float | None = None
     ttl_seconds: int | None = None
@@ -246,6 +253,16 @@ class QueueResult:
             raise ValueError("QueueResult requires a valid workflow_id")
         if self.status is None:
             raise ValueError("QueueResult requires a valid status")
+        # execution_id should always be set on the HITL enqueue path; it's not a
+        # hard requirement (the connector's value is nullable on some paths and a
+        # raise would break manual-review enqueue), so surface a missing value
+        # rather than silently writing a NULL hitl_queue.execution_id.
+        if not self.execution_id:
+            logger.warning(
+                "QueueResult for file=%r has no execution_id; "
+                "hitl_queue.execution_id will be NULL",
+                self.file,
+            )
 
     def to_dict(self) -> Any:
         result_dict = {
--- a/workers/shared/workflow/destination_connector.py
+++ b/workers/shared/workflow/destination_connector.py
@@ -1833,6 +1833,7 @@ class WorkerDestinationConnector:
                 workflow_id=str(self.workflow_id),
                 whisper_hash=whisper_hash,
                 file_execution_id=file_execution_id,
+                execution_id=self.execution_id,
                 extracted_text=extracted_text,
                 ttl_seconds=ttl_seconds,
                 hitl_reason=hitl_reason,
--- a/workers/tests/test_queue_result_execution_id.py
+++ b/workers/tests/test_queue_result_execution_id.py
@@ -1,0 +1,47 @@
+"""UN-3655: ``QueueResult`` carries ``execution_id`` into the HITL queue message.
+
+The ETL manual-review push serialises a ``QueueResult`` to the queue message; the
+backend manual-review enqueue persists ``hitl_queue.execution_id`` from
+``message["execution_id"]``. These tests pin that contract — the key must be
+present (and named exactly ``execution_id``) so the column stops being NULL.
+"""
+
+from __future__ import annotations
+
+from shared.enums import QueueResultStatus
+from shared.models.result_models import QueueResult
+
+
+def _result(**overrides):
+    kwargs = {
+        "file": "doc.pdf",
+        "status": QueueResultStatus.SUCCESS,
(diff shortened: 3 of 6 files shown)


workers/shared/models/result_models.py
```diff
@@ -223,6 +223,11 @@ class QueueResult:
     file_content: str | None = None
     whisper_hash: str | None = None
     file_execution_id: str | None = None
+    # Workflow execution id — carried into the HITL queue message so the backend
+    # can persist hitl_queue.execution_id (it reads message["execution_id"]).
+    # Without it the column is left NULL and a review item can't be traced back
+    # to its execution without joining through workflow_file_execution.
+    execution_id: str | None = None
```
muhammad-ali-e, line 237: **Type design (High value):** `execution_id: str | None = None` models the field as optional with a `None` default — which re-permits the exact NULL state this PR exists to eliminate. Its sibling execution-scoped id `workflow_id` (line 222) is required (`str`, no default) and validated in `__post_init__`; `execution_id` is always available at the only call site (`destination_connector.py:1836` reads `self.execution_id` unconditionally), so it has the same lifecycle.

Suggested: make it required and move it next to `workflow_id`, then add to the existing `__post_init__` validation block:
```python
if not self.execution_id:
    raise ValueError("QueueResult requires a valid execution_id")
```
This closes the NULL hole at construction instead of leaving it to chance. (The `test_execution_id_defaults_to_none_but_key_present` test currently encodes the optional behavior as acceptable and would need updating — which is correct.)

If backward-compat with unknown external callers is mandatory, at minimum add a `logger.warning` in `__post_init__` when `execution_id` is falsy, so a NULL write is logged rather than silent.

workers/shared/models/result_models.py
```diff
@@ -251,6 +256,7 @@ def to_dict(self) -> Any:
             "workflow_id": self.workflow_id,
             "file_content": self.file_content,
             "file_execution_id": self.file_execution_id,
+            "execution_id": self.execution_id,
```
muhammad-ali-e, line 276: **Maintainability (Medium):** `to_dict()` is a hand-maintained serializer — every field that the consumer reads must be manually mirrored here. A field added to the dataclass but forgotten here fails silently as a downstream NULL, which is precisely this ticket's failure mode. Consider replacing the hand-written body with `dataclasses.asdict(self)` plus the one transform it needs (`status` → `status.value`), so adding a field automatically propagates to the wire format. Not required for this PR, but it removes the recurring two-place-edit hazard.

workers/shared/workflow/destination_connector.py
```diff
@@ -1833,6 +1833,7 @@ def _push_data_to_queue(
                 workflow_id=str(self.workflow_id),
                 whisper_hash=whisper_hash,
                 file_execution_id=file_execution_id,
+                execution_id=self.execution_id,
```
muhammad-ali-e, line 1836: **Test gap (criticality 7 — highest-value missing test):** This line is the entire point of the PR, and it has no test. The two new tests only exercise `QueueResult(...).to_dict()` with `execution_id` passed in by hand, so they still pass if this kwarg is dropped (→ defaults to `None` → `hitl_queue.execution_id` back to NULL) or fat-fingered to `self.file_execution_id` (the adjacent line). Suggest a test that exercises `_push_data_to_queue` (line 1759) with the I/O helpers mocked and asserts the dict handed to the enqueue boundary carries the connector's `execution_id` — using a value distinct from `file_execution_id` so an adjacent-field swap is detectable.

workers/shared/models/result_models.py
```diff
@@ -223,6 +223,11 @@ class QueueResult:
     file_content: str | None = None
     whisper_hash: str | None = None
     file_execution_id: str | None = None
+    # Workflow execution id — carried into the HITL queue message so the backend
```
muhammad-ali-e: **Comment accuracy (comment-rot risk):** This comment asserts backend behavior — that it `reads message["execution_id"]` and persists `hitl_queue.execution_id`. That consumer is `pluggable_apps.manual_review_v2`, which is not in this repo (only referenced via guarded imports), so the claim can't be verified or kept honest here. It also contradicts the test file's comment, which describes the access as `.get("execution_id")` (no-raise) rather than subscript. Suggest scoping the comment to what this repo controls and naming where the other half lives, e.g.: `# Carried into the HITL queue message so the manual-review consumer (pluggable_apps.manual_review_v2, separate codebase) can populate hitl_queue.execution_id. The key name must stay in sync with that consumer.` Drop the assertion about which accessor the backend uses; the durable local fact is that `to_dict()` always emits the key (None when unset).

workers/tests/test_queue_result_execution_id.py
```diff
@@ -0,0 +1,47 @@
+"""UN-3655: ``QueueResult`` carries ``execution_id`` into the HITL queue message.
+
+The ETL manual-review push serialises a ``QueueResult`` to the queue message; the
+backend manual-review enqueue persists ``hitl_queue.execution_id`` from
+``message["execution_id"]``. These tests pin that contract — the key must be
```
muhammad-ali-e: **Docstring scope:** "These tests pin that contract" overstates what's covered — the tests only exercise the producer's `QueueResult.to_dict()` serialization, not the backend persistence half (consumer + `hitl_queue.execution_id` column live out-of-tree). Suggest scoping to the verifiable half: "pin the producer half of that contract — that `to_dict()` emits a key named exactly `execution_id`." Leave the consumer/column behavior as context rather than something the tests guarantee.

muhammad-ali-e (commented):

**Verdict:** The change is correct, additive, and safe at the wire level — `to_dict()` always emits the `execution_id` key, the call site wiring matches the sibling `file_execution_id`, and `self.execution_id` is structurally non-None on the live HITL enqueue path. Code-reviewer and code-simplifier found nothing to change.

The inline comments below are improvement opportunities, not blockers. Highest value: (1) the integration point at `destination_connector.py:1836` has no test, and (2) the field is modeled as optional with a `None` default, which re-permits the exact NULL state this PR exists to eliminate.

**Out-of-diff note (cannot inline):** `backend/workflow_manager/endpoint_v2/queue_utils.py:231-267` has a parallel `QueueResult` dataclass that still lacks `execution_id`. If any HITL enqueue still flows through that backend path (rather than the workers path), those rows keep getting NULL `execution_id` — worth confirming that path is dead, else mirror the fix there.

muhammad-ali-e:
Thanks for the review — addressed in the latest commit:

**Highest-value gap (test for the integration line)** — added `test_push_data_to_queue_wires_connector_execution_id`: it drives `_push_data_to_queue` with the I/O helpers mocked and asserts the dict at the enqueue boundary carries the connector's `execution_id`, with `execution_id != file_execution_id` so a dropped kwarg or an adjacent-field swap is caught. The `to_dict()` tests couldn't see that line.

**Type design (None default re-permits NULL)** — kept the field optional but `__post_init__` now logs a **WARNING** when `execution_id` is missing, so a latent NULL write is visible rather than silent. I went with the warning (your "at minimum" option) rather than a required-field raise on purpose: the connector's `execution_id` is genuinely nullable on some paths (`destination_connector.py:228` inits `None`; `:1003-1007` explicitly handles the falsy case), so a hard raise would convert a cosmetic NULL column into a broken HITL enqueue — strictly worse. Added a positive + negative caplog test for the warning.

**Comment accuracy / docstring scope** — rescoped both to what this repo controls (the producer half): named the out-of-tree consumer (`pluggable_apps.manual_review_v2`), dropped the backend-accessor (`message[...]` vs `.get`) assertion, and the durable local fact (`to_dict()` always emits the key, None when unset).

**Out-of-diff: parallel backend `QueueResult`** — good catch. Rather than trace liveness, I mirrored the additive fix into `endpoint_v2/queue_utils.py` + `destination._create_queue_result` (the backend connector has `self.execution_id`), so the column is populated whichever destination path enqueues. Backend parity test added.

**Deferred: `to_dict()` → `dataclasses.asdict()`** — left as-is for this PR; switching the serialization of every field has its own regression surface (field set/ordering on the wire) and is orthogonal to the NULL fix. Happy to do it as a follow-up if you'd like.

All workers (4) + backend (2) tests green; pre-commit clean.

PR #2126, opened 2026-06-30
UN-3657 [FIX] Execution serializer — terminal-failure run reports unaccounted files as failed, not in-progress

muhammad-ali-e:
## What

A terminal **ERROR/STOPPED** execution no longer shows files as still "in progress" in the UI.

## Why

Observed on PG exec `4b180f05` (ERROR, `total_files=2`) which displayed **"2 in progress"** for a finished run. `ExecutionSerializer` derives `successful_files`/`failed_files` as `SerializerMethodField`s that **count `workflow_file_execution` rows** (`status=COMPLETED`/`ERROR`). When orchestration fails **before** those rows are created (0 rows), both read 0, and the UI derives in-progress as `total - successful - failed = total` → phantom in-progress for a finished run. The persisted `workflow_execution.failed_files` column (set by UN-3652) is ignored by this serializer.

This affects **both transports** — a Celery orchestration that fails pre-row-creation shows the same artifact. It is not a PG regression; it's a latent display bug.

## How

`get_failed_files` now returns `total_files - successful_files` for a terminal **failure** run (`ExecutionStatus.is_failure` → ERROR/STOPPED): every file that didn't succeed in a finished failure run is a failure, including ones that never got a row or were left PENDING/EXECUTING by the abort.

Scoped to `is_failure`, so:
- **COMPLETED** runs → unchanged (row count) — success path untouched.
- **Live** runs (PENDING/EXECUTING) → unchanged — genuinely-pending files are not prematurely marked failed.
- No-op whenever the rows already account for every file.

`max(0, …)` guards a stale `successful > total` count. OSS read-path only; transport-agnostic.

> Note: Fix UN-3654 already **prevents** the specific failure that produced this display; this is defense-in-depth for *any* early-orchestration failure (bad source, permissions, validation).

## Tests

`test_execution_serializer.py` (6, ORM mocked — no test DB, mirroring the repo's existing mocked-ORM serializer tests): error/stopped reconcile (incl. the `4b180f05` 0-row case + partial rows), COMPLETED + EXECUTING unchanged, negative-clamp guard.

## Validation

Deterministic read-path logic, fully unit-covered across every branch. E2E validates on deploy: a failed execution will read "**N failed**, 0 in progress" instead of phantom in-progress.

### DIFF
--- a/backend/workflow_manager/execution/serializer/execution.py
+++ b/backend/workflow_manager/execution/serializer/execution.py
@@ -1,7 +1,11 @@
+import logging
+
 from rest_framework import serializers
 
 from workflow_manager.workflow_v2.enums import ExecutionStatus
 from workflow_manager.workflow_v2.models import WorkflowExecution
+
+logger = logging.getLogger(__name__)
 
 
 # TODO: Optimize with select_related / prefetch_related to reduce DB queries
@@ -25,30 +29,59 @@ class ExecutionSerializer(serializers.ModelSerializer):
         """Fetch the pipeline or API deployment name"""
         return obj.pipeline_name
 
+    def _status_count(self, obj: WorkflowExecution, status_value: str) -> int:
+        """COUNT of this execution's file rows in ``status_value``, memoised per
+        (execution, status) so the successful/failed fields don't issue
+        duplicate COUNTs for the same status within one serialization pass.
+        """
+        cache = self.__dict__.setdefault("_status_count_cache", {})
+        key = (str(obj.id), status_value)
+        if key not in cache:
+            cache[key] = obj.file_executions.filter(status=status_value).count()
+        return cache[key]
+
     def get_successful_files(self, obj: WorkflowExecution) -> int:
         """Return the count of successfully executed files"""
-        return obj.file_executions.filter(status=ExecutionStatus.COMPLETED.value).count()
+        return self._status_count(obj, ExecutionStatus.COMPLETED.value)
 
     def get_failed_files(self, obj: WorkflowExecution) -> int:
-        """Return the count of failed executed files.
+        """Return the count of failed files.
 
-        For a terminal *failure* run (ERROR/STOPPED), every file that did not
-        succeed is a failure — including files that never got a file-execution
-        row because orchestration aborted before creating them, or were left
-        PENDING/EXECUTING by the abort. The UI derives "in progress" as
-        ``total - successful - failed``, so without this a finished failure run
-        shows phantom in-progress files (e.g. an early barrier-dispatch failure
-        with 0 rows reads as "N in progress" instead of "N failed").
+        For a terminal *failure* run (ERROR/STOPPED), files that never reached a
+        terminal row — orchestration aborted before creating them, or they were
+        left PENDING/EXECUTING — are failures, not "in progress". The UI derives
+        in-progress as ``total - successful - failed`` (frontend
+        ``DetailedLogs.jsx``), so without this a finished failure run shows
+        phantom in-progress files. A no-op whenever every file already has a
+        terminal COMPLETED/ERROR row.
 
-        Scoped to ``is_failure`` (ERROR/STOPPED) so COMPLETED runs and live
-        runs (PENDING/EXECUTING) keep the exact row-count behaviour — a no-op
-        whenever the rows already account for every file, so the success path
-        and real-time progress are untouched on every transport.
+        Never under-reports below the real ERROR-row count: if the terminal rows
+        already exceed ``total_files`` (counter drift / an impossible count) the
+        derived value would be too low, so fall back to the ERROR rows and log
+        the drift instead of silently clamping it away.
         """
-        if ExecutionStatus.is_failure(obj.status):
-            total = obj.total_files or 0
-            return max(0, total - self.get_successful_files(obj))
-        return obj.file_executions.filter(status=ExecutionStatus.ERROR.value).count()
+        error_rows = self._status_count(obj, ExecutionStatus.ERROR.value)
+        # is_failure swallows ValueError -> False for an unrecognised status, so a
+        # malformed/legacy status safely takes the plain row-count path (no raise,
+        # so it can't 500 the executions list — but also no reconcile for it).
+        if not ExecutionStatus.is_failure(obj.status):
+            return error_rows
+        successful = self._status_count(obj, ExecutionStatus.COMPLETED.value)
+        total = obj.total_files or 0
+        if successful + error_rows > total:
+            logger.warning(
+                "Execution %s terminal file counts exceed total_files "
+                "(status=%s total=%s successful=%s error_rows=%s); reporting "
+                "failed=%s",
+                obj.id,
+                obj.status,
+                total,
+                successful,
+                error_rows,
+                error_rows,
+            )
+            return error_rows
+        return total - successful
 
     def get_aggregated_total_pages_processed(self, obj: WorkflowExecution) -> int | None:
         """Return the total pages processed across all file executions."""
--- a/backend/workflow_manager/execution/serializer/execution.py
+++ b/backend/workflow_manager/execution/serializer/execution.py
@@ -30,7 +30,24 @@ class ExecutionSerializer(serializers.ModelSerializer):
         return obj.file_executions.filter(status=ExecutionStatus.COMPLETED.value).count()
 
     def get_failed_files(self, obj: WorkflowExecution) -> int:
-        """Return the count of failed executed files"""
+        """Return the count of failed executed files.
+
+        For a terminal *failure* run (ERROR/STOPPED), every file that did not
+        succeed is a failure — including files that never got a file-execution
+        row because orchestration aborted before creating them, or were left
+        PENDING/EXECUTING by the abort. The UI derives "in progress" as
+        ``total - successful - failed``, so without this a finished failure run
+        shows phantom in-progress files (e.g. an early barrier-dispatch failure
+        with 0 rows reads as "N in progress" instead of "N failed").
+
+        Scoped to ``is_failure`` (ERROR/STOPPED) so COMPLETED runs and live
+        runs (PENDING/EXECUTING) keep the exact row-count behaviour — a no-op
+        whenever the rows already account for every file, so the success path
+        and real-time progress are untouched on every transport.
+        """
+        if ExecutionStatus.is_failure(obj.status):
+            total = obj.total_files or 0
+            return max(0, total - self.get_successful_files(obj))
         return obj.file_executions.filter(status=ExecutionStatus.ERROR.value).count()
 
     def get_aggregated_total_pages_processed(self, obj: WorkflowExecution) -> int | None:
--- a/backend/workflow_manager/execution/tests/test_execution_serializer.py
+++ b/backend/workflow_manager/execution/tests/test_execution_serializer.py
@@ -1,0 +1,88 @@
+"""ExecutionSerializer file-count reconciliation (UN-3657).
+
+A terminal *failure* run (ERROR/STOPPED) must not report files as still
+"in progress": the UI derives in-progress as ``total - successful - failed``,
+so every non-successful file in a finished failure run has to land in
+``failed`` — including files that never got a ``file_execution`` row because
+orchestration aborted before creating them. COMPLETED and live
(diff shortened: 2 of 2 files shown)


backend/workflow_manager/execution/serializer/execution.py
```diff
@@ -30,7 +30,24 @@ def get_successful_files(self, obj: WorkflowExecution) -> int:
...
+        ``total - successful - failed``, so without this a finished failure run
+        shows phantom in-progress files (e.g. an early barrier-dispatch failure
+        with 0 rows reads as "N in progress" instead of "N failed").
+
+        Scoped to ``is_failure`` (ERROR/STOPPED) so COMPLETED runs and live
+        runs (PENDING/EXECUTING) keep the exact row-count behaviour — a no-op
+        whenever the rows already account for every file, so the success path
+        and real-time progress are untouched on every transport.
+        """
+        if ExecutionStatus.is_failure(obj.status):
```
muhammad-ali-e: **P1 — failure branch trusts `total_files` and discards the actual ERROR-row count, so a stale/low `total_files` silently *under*-reports failures (the inverse of the bug being fixed).**

The new branch returns `total_files - successful` and never looks at real `ERROR` rows. If `total_files` lags the true row count (counter drift — exactly what UN-3652/UN-3657 are about — or a partial-write race during abort), genuine failures vanish and reappear as phantom in-progress via the UI's `total - successful - failed`.

Example: `total_files=2` but 5 rows exist (3 COMPLETED, 2 ERROR) → `max(0, 2-3)=0` failed; the old code returned 2. Two real failures disappear, silently.

Secondary: `ExecutionStatus.is_failure` swallows `ValueError` and returns `False` for any unrecognized status (data_models.py:576-581), so a malformed/legacy status string silently takes the `else` row-count path — reintroducing phantom-in-progress for exactly the corrupt data most worth flagging.

Suggested hardening — take the larger of the derived value and the real ERROR count, and don't let the divergence go unrecorded:
```python
error_rows = obj.file_executions.filter(status=ExecutionStatus.ERROR.value).count()
derived = total - self.get_successful_files(obj)
if derived != error_rows:
    logger.warning("exec %s failed-file reconcile mismatch: total=%s successful=%s error_rows=%s", obj.id, total, ..., error_rows)
return max(0, derived, error_rows)
```
(There is no `logger` in this module today — its siblings `views/execution.py` and `workflow_v2/serializers.py` both define one.)

**Broader fix worth considering:** the partition invariant `successful + failed + in_progress == total` is currently split across this serializer (two terms) and the frontend (`DetailedLogs.jsx` derives the third). Emitting an explicit `in_progress_files = SerializerMethodField` (= `max(0, total - successful - failed)`) and having the UI render it would let the server own all three buckets, making this whole class of phantom-count bug impossible regardless of how `failed` is computed.

backend/workflow_manager/execution/serializer/execution.py
```diff
@@ -30,7 +30,24 @@ def get_successful_files(self, obj: WorkflowExecution) -> int:
...
+            total = obj.total_files or 0
+            return max(0, total - self.get_successful_files(obj))
```
muhammad-ali-e: **P2 — `max(0, …)` silently clamps the impossible `successful > total` state with no signal.** `successful > total` means more files completed than the run claims to have had — an impossible count that indicates real counter corruption (double-counted completions, `total_files` written too low at dispatch, concurrent updates). The clamp returns `0` and erases the most diagnostic inconsistency there is. `test_successful_over_total_is_clamped_not_negative` even enshrines this silent `0` as correct.

Keep the clamp as the safe return value, but `logger.warning(...)` the anomaly first (execution id + both counts) so the drift is debuggable months later instead of invisible. This is consistent with the project's error-surfacing conventions.

backend/workflow_manager/execution/serializer/execution.py
```diff
@@ -30,7 +30,24 @@ def get_successful_files(self, obj: WorkflowExecution) -> int:
...
+        For a terminal *failure* run (ERROR/STOPPED), every file that did not
+        succeed is a failure — including files that never got a file-execution
+        row because orchestration aborted before creating them, or were left
+        PENDING/EXECUTING by the abort. The UI derives "in progress" as
+        ``total - successful - failed``, so without this a finished failure run
+        shows phantom in-progress files (e.g. an early barrier-dispatch failure
+        with 0 rows reads as "N in progress" instead of "N failed").
+
+        Scoped to ``is_failure`` (ERROR/STOPPED) so COMPLETED runs and live
+        runs (PENDING/EXECUTING) keep the exact row-count behaviour — a no-op
```
muhammad-ali-e: **P3 — docstring precision / verbosity.**

1. "a no-op whenever the rows already account for every file" is imprecise: the new branch (`total - COMPLETED`) equals the old (`ERROR count`) only when `completed + errored == total`. A file with a row that is still PENDING/EXECUTING *does* account for the file in the plain reading, yet the new code now counts it as failed — so behavior *does* change there. Suggest: "a no-op whenever every file already has a terminal COMPLETED/ERROR row."
2. "...untouched on every transport" — this method reads only `obj.status`/`total_files`/row counts and has no transport (Celery vs PG) dependency; mentioning transport implies a conditional that doesn't exist. Drop it.
3. 14 lines of prose for a 3-line method, with the rationale repeated again in the test-module docstring. Consider trimming to ~6 lines, keeping the load-bearing "why" (UI derives in-progress as `total - successful - failed`). Ideally anchor that claim to the specific frontend component so it doesn't rot silently.

backend/workflow_manager/execution/tests/test_execution_serializer.py
```diff
@@ -0,0 +1,88 @@
...
+
+from __future__ import annotations
+
+import os
+from unittest.mock import MagicMock
+
+import django
+from django.apps import apps
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "backend.settings.test")
```
muhammad-ali-e: **P2 — `os.environ.setdefault("DJANGO_SETTINGS_MODULE", …)` is a no-op when the var is already exported, which is extremely common in dev shells (`backend.settings.dev`, or an enterprise `backend.settings.cloud`).** A developer with `DJANGO_SETTINGS_MODULE` already set gets *their* module instead of `backend.settings.test`, breaking collection (reproduced: `ModuleNotFoundError: No module named 'backend.settings.cloud'` on a checkout with that env). Force the assignment: `os.environ["DJANGO_SETTINGS_MODULE"] = "backend.settings.test"`.

backend/workflow_manager/execution/tests/test_execution_serializer.py
```diff
@@ -0,0 +1,88 @@
...
+if not apps.ready:
+    django.setup()
```
muhammad-ali-e: **P2 — `django.setup()` contradicts the module docstring's "no test database is required" (line 10) and diverges from this repo's DB-free test convention.** `django.setup()` imports the full `backend.settings.base` stack and runs every app's `AppConfig.ready()` — which (a) requires a complete prod-like env (from a clean checkout with only `sample.test.env` it aborts on missing `CELERY_BROKER_BASE_URL`/`INDEXING_FLAG_TTL`), and (b) actually opens a live DB connection during startup (`execution_log_utils.ready()`). So the bootstrap is neither env-free nor DB-free as claimed, even though the serializer methods under test are pure.

Sibling DB-free tests avoid this by stubbing the models in `sys.modules` before import (`usage_v2/tests/test_helper.py`, `pg_queue/tests/test_producer.py`). Recommend following that pattern — stub `workflow_manager.workflow_v2.models` (+ enums), drop `django.setup()`, and the docstring becomes true. (Relevant since there is no backend pytest job in CI, so the clean-environment fragility is the primary failure mode.)

backend/workflow_manager/execution/tests/test_execution_serializer.py
```diff
@@ -0,0 +1,88 @@
...
+        # total 3: 1 completed, 1 errored row, 1 never created -> 2 failed.
+        obj = _execution(ExecutionStatus.ERROR.value, 3, completed=1, errored=1)
+        assert self.serializer.get_failed_files(obj) == 2
+
+    def test_stopped_run_reconciles(self):
+        # STOPPED is a failure terminal state: 3 of 5 done -> 2 failed, 0 in-progress.
+        obj = _execution(ExecutionStatus.STOPPED.value, 5, completed=3)
+        assert self.serializer.get_failed_files(obj) == 2
+
+    def test_completed_run_uses_row_count_unchanged(self):
```
muhammad-ali-e, line 98: **P3 — test gaps that would lock the intended contract:**

1. **COMPLETED-not-reconciled asymmetry is untested.** This case only covers `completed + errored == total`. The deliberate decision *not* to reconcile COMPLETED is invisible: add `_execution(COMPLETED, total=3, completed=1, errored=1)` asserting `failed == 1` (the ERROR-row count), not `total - completed == 2`, so a future "fix" that reconciles COMPLETED fails here.
2. **Assert the real invariant, not just `failed`.** `test_stopped_run`/`test_error_run_partial` assert only `get_failed_files`; they don't verify the actual contract (`total - successful - failed == 0`, i.e. zero phantom in-progress). Add that derived assertion so the test fails if either count regresses.
3. **Unknown/garbage status** (`is_failure`'s swallowed `ValueError` → `else` path) is the untested branch most worth pinning: `_execution("GARBAGE", 5, completed=1, errored=1)` should return `1` and must never raise (a raise would 500 the executions list endpoint).

Minor: the `4b180f05` exec id in the line-55 comment is an opaque one-off production UUID — the rest of the comment already describes the scenario; consider dropping the id.

muhammad-ali-e (commented):

Multi-agent review of the `get_failed_files` reconciliation change. The production fix is **logically correct and well-scoped** for the happy path it targets — finished ERROR/STOPPED runs no longer show phantom in-progress files. Findings below are about robustness of the new derivation, observability of the defensive guards, the test harness, and docstring precision. Inline comments follow; a prioritized summary is in the PR thread.

_Note: the existing greptile P2 on line 50 (double `COUNT` query) is a separate perf point and is not repeated here._

muhammad-ali-e:
Thanks — this was a strong review; addressed in the latest commit.

**P1 (under-reporting on drift) — fixed.** `get_failed_files` no longer blindly trusts `total_files`. It now also reads the real ERROR-row count and, when the terminal rows already exceed `total_files` (counter drift / impossible count), returns the ERROR rows instead of `total - successful` — so genuine failures can't silently vanish. Your example (`total=2`, 3 COMPLETED + 2 ERROR) now correctly returns 2 (and logs the drift). The common pre-row-creation case still derives `total - successful`.

**P2 (double COUNT) — fixed.** Added a per-(execution, status) memo so `successful_files`/`failed_files` don't issue duplicate COUNTs for the same status — which also tightens the pre-existing N+1 the TODO flags (failure reads now do 2 counts, not 3).

**P2 (silent clamp) — fixed.** `successful > total` (and any terminal-rows-exceed-total) now logs a WARNING with the execution id + both counts before returning the safe value, instead of erasing the most diagnostic inconsistency silently. (Added a module `logger`.)

**is_failure swallows ValueError** — documented inline: an unrecognised status takes the plain row-count path and never raises (so it can't 500 the list endpoint). Pinned by a garbage-status test.

**Test harness (both P2s) — fixed.** Forced `DJANGO_SETTINGS_MODULE` (the `setdefault` was a no-op when a dev shell already exports `.dev`/`.cloud`), and switched to the sibling `sys.modules` model-stub pattern — dropped `django.setup()` entirely. The test is now genuinely DB-free / app-registry-free (runs in 0.26s, no `AppConfig.ready()` DB connection), so the docstring is true.

**P3 docstring** — fixed: corrected the 'no-op' claim (only when every file has a terminal COMPLETED/ERROR row), dropped the transport mention and the opaque exec UUID, anchored in-progress to `DetailedLogs.jsx`.

**P3 test gaps** — added: drift-no-under-report, successful>total clamp+warn, the COMPLETED-not-reconciled asymmetry, unknown-status-never-raises, and the real `total - successful - failed == 0` invariant on failure runs (not just `failed`).

**Deferred — explicit `in_progress_files` server field.** Agree it's the durable fix (server owns all three buckets, makes the whole class impossible). Left out of this PR because it needs a matching frontend change in `DetailedLogs.jsx`; happy to do it as a focused follow-up.

All 9 tests green; pre-commit clean.

PR #2129, opened 2026-06-30
UN-3659 [FIX] PG result backend — reconnect-retry on stale store_result connection

muhammad-ali-e:
## What
`PgResultBackend.store_result()` now survives a connection that PgBouncer reaped while the executor sat idle — a one-shot reconnect-retry. Sibling of UN-3654.

## Why
A PG execution **hung permanently** (exec `b11ba2f3`, ali dev) after the worker was idle ~56 min (flag off → enabled + run, past PgBouncer `server_idle_timeout`). The logs showed every file **extracted text and answered 4/4 prompts successfully**, then:

```
PgResultBackend: rollback failed; treating connection as dead
psycopg2.OperationalError: server closed the connection unexpectedly
FAILED to store request-reply result (...) — acking anyway; caller will time out
```

The executor consumer caches **one** `PgResultBackend` for its lifetime, so its write connection is long-lived and idle-reapable. `_cursor` discards a dead connection so the *next* call reconnects, but `store_result()`'s *current* call failed with no retry → the result was dropped → the blocking file-processing callers waited forever, got SIGKILLed, and the execution was orphaned in `EXECUTING`. UN-3654 added the retry to dispatch `send()` and the consumer's claim connection but **not** to this result-write path.

## How
`store_result()` retries once on `OperationalError`/`InterfaceError`. **Unconditional** (no reused-vs-fresh guard, unlike `send()`) because the write is `INSERT … ON CONFLICT (task_id) DO NOTHING` — re-running can't duplicate or clobber a recorded result. `_cursor` already drops the dead conn, so the retry reconnects and writes the result, unblocking the caller.

## Scope (audited the other PG paths)
- `store_result` — **fixed** (long-lived cached conn).
- `get_result`/`wait_for_result` — **safe**, left as-is: its only caller (`executor_rpc.wait_for_result`) makes a *fresh* `PgResultBackend` per wait and polls every ~0.2–2s, so no idle window. (Documented inline.)
- Consumer `read`/`delete` + periodic leader/reaper/scheduler loops — self-heal next cycle.
- **Related follow-up (separate ticket):** the `PgBarrier` decrement / `claim_batch` (cached thread-local conn) has the same idle-reap risk and was deliberately left non-retried by UN-3651 (non-idempotent); UN-3654's reused-guard would resolve it. Not the trigger here.

## Tests
- **Dev-test:** real `pg_terminate_backend` A/B — bare INSERT fails on the reaped conn with the exact `server closed the connection` that stranded `b11ba2f3`; `store_result()` self-heals and the row is verified in the DB.
- **6 unit tests:** retry+succeed, both `Operational`/`Interface` error types, backoff fires once, retry-also-fails raises after exactly one reconnect, non-connection error not retried. Full file: 13 passed.

## Op note
`b11ba2f3` won't self-recover (no execution-level reaper for stuck `EXECUTING`) — mark it FAILED on the primary or re-run.

### DIFF
--- a/workers/queue_backend/pg_queue/result_backend.py
+++ b/workers/queue_backend/pg_queue/result_backend.py
@@ -37,7 +37,7 @@ import contextlib
 import json
 import logging
 import time
-from collections.abc import Iterator
+from collections.abc import Callable, Iterator
 from typing import TYPE_CHECKING, Any, Final, Self
 
 import psycopg2
@@ -52,6 +52,17 @@ if TYPE_CHECKING:
     from psycopg2.extensions import connection as PgConnection
 
 logger = logging.getLogger(__name__)
+
+# psycopg2 errors that mean the connection itself is dead (dropped socket /
+# PgBouncer recycle / server termination). Shared by ``_cursor`` (decides whether
+# to discard the cached handle) and ``_store_with_reconnect`` (decides whether the
+# one-shot retry is eligible) so the two can't drift — narrowing one without the
+# other would silently break the retry's "was this a connection death?" test.
+# Mirrors ``PgQueueClient._CONN_DEAD_ERRORS`` (UN-3654), the sibling this models on.
+_CONN_DEAD_ERRORS: Final[tuple[type[Exception], ...]] = (
+    psycopg2.OperationalError,
+    psycopg2.InterfaceError,
+)
 
 # How long a stored result lives before the reaper's retention sweep may delete
 # it. Defaults to the executor caller-timeout default so a result always
@@ -119,9 +130,7 @@ class PgResultBackend:
                 yield cur
             conn.commit()
         except Exception as exc:
-            conn_dead = isinstance(
-                exc, (psycopg2.OperationalError, psycopg2.InterfaceError)
-            )
+            conn_dead = isinstance(exc, _CONN_DEAD_ERRORS)
             try:
                 conn.rollback()
             except Exception:
@@ -139,7 +148,7 @@ class PgResultBackend:
                 self._conn = None
             raise
 
-    def _store_with_reconnect(self, operation: Any) -> None:
+    def _store_with_reconnect(self, operation: Callable[[Any], None]) -> None:
         """Run an idempotent store ``operation(cur)`` in a committed cursor,
         retrying ONCE if the cached connection was reaped while idle.
 
@@ -150,7 +159,14 @@ class PgResultBackend:
         heals the *next* call, so without a retry the *current* ``store_result``
         fails, the result is silently dropped, and the blocking caller is
         stranded forever (the exec-b11ba2f3 hang). On a dead-connection error
-        ``_cursor`` has already dropped the handle, so the retry reconnects.
+        ``_cursor`` (when it OWNS the connection) has already dropped the handle,
+        so the retry reconnects.
+
+        Only **owned** connections are retried: ``_cursor`` clears a dead handle
+        only when ``_owns_conn`` (an injected connection is the caller's), so for
+        an injected connection the retry would re-acquire the same dead handle and
+        buy nothing — re-raise immediately. Production is always owned (the
+        executor consumer constructs ``PgResultBackend()``).
 
         Safe to retry unconditionally (no reused-vs-fresh guard needed, unlike
         the non-idempotent ``PgQueueClient.send``): the write is
@@ -162,15 +178,17 @@ class PgResultBackend:
                 with self._cursor() as cur:
                     operation(cur)
                 return
-            except (psycopg2.OperationalError, psycopg2.InterfaceError) as exc:
-                if attempt >= _STORE_RETRY_ATTEMPTS:
+            except _CONN_DEAD_ERRORS:
+                # Last attempt, or an injected (non-owned) conn _cursor won't
+                # have dropped — retrying can't reconnect, so re-raise.
+                if attempt >= _STORE_RETRY_ATTEMPTS or not self._owns_conn:
                     raise
+                # Describe what was observed, not an inferred cause: this catch
+                # also covers deadlock / statement-timeout / admin-shutdown, not
+                # only a stale idle reap. exc_info carries the type + message.
                 logger.warning(
                     "PgResultBackend: result write failed with a connection-level "
-                    "error (%s: %s); dropping the cached connection and retrying "
-                    "once (stale idle reap or DB unavailable)",
-                    type(exc).__name__,
-                    exc,
+                    "error; dropping the cached connection and retrying once",
                     exc_info=True,
                 )
                 time.sleep(_STORE_RETRY_BACKOFF_SECONDS)
@@ -208,11 +226,14 @@ class PgResultBackend:
         ``result`` is the decoded JSONB dict (psycopg2 parses ``jsonb`` to a
         Python ``dict``); ``None`` means the task has not finished yet.
 
-        No reconnect-retry here (unlike :meth:`store_result`): the only caller,
-        ``executor_rpc.wait_for_result``, makes a FRESH ``PgResultBackend`` per
-        wait and then polls every ~0.2–2s, so this connection is new and kept
-        warm — it has no idle window to be reaped. A failure on a fresh
-        connection is a genuine error, not a stale-reap.
+        No reconnect-retry here (unlike :meth:`store_result`). The invariant it
+        relies on: the sole entry point into a wait (``executor_rpc.wait_for_result``
+        → this class's :meth:`wait_for_result` → ``get_result``) constructs a
+        FRESH ``PgResultBackend`` per wait and polls it every ~0.2–2s, so this
+        connection is new and kept warm — no idle window to be reaped, and a
+        failure on a fresh connection is a genuine error, not a stale reap. If a
+        long-lived backend ever gains a one-shot ``get_result`` lookup, revisit
+        this.
         """
         with self._cursor() as cur:
             cur.execute(_get_sql(), (str(task_id),))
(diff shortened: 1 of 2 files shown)


workers/queue_backend/pg_queue/result_backend.py
```diff
@@ -129,6 +139,42 @@ def _cursor(self) -> Iterator[Any]:
...
+        the non-idempotent ``PgQueueClient.send``): the write is
+        ``INSERT … ON CONFLICT (task_id) DO NOTHING``, so re-running after an
+        ambiguous failure can neither duplicate nor clobber a recorded result.
+        """
+        for attempt in range(1, _STORE_RETRY_ATTEMPTS + 1):
+            try:
+                with self._cursor() as cur:
+                    operation(cur)
+                return
+            except (psycopg2.OperationalError, psycopg2.InterfaceError) as exc:
```
muhammad-ali-e: **P2 — drift hazard + over-broad catch.** The connection-dead tuple `(psycopg2.OperationalError, psycopg2.InterfaceError)` is now inlined in **two** coupled sites: here and in `_cursor` (`isinstance(exc, (...))`, line 122-124). They must stay in lockstep — narrowing one without the other silently breaks the retry's "was this a connection death?" test. The sibling `client.py` (UN-3654) deliberately hoisted this into a named `_CONN_DEAD_ERRORS: Final[tuple[...]] = (...)` constant *precisely to prevent this*, with a comment to that effect. This PR re-introduces the duplication the sibling removed. Suggest mirroring it:
```python
_CONN_DEAD_ERRORS: Final[tuple[type[Exception], ...]] = (
    psycopg2.OperationalError, psycopg2.InterfaceError,
)
```
then `except _CONN_DEAD_ERRORS as exc:` here and `isinstance(exc, _CONN_DEAD_ERRORS)` in `_cursor`.

Secondary (P3): `OperationalError` is also the base class for deadlock/serialization (class 40), statement-timeout/cancel, and disk-full/too-many-connections (class 53). Those will trip this retry + the 0.5s sleep and tear down a healthy cached connection, while being logged as "stale idle reap or DB unavailable" — a misleading label for a deadlock. Idempotency keeps it correctness-safe, but consider either softening the warning wording to describe what was observed rather than an inferred cause, or gating the retry on `exc.pgcode` SQLSTATE class `08` (plus `InterfaceError`, which has no pgcode).

workers/queue_backend/pg_queue/result_backend.py
```diff
@@ -129,6 +139,42 @@ def _cursor(self) -> Iterator[Any]:
...
+        fails, the result is silently dropped, and the blocking caller is
+        stranded forever (the exec-b11ba2f3 hang). On a dead-connection error
+        ``_cursor`` has already dropped the handle, so the retry reconnects.
+
+        Safe to retry unconditionally (no reused-vs-fresh guard needed, unlike
+        the non-idempotent ``PgQueueClient.send``): the write is
+        ``INSERT … ON CONFLICT (task_id) DO NOTHING``, so re-running after an
+        ambiguous failure can neither duplicate nor clobber a recorded result.
+        """
+        for attempt in range(1, _STORE_RETRY_ATTEMPTS + 1):
```
muhammad-ali-e, line 176: **P3 — retry is a silent no-op for injected (non-owned) connections.** `_cursor` only discards the dead handle when `self._owns_conn` (line 136). For an injected connection (`_owns_conn=False`), `self._conn` is *not* cleared on a connection-level error, so attempt 2 re-acquires the **same** dead handle via `self.conn`, burns the 0.5s backoff, fails identically, and re-raises — the retry buys nothing. Production is safe (the executor consumer always constructs `PgResultBackend()` with `_owns_conn=True`), but the asymmetry is undocumented. Either short-circuit `if not self._owns_conn` to raise immediately without the spurious sleep, or add a one-line note that the retry only heals owned connections. (Relatedly, the docstring's "`_cursor` has already dropped the handle" at line 152-153 is only true for owned connections — worth a half-clause clarifying that.)

workers/queue_backend/pg_queue/result_backend.py
```diff
@@ -141,23 +187,32 @@ def store_result(
...
             )
+        )
 
     def get_result(self, task_id: str) -> dict[str, Any] | None:
         """Return ``{status, result, error}`` if the row exists, else ``None``.
 
         ``result`` is the decoded JSONB dict (psycopg2 parses ``jsonb`` to a
         Python ``dict``); ``None`` means the task has not finished yet.
+
+        No reconnect-retry here (unlike :meth:`store_result`): the only caller,
```
muhammad-ali-e: **P3 — comment-rot trap.** This says the retry is omitted because "the only caller, `executor_rpc.wait_for_result`, makes a FRESH `PgResultBackend` per wait." Two precisions: (1) `get_result`'s *direct* caller is `PgResultBackend.wait_for_result` (same class); `executor_rpc.wait_for_result` is the reachable entry point two hops up. (2) This is a load-bearing "only caller" safety assertion — if someone later adds a one-shot `get_result` lookup on a long-lived backend, the rationale silently goes stale. Suggest rewording to the invariant it actually depends on, e.g. "the sole entry point creates a fresh backend per wait and polls it every ~0.2–2s, so the connection is new and kept warm." The 0.2–2s figures and "fresh per wait" themselves check out.

workers/tests/test_pg_result_backend.py
```diff
@@ -101,3 +105,126 @@ def write_after_delay() -> None:
...
+        self._cursor = cursor
+
+    def __enter__(self):
+        return self._cursor
+
+    def __exit__(self, *_):
+        return False
+
+
+class TestStoreResultReconnectRetry:
```
muhammad-ali-e, line 125: **P2 — no CI test proves the reconnect actually heals against real PG.** Every test here drives failure through `cur.execute.side_effect` against a fully mocked conn/cursor/`create_pg_connection`, asserting orchestration (close/factory/execute/commit were *called*). If `_cursor`'s discard stopped nulling `self._conn`, or the real reconnect produced an unusable handle, these would still pass. The genuine heal is only covered by the manual `pg_terminate_backend` dev-test, which is not in CI. Consider one DB-gated test (no `pg_terminate_backend` needed — a client-side `close()` makes psycopg2 raise `InterfaceError` on next use):
```python
rb = PgResultBackend()          # owned, real factory
rb.conn                          # materialize
rb._conn.close()                 # next use -> InterfaceError
rb.store_result(k, result={"healed": True})   # must self-heal
# then assert the row landed via a fresh backend's get_result
```
This is the test that would actually fail if the reconnect were broken.

workers/tests/test_pg_result_backend.py
```diff
@@ -101,3 +105,126 @@ def write_after_delay() -> None:
...
+        conn.cursor.return_value = _CursorCtx(cur)
+        return conn, cur
+
+    @staticmethod
+    def _no_sleep(monkeypatch):
+        sleep = MagicMock()
+        monkeypatch.setattr("queue_backend.pg_queue.result_backend.time.sleep", sleep)
+        return sleep
+
+    def test_stale_conn_retries_and_writes_result(self, monkeypatch):
```
muhammad-ali-e: **P3 — two retry-path gaps worth closing.** (1) **Failed-status branch is never retried:** all six tests call `store_result(..., result={...})` (completed branch); the `result is None -> STATUS_FAILED` path is only covered by the no-retry real-PG round-trip. Since the PR's whole motivation is "a failed task's result silently dropped -> caller stranded," delivering the *failed* outcome through a stale connection deserves coverage — easily added by parametrizing this test over `{"result": {...}}` and `{"error": "boom"}`. (2) **Commit-time reap untested:** `_cursor` commits *after* the `yield`, and an idle reap typically surfaces when the buffered INSERT flushes at `conn.commit()`, not at `execute`. A variant with `conn.commit.side_effect = OperationalError(...)` on attempt 1 (fresh conn succeeds) would cover the most realistic failure point.

muhammad-ali-e (commented):

The fix is correct and well-reasoned — the idempotency argument for the unconditional retry holds, the one-shot bound is right, and propagation/logging are sound. No functional bugs found. The inline notes below are consistency, error-specificity, and test-coverage improvements; none are merge-blockers except optionally the first (drift parity with the sibling `PgQueueClient` that this PR explicitly models itself on). Two findings overlap existing greptile comments at lines 142 and 175 and were intentionally omitted.

muhammad-ali-e:
Thanks — all addressed in the latest commit.

**P2 — shared `_CONN_DEAD_ERRORS` (parity with the sibling).** Extracted `_CONN_DEAD_ERRORS: Final[tuple[...]] = (OperationalError, InterfaceError)` and used it in both `_cursor` and `_store_with_reconnect` — exactly the named constant `PgQueueClient` (UN-3654) introduced. The duplication this PR re-introduced is gone.

**P2 — typing.** `operation: Callable[[Any], None]`.

**P2 — over-broad catch / misleading label.** Kept the catch (idempotency makes it correctness-safe) but reworded the WARNING to describe the *observation* ("a connection-level error") rather than asserting "stale idle reap" — since the same classes cover deadlock/timeout/admin-shutdown. Dropped the duplicate inline `%s: %s`; `exc_info=True` alone carries type+message (also resolves greptile's double-log note at line 175).

**P3 — injected (non-owned) connection.** Good catch on the silent no-op. Now short-circuits: the retry only runs for **owned** connections (since `_cursor` only clears the handle when `_owns_conn`); an injected conn re-raises immediately with no spurious backoff. Documented + a `test_injected_connection_not_retried` test.

**P3 — get_result comment-rot.** Reworded to the actual invariant (the sole wait entry point builds a fresh backend per wait and polls it ~0.2–2s → connection stays warm), and fixed the caller precision.

**P2 — no real-PG heal test.** Added `test_real_reconnect_heals_closed_connection` (DB-gated): owned backend, `conn.close()` → `InterfaceError` on next use → `store_result` must self-heal → the row is read back on a separate connection. This is the one that fails if the reconnect were actually broken.

**P3 — test gaps.** Parametrized the retry over **completed/failed** (the failed result was the whole point — silently dropped → caller stranded), and added a **commit-time reap** test (`conn.commit.side_effect` on attempt 1, since an idle reap usually surfaces at flush, not execute).

17 tests green (incl. the DB-gated heal); pre-commit clean.

PR #2130, opened 2026-06-30
UN-3660 [FIX] PG barrier decrement — phase-split reconnect-retry on stale connection

muhammad-ali-e:
## What

Closes the **last** uncovered stale-connection-after-idle hang path in the PG-queue execution engine, after UN-3654 (dispatch `send`), UN-3651 (barrier enqueue) and UN-3659 (`store_result`).

The `PgBarrier` decrement —
```
UPDATE pg_barrier_state SET remaining = remaining - 1,
       results = results || jsonb_build_array(...) WHERE execution_id = ... RETURNING ...
```
runs on the worker's cached thread-local connection. After a worker sits idle > `server_idle_timeout` (PgBouncer) the connection is reaped server-side. The decrement's `try/except` caught only `psycopg2.DataError`, so an `OperationalError`/`InterfaceError` propagated, the batch's decrement was lost, `remaining` never reached 0, and the execution hung at the fan-in (all files `COMPLETED` but execution stuck `EXECUTING`) until the ~6h barrier expiry. Same class as the earlier hangs, one step later in the pipeline.

## Why not a plain retry (or `send`'s reused-guard)

The decrement is **non-idempotent**, and a double-apply is *harmful*: it can fire the callback early with incomplete results, or drive `remaining` past 0 and strand the barrier. Even `send()`'s reused-guard is not enough on its own here — a reused-connection failure at **commit** time is ambiguous (the server may have applied it).

## Fix — phase-split retry

Split the statement into its two phases and retry only the one where a re-apply is provably exactly-once:

- **EXECUTE** phase fails on a **reused** (cached) connection → idle-reap; the statement never committed → reconnect and re-apply **once**.
- **COMMIT** phase fails → ambiguous → **never** retry; propagate so the caller tears the barrier down (fail-fast — unchanged behaviour).
- Fresh-connection failure / `DataError` / non-connection error → propagate unchanged.

Safety rests on the connection being **non-autocommit** (verified in `create_pg_connection`): an execute-phase failure is never durable even if the server ran the `UPDATE`, because the open transaction is rolled back on disconnect. Durability happens only at `commit()`, the phase we never retry — so re-applying can never double-count.

Also adds the shared `_CONN_DEAD_ERRORS` constant (parity with `client.py` / `result_backend.py`, so the "is this a connection death?" test can't drift) and a `_recover_after_error` helper now reused by `_cursor`.

## Tests

- **6 unit** (`TestDecrementPhaseSplitRetry`): execute-retry-heals (both error types), **commit-fail-does-NOT-retry**, fresh-conn-no-retry, DataError-no-retry, one-shot bound.
- **Real-DB**: idle-reaped connection self-heals through the production entry and decrements **exactly once** (2→1, one result appended) — the no-double-decrement pin.
- **Real `pg_terminate_backend`**: a genuinely killed backend mid-decrement self-heals and fires the callback **exactly once**.
- Full `test_pg_barrier.py` (63) + sibling barrier suites (35) green; pre-commit clean.

Dev-tested end-to-end inside the live PG worker fleet against the real database (real backend termination mid-decrement → self-heal → callback once → barrier consumed).

After this, no stale-connection-after-idle path can hang an execution.

### DIFF
--- a/workers/queue_backend/pg_barrier.py
+++ b/workers/queue_backend/pg_barrier.py
@@ -94,6 +94,19 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+# psycopg2 errors that mean the connection itself is dead (dropped socket /
+# PgBouncer recycle / server termination), as opposed to a logical/data error on
+# a live connection. Shared by ``_cursor`` / ``_recover_after_error`` (discard the
+# stale thread-local handle), the idempotent pre-dispatch write retry, and the
+# decrement phase-split retry, so the "was this a connection death?" test can't
+# drift across them. Mirrors ``PgQueueClient._CONN_DEAD_ERRORS`` (UN-3654) and
+# ``PgResultBackend._CONN_DEAD_ERRORS`` (UN-3659), the siblings this models on.
+_CONN_DEAD_ERRORS: Final[tuple[type[Exception], ...]] = (
+    psycopg2.OperationalError,
+    psycopg2.InterfaceError,
+)
+
+
 # Thread-local owned connection (prefork → one per child; thread-local keeps it
 # correct under -P threads too, since a libpq connection is not concurrency-safe
 # across threads). Self-recovers a dropped socket / PgBouncer recycle — same
@@ -109,6 +122,27 @@ def _get_conn() -> PgConnection:
     return conn
 
 
+def _recover_after_error(conn: PgConnection, exc: BaseException) -> bool:
+    """Roll back ``conn``; if it is dead, discard the thread-local handle so the
+    next :func:`_get_conn` reconnects. Returns whether the connection was judged
+    dead (a failed rollback proves it dead regardless of the original error).
+
+    Factored out so ``_cursor`` and the decrement phase-split (which can't use
+    ``_cursor`` because it must distinguish execute-phase from commit-phase
+    failures) share one definition of "recover a connection after an error".
+    """
+    conn_dead = isinstance(exc, _CONN_DEAD_ERRORS)
+    try:
+        conn.rollback()
+    except Exception:
+        conn_dead = True
+    if conn_dead or conn.closed:
+        with contextlib.suppress(Exception):
+            conn.close()
+        _local.conn = None
+    return conn_dead
+
+
 @contextlib.contextmanager
 def _cursor() -> Iterator[Any]:
     """Yield a cursor; commit on success, roll back + recover on error."""
@@ -118,15 +152,7 @@ def _cursor() -> Iterator[Any]:
             yield cur
         conn.commit()
     except Exception as exc:
-        conn_dead = isinstance(exc, (psycopg2.OperationalError, psycopg2.InterfaceError))
-        try:
-            conn.rollback()
-        except Exception:
-            conn_dead = True
-        if conn_dead or conn.closed:
-            with contextlib.suppress(Exception):
-                conn.close()
-            _local.conn = None
+        _recover_after_error(conn, exc)
         raise
 
 
@@ -145,6 +171,11 @@ _BARRIER_WRITE_ATTEMPTS: Final = 2  # total attempts: 1 initial + 1 retry
 # avoids immediately re-hammering a struggling server on the rarer server-side
 # errors the broad psycopg2.OperationalError catch also covers.
 _BARRIER_RETRY_BACKOFF_SECONDS: Final = 0.5
+
+# One retry for the NON-idempotent barrier DECREMENT — but only on an
+# execute-phase failure on a reused connection (see :func:`_apply_decrement`),
+# never on commit (ambiguous). Same one-shot bound as the idempotent write.
+_BARRIER_DECREMENT_ATTEMPTS: Final = 2  # total attempts: 1 initial + 1 retry
 
 
 def _run_idempotent_pre_dispatch_write(
@@ -174,7 +205,7 @@ def _run_idempotent_pre_dispatch_write(
             with _cursor() as cur:
                 operation(cur)
             return
-        except (psycopg2.OperationalError, psycopg2.InterfaceError) as exc:
+        except _CONN_DEAD_ERRORS as exc:
             # _cursor already dropped the dead thread-local conn → the next
             # _get_conn() reconnects. Retry once; re-raise if it still fails
             # (a genuinely-down DB surfaces as ERROR, as before). Name the real
@@ -597,6 +628,99 @@ def _fire_barrier_callback(
     return str(callback_signature.apply_async().id)
 
 
+def _apply_decrement(execution_id: str, result_json: str) -> tuple[int, Any] | None:
+    """Apply the barrier decrement ``UPDATE`` and return its ``(remaining,
+    results)`` RETURNING row (``None`` if the barrier row is already gone).
+
+    The decrement is NON-idempotent — re-applying it double-counts (premature
+    callback fire with incomplete results, or a strand past 0) — so it is split
+    into its two phases and retried in ONLY the one phase where a re-apply is
+    provably safe:
+
+    - **EXECUTE phase** (``UPDATE … RETURNING`` + ``fetchone``) — the statement
+      runs but is NOT yet committed. A connection-level failure here on a
+      **reused** (cached) connection is the PgBouncer idle-reap: the statement
+      never reached the server, no commit was issued, and the open transaction is
+      rolled back on disconnect. So the decrement provably never landed →
+      reconnect and re-apply it exactly once. (Only a *cached* connection can be a
+      stale idle-reaped handle; a freshly-created one failing is a genuine DB
+      error, not a reap, so it is not retried.) **This safety relies on the
+      connection being non-autocommit** (``create_pg_connection`` opens it that
+      way): even if the server actually ran the ``UPDATE`` before the socket
+      dropped, it sits in an uncommitted transaction that is rolled back on
+      disconnect — durability happens ONLY at the commit below — so re-applying
+      can never double-count.
+    - **COMMIT phase** — ``conn.commit()``. A failure here is AMBIGUOUS: the
+      server may have applied the commit before the socket dropped. Re-applying
+      could double-count, so it is NEVER retried — it propagates, and the caller
+      (:func:`run_batch_with_barrier`) tears the barrier down so the execution
+      fails fast rather than risk a corrupted counter.
+
+    Any non-connection error (e.g. the NUL-byte ``psycopg2.DataError`` from the
+    jsonb cast) also propagates unchanged — only the idle-reap self-heals. This is
+    the decrement's counterpart to ``PgQueueClient.send``'s reused-guard
+    (UN-3654) and the idempotent pre-dispatch write's retry, but phase-split
+    because the decrement, unlike those, must never replay a committed write.
+    """
+    sql = (
+        f"UPDATE {qualified('pg_barrier_state')} "
+        "   SET remaining = remaining - 1, "
+        "       results = results || jsonb_build_array(%s::jsonb) "
+        " WHERE execution_id = %s "
+        "RETURNING remaining, results"
+    )
+    for attempt in range(1, _BARRIER_DECREMENT_ATTEMPTS + 1):
+        # Only a connection cached BEFORE this attempt can be a stale idle-reaped
+        # handle; a freshly-reconnected one failing is a genuine error, not a reap.
+        reused = getattr(_local, "conn", None) is not None and not _local.conn.closed
+        conn = _get_conn()
+        try:
+            with conn.cursor() as cur:
+                cur.execute(sql, (result_json, execution_id))
+                row = cur.fetchone()
+        except Exception as exc:
+            conn_dead = _recover_after_error(conn, exc)
+            # Retry ONLY a reused-conn death on the execute phase: it never
+            # committed (re-applying lands exactly once) and reconnecting can
+            # actually help. A fresh-conn death (DB genuinely down), a DataError,
+            # or the last attempt all propagate unchanged.
+            if conn_dead and reused and attempt < _BARRIER_DECREMENT_ATTEMPTS:
+                logger.warning(
+                    "[exec:%s] PgBarrier decrement: execute failed on a cached "
+                    "connection (%s) — reconnecting and re-applying once "
+                    "(attempt %d/%d); the decrement never committed, so it lands "
+                    "exactly once.",
+                    execution_id,
+                    type(exc).__name__,
+                    attempt,
+                    _BARRIER_DECREMENT_ATTEMPTS,
+                    exc_info=True,
+                )
+                time.sleep(_BARRIER_RETRY_BACKOFF_SECONDS)
+                continue
+            raise
+        try:
+            conn.commit()
+        except Exception as exc:
+            _recover_after_error(conn, exc)
+            # AMBIGUOUS — the server may have committed. Re-applying could
+            # double-count, so do NOT retry: propagate so the caller tears the
+            # barrier down (fail-fast) rather than corrupt the counter.
+            logger.warning(
+                "[exec:%s] PgBarrier decrement: commit failed (%s) — NOT retrying "
+                "(the server may already have applied it; a re-apply would "
+                "double-count). Propagating so the barrier is torn down.",
+                execution_id,
+                type(exc).__name__,
+                exc_info=True,
+            )
+            raise
+        return row
+    # Unreachable: the loop either returns the row or raises. The bare raise keeps
+    # the type checker honest about the no-fallthrough invariant.
+    raise AssertionError("unreachable: _apply_decrement loop exited without return")
+
+
 def _barrier_pg_decrement(
     result: Any,
     *,
@@ -616,12 +740,19 @@ def _barrier_pg_decrement(
       the header task's body — a PG-consumed task fires no ``.link``, so the
       decrement must run in-body.
 
-    Callers MUST run each decrement in its own committed transaction (the
-    decrement ``UPDATE`` runs in its own ``_cursor()`` txn) and MUST NOT retry
-    it: a replay re-runs the decrement and corrupts the count. This is enforced
-    loudly, not just in prose — entry raises if the shared connection is already
-    mid-transaction (see the guard below). The Celery wrapper additionally pins
-    ``max_retries=0``; an orphaned barrier is bounded by ``expires_at`` instead.
+    Callers MUST run each decrement in its own committed transaction and MUST NOT
+    replay a *committed* decrement: re-applying one that already landed corrupts
+    the count (fires the callback early with incomplete results, or skips past 0
+    and strands the barrier). This is enforced loudly, not just in prose — entry
+    raises if the shared connection is already mid-transaction (see the guard
+    below), and the Celery wrapper pins ``max_retries=0`` so a task-level replay
+    can't re-drive it; an orphaned barrier is bounded by ``expires_at`` instead.
+
+    The decrement DOES self-heal one narrow, provably-safe case via
+    :func:`_apply_decrement`: an execute-phase failure on a cached connection that
+    PgBouncer idle-reaped (the statement never reached the server, so nothing
+    committed) is reconnected and re-applied exactly once. A commit-phase failure
+    is ambiguous and is NEVER retried — see :func:`_apply_decrement`.
     """
     # Enforce the "own committed transaction" contract loudly. A caller that
     # invokes this inside an already-open transaction on the shared thread-local
@@ -656,16 +787,7 @@ def _barrier_pg_decrement(
     # jsonb_build_array(...) appends exactly one element regardless of the
     # result's shape (``||`` would concatenate if the result were itself a list).
     try:
-        with _cursor() as cur:
-            cur.execute(
-                f"UPDATE {qualified('pg_barrier_state')} "
-                "   SET remaining = remaining - 1, "
-                "       results = results || jsonb_build_array(%s::jsonb) "
-                " WHERE execution_id = %s "
-                "RETURNING remaining, results",
-                (result_json, execution_id),
-            )
-            row = cur.fetchone()
+        row = _apply_decrement(execution_id, result_json)
     except psycopg2.DataError:
         # json.dumps accepts a few bytes jsonb rejects — notably a NUL (0x00)
         # in a string. The cast above then raises, the decrement never lands, and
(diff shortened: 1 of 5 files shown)


workers/queue_backend/pg_barrier.py
```diff
@@ -597,6 +628,99 @@ def _fire_barrier_callback(
...
+        f"UPDATE {qualified('pg_barrier_state')} "
+        "   SET remaining = remaining - 1, "
+        "       results = results || jsonb_build_array(%s::jsonb) "
+        " WHERE execution_id = %s "
+        "RETURNING remaining, results"
+    )
+    for attempt in range(1, _BARRIER_DECREMENT_ATTEMPTS + 1):
+        # Only a connection cached BEFORE this attempt can be a stale idle-reaped
+        # handle; a freshly-reconnected one failing is a genuine error, not a reap.
+        reused = getattr(_local, "conn", None) is not None and not _local.conn.closed
```
muhammad-ali-e: **[Medium] The `reused` fresh-vs-cached distinction is inoperative in the production path.**

The only production caller, `_barrier_pg_decrement`, unconditionally runs the entry-guard `_get_conn()` (~line 765) *before* `_apply_decrement`. That call populates `_local.conn` with a live, open connection. So by the time `reused` is sampled here, `_local.conn` is **always** non-None and open → `reused` is always `True` on attempt 1 — including the case the design explicitly wants to exclude: a brand-new connection whose first statement dies against a genuinely-down DB.

This is **not** a double-count bug (the execute-phase retry is unconditionally safe since nothing committed). The practical impact is bounded: one wasted reconnect + a 0.5s `time.sleep` before surfacing a real DB-down error. But:
- the documented invariant ("a freshly-created one failing is a genuine DB error… so it is not retried", lines 645-647) is not realized through the production entry point — here `reused`'s only real effect is the one-shot bound, which `attempt < _BARRIER_DECREMENT_ATTEMPTS` already provides;
- `test_fresh_conn_execute_failure_is_not_retried` passes only because it calls `_apply_decrement` directly with `_local.conn = None`, bypassing the guard — false confidence;
- the "Mirrors `PgQueueClient.send`'s reused-guard" claim is operationally inaccurate: in `send`, `reused` is sampled before any `_get_conn`.

**Fix:** capture `reused` in `_barrier_pg_decrement` *before* the entry-guard `_get_conn()` and thread it into `_apply_decrement`. Then add an end-to-end test (through `_barrier_pg_decrement`) for the fresh-conn case. Alternatively, if the team accepts the execute retry is always safe regardless, soften the docstrings/comments to stop claiming a fresh conn isn't retried.

_Minor (same line): capture `_local.conn` once to avoid the double lookup —_
```python
cached = getattr(_local, "conn", None)
reused = cached is not None and not cached.closed
```

workers/queue_backend/pg_barrier.py
```diff
@@ -109,6 +122,27 @@ def _get_conn() -> PgConnection:
...
+    dead (a failed rollback proves it dead regardless of the original error).
+
+    Factored out so ``_cursor`` and the decrement phase-split (which can't use
+    ``_cursor`` because it must distinguish execute-phase from commit-phase
+    failures) share one definition of "recover a connection after an error".
+    """
+    conn_dead = isinstance(exc, _CONN_DEAD_ERRORS)
+    try:
+        conn.rollback()
+    except Exception:
```
muhammad-ali-e, line 131: **[Medium] Rollback failure is swallowed with no log, and can silently reclassify a non-connection error as a connection death.**

Two compounding issues:
1. The rollback exception is discarded with zero logging. The original op error does surface later via the caller's `raise`, but the rollback failure itself leaves no trace — if rollback fails for an unexpected reason (e.g. an `InternalError`), that diagnostic is gone.
2. `conn_dead = True` on a failed rollback is the exact gate that decides whether `_apply_decrement` retries the execute phase (line 687). The docstring at lines 659-660 promises a `DataError` "propagates unchanged", but a non-`_CONN_DEAD_ERRORS` error whose rollback *also* fails gets relabeled `conn_dead=True` and routed into the retry path — contradicting the stated invariant. The counter stays safe (execute phase committed nothing), but the real error is masked behind a misleading "cached connection died, reconnecting" warning and re-run.

**Fix:** log the rollback failure with `exc_info` before flipping `conn_dead`:
```python
except Exception:
    logger.warning(
        "PgBarrier: rollback during error recovery failed (original error %s) "
        "— treating the connection as dead and discarding it.",
        type(exc).__name__, exc_info=True,
    )
    conn_dead = True
```

workers/queue_backend/pg_barrier.py
```diff
@@ -597,6 +628,99 @@ def _fire_barrier_callback(
...
+    provably safe:
+
+    - **EXECUTE phase** (``UPDATE … RETURNING`` + ``fetchone``) — the statement
+      runs but is NOT yet committed. A connection-level failure here on a
+      **reused** (cached) connection is the PgBouncer idle-reap: the statement
+      never reached the server, no commit was issued, and the open transaction is
+      rolled back on disconnect. So the decrement provably never landed →
+      reconnect and re-apply it exactly once. (Only a *cached* connection can be a
+      stale idle-reaped handle; a freshly-created one failing is a genuine DB
+      error, not a reap, so it is not retried.) **This safety relies on the
```
muhammad-ali-e: **[Low] The load-bearing "non-autocommit" invariant is documented but unenforced.**

The entire exactly-once argument for the execute-phase retry rests on the connection being non-autocommit (so the uncommitted `UPDATE` rolls back on disconnect). This is currently true only because `create_pg_connection` (`pg_queue/connection.py`) never sets `autocommit` and thus inherits psycopg2's `False` default — it does **not** actively open non-autocommit, so the phrase "`create_pg_connection` opens it that way" slightly overstates. There is no runtime guard: a future change adding `conn.autocommit = True` there (e.g. to match the queue's commit-immediately pattern) would silently reintroduce the double-count this function exists to prevent — and the entry guard's `TRANSACTION_STATUS_IDLE` check does *not* imply non-autocommit.

**Fix:** (a) soften the wording to "non-autocommit (psycopg2's default; `create_pg_connection` does not override it)"; and (b) as a code change, assert the invariant loudly, e.g. `assert not conn.autocommit` at the top of `_apply_decrement`, so it fails loudly rather than corrupting the counter.

muhammad-ali-e (commented):

Reviewed via six specialised agents (Code Reviewer, Silent-Failure Hunter, Type-Design Analyzer, PR Test Analyzer, Comment Analyzer, Code Simplifier) scoped to the two changed files. **Overall: a well-reasoned, safety-critical change.** The execute-vs-commit phase split is structurally sound, the no-double-count argument holds, and the real-DB `pg_terminate_backend` test is a genuine exactly-once proof.

The findings below are mostly Low/Nit. The one that matters most is **#1**: the `reused` fresh-vs-cached distinction is *inoperative in the production path* (the entry guard's `_get_conn()` pre-populates `_local.conn`, so `reused` is always `True` on attempt 1). Not a correctness/double-count bug — the execute-phase retry is unconditionally safe — but the documented invariant isn't realized in prod and the unit test asserting it bypasses the real entry point.

See inline comments for specifics.

### Priority summary
| # | Sev | Location | Finding |
|---|-----|----------|---------|
| 1 | Med | pg_barrier.py:675 | `reused` always True in prod path; fresh-conn-not-retried invariant inoperative |
| 2 | Med | pg_barrier.py:137 | `_recover_after_error` swallows rollback failure with no log; can reclassify a non-conn error as conn-death |
| 3 | Low | pg_barrier.py:647 | Load-bearing non-autocommit invariant is documented but unenforced — add `assert not conn.autocommit` |
| 4 | Low | pg_barrier.py:656 | Docstring names wrong caller for commit-phase teardown (Celery `.link` path has no teardown) |
| 5 | Low | pg_barrier.py:104 | `_CONN_DEAD_ERRORS` triplicated across 3 modules — hoist to shared module to make the no-drift invariant real |
| 6 | Low | pg_barrier.py:631 | Return `tuple[int, Any] \| None` → `NamedTuple` for clarity / typed `results` |
| 7 | Low | pg_barrier.py:701 | Execute-phase give-up `raise` emits no decrement-level log (asymmetric trail) |
| 8 | Nit | pg_barrier.py:719 | `AssertionError` comment's "keeps the type checker honest" rationale is wrong (`None` is a valid return) |
| 9 | Test | test_pg_barrier.py:275 | `time.sleep` backoff never asserted — deleting it wouldn't fail any test |
| 10 | Test | test_pg_barrier.py:329 | Fresh-conn test bypasses the real entry point (ties to #1) |
| 11 | Test | test_pg_barrier.py:368 | Test rationale ("one-shot bound") doesn't match the actual gating branch |

workers/queue_backend/pg_barrier.py
```diff
@@ -597,6 +628,99 @@ def _fire_barrier_callback(
...
+      connection being non-autocommit** (``create_pg_connection`` opens it that
+      way): even if the server actually ran the ``UPDATE`` before the socket
+      dropped, it sits in an uncommitted transaction that is rolled back on
+      disconnect — durability happens ONLY at the commit below — so re-applying
+      can never double-count.
+    - **COMMIT phase** — ``conn.commit()``. A failure here is AMBIGUOUS: the
+      server may have applied the commit before the socket dropped. Re-applying
+      could double-count, so it is NEVER retried — it propagates, and the caller
+      (:func:`run_batch_with_barrier`) tears the barrier down so the execution
```
muhammad-ali-e: **[Low / comment accuracy] This docstring names the wrong caller for commit-phase teardown.**

The direct caller, `_barrier_pg_decrement`, does **not** catch a commit-phase error — its only `except` is `psycopg2.DataError` (~line 791). So a commit-phase failure propagates to whichever ultimate caller drove it:
- `run_batch_with_barrier` (in-body PG path) — caught and `_abort_barrier_in_body` runs → teardown. Fail-fast holds. ✅
- `barrier_pg_decr_and_check` (the Celery `.link` path, today's live path) — **no** teardown; the barrier strands to `expires_at` (`max_retries=0`, no replay).

The no-double-count guarantee still holds (commit is never retried), but the "tears the barrier down so the execution fails fast" claim is false for the path that's actually live today. Suggest rewording to cover both paths and drop the single-caller framing.

workers/queue_backend/pg_barrier.py
```diff
@@ -94,6 +94,19 @@
...
 
 
+# psycopg2 errors that mean the connection itself is dead (dropped socket /
+# PgBouncer recycle / server termination), as opposed to a logical/data error on
+# a live connection. Shared by ``_cursor`` / ``_recover_after_error`` (discard the
+# stale thread-local handle), the idempotent pre-dispatch write retry, and the
+# decrement phase-split retry, so the "was this a connection death?" test can't
+# drift across them. Mirrors ``PgQueueClient._CONN_DEAD_ERRORS`` (UN-3654) and
+# ``PgResultBackend._CONN_DEAD_ERRORS`` (UN-3659), the siblings this models on.
+_CONN_DEAD_ERRORS: Final[tuple[type[Exception], ...]] = (
```
muhammad-ali-e: **[Low] `_CONN_DEAD_ERRORS` is now copy #3 of an identical constant** (`pg_queue/client.py`, `pg_queue/result_backend.py`). The comment here explicitly worries about the "was this a connection death?" test drifting — but the three module-level copies can drift *independently*, which is the exact risk the constant was meant to kill, just moved up to the module boundary.

**Fix:** hoist a single `_CONN_DEAD_ERRORS` into a shared location (e.g. `pg_queue/connection.py`, which all three already import for `create_pg_connection`) and import it in all three. If that's deemed too invasive, at minimum add a test asserting the three tuples are equal.

_Comment nit:_ `PgQueueClient._CONN_DEAD_ERRORS` / `PgResultBackend._CONN_DEAD_ERRORS` are module-level constants, not class attributes — `PgQueueClient._CONN_DEAD_ERRORS` would `AttributeError`. Refer to them by module (`pg_queue.client` / `pg_queue.result_backend`).

workers/queue_backend/pg_barrier.py
```diff
@@ -597,6 +628,99 @@ def _fire_barrier_callback(
     return str(callback_signature.apply_async().id)
 
 
+def _apply_decrement(execution_id: str, result_json: str) -> tuple[int, Any] | None:
```
muhammad-ali-e: **[Low / type design] Consider a `NamedTuple` for the return type.**

`tuple[int, Any] | None` is positional and shape-lossy: it names neither element and `Any` discards the known `list` shape of `results`. The caller decodes it positionally (`remaining, all_results = int(row[0]), row[1]`), and the defensive `int(row[0])` is a tell the declared `int` isn't trusted.

```python
class _DecrementRow(NamedTuple):
    remaining: int
    results: list[Any]
```
Return `_DecrementRow | None`. Callers become `row.remaining` / `row.results` (kills the positional indexing), the `results` shape is documented in the type, and it still unpacks/indexes like a plain tuple for any psycopg2-row consumers — near-zero cost. (No need to go further into a tagged union for the four outcomes — that interpretation rightly belongs to `_barrier_pg_decrement`.)

workers/queue_backend/pg_barrier.py
```diff
@@ -597,6 +628,99 @@ def _fire_barrier_callback(
...
+                    "exactly once.",
+                    execution_id,
+                    type(exc).__name__,
+                    attempt,
+                    _BARRIER_DECREMENT_ATTEMPTS,
+                    exc_info=True,
+                )
+                time.sleep(_BARRIER_RETRY_BACKOFF_SECONDS)
+                continue
+            raise
```
muhammad-ali-e, line 739: **[Low] The execute-phase give-up `raise` emits no decrement-level log.**

The retry branch logs a warning (lines 688-698) and the commit branch logs (709-716), but this propagate path is silent. For a fresh-conn death (`reused` False) or a non-conn error, `_barrier_pg_decrement` only catches `DataError`, so an `OperationalError` here propagates with no barrier-specific breadcrumb — and on the Celery `.link` path the barrier isn't torn down, so it hangs to `expires_at`. The result is an asymmetric trail: retryable blips are logged, but the terminal "gave up / wasn't retryable" decision isn't. Consider one warning before this `raise` naming the execution, error type, and that the decrement is propagating without retry (mirroring the commit-phase log).

workers/queue_backend/pg_barrier.py
```diff
@@ -597,6 +628,99 @@ def _fire_barrier_callback(
...
+                "[exec:%s] PgBarrier decrement: commit failed (%s) — NOT retrying "
+                "(the server may already have applied it; a re-apply would "
+                "double-count). Propagating so the barrier is torn down.",
+                execution_id,
+                type(exc).__name__,
+                exc_info=True,
+            )
+            raise
+        return row
+    # Unreachable: the loop either returns the row or raises. The bare raise keeps
```
muhammad-ali-e: **[Nit / comment accuracy] The "keeps the type checker honest" rationale is wrong.**

The return annotation is `tuple[int, Any] | None`, so an implicit fall-through off the loop would return `None` — a *valid* return (the function legitimately returns `None` at line 718 when `fetchone()` finds no row). A type checker would therefore not flag a missing return whether or not this line exists. The `AssertionError` is a pure defensive runtime guard, not a type-checker aid. Reword to e.g. "Defensive: the loop always returns or raises before here." (Keeping the guard is fine — cheap insurance in concurrency code; just fix the justification.)

workers/tests/test_pg_barrier.py
```diff
@@ -250,6 +261,125 @@ def test_reraises_after_exhausting_attempts(self, _clean_local, monkeypatch):
...
+    idempotent pre-dispatch write it CANNOT retry freely. It self-heals exactly
+    one provably-safe case — an execute-phase failure on a *cached* connection
+    (the PgBouncer idle-reap: the statement never reached the server, so nothing
+    committed) — and refuses every other: a commit-phase failure is ambiguous
+    (the server may have applied it) and a fresh-conn failure is a real DB error.
+    The count is therefore never double-applied. (UN-3660)
+    """
+
+    @pytest.fixture(autouse=True)
+    def _no_sleep(self, monkeypatch):
```
muhammad-ali-e: **[Test gap] The `time.sleep(_BARRIER_RETRY_BACKOFF_SECONDS)` backoff is never asserted.**

This `_no_sleep` fixture (and the inline monkeypatches at ~456/601/631) replace `pg_barrier.time.sleep` with a bare no-op purely to skip the wall-clock wait — none record or assert the call. If someone deleted the `time.sleep` in `_apply_decrement` (line 699), every test would still pass, and a "tighten the retry loop" refactor could silently start hammering a struggling DB.

**Fix:** have the fixture append to a list, then assert `sleeps == [pg_barrier._BARRIER_RETRY_BACKOFF_SECONDS]` in the retry test and `sleeps == []` in the no-retry tests (commit-phase / fresh-conn).

workers/tests/test_pg_barrier.py
```diff
@@ -250,6 +261,125 @@ def test_reraises_after_exhausting_attempts(self, _clean_local, monkeypatch):
...
+
+        with pytest.raises(psycopg2.OperationalError, match="during commit"):
+            pg_barrier._apply_decrement("exec-2", '{"ok": true}')
+
+        assert conn.executes == 1  # the UPDATE ran exactly once
+        assert conn.commits == 1  # commit was attempted exactly once
+        assert reconnects == []  # NEVER reconnected/retried after a commit failure
+        assert conn.closed is True  # the dead conn was discarded
+
+    def test_fresh_conn_execute_failure_is_not_retried(
```
muhammad-ali-e, line 339: **[Test gap — ties to the `reused` finding] This test bypasses the real entry point.**

It calls `_apply_decrement(...)` directly with `_local.conn = None`, so it exercises a fresh-conn state that the production wrapper (`_barrier_pg_decrement` → `barrier_pg_decr_and_check`) never produces: that wrapper's entry-guard `_get_conn()` always pre-populates `_local.conn`, making `reused` always `True` on attempt 1. So this test asserts a property the production path does not actually provide. Add an end-to-end test through `_barrier_pg_decrement` for the fresh-conn-not-retried case (or, if `reused` is fixed to sample before the entry guard, this becomes a true assertion).

workers/tests/test_pg_barrier.py
```diff
@@ -250,6 +261,125 @@ def test_reraises_after_exhausting_attempts(self, _clean_local, monkeypatch):
...
+            lambda **_k: reconnects.append(1) or _FakeConn(),
+        )
+
+        with pytest.raises(psycopg2.DataError):
+            pg_barrier._apply_decrement("exec-4", '{"ok": true}')
+
+        assert reconnects == []  # no retry on a logical/data error
+        assert conn.closed is False  # a live conn after a data error is kept
+
+    def test_reraises_after_second_execute_failure(self, _clean_local, monkeypatch):
```
muhammad-ali-e: **[Test quality / informational] The stated rationale doesn't match the actual gating branch.**

This test is framed around "the one-shot bound" (`attempt < _BARRIER_DECREMENT_ATTEMPTS`), but trace the path: on attempt 1, `_recover_after_error` sets `_local.conn = None` before the `continue`; on attempt 2, `reused` recomputes to `False`, so the retry is refused by **`reused`**, not by the attempt bound — both independently block it. The `attempt <` term can never be the *sole* reason a retry is refused, because every retry discards the connection. Net: the decrement is hard-capped at one self-heal by the `reused` logic regardless of the attempt constant, and bumping `_BARRIER_DECREMENT_ATTEMPTS` would NOT increase self-heals — an untested assumption worth a comment. The propagation *outcome* asserted here is correct; only the rationale needs fixing.

muhammad-ali-e:
Thanks for the thorough review — addressed in `126a3da2e`. Summary:

**[Medium] `reused` inoperative in the production path** — fixed. `_barrier_pg_decrement` now samples `conn_was_cached` **before** the entry-guard `_get_conn()` and threads it into `_apply_decrement(..., reused=...)`, so a genuinely fresh connection is no longer misclassified as cached. Added `test_wrapper_fresh_conn_not_retried_end_to_end` that drives the real wrapper (not `_apply_decrement` directly) and asserts no retry reconnect — the false-confidence gap you flagged.

**[Medium] Rollback failure swallowed** — fixed. `_recover_after_error` now logs the rollback failure with `exc_info` before flipping `conn_dead`, naming that it was the rollback (not the original error) that condemned the connection.

**[Low] `_CONN_DEAD_ERRORS` copy #3** — fixed. Hoisted to `pg_queue.connection` (the module all three import) as `CONN_DEAD_ERRORS`; `client.py`, `result_backend.py` and `pg_barrier.py` now import it, so they can't drift. (`psycopg2` drops out of the two that only used it for the constant.) Also fixed the comment nit — referred to siblings by module, not as class attributes.

**[Low] Non-autocommit invariant unenforced** — softened the wording ("psycopg2's default; `create_pg_connection` does not override it") and pinned it at the **source** with `test_create_pg_connection_is_non_autocommit`. I did *not* add `assert not conn.autocommit` in `_apply_decrement`: the `barrier_db` test fixture deliberately uses an autocommit connection, so the assert would break the real-DB tests — the source-level test is the stronger guard.

**[Low] Commit-phase docstring names wrong caller** — fixed. Now covers both paths: in-body (`run_batch_with_barrier` → teardown, fail-fast) and the live Celery `.link` path (`barrier_pg_decr_and_check`, `max_retries=0` → reclaimed at `expires_at` by the reaper).

**[Low] NamedTuple return** — done. `_apply_decrement` returns `_DecrementRow(remaining, results)`; the caller reads `row.remaining` / `row.results`.

**[Low] Silent execute-phase give-up** — fixed. A conn-dead give-up (fresh-conn death / retry exhausted) now logs a warning before propagating, matching the retryable and commit paths. (A non-conn `DataError` stays logged by the caller's teardown to avoid a double-log.)

**[Nit] Unreachable-raise rationale** — reworded to "defensive runtime guard" (the `| None` return means a type checker wouldn't flag a fall-through anyway).

**[Test gap] Backoff never asserted** — fixed. The `sleeps` fixture records `time.sleep` calls; the retry test asserts `sleeps == [_BARRIER_RETRY_BACKOFF_SECONDS]` and the no-retry tests assert `sleeps == []`.

**[Test gap / quality] Fresh-conn bypass + one-shot rationale** — the new end-to-end test covers the real fresh-conn path; `test_reraises_after_one_retry` now asserts exactly one reconnect + one backoff and its comment explains the attempt-bound vs `reused` interaction.

Re-verified: full `test_pg_barrier.py` + `test_pg_queue_client.py` + `test_pg_result_backend.py` + sibling barrier suites = 158 passed; pre-commit clean. Re-ran the live in-container `pg_terminate_backend` E2E on the final code (all 4 changed files patched into the PG worker fleet, clean boot, cross-module import resolves) → self-heal → decrement exactly once → callback once → barrier consumed.

PR #2134, opened 2026-07-01
UN-3661 [FIX] PG reaper — fast stuck-batch detection + file-execution cascade

muhammad-ali-e:
## Problem

The PG reaper only recovered a stranded execution at the **~6h barrier expiry**, and marked the `workflow_execution` ERROR **without touching its file executions**. So a worker that crashed mid-batch (SIGKILL / OOM / eviction) left the execution stuck `EXECUTING` for ~6h, then flipped to ERROR while its files stayed `EXECUTING` — the exec-b11ba2f3 UI inconsistency (execution ERROR, files "Pipeline completed… EXECUTING").

This closes the crash/stuck-recovery gap and brings PG to parity with Celery, whose per-task `FILE_PROCESSING_TASK_TIME_LIMIT` (2.5h) already fails a stuck task fast.

## Part 1 — fast, progress-based detection

- Add `pg_barrier_state.last_progress_at` (**unindexed**), re-stamped to `now()` on enqueue and on every decrement.
  - Unindexed on purpose: the decrement writes it every batch-completion, so indexing it would break the decrement's heap-only-tuple (HOT) update. Unindexed → **no index churn on the hot path**; the reaper seq-scans this tiny (in-flight-only) table cheaply, and it's directly queryable ("when did this barrier last progress?").
- The reaper marks a barrier stranded once `last_progress_at` is older than `WORKER_PG_BATCH_STUCK_TIMEOUT_SECONDS` (default **9000** = 2.5h, matching Celery's file-processing limit) **OR** it passes the fixed 6h `expires_at` cap (last-resort backstop). One shared `_STRANDED_PREDICATE` drives the SELECT, the pre-mark re-check, and the cleanup DELETE.
- A **per-progress** window is invariant to `MAX_PARALLEL_FILE_BATCHES` (a dynamic per-org config), so it **never false-fails a legitimately long multi-batch run** — any completed batch refreshes the deadline; only a genuinely stalled one is reaped.

## Part 2 — atomic file-execution cascade

- The internal `update_status` endpoint gains a `cascade_terminal_files` flag: when set **and** the new status is terminal, it also marks the execution's non-terminal (PENDING/EXECUTING) file executions to the same terminal status **in the same transaction** — so execution + files can't diverge. Terminal files (COMPLETED/ERROR/STOPPED) are left untouched; idempotent.
- The reaper passes the flag on recovery, so a recovered strand's execution and its files reach a consistent terminal state.

Chose the atomic backend endpoint (one transaction, one round trip) over a reaper-side multi-call cascade — better for long-term maintainability, performance under mass-strand recovery, and it has no partial-state window.

## Tests

- **Worker (227 pass):** barrier enqueue/decrement stamp `last_progress_at`; reaper OR-detection (stalled reaped, progressing untouched, all existing recovery guards intact) + cascade flag asserted.
- **Backend:** `test_cascade_terminal_files.py` — non-terminal → terminal, terminal untouched, idempotent.
- **Dev-tested E2E:** reaper detection against **real Postgres** (a stalled barrier reaped, a *progressing* one untouched, marked ERROR + `cascade_terminal_files=True`); the cascade against the **real DB** (2 non-terminal files → ERROR, COMPLETED untouched).

## Notes

- Migration `0011` adds the column (no index — deliberate). PG barrier is not yet on staging, so this is additive.
- No behaviour change on the Celery/chord path; the barrier's fixed 6h `expires_at` backstop is preserved.

### DIFF
--- a/backend/pg_queue/models.py
+++ b/backend/pg_queue/models.py
@@ -185,9 +185,10 @@ class PgBarrierState(models.Model):
     drives ``remaining`` to 0 dispatches the aggregating callback and deletes the
     row. A header-task failure aborts the barrier by deleting the row outright
     (``DELETE … RETURNING`` — atomic claim+teardown), so the callback can never
-    fire with partial results. ``expires_at`` bounds an orphaned barrier (header
-    tasks that never complete); a periodic sweep job (not yet implemented) is the
-    intended reclaim backstop.
+    fire with partial results. The ``pg-queue-reaper`` marks a stranded execution
+    ERROR when ``last_progress_at`` goes stale (no batch completed for
+    ``WORKER_PG_BATCH_STUCK_TIMEOUT_SECONDS``) or, as an absolute backstop, when
+    ``expires_at`` passes (see ``workers/queue_backend/pg_queue/reaper.py``).
 
     Managed=True / generated migration — no DB-side function, extension-free
     (UN-3533), same posture as ``PgQueueMessage``.
@@ -206,10 +207,22 @@ class PgBarrierState(models.Model):
     # Aggregated header-task results, appended in completion order (JSONB array).
     results = models.JSONField(default=list)
     created_at = models.DateTimeField(default=timezone.now)
-    # Orphan bound (Redis-TTL equivalent): a barrier whose header tasks never
-    # complete is reclaimable past this. Must exceed the longest execution
-    # wall-clock, same budgeting as WORKER_BARRIER_KEY_TTL_SECONDS.
+    # Absolute orphan cap (Redis-TTL equivalent, WORKER_BARRIER_KEY_TTL_SECONDS,
+    # default 6h): a last-resort backstop the reaper also sweeps. Fixed at enqueue
+    # — the *fast* stuck signal is last_progress_at below.
     expires_at = models.DateTimeField()
+    # Progress liveness (UN-3661): re-stamped to now() on enqueue AND on every
+    # decrement, so it tracks when a batch last completed. The reaper marks the
+    # execution ERROR once it is older than WORKER_PG_BATCH_STUCK_TIMEOUT_SECONDS
+    # (default 2.5h, = Celery's per-task FILE_PROCESSING_TASK_TIME_LIMIT) — a
+    # per-PROGRESS window that is invariant to MAX_PARALLEL_FILE_BATCHES (dynamic
+    # per-org), so it never false-fails a legitimately long multi-batch run.
+    # Deliberately UNINDEXED: it's written on every decrement, so indexing it would
+    # break the heap-only-tuple (HOT) update the decrement relies on — and the
+    # reaper's sweep of this tiny (in-flight-only) table is a cheap seq scan.
+    # Directly queryable ("when did this barrier last progress?") — the PG-queue
+    # "stuck-detection = plain SQL" property.
+    last_progress_at = models.DateTimeField(default=timezone.now)
 
     class Meta:
         db_table = "pg_barrier_state"
--- a/backend/pg_queue/models.py
+++ b/backend/pg_queue/models.py
@@ -214,8 +214,9 @@ class PgBarrierState(models.Model):
     # Progress liveness (UN-3661): re-stamped to now() on enqueue AND on every
     # decrement, so it tracks when a batch last completed. The reaper marks the
     # execution ERROR once it is older than WORKER_PG_BATCH_STUCK_TIMEOUT_SECONDS
-    # (default 2.5h, = Celery's per-task FILE_PROCESSING_TASK_TIME_LIMIT) — a
-    # per-PROGRESS window that is invariant to MAX_PARALLEL_FILE_BATCHES (dynamic
+    # (default 2.5h — in the same band as Celery's per-task
+    # FILE_PROCESSING_TASK_TIME_LIMIT, which ships 2h–3h) — a per-PROGRESS window
+    # that is invariant to MAX_PARALLEL_FILE_BATCHES (dynamic
     # per-org), so it never false-fails a legitimately long multi-batch run.
     # Deliberately UNINDEXED: it's written on every decrement, so indexing it would
     # break the heap-only-tuple (HOT) update the decrement relies on — and the
--- a/backend/workflow_manager/internal_serializers.py
+++ b/backend/workflow_manager/internal_serializers.py
@@ -175,6 +175,11 @@ class WorkflowExecutionStatusUpdateSerializer(serializers.Serializer):
 
     status = serializers.ChoiceField(choices=ExecutionStatus.choices)
     error_message = serializers.CharField(required=False, allow_blank=True)
+    # When true AND the new status is terminal, also mark this execution's
+    # non-terminal (PENDING/EXECUTING) file executions to the same terminal status,
+    # atomically. The reaper sets this so a stranded execution and its files reach a
+    # consistent terminal state (else execution=ERROR while its files stay EXECUTING).
+    cascade_terminal_files = serializers.BooleanField(required=False, default=False)
     total_files = serializers.IntegerField(
         required=False, min_value=0
     )  # Allow 0 but backend will only update if > 0
--- a/backend/workflow_manager/internal_serializers.py
+++ b/backend/workflow_manager/internal_serializers.py
@@ -189,12 +189,13 @@ class WorkflowExecutionStatusUpdateSerializer(serializers.Serializer):
     execution_time = serializers.FloatField(required=False, min_value=0)
 
     def validate(self, attrs):
-        """Reject impossible file-count aggregates + a meaningless cascade combo.
+        """Reject a meaningless cascade combo + impossible file-count aggregates."""
+        self._validate_cascade(attrs)
+        self._validate_file_aggregates(attrs)
+        return attrs
 
-        Per-field min_value=0 catches negatives, but successful + failed >
-        total or either component > total slips through and skews the
-        outcome-based notification filter downstream.
-        """
+    @staticmethod
+    def _validate_cascade(attrs) -> None:
         # cascade_terminal_files only makes sense when the new status is terminal
         # (it's a silent no-op otherwise) — reject the illegal combo at the boundary
         # rather than accepting-and-ignoring it.
@@ -210,6 +211,12 @@ class WorkflowExecutionStatusUpdateSerializer(serializers.Serializer):
                 }
             )
 
+    @staticmethod
+    def _validate_file_aggregates(attrs) -> None:
+        """Per-field min_value=0 catches negatives, but successful + failed > total
+        or either component > total slips through and skews the outcome-based
+        notification filter downstream.
+        """
         total = attrs.get("total_files")
         successful = attrs.get("successful_files")
         failed = attrs.get("failed_files")
@@ -221,23 +228,18 @@ class WorkflowExecutionStatusUpdateSerializer(serializers.Serializer):
                         "total_files": "total_files is required when file aggregates are provided."
                     }
                 )
-        else:
-            if successful is not None and successful > total:
-                raise serializers.ValidationError(
-                    {"successful_files": "successful_files cannot exceed total_files."}
-                )
-            if failed is not None and failed > total:
-                raise serializers.ValidationError(
-                    {"failed_files": "failed_files cannot exceed total_files."}
-                )
-            if (
-                successful is not None
-                and failed is not None
-                and successful + failed > total
-            ):
-                msg = "successful_files + failed_files cannot exceed total_files."
-                raise serializers.ValidationError({"non_field_errors": msg})
-        return attrs
+            return
+        if successful is not None and successful > total:
+            raise serializers.ValidationError(
+                {"successful_files": "successful_files cannot exceed total_files."}
+            )
+        if failed is not None and failed > total:
+            raise serializers.ValidationError(
+                {"failed_files": "failed_files cannot exceed total_files."}
+            )
+        if successful is not None and failed is not None and successful + failed > total:
+            msg = "successful_files + failed_files cannot exceed total_files."
+            raise serializers.ValidationError({"non_field_errors": msg})
 
 
 class OrganizationContextSerializer(serializers.Serializer):
--- a/backend/workflow_manager/internal_serializers.py
+++ b/backend/workflow_manager/internal_serializers.py
@@ -189,12 +189,27 @@ class WorkflowExecutionStatusUpdateSerializer(serializers.Serializer):
     execution_time = serializers.FloatField(required=False, min_value=0)
 
     def validate(self, attrs):
-        """Reject impossible file-count aggregates.
+        """Reject impossible file-count aggregates + a meaningless cascade combo.
 
         Per-field min_value=0 catches negatives, but successful + failed >
         total or either component > total slips through and skews the
         outcome-based notification filter downstream.
         """
+        # cascade_terminal_files only makes sense when the new status is terminal
+        # (it's a silent no-op otherwise) — reject the illegal combo at the boundary
+        # rather than accepting-and-ignoring it.
+        if attrs.get("cascade_terminal_files") and not ExecutionStatus.is_completed(
+            attrs["status"]
+        ):
+            raise serializers.ValidationError(
+                {
+                    "cascade_terminal_files": (
+                        "cascade_terminal_files=True requires a terminal status "
+                        "(COMPLETED/ERROR/STOPPED)."
+                    )
+                }
+            )
+
         total = attrs.get("total_files")
         successful = attrs.get("successful_files")
         failed = attrs.get("failed_files")
--- a/backend/workflow_manager/internal_views.py
+++ b/backend/workflow_manager/internal_views.py
@@ -550,13 +550,15 @@ class WorkflowExecutionInternalViewSet(viewsets.ReadOnlyModelViewSet):
         from workflow_manager.file_execution.models import WorkflowFileExecution
         from workflow_manager.workflow_v2.enums import ExecutionStatus
 
-        terminal = {
-            ExecutionStatus.COMPLETED.value,
-            ExecutionStatus.ERROR.value,
-            ExecutionStatus.STOPPED.value,
-        }
+        terminal = ExecutionStatus.terminal_values()  # canonical COMPLETED/ERROR/STOPPED
         if status_enum.value not in terminal:
             return
+        # Bulk .update() deliberately bypasses WorkflowFileExecution.update_status():
+        # this is a give-up recovery, so per-file execution_time is unknown (stays
+        # NULL) and the parent's failed/successful_files aggregates are NOT
+        # reconciled here — the execution's own terminal status (set above) is the
+        # source of truth (is_failure_run / notifications read it), and the
+        # serializer already derives failed counts for a terminal-failure run.
         cascaded = (
             WorkflowFileExecution.objects.filter(workflow_execution=execution)
             .exclude(status__in=terminal)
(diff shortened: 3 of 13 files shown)


workers/queue_backend/barrier.py
```diff
@@ -81,6 +81,47 @@ def barrier_ttl_seconds() -> int:
     return value
 
 
+# PG barrier stuck-timeout — the "no progress" deadline for a PG-backed barrier.
+# Unlike the Redis backend's fixed key TTL, the PG barrier's ``expires_at`` SLIDES:
```
muhammad-ali-e: **[Comment accuracy · high-value]** This says `expires_at` *SLIDES* and is re-stamped `now() + this` on every decrement. That is wrong on the field: `_apply_decrement` (pg_barrier.py:699-715) only re-stamps **`last_progress_at`** and deliberately leaves `expires_at` untouched to preserve the HOT update. `expires_at` is fixed at enqueue to `now() + WORKER_BARRIER_KEY_TTL_SECONDS` (6h). The sliding column that this stuck-timeout is compared against is `last_progress_at`. Every other comment in this PR (models.py:207-225, pg_barrier.py docstring item 4) gets this right — only this one, sitting on the authoritative config function, describes the pre-refactor design. Suggest: rewrite to say `last_progress_at` slides (re-stamped to bare `now()`), and note the reaper applies the timeout at query time via `_STRANDED_PREDICATE`.

workers/queue_backend/barrier.py
```diff
@@ -81,6 +81,47 @@ def barrier_ttl_seconds() -> int:
...
+# it's re-stamped to ``now() + this`` on enqueue AND on every decrement, so a
+# barrier only "expires" (→ the reaper marks it ERROR) after this long with NO
+# batch completing. Default 2.5h matches Celery's per-task ``FILE_PROCESSING_TASK_
+# TIME_LIMIT`` (9000s), giving crash / runaway parity: a batch whose worker died
+# (or that runs longer than Celery would allow) stalls progress and is reaped at
+# ~this bound, PROMPTLY (not at the old ~6h barrier expiry). A per-PROGRESS window
+# is invariant to how many batches run in parallel (``MAX_PARALLEL_FILE_BATCHES``
+# is dynamic per-org), so — unlike a per-execution age cap — it never false-fails a
+# legitimately long multi-batch run: any completion refreshes the deadline.
+_DEFAULT_BARRIER_STUCK_TIMEOUT_SECONDS = 9000  # 2.5h, = Celery FILE_PROCESSING hard limit
```
muhammad-ali-e: **[Comment accuracy]** `= Celery FILE_PROCESSING hard limit` overstates an equality that isn't true. `FILE_PROCESSING_TASK_TIME_LIMIT` is a tunable env var shipping **7200 (2h)** in `workers/sample.env` and **10800 (3h)** in `docker/sample.env` — never 9000. 9000 is a midpoint, not an equality, and the claim silently rots the moment an operator tunes the var. Reframe as an independent default that *sits in the same band* (2h–3h). Same inaccurate equivalence recurs at models.py:217, pg_barrier.py:46, and test_pg_barrier.py:97 — worth fixing all four together. (Compare the accurate range comment already at redis_barrier.py:107-108.)

workers/queue_backend/barrier.py
```diff
@@ -81,6 +81,47 @@ def barrier_ttl_seconds() -> int:
...
+
+
+def barrier_stuck_timeout_seconds() -> int:
```
muhammad-ali-e, line 111: **[Simplification]** `barrier_stuck_timeout_seconds()` is a line-for-line clone of `barrier_ttl_seconds()` (read env → default → `int()` parse-raise → `<=0` raise). Extract a private `_positive_int_env(var, default, human_default) -> int` and have both call it — removes ~20 duplicated lines and guarantees the two validators can't drift. Behavior (including the two distinct error phrasings and the `WORKER_*` substrings the tests match on) is preserved.

backend/pg_queue/models.py
```diff
@@ -206,10 +207,22 @@ class PgBarrierState(models.Model):
...
-    # complete is reclaimable past this. Must exceed the longest execution
-    # wall-clock, same budgeting as WORKER_BARRIER_KEY_TTL_SECONDS.
+    # Absolute orphan cap (Redis-TTL equivalent, WORKER_BARRIER_KEY_TTL_SECONDS,
+    # default 6h): a last-resort backstop the reaper also sweeps. Fixed at enqueue
+    # — the *fast* stuck signal is last_progress_at below.
     expires_at = models.DateTimeField()
+    # Progress liveness (UN-3661): re-stamped to now() on enqueue AND on every
+    # decrement, so it tracks when a batch last completed. The reaper marks the
+    # execution ERROR once it is older than WORKER_PG_BATCH_STUCK_TIMEOUT_SECONDS
+    # (default 2.5h, = Celery's per-task FILE_PROCESSING_TASK_TIME_LIMIT) — a
```
muhammad-ali-e: **[Comment accuracy]** `= Celery's per-task FILE_PROCESSING_TASK_TIME_LIMIT` — same inaccurate equality as barrier.py:95. That limit ships 7200/10800 across deployments, never 9000. State it as an independent default in the same band, not an equality.

workers/queue_backend/pg_barrier.py
```diff
@@ -38,10 +39,16 @@
...
    header failure).
-4. Orphan bound: like the Redis backend's key TTL, ``expires_at`` bounds a
-   barrier whose header tasks never complete. **No expiry reclaim ships in this
-   phase** — a periodic sweep job (keyed on ``pg_barrier_expires_idx``) is the
-   intended backstop and is future work.
+4. Stuck bound: ``last_progress_at`` is re-stamped to ``now()`` on enqueue AND on
+   every decrement, tracking when a batch last completed. The
+   :mod:`~queue_backend.pg_queue.reaper` marks the stranded execution ERROR once
+   ``last_progress_at`` is older than ``WORKER_PG_BATCH_STUCK_TIMEOUT_SECONDS``
+   (default 2.5h, = Celery's per-task ``FILE_PROCESSING_TASK_TIME_LIMIT``) — a
```
muhammad-ali-e: **[Comment accuracy]** `= Celery's per-task FILE_PROCESSING_TASK_TIME_LIMIT` — same inaccurate equality as barrier.py:95 (ships 7200/10800, not 9000). Reframe as an independent default in the same band.

workers/tests/test_pg_barrier.py
```diff
@@ -92,6 +92,24 @@ def test_invalid_raises(self, monkeypatch, bad):
             barrier_mod.barrier_ttl_seconds()
 
 
+class TestStuckTimeoutEnv:
+    # The PG barrier's sliding expires_at window (UN-3661) — distinct from the
```
muhammad-ali-e: **[Comment accuracy]** "The PG barrier's sliding **expires_at** window" — the sliding column is `last_progress_at`; `expires_at` is the fixed 6h cap. The test body below (`test_decrement_refreshes_last_progress_at`) correctly asserts on `last_progress_at`, so this comment contradicts its own test. Change to "sliding `last_progress_at` window".

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -441,11 +468,14 @@ def recover_expired_barriers(
     ``conn`` runs in manual-commit mode; on any error we roll back before
     continuing/re-raising so the connection isn't left in an aborted-txn state.
     """
+    if stuck_timeout_seconds is None:
+        stuck_timeout_seconds = barrier_stuck_timeout_seconds()
```
muhammad-ali-e, line 483: **[Silent failure · HIGH]** `barrier_stuck_timeout_seconds()` is resolved here, *outside* the `try`, and it raises `ValueError` on any non-integer / non-positive `WORKER_PG_BATCH_STUCK_TIMEOUT_SECONDS`. `tick()` re-raises to `run()`, which logs only a generic `logger.exception("cycle failed; continuing")`. Net effect: one fat-fingered env var makes **every** sweep fail and recover **nothing** — the entire recovery net this module exists to be is silently disabled, indistinguishable from a transient DB blip. This is a robustness regression (the pre-PR reaper read no such var). Fix: validate this knob **once at reaper startup** (in `PgReaper.__init__`, alongside the existing interval/lease validation) so a bad value crashes loudly at boot; or at least `except ValueError` here and log a distinct `recovery disabled — invalid WORKER_PG_BATCH_STUCK_TIMEOUT_SECONDS`.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -319,6 +331,10 @@ def _mark_stranded_error(
         status=ExecutionStatus.ERROR.value,
         error_message=f"[reaper-recovery] Execution stranded: {reason}.",
         organization_id=organization_id,
+        # Cascade ERROR to the execution's non-terminal file executions in the same
+        # backend transaction — else the execution goes ERROR while its files stay
+        # EXECUTING (the b11ba2f3 inconsistency).
+        cascade_terminal_files=True,
```
muhammad-ali-e, line 337: **[Silent failure · MEDIUM]** `_mark_stranded_error` ignores the return of `update_workflow_execution_status`, then the caller proceeds to `DELETE FROM pg_barrier_state`. This trusts the write path to *raise* on failure — currently true, but note the deliberate asymmetry: the sibling read path `_execution_status` (reaper.py:284) explicitly checks `response.success` and raises *because the client returns `success=False` rather than raising*. If the update endpoint ever returns a 2xx envelope with `success=False` (or the write client is made non-raising like the read client), the reaper sails into the DELETE and permanently erases the only recovery handle while the execution stays non-terminal forever — the exact catastrophic strand this module prevents. Mirror the read path: assert `response.success` here and raise if falsy.

backend/workflow_manager/internal_views.py
```diff
@@ -531,6 +537,44 @@ def update_status(self, request, id=None):
...
+            ExecutionStatus.COMPLETED.value,
+            ExecutionStatus.ERROR.value,
+            ExecutionStatus.STOPPED.value,
+        }
+        if status_enum.value not in terminal:
+            return
+        cascaded = (
+            WorkflowFileExecution.objects.filter(workflow_execution=execution)
+            .exclude(status__in=terminal)
+            .update(
```
muhammad-ali-e, line 565: **[Silent failure · LOW-MED]** This bulk `QuerySet.update()` bypasses `WorkflowFileExecution.update_status()`, the only place that computes `execution_time` for terminal states and reconciles the parent's `failed_files`/`successful_files` aggregates. Consequences for reaper-recovered files: `execution_time` stays NULL (unlike the normal path — billing/analytics reading it directly will see nulls), and an ERROR run shows `failed_files=0` in the UI, a confusing inconsistency during incident debugging. The top-line status is correct (so `is_failure_run` / notifications still fire — verified). Suggest either reconciling the aggregates + `execution_time` here, or a code comment documenting that they are intentionally not reconciled on reaper recovery.

backend/workflow_manager/internal_serializers.py
```diff
@@ -175,6 +175,11 @@ class WorkflowExecutionStatusUpdateSerializer(serializers.Serializer):
 
     status = serializers.ChoiceField(choices=ExecutionStatus.choices)
     error_message = serializers.CharField(required=False, allow_blank=True)
+    # When true AND the new status is terminal, also mark this execution's
+    # non-terminal (PENDING/EXECUTING) file executions to the same terminal status,
+    # atomically. The reaper sets this so a stranded execution and its files reach a
+    # consistent terminal state (else execution=ERROR while its files stay EXECUTING).
+    cascade_terminal_files = serializers.BooleanField(required=False, default=False)
```
muhammad-ali-e, line 182: **[Type design · MEDIUM]** `cascade_terminal_files=True` has *partial* semantics: it is a silent no-op unless `status` is terminal (internal_views.py:558 early-returns). So the serializer accepts a meaningless-but-illegal combination — "cascade to files, but the new status is non-terminal" — and quietly ignores it. Add a `validate()` that rejects `cascade_terminal_files=True` when `status` is non-terminal, turning a silent no-op into a clear 400. Makes the terminality invariant (currently only in a comment) enforced at the boundary.

backend/workflow_manager/internal_views.py
```diff
@@ -531,6 +537,44 @@ def update_status(self, request, id=None):
...
+        terminal while its files stay EXECUTING (the b11ba2f3 inconsistency).
+
+        No-op unless *status_enum* is terminal. Runs in the caller's transaction, so
+        the execution status and the file cascade commit together. Idempotent — a
+        re-run finds those files already terminal and updates nothing.
+        """
+        from workflow_manager.file_execution.models import WorkflowFileExecution
+        from workflow_manager.workflow_v2.enums import ExecutionStatus
+
+        terminal = {
```
muhammad-ali-e: **[Type design / DRY · LOW]** This inline `terminal = {COMPLETED, ERROR, STOPPED}` duplicates the canonical terminal definition in `ExecutionStatus.is_completed` (core/data_models.py:557-568) — which the reaper itself already uses. Add a fourth terminal status and `is_completed` updates while this set silently does not. Clean fix: expose a canonical `ExecutionStatus.terminal_values()` (or `TERMINAL` frozenset) that both `is_completed` and this `.exclude(status__in=...)` consume. (The ORM `exclude` genuinely needs the *set*, which is why swapping only the guard to `is_completed()` wouldn't be an improvement — hence a shared constant.)

workers/tests/test_pg_reaper.py
```diff
@@ -700,6 +701,9 @@ def test_recovers_only_expired_marks_error_and_cleans_up(self, barrier_conn):
         assert call.status == "ERROR"
         assert call.organization_id == "org-1"
         assert "never completed" in call.error_message  # remaining>0
+        # Cascade the terminal status to the stranded execution's non-terminal
+        # files (the b11ba2f3 fix) — else execution=ERROR while files stay EXECUTING.
+        assert call.cascade_terminal_files is True
```
muhammad-ali-e, line 722: **[Test gap · HIGH]** Good that the cascade flag is asserted here. But the headline behavior of this PR — reaping via **stale `last_progress_at` while `expires_at` is still in the future** — has *no* end-to-end reaper test. `_seed` (test_pg_reaper.py:652) only ever sets `expires_at`; every "stranded" row here matches `_STRANDED_PREDICATE` exclusively through the `expires_at < now()` clause, with `last_progress_at` always fresh. The whole `last_progress_at` clause could be deleted from the SELECT / `_still_stranded` / DELETE and this suite would still pass. Please add (real-Postgres): (1) seed `expires_at = now()+6h`, `last_progress_at = now()-3h`, call with `stuck_timeout_seconds=120` → asserts ERROR + `cascade_terminal_files=True` + rows deleted; (2) a boundary negative (fresh progress, long run → NOT reaped); (3) the `_still_stranded` re-arm race where a decrement refreshes `last_progress_at` mid-recovery must abort the ERROR mark (existing re-arm test only refreshes `expires_at`).

muhammad-ali-e (commented):
## PR Review Toolkit — automated multi-agent review (UN-3661)

Ran six agents (Code Reviewer, Silent Failure Hunter, Type Design Analyzer, PR Test Analyzer, Comment Analyzer, Code Simplifier) over the 12-file diff vs `feat/UN-3445-pg-queue-integration`. The change is coherent and correct in its core mechanics (the stranded predicate binds params correctly, 9000s stuck-timeout fires before the 6h cap, the decrement stays a HOT update, the terminal set matches the enum). No Critical/blocking bugs.

**Highest-value fixes**
1. **Comment error (3 agents, high-confidence):** barrier.py:85 says `expires_at` slides — it's `last_progress_at` that slides; `expires_at` is fixed at enqueue. Contradicts this PR's own correct comments.
2. **Silent failure (HIGH):** reaper.py:472 resolves `barrier_stuck_timeout_seconds()` outside the `try`; a bad env var disables the *entire* recovery net as a generic repeating traceback. Validate at startup.
3. **Test gap (HIGH):** the stuck-via-`last_progress_at`-but-not-expired reaping path — the reason this PR exists — has no end-to-end reaper test.
4. **Inaccurate equivalence (4 sites):** `9000 = FILE_PROCESSING_TASK_TIME_LIMIT` — that var ships 7200/10800, never 9000.

**Also noted (not inline, outside diff hunks):**
- reaper.py:366/379-380 — `_recover_one_barrier` docstring still describes the DELETE guard as `expires_at < now()`; it's now `_STRANDED_PREDICATE`. (The sibling `_still_stranded` docstring *was* updated — this one was missed.)
- reaper.py:8-9/27 — module docstring still presents `expires_at`/~6h as the sole trigger/bottom-out; update to describe the new fast progress-staleness path.
- internal_views.py:534 — the status-update handler (now also wrapping the cascade) uses `logger.error(...)` without `exc_info`; a cascade DB error loses its traceback. Prefer `logger.exception`.
- Optional hardening: add a HOT-safe `CHECK (last_progress_at >= created_at)` to match the existing `expires_at > created_at` constraint.
- `b11ba2f3` referenced in 4 new comments doesn't resolve as a git object; elsewhere it's written `exec-b11ba2f3` (an execution id). Keep the `exec-` prefix or cite UN-3661.

Inline comments below carry file:line and suggested fixes. All are non-blocking; (1)-(3) are worth addressing before merge.

muhammad-ali-e:
Thanks — addressed in `a620ca480`. Summary:

**[Silent failure · HIGH] reaper.py:472** — the stuck-timeout is now resolved + validated **once at `PgReaper.__init__`** (alongside the interval/lease validation), so a garbled `WORKER_PG_BATCH_STUCK_TIMEOUT_SECONDS` crashes loudly at boot instead of silently failing every sweep through `run()`'s generic catch. `tick()` passes the resolved value.

**[Silent failure · HIGH] reaper.py:337** — `_mark_stranded_error` now checks the update `response.success` and raises if falsy (mirroring the read path), so a non-raising `success=False` can never let the reaper DELETE the barrier (the only recovery handle) while the execution stays non-terminal. Added a test.

**[Test gap · HIGH] test_pg_reaper.py:706** — the headline `last_progress_at` path now has real-Postgres coverage: (1) stale `last_progress_at` + **future** `expires_at` is reaped (proves the clause is load-bearing — deleting it would now fail this test); (2) a *progressing* barrier (fresh progress, future cap) is NOT reaped; (3) a progress-refresh mid-recovery aborts the ERROR mark (`_still_stranded` re-arm via `last_progress_at`). `_seed` gained a `last_progress` param.

**[Type design · MEDIUM] serializer** — `validate()` now rejects `cascade_terminal_files=True` with a non-terminal status (400) instead of accept-and-ignore.

**[Type design / DRY · LOW] terminal set** — added `ExecutionStatus.terminal_values()` (core) as the single source; `is_completed()` and the cascade's `status__in` exclude both consume it (behaviour identical — verified COMPLETED/STOPPED/ERROR).

**[Simplification] barrier.py** — extracted `_positive_int_env()` shared by `barrier_ttl_seconds` / `barrier_stuck_timeout_seconds` (removed the duplicated parser; error phrasings + `WORKER_*` substrings preserved).

**[Silent failure · LOW-MED] cascade bulk-update** — documented that it intentionally does not reconcile per-file `execution_time` / the parent aggregates on reaper recovery (the execution's terminal status is the source of truth; the serializer derives failed counts).

**[Comment accuracy ×5]** — fixed the `last_progress_at` (not `expires_at`) sliding-column description on the authoritative config function + models/pg_barrier/test comments, and reframed the `= FILE_PROCESSING_TASK_TIME_LIMIT` equality as "same band (ships 2h–3h)".

**Feature-flag / common-code safety** (double-checked per an offline review point): the `workflow_manager` cascade is opt-in — `cascade_terminal_files=True` is set by exactly one caller (the reaper), every other (Celery-path) status-update caller uses the default `False`, and the reaper only ever touches PG-path executions (`pg_barrier_state` rows). `is_completed` is behaviour-identical. So no change for a non-PG / flag-off execution.

234 worker tests pass; pre-commit clean.

PR #2137, opened 2026-07-01
UN-3662 [FIX] PG queue — skip stale/redelivered batch for a reaper-recovered (terminal) execution

muhammad-ali-e:
## What

On the PG queue (at-least-once delivery), a file batch can be redelivered — or a stale batch consumed — *after* the reaper has recovered the execution (marked it ERROR) and deleted its barrier. Reprocessing **resurrected** the execution back to EXECUTING and re-ran its files (double LLM / destination write), ending in an inconsistent state: execution=EXECUTING, files=COMPLETED, with a stale reaper-recovery error message. This breaks the reaper's terminal state being one-way.

This adds a **validate-first terminal guard** on the file-processing batch path:

- `_setup_execution_context` raises `_TerminalExecutionSkip` when the fetched execution is already terminal (COMPLETED/ERROR/STOPPED), *before* touching status or processing.
- `_run_batch_stages` catches it and returns a benign zero-work result — no EXECUTING flip, no file pre-create, no LLM/destination work.

## Why it's safe for the Celery path

The guard is gated to the PG path via `is_pg` (`barrier_context` present == the transport the `pg_queue` flag selects). On Celery, `is_pg=False` and the guard returns immediately, so the existing flow is byte-for-byte unchanged. The resurrection it prevents is PG-specific: Celery lost tasks are never redelivered, so they can't be resurrected.

## Testing

- New `test_terminal_execution_guard.py`: guard raises for terminal on PG, **no-ops for terminal on Celery** (regression-safety), passes for active/unknown status, skips-without-processing, and proceeds normally when active.
- `test_chord_callback_boundary.py` transport-routing tests strengthened to assert `is_pg=False` on Celery / `is_pg=True` on PG.
- Full worker suite shows **no new failures vs baseline** — the pre-existing order-dependent flakes and one hanging consumer test are unrelated (verified identical with these changes stashed).

Manual verification of the resurrection scenario will follow in the dev environment (reap → redeliver → confirm the execution stays ERROR).

Part of the PG-queue crash-recovery work (follows UN-3661). Companion exactly-once hardening tracked separately (UN-3663).

### DIFF
--- a/workers/file_processing/tasks.py
+++ b/workers/file_processing/tasks.py
@@ -12,7 +12,10 @@ from typing import Any
 
 from queue_backend import worker_task
 from queue_backend.barrier import BarrierContext
-from queue_backend.pg_barrier import run_batch_with_barrier
+from queue_backend.pg_barrier import (
+    SKIPPED_TERMINAL_EXECUTION_KEY,
+    run_batch_with_barrier,
+)
 
 # Import shared worker infrastructure
 from shared.api import InternalAPIClient
@@ -267,16 +270,24 @@ def _raise_if_execution_terminal(
 
 
 def _terminal_skip_result(batch_data: FileBatchData) -> dict[str, Any]:
-    """Benign zero-work batch result for a batch skipped because its execution
-    is already terminal (UN-3662) — no files processed, all counts zero.
+    """Batch result for a batch skipped because its execution is already terminal
+    (UN-3662). The files were never attempted; per the accounting contract from
+    68d137a8 (unaccounted files count as failed, not in-progress) they are
+    reported as failed, so a consumer that ever aggregates this result does not
+    render them as perpetually in-progress. The ``SKIPPED_TERMINAL_EXECUTION_KEY``
+    marker tells ``run_batch_with_barrier`` to bypass the barrier decrement (the
+    reaper has by definition already torn the barrier down).
     """
-    return BatchExecutionResult(
-        total_files=len(batch_data.files),
+    n_files = len(batch_data.files)
+    result = BatchExecutionResult(
+        total_files=n_files,
         successful_files=0,
-        failed_files=0,
+        failed_files=n_files,
         execution_time=0.0,
         organization_id=batch_data.file_data.organization_id,
     ).to_dict()
+    result[SKIPPED_TERMINAL_EXECUTION_KEY] = True
+    return result
 
 
 def _run_batch_stages(
--- a/workers/file_processing/tasks.py
+++ b/workers/file_processing/tasks.py
@@ -224,8 +224,63 @@ def _enhance_batch_with_mrq_flags(
         )
 
 
+class _TerminalExecutionSkip(Exception):
+    """A batch was delivered for an execution already in a terminal state.
+
+    Raised by :func:`_setup_execution_context` when the execution's status is
+    COMPLETED/ERROR/STOPPED. On the PG queue (at-least-once), a batch can be
+    redelivered — or a stale batch consumed — *after* the reaper has recovered
+    the execution (marked it ERROR) and deleted its barrier. Reprocessing would
+    resurrect the execution back to EXECUTING and re-run its files (double LLM /
+    destination write). The batch is skipped instead (UN-3662).
+
+    The guard is gated to the PG path (``is_pg``) so the Celery flow is
+    byte-for-byte unchanged — the resurrection it prevents is PG-specific
+    (Celery lost tasks are never redelivered, so they can't be resurrected).
+    """
+
+    def __init__(self, execution_id: str, status: str | None) -> None:
+        self.execution_id = execution_id
+        self.status = status
+        super().__init__(
+            f"Execution {execution_id} already terminal ({status}); "
+            f"skipping stale/redelivered batch"
+        )
+
+
+def _raise_if_execution_terminal(
+    workflow_execution: dict[str, Any], execution_id: str, *, is_pg: bool
+) -> None:
+    """PG-path-only validate-first guard: refuse to (re)process a batch whose
+    execution is already terminal. See :class:`_TerminalExecutionSkip`.
+
+    No-op on the Celery path (``is_pg=False``) — the resurrection this prevents
+    is PG-specific, so gating keeps the Celery flow byte-for-byte unchanged
+    (UN-3662). A missing/unknown status is treated as non-terminal (fail open —
+    proceed as normal).
+    """
+    if not is_pg:
+        return
+    status = workflow_execution.get("status")
+    if status in ExecutionStatus.terminal_values():
+        raise _TerminalExecutionSkip(execution_id, status)
+
+
+def _terminal_skip_result(batch_data: FileBatchData) -> dict[str, Any]:
+    """Benign zero-work batch result for a batch skipped because its execution
+    is already terminal (UN-3662) — no files processed, all counts zero.
+    """
+    return BatchExecutionResult(
+        total_files=len(batch_data.files),
+        successful_files=0,
+        failed_files=0,
+        execution_time=0.0,
+        organization_id=batch_data.file_data.organization_id,
+    ).to_dict()
+
+
 def _run_batch_stages(
-    file_batch_data: dict[str, Any], celery_task_id: str
+    file_batch_data: dict[str, Any], celery_task_id: str, is_pg: bool = False
 ) -> dict[str, Any]:
     """The actual batch work (validate → setup → pre-create → process → compile).
 
@@ -237,8 +292,21 @@ def _run_batch_stages(
     # Step 1: Validate and parse input data
     batch_data = _validate_and_parse_batch_data(file_batch_data)
 
-    # Step 2: Setup execution context
-    context = _setup_execution_context(batch_data, celery_task_id)
+    # Step 2: Setup execution context. Validate-first (PG path only): a batch
+    # delivered for an already-terminal execution is a stale/redelivered task
+    # arriving after the reaper recovered the execution — skip it rather than
+    # resurrect the execution and re-run its files (UN-3662). The guard is gated
+    # inside _setup_execution_context via is_pg, so on Celery this never fires.
+    try:
+        context = _setup_execution_context(batch_data, celery_task_id, is_pg=is_pg)
+    except _TerminalExecutionSkip as skip:
+        logger.warning(
+            f"[exec:{skip.execution_id}] Skipping batch of "
+            f"{len(batch_data.files)} file(s): execution already terminal "
+            f"({skip.status}) — stale/redelivered after reaper-recovery; not "
+            f"reprocessing (UN-3662)."
+        )
+        return _terminal_skip_result(batch_data)
 
     # Step 3: Handle manual review logic
     # context = _handle_manual_review_logic(context)
@@ -281,13 +349,16 @@ def _process_file_batch_core(
 
     if barrier_context is None:
         # Celery chord path — the chord's .link runs the decrement after this.
-        return _run_batch_stages(file_batch_data, celery_task_id)
+        # is_pg=False disables the UN-3662 terminal guard → Celery flow unchanged.
+        return _run_batch_stages(file_batch_data, celery_task_id, is_pg=False)
 
     # PG fire-and-forget path — claim the batch (idempotent on redelivery), run
     # the stages, then decrement the barrier in-body / self-chain the callback.
+    # is_pg=True enables the UN-3662 terminal guard (skip stale/redelivered
+    # batches for a reaper-recovered execution).
     return run_batch_with_barrier(
         barrier_context,
-        lambda: _run_batch_stages(file_batch_data, celery_task_id),
+        lambda: _run_batch_stages(file_batch_data, celery_task_id, is_pg=True),
     )
 
 
@@ -359,7 +430,7 @@ def _validate_and_parse_batch_data(file_batch_data: dict[str, Any]) -> FileBatch
 
 
 def _setup_execution_context(
-    batch_data: FileBatchData, celery_task_id: str
+    batch_data: FileBatchData, celery_task_id: str, is_pg: bool = False
 ) -> WorkflowContextData:
     """Setup execution context with validation and API client initialization.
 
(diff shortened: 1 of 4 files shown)


muhammad-ali-e (commented):
Automated PR Review Toolkit pass (code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier). Findings below are grouped inline. The count-semantics note on `_terminal_skip_result` refines @greptile-apps's existing P2 rather than repeating it.

workers/file_processing/tasks.py
```diff
@@ -281,13 +349,16 @@ def _process_file_batch_core(
...
         # Celery chord path — the chord's .link runs the decrement after this.
-        return _run_batch_stages(file_batch_data, celery_task_id)
+        # is_pg=False disables the UN-3662 terminal guard → Celery flow unchanged.
+        return _run_batch_stages(file_batch_data, celery_task_id, is_pg=False)
 
     # PG fire-and-forget path — claim the batch (idempotent on redelivery), run
     # the stages, then decrement the barrier in-body / self-chain the callback.
+    # is_pg=True enables the UN-3662 terminal guard (skip stale/redelivered
+    # batches for a reaper-recovered execution).
     return run_batch_with_barrier(
```
muhammad-ali-e, line 370: **[P2 — Silent failure / log noise] The terminal skip is handled one layer too low, so the barrier is still decremented and logs a spurious `ERROR` on every benign skip.**

`_run_batch_stages` swallows `_TerminalExecutionSkip` and returns a normal-looking result (`tasks.py:302-309`). `run_batch_with_barrier` can't tell it was a skip, so it unconditionally calls `_barrier_pg_decrement` here. In the guard's *intended* scenario the reaper has already **deleted the barrier**, so the decrement hits `row is None` and emits:

> `PgBarrier decrement found no row — barrier already torn down…` (`queue_backend/pg_barrier.py`, ERROR level)

So the normal, benign operation of this guard logs an ERROR every time it fires — false Sentry/alert noise that also desensitizes operators to the *same* line when it signals genuine barrier corruption.

Separately, for the **STOPPED** case (user stop) I couldn't find a barrier teardown on the stop path. If the barrier is still live, each skipped batch decrements with a 0/0 result and one of them eventually drives `remaining→0`, firing the aggregating callback with all-zero-work results against an already-terminal execution.

**Suggested fix:** let `_TerminalExecutionSkip` propagate to `run_batch_with_barrier` (or return a sentinel) and short-circuit *before* `_barrier_pg_decrement` — a terminal execution's barrier is by definition already gone, so the decrement is a wasted round-trip that only logs an error. Confirm the STOP path tears the barrier down before relying on the skip for STOPPED.

workers/file_processing/tasks.py
```diff
@@ -224,8 +224,63 @@ def _enhance_batch_with_mrq_flags(
...
+
+
+def _terminal_skip_result(batch_data: FileBatchData) -> dict[str, Any]:
+    """Benign zero-work batch result for a batch skipped because its execution
+    is already terminal (UN-3662) — no files processed, all counts zero.
+    """
+    return BatchExecutionResult(
+        total_files=len(batch_data.files),
+        successful_files=0,
+        failed_files=0,
```
muhammad-ali-e: **[P2 — refining @greptile-apps's P2 on line 279 with new info; not a duplicate]** greptile suggests `total_files=0`. I'd push the other way: **`failed_files=len(batch_data.files)`**.

Sibling commit `68d137a8` ("terminal-failure run reports unaccounted files as failed, not in-progress") established the accounting contract that the UI derives in-progress as `total - successful - failed`. With the current `total=N, successful=0, failed=0`, any consumer that *does* aggregate this result (the STOPPED / barrier-still-live case in the comment above) renders N files as **perpetually in-progress** — exactly the defect `68d137a8` fixed. greptile's `total_files=0` makes the N files vanish entirely (never attempted, never failed).

Reporting `failed_files=N` keeps the invariant `successful + failed == total`, matches `orchestration_utils.py`'s `successful=0, failed=total` convention, and means a stranded batch surfaces as failed rather than silently in-progress. A batch never processed for a terminal execution is not a success.

(If the skip is instead short-circuited before the barrier decrement per the comment above, the result is discarded and this is moot — but the safe value is still `failed=N`, not `0`.)

workers/file_processing/tasks.py
```diff
@@ -411,6 +482,14 @@ def _setup_execution_context(
     execution_context = execution_response.data
     workflow_execution = execution_context.get("execution", {})
 
+    # UN-3662: validate-first terminal guard (PG path only, gated by is_pg). If
+    # the execution is already terminal (COMPLETED/ERROR/STOPPED), this batch is
+    # a stale/redelivered task arriving after the reaper recovered the execution
+    # — reprocessing would resurrect it to EXECUTING and re-run its files. Skip
+    # before we touch status or process anything. On Celery (is_pg=False) this is
+    # a no-op, so the existing Celery flow is unchanged.
+    _raise_if_execution_terminal(workflow_execution, execution_id, is_pg=is_pg)
```
muhammad-ali-e, line 502: **[P3 — residual TOCTOU; wording overstates the guarantee]** This guard reads `status` (line 264) and re-asserts non-terminal here, but the status→`EXECUTING` write happens unconditionally at `tasks.py:508`. If a concurrent reaper transitions the execution to ERROR *after* this check but *before* line 508, the batch still resurrects the execution and re-runs its files. The guard **narrows** the window; it can't close it (read-then-act, not a transactional CAS).

That's inherent to at-least-once + reaper-recovery and is acceptable — but the docstrings say the guard is applied "before we touch status" / that it "prevents" the resurrection, which reads as a hard guarantee. Suggest documenting this as a known residual window, or (stronger) gating the line-508 `EXECUTING` write on a conditional update that only fires when the current status is non-terminal.

workers/file_processing/tasks.py
```diff
@@ -224,8 +224,63 @@ def _enhance_batch_with_mrq_flags(
...
+    execution is already terminal. See :class:`_TerminalExecutionSkip`.
+
+    No-op on the Celery path (``is_pg=False``) — the resurrection this prevents
+    is PG-specific, so gating keeps the Celery flow byte-for-byte unchanged
+    (UN-3662). A missing/unknown status is treated as non-terminal (fail open —
+    proceed as normal).
+    """
+    if not is_pg:
+        return
+    status = workflow_execution.get("status")
```
muhammad-ali-e, line 267: **[P3 — silent fail-open masks a malformed execution-fetch response]** `workflow_execution.get("status")` returns `None` when the execution dict came back empty/malformed, and an unrecognized status string is likewise treated as non-terminal. Fail-open is the right call for a *genuinely active* execution, but a **missing** `status` specifically signals an API-contract regression, and it's swallowed with no log — the guard silently stops protecting against resurrection.

Keep fail-open on the processing decision, but emit a `logger.warning` when `is_pg and not status` (and ideally when the status isn't a recognized `ExecutionStatus` value) so a degraded response is visible instead of silently masked. Note the tests currently *pin* the silent behavior (`None`/`""` params).

workers/file_processing/tasks.py
```diff
@@ -224,8 +224,63 @@ def _enhance_batch_with_mrq_flags(
...
+        successful_files=0,
+        failed_files=0,
+        execution_time=0.0,
+        organization_id=batch_data.file_data.organization_id,
+    ).to_dict()
+
+
 def _run_batch_stages(
-    file_batch_data: dict[str, Any], celery_task_id: str
+    file_batch_data: dict[str, Any], celery_task_id: str, is_pg: bool = False
```
muhammad-ali-e, line 294: **[P3 — `is_pg` default is a foot-gun and names the mechanism, not the invariant]** Two related nits on the new flag threaded through `_run_batch_stages` / `_setup_execution_context`:

1. **Default `is_pg=False`** means a future PG-path caller that forgets to pass `is_pg=True` silently disables the guard with no error. `_raise_if_execution_terminal` already gets this right (keyword-only, no default). Consider making these two helpers keyword-only with no default so PG call sites must state intent.
2. **Naming:** the real precondition is "redelivery-possible transport," not "Postgres." A boolean named for the mechanism is boolean-blind; `redelivery_possible` / `at_least_once` (or a small `Transport` enum) would express *why* the guard runs. It's also a re-derivation of transport that `barrier_context is None` already encodes in `_process_file_batch_core` — two independent transport signals can drift.

workers/file_processing/tasks.py
```diff
@@ -359,7 +430,7 @@ def _validate_and_parse_batch_data(file_batch_data: dict[str, Any]) -> FileBatch
 
 
 def _setup_execution_context(
-    batch_data: FileBatchData, celery_task_id: str
+    batch_data: FileBatchData, celery_task_id: str, is_pg: bool = False
```
muhammad-ali-e, line 444: **[P3 — contract/docstring not updated for the new behavior]** With `is_pg` added, this function now (a) takes an `is_pg` parameter and (b) can raise `_TerminalExecutionSkip`. The docstring below (`Args:` / `Raises:` at lines 437-445) documents neither — it still lists only `batch_data`/`celery_task_id` and `Raises: ValueError`. Add `is_pg` to `Args` and `_TerminalExecutionSkip` to `Raises` so the new part of the contract is visible to callers.

workers/file_processing/tasks.py
```diff
@@ -224,8 +224,63 @@ def _enhance_batch_with_mrq_flags(
...
+    """A batch was delivered for an execution already in a terminal state.
+
+    Raised by :func:`_setup_execution_context` when the execution's status is
+    COMPLETED/ERROR/STOPPED. On the PG queue (at-least-once), a batch can be
+    redelivered — or a stale batch consumed — *after* the reaper has recovered
+    the execution (marked it ERROR) and deleted its barrier. Reprocessing would
+    resurrect the execution back to EXECUTING and re-run its files (double LLM /
+    destination write). The batch is skipped instead (UN-3662).
+
+    The guard is gated to the PG path (``is_pg``) so the Celery flow is
```
muhammad-ali-e, line 240: **[P3 — comment accuracy]** Three small precision issues in this docstring block:

- **"byte-for-byte unchanged"** (repeated at lines 237-238, 258, and in the test at `test_terminal_execution_guard.py:51`): the Celery path is *not* literally byte-for-byte — `_run_batch_stages`/`_setup_execution_context` gained an `is_pg` param and an added `_raise_if_execution_terminal` call that early-returns. The *observable behavior* is unchanged; "behaviorally unchanged (the guard is a no-op when `is_pg=False`)" is the accurate claim.
- **"Celery lost tasks are never redelivered"**: this is broker/`acks_late`/`task_reject_on_worker_lost`-config-dependent, so "never" can rot. More importantly it understates the real reason the guard is PG-only — the **reaper** marking an in-flight execution terminal and reclaiming its dedup markers is the PG-specific mechanism; the Celery chord has no equivalent.
- **line 230**: "Raised by `_setup_execution_context`" — the `raise` actually lives in `_raise_if_execution_terminal`; `_setup_execution_context` only propagates it.

workers/file_processing/tasks.py
```diff
@@ -237,8 +292,21 @@ def _run_batch_stages(
     # Step 1: Validate and parse input data
     batch_data = _validate_and_parse_batch_data(file_batch_data)
 
-    # Step 2: Setup execution context
-    context = _setup_execution_context(batch_data, celery_task_id)
+    # Step 2: Setup execution context. Validate-first (PG path only): a batch
```
muhammad-ali-e, line 306: **[P3 — comment duplication / rot risk]** The UN-3662 rationale ("stale/redelivered batch after reaper recovery → would resurrect to EXECUTING and re-run files → skip; no-op on Celery") is fully restated ~5 times: the class docstring (227-239), `_raise_if_execution_terminal` (254-260), this block (295-299), the block in `_setup_execution_context` (485-490), and the `_process_file_batch_core` comments (352, 355-357). Any change to the reasoning must be edited in five places or they silently diverge.

Keep the full explanation in the class docstring (the canonical home) and collapse the two multi-line inline blocks here and at 485-490 to a one-liner that defers to it, e.g. `# UN-3662: PG-only terminal guard — skip stale/redelivered batch (see _TerminalExecutionSkip).` The terse call-site comments at 352/355-357 are already a good model — leave those.

workers/tests/test_terminal_execution_guard.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def test_terminal_skip_result_is_zero_work():
+    result = _terminal_skip_result(_fake_batch_data(n_files=4, org="org-9"))
+    assert result["total_files"] == 4
+    assert result["successful_files"] == 0
+    assert result["failed_files"] == 0
+
+
+def test_run_batch_stages_skips_terminal_without_processing():
```
muhammad-ali-e, line 75: **[P3 — test coverage gap: the guard's real invocation site is never executed]** Every test mocks `_setup_execution_context` (here and line 104), so the actual production wiring — `tasks.py:491` `_raise_if_execution_terminal(workflow_execution, execution_id, is_pg=is_pg)`, placed after the execution fetch and *before* the status write — is never run. Two regressions would pass green: the `is_pg` forwarding chain (`_run_batch_stages`→`_setup_execution_context`→`_raise_if_execution_terminal`), and the guard's placement before the `EXECUTING` write.

Suggested additions:
1. Drive `_setup_execution_context` with a mocked API client returning `execution={"status":"ERROR"}`; assert it raises `_TerminalExecutionSkip` when `is_pg=True` and proceeds when `is_pg=False`. (Catches a hardcoded flag or a mis-placed guard.)
2. An end-to-end `run_batch_with_barrier(ctx, lambda: _terminal_skip_result(bd))` test asserting the barrier still decrements exactly once (the no-hang invariant this PR exists to protect) — natural home is `tests/test_pg_barrier.py`.
3. A one-line `json.dumps(result)` assertion in `test_terminal_skip_result_is_zero_work` — `_barrier_pg_decrement` does `json.dumps(result)` with no `default=str`, so a `to_dict()` boundary regression would strand the barrier.

muhammad-ali-e:
Addressed all three review findings in `c1f6acc26`:

**1. [P2 — architectural] Spurious `ERROR` on every benign skip (barrier decrement on a torn-down barrier).** Fixed. The skip result now carries a `SKIPPED_TERMINAL_EXECUTION_KEY` marker, and `run_batch_with_barrier` detects it and **bypasses `_barrier_pg_decrement` entirely** (and does not abort). In the guard's intended scenario the reaper has already deleted the barrier, so we no longer make the gone-barrier round-trip that logged `PgBarrier decrement found no row — barrier already torn down`. This also removes the STOPPED (barrier-still-live) risk of driving `remaining→0` and firing the aggregating callback with zero-work results — we simply never decrement on a skip.

**2. + 3. [P2 — count semantics] `_terminal_skip_result` counts.** Resolved by taking @muhammad-ali-e's refinement over the `total_files=0` suggestion: the result now reports **`failed_files=N`** (total=N, successful=0). This honours the `68d137a8` accounting contract (unaccounted files count as failed, not in-progress) — in-progress `(total − successful − failed)` is 0, so the files neither show as perpetually in-progress nor silently vanish.

**On the STOPPED barrier-teardown you flagged:** I couldn't find a dedicated PG-barrier teardown on the stop path either. With this fix the skip no longer fires a zero-work callback for a STOPPED-with-live-barrier execution — the barrier instead strands and is reclaimed by the reaper (the existing cleanup net). A dedicated stop-path teardown (so STOPPED barriers are cleaned promptly rather than at reaper time) is worth a small follow-up; out of scope here.

**Tests added/updated:** skip result asserts `failed=N` + marker; two new tests that `run_batch_with_barrier` bypasses the decrement/abort on a marked result and still decrements on a normal one. Full targeted suite green (51 passed).

PR #2138, opened 2026-07-01
UN-3663 [FIX] PG queue — skip duplicate destination write on re-run (Gap A) + UN-3662 review nits

muhammad-ali-e:
Combined PR (onto the integration branch): the **UN-3663** exactly-once hardening plus the deferred **UN-3662** review nits. Every behavioral change is gated to the PG transport — the Celery path is byte-unchanged, so there's no regression to the existing staging/main flow.

## UN-3663 — duplicate destination write on re-run (Gap A)
On the PG (at-least-once) transport a batch can be re-run after a crash / reaper-recovery, and such a re-run **bypasses discovery's FileHistory filter** — so if the file's destination write already completed in a prior run, the write is repeated (duplicate).

- **Destination re-check (PG only):** before `destination.handle_output()`, re-check FileHistory by hash; if a **COMPLETED** record exists, skip the write and reuse the existing duplicate-skip path. Gated on `is_pg_transport(file_data.transport)` → **Celery destination path byte-unchanged**; **fail-open** so any lookup error never blocks a legitimate write. No DB migration.
- **Active-file TTL cap** raised 7200 → 9000 (≥ the PG stuck-batch reaper timeout) so an operator can configure the marker to outlive a stalled-but-not-yet-reaped batch. Default (300s) unchanged; only the allowed maximum widens → can't regress either transport.

**Deferred to their own follow-ups** (can't be cleanly PG-gated / out of scope here): making the backend FileHistory-create + file-status write atomic (shared endpoint — would touch the Celery path); a post-reaper reconciliation sweep for the write-then-crash-before-record window.

## UN-3662 review nits (follow-up to the merged terminal-guard change)
No behavior change; polish only:
- `is_pg` is now **keyword-only, no default** on the batch-stage helpers, so a PG call site can't silently disable the guard.
- **Fail-open warning** when a PG execution-fetch returns no status (was silently masked).
- Docstrings updated (Args/Raises); duplicated rationale collapsed to the class docstring; wording corrected (behaviorally-unchanged, accurate raise site); documented the residual read-then-act window (guard narrows, not a transactional guarantee).
- Stripped internal tracker/commit refs from code comments (public repo).
- New test drives the guard at its real call site (asserts it raises **and** the EXECUTING write never happens) + a fail-open-warning test.

## Tests
New `test_pg_duplicate_write_guard.py` (PG skips on COMPLETED FileHistory hit; proceeds otherwise; no-op on Celery; fail-open on lookup error; hash-only match for API; TTL cap honors 9000) + the terminal-guard tests. Full worker suite shows no new failures vs baseline (pre-existing suite pollution + one hanging consumer test are unrelated — verified identical with these changes stashed).

### DIFF
--- a/workers/file_processing/tasks.py
+++ b/workers/file_processing/tasks.py
@@ -230,16 +230,23 @@ def _enhance_batch_with_mrq_flags(
 class _TerminalExecutionSkip(Exception):
     """A batch was delivered for an execution already in a terminal state.
 
-    Raised by :func:`_setup_execution_context` when the execution's status is
+    Raised by :func:`_raise_if_execution_terminal` (called from
+    :func:`_setup_execution_context`) when the execution's status is
     COMPLETED/ERROR/STOPPED. On the PG queue (at-least-once), a batch can be
     redelivered — or a stale batch consumed — *after* the reaper has recovered
     the execution (marked it ERROR) and deleted its barrier. Reprocessing would
     resurrect the execution back to EXECUTING and re-run its files (double LLM /
-    destination write). The batch is skipped instead (UN-3662).
+    destination write). The batch is skipped instead.
 
-    The guard is gated to the PG path (``is_pg``) so the Celery flow is
-    byte-for-byte unchanged — the resurrection it prevents is PG-specific
-    (Celery lost tasks are never redelivered, so they can't be resurrected).
+    Gated to the PG path (``is_pg``): a no-op on Celery, so the existing Celery
+    flow is behaviorally unchanged. It's PG-specific because the **reaper** that
+    marks an in-flight execution terminal (and tears its barrier down) is a
+    PG-only mechanism — the Celery chord has no equivalent.
+
+    This is a validate-first *narrowing*, not a transactional guarantee: the
+    status read and the later ``EXECUTING`` write are not atomic, so a reaper
+    transition landing in that window can still resurrect the execution. The
+    reaper's terminal file-cascade is the backstop for that residual case.
     """
 
     def __init__(self, execution_id: str, status: str | None) -> None:
@@ -258,25 +265,36 @@ def _raise_if_execution_terminal(
     execution is already terminal. See :class:`_TerminalExecutionSkip`.
 
     No-op on the Celery path (``is_pg=False``) — the resurrection this prevents
-    is PG-specific, so gating keeps the Celery flow byte-for-byte unchanged
-    (UN-3662). A missing/unknown status is treated as non-terminal (fail open —
-    proceed as normal).
+    is PG-specific, so gating leaves the Celery flow behaviorally unchanged. A
+    missing/unrecognized status is treated as non-terminal (fail open — proceed
+    as normal), but a *missing* status on the PG path signals a degraded
+    execution-fetch response and is logged so it isn't silently masked.
     """
     if not is_pg:
         return
     status = workflow_execution.get("status")
+    if not status:
+        # Fail open (proceed), but surface it — a missing status is an
+        # API-contract regression, not a normal active execution.
+        logger.warning(
+            f"[exec:{execution_id}] terminal guard: execution-fetch returned no "
+            f"status; proceeding (fail-open) but the response looks degraded."
+        )
+        return
     if status in ExecutionStatus.terminal_values():
         raise _TerminalExecutionSkip(execution_id, status)
 
 
 def _terminal_skip_result(batch_data: FileBatchData) -> dict[str, Any]:
-    """Batch result for a batch skipped because its execution is already terminal
-    (UN-3662). The files were never attempted; per the accounting contract from
-    68d137a8 (unaccounted files count as failed, not in-progress) they are
-    reported as failed, so a consumer that ever aggregates this result does not
-    render them as perpetually in-progress. The ``SKIPPED_TERMINAL_EXECUTION_KEY``
-    marker tells ``run_batch_with_barrier`` to bypass the barrier decrement (the
-    reaper has by definition already torn the barrier down).
+    """Batch result for a batch skipped because its execution is already terminal.
+
+    The files were never attempted; they are counted as failed (not left
+    unaccounted). The UI derives in-progress as ``total - successful - failed``,
+    so reporting ``failed=total`` keeps a consumer that ever aggregates this
+    result from rendering the files as perpetually in-progress. The
+    ``SKIPPED_TERMINAL_EXECUTION_KEY`` marker tells ``run_batch_with_barrier`` to
+    bypass the barrier decrement (a terminal execution's barrier is by definition
+    already torn down).
     """
     n_files = len(batch_data.files)
     result = BatchExecutionResult(
@@ -291,7 +309,7 @@ def _terminal_skip_result(batch_data: FileBatchData) -> dict[str, Any]:
 
 
 def _run_batch_stages(
-    file_batch_data: dict[str, Any], celery_task_id: str, is_pg: bool = False
+    file_batch_data: dict[str, Any], celery_task_id: str, *, is_pg: bool
 ) -> dict[str, Any]:
     """The actual batch work (validate → setup → pre-create → process → compile).
 
@@ -299,15 +317,16 @@ def _run_batch_stages(
     fire-and-forget path. The task instance isn't needed here — its only use
     (deriving ``celery_task_id``) happens in the caller. Returns the
     JSON-serialisable batch result.
+
+    ``is_pg`` (keyword-only, no default so a PG call site can't silently forget
+    it) forwards to the terminal guard; see :class:`_TerminalExecutionSkip`.
     """
     # Step 1: Validate and parse input data
     batch_data = _validate_and_parse_batch_data(file_batch_data)
 
-    # Step 2: Setup execution context. Validate-first (PG path only): a batch
-    # delivered for an already-terminal execution is a stale/redelivered task
-    # arriving after the reaper recovered the execution — skip it rather than
-    # resurrect the execution and re-run its files (UN-3662). The guard is gated
-    # inside _setup_execution_context via is_pg, so on Celery this never fires.
+    # Step 2: Setup execution context. The terminal guard inside it may skip a
+    # stale/redelivered batch for an already-terminal execution — see
+    # :class:`_TerminalExecutionSkip`.
     try:
         context = _setup_execution_context(batch_data, celery_task_id, is_pg=is_pg)
     except _TerminalExecutionSkip as skip:
@@ -315,7 +334,7 @@ def _run_batch_stages(
             f"[exec:{skip.execution_id}] Skipping batch of "
             f"{len(batch_data.files)} file(s): execution already terminal "
             f"({skip.status}) — stale/redelivered after reaper-recovery; not "
-            f"reprocessing (UN-3662)."
+            f"reprocessing."
         )
         return _terminal_skip_result(batch_data)
 
@@ -360,13 +379,13 @@ def _process_file_batch_core(
 
     if barrier_context is None:
         # Celery chord path — the chord's .link runs the decrement after this.
-        # is_pg=False disables the UN-3662 terminal guard → Celery flow unchanged.
+        # is_pg=False disables the terminal guard → Celery flow unchanged.
         return _run_batch_stages(file_batch_data, celery_task_id, is_pg=False)
 
     # PG fire-and-forget path — claim the batch (idempotent on redelivery), run
     # the stages, then decrement the barrier in-body / self-chain the callback.
-    # is_pg=True enables the UN-3662 terminal guard (skip stale/redelivered
-    # batches for a reaper-recovered execution).
+    # is_pg=True enables the terminal guard (skip a stale/redelivered batch for a
+    # reaper-recovered execution).
     return run_batch_with_barrier(
         barrier_context,
         lambda: _run_batch_stages(file_batch_data, celery_task_id, is_pg=True),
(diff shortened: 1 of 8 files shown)


workers/shared/workflow/execution/service.py
```diff
@@ -1361,6 +1362,50 @@ def _resolve_connector_config(
...
+
+        On the PG (at-least-once) transport a batch can be re-run after a crash
+        or reaper-recovery, and such a re-run bypasses discovery's FileHistory
+        filter — so re-check by hash at the write boundary. Returns True only if
+        a COMPLETED FileHistory record already exists for this file.
+
+        Always False on the Celery transport (so that path is byte-unchanged),
+        and fail-open on any lookup error (never block a legitimate write).
+        """
+        if not is_pg_transport(getattr(file_data, "transport", None)):
```
muhammad-ali-e: **🔴 CRITICAL — this guard is dead code; the PR's headline fix never fires.**

`file_data` here is a `WorkerFileData` (via `file_processing_context.file_data`), and `WorkerFileData` (unstract/core `data_models.py:1613`) has **no `transport` field** — and `from_dict` drops unknown keys. So `getattr(file_data, "transport", None)` is always `None`, `is_pg_transport(None)` is always `False`, and `_pg_destination_already_written` returns `False` on every real call. The duplicate-destination-write on PG re-run that this PR exists to close is **not actually closed**.

Verified: `WorkerFileData` fields carry no transport; nothing sets `file_data.transport` in `workers/`. The `getattr(..., None)` default is exactly what hid the contract violation — a typed `file_data: WorkerFileData` with direct access would have been flagged statically.

**Fix:** source the transport from where it is authoritative (thread `is_pg`/`WorkflowContextData.transport` down through `processor.execute_workflow_for_file` → `_handle_destination_processing` → this method, the same way `is_pg` is already threaded for the terminal guard), then type the parameter and drop the `getattr` default. Note: transport is *not* currently available at the service layer at all, so this needs real plumbing, not a one-liner.

_(Corroborated independently by code-reviewer, silent-failure-hunter, type-design-analyzer, and pr-test-analyzer.)_

muhammad-ali-e (commented):

Ran six specialist agents (code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier) over the 6 in-scope files. Overall the terminal-guard hardening and comment self-containment are solid, and several choices are genuinely good (keyword-only `is_pg` with no default; skipped files counted as `failed` not unaccounted; the honest "validate-first narrowing, not a transactional guarantee" docstring).

**However, four of the six agents independently found — and I verified — that the PR's headline change (the Gap A duplicate-write guard) is dead code in production.** See the inline comment on `service.py:1384`. That is the one blocking item; everything else is hardening/polish.

Findings are posted inline below, prioritised Critical → Low.

workers/tests/test_pg_duplicate_write_guard.py
```diff
@@ -0,0 +1,130 @@
...
+    assert already is False
+    api.get_file_history_by_cache_key.assert_not_called()  # no-op on Celery
+
+
+def test_pg_skips_when_completed_history_exists():
+    api = mock.Mock()
+    api.get_file_history_by_cache_key.return_value = _history(status="COMPLETED")
+    already = _svc(api)._pg_destination_already_written(
+        file_hash=_file_hash(),
+        file_data=mock.Mock(transport=PG),
```
muhammad-ali-e: **🟠 HIGH — these tests mask the Critical bug above.** Every case builds `file_data=mock.Mock(transport=PG|CELERY)`. A spec-less `Mock` fabricates *any* attribute, so `.transport` "exists" only in the test — the real `WorkerFileData` has no such field, so the suite is green while the feature is inert.

**Fix:** exercise the guard against a real `WorkerFileData` (or `mock.Mock(spec=WorkerFileData)`); with a spec the current code surfaces the missing `transport` immediately. Also add a wiring test for `_handle_destination_processing` (guard→`True` ⇒ `destination.handle_output` not called and `processed=False` returned; guard→`False` ⇒ it *is* called) — the terminal guard got a real-call-site test, this one has none.

workers/shared/workflow/execution/service.py
```diff
@@ -1361,6 +1362,50 @@ def _resolve_connector_config(
...
+            return False
+        # API executions use unique per-execution paths, so match on hash only.
+        lookup_file_path = None if is_api else getattr(file_hash, "file_path", None)
+        try:
+            history_result = self.api_client.get_file_history_by_cache_key(
+                cache_key=cache_key,
+                workflow_id=workflow_id,
+                file_path=lookup_file_path,
+            )
+        except Exception as exc:
```
muhammad-ali-e: **🟠 HIGH — broad `except Exception` + `warning`-only can silently disable the guard.** Fail-open is the right *direction* (failing closed would skip a legitimate first write = silent data loss). The problem is the catch doesn't distinguish a transient API/DB blip from a programmer bug (`AttributeError`, changed `get_file_history_by_cache_key` signature, deserialization error). If such a bug lands, this catches on *every* file, logs at `warning` (not Sentry-visible), returns `False`, and the dedup protection becomes a permanent no-op with no page.

**Fix:** narrow the catch to expected transient types (keep `warning` + proceed); let unexpected types `logger.error` (or re-raise) so a regression is loud. Consider a metric/counter on skip-vs-proceed so a permanently-dead guard is observable.

workers/shared/workflow/execution/service.py
```diff
@@ -1471,20 +1516,43 @@ def _handle_destination_processing(
...
-                    is_success=is_success,
+                # Gap A guard (PG transport only): a re-run after a crash /
+                # reaper-recovery bypasses discovery's FileHistory filter, so
+                # re-check by hash at the write boundary. If this file's
+                # destination write already completed in a prior run (a COMPLETED
+                # FileHistory record exists), skip the write and reuse the
+                # duplicate-skip path below, avoiding a duplicate destination
+                # write. A no-op on the Celery transport, so that path is
+                # unchanged; fail-open so a lookup error never blocks a write.
+                if self._pg_destination_already_written(
```
muhammad-ali-e, line 1537: **🟠 HIGH — guard ignores `use_file_history`, suppressing legitimate re-writes for ETL.** The discovery filter this mirrors (`FileHistoryFilter.apply`) is gated on `use_file_history`. But `_should_create_file_history` creates a COMPLETED record for ETL/TASK workflows even when `use_file_history=False` (service.py ~1694). So an ETL workflow with `use_file_history=False` on PG — whose contract is "rewrite every run" — will have its destination write silently skipped because a prior run's COMPLETED FileHistory exists, diverging from Celery and dropping expected output. (Becomes live once the Critical is fixed.)

`_handle_destination_processing` already has `use_file_history` in scope. **Fix:** return `False` from the guard when `use_file_history` is `False`, and add a regression test.

workers/file_processing/tasks.py
```diff
@@ -258,25 +265,36 @@ def _raise_if_execution_terminal(
...
-    proceed as normal).
+    is PG-specific, so gating leaves the Celery flow behaviorally unchanged. A
+    missing/unrecognized status is treated as non-terminal (fail open — proceed
+    as normal), but a *missing* status on the PG path signals a degraded
+    execution-fetch response and is logged so it isn't silently masked.
     """
     if not is_pg:
         return
     status = workflow_execution.get("status")
+    if not status:
```
muhammad-ali-e, line 278: **🟡 MEDIUM — docstring over-claims; an *unrecognized* (non-empty) status still passes silently.** This block correctly warns on a *missing* status, but the docstring says "missing/**unrecognized** status … is logged so it isn't silently masked." A non-empty value not in `ExecutionStatus.terminal_values()` falls straight through at line 284 with no log. If a new terminal-ish state or an enum typo ever ships, a stale/redelivered PG batch for it won't be skipped — the exact resurrection this guard prevents — with zero trace.

**Fix:** add a branch that `logger.warning`s when `status` is non-empty but unknown to `ExecutionStatus`, before treating it as non-terminal — matching the docstring's claim.

workers/shared/workflow/execution/active_file_manager.py
```diff
@@ -20,7 +20,11 @@
 
 # Constants for cache configuration
 DEFAULT_ACTIVE_FILE_CACHE_TTL = 300  # 5 minutes
-MAX_ACTIVE_FILE_CACHE_TTL = 7200  # 2 hours maximum
+# Cap the active-file marker TTL at (>=) the PG stuck-batch reaper timeout so a
+# marker can be configured to survive a stalled-but-not-yet-reaped batch — below
+# it there is a window where the marker has expired but the reaper has not yet
+# marked the execution terminal. Default stays 300s; operators opt into longer.
+MAX_ACTIVE_FILE_CACHE_TTL = 9000  # 2.5 hours maximum
```
muhammad-ali-e, line 30: **🟡 MEDIUM — cap is pinned to the reaper *default*, not its configurable max; plus silent clamping.**

1. `9000` matches `_DEFAULT_BARRIER_STUCK_TIMEOUT_SECONDS`, but that reaper timeout is env-tunable via `WORKER_PG_BATCH_STUCK_TIMEOUT_SECONDS` (up to 10800/3h per its comments). If an operator raises the reaper timeout above 9000, the "marker outlives a stalled-but-not-yet-reaped batch" invariant this comment asserts silently breaks. Prefer deriving the cap from the same source (`max(9000, get_barrier_stuck_timeout_seconds())`) or add a test asserting `MAX_ACTIVE_FILE_CACHE_TTL >= _DEFAULT_BARRIER_STUCK_TIMEOUT_SECONDS`.
2. `get_active_file_cache_ttl()` (below) silently clamps out-of-range values and silently defaults on parse error. An operator who sets `ACTIVE_FILE_CACHE_TTL` above 9000 (or a typo) gets a shorter effective TTL than intended, re-opening the very window this raise was meant to close — `logger.warning` on clamp/parse-fail.
3. Nit: the inline comment on the return (line 41) still reads "Between 1 minute and 2 hours" — now 2.5h.

workers/shared/workflow/execution/service.py
```diff
@@ -1471,20 +1516,43 @@ def _handle_destination_processing(
...
-                    execution_error=execution_error,
-                )
+                    is_api=destination.is_api,
+                ):
+                    logger.info(
+                        f"[exec:{execution_id}] Skipping destination write for "
+                        f"'{file_hash.file_name}' — already completed in a prior "
+                        f"run (FileHistory hit); avoiding a duplicate write."
+                    )
+                    handle_output_result = None
```
muhammad-ali-e: **🟢 LOW — Gap A skip borrows the "race condition, not an error" log path.** Setting `handle_output_result = None` funnels into the shared `if handle_output_result is None:` branch below, which logs `DUPLICATE SKIP … internal race condition, not an error.` For a deliberate re-run dedup that message is misleading, and an operator sees two overlapping lines (the clear one at 1533 + the generic race one). Consider returning the `processed=False` result directly or a distinct flag so the two skip reasons stay distinguishable in logs.

workers/shared/workflow/execution/service.py
```diff
@@ -1361,6 +1362,50 @@ def _resolve_connector_config(
...
 
+    def _pg_destination_already_written(
+        self,
+        *,
+        file_hash,
+        file_data,
+        workflow_id: str,
+        is_api: bool,
+    ) -> bool:
+        """PG-only Gap A guard: True if this file's destination write already
```
muhammad-ali-e: **🟢 LOW — comment/type polish (comment-analyzer + type-design):**
- "Gap A" appears nowhere else in the codebase (no def/doc/ticket) — an unresolvable tag, and inconsistent with this PR's own stripping of `UN-3662`. Either spell it out ("the re-run duplicate-write gap") or drop the label.
- "re-check by hash" understates the lookup: for non-API executions it also passes `file_path` (line 1390). Suggest "by hash, plus path for non-API".
- Params `file_hash` / `file_data` are untyped and read via `getattr`-with-default throughout; the real objects are `FileHashData` / `WorkerFileData` with all fields guaranteed. Annotating them and using direct access removes the concealment that allowed the Critical bug.
- The 8-line call-site comment at 1519 largely duplicates this docstring — trim it to the one local fact (why `None` is assigned) to avoid drift.

muhammad-ali-e:
Addressed all review findings in `a1c5a6e76`.

**🔴 CRITICAL — guard was dead code.** You were right: `file_data` is a `WorkerFileData` with no `transport` field, so `getattr(file_data, 'transport', None)` was always `None` → the guard never fired, and the spec-less mock hid it. Fixed by threading the **authoritative** transport: set from `is_pg` onto `WorkflowContextData.transport`, through `_process_file → FileProcessor.process_file → FileProcessingContext`, and read in `_handle_destination_processing`. The guard now takes a **typed `transport` param** (no `getattr` default).

**🟠 HIGH — tests masked it.** Guard tests now pass a real `transport` string and use `spec=FileHashData`; added a **real-call-site wiring test** that drives `_handle_destination_processing` and asserts it forwards the context transport to the guard and does **not** call `destination.handle_output` on a hit (the exact regression the dead code caused).

**🟠 HIGH — broad except.** Fail-open now logs at **error** (Sentry-visible, `exc_info=True`) so a persistent lookup regression is loud instead of a silent permanent no-op.

**🟠 HIGH — `use_file_history`.** The guard now returns `False` when `use_file_history` is off (ETL/TASK 'rewrite every run' contract), matching the discovery filter. Added a regression test.

**🟡 MED — unknown status.** `_raise_if_execution_terminal` now also warns on an *unrecognized non-empty* status, matching its docstring.

**🟡 MED — TTL cap.** `get_active_file_cache_ttl` now warns on clamp/parse-fail (a silently shortened TTL reopens the window); fixed the stale '2 hours' comment; noted the cap tracks the reaper **default** and must be raised alongside it. Added a clamp-warning test.

**🟢 LOW.** Deliberate re-run skip now returns a distinct `processed=False` result instead of borrowing the 'internal race' `None`-path log; stripped the unresolvable `Gap A` tag; typed `file_hash`/params (direct access, no concealing `getattr`).

28 guard tests green; broader suite shows no new failures vs baseline (the pre-existing `test_sanity_phase3` pollution is unrelated — same with changes stashed).

PR #2139, opened 2026-07-02
UN-3667 [TEST] PG queue — make workers suite deterministic + wire into CI (UN-3666)

muhammad-ali-e:
## What & why

The workers test suite has **never run in CI**: the rig's `unit-workers` group filtered `-m "unit"`, but no workers test carries that marker — so zero tests were collected and exit-5 mapped to success. ~1,100 tests were silently not running. Turning the lane on surfaced a **hang** plus **~28 order-dependent failures**.

This PR makes the suite deterministic in a single process and wires it into the rig. **Tests/CI only — no runtime or production code changes, no Celery-path impact.**

## Changes

**Determinism (`workers/tests/conftest.py`)**
- Reset `os.environ` to the `.env.test` baseline around every test — kills the ambient dev-`.env` bleed (green in CI, red locally) and bare-env-write leaks.
- Reset `queue_backend` thread-locals + one-shot log-once flags between tests.
- `isolated_celery_registry` fixture — a genuinely-empty Celery app for the empty-registry guard tests. Fixes the **hang**: Celery's process-global `@shared_task` finalizer backlog injects the worker's tasks into every new app, defeating the guard and sending `run()` into an infinite poll loop.
- Collection hook auto-marks real-Postgres tests `integration`; the unit lane runs `-m "not integration"`.

**Isolation fixes**
- Empty-app consumer tests use `set_as_current=False` (stop `current_app` hijack).
- `test_legacy_executor_scaffold`: snapshot/restore `ExecutorRegistry` (was wiping the global registry on teardown); restore the exceptions-module namespace after `importlib.reload` (was rebinding exception classes and breaking `pytest.raises` identity suite-wide).
- Barrier protocol test binds `AsyncResult` to a backend-less probe app (was connecting to a polluted PG result backend).

**Drifted tests** (production refactored, tests never updated because they never ran)
- `ExecutionDispatcher` → `get_executor_dispatcher` (sanity phase3/phase5); new `execution_id` arg (phase6h); `lookup-enrichment` plugin (phase6c); `flush_pending_usage` stub (phase2); shim-kwargs subset (phase1); PG-barrier `__all__` exports (queue_backend_seam).

**CI wiring**
- `groups.yaml`: `unit-workers` → `-m "not integration"`; new `integration-workers` lane (requires postgres; `optional` until CI provisions a migrated DB and sets `REQUIRE_PG_TESTS=1`).
- `rig/cli.py`: exit-5 (no tests collected) now **fails** a required group — prevents this class of silent regression.
- `pyproject.toml`: `--timeout=120` hang safety net + `pytest-timeout` dep.

## Results
- **Unit lane: 1133 passed, 0 failed, deterministic** (was 28 failing + a hang) — verified across repeated runs.
- **Integration lane: 103/104** — the one failure is a pre-existing real-PG concurrency test, now correctly isolated to the integration lane (passes alone).

## Follow-ups (not blockers)
- The flaky integration concurrency test needs DB-state isolation between integration tests.
- Activating the integration lane for real needs the `pg_queue` migration + `REQUIRE_PG_TESTS=1` wired into CI.

### DIFF
--- a/tests/groups.yaml
+++ b/tests/groups.yaml
@@ -70,9 +70,32 @@ groups:
     tier: unit
     workdir: workers
     paths: [tests, shared/tests]
-    markers: "unit"
+    # Run everything that isn't a real-Postgres test. The workers suite doesn't
+    # tag its unit tests with an explicit `unit` marker (there are ~1100 of
+    # them), so select by exclusion: `conftest.py` marks the real-DB tests
+    # `integration`, and this lane runs the rest. (`markers: "unit"` here
+    # previously matched *zero* tests — the suite silently never ran in CI.)
+    markers: "not integration"
     uv_sync_group: test
     coverage_source: shared
+
+  integration-workers:
+    tier: integration
+    workdir: workers
+    paths: [tests, shared/tests]
+    # Real-Postgres tests: barrier-decrement races, reaper, batch dedup, leader
+    # election, result backend. `conftest.py` marks these `integration` (any
+    # test using a real-connection fixture). They skip when Postgres is
+    # unreachable or the pg_queue schema isn't migrated.
+    markers: "integration"
+    uv_sync_group: test
+    coverage_source: shared
+    requires_services: [postgres]
+    # Optional until CI provisions a *migrated* database and sets
+    # REQUIRE_PG_TESTS=1 — that env flips the graceful skips into hard failures
+    # (see conftest `_skip_or_fail_no_pg`) so the lane can't pass having run
+    # nothing. Until then it runs-and-skips against a bare Postgres.
+    optional: true
 
   unit-backend:
     tier: unit
--- a/workers/tests/conftest.py
+++ b/workers/tests/conftest.py
@@ -58,7 +58,7 @@ def pytest_collection_modifyitems(config, items):
             item.add_marker(pytest.mark.integration)
 
 
-@pytest.hookimpl(hookwrapper=True)
+@pytest.hookimpl(wrapper=True)
 def pytest_runtest_makereport(item, call):
     """Under ``REQUIRE_PG_TESTS``, a *skipped* integration test is a failure.
 
@@ -71,13 +71,14 @@ def pytest_runtest_makereport(item, call):
     turned into a failure. This closes the "green having exercised none of them"
     gap for the barrier / leader-election / reaper suites regardless of skip
     mechanism.
+
+    Uses the modern ``wrapper=True`` hook (pytest>=8) — ``yield`` returns the
+    report directly, and the wrapper returns it back.
     """
-    outcome = yield
-    if not os.getenv("REQUIRE_PG_TESTS"):
-        return
-    report = outcome.get_result()
+    report = yield
     if (
-        report.when == "setup"
+        os.getenv("REQUIRE_PG_TESTS")
+        and report.when == "setup"
         and report.skipped
         and item.get_closest_marker("integration") is not None
     ):
@@ -87,6 +88,7 @@ def pytest_runtest_makereport(item, call):
             f"(Postgres unreachable/unmigrated): {item.nodeid}. The integration "
             f"lane must exercise the real-Postgres paths, not skip them."
         )
+    return report
 
 
 # --- Isolation: keep the suite deterministic in a single process ---
--- a/workers/tests/conftest.py
+++ b/workers/tests/conftest.py
@@ -6,6 +6,8 @@ shared/constants/api_endpoints.py raises ValueError at import
 time if INTERNAL_API_BASE_URL is not set.
 """
 
+import contextlib
+import importlib
 import os
 from pathlib import Path
 
@@ -54,6 +56,37 @@ def pytest_collection_modifyitems(config, items):
     for item in items:
         if pg_fixtures & set(getattr(item, "fixturenames", ())):
             item.add_marker(pytest.mark.integration)
+
+
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    """Under ``REQUIRE_PG_TESTS``, a *skipped* integration test is a failure.
+
+    The integration lane sets ``REQUIRE_PG_TESTS`` precisely so it can't pass
+    having run nothing. Real-Postgres fixtures skip when the DB is unreachable or
+    unmigrated, and not all of them route through ``_skip_or_fail_no_pg`` (some
+    call ``pytest.skip`` directly). Rather than depend on every fixture — present
+    and future — using the right helper, enforce the guarantee centrally: any
+    ``integration``-marked test that skips during setup while the flag is set is
+    turned into a failure. This closes the "green having exercised none of them"
+    gap for the barrier / leader-election / reaper suites regardless of skip
+    mechanism.
+    """
+    outcome = yield
+    if not os.getenv("REQUIRE_PG_TESTS"):
+        return
+    report = outcome.get_result()
+    if (
+        report.when == "setup"
+        and report.skipped
+        and item.get_closest_marker("integration") is not None
+    ):
+        report.outcome = "failed"
+        report.longrepr = (
+            f"REQUIRE_PG_TESTS is set but integration test skipped "
+            f"(Postgres unreachable/unmigrated): {item.nodeid}. The integration "
+            f"lane must exercise the real-Postgres paths, not skip them."
+        )
 
 
 # --- Isolation: keep the suite deterministic in a single process ---
@@ -166,8 +199,10 @@ def isolated_celery_registry():
     enters its poll loop, and the test hangs).
 
     This fixture snapshots and clears that backlog for the test's duration, so a
-    ``Celery(...)`` created inside the test starts with only ``celery.*``
-    built-ins, then restores it so the backlog isn't mutated for other tests.
+    ``Celery(...)`` created inside the test starts with *no tasks at all* —
+    neither the worker's shared tasks nor Celery's own ``celery.*`` built-ins,
+    since both are registered through this same finalizer backlog — then restores
+    it so the backlog isn't mutated for other tests.
     """
     import celery._state as celery_state
 
@@ -192,29 +227,36 @@ def _reset_queue_backend_state():
     order-dependent (green alone, failing in a full run). Reset on teardown so
     each test starts from the same clean module state regardless of order.
 
-    Each reset is guarded independently: a module that fails to import (missing
-    optional dep in a given lane) must not break the autouse fixture for the
-    whole suite.
+    Only the *import* is guarded (a module absent in a given lane must not break
+    the autouse fixture). The resets run outside that guard on purpose: if a
+    reset target is ever renamed/removed, the resulting AttributeError surfaces
+    loudly here rather than turning the whole fixture into a silent no-op — which
+    would let order-dependence and hangs creep back with no failing test.
     """
     yield
-    try:
+    with contextlib.suppress(ImportError):
         import queue_backend.routing as _routing
 
         _routing._allow_list_logged = False
-    except Exception:
-        pass
-    try:
-        import queue_backend.dispatch as _dispatch
+    with contextlib.suppress(ImportError):
+        # queue_backend re-exports a ``dispatch`` *function*, which shadows the
+        # submodule as a package attribute — so ``import queue_backend.dispatch``
+        # would bind the function, not the module. Load the module explicitly.
+        _dispatch = importlib.import_module("queue_backend.dispatch")
 
         _dispatch._pg_routing_logged.clear()
-        if hasattr(_dispatch._pg_local, "client"):
+        client = getattr(_dispatch._pg_local, "client", None)
+        if client is not None:
+            # Close before dropping so a real libpq connection isn't leaked to
+            # GC/__del__ (matters in a large real-Postgres run).
+            with contextlib.suppress(Exception):
+                client.close()
             del _dispatch._pg_local.client
-    except Exception:
-        pass
-    try:
+    with contextlib.suppress(ImportError):
         import queue_backend.pg_barrier as _pg_barrier
 
-        if hasattr(_pg_barrier._local, "conn"):
-            del _pg_barrier._local.conn
-    except Exception:
-        pass
+        conn = getattr(_pg_barrier._local, "conn", None)
+        if conn is not None:
+            with contextlib.suppress(Exception):
+                conn.close()
+            _pg_barrier._local.conn = None
--- a/workers/tests/conftest.py
+++ b/workers/tests/conftest.py
@@ -16,6 +16,77 @@ from dotenv import load_dotenv
 _env_test = Path(__file__).resolve().parent.parent / ".env.test"
 load_dotenv(_env_test)
 
+# Pristine baseline: the environment right after .env.test loads, captured before
+# pytest collection imports any worker module. Some worker modules call
+# ``load_dotenv(<workers>/.env)`` at import time, so on a developer machine the
+# ambient dev ``.env`` (e.g. WORKER_BARRIER_KEY_TTL_SECONDS=180) bleeds into the
+# process during collection and silently overrides test defaults — green in CI
+# (no ``.env``), red locally. ``_restore_os_environ`` resets to this baseline
+# around every test so the suite is deterministic regardless of ambient ``.env``.
+_PRISTINE_ENVIRON = dict(os.environ)
+
+
+# --- Collection: separate real-Postgres tests from the unit lane ---
+
+
+def pytest_collection_modifyitems(config, items):
+    """Auto-mark real-Postgres tests as ``integration``.
+
+    Any test that requests a real-Postgres fixture needs a live database.
+    Marking those ``integration`` here — rather than by hand on each of ~100 tests
+    scattered across mixed unit/integration files — lets the fast lane run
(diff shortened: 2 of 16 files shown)


muhammad-ali-e (commented):

This is careful, well-commented test-infra work: the determinism fixtures target real, correctly-identified process-global leaks (all reset targets verified against `queue_backend/{routing,dispatch,pg_barrier}.py`), the assertion changes are principled narrowings rather than blind weakenings, and the exit-5 CI fix is real and correct. Findings below are ranked; the top two concern the *integration lane's core guarantee* and a live auto-marking gap. Nothing here is a blocker for the fast-lane win, but F1/F2 undercut the safety net this PR advertises.

**Priority order:** F1 (high) → F2 (med) → F3/F4 (med) → F5/F6/F8 (low). Details inline.

workers/tests/conftest.py
```diff
@@ -33,18 +104,34 @@ def integration_pg_conn():
     return create_pg_connection(env_prefix="TEST_DB_")
 
 
+def _skip_or_fail_no_pg(reason: str):
```
muhammad-ali-e, line 142: **[HIGH] The `REQUIRE_PG_TESTS` hard-fail guarantee is false for the barrier / leader-election / reaper suites — the exact suites the integration-lane comment names.**

`_skip_or_fail_no_pg` (which honors `REQUIRE_PG_TESTS`) is only wired into `pg_conn`/`pg_client` here and into `dedup_db` (`test_pg_batch_dedup.py`). The other real-Postgres fixtures still use a plain `pytest.skip`, which ignores the flag:
- `barrier_db` — `test_pg_barrier.py:484,490`
- `lock_db` — `test_leader_election.py:160,165`
- `barrier_conn` — `test_pg_reaper.py:638,644`

So with `REQUIRE_PG_TESTS=1` against a bare/unmigrated Postgres, barrier-decrement, leader-election, and reaper tests **still silently skip** and the lane goes green having exercised none of them — the precise "pass having run nothing" failure the groups.yaml comment (`tests/groups.yaml:95`) claims to have closed. Those fixture files aren't touched by this PR, but this PR introduces the guarantee, so the incomplete wiring is now a correctness gap.

**Fix:** route every real-PG fixture through the shared `_skip_or_fail_no_pg` (or a single shared helper), or add a collection-time assertion that when `REQUIRE_PG_TESTS` is set, zero `integration`-marked tests end up `skipped`.

workers/tests/conftest.py
```diff
@@ -16,6 +16,77 @@
...
+    scattered across mixed unit/integration files — lets the fast lane run
+    ``-m "not integration"`` (DB-free, deterministic) while a dedicated
+    integration lane runs ``-m integration`` against a provisioned Postgres.
+
+    The set lists every fixture that opens a real connection directly; fixtures
+    layered on top of these (e.g. ``pg_client``/``result_backend`` build on
+    ``pg_conn``) are covered transitively because ``fixturenames`` includes the
+    whole dependency graph.
+    """
+    pg_fixtures = {
```
muhammad-ali-e, line 48: **[MEDIUM] Auto-marking is a hand-maintained fixture allow-list with a live gap.**

A test is marked `integration` only if it *requests* one of these six fixtures. `test_create_pg_connection_is_non_autocommit` (`test_pg_barrier.py:451`) opens a real connection inline via `create_pg_connection(env_prefix="TEST_DB_")` and requests **no** fixture — so it is never marked. Consequences today:
1. It lands in the fast `unit-workers` lane (`-m "not integration"`), which is meant to be DB-free, and makes a real connect attempt on every run (a connect-timeout stall — now bounded by the new `--timeout=120`, i.e. up to a 120s hang before it trips).
2. It is excluded from `integration-workers`, and its own `pytest.skip` bypasses `REQUIRE_PG_TESTS` — so it can silently skip even in the lane built to hard-fail on a missing DB.

This is a hole in the invariant the PR establishes ("every real-PG test is marked integration"), and the allow-list will drift silently as new PG fixtures are added.

**Fix:** mark by connection provenance rather than fixture name (a single `create_pg_connection` seam all PG tests go through, marked centrally), or add a guard asserting `pg_fixtures` covers every fixture/test that opens a real connection. At minimum, give this specific test an explicit `@pytest.mark.integration` and route its skip through `_skip_or_fail_no_pg`.

workers/tests/conftest.py
```diff
@@ -56,3 +143,78 @@ def pg_client(pg_conn):
...
+    Each reset is guarded independently: a module that fails to import (missing
+    optional dep in a given lane) must not break the autouse fixture for the
+    whole suite.
+    """
+    yield
+    try:
+        import queue_backend.routing as _routing
+
+        _routing._allow_list_logged = False
+    except Exception:
```
muhammad-ali-e: **[MEDIUM] `except Exception: pass` here is too broad — it turns the determinism reset into a silent no-op if a global is renamed.**

The docstring's intent is to tolerate a *module that fails to import* in a lane. But `except Exception` also swallows `AttributeError`/`NameError` from a future rename or removal of the reset targets (`_routing._allow_list_logged`, `_dispatch._pg_routing_logged`, `_dispatch._pg_local.client`, `_pg_barrier._local.conn`). All four exist today (verified), so the reset is correct now — but if any is renamed, this autouse fixture silently stops resetting, order-dependence/flakiness/hangs return, and **no test fails**. That's the exact failure this fixture exists to prevent, made undetectable.

**Fix:** narrow the guard to `except ImportError` (or `ModuleNotFoundError`) around the import, and do the attribute resets *outside* the import guard so an `AttributeError` from a rename surfaces loudly. A small meta-test that sets each flag/thread-local and asserts it was reset would make the fixture self-verifying. (Cosmetic: `contextlib.suppress(ImportError)` reads cleaner than `try/except: pass`.)

workers/tests/conftest.py
```diff
@@ -56,3 +143,78 @@ def pg_client(pg_conn):
...
+    A truly empty ``Celery(...)`` app is impossible once any ``@shared_task``
+    has been imported: Celery keeps a *process-global* finalizer backlog
+    (``celery._state._on_app_finalizers``) and injects every declared shared
+    task into every *new* app on finalize. So ``Celery("empty")`` created in a
+    full-suite run actually carries the worker's shared tasks — which silently
+    defeats the consumer's empty-registry guard (the guard passes, ``run()``
+    enters its poll loop, and the test hangs).
+
+    This fixture snapshots and clears that backlog for the test's duration, so a
+    ``Celery(...)`` created inside the test starts with only ``celery.*``
```
muhammad-ali-e: **[MEDIUM — comment accuracy] "starts with only `celery.*` built-ins" is factually wrong.**

Celery's built-in tasks (`celery.chord`, `celery.group`, `celery.backend_cleanup`, …) are themselves registered via `@connect_on_app_finalize`, i.e. they live in the very `_on_app_finalizers` set this fixture clears. Clearing it strips the built-ins too — a `Celery(...)` finalized during the fixture has **zero** tasks, not "only `celery.*` built-ins" (verified empirically in the workers venv: a fresh finalized app reports `[]`). The test still passes because `_registered_task_count()` filters `celery.*` and counts 0 either way — only the stated reason is wrong.

**Fix:** reword to "…starts with **no tasks at all** — neither the worker's shared tasks nor Celery's `celery.*` built-ins, since both are registered through the same finalizer backlog." Same wrong model appears at `test_pg_queue_consumer.py:227`.

workers/tests/test_pg_queue_consumer.py
```diff
@@ -217,12 +217,14 @@ def _sleep(secs):
...
+    def test_run_refuses_to_start_with_empty_registry(self, isolated_celery_registry):
         # A non-bootstrapped consumer would drop every message as "unknown" —
-        # fail loudly instead.
+        # fail loudly instead. isolated_celery_registry makes "empty" genuinely
+        # empty (Celery's global shared_task backlog would otherwise leak the
+        # worker's tasks in and let the guard pass → infinite poll loop / hang).
         from celery import Celery
 
-        empty_app = Celery("empty-no-tasks")  # only celery.* built-ins
+        empty_app = Celery("empty-no-tasks", set_as_current=False)  # only celery.* built-ins
```
muhammad-ali-e: **[LOW — comment accuracy] `# only celery.* built-ins` is inaccurate here.** This test now requests `isolated_celery_registry`, which clears the finalizer backlog, so this app has **zero** tasks — not "only celery.* built-ins" (clearing `_on_app_finalizers` removes the built-ins too; see conftest.py:169). Suggest: `# no tasks at all (isolated_celery_registry cleared the finalizer backlog)`.

Related: the comment on `test_registered_task_count_excludes_celery_builtins` (a few lines below) says built-ins are "present in every app" and must not mask an un-bootstrapped registry — but adding `isolated_celery_registry` to that test removes the built-ins, so it no longer exercises "built-ins present but excluded." Either drop the fixture from that specific test (so built-ins are genuinely present) or reconcile the comment.

tests/groups.yaml
```diff
@@ -70,10 +70,33 @@ groups:
...
+    # unreachable or the pg_queue schema isn't migrated.
+    markers: "integration"
+    uv_sync_group: test
+    coverage_source: shared
+    requires_services: [postgres]
+    # Optional until CI provisions a *migrated* database and sets
+    # REQUIRE_PG_TESTS=1 — that env flips the graceful skips into hard failures
+    # (see conftest `_skip_or_fail_no_pg`) so the lane can't pass having run
+    # nothing. Until then it runs-and-skips against a bare Postgres.
+    optional: true
```
muhammad-ali-e, line 98: **[LOW] `integration-workers` is a permanently-green no-op today, and nothing enforces the promised follow-up.**

`REQUIRE_PG_TESTS` is set nowhere in `.github/` (grep-confirmed — only read in conftest/tests), and `requires_services: [postgres]` is cosmetic: `tests/rig/cli.py` only appends it to a display flag; it provisions nothing and gates nothing. CI's integration tier runs `-m integration` with no Postgres and no `REQUIRE_PG_TESTS`, so the lane collects, skips everything, exits 0, and — being `optional: true` — reports green. It looks like coverage but exercises none of the PG-queue concurrency-correctness tests (the exact areas the surrounding commits are fixing).

This is intentional-for-now ("Optional until CI provisions a *migrated* database"), but it's easy to ship a lane that's a no-op indefinitely. **Fix:** link a tracking ticket in this comment, or add a guard (xfail/TODO or a CI assertion) that fails once a migrated DB is provisioned but the `REQUIRE_PG_TESTS=1` + `optional: false` flip is still missing — so the placeholder can't silently become permanent.

workers/tests/conftest.py
```diff
@@ -56,3 +143,78 @@ def pg_client(pg_conn):
...
+
+        _routing._allow_list_logged = False
+    except Exception:
+        pass
+    try:
+        import queue_backend.dispatch as _dispatch
+
+        _dispatch._pg_routing_logged.clear()
+        if hasattr(_dispatch._pg_local, "client"):
+            del _dispatch._pg_local.client
```
muhammad-ali-e, line 256: **[LOW] Thread-local real connections are dropped without closing.** `del _dispatch._pg_local.client` (here) and `del _pg_barrier._local.conn` (below) discard handles that may hold open libpq connections, relying on GC/`__del__` to close them. In a large real-Postgres run this can leak connections and, worst case, approach `max_connections`. Low likelihood (most tests inject monkeypatched connections whose own fixtures manage lifecycle), but cheap to fix: `close()` the handle (under `contextlib.suppress(Exception)`) before deleting, then set the thread-local to `None`.

workers/tests/test_sanity_phase6c.py
```diff
@@ -432,8 +432,16 @@ def test_highlight_skipped_when_disabled(
...
-        # Plugin loader should NOT have been called
-        mock_plugin_get.assert_not_called()
+        # The highlight plugin must NOT be loaded when highlight is disabled.
+        # Other plugins (e.g. lookup-enrichment) load independently of highlight,
+        # so assert on the highlight plugin specifically rather than "never
+        # called".
+        highlight_loads = [
+            call
+            for call in mock_plugin_get.call_args_list
+            if call.args and call.args[0] == "highlight-data"
```
muhammad-ali-e: **[LOW] This filter can pass vacuously.** `call.args and call.args[0] == "highlight-data"` only matches the *positional* form. The narrowing itself is correct — production loads the highlight plugin via a positional `ExecutorPluginLoader.get("highlight-data")` (`legacy_executor.py:1394`), and other plugins now legitimately load during `_handle_answer_prompt`, so the old blanket `assert_not_called()` had to go. But if that call ever becomes keyword (`get(name="highlight-data")`), the list comprehension silently yields empty and the test passes as a false negative. **Fix:** match both positional and keyword forms, and/or additionally assert the loader *was* called for an expected sibling plugin so the filter can't pass vacuously.

workers/tests/test_phase1_log_streaming.py
```diff
@@ -347,11 +347,18 @@ def test_extract_passes_log_info_to_shim(
...
         assert result.success
-        mock_shim_cls.assert_called_once_with(
-            platform_api_key="sk-test",
-            log_events_id="session-abc",
-            component={"tool_id": "t1", "run_id": "r1", "doc_name": "test.pdf"},
-        )
+        # Assert the log-info kwargs specifically (the point of this test) rather
+        # than the full constructor — the shim gained execution/org/file-exec ids
+        # unrelated to log passthrough.
+        mock_shim_cls.assert_called_once()
```
muhammad-ali-e, line 353: **[LOW] Narrowing is sound; one small residual gap.** Switching from `assert_called_once_with(...)` to `assert_called_once()` + per-kwarg checks still verifies all three log-passthrough kwargs (`platform_api_key`, `log_events_id`, `component`) — the point of this test — so nothing load-bearing is now unverified. Two notes: (a) no test now verifies the shim's *new* constructor args (`execution_id`/`organization_id`/`file_execution_id`) are passed correctly — a small coverage gap for the new plumbing; consider one assertion elsewhere covering that path. (b) `call_args.kwargs[...]` assumes keyword passing — fine since production uses kwargs, but it would `KeyError` rather than assert cleanly if that ever changed.

muhammad-ali-e:
Thanks for the thorough review — addressed in 21d376c. Summary per finding:

**F1 (HIGH) — `REQUIRE_PG_TESTS` guarantee was incomplete.** Correct catch: only `pg_conn`/`pg_client`/`dedup_db` routed through the fail-loud helper; `barrier_db`/`lock_db`/`barrier_conn` used plain `pytest.skip` and ignored the flag. Rather than chase every fixture (and every future one), enforced it **centrally**: a `pytest_runtest_makereport` hook turns any skipped `integration`-marked test into a failure when `REQUIRE_PG_TESTS` is set. Verified: bad DB + flag → non-passing error; no flag → skips green.

**F2 (MED) — auto-marking gap.** Marked `test_create_pg_connection_is_non_autocommit` `@pytest.mark.integration` explicitly (it opens a real connection inline via `create_pg_connection`, no fixture, so the name-based hook missed it). It's now out of the DB-free unit lane, and the F1 hook covers its skip. The broader "mark by connection provenance" idea is a good hardening but a larger change — noted for follow-up.

**F3 (MED) + F7 (LOW).** Narrowed the reset guard to `suppress(ImportError)` with the attribute resets outside the import catch. This **immediately surfaced a latent bug the broad `except Exception` was hiding**: `queue_backend` re-exports a `dispatch` *function* that shadows the submodule, so `import queue_backend.dispatch` bound the function — the dispatch reset had silently never run. Fixed via `importlib.import_module`. Also close cached connections before dropping the thread-locals.

**F4 (MED) + F5 (LOW) — comment accuracy.** Correct: clearing the finalizer backlog removes the `celery.*` built-ins too, so an isolated app has **no** tasks. Fixed the wording in both places. Also reworked `test_registered_task_count_excludes_celery_builtins` to register a `celery.*`-named task and assert it's excluded — so it genuinely exercises the filter its name claims (it no longer passes vacuously on an empty registry).

**F6 (LOW) — placeholder lane.** The F1 hook substantially closes this: the moment `REQUIRE_PG_TESTS=1` is set, the lane can no longer be green-having-run-nothing. Activation (migrate the schema + set the flag + flip `optional`) is tracked in UN-3677; kept the ticket ref out of the code comment per the repo's no-internal-refs-in-code convention.

**F8 (LOW).** Filter now reads the plugin name from positional **or** keyword args, and additionally asserts the sibling `lookup-enrichment` plugin *was* loaded, so it can't pass vacuously.

**F9 (LOW).** Fair point that the shim's new `execution_id`/`organization_id`/`file_execution_id` args aren't verified. Left it out of this test-hygiene PR to avoid scope creep — a small dedicated assertion belongs with the feature that added those args; noting it as a minor coverage follow-up.

Unit lane still 1107 passed / deterministic after the changes.

muhammad-ali-e:
Thanks @greptile-apps — both addressed in 55f4e48.

- **G1 (hookwrapper deprecation):** switched `pytest_runtest_makereport` to the modern `@pytest.hookimpl(wrapper=True)` API — `report = yield` directly, no `outcome.get_result()`. Verified the guarantee still holds: skip→fail under `REQUIRE_PG_TESTS`, skip (green) without it; unit lane 1107 passed / deterministic.
- **G2 (stale exit-5 comment):** added a note on `_NON_FAILING_PYTEST_EXIT_CODES` clarifying exit-5 is non-failing for *optional* groups only — required groups that collect nothing fail via the explicit guard in `cmd_run`.

PR #2140, opened 2026-07-02
UN-3669 [FIX] PG queue — barrier abort + poison-drop leave a recoverable terminal handle (UN-3670)

muhammad-ali-e:
## What & why

Two PG-path failures could strand an execution `EXECUTING` forever with no handle the reaper can find. Both are pre-rollout blockers (HIGH).

### UN-3669 — barrier abort leaves no recoverable handle
On any batch exception, the in-body abort deleted the `pg_barrier_state` row **and** cleared the per-batch `pg_batch_dedup` markers **without** marking the execution terminal. Consequences:
1. **Permanent EXECUTING strand** — after abort nothing marks the execution, and the reaper only recovers executions that still have a barrier row.
2. **Zombie batch re-run** — because the markers were cleared, a vt-redelivered message re-claimed and re-ran the **whole batch** (real LLM spend + destination writes) against a torn-down barrier.

### UN-3670 — poison-drop silently discards the execution
When `read_ct > max_attempts`, the consumer deleted the message with only a log line. Pipeline headers and barrier callbacks have no `reply_key`/`on_error`, so their work was silently discarded. Sharpest case: the aggregating callback — the barrier row is deleted right after the callback is *enqueued*, so a ~3-minute backend/internal-API outage while callbacks are in flight permanently stranded every affected execution.

## Fix

Both paths now mark the execution **ERROR** (`+cascade_terminal_files`) **before** erasing their recovery handle:

- **Barrier abort** (`pg_barrier.py`): mark ERROR first; only on a *confirmed* mark tear the barrier row down, keeping the dedup markers so a redelivered batch is skipped by `claim_batch` (belt-and-braces with the terminal-execution guard) rather than re-run wholesale. If the mark can't be confirmed (backend down / no org) the barrier row is **left intact for the reaper**.
- **Poison-drop** (`consumer.py`): for a message carrying `execution_id` with no failure channel, mark ERROR; if the mark can't be confirmed, **re-park** the message with a long vt (bounded budget) instead of deleting into a void, so the drop never races a dead backend. Messages with `reply_key`/`on_error` keep existing behavior.

Shared `mark_execution_error` helper (`queue_backend/pg_queue/recovery.py`) + new `PgQueueClient.set_vt` re-park primitive.

Everything lives in `queue_backend/` + the internal-API client, reachable only on the PG transport. `barrier_pg_abort` defaults `preserve_dedup_markers=False`, so the Celery `link_error` path is byte-unchanged; the reply-channel poison path is unchanged. Both covered by tests.

## Testing
Dev-tested locally (worker venv against local Postgres):
- **Unit lane: 1128 passed / 0 failed** (deterministic).
- **Touched integration lane green** against local Postgres (`test_pg_barrier`, `test_pg_batch_dedup`, `test_pg_queue_client`).
- New tests: recovery helper (4); barrier mark-first + preserve-markers + unconfirmed-strand (7); consumer poison mark/re-park/budget/reply-channel (5) + `_pipeline_identity` (5); client `set_vt` (3).

Live acceptance repro (kill a batch mid-run; block the internal API mid-callback) is deferred to the Integration deploy + load-test phase, where the PG fleet and raised VT are available.

## Follow-up
PR-C (UN-3671, orchestration idempotency claim) lands separately.

### DIFF
--- a/workers/queue_backend/pg_queue/consumer.py
+++ b/workers/queue_backend/pg_queue/consumer.py
@@ -26,12 +26,14 @@ import os
 import signal
 import time
 from collections.abc import Callable
+from enum import Enum
 from typing import TYPE_CHECKING, TypeVar
 
 from celery import current_app
 
 from unstract.core.data_models import ContinuationSpec, TaskPayload
 
+from ..barrier import callback_recovery_identity
 from ..fairness import FAIRNESS_HEADER_NAME
 from .client import PgQueueClient
 from .liveness import LivenessServer as _BaseLivenessServer
@@ -92,25 +94,43 @@ def _json_safe(value: object) -> object:
     return json.loads(json.dumps(value, default=str))
 
 
+class _PoisonMarkOutcome(Enum):
+    """Result of trying to mark a poison-dropped execution ERROR (drives whether
+    the consumer drops the message now or re-parks it).
+    """
+
+    CONFIRMED = "confirmed"  # marked terminal → safe to drop the message
+    UNMARKABLE = "unmarkable"  # permanent (no org) → drop now; re-park can't help
+    TRANSIENT = "transient"  # backend down / client build failed → re-park
+
+
 def _pipeline_identity(payload: TaskPayload) -> tuple[str | None, str]:
     """Best-effort ``(execution_id, organization_id)`` from a fire-and-forget
     payload, for marking the execution ERROR on a poison drop.
 
-    Barrier callbacks carry both directly in ``kwargs`` (the sharpest strand
-    case). Pipeline headers carry ``execution_id`` in the injected
-    ``_barrier_context`` and the org on the ``fairness`` payload. Returns
-    ``(None, "")`` when the payload isn't a pipeline message — nothing to mark.
-    ``organization_id`` may be ``""`` even with an ``execution_id`` (the caller
-    logs and skips the mark, since the status API is org-scoped).
+    The identity lives in one of three places, tried in order: directly on a
+    callback payload's ``kwargs`` (the aggregating-callback case — the sharpest
+    strand); on the injected ``_barrier_context``'s callback descriptor (a
+    pipeline header); or the org on the ``fairness`` payload. The callback-
+    descriptor dig goes through :func:`callback_recovery_identity` so it can't
+    drift from the barrier abort site. Returns ``(None, "")`` when the payload
+    isn't a pipeline message — nothing to mark. ``organization_id`` may be ``""``
+    even with an ``execution_id`` (the caller drops, since the status API is
+    org-scoped and re-parking can't conjure an org).
     """
     kwargs = payload.get("kwargs") or {}
     barrier_ctx = kwargs.get("_barrier_context") or {}
-    execution_id = kwargs.get("execution_id") or barrier_ctx.get("execution_id")
-    organization_id = kwargs.get("organization_id") or (
-        (barrier_ctx.get("callback_descriptor") or {}).get("kwargs") or {}
-    ).get("organization_id")
-    if not organization_id:
-        organization_id = (payload.get("fairness") or {}).get("org_id")
+    cb_execution_id, cb_org = callback_recovery_identity(
+        barrier_ctx.get("callback_descriptor") or {}
+    )
+    execution_id = (
+        kwargs.get("execution_id") or barrier_ctx.get("execution_id") or cb_execution_id
+    )
+    organization_id = (
+        kwargs.get("organization_id")
+        or cb_org
+        or (payload.get("fairness") or {}).get("org_id")
+    )
     return (
         str(execution_id) if execution_id else None,
         str(organization_id or ""),
@@ -251,66 +271,7 @@ class PgQueueConsumer:
         # it (with the payload, so it's recoverable from logs) instead of
         # redelivering on every vt expiry forever.
         if message.read_ct > self.max_attempts:
-            execution_id, organization_id = _pipeline_identity(payload)
-            logger.error(
-                "PG-queue consumer: task %r (msg_id=%s) exceeded max_attempts=%s "
-                "(read_ct=%s, execution_id=%s) — poison; full payload: %r",
-                task_name,
-                message.msg_id,
-                self.max_attempts,
-                message.read_ct,
-                execution_id,
-                payload,
-            )
-            # Messages with a failure channel (request-reply / on_error callback)
-            # surface the failure there — existing behavior, drop after.
-            if reply_key or on_error:
-                self._fail_dispatch(
-                    payload,
-                    error=f"task {task_name} exceeded max_attempts={self.max_attempts}",
-                )
-                self._client.delete(message.msg_id)
-                return
-            # No failure channel. A pipeline header / barrier callback carries an
-            # execution_id but neither reply_key nor on_error, so a bare delete
-            # silently strands the execution EXECUTING-forever (barrier row already
-            # gone → the reaper has no handle). Mark it ERROR first so the failure
-            # is visible and re-runnable.
-            if execution_id is None:
-                # Not a pipeline message and no failure channel — nothing to mark.
-                self._client.delete(message.msg_id)
-                return
-            if self._mark_poison_execution_error(
-                execution_id, organization_id, task_name
-            ):
-                self._client.delete(message.msg_id)  # terminal → safe to drop
-                return
-            # Mark unconfirmed (backend down): re-park with a long vt rather than
-            # delete into a void, so the poison drop never races a dead backend —
-            # bounded so a permanently-unmarkable message can't re-park forever.
-            if message.read_ct > self.max_attempts + self._poison_repark_budget:
-                logger.error(
-                    "PG-queue consumer: exhausted poison re-park budget for "
-                    "execution %s (msg_id=%s, read_ct=%s) — dropping despite an "
-                    "unconfirmed ERROR mark; the reaper is the last net. Full "
-                    "payload: %r",
-                    execution_id,
-                    message.msg_id,
-                    message.read_ct,
-                    payload,
-                )
-                self._client.delete(message.msg_id)
-                return
-            self._client.set_vt(message.msg_id, self._poison_repark_vt_seconds)
-            logger.warning(
-                "PG-queue consumer: could not confirm ERROR mark for poison "
-                "execution %s (msg_id=%s, read_ct=%s) — re-parked for %ss "
-                "(backend down?) instead of dropping.",
-                execution_id,
-                message.msg_id,
-                message.read_ct,
-                self._poison_repark_vt_seconds,
-            )
+            self._drop_poison_message(message, payload, task_name)
             return
 
         task = self._app.tasks.get(task_name)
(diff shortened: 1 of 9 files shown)


workers/queue_backend/pg_queue/consumer.py
```diff
@@ -206,20 +251,66 @@ def _handle(self, message: QueueMessage) -> None:
...
+                return
+            if self._mark_poison_execution_error(
+                execution_id, organization_id, task_name
+            ):
+                self._client.delete(message.msg_id)  # terminal → safe to drop
+                return
+            # Mark unconfirmed (backend down): re-park with a long vt rather than
+            # delete into a void, so the poison drop never races a dead backend —
+            # bounded so a permanently-unmarkable message can't re-park forever.
+            if message.read_ct > self.max_attempts + self._poison_repark_budget:
```
muhammad-ali-e: **[MEDIUM · Silent-Failure Hunter] The "reaper is the last net" guarantee is false for the aggregating-callback sub-case.**

Budget exhaustion here deletes the poison message despite an *unconfirmed* ERROR mark, and the log/comment claim the reaper will recover it. But `recovery.py` states the reaper only recovers executions that **still have a `pg_barrier_state` row** — and an aggregating-callback message (UN-3670, the sharpest strand case) is enqueued *after* the barrier row is deleted. So when that callback message poisons and budget is exhausted, there is no barrier row and no queue message left → the execution is stranded EXECUTING forever, i.e. the exact bug this PR fixes, re-introduced in bounded form (default ~25 min of backend downtime = 5×300s before giving up).

It is logged at ERROR with the full payload, so it is manually recoverable — not fully silent. But the guarantee is overstated.

**Fix:** correct the log/comment to stop asserting guaranteed reaper recovery for the callback sub-case (e.g. "the reaper recovers this only if a barrier row still exists; an aggregating-callback drop here has no handle and needs manual replay"). Optionally raise `poison_repark_budget` for callback-shaped payloads, since those are precisely the ones with no reaper fallback.

muhammad-ali-e (commented):
## Automated PR review — PR Review Toolkit (6 agents)

Code Reviewer, Silent Failure Hunter, Type Design Analyzer, PR Test Analyzer, Comment Analyzer, and Code Simplifier all ran against this diff. **No Critical/High correctness bugs** — the mark-before-teardown ordering, the unconfirmed-leaves-handle paths, and the re-park budget math were independently verified sound (unit lane: 24 passed). Findings below are MEDIUM/LOW hardening, type-design, comment-accuracy, and test-coverage items, posted inline. See the PR conversation for the prioritized summary.

_Note: one raised concern — `recovery.py:68` `getattr(response, "success", True)` defaulting "unknown → success" — was **verified safe** by three agents against the real `APIResponse.success` contract (`.post` raises on HTTP failure; the `True` default only applies to a legacy raise-on-failure response that would already have raised). Not flagged inline._

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -206,20 +251,66 @@ def _handle(self, message: QueueMessage) -> None:
...
+            if execution_id is None:
+                # Not a pipeline message and no failure channel — nothing to mark.
+                self._client.delete(message.msg_id)
+                return
+            if self._mark_poison_execution_error(
+                execution_id, organization_id, task_name
+            ):
+                self._client.delete(message.msg_id)  # terminal → safe to drop
+                return
+            # Mark unconfirmed (backend down): re-park with a long vt rather than
```
muhammad-ali-e: **[LOW · Comment Analyzer] Comment narrows the re-park trigger to "backend down".**

This branch is reached for **every** `False` return of `_mark_poison_execution_error` — which also includes the no-org case (`consumer.py:556`) and an API-client build failure (`consumer.py:566`), not only a down backend.

**Fix:** `# Mark unconfirmed (backend down, no org, or client-build failure): re-park with a long vt rather than delete into a void...`

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -206,20 +251,66 @@ def _handle(self, message: QueueMessage) -> None:
...
+                    "unconfirmed ERROR mark; the reaper is the last net. Full "
+                    "payload: %r",
+                    execution_id,
+                    message.msg_id,
+                    message.read_ct,
+                    payload,
+                )
+                self._client.delete(message.msg_id)
+                return
+            self._client.set_vt(message.msg_id, self._poison_repark_vt_seconds)
```
muhammad-ali-e: **[LOW · Code Reviewer + Silent-Failure Hunter + Type Design — 3-agent consensus] `set_vt` return value is ignored, so the re-park log can lie.**

`set_vt` returns `False` when the row is already gone (vt expired and another reader deleted it), but the result is discarded and the `logger.warning("...re-parked for %ss...")` fires unconditionally — claiming a re-park that never happened. Functionally harmless (a gone row means another reader already handled it), log-accuracy only.

**Fix:** `if not self._client.set_vt(message.msg_id, self._poison_repark_vt_seconds): logger.info("poison msg %s already gone before re-park", message.msg_id); return` before emitting the re-park warning.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -443,6 +534,54 @@ def _fail_dispatch(self, payload: TaskPayload, *, error: str) -> None:
...
+    def _mark_poison_execution_error(
+        self, execution_id: str, organization_id: str, task_name: str | None
+    ) -> bool:
+        """Best-effort: mark a poison-dropped pipeline execution ERROR (+cascade).
+
+        Returns ``True`` iff the mark is confirmed (safe to delete the message);
+        ``False`` when it can't be confirmed (no org, or the backend is down) so
+        the caller re-parks instead of dropping into a void.
+        """
+        if not organization_id:
```
muhammad-ali-e, line 608: **[MEDIUM · Type Design + Comment Analyzer] No-org poison message re-parks for the full budget uselessly, and this log says "dropping" while the caller actually re-parks.**

Two linked issues on the no-org branch:
1. **Behavior:** returning `False` sends the caller into the re-park loop (`consumer.py:283→291`). But a *missing org is permanent* — the org-scoped status API can never be called for this message, so re-parking cannot change the outcome. It burns up to `poison_repark_budget × poison_repark_vt_seconds` (~25 min of dead polls) before finally dropping, when it could drop immediately. (The barrier caller doesn't loop, so only the consumer is bitten.)
2. **Comment/log drift:** this log says `"...dropping (the reaper is the recovery net)."`, but on `False` the caller re-parks, not drops.

**Fix (preferred):** distinguish *permanent* (no-org → drop now) from *transient* (backend down → re-park) — e.g. return a small 3-state result (`CONFIRMED`/`TRANSIENT`/`UNMARKABLE`) instead of a bool, and drop-now on `UNMARKABLE`. Minimal fix: special-case the no-org poison message to `delete` immediately and correct this log.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -83,6 +92,31 @@ def _json_safe(value: object) -> object:
...
+    case). Pipeline headers carry ``execution_id`` in the injected
+    ``_barrier_context`` and the org on the ``fairness`` payload. Returns
+    ``(None, "")`` when the payload isn't a pipeline message — nothing to mark.
+    ``organization_id`` may be ``""`` even with an ``execution_id`` (the caller
+    logs and skips the mark, since the status API is org-scoped).
+    """
+    kwargs = payload.get("kwargs") or {}
+    barrier_ctx = kwargs.get("_barrier_context") or {}
+    execution_id = kwargs.get("execution_id") or barrier_ctx.get("execution_id")
+    organization_id = kwargs.get("organization_id") or (
```
muhammad-ali-e: **[MEDIUM · Type Design (highest-value type finding) + Code Simplifier] The load-bearing recovery contract is an untyped, stringly-keyed deep-dig duplicated across two modules.**

The one value the whole strand-fix depends on — `organization_id` (and `execution_id`) living in the barrier callback's kwargs — is reached via a 5-level `.get(...) or {}` chain here (`payload → kwargs → _barrier_context → callback_descriptor → kwargs → organization_id`) **and again** in `pg_barrier.py:_mark_execution_error_on_abort`. Neither is a typed field of `BarrierContext`/`CallbackDescriptor`; the magic strings are enforced only by matching test fixtures. Because every miss is best-effort-swallowed → `False`, a producer-side key rename would **silently** regress the entire recovery net with no test failing — the worst failure mode for a safety net.

**Fix (high value):** promote `organization_id`/`execution_id` to typed fields on `CallbackDescriptor`/`BarrierContext`, and route both recovery sites through one shared typed accessor (e.g. `barrier_recovery_identity(ctx) -> (execution_id, organization_id)`) so a rename is caught at the type layer and the deep-dig lives in one place. Minimal: extract the magic keys to shared constants + one helper so the two sites can't drift.

_Simplifier's lighter companion suggestion: even without typing, hoist the intermediate dicts so the org fallback reads as a flat 3-source chain (`kwargs → callback_kwargs → fairness`) instead of the nested parenthesization + separate reassignment._

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -83,6 +92,31 @@ def _json_safe(value: object) -> object:
     return json.loads(json.dumps(value, default=str))
 
 
+def _pipeline_identity(payload: TaskPayload) -> tuple[str | None, str]:
```
muhammad-ali-e, line 107: **[LOW · Type Design] `_pipeline_identity` return `tuple[str | None, str]` uses two different "absent" sentinels with an unenforced cross-field rule.**

Slot 0 uses `None` for "not a pipeline message"; slot 1 uses `""` for "org unknown". The illegal state `(None, "org-x")` is representable but nonsensical, and callers must remember positional order + the two distinct absences.

**Fix:** model as `PipelineIdentity | None`, where `None` = "not a pipeline message" and the object carries `execution_id: str` (guaranteed present) + `organization_id: str | None`. Makes the two absences structurally distinct and `(None, "org")` unrepresentable. The caller already unpacks and branches on `execution_id is None`, so it's near-mechanical.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -206,20 +251,66 @@ def _handle(self, message: QueueMessage) -> None:
...
+            # No failure channel. A pipeline header / barrier callback carries an
+            # execution_id but neither reply_key nor on_error, so a bare delete
+            # silently strands the execution EXECUTING-forever (barrier row already
+            # gone → the reaper has no handle). Mark it ERROR first so the failure
+            # is visible and re-runnable.
+            if execution_id is None:
+                # Not a pipeline message and no failure channel — nothing to mark.
+                self._client.delete(message.msg_id)
+                return
+            if self._mark_poison_execution_error(
```
muhammad-ali-e: **[Test coverage · PR Test Analyzer] Three untested branches on the exact silent-strand path this PR fixes.**

Coverage is strong overall, but these poison branches have no test and would let a real regression pass green:
- **Gap A (sev 7) — re-park budget boundary is off-by-one untested.** Tests use `read_ct=6` (re-park) and `read_ct=11` (drop); the exact boundary `read_ct == max_attempts + budget` (=10, which must **re-park**) is never asserted. A `>`→`>=` regression drops one cycle early into the unconfirmed-delete void and passes every current test. Add `read_ct=10` asserting `set_vt` called / `delete` not.
- **Gap B (sev 6) — no-org poison path untested.** Every poison test carries `organization_id="org-1"`. The `execution_id`-present-but-empty-org shape (`_pipeline_identity` returns `("e4", "")`) never runs through the poison flow. Add a poison payload `{kwargs: {execution_id: "e4"}}` (no org, no reply channel) asserting `mark_execution_error` **not** called and the message re-parked (or dropped, if the no-org behavior fix above lands).
- **Gap C (sev 6) — consumer's own `_get_api_client` build-failure → re-park untested.** All consumer poison tests inject `api_client=MagicMock()`, short-circuiting the lazy build, so `consumer.py:566`'s `except` is never executed. Construct the consumer *without* `api_client`, monkeypatch `shared.api.InternalAPIClient` to raise, assert re-park (not drop).

workers/queue_backend/pg_queue/recovery.py
```diff
@@ -0,0 +1,80 @@
...
+        return False
+    # ``success`` absent → assume the raised-on-failure legacy contract (True).
+    if not getattr(response, "success", True):
+        logger.error(
+            "Marking execution %s ERROR reported success=False — leaving the "
+            "recovery handle in place for the reaper.",
+            execution_id,
+        )
+        return False
+    logger.error(
```
muhammad-ali-e: **[LOW · Silent-Failure Hunter + Code Reviewer] A *successful* recovery mark is logged at ERROR level.**

Every confirmed recovery emits `logger.error("Marked stranded execution ... ERROR (+cascade files)")`, so a success shows up as an ERROR-level line (Sentry/alert noise for a non-error outcome). This mirrors the reaper's `_mark_stranded_error`, so it's consistent and arguably intentional (a strand is noteworthy).

**Fix:** consider `logger.warning` for this confirmed-success case (the failure branches above already use ERROR correctly), or leave as-is for reaper parity. Non-blocking.

muhammad-ali-e:
## Review feedback addressed (commit 49d1d6e91)

Thanks for the 6-agent pass. Fixes below; unit lane 1131 green + touched integration lane green vs local Postgres.

**Poison-drop correctness (the cluster of MEDIUM/LOW around `consumer.py`)**
- **[MEDIUM] "reaper is the last net" overstated for the callback sub-case** — corrected. The budget-exhaustion log no longer claims guaranteed reaper recovery: an aggregating-callback message has no `pg_barrier_state` row left, so it explicitly says manual replay from the logged payload may be required. The bound is now documented as a backstop, not a reaper net for that case.
- **[MEDIUM] no-org message re-parks uselessly + "dropping" log drift** — fixed by replacing the bool return of `_mark_poison_execution_error` with a 3-state `_PoisonMarkOutcome`: `CONFIRMED` → drop, `UNMARKABLE` (no org, permanent) → **drop now** (no wasted ~25 min budget), `TRANSIENT` (backend down / client build failed) → re-park. Logs now match behavior.
- **[LOW] `set_vt` return ignored → log can lie** — now checked; a message already gone before re-park logs an INFO "already gone" and skips the re-park warning.
- **[LOW] "backend down" comment too narrow** — widened to "backend down, no org, or client-build failure".

**[MEDIUM · type-design] load-bearing identity was a duplicated stringly deep-dig** — extracted `barrier.callback_recovery_identity(callback_descriptor)`, now used by **both** recovery sites (`consumer._pipeline_identity` and `pg_barrier._mark_execution_error_on_abort`), so a producer-side kwarg rename can't silently regress the net in two places. (Went with the shared-accessor "minimal" option rather than promoting to typed `TypedDict` fields, since `execution_id`/`organization_id` are legitimately the callback's own kwargs — duplicating them as top-level fields would fork the source of truth.)

**[LOW] successful recovery logged at ERROR** — `recovery.mark_execution_error` now logs a *confirmed* mark at WARNING (the underlying failure is already logged ERROR by the caller); the failure branches keep ERROR.

**[Test coverage — all three gaps added]**
- Gap A: re-park budget boundary `read_ct == max_attempts + budget` (=10) asserts **re-park**, guarding a `>`→`>=` regression.
- Gap B: no-org poison asserts drop-now + `mark_execution_error` **not** attempted.
- Gap C: consumer's own lazy `_get_api_client` build-failure (no injected client, `InternalAPIClient` raises) asserts **re-park**.

**SonarCloud (3)**
- `consumer._handle` cognitive complexity 26→under 15 by extracting `_drop_poison_message`.
- Two `pytest.raises` one-throwing-call nits: hoisted the shared mock/client out of the `with` block.

**Deferred (LOW, noted):** modelling `_pipeline_identity`'s return as `PipelineIdentity | None` (the two-sentinel `tuple[str | None, str]` finding) — the 3-state outcome enum covers the mark path; happy to fold the identity dataclass into a follow-up if you'd prefer it here.

**Verified-safe, not changed:** `recovery.py` `getattr(response, "success", True)` default — confirmed against the real `APIResponse` contract (a real failure raises or returns `success=False`; the `True` default only covers a legacy raise-on-failure response that would already have raised), matching the reaper's read path.

muhammad-ali-e:
## Greptile + SonarCloud feedback addressed (commit 86cf05843)

**Greptile P2 — `pg_barrier.py` bare `callback_descriptor` key access ✅ fixed**
Good catch: `_mark_execution_error_on_abort` runs inside `run_batch_with_barrier`'s `except` block, so a missing `callback_descriptor` would raise a `KeyError` that masks the original batch exception before the `raise`. Switched to `barrier_context.get("callback_descriptor") or {}` (and `callback_recovery_identity` already tolerates an empty dict → `(None, None)` → leaves the barrier row for the reaper).

**Greptile P2 — `consumer._get_api_client` caching build failures — intentionally not taken**
The current builder mirrors the reaper's `_get_api_client` (caches success, retries on failure). Declining the failure-caching sentinel because:
- The cost is poison-drop-only and bounded by the re-park budget, and `from shared.api import InternalAPIClient` is already import-cached — so the repeated cost is a fast-failing constructor call (you noted it's negligible).
- Caching a *permanent-failure* sentinel would prevent in-process recovery if the build failure were ever transient, and would diverge from the reaper's behavior for no real gain.

Happy to revisit if you'd rather have reaper + consumer both cache failures behind one shared builder — that'd be the consistent way to do it.

**SonarCloud (test_pg_queue_consumer.py L387) ✅ fixed**
Hoisted the shared `MagicMock` client out of `test_rejects_backoff_max_below_poll_interval`'s `pytest.raises` block so only the invocation under test can throw (the twin of the earlier `test_rejects_non_positive_params` fix — this was the actual flagged line).

Unit lane green (140 in the two touched files); full lane 1131 unchanged.

PR #2141, opened 2026-07-02
UN-3671 [FIX] PG orchestrator — idempotency claim for async_execute_bin (double-orchestration guard)

muhammad-ali-e:
## What & why

`async_execute_bin` (the orchestration task) has **no idempotency guard** — its docstring explicitly allows concurrent executions. On the PG queue this is unsafe under redelivery: if the orchestration runs longer than the consumer's visibility timeout and a sibling replica re-claims the message, the orchestration **runs twice**:

- the barrier UPSERT re-arms `remaining = N` and deletes the first run's `pg_batch_dedup` markers mid-flight,
- a second set of N batch headers dispatches,
- up to 2N decrements race `remaining = N` → early/incorrect callback + double LLM spend.

Today this can't happen only because every PG consumer is single-replica / concurrency-1 — correctness by deployment accident. The documented scaling path ("scale replicas") arms it. This closes the window.

## Approach — dedicated claim table

New `pg_orchestration_claim` table (one row per `execution_id`), claimed at orchestration entry with `INSERT … ON CONFLICT DO NOTHING RETURNING`, analogous to `claim_batch` one level up.

Chosen over a `pg_batch_dedup` sentinel because a sentinel would need **two** changes to already-merged code (relax the `batch_index >= 0` CHECK **and** exclude the sentinel from the barrier-arm blanket `DELETE`). The dedicated table touches **zero** merged SQL — lower regression risk.

## Changes

- **Backend:** `PgOrchestrationClaim` model + migration `0012` (additive).
- **Workers:** `claim_orchestration()` in `pg_barrier.py`; `_should_skip_duplicate_orchestration()` gate wired into `async_execute_bin_general` **before** setup / barrier arm / batch dispatch; new table registered in `schema.QUEUE_TABLES`.
- **Semantics:** claim-before-work (a crash mid-orchestration leaves the claim → redelivery no-ops → the reaper recovers the strand, same as `claim_batch`). The claim is a **tombstone** — deliberately not cleared at finalise, so a redelivery *after* completion can't re-orchestrate a finished execution. Reclaimed by the future dedup-orphan sweep (same leak posture as `pg_batch_dedup`, one row per execution).

The gate returns `False` on Celery (`is_pg_transport`), which has no at-least-once redelivery, so the Celery path never claims and is behaviorally unchanged (proven by `test_celery_transport_never_claims`).

## Testing (local)

Dev-tested locally (migration applied to the local dev DB; worker venv):
- **Unit lane: 1135 passed / 0 failed.**
- **Real-Postgres integration** for `claim_orchestration`: first-wins / duplicate-loses / distinct-executions-independent / tombstone.
- Transport-gated gate (Celery never claims; PG first proceeds, duplicate skips) + `async_execute_bin_general` short-circuits to the skip result without setup.

Live acceptance (2 orchestrator replicas + an artificially long orchestration > vt → exactly one set of batches dispatches) is deferred to the Integration deploy + load-test phase.

## Deploy note

Apply the migration **before** PG workers claim (migrate-first — matches the chart's existing migration-Job order). Flag-gated: inert until `pg_queue_enabled`.


workers/general/tasks.py
```diff
@@ -196,6 +227,17 @@ def async_execute_bin_general(
         )
 
         try:
+            # PG-only idempotency gate (claim-before-work): a duplicate /
+            # redelivered orchestration no-ops here BEFORE any setup, barrier arm,
+            # or batch dispatch — so exactly one delivery orchestrates the
+            # execution even with replicas > 1. No-op on Celery.
+            if _should_skip_duplicate_orchestration(execution_id, transport):
```
muhammad-ali-e: **BLOCKER — the committed-before-work claim silently defeats `autoretry_for` and permanently loses the execution.**

`claim_orchestration` commits its row in its own transaction (`_cursor` commits on success) *before* any orchestration work. But this task is decorated `autoretry_for=(Exception,), max_retries=3` and the `except` block re-raises "for Celery retry mechanism". Trace of a transient first-attempt failure on the PG path:
1. Attempt 1: claim row inserted + committed, then setup/`get_workflow_execution`/status-update throws.
2. `except` marks ERROR and re-raises → retry is scheduled.
3. Attempt 2: `claim_orchestration` returns `False` → gate returns `True` → task returns `{"status": "skipped_duplicate_orchestration", ...}` and **never orchestrates.**

The configured retry is nullified and the task returns a *success-shaped* dict masking a lost execution — a silent divergence from the Celery path, where the same blip would retry and usually succeed.

**Fix:** distinguish a concurrent duplicate from this task's own retry. Cheapest: add `release_orchestration_claim(execution_id)` and call it in the `except` block before re-raise when the barrier was never armed and `self.request.retries < self.max_retries`. If "never retry orchestration" is truly intended, then drop `autoretry_for`/`max_retries` for the PG path and don't re-raise "for retry".

workers/queue_backend/pg_barrier.py
```diff
@@ -305,6 +305,39 @@ def clear_execution_batches(execution_id: str) -> int:
...
+
+    Atomically inserts the single ``pg_orchestration_claim`` row; returns ``True``
+    if THIS call inserted it (first delivery — the caller orchestrates) and
+    ``False`` if it already existed (a duplicate / redelivered orchestration — the
+    caller must no-op). ``ON CONFLICT DO NOTHING`` makes the check-and-set a single
+    statement, so two replicas racing the same orchestration resolve to exactly
+    one ``True`` (the row's existence is the token).
+
+    Claimed BEFORE the orchestration work (arm barrier / dispatch batches), so a
+    crash mid-orchestration leaves the claim, a redelivery no-ops, and the reaper
```
muhammad-ali-e: **BLOCKER — crash between claim and barrier-arm is NOT reaper-recoverable; the docstring's "same crash-window semantics as `claim_batch`" is false.**

The claim is taken at the very top of `async_execute_bin_general` (`tasks.py:234`), *before* `setup_execution_context`, status update, batch creation, and barrier arm. The reaper only recovers executions that have a `pg_barrier_state` row (`reaper.py` — barrier-orphan recovery, per-stage re-enqueue explicitly deferred). For `claim_batch` the barrier is *always* armed before any batch task runs, so the reaper always has a handle. Here, a SIGKILL / pod eviction / VT expiry in the **claim→arm window** commits the tombstone but leaves **no** `pg_barrier_state` row — so the reaper never sees it, the redelivery permanently no-ops, and the execution strands `EXECUTING` forever.

This is a *new* unrecoverable window that did not exist before this PR. **Fix:** either arm a recoverable handle atomically with/before the claim, or extend the reaper to sweep claims with no matching barrier state. At minimum, correct this docstring — do not equate it with `claim_batch`.

backend/pg_queue/models.py
```diff
@@ -343,3 +343,46 @@ class Meta:
...
+    ``remaining = N`` and wipes the ``pg_batch_dedup`` markers mid-flight),
+    dispatches a second set of N batch headers, and races the decrements. This
+    table is the durable single-winner token that closes that window: one row per
+    ``execution_id``.
+
+    The worker claims by inserting its row atomically
+    (``INSERT … ON CONFLICT DO NOTHING RETURNING``): a fresh insert means "first
+    delivery — orchestrate"; a conflict means "already claimed — a duplicate /
+    redelivery → no-op". Claimed BEFORE the work, so a crash mid-orchestration
+    leaves the claim and the redelivery no-ops; the reaper recovers the stranded
```
muhammad-ali-e: **BLOCKER (same root cause as the `claim_orchestration` finding).** This docstring repeats "the reaper recovers the stranded execution — same crash-window semantics as `pg_batch_dedup`." That parity does not hold: the claim is committed *before* the barrier is armed, so a crash in the claim→arm window leaves this tombstone with no `pg_barrier_state` row, and the reaper (which only sweeps `pg_barrier_state`) cannot recover it — the redelivery permanently no-ops. Please correct the docstring and address the underlying strand window.

workers/queue_backend/pg_barrier.py
```diff
@@ -305,6 +305,39 @@ def clear_execution_batches(execution_id: str) -> int:
...
+    recovers the stranded execution — the same crash-window semantics as
+    :func:`claim_batch`. Unlike the per-batch markers, this claim is a tombstone:
+    it is deliberately NOT cleared at barrier finalise, so a redelivery AFTER the
+    execution completes can't re-orchestrate a finished execution. It is reclaimed
+    by the (future) dedup-orphan sweep, same as ``pg_batch_dedup``.
```
muhammad-ali-e: **Inaccurate docstring — factually wrong on two counts.** "reclaimed by the (future) dedup-orphan sweep, same as `pg_batch_dedup`":
1. The sweep is **not future** — `sweep_orphan_dedup` already exists and runs on every leader tick (present on the base branch).
2. That live sweep hardcodes `DELETE FROM pg_batch_dedup` and never touches `pg_orchestration_claim` (zero references in the reaper).

So this new table is currently reclaimed by **nothing** — a permanent tombstone with no GC, which is *strictly worse* than the sibling it's compared to, not "the same." Reword to: "not reclaimed yet — `sweep_orphan_dedup` covers only `pg_batch_dedup` and must be extended to also sweep `pg_orchestration_claim` by `claimed_at` age (add an index then)." (The same stale "future" framing exists on `PgBatchDedup`/`clear_execution_batches` and is worth a follow-up.)

workers/queue_backend/pg_barrier.py
```diff
@@ -305,6 +305,39 @@ def clear_execution_batches(execution_id: str) -> int:
...
+
+    PG path only — the caller gates on the transport (Celery has no redelivery, so
+    no double-orchestration to guard).
+    """
+    with _cursor() as cur:
+        cur.execute(
+            f"INSERT INTO {qualified('pg_orchestration_claim')} "
```
muhammad-ali-e: **Deploy-ordering hazard — missing table (migration 0012 not applied) is a silent-until-first-run, fleet-wide outage.** If PG transport is enabled before `0012_pgorchestrationclaim` is migrated, every general orchestration hits `UndefinedTable` here. It propagates (good — no swallow), but surfaces only as the generic `"General workflow execution failed for {execution_id}: {e}"` with no errorId and recurs per-execution; after 3 retries each execution is ERROR. Root cause (unapplied migration) is buried in hundreds of identical generic stack traces.

**Fix:** add the table to the PG-queue startup/preflight existence check (fail fast: "migration 0012 not applied — run migrations before enabling PG transport"), or catch `psycopg2.errors.UndefinedTable` here and re-raise as a typed, actionable error. Do **not** add a silent "table missing → proceed" fallback — that reintroduces the double-orchestration this PR prevents.

workers/tests/test_orchestration_idempotency.py
```diff
@@ -0,0 +1,146 @@
...
+        )
+        return cur.fetchone()[0]
+
+
+@pytest.mark.integration
+class TestClaimOrchestration:
+    def test_first_claim_wins(self, claim_db):
+        assert claim_orchestration("exec-1") is True
+
+    def test_duplicate_claim_loses(self, claim_db):
```
muhammad-ali-e, line 215: **Missing the test for the property this feature exists for — concurrent racing claims.** This test issues two *sequential* claims, which exercises `ON CONFLICT DO NOTHING` but not the contended path. The PR's stated reason to exist is "two replicas racing the same orchestration resolve to exactly one `True`." A regression that swapped the single-statement `INSERT ... ON CONFLICT ... RETURNING` for a check-then-insert would still pass this test while two real replicas both won.

The sibling `test_pg_batch_dedup.py::test_concurrent_claims_resolve_to_exactly_one_winner` already has the right pattern: N threads, each with its own connection injected into `pg_barrier._local.conn`, all released via a `threading.Barrier`; assert `sum(results) == 1` and row count == 1, repeated over several execution_ids. Please add the equivalent for `claim_orchestration`.

workers/general/tasks.py
```diff
@@ -138,6 +139,36 @@ def _log_batch_creation_statistics(
...
+    ``True`` so the caller no-ops.
+
+    No-op on Celery (returns ``False`` — always proceeds): Celery has no
+    at-least-once redelivery, and the router owns terminal status there, so there
+    is no double-orchestration to guard. Gated exactly like the terminal guard so
+    the Celery path is behaviorally unchanged.
+    """
+    if not is_pg_transport(transport):
+        return False
+    if claim_orchestration(execution_id):
```
muhammad-ali-e: **Untested error path — what happens when `claim_orchestration` raises.** The gate has no try/except, so a DB error propagates (correct, safe choice — no swallow). But nothing pins that behavior, and one stray `try/except` would flip it into a silent failure (swallow-and-proceed → destructive double-orchestration; swallow-and-return-`True` → a legitimate first delivery silently skipped). Add a Layer-1 test: patch `claim_orchestration` to raise `psycopg2.OperationalError` and assert `_should_skip_duplicate_orchestration` propagates (does not return a bool) and that `async_execute_bin_general` does not return the skip dict.

workers/queue_backend/pg_barrier.py
```diff
@@ -305,6 +305,39 @@ def clear_execution_batches(execution_id: str) -> int:
         return cur.rowcount
 
 
+def claim_orchestration(execution_id: str) -> bool:
```
muhammad-ali-e: **Readability nit — the bare `-> bool` is not self-documenting, and its polarity is inverted vs. its caller.** `claim_orchestration` returns `True` = "I won, proceed", while the adjacent `_should_skip_duplicate_orchestration` returns `True` = "skip". The call site reads `if claim_orchestration(...): return False` — a double negative that leans on the inline comment. Consider renaming to `try_claim_orchestration` (the `try_` idiom conventionally means "returns whether it won"), keeping the bool to stay consistent with `claim_batch`.

workers/general/tasks.py
```diff
@@ -138,6 +139,36 @@ def _log_batch_creation_statistics(
...
+        return False  # first delivery — this replica owns the orchestration
+    logger.info(
+        f"[exec:{execution_id}] orchestration already claimed — duplicate/"
```
muhammad-ali-e: **Observability nit (depends on the retry-defeat fix).** This skip is logged at INFO as a benign "duplicate/redelivered orchestration." Once the `autoretry_for` interaction is fixed, this same line can also fire when the task's *own* retry is being suppressed — a serious event with identical wording and no errorId for alerting. Include `self.request.retries` in the log and escalate to an error-level log with an errorId when `retries > 0`, so a defeated retry is distinguishable from a genuine duplicate.

backend/pg_queue/models.py
```diff
@@ -343,3 +343,46 @@ class Meta:
             # expires_at <= now()); no sweeper reads it yet (see class docstring).
             models.Index(fields=["expires_at"], name="pg_task_result_expires_idx"),
         ]
+
+
+class PgOrchestrationClaim(models.Model):
```
muhammad-ali-e, line 348: **Type-design nit — the tombstone invariant is enforced only by prose + the *absence* of a clear-function.** The single most important behavioral difference from `claim_batch` (never deleted at finalise) has no structural guard; a future maintainer could add a "cleanup" DELETE and silently break post-completion idempotency. Cheap safeguard: add a `# tombstone: never delete on finalise` marker here, and a one-line note in `clear_execution_batches` explaining there is deliberately no orchestration-claim analog.

workers/tests/test_orchestration_idempotency.py
```diff
@@ -0,0 +1,146 @@
...
+        # The worker bootstraps ``async_execute_bin_general`` under a bare
+        # ``tasks`` module object (a sys.path quirk), which is a DIFFERENT object
+        # than ``from general import tasks``. So patch the claim in the task's OWN
+        # globals (via the unwrapped function), not on our module alias — otherwise
+        # the task would resolve the real claim and hit Postgres. Reaching the skip
+        # dict is itself proof setup was never called (it runs after the gate).
+        fn = tasks.async_execute_bin_general
+        while hasattr(fn, "__wrapped__"):
+            fn = fn.__wrapped__
+        with patch.dict(fn.__globals__, {"claim_orchestration": lambda _eid: False}):
```
muhammad-ali-e: **Comment nit — the `__globals__` patching rationale doesn't hold in this test.** The production fact (the worker bootstraps a bare `tasks` module distinct from `general.tasks` via `spec_from_file_location`) is correct, but this test calls the `general.tasks` copy, so `fn.__globals__` *is* `general.tasks.__dict__` — identical to what the simpler `patch("general.tasks.claim_orchestration")` targets. The worker bootstrap never runs under pytest, so "otherwise the task would resolve the real claim and hit Postgres" isn't true here and could mislead. Keep the robust technique but reframe it as belt-and-braces rather than necessity.

muhammad-ali-e (commented):
Automated PR review (PR Review Toolkit: Code Reviewer, Silent Failure Hunter, Type Design Analyzer, PR Test Analyzer, Comment Analyzer, Code Simplifier).

**2 blockers** (one root cause — the claim is committed before the work and never released on failure): it defeats `autoretry_for` (permanently-lost executions) and strands crash-window executions the reaper can't recover. **2 correctness/observability** issues (a factually-wrong GC docstring; a silent missing-table outage) and **several test/design/comment nits**. See inline comments.

muhammad-ali-e:
## Review feedback addressed (commit 841847df8)

Thanks — the 2 blockers were spot on. SonarCloud is clean; CodeRabbit skipped (draft).

**BLOCKER 1 — committed-before-work claim defeats retries / silently loses the execution → FIXED.**
Added `release_orchestration_claim()`, called on the orchestrator's error path (PG-gated, best-effort). A failed attempt now frees the claim so the redelivery/retry re-orchestrates; the claim persists only after a *successful* orchestration (the post-completion tombstone). New integration tests: `test_release_allows_reclaim`, `test_release_is_safe_when_absent`.

**BLOCKER 2 — crash-window strand not reaper-recoverable (+ the false "same as `pg_batch_dedup`" docstrings) → docstrings corrected; reaper fix filed as follow-up.**
You're right on both facts: the claim is taken *before* the barrier is armed (so a hard crash in the claim→arm window leaves a claim the reaper's barrier sweep can't see), and the **live** `sweep_orphan_dedup` covers only `pg_batch_dedup` — this table has **no GC** today. I corrected all three docstrings to say exactly that. The proper fix (extend the reaper to GC + recover orphan claims, which also needs an `organization_id` column on the claim so the reaper can call the org-scoped status API) touches the merged reaper + a schema change, so I scoped it out to **UN-3679** and noted that the `pg_queue_enabled` gate stays **off** until it lands. Software-exception strands are already covered now by the release-on-failure fix; only the hard-crash window remains, tracked there.

**Correctness/observability + nits — all addressed:**
- **Deploy hazard:** `try_claim_orchestration` catches `UndefinedTable` → re-raises an actionable *"migration 0012 not applied — run migrations before enabling PG transport"* error (no silent table-missing fallback).
- **Observability:** the skip logs at ERROR with `errorId=ORCH_CLAIM_SUPPRESSED_RETRY` when `retries > 0` (a suppressed retry, distinct from a benign duplicate); `self.request.retries` threaded into the gate.
- **Concurrency test:** N threads + `threading.Barrier`, assert exactly one winner over several executions (the contended `ON CONFLICT` path), mirroring the dedup sibling.
- **Error-path test:** `claim raises → _should_skip propagates` (pins the no-swallow contract).
- **Rename:** `claim_orchestration` → `try_claim_orchestration` (kills the double-negative at the call site).
- **Type-design:** tombstone invariant marked in the model (a future maintainer must not add a finalise-time DELETE).
- **Comment:** reframed the `__globals__` patch as belt-and-braces (correct that under pytest `fn.__globals__` *is* `general.tasks.__dict__`).

Unit lane 1137 passed; orchestration integration 7 green vs local Postgres.

muhammad-ali-e:
## Greptile feedback addressed (commit 361d73ea8)

Both P2s were valid — thanks.

**G1 — no stale-connection retry on the claim (first write after idle) ✅ fixed.**
Good catch that this is the *first* DB write of the orchestration task, so it's the one most exposed to a PgBouncer idle-reap. I didn't take the plain one-shot retry, though, because a claim is a `RETURNING` read: a **commit-phase** conn-death is ambiguous (the server may have committed), and re-running would flip a real winner to a loser → strand — which is exactly why `_run_idempotent_pre_dispatch_write` is `-> None` and why `claim_batch` stays retry-free. Instead I added the same **execute-phase-only phase-split** as `_apply_decrement`: retry once only on an execute-phase failure of a *cached* connection (the INSERT provably never committed — non-autocommit rolls back on disconnect — so the retry's answer is authoritative). Commit-phase and fresh-conn failures propagate. Tests mirror the decrement's: reaped-cached-conn retries once; commit-phase not retried; fresh-conn not retried.

**G2 — `release_orchestration_claim` called even when the claim was never acquired ✅ fixed.**
Now tracked with a `claimed_orchestration` flag set only after the gate returns "proceed" on PG (i.e. we actually won the claim). If the claim's own INSERT raises, the flag stays False and the error path skips the release — no spurious reconnect-to-delete-nothing, no misleading warning. Tests: no release when the claim never acquired; release when claimed then the work fails.

Unit lane 1142 passed; orchestration + claim-retry integration green (the one red in the shared integration lane is the pre-existing UN-3676 flake — passes in isolation).

PR #2142, opened 2026-07-02
UN-3679 [FIX] PG reaper — sweep pg_orchestration_claim (GC + crash-window recovery)

muhammad-ali-e:
## What & why

Follow-up to UN-3671, closing the two recovery gaps its review flagged (blocker 2 + the no-GC tombstone). **The `pg_queue_enabled` gate stays off until this lands.**

The orchestration idempotency claim (`pg_orchestration_claim`) is taken **before** the barrier is armed, so — unlike `pg_batch_dedup`, whose barrier always exists — two things the barrier sweep can't handle:
- **Gap 2 (crash-window strand):** a hard crash in the claim→arm window leaves a claim with **no `pg_barrier_state` row**; the reaper's barrier sweep never sees it → the execution strands `EXECUTING` forever.
- **Gap 1 (no GC):** a successful claim's tombstone (kept to block post-completion re-orchestration) has no natural GC — the live `sweep_orphan_dedup` only touches `pg_batch_dedup`.

## Fix

New `sweep_orphan_claims` in the reaper's retention sweep (cadence-gated, its own `_run_sweep`, leader-only). For each claim with **no matching `pg_barrier_state` row** older than the stuck-timeout:
- **terminal** execution → GC the tombstone (delete);
- **non-terminal** → mark ERROR (`mark_execution_error`, +cascade) then delete — the crash-window recovery.

A claim younger than the stuck-timeout, or one whose barrier is armed (a live run), is left alone. The mark re-checks *still-orphan* right before firing (so a slow-but-live orchestration that just armed isn't flagged ERROR). Per-row best-effort — one failure (API down / unreadable status / unconfirmed mark) leaves that row for the next sweep and never blocks the others.

**Correctness note:** a crash-window strand's message is acked on the first skip-redelivery (the gate returns a dict, not a raise), so by the stuck-timeout (~2.5h) no message exists — the reaper can mark+delete without re-orchestration risk.

## Schema

Added `organization_id` to `pg_orchestration_claim` (model + migration `0013`) — the reaper needs it for the org-scoped status/mark API (the row had only `execution_id`). `try_claim_orchestration` stamps it; the gate passes `schema_name`. Kept as `TextField` (no FK), consistent with every PG-queue table (`PgQueueMessage.org_id`, `PgBarrierState.organization_id`): the queue is self-contained infra written via raw SQL, the value is the org identifier string (matches `organization.organization_id`, not the BIGINT PK), and `""` is a valid "no org" sentinel an FK can't reference.

The claim/gate are PG-transport-gated; the sweep is leader-only and only ever touches `pg_orchestration_claim` (a table the Celery path never writes). Celery path untouched.

## Testing (local)

- **Unit lane: 1142 passed.**
- **7 real-Postgres claim-sweep tests:** GC-terminal, recover-crash-window (marks ERROR + cascade), leave-young, leave-armed-barrier, unconfirmed-mark-leaves, no-org-leaves, one-failure-isolation; plus a `_maybe_sweep` wiring assertion. Existing claim/gate tests updated for the org arg.
- Backend `makemigrations --check`: clean.

## Deploy

Apply migration `0013` before enabling PG transport (migrate-first; flag-gated, inert until `pg_queue_enabled`).


workers/queue_backend/pg_barrier.py
```diff
@@ -361,10 +360,11 @@ def try_claim_orchestration(execution_id: str) -> bool:
                 with conn.cursor() as cur:
                     cur.execute(
                         f"INSERT INTO {qualified('pg_orchestration_claim')} "
-                        "(execution_id, claimed_at) VALUES (%s, now()) "
+                        "(execution_id, organization_id, claimed_at) "
```
muhammad-ali-e, line 363: **MEDIUM (correctness + stale comment).** This INSERT now writes `organization_id`, a column added by **migration 0013** in this same PR — but the guard at the bottom of this function (`except psycopg2.errors.UndefinedTable`, ~L406) only catches a **missing table**, and its message still names only `0012_pgorchestrationclaim`.

On a deploy where 0012 ran but **0013 has not**, the table exists and the column is missing → psycopg2 raises `UndefinedColumn` (SQLSTATE 42703), a *sibling* of `UndefinedTable` (42P01). It is **not caught**, so the execution gets exactly the generic per-execution stack trace this guard exists to prevent — and even the caught fresh-install case points at 0012, which alone leaves the column absent.

**Fix:** broaden the `except` to also cover `psycopg2.errors.UndefinedColumn`, and update the message, e.g. *"pg_orchestration_claim (or its organization_id column) is missing — apply pg_queue migrations through 0013_pgorchestrationclaim_organization_id before enabling PG transport."*

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -543,6 +553,212 @@ def recover_expired_barriers(
...
+        except Exception:
+            with contextlib.suppress(Exception):
+                conn.rollback()
+            logger.exception(
+                "Reaper: failed to recover/GC orphan orchestration claim for "
+                "execution %s — leaving the row for the next sweep.",
+                execution_id,
+            )
+
+    removed = gc_count + recovered
```
muhammad-ali-e, line 787: **MEDIUM (silent failure / observability).** Unlike its sibling `recover_expired_barriers` (which tracks a `failed` counter and escalates `logger.error("… likely systemic (internal API down / bad migration)")` when `failed and not recovered`), this sweep has no `failed` counter and no systemic escalation.

When the internal API is down, every row raises inside `_execution_status`, is caught per-row (L740–747), and `gc_count`/`recovered` stay 0 → `removed == 0` → the `if removed:` rollup is skipped **and the function returns 0 normally**. `_run_sweep` treats a clean return as success and **resets `self._sweep_fail_streak` to 0** (L1001), so the consecutive-failure detector — the one signal that would surface "the crash-window recovery net is completely down and executions are stranding" — stays blind.

**Fix:** mirror `recover_expired_barriers`: increment a `failed` counter in the per-row `except`, and after the loop escalate `logger.error(... systemic ...)` when `failed and not removed`, `logger.warning` when `failed` with partial progress, and include `skipped = len(rows) - removed - failed` so the aggregate is visible even when `removed == 0`.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -543,6 +553,212 @@ def recover_expired_barriers(
...
+            f"SELECT 1 FROM {qualified('pg_orchestration_claim')} AS c "
+            f"WHERE c.execution_id = %s AND {_orphan_claim_where('c')}",
+            (execution_id, stuck_timeout_seconds),
+        )
+        found = cur.fetchone() is not None
+    conn.commit()
+    return found
+
+
+def _delete_orphan_claim(
```
muhammad-ali-e, line 604: **LOW/MEDIUM (miscount + lost signal).** `_delete_orphan_claim` returns `None` and never inspects `cur.rowcount`. Its DELETE is re-guarded on `_orphan_claim_where` (still-orphan-and-old), so if a barrier is armed or the claim is released+re-claimed in the window since the re-check, the DELETE matches **0 rows** — yet `_recover_one_claim` still returns `"recovered"`/`"gc"` unconditionally (L660, L686), so `removed` over-reports and nothing records the race.

The barrier sibling does this correctly: `_recover_one_barrier` inspects `cur.rowcount > 0` and emits `logger.warning("… was re-armed during recovery — leaving its rows for the new run")` (L449–460).

**Fix:** have `_delete_orphan_claim` return `cur.rowcount`; in `_recover_one_claim`, treat a 0-row delete as a benign skip (`return None`, don't count) with a warning mirroring the barrier path. Not a lost execution (the barrier sweep picks up a re-armed run next cycle) but a silent miscount of the exact race the code takes pains to guard elsewhere.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -543,6 +553,212 @@ def recover_expired_barriers(
...
+        error_message=(
+            "[reaper-recovery] Execution stranded: the orchestration claimed its "
+            "slot but the barrier was never armed (crash before dispatch)."
+        ),
+    ):
+        # Unconfirmed mark — do NOT delete the claim (it is the only recovery
+        # handle); leave it for the next sweep to retry.
+        return None
+
+    _delete_orphan_claim(conn, execution_id, stuck_timeout_seconds)
```
muhammad-ali-e: **LOW (advisory — document the safety assumption).** The crash-window branch deletes the claim tombstone *after* marking the execution ERROR. Unlike the terminal-GC branch just above (which carries the inline note *"the message was acked on completion or on the first skip-redelivery, so there is nothing left to re-orchestrate once the row is gone"*), this branch has no comment asserting the message is already acked.

Deleting the tombstone re-opens the double-orchestration gate the claim exists to close: if the orchestration message is still un-acked past the stuck-timeout (e.g. the whole worker fleet was down longer than the stuck-timeout, so no consumer ever hit the skip-and-ack path), a later redelivery re-wins the claim and re-orchestrates a now-ERROR execution.

This looks like an accepted design tradeoff, but please confirm the ack-by-stuck-timeout assumption against the queue's redelivery/ack semantics and add a one-line inline comment here documenting why the delete is safe (as the terminal branch does).

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -543,6 +553,212 @@ def recover_expired_barriers(
...
+    conn.commit()
+
+
+def _recover_one_claim(
+    conn: PgConnection,
+    api_client: InternalAPIClient,
+    execution_id: str,
+    organization_id: str,
+    stuck_timeout_seconds: int,
+) -> str | None:
```
muhammad-ali-e: **LOW (type precision).** The return is `str | None` but the only non-`None` values are the literals `"recovered"` and `"gc"`, compared at L734–737. A typo in a returned literal or a caller comparison silently falls through to the skip/`None` path — the row is still deleted inside, but the counter isn't incremented, so `removed` under-counts and the summary log misreports.

**Fix (one line, zero runtime cost):** `def _recover_one_claim(...) -> Literal["recovered", "gc"] | None:`. This gives the type checker the closed domain and aligns with the module's own conventions for closed outcomes (`TickOutcome`, `ExecutionStatus`). Optionally hoist the two literals to module constants shared by producer and consumer.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -543,6 +553,212 @@ def recover_expired_barriers(
...
+    """Recover or GC one orphan claim; return ``"recovered"`` (execution marked
+    ERROR — a crash-window recovery), ``"gc"`` (a terminal execution's tombstone
+    deleted), or ``None`` (a benign skip). The claim row is deleted on GC and on a
+    confirmed recovery; it is LEFT for the next sweep on any unconfirmed step (no
+    org, unreadable status, re-armed during recovery, or an unconfirmed mark) so
+    nothing is lost.
+
+    A terminal execution (COMPLETED / ERROR) → GC the tombstone. A non-terminal one
```
muhammad-ali-e: **LOW (comment accuracy).** The docstring says *"A terminal execution (COMPLETED / ERROR)"* and the inline comment at L651 says *"Completed / failed"*, but the code gates on `ExecutionStatus.is_completed(status)` (L650), which treats **COMPLETED, STOPPED, and ERROR** as terminal. A STOPPED execution's tombstone is also GC'd, which neither phrasing conveys.

**Fix:** mirror the sibling `_recover_one_barrier` docstring, which defers to `is_completed` as the single source of truth — e.g. *"A terminal execution (per `ExecutionStatus.is_completed`: COMPLETED / STOPPED / ERROR) → GC the tombstone"* — so the prose can't drift from the predicate.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -543,6 +553,212 @@ def recover_expired_barriers(
     return recovered
 
 
+def _orphan_claim_where(claim_alias: str) -> str:
```
muhammad-ali-e, line 564: **NIT (optional, consistency).** `_orphan_claim_where(claim_alias)` is parameterized on the alias, but all three call sites pass the literal `'c'` and every query hardcodes `AS c` (L587–589, L605–607, L716–718). Its own docstring says it's defined *"like `_STRANDED_PREDICATE` for barriers"* — yet `_STRANDED_PREDICATE` (L100) is a plain module-level string constant.

Collapsing this to a `_ORPHAN_CLAIM_WHERE` constant would produce byte-identical SQL (every caller already aliases `AS c`) and match the very sibling it cites. Purely a judgment call — the alias-parameter form is a defensible "explicit over implicit" choice; only worth it if the team values parity with `_STRANDED_PREDICATE`.

muhammad-ali-e (commented):

Ran six specialized agents — Code Reviewer, Silent Failure Hunter, Type Design Analyzer, PR Test Analyzer, Comment Analyzer, Code Simplifier. Overall this is a careful, well-tested extension that closely follows the existing reaper/barrier patterns; **no blocking defects**. Findings below are graded MEDIUM → NIT and posted inline. Highlights:

- **MEDIUM** — migration guard misses the new `organization_id` column (0012-but-not-0013 deploys raise an uncaught `UndefinedColumn`).
- **MEDIUM** — the orphan-claim sweep has no systemic-failure escalation and its clean `return 0` on a total outage resets `_run_sweep`'s fail-streak, blinding the outage detector (diverges from `recover_expired_barriers`).
- **LOW/MED** — `_delete_orphan_claim` ignores `rowcount`, so a raced no-op delete over-counts and drops the re-armed warning the barrier path emits.
- Plus low/nit items (return-type precision, terminal-status wording, `_orphan_claim_where` as a constant) and several test-coverage gaps on the race/rollback branches.

workers/tests/test_pg_reaper.py
```diff
@@ -1159,5 +1174,138 @@ def test_main_module_reexports_real_main(self):
...
+def _claim_ids(conn):
+    with conn.cursor() as cur:
+        cur.execute("SELECT execution_id FROM pg_orchestration_claim ORDER BY 1")
+        rows = [r[0] for r in cur.fetchall()]
+    conn.commit()
+    return rows
+
+
+@pytest.mark.integration
+class TestSweepOrphanClaims:
```
muhammad-ali-e, line 1228: **Test coverage — the race/rollback branches this PR exists to defend are not directly exercised.** Happy paths and static-state skips are solid, but the `on_get` hook already used by the barrier suite (`TestRecoverExpiredBarriers`) makes the missing cases cheap to port:

1. **`_claim_still_orphan` re-check race (reaper.py L665–671, rate 8).** `test_leaves_claim_with_armed_barrier` seeds the barrier *before* the sweep, so the row is filtered by the SELECT and never reaches the re-check. Add a test where `on_get` INSERTs a `pg_barrier_state` row (or re-claims with a fresh `claimed_at`) *between* SELECT and mark; assert `removed == 0`, `update_calls == []`, row left.
2. **`_delete_orphan_claim` re-guard (L596–609, rate 8).** No test drives a fresh `claimed_at`/just-armed barrier into the window between a confirmed mark and the delete; assert the re-claimed row survives.
3. **SELECT-level rollback + re-raise (L723–725, rate 7).** Mirror `test_rolls_back_on_select_error`: `cur.execute.side_effect = OperationalError`, assert it raises, `conn.rollback` once, `conn.commit` not called.
4. **Cadence-gating / independence (rate 6).** `test_cadence_gates_repeat_within_interval` (L424) and `test_one_sweep_failing_does_not_starve_the_other` (L448) never reference `stub_retention_sweep.claims`; add `claims.call_count`/`claims.assert_called_once()` assertions to pin the independence the production comment (L958–968) claims.
5. **GC only tested with COMPLETED (L650, rate 3).** Parametrize over `("COMPLETED", "ERROR", "STOPPED")` so a terminal ERROR/STOPPED GCs and does not re-mark.
6. **"success read but status is None" (L642–648, rate 3).** `_FakeApiClient(status=None)` → assert `removed == 0`, no mark, row left (barrier path has `test_status_none_on_success_does_not_mark`).

muhammad-ali-e:
## Review feedback addressed (commit 3cb280b39)

Thanks — good, sharp findings. SonarCloud clean; CodeRabbit skipped (non-default base).

**MEDIUM — `UndefinedColumn` not caught (0012-but-not-0013 deploy) ✅ fixed.**
Broadened `try_claim_orchestration`'s guard to `(UndefinedTable, UndefinedColumn)` and updated the message to name both migrations. A deploy with the table but no `organization_id` column now fails fast with the actionable error instead of a generic per-execution trace. Test: parametrized over both SQLSTATEs.

**MEDIUM — sweep had no systemic-failure signal ✅ fixed.**
Added a `failed` counter; `sweep_orphan_claims` now **raises** when a non-empty sweep swept nothing because every row raised (systemic — API down / bad migration), so `_run_sweep` records it on the consecutive-failure streak rather than a clean return resetting it and hiding "the recovery net is down". Partial failure → warning. Mirrors `recover_expired_barriers`. Test: `test_all_rows_failing_raises_systemic`.

**MEDIUM — `_delete_orphan_claim` ignored `rowcount` (both reviewers) ✅ fixed.**
It now returns `cur.rowcount`; both the GC and recovery branches treat a 0-row delete (a concurrent release+re-claim or barrier arm won the window) as a benign skip + warning, so `removed` never over-reports. Tests: `test_delete_reguard_reclaim_leaves_fresh_claim` (GC path) + the existing recovery path.

**LOW — all addressed.**
- Documented the ack-safety assumption on the crash-window delete branch, incl. the accepted residual (a whole-fleet outage longer than the stuck-timeout could leave the message un-acked — a tradeoff for a catastrophic multi-hour outage).
- `_recover_one_claim` return typed `Literal["recovered", "gc"] | None` with shared module constants (`_CLAIM_RECOVERED` / `_CLAIM_GC`) used at both producer and consumer.
- GC docstring/comment now defers to `ExecutionStatus.is_completed` (COMPLETED / STOPPED / ERROR) as the single source of truth.

**Test coverage — the race branches are now directly exercised:** re-check race (barrier armed between SELECT and mark → not marked, row left) and the delete re-guard (re-claim between SELECT and DELETE → 0-row delete, fresh claim survives), using the `on_get` hook like the barrier suite.

**NIT (declined, with rationale):** keeping `_orphan_claim_where(alias)` as a function rather than a `_STRANDED_PREDICATE`-style constant — the alias parameter documents the correlation explicitly, and all three call sites reading `AS c` makes the intent clear at each site. Happy to collapse it to a constant if you'd rather match the sibling exactly.

Unit lane 1144; reaper + claim integration green.

PR #2143, opened 2026-07-02
UN-3672 [FEAT] PG queue — application-level /metrics exporter (queue depth, message age, barriers, reaper outcomes)

muhammad-ali-e:
## What

**UN-3672** — the PG queue path emitted **zero metrics**; detecting an outage meant "a user complains". This adds an **application-level Prometheus `/metrics` exporter** on the existing health server of every PG-queue process. Application-level ONLY by design — no chart/infra changes; the scrape config (PodMonitoring) + alert rules are deferred to a follow-up pending infra alignment. Until then the endpoint is curl-able (incident debugging; the load-test harness consumes it directly), but nothing collects it in cloud.

## Why the exporter must be ours

Today's queue observability comes from **RabbitMQ's built-in Prometheus plugin** (scraped via a GMP PodMonitoring). Postgres has no such plugin — the pg_queue tables ARE the broker — so replacing RabbitMQ means replacing its metrics source too. This is that replacement.

## Design

- **`/metrics` on the existing LivenessServer** — same port as `/health` (consumer `WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT`, reaper `WORKER_PG_REAPER_HEALTH_PORT`), opt-in via `metrics_fn`, kept as an opaque callable so `liveness.py` stays prometheus-free. A broken renderer 500s `/metrics` only — the `/health` verdict is untouched (tested).
- **Per-pod (every PG worker):** `pg_consumer_heartbeat_age_seconds` — the same poll-loop freshness `/health` verdicts on, as a scrapeable number. The fleet supervisor adds `pg_consumer_alive_children` / `pg_consumer_configured_concurrency`.
- **Queue-wide (reaper only — the leader-elected singleton, so N pods don't run identical SQL and emit duplicate series):**
  - `pg_queue_depth{queue}` + `pg_queue_oldest_message_age_seconds{queue}`
  - `pg_barrier_live` / `pg_barrier_stranded` (same stranded predicate the recovery uses)
  - `pg_reaper_barrier_recovered_total`, `pg_reaper_claim_{recovered,gc}_total`, `*_recovery_failures_total`, `pg_reaper_sweep_failures_total{table}`, `pg_reaper_is_leader`
- **Cached snapshots, not scrape-time SQL:** the leader refreshes the queue gauges every 60s inside its tick; a scrape (or a curl loop mid-incident) never touches the DB. `pg_queue_gauges_age_seconds` exposes snapshot staleness. Snapshot is cleared on losing leadership (a standby must not export a frozen stale snapshot); a refresh failure is counted + logged and never fails the tick; a drained queue's series drops out instead of freezing at its last non-zero value.
- **Instance-owned registries** (never the `prometheus_client` default `REGISTRY`): the workers tree imports under two module names in places (the bare-`tasks` sys.path quirk) — module-level collectors would double-register.
- `prometheus-client` was already a declared (unused) dependency — no new deps. `OPERATIONS.md`'s false "metrics endpoint" claim replaced with the real endpoints.

## Celery safety

All changes live in PG-only modules (`queue_backend/pg_queue/*`, `pg_queue_consumer/`) that Celery workers never import. The full workers unit lane (1192 tests) passes untouched.

## Testing

- **23 new tests**: route isolation (404/200/500, `/health` unaffected), snapshot set/clear/drain semantics, counter threading through the recovery/sweep functions, tick wiring (cadence-gated, standby-never-refreshes, failure-never-fails-the-tick, leadership-loss clears), end-to-end HTTP on both server subclasses.
- **1192 unit + 123 integration** (real Postgres) pass. (The 1 pre-existing flaky `test_exactly_one_task_fires_the_callback` passes solo — known UN-3676 shared-lane race.)
- **Live smoke against the real dev Postgres**: rendered real values — and immediately surfaced a 15-day-old stranded dev `notifications` backlog (41 messages), the exact silent-failure class this exporter exists to catch:

```
pg_reaper_is_leader 1.0
pg_queue_depth{queue="notifications"} 41.0
pg_queue_oldest_message_age_seconds{queue="notifications"} 1.328512e+06
pg_barrier_live 0.0
pg_barrier_stranded 0.0
```

## Out of scope (deferred follow-up)

- GMP PodMonitoring scrape config in the cloud chart + Cloud Monitoring alert rules (infra alignment).
- Grafana dashboards.

### DIFF
--- a/workers/OPERATIONS.md
+++ b/workers/OPERATIONS.md
@@ -70,11 +70,33 @@ curl http://localhost:8083/health  # Callback worker
 
 ### Metrics (Prometheus)
 
-Workers expose metrics on their health ports at `/metrics`:
-- Task execution counts
-- Processing times
-- Queue depths
-- Error rates
+**PG-queue processes** serve Prometheus text-format metrics at `/metrics` on
+their existing health port (`WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT` for
+consumers, `WORKER_PG_REAPER_HEALTH_PORT` for the reaper — same server as
+`/health`; no port configured means neither endpoint):
+
+- Every PG worker (per-pod): `pg_consumer_heartbeat_age_seconds` (poll-loop
+  freshness — the same signal `/health` verdicts on); the fleet supervisor adds
+  `pg_consumer_alive_children` / `pg_consumer_configured_concurrency`.
+- Reaper only (the leader-elected singleton — queue-WIDE state comes from one
+  process): `pg_queue_depth{queue}`, `pg_queue_oldest_message_age_seconds{queue}`,
+  `pg_barrier_live` / `pg_barrier_stranded`, recovery outcome counters
+  (`pg_reaper_barrier_recovered_total`, `pg_reaper_claim_*_total`,
+  `pg_reaper_sweep_failures_total{table}`) and `pg_reaper_is_leader`.
+  Queue gauges are cached snapshots refreshed on the reaper's own cadence —
+  scrapes never touch the DB; `pg_queue_gauges_age_seconds` exposes staleness.
+
+```bash
+curl http://localhost:8090/metrics  # PG consumer / supervisor
+curl http://localhost:8086/metrics  # PG reaper
+```
+
+Nothing scrapes these in cloud yet (no PodMonitoring/alerts — deferred);
+they're for direct curl during incidents, dashboards-to-come, and the
+load-test harness.
+
+**Celery workers** expose no app-level `/metrics`; their queue observability
+comes from RabbitMQ's Prometheus plugin and Flower (below).
 
 ### Logging
 
--- a/workers/OPERATIONS.md
+++ b/workers/OPERATIONS.md
@@ -75,16 +75,22 @@ their existing health port (`WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT` for
 consumers, `WORKER_PG_REAPER_HEALTH_PORT` for the reaper — same server as
 `/health`; no port configured means neither endpoint):
 
-- Every PG worker (per-pod): `pg_consumer_heartbeat_age_seconds` (poll-loop
-  freshness — the same signal `/health` verdicts on); the fleet supervisor adds
-  `pg_consumer_alive_children` / `pg_consumer_configured_concurrency`.
+- Every PG consumer / the fleet supervisor (per-pod):
+  `pg_consumer_heartbeat_age_seconds` (poll-loop freshness — the same signal
+  `/health` verdicts on); the supervisor adds `pg_consumer_alive_children` /
+  `pg_consumer_configured_concurrency`.
 - Reaper only (the leader-elected singleton — queue-WIDE state comes from one
-  process): `pg_queue_depth{queue}`, `pg_queue_oldest_message_age_seconds{queue}`,
-  `pg_barrier_live` / `pg_barrier_stranded`, recovery outcome counters
-  (`pg_reaper_barrier_recovered_total`, `pg_reaper_claim_*_total`,
-  `pg_reaper_sweep_failures_total{table}`) and `pg_reaper_is_leader`.
-  Queue gauges are cached snapshots refreshed on the reaper's own cadence —
-  scrapes never touch the DB; `pg_queue_gauges_age_seconds` exposes staleness.
+  process): `pg_reaper_heartbeat_age_seconds` (tick-loop freshness),
+  `pg_queue_depth{queue}`, `pg_queue_oldest_message_age_seconds{queue}`,
+  `pg_barrier_live` / `pg_barrier_stranded`, outcome + failure counters
+  (`pg_reaper_barrier_*_total`, `pg_reaper_claim_*_total`,
+  `pg_reaper_sweep_failures_total{table}`, `pg_reaper_tick_failures_total`,
+  `pg_reaper_gauge_refresh_failures_total`) and `pg_reaper_is_leader`.
+  Queue gauges are cached snapshots refreshed on the reaper's own cadence
+  (60s — `_GAUGE_REFRESH_INTERVAL_SECONDS` in `reaper.py`) — scrapes never
+  touch the DB; `pg_queue_gauges_age_seconds` exposes snapshot staleness and
+  keeps growing while refreshes fail or the pod is a standby (alert on it
+  together with `pg_reaper_is_leader == 1`).
 
 ```bash
 curl http://localhost:8090/metrics  # PG consumer / supervisor
--- a/workers/pg_queue_consumer/supervisor.py
+++ b/workers/pg_queue_consumer/supervisor.py
@@ -450,6 +450,13 @@ def _maybe_start_supervisor_health(fleet: _Fleet) -> LivenessServer | None:
             "liveness_probe_bound": True,
         }
 
+    from queue_backend.pg_queue.metrics import ConsumerMetrics
+
+    metrics = ConsumerMetrics(
+        freshness_fn=fleet.freshness,
+        alive_children_fn=lambda: float(fleet.alive_count()),
+        concurrency_fn=lambda: float(fleet.concurrency),
+    )
     server = LivenessServer(
         freshness_fn=fleet.freshness,
         stale_after=stale_after,
@@ -457,6 +464,7 @@ def _maybe_start_supervisor_health(fleet: _Fleet) -> LivenessServer | None:
         check_name="pg_queue_fleet",
         age_key="oldest_child_seconds_since_poll",
         extra_status_fn=_extra_status,
+        metrics_fn=metrics.render,
         thread_name="pg-supervisor-liveness",
         log_label="pg-queue supervisor",
     )
--- a/workers/queue_backend/pg_queue/liveness.py
+++ b/workers/queue_backend/pg_queue/liveness.py
@@ -39,9 +39,19 @@ class LivenessServer:
     ``check_name`` / ``age_key`` label the JSON payload; ``extra_status_fn``
     (optional) merges extra fields in (informational — it never affects the
     200/503 verdict, which is purely the freshness heartbeat).
+
+    ``metrics_fn`` (optional) additionally serves ``/metrics``: it returns the
+    Prometheus text-exposition body (bytes). Kept as an opaque callable so this
+    module stays free of the prometheus dependency; the metric definitions live
+    in :mod:`queue_backend.pg_queue.metrics`. A ``metrics_fn`` failure returns
+    500 on ``/metrics`` only — it can never affect the ``/health`` verdict.
     """
 
     _PATHS = frozenset({"/health", "/healthz", "/livez"})
+    _METRICS_PATH = "/metrics"
+    # Prometheus text exposition format (metrics.METRICS_CONTENT_TYPE — inlined
+    # so this module keeps zero imports from the metrics side).
+    _METRICS_CONTENT_TYPE = "text/plain; version=0.0.4; charset=utf-8"
 
     def __init__(
         self,
@@ -52,6 +62,7 @@ class LivenessServer:
         check_name: str,
         age_key: str,
         extra_status_fn: Callable[[], dict[str, Any]] | None = None,
+        metrics_fn: Callable[[], bytes] | None = None,
         thread_name: str = "pg-queue-liveness",
         log_label: str = "pg-queue",
     ) -> None:
@@ -66,6 +77,7 @@ class LivenessServer:
         self._check_name = check_name
         self._age_key = age_key
         self._extra_status_fn = extra_status_fn
+        self._metrics_fn = metrics_fn
         self._thread_name = thread_name
         # Prefixes the (now-shared) log messages so they stay attributable to the
         # source process after the consumer/reaper extraction (e.g. "pg-queue
@@ -86,6 +98,9 @@ class LivenessServer:
         freshness_fn = self._freshness_fn
         stale_after = self._stale_after
         paths = self._PATHS
+        metrics_path = self._METRICS_PATH
+        metrics_content_type = self._METRICS_CONTENT_TYPE
+        metrics_fn = self._metrics_fn
         check_name = self._check_name
         age_key = self._age_key
         extra_status_fn = self._extra_status_fn
@@ -94,7 +109,11 @@ class LivenessServer:
         class _Handler(BaseHTTPRequestHandler):
             def do_GET(self) -> None:
                 # Strip any query string — a probe like /health?foo=bar must match.
-                if urlsplit(self.path).path not in paths:
+                path = urlsplit(self.path).path
+                if metrics_fn is not None and path == metrics_path:
+                    self._serve_metrics()
+                    return
+                if path not in paths:
                     self.send_response(404)
                     self.end_headers()
                     return
@@ -125,6 +144,24 @@ class LivenessServer:
                     self.wfile.write(body)
                 except (BrokenPipeError, ConnectionResetError):
                     pass  # client (probe) hung up mid-response — not our problem
+
+            def _serve_metrics(self) -> None:
+                # A broken metrics renderer must degrade to a 500 on /metrics
+                # alone — the probe verdict on /health stays untouched.
+                try:
+                    body = metrics_fn()  # type: ignore[misc]  # guarded by caller
+                except Exception:
+                    logger.exception("%s: /metrics render failed", log_label)
+                    self.send_response(500)
+                    self.end_headers()
+                    return
+                self.send_response(200)
+                self.send_header("Content-Type", metrics_content_type)
+                self.end_headers()
+                try:
+                    self.wfile.write(body)
+                except (BrokenPipeError, ConnectionResetError):
+                    pass  # scraper hung up mid-response — not our problem
 
             def log_message(self, *_: object) -> None:
                 pass  # silence per-request access logging
--- a/workers/queue_backend/pg_queue/liveness.py
+++ b/workers/queue_backend/pg_queue/liveness.py
@@ -21,6 +21,7 @@ Bind ``port=0`` to let the OS pick a free port (read back via :attr:`bound_port`
 
 from __future__ import annotations
 
+import contextlib
 import logging
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
@@ -137,10 +138,13 @@ class LivenessServer:
                     }
                 )
                 body = json.dumps(payload).encode()
-                self.send_response(503 if stale else 200)
-                self.send_header("Content-Type", "application/json")
-                self.end_headers()
+                # Headers too, not just the body write: a client that hangs up
+                # before headers go out would otherwise raise into socketserver's
+                # handle_error, which prints an unattributed traceback to stderr.
                 try:
+                    self.send_response(503 if stale else 200)
+                    self.send_header("Content-Type", "application/json")
+                    self.end_headers()
                     self.wfile.write(body)
                 except (BrokenPipeError, ConnectionResetError):
                     pass  # client (probe) hung up mid-response — not our problem
@@ -152,13 +156,15 @@ class LivenessServer:
                     body = metrics_fn()  # type: ignore[misc]  # guarded by caller
                 except Exception:
                     logger.exception("%s: /metrics render failed", log_label)
-                    self.send_response(500)
-                    self.end_headers()
+                    with contextlib.suppress(BrokenPipeError, ConnectionResetError):
+                        self.send_response(500)
+                        self.end_headers()
                     return
-                self.send_response(200)
-                self.send_header("Content-Type", metrics_content_type)
-                self.end_headers()
+                # Same hung-up-client guard as the /health path above.
                 try:
+                    self.send_response(200)
+                    self.send_header("Content-Type", metrics_content_type)
+                    self.end_headers()
                     self.wfile.write(body)
                 except (BrokenPipeError, ConnectionResetError):
                     pass  # scraper hung up mid-response — not our problem
--- a/workers/queue_backend/pg_queue/metrics.py
+++ b/workers/queue_backend/pg_queue/metrics.py
@@ -1,0 +1,246 @@
+"""Prometheus metrics for PG-queue processes (application-level exporter).
+
+Today's queue observability comes from RabbitMQ's built-in Prometheus plugin;
+Postgres has no such plugin — the pg_queue tables ARE the broker — so the
+exporter is ours. This module holds the metric *definitions*; the SQL that
+feeds the queue-wide gauges lives with its sibling queries in ``reaper.py``
+(it shares the stranded-barrier predicate), and the HTTP surface is the
+existing :class:`~queue_backend.pg_queue.liveness.LivenessServer` ``/metrics``
+route — no new port, no new server, no execution-path changes.
+
+Two exporters, matching the two process shapes:
+
+- :class:`ConsumerMetrics` — per-pod, on every PG worker: poll-loop heartbeat
+  freshness (the same signal ``/health`` verdicts on, as a scrapeable number).
+- :class:`ReaperMetrics` — queue-WIDE state, exported only by the reaper: it is
+  the leader-elected singleton, so queue depth / oldest-message age / barrier
+  counts come from one process instead of N pods running identical SQL and
+  emitting duplicate series. Standbys export zeroed queue gauges and
+  ``is_leader 0``.
+
+Registries are instance-owned (never the ``prometheus_client`` default
+``REGISTRY``): the workers tree is imported under more than one module name in
+places (the bare-``tasks`` sys.path quirk), and module-level collectors would
+double-register on the second import. An instance per process cannot.
+"""
+
+from __future__ import annotations
+
+import logging
+import time
+from collections.abc import Callable
+from typing import TYPE_CHECKING, Final
+
+if TYPE_CHECKING:
+    from prometheus_client import CollectorRegistry
+
+logger = logging.getLogger(__name__)
+
+# Prometheus text exposition format (the /metrics Content-Type).
+METRICS_CONTENT_TYPE: Final = "text/plain; version=0.0.4; charset=utf-8"
+
+
+def _new_registry() -> CollectorRegistry:
+    from prometheus_client import CollectorRegistry
+
+    return CollectorRegistry()
+
+
+def _render(registry: CollectorRegistry) -> bytes:
+    from prometheus_client import generate_latest
+
+    return generate_latest(registry)
+
+
+class ConsumerMetrics:
+    """Per-pod metrics for a PG-queue consumer (or the fleet supervisor).
+
+    ``freshness_fn`` is the same heartbeat the liveness probe reads —
+    seconds since the poll loop last made progress (for the supervisor, the
+    OLDEST child's, so one wedged child surfaces). The optional fleet hooks
+    exist because the supervisor's ``/health`` JSON already reports them and
+    an operator graphing the fleet needs them as numbers, not JSON.
+    """
+
+    def __init__(
+        self,
+        *,
+        freshness_fn: Callable[[], float],
+        alive_children_fn: Callable[[], float] | None = None,
+        concurrency_fn: Callable[[], float] | None = None,
+    ) -> None:
+        from prometheus_client import Gauge
+
+        self.registry = _new_registry()
+        heartbeat = Gauge(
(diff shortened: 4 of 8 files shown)


workers/queue_backend/pg_queue/reaper.py
```diff
@@ -951,13 +1023,17 @@ def tick(self) -> TickOutcome:
...
                 raise
             if not still_leader:
                 logger.warning(
                     "Reaper: lost leadership (lease taken over) — stepping down "
                     "to standby"
                 )
                 self._is_leader = False
+                # A standby must not keep exporting a frozen queue snapshot as if
+                # it were live — the new leader owns those series now.
+                self._metrics.clear_queue_snapshot()
```
muhammad-ali-e: **[Critical — silent-failure-hunter, code-reviewer, type-design-analyzer]** Leadership flap → up to 60s of a leader exporting a false "empty queue".

`clear_queue_snapshot()` wipes the gauges and resets `ReaperMetrics._last_refresh_monotonic`, but the reaper's own cadence gate `self._last_gauge_refresh_monotonic` is **not** reset. If this process re-acquires leadership within `_GAUGE_REFRESH_INTERVAL_SECONDS` (a normal lease-flap scenario — the renew-raise at L1025-1027 is caught by `run()` and the loop continues), `_maybe_refresh_gauges` gates out, so the just-cleared gauges are not repopulated. During that window the process exports `pg_reaper_is_leader 1`, `pg_barrier_live 0`, `pg_barrier_stranded 0`, no `pg_queue_depth` series, and `pg_queue_gauges_age_seconds 0` — a leader reporting a fresh-looking empty queue while an arbitrary backlog / stranded barriers may exist. The docstring's own disambiguation rule ("age 0 → check `pg_reaper_is_leader`") fails here because is_leader is 1, and it contradicts the "first leader tick refreshes immediately" comment near L957.

**Fix:** reset the gate wherever the snapshot is cleared — e.g. a small `_step_down()` helper doing `self._metrics.clear_queue_snapshot(); self._last_gauge_refresh_monotonic = None`, called at both this site and L1026.

*Deeper fix (type-design):* `ReaperMetrics` already owns `is_leader_fn` — enforcing "standby exports no snapshot" at collect time (custom collector) would make the frozen-standby export unrepresentable and delete both call sites, covering any future code path that flips `_is_leader` without remembering to clear.

workers/queue_backend/pg_queue/metrics.py
```diff
@@ -0,0 +1,246 @@
...
+        gauges_age = Gauge(
+            "pg_queue_gauges_age_seconds",
+            "Seconds since the queue gauges were last refreshed (0 = never; "
+            "check pg_reaper_is_leader)",
+            registry=self.registry,
+        )
+        gauges_age.set_function(self._seconds_since_refresh)
+
+    def _seconds_since_refresh(self) -> float:
+        if self._last_refresh_monotonic is None:
```
muhammad-ali-e: **[Critical — type-design-analyzer]** TOCTOU race: `self._last_refresh_monotonic` is read twice (None check, then arithmetic). The reaper tick thread calls `clear_queue_snapshot()` (sets it to `None`) while the HTTP scrape thread executes this callback inside `generate_latest`; interleaving between check and use raises `TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'`, which fails the **whole** render and turns the scrape into a 500 — during leadership churn, exactly when an operator is watching.

**Fix:** read once into a local:
```python
def _seconds_since_refresh(self) -> float:
    last = self._last_refresh_monotonic
    return 0.0 if last is None else time.monotonic() - last
```

workers/queue_backend/pg_queue/metrics.py
```diff
@@ -0,0 +1,246 @@
...
+        )
+        self._barriers_stranded = Gauge(
+            "pg_barrier_stranded",
+            "Live barriers past the stuck-timeout / expiry (reaper will recover)",
+            registry=self.registry,
+        )
+        self._last_refresh_monotonic: float | None = None
+        gauges_age = Gauge(
+            "pg_queue_gauges_age_seconds",
+            "Seconds since the queue gauges were last refreshed (0 = never; "
```
muhammad-ali-e: **[Important — silent-failure-hunter]** `pg_queue_gauges_age_seconds` — the metric whose entire purpose is exposing staleness — reports **maximal freshness (0)** in the maximal-staleness case:

- A leader whose `refresh_queue_gauges` has failed since process start (missing grant, bad migration) exports age `0` forever; a staleness alert `pg_queue_gauges_age_seconds > X` can never fire. Only `pg_reaper_gauge_refresh_failures_total` reveals it.
- Combined with `clear_queue_snapshot()` zeroing the barrier gauges, a fleet with **no leader at all** (stuck lease) exports "0 stranded barriers, age 0" from every pod — the stranded-barrier alert goes dark exactly when the recovery net is down.

A documented ambiguous sentinel is still an ambiguous sentinel that defeats alerting.

**Fix:** don't overload 0. Either return `float("nan")` when `_last_refresh_monotonic is None` (NaN comparisons are always false — distinguishable from fresh), or follow Prometheus convention and export `pg_queue_gauges_last_refresh_timestamp_seconds` (unix time; `time() - metric > X` fires naturally on "never"). Consider likewise leaving the barrier gauges unset on standby instead of 0. Note the wording nit too: "0 = never" is also wrong for a demoted ex-leader (cleared, not never-taken).

workers/queue_backend/pg_queue/metrics.py
```diff
@@ -0,0 +1,246 @@
...
+        barriers_live: int,
+        barriers_stranded: int,
+    ) -> None:
+        """Replace the cached queue-wide snapshot.
+
+        ``depths`` maps queue name -> (message count, oldest-message age in
+        seconds). Series are cleared first so a queue that drained to zero
+        drops out rather than freezing at its last non-zero value.
+        """
+        self._queue_depth.clear()
```
muhammad-ali-e: **[Important — type-design-analyzer, code-reviewer, silent-failure-hunter, pr-test-analyzer]** The snapshot swap is not atomic with respect to scrapes. `prometheus_client` locks per metric, not across metrics: a scrape landing between `clear()` and the repopulation loop sees empty/partial `pg_queue_depth` (a transient false "queue drained" sample), and one landing between the loop and the barrier sets sees new depths paired with old barrier counts and the previous refresh age. Same torn window in `clear_queue_snapshot`. Rare per scrape, but it fires eventually at scale and produces unexplainable dashboard dips.

**Fix options** (in increasing robustness):
1. Set-then-prune: set new values first, then `remove()` only the label sets absent from the new snapshot.
2. Hold the snapshot as an immutable frozen dataclass swapped by a single attribute assignment (GIL-atomic) and export via a custom `Collector` on the instance registry — collect reads one consistent object, `clear` becomes swapping in an empty snapshot. This also subsumes the `_seconds_since_refresh` TOCTOU fix.

*Pairs well with:* `depths: dict[str, tuple[int, float]]` uses a positional tuple — the builder in `reaper.py` could swap depth/oldest_age without a type error. A small `NamedTuple QueueStat(depth: int, oldest_age_seconds: float)` makes the call site keyword-checked.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -969,6 +1045,7 @@ def tick(self) -> TickOutcome:
                     self._get_sweep_conn(),
                     self._get_api_client(),
                     self._stuck_timeout_seconds,
+                    metrics=self._metrics,
```
muhammad-ali-e, line 1049: **[Important — silent-failure-hunter]** Whole-tick recovery failures have no metric. This PR adds failure counters for per-row barrier recovery, per-row claim recovery, retention sweeps, and gauge refreshes — but the most severe failure mode has none: `recover_expired_barriers`'s initial SELECT raising (re-raise → tick raises here), or `dispatch_due_schedules` raising. These propagate to `run()`, which logs "cycle failed; continuing" and loops. Meanwhile `_last_tick_monotonic` is set at tick **start**, so `pg_reaper_heartbeat_age_seconds` stays fresh and `/health` stays 200 while every tick fails and zero recovery work happens. A schema/permission problem confined to `pg_barrier_state` would be invisible in every metric this exporter ships (the gauge-refresh queries could still succeed).

**Fix:** add a `pg_reaper_tick_failures_total` counter to `ReaperMetrics` and increment it in `run()`'s `except Exception` alongside the existing `logger.exception`.

workers/queue_backend/pg_queue/metrics.py
```diff
@@ -0,0 +1,246 @@
...
+        )
+        self._oldest_age = Gauge(
+            "pg_queue_oldest_message_age_seconds",
+            "Age of the oldest message in the queue (cached snapshot)",
+            ["queue"],
+            registry=self.registry,
+        )
+        self._barriers_live = Gauge(
+            "pg_barrier_live",
+            "pg_barrier_state rows with remaining > 0 (in-flight fan-outs)",
```
muhammad-ali-e: **[Important — comment-analyzer, code-reviewer]** Help-text/SQL mismatch: `pg_barrier_live` claims "pg_barrier_state rows with **remaining > 0**", but the feeding query in `refresh_queue_gauges` is an unfiltered `count(*)` over `pg_barrier_state`. Completed barriers are deleted synchronously in the happy path, so the counts usually match — but delete-failure lingerers (`remaining == 0`, reclaimed only at expiry) and abandoned rows (`remaining < 0`) are counted as "live", making the gauge overstate in-flight fan-outs precisely during failure conditions, and an operator debugging a nonzero gauge against a `remaining > 0` query sees a discrepancy.

**Fix:** either add `count(*) FILTER (WHERE remaining > 0)` to the SQL (keeping the stranded `FILTER`) so the metric matches its documented semantics, or reword the help to "all pg_barrier_state rows (in-flight fan-outs; may include torn-down rows awaiting cleanup)".

workers/OPERATIONS.md
```diff
@@ -70,11 +70,33 @@ curl http://localhost:8083/health  # Callback worker
...
-- Task execution counts
-- Processing times
-- Queue depths
-- Error rates
+**PG-queue processes** serve Prometheus text-format metrics at `/metrics` on
+their existing health port (`WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT` for
+consumers, `WORKER_PG_REAPER_HEALTH_PORT` for the reaper — same server as
+`/health`; no port configured means neither endpoint):
+
+- Every PG worker (per-pod): `pg_consumer_heartbeat_age_seconds` (poll-loop
```
muhammad-ali-e: **[Important — comment-analyzer]** "Every PG worker (per-pod): `pg_consumer_heartbeat_age_seconds`" is false for the reaper, which the section intro ("PG-queue processes") includes. The reaper exports `pg_reaper_heartbeat_age_seconds`, and that metric isn't documented anywhere in this section — an operator curling the reaper's `:8086/metrics` per the example below will not find the metric this line promises.

**Fix:** scope this bullet to "Every PG consumer / the fleet supervisor", and add `pg_reaper_heartbeat_age_seconds` to the reaper bullet.

workers/tests/test_pg_metrics.py
```diff
@@ -0,0 +1,389 @@
...
+    def test_leader_tick_refreshes_gauges(self):
+        reaper = self._reaper(_FakeLease(acquires=True))
+        with (
+            patch.object(reaper_mod, "recover_expired_barriers", return_value=[]),
+            patch.object(reaper_mod, "refresh_queue_gauges") as refresh,
+        ):
+            reaper.tick()
+        refresh.assert_called_once()
+
+    def test_refresh_is_cadence_gated(self):
```
muhammad-ali-e: **[Critical test gap — pr-test-analyzer]** Only the *suppression* half of the cadence gate is tested (two immediate ticks → one refresh). Two untested behaviors:

1. **Resumption:** no test that a refresh happens *again* once the interval elapses. A bug in the gate condition (inverted comparison, wrong sentinel handling, unit mix-up) would pass every existing test — the first refresh fires via the `None` sentinel — while production queue gauges silently freeze forever after the first snapshot.
2. **Failure consumes the interval:** the documented "cadence is advanced BEFORE the read so a persistent failure retries once per interval rather than every tick" is untested. Moving the timestamp assignment after the `try` (the "obvious" success-only placement) would make every ~5s tick hammer a down DB, and no test would fail.

**Fix:** (1) after a leader tick, rewind `reaper._last_gauge_refresh_monotonic -= _GAUGE_REFRESH_INTERVAL_SECONDS + 1` and assert a second `refresh_queue_gauges` call on the next tick (include the exact-boundary case). (2) after the failing tick, tick again with the patch still failing and assert exactly one call total.

workers/tests/test_pg_metrics.py
```diff
@@ -0,0 +1,389 @@
...
+        assert age is not None and 0.0 <= age < 5.0
+
+
+class _FakeCursor:
+    """Cursor returning one preloaded result set per execute() call."""
+
+    def __init__(self, result_sets):
+        self._result_sets = list(result_sets)
+        self._current = None
+        self.executed: list[str] = []
```
muhammad-ali-e: **[Important — pr-test-analyzer]** `_FakeCursor.executed` is collected but never asserted, and the fake ignores `params` entirely. `test_snapshot_from_sql_rows` therefore verifies only row-to-gauge mapping: `refresh_queue_gauges` could query the wrong table, drop the `GROUP BY`, omit `_STRANDED_PREDICATE`, or pass no/wrong bind params for `stuck_timeout_seconds` and the test would still pass — the stranded count would silently diverge from what the reaper actually recovers.

**Fix:** record `(sql, params)` tuples and assert the first execute targets `pg_queue_message` with `GROUP BY queue_name` and no params, and the second targets `pg_barrier_state`, embeds `_STRANDED_PREDICATE`, and binds the timeout.

workers/pg_queue_consumer/supervisor.py
```diff
@@ -450,13 +450,21 @@ def _extra_status() -> dict[str, object]:
             "liveness_probe_bound": True,
         }
 
+    from queue_backend.pg_queue.metrics import ConsumerMetrics
+
+    metrics = ConsumerMetrics(
```
muhammad-ali-e, line 455: **[Important — pr-test-analyzer]** The supervisor's metrics wiring has no test at all — and it is the only producer of `pg_consumer_alive_children` / `pg_consumer_configured_concurrency`. The PR's end-to-end HTTP tests cover the consumer and reaper server subclasses but not this path. Because the lambdas defer evaluation to scrape time, a `_Fleet` API rename (`alive_count` → something else) passes at import and turns every supervisor `/metrics` scrape into a 500 at runtime only.

**Fix:** add a supervisor health test that sets the health-port env to 0, calls `_maybe_start_supervisor_health(...)` for real, GETs `/metrics` on `server.bound_port`, and asserts `pg_consumer_alive_children` and `pg_consumer_configured_concurrency` appear with the fleet's values.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -1047,6 +1130,7 @@ def _run_sweep(self, table: str, fn: Callable[[PgConnection], int]) -> int:
         except Exception:
             streak = self._sweep_fail_streak.get(table, 0) + 1
             self._sweep_fail_streak[table] = streak
+            self._metrics.sweep_failures.labels(table=table).inc()
```
muhammad-ali-e, line 1134: **[Important — pr-test-analyzer]** `pg_reaper_sweep_failures_total{table=...}` is the only new counter in the diff with zero test coverage, including its `table` label value. Since this counter is the alerting signal for the "sweep silently failing for days" scenario the fail-streak log was built for, a wiring regression (wrong label, inc on success path) would go unnoticed.

**Fix:** extend the existing sweep-failure test in `test_pg_reaper.py` (the `OperationalError` side-effect case) to assert `reaper.metrics.registry.get_sample_value("pg_reaper_sweep_failures_total", {"table": ...}) == 1.0`, and that a subsequent successful sweep does not reset it.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -951,13 +1023,17 @@ def tick(self) -> TickOutcome:
                 # A raised renew == "leadership unknown": stop acting (honour the
                 # lease's documented contract) before letting it propagate.
                 self._is_leader = False
+                self._metrics.clear_queue_snapshot()
```
muhammad-ali-e: **[Important test gap — pr-test-analyzer]** This renew-*raises* step-down path clears the snapshot before re-raising, but `test_lost_leadership_clears_snapshot` only covers renew returning `False` (the branch below). This is a separate code path added by this PR; if the `clear` call were dropped here, a reaper that lost DB connectivity mid-lease would keep exporting its frozen queue snapshot as live data — precisely during an incident.

**Fix:** add a twin test with `renew` raising (same `_FakeLease` + `side_effect=OperationalError` pattern as in `test_pg_reaper.py`), asserting the tick raises AND the `pg_queue_depth{queue=...}` sample is `None` afterwards.

workers/queue_backend/pg_queue/liveness.py
```diff
@@ -126,6 +145,24 @@ def do_GET(self) -> None:
...
+                # A broken metrics renderer must degrade to a 500 on /metrics
+                # alone — the probe verdict on /health stays untouched.
+                try:
+                    body = metrics_fn()  # type: ignore[misc]  # guarded by caller
+                except Exception:
+                    logger.exception("%s: /metrics render failed", log_label)
+                    self.send_response(500)
+                    self.end_headers()
+                    return
+                self.send_response(200)
```
muhammad-ali-e: **[Suggestion — silent-failure-hunter]** Only `self.wfile.write(body)` is guarded against `BrokenPipeError`/`ConnectionResetError`. If the scraper disconnects before headers go out, `send_response`/`end_headers` (both the 500 and 200 paths) raise out of `do_GET` into `socketserver`'s `handle_error`, which prints a raw traceback to stderr — bypassing the module logger and the `log_label` attribution this class carefully maintains. Unattributed stderr noise that looks like a crash during scrape-timeout incidents. (The `/health` path shares this pre-existing exposure; this PR adds a second instance.)

**Fix:** wrap the header/write sequence in the same `(BrokenPipeError, ConnectionResetError)` guard.

workers/queue_backend/pg_queue/metrics.py
```diff
@@ -0,0 +1,246 @@
...
+        self._oldest_age.clear()
+        for queue, (depth, oldest_age) in depths.items():
+            self._queue_depth.labels(queue=queue).set(depth)
+            self._oldest_age.labels(queue=queue).set(oldest_age)
+        self._barriers_live.set(barriers_live)
+        self._barriers_stranded.set(barriers_stranded)
+        self._last_refresh_monotonic = time.monotonic()
+
+    def clear_queue_snapshot(self) -> None:
+        """Zero the queue-wide gauges (called on losing leadership, so a standby
```
muhammad-ali-e: **[Suggestion — comment-analyzer]** "Zero the queue-wide gauges" is imprecise: the labeled series (`pg_queue_depth`, `pg_queue_oldest_message_age_seconds`) are *removed* via `.clear()`, not set to 0 — only the unlabeled barrier gauges are zeroed. The PR's own `test_clear_zeroes_the_snapshot` asserts the labeled sample is `None`, contradicting "zero". Absent series vs 0 samples behave differently in PromQL.

**Fix:** "Drop the per-queue series and zero the barrier gauges (called on losing leadership, ...)". Same imprecision in the module docstring ("Standbys export zeroed queue gauges" — the per-queue series are absent on a standby, not zero-valued).

workers/OPERATIONS.md
```diff
@@ -70,11 +70,33 @@ curl http://localhost:8083/health  # Callback worker
...
+  freshness — the same signal `/health` verdicts on); the fleet supervisor adds
+  `pg_consumer_alive_children` / `pg_consumer_configured_concurrency`.
+- Reaper only (the leader-elected singleton — queue-WIDE state comes from one
+  process): `pg_queue_depth{queue}`, `pg_queue_oldest_message_age_seconds{queue}`,
+  `pg_barrier_live` / `pg_barrier_stranded`, recovery outcome counters
```
muhammad-ali-e: **[Suggestion — comment-analyzer]** Two doc nits in this block:

1. The "recovery outcome counters" enumeration reads exhaustive but omits `pg_reaper_barrier_recovery_failures_total` (not covered by the `pg_reaper_claim_*_total` glob) and `pg_reaper_gauge_refresh_failures_total` — the failure twins are the ones an operator alerts on. Use globs symmetrically (`pg_reaper_barrier_*_total`) or append the missing names.
2. "refreshed on the reaper's own cadence" never states the bound. An operator deciding whether `pg_queue_gauges_age_seconds` is cached-fresh or cached-dead needs the expected ceiling — add "(60s — `_GAUGE_REFRESH_INTERVAL_SECONDS` in `reaper.py`)".

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -816,6 +826,56 @@ def sweep_orphan_claims(
     return removed
 
 
+# Queue-gauge snapshot cadence. Deliberately a module constant, not an env knob:
+# it only bounds metrics staleness (the tick already runs every ~5s), and the
```
muhammad-ali-e: **[Suggestion — comment-analyzer]** "(the tick already runs every ~5s)" hardcodes the default of `_DEFAULT_REAPER_INTERVAL_SECONDS`, which is env-overridable via `WORKER_PG_REAPER_INTERVAL_SECONDS` — a rot-prone literal that silently goes stale if the default or deployment config changes.

**Fix:** reference the knob instead of the number: "(the tick already runs every `_DEFAULT_REAPER_INTERVAL_SECONDS` by default)".

workers/queue_backend/pg_queue/metrics.py
```diff
@@ -0,0 +1,246 @@
...
+            alive.set_function(alive_children_fn)
+        if concurrency_fn is not None:
+            conc = Gauge(
+                "pg_consumer_configured_concurrency",
+                "Configured child-process concurrency of the supervisor fleet",
+                registry=self.registry,
+            )
+            conc.set_function(concurrency_fn)
+
+    def render(self) -> bytes:
```
muhammad-ali-e: **[Suggestion — code-simplifier]** The `g = Gauge(name, doc, registry=self.registry); g.set_function(fn)` pattern is written out 6 times across the two exporters, and both classes independently repeat `self.registry = _new_registry()` plus an identical `render()`. A tiny shared base collapses all of it:

```python
class _Exporter:
    def __init__(self) -> None:
        self.registry = _new_registry()

    def _function_gauge(self, name: str, doc: str, fn: Callable[[], float]) -> None:
        from prometheus_client import Gauge
        Gauge(name, doc, registry=self.registry).set_function(fn)

    def render(self) -> bytes:
        return _render(self.registry)
```

One place then owns the function-gauge idiom and the registry/exposition lifecycle; the two classes only declare *what* they export. Minor, same file: the `pg_reaper_claim_gc_total` help text splits a string as two adjacent literals on one line (`"...already " "terminal)"`) with no line-length reason — collapse to one literal.

workers/tests/test_pg_metrics.py
```diff
@@ -0,0 +1,389 @@
...
+            stale_after=60,
+            port=0,
+            check_name="t",
+            age_key="age",
+            metrics_fn=metrics_fn,
+        )
+
+    def test_404_when_no_metrics_fn(self):
+        server = self._server(None)
+        server.start()
```
muhammad-ali-e: **[Suggestion — code-simplifier]** The `server.start(); try: ...; finally: server.stop()` dance repeats 5 times in this file, often burying a one-line assertion. A small context manager makes each test read as its intent:

```python
@contextlib.contextmanager
def _running(server):
    server.start()
    try:
        yield server
    finally:
        server.stop()
```

Also worth a cheap line each (pr-test-analyzer): `/metrics?foo=bar` query-string stripping on the new route is untested (unlike `/health?...`), and `test_leader_tick_refreshes_gauges` asserts `assert_called_once()` with no argument check — pinning the args would catch the reaper being wired to a different `ReaperMetrics` instance than the one `/metrics` serves.

muhammad-ali-e (commented):
## Comprehensive PR review (PR Review Toolkit)

Six specialized review passes: code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier. Findings are posted inline, tagged by severity.

**Counts:** 2 critical, 1 critical test gap, 9 important, 6 suggestions.

**Top items:**
1. Leadership flap leaves a re-elected leader exporting a false "empty queue" for up to 60s (cadence gate not reset on snapshot clear) — found independently by 3 agents.
2. TOCTOU race in `_seconds_since_refresh` can 500 the entire `/metrics` render.
3. The `0.0` "never refreshed" sentinel defeats staleness alerting exactly when the snapshot pipeline is broken.
4. Whole-tick recovery failures have no metric — the recovery net going fully down is the one outage this exporter cannot show.
5. Only the suppression half of the 60s cadence gate is tested; gauges freezing forever after the first snapshot would pass the suite.

**Verified strengths:** `/metrics` 500-isolation from `/health` is structurally enforced and test-pinned; instance-owned registries genuinely prevent the dual-import double-registration; counter increments sit at the same sites that log, before the systemic-failure re-raise; SQL parameterization and rollback-before-reraise are correct.

muhammad-ali-e:
## Review + Sonar feedback addressed (commit 6d07f0f63)

All 18 toolkit findings + the SonarCloud quality-gate failure. The two criticals and the torn-scrape finding shared one root cause, fixed with a redesign rather than three patches.

**Root-cause fix — atomic snapshot collector.** The queue gauges were per-metric `Gauge`s mutated non-atomically, which produced: **[C1]** leadership flap → cleared gauges + un-reset cadence gate → a re-elected leader exporting a false "empty queue" for up to 60s; **[C2]** TOCTOU in `_seconds_since_refresh` (double read of the mutable timestamp → `TypeError` 500ing the whole scrape during churn); **[I2]** torn scrapes (empty/partial depths, new depths with old barrier counts). Now the gauges render from **one immutable `_QueueSnapshot` swapped by reference** — a scrape reads a single snapshot object and can never tear or race a clear. Step-down goes through `_step_down_metrics()` which clears **and resets the refresh cadence**, so a re-elected leader refreshes in the same tick (tested).

**Important:**
- **[I1] staleness metric read 0 in the maximal-staleness case** ✅ the snapshot carries `reference_monotonic` (construction/refresh/step-down time) — `pg_queue_gauges_age_seconds` now GROWS while refreshes fail or no leader exists, so an age alert can actually fire. Help text: alert on it together with `pg_reaper_is_leader == 1`.
- **[I3] whole-tick failures invisible** ✅ added `pg_reaper_tick_failures_total`, incremented in `run()`'s catch — the one machine-readable signal for "every tick fails but the heartbeat stays fresh".
- **[I4] `pg_barrier_live` help/SQL mismatch** ✅ SQL now `FILTER (WHERE remaining > 0)`; stranded stays predicate-only with help noting it includes `remaining==0` lingerers.
- **[I5] OPERATIONS.md** ✅ reaper heartbeat metric documented, consumer bullet rescoped, failure counters enumerated symmetrically, 60s cadence bound named.

**Test gaps (all filled):** cadence resumption + failure-consumes-the-interval [T1]; SQL contract pinned (tables, `GROUP BY`, stranded predicate, bound timeout param) [T2]; supervisor fleet-metrics end-to-end via the real `_maybe_start_supervisor_health` [T3]; `sweep_failures{table}` increment + monotonic-across-success [T4]; renew-**raises** step-down clears the snapshot [T5]; plus `/metrics?query` stripping and arg-pinning on the leader-tick refresh. 31 tests now (was 23).

**Suggestions:** header writes guarded against client hangup on both routes [S1]; "zeroed" wording corrected to dropped-series (moot under the redesign) [S2]; interval comment references the knob [S4]; shared `_Exporter` base collapses the triplication [S5]; `_running()` context manager de-boilerplates the tests [S6].

**SonarCloud (C reliability):** all 24 float-equality assertions → `pytest.approx`; the implicit string concat is gone. Gate should pass on re-analysis.

1200 unit + 123 integration pass.

muhammad-ali-e:
## Greptile + Sonar round addressed (commit 64feedc52)

- **Greptile — missing `describe()` on the custom collector ✅** `_QueueSnapshotCollector` now implements the registration protocol: `register()` checks names against stable descriptors instead of invoking the full `collect()` render (clock read + five family allocations) as a side effect. A single `_families()` builder feeds both `describe()` (names only) and `collect()` (names + samples) so the two can't drift; a new test pins same-names + no-samples.
- **Sonar S5799 — implicit string concat ✅** merged the split help string on the claim-GC counter (a formatter line-wrap artifact).

32 tests pass; pre-commit clean.

PR #2144, opened 2026-07-03
UN-3680 [FIX] PG queue — seam-audit fixes: poison identity, ack/release retries, callback duplicate guard, compose VT

muhammad-ali-e:
## What

Post-merge **seam audit** of the PG-queue epic surfaced interaction bugs *between* individually-reviewed features — the class no single-PR review could see. This closes the **runtime (OSS)** findings before the UN-3675 load-test gate. Every runtime change is PG-gated with a Celery no-op proven by test. The chart-side vt/runbook fixes ship as a separate unstract-cloud PR (repo boundary).

## Fixes

| | Severity | Fix |
|---|---|---|
| **A1** | blocker | **Poison drop couldn't identify orchestration messages.** The backend dispatches `async_execute_bin` with `execution_id` **positionally** (`args[2]`); the poison-drop identity reader only read kwargs. A poisoned orchestration (a circuit-breaker-open drop, *before* the barrier is armed and *before* the claim is taken) had no barrier row and no claim row → invisible to every reaper sweep, a **permanent silent strand**. `_pipeline_identity` now takes a positional fallback keyed by task name. |
| **A2** | medium | **Consumer ack (`client.delete`) lacked the reconnect-retry** every producer-side write already had. The consumer connection idles the whole task wall-clock, so the ack is the likeliest statement to meet a reaped connection; a lost ack redelivers an already-completed message. One-shot reused-gate retry; the DELETE is idempotent (safe even post-commit, unlike `send`'s at-least-once INSERT). |
| **A3** | medium | **`release_orchestration_claim` lacked the retry.** A first-write-after-idle on the failure path whose raise the caller swallows → an idle-reap left the claim committed and suppressed every redelivery. Wrapped in the existing idempotent retry helper. |
| **A4** | medium | **docker-compose PG consumers at the 30s vt default.** The callback + both orchestrators + scheduler had no vt env → a >30s task redelivered mid-run. Added vt/health-stale (env-overridable), mirroring the k8s chart. fileproc/executor left as-is (correctly sized to `EXECUTOR_RESULT_TIMEOUT`). |
| **L1** | low | **Orchestration-entry terminal skip (defense-in-depth).** The claim tombstone normally blocks re-orchestrating a COMPLETED execution; if it was GC'd (a lowered stuck-timeout) a redelivery could re-arm/re-dispatch a finished execution. Skip if already terminal, keeping the freshly-won claim to re-establish the tombstone. PG-gated. |
| **H1** | blocker | **The aggregating callback re-ran webhooks + subscription-usage billing wholesale on redelivery** (the `send` commit-retry double-enqueue, an idle-reaped ack, a vt overrun). Added a terminal re-check that skips the side effects when a prior delivery already finalized the execution. PG-gated via a `_pg_transport` marker the barrier stamps on the PG callback dispatch; the Celery `.link` path never injects it → strict no-op. |

## Celery safety

Every runtime change gates on the PG transport (`is_pg_transport` / the barrier marker) and ships with a Celery-no-op test. The full workers unit lane is untouched.

## Testing

- **1223 unit + real-Postgres integration pass**; pre-commit clean.
- New tests: positional-identity poison recovery, ack-retry (×3 error arms), release-retry, L1 PG-skip + Celery-no-op, H1 duplicate-guard (helper + both callback entry points + barrier marker injection/omission).
- The one in-lane integration failure (`test_exactly_one_task_fires_the_callback`) is the pre-existing UN-3676 shared-lane flaky (passes solo).

## Notes / follow-ups (filed separately, don't block)

- Confirmed while sizing A4: `process_file_batch` iterates files serially at `EXECUTOR_TIMEOUT`, and the PG path's `task.apply()` doesn't enforce the Celery `time_limit` SIGKILL — so a batch can run toward the task time limit. This validates the separate chart PR's fileproc-vt bump.
- H1's minimal terminal re-check does not close the *concurrent*-redelivery race (two callbacks in flight at once); a per-execution callback claim would — deferred unless review wants it here.
- Follow-ups: M1 file-execution terminal-status guard, memory-recycle parity, dead `process_file_batch` endpoint, CI `REQUIRE_PG_TESTS` + real-PG tests for the newest guards, rollback-drain runbook.

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -603,7 +603,7 @@ services:
       # bounded-minutes orchestration, and idempotency-guarded (UN-3671) so a vt
       # overrun can't double-orchestrate. Without an explicit vt the consumer
       # defaults to 30s and a slow dispatch redelivers mid-run. 600/660 mirrors the
-      # k8s chart (grace-equivalent handled by restart:unless-stopped here).
+      # k8s deployment defaults (grace-equivalent handled by restart:unless-stopped here).
       - WORKER_PG_QUEUE_CONSUMER_VT_SECONDS=${WORKER_PG_ORCHESTRATOR_VT_SECONDS:-600}
       - WORKER_PG_QUEUE_CONSUMER_HEALTH_STALE_SECONDS=${WORKER_PG_ORCHESTRATOR_HEALTH_STALE_SECONDS:-660}
     labels:
@@ -636,7 +636,7 @@ services:
       - WORKER_PG_QUEUE_CONSUMER_QUEUE=celery
       - WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT=8090
       # Orchestration is bounded-minutes + claim-guarded (see orchestrator-api).
-      # Without an explicit vt it defaults to 30s. 600/660 mirrors the k8s chart.
+      # Without an explicit vt it defaults to 30s. 600/660 matches our k8s deployment defaults (workerPg* values, separate chart repo).
       - WORKER_PG_QUEUE_CONSUMER_VT_SECONDS=${WORKER_PG_ORCHESTRATOR_VT_SECONDS:-600}
       - WORKER_PG_QUEUE_CONSUMER_HEALTH_STALE_SECONDS=${WORKER_PG_ORCHESTRATOR_HEALTH_STALE_SECONDS:-660}
     labels:
@@ -713,7 +713,7 @@ services:
       # The aggregating callback (fan-in) runs after all batches: aggregate results,
       # update status, push destination, notify. Without an explicit vt it defaults
       # to 30s and a >30s callback redelivers — re-firing webhooks + subscription
-      # billing. 3660/3720 mirrors the k8s chart.
+      # billing. 3660/3720 matches our k8s deployment defaults (separate chart repo).
       - WORKER_PG_QUEUE_CONSUMER_VT_SECONDS=${WORKER_PG_CALLBACK_VT_SECONDS:-3660}
       - WORKER_PG_QUEUE_CONSUMER_HEALTH_STALE_SECONDS=${WORKER_PG_CALLBACK_HEALTH_STALE_SECONDS:-3720}
     labels:
@@ -750,7 +750,7 @@ services:
       - WORKER_PG_QUEUE_CONSUMER_QUEUE=scheduler
       - WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT=8090
       # Scheduled-trigger dispatch is quick (enqueue the execution). Without an
-      # explicit vt it defaults to 30s. 180/240 mirrors the k8s chart.
+      # explicit vt it defaults to 30s. 180/240 matches our k8s deployment defaults (separate chart repo).
       - WORKER_PG_QUEUE_CONSUMER_VT_SECONDS=${WORKER_PG_SCHEDULER_VT_SECONDS:-180}
       - WORKER_PG_QUEUE_CONSUMER_HEALTH_STALE_SECONDS=${WORKER_PG_SCHEDULER_HEALTH_STALE_SECONDS:-240}
     labels:
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -599,6 +599,13 @@ services:
       - WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE=api_deployment
       - WORKER_PG_QUEUE_CONSUMER_QUEUE=celery_api_deployments
       - WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT=8090
+      # async_execute_bin arms the barrier + dispatches batches, then returns — a
+      # bounded-minutes orchestration, and idempotency-guarded (UN-3671) so a vt
+      # overrun can't double-orchestrate. Without an explicit vt the consumer
+      # defaults to 30s and a slow dispatch redelivers mid-run. 600/660 mirrors the
+      # k8s chart (grace-equivalent handled by restart:unless-stopped here).
+      - WORKER_PG_QUEUE_CONSUMER_VT_SECONDS=${WORKER_PG_ORCHESTRATOR_VT_SECONDS:-600}
+      - WORKER_PG_QUEUE_CONSUMER_HEALTH_STALE_SECONDS=${WORKER_PG_ORCHESTRATOR_HEALTH_STALE_SECONDS:-660}
     labels:
       - traefik.enable=false
     volumes:
@@ -628,6 +635,10 @@ services:
       - WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE=general
       - WORKER_PG_QUEUE_CONSUMER_QUEUE=celery
       - WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT=8090
+      # Orchestration is bounded-minutes + claim-guarded (see orchestrator-api).
+      # Without an explicit vt it defaults to 30s. 600/660 mirrors the k8s chart.
+      - WORKER_PG_QUEUE_CONSUMER_VT_SECONDS=${WORKER_PG_ORCHESTRATOR_VT_SECONDS:-600}
+      - WORKER_PG_QUEUE_CONSUMER_HEALTH_STALE_SECONDS=${WORKER_PG_ORCHESTRATOR_HEALTH_STALE_SECONDS:-660}
     labels:
       - traefik.enable=false
     volumes:
@@ -699,6 +710,12 @@ services:
       - WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE=callback
       - WORKER_PG_QUEUE_CONSUMER_QUEUE=file_processing_callback,api_file_processing_callback
       - WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT=8090
+      # The aggregating callback (fan-in) runs after all batches: aggregate results,
+      # update status, push destination, notify. Without an explicit vt it defaults
+      # to 30s and a >30s callback redelivers — re-firing webhooks + subscription
+      # billing. 3660/3720 mirrors the k8s chart.
+      - WORKER_PG_QUEUE_CONSUMER_VT_SECONDS=${WORKER_PG_CALLBACK_VT_SECONDS:-3660}
+      - WORKER_PG_QUEUE_CONSUMER_HEALTH_STALE_SECONDS=${WORKER_PG_CALLBACK_HEALTH_STALE_SECONDS:-3720}
     labels:
       - traefik.enable=false
     volumes:
@@ -732,6 +749,10 @@ services:
       - WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE=scheduler
       - WORKER_PG_QUEUE_CONSUMER_QUEUE=scheduler
       - WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT=8090
+      # Scheduled-trigger dispatch is quick (enqueue the execution). Without an
+      # explicit vt it defaults to 30s. 180/240 mirrors the k8s chart.
+      - WORKER_PG_QUEUE_CONSUMER_VT_SECONDS=${WORKER_PG_SCHEDULER_VT_SECONDS:-180}
+      - WORKER_PG_QUEUE_CONSUMER_HEALTH_STALE_SECONDS=${WORKER_PG_SCHEDULER_HEALTH_STALE_SECONDS:-240}
     labels:
       - traefik.enable=false
     volumes:
--- a/workers/callback/tasks.py
+++ b/workers/callback/tasks.py
@@ -8,6 +8,7 @@ import time
 from typing import Any
 
 from queue_backend import worker_task
+from queue_backend.pg_barrier import PG_TRANSPORT_CALLBACK_KWARG
 
 # Import shared worker infrastructure
 from shared.api import InternalAPIClient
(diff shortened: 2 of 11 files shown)


workers/callback/tasks.py
```diff
@@ -1360,6 +1361,37 @@ def _track_subscription_usage_if_available(
...
+    real callback's side effects, which is worse. PG only — the caller gates on the
+    transport marker, so the Celery path never reaches here.
+    """
+    try:
+        response = api_client.get_workflow_execution(execution_id, file_execution=False)
+        if not response.success:
+            return False
+        status = (response.data.get("execution") or {}).get("status")
+        return status in ExecutionStatus.terminal_values()
+    except Exception:
```
muhammad-ali-e: **[Medium — silent failure]** The `except Exception: … return False` catches *every* error, so a **programming** error (e.g. `response.data is None` → `AttributeError`, or a future rename of `terminal_values()`) silently degrades this guard to a permanent no-op — re-enabling the exact double-webhook / double-billing this feature prevents, with only a WARNING that's indistinguishable from a transient "API down".

**Fix:** narrow the catch to the expected transport/HTTP failure types (so a genuinely unreadable backend still proceeds), and/or log the unexpected case at `ERROR` so a persistently-disabled guard is visible rather than blending into transient noise.

workers/callback/tasks.py
```diff
@@ -1401,6 +1437,23 @@ def _process_batch_callback_core(
                 f"Starting batch callback processing for execution {context.execution_id}"
             )
 
+            # Skip a redelivered callback whose execution a prior delivery already
+            # finalized — don't re-fire webhooks / re-count billing. (Returns
+            # through the outer finally, which closes the api_client.)
+            if is_pg and _pg_execution_already_finalized(
```
muhammad-ali-e: **[Low — design edge]** `_pg_execution_already_finalized` treats any status in `terminal_values()` = `{COMPLETED, STOPPED, ERROR}` as "a prior callback finalized this". But STOPPED/ERROR can be set by **other** paths (external stop, upstream error) — so the *first, legitimate* callback delivery for such an execution would also skip its side effects, including `_cleanup_execution_resources`. The docstring's premise ("the callback is what SETS the terminal status") only strictly holds for COMPLETED.

**Consider:** gating the skip on COMPLETED-only, or confirm that skipping cleanup/notify for an externally-stopped/errored run is intended (re-firing webhooks on a stopped run is admittedly also undesirable — hence Low).

workers/callback/tasks.py
```diff
@@ -1379,6 +1411,10 @@ def _process_batch_callback_core(
     # Initialize performance optimizations
     _initialize_performance_managers()
 
+    # PG at-least-once duplicate guard (see _pg_execution_already_finalized).
+    # Popped BEFORE parameter extraction so the marker never flows into the context.
+    is_pg = bool(kwargs.pop(PG_TRANSPORT_CALLBACK_KWARG, False))
```
muhammad-ali-e, line 1422: **[Low — test gap]** The comment promises the marker is "Popped BEFORE parameter extraction so the marker never flows into the context", but no test asserts this: `TestCoreCallbackGuard` patches out `_extract_callback_parameters` entirely. A reorder of the pop (or forgetting it on the `_api` path at line 1632) would leak `_pg_transport` into downstream task kwargs and pass CI.

**Fix:** in `test_pg_non_terminal_proceeds`, assert `PG_TRANSPORT_CALLBACK_KWARG not in _extract_callback_parameters.call_args.kwargs`.

workers/callback/tasks.py
```diff
@@ -1401,6 +1437,23 @@ def _process_batch_callback_core(
...
+            if is_pg and _pg_execution_already_finalized(
+                context.api_client, context.execution_id
+            ):
+                logger.warning(
+                    f"PG callback: execution {context.execution_id} already terminal "
+                    f"— a previous callback completed; skipping duplicate side effects "
+                    f"(status update, subscription billing, webhooks)."
+                )
+                return {
+                    "status": "skipped_duplicate_callback",
```
muhammad-ali-e: **[Low — DRY]** This `skipped_duplicate_callback` result dict (+ the redundant `duplicate_callback_skipped: True` flag) is byte-identical to the one at line 1675, and `general/tasks.py:328` builds a sibling `skipped_terminal_execution` dict — three hand-built contracts with magic-string statuses that can drift. Any monitoring keying on these fields would break on one path only.

**Fix:** a one-line helper `_skipped_duplicate_callback(execution_id)` (and/or module/enum constants for the status strings) shared by both callback sites.

workers/queue_backend/pg_queue/client.py
```diff
@@ -392,7 +392,41 @@ def delete(self, msg_id: int) -> bool:
...
+        ERRORS`` and only when the failing connection was **reused** (a fresh conn
+        failing is a genuine error). Unlike :meth:`send`'s at-least-once INSERT this
+        DELETE is idempotent — a re-run can only no-op (the row is already gone),
+        never duplicate — so the retry is safe even in the ambiguous post-commit
+        case (it just returns ``False``, "already gone", within the contract above).
         """
+        reused = self._conn is not None and self._owns_conn
+        try:
+            return self._delete_row(msg_id)
+        except _CONN_DEAD_ERRORS as exc:
```
muhammad-ali-e, line 412: **[Low — test parity gap]** `send()` has `test_reused_conn_non_connection_error_not_retried` pinning this narrowed `except _CONN_DEAD_ERRORS`, but the new `TestDeleteReconnectRetry` omits it. If someone broadened this to a bare `except Exception`, a logic bug would be silently retried with no test catching it.

**Fix:** add a test — reused conn raises `RuntimeError("logic")` → `pytest.raises(RuntimeError)`, `factory.assert_not_called()`, `sleep.assert_not_called()`.

workers/queue_backend/pg_queue/client.py
```diff
@@ -392,7 +392,41 @@ def delete(self, msg_id: int) -> bool:
...
+                "on a reused cached connection (%s: %s); dropping it and retrying "
+                "the ack once so the completed message isn't redelivered",
+                msg_id,
+                type(exc).__name__,
+                exc,
+                exc_info=True,
+            )
+            time.sleep(_SEND_RETRY_BACKOFF_SECONDS)
+            # _cursor already dropped the dead owned conn, so this reconnects.
+            return self._delete_row(msg_id)
```
muhammad-ali-e, line 426: **[Low — test parity gap]** This second `_delete_row` is intentionally un-wrapped so a second connection death re-raises (correct, bounded to one retry), but — unlike `send()`'s `test_retry_failure_reraises_once` — no test proves the retry happens **exactly once** with no loop.

**Fix:** add `test_retry_failure_reraises_once` to `TestDeleteReconnectRetry`: cached dead conn + reconnect factory returns a *second* dead conn → assert `pytest.raises(OperationalError)` and `factory.assert_called_once()`.

workers/queue_backend/pg_barrier.py
```diff
@@ -191,6 +191,13 @@ def _cursor() -> Iterator[Any]:
 # errors the broad psycopg2.OperationalError catch also covers.
 _BARRIER_RETRY_BACKOFF_SECONDS: Final = 0.5
 
+# Callback kwarg the PG barrier stamps on the aggregating callback's dispatch so
+# the callback can PG-gate its at-least-once duplicate guard (see
+# _fire_barrier_callback). Matched by the same literal in callback/tasks.py — a
```
muhammad-ali-e: **[Low — comment accuracy]** This says the kwarg is "Matched by the **same literal** in callback/tasks.py — a drift is caught by the callback's PG-guard test". But `callback/tasks.py:11` does `from queue_backend.pg_barrier import PG_TRANSPORT_CALLBACK_KWARG` and uses that symbol at both sites — there is no second `"_pg_transport"` literal anywhere (grep-confirmed), so drift is *structurally impossible* and the described failure mode can't occur.

**Fix:** reword to reflect reality, e.g. "Imported (not re-typed) by callback/tasks.py so producer and consumer can't drift. Underscore-prefixed so it can't collide with a real task kwarg."

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -104,14 +104,51 @@ class _PoisonMarkOutcome(Enum):
...
+# ``kwargs`` / ``_barrier_context``), mapped to ``(execution_id_index,
+# organization_id_index)``. ``async_execute_bin`` is dispatched
+# ``args=[schema_name, workflow_id, execution_id, hash_values]`` — and its poison
+# drop happens BEFORE the barrier is armed AND before the orchestration claim is
+# taken (a circuit-breaker-open drop, or repeated pre-claim failures), so the
+# strand has NO ``pg_barrier_state`` row and NO ``pg_orchestration_claim`` row: it
+# is invisible to every reaper sweep. Without recovering the execution_id here the
+# poison drop can only bare-delete → the execution hangs EXECUTING forever with no
+# handle. ``args[0]`` (org schema) doubles as the org.
+_POSITIONAL_IDENTITY_ARGS: dict[str, tuple[int, int]] = {
```
muhammad-ali-e, line 119: **[Low — invariant not in types]** The load-bearing invariant "`args[2]` is `execution_id`, `args[0]` is org schema" lives only as this hand-maintained tuple + comment, unbound to the actual `async_execute_bin` dispatch site (`args=[schema, workflow_id, execution_id, hash]`). Verified correct **today**, but a future reorder of the dispatch args would silently make the poison-drop path mark the *wrong* execution ERROR (or swap org/exec) with no test or type failure.

**Fix (optional hardening):** derive the indices from a single shared arg-layout definition (NamedTuple / index enum) imported by both the enqueue site and here, so they can't drift independently.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -104,14 +104,51 @@ class _PoisonMarkOutcome(Enum):
...
+    if indices is None:
+        return (None, None)
+    args = payload.get("args") or []
+    exec_idx, org_idx = indices
+    execution_id = args[exec_idx] if exec_idx < len(args) else None
+    organization_id = args[org_idx] if org_idx < len(args) else None
+    return (execution_id, organization_id)
+
+
+def _pipeline_identity(
```
muhammad-ali-e: **[Low — API design]** `task_name` is threaded in as a second, defaulted parameter that duplicates `payload["task_name"]` (already present in the payload and used elsewhere). The two can diverge, and the `None` default silently makes the positional fallback **inert** — an implicit "caller must pass the same task_name that's in the payload" precondition that isn't enforced.

**Fix:** read `payload.get("task_name")` inside `_positional_identity`; if a decoupled override is genuinely wanted, name it explicitly (`task_name_override`) rather than a silently-defaulting positional. (Prod call site at line 555 does pass it, so no runtime bug today.)

workers/tests/test_pg_callback_duplicate_guard.py
```diff
@@ -0,0 +1,146 @@
...
+        with pytest.raises(AssertionError, match="side effects must not run"):
+            self._run(is_pg=True, terminal=False)
+
+    def test_celery_never_checks_or_skips(self):
+        # No marker → the guard is skipped entirely (finalized never called) and
+        # the callback proceeds — proving zero Celery regression.
+        import pytest
+
+        with pytest.raises(AssertionError, match="side effects must not run"):
+            _, finalized, _ = self._run(is_pg=False, terminal=True)
```
muhammad-ali-e: **[Low — test asserts less than it claims]** `self._run(...)` raises the `AssertionError` sentinel *before returning*, so `_, finalized, _ = self._run(...)` never unpacks and the captured `finalized` mock is never asserted on — the comment's claim "finalized never called" is unverified. A regression where the Celery path performs the extra `_pg_execution_already_finalized` round-trip (but still proceeds) would pass.

**Fix:** patch `_pg_execution_already_finalized` at test scope, catch the sentinel, then `finalized.assert_not_called()`. (The test does still catch a wrongful Celery *skip* — this closes the narrower round-trip gap.)

docker/docker-compose.yaml
```diff
@@ -699,6 +710,12 @@ services:
       - WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE=callback
       - WORKER_PG_QUEUE_CONSUMER_QUEUE=file_processing_callback,api_file_processing_callback
       - WORKER_PG_QUEUE_CONSUMER_HEALTH_PORT=8090
+      # The aggregating callback (fan-in) runs after all batches: aggregate results,
+      # update status, push destination, notify. Without an explicit vt it defaults
+      # to 30s and a >30s callback redelivers — re-firing webhooks + subscription
+      # billing. 3660/3720 mirrors the k8s chart.
+      - WORKER_PG_QUEUE_CONSUMER_VT_SECONDS=${WORKER_PG_CALLBACK_VT_SECONDS:-3660}
```
muhammad-ali-e, line 717: **[Low — comment rot risk]** Four comments assert the vt pairs (`600/660`, `3660/3720`, `180/240`) "mirror the k8s chart", but there's no Helm chart in this repo (the chart ships in a separate repo per the PR description), so a maintainer here can't cross-check the numbers and they'll silently rot if the chart's values change — with no test guarding the parity.

**Fix:** name the authoritative source once (external repo/path or values key) and reference it, or soften to "match our k8s deployment defaults" so it doesn't read as a checkable in-repo invariant.

muhammad-ali-e (commented):
## Automated PR review — PG-queue seam-audit (UN-3680)

Reviewed with the PR Review Toolkit (code review, silent-failure, type design, test coverage, comments, simplification). The runtime changes are correct, well PG-gated, and each seam has targeted Celery no-op coverage — the six agents found **no blocker/high correctness defect**. The inline comments below are hardening / test-quality / accuracy improvements, most **Low**, one **Medium** (the duplicate-guard's broad `except`). None need block merge; the Medium and the two `delete()` test-parity gaps are the ones most worth addressing.

_Verified against the real dispatch sites: the positional-identity map `(2,0)` matches `async_execute_bin`'s `args=[schema, workflow_id, execution_id, hash]`, the `delete()` retry faithfully mirrors `send()`, and the terminal guard keeps the claim correctly._

muhammad-ali-e:
## Review feedback addressed (commit 273e9dd0)

Thanks — all actionable items fixed. Highlights:

**SonarCloud (Critical) — `_positional_identity` IndexError** ✅ Now bounds- AND type-checks (`isinstance(args, (list,tuple)) and len(args) > max(idx)`) before any indexing → returns `(None, None)` instead of risking `IndexError`/`TypeError`. Also dropped the redundant `task_name` parameter (toolkit Low): it reads the payload's own `task_name`, so it can't drift and there's no silently-inert `None` default. +test for non-sequence `args`.

**Callback duplicate guard** — combined Greptile's double-fetch + the Medium + the COMPLETED-only edge into one cleaner shape:
- **Greptile (double-fetch)** ✅ The core guard no longer makes a second API call — `_extract_callback_parameters` already fetches the execution, so I store `context.execution_status` (once, at the both-paths convergence) and the guard reads that.
- **Medium (broad `except`)** ✅ Gone — the fetch-helper is replaced by a **pure predicate** `_callback_already_ran(status)`, so there's no `except Exception` that could silently degrade the guard on a programming error.
- **Low (terminal edge)** ✅ Gate on **COMPLETED only**, not all of `terminal_values()`. ERROR/STOPPED can be set by other paths (external stop, reaper), so the *first legitimate* callback for such an execution must still run — COMPLETED is the unambiguous "a prior callback finalized this".
- **Low (DRY)** ✅ Shared `_skipped_duplicate_callback()` result for both callback sites.

**Comments** ✅ `PG_TRANSPORT_CALLBACK_KWARG` reworded (it's *imported* by callback/tasks.py, not a re-typed literal — no drift possible); compose "mirrors the k8s chart" softened (no chart in this repo).

**Test parity (Low)** ✅ `delete()` gains the `non-connection-error-not-retried` + `retry-reraises-once` tests (send() parity); the core-guard test now asserts the marker is popped *before* extraction and that the Celery path never skips.

1230 unit + real-PG integration pass; pre-commit clean. (Deferred, noted in the PR: the *concurrent*-redelivery race and the cross-file status-string constants — small follow-ups, not blockers.)

PR #2147, opened 2026-07-06
UN-3683 [FIX] PG executor result backend — clear result on consume + schema-drift guard

muhammad-ali-e:
## What & why

The PG executor RPC reply store (`pg_task_result`) holds the executor's full return value — the extraction output, including raw document text — for the whole retention TTL (~1h) before the reaper sweeps it. It's a **transient request-reply channel**: once the blocking caller has read the reply, the payload no longer needs to exist. This shrinks that window, and adds a guard so a model schema change can't silently break the workers' raw SQL.

## Changes

**1. Clear-on-consume for `pg_task_result`**
- `PgResultBackend.forget(task_id)` → `UPDATE pg_task_result SET result = NULL`. Nulls the payload but **keeps the row as a tombstone** so an at-least-once redelivery can't re-insert it (`store_result` is `ON CONFLICT DO NOTHING`); `expires_at` is untouched so the reaper still deletes the empty row at TTL.
- Wired at the single consume point (`PgClientQueueTransport.wait_for_result`), right after the reply is read; skipped on timeout.
- **Best-effort — never raises**: it runs after the caller already holds the result, inside `PgExecutionDispatcher.dispatch`'s `except → failure` guard, so a cleanup miss can't turn a successful RPC into a failure. The TTL sweep stays the backstop for unconsumed rows (crash/timeout).
- Net: the window the payload sits in `pg_task_result` drops from ~1h to read-and-clear latency. Flag off → Celery path → no `pg_task_result` row → **zero change**.

**2. Schema-drift guard for the queue tables**
- The workers issue raw psycopg2 SQL (Docker-decoupled from Django), so a model column rename/removal fails only at runtime, not at compile time. New DB-gated test asserts every worker-required column (per-table manifest for all 7 `QUEUE_TABLES`) exists in the live schema — **expected ⊆ actual**, so it catches renames/removals but ignores safe additions (queue SQL uses explicit column lists, never `SELECT *`) — plus a guard that the manifest covers exactly `QUEUE_TABLES`.

## Tests
- `forget`: unit + DB-gated (null-keeps-row, absent-noop, redelivery-blocked, best-effort swallow)
- transport wiring (forgets after consume / not on timeout)
- schema-drift (per-table + registry coverage)
- Full local suite green; DB-gated tests ran against real Postgres.

## Out of scope (follow-ups)
- Celery-side `AsyncResult.forget()` symmetry (touches shared sdk1 / Prompt Studio — larger blast radius).
- `file_history.result` cleartext-at-rest review (durable, always-on for ETL, pre-existing — not introduced here).

## Remaining before ready
- Dev-test against the running integration stack: run one PG ETL, confirm `pg_task_result.result` nulls out after the execution consumes the reply.

### DIFF
--- a/workers/queue_backend/pg_queue/executor_rpc.py
+++ b/workers/queue_backend/pg_queue/executor_rpc.py
@@ -96,6 +96,13 @@ class PgClientQueueTransport(QueueTransport):
         """
         with PgResultBackend() as rb:
             row = rb.wait_for_result(reply_key, timeout)
+            if row is not None:
+                # Reply consumed: drop the payload now so customer extraction data
+                # (PII) doesn't sit in pg_task_result for the whole retention TTL.
+                # Keep the row as a tombstone (blocks at-least-once re-insert); the
+                # reaper deletes it at expires_at. Best-effort — forget never raises,
+                # so a cleanup miss can't fail an otherwise successful dispatch.
+                rb.forget(reply_key)
         if row is None:
             return None
         return ExecResultRow(
--- a/workers/queue_backend/pg_queue/executor_rpc.py
+++ b/workers/queue_backend/pg_queue/executor_rpc.py
@@ -97,11 +97,8 @@ class PgClientQueueTransport(QueueTransport):
         with PgResultBackend() as rb:
             row = rb.wait_for_result(reply_key, timeout)
             if row is not None:
-                # Reply consumed: drop the payload now so customer extraction data
-                # (PII) doesn't sit in pg_task_result for the whole retention TTL.
-                # Keep the row as a tombstone (blocks at-least-once re-insert); the
-                # reaper deletes it at expires_at. Best-effort — forget never raises,
-                # so a cleanup miss can't fail an otherwise successful dispatch.
+                # Reply consumed: drop the payload now so PII doesn't sit in
+                # pg_task_result for the full retention TTL. Best-effort (see forget()).
                 rb.forget(reply_key)
         if row is None:
             return None
--- a/workers/queue_backend/pg_queue/result_backend.py
+++ b/workers/queue_backend/pg_queue/result_backend.py
@@ -83,6 +83,14 @@ def _get_sql() -> str:
     )
 
 
+def _forget_sql() -> str:
+    # Null the payload but KEEP the row: the tombstone makes a redelivered
+    # ``store_result`` a ``ON CONFLICT DO NOTHING`` no-op (can't re-insert the
+    # payload), and ``expires_at`` is left untouched so the reaper still deletes
+    # the empty row at retention.
+    return f"UPDATE {qualified('pg_task_result')} SET result = NULL WHERE task_id = %s"
+
+
 # Poll cadence for wait_for_result: start tight (low latency for fast tasks),
 # back off to a ceiling so a long-running task doesn't hammer the DB.
 _POLL_INITIAL_SECONDS = 0.2
@@ -257,6 +265,34 @@ class PgResultBackend:
             initial=poll_interval,
             maximum=_POLL_MAX_SECONDS,
         )
+
+    def forget(self, task_id: str) -> None:
+        """Drop the stored payload once the blocking caller has consumed it.
+
+        Nulls the ``result`` column but keeps the row as a tombstone (see
+        :func:`_forget_sql`), so a redelivered executor message can't re-insert
+        the payload and the reaper still flushes the empty row at ``expires_at``.
+        Cutting the payload here shrinks the window that customer extraction data
+        (PII) sits in ``pg_task_result`` from the full retention TTL down to the
+        read-and-clear latency; the TTL sweep remains the backstop for any result
+        that is never consumed (caller crash / timeout).
+
+        **Best-effort by contract — never raises.** It runs AFTER the caller
+        already holds the result, and its call site is inside
+        ``PgExecutionDispatcher.dispatch``'s ``except Exception -> failure`` guard;
+        a raise here would turn a successful RPC into a failure. Any error is
+        logged and swallowed — the retention sweep still bounds the exposure.
+        """
+        try:
+            with self._cursor() as cur:
+                cur.execute(_forget_sql(), (str(task_id),))
+        except Exception:
+            logger.warning(
+                "PgResultBackend.forget: could not clear result for task_id=%s; "
+                "the retention sweep will flush it at expires_at",
+                task_id,
+                exc_info=True,
+            )
 
     def close(self) -> None:
         """Close an owned connection (injected connections are the caller's)."""
--- a/workers/queue_backend/pg_queue/result_backend.py
+++ b/workers/queue_backend/pg_queue/result_backend.py
@@ -161,7 +161,7 @@ class PgResultBackend:
             raise
 
     def _store_with_reconnect(self, operation: Callable[[Any], None]) -> None:
-        """Run an idempotent store ``operation(cur)`` in a committed cursor,
+        """Run an idempotent write ``operation(cur)`` in a committed cursor,
         retrying ONCE if the cached connection was reaped while idle.
 
         The executor consumer caches one ``PgResultBackend`` for its lifetime
@@ -181,9 +181,13 @@ class PgResultBackend:
         executor consumer constructs ``PgResultBackend()``).
 
         Safe to retry unconditionally (no reused-vs-fresh guard needed, unlike
-        the non-idempotent ``PgQueueClient.send``): the write is
-        ``INSERT … ON CONFLICT (task_id) DO NOTHING``, so re-running after an
-        ambiguous failure can neither duplicate nor clobber a recorded result.
+        the non-idempotent ``PgQueueClient.send``) because **every** write routed
+        through here is idempotent: ``store_result``'s
+        ``INSERT … ON CONFLICT (task_id) DO NOTHING`` and ``forget``'s
+        ``UPDATE … SET result = NULL, error = ''`` (re-nulling an already-nulled
+        tombstone is a no-op). Re-running either after an ambiguous failure can
+        neither duplicate nor clobber a recorded result. Keep this invariant if a
+        third caller is added — a non-idempotent write must NOT use this helper.
         """
         for attempt in range(1, _STORE_RETRY_ATTEMPTS + 1):
             try:
--- a/workers/queue_backend/pg_queue/result_backend.py
+++ b/workers/queue_backend/pg_queue/result_backend.py
@@ -11,6 +11,11 @@ A row appears ONLY when the task finishes — ``status="completed"`` carrying th
 text if the task raised. Absence of a row means "not done yet"; there is
 deliberately no ``pending`` state to maintain.
 
+Once the blocking caller consumes the reply, :meth:`PgResultBackend.forget` nulls
+the payload in place — a third legal shape: ``completed``/``failed`` with
+``result`` and ``error`` cleared (a tombstone the reaper deletes at
+``expires_at``). Readers must not assume a finished row still carries a payload.
+
 Two deliberate properties:
 
 - **Idempotent writes** (``INSERT … ON CONFLICT DO NOTHING``): the PG queue is
@@ -84,11 +89,17 @@ def _get_sql() -> str:
 
 
 def _forget_sql() -> str:
-    # Null the payload but KEEP the row: the tombstone makes a redelivered
-    # ``store_result`` a ``ON CONFLICT DO NOTHING`` no-op (can't re-insert the
-    # payload), and ``expires_at`` is left untouched so the reaper still deletes
-    # the empty row at retention.
-    return f"UPDATE {qualified('pg_task_result')} SET result = NULL WHERE task_id = %s"
+    # Null BOTH payload channels but KEEP the row: ``result`` (success payload) and
+    # ``error`` (failure text — an extraction error can embed document content too),
+    # so the PII scrub covers the failure channel, not just the success one. The
+    # tombstone makes a redelivered ``store_result`` a ``ON CONFLICT DO NOTHING``
+    # no-op (can't re-insert the payload); ``status`` still distinguishes the outcome
+    # and ``expires_at`` is left untouched so the reaper still deletes the row at
+    # retention.
+    return (
+        f"UPDATE {qualified('pg_task_result')} "
+        "SET result = NULL, error = '' WHERE task_id = %s"
+    )
 
 
 # Poll cadence for wait_for_result: start tight (low latency for fast tasks),
@@ -225,7 +236,11 @@ class PgResultBackend:
         """Return ``{status, result, error}`` if the row exists, else ``None``.
 
         ``result`` is the decoded JSONB dict (psycopg2 parses ``jsonb`` to a
-        Python ``dict``); ``None`` means the task has not finished yet.
+        Python ``dict``) or ``None``. A *missing* row (this returns ``None``) means
+        the task has not finished. A *present* row with ``result is None`` is either
+        a ``failed`` outcome or a ``completed`` **tombstone** whose payload was
+        already cleared by :meth:`forget` post-consume — ``status`` distinguishes
+        them, so a payload-poll must not treat a completed row as carrying a result.
 
         No reconnect-retry here (unlike :meth:`store_result`). The invariant it
         relies on: the sole entry point into a wait (``executor_rpc.wait_for_result``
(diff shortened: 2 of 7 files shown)


workers/tests/test_pg_schema_drift.py
```diff
@@ -0,0 +1,103 @@
...
+in Docker decoupled from the backend by design, so they cannot import the models
+as a source of truth. Column names therefore live as **string literals** across
+``queue_backend/pg_queue/*``. A change in ``backend/pg_queue`` that renames or
+drops one of those columns would NOT fail at import/compile time — it would only
+surface at runtime as a SQL error.
+
+This DB-gated test closes that gap: it asserts every column the workers depend on
+still exists in the live schema, so a model rename/removal fails in CI instead of
+in production.
+
```
muhammad-ali-e, line 19: **Important — the guard doesn't actually run in CI, so this claim doesn't hold yet.** The `integration-workers` lane in `tests/groups.yaml` is `optional: true` and provisions a bare Postgres without the pg_queue migration, so the `pg_conn` fixture skips (`to_regclass('pg_queue_message')` is NULL) and nothing in CI sets `REQUIRE_PG_TESTS`. Only `test_contract_covers_every_registered_queue_table` runs in the unit lane, and it checks the manifest against `QUEUE_TABLES`, not the live schema. A backend column rename would still reach production unflagged.

Suggest provisioning a migrated DB + `REQUIRE_PG_TESTS=1` in the integration lane (conftest already has the skip-becomes-failure enforcement waiting for it), or softening this docstring to "fails on migrated dev machines / once the CI lane is wired" until then.

workers/queue_backend/pg_queue/executor_rpc.py
```diff
@@ -96,6 +96,13 @@ def wait_for_result(self, reply_key: str, timeout: float) -> ExecResultRow | Non
         """
         with PgResultBackend() as rb:
             row = rb.wait_for_result(reply_key, timeout)
+            if row is not None:
+                # Reply consumed: drop the payload now so customer extraction data
+                # (PII) doesn't sit in pg_task_result for the whole retention TTL.
+                # Keep the row as a tombstone (blocks at-least-once re-insert); the
+                # reaper deletes it at expires_at. Best-effort — forget never raises,
+                # so a cleanup miss can't fail an otherwise successful dispatch.
+                rb.forget(reply_key)
```
muhammad-ali-e, line 102: **Important — PII-retention parity gap with the backend transport.** The backend's `DjangoQueueTransport.wait_for_result` (`backend/pg_queue/executor_rpc.py:84-102`) polls the same `pg_task_result` table but never forgets, so a backend-dispatched RPC still leaves the extraction payload for the full retention TTL — the exact exposure this PR closes on the workers side. There is also no shared transport-contract test pinning "payload cleared after consume", so the two implementations have already diverged silently.

Suggest mirroring the consume-then-clear on the backend transport (follow-up PR is fine) and/or adding a contract test parametrized over both transports. If backend-side forget is deliberately deferred, a one-line note here would stop the next reader from assuming parity.

workers/tests/test_pg_schema_drift.py
```diff
@@ -0,0 +1,103 @@
...
+    },
+}
+
+
+def _actual_columns(pg_conn, table: str) -> set[str]:
+    """Columns present for *table* in the live schema (search-path resolved, so it
+    matches how the workers' unqualified/qualified SQL sees the table)."""
+    with pg_conn.cursor() as cur:
+        cur.execute(
+            "SELECT column_name FROM information_schema.columns WHERE table_name = %s",
```
muhammad-ali-e: **Important — missing `table_schema` filter (and the docstring claim above is inaccurate).** `information_schema.columns` is privilege-filtered, not search_path-filtered: this query returns columns from *every* visible schema containing a table named `table`. In the multi-schema cloud-dev DB that `schema.py` itself describes (per-developer schemas alongside `unstract`), a column dropped in the target schema but still present in a stale sibling schema unions in and the guard passes — exactly the drift it exists to catch. The connection's `search_path` option has no effect on this view, so "search-path resolved" (line 77) is false.

Suggest:
```sql
SELECT column_name FROM information_schema.columns
WHERE table_name = %s AND table_schema = %s
```
bound to the resolved queue schema (e.g. `queue_schema()` from `queue_backend.pg_queue.schema`), and rewording the docstring to say the schema is pinned explicitly.

muhammad-ali-e (commented):
## Comprehensive PR review (6-agent pass: code review, silent-failure, type design, test coverage, comment accuracy, simplification)

**Overall: the core mechanism is sound.** Verified against the code: `store_result` really is `ON CONFLICT (task_id) DO NOTHING` so the tombstone blocks redelivery re-insert; `_cursor` commits on success so the UPDATE is durable; the reaper deletes expired `pg_task_result` rows; `reply_key` is a fresh UUID per dispatch and the dispatcher already maps completed-with-NULL-result to an explicit failure, so a tombstone can never masquerade as a successful reply. No critical correctness bugs found.

**Top findings (details inline):**
1. **The schema-drift guard does not actually run in CI today** — the integration lane provisions an unmigrated Postgres and nothing sets `REQUIRE_PG_TESTS`, so the DB-gated tests skip and the docstring's "fails in CI instead of in production" claim doesn't hold yet.
2. **PII-retention parity gap**: the backend's `DjangoQueueTransport.wait_for_result` never forgets, so backend-dispatched RPCs still leave payloads for the full TTL; no contract test pins the behavior across both transports.
3. **`_actual_columns` doesn't filter by `table_schema`** — in a multi-schema DB a stale sibling schema can mask exactly the drift the guard exists to catch.
4. **`forget()` failures have no metric** — a systematic failure silently reverts PII retention to the full TTL with nothing alertable; the pg_queue metrics module's convention is a counter at each such log site.
5. Two of the new tests are vacuous as written (the catch-all guarantees no-raise regardless of behavior); `caplog` assertions would give them teeth.

**Out-of-diff note:** `backend/pg_queue/models.py:335-343` still says the retention sweep is "future" / "no sweeper reads it yet", which now directly contradicts this PR's (correct) comments that the reaper deletes at `expires_at` — worth fixing while in this area.

**Strengths:** the tombstone/redelivery test asserts real DB state; the manifest-completeness test pins `WORKER_SCHEMA_CONTRACT == QUEUE_TABLES` so a new queue table can't silently skip the guard; the never-raises contract is correctly scoped to `Exception` (not `BaseException`) and cannot poison the cached connection (`_cursor` discards a dead owned handle before the swallow runs).

workers/queue_backend/pg_queue/result_backend.py
```diff
@@ -258,6 +266,34 @@ def wait_for_result(
...
+        already holds the result, and its call site is inside
+        ``PgExecutionDispatcher.dispatch``'s ``except Exception -> failure`` guard;
+        a raise here would turn a successful RPC into a failure. Any error is
+        logged and swallowed — the retention sweep still bounds the exposure.
+        """
+        try:
+            with self._cursor() as cur:
+                cur.execute(_forget_sql(), (str(task_id),))
+        except Exception:
+            logger.warning(
```
muhammad-ali-e, line 318: **Important — no aggregate signal when forget fails.** A single `logger.warning` is fine for a one-off miss (the reaper backstop is real), but a *systematic* failure — revoked UPDATE privilege, recurring transient errors — silently extends PII retention from read-and-clear latency back to the full 3600s TTL on every dispatch, and nothing alertable fires. The pg_queue metrics module (b84980b0) established the pattern of incrementing an outcome counter at the same sites that log the outcome (`sweep_failures`, `barrier_recovery_failures`, …); `forget` breaks that pattern.

Suggest a `forget_failures` counter plumbed to the exporter, or at minimum a documented log-based alert on this warning string.

Related (low): `store_result` gets a one-shot reconnect-retry precisely because an idempotent write on a dead owned connection is rescuable — `forget` is equally idempotent and could route through the same helper, so a transient blip doesn't cost the full TTL.

workers/queue_backend/pg_queue/result_backend.py
```diff
@@ -83,6 +83,14 @@ def _get_sql() -> str:
     )
 
 
+def _forget_sql() -> str:
+    # Null the payload but KEEP the row: the tombstone makes a redelivered
+    # ``store_result`` a ``ON CONFLICT DO NOTHING`` no-op (can't re-insert the
+    # payload), and ``expires_at`` is left untouched so the reaper still deletes
+    # the empty row at retention.
+    return f"UPDATE {qualified('pg_task_result')} SET result = NULL WHERE task_id = %s"
```
muhammad-ali-e: **Suggestion — `error` survives forget, so the failure channel is excluded from the PII scrub.** Failed replies store `error = f"{type(exc).__name__}: {exc}"` (`consumer.py:357`), and exception text from extraction failures can embed customer document content. The call site forgets on failed rows too, but for them this UPDATE is a no-op (`result` is already NULL) and the error text sits for the full TTL — the docstring's PII-shrink claim quietly applies only to the success channel.

Consider `SET result = NULL, error = ''` (status alone distinguishes the outcome, and no reader consumes the row after forget), or document why error text is out of scope.

workers/queue_backend/pg_queue/result_backend.py
```diff
@@ -258,6 +266,34 @@ def wait_for_result(
             maximum=_POLL_MAX_SECONDS,
         )
 
+    def forget(self, task_id: str) -> None:
+        """Drop the stored payload once the blocking caller has consumed it.
```
muhammad-ali-e, line 289: **Suggestion — the surrounding docs still describe a two-state table.** The module docstring (lines 9-12: a row is either `completed` *carrying the payload* or `failed` carrying error text) and `get_result`'s docstring ("`result` is the decoded JSONB dict") don't admit the third legal shape this method introduces: `completed` with `result=NULL` (consumed tombstone). A future `get_result`-based feature (e.g. poll-by-task-id for `dispatch_async`) could reasonably code against the stale two-state invariant. One sentence in each would close that.

Also a rot risk: this docstring hardcodes `PgExecutionDispatcher.dispatch`'s guard, but the immediate caller is `PgClientQueueTransport.wait_for_result` three modules away. Stating the invariant instead ("every caller must sit under the dispatcher's never-raises guard") survives a second caller being added.

workers/tests/test_pg_result_backend.py
```diff
@@ -106,6 +106,45 @@ def write_after_delay() -> None:
...
+        k = _key()
+        result_backend.store_result(k, result={"data": {"pii": "secret"}})
+        result_backend.forget(k)
+        row = result_backend.get_result(k)
+        assert row is not None  # row kept (tombstone), not deleted
+        assert row["result"] is None  # payload dropped
+        assert row["status"] == STATUS_COMPLETED  # status preserved
+
+    def test_forget_absent_is_noop(self, result_backend):
+        result_backend.forget(_key())  # must not raise; UPDATE matches no row
```
muhammad-ali-e: **Suggestion — vacuous as written.** Because `forget()` catches `Exception` unconditionally, "must not raise" passes even if `_forget_sql()` were invalid SQL or the DB rejected the UPDATE outright — the failure mode this test claims to pin is indistinguishable from total breakage. Asserting via `caplog` that **no** warning was emitted would prove the UPDATE actually executed cleanly against an absent row.

workers/tests/test_pg_result_backend.py
```diff
@@ -106,6 +106,45 @@ def write_after_delay() -> None:
...
+    def test_forget_is_best_effort_swallows_errors(self, monkeypatch):
+        """A cleanup failure must not propagate: forget runs inside
+        PgExecutionDispatcher.dispatch's ``except -> failure`` guard AFTER the
+        caller already holds the result, so a raise would fail a good RPC.
+        """
+        rb = PgResultBackend()
+        monkeypatch.setattr(
+            rb, "_cursor", MagicMock(side_effect=psycopg2.OperationalError("dead"))
+        )
+        rb.forget("k")  # must NOT raise
```
muhammad-ali-e: **Suggestion — pins "swallowed" but not "logged".** The stated contract is *logged and swallowed*; a refactor to `contextlib.suppress` (the forbidden fully-silent shape) would pass this test unchanged. Add a `caplog` assertion that a WARNING containing "could not clear result" and the task_id was emitted.

workers/tests/test_executor_rpc.py
```diff
@@ -323,6 +323,30 @@ def test_wait_for_result_none_passes_through(self):
...
+        rb = MagicMock()
+        rb.__enter__.return_value = rb
+        rb.wait_for_result.return_value = {
+            "status": "completed",
+            "result": {"a": 1},
+            "error": "",
+        }
+        with patch(f"{_WMOD}.PgResultBackend", return_value=rb):
+            PgClientQueueTransport().wait_for_result("rk", 5)
+        rb.forget.assert_called_once_with("rk")
```
muhammad-ali-e, line 338: **Suggestion — doesn't pin that forget runs *inside* the context.** This assertion passes even if `rb.forget(reply_key)` were moved after the `with` block. In production that code motion is a real defect: `PgResultBackend.conn` lazily reopens a **new owned connection** after `close()`, which nothing would ever close — a connection leak per dispatch, invisible here. Consider asserting order via `rb.mock_calls` (forget before `__exit__`).

workers/tests/test_pg_result_backend.py
```diff
@@ -106,6 +106,45 @@ def write_after_delay() -> None:
...
+    """
+
+    def test_forget_nulls_result_keeps_row(self, result_backend):
+        k = _key()
+        result_backend.store_result(k, result={"data": {"pii": "secret"}})
+        result_backend.forget(k)
+        row = result_backend.get_result(k)
+        assert row is not None  # row kept (tombstone), not deleted
+        assert row["result"] is None  # payload dropped
+        assert row["status"] == STATUS_COMPLETED  # status preserved
```
muhammad-ali-e, line 134: **Nit — consider also asserting `expires_at` is unchanged.** The tombstone design relies on the reaper deleting at the *original* `expires_at` (`_forget_sql` comment); a future edit that also touches `expires_at` would pass this test.

workers/tests/test_pg_schema_drift.py
```diff
@@ -0,0 +1,103 @@
...
+        "organization_id",
+        "workflow_id",
+        "pipeline_name",
+        "cron_string",
+        "enabled",
+        "pg_owned",
+        "last_run_at",
+        "next_run_at",
+        "created_at",
+        "updated_at",
```
muhammad-ali-e, line 79: **Nit — contract over-mirrors for this table.** No worker raw SQL reads `pg_periodic_schedule.created_at` or `updated_at` (the scheduler SQL uses neither), so renaming them would fail this guard even though no worker is affected — contradicting the docstring's own semantics ("a renamed/removed column **that the workers use** → fails"). Either trim these two or reword the header comment to "columns the workers expect to exist (full model mirror)".

workers/tests/test_pg_schema_drift.py
```diff
@@ -0,0 +1,103 @@
...
+
+@pytest.mark.parametrize("table", sorted(WORKER_SCHEMA_CONTRACT))
+def test_worker_required_columns_exist(pg_conn, table):
+    actual = _actual_columns(pg_conn, table)
+    assert actual, f"{table} not found in the live schema (migration applied?)"
+    missing = WORKER_SCHEMA_CONTRACT[table] - actual
+    assert not missing, (
+        f"{table} is missing worker-required column(s) {sorted(missing)}: a model "
+        f"rename/removal in backend/pg_queue broke the workers' raw SQL. Update the "
+        f"SQL in queue_backend/pg_queue/ and WORKER_SCHEMA_CONTRACT together."
```
muhammad-ali-e: **Nit — pointer misses `pg_barrier.py`.** The raw SQL for `pg_barrier_state` and `pg_orchestration_claim` lives in `queue_backend/pg_barrier.py`, one directory above `queue_backend/pg_queue/` — a maintainer following this failure message would grep the wrong directory. Say `queue_backend/` (or name both paths). Same applies to the module docstring's "across ``queue_backend/pg_queue/*``" (line 6).

workers/queue_backend/pg_queue/executor_rpc.py
```diff
@@ -96,6 +96,13 @@ def wait_for_result(self, reply_key: str, timeout: float) -> ExecResultRow | Non
         """
         with PgResultBackend() as rb:
             row = rb.wait_for_result(reply_key, timeout)
+            if row is not None:
+                # Reply consumed: drop the payload now so customer extraction data
+                # (PII) doesn't sit in pg_task_result for the whole retention TTL.
+                # Keep the row as a tombstone (blocks at-least-once re-insert); the
```
muhammad-ali-e: **Nit — this comment duplicates `forget()`'s docstring.** The tombstone/PII/best-effort rationale is authored in full ~5 times across the PR (here, the `forget()` docstring, `_forget_sql()`, and two test docstrings). `forget()`'s docstring is the right canonical home; this call-site comment could shrink to the locally relevant part, e.g.:
```python
# Reply consumed: drop the payload now so PII doesn't sit in
# pg_task_result for the full retention TTL. Best-effort (see forget()).
```

muhammad-ali-e:
Thanks for the thorough review — all valid, addressed in `cbf260d00`.

### Important
- **Backend-transport parity** — `DjangoQueueTransport.wait_for_result` now clears `result`+`error` after consume, mirroring the workers' `forget`, so a backend-dispatched RPC no longer leaves the payload for the full TTL. Added `test_present_row_clears_payload_after_consume` + `test_timeout_does_not_clear`. (A single shared parametrized-over-both-transports contract test is awkward across the two packages; the two mirrored suites cover it for now.)
- **`information_schema` missing `table_schema`** — real bug, fixed. Query now pins `table_schema = queue_schema()` (the exact schema the workers' `qualified()` SQL targets); rewrote the docstring — it's privilege-filtered, not search-path-filtered.
- **Drift guard doesn't run in CI** — correct; softened the claim. Docstring now says it enforces only against a migrated Postgres (dev machine or a CI lane that provisions one + `REQUIRE_PG_TESTS=1`) and calls out that the `integration-workers` lane is optional/un-migrated today, so it's a dev/CI-lane guard, not a prod merge gate. Wiring a migrated lane noted as follow-up.
- **No aggregate signal on forget failure** — `forget` now routes through the idempotent `_store_with_reconnect`, so a transient blip self-heals on the one-shot retry instead of costing the full TTL (your "Related (low)" point). A surviving failure is logged; docstring says to alert on the `"could not clear result"` string until a `forget_failures` metric is wired — deferred to the pg_queue metrics work (no executor-RPC exporter exists to hang it on yet).

### Suggestions
- **`error` survives forget** — good catch; `_forget_sql` now `SET result = NULL, error = ''` so the failure channel is scrubbed too. Added `test_forget_clears_error_channel_too`.
- **Two-state docstrings** — module + `get_result` docstrings now admit the consumed-tombstone third shape (`completed`/`failed` with payload cleared); `forget` states the never-raises **invariant** instead of hardcoding the dispatcher caller.
- **Vacuous best-effort tests** — `test_forget_absent_is_clean_noop` now asserts (via `caplog`) that **no** warning fires (proves the UPDATE ran), and `test_forget_is_best_effort_logs_and_swallows` asserts the WARNING with the task_id **is** emitted (pins logged-and-swallowed vs a silent `suppress`).
- **Wiring test order** — now asserts `forget` runs before `__exit__` via `rb.mock_calls` (guards against the reopen-after-close leak).

### Nits
- `expires_at`-unchanged now asserted in the tombstone test.
- Drift docstring reworded to "full model mirror" (consistent with including `pg_periodic_schedule.created_at`/`updated_at`).
- Path pointers fixed to `queue_backend/` (naming `pg_barrier.py`), in both the module docstring and the failure message.
- Call-site comment shrunk to the local bit + "see forget()".

Tests: workers 69 passed (DB-gated ran against real PG), backend `DjangoQueueTransportWait` 5 passed.

muhammad-ali-e:
@greptile-apps addressed the one comment in `1f344a366`: generalized `_store_with_reconnect`'s docstring — the unconditional-retry justification now covers **both** idempotent callers (`store_result`'s `INSERT … ON CONFLICT DO NOTHING` and `forget`'s `UPDATE SET result = NULL, error = ''`), with a note to keep the idempotency invariant for any future caller. Docs-only, no behaviour change. Thanks for the catch — and for the merge-safety read.

PR #2148, opened 2026-07-07
UN-3684 [FIX] PG barrier — claim_batch idle-reap: execute-phase reconnect-retry

muhammad-ali-e:
## What & why

Scheduled ETL executions on the PG queue intermittently **stranded for ~2.5h** and then got reaper-marked **ERROR** (`[reaper-recovery] … barrier expired`). Root cause (confirmed via GCP logs + DB + code): the `pg_barrier` module's **thread-local DB connection sits idle between infrequent scheduled batches** — a pipeline fires one only ~every 30 min, past PgBouncer's `server_idle_timeout`. The server reaps it; the next batch's `claim_batch` runs its `INSERT … RETURNING` on the **stale handle** → `server closed the connection unexpectedly` (execute phase — nothing committed). `claim_batch` was on plain `_cursor()` with **no reconnect-retry**, so it failed outright; the batch never started, and (fileproc `vt` ≈ reaper timeout ≈ 2.5h, no fast retry) it stranded until the reaper.

This is the **idle-reap connection-resilience family** (UN-3654/3651/3659/3660) — `claim_batch` is the one member that never got hardened. Its sibling one level up, `try_claim_orchestration`, already implements the correct pattern.

## Change

Give `claim_batch` the same **execute-phase reconnect-retry** as `try_claim_orchestration` (phase-split):

| Failure | Behavior | Why safe |
|---|---|---|
| **reused-conn, execute-phase** (the idle-reap) | reconnect + retry once | the INSERT never committed → the retry's `RETURNING` claim answer is authoritative |
| **commit-phase** (ambiguous) | **not** retried, propagates | a re-run could flip the claim (`True`→`False` → skip-and-strand, or re-claim → double-decrement) |
| **fresh-conn** failure | **not** retried, propagates | a real DB error, not a reap |

An idle-reaped batch connection now self-heals in ~0.5s at claim time — the batch starts on first delivery, **no 2.5h strand, the reaper never involved.**

## Tests
`TestClaimBatchRetry`: execute-phase reaped-cached-conn retries once and returns the authoritative claim; commit-phase failure is **not** retried; fresh-conn failure is **not** retried. Full `pg_barrier` suite: **91 passed**.

## Considered and dropped
A consumer-side fast-redeliver + dedicated conn-death budget. `claim_batch`'s dedup marker is an at-most-once gate, so it can't recover a *post-claim* strand anyway — this fix covers 100% of the observed failure (idle-reap at the claim) with far less risk. Revisit only if strands persist after this.

## Remaining before ready
Dev-test against the running stack: let a scheduled ETL run after an idle gap (or bounce PgBouncer to force a reap); confirm the `claim_batch(...): execute failed on a cached connection — reconnecting` warning fires and the execution **COMPLETES** instead of stranding.

### DIFF
--- a/workers/queue_backend/pg_barrier.py
+++ b/workers/queue_backend/pg_barrier.py
@@ -278,17 +278,77 @@ def claim_batch(execution_id: str, batch_index: int) -> bool:
     Called on the in-body PG path from :func:`run_batch_with_barrier` (claim at
     batch start), alongside the in-body decrement. The companion
     :func:`clear_execution_batches` reclaims the markers at barrier finalise.
+
+    **Execute-phase reconnect-retry** (mirrors :func:`try_claim_orchestration`,
+    the execution-level claim one level up). This module's thread-local connection
+    sits idle between batches — a scheduled pipeline fires one only ~every 30 min,
+    well past PgBouncer's ``server_idle_timeout`` — so the FIRST ``INSERT`` after
+    the gap can hit an idle-reaped handle (``server closed the connection
+    unexpectedly``). Retrying a **reused-conn execute-phase** death is safe: the
+    INSERT never committed, so the retry's win/lost ``RETURNING`` answer is
+    authoritative. A **commit-phase** failure is ambiguous (the row may be
+    persisted) — re-running could flip the claim (``True`` → ``False`` → the caller
+    skips and strands the barrier, or a re-claim double-decrements), so it is NOT
+    retried; it propagates to vt-expiry redelivery. A fresh-conn failure is a real
+    DB error, likewise not retried.
     """
-    with _cursor() as cur:
-        cur.execute(
-            f"INSERT INTO {qualified('pg_batch_dedup')} "
-            "(execution_id, batch_index, created_at) "
-            "VALUES (%s, %s, now()) "
-            "ON CONFLICT (execution_id, batch_index) DO NOTHING "
-            "RETURNING execution_id",
-            (execution_id, batch_index),
-        )
-        return cur.fetchone() is not None
+    # Only a CACHED handle can be a stale idle-reap; sample before _get_conn()
+    # below materialises one, so a fresh-conn failure isn't misread as a reap
+    # (mirrors try_claim_orchestration / the decrement phase-split).
+    reused = getattr(_local, "conn", None) is not None and not _local.conn.closed
+    for attempt in range(1, _BARRIER_WRITE_ATTEMPTS + 1):
+        conn = _get_conn()
+        try:
+            with conn.cursor() as cur:
+                cur.execute(
+                    f"INSERT INTO {qualified('pg_batch_dedup')} "
+                    "(execution_id, batch_index, created_at) "
+                    "VALUES (%s, %s, now()) "
+                    "ON CONFLICT (execution_id, batch_index) DO NOTHING "
+                    "RETURNING execution_id",
+                    (execution_id, batch_index),
+                )
+                claimed = cur.fetchone() is not None
+        except Exception as exc:
+            conn_dead = _recover_after_error(conn, exc)
+            # Reused-conn execute-phase death (idle reap): the INSERT never
+            # committed, so the retry on a fresh conn is authoritative. A
+            # fresh-conn failure or a non-conn error propagates.
+            if conn_dead and reused and attempt < _BARRIER_WRITE_ATTEMPTS:
+                logger.warning(
+                    "[exec:%s] claim_batch(%d): execute failed on a cached "
+                    "connection (%s) — reconnecting and retrying once (attempt "
+                    "%d/%d); the INSERT never committed.",
+                    execution_id,
+                    batch_index,
+                    type(exc).__name__,
+                    attempt,
+                    _BARRIER_WRITE_ATTEMPTS,
+                    exc_info=True,
+                )
+                time.sleep(_BARRIER_RETRY_BACKOFF_SECONDS)
+                continue
+            raise
+        try:
+            conn.commit()
+        except Exception as exc:
+            # AMBIGUOUS commit — the dedup row may be persisted. Re-running could
+            # flip the RETURNING claim (skip-and-strand, or double-decrement), so
+            # NEVER retry; propagate to redelivery.
+            _recover_after_error(conn, exc)
+            logger.warning(
+                "[exec:%s] claim_batch(%d): commit failed (%s) — NOT retrying (the "
+                "server may already have committed; a re-run could flip the claim). "
+                "Propagating.",
+                execution_id,
+                batch_index,
+                type(exc).__name__,
+                exc_info=True,
+            )
+            raise
+        return claimed
+    # Unreachable: the loop either returns or raises.
+    raise AssertionError("claim_batch loop fell through")
 
 
 def clear_execution_batches(execution_id: str) -> int:
--- a/workers/queue_backend/pg_barrier.py
+++ b/workers/queue_backend/pg_barrier.py
@@ -183,7 +183,11 @@ def _cursor() -> Iterator[Any]:
 # dead conn, so a single retry runs against a freshly reconnected one. This turns
 # a transient blip (which previously aborted the whole execution at barrier
 # enqueue) into a self-heal. Kept a literal (not env-driven) so the idempotency
-# bound can't be weakened operationally.
+# bound can't be weakened operationally. Also bounds the phase-split RETURNING
+# claims (``claim_batch`` / ``try_claim_orchestration`` via
+# :func:`_run_returning_claim_with_reconnect`); they are non-idempotent but retry
+# ONLY the provably-uncommitted execute phase, so the same one-retry bound holds
+# without reopening a claim-flip.
 _BARRIER_WRITE_ATTEMPTS: Final = 2  # total attempts: 1 initial + 1 retry
 # Small fixed pause before the retry. The idle-reap case reconnects instantly
 # regardless; this only widens the self-heal window for a brief DB failover, and
@@ -223,8 +227,11 @@ def _run_idempotent_pre_dispatch_write(
     remaining - 1``): it is not idempotent — a re-applied decrement can fire the
     callback **prematurely with incomplete results**, or skip past 0 and
     **strand the barrier** to expiry — so it stays on the plain :func:`_cursor`
-    (recover-but-don't-retry). Same for ``claim_batch``, whose ``RETURNING``
-    answer flips on a retry.
+    (recover-but-don't-retry). The RETURNING claims (``claim_batch`` /
+    ``try_claim_orchestration``) are likewise non-idempotent — a committed claim's
+    win/lost answer flips on a naive re-run — so they retry via
+    :func:`_run_returning_claim_with_reconnect`, which retries ONLY the
+    provably-uncommitted execute phase.
     """
     for attempt in range(1, _BARRIER_WRITE_ATTEMPTS + 1):
         try:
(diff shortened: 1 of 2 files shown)


workers/queue_backend/pg_barrier.py
```diff
@@ -278,17 +278,77 @@ def claim_batch(execution_id: str, batch_index: int) -> bool:
     Called on the in-body PG path from :func:`run_batch_with_barrier` (claim at
     batch start), alongside the in-body decrement. The companion
     :func:`clear_execution_batches` reclaims the markers at barrier finalise.
+
+    **Execute-phase reconnect-retry** (mirrors :func:`try_claim_orchestration`,
```
muhammad-ali-e: **[Should fix] This PR makes two earlier docstrings in this file stale** (they're outside the diff, so anchoring here):

1. `_run_idempotent_pre_dispatch_write` (L225–227) still reads: *"…so it stays on the plain `_cursor` (recover-but-don't-retry). Same for `claim_batch`, whose `RETURNING` answer flips on a retry."* Before this PR that was the authoritative no-retry policy statement for `claim_batch`; now the file asserts in one place that `claim_batch` never retries and here that it retries once. Suggested rewording: *"Deliberately NOT used for the barrier decrement or the claims (`claim_batch` / `try_claim_orchestration`): they are non-idempotent — a re-applied decrement double-counts, and a re-run claim's committed `RETURNING` answer can flip — so each carries its own phase-split loop that retries ONLY an execute-phase failure on a cached connection."*

2. `_BARRIER_WRITE_ATTEMPTS` (L178–187) is documented as bounding "an IDEMPOTENT, pre-dispatch barrier write… so the idempotency bound can't be weakened", but it now also bounds two non-idempotent RETURNING claims (this one and `try_claim_orchestration` — drift pre-existed, this PR widens it). Worth a one-line addendum that it also bounds the phase-split claims, which retry only the provably-uncommitted execute phase.

workers/queue_backend/pg_barrier.py
```diff
@@ -278,17 +278,77 @@ def claim_batch(execution_id: str, batch_index: int) -> bool:
...
+    the execution-level claim one level up). This module's thread-local connection
+    sits idle between batches — a scheduled pipeline fires one only ~every 30 min,
+    well past PgBouncer's ``server_idle_timeout`` — so the FIRST ``INSERT`` after
+    the gap can hit an idle-reaped handle (``server closed the connection
+    unexpectedly``). Retrying a **reused-conn execute-phase** death is safe: the
+    INSERT never committed, so the retry's win/lost ``RETURNING`` answer is
+    authoritative. A **commit-phase** failure is ambiguous (the row may be
+    persisted) — re-running could flip the claim (``True`` → ``False`` → the caller
+    skips and strands the barrier, or a re-claim double-decrements), so it is NOT
```
muhammad-ali-e: **[Should fix] The "re-claim double-decrements" hazard is unreachable — drop the clause** (here, in the inline comment at L336–337, and in the test docstring at `test_pg_barrier.py` L561).

Walking the hypothetical commit-phase retry: if the commit **persisted**, the retry's `INSERT … ON CONFLICT DO NOTHING` finds the row → `False` → caller skips → barrier under-decremented (the strand — correctly documented). If it **didn't persist**, the retry returns `True` → exactly one decrement. The first attempt never returned `True` to the caller (the commit raised), so no path decrements twice. The sibling comment in `try_claim_orchestration` (L458–459) correctly cites only the winner→loser flip/strand.

The strand hazard alone fully justifies never-retry; stating an impossible failure mode is the kind of claim a future maintainer burns time trying to reproduce. Suggested: *"re-running could flip `True` → `False` — the caller would skip, the decrement would never run, and the barrier strands."*

workers/queue_backend/pg_barrier.py
```diff
@@ -278,17 +278,77 @@ def claim_batch(execution_id: str, batch_index: int) -> bool:
     Called on the in-body PG path from :func:`run_batch_with_barrier` (claim at
     batch start), alongside the in-body decrement. The companion
     :func:`clear_execution_batches` reclaims the markers at barrier finalise.
+
+    **Execute-phase reconnect-retry** (mirrors :func:`try_claim_orchestration`,
+    the execution-level claim one level up). This module's thread-local connection
+    sits idle between batches — a scheduled pipeline fires one only ~every 30 min,
```
muhammad-ali-e: **[Minor] Two rot-prone specifics in this sentence:**

- *"sits idle between batches"* is imprecise — batches within one execution run back-to-back; the long idle gap is between **pipeline runs** (each scheduled run's first `claim_batch` is the first-write-after-idle). A reader could infer batches within an execution are 30 minutes apart.
- *"~every 30 min"* is a property of one deployment's schedule, not the code, and appears in two files.

Suggested: *"sits idle between pipeline runs — scheduled pipelines can idle for tens of minutes, past PgBouncer's `server_idle_timeout` — so the first `INSERT` of a run can hit an idle-reaped handle."* Keep the concrete ~30-min incident detail in the test docstring (incident context belongs there), anchored by the ticket id UN-3684.

workers/queue_backend/pg_barrier.py
```diff
@@ -278,17 +278,77 @@ def claim_batch(execution_id: str, batch_index: int) -> bool:
...
-            "VALUES (%s, %s, now()) "
-            "ON CONFLICT (execution_id, batch_index) DO NOTHING "
-            "RETURNING execution_id",
-            (execution_id, batch_index),
-        )
-        return cur.fetchone() is not None
+    # Only a CACHED handle can be a stale idle-reap; sample before _get_conn()
+    # below materialises one, so a fresh-conn failure isn't misread as a reap
+    # (mirrors try_claim_orchestration / the decrement phase-split).
+    reused = getattr(_local, "conn", None) is not None and not _local.conn.closed
```
muhammad-ali-e, line 297: **[Suggested, latent] `reused` goes stale on attempt 2+ — one line makes the invariant structural.**

The flag is sampled once but tested every attempt. With `_BARRIER_WRITE_ATTEMPTS = 2` the `attempt < N` guard masks it, but if the constant is ever bumped to 3, attempt 2 runs on a **fresh** connection while `reused` still reads `True` — a fresh-conn death would be retried (violating the documented "fresh-conn failures propagate" policy) with a log claiming a cached connection. Correctness currently depends on an unstated coupling to the constant's value.

Fix: `reused = False` immediately before the `continue` at L330 (or resample at the top of the loop — after a discard `_local.conn` is `None`, so attempt 2 correctly reads `False`). Same latent shape pre-exists at `try_claim_orchestration` L422 — worth fixing both for symmetry, or a comment tying safety to `_BARRIER_WRITE_ATTEMPTS == 2`.

workers/queue_backend/pg_barrier.py
```diff
@@ -278,17 +278,77 @@ def claim_batch(execution_id: str, batch_index: int) -> bool:
...
+            raise
+        try:
+            conn.commit()
+        except Exception as exc:
+            # AMBIGUOUS commit — the dedup row may be persisted. Re-running could
+            # flip the RETURNING claim (skip-and-strand, or double-decrement), so
+            # NEVER retry; propagate to redelivery.
+            _recover_after_error(conn, exc)
+            logger.warning(
+                "[exec:%s] claim_batch(%d): commit failed (%s) — NOT retrying (the "
```
muhammad-ali-e: **[Minor, observability] Close the loop for the operator on the persisted-row half of the ambiguity.**

"Propagating" (and the docstring's "propagates to vt-expiry redelivery") reads as if redelivery repairs it. But if the commit actually persisted server-side, the redelivery's `claim_batch` returns `False` → `skipped_redelivery` → the barrier is never decremented for this batch and hangs until the **reaper** recovers it at expiry. Suggested addendum to the message: *"if the row was persisted, the redelivery will skip and the reaper recovers the barrier at expiry"* — so an operator staring at a stranded barrier can connect it to this WARNING.

workers/tests/test_pg_barrier.py
```diff
@@ -550,6 +550,76 @@ def test_schema_behind_raises_actionable_error(self, _clean_local, exc_cls):
             try_claim_orchestration("exec-1", "org-1")
 
 
+class TestClaimBatchRetry:
```
muhammad-ali-e: **[Suggested] Two cheap tests would close the real gaps here (both have direct precedents in sibling classes); coverage otherwise strong (7.5/10).**

1. **Retry exhaustion** — no test pins that a second conn-dead failure propagates. Boundedness rests solely on `attempt < _BARRIER_WRITE_ATTEMPTS`; an off-by-one (`<=`) or `reused`-refresh bug would add a third attempt or hit the `AssertionError` fall-through, and all three current tests still pass. Precedent: `TestDecrementPhaseSplitRetry::test_reraises_after_one_retry` (L399). Shape: cached dead `_FakeConn`, `create_pg_connection` returns a second failing `_FakeConn`; assert `pytest.raises(OperationalError)`, `executes == 1` on each conn, `sleeps == [pg_barrier._BARRIER_RETRY_BACKOFF_SECONDS]`.

2. **Non-connection execute error on a reused conn is NOT retried** — the only guard against retrying a real SQL error is the `conn_dead` conjunct; a regression dropping `conn_dead and` from the gate would pass every current test. Precedents: `TestIdempotentWriteRetry` L253, `TestDecrementPhaseSplitRetry` L379. Shape: `_local.conn = _FakeConn(execute_error=psycopg2.DataError(...))`, `create_pg_connection = pytest.fail`; expect `DataError`, `sleeps == []`.

(Optional 3rd: the happy-retry test can only ever see `claimed is True` because the `MagicMock` cursor's `fetchone()` is always truthy — the "win/**lost** answer is authoritative" half is untested. Needs a small `fetchone_result=None` option on `_FakeConn`.)

workers/tests/test_pg_barrier.py
```diff
@@ -550,6 +550,76 @@ def test_schema_behind_raises_actionable_error(self, _clean_local, exc_cls):
...
+    """``claim_batch`` is a RETURNING claim, sibling to ``try_claim_orchestration``
+    one level up, so it self-heals ONLY the provably-safe case: an execute-phase
+    failure on a *cached* connection (the idle reap that stranded scheduled ETL
+    batches — the ~30-min-idle barrier connection past ``server_idle_timeout``; the
+    INSERT never committed, so the retry's claim answer is authoritative). A
+    commit-phase failure is ambiguous and a fresh-conn failure is a real error —
```
muhammad-ali-e: **[Minor] The docstring conflates the two no-retry rationales** — the flip hazard only applies to the commit-phase case. Retrying a fresh-conn *execute* failure could not flip anything (nothing committed); it isn't retried because a reconnect buys nothing against a genuine DB error (as `_apply_decrement`'s docstring states precisely). And "(or double-decrement)" on L561 is unreachable (see the source-side comment).

Suggested: *"A commit-phase failure is ambiguous — retrying could flip `True`→`False` and strand the barrier. A fresh-conn failure is a real DB error — reconnecting buys nothing. Neither is retried."*

muhammad-ali-e (commented):
## Comprehensive PR review (6 review agents: code review, silent-failure, type design, test coverage, comment accuracy, simplification)

**Verdict: the change is correct.** The `claim_batch` phase-split faithfully mirrors `try_claim_orchestration` — `reused` sampled before the first `_get_conn()`, retry gated on `conn_dead and reused and attempt < _BARRIER_WRITE_ATTEMPTS`, commit-phase never retried, fresh-conn/non-conn errors propagate, `AssertionError` fall-through guard. No silent failures: every catch retries once (bounded, `exc_info=True`) or re-raises, and the propagated exception surfaces at the PG consumer as ERROR with the row left for vt-expiry redelivery. Full barrier + dedup suites pass (100/100).

**No critical or important functional issues.** The inline comments below are prioritised:

1. **Should fix — stale docstrings this PR contradicts** (`_run_idempotent_pre_dispatch_write` L225–227 still says `claim_batch` "stays on the plain `_cursor` (recover-but-don't-retry)"; `_BARRIER_WRITE_ATTEMPTS` L178–187 still says "IDEMPOTENT").
2. **Should fix — the "double-decrement" hazard cited at L291/L336–337 and in the test docstring is unreachable**; the strand is the real (and sufficient) hazard.
3. **Suggested — harden the `reused` flag** against a future `_BARRIER_WRITE_ATTEMPTS` bump (one line).
4. **Suggested — two cheap tests with sibling precedents**: retry exhaustion, non-connection error not retried.
5. **Minor wording/observability** — the "~30 min between batches" figure, the commit-failure log breadcrumb, the test-docstring rationale conflation.

**Explicitly considered and endorsed as-is:** the bespoke loop + unreachable `AssertionError` (established file idiom; a shared phase-split helper across the three copies is worthwhile **follow-up**, not this PR), the `-> bool` contract, calling `claim_batch` outside `run_batch_with_barrier`'s try block, and the commit-ambiguity → strand → reaper design.

**Strengths:** the `reused` pre-sampling comment is load-bearing and accurate; tests assert behaviour (per-conn execute/commit counts, `sleeps` as a retry-count witness, `create_pg_connection → pytest.fail` making "no reconnect" a hard assertion) rather than internals; log messages honestly say the server *may* have committed rather than overclaiming.

muhammad-ali-e:
Thanks — all addressed in `afc8b03cf`, and it converged nicely with the SonarCloud gate.

### SonarCloud (15.9% duplication on new code)
`claim_batch`'s phase-split loop was a near-copy of `try_claim_orchestration`. Extracted the shared retry into **`_run_returning_claim_with_reconnect(operation, what)`**; both are now thin wrappers. One loop → duplication gone, and it fixes the two callers' shared points in one place. The sibling's 5 existing tests still pass, confirming behavior is preserved.

### Should-fix
- **Stale docstrings (`pg_barrier.py:282`)** — `_run_idempotent_pre_dispatch_write` ("Same for `claim_batch`, whose RETURNING answer flips") and `_BARRIER_WRITE_ATTEMPTS` (IDEMPOTENT-only) both updated to reflect that the RETURNING claims now retry via the phase-split helper (execute-phase only).
- **Unreachable "re-claim double-decrements" (`:291`)** — dropped in all three places (docstring, inline comment, test docstring). The first attempt never returned `True` to the caller, so no path double-decrements; the strand hazard alone justifies never-retry.

### Suggested
- **`reused` goes stale (`:298`)** — now `reused = False` before the retry `continue` (in the helper), so a fresh-conn death on a later attempt is never retried — structural, not masked by `_BARRIER_WRITE_ATTEMPTS == 2`. Fixes both callers at once.
- **Commit-phase observability (`:340`)** — the WARNING now spells out: if the row persisted, the redelivery's claim skips and the reaper recovers the barrier at expiry — so an operator staring at a stranded barrier connects it to this log.
- **Two tests (`test:553`)** — added `test_second_conn_dead_failure_propagates` (retry bounded to one, no fall-through) and `test_non_connection_execute_error_on_reused_conn_not_retried` (pins the `conn_dead` guard). Also did the optional 3rd — added a `fetchone_result` option to `_FakeConn` and `test_retry_returns_the_lost_answer_authoritatively`, so the "win/**lost** answer is authoritative" half is now covered (not just the truthy-MagicMock default).

### Minor
- **Docstring precision (`:284`)** — "idle **between pipeline runs**" (not "between batches"); the concrete ~30-min incident detail now lives only in the test docstring, anchored to UN-3684.
- **Test docstring (`:559`)** — reworded to separate the two distinct no-retry rationales (commit-phase = flip/strand; fresh-conn = reconnect buys nothing), double-decrement removed.

Full `pg_barrier` suite: **94 passed** (was 91).

muhammad-ali-e:
Follow-up (`a6692cbb2`): the source dedup pushed Sonar's duplication into the **test** file (16.5% — `TestClaimBatchRetry` mirrored `TestClaimOrchestrationRetry`). Collapsed both into one **parametrized** `TestReturningClaimReconnect` over both claim callers, so each retry assertion is authored once; the schema-behind test stays in its own `TestClaimOrchestrationSchemaGuard`. Coverage went **up** (94→97) — the newer edge tests (retry-exhaustion, non-conn-not-retried, authoritative lost answer) now run for `try_claim_orchestration` too. Both source- and test-side duplication now removed.

muhammad-ali-e:
@greptile-apps addressed in the latest commit: the execute-phase retry happy-path test is now parametrized over **both** `_CONN_DEAD_ERRORS` symptoms — `OperationalError` and `InterfaceError` ("connection already closed") — mirroring `TestDecrementPhaseSplitRetry`, and it runs across both claim callers via the shared `TestReturningClaimReconnect`. So the idle-reap-as-InterfaceError path is now pinned for `claim_batch` (and `try_claim_orchestration`). Suite: 99 passed. Thanks for the merge-safety read.

PR #2149, opened 2026-07-07
UN-3690 [FIX] PG scheduler — recompute next_run_at on cron edit (Beat parity)

muhammad-ali-e:
### Problem
Editing a `pg_owned` pipeline's schedule time didn't take effect until one stale cycle — the pipeline fired once more at the **old** cron's `next_run_at`, ignoring the new time.

Confirmed on integration via GCP logs (pipeline `438ac6d8`): baselined to `04:00` off the original `0 * * * *`; edits to `12`/`52`/`58` were ignored; it fired at the **stale** `04:00`, and only then recomputed the next one off the new cron.

### Root cause
`mirror_periodic_schedule_upsert` upserts `cron_string` but **never touches `next_run_at`** — which the PG scheduler fires on. Celery Beat has no such staleness: it recomputes due-ness live from the crontab each tick, so an edit takes effect within one beat interval. So this is a **PG-only, flag-gated (`pg_owned`) Beat-parity regression** — not present on staging/prod (still Beat).

### Fix
When the cron **changes** on an **already-baselined** row (`next_run_at IS NOT NULL`), retarget `next_run_at` to the new cron's next match. Deliberately left untouched (NULL) for:
- a **brand-new** row — the scheduler's no-burst baseline is correct (a new schedule fires at its next match);
- a **Beat→PG hand-over** — ownership nulls it.

The `next_run_at IS NOT NULL` guard means the branch only ever fires for a row the PG scheduler has already baselined — i.e. an actively PG-owned schedule — so **Celery-only rows are never touched**. Still best-effort (swallowed on error) and never touches the Beat `PeriodicTask`/`CrontabSchedule`.

### Testing
- **5 new mirror unit tests**: retarget-on-change, no-touch-when-cron-unchanged, NULL-left-for-baseline, new-row-no-set, invalid-cron-still-upserts. Existing mirror suite green (17 total).
- **Real-DB dev-test** (sqlite in-memory, hard-guarded off the real DB): exercised the actual function end-to-end (real ORM + croniter + model) — all 5 scenarios pass, including the Celery-untouched (`NULL` stays `NULL`) property.

### Live confirm after deploy
Edit a `pg_owned` pipeline's cron → `pg_periodic_schedule.next_run_at` reflects the new cron's next match immediately, and it fires at the intended minute (no stale cycle).

### DIFF
--- a/backend/scheduler/tasks.py
+++ b/backend/scheduler/tasks.py
@@ -34,6 +34,49 @@ logger = logging.getLogger(__name__)
 # ---------------------------------------------------------------------------
 
 
+def _retargeted_next_run_at(pipeline_id: str, cron_string: str) -> datetime | None:
+    """The ``next_run_at`` a cron EDIT should set for Beat parity, or ``None`` to
+    leave it as the PG scheduler's baseline owns it.
+
+    Returns a recomputed next-match ONLY when ``cron_string`` differs from an
+    already-baselined mirror row (``next_run_at`` set). Returns ``None`` for a
+    brand-new row (no mirror yet) and a not-yet-baselined one (``next_run_at``
+    NULL) so the scheduler's no-burst baseline records the first next-time.
+
+    Fully guarded on purpose: this is an OPTIONAL enhancement over the mandatory
+    ``cron_string``/``enabled`` mirror write, so a read failure, a ``None``/invalid
+    cron (normally rejected upstream by ``PipelineSerializer.validate_cron_string``
+    in ``pipeline_v2``), or a croniter error must degrade to ``None`` — never take
+    the base mirror write down with it.
+    """
+    try:
+        existing = (
+            PgPeriodicSchedule.objects.filter(pipeline_id=pipeline_id)
+            .values("cron_string", "next_run_at")
+            .first()
+        )
+        if (
+            existing is None
+            or existing["next_run_at"] is None
+            or existing["cron_string"] == cron_string
+        ):
+            return None
+        return croniter(cron_string, timezone.now()).get_next(datetime)
+    except Exception as exc:
+        # Log the ACTUAL cause (bad read; cron=None → AttributeError;
+        # CroniterBadCronError / CroniterBadDateError; a croniter API change) — a
+        # generic "could not recompute" is a dead end when debugging a stale fire.
+        logger.warning(
+            "pg_periodic_schedule: could not retarget next_run_at for pipeline %s "
+            "(cron %r): %s",
+            pipeline_id,
+            cron_string,
+            exc,
+            exc_info=True,
+        )
+        return None
+
+
 def mirror_periodic_schedule_upsert(
     *,
     pipeline_id: str,
@@ -44,46 +87,26 @@ def mirror_periodic_schedule_upsert(
     enabled: bool,
 ) -> None:
     try:
-        defaults: dict = {
+        defaults: dict[str, Any] = {
             "organization_id": organization_id or "",
             "workflow_id": workflow_id or None,
             "pipeline_name": pipeline_name or "",
             "cron_string": cron_string,
             "enabled": enabled,
         }
-        # Beat parity: when the cron CHANGES on an already-baselined row
-        # (next_run_at set), retarget next_run_at to the new cron's next match so
-        # the edit takes effect this cycle. The PG scheduler fires on the
-        # materialised next_run_at; Beat instead recomputes due-ness live from the
-        # crontab each tick, so without this the pipeline fires once more at the
-        # STALE old-cron time and the edit is ignored for a cycle (UN-3690).
-        # Deliberately left untouched (NULL) for a brand-new row — the scheduler's
-        # no-burst baseline is correct there (a new schedule fires at its next
-        # match) — and for a Beat→PG hand-over, where ownership nulls it.
-        existing = (
-            PgPeriodicSchedule.objects.filter(pipeline_id=pipeline_id)
-            .values("cron_string", "next_run_at")
-            .first()
-        )
-        if (
-            existing is not None
-            and existing["next_run_at"] is not None
-            and existing["cron_string"] != cron_string
-        ):
-            try:
-                defaults["next_run_at"] = croniter(cron_string, timezone.now()).get_next(
-                    datetime
-                )
-            except Exception:
-                # Unparseable cron (should be caught by the serializer's validation
-                # upstream) — leave next_run_at stale rather than block the mirror;
-                # the scheduler quiesces a bad cron on its own tick.
-                logger.warning(
-                    "pg_periodic_schedule: could not recompute next_run_at for "
-                    "pipeline %s (cron %r) — leaving the existing value",
-                    pipeline_id,
-                    cron_string,
-                )
+        # Beat parity: a cron EDIT on an already-baselined row must retarget
+        # next_run_at to the new cron's next match, else the PG scheduler fires
+        # once more at the STALE old-cron time — Beat has no such staleness, it
+        # recomputes due-ness live from the crontab each tick (UN-3690). Left
+        # untouched (NULL) for a brand-new row — the scheduler's no-burst baseline
+        # is correct there (a new schedule fires at its next match) — and for a
+        # Beat→PG hand-over, where next_run_at is already NULL (never set while
+        # Beat-owned; ownership clears it only on rollback to Beat, see
+        # reconcile_ownership_for). Computed via a fully-guarded helper so it can
+        # never take down the mandatory cron_string/enabled mirror write below.
+        next_run_at = _retargeted_next_run_at(pipeline_id, cron_string)
+        if next_run_at is not None:
+            defaults["next_run_at"] = next_run_at
         PgPeriodicSchedule.objects.update_or_create(
             pipeline_id=pipeline_id,
             defaults=defaults,
--- a/backend/scheduler/tasks.py
+++ b/backend/scheduler/tasks.py
@@ -1,9 +1,11 @@
 import json
 import logging
 import traceback
+from datetime import datetime
 from typing import Any
 
 from celery import shared_task
+from croniter import croniter
 from django.db import transaction
 from django.utils import timezone
 from django_celery_beat.models import CrontabSchedule, PeriodicTask
@@ -42,15 +44,49 @@ def mirror_periodic_schedule_upsert(
     enabled: bool,
 ) -> None:
     try:
+        defaults: dict = {
+            "organization_id": organization_id or "",
+            "workflow_id": workflow_id or None,
+            "pipeline_name": pipeline_name or "",
+            "cron_string": cron_string,
+            "enabled": enabled,
+        }
+        # Beat parity: when the cron CHANGES on an already-baselined row
+        # (next_run_at set), retarget next_run_at to the new cron's next match so
+        # the edit takes effect this cycle. The PG scheduler fires on the
+        # materialised next_run_at; Beat instead recomputes due-ness live from the
+        # crontab each tick, so without this the pipeline fires once more at the
+        # STALE old-cron time and the edit is ignored for a cycle (UN-3690).
+        # Deliberately left untouched (NULL) for a brand-new row — the scheduler's
+        # no-burst baseline is correct there (a new schedule fires at its next
+        # match) — and for a Beat→PG hand-over, where ownership nulls it.
+        existing = (
+            PgPeriodicSchedule.objects.filter(pipeline_id=pipeline_id)
+            .values("cron_string", "next_run_at")
+            .first()
+        )
+        if (
+            existing is not None
+            and existing["next_run_at"] is not None
+            and existing["cron_string"] != cron_string
+        ):
+            try:
+                defaults["next_run_at"] = croniter(cron_string, timezone.now()).get_next(
+                    datetime
+                )
+            except Exception:
+                # Unparseable cron (should be caught by the serializer's validation
+                # upstream) — leave next_run_at stale rather than block the mirror;
+                # the scheduler quiesces a bad cron on its own tick.
+                logger.warning(
+                    "pg_periodic_schedule: could not recompute next_run_at for "
+                    "pipeline %s (cron %r) — leaving the existing value",
+                    pipeline_id,
+                    cron_string,
+                )
         PgPeriodicSchedule.objects.update_or_create(
             pipeline_id=pipeline_id,
-            defaults={
-                "organization_id": organization_id or "",
-                "workflow_id": workflow_id or None,
-                "pipeline_name": pipeline_name or "",
-                "cron_string": cron_string,
-                "enabled": enabled,
-            },
+            defaults=defaults,
         )
     except Exception:
         logger.exception(
--- a/backend/scheduler/tests/test_pg_periodic_schedule_mirror.py
+++ b/backend/scheduler/tests/test_pg_periodic_schedule_mirror.py
@@ -13,8 +13,6 @@ from datetime import datetime
 from datetime import timezone as dt_timezone
 from unittest.mock import MagicMock, patch
 
-from croniter import croniter
-
 from scheduler import tasks
 from scheduler.helper import SchedulerHelper
 
@@ -28,10 +26,18 @@ class _DoesNotExist(Exception):
     """Stand-in for PeriodicTask.DoesNotExist when the model is mocked."""
 
 
+def _stub_existing_row(sched, row=None):
+    """Stub the ``.filter().values().first()`` read of the existing mirror row
+    (``None`` = no row yet). The one place the read shape lives, so a change to how
+    the code reads the row (e.g. ``.only()`` instead of ``.values()``) is a
+    one-line edit here."""
+    sched.objects.filter.return_value.values.return_value.first.return_value = row
+
+
 class TestUpsertMirror:
     def test_upserts_with_given_fields(self):
         with patch("scheduler.tasks.PgPeriodicSchedule") as sched:
-            sched.objects.filter.return_value.values.return_value.first.return_value = None
+            _stub_existing_row(sched)
             tasks.mirror_periodic_schedule_upsert(
                 pipeline_id=_PIPELINE_ID,
                 organization_id=_ORG,
@@ -51,7 +57,7 @@ class TestUpsertMirror:
 
     def test_disabled_pipeline_mirrors_enabled_false(self):
         with patch("scheduler.tasks.PgPeriodicSchedule") as sched:
-            sched.objects.filter.return_value.values.return_value.first.return_value = None
+            _stub_existing_row(sched)
             tasks.mirror_periodic_schedule_upsert(
                 pipeline_id=_PIPELINE_ID,
                 organization_id=_ORG,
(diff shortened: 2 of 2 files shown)


backend/scheduler/tasks.py
```diff
@@ -42,15 +44,49 @@ def mirror_periodic_schedule_upsert(
...
+        ):
+            try:
+                defaults["next_run_at"] = croniter(cron_string, timezone.now()).get_next(
+                    datetime
+                )
+            except Exception:
+                # Unparseable cron (should be caught by the serializer's validation
+                # upstream) — leave next_run_at stale rather than block the mirror;
+                # the scheduler quiesces a bad cron on its own tick.
+                logger.warning(
```
muhammad-ali-e: **[Important — silent-failure-hunter]** This warning drops the exception entirely, and `except Exception` catches far more than a parse error: `cron_string=None` reaches here as `AttributeError` (the upstream guard in `helper.py` only checks key *presence*), a croniter API regression raises `TypeError`, and `CroniterBadDateError` (valid syntax, no matching date, e.g. `0 0 30 2 *`) is a distinct cause. In every case the operator sees the same "could not recompute" message implying a bad cron — a dead end when debugging why a pipeline fired at a stale time. Compare `PipelineSerializer._validate_basic_cron_format`, which logs the error.

```python
except Exception as exc:
    logger.warning(
        "pg_periodic_schedule: could not recompute next_run_at for "
        "pipeline %s (cron %r) — leaving the existing value %s: %s",
        pipeline_id,
        cron_string,
        existing["next_run_at"],
        exc,
    )
```

Including the stale `existing["next_run_at"]` tells the operator exactly when the pipeline will still fire. A `caplog` assertion in `test_invalid_cron_leaves_stale_next_run_at_but_still_upserts` would pin the log so a refactor can't silently drop it.

backend/scheduler/tasks.py
```diff
@@ -42,15 +44,49 @@ def mirror_periodic_schedule_upsert(
...
+        # Beat parity: when the cron CHANGES on an already-baselined row
+        # (next_run_at set), retarget next_run_at to the new cron's next match so
+        # the edit takes effect this cycle. The PG scheduler fires on the
+        # materialised next_run_at; Beat instead recomputes due-ness live from the
+        # crontab each tick, so without this the pipeline fires once more at the
+        # STALE old-cron time and the edit is ignored for a cycle (UN-3690).
+        # Deliberately left untouched (NULL) for a brand-new row — the scheduler's
+        # no-burst baseline is correct there (a new schedule fires at its next
+        # match) — and for a Beat→PG hand-over, where ownership nulls it.
+        existing = (
```
muhammad-ali-e: **[Important — silent-failure-hunter / pr-test-analyzer]** Structural regression vs. before: previously only `update_or_create` could fail; now a failure in this read (or the recompute) aborts the **mandatory** `enabled`/`cron_string` mirror write too, because the outer swallow catches everything. In practice the delta is small (a realistic read failure would also fail the write, and the outer handler logs with traceback), but the optional enhancement shouldn't be able to take down the base behaviour. Cheap hardening consistent with the schema-drift-guard style used elsewhere in this feature branch: wrap the read+recompute (lines 63–86) in its own try/except that logs and falls through to the plain upsert — a read failure then degrades to exactly the pre-PR behaviour.

Also: this is the one new code path with zero failure-mode coverage. `test_failure_is_swallowed` only covers `update_or_create` raising — add a sibling test with `sched.objects.filter.side_effect = RuntimeError("db down")` asserting no raise, to pin the "never break Beat" guarantee at the read seam.

backend/scheduler/tasks.py
```diff
@@ -42,15 +44,49 @@ def mirror_periodic_schedule_upsert(
...
+                # Unparseable cron (should be caught by the serializer's validation
+                # upstream) — leave next_run_at stale rather than block the mirror;
+                # the scheduler quiesces a bad cron on its own tick.
+                logger.warning(
+                    "pg_periodic_schedule: could not recompute next_run_at for "
+                    "pipeline %s (cron %r) — leaving the existing value",
+                    pipeline_id,
+                    cron_string,
+                )
         PgPeriodicSchedule.objects.update_or_create(
```
muhammad-ali-e, line 110: **[Suggestion — code-reviewer / type-design-analyzer]** Read-then-write race: the `existing` read (lines 63–67) and this `update_or_create` are not in one transaction, so a concurrent scheduler tick can interleave. Worst case: the mirror reads `next_run_at IS NULL`, the tick baselines to the *old* cron's next match, the mirror then writes without the key — i.e. the stale-for-one-cycle behaviour this PR fixes survives for that rare window, and it self-heals next cycle. Acceptable for a best-effort mirror; if you ever want to close it, `transaction.atomic()` + `select_for_update()` matches the pattern `enable_task` already uses (tasks.py:296–304).

backend/scheduler/tasks.py
```diff
@@ -42,15 +44,49 @@ def mirror_periodic_schedule_upsert(
...
+        }
+        # Beat parity: when the cron CHANGES on an already-baselined row
+        # (next_run_at set), retarget next_run_at to the new cron's next match so
+        # the edit takes effect this cycle. The PG scheduler fires on the
+        # materialised next_run_at; Beat instead recomputes due-ness live from the
+        # crontab each tick, so without this the pipeline fires once more at the
+        # STALE old-cron time and the edit is ignored for a cycle (UN-3690).
+        # Deliberately left untouched (NULL) for a brand-new row — the scheduler's
+        # no-burst baseline is correct there (a new schedule fires at its next
+        # match) — and for a Beat→PG hand-over, where ownership nulls it.
```
muhammad-ali-e: **[Suggestion — comment-analyzer]** "where ownership nulls it" misattributes the mechanism: in `reconcile_ownership_for` (scheduler/ownership.py:104–107), `next_run_at` is nulled only on the **rollback** branch (`if not pg_owned`) — i.e. on a PG→Beat hand-back. On the Beat→PG hand-over branch nothing touches `next_run_at`; it is NULL at hand-over because it was never set while Beat-owned (or an earlier rollback cleared it). A maintainer reading this and then the `pg_owned=True` path will find no nulling and conclude the comment is wrong. Suggested rewording:

> …and for a Beat→PG hand-over, where next_run_at is already NULL (never set while Beat-owned; ownership clears it on rollback to Beat — see reconcile_ownership_for).

backend/scheduler/tasks.py
```diff
@@ -42,15 +44,49 @@ def mirror_periodic_schedule_upsert(
...
+            existing is not None
+            and existing["next_run_at"] is not None
+            and existing["cron_string"] != cron_string
+        ):
+            try:
+                defaults["next_run_at"] = croniter(cron_string, timezone.now()).get_next(
+                    datetime
+                )
+            except Exception:
+                # Unparseable cron (should be caught by the serializer's validation
```
muhammad-ali-e: **[Suggestion — comment-analyzer]** The claim is accurate but the reference is ambiguous: the validation that actually guards this path is `PipelineSerializer.validate_cron_string` → `_validate_basic_cron_format` (pipeline_v2/serializers/crud.py:90–105). The *nearest* serializer, `scheduler/serializer.py` (`AddJobSerializer`), validates only length — a reader checking there will think this comment is wrong. Suggest naming it: "should be caught upstream by PipelineSerializer.validate_cron_string (pipeline_v2)".

backend/scheduler/tasks.py
```diff
@@ -42,15 +44,49 @@ def mirror_periodic_schedule_upsert(
...
+            .values("cron_string", "next_run_at")
+            .first()
+        )
+        if (
+            existing is not None
+            and existing["next_run_at"] is not None
+            and existing["cron_string"] != cron_string
+        ):
+            try:
+                defaults["next_run_at"] = croniter(cron_string, timezone.now()).get_next(
```
muhammad-ali-e: **[Suggestion — code-simplifier]** Optional: extracting the recompute into a small module-level helper flattens the two-level try nesting and separates the parse-failure concern from the outer "never break Beat" guard, matching the file's existing small-helper style (`_mirror_periodic_schedule_set_enabled` / `_delete`):

```python
def _next_run_for_cron_edit(pipeline_id: str, cron_string: str) -> datetime | None:
    try:
        return croniter(cron_string, timezone.now()).get_next(datetime)
    except Exception:
        logger.warning(...)
        return None
```

Then at the call site: `if (next_run := _next_run_for_cron_edit(pipeline_id, cron_string)) is not None: defaults["next_run_at"] = next_run`. Behaviour identical; pairs well with the log-the-exception fix suggested above.

backend/scheduler/tasks.py
```diff
@@ -42,15 +44,49 @@ def mirror_periodic_schedule_upsert(
     enabled: bool,
 ) -> None:
     try:
+        defaults: dict = {
```
muhammad-ali-e: **[Nit — type-design-analyzer]** Bare `dict` is `dict[Any, Any]`; tighten to `defaults: dict[str, Any]`. (A TypedDict is *not* recommended — `update_or_create(defaults=...)` is inherently stringly-keyed at the Django boundary and a TypedDict wouldn't be validated against the model's fields anyway.)

backend/scheduler/tests/test_pg_periodic_schedule_mirror.py
```diff
@@ -70,6 +77,82 @@ def test_failure_is_swallowed(self):
             )
 
 
+class TestUpsertNextRunRecompute:
```
muhammad-ali-e, line 86: **[Important — pr-test-analyzer]** These tests never run in CI: `tests/groups.yaml` (`unit-backend.paths`, lines ~106–118) does not include `scheduler/tests`, no other lane references it, and a plain `pytest` run dies with `AppRegistryNotReady` (`pytest-django` is not a backend dependency and no conftest calls `django.setup()`). The tests do pass — 17/17 in this file, 28/28 in `backend/scheduler/tests/` — but only with a hand-injected Django bootstrap. Pre-existing gap, but this PR's entire regression net for a scheduler-correctness fix is inert until `scheduler/tests` is added to a CI lane (plus a Django setup shim). Worth fixing or explicitly acknowledging before merge.

backend/scheduler/tests/test_pg_periodic_schedule_mirror.py
```diff
@@ -70,6 +77,82 @@ def test_failure_is_swallowed(self):
...
+    def test_cron_change_on_baselined_row_retargets_next_run_at(self):
+        now = datetime(2026, 1, 1, 8, 0, tzinfo=dt_timezone.utc)
+        with (
+            patch("scheduler.tasks.PgPeriodicSchedule") as sched,
+            patch("scheduler.tasks.timezone.now", return_value=now),
+        ):
+            self._mock_existing(sched, cron_string="0 9 * * *", next_run_at=self._OLD)
+            self._upsert("0 10 * * *")  # edited to a new time
+        # recomputed from the NEW cron + now (not the stale 09:00)
+        assert self._defaults(sched)["next_run_at"] == croniter(
```
muhammad-ali-e: **[Suggestion — pr-test-analyzer]** Self-referential assertion: the expected value is derived with the exact `croniter(...).get_next(datetime)` expression the implementation uses, so it can't catch croniter misuse — e.g. if both sides went naive, it would still pass, and a naive value saved to the aware `next_run_at` field under `USE_TZ=True` only warns. With `now` frozen at 08:00 UTC the expectation is a known constant; assert it directly:

```python
assert self._defaults(sched)["next_run_at"] == datetime(
    2026, 1, 1, 10, 0, tzinfo=dt_timezone.utc
)
```

This pins both the value and tz-awareness. Nice-to-have: also assert `defaults["cron_string"] == "0 10 * * *"` here, pinning that the retargeted time ships *with* the new cron in the same upsert.

backend/scheduler/tests/test_pg_periodic_schedule_mirror.py
```diff
@@ -70,6 +77,82 @@ def test_failure_is_swallowed(self):
...
+    the stale old-cron time. Only for an already-baselined row: a fresh row and a
+    Beat→PG hand-over keep ``next_run_at`` NULL for the scheduler's no-burst
+    baseline.
+    """
+
+    _OLD = datetime(2026, 1, 1, 9, 0, tzinfo=dt_timezone.utc)
+
+    @staticmethod
+    def _mock_existing(sched, *, cron_string, next_run_at):
+        sched.objects.filter.return_value.values.return_value.first.return_value = (
```
muhammad-ali-e: **[Suggestion — code-simplifier]** The deep chain `sched.objects.filter.return_value.values.return_value.first.return_value = ...` is now written four times (verbatim in the three `TestUpsertMirror` tests at lines 34/54/67, and again here). A single module-level helper collapses it, so a future change to how the code reads the existing row (e.g. `.only()` instead of `.values()`) is a one-line edit:

```python
def _stub_existing_row(sched, row=None):
    """Stub the `.filter().values().first()` read of the existing mirror row."""
    sched.objects.filter.return_value.values.return_value.first.return_value = row
```

It also lets you drop `_mock_existing`'s non-obvious `cron_string is None → None` ternary — the new-row case becomes a literal `_stub_existing_row(sched)`.

muhammad-ali-e (commented):
## Multi-agent PR review (code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier)

**Verdict: the fix is correct.** Timezone handling verified end-to-end (`USE_TZ=True`/UTC, croniter over an aware base returns aware UTC, matching the PG scheduler tick's own `compute_next_run` semantics), the `next_run_at IS NOT NULL` guard correctly selects only baselined/PG-owned rows, and all 28 scheduler tests pass locally. No critical correctness issues found.

### Prioritised findings (inline comments below)

**Important**
1. `tasks.py:81` — the invalid-cron warning omits the exception itself, so any non-parse failure (`AttributeError` from a `None` cron, `TypeError`, `CroniterBadDateError`) is misattributed to a bad cron.
2. `tasks.py:63` — a failure in the new read/recompute (an optional enhancement) now aborts the mandatory `enabled`/`cron_string` mirror write; the read-failure path is also untested.
3. `test_pg_periodic_schedule_mirror.py:80` — these tests are not wired into any CI lane (`tests/groups.yaml` `unit-backend.paths` does not include `scheduler/tests`, and `pytest-django`/a Django bootstrap is missing), so the regression net for this fix is inert.

**Suggestions**
4. `tasks.py:87` — non-transactional read-then-write can race a concurrent scheduler tick (bounded to the pre-PR stale-fire for one cycle; optional `select_for_update`).
5. `tasks.py:62` — comment misattributes the mechanism: ownership nulls `next_run_at` on the PG→Beat rollback branch, not on the Beat→PG hand-over.
6. `tasks.py:78` — name `PipelineSerializer.validate_cron_string` explicitly (the nearer `scheduler/serializer.py` does not validate crons).
7. `tasks.py:74` — optionally extract the croniter recompute into a small helper to flatten the nested try/except.
8. `tests:122` — the recompute assertion re-derives the expected value with the same croniter expression the implementation uses; assert the concrete datetime instead.
9. `tests:92` — the deep `.filter().values().first()` mock chain is now written four times; a shared module-level helper would make future ORM changes a one-line edit.
10. `tasks.py:47` — nit: `defaults: dict` → `dict[str, Any]`.

### Verified strengths
- All four branches of the (existing, baselined, cron-changed) decision matrix plus the invalid-cron fallback are covered by the new tests, which correctly assert **key absence** in `defaults` (matching `update_or_create` semantics).
- The block comment's load-bearing claims were verified against the code: the PG scheduler does fire on the materialised `next_run_at` (`pg_scheduler.py:144-154`), `_quiesce_invalid_cron` does disable a bad-cron row on its own tick, and the no-burst baseline for NULL rows is real.
- The mirror's best-effort/never-break-Beat contract is preserved and consistently documented.

muhammad-ali-e:
Thanks — thorough pass. All addressed in `f52466ed` (18 unit tests + the real-DB dev-test still green).

### Important
- **Structural regression — read failure could take down the mandatory mirror write** (`tasks.py:63`): extracted the existing-row read + recompute into a fully-guarded `_retargeted_next_run_at()` that returns `None` on any failure, so a read/parse error now degrades to **exactly the pre-PR behaviour** (plain `cron_string`/`enabled` upsert) instead of the outer swallow eating the base write. New test `test_read_failure_degrades_to_plain_upsert` (`filter.side_effect = RuntimeError` → no raise **and** `update_or_create` still called).
- **Swallowed exception** (`tasks.py:81`): now logs the **actual** exception (`exc` + `exc_info=True`), so a bad read, `cron=None` (`AttributeError`), `CroniterBadCronError`/`BadDateError`, or an API change are distinguishable — no more generic dead-end message.

### Suggestions / nits
- **Ownership comment** (`:62`): corrected — Beat→PG hand-over has `next_run_at` already NULL (never set while Beat-owned); ownership only clears it on **rollback to Beat**.
- **Serializer reference** (`:78`): named it — `PipelineSerializer.validate_cron_string` (`pipeline_v2`), not the nearest `scheduler/serializer.py`.
- **Type** (`:47`): `defaults: dict[str, Any]`.
- **Extract helper** (`:74`): done — that's `_retargeted_next_run_at`, which also flattens the nesting.
- **Self-referential assertion** (test `:122`): now asserts the literal `datetime(2026, 1, 1, 10, 0, tzinfo=utc)` (pins value **and** tz-awareness) plus `cron_string == "0 10 * * *"`.
- **Deep-chain dedup** (test `:92`): added a module-level `_stub_existing_row(sched, row=None)` — the `.filter().values().first()` chain now lives in one place; the 3 `TestUpsertMirror` stubs and `_mock_existing` both route through it.

### Acknowledged (not changed)
- **Read-then-write race** (`:87`): agreed it's acceptable for a best-effort mirror — the worst case (mirror reads NULL, tick baselines to old cron) self-heals next cycle. Left as-is; `transaction.atomic()` + `select_for_update()` noted as the close-it path if we ever want it.
- **Tests don't run in CI** (test `:80`): correct and a real pre-existing gap — `scheduler/tests` isn't in any `groups.yaml` lane and there's no `pytest-django`/`django.setup()` shim, so this net is inert in CI (I verified locally via a hand-injected bootstrap: 18/18 here, and the real-DB dev-test). Same theme as the workers CI-gating gap I filed earlier. I'll file a focused follow-up to add `scheduler/tests` to a backend lane with a Django setup shim — flagging rather than silently expanding this PR's scope.

backend/scheduler/tasks.py
```diff
@@ -32,6 +34,49 @@
...
+        # generic "could not recompute" is a dead end when debugging a stale fire.
+        logger.warning(
+            "pg_periodic_schedule: could not retarget next_run_at for pipeline %s "
+            "(cron %r): %s",
+            pipeline_id,
+            cron_string,
+            exc,
+            exc_info=True,
+        )
+        return None
```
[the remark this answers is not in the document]
muhammad-ali-e: Fair observation, but I'm going to keep this as-is — the read is load-bearing, and the path is cold:

- **The SELECT is the feature, not incidental.** Detecting whether the cron *actually changed* is the whole point of the fix, and the caller (`_schedule_task_job`) doesn't hold the previous cron — so there's no cheaper in-memory signal to short-circuit on. Passing the old cron down would just move the same read upstream.
- **Cold path.** `mirror_periodic_schedule_upsert` runs on user-driven schedule saves (create / edit / pause), not a hot loop — and it's a single PK-indexed, two-column `.values()` read.
- **The clean alternative costs more than it saves here.** The function already sits in front of `update_or_create`, which does its own `get`. Collapsing to a single read means replacing the atomic, race-safe `update_or_create` with a manual `filter → create/update` (plus its concurrent-create race, swallowed by the best-effort guard) and rewriting the mirror tests that assert on `update_or_create`. Trading concurrency-safety + churn to shave one cold-path round-trip isn't worth it.

If schedule saves ever become hot, the clean version is exactly what you suggested — one `.values().first()` → `filter().update()` / `create()` (one read, one write). Happy to do it then; not warranted now.

PR #2150, opened 2026-07-07
UN-3693 [FIX] PG — Prompt Studio task_status resolves from pg_task_result (consumer-writes)

muhammad-ali-e:
### Problem
Under `pg_queue_enabled`, Prompt Studio's `dispatch_with_callback` tasks never register a Celery task, so the REST `task_status` endpoint (`AsyncResult` → `celery_taskmeta`) reported **`"processing"` forever** even though the work succeeded — hanging every REST-polling client. The flag-ON regression suite died at `createEtlProject` setup → **0 tests** (flag-OFF ran 208/17/23). Browser UI unaffected (Socket.IO event).

### Fix — PG-native, gated by the same per-org `pg_queue_enabled` flag the dispatch uses
- **Write (one site):** the PG consumer records the terminal status in **`pg_task_result`** from `_chain_continuation` — the single choke point for both success (`error=None`, executor result) and failure (`error` set) of a callback dispatch — keyed by the dispatch `task_id`. `completed` unless the run raised or the executor reported `success=False`; written **before** the enqueue (resolves even if the callback is lost); TTL 24h; best-effort (never wedges the ack). PG-only by construction.
- **Read (`task_status`):** flag **ON** → read `pg_task_result` (completed/failed/none→processing; DB-read guarded → a blip degrades to "processing", not a bare 500); flag **OFF** (default) → Celery `AsyncResult` **unchanged**, PG never queried.

**Clean table separation:** PG tasks live entirely in `pg_task_result`; Celery tasks entirely in `celery_taskmeta`; neither bleeds into the other. **No change to the Celery flow** (flag-off is the original `AsyncResult` code; `ide_callback` untouched).

_Note:_ completed rows are status-only (PII-free); failed rows carry the executor error text (already shown to the owner via the socket event + `task_status` response), TTL'd.

### Testing
Consumer: completed / executor-failed / run-raised / no-id / swallow / records-even-if-enqueue-fails. Backend flag-gate: on completed/failed/pending/DB-blip; off Celery + PG-untouched assertion. DB-gated `store_result(result={}) → completed` guard.

### Deploy
Touches backend + workers → next snapshot rebuilds both; then the flag-ON regression suite should clear `createEtlProject`.

### DIFF
--- a/backend/prompt_studio/prompt_studio_core_v2/tests/test_task_status.py
+++ b/backend/prompt_studio/prompt_studio_core_v2/tests/test_task_status.py
@@ -1,0 +1,104 @@
+"""UN-3693: PromptStudioCoreView.task_status — flag-gated.
+
+Gate on the per-org ``pg_queue_enabled`` flag (the same the executor dispatch uses):
+flag ON → read ``pg_task_result`` (completed / failed / none→processing); flag OFF
+(default) → Celery ``AsyncResult`` unchanged, and the PG table is never queried on
+the hot poll path. ``check_feature_flag_status`` already fails closed to ``False`` on
+any Flipt error, so "flag off / Flipt down" is one path. A transient PG DB error
+degrades to "processing", not a bare 500.
+
+NOTE: run via a Django-bootstrapped harness (no pytest-django in this repo yet — the
+CI-gating of these view tests is tracked in UN-3692). Verified locally green.
+"""
+
+from unittest.mock import MagicMock, patch
+
+from prompt_studio.prompt_studio_core_v2.views import PromptStudioCoreView
+
+_VIEWS = "prompt_studio.prompt_studio_core_v2.views"
+_TID = "task-abc"
+
+
+def _view():
+    view = PromptStudioCoreView()
+    view.get_object = MagicMock()  # bypass permission/object lookup
+    return view
+
+
+def _flag(enabled):
+    return patch(f"{_VIEWS}.check_feature_flag_status", return_value=enabled)
+
+
+def _org():
+    return patch(f"{_VIEWS}.UserSessionUtils.get_organization_id", return_value="org1")
+
+
+def _stub_pg(pg_row=None, *, error=None):
+    """Patch pg_task_result read: .filter().values().first() → pg_row, or raise."""
+    m = MagicMock()
+    if error is not None:
+        m.objects.filter.side_effect = error
+    else:
+        m.objects.filter.return_value.values.return_value.first.return_value = pg_row
+    return patch("pg_queue.models.PgTaskResult", m), m
+
+
+def _async(ready, successful=True, result=""):
+    ar = MagicMock()
+    ar.return_value.ready.return_value = ready
+    ar.return_value.successful.return_value = successful
+    ar.return_value.result = result
+    return ar
+
+
+class TestTaskStatusFlagOn:
+    def test_pg_completed(self):
+        p_pg, _ = _stub_pg({"status": "completed", "error": ""})
+        with _org(), _flag(True), p_pg:
+            resp = _view().task_status(MagicMock(), task_id=_TID)
+        assert resp.data == {"task_id": _TID, "status": "completed"}
+
+    def test_pg_failed(self):
+        p_pg, _ = _stub_pg({"status": "failed", "error": "extraction blew up"})
+        with _org(), _flag(True), p_pg:
+            resp = _view().task_status(MagicMock(), task_id=_TID)
+        assert resp.data["status"] == "failed"
+        assert resp.data["error"] == "extraction blew up"
+        assert resp.status_code == 500
+
+    def test_no_pg_row_is_processing(self):
+        p_pg, _ = _stub_pg(None)  # PG task not done yet — no "pending" row
+        with _org(), _flag(True), p_pg:
+            resp = _view().task_status(MagicMock(), task_id=_TID)
+        assert resp.data == {"task_id": _TID, "status": "processing"}
+
+    def test_pg_db_error_degrades_to_processing(self):
+        # A transient PG read error must not 500; degrade to "processing".
+        p_pg, _ = _stub_pg(error=RuntimeError("db blip"))
+        with _org(), _flag(True), p_pg:
+            resp = _view().task_status(MagicMock(), task_id=_TID)
+        assert resp.data == {"task_id": _TID, "status": "processing"}
+
+
+class TestTaskStatusFlagOff:
+    def test_celery_completed_never_queries_pg(self):
+        # flag off (also the Flipt-error fail-closed path): Celery + PG untouched.
+        p_pg, pg = _stub_pg({"status": "completed"})  # would resolve, if queried
+        with _org(), _flag(False), p_pg, patch(f"{_VIEWS}.AsyncResult", _async(ready=True)):
+            resp = _view().task_status(MagicMock(), task_id=_TID)
+        assert resp.data == {"task_id": _TID, "status": "completed"}
+        pg.objects.filter.assert_not_called()  # default path never touches PG
+
+    def test_celery_failed(self):
+        with _org(), _flag(False), patch(
+            f"{_VIEWS}.AsyncResult", _async(ready=True, successful=False, result="celery boom")
+        ):
+            resp = _view().task_status(MagicMock(), task_id=_TID)
+        assert resp.data["status"] == "failed"
+        assert "celery boom" in resp.data["error"]
+        assert resp.status_code == 500
+
+    def test_pending(self):
+        with _org(), _flag(False), patch(f"{_VIEWS}.AsyncResult", _async(ready=False)):
+            resp = _view().task_status(MagicMock(), task_id=_TID)
+        assert resp.data == {"task_id": _TID, "status": "processing"}
--- a/backend/prompt_studio/prompt_studio_core_v2/views.py
+++ b/backend/prompt_studio/prompt_studio_core_v2/views.py
@@ -19,6 +19,7 @@ from file_management.constants import FileInformationKey as FileKey
 from file_management.exceptions import FileNotFound
 from permissions.permission import IsOwner, IsOwnerOrSharedUserOrSharedToOrg
 from permissions.resource_share_views import ResourceShareManagementMixin
+from pg_queue.flags import PG_QUEUE_FLAG_KEY
 from pipeline_v2.models import Pipeline
 from plugins import get_plugin
 from rest_framework import status, viewsets
@@ -81,6 +82,8 @@ from prompt_studio.prompt_studio_registry_v2.serializers import (
 from prompt_studio.prompt_studio_v2.constants import ToolStudioPromptErrors
 from prompt_studio.prompt_studio_v2.models import ToolStudioPrompt
 from prompt_studio.prompt_studio_v2.serializers import ToolStudioPromptSerializer
+from unstract.core.data_models import PgTaskStatus
+from unstract.flags.feature_flag import check_feature_flag_status
 from unstract.sdk1.utils.common import Utils as CommonUtils
 
 from .models import CustomTool
@@ -872,6 +875,51 @@ class PromptStudioCoreView(ResourceShareManagementMixin, viewsets.ModelViewSet):
         """
         # Verify the user has access to this tool (triggers permission check)
         self.get_object()
+
+        # Gate on the same per-org ``pg_queue_enabled`` flag the executor dispatch
+        # uses (``resolve_pg_transport`` — bucketed per org): flag ON → the task rode
+        # PG, which records its terminal state in ``pg_task_result`` (the eager PG
+        # executor never writes a Celery result backend, so ``AsyncResult`` alone
+        # would poll "processing" forever — UN-3693); flag OFF (default) → Celery,
+        # unchanged, and PG is not touched. ``entity_id``/``context`` match
+        # ``resolve_pg_transport`` so the per-org bucket agrees with the dispatch.
+        # No wrapper: ``check_feature_flag_status`` already fails closed to ``False``
+        # (its own try/except + warning) on any Flipt error → reads as "off" → Celery.
+        org_id = str(UserSessionUtils.get_organization_id(request) or "")
+        pg_enabled = check_feature_flag_status(
+            flag_key=PG_QUEUE_FLAG_KEY,
+            entity_id=org_id or "default",
+            context={"organization_id": org_id},
+        )
+        if pg_enabled:
+            from pg_queue.models import PgTaskResult
+
+            try:
+                row = (
+                    PgTaskResult.objects.filter(task_id=task_id)
+                    .values("status", "error")
+                    .first()
+                )
+            except Exception:
+                # A transient DB blip degrades to "processing" (a poll retries) rather
+                # than a bare 500 that a status-code-keyed client would misread as a
+                # terminal task failure.
+                logger.exception(
+                    "task_status: pg_task_result read failed for %s; treating as "
+                    "pending",
+                    task_id,
+                )
+                row = None
+            # No row → not finished yet (there is no "pending" row). Status compared
+            # against PgTaskStatus (the writer's source of truth), not a bare literal.
+            if row is None:
+                return Response({"task_id": task_id, "status": "processing"})
+            if row["status"] == PgTaskStatus.COMPLETED.value:
+                return Response({"task_id": task_id, "status": "completed"})
+            return Response(
+                {"task_id": task_id, "status": "failed", "error": row["error"] or ""},
+                status=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            )
 
         result = AsyncResult(task_id, app=celery_app)
         if not result.ready():
--- a/workers/queue_backend/pg_queue/consumer.py
+++ b/workers/queue_backend/pg_queue/consumer.py
@@ -70,6 +70,15 @@ _DEFAULT_POISON_REPARK_VT_SECONDS = 300
 # last net) after this many extra reads past max_attempts, so a permanently
 # unmarkable pipeline message can't re-park forever.
 _DEFAULT_POISON_REPARK_BUDGET = 5
+# Retention for the pg_task_result rows the REST task_status poll reads (UN-3693).
+# 24h (Celery's default result_expires) — the 1h default would expire a completed
+# task's row while a late poll (a browser resumed from sleep, an API client checking
+# back) is still querying it. Completed rows are status-only (no payload); FAILED rows
+# carry the executor error text, which can embed document content (cf. the _forget_sql
+# note in result_backend.py / UN-3683) — accepted at this horizon because that same
+# text is already shown to the owner via the WebSocket event + the task_status
+# response, and the row is TTL'd.
+_TASK_STATUS_RETENTION_SECONDS = 86400
 # Liveness: a poll loop that hasn't cycled in this many seconds is reported
 # unhealthy. The heartbeat is stamped at the top of each poll_once and frozen
 # during task execution, so this threshold doubles as an UPPER BOUND on a single
@@ -434,6 +443,56 @@ class PgQueueConsumer:
             self._result_backend = PgResultBackend()
         self._result_backend.store_result(reply_key, result=result, error=error)
 
+    def _record_task_status(
+        self, payload: TaskPayload, *, error: str | None, executor_result: object
+    ) -> None:
+        """Record a ``dispatch_with_callback`` task's terminal status in
+        ``pg_task_result`` so the REST ``PromptStudio.task_status`` poll resolves
+        under the PG transport (UN-3693) — the eager PG executor never writes a Celery
+        result backend under the dispatch id, so ``AsyncResult`` alone would poll
+        "processing" forever.
+
+        Keyed by the dispatch ``task_id``; ``completed`` unless the run raised
+        (``error`` set) or the executor reported an application failure
+        (``executor_result["success"]`` false). Status-only + PII-free (no payload
+        stored) + TTL'd. Best-effort — never raises, so it can't wedge the ack.
+        """
+        task_id = payload.get("task_id")
+        if not task_id:
+            return
+        executor_failed = isinstance(executor_result, dict) and not executor_result.get(
+            "success", True
+        )
+        try:
+            with PgResultBackend() as rb:
+                if error is not None or executor_failed:
+                    msg = (
+                        error
+                        or (
+                            executor_result.get("error")
+                            if isinstance(executor_result, dict)
+                            else None
+                        )
+                        or "Task failed"
+                    )
+                    rb.store_result(
+                        task_id,
+                        error=msg,
+                        retention_seconds=_TASK_STATUS_RETENTION_SECONDS,
+                    )
+                else:
+                    rb.store_result(
+                        task_id,
+                        result={},  # status-only → completed (PII-free)
+                        retention_seconds=_TASK_STATUS_RETENTION_SECONDS,
+                    )
+        except Exception:
+            logger.exception(
+                "PG-queue consumer: could not record pg_task_result status for task "
+                "%s — the REST task_status poll may report 'processing'",
+                task_id,
+            )
+
     def _chain_continuation(
         self,
         spec: ContinuationSpec,
@@ -467,6 +526,18 @@ class PgQueueConsumer:
         its user-facing WebSocket event — is lost, logged loud with the run/task/org
         so the stranded session is correlatable).
         """
+        # Record the terminal status for the REST PromptStudio.task_status poll
+        # (UN-3693) before the enqueue, so it resolves under PG even if the callback
+        # enqueue below fails. This self-chain is the single terminal choke point for
+        # both success (``error`` None) and failure (``error`` set) of a callback
+        # dispatch. The recorded status reflects the EXECUTOR outcome, not callback
+        # delivery: on a rare on_success-enqueue-failure ``_fail_dispatch`` re-enters
+        # here with on_error, but ``store_result`` is first-write-wins (ON CONFLICT DO
+        # NOTHING) so the "completed" already written stands — a REST poll then reads
+        # "completed" (the executor did succeed) even though the callback's socket
+        # event / bookkeeping was lost. Accepted (the browser also relies on that
+        # socket event, so this window is REST-poll-only and rare).
+        self._record_task_status(payload, error=error, executor_result=prepend)
         try:
             queue = spec["queue"]
             kwargs = dict(spec.get("kwargs") or {})
--- a/workers/tests/test_pg_queue_consumer.py
+++ b/workers/tests/test_pg_queue_consumer.py
@@ -978,3 +978,80 @@ class TestSelfChain:
 
 if __name__ == "__main__":
     pytest.main([__file__, "-v"])
+
+
+class TestRecordTaskStatus:
+    """UN-3693: a dispatch_with_callback task's terminal status is recorded in
+    pg_task_result so the REST PromptStudio.task_status poll resolves under PG.
+    completed unless the run raised (error) or the executor reported success=False
+    (completed rows are status-only; failed rows carry the executor error text). TTL'd
+    + best-effort — never wedges the ack."""
+
+    _RB = "queue_backend.pg_queue.consumer.PgResultBackend"
+    _RET = 86400
+
+    @staticmethod
+    def _consumer(client=None):
+        return PgQueueConsumer(["q"], client=client or MagicMock())
+
+    @staticmethod
+    def _entered_rb(patched):  # object bound by `with PgResultBackend() as rb:`
+        return patched.return_value.__enter__.return_value
+
+    def test_success_records_completed(self):
+        with patch(self._RB) as RB:
+            self._consumer()._record_task_status(
+                {"task_id": "tid"}, error=None, executor_result={"success": True, "data": {}}
+            )
+        self._entered_rb(RB).store_result.assert_called_once_with(
+            "tid", result={}, retention_seconds=self._RET
+        )
+
+    def test_executor_reported_failure_records_failed(self):
+        with patch(self._RB) as RB:
+            self._consumer()._record_task_status(
+                {"task_id": "tid"}, error=None,
+                executor_result={"success": False, "error": "boom"},
+            )
+        self._entered_rb(RB).store_result.assert_called_once_with(
+            "tid", error="boom", retention_seconds=self._RET
+        )
+
+    def test_run_raised_records_failed(self):
+        with patch(self._RB) as RB:
+            self._consumer()._record_task_status(
+                {"task_id": "tid"}, error="RuntimeError: kaboom", executor_result="tid"
+            )
+        self._entered_rb(RB).store_result.assert_called_once_with(
+            "tid", error="RuntimeError: kaboom", retention_seconds=self._RET
+        )
+
+    def test_no_task_id_is_noop(self):
+        with patch(self._RB) as RB:
+            self._consumer()._record_task_status(
+                {}, error=None, executor_result={"success": True}
+            )
+        RB.assert_not_called()
+
+    def test_backend_error_is_swallowed(self):
+        with patch(self._RB) as RB:
+            RB.side_effect = RuntimeError("pg down")
+            self._consumer()._record_task_status(
+                {"task_id": "tid"}, error=None, executor_result={"success": True}
+            )  # must not raise
+
+    def test_chain_continuation_records_status_even_if_enqueue_fails(self):
+        # Single-site wiring: status is recorded before the enqueue, so a REST poll
+        # resolves even when the callback continuation is lost.
+        client = MagicMock()
+        client.send.side_effect = RuntimeError("enqueue down")
+        with patch(self._RB) as RB:
+            ok = self._consumer(client)._chain_continuation(
+                {"task_name": "cb.done", "queue": "ide_callback", "kwargs": {}},
+                prepend={"success": True},
+                payload={"task_id": "tid", "args": [{}]},
+            )
+        assert ok is False
+        self._entered_rb(RB).store_result.assert_called_once_with(
+            "tid", result={}, retention_seconds=self._RET
+        )
--- a/workers/tests/test_pg_result_backend.py
+++ b/workers/tests/test_pg_result_backend.py
@@ -66,6 +66,17 @@ class TestStoreGet:
         assert row["error"] == "boom"
         assert row["result"] is None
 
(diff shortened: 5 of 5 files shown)


backend/prompt_studio/prompt_studio_core_v2/tests/test_task_status.py
```diff
@@ -0,0 +1,89 @@
...
+Flag ON → read the PG ``pg_task_result`` row (completed / failed / none→processing);
+the eager PG executor never writes a Celery result backend, so ``AsyncResult`` alone
+would poll "processing" forever. Flag OFF (default) → Celery ``AsyncResult`` exactly
+as before, and the PG table is NOT queried on the hot poll path. Flipt errors fail
+closed to Celery.
+"""
+
+from unittest.mock import MagicMock, patch
+
+from prompt_studio.prompt_studio_core_v2.views import PromptStudioCoreView
```
muhammad-ali-e, line 16: **[Critical — pr-test-analyzer]** These 6 tests never execute in this repo's harness. Importing the real view pulls the full Django model chain and collection dies with `AppRegistryNotReady`: there is no `pytest-django` in `backend/pyproject.toml`, and the `unit-backend` group in `tests/groups.yaml:100-127` (a) doesn't list `prompt_studio/prompt_studio_core_v2/tests` in `paths`, (b) is `optional: true`, and (c) points at `backend.settings.test_cases`, which doesn't exist. So the flag-gating, fail-closed branch, and completed/failed mapping in `views.py` can regress with these tests still "green" (never run).

**Fix**: either wire `pytest-django` + a real settings module and add this path to `groups.yaml`, or follow the sibling `test_build_index_payload.py` pattern in this same directory, which documents this exact constraint and stubs `sys.modules` so the tests run in the bare environment.

backend/prompt_studio/prompt_studio_core_v2/views.py
```diff
@@ -873,6 +875,48 @@ def task_status(
...
+        if pg_enabled:
+            from pg_queue.models import PgTaskResult
+
+            pg_row = (
+                PgTaskResult.objects.filter(task_id=task_id)
+                .values("status", "error")
+                .first()
+            )
+            # No row yet → the PG task hasn't finished (there is no "pending" row).
+            if pg_row is None:
```
muhammad-ali-e: **[Important — silent-failure-hunter]** "No row yet → the PG task hasn't finished" is only true for the first hour. `_record_pg_task_status` calls `store_result` without `retention_seconds`, so rows get `expires_at = now() + 3600` (`DEFAULT_RETENTION_SECONDS`) and the reaper hard-deletes them (`reaper.py:246`). Any poll arriving >1h after completion — browser resumed from sleep, API client checking late — sees a completed/failed task report **"processing" forever**: the exact UN-3693 symptom, with nothing logged anywhere. The "absence = not done" invariant was designed for the blocking RPC waiter (bounded timeout ≤ retention); this REST poll endpoint has no timeout.

**Fix**: pass an explicit longer retention for these status-only rows (e.g. `retention_seconds=86400`, matching Celery's default `result_expires`) — they're PII-free by construction — and/or hedge this comment: `# No row → not finished yet, OR the row already expired at retention (reaper) — both degrade to "processing".`

backend/prompt_studio/prompt_studio_core_v2/views.py
```diff
@@ -873,6 +875,48 @@ def task_status(
...
+        # over PG, which records its terminal state in ``pg_task_result`` (the eager
+        # PG executor never writes a Celery result backend, so ``AsyncResult`` would
+        # poll "processing" forever — UN-3693). Flag OFF (default) → Celery: read
+        # ``AsyncResult`` exactly as before, and skip the PG lookup entirely on the
+        # hot poll path. Resolving with the same org bucket the dispatch used keeps
+        # read and dispatch consistent (an org is all-on or all-off, even mid-ramp);
+        # fail closed to Celery on any Flipt error.
+        org_id = UserSessionUtils.get_organization_id(request)
+        try:
+            pg_enabled = check_feature_flag_status(
```
muhammad-ali-e: **[Important — silent-failure-hunter, type-design-analyzer]** The transport isn't bound to the task; it's re-derived from Flipt on every poll, so any dispatch↔poll drift strands in-flight tasks at permanent "processing": (1) ramp OFF→ON mid-flight → Celery-dispatched task polled against PG, no row ever appears; (2) ramp ON→OFF (i.e. a rollback — exactly when you'd flip it) → PG task polled via `AsyncResult` → PENDING forever; (3) rollout % change re-buckets orgs, same outcome; (4) Flipt outage at poll → fail-closed → Celery read for a PG task. The comment's "an org is all-on or all-off, even mid-ramp" covers entity bucketing, not flips while a task is in flight.

Separately, the Flipt **context** differs from dispatch: `resolve_pg_transport` (`executor_rpc.py:155`) sends `{"executor_name", "organization_id"}`; this call sends only `{"organization_id"}`. A segment rule keyed on `executor_name` would make dispatch and poll permanently disagree. Also `org_id` is resolved *outside* the fail-closed try — `None` becomes `entity_id="None"`.

**Fix (cheap, self-healing)**: cross-check the other store only in the non-terminal branch — PG branch: if `pg_row is None`, check `AsyncResult` before answering "processing"; Celery branch: if `not result.ready()`, check `PgTaskResult`. Costs nothing on the terminal path and removes all four scenarios. Longer term, extract a shared `resolve_pg_for_org(org_id)` helper next to `PG_QUEUE_FLAG_KEY` in `backend/pg_queue/flags.py` so dispatch and poll structurally can't drift. Note also `check_feature_flag_status` (`unstract/flags/feature_flag.py:36`) does an extra `list_feature_flags()` RPC + INFO log on every call — pre-existing, but this PR puts it on a ~2s poll loop.

backend/prompt_studio/prompt_studio_core_v2/views.py
```diff
@@ -873,6 +875,48 @@ def task_status(
...
+
+            pg_row = (
+                PgTaskResult.objects.filter(task_id=task_id)
+                .values("status", "error")
+                .first()
+            )
+            # No row yet → the PG task hasn't finished (there is no "pending" row).
+            if pg_row is None:
+                return Response({"task_id": task_id, "status": "processing"})
+            if pg_row["status"] == "completed":
```
muhammad-ali-e: **[Important — type-design-analyzer]** Bare `"completed"` literal. `PgTaskStatus(str, Enum)` in `unstract/core/src/unstract/core/data_models.py:369` exists precisely so the writer (`PgResultBackend` → `PgTaskStatus.COMPLETED.value`) and readers "agree on one source of truth rather than matching bare literals by eye across the process boundary" (its own docstring); the sibling reader `executor_rpc.py` uses the enum, and this backend already imports from `unstract.core.data_models` elsewhere (e.g. `workflow_v2/transport.py:38`).

**Fix**: `from unstract.core.data_models import PgTaskStatus` and compare `pg_row["status"] == PgTaskStatus.COMPLETED.value`. Consider logging on an unrecognised status instead of the blind else → `failed`/500 (a future status value would silently surface as a task failure with an empty error).

backend/prompt_studio/prompt_studio_core_v2/views.py
```diff
@@ -873,6 +875,48 @@ def task_status(
...
+                org_id,
+                exc_info=True,
+            )
+            pg_enabled = False
+
+        if pg_enabled:
+            from pg_queue.models import PgTaskResult
+
+            pg_row = (
+                PgTaskResult.objects.filter(task_id=task_id)
```
muhammad-ali-e: **[Suggestion — silent-failure-hunter]** This query is unguarded: a transient DB blip raises out of the view → DRF generic 500 *without* the `{task_id, status}` body. Since the legitimate failed branch below also returns 500 (*with* `status: "failed"`), a frontend keying off the status code will misread a transient blip as terminal task failure. Consider catching DB errors here and returning `"processing"` with an error-level log — or using a non-500 code for the legitimate failed branch.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -470,10 +470,16 @@ def _chain_continuation(
         try:
             queue = spec["queue"]
             kwargs = dict(spec.get("kwargs") or {})
+            # Stamp the transport marker so a PG-dispatched callback records its
+            # terminal status in pg_task_result (never celery_taskmeta) — the REST
+            # task_status poll then resolves under PG (UN-3693). PG-only by
+            # construction: this self-chain is the PG analogue of a Celery link, so
+            # the marker is absent on the Celery link path.
+            cb = dict(kwargs.get("callback_kwargs") or {})
+            cb["_pg_transport"] = True
```
muhammad-ali-e: **[Important — comment-analyzer, type-design-analyzer]** This re-types the marker as a literal, but a shared constant already exists: `PG_TRANSPORT_CALLBACK_KWARG: Final = "_pg_transport"` at `workers/queue_backend/pg_barrier.py:203`, whose comment documents the invariant this bypasses — "callback/tasks.py IMPORTS this symbol (not a re-typed literal), so the producer and consumer can't drift." The PG barrier stamps the same marker onto its aggregating callbacks (`pg_barrier.py:844`), so there are now two producers and a rename would miss these two new sites. Worse, drift here fails silently in the worst direction: reader stops matching → `_record_pg_task_status` no-ops → REST poll reports "processing" forever — the exact bug this PR fixes — with the swallow-all posture guaranteeing no error surfaces.

**Fix**: import `PG_TRANSPORT_CALLBACK_KWARG` here and in `ide_callback/tasks.py:83` (or move the constant somewhere both can import cleanly), and mention the barrier as the second producer in the comment above.

workers/ide_callback/tasks.py
```diff
@@ -63,20 +63,63 @@ def _emit_websocket(
...
+    (UN-3693). We record it in the **PG** table (never ``celery_taskmeta``), keyed
+    by ``executor_task_id``. **Gated on the ``_pg_transport`` marker** the PG
+    consumer stamps (``_chain_continuation``), so only PG-dispatched tasks touch PG
+    tables — Celery tasks are recorded by Celery's own result backend, as before.
+    **Status-only** (``result={}``): no extracted payload is stored (PII-safe,
+    consistent with UN-3683). Best-effort: a write failure can never fail the
+    callback (the socket event above is the primary channel).
+    """
+    cb = callback_kwargs or {}
+    if not cb.get("_pg_transport") or not task_id:
```
muhammad-ali-e: **[Important — comment-analyzer, type-design-analyzer]** Reader side of the re-typed literal — import `PG_TRANSPORT_CALLBACK_KWARG` (`workers/queue_backend/pg_barrier.py:203`) instead of `"_pg_transport"`; see the companion comment on `consumer.py:479`. The existing tests pin the literal on each side independently, so a rename of one production file plus its own test file would pass CI while breaking the pair.

Minor: the docstring's "the socket event **above** is the primary channel" dangles — nothing above in this function emits a socket event; the emit happens in the caller (`_emit_event`, which calls this helper after `_emit_websocket`). Suggest "(the socket event emitted by the caller, `_emit_event`, is the primary channel)".

workers/ide_callback/tasks.py
```diff
@@ -63,20 +63,63 @@ def _emit_websocket(
...
+        return
+    status = event_fields.get("status")
+    if status not in ("completed", "failed"):
+        return
+    try:
+        from queue_backend.pg_queue.result_backend import PgResultBackend
+
+        with PgResultBackend() as rb:
+            if status == "completed":
+                rb.store_result(task_id, result={})  # status-only → completed
```
muhammad-ali-e: **[Important — silent-failure-hunter, pr-test-analyzer]** Two things at this write:

1. **Retention**: no `retention_seconds` → default 1h TTL, after which the reaper deletes the row and the REST poll regresses to permanent "processing" (see companion comment on `views.py`). These rows are status-only/PII-free — pass a longer retention (e.g. `86400`, Celery's default `result_expires`).
2. **Untested truthiness hazard**: `result={}` flips `store_result` between completed and failed via its internal `result is not None` check. That exact payload is only mock-asserted — the DB-gated suite (`test_pg_result_backend.py`) never round-trips `result={}`. If `store_result` ever changed to `if result:`, **every successful PG prompt would be recorded as `failed`** with zero test failures. Add one DB-gated case: `store_result(k, result={})` → `get_result(k)["status"] == "completed"`.

workers/ide_callback/tasks.py
```diff
@@ -63,20 +63,63 @@ def _emit_websocket(
...
+            else:
+                rb.store_result(task_id, error=event_fields.get("error") or "Task failed")
+    except Exception:
+        logger.warning(
```
muhammad-ali-e: **[Important — silent-failure-hunter]** The swallow is right (raising would re-run the callback and duplicate the already-sent websocket emit), but WARNING is too quiet for the only write the REST read path depends on. What lands here — revoked INSERT grant, schema drift, missing table, pool exhaustion, even an import error of `result_backend` itself (the import is inside the try) — is systematic, and its failure mode is exactly the bug this PR fixes. WARNINGs on workers don't page, and there's no metric. The log also omits the terminal `status` being dropped and any correlation (`cb` is in scope with `room`/document ids).

**Fix**: `logger.error`, include `status` and correlation from `cb`; cf. the codebase's own `forget()` precedent (`result_backend.py:303-311`) which explicitly flags WARNING-only observability as a stopgap, and `_chain_continuation`'s loud loss-logging (`consumer.py:494-506`) as the model.

workers/ide_callback/tasks.py
```diff
@@ -92,6 +135,9 @@ def _emit_event(
         event=PROMPT_STUDIO_RESULT_EVENT,
         data=_json_safe(payload),
     )
+    # Mirror the terminal outcome into pg_task_result so the REST task_status poll
```
muhammad-ali-e: **[Nit — comment-analyzer]** Third statement of the same rationale within ~35 lines — the `_emit_event` docstring and the `_record_pg_task_status` docstring both already say this, and the callee's name is self-describing. Suggest dropping this inline comment.

workers/tests/test_pg_queue_consumer.py
```diff
@@ -843,7 +843,11 @@ def test_success_chains_on_success_with_result_and_acks(self):
         assert queue_arg == "ide_callback"
         assert payload_arg["task_name"] == "cb.done"
         assert payload_arg["args"] == [7]  # x + y prepended (Celery link parity)
-        assert payload_arg["kwargs"] == {"callback_kwargs": {"room": "r1"}}
+        # ``_pg_transport`` is stamped so the callback records its terminal status
+        # in pg_task_result under the PG transport (UN-3693).
+        assert payload_arg["kwargs"] == {
```
muhammad-ali-e: **[Important — pr-test-analyzer]** Only this success-path test asserts the `_pg_transport` stamp. `test_failure_chains_on_error_with_task_id_and_acks` (below) checks `callback_kwargs["error"]` but not the marker — and the pre-PR code built `cb` only inside the `if error is not None:` branch, so a refactor scoping the stamp back there is a plausible mistake that would silently break exactly the failure half of this fix (failed PG tasks never write a `failed` row → REST pollers spin forever).

**Fix**: add `assert payload_arg["kwargs"]["callback_kwargs"]["_pg_transport"] is True` to the failure-path test (and ideally to `test_early_drop_branches_still_chain_on_error`).

workers/tests/test_ide_callback.py
```diff
@@ -779,3 +779,85 @@ def test_not_called_on_executor_failure(
...
+    def test_marker_failed_stores_error(self):
+        from ide_callback.tasks import _record_pg_task_status
+
+        with patch(self._PATCH_RB) as RB:
+            _record_pg_task_status(
+                {"_pg_transport": True}, "tid-1", {"status": "failed", "error": "boom"}
+            )
+        self._rb(RB).store_result.assert_called_once_with("tid-1", error="boom")
+
+    def test_no_marker_writes_nothing_to_pg(self):
```
muhammad-ali-e: **[Suggestion — code-simplifier, pr-test-analyzer]** Three cleanups for this class:

1. The four "writes nothing" tests (`no_marker`, `none_kwargs`, `non_terminal`, `empty_task_id`) share an identical body — a single `@pytest.mark.parametrize` with ids keeps each case's intent while dropping the duplication.
2. `from ide_callback.tasks import _record_pg_task_status` is repeated in 7 methods — hoist to module level.
3. Gap: these unit tests plus the mocked `_emit_event` pin the helper, but no test exercises a *task-level* path (e.g. `ide_prompt_complete` with `{"_pg_transport": True}` in `cb`, asserting `store_result(executor_task_id, result={})`). The `callback_kwargs=cb` threading at 9 call sites defaults to `None` and no-ops silently, so omitting it at a future `_emit_event` call site reproduces the "processing forever" symptom with all current tests green. One wiring test per task family closes that.

muhammad-ali-e (commented):
## Multi-agent PR review (code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier)

**Core mechanism verified correct**: status-string contract worker↔backend aligns; poll key == record key on both transports; `reply_key` XOR `on_success` means the full result payload never lands in `pg_task_result` (only the status-only `result={}`); fail-closed behaviour is consistent; all terminal `_emit_event` sites are threaded with `callback_kwargs=cb`; the marker is absent on the Celery link path by construction.

### Prioritised findings (details inline)

**Critical**
1. New backend tests never execute — no `pytest-django` in `backend/pyproject.toml`, and `tests/groups.yaml` `unit-backend` excludes `prompt_studio/prompt_studio_core_v2/tests` (group is also `optional: true` and points at a nonexistent `backend.settings.test_cases`). Half the PR's claimed coverage is dead in CI.

**Important**
2. 1h TTL: the reaper deletes `pg_task_result` rows after `DEFAULT_RETENTION_SECONDS=3600`; the view maps row-absence to `processing` → a poll >1h after completion reports "processing" forever (the exact UN-3693 symptom, silently).
3. Flag re-resolved at poll time: any dispatch↔poll drift (ramp flip mid-flight, rollout % change, Flipt outage at poll) strands in-flight tasks at permanent "processing". A cheap cross-check of the other store in the non-terminal branch removes all four scenarios.
4. `_pg_transport` re-typed as a literal at both new sites although `PG_TRANSPORT_CALLBACK_KWARG` (`workers/queue_backend/pg_barrier.py:203`) exists with a documented "IMPORTS this symbol… so the producer and consumer can't drift" invariant.
5. `views.py` compares bare `"completed"` although `PgTaskStatus` (`unstract/core/src/unstract/core/data_models.py:369`) exists precisely so writer/reader don't match literals by eye across trees.
6. Consumer failure-path test doesn't assert the marker — a refactor scoping the stamp back into the `if error` branch would break exactly the failed half of this fix, silently.
7. Flipt context mismatch: dispatch sends `{executor_name, organization_id}`, poll sends only `{organization_id}` — a segment rule on `executor_name` would make dispatch and poll disagree (PG dispatch, Celery poll → eternal "processing").
8. Best-effort PG write failure is WARNING-only with no status/correlation — a systematic failure (revoked grant, schema drift, import error) silently reintroduces UN-3693.

**Suggestions** — retention arg + `result={}` DB round-trip test, org-id resolution outside the fail-closed try, unguarded PG query → shape-less 500 on DB blips, hot-poll-path Flipt helper overhead (`unstract/flags/feature_flag.py:36` does an extra `list_feature_flags()` RPC + INFO log per call — pre-existing, but this PR puts it on a 2-second poll loop), task-level wiring tests, test parametrisation/import hoisting, a dangling docstring referent ("the socket event above" — the emit happens in the caller), and a stale docstring: `backend/pg_queue/models.py:319` still says the retention sweep is "not wired yet", but `reaper.py:246/1091` implements it.

### Strengths
Write-side encapsulation is genuinely good (`_record_pg_task_status` delegates status derivation to `store_result`, which uses the enum); swallow ordering in `_emit_event` is correct (websocket first, PG mirror second, independently guarded); `test_flag_off_uses_celery_and_never_queries_pg` pins the hot-path property behaviourally; status-only `result={}` write is PII-safe and idempotent.

muhammad-ali-e:
Thanks — excellent, deep pass. All addressed in `10cd23525` (worker suites 135, backend self-heal matrix 9, all green via the Django shim).

### Important
- **Dispatch↔poll flag drift** (`views.py:889`): the pure gate could strand in-flight tasks on a mid-flight flip / **rollback (ON→OFF)** / re-bucket / Flipt outage. Fixed with a **self-heal cross-check** — extracted `_pg_task_status_response` / `_celery_task_status_response`; read the flag-selected store first and cross-check the OTHER **only when the primary is non-terminal**. Removes all four scenarios at **zero cost on the terminal path** (a completed task on flag-OFF still never queries PG — asserted). `org_id` is now str-coerced; the fail-closed context caveat is moot now that a Celery segment-rule mismatch self-heals.
- **Retention** (`views.py:911` / `tasks.py:93`): status-only rows are written with **24h retention** (`_TASK_STATUS_RETENTION_SECONDS`), not the 1h default — a late poll after expiry no longer regresses to "processing forever". PII-free by construction, so the longer horizon is safe.
- **Bare literal → enum** (`views.py:913`): compares `PgTaskStatus.COMPLETED/FAILED.value`; unrecognised status now logs and falls through (not a blind → failed/500).
- **Shared marker** (`consumer.py:479`, `tasks.py:83`): both now `import PG_TRANSPORT_CALLBACK_KWARG` (`pg_barrier`) — thanks for catching the **second producer** (the barrier); comment notes it. No more re-typed literal that could silently drift → no-op → "processing forever".
- **Swallow log** (`tasks.py:97`): `logger.error` (not warning), with the dropped `status` and `run_id` correlation.
- **`result={}` truthiness** (`tasks.py:93`): added a **DB-gated** `store_result(result={}) → get_result → completed` guard in `test_pg_result_backend.py` — so an `if result:` regression can't silently flip every successful PG prompt to `failed`.
- **Consumer failure-path marker** (`test_pg_queue_consumer.py:848`): added `assert ...["_pg_transport"] is True` on the failure path.

### Suggestion / Nit
- **Unguarded PG read** (`views.py:906`): guarded — a transient DB error degrades to `"processing"`, not a bare 500 a status-code-keyed client would misread as failure.
- **Docstring dangle** (`tasks.py:83`): reworded to "the socket event emitted by the caller, `_emit_event`". Redundant inline comment at the hook (`:138`) dropped.
- **Test cleanups** (`test_ide_callback.py:812`): parametrized the four "writes nothing" cases, hoisted the import, and added a **task-level** `ide_prompt_complete` wiring test that asserts `store_result(executor_task_id, result={}, retention_seconds=…)` — guarding the `callback_kwargs=cb` threading at the call sites (a future site omitting it would otherwise no-op silently with all helper tests green).

### Critical — test harness (`test_task_status.py:13`)
Fair, and I verified they were green only via a Django-bootstrap shim. I updated them to the full self-heal matrix, but making these **view** tests run bare needs either pytest-django wiring or a large (~30-module) `sys.modules` stub surface for `views.py` — disproportionate to bolt onto this PR. This is the same backend CI-gating gap I filed as **UN-3692**; I've extended it to explicitly cover the `prompt_studio_core_v2` view tests. Flagging rather than shipping a brittle stub. Happy to prioritize UN-3692 if you'd rather gate these before merge.

backend/prompt_studio/prompt_studio_core_v2/views.py
```diff
@@ -120,6 +123,60 @@ def _multi_var_lookup_block_response(custom_tool, prompt_ids=None):
...
+        task_id,
+    )
+    return None
+
+
+def _celery_task_status_response(task_id: str) -> Response | None:
+    """Terminal-status ``Response`` from the Celery result backend, or ``None`` while
+    the task is still pending (an unknown id resolves to PENDING → not ready).
+    """
+    result = AsyncResult(task_id, app=celery_app)
```
muhammad-ali-e: **[Important — silent-failure-hunter]** `_celery_task_status_response` is unguarded, and on the flag-ON path the cross-check runs it on **every in-flight poll** (no `pg_task_result` row yet → primary returns `None` → secondary fires). `result.ready()` hits the Celery result backend — a separate SQLAlchemy Postgres connection (`db+postgresql://…CELERY_BACKEND_DB_NAME`, `backend/backend/celery_config.py`), a different failure domain from the Django-ORM read this PR guarded. A blip there raises out of the view → bare DRF 500 → a status-code-keyed client shows the run as *failed* while it is still running and will complete — the exact misread the `_pg_task_status_response` guard was written to prevent, reintroduced through the other store. It also makes the Celery backend a hard availability dependency of PG-org polls, on the transport whose point is Celery-result-backend independence.

Suggested fix — mirror the PG guard:
```python
def _celery_task_status_response(task_id: str) -> Response | None:
    try:
        result = AsyncResult(task_id, app=celery_app)
        if not result.ready():
            return None
        ...
    except Exception:
        logger.error(
            "task_status: celery result backend read failed for %s; treating as pending",
            task_id, exc_info=True,
        )
        return None
```

backend/prompt_studio/prompt_studio_core_v2/views.py
```diff
@@ -873,14 +930,38 @@ def task_status(
...
+        # PG, which records its terminal state in ``pg_task_result`` (the eager PG
+        # executor never writes a Celery result backend, so ``AsyncResult`` alone
+        # polls "processing" forever — UN-3693); flag OFF (default) → Celery. The
+        # ``entity_id``/``context`` match ``resolve_pg_transport`` so the per-org
+        # bucket agrees with the dispatch. No wrapper: ``check_feature_flag_status``
+        # already fails closed to ``False`` (its own try/except + warning) on any
+        # Flipt error → reads as "off" → Celery, and the cross-check below still
+        # self-heals a PG task.
+        org_id = str(UserSessionUtils.get_organization_id(request) or "")
+        pg_enabled = check_feature_flag_status(
```
muhammad-ali-e, line 889: **[Important — code-reviewer]** Perf: this puts a Flipt evaluation on every poll tick of a hot endpoint. `check_feature_flag_status` constructs a fresh `FliptClient` and makes **two** round-trips per call — `client.list_feature_flags()` purely to emit an INFO log (`unstract/flags/src/unstract/flags/feature_flag.py:36`) plus the actual `evaluate_boolean`. Dispatch pays this once per task; the read path now pays it per poll (~every couple of seconds per active Prompt Studio session), so a slow/degraded Flipt adds latency to every status poll for every org (it fails closed, so no correctness loss — just latency and sustained Flipt load).

Note the flag gate can't simply be dropped: on a non-migrated/Celery-only deployment an unconditional PG-first read would ERROR-log on every poll (missing table). Suggested: a short-TTL per-org cache of the flag result in this view, and/or fix the wasteful `list_feature_flags()` log call in the shared helper (pre-existing, but this PR is what makes it hot). Fine as a fast-follow.

workers/ide_callback/tasks.py
```diff
@@ -16,12 +16,20 @@
...
 
 logger = logging.getLogger(__name__)
 
 PROMPT_STUDIO_RESULT_EVENT = "prompt_studio_result"
 
+# Retention for the status-only pg_task_result rows the REST task_status poll reads.
+# The default (1h) would expire a completed task's row while a late poll (a browser
+# resumed from sleep, an API client checking back) is still querying it → the exact
+# "processing forever" symptom UN-3693 fixes. 24h (Celery's default result_expires);
+# the rows are PII-free by construction, so the longer horizon is safe.
```
muhammad-ali-e: **[Important — comment-analyzer]** "PII-free by construction" only holds for the *completed* path. The *failed* path stores `error=event_fields.get("error")` (line 108) — free-form executor error text, which this codebase itself documents can embed document content: the `_forget_sql` comment in `workers/queue_backend/pg_queue/result_backend.py` ("an extraction error can embed document content too") is the very reason UN-3683 scrubs `error` on consume. These ide-callback rows are never scrubbed (the REST poll reads via ORM; no forget-on-consume), and this constant extends their life from 1h to 24h.

Either correct the claim — e.g. "completed rows are status-only; failed rows carry the executor error text (which can embed document content), accepted at 24h because the same text already rides the WebSocket event and the task_status response" — or truncate/sanitize the stored error text before writing. Same overstatement in the `_record_pg_task_status` docstring (line 86) and the `TestRecordPgTaskStatus` docstring.

workers/tests/test_ide_callback.py
```diff
@@ -779,3 +781,111 @@ def test_not_called_on_executor_failure(
...
+                "index_document",
+                callback_kwargs={PG_TRANSPORT_CALLBACK_KWARG: True},
+                status="completed",
+            )
+        rec.assert_called_once()
+        assert rec.call_args.args[0] == {PG_TRANSPORT_CALLBACK_KWARG: True}
+        assert rec.call_args.args[1] == "tid-9"
+        assert rec.call_args.args[2].get("status") == "completed"
+
+    def test_task_level_ide_prompt_complete_records_pg_status(
```
muhammad-ali-e: **[Important — pr-test-analyzer]** This end-to-end guard covers only the success path through `ide_prompt_complete`. None of the error callbacks (`ide_prompt_error` — tasks.py:603, `ide_index_error` — :424, the failure branches of `ide_index_complete` — :262/:361/:376) are exercised with the marker: their pre-existing tests run the real `_emit_event` but never set `PG_TRANSPORT_CALLBACK_KWARG`, so `_record_pg_task_status` no-ops unasserted. A refactor dropping `callback_kwargs=cb` at one of those sites leaves every test green while failed PG runs never write a `failed` row — REST pollers spin at "processing" forever, on exactly the path where stopping the spinner matters most (the consumer-side test at `test_pg_queue_consumer.py:869` guards its half of this chain; the callback half is unguarded).

Suggested: a sibling of this test calling `ide_prompt_error("tid", cb)` with `cb = {**base_prompt_kwargs, PG_TRANSPORT_CALLBACK_KWARG: True, "executor_task_id": "exec-9", "error": "boom"}`, asserting `store_result("exec-9", error=<contains "boom">, retention_seconds=86400)` — optionally parametrized over `ide_index_error` and the `ide_index_complete` failure branch.

muhammad-ali-e (commented):
Second-pass review (Code Reviewer, Silent Failure Hunter, Type Design Analyzer, PR Test Analyzer, Comment Analyzer, Code Simplifier). All prior-round feedback verified as addressed — marker constant shared, retention set, enum comparison, guarded PG read, cross-store self-heal all check out end-to-end (poll key = write key, `result={}` → COMPLETED, no key collision with the blocking RPC path, no connection leak). 4 new findings below: 1 error-handling gap, 1 data-retention/comment mismatch, 1 hot-path perf cost, 1 test gap on the failure path.

muhammad-ali-e:
Heads-up: I **pivoted the approach to consumer-writes** (force-pushed `803352059`) — it's much simpler and, usefully, it erases two of your latest comments by construction. Mapping the round:

- **`test_ide_callback.py:867` (callback wiring only covers success)** → **gone.** `ide_callback/tasks.py` is reverted to base; the write is now a single consumer method (`_record_task_status` from `_chain_continuation`, the one choke point for both success and failure), tested for completed / executor-failed / run-raised / no-id / swallow / records-even-if-enqueue-fails. No 8-site threading to leave unguarded.
- **`views.py:169` (self-heal runs the unguarded Celery read on every PG poll → a Celery-backend blip 500s a running task)** → **gone.** The self-heal cross-check is dropped; on flag-ON the read never touches the Celery backend. `task_status` is a pure flag-gate now (flag off = the original `AsyncResult` code, byte-for-byte).
- **`tasks.py:30` (PII claim — failed rows carry executor error text that can embed doc content)** → **fixed.** Corrected the claim in all three places (the retention-constant comment, the write docstring, the test docstring): completed rows are status-only; failed rows carry the executor error text, accepted at 24h because that same text is already shown to the owner via the socket event + the `task_status` response, and the row is TTL'd.
- **`views.py:944` (Flipt eval per poll — 2 round-trips)** → still applies to the flag-gate; taking it as the **fast-follow** you suggested (short-TTL per-org flag cache + fixing the wasteful `list_feature_flags()` log in the shared helper). Filing it.

Also fixed the two SonarCloud smells (`logging.error(..., exc_info=True)` → `logging.exception()` in the consumer write + the backend DB-guard).

Net: the change is ~half the size (one write site, `ide_callback` untouched, pure flag-gate read), no Celery-flow change, clean PG/Celery table separation. The one accepted trade-off vs the callback approach: the consumer records the *executor's* outcome, so the rare "extraction succeeds but the callback's own bookkeeping then fails" reads as `completed` — the browser still shows the true status via Socket.IO.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -467,6 +526,12 @@ def _chain_continuation(
         its user-facing WebSocket event — is lost, logged loud with the run/task/org
         so the stranded session is correlatable).
         """
+        # Record the terminal status for the REST PromptStudio.task_status poll
+        # (UN-3693) before the enqueue, so it resolves under PG even if the callback
+        # enqueue below fails. This self-chain is the single terminal choke point for
+        # both success (``error`` None) and failure (``error`` set) of a callback
+        # dispatch.
+        self._record_task_status(payload, error=error, executor_result=prepend)
```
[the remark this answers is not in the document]
muhammad-ali-e: Good catch on the exact semantics — added a clarifying comment at the write site (`8b2e2f066`). It documents that the recorded status reflects the **executor** outcome, not callback delivery: on the rare on_success-enqueue-failure, `_fail_dispatch` re-enters with on_error but the first-write-wins `ON CONFLICT DO NOTHING` keeps the "completed" — so a REST poll reads "completed" (the executor did succeed) even though the callback's socket event/bookkeeping was lost. Accepted as-is: it's a rare, REST-poll-only window (the browser relies on that same socket event), and forcing the on_error re-write to override would trade the clean idempotency for a narrower edge.

backend/prompt_studio/prompt_studio_core_v2/tests/test_task_status.py
```diff
@@ -0,0 +1,104 @@
...
+the hot poll path. ``check_feature_flag_status`` already fails closed to ``False`` on
+any Flipt error, so "flag off / Flipt down" is one path. A transient PG DB error
+degrades to "processing", not a bare 500.
+
+NOTE: run via a Django-bootstrapped harness (no pytest-django in this repo yet — the
+CI-gating of these view tests is tracked in UN-3692). Verified locally green.
+"""
+
+from unittest.mock import MagicMock, patch
+
```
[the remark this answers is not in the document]
muhammad-ali-e: Correct — genuinely offline until UN-3692. There's no conftest/Django harness for these view tests in this repo yet (no pytest-django; `prompt_studio/**/tests` isn't in a `tests/groups.yaml` lane), so `from ...views import PromptStudioCoreView` needs a Django bootstrap. I verified them locally via a `django.setup()` shim (all green), and extended **UN-3692** to cover the `prompt_studio_core_v2` view tests alongside the scheduler ones — wiring pytest-django + a lane there gates both. Flagged rather than bolting a ~30-module `sys.modules` stub onto this PR.

PR #2153, opened 2026-07-08
UN-3695 [FIX] PG consumer — supervisor shutdown grace = VT, single shared join deadline

muhammad-ali-e:
## What
The PG-queue prefork **supervisor** SIGKILLed its own children after a **hardcoded 30s** on shutdown (`_SHUTDOWN_GRACE_SECONDS = 30.0`), even though the pod's `terminationGracePeriodSeconds` is set to `VT + buffer` (9120s for `worker-pg-file-processing`, VT=9060). So on a graceful SIGTERM (deploy / HPA pod scale-down) the supervisor knifed an in-flight batch ~300× sooner than the grace the pod actually had — **orphaning the batch**: the executor finished the work, but nothing was left to consume the reply, mark the file COMPLETE, decrement the barrier, or ack the queue message.

Two changes:
- **`shutdown_grace_from_env()`** — the per-child drain grace now defaults to the consumer **VT** (`WORKER_PG_QUEUE_CONSUMER_VT_SECONDS`), **floored at 30** so it can never be *shorter* than before (non-regressive: OSS/dev on the default VT stays 30; fileproc gets 9060 with **no chart change**, since the chart already sets VT). Explicit `WORKER_PG_QUEUE_CONSUMER_SHUTDOWN_GRACE_SECONDS` overrides.
- **`_join_children` → single shared deadline** (was per-child recomputed). Children are SIGTERM'd together and drain in parallel, so one shared window still gives each its full grace while bounding total drain to ~grace. A per-child deadline sums to N×grace and, at grace=VT, blows past the pod grace → k8s SIGKILLs the whole pod, hard-killing siblings that were draining cleanly.

## Why
Deploys and autoscaling routinely send graceful SIGTERM to workers. The 30s app-grace silently defeated the ~2.5h drain budget the chart deliberately provisions, so every scale-down/rollout landing mid-extraction stranded a batch until visibility-timeout redelivery (~2.5h) or the reaper.

## Scope / safety
- The supervisor (`pg_queue_consumer`) is a **PG-worker-only component** — it never runs for Celery workers, so **no feature-flag branch is needed and the Celery path is byte-for-byte unchanged**.
- `max(30, VT)` floor guarantees the drain is never shorter than today.

## How tested
- Unit: `shutdown_grace_from_env` (VT-tracking / floor / override) + a shared-deadline regression test — **38 supervisor tests pass**; `test_pg_consumer_supervisor.py` + `test_pg_queue_consumer.py` = 109 pass.
- **e2e with real forked children**: (1) grace tracks VT (40) and the supervisor **drains** a live child with no premature SIGKILL; (2) two *wedged* children are SIGKILLed together at the **shared** deadline, not 2× it.

Part of UN-3695 (PG crash-safe worker shutdown). Follow-ups tracked there: heartbeat-lease for hard-kill recovery (next); the idempotency track was found already handled (billing is idempotent in the cloud plugins) and needs no code.

### DIFF
--- a/workers/pg_queue_consumer/supervisor.py
+++ b/workers/pg_queue_consumer/supervisor.py
@@ -36,6 +36,7 @@ from __future__ import annotations
 
 import contextlib
 import logging
+import math
 import multiprocessing
 import os
 import signal
@@ -63,8 +64,9 @@ _RESTART_MAX_BACKOFF_SECONDS = 30.0
 _MIN_HEALTHY_UPTIME_SECONDS = 10.0
 # Consecutive immediate crashes after which the fleet probe is forced unhealthy.
 _CRASH_LOOP_THRESHOLD = 3
-# Fallback per-child graceful-drain budget (s) on shutdown, used only when neither
-# an explicit override nor the consumer VT is set — see shutdown_grace_from_env().
+# Fallback graceful-drain budget (s, shared across all children) on shutdown, used
+# only when neither an explicit override nor the consumer VT is set — see
+# shutdown_grace_from_env().
 # This is NOT the live value: the live grace defaults to the consumer's visibility
 # timeout so a SIGTERM (deploy / HPA scale-down) lets an in-flight batch finish
 # instead of a mid-flight SIGKILL that orphans it (UN-3695).
@@ -98,7 +100,8 @@ def concurrency_from_env() -> int:
 
 
 def shutdown_grace_from_env() -> float:
-    """Per-child graceful-drain budget (seconds) on shutdown, before SIGKILL.
+    """Graceful-drain budget (seconds) on shutdown, shared across all children,
+    before SIGKILL.
 
     Defaults to the consumer's visibility timeout (``WORKER_PG_QUEUE_CONSUMER_VT_SECONDS``)
     — the by-design upper bound on a single task's runtime — so a graceful SIGTERM
@@ -115,11 +118,20 @@ def shutdown_grace_from_env() -> float:
     """
     from queue_backend.pg_queue.consumer import _DEFAULT_VT_SECONDS, consumer_env
 
-    override = consumer_env("SHUTDOWN_GRACE_SECONDS", None, float)
+    override: float | None = consumer_env("SHUTDOWN_GRACE_SECONDS", None, float)
     if override is not None:
-        return max(0.0, override)
+        # A negative / non-finite drain budget is never intentional and would
+        # re-introduce the exact failure this module prevents: <0 or nan → 0 → every
+        # child SIGKILLed with no drain; inf → shutdown hangs until k8s hard-kills the
+        # pod. Fail fast at startup, mirroring concurrency_from_env()'s validation.
+        if not math.isfinite(override) or override < 0:
+            raise ValueError(
+                "WORKER_PG_QUEUE_CONSUMER_SHUTDOWN_GRACE_SECONDS must be a finite "
+                f"number >= 0, got {override!r}"
+            )
+        return override
     vt = consumer_env("VT_SECONDS", _DEFAULT_VT_SECONDS, int)
-    return max(float(_DEFAULT_SHUTDOWN_GRACE_SECONDS), float(vt))
+    return max(_DEFAULT_SHUTDOWN_GRACE_SECONDS, float(vt))
 
 
 class _Fleet:
@@ -363,7 +375,9 @@ def run_supervised(concurrency: int) -> None:
     fleet = _Fleet(concurrency)
     grace_seconds = shutdown_grace_from_env()
     logger.info(
-        "PG-queue consumer supervisor: shutdown drain grace = %.0fs/child", grace_seconds
+        "PG-queue consumer supervisor: shutdown drain grace = %.0fs (shared across "
+        "children)",
+        grace_seconds,
     )
     stopping = threading.Event()
 
@@ -415,16 +429,22 @@ def run_supervised(concurrency: int) -> None:
 def _wait_for_exit(pid: int, deadline: float) -> bool:
     """Poll ``pid`` until it exits or ``deadline`` (monotonic) passes. True if it
     exited (or was already reaped).
+
+    Polls ``waitpid`` at least once regardless of the deadline: with the single
+    SHARED shutdown deadline, a child iterated after the window has already elapsed
+    would otherwise be reported "did not drain" and SIGKILLed — a false hard-kill
+    alarm — even though it exited cleanly within the grace.
     """
-    while time.monotonic() < deadline:
+    while True:
         try:
             reaped, _status = os.waitpid(pid, os.WNOHANG)
         except ChildProcessError:
             return True
         if reaped != 0:
             return True
+        if time.monotonic() >= deadline:
+            return False
         time.sleep(0.1)
-    return False
 
 
 def _join_children(fleet: _Fleet, grace_seconds: float) -> None:
--- a/workers/pg_queue_consumer/supervisor.py
+++ b/workers/pg_queue_consumer/supervisor.py
@@ -63,8 +63,12 @@ _RESTART_MAX_BACKOFF_SECONDS = 30.0
 _MIN_HEALTHY_UPTIME_SECONDS = 10.0
 # Consecutive immediate crashes after which the fleet probe is forced unhealthy.
 _CRASH_LOOP_THRESHOLD = 3
-# How long to wait per child for a graceful drain on shutdown before SIGKILL.
-_SHUTDOWN_GRACE_SECONDS = 30.0
+# Fallback per-child graceful-drain budget (s) on shutdown, used only when neither
+# an explicit override nor the consumer VT is set — see shutdown_grace_from_env().
+# This is NOT the live value: the live grace defaults to the consumer's visibility
+# timeout so a SIGTERM (deploy / HPA scale-down) lets an in-flight batch finish
+# instead of a mid-flight SIGKILL that orphans it (UN-3695).
+_DEFAULT_SHUTDOWN_GRACE_SECONDS = 30.0
 
 
 def concurrency_from_env() -> int:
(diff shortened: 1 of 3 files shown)


workers/pg_queue_consumer/supervisor.py
```diff
@@ -93,6 +97,31 @@ def concurrency_from_env() -> int:
...
+    ``WORKER_PG_QUEUE_CONSUMER_SHUTDOWN_GRACE_SECONDS`` overrides (and is honoured
+    as-is, e.g. a short dev drain); otherwise the VT is floored at
+    ``_DEFAULT_SHUTDOWN_GRACE_SECONDS`` so a tiny/unset VT still gets a sane drain.
+
+    A hardcoded 30s here previously undercut the chart's ~2.5h budget by ~300×,
+    SIGKILLing in-flight batches on every scale-down/rollout.
+    """
+    from queue_backend.pg_queue.consumer import _DEFAULT_VT_SECONDS, consumer_env
+
+    override = consumer_env("SHUTDOWN_GRACE_SECONDS", None, float)
```
muhammad-ali-e: **[med] `SHUTDOWN_GRACE_SECONDS` override is unvalidated — a bad value silently produces the exact failure UN-3695 fixes.**

`consumer_env(..., float)` accepts negative and non-finite values, and `max(0.0, override)` then collapses them:
- `SHUTDOWN_GRACE_SECONDS=-30` → `max(0.0, -30)=0.0` → deadline `now+0` → **every child SIGKILLed with zero drain** on the next scale-down (the orphaned-batch bug this PR exists to prevent), with no log.
- `nan` → `max(0.0, nan)=0.0` → same immediate kill.
- `inf`/`1e999` → `max(0.0, inf)=inf` → `_join_children` deadline is `now+inf` → shutdown **hangs forever** until k8s hard-kills the whole pod.

A negative/non-finite drain budget is never intentional. Sibling `concurrency_from_env()` already *raises* on `n < 1` — mirror that here:
```python
override: float | None = consumer_env("SHUTDOWN_GRACE_SECONDS", None, float)
if override is not None:
    if not math.isfinite(override) or override < 0:
        raise ValueError(
            f"WORKER_PG_QUEUE_CONSUMER_SHUTDOWN_GRACE_SECONDS must be a finite "
            f"number >= 0, got {override!r}"
        )
    return override
```
(The `override:` annotation also restores the `X | None` intent-declaration convention this module documents at `consumer.py:858`.)

workers/pg_queue_consumer/supervisor.py
```diff
@@ -395,15 +428,22 @@ def _wait_for_exit(pid: int, deadline: float) -> bool:
...
+    window still gives every child its full grace from that common SIGTERM — while
+    bounding the total wait to ~``grace_seconds``. A per-child deadline would instead
+    sum to N×grace and, at grace≈VT (thousands of seconds), blow past the pod's
+    ``terminationGracePeriodSeconds`` — so k8s SIGKILLs the whole pod, hard-killing
+    siblings that were still draining cleanly (UN-3695 / debate H2).
     """
+    deadline = time.monotonic() + grace_seconds
     for slot, pid in fleet.alive_items():
-        if _wait_for_exit(pid, time.monotonic() + grace_seconds):
+        if _wait_for_exit(pid, deadline):
```
muhammad-ali-e, line 462: **[med] New regression: a clean-exited child iterated *after* the shared deadline elapses is spuriously SIGKILLed and logged as "did not drain".**

With one wedged child consuming the whole shared window, every subsequent child hits `_wait_for_exit(pid, deadline)` with `deadline` already in the past. `_wait_for_exit`'s loop is `while time.monotonic() < deadline:` — so it returns `False` **without ever calling `waitpid`**, even for a child that exited cleanly at T+5s. `_join_children` then SIGKILLs it and logs `"did not drain within the shared %ss grace"`.

The SIGKILL to the (already-zombie) child is harmless, but the WARNING is a false "batch was hard-killed" alarm — ironic for the PR whose whole point is to stop hard-killing in-flight batches, and it can trip exactly that alert. This is new: the old per-child code recomputed `time.monotonic() + grace` per iteration, so the deadline was never already-elapsed at entry.

Fix in `_wait_for_exit` — poll `waitpid` at least once regardless of the deadline:
```python
def _wait_for_exit(pid: int, deadline: float) -> bool:
    while True:
        try:
            reaped, _status = os.waitpid(pid, os.WNOHANG)
        except ChildProcessError:
            return True
        if reaped != 0:
            return True
        if time.monotonic() >= deadline:
            return False
        time.sleep(0.1)
```

workers/pg_queue_consumer/supervisor.py
```diff
@@ -332,6 +361,10 @@ def run_supervised(concurrency: int) -> None:
     logging.basicConfig(level=os.getenv("LOG_LEVEL", "INFO"))
 
     fleet = _Fleet(concurrency)
+    grace_seconds = shutdown_grace_from_env()
+    logger.info(
+        "PG-queue consumer supervisor: shutdown drain grace = %.0fs/child", grace_seconds
```
muhammad-ali-e: **[low] Startup log says `%.0fs/child` but the drain is now a single *shared total*, not per-child.**

After this PR `grace_seconds` bounds the total drain to ~grace across all children (see the `_join_children` docstring, which explicitly rejects the per-child model). An operator reading `grace = 9060s/child` with concurrency=4 will mis-model the total window as 4×9060s — the precise per-child-vs-shared misconception the rework fixes. Drop the `/child`:
```python
"PG-queue consumer supervisor: shutdown drain grace = %.0fs (shared across children)", grace_seconds
```

workers/pg_queue_consumer/supervisor.py
```diff
@@ -93,6 +97,31 @@ def concurrency_from_env() -> int:
     return n
 
 
+def shutdown_grace_from_env() -> float:
+    """Per-child graceful-drain budget (seconds) on shutdown, before SIGKILL.
```
muhammad-ali-e: **[low] "Per-child" wording left stale after the switch to a shared deadline (3 sites).**

This docstring's first line still calls the value a *"Per-child graceful-drain budget"*, and so does the constant comment at line 66 and the README table row at `README.md:25` — yet `_join_children` now applies it as a **single shared total** and its own docstring explicitly warns that a per-child deadline would be a bug. Same knob described two contradictory ways within one PR. Suggest a find-and-replace to "graceful-drain budget, shared across all children on shutdown" at all three sites (the `_join_children` docstring and the SIGKILL warning at line 445 are already correct — align the rest to them).

workers/pg_queue_consumer/supervisor.py
```diff
@@ -93,6 +97,31 @@ def concurrency_from_env() -> int:
...
+    A hardcoded 30s here previously undercut the chart's ~2.5h budget by ~300×,
+    SIGKILLing in-flight batches on every scale-down/rollout.
+    """
+    from queue_backend.pg_queue.consumer import _DEFAULT_VT_SECONDS, consumer_env
+
+    override = consumer_env("SHUTDOWN_GRACE_SECONDS", None, float)
+    if override is not None:
+        return max(0.0, override)
+    vt = consumer_env("VT_SECONDS", _DEFAULT_VT_SECONDS, int)
+    return max(float(_DEFAULT_SHUTDOWN_GRACE_SECONDS), float(vt))
```
muhammad-ali-e: **[nit] Redundant `float()` on an already-float constant.** `_DEFAULT_SHUTDOWN_GRACE_SECONDS` is `30.0` (line 71), so `float(...)` around it is a no-op. `float(vt)` is the meaningful cast (`vt` is an `int`):
```python
return max(_DEFAULT_SHUTDOWN_GRACE_SECONDS, float(vt))
```

workers/queue_backend/pg_queue/README.md
```diff
@@ -0,0 +1,147 @@
...
+`consumer_env("<SUFFIX>", …)` → `WORKER_PG_QUEUE_CONSUMER_<SUFFIX>`. Authoritative
+defaults live in the code constants (`consumer.py`, `pg_queue_consumer/supervisor.py`,
+`reaper.py`); cloud overrides live in the Helm chart `values.yaml`.
+
+### Consumer (`WORKER_PG_QUEUE_CONSUMER_*`)
+| Env suffix | One-line | Default |
+|---|---|---|
+| `CONCURRENCY` | Prefork consumer children per pod (1 = plain single process) | `1` |
+| `VT_SECONDS` | **Visibility timeout** — how long a claimed message stays hidden before it can redeliver | `30` (chart: `9060` for file-processing) |
+| `SHUTDOWN_GRACE_SECONDS` | Per-child graceful-drain budget on SIGTERM before SIGKILL | `= VT`, floored at `30` |
```
muhammad-ali-e: **[low] This row contradicts the glossary in the same file.** It calls `SHUTDOWN_GRACE_SECONDS` a *"Per-child graceful-drain budget"*, but the "Shutdown grace" glossary entry at line 87 correctly says the supervisor drains *"using a single shared deadline."* Reword to `Graceful-drain budget (shared across all children) on SIGTERM before SIGKILL` so the two agree.

workers/tests/test_pg_consumer_supervisor.py
```diff
@@ -258,6 +297,25 @@ def test_straggler_is_sigkilled_and_reaped(self):
         kill.assert_called_once_with(111, _signal.SIGKILL)
         waitpid.assert_called_once_with(111, 0)
 
+    def test_shared_deadline_not_per_child(self):
```
muhammad-ali-e, line 326: **[test] Coverage gaps around the change's load-bearing paths.**

1. **Multi-child straggler on the shared deadline is untested.** `test_shared_deadline_not_per_child` only exercises the clean-exit path (`_capture` always returns `True`); `test_straggler_is_sigkilled_and_reaped` uses a single child. A regression re-introducing a per-child deadline inside the SIGKILL loop would pass both. Add a 3-child, all-`False` (none drain) case asserting one shared deadline **and** `kill`/`waitpid` called 3×.
2. **This assertion is probabilistic.** `len(set(deadlines)) == 1` only distinguishes shared-vs-per-child if `time.monotonic()` ticks between iterations; with `_wait_for_exit` patched to return instantly a per-child bug could yield 3 identical values on a coarse clock and pass. Harden by patching `time.monotonic` with an advancing `side_effect=[100.0, 200.0, 300.0, 400.0]` and asserting `deadlines == [105.0, 105.0, 105.0]`.
3. **Negative override branch untested** — add `SHUTDOWN_GRACE_SECONDS=-5` (pairs with the validation fix suggested on `supervisor.py:118`).
4. **Malformed env is untested** — assert `shutdown_grace_from_env()` raises `ValueError` matching the offending var name for `SHUTDOWN_GRACE_SECONDS=abc` and `VT_SECONDS=9060s`, pinning the fail-fast-at-startup contract.

muhammad-ali-e (commented):
Automated PR review (PR Review Toolkit: code-reviewer, silent-failure-hunter, type-design, comment, test, simplifier). The core fix is correct — grace defaults to VT floored at 30s, override honoured as-is, and the single shared deadline correctly bounds total drain to ~grace instead of N×grace. Findings below are ranked; the two `[med]` items are the ones worth addressing before merge. Everything else is low/nit.

muhammad-ali-e:
Thanks — all 7 addressed in `67f7d6e83`:

- **[med] `supervisor.py:118` unvalidated override** — `shutdown_grace_from_env` now rejects a negative / non-finite `SHUTDOWN_GRACE_SECONDS` (`math.isfinite` + `>= 0`) and raises with the offending var name, mirroring `concurrency_from_env`. No more `-30`/`nan` → 0 (immediate SIGKILL) or `inf` → hang. Added the `override: float | None` annotation.
- **[med] `supervisor.py:442` `_wait_for_exit` false-alarm** — it now polls `waitpid` **at least once** regardless of the deadline (`while True` + explicit exits), so a child iterated after the shared window elapsed isn't spuriously reported "did not drain" / SIGKILLed.
- **[low] `:366` log** — now `shutdown drain grace = %.0fs (shared across children)` (dropped `/child`).
- **[low] "Per-child" wording (3 sites)** — corrected the `shutdown_grace_from_env` docstring, the constant comment (line 66), and `README.md` to "shared across all children", matching the `_join_children` docstring.
- **[nit] `:122` redundant `float()`** — dropped; now `max(_DEFAULT_SHUTDOWN_GRACE_SECONDS, float(vt))`.
- **[low] `README.md:25`** — row reworded to agree with the glossary ("shared across all children").
- **[test] coverage** — (1) shared-deadline test hardened with an advancing `time.monotonic` (asserts exact `105/105/105`, no longer probabilistic); (2) new 3-child all-wedged straggler case asserting `kill`/`waitpid` ×3 **and** one shared deadline; (3)+(4) negative / non-finite / malformed override + malformed `VT_SECONDS=9060s` fail-fast tests. Replaced the `… or True` lambdas with named helpers. **43 tests pass; e2e re-verified** (drain + wedge).

SonarCloud's "C Reliability on New Code" should clear on this commit — it was the same override/hang class + the lambda idiom, both now fixed.

muhammad-ali-e:
Addressed the Greptile nit (`20462e044` → amended): the SIGKILL warning in `_join_children` now uses `%.0fs` for `grace_seconds`, matching the startup log (was `%s` → printed `9060.0s` vs the startup log's `9060s`). Cosmetic only, no behavior change.

PR #2154, opened 2026-07-08
UN-3695 [FIX] PG consumer — renewable lease (dead worker redelivers in minutes, not the full VT)

muhammad-ali-e:
## What
Follow-on to the merged grace=VT fix (#2153). That covered **graceful** kills (drain on SIGTERM). This covers **hard** kills (OOM / node loss), where the worker gets no SIGTERM: the message was claimed for the full visibility timeout (VT=9060s / 2.5h for file-processing), so a hard-killed worker stranded its in-flight message for ~2.5h before the vt expired and it redelivered.

- **New `LEASE_SECONDS` (default 120):** the message is claimed for the short lease, and a daemon thread **renews** it (`client.set_vt`) every ~`lease/3` while the task runs. A live-but-slow task keeps its claim; a **dead** worker's renewal stops with the process, so its lease expires in ~lease → **redelivery in minutes, not 2.5h**.
- **`VT_SECONDS` is retained as the drain / max-runtime bound** — it still drives PR A's shutdown grace and the chart's `health-stale` / render guards. It is **no longer the claim window**; the claim now uses `LEASE_SECONDS` (capped at VT).
- **Safety:** the renewal thread is started only around `task.apply`, **joined before the ack** so the two never race the row; it uses its **own DB connection** (the main thread owns the claim/ack connection); it's **best-effort** — a transient `set_vt` failure retries within the ~2× slack the `lease/3` interval leaves before expiry, and a `False` return (row already acked/expired) stops it. The renew connection is opened only on the **first actual renewal**, so a task shorter than the interval opens no second connection.

## Why
PR A made deploy/scale-down orphans impossible, but OOM/node-death (no SIGTERM) still stranded a batch for the full VT. Shortening recovery to minutes closes the hard-kill gap and brings PG closer to Celery's broker-requeue latency.

## Scope / safety
- PG-worker-only path (the consumer); Celery untouched.
- Default `LEASE_SECONDS=120` works with no chart change; the paired unstract-cloud change sets it explicitly per pg-worker + adds a `LEASE ≤ VT` render guard.

## How tested
- Unit (deterministic, controlling `stop.wait`): lease cap-at-VT / renew-interval (`lease//3`) / **claim uses lease not VT** / renew-until-stopped / stop-on-row-gone / swallow-transient-error-and-retry / ctx-sets-stop-on-exit. **78 consumer tests pass.**
- The end-to-end "dead worker redelivers in ~120s" **wall-clock timing is validated in the integration environment** (real Postgres + worker + kill), alongside the flag-ON regression.

Part of UN-3695. Depends on nothing; complements the merged #2153.

### DIFF
--- a/workers/queue_backend/pg_queue/README.md
+++ b/workers/queue_backend/pg_queue/README.md
@@ -27,7 +27,8 @@ integers.
 | Env suffix | One-line | Default |
 |---|---|---|
 | `CONCURRENCY` | Prefork consumer children per pod (1 = plain single process) | `1` |
-| `VT_SECONDS` | **Visibility timeout** — how long a claimed message stays hidden before it can redeliver | `30` (chart: `9060` for file-processing ≈ 2.5h) |
+| `VT_SECONDS` | Since PR C, the **drain / max-runtime bound** (drives `SHUTDOWN_GRACE`, `HEALTH_STALE`, the chart guards) — the *claim* window is `LEASE_SECONDS`, not this | `30` (chart: `9060` for file-processing ≈ 2.5h) |
+| `LEASE_SECONDS` | **Renewable claim window** — a claimed message is hidden only this long and renewed every ~LEASE/3 while the task runs; a dead worker's claim expires in ~LEASE → fast redelivery (capped at VT) | `120` |
 | `SHUTDOWN_GRACE_SECONDS` | Graceful-drain budget (shared across all children) on SIGTERM before SIGKILL | `= VT` (floored at `30`) |
 | `QUEUE` | Queue name(s) this consumer polls (comma-separated) | `default` |
 | `BATCH` | Messages claimed per poll | `1` |
@@ -64,10 +65,19 @@ integers.
 ## 2. Concepts — glossary
 
 **Visibility timeout (VT)** — when a consumer *claims* a message it becomes invisible
-to other consumers for `VT` seconds. If the worker finishes and deletes (acks) it, it's
-gone; if the worker dies without acking, the VT expires and the message **redelivers**.
-VT is the queue's only "worker died" signal (there's no live broker connection like
-RabbitMQ), so recovery latency ≈ VT.
+to other consumers until its `vt` passes. If the worker finishes and deletes (acks) it,
+it's gone; if the worker dies without acking, the `vt` expires and the message
+**redelivers**. There's no live broker connection like RabbitMQ, so the `vt` is the
+queue's only "worker died" signal — recovery latency ≈ how far out the `vt` is.
+
+**Renewable lease (`LEASE_SECONDS`, UN-3695 PR C)** — rather than claim for the full
+`VT` (up to 2.5h), the consumer claims for a **short** `LEASE` and a background thread
+**renews** it (`set_vt`) every ~`LEASE/3` while the task runs. A live-but-slow task
+keeps its claim; a **dead** worker's renewal stops, so its `vt` expires in ~`LEASE` →
+redelivery in **minutes, not hours**. `VT_SECONDS` is retained as the *drain /
+max-runtime bound* (grace, health-stale), and the lease is capped at it. The renewal
+uses its own DB connection and is best-effort (a transient failure retries within the
+`~2×` slack the `LEASE/3` interval leaves before expiry).
 
 **Claim** — an atomic `SELECT … FOR UPDATE SKIP LOCKED` that hides up to `BATCH`
 ready rows for `VT` seconds and hands them to one consumer. `SKIP LOCKED` distributes
--- a/workers/queue_backend/pg_queue/README.md
+++ b/workers/queue_backend/pg_queue/README.md
@@ -27,8 +27,8 @@ integers.
 | Env suffix | One-line | Default |
 |---|---|---|
 | `CONCURRENCY` | Prefork consumer children per pod (1 = plain single process) | `1` |
-| `VT_SECONDS` | Since PR C, the **drain / max-runtime bound** (drives `SHUTDOWN_GRACE`, `HEALTH_STALE`, the chart guards) — the *claim* window is `LEASE_SECONDS`, not this | `30` (chart: `9060` for file-processing ≈ 2.5h) |
-| `LEASE_SECONDS` | **Renewable claim window** — a claimed message is hidden only this long and renewed every ~LEASE/3 while the task runs; a dead worker's claim expires in ~LEASE → fast redelivery (capped at VT) | `120` |
+| `VT_SECONDS` | The **drain / max-runtime bound** (drives `SHUTDOWN_GRACE`, `HEALTH_STALE`, the chart guards) — the *claim* window is `LEASE_SECONDS`, not this | `30` (chart: `9060` for file-processing ≈ 2.5h) |
+| `LEASE_SECONDS` | **Renewable claim window** — the effective claim window is `min(LEASE, VT)`, renewed every ~that/3 while the task runs; a dead worker's claim expires in ~that → fast redelivery. With the defaults (`LEASE=120`, `VT=30`) it clamps to 30; the chart raises VT so the full 120 applies | `120` |
 | `SHUTDOWN_GRACE_SECONDS` | Graceful-drain budget (shared across all children) on SIGTERM before SIGKILL | `= VT` (floored at `30`) |
 | `QUEUE` | Queue name(s) this consumer polls (comma-separated) | `default` |
 | `BATCH` | Messages claimed per poll | `1` |
@@ -70,18 +70,23 @@ it's gone; if the worker dies without acking, the `vt` expires and the message
 **redelivers**. There's no live broker connection like RabbitMQ, so the `vt` is the
 queue's only "worker died" signal — recovery latency ≈ how far out the `vt` is.
 
-**Renewable lease (`LEASE_SECONDS`, UN-3695 PR C)** — rather than claim for the full
-`VT` (up to 2.5h), the consumer claims for a **short** `LEASE` and a background thread
+**Renewable lease (`LEASE_SECONDS`, UN-3695)** — rather than claim for the full `VT`
+(up to 2.5h), the consumer claims for a **short** `LEASE` and a background thread
 **renews** it (`set_vt`) every ~`LEASE/3` while the task runs. A live-but-slow task
 keeps its claim; a **dead** worker's renewal stops, so its `vt` expires in ~`LEASE` →
 redelivery in **minutes, not hours**. `VT_SECONDS` is retained as the *drain /
-max-runtime bound* (grace, health-stale), and the lease is capped at it. The renewal
-uses its own DB connection and is best-effort (a transient failure retries within the
-`~2×` slack the `LEASE/3` interval leaves before expiry).
+max-runtime bound* (grace, health-stale), and the lease is clamped to it. Note the
+lease is **not** a hard cap on runtime — a live-but-hung task keeps renewing forever;
+the backstop for that is the liveness probe restarting the pod (process death stops
+renewal). The renewal owns its own DB connection (closed on exit) and is best-effort:
+a connection death retries within the `~2×` slack the `LEASE/3` interval leaves before
+expiry, and escalates to an ERROR log once it keeps failing past `LEASE` (the lease is
+then genuinely lost and the message may double-run). Because renewal covers only the
+in-flight message, `BATCH_SIZE` is forced to 1 whenever the lease is the short window.
 
 **Claim** — an atomic `SELECT … FOR UPDATE SKIP LOCKED` that hides up to `BATCH`
-ready rows for `VT` seconds and hands them to one consumer. `SKIP LOCKED` distributes
-work across children and replicas without contention.
+ready rows for the claim window (`min(LEASE, VT)`) and hands them to one consumer.
+`SKIP LOCKED` distributes work across children and replicas without contention.
 
 **Redelivery / at-least-once** — a message can be delivered more than once (VT expiry
 after a crash, or a poison re-park). Handlers must tolerate re-execution.
--- a/workers/queue_backend/pg_queue/consumer.py
+++ b/workers/queue_backend/pg_queue/consumer.py
@@ -27,7 +27,7 @@ import os
 import signal
 import threading
 import time
-from collections.abc import Callable
+from collections.abc import Callable, Iterator
 from enum import Enum
 from typing import TYPE_CHECKING, TypeVar
 
@@ -38,6 +38,7 @@ from unstract.core.data_models import ContinuationSpec, TaskPayload
 from ..barrier import callback_recovery_identity
 from ..fairness import FAIRNESS_HEADER_NAME
 from .client import PgQueueClient
+from .connection import CONN_DEAD_ERRORS
 from .liveness import LivenessServer as _BaseLivenessServer
 from .result_backend import PgResultBackend
 from .task_payload import to_payload
@@ -59,7 +60,7 @@ _DEFAULT_QUEUE = "default"
 # raise it, keep vt_seconds > batch_size x worst-case task duration.
 _DEFAULT_BATCH = 1
 _DEFAULT_VT_SECONDS = 30
-# Renewable-lease claim window (UN-3695 PR C). A claimed message is hidden for only
+# Renewable-lease claim window (UN-3695). A claimed message is hidden for only
 # this long; a background thread renews it (~every LEASE/3) while the task runs, so a
 # live-but-slow task stays claimed but a DEAD worker's claim expires in ~LEASE →
 # redelivery in minutes instead of the full VT. VT_SECONDS is now the drain /
@@ -258,16 +259,33 @@ class PgQueueConsumer:
         self._api_client = api_client
         self.batch_size = batch_size
         self.vt_seconds = vt_seconds
-        # Renewable lease (UN-3695 PR C): claim for lease_seconds (short) and renew
-        # every ~lease/3 while a task runs. Capped at vt_seconds so the lease is
-        # genuinely the shorter window; setting lease >= vt degrades to the old
-        # single-long-claim behaviour (renewal fires rarely).
+        # Renewable lease (UN-3695): the claim is taken for a short LEASE and renewed
+        # every ~lease/3 while a task runs, so a dead worker's claim expires in ~LEASE
+        # (fast redelivery) but a live one is never redelivered. VT_SECONDS is the
+        # drain / max-runtime bound; a lease longer than it is meaningless, so clamp —
+        # loudly, since a lease > VT is a misconfig, not a silent no-op.
         self.lease_seconds = min(lease_seconds, vt_seconds)
+        if lease_seconds > vt_seconds:
+            logger.warning(
+                "WORKER_PG_QUEUE_CONSUMER_LEASE_SECONDS=%s > VT_SECONDS=%s; clamped the "
+                "claim window to %ss (the lease can't outlast the max-runtime bound)",
+                lease_seconds,
+                vt_seconds,
+                self.lease_seconds,
+            )
         self._lease_renew_interval = max(1, self.lease_seconds // 3)
-        # Second client (own connection) for lease renewals — the main thread owns
-        # self._client for the read/ack, so a renewal must not share that connection.
-        # Lazily created on the first task run.
-        self._renew_client: PgQueueClient | None = None
+        # The renewal only extends the IN-FLIGHT message; a batch tail sits claimed but
+        # unrenewed, so with batch>1 a slow head lets the tail's lease lapse and another
+        # consumer double-runs it. Force serial claims whenever the lease is the short
+        # (renewed) claim window; batch>1 stays available when lease == vt (no renewal).
+        if self.batch_size > 1 and self.lease_seconds < vt_seconds:
+            logger.warning(
+                "WORKER_PG_QUEUE_CONSUMER_BATCH_SIZE=%s forced to 1: the renewable lease "
+                "only covers the in-flight message, so a batch tail could lapse and "
+                "double-run (UN-3695)",
+                self.batch_size,
+            )
+            self.batch_size = 1
         self.poll_interval = poll_interval
         self.backoff_max = backoff_max
         self.max_attempts = max_attempts
@@ -313,16 +331,18 @@ class PgQueueConsumer:
         return total
 
     @contextlib.contextmanager
-    def _lease_renewal(self, msg_id: int):
-        """Keep ``msg_id``'s claim alive while a task runs (UN-3695 PR C).
+    def _lease_renewal(self, msg_id: int) -> Iterator[None]:
+        """Keep ``msg_id``'s claim alive while a task runs (UN-3695).
 
-        The claim was taken for the short ``lease_seconds``; this starts a daemon
-        thread that renews it (``set_vt``) every ``lease/3`` — so a live-but-slow task
-        stays claimed — then stops + joins on exit. If the worker DIES, the thread
-        dies with it → the lease expires in ~``lease_seconds`` → the message
-        redelivers in minutes instead of the full VT. The renewal uses its OWN
-        connection (``_renew_client``); the main thread owns ``_client`` for the ack,
-        and the thread is joined here before that ack so the two never race the row.
+        Starts a daemon thread that renews the short lease (``set_vt``) every
+        ``lease/3`` — so a live-but-slow task stays claimed — then signals stop and
+        joins on exit, before the caller acks. If the worker DIES, the thread dies
+        with it → the lease expires in ~``lease_seconds`` → the message redelivers in
+        minutes instead of the full VT. The join is bounded by
+        ``_LEASE_JOIN_TIMEOUT_SECONDS``; a thread still alive after that (wedged in a
+        stalled ``set_vt``) is logged and abandoned — it owns its own connection (see
+        the loop), so it can't corrupt the ack path, and a late ``set_vt`` on the
+        about-to-be-acked row is a benign no-op.
         """
         stop = threading.Event()
         thread = threading.Thread(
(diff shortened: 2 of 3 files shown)


workers/queue_backend/pg_queue/consumer.py
```diff
@@ -276,7 +299,7 @@ def poll_once(self) -> int:
         for queue_name in self.queue_names:
             try:
                 messages = self._client.read(
-                    queue_name, vt_seconds=self.vt_seconds, qty=self.batch_size
+                    queue_name, vt_seconds=self.lease_seconds, qty=self.batch_size
```
muhammad-ali-e, line 320: **[High] BATCH>1 tail messages can double-process.** `read()` now claims the whole batch for `lease_seconds`, but `_handle` processes messages sequentially and `_lease_renewal` only renews the *in-flight* `msg_id`. Messages #2..N sit in the local list unrenewed; if the head task runs longer than `lease_seconds` (routine for file-processing) their `vt` expires, another consumer reclaims them, and they run twice. Before this PR the tail was hidden for the full `vt_seconds` (~2.5h). Safe at the default `BATCH=1`, but `BATCH` is a supported knob on exactly the long-task/large-VT profile. Fix options: renew every still-unacked `msg_id` in the batch; or claim the tail for `vt_seconds`; or force `batch_size=1` when `lease_seconds < vt_seconds`. Coupled to the stale comment at line 59.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -57,6 +59,15 @@
 # raise it, keep vt_seconds > batch_size x worst-case task duration.
```
muhammad-ali-e, line 60: **[High] This mitigation note is now wrong and will mislead operators.** The batch claim window is `lease_seconds` (line 302), not `vt_seconds`, so 'keep vt_seconds > batch_size × worst-case task duration' no longer protects the batch tail — raising `VT_SECONDS` does nothing for queued-but-unstarted messages once the lease caps the claim window. Update to reference `LEASE_SECONDS` and tie it to the batch-renewal fix (line 302).

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -289,6 +312,59 @@ def poll_once(self) -> int:
...
+        """
+        while not stop.wait(self._lease_renew_interval):
+            try:
+                # Created on the FIRST actual renewal (not on thread start), so a
+                # task shorter than the interval opens no second connection.
+                if self._renew_client is None:
+                    self._renew_client = PgQueueClient()
+                if not self._renew_client.set_vt(msg_id, self.lease_seconds):
+                    return  # row gone — nothing left to renew
+            except Exception:
```
muhammad-ali-e: **[High] `except Exception` is too broad, and repeated failures never escalate.** (1) A deterministic bug (an `AttributeError` from a rename, or `set_vt`'s own `ValueError`) is caught every tick and logged as a self-healing 'blip' — it never self-heals, the lease is never renewed, and every long task silently double-runs. Narrow the catch to the connection-dead types (`_CONN_DEAD_ERRORS`) and log unexpected types at ERROR, then stop retrying. (2) Even for genuine transient failures: if the DB is down longer than `lease_seconds` the lease is *lost* and the task double-runs, yet every tick emits the same reassuring WARNING ('a dead connection self-heals on the next tick'). Track last-success time and escalate to ERROR with a distinct, greppable message once the failure window exceeds `lease_seconds` (the point past which a double-run is provably possible).

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -289,6 +312,59 @@ def poll_once(self) -> int:
...
+        current lease expires); a ``False`` return means the row is already gone
+        (acked by us / expired + reclaimed) → stop.
+        """
+        while not stop.wait(self._lease_renew_interval):
+            try:
+                # Created on the FIRST actual renewal (not on thread start), so a
+                # task shorter than the interval opens no second connection.
+                if self._renew_client is None:
+                    self._renew_client = PgQueueClient()
+                if not self._renew_client.set_vt(msg_id, self.lease_seconds):
```
muhammad-ali-e: **[High] `set_vt` returning `False` mid-run is an unlogged double-run signal.** The renewal thread is joined *before* the ack (line 339; the `delete` happens after the `with` block exits), so it can never observe our own ack — the docstring's 'acked by us' case is impossible in this thread. Mid-run, `False` means the lease already expired and another worker reclaimed the row, i.e. the task is double-running *right now*. Yet the loop just `return`s silently with no log. Emit a WARNING (arguably ERROR) before returning, and correct the docstring at lines 347-348 — 'acked by us' should read 'expired + reclaimed → likely double-running'.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -289,6 +312,59 @@ def poll_once(self) -> int:
...
+            args=(msg_id, stop),
+            name=f"pg-lease-{msg_id}",
+            daemon=True,
+        )
+        thread.start()
+        try:
+            yield
+        finally:
+            stop.set()
+            thread.join(timeout=_LEASE_JOIN_TIMEOUT_SECONDS)
```
muhammad-ali-e, line 359: **[Medium] Wedged renewal thread: the join can time out silently, and the shared `_renew_client` can then be used by two threads.** If `set_vt` is blocked in a hung DB call, `join(timeout=10)` returns with the thread still alive and execution falls through to the ack with no `is_alive()` check and no log. Compounding hazard: the next message starts a *new* renewal thread that reuses the same `self._renew_client` connection (line 355) — two threads on one non-thread-safe psycopg connection → protocol corruption that surfaces later on an unrelated statement. Add an `if thread.is_alive():` ERROR log after the join, and consider not sharing `_renew_client` across tasks while a prior thread may still be alive.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -245,6 +258,16 @@ def __init__(
         self._api_client = api_client
         self.batch_size = batch_size
         self.vt_seconds = vt_seconds
+        # Renewable lease (UN-3695 PR C): claim for lease_seconds (short) and renew
+        # every ~lease/3 while a task runs. Capped at vt_seconds so the lease is
+        # genuinely the shorter window; setting lease >= vt degrades to the old
+        # single-long-claim behaviour (renewal fires rarely).
+        self.lease_seconds = min(lease_seconds, vt_seconds)
```
muhammad-ali-e, line 267: **[Medium] Silent cap hides operator misconfig, and the adjacent comment is factually wrong.** (a) `min(lease_seconds, vt_seconds)` silently discards intent — `LEASE_SECONDS=200` with `VT=100` yields 100 with no log or error, inconsistent with the neighboring `backoff_max >= poll_interval` check ~30 lines up that raises loudly. Prefer a `raise`, or at least a `logger.warning` naming both values so the clamp is observable. (b) The comment at 263-264 ('setting lease >= vt degrades to the old single-long-claim behaviour (renewal fires rarely)') is wrong: with `min()` the effective lease becomes `vt` and the interval `vt // 3`, so the renewal thread still fires ~3×/window; only the dead-worker recovery latency degrades back to ≈VT. Reword.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -245,6 +258,16 @@ def __init__(
...
+        self._lease_renew_interval = max(1, self.lease_seconds // 3)
+        # Second client (own connection) for lease renewals — the main thread owns
+        # self._client for the read/ack, so a renewal must not share that connection.
+        # Lazily created on the first task run.
+        self._renew_client: PgQueueClient | None = None
```
muhammad-ali-e: **[Medium] `_renew_client` is never closed.** It's created lazily (line 355) and no path closes it — `stop()` only flips `self._running`. Each consumer child that runs a task longer than `lease/3` holds a second Postgres connection for the whole process lifetime, doubling steady-state connections against PgBouncer with no release on graceful shutdown. (The main `_client` is also unclosed, so it's symmetric — but the PgBouncer-reap care taken elsewhere suggests this should be a deliberate choice, not an omission.) Add a guarded `self._renew_client.close()` to `stop()`.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -289,6 +312,59 @@ def poll_once(self) -> int:
...
+    def _lease_renewal(self, msg_id: int):
+        """Keep ``msg_id``'s claim alive while a task runs (UN-3695 PR C).
+
+        The claim was taken for the short ``lease_seconds``; this starts a daemon
+        thread that renews it (``set_vt``) every ``lease/3`` — so a live-but-slow task
+        stays claimed — then stops + joins on exit. If the worker DIES, the thread
+        dies with it → the lease expires in ~``lease_seconds`` → the message
+        redelivers in minutes instead of the full VT. The renewal uses its OWN
+        connection (``_renew_client``); the main thread owns ``_client`` for the ack,
+        and the thread is joined here before that ack so the two never race the row.
```
muhammad-ali-e: **[Low] Docstring overstates the guarantee.** 'the thread is joined here before that ack so the two never race the row' — the join has a 10s timeout (line 339); on timeout the thread is still alive and `set_vt` can run after the `delete`. The race is benign (`set_vt` on a deleted row updates 0 rows → `False`), but 'never race' contradicts the honest constant comment at lines 66-67. Soften to: 'joined best-effort, bounded by `_LEASE_JOIN_TIMEOUT_SECONDS`; a late `set_vt` on the acked row no-ops.'

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -289,6 +312,59 @@ def poll_once(self) -> int:
                 )
         return total
 
+    @contextlib.contextmanager
+    def _lease_renewal(self, msg_id: int):
```
muhammad-ali-e: **[Low] Missing return-type annotation.** This is the only method in the file without one. A `@contextmanager` generator should be `-> Iterator[None]` (add `Iterator` to the `collections.abc` import at line 30).

workers/queue_backend/pg_queue/README.md
```diff
@@ -27,7 +27,8 @@ integers.
 | Env suffix | One-line | Default |
 |---|---|---|
 | `CONCURRENCY` | Prefork consumer children per pod (1 = plain single process) | `1` |
-| `VT_SECONDS` | **Visibility timeout** — how long a claimed message stays hidden before it can redeliver | `30` (chart: `9060` for file-processing ≈ 2.5h) |
+| `VT_SECONDS` | Since PR C, the **drain / max-runtime bound** (drives `SHUTDOWN_GRACE`, `HEALTH_STALE`, the chart guards) — the *claim* window is `LEASE_SECONDS`, not this | `30` (chart: `9060` for file-processing ≈ 2.5h) |
+| `LEASE_SECONDS` | **Renewable claim window** — a claimed message is hidden only this long and renewed every ~LEASE/3 while the task runs; a dead worker's claim expires in ~LEASE → fast redelivery (capped at VT) | `120` |
```
muhammad-ali-e: **[Low] `LEASE_SECONDS` default 120 is misleading under the default VT.** Because the effective lease is `min(LEASE, VT)`, with the default `VT=30` the lease clamps to 30 and the renew interval to 10s — not the 120 / ~40s a reader infers from this row. The minutes-not-hours benefit only materializes once VT is raised (the `9060` file-processing chart). Add a parenthetical, e.g. `120` (effective = `min(LEASE, VT)`; with the default `VT=30` this clamps to `30`).

workers/queue_backend/pg_queue/README.md
```diff
@@ -64,10 +65,19 @@ integers.
...
+it's gone; if the worker dies without acking, the `vt` expires and the message
+**redelivers**. There's no live broker connection like RabbitMQ, so the `vt` is the
+queue's only "worker died" signal — recovery latency ≈ how far out the `vt` is.
+
+**Renewable lease (`LEASE_SECONDS`, UN-3695 PR C)** — rather than claim for the full
+`VT` (up to 2.5h), the consumer claims for a **short** `LEASE` and a background thread
+**renews** it (`set_vt`) every ~`LEASE/3` while the task runs. A live-but-slow task
+keeps its claim; a **dead** worker's renewal stops, so its `vt` expires in ~`LEASE` →
+redelivery in **minutes, not hours**. `VT_SECONDS` is retained as the *drain /
+max-runtime bound* (grace, health-stale), and the lease is capped at it. The renewal
```
muhammad-ali-e: **[Low] The lease does not self-cap a live-but-hung task.** The renewal loop has no absolute deadline — a live-but-wedged task renews forever and is never redelivered, so `VT_SECONDS` is not a max-runtime bound *via the lease*. In practice `task.apply` runs synchronously on the poll thread, so a hung task stalls the consumer and the external health-stale / liveness probe kills the pod (process death stops renewal → redelivery). Consider rewording so operators understand the bound is enforced by the probe killing the pod, not by the lease self-capping.

workers/tests/test_pg_queue_consumer.py
```diff
@@ -1055,3 +1056,65 @@ def test_chain_continuation_records_status_even_if_enqueue_fails(self):
         self._entered_rb(RB).store_result.assert_called_once_with(
             "tid", result={}, retention_seconds=self._RET
         )
+
+
+class TestLeaseRenewal:
```
muhammad-ali-e, line 1063: **[Medium] Test gaps in the new suite.** Every test pre-sets `_renew_client = MagicMock()`, so the load-bearing paths are untested: (1) lazy `PgQueueClient()` creation — that it's built *once* and *not* built for a sub-interval task (the 'no second connection' guarantee); (2) `_handle` integration — that `task.apply` runs inside `_lease_renewal` and the thread is joined *before* the ack (deleting the `with` wrapper wouldn't fail any current test); (3) the join-timeout / wedged-thread path; (4) `_renew_client` close on shutdown (see consumer.py:270); (5) boundary values — `lease_seconds<=0` rejected at construction, `lease == vt`, and the tiny-lease `max(1, //3)` no-slack edge. Also `stop.wait` is a bare mock, so the timing invariant `_lease_renew_interval < lease_seconds` is never asserted.

muhammad-ali-e (commented):

The renewable-lease change is well-structured overall: the daemon-thread + `Event` + context-manager shape is idiomatic, the lazy second connection and join-before-ack are well-reasoned, and the happy-path tests are solid. Findings are grouped below and posted as inline comments.

**High**
- BATCH>1 tail double-processing (`consumer.py:302`) + now-stale mitigation comment (`consumer.py:59`)
- Renewal error handling: over-broad `except` + no escalation on prolonged failure (`consumer.py:358`)
- `set_vt`→`False` mid-run is an unlogged double-run signal (`consumer.py:356`)

**Medium**
- Wedged-join not logged + shared `_renew_client` reuse race (`consumer.py:339`)
- Silent `lease>vt` cap + factually wrong 'fires rarely' comment (`consumer.py:265`)
- `_renew_client` never closed (`consumer.py:270`)
- Test gaps: lazy-init, `_handle` integration, close, boundaries (test file)

**Low**
- 'never race' docstring overstated (`consumer.py:325`)
- Missing `-> Iterator[None]` (`consumer.py:316`)
- README: LEASE default vs default VT (`README.md:31`); lease is not the max-runtime bound (`README.md:78`)

The default config (`BATCH=1`) is safe from the top item; the batch and error-handling findings bite specifically under the file-processing profile this feature targets (large VT, batching, long tasks).

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e, line 359: **[Important] Bounded join can leave a wedged renewal thread sharing `_renew_client` with the next task's thread → concurrent use of one psycopg2 connection.**

If a renewal thread is blocked inside `set_vt` on a stalled connection (DB failover / PgBouncer reap / network partition — exactly the window this feature targets) for >10s, `join(timeout=…)` returns silently and `_handle` proceeds. The main loop then advances to the next message, whose renewal thread reuses the **same** `self._renew_client` (`if self._renew_client is None` at :354 is already non-None). Two threads issuing `set_vt` on one libpq connection is undefined behavior (protocol desync), which can then make renewals fail and reintroduce the very redelivery this PR prevents. The docstring's "joined here before that ack so the two never race the row" (:325) also does not hold once the join times out.

Suggested fix: give each renewal its own `PgQueueClient` scoped to `_lease_renewal` (and `close()` it), **or** guard `_renew_client` use with a `threading.Lock`. Also log + increment a metric when `thread.is_alive()` after the join — an abandoned thread is currently completely silent.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -289,6 +312,59 @@ def poll_once(self) -> int:
...
+        (acked by us / expired + reclaimed) → stop.
+        """
+        while not stop.wait(self._lease_renew_interval):
+            try:
+                # Created on the FIRST actual renewal (not on thread start), so a
+                # task shorter than the interval opens no second connection.
+                if self._renew_client is None:
+                    self._renew_client = PgQueueClient()
+                if not self._renew_client.set_vt(msg_id, self.lease_seconds):
+                    return  # row gone — nothing left to renew
```
muhammad-ali-e: **[Important] `set_vt` → `False` (row gone) returns with zero logging — the most definitive double-run signal is silent.**

During an in-flight renewal the ack has not run yet (the thread is joined *before* the ack at :504). Verified against `client.py`: `read()` reclaim only UPDATEs `vt`/`read_ct`, it does **not** delete — so `set_vt` returns `False` only after the row was actually DELETED, i.e. another worker already reclaimed **and acked** it → this task is being/has-been double-processed. Contrast the ack path at :504-510, which logs a WARNING for the *same* condition. Here it's a bare `return`.

Suggested fix: `logger.error(...)` naming `msg_id` + a `lease_lost` counter before returning.

Also: the docstring at :347 lists "acked by us" as a cause of `False` — that's impossible inside the loop (ack happens after join), and "expired + reclaimed" does **not** by itself return `False` (reclaim doesn't delete). Reword to: "`False` = the row was deleted (acked/dropped by its current owner)."

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -289,6 +312,59 @@ def poll_once(self) -> int:
...
+            except Exception:
+                logger.warning(
```
muhammad-ali-e: **[Important] The loop can't distinguish a transient blip from a definitively-lost lease — sustained failure silently allows a concurrent double-run.**

Every `set_vt` exception logs an identical WARNING and loops with no state about how long renewals have been failing. If `set_vt` fails on every tick for longer than `lease_seconds` (dead conn / failover), the lease expires **while `task.apply` is still running**, another worker reclaims the row, and the task runs twice — the worker that lost the lease emits only look-alike WARNINGs, never escalates. Given this subsystem's own `delete()` docstring ties a lost lease to duplicate webhooks + subscription-usage billing, a blip and a total loss should not be indistinguishable in the logs.

Suggested fix: track `last_ok = time.monotonic()`; once `now - last_ok >= lease_seconds`, escalate to `logger.error` + a `lease_renewal_giveup` metric so it pages, while keeping the sub-lease case at WARNING.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -245,6 +258,16 @@ def __init__(
         self._api_client = api_client
         self.batch_size = batch_size
         self.vt_seconds = vt_seconds
+        # Renewable lease (UN-3695 PR C): claim for lease_seconds (short) and renew
+        # every ~lease/3 while a task runs. Capped at vt_seconds so the lease is
+        # genuinely the shorter window; setting lease >= vt degrades to the old
+        # single-long-claim behaviour (renewal fires rarely).
```
muhammad-ali-e: **[Important] This comment is factually wrong.** After `self.lease_seconds = min(lease_seconds, vt_seconds)` (:265) the effective lease can never exceed `vt`, so `_lease_renew_interval = vt//3` and renewal fires every `vt/3` — *frequently*, re-parking on every tick. That is neither "single-long-claim behaviour" (which never renews) nor "renewal fires rarely."

Suggested: "setting `lease >= vt` just clamps the lease to `vt`; renewal still fires every `vt/3` (redundantly, each extending by `vt`)."

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e, line 267: **[Minor] With the shipped code defaults the feature is inert.** `_DEFAULT_LEASE_SECONDS = 120` (:67) but `_DEFAULT_VT_SECONDS = 30` (:66-ish), so a default-constructed consumer gets `min(120, 30) = 30` — the advertised 120s lease is silently discarded and lease == vt. It only does something once `VT_SECONDS` is raised well above the lease (which the chart does in prod: VT=9060). This is a dev/local sharp edge, not a prod bug — but worth either making the defaults coherent (lower the default lease below the default VT, or raise the default VT to match its new "max-runtime bound" role) or `logger.warning` when `lease_seconds > vt_seconds` so the clamp is visible. Note the constructor rejects `backoff_max < poll_interval` explicitly (:236) yet clamps this cross-field invariant silently — inconsistent.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -289,6 +312,59 @@ def poll_once(self) -> int:
...
+            daemon=True,
+        )
+        thread.start()
+        try:
+            yield
+        finally:
+            stop.set()
+            thread.join(timeout=_LEASE_JOIN_TIMEOUT_SECONDS)
+
+    def _renew_lease_loop(self, msg_id: int, stop: threading.Event) -> None:
```
muhammad-ali-e, line 373: **[Suggestion] No lease-failure mode is wired to metrics/alerting**, though a `ConsumerMetrics` Prometheus facility is already used by this consumer. Every failure here is log-only, so operators can't see how often leases are lost in prod (i.e. how often duplicate execution occurs). Suggest adding counters — `pg_lease_renewal_failures_total`, `pg_lease_lost_total`, `pg_lease_renewal_giveup_total`, `pg_lease_thread_wedged_total` — incremented at the sites flagged above, and alerting on any non-zero `lease_lost` / `giveup` rate.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: **[Suggestion] `_renew_client` is created but never `close()`d.** Not an unbounded leak (one connection per consumer lifetime, and consistent with `self._client` also not being closed), but the create-without-close asymmetry is a smell — it survives as an idle connection between tasks. Consider closing both `self._client` and `self._renew_client` (guard for `None`) on a shutdown/`stop()` path. Scoping the renew client into `_lease_renewal` (see the concurrency comment above) would remove this field entirely and make the lifetime obvious.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -245,6 +258,16 @@ def __init__(
...
         self.vt_seconds = vt_seconds
+        # Renewable lease (UN-3695 PR C): claim for lease_seconds (short) and renew
+        # every ~lease/3 while a task runs. Capped at vt_seconds so the lease is
+        # genuinely the shorter window; setting lease >= vt degrades to the old
+        # single-long-claim behaviour (renewal fires rarely).
+        self.lease_seconds = min(lease_seconds, vt_seconds)
+        self._lease_renew_interval = max(1, self.lease_seconds // 3)
+        # Second client (own connection) for lease renewals — the main thread owns
+        # self._client for the read/ack, so a renewal must not share that connection.
+        # Lazily created on the first task run.
```
muhammad-ali-e: **[Nit] Imprecise + duplicated comment.** "Lazily created on the first task **run**" is slightly wrong — it's created on the first *renewal* (:352-353 states this accurately), which only happens when a task outlives the interval; a short task opens no second connection. Suggest: "Lazily created on the first lease renewal (never for tasks shorter than the interval)." Keep only one authoritative statement of this — :352-353 already has it.

workers/queue_backend/pg_queue/consumer.py
muhammad-ali-e: **[Nit] Missing return-type annotation.** The rest of the file annotates returns (`poll_once -> int`, `_renew_lease_loop -> None`); this contextmanager omits it. Add `-> Iterator[None]` (`from collections.abc import Iterator`) so the generator contract is expressed and type-checkable.

workers/tests/test_pg_queue_consumer.py
muhammad-ali-e, line 1063: **[Tests] Good loop-logic coverage; the gaps are at the seams.** The state-machine tests (cap, interval, claim-uses-lease, renew-until-stopped, stop-on-row-gone, swallow-and-retry, ctx stop-on-exit) are solid. Missing, roughly in priority order:

1. **Integration** — nothing drives `_handle`/`poll_once` to assert `task.apply` is actually wrapped in `_lease_renewal(msg_id)` (consumer.py:424-430). A refactor dropping the `with` would keep all 7 tests green.
2. **Ordering** — the "joined before ack" safety claim (thread stopped/joined at :338-339 before `delete` at :504) is unasserted.
3. **Lazy `_renew_client`** — every loop test pre-sets `c._renew_client = MagicMock()`, so the `if self._renew_client is None: PgQueueClient()` branch (:354-355) and the "no connection for short tasks" promise have zero coverage.
4. **`lease_seconds <= 0` validation** — `TestConstruction.test_rejects_non_positive_params` enumerates every other positive-int param but omits `lease_seconds`; add `0` and `-5` cases (trivial, guards the zero-lease continuous-double-run foot-gun).
5. **`LEASE_SECONDS` env wiring** — `build_consumer_from_env` (consumer.py:999) has no test at all; the operator-facing knob could be mis-spelled/dropped with CI green.
6. **Default cap 120→30** and the `max(1, lease//3)` floor for tiny leases (1/2 → 1) and the `lease == vt` boundary are untested.
7. **Sustained `set_vt` failure past the lease** (the double-run boundary) is untested.

workers/queue_backend/pg_queue/README.md
```diff
@@ -27,7 +27,8 @@ integers.
 | Env suffix | One-line | Default |
 |---|---|---|
 | `CONCURRENCY` | Prefork consumer children per pod (1 = plain single process) | `1` |
-| `VT_SECONDS` | **Visibility timeout** — how long a claimed message stays hidden before it can redeliver | `30` (chart: `9060` for file-processing ≈ 2.5h) |
+| `VT_SECONDS` | Since PR C, the **drain / max-runtime bound** (drives `SHUTDOWN_GRACE`, `HEALTH_STALE`, the chart guards) — the *claim* window is `LEASE_SECONDS`, not this | `30` (chart: `9060` for file-processing ≈ 2.5h) |
```
muhammad-ali-e: **[Docs] These two rows are misleading for the default/local config.** `VT_SECONDS` row asserts "the *claim* window is `LEASE_SECONDS`, not this" — but with the default `VT=30 < LEASE=120` the effective claim window **is** VT (30), because of `min(lease, vt)`. And the `LEASE_SECONDS` Default column says `120` while the effective default is 30. Suggest: VT row → "the claim window is `min(LEASE_SECONDS, VT_SECONDS)`"; LEASE row Default → "`120` (effective = `min(LEASE, VT)`; with the default `VT=30`, 30)".

Also: "Since PR C" / "UN-3695 PR C" shorthand (:30, :73, and consumer.py:62/261/317/423) will rot after merge — no artifact called "PR C" survives. Prefer the `UN-3695` Jira ref or describe the behavior directly, especially in the README where a standalone reader can't resolve it.

workers/queue_backend/pg_queue/README.md
```diff
@@ -64,10 +65,19 @@ integers.
...
+**Renewable lease (`LEASE_SECONDS`, UN-3695 PR C)** — rather than claim for the full
+`VT` (up to 2.5h), the consumer claims for a **short** `LEASE` and a background thread
+**renews** it (`set_vt`) every ~`LEASE/3` while the task runs. A live-but-slow task
+keeps its claim; a **dead** worker's renewal stops, so its `vt` expires in ~`LEASE` →
+redelivery in **minutes, not hours**. `VT_SECONDS` is retained as the *drain /
+max-runtime bound* (grace, health-stale), and the lease is capped at it. The renewal
+uses its own DB connection and is best-effort (a transient failure retries within the
+`~2×` slack the `LEASE/3` interval leaves before expiry).
 
 **Claim** — an atomic `SELECT … FOR UPDATE SKIP LOCKED` that hides up to `BATCH`
```
muhammad-ali-e, line 87: **[Docs] Stale glossary entry.** The "Claim" definition still says it hides rows "for `VT` seconds," but `poll_once` now reads with `vt_seconds=self.lease_seconds` (consumer.py:302). Update to "...hides up to `BATCH` ready rows for the **lease** window (`min(LEASE, VT)`)..." so it matches the PR-C model introduced just above.

muhammad-ali-e (commented):

Ran 6 specialized agents (code review, silent-failure hunt, type design, test coverage, comment accuracy, simplification) over the renewable-lease change. The design is sound and the happy path (live worker renews / dead worker's lease expires) is correct and well-tested. Findings below are grouped as inline comments.

**Priority order:**
1. **Important** — shared `_renew_client` concurrency race when the bounded join times out (2 agents found this independently).
2. **Important** — three silent/under-observed unhappy-path renewal failures (row-gone return, sustained-failure escalation, wedged-thread) — a lost lease here means duplicate execution / double-billing, so these deserve a log + metric.
3. **Important** — an inaccurate code comment + misleading README/default interaction around `min(lease, vt)`.
4. **Suggestions/nits** — metrics wiring, connection close, comment precision, return annotation.
5. **Tests** — integration/ordering/edge-case coverage gaps.

Note: one agent's finding that a dead `_renew_client` "never rebuilds / self-heal is false" was **verified and dropped** — `_cursor()` (client.py:219-222) discards a dead owned connection, so `set_vt` does reconnect on the next tick; the comment is accurate.

muhammad-ali-e:
Thanks for the thorough pass — addressed the review in `fbbf03ef5`. Summary mapped to the findings:

**Correctness / safety (the top-3 + high items)**
- **Shared `_renew_client` race (#1) + never-closed (#7):** the renewal thread now owns a **thread-local** connection, built lazily on the first renewal via `_make_renew_client()` and **closed in a `finally`**. No shared state → a join-timeout wedge can't leave the next message's thread reusing a half-used psycopg2 connection.
- **Wedged join silent (#6):** after the bounded join, a still-alive thread is logged at **ERROR** (connection abandoned until process exit).
- **`set_vt`→False unlogged double-run signal (#2):** now logged (the ack runs only after the thread is joined, so a gone row means another consumer reclaimed+acked → this task is double-running). Docstring corrected — "acked by us" was impossible here.
- **Over-broad `except` + no escalation (#3):** narrowed to `CONN_DEAD_ERRORS` (a non-connection error now **propagates** instead of being swallowed as a "self-healing blip" forever), and added escalation — once renewal keeps failing **past `lease_seconds`** the lease is genuinely lost → **ERROR**.
- **BATCH>1 tail double-process (#4):** `BATCH_SIZE` is **forced to 1 when `lease < vt`** (renewal only covers the in-flight message). All shipped workers are already batch=1; this guards the knob. (`lease == vt` keeps batch>1 — no short window.)
- **Silent `min()` clamp (#5/#9):** now **warns** on `lease > vt`; the wrong comment ("renewal fires rarely") is corrected.

**Docs / nits**
- **#10/#11** softened the "never race" docstring (bounded join, late `set_vt` no-ops) and fixed the "lazily created on first task run" → "first renewal" wording. **#12** added `-> Iterator[None]`. **#14** README now says the effective claim window is `min(LEASE, VT)`, the lease is **not** a hard runtime cap (liveness probe is the backstop), plus the batch-forced note; `"PR C"` shorthand → `UN-3695`.

**Tests (#13):** +9 covering loud clamp / tiny-interval floor / non-positive reject / batch-forced / batch-kept-at-`lease==vt` / `LEASE_SECONDS` env wiring / renew+close / no-connection-for-sub-interval-task / row-gone-logs / conn-death-retry-then-escalate / non-connection-propagates / wedged-thread-log / `_handle`-actually-wraps-in-`_lease_renewal`. **87 consumer tests pass.**

**Deferred (1):** lease **metrics counters** (#8) — `ConsumerMetrics` isn't wired to the consumer instance (it's constructed in `run()` around the liveness gauge), so emitting `pg_lease_lost_total` etc. is a separate small observability change rather than squeezing a metrics hook in here. Happy to file it as a fast follow.

The one round-1 finding your verifier already dropped ("dead client never self-heals") stays dropped — `_cursor()` discards a dead owned connection so `set_vt` reconnects next tick.

muhammad-ali-e:
Thanks Greptile (4/5, safe-to-merge) — addressed the one P2 in `e3d72d625`: the startup log in `run()` now shows **`lease=%ss`** alongside `vt=%ss`, so an operator sees the real claim window (e.g. 120s) rather than assuming the 9060s VT is the claim window. `VT_SECONDS` is now the drain/max-runtime bound, not the claim window — the log wording matches that.

PR #2156, opened 2026-07-08
Add MiniMax LLM adapter

octo-patch:
## What

- Adds a branded MiniMax LLM adapter for the compatible chat API.
- Supports MiniMax-M3 and MiniMax-M2.7 with native routing, regional endpoint configuration, thinking controls, an official icon, and focused tests.

## Why

- MiniMax users need a first-class adapter with correct request transformation and usage tracking.

## How

- Registers the adapter and prefixes model IDs with `minimax/` for native routing.
- Defaults to the global API endpoint while documenting the China endpoint as an override.
- Maps the UI thinking toggle to the official `thinking.type` request field and prevents disabling thinking for MiniMax-M2.7.
- Replaces the placeholder icon with the official MiniMax asset.

- No existing adapter behavior is changed. The implementation is isolated to the new MiniMax adapter, its registration, schema, icon, and tests.

## Database Migrations

- None.

## Env Config

- None. Users provide their API key in the adapter configuration and may override the regional API base.

## Relevant Docs

- https://platform.minimax.io/docs/guides/text-generation
- https://platform.minimax.io/docs/guides/pricing-paygo

## Related Issues or PRs

- None.

## Dependencies Versions

- No dependency changes. The implementation uses the repository's pinned `litellm==1.85.1`.

## Notes on Testing

- Focused adapter tests: 19 passed.
- Ruff lint and format checks passed for the changed Python files.
- JSON parsing, Python compilation, secret scanning, and `git diff --check` passed.
- The full adapter test file reached 35 passed; three existing embedding tests could not load the optional system `libmagic` library on this machine.

## Screenshots

- Not applicable.

## Checklist

- [x] I have read and understood the [Contribution Guidelines](https://docs.unstract.com/unstract/contributing/unstract/).


unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/minimax.json
```diff
@@ -0,0 +1,113 @@
...
+      "title": "Timeout",
+      "default": 900,
+      "description": "Timeout in seconds."
+    },
+    "enable_reasoning": {
+      "type": "boolean",
+      "title": "Enable Reasoning",
+      "default": false,
+      "description": "Toggle on for reasoning models to route reasoning_effort via the OpenAI-compatible request body."
+    }
```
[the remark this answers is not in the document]
ritwik-g: Confirmed — this is a real issue, not just a doc mismatch.

The reasoning path in `OpenAICompatibleLLMParameters.validate()` (`base1.py`) does two things when `enable_reasoning` is on: it puts `reasoning_effort` into `extra_body` **and** pops `temperature` (`validated.pop("temperature", None)`). MiniMax's OpenAI-compatible endpoint doesn't recognise `reasoning_effort` — it gates deep thinking via `thinking: {type: "adaptive"}` — so the key is silently ignored while the user's `temperature` is stripped. The toggle activates no reasoning but changes generation behaviour.

Severity is bounded to the opt-in case: it only fires if a user flips the toggle. It won't auto-enable, since `_is_openai_reasoning_model` matches only `o1|o3|o4|gpt-5`, not `MiniMax-M3`.

Fix: dropping the `enable_reasoning` block (and its `allOf` conditional) from `minimax.json` for now, since the shared OpenAI-compatible validate path has no way to emit MiniMax's `thinking` field. A proper `enable_reasoning → thinking:{type:"adaptive"}` mapping would be a follow-up in the validate layer.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -528,6 +529,17 @@ def validate(adapter_metadata: dict[str, "Any"]) -> dict[str, "Any"]:
         )
 
 
+class MiniMaxLLMParameters(OpenAICompatibleLLMParameters):
```
chandrasekharan-zipstack: **Routing through `custom_openai/` silently zeroes out cost tracking for MiniMax.**

`_validate_branded_openai_compatible` delegates to `OpenAICompatibleLLMParameters.validate()`, whose `validate_model()` unconditionally prepends `custom_openai/`. `cost_model` is only populated when the endpoint is OpenAI's own (`base1.py:483`), so it stays unset here, and `audit.py:98` ends up calling `cost_per_token(model="custom_openai/MiniMax-M3")` — which raises, gets swallowed by the bare `except`, and records `cost_in_dollars = 0.0` at `debug` level.

LiteLLM already ships a **native MiniMax provider**, and it prices M3:

```
litellm.get_llm_provider("minimax/MiniMax-M3")  -> ('MiniMax-M3', 'minimax', None, None)

cost_per_token("custom_openai/MiniMax-M3") -> raises        => cost = 0.0
cost_per_token("minimax/MiniMax-M3")       -> $0.30 / $1.20 per 1M tokens
```

So every MiniMax token would bill as $0 in usage tracking, silently.

This is why `OpenRouterLLMParameters` extends `BaseChatCompletionParameters` instead of `OpenAICompatibleLLMParameters` — its own docstring says so: *"Routed through LiteLLM's native `openrouter/` provider so per-token costs resolve and reasoning params map without provider-specific workarounds."* MiniMax should follow **OpenRouter's** template, not NVIDIA Build's.

Worth noting NVIDIA Build's `custom_openai/` routing is *correct* for NVIDIA — LiteLLM prices zero `nvidia_nim/` chat models (3 entries, all rerankers, all `$0.0`), so there's nothing to forfeit there. That reason doesn't carry over to MiniMax.

Bonus: `MinimaxChatConfig.get_api_base()` already returns `https://api.minimax.io/v1`, so going native lets you drop the hardcoded `_MINIMAX_API_BASE` constant entirely.

unstract/sdk1/src/unstract/sdk1/adapters/llm1/minimax.py
```diff
@@ -0,0 +1,45 @@
...
+    def get_description() -> str:
+        return DESCRIPTION
+
+    @staticmethod
+    def get_provider() -> str:
+        return "minimax"
+
+    @staticmethod
+    def get_icon() -> str:
+        return "/icons/adapter-icons/MiniMax.png"
```
chandrasekharan-zipstack, line 41: **The shipped `MiniMax.png` is the OpenRouter logo.**

`frontend/public/icons/adapter-icons/MiniMax.png` is byte-for-byte identical to `OpenRouter.png`:

```
0980b0d3102c84da9d332862c563c44f  MiniMax.png
0980b0d3102c84da9d332862c563c44f  OpenRouter.png
676be9d2fc5cec78d5be80af03eb9fd4  NvidiaBuild.png
```

Both 29953 bytes. Rendering it confirms it's the OpenRouter fork-arrow mark. `NvidiaBuild.png` differs, so this isn't a shared-placeholder convention — it looks like the icon was copied along with the adapter template and never swapped.

Users would see OpenRouter's branding on the MiniMax card in the adapter picker. Needs the real MiniMax logo.

octo-patch:
Thanks for the review. I switched the MiniMax adapter to native routing, mapped the UI toggle to `thinking.type`, covered MiniMax-M3 and MiniMax-M2.7, documented both regional API endpoints, and replaced the icon with the official asset. I ran 19 focused tests plus Ruff and JSON validation.

octo-patch:
This work continues in #2166, rebuilt from the latest `main` with a clean commit history. The replacement contains the same scoped MiniMax adapter changes and focused checks.

Issue #2014, opened 2026-06-05
fix: [ISSUE] Bundled qdrant server (v1.8.3) incompatible with qdrant-client 1.16, vector db test connection fails with 404 error


amanattrish:
## Describe the bug
On a fresh setup (`./run-platform.sh`), adding the bundled Qdrant as a vector database during onboarding fails. "Test Connection" returns: "Error testing '<vector-db-name>'. Unexpected Response: 404 (Not Found)..."

## To reproduce
1. Run `./run-platform.sh` and log in.
2. Onboarding >> Connect a Vector Database >> Qdrant.
3. Add Name of vector DB, URL: `http://unstract-vector-db:6333`, no API key.
4. Click Test Connection.

## Expected behavior
It should connect without any problem.

## Environment details
- Version v0.173.0 of unstarct
- OS: Ubuntu

## Screenshots

<img width="1077" height="458" alt="Image" src="https://github.com/user-attachments/assets/4dfe81b0-94f0-4074-b0bb-883b91c855dd" />

amanattrish:
Root Cause:

The Python client and the bundled server are far apart in version. Backend logs:
qdrant.py:67: UserWarning: Qdrant client version 1.16.2 is incompatible with server version 1.8.3. Major versions should match and minor version difference must not exceed 1.
...
Error occured while testing adapter Unexpected Response: 404 (Not Found)


Network connectivity is fine (curl to the container returns the version JSON). The 1.16 client calls an API path the 1.8.3 server does not have, so it gets a 404


The two pins have drifted:
  - `docker/docker-compose-dev-essentials.yaml` pins `qdrant/qdrant:v1.8.3`, last touched in #185 (Mar 2024)
  - `unstract/sdk1/pyproject.toml` pins `qdrant-client>=1.16.0,<1.17.0`, bumped in #1845 (Mar 2026)

amanattrish:
Suggested fix:
Bump the Qdrant server image to a 1.16.x tag to match the client. `v1.16.1` lines up with the installed client version.

chandrasekharan-zipstack:
@amanattrish apologies for getting to this late and thanks for taking the time to raise this issue and a PR for it.

This was fixed in `v0.174.0` by #2046 

<img width="2252" height="714" alt="Image" src="https://github.com/user-attachments/assets/a72c47bf-398c-47ee-8bdd-94d6a4d0c5ed" />

PR #2167, opened 2026-07-12
Polish docs and comments in unstract (#2014)

bglglzd:
This is a focused change for the cited issue with minimal side effects.

Related to #2014.

### DIFF
--- a/backend/file_management/api_doc.md
+++ b/backend/file_management/api_doc.md
@@ -36,5 +36,5 @@ Sample URL:
     "file": <File> "<multiple files>",
     "connector_id": "<Connector Id>",
     "path": "<File location>",
-}   
+}
 ```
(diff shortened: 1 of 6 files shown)


jaseemjaskp:
Thanks for the contribution, but this PR does not address #2014. That issue is about the bundled Qdrant server (v1.8.3) being incompatible with qdrant-client 1.16, which causes the vector DB "Test Connection" to fail with a 404 — a code/dependency fix, not a documentation issue.
This PR only removes trailing whitespace from a few Markdown files and makes no functional changes. It also breaks intentional Markdown hard breaks in docs/ARCHITECTURE.md (as flagged in review), so it's a net negative rather than an improvement.
Closing this PR as it doesn't resolve the linked issue and adds no meaningful value. Please avoid linking unrelated issues to trivial formatting changes.

PR #2171, opened 2026-07-14
UN-3445 [FIX] workflow_execution — partial index on active executions (cut per-execution history scan)

muhammad-ali-e:
## What

Adds a **partial index** on `workflow_execution` over active (non-terminal) rows only, to speed up the hot "is there an in-flight execution of this workflow?" lookup.

```sql
CREATE INDEX CONCURRENTLY we_active_by_workflow_idx
  ON workflow_execution (workflow_id)
  WHERE status NOT IN ('COMPLETED', 'STOPPED', 'ERROR');
```

## Why

`_get_active_workflow_executions` (backend `endpoint_v2/source.py`) and the worker's `get_workflow_executions_by_status` run, per file / execution:

```sql
WHERE workflow_id = … AND status IN ('PENDING', 'EXECUTING')
```

The existing `(workflow_id, created_at)` index returns **all** of a workflow's executions (99%+ terminal history) and then filters down to the 1–2 active rows. That cost grows **unbounded with history**:

- Measured **~44× amplification** on a 35k-execution workflow (index reads 34,913 rows to return the active ones); production workflows reach 100k+.
- `EXPLAIN ANALYZE`: **21.0 ms → 1.4 ms** per call with the partial index.
- Under load this surfaces as DB CPU — observed pegging a 4-vCPU Postgres at ~99% at 800 concurrent users, with the queue-wait stage ballooning.

The partial index makes this lookup's cost **independent of workflow history**.

## How

- **Partial** — indexes only non-terminal rows (a few thousand regardless of table size), so it stays tiny (~2 MB on a 3M-row table vs ~20 MB for a full `(workflow_id, status)` composite) and never includes the unbounded terminal history.
- **Complement predicate** (`NOT IN (terminal)`) rather than a frozen `IN ('PENDING','EXECUTING')` list: a query for any subset of active statuses still provably implies `NOT IN (terminal)`, so the planner keeps using the index if a **new active status** (e.g. a future `PAUSE`) is introduced. A frozen active-list index would silently stop being used in that case. Only a new **terminal** status requires touching the predicate.
- **`CONCURRENTLY` + `atomic = False`** — the build takes no write lock, so live execution traffic is not blocked on the large `workflow_execution` table. `SeparateDatabaseAndState` (with `RunSQL` + `IF NOT EXISTS`, mirroring migration 0019) keeps Django's model state in sync with `Meta.indexes`.
- **No query/code changes** — the index is transparent to the ORM; the planner picks it up automatically.

## Testing

Dev-tested against a local Postgres:

- `migrate` applies cleanly; index is `indisvalid = true`.
- Planner uses it for `status IN ('EXECUTING', 'PENDING')` (`Index Scan using we_active_by_workflow_idx`).
- Still used for a hypothetical future `status IN (…, 'PAUSE')` query (complement future-proofing).
- Reverse migration drops it; re-apply recreates it.
- `makemigrations --check` → no drift; `sqlmigrate` emits the expected `CREATE INDEX CONCURRENTLY …`.

## Deployment notes

`workflow_execution` is large in production (multi-million rows). `CREATE INDEX CONCURRENTLY` scans the table and can run for minutes — long enough to time out a deploy's `migrate` step. **Recommended: build the index out of band first** (run the `CREATE INDEX CONCURRENTLY IF NOT EXISTS …` manually before the deploy); the migration then no-ops via `IF NOT EXISTS`.

**Recovery:** an interrupted `CREATE INDEX CONCURRENTLY` leaves an `INVALID` index; `IF NOT EXISTS` will not rebuild over it, so recover with `DROP INDEX CONCURRENTLY IF EXISTS we_active_by_workflow_idx;` then re-run.

### DIFF
--- a/backend/workflow_manager/workflow_v2/migrations/0023_workflowexecution_active_by_workflow_idx.py
+++ b/backend/workflow_manager/workflow_v2/migrations/0023_workflowexecution_active_by_workflow_idx.py
@@ -1,0 +1,140 @@
+"""Add a partial index over ACTIVE workflow executions.
+
+Speeds up the hot "is there an in-flight execution of this workflow?" lookup
+(``_get_active_workflow_executions`` in ``endpoint_v2/source.py`` and the worker's
+``get_workflow_executions_by_status``), which runs
+``WHERE workflow_id = … AND status IN ('PENDING', 'EXECUTING')`` per file /
+execution. The existing ``(workflow_id, created_at)`` index returns *all* of a
+workflow's executions (99%+ terminal history) and then filters down to the 1-2
+active rows — a cost that grows without bound as a workflow accumulates history.
+
+Measured once at authoring on a 35k-execution workflow (figures illustrative,
+as-measured): ~44x fewer rows scanned (the index walks the active rows, not all
+34,913) and 21ms -> 1.4ms latency per call. Production workflows reach 100k+.
+
+Design
+------
+* PARTIAL — the index only covers non-terminal rows, so it stays tiny (a few
+  thousand active rows regardless of table size) and its cost is independent of
+  history. On a ~3M-row table it is ~2 MB vs ~20 MB for a full
+  ``(workflow_id, status)`` composite.
+* COMPLEMENT PREDICATE (``status NOT IN (terminal)``) rather than a frozen
+  ``status IN ('PENDING', 'EXECUTING')`` list. This is an *index-level* property:
+  a query for any subset of non-terminal statuses still provably implies
+  ``NOT IN (terminal)``, so the index stays usable if a new *active* status
+  (e.g. a future PAUSE) is added. It does NOT make a new active status work
+  end-to-end — the callers use a frozen positive ``status IN ('PENDING',
+  'EXECUTING')`` list, so such a row is *indexed* but not *returned* until every
+  caller's list is updated too. Only a new *terminal* status requires touching
+  this predicate.
+* CONCURRENTLY + ``atomic = False`` — builds without a write lock, so live
+  execution traffic is not blocked. ``workflow_execution`` is a large
+  (multi-million-row) table in production; a plain ``AddIndex`` would hold a SHARE
+  lock and block writes for the full build.
+* INVALID-INDEX GUARD — ``IF NOT EXISTS`` would silently no-op over a leftover
+  INVALID index from a prior interrupted CONCURRENTLY build, and Django would then
+  record this migration as applied while the index is physically unusable
+  (never read, only write overhead). The second statement RAISEs in that case so
+  the failure is loud instead of green-but-broken.
+
+Deployment
+----------
+``CREATE INDEX CONCURRENTLY`` scans the whole table (twice) and can run for several
+minutes on a large table — long enough to time out a deploy's ``migrate`` step.
+Prefer building the index OUT OF BAND *before* the deploy; the migration then
+becomes a no-op via ``IF NOT EXISTS``::
+
+    CREATE INDEX CONCURRENTLY IF NOT EXISTS we_active_by_workflow_idx
+      ON workflow_execution (workflow_id)
+      WHERE status NOT IN ('COMPLETED', 'STOPPED', 'ERROR');
+
+After building, verify the planner actually chooses it — the partial-index proof
+depends on the ``status`` values reaching the planner as constants, so re-verify
+if the DB driver is ever changed (e.g. a psycopg3 server-side-binding move)::
+
+    EXPLAIN SELECT * FROM workflow_execution
+      WHERE workflow_id = '…' AND status IN ('PENDING', 'EXECUTING');
+    -- expect: Index Scan using we_active_by_workflow_idx
+
+Recovery
+--------
+An interrupted ``CREATE INDEX CONCURRENTLY`` leaves an INVALID index that still
+adds write overhead but is not used for reads. ``IF NOT EXISTS`` will NOT rebuild
+over it (and the guard below RAISEs on it), so recover by dropping it first, then
+re-running the migration::
+
+    DROP INDEX CONCURRENTLY IF EXISTS we_active_by_workflow_idx;
+"""
+
+from django.db import migrations, models
+from django.db.models import Q
+
+INDEX_NAME = "we_active_by_workflow_idx"
+
+# Terminal ExecutionStatus values (COMPLETED, STOPPED, ERROR). Kept as literals —
+# migrations must not import app enums, so they stay frozen if the enum changes.
+# The model-side copy (WorkflowExecution.Meta.indexes) is asserted against
+# ExecutionStatus.terminal_values() by tests/test_active_execution_index.py so it
+# cannot silently drift. Update ONLY for a new *terminal* status.
+TERMINAL_STATUSES = ["COMPLETED", "STOPPED", "ERROR"]
+
+# Fail loudly if a prior interrupted CONCURRENTLY build left an INVALID index of
+# this name: the IF NOT EXISTS above would otherwise silently keep it and Django
+# would record this migration as applied while the index is physically unusable.
+_ASSERT_INDEX_VALID = f"""
+DO $$
+BEGIN
+    IF EXISTS (
+        SELECT 1 FROM pg_class c
+        JOIN pg_index i ON i.indexrelid = c.oid
+        WHERE c.relname = '{INDEX_NAME}' AND NOT i.indisvalid
+    ) THEN
+        RAISE EXCEPTION 'Index {INDEX_NAME} exists but is INVALID (a prior CREATE INDEX CONCURRENTLY was interrupted). Drop it and re-run: DROP INDEX CONCURRENTLY IF EXISTS {INDEX_NAME};';
+    END IF;
+END
+$$;
+"""
+
+
+class Migration(migrations.Migration):
+    # CREATE / DROP INDEX CONCURRENTLY cannot run inside a transaction block.
+    atomic = False
+
+    dependencies = [
+        ("workflow_v2", "0022_merge_20260626_1131"),
+    ]
+
+    operations = [
+        migrations.SeparateDatabaseAndState(
+            database_operations=[
+                # Build the real index safely: lock-free (CONCURRENTLY), idempotent
+                # (IF [NOT] EXISTS), a no-op if it was created out of band first.
+                migrations.RunSQL(
+                    sql=(
+                        f"CREATE INDEX CONCURRENTLY IF NOT EXISTS {INDEX_NAME} "
+                        "ON workflow_execution (workflow_id) "
+                        "WHERE status NOT IN ('COMPLETED', 'STOPPED', 'ERROR');"
+                    ),
+                    reverse_sql=f"DROP INDEX CONCURRENTLY IF EXISTS {INDEX_NAME};",
+                ),
+                # Guard: turn a masked INVALID index (from a prior failed build)
+                # into a loud failure instead of a green-but-unused index.
+                migrations.RunSQL(
+                    sql=_ASSERT_INDEX_VALID,
+                    reverse_sql=migrations.RunSQL.noop,
+                ),
+            ],
+            # Keep Django's model state aware of the index (mirrors Meta.indexes on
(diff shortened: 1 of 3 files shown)


muhammad-ali-e (commented):
Automated review (PR Review Toolkit: code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier).

The index design itself is sound — partial complement-predicate, `CONCURRENTLY` + `atomic=False`, `SeparateDatabaseAndState`, model/DB state in sync — and I confirmed the index is genuinely exercised by every active-execution query path (`_get_active_workflow_executions`, the worker's internal viewset).

The one blocking defect is the **broken migration dependency on the nonexistent `0022_merge_20260626_1131`** (should be `0021_alter_workflow_organization`) — already reported by greptile on line 68, so not repeated here. `0022_merge` lives only on `feat/UN-3445-pg-queue-integration`, not on this branch or `main`, so `migrate` / `makemigrations --check` will raise `NodeNotFoundError`.

The inline comments below are the remaining, non-duplicate findings — all maintainability/robustness (P2) or advisory (P3); none block beyond the dependency fix.

backend/workflow_manager/workflow_v2/models/execution.py
```diff
@@ -214,6 +214,18 @@ class Meta:
         indexes = [
             models.Index(fields=["workflow_id", "-created_at"]),
             models.Index(fields=["pipeline_id", "-created_at"]),
+            # Partial index over ACTIVE (non-terminal) executions only — see
+            # migration 0023. Keeps the hot "is there an in-flight execution of
+            # this workflow?" lookup (WHERE workflow_id=… AND status IN active)
+            # O(active) instead of O(all-history-for-workflow). The predicate is
+            # the COMPLEMENT of the terminal set, so a query for any subset of the
+            # active states still uses the index and a new *active* status is
+            # auto-covered; only a new *terminal* status needs this list updated.
```
muhammad-ali-e: **P2 — “auto-covered” is misleading; a new *active* status is NOT covered end-to-end.** This is true only at the *index* level: the partial predicate would still be usable. But both real callers use a **frozen positive list** `status IN ('PENDING','EXECUTING')` — `_get_active_workflow_executions` (`endpoint_v2/source.py:586`) and the worker (`workers/shared/api/internal_client.py:399`, `workers/shared/workflow/connectors/source.py:377`). So if someone adds a future `PAUSE` active status trusting this comment, the row is *indexed* but never *returned* by the in-flight lookup → duplicate file processing, silently. Same wording in the migration docstring (lines 19-25). Suggest softening to: “the index auto-covers a new active status, but every caller’s active-status list must still be updated.”

backend/workflow_manager/workflow_v2/models/execution.py
```diff
@@ -214,6 +214,18 @@ class Meta:
...
+            models.Index(
+                fields=["workflow_id"],
+                name="we_active_by_workflow_idx",
+                condition=~Q(status__in=["COMPLETED", "STOPPED", "ERROR"]),
```
muhammad-ali-e, line 237: **P2 — hardcoded terminal set bypasses the canonical `ExecutionStatus.terminal_values()`.** This file already imports `ExecutionStatus` (line 21), and a canonical helper exists — `ExecutionStatus.terminal_values()` (`unstract/core/.../data_models.py:581`) returns `frozenset({COMPLETED, STOPPED, ERROR})`, and its own comment says it exists “so every terminal check … and ORM `status__in` filters … stays in sync.” This inline `["COMPLETED","STOPPED","ERROR"]` is a 3rd hand-maintained copy (model + migration `state_operations` + migration RunSQL literal).

Caveat: a naive swap breaks `makemigrations --check` — `terminal_values()` is an unordered `frozenset`, so the deconstructed condition won’t match the migration’s frozen list. To use the helper you must produce an identical deconstruction (e.g. a fixed `sorted(...)`) **and** regenerate the migration `state_operations` to match. If that’s not worth it, add a unit test asserting `set(model_index_predicate) == ExecutionStatus.terminal_values()` so the literal can’t silently drift from the enum. The migration RunSQL literal is legitimately frozen and should stay a literal.

backend/workflow_manager/workflow_v2/migrations/0023_workflowexecution_active_by_workflow_idx.py
```diff
@@ -0,0 +1,98 @@
...
+    ]
+
+    operations = [
+        migrations.SeparateDatabaseAndState(
+            # Build the real index safely: lock-free (CONCURRENTLY), idempotent
+            # (IF [NOT] EXISTS), and a no-op if it was created out of band first.
+            database_operations=[
+                migrations.RunSQL(
+                    sql=(
+                        f"CREATE INDEX CONCURRENTLY IF NOT EXISTS {INDEX_NAME} "
```
muhammad-ali-e, line 114: **P2 — `IF NOT EXISTS` silently no-ops over a leftover INVALID index.** This is exactly the scenario the Recovery docstring (lines 42-51) describes, but nothing in the code defends against it. If a prior `CREATE INDEX CONCURRENTLY` was interrupted (deploy `statement_timeout`, OOM, SIGTERM during a rolling deploy), Postgres leaves an **INVALID** index of the same name. On the next `migrate`, `IF NOT EXISTS` matches on name and no-ops, Django records 0023 as *applied* and the model state as having the index — while physically it is invalid: never used for reads (the O(all-history) scan this PR set out to kill persists) yet still carrying write overhead. `migrate` exits 0; the only way to notice is manually querying `pg_index.indisvalid`. With `atomic=False` there is no rollback to lean on, so this guard is the only safety net. Suggest a post-create assertion (a `DO $$ ... $$` block that `RAISE`s if the index exists but `NOT indisvalid`) so a masked prior failure fails loudly instead of shipping green-but-broken.

backend/workflow_manager/workflow_v2/migrations/0023_workflowexecution_active_by_workflow_idx.py
```diff
@@ -0,0 +1,98 @@
+"""Add a partial index over ACTIVE workflow executions.
+
+Speeds up the hot "is there an in-flight execution of this workflow?" lookup
+(``_get_active_workflow_executions`` in ``endpoint_v2/source.py`` and the worker's
+``get_workflow_executions_by_status``), which runs
+``WHERE workflow_id = … AND status IN ('PENDING', 'EXECUTING')`` on every file /
+execution. The existing ``(workflow_id, created_at)`` index returns *all* of a
+workflow's executions (99%+ terminal history) and then filters down to the 1-2
+active rows — a cost that grows without bound as a workflow accumulates history
+(measured ~44x amplification, 21ms -> 1.4ms per call on a 35k-execution workflow;
```
muhammad-ali-e: **P3 — comment: `~44x amplification` reads as the same metric as `21ms -> 1.4ms` (~15x).** These are two different quantities placed side by side (44× is presumably *rows scanned*, 15× is *latency*), so a reader takes them as one and finds them contradictory. Suggest labeling them, e.g. “~44× fewer rows scanned; 21ms -> 1.4ms latency (figures illustrative, measured once at authoring).” Since a migration is an immutable historical artifact, framing the numbers as as-measured avoids future confusion.

backend/workflow_manager/workflow_v2/migrations/0023_workflowexecution_active_by_workflow_idx.py
```diff
@@ -0,0 +1,98 @@
...
+
+Design
+------
+* PARTIAL — the index only covers non-terminal rows, so it stays tiny (a few
+  thousand active rows regardless of table size) and its cost is independent of
+  history. On a ~3M-row table it is ~2 MB vs ~20 MB for a full
+  ``(workflow_id, status)`` composite.
+* COMPLEMENT PREDICATE (``status NOT IN (terminal)``) rather than a frozen
+  ``status IN ('PENDING', 'EXECUTING')`` list: a query for any *subset* of the
+  active states still provably implies ``NOT IN (terminal)``, so the planner keeps
```
muhammad-ali-e: **P3 — advisory: the complement-predicate proof depends on plan-time constant statuses (driver-sensitive).** For Postgres to use this partial index, the planner must prove `status IN (active)` ⇒ `status NOT IN (terminal)`; that proof only fires when the `status` values reach the planner as `Const`s. Today that holds because the project pins `psycopg2-binary` (`backend/pyproject.toml`), which interpolates params client-side, so the server sees literals. Note for the future: a move to psycopg3 in server-side-binding mode would send statuses as bind `Param`s, the implication proof can fail, and the planner would silently fall back to `(workflow_id, -created_at)` — making this index dead write-overhead. No change needed now; worth adding a one-line `EXPLAIN` check to the deploy verification (confirm `we_active_by_workflow_idx` is actually chosen) and re-verifying if the DB driver is ever upgraded.

muhammad-ali-e:
Addressed all review comments in `82aa895ec`:

**P1 — migration graph missing parent (greptile).** Correct — `0022_merge` and `0020_queue_message_id` live only on `feat/UN-3445-pg-queue-integration`, not on `main` (this PR's base, whose leaf is `0021_alter_workflow_organization`). Fixed the dependency to `0021` and renumbered the migration `0023 → 0022`. Verified against `main`: leaf is `0021`; the graph now loads.

**P2 — "auto-covered" is misleading.** Softened in both the model comment and the migration docstring: the *index* stays usable for a new active status, but the callers use a frozen `status IN ('PENDING','EXECUTING')` list, so a new active status is *indexed yet not returned* until every caller's list is updated too.

**P2 — hardcoded terminal set can drift.** Added `tests/test_active_execution_index.py` asserting the index's excluded set equals `{all ExecutionStatus values} − {PENDING, EXECUTING}`, so a new enum status fails the test until it is consciously classified. Note: `ExecutionStatus.terminal_values()` is **feat-branch-only** (not on `main`), so the test keys off the enum *members* (which exist on main) rather than that helper — same drift protection, base-compatible.

**P2 — `IF NOT EXISTS` no-ops over a leftover INVALID index.** Added a second `RunSQL` (`DO $$ … $$` block) that `RAISE`s if an index of this name exists but is `NOT indisvalid`, so a masked prior failure fails loudly instead of shipping green-but-unused. Tested the guard SQL against a valid index (no false raise).

**P3 — `44x` vs `21ms→1.4ms` read as contradictory.** Relabeled as two distinct as-measured quantities: `~44x fewer rows scanned` and `21ms -> 1.4ms latency`.

**P3 — complement-predicate proof is driver-sensitive (psycopg2 vs psycopg3).** Added an `EXPLAIN` verification step to the Deployment notes (confirm `we_active_by_workflow_idx` is chosen; re-verify if the DB driver changes).

Note on local validation: this branch's workspace can't complete a full `django.setup()` (an unrelated `unstract.sdk1.prompt` import in the `simple_prompt_studio` plugin), so `makemigrations --check` and the pytest run happen in CI. Locally verified: `py_compile` on all files, the guard SQL against the DB, the drift-test set math, and the `0021` dependency resolves.

muhammad-ali-e:
**Retargeted this PR from `main` to `feat/UN-3445-pg-queue-integration`** (where the rest of the epic lives), and rebased onto it. This resolves the earlier migration-dependency question cleanly:

- **Dependency** is now `0022_merge_20260626_1131` (the leaf on this branch), and the migration is numbered **`0023`**. No later `makemigrations --merge` is needed — it's a linear chain here.
- **Drift test** now uses the canonical `ExecutionStatus.terminal_values()` (which exists on this branch), per the earlier review suggestion, instead of keying off enum members.

Full validation now passes locally on this branch (the earlier `sdk1.prompt` env block was specific to the `main`-based checkout):
- `makemigrations workflow_v2 --check` → **No changes detected** (graph loads, `0022_merge` dependency resolves, no drift)
- `sqlmigrate workflow_v2 0023` → emits the `CREATE INDEX CONCURRENTLY … WHERE status NOT IN (...)` plus the INVALID-index `DO` guard
- `pytest tests/test_active_execution_index.py` → **2 passed** (index is a negated partial; excluded set == `ExecutionStatus.terminal_values()`)

All prior review points (INVALID-index guard, softened "auto-covered" wording, 44×-rows-vs-latency labeling, EXPLAIN deploy-verify note) are retained.

PR #2172, opened 2026-07-14
UN-3445 [FIX] PG queue — finalization strand: race guard + PG re-raise + reaper self-heal + N+1 dedup

muhammad-ali-e:
## Problem

Under the PG path at high concurrency, executions strand **permanently in `EXECUTING`**: every file execution reaches `COMPLETED` (work done) but the parent `workflow_execution` is never marked terminal, with `successful_files`/`failed_files` left `NULL`. On a load-test workflow, **783 of 860** stuck rows were PG (64 created in a 6-hour window); the rest were old Celery cruft.

**Root cause (log + DB confirmed):** a **lost-update race**. The callback's terminal write *succeeds* (backend returns 200, logs `COMPLETED`), but a concurrent **unlocked read-modify-write** on the same status column commits *after* it and clobbers it back to `EXECUTING` (the NULL counters are the tell — written to 1, then reverted). A **swallowed** finalization error is a second, latent path to the same strand. Stranded execs then fuel an **O(active) N+1** in the dedup check that collapsed 800u throughput (6 → 2 rps).

Customer impact: a client polling the API-deployment status endpoint (which keys off `execution_status`) sees `EXECUTING` forever even though the result is ready.

## Fixes (all PG-scoped; the Celery flow is unchanged)

**L1 — race guard** (`internal_views.update_status`): flag-gated on `pg_queue_enabled` (per-org, like transport dispatch). Under the flag: `SELECT FOR UPDATE` the row and enforce **terminal one-way** — a non-terminal write can no longer overwrite an already-terminal execution.

**L2 — durable finalization** (`callback/_update_execution_status_unified`): **re-raise** the finalization error on the PG path (`is_pg`) instead of swallowing, so the message vt-redelivers and, if it keeps failing, poison-drops to a terminal `ERROR`. Idempotency-guarded by the existing `_callback_already_ran`. Celery keeps the legacy swallow.

**L3 — retry hardening** (`base_client`): equal-jitter backoff on 5xx/timeout retries, so concurrent workers stop retrying in synchronized waves against a saturated backend.

**L4 — reaper self-heal** (opt-in `WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED`): the leader-elected reaper triggers a new backend endpoint `recover_stuck_pg_executions` that finds PG executions (**`queue_message_id` set** — Celery rows are never touched) stuck past a long window, **recomputes the correct terminal status from their files** (or `ERROR`s a never-dispatched, file-less one), and finalizes them. The barrier self-deletes at fan-in, so the existing PG-table sweeps cannot see these — this is the guard for that gap. Reaper reads business state via the internal API only (never direct-DBs `workflow_execution`).

**N+1 — batched dedup** (`active_file_executions` endpoint + worker rewire): one indexed join replaces the workers' per-active-execution dedup loop (`_get_existing_file_executions`); the worker falls back to the legacy loop if the endpoint is unavailable.

## Tests

- `workers/tests/test_pg_finalization_fixes.py` — L2 (PG re-raise / Celery swallow / success), L3 (jitter band / non-constant / zero). **6 pass.**
- `backend/workflow_manager/execution/tests/test_pg_finalization_fixes.py` — L1 guard (reject non-terminal-over-terminal / flag-off legacy / terminal-write-allowed), L4 (recompute→COMPLETED / →ERROR / file-less→ERROR / **Celery-never-scanned** / still-processing-skipped / recent-not-recovered), N+1 mapping. **10 pass** (Django `TestCase`, run via `manage.py test`).

## Deploy / verification notes

- L4 and the N+1 query use the `workflow_execution(workflow_id) WHERE status NOT IN (terminal)` partial index (already merged on this base) — EXPLAIN-verified **26 ms / 29 ms** on a 3M-row table.
- Full E2E (dispatch → worker → callback → finalize, plus the reaper sweep) can't be run outside a live worker fleet; plan is to merge, deploy to integration, and validate with a 400u/800u load test there (expect: 0 executions left EXECUTING-with-all-files-terminal after drain, and the dedup logged as a single batched call).
- Rollback: L1 gated by `pg_queue_enabled`; L4 gated by its env flag (default off).

### DIFF
--- a/backend/workflow_manager/internal_views.py
+++ b/backend/workflow_manager/internal_views.py
@@ -589,7 +589,6 @@ class WorkflowExecutionInternalViewSet(viewsets.ReadOnlyModelViewSet):
         """
         from datetime import timedelta
 
-        from django.db.models import Count, Q
         from django.utils import timezone
 
         from workflow_manager.workflow_v2.enums import ExecutionStatus
@@ -623,92 +622,95 @@ class WorkflowExecutionInternalViewSet(viewsets.ReadOnlyModelViewSet):
             .values_list("id", flat=True)[:limit]
         )
 
-        recovered = 0
-        skipped = 0
-        failed_recoveries = 0
+        outcomes = {"recovered": 0, "skipped": 0, "failed": 0}
         for exec_id in stuck_ids:
-            try:
-                with transaction.atomic():
-                    execution = WorkflowExecution.objects.select_for_update().get(
-                        pk=exec_id
-                    )
-                    # Re-check under lock: skip if it reached terminal or was just
-                    # touched (no longer stuck) — never race a live execution.
-                    if execution.status in terminal or execution.modified_at >= cutoff:
-                        skipped += 1
-                        continue
-                    # One conditional aggregate instead of three COUNT round-trips.
-                    counts = WorkflowFileExecution.objects.filter(
-                        workflow_execution=execution
-                    ).aggregate(
-                        total=Count("id"),
-                        terminal=Count("id", filter=Q(status__in=terminal)),
-                        errored=Count("id", filter=Q(status=ExecutionStatus.ERROR.value)),
-                        stopped=Count(
-                            "id", filter=Q(status=ExecutionStatus.STOPPED.value)
-                        ),
-                        completed=Count(
-                            "id", filter=Q(status=ExecutionStatus.COMPLETED.value)
-                        ),
-                    )
-                    total = counts["total"]
-                    if total == 0 or counts["terminal"] < total:
-                        # total==0: a valid PG message may still be QUEUED (a
-                        # backlog/outage can outlast stuck_seconds), so failing it
-                        # would wrongly terminalize a live execution the one-way guard
-                        # then blocks from recovery — file-less recovery needs a
-                        # queue-state check (reaper-side, follow-up). terminal < total:
-                        # a file is still non-terminal → genuinely processing. Skip.
-                        skipped += 1
-                        continue
-                    # Recompute the correct terminal status: ERROR if any file
-                    # errored, else STOPPED if any was stopped (a cancelled run must
-                    # not be turned into ERROR), else COMPLETED.
-                    if counts["errored"]:
-                        computed = ExecutionStatus.ERROR
-                    elif counts["stopped"]:
-                        computed = ExecutionStatus.STOPPED
-                    else:
-                        computed = ExecutionStatus.COMPLETED
-                    execution.update_execution(status=computed)
-                    self._update_file_aggregates(
-                        execution,
-                        {
-                            "total_files": total,
-                            "successful_files": counts["completed"],
-                            "failed_files": counts["errored"],
-                        },
-                    )
-                    logger.warning(
-                        "Reaper safety-net: finalized stranded PG execution %s to %s "
-                        "(files %s ok / %s errored / %s stopped)",
-                        exec_id,
-                        computed.value,
-                        counts["completed"],
-                        counts["errored"],
-                        counts["stopped"],
-                    )
-                    recovered += 1
-            except Exception:
-                failed_recoveries += 1
-                logger.exception(
-                    "recover_stuck_pg_executions: failed to recover %s", exec_id
-                )
+            outcomes[self._recover_one_stuck_pg_execution(exec_id, cutoff, terminal)] += 1
         logger.info(
             "recover_stuck_pg_executions: scanned=%s recovered=%s skipped=%s failed=%s",
             len(stuck_ids),
-            recovered,
-            skipped,
-            failed_recoveries,
+            outcomes["recovered"],
+            outcomes["skipped"],
+            outcomes["failed"],
         )
         return Response(
             {
-                "recovered": recovered,
-                "skipped": skipped,
+                "recovered": outcomes["recovered"],
+                "skipped": outcomes["skipped"],
                 "scanned": len(stuck_ids),
-                "failed": failed_recoveries,
+                "failed": outcomes["failed"],
             }
         )
+
+    def _recover_one_stuck_pg_execution(self, exec_id, cutoff, terminal) -> str:
+        """Finalize one stuck PG execution under a row lock.
+
+        Returns the outcome — ``"recovered"`` | ``"skipped"`` | ``"failed"``.
+        Extracted from :meth:`recover_stuck_pg_executions` to keep that action's
+        cognitive complexity low; see it for the full rationale.
+        """
+        from django.db.models import Count, Q
+
+        from workflow_manager.workflow_v2.enums import ExecutionStatus
+
+        try:
+            with transaction.atomic():
+                execution = WorkflowExecution.objects.select_for_update().get(pk=exec_id)
+                # Re-check under lock: skip if it reached terminal or was just
+                # touched (no longer stuck) — never race a live execution.
+                if execution.status in terminal or execution.modified_at >= cutoff:
+                    return "skipped"
+                # One conditional aggregate instead of three COUNT round-trips.
+                counts = WorkflowFileExecution.objects.filter(
+                    workflow_execution=execution
+                ).aggregate(
+                    total=Count("id"),
+                    terminal=Count("id", filter=Q(status__in=terminal)),
+                    errored=Count("id", filter=Q(status=ExecutionStatus.ERROR.value)),
+                    stopped=Count("id", filter=Q(status=ExecutionStatus.STOPPED.value)),
+                    completed=Count(
+                        "id", filter=Q(status=ExecutionStatus.COMPLETED.value)
+                    ),
+                )
+                total = counts["total"]
+                if total == 0 or counts["terminal"] < total:
+                    # total==0: a valid PG message may still be QUEUED (a
+                    # backlog/outage can outlast stuck_seconds), so failing it would
+                    # wrongly terminalize a live execution the one-way guard then
+                    # blocks from recovery — file-less recovery needs a queue-state
+                    # check (reaper-side, follow-up). terminal < total: a file is
+                    # still non-terminal → genuinely processing. Skip either way.
+                    return "skipped"
+                # Recompute the correct terminal status: ERROR if any file errored,
+                # else STOPPED if any was stopped (a cancelled run must not be turned
+                # into ERROR), else COMPLETED.
+                if counts["errored"]:
+                    computed = ExecutionStatus.ERROR
+                elif counts["stopped"]:
+                    computed = ExecutionStatus.STOPPED
+                else:
+                    computed = ExecutionStatus.COMPLETED
+                execution.update_execution(status=computed)
+                self._update_file_aggregates(
+                    execution,
+                    {
+                        "total_files": total,
+                        "successful_files": counts["completed"],
+                        "failed_files": counts["errored"],
+                    },
+                )
+                logger.warning(
+                    "Reaper safety-net: finalized stranded PG execution %s to %s "
+                    "(files %s ok / %s errored / %s stopped)",
+                    exec_id,
+                    computed.value,
+                    counts["completed"],
+                    counts["errored"],
+                    counts["stopped"],
+                )
+                return "recovered"
+        except Exception:
+            logger.exception("recover_stuck_pg_executions: failed to recover %s", exec_id)
+            return "failed"
 
     @staticmethod
     def _cascade_terminal_files(execution, status_enum, error_message) -> None:
--- a/backend/workflow_manager/internal_views.py
+++ b/backend/workflow_manager/internal_views.py
@@ -510,35 +510,44 @@ class WorkflowExecutionInternalViewSet(viewsets.ReadOnlyModelViewSet):
                 # PG execution is still finalizing (per the "transport carried on the
                 # row, never re-read from Flipt" rule).
                 #
-                # Guard COMPLETED specifically — it is the codebase's only truly-final
-                # status (set exclusively by a successful callback; a redelivered
-                # callback on a COMPLETED row is skipped, see _callback_already_ran).
-                # Under the PG path's high concurrency ~15 call sites write status via
-                # unlocked read-modify-write, so a stale writer can commit AFTER the
-                # callback's COMPLETED and clobber it (lost update → execution stranded
-                # in EXECUTING, or a confusing success→failed flip). Lock the row and
-                # refuse to overwrite COMPLETED with anything else.
+                # Protect COMPLETED and STOPPED — the two statuses that assert a
+                # deliberate final outcome. COMPLETED is set exclusively by a
+                # successful callback (its redelivery is skipped, see
+                # _callback_already_ran); STOPPED records an explicit user/operator
+                # stop. Under the PG path's high concurrency ~15 call sites write
+                # status via unlocked read-modify-write, so a stale writer can commit
+                # AFTER one of these and clobber it (lost update → execution stranded
+                # in EXECUTING, a confusing success→failed flip, or a straggler
+                # callback erasing a user's stop). Lock the row and refuse to overwrite
+                # either with a different status.
                 #
-                # ERROR/STOPPED are deliberately NOT protected: other paths (upstream
-                # error, external stop, the reaper) can set them BEFORE the first real
-                # callback, which then legitimately corrects the row to COMPLETED when
-                # the files actually succeeded — blocking that would freeze a genuinely
-                # successful run at a premature ERROR. Celery executions
-                # (queue_message_id NULL) keep the legacy path.
+                # ERROR is deliberately NOT protected: other paths (upstream error,
+                # the reaper) can set it BEFORE the first real callback, which then
+                # legitimately corrects the row to COMPLETED when the files actually
+                # succeeded — blocking that would freeze a genuinely successful run at
+                # a premature ERROR. Celery executions (queue_message_id NULL) keep the
+                # legacy path.
                 if execution.queue_message_id is not None:
                     execution = WorkflowExecution.objects.select_for_update().get(
                         pk=execution.pk
                     )
-                    completed = ExecutionStatus.COMPLETED.value
-                    if execution.status == completed and status_enum.value != completed:
+                    protected = {
+                        ExecutionStatus.COMPLETED.value,
+                        ExecutionStatus.STOPPED.value,
+                    }
+                    if (
+                        execution.status in protected
+                        and status_enum.value != execution.status
+                    ):
                         logger.warning(
-                            f"update_status: refusing to overwrite COMPLETED status "
-                            f"with '{status_enum.value}' for execution {id}"
+                            f"update_status: refusing to overwrite protected status "
+                            f"'{execution.status}' with '{status_enum.value}' "
+                            f"for execution {id}"
                         )
                         return Response(
                             {
                                 "status": "skipped",
-                                "reason": "already_completed",
+                                "reason": "already_final",
                                 "execution_id": str(execution.id),
                                 "new_status": execution.status,
                             }
--- a/backend/workflow_manager/internal_views.py
+++ b/backend/workflow_manager/internal_views.py
@@ -589,12 +589,24 @@ class WorkflowExecutionInternalViewSet(viewsets.ReadOnlyModelViewSet):
         """
         from datetime import timedelta
 
+        from django.db.models import Count, Q
         from django.utils import timezone
 
         from workflow_manager.workflow_v2.enums import ExecutionStatus
 
-        stuck_seconds = int(request.data.get("stuck_seconds") or 9000)
-        limit = int(request.data.get("limit") or 100)
+        # Validate/clamp inputs: a negative stuck_seconds would push the cutoff into
+        # the future and match LIVE executions (finalizing running work); an
+        # unbounded limit would let one call scan the whole table.
+        try:
+            stuck_seconds = int(request.data.get("stuck_seconds") or 9000)
+            limit = int(request.data.get("limit") or 100)
+        except (TypeError, ValueError):
+            return Response(
+                {"error": "stuck_seconds and limit must be integers"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+        stuck_seconds = max(1, stuck_seconds)
+        limit = max(1, min(limit, 1000))
         cutoff = timezone.now() - timedelta(seconds=stuck_seconds)
         terminal = ExecutionStatus.terminal_values()
 
@@ -613,6 +625,7 @@ class WorkflowExecutionInternalViewSet(viewsets.ReadOnlyModelViewSet):
 
         recovered = 0
         skipped = 0
+        failed_recoveries = 0
         for exec_id in stuck_ids:
             try:
                 with transaction.atomic():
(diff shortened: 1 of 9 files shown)


muhammad-ali-e:
Addressed the review in `089e93a98` (+ `5694a197e` for the earlier float-equality nit):

**P1 — "In-Flight PG Writes Lose Guard" (`internal_views.py`).** Fixed. The L1 guard now keys on the row's own **`queue_message_id`** marker instead of a re-evaluated `pg_queue_enabled` Flipt lookup, so the terminal-one-way invariant holds even if the rollout flag is toggled off or Flipt is unreachable while a PG execution is still finalizing — consistent with the "transport carried on the row, never re-read from Flipt" rule (and with L2's `is_pg` / L4's `queue_message_id` gating). Celery rows (`queue_message_id` NULL) keep the legacy path; the per-call Flipt lookup + org derivation are gone.

**P1 — "Queued Execution Marked Failed" (`internal_views.py`).** Fixed. L4 no longer marks a file-less stuck execution ERROR — a valid PG message can remain queued past the stuck window during a backlog/outage, and failing it would wrongly terminalize a live execution that the one-way guard would then block from recovery. It is now **skipped** for a later sweep. A queue-state-aware file-less recovery (check `pg_queue_message` for the execution's `queue_message_id`) belongs reaper-side and is deferred to a follow-up.

**P2 — "Poison Drop Can Preserve Strand" (`callback/tasks.py`).** Acknowledged; the layers cover it. If the finalization API is persistently unavailable, the PG re-raise vt-redelivers and the poison-drop attempts a terminal ERROR write (`recovery.mark_execution_error`). If that ERROR write *also* fails (same outage), the execution stays EXECUTING but is **not** permanently stranded — the L4 reaper safety-net re-scans stuck PG executions and finalizes them to the correct terminal status once the API recovers. So L2 + L4 together close the persistent-failure case; no code change needed.

**SonarCloud.** Resolved: `logging.exception()` in the reaper sweep and `active_file_executions` except handlers; `pytest.approx(0.0)` for the zero-base jitter assertion.

Tests updated accordingly (L1 guard keyed on `queue_message_id` — PG vs Celery; file-less execution asserted skipped, not failed). 16 tests green.

workers/shared/workflow/connectors/source.py
```diff
@@ -371,6 +371,28 @@ def _get_existing_file_executions(self) -> dict[str, str]:
 
             from unstract.core.data_models import ExecutionStatus
 
+            # Fast path: one batched call returns {provider_file_uuid: status} for
+            # every file being processed across the workflow's active executions,
+            # replacing the per-active-execution N+1 below (the O(active) round-trips
+            # that saturated the backend at 800 concurrent users). Fall through to
+            # the legacy loop only if the batch endpoint is unavailable/errors.
+            batch = self.api_client.get_active_file_executions(self.workflow_id)
```
muhammad-ali-e: **[P1] The N+1 batched-dedup fix is inert — it's wired into dead code.**

This new fast-path (and its backing `get_active_file_executions` client method + `active_file_executions` backend endpoint) lives inside `_get_existing_file_executions` (def at `source.py:355`). A repo-wide search finds **zero callers** of `_get_existing_file_executions` — or its sibling `_get_existing_file_executions_optimized` (`source.py:297`) — anywhere; only the definitions and one docstring mention. Both methods were already call-less at the base commit.

The **live** dedup path is `check_files_active_processing`, invoked at `workers/shared/processing/filter_pipeline.py:692` and `workers/file_processing/tasks.py:1204`, and it is *already* a single batched call. So the PR's headline claim — "replaces the per-active-execution N+1 (the 800u backend-saturation collapse)" — is not achieved in production: none of the new endpoint / client method / fast-path is ever reached. The added backend unit test exercises the endpoint directly, so CI stays green while the code never runs.

No regression (dead code stays dead, fallback is safe), but the intended optimisation does nothing. **Fix:** either wire `get_active_file_executions` into the live path (`filter_pipeline.py` / `file_processing/tasks.py`), or drop the new endpoint + methods if `check_files_active_processing` already covers it. At minimum, confirm which method the live path uses before relying on this PR to resolve the saturation.

backend/workflow_manager/internal_views.py
```diff
@@ -503,6 +505,45 @@ def update_status(self, request, id=None):
...
+                # a non-terminal status with NULL counters (lost update → execution
+                # stranded in EXECUTING forever). When enabled: lock the row and
+                # make terminal one-way — never let a non-terminal write overwrite
+                # an already-terminal status.
+                org_id = ""
+                if execution.workflow and getattr(
+                    execution.workflow, "organization", None
+                ):
+                    org_id = str(execution.workflow.organization.organization_id or "")
+                if check_feature_flag_status(
```
muhammad-ali-e: **[P2] The guard evaluates `pg_queue_enabled` differently than the dispatcher that routed the execution.** (Extends the existing greptile note at :528 on keying the invariant off the PG marker — adds two distinct failure modes.)

`resolve_transport` (`workflow_v2/transport.py:124-127`, the authority that decides PG-vs-Celery) evaluates the *same flag* with `entity_id=str(execution_id)` and `context={organization_id, workflow_id, pipeline_id}`. This guard evaluates it with `entity_id=org_id or "default"` and `context={organization_id}` only. The inline comment "evaluated per-org exactly like transport dispatch" is therefore inaccurate — dispatch is per-*execution*, not per-org.

1. **Divergence even when the flag is globally on:** with a Flipt **percentage rollout** (hashes `entity_id`) or a **segment constraint on `workflow_id`/`pipeline_id`**, an execution that `resolve_transport` routed to PG (its `execution_id` hashed in / matched a workflow constraint) can evaluate the flag as **False** here → the terminal-one-way guard is silently skipped for exactly the PG executions it exists to protect.
2. **Unconditional Flipt round-trips on the hot status-write path:** `resolve_transport` short-circuits on `FLIPT_SERVICE_AVAILABLE != "true"` before touching Flipt; this guard calls `check_feature_flag_status` on *every* `update_status`. When Flipt is blind, each status write pays a client init + failed `list_feature_flags()` + exception + `exc_info` traceback log — amplifying load under the saturation this PR remediates.

**Fix (agrees with greptile):** gate the guard off the deterministic PG marker `queue_message_id IS NOT NULL` (as `recover_stuck_pg_executions` already does, with no Flipt call) rather than re-deriving the flag; or at least match the dispatcher's `entity_id`/context and short-circuit on `FLIPT_SERVICE_AVAILABLE`. Also fix the comment.

backend/workflow_manager/internal_views.py
```diff
@@ -537,6 +578,162 @@ def update_status(self, request, id=None):
...
+        Celery rows carry ``task_id`` and ``queue_message_id=NULL``), so it needs no
+        Flipt context and never touches Celery executions.
+        """
+        from datetime import timedelta
+
+        from django.utils import timezone
+
+        from workflow_manager.workflow_v2.enums import ExecutionStatus
+
+        stuck_seconds = int(request.data.get("stuck_seconds") or 9000)
```
muhammad-ali-e: **[P2] Unvalidated `request.data` with bare `int()` outside any try; no bounds.**

These coercions sit *before* the per-execution try (which starts at :625):
- `int("abc")` on a malformed body raises `ValueError` → unhandled 500 with a traceback, instead of a 400. The sibling `update_status` routes input through a serializer; this cross-org endpoint should too.
- **No bounds.** A negative `stuck_seconds` moves `cutoff = now() - timedelta(seconds=stuck_seconds)` into the *future*, so `modified_at__lt=cutoff` matches **live** executions — the sweep would then finalize/ERROR currently-running work. `limit` is likewise unbounded.

The reaper is the only caller today (sends validated ints), so exploit risk is low — but the endpoint's contract shouldn't rely on that. **Fix:** a small DRF serializer or explicit clamp — `stuck_seconds` min 1 (default 9000), `limit` min 1 / max ~1000 (default 100), returning 400 on invalid input.

backend/workflow_manager/internal_views.py
```diff
@@ -537,6 +578,162 @@ def update_status(self, request, id=None):
...
+                            "failed_files": failed,
+                        },
+                    )
+                    logger.warning(
+                        f"Reaper safety-net: finalized stranded PG execution "
+                        f"{exec_id} to {computed.value} "
+                        f"(files {successful} ok / {failed} failed)"
+                    )
+                    recovered += 1
+            except Exception as e:
```
muhammad-ali-e: **[P2] Per-execution recovery failures are swallowed and dropped from the response contract.**

This best-effort `except` (continuing the loop on one bad row is correct) has two observability gaps that matter for a *last-resort* safety net:
- **No `failed` counter.** The response returns only `recovered`/`skipped`/`scanned` (:697-699), so a caller can't distinguish "10 recovered, 0 failed" from "10 recovered, 90 silently failed." The `scanned != recovered + skipped` arithmetic gap is the only clue, and it isn't surfaced. Combined with the reaper only logging when `recovered > 0` (see `reaper.py:1204`), a wholesale net failure is invisible on **both** ends — the exact "silent net defeated" class this PR set out to kill.
- **No `exc_info`.** For a stranded-forever execution, *why* it can't be recovered is the important diagnostic, and it's degraded to a one-line message.

**Fix:** track `failed += 1`, add `exc_info=True`, and return `"failed": failed`.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -1177,6 +1190,29 @@ def _maybe_sweep(self) -> None:
...
+        # can't see them — this asks the backend (org-agnostic; endpoint scopes to
+        # PG rows by queue_message_id and heals server-side to the CORRECT terminal
+        # status). Cadence-gated (per-row API work) like the claims sweep; opt-in.
+        if self._stuck_recovery_enabled:
+            try:
+                resp = self._get_api_client().recover_stuck_pg_executions(
+                    stuck_seconds=self._stuck_recovery_seconds,
+                )
+                data = getattr(resp, "data", None) or {}
+                if data.get("recovered"):
```
muhammad-ali-e: **[P2] The reaper hides a wholesale safety-net failure — it only logs when `recovered > 0`.**

If the endpoint scans 100 stranded executions and every recovery attempt throws inside the per-exec loop (a regression in `update_execution`, a persistent lock/DB issue), it returns `{"recovered": 0, "skipped": 0, "scanned": 100}`. `data.get("recovered")` is `0` → falsy → **the reaper logs nothing**, while 100 executions stay stranded forever. The one place operating this last-resort net sees complete silence — the same failure mode the PR exists to eliminate, reintroduced one level up.

**Fix:** also surface the non-happy signal, e.g. `elif scanned: logger.warning("safety-net scanned %s but recovered 0 (skipped %s, failed %s) — investigate", ...)`, and include the `failed` count once the endpoint returns it (see `internal_views.py:687`).

workers/callback/tasks.py
```diff
@@ -419,6 +420,17 @@ def _update_execution_status_unified(
         logger.error(
             f"Failed to update execution status for {execution_id}: {e}", exc_info=True
         )
+        # On the PG path this finalization write is the single terminalizing step.
+        # Swallowing a failure here strands the execution in EXECUTING forever: the
+        # consumer acks/deletes the message, the barrier row is already gone, and
+        # the reaper has no handle — every safety net is silently defeated. Re-raise
+        # so PG's at-least-once vt-redelivery retries the write and, if it keeps
+        # failing, poison-drops the batch to a terminal ERROR. The re-run is
+        # idempotency-guarded by _callback_already_ran (skips once terminal). The
```
muhammad-ali-e: **[P3] Comment inaccuracy: "skips once terminal" overstates `_callback_already_ran`.**

`_callback_already_ran` (def at `tasks.py:1382`) returns `True` **only** for `COMPLETED` — its own docstring is explicit that `ERROR`/`STOPPED` are deliberately excluded (they can be set by other paths, so treating them as "already ran" would make the first legitimate callback skip its side effects). So when the failing finalization write is for an `ERROR` execution, a vt-redelivery is **not** short-circuited by this guard and the whole callback re-runs. Writing "skips once terminal" tells a maintainer the redelivery is idempotency-guarded for all terminal outcomes, which is false.

**Fix:** "(skips once **COMPLETED**; ERROR/STOPPED intentionally re-run — see `_callback_already_ran`)".

backend/workflow_manager/internal_views.py
```diff
@@ -537,6 +578,162 @@ def update_status(self, request, id=None):
...
+                with transaction.atomic():
+                    execution = WorkflowExecution.objects.select_for_update().get(
+                        pk=exec_id
+                    )
+                    # Re-check under lock: skip if it reached terminal or was just
+                    # touched (no longer stuck) — never race a live execution.
+                    if execution.status in terminal or execution.modified_at >= cutoff:
+                        skipped += 1
+                        continue
+                    files = WorkflowFileExecution.objects.filter(
```
muhammad-ali-e: **[P3 / simplify] Collapse the 3 per-execution `COUNT` round-trips into one conditional aggregate.**

This recovery loop issues three separate counts against the same queryset per stuck execution (`files.count()`, `files.filter(status__in=terminal).count()`, `files.exclude(status=COMPLETED).count()`). On a concurrency-sensitive path, one query does it:

```python
from django.db.models import Count, Q
counts = files.aggregate(
    total=Count("id"),
    terminal=Count("id", filter=Q(status__in=terminal)),
    failed=Count("id", filter=~Q(status=ExecutionStatus.COMPLETED.value)),
)
```

Behaviour-preserving: `failed` is only *used* after the `terminal_files < total` gate has proven every file is terminal, so `~Q(status=COMPLETED)` equals the ERROR/STOPPED count exactly as `.exclude(status=COMPLETED).count()` did; `total == 0` yields all-zero counts. Cuts 3 DB round-trips → 1.

muhammad-ali-e (commented):
Automated PR review (PR Review Toolkit: code-reviewer, silent-failure-hunter, type-design, test-analyzer, comment-analyzer, code-simplifier). Findings verified against the code before posting; duplicates of existing greptile comments dropped. Priority-ordered inline comments below. Full prioritised summary is in the PR conversation.

Headline: **the N+1 batched-dedup optimisation (the PR's stated primary fix) is wired into dead code and has no runtime effect** — see the comment on `source.py`.

muhammad-ali-e:
Addressed review round 2 in `2643b7a2c`:

**P1 — N+1 batched fix wired into dead code.** Correct — removed it entirely. `_get_existing_file_executions` has zero callers; the live dedup path is `check_files_active_processing`, which is already a single batched call. Dropped the `active_file_executions` endpoint, the `get_active_file_executions` client method, the `source.py` fast-path, and the N+1 test. (Aside: this also means the change_7 throughput collapse wasn't this N+1 — I'll re-investigate separately.)

**Greptile 4/5 — wrong parent status for cancelled executions.** Fixed. Recovery now recomputes the correct terminal status: **ERROR** if any file errored, else **STOPPED** if any was stopped (a cancelled run is no longer turned into ERROR), else **COMPLETED**. `successful_files`/`failed_files` follow (completed / errored counts).

**P2 — unvalidated `int()` / no bounds.** Fixed. Non-integer body → 400; `stuck_seconds` clamped to ≥1 (a negative value would push the cutoff into the future and finalize live work); `limit` clamped to 1..1000.

**P2 — silent net failure (endpoint + reaper).** Fixed on both ends. The endpoint now tracks a `failed` counter, returns it, and logs per-row failures with `exc_info`. The reaper logs whenever `recovered` **or** `failed` > 0, and separately surfaces "scanned N, recovered 0" so a wholesale failure isn't invisible.

**P2 — guard evaluates the flag differently than the dispatcher.** Resolved by the round-1 change: the guard no longer evaluates Flipt at all — it keys on the row's `queue_message_id` marker, so there's no per-execution vs per-org divergence.

**P3 — 3 COUNTs → one aggregate.** Done (single conditional `aggregate`).

**P3 — `_callback_already_ran` comment.** Fixed: "skips once COMPLETED; ERROR/STOPPED intentionally re-run (a failed-write ERROR execution simply retries the write)".

Tests: 6 worker + 12 backend (added STOPPED-recovers-to-STOPPED, ERROR-over-STOPPED priority, and negative-`stuck_seconds` clamp; removed the N+1 test). All green.

backend/workflow_manager/internal_views.py
```diff
@@ -537,6 +570,146 @@ def update_status(self, request, id=None):
...
+                        ),
+                    )
+                    total = counts["total"]
+                    if total == 0 or counts["terminal"] < total:
+                        # total==0: a valid PG message may still be QUEUED (a
+                        # backlog/outage can outlast stuck_seconds), so failing it
+                        # would wrongly terminalize a live execution the one-way guard
+                        # then blocks from recovery — file-less recovery needs a
+                        # queue-state check (reaper-side, follow-up). terminal < total:
+                        # a file is still non-terminal → genuinely processing. Skip.
```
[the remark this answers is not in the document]
muhammad-ali-e: Deliberate, and documented in the code comment on this branch. This is the exact mirror of your earlier P1 ("Queued Execution Marked Failed") on the previous revision: `total==0 → ERROR` gets flagged as *wrongly failing* a still-queued execution; `total==0 → skip` gets flagged as *stranded forever*. Both are correct — which is the point: **no file-count-only rule can distinguish "message still queued/claimed (live)" from "message lost/exhausted (dead)".** The only sound discriminator is the PG queue-message state itself (`pg_queue_message` / claim rows), not the file aggregate.

Why skip is the safe choice to ship now:
- **Non-regressive.** Before this PR there is no reaper recovery at all, so a file-less stuck PG execution already stays PENDING indefinitely. Skipping preserves exactly today's behaviour for that sub-case while fixing the *main* strand (files all terminal, execution stuck) — which is 91% of observed orphans.
- **Failing it is unsafe.** As your prior revision correctly noted, ERROR-ing a live queued execution trips the terminal-one-way guard and permanently blocks the delayed worker from finalizing it — turning a transient backlog into data loss.

The queue-state check belongs reaper-side (it reads PG-queue tables, which the backend endpoint deliberately does not touch — see the reaper "PG-tables only" principle), so it's tracked as a follow-up rather than bolted onto this cross-org status endpoint. The `total==0` branch comment states this explicitly.

backend/workflow_manager/internal_views.py
```diff
@@ -537,6 +570,146 @@ def update_status(self, request, id=None):
...
+                        skipped += 1
```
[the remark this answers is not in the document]
muhammad-ali-e: Deliberate, and documented in the code comment on this branch. This is the exact mirror of your earlier P1 ("Queued Execution Marked Failed") on the previous revision: `total==0 → ERROR` gets flagged as *wrongly failing* a still-queued execution; `total==0 → skip` gets flagged as *stranded forever*. Both are correct — which is the point: **no file-count-only rule can distinguish "message still queued/claimed (live)" from "message lost/exhausted (dead)".** The only sound discriminator is the PG queue-message state itself (`pg_queue_message` / claim rows), not the file aggregate.

Why skip is the safe choice to ship now:
- **Non-regressive.** Before this PR there is no reaper recovery at all, so a file-less stuck PG execution already stays PENDING indefinitely. Skipping preserves exactly today's behaviour for that sub-case while fixing the *main* strand (files all terminal, execution stuck) — which is 91% of observed orphans.
- **Failing it is unsafe.** As your prior revision correctly noted, ERROR-ing a live queued execution trips the terminal-one-way guard and permanently blocks the delayed worker from finalizing it — turning a transient backlog into data loss.

The queue-state check belongs reaper-side (it reads PG-queue tables, which the backend endpoint deliberately does not touch — see the reaper "PG-tables only" principle), so it's tracked as a follow-up rather than bolted onto this cross-org status endpoint. The `total==0` branch comment states this explicitly.

muhammad-ali-e:
### Review round 3 addressed

**SonarCloud — Cognitive Complexity (16 → below 15) on `recover_stuck_pg_executions`** (`482f641c`)
Extracted the per-execution recovery body (try/atomic + lock re-check + conditional aggregate + status recompute + finalize) into a new helper `_recover_one_stuck_pg_execution()` returning an outcome string (`"recovered"` / `"skipped"` / `"failed"`). The action now just tallies outcomes over the stuck-id list. No behaviour change; the 12 backend tests remain green.

**greptile P1 ×2 — "file-less executions remain stranded"** (`total==0 → skip`): replied in-thread. Deliberate and non-regressive. This is the mirror of the earlier P1 that flagged `total==0 → ERROR` as *wrongly failing a live execution* — together they prove no file-count-only rule is correct; the sound discriminator is PG queue-message state, which is a reaper-side follow-up (the backend endpoint intentionally doesn't read PG-queue tables). Skipping preserves today's pre-reaper behaviour for that sub-case while fixing the main strand (files all terminal, execution stuck).

**greptile P2 — "poison drop can preserve strand"** (`tasks.py`): acknowledged as an inherent property of the poison-drop path, out of scope for this PR — the re-raise's job is to stop *silently swallowing* the finalization failure so the redelivery/poison-drop machinery engages at all; terminalizing on ultimate poison-drop is a separate concern.

All earlier rounds (guard keyed on `queue_message_id`, inert-N+1 removal, input validation/clamps, `failed` counter + `logger.exception`, reaper scanned-but-recovered-0 log, Flipt-gated jitter, comment corrections) remain as committed.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -1177,6 +1190,42 @@ def _maybe_sweep(self) -> None:
...
+        # completed. Those have no barrier/claim row, so the PG-table sweeps above
+        # can't see them — this asks the backend (org-agnostic; endpoint scopes to
+        # PG rows by queue_message_id and heals server-side to the CORRECT terminal
+        # status). Cadence-gated (per-row API work) like the claims sweep; opt-in.
+        if self._stuck_recovery_enabled:
+            try:
+                resp = self._get_api_client().recover_stuck_pg_executions(
+                    stuck_seconds=self._stuck_recovery_seconds,
+                )
+                data = getattr(resp, "data", None) or {}
```
muhammad-ali-e: **[P2] The safety-net observability block is dead — `recovered`/`failed`/`scanned` are always `0`, so no log branch can ever fire.** (This is a *different* root cause from my earlier `:1204` note about logging only when `recovered>0`; the `elif scanned` fix added for that is itself inert.)

Response-shape mismatch across three files:
- The endpoint returns a **flat** body — `internal_views.py:706` → `Response({"recovered":.., "skipped":.., "scanned":.., "failed":..})`. DRF is configured with the plain `JSONRenderer` (`settings/base.py`), so there is no `{"data": ...}` envelope.
- `execution_client.recover_stuck_pg_executions` wraps that via `convert_dict_response(response, APIResponse)` (`execution_client.py:349`), which sets `APIResponse.data = response_dict.get("data")` → **`None`** (`response_models.py`), because the counters sit at the top level, not under `"data"`.
- Here `data = getattr(resp, "data", None) or {}` collapses to `{}`, so `recovered`/`failed`/`scanned` are permanently `0`. Neither `if recovered or failed:` nor `elif scanned:` can ever run.

Recovery still happens server-side (and the backend logs its own counts), but the reaper-side block — whose comment claims to "surface it so a wholesale-failure regression isn't invisible on this end" — is dead. The systemic-failure case (endpoint inflates `failed` to ~100 but returns HTTP 200) is exactly what stays invisible here.

**Fix** (either): backend returns `Response({"data": {"recovered":.., ...}})` so `resp.data` is populated; or the client returns the raw body (via `APIResponse.success_response(data=response)`) and the reaper reads from it. Add a loud guard — e.g. `if not data: logger.warning("safety-net: empty/unexpected recovery response: %r", resp)` — so the next shape drift isn't silent again. No worker test exercises this path, which is why it slipped through.

backend/workflow_manager/internal_views.py
```diff
@@ -503,6 +503,39 @@ def update_status(self, request, id=None):
...
+                # non-terminal status with NULL counters (lost update → execution
+                # stranded in EXECUTING forever). Lock the row and make terminal
+                # one-way. Celery executions (queue_message_id NULL) keep the legacy
+                # path unchanged.
+                if execution.queue_message_id is not None:
+                    execution = WorkflowExecution.objects.select_for_update().get(
+                        pk=execution.pk
+                    )
+                    terminal = ExecutionStatus.terminal_values()
+                    if execution.status in terminal and status_enum.value not in terminal:
```
muhammad-ali-e: **[P3] The terminal-one-way guard still allows a terminal→terminal clobber.** The predicate blocks only terminal→*non*-terminal:

```python
if execution.status in terminal and status_enum.value not in terminal:
```

A stale writer computing a *different* terminal value can still overwrite the row — e.g. the reaper safety-net writing `COMPLETED` while a redelivered `ERROR` callback writes `ERROR` (or vice-versa). This path is reachable: `_callback_already_ran` only short-circuits on `COMPLETED`, so `ERROR`/`STOPPED` redeliveries re-run and re-write, and the reaper's `recover_stuck_pg_executions` independently writes a recomputed terminal status. The result is a final status that flips based on commit ordering.

Rarer than the non-terminal strand this guard targets, hence P3 — but if the intent is truly "terminal is one-way / immutable," also skip when both current and incoming are terminal and differ (allow only the idempotent same-value rewrite).

muhammad-ali-e (commented):
Follow-up review pass on the finalization-strand fixes. The re-raise, terminal-one-way guard, STOPPED-aware recompute, input validation, and cadence-gating all look correct, and the comments are accurate. Two remaining items below — one of which silently defeats the observability that the last round of commits was written to add.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Fixed in `77015dbef`. Confirmed the exact chain you traced: the endpoint returns a flat `{"recovered":..,"skipped":..,"scanned":..,"failed":..}` body, `convert_dict_response` reads `response["data"]` → `None`, so `resp.data or {}` collapsed to `{}` and every counter was `0` → neither `if recovered or failed` nor `elif scanned` could ever run.

Went with the client-side fix (localized, leaves the HTTP contract and the backend tests untouched): `recover_stuck_pg_executions` now returns `APIResponse.success_response(data=response)` so the whole body lands in `.data`. Added the loud guard you suggested — `if not data: logger.warning("safety-net recovery returned an empty/unexpected response (%r) …")` — so the next shape drift is surfaced, not silent. And added the missing worker test (`TestRecoverStuckClientResponseShape`) covering both a flat body → populated `.data` and a non-dict body → empty `.data` (guard path).

backend/workflow_manager/internal_views.py
muhammad-ali-e: Fixed in `77015dbef`. Tightened the predicate to `execution.status in terminal and status_enum.value != execution.status` — terminal is now fully one-way: only an idempotent same-value rewrite is allowed, so terminal→non-terminal (the strand) **and** terminal→a different terminal are both rejected. That closes the commit-order flip (redelivered `ERROR` over an earlier `COMPLETED`) and, notably, stops a late `COMPLETED` from erasing a user's `STOPPED` abort.

PG-scoped as before (`queue_message_id`), so the Celery path is unchanged. The reaper's own recovery write can't conflict — it already skips any row that's terminal under its lock before writing. Added 3 backend tests: different-terminal rejected, STOPPED-abort survives a late COMPLETED, and same-value rewrite still allowed.

backend/workflow_manager/internal_views.py
```diff
@@ -503,6 +503,39 @@ def update_status(self, request, id=None):
...
+                # is unreachable while a PG execution is still finalizing (per the
+                # "transport carried on the row, never re-read from Flipt" rule).
+                # Under the PG path's high concurrency ~15 call sites write this
+                # status via unlocked read-modify-write, so a stale writer can commit
+                # AFTER the callback's terminal write and clobber it back to a
+                # non-terminal status with NULL counters (lost update → execution
+                # stranded in EXECUTING forever). Lock the row and make terminal
+                # one-way. Celery executions (queue_message_id NULL) keep the legacy
+                # path unchanged.
+                if execution.queue_message_id is not None:
```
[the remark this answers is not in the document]
muhammad-ali-e: Confirmed real and structural — thanks. The ordering is: `_dispatch_orchestrator_task` enqueues (message claimable) → `dispatched=True` → **then** `_record_dispatch_handle` writes `queue_message_id` (best-effort, its own try/except at `workflow_helper.py:671-684`). So there's a genuine window where a PG message is claimable while `queue_message_id` is still NULL, and if that best-effort write *fails* it stays NULL indefinitely (guard never fires, and the reaper's `queue_message_id__isnull=False` filter never sees it either).

I'm **not** fixing it in this PR, deliberately:
- The only sound close is a **creation-time transport marker** (persist the resolved transport on the row *before* enqueue, or a two-phase "enqueue-invisible → record → make-visible"). There is no such marker today — `task_id`/`queue_message_id` are both written post-dispatch (`models/execution.py:143-152`) — so it's a schema + dispatch/creation-path change, its own ticket.
- I can't key the guard on `task_id IS NULL` as a proxy, because a genuinely-Celery execution is *also* NULL during its own post-dispatch window → that would rope the Celery path into the guard, which this PR is committed not to disturb.
- Scope/severity: the window is dispatch→post-dispatch-write (typically ms, same request), so only an execution that fully finalizes inside it is exposed. The guard is still a net improvement over none, and this is the same "needs infra the endpoint deliberately doesn't own → follow-up" boundary as the file-less queue-state check.

Filing a follow-up ticket for the creation-time transport marker (closes both this window and the write-failed-→-NULL-forever case). Will link it here.

muhammad-ali-e:
### Review round 4 addressed (`77015dbef`)

**P2 — reaper safety-net observability was dead** (`reaper.py`, `execution_client.py`): the endpoint returns a flat body with no `{"data": …}` envelope, so `convert_dict_response` yielded `resp.data = None` → every counter `0` → no log branch could fire. Fixed client-side (surface the whole body as `.data`) + added the loud empty/unexpected-response guard. New worker test `TestRecoverStuckClientResponseShape` covers flat-body and non-dict-body shapes.

**P3 — terminal→terminal clobber** (`internal_views.py`): the guard now allows only an idempotent same-value terminal rewrite, so terminal→non-terminal **and** terminal→a different terminal are both rejected (stops commit-order flips and a late `COMPLETED` erasing a user's `STOPPED` abort). Still PG-scoped; Celery unchanged. 3 new backend tests.

**P1 (greptile `:519`) — guard starts too late** (`queue_message_id` written post-dispatch): confirmed real and structural. Not fixed here by design — the sound close is a creation-time transport marker (schema + dispatch/creation-path change), and keying the guard on `task_id IS NULL` would wrongly rope in the Celery path. Deferred to a follow-up ticket (also covers the "best-effort record fails → NULL forever" case); rationale on the thread.

**File-less-strand P1 ×2:** greptile withdrew both after the queue-state-discriminator explanation.

Tests: backend 15 green, workers 10 green.

backend/workflow_manager/internal_views.py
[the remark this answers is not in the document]
muhammad-ali-e: Follow-up filed: **UN-3715** — *creation-time transport marker (close the queue_message_id NULL window in the terminal-one-way guard)*, covering both the startup window and the write-failed-→-NULL-forever case.

backend/workflow_manager/internal_views.py
muhammad-ali-e: Correction to my round-4 fix (`c9b88a1e4`). Making *all* terminal states immutable was too aggressive — it blocked a **legitimate** transition. Per `_callback_already_ran`, `ERROR`/`STOPPED` can be set by other paths (upstream error, external stop, the reaper) **before** the first real callback runs; that callback then correctly writes `COMPLETED` when the files actually succeeded. Round-4 froze that at a premature `ERROR` — i.e. a genuinely successful run showing as failed.

Re-scoped the guard to **`COMPLETED` specifically** (the codebase's only truly-final status): refuse `COMPLETED → anything-else` (the strand + the confusing success→failed flip), but allow `ERROR`/`STOPPED → COMPLETED` (correct a premature failure). This is stricter than the original terminal→non-terminal guard (it also blocks a stale `COMPLETED → ERROR`) and correctly narrower than round-4. Matches `_callback_already_ran` exactly. Reason code `already_terminal` → `already_completed`; tests updated (COMPLETED protected both directions; ERROR→COMPLETED now allowed).

muhammad-ali-e:
### Review round 5 — terminal guard re-scoped to COMPLETED (`c9b88a1e4`)

Round 4 made **all three** terminal states immutable. That was too aggressive — it blocked a **legitimate** transition: per `_callback_already_ran`, `ERROR`/`STOPPED` can be set by other paths (upstream error, external stop, the reaper) *before* the first real callback runs, and that callback then correctly writes `COMPLETED` when the files actually succeeded. Freezing it at `ERROR` would show a genuinely successful run as failed.

**Fix:** guard `COMPLETED` specifically — the codebase's only truly-final status (set exclusively by a successful callback; a redelivered callback on a `COMPLETED` row is skipped).

| Transition | Guard | Why |
|---|---|---|
| `COMPLETED → EXECUTING/PENDING` | **blocked** | the strand bug (stale clobber) |
| `COMPLETED → ERROR` | **blocked** | the confusing success→failed flip — never legitimate |
| `COMPLETED → STOPPED` | **blocked** | nonsensical (can't abort a finished run) |
| `ERROR`/`STOPPED → COMPLETED` | **allowed** | premature failure corrected to the true result |
| `COMPLETED → COMPLETED` | allowed | idempotent redelivery |

This is stricter than the original terminal→non-terminal guard (it also blocks a stale `COMPLETED → ERROR`) and correctly narrower than round 4. Still PG-scoped by `queue_message_id`; Celery unchanged. Reason code `already_terminal` → `already_completed`; backend tests updated (COMPLETED protected in both directions, `ERROR → COMPLETED` now allowed) — 15 green.

Checked the reverse `STOPPED → COMPLETED` against the abort path: there's no live mid-flight cancel primitive on the PG path today (STOPPED is a computed terminal outcome, not a real-time abort — per-file cancel is a *future* epic item), so leaving that direction as last-writer is non-regressive and needs no change now.

backend/workflow_manager/internal_views.py
```diff
@@ -503,6 +503,47 @@ def update_status(self, request, id=None):
...
+                # callback, which then legitimately corrects the row to COMPLETED when
+                # the files actually succeeded — blocking that would freeze a genuinely
+                # successful run at a premature ERROR. Celery executions
+                # (queue_message_id NULL) keep the legacy path.
+                if execution.queue_message_id is not None:
+                    execution = WorkflowExecution.objects.select_for_update().get(
+                        pk=execution.pk
+                    )
+                    completed = ExecutionStatus.COMPLETED.value
+                    if execution.status == completed and status_enum.value != completed:
```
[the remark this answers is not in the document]
muhammad-ali-e: Good catch — adopted in `a29c51315`. You're right that a straggler callback finishing after an explicit stop shouldn't silently erase it. Now protecting both COMPLETED and STOPPED (the two statuses that assert a *deliberate* final outcome) via a `protected` set, exactly as suggested. ERROR stays unprotected so the premature-ERROR → real-COMPLETED correction still works. reason code `already_completed` → `already_final`; added a backend test (`test_pg_rejects_completed_over_stopped`).

The asymmetry is deliberate: an automatic ERROR should be correctable upward, but a human/operator STOPPED — like a successful COMPLETED — should not be overwritten by a late writer.

muhammad-ali-e:
### Review round 6 (`a29c51315`)

**greptile P1 — preserve explicit stopped state:** a straggler callback finishing after a user/operator stop could overwrite `STOPPED` with `COMPLETED`, erasing the stop. Now the guard protects **both `COMPLETED` and `STOPPED`** (the two deliberate-final statuses); `ERROR` stays unprotected so the premature-`ERROR` → real-`COMPLETED` correction still works. `reason` `already_completed` → `already_final`; new test `test_pg_rejects_completed_over_stopped`.

| Existing → incoming | Guard |
|---|---|
| `COMPLETED` / `STOPPED` → any different status | **blocked** |
| `ERROR` → `COMPLETED` (premature failure corrected) | allowed |
| same-value rewrite | allowed |
| Celery row (`queue_message_id` NULL) | legacy, unchanged |

**SonarCloud S3776** — `_maybe_sweep` cognitive complexity 17>15: extracted the opt-in stuck-execution recovery block into `_maybe_recover_stuck_executions()`. No behaviour change.

Backend tests 16 green.

PR #2173, opened 2026-07-14
UN-3569 [FEAT] Pass group_key through agentic_params in structure_tool_task

harini-venkataraman:
## What

- Pass the `group_key` setting from agentic table tool metadata through to the worker's `agentic_params` dict, enabling the opt-in section-level field extraction feature in the agentic table plugin

## Why

The cloud repo's `payload_modifier` now exports `group_key` in the tool settings. The OSS `structure_tool_task.py` needs to read it from `at_settings` and include it in `agentic_params` so the agentic table worker receives the configuration.

## How

- **`workers/file_processing/structure_tool_task.py`**:
  - Added type annotation `dict[str, Any]` to `agentic_params` declaration
  - After building the base `agentic_params` dict, reads `group_key` from `at_settings` (default `""`)
  - Conditionally adds `group_key` to `agentic_params` only when non-empty (matching the cloud payload modifier pattern)

```python
group_key = at_settings.get("group_key", "")
if group_key:
    agentic_params["group_key"] = group_key
```

No. The change is purely additive:
- `at_settings.get("group_key", "")` returns `""` when the key doesn't exist (all current deployments), so the `if group_key:` guard is `False` and `agentic_params` is unchanged
- The type annotation change (`dict[str, Any]`) is cosmetic and has no runtime effect
- The agentic table worker already handles the absence of `group_key` gracefully (defaults to `""` via `params.get(ParamKeys.GROUP_KEY, "")`)

## Relevant Docs

- [UN-3569](https://zipstack.atlassian.net/browse/UN-3569)

## Related Issues or PRs

- Cloud repo PR: UN-3569 [FEAT] Add opt-in group_key for unlabeled section-level field extraction

- No new dependencies or env variables
- Requires the corresponding cloud repo PR to be deployed for `group_key` to appear in `at_settings`

## Notes on Testing

- No unit test changes needed — the `structure_tool_task.py` function is integration-tested via end-to-end pipeline runs
- Verify by configuring `group_key` in agentic table settings and running an extraction — the worker should receive `group_key` in its params

## Screenshots

N/A — backend-only change

## Checklist

[UN-3569]: https://zipstack.atlassian.net/browse/UN-3569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


workers/file_processing/structure_tool_task.py
```diff
@@ -435,7 +435,7 @@ def _execute_structure_tool_impl(params: dict) -> dict:
         json_structure = at_settings.get("json_structure")
         if isinstance(json_structure, dict):
             json_structure = json.dumps(json_structure)
-        agentic_params = {
+        agentic_params: dict[str, Any] = {
```
jaseemjaskp: [nit] This annotation is a no-op — it doesn't buy the type safety it looks like it does.

`_execute_structure_tool_impl(params: dict)` (line 213) takes a bare `dict`, so `at_output["llm"]` and every `at_settings.get(...)` here are `Any`. mypy joins `Any` with `str` to `Any`, so this literal **already** infers `dict[str, Any]`, and the `agentic_params["group_key"] = ...` assignment on line 457 type-checks clean without the annotation. Verified empirically under this package's own strict flags (`workers/pyproject.toml:122-135`): `Revealed type is "dict[str, Any]"`, `Success: no issues found`.

It's also unenforced either way — mypy is commented out at `.pre-commit-config.yaml:71-78` and doesn't run in any CI workflow.

Harmless as self-documentation, and it does mirror `agentic_results: dict[str, Any]` on line 432, so keeping it is fine. Two things worth knowing: it pins the *weakest* possible type (if `params` were ever typed properly, this would preserve the `Any` bag rather than let inference tighten), and its siblings in this same function — `answer_params`, `extract_params`, `pipeline_options` (lines ~364-405) — are all unannotated. If the conditional below collapses into the literal, this should go with it.

workers/file_processing/structure_tool_task.py
```diff
@@ -452,6 +452,9 @@ def _execute_structure_tool_impl(params: dict) -> dict:
             "execution_id": execution_id,
             "PLATFORM_SERVICE_API_KEY": platform_service_api_key,
         }
+        group_key = at_settings.get("group_key", "")
```
jaseemjaskp: [nit] **This is the main review comment — the other two are minor.**

Two things here, one substantive:

**1. Absent-vs-empty is undocumented and unverifiable.** All eleven sibling keys use `at_settings.get(key, default)` unconditionally — including `"target_table": at_settings.get("target_table", "")` (446) and `"instructions": at_settings.get("instructions", "")` (448), which forward `""` happily. `group_key` alone is truthiness-gated, so it's *absent* from the payload rather than empty. That's a different wire payload, and it makes the key set variadic where it was previously fixed at 13.

Whether that matters depends entirely on the executor:
- consumer does `.get("group_key", "")` → the conditional is **pure noise**; the inline form is behaviourally identical
- consumer does `"group_key" in params` → the conditional is **load-bearing and correct**

`group_key` appears nowhere else in this repo and the executor is out-of-tree, so neither I nor the next maintainer can tell which holds. That's the actual problem: someone normalising this block for consistency would flatten it to the one-liner and silently change behaviour.

**Please pick one:**

*(a) If the executor defaults `group_key` to `""`* — delete 455-457, drop the annotation on 438, and add to the literal after line 448:
```python
            "group_key": at_settings.get("group_key", ""),
```

*(b) If absence is genuinely meaningful* — keep it, but record why, since nothing in this repo can:
```python
        # Omitted rather than blanked when unset: the agentic_table executor
        # branches on key *presence* to enable section-level extraction, so an
        # empty group_key is not equivalent to an absent one. Do not fold into
        # the literal above.
```
This is exactly the kind of constraint the excellent `Important:` comment at 425-430 already models for the SOURCE-vs-INFILE choice — an invariant whose violation fails far from this line. That block sets the bar for this file, and it's why the silence here reads as an outlier rather than house style.

**Do not do (a) on my say-so** — you have executor access and I don't. If the conditional is redundant, keeping it costs nothing; if collapsing it is wrong, it breaks every non-grouped agentic_table prompt.

**2. Dead default (minor, applies only if you keep the conditional).** The `""` on this line is never observable — the value is only read through `if group_key:` on 456, and `None` is equally falsy. `at_settings.get("group_key")` is identical. Optionally tighter:
```python
        if group_key := at_settings.get("group_key"):
            agentic_params["group_key"] = group_key
```
Walrus is rare but not unprecedented in `workers/` (`shared/enums/worker_enums_base.py:100`, `shared/workflow/execution/service.py:1132,1219`). Dropping just the `, ""` is equally fine if you'd rather stay conventional.

workers/file_processing/structure_tool_task.py
```diff
@@ -452,6 +452,9 @@ def _execute_structure_tool_impl(params: dict) -> dict:
...
+        if group_key:
+            agentic_params["group_key"] = group_key
         at_ctx = ExecutionContext(
```
jaseemjaskp, line 478: [nit] Optional, pre-existing, and **not** something to fix in this PR — flagging only because it's adjacent and this diff is what makes it matter.

Unlike the `structure_pipeline` dispatch at `:481-488`, which logs tool/skip/summarize/single-pass state before dispatching, this `agentic_table` dispatch has no log line at all — no dispatch log, no result log. So there's currently no breadcrumb to tell you, from a deployed run, whether `group_key` was forwarded or quietly not configured. That's the one observability gap I'd actually care about here: if the `payload_modifier` plugin ever fails to export a configured `group_key`, this layer genuinely cannot distinguish it from "user didn't configure grouping", and there'd be no trace either way.

I'm deliberately *not* asking for a `group_key`-only log line — it'd be an orphan in an otherwise silent block. If you or a follow-up wants to close this, the whole-dispatch version gets `group_key` observability for free:

```python
logger.info(
    "Dispatching agentic_table: prompt=%s target_table=%s pages=%s-%s group_key=%s",
    at_output[_SK.NAME],
    agentic_params["target_table"],
    agentic_params["starting_page"],
    agentic_params["ending_page"],
    group_key or "<none>",
)
```

Separate PR is the right home.

jaseemjaskp (approved):
Reviewed with the PR Review Toolkit (code review, silent-failure, type design, test coverage, comments, simplification).

**Verdict: approve.** The change is additive, conservative, and cannot regress existing behaviour — `if group_key:` drops only falsy values, which is exactly the set meaning "not configured". No correctness, security, concurrency, or performance findings. Everything below is a nit; none blocks merge.

One theme runs through every finding, and it's worth one reply from you rather than three:

> **Does the `agentic_table` executor distinguish an absent `group_key` from `""`?**

`group_key` appears nowhere else in this repo — the producer (`export_agentic_table_settings`, via the `payload_modifier` plugin at `prompt_studio_registry_helper.py:391`) and the consumer (the `agentic_table` executor, reached across a Celery boundary) are both out-of-tree. So the one genuinely non-obvious decision in this diff is unverifiable by anyone reading this repository. If the answer is "no", the four lines collapse to one; if "yes", they need a comment saying so. Either way the next reader shouldn't have to guess.

A note against over-reacting to the nits: **please don't collapse this blind.** The failure modes are asymmetric — if the conditional is redundant, keeping it costs nothing; if collapsing it is wrong, it breaks every *non-grouped* agentic_table prompt, which is the majority path. The current form is the safe default.

We also explicitly considered and **rejected** two things you might otherwise expect from a review like this:
- **No test requested.** The agentic_table dispatch block (`structure_tool_task.py:431-474`) has zero coverage today, but that gap is pre-existing and closing it here would cost a ~50-line fixture to assert that four lines do what they plainly say, mock-to-mock, with no in-repo consumer. Worth its own PR; not a gate on this one.
- **No validation / no fail-loud requested.** `group_key` is optional, and the fail-loud block at `:319-330` correctly covers only the *required* fields (`target_table`, `json_structure`). This PR matches the convention of the five optional settings above it.

Two pre-existing issues surfaced that are out of scope here but worth filing separately:
1. The `agentic_table` dispatch (`:458-470`) is entirely unlogged, unlike the `structure_pipeline` dispatch at `:481-488`.
2. `prompt_studio_registry_helper.py:387-395` has no `else` — if the `payload_modifier` plugin is missing, export silently succeeds with unpopulated settings, and the eventual error at `structure_tool_task.py:322-330` tells the user to "re-export", which will never help. That one is a real silent failure and the most valuable thing this review turned up.

pk-zipstack (approved):

chandrasekharan-zipstack (approved):

PR #2174, opened 2026-07-14
UN-3717 [FIX] default PG-reaper stuck-execution recovery ON (kill-switch env)

muhammad-ali-e:
## What
Default the PG reaper's stranded-execution recovery sweep to **ON**, disabled only by an explicit `WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED="false"` kill-switch.

## Why
The recovery sweep — the safety-net that finalizes PG executions stranded non-terminal after all their files completed (the finalization strand, fixed in #2172) — was gated behind a default-**off** env. Because it was never set, the one net that heals those strands did nothing, and they silently accumulated (a batch of stranded `EXECUTING`+NULL-counter executions was observed in integration, ~4.5% of a load-test run).

Every *other* reaper sweep (barrier recovery, orphan-claim GC, dedup GC, VT re-arm) is always-on. Recovery is equally core to the reaper's job, so a default-off gate is an inconsistent footgun.

Making it default-on in the **code** also means no chart/`values.yaml` change is needed to enable it — deployments inherit it; the env only needs to be set to *disable* it.

## How
- Extract `stuck_recovery_enabled_from_env()` (mirrors `lease_seconds_from_env` / `_positive_duration_from_env`): enabled unless the env is explicitly `"false"`.
- Wire it into `PgReaper.__init__`.

## Safety
- Recovery only acts when **all** of an execution's files are already terminal (skips still-processing / file-less) — it reconciles the parent status, never re-runs work. So default-on is safe.
- Kill-switch preserved (`...ENABLED="false"`) for instant rollback.
- Reaper-only; no other worker is affected.

## Tests
- 5 unit tests: default-on, explicit-`false` disables, case-insensitive, non-`false` values stay on.
- Full worker finalization suite green (15 passed).

UN-3717 — follow-up to #2172.

### DIFF
--- a/workers/queue_backend/pg_queue/reaper.py
+++ b/workers/queue_backend/pg_queue/reaper.py
@@ -218,7 +218,8 @@ def stuck_recovery_enabled_from_env() -> bool:
     one safety-net that heals them did nothing until someone remembered the env.
     """
     return (
-        os.getenv("WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED", "true").lower() != "false"
+        os.getenv("WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED", "true").strip().lower()
+        != "false"
     )
 
 
--- a/workers/queue_backend/pg_queue/reaper.py
+++ b/workers/queue_backend/pg_queue/reaper.py
@@ -207,6 +207,21 @@ def dedup_retention_from_env() -> int:
     )
 
 
+def stuck_recovery_enabled_from_env() -> bool:
+    """Whether the reaper's stranded-execution recovery sweep runs — ON by default.
+
+    Like every other reaper sweep (barrier / orphan-claim / dedup GC), finalizing PG
+    executions stranded non-terminal after all their files completed is core to the
+    reaper's job, so it defaults on. ``WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED=
+    "false"`` is a kill-switch; any other value (or unset) leaves it enabled. A
+    default-off gate was a footgun — it let strands silently accumulate because the
+    one safety-net that heals them did nothing until someone remembered the env.
+    """
+    return (
+        os.getenv("WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED", "true").lower() != "false"
+    )
+
+
 def _rollback_after_sweep_failure(conn: PgConnection, table: str) -> None:
     """Roll back after a failed sweep DELETE; surface a rollback that itself fails.
 
@@ -974,13 +989,11 @@ class PgReaper:
         if self._dedup_retention <= 0:
             raise ValueError("dedup_retention_seconds must be positive")
         # Safety-net recovery of PG executions stranded non-terminal after all files
-        # completed (barrier gone → invisible to the PG-table sweeps). Opt-in
-        # (default off) for a controlled rollout; the stuck window defaults to the
-        # barrier stuck-timeout so it never races a legitimately long execution.
-        self._stuck_recovery_enabled = (
-            os.getenv("WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED", "false").lower()
-            == "true"
-        )
+        # completed (barrier gone → invisible to the PG-table sweeps). ON by default
+        # (kill-switch via WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED="false"); the
+        # stuck window defaults to the barrier stuck-timeout so it never races a
+        # legitimately long execution.
+        self._stuck_recovery_enabled = stuck_recovery_enabled_from_env()
         self._stuck_recovery_seconds = _positive_duration_from_env(
             "WORKER_PG_STUCK_EXECUTION_RECOVERY_SECONDS",
             self._stuck_timeout_seconds,
--- a/workers/queue_backend/pg_queue/reaper.py
+++ b/workers/queue_backend/pg_queue/reaper.py
@@ -210,16 +210,18 @@ def dedup_retention_from_env() -> int:
 def stuck_recovery_enabled_from_env() -> bool:
     """Whether the reaper's stranded-execution recovery sweep runs — ON by default.
 
-    Like every other reaper sweep (barrier / orphan-claim / dedup GC), finalizing PG
-    executions stranded non-terminal after all their files completed is core to the
-    reaper's job, so it defaults on. ``WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED=
-    "false"`` is a kill-switch; any other value (or unset) leaves it enabled. A
-    default-off gate was a footgun — it let strands silently accumulate because the
-    one safety-net that heals them did nothing until someone remembered the env.
+    Finalizing PG executions stranded non-terminal after all their files completed is
+    core to the reaper's job (like the barrier / orphan-claim / dedup sweeps), so it
+    defaults on. Parsed like the codebase's other default-on flags (``CACHE_REDIS_
+    ENABLED``, ``ENABLE_METRICS``): unset → on; set the env to any non-``"true"``
+    value (conventionally ``"false"``) to disable — a kill-switch that ``strip()`` +
+    ``lower()`` make robust to a stray space, newline, or case from a ConfigMap. A
+    default-off gate was a footgun: it let strands silently accumulate because the one
+    safety-net that heals them did nothing until someone set the env.
     """
     return (
         os.getenv("WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED", "true").strip().lower()
-        != "false"
+        == "true"
     )
 
 
@@ -990,9 +992,9 @@ class PgReaper:
         if self._dedup_retention <= 0:
             raise ValueError("dedup_retention_seconds must be positive")
         # Safety-net recovery of PG executions stranded non-terminal after all files
-        # completed (barrier gone → invisible to the PG-table sweeps). ON by default
-        # (kill-switch via WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED="false"); the
-        # stuck window defaults to the barrier stuck-timeout so it never races a
+        # completed (barrier gone → invisible to the PG-table sweeps). Enablement +
+        # kill-switch semantics live in stuck_recovery_enabled_from_env(); the stuck
+        # window (below) defaults to the barrier stuck-timeout so it never races a
         # legitimately long execution.
         self._stuck_recovery_enabled = stuck_recovery_enabled_from_env()
         self._stuck_recovery_seconds = _positive_duration_from_env(
--- a/workers/tests/test_pg_finalization_fixes.py
+++ b/workers/tests/test_pg_finalization_fixes.py
@@ -114,3 +114,38 @@ class TestRecoverStuckClientResponseShape:
         # A malformed (non-dict) body → empty data → the reaper's loud-guard fires.
         resp = self._client(None).recover_stuck_pg_executions()
         assert not resp.data
+
+
+class TestStuckRecoveryDefaultOn:
+    """Stranded-execution recovery is ON by default in the PG reaper; only an
+    explicit 'false' kill-switch disables it (a default-off gate previously let
+    strands silently accumulate)."""
+
+    def _flag(self, value):
+        import os
+
+        from queue_backend.pg_queue.reaper import stuck_recovery_enabled_from_env
+
+        key = "WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED"
+        with patch.dict(os.environ, {}, clear=False):
+            if value is None:
+                os.environ.pop(key, None)
+            else:
+                os.environ[key] = value
+            return stuck_recovery_enabled_from_env()
+
+    def test_default_unset_is_enabled(self):
+        assert self._flag(None) is True
+
+    def test_explicit_false_disables(self):
+        assert self._flag("false") is False
+
+    def test_false_is_case_insensitive(self):
+        assert self._flag("FALSE") is False
+
+    def test_true_stays_enabled(self):
+        assert self._flag("true") is True
+
+    def test_only_false_kills_it(self):
+        # kill-switch semantics: a typo/other value must not silently disable it.
+        assert self._flag("1") is True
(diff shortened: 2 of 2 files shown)


muhammad-ali-e (commented):
## PR Review Toolkit — automated review

Ran six agents (Code Reviewer, Silent-Failure Hunter, Type-Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier) over the diff vs `feat/UN-3445-pg-queue-integration`.

The default-OFF → default-ON flip is correct and well-motivated, and line-length/test-isolation are clean. The one theme surfaced independently by four agents: the **kill-switch string→bool decode is brittle** (no `.strip()`, only exact `"false"` disables). Plus two behavioral test gaps. None are strictly blocking; the `.strip()` fix is the highest-value one. Details inline.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -207,6 +207,21 @@ def dedup_retention_from_env() -> int:
...
+
+    Like every other reaper sweep (barrier / orphan-claim / dedup GC), finalizing PG
+    executions stranded non-terminal after all their files completed is core to the
+    reaper's job, so it defaults on. ``WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED=
+    "false"`` is a kill-switch; any other value (or unset) leaves it enabled. A
+    default-off gate was a footgun — it let strands silently accumulate because the
+    one safety-net that heals them did nothing until someone remembered the env.
+    """
+    return (
+        os.getenv("WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED", "true").lower() != "false"
```
muhammad-ali-e: **[High] Kill-switch decode is brittle — flagged independently by the silent-failure, type-design, and test agents.**

```python
os.getenv("WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED", "true").lower() != "false"
```

Two operational footguns, both in the dangerous direction (fails toward *enabled*):

1. **No `.strip()`** — `" false "` or a trailing newline (routine in k8s ConfigMap/YAML env values) → `!= "false"` → stays **ON**. An operator disabling recovery in an incident with a stray space silently fails to disable it. `workers/log_consumer/tasks.py:34` already strips for exactly this reason.
2. **Only the exact token `"false"` disables** — `0`, `no`, `off`, `disabled` all leave the sweep armed with no warning. This also **inverts the codebase convention**: every other boolean env flag uses affirmative `== "true"` parsing (e.g. `shared/infrastructure/config/worker_config.py:160,272,301`); this is the only `!= "false"` in the tree.

The "a typo like `1` must not disable" intent (per `test_only_false_kills_it`) is fine — but robustness against whitespace and common falsey tokens doesn't conflict with it. Matching the sibling `_positive_duration_from_env` "loud-on-misconfig" posture:

```python
def stuck_recovery_enabled_from_env() -> bool:
    raw = os.getenv("WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED")
    if raw is None:
        return True
    token = raw.strip().lower()
    if token in ("true", "1", "yes", "on"):
        return True
    if token in ("false", "0", "no", "off"):
        return False
    raise ValueError(
        f"WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED={raw!r} is not a boolean "
        "(use true/false). Unset it to default to enabled."
    )
```

At minimum, add `.strip()`.

workers/tests/test_pg_finalization_fixes.py
```diff
@@ -114,3 +114,38 @@ def test_non_dict_body_yields_empty_data(self):
         # A malformed (non-dict) body → empty data → the reaper's loud-guard fires.
         resp = self._client(None).recover_stuck_pg_executions()
         assert not resp.data
+
+
+class TestStuckRecoveryDefaultOn:
```
muhammad-ali-e: **[Medium] Coverage gaps — the tests exercise the parser in isolation but not the two things that actually change behavior.**

1. **Sweep gating untested** — nothing asserts that `_maybe_recover_stuck_executions` (`reaper.py:1218`) early-returns when the flag is off. A regression that drops or inverts that guard passes every current test. Use the same `__new__` construction pattern as `TestRecoverStuckClientResponseShape` above: with `_stuck_recovery_enabled=False` and a mock API client, assert `recover_stuck_pg_executions` is **not** called; with `True`, assert it **is**.
2. **Constructor wiring untested** — nothing pins that `PgReaper.__init__` (`reaper.py:996`) actually sources `self._stuck_recovery_enabled` from this helper; hardcoding `= True` would go unnoticed. Patch `queue_backend.pg_queue.reaper.stuck_recovery_enabled_from_env` and assert the constructor calls it and assigns the result.

Also add two edge cases to `_flag`, since both stay ON and matter for the kill-switch contract:
- `assert self._flag("  false  ") is True` — pins the no-strip behavior (or flips to `False` if you add `.strip()` per the reaper.py comment).
- `assert self._flag("") is True` — explicitly-set-empty env is a common deployment artifact.

workers/tests/test_pg_finalization_fixes.py
```diff
@@ -114,3 +114,38 @@ def test_non_dict_body_yields_empty_data(self):
...
+    """Stranded-execution recovery is ON by default in the PG reaper; only an
+    explicit 'false' kill-switch disables it (a default-off gate previously let
+    strands silently accumulate)."""
+
+    def _flag(self, value):
```
muhammad-ali-e: **[Low] Simplify env juggling with `monkeypatch`.** The empty `patch.dict(..., {}, clear=False)` opened purely for teardown, plus manual `pop`/set and the local `import os`, all collapse into pytest's `monkeypatch` (same isolate-and-restore intent):

```python
def _flag(self, monkeypatch, value):
    from queue_backend.pg_queue.reaper import stuck_recovery_enabled_from_env

    key = "WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED"
    monkeypatch.delenv(key, raising=False)
    if value is not None:
        monkeypatch.setenv(key, value)
    return stuck_recovery_enabled_from_env()
```

Each test passes `monkeypatch` through, e.g. `self._flag(monkeypatch, None)`. (The current isolation is *correct* — this is pure readability.)

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -974,13 +989,11 @@ def __init__(
...
             raise ValueError("dedup_retention_seconds must be positive")
         # Safety-net recovery of PG executions stranded non-terminal after all files
-        # completed (barrier gone → invisible to the PG-table sweeps). Opt-in
-        # (default off) for a controlled rollout; the stuck window defaults to the
-        # barrier stuck-timeout so it never races a legitimately long execution.
-        self._stuck_recovery_enabled = (
-            os.getenv("WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED", "false").lower()
-            == "true"
-        )
+        # completed (barrier gone → invisible to the PG-table sweeps). ON by default
```
muhammad-ali-e: **[Low] Comment now duplicates the helper's docstring.** Since the enable/kill-switch logic moved into `stuck_recovery_enabled_from_env()` (whose docstring already explains default-on + the `="false"` kill-switch), this constructor comment repeats it. Its remaining *unique* value is the stuck-window half ("defaults to the barrier stuck-timeout so it never races a legitimately long execution"), which sits above `self._stuck_recovery_seconds`. Consider trimming the enable/kill-switch sentence here and letting the docstring own it — keeps each fact in one place.

workers/queue_backend/pg_queue/reaper.py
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in `89a8d7c2e` — added `.strip()` before the compare, so `false ` / `  FALSE\n` (a trailing space or newline from a configmap value) still disables recovery. Test covers both.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -974,13 +989,11 @@ def __init__(
...
+        # (kill-switch via WORKER_PG_STUCK_EXECUTION_RECOVERY_ENABLED="false"); the
+        # stuck window defaults to the barrier stuck-timeout so it never races a
+        # legitimately long execution.
+        self._stuck_recovery_enabled = stuck_recovery_enabled_from_env()
```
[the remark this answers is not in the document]
muhammad-ali-e: Real property, and worth stating explicitly — but it isn't introduced by this PR. Recovery has always been status-reconciliation-only (it finalizes the parent row from the already-terminal files; it does **not** re-fire notifications / usage accounting / destination push). This PR only flips the *default*; the trade-off is the same one recovery has always had.

Why it's acceptable as the default:

1. **The window makes racing a live callback unlikely.** `stuck_recovery_seconds` defaults to the barrier stuck-timeout (hours). A finalization callback still un-run after that means its worker is effectively dead — at which point recovery is the *only* path that ever finalizes the execution.
2. **The strands this exists to heal already ran their side effects.** In the observed cases the callback logged `completed successfully` and did the destination push / usage write — only the *status write* was lost to the race. Recovery just repairs that lost write; nothing is skipped.
3. **Finalized-without-notification beats stuck-forever.** For the residual case (callback genuinely never runs, files all terminal), an execution pinned in `EXECUTING` indefinitely is the worse outcome — and staying default-off is exactly why 787 of them accumulated.

The proper long-term fix — re-firing the finalization side effects from the reaper, or gating recovery on "no live queue message/claim for this execution so the callback can't still run" — is the same queue-state check already tracked as a follow-up (it needs PG-queue-table reads the recovery endpoint deliberately doesn't do). Until then, lowering `WORKER_PG_STUCK_EXECUTION_RECOVERY_SECONDS` (e.g. to drain a backlog) is the knob that increases this risk, so that stays a deliberate, documented choice — the default window is conservative.

muhammad-ali-e:
### Review addressed (`89a8d7c2e`)

**greptile P2 — whitespace defeats the kill-switch:** fixed. `.strip()` before the compare, so an emergency `false ` (trailing space/newline from a configmap) still disables recovery. Test covers `false ` and `  FALSE\n`. Worker suite green (16 passed).

**greptile P1 — recovery can bypass finalization side effects:** replied in-thread. It's a real, pre-existing property of the recovery path (status-reconciliation only — no notifications/usage/destination replay), not introduced here; this PR only flips the default. Mitigated by the stuck window defaulting to the barrier stuck-timeout (hours, so it won't race a live callback), and by the fact that the observed strands already ran their side effects — only the status write was lost. Finalized-without-notification also beats stuck-forever (the default-off gate is exactly why the backlog accumulated). Full side-effect replay / queue-state gating is the tracked follow-up.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Addressed in `e104a8f03`. Switched to the codebase convention for default-on flags — `os.getenv(name, "true").strip().lower() == "true"` (matches `CACHE_REDIS_ENABLED` / `ENABLE_METRICS`). This resolves both footguns *and* the convention inversion in one move:
- `.strip()` handles the whitespace/newline.
- `== "true"` means **any** non-`"true"` value disables — `false`/`0`/`no`/`off`/`disabled`/typo — so an operator's emergency rollback reliably turns it off rather than silently failing. (The earlier `!= "false"` / `test_only_false_kills_it` "a typo stays on" behavior is dropped — you're right that reliable-disable matters more, and this is now the only sane reading given the key stays unset in cloud → on by default.)

workers/tests/test_pg_finalization_fixes.py
muhammad-ali-e: Both gaps covered in `e104a8f03`:
- **Sweep gating** — new `TestStuckRecoverySweepGating`: builds the reaper via `__new__`, sets `_stuck_recovery_enabled` False/True + a mock API client, asserts `recover_stuck_pg_executions` is **not** called when off and **is** called (`stuck_seconds=9000`) when on.
- **Constructor wiring** — new `TestReaperConstructorWiring`: patches `stuck_recovery_enabled_from_env` → False, constructs `PgReaper(MagicMock(lease_seconds=30))`, asserts `_stuck_recovery_enabled is False`. So a hardcoded `= True` in `__init__` would now fail.
- Edge cases folded into `test_any_non_true_value_disables` (`0`/`no`/`off`/`disabled`/`""`/typo → False) and `test_disable_is_whitespace_and_case_tolerant`.

workers/tests/test_pg_finalization_fixes.py
muhammad-ali-e: Done in `e104a8f03` — `_flag` now takes `monkeypatch` (delenv + conditional setenv), dropping the empty `patch.dict(..., {}, clear=False)` and the local `import os`. Also trimmed the constructor comment so the enable/kill-switch explanation lives only in the helper docstring (the comment keeps just the stuck-window note that belongs to `_stuck_recovery_seconds`).

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Done — trimmed in `e104a8f03`; the enable/kill-switch explanation now lives only in the helper docstring, the comment keeps just the stuck-window note.

muhammad-ali-e:
### PR Review Toolkit addressed (`e104a8f03`)

**[High] Brittle kill-switch decode** — adopted the codebase convention for default-on flags: `os.getenv(name, "true").strip().lower() == "true"` (same as `CACHE_REDIS_ENABLED` / `ENABLE_METRICS`). Fixes the whitespace footgun *and* the convention inversion, and makes disable reliable — any non-`"true"` value (`false`/`0`/`no`/`off`/typo, with any spacing/case) turns it off, so an emergency rollback can't silently fail.

**[Medium] Coverage gaps** — added the two behavioral tests the parser tests missed:
- `TestStuckRecoverySweepGating` — `_maybe_recover_stuck_executions` skips the API call when the flag is off, makes it when on.
- `TestReaperConstructorWiring` — `__init__` sources the attr from the helper (a hardcoded `= True` would now fail).
- Plus `0`/`no`/`off`/`disabled`/`""`/typo + whitespace/case edge cases.

**[Low] monkeypatch + comment dup** — tests use `monkeypatch`; the constructor comment no longer duplicates the helper docstring.

18 worker tests pass. Thanks — the "parser tested in isolation but not the behavior" point was the valuable catch.

PR #2175, opened 2026-07-14
UN-3718 [FIX] reaper recovery — add missing InternalAPIClient.recover_stuck_pg_executions delegator

muhammad-ali-e:
## What
Add the missing `recover_stuck_pg_executions` delegator to the `InternalAPIClient` facade.

## Why
The reaper's recovery sweep calls `self._get_api_client().recover_stuck_pg_executions(...)`, where `_get_api_client()` returns an **`InternalAPIClient`**. That facade delegates every method to its `ExecutionAPIClient` via an explicit per-method wrapper — but #2172 added `recover_stuck_pg_executions` only to `ExecutionAPIClient` and **never added the facade delegator**. So the sweep threw on every cycle:

```
AttributeError: 'InternalAPIClient' object has no attribute 'recover_stuck_pg_executions'
  File "queue_backend/pg_queue/reaper.py", line 1224, in _maybe_recover_stuck_executions
```

Observed live on `snapshot.964`: the reaper is on the new image, is the leader, and sweeps every 5 min, but the recovery call died before any HTTP request (backend endpoint never hit), so a backlog of 787 EXECUTING+NULL executions was not draining.

## Why the tests missed it
`TestStuckRecoverySweepGating` mocked `_get_api_client()` — a `MagicMock` has every attribute, so the missing real delegator was invisible.

## How
- 13-line delegator on `InternalAPIClient` → forwards to `self.execution_client.recover_stuck_pg_executions(...)`.
- `TestFacadeExposesRecovery` exercises the **real** facade (constructed via `__new__`, mock sub-client) and asserts it delegates — this fails without the fix.

## Tests
19 worker tests pass.

## Deploy
Code bug → merge → rebuild snapshot → redeploy. The fixed reaper then drains the 787 automatically (~100/sweep × 5-min cadence ≈ 40 min; all are past the 2.5h window).

UN-3718 — follow-up to #2172 / #2174.

### DIFF
--- a/workers/tests/test_pg_finalization_fixes.py
+++ b/workers/tests/test_pg_finalization_fixes.py
@@ -186,6 +186,30 @@ class TestStuckRecoverySweepGating:
         api.recover_stuck_pg_executions.assert_called_once_with(stuck_seconds=9000)
 
 
+class TestFacadeExposesRecovery:
+    """The reaper calls recover_stuck_pg_executions on the InternalAPIClient FACADE
+    (what _get_api_client() returns), which must delegate to its ExecutionAPIClient.
+    The mock-based gating test above CANNOT catch a missing delegator (a MagicMock
+    has every attribute) — this pins the real facade wiring, which once shipped an
+    AttributeError to prod because the delegator was never added."""
+
+    def test_facade_delegates_to_execution_client(self):
+        from shared.api.internal_client import InternalAPIClient
+
+        # __new__ (not __init__) → no network/env setup; inject a mock sub-client.
+        client = InternalAPIClient.__new__(InternalAPIClient)
+        client.execution_client = MagicMock()
+        sentinel = object()
+        client.execution_client.recover_stuck_pg_executions.return_value = sentinel
+
+        result = client.recover_stuck_pg_executions(stuck_seconds=600, limit=50)
+
+        client.execution_client.recover_stuck_pg_executions.assert_called_once_with(
+            stuck_seconds=600, limit=50
+        )
+        assert result is sentinel
+
+
 class TestReaperConstructorWiring:
     """__init__ sources _stuck_recovery_enabled from the helper (not a hardcoded
     literal) — pins the wiring so a stray `= True` can't slip through."""
--- a/workers/tests/test_pg_finalization_fixes.py
+++ b/workers/tests/test_pg_finalization_fixes.py
@@ -190,8 +190,9 @@ class TestFacadeExposesRecovery:
     """The reaper calls recover_stuck_pg_executions on the InternalAPIClient FACADE
     (what _get_api_client() returns), which must delegate to its ExecutionAPIClient.
     The mock-based gating test above CANNOT catch a missing delegator (a MagicMock
-    has every attribute) — this pins the real facade wiring, which once shipped an
-    AttributeError to prod because the delegator was never added."""
+    has every attribute) — this pins the real facade wiring; the delegator was
+    missing (added in UN-3718), a real AttributeError the MagicMock-based test can't
+    surface."""
 
     def test_facade_delegates_to_execution_client(self):
         from shared.api.internal_client import InternalAPIClient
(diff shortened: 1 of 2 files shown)


workers/tests/test_pg_finalization_fixes.py
```diff
@@ -186,6 +186,30 @@ def test_enabled_makes_the_recovery_call(self):
         api.recover_stuck_pg_executions.assert_called_once_with(stuck_seconds=9000)
 
 
+class TestFacadeExposesRecovery:
+    """The reaper calls recover_stuck_pg_executions on the InternalAPIClient FACADE
+    (what _get_api_client() returns), which must delegate to its ExecutionAPIClient.
+    The mock-based gating test above CANNOT catch a missing delegator (a MagicMock
+    has every attribute) — this pins the real facade wiring, which once shipped an
+    AttributeError to prod because the delegator was never added."""
```
muhammad-ali-e: **[Comment Analyzer — minor]** The rest of this docstring is verifiable and accurate, but "once shipped an AttributeError **to prod**" is a historical claim the codebase can't substantiate — git history confirms the delegator was missing and later added (UN-3718), but not that it reached production vs. being caught pre-release.

Suggest softening while keeping the valuable "why this test exists" rationale, e.g.:
> …this pins the real facade wiring, which the mock-based test could not — the delegator was missing (added in UN-3718), a real AttributeError the MagicMock-based test cannot surface.

workers/tests/test_pg_finalization_fixes.py
```diff
@@ -186,6 +186,30 @@ def test_enabled_makes_the_recovery_call(self):
...
+        client.execution_client = MagicMock()
+        sentinel = object()
+        client.execution_client.recover_stuck_pg_executions.return_value = sentinel
+
+        result = client.recover_stuck_pg_executions(stuck_seconds=600, limit=50)
+
+        client.execution_client.recover_stuck_pg_executions.assert_called_once_with(
+            stuck_seconds=600, limit=50
+        )
+        assert result is sentinel
```
muhammad-ali-e, line 211: **[Test Analyzer — optional, low priority]** This test always passes both args explicitly, but the only production caller (`reaper.py`) passes just `stuck_seconds` and relies on `limit` defaulting to `None`. That default-forwarding path is never exercised through the facade. Consider a second small test mirroring real usage:

```python
def test_facade_forwards_defaults(self):
    client = InternalAPIClient.__new__(InternalAPIClient)
    client.execution_client = MagicMock()
    client.recover_stuck_pg_executions(stuck_seconds=600)
    client.execution_client.recover_stuck_pg_executions.assert_called_once_with(
        stuck_seconds=600, limit=None
    )
```

Low criticality since forwarding is explicit keyword-to-keyword.

muhammad-ali-e (commented):
**Automated PR Review Toolkit summary** (Code Reviewer, Silent Failure Hunter, Type Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier).

The change itself is clean and correct: the new `recover_stuck_pg_executions` delegator on `InternalAPIClient` is a faithful keyword pass-through to `ExecutionAPIClient.recover_stuck_pg_executions` — signatures match exactly, `APIResponse` return type resolves and is appropriate, and it mirrors the established facade delegator pattern (e.g. `update_workflow_execution_status`). No correctness, type-design, silent-failure, or simplification issues found. Errors from the target raise (they are not swallowed) and are logged by the reaper.

Two minor, non-blocking suggestions are left as inline comments. Both are nice-to-haves, not blockers.

Informational (outside this diff, no action required here):
- `ExecutionAPIClient.recover_stuck_pg_executions` hardcodes `success=True`/`status_code=200` in its `APIResponse.success_response(...)` wrapper; harmless today because the reaper inspects `.data` counters rather than `.success`, but a future caller trusting `.success` could be misled.
- The reaper's empty-payload guard (`if not data`) only catches `None`/`{}`; a truthy-but-non-counter 200 body would skip the guard and silently log zero counters.

workers/tests/test_pg_finalization_fixes.py
muhammad-ali-e: Fixed in `1b8c73631` — softened to your exact suggestion: "the delegator was missing (added in UN-3718), a real AttributeError the MagicMock-based test can't surface." (Fair point: we saw it live in *integration*, but the codebase can't substantiate "prod", so the claim is now what git can back.)

workers/tests/test_pg_finalization_fixes.py
muhammad-ali-e: Added `test_facade_forwards_defaults` in `1b8c73631` — mirrors the real caller (reaper passes only `stuck_seconds`) and asserts `limit=None` is forwarded. Good catch on the untested default path.

muhammad-ali-e:
### PR Review Toolkit addressed (`1b8c73631`)

Both minor comments handled:
- **Comment Analyzer** — softened the `TestFacadeExposesRecovery` docstring: dropped the unverifiable "shipped to prod" phrasing for what git can substantiate ("delegator was missing, added in UN-3718; a real AttributeError the MagicMock-based test can't surface").
- **Test Analyzer** — added `test_facade_forwards_defaults`, exercising the real caller's path (reaper passes only `stuck_seconds`; asserts `limit=None` is forwarded).

20 worker tests pass. The delegator change itself was flagged clean/correct by all six agents — no correctness/type/silent-failure/simplification issues.

PR #2176, opened 2026-07-14
UN-3636 [MISC] enable the unit-workers rig group and de-flake its suite

chandrasekharan-zipstack:
## What

The `unit-workers` rig group filtered `markers: "unit"`, but **no worker test carries that marker** — so it collected zero tests and silently never ran in CI. This switches it to the negative filter the other unit groups use, so all **734 worker tests** actually run and gate.

Enabling collection surfaced pre-existing debt (the suite was never CI-gated). All fixed here — no product code changed.

## Fixes

**Test isolation (parallel/xdist pollution)** — in `workers/tests/conftest.py`:
- Worker Celery apps build a live-Postgres result backend from ambient `DB_*` / `CELERY_BACKEND_DB_*`; stripped before any app import so eager results stay in-memory.
- Building a Celery app hijacks `current_app`; pin celery's finalized default app as current around each test so `@worker_task` proxies resolve (was `NotRegistered` / psycopg2 errors, order-dependent).

**Stale assertions** (prod drifted, tests never caught it since the suite wasn't running):
- `ExecutorToolShim` gained `execution_id`/`organization_id`/`file_execution_id`.
- `_run_agentic_extraction()` gained required `execution_id`.
- Index result now carries `usage_records` from `embedding.flush_pending_usage()`.
- Highlight-disabled test over-asserted "plugin loader untouched" — `lookup-enrichment` is queried unconditionally and returns `None` in OSS; narrowed to assert the highlight plugin specifically.

**Vanity removal** — dropped tautological `*_enum_exists` tests that just mirror the source `Operation` enum (kept the routing/registration/removed-guard tests).

**Renames** — `test_sanity_phase6*` / `test_phaseN*` → intent-revealing names (`git mv`, history preserved).

## Verification

`tox -e groups -- unit-workers` → **734 passed**, stable across repeated parallel runs. `tox -e rig -- validate` OK. Lint clean under the repo's pinned ruff/ruff-format.

### DIFF
--- a/tests/groups.yaml
+++ b/tests/groups.yaml
@@ -44,7 +44,9 @@ groups:
     tier: unit
     workdir: workers
     paths: [tests, shared/tests]
-    markers: "unit"
+    # Negative filter like the other unit groups: worker tests aren't tagged
+    # `unit`; live-infra tests carry `integration`/`slow` and run elsewhere.
+    markers: "not integration and not slow"
     uv_sync_group: test
     coverage_source: shared
 
--- a/tests/groups.yaml
+++ b/tests/groups.yaml
@@ -17,8 +17,7 @@ defaults:
 groups:
   # ── Unit tier: pure pytest, no external services ───────────────────────────
   unit-rig:
-    # Self-tests for the rig's pure logic (dep-graph, evaluate, junit parse).
-    # Run from the repo root so the `tests.rig` package import resolves.
+    # Rig self-tests. Run from repo root so the `tests.rig` import resolves.
     tier: unit
     workdir: .
     paths: [tests/rig/tests]
@@ -53,20 +52,17 @@ groups:
   unit-backend:
     tier: unit
     workdir: backend
-    # Pure backend tests — no DB. Collect the whole tree and let markers, not a
-    # hand-kept file list, decide membership: tests needing live infra carry
-    # `@pytest.mark.integration` (see integration-backend) and are excluded here.
+    # Marker-selected, not a hand-kept file list: live-infra tests carry
+    # `integration` and run in integration-backend.
     paths: ["."]
     markers: "not integration"
     uv_sync_group: test
-    # Anchored: integration-backend reuses the identical Django settings env so
-    # the two halves of the backend suite can't drift apart.
+    # Shared with integration-backend so both halves use identical settings.
     env: &backend_test_env
       DJANGO_SETTINGS_MODULE: backend.settings.test
       # Tenancy is row-level, not schema-per-tenant; tests run in public.
       DB_SCHEMA: public
-      # base.py resolves these at import time with no default; supply test-safe
-      # values here.
+      # Resolved at import with no default; supply test-safe values.
       DJANGO_SECRET_KEY: test-secret-key-not-for-production
       # All-zero Fernet key: valid format, obviously a placeholder.
       ENCRYPTION_KEY: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
@@ -81,8 +77,7 @@ groups:
       SYSTEM_ADMIN_USERNAME: admin
       SYSTEM_ADMIN_PASSWORD: admin
       SYSTEM_ADMIN_EMAIL: admin@example.com
-      # ExecutionFileHandler builds execution-dir paths from this at init; only
-      # constructed, never written to in these tests.
+      # Only used to build paths at init; never written in these tests.
       WORKFLOW_EXECUTION_DIR_PREFIX: /tmp/unstract-workflow-exec
     coverage_source: .
 
@@ -108,13 +103,9 @@ groups:
   integration-backend:
     tier: integration
     workdir: backend
-    # Backend ORM tests — need live infra. The rig provisions the declared
-    # requires_services via testcontainers and injects their connection env into
-    # pytest (tests/rig/cli.py:_inject_infra_env). Not optional: these gate the
-    # integration tier.
-    # Marker-only selection — exact complement of unit-backend: conftest.py
-    # auto-marks DB-bound tests `integration`, so new DB tests join this
-    # group with no edit here. Tests needing external creds skip without them.
+    # Complement of unit-backend: conftest auto-marks DB-bound tests
+    # `integration`, so new ones join here with no edit. The rig provisions
+    # requires_services and injects their env; cred-gated tests skip without.
     paths: ["."]
     markers: "integration"
     uv_sync_group: test
@@ -127,10 +118,8 @@ groups:
     workdir: unstract/connectors
     paths: [tests]
     markers: "integration"
-    # Most connector integration tests need real third-party credentials
-    # (Snowflake, GDrive, Box, Dropbox, …) and skip when those are absent. The
-    # MinIO test actually runs: the rig provisions MinIO via testcontainers and
-    # injects MINIO_* creds (tests/rig/cli.py).
+    # Most need real third-party creds and skip when absent; MinIO runs
+    # because the rig provisions it and injects the creds.
     requires_services: [minio]
     uv_sync_group: test
     coverage_source: src
@@ -143,12 +132,8 @@ groups:
     optional: true   # placeholder until tests are authored
 
   # ── E2E tier: real HTTP against running platform ───────────────────────────
-  # Every group below has `requires_platform: true` and must (transitively)
-  # depend on `e2e-smoke` — the rig's `_validate_platform_groups_depend_on_gate`
-  # enforces this structural invariant so the manifest can't ship a platform
-  # test that bypasses the smoke gate, and so smoke runs first in topo order.
-  # (Runtime skip-on-gate-failure is a future enhancement; see that function's
-  # docstring and the TODO in cmd_run.)
+  # Platform groups must depend (transitively) on e2e-smoke; the rig enforces
+  # this so smoke gates and runs first. Nothing ships bypassing it.
   e2e-smoke:
     tier: e2e
     paths: [tests/e2e/smoke]
@@ -157,8 +142,7 @@ groups:
     timeout_seconds: 1200
     depends_on: [unit-sdk1, unit-workers]
 
-  # Not optional: login is the first e2e critical path we gate on. A red login
-  # test fails the build and (on main) surfaces as an in-scope critical gap.
+  # Not optional: login is a gated critical path; red fails the build.
   e2e-login:
     tier: e2e
     paths: [tests/e2e/auth]
--- a/tests/groups.yaml
+++ b/tests/groups.yaml
@@ -44,7 +44,8 @@ groups:
     workdir: workers
     paths: [tests, shared/tests]
     # Negative filter like the other unit groups: worker tests aren't tagged
-    # `unit`; live-infra tests carry `integration`/`slow` and run elsewhere.
+    # `unit`. Tests tagged `integration`/`slow` are excluded here but not yet
+    # routed to any group — add one before relying on those markers.
     markers: "not integration and not slow"
     uv_sync_group: test
     coverage_source: shared
--- a/workers/tests/conftest.py
+++ b/workers/tests/conftest.py
@@ -6,9 +6,45 @@ shared/constants/api_endpoints.py raises ValueError at import
 time if INTERNAL_API_BASE_URL is not set.
 """
 
+import os
 from pathlib import Path
 
+import pytest
 from dotenv import load_dotenv
 
 _env_test = Path(__file__).resolve().parent.parent / ".env.test"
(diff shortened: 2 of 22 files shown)


tests/groups.yaml
```diff
@@ -44,27 +43,26 @@ groups:
     tier: unit
     workdir: workers
     paths: [tests, shared/tests]
-    markers: "unit"
+    # Negative filter like the other unit groups: worker tests aren't tagged
+    # `unit`; live-infra tests carry `integration`/`slow` and run elsewhere.
+    markers: "not integration and not slow"
```
athul-rs, line 49: **The mechanism that hid this bug is still in place.**

Flipping the marker fixes *this* group, but not the reason a zero-collect group reported green. `tests/rig/cli.py:52` has `_NON_FAILING_PYTEST_EXIT_CODES = (0, 5)`, and pytest exit 5 means "no tests collected". The aggregation at `cli.py:481-486`:

```python
if (
    exit_code not in _NON_FAILING_PYTEST_EXIT_CODES   # exit 5 -> False, short-circuits
    and not group.optional
    and overall_exit == 0
):
```

The `not group.optional` guard is never reached for exit 5, so a **non-optional** group that collects nothing folds in as a pass. That is precisely what `unit-workers` did. `tests/rig/` is untouched by this PR, so the next marker typo, renamed marker, or moved directory goes green the same way.

Worth knowing: the rig already has a guard built for exactly this. `_coverage_attesting_groups()` (`cli.py:664`) excludes `empty` groups from attesting coverage, with the comment *"a broken marker expression would otherwise report OK with zero tests run"*. It never fired here only because it applies to groups named in `critical_paths.yaml` `covered_by`, and `unit-workers` is in none of them.

Cheapest close: add `unit-workers` to a critical path's `covered_by` so the existing gate does its job. Note a blanket `_NON_FAILING_PYTEST_EXIT_CODES = (0,)` is **not** safe — hurl groups synthesize `exit 5` at `cli.py:897`, and `--marker`/`--paths` overrides legitimately zero-collect.

Happy for this to be a follow-up ticket if you'd rather keep the PR scoped, but as it stands UN-3636 fixes the instance and leaves the class open.

tests/groups.yaml
```diff
@@ -44,27 +43,26 @@ groups:
     tier: unit
     workdir: workers
     paths: [tests, shared/tests]
-    markers: "unit"
+    # Negative filter like the other unit groups: worker tests aren't tagged
+    # `unit`; live-infra tests carry `integration`/`slow` and run elsewhere.
```
athul-rs: For workers, there is no "elsewhere" — this comment points at a group that doesn't exist.

No group in this manifest runs workers `integration`/`slow` tests. `integration-workflow-execution` is root-workdir (`paths: [tests/integration/workflow_execution]`), optional, and its path isn't in the tree, so it's skipped as a missing placeholder. Both markers *are* registered in `workers/pyproject.toml:173-177`, so `--strict-markers` won't stop anyone from reaching for them — and a worker test marked `@pytest.mark.integration` would be excluded here and collected by **no group at all**.

Not hypothetical: `unit-sdk1` already uses the same negative filter, and `unstract/sdk1/tests/test_platform.py:106` and `:175` carry `@pytest.mark.slow` today. Those two tests are currently run by nothing.

This is the pre-existing manifest convention rather than something you introduced, so I wouldn't block on it — but the comment as written actively invites the next person into the gap. Either reword to say the markers are currently unrouted, or add the group that claims them.

workers/tests/test_plugin_migration_regression.py
```diff
@@ -553,15 +553,17 @@ def _build_log_component(self, operation, executor_params=None):
...
+                "operation": operation,
+                "run_id": "run-log",
+                "execution_source": "tool",
+                "executor_params": params,
+                "request_id": "req-1",
+                "log_events_id": "evt-1",
+            }
+        )
 
         # Replicate tasks.py logic
```
athul-rs: **These are the tests worth deleting — and one is asserting behaviour production doesn't have.**

`_build_log_component` copies the `tasks.py` if/elif into the test body and asserts on its own copy. Nothing in this class imports `executor.tasks`, so no change to `tasks.py` can ever fail it. Same pattern in `test_table_extract_operation.py:17`, `test_smart_table_extract_operation.py:19`, `test_simple_prompt_studio_operation.py:214`, `test_agentic_operations.py:239`.

Worse: the replica has **drifted from the original and stays green**. Real `tasks.py:79-88`:

```python
if context.operation == "ide_index":
    index_params = params.get("index_params", {})
    extract_params = params.get("extract_params", {})
    usage_kwargs = extract_params.get("usage_kwargs", {})
    "tool_id":  index_params.get("tool_id", ""),
    "doc_name": str(usage_kwargs.get("file_name", "")),
```

The replica below (L569-575) reads `extract_params["tool_id"]` and `extract_params["file_name"]`. So `test_ide_index_extracts_nested_params` (L594) feeds `{"extract_params": {"tool_id": "t-nested", "file_name": "nested.pdf"}}` and asserts `tool_id == "t-nested"` — but that same payload through the real `tasks.py` yields `tool_id=""` and `doc_name=""`. The test passes while asserting the opposite of production. That's negative signal, not merely zero.

The right pattern already exists in this suite: `test_log_streaming.py:241-293` imports the real `execute_extraction`, patches `ExecutionOrchestrator`, and asserts on `ctx._log_component`. It covers the default and empty-`log_events_id` branches — and the only two branches with *no* real coverage anywhere are `ide_index` and `structure_pipeline`, i.e. exactly the ones the replica fakes and gets wrong.

Since this PR is explicitly pruning vanity tests: these are the vanity tests. Either extend the `test_log_streaming.py` pattern to the two nested branches, or delete the replicas.

workers/tests/test_completion_and_highlight.py
```diff
@@ -432,8 +433,15 @@ def test_highlight_skipped_when_disabled(
...
-        mock_plugin_get.assert_not_called()
+        # Highlight disabled → highlight plugin must not be fetched. Other plugin
+        # lookups (e.g. lookup-enrichment) run unconditionally and return None in
+        # OSS, so assert on the highlight key rather than "loader untouched".
+        highlight_fetches = [
+            c
+            for c in mock_plugin_get.call_args_list
+            if c.args and c.args[0] == "highlight-data"
+        ]
+        assert highlight_fetches == []
```
athul-rs, line 444: Nit, but on-theme: this assertion stops asserting if the call style ever changes.

`c.args and c.args[0] == "highlight-data"` only matches **positional** calls. `legacy_executor.py:1394` does `ExecutorPluginLoader.get("highlight-data")` today, so it works. But if that ever becomes `get(name="highlight-data")`, `c.args` is empty, `highlight_fetches` is `[]`, and `assert highlight_fetches == []` passes vacuously. A test that silently stops testing is the same failure class this PR exists to fix.

```python
assert not [
    c
    for c in mock_plugin_get.call_args_list
    if "highlight-data" in c.args or c.kwargs.get("name") == "highlight-data"
]
```

athul-rs (commented):
Reviewed the marker fix, the conftest isolation work, the stale-assertion fixes, and the deletions/renames. Net: the direction is right and this is a strict improvement — enabling 734 previously-dead tests is a clear win, and both conftest mechanisms (env strip + `current_app` pinning) are correct as written.

I ran an adversarial pass over my own findings before posting, and **three didn't survive**. Recording them so nobody re-litigates:

- **The `*_enum_exists` deletions are fine.** I first read them as the only pin on sdk1's `Operation` values — they aren't. `test_plugin_migration_regression.py:131-176` already pins the whole set bidirectionally (`test_every_operation_is_mapped`, `test_no_extra_mappings`, `len(Operation) == 19`). Genuine dedup, and sdk1 is an in-repo editable path dep, not a versioned external provider. Agreed with the call.
- **`shared/tests` is *not* missed by the env strip.** `os.environ.pop()` is process-global and both paths run in one pytest process with conftests loaded before any test executes. Only the autouse `current_app` fixture doesn't reach `shared/tests`, and nothing there touches celery — latent, zero impact.
- **The env strip is the right mechanism, not a workaround** — it's the only thing that handles an ambient exported `DB_HOST` and the only thing covering `CELERY_BACKEND_DB_*`. See the conftest note.

Four comments left. Two I think are worth acting on:

1. **The rig still scores "collected zero tests" as a pass for non-optional groups.** This PR fixes the instance, not the class — the next marker typo goes green the same way.
2. **The `log_component` replica tests are the real tautologies** — and one is currently green while asserting the *opposite* of what production does.

Plus two nits. Nothing blocking; CI is green.

workers/tests/conftest.py
```diff
@@ -6,9 +6,45 @@
...
 from pathlib import Path
 
+import pytest
 from dotenv import load_dotenv
 
 _env_test = Path(__file__).resolve().parent.parent / ".env.test"
 load_dotenv(_env_test)
+
+# Worker Celery apps build a Postgres result backend from DB_*/CELERY_BACKEND_DB_*.
+# Strip these before any app is imported so tests don't reach (or leak) a live DB
```
athul-rs, line 19: Nit / maintainability: the root conftest does the exact opposite of this block, four lines at a time.

`workers/conftest.py:15-18` (which also covers `shared/tests`):

```python
os.environ.setdefault("DB_HOST", "localhost")
os.environ.setdefault("DB_USER", "test")
os.environ.setdefault("DB_PASSWORD", "test")
os.environ.setdefault("DB_NAME", "testdb")
```

Two adjacent conftests with opposing intent on the same four vars, resolving correctly only because pytest loads the root one first.

To be clear, **the pop is the right mechanism — keep it.** It's the only thing that neutralises an ambient exported `DB_HOST` (which `setdefault` would happily leave pointing at a live DB, and the rig copies the ambient env into every group subprocess at `tests/rig/cli.py:83`), and the only thing covering `CELERY_BACKEND_DB_*`, which the root conftest never sets and `worker_config.py:129-152` checks *first*.

But those four `setdefault`s are now dead weight during the group run, and the next reader will reasonably conclude a DB is configured. Worth dropping them **in addition to** this pop (never instead of), or at minimum cross-referencing between the two files.

tests/groups.yaml
chandrasekharan-zipstack: Reworded in f66b9dbd — the note now says integration/slow worker tests are excluded here but **not yet routed to any group**, instead of implying they run elsewhere.

workers/tests/test_plugin_migration_regression.py
chandrasekharan-zipstack: Agreed — these were the vanity tests. Deleted the replica class and replaced it with tests that drive the real `execute_extraction` (orchestrator patched out), covering the two special-cased branches plus default. The `ide_index` case now feeds `index_params` + `extract_params.usage_kwargs` and asserts what production actually builds, so drift in `tasks.py` fails here instead of a private copy. f66b9dbd.

workers/tests/test_completion_and_highlight.py
chandrasekharan-zipstack: Fixed in f66b9dbd — assertion now matches `"highlight-data" in c.args or c.kwargs.get("name") == "highlight-data"`, so a kwarg-style call cannot pass it vacuously.

workers/tests/conftest.py
chandrasekharan-zipstack: Kept the pop and cross-referenced both conftests (f66b9dbd). One correction: those `setdefault`s are not dead during the group run — the pop only covers `workers/tests/`, so `shared/tests/` (no local conftest) still relies on them. Dropping them would break shared/tests, so I left them and added the cross-reference instead.

tests/groups.yaml
chandrasekharan-zipstack: Good catch on the class-level gap. The fix lives in `tests/rig/` (exit-5 handling), which this PR deliberately does not touch, and as you note a blanket `(0,)` is unsafe given hurl’s synthesized exit 5 and legitimate marker/path zero-collects. Taking your offer to track it as a follow-up rather than widen scope — will file a ticket for a rig-level empty-collect guard on non-optional groups.

chandrasekharan-zipstack: Decided to close the class here rather than defer. `2a1acc63` adds a runtime gate: a **non-optional pytest** group that collects zero tests (exit 5) now fails `overall_exit` instead of folding in green. Exemptions match the legit zero-collects you called out — `optional` groups, hurl (its synthesized exit 5 = "no files"), and dev runs with a `--marker`/`--paths` override. So `_NON_FAILING_PYTEST_EXIT_CODES` stays `(0, 5)` for those; the new guard only fires for the misconfiguration case. Two rig self-tests added (`test_empty_nonoptional_group_fails_build`, `test_empty_group_with_marker_override_does_not_fail`).

athul-rs (approved):

PR #2177, opened 2026-07-14
UN-3719 [FIX] reaper poison orphan-claims — 404 (not 500) for a deleted execution → GC the claim

muhammad-ali-e:
## What
The reaper's orphan-claim sweep can now GC claims whose execution was deleted, instead of failing on them every cycle.

## Why
`pg_orchestration_claim` has **no FK** to `workflow_execution`, so retention/cleanup that deletes old executions leaves orphaned claim rows. The sweep reads a claim's `execution_id` and looks it up (in `workflow_execution`) to decide GC-vs-recover — but the backend `retrieve` swallowed `Http404` in a bare `except Exception` and returned **500**. The reaper's `_execution_status` treats any failure as *transient* (raises → retain the claim), so a permanently-deleted execution poisoned every sweep:

```
RuntimeError: status read failed for execution 64d791a2… (refusing to mark ERROR on an unconfirmed status)
Reaper: orphan-claim sweep — 3 of 4 row(s) failed (left for the next sweep); 1 swept.
```

Observed live in integration: 3 claim rows (dated Apr–Oct) whose execution + file-execution rows are gone (0 rows), failing every 5-min sweep.

## How
1. **Backend** `retrieve()`: catch `Http404` / `WorkflowExecution.DoesNotExist` → **404** (before the generic 500).
2. **Client plumbing**: `APIRequestError` carries `status_code`; `get_workflow_execution` propagates it into `ExecutionResponse.status_code`.
3. **Reaper**: a confirmed **404** → `_EXECUTION_GONE` sentinel → GC the orphan claim (re-guarded on still-orphan-and-old). A **5xx / transport** failure still raises → claim retained (unchanged — never terminalize on an unconfirmed read).

## Safety
- The 404-GC only removes the queue-infra claim tombstone, never touches an execution/files, and re-guards on `_orphan_claim_where` (old + still-orphan) before deleting.
- Transient-vs-gone is a *deterministic HTTP status*, not string matching.
- Returning 404 for a missing execution is strictly more correct for every caller of the endpoint.

## Tests
- Backend: `retrieve` returns 404 for a missing execution.
- Worker: `_execution_status` → GONE on 404, raises on 5xx/transport, returns status on success; `_recover_one_claim` GC's a gone execution and leaves it on a re-claim race.
- Worker 24 + backend 17 green.

UN-3719 — follow-up to #2172 / #2174 / #2175.

### DIFF
--- a/backend/workflow_manager/internal_views.py
+++ b/backend/workflow_manager/internal_views.py
@@ -6,6 +6,7 @@ import logging
 import uuid
 
 from django.db import transaction
+from django.http import Http404
 from django.shortcuts import get_object_or_404
 from rest_framework import status, viewsets
 from rest_framework.decorators import action
@@ -148,6 +149,20 @@ class WorkflowExecutionInternalViewSet(viewsets.ReadOnlyModelViewSet):
             serializer = WorkflowExecutionContextSerializer(context_data)
             return Response(serializer.data)
 
+        except (Http404, WorkflowExecution.DoesNotExist):
+            # A deleted / never-existent execution is a deterministic 404, not a 500.
+            # The reaper's orphan-claim sweep depends on this: it must distinguish
+            # "execution gone → GC the orphan claim" from a transient server error
+            # ("retain, retry next sweep"). Returning 500 here made the reaper retry
+            # forever and never GC claims for deleted executions.
+            logger.info(
+                f"WorkflowExecution {kwargs.get('pk') or kwargs.get('id')} not found "
+                "(retrieve) — returning 404"
+            )
+            return Response(
+                {"error": "WorkflowExecution not found"},
+                status=status.HTTP_404_NOT_FOUND,
+            )
         except Exception as e:
             logger.error(
                 f"Failed to retrieve workflow execution {kwargs.get('id')}: {str(e)}"
--- a/backend/workflow_manager/internal_views.py
+++ b/backend/workflow_manager/internal_views.py
@@ -150,18 +150,30 @@ class WorkflowExecutionInternalViewSet(viewsets.ReadOnlyModelViewSet):
             return Response(serializer.data)
 
         except (Http404, WorkflowExecution.DoesNotExist):
-            # A deleted / never-existent execution is a deterministic 404, not a 500.
-            # The reaper's orphan-claim sweep depends on this: it must distinguish
-            # "execution gone → GC the orphan claim" from a transient server error
-            # ("retain, retry next sweep"). Returning 500 here made the reaper retry
-            # forever and never GC claims for deleted executions.
-            logger.info(
-                f"WorkflowExecution {kwargs.get('pk') or kwargs.get('id')} not found "
-                "(retrieve) — returning 404"
+            # A deleted execution must return 404 (deterministic) so the reaper's
+            # orphan-claim sweep can GC it, instead of 500 (which it treats as
+            # transient → retry forever). BUT get_object() is org-scoped and this try
+            # wraps the whole method, so an Http404 here can also mean "exists but
+            # scoped out of this request" or a nested Http404 for a *different*
+            # resource — NOT "execution deleted". Since the reaper turns a genuine 404
+            # into an irreversible claim GC, confirm the row is truly absent (unscoped)
+            # before saying not-found; otherwise 500 so the reaper retains the claim.
+            execution_id = kwargs.get("pk") or kwargs.get("id")
+            if not WorkflowExecution.objects.filter(id=execution_id).exists():
+                logger.info(
+                    f"WorkflowExecution {execution_id} not found (retrieve) — 404"
+                )
+                return Response(
+                    {"error": "WorkflowExecution not found"},
+                    status=status.HTTP_404_NOT_FOUND,
+                )
+            logger.warning(
+                f"WorkflowExecution {execution_id} exists but was not retrievable in "
+                "this context (org scope / nested lookup) — returning 500 (retain)"
             )
             return Response(
-                {"error": "WorkflowExecution not found"},
-                status=status.HTTP_404_NOT_FOUND,
+                {"error": "Failed to retrieve workflow execution"},
+                status=status.HTTP_500_INTERNAL_SERVER_ERROR,
             )
         except Exception as e:
             logger.error(
--- a/workers/queue_backend/pg_queue/reaper.py
+++ b/workers/queue_backend/pg_queue/reaper.py
@@ -359,7 +359,17 @@ def _execution_status(
         execution_id, organization_id=organization_id, file_execution=False
     )
     if not getattr(response, "success", False):
-        if getattr(response, "status_code", None) == 404:
+        # GONE only when the backend itself confirms the row is absent: require BOTH
+        # the 404 status AND the app-level marker in the body. A bare 404 from a
+        # proxy / gateway / rolling-deploy URL-or-version skew would 404 *every* read
+        # at once — treating that as "deleted" would GC every orphan claim in a
+        # single sweep. And an org-scoped 404 for an execution that still exists is
+        # likewise not "gone". The dual signal gates GC to the real thing.
+        if getattr(
+            response, "status_code", None
+        ) == 404 and "WorkflowExecution not found" in (
+            getattr(response, "error", "") or ""
+        ):
             return _EXECUTION_GONE
         raise RuntimeError(
             f"status read failed for execution {execution_id} "
@@ -485,7 +495,18 @@ def _recover_one_barrier(
         return False
 
     status = _execution_status(api_client, execution_id, organization_id)
-    if status is None:
+    if status is _EXECUTION_GONE:
+        # The execution row was deleted — there is nothing to mark ERROR. Fall through
+        # to the cleanup below and GC the orphaned barrier / dedup rows (same "no
+        # status overwrite" path as an already-terminal execution). Without this
+        # branch, ExecutionStatus.is_completed(_EXECUTION_GONE) would raise
+        # ValueError, churn every sweep, and could trip the systemic-failure guard.
+        logger.info(
+            "Reaper: barrier for deleted execution %s — cleaning up the orphaned "
+            "row only (no status overwrite).",
+            execution_id,
+        )
+    elif status is None:
         # A successful read with no status is anomalous — don't mark on it; leave
         # the row for the next sweep rather than risk a wrong ERROR.
         logger.warning(
@@ -494,7 +515,7 @@ def _recover_one_barrier(
             execution_id,
         )
         return False
-    if ExecutionStatus.is_completed(status):
+    elif ExecutionStatus.is_completed(status):
         logger.warning(
             "Reaper: barrier for execution %s expired but the execution is already "
             "%s — cleaning up the orphaned row only (no status overwrite).",
--- a/workers/queue_backend/pg_queue/reaper.py
+++ b/workers/queue_backend/pg_queue/reaper.py
@@ -112,6 +112,12 @@ _CLAIM_RECOVERED: Final = "recovered"  # execution marked ERROR (crash-window)
 _CLAIM_GC: Final = "gc"  # terminal execution's tombstone deleted
 _ClaimOutcome = Literal["recovered", "gc"]
 
+# Sentinel: _execution_status returns this when the execution row no longer exists
+# (a deterministic 404 — deleted by retention/cleanup). Distinct from None (a
+# readable execution with no status) and from a status string, so _recover_one_claim
+# can GC the pure-orphan claim instead of retaining it forever on an unreadable read.
+_EXECUTION_GONE: Final = object()
+
 
 class LeaderLeaseLike(Protocol):
     """Structural contract :class:`PgReaper` needs from a lease.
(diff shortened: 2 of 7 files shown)


workers/queue_backend/pg_queue/reaper.py
```diff
@@ -112,6 +112,12 @@
 _CLAIM_GC: Final = "gc"  # terminal execution's tombstone deleted
 _ClaimOutcome = Literal["recovered", "gc"]
 
+# Sentinel: _execution_status returns this when the execution row no longer exists
+# (a deterministic 404 — deleted by retention/cleanup). Distinct from None (a
+# readable execution with no status) and from a status string, so _recover_one_claim
+# can GC the pure-orphan claim instead of retaining it forever on an unreadable read.
+_EXECUTION_GONE: Final = object()
```
muhammad-ali-e, line 119: **[P1] Second caller of `_execution_status` doesn't handle `_EXECUTION_GONE` → misleading `ValueError`, barrier path never GC'd.**

`_execution_status` now has two callers, but only `_recover_one_claim` (line 725) learned about the new sentinel. `_recover_one_barrier` (line 487) still guards only `status is None` and then calls `ExecutionStatus.is_completed(status)` (line 497). For a deleted execution that still has a stranded barrier row, `status` is `_EXECUTION_GONE` (a bare `object()`), so `is_completed()` does `ExecutionStatus(<object>)` → `ValueError: Invalid status: <object …>`. The per-row `except Exception` (line 588) catches it, rolls back, increments `failed`, and retains the row — so the barrier for a gone execution churns every sweep and can trip the "recovered nothing → systemic failure" escalation.

This isn't a fresh regression (pre-PR that path raised `RuntimeError` and also retained), but the fix is applied to the claim path and left uncovered on the sibling barrier path, and the failure now surfaces as a misleading exception rather than the deliberate one.

Root cause is the type: `_execution_status(...) -> str | object | None` (line 344) — `object` supersets `str` and `None`, so the union collapses to `object` and no checker flags the second call site. Fix both together:
1. Make the sentinel a closed type so call sites are exhaustively checkable, e.g. a single-member `Enum` (`str | _StatusRead | None` does not collapse). Note `enum` is **not** currently imported in this module.
2. Handle `status is _EXECUTION_GONE` in `_recover_one_barrier` before the `is_completed` call — delete the stranded barrier / `pg_batch_dedup` rows for the gone execution (mirror the claim-path GC), or at minimum skip cleanly with a clear log.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -335,20 +341,26 @@ def rearm_expired_claims(conn: PgConnection) -> int:
...
+    ``status=None`` — not terminal — and the caller would then mark a possibly-
+    COMPLETED execution ERROR. Treating a transient "couldn't read" as a hard stop
+    (the caller's ``except`` retains the row for retry) keeps the terminal-skip guard
+    honest; only a confirmed 404 is safe to act on.
     """
     response = api_client.get_workflow_execution(
         execution_id, organization_id=organization_id, file_execution=False
     )
     if not getattr(response, "success", False):
+        if getattr(response, "status_code", None) == 404:
```
muhammad-ali-e: **[P1] Extends @greptile-apps (org-scoped 404): a bare HTTP 404 is overloaded beyond "row deleted".**

Greptile already flagged that an org-scope mismatch 404s a live execution here. The same `status_code == 404`-only check also fires for 404s that have nothing to do with the execution being gone: a rolling-deploy URL/API-version skew, a reverse proxy / load balancer, or an auth gateway. Any of these → `_EXECUTION_GONE` → `_delete_orphan_claim`. Blast radius is large: during a bad deploy *every* status read 404s, so *every* orphan claim is GC'd in a single sweep.

The fix data is already on the response: the backend returns a distinguishing body `{"error": "WorkflowExecution not found"}` (internal_views.py:163), and `execution_client` puts the 404 body into `response.error` (execution_client.py:140), which survives the `to_dict()` round-trip onto the object read here. Require **both** signals before acting:
```python
if not getattr(response, "success", False):
    if (getattr(response, "status_code", None) == 404
            and "WorkflowExecution not found" in (getattr(response, "error", "") or "")):
        return _EXECUTION_GONE
    raise RuntimeError(...)  # any other 404 (route/proxy/auth) is transient → retain
```
A bare 404 with no app-level marker is exactly the "unconfirmed status" the docstring says must be retained. (Minor: `HTTPStatus.NOT_FOUND` reads better than the literal `404` and compares equal.)

backend/workflow_manager/internal_views.py
```diff
@@ -148,6 +149,20 @@ def retrieve(self, request, *args, **kwargs):
             serializer = WorkflowExecutionContextSerializer(context_data)
             return Response(serializer.data)
 
+        except (Http404, WorkflowExecution.DoesNotExist):
```
muhammad-ali-e, line 152: **[P1/P2] `retrieve` emits an ambiguous 404 — root-cause backend fix for the org-scoped-404 issue.**

Two problems make this 404 mean more than "row deleted", and the reaper turns any such 404 into an irreversible claim GC:
1. **Org scoping.** `get_object()` runs against the org-filtered queryset, and `resolve_organization` returns `queryset.none()` when the claim's `organization_id` doesn't resolve (or is stored in the wrong PK-vs-string format) → `get_object_or_404` raises `Http404` for an execution that still exists. Before returning 404, confirm the row is genuinely gone irrespective of org scope:
```python
if not WorkflowExecution.objects.filter(id=kwargs.get('id')).exists():
    return Response({"error": "WorkflowExecution not found"}, status=404)
# else the row exists but is scoped out / another failure → 500 (retain), not 404
```
2. **Over-broad `try`.** This `except` wraps the entire method body, so a downstream `Http404` (e.g. a nested `get_object_or_404` in `_get_source_config` / `_get_destination_config`) is masked as "WorkflowExecution not found" → false 404 → GC of a live execution's claim. Narrow the `try` to just the `get_object()` call and run the context-building outside it, so downstream failures surface as 500 (retain).

Also: `WorkflowExecution.DoesNotExist` in the tuple is unreachable — DRF's `get_object()` raises `Http404`, never the bare model exception; drop it. And add `organization_id` to the log line so a wrongful GC is traceable.

workers/shared/clients/base_client.py
```diff
@@ -86,9 +86,16 @@ class AuthenticationError(InternalAPIClientError):
 
 
 class APIRequestError(InternalAPIClientError):
-    """Raised when API request fails."""
+    """Raised when an API request fails.
 
-    pass
+    ``status_code`` carries the HTTP status when the failure was an HTTP error
```
muhammad-ali-e: **[P2] `APIRequestError` docstring overstates the `status_code` contract (rot risk).**

The docstring presents a clean dichotomy — HTTP error response → carries a code; transport failure → `None`. But several branches raise `APIRequestError` **without** a `status_code` even though they are HTTP-level or request-level failures: rate-limit exhaustion (429, ~line 384), the catch-all "Other errors" branch (~line 396), and data-serialization failure (~line 325). So an HTTP error can also yield `status_code=None`.

This doesn't affect the reaper (404 is always populated on the client-error branch), but a future caller who trusts "any HTTP error carries a code" would be misled. Suggest narrowing the wording to: *"carries the HTTP status for a non-timeout 4xx or a retry-exhausted 5xx (e.g. 404 / 500); `None` for transport failures and for a few paths that don't thread it (429-exhausted, serialization, generic fallback). Only rely on it for those 4xx/5xx cases."*

(Related maintainability note, no change required: this reuses `BaseResponse.status_code`, which elsewhere defaults to `200`/`400` and means "this response's HTTP status" — here on the error path it means "the upstream call's status, or `None`." Same field, two null-semantics; worth a one-line note on the field, or a distinct name like `upstream_status_code`.)

workers/shared/clients/execution_client.py
```diff
@@ -134,10 +134,13 @@ def get_workflow_execution(
                 message="Successfully retrieved workflow execution",
             )
         except Exception as e:
+            # Preserve the HTTP status (404 vs 5xx) so callers — the reaper's
+            # orphan-claim sweep — can tell "execution deleted" from a transient blip.
             return ExecutionResponse.error_response(
                 error=str(e),
                 execution_id=str(execution_id),
                 message="Failed to retrieve workflow execution",
+                status_code=getattr(e, "status_code", None),
```
muhammad-ali-e, line 143: **[P2] The status_code propagation chain this feature depends on has no test coverage.**

Every 404 test (`test_reaper_orphan_claim_404.py`) hand-sets `.status_code` on a `MagicMock`; none constructs a real `ExecutionResponse`, raises a real `APIRequestError`, or calls this method. So the three cooperating production pieces are untested end-to-end:
- `base_client.py:345,366` — `APIRequestError(..., status_code=response.status_code)`
- this line — `error_response(..., status_code=getattr(e, "status_code", None))`
- `response_models.py:220` — the new `status_code` param

A rename of `APIRequestError.status_code`, or a change to the `getattr` attribute name here, leaves every unit test green while `_execution_status` sees `status_code=None` on a real 404 → raises → the reaper is back to retrying deleted-execution claims forever (the exact bug this PR fixes). Add unit tests: (a) `APIRequestError("m", status_code=404).status_code == 404` and default `None`; (b) patch the HTTP layer to raise `APIRequestError(status_code=404/500)` and assert the returned `ExecutionResponse.status_code` matches, and a plain `Exception` → `None`; (c) a `base_client` test that a 404 response raises `APIRequestError` whose `.status_code == 404`.

workers/tests/test_reaper_orphan_claim_404.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def test_success_returns_status(self):
+        assert self._call(_resp(True, status="COMPLETED")) == "COMPLETED"
+
+    def test_404_returns_gone_sentinel(self):
+        from queue_backend.pg_queue.reaper import _EXECUTION_GONE
+
+        assert self._call(_resp(False, status_code=404)) is _EXECUTION_GONE
+
+    def test_500_raises_transient(self):
```
muhammad-ali-e, line 64: **[P2] Add a non-404 4xx case to pin the "only 404 is GONE" boundary.**

The suite covers 404 (GONE), 500 (raise), and transport/`None` (raise), but never a non-404 4xx such as 403/401/409. `_execution_status` special-cases *only* `== 404`; if someone later loosens that to `>= 400` or `in client_error_range`, live claims sitting behind a transient 403 (auth blip) or 409 would be silently GC'd and **no test would fail**. Add e.g. `test_403_raises_and_is_not_gone`: `_resp(False, status_code=403)` must `pytest.raises(RuntimeError)` and must not return `_EXECUTION_GONE`.

muhammad-ali-e (commented):
Automated PR review (PR Review Toolkit: code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier). The status_code propagation chain (base_client → execution_client → ExecutionResponse → reaper) is correct and well-documented. Findings below are ranked; the org-scoped-404 issue already raised by @greptile-apps is not re-posted (comment 1 extends it with the infra/deploy-skew vector + a concrete fix). Highest-priority item is a second, un-updated caller of `_execution_status`.

workers/queue_backend/pg_queue/reaper.py
```diff
@@ -335,20 +341,26 @@ def rearm_expired_claims(conn: PgConnection) -> int:
...
+            return _EXECUTION_GONE
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in `0a740136b` — two layers. Reaper side: `_execution_status` now requires **both** `status_code==404` **and** the backend marker `"WorkflowExecution not found"` in the body before returning GONE, so an org-scoped 404 (or proxy/gateway/deploy 404) can't trigger GC. Backend side: `retrieve` confirms the row is genuinely absent via an **unscoped** `WorkflowExecution.objects.filter(id=…).exists()` before returning 404 — a scoped-out-but-existing execution now returns 500 (reaper retains the claim).

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Good catch — fixed in `0a740136b`. `_recover_one_barrier` now has a `status is _EXECUTION_GONE` branch **before** `is_completed()`, so the sentinel no longer hits `ExecutionStatus(<object>)` → `ValueError`. GONE → log + fall through to the existing cleanup (GC the orphaned `pg_barrier_state` + `pg_batch_dedup`, no status overwrite), symmetric with the claim path.

workers/queue_backend/pg_queue/reaper.py
muhammad-ali-e: Adopted your dual-signal in `0a740136b`: `status_code==404 AND "WorkflowExecution not found" in (response.error or "")`. A bare 404 (proxy/gateway/version skew) now raises → retained, so a bad deploy can't GC every claim in one sweep. Added tests: 404-with-marker→GONE, 404-without-marker→raise.

backend/workflow_manager/internal_views.py
muhammad-ali-e: Fixed in `0a740136b` — took the unscoped-existence approach: on `Http404`/`DoesNotExist`, `retrieve` checks `WorkflowExecution.objects.filter(id=…).exists()` (the manager only scopes on *user* context, so this is unscoped in the service-to-service internal path). Genuinely absent → 404; exists-but-scoped-out **or** a nested downstream `Http404` → 500 (retain). This closes both the org-scope and the over-broad-`try` holes at once. Backend test added for the scoped-out→500 case.

workers/shared/clients/base_client.py
muhammad-ali-e: Narrowed the docstring in `0a740136b` to your wording: status_code is populated only for a non-timeout 4xx / retry-exhausted 5xx; None for transport failures and the 429-exhausted / serialization / generic-fallback paths — 'don't assume any HTTP error carries a code.'

workers/shared/clients/execution_client.py
muhammad-ali-e: Added `TestStatusCodePropagationChain` in `0a740136b`: constructs a real `ExecutionAPIClient`, patches the HTTP layer to raise a real `APIRequestError(status_code=404/500)`, calls `get_workflow_execution`, and asserts the returned `ExecutionResponse.status_code` + `.error` marker. Plus `TestApiRequestErrorStatusCode` (carries 404 / defaults None). A rename in the chain now fails a test.

workers/tests/test_reaper_orphan_claim_404.py
muhammad-ali-e: Added `test_403_raises_and_is_not_gone` (and `test_404_without_marker_raises`) in `0a740136b` — pins 'only the app-level 404 is GONE'; loosening to `>=400` would now fail here instead of silently GC-ing live claims behind a 403/409.

muhammad-ali-e:
### Review addressed (`0a740136b`)

Thanks — several real P1s. All fixed:

**[P1] Org-scoped / overloaded 404 → GC of a live claim** (greptile + review). Two layers now guard it:
- Reaper `_execution_status` requires **both** `status_code==404` **and** the backend marker `"WorkflowExecution not found"` in the body → a proxy / gateway / deploy-skew 404 (or org-scoped 404) can't map to GONE.
- Backend `retrieve` confirms the row is **genuinely absent (unscoped `.exists()`)** before 404; exists-but-scoped-out → 500 (reaper retains). This also closes the over-broad-`try` (a nested downstream `Http404` for a different resource → 500, not a spurious "not found").

**[P1] `_recover_one_barrier` didn't handle `_EXECUTION_GONE`** → `ExecutionStatus(<object>)` `ValueError` churning every sweep. Added the sentinel branch (GC the orphaned barrier/dedup rows, no status overwrite), symmetric with the claim path.

**[P2] `APIRequestError` docstring** narrowed (429-exhausted / serialization / fallback don't thread `status_code`).

**[P2] Propagation-chain coverage** — new `TestStatusCodePropagationChain` exercises the real `APIRequestError(status_code) → get_workflow_execution → ExecutionResponse.status_code/.error` path (the mock-based tests couldn't catch a rename); plus `TestApiRequestErrorStatusCode`.

**[P2] Boundary test** — `test_403_raises_and_is_not_gone` + `test_404_without_marker_raises` pin "only the app-level 404 is GONE".

**SonarCloud S5778 ×2** — one throwing call per `pytest.raises` (the api is built outside the `with`).

Worker 30 + backend 18 green.

PR #2178, opened 2026-07-14
UN-3721 [FIX] PG finalization strand — terminal guard at the model layer (backend stale-object clobber)

muhammad-ali-e:
## What
Move the terminal-one-way guard from the HTTP endpoint (#2172) down to the **model layer**, so backend-internal writers can't revert a finalized PG execution.

## Why
An 800-user load test still stranded ~4% of PG executions (`EXECUTING` + NULL counters after the callback finalized `COMPLETED`). A log trace of 4 stranded IDs proved it is **not** a worker/guard-path race:
- callback wrote `COMPLETED` (HTTP 200), then ~12ms later the row reverted to `EXECUTING`+NULL, with **no status-transition log in any container**.
- That's a **backend-internal stale-object save**: a `WorkflowExecution` created as `EXECUTING` (NULL counters) is re-`save()`d after the callback committed `COMPLETED`; Django's full-row `.save()` reverts status + counters from the stale in-memory copy.

The #2172 guard lives only in the `update_status` HTTP endpoint (worker→backend), so it cannot see backend-internal `update_execution` / `.save()` writes — and those don't log the transition (why the trace looked "clean").

## How
- **`WorkflowExecution.update_execution`** (`models/execution.py`) and **`WorkflowExecutionServiceHelper.update_execution`** (`execution.py`): terminal-one-way guard, PG-scoped via `queue_message_id`. Re-read the committed status; if it is `COMPLETED`/`STOPPED` and the write would change it, **skip the entire save** — which also stops the stale NULL counters from clobbering the real ones. `ERROR` stays correctable to `COMPLETED`; Celery rows (NULL `queue_message_id`) are untouched.
- **`WorkflowHelper._set_result_acknowledge`** (`workflow_helper.py`): `save(update_fields=["result_acknowledged"])` — acknowledging a result must never rewrite status/counters (the confirmed full-save-after-completion site).

## Tests
6 new backend tests: refuses stale revert of `COMPLETED`/`STOPPED` (status **and** counters preserved), allows `ERROR→COMPLETED`, allows idempotent rewrite, Celery row unaffected, acknowledge doesn't touch status/counters. Full backend finalization suite: 22 green.

## Safety
PG-scoped (`queue_message_id`), so the Celery path is unchanged. The guard skips only a *reverting* write of a protected-terminal status — legitimate finalization (`EXECUTING→COMPLETED`) and the premature-`ERROR→COMPLETED` correction still work.

UN-3721 — follow-up to #2172 (endpoint guard was necessary but at the wrong layer for this clobber).

### DIFF
--- a/backend/workflow_manager/execution/tests/test_pg_finalization_fixes.py
+++ b/backend/workflow_manager/execution/tests/test_pg_finalization_fixes.py
@@ -231,3 +231,106 @@ class TerminalOneWayGuardTests(TestCase):
         ex.refresh_from_db()
         assert out.get("status") == "updated"
         assert ex.status == ExecutionStatus.COMPLETED.value
+
+
+class ModelStaleWriterGuardTests(TestCase):
+    """The terminal-one-way guard at the MODEL layer (update_execution) — where the
+    HTTP-endpoint guard can't reach. A stale backend object (created EXECUTING, NULL
+    counters, re-saved after the callback set COMPLETED) must not revert a
+    protected-terminal (COMPLETED/STOPPED) PG execution back to EXECUTING+NULL."""
+
+    def setUp(self):
+        self.wf = Workflow.objects.create(workflow_name="wf-stale")
+
+    def _stale(self, ex, status, **fields):
+        """A separate in-memory instance of ex with stale field values."""
+        stale = WorkflowExecution.objects.get(pk=ex.pk)
+        stale.status = status.value
+        for k, v in fields.items():
+            setattr(stale, k, v)
+        return stale
+
+    def test_pg_refuses_stale_revert_of_completed(self):
+        ex = WorkflowExecution.objects.create(
+            workflow=self.wf,
+            status=ExecutionStatus.COMPLETED,
+            queue_message_id=11,
+            total_files=1,
+            successful_files=1,
+            failed_files=0,
+        )
+        stale = self._stale(
+            ex, ExecutionStatus.EXECUTING, successful_files=None, failed_files=None
+        )
+        stale.update_execution(status=ExecutionStatus.EXECUTING)  # the clobber
+        ex.refresh_from_db()
+        assert ex.status == ExecutionStatus.COMPLETED.value  # not reverted
+        assert ex.successful_files == 1  # counters not nulled
+
+    def test_pg_refuses_stale_revert_of_stopped(self):
+        ex = WorkflowExecution.objects.create(
+            workflow=self.wf, status=ExecutionStatus.STOPPED, queue_message_id=12
+        )
+        self._stale(ex, ExecutionStatus.EXECUTING).update_execution(
+            status=ExecutionStatus.EXECUTING
+        )
+        ex.refresh_from_db()
+        assert ex.status == ExecutionStatus.STOPPED.value
+
+    def test_pg_allows_error_corrected_to_completed(self):
+        # ERROR is not protected — a premature ERROR must stay correctable.
+        ex = WorkflowExecution.objects.create(
+            workflow=self.wf, status=ExecutionStatus.ERROR, queue_message_id=13
+        )
+        WorkflowExecution.objects.get(pk=ex.pk).update_execution(
+            status=ExecutionStatus.COMPLETED
+        )
+        ex.refresh_from_db()
+        assert ex.status == ExecutionStatus.COMPLETED.value
+
+    def test_pg_allows_idempotent_completed_rewrite(self):
+        ex = WorkflowExecution.objects.create(
+            workflow=self.wf, status=ExecutionStatus.COMPLETED, queue_message_id=14
+        )
+        WorkflowExecution.objects.get(pk=ex.pk).update_execution(
+            status=ExecutionStatus.COMPLETED
+        )
+        ex.refresh_from_db()
+        assert ex.status == ExecutionStatus.COMPLETED.value
+
+    def test_celery_row_unaffected(self):
+        # queue_message_id NULL → legacy behavior, the stale write goes through.
+        ex = WorkflowExecution.objects.create(
+            workflow=self.wf,
+            status=ExecutionStatus.COMPLETED,
+            queue_message_id=None,
+            task_id=uuid.uuid4(),
+        )
+        self._stale(ex, ExecutionStatus.EXECUTING).update_execution(
+            status=ExecutionStatus.EXECUTING
+        )
+        ex.refresh_from_db()
+        assert ex.status == ExecutionStatus.EXECUTING.value  # legacy: overwritten
+
+    def test_result_acknowledge_does_not_touch_status_or_counters(self):
+        # A stale object acknowledging a result must not rewrite status/counters
+        # (update_fields scoping).
+        from workflow_manager.workflow_v2.workflow_helper import WorkflowHelper
+
+        ex = WorkflowExecution.objects.create(
+            workflow=self.wf,
+            status=ExecutionStatus.COMPLETED,
+            queue_message_id=15,
+            total_files=1,
+            successful_files=1,
+            failed_files=0,
+            result_acknowledged=False,
+        )
+        stale = self._stale(
+            ex, ExecutionStatus.EXECUTING, successful_files=None, failed_files=None
+        )
+        WorkflowHelper._set_result_acknowledge(stale)
+        ex.refresh_from_db()
+        assert ex.status == ExecutionStatus.COMPLETED.value
+        assert ex.successful_files == 1
+        assert ex.result_acknowledged is True
--- a/backend/workflow_manager/execution/tests/test_pg_finalization_fixes.py
+++ b/backend/workflow_manager/execution/tests/test_pg_finalization_fixes.py
@@ -269,13 +269,98 @@ class ModelStaleWriterGuardTests(TestCase):
 
     def test_pg_refuses_stale_revert_of_stopped(self):
         ex = WorkflowExecution.objects.create(
-            workflow=self.wf, status=ExecutionStatus.STOPPED, queue_message_id=12
+            workflow=self.wf,
+            status=ExecutionStatus.STOPPED,
+            queue_message_id=12,
+            total_files=2,
+            successful_files=1,
+            failed_files=1,
         )
-        self._stale(ex, ExecutionStatus.EXECUTING).update_execution(
-            status=ExecutionStatus.EXECUTING
-        )
+        self._stale(
+            ex, ExecutionStatus.EXECUTING, successful_files=None, failed_files=None
+        ).update_execution(status=ExecutionStatus.EXECUTING)
         ex.refresh_from_db()
         assert ex.status == ExecutionStatus.STOPPED.value
+        assert ex.successful_files == 1 and ex.failed_files == 1  # counters preserved
+
+    def test_pg_same_status_does_not_clobber_counters(self):
+        # Same-status COMPLETED→COMPLETED: the guard is a no-op, so update_fields must
+        # be what stops the stale NULL counters from clobbering the real ones.
+        ex = WorkflowExecution.objects.create(
+            workflow=self.wf,
+            status=ExecutionStatus.COMPLETED,
+            queue_message_id=17,
+            total_files=1,
+            successful_files=1,
+            failed_files=0,
+        )
+        self._stale(
+            ex, ExecutionStatus.COMPLETED, successful_files=None, failed_files=None
+        ).update_execution(status=ExecutionStatus.COMPLETED)
+        ex.refresh_from_db()
+        assert ex.successful_files == 1  # not nulled
+
+    def test_pg_stale_null_marker_still_guarded_and_not_nulled(self):
+        # A stale object whose in-memory queue_message_id is None (snapshotted before
+        # dispatch recorded it) must STILL be guarded (routing reads the persisted
+        # marker) AND the marker must not be nulled by the write.
+        ex = WorkflowExecution.objects.create(
+            workflow=self.wf,
+            status=ExecutionStatus.COMPLETED,
+            queue_message_id=99,
+            successful_files=1,
+        )
+        self._stale(
+            ex, ExecutionStatus.EXECUTING, queue_message_id=None, successful_files=None
+        ).update_execution(status=ExecutionStatus.EXECUTING)
+        ex.refresh_from_db()
+        assert ex.status == ExecutionStatus.COMPLETED.value  # still guarded
+        assert ex.queue_message_id == 99  # marker NOT nulled
+        assert ex.successful_files == 1
+
+    def test_pg_refused_status_still_applies_error_and_attempt(self):
+        # The refused status must NOT silently drop error / increment_attempt.
+        ex = WorkflowExecution.objects.create(
+            workflow=self.wf,
+            status=ExecutionStatus.COMPLETED,
+            queue_message_id=16,
+            attempts=0,
+        )
+        self._stale(ex, ExecutionStatus.EXECUTING).update_execution(
+            status=ExecutionStatus.EXECUTING,
+            error="late failure",
+            increment_attempt=True,
+        )
+        ex.refresh_from_db()
+        assert ex.status == ExecutionStatus.COMPLETED.value  # status refused
+        assert ex.error_message == "late failure"  # error applied
+        assert ex.attempts == 1  # increment applied
+
+    def test_service_helper_update_execution_is_guarded(self):
+        from workflow_manager.workflow_v2.execution import (
+            WorkflowExecutionServiceHelper,
+        )
+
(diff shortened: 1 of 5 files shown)


backend/workflow_manager/workflow_v2/models/execution.py
```diff
@@ -373,6 +373,35 @@ def update_execution(
...
+            # created as EXECUTING (with NULL file counters) and re-saved by a backend
+            # path after the callback already committed COMPLETED — must not revert a
+            # protected-terminal status. Re-read the committed status and, if it is
+            # COMPLETED/STOPPED and this write would change it, skip the ENTIRE save
+            # (which also prevents the stale object's NULL counters from clobbering
+            # the real ones). This guards the backend-internal writers the HTTP
+            # update_status guard can't see. ERROR is deliberately NOT protected (a
+            # premature ERROR stays correctable to COMPLETED). Celery rows
+            # (queue_message_id NULL) are unaffected.
+            if self.queue_message_id is not None:
```
muhammad-ali-e: **[High] Guard gates on the stale `self.queue_message_id` — can fail open and NULL out the PG marker.**

The guard re-reads the committed *status* fresh (lines 387-391, correct), but decides *whether to run at all* from `self.queue_message_id` — a field on the same stale in-memory object the guard exists to defend against. If this object was snapshotted before `_record_dispatch_handle` recorded the handle (`queue_message_id` is written via a scoped `save(update_fields=["queue_message_id"])` at `execution.py:466`), then `self.queue_message_id is None`, the gate is skipped entirely, and the unconditional `self.save()` at line 419 not only reverts the protected status but writes `NULL` back into the `queue_message_id` column — permanently making the row look like a Celery row and disabling *this* guard plus the sibling guards (all gate on `queue_message_id is not None`) for every later write.

**Fix:** read the marker from the same query as the status and gate on the persisted value:
```python
committed_qmid, committed = (
    WorkflowExecution.objects.filter(pk=self.pk)
    .values_list("queue_message_id", "status")
    .first()
) or (None, None)
if committed_qmid is not None and committed in protected and status.value != committed:
    ...
    return
```
No current test exercises this (every `_stale()` helper reads the row *after* `queue_message_id` is set).

backend/workflow_manager/workflow_v2/models/execution.py
```diff
@@ -373,6 +373,35 @@ def update_execution(
...
+                }
+                if committed in protected and status.value != committed:
+                    logger.warning(
+                        "update_execution: refusing to overwrite protected status "
+                        "'%s' with '%s' for execution %s (stale-writer guard)",
+                        committed,
+                        status.value,
+                        self.id,
+                    )
+                    return
```
muhammad-ali-e: **[High] Early `return` silently drops `error` and `increment_attempt`, not just the status.**

The guard `return`s from *inside* the `if status is not None:` block, above the `if error:` (line ~430) and `if increment_attempt:` (line ~432) handling. So a call on a protected-terminal row like:
```python
update_execution(status=ExecutionStatus.ERROR, error="post-finalization cleanup failed", increment_attempt=True)
```
correctly refuses the status revert **but also** silently discards `error_message` (a real post-terminal failure diagnostic vanishes) and skips `attempts += 1` (retry accounting corrupted) — with only a WARNING that mentions status. This re-introduces a silent failure on a different field.

**Fix:** protect only the status field — on a protected row skip the status assignment / `execution_time` / rate-limit logic, but fall through to apply `error`/`increment_attempt` and `save(update_fields=[...])` only the fields you touched. At minimum, include the dropped `error`/`increment_attempt` payload in the WARNING so it isn't lost.

backend/workflow_manager/workflow_v2/execution.py
```diff
@@ -177,6 +177,27 @@ def update_execution(
...
+                and status.value != execution.status
+            ):
+                logger.warning(
+                    "update_execution: refusing to overwrite protected status '%s' "
+                    "with '%s' for execution %s (stale-writer guard)",
+                    execution.status,
+                    status.value,
+                    execution.id,
+                )
+                return
```
muhammad-ali-e: **[High] HTTP-layer guard: (a) early `return` drops `error`/`increment_attempt`, and (b) unlocked read-modify-write (TOCTOU).**

(a) Same issue as the model layer — this `return` sits inside `if status is not None:`, above the `if error:` / `if increment_attempt:` blocks (lines ~215-218), so a refused call also silently drops the error message and attempt increment.

(b) This method reads `execution` at line 177, evaluates the guard, then `execution.save()` at line 220 — with **no** `transaction.atomic()` / `select_for_update()`, unlike the `internal_views` guard it says it "mirrors" (which locks the row precisely to close this window). A callback can commit `COMPLETED` between the read and the save; the guard saw a pre-terminal status, passes, and the full-row `save()` clobbers the finalized row. Wrap the read-check-save in `with transaction.atomic():` and re-fetch via `select_for_update()` when `queue_message_id is not None`.

(This is the same *class* as the greptile model-layer race @388-392, but a distinct, unflagged location.)

backend/workflow_manager/workflow_v2/execution.py
```diff
@@ -177,6 +177,27 @@ def update_execution(
         execution = WorkflowExecution.objects.get(pk=self.execution_id)
 
         if status is not None:
+            # Terminal-one-way for PG executions (mirrors the model method): don't let
+            # a write revert a protected-terminal status. execution is freshly read
+            # above, so its status is the committed one. ERROR stays correctable to
+            # COMPLETED; Celery rows (queue_message_id NULL) are unaffected.
+            if (
+                execution.queue_message_id is not None
```
muhammad-ali-e: **[Med] Design/simplification: the protected-terminal rule is duplicated as inline literals and has already diverged.**

The protected set `{COMPLETED, STOPPED}` is written two different ways — a **tuple** here (`execution.py:187-190`) and a **set literal** in `models/execution.py:392-395` — with duplicated warning text and duplicated 4-/10-line rationale comments. `ExecutionStatus` already owns every *other* status subset as a single-sourced classmethod (`terminal_values()`, `failure_statuses()`), each with a "one definition so checks stay in sync" comment. This third subset is the only one left inline, so adding a future terminal status (e.g. `CANCELLED`) silently misses these two spots.

**Suggest:** add `ExecutionStatus.protected_terminal_values()` (derive as `terminal_values() - {ERROR.value}` so new terminals are protected-by-default) and a shared `refuses_overwrite(committed, target)` predicate; name the PG discriminator as a `WorkflowExecution.is_pg_execution` property (`queue_message_id is not None` is repeated at execution.py:185/458, models/execution.py:386, internal_views.py:530). Note the model layer's committed-status *re-read* is load-bearing (must stay) — only the set/predicate/naming should be shared, so a "make these consistent" refactor mustn't collapse the re-read away.

backend/workflow_manager/execution/tests/test_pg_finalization_fixes.py
```diff
@@ -231,3 +231,106 @@ def test_pg_allows_idempotent_completed_rewrite(self):
...
+        ex = WorkflowExecution.objects.create(
+            workflow=self.wf, status=ExecutionStatus.ERROR, queue_message_id=13
+        )
+        WorkflowExecution.objects.get(pk=ex.pk).update_execution(
+            status=ExecutionStatus.COMPLETED
+        )
+        ex.refresh_from_db()
+        assert ex.status == ExecutionStatus.COMPLETED.value
+
+    def test_pg_allows_idempotent_completed_rewrite(self):
```
muhammad-ali-e, line 376: **[Med] Test coverage gaps.**

1. **Idempotent same-status counter clobber is not exercised.** This test builds the writer with a *fresh* `WorkflowExecution.objects.get(pk=ex.pk)` (correct counters) and asserts only `status` — so it can't catch the greptile-reported counter clobber, where the guard is a no-op (`status.value == committed`) and the trailing full `save()` writes stale/NULL counters. Add a stale-instance variant that nulls the counters and asserts `ex.successful_files == 1` survives.
2. **The HTTP-layer guard (`WorkflowExecutionServiceHelper.update_execution`, execution.py:171-220) has no direct test** — only the model method and the `internal_views` view are covered. It carries its own hand-copied guard that can drift; mirror the model cases against the service helper.
3. **Silent side-effect drop** (error/increment_attempt on a protected row) is untested at both layers.
4. **STOPPED counter preservation** isn't asserted (`test_pg_refuses_stale_revert_of_stopped` creates no counters and checks only status) — make it symmetric with the COMPLETED case.

muhammad-ali-e (commented):
**PR Review Toolkit — consolidated findings (Code Reviewer · Silent-Failure Hunter · Type-Design · Test Analyzer · Comment Analyzer · Simplifier).**

The terminal-one-way guard is a sound fix for the reported clobber. Below are findings that are *new* relative to the existing greptile P1 comments (same-status counter clobber @397, check-save race @388-392, stale-cache publish @helper:443, `update_execution_err` bypass @execution:200) — I've deliberately not repeated those. Highest-value items first. Inline comments follow.

backend/workflow_manager/workflow_v2/models/execution.py
muhammad-ali-e: Fixed (`20e1c8ec0`). The write now ROUTES on the **persisted** `queue_message_id` (fresh `values_list` read), never stale `self`. A PG row snapshotted before dispatch → routed to the guarded path; and PG writes are field-scoped (`update_fields`) so `queue_message_id` is never written → can't be nulled. New test `test_pg_stale_null_marker_still_guarded_and_not_nulled`.

backend/workflow_manager/workflow_v2/models/execution.py
```diff
@@ -373,6 +373,35 @@ def update_execution(
...
+            # update_status guard can't see. ERROR is deliberately NOT protected (a
+            # premature ERROR stays correctable to COMPLETED). Celery rows
+            # (queue_message_id NULL) are unaffected.
+            if self.queue_message_id is not None:
+                committed = (
+                    WorkflowExecution.objects.filter(pk=self.pk)
+                    .values_list("status", flat=True)
+                    .first()
+                )
+                protected = {
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed. The PG path now does the committed-status read and the write inside `transaction.atomic()` with `select_for_update()` — the read-check-save is atomic, so a concurrent callback can't commit COMPLETED in the window.

backend/workflow_manager/workflow_v2/models/execution.py
```diff
@@ -373,6 +373,35 @@ def update_execution(
...
+                    ExecutionStatus.COMPLETED.value,
+                    ExecutionStatus.STOPPED.value,
+                }
+                if committed in protected and status.value != committed:
+                    logger.warning(
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed. Even a same-status write no longer full-saves: the PG path writes `update_fields` (status/execution_time/error/attempts + modified_at) only, so stale NULL counters can't clobber. New test `test_pg_same_status_does_not_clobber_counters`.

backend/workflow_manager/workflow_v2/models/execution.py
muhammad-ali-e: Fixed. The guard now refuses ONLY the status field — it sets `self.status = committed` and falls through, so `error` / `increment_attempt` still apply (attempts incremented from the locked value). Warning names both. New test `test_pg_refused_status_still_applies_error_and_attempt`.

backend/workflow_manager/workflow_v2/execution.py
muhammad-ali-e: Both fixed. The service-helper method now just **delegates** to the model `update_execution` (single guard, atomic + field-scoped), eliminating the hand-copied guard, its early-return drop, and its unlocked TOCTOU. New test `test_service_helper_update_execution_is_guarded`.

backend/workflow_manager/workflow_v2/workflow_helper.py
```diff
@@ -436,7 +436,11 @@ def _set_result_acknowledge(execution: WorkflowExecution) -> None:
         """
         if not execution.result_acknowledged:
             execution.result_acknowledged = True
-            execution.save()
+            # Scope the write to ONLY result_acknowledged. A full save() here would
+            # rewrite the whole row from this (possibly stale) object — the exact
+            # mechanism that reverted COMPLETED executions back to EXECUTING+NULL under
+            # load. Acknowledging a result must never touch status / counters.
+            execution.save(update_fields=["result_acknowledged"])
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed. `_set_result_acknowledge` now does a DB-side `WorkflowExecution.objects.filter(pk=...).update(result_acknowledged=True)` — no `save()`, so `_handle_execution_cache()` never runs and can't republish a stale status to Redis. Test asserts status/counters untouched.

backend/workflow_manager/workflow_v2/execution.py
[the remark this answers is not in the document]
muhammad-ali-e: Fixed. `update_execution_err` now routes through `execution.update_execution(status=ERROR, error=...)` instead of a direct save, so it can't revert a PG execution the callback finalized COMPLETED/STOPPED. New test `test_update_execution_err_is_guarded`.

backend/workflow_manager/workflow_v2/execution.py
muhammad-ali-e: Done. Added `ExecutionStatus.protected_terminal_values()` (= `terminal_values() - {ERROR}`, so a new terminal is protected-by-default) in shared core; both guard sites use it. The service-helper's duplicated guard is gone (now delegates).

backend/workflow_manager/execution/tests/test_pg_finalization_fixes.py
muhammad-ali-e: All added (`20e1c8ec0`): same-status counter clobber, STOPPED counter preservation, dropped error/increment on a protected row, the service-helper guard, and `update_execution_err`. 11 stale-writer tests; full suite 29 green.

muhammad-ali-e:
### Review addressed (`20e1c8ec0`) — the guard is now correct, not just present

Thanks — this round found real holes in the first pass. All fixed:

| Finding | Fix |
|---|---|
| **Gated on stale `self.queue_message_id`** → could NULL the marker & disable all guards | Route on the **persisted** marker (fresh read); PG writes are `update_fields`-scoped so `queue_message_id` is never written |
| **Check-save race (not atomic)** | PG path reads committed status + writes inside `transaction.atomic()` + `select_for_update()` |
| **Same-status full save clobbers counters** | `update_fields` only — never touches counters |
| **Early `return` drops `error`/`increment_attempt`** | Refuse ONLY the status field; error/increment still apply (attempts from the locked value) |
| **`_set_result_acknowledge` republishes stale cache** | DB-side `.update(result_acknowledged=True)` — no `save()`, no cache write |
| **`update_execution_err` bypasses guard** | Routes through `update_execution` |
| **Service-helper hand-copied guard (TOCTOU + dropped fields)** | Now **delegates** to the one model method |
| **Duplicated protected set** | `ExecutionStatus.protected_terminal_values()` (= `terminal_values() − {ERROR}`) in shared core |

Also merged the latest `feat` (resolved the test-file conflict — kept #2177's `RetrieveNotFoundTests` alongside the new `ModelStaleWriterGuardTests`).

11 new stale-writer tests; full backend finalization suite **29 green**. SonarCloud was already clean.

muhammad-ali-e:
SonarCloud S3776 resolved in `2cb8f6e64` — extracted `_apply_guarded_status` from `_update_execution_pg_guarded` (complexity 16→<15) and removed a duplicated rate-limit-release block. 29 backend finalization tests green.

backend/workflow_manager/workflow_v2/models/execution.py
```diff
@@ -369,8 +369,31 @@ def update_execution(
...
+        # handle would otherwise be mis-routed to the legacy path, whose full save()
+        # would revert the row AND write NULL back into queue_message_id — permanently
+        # disabling this and every sibling guard (all gate on queue_message_id).
+        committed_qmid = (
+            WorkflowExecution.objects.filter(pk=self.pk)
+            .values_list("queue_message_id", flat=True)
+            .first()
+        )
+        if committed_qmid is None:
+            self._update_execution_legacy(status, error, increment_attempt)
```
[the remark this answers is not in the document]
muhammad-ali-e: Fixed in `35da92fc6`. `update_execution` now takes the row lock **first**: it reads the persisted `queue_message_id` + status under `select_for_update()` and dispatches to the legacy or guarded write **inside the same transaction**. So routing and the write are atomic — a concurrent `_record_dispatch_handle` can't interleave between them; it serializes after our commit (and if it had already committed, we'd have read the marker as set and taken the guarded path). Split into `_apply_legacy_update` / `_apply_pg_guarded_update`, both under the lock. 29 tests green.

PR #2180, opened 2026-07-15
UN-3740 [FIX] pg_queue — squash 0001-0014 into a single initial migration

muhammad-ali-e:
## What
Collapses the 14 incremental `pg_queue` migrations (`0001`-`0014`) into a single `0001_initial_squashed`, before the app reaches `main`/staging. Only integration has ever applied these, so this is the clean window to do it. Removes the create-then-drop index churn (`pg_queue_message_dequeue_idx`, `pg_periodic_schedule_due_idx`).

## How
- `squashmigrations pg_queue 0014`, keeping `replaces=[0001…0014]`, then **delete the 14 originals**. Safe because integration is fully at `0014` and every other env is fresh — no mid-way installs, so `replaces` reconciles integration with **zero SQL**.
- Ported the `0005` `RunPython` seed (seeds the single `pg_orchestrator_lock` row — leader-election needs it) into the squash.
- Removed the leftover `CreateModel`-declared `dequeue_idx` + its `RemoveIndex`: Django defers the `CreateModel` index to `deferred_sql`, so on a fresh apply it was re-created **after** the `RemoveIndex` drop and survived. Dropping the pair yields the correct final schema with no churn.
- Removed `test_migration_0014_backfill.py` (it imported the squashed-away 0014 module). The backfill's behavioural guard remains in `test_pg_queue_client.py::test_inflight_backfill_sql_classifies_by_vt`.

## Effect on the integration DB
**Zero schema change.** Integration has `0001`-`0014` recorded; the squash's `replaces` makes Django mark it applied without running any SQL.

## Validation — run against a real local Postgres (existing DB + new DB)
- `makemigrations --check` → No changes (squash faithfully represents the models).
- **New DB** (main/staging/prod path): fresh apply → schema **index-for-index identical** to applying 0001-14; `pg_orchestrator_lock` seed row present.
- **Existing DB @0014** (integration path): squash-only migrate → `showmigrations` = `[X] 0001_initial_squashed (14 squashed migrations)`, no SQL run.

## Follow-up
Once feat→main lands and every env has the squash recorded, a later cleanup can drop the `replaces` attribute (plain initial migration).

Ticket: UN-3740


muhammad-ali-e (commented):
## PR Review — UN-3740 `pg_queue` migration squash

**Verdict: the squash is faithful and safe to merge.** A full line-by-line reconciliation of `0001_initial_squashed` against all 14 originals (+ `models.py`) confirms every model, field, index, and constraint matches the final schema; column ordering is preserved on a fresh apply; `replaces` is complete and exact (all 14, names character-for-character); and both apply paths are correct — fresh DB runs the single squash against empty tables, integration already-at-0014 reconciles with **zero SQL**. The `RunPython` seed and the `AddField(state) → SET DEFAULT → backfill → indexes → constraint` ordering are carried over correctly. The net-removed `pg_queue_message_dequeue_idx` and the folded `pg_periodic_schedule_due_idx` churn are both handled correctly.

All inline comments below are **non-blocking polish** — no correctness defect was found.

| # | Severity | Finding | Location |
|---|---|---|---|
| 1 | Low | "bare UPDATE" undersells the real lease predicate | `0001_initial_squashed.py:8` |
| 2 | Medium | `SET DEFAULT` rationale dropped | `:216` |
| 3 | Med–High | backfill rationale dropped + structural test deleted w/o replacement | `:220` |
| 4 | Low | `reverse_sql=""` → prefer `RunSQL.noop` | `:221` |
| 5 | Low–Med | fillfactor/autovacuum `NOTE (UN-3445)` dropped | `:247` |

**Two findings outside this PR's diff (can't inline-anchor):**
- **Comment rot (Low, please fix):** `workers/tests/test_pg_queue_client.py:696` still reads `# structural guard is backend/pg_queue/tests/test_migration_0014_backfill` — a dangling reference to the file this PR deletes. Update it so the next reader doesn't chase a ghost file.
- **Type design (out of scope, optional):** `PgTaskResult.status` is a documented closed enum (`"completed" | "failed"`) but, unlike `PgQueueMessage.state`, carries no `CheckConstraint` — the raw-SQL writer can persist a typo'd status silently. Mirror `state_valid` in `models.py` via a **new** migration (never by editing this squash) if you want to close it.

<sub>Generated by the PR Review Toolkit — code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier.</sub>

backend/pg_queue/migrations/0001_initial_squashed.py
```diff
@@ -0,0 +1,248 @@
+# Generated by Django 4.2.30 on 2026-07-15 12:31
+
+import django.utils.timezone
+from django.db import migrations, models
+
+
+# Seed function ported from the squashed 0005_pgorchestratorlock: leader-election
+# is a bare `UPDATE ... WHERE id = 1`, so the single lock row must exist or every
```
muhammad-ali-e: **[Low · Comment accuracy]** This calls leader-election "a bare `UPDATE ... WHERE id = 1`", but the real statement in `workers/queue_backend/pg_queue/leader_election.py:205-213` also carries a lease predicate:

```sql
WHERE id = 1 AND (leader = '' OR acquired_at < now() - make_interval(...))
```

The load-bearing claim — a plain `UPDATE` with **no** INSERT/upsert fallback, so the `id=1` row must pre-exist or `try_acquire` returns `False` forever — is fully correct. Only the word "bare" undersells the WHERE clause and may make a future reader distrust the comment after grepping the real SQL.

Suggest: *"leader-election is a plain `UPDATE` keyed on `id = 1` (no upsert/INSERT fallback), so the single lock row must exist or every `try_acquire` silently returns `False` forever."*

backend/pg_queue/migrations/0001_initial_squashed.py
```diff
@@ -0,0 +1,248 @@
...
+                "db_table": "pg_orchestration_claim",
+            },
+        ),
+        migrations.AddField(
+            model_name="pgqueuemessage",
+            name="state",
+            field=models.TextField(default="ready"),
+        ),
+        migrations.RunSQL(
+            sql="ALTER TABLE pg_queue_message ALTER COLUMN state SET DEFAULT 'ready'",
```
muhammad-ali-e, line 224: **[Medium · Lost rationale]** The original `0014` documented *why* this DB-level default exists: Django's field `default=` is app-level only and never emits a DB default, so this `SET DEFAULT` is the **sole** thing making the schema self-protecting against a raw `INSERT` that omits `state`. The squash kept the operation but dropped the explanation. This is a "why" comment worth restoring — the text is preserved in git at `origin/feat/UN-3445-pg-queue-integration:backend/pg_queue/migrations/0014_...`.

backend/pg_queue/migrations/0001_initial_squashed.py
```diff
@@ -0,0 +1,248 @@
...
+            reverse_sql="ALTER TABLE pg_queue_message ALTER COLUMN state DROP DEFAULT",
+        ),
+        migrations.RunSQL(
+            sql="UPDATE pg_queue_message SET state = 'claimed' WHERE vt > now()",
```
muhammad-ali-e, line 236: **[Medium–High · Lost safety rationale + deleted structural guard]** Two related regressions converge on this backfill:

1. **Rationale dropped.** Original `0014` explained this at length: `AddField(state)` backfills *every* row — including in-flight, claimed-but-unacked rows carrying a future `vt` — to `'ready'`; without this `UPDATE`, a rolling deploy over a live queue makes still-processing messages instantly re-claimable → one-time double processing. Restore that comment so this line reads as safety-critical and order-sensitive.

2. **Structural test deleted without replacement.** The removed `test_migration_0014_backfill.py` imported the migration module and asserted the backfill (a) exists, (b) has the exact `state='claimed'` / `vt > now()` predicate, and (c) runs **after** `AddField(state)`. The surviving `workers/tests/test_pg_queue_client.py::test_inflight_backfill_sql_classifies_by_vt` runs a *hardcoded copy* of the SQL — it proves predicate semantics but never reads this file, so a future edit that drops/reorders/inverts this op passes every test. Deleting (rather than **re-pointing**) the guard is the gap.

Suggested fix — repoint the guard at the squash:

```python
from importlib import import_module
from django.db import migrations
_M = import_module("pg_queue.migrations.0001_initial_squashed")

def test_backfill_present_correct_and_ordered_after_addfield():
    ops = _M.Migration.operations
    add = next(i for i,o in enumerate(ops)
               if isinstance(o, migrations.AddField) and o.name == "state")
    bf  = next(i for i,o in enumerate(ops)
               if isinstance(o, migrations.RunSQL)
               and "state = 'claimed'" in str(o.sql) and "vt > now()" in str(o.sql))
    assert bf > add
```

_Note:_ in practice the loss is low-severity — this is now a squashed *initial* migration (empty table on fresh apply, never re-run on existing DBs via `replaces`), so the backfill's real protective window is already past. But the guard is cheap to restore.

backend/pg_queue/migrations/0001_initial_squashed.py
```diff
@@ -0,0 +1,248 @@
...
+            reverse_sql="",
```
muhammad-ali-e: **[Low · Cosmetic]** `reverse_sql=""` is functionally identical to the original's `migrations.RunSQL.noop` (Django defines `RunSQL.noop == ""`), so this is **not** a behavioural change. But sitting next to the sibling `RunSQL` above that has a real reverse (`DROP DEFAULT`), the bare `""` reads like a forgotten reverse. Prefer `reverse_sql=migrations.RunSQL.noop` to make the intentional non-reversal self-documenting.

backend/pg_queue/migrations/0001_initial_squashed.py
```diff
@@ -0,0 +1,248 @@
...
+                name="pg_queue_message_claimed_idx",
+            ),
+        ),
+        migrations.AddConstraint(
+            model_name="pgqueuemessage",
+            constraint=models.CheckConstraint(
+                check=models.Q(("state__in", ["ready", "claimed"])),
+                name="pg_queue_message_state_valid",
+            ),
+        ),
```
muhammad-ali-e, line 263: **[Low–Medium · Lost rationale]** Original `0014` ended with a `NOTE (UN-3445)` explaining that fillfactor + aggressive per-table autovacuum tuning was **deliberately kept out** of this migration: it's env-specific ops tuning, and a failed `ALTER TABLE SET (autovacuum_*)` under a locked-down managed role (Cloud SQL / RDS / Azure) would abort the whole migrate. The squash dropped that note. Worth restoring so nobody "helpfully" re-adds the tuning here and breaks a managed-Postgres migration.

muhammad-ali-e:
Thanks for the thorough line-by-line review (+ SonarCloud green, Greptile clean). All inline comments addressed in `26e9d1b26` — **comment/cosmetic only, zero schema/behaviour change** (`makemigrations --check` = *No changes detected*), so the migration and both apply paths are untouched.

| Comment | Resolution |
|---|---|
| L8 — "bare `UPDATE`" undersells the WHERE | Reworded → *"plain `UPDATE` keyed on `id = 1` (with a lease predicate) and has NO upsert/INSERT fallback…"* |
| L216 — lost *why* for `SET DEFAULT` | **Restored** verbatim from `0014` (ORM `default=` is app-level only; this DB default is the sole guard against a raw `INSERT` omitting `state`) |
| L220 — lost backfill safety rationale | **Restored** verbatim (backfill lands genuinely in-flight rows as `claimed` → prevents one-time double-processing on a rolling deploy; order-sensitive after `AddField`) |
| L221 — `reverse_sql=""` reads like a forgotten reverse | Changed to `migrations.RunSQL.noop` (identical value, self-documenting) |
| L247 — lost the fillfactor/autovacuum `NOTE` | **Restored** verbatim (tuning deliberately kept out; would abort migrate on a locked-down managed role) |

**On the deleted `test_migration_0014_backfill.py` (L220 part 2):** intentional — the structural migration test was dropped by decision (the migration is a one-time squash, not iterated), and the **behavioural** guard remains in `test_pg_queue_client.py::test_inflight_backfill_sql_classifies_by_vt`. Keeping it removed.

**Deployment validation (real DBs, both paths):**
- **Fresh / staging path** — on a DB with all of main's migrations applied (staging shape, no pg_queue), `migrate` applied `pg_queue.0001_initial_squashed` + `workflow_v2 0020_queue_message_id → 0022_merge → 0023` cleanly on top.
- **Existing DB @0014 (integration path)** — squash marked applied via `replaces` with **zero SQL** (`[-] → [X]`).

PR #2182, opened 2026-07-16
UN-3741 [FIX] Show created/modified metadata on adapter and Prompt Studio list pages

athul-rs:
## What

- Adapter list and Prompt Studio list rows now show an inline `Updated <relative time>` hint next to the owner, plus an info icon whose hover tooltip shows absolute **Created** / **Modified** times (browser timezone), the **Model** (adapters) and **Prompt count** (Prompt Studio).
- Backend: `AdapterListSerializer` now returns `created_at` / `modified_at`; the Prompt Studio list returns `modified_at` as `max(tool.modified_at, latest prompt.modified_at)`.
- Owner email in list rows now ellipsizes with a full-address tooltip instead of wrapping mid-address when the row is tight.

## Why

- Customer request: richer metadata on the adapter and Prompt Studio list pages (created / modified times) to tell stale items from recently-touched ones.
- Prompt edits never touch the `CustomTool` row, so exposing the raw `modified_at` would show misleadingly stale times on actively-edited projects — hence the latest-prompt subquery.

## How

- All display logic lives in the shared `ListView` widget (`frontend/src/components/widgets/list-view/`), which the adapter, Prompt Studio, Workflows and Connectors pages all render through. Rendering is presence-keyed on the item fields, so pages whose APIs lack a field simply don't show it.
- No new endpoints and no extra frontend requests — everything rides the existing list responses. The Prompt Studio annotation is a correlated subquery in the same single list query, mirroring the existing `_prompt_count` subquery.

- No. API changes are additive (two extra fields on the adapter list; Prompt Studio already returned `modified_at`, only its value becomes more accurate). The new UI block renders only when `modified_at` is present, so any consumer without the field renders exactly as before. Workflows and Connectors pages (same widget, serializers already return timestamps) gain the same display — verified to follow the same row anatomy.
- Known nuance: the Prompt Studio tooltip's prompt count reuses the existing `prompt_count` field, which counts all prompt rows including notes (pre-existing semantics, also consumed elsewhere).

## Database Migrations

- None (uses existing `BaseModel.created_at` / `modified_at` columns).

## Env Config

- None.

## Relevant Docs

- N/A

## Related Issues or PRs

- [UN-3741](https://zipstack.atlassian.net/browse/UN-3741)

## Dependencies Versions

- No changes (frontend reuses the existing `moment` dependency).

## Notes on Testing

- New test `test_list_modified_at_reflects_prompt_edits` (verified red on unpatched code, green with the fix); prompt studio suite 2/2 and adapter suite 22/22 against live Postgres.
- `ruff` (pinned 0.3.4) clean, `biome check` clean, `vite build` clean.
- Visual verification via the real `ListView` component rendered with representative data at standard width: rows stay single-line; long owner emails ellipsize with a full-address tooltip.

## Screenshots

_Screenshots to

image

 be attached below (list view + hover tooltips for adapters and Prompt Studio)._

## Checklist

[UN-3741]: https://zipstack.atlassian.net/browse/UN-3741?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/prompt_studio/prompt_studio_core_v2/serializers.py
```diff
@@ -46,6 +46,7 @@ class CustomToolListSerializer(serializers.ModelSerializer):
 
     created_by_email = serializers.SerializerMethodField()
     prompt_count = serializers.SerializerMethodField()
+    modified_at = serializers.SerializerMethodField()
```
ritwik-g: **P2 — `?ordering=modified_at` on this list endpoint is now silently ignored.**

Converting `modified_at` to a `SerializerMethodField` removes it from `OrderingFilter`'s valid fields: the backend is enabled globally (`backend/settings/base.py:628`), this view declares no `ordering_fields`, so DRF derives orderable fields from the serializer and excludes `source='*'` fields (which every `SerializerMethodField` is). `remove_invalid_fields` then drops the param silently — `GET /prompt-studio/?ordering=-modified_at` returns 200 in DB-arbitrary order instead of sorting, with no error. The frontend doesn't send this param today (grepped), so it's a latent public-API regression rather than a live UI bug.

[verified: settings + view checked in the PR head; DRF `get_default_valid_fields` excludes `source='*'`]

Fix options: annotate the effective timestamp on the queryset (e.g. `Greatest`/`Coalesce` over the subquery) and order off the annotation, or declare explicit `ordering_fields` on the view.

backend/prompt_studio/prompt_studio_core_v2/serializers.py
```diff
@@ -70,6 +71,15 @@ class Meta:
     def get_created_by_email(self, instance):
         return instance.created_by.email if instance.created_by else ""
 
+    def get_modified_at(self, instance):
+        """Latest of the tool's own modified_at and its prompts' — prompt
+        edits never touch the CustomTool row (UN-3741).
+        """
+        last_prompt_modified = getattr(instance, "_last_prompt_modified", None)
```
ritwik-g: **P2 — missing-annotation fallback silently reproduces the exact UN-3741 stale value; the sibling method handles the same case loudly.**

`get_prompt_count` (a few lines below) guards the identical missing-annotation case with `hasattr` + `logger.warning` + a correct per-instance fallback query. `get_modified_at` instead does a truthy `getattr` and silently returns stale `instance.modified_at` — the very bug this PR fixes — with no log line and no way to tell it from correct output. Today no path lacks the annotation (the serializer is only used for `list`, which always annotates — verified), so this isn't live; but any future caller (new action, plugin, direct serializer use) regresses UN-3741 invisibly. Truthiness also can't distinguish "no annotation" from "no prompts" — the `hasattr` pattern can.

Suggest mirroring the sibling: `hasattr` guard, warn, fall back to `instance.mapped_prompt.aggregate(Max('modified_at'))`.

frontend/src/components/widgets/list-view/ListView.jsx
```diff
@@ -228,21 +269,27 @@ function ListView({
             }}
             className={`cur-pointer ${centered ? "centered" : ""}`}
```
ritwik-g: **P2 — `centered` is now a dead prop, and vertical alignment silently changed for the consumers that never opted in.**

The prop only ever worked via `.cur-pointer.centered .ant-list-item-meta` (`ListView.css:40-46`), and this PR removes `List.Item.Meta` from the markup — so both CSS rules are dead and the `centered` passed by `ToolSettings.jsx:337` / `ConnectorsPage.jsx:259` (through `ViewTools`) does nothing. Alignment is now hardcoded by the new `<Flex align="center">`, which also flips the previously top-aligned Workflows and Prompt Studio lists to center.

[verified: `List.Item.Meta` gone from the JSX; no other `.ant-list-item-meta` / `centered` references reach this widget]

Cosmetically minor since rows are single-line now, but please either wire the prop into the Flex `align` value or remove the prop and the two dead CSS rules so the API doesn't lie to consumers.

frontend/src/components/widgets/list-view/ListView.jsx
```diff
@@ -12,14 +12,34 @@ import "./ListView.css";
...
+const renderItemMetadata = (item) => (
+  <div>
+    {item?.created_at && (
+      <div>Created: {formattedDateTime(item.created_at)}</div>
+    )}
+    {item?.modified_at && (
+      <div>Modified: {formattedDateTime(item.modified_at)}</div>
+    )}
+    {item?.model && <div>Model: {item.model}</div>}
+    {item?.prompt_count !== undefined && (
```
ritwik-g: **P3 — after editing a tool from the list page, the edited row's tooltip silently loses its "Prompts" line.**

`ListOfTools.updateList` (`ListOfTools.jsx:192`) replaces the row wholesale with the PATCH response, which is serialized by `CustomToolSerializer` — it carries the timestamps but not `prompt_count`. With presence-keyed rendering, the just-edited row drops the line until a full reload while every other row keeps it. Merging instead of replacing in `updateList` (`{...item, ...data}`) preserves list-only fields.

[verified: read the edit → `updateList` path and both serializers]

backend/prompt_studio/prompt_studio_core_v2/tests/test_prompt_studio_author.py
```diff
@@ -83,3 +83,42 @@ def test_create_project_then_add_prompt(self) -> None:
         persisted = ToolStudioPrompt.objects.get(prompt_id=prompt.data["prompt_id"])
         assert persisted.tool_id_id == tool.tool_id
         assert persisted.active
+
+    def test_list_modified_at_reflects_prompt_edits(self) -> None:
```
ritwik-g: **P3 — two cheap additions would harden this test; docstring slightly overstates coverage.**

- **Zero-prompt tool:** list immediately after project creation and assert `modified_at == tool.modified_at`. This protects the `if last_prompt_modified and ...` guard — if a refactor ever drops it, `None > datetime` raises `TypeError` and the whole list endpoint 500s for any org with a fresh promptless project (the default state of every new project).
- **Tool newer than prompts:** save the tool after creating the prompt and assert the tool's own timestamp wins — the fallback branch of `get_modified_at` is currently untested.
- The docstring says "prompt edits", but the test only exercises prompt **creation**; an update-path regression would still pass. Reword or add an update step.

(The equal-timestamp flakiness point is already raised by CodeRabbit/Greptile above — not repeating it; note the final assertion would still hold even in that case, only the guard assert is exposed.)

frontend/src/components/widgets/list-view/ListView.css
```diff
@@ -47,8 +47,35 @@
...
+/* Left column: title + description stacked; shrinks with ellipsis */
+.list-view-left {
+  display: flex;
+  flex: 1 1 45%;
+  flex-direction: column;
+  gap: 4px;
+  min-width: 0;
+}
+
+/* Middle column: owner + updated, vertically centered in the entry */
```
ritwik-g: **P3 (nit)** — the entry-level vertical centering actually comes from the parent `<Flex align="center">` (`ListView.jsx:272`), not this rule; `align-items: center` here only aligns the badge and the updated-text against each other. Worth rewording so nobody deletes the parent `align` believing this class covers it.

ritwik-g (commented):
## Review — UN-3741 list metadata (no P1s; 3 P2, 3 P3 inline)

Severity: **P1** must fix / **P2** should fix / **P3** optional. Every P2 was independently verified against the code (details in each inline comment).

**Checked and cleared:**
- Backend annotation/serializer pairing is sound: `CustomToolListSerializer` is only returned for the `list` action, and `get_queryset` always annotates `_last_prompt_modified` for it; the datetime comparison is tz-aware-safe; a `SerializerMethodField` shadowing a `Meta.fields` model field is valid DRF.
- All four ListView consumers (adapters, Prompt Studio, Workflows, Connectors) get both timestamps from their list serializers — no page silently loses the new meta column. The removed CSS classes (`list-view-content`, `adapters-list-title-container`, `list-item-desc`) have no other references.
- `moment(...)` / `formattedDateTime` parse DRF's ISO-8601 output cleanly (no `DATETIME_FORMAT` override, `USE_TZ=True`) — no "Invalid date" path.
- Prompt reorders / bulk ops **do** bump `modified_at`: `BaseModelQuerySet` patches `update()`/`bulk_update()` (`utils/models/base_model.py:40-54`) and `ToolStudioPrompt` inherits that manager — so the "honest last modified" holds for those paths too.
- No existing tests were modified or deleted; the new test genuinely exercises the list-action → annotation → serializer path and fails without the fix.
- CodeRabbit's "critical" `sessionDetails` render-crash is a false positive — `ListView.jsx:85` already uses `sessionDetails?.email` and this PR doesn't touch that line. The deprecated-icon onClick behavior it flagged as major is pre-existing code moved verbatim, out of this PR's scope.

**P3 (not inline):** title type — this is customer-requested new metadata display, arguably `[FEAT]` rather than `[FIX]` per the release-notes standard; author's call since it also corrects the misleadingly stale `modified_at`.

ritwik-g (dismissed):
No blocking issues — zero P1s. The three P2s in my review (silently-ignored `?ordering=modified_at`, silent stale fallback in `get_modified_at`, dead `centered` prop) are latent/hardening/cosmetic and fine as a follow-up. Backend annotation path, all four ListView consumers, and the new test were verified against the PR head.

backend/prompt_studio/prompt_studio_core_v2/serializers.py
ritwik-g: Correction after a deeper look — this is even less severe than stated above. Pre-PR, `?ordering=modified_at` only worked for **org admins / service accounts**: the regular-user queryset ends in `.distinct("tool_id")` (`prompt_studio_core_v2/models.py:51`), and Postgres `DISTINCT ON` requires ORDER BY to start with the same expression, so the param **500'd** (`ProgrammingError`) for everyone else. Post-PR it's silently ignored for all callers — for regular users that's arguably an improvement.

With no frontend or known client sending the param, this is confirmed non-blocking. Follow-up suggestion stands: if "sort by last modified" should be a supported feature, annotate with `Greatest(...)` + declare explicit `ordering_fields` — that also fixes the pre-existing regular-user 500.

ritwik-g (approved):
Re-approving after re-examining the ordering finding: pre-PR, `?ordering=modified_at` only worked for org admins (regular users 500'd on the `distinct("tool_id")` queryset) and no client sends it — confirmed non-blocking, details in the thread on serializers.py:49. All other findings remain follow-up material (no P1s).

frontend/src/components/widgets/list-view/ListView.jsx
```diff
@@ -136,76 +159,94 @@ function ListView({
...
-                cursor: item?.is_deprecated ? "not-allowed" : "pointer",
-                opacity: item?.is_deprecated ? 0.4 : 1,
-              }}
-            />
-          </Tooltip>
-          {handleShare && (
-            <Tooltip
-              title={item?.is_deprecated ? "This adapter is deprecated" : ""}
+        {showOwner && renderOwnerBadge(item)}
+        {item?.modified_at && (
```
jaseemjaskp: [blocking] **The "Updated" block is presence-keyed, so it silently ships to Workflows and Connectors — where the value is not an honest "last modified".**

`renderMeta` gates only on `item?.modified_at`, and `ListView` is reached by four pages via `ViewTools` (`ListOfTools`, `ToolSettings`, `Workflows`, `ConnectorsPage`). Both `WorkflowSerializer` (`workflow_v2/serializers.py:46`) and `ConnectorInstanceSerializer` (`connector_v2/serializers.py:45`) use `fields = "__all__"`, and both models extend `BaseModel` (`modified_at = auto_now=True`) — so those two pages gain the label with no backend change and no test.

The PR notes these pages "gain the same display", but the value there is misleading in exactly the way this PR set out to fix elsewhere:

- **Workflows — structurally stale.** A workflow's content lives in `ToolInstance` / `WorkflowEndpoint` rows. `WorkflowEndpointViewSet` and `tool_instance_v2/views.py` never save the `Workflow` row. So a user reconfigures endpoints and tool settings today and the list still reads *"Updated 8 months ago"* — the same class of bug as the `CustomTool` one this PR fixes. Separately, `WorkflowHelper.active_project_workflow` (`workflow_helper.py:85-86`) does `is_active = True; save()`, bumping `modified_at` on *activation* rather than a content edit.
- **Connectors — bumped by automation, not humans.** `connector_auth_v2/models.py:104-114` writes `connector_instance.connector_metadata` and calls `save()` during **automatic OAuth token refresh**. A user sees "Updated 2 minutes ago" on a connector nobody has touched in months.

Presence of a field is not evidence of its meaning. Suggest making the opt-in explicit rather than inferred:

```jsx
function ListView({ ..., showModified = false, ... }) {
  const renderMeta = (item) => {
    if (!showOwner && !(showModified && item?.modified_at)) return null;
```

thread `showModified` through `ViewTools`, and pass it only from `ToolSettings` and `ListOfTools` — the two surfaces this PR actually verified. That also leaves room to add the workflow-side annotation as a follow-up instead of shipping the stale value now.

backend/prompt_studio/prompt_studio_core_v2/views.py
```diff
@@ -164,8 +164,16 @@ def get_queryset(self) -> QuerySet | None:
                 .annotate(cnt=Count("prompt_id"))
                 .values("cnt")
             )
+            # Prompt edits don't touch the CustomTool row, so surface the
+            # latest prompt modified_at for an honest "last modified" (UN-3741)
+            last_prompt_modified_sq = (
+                ToolStudioPrompt.objects.filter(tool_id=OuterRef("pk"))
+                .order_by("-modified_at")
+                .values("modified_at")[:1]
+            )
```
jaseemjaskp: [blocking] **The derived `modified_at` can travel backwards — deleting the newest prompt makes a tool look older than it did a second ago.**

The subquery takes the newest *surviving* prompt, and nothing bumps the parent row on delete: `ToolStudioPrompt` has no `save()`/`delete()` override and no `post_save`/`post_delete` receiver (verified across the `prompt_studio` tree — the only signal there is a `pre_delete` on `IndexManager`).

Failure scenario: tool row last written at T1 (rename). Prompts edited at T2 and T5. List shows "Updated" = T5. The user **deletes the T5 prompt** — a real edit, just now — and the list immediately reports T2: *"Updated 3 days ago"*. The timestamp moves backwards in response to a user action, which is a contradiction for a field named "last modified".

The root fix also collapses several other findings on this PR. Bump the parent at the source:

```python
# ToolStudioPrompt.save()/delete(), or post_save/post_delete receivers
CustomTool.objects.filter(pk=self.tool_id_id).update(modified_at=timezone.now())
```

`modified_at` then becomes honest for creates, edits **and deletes**, and the annotation, the `SerializerMethodField`, the view↔serializer coupling and the reported `?ordering=modified_at` breakage all disappear together.

If you prefer to keep deriving it, at minimum compute the max in SQL so it stays sortable — but note this still does not fix the delete case:

```python
from django.db.models.functions import Coalesce, Greatest
qs.annotate(_effective_modified_at=Greatest(
    F("modified_at"), Coalesce(Subquery(last_prompt_modified_sq), F("modified_at"))))
```

backend/prompt_studio/prompt_studio_core_v2/serializers.py
```diff
@@ -70,6 +71,15 @@ class Meta:
     def get_created_by_email(self, instance):
         return instance.created_by.email if instance.created_by else ""
 
+    def get_modified_at(self, instance):
```
jaseemjaskp: [blocking] **`modified_at` now means two different things on two endpoints of the same resource.**

`get_serializer_class` (`views.py:142-145`) routes `list` → `CustomToolListSerializer` (this derived `max(tool, latest prompt)`) and every other action → `CustomToolSerializer`, which is `fields = "__all__"` (`serializers.py:102-110`) and returns the **raw row value**.

Failure scenario: a client lists tools and sees `modified_at = T5` (from a prompt edit), then GETs that same tool and receives `modified_at = T1` (the raw row). Same field name, same resource, two semantics, and nothing on the wire tells the client which one it got — the timestamp appears to jump backwards on navigation. Any client that caches list rows and reconciles against detail responses will flip-flop.

Two ways out, both better than shadowing the model field:

1. **Don't shadow it.** Expose the derived value under its own name (`last_activity_at`), leave `modified_at` meaning "this row was written", and have the UI read the new field. The two fields then have distinct, defensible meanings on every endpoint.
2. **Make it true at the source** — the parent-row bump described on `views.py:173`. Then both serializers agree because there is only one value.

frontend/src/components/widgets/list-view/ListView.jsx
```diff
@@ -136,76 +159,94 @@ function ListView({
...
+    if (!showOwner && !item?.modified_at) {
+      return null;
+    }
     return (
-      <Flex
-        gap={20}
-        align="center"
-        justify="space-between"
-        className="list-view-container"
+      <div
```
jaseemjaskp: [low] **The meta column swallows row clicks, leaving a dead strip through the middle of every row.**

`.list-view-meta` wraps *all* its children in `onClick={(event) => event.stopPropagation()}`. Previously `renderOwnerBadge` lived in `<div className="list-view-content">`, which had no handler, so that region bubbled to the `List.Item` `onClick`.

Failure scenario: on Prompt Studio / Workflows the row click navigates; on Adapters / Connectors (`isClickable={false}`) it opens the Share modal. After this change, clicking the "Updated 2 days ago" text or the whitespace around the owner badge does **nothing**, while every pixel to its left and right responds. It reads as an unresponsive band rather than an intentional boundary.

The only child that needs click isolation is the co-owner button, and it already stops propagation itself in `handleCoOwnerClick` (line 71). Suggest dropping `onClick` and `role="none"` from this div.

frontend/src/components/widgets/list-view/ListView.css
```diff
@@ -47,8 +47,35 @@
...
+  margin-bottom: 0 !important;
+  /* keep rows uniform; full text is in the ellipsis tooltip */
+  max-width: 520px;
+}
+
+.list-view-row {
+  width: 100%;
+}
+
+/* Left column: title + description stacked; shrinks with ellipsis */
```
jaseemjaskp, line 51: [low] **This comment claims the title ellipsizes, but only the description does — a long title wraps and breaks the row uniformity the PR is after.**

`.adapters-list-title` (line ~106) sets only `font-size: 16px` — no `overflow`, `text-overflow`, or `white-space` — and `renderTitle` (`ListView.jsx:137-163`) renders a plain `Typography.Text` with no `ellipsis` prop on all three branches. Since `.list-view-left` is `flex-direction: column` with `min-width: 0`, a long title **wraps to a second line**, defeating both this comment and the "keep rows uniform" goal stated at line 51 (which is the same goal `max-width: 520px` on the description was added to protect).

Preferred fix — make the comment true, matching the description's treatment:

```jsx
<Typography.Text className="adapters-list-title" ellipsis={{ tooltip: true }}>
```

Otherwise correct the comment to say the title wraps, so the next reader doesn't trust it.

backend/adapter_processor_v2/serializers.py
```diff
@@ -170,6 +170,8 @@ class Meta(BaseAdapterSerializer.Meta):
             "adapter_name",
             "adapter_type",
             "created_by",
+            "created_at",
```
jaseemjaskp, line 173: [low] **These two fields have no test coverage, and neither does the endpoint they ship on.**

`AdapterListSerializer` appears in exactly three places repo-wide: its definition and two references in `views.py`. No test constructs it or exercises `AdapterInstanceViewSet.list` — `test_adapter_api.py` covers only `create`, and asserts against the ORM instance (`AdapterInstance.objects.get(...)`), never a serialized response.

The fields themselves are safe (`AdapterInstance` extends `BaseModel`). The risk is silence: `Meta.fields` is a hand-maintained tuple, so a merge or refactor dropping these two lines removes the "Modified" line from the UI with no test failing and no server error. Worth noting the untested blast radius is wider than the diff — `to_representation` (lines ~178-200) also injects `is_available` / `is_deprecated` / `icon` and makes an SDK call per row.

One test in `test_adapter_api.py`, reusing the existing `setUp`, would pin it:

```python
list_view = AdapterInstanceViewSet.as_view({"get": "list"})
# ... create via self.create_view as the existing test does
assert row["created_at"] and row["modified_at"]
# then rename via the API: modified_at advances, created_at does not
```

The rename leg is the part that matters — it pins `modified_at` as mutable and `created_at` as stable, which is what a `fields`-tuple typo would break.

backend/prompt_studio/prompt_studio_core_v2/tests/test_prompt_studio_author.py
```diff
@@ -83,3 +83,42 @@ def test_create_project_then_add_prompt(self) -> None:
...
+            "/api/v1/prompt-studio/",
+            {
+                "tool_name": "quote-parser",
+                "description": "extracts quote fields",
+                "author": "test",
+            },
+        )
+        assert project.status_code == status.HTTP_201_CREATED, project.data
+        tool = CustomTool.objects.get(tool_id=project.data["tool_id"])
+        prompt = ToolStudioPrompt.objects.create(
```
jaseemjaskp, line 129: [low] **One tool + one prompt means the two lines that make the annotation correct are pinned by nothing.**

(Distinct from the zero-prompt/timestamp points already raised — this is about the subquery's *discriminating* properties.)

With a single tool and a single prompt in a `TestCase` transaction, both of these regressions still pass this test:

- **Drop `filter(tool_id=OuterRef("pk"))`** (`views.py:170`) and the subquery returns the globally-latest prompt `modified_at` across every tool in every org. The test can't see it — the only prompt in the DB belongs to the only tool. Note this would be a cross-org leak of activity timing, since `for_user()` org-scoping does not reach inside a `Subquery`.
- **Flip `order_by("-modified_at")` to ascending** (`views.py:171`) and the test passes identically — with one prompt, first == last.

Fix — two tools × two prompts, with interleaved timestamps so a wrong answer is a *different* value rather than the same one:

```
tool_a ── prompt_a1 (old), prompt_a2 (mid)
tool_b ── prompt_b1 (newest in DB)
```

then `assert row_a["modified_at"] == prompt_a2.modified_at`  # not prompt_b1's

Lost correlation now fails on `prompt_a2` vs `prompt_b1`; a flipped sort fails on `prompt_a2` vs `prompt_a1`. Set timestamps explicitly via `ToolStudioPrompt.objects.filter(pk=...).update(modified_at=<fixed>)` — `BaseModelQuerySet` (`utils/models/base_model.py:40`) honours an explicit `modified_at` — which also removes the wall-clock dependence.

frontend/src/components/widgets/list-view/ListView.jsx
```diff
@@ -136,76 +159,94 @@ function ListView({
       );
     }
 
+    return title;
+  };
+
+  const renderMeta = (item) => {
+    if (!showOwner && !item?.modified_at) {
```
jaseemjaskp: [nit] **This guard can never fire.**

`showOwner` defaults to `true` (line 55) and no caller passes `false` — `ToolSettings` and `ConnectorsPage` pass `true` explicitly; `ListOfTools` and `Workflows` omit it and get the default. So `!showOwner` is always false and this never returns `null`, while an empty `.list-view-meta` flex container renders unconditionally.

Either drop it, or make the emptiness check real:

```js
if (!showOwner && !item?.modified_at && item?.prompt_count === undefined) return null;
```

frontend/src/components/widgets/list-view/ListView.jsx
jaseemjaskp: [nit] **Four fields, two presence conventions, no stated reason.**

`item?.model &&` is a truthiness test while `item?.prompt_count !== undefined` is an explicit-undefined test. So `model: ""` hides its line but `prompt_count: 0` correctly shows `Prompts: 0`. The inconsistency invites the next field to pick whichever rule the author happened to copy.

One rule handles both correctly (`0` shows, `null`/`undefined` hide):

```js
{item?.model != null && <div>Model: {item.model}</div>}
{item?.prompt_count != null && <div>Prompts: {item.prompt_count}</div>}
```

frontend/src/components/widgets/list-view/ListView.jsx
```diff
@@ -136,76 +159,94 @@ function ListView({
...
-                  item?.is_deprecated ? "disabled-icon" : ""
-                }`}
-                onClick={(event) => handleShareClick(event, item, true)}
-                style={{
-                  cursor: item?.is_deprecated ? "not-allowed" : "pointer",
-                  opacity: item?.is_deprecated ? 0.4 : 1,
-                }}
+              Updated {moment(item.modified_at).fromNow()}
+            </Typography.Text>
+            <Tooltip title={renderItemMetadata(item)}>
```
jaseemjaskp: [nit] **The tooltip body is built eagerly for every row, even while closed.**

`title={renderItemMetadata(item)}` constructs the full metadata tree on every render of every row in the list, though it's only visible on hover. antd's `Tooltip` accepts a thunk:

```jsx
<Tooltip title={() => renderItemMetadata(item)}>
```

frontend/src/components/widgets/list-view/ListView.jsx
```diff
@@ -136,76 +159,94 @@ function ListView({
...
-                key={`${item.id}-share`}
-                className={`action-icon-buttons share-icon ${
-                  item?.is_deprecated ? "disabled-icon" : ""
-                }`}
-                onClick={(event) => handleShareClick(event, item, true)}
-                style={{
-                  cursor: item?.is_deprecated ? "not-allowed" : "pointer",
-                  opacity: item?.is_deprecated ? 0.4 : 1,
-                }}
+              Updated {moment(item.modified_at).fromNow()}
```
jaseemjaskp: [nit] **A malformed `modified_at` renders the literal string "Invalid date" to the user, silently.**

The `item?.modified_at &&` guard catches *absence*, never *malformation* — a garbage string is truthy and flows straight to the DOM. Verified against the pinned `moment@2.29.4`: `"17/07/2026"`, `"not-a-date"` and `"2026-13-45T99:99:99Z"` all yield `"Invalid date"` from both `.fromNow()` and `formattedDateTime`, with nothing thrown and nothing logged. The two helpers also disagree on `0` (`""` vs `"57 years ago"`), so the row label and its own tooltip can contradict each other.

Latent today since the backend emits ISO — it goes live the moment a plugin serializer or cached payload returns a different shape. Non-ISO input also trips moment's deprecated fallback to browser-native `Date` parsing, making the result browser-dependent. Strict parsing plus a log would surface a backend regression instead of shipping the word "Invalid" to users:

```js
const m = moment(item.modified_at, moment.ISO_8601, true);
if (!m.isValid()) { logError(...); return null; }
```

Minor related point: `modified_at` is stamped server-side and read right after a mutation, so a slightly fast server clock renders "Updated **in** a few seconds". `moment.min(m, moment())` clamps it.

frontend/src/components/widgets/list-view/ListView.jsx
```diff
@@ -12,14 +12,34 @@ import "./ListView.css";
 import {
   DeleteOutlined,
   EditOutlined,
+  InfoCircleOutlined,
   QuestionCircleOutlined,
   ShareAltOutlined,
   UserOutlined,
 } from "@ant-design/icons";
+import moment from "moment";
```
jaseemjaskp: [nit] **The repo now renders "N ago" through two different date libraries.**

To be clear, `moment` is a fair choice here — it's already a dependency (`package.json:35`) and already in the bundle via `helpers/GetStaticData.js`, so this adds no weight, and there's no existing relative-time helper to reuse.

The inconsistency is that `metrics-dashboard/RecentActivity.jsx:12-20,159` already does this job on the other library (`dayjs.extend(relativeTime)` → `dayjs(item.created_at).fromNow()`). Suggest adding a `timeAgo` next to `formattedDateTime` in `helpers/GetStaticData.js` and importing it alongside the `formattedDateTime` this file already pulls in. That keeps date handling centralised the way the rest of the codebase does it, and makes a future moment→dayjs migration a one-file change rather than a grep across components.

frontend/src/components/widgets/list-view/ListView.css
```diff
@@ -83,18 +113,20 @@
   gap: 5px;
 }
 
-.list-view-content {
+.adapters-list-modified-container {
```
jaseemjaskp: [nit] **New classes use an `adapters-list-*` prefix inside a now-generic widget.**

`.adapters-list-modified-container` / `-text` / `-info-icon` are added by this PR, but the same restructure standardises on `list-view-*` for its other new classes (`list-view-row`, `list-view-left`, `list-view-meta`). This widget is shared — Prompt Studio renders through it, and `prompt_count` in `renderItemMetadata` is Prompt-Studio-only — so `adapters-list-*` is a misnomer on brand-new code. The pre-existing `adapters-list-*` classes are understandable legacy; these would be better as `list-view-modified-container` / `-text` / `-info-icon`.

frontend/src/components/widgets/list-view/ListView.css
```diff
@@ -66,11 +93,14 @@
   align-items: center;
   display: flex;
   justify-content: center;
+  /* allow the email to ellipsize instead of wrapping when the row is tight */
```
jaseemjaskp: [nit] **The comment names a failure mode that can't occur, which hides why this rule is load-bearing.**

The email can't wrap with or without `min-width: 0` — `ListView.jsx:103-106` gives `.shared-username` antd's `ellipsis={{ tooltip: true }}` (which emits `white-space: nowrap`), and line 77 caps it at `max-width: 200px`. What `min-width: 0` actually buys is letting this container shrink below its min-content width as a flex item of `.list-view-meta`, so the ellipsis engages instead of the text **overflowing and forcing the row wider**.

Worth fixing because the current wording invites deletion: a maintainer who tests for wrapping, sees none, and removes the rule will reintroduce the overflow. Suggest: `/* let this shrink below its content width so the email ellipsizes instead of forcing the row wider */`

jaseemjaskp (commented):
**PR Review Toolkit** — Code Reviewer, Silent Failure Hunter, Type Design Analyzer, PR Test Analyzer, Comment Analyzer and Code Simplifier. 14 findings below; items already raised by other reviewers are excluded.

The feature is well-motivated and the latest-prompt subquery is the right instinct — the `_prompt_count` mirroring is clean, the UN-2202 comment in the test documents a genuine trap, and the removed CSS classes were verified unreferenced. Three findings are blocking.

**Blocking**

1. **`ListView.jsx:176` — the "Updated" block is presence-keyed, so it silently ships to Workflows and Connectors.** Both serializers are `fields = "__all__"` over `BaseModel`, so both pages gain the label with no backend change and no test. On **Workflows** the value is structurally stale in exactly the way this PR fixes for Prompt Studio (content lives in `ToolInstance`/`WorkflowEndpoint`, which never write the `Workflow` row) — a user reconfigures endpoints today and reads "Updated 8 months ago". On **Connectors** it's bumped by automatic OAuth token refresh (`connector_auth_v2/models.py:114`), so "Updated 2 minutes ago" appears on connectors nobody touched. Presence of a field isn't evidence of its meaning — make the opt-in explicit.

2. **`views.py:173` — the derived timestamp travels backwards on prompt delete.** `ToolStudioPrompt` has no save/delete override and no signals, so deleting the newest prompt drops the tool's reported `modified_at` to the next-newest. The user performs an edit and the row ages by three days.

3. **`serializers.py:74` — `modified_at` means two different things on two endpoints.** `list` returns the derived max; every other action goes through `CustomToolSerializer` (`fields = "__all__"`) and returns the raw row. List then GET makes the timestamp jump backwards, with nothing on the wire to distinguish them.

**One fix collapses all three, plus the `?ordering=modified_at` breakage already reported:** bump the parent row at the source (`post_save`/`post_delete` on `ToolStudioPrompt` → `CustomTool.objects.filter(pk=...).update(modified_at=now())`). `modified_at` becomes honest for creates, edits *and* deletes; both serializers agree because there's only one value; ordering works because it's a real column again; and the annotation, the `SerializerMethodField` and the view↔serializer coupling all disappear. If you'd rather keep deriving, computing the max in SQL via `Greatest`/`Coalesce` restores sortability — but note it does **not** fix the delete case (#2).

**Low** — a dead click strip through the middle of every row (`ListView.jsx:170`); the title doesn't ellipsize despite the comment, so long titles wrap and break row uniformity (`ListView.css:59`); the adapter fields ship with zero coverage of the list endpoint (`adapter_processor_v2/serializers.py:173`); the new test's single-tool/single-prompt fixture can't detect a lost `OuterRef` correlation or a flipped sort (`test_prompt_studio_author.py:106`).

**Nits** — dead `showOwner` guard, mixed presence conventions, eagerly-built tooltip, unguarded "Invalid date" rendering, moment/dayjs split, `adapters-list-*` prefix on generic classes, and a misleading `min-width: 0` comment.

Also worth folding into the `centered` cleanup already raised: `ListView.css:40-46` still targets `.ant-list-item-meta`, which no longer renders anywhere now that `List.Item.Meta` is gone — both rules are unreachable. (Outside the diff, so no inline comment.)

backend/prompt_studio/prompt_studio_core_v2/tests/test_prompt_studio_author.py
```diff
@@ -83,3 +83,33 @@ def test_create_project_then_add_prompt(self) -> None:
...
+            created_by=self.user,
+        )
+        prompt = ToolStudioPrompt.objects.create(
+            tool_id=tool,
+            prompt_key="quote_number",
+            prompt="What is the quote number?",
+            prompt_type=ToolStudioPrompt.PromptType.PROMPT,
+            sequence_number=1,
+        )
+        assert prompt.modified_at > tool.modified_at
```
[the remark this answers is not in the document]
athul-rs: Fixed in f8ca180f (timestamp pinned via .update()), then superseded in 057101b2: the annotation is gone entirely — prompt writes now bump the parent CustomTool row at the source, and the tests backdate deterministically with .update().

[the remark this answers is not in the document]
athul-rs: Fixed in f8ca180f (pinned via .update() to a known instant), then superseded in 057101b2 — the subquery is deleted; the parent-bump tests backdate explicitly so no assertion depends on wall-clock advancement.

frontend/src/components/widgets/list-view/ListView.jsx
```diff
@@ -71,94 +91,115 @@ function ListView({
...
+              item?.is_deprecated ? "disabled-icon" : ""
+            }`}
+            onClick={(event) => handleShareClick(event, item, true)}
+            style={{
+              cursor: item?.is_deprecated ? "not-allowed" : "pointer",
+              opacity: item?.is_deprecated ? 0.4 : 1,
+            }}
+          />
+        </Tooltip>
+      )}
```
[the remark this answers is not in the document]
athul-rs: Fixed in f8ca180f — both handlers early-return when item.is_deprecated, matching the disabled styling.

backend/prompt_studio/prompt_studio_core_v2/serializers.py
athul-rs: Fixed in 057101b2 by taking the root fix from the sibling thread: prompt writes bump the parent row (ToolStudioPrompt.save/delete), so modified_at is a plain model field again — orderable by DRF's OrderingFilter with no annotation involved.

backend/prompt_studio/prompt_studio_core_v2/serializers.py
athul-rs: Moot in 057101b2 — get_modified_at is deleted along with the annotation; modified_at is served as a plain model field, so there is no fallback path left to get wrong. (The hasattr+warn pattern survives on get_prompt_count, unchanged.)

frontend/src/components/widgets/list-view/ListView.jsx
athul-rs: Fixed in 057101b2 — centered is wired into the row Flex (align={centered ? "center" : "flex-start"}), restoring the pre-PR opt-in semantics for ToolSettings/ConnectorsPage, and both dead .ant-list-item-meta rules plus the dead className are removed.

frontend/src/components/widgets/list-view/ListView.jsx
athul-rs: Fixed in 057101b2 — updateList now merges ({...item, ...data}) so list-only fields like prompt_count survive an edit. With the parent-bump change the PATCH response's modified_at is also already correct.

backend/prompt_studio/prompt_studio_core_v2/tests/test_prompt_studio_author.py
athul-rs: Addressed in 057101b2 by the redesign: the promptless-project case is pinned (list serves the tool's own timestamp — no annotation, so no None>datetime hazard), prompt update is exercised directly (edit bumps the parent), and the docstring now describes exactly what runs. The tool-newer-than-prompts fallback branch no longer exists to test.

frontend/src/components/widgets/list-view/ListView.css
athul-rs: Fixed in 057101b2 — comment now says the column's vertical position comes from the row Flex align in ListView.jsx, and this rule only aligns badge vs updated-text.

frontend/src/components/widgets/list-view/ListView.jsx
athul-rs: Fixed in 057101b2 exactly as suggested: showModified prop (default false) threaded through ViewTools and passed only by ToolSettings and ListOfTools — Workflows and Connectors no longer inherit the label by field presence. The workflow-side annotation stays open as a follow-up if that page ever wants the label honestly.

backend/prompt_studio/prompt_studio_core_v2/views.py
athul-rs: Took the root fix in 057101b2: ToolStudioPrompt.save() and delete() bump the parent CustomTool row (queryset update, no recursion), the subquery/annotation and SerializerMethodField are deleted, and a test pins that DELETE advances the tool's modified_at — it can no longer travel backwards. Cross-tool isolation is pinned too. As you predicted, the ordering and two-semantics findings collapsed with it. One caveat noted in code: the collateral renumbering rows in reorder go through bulk_update and don't bump, but the dragged prompt's save() already bumps once per reorder.

backend/prompt_studio/prompt_studio_core_v2/serializers.py
athul-rs: Resolved by the source fix in 057101b2 (option 2): there is only one value now — list and detail both serve the raw row, which prompt writes keep honest. Incidental bonus: the list endpoint's modified_at is now rendered by DRF's DateTimeField, so its wire format finally matches created_at.

frontend/src/components/widgets/list-view/ListView.jsx
athul-rs: Fixed in 057101b2 — the onClick/role are gone from .list-view-meta; only the co-owner button isolates itself, everything else bubbles to the row.

frontend/src/components/widgets/list-view/ListView.css
athul-rs: Fixed in 057101b2 with the preferred option — ellipsis={{tooltip:true}} on all three title branches, plus min-width:0 for the icon-row flex case so the ellipsis can actually engage.

backend/adapter_processor_v2/serializers.py
athul-rs: Fixed in 057101b2 — test_list_serves_timestamps_and_rename_bumps_modified exercises AdapterInstanceViewSet.list end-to-end: both fields present, and the rename leg pins modified_at as mutable (backdated first so the bump is strict) with created_at stable.

backend/prompt_studio/prompt_studio_core_v2/tests/test_prompt_studio_author.py
athul-rs: Superseded in 057101b2 — the subquery is deleted (parent-bump at source), so the correlation and ordering lines no longer exist. The property you wanted pinned survives as test_prompt_bump_only_touches_its_own_tool: a prompt write on tool A must not move tool B, with explicit .update() timestamps.

frontend/src/components/widgets/list-view/ListView.jsx
athul-rs: Fixed in 057101b2 — the guard is now real: renderMeta computes the (opt-in, validity-checked) updated label first and returns null when neither it nor the owner badge will render.

frontend/src/components/widgets/list-view/ListView.jsx
athul-rs: Fixed in 057101b2 — both lines use != null, so 0 renders and null/undefined hide; the comment above now states the convention.

frontend/src/components/widgets/list-view/ListView.jsx
athul-rs: Fixed in 057101b2 — title={() => renderItemMetadata(item)}.

frontend/src/components/widgets/list-view/ListView.jsx
athul-rs: Fixed in 057101b2 via a shared timeAgo helper: strict ISO_8601 parsing (malformed input hides the label entirely instead of rendering "Invalid date") and clamped with moment.min(parsed, now) so a fast server clock can't produce "in a few seconds". No logError exists at this layer, so silent-hide is the chosen failure mode — the tooltip's formattedDateTime already returns "" on the same inputs, so label and tooltip can no longer contradict.

frontend/src/components/widgets/list-view/ListView.jsx
athul-rs: Fixed in 057101b2 — timeAgo lives in helpers/GetStaticData.js next to formattedDateTime (that helper file is moment-based, so no new library surface), and ListView imports it instead of importing moment directly. A future moment→dayjs migration is now a one-file change for this label.

frontend/src/components/widgets/list-view/ListView.css
athul-rs: Fixed in 057101b2 — the three new classes are renamed to list-view-modified-container/-text/-info-icon; pre-existing adapters-list-* classes left as legacy.

frontend/src/components/widgets/list-view/ListView.css
athul-rs: Fixed in 057101b2 with your suggested wording — the comment now explains the rule lets the container shrink so the email ellipsizes instead of forcing the row wider, and notes the email itself can never wrap.

backend/prompt_studio/prompt_studio_v2/models.py
```diff
@@ -153,6 +154,32 @@ class RequiredType(models.TextChoices):
...
+        super().save(*args, **kwargs)
+        self._touch_tool()
+
+    def delete(self, *args, **kwargs):
+        # Capture before delete — SET_NULL FKs survive, but be defensive
+        tool_pk = self.tool_id_id
+        result = super().delete(*args, **kwargs)
+        if tool_pk:
+            CustomTool.objects.filter(pk=tool_pk).update(modified_at=timezone.now())
+        return result
```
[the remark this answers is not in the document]
athul-rs: Fixed in ea3b3d32 — verified live first: a prompts-clearing sync left modified_at byte-identical. sync_prompts now bumps the tool explicitly when the queryset delete removed rows, pinned by test_sync_prompts_clear_bumps_tool_modified_at (clearing sync with no tool-settings payload → bump asserted). Also re-swept both repos for other ToolStudioPrompt queryset deletes: this was the only one.

backend/prompt_studio/prompt_studio_v2/models.py
```diff
@@ -153,6 +154,32 @@ class RequiredType(models.TextChoices):
...
+        """Bump the parent tool's modified_at so it reflects prompt activity.
+
+        Prompt writes never touch the CustomTool row, so without this the
+        tool's "last modified" goes stale (UN-3741) — and a derived
+        max(tool, prompts) can travel backwards on prompt delete. A
+        queryset update bypasses CustomTool.save() (no auto_now recursion,
+        no signals).
+        """
+        if self.tool_id_id:
+            CustomTool.objects.filter(pk=self.tool_id_id).update(
```
jaseemjaskp: **`CustomTool.objects` is org-scoped, so this bump silently no-ops the moment a prompt is written outside request context.**

`CustomTool.objects` is `CustomToolModelManager(DefaultOrganizationManagerMixin, BaseModelManager)` (`prompt_studio_core_v2/models.py:31`). The mixin's `get_queryset()` appends `.filter(organization=UserContext.get_organization())` (`utils/models/organization_mixin.py:28-30`), and `UserContext.get_organization()` reads a **thread-local** that returns `None` when unset. So this expands to:

```sql
UPDATE custom_tool SET modified_at = %s WHERE tool_id = %s AND organization_id IS NULL
```

Zero rows matched. `.update()` returns `0`, raises nothing, and the return value is discarded — the bump just doesn't happen, with no signal.

To be fair: I traced every current `ToolStudioPrompt` write path and they all run inside an org-scoped request (DRF create/update/destroy, `import_prompts`, `reorder_prompts_helper`), so **this is latent, not a live bug today**. I'm raising it because the exposure is structural rather than hypothetical: this is a *model-level* invariant that quietly depends on ambient request state, so it's silently untrue from the Django admin (`prompt_studio_v2/admin.py:5`), a shell, a management command, or any future Celery task. Prompt writes moving to a worker is not exotic. And the docstring right above reinforces the blind spot — it carefully explains why the queryset form is safe (no `auto_now` recursion, no signals) but never mentions the manager is filtered, so the next reader has no reason to suspect it.

The fix is one token — `_base_manager` is Django's unfiltered manager and is the documented idiom for exactly this (internal FK-following writes that must not inherit a filtering default manager):

```python
CustomTool._base_manager.filter(pk=self.tool_id_id).update(modified_at=timezone.now())
```

Same change at `prompt_studio_helper.py:3111`. (`_base_manager` is a plain `Manager`, so `BaseModelQuerySet.update()`'s `setdefault` doesn't apply — but `modified_at` is passed explicitly here, so behaviour is identical.)

**Related, and why this matters more than it looks:** `delete()` (line 179-180) inlines its own copy of this statement rather than calling `_touch_tool()`. The two are behaviourally identical today — I checked, Django's collector nulls only the instance's PK, never its FK attnames, so `self.tool_id_id` survives `super().delete()` (which incidentally makes the "SET_NULL FKs survive, but be defensive" comment on line 176 cite the wrong mechanism; `on_delete` governs what happens to prompts when the *tool* is deleted, not this). But that duplication is a live trap for exactly this fix: applying `_base_manager` to `_touch_tool` would fix `save()` and silently miss `delete()`. Collapse it first:

```python
def delete(self, *args, **kwargs):
    result = super().delete(*args, **kwargs)
    self._touch_tool()
    return result
```

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -3103,6 +3104,13 @@ def sync_prompts(tool: CustomTool, import_data: dict, user) -> dict:
         with transaction.atomic():
             # Delete all existing prompts
             deleted_count, _ = ToolStudioPrompt.objects.filter(tool_id=tool).delete()
+            if deleted_count:
```
jaseemjaskp, line 3110: **`deleted_count` is inflated by cascaded rows — the sync API reports a prompt count that counts prompt *outputs*, and the new test pins the wrong number.**

The guard you added here is correct, so the bump is fine — `deleted_count` is non-zero iff at least one prompt was deleted. The problem is the value itself, which this PR now leans on and asserts against.

`QuerySet.delete()` returns `(total_across_all_models, {label: per_model_count})` — the **total**, not the prompt count. `PromptStudioOutputManager.prompt_id` is `on_delete=CASCADE` (`prompt_studio_output_manager_v2/models.py:57-60`), so every prompt that has ever been run drags its output rows into that total. Syncing a project with 3 prompts × 6 documents reports `prompts_deleted: 21`.

That value isn't internal — it's returned to the API client: `{"prompts_deleted": deleted_count, ...}` at line 3143, merged into the response at `views.py:1357`. So a user syncing a project that has been run sees a wildly wrong "prompts deleted" figure.

The count line itself predates this PR, but the new test at `tests/test_prompt_studio_author.py:202` asserts `result["prompts_deleted"] == 1` and passes **only** because the fixture prompt was never run and has zero outputs. So the test now certifies the bug — it reads as coverage of the count while pinning nothing.

```python
_, per_model = ToolStudioPrompt.objects.filter(tool_id=tool).delete()
deleted_count = per_model.get(ToolStudioPrompt._meta.label, 0)
```

Worth adding a `PromptStudioOutputManager` row to the fixture in that test so the assertion actually discriminates — otherwise it keeps passing whichever way this goes.

backend/prompt_studio/prompt_studio_core_v2/views.py
```diff
@@ -164,8 +164,11 @@ def get_queryset(self) -> QuerySet | None:
                 .annotate(cnt=Count("prompt_id"))
                 .values("cnt")
             )
+            # modified_at needs no annotation: prompt writes bump the parent
+            # row at the source (ToolStudioPrompt.save/delete), so the plain
+            # field is honest and stays orderable (UN-3741)
```
jaseemjaskp: **"the plain field is honest" is asserted as a general property, but it holds only for *prompt* writes — and the label this drives doesn't say "prompts".**

Two first-class project mutations don't bump `CustomTool.modified_at`, because neither child model has the `_touch_tool` treatment:

- **`ProfileManager`** (`prompt_profile_manager_v2/models.py:126` — has the `prompt_studio_tool` FK, no `save()`/`delete()` override). Changing the LLM, vector store, chunk size or retrieval strategy leaves the tool untouched. Also `views.py:969` does `ToolStudioPrompt.objects.filter(tool_id=...).update(profile_manager=...)` on first-profile creation — a real write to every prompt in the project, via `QuerySet.update()`, so no bump.
- **`DocumentManager`** (`prompt_studio_document_manager_v2/models.py:22`). Uploading or deleting a document doesn't bump either.

User-visible result: change a project's LLM adapter, reload the list, and it still reads "Updated 3 weeks ago" — with the tooltip stating a specific, confidently wrong `Modified:` datetime. That's the same failure UN-3741 exists to fix, reached through a different child table.

**And it moves in the other direction too:** `views.py:1178-1179` does `custom_tool.save(update_fields=["last_exported_at"])`, and `BaseModel.save()` injects `modified_at` into any non-empty `update_fields` (`utils/models/base_model.py:114-115`). So **clicking Export makes a project read "Updated a few seconds ago"** for an action that modifies nothing. That mechanism predates this PR, but this PR is what makes it visible — `modified_at` wasn't served on the list page before. It directly undercuts the feature being shipped. If the bump isn't wanted there, preserve it explicitly:

```python
CustomTool._base_manager.filter(pk=custom_tool.pk).update(
    last_exported_at=timezone.now(), modified_at=custom_tool.modified_at
)
```

(The adapter side has the same shape: `manage_deprecated_adapters.py:163,197` save with `update_fields`, so an operator deprecating an adapter makes it read "Updated just now".)

I don't think any of this has to block the merge — but the comment shouldn't tell the next maintainer the field is trustworthy in general, because someone adding a third child table or bulk path will read this and assume the invariant is self-maintaining. Either extend the bump to `ProfileManager`/`DocumentManager`, or scope the claim honestly:

```python
# modified_at needs no annotation: instance-level prompt writes bump the parent
# row at the source (ToolStudioPrompt.save/delete), plus the explicit bump in
# sync_prompts, so the plain field is orderable (UN-3741).
# Caveat: only *prompt* writes bump. ProfileManager/DocumentManager edits and
# queryset-level prompt writes (views.py:969) do NOT — any new path must bump
# CustomTool itself.
```

Worth a deliberate call either way rather than inheriting it by accident.

jaseemjaskp (commented):
Second pass. The parent-bump-at-source redesign is the right call and the backend tests are genuinely good — `test_prompt_bump_only_touches_its_own_tool` is the one most people omit, and it's the one that would catch a `_touch_tool` that lost its `pk` filter. Three findings that clear the bar; the rest of what I checked came back clean (details below).

Verified clean, so nobody re-treads them: reorder **does** bump (the `bulk_update` at `prompt_studio_v2/helper.py:112` looks like a bypass, but `reorder_prompts_helper` finishes with `prompt_instance.save()` at `helper.py:64`); `import_prompts` uses `objects.create()` so the sync ADD path bumps via `save()`; the deleted CSS classes (`.list-view-content`, `.adapters-list-title-container`, `.list-item-desc`) have zero references left repo-wide; `centered` is still live and correctly rewired to the row `Flex` align; `AdapterListSerializer` leaks nothing (secrets live in `adapter_metadata_b`, never in that tuple), and the detail serializer already exposed both fields via `fields = "__all__"`, so list is now a strict subset. No deadlock risk on the N-prompt import loop either — I looked for an inverted lock order and there isn't one.

backend/prompt_studio/prompt_studio_v2/models.py
athul-rs: Fixed in 5cd36425 — `_touch_tool` uses `CustomTool._base_manager` (matching the tenant_account_v2 / platform_api precedent), pinned by test_prompt_bump_survives_missing_org_context which clears the thread-local org and asserts the bump still lands. Also took both secondary points: `delete()` now calls `_touch_tool()` after `super().delete()` (single copy, no drift trap) and the wrong-mechanism comment is replaced with the collector-nulls-only-the-PK fact. Same `_base_manager` change applied to the sync bump.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
athul-rs: Fixed in 5cd36425 — `deleted_count` now comes from `per_model.get(ToolStudioPrompt._meta.label, 0)`, and the sync test fixture carries a PromptStudioOutputManager row so the assertion discriminates: with the old code it reports 2, the test demands 1. Good catch that my test was certifying the bug.

backend/prompt_studio/prompt_studio_core_v2/views.py
athul-rs: Took the deliberate-call option in 5cd36425: the comment is scoped honestly per your suggested text (only prompt writes bump; ProfileManager/DocumentManager and queryset-level prompt writes do not), and extending the bump — plus the export-path and adapter-deprecation spurious-bump nuances you listed — is filed as UN-3759 rather than growing this PR further. Agreed it shouldn't block the merge.

chandrasekharan-zipstack (approved):
Consider making code comments concise and generic and removing the JIRA dependency if its not truly required

jaseemjaskp (approved):
Six-agent PR Review Toolkit pass (code review, silent-failure, type design, tests, comments, simplification) over the backend bump logic and the ListView refactor. **No blocking findings — approving.**

I verified the load-bearing claims independently rather than trusting them:

- **Deletion collector** — Django 4.2.30 `deletion.py` nulls only `model._meta.pk.attname` (lines 456, 519-521); `tool_id_id` survives `super().delete()`, so the delete-bump works. Comment is accurate.
- **`_base_manager` / org scope** — `CustomToolModelManager` mixes in `DefaultOrganizationManagerMixin` (thread-local scoped); no `base_manager_name` anywhere in the MRO, so `_base_manager` is a plain unfiltered `Manager`. `tool_id` is a UUID pk. All three sub-claims hold, and `test_prompt_bump_survives_missing_org_context` genuinely pins it.
- **Prompt reorder** — flagged by one agent as bypassing the bump via `bulk_update`; **that is wrong**. `reorder_prompts_helper` ends in `prompt_instance.save()` (`prompt_studio_v2/helper.py:64`), which fires `_touch_tool()`. Reorder does bump.
- **`per_model.get(ToolStudioPrompt._meta.label, 0)`** — `Collector.delete()` keys by `_meta.label`; the `0` default is only hit at zero rows. Correct, and the sync test builds a real CASCADE fixture so the assertion actually discriminates 1 from 2.
- **Backdating helper** — works only because `BaseModelQuerySet.update` uses `setdefault`; had it been an unconditional assign, every `>` assertion would be vacuous. It is not. Tests are honest.
- **CSS** — `.list-view-content`, `.adapters-list-title-container`, `.list-item-desc` have zero remaining references; `centered` is still honored via the new Flex `align`.

Non-blocking notes, no action needed for this PR:

1. `views.py:971` (`.update(profile_manager=...)` on first-profile creation) bypasses the bump, leaving a stale label on that one path. Pre-existing, display-only, self-healing on the next prompt edit, and the `views.py:167-171` comment already warns about this class.
2. Source-bumping is prospective only — existing projects keep their current `modified_at` until the next prompt write. Worth a thought, but the field never included prompt activity before, so this is strictly an improvement.
3. `sync_prompts` fires N+2 writes to one `CustomTool` row per sync. Real write amplification, but prompts-per-tool is in the tens and sync is infrequent — not a throughput concern.
4. `timeAgo()` is untested and the harness exists (`vitest`, sibling `helpers/requestId.test.js`). The strict-parse and `moment.min` clamp are exactly the non-obvious bits a future cleanup would strip. Cheap to add whenever.

**One thing worth fixing outside the diff: the PR description is stale.** It still says the Prompt Studio list returns `modified_at` as `max(tool.modified_at, latest prompt.modified_at)` via a correlated subquery, and cites a test `test_list_modified_at_reflects_prompt_edits` that no longer exists. The implementation moved to source-bumping. Worth updating before merge so the description matches what ships.

PR #299, opened 2024-04-29
[FEAT] Tool Export Validation

harini-venkataraman:
## What
This PR enables support for exporting only valid tools that have prompts and that are executed. 
...
## Why
This prevents export of dummy tools developed using Prompt Studio.
...
## How
Adding validations to see if the Output Manager has a valid field against the prompt.
...

## Can this PR break any existing features. If yes please list of possible items. If no please explain why. (PS: Admins do not merge the PR without this section filled)
1. Export validation is an additional filtering check and does not break features.
2. Export functional logic has these changes made, but will prevent export only if there is not valid prompt.

...
## Database Migrations
Not applicable.

...
## Env Config
Not applicable.
... 

## Notes on Testing
1. Test by exporting a properly executed tool.
2. Create a prompt with no text prompt entered, try exporting.
3. Create a prompt and without executing it, try exporting.

...
## Screenshots
Screenshot from 2024-04-29 16-13-53
GT

Screenshot from 2024-04-29 15-59-04

image

## Checklist

I have read and understood the [Contribution Guidelines]().

PR #2183, opened 2026-07-16
UN-3739 [FIX] Prompt Studio owner-access check validates requester, not just profile creator

athul-rs:
## What

- Prompt Studio's `validate_profile_manager_owner_access()` now receives the **requesting user** (`request_user`, the `User` object) and passes immediately when the requester is an org admin.
- Adds a **service-account bypass** for the profile creator (platform-API-created projects).
- Denial messages now name the **profile creator** who actually lacks access (flagging offboarded creators explicitly) with remediation steps, instead of the misleading "You do not have access"; also fixes the message being raised as a 1-tuple.
- Denials log both identities: profile creator and requester.

## Why

An org admin indexing a Prompt Studio project could get `Permission Error: You do not have access to <adapter>`. The check validated only the profile **creator's** adapter access — the requester was never passed in — so admins were denied on any project whose profile creator lost adapter access (share revoked, creator offboarded — offboarding auto-wipes share rows — or profile created by a service account, which holds no shares by design). The UN-3479 admin bypass (#1993) couldn't help: it was applied to the creator, the only identity the function received. This contradicts the admin model everywhere else (admins see all adapters and are hidden from share dialogs since UN-3318), and the only workaround was sharing adapters org-wide.

## How

- `validate_profile_manager_owner_access(profile_manager, request_user=None)` — evaluation order: requester-is-admin → creator None → creator service-account → creator admin → creator has all-4 adapter access → else deny.
- The requester is plumbed as a **separate `request_user: User` parameter** (the object views already hold — no re-resolution from a non-unique `user_id` string) through all four view entry points (`build_index_payload`, `build_fetch_response_payload`, `build_bulk_fetch_response_payload`, `build_single_pass_payload`) and their internal chains; views pass `request.user`.
- On the four builders `request_user` is **keyword-only with no default** — dropping the plumb is a `TypeError`, not a silent revert to pre-fix behavior; a signature test pins this, and each builder has a forwarding test.
- Deliberately separate from the existing `user_id` parameter, which is the **file-path owner** (`tool.created_by.user_id`) and must not change — a regression test pins that the two identities are never conflated.

- No. The requester parameter is deliberately **two-tier**: on the four view-facing builders `request_user` is **required keyword-only with no default** — omitting it is a `TypeError`, so a dropped or refactored-away plumb fails loudly instead of silently reverting UN-3739 (do **not** re-add `= None` there; a signature test pins this). On the internal/worker-path helpers (`validate_profile_manager_owner_access`, `_build_summarize_params`, `index_document`, `prompt_responder`, and the single-prompt/single-pass chain) it stays `None`-defaulted, so callers without a request context get exactly the pre-fix validation behavior. The revocation guard is kept — a non-admin requester is still denied when the profile creator lacks adapter access — and delegated use (shared-project users piggybacking on the creator's adapter access) is unchanged and covered by tests.
- **Service-account-creator bypass — intended semantic (sign-off):** a profile created by a platform-API service account passes the guard for any requester. This is a deliberate authz decision, not a side effect: platform-API profiles hold no adapter shares by design (#1987 trusts service accounts on every other permission path), and only org admins can mint API keys, which bounds the reach. The interaction with `created_by=SET_NULL` on key deletion is pre-existing and now tracked as UN-3750.
- Note: denial messages are PII-free — the profile creator is referenced generically ("another user") and identified only in server logs; remediation points at an org admin, since after the requester-admin bypass only non-admin collaborators can hit these errors.

## Database Migrations

- None.

## Env Config

- None.

## Relevant Docs

- N/A

## Related Issues or PRs

- [UN-3739](https://zipstack.atlassian.net/browse/UN-3739)
- Related: UN-3479 (#1989/#1993 admin bypass — applied to creator only), UN-3318 (#2000 admins hidden from share dialog), #1987 (service-account bypasses on request-layer paths)
- Follow-up filed from review: [UN-3750](https://zipstack.atlassian.net/browse/UN-3750) — `_BUSINESS_APP_LABELS` omits `prompt_profile_manager_v2`, so API-key deletion NULLs profile creators

## Dependencies Versions

- No changes.

## Notes on Testing

- New unit suite `test_validate_profile_manager_owner_access.py` covers the full case matrix: admin requester passes on lapsed-creator profiles, non-admin requester still blocked, service-account creator passes, admin creator passes, `created_by=None` passes, delegated use passes, plus message contract (names creator + adapter, flags former members, plain-string detail).
- `test_build_index_payload.py::TestOwnerAccessPlumbing` pins that the validator receives the `request_user` object, distinct from the file-path `user_id`.
- 59/59 backend tests pass at HEAD (`prompt_studio/`, `adapter_processor_v2/`); ruff 0.3.4 check + format clean.
- QA repro/verify steps (revoke-share, offboarded-creator, and API-created variants) documented in UN-3739.

[UN-3739]: https://zipstack.atlassian.net/browse/UN-3739?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -183,105 +183,111 @@ def validate_adapter_status(
...
 
+    @staticmethod
+    def _is_org_admin_by_user_id(user_id: str | None) -> bool:
+        """Return True if ``user_id`` resolves to an org member with admin role."""
+        if not user_id:
+            return False
+        member = OrganizationMemberService.get_user_by_user_id(user_id)
+        return bool(
+            member and OrganizationMemberService.is_user_organization_admin(member.user)
+        )
```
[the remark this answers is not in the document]
athul-rs: Declining this one deliberately: the suggested short-circuit drops the service-account exclusion that `is_user_organization_admin` enforces (service accounts must not be treated as org admins) and duplicates admin-role resolution outside its single home. The cost is two indexed PK lookups on a request that goes on to do extraction/LLM work — negligible. Keeping the delegation preserves one source of truth for "is this user an org admin".

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -1563,6 +1593,7 @@ def _execute_single_prompt(
         document_id,
         run_id,
         profile_manager_id,
+        request_user_id=None,
     ):
         prompt_instance = PromptStudioHelper._fetch_prompt_from_id(id)
```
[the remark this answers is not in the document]
athul-rs: Fixed in 9d33951e — annotated `request_user_id: str | None = None`.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -1668,6 +1700,7 @@ def _execute_prompts_in_single_pass(
         org_id,
         document_id,
         run_id,
+        request_user_id=None,
     ):
         prompts = PromptStudioHelper.fetch_prompt_from_tool(tool_id)
```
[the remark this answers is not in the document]
athul-rs: Fixed in 9d33951e — annotated `request_user_id: str | None = None`.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -183,105 +183,111 @@ def validate_adapter_status(
...
+            )
+        else:
+            error_msg = (
+                f"Permission Error: {profile_ref}, who no longer has access to"
+                f" the adapter '{denied_names}'. Re-share the adapter with"
+                f" them, share it with everyone, or recreate the profile using"
+                f" adapters you have access to."
+            )
+
+        raise PermissionError(error_msg)
```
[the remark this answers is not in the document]
athul-rs: Fixed in 9d33951e. Agreed — and the point is sharper than stated: after the requester-admin bypass, admins never see this error, so the email's only audience was non-admin collaborators who can't act on it. Messages now reference the creator generically, remediation points at an org admin, and the creator's identity stays in the server-side log line. A test now pins that no email appears in user-facing messages.

ritwik-g (commented):
Reviewed against `fcf48ba5`. I ran an independent RCA of the underlying bug first (call chain, permission model, `created_by` provenance, git history of the check), then an adversarial verification pass over this diff.

**Verdict: the fix is correct and lands at the right layer.** Both real gaps are closed — the admin *requester* bypass (`prompt_studio_helper.py:227`) and the service-account *creator* bypass (`:234`). The SA bypass is load-bearing rather than belt-and-braces: `is_user_organization_admin` returns `False` for service accounts (`organization_member_service.py:29-30`), so the creator-admin bypass could never have fired for this customer's profile.

Particular credit for carrying the requester as a **separate** `request_user_id` instead of reusing `user_id`. `user_id` is `tool.created_by.user_id` — the file-path owner — at all four entry points (`views.py:451/530/683/789`), so the ticket's own "Proposed fix" wording ("call sites already have `user_id`") would have re-validated the creator and no-oped on the exact reported scenario. `test_build_index_payload.py:212` pinning `request_user_id != user_id` is the right guard against that regressing.

The tests are genuinely behavioral, not theatre — I checked by installing `main`'s pre-fix body behind the new signature: 5 of 10 fail, including both bypass tests and all three message tests. The 1-tuple `error_msg` defect is real (DRF renders `.detail` as a JSON array; confirmed by executing against the pinned DRF 3.17.1), and `isinstance(exc.detail, str)` is the only assertion that catches it — `str(exc)` on a list-detail still contains the email substring. Good catch.

Comments below: one **P2** worth closing pre-merge (a test gap that makes future breakage silent), one sign-off ask, and three follow-ups. Nothing blocking beyond the P2.

Nit: "49/49 backend tests pass" in the description doesn't match a reproducible scope (I get 15 for the subtree, 137 for the backend unit tier).

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -706,6 +719,7 @@ def build_fetch_response_payload(
         document_id: str,
         run_id: str,
         profile_manager_id: str | None = None,
+        request_user_id: str | None = None,
```
ritwik-g: **P2 — `request_user_id` defaults to `None` on every builder, and only 1 of 4 has a forwarding test.**

`build_fetch_response_payload` (here), `build_bulk_fetch_response_payload` (`:943`) and `build_single_pass_payload` (`:1135`) all declare `request_user_id: str | None = None`, and `_is_org_admin_by_user_id(None)` returns `False` (`:189-190`). So a dropped or wrong plumb raises no TypeError, logs nothing, and fails no test — it silently reverts to exact pre-fix behavior. Only `build_index_payload` has a forwarding test (`test_build_index_payload.py:203-213`); grep over the tests dir returns zero for the other three.

Failure scenario: someone later swaps `request_user_id=request.user.user_id` back to `request_user_id=user_id` at `views.py:613` — the two params sit adjacent — or drops it entirely. CI stays green, the admin is denied again, and UN-3739 reopens as a "new" bug.

Suggest: make `request_user_id` keyword-only with **no default** on the four builders — free today, their only callers repo-wide are `views.py:445/603/716/824` and there are zero in unstract-cloud — which turns silent degradation into a loud TypeError. Plus 3 forwarding tests reusing the existing `validate_owner_mock` pattern (~15 lines).

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -183,105 +183,111 @@ def validate_adapter_status(
...
-                adapter_names.add(profile_manager.x2text.adapter_name)
-            if len(adapter_names) > 1:
-                error_msg = (
-                    f"Multiple permission errors were encountered with {', '.join(adapter_names)}",  # noqa: E501
-                )
-            else:
-                error_msg = (
-                    f"Permission Error: You do not have access to {adapter_names.pop()}",  # noqa: E501
-                )
+        if getattr(owner, "is_service_account", False):
```
ritwik-g, line 238: **P3 — worth one explicit line of sign-off in the PR body.**

This is the only new early-return that admits a **non-admin** requester who was previously denied, and it passes regardless of which adapters the profile references or who owns them. The reported bug is fully fixed by the requester-admin bypass at `:227` alone — this is additional reach.

Scoping it honestly, because I initially thought it was worse than it is: the "post-migration the guard is off org-wide" worry does **not** hold — if every resource lands SA-owned then the adapters do too, so `adapter.created_by == owner` (`:200`) would pass anyway. The real reach is the platform-API path only: an admin-issued key creates profiles referencing adapters it doesn't own. And a non-admin can't mint a service account (`platform_api/permissions.py:11` gates key creation on `IsOrganizationAdmin`), which bounds it further.

So: defensible. But it's a permanent authz change rather than a bug fix, and the docstring alone isn't sign-off. A line in the PR body confirming this is the intended semantic would close it.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -183,105 +183,111 @@ def validate_adapter_status(
...
+              one or more adapters and no bypass applies.
         """
-        profile_manager_owner = profile_manager.created_by
-        if profile_manager_owner is None:
+        if PromptStudioHelper._is_org_admin_by_user_id(request_user_id):
             return
 
-        if OrganizationMemberService.is_user_organization_admin(profile_manager_owner):
+        owner = profile_manager.created_by
+        if owner is None:
```
ritwik-g, line 235: **Pre-existing, not this diff's doing — but this PR is what codifies the guard's semantics, so it's the right moment to decide.**

The two bypasses **stack over a service account's lifetime**, leaving the revocation guard permanently off for platform-API profiles:

- `_BUSINESS_APP_LABELS` (`platform_api/services.py:20-30`) omits **`prompt_profile_manager_v2`** while including its sibling `prompt_studio_core_v2`.
- `transfer_ownership` (`:110-119`) skips any model outside that set, so it never touches `ProfileManager`.
- `delete_api_user_for_key` (`:131-137`) then calls `api_user.delete()`, and `created_by` is `on_delete=models.SET_NULL` (`prompt_profile_manager_v2/models.py:109-116`) → `created_by = NULL`.

So while the key lives, `:234` passes; once the key is rotated or deleted, this `owner is None` branch passes. Both states, every requester — reachable through the ordinary key lifecycle, not a hand-edited DB.

`main` has the same `if profile_manager_owner is None: return`, so I'm not asking you to fix it here. But the new docstring at `:210-217` enumerates the bypasses and omits this one. Either document the None branch honestly, or file `_BUSINESS_APP_LABELS` as a follow-up — the label asymmetry reads like an oversight rather than a decision.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -183,105 +183,111 @@ def validate_adapter_status(
...
+        )
+
+    @staticmethod
+    def _user_has_adapter_access(user: Any, adapter: Any) -> bool:
+        return (
+            adapter.shared_to_org
+            or adapter.created_by == user
+            or adapter.shared_users.filter(pk=user.pk).exists()
+            or has_group_access(user, adapter)
+        )
```
ritwik-g: **P3 / follow-up — 3 of the 4 disjuncts are unpinned.**

Proven by sabotage: reducing this helper to `return adapter.created_by == user` — dropping `shared_to_org`, `shared_users` and `has_group_access` — leaves **all 15 tests passing**. The fixtures hardcode every disjunct but `created_by` to `False` (`test_validate_profile_manager_owner_access.py:51-52, 93-94`). Line coverage reports 100% and actively hides this, since Python branch coverage doesn't track short-circuit sub-conditions.

Follow-up rather than a gate: the extraction itself is faithful — I diffed it against `origin/main:207-238` disjunct-for-disjunct — and `main` had *zero* tests for this function, so this PR moves coverage 0 → substantial.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
ritwik-g: **P3 — design nit: this re-resolves a User the view already holds, via a non-unique CharField.**

`request.user` → `user_id` string → back to a User costs 3 queries: `get_user_by_user_id` (no `select_related`), the `member.user` lazy deref, then `is_user_organization_admin(member.user)` re-fetching the same row at `organization_member_service.py:32`. That's ×2 per index (default + summary profile).

Two consequences beyond the query count:
- it bypasses `permissions/permission.py:47-58` `_is_organization_admin(request)`, which already does exactly this from `request.user` **with a per-request cache**;
- it newly keys an authz decision on `User.user_id` — `models.CharField()` with no `unique=True` (`account_v2/models.py:70`) — where the pre-fix code used the PK.

Passing `request_user_id=request.user` (the object the views already have) would delete this helper and cut 3 queries to 0–1. The cost is cosmetic next to extraction + embedding + inference, so this is about the design, not the perf.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
athul-rs: Fixed in 153e4dcc, both halves: `request_user` is now keyword-only with **no default** on all four builders (dropped plumb = TypeError; `TestBuilderSignatures` pins the signature shape), and all three previously-untested builders got forwarding tests (parametrized sentinel-abort via the patched validator). The index builder's existing forwarding test now asserts an object sentinel, so the adjacent-param swap you describe can't pass either.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
athul-rs: Fixed in 153e4dcc — took your suggestion wholesale: views pass `request.user`, the chains carry `request_user: User | None`, and `_is_org_admin_by_user_id` is deleted. The validator calls `is_user_organization_admin(request_user)` directly (single lookup, service-account exclusion stays in its one home), and no authz decision keys on the non-unique `user_id` CharField anymore. I stopped short of the `_is_organization_admin(request)` per-request cache since the validator sits below the request layer; at 1 query per validation that seemed acceptable — shout if you disagree.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
athul-rs: Fixed in 153e4dcc — `TestCreatorAccessDisjuncts` pins the three previously-unpinned paths (org-share, user-share, group-share), each with only that disjunct truthy, so your sabotage (`return adapter.created_by == user`) now fails 3 tests. Nice catch on line coverage actively hiding this.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
athul-rs: Sign-off added to the PR body ("Can this PR break…" section) in exactly those terms: the SA-creator bypass is a deliberate authz decision, scoped as you framed it — platform-API path only, bounded by admin-gated key creation. Thanks for scoping it honestly rather than flagging the org-wide-migration version.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
athul-rs: Both options taken: the validator docstring now enumerates the `created_by is None` branch honestly (including the key-deletion → SET_NULL route), and the `_BUSINESS_APP_LABELS` omission of `prompt_profile_manager_v2` is filed as UN-3750 with your full chain (labels set → transfer_ownership skip → delete_api_user_for_key → SET_NULL) and a proposed fix + backfill question. Agreed the asymmetry with `prompt_studio_core_v2` reads like an oversight.

ritwik-g (approved):
Approving — the fix is correct, and I re-verified it still holds after the UN-2202 merge.

Verified at `4c8bbe13` by execution rather than reading. All four builders are genuinely keyword-only with no default (`inspect.signature` → `KEYWORD_ONLY, default=empty`), and every bypass is load-bearing under sabotage: deleting the admin-requester branch, the service-account branch, or the None-creator branch each fails exactly one targeted test. The rewired disjunct tests reproduce the original signature under the new model — reducing `_adapter_accessible_by` to owner-only fails exactly 3, and exactly the right 3. `_adapter_accessible_by` is byte-identical to main's (adopted, not reinvented), and no coverage was lost in the rewire: 15 test defs on both sides, the only delta a rename and a forced mock rewire. Backend unit tier green.

Worth calling out two judgement calls that went the right way: declining the Greptile suggestion (it would have dropped the service-account exclusion from `is_user_organization_admin` and duplicated admin-role resolution outside its one home), and catching during implementation that `user_id` is the file-path owner rather than plumbing it through blind — that would have no-oped on the exact reported bug.

## One thing to fix before merge — the description, not the code

The **"Can this PR break any existing features"** section still says:

> "All new parameters are trailing and optional (`None` default): callers that don't pass a requester (worker/legacy paths) get exactly the pre-fix validation behavior."

That was true at `fcf48ba5` and was falsified by `153e4dcc` — the very commit that hardened the builders. It wants **rewriting rather than deleting**, because the two-tier truth is the useful part:

- the four view-facing builders take `request_user` as **required keyword-only** — a dropped plumb is a `TypeError`;
- internal/worker paths remain `None`-defaulted and get pre-fix behavior.

As written it's the sentence a future contributor reads before re-adding `= None` and silently reverting UN-3739 — and the service-account sign-off currently sits underneath it, which undercuts the sign-off too.

Two smaller ones while you're in there:

- **3× `request_user_id`** in What / How / Notes on Testing — it's a `User` object now. That's the exact confusion your own commit removed from the code; leaving it in the body invites its return.
- **"49/49 backend tests pass (`prompt_studio/`, `adapter_processor_v2/`)"** — that scope collects **59** at HEAD. Also worth naming the three classes that *are* the review response (`TestCreatorAccessDisjuncts`, `TestBuilderSignatures`, `TestBuilderForwarding`), since they're the interesting part of the test plan.

## Best follow-up if you want one (not a gate)

The plumb **type** is unpinned. Changing any view to `request_user=request.user.user_id` passes the entire suite green — but a `str` is truthy and lacks `.is_authenticated`, so `is_user_organization_admin` short-circuits to `False`, the admin bypass silently dies, and UN-3739 is back with no failing test and no log line. It's the same shape as the hole you just closed, one layer up, and it's invited by `user_id=tool.created_by.user_id` sitting on the adjacent line. One thin view-level test closes it.

Related, same family: `request_user` is still `None`-defaulted on 8 functions including the guard itself — only the 4 builders are hardened. Latent today (every forward is correct), but it's the remediation surface if you ever want to make the contract uniform.

athul-rs:
Thanks for the approval and the triage framing — verified all three before acting:

**Dead code (mine): confirmed, filed as UN-3756.** Re-ran the caller sweep on both repos: cloud's `prompt_responder` hits are Simple Prompt Studio's own `SPSProjectHelper.prompt_responder` (no delegation), cloud imports only `get_select_fields`/`get_tool_from_tool_id`/`fetch_prompt_from_tool` from this helper, and the sdk/worker `index_document` hits are `vector_db.index_document`. So your 4-of-9 unreachable-call-sites count stands. Kept the deletion out of this PR since it's approved — UN-3756 scopes the 7 functions plus their now-pointless `request_user` plumbing.

**UN-1258 (mine): confirmed Jira-only.** It exists — "BUG_REG (0.60.1): shared LLM selected in profile manager → permission error", May 2024, Closed — but there's no corresponding commit in this repo's history, so it's cited in UN-3739 as same-failure-family context, not code precedent. I've annotated the ticket's Related section accordingly.

**Agentic-table (Hari's call): one data point to help the triage** — it's not just the view branch skipping `build_fetch_response_payload`; the cloud plugin's `build_agentic_table_payload` doesn't run `validate_adapter_status` or the owner-access check on its side either, so the path currently has zero adapter validation end-to-end. Leaving the ticket/ownership decision with Hari as you framed it.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -205,79 +205,98 @@ def validate_adapter_status(
...
+            " requester %s",
+            profile_manager.profile_name,
+            owner.user_id,
+            getattr(request_user, "user_id", None),
+        )
+
+        # Creator identity stays in server logs only — this error is shown
+        # to non-admin collaborators (admins bypass above), so no PII here.
+        denied_names = ", ".join(adapter.adapter_name for adapter in denied)
+        profile_ref = (
```
jaseemjaskp, line 289: [low] **The denial message tells users their own profile "was created by another user".**

`profile_ref` hardcodes *another user* into every branch, but the requester can be the creator.

**Trigger:** non-admin user B creates a project whose default profile uses an adapter that X shared to B. X unshares it. B runs their own project → requester B is not an admin, creator is B, `_adapter_accessible_by` is now False. B is told their own profile "was created by another user" and to "ask an org admin to re-share it with the profile's creator" — B *is* the creator.

This is worth flagging because it's a regression against this PR's own stated goal. The old `"You do not have access to {adapter}"` wording was **correct** for exactly this case, and share-revocation-on-own-profile is the most common revocation shape. The generic wording adopted to fix the PII leak (#299 thread) is what introduced the falsehood — so the two threads pull in opposite directions, and identity-branching resolves both.

```python
if request_user is not None and owner.pk == request_user.pk:
    profile_ref = (
        f"This project's LLM profile '{profile_manager.profile_name}' uses"
        f" adapters you no longer have access to"
    )
else:
    profile_ref = (...existing...)
```

The `no longer a member` branch below can't collide: B is querying as a member. Not blocking — the deny decision is right and remediation is still partially actionable.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -317,6 +336,7 @@ def _build_summarize_params(
         stem: str,
         extract_file_path: str,
         platform_api_key: str,
+        request_user: User | None = None,
```
jaseemjaskp: [low] **This is the one *live* hop where dropping the plumb silently reverts to pre-fix behavior.**

`TestBuilderSignatures` deliberately pins the four view-facing builders as keyword-only **with no default**, so "a dropped plumb must fail loudly (TypeError), never silently revert". `_build_summarize_params` runs the *same guard* against the summary profile (line ~363) but takes `request_user: User | None = None`.

Change the call site at line ~550 to drop `request_user=` and the summarize-profile check silently reverts — **no test fails** (mutation-verified: both this default and the inner validator call survive sabotage with all 25 tests green). The regression it admits is UN-3739 itself: an org admin indexing a **summarize-enabled** project with a distinct summarize profile gets denied again.

Your PR body scopes the fail-loud claim to the four builders, so the stated contract isn't broken — it's just narrower than the risk. Suggest `*, request_user: User | None` (no default) here and adding it to the `TestBuilderSignatures` parametrize list, which closes the gap structurally rather than with a test.

backend/prompt_studio/prompt_studio_core_v2/tests/test_build_index_payload.py
```diff
@@ -173,3 +192,24 @@ def test_check_extraction_status_raises_is_swallowed(self, caplog) -> None:
...
+
+    def test_owner_access_receives_requesting_user(self) -> None:
+        validate_owner_mock = MagicMock(return_value=None)
+        _dispatch_build(
+            check_return=True,
+            read_return="extracted text",
+            validate_owner_mock=validate_owner_mock,
+        )
+        _args, kwargs = validate_owner_mock.call_args
+        assert kwargs.get("request_user") is _REQUEST_USER
```
jaseemjaskp, line 215: [low] **The summarize-profile validator call has zero coverage — proven by mutation.**

`build_index_payload` forwards `request_user` into `_build_summarize_params`, which makes a **second** `validate_profile_manager_owner_access` call against `summary_profile`. Both hops can be reverted to `request_user=None` and **all 25 tests still pass**.

Two causes: `_build_summarize_params` is patched wholesale (lines 116-119), and `_make_tool()` defaults to `summarize_context=False`, so the branch never executes anyway.

Cheapest fix — a direct unit test on `_build_summarize_params` in the new suite asserting the validator receives `request_user=<sentinel>` when `summary_profile != default_profile`. Fuller version needs `_dispatch_build` to grow a `patch_summarize: bool = True` escape hatch so the real path runs:

```python
def test_summarize_profile_check_also_receives_requesting_user(self) -> None:
    validate_owner_mock = MagicMock(return_value=None)
    _dispatch_build(
        check_return=True, read_return="extracted text",
        tool=_make_tool(summarize_context=True),
        validate_owner_mock=validate_owner_mock,
    )
    assert validate_owner_mock.call_count == 2
    for call in validate_owner_mock.call_args_list:
        assert call.kwargs.get("request_user") is _REQUEST_USER
```

Credit where due: the rest of this suite is genuinely strong — 11 of 15 mutants died, including every bypass removal and the tuple-detail revert.

backend/prompt_studio/prompt_studio_core_v2/tests/test_validate_profile_manager_owner_access.py
```diff
@@ -0,0 +1,348 @@
...
+                _psh_mod, "has_group_access", MagicMock(return_value=group_access)
+            )
+        )
+        PromptStudioHelper.validate_profile_manager_owner_access(
+            profile, request_user=request_user
+        )
+
+
+class TestRequesterBypass:
+    """Cases 1 and 7 of the UN-3739 matrix: the reported bug."""
```
jaseemjaskp: [low] **Every `Case N` citation in this file disagrees with the module docstring's own contract, and three cite cases that don't exist.**

The docstring (lines 11-17) defines a 6-item contract in evaluation order. The per-test citations use a different, unpublished numbering:

| Line | Cites | Docstring actually says | Should be |
|---|---|---|---|
| 117 | "Cases 1 and 7" | no Case 7 exists | Cases 1 and 6 |
| 129 | "Case 2" (non-admin blocked) | Case 2 = `created_by is None` → pass | Case 6 |
| 148 | "Case 4" (service account) | Case 4 = creator is org admin | Case 3 |
| 156 | "Case 3" (admin creator) | Case 3 = service account | Case 4 |
| 164 | "Case 8" (None creator) | no Case 8 exists | Case 2 |
| 171 | "Cases 5/6" (creator has access) | Case 6 = the denial branch | Case 5 |
| 246 | "Case 7" (offboarded creator) | no Case 7 exists | Case 6 (sub-branch) |

Lines 148 and 156 are a straight swap of each other. Cases 7/8 point at "the UN-3739 matrix" (line 117) that isn't in the repo, so a future maintainer can't resolve them.

Suggest renumbering against the docstring's 1-6 and dropping the "UN-3739 matrix" phrasing — the docstring *is* the contract, so a second numbering has no reason to exist. If you want to cite the two message sub-branches individually, split item 6 into 6a/6b rather than inventing 7/8.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -205,79 +205,98 @@ def validate_adapter_status(
...
+            "Adapter access denied for profile '%s': creator %s lacks access,"
+            " requester %s",
+            profile_manager.profile_name,
+            owner.user_id,
+            getattr(request_user, "user_id", None),
+        )
+
+        # Creator identity stays in server logs only — this error is shown
+        # to non-admin collaborators (admins bypass above), so no PII here.
+        denied_names = ", ".join(adapter.adapter_name for adapter in denied)
```
jaseemjaskp: [nit] **Dropping the `set()` lost name dedupe — a denial can now read "these adapters: prod, prod".**

The old code accumulated `adapter_names` in a `set()`. The list comprehension is the right call (it fixed the set's nondeterministic ordering — don't go back), but it also dropped dedupe. Two adapters of different types can share a name.

```python
denied_names = ", ".join(dict.fromkeys(a.adapter_name for a in denied))
```

Order-preserving, keeps the ordering fix. Note `len(denied) > 1` at line 284 keys off adapter *count*, not name count, so a two-adapter/one-name denial reads "these adapters: prod" — minor, and arguably fine.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -205,79 +205,98 @@ def validate_adapter_status(
...
+        profiles whose creating user was deleted, which includes
+        platform-API users removed on key deletion since ``created_by``
+        is SET_NULL; a service-account creator (platform-API profiles
+        hold no adapter shares by design); and an org-admin creator.
 
         Args:
-            profile_manager (ProfileManager): The profile manager instance to
-              validate.
+            profile_manager: The profile whose adapters will be used.
+            request_user: The user triggering the action, if in a request
```
jaseemjaskp: [nit] **"Worker paths pass None and rely on creator checks" describes an architecture that doesn't exist.**

Nothing in `workers/` calls this validator, and nothing passes `request_user=None` except tests and the legacy `index_document` / `prompt_responder` chain — both of which have **zero callers** in the repo. Every live call arrives via the four builders, which all pass a requester.

This is the one place the PR asserts something false, and it's load-bearing: it's the stated justification for the `= None` defaults, so the next person threading a `None` through a live path will cite it.

Suggest the truth: *"Defaults to None for the legacy, currently-uncalled `index_document`/`prompt_responder` chain. All live callers (the four builders) pass a requester."*

Same root cause makes the "(admins bypass above)" rationale at line 270-271 slightly leaky — it holds only when `request_user is not None`. The no-PII conclusion is still correct and correctly conservative; only the justification is.

backend/prompt_studio/prompt_studio_core_v2/tests/test_validate_profile_manager_owner_access.py
```diff
@@ -0,0 +1,348 @@
...
+        # Creator has access to all but the LLM adapter — mirrors the
+        # reported single-adapter denial.
+        for attr in ("vector_store", "embedding_model", "x2text"):
+            getattr(profile, attr).owner_set.add(creator)
+
+        with pytest.raises(PSPermissionError) as exc_info:
+            _run_check(profile, creator_is_member=creator_is_member)
+        return exc_info.value
+
+    def test_message_blames_creator_without_pii(self) -> None:
```
jaseemjaskp, line 232: [nit] **The single-vs-multi adapter message branch is unpinned in both directions.**

`elif len(denied) > 1:` can be forced to `elif True` *or* `elif False` and all tests pass. `TestDenialMessage` only asserts adapter *names appear*, never the wording that distinguishes the branches — though the branches produce materially different copy:

- 1 denied → `...no longer has access to the adapter 'llm-1'. Ask an org admin to re-share it...`
- 4 denied → `...no longer has access to these adapters: llm-1, vdb-1, .... Ask an org admin to re-share them...`

Two asserts close it:

```python
# test_message_blames_creator_without_pii (single-denial fixture):
assert "the adapter 'llm-1'" in message
assert "these adapters" not in message

# test_message_lists_all_inaccessible_adapters (4 denied):
assert "these adapters:" in message
assert "the adapter '" not in message
```

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -205,79 +205,98 @@ def validate_adapter_status(
...
+        # Access resolution via the UN-2202 membership bridges — created_by
+        # is audit-only for adapters; owner/viewer roles carry access.
+        denied = [
+            adapter for adapter in adapters if not _adapter_accessible_by(adapter, owner)
+        ]
+        if not denied:
+            return
 
-            raise PermissionError(error_msg)
+        for adapter in denied:
```
jaseemjaskp: [nit] **Up to 5 `error` records per denial, and neither log tells the whole story.**

The two blocks aren't redundant — they're *complementary*, which is the actual problem: the loop has `adapter.id` but no profile/requester; the summary has profile/requester but no adapter identity. An on-call reader has to stitch them across records.

```python
logger.error(
    "Adapter access denied for profile '%s': creator %s lacks access to"
    " adapters %s, requester %s",
    profile_manager.profile_name,
    owner.user_id,
    [str(adapter.id) for adapter in denied],
    getattr(request_user, "user_id", None),
)
```

One record, strictly more information than either had. This makes `ERROR_MSG` (line 91) dead — the only other copy is inlined at `tool_instance_helper.py:519`, so there's no shared-constant coupling to break.

**Weigh first:** if any log-based alerting greps `"doesn't have access to adapter"`, this moves the needle. Only you know whether that exists.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -205,79 +205,98 @@ def validate_adapter_status(
...
+        if OrganizationMemberService.is_user_organization_admin(owner):
+            return
+
+        adapters = [
+            profile_manager.llm,
+            profile_manager.vector_store,
+            profile_manager.embedding_model,
+            profile_manager.x2text,
+        ]
+        # Access resolution via the UN-2202 membership bridges — created_by
```
jaseemjaskp: [nit] **This comment enumerates 2 of the 4 access disjuncts, and duplicates a docstring 150 lines up.**

"owner/viewer roles carry access" omits `shared_to_org` and `has_group_access` — both checked by `_adapter_accessible_by`, and both pinned by this PR's own `test_org_shared_adapters_pass` / `test_group_shared_adapters_pass` as independently sufficient. "Membership bridges" also doesn't describe org-share or group access.

`_adapter_accessible_by` already carries an accurate, complete docstring. Two descriptions of one predicate is how they drift apart. Suggest deleting, or reducing to a pointer:

```python
# Access = org share, owner/viewer role, or group (UN-2202); see _adapter_accessible_by.
```

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -205,79 +205,98 @@ def validate_adapter_status(
...
 
         Raises:
-            PermissionError: If the owner does not have permission to perform
-              the action.
+            PermissionError: If the profile creator no longer has access to
+              one or more adapters and no bypass applies.
         """
-        profile_manager_owner = profile_manager.created_by
-        if profile_manager_owner is None:
+        if (
```
jaseemjaskp: [nit] **`request_user is not None` is dead weight and implies the callee can't handle None.**

`is_user_organization_admin` already opens with `if not user or not getattr(user, "is_authenticated", False): return False` (`organization_member_service.py:27`) — it handles `None`, `AnonymousUser`, and service accounts.

```python
if OrganizationMemberService.is_user_organization_admin(request_user):
    return
```

Minor, but it removes the false implication that invites the same defensive wrapper at the next call site.

jaseemjaskp (approved):
## Review summary — approving with nits

Reviewed with six agents (code review, silent-failure, type design, test coverage, comments, simplification). **No blocking findings**; the authz core is sound and I could not break it:

- **The admin/service-account bypasses grant nothing new.** Both match the platform's existing adapter model — `AdapterInstanceModelManager.for_user` already returns the unfiltered queryset for org admins and grants service accounts every non-frictionless adapter. Not a privilege escalation.
- **Every failure mode fails closed.** `is_user_organization_admin` swallowing exceptions → `False` only ever *removes* a bypass. Under a real DB outage `_adapter_accessible_by` issues uncaught queries, so you get a loud 500, not a bogus 403. `MultipleObjectsReturned` is ruled out by the `(organization, user)` unique constraint.
- **No breaking change.** All four builders are called only from `views.py`, and all four pass `request_user`.
- **Real bug fixed in passing:** the old `error_msg = (f"...",)` trailing comma made the DRF detail a tuple; now a plain string, and pinned.
- **The tests are strong.** Mutation testing killed 11/15 mutants, including every bypass removal. `_adapter()` setting `shared_to_org = False` and `_user()` setting `is_service_account` explicitly are load-bearing — a bare `MagicMock` is auto-truthy and would have made every denial test pass vacuously. Correctly handled.

The findings worth your time, in order:

1. **`profile_ref` tells users their own profile "was created by another user"** (helper.py:273) — reachable via the most common revocation shape, and a regression against this PR's own goal.
2. **`_build_summarize_params` is the one live hop that silently reverts** (helper.py:339) — and mutation testing shows nothing catches it.
3. **The summarize-profile validator call is entirely uncovered** (test_build_index_payload.py) — same bug class as UN-3739, for summarize-enabled projects.
4. **The new suite's `Case N` citations are systematically wrong** (test_validate_...py:117) — two are a straight swap, three cite cases that don't exist.

The rest are genuine nits. Deliberately **not** raising: the `getattr(owner, "is_service_account", False)` idiom (matches the house convention at 8+ sites incl. `organization_member_service.py:29`; `getattr(request_user, "user_id", None)` is outright correct since `request_user` is legitimately None), and any value-object/NewType wrapper for the two identities — `str` vs `User` plus keyword-only already covers the residual conflation risk. The explicit-parameter plumb is the right call over a `StateStore` thread-local; the wider diff buys you an identity flow that can't be silently lost again.

Follow-up ticket material, not this PR: `PromptStudioHelper.index_document` / `prompt_responder` and their chain have zero callers — the `request_user` additions there are inert. Deleting the chain would also let you drop the `= None` defaults. Renaming `user_id` → `file_owner_user_id` on the four builders would make the invariant self-evident rather than test-enforced.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
athul-rs: Fixed in ba313f4f — identity-branched exactly as you suggested: when the requester is the creator they're addressed directly ('uses the adapter …, which you no longer have access to'), still PII-free; pinned by test_message_addresses_creator_requester_directly (asserts 'another user' absent). Good catch — this was a regression against the PR's own goal.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
athul-rs: Fixed in ba313f4f — `_build_summarize_params` is now `*, request_user: User | None` with no default and is in the TestBuilderSignatures parametrize list. Agreed it was the one live hop outside the stated contract.

backend/prompt_studio/prompt_studio_core_v2/tests/test_build_index_payload.py
athul-rs: Fixed in ba313f4f via your cheapest-fix variant: TestSummarizeHopForwarding drives the real `_build_summarize_params` (summarize_context=True, distinct summary profile) and asserts the validator receives the requester sentinel. Combined with the no-default signature, both of your surviving mutants are dead.

backend/prompt_studio/prompt_studio_core_v2/tests/test_validate_profile_manager_owner_access.py
athul-rs: Fixed in ba313f4f — all citations renumbered against the module docstring's 1–6 contract (your table was exact, including the 3↔4 swap), 'UN-3739 matrix' phrasing dropped, and the former-member test cites case 6's sub-branch instead of an invented 7.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
athul-rs: Fixed in ba313f4f with your `dict.fromkeys` one-liner (ordering kept); pinned by test_message_dedupes_adapter_names. The two-adapter/one-name plural-copy edge you flagged is left as-is per your own 'arguably fine'.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
athul-rs: Fixed in ba313f4f with your suggested wording — the docstring now says None exists only for the legacy, currently-uncalled index_document/prompt_responder chain (UN-3756 tracks its deletion) and that all live callers pass a requester. The adjacent no-PII comment was reworded too, so the leaky '(admins bypass above)' rationale is gone.

backend/prompt_studio/prompt_studio_core_v2/tests/test_validate_profile_manager_owner_access.py
athul-rs: Fixed in ba313f4f — both directions pinned with your exact asserts: singular fixture requires "the adapter 'llm-1'" and forbids 'these adapters'; the 4-denied fixture requires 'these adapters:' and forbids "the adapter '".

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
athul-rs: Fixed in ba313f4f — single error record with profile, creator, denied adapter ids, and requester; ERROR_MSG deleted (verified: the only other 'doesn't have access' string is inlined in tool_instance_helper, untouched). On your weigh-first caveat: no automated alerting greps that line to my knowledge — the only consumers were ad-hoc support/QA grep instructions from this ticket, which I'm updating to the new format.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
athul-rs: Fixed in ba313f4f — reduced to your pointer one-liner; _adapter_accessible_by's docstring is the single source of truth.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
athul-rs: Fixed in ba313f4f — guard dropped; verified is_user_organization_admin handles None/anonymous at organization_member_service.py:27. The one remaining None check (owner.pk == request_user.pk in the new identity branch) stays because None has no pk.

PR #2170, opened 2026-07-13
UN-3636 [DEV] Hermetic LLM mock hook for execute-path critical tests

chandrasekharan-zipstack:
Stacked on #2164. Adds the deterministic-LLM capability the execute-path critical-path gaps need, without dragging real provider secrets into CI.

## What
- **sdk1 hook** (`llm.py`): one module-level `_inject_mock_response` + a one-line call at all four `litellm.completion`/`acompletion` sites. When `UNSTRACT_LLM_MOCK_RESPONSE` is set, litellm returns that string as the completion with fixed usage (10/20/30), tunable via `DEFAULT_MOCK_RESPONSE_*`; error sentinels like `litellm.RateLimitError` force error paths. Unset in production → no-op.
- **compose overlay**: pass the env through `worker-executor-v2` / `worker-file-processing-v2` (the processes that run the LLM). Empty unless the rig/CI sets it.
- **unit tests**: pin both our injector and the litellm mock contract (string + deterministic usage + error sentinel), so a litellm bump can't silently break hermetic execute-path coverage.

## Why this shape
- Single-place change, no cross-codebase adapter churn (unlike the retired NoOp adapter).
- Hermetic + secret-free by construction — a cache-miss can't fall through to a real paid call (contrast: a record/replay proxy needs real keys to warm + as miss fallback).
- Execute path tolerates a plain string (TEXT prompts stored verbatim; JSON prompts use tolerant repair) and never calls `test_connection()`, so a fixed mock runs every execute path to completion.

## Not in this PR
The actual execute-path e2e tests (workflow-create-execute, usage-token-tracking, …) that consume this hook — those need the full compose stack booted to author the API flow correctly and will land next. Critical-path mappings are intentionally left unflipped until a real passing test exists (no false greens).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2188, opened 2026-07-20
UN-3636 [MISC] Merge overlay manifests via UNSTRACT_RIG_EXTRA_MANIFESTS

chandrasekharan-zipstack:
## What

Three rig changes, bundled because they all land on the same test-infrastructure surface:

1. **Overlay manifests.** `load_groups()` merges extra group manifests listed in the `UNSTRACT_RIG_EXTRA_MANIFESTS` env var (os.pathsep-separated, REPO_ROOT-relative) onto the base `tests/groups.yaml` **before** validation — so cross-manifest `depends_on` and the platform-gate invariant are checked over the union. Group-name collisions are an error, not a silent override.
2. **Cross-tier deps no longer re-run.** `--tier X` now bounds `depends_on` expansion to tier X.
3. **The `ENVIRONMENT` gate on the LLM mock is gone** (merged in via #2191), along with the now-unread `ENVIRONMENT` variable itself.

### 1. Overlay manifests

Lets a downstream repo contribute its own test groups without editing the OSS manifest. First consumer is the **cloud build**: it copies a `tests/groups.cloud.yaml` into the merged tree (via `copy_cloud_deps.py`) and points the env var at it, so cloud-only plugin test groups (workers, etc.) join the rig without forking `groups.yaml`. No cloud specifics leak into OSS.

### 2. Cross-tier deps

`integration-workflow-execution` and `e2e-smoke` both declare `depends_on: [unit-sdk1, unit-workers]`. Because `--tier` expanded deps transitively with no tier bound, those two unit groups ran **three times per CI run** — once in the unit leg, again in integration, again in e2e. Measured on run `29891886572` (main):

| group | unit leg | integration leg | e2e leg |
|---|---:|---:|---:|
| `unit-sdk1` | 21s | 27s | 27s |
| `unit-workers` | 37s | 51s | 51s |

≈136s of duplicate execution per run. On the cloud CI runner (`custom-k8s-runner`, 2 xdist workers) the same duplication costs ~350s.

### 3. Dropping the `ENVIRONMENT` gate (#2191)

Follow-up to #2170, removing a guard that PR added. The gate required `ENVIRONMENT ∈ {test, development}` on top of `UNSTRACT_LLM_MOCK_RESPONSE` before the SDK would mock a completion. It was added for [this review comment](https://github.com/Zipstack/unstract/pull/2170#discussion_r3614772911) — the hazard being a worker env block copied out of the test overlay into a real deployment, where litellm's synthetic 10/20/30 token usage would flow into the usage tables indistinguishable from real spend.

It did not defend that case:

- The gate was written **into that same overlay block**, next to the mock var, so a copy carries the gate along with it.
- `docker/docker-compose.yaml` set `ENVIRONMENT=development` on both workers that run the injection, so any deployment derived from base compose — which is how Unstract is self-hosted — satisfied the gate anyway.

That left one case it genuinely caught: the mock var set alone somewhere that sets no `ENVIRONMENT` at all. That held by accident (nothing sets the variable in k8s) rather than by design, and it cost a dependency on a variable **nothing else in the codebase reads** — so the variable was dropped from compose too, rather than left looking load-bearing.

What still guards the hatch is unchanged: it is off unless someone sets `UNSTRACT_LLM_MOCK_RESPONSE`, and it logs a warning once per process while active. The defence worth having is making mocked spend *distinguishable downstream* — an `is_mock` marker on the usage record, which config checks can't be copy-pasted around. Not in this PR.

## Changes

**Overlays**
- `tests/rig/groups.py`: `_extra_manifest_paths()`, `_load_manifest_dict()`, `_merge_manifest()`, wired into `load_groups()`.
- `tests/rig/tests/test_groups.py`: +6 tests (merge with cross-manifest `depends_on`, name collision, explicit-manifest isolation, overlay defaults, malformed overlay, missing overlay path).

**Cross-tier deps**
- `tests/rig/selection.py`: with `--tier`, drop dep-expanded groups outside that tier.
- `tests/rig/tests/test_selection.py`: +3 tests (cross-tier dep not pulled in, explicit request survives the filter, expansion unchanged without a tier filter).

**LLM mock gate (#2191)**
- `unstract/sdk1/src/unstract/sdk1/llm.py`: drop `_ENVIRONMENT_ENV`, `_MOCK_ALLOWED_ENVIRONMENTS`, `_warn_mock_refused`, and the gate.
- `unstract/sdk1/tests/test_mock_response.py`: drop the 4 gate tests and the autouse env fixture (9 tests remain, all passing).
- `tests/compose/docker-compose.test.yaml`: drop the two `ENVIRONMENT=test` lines added for the gate.
- `docker/docker-compose.yaml`: drop the unread `ENVIRONMENT` declarations.
- `tests/README.md`: record why it was removed, so it doesn't get re-added.

**No-org short-circuit**
- `backend/utils/user_context.py`: `get_organization()` returns `None` when no org is in context (import time, management commands with no request) instead of querying, so evaluating it on a DB-less/unmigrated setup can't fail — which is what the rig's backend groups do at collection time.
- `backend/utils/tests/test_user_context.py`: cover both branches, including the truthy one so inverting the guard can't pass unnoticed.

## Safety

- Overlays are **no-op when the env var is unset** → zero effect on existing OSS runs, and they apply only to the default manifest, so an ad-hoc `load_groups(path)` stays isolated.
- Intra-tier deps still expand and topo-order (`e2e-smoke` → `e2e-login`); explicitly named groups are never dropped by the tier filter.
- Gating is unaffected: `blocked_by` intersects with groups that failed *in the same run*, so a dep that wasn't run simply doesn't gate — its own tier leg gates it.
- The LLM mock stays opt-in behind `UNSTRACT_LLM_MOCK_RESPONSE` and warns once per process while active.
- `tests.rig validate` OK — 18 groups, 16 critical paths. Full rig suite: 80 passed, 1 skipped.


tests/rig/groups.py
```diff
@@ -135,6 +146,40 @@ def load_groups(path: Path | None = None) -> GroupManifest:
...
+def _merge_manifest(
+    groups: dict[str, GroupDefinition], manifest_path: Path, base_defaults: dict[str, Any]
+) -> None:
+    """Merge an overlay manifest into ``groups`` in place. Overlay groups inherit
+    the base ``defaults`` unless they declare their own; a name collision is an
+    error rather than a silent override."""
+    raw = yaml.safe_load(manifest_path.read_text())
+    if not isinstance(raw, dict) or "groups" not in raw:
+        raise ValueError(f"{manifest_path}: expected top-level `groups:` mapping")
+    defaults = {**base_defaults, **(raw.get("defaults") or {})}
```
jaseemjaskp, line 190: [nit] An overlay's `defaults.platform_gate_group` is silently ignored. `_merge_manifest` builds a merged `defaults` here, but it only feeds `_build_group` — the gate name in `load_groups` (`smoke_gate = defaults.get("platform_gate_group", "e2e-smoke")`) is resolved from the **base** manifest's defaults only. So an overlay that both introduces `requires_platform` groups and renames the gate under its own defaults will fail validation with `platform gate 'e2e-smoke' is not defined`, even though it's internally consistent. Gate-renaming is an advertised feature, so this is an inconsistency between two supported behaviors. Either fold overlay defaults into the gate lookup, or add a one-line comment stating overlays cannot rename the gate.

tests/rig/groups.py
```diff
@@ -135,6 +146,40 @@ def load_groups(path: Path | None = None) -> GroupManifest:
...
+    return paths
+
+
+def _merge_manifest(
+    groups: dict[str, GroupDefinition], manifest_path: Path, base_defaults: dict[str, Any]
+) -> None:
+    """Merge an overlay manifest into ``groups`` in place. Overlay groups inherit
+    the base ``defaults`` unless they declare their own; a name collision is an
+    error rather than a silent override."""
+    raw = yaml.safe_load(manifest_path.read_text())
```
jaseemjaskp, line 175: [low] Two small robustness/clarity points on this parse block:

1. **Opaque error for a bad overlay path.** `read_text()` has no existence guard, so a typo in `UNSTRACT_RIG_EXTRA_MANIFESTS` surfaces a bare `FileNotFoundError` (and a directory → `IsADirectoryError`, a top-level YAML list → `AttributeError` on `raw["groups"]`) that never names the env var — unlike every other manifest error here, which uses `ValueError(f"{manifest_path}: ...")`. Consider guarding with a clear `ValueError` naming `UNSTRACT_RIG_EXTRA_MANIFESTS`.
2. **Duplication.** Lines 171-173 (`yaml.safe_load` + the `isinstance/`"groups" not in raw`` check + message) are byte-identical to `load_groups` (127-129). Extracting a `_load_manifest_dict(path) -> dict` helper single-sources both the parse and the error message; each caller keeps its own (differing) `defaults` logic.

backend/utils/tests/test_user_context.py
```diff
@@ -0,0 +1,30 @@
...
+class TestGetOrganizationNoContext:
+    def test_returns_none_without_hitting_db(self):
+        with (
+            patch("utils.user_context.StateStore.get", return_value=None),
+            patch("utils.user_context.Organization.objects.get") as mock_get,
+        ):
+            assert UserContext.get_organization() is None
+            mock_get.assert_not_called()
+
+    def test_empty_identifier_short_circuits(self):
```
jaseemjaskp, line 23: [nit] Both tests assert only the falsy branch of `if not organization_id`. Nothing pins the complement — that a truthy id falls through to `Organization.objects.get`. A mutation of `if not organization_id:` → `if organization_id:` passes both existing tests while breaking every real request. Cheap to close with the same mock style: set `StateStore.get` to return `"org-123"`, mock `Organization.objects.get` to return a sentinel, and assert the sentinel is returned and `get` was called with `organization_id="org-123"`.

tests/rig/tests/test_groups.py
```diff
@@ -232,3 +232,58 @@ def test_group_env_is_frozen() -> None:
...
+    base = _base_manifest(tmp_path)
+    overlay = tmp_path / "groups.cloud.yaml"
+    overlay.write_text(
+        """
+        version: 1
+        groups:
+          unit-cloud:
+            tier: unit
+            paths: [y]
+            depends_on: [unit-base]
```
jaseemjaskp, line 277: [nit] The two new overlay tests cover merge + name-collision well. Two uncovered overlay-specific branches worth pinning cheaply (each mirrors `test_extra_manifest_name_collision_rejected`):
- **Overlay with its own `defaults`** — `_merge_manifest` does `{**base_defaults, **overlay_defaults}` (groups.py:174); if it broke, an overlay's `parallel`/`timeout_seconds`/`runner` defaults would be silently dropped. Assert a merged group reflects an overlay default.
- **Malformed overlay** (non-dict / missing `groups:`) — the `ValueError` at groups.py:172-173 is a distinct branch from the same check in `load_groups` and is currently unexercised.

jaseemjaskp (approved):

Clean, well-tested change. The `user_context` short-circuit is behavior-preserving (all 42 callers already treat `None` as "no org") and adds no new silent failures; the overlay-manifest merge fails loud on every error path and rejects name collisions rather than silently overriding. No new types introduced. **Approving** — all findings below are non-blocking nits.

One pre-existing item worth a follow-up (not from this diff): `get_organization` catches a broad `django.db.utils.ProgrammingError` and returns `None` with no logging, which also swallows genuine runtime schema/permission errors under runserver — consider narrowing or logging.

Note: I did **not** repost @greptile-apps' P1 at `groups.py:139` (overlay loop applies to any custom `load_groups(path)` caller) — it stands and is worth resolving; whatever fix lands there will need the two new rig tests updated, since they rely on overlays applying with an explicit `path`.

tests/rig/groups.py
chandrasekharan-zipstack: Folded overlay defaults into the gate lookup rather than documenting the limitation.

`_merge_manifest` now returns the merged `defaults` and `load_groups` reassigns from it, so `smoke_gate = defaults.get("platform_gate_group", ...)` sees an overlay's rename. With multiple overlays each one inherits the previous merge, so the last writer wins — consistent with how group specs already merge.

tests/rig/groups.py
chandrasekharan-zipstack: Both done.

1. `_extra_manifest_paths()` now validates each resolved path and raises `ValueError(f"{EXTRA_MANIFESTS_ENV}: {entry!r} is not a file (resolved to {p})")`, so the message names the env var *and* shows both the raw entry and where it resolved to. Covers the typo, directory, and missing-file cases before `read_text()` is ever reached. Verified:

```
$ UNSTRACT_RIG_EXTRA_MANIFESTS=tests/nope.yaml python -m tests.rig validate
ERROR: UNSTRACT_RIG_EXTRA_MANIFESTS: 'tests/nope.yaml' is not a file (resolved to /.../tests/nope.yaml)
```

2. Extracted `_load_manifest_dict(path) -> dict` — single-sources the `yaml.safe_load` + shape check + error message for both `load_groups` and `_merge_manifest`. The top-level-list case you mentioned now hits the same `expected top-level `groups:` mapping` ValueError instead of an `AttributeError`. Each caller keeps its own `defaults` logic.

backend/utils/tests/test_user_context.py
chandrasekharan-zipstack: Added `test_identifier_present_looks_up_organization` with exactly that shape — `StateStore.get` returns `"org-123"`, `Organization.objects.get` is mocked to a sentinel, and it asserts both the sentinel is returned and `get` was called with `organization_id="org-123"`.

Confirmed it closes the mutation: inverting the guard to `if organization_id:` now fails this test. All 3 pass via the rig runner.

tests/rig/tests/test_groups.py
chandrasekharan-zipstack: Both added, plus two more.

- `test_extra_manifest_defaults_applied` — overlay declares `parallel: false` / `timeout_seconds: 42`, asserts they reach the merged group (pins the `{**base, **overlay}` spread).
- `test_extra_manifest_malformed_rejected` — overlay is a top-level list, expects the `expected top-level `groups:` mapping` ValueError.
- `test_extra_manifest_missing_path_names_env_var` — covers the new env-var-naming guard.
- `test_explicit_manifest_ignores_overlays` — pins the isolation fix from greptile's thread above.

Worth flagging: applying that isolation fix broke the two original overlay tests, since they passed an explicit fixture path *and* expected the overlay to merge. Reworked them to monkeypatch `DEFAULT_MANIFEST` and call `load_groups()` with no argument, so they now exercise the real production path instead of one that no longer takes overlays.

tests/rig/groups.py
```diff
@@ -118,15 +123,20 @@ def expand(self, selected: list[str]) -> list[str]:
...
     groups: dict[str, GroupDefinition] = {}
     for name, spec in (raw["groups"] or {}).items():
         groups[name] = _build_group(name, spec, defaults)
 
+    # Merge before validation so cross-manifest `depends_on` and the platform
+    # gate are checked over the union. An explicit `path` stays isolated from
+    # the ambient env var.
+    if manifest_path == DEFAULT_MANIFEST:
+        for extra in _extra_manifest_paths():
+            defaults = _merge_manifest(groups, extra, defaults)
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Valid — taken, in 8d1442d0.

Confirmed the gap is real and not just theoretical about the docstring: `test_explicit_manifest_ignores_overlays` only passed because its fixture manifest happens not to be the default one (`# base is deliberately not the default manifest here`). Pass the default path explicitly and overlays merged.

Two reasons `path is None` is the better key, beyond matching what the docstring already claimed:

1. `Path` equality is spelling-sensitive. `tests/groups.yaml` and `/abs/…/tests/groups.yaml` are the same file but compare unequal, so the old check gave two different behaviours for one manifest depending on how the caller wrote it. `path is None` has no such edge.
2. No caller in the repo passes an explicit path — `cli.py` always calls `load_groups()` — so this is behaviour-neutral for every real call site.

Added `test_explicit_default_path_ignores_overlays` to pin it; verified it fails against the old condition (`assert 'unit-cloud' not in ['unit-base', 'unit-cloud']`) and passes with the new one. Also updated the `_overlay_env` helper docstring, which said overlays apply "to the default manifest" — now "when `load_groups` is called with no path".

`tests/rig` unit tier: 81 passed, 1 skipped. Ruff clean (fixed a `D209` on `_merge_manifest`'s docstring while in there).

PR #2189, opened 2026-07-20
UN-3575 [FEAT] Lookup enrichment support for agentic-table prompts (OSS side)

pk-zipstack:
## What

- Thread the per-prompt `lookup_config` into the agentic-table executor's `agentic_params` in `structure_tool_task`, so post-extraction lookup enrichment runs for agentic-table prompts on the **deployment path**.
- Add a generic, length-preserving bounded-concurrency map at `workers/shared/parallel_map.py`, used by the agentic-table lookup per-row enrichment (and reusable by other LLM call sites).

## Why

Ref [UN-3575](https://zipstack.atlassian.net/browse/UN-3575). The cloud repo enables lookups on agentic-table prompts; the enrichment runs inside the dedicated `agentic_table` executor. For that to fire on **deployed** pipelines (not just the IDE), the OSS `structure_tool_task.py` — which dispatches agentic-table prompts to the executor — must forward the `lookup_config` that is attached to the prompt output upstream at export.

The per-row enrichment mode needs bounded concurrency (N rows → N LLM calls, capped at K workers). `parallel_map` is generic OSS-side infrastructure (like the existing `run_lookup_enrichment` hook), so it lives in the OSS `workers/shared` and is imported from there by the cloud agentic executor at runtime.

## How

- **`workers/file_processing/structure_tool_task.py`**: after building the base `agentic_params`, forward `at_output.get("lookup_config")`. It is `None` when no lookup is assigned to the prompt, so the executor runs enrichment only when set.

  ```python
  "lookup_config": at_output.get("lookup_config"),
  ```

- **`workers/shared/parallel_map.py`** (new): `parallel_map(items, worker, *, max_workers, on_error=None, label="")` — a `ThreadPoolExecutor` map that returns results **in input order and always the same length as the input** (failed items hold the `on_error` fallback, never dropped), so callers realign results to inputs by index. `max_workers=1` runs effectively sequentially, so one knob covers both sequential and bounded-parallel strategies. Retries/rate-limits are intentionally left to the SDK LLM client (which already backs off on 429/5xx and honours `Retry-After`), so concurrent callers self-throttle. No early-abort (documented).

No. Both changes are additive:
- `at_output.get("lookup_config")` returns `None` for all current deployments (no lookup on agentic-table), so the executor's enrichment path is skipped and `agentic_params` is functionally unchanged.
- `workers/shared/parallel_map.py` is a brand-new module with no existing importers in this repo; it introduces no behavioural change on its own.

## Database Migrations

- None (migrations for the `per_row` field live in the cloud repo).

## Env Config

- New optional env var **`LOOKUP_PER_ROW_MAX_WORKERS`** (default `8`) is read by the cloud agentic-table worker for per-row concurrency; no OSS-side config change required.

## Relevant Docs

- [UN-3575](https://zipstack.atlassian.net/browse/UN-3575)

## Related Issues or PRs

- Cloud companion PR: **Zipstack/unstract-cloud#1662** — UN-3575 [FEAT] Lookup enrichment for agentic-table prompts (whole-output + per-row). **Both must be merged/deployed together** — the cloud agentic executor consumes the `lookup_config` param forwarded here and imports `parallel_map` from this repo's `workers/shared`.

## Dependencies Versions

- No new package dependencies.

## Notes on Testing

- Integration-tested via end-to-end pipeline runs (this function is not unit-tested).
- Validated on a full local stack: a real agentic-table extraction with a per-row lookup logged `parallel_map: N items across up to 8 workers` and `Per-row lookup enrichment complete: N/N rows enriched, 0 failed`.
- Backward-compat verified: agentic-table prompts without a lookup dispatch with `lookup_config=None` and behave exactly as before.

## Screenshots

N/A — backend/worker-only change.

## Checklist

[UN-3575]: https://zipstack.atlassian.net/browse/UN-3575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[UN-3575]: https://zipstack.atlassian.net/browse/UN-3575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


workers/shared/parallel_map.py
```diff
@@ -0,0 +1,94 @@
...
+            "parallel_map: %d items across up to %d workers%s",
+            n,
+            effective_workers,
+            suffix,
+        )
+
+    with ThreadPoolExecutor(max_workers=effective_workers) as executor:
+        future_to_idx = {
+            executor.submit(worker, item): idx for idx, item in enumerate(items)
+        }
```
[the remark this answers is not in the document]
praveen-formido: Good catch on the queued futures. Keeping this as-is: the row payloads are already fully materialized in memory — the extracted `tables` array lives in RAM before enrichment starts — so submitting them to the pool doesn't duplicate the data; it only adds lightweight `Future`/work-item wrappers, which are negligible next to the table itself. `max_workers` bounds the thing that actually matters here (concurrent LLM calls vs. provider rate limits). If a future use case needs to cap submission for very large tables, chunked submission can be added behind the same interface without touching callers. Resolving as a nice-to-have.

workers/shared/parallel_map.py
```diff
@@ -0,0 +1,94 @@
...
+
+``max_workers=1`` runs effectively sequentially, so a single knob covers both the
+sequential and bounded-parallel strategies with no code change.
+
+Rate limiting / retries are intentionally NOT handled here — the SDK LLM client
+already retries transient errors (429/500/503/... and provider "overloaded")
+with backoff and honors ``Retry-After``, so ``max_workers`` concurrent callers
+self-throttle without any extra coordination in this utility.
+
+NOTE: there is no early-abort. All submitted work runs to completion even if the
```
chandrasekharan-zipstack, line 16: @muhammad-ali-e do you have any thoughts on this behaviour?

workers/file_processing/structure_tool_task.py
```diff
@@ -474,6 +474,10 @@ def _execute_structure_tool_impl(params: dict) -> dict:
             "execution_id": execution_id,
             "PLATFORM_SERVICE_API_KEY": platform_service_api_key,
             "group_key": at_settings.get("group_key", ""),
+            # Lookup enrichment config, attached to the output upstream when a
+            # lookup is assigned to this agentic-table prompt. None otherwise —
+            # the agentic_table executor runs enrichment only when set.
```
chandrasekharan-zipstack:
```suggestion
```
NIT: Consider removing such comments since it doesn't add much value

workers/shared/parallel_map.py
```diff
@@ -0,0 +1,94 @@
+"""Generic bounded-concurrency map for LLM (and other I/O-bound) call sites.
+
+Runs ``worker(item)`` across a thread pool capped at ``max_workers`` and returns
+results in input order. Unlike the agentic extractor's older ``parallel_page_map``
```
chandrasekharan-zipstack: ⚪ `parallel_page_map` exists only in the cloud repo (`agentic_table/src/utils/parallel_map.py`), so an OSS reader can't resolve the symbol and neither repo's CI catches it when that helper is renamed or deleted. "older" is also not quite right — both coexist. Suggest stating the property instead: "length-preserving: failed items keep a slot, so callers can realign by index."

workers/shared/parallel_map.py
```diff
@@ -0,0 +1,94 @@
...
+(which drops ``None`` results), this is **length-preserving**: the output list is
+always the same length as the input, so callers can realign results to inputs by
+index — essential for per-row enrichment where row *i* in must map to row *i* out.
+
+``max_workers=1`` runs effectively sequentially, so a single knob covers both the
+sequential and bounded-parallel strategies with no code change.
+
+Rate limiting / retries are intentionally NOT handled here — the SDK LLM client
```
chandrasekharan-zipstack: 🟠 **This claim doesn't hold, and it's the entire stated reason the utility ships without rate limiting.**

Checked `unstract/sdk1/src/unstract/sdk1/utils/retry_utils.py`:
- `RETRYABLE_STATUS_CODES = frozenset({408, 429, 500, 502, 503, 504})` — 429/500/503 are right, but **provider "overloaded" is not covered**: Anthropic's `overloaded_error` is HTTP 529, which isn't in that set, and `grep -i overloaded` over the SDK returns zero hits.
- Retries are **config-gated, not unconditional**: `pop_litellm_retry_kwargs` (`:176`) is `kwargs.pop("max_retries", None) or 0`, and `call_with_retry` with `max_retries=0` does exactly one attempt. Adapter schemas default to 3–5 so UI-created instances retry, but a config without the key gets zero.
- "so `max_workers` concurrent callers self-throttle" is a non-sequitur regardless — retry-with-backoff is failure recovery, not rate limiting. Concurrency stays at `max_workers` either way.

At 8 concurrent × 500 rows against a provider that 529s, the assumed safety net isn't there. Suggest cutting the paragraph to one line: *"Retries/rate limiting are the caller's concern; this utility only bounds concurrency."*

workers/shared/parallel_map.py
chandrasekharan-zipstack, line 16: ✅ This NOTE is a genuinely good comment — a non-obvious behavioural trap, stated generically, actionable, and it won't rot.

workers/shared/parallel_map.py
```diff
@@ -0,0 +1,94 @@
...
+) -> list[R]:
+    """Apply ``worker`` to each item with bounded concurrency, order preserved.
+
+    Args:
+        items: Inputs to process. An empty list returns ``[]``.
+        worker: Called as ``worker(item)`` for each item.
+        max_workers: Thread-pool cap. ``<= 1`` runs effectively sequentially.
+        on_error: Called as ``on_error(index, item, exception)`` to produce a
+            fallback result when a worker raises. If omitted, a failed item's
+            slot is left as ``None`` (still counted — length is preserved).
```
chandrasekharan-zipstack, line 53: 🟠 Failure is never returned as a signal — the function returns only `list[R]`, so with `on_error` omitted a failed slot is `None`, indistinguishable from a legitimate `None` result. Documenting the trap doesn't remove it. Worth considering returning `(results, failures: list[tuple[int, Exception]])`, or requiring an explicit error policy so a silent `None` is impossible.

workers/shared/parallel_map.py
```diff
@@ -0,0 +1,94 @@
...
+    Returns:
+        A list the SAME LENGTH as ``items``, results in input order. Failed
+        items hold either the ``on_error`` fallback or ``None``.
+    """
+    if not items:
+        return []
+
+    n = len(items)
+    effective_workers = max(max_workers, 1)
+    results: list[R | None] = [None] * n
```
chandrasekharan-zipstack, line 68: 🟡 Annotated `-> list[R]`, but the body declares `results: list[R | None]` and the docstring correctly says failed slots hold `None` when `on_error` is omitted. Type checkers will trust the annotation. Should be `-> list[R | None]`.

workers/shared/parallel_map.py
```diff
@@ -0,0 +1,94 @@
...
+        future_to_idx = {
+            executor.submit(worker, item): idx for idx, item in enumerate(items)
+        }
+        for future in as_completed(future_to_idx):
+            idx = future_to_idx[future]
+            try:
+                results[idx] = future.result()
+            except Exception as e:
+                with log_lock:
+                    logger.error(
```
chandrasekharan-zipstack: ⚪ `log_lock` isn't needed. Beyond `logging` already being thread-safe, the `as_completed` loop body runs **entirely on the calling thread** — `results[idx]` and this `logger.error` are single-threaded code, so there's no concurrent access at all. As written it signals a threading hazard that doesn't exist.

workers/shared/parallel_map.py
```diff
@@ -0,0 +1,94 @@
...
+                        "parallel_map: item %d/%d failed: %s",
```
chandrasekharan-zipstack: 🟠 `logger.error(..., exc_info=True)` fires once per failed item with no cap and no dedup. A 500-row table against a dead adapter or an exhausted quota emits 500 full stack traces — and 500 Sentry events for one logically identical failure. Suggest logging the first N with `exc_info`, the rest at debug, and one aggregated summary (`k/n failed, first error: ...`) after the pool drains.

workers/shared/parallel_map.py
```diff
@@ -0,0 +1,94 @@
...
+                        idx + 1,
+                        n,
+                        e,
+                        exc_info=True,
+                    )
+                if on_error is not None:
```
chandrasekharan-zipstack, line 101: 🟡 `on_error(...)` is invoked unguarded inside the `with ThreadPoolExecutor` block. If it raises, the exception propagates through `__exit__` → `shutdown(wait=True)`, so the call blocks until every remaining in-flight LLM call finishes before the error surfaces — a multi-minute hang on a large table with no log explaining it.

workers/file_processing/structure_tool_task.py
```diff
@@ -474,6 +474,10 @@ def _execute_structure_tool_impl(params: dict) -> dict:
             "execution_id": execution_id,
             "PLATFORM_SERVICE_API_KEY": platform_service_api_key,
             "group_key": at_settings.get("group_key", ""),
```
chandrasekharan-zipstack, line 481: ⚪ Three lines for one dict entry — "Lookup enrichment config" restates the key name and "None otherwise" restates `.get()`. Only "attached to the output upstream" carries information, and that's worth naming concretely (`prompt_studio_registry_helper.py:367`). Suggest collapsing to one line.

chandrasekharan-zipstack (commented):
## Review — UN-3575 OSS side

Reviewed together with the cloud companion Zipstack/unstract-cloud#1662.

Both changes are appropriately small and the `structure_tool_task.py` forwarding is exactly right. `parallel_map`'s core design — length-preserving, index-realigned, `max_workers=1` collapsing to sequential so one knob covers both strategies — is the correct shape for the per-row use case, and the no-early-abort NOTE is a genuinely good comment.

Main items inline. The one that matters most:

**`parallel_map.py:12-15`** — the docstring's justification for omitting rate limiting doesn't hold up against the SDK. 529 (Anthropic `overloaded_error`) isn't in `RETRYABLE_STATUS_CODES`, retries are config-gated rather than unconditional, and "concurrent callers self-throttle" doesn't follow from retry-with-backoff. Since that paragraph is the entire stated reason this utility ships without rate limiting, it's worth either correcting or just scoping the claim down.

Also worth a thought, though not blocking: `parallel_map` is a near-verbatim copy of the pre-existing `workers/plugins/agentic_table/src/utils/parallel_map.py` in the cloud repo (same `ThreadPoolExecutor` + `as_completed` + index-map structure), and today the only consumer is that same cloud plugin. Generalizing the plugin's own utility would avoid both the duplication and the new cross-repo import dependency. The OSS placement is defensible if other call sites are planned — just flagging the trade-off.

### Tests

Not asking for full coverage, but `parallel_map` is pure, dependency-free and trivially testable, and it lands in a CI-gated path (`unit-workers`, `coverage_source: shared`). Four asserts would pin the whole contract: order preserved under reversed completion, **length preserved on failure with `on_error=None`**, `on_error` fallback lands at the failed index, and `max_workers <= 1`. The middle one is the single behaviour that differs from `parallel_page_map` (which ends with `[r for r in results if r is not None]`) and therefore the module's entire reason to exist. `workers/plugins/agentic_table/tests/test_parallel_map.py` in the cloud repo already covers all of these for the predecessor and ports almost directly.

### Comments

Could you keep the new comments concise and generic? A few here narrate the PR's own context or reference symbols in the other repo, which won't read correctly to an OSS-only reader and will go stale as the code moves. Flagged inline.

workers/file_processing/structure_tool_task.py
praveen-formido: Done in e2c922f6d — collapsed to one line naming the source (`prompt_studio_registry_helper.py`).

workers/shared/parallel_map.py
praveen-formido: Fixed (e2c922f6d) — dropped the cloud-only `parallel_page_map` reference; the docstring now states the length-preserving property directly instead of comparing.

workers/shared/parallel_map.py
praveen-formido: You're right, thanks for checking the SDK. Reworded (e2c922f6d) to not overstate coverage: it no longer claims 429/500/503/overloaded are handled — it says bounding `max_workers` caps concurrency and that rate-limit/`Retry-After`/backoff is the LLM client's job, and to size `max_workers` conservatively.

workers/shared/parallel_map.py
praveen-formido: Thanks — kept it.

workers/shared/parallel_map.py
praveen-formido: Made the type honest (`list[R | None]`) and the docstring now recommends passing `on_error` so a failed slot is distinguishable from a legitimate `None`. Kept `on_error` optional rather than mandatory to preserve the generic contract — our per-row caller always passes it, so no ambiguous `None` arises in practice.

workers/shared/parallel_map.py
praveen-formido: Fixed — return type is now `list[R | None]`.

workers/shared/parallel_map.py
praveen-formido: Removed the lock — you're right, the `as_completed` loop body runs on the calling thread, so there's no concurrent access.

workers/shared/parallel_map.py
praveen-formido: Fixed — first 5 failures log with `exc_info`, the rest as one-liners, plus a single `k/n failed` summary after the pool drains.

workers/shared/parallel_map.py
praveen-formido: Fixed — `on_error` is now wrapped in try/except so a raising fallback can't propagate through the pool's `__exit__` (`shutdown(wait=True)`) and hang the caller.

workers/file_processing/structure_tool_task.py
praveen-formido: Collapsed to one line and named the source (`prompt_studio_registry_helper.py`).

workers/executor/executors/lookup_enrichment.py
```diff
@@ -32,6 +32,24 @@ def is_blank(value: Any) -> bool:
...
+    Returns ``None`` (fall back to per-call loading) when the plugin/config is
+    unavailable or the read fails.
+    """
+    lookup_cls = ExecutorPluginLoader.get("lookup-enrichment")
+    if not (lookup_config and lookup_cls):
+        return None
+    preload = getattr(lookup_cls, "preload_reference_texts", None)
+    if preload is None:
+        return None
+    return preload(lookup_config)
```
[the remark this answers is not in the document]
praveen-formido: Good catch — fixed in f2bf64014. Wrapped the plugin's `preload_reference_texts` call in try/except (and switched to `callable()`), so a raising hook now falls back to per-call reference loading per the documented None-on-failure contract instead of aborting enrichment.

chandrasekharan-zipstack (approved):

workers/shared/parallel_map.py
praveen-formido: The related concern on this NOTE — whether the SDK's retry actually covers rate-limits/overloaded — was addressed in the thread just below: the docstring was reworded to stop claiming 429/500/503/overloaded coverage and to say bounding `max_workers` caps concurrency while rate-limit/`Retry-After`/backoff is the LLM client's job (size `max_workers` conservatively). Resolving to keep the PR tidy — @muhammad-ali-e, please reopen or drop a note if you'd still like to weigh in on the behaviour.

muhammad-ali-e: I think here we talking about cancelling an action, right?. Since we don't have cancel or aborttin in existing system , we should do that but not required now. We might implement it during pg queue implemenation later. But Ithink this parallel concept should be a optional one.

PR #2192, opened 2026-07-22
UN-3636 [FIX] Stop re-running cross-tier deps in every tier leg

chandrasekharan-zipstack:
## What

Bound the rig's `depends_on` expansion to the requested tier when `--tier` is passed.

## Why

`integration-workflow-execution` and `e2e-smoke` both declare `depends_on: [unit-sdk1, unit-workers]`. Because `--tier` expanded deps transitively with no tier bound, those two unit groups ran **three times per CI run** — once in the unit leg, again in the integration leg, again in the e2e leg.

Measured on run `29891886572` (main):

| group | unit leg | integration leg | e2e leg |
|---|---:|---:|---:|
| `unit-sdk1` | 21s | 27s | 27s |
| `unit-workers` | 37s | 51s | 51s |

≈136s of duplicate execution per run. On the cloud CI runner (`custom-k8s-runner`, ~2 xdist workers) the same duplication costs **~350s**.

## Changes

- `tests/rig/selection.py`: with `--tier`, drop dep-expanded groups outside that tier. Explicitly named groups are never dropped.
- `tests/rig/tests/test_selection.py`: +3 tests — cross-tier dep not pulled in, explicit request survives the filter, expansion unchanged without a tier filter.

## Safety

- Intra-tier deps still expand and topo-order (`e2e-smoke` → `e2e-login`).
- Gating is unaffected: `blocked_by` intersects with groups that failed *in the same run*, so a dep that wasn't run simply doesn't gate — and its own tier leg gates it.
- `tests.rig validate` OK — 18 groups, 16 critical paths. Full rig suite: 74 passed, 1 skipped.

### DIFF
--- a/tests/rig/selection.py
+++ b/tests/rig/selection.py
@@ -7,6 +7,10 @@ Resolution order (de-duped, then dep-expanded, then topo-sorted by the manifest)
 The literal ``all`` expands to every group in the manifest. When the result is
 empty, callers should treat that as "do nothing" rather than silently running
 everything — the CLI surfaces a clear error.
+
+With ``--tier``, dep expansion never reaches outside that tier: each tier runs as
+its own CI leg, so a cross-tier ``depends_on`` would otherwise re-run the same
+group in every leg that transitively depends on it.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


chandrasekharan-zipstack:
Folding these changes into #2188 instead of carrying a second rig PR — the cross-tier dep fix now rides on `feat/rig-extra-manifests` (commit 793b32f0c).

PR #2197, opened 2026-07-23
UN-3798 [FIX] Skip broken file-path task load for PG pluggable workers

muhammad-ali-e:
## What
- For pluggable `WORKER_TYPE`s, the top-level `workers/worker.py` no longer loads the plugin's `tasks.py` by file path under a bare `"tasks"` module name. Their tasks are already registered by `WorkerBuilder.build_celery_app()`; the file-path load is skipped.
- Non-pluggable (top-level) workers keep the existing file-path load, unchanged.

## Why
- The file-path load used `spec_from_file_location("tasks", ...)` — a bare name with **no parent package** — so every plugin `tasks.py` relative import (`from .clients import ...`) failed with `ImportError: attempted relative import with no known parent package`, crash-looping the worker on startup under `python -m pg_queue_consumer`.
- This blocks **every** cloud PG pluggable worker (bulk-download UN-3752, agentic-callback UN-3754, agentic_studio UN-3779). It was never caught because these workers are disabled in all base values and no test had started one via the PG consumer.
- The Celery path is unaffected: Celery pluggable workers run via `celery -A pluggable_worker.{type}.worker` (a proper dotted package import), which resolves relative imports and never touches this loader. The `is_pluggable()` file-path branch never actually ran successfully.

## How
- `build_celery_app()` (called just above the loader) verifies a pluggable type via `_verify_pluggable_worker_exists` → `importlib.import_module("pluggable_worker.{type}.worker")` — a package import that runs the plugin's `from . import tasks` and registers the tasks on this same app. So the file-path load is redundant + broken; skip it for pluggable workers.

- No. The Celery path and non-pluggable file-path path are untouched (non-pluggable load simply moved into the `else` branch). Removes a code path that never ran successfully. PG and Celery register identical tasks (both via `build_celery_app`).

## Database Migrations
- None.

## Env Config
- None.

## Notes on Testing
- Validated on a running dev stack via `python -m pg_queue_consumer`: `agentic_callback` and `agentic_studio` now start clean (`tasks already registered … skipping file-path task load` → `ready for Celery`); `general` (non-pluggable) loads unchanged via the file-path path.
- No unit test: `worker.py` is module-level startup code and this repo has no pluggable-worker fixture; the runtime worker-start is the validation.

## Related Issues or PRs
- Prerequisite for UN-3752, UN-3754, UN-3779.

### DIFF
--- a/workers/worker.py
+++ b/workers/worker.py
@@ -444,36 +444,50 @@ logger.info("✅ Worker infrastructure initialized successfully")
 # Determine worker path dynamically based on worker type
 base_dir = os.path.dirname(os.path.abspath(__file__))
 if worker_type.is_pluggable():
-    # Pluggable workers live inside workers/pluggable_worker/{worker_name}
-    worker_directory = os.path.join("pluggable_worker", worker_type.value)
-    worker_path = os.path.join(base_dir, worker_directory)
+    # Pluggable workers' tasks are ALREADY registered by this point:
+    # build_celery_app() above verifies the plugin via
+    # WorkerBuilder._verify_pluggable_worker_exists, which does
+    # `importlib.import_module("pluggable_worker.{type}.worker")` — a proper
+    # PACKAGE import that runs the plugin's own `from . import tasks`, resolving
+    # its relative imports (`from .clients import ...`) and registering the tasks
+    # on this same app. The generic file-path load below loads tasks.py under a
+    # bare "tasks" spec (no parent package), which BREAKS those relative imports
+    # ("attempted relative import with no known parent package"), so it MUST be
+    # skipped for pluggable workers — otherwise every cloud PG pluggable worker
+    # crashes on startup.
+    logger.info(
+        f"✅ Pluggable worker {worker_type.value} tasks already registered via "
+        "WorkerBuilder (package import); skipping file-path task load"
+    )
 else:
-    # Enum values use underscores (Python module names); a few on-disk dirs
-    # still use hyphens (e.g. api-deployment). Derive the directory from the
+    # Non-pluggable (top-level) workers register tasks via a file-path load of
+    # their tasks.py — they use absolute imports, with their dir on sys.path.
+    # Enum values use underscores (Python module names); a few on-disk dirs still
+    # use hyphens (e.g. api-deployment). Derive the directory from the
     # authoritative import-path map on WorkerType instead of a blind replace.
     worker_directory = worker_type.to_import_path().rsplit(".", 1)[0]
     worker_path = os.path.join(base_dir, worker_directory)
 
-# Add worker directory to path for task imports
-if os.path.exists(worker_path):
-    sys.path.append(worker_path)
-    logger.info(f"✅ Added {worker_directory} to Python path for task imports")
+    # Add worker directory to path for task imports
+    if os.path.exists(worker_path):
+        sys.path.append(worker_path)
+        logger.info(f"✅ Added {worker_directory} to Python path for task imports")
 
-    # Import tasks module to register tasks
-    tasks_file = os.path.join(worker_path, "tasks.py")
-    if os.path.exists(tasks_file):
-        logger.info(f"📋 Loading tasks from: {tasks_file}")
-        # Import the tasks module to register tasks with the app
-        import importlib.util
+        # Import tasks module to register tasks
+        tasks_file = os.path.join(worker_path, "tasks.py")
+        if os.path.exists(tasks_file):
+            logger.info(f"📋 Loading tasks from: {tasks_file}")
+            # Import the tasks module to register tasks with the app
+            import importlib.util
 
-        spec = importlib.util.spec_from_file_location("tasks", tasks_file)
-        tasks_module = importlib.util.module_from_spec(spec)
-        spec.loader.exec_module(tasks_module)
-        logger.info(f"✅ Tasks loaded successfully from {worker_directory}")
+            spec = importlib.util.spec_from_file_location("tasks", tasks_file)
+            tasks_module = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(tasks_module)
+            logger.info(f"✅ Tasks loaded successfully from {worker_directory}")
+        else:
+            logger.warning(f"⚠️ No tasks.py found at: {tasks_file}")
     else:
-        logger.warning(f"⚠️ No tasks.py found at: {tasks_file}")
-else:
-    logger.error(f"❌ Worker directory not found: {worker_path}")
+        logger.error(f"❌ Worker directory not found: {worker_path}")
 
 # Log successful configuration
 logger.info(f"✅ Successfully loaded {worker_type} worker using WorkerBuilder")
--- a/workers/worker.py
+++ b/workers/worker.py
@@ -5,6 +5,7 @@ This module serves as the main entry point for all Celery workers.
 It uses WorkerBuilder to ensure proper configuration including chord retry settings.
 """
 
+import importlib.util
 import logging
 import os
 import sys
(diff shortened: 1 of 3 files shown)


muhammad-ali-e (commented):
Automated PR review (PR Review Toolkit: Code Reviewer, Silent-Failure Hunter, Type-Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier).

**Verdict:** the fix itself is correct and behavior-preserving — the file-path load is cleanly narrowed to the non-pluggable branch, and the pluggable skip is safe (a missing/broken plugin already raises `ImportError` in `build_celery_app` before this code runs). No blocking bugs. The comments below are mostly comment-accuracy fixes plus optional hardening/refactor suggestions.

workers/worker.py
```diff
@@ -444,36 +444,50 @@ def on_task_postrun(sender=None, task_id=None, **kwargs):
...
-    # Pluggable workers live inside workers/pluggable_worker/{worker_name}
-    worker_directory = os.path.join("pluggable_worker", worker_type.value)
-    worker_path = os.path.join(base_dir, worker_directory)
+    # Pluggable workers' tasks are ALREADY registered by this point:
+    # build_celery_app() above verifies the plugin via
+    # WorkerBuilder._verify_pluggable_worker_exists, which does
+    # `importlib.import_module("pluggable_worker.{type}.worker")` — a proper
+    # PACKAGE import that runs the plugin's own `from . import tasks`, resolving
+    # its relative imports (`from .clients import ...`) and registering the tasks
+    # on this same app. The generic file-path load below loads tasks.py under a
```
muhammad-ali-e: **[Comment accuracy — MEDIUM]** "registering the tasks on this same app" is contradicted by call ordering. `_verify_pluggable_worker_exists()` performs the `import_module` at `builder.py:60`, **before** `app = Celery(app_name)` is created at `builder.py:74`. The app object this file uses does not exist yet when the plugin is imported, so registration must go through Celery's deferred binding (`@shared_task` / `connect_on_app_finalize`), not a direct attach to "this same app".

_Suggested wording:_ "…runs the plugin's task-registration code. Celery binds those tasks to the worker's app when it finalizes, so the tasks are already registered by this point."

workers/worker.py
```diff
@@ -444,36 +444,50 @@ def on_task_postrun(sender=None, task_id=None, **kwargs):
...
+    # bare "tasks" spec (no parent package), which BREAKS those relative imports
+    # ("attempted relative import with no known parent package"), so it MUST be
+    # skipped for pluggable workers — otherwise every cloud PG pluggable worker
```
muhammad-ali-e: **[Comment accuracy — LOW]** "otherwise **every** cloud PG pluggable worker crashes on startup" overstates the guarantee. The `attempted relative import with no known parent package` crash only fires for plugins whose `tasks.py` actually uses relative imports; a plugin written with absolute imports would not crash. Suggest softening to the real precondition, e.g. "…which breaks any relative imports in the plugin's `tasks.py`, so the file-path load must be skipped for pluggable workers."

workers/worker.py
```diff
@@ -444,36 +444,50 @@ def on_task_postrun(sender=None, task_id=None, **kwargs):
...
 base_dir = os.path.dirname(os.path.abspath(__file__))
 if worker_type.is_pluggable():
-    # Pluggable workers live inside workers/pluggable_worker/{worker_name}
-    worker_directory = os.path.join("pluggable_worker", worker_type.value)
-    worker_path = os.path.join(base_dir, worker_directory)
+    # Pluggable workers' tasks are ALREADY registered by this point:
+    # build_celery_app() above verifies the plugin via
+    # WorkerBuilder._verify_pluggable_worker_exists, which does
+    # `importlib.import_module("pluggable_worker.{type}.worker")` — a proper
+    # PACKAGE import that runs the plugin's own `from . import tasks`, resolving
```
muhammad-ali-e: **[Maintainability / rot risk — LOW]** This comment hard-codes the internal structure of cloud-only plugin code that is not in this repo (`pluggable_worker/` is gitignored in OSS): `from . import tasks`, `from .clients import ...`. Nobody editing the OSS side can validate these specifics, and they silently rot if a plugin registers tasks differently. Prefer stating the *contract* ("pluggable workers are proper packages; verification imports `pluggable_worker.<type>.worker` as a package, running its own task registration incl. relative imports") and marking the concrete `from . import ...` examples as illustrative ("e.g.").

workers/worker.py
```diff
@@ -444,36 +444,50 @@ def on_task_postrun(sender=None, task_id=None, **kwargs):
 # Determine worker path dynamically based on worker type
 base_dir = os.path.dirname(os.path.abspath(__file__))
 if worker_type.is_pluggable():
-    # Pluggable workers live inside workers/pluggable_worker/{worker_name}
-    worker_directory = os.path.join("pluggable_worker", worker_type.value)
-    worker_path = os.path.join(base_dir, worker_directory)
+    # Pluggable workers' tasks are ALREADY registered by this point:
```
muhammad-ali-e: **[Structure / testability — suggestion]** Three agents (Simplifier, Test Analyzer, Type-Design) converged here: extract lines 445–490 into a `load_worker_tasks(worker_type)` helper. Benefits: (1) the giant explanatory comment becomes a proper docstring; (2) nesting collapses from depth 3 to depth 1 via guard-clause `return`s (impossible at module scope today); (3) it creates the seam for the single highest-value regression test — assert `spec_from_file_location` is **never** called when `is_pluggable()` is True — which permanently pins this fix against a future re-merge of the branches. All locals (`base_dir`, `worker_directory`, `worker_path`, `tasks_file`, `spec`, `tasks_module`) are used only in this block, so extraction leaks nothing. Reasonable to defer as a fast-follow given this is a hotfix.

workers/worker.py
```diff
@@ -444,36 +444,50 @@ def on_task_postrun(sender=None, task_id=None, **kwargs):
...
+            tasks_module = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(tasks_module)
+            logger.info(f"✅ Tasks loaded successfully from {worker_directory}")
+        else:
+            logger.warning(f"⚠️ No tasks.py found at: {tasks_file}")
     else:
-        logger.warning(f"⚠️ No tasks.py found at: {tasks_file}")
-else:
-    logger.error(f"❌ Worker directory not found: {worker_path}")
+        logger.error(f"❌ Worker directory not found: {worker_path}")
```
muhammad-ali-e, line 474: **[Silent failure — MEDIUM]** Two zero-task boot paths survive here: (a) this `else` logs an error but **execution continues** to the `✅ Successfully loaded` line at 492 — a misleading success after a fatal misconfig (pre-existing, just re-indented); (b) after removing the file-path load for pluggable workers, plugin-side import is now the *sole* task-registration mechanism with no post-check. Celery does not error on an empty task registry — the worker starts and silently processes nothing. Consider one guard after the if/else that covers both paths:
```python
registered = [n for n in app.tasks if not n.startswith("celery.")]
if not registered:
    raise RuntimeError(
        f"No tasks registered for worker '{worker_type.value}' "
        f"(pluggable={worker_type.is_pluggable()}) — worker would start but process nothing."
    )
```

workers/worker.py
```diff
@@ -444,36 +444,50 @@ def on_task_postrun(sender=None, task_id=None, **kwargs):
...
+        f"✅ Pluggable worker {worker_type.value} tasks already registered via "
+        "WorkerBuilder (package import); skipping file-path task load"
+    )
 else:
-    # Enum values use underscores (Python module names); a few on-disk dirs
-    # still use hyphens (e.g. api-deployment). Derive the directory from the
+    # Non-pluggable (top-level) workers register tasks via a file-path load of
+    # their tasks.py — they use absolute imports, with their dir on sys.path.
+    # Enum values use underscores (Python module names); a few on-disk dirs still
+    # use hyphens (e.g. api-deployment). Derive the directory from the
```
muhammad-ali-e: **[Type design — LOW]** Deriving the on-disk directory as `worker_type.to_import_path().rsplit(".", 1)[0]` (line 468) reuses an import-path string as a filesystem path — it only works because every path is `"{dir}.tasks"`. Consider an explicit `WorkerType.to_directory() -> str` that owns the hyphen/underscore mapping, so intent is named and the mapping lives in one place. (Keep the task-loading logic itself out of the enum — the other `WorkerType` methods are pure derivations; a `load_tasks()` would drag `sys.path`/`importlib`/the Celery app into a lightweight `str, Enum`.)

workers/worker.py
```diff
@@ -444,36 +444,50 @@ def on_task_postrun(sender=None, task_id=None, **kwargs):
...
+    if os.path.exists(worker_path):
+        sys.path.append(worker_path)
+        logger.info(f"✅ Added {worker_directory} to Python path for task imports")
+
+        # Import tasks module to register tasks
+        tasks_file = os.path.join(worker_path, "tasks.py")
+        if os.path.exists(tasks_file):
+            logger.info(f"📋 Loading tasks from: {tasks_file}")
+            # Import the tasks module to register tasks with the app
+            import importlib.util
```
muhammad-ali-e: **[Nit]** `import importlib.util` is done inside the innermost `if`. Since `os`/`sys` are already imported at module top, moving this up with them is more consistent. Style-only.

muhammad-ali-e:
Addressed all 7 threads in `f7905a3e9`.

- **Deferred-binding comment accuracy** — reworded: pluggable tasks register via WorkerBuilder's package import and **Celery binds them to the app on finalize**, so they're registered by this point (dropped the "on this same app" phrasing that contradicted the `import_module`-before-`Celery()` ordering).
- **"every worker crashes" overstated** — softened to the real precondition: the file-path load "breaks **any relative imports** in the plugin's `tasks.py`", so it's skipped for pluggable workers.
- **Hard-coded cloud-plugin internals** — now states the contract (pluggable workers are proper packages; verification imports the package and runs its own registration) and marks `from . import tasks` as illustrative (`e.g.`).
- **Extract to a helper** — extracted `load_worker_tasks(worker_type)` with guard-clause returns (nesting depth 3 → 1) and the explanatory comment moved into a docstring.
- **Silent zero-task boot** — added a post-load check, but as a **WARNING, not a hard raise**: pluggable tasks bind on app finalize (`@shared_task`/`connect_on_app_finalize`), which can be *after* this point, so a raise here would false-positive on a correctly-configured pluggable worker — i.e. re-introduce the very crash this PR fixes. The warning surfaces the misconfig without that risk.
- **`to_directory()` type design** — added `WorkerType.to_directory()` as the single source of the underscore→hyphen mapping; `to_import_path()` and the loader both read it (no more slicing the import path). Pinned by tests.
- **`import importlib.util` nit** — moved to the module-top imports.

Deferred (per your hotfix note): the "`spec_from_file_location` never called when `is_pluggable()`" regression test. `worker.py` runs infra init + builds the Celery app at import, so there's no clean seam to call the loader in isolation without a larger `main()`-guard refactor. The extraction creates that seam for a fast-follow; in the meantime `to_directory()` is covered by `tests/test_worker_enums_directory.py` (3 tests pass).

PR #2198, opened 2026-07-23
UN-3753 [GATED-FEAT] Route webhook notifications through PG-queue transport (flag-gated)

muhammad-ali-e:
## What

- Adds `notification_v2/notification_dispatch.py` — a flag-gated transport seam for the buffered-webhook dispatch, mirroring `pipeline_dispatch.py`.
- Routes `_send_clubbed` (the notification buffer-flush path) through the seam: PG queue when `pg_queue_enabled` is on for the org, else the prior `celery_app.send_task` (byte-identical).
- `_org_identifier()` resolves the buffer's org **pk** → the string `organization_id` that `resolve_transport` needs for its Flipt decision (the org pk stays in the task kwargs — the worker's buffer-mark contract).
- 6 unit tests (`test_notification_dispatch.py`) pinning both routes + fail-closed behavior.

## Why

- Part of migrating the notification worker off Celery onto the PG queue (UN-3753, under the PG-queue epic UN-3445), so Celery can eventually be decommissioned.
- Universal pattern: flag-gate the enqueue point; flag OFF keeps Celery byte-unchanged, flag ON enqueues onto PG (drained by the PG notification consumer).

## How

- `resolve_transport(execution_id, organization_id)` decides the transport; `is_pg_transport()` is the single "counts as PG" check.
- PG path: `enqueue_task(task_name="send_webhook_notification", queue="notifications", …)` — the exact args/kwargs the Celery call used, JSON-normalized by the producer.
- Celery path: unchanged `send_task` with identical args/kwargs/queue.
- Scope: only site 1 (`_send_clubbed`, the production flush path) is migrated. The `WebhookSend` / `WebhookBatch` internal endpoints stay on Celery — they use a `countdown` stagger the PG queue has no delayed-visibility for (a follow-up).

- No. With `pg_queue_enabled` OFF (the default) `resolve_transport` fails closed to Celery and the seam calls the same `send_task` as before — byte-identical. The PG branch only runs when the flag is explicitly enabled for an org. Dev-tested on a live stack: flag-off produced 0 PG rows (Celery taken); an enqueued task was drained by the PG notification consumer and delivered end-to-end.

## Database Migrations

- None.

## Env Config

- None (uses the existing `pg_queue_enabled` Flipt flag).

## Notes on Testing

- `pytest notification_v2/tests/test_notification_dispatch.py` — 6 tests pass; `pipeline_dispatch` regression suite green.
- Live dev-test: flag-off Celery parity (0 PG rows); PG round-trip enqueue → `pg_queue_consumer` (WORKER_TYPE=notification) → webhook delivered (200) → row acked.

## Related Issues or PRs

- UN-3753 (sub-task of UN-3445). Cloud counterpart: `workerPgNotification` chart (batched into the aux-workers cloud PR). Depends on the epic branch merging first, then rebased to main.

### DIFF
--- a/backend/notification_v2/internal_api_views.py
+++ b/backend/notification_v2/internal_api_views.py
@@ -515,19 +515,32 @@ def _reclaim_stale_sending() -> int:
     return int(reclaimed)
 
 
-def _org_identifier(org_pk: Any) -> str | None:
+def _org_identifier(org_pk: int) -> str | None:
     """Resolve the string ``Organization.organization_id`` from the buffer's org pk.
 
     ``resolve_transport`` keys its Flipt decision on the org's string identifier,
     but the buffer stores/uses the Organization pk. One indexed pk lookup per
-    dispatch group (post-commit) — negligible before the webhook HTTP call it
-    precedes. ``None`` (org deleted) fails closed to Celery in resolve_transport.
+    dispatch group (post-commit) — negligible relative to the downstream webhook
+    dispatch.
+
+    Data-anomaly guard: ``NotificationBuffer.organization`` is
+    ``on_delete=CASCADE``, so a live buffer row with a missing org is unreachable
+    in normal operation. A ``None`` here therefore signals a dangling FK — we log
+    it (the only org-traceable breadcrumb; resolve_transport's own warning is keyed
+    on the random dispatch uuid) and fail closed to Celery in resolve_transport.
     """
-    return (
+    org_string_id = (
         Organization.objects.filter(pk=org_pk)
         .values_list("organization_id", flat=True)
         .first()
     )
+    if org_string_id is None:
+        logger.warning(
+            "metric=notification_org_identifier_missing_total org_pk=%s "
+            "(dangling FK; notification routing falls back to Celery)",
+            org_pk,
+        )
+    return org_string_id
 
 
 def _send_clubbed(
@@ -578,7 +591,7 @@ def _send_clubbed(
                 "organization_id": org_id,
             },
             queue="notifications",
-            organization_id=_org_identifier(org_id),
+            org_string_id=_org_identifier(org_id),
         )
         logger.info(
             "metric=notification_batch_dispatched_total platform=%s result=success "
@@ -588,7 +601,31 @@ def _send_clubbed(
             webhook_url_hash(url),
             len(buffer_ids),
         )
+    except (ValueError, TypeError):
+        # PERMANENT dispatch errors — enqueue_task validation (priority range /
+        # reply_key+callback exclusivity) or payload serialization — fail
+        # identically every flush tick. Dead-letter now (distinct metric) instead
+        # of reverting to PENDING and re-rendering + re-dispatching + emitting a
+        # broker_failure traceback until the attempt cap. Guard on SENDING so a row
+        # the worker already resolved isn't clobbered.
+        logger.exception(
+            "metric=notification_batch_dispatched_total platform=%s "
+            "result=dispatch_error org_id=%s webhook_url_hash=%s rows=%d",
+            platform,
+            org_id,
+            webhook_url_hash(url),
+            len(buffer_ids),
+        )
+        NotificationBuffer.objects.filter(
+            id__in=buffer_ids,
+            status=BufferStatus.SENDING.value,
+        ).update(status=BufferStatus.DEAD_LETTER.value)
     except Exception:
+        # TRANSIENT transport/broker failure — revert to PENDING (outside the
+        # committed txn) so the next flush tick retries; refund the SENDING-claim
+        # attempt since nothing was queued or sent. Guard on SENDING so a row the
+        # worker already marked terminal (broker raised post-delivery) isn't
+        # resurrected into a duplicate.
         logger.exception(
             "metric=notification_batch_dispatched_total platform=%s "
             "result=broker_failure org_id=%s webhook_url_hash=%s rows=%d",
@@ -597,10 +634,6 @@ def _send_clubbed(
             webhook_url_hash(url),
             len(buffer_ids),
         )
-        # Revert to PENDING (outside the committed txn) so a transient broker
-        # outage retries next tick; refund the SENDING-claim attempt since nothing
-        # was queued or sent. Guard on SENDING so a row the worker already marked
-        # terminal (broker raised post-delivery) isn't resurrected into a duplicate.
         NotificationBuffer.objects.filter(
             id__in=buffer_ids,
             status=BufferStatus.SENDING.value,
--- a/backend/notification_v2/internal_api_views.py
+++ b/backend/notification_v2/internal_api_views.py
@@ -38,7 +38,10 @@ from notification_v2.helper import (
     webhook_url_hash,
 )
 from notification_v2.models import Notification, NotificationBuffer
-from notification_v2.notification_dispatch import dispatch_webhook_notification
+from notification_v2.notification_dispatch import (
+    PermanentDispatchError,
+    dispatch_webhook_notification,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -601,13 +604,16 @@ def _send_clubbed(
             webhook_url_hash(url),
             len(buffer_ids),
         )
-    except (ValueError, TypeError):
-        # PERMANENT dispatch errors — enqueue_task validation (priority range /
-        # reply_key+callback exclusivity) or payload serialization — fail
+    except PermanentDispatchError:
+        # PG-ONLY permanent failure — enqueue_task validation (priority range /
+        # reply_key+callback exclusivity) or payload serialization — fails
         # identically every flush tick. Dead-letter now (distinct metric) instead
         # of reverting to PENDING and re-rendering + re-dispatching + emitting a
-        # broker_failure traceback until the attempt cap. Guard on SENDING so a row
-        # the worker already resolved isn't clobbered.
+        # broker_failure traceback until the attempt cap. The Celery path never
+        # raises this, so the flag-off flow is unchanged: a Celery send_task failure
+        # is an ordinary Exception handled by the transient branch below, exactly as
+        # before UN-3753. Guard on SENDING so a row the worker already resolved
+        # isn't clobbered.
         logger.exception(
             "metric=notification_batch_dispatched_total platform=%s "
             "result=dispatch_error org_id=%s webhook_url_hash=%s rows=%d",
--- a/backend/notification_v2/internal_api_views.py
+++ b/backend/notification_v2/internal_api_views.py
@@ -14,6 +14,7 @@ import logging
 from datetime import timedelta
 from typing import Any, cast
 
+from account_v2.models import Organization
 from api_v2.models import APIDeployment
 from django.conf import settings
 from django.db import transaction
@@ -37,6 +38,7 @@ from notification_v2.helper import (
     webhook_url_hash,
 )
 from notification_v2.models import Notification, NotificationBuffer
+from notification_v2.notification_dispatch import dispatch_webhook_notification
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


backend/notification_v2/internal_api_views.py
```diff
@@ -540,8 +557,12 @@ def _send_clubbed(
     ``buffer_row_ids`` + ``organization_id`` to the worker so it can mark them.
     """
     try:
-        celery_app.send_task(
-            "send_webhook_notification",
+        # Flag-gated transport (UN-3753): PG queue when pg_queue_enabled for this
+        # org, else Celery (byte-identical to the prior send_task). resolve_transport
+        # keys on the org STRING id, but the buffer/worker contract below uses the
+        # org pk — hence _org_identifier(org_id) for routing, org_id in kwargs.
+        dispatch_webhook_notification(
```
muhammad-ali-e, line 580: **[P1 · Robustness] Broad `except Exception` mislabels permanent dispatch errors as `broker_failure` and retries them forever.**

The `except Exception` wrapping this call (below, ~L591) assumes a *transient broker outage*: it logs `result=broker_failure` and reverts the rows `SENDING → PENDING` to retry next tick. But it catches **every** exception type from `dispatch_webhook_notification` identically, including permanent ones:

- `ValueError` from `enqueue_task` (priority range / reply_key+callback mutual-exclusion validation)
- `TypeError`/serialization errors from `_json_safe(...)` on a non-encodable payload
- programming errors (`AttributeError`/`KeyError`) in the seam, or a DB error from the `_org_identifier(org_id)` lookup (evaluated inside the `try` as a call arg)

All of these are permanent — they will fail identically every flush tick, re-render + re-dispatch the group, and emit a full `logger.exception` traceback to Sentry mislabeled as a transport failure, until the dispatch-attempt cap eventually dead-letters it.

**Fix:** branch on error class — mark permanent errors (`ValueError`, serialization `TypeError`) `DEAD_LETTER` immediately with a distinct metric (e.g. `result=dispatch_error`), and keep the revert-to-PENDING path only for genuine transport/broker exceptions.

_Minor related: a PG enqueue failure is logged with `logger.exception` both inside `enqueue_task` and again here → two Sentry issues per PG failure vs one on the Celery path. Pick one owner for the traceback._

backend/notification_v2/internal_api_views.py
```diff
@@ -557,6 +578,7 @@ def _send_clubbed(
                 "organization_id": org_id,
             },
             queue="notifications",
+            organization_id=_org_identifier(org_id),
```
muhammad-ali-e: **[P1 · Tests] The PR's actual behavioral change here is untested.**

The six new tests exercise `dispatch_webhook_notification` in isolation, but nothing covers this call site — where the two highest-risk regressions live:

1. **The two-org-values contract.** Routing uses the **string** id (`organization_id=_org_identifier(org_id)`) while the worker's buffer-mark contract uses the **pk** (`kwargs["organization_id"]=org_id`). Swapping them would pass every existing test yet silently (a) fail routing closed to Celery for every org (a pk never matches a Flipt string-id rule) and (b) strand rows in `SENDING` (the mark endpoint filters on the pk). Add a mock-based test asserting `call.kwargs["organization_id"] == "<string id>"` and `call.kwargs["kwargs"]["organization_id"] == <pk>`.
2. **The revert-on-failure recovery.** `test_pg_enqueue_failure_propagates...` only asserts the dispatch *raises*; the safety-critical half — `_send_clubbed` reverting only `SENDING` rows to `PENDING`, `dispatched_at=None`, `dispatch_attempts` F(-1), leaving already-`DISPATCHED` rows untouched — is untested. Best as a real DB test (seed one `SENDING` + one `DISPATCHED` row, patch dispatch to raise, assert the conditional update).
3. `_org_identifier` itself has no test: org found → string id (not `"None"`), org missing → `None`.

backend/notification_v2/internal_api_views.py
```diff
@@ -513,6 +515,21 @@ def _reclaim_stale_sending() -> int:
     return int(reclaimed)
 
 
+def _org_identifier(org_pk: Any) -> str | None:
+    """Resolve the string ``Organization.organization_id`` from the buffer's org pk.
+
+    ``resolve_transport`` keys its Flipt decision on the org's string identifier,
+    but the buffer stores/uses the Organization pk. One indexed pk lookup per
+    dispatch group (post-commit) — negligible before the webhook HTTP call it
+    precedes. ``None`` (org deleted) fails closed to Celery in resolve_transport.
```
muhammad-ali-e: **[P2 · Observability + comment accuracy] Silent `None` return, and the "org deleted" framing is misleading given the CASCADE FK.**

`NotificationBuffer.organization` is `on_delete=models.CASCADE`, so deleting an Organization cascade-deletes its buffer rows — the "live buffer row + missing org" state this docstring describes is essentially unreachable in normal operation. If `_org_identifier` *does* return `None`, it signals a real data-integrity anomaly (dangling FK), yet it's swallowed with no log here; the only trace is `resolve_transport`'s warning keyed on the random dispatch UUID, untraceable back to the org/rows.

**Fix:** log a `warning` (including `org_pk`) when the lookup returns `None`, and reword the comment to "data-anomaly guard" rather than an expected-deletion path.

_Nits: "negligible before the webhook HTTP call it precedes" — the HTTP call happens later in the worker, not here; prefer "negligible relative to the downstream webhook dispatch." Also `org_pk: Any` could be narrowed to `int` (it's a pk throughout)._

backend/notification_v2/notification_dispatch.py
```diff
@@ -0,0 +1,95 @@
...
+WEBHOOK_NOTIFICATION_TASK = "send_webhook_notification"
+
+
+def dispatch_webhook_notification(
+    *,
+    celery_app: Any,
+    args: list[Any],
+    kwargs: dict[str, Any],
+    queue: str,
+    organization_id: str | None,
```
muhammad-ali-e: **[P2 · Type design] The `organization_id` name is overloaded across the boundary — the core latent trap in this seam.**

This param is the org **string** identifier (for Flipt routing), while `kwargs["organization_id"]` (passed by the caller) is the org **pk** (int, the worker buffer-mark contract). Same name, two incompatible meanings, both effectively untyped (`str | None` and a value inside `dict[str, Any]`). A future maintainer who "unifies" or swaps them gets **silent mis-routing**, not a type error: `resolve_transport` would `str()` the pk to `"5"`, match no org segment, and quietly force Celery.

**Cheapest high-value fix:** rename the routing param to `org_string_id` / `routing_org_identifier` (one call-site edit) so the two identities can't be confused. Optionally encode them as distinct `NewType`s (`OrgStringId`, `OrgPk`) to make a swap a type error at zero runtime cost.

backend/notification_v2/notification_dispatch.py
```diff
@@ -0,0 +1,95 @@
...
+            buffer-mark contract) — distinct from the ``organization_id`` param
+            below, which is the string identifier used only for flag routing.
+        queue: Target queue name, forwarded verbatim.
+        organization_id: The org's **string** identifier
+            (``Organization.organization_id``), used solely for the Flipt
+            transport decision. ``None`` (or empty) fails closed to Celery.
+
+    Returns:
+        A task id string — the Celery ``AsyncResult`` id on the Celery path, or
+        the minted PG task id on the PG path (usable by callers that surface it
```
muhammad-ali-e: **[P2 · Comment accuracy] The return value's "usable by callers that surface it in an API response" is aspirational — no caller uses it.**

The sole caller, `_send_clubbed`, is a fire-and-forget `transaction.on_commit` path and discards the return entirely; there is no API response in sight. The claim invites a future maintainer to believe a real consumer depends on it.

_The rest of the `Returns:` block is accurate and worth keeping: I confirmed the function returns the minted `dispatch_id` (uuid `str`) on the PG path — deliberately **not** `enqueue_task`'s `int` `msg_id` — giving Celery-parity `str` semantics, and three tests pin the `dispatch_id == task_id == Flipt bucket` invariant._

**Fix:** drop the parenthetical or mark it explicitly hypothetical ("returned for symmetry / any future caller").

backend/notification_v2/notification_dispatch.py
```diff
@@ -0,0 +1,95 @@
...
+        in an API response).
+    """
+    # A buffered notification is a single fire-and-forget task with no natural
+    # sticky entity, so mint a fresh id to drive Flipt's percentage bucketing and
+    # to serve as the PG task id.
+    dispatch_id = str(uuid.uuid4())
+    transport = resolve_transport(
+        execution_id=dispatch_id,
+        organization_id=organization_id or None,
```
muhammad-ali-e: **[P3 · Simplify] `organization_id or None` is redundant.**

`resolve_transport` already normalizes falsy input on its first line (`if not organization_id: return celery`), so `""` and `None` resolve identically — the `or None` (converting `""` → `None`) can't change the outcome. It's also unreachable in practice: the sole caller passes `_org_identifier(...)`, whose `.values_list(...).first()` yields `str | None`, never `""`. Passing `organization_id` directly is behavior-preserving (the `test_none_org_fails_closed_to_celery` identity assertion still holds).

_Note: the sibling `org_id=organization_id or ""` on the PG path is **not** redundant — it satisfies `enqueue_task`'s `org_id: str` contract, so leave that one._

muhammad-ali-e (commented):

No blocking correctness bugs found — the flag-gated transport seam is correct, fail-closed, and byte-identical on the Celery path. The comments below are ranked robustness / test-coverage / clarity improvements. Highest-value items: (1) the broad `except` in `_send_clubbed` mislabels permanent dispatch errors as transient broker failures and retries them forever, and (2) the PR's actual behavioral change in `internal_api_views.py` has no test coverage.

muhammad-ali-e:
Addressed all 6 threads in `9f13db367`.

- **P1 · broad `except` mislabels permanent errors** — `_send_clubbed` now branches on class: `except (ValueError, TypeError)` (enqueue_task validation / payload serialization) → **dead-letter now** with a distinct `result=dispatch_error` metric; `except Exception` keeps the transient revert-to-PENDING + attempt refund. So a permanent error no longer re-renders/re-dispatches and emits a mislabeled `broker_failure` traceback every tick until the cap. (On the double-traceback: `_send_clubbed` owns the domain metric+traceback; `enqueue_task`'s log is a producer-level breadcrumb shared by all callers — leaving that as-is rather than touching the shared producer here.)
- **P1 · call site untested** — new `test_send_clubbed.py`: (1) the two-org-identifier contract — asserts `org_string_id == "<string>"` for routing while `kwargs["organization_id"] == <pk>`; (2) recovery — transient → SENDING-guarded revert to PENDING, permanent → DEAD_LETTER; (3) `_org_identifier` — string id, and None→warn. 11 tests pass.
- **P2 · overloaded `organization_id`** — renamed the seam's routing param to **`org_string_id`** (one call-site + test edit) so it can't be confused with the pk in `kwargs["organization_id"]`.
- **P2 · silent None + "org deleted" framing** — `_org_identifier` now logs a warning with `org_pk` on None, reworded to a "data-anomaly guard" (CASCADE FK makes the missing-org state otherwise unreachable), narrowed `org_pk: int`, and fixed the "before the webhook HTTP call" wording.
- **P2 · aspirational Returns comment** — dropped "usable by callers that surface it in an API response"; now "returned for symmetry / any future caller; the sole current caller discards it."
- **P3 · redundant `or None`** — routing arg now passes `org_string_id` straight through (resolve_transport normalizes falsy); kept the load-bearing `or ""` on `enqueue_task`'s `org_id`.

muhammad-ali-e:
Follow-up (85b927cae): gated the dead-letter classing to the **PG path only** so the flag-off (Celery) error flow stays byte-for-byte unchanged. The seam now raises a typed `PermanentDispatchError` only when `enqueue_task` rejects a message permanently (validation / non-serializable payload); `_send_clubbed` dead-letters on that, while any Celery `send_task` failure remains an ordinary `Exception` routed to the transient PENDING branch, exactly as before. +seam test for the wrap; call-site test now raises the real `PermanentDispatchError`. 12 tests pass.

PR #1797, opened 2026-02-17
UN-2202 [FEAT] Add co-owner management for shared resources

kirtimanmishrazipstack:
## What

- Co-owner management (add/remove owners) for Adapters, API Deployments, Connectors, Pipelines, Workflows, and Prompt Studio, via a shared `CoOwnerManagement` widget on each resource page.
- Consolidates the six per-resource `shared_users` M2M tables into one polymorphic `ResourceMembership` table — `OWNER` = creator/co-owner, `VIEWER` = former shared user.
- New `permissions` internals: `HasMembersMixin`, `OwnerManagementMixin` (the `owners/` endpoints), plus membership serializers and permission classes.

## Why

- Only the creator could manage a resource; teams need multiple users with full CRUD access.
- Seven parallel `*_member` through-tables duplicated the same shape and query logic — one table removes that drift.
- `created_by` becomes audit-only, so ownership no longer depends on who first created the resource.

## How

- `ResourceMembership(user, role, content_type, object_id, organization)` with a `GenericRelation("memberships")` on each resource; `save()` auto-derives `organization` so it is safe to read from Celery workers.
- Access querysets union `memberships__user`; owner-locked writes are gated by `IsOwner` and exposed through new `owners/` REST endpoints per resource.
- Per-app data migrations backfill creator→OWNER and `shared_users`→VIEWER through one shared helper, then drop the old M2M (idempotent, `get_or_create`-keyed).

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

- Risk is scoped to the ownership-storage migration and the permission/query swap. The backfill is idempotent and preserves every existing creator (→OWNER) and shared user (→VIEWER); the `memberships` read surface, the two-modal API, and the frontend are unchanged, so no existing access is lost.

## Database Migrations

- `tenant_account_v2/0005_resource_membership` — creates the `resource_membership` table.
- `adapter_processor_v2/0005`, `api_v2/0005`, `connector_v2/0007`, `pipeline_v2/0005`, `prompt_studio_core_v2/0009`, `workflow_v2/0022` — backfill from creator + `shared_users`, then drop the per-resource `*_member` M2M.

## Env Config

- None added in this repo. Co-owner add/remove email notifications are cloud-side and read two SendGrid template IDs from cloud settings (see [unstract-cloud#1295](https://github.com/Zipstack/unstract-cloud/pull/1295)):
[...]

PR #1462, opened 2025-07-30
UN-2649 [FEATURE] Add workflow sharing functionality

johnyrahul:
Implement workflow sharing feature allowing users to share workflows with organization members and specific users.

🤖 Generated with [Claude Code](https://claude.ai/code)

## What

-

## Why

-

## How

-

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

-

## Database Migrations

- 

## Env Config

- 

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

-

## Notes on Testing

-

## Checklist

I have read and understood the [Contribution Guidelines](https://docs.unstract.com/unstract/contributing/unstract/).

PR #2200, opened 2026-07-23
UN-3769 [FEAT] Sortable resource lists with co-owner ownership

kirtimanmishrazipstack:
## What

- New shared `ResourceTable` (Name / Owned By / Created / Modified / Actions) replaces `ListView`/`ViewTools` across Adapters, Workflows, Prompt Studio and Connectors.
- The Owned By column shows the owner avatar/name/email plus co-owner count and opens the co-owner modal.
- Prompt Studio rows show a `Prompts: N` count; both Created and Modified columns are sortable.
- Server-side owner-inclusive search, per-column sort and pagination on all four list endpoints. Lists default to most-recently-modified first (`-modified_at`) so the visible Modified column matches the sort order.

## Why

- List views didn't surface ownership, so UN-2202 co-owners were invisible there.
- Sort/search/pagination were client-only or absent, making large resource lists hard to navigate.

## How

- Builds on #2208's pagination work: the `for_user()` managers use plain `.distinct()` and each viewset is vanilla DRF (`ordering = ["-modified_at", "pk"]` + `ordering_fields`), filtered by the global `DeterministicOrderingFilter`. Sort/search is standard `?ordering=`/`?search=`; Prompt Studio annotates `prompt_count`.
- Frontend `ResourceTable` uses custom sort-dropdown headers; `usePaginatedList` owns sort state and pages fetch `?ordering=`/`?search=` server-side, defaulting to `modified_at` desc.
- Deleted the now-unused `ListView` and `ViewTools` (0 remaining consumers).

- Scoped to resource list views; all queries stay org-scoped and `for_user()`-filtered. Endpoints keep the bare-array response unless `?page`/`?page_size` is sent, so existing API callers are unaffected.

## Database Migrations

- None

## Env Config

- None

## Relevant Docs

-

## Related Issues or PRs

- Follows UN-2202 co-owners (#1797). Cloud counterpart: Zipstack/unstract-cloud#1672 — merge this PR first (the cloud frontend build pulls OSS `main` for the shared `ResourceTable`).

## Dependencies Versions

- None

## Notes on Testing

- Behavior change: resource lists now default to most-recently-modified first (`-modified_at`), instead of `name asc`. This restores the `-modified_at` ordering #2187 gave Workflows. Both Created and Modified columns are sortable; the default order shows as active on the Modified header on load.

## Screenshots

_Add before merge._

## Checklist


frontend/src/hooks/usePaginatedList.js
```diff
@@ -1,44 +1,141 @@
...
+  const results = data?.results ?? data ?? [];
+  const total = data?.count ?? results.length;
+  // Deleting the last row on a page leaves it empty; step back a page.
+  if (results.length === 0 && page > 1 && total > 0) {
+    return refetchPrevPage();
+  }
+  setList(results);
+  setPagination((prev) => ({ ...prev, current: page, pageSize, total }));
+  return undefined;
+}
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @coderabbitai Valid catch — the sequence guard only covered the `.then`. Fixed in b76e2e297: each list page's `.catch` now early-returns when `seq !== seqRef.current`, and `.finally` only clears loading when `seq === seqRef.current`, so a superseded request no longer clears loading or surfaces a stale error while a newer one is still pending.

frontend/src/components/custom-tools/list-of-tools/ListOfTools.jsx
```diff
@@ -107,61 +120,101 @@ function ListOfTools({ segmentOptions, segmentValue, onSegmentChange }) {
...
+          // A newer request superseded this one — don't surface its error.
+          if (seq !== seqRef.current) {
+            return;
+          }
+          setAlertDetails(
+            handleException(err, "Failed to get the list of tools"),
+          );
+          // Avoid an indefinite spinner when the first fetch fails.
+          setDisplayList((prev) => prev ?? []);
+        })
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @coderabbitai Fixed in c435b06cf. All five list pages now track an explicit `loadError` (cleared at each fetch start, set in the seq-gated `catch`) instead of forcing `displayList` to `[]` on failure, and render a retryable error state (Retry refetches the current page) shown only when there is no data. A failed initial load therefore no longer renders the "No … available" empty state; genuine empty states are unchanged, and a failed refetch keeps the existing rows plus the error toast.

frontend/src/components/widgets/resource-table/ResourceTable.jsx
```diff
@@ -0,0 +1,393 @@
...
+            </Typography.Text>
+          )}
+        </div>
+      </div>
+    );
+  };
+
+  const renderOwner = (item) => {
+    const email = item?.[ownerEmailProp];
+    const isMe = item?.is_owner || (email && email === sessionDetails?.email);
```
kirtimanmishrazipstack: **🔴 `is_owner` mislabels co-owned rows as "Me"**

`is_owner` is true for **any** OWNER membership — co-owners included — but this cell always displays the *primary* owner (`created_by_email` here, `owner_email` on the cloud agentic page). So a co-owner opening a resource they didn't create sees the row labeled **"Me"** with the creator's avatar + email rendered right below it — a self-contradicting "Owned By" cell for every co-owned-but-not-created resource, on all four OSS lists and the shared cloud Projects page.

Display-only (the co-owner already sees the roster), but it's wrong in exactly the column this feature adds. The "Me" label should track the *displayed* owner, not the viewer's ownership status:

```js
const isMe = email && email === sessionDetails?.email;
```

Creator-viewing-own stays "Me" via the email match.

kirtimanmishrazipstack: Fixed in f6c016003. `isMe` now keys on the displayed owner email only — `const isMe = Boolean(email) && email === sessionDetails?.email;` — dropping the `is_owner` disjunct. A co-owner viewing a resource they did not create now sees the actual owner, while the creator viewing their own still reads "Me". Since `ResourceTable` is shared, this covers all four OSS lists and the cloud Projects page.

frontend/src/components/tool-settings/tool-settings/ToolSettings.jsx
```diff
@@ -87,95 +101,132 @@ function ToolSettings({ type }) {
...
+      headers: { "X-CSRFToken": sessionDetails?.csrfToken },
+    })
+      .then(() => handleListRefresh())
       .catch((err) => {
         setAlertDetails(handleException(err));
-      })
-      .finally(() => {
+        // Refresh only runs on success; clear loading here so a failed delete
+        // doesn't leave the table stuck under its spinner.
         setIsLoading(false);
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Fixed in db2415152. The delete handler now snapshots the request token at start (`const seq = seqRef.current;`) — a fetch started while the delete is in flight bumps it — and the `catch` clears loading only when `seq === seqRef.current`, so a failed delete no longer hides the spinner for a newer in-flight request. Same seq-guard as the list fetches. ToolSettings is the only delete that sets loading; the other pages' deletes don't, so no race there.

frontend/src/components/tool-settings/tool-settings/ToolSettings.jsx
```diff
@@ -87,95 +101,138 @@ function ToolSettings({ type }) {
...
-    };
+  // New/edited adapters land on some page under the active sort — refetch the
+  // current page to reflect server truth rather than splicing a stale array.
+  const addNewItem = () => handleListRefresh();
+
+  const handleDelete = (_event, adapter) => {
+    // Snapshot the request token: a fetch (search/sort/paginate/refresh) that
+    // starts while this delete is in flight bumps it, so we don't clear its
+    // loading on a failed delete.
+    const seq = seqRef.current;
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Fixed at the root in 52d9d1173. You are right — two overlapping deletes shared the snapshotted `seqRef`, so the first failure cleared loading for both. Rather than add another guard, the delete no longer touches the shared list-loading at all (matching the other four lists): success refetches via `handleListRefresh`, which owns the spinner, and failure just toasts. With deletes out of the shared-loading contention, the stuck / newer-fetch / concurrent-delete edges are all gone by construction.

frontend/src/components/tool-settings/tool-settings/ToolSettings.jsx
```diff
@@ -87,95 +101,129 @@ function ToolSettings({ type }) {
...
+    axiosPrivate({
       method: "DELETE",
       url: `/api/v1/unstract/${sessionDetails?.orgId}/adapter/${adapter?.id}/`,
-      headers: {
-        "X-CSRFToken": sessionDetails?.csrfToken,
-      },
-    };
+      headers: { "X-CSRFToken": sessionDetails?.csrfToken },
+    })
+      .then(() => handleListRefresh())
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Fixed at the root in 5947bcc64. You are right: `handleListRefresh` closed over pagination/search/sort, so a refresh captured in a pending delete's `.then` refetched the stale view and overwrote the user's current one. Rather than patch the delete, `handleListRefresh` is now a stable callback that reads the latest params from a ref, so a captured refresh always targets the view the user is on now. Fix lives in the shared `usePaginatedList` hook, so it covers create/edit/import/delete across all five lists.

frontend/src/components/widgets/resource-table/ResourceTable.jsx
```diff
@@ -0,0 +1,396 @@
...
+              {item[descriptionProp]}
+            </Typography.Text>
+          )}
+        </div>
+      </div>
+    );
+  };
+
+  const renderOwner = (item) => {
+    const email = item?.[ownerEmailProp];
```
kirtimanmishrazipstack, line 184: **High — the "Owned By" column can name someone who is not an owner.**

All four OSS pages pass `ownerEmailProp="created_by_email"` (`Workflows.jsx:442`, `ToolSettings.jsx:401`, `ListOfTools.jsx:495`, `ConnectorsPage.jsx:338`), and `OWNER_SORT_FIELD = "created_by__email"` keys the owner sort off the same column.

Repro: Alice creates workflow W → adds Bob as co-owner → Bob removes Alice. That removal is permitted — `RemoveOwnerSerializer` only requires another *live* owner (`membership_serializers.py:92`), and the co-owner modal renders the remove button once `totalOwners > 1`. Memberships are now `{Bob: OWNER}` but `created_by` is still Alice. The row renders **"alice / alice@x.com"** with no `+N`, to the whole org — and Alice no longer appears in `for_user()` for W at all. `?sort_by=owner` sorts by that same wrong column.

The cloud side already solved this, and its docstring states the rule we're breaking here — `agentic_studio_v1/serializers.py:223`:

> `created_by` is audit-only (UN-2202) and the creator can be removed as owner, so it must not be used to name the owner.

Cloud has `get_owner_email()` (earliest live OWNER, service accounts excluded) and passes `ownerEmailProp="owner_email"`.

Suggested fix: mirror that method onto `AdapterListSerializer` / `CustomToolListSerializer` / `ConnectorInstanceSerializer` / `WorkflowSerializer` and switch the four `ownerEmailProp` values. The `memberships__user` prefetch is already attached in all four `get_queryset`s, so it stays query-free. Pointing `OWNER_SORT_FIELD` at the same roster would also close the sort/display divergence already documented on the cloud PR.

frontend/src/pages/ConnectorsPage.jsx
```diff
@@ -235,30 +299,52 @@ function ConnectorsPage() {
...
-            descriptionProp="connector_type"
-            iconProp="icon"
-            showOwner={true}
-            type="Connector"
-            isEmpty={!listRef.current.length}
-            centered
-            isClickable={false}
-          />
+          {displayList === undefined && !loadError && <SpinnerLoader />}
+          {displayList === undefined && loadError && (
```
kirtimanmishrazipstack: **High — `loadError` is unreachable after the first successful load.**

This gate gets `displayList === undefined`, which is only true before the first response lands. After that, `setLoadError(true)` writes state that **no branch can render**, and the page falls through to a length-based branch that asserts something false.

Same gating in `Workflows.jsx:408`, `ToolSettings.jsx:370`, `ListOfTools.jsx:464`, and cloud `Projects.jsx:320`.

Repro on ToolSettings: open `/settings/llm` with zero adapters (first fetch succeeds → `displayList = []`), type `gpt` in the search → `setSearchTerm("gpt")` commits immediately, the request 500s. The catch sets `loadError` (unrenderable) and the screen renders **"No results found for this search"** — a claim that was never established. The toast is transient; the on-screen state persists and is wrong.

Worth calling out that the commit adding this was titled "Show a retryable error on list-fetch failure" — as gated it only ever fires on cold-start failure, the one case users already handle by refreshing.

Suggested shape — hoist the error above the length branches and negate it on the rest:

```jsx
{loadError && <EmptyState text="Couldn't load. Please try again." btnText="Retry" handleClick={handleListRefresh} />}
{!loadError && displayList === undefined && <SpinnerLoader />}
{!loadError && displayList?.length === 0 && !searchTerm && (/* onboarding */)}
{!loadError && displayList?.length === 0 && searchTerm && (/* no results */)}
{!loadError && displayList?.length > 0 && (/* table */)}
```

frontend/src/components/workflows/workflow/Workflows.jsx
```diff
@@ -115,46 +106,53 @@ function Workflows() {
...
+      )
       .catch(() => {
+        // A newer request superseded this one — don't surface its error.
+        if (seq !== seqRef.current) {
+          return;
+        }
         console.error("Unable to get project list");
-        // Avoid an indefinite spinner when the first fetch fails.
-        setProjectList((prev) => prev ?? []);
+        // Surface a retryable error instead of a misleading empty state.
```
kirtimanmishrazipstack, line 141: **High — this is the only list-fetch catch of the five that reports nothing to the user or to Sentry.**

Line 131 is `.catch(() => {` with no `err` parameter, and line 136 is a bare `console.error("Unable to get project list")` — no error object, no HTTP status, no URL, no page number.

The other four all surface it: `ToolSettings.jsx:170` and `ListOfTools.jsx:187` call `setAlertDetails(handleException(err))`, `ConnectorsPage.jsx:136` calls `setAlertDetails(handleException(err, "Failed to load connectors"))`, cloud `Projects.jsx:141` calls `showError(err, "Failed to load projects")`. `handleException` is already imported here (line 46).

On its own this is minor. Combined with the `loadError` gating issue it means a Workflows list failure is **completely invisible** — no toast, no error state, and the pager silently snapping back to its previous value because `setPagination` never ran. Clicking page 2 on a 500 looks exactly like the app ignoring the click.

```js
.catch((err) => {
  if (seq !== seqRef.current) return;
  setAlertDetails(handleException(err, "Unable to load workflows"));
  setLoadError(true);
  syncRequested();
})
```

frontend/src/hooks/usePaginatedList.js
```diff
@@ -1,44 +1,203 @@
...
+    order: defaultOrder,
+  });
 
-  const fetchRef = useRef(fetchData);
-  fetchRef.current = fetchData;
+  // Page assigns its fetch fn here; handlers call the latest one via the ref.
+  const fetchRef = useRef(null);
+
+  // Mirrors the on-screen view each render (pagination only advances when a
+  // response applies), so a FAILED request can realign requestedRef with what's
```
kirtimanmishrazipstack: **Medium — this comment is false for three of the five fields, and `syncRequested` depends on it being true.**

"Mirrors the on-screen view each render" holds for `page`/`pageSize` — `setPagination` is only ever called from `applyPagedResponse`, so those genuinely advance at apply time. It does **not** hold for `search`/`sortBy`/`order`: `handleSearch` (line 145) and `handleSortChange` (line 153) commit `setSearchTerm`/`setSort` *before* firing the request, and neither rolls back on failure.

So `appliedRef` is a hybrid of last-**applied** pagination and last-**requested** controls, and `syncRequested` copies that hybrid into `requestedRef`.

Consequence — on the two axes users touch most, `syncRequested` writes back the failed request byte-for-byte and its stated purpose ("stops the next handleListRefresh from replaying the failed target") isn't achieved. Click *Owned By → Z-A*, request 500s, `syncRequested()` records `{sortBy: "owner", order: "desc"}` — identical to what just failed. Delete a row: success toast fires, `handleListRefresh()` replays the same failing request, a second error toast appears, and the deleted row is still on screen because `setList` never ran.

Mixed-axis case: on page 3, type `zzz`, request fails → `requestedRef = {page: 3, search: "zzz"}`, a view that was neither requested (page 1) nor displayed (page 3, unsearched). It self-heals via the stepback cascade 3→2→1, but that's three round trips and two list flashes.

Either snapshot `searchTerm`/`sort` before the request and restore them in the page's catch, or record params in the success path where `setList`/`setPagination` actually commit. At minimum the comment needs to stop claiming an invariant the next maintainer will rely on.

frontend/src/hooks/usePaginatedList.js
```diff
@@ -1,44 +1,203 @@
...
+    pageSize: defaultPageSize,
+    search: "",
+    sortBy: defaultSortBy,
+    order: defaultOrder,
+  });
+
+  // Single fetch entry: records the requested view before firing, so every
+  // path (navigation, stepback, reset) keeps requestedRef in sync with what
+  // ends up on screen. Pages route their stepback (refetchPrevPage) through it.
+  const requestList = (page, pageSize, search, sortBy, order) => {
```
kirtimanmishrazipstack, line 130: **Medium — no `return` here, which voids the contract `applyPagedResponse` documents.**

`applyPagedResponse`'s JSDoc (lines 33-34, 44-45) says it steps back "*returning that refetch promise so the caller's `finally` waits for replacement data instead of clearing loading early*", and declares `@return {Promise|undefined}`.

All five callers now pass `refetchPrevPage: () => requestList(page - 1, …)`, and `requestList` discards the value of `fetchRef.current?.(…)`. So `applyPagedResponse` returns `undefined` on every path, the `.then` adopts nothing, and the `Promise` arm of the `@return` tag is unreachable.

Nothing breaks today — but only by accident. The stepback runs `++seqRef.current` **synchronously** inside the `.then`, so the outer `.finally` sees `seq !== seqRef.current` and skips `setLoading(false)`. The seq guard is the entire mechanism; the documented one doesn't exist.

That's the hazard: someone reading this JSDoc reasonably concludes the promise chain holds `loading` and that `if (seq === seqRef.current)` in `.finally` is redundant belt-and-braces. Removing it reintroduces exactly the early-clear bug these commits fixed. The doc also invites `await applyPagedResponse(...)` expecting the stepback data to have landed.

Either `return fetchRef.current?.(page, pageSize, search, sortBy, order);` — one word, and the doc becomes true — or delete both doc sentences and the `@return` tag so the seq guard is documented as the mechanism it actually is.

backend/prompt_studio/prompt_studio_core_v2/views.py
```diff
@@ -161,27 +162,30 @@ def get_queryset(self) -> QuerySet | None:
...
+            # column sorts.
+            qs = apply_search_and_sort(
+                qs,
+                model=CustomTool,
+                name_field="tool_name",
+                request=self.request,
+                select_related=("created_by",),
+                prefetch_related=("memberships__user",),
+            )
+            # The pk__in re-wrap drops annotations, so re-apply prompt_count on
```
kirtimanmishrazipstack: **Medium — both halves of this rationale are stale.** The code is correct; the stated reasons aren't, which will mislead anyone deciding whether the Subquery can be simplified away.

1. "**re-apply** prompt_count" — nothing was annotated before the re-wrap. `qs` at line 161 is `CustomTool.objects.for_user(...).prefetch_related("memberships__user")` only. The annotation is applied here for the first time, not re-applied. (Pre-refactor the `.annotate()` did come first; the comment survived the reordering.)

2. "**Subquery avoids conflict with the `distinct("tool_id")` that `for_user()` carries**" — the queryset being annotated here has no `DISTINCT ON`. `apply_search_and_sort` returns a fresh `CustomTool.objects.filter(pk__in=…)` chain; the `distinct("tool_id")` lives only on the inner `.values("pk")` subquery, which is never annotated. `for_user()` also returns `self.all()` for service accounts and org admins, with no distinct at all.

Something like:

```python
# apply_search_and_sort returns a fresh CustomTool.objects chain, so the
# annotation goes on afterwards. Subquery keeps the count out of the outer
# GROUP BY, which would otherwise collide with the sort column.
```

kirtimanmishrazipstack (commented):
Second-pass review (self-review). Pass-1 findings are both closed — the `is_owner` fix at `ResourceTable.jsx:190` holds and preserves the creator-sees-"Me" case. CI is fully green here.

This pass focused on the 5 commits reworking the list fetch state machine, plus a fresh look at the owner axis. **3 High + 3 Medium**, all verified against the source rather than inferred from the diff.

- **High 1** — the OSS "Owned By" column can name someone who is no longer an owner
- **High 2** — `loadError` is unreachable after the first successful load, on all 5 pages
- **High 3** — the Workflows list-fetch catch reports nothing (compounds High 2 into total silence)
- **Medium 1** — `syncRequested` can't undo a failed search/sort; its comment claims otherwise
- **Medium 2** — `requestList` voids the stepback contract `applyPagedResponse` documents
- **Medium 3** — stale rationale on the `_prompt_count` Subquery

Also noted, not blocking: the `?order_by` param on `/workflow/` is now silently ignored (main honoured it for `modified_at`). No frontend senders remain, so it just needs a line in the breaking-change section rather than a code change.

Separately — the `.distinct()` calls in all five `for_user()` managers are **no-ops** (the sharing helpers return `ValuesQuerySet` subqueries used as `IN`, which can't multiply outer rows; they're leftovers from the pre-`ResourceMembership` join era). That does *not* make the `pk__in` re-wrap wrong — while those `.distinct()` calls stand, Postgres still requires `ORDER BY` to lead with the `DISTINCT ON` expression, so the re-wrap is load-bearing exactly as written. Worth a follow-up ticket that deletes the `.distinct()` calls, the re-wrap, both hint params, and the `_prompt_count` Subquery together.

frontend/src/components/widgets/resource-table/ResourceTable.jsx
kirtimanmishrazipstack: Fixed in db133f5b5. Added `owner_email()` to `HasMembersMixin` (earliest live OWNER, service accounts excluded — mirrors the cloud serializer), emitted it on all four list serializers, and switched the four `ownerEmailProp` values to `owner_email`. Query-free since the `memberships__user` prefetch is already attached in every `get_queryset`. Left `OWNER_SORT_FIELD` on `created_by__email` — the sort/display divergence stays the documented follow-up, same as cloud.

frontend/src/pages/ConnectorsPage.jsx
kirtimanmishrazipstack: Fixed in db133f5b5 (OSS) and a600eed3 (cloud). Hoisted the `loadError` branch ahead of the length branches and gated the spinner/length branches on `!loadError` across all four OSS pages and cloud `Projects.jsx`, so a failure after the first successful load shows the retry state instead of a false "No results" claim.

frontend/src/components/workflows/workflow/Workflows.jsx
kirtimanmishrazipstack: Fixed in db133f5b5. The catch now takes `err` and calls `setAlertDetails(handleException(err, "Unable to load workflows"))`, matching the other four pages; it keeps `setLoadError(true)` and `syncRequested()`.

frontend/src/hooks/usePaginatedList.js
kirtimanmishrazipstack: Fixed in db133f5b5 — comment only. Rewrote it to say page/pageSize mirror the applied view while search/sortBy/order are the last-requested values (committed before the request fires, not rolled back on failure). The behavior self-heals, so a snapshot/rollback redesign is left as a follow-up rather than expanding this PR; the comment no longer claims an invariant that does not hold.

frontend/src/hooks/usePaginatedList.js
kirtimanmishrazipstack: Fixed in db133f5b5. `requestList` now returns `fetchRef.current?.(...)`, so a stepback promise propagates through `applyPagedResponse` as its JSDoc documents. The seq guard remains the actual loading-hold mechanism (belt-and-braces, not redundant).

backend/prompt_studio/prompt_studio_core_v2/views.py
kirtimanmishrazipstack: Fixed in db133f5b5. Rewrote the comment: `apply_search_and_sort` returns a fresh `CustomTool.objects` chain, so the `prompt_count` annotation is applied afterwards (not re-applied), and the Subquery keeps the count out of the outer GROUP BY — not a `DISTINCT ON` conflict, which the re-wrapped queryset does not carry.

chandrasekharan-zipstack:
Reviewed this alongside #2208, which overlaps heavily. Summary plus a suggested path.

**tl;dr** — the feature here (owner column, unified `ResourceTable`) is the right direction. The backend plumbing underneath it is solving a problem #2208 removes, so this gets considerably smaller stacked on top of that PR.

### Why stack

Both PRs hit the same wall: `for_user()` ends in `DISTINCT ON`, so Postgres pins `ORDER BY` to the distinct expression and neither `adapter_name` nor `modified_at` is sortable.

- This PR routes around it — `apply_search_and_sort()` re-wraps every queryset as `model.objects.filter(pk__in=qs.order_by("pk").values("pk"))` so the outer query is free to sort. Correct, but it costs an extra subquery on every list *and* detail call, and needs a bespoke `?sort_by`/`?order` vocabulary limited to three columns.
- #2208 deletes the wall — `.distinct("id")` / `.distinct("tool_id")` become plain `.distinct()` on the three managers. Every arm of the sharing predicate is a `pk__in` subquery rather than a join, so there are no duplicate rows to collapse (Workflow has shipped on plain `.distinct()` since #1462). Once it's gone the viewsets are vanilla DRF: `ordering = ["-modified_at", "pk"]`, `ordering_fields = [...]`, standard `?ordering=`.

Stacked, `backend/utils/list_query.py` goes away entirely and the sort headers move to `?ordering=` — which also unlocks `modified_at`, something the three-key vocabulary can't express today.

### Suggested order

1. #2208 as the base
2. this PR rebased onto it
3. cloud: Zipstack/unstract-cloud#1673, then Zipstack/unstract-cloud#1672

Merging the two OSS PRs back-to-back rather than separately, so `main` never sits in an intermediate state where the lists have pagers but not the new table.

### What the rebase looks like

Mostly "take this PR's version":

| File | Resolution |
|---|---|
| 4 × `views.py` | take #2208's; drop the `apply_search_and_sort` import + call, keep `ordering_fields` |
| `ListOfTools` / `ToolSettings` / `ConnectorsPage` / `Workflows.jsx` | take this PR's — #2208's edits are superseded |
| selector files (`AdapterSelectionModal`, `AddLlmProfile`, `DefaultTriad`, `CombinedOutput`, `ConfigureConnectorModal`, `EtlTaskDeploy`, `CustomToolsHelper`) | no conflict — this PR doesn't touch them; keep #2208's `fetchAllPages` |
| hooks | keep this PR's `usePaginatedList` (it's the superset); drop #2208's `usePaginatedResource`, port its tests onto the survivor |
| `ViewTools` / `ListView` | this PR's deletion wins |
| `useListSearch.js` | take #2208's deletion — this PR orphans it (61 dead lines) |
| `backend/utils/tests/test_list_pagination.py` | keeps passing, `?ordering=` is unchanged |

`HasMembersMixin.owner_email()` and the four serializer fields stay exactly as they are — they're independent of all of the above and are what the Owned By column needs.

### Review points

1. **Owner sort/search is keyed on the wrong field.** `OWNER_SORT_FIELD = "created_by__email"`, but the column renders `owner_email()` — the earliest live OWNER membership. `owner_email()`'s own comment says `created_by` "is audit-only … must not name the owner". So "Owned By → A-Z" sorts by a column that isn't on screen, and `?search=alice@` misses everything Alice owns but didn't create. unstract-cloud#1672 documents this as accepted; worth fixing properly with a Subquery annotation of the earliest OWNER email, added to `ordering_fields`.

2. **`prompt_count` is no longer displayed anywhere.** The old `ListView` tooltip showed `Prompts: N` (and `Model:`, `Modified:`). `ResourceTable` has no equivalent, but the backend still runs the `_prompt_count` Subquery per row and serializes it. Either surface it in the table or drop the annotation — right now it's dead payload and a visible regression for the Prompt Studio list.

3. **Default sort should be descending / most-recent-first.** No page sets `defaultSortBy`, so the backend default `name asc` applies. Workflows loses the `-modified_at` ordering #2187 gave it, and the `EtlTaskDeploy` dropdown changes too. Related: the table shows a **Created Date** column, so defaulting to `-modified_at` would order by a column that isn't visible — either add a Last Modified column or default to `-created_at`.

4. **No tests.** `list_query.py` is 75 lines of DISTINCT-ON-sensitive query rewriting with no coverage. If it survives the rebase it needs some; if it doesn't, `backend/utils/tests/test_list_pagination.py` from #2208 already covers the replacement.

5. **The helper runs on every action, not just `list`.** Prompt Studio correctly gates on `self.action == "list"`; adapters, connectors and workflows don't — so `retrieve`/`update`/`destroy` and the hot `?adapter_type=LLM` selector fetch all pay the `pk__in` re-wrap plus the `memberships__user` prefetch.

6. **`owner_email()` off the list path.** `is_owner()`, `co_owners_count()` and `owner_email()` each evaluate `self.memberships.all()` separately, so detail/create/update responses now do three queryset evaluations per instance instead of two. Worth caching on the instance.

7. **`default_sort_by` kwarg is unused** by all five call sites.

8. **Pager alignment and nested scroll.** `.workflows-pg-body` is `align-items: center`, which shrink-wraps `.workflows-pagination` and neutralises its `justify-content: flex-end` — that's why the pager reads as centred rather than right-aligned on staging. This PR doesn't touch `Workflows.css`, so that rule ends up orphaned (#2208 deletes it). Worth sorting out while the list markup is being replaced, along with the inner `overflow-y: auto` that makes even a 10-row page scroll.

Verified separately, both safe: `ListView`/`ViewTools` have zero remaining importers in OSS *and* cloud, and the `usePaginatedList` API change (`fetchData` → `fetchRef`) has no unmigrated consumer in either repo.

chandrasekharan-zipstack:
Re-reviewed at current HEAD. The rebase onto `feat/list-pagination-consistency` landed cleanly and most of my earlier points are addressed — prompt count is back, Created and Modified are both columns, all four pages default to `modified_at` desc matching the viewset `ordering`, `list_query.py` is gone in favour of plain `?ordering=`, and the pager is antd's own so the alignment problem is fixed. Prefetching is correct too: `memberships__user` is on all four viewsets, so `owner_email()` costs no extra queries.

Remaining concerns, roughly in order:

**1. Search matches the creator, the column shows the owner.** All four viewsets now filter on `Q(<name>__icontains) | Q(created_by__email__icontains)`, but the Owned By cell renders `owner_email()` — the earliest live OWNER membership. The docstring on that method says `created_by` is audit-only and "must not name the owner", which is exactly right, and that makes the search predicate inconsistent with it. Searching someone's email returns rows whose Owned By shows a different person, and misses rows they actually own. It only diverges once ownership is transferred or the creator is removed as owner — which is the case UN-2202 exists for.

**2. `ordering_fields` exposes `created_by__email`, but nothing sorts by owner.** The Owned By header is a static `<span>`, not a `SortHeader`. So this is API surface that no UI reaches and that sorts by the wrong field if hit directly. Either drop it, or annotate the real owner so sort and search share one definition:

```python
owner_sq = (
    ResourceMembership.objects.filter(
        object_id=OuterRef("pk"),
        content_type=ct,
        role=ResourceRole.OWNER,
        user__is_service_account=False,
    )
    .order_by("created_at")
    .values("user__email")[:1]
)
```

`annotate(owner_email_sort=Subquery(owner_sq))` then covers both this and (1).

**3. No tests.** Nothing added or touched across either PR. `owner_email()` has real branching — service-account filter, earliest-by-`created_at` tiebreak, the no-owner `None` case — and the widened search and new `ordering_fields` entries are all unpinned. `backend/utils/tests/test_list_pagination.py` is already set up for the four endpoints and would take these cheaply.

**4. Displayed sort and requested sort disagree.** Two symptoms, one cause:

- Each page mounts with `getListOfTools(1, DEFAULT_PAGE_SIZE, "", "", "asc")` — no `ordering` sent — while hook state is `{sortBy: "modified_at", order: "desc"}`. So the header paints Modified as active-descending for a request that never asked for it. It also bypasses `requestList`, so `requestedRef` keeps its constructor defaults and a later `handleListRefresh` replays `-modified_at` when mount sent nothing. Rows match today only because the viewset default happens to agree.
- "Clear Sort" sets `sortBy` to `""`, which drops `ordering` and falls back to the same `-modified_at` — the header un-highlights but nothing reorders.

Both go away if "no sort" stops being a state. The viewset always applies an ordering, so clearing should restore the default rather than clear it:

```js
const nextSort = sortBy
  ? { sortBy, order: order || "asc" }
  : { sortBy: defaultSortBy, order: defaultOrder };
```

and mount should route through `requestList(1, DEFAULT_PAGE_SIZE, "", sort.sortBy, sort.order)` — `sort` is already the defaults there, so no duplicated literals.

**Minor**

- `Workflows.css` is untouched, so `.listWrapper { height: 92%; overflow: hidden auto }` still nests inside `.list-of-workflows-body { overflow-y: auto }`. The pager now sits at the bottom of the table inside that inner scroll box rather than pinned, and the double scrollbar is still there.
- `useListSearch.js` is orphaned again — no importers left.
- In `renderOwner`, the `initials` ternary is dead: `isMe ? email : name` both slice to the same two characters.
- `getListOfTools`' `useCallback` deps omit `requestList` / `syncRequested`. Harmless since both only touch refs, but exhaustive-deps will flag it.

**On comments** — could you trim the new ones to be shorter and less situational? Several read as design notes and reference things that will drift. A few examples:

- The `modified_at needs no annotation: ...` block in `prompt_studio_core_v2/views.py` names specific call sites (`ToolStudioPrompt.save/delete`, `sync_prompts`) and then instructs future authors — that goes stale the first time one of them moves.
- `owner_email()`'s comment explains the ticket rationale, the prefetch, and the tiebreak in one paragraph.
- `usePaginatedList.js` has several multi-sentence blocks around `appliedRef` / `requestedRef` that restate what the code does, plus one that cross-references another function's JSDoc.

A single line on *why* is usually enough, and it survives refactors better than a paragraph on *how*. Same for the JSDoc — worth keeping where a param is non-obvious, but a lot of it currently narrates the implementation.

Nothing above blocks the stack. (1) and (3) are the two I'd want resolved before merge.

kirtimanmishrazipstack:
@chandrasekharan-zipstack thanks, went through all of it. Pushed `8401a95ce` here (cloud mirror in Zipstack/unstract-cloud#1672 at `59e03e0e`).

**1 + 2 — Owned By is now display-only.** Dropped `created_by__email` from both the search `Q` and `ordering_fields` across all four viewsets, so search matches the owner shown in the column (name-only) and there's no dead owner-sort surface left. I went with the deletion over the annotated-owner subquery: we already made the header static last round so owner sort isn't coming back, and the old owner search keyed on the creator anyway — it was never searching the displayed owner.

**4 — sort-state fixed.** `handleSortChange` now restores the default ordering on Clear Sort instead of an empty one, and every list mount requests the seeded sort via `requestList(...)` rather than the hardcoded `"","asc"`. So the header and rows agree on first load, and Clear Sort actually re-sorts to default instead of un-highlighting a no-op.

**Minors**
- `.listWrapper` double-scroll — at HEAD that rule is unreferenced: the table renders directly in `.workflows-pg-body` (same structure as `ConnectorsPage`, which is verified working), so there's no inner scroll box around the table and no double scrollbar. Removed the dead rule.
- `useListSearch.js` — deleted (0 importers either repo).
- dead `initials` ternary — simplified to `(email || name)`.
- `getListOfTools` useCallback deps — left as-is; `requestList`/`syncRequested` only touch refs, so pulling them into deps just re-creates the fetch fn every render for no behavior change.

**Comments** — trimmed the ones you flagged (the `modified_at` block, `owner_email()`, and the `appliedRef`/`requestedRef` + cross-referencing blocks in `usePaginatedList`) down to a one-line why.

chandrasekharan-zipstack (approved):
Re-reviewed at `8401a95c`. All the points from my last pass are addressed: owner search/sort divergence removed by scoping search to name-only and dropping `created_by__email` from `ordering_fields` across the four viewsets, clear-sort now restores the default ordering, list mounts request the seeded sort so header and rows agree, dead code (`useListSearch.js`, the initials branch, the unreferenced `.listWrapper` rule) gone, and the new comments trimmed. CI is green.

One thing left open, not blocking: there are still no tests for `owner_email()` (service-account filter, earliest-OWNER tiebreak, the `None` case) or the new `ordering_fields`. `backend/utils/tests/test_list_pagination.py` already covers the four endpoints and would take these cheaply — worth a follow-up.

Also noting search no longer matches owners at all. Consistent, and the right call for now, but if UN-3769 wanted owner search it's deferred with nothing tracking it.

Approving — merging into `feat/list-pagination-consistency` so the stack goes to main as one change.

kirtimanmishrazipstack:
@chandrasekharan-zipstack #3 (tests) is done — added in `98062a12a` to `utils/tests/test_list_pagination.py` (parametrized over all four endpoints):

- `test_search_matches_name_not_owner_email` — an owner's email substring no longer returns their rows; the name still does.
- `test_dropped_owner_ordering_field_is_ignored` — `?ordering=created_by__email` is ignored and falls back to `-modified_at, pk` (dropped field is API-safe, not a 400); with one creator, had it survived the rows would be pk-ordered.
- `test_owner_email_is_earliest_live_owner` — earliest live OWNER, skips service accounts, `None` with no owner (shared mixin, pinned once).

All pass locally: `7 passed, 24 subtests` (`uv run --env-file .env pytest utils/tests/test_list_pagination.py --reuse-db`). Cloud counterpart committed on #1672.

PR #1618, opened 2025-10-30
UN-2930 [MISC] Add plugin infrastructure to unstract-core

chandrasekharan-zipstack:
## NOTE
- This PR has the below stacked changes which have been reviewed and merged to this already
#1620 
#1626 

## What

- Introduce a generic, framework-agnostic plugin system for Django, Flask, and Workers
- Add core plugin infrastructure with discovery, validation, and loading capabilities
- Support for both singleton and non-singleton plugin patterns
- Redis client enhancement to support REDIS_USERNAME environment variable
- Support to add plugins to `unstract-core` as well
- MINOR: Added debupgy dependency to worker to allow debugging

## Why

- UN-2930 requires a flexible plugin architecture to separate OSS and cloud-specific functionality
- Need a standardized way to load and manage plugins across different frameworks (Django, Flask, Workers)
- Enable dynamic feature loading without hardcoding cloud-specific logic in OSS codebase

## How

**Core Plugin Infrastructure:**
- `unstract/core/src/unstract/core/plugins/plugin_manager.py` - Generic PluginManager with:
  - Plugin discovery from directories
  - Metadata validation (version, dependencies)
  - Support for `.so` compiled extensions
  - Framework-specific registration callbacks

**Framework-Specific Wrappers:**
- `unstract/core/src/unstract/core/django/plugin.py` - DjangoPluginManager wrapper
  - Singleton pattern for Django apps
  - Django AppConfig integration
  - Convenience `plugin_loader()` function

- `unstract/core/src/unstract/core/flask/plugin.py` - FlaskPluginManager wrapper
  - Flask blueprint registration support
  - Flask app.logger integration
  - Singleton pattern for Flask apps

**Supporting Changes:**
- Updated `.gitignore` to exclude plugin directories that will contain cloud-specific code
- Added `debugpy>=1.8.14` to workers dev dependencies for debugging support
- Enhanced Redis client to accept `REDIS_USERNAME` env var (alternative to `REDIS_USER`)

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

**No, this PR will not break any existing features.**

**Reasoning:**
1. **Pure additive changes** - Only adds new plugin infrastructure, doesn't modify existing code paths
2. **No existing code depends on this** - This is net-new functionality
3. **Backward compatible** - All changes are additions, no deletions or modifications to existing APIs
4. **Isolated changes** - Plugin system is opt-in and doesn't affect services that don't use it
[...]

PR #2201, opened 2026-07-23
[FIX] Propagate all *_cls/*_class plugin metadata fields in PluginManager

jaseemjaskp:
## What

- `PluginManager.load_plugins()` (in `unstract/core`) only copied a hardcoded 3-field allowlist (`entrypoint_cls`, `exception_cls`, `service_class`) from a plugin's `metadata` dict into the registered plugin dict.
- Generalized this to copy any metadata key ending in `_cls` or `_class`, so new plugin fields don't require this shared loader to be updated every time.

## Why

- The `verticals_usage` plugin (enterprise/cloud-only, feeds API deployment usage into the API Hub subscription dashboard) defines a `headers_cache_class` metadata field that the loader's hardcoded allowlist didn't know about. It was silently dropped, so every `plugin["headers_cache_class"]` lookup in `backend/plugins/workflow_manager/workflow_v2/api_hub_usage_utils.py` raised `KeyError` — swallowed by broad `except Exception` blocks.
- Net effect: API Hub request headers were never cached in Redis, `store_usage()` (the only path that writes to the `verticals.subscription_usage` table) was never reached, and the API Hub portal's subscription dashboard stopped showing usage — with no visible errors anywhere.
- Traced the regression to PR #1736 ("refactor: Add dynamic plugin loading for enterprise components"), which switched `api_hub_usage_utils.py` from a build-time file-replacement model to a runtime `get_plugin()` dict lookup, without the shared `PluginManager` (added in #1618) being updated to support the `headers_cache_class` field.

## How

- Replaced the three explicit `if "x" in metadata: plugin_data["x"] = metadata["x"]` checks in `unstract/core/src/unstract/core/plugins/plugin_manager.py` with a loop that copies any metadata key matching `*_cls`/`*_class`.
- Added `unstract/core/tests/test_plugin_manager.py` covering: unknown `_class`/`_cls` fields (e.g. `headers_cache_class`) now propagate, existing `entrypoint_cls`/`exception_cls` still work, non-class metadata isn't duplicated at the top level, and disabled plugins are still skipped.

- No. Verified via grep across `backend/` that no other plugin's metadata relies on a field outside the previous three-field allowlist (only `service_class` is used elsewhere — notifications, hubspot, subscription_usage, file_converter, payload_modifier, etc.), so existing plugin behavior is unchanged; this purely adds support for previously-dropped fields.

## Database Migrations

- None

## Env Config

- None

## Relevant Docs

- None

## Related Issues or PRs

- Regression introduced in #1736

## Dependencies Versions

- None

## Notes on Testing

- Added `unstract/core/tests/test_plugin_manager.py` (4 new tests, all passing).
- Ran full `unstract/core` test suite (`uv run pytest unstract/core/tests/`) — no new regressions; 2 pre-existing unrelated failures were already broken before this change.
- Ran `ruff check` / `ruff format --check` on changed files — clean.

## Checklist

### DIFF
--- a/unstract/core/src/unstract/core/plugins/plugin_manager.py
+++ b/unstract/core/src/unstract/core/plugins/plugin_manager.py
@@ -215,13 +215,13 @@ class PluginManager:
                     "metadata": metadata,
                 }
 
-                # Add optional fields if present
-                if "entrypoint_cls" in metadata:
-                    plugin_data["entrypoint_cls"] = metadata["entrypoint_cls"]
-                if "exception_cls" in metadata:
-                    plugin_data["exception_cls"] = metadata["exception_cls"]
-                if "service_class" in metadata:
-                    plugin_data["service_class"] = metadata["service_class"]
+                # Add optional class/entrypoint fields if present (any key ending in
+                # _cls or _class, e.g. entrypoint_cls, exception_cls, service_class,
+                # headers_cache_class) so new plugin metadata fields don't need this
+                # loader updated every time.
+                for key, value in metadata.items():
+                    if key.endswith(("_cls", "_class")) and key not in plugin_data:
+                        plugin_data[key] = value
 
                 self.plugins[plugin_name] = plugin_data
 
--- a/unstract/core/src/unstract/core/plugins/plugin_manager.py
+++ b/unstract/core/src/unstract/core/plugins/plugin_manager.py
@@ -11,6 +11,11 @@ from collections.abc import Callable
 from pathlib import Path
 from typing import Any
 
+# Metadata keys ending in either suffix are treated as class/entrypoint handles
+# and promoted onto the registered plugin dict (e.g. entrypoint_cls,
+# exception_cls, service_class).
+_CLASS_HANDLE_SUFFIXES = ("_cls", "_class")
+
 
 class PluginManager:
     """Generic plugin manager for loading and managing application plugins.
@@ -215,12 +220,15 @@ class PluginManager:
                     "metadata": metadata,
                 }
 
-                # Add optional class/entrypoint fields if present (any key ending in
-                # _cls or _class, e.g. entrypoint_cls, exception_cls, service_class,
-                # headers_cache_class) so new plugin metadata fields don't need this
-                # loader updated every time.
+                # Add optional class/entrypoint fields if present, so new plugin
+                # metadata fields don't need this loader updated every time.
                 for key, value in metadata.items():
-                    if key.endswith(("_cls", "_class")) and key not in plugin_data:
+                    if not isinstance(key, str):
+                        continue
+                    # `key not in plugin_data` is defensive: none of the fixed
+                    # keys above (version/module/metadata) end in a class-handle
+                    # suffix, so this never actually excludes anything today.
+                    if key.endswith(_CLASS_HANDLE_SUFFIXES) and key not in plugin_data:
                         plugin_data[key] = value
 
                 self.plugins[plugin_name] = plugin_data
--- a/unstract/core/tests/test_plugin_manager.py
+++ b/unstract/core/tests/test_plugin_manager.py
@@ -93,3 +93,31 @@ def test_disabled_plugin_is_not_loaded(plugins_pkg_dir):
     manager = _load(plugins_pkg_dir)
 
     assert not manager.has_plugin("sample_plugin")
+
+
+def test_inactive_plugin_is_not_loaded(plugins_pkg_dir):
+    _write_plugin(
+        plugins_pkg_dir,
+        "sample_plugin",
+        '{"name": "sample_plugin", "is_active": False}',
+    )
+
+    manager = _load(plugins_pkg_dir)
+
+    assert not manager.has_plugin("sample_plugin")
+
+
+def test_class_field_propagates_regardless_of_value(plugins_pkg_dir):
+    """A *_class/*_cls field is propagated by name, not filtered by its value -
+    a plugin can legitimately set one to None before it's populated at runtime.
+    """
+    _write_plugin(
+        plugins_pkg_dir,
+        "sample_plugin",
+        '{"name": "sample_plugin", "headers_cache_class": None}',
+    )
+
+    plugin = _load(plugins_pkg_dir).get_plugin("sample_plugin")
+
+    assert "headers_cache_class" in plugin
+    assert plugin["headers_cache_class"] is None
--- a/unstract/core/tests/test_plugin_manager.py
+++ b/unstract/core/tests/test_plugin_manager.py
@@ -1,0 +1,95 @@
+import importlib
+import sys
+
+import pytest
+from unstract.core.plugins.plugin_manager import PluginManager
+
+
+def _write_plugin(plugins_pkg_dir, plugin_name, metadata_body):
+    plugin_dir = plugins_pkg_dir / plugin_name
+    plugin_dir.mkdir()
+    (plugin_dir / "__init__.py").write_text(f"metadata = {metadata_body}\n")
+
+
+@pytest.fixture
+def plugins_pkg_dir(tmp_path, monkeypatch):
+    """Create an importable package directory to hold fake plugins."""
+    pkg_dir = tmp_path / "fake_plugins_pkg"
+    pkg_dir.mkdir()
+    (pkg_dir / "__init__.py").write_text("")
+
+    monkeypatch.syspath_prepend(str(tmp_path))
+    importlib.invalidate_caches()
+
+    yield pkg_dir
+
+    sys.modules.pop("fake_plugins_pkg", None)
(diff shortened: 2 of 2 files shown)


unstract/core/src/unstract/core/plugins/plugin_manager.py
```diff
@@ -215,13 +215,13 @@ def load_plugins(self) -> None:
...
-                    plugin_data["entrypoint_cls"] = metadata["entrypoint_cls"]
-                if "exception_cls" in metadata:
-                    plugin_data["exception_cls"] = metadata["exception_cls"]
-                if "service_class" in metadata:
-                    plugin_data["service_class"] = metadata["service_class"]
+                # Add optional class/entrypoint fields if present (any key ending in
+                # _cls or _class, e.g. entrypoint_cls, exception_cls, service_class,
+                # headers_cache_class) so new plugin metadata fields don't need this
+                # loader updated every time.
+                for key, value in metadata.items():
```
jaseemjaskp, line 225: **[Silent failure — Medium]** This loop propagates class handles *by name only* — it copies whatever keys end in `_cls`/`_class`, with no notion of a *required* field. That fixes the `headers_cache_class` drop, but the mechanism that made it invisible is untouched: if a plugin declares a misnamed field (`header_cache_class`, `headers_cache_clss`), the correct key is simply absent and `verticals_usage_plugin["headers_cache_class"]()` in `backend/plugins/workflow_manager/workflow_v2/api_hub_usage_utils.py:38,99` raises `KeyError` again — swallowed by the same broad `except Exception` (lines 51/74/101) this PR was written to defeat.

Consider having consumers read with `.get()` and log an explicit failure, or validating a plugin's required class fields at load time, so a typo errors loudly at startup instead of failing silently per-request. (Narrowing those `except Exception` blocks to distinguish a permanent misconfiguration from a transient runtime error is the more complete fix, but is outside this PR's changed lines.)

unstract/core/src/unstract/core/plugins/plugin_manager.py
```diff
@@ -215,13 +215,13 @@ def load_plugins(self) -> None:
...
                 }
 
-                # Add optional fields if present
-                if "entrypoint_cls" in metadata:
-                    plugin_data["entrypoint_cls"] = metadata["entrypoint_cls"]
-                if "exception_cls" in metadata:
-                    plugin_data["exception_cls"] = metadata["exception_cls"]
-                if "service_class" in metadata:
-                    plugin_data["service_class"] = metadata["service_class"]
+                # Add optional class/entrypoint fields if present (any key ending in
```
jaseemjaskp: **[Maintainability — Low]** The `_cls`/`_class` suffix is now load-bearing behavior but is documented only in this comment. Consider promoting it to a module-level constant (e.g. `_CLASS_HANDLE_SUFFIXES = ("_cls", "_class")`) so the contract is greppable and testable. Also consider dropping the `headers_cache_class` example from the comment — it names a single enterprise plugin's field, so it is the part most likely to rot; the commit message already records the motivating case.

unstract/core/src/unstract/core/plugins/plugin_manager.py
```diff
@@ -215,13 +215,13 @@ def load_plugins(self) -> None:
...
+                # _cls or _class, e.g. entrypoint_cls, exception_cls, service_class,
+                # headers_cache_class) so new plugin metadata fields don't need this
+                # loader updated every time.
+                for key, value in metadata.items():
+                    if key.endswith(("_cls", "_class")) and key not in plugin_data:
```
jaseemjaskp: **[Nit — Low]** `key not in plugin_data` can never be false here: at this point `plugin_data` holds only `version`/`module`/`metadata` (none end in `_cls`/`_class`) and metadata keys are already unique, so the guard is effectively dead. Fine to keep as future-proofing against a reserved key that someday ends in `_class`, but if so, add a one-line note saying it's purely defensive — otherwise it reads as load-bearing when it isn't. Flagged independently by the code-review, simplifier, and test passes.

unstract/core/tests/test_plugin_manager.py
```diff
@@ -0,0 +1,95 @@
...
+    _write_plugin(
+        plugins_pkg_dir,
+        "sample_plugin",
+        '{"name": "sample_plugin", "service_class": str, "headers_cache_class": dict, }',
+    )
+
+    plugin = _load(plugins_pkg_dir).get_plugin("sample_plugin")
+
+    assert plugin["service_class"] is str
+    assert plugin["headers_cache_class"] is dict
```
jaseemjaskp, line 56: **[Test gap — High value]** The real behavioral widening in this change is that the loop propagates a `*_cls`/`*_class` key *regardless of its value*, whereas the old code copied 3 known class fields. No test pins that contract. A future "only propagate real classes" `isinstance`/callable check would pass all 4 tests here while silently reverting the fix for a field whose value is `None`/lazy at load time.

Suggest adding a test with metadata `{"name": "sample_plugin", "headers_cache_class": None}` asserting `plugin["headers_cache_class"] is None` — this locks "propagate by name, not by value."

unstract/core/tests/test_plugin_manager.py
```diff
@@ -0,0 +1,95 @@
...
+def test_disabled_plugin_is_not_loaded(plugins_pkg_dir):
+    _write_plugin(
+        plugins_pkg_dir,
+        "sample_plugin",
+        '{"name": "sample_plugin", "disable": True}',
+    )
+
+    manager = _load(plugins_pkg_dir)
+
+    assert not manager.has_plugin("sample_plugin")
```
jaseemjaskp, line 95: **[Test gap — Medium]** The skip condition at `plugin_manager.py:202` is `metadata.get("disable", False) or not metadata.get("is_active", True)`. This test exercises only the `disable=True` operand; the `is_active=False` branch is separate and could regress independently (e.g. someone flips the default or drops the operand). Suggest mirroring this test with metadata `{"name": "sample_plugin", "is_active": False}` and asserting `not manager.has_plugin("sample_plugin")`.

jaseemjaskp (commented):
Automated PR review (PR Review Toolkit: code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier). The core fix is correct and well-scoped; these are refinements. The non-string-key `AttributeError` case is already covered by CodeRabbit's comment and is not repeated here.

unstract/core/src/unstract/core/plugins/plugin_manager.py
```diff
@@ -215,13 +215,13 @@ def load_plugins(self) -> None:
...
+                        plugin_data[key] = value
```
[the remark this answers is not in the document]
jaseemjaskp: Fixed in f0ccfc8ce — added an `isinstance(key, str)` guard before the suffix check, so a non-string metadata key is skipped instead of raising an uncaught `AttributeError` that would otherwise crash `load_plugins()` for every plugin.

unstract/core/src/unstract/core/plugins/plugin_manager.py
jaseemjaskp: Leaving this as a follow-up rather than fixing here — a real gap, but the proper fix (validating required class fields at load time, or narrowing the `except Exception` blocks in `api_hub_usage_utils.py` to distinguish a misconfigured plugin from a transient error) touches consumer files outside this PR's diff. Noted in the PR description.

unstract/core/src/unstract/core/plugins/plugin_manager.py
jaseemjaskp: Fixed in f0ccfc8ce — hoisted the suffix tuple to a module-level `_CLASS_HANDLE_SUFFIXES` constant and trimmed the comment so it no longer calls out `headers_cache_class` specifically.

unstract/core/src/unstract/core/plugins/plugin_manager.py
jaseemjaskp: Fixed in f0ccfc8ce — added a one-line comment noting `key not in plugin_data` is defensive/future-proofing and not currently reachable.

unstract/core/tests/test_plugin_manager.py
jaseemjaskp: Added `test_class_field_propagates_regardless_of_value` in f0ccfc8ce, which asserts `headers_cache_class: None` still gets propagated onto the plugin dict.

unstract/core/tests/test_plugin_manager.py
jaseemjaskp: Added `test_inactive_plugin_is_not_loaded` in f0ccfc8ce, mirroring the existing `disable=True` test for the `is_active=False` branch.

chandrasekharan-zipstack (approved):

athul-rs (approved):

PR #2202, opened 2026-07-23
UN-3636 [FIX] Two OSS test fixes blocking the cloud rig integration tier

chandrasekharan-zipstack:
Two OSS test issues that block the cloud test-rig integration tier (which consumes OSS `main`). Independent commits — splittable if preferred.

## 1. Leaked `sys.modules` stub (`test_execution_serializer`)

The test stubbed `workflow_manager.workflow_v2.models` in `sys.modules` at import time and never restored it. Under pytest-xdist a worker process is shared across many tests, so the fake `WorkflowExecution` (no real `Workflow`) shadowed the real module for **every later test on that worker** → `cannot import name 'Workflow'` via the URLconf. Latent until the rig ran real multi-worker parallelism, then deterministic (~30 failures/run).

**Fix:** delete the stub — the suite runs under `pytest-django` (app registry loaded), so the real serializer imports directly. Deleted rather than restored-in-`finally` so the footgun can't be copy-pasted; it was the only `sys.modules`-mutating test in the backend.

## 2. `None` DB value serialized as the string `"None"` (`get_sql_values_for_query`)

Fell through to `f"{value}"` for `None`, writing the literal `"None"` — invalid JSON for a `jsonb` column (e.g. `data` on an errored record). `test_insert_into_db_with_error_postgresql` was marked `xfail(strict)` for exactly this; the bug is env-fragile (reproduces on some Postgres builds, XPASSes on others), which the cloud rig surfaced as a hard `XPASS(strict)` failure.

**Fix:** bind `None` as SQL NULL (deterministic across environments); drop the obsolete strict xfail + unused `pytest` import.

## Verification
- serializer test, unit tier: **9 passed**
- full integration tier, `-n auto` (real rig config), ×2: **0 `Workflow` import errors, 708 passed**
- `get_sql_values_for_query({"data": None, ...})` → `data` is SQL `NULL`, other values unchanged
- cloud rig integration tier now **712 passed** (was 30 failed / 461 errors), lone remaining failure was this XPASS

### DIFF
--- a/backend/workflow_manager/endpoint_v2/tests/destination-connectors/test_destination_connector_postgres.py
+++ b/backend/workflow_manager/endpoint_v2/tests/destination-connectors/test_destination_connector_postgres.py
@@ -1,7 +1,6 @@
 import os
 from unittest.mock import Mock, patch
 
-import pytest
 from django.test import TestCase
 from unstract.connectors.databases.postgresql import PostgreSQL
 from workflow_manager.endpoint_v2.constants import DestinationKey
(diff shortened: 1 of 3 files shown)


pk-zipstack (approved):
LGTM

PR #2203, opened 2026-07-24
[FIX] Resolve Prompt Studio default LLM profile and unblock deploy when LLMChallenge is off

hari-kuriakose:
## Purpose

Fixes two independent server-side failures in the Prompt Studio → export → deploy path.

---

## 1. `fetch_response` ignores the project default LLM profile

`build_fetch_response_payload` (and the synchronous `_fetch_response`) resolve the profile from the **prompt's own FK** and raise `DefaultProfileError` when it is null. Both sibling paths already fall back to the project default:

| Path | Profile resolution |
| --- | --- |
| `index_document` | `ProfileManager.get_default_llm_profile(tool)` |
| `single_pass` | `ProfileManager.get_default_llm_profile(tool)` |
| `fetch_response` | **`prompt.profile_manager` only** |

`fetch_response` was the lone omission. Both call sites now fall back to the project default.

`DefaultProfileError` previously read *"Default LLM profile is not configured. Please set an LLM profile as default to continue."* — naming a setting the code never read, so setting the project default did not resolve the error. The message now names both places a profile can come from, and no longer says "for this prompt", since `_resolve_llm_ids` raises the same class for monitor/challenge resolution. Its `status_code` drops 500 → 400: it is a user-fixable configuration error, not a server fault.

## 2. An exported tool fails deployment validation

`deployment run` ends in pipeline status `ERROR` with `422 Unprocessable Entity: Tool validation failed`, at the final step after everything else succeeds.

The failure is an **`enum`** violation, not a `required` one — a present-but-empty key satisfies `required`. The chain:

1. Export resolves `challenge_llm` correctly, falling back to the default profile's LLM.
2. Tool-instance creation discards it — `get_default_settings` seeds a `"type": "string"` property with no spec `default` as `""`.
3. Because `challenge_llm` declares `adapterType: "LLM"`, `_update_schema_for_adapter_type` injects `enum: [<real adapter ids>]`.
4. `DefaultsGeneratingValidator.validate(tool_meta)` rejects `""` — not in the enum.

`validate_adapter_access` does not catch it first: it filters `AdapterInstance.objects.filter(id__in=adapter_ids)`, so `""` matches no row and passes silently.

**Fix:** `frame_spec` seeds `challenge_llm`'s spec `default` with a resolved adapter ID — the tool's own `challenge_llm` FK, else the default profile's LLM, the same precedence `frame_export_json` already uses. `get_default_settings` then seeds a real ID instead of `""`, and validation passes.

### Why not make `adapterType` conditional

An earlier revision of this PR dropped `adapterType` when `enable_challenge` was off. That suppresses the enum, but `adapterType` is the **sole** discriminator in `Spec.get_adapter_properties()` (`tool-registry/dto.py`), so dropping it also unregisters `challenge_llm` from every adapter-aware path:

- `update_metadata_with_adapter_instances` / `update_metadata_with_default_adapter` stop writing `challenge_llm_adapter_id` — which is what runtime actually reads (`structure_tool_task.py` overwrites the exported value with that key). A tool exported with challenge off and later toggled on in the workflow UI would resolve `challenge_llm = ""` and silently skip the challenge.
- The settings form loses its adapter dropdown (RJSF renders it from `enum`/`enumNames`).
- `has_adapter_ids_in_metadata` always returns False, so the lazy-migration path logs on every validation.

Thanks to @chandrasekharan-zipstack for catching this.

The suggested alternative — keep `adapterType`, drop `required`, add `default: ""` — was measured against the real `frame_spec` composed with the real `_update_schema_for_adapter_type`, and **does not clear the enum**: `get_default_settings` seeds `""` whether or not a `""` default is declared, so validation still fails. Seeding a real ID is what actually resolves it, and it lets `adapterType` and `required` return to their unconditional form — the schema shape is now unchanged from `main`.

## 3. Output was booked against the wrong profile

`cb_kwargs["profile_manager_id"]` carried the raw (possibly `None`) argument rather than the profile actually resolved. `OutputManagerHelper.get_default_profile` treats `None` as "re-resolve the project default", so a prompt carrying its own profile FK **ran** under that FK while its `PromptStudioOutputManager` row was **stored** under the project default. The read path (`_resolve`) prefers the prompt FK, so the row was written where it would not be read back.

Fixed at all three call sites — `build_fetch_response_payload`, `build_bulk_fetch_response_payload`, and the synchronous `_execute_single_prompt` → `_handle_response` path.

---

## Known limitation (not a regression)

The seeded `challenge_llm` default is resolved at export time from the exporting user's context and frozen into the persisted `tool_spec`, while `_update_schema_for_adapter_type` rebuilds the `enum` scoped to the *requesting* user. So a user the tool is shared with, who lacks access to the exporter's LLM adapter, still gets a default outside their enum and hits the same 422 — as does the exporter if that adapter is later deleted.

This is a strict improvement rather than a regression: before this PR the seed was `""`, which failed the enum for **every** user including the exporter. The set of failing cases shrinks; nobody who previously succeeded now fails. Re-resolving the default at schema-hydration time (alongside the enum, where the user-scoped adapter list is already in hand) would close it, and is better placed as a follow-up next to #2209's instance-layer overlay.

---

## Verification

Composing the real `frame_spec` with the real `_update_schema_for_adapter_type` and the real `get_default_settings`:

```
enable_challenge=False   seeded=<uuid>  deploys=True  adapter-id write=True  dropdown=True  required=True
enable_challenge=True    seeded=<uuid>  deploys=True  adapter-id write=True  dropdown=True  required=True

stored "" is still rejected (enum) in both modes
```

A stored `""` — the pre-fix state — still fails validation, so an enabled challenge cannot silently run without an LLM.

Tests: 16 challenge_llm tests plus 12 profile-resolution tests, verified in three configurations (with `pytest-django`, without it, and with a bare environment), and the real `unstract.sdk1` / `unstract.tool_registry` remain importable afterwards. `ruff` and `ruff-format` pass at the version pinned in `.pre-commit-config.yaml` (v0.3.4).

## Impact

- Prompts with no profile FK become runnable when a project default exists.
- Tools exported from Prompt Studio now deploy instead of 422-ing at the final step.
- Existing tool instances already storing `challenge_llm: ""` are fixed on re-export; no migration.

## Related

- **#2209** — fixes the same seeding bug at the tool-instance layer, by overlaying the export's resolved `tool_settings` onto the spec-seeded defaults. The two are complementary and both land a real adapter ID: this PR sets the spec `default` so seeding never produces `""`; #2209 overlays the exported value so the instance carries exactly what export resolved. Both touch `prompt_studio_registry_helper.py`, so expect a trivial conflict depending on merge order.


backend/prompt_studio/prompt_studio_registry_v2/prompt_studio_registry_helper.py
```diff
@@ -47,14 +47,25 @@ def frame_spec(tool: CustomTool) -> Spec:
...
+            "adapterIdKey": "challenge_llm_adapter_id",
+        }
+        if not tool.enable_challenge:
+            # With LLMChallenge off the tool instance stores challenge_llm as ""
+            # (no spec default to seed it). Declaring adapterType here would make
+            # the tool instance schema carry an enum of real adapter IDs only,
+            # which "" can never satisfy - so deployment validation rejects a
+            # tool that never used LLMChallenge in the first place. Leave the
+            # property a free-form string until the feature is switched on.
+            challenge_llm_property.pop("adapterType")
```
chandrasekharan-zipstack: Dropping `adapterType` does more than suppress the enum — it also disables the only two paths that **write** `challenge_llm_adapter_id` into tool-instance metadata.

Every read of `adapterIdKey` sits inside an `adapterType`-filtered loop (`tool-registry/src/unstract/tool_registry/dto.py:68-76`), so removing it takes `challenge_llm` out of `get_llm_adapter_properties()` entirely. That kills `update_metadata_with_default_adapter` (`tool_instance_helper.py:214-219`) and `update_metadata_with_adapter_instances` (`:142-148`).

Runtime reads `challenge_llm_adapter_id`, not `challenge_llm` — `workers/file_processing/structure_tool_task.py:357` unconditionally overwrites the exported value with whatever the adapter-id key holds. So:

1. Export with LLMChallenge **off** → `challenge_llm_adapter_id` never seeded.
2. User flips `enable_challenge` **on** in the workflow tool-settings UI (the spec still exposes that boolean) without re-exporting.
3. Runtime resolves `challenge_llm = ""` → `workers/executor/executors/legacy_executor.py:1564` returns silently, no log.

Pre-PR this was covered by default-adapter seeding. This contradicts the PR body's *"an enabled challenge cannot silently run without an LLM"* — that holds only when the tool was **exported** with challenge on. The verification table models the schema-validation layer correctly; the regression is in the metadata-write layer, which is why the tests pass.

Two side effects of the same change: Challenger LLM renders as a free-text box instead of a dropdown (no enum/enumNames), and `has_adapter_ids_in_metadata` (`:300-322`) always returns False, so the lazy-migration path logs on every validation.

Suggested alternative: keep `adapterType` and scope the fix to `required` + `default: ""`. If that alone doesn't clear the enum, also `pop("adapterIdKey")` and add a runtime fallback so `challenge_llm` resolves when the adapter-id key is absent — otherwise `adapterIdKey` is left as a dead key that reads as if it's still being written.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -721,6 +721,11 @@ def build_fetch_response_payload(
                 profile_manager_id=profile_manager_id
             )
 
+        # A prompt need not carry its own profile FK - fall back to the project
+        # default, matching index_document and single-pass extraction.
+        if not profile_manager:
+            profile_manager = ProfileManager.get_default_llm_profile(tool)
```
chandrasekharan-zipstack: The fallback itself is correct and worth having — `tool` is in scope, placement mirrors `_resolve_llm_ids`, and the downstream persistence path already falls back the same way (`output_manager_helper.py:232-238` → `fetch_default_llm_profile(tool)`), so outputs land against the right profile and read back.

But the `if not profile_manager: raise DefaultProfileError()` immediately below is now unreachable: `get_default_llm_profile` **raises**, it never returns `None` (`prompt_profile_manager_v2/models.py:160-167`). Harmless in effect — it raises the same class, so the improved message still surfaces — but it reads as a live guard.

Either drop the now-dead check, or make the helper return `None` so the check does the work. Same applies at line 1823.

Minor while you're in `exceptions.py`: `DefaultProfileError.status_code` is 500 for what is purely a user-configuration error. 400 would be more honest, and you're already editing those three lines.

backend/prompt_studio/prompt_studio_registry_v2/tests/test_frame_spec_challenge_llm.py
```diff
@@ -0,0 +1,323 @@
...
+            "restructured, update this test rather than deleting it."
+        )
+    body = source[source.index(start_marker) :]
+    end = body.find("\n    @staticmethod")
+    if end != -1:
+        body = body[:end]
+    return textwrap.dedent(body)
+
+
+SPEC_CLS = _load_spec_class()
```
chandrasekharan-zipstack: `SPEC_CLS = _load_spec_class()` runs at **module import** time and calls `_stub_package()` on `unstract`, `unstract.sdk1`, `unstract.sdk1.constants`, `unstract.tool_registry` — each with `__path__ = []`, replacing the real modules in `sys.modules` for the rest of the pytest process. pytest imports every test module during collection before running anything, so this lands ahead of every other backend test.

Reproduced in `backend/.venv`:

```
REAL LogLevel present: True
AFTER STUB FAILS: ImportError cannot import name 'LogLevel' from 'unstract.sdk1.constants'
adapters FAILS: ModuleNotFoundError No module named 'unstract.sdk1.adapters'
```

Production code does lazy `unstract.*` imports inside functions (e.g. `prompt_studio_core_v2/internal_views.py:412-415`, `workflow_manager/workflow_v2/views.py:1292`), so any unit test reaching those paths breaks. Green CI here is collection-order luck; under xdist it becomes order-dependent flakiness.

This is the exact pattern the repo has been removing:

- `9a76a745` (Jul 7) rewrote `test_build_index_payload.py` from 421 lines to 50, dropping ~30 stub modules — *"the clobber/restore dance was import-order dependent and silently skipped all tests when the stub surface drifted."*
- `b60dd6f3` / #2202 (Jul 24) removed the last one, from `test_execution_serializer.py`, after it *"shadowed the real module for every later test on that worker"* under xdist. That commit deliberately **deleted** the stub rather than restoring it in a `finally`, explicitly *"so the pattern can't be copied."*

`git grep 'sys.modules\[' -- 'backend/**/test_*.py'` on `main` currently returns **zero** hits. This PR would be the first to reintroduce it.

The scaffolding also isn't needed: `pytest-django>=4.12.0` is in `backend/pyproject.toml:76`, and the `unit-backend` rig group sets `DJANGO_SETTINGS_MODULE=backend.settings.test` and collects `paths: ["."]` (`tests/groups.yaml:51-76`). Importing the real modules and using `mock.patch` works — that is what `test_build_index_payload.py` does today, post-cleanup.

Same issue in #2209's test file.

chandrasekharan-zipstack (commented):
Review of the two fixes. The profile-fallback half is good; the `frame_spec` half has a regression that the schema-level tests can't see.

Also flagging the test harness: it reintroduces module-level `sys.modules` stubbing, which this repo has been systematically removing (see the inline comment on the test file).

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
Deepak-Kesavan: **[major] #2203's headline half -- "Resolve Prompt Studio default LLM profile" -- has no test at all; the PR's only new test file covers `frame_spec`/`challenge_llm`, and deleting both new `get_default_llm_profile` fallbacks leaves the suite fully green.**

Revert the fix by deleting both inserted blocks (`if not profile_manager: profile_manager = ProfileManager.get_default_llm_profile(tool)` at prompt_studio_helper.py:724-728 and :1820-1824). A prompt with no `profile_manager` FK on a project that has a default profile then raises `DefaultProfileError` (500) on `fetch_response`/single-pass -- the original CLI-surfaced bug -- and the test suite reports `9 passed in 0.04s`. There is no test that would fail if this half of the PR were reverted, and the reworded `DefaultProfileError.default_detail` in exceptions.py is likewise never asserted.

**Suggested fix:** Add a test for the resolution ladder in `PromptStudioHelper` (prompt-level `profile_manager_id` wins > prompt's own FK > project default > `DefaultProfileError`), at minimum the case the CLI hit: prompt with `profile_manager=None` on a project that has a default, asserting no `DefaultProfileError` and that the project default is selected.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:fc9c4eb4ab18</sub>

Deepak-Kesavan (commented):
## 🤖 Automated review

Generated by **Unstract PR review kit (Claude Code)** — an automated pass, not a human review. Findings are starting points, not verdicts: judge each one on the code, and close anything that is wrong or not worth doing. No approval or change-request is implied, and nothing here blocks the merge.

1 inline comment(s).

backend/prompt_studio/prompt_studio_registry_v2/prompt_studio_registry_helper.py
hari-kuriakose: You were right, and the consequence was worse than the enum suppression I was aiming at. Reworked in 83bd7d7f.

Confirmed all three casualties you named: `get_llm_adapter_properties()` filters on `adapterType`, so popping it took `challenge_llm` out of `update_metadata_with_adapter_instances` / `update_metadata_with_default_adapter` (no `challenge_llm_adapter_id` write, which is what `structure_tool_task.py:357` actually reads), out of the settings-form dropdown, and out of `has_adapter_ids_in_metadata`. Ran the real `frame_spec` and confirmed `get_llm_adapter_properties_keys()` returned an empty set with challenge off.

**Your suggested alternative doesn't clear the enum, though** — worth recording since you hedged on it. I measured `keep adapterType + drop required + default: ""` against the real `frame_spec` composed with the real `_update_schema_for_adapter_type`:

```
cand      chal    seeded  deploys  adapterID-write  dropdown  errs
CURRENT   False   ''      True     False            False     []
A         False   ''      False    True             True      ['enum']   <- your suggestion
B         False   <uuid>  True     True             True      []         <- shipped
```

`get_default_settings` seeds `""` whether or not a `""` default is declared (it only reads `default` when present — and `""` *is* the value it would seed anyway), so the enum still rejects it.

So the fix is to seed a **real** adapter ID as the spec `default`, using the same precedence `frame_export_json` already uses (tool's `challenge_llm` FK, else default profile's LLM). `adapterType` and `required` both go back to unconditional — the schema shape is now identical to `main`, and only the seeded value changes. That also avoids your `pop("adapterIdKey")` fallback, which I think would have misfired: `update_metadata_with_adapter_properties` falls back to `AdapterPropertyKey.ADAPTER_ID` when `adapterIdKey` is absent, so the ID would land under the wrong key.

Verified in both modes: deploys, adapter-id write registers, dropdown present, and a stored `""` is still rejected with an enum violation. `test_adapter_type_is_always_declared` now fails loudly if anyone pops it again.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
hari-kuriakose: Both fixed in 83bd7d7f.

Dropped the dead guard at both sites rather than changing the helper — `get_default_llm_profile` has 15 call sites that rely on it raising, so making it return `None` would be the larger blast radius. Left a comment at each site noting the helper raises, so the absence of a guard reads as deliberate.

`status_code` 500 → 400, agreed. Also took the message off "for this prompt": as CodeRabbit noted separately, `_resolve_llm_ids` raises the same class for monitor/challenge resolution, where attaching a profile to the prompt wouldn't help. It now reads "No LLM profile could be resolved. Set an LLM profile as the project default, or attach one to the prompt, to continue." Checked for consumers asserting on the 500 — none in the frontend or tests.

Left the identical dead guard at `prompt_studio_helper.py:947-950` (`build_bulk_fetch_response_payload`) alone since it predates this PR — happy to fold it in if you'd rather it go now.

backend/prompt_studio/prompt_studio_registry_v2/tests/test_frame_spec_challenge_llm.py
hari-kuriakose: Rewritten in 83bd7d7f and hardened in 27dd539f. The stubbing is gone entirely — real imports plus `mock.patch` on the two DB lookups.

Confirmed your reproduction independently: `git grep 'sys.modules\[' -- 'backend/**/test_*.py'` on `main` returns zero, and the pattern breaks things in exactly the way you describe. What I hit while verifying it: `prompt_studio_core_v2/tests/test_build_index_payload.py` still installs stubs for `account_v2` and `unstract.*` at import time via its `_install()` helpers (it assigns to `sys.modules` inside the helper, which is why the grep misses it). Collecting it alongside `test_models.py` breaks that file today, with none of my changes present:

```
ERROR test_models.py - ImportError: cannot import name 'Organization' from 'account_v2.models' (unknown location)
```

So the rewrite alone wasn't enough — my module-level `unstract.sdk1` import was also resolving to a mock depending on collection order. Everything now imports inside `_deps()` at call time.

Two corrections to the surrounding facts, since they changed what the fix had to be:

- **`pytest-django` was not in `backend/pyproject.toml`.** Line 76 was the `deploy` group. The `unit-backend` lane ran with only `django` installed and no `DJANGO_SETTINGS_MODULE`, so real imports at module scope raised `ImproperlyConfigured` during collection — I verified this against the CI log for `5a4da0e7`. (`main` has since added it, and the merge in 77c857b4 picks it up.) `_deps()` calls `django.setup()` itself so the file doesn't depend on that.
- **`unit-backend` uses an explicit path list, not `paths: ["."]`** — and `prompt_studio_core_v2/tests` isn't in it, which is why the stubbing sibling above doesn't currently break CI.

`_deps()` skips only when `django` is genuinely absent; anything else raises, so a poisoned `sys.modules` gets reported rather than silently becoming 16 skips.

Also dropped `_seeded_default` (now calls the real `ToolUtils.get_default_settings`) and `test_other_adapter_properties_are_untouched`, which only exercised code this PR doesn't modify and so could never fail.

Verified: 16 pass standalone, with and without `DJANGO_SETTINGS_MODULE`, and 135 pass across the `unit-backend` path list. Also asserted the real `unstract.sdk1` / `unstract.tool_registry` stay importable after this module loads.

Re CodeRabbit's related point: `importorskip` is gone and `jsonschema` is declared in the backend test group, so this suite can't silently skip.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
hari-kuriakose: Confirmed and fixed in 74c3aa6d.

I reproduced your revert exactly — deleting both inserted blocks left `prompt_studio` at **52 passed**, nothing red. You were right that the headline half of this PR was entirely unpinned.

Added `prompt_studio_core_v2/tests/test_profile_resolution_fallback.py` covering the full ladder you specified, at **both** call sites (`build_fetch_response_payload` and `_fetch_response`):

1. explicit `profile_manager_id` wins over everything
2. else the prompt's own `profile_manager` FK
3. else the project default — the rung this PR added, and the case the CLI hit
4. else `DefaultProfileError` propagates (the fallback must not swallow a genuinely unconfigured project)

Checked the tests actually bite rather than just passing alongside the fix:

- reverting **both** fallbacks → fails collection with an explicit message naming the missing call sites
- reverting **one** → 4 of 9 fail, so fixing one site and not the other is caught too

The ladder is extracted from the real source and anchored on the exact lines the fix added, rather than reimplemented — a reworded or deleted fallback fails loudly instead of quietly passing against a stale copy. Driving the whole function end-to-end would pull in storage, indexing and an LLM round-trip for what is a four-branch resolution.

Not addressed: you're right that the reworded `DefaultProfileError.default_detail` still isn't asserted anywhere. It's a user-facing string with no branching behaviour, so I left it — say the word if you'd like it pinned.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -907,7 +911,11 @@ def build_fetch_response_payload(
             "document_id": document_id,
             "tool_id": tool_id,
             "prompt_ids": [str(prompt.prompt_id)],
-            "profile_manager_id": profile_manager_id,
+            # Record the profile actually used, not the (possibly None) argument.
+            # The callback otherwise re-resolves the project default, so a
+            # default change mid-run would book output against a different
+            # profile than the one that produced it.
+            "profile_manager_id": str(profile_manager.profile_id),
```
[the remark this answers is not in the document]
hari-kuriakose: Already fixed in c5c2289 — same change as your proposed diff (`str(profile_manager.profile_id)` at what is now line 1108). I'd spotted the bulk site while addressing your other comment on the single-prompt path, since the two carry identical `cb_kwargs` construction.

backend/prompt_studio/prompt_studio_registry_v2/tests/test_frame_spec_challenge_llm.py
```diff
@@ -0,0 +1,234 @@
...
+
+        os.environ.setdefault("DJANGO_SETTINGS_MODULE", "backend.settings.test")
+        django.setup()
+
+        from tool_instance_v2.tool_processor import ToolProcessor
+        from prompt_studio.prompt_studio_registry_v2.prompt_studio_registry_helper import (  # noqa: E501
+            PromptStudioRegistryHelper,
+        )
+    except Exception as exc:  # pragma: no cover - environment guard
+        pytest.skip(f"Django app registry unavailable: {type(exc).__name__}: {exc}")
```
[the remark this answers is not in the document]
hari-kuriakose: Agreed, and already narrowed in 27dd539f — the diff you're quoting is from 83bd7d7f.

The guard now catches only `ImportError` on `import django`; `django.setup()` and every application import run unguarded, so configuration errors and import regressions propagate as failures rather than skips:

```python
try:
    import django
except ImportError as exc:
    pytest.skip(f"django is not installed: {exc}")

os.environ.setdefault("DJANGO_SETTINGS_MODULE", "backend.settings.test")
django.setup()
...
```

Same effect as your `importorskip` suggestion. I kept the explicit form because the docstring explains at some length *why* nothing is imported at module scope (a sibling test module stubs `unstract.*` and `account_v2` into `sys.modules` during collection), and the `try/except` keeps that reasoning adjacent to the code it justifies.

chandrasekharan-zipstack (approved):

PR #2204, opened 2026-07-24
[FIX] Keep the Prompt Studio back-reference in tool registry listing; drop dead adapter-choices route

hari-kuriakose:
## Purpose

Fixes two small, self-contained backend bugs in the tool registry surface.

---

## 1. Registry listing drops the back-reference to the Prompt Studio project

There is no way to correlate a registry entry with the project that produced it. `fetch_json_for_registry` projects only four keys — `name`, `description`, `icon`, `function_name` — where `function_name` is the `prompt_registry_id`, **a fresh UUID, not the Prompt Studio `tool_id`**. Callers are forced to match on `name`, which is ambiguous whenever two projects share one.

The data is already present and simply discarded: `PromptStudioRegistry` has a `custom_tool` `OneToOneField` pointing back at the project, and `PromptStudioRegistrySerializer` is `fields = "__all__"`, so `custom_tool` is already in `pi_serializer.data`. The projection just never copied it across.

The frontend already pays for the omission. In `CreateApiDeploymentFromPromptStudio.jsx`:

```js
tool.function_name === toolDetails.tool_id || tool.name === toolDetails.tool_name
```

The first clause can never be true (a registry UUID compared against a Prompt Studio `tool_id`), so correlation silently falls through to the ambiguous name comparison. Exposing `custom_tool` lets callers correlate exactly.

`custom_tool` is `null=True`, so legacy rows report `None` rather than failing. **Additive only** — no existing key changes, so nothing consuming this listing breaks.

## 2. `prompt-studio/adapter-choices/` routes to a method that does not exist

Returns `500 server_error` on every call.

```python
prompt_studio_adapter_choices = PromptStudioCoreView.as_view(
    {"get": "get_adapter_choices"}
)
```

`get_adapter_choices` is **not defined** — not on `PromptStudioCoreView`, not anywhere in the backend. The route resolves, the handler is missing, every request 500s.

**Fix: remove it**, so the path 404s honestly instead of 500ing. The only two references in the repo were the `as_view` registration and the URL entry itself, and no frontend code calls it. Listing LLM adapters is already available server-side via the adapter list endpoint filtered by type.

---

## Verification

- A repo-wide search for `adapter_choices` / `adapter-choices` returns nothing after the change.
- `urls.py` parses cleanly.
- `ruff` and `ruff-format` pass at the version pinned in `.pre-commit-config.yaml` (v0.3.4).

## Impact

- The registry listing can offer real `tool_id` filtering, and the frontend's dead first clause can be replaced with a correct comparison in a follow-up.
- A dead endpoint stops returning 500.

## Related

Part of a set of independent Prompt Studio / registry fixes: **#2203** (profile fallback + challenge-off deploy), **#2206** (registry delete + API-key ergonomics), **#2209** (challenge-on deploy). No code dependency between them.


backend/prompt_studio/prompt_studio_core_v2/urls.py
```diff
@@ -125,11 +120,6 @@
             prompt_studio_bulk_fetch_response,
             name="prompt-studio-bulk-fetch-response",
```
chandrasekharan-zipstack, line 126: This is mention in the public facing docs too and can be removed there as well: https://docs.unstract.com/unstract/unstract_platform/api_documentation/versions/v1-prompt-studio/

backend/prompt_studio/prompt_studio_registry_v2/prompt_studio_registry_helper.py
```diff
@@ -443,6 +443,12 @@ def fetch_json_for_registry(user: User) -> list[dict[str, Any]]:
             tool_metadata[JsonSchemaKey.FUNCTION_NAME] = prompts.get(
                 JsonSchemaKey.PROMPT_REGISTRY_ID
             )
+            # Back-reference to the Prompt Studio project that produced this
+            # entry. Without it callers can only correlate on `name`, which is
+            # ambiguous when two projects share one.
+            tool_metadata[JsonSchemaKey.CUSTOM_TOOL] = prompts.get(
```
chandrasekharan-zipstack: Additive and safe. Two notes for whoever consumes it:

1. **Nothing consumes it yet.** `CreateApiDeploymentFromPromptStudio.jsx:110-114` still runs the dead `tool.function_name === toolDetails.tool_id` clause plus the ambiguous name fallback. If the name-collision bug is the motivation, that `find` needs to switch to `tool.custom_tool === toolDetails.tool_id` — otherwise this is a payload change with no behavioural effect. Worth either riding along or being explicitly tracked.

2. **The key is absent on the other two producers** of entries in this same list — `ToolRegistry.fetch_tools_descriptions()` (`tool_registry.py:162`) and cloud's `AgenticRegistryHelper.fetch_registry_list`. Combined with `custom_tool` being `null=True`, a consumer needs `tool.custom_tool && tool.custom_tool === …`, not bare equality — otherwise `undefined === undefined` can match on static/agentic entries.

Optional hardening: `str(...)` the value when non-None. Not needed today (DRF's `JSONRenderer` encodes `UUID`), but it would become a real break if anyone later `json.dumps` or caches `get_tool_list()` output.

backend/prompt_studio/prompt_studio_core_v2/urls.py
```diff
@@ -30,9 +30,6 @@
 
 prompt_studio_prompt_index = PromptStudioCoreView.as_view({"post": "index_document"})
 prompt_studio_prompt_response = PromptStudioCoreView.as_view({"post": "fetch_response"})
-prompt_studio_adapter_choices = PromptStudioCoreView.as_view(
-    {"get": "get_adapter_choices"}
-)
 prompt_studio_bulk_fetch_response = PromptStudioCoreView.as_view(
```
chandrasekharan-zipstack, line 33: The route is dead in code, but it is **publicly documented** — `unstract-docs/docs/unstract_platform/api_documentation/versions/v1-prompt-studio.mdx:630` carries an `<ApiEndpoint>` block for `/api/v1/unstract/{org_id}/prompt-studio/adapter-choices/`, described as *"Returns available adapter choices for use in LLM profiles"*.

It was added in unstract-docs `97a8538` ("Add Platform API Documentation"), i.e. written from the URLconf, documenting an endpoint that has never worked. Without a companion docs PR this change swaps a documented 500 for a documented 404.

Grep is otherwise clean — zero hits in `frontend/src`, all of `unstract-cloud`, `unstract-python-client`, the Postman collections, and the e2e/rig suites.

chandrasekharan-zipstack (commented):
Both changes check out. `custom_tool` really is in `pi_serializer.data`, the single caller (`tool_processor.py:136-138` → `tool_instance_v2/views.py:60`) is DRF-rendered so the `uuid.UUID` value encodes fine, and no consumer does strict parsing. The dead `function_name === tool_id` clause is confirmed unreachable.

On the removed route: `get_adapter_choices` was **never defined in this repo's history** — `git log --all -G 'def get_adapter_choices'` returns nothing, so it has 500'd since the initial commit rather than being defined and later dropped. Removing it is right. Two loose ends inline.

backend/prompt_studio/prompt_studio_registry_v2/tests/test_fetch_json_for_registry.py
```diff
@@ -0,0 +1,226 @@
...
+
+    class _Manager:
+        @staticmethod
+        def list_tools(user: Any) -> list[dict[str, Any]]:
+            return serialized_rows
+
+    class _PromptStudioRegistry:
+        objects = _Manager()
+
+    class _Serializer:
```
Deepak-Kesavan: **[minor] The test harness replaces `PromptStudioRegistrySerializer` with an identity pass-through, so the one assumption that can actually regress — that the serializer still emits `custom_tool` — is never exercised; all six tests stay green if the field stops being serialized.**

The file's own docstring states the fix works because "`custom_tool` is a OneToOneField and the serializer is `fields = \"__all__\"`". If someone narrows `PromptStudioRegistrySerializer.Meta` to an explicit field list or adds `exclude` — a plausible optimization, since `__all__` currently ships the large `tool_property`/`tool_spec`/`tool_metadata` blobs on every listing — `prompts.get(JsonSchemaKey.CUSTOM_TOOL)` silently returns None for every row and the API regresses to the pre-fix behaviour, while this regression-pinning suite still reports 6 passed. The stub also feeds `str` values where the real serializer yields `uuid.UUID` (I confirmed via a standalone DRF probe: PrimaryKeyRelatedField.to_representation returns `value.pk`), so any consumer assumption about the value's Python type is likewise unpinned.

**Suggested fix:** Add one Django-backed test in this group that instantiates `PromptStudioRegistrySerializer(instance=<unsaved PromptStudioRegistry with custom_tool set>)` and asserts `"custom_tool" in serializer.data` — it needs no DB write and directly pins the assumption the rest of the file rests on.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:418a143b7fc9</sub>

Deepak-Kesavan (commented):
## 🤖 Automated review

Generated by **Unstract PR review kit (Claude Code)** — an automated pass, not a human review. Findings are starting points, not verdicts: judge each one on the code, and close anything that is wrong or not worth doing. No approval or change-request is implied, and nothing here blocks the merge.

1 inline comment(s).

hari-kuriakose:
Pushed `b7e17d9` addressing the review. Summary of what changed and why.

### 1. Publish as `tool_id`, not `custom_tool`

`custom_tool` is the ORM column name; the value it carries is a Prompt Studio `tool_id`, which is what this identifier is called on every other public surface — the API docs, and `toolDetails.tool_id` in the deployment UI. This projection already renames on the way out (`function_name` is `prompt_registry_id`), so leaking the column name broke its own convention. The key was unshipped, so the rename is free now and a breaking wire change later. `JsonSchemaKey.TOOL_ID` already existed, so no new constant.

### 2. Stringified the value

Adopted the `str()` treatment used by the neighbouring `TOOL_ID` assignment in `frame_export_json`. Closes the optional hardening noted in the review — the listing no longer depends on DRF's renderer to turn a `uuid.UUID` into JSON if this output is ever cached or re-serialized.

### 3. Documented the heterogeneity

The comment now records that `fetch_tools_descriptions()` and cloud's agentic registry contribute entries to this same list with no such key at all, so consumers must treat a missing key and a `None` value alike — i.e. `tool.tool_id && tool.tool_id === …`, not bare equality. Confirmed the merged list was homogeneous (`name`/`description`/`icon`/`function_name`) before this PR, so this change is what introduces the difference.

### 4. Rewrote the tests to import for real

The harness read the helper's source, string-sliced the function body and `exec`'d it. Its stated rationale was wrong: `pytest-django>=4.12.0` is a declared test dependency (`backend/pyproject.toml:76`), `tests/groups.yaml:83` sets `DJANGO_SETTINGS_MODULE` for the `unit-backend` tier, and the cited precedent `test_build_index_payload.py:27` imports the real module — its own docstring says so. Now imports the helper and patches its collaborators, matching the sibling suite. The extraction's end marker (`"\n    @staticmethod"`) is gone with it; `fetch_json_for_registry` is the last static method in the class, so that marker never matched and the slice silently ran to EOF.

### 5. Added the test that was missing

@Deepak-Kesavan's point was correct, and I verified it before and after. The old suite stubbed the serializer with an identity pass-through, so the assumption everything rests on — that `Meta.fields = "__all__"` still emits `custom_tool` — was never exercised:

| Mutation | Old suite | New suite |
|---|---|---|
| Narrow `Meta.fields` to drop `custom_tool` | **6 passed** (blind) | **1 failed** |
| Delete the projection line | 5 failed | 5 failed |

`test_serializer_emits_the_custom_tool_column` closes that gap. All six original scenarios are kept.

### 6. Docs

Zipstack/unstract-docs#223 (draft) removes the `<ApiEndpoint>` block at `v1-prompt-studio.mdx:630`. Merge order doesn't matter — both are corrections to the same never-functional endpoint.

### Verification

`7 passed` under the rig's unit-backend env; `ruff` and `ruff-format` clean at the pinned v0.3.4.

One thing left open for a follow-up, unchanged by this push: the frontend `find` in `CreateApiDeploymentFromPromptStudio.jsx:110-114` still runs the dead `tool.function_name === toolDetails.tool_id` clause. It can now become `tool.tool_id && tool.tool_id === toolDetails.tool_id`.

backend/prompt_studio/prompt_studio_core_v2/urls.py
hari-kuriakose: Done — opened Zipstack/unstract-docs#223 (draft), which removes the `<ApiEndpoint>` block for this path from `v1-prompt-studio.mdx`. `grep -rn "adapter-choices" docs/` returns nothing after the change.

Merge order doesn't matter: both are corrections to the same endpoint, which never worked in either state.

backend/prompt_studio/prompt_studio_core_v2/urls.py
hari-kuriakose: Addressed in `b7e17d9` + Zipstack/unstract-docs#223 (draft).

Confirmed your finding independently: the block is at `v1-prompt-studio.mdx:630`, and it was indeed written from the URLconf rather than from working behaviour. The docs PR removes it, so this stops being "a documented 500 becomes a documented 404" and becomes "an endpoint that never existed is no longer advertised."

Also confirmed your point that the handler was never defined at any commit — which is why removal is the right call rather than restoring a handler.

backend/prompt_studio/prompt_studio_registry_v2/prompt_studio_registry_helper.py
hari-kuriakose: All three points addressed in `b7e17d9`.

**1. Nothing consumes it yet** — still true, and now cheaper to fix. The key is published as `tool_id` rather than `custom_tool`, so the frontend `find` becomes `tool.tool_id && tool.tool_id === toolDetails.tool_id` — comparing like with like instead of a registry UUID against a project id. I left the JSX itself out of this PR to keep the diff to the backend surface; flagged it as the follow-up rather than letting it drift.

The rename was worth doing now: `custom_tool` is the ORM column name, while the value is a Prompt Studio `tool_id` — which is what it's called in the API docs and in `toolDetails.tool_id`. This projection already renames on the way out (`function_name` is `prompt_registry_id`), so shipping the raw column name broke its own convention. Unshipped key, so free now and a breaking wire change later. `JsonSchemaKey.TOOL_ID` already existed.

**2. Key absent on the other two producers** — confirmed and documented at the projection. I checked `fetch_tools_descriptions` (`tool_registry.py:191-196, 205-210`) and it emits exactly `name`/`description`/`icon`/`function_name`, so the merged list was homogeneous *before* this PR — this change is what introduces the difference, and it can't be waved off as pre-existing. The comment now states that consumers must treat a missing key and a `None` value alike, i.e. the `&&` guard you described, not bare equality.

**3. Optional `str()` hardening** — taken, rather than deferred. It matches the neighbouring `TOOL_ID` use in `frame_export_json`, and it means the value no longer depends on DRF's renderer to leave the dict. You were right that it doesn't break anything today (I traced every consumer: only `tool_processor.py:133-147` → the DRF view, no `json.dumps`/cache/Celery path) — but it's one line and removes the trap for whoever caches `get_tool_list()` later.

backend/prompt_studio/prompt_studio_registry_v2/tests/test_fetch_json_for_registry.py
hari-kuriakose: You were right, and I verified it both ways before and after the fix. Addressed in `b7e17d9`.

**Confirmed empirically.** I applied exactly the regression you described — narrowing `Meta.fields` away from `"__all__"` so `custom_tool` is no longer serialized — and ran the suite:

| Mutation | Old suite | New suite |
|---|---|---|
| Narrow `Meta.fields` to drop `custom_tool` | **6 passed** — blind | **1 failed** |
| Delete the projection line | 5 failed | 5 failed |

So the suite did catch a direct revert of the line, but was completely blind to the serializer-side regression it advertised pinning — which is the one that would silently restore the original bug in production.

**Fix:** added `test_serializer_emits_the_custom_tool_column`, essentially your suggestion — it asserts against the real `PromptStudioRegistrySerializer().get_fields()`, needs no DB write, and fails with a diagnostic message pointing at the consequence.

**Went a bit further on the harness itself.** The reason the stub was there at all was the docstring's claim that "Django is not importable in a plain checkout (no `pytest-django`)". That turned out to be false: `pytest-django>=4.12.0` is a declared test dependency (`backend/pyproject.toml:76`), `tests/groups.yaml:83` sets `DJANGO_SETTINGS_MODULE` for the `unit-backend` tier, and the file it cited as precedent (`test_build_index_payload.py:27`) imports the real module — its own docstring says so explicitly. So the whole read-source-and-`exec` harness is gone; the tests now import the helper and patch its collaborators, and CI's unit tier passes, which is the proof that Django was importable all along.

That also removed a latent trap you didn't mention: the extraction's end marker was `"\n    @staticmethod"`, but `fetch_json_for_registry` is the last static method in the class, so the marker never matched and the slice silently ran to EOF.

Your point about type fidelity was also correct — `PrimaryKeyRelatedField.to_representation` returns `value.pk`, a `uuid.UUID`, not a `str`. The fixtures now use `uuid.UUID` objects, and the projection stringifies explicitly, so the wire type is pinned rather than incidental.

All six original scenarios are kept.

hari-kuriakose: Follow-up on `b79bc1ef` — including a correction to something I asserted in my previous reply.

### Correction: my "the cited precedent imports the real module" claim was wrong

I said `test_build_index_payload.py` imports the real helper and that the old harness's rationale was false on that basis. The first half is inaccurate. That file stubs collaborators on `sys.modules` *before* importing, and its own docstring (line 16) says the environment has no `pytest-django` — the same claim I called false. It degrades to `4 skipped` under a bare `pytest` rather than erroring.

What still stands: `pytest-django>=4.12.0` **is** a declared test dependency (`backend/pyproject.toml:76`), `tests/groups.yaml:83` **does** set `DJANGO_SETTINGS_MODULE` for the `unit-backend` tier, and importing for real works there — CI's unit tier passes. So the rewrite is right; my justification named the wrong precedent. The docstring now states the rig requirement directly instead of citing a file that does something else.

Related: a review flagged that this file errors at collection under bare `pytest`. It does — but so does the pre-existing `test_models.py` in the same directory, untouched by this PR, with the identical `Requested setting INSTALLED_APPS`. It's a property of every Django-coupled suite here, not a regression.

### The `tool_id` name was a mistake; it is now `prompt_studio_tool_id`

I checked for a collision by looking at listing entries and reported it clear. That was the wrong surface. `ToolInstance.tool_id` is `db_comment="Function name of the tool being used"` and resolves via `ToolProcessor.get_tool_by_uid` — and `CreateApiDeploymentFromPromptStudio.jsx:375`, in the same component that reads this listing, posts `tool_id: toolFunctionName || toolDetails.tool_id` to `tool_instance/`. Publishing a CustomTool pk under `tool_id` invites feeding it back as a function name, which raises `ToolDoesNotExist`.

The frontend follow-up is therefore `tool.prompt_studio_tool_id && tool.prompt_studio_tool_id === toolDetails.tool_id`.

### The assertion I added last round was itself vacuous

Worth stating plainly, since I criticized the previous suite for exactly this. `prompt_registry_id` is a `UUIDField` → DRF renders a **`str`**; `custom_tool` is a `OneToOneField` → `PrimaryKeyRelatedField` renders a raw **`UUID`**. I typed both fixtures as `uuid.UUID`, so the guard `function_name != tool_id` was passing on the *type mismatch* — it would have passed even with both keys carrying the same identifier, which is the precise thing it exists to forbid. CodeRabbit flagged the type issue; verified against the live serializer:

```
prompt_registry_id -> UUIDField           -> '52f4e38a-…'  (str)
custom_tool        -> PrimaryKeyRelatedField -> UUID('52f4e38a-…')
```

Fixed, and `test_the_two_ids_are_compared_like_for_like` now fails if the mismatch ever comes back.

Also strengthened per your original point: the serializer test drove `get_fields()`, which reflects declared fields rather than output — a `to_representation` override dropping the key would have passed. It now renders through the real per-field `to_representation`, staying DB-free.

Mutation results, all four caught (was 2/4):

| Mutation | Before | Now |
|---|---|---|
| Narrow `Meta.fields` | 6 passed | 1 failed |
| Delete projection line | 5 failed | 6 failed |
| Make `function_name` == back-ref | *would pass* | 6 failed |
| Drop `many=True` | 6 passed | 7 failed |

backend/prompt_studio/prompt_studio_registry_v2/tests/test_fetch_json_for_registry.py
```diff
@@ -0,0 +1,256 @@
...
+    fields = {
+        name: field
+        for name, field in serializer.get_fields().items()
+        if not field.write_only and hasattr(stub, name)
+    }
+
+    rendered = {
+        name: field.to_representation(getattr(stub, name))
+        for name, field in fields.items()
+    }
```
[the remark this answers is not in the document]
hari-kuriakose: Valid finding — fixed in `2694de02`.

You're right that the field-by-field reconstruction skipped `Serializer.to_representation` itself. That's the same blind spot as the `get_fields()` version it replaced, just one level up: the field layer was exercised, the serializer layer wasn't.

Now renders in a single `PromptStudioRegistrySerializer().to_representation(stub)` call. Verified with exactly the mutation you described — a serializer-level override that pops the key after the fields have run:

```python
def to_representation(self, instance):
    rep = super().to_representation(instance)
    rep.pop("custom_tool", None)
    return rep
```

| | Result |
|---|---|
| `b79bc1ef` (field-by-field) | **8 passed** — blind, as you said |
| `2694de02` (serializer-level) | **1 failed** |

On the "populated unsaved model instance" part of the suggestion: I kept the `SimpleNamespace` stub rather than instantiating `PromptStudioRegistry`. A real instance — even unsaved — makes `PrimaryKeyRelatedField` evaluate its queryset, which raises `RuntimeError: Database access not allowed` under the unit tier. I hit that when first strengthening this assertion. The stub now carries every field the serializer reads (`url`, the three JSON blobs, `shared_to_org`, `shared_users`, `organization`, and the audit columns), so it renders identically while staying DB-free, and a newly added serializer field shows up here as a loud `AttributeError` rather than a silent coverage gap.

chandrasekharan-zipstack (approved):

PR #2205, opened 2026-07-24
[MISC] Fail rig groups that inherit an unrelated pytest config; skip order-dependent connector tests

chandrasekharan-zipstack:
## Why

Two related test-rig defects surfaced while wiring cloud CI onto the rig.

**1. A group can silently inherit another project's pytest config.**
The rig runs `cd <workdir> && pytest <paths>` and never pins the config file. pytest resolves config by walking *up* from the workdir, so a group whose workdir is a nested project without its own config adopts its parent's `addopts` — coverage targets, `--strict-config` — and fails for reasons that have nothing to do with its tests. Observed symptom: `ERROR: Unknown config option: asyncio_mode` followed by `no tests ran`, and a tier exit 4.

`--rootdir` does **not** fix this (verified empirically — it moves rootdir but leaves ini-file discovery alone). Only an own config section, or `-c`, does.

This matters beyond the one case: `workers/plugins/` holds 11 nested projects, each a future instance of the same trap.

**2. Several connector tests are order-dependent and dormant.**
They share a fixed table (or remote path) across test methods with no working teardown — several `tearDown` bodies are a bare `pass` with the drops commented out. They are inert only because their credentials aren't provisioned, so the breakage would appear as a surprise the moment someone adds them.

## What

- `_resolve_pytest_configfile()` mirrors pytest's upward search; `_config_isolation_error()` fails a group up front when the resolved config belongs to neither its workdir nor the repo root, with a message naming the offending file and the fix.
  - Inheriting the **repo-root** config stays allowed — that's the normal monorepo case, and `unit-core` / `unit-platform-service` rely on it today.
- Explicit `@unittest.skip` with the reason stated on the 7 dormant connector test classes and the 2 SharePoint tests, so enabling credentials becomes a deliberate step.
  - The **postgres** variant is deliberately untouched: it runs today and is being fixed rather than skipped.

## Verification

- Guardrail checked against all **18** existing pytest groups — none rejected.
- Rig self-tests: **86 passed, 1 skipped** (5 new tests cover own-config, root-config, and nested-sibling-config cases).

### DIFF
--- a/tests/rig/cli.py
+++ b/tests/rig/cli.py
@@ -30,6 +30,7 @@ from tests.rig.groups import (
     REPO_ROOT,
     TIERS,
     GroupDefinition,
+    MigrateSpec,
     load_groups,
 )
 from tests.rig.reporting import (
@@ -63,6 +64,11 @@ _PYTEST_PLUGIN_DIR = REPO_ROOT / "tests" / "rig" / "pytest_plugin"
 
 # saxutils.escape leaves quotes alone, which is wrong inside an attribute.
 _XML_ATTR_ESCAPES = {'"': "&quot;"}
+
+# Cap the pre-group `manage.py migrate`. It runs outside the group's own
+# timeout budget, so without this a wedged connection hangs the job to the CI
+# ceiling.
+_MIGRATE_TIMEOUT_SECONDS = 300
 
 
 @lru_cache(maxsize=1)
@@ -456,14 +462,31 @@ def cmd_run(args: argparse.Namespace) -> int:
             endpoints = runtime.up()
         elif needs_services and not args.dry_run:
             # Infra-only: testcontainers Postgres/Redis/etc., no platform
-            # services. ponytail: up() starts the full infra set even if a run
-            # only needs Postgres; trim to the requested services if startup
-            # cost ever matters.
+            # services. up() starts the full infra set even if a run only needs
+            # Postgres; trim to the requested services if startup cost matters.
             runtime = TestcontainersRuntime()
             print(
                 f"[rig] bringing infra up via runtime={runtime.name} (requires_services)"
             )
             endpoints = runtime.up()
+
+        # Schema for the database the rig just provisioned. Guarded on
+        # `postgres_url` so it only fires for a rig-owned container: under the
+        # compose runtime the platform's own backend migrates on startup, and
+        # migrating its database from here would be both redundant and wrong.
+        if endpoints is not None and manifest.postgres_migrate is not None:
+            postgres_url = endpoints.infra.postgres_url
+            if postgres_url:
+                migrate_exit = _run_migrations(
+                    manifest.postgres_migrate, postgres_url=postgres_url
+                )
+                if migrate_exit != 0:
+                    # Every DB-backed test would skip on a missing table, which
+                    # reads as "nothing to run" rather than a broken database.
+                    raise RuntimeError(
+                        f"migrations failed against the provisioned database "
+                        f"(exit {migrate_exit}); aborting before any group runs"
+                    )
 
         # Groups run in topo order, so a dependency's result is always known by
         # the time its dependents come up. A red dep means the precondition it
@@ -730,7 +753,17 @@ def _inject_infra_env(
         return
     infra = endpoints.infra
     if "postgres" in group.requires_services and infra.postgres_url:
-        env.update(_db_env_from_postgres_url(infra.postgres_url))
+        db_env = _db_env_from_postgres_url(infra.postgres_url)
+        env.update(db_env)
+        # A rig-provisioned Postgres is bare — `public` is the only schema it
+        # has. Groups needing another one declare DB_SCHEMA themselves.
+        env.setdefault("DB_SCHEMA", "public")
+        db_env["DB_SCHEMA"] = env["DB_SCHEMA"]
+        # The workers' real-Postgres fixtures read TEST_DB_* so that a developer
+        # run keeps pointing at their own compose database. Mirror the
+        # provisioned values onto that prefix, or those tests silently connect
+        # to the dev-compose defaults instead of the container the rig started.
+        env.update({f"TEST_{key}": value for key, value in db_env.items()})
     if "minio" in group.requires_services and infra.minio_endpoint:
         # http: this is a local, throwaway testcontainers MinIO with no TLS.
         env.setdefault(
@@ -840,27 +873,12 @@ def _execute_group(
     if group.runner == "hurl":
         cmd = _hurl_command(group, workdir)
         exit_code = _spawn(cmd, env=env, cwd=workdir, timeout=timeout)
-        # Match the exit.txt write's defensive handling: a read-only reports
-        # dir or full disk shouldn't abort the whole run and orphan completed
-        # groups before the summary renders.
-        try:
-            _write_synthetic_junit(junit, group.name, exit_code)
-        except OSError as err:
-            print(
-                f"[rig] could not write synthetic junit for {group.name}: {err}",
-                file=sys.stderr,
-            )
+        _write_synthetic_junit_safe(junit, group.name, exit_code)
     elif (isolation_error := _config_isolation_error(group, workdir)) is not None:
         print(isolation_error, file=sys.stderr)
         # 4 is pytest's usage-error code, so this gates like any hard failure.
         exit_code = 4
-        try:
-            _write_synthetic_junit(junit, group.name, exit_code)
-        except OSError as err:
-            print(
-                f"[rig] could not write synthetic junit for {group.name}: {err}",
-                file=sys.stderr,
-            )
+        _write_synthetic_junit_safe(junit, group.name, exit_code)
     else:
         cmd = _pytest_command(
             group,
@@ -928,6 +946,45 @@ def _prepare_group_env(group: GroupDefinition, *, env: dict[str, str]) -> None:
     # NOTE: rig pytest plugins (pytest-timeout, pytest-md-report, etc.) are
     # injected via `uv run --with ...` in _pytest_command, not installed here.
     # That avoids losing them on the next `uv run` (which re-syncs the venv).
+
+
+def _run_migrations(spec: MigrateSpec, *, postgres_url: str) -> int:
+    """Apply Django migrations to the rig-provisioned database.
+
+    Runs once, right after the container is up and before any group, because the
+    schema is a property of the database rather than of a group. Only the
+    migrating project's own Django settings are layered over the connection env
+    — a group's environment is irrelevant here and may not even exist yet.
+    """
+    env = {
+        **os.environ,
+        **_db_env_from_postgres_url(postgres_url),
+        # A rig-provisioned Postgres is bare; `public` is the only schema it has.
+        "DB_SCHEMA": "public",
+        **spec.env,
+    }
+    workdir = spec.absolute_workdir()
+    base = ["uv", "run", "python"] if shutil.which("uv") else [sys.executable]
+    cmd = [*base, "manage.py", "migrate", *spec.apps, "--noinput"]
+    print(
+        f"[rig] migrating provisioned database via {spec.workdir}: "
+        f"{' '.join(spec.apps) or 'all apps'}"
+    )
+    return _spawn(cmd, env=env, cwd=workdir, timeout=_MIGRATE_TIMEOUT_SECONDS)
+
+
+def _write_synthetic_junit_safe(path: Path, group_name: str, exit_code: int) -> None:
+    """Mirror the exit.txt write's defensive handling: a read-only reports dir
+    or full disk shouldn't abort the whole run and orphan completed groups
+    before the summary renders.
+    """
+    try:
+        _write_synthetic_junit(path, group_name, exit_code)
+    except OSError as err:
+        print(
+            f"[rig] could not write synthetic junit for {group_name}: {err}",
+            file=sys.stderr,
+        )
 
 
 def _pytest_base_cmd(group: GroupDefinition, workdir: Path) -> list[str]:
@@ -1074,7 +1131,8 @@ def _hurl_command(group: GroupDefinition, workdir: Path) -> list[str]:
 
 
 def _write_synthetic_junit(path: Path, group_name: str, exit_code: int) -> None:
-    """Synthesise a JUnit XML for hurl runs.
+    """Synthesise a JUnit XML for a group pytest never wrote one for (hurl runs,
+    or a group whose pre-run migration failed).
 
     Exit 5 ("no tests collected") must produce failures=0; otherwise an empty
     hurl group would show ⚪ via :class:`GroupResult` while also being counted
--- a/tests/rig/cli.py
+++ b/tests/rig/cli.py
@@ -850,6 +850,17 @@ def _execute_group(
                 f"[rig] could not write synthetic junit for {group.name}: {err}",
                 file=sys.stderr,
             )
+    elif (isolation_error := _config_isolation_error(group, workdir)) is not None:
+        print(isolation_error, file=sys.stderr)
+        # 4 is pytest's usage-error code, so this gates like any hard failure.
+        exit_code = 4
+        try:
+            _write_synthetic_junit(junit, group.name, exit_code)
+        except OSError as err:
+            print(
+                f"[rig] could not write synthetic junit for {group.name}: {err}",
+                file=sys.stderr,
+            )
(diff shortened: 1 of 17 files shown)


workers/tests/conftest.py
```diff
@@ -200,6 +200,80 @@ def pg_client(pg_conn):
...
+# one physical table, and the reaper/sweeper paths scan *all* rows — so a whole-
+# table assertion in one worker sees another worker's rows. Give each worker its
+# own schema (the raw-connection equivalent of Django's per-worker database) and
+# truncate between tests, since a raw connection carries committed rows to the
+# next test with no rollback.
+
+
+def _worker_pg_schema() -> str:
+    worker = os.getenv("PYTEST_XDIST_WORKER", "")
+    return f"test_{worker}" if worker else "public"
```
praveen-formido: **Hardcoding `public` here disables the whole real-PG suite on a developer machine.**

`create_pg_connection` defaults `{prefix}SCHEMA` to `unstract` (`queue_backend/pg_queue/connection.py`), and nothing else in the tree sets `TEST_DB_SCHEMA`. So before this PR a local `pytest workers/tests -m integration` connected with `search_path=unstract` — where `backend migrate` actually put the pg_queue tables in dev.

After this PR, `_pg_worker_schema_env` unconditionally sets `TEST_DB_SCHEMA=public` (serial run → `_worker_pg_schema()` returns `"public"`). `pg_conn`'s `SELECT to_regclass('pg_queue_message')` then resolves against `public`, finds nothing, and every one of the ~100 real-Postgres tests hits `_skip_or_fail_no_pg("pg_queue migration not applied")`. Under xdist locally it's the same outcome by a different route: the clone source `public.pg_barrier_state` doesn't exist, so no schema is created and the tests skip anyway.

CI is fine because the rig injects `DB_SCHEMA=public` and migrates into `public` — but the new test's own docstring in `test_cli.py` says "a developer run keeps hitting their own compose DB", and this breaks precisely that.

Suggest deriving the base schema instead of literalising it:

```python
def _base_pg_schema() -> str:
    return os.getenv("TEST_DB_SCHEMA") or os.getenv("DB_SCHEMA") or "unstract"

def _worker_pg_schema() -> str:
    worker = os.getenv("PYTEST_XDIST_WORKER", "")
    return f"test_{worker}" if worker else _base_pg_schema()
```

and using that same base as the clone source in `_pg_worker_schema` (both the `to_regclass('public.pg_barrier_state')` probe and `LIKE public.{table}`), rather than `public`.

workers/tests/conftest.py
```diff
@@ -200,6 +200,80 @@ def pg_client(pg_conn):
...
+                                f"(LIKE public.{table} INCLUDING ALL)"
+                            )
+                conn.commit()
+            finally:
+                conn.close()
+    yield schema
+
+
+@pytest.fixture(autouse=True)
+def _pg_worker_schema_env(_restore_os_environ, _pg_worker_schema):
```
praveen-formido: **This is `autouse` with no marker/fixture gate, so it opens a Postgres connection before every test in the suite — including the DB-free unit lane.**

`workers/tests` has ~660 test functions and the `unit-workers` group runs `-m "not integration"` with no Postgres provisioned. `integration_pg_conn()` → `create_pg_connection()` retries transient `OperationalError` 3× with 0.5 s → 1.0 s backoff before raising, so each of those tests pays ~1.5 s of pure `time.sleep` inside `contextlib.suppress`. That's ~16 minutes of sleep spread across the xdist workers, against a 600 s group timeout — and it buys nothing, because those tests never touch a database.

Three ways out, roughly in order of preference:

1. Resolve reachability once in the session fixture and short-circuit — `_pg_worker_schema` already connects; have it yield `None`/a sentinel when Postgres is unreachable and skip the truncate entirely.
2. Drop `autouse` and have the real-PG fixtures (`pg_conn`, `barrier_db`, `dedup_db`, `lock_db`, `barrier_conn`) depend on it — that's the same set `pytest_collection_modifyitems` already enumerates for auto-marking, so it stays in one place.
3. At minimum, set `WORKER_PG_QUEUE_CONNECT_RETRIES=1` for this path so a miss costs one refused connect rather than 1.5 s.

Option 2 also stops the fixture mutating `DB_SCHEMA`/`TEST_DB_SCHEMA` for hundreds of tests that have no interest in either.

tests/rig/cli.py
```diff
@@ -933,6 +1003,62 @@ def _pytest_base_cmd(group: GroupDefinition, workdir: Path) -> list[str]:
...
+)
+
+
+def _resolve_pytest_configfile(workdir: Path) -> Path | None:
+    """Find the config file pytest would adopt when invoked from `workdir`.
+
+    Mirrors pytest's upward search. `pytest.ini` counts even when empty; the
+    shared files only count when they carry a pytest section.
+    """
+    for directory in [workdir, *workdir.parents]:
```
praveen-formido, line 1040: The walk starts at `workdir`, but pytest starts ini discovery at **the common ancestor of the args** and only falls back to cwd when no path args exist. The rig always passes `paths`, so the two disagree whenever `paths` reach into a subdirectory.

Concretely, for the very shape this guardrail exists to catch:

```yaml
workdir: workers
paths: [plugins/agentic_table/tests]
```

pytest resolves from `workers/plugins/agentic_table/tests` upward and adopts `workers/pyproject.toml` — the inherited-parent-config bug. The guardrail resolves from `workers`, finds `workers/pyproject.toml`, sees `configfile.parent == workdir`, and passes it.

It only fires when the *workdir itself* is the nested project. Resolving from `os.path.commonpath(group.absolute_paths())` instead would match pytest and cover both spellings.

tests/rig/cli.py
```diff
@@ -840,16 +875,12 @@ def _execute_group(
...
-        # groups before the summary renders.
-        try:
-            _write_synthetic_junit(junit, group.name, exit_code)
-        except OSError as err:
-            print(
-                f"[rig] could not write synthetic junit for {group.name}: {err}",
-                file=sys.stderr,
-            )
+        _write_synthetic_junit_safe(junit, group.name, exit_code)
+    elif (isolation_error := _config_isolation_error(group, workdir)) is not None:
```
praveen-formido, line 892: This check depends only on the manifest and the filesystem, yet it runs here — after `_prepare_group_env` has already done `uv sync`, and after infra/platform bring-up in `cmd_run`. A misconfigured group burns a container stack and a dep sync before being told its config is wrong.

Since `load_groups` already does this class of validation (`_validate_paths`, `_validate_no_cycles`, `_validate_platform_groups_depend_on_gate`), running it there too would make `python -m tests.rig validate` catch it in seconds — and catch it in review rather than mid-run. Keeping this run-time gate as well is fine; it's the only place that can produce a per-group junit.

pk-zipstack (commented):
## Review

The config-isolation guardrail is a good catch and well-tested — `_resolve_pytest_configfile` mirrors pytest's file precedence correctly, allowing repo-root inheritance is the right call, and `_build_postgres_migrate` validating `manage.py` at load time is exactly where that check belongs.

Most of my concerns are in the parts the PR description doesn't cover. The title/body describe (1) the guardrail and (2) the connector skips, but the diff also adds a new manifest feature (`defaults.postgres_migrate` + parsing/validation/execution), the `TEST_DB_*` env mirror, and per-worker Postgres schema isolation in `workers/tests/conftest.py`. That last one carries the real risk here, and a reviewer keying off the description would skip straight past it. Worth expanding the body.

### Ranked

1. **`TEST_DB_SCHEMA` is forced to `public`, which silently disables the real-PG suite on developer machines** — `workers/tests/conftest.py` (inline).
2. **Autouse per-test Postgres connect adds ~1.5 s/test of retry backoff to the DB-free unit lane** — `workers/tests/conftest.py` (inline).
3. Guardrail resolves config from `workdir`, pytest resolves from the args' common ancestor — a false negative on the `workers/plugins/` shape the PR is guarding against (inline).
4. Guardrail is a run-time check behind `uv sync` + infra bring-up rather than a load-time validation (inline).
5. Smaller items inline: `spec.env` precedence in `_run_migrations`, `update` vs `setdefault` for the `TEST_*` mirror, substring config-section detection, `_write_synthetic_junit` docstring drift.

### Nits (not worth blocking)

- `test_sharepoint_fs.py`: stacking `@unittest.skip` above `@unittest.skipUnless` makes the inner decorator dead — the outer skip always wins and overwrites the reason. Same net effect, but the intent reads better if the `skipUnless` goes or gets a note saying it's kept for when the outer one is lifted.
- `_pg_worker_schema` imports `QUEUE_TABLES` unconditionally but only uses it when `schema != "public"`.
- `tests/README.md`: the `postgres_migrate` row sits in the per-group options table with a parenthetical explaining it is *not* a per-group key. That's a sign it wants its own short subsection.
- The connector-test diffs mix the `@unittest.skip` change with unrelated import reordering and line-length reflows. Harmless, just noise in the blame.

### Things that check out

- `LIKE ... INCLUDING ALL` does preserve the `expires_at > created_at` CHECK and the partial `state='ready'` index, so `test_expires_at_must_exceed_created_at` stays meaningful against the clone.
- The migration guard on `endpoints.infra.postgres_url` is correct — only `TestcontainersRuntime` populates it, so compose/local runs can't trigger a migrate against a real database.
- `last_progress_at` added to the raw inserts in `test_pg_barrier.py` matches `0001_initial_squashed` (`DateTimeField(default=timezone.now)`, no DB-level default after Django's AddField).

tests/rig/cli.py
```diff
@@ -919,6 +950,45 @@ def _prepare_group_env(group: GroupDefinition, *, env: dict[str, str]) -> None:
...
+    schema is a property of the database rather than of a group. Only the
+    migrating project's own Django settings are layered over the connection env
+    — a group's environment is irrelevant here and may not even exist yet.
+    """
+    env = {
+        **os.environ,
+        **_db_env_from_postgres_url(postgres_url),
+        # A rig-provisioned Postgres is bare; `public` is the only schema it has.
+        "DB_SCHEMA": "public",
+        **spec.env,
```
praveen-formido, line 980: `**spec.env` lands *after* `_db_env_from_postgres_url(postgres_url)`, so a `DB_HOST`/`DB_NAME`/`DB_USER` in `defaults.postgres_migrate.env` silently wins over the provisioned container's connection — migrations would land in a different database than the groups then test against, with nothing in the output saying so.

Today's manifest is safe (the `backend_test_env` anchor carries no `DB_*` connection keys), but that anchor is now shared with `unit-backend`/`integration-backend`, so a var added for one purpose silently changes the other. Either put `**spec.env` first and let the derived connection vars win, or reject connection keys in `_build_postgres_migrate` — that function already validates shape, so it's the natural place.

tests/rig/cli.py
```diff
@@ -730,7 +757,15 @@ def _inject_infra_env(
...
-        env.update(_db_env_from_postgres_url(infra.postgres_url))
+        db_env = _db_env_from_postgres_url(infra.postgres_url)
+        env.update(db_env)
+        # A bare Postgres only has `public`; groups needing another declare it.
+        env.setdefault("DB_SCHEMA", "public")
+        db_env["DB_SCHEMA"] = env["DB_SCHEMA"]
+        # Fixtures read TEST_DB_* so a developer run keeps pointing at their own
+        # compose DB. Mirror onto that prefix or those tests connect to the
+        # dev-compose defaults instead of the provisioned container.
+        env.update({f"TEST_{key}": value for key, value in db_env.items()})
```
praveen-formido: `update` rather than `setdefault` here, two lines after `DB_SCHEMA` gets the `setdefault` treatment. A group that declares its own `TEST_DB_HOST`/`TEST_DB_NAME` in `env:` — the escape hatch for pointing a group at a specific database — gets silently overwritten. Worth picking one convention; `setdefault` matches the surrounding `_inject_infra_env` code and the `UNSTRACT_*` handling in `_execute_group`.

tests/rig/cli.py
```diff
@@ -1007,7 +1133,8 @@ def _hurl_command(group: GroupDefinition, workdir: Path) -> list[str]:
 
 
 def _write_synthetic_junit(path: Path, group_name: str, exit_code: int) -> None:
-    """Synthesise a JUnit XML for hurl runs.
+    """Synthesise a JUnit XML for a group pytest never wrote one for (hurl runs,
+    or a group whose pre-run migration failed).
```
praveen-formido: Docstring drift: a failed pre-run migration raises `RuntimeError` in `cmd_run` and aborts before any group runs, so no synthetic junit is ever written for that case. The actual second caller added by this PR is the config-isolation branch in `_execute_group`. Suggest "(hurl runs, or a group rejected by the config-isolation guardrail)".

tests/rig/cli.py
```diff
@@ -933,6 +1003,62 @@ def _pytest_base_cmd(group: GroupDefinition, workdir: Path) -> list[str]:
...
+    """
+    for directory in [workdir, *workdir.parents]:
+        for name, required_section in _PYTEST_CONFIG_FILES:
+            candidate = directory / name
+            if not candidate.is_file():
+                continue
+            if required_section is None:
+                return candidate
+            try:
+                if required_section in candidate.read_text(encoding="utf-8"):
```
praveen-formido: Minor: a plain substring test means a commented-out `# [tool.pytest.ini_options]`, or the string appearing in a docstring/comment, counts as a real config file — and conversely `[pytest]` in `tox.ini` matches anywhere in the file. Matching at line start after `strip()` would be closer to what the TOML/INI parsers actually see:

```python
if any(line.strip() == required_section for line in candidate.read_text(encoding="utf-8").splitlines()):
```

Also worth broadening the fix hint in `_config_isolation_error`: it names `{workdir}/pyproject.toml`, but for a workdir with no `pyproject.toml` an empty `pytest.ini` is the lighter fix and is what pytest documents for exactly this.

chandrasekharan-zipstack:
@pk-zipstack thanks — all addressed in 6ef160ec. Mapping:

- **conftest `public` hardcode disables local real-PG suite** — added `_base_pg_schema()` deriving from `TEST_DB_SCHEMA`/`DB_SCHEMA`/`unstract`; the per-worker clone now sources from that base (both the `to_regclass` probe and `LIKE`).
- **autouse fixture connects before every unit test** — the per-test schema-env/truncate fixture now short-circuits unless the test is `integration`-marked *and* the session fixture found Postgres reachable (it now yields `None` when not). Unit lane opens no connection and its `DB_SCHEMA` env is left untouched.
- **guardrail resolves from workdir, not the paths' ancestor** — now resolves from `commonpath(paths)`, matching pytest; catches a nested project whose paths reach past the workdir (added a test).
- **guardrail runs after uv sync / infra** — now also runs in `python -m tests.rig validate` (skips optional placeholders), so misconfig is caught in review; the run-time gate stays for the per-group junit.
- **`**spec.env` overrides provisioned connection** — reordered so the derived connection wins; only `DB_SCHEMA` stays spec-overridable.
- **`update` vs `setdefault` for `TEST_DB_*`** — switched to `setdefault` so a group's own `TEST_DB_*` escape hatch survives.
- **synthetic-junit docstring drift** — now names the config-isolation branch as the second caller.
- **substring config-section match** — now matches the section header on its own line; fix hint falls back to `pytest.ini` when the workdir has no `pyproject.toml`.

workers/tests/conftest.py
```diff
@@ -200,6 +200,99 @@ def pg_client(pg_conn):
...
+            with conn.cursor() as cur:
+                cur.execute(f"SELECT to_regclass('{base}.pg_barrier_state')")
+                if cur.fetchone()[0] is not None:
+                    reachable = True
+                    if schema != base:
+                        cur.execute(f'CREATE SCHEMA IF NOT EXISTS "{schema}"')
+                        for table in QUEUE_TABLES:
+                            cur.execute(
+                                f'CREATE TABLE IF NOT EXISTS "{schema}".{table} '
+                                f"(LIKE {base}.{table} INCLUDING ALL)"
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Fixed in 3d142f30 — DDL now built via `psycopg2.sql` identifiers/literals instead of f-strings.

workers/tests/conftest.py
```diff
@@ -200,6 +200,99 @@ def pg_client(pg_conn):
...
+                            )
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Fixed in 3d142f30 — the per-worker schema is now dropped (CASCADE) and recreated each run, so a reused local DB can't keep a stale clone.

pk-zipstack (approved):
LGTM

PR #2209, opened 2026-07-25
[FIX] Seed challenge_llm from the exported tool's resolved value so challenge-enabled deploys pass

hari-kuriakose:
## Purpose

A tool exported with **LLMChallenge enabled** fails at the final `deployment run` with `422 Unprocessable Entity: Tool validation failed`, even though export has already resolved a valid challenge LLM.

This is the challenge-**enabled** counterpart to **#2203**, which unblocked the challenge-**off** case by omitting the adapter enum when the feature is disabled. With challenge **on** the enum is present and required, so the seeding bug still bites.

## The chain

1. Export resolves a real `challenge_llm` — falling back to the default profile's LLM when the project set none — and stores it under `tool_metadata[tool_settings]` (`prompt_studio_registry_helper.py:285`).
2. Tool-instance creation seeds metadata from `get_default_settings`, which walks the **spec**: a `"type": "string"` property with no `default` is seeded `""`. `challenge_llm` has no spec default, so the instance stores `""` — discarding the resolved value.
3. Because `challenge_llm` declares `adapterType: "LLM"`, `_update_schema_for_adapter_type` injects `enum: [<real adapter ids>]` into the tool-instance schema.
4. Deployment validation rejects `""` — an **enum** violation.

**There is a second half**, surfaced in review, and it is why seeding alone was not enough. `ToolInstanceViewSet.create` calls `ToolInstanceHelper.update_metadata_with_default_values` immediately after `perform_create` (`views.py:141-143`). `challenge_llm` declares `adapterType: "LLM"`, so it is one of the properties `update_metadata_with_default_adapter` walks — where two **unconditional** assignments (`tool_instance_helper.py:214-219`) replaced the seeded value with the user's default LLM.

So the seed only survived for users with **no** default LLM adapter. Both halves are fixed here.

## The fix

- **New** `PromptStudioRegistryHelper.get_resolved_settings(prompt_registry_id)` reads `tool_metadata[tool_settings]`, returning `{}` when the row or the block is absent.
- Tool-instance `create()` overlays the resolved `challenge_llm` — and its companion `challenge_llm_adapter_id` — onto the spec-seeded defaults.
- `update_metadata_with_default_adapter` now **fills gaps only**: a key already carrying a value resolved for that specific tool is preserved rather than overwritten with the user's global default. The preserved value is mirrored into the companion ID key so the pair is never half-written.

## Scope

**The overlay is deliberately restricted to `challenge_llm` alone.**

An earlier revision iterated every spec-seeded key, which silently widened the change: `enable_challenge`, `summarize_as_source`, `enable_highlight` and `enable_word_confidence` are all *both* spec properties and export keys, so all four were being overlaid. That would make every new Prompt Studio tool instance inherit the project's real challenge / summarized-extraction / highlighting settings instead of `False` — cost-, latency- and output-affecting behaviour, and a separate decision from fixing a validation failure. It is now scoped explicitly to the one key that causes the 422.

This also makes the `single_pass_extraction_mode` ↔ `enable_single_pass_extraction` key-name mismatch moot, rather than leaving a silent 4-of-5 asymmetry where single-pass alone stays stale.

`get_resolved_settings` returns `{}` for a non-UUID or missing registry row, so regular registry tools and cloud agentic tools are a no-op.

## The registry lookup

- **Screens non-UUID ids first.** `ToolInstance.tool_id` is a free-form `CharField` and registry tools use slugs (`"text_extractor"`), which a UUID pk lookup rejects — so the common non-Prompt-Studio path was logging a WARNING on *every* tool-instance creation. Checked inline rather than via `ToolInstanceHelper.is_uuid_format`, because `tool_instance_v2.tool_processor` already imports this module at module level and importing back would cycle.
- **Narrowed `except Exception` to `DoesNotExist`.** A swallowed `OperationalError` would seed `""` and resurface later as the same confusing 422 this PR exists to prevent; `ATOMIC_REQUESTS` is `False`, so `create()` still commits. Verified against the real queryset that `values_list(...).get()` raises the model's own `DoesNotExist`.
- **Fetches via `values_list("tool_metadata", flat=True)`**, so the largest JSONB column on the row is not pulled a second time — the caller has already loaded this row to build the `Tool`.
- Remaining log demoted to `debug`.

## Explicitly not in scope

- A resolved `challenge_llm` id only validates if the **deploying** user has access to that adapter (the enum is built per-user). Cross-user deploy is a pre-existing residual, not introduced by this change.

## Tests

`tool_instance_v2/tests/test_challenge_llm_seed_overlay.py` — 16 tests in the unit tier, no DB.

An earlier revision of this file used module-level `sys.modules` stubbing. That was wrong: it replaced `unstract.*` for the whole pytest process at collection time, and the repo had deliberately removed the pattern in `9a76a745` and `b60dd6f3`. It is rewritten to import the real modules and patch per-test, matching `test_build_index_payload.py` as it exists today. `git grep 'sys.modules\[' -- 'backend/**/test_*.py'` is back to **zero** hits, and `LogLevel` / `unstract.sdk1.adapters` both import fine after the test module loads — the two the review showed breaking.

Coverage:

- The `""` seed is replaced by the real adapter id, and the companion `challenge_llm_adapter_id` is written alongside it.
- **Wiring**: `create()` is asserted to invoke the overlay. The previous tests bound `resolved_settings` themselves, so deleting the call from `create()` left them all green — that gap is closed.
- **Clobber**: a seeded value survives the default-adapter walk, while an unset one still receives the user's default (the guard fills gaps, it does not disable defaults).
- **Half-written pair**: preserving a value must not strand a missing ID key.
- Scope guard: the four other spec/export overlaps stay at their spec defaults, and non-spec export keys never reach instance metadata.
- Registry lookup: reads the `tool_settings` block specifically, short-circuits a slug without querying, and does **not** swallow `OperationalError`.

Verified by mutation — removing the overlay call, the no-stomp guard, the `setdefault`, the companion adapter-id write, re-widening the overlay, and restoring `except Exception` each fail a test.

Full `unit-backend` suite passes locally: **293 passed, 1 skipped**, with no collection-order damage. `ruff` and `ruff-format` pass at the version pinned in `.pre-commit-config.yaml` (v0.3.4).

## Related

- **#2203** — the challenge-**off** counterpart. Independent change; either can merge first.


backend/tool_instance_v2/serializers.py
```diff
@@ -170,11 +173,26 @@ def create(self, validated_data: dict[str, Any]) -> Any:
...
+        # already resolved a real value for it, so prefer that - otherwise the
+        # instance stores "", which fails deployment validation against the
+        # adapter enum. Only spec-declared keys are overlaid; the export's
+        # settings are a superset (llm, vector-db, ...) and the rest are not
+        # part of the instance schema. Non-Prompt-Studio tools resolve to {}
+        # here, making this a no-op for them.
+        resolved_settings = PromptStudioRegistryHelper.get_resolved_settings(
+            prompt_registry_id=str(tool_uid)
+        )
+        for key in tool_settings:
```
chandrasekharan-zipstack: **The overlay is overwritten a few lines later, so this is a no-op for most users.**

`ToolInstanceViewSet.create` calls `ToolInstanceHelper.update_metadata_with_default_values(instance, user)` immediately after `perform_create` (`tool_instance_v2/views.py:141-143`). That walks the user's default adapters into `update_metadata_with_default_adapter`, which ends in two **unconditional** assignments (`tool_instance_helper.py:214-219`):

```python
for adapter_key, adapter_property in properties.items():
    metadata[adapter_key] = str(adapter.id)
    metadata[metadata_key_for_id] = str(adapter.id)
```

`properties` is `schema_spec.get_llm_adapter_properties()`, which **includes `challenge_llm`** (it declares `adapterType: "LLM"` on this branch). So whenever the user has a `UserDefaultAdapter.default_llm_adapter`, the value overlaid here is clobbered with their default LLM id.

Corollary worth stating explicitly: the reported 422 can only occur for users with **no** default LLM adapter — which is exactly the population this PR helps, and the case the tests should pin. It also suggests the cleaner fix is in `update_metadata_with_default_adapter` (don't stomp an already-populated adapter key), which would subsume this and interacts with #2203.

Separately: the overlay sets `challenge_llm` but never its companion `challenge_llm_adapter_id` (not a spec property, so not in `get_default_settings`, and not in the export's `tool_settings`). Every other code path sets the two together, so this produces a metadata shape nothing else does.

backend/tool_instance_v2/serializers.py
```diff
@@ -170,11 +173,26 @@ def create(self, validated_data: dict[str, Any]) -> Any:
         validated_data[TIKey.PK] = uuid.uuid4()
         # TODO: Use version from tool props
         validated_data[TIKey.VERSION] = ""
+        tool_settings = ToolProcessor.get_default_settings(tool)
```
chandrasekharan-zipstack, line 279: **The overlay touches five keys, not one — and the description says the opposite.**

`ToolUtils.get_default_settings` returns exactly `tool.spec.properties.keys()` (`tool_registry/tool_utils.py:124-143`), so the iteration set is spec ∩ export:

| spec key | overlaid? |
|---|---|
| `challenge_llm` | yes |
| `enable_challenge` | **yes** |
| `summarize_as_source` | **yes** |
| `enable_highlight` | **yes** |
| `enable_word_confidence` | **yes** |
| `single_pass_extraction_mode` | no — export writes `enable_single_pass_extraction` (`constants.py:95`) |

The description says *"the instance still seeds `enable_challenge` from the spec default (`False`) rather than the tool's real setting … noted so it isn't assumed fixed."* That isn't right — `JsonSchemaKey.ENABLE_CHALLENGE == "enable_challenge"` (`constants.py:93`) is both a spec property (`prompt_studio_registry_helper.py:58`) and an export key (`:287`), so it **is** overlaid.

So every newly created Prompt Studio tool instance now inherits the project's real LLMChallenge / summarized-extraction / highlighting / word-confidence settings instead of `False`. That may well be the behaviour you want, but it is cost-, latency- and output-affecting, and right now it's undisclosed and untested (the tests only exercise `challenge_llm` plus a passive `enable_highlight: False`).

Also note the asymmetry: single-pass silently stays stale because of the key-name mismatch, so a project exported with all five enabled yields an instance with 4-of-5 real. Either own the wider scope with tests and fix the single-pass name, or restrict the overlay to `{challenge_llm}` explicitly.

backend/prompt_studio/prompt_studio_registry_v2/prompt_studio_registry_helper.py
```diff
@@ -143,6 +143,31 @@ def get_tool_by_prompt_registry_id(
...
+        ``tool_metadata[tool_settings]``. ``Tool`` (built from ``tool_spec`` /
+        ``tool_property``) does not carry them, so callers that only have a
+        ``Tool`` cannot see the resolved values.
+
+        Returns an empty dict when the registry row is missing or carries no
+        settings, so callers can treat "no resolved settings" as a no-op.
+        """
+        try:
+            prompt_registry_tool = PromptStudioRegistry.objects.get(pk=prompt_registry_id)
+        except Exception as e:
```
chandrasekharan-zipstack: Three issues with this lookup:

1. **`except Exception` swallows real DB errors.** An `OperationalError` here returns `{}` → seeds `""` → surfaces much later as the same confusing 422 this PR is fixing. `ATOMIC_REQUESTS` is `False` (`settings/base.py:258-259`), so `super().create()` can still succeed after a swallowed read error. Narrow to `PromptStudioRegistry.DoesNotExist` (+ `ValidationError`).

2. **It logs a WARNING on a fully normal path.** `ToolInstance.tool_id` is a free-form `CharField` (`tool_instance_v2/models.py:35`), and registry tools use slugs — `"classify"`, `"text_extractor"`. `prompt_registry_id` is a `UUIDField` pk, so `UUIDField.to_python("classify")` raises `ValidationError: "classify" is not a valid UUID` → caught → warning. Cloud agentic tools get a second one (they already log at `:132`). `ToolInstanceHelper.is_uuid_format` already exists and is used in the calling file at `serializers.py:99` — guard on it and demote the remaining log to `debug`.

3. **Duplicate fetch of the same row.** `serializers.py:168` already pulled this row via `ToolProcessor.get_tool_by_uid` → `get_tool_by_prompt_registry_id` → `PromptStudioRegistry.objects.get(pk=…)` (`:129`). This re-issues the identical query, and `tool_metadata` is the full JSONB blob of every prompt and output — the largest column on the row — so it gets pulled twice. Either return the metadata alongside the `Tool` from the existing fetch, or use `.values_list("tool_metadata", flat=True)`.

backend/tool_instance_v2/tests/test_challenge_llm_seed_overlay.py
```diff
@@ -0,0 +1,352 @@
...
+    sys.modules["unstract.tool_registry.constants"] = constants
+    dto_spec = importlib.util.spec_from_file_location(
+        "unstract.tool_registry.dto", DTO_PATH / "dto.py"
+    )
+    dto = importlib.util.module_from_spec(dto_spec)
+    dto_spec.loader.exec_module(dto)
+    return dto.Spec
+
+
+SPEC_CLS = _load_spec_class()
```
chandrasekharan-zipstack: **This reintroduces a pattern the repo deliberately removed — and cites as precedent the very file that was cleaned of it.**

`SPEC_CLS = _load_spec_class()` runs at **module import** time and `_stub_package()`s `unstract`, `unstract.sdk1`, `unstract.sdk1.constants`, `unstract.tool_registry` into `sys.modules`, each with `__path__ = []`. pytest imports all test modules during collection before running anything, so this replaces the real modules for the whole process, ahead of every other backend test. Reproduced in `backend/.venv`:

```
REAL LogLevel present: True
AFTER STUB FAILS: ImportError cannot import name 'LogLevel' from 'unstract.sdk1.constants'
adapters FAILS: ModuleNotFoundError No module named 'unstract.sdk1.adapters'
```

Production code does lazy `unstract.*` imports inside functions (`prompt_studio_core_v2/internal_views.py:412-415`, `workflow_manager/workflow_v2/views.py:1292`), so any unit test reaching those breaks. Green CI is collection-order luck; under xdist it's order-dependent flakiness.

The history:

- **`9a76a745`** (Jul 7) rewrote `test_build_index_payload.py` from 421 lines to 50, deleting ~30 stub modules: *"the clobber/restore dance was import-order dependent and silently skipped all tests when the stub surface drifted."*
- **`b60dd6f3` / #2202** (Jul 24) removed the last remaining one from `test_execution_serializer.py`, after it *"shadowed the real module for every later test on that worker"* under xdist — 30 unrelated integration failures. That commit **deleted** the stub rather than restoring it in a `finally`, explicitly *"so the pattern can't be copied."*

`git grep 'sys.modules\[' -- 'backend/**/test_*.py'` on `main` returns **zero** hits today. This PR and #2203 would be the first to bring it back.

The file's stated precedent — *"matching the existing `test_build_index_payload.py` approach"* — is the post-`9a76a745` file, which does the opposite: it imports the real module and patches with `ExitStack`. And the stated rationale (*Django not importable in a plain checkout*) is contradicted by `pytest-django>=4.12.0` in `backend/pyproject.toml:76` and the `unit-backend` rig group setting `DJANGO_SETTINGS_MODULE=backend.settings.test` over `paths: ["."]` (`tests/groups.yaml:51-76`).

There's also a coverage hole the style creates: the test binds `resolved_settings` itself (`:172-176`), so deleting the `get_resolved_settings(...)` call at `serializers.py:185-186` while keeping the loop leaves all six tests green. The mutation check covered the loop, not the wiring.

Suggest rewriting as an import-and-`mock.patch` test of the real `create()` — that removes the pollution, the string-anchor brittleness, and the wiring gap in one go. (`tests/__init__.py` is fine and matches convention; no issue there.)

chandrasekharan-zipstack (commented):
Three things to resolve before this lands: the overlay is discarded downstream in the common case, its scope is wider than the description says, and the test file reintroduces module-level `sys.modules` stubbing that the repo removed three days ago.

One claim in the description that does check out: the circular-import worry is a non-issue — `tool_processor.py:7-10` already imports `PromptStudioRegistryHelper` at module level, so the `tool_instance_v2 → prompt_studio_registry_helper` edge predates this PR.

backend/tool_instance_v2/serializers.py
hari-kuriakose: Confirmed and fixed — this was the important one. The overlay was dead code for anyone with a default LLM adapter, which is most users.

Fixed where you suggested, in `update_metadata_with_default_adapter`: it now fills gaps only, so a value already resolved for that specific tool survives the walk.

I checked the precondition that makes changing the shared helper safe: `update_metadata_with_default_values` has exactly one call site (`views.py:142`, right after `perform_create`), and `update_metadata_with_default_adapter` has exactly one caller (the loop inside it). Nothing uses either as "reset to the user's defaults", so guarding the write doesn't change any other path.

On the companion key — also fixed, and it turned out to cut both ways. The overlay now writes `challenge_llm_adapter_id` alongside `challenge_llm`, reading the key from the spec's `adapterIdKey` rather than hardcoding it. But my first version of the guard `continue`d past the *whole* write when the adapter key was populated, which stranded the ID key and produced the same half-written shape from the other direction. Reproduced against the real helper before fixing:

```
result: {'challenge_llm': '11111111-2222-3333-4444-555555555555'}
companion id key present? False
```

That's precisely what `update_metadata_with_adapter_properties:120-122` warns about ("stale UUID-shaped values at adapter_key bypass the lazy migrator and fail schema enum validation later"). The guard now mirrors the preserved value into the ID key via `setdefault`, so the pair is consistent whichever path populated it. Pinned by `test_kept_value_never_leaves_a_half_written_pair`.

backend/tool_instance_v2/serializers.py
hari-kuriakose: You're right on both counts, and the description was wrong. `JsonSchemaKey.ENABLE_CHALLENGE == "enable_challenge"` is both a spec property and an export key, so it *was* being overlaid — the "noted so it isn't assumed fixed" line claimed the opposite of what the code did. Your table is accurate.

I took the second option: **restricted the overlay to `challenge_llm` explicitly.**

Reasoning — the other four are cost-, latency- and output-affecting. Making every new instance inherit the project's real challenge / summarized-extraction / highlighting settings may well be desirable, but it's a product decision, not a side effect of fixing a validation failure, and it doesn't belong in a bugfix PR under a title about deploy validation. Narrowing also dissolves the asymmetry you point out: with single-pass out of scope there's no 4-of-5 to explain, and CodeRabbit's separate `enable_single_pass_extraction` alias finding becomes moot rather than a second thing to get right.

The overlay is now a named helper with the scope decision stated in its docstring, and `test_only_challenge_llm_is_overlaid` pins it: the export fixture deliberately carries `enable_challenge: True`, `summarize_as_source: True` and `enable_highlight: True`, and the test asserts the instance still gets the spec defaults. Re-widening the loop fails it.

Description rewritten to describe what actually ships.

backend/prompt_studio/prompt_studio_registry_v2/prompt_studio_registry_helper.py
hari-kuriakose: All three fixed.

**1. Narrowed to `DoesNotExist`.** Agreed on the failure mode — a swallowed `OperationalError` returning `{}` seeds `""` and resurfaces as the same 422 this PR exists to prevent, and with `ATOMIC_REQUESTS = False` the create still commits. I verified the exception type against the real queryset before narrowing, since I changed the query shape in the same edit:

```
queryset type: BaseModelQuerySet
raises DoesNotExist? True
```

`values_list(...).get()` raises the model's own `DoesNotExist`, so a missing row can't become an unhandled 500. `test_db_errors_are_not_swallowed` pins that `OperationalError` propagates.

**2. Guarded, and the remaining log demoted to `debug`.** One deviation from your suggestion worth flagging: I did *not* use `ToolInstanceHelper.is_uuid_format`, even though it exists and is used in the calling file. `tool_instance_v2.tool_processor` imports this module at module level, so importing back into `tool_instance_v2` from here would cycle. Inlined the `uuid.UUID()` check instead, with that reason in a comment so the next person doesn't "clean it up" into an import. `test_non_uuid_tool_id_short_circuits_without_a_query` asserts the slug path never touches `objects`.

**3. Duplicate fetch — now `values_list("tool_metadata", flat=True)`.** I went with this over threading metadata out of `get_tool_by_prompt_registry_id`, since changing that function's return type ripples into every caller for no gain here. The row is still read twice, but the second read no longer drags the full JSONB blob.

backend/tool_instance_v2/tests/test_challenge_llm_seed_overlay.py
hari-kuriakose: You're right, and the rationale I gave for it was false. Rewritten as an import-and-patch test of the real modules.

The stated reason — "Django is not importable in a plain checkout" — doesn't survive contact with `tests/groups.yaml:51-76`, which sets `DJANGO_SETTINGS_MODULE=backend.settings.test` for `unit-backend` over `paths: ["."]`. I ran the real imports under that env: they load fine and the tests touch no database. The stubbing was never necessary, so it was pure cost. And citing `test_build_index_payload.py` was doubly wrong given `9a76a745` is the commit that removed exactly this from it.

Verification against the two things you reproduced:

```
git grep 'sys.modules[' -- 'backend/**/test_*.py'   ->  0 hits
LogLevel importable AFTER test module import: True
unstract.sdk1.adapters importable: True
```

**The wiring hole is closed too** — that was the sharpest part of the finding. `TestOverlayIsWiredIntoCreate::test_create_invokes_the_overlay` patches the overlay and asserts `create` calls it with the right tool id. I re-ran the mutation you described: deleting the `_overlay_resolved_challenge_llm(...)` call from `create()` while leaving the helper intact now fails 1 test, where before it left all six green.

Full mutation set, each failing at least one test: remove the overlay call; remove the no-stomp guard; drop the `setdefault` that keeps the adapter pair consistent; drop the companion adapter-id write; re-widen the overlay to all spec keys; restore `except Exception`.

16 tests, no DB. Full `unit-backend` passes locally at **293 passed, 1 skipped** — I ran the whole tier rather than just this file, since the failure mode you described is cross-test.

One gap I can't close locally and would rather state than paper over: neither `pytest-xdist` nor `pytest-randomly` is installed in my venv, so I could not exercise the multi-worker ordering directly. The argument that it's safe is structural rather than empirical — with zero `sys.modules` mutation there is no cross-worker state left to leak. Worth a CI eye if this branch runs under xdist.

backend/prompt_studio/prompt_studio_registry_v2/prompt_studio_registry_helper.py
```diff
@@ -143,6 +143,31 @@ def get_tool_by_prompt_registry_id(
...
+            logger.warning(
+                f"Error while fetching resolved settings for prompt registry ID "
+                f"{prompt_registry_id}: {e}"
+            )
+            return {}
+        metadata = prompt_registry_tool.tool_metadata or {}
+        return metadata.get(JsonSchemaKey.TOOL_SETTINGS, {}) or {}
+
```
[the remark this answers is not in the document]
hari-kuriakose: Valid — fixed. The warning-per-creation was real: `ToolInstance.tool_id` is a free-form `CharField` and registry tools use slugs, so every built-in tool creation hit the `except` and logged.

Implemented the short-circuit as suggested, with two adjustments:

- The check is inline rather than via `ToolInstanceHelper.is_uuid_format` (which a human reviewer suggested): `tool_instance_v2.tool_processor` imports this module at module level, so importing back would create a cycle.
- The remaining `except` is narrowed to `PromptStudioRegistry.DoesNotExist` and its log demoted to `debug` — the blind `except Exception` you flagged via BLE001 would swallow an `OperationalError` into `{}`, seeding `""` and resurfacing as the same 422 this PR fixes.

Your note about the old test still passing was accurate, and points at a real hole — it only exercised the DB-lookup branch. There's now a test asserting the slug path never touches `objects` at all.

backend/tool_instance_v2/serializers.py
```diff
@@ -170,11 +173,26 @@ def create(self, validated_data: dict[str, Any]) -> Any:
...
+        )
+        for key in tool_settings:
+            if key in resolved_settings:
+                tool_settings[key] = resolved_settings[key]
         validated_data[TIKey.METADATA] = {
             # TODO: Review and remove tool instance ID
             WorkflowKey.WF_TOOL_INSTANCE_ID: str(validated_data[TIKey.PK]),
             PromptStudioRegistryKeys.PROMPT_REGISTRY_ID: str(tool_uid),
-            **ToolProcessor.get_default_settings(tool),
+            **tool_settings,
```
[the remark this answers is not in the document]
hari-kuriakose: The key-name mismatch you identified is real — `single_pass_extraction_mode` (spec) vs `enable_single_pass_extraction` (export, `constants.py:95`) — but I resolved it the other way, so the alias map is **not** implemented.

The overlay is now restricted to `challenge_llm` alone. The wider loop was overlaying four keys beyond the one this PR is about (`enable_challenge`, `summarize_as_source`, `enable_highlight`, `enable_word_confidence`), which would change whether challenge, summarization and highlighting actually *run* on new instances — cost-, latency- and output-affecting behaviour that shouldn't ride along in a fix for a validation failure.

With single-pass out of scope, there's no gap to close: the setting keeps its existing spec-default behaviour rather than being half-migrated. Adding the alias would have made the overlay *more* correct in isolation while widening undisclosed behaviour change — the asymmetry you spotted was a symptom of the scope being too wide, not of a missing mapping.

If the team later wants exported instances to inherit all their real settings, the alias map is the right fix and this comment is the reference for it.

hari-kuriakose:
Pushed `cbab0eef5`. All four review findings and both CodeRabbit findings are addressed, with per-thread replies.

**What changed materially:**

1. **The overlay was a no-op for most users.** `update_metadata_with_default_adapter` clobbered the seeded `challenge_llm` with the user's default LLM right after creation, so the fix only worked for users with no default adapter. Guarded to fill gaps only — verified first that it has a single call site, so nothing relies on the overwrite.
2. **Scope narrowed to `challenge_llm`.** It was silently overlaying four more keys and changing whether challenge / summarization / highlighting actually run. The description's `enable_challenge` claim was wrong; it's rewritten.
3. **Registry lookup** — non-UUID short-circuit (no more warning on every built-in tool creation), `except` narrowed to `DoesNotExist`, `values_list` so the big JSONB column isn't pulled twice.
4. **Tests rewritten** to import the real modules and patch per-test. Zero `sys.modules` mutation, and the wiring gap is closed — deleting the overlay call from `create()` now fails a test.

**Verification:** 16 tests, no DB. Six mutations each fail at least one test. Full `unit-backend` passes locally at 293 passed / 1 skipped. `ruff` + `ruff-format` clean at the pinned v0.3.4.

One thing I couldn't verify locally, stated rather than glossed: xdist isn't installed in my venv, so the multi-worker ordering argument is structural (no `sys.modules` mutation left to leak) rather than empirical.

chandrasekharan-zipstack (approved):

PR #2211, opened 2026-07-26
[P0+P1] shadcn/ui foundation + lucide icon migration

hari-kuriakose:
Implements **P0-01 … P0-16** of `UN_SHADCN_IMPL_PLAN.md` (spec: `UN_SHADCN_SPEC.md`, palette decision **D8** = Midnight Bloom).

**This phase is intentionally a visual no-op.** antd still renders every screen; this PR only installs the shadcn/ui stack alongside it (spec §7 coexistence). No screen layout or IA changes (C4 / D10).

## What's here

- **Deps** — Radix primitives, CVA/clsx/tailwind-merge, lucide-react, next-themes, sonner, react-hook-form + zod, Tailwind v4. `antd` deliberately retained until P4.
- **Fonts** — self-hosted `@fontsource` Inter + Geist Mono. No CDN: prod serves through nginx and must not depend on an external host.
- **Tokens** — `src/index.css` carries the Midnight Bloom light + dark palette verbatim. Type/spacing/radius scales are *not* declared: they were verified identical to Tailwind v4 stock, and hardcoding them in px would lose rem-based user-font-size scaling.
- **32 primitives** in `src/components/ui/`, plus hand-written `spinner` + `kbd` and `success`/`warning` badge variants.
- **Theme** — `next-themes` mirrors the existing session theme onto `.dark`. Persistence and the toggle location are unchanged.
- **Toasts** — sonner mounted; shared `useAppToast` helper for cloud plugins to import (D9). `ALERT_SURFACE` keeps antd as the single active surface until P2-06, so alerts aren't double-rendered.

## Gates

| Gate | Result |
|---|---|
| P0-G1 build (plugins absent) | ✅ `src/plugins/` is absent by design in an OSS checkout, so this exercises the `optionalPluginImports` path |
| P0-G3 dark mode | ✅ see below |
| P0-G4 lint | ✅ no new findings |
| P0-G5 tests | ✅ 4 files / 16 tests |
| P0-G6 no visual regression | ✅ see below |

**Dark mode** — the load-bearing check is that Tailwind *utilities* flip, not just CSS variables:

| | light | dark |
|---|---|---|
| `bg-background` | `rgb(250,250,250)` | `rgb(26,26,26)` |
| `text-foreground` | `rgb(26,26,26)` | `rgb(250,250,250)` |

A plain `@theme` would have frozen these at the light value — that's the trap in spec §2.5.3a.

**No visual regression** — antd element count (33), button count (5), height (50px), background `rgb(47,147,246)`, radius (6px) and page background are all identical before/after. Only the body font moves to Inter, which is P0-12's intent.

## Two fixes the plan didn't anticipate

1. **`.gitignore`** — the Python `lib/` rule also matched `frontend/src/lib/`, where `components.json` points `@/lib/utils`. Without the negation, `cn()` would never reach the repo and every primitive would fail to resolve in CI. Worth a sanity check from a reviewer.
2. **`biome.json`** — `css.parser.tailwindDirectives: true`. Biome can't parse `@theme`/`@plugin`/`@custom-variant` without it and fails CI with 4 parse errors.

## Pre-existing lint findings

`biome ci src/` reports 3 errors / 24 warnings, all in files this PR never touches. Pristine `main` reports **227 errors / 261 warnings** with the same binary.

## Deferred deliberately

- `body { overflow: hidden }` left as-is; it interacts with Radix's dialog scroll-lock and is re-tested at the first P2 dialog conversion.
- P1–P4 and the cloud plugin phase not started, per the 🛑 STOP gate after P0.


hari-kuriakose:
### Scope note — this PR now covers P0 **and** P1-01/P1-02

The plan's topology put P0 on `feat/shadcn-foundation` (W1) and P1–P4 on `feat/shadcn-oss-migration` (W2) so the foundation could be reviewed on its own. I pushed the P1 icon commit to the foundation branch by mistake, and separating them after the fact would require a force-push, which I'm not doing to an already-pushed branch.

So this PR carries **two reviewable commits**:

- `a0473a1` **[P0]** foundation — deps, tokens, 32 primitives, theme provider, toast bridge. Zero visual change.
- `26a8c64` **[P1-01/P1-02]** icon migration — 91 files, 87 icons, `@ant-design/icons` → `lucide-react`.

They're cleanly separated by commit, so `git show a0473a1` reviews the foundation in isolation. `feat/shadcn-oss-migration` exists and points at the same head; P1-03 onward will land there.

### P1 highlights worth a look

**Non-exact icon mappings** (full table in `frontend/docs/icon-map.md`):
- `CheckCircleFilled` / `PlayCircleFilled` / `InfoCircleFilled` → lucide has **no filled variants**; these now render as outlines.
- `MoreOutlined` → `EllipsisVertical` (antd's is vertical; plain `Ellipsis` is horizontal).
- `SlackOutlined` → `MessagesSquare` — lucide **dropped brand icons**, so the Slack mark is gone.
- `CaretDownOutlined` → `ChevronDown` (solid triangle → stroke).

**Three collisions the rename introduced**, two of them more than cosmetic:
- `FileUpload.jsx` / `FileWidget.jsx` import antd's `Upload` **component**, shadowed by the lucide `Upload` icon — would have broken both upload widgets.
- `Workflows.jsx` defines its own `User` component; the lucide import made it render itself (infinite recursion, caught by the build).

**One follow-on fix:** `useRetrievalStrategies.js` emits icon name strings that `RetrievalStrategyModal`'s `ICON_MAP` consumes. The modal's keys became lucide names while the hook still emitted antd names — every lookup would have missed and silently fallen back to a default icon. Backend contract unchanged.

### Plan estimates are running low

Plan said "162 files / 43 icons"; actual was **91 files / 87 icons** — the original enumeration used a single-line regex that missed multi-line imports. Typography (P1-03) is similarly **295 sites, not 158**. Re-measuring each phase before starting it.

### P1-03 is paused pending a decision

`Typography` can't be regex-converted: 10+ call-sites use `ellipsis={{ tooltip, rows }}`, which does JS measurement and hover tooltips. Tailwind's `truncate` is CSS-only, so a naive swap would silently drop behavior — a regression, which C4 forbids. Recommended approach is a small `@/components/ui/typography.jsx` shim presenting antd's API on shadcn tokens, turning 295 sites into a mechanical import swap. Flagging before building it.

hari-kuriakose:
Closing this as superseded by #2212.

`feat/shadcn-foundation` is a strict ancestor of `feat/shadcn-oss-migration`: both of its commits are already contained in #2212, and there is nothing on this branch that is not on that one.

```
commits vs main — foundation:     2
commits vs main — oss-migration:  92  (includes those 2)
in foundation but NOT in oss-migration: 0
```

Both PRs target `main` directly rather than being stacked, so merging this one first would only shrink #2212's diff — it would not change the final state of `main`. Keeping it open mainly risks someone reviewing 163 files that are a subset of the other PR.

All subsequent work — the Transfer rewrite, Tabs pane keys, the app-level confirm dialog host, sidebar fly-out fixes, prompt-card icons — is on `feat/shadcn-oss-migration` only.

Please review #2212 instead.

PR #2215, opened 2026-07-27
UN-3794 [MISC] Remove x2text-service from the stack

athul-rs:
## What

Stops shipping `x2text-service`: removed from `docker-compose.yaml`, the build and override compose files, the CI build matrix and cache config, and the e2e readiness check. Adds a test that fails if it comes back.

## Why

`x2text-service` is a thin HTTP proxy in front of unstructured.io. Only two adapters route through it — `unstructured_community` and `unstructured_enterprise` (via `UnstructuredHelper.make_request`). LLMWhisperer, llama_parse and no_op do not touch it. Both adapters are sunset, so this is a container we build, publish, run and health-check for nothing.

## How

- Dropped the service block and its `3004:3004` publish from `docker/docker-compose.yaml`, plus the backend's `depends_on` entry.
- Dropped the build entries from `docker/docker-compose.build.yaml` and `docker/sample.compose.override.yaml`.
- Dropped it from the `production-build.yaml` build matrix and its summary loop, and from the `ci-test.yaml` build cache config.
- Dropped the readiness probe from `tests/rig/runtime.py`, so e2e no longer waits on a service that is not started.

Left in place deliberately:

- **`X2TEXT_HOST` / `X2TEXT_PORT` stay in the sample envs.** `ToolsUtils.__init__` reads both with `raise_exception=True` on *every* workflow execution, not only for the two sunset adapters. Removing them breaks all execution, not just the paths being retired.
- **`x2text-service/`, its Dockerfile and the two adapters stay.** Deleting them strands configured `AdapterInstance` rows of those types, which makes it a migration decision rather than a packaging one. Worth a separate ticket owned by whoever sequences the sunset.

Yes, narrowly and by design: any deployment still using an `unstructured_community` or `unstructured_enterprise` adapter loses text extraction through it, because nothing will be listening on 3004. That is the intent — both adapters are sunset. No other extraction path is affected; LLMWhisperer and the rest never used the service.

Two operational notes:

- Existing installs keep the container until it is explicitly removed — `docker compose up` does not delete a service that is gone from the file without `--remove-orphans`. Redeploys need to drop the container and unpublish 3004 rather than assume the file change did it.
- `UNSTRACT_X2TEXT_URL` is still threaded through the rig config; it is now unused by the readiness check. Left alone to avoid churning `tests/rig/tests/test_runtime.py` in this PR.

## Database Migrations

None.

## Env Config

No removals. `X2TEXT_HOST` / `X2TEXT_PORT` intentionally retained — see above.

## Relevant Docs

`docs/local-dev-setup-executor-migration.md` still lists the service on port 3004 and describes it as needed for text extraction. Not updated here to keep this PR to packaging; worth a follow-up.

## Related Issues or PRs

UN-3794

## Dependencies Versions

Unchanged.

## Notes on Testing

- `tests/rig/tests/test_x2text_decommissioned.py` asserts the service name is absent from every file that would deploy, build or publish it, and that 3004 is no longer published. Confirmed to fail against `main` before this change, so a partial revert fails rather than quietly reintroducing the container.
- `docker compose config` validates clean across the merged compose files, with no `x2text` or `3004` in the rendered output.

Manual check worth doing before merge: bring the stack up and confirm no `unstract-x2text-service` container, the backend healthy without the `depends_on`, port 3004 refused, and one non-unstructured extraction path (LLMWhisperer) still working end to end.

## Checklist

### DIFF
--- a/.github/workflows/ci-test.yaml
+++ b/.github/workflows/ci-test.yaml
@@ -206,8 +206,6 @@ jobs:
             tool-sidecar.cache-to=type=gha,mode=max,scope=tool-sidecar
             platform-service.cache-from=type=gha,scope=platform-service
             platform-service.cache-to=type=gha,mode=max,scope=platform-service
-            x2text-service.cache-from=type=gha,scope=x2text-service
-            x2text-service.cache-to=type=gha,mode=max,scope=x2text-service
             tool-classifier.cache-from=type=gha,scope=tool-classifier
             tool-classifier.cache-to=type=gha,mode=max,scope=tool-classifier
             tool-text_extractor.cache-from=type=gha,scope=tool-text_extractor
--- a/.github/workflows/production-build.yaml
+++ b/.github/workflows/production-build.yaml
@@ -103,7 +103,6 @@ jobs:
             platform-service,
             runner,
             worker-unified,
-            x2text-service,
           ]
 
     steps:
@@ -315,7 +314,7 @@ jobs:
             echo "|---------|--------|" >> $GITHUB_STEP_SUMMARY
 
             # Define services in order
-            for service in backend frontend platform-service runner worker-unified x2text-service; do
+            for service in backend frontend platform-service runner worker-unified; do
               if [ -f "build-status/${service}.json" ]; then
                 STATUS=$(jq -r '.status' "build-status/${service}.json")
                 if [ "$STATUS" = "success" ]; then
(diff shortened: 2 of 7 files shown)


athul-rs:
Closing for now — holding the x2text-service removal until it is explicitly agreed. Nothing else was bundled in this PR; the work is preserved locally and can be reopened unchanged.

PR #2217, opened 2026-07-28
UN-3445 [GATED-FEAT] PG-queue auxiliary-worker migration (OSS): loader fix + notification seam

muhammad-ali-e:
## What

The OSS slice of the PG-queue **auxiliary-worker migration** wave (epic UN-3445). Two flag-gated changes, both inert with `pg_queue_enabled` off:

- **UN-3798** — fix the PG pluggable-worker task loader (`workers/worker.py`): skip the broken file-path task load for pluggable workers (their tasks are already registered by `build_celery_app`'s package import). Non-pluggable workers keep the file-path load. Fixes a startup crash-loop for cloud PG pluggable workers; the Celery path is untouched.
- **UN-3753** — route the buffered webhook-notification dispatch through the PG-queue transport seam (`notification_v2/notification_dispatch.py`): PG when `pg_queue_enabled` for the org, else `celery_app.send_task` (byte-identical). `PermanentDispatchError` (raised only on the PG branch) dead-letters permanent enqueue errors; the Celery error path is unchanged.

## Why

- Part of migrating the remaining dispatch/auxiliary workers off Celery onto the PG queue so Celery can eventually be decommissioned (sub-tasks of UN-3445).

- No. Both changes are gated by the `pg_queue_enabled` Flipt flag; with it off (the production default) `resolve_transport` fail-closes to Celery and the seams call the same `send_task` as before — byte-identical. The loader fix only skips a load path for pluggable workers (not deployed with the flag off). Dev-tested on a live stack: flag-off produced 0 PG rows (Celery taken); PG round-trips delivered end-to-end.

## Notes on Testing

- `notification_v2/tests/` (12 tests) pass; `pipeline_dispatch` regression suite green; worker enum/loader tests pass. Merged current `main` in (clean, no conflicts).

## Related Issues or PRs

- UN-3445 (epic), UN-3798, UN-3753. Cloud counterpart PR covers UN-3752/3754/3779 + the chart. Flag-gated wave; merge-safe with the flag off.


workers/worker.py
```diff
@@ -440,40 +441,69 @@ def on_task_postrun(sender=None, task_id=None, **kwargs):
...
+
+load_worker_tasks(worker_type)
+
+# A worker that boots with no registered tasks starts but silently processes
+# nothing (Celery does not error on an empty registry). Surface that misconfig —
+# as a WARNING, not a hard failure: pluggable tasks bind on app finalize
+# (@shared_task / connect_on_app_finalize), which can be after this point, so a
+# raise here would false-positive on a correctly-configured pluggable worker.
+_registered_tasks = [name for name in app.tasks if not name.startswith("celery.")]
+if not _registered_tasks:
```
muhammad-ali-e: **[Medium — silent failure]** The empty-registry check only ever *warns*, but for a **non-pluggable** worker this state is terminal, not transient. `load_worker_tasks` loads a non-pluggable `tasks.py` synchronously above (or early-returns on a missing dir / `tasks.py` at L474 / L482), so an empty registry here has no late `app.finalize()` binding coming — the worker boots a healthy-looking Celery process with zero registered tasks, silently rejects every delivered message, and L509 still logs `✅ Successfully loaded`. To k8s the pod is Ready. The WARNING-not-raise is justified *only* for pluggable workers (whose `@shared_task`s bind on finalize).

Suggest splitting severity by `is_pluggable()`: warn for pluggable, `raise RuntimeError(...)` for non-pluggable — and likewise hard-fail the missing-dir / missing-`tasks.py` early-returns (L474 / L482) rather than `return`.

backend/notification_v2/notification_dispatch.py
```diff
@@ -0,0 +1,119 @@
...
+                task_name=WEBHOOK_NOTIFICATION_TASK,
+                queue=queue,
+                args=args,
+                kwargs=kwargs,
+                # enqueue_task's org_id is str-typed — coerce None→"" (unlike the
+                # routing arg above, this `or ""` is load-bearing).
+                org_id=org_string_id or "",
+                task_id=dispatch_id,
+            )
+        except (ValueError, TypeError) as exc:
```
muhammad-ali-e, line 159: **[Medium — error classification]** `except (ValueError, TypeError)` catches only Python-layer permanent failures, but a permanent **DB-layer** rejection is neither type. Concretely: `_json_safe` uses `json.dumps(default=str)`, which *permits* `NaN`/`Infinity`, so a webhook body carrying a `NaN` float round-trips un-sanitized and Postgres `JSONB` rejects it at insert with `django.db.DataError`. That is permanent (fails identically every flush tick) yet slips past this catch → the caller's generic `except Exception` reverts to PENDING and re-renders/re-dispatches until the attempt cap, logging a misleading `result=broker_failure` each tick — the exact loop `PermanentDispatchError` exists to prevent, just for DB-layer errors instead of Python-layer ones.

Fix options: `json.dumps(..., allow_nan=False)` inside `_json_safe` (so it raises `ValueError` at the intended seam), and/or widen the wrap to include `django.db.DataError`/`IntegrityError` while leaving `OperationalError`/`InterfaceError` (genuinely transient) to propagate. Self-heals at the cap, hence Medium not High.

backend/notification_v2/internal_api_views.py
```diff
@@ -557,6 +594,7 @@ def _send_clubbed(
                 "organization_id": org_id,
             },
             queue="notifications",
+            org_string_id=_org_identifier(org_id),
```
muhammad-ali-e, line 610: **[Low — type safety]** The PR's central invariant is that the two org identifiers must never be conflated: the org **string** id routes Flipt (`org_string_id`), while the org **pk** stays in `kwargs['organization_id']`. That's well *expressed* (keyword-only param, divergent naming, strong docstrings) but weakly *enforced*: `_send_clubbed` (L557), `_penalize_render_failure` (L653) and `_dispatch_group` (L673) all type `org_id: Any`, and `resolve_transport` does `str(organization_id)`, so an accidental pk↔string swap silently mis-routes rather than failing loudly. Cheapest hardening: type those three params `org_id: int` to match `_org_identifier(org_pk: int)`, letting mypy flag a mix-up along the whole caller chain.

workers/shared/enums/worker_enums_base.py
```diff
@@ -58,13 +58,22 @@ def to_import_path(self) -> str:
...
+        file-path task loader in ``worker.py`` reads it directly rather than
+        slicing the import path.
+        """
         directory_mapping = {
             "api_deployment": "api-deployment",
             # All others use same name for directory and module
         }
-        directory = directory_mapping.get(self.value, self.value)
-        return f"{directory}.tasks"
+        return directory_mapping.get(self.value, self.value)
```
muhammad-ali-e: **[Low — enforcement]** `directory_mapping` keys on `self.value` (the string `"api_deployment"`), so if an enum value were ever renamed the key would fall through `.get(self.value, self.value)` and silently return the *wrong* directory. The sibling `to_health_port` keys its `port_mapping` on the enum **member** (`WorkerType.API_DEPLOYMENT`). Keying this dict the same way ties the override to member identity and removes the stringly-typed rename hazard — one-line change, no behavior change.

backend/notification_v2/notification_dispatch.py
```diff
@@ -0,0 +1,119 @@
...
+    # Use the shared is_pg_transport() — the single source for "what counts as PG
+    # transport" — rather than opening a second comparison site.
+    if is_pg_transport(transport):
+        try:
+            enqueue_task(
+                task_name=WEBHOOK_NOTIFICATION_TASK,
+                queue=queue,
+                args=args,
+                kwargs=kwargs,
+                # enqueue_task's org_id is str-typed — coerce None→"" (unlike the
```
muhammad-ali-e: **[Low — comment accuracy]** This comment calls the `or ""` "load-bearing (unlike the routing arg above)", but `enqueue_task` already re-coerces internally — `producer.py:130` does `org_id=org_id or ""` — so passing `None` straight through would produce an identical row. The `or ""` is needed only for static-type correctness (`enqueue_task`'s param is typed `org_id: str`), not at runtime; the claimed runtime asymmetry with the routing arg doesn't exist. Suggest: `# enqueue_task is typed org_id: str; pass "" not None to satisfy the type (it re-coerces internally anyway).`

workers/worker.py
```diff
@@ -440,40 +441,69 @@ def on_task_postrun(sender=None, task_id=None, **kwargs):
...
+def load_worker_tasks(worker_type: WorkerType) -> None:
+    """Register the worker type's Celery tasks.
+
+    Pluggable workers are already registered by this point: ``build_celery_app()``
+    above verified the plugin via ``WorkerBuilder._verify_pluggable_worker_exists``,
+    which does ``importlib.import_module("pluggable_worker.<type>.worker")`` — a
+    proper PACKAGE import that runs the plugin's own task-registration code (e.g.
+    ``from . import tasks``, which may use relative imports such as
+    ``from .clients import ...``). Celery binds those tasks to the worker's app
+    when it finalizes, so they are registered by this point. The generic file-path
```
muhammad-ali-e: **[Low — comment accuracy]** This docstring asserts pluggable tasks are "registered by this point", but the empty-registry warning ~40 lines below (L497) correctly says they "bind on app finalize … which can be **after** this point." Both can't be true — `@shared_task` / `connect_on_app_finalize` binding is deferred to `app.finalize()`. The skip is still correct, but the real justification is the relative-import breakage (loading `tasks.py` under a bare `"tasks"` spec), not that tasks are already bound. Suggest aligning the docstring with the sibling comment: the package import *runs* the plugin's registration code; binding completes at finalize.

backend/notification_v2/internal_api_views.py
```diff
@@ -566,7 +604,34 @@ def _send_clubbed(
             webhook_url_hash(url),
             len(buffer_ids),
         )
+    except PermanentDispatchError:
+        # PG-ONLY permanent failure — enqueue_task validation (priority range /
```
muhammad-ali-e: **[Low — comment accuracy]** This call-site comment lists "priority range / reply_key+callback exclusivity" as the permanent failure modes, but `dispatch_webhook_notification` passes neither `priority` (defaults in-range) nor `reply_key`/callbacks, so those `enqueue_task` `ValueError`s are unreachable from this path. The only reachable permanent failure here is payload JSON-serialization. Suggest narrowing this comment to the reachable mode — the general list is fine to keep on the `PermanentDispatchError` *class* docstring.

backend/notification_v2/internal_api_views.py
```diff
@@ -513,6 +518,34 @@ def _reclaim_stale_sending() -> int:
...
+    it (the only org-traceable breadcrumb; resolve_transport's own warning is keyed
+    on the random dispatch uuid) and fail closed to Celery in resolve_transport.
+    """
+    org_string_id = (
+        Organization.objects.filter(pk=org_pk)
+        .values_list("organization_id", flat=True)
+        .first()
+    )
+    if org_string_id is None:
+        logger.warning(
```
muhammad-ali-e: **[Low — observability]** A `None` here is documented as a dangling FK — DB corruption "unreachable in normal operation" under `on_delete=CASCADE`. A genuine data-integrity anomaly ops must investigate is better surfaced as `logger.error` (Sentry-routed) than a WARNING no alert is keyed on. Failing closed to Celery is the right behavior; just consider bumping the severity so it's actually noticed.

workers/worker.py
```diff
@@ -440,40 +441,69 @@ def on_task_postrun(sender=None, task_id=None, **kwargs):
...
+    load below is therefore SKIPPED for pluggable workers — it loads ``tasks.py``
+    under a bare ``"tasks"`` spec with no parent package, which breaks any relative
+    imports in the plugin's ``tasks.py`` ("attempted relative import with no known
+    parent package").
+
+    Non-pluggable (top-level) workers use absolute imports; their ``tasks.py`` is
+    loaded by file path with the worker directory on ``sys.path``.
+    """
+    if worker_type.is_pluggable():
```
muhammad-ali-e, line 467: **[Low — test gap]** This pluggable-skip early-return is the core fix the PR ships (it avoids the bare-`"tasks"` spec that breaks a plugin's relative imports), yet it has no direct test — only `to_directory()` is covered. If a future edit drops or inverts this guard, no test fails and every pluggable worker regresses to "attempted relative import with no known parent package" at startup (caught only by the soft warning). The blocker is that `load_worker_tasks` is invoked at module scope alongside heavy side effects, so `import worker` boots a real worker. Consider guarding the module-level orchestration behind a `main()` / `__name__` guard so the function is importable and can be unit-tested (assert `exec_module` is never called for a pluggable type).

workers/worker.py
```diff
@@ -440,40 +441,69 @@ def on_task_postrun(sender=None, task_id=None, **kwargs):
...
+
+
+load_worker_tasks(worker_type)
+
+# A worker that boots with no registered tasks starts but silently processes
+# nothing (Celery does not error on an empty registry). Surface that misconfig —
+# as a WARNING, not a hard failure: pluggable tasks bind on app finalize
+# (@shared_task / connect_on_app_finalize), which can be after this point, so a
+# raise here would false-positive on a correctly-configured pluggable worker.
+_registered_tasks = [name for name in app.tasks if not name.startswith("celery.")]
```
muhammad-ali-e: **[Nit]** `_registered_tasks` is materialized solely to test emptiness. `if not any(not n.startswith("celery.") for n in app.tasks):` short-circuits and drops the throwaway list. Purely optional — the named list reads fine as-is.

muhammad-ali-e (commented):

Ran six specialist agents (Code Reviewer, Silent-Failure Hunter, Type-Design Analyzer, Test Analyzer, Comment Analyzer, Code Simplifier) plus an independent verification pass against `enqueue_task` / `resolve_transport` / `is_pg_transport`.

**Verdict: no blocking issues.** The transport-routing seam is correct, the flag-off (Celery) path is byte-identical to the prior `send_task`, the permanent-vs-transient error taxonomy is sound and `SENDING`-guarded, and the two-org-identifier contract is well-documented and pinned by tests. Everything below is an improvement, not a defect — the two `[Medium]` items are the only ones I'd act on before merge.

Findings are inline. Priority: 2 Medium, 6 Low, 1 Nit.

muhammad-ali-e:
## Review feedback addressed (commit 02dcdcb4a)

Worked through the automated review. All changes are either new-in-this-PR code or gated to the PG branch, so the flag-off (Celery) path stays byte-identical.

**Addressed (9):**
- **[Medium] loader empty-registry** — split by \`is_pluggable()\`: non-pluggable now \`raise RuntimeError\` on empty registry / missing dir / missing \`tasks.py\` (terminal — no late finalize binding); pluggable stays a WARN. Celery hot path unaffected (general/executor ship a \`tasks.py\`).
- **[Medium] NaN → re-dispatch loop** — \`_json_safe\` now uses \`json.dumps(..., allow_nan=False)\` so NaN/Infinity raise \`ValueError\` at the enqueue seam and funnel into \`PermanentDispatchError\`, instead of surfacing later as a permanent \`django.db.DataError\` at the JSONB insert. Except left narrow (transient DB errors still propagate). New unit test added. PG-branch only.
- **[Low] org-id typing** — \`_send_clubbed\` / \`_penalize_render_failure\` / \`_dispatch_group\` now type \`org_id: int\` to match \`_org_identifier(org_pk: int)\` (annotation-only).
- **[Low] observability** — dangling-FK org resolution bumped WARNING → \`logger.error\` (Sentry-routed); fail-closed-to-Celery unchanged.
- **[Low/Nit] comments + short-circuit** — corrected the loader docstring (import runs registration; binding completes at \`app.finalize\`; skip exists because a bare \`\"tasks\"\` spec breaks relative imports), the \`or \"\"\` note, and the \`PermanentDispatchError\` call-site comment; replaced the materialized registry list with a short-circuiting \`any(...)\`; re-keyed \`directory_mapping\` on the enum member.

**Deferred (follow-up ticket):** \`worker.py\` loader test-gap — needs a \`main()\`/\`__name__\` guard refactor to make the module import-safe for unit testing.

Tests: 28 backend + 3 worker green; pre-commit clean.

backend/pg_queue/producer.py
```diff
@@ -50,8 +50,15 @@ def _json_safe(value: Any) -> Any:
...
+
+    ``allow_nan=False`` rejects ``NaN``/``Infinity`` here with a ``ValueError``
+    rather than letting Python's default lenient encoder emit the non-standard
+    ``NaN``/``Infinity`` tokens: Postgres ``jsonb`` rejects those at insert with a
+    ``django.db.DataError`` (a permanent failure the notification dispatcher's
+    ``(ValueError, TypeError)`` seam would otherwise miss, looping the row). Fail
+    at the intended seam instead.
     """
-    return json.loads(json.dumps(value, default=str))
+    return json.loads(json.dumps(value, default=str, allow_nan=False))
```
muhammad-ali-e, line 61: **[Medium — silent failure / comment inaccuracy]** The new `allow_nan=False` correctly turns `NaN`/`Infinity` into a `ValueError` at this seam — but `_json_safe` runs during message construction (lines 109–126), *before* the `try/except` that wraps `PgQueueMessage.objects.create()` (line 133). So a serialization failure — the exact case this change newly makes reachable — propagates out of `enqueue_task` with **no producer-side breadcrumb**: the `logger.exception("PG-queue: failed to enqueue task=… queue=… org=…")` only ever wraps the DB insert.

This contradicts that handler's own comment ("so a DB/constraint/**serialization** error isn't mislabeled by the caller's broad handler … log the failure with breadcrumbs before it propagates") — serialization errors never reach it. On the notification path the miss is partly masked (it becomes `PermanentDispatchError` → DEAD_LETTER upstream), but on the orchestrator path (`workflow_helper` → `enqueue_task`, which per the docstring has "no silent Celery fallback") a bad-float payload is dropped with **zero** `task_name`/`queue`/`org` context for on-call — precisely the permanent-drop-without-a-log class this PR is otherwise trying to close.

Fix: move the `_json_safe(…)` coercions inside the `try` (or wrap message construction in the same `except … log … raise`), so serialization failures get the same breadcrumb the comment promises. If logging isn't wanted here, narrow the comment to DB-only so it stops advertising serialization coverage it doesn't provide.

muhammad-ali-e (commented):
Follow-up automated review pass (PR Review Toolkit). One net-new finding below; everything else surfaced was observability-level, a comment nit, a test-coverage gap, or on a file outside this diff — reported to the author out-of-band rather than posted here to keep the bar high.

muhammad-ali-e:
Addressed the CodeRabbit nitpick (commit 617b2fbb0): `test_permanent_pg_error_dead_letters` now also asserts the dead-letter `.update()` is guarded to rows still `status=SENDING` (via `filter.call_args.kwargs`), mirroring the transient-revert test's clobber-guard assertion. 5 tests green.

muhammad-ali-e:
Addressed the follow-up review finding (commit 1eb15c6dd): moved the `_json_safe` message construction inside `enqueue_task`'s `try/except`, so a serialization `ValueError` (now reachable via `allow_nan=False`) gets the same `task=/queue=/org=` breadcrumb as a DB-insert failure — no more context-free permanent drop on the orchestrator path. The comment's "serialization" coverage claim is now accurate. Test asserts the breadcrumb fires and the DB insert is never reached; 16 producer tests green.

backend/notification_v2/notification_dispatch.py
```diff
@@ -0,0 +1,121 @@
...
+    )
+    # Use the shared is_pg_transport() — the single source for "what counts as PG
+    # transport" — rather than opening a second comparison site.
+    if is_pg_transport(transport):
+        try:
+            enqueue_task(
+                task_name=WEBHOOK_NOTIFICATION_TASK,
+                queue=queue,
+                args=args,
+                kwargs=kwargs,
```
muhammad-ali-e: **[High — broken transport contract / duplicate external delivery]** The PG branch forwards `kwargs` verbatim, including `raise_on_final_failure: True` (set unconditionally in `_send_clubbed`, `internal_api_views.py:594`). That flag was designed for **Celery's** semantics: on retry exhaustion the task re-raises so it ends in `FAILURE` and `link_error` dead-letters the rows — with **no message-level redelivery**.

On the PG consumer the same re-raise means the opposite. `send_webhook_notification` marks the buffers `DEAD_LETTER` via `_mark_buffer_outcome(dispatched=False)` and then re-raises (`workers/notification/tasks.py:328,353`). That exception propagates out of `task.apply(throw=True)`; since this dispatch carries no `reply_key`/`on_success`/`on_error`, the consumer takes the fire-and-forget branch and **leaves the row for vt-expiry redelivery, bounded by `max_attempts`** (`consumer.py:538-547`). On redelivery the task re-runs from the top — `webhook_provider.send()` is called again with **no buffer-status idempotency guard** — so the subscriber is re-POSTed on every redelivery, up to `max_attempts` (default 5) × the in-task `max_retries` eager attempts, and the run finally trips `_drop_poison_message` (a scary `error`-level "poison-dropped" log) for what is an ordinary, already-dead-lettered notification failure.

**Net:** flag-off (Celery) POSTs the endpoint ~`max_retries` times then stops; flag-on (PG) hammers it ~`max_attempts × max_retries` times, **defeating the very dead-letter design this PR builds** and emitting false poison alarms to on-call. Independently confirmed by two review passes.

**Fix (contained, at this seam):** on the PG path the worker marks the buffers directly, so the re-raise buys nothing (there is no `link_error` and no consumed `FAILURE`). Override it before enqueue so the task returns `None` on terminal failure → the consumer acks (deletes) the row → single pass, matching Celery's external behavior:

```python
if is_pg_transport(transport):
    pg_kwargs = {**kwargs, "raise_on_final_failure": False}
    enqueue_task(..., kwargs=pg_kwargs, ...)
```

Please also add a regression test that drives a permanently-failing clubbed webhook through the PG consumer and asserts the endpoint is hit once (not `max_attempts` times) and no poison-drop is logged.

backend/notification_v2/notification_dispatch.py
```diff
@@ -0,0 +1,132 @@
...
+        # terminal failure re-raises so the task ends FAILURE and link_error
+        # dead-letters the rows, with NO message-level redelivery. On the PG
+        # consumer the same re-raise means the OPPOSITE — the worker already marked
+        # the buffers DEAD_LETTER directly, so re-raising only leaves the row for
+        # vt-expiry redelivery (bounded by max_attempts), re-POSTing the subscriber
+        # on every redelivery and tripping a false poison-drop. Force it False on the
+        # PG branch so a terminal failure returns None → the consumer acks (deletes)
+        # the row → single pass, matching Celery's external behaviour. (Celery branch
+        # below keeps kwargs verbatim — byte-identical.)
+        pg_kwargs = {**kwargs, "raise_on_final_failure": False}
```
vishnuszipstack: **[High — the override is unreachable for any org with `max_retries >= 1`]**

This override (added to address the earlier `raise_on_final_failure` thread) is a no-op for exactly the orgs it matters for. Verified against the installed Celery and by running the task shape directly.

The PG consumer runs the task with `task.apply(..., throw=True)` (`workers/queue_backend/pg_queue/consumer.py:509`). Under `apply()`:

* `self.request.retries` is **always 0**, so the guard in `send_webhook_notification` — `if max_retries is not None and self.request.retries < max_retries:` (`workers/notification/tasks.py:317`, `:342`) — is true on the *first* delivery failure for any `max_retries >= 1`;
* that branch calls `raise self.retry(exc=e, countdown=retry_delay)`, which raises `Retry`;
* with `throw=True` the tracer **propagates `Retry` straight out of `apply()`** — the `isinstance(retval, Retry) -> retval.sig.apply(...)` re-apply branch in `Task.apply` is never reached.

Reproduced with the consumer's exact call shape:

```
apply(throw=True), max_retries=3  ->  RAISED out of apply: Retry   POSTs: 1  buffer marks: []
apply(throw=True), max_retries=0  ->  returned: SUCCESS None       POSTs: 1  buffer marks: ['DEAD_LETTER']
```

So for `max_retries >= 1` the task **never reaches** the terminal branch — `_mark_buffer_outcome(dispatched=False)` never runs and `raise_on_final_failure` is never read. `Retry` is an `Exception`, so the consumer's `except Exception` takes the fire-and-forget branch (`consumer.py:539-546`) and leaves the row for vt-expiry redelivery: the subscriber is re-POSTed on every redelivery up to `max_attempts`, then poison-dropped — precisely the outcome the comment above says this override prevents. Rows stay `SENDING` until `_reclaim_stale_sending` returns them to `PENDING`, and the cycle repeats.

It only behaves as documented when `max_retries` is 0 — which is the model default (`backend/notification_v2/models.py:39-41`), so this passes dev testing and breaks for any subscriber that configured retries.

**Fix:** neutralize the in-task retry loop on PG as well, so the task takes the terminal branch and acks once:

```python
pg_kwargs = {**kwargs, "raise_on_final_failure": False, "max_retries": 0}
```

Retry spacing then belongs to the PG layer (or have the consumer honour `Retry.when` via `set_vt(countdown)` instead of treating it as a failure). Either way `test_pg_forces_raise_on_final_failure_false` should grow a case that drives a real `task.apply()` with `max_retries=3` and asserts one POST + one dead-letter mark — today it asserts the kwarg value, not the behaviour the kwarg is supposed to buy.

backend/notification_v2/notification_dispatch.py
```diff
@@ -0,0 +1,132 @@
+"""Transport-routed dispatch for buffered webhook notifications (UN-3753).
+
+Routes the ``send_webhook_notification`` task through the same
+:func:`resolve_transport` flag as the execution path: the PG queue when
+``pg_queue_enabled`` for this org, else Celery. **Fail-closed** — with the gate
+off (the production default) it resolves to Celery, behaving exactly like the
+prior unconditional ``celery_app.send_task`` (zero regression). On PG the task
+is drained by the PG notification consumer on the ``notifications`` queue.
```
vishnuszipstack: **[High — no consumer drains this queue in the OSS repo]**

This sentence states as present-tense fact that a PG notification consumer drains the `notifications` queue. Nothing in this repo does:

* `docker/docker-compose.yaml` defines seven `pg-queue-consumer` services; their `WORKER_PG_QUEUE_CONSUMER_QUEUE` values are `celery_api_deployments`, `celery`, `file_processing,api_file_processing`, `file_processing_callback,api_file_processing_callback`, `scheduler`, `celery_executor_*`, `ide_callback` (lines 571, 607, 641, 682, 721, 760, 812). **None is `notifications`.** Line 676 still reads "until notifications move to PG".
* `workers/run-worker.sh` `PG_CONSUMER_ROLES` has no notification role, so `-L pg` / `run_all_workers` never starts one.
* The consumer's own default is `_DEFAULT_QUEUE = "default"` (`consumer.py:59`).

Turning `pg_queue_enabled` on for an org therefore inserts every buffered webhook into `pg_queue_message(queue_name='notifications')` where nothing reads it. `_send_clubbed` logs `result=success`, rows sit `SENDING` → reclaimed → re-dispatched → dead-lettered by the attempt cap, and each cycle leaves an orphan row behind (there is no TTL sweep on `pg_queue_message` — the reaper's retention pass only covers `pg_task_result` and `pg_batch_dedup`). Net effect: **every webhook for that org is silently lost**, and starting the consumer later floods subscribers with the accumulated backlog.

I appreciate the cloud counterpart PR carries the chart, so this may be deployed there — but the docstring is making a claim about *this* repo, and an OSS operator reading it has no consumer. Please either add the missing wiring in this PR (a `notification;notifications` entry in `PG_CONSUMER_ROLES` plus a matching compose service with `WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE=notification`), or reword to state the prerequisite explicitly, e.g.:

> On PG the task lands on the `notifications` queue and requires a pg-queue consumer configured with `WORKER_PG_QUEUE_CONSUMER_WORKER_TYPE=notification` / `WORKER_PG_QUEUE_CONSUMER_QUEUE=notifications`. No such service exists in this repo's compose yet — the flag must stay off until one is deployed.

backend/notification_v2/notification_dispatch.py
```diff
@@ -0,0 +1,132 @@
...
+    # straight through (no `or None` needed).
+    transport = resolve_transport(
+        execution_id=dispatch_id,
+        organization_id=org_string_id,
+    )
+    # Use the shared is_pg_transport() — the single source for "what counts as PG
+    # transport" — rather than opening a second comparison site.
+    if is_pg_transport(transport):
+        # ``raise_on_final_failure`` is Celery semantics: on the Celery path a
+        # terminal failure re-raises so the task ends FAILURE and link_error
```
vishnuszipstack: **[Medium — comment premise is stale: `link_error` was removed]**

"on the Celery path a terminal failure re-raises so the task ends FAILURE and **link_error dead-letters the rows**" describes a mechanism this codebase no longer uses, and the PR's own neighbouring code says so:

* `internal_api_views.py:571-577` — "reported back by the notification worker over the internal API (`buffer/mark/dispatched` / `buffer/mark/dead-letter`), **not via Celery `link`/`link_error`**: those callbacks routed to the `celery` queue … so ~half were dropped as 'unregistered task'";
* `backend/notification_v2/tasks.py:4` — "The clubbed dispatch no longer wires these as `link`/`link_error`";
* `_send_clubbed` passes no `link_error=` argument at all.

What actually dead-letters the rows on **both** transports is the worker: `workers/notification/tasks.py:327/333/352/358` call `_mark_buffer_outcome(..., dispatched=False)` *before* `if raise_on_final_failure: raise`. The call site states this correctly at `internal_api_views.py:591-593`, so two comments in the same PR contradict each other.

This matters because the "PG means the OPPOSITE" argument rests on an asymmetry that doesn't exist — the dead-lettering is identical on both paths; only the *post-raise disposition* differs (Celery: FAILURE state, no redelivery; PG: vt-expiry redelivery). Suggested rewrite:

```
# ``raise_on_final_failure`` is Celery semantics. On BOTH transports the worker
# has already marked the buffers DEAD_LETTER over the internal API before it
# re-raises (workers/notification/tasks.py::send_webhook_notification), so the
# re-raise is purely a FAILURE-state signal for Celery monitoring — no
# redelivery. On the PG consumer the same re-raise is not a signal but a
# retry: the fire-and-forget branch leaves the row for vt-expiry redelivery
# (bounded by max_attempts), re-POSTing the subscriber each time. Force it
# False on the PG branch so the task returns None -> the consumer acks.
```

The same stale claim exists upstream at `workers/notification/tasks.py:254-256` — worth fixing together.

backend/notification_v2/notification_dispatch.py
```diff
@@ -0,0 +1,132 @@
+"""Transport-routed dispatch for buffered webhook notifications (UN-3753).
+
+Routes the ``send_webhook_notification`` task through the same
+:func:`resolve_transport` flag as the execution path: the PG queue when
+``pg_queue_enabled`` for this org, else Celery. **Fail-closed** — with the gate
+off (the production default) it resolves to Celery, behaving exactly like the
+prior unconditional ``celery_app.send_task`` (zero regression). On PG the task
+is drained by the PG notification consumer on the ``notifications`` queue.
+
+The same ``args``/``kwargs``/``queue`` are forwarded on both paths; on PG they're
```
vishnuszipstack: **[Low — module docstring contradicts the code]**

"The same `args`/`kwargs`/`queue` are forwarded on both paths" is no longer true: line 100 forces `raise_on_final_failure=False` on the PG branch. The same over-claim repeats in the function docstring at `:55` ("forwarded unchanged on both paths") and in the `Args:` entry at `:62` ("forwarded verbatim").

The inline comment at `:90-99` and the tests both document the override correctly — these three summary statements are the only places still claiming otherwise, and they are exactly what a reader skimming the docstring takes away. This is also the kwarg a future maintainer must not "clean up" as redundant.

```
``args`` and ``queue`` are forwarded verbatim on both paths. ``kwargs`` differ
in exactly one key: ``raise_on_final_failure`` is forced ``False`` on the PG
branch (see the inline note there). On PG the payload is JSON-normalized by
``enqueue_task`` (UUIDs/datetimes -> str).
```

and at `:62`, "forwarded as-is except `raise_on_final_failure`, which the PG branch overrides to `False`."

backend/notification_v2/internal_api_views.py
```diff
@@ -566,7 +608,37 @@ def _send_clubbed(
...
+            platform,
+            org_id,
+            webhook_url_hash(url),
+            len(buffer_ids),
+        )
+        NotificationBuffer.objects.filter(
+            id__in=buffer_ids,
+            status=BufferStatus.SENDING.value,
+        ).update(status=BufferStatus.DEAD_LETTER.value)
     except Exception:
```
vishnuszipstack, line 650: **[Medium — the unconditional refund makes `NOTIFICATION_MAX_DISPATCH_ATTEMPTS` unreachable]**

The `SENDING` claim charges `+1` (`internal_api_views.py:769`) and this handler refunds `-1` (`:656`). Net zero per cycle — so for **any** failure that keeps landing in this branch, `dispatch_attempts` never grows and the cap can never dead-letter the group. It re-renders and re-dispatches on every flush tick, forever, emitting a `logger.exception` traceback each time.

That was a reasonable trade when the sole occupant was "RabbitMQ is down" (self-healing, and refunding is correct because nothing was sent). This PR widens the branch's population: any PG enqueue failure that isn't `ValueError`/`TypeError` lands here, including ones that are permanent and will recur identically — e.g. a payload string containing `\u0000`, which passes `allow_nan=False` cleanly and then fails at insert with `django.db.DataError` on every single tick.

(The classification question itself is already covered in the `notification_dispatch.py:114` thread; this is specifically about the refund making termination impossible even once a failure is known to repeat.)

**Suggested:** bound the refund rather than making it unconditional — e.g. only refund while `dispatch_attempts` is below a small threshold, or don't refund at all and let the cap terminate the group. As written, no failure reaching line 636 has a termination condition.

backend/notification_v2/internal_api_views.py
```diff
@@ -557,6 +598,7 @@ def _send_clubbed(
                 "organization_id": org_id,
             },
             queue="notifications",
+            org_string_id=_org_identifier(org_id),
         )
         logger.info(
             "metric=notification_batch_dispatched_total platform=%s result=success "
```
vishnuszipstack, line 616: **[Low — the rollout's own metric can't distinguish the two transports]**

`result=success` is emitted identically whether the dispatch went to PG or Celery, and `dispatch_webhook_notification` returns a bare `str` whose provenance (minted PG task id vs Celery `AsyncResult.id`) is indistinguishable to the caller — which discards it anyway.

For a percentage-rollout flag this is the one metric you'll want during the ramp: "are PG-routed notifications succeeding at the same rate as Celery-routed ones?" is currently unanswerable from the logs. Since the seam already knows which branch it took, returning it costs almost nothing:

```python
class DispatchResult(NamedTuple):
    transport: str
    task_id: str
```

then add `transport=%s` to this line. Related: the seam discards `enqueue_task`'s returned `msg_id`, so the producer's `msg_id=` log (`producer.py:152`) and the seam's `task_id=` log (`notification_dispatch.py:120`) can't be correlated when chasing a dropped notification.

backend/pg_queue/producer.py
```diff
@@ -97,33 +104,37 @@ def enqueue_task(
...
-    # the caller's broad handler.
+    # the caller's broad handler. Message construction (the _json_safe coercions)
+    # lives inside the try because serialization failures surface there — e.g. a
+    # NaN/Infinity float raises ValueError under allow_nan=False — and on the
+    # orchestrator path (no silent fallback) would otherwise drop the task with no
+    # task/queue/org breadcrumb for on-call.
     try:
+        message: TaskPayload = {
+            "task_name": task_name,
+            "args": _json_safe(list(args) if args is not None else []),
```
vishnuszipstack, line 117: **[Medium — test gap: the NaN guard's real trigger rides in `args`, but only `kwargs` is covered]**

`allow_nan=False` is applied to `args` (this line) and `kwargs` (`:118`), but `test_json_safe_rejects_nan_with_value_error` only exercises `kwargs`. The scenario this change exists for is a webhook **body** carrying a NaN float — and `notification_dispatch.py:105` puts the body in `args[1]`, not `kwargs`. The one code path the guard was written for is the untested one.

Also uncovered on the same lines: `float("inf")` / `float("-inf")` (equally rejected by `allow_nan=False`), and NaN nested inside `fairness` (`:123`) or `on_success`/`on_error` (`:133-135`) — all of which are `_json_safe`-coerced here and would otherwise reach the `jsonb` insert.

Second, smaller point on that test: it patches `producer.logger` wholesale and asserts only `log.exception.assert_called_once()`. The stated purpose of moving message construction inside the `try` (`:109-113`) is that on-call gets a **task/queue/org breadcrumb** — but the format args are never checked, so a breadcrumb that silently lost `task_name` would still pass. Prefer `caplog` and assert the rendered record contains the task name, queue and org.

**Suggested:** parametrize over `[nan, inf, -inf]` x `[args, kwargs, fairness]`.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -789,13 +789,17 @@ def _drop_poison_message(
...
             task_name,
+            message.read_ct,
             message.msg_id,
-            self.max_attempts,
             message.read_ct,
+            payload.get("queue"),
             execution_id,
+            organization_id or None,
+            self.max_attempts,
             payload,
```
vishnuszipstack: **[Medium — this log now carries subscriber credentials into Sentry, and the queue row stores them at rest]**

`full payload: %r` predates this PR, but this PR is what first routes **credential-bearing** payloads through the PG consumer, so the exposure is new in practice.

`_send_clubbed` puts `headers` into `args[2]` (`internal_api_views.py:586`), and `build_webhook_headers` (`backend/notification_v2/helper.py:93-107`) writes the subscriber's secret straight into it:

```python
headers["Authorization"] = f"Bearer {auth_key}"   # BEARER
headers["Authorization"] = auth_key               # API_KEY
headers[auth_header] = auth_key                   # CUSTOM_HEADER
```

So every poison drop of a buffered notification dumps the subscriber's bearer token / API key **plus the full clubbed webhook body (customer pipeline data)** to stdout and Sentry at ERROR.

Separately and arguably worse: `pg_queue_message.message` is a plain JSONB column, so those same secrets are now persisted **unencrypted at rest** for as long as the row exists — and rows are only removed on ack, with no TTL sweep on `pg_queue_message` (the reaper's retention pass covers only `pg_task_result` and `pg_batch_dedup`). Combined with the missing `notifications` consumer, undrained rows accumulate indefinitely.

**Suggested:** redact before logging — a small `_redact_payload()` that masks any `args`/`kwargs` entry whose key matches `authorization|api[-_]?key|token|secret`, and log a hash or length for the body rather than the body itself.

workers/queue_backend/pg_queue/consumer.py
```diff
@@ -789,13 +789,17 @@ def _drop_poison_message(
         """
         execution_id, organization_id = _pipeline_identity(payload)
         logger.error(
-            "PG-queue consumer: task %r (msg_id=%s) exceeded max_attempts=%s "
-            "(read_ct=%s, execution_id=%s) — poison; full payload: %r",
+            "PG-queue consumer: poison-dropped task %r after %s attempts "
+            "(msg_id=%s, read_ct=%s, queue=%s, execution_id=%s, org_id=%s) — "
+            "exceeded max_attempts=%s; full payload: %r",
             task_name,
+            message.read_ct,
```
vishnuszipstack: **[Low — `read_ct` is printed twice under two different names, and `org_id` mixes identifier spaces]**

Two issues in the enriched format string:

1. `message.read_ct` is passed **twice** — here as `after %s attempts` and again at `:798` as `read_ct=%s`. Every drop renders the same number in both slots, e.g. `... after 6 attempts (msg_id=5, read_ct=6, ...) — exceeded max_attempts=5`. "attempts" is also misleading: `read_ct` counts *reads*, and a lease expiry with no execution still increments it. Drop the phrase and keep the labelled field (which is what `test_poison_log_enriched_with_org_and_read_ct` actually asserts on).

2. `org_id=%s` comes from `_pipeline_identity`, which reads `kwargs["organization_id"]` (`consumer.py:195`). For the notification payload this PR introduces, that key holds the org **pk (an int)** — `internal_api_views.py:598` — whereas every other consumer log line and the entire Flipt/routing path use the org **string** id. Grepping `org_id=` across consumer logs will silently mix two identifier spaces, which is the exact conflation the rest of this PR is careful to avoid. The new test uses a pipeline payload where `org_id` happens to be the string id, so it can't catch this.

**Suggested:** drop the duplicate arg, and either label the notification case distinctly or note in `_pipeline_identity`'s docstring that the value's flavour is payload-dependent.

workers/worker.py
```diff
@@ -440,40 +441,83 @@ def on_task_postrun(sender=None, task_id=None, **kwargs):
...
+
+
+load_worker_tasks(worker_type)
+
+# A worker that boots with no registered tasks starts but silently processes
+# nothing (Celery does not error on an empty registry). Surface that misconfig,
+# but split by worker kind: pluggable tasks bind on app finalize (@shared_task /
+# connect_on_app_finalize), which can be after this point, so a raise here would
+# false-positive on a correctly-configured pluggable worker — only WARN. A
+# non-pluggable worker has no later binding step (load_worker_tasks bound its
```
vishnuszipstack: **[Medium — the stated rationale for warn-only is incorrect: `app.tasks` auto-finalizes]**

The comment argues that a raise here would false-positive on a correctly-configured pluggable worker, because "pluggable tasks bind on app finalize … which can be after this point". Accessing `app.tasks` on line 505 is itself what triggers finalization. From Celery's own source:

```python
@property
def tasks(self):
    # docstring: "Task registry.
    #   Warning: Accessing this attribute will also auto-finalize the app."
    self.finalize(auto=True)
    return self._tasks
```

Confirmed empirically — a pending `@shared_task` is invisible before touching `.tasks` and visible after, with `app.finalized` flipping `False -> True`:

```
finalized before .tasks: False
finalized after  .tasks: True
non-celery tasks visible: ['probe.later']
```

Two consequences:

1. The premise is false. Iterating `app.tasks` applies every pending `@shared_task` / `connect_on_app_finalize` registration *before* the emptiness test is evaluated, so the check is exactly as accurate for pluggable workers as for non-pluggable ones. The warn-only branch therefore lets a genuinely broken pluggable worker boot as a silent no-op — the precise failure mode this guard was added to prevent, and the one that matters most since pluggable workers are the cloud deployment.
2. There's an unremarked side effect: the app is now finalized at **import** time rather than at worker start.

The same premise appears in the `load_worker_tasks` docstring at `:452-459`.

**Suggested:** correct both comments, and either raise for pluggable workers too, or state the real reason for the softer treatment (e.g. cloud plugins that register tasks from a module imported after this point) — right now the justification given doesn't hold.

workers/shared/enums/worker_enums_base.py
```diff
@@ -58,13 +58,22 @@ def to_import_path(self) -> str:
         if self.is_pluggable():
             return f"pluggable_worker.{self.value}.tasks"
 
-        # Map to actual directory structure
+        return f"{self.to_directory()}.tasks"
+
+    def to_directory(self) -> str:
```
vishnuszipstack, line 63: **[Low — the "(non-pluggable)" precondition is documented but unenforced, and the failure is silent]**

`to_directory()` returns `directory_mapping.get(self, self.value)`, so calling it on a pluggable type yields the bare enum value (e.g. `"my_plugin"`) — but a pluggable worker's code lives under `pluggable_worker/<value>`. The returned path is wrong and *plausible-looking*, so nothing surfaces it.

Today that's safe only by call ordering: `to_import_path` guards with `is_pluggable()` at `:58`, and `worker.py:475` is reached only after the early return at `worker.py:467`. Nothing enforces that ordering, and this PR just made the consequence sharper — `worker.py:477-480` now **raises** `RuntimeError` on a missing directory instead of logging, so a future caller that skips the guard turns a silently-wrong path into a boot crash-loop.

**Suggested:** make the precondition real rather than prose —

```python
if self.is_pluggable():
    raise ValueError(f"{self!r} is pluggable; use to_import_path()")
```

(or return `f"pluggable_worker/{self.value}"`). Minor docstring nit while you're here: "a few on-disk dirs use hyphens" describes a mapping with exactly one entry.

backend/notification_v2/tests/test_send_clubbed.py
```diff
@@ -0,0 +1,117 @@
...
+            patch.object(views, "_org_identifier", return_value="o"),
+            patch.object(
+                views,
+                "dispatch_webhook_notification",
+                side_effect=RuntimeError("broker down"),
+            ),
+            patch.object(views, "NotificationBuffer") as buf,
+        ):
+            _send(buffer_ids=["b1", "b2"])
+        # Guarded on SENDING, reverted to PENDING (attempt refunded).
```
vishnuszipstack: **[Medium — the comment claims a refund the assertions never check]**

This comment says "(attempt refunded)", but lines 71-73 assert only `status` and `dispatched_at`. The refund itself —

```python
dispatch_attempts=F("dispatch_attempts") - 1,
```

(`internal_api_views.py:656`) — is unasserted. Delete that kwarg from production and this test still passes, yet a transient broker/PG outage would then **burn a dispatch attempt per flush tick**, so a 5-tick outage silently dead-letters live customer notifications that were never sent.

A comment asserting coverage that doesn't exist is worse than a plain gap — it reads as covered in review. The dead-letter sibling at `:87` already does the rigorous thing (exact-dict `==`, which correctly proves the *absence* of a refund); this test should match it:

```python
assert set(ukw) == {"status", "dispatched_at", "dispatch_attempts"}
assert "dispatch_attempts" in str(ukw["dispatch_attempts"])  # F(...) - 1
```

Worth adding a paired `test_permanent_error_does_not_refund_attempt` so the asymmetry between the two branches is explicit rather than incidental.

Related gap in `test_routes_string_id_and_keeps_pk_in_kwargs` above: it checks `org_string_id`, `kwargs["organization_id"]` and `queue`, but not `args == [url, body, headers, settings.NOTIFICATION_TIMEOUT]` nor `kwargs["buffer_row_ids"] == buffer_ids`. The latter is what replaced Celery `link`/`link_error` — drop it and the worker can never call the mark endpoints, so **every row strands in SENDING** until the reclaim lease expires. Silent, and visible only as a growing SENDING backlog.

workers/tests/test_worker_enums_directory.py
```diff
@@ -0,0 +1,30 @@
...
+to_import_path(); these pin the hyphen/underscore mapping and that to_import_path
+is built on top of it, so the two can't drift.
+"""
+
+from celery import Celery
+from shared.enums.worker_enums_base import WorkerType
+
+# The workers autouse conftest fixture finalizes celery's default_app around every
+# test; this pure-enum test builds no Celery app of its own, so establish one.
+Celery("test-worker-enums").set_default()
```
vishnuszipstack: **[Low — a leaf test module mutates process-global Celery state at import time]**

`set_default()` writes `celery._state.default_app` **unconditionally**, and this runs at *collection* time, not inside a test. From this module's import onward the session-wide `default_app` is a bare stub named `test-worker-enums`, which the autouse `_restore_current_celery_app` fixture (`workers/tests/conftest.py:404-418`) then finalizes and pins as `current_app` for **every subsequent test**. This is the only `set_default()` in the entire workers tree, and whether it wins depends on collection order.

It currently passes — I ran this module together with a later-alphabetical one and got 6 passed — so this is latent rather than breaking. The failure mode when it does bite is nasty: a test that relies on the pre-existing default app carrying app-specific (non-`shared_task`) registrations fails with `NotRegistered`, and the cause is a module it never imports.

The line is genuinely load-bearing today, though — removing it makes these tests ERROR when the file runs alone, because the fixture calls `default_app.finalize()` on a `default_app` that is `None`. So the root cause is a **conftest defect being worked around in the wrong place**: a pure-enum test needs no Celery app at all.

**Suggested:** guard the fixture (`if default_app is not None:`) at `conftest.py:411-413` and delete lines 8-13 here.

workers/tests/test_worker_enums_directory.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def test_to_directory_maps_underscored_value_to_hyphenated_dir():
+    # The one worker whose on-disk dir differs from its enum value.
+    assert WorkerType.API_DEPLOYMENT.value == "api_deployment"
+    assert WorkerType.API_DEPLOYMENT.to_directory() == "api-deployment"
```
vishnuszipstack, line 25: **[Low — cover the whole mapping, not two members]**

These tests pin `API_DEPLOYMENT` and `GENERAL`. But UN-3798's actual failure mode is enum-vs-disk **drift**, and this PR raised the stakes: `worker.py:477-480` now raises `RuntimeError` on a missing directory where it previously logged and continued, so drift is a container crash-loop across every PG consumer replica rather than a degraded worker.

A parametrized test closes the class of bug instead of two instances:

```python
@pytest.mark.parametrize("wt", list(WorkerType))
def test_every_worker_type_maps_to_an_existing_directory_with_tasks(wt):
    d = Path(__file__).parents[1] / wt.to_directory()
    assert d.is_dir(), f"{wt.value} -> {wt.to_directory()} does not exist"
    assert (d / "tasks.py").is_file()
```

I verified this passes today for all nine members (`api-deployment`, `callback`, `executor`, `file_processing`, `general`, `ide_callback`, `log_consumer`, `notification`, `scheduler` all have a `tasks.py`), so it's a free lock-in that converts a production `RuntimeError` into a red build the moment someone adds a `WorkerType` without a directory.

vishnuszipstack (commented):
## PR Review Toolkit — consolidated findings

Ran the Code Reviewer, Silent Failure Hunter, Type Design Analyzer, PR Test Analyzer, Comment Analyzer and Code Simplifier agents over the 12 changed files, then verified each surviving finding against the source. Findings already raised on this PR are omitted.

**Two blockers before the flag can be turned on for any org:**

1. The `raise_on_final_failure=False` override is unreachable for any subscriber with `max_retries >= 1` — verified by running the consumer's exact `task.apply(throw=True)` shape. The buffers are never dead-lettered and the subscriber is re-POSTed on every vt-expiry redelivery.
2. Nothing in this repo drains the PG `notifications` queue — none of the seven `pg-queue-consumer` compose services polls it, and `run-worker.sh` has no notification PG role. Flag-on means every buffered webhook for that org is enqueued and silently lost.

Also flagged: an unconditional attempt-refund that makes the dispatch cap unreachable, subscriber credentials reaching Sentry and PG-at-rest via the poison-drop path, an empty-registry guard whose stated rationale is contradicted by Celery's auto-finalize, and several test/comment accuracy gaps.

Nice work on the seam tests themselves — they assert forwarded payloads and negatives rather than mock tautologies, and `test_permanent_pg_error_dead_letters` proving the *absence* of a refund via exact-dict equality is the right instinct.

vishnuszipstack (approved):

PR #2210, opened 2026-07-25
UN-2646 [FEAT] LLMWhisperer image output mode adapter

praveen-formido:
## What

- Adds an **image output mode** to the LLMWhisperer V2 X2Text adapter (`unstract-sdk1`). When the adapter is configured with `output_mode = image`, it converts a **PDF** into per-page images via the LLMWhisperer **pdf-to-images** endpoints (submit → poll → retrieve + unzip) instead of extracting text.
- Introduces `PageImageReference` (a dedicated DTO) and an additive, `None`-defaulted `page_images` field on `TextExtractionMetadata`. Page images are persisted to Unstract FileStorage and referenced by these objects — they are **never** smuggled into the `extracted_text` string, so every existing text-mode consumer is byte-unaffected.
- Wires the mode through the adapter's JSON schema/UI: a new `image` enum value labelled **"Image (PDF only)"**, plus a conditional description that shows the PDF-only guidance **only** when Image is selected (UNS-759).
- Enforces the **PDF-only** constraint in two layers with one shared message:
  - Runtime guard `LLMWhispererV2._validate_pdf_only` inside `process()` (SDK, MUNS-195).
  - Fail-fast check at index time in the backend's `build_index_payload` (the live pre-dispatch path) so the user is rejected **before** an executor task is dispatched (UNS-757).
- Kept **strictly independent** of the `document_insights`/signature feature (#1967) — this branch is image-output only.

## Why

- Downstream workflows need the raw page images of a PDF (e.g. image-first pipelines / vision models), which LLMWhisperer now exposes via its `pdf-to-images` API. Surfacing it as an adapter `output_mode` lets Prompt Studio and deployments opt in with no new adapter type.
- Modelling the result as a typed `PageImageReference` list (rather than overloading `extracted_text`) keeps the change **additive and non-breaking**: text/layout modes serialize exactly as before (`page_images` is omitted when `None`).
- Validating PDF-only in both the SDK runtime and the backend index path means the user gets an identical, actionable error at the earliest possible point instead of waiting for a worker to fail extraction.

## How

- **DTO & constants (MUNS-193):** `PageImageReference` (`page_number`, `path`, optional `filename`/`size_bytes`/`provider`) with `to_dict`/`from_dict`; `OutputModes.IMAGE`; `ImageOutputConfig` (endpoint/response contract, `PDF_EXTENSION`, single-source `PDF_ONLY_ERROR`); `WhispererConfig.OUTPUT_MODE`.
[...]

PR #2218, opened 2026-07-28
UN-2646 [FEAT] OSS half of the VLM image-answer feature (path contract, loader, vision policy, bridge)

pk-zipstack:
## What

The complete **OSS half of the VLM image-answer feature** (the cloud-only consumer that answers prompts against LLMWhisperer image-output documents by sending the persisted page images to a vision LLM). Four commits, reviewable independently:

1. **Shared page-path contract** — promotes `build_page_store_dir` (`{extract_dir}/{stem}/pages`) and the page-naming constants (`PAGE_GLOB_PATTERN`, `PAGE_NUMBER_REGEX`) from the LLMWhisperer helper to the shared x2text surface; the adapter binds the shared function by identity. The tolerant ZIP-ingest pattern stays adapter-local as a named `ZIP_PAGE_MEMBER_REGEX` (service wire contract ≠ storage contract).
2. **sdk1 reader utilities** — a FileStorage-backed page-image loader (natural sort by integer page index, fail-fast page cap, base64 + `complete_vision` content blocks with "Page N" labels, typed empty/partial/over-cap errors with cache-bypass + re-billing remediation copy) and a vision-capability policy (LiteLLM local-registry classification: SUPPORTED / UNSUPPORTED / UNKNOWN; hard-block only definitive UNSUPPORTED, warn-and-allow unknown/self-hosted — never `get_model_info`, which can make network calls for Ollama).
3. **OSS bridge** (`vlm_image_answer`, mirroring the `lookup_enrichment` bridge with the opposite error policy): detects image mode in the `answer_prompt` flow by resolving the x2text adapter config via the platform service (payload carries only the instance id), dispatches to the cloud `vlm-image-answer` executor plugin, and raises structured code-prefixed errors (`IMAGE_OUTPUT_REQUIRES_CLOUD`, `IMAGE_OUTPUT_MISSING`, `IMAGE_PAGE_CAP_EXCEEDED`, `VISION_LLM_REQUIRED`) when it cannot serve — never a silent fallthrough to the text path. Single-pass extraction is rejected for image-mode profiles. Backend `vlm_utils` no-op bridge wires: profile `vision_warning`, deploy-time guard, and re-extraction invalidation.
4. **FE**: profile save surfaces a backend-computed `vision_warning` toast (field never present in OSS responses).

## Why

- Image mode produces per-page PNGs and no text; without a consumer the executor would answer prompts against the one-line extraction summary — the prohibited silent-wrong-answer mode. The bridge guarantees image-mode prompts either reach the vision path (cloud) or fail loudly with an actionable error (OSS).
- Writer (adapter) and reader (consumer) share one path derivation and one naming contract by identity, so they can never drift — this is what makes the no-transport/no-manifest design safe.
- The consumer itself is a paid cloud feature; OSS ships only neutral sdk1 utilities, the bridge, and gating (companion cloud PR: Zipstack/unstract-cloud — `feat/vlm-image-answer`).

## How

- Hook placement: inside `_execute_single_prompt`, the image-mode branch replaces **only** retrieval + completion; type conversion, lookup enrichment, webhooks, challenge, and output persistence run unchanged, so image-mode answers behave identically downstream.
- Detection cache: per `(execution_id, adapter_instance_id)`, bounded — one platform-service call per adapter per run, and adapter edits are picked up on the next execution.
- Error codes are prefixed onto messages because `ExecutionResult.error` is string-only; the code survives verbatim to the Prompt Studio socket events and deployment API responses.

- **No.** Text/layout-mode prompts take the `vlm_answer is None` branch and behave byte-identically (the bridge returns `None` without a platform call when no x2text adapter id is present, and after one cached resolution otherwise). The path-contract refactor is behavior-preserving — all 57 pre-existing sdk1 image tests pass unmodified. The backend hooks are no-ops without the cloud package. The FE change keys off a response field OSS never emits.

## Database Migrations

- None.

## Env Config

- None required. (`VLM_IMAGE_ANSWER_PAGE_CAP` is read by the cloud plugin, not OSS.)

## Relevant Docs

- LLMWhisperer output modes: https://docs.unstract.com/llmwhisperer/llm_whisperer/apis/llm_whisperer_text_extraction_api/#output-modes

## Related Issues or PRs

- Jira: **UN-2646** (VLM consumer phase — modules MUNS-206..212).
- **Stacked on #2210** (adapter image output mode) — retargets to `main` when #2210 merges.
- Companion cloud PR: `unstract-cloud` `feat/vlm-image-answer` (executor plugin + backend hooks; the two halves meet only at runtime via the `unstract.executor.plugins` entry point and the `plugins.vlm_image_answer` package presence).

## Dependencies Versions

- None added.

### Unit suites

- **sdk1: 123 pass** (26 path-contract + 23 loader incl. stale-listing/TOCTOU regressions + 13 vision-capability + pre-existing image suite with 5 new PDF-sniffing cases, rest unmodified).
- **workers: 13 pass** (`executor/tests/test_vlm_image_answer_bridge.py` — detection, per-execution caching, plugin-absent structured error, error-code mapping, single-pass guard).
- **backend: 20 pass** (8 `vlm_utils` bridge + 12 image-output gating).
- All with pinned ruff v0.3.4 clean.

### Live end-to-end validation (merged-tree Docker stack, MinIO object storage)

Run against a local rig with the companion cloud plugin installed, then removed — every path below verified in the running product:

| Path | Result |
|---|---|
| Prompt Studio happy path (1-page invoice) | ✅ Correct VLM answer **with spatial reasoning** ("upper right corner of the invoice…") — conclusive proof the answer came from the page image, since image mode has no extracted text |
| Multi-page (11-page doc, stem with spaces) | ✅ Ordered "Page N" delivery; token cost linear (1,559/page vs 1,570 single-page) |
| **API deployment E2E** (7-page PDF via `POST /deployment/api/...`) | ✅ After fixing the extension-less input bug (below); pages persisted under the execution dir on MinIO, answered by the vision LLM |
| Missing images — empty `pages/` | ✅ `IMAGE_OUTPUT_MISSING` with cache-bypass + per-page re-billing remediation |
| Missing images — partial (2 of 11 purged) | ✅ Distinct copy with exact missing list `[6, 9]`; **zero VLM cost** on failure |
| Page cap (cap=2 vs 7-page doc) | ✅ `IMAGE_PAGE_CAP_EXCEEDED`; **zero image reads** (fail-fast before any bytes) |
| Vision gate — run time (Bedrock `nova-micro`, non-vision) | ✅ `VISION_LLM_REQUIRED` naming the model |
| Vision gate — config time (same model) | ✅ Non-blocking `vision_warning` toast on profile save |
| Single-pass rejection (with the cloud single-pass executor installed) | ✅ `IMAGE_OUTPUT_UNSUPPORTED_OPERATION` before the executor is invoked |
| Service-side failure (password-protected PDF) | ✅ Fail-closed poll surfaced the PDFium error verbatim on the first `error` status |
| Text/layout-mode regression | ✅ No behavior change |
| **Plugin-absent (pure OSS)** | ✅ `image` stripped from the served schema (enum + enumNames + conditional block); existing image-mode profile → `IMAGE_OUTPUT_REQUIRES_CLOUD` (never a silent answer from the summary); adapter save backstop → clean 400 |

### Bugs found and fixed by the live testing (each with regression tests)

1. **Stale object-store listings / TOCTOU** (`ff7a4774`): fsspec's directory cache in the long-lived worker served a stale page listing, so a purged page passed discovery and blew up at read time as a raw `FileNotFoundError`. Discovery now refreshes the backend listing cache, and a read-time miss maps to the typed `PageImageSetIncompleteError`.
2. **Extension-less deployment inputs** (`801ce759`): workflow executions store the source file as `SOURCE` (no extension), so the extension-only PDF guard false-rejected every API-deployment input. The guard now falls back to `%PDF-` magic-byte content sniffing (fail-closed; no extra read on `.pdf`-named paths).
3. **Gating 500 on adapter save** (`4d40c473`): the rejection was raised with a bare-string `ValidationError` from inside `to_internal_value`, which crashes DRF's error collection into a 500. Now a field-keyed dict detail → clean 400 with the "available only on Unstract Cloud" message.

### Known accepted edge

An adapter **saved in image mode on cloud** and later opened on a plugin-less deployment renders a misleading dropdown state (RJSF displays the schema default while `formData` still holds the removed `image` value); submit is correctly rejected with the clean 400, and actively re-picking any mode recovers. Cloud→OSS downgrade only; documented rather than fixed (a fix would mean mutating stored data on read or custom widget work for a rare corner).

## Checklist

### DIFF
--- a/backend/prompt_studio/vlm_utils.py
+++ b/backend/prompt_studio/vlm_utils.py
@@ -21,6 +21,21 @@ try:
 except ImportError:
     _hooks = None
     VLM_IMAGE_ANSWER_AVAILABLE = False
+    # Distinguish "running OSS" (package absent — expected, silent) from
+    # "cloud hooks are broken" (package present but backend_hooks failed
+    # to import): in the latter case the adapter gating still enables
+    # image output mode while these hooks quietly stop existing.
+    try:
+        import plugins.vlm_image_answer  # noqa: F401
+    except ImportError:
+        pass
+    else:
+        logger.warning(
+            "plugins.vlm_image_answer is present but backend_hooks failed "
+            "to import — VLM profile warnings, deploy-time validation and "
+            "re-extraction invalidation are disabled while image output "
+            "mode remains enabled"
+        )
 
 
 def get_profile_vision_warning(profile_manager: Any) -> str | None:
--- a/backend/prompt_studio/vlm_utils.py
+++ b/backend/prompt_studio/vlm_utils.py
@@ -1,0 +1,71 @@
+"""Bridge helpers for the cloud-only VLM image-answer feature. No-ops in OSS.
+
+Image output mode is answered by a vision LLM through the cloud-only
+``vlm-image-answer`` plugin. The backend touch points below (profile-save
+vision warning, deploy-time validation, answer-cache invalidation on
+re-extraction) delegate to ``plugins.vlm_image_answer.backend_hooks`` when
+that cloud package is present and degrade to no-ops when it is not — OSS
+additionally hides the image output mode entirely via
+``adapter_processor_v2.image_output_gating``.
+"""
+
+import logging
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+try:
+    from plugins.vlm_image_answer import backend_hooks as _hooks
+
+    VLM_IMAGE_ANSWER_AVAILABLE = True
+except ImportError:
+    _hooks = None
+    VLM_IMAGE_ANSWER_AVAILABLE = False
+
+
+def get_profile_vision_warning(profile_manager: Any) -> str | None:
+    """Non-blocking warning when an image-mode profile's LLM lacks vision.
+
+    Returns a human-readable warning string, or None (always None in OSS).
+    Never raises — a warning must not break profile save/read.
+    """
+    if not VLM_IMAGE_ANSWER_AVAILABLE:
+        return None
+    try:
+        return _hooks.get_profile_vision_warning(profile_manager)
+    except Exception:
+        logger.exception("VLM vision warning check failed; skipping warning")
+        return None
+
+
+def validate_workflow_for_deployment(workflow: Any) -> None:
+    """Deploy-time guard: reject deployments that cannot serve image mode.
+
+    The cloud hook raises ``rest_framework.serializers.ValidationError``
+    for a definitive misconfiguration (e.g. image-mode profile with a
+    known non-vision LLM); OSS is a no-op (image mode is gated off).
+    """
+    if not VLM_IMAGE_ANSWER_AVAILABLE:
+        return
+    _hooks.validate_workflow_for_deployment(workflow)
+
+
+def invalidate_vlm_answers_on_reextraction(
+    document_id: str, profile_manager: Any, extract_file_path: str
+) -> None:
+    """Invalidate stored VLM answers after a re-extraction rewrote pages/.
+
+    Called from the extraction choke point right after a successful
+    (non-cache-hit) extraction. Never raises — invalidation failure must
+    not fail the extraction itself; the cloud hook logs and degrades.
+    """
+    if not VLM_IMAGE_ANSWER_AVAILABLE:
+        return
+    try:
+        _hooks.invalidate_vlm_answers_on_reextraction(
+            document_id=document_id,
+            profile_manager=profile_manager,
+            extract_file_path=extract_file_path,
+        )
+    except Exception:
+        logger.exception("VLM answer invalidation failed after re-extraction")
--- a/frontend/src/components/custom-tools/add-llm-profile/AddLlmProfile.jsx
+++ b/frontend/src/components/custom-tools/add-llm-profile/AddLlmProfile.jsx
@@ -360,6 +360,16 @@ function AddLlmProfile({
           type: "success",
           content: "Saved successfully",
         });
+        // Backend-computed advisory (e.g. image output mode selected with
+        // an LLM that may not support vision). Absent in OSS responses.
+        if (data?.vision_warning) {
+          setAlertDetails({
+            type: "warning",
+            title: "Check LLM compatibility",
+            content: data.vision_warning,
+            duration: 10,
+          });
+        }
 
         if (newLlmProfiles?.length === 1) {
           // Set the first LLM profile as default
--- a/frontend/src/components/custom-tools/add-llm-profile/AddLlmProfile.jsx
+++ b/frontend/src/components/custom-tools/add-llm-profile/AddLlmProfile.jsx
@@ -356,18 +356,21 @@ function AddLlmProfile({
           llmProfiles: newLlmProfiles,
         };
         updateCustomTool(updatedState);
-        setAlertDetails({
-          type: "success",
-          content: "Saved successfully",
-        });
-        // Backend-computed advisory (e.g. image output mode selected with
-        // an LLM that may not support vision). Absent in OSS responses.
+        // Single alert: the store holds one alertDetails object, so two
+        // synchronous calls would batch and only the last would render.
+        // vision_warning is a backend-computed advisory (image output
+        // mode with an LLM that may not support vision); absent in OSS.
         if (data?.vision_warning) {
           setAlertDetails({
             type: "warning",
-            title: "Check LLM compatibility",
+            title: "Saved — check LLM compatibility",
             content: data.vision_warning,
             duration: 10,
+          });
+        } else {
+          setAlertDetails({
+            type: "success",
+            content: "Saved successfully",
           });
         }
 
--- a/unstract/sdk1/src/unstract/sdk1/adapters/x2text/constants.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/x2text/constants.py
@@ -39,14 +39,13 @@ class ImageOutputConstants:
     # The adapter (writer) persists one PNG per page under a ``pages``
     # subfolder as ``page_NNN.png`` (zero-padded to PAGE_NUMBER_PADDING
     # digits; four or more digits appear naturally past page 999). Readers
-    # discover pages with PAGE_GLOB_PATTERN and MUST order them by the
-    # integer captured by PAGE_NUMBER_REGEX — never lexicographically,
-    # which silently misorders once page numbers outgrow the padding.
+    # list the directory and MUST order pages by the integer captured by
+    # PAGE_NUMBER_REGEX — never lexicographically, which silently
+    # misorders once page numbers outgrow the padding.
     PAGES_SUBFOLDER = "pages"
     PAGE_IMAGE_PREFIX = "page_"
     PAGE_IMAGE_EXTENSION = ".png"
     PAGE_NUMBER_PADDING = 3
-    PAGE_GLOB_PATTERN = "page_*.png"
     # First capture group is the numeric page index (as a string, possibly
     # zero-padded) — cast to int before sorting.
     PAGE_NUMBER_REGEX = r"page_(\d+)\.png"
--- a/unstract/sdk1/src/unstract/sdk1/adapters/x2text/constants.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/x2text/constants.py
@@ -35,7 +35,43 @@ class ImageOutputConstants:
         "Please provide a PDF file or select a text output mode."
     )
 
+    # --- Page image storage layout (writer/reader contract) ---
+    # The adapter (writer) persists one PNG per page under a ``pages``
+    # subfolder as ``page_NNN.png`` (zero-padded to PAGE_NUMBER_PADDING
+    # digits; four or more digits appear naturally past page 999). Readers
+    # discover pages with PAGE_GLOB_PATTERN and MUST order them by the
+    # integer captured by PAGE_NUMBER_REGEX — never lexicographically,
+    # which silently misorders once page numbers outgrow the padding.
+    PAGES_SUBFOLDER = "pages"
+    PAGE_IMAGE_PREFIX = "page_"
+    PAGE_IMAGE_EXTENSION = ".png"
+    PAGE_NUMBER_PADDING = 3
+    PAGE_GLOB_PATTERN = "page_*.png"
+    # First capture group is the numeric page index (as a string, possibly
+    # zero-padded) — cast to int before sorting.
+    PAGE_NUMBER_REGEX = r"page_(\d+)\.png"
+
     @staticmethod
     def is_pdf(file_name: str) -> bool:
         """Return True when ``file_name`` is a PDF (case-insensitive suffix)."""
         return Path(file_name).suffix.lower() == ImageOutputConstants.PDF_EXTENSION
+
+
+def build_page_store_dir(output_file_path: str | None, input_file_path: str) -> str:
+    """Per-document folder for page images: ``{extract_dir}/{stem}/pages``.
+
+    The single canonical derivation shared by the adapter (writer) and any
+    page-image reader, so both sides agree on the location without metadata
+    persistence or a manifest sidecar. Keyed on the document ``stem`` (the
+    same discriminator the extract ``.txt`` files alongside use), not the
+    per-run whisper_hash. This is collision-safe against concurrent documents
+    in the same project, is reconstructible from ``output_file_path`` alone,
+    and — being stable across runs — makes a re-extraction overwrite its own
+    pages instead of orphaning a fresh tree in FileStorage on every run.
+
+    Pure and deterministic: no I/O, no lookups.
+    """
+    reference = output_file_path or input_file_path
+    base_dir = str(Path(reference).parent) if reference else "."
+    stem = Path(reference).stem if reference else "document"
+    return str(Path(base_dir) / stem / ImageOutputConstants.PAGES_SUBFOLDER)
--- a/unstract/sdk1/src/unstract/sdk1/adapters/x2text/constants.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/x2text/constants.py
@@ -51,10 +51,20 @@ class ImageOutputConstants:
     # zero-padded) — cast to int before sorting.
     PAGE_NUMBER_REGEX = r"page_(\d+)\.png"
 
+    # Leading bytes of every PDF file — the content-based check for inputs
+    # whose storage name carries no extension (workflow executions store the
+    # source file under an extension-less name like ``SOURCE``).
+    PDF_MAGIC_BYTES = b"%PDF-"
+
     @staticmethod
     def is_pdf(file_name: str) -> bool:
         """Return True when ``file_name`` is a PDF (case-insensitive suffix)."""
         return Path(file_name).suffix.lower() == ImageOutputConstants.PDF_EXTENSION
+
+    @staticmethod
+    def is_pdf_bytes(header: bytes) -> bool:
+        """Return True when ``header`` starts with the PDF magic bytes."""
+        return bytes(header).startswith(ImageOutputConstants.PDF_MAGIC_BYTES)
 
 
 def build_page_store_dir(output_file_path: str | None, input_file_path: str) -> str:
--- a/unstract/sdk1/src/unstract/sdk1/adapters/x2text/page_image_loader.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/x2text/page_image_loader.py
@@ -113,6 +113,17 @@ def discover_page_images(fs: FileStorage, page_store_dir: str) -> list[tuple[int
         PageImagesNotFoundError: directory missing or no page images in it.
         PageImageSetIncompleteError: duplicate or missing page numbers.
     """
+    # Object-store backends serve listings from fsspec's directory cache in
+    # long-lived worker processes; a page purged since the last listing would
+    # still "exist" here and only blow up at read time. Refresh the cache
+    # first so discovery reflects reality (no-op for backends without one).
+    invalidate = getattr(getattr(fs, "fs", None), "invalidate_cache", None)
+    if callable(invalidate):
+        try:
+            invalidate(page_store_dir)
+        except Exception:  # pragma: no cover - cache refresh is best-effort
+            logger.debug("Could not invalidate listing cache for %s", page_store_dir)
+
     try:
         entries = fs.ls(page_store_dir) if fs.exists(page_store_dir) else None
     except FileNotFoundError:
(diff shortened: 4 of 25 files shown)


workers/executor/executors/vlm_image_answer.py
```diff
@@ -0,0 +1,201 @@
...
+) -> dict[str, Any] | None:
+    """Return the x2text adapter config when it is in image mode, else None.
+
+    Resolution goes through the platform service (the payload has no
+    adapter metadata); results are cached per (execution, adapter).
+    """
+    cache_key = (execution_id, adapter_instance_id)
+    if cache_key in _MODE_CACHE:
+        _MODE_CACHE.move_to_end(cache_key)
+        return _MODE_CACHE[cache_key]
```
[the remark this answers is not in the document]
praveen-formido: Fixed in 5151efafa. Valid catch — IDE payloads carry no `execution_id`, so all IDE runs shared one `("", adapter)` entry and an adapter edit between runs kept serving the stale mode. The cache key now scopes to `execution_id or run_id` (still deduplicating the N per-prompt resolutions within one run — the cache's actual purpose — with no cross-run reuse), and caching is skipped entirely when no scope id exists. Covered by `test_run_id_scopes_cache_when_execution_id_missing` and `test_no_scope_id_never_caches`.

unstract/sdk1/src/unstract/sdk1/adapters/x2text/page_image_loader.py
```diff
@@ -0,0 +1,256 @@
...
+                "page set is incomplete. Re-extract the document with cache "
+                "bypass to regenerate it (billed per page).",
+                page_store_dir=page_store_dir,
+                found_pages=[n for n, _ in discovered if n != page_number],
+                missing_pages=[page_number],
+            ) from e
+        encoded = base64.b64encode(bytes(data)).decode("ascii")
+        loaded.append(
+            LoadedPageImage(page_number=page_number, path=path, base64_data=encoded)
+        )
```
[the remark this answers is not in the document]
praveen-formido: Fixed in 5151efafa. `load_page_images` now enforces an aggregate raw-byte budget (default 50MB, `max_total_bytes=None` to disable) while reading — it stops at the page that crosses the budget, before encoding/retaining more, and raises the typed `PageImageSetTooLargeError` with page-range remediation. The bridge maps it to a distinct `IMAGE_PAGES_TOO_LARGE` error code. Covered by `TestByteBudget` (stop-at-page accounting, disable, message) and the bridge mapping test.

praveen-formido:
**Re: Confidence 3/5 — "an oversized individual page can still cause an uncontrolled worker-memory allocation before the new byte limit rejects it"**

Addressed in 46c34f2fd. The byte budget is now enforced **from storage size metadata (object HEAD) before any read** — an oversized page set is rejected with zero image bytes entering worker memory. The read-time accounting remains as a belt-and-braces guard for backends without a `size()` API and for stat/read races. Regression tests pin all three paths: metadata pre-check rejects with `reads == []`, no-`size()` backends still enforce at read time, and unreadable metadata falls back cleanly.

workers/executor/executors/vlm_image_answer.py
```diff
@@ -0,0 +1,213 @@
...
+            "This document was extracted in image output mode, which is "
+            "answered by a vision LLM available only on Unstract Cloud. "
+            "Switch the profile's text extractor to a text output mode, "
+            "or run this on Unstract Cloud.",
+            error_code=IMAGE_OUTPUT_REQUIRES_CLOUD,
+        )
+
+    shim.stream_log(f"Answering `{prompt_name}` from page images via vision LLM")
+    fs = FileUtils.get_fs_instance(execution_source=execution_source)
+    page_store_dir = build_page_store_dir(file_path, file_path)
```
chandrasekharan-zipstack: **1. `file_path` is not always the extract path — summarize and smart-table silently break the pages lookup.** (blocker)

This derives the pages directory from the executor's `file_path`, but `answer_params["file_path"]` gets rewritten by the pipeline in two places before the answer step runs:

- `legacy_executor.py:898` — summarization enabled → `summarize_file_path`
- `legacy_executor.py:658` and `:791` — `skip_extraction` / smart table → the original source file

The writer derived the directory from `output_file_path` (the extract `.txt`). So with summarize on, the reader looks in `{summarize_dir}/{hash}/pages`, finds nothing, and raises `IMAGE_OUTPUT_MISSING` — whose remediation copy tells the user to *re-extract with cache bypass, billed per page*. That will not fix it, and they pay to find that out.

Same hazard class as single-pass, and the mechanism is already here: `raise_if_image_mode_unsupported`. Either reject summarize (and `skip_extraction`) for image mode explicitly, or thread the un-rewritten extract path through to the bridge so the derivation cannot be poisoned.

workers/executor/executors/vlm_image_answer.py
```diff
@@ -0,0 +1,213 @@
...
+    adapter metadata); results are cached per (scope, adapter). With no
+    scope id at all, caching is skipped entirely — a shared ("", adapter)
+    entry would serve a stale mode to every later run on this worker.
+    """
+    cache_key = (scope_id, adapter_instance_id)
+    if scope_id and cache_key in _MODE_CACHE:
+        _MODE_CACHE.move_to_end(cache_key)
+        return _MODE_CACHE[cache_key]
+
+    config = PlatformHelper.get_adapter_config(shim, adapter_instance_id) or {}
```
chandrasekharan-zipstack, line 85: **2. Every prompt run now depends on the platform service, including pure text mode.**

`run_vlm_image_answer` is called unconditionally at the top of `_execute_single_prompt`, so any prompt carrying an `x2text_adapter` triggers this call once per (execution, adapter). Text-mode runs never made a platform call in `answer_prompt` before. A platform-service blip now fails prompts that have no stake in image mode.

The caching itself is well-reasoned — run-scoped so an adapter edit lands on the next run, and no-scope skips caching entirely rather than pinning a stale mode forever. My question is the sourcing: the extract step in the same pipeline already resolves the x2text adapter, so could the output mode be forwarded via `answer_params`, leaving this platform call as the fallback only for a standalone "run prompt" against a cached extraction?

If that is not workable, fine — but then the fail-closed behaviour should be a stated decision rather than a consequence of where the call sits.

workers/executor/executors/legacy_executor.py
```diff
@@ -1769,9 +1770,27 @@ def _execute_single_prompt(
...
+                metadata=metadata,
+                metrics=metrics,
+                usage_kwargs=usage_kwargs,
+            )
             retrieval_strategy = output.get(PSKeys.RETRIEVAL_STRATEGY)
             valid_strategies = {s.value for s in RetrievalStrategy}
-            if retrieval_strategy in valid_strategies:
+            if vlm_answer is not None:
+                answer = vlm_answer
+                metadata[PSKeys.CONTEXT][prompt_name] = []
```
chandrasekharan-zipstack, line 1802: **3. Challenge and evaluation run against empty context in image mode.**

`context_list` stays `[]` on this branch, and both `_run_challenge_if_enabled` and `_run_evaluation_if_enabled` further down still execute with it. Challenge asks a second LLM to verify the answer *against the context* — handed none, it will either reject everything or emit noise, and the user is billed for the call either way.

The PR description says challenge "runs unchanged", which is mechanically true but not semantically. Either skip both in image mode, or hand challenge the page images too.

workers/executor/executors/legacy_executor.py
```diff
@@ -1769,9 +1770,27 @@ def _execute_single_prompt(
         records: list[dict[str, Any]] = []
         try:
             answer = "NA"
+            # Image output mode: the document has page images, not text —
+            # the answer comes from a vision LLM (cloud plugin) and RAG
+            # retrieval is skipped entirely. Returns None when the
+            # prompt's x2text adapter is not in image mode; raises rather
+            # than falling through when it is but cannot be served.
+            vlm_answer = run_vlm_image_answer(
```
chandrasekharan-zipstack: **4. Retrieval adapters and indexing still run for image-mode profiles.**

`_init_llm_and_retrieval` runs before this call, so with `chunk_size > 0` the embedding and vector DB adapters are constructed for a prompt that will never retrieve. Pipeline Step 3 also indexes the one-line extraction summary. Net: wasted embedding spend, a polluted vector store, and an unreachable vector DB failing an image-mode prompt for a reason unrelated to it.

Given the base branch already carries a "survive indexing" commit this may be a known accepted cost for now — worth confirming it is deliberate rather than incidental.

frontend/src/components/custom-tools/add-llm-profile/AddLlmProfile.jsx
```diff
@@ -360,6 +360,16 @@ function AddLlmProfile({
           type: "success",
           content: "Saved successfully",
         });
+        // Backend-computed advisory (e.g. image output mode selected with
+        // an LLM that may not support vision). Absent in OSS responses.
+        if (data?.vision_warning) {
```
chandrasekharan-zipstack, line 363: **5. The "Saved successfully" toast is swallowed whenever a vision warning fires.**

Two synchronous `setAlertDetails` calls, but the store holds a single `alertDetails` object (`store/alert-store.js`) and `App.jsx` renders off `useEffect(..., [alertDetails])`. React batches both updates into one render, the effect fires once with the final value, and only the warning notification is shown — the success confirmation never appears.

Fold it into a single alert ("Saved — check LLM compatibility") or defer the second call.

backend/prompt_studio/vlm_utils.py
```diff
@@ -0,0 +1,71 @@
...
+import logging
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+try:
+    from plugins.vlm_image_answer import backend_hooks as _hooks
+
+    VLM_IMAGE_ANSWER_AVAILABLE = True
+except ImportError:
```
chandrasekharan-zipstack, line 21: **6. This cannot distinguish "running OSS" from "cloud hooks are broken".**

The `except ImportError` is silent. Note the asymmetry with the sibling gate: `adapter_processor_v2.image_output_gating._consumer_plugin_available()` imports the *package* (`plugins.vlm_image_answer`), while this imports the *submodule* (`.backend_hooks`). So on cloud, a `backend_hooks` import failure — a bad sdk1 import, a typo, a partial deploy — leaves image output mode fully enabled in the schema while the deploy-time guard and the profile vision warning quietly stop existing.

One `logger.warning` in the except closes the gap, or check package presence first and log distinctly when the submodule import is what failed.

unstract/sdk1/src/unstract/sdk1/adapters/x2text/constants.py
```diff
@@ -35,7 +35,53 @@ class ImageOutputConstants:
...
+    # subfolder as ``page_NNN.png`` (zero-padded to PAGE_NUMBER_PADDING
+    # digits; four or more digits appear naturally past page 999). Readers
+    # discover pages with PAGE_GLOB_PATTERN and MUST order them by the
+    # integer captured by PAGE_NUMBER_REGEX — never lexicographically,
+    # which silently misorders once page numbers outgrow the padding.
+    PAGES_SUBFOLDER = "pages"
+    PAGE_IMAGE_PREFIX = "page_"
+    PAGE_IMAGE_EXTENSION = ".png"
+    PAGE_NUMBER_PADDING = 3
+    PAGE_GLOB_PATTERN = "page_*.png"
```
chandrasekharan-zipstack: **NIT** — small, non-blocking cleanups, none of them merge-gating:

- **`PAGE_GLOB_PATTERN` is dead.** No reader uses it — `page_image_loader` does `fs.ls` + `PAGE_NUMBER_REGEX`, never a glob. The only references are this module's own docstring and two tests in `test_x2text_shared_page_path.py` that pin the constant to itself. Delete all three.
- **The byte budget is enforced twice** — a duck-typed `fs.size` pre-pass plus read-time accounting, roughly 40 lines and 6 tests. The page cap already bounds the count at 20, and `max_total_bytes` is never wired to env like the cap is, so it is fixed at 50MB regardless. The read-time check alone covers this; add the metadata pre-pass when a single 50MB page render actually turns up.
- **The `PageImagesNotFoundError` message is duplicated verbatim** at both raise sites in `discover_page_images` — hoist it to a local.
- **Function-level imports in `_handle_single_pass_extraction`** (`legacy_executor.py:2315`): `PSKeys` is already imported at module top, and `run_vlm_image_answer` from that same module is imported at line 29. No circular-dependency reason, and CLAUDE.md asks for imports at the top.
- **Duplicate page numbers** raise `PageImageSetIncompleteError` with `missing_pages=[]`, which the bridge maps to `IMAGE_OUTPUT_MISSING`. The message says "corrupt", the code says "missing" — the type is doing two jobs. Low priority.
- **Cross-repo (unstract-cloud#1690):** the `pytest.importorskip` guards in `workers/plugins/vlm_image_answer/tests/test_base.py` and `backend/plugins/vlm_image_answer/tests/test_backend_hooks.py` are correct today, but will skip silently forever if those sdk1 modules ever move. Worth a follow-up to strip them once this PR lands.

chandrasekharan-zipstack (commented):
Reviewed both halves (#2218 + Zipstack/unstract-cloud#1690) against the live code paths.

The architecture holds up well. Binding writer and reader to *one* `build_page_store_dir` (with a test asserting `is` identity, not equality) is the right way to make a two-sided storage contract undriftable, and the fail-loud-never-fall-through error policy is applied consistently across the bridge, the loader and the gating. Gate-before-read with a test asserting `fs.read_calls == []`, and the error-code prefixing to survive string-only `ExecutionResult.error` propagation, are both nicely done.

Verified and holding: `fs.ls` / `fs.size` / `fs.read(length=)` all exist on `FileStorage`; `complete_vision` / `get_model_name` / `get_metrics` match the SDK1 signatures; the TOCTOU `except FileNotFoundError` works because `skip_local_cache` re-raises `FileNotFoundError` after its invalidate-and-retry, and the new `invalidate_cache` in discovery covers the stale-*positive* listing case that decorator does not; the DRF dict-detail fix is the correct shape for a raise inside `to_internal_value`.

Six comments inline plus one NIT roll-up. My ranking: **#1 is a blocker** — user-reachable, misleading, and costs money to discover. **#2 and #3 are design decisions I would like stated** rather than left as defaults. Everything else is cleanup.

workers/executor/executors/vlm_image_answer.py
praveen-formido: Fixed in bbdeaad11 — you're right, this was the real blocker. The payload builders now stamp an explicit `extract_file_path` captured **before** any rewrite (IDE single + bulk builders capture it ahead of the summarize override; `structure_tool_task` stamps `extracted_input_file`, which the summarize/smart-table mutations never touch), and the bridge derives `pages/` from that key with the old `file_path` as fallback for older payloads. Pinned by `test_pages_dir_keys_on_extract_path_not_rewritten_file_path`.

workers/executor/executors/vlm_image_answer.py
praveen-formido: Fixed in bbdeaad11, exactly along the line you suggested: the backend now stamps the adapter's `output_mode` per prompt (`_stamp_x2text_output_mode` — it already holds the decrypted metadata for the config hash), so detection is payload-only and text-mode prompts make **zero** platform calls. The run-scoped platform resolution remains solely as the fallback for unstamped payloads (API deployments build `outputs` worker-side without metadata access, and older in-flight payloads). Covered by `TestStampedDetection` (stamped image/text/absent × platform-call assertions).

workers/executor/executors/legacy_executor.py
praveen-formido: Fixed in bbdeaad11 — challenge and evaluation are now skipped in image mode with a visible stream-log line ("Skipped challenge/evaluation … image output mode has no retrieval context"). Agreed that "runs unchanged" was mechanically true but a billed, doomed second LLM call semantically. Handing challenge the page images is the right future shape but it's a product decision (vision challenge = double image-token cost per prompt) — deferred to the windowing/production phase rather than invented here.

workers/executor/executors/legacy_executor.py
praveen-formido: Half fixed, half deliberate — called out separately now:

- **Adapter construction (fixed in bbdeaad11):** detection moved ahead of `_init_llm_and_retrieval`, which now runs LLM-only (`chunk_size=0` path) for image-mode prompts — no embedding/vector DB construction, and an unreachable vector DB can no longer fail an image-mode prompt.
- **Indexing the summary (deliberate):** the one-line summary as `extracted_text` is what keeps the extraction cache consistent (a re-run is a cache hit — no re-submit to LLMWhisperer, no per-page re-billing) and keeps the document present in doc coverage. The indexed footprint is a single tiny chunk per doc. Documented as a locked design decision in the PR body; happy to revisit if the vector-store pollution ever matters in practice.

frontend/src/components/custom-tools/add-llm-profile/AddLlmProfile.jsx
praveen-formido: Fixed in bbdeaad11 — single alert now: `Saved — check LLM compatibility` (warning, with the backend copy) when `vision_warning` is present, plain `Saved successfully` otherwise. Good catch on the store/batching interaction; the folded copy also reads better than two stacked toasts.

backend/prompt_studio/vlm_utils.py
praveen-formido: Fixed in bbdeaad11 — on `ImportError`, `vlm_utils` now probes the package: absent → silent (OSS, expected); present but `backend_hooks` failed → a distinct `logger.warning` spelling out exactly what's broken ("profile warnings, deploy-time validation and re-extraction invalidation disabled while image output mode remains enabled"). The asymmetry with `image_output_gating`'s package-level probe was precisely the gap — thanks.

unstract/sdk1/src/unstract/sdk1/adapters/x2text/constants.py
praveen-formido: Acknowledged, leaving these as-is for this round per the non-blocking flag — with one note on the byte budget: the metadata pre-pass was added at Greptile's explicit insistence in the previous round ("an oversized individual page can still cause an uncontrolled worker-memory allocation before the byte limit rejects it") — the read-time check alone does pull each full object into memory before accounting. If you'd rather drop the pre-pass (or the whole budget down to read-time only) I'll follow your call — you outrank the bot. `PAGE_GLOB_PATTERN` removal and the duplicated `PageImagesNotFoundError` message + env wiring for `max_total_bytes` are all fair; happy to fold them into the next commit if you want them in this PR rather than a follow-up.

praveen-formido: All three folded in after your call, in 0c4fd750f (OSS) + 61449599 (cloud): dead `PAGE_GLOB_PATTERN` deleted (readers list + regex, never glob); the duplicated not-found message extracted to one helper; `VLM_IMAGE_ANSWER_MAX_TOTAL_MB` env now overrides the byte budget like the page cap's env does. On the double enforcement: resolved the Greptile-vs-you tension with a mechanism you'll both prefer — the metadata pre-pass is **gone**, replaced by bounded reads (`length = remaining + 1` per read), which is less code than either previous layer and gives a hard `budget + 1` allocation ceiling that the read-time accounting alone never had.

chandrasekharan-zipstack (approved):

PR #2222, opened 2026-07-30
feat(org): make the default organization name configurable (#1749)

### LINKED ISSUE #1749: feat: [ISSUE] change default organization name "mock_org" to a meaningful word
**Is your feature request related to a problem? Please describe.**
is it possible for us to change default organization "mock_org"?

I'd like to change default organization name "mock_org" in the URL to a meaningful word, e.g. us_org, 
for example, change below URL:
http://frontend.unstract.localhost/mock_org/tools
to:
http://frontend.unstract.localhost/us_org/tools

**Describe the solution you'd like**
could you add a config to let us to change the "mock_org"?

**Describe alternatives you've considered**

**Additional context**

aniketshukla1:
## Summary

The default organization for the open-source (single-tenant) deployment was hardcoded as `mock_org` — it fixes the org segment of every URL (e.g. `/mock_org/tools`) and also doubles as the frontend's open-source sentinel. This makes the name configurable while staying fully backward-compatible.

Closes #1749.

## Changes

**Backend**
- Add `DEFAULT_ORGANIZATION_NAME` (env var, default `mock_org`) in `settings/base.py`, mirroring the existing `DEFAULT_AUTH_USERNAME` pattern.
- `DefaultOrg.ORGANIZATION_NAME` / `MOCK_ORG` now read that setting. The org name already flows to the frontend via `sessionDetails.orgName`, so URLs follow automatically.

**Frontend**
- Expose `defaultOrgName` through the runtime config (`generate-runtime-config.sh`) and `config.js` (env `VITE_DEFAULT_ORG_NAME`, default `mock_org`).
- The four `isOpenSource = orgName === "mock_org"` checks (top nav bar, admin route guard, PostHog events, administration page) now compare against `config.defaultOrgName` instead of the literal string, so open-source detection stays correct after a rename.

**Docs**
- Document both env vars in `backend/sample.env` and `frontend/sample.env`.

## Backward compatibility

Every default remains `mock_org`, so existing deployments are unaffected. To rename the org, set `DEFAULT_ORGANIZATION_NAME` (backend) and `VITE_DEFAULT_ORG_NAME` (frontend) to the same value.

## Testing

- Backend compiles; the settings → constants wiring mirrors the existing `DEFAULT_AUTH_USERNAME`.
- Frontend passes `biome check` (import organization + formatting) on all changed files.


aniketshukla1:
Thanks @greptileai and @coderabbitai for the review. Pushed `b155ce9` addressing all findings:

**1. Unvalidated names break org routing (Greptile · P1)**
`DEFAULT_ORGANIZATION_NAME` is now validated in `backend/backend/settings/base.py` to be a single URL-safe segment (`^[A-Za-z0-9_-]+$`), raising `ValueError` on anything containing `/`, whitespace or other delimiters. This stops a value from being accepted that the org middleware (`^/api/v[12]/unstract/(?P<org_id>[^/]+)/`) and the frontend route can't resolve.

**2. Renaming strands persisted tenant data (Greptile · P1)**
This value *is* the persisted `organization_id` and the URL segment, so an automatic rename migration can't safely distinguish a fresh install with a custom name from an existing install being renamed, and rewriting a tenant identifier risks data loss. Instead it's now documented as **set-once**, with an explicit data-stranding warning in both `settings/base.py` and `backend/sample.env` noting that migrating existing records to a new identifier must be done manually. Glad to add an explicit, opt-in management command for renaming if you'd prefer that over documentation.

**3. Keep backend and frontend configuration in lockstep (CodeRabbit · Major)**
The backend now normalizes blank/whitespace values to `mock_org` (`os.environ.get("DEFAULT_ORGANIZATION_NAME", "").strip() or "mock_org"`), matching the frontend default. An empty backend value can no longer diverge from the frontend and misclassify the open-source org as SaaS. I intentionally kept the frontend's `mock_org` fallback rather than failing fast, because the zero-config default deployment never sets `VITE_DEFAULT_ORG_NAME` and must keep resolving to `mock_org`; normalizing both sides to the same default is what puts them in lockstep.

**4. Escape `defaultOrgName` before embedding it in JavaScript (CodeRabbit · Major)**
`frontend/generate-runtime-config.sh` now passes the resolved value through the existing `js_escape` helper (preserving the `VITE_ → REACT_APP_ → mock_org` fallback order), so a value containing `"` or `\` can no longer generate invalid JavaScript.

Verified before pushing: `base.py` byte-compiles and passes `ruff`; the validation accepts valid slugs and rejects `/`, whitespace, dots and non-ASCII; `generate-runtime-config.sh` passes `sh -n` and escapes `"`/`\` correctly.

@greptileai review
@coderabbitai review

aniketshukla1:
Pushed `22611ee` addressing the remaining P1 (**Whitespace desynchronizes OSS detection**, `generate-runtime-config.sh:22`).

Root cause: the backend now `.strip()`s `DEFAULT_ORGANIZATION_NAME`, but the frontend preserved surrounding whitespace — so matching values like `" acme "` resolved to `acme` on the backend vs `" acme "` on the frontend, and the `orgName === config.defaultOrgName` check misclassified the open-source deployment as SaaS.

Both frontend paths now normalize identically to the backend:

- **`generate-runtime-config.sh`** (container/runtime): trims the resolved value via a `trim()` helper before `js_escape`, with a blank → `mock_org` fallback.
- **`config.js`** (dev/build): trims the resolved value with the same blank → `mock_org` fallback.

Verified: `sh -n` passes and whitespace/quote/empty inputs normalize correctly (`"  acme  "` → `acme`; whitespace-only/empty → `mock_org`; `"` and `\` still escaped); `config.js` parses as an ES module.

@greptileai review
@coderabbitai review

PR #2229, opened 2026-08-07
UN-2123 [FEAT] Propagate request_id across services and workers

Deepak-Kesavan:
## What

Completes end-to-end `request_id` (`X-Request-ID`) correlation so a **single ID
can filter logs across the backend, Celery workers and services** in gcloud.

- **backend → workers:** a `before_task_publish` signal
  (`backend/backend/celery_signals.py`) injects the request-scoped `request_id`
  (from `StateStore`) into every published Celery task's message headers — no
  per-`send_task` changes required.
- **workers:** `task_prerun` now prefers an explicit `request_id` from the
  message headers over payload-derived ids (`file_execution_id`, etc.), and a
  new `before_task_publish` handler re-propagates it onto downstream
  worker→worker task chains.
- **workers → backend:** the internal-API HTTP client now forwards
  `X-Request-ID` from the worker log context, so backend callbacks share the
  originating request's id.
- **x2text-service:** reads/mints `X-Request-ID` and logs it via a
  self-contained logging module (no new dependency).
- **logging:** all services unified onto one canonical log format so
  `request_id` / `trace_id` / `span_id` parse identically in gcloud.

## Why

Debugging across services is painful when logs can't be correlated. The backend
already assigned a `request_id` per request (and the frontend forwards one), but
it died at the backend boundary — it was never propagated to the Celery workers
or echoed on worker→backend calls, so worker logs showed `request_id:-` (or an
`execution_id` that didn't match the originating request). This threads the same
id through the whole chain.

Jira: [UN-2123](https://zipstack.atlassian.net/browse/UN-2123)

## How

The worker-side receiving/logging machinery already existed (UN-3435) and was
explicitly designed to accept a real `request_id` from the producer — this PR
supplies the missing producer side and closes the remaining hops:

1. `before_task_publish` (backend) reads `StateStore.get(Common.REQUEST_ID)` and
   sets `headers["request_id"]`. Guarded so it can never break task publishing.
2. Worker `_bind_task_context` (task_prerun) reads `task.request.request_id`
   (Celery 5.6.2 exposes custom message headers on the task `Context`; a raw
   `request.headers` mapping is used as a version-safe fallback). Resolution
   order: message-header `request_id` → payload-derived id → Celery `task_id`.
3. Worker `_propagate_request_id_on_publish` (before_task_publish) forwards the
   bound `request_id` onto tasks the worker itself publishes.
4. `base_client._make_request` adds `X-Request-ID` from the worker log context.
5. x2text-service `create_app()` wires `setup_logging` +
   `register_request_id_middleware` from a new self-contained `app/logging_util.py`.
6. The shared Flask formatter (`unstract/core/flask/logging.py`) is aligned to
   the canonical backend/worker format.

**Out of scope (recommended as separate tickets):**
- **LLM Whisperer (LLMW)** adoption of this standard — separate repo/product.
- **Full OpenTelemetry trace activation.** The OTel scaffolding (`trace_id`
  fields in every formatter, meta-deps) is currently inert — no service runs
  under `opentelemetry-instrument`, exporters are hardcoded to `none`, and there
  is no OTLP endpoint — so `trace_id`/`span_id` always render as `-`. Activating
  real distributed tracing is a larger DevOps/observability effort. This PR uses
  `request_id` (which the frontend/backend already surface via the
  `X-Request-ID` response header) as the pragmatic correlation key.
- **SDK1 x2text-adapter → x2text-service** header forwarding — needs the SDK
  execution `request_id` threaded through several tool-side layers; belongs with
  the tool-execution/OTel correlation follow-up. (x2text still *logs* any
  `X-Request-ID` it receives after this PR.)
- **`tool-sandbox → runner`** currently sends `X-Request-ID = file_execution_id`,
  inconsistent with the HTTP request_id standardized here; reconciling it means
  threading the request_id through tool-sandbox — same tool-execution/OTel bucket.
- **PGMQ transport** (`pg_queue_enabled`) — a separate, non-Celery transport
  (DB-INSERT enqueue + custom poll-loop consumer) that this PR's Celery-signal
  propagation does not cover. Celery is the live default; a matching carrier
  (`request_id` on `TaskPayload`, populated in the PG producers, read into the
  consumer's `apply(headers=…)`) is handed to the PGMQ owner so correlation
  stays complete as `pg_queue_enabled` ramps.

Low risk — the change is additive and defensively guarded:

- Both `before_task_publish` handlers no-op safely (null-guarded, `StateStore`
  read wrapped in try/except) so they **cannot break task publishing** even if
  no `request_id` is in scope (e.g. beat-scheduled tasks fall back to the
  existing `execution_id`/`task_id` behaviour).
- Reading `task.request.request_id` uses `getattr(..., None)` and only overrides
  the previously-working payload-derived id when a header is actually present —
  so existing worker correlation is preserved when the header is absent.
- **One intentional log-format change:** the shared Flask formatter now matches
  the canonical backend/worker format, which means Flask-service log lines
  (platform-service, runner) show `module:<source-module>` instead of the
  logger `name`. Log *content* is otherwise unchanged; any dashboard keying on
  the logger name string would need updating (grep-by-`request_id`/`trace_id` is
  unaffected and is the point of the change).
- No API contract, DB, or schema changes.

## Database Migrations

None.

## Env Config

None. (x2text-service reads the existing `LOG_LEVEL` env if set; defaults to `INFO`.)

## Relevant Docs

-

## Related Issues or PRs

- Jira: [UN-2123](https://zipstack.atlassian.net/browse/UN-2123)
- Builds on the worker-side machinery from UN-3435 (#1932).

## Dependencies Versions

None added. (x2text-service deliberately avoids taking on `unstract-core`; it
uses a small self-contained logging module instead.)

## Notes on Testing

- Static: all touched files `py_compile`-clean; pre-commit (ruff, ruff-format,
  pycln, pyupgrade, secret-scan, test-selection) green.
- Verified against Celery 5.6.2 that a custom key in the message headers is
  exposed on the task `Context` via both attribute and `.get()` access.
- **One runtime smoke test still recommended on a live cluster** (not yet run):
  trigger a workflow/API execution and confirm a single `X-Request-ID` from the
  originating HTTP request appears in the **worker** log lines (`request_id:<id>`)
  and on the worker→backend internal-API calls. If Celery header→Context
  attribute promotion ever regressed, the `request.headers` fallback covers it,
  but the smoke test is the real guarantee.

## Screenshots

N/A (no UI surface).

## Checklist

[UN-2123]: https://zipstack.atlassian.net/browse/UN-2123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/backend/celery_signals.py
+++ b/backend/backend/celery_signals.py
@@ -1,14 +1,8 @@
-"""Celery signal handlers for the backend (producer side).
+"""Celery signal handlers carrying the HTTP ``request_id`` onto published tasks.
 
-Propagates the HTTP ``request_id`` (correlation ID assigned by
-``CustomRequestIDMiddleware``) onto every published Celery task so that worker
-logs can be correlated back to the originating request.
-
-The value is placed in the task message headers under ``request_id``. Workers
-read it from ``task.request`` in ``task_prerun`` and bind it onto their log
-context -- see ``workers/shared/infrastructure/logging/logger.py``. Using the
-``before_task_publish`` signal means this works for *every* ``send_task`` /
-``.delay`` / ``.apply_async`` call with no per-call-site changes.
+The id travels in the task message headers under ``Common.REQUEST_ID``. Hooking
+``before_task_publish`` rather than each producer is deliberate: it covers every
+``send_task`` / ``.delay`` / ``.apply_async`` with no per-call-site change.
 """
 
 import logging
@@ -25,11 +19,9 @@ logger = logging.getLogger(__name__)
 def propagate_request_id(headers=None, **kwargs):
     """Inject the current request_id into the outgoing task's message headers.
 
-    Fires in the producer thread (the web request thread for API-triggered
-    tasks), where ``StateStore`` still holds the request_id set by
-    ``CustomRequestIDMiddleware``. No-ops when there is no request_id in scope
-    (e.g. beat-scheduled publishes), leaving the worker to fall back to its
-    own correlation id (execution_id / task_id).
+    Relies on firing in the *producer* thread, where ``StateStore`` still holds
+    the id set by ``CustomRequestIDMiddleware``. No-ops without one (e.g. beat
+    publishes), leaving the worker to derive its own correlation id.
     """
     if headers is None:
         return
@@ -45,7 +37,6 @@ def propagate_request_id(headers=None, **kwargs):
 
 
 def _request_id_from_task(task) -> str | None:
-    """Read a propagated request_id off a Celery task's message context."""
     request = getattr(task, "request", None)
     if request is None:
         return None
@@ -59,15 +50,13 @@ def _request_id_from_task(task) -> str | None:
 
 @task_prerun.connect
 def bind_request_id(task=None, **kwargs):
-    """Bind the propagated request_id for tasks executed by the backend's OWN
-    Celery workers (beat, dashboard-metric tasks, etc.).
+    """Bind the propagated request_id for tasks run by the backend's own workers.
 
-    The separate ``workers/`` fleet has its own ``task_prerun`` reader; the
-    backend Celery app previously injected the header (``propagate_request_id``)
-    but never consumed it, so backend-executed tasks logged ``request_id:-``.
-    Binding it onto ``log_request_id``'s thread-local makes
-    ``log_request_id.filters.RequestIDFilter`` emit it, and onto ``StateStore``
-    so any task this worker itself publishes re-propagates it.
+    Two writes, both required: ``log_request_id``'s thread-local is the only
+    thing ``log_request_id.filters.RequestIDFilter`` reads (the HTTP middleware
+    is otherwise its sole writer, so a task would log ``request_id:-``), and
+    ``StateStore`` is what ``propagate_request_id`` reads, so tasks this worker
+    itself publishes carry the id onward.
     """
     request_id = _request_id_from_task(task)
     if not request_id:
@@ -81,7 +70,7 @@ def bind_request_id(task=None, **kwargs):
 
 @task_postrun.connect
 def clear_request_id(**kwargs):
-    """Clear the task-scoped request_id bound in ``bind_request_id``."""
+    """Clear the task-scoped request_id -- worker threads are pooled and reused."""
     if hasattr(log_request_id_local, "request_id"):
         try:
             del log_request_id_local.request_id
--- a/backend/backend/celery_signals.py
+++ b/backend/backend/celery_signals.py
@@ -14,7 +14,8 @@ context -- see ``workers/shared/infrastructure/logging/logger.py``. Using the
 import logging
 
 from account_v2.constants import Common
-from celery.signals import before_task_publish
+from celery.signals import before_task_publish, task_postrun, task_prerun
+from log_request_id import local as log_request_id_local
 from utils.local_context import StateStore
 
 logger = logging.getLogger(__name__)
@@ -41,3 +42,52 @@ def propagate_request_id(headers=None, **kwargs):
         return
     if request_id and not headers.get(Common.REQUEST_ID):
         headers[Common.REQUEST_ID] = request_id
+
+
+def _request_id_from_task(task) -> str | None:
+    """Read a propagated request_id off a Celery task's message context."""
+    request = getattr(task, "request", None)
+    if request is None:
+        return None
+    request_id = getattr(request, Common.REQUEST_ID, None)
+    if not request_id:
+        task_headers = getattr(request, "headers", None)
+        if isinstance(task_headers, dict):
+            request_id = task_headers.get(Common.REQUEST_ID)
+    return request_id or None
+
+
+@task_prerun.connect
+def bind_request_id(task=None, **kwargs):
+    """Bind the propagated request_id for tasks executed by the backend's OWN
+    Celery workers (beat, dashboard-metric tasks, etc.).
+
+    The separate ``workers/`` fleet has its own ``task_prerun`` reader; the
+    backend Celery app previously injected the header (``propagate_request_id``)
+    but never consumed it, so backend-executed tasks logged ``request_id:-``.
+    Binding it onto ``log_request_id``'s thread-local makes
+    ``log_request_id.filters.RequestIDFilter`` emit it, and onto ``StateStore``
+    so any task this worker itself publishes re-propagates it.
+    """
+    request_id = _request_id_from_task(task)
+    if not request_id:
+        return
+    log_request_id_local.request_id = request_id
+    try:
+        StateStore.set(Common.REQUEST_ID, request_id)
+    except Exception:
+        logger.debug("Unable to set request_id on StateStore", exc_info=True)
+
+
+@task_postrun.connect
+def clear_request_id(**kwargs):
+    """Clear the task-scoped request_id bound in ``bind_request_id``."""
+    if hasattr(log_request_id_local, "request_id"):
+        try:
+            del log_request_id_local.request_id
+        except AttributeError:
+            pass
+    try:
+        StateStore.clear(Common.REQUEST_ID)
+    except Exception:
+        logger.debug("Unable to clear request_id from StateStore", exc_info=True)
--- a/backend/backend/celery_signals.py
+++ b/backend/backend/celery_signals.py
@@ -1,0 +1,43 @@
+"""Celery signal handlers for the backend (producer side).
+
+Propagates the HTTP ``request_id`` (correlation ID assigned by
+``CustomRequestIDMiddleware``) onto every published Celery task so that worker
+logs can be correlated back to the originating request.
+
+The value is placed in the task message headers under ``request_id``. Workers
+read it from ``task.request`` in ``task_prerun`` and bind it onto their log
+context -- see ``workers/shared/infrastructure/logging/logger.py``. Using the
+``before_task_publish`` signal means this works for *every* ``send_task`` /
+``.delay`` / ``.apply_async`` call with no per-call-site changes.
+"""
+
+import logging
+
+from account_v2.constants import Common
+from celery.signals import before_task_publish
+from utils.local_context import StateStore
+
+logger = logging.getLogger(__name__)
+
+
+@before_task_publish.connect
+def propagate_request_id(headers=None, **kwargs):
+    """Inject the current request_id into the outgoing task's message headers.
+
+    Fires in the producer thread (the web request thread for API-triggered
+    tasks), where ``StateStore`` still holds the request_id set by
+    ``CustomRequestIDMiddleware``. No-ops when there is no request_id in scope
+    (e.g. beat-scheduled publishes), leaving the worker to fall back to its
+    own correlation id (execution_id / task_id).
+    """
+    if headers is None:
+        return
+    try:
+        request_id = StateStore.get(Common.REQUEST_ID)
+    except Exception:
+        # StateStore can raise if CONCURRENCY_MODE is misconfigured; never let
+        # correlation plumbing break task publishing.
+        logger.debug("Unable to read request_id from StateStore", exc_info=True)
+        return
+    if request_id and not headers.get(Common.REQUEST_ID):
+        headers[Common.REQUEST_ID] = request_id
--- a/workers/shared/clients/base_client.py
+++ b/workers/shared/clients/base_client.py
@@ -32,6 +32,20 @@ logger = WorkerLogger.get_logger(__name__)
 APPLICATION_JSON = "application/json"
 
 
+def _current_request_id() -> str | None:
+    """Return the request_id bound on the current worker log context, if any.
+
+    Bound by the ``task_prerun`` handler in the logging module; used to
+    propagate ``X-Request-ID`` onto outbound calls to the backend internal API.
+    Returns ``None`` for the ``"-"`` placeholder so no empty header is sent.
+    """
+    ctx = WorkerLogger.get_context()
+    request_id = getattr(ctx, "request_id", None) if ctx else None
+    if not request_id or request_id == "-":
+        return None
+    return request_id
+
+
 # Single PG-queue rollout flag (same key as pg_queue.flags / executor_rpc).
 _PG_QUEUE_FLAG_KEY = "pg_queue_enabled"
 
@@ -315,6 +329,12 @@ class BaseAPIClient:
                 current_org_id = organization_id or self.organization_id
                 if current_org_id:
                     headers["X-Organization-ID"] = current_org_id
+
+                # Propagate the correlation id back to the backend so worker
+                # callbacks share the originating request's request_id in logs.
+                request_id = _current_request_id()
+                if request_id:
+                    headers["X-Request-ID"] = request_id
 
                 if headers:
                     kwargs["headers"] = headers
--- a/workers/shared/infrastructure/logging/logger.py
+++ b/workers/shared/infrastructure/logging/logger.py
@@ -703,22 +703,67 @@ def _extract_request_id(
     return None
 
 
+def _request_id_from_message(task: Any) -> str | None:
+    """Read an explicit request_id propagated via Celery message headers.
+
+    The task producer (backend ``before_task_publish`` handler, or a worker
+    re-publishing a downstream task) injects ``request_id`` into the message
+    headers. Celery exposes custom headers on ``task.request`` -- as a direct
+    attribute under protocol v2, and via the raw ``headers`` mapping as a
+    version-safe fallback. This is the authoritative cross-service correlation
+    id and takes precedence over payload-derived ids (file_execution_id, etc.).
+    """
+    request = getattr(task, "request", None)
+    if request is None:
+        return None
+    value = getattr(request, "request_id", None)
+    if not value:
+        headers = getattr(request, "headers", None)
+        if isinstance(headers, Mapping):
+            value = headers.get("request_id")
+    return _coerce_id(value)
+
+
 def _bind_task_context(task_id, task, args, kwargs, **_):
     """Celery ``task_prerun`` handler: bind request_id onto the log context.
+
+    Resolution order: an explicit request_id propagated on the message headers
+    (cross-service correlation), then a payload-derived id
+    (``_extract_request_id``), then the Celery ``task_id``.
 
     Catches any extraction failure so a malformed payload can never leave
     the previous task's id bound on the thread.
     """
-    try:
-        request_id = _extract_request_id(args or (), kwargs or {}, task) or task_id
-    except Exception:
-        logging.getLogger(__name__).debug(
-            "request_id extraction failed for task %s; falling back to task_id",
-            task_id,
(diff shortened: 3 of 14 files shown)


x2text-service/app/logging_util.py
```diff
@@ -0,0 +1,92 @@
...
+    " :- %(message)s"
+)
+
+
+class RequestIDFilter(logging.Filter):
+    """Inject the current request's ``request_id`` into log records."""
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        record.request_id = getattr(g, "request_id", "-") if g else "-"
+        return True
```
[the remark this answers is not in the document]
Deepak-Kesavan: Good call — applied in 2fd885a3c. Switched to `has_request_context()` before reading `g`, so log records outside an active request context (e.g. gunicorn startup) deterministically get `request_id:-` rather than relying on `LocalProxy.__bool__` semantics. Thanks!

workers/shared/clients/base_client.py
```diff
@@ -316,6 +330,12 @@ def _make_request(
                 if current_org_id:
                     headers["X-Organization-ID"] = current_org_id
 
+                # Propagate the correlation id back to the backend so worker
+                # callbacks share the originating request's request_id in logs.
+                request_id = _current_request_id()
+                if request_id:
+                    headers["X-Request-ID"] = request_id
```
praveen-formido, line 337: **The worker→backend hop does not actually correlate — the backend drops this header.**

`backend/backend/settings/base.py:279` sets `LOG_REQUEST_ID_HEADER = "X-Request-ID"`, but `django-log-request-id` looks that value up in `request.META`, where WSGI maps HTTP headers to `HTTP_X_REQUEST_ID`:

```python
# log_request_id/middleware.py::_get_request_id
return request.META.get(request_id_header, default_request_id)
```

The lookup therefore always misses, and because `GENERATE_REQUEST_ID_IF_NOT_IN_HEADER = True` the backend mints a fresh uuid4 for every request instead. Verified against the installed package:

```
LOG_REQUEST_ID_HEADER='X-Request-ID',      META={'HTTP_X_REQUEST_ID': 'abc-123-from-worker'} -> 98701742cad145069f9d1df02327af42
LOG_REQUEST_ID_HEADER='HTTP_X_REQUEST_ID', META={'HTTP_X_REQUEST_ID': 'abc-123-from-worker'} -> abc-123-from-worker
```

Concrete scenario: a file-processing worker calls the internal API with `X-Request-ID: <originating id>`; `CustomRequestIDMiddleware` ignores it, `request.id` becomes a new uuid4, `CustomAuthMiddleware` stores *that* in `StateStore`, and the backend log lines for the callback show an id unrelated to the worker's. So the "workers → backend" bullet in the PR description isn't achieved, and the same bug already silently discards the `X-Request-ID` the frontend forwards (`frontend/src/helpers/requestId.js`).

Fix is one line: `LOG_REQUEST_ID_HEADER = "HTTP_X_REQUEST_ID"` (keep `REQUEST_ID_RESPONSE_HEADER = "X-Request-ID"`, which is a real header name and is correct as-is). Worth adding a test that posts `X-Request-ID` and asserts the response echoes the same value.

workers/shared/infrastructure/logging/logger.py
```diff
@@ -703,24 +703,69 @@ def _extract_request_id(
...
+    file-processing task enqueuing a callback). Reads the request_id bound onto
+    the thread-local log context by ``_bind_task_context``; no-ops when absent
+    or when the caller already set the header.
+    """
+    if headers is None or headers.get("request_id"):
+        return
+    ctx = WorkerLogger.get_context()
+    request_id = _coerce_id(getattr(ctx, "request_id", None)) if ctx else None
+    if request_id:
+        headers["request_id"] = request_id
```
praveen-formido, line 776: **This re-publish handler combined with the new header-first precedence regresses correlation for beat-scheduled pipelines.**

`_bind_task_context` falls back to `request_id or task_id`, so a task with no extractable payload id binds its own opaque Celery `task_id`. This handler then stamps that `task_id` onto every task it publishes, and the child's `_request_id_from_message` gives the header *top* priority over its own payload (above `file_execution_id` / `execution_id`).

Concrete scenario — scheduled ETL/TASK pipelines, the most common production flow:

1. `workers/scheduler/tasks.py:262` `execute_pipeline_task_v2(organization_id, pipeline_id, pipeline_name)` — none of `_REQUEST_ID_KEYS` (`request_id`/`file_execution_id`/`execution_id`/`run_id`) is present, so it binds its Celery `task_id`.
2. It then dispatches `async_execute_bin` (`workers/scheduler/tasks.py:172`), whose payload *does* carry the real `execution_id`.
3. Post-change that child — and every downstream file batch and callback — logs `request_id:<scheduler-celery-task-id>` instead of `request_id:<execution_id>`.

The PR's risk section says beat-scheduled tasks "fall back to the existing `execution_id`/`task_id` behaviour", but the header is now present on those chains, so they don't. Same shape applies to the per-file granularity `_extract_request_id` documents ("Workers bind `file_execution_id` … giving per-file granularity"): once any ancestor stamps a header, descendants can no longer surface their own id.

Suggest only propagating an id that is genuinely a cross-service correlation id — e.g. have `_bind_task_context` record whether the bound value came from a message header (or from a real `request_id` payload key) and have this handler forward it only in that case, leaving payload-derived / `task_id` fallbacks local to their own task.

workers/shared/infrastructure/logging/logger.py
```diff
@@ -703,24 +703,69 @@ def _extract_request_id(
...
+                "request_id extraction failed for task %s; falling back to task_id",
+                task_id,
+                exc_info=True,
+            )
+            request_id = None
+    request_id = request_id or task_id
     WorkerLogger.update_context(request_id=request_id, task_id=task_id)
 
 
+def _propagate_request_id_on_publish(headers=None, **_):
```
praveen-formido, line 760: **The PG-queue transport bypasses this hop entirely, so correlation breaks there.**

`workers/queue_backend/dispatch.py::enqueue` routes to `_enqueue_pg` when `resolve_backend(...) is QueueBackend.PG`, which serialises via `to_payload()` and writes straight to `pg_queue_message` — it never calls `current_app.send_task`, so `before_task_publish` never fires and no `request_id` is carried. On the consume side `consumer.py:575` runs `task.apply(headers={FAIRNESS_HEADER_NAME: ...})`, i.e. the only headers reaching `Context` are the fairness ones, so `_request_id_from_message` returns `None` and the task falls back to its payload id.

Concrete scenario: an API-triggered execution whose task name is in `WORKER_PG_QUEUE_ENABLED_TASKS` (or that rides a `backend=` pipeline override) loses the HTTP `request_id` at the first PG hop and never regains it for the rest of the execution — exactly the chains the PG rollout is moving onto. Worth threading `request_id` into `to_payload()` / restoring it into the `headers` passed to `task.apply()`, or at minimum calling this gap out in the PR's out-of-scope list.

backend/backend/celery_signals.py
```diff
@@ -0,0 +1,43 @@
...
+        return
+    try:
+        request_id = StateStore.get(Common.REQUEST_ID)
+    except Exception:
+        # StateStore can raise if CONCURRENCY_MODE is misconfigured; never let
+        # correlation plumbing break task publishing.
+        logger.debug("Unable to read request_id from StateStore", exc_info=True)
+        return
+    if request_id and not headers.get(Common.REQUEST_ID):
+        headers[Common.REQUEST_ID] = request_id
```
praveen-formido, line 36: **The header is injected for the Django Celery workers too, but nothing there consumes it — and those workers don't re-propagate.**

The `workers/` package gets its receiving side from `_bind_task_context` / `_propagate_request_id_on_publish`, but the backend Celery app has no `task_prerun` equivalent: its logging uses `log_request_id.filters.RequestIDFilter`, which reads `log_request_id.local.request_id` — only ever populated by the HTTP middleware. So for `-A backend worker` services (e.g. `worker-metrics` on `dashboard_metric_events` in `docker/docker-compose.yaml:49`, plus the `celery` / `celery_api_deployments` / `celery_periodic_logs` queues) the injected header is carried and then dropped: those task logs still show `request_id:-`.

Second-order effect: since `StateStore` is only written by `CustomAuthMiddleware` on an HTTP request, this handler is a no-op inside those worker processes, so any task they publish downstream loses the id too — the chain breaks at the first backend-worker hop.

Also a doc nit on the docstring above: `StateStore[request_id]` is populated by `CustomAuthMiddleware` (`backend/account_v2/custom_auth_middleware.py:26`), not by `CustomRequestIDMiddleware` (which only sets `request.id`). Worth correcting since the distinction matters for which requests actually have an id in scope.

praveen-formido (commented):
Code review — 4 findings on the `request_id` propagation chain. The producer/consumer plumbing itself looks correct (verified against Celery 5.5.3 that a custom key in the v2 message headers is promoted onto `Task.request` both as an attribute and via `request.headers`, and that mutating the `headers` dict inside `before_task_publish` reaches `producer.publish`). The issues are at the two ends of the chain and in the precedence rule.

chandrasekharan-zipstack (commented):
A few points on hops and guarantees that I think are still open — details inline.

workers/shared/clients/base_client.py
chandrasekharan-zipstack, line 337: **The worker → runner hop stamps a different id, and it isn't in the out-of-scope list.**

This closes the worker → backend hop, but the worker → runner hop already sends an `X-Request-ID` and it carries the wrong value:

```python
# unstract/tool-sandbox/src/unstract/tool_sandbox/helper.py:439 (and :592)
headers = {
    "X-Request-ID": file_execution_id,
}
```

The runner reads that header via `unstract.core.flask.register_request_id_middleware` and renders it as `request_id:<file_execution_id>` — so after this PR the runner is the one service actively logging a *different* id under the same field name, which is exactly what breaks the single-`request_id` gcloud query the PR is built for.

This is called from worker context, where `WorkerLogger.get_context().request_id` is already bound by `_bind_task_context`, so the fix is the same shape as `_current_request_id()` here — prefer the propagated id and keep `file_execution_id` as the fallback.

Either close it or add it alongside the SDK1 x2text-adapter gap in the PR's out-of-scope list; right now it reads as covered.

x2text-service/app/config.py
```diff
@@ -1,17 +1,25 @@
...
 from app.controllers import api
+from app.logging_util import register_request_id_middleware, setup_logging
 from app.models import X2TextAudit, be_db
 
 load_dotenv()
 
 
 def create_app() -> Flask:
+    log_level = getattr(logging, env.get("LOG_LEVEL", "INFO").upper(), logging.INFO)
+    setup_logging(log_level)
```
chandrasekharan-zipstack, line 16: **The old logging config it replaces is still in the tree.**

```python
# x2text-service/app/controllers/controller.py:19
logging.basicConfig(
    level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s"
)
```

That runs at import of `app.controllers` (line 7 above), i.e. before `create_app()` is ever called. It happens to be harmless today only because `dictConfig` then clears and replaces the root handlers — but any path that imports the controller without going through `create_app()` (tests, an alternate entrypoint, a REPL) silently gets the pre-PR format with no `request_id` field at all.

Worth deleting as part of this change — it's the line `setup_logging()` is meant to supersede, and leaving both means the service has two competing logging configs whose behaviour depends on import order.

x2text-service/app/logging_util.py
```diff
@@ -0,0 +1,96 @@
...
+        }
+    )
+
+
+def register_request_id_middleware(app: Flask) -> None:
+    """Read ``X-Request-ID`` from each request (or mint one) onto Flask ``g``."""
+
+    @app.before_request
+    def _assign_request_id() -> None:
+        g.request_id = request.headers.get("X-Request-ID", str(uuid.uuid4()))
```
chandrasekharan-zipstack: **x2text mints a request_id that nothing can ever learn.**

There's no `after_request` echoing `X-Request-ID` back on the response — the Django backend does this via `REQUEST_ID_RESPONSE_HEADER`, and the frontend relies on it. Combined with the SDK1 x2text-adapter not forwarding the header (acknowledged as out of scope), the practical outcome today is: x2text generates a fresh uuid4 per request, receives it from nobody, and returns it to nobody. The id appears only in x2text's own log lines, where it correlates with nothing.

A three-line `after_request` at least makes it retrievable from a live call now, ahead of the SDK-side follow-up:

```python
    @app.after_request
    def _echo_request_id(response):
        response.headers["X-Request-ID"] = g.get("request_id", "-")
        return response
```

Same gap exists in the shared `unstract/core/src/unstract/core/flask/middleware.py:15`, so fixing it in both keeps the two copies aligned.

workers/shared/infrastructure/logging/logger.py
```diff
@@ -703,24 +703,69 @@ def _extract_request_id(
...
-    try:
-        request_id = _extract_request_id(args or (), kwargs or {}, task) or task_id
-    except Exception:
-        logging.getLogger(__name__).debug(
-            "request_id extraction failed for task %s; falling back to task_id",
-            task_id,
-            exc_info=True,
-        )
-        request_id = task_id
+    request_id = _request_id_from_message(task)
```
chandrasekharan-zipstack: **This call sits outside the try, which breaks the guarantee the docstring three lines above still promises.**

> Catches any extraction failure so a malformed payload can never leave the previous task's id bound on the thread.

Before this PR the whole resolution — including the `or task_id` fallback — was inside the `try`, so `update_context` always ran. Now `_request_id_from_message(task)` runs unguarded: if it raises, `WorkerLogger.update_context(...)` on line 749 is never reached, and the **previous** task's `request_id` *and* `task_id` stay bound on the thread-local context. That is precisely the failure mode the guard was written for, and it's worse than the old one because a stale id is silently plausible where a missing one is obvious.

A raising `task_prerun` receiver also fails the task itself (`send_prerun` is inside the traced task body in `celery/app/trace.py`), so the blast radius isn't only logging.

Moving the call inside the existing `try` keeps the invariant with no other change:

```python
    try:
        request_id = _request_id_from_message(task) or _extract_request_id(
            args or (), kwargs or {}, task
        )
    except Exception:
        ...
        request_id = None
```

workers/shared/clients/base_client.py
Deepak-Kesavan: Fixed in 713f9a474: `LOG_REQUEST_ID_HEADER` is now `HTTP_X_REQUEST_ID`. Spot on — and this also explains a red herring in our own dev smoke test, where the client `X-Request-ID` we sent was dropped and a fresh id minted (we had wrongly blamed the edge LB). The incoming header from the frontend and from workers is now actually honored. Thanks.

workers/shared/infrastructure/logging/logger.py
Deepak-Kesavan: Fixed in 713f9a474. Added a `request_id_propagatable` marker to `LogContext`: `_bind_task_context` sets it True only when the id came from an upstream message header, and `_propagate_request_id_on_publish` now re-propagates only those — never a `task_id` or `execution_id` fallback. So beat/scheduler-originated child tasks keep deriving their own `file_execution_id`; only a genuine HTTP-origin id fans out across the chain.

workers/shared/infrastructure/logging/logger.py
Deepak-Kesavan: Confirmed — the PG transport enqueues via `PgQueueMessage.objects.create()`, so `before_task_publish` never fires and `TaskPayload` carries no request_id. Keeping this out of scope for this PR (Celery is the live default; `pg_queue_enabled` is fail-closed). The PGMQ owner has a precise carrier spec (add `request_id` to `TaskPayload`, populate in the backend and worker PG producers, inject into the consumer `apply(headers=...)`). Added to the PR out-of-scope section.

backend/backend/celery_signals.py
Deepak-Kesavan: Fixed in 713f9a474. Added `task_prerun`/`task_postrun` handlers in `celery_signals.py` that bind the injected header onto `log_request_id.local` (so the backend `RequestIDFilter` emits it) and onto `StateStore` (so a task the backend worker itself publishes re-propagates). No longer a no-op on that side.

workers/shared/clients/base_client.py
Deepak-Kesavan: Good catch. `tool-sandbox -> runner` sends `X-Request-ID = file_execution_id`, which is inconsistent with standardizing on the HTTP request_id. Reconciling it means threading the request_id through tool-sandbox, which belongs with the tool-execution / OTel follow-up (same bucket as the SDK1 x2text-adapter forwarding). Added it explicitly to the PR out-of-scope list rather than leaving it silent.

x2text-service/app/config.py
Deepak-Kesavan: Fixed in 713f9a474 — removed the stale `logging.basicConfig` at `controller.py` import. `setup_logging` (dictConfig) is the single source now.

x2text-service/app/logging_util.py
Deepak-Kesavan: Fixed in 713f9a474 — added an `after_request` that echoes `X-Request-ID`, and put it in the shared `unstract.core.flask` middleware too, so platform-service and runner echo consistently, not just x2text.

workers/shared/infrastructure/logging/logger.py
Deepak-Kesavan: Fixed in 713f9a474 — moved the whole resolution (header read, payload extraction, and the `or task_id` fallback) inside the `try`, so `update_context` always runs and the docstring guarantee holds again.

workers/shared/clients/base_client.py
Deepak-Kesavan: Follow-up on the second half of this comment — the test you suggested is now in `9bd918a3d`.

`backend/middleware/test_request_id.py` posts an `X-Request-ID` through a real middleware chain and asserts the backend adopts it as `request.id` *and* echoes the same value back on the response. Settings are deliberately not overridden, so it reads the production `LOG_REQUEST_ID_HEADER`. I verified it has teeth by reverting the setting to the old raw header name — 3 tests fail, reproducing exactly the mint-a-fresh-uuid4 behaviour you diagnosed:

```
AssertionError: 8251fb90-492f-4cc0-a652-d3ee5fca1127 != aaaaaaaa-bbbb-4ccc-8ddd-eeeeeeeeeeee
```

Also added `backend/backend/test_celery_signals.py` covering the carrier itself: header injection from `StateStore`, the no-clobber and no-op-without-an-id paths, binding from both the `Context` attribute and the raw `headers` mapping, re-propagation from a backend worker, and `task_postrun` clearing so a pooled thread cannot leak the previous task's id. 17 tests, DB-free so they run in the unit tier.

chandrasekharan-zipstack (approved):
Consider making code comments and generic to avoid rot over time

PR #2207, opened 2026-07-24
feat(mcp): host deployment-scoped and organization-scoped MCP servers

hari-kuriakose:
## What

Hosts **two** MCP servers for Unstract, so coding agents can discover and run
document extraction as tool calls instead of hand-rolling HTTP requests.

| | Deployment server | Platform server |
| --- | --- | --- |
| Scope | one API deployment | one organization |
| Credential | that deployment's API key | a `PlatformApiKey` |
| Authenticated by | the view itself | `CustomAuthMiddleware` |
| URL | `/deployment/api/<org>/<api_name>/mcp` | `/api/v1/unstract/<org>/mcp/` |
| Tools | extract, poll status | discovery + state changes |

They share the JSON-RPC transport (`transport.py`) and differ only in
authentication and tool surface. Neither key works on the other server.

## Approach

A hand-rolled JSON-RPC 2.0 endpoint mounted in the existing Django app rather
than a separate service to deploy and scale, with a declarative tool registry
behind it:

- Scoped via the URL, so an MCP session inherits the deployment or organization
  the caller already addresses.
- Authenticated with the credential that already exists for that scope — no
  second key to mint or revoke.
- `MCPToolRegistry` maps a tool name to a callable plus the JSON schema clients
  need, so `tools/list` is generated rather than maintained by hand.

Only the **hosted HTTP** transport is implemented; a stdio server is a separate
concern and out of scope here.

The app is `backend/mcp_server/` — deliberately **not** `_v2` (that suffix marks
apps carried through the v0.93.0 v1→v2 multitenancy migration; every app added
since is unsuffixed), and not bare `mcp`, which would shadow the PyPI `mcp` SDK
package on the import path.

## Endpoint

An MCP session is scoped to one API deployment and mirrors that deployment's
REST URL:

```
POST /deployment/api/<org_name>/<api_name>/     # REST (existing)
POST /deployment/api/<org_name>/<api_name>/mcp  # MCP  (new)
```

Auth is the deployment's **existing API key** — same key, same management UI.
There is no second credential to mint or revoke.

```bash
claude mcp add --transport http unstract \
  https://<host>/deployment/api/<org_name>/<api_name>/mcp \
  --header "Authorization: Bearer <api_key>"
```

A `/<api_key>` path variant exists for MCP clients that cannot set headers.
It is a workaround for client limitations, not a recommended path — credentials
in a URL can reach access logs and proxies, so prefer the header form.

## Tools

| Tool | Purpose |
| --- | --- |
[...]

PR #2231, opened 2026-08-07
fix(mcp): skip the platform auth tests where their preconditions do not hold

Deepak-Kesavan:
Unblocks Unstract Cloud `main`, red since #2207 merged (cloud run 31159495978 and every run since).

## What is broken

`mcp_server/tests/test_platform_auth.py` drives requests through the real URL stack so `CustomAuthMiddleware` runs. On cloud neither thing it depends on is true:

- **The route is not mounted.** `copy_cloud_deps.py` overwrites the OSS `settings/test.py` with a redirect to `test_cloud.py`, which derives from `settings/cloud.py` → `base.py`. So `MCP_PLATFORM_SERVER_ENABLED = True` never reaches the cloud suite, and `urls_v2` skips the mount. Five tests 404.
- **The auth middleware is absent.** The same `test_cloud.py` drops `CUSTOM_AUTH_MIDDLEWARE` from the default test `MIDDLEWARE` (it is kept only in `MIDDLEWARE_WITH_AUTH`). This view has `permission_classes = []` and deliberately does not re-authenticate.

Why it passed pre-merge: cloud `ci-test.yaml` checks out OSS at a floating `ref: main`. #1712 was green at 07:42 against an OSS main where `mcp_server/` did not exist; #2207 landed at 07:54:05 and the next cloud run was red. Nothing re-validates a merged cloud PR against a newer OSS main — worth addressing separately.

## The sixth test is why this is a skip, not a deselect

`test_bad_credentials_are_rejected` **passed** on cloud. A 404 satisfies "this request is refused" exactly as well as the 401 it was written to assert, so it reported green for auth that was never reached. `--ignore`-ing the file keeps that hidden; a skip names it.

## Why not just enable the flag on cloud

It mounts the route but does not go green — with the middleware still absent the endpoint goes from *unmounted* to *unauthenticated*, trading five failures for a different five. I verified this on a merged tree before discarding the approach.

## Why not override the group in `groups.cloud.yaml`

`pytest_extra: ["--ignore=..."]` on `integration-backend` cannot work. The rig rejects redefinition:

```
ERROR: tests/groups.cloud.yaml: group 'integration-backend' already defined in a prior manifest
```

`_merge_manifest` raises on name collision by design ("a name collision is an error rather than a silent override"), and it fails at the `Validate rig manifests` step before any test runs. There is no ignore/exclude/disable key in the rig either. So the fix has to live here.

## What this does

Checks the two preconditions directly and skips with a named reason otherwise — deliberately **not** by reading `MCP_PLATFORM_SERVER_ENABLED`. That flag is what *causes* the mount in the OSS URLconf, but it is not what these tests need; asking for the route and the middleware stays correct for any tree that mounts the server another way, and re-arms on its own if cloud ever runs this endpoint with auth.

The route probe resolves a path with **no organization segment**, unlike the URL the tests request: tenant middleware strips the org from `path_info` before resolution, so resolving the request URL as written 404s even on OSS where these tests pass.

No.

- Test-only change; no production code is touched.
- On OSS both preconditions hold, so the suite runs exactly as before — verified, no unmet precondition reported.
- The deployment-scoped MCP server, the one required for release, is mounted unconditionally in `api_v2/execution_urls.py` and is unaffected by any of this.
- Nothing enables the organization-scoped server anywhere; it stays off in staging and production, as intended.

## Notes on Testing

Verified against real `copy_cloud_deps`-merged trees rather than by reasoning:

| tree | verdict |
|---|---|
| OSS `main` | no unmet precondition → suite runs unchanged |
| cloud merged, flag off (today) | skip, reporting the **route** reason |
| cloud merged, flag on | skip, reporting the **middleware** reason |

Both guard branches therefore exercised. `198 passed` in the OSS unit tier.

## Related Issues or PRs

- Zipstack/unstract#2207 — introduced these tests
- Zipstack/unstract-cloud#1712 — deployment-scoped allowlist; green, unrelated to this failure
- Unblocks Zipstack/unstract-cloud#1705

### DIFF
--- a/backend/mcp_server/tests/test_platform_auth.py
+++ b/backend/mcp_server/tests/test_platform_auth.py
@@ -18,6 +18,7 @@ import uuid
 
 import pytest
 from account_v2.models import Organization, User
+from django.conf import settings
 from django.test import Client, TestCase
 from platform_api.models import PlatformApiKey
 from tenant_account_v2.models import OrganizationMember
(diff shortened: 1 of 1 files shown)


Deepak-Kesavan:
You're right, and it's the more important half of the objection — taking it.

`1acd357a` inverts the gate. The flag now decides whether the server is *expected*, and the route and middleware become assertions rather than skip conditions:

```
MCP_PLATFORM_SERVER_ENABLED off -> skip; deliberately absent, nothing to authenticate
MCP_PLATFORM_SERVER_ENABLED on  -> route must resolve AND CUSTOM_AUTH_MIDDLEWARE
                                   must be in MIDDLEWARE, or fail naming which
```

My reasoning for the 404 check was that the flag causes the mount but isn't what the tests need. That held for the case in front of me and missed yours: where the server is meant to be on, an unmounted route is a regression, and my version would have gone quiet exactly when it should have shouted — the same failure this suite exists to catch, one level up.

I kept the middleware assertion, because the flag alone doesn't make these tests meaningful. Cloud's `test_cloud.py` drops `CUSTOM_AUTH_MIDDLEWARE` from the default test `MIDDLEWARE`, and this view has `permission_classes = []` and doesn't re-authenticate — so enabling the flag there *without* the middleware moves the endpoint from unmounted to unauthenticated, and the rejection tests would pass against an open endpoint. With the assertion it fails and says so. I verified that specific case rather than assuming it.

Four states, all exercised:

| state | outcome |
|---|---|
| enabled, wired up (OSS) | runs |
| disabled (cloud today) | skips |
| enabled, middleware dropped | **fails**, naming the middleware |
| enabled, URLconf without the mount | **fails**, naming the path |

`198 passed` in the unit tier.

---

On the broader class — agreed that's the real problem, and I'd split it in two:

**Settings divergence.** `copy_cloud_deps` overwrites OSS `settings/test.py` with a redirect to `test_cloud.py`, which derives from `settings/cloud.py`. So every key OSS sets for tests is silently dropped downstream. I diffed the two: `MCP_PLATFORM_SERVER_ENABLED` is currently the *only* divergence, so we caught this at one. A test asserting the two files agree on a declared set of keys would turn the next one into a named failure instead of a mystery 404 in an unrelated suite.

**Floating OSS ref.** Cloud CI resolves `Zipstack/unstract@main` at run time, so a cloud PR's green tick is only valid against whatever OSS main was at that instant — which is exactly how #1712 merged clean and cloud went red 9 seconds later. Nothing re-validates a merged cloud PR against a newer OSS main. Worth its own issue; options range from pinning the ref to a scheduled cloud-main run against latest OSS.

Happy to take either if useful — the settings-drift guard is small and would have caught this one before it merged.

harini-venkataraman (approved):

Issue #1684, opened 2025-11-28
fix: [ISSUE] "error": "Final output processing failed: " when using API deployment


freddujour:
## Describe the bug
When deploying a Prompt Studio project as an API Deployment, the workflow executes successfully through all stages (SOURCE → INITIALIZE → COMPILE → BUILD → RUN), and the prompt output is correctly generated and visible under “Combined Output” in Prompt Studio. However, the API Deployment consistently fails at the final step with the error:

```
Final output processing failed:
```

The resulting payload shows `execution_status: "ERROR"` and the `result` array contains:
```
{
  "file": "document_xxx",
  "status": "Failed",
  "result": null,
  "error": "Final output processing failed:",
  "metadata": null
}
```

The prompt output itself is valid JSON/text and is correctly written to workflow storage. The failure appears to occur only inside the API Deployment wrapper.

## To reproduce
1. Install Unstract OSS using the official Docker Compose stack (default configuration).
2. Create a Prompt Studio project that extracts structured fields from a PDF (e.g. `"provider": "BRENNTAG"`).
3. Test the prompt manually → valid output is produced.
4. Click **Deploy as API** and create a new deployment.
5. Send a PDF to:
   ```
   POST /deployment/api/mock_org/<deployment>/
   ```
   with a valid API key.
6. Poll the returned `status_api` endpoint once.
7. The workflow completes RUN, writes the prompt output to storage, then FINALIZE fails with:
   ```
   Final output processing failed:
   ```
8. API returns:
   ```
   {
     "execution_status": "ERROR",
     "result": [
       {
         "status": "Failed",
         "result": null,
         "error": "Final output processing failed:",
         "metadata": null
       }
     ]
   }
   ```

## Expected behavior
The API Deployment should return the prompt output generated by the workflow. Since the LLM execution succeeds and the prompt output is correctly stored, the API Deployment should assemble and return the final structured result instead of failing.

## Environment details
- Unstract Open Source Edition  
- Version: latest OSS (Docker images pulled November 2025)  
- Host: Windows 11 (Docker Desktop)  
- Containers: `unstract-backend`, `unstract-frontend`, `unstract-worker`, `unstract-worker-file-processing`, `unstract-platform-service`, etc.  
- API used: `/deployment/api/mock_org/<deployment>/`

## Additional context
- Issue reproduces whether prompt output is `json` or `text`.  
- LLM profile (`qwen2.5-3b-instruct`) runs without errors.  
[...]

PR #2232, opened 2026-08-07
UN-3638 [MISC] Scope and gate critical-path regressions in the rig report

chandrasekharan-zipstack:
## What

- Pass `scope_groups` to `evaluate()` in `cmd_report`, derived from the groups that actually emitted junit.
- Gate the report job on surviving regressions instead of returning 0 regardless.
- Split the report's gap section so a path covered by a tier that did not run is no longer listed under "not yet covered".

## Why

Spotted on #2230, a frontend-only PR. Its report comment listed nine critical paths under **❌ Regressions (must be zero)** — `adapter-register-llm`, `workflow-author`, `api-deployment-provision`, `api-deployment-auth`, `mcp-server-auth`, `mcp-platform-auth`, `prompt-studio-author`, `connector-register-test`, `usage-aggregate-read` — while every check on the PR was green.

Two independent defects, which masked each other.

**1. False positives.** All nine paths are covered solely by `integration-backend`. The `changes` filter marks `frontend/**` as not `relevant`, so the `test` job (unit + integration) skips and emits no junit. `cmd_report` called `evaluate()` without `scope_groups`, which the docstring documents as "no scoping is applied (back-compat)", making `in_scope` unconditionally true. Every one of those paths was in main's baseline and not covered in this build, so:

```python
elif path.id in previously_covered and in_scope:
    state = "regression"
```

`cmd_run` already passes `scope_groups` and degrades out-of-tier paths to `gap`. `cmd_report` never got the same treatment, even though the workflow comment names it "the sole regression authority".

**2. No gate.** `cmd_report`'s only exit path was `return 1 if unknown_marker_ids else 0` — it rendered "must be zero" into the comment and exited 0 regardless. The `report` job's fail step only inspects `needs.*.result`, and a path-filtered skip counts as a pass, so nothing failed.

Fixing either alone is wrong: gate-only turns every frontend PR red, scope-only leaves genuine regressions unenforced.

The practical cost today is a comment that cries wolf under a heading reading "must be zero", which trains reviewers to scroll past the section that is supposed to stop a merge.

## How

- `scope_groups = [r.name for r in group_results]` — a skipped tier contributes no groups, so its paths fall to `gap`. A group that ran and went red still reports, stays in scope, and is still classified as a regression.
- Collect `regressions` from the evaluated statuses, print the ids to stderr, and fold them into the return code.
- `reporting.py` splits out-of-scope gaps that have declared covering groups into a collapsed "💤 Covered, but not exercised in this build" section. Paths with no declared groups (e.g. `workflow-execution-fan-out`) stay under "not yet covered", where they belong.

It changes when a required check fails, so it is worth reviewing on that basis rather than as a no-op.

- **On a full run the scoping is a no-op.** Every group reports, so `scope_groups` covers all of them and `in_scope` matches the previous unconditional `True`. Only builds with a skipped tier change classification. `main` always runs both tiers.
- **The gate is new behaviour by design.** Any genuine regression now fails `report`, where it previously only printed. If main's baseline is currently out of step with reality, the first build after this merges will surface it — that is the intended effect, but it is the reason to land this deliberately rather than fold it into an unrelated PR.
- **One edge case worth naming:** a non-optional group that runs green but collects zero tests does not attest coverage (`_coverage_attesting_groups` excludes `empty`), yet it does report, so it stays in scope. Its baseline-covered paths will now be regressions and will fail the job. That is the correct reading of a group that silently stopped collecting, and `cmd_run` already fails empty non-optional groups, but it is a new failure mode for the report job.
- **A corrupt baseline now fails the report job too.** Raised in review: a truncated `previous-summary.json` empties `previously_covered`, which makes the `regression` state unreachable and would have let the new gate pass vacuously. `cmd_report` now flips its exit code on `baseline_corrupt`, matching `cmd_run`. A *missing* baseline still exits 0, so first runs and post-eviction runs are unaffected — only a file that exists and is unreadable fails.
- **A skipped marked test now fails the report job, with a green tier.** Only *passing* `@pytest.mark.critical_path` tests attest, so an env-dependent `skipTest` leaves its group green and in scope while attesting nothing — which reads as a regression. Raised in review; kept as a failure (a silently-skipped critical-path test should shout), but the stderr line now distinguishes "covering group ran green but nothing attested" from "no covering group ran green", since the fixes differ.
- **`optional` groups are excluded from the gate's scope.** They are documented as non-blocking and `cmd_run` honours that; leaving them in scope let a red optional group gate the build through a regression instead. Latent today (no critical path names one), fixed before it isn't.
- **The baseline is now pruned to the registry on merge.** The union never forgot, so a deliberately retired path could not leave the cache. Editing `critical_paths.yaml` is what accepts a removal, and the failure message says so.
- Baseline writing is untouched: `--update-baseline` remains main-only and green-only, so no PR can poison it.

## Database Migrations

- None.

## Env Config

- None.

## Related Issues or PRs

- Found while reviewing CI output on #2230. No code dependency in either direction; that PR is frontend-only.

## Notes on Testing

- Two new tests in `tests/rig/tests/test_cli.py` drive `cmd_report` end to end: `test_cmd_report_skipped_tier_is_a_gap_not_a_regression` (covering group emits no junit → `gap`, exit 0) and `test_cmd_report_gates_on_a_real_regression` (covering group reports red → `regression`, exit 1).
- Both were confirmed to fail against the unmodified `cli.py` and pass with it, so they exercise the changed branch rather than passing vacuously.
- Full rig self-test suite: 106 passed, 1 skipped.
- Note that this PR cannot demonstrate the fix in its own CI run: touching `tests/rig/**` flips `relevant` to true, so the integration tier runs and those paths are attested normally. The live confirmation is #2230's next run once this lands — its nine paths should move out of the regression section.

## Checklist

### DIFF
--- a/backend/backend/settings/test.py
+++ b/backend/backend/settings/test.py
@@ -1,16 +1,8 @@
+"""OSS test settings: the production base plus the shared test-only deltas.
+
+The deltas live in `test_base` rather than here because `copy_cloud_deps`
+replaces this file on a cloud build.
+"""
+
 from backend.settings.base import *  # noqa: F401, F403
-
-DEBUG = True
-
-# Django's default PBKDF2 hasher is deliberately slow; suites that seed several
-# users per test spend most of their time there. Test fixtures need speed, not
-# resistance to offline cracking.
-PASSWORD_HASHERS = ["django.contrib.auth.hashers.MD5PasswordHasher"]
-
-# The organization-scoped MCP server ships disabled (see
-# MCP_PLATFORM_SERVER_ENABLED in base.py), but its tests must still exercise
-# it: several drive requests through the full URL stack precisely so the auth
-# middleware runs, and an unmounted route would make them 404 — turning a suite
-# that checks a credential is *rejected* into one that passes because nothing
-# is there. Shipping-off and untested are different things.
-MCP_PLATFORM_SERVER_ENABLED = True
+from backend.settings.test_base import *  # noqa: F401, F403
--- a/backend/backend/settings/test_base.py
+++ b/backend/backend/settings/test_base.py
@@ -14,10 +14,6 @@ PASSWORD_HASHERS = ["django.contrib.auth.hashers.MD5PasswordHasher"]
 SESSION_COOKIE_SECURE = False
 CSRF_COOKIE_SECURE = False
 
-# Env-driven, so unset under test: without it internal-API requests fail as
-# "not configured" rather than on their own merits.
-INTERNAL_SERVICE_API_KEY = "test-internal-service-key"
-
 # Ships disabled, but its tests drive the full URL stack to exercise the auth
 # middleware — an unmounted route would 404 and pass them vacuously.
 MCP_PLATFORM_SERVER_ENABLED = True
--- a/backend/backend/settings/test_base.py
+++ b/backend/backend/settings/test_base.py
@@ -1,0 +1,32 @@
+"""Test-only setting deltas, shared by the OSS and cloud test settings.
+
+This module imports nothing on purpose. `copy_cloud_deps` replaces the OSS
+`settings/test.py` on a cloud build, so a delta that lives only in that file is
+silently lost there; keeping the deltas here is what stops the two trees
+diverging. Star-importing a base settings module would re-export its names and
+clobber whatever the importer derived from `cloud`, so this file defines only
+what it overrides.
+"""
+
+DEBUG = True
+
+# Django's default PBKDF2 hasher is deliberately slow; suites that seed several
+# users per test spend most of their time there. Test fixtures need speed, not
+# resistance to offline cracking.
+PASSWORD_HASHERS = ["django.contrib.auth.hashers.MD5PasswordHasher"]
+
+# Prod enforces HTTPS-only cookies; tests run over plain HTTP.
+SESSION_COOKIE_SECURE = False
+CSRF_COOKIE_SECURE = False
+
+# Env-driven and unset under test, which would make any request to an internal
+# API fail as "not configured" rather than on its own merits.
+INTERNAL_SERVICE_API_KEY = "test-internal-service-key"
+
+# The organization-scoped MCP server ships disabled (see
+# MCP_PLATFORM_SERVER_ENABLED in base.py), but its tests must still exercise
+# it: several drive requests through the full URL stack precisely so the auth
+# middleware runs, and an unmounted route would make them 404 — turning a suite
+# that checks a credential is *rejected* into one that passes because nothing
+# is there. Shipping-off and untested are different things.
+MCP_PLATFORM_SERVER_ENABLED = True
--- a/backend/backend/settings/test_base.py
+++ b/backend/backend/settings/test_base.py
@@ -1,32 +1,23 @@
 """Test-only setting deltas, shared by the OSS and cloud test settings.
 
-This module imports nothing on purpose. `copy_cloud_deps` replaces the OSS
-`settings/test.py` on a cloud build, so a delta that lives only in that file is
-silently lost there; keeping the deltas here is what stops the two trees
-diverging. Star-importing a base settings module would re-export its names and
-clobber whatever the importer derived from `cloud`, so this file defines only
-what it overrides.
+Imports nothing on purpose: `copy_cloud_deps` replaces the OSS `settings/test.py`
+on a cloud build, so deltas kept only there are lost, and star-importing a base
+module here would re-export its names over whatever the importer derived.
 """
 
 DEBUG = True
 
-# Django's default PBKDF2 hasher is deliberately slow; suites that seed several
-# users per test spend most of their time there. Test fixtures need speed, not
-# resistance to offline cracking.
+# PBKDF2 is deliberately slow, and suites seeding users per test pay it repeatedly.
 PASSWORD_HASHERS = ["django.contrib.auth.hashers.MD5PasswordHasher"]
 
 # Prod enforces HTTPS-only cookies; tests run over plain HTTP.
 SESSION_COOKIE_SECURE = False
 CSRF_COOKIE_SECURE = False
 
-# Env-driven and unset under test, which would make any request to an internal
-# API fail as "not configured" rather than on its own merits.
+# Env-driven, so unset under test: without it internal-API requests fail as
+# "not configured" rather than on their own merits.
 INTERNAL_SERVICE_API_KEY = "test-internal-service-key"
 
-# The organization-scoped MCP server ships disabled (see
-# MCP_PLATFORM_SERVER_ENABLED in base.py), but its tests must still exercise
-# it: several drive requests through the full URL stack precisely so the auth
-# middleware runs, and an unmounted route would make them 404 — turning a suite
-# that checks a credential is *rejected* into one that passes because nothing
-# is there. Shipping-off and untested are different things.
+# Ships disabled, but its tests drive the full URL stack to exercise the auth
+# middleware — an unmounted route would 404 and pass them vacuously.
 MCP_PLATFORM_SERVER_ENABLED = True
--- a/tests/rig/cli.py
+++ b/tests/rig/cli.py
@@ -374,7 +374,10 @@ def cmd_report(args: argparse.Namespace) -> int:
     # A skipped tier emits no junit, so scoping to the groups that reported keeps
     # its paths as gaps rather than regressions — nothing regressed, it never ran.
     # A group that ran and went red still reports, so real regressions survive.
-    scope_groups = [r.name for r in group_results]
+    # `optional` groups are excluded: they are documented as non-blocking, and
+    # leaving them in scope would let a red one gate the build through a
+    # regression instead.
+    scope_groups = [r.name for r in group_results if not manifest.get(r.name).optional]
     statuses = cp.evaluate(
         registry,
         groups_run_green=green,
@@ -399,11 +402,37 @@ def cmd_report(args: argparse.Namespace) -> int:
     # cmd_run gates on the regressions it can see, but only within its own tier.
     # This is the only cross-tier evaluation, so it is the only place a regression
     # spanning tiers can be gated on.
-    regressions = [s for s in statuses if s.state == "regression"]
-    if regressions:
-        ids = ", ".join(s.path.id for s in regressions)
+    # A covering group that ran green without attesting means its marked test
+    # was skipped or unmarked; a group that went red means the test failed. The
+    # remedies differ, so the two are reported apart rather than as one count.
+    unproven: list[cp.CriticalPathStatus] = []
+    uncovered: list[cp.CriticalPathStatus] = []
+    for s in statuses:
+        if s.state != "regression":
+            continue
+        target = unproven if any(g in green for g in s.path.covered_by) else uncovered
+        target.append(s)
+    regressions = unproven + uncovered
+    if uncovered:
+        ids = ", ".join(s.path.id for s in uncovered)
         print(
-            f"\n[rig] ❌ {len(regressions)} critical-path regression(s) detected: {ids}",
+            f"\n[rig] ❌ {len(uncovered)} critical-path regression(s) — no covering "
+            f"group ran green: {ids}",
+            file=sys.stderr,
+        )
+    if unproven:
+        ids = ", ".join(s.path.id for s in unproven)
+        print(
+            f"\n[rig] ❌ {len(unproven)} critical-path regression(s) — covering group "
+            f"ran green but no passing @pytest.mark.critical_path test attested "
+            f"them (skipped or unmarked?): {ids}",
+            file=sys.stderr,
+        )
+    if regressions:
+        print(
+            "[rig] to accept a deliberate removal, drop or re-point the path in "
+            "tests/critical_paths.yaml in the same PR — the baseline is keyed off "
+            "that registry.",
             file=sys.stderr,
         )
     if args.update_baseline:
--- a/tests/rig/cli.py
+++ b/tests/rig/cli.py
@@ -398,9 +398,9 @@ def cmd_report(args: argparse.Namespace) -> int:
             f"(not in tests/critical_paths.yaml)",
             file=sys.stderr,
         )
-    # This is the only cross-tier evaluation, so it is also the only place a
-    # regression can be gated on. cmd_run deliberately doesn't: a single tier
-    # can't tell "another tier's path went red" from "another tier didn't run".
+    # cmd_run gates on the regressions it can see, but only within its own tier.
+    # This is the only cross-tier evaluation, so it is the only place a regression
+    # spanning tiers can be gated on.
     regressions = [s for s in statuses if s.state == "regression"]
     if regressions:
         ids = ", ".join(s.path.id for s in regressions)
@@ -424,7 +424,10 @@ def cmd_report(args: argparse.Namespace) -> int:
             return 1
         cp.merge_into_baseline(statuses, baseline_path)
         print(f"[rig] merged into baseline: {baseline_path}")
-    return 1 if unknown_marker_ids or regressions else 0
+    # A corrupt baseline empties previously_covered, so no path can reach
+    # "regression" and the gate above silently passes. Fail on it directly
+    # rather than leaving a required check green behind an advisory banner.
+    return 1 if unknown_marker_ids or regressions or baseline_corrupt else 0
 
 
 def cmd_run(args: argparse.Namespace) -> int:
--- a/tests/rig/cli.py
+++ b/tests/rig/cli.py
@@ -371,8 +371,18 @@ def cmd_report(args: argparse.Namespace) -> int:
         print(f"[rig] {exc}", file=sys.stderr)
         baseline = None
         baseline_corrupt = True
+    # A tier the path filter skipped emits no junit, so its groups are absent
+    # here. Scoping to the groups that actually reported keeps their paths as
+    # gaps rather than regressions — nothing regressed, the tier never ran. A
+    # group that ran and went red still reports, so it stays in scope and a
+    # genuine regression is still caught.
+    scope_groups = [r.name for r in group_results]
     statuses = cp.evaluate(
-        registry, groups_run_green=green, baseline=baseline, marker_proven=proven
+        registry,
+        groups_run_green=green,
+        baseline=baseline,
+        scope_groups=scope_groups,
+        marker_proven=proven,
     )
     write_summary(
         reports_dir=reports_dir,
@@ -386,6 +396,16 @@ def cmd_report(args: argparse.Namespace) -> int:
             f"[rig] ❌ @pytest.mark.critical_path references unknown path "
             f"id(s): {', '.join(unknown_marker_ids)} "
             f"(not in tests/critical_paths.yaml)",
+            file=sys.stderr,
+        )
+    # This is the only cross-tier evaluation, so it is also the only place a
+    # regression can be gated on. cmd_run deliberately doesn't: a single tier
+    # can't tell "another tier's path went red" from "another tier didn't run".
+    regressions = [s for s in statuses if s.state == "regression"]
+    if regressions:
+        ids = ", ".join(s.path.id for s in regressions)
+        print(
+            f"\n[rig] ❌ {len(regressions)} critical-path regression(s) detected: {ids}",
             file=sys.stderr,
         )
     if args.update_baseline:
@@ -404,7 +424,7 @@ def cmd_report(args: argparse.Namespace) -> int:
             return 1
         cp.merge_into_baseline(statuses, baseline_path)
         print(f"[rig] merged into baseline: {baseline_path}")
-    return 1 if unknown_marker_ids else 0
+    return 1 if unknown_marker_ids or regressions else 0
 
 
 def cmd_run(args: argparse.Namespace) -> int:
--- a/tests/rig/cli.py
+++ b/tests/rig/cli.py
@@ -371,11 +371,9 @@ def cmd_report(args: argparse.Namespace) -> int:
         print(f"[rig] {exc}", file=sys.stderr)
         baseline = None
         baseline_corrupt = True
-    # A tier the path filter skipped emits no junit, so its groups are absent
-    # here. Scoping to the groups that actually reported keeps their paths as
-    # gaps rather than regressions — nothing regressed, the tier never ran. A
-    # group that ran and went red still reports, so it stays in scope and a
-    # genuine regression is still caught.
+    # A skipped tier emits no junit, so scoping to the groups that reported keeps
+    # its paths as gaps rather than regressions — nothing regressed, it never ran.
+    # A group that ran and went red still reports, so real regressions survive.
     scope_groups = [r.name for r in group_results]
     statuses = cp.evaluate(
         registry,
@@ -424,9 +422,8 @@ def cmd_report(args: argparse.Namespace) -> int:
             return 1
         cp.merge_into_baseline(statuses, baseline_path)
         print(f"[rig] merged into baseline: {baseline_path}")
-    # A corrupt baseline empties previously_covered, so no path can reach
-    # "regression" and the gate above silently passes. Fail on it directly
-    # rather than leaving a required check green behind an advisory banner.
+    # A corrupt baseline makes "regression" unreachable, so the gate above would
+    # pass vacuously; it has to fail on its own.
     return 1 if unknown_marker_ids or regressions or baseline_corrupt else 0
 
 
--- a/tests/rig/critical_paths.py
+++ b/tests/rig/critical_paths.py
@@ -234,9 +234,8 @@ def merge_into_baseline(statuses: list[CriticalPathStatus], destination: Path) -
     existing: set[str] = set()
     if destination.exists():
         try:
-            parsed = json.loads(destination.read_text())
-            existing = set(parsed.get("covered_paths") or [])
-        except (json.JSONDecodeError, OSError) as exc:
+            existing = set(_read_baseline(destination)["covered_paths"])
+        except (OSError, ValueError) as exc:
             raise BaselineCorruptError(
                 f"refusing to merge into corrupt baseline {destination}: {exc}. "
                 "Delete the cache entry and re-run."
@@ -248,18 +247,35 @@ def merge_into_baseline(statuses: list[CriticalPathStatus], destination: Path) -
     destination.write_text(json.dumps(payload, indent=2))
 
 
+def _read_baseline(source: Path) -> dict[str, Any]:
+    """Parse a baseline file and check its shape.
+
+    Well-formed JSON of the wrong shape is as unusable as a truncated file, and
+    would otherwise surface as an ``AttributeError``/``TypeError`` from whichever
+    caller touched it first. Raises ``OSError`` or ``ValueError``
+    (``JSONDecodeError`` is one) for callers to translate.
+    """
+    parsed = json.loads(source.read_text())
+    if not isinstance(parsed, dict):
+        raise ValueError(f"expected a JSON object, got {type(parsed).__name__}")
+    covered = parsed.get("covered_paths") or []
+    if not isinstance(covered, list) or not all(isinstance(p, str) for p in covered):
+        raise ValueError("'covered_paths' must be a list of strings")
+    return {**parsed, "covered_paths": covered}
+
+
 def load_baseline(source: Path) -> dict[str, Any] | None:
     """Load the cached baseline.
 
     Returns None if the file doesn't exist (first build / fresh cache).
-    Raises :class:`BaselineCorruptError` if the file exists but is unreadable
-    or unparseable — see :func:`merge_into_baseline` for the rationale.
+    Raises :class:`BaselineCorruptError` if the file exists but is unreadable,
+    unparseable, or misshapen — see :func:`merge_into_baseline` for the rationale.
     """
     if not source.exists():
         return None
     try:
-        return json.loads(source.read_text())
-    except (json.JSONDecodeError, OSError) as exc:
+        return _read_baseline(source)
+    except (OSError, ValueError) as exc:
         raise BaselineCorruptError(
             f"baseline at {source} is unreadable: {exc}. "
             "Delete the cache entry and re-run."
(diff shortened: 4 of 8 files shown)


tests/rig/cli.py
```diff
@@ -404,7 +424,7 @@ def cmd_report(args: argparse.Namespace) -> int:
             return 1
         cp.merge_into_baseline(statuses, baseline_path)
         print(f"[rig] merged into baseline: {baseline_path}")
-    return 1 if unknown_marker_ids else 0
+    return 1 if unknown_marker_ids or regressions else 0
```
Deepak-Kesavan: **[major] A corrupt cached baseline silently disables the new regression gate: `baseline = None` makes `previously_covered` empty, so no path can reach `state == "regression"` and line 427 always returns 0. `cmd_run` guards the identical condition at cli.py:688-689; `cmd_report` did not copy that guard.**

Precondition: the restored `reports/previous-summary.json` is truncated or unparseable (cache write interrupted, partial artifact).

Trace: `load_baseline` raises `BaselineCorruptError`; cli.py:368-373 swallows it into `baseline = None`, `baseline_corrupt = True`, and does not touch the return path. critical_paths.py:164 then yields an empty `previously_covered`, and critical_paths.py:191 (`elif path.id in previously_covered and in_scope`) is the ONLY producer of `state = "regression"` — so `regressions` at cli.py:404 is always `[]` and line 427 returns 0 no matter what regressed.

Why nothing else catches it: the one branch that does return 1 on `baseline_corrupt` is inside `if args.update_baseline:` (cli.py:417), and ci-test.yaml:295-299 adds `--update-baseline` only on `refs/heads/main`, so it never runs on a PR. The workflow backstop `Fail if any tier failed` (ci-test.yaml:337-345) reads only `needs.*.result` and cannot observe the report job's own verdict.

Why this is material rather than theoretical: every entry in tests/critical_paths.yaml uses `proof: marker`. For a marker path, deleting or un-marking the attesting test blanks `covering` (critical_paths.py:185-186) while the group and the tier stay green — so the tier-result backstop sees nothing, and this gate is the sole detector. That is precisely the regression the gate was added to catch.

Reproduced by driving the real `cmd_report` (harness shape copied from the PR's own `_drive_cmd_report`, test_cli.py:960) with one group `int-g` covering `int-path`, `int-g` red in every run, varying only the baseline blob:
  - `{"covered_paths": ["int-path"]}`   -> state `regression`, exit 1
  - `{"covered_paths": ["int-pat`  (truncated) -> state `gap`, exit 0

Two things this is NOT, both checked: a *missing* baseline exiting 0 is correct (nothing to regress against, and failing would break the first run and every post-eviction run); and the corrupt case is not invisible — reporting.py:267-270 does surface the "Baseline cache was corrupt; regression detection disabled this run" banner in the sticky comment and job summary. The defect is that a required check goes green on an advisory banner a human has to notice, which is the failure mode `merge_into_baseline`'s own docstring argues against.

**Suggested fix:** Mirror `cmd_run`: make line 427 `return 1 if unknown_marker_ids or regressions or baseline_corrupt else 0`. cli.py:688-689 already carries the rationale in a comment — "A red-then-fixed cycle without this flip would silently swallow the corrupt cache and disable regression detection on the next N builds" — which applies verbatim here.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:325820f2c202</sub>

Deepak-Kesavan (commented):
## 🤖 Automated review

2 inline comment(s).

tests/rig/cli.py
```diff
@@ -388,6 +398,16 @@ def cmd_report(args: argparse.Namespace) -> int:
             f"(not in tests/critical_paths.yaml)",
             file=sys.stderr,
         )
+    # This is the only cross-tier evaluation, so it is also the only place a
```
Deepak-Kesavan: **[minor] The new comment at cli.py:401-403 states "cmd_run deliberately doesn't" gate on regressions, but `cmd_run` in the same file does exactly that at cli.py:697-704.**

cli.py:401-403 reads: "This is the only cross-tier evaluation, so it is also the only place a regression can be gated on. cmd_run deliberately doesn't: a single tier can't tell 'another tier's path went red' from 'another tier didn't run'."

cli.py:697-704 in `cmd_run`:
```python
regressions = [s for s in statuses if s.state == "regression"]
if regressions:
    print(f"\n[rig] \u274c {len(regressions)} critical-path regression(s) detected", ...)
    if overall_exit == 0:
        overall_exit = 1
```
`cmd_run` gates on regressions today. What is actually true is the narrower claim the second sentence gestures at: `cmd_run` scopes to its own tier (cli.py:465), so it can only gate on regressions within that tier, while `cmd_report` is the only place cross-tier regressions can be gated.

This is worth fixing because the comment is the stated justification for the whole change, and it is the first thing the next reader will use to decide whether `cmd_report`'s exit path needs to match `cmd_run`'s — which is exactly the question behind the corrupt-baseline gap above, where `cmd_run` has a guard (cli.py:688-689) that `cmd_report` does not.

**Suggested fix:** Reword to the accurate claim: cmd_run gates on regressions it can see within its own tier; cmd_report is the only place a *cross-tier* regression can be gated.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:93f3d6831684</sub>

tests/rig/cli.py
chandrasekharan-zipstack: Confirmed and fixed in cb7ae4a0c.

Traced it through the same path: `load_baseline` raises → cli.py:372 sets `baseline = None` → `previously_covered` is empty at critical_paths.py:164 → critical_paths.py:191 is the only producer of `regression`, so `regressions` is always `[]`. The `baseline_corrupt` return-1 sits inside `if args.update_baseline:`, which ci-test.yaml only passes on `refs/heads/main`, so nothing on a PR observes it.

Applied the suggested fix verbatim:

```python
return 1 if unknown_marker_ids or regressions or baseline_corrupt else 0
```

Added `test_cmd_report_gates_on_a_corrupt_baseline`, which drives the real `cmd_report` with a truncated baseline blob and a red covering group. It asserts both halves of the defect — the path lands on `gap` (regression classification is genuinely unreachable) and the exit code is still 1. Confirmed it fails with `assert 0 == 1` against the unfixed return line, so it exercises the branch rather than passing vacuously.

The missing-baseline case stays at exit 0 as you noted it should: `load_baseline` returns `None` without raising when the file is absent, so `baseline_corrupt` is only set for a file that exists and is unreadable.

Rig self-tests: 107 passed, 1 skipped.

tests/rig/cli.py
chandrasekharan-zipstack: Correct — fixed in cb7ae4a0c. `cmd_run` does gate on regressions (cli.py:697-704); what is actually narrower is its scope, since it evaluates against its own tier only (cli.py:465).

Reworded to the accurate claim:

```python
# cmd_run gates on the regressions it can see, but only within its own tier.
# This is the only cross-tier evaluation, so it is the only place a regression
# spanning tiers can be gated on.
```

tests/rig/cli.py
praveen-formido: **[high] The gate is now unrecoverable from a deliberate coverage change, and main cannot clear it.**

`merge_into_baseline` only ever unions (`existing | fresh`) — it never prunes. So a path that stops being covered *on purpose* (test retired, `covered_by` re-pointed at another group, a `critical_path` marker renamed, a path ratcheted from `group` to `marker` before its tests are marked) stays in `covered_paths` forever, and from this PR on that is a hard failure of the `report` job on **every** subsequent PR.

Main can't repair it either, and the loop is closed by the workflow:

```yaml
- name: Save updated baseline
  if: github.ref == 'refs/heads/main'
```

That step carries no `always()`, so its implicit `success()` applies. On main the combine step now exits 1 for the same regression, the save step is skipped, and the stale baseline entry survives to the next run — where it fails again. The only exit is someone manually deleting the `unstract-test-baseline-main-*` cache entry, which is not documented anywhere and is not obvious from the failure message.

Two things would close this, either one sufficient:
- prune from the baseline any id that is no longer in the registry *and* any id whose `covered_by` no longer names a group that can attest it, in `merge_into_baseline`; and/or
- an explicit escape hatch (`--allow-regression <id>` or an `expected_regressions:` key in `critical_paths.yaml`) so a deliberate removal is expressible in the same PR that makes it.

Worth resolving before this lands, because the failure only shows up once someone legitimately retires a covered test, at which point every open PR goes red at the same time.

tests/rig/cli.py
```diff
@@ -371,8 +371,18 @@ def cmd_report(args: argparse.Namespace) -> int:
         print(f"[rig] {exc}", file=sys.stderr)
         baseline = None
         baseline_corrupt = True
+    # A tier the path filter skipped emits no junit, so its groups are absent
+    # here. Scoping to the groups that actually reported keeps their paths as
+    # gaps rather than regressions — nothing regressed, the tier never ran. A
+    # group that ran and went red still reports, so it stays in scope and a
+    # genuine regression is still caught.
+    scope_groups = [r.name for r in group_results]
```
praveen-formido: **[medium] `optional: true` silently stops meaning optional once a critical path points at an optional group.**

`group_results` includes optional groups (they run and emit junit), so they land in `scope_groups` here. A red optional group is not in `_coverage_attesting_groups` (that needs `status == STATUS_PASS`), so any baseline-covered path it covers becomes `in_scope=True` + not covered → `regression` → the new `return 1` fails the `report` job, which is the branch-protection check.

That contradicts the documented contract — `tests/README.md:115`:

> **non-blocking** — if the group runs and fails, its red result still shows in the summary but does not gate the overall exit code. Use for groups that need infra CI doesn't provision (e.g. live-DB connector tests) where a red run shouldn't block merge.

and `cmd_run` honours it (both exit-folds are gated on `not group.optional`), so the two commands would now disagree about the same red group.

Latent today — no path in `critical_paths.yaml` names `unit-platform-service`, `integration-workers`, `integration-workflow-execution` or `e2e-hurl`. But `integration-workflow-execution` is an optional placeholder explicitly marked *"until tests are authored"*, and the natural next step is to point `workflow-execution-fan-out` at it, at which point a flaky live-DB group starts blocking merges with no signal that `optional` was overridden.

Suggest dropping optional groups from the scope used for the gate (keep them in scope for reporting), or at minimum documenting the carve-out in README:115 so the contract is honest.

tests/rig/cli.py
```diff
@@ -388,6 +398,16 @@ def cmd_report(args: argparse.Namespace) -> int:
             f"(not in tests/critical_paths.yaml)",
             file=sys.stderr,
         )
+    # This is the only cross-tier evaluation, so it is also the only place a
+    # regression can be gated on. cmd_run deliberately doesn't: a single tier
+    # can't tell "another tier's path went red" from "another tier didn't run".
+    regressions = [s for s in statuses if s.state == "regression"]
```
praveen-formido: **[medium] A *skipped* marked test is now a hard build failure, not a warning.**

Every path in `critical_paths.yaml` is `proof: marker`, and `_marker_proven_paths` unions `passed_critical_path_ids(...)` — only **passing** marked tests attest. A test that skips leaves its group green (skips don't fail), so the group stays in `groups_run_green`… no, it stays in `scope_groups` and *out* of `covering`, giving exactly `in_scope=True` + not covered + in baseline → `regression` → exit 1, while the tier job itself is green. So this is a new way for `report` to go red on its own.

Concretely, from the commit right before this one (#2231), `mcp_server/tests/test_platform_auth.py`:

```python
if not settings.MCP_PLATFORM_SERVER_ENABLED:
    self.skipTest("MCP_PLATFORM_SERVER_ENABLED is off, ...")
```

On any tree where that flag is off, `integration-backend` runs green and in scope, and `mcp-platform-auth` flips from a printed warning to a required-check failure the moment that id is in that tree's baseline. The cloud tree is precisely such a tree — see my comment on `settings/test.py`, where commit 2's fix doesn't actually reach it. Same shape applies to any `skipif` that depends on the environment (`connector-register-test` is the obvious candidate: "validated against the live system").

Not necessarily wrong — a silently-skipped critical-path test arguably *should* shout — but it's a distinct new failure mode from the "empty group" case called out in the PR description, it fires with a green tier, and the message (`critical-path regression(s) detected`) gives no hint that the cause was a skip. Worth naming in the description at minimum, and ideally distinguishing "covering group green but marker unproven" from "covering group red" in the stderr line, since the fix is completely different.

backend/backend/settings/test.py
```diff
@@ -1,16 +1,8 @@
...
 
-# The organization-scoped MCP server ships disabled (see
-# MCP_PLATFORM_SERVER_ENABLED in base.py), but its tests must still exercise
-# it: several drive requests through the full URL stack precisely so the auth
-# middleware runs, and an unmounted route would make them 404 — turning a suite
-# that checks a credential is *rejected* into one that passes because nothing
-# is there. Shipping-off and untested are different things.
-MCP_PLATFORM_SERVER_ENABLED = True
+from backend.settings.base import *  # noqa: F401, F403
+from backend.settings.test_base import *  # noqa: F401, F403
```
praveen-formido, line 8: **[medium] The cloud half is missing, so this commit does not yet deliver the fix it describes.**

The commit message states the shape as:

```
OSS    test.py       = base  + test_base
cloud  test_cloud.py = cloud + test_base
```

The second line isn't true anywhere yet. In `unstract-cloud`, `backend/backend/settings/test_cloud.py` is still:

```python
from backend.settings.cloud import *  # noqa: F401, F403

DEBUG = True
PASSWORD_HASHERS = [...]
SESSION_COOKIE_SECURE = False
CSRF_COOKIE_SECURE = False
```

— no `test_base` import anywhere in that repo (grepped). `copy_cloud_deps` overwrites `settings/test.py` with the redirect to `test_cloud`, so on a cloud build nothing on this new import chain executes. `MCP_PLATFORM_SERVER_ENABLED` still doesn't reach the cloud suite, the platform-auth tests keep skipping there, and the cloud `test_cloud.py` keeps its duplicated copies of the three deltas this commit was meant to de-duplicate. As it stands this half is a pure OSS refactor.

Landing order matters here because of commit 1: with the gate on, `mcp-platform-auth` skipping on cloud is now a `report` failure rather than a warning as soon as that id is in the cloud baseline (see my comment on `cmd_report`). Please land the companion cloud PR in lockstep, and link it here — and note the cloud side needs `from backend.settings.test_base import *` placed **after** `from backend.settings.cloud import *`, since `test_cloud` derives `MIDDLEWARE` from `PRODUCTION_MIDDLEWARE` at import time.

backend/backend/settings/test_base.py
```diff
@@ -0,0 +1,32 @@
...
+# resistance to offline cracking.
+PASSWORD_HASHERS = ["django.contrib.auth.hashers.MD5PasswordHasher"]
+
+# Prod enforces HTTPS-only cookies; tests run over plain HTTP.
+SESSION_COOKIE_SECURE = False
+CSRF_COOKIE_SECURE = False
+
+# Env-driven and unset under test, which would make any request to an internal
+# API fail as "not configured" rather than on its own merits.
+INTERNAL_SERVICE_API_KEY = "test-internal-service-key"
```
praveen-formido: **[low] This delta does more than the comment says: it opens a blanket auth bypass across the whole suite.**

The stated motive is that an internal-API request should "fail on its own merits rather than as *not configured*". But `INTERNAL_SERVICE_API_KEY` is also read by `CustomAuthMiddleware`, which is in `MIDDLEWARE` for the OSS test settings (base.py:483) and is not stripped here:

```python
# backend/account_v2/custom_auth_middleware.py:37
if (
    settings.INTERNAL_SERVICE_API_KEY
    and x_api_key == settings.INTERNAL_SERVICE_API_KEY
):
    return self.get_response(request)   # fully authenticated, no user, no org check
```

Previously that branch was dead under test (the setting was `None`). Now a single `X-API-Key: test-internal-service-key` header authenticates any request in any OSS test. Nothing sends it today, so this isn't a live break — but `api-deployment-auth`, `mcp-server-auth` and `mcp-platform-auth` are exactly the critical paths that exist to prove credentials are *refused*, and this hands the suite a global skeleton key that would make a future negative test pass for the wrong reason. That's the same class of bug the `MCP_PLATFORM_SERVER_ENABLED` comment four lines below is guarding against.

`override_settings(INTERNAL_SERVICE_API_KEY=...)` on the handful of tests that actually drive `/internal/` would keep the fix and not widen the surface. If it does stay global, it's worth a second sentence in the comment naming the `CustomAuthMiddleware` side effect, since the next reader will only see the internal-API rationale.

Separately: `SESSION_COOKIE_SECURE` / `CSRF_COOKIE_SECURE` are also new here and already default to `False` in `base.py:109-113`, so they're a no-op on the OSS side — they only matter once cloud imports this. Fine, just noting they aren't a straight move like the other two.

tests/rig/reporting.py
```diff
@@ -339,6 +343,21 @@ def _render_markdown(
...
+            lines.append(
+                "<details><summary>💤 Covered, but not exercised in this "
+                "build</summary>"
+            )
+            lines.append("")
+            for s in unexercised:
+                covers = ", ".join(s.path.covered_by)
+                lines.append(
+                    f"- **{s.path.id}** — {s.path.description} "
+                    f"(covered by {covers}; that tier did not run here)"
```
praveen-formido: **[low] "that tier did not run here" asserts a cause the data doesn't support.**

In `cmd_report`, scope is inferred purely from junit presence (`scope_groups = [r.name for r in group_results]`), which conflates several causes:

- the tier was path-filtered out (the case this section is written for);
- the group ran but died before pytest wrote `junit.xml` — segfault/OOM/timeout kill. `_execute_group` returns `result is None` there;
- the tier ran but its artifact didn't make it back: `upload-artifact` uses `if-no-files-found: ignore`, so a missing `reports/` uploads nothing and still succeeds, and `download-artifact` then merges nothing for that tier.

In the second and third cases the report tells the reader the tier didn't run, when in fact it ran and its result was lost — and with commit 1's scoping that same inference also demotes those paths from `regression` to `gap`, so the report goes green. The tier-death case is caught by the workflow's `Fail if any tier failed` step, so this is mostly a wording problem today rather than a hole.

Suggest softening to something that doesn't claim a cause — e.g. `(covered by {covers}; no result reported in this build)` — or, better, deriving scope from the tiers the workflow actually selected rather than from junit presence, which would make the sentence true and close the artifact-loss case at the same time.

tests/rig/reporting.py
```diff
@@ -318,7 +318,11 @@ def _render_markdown(
 
     if critical_statuses:
         regressions = [s for s in critical_statuses if s.state == "regression"]
-        gaps = [s for s in critical_statuses if s.state == "gap"]
+        all_gaps = [s for s in critical_statuses if s.state == "gap"]
+        # A path whose covering groups exist but sat out this build is not
+        # "not yet covered" — reporting it as such reads as missing tests.
+        unexercised = [s for s in all_gaps if not s.in_scope and s.path.covered_by]
+        gaps = [s for s in all_gaps if s not in unexercised]
```
praveen-formido: **[low] `s not in unexercised` is an O(n²) scan that leans on `CriticalPathStatus.__eq__`.**

It works today only because `CriticalPathStatus` and `CriticalPath` are both `frozen=True` dataclasses, so `in` falls back to generated structural equality and each row is unique by `path.id`. Drop `frozen`/`eq` on either dataclass, or add a field that isn't comparable, and this silently mis-partitions rather than failing.

A single pass says the same thing and can't drift:

```python
unexercised, gaps = [], []
for s in all_gaps:
    (unexercised if not s.in_scope and s.path.covered_by else gaps).append(s)
```

praveen-formido (commented):
Careful pass over both commits. The scoping fix reads correct — `cmd_run` already did exactly this and `cmd_report` was the odd one out — and the two new tests genuinely drive the changed branch rather than passing vacuously.

My concerns are all downstream of the *gate*. Promoting a printed warning to a required-check failure changes the blast radius of several behaviours that were previously only noisy, and two of them are documented contracts that now quietly don't hold. Plus the settings commit's fix isn't reachable from the cloud tree yet, which interacts badly with the gate.

7 comments, roughly in severity order. Nothing here disputes the diagnosis in the description — it's about what else becomes blocking once this is in.

tests/rig/cli.py
chandrasekharan-zipstack: Half confirmed, half narrower than described — fixed the confirmed half in 3c9c31fe1.

**Confirmed:** `merge_into_baseline` only ever unions, so nothing ever leaves the cache. Now pruned to the registry:

```python
known = {s.path.id for s in statuses}
payload = {"covered_paths": sorted((existing | fresh) & known)}
```

`statuses` always spans the whole registry (`evaluate` iterates `registry.paths`), so this is safe across tier invocations — each one knows every id, and only differs in which are `covered`. That makes editing `critical_paths.yaml` the way to accept a deliberate removal, which is your second suggestion in a form that needs no new key.

**Narrower than described:** a stale id that is no longer in the registry was already inert, not a permanent failure. `_status_for` is only called per registry path, so a baseline id with no registry entry never produces a status and cannot reach `regression`. So of the four scenarios you list, "test retired" and "`covered_by` re-pointed" only bite if the path *stays* declared in `critical_paths.yaml` — at which point a hard failure is the ratchet working, and the fix is to drop or re-point the entry in the same PR.

I did not add `--allow-regression`/`expected_regressions:`. Editing the registry expresses the same intent in the file that already owns it, and a suppression list is a second source of truth that tends to accumulate. Happy to add it if you would rather have the removal be reviewable as a diff separate from the registry.

Also left `Save updated baseline` gated on `success()` deliberately: on a red main, saving a baseline computed from that build is what would silently absorb the regression. The stale-entry accumulation it was protecting against is what the prune now handles.

What the failure message was missing is the way out, which it now says:

```
[rig] to accept a deliberate removal, drop or re-point the path in
tests/critical_paths.yaml in the same PR — the baseline is keyed off that registry.
```

New test `test_baseline_merge_prunes_paths_dropped_from_the_registry`; confirmed it fails `assert ['p1', 'p2'] == ['p1']` against the un-pruned merge. Also fixed `test_baseline_merge_unions_with_existing`, which simulated two tiers with two *different* registries — real invocations share one registry file and differ only in which groups ran, and the union guarantee still holds under that shape.

tests/rig/cli.py
chandrasekharan-zipstack: Confirmed and fixed in 3c9c31fe1 — this was the one I would have been sorriest to ship.

Verified the mechanism exactly as you describe: `group_results` includes optional groups, `_coverage_attesting_groups` needs `status == STATUS_PASS`, so a red optional group is in scope and not attesting → `regression` → the gate fails the branch-protection check, while `cmd_run` folds its exit code only on `not group.optional`. Two commands, same red group, opposite verdicts.

Took your first suggestion rather than documenting the carve-out — the README contract is the one that should survive:

```python
scope_groups = [r.name for r in group_results if not manifest.get(r.name).optional]
```

Scope only governs the `elif` in `_status_for`, so a green optional group still marks its paths `covered` (it is in `groups_run_green`, and `covering` is checked first). Only the red case changes: those paths fall to `gap` and report without gating.

Also confirmed it is latent today — the `covered_by` union across `critical_paths.yaml` is `{integration-backend, e2e-*}` with zero overlap against `{unit-platform-service, integration-workers, integration-workflow-execution, e2e-hurl}`. Which is precisely why it needed catching now: pointing `workflow-execution-fan-out` at that placeholder is a one-line change with no hint attached that it would start gating merges.

New test `test_cmd_report_does_not_gate_on_an_optional_group`, confirmed failing with `an optional group's red result must not read as a regression` before the fix.

tests/rig/cli.py
chandrasekharan-zipstack: Confirmed — `passed_critical_path_ids` skips any testcase with a `skipped` child, so a skipped marked test leaves its group green and in scope while attesting nothing. Green tier, red report job.

I agree with your read that this *should* shout rather than be silently accepted; the defect was the message, which named neither the cause nor the fix. Split in 3c9c31fe1:

```
[rig] ❌ N critical-path regression(s) — no covering group ran green: <ids>
[rig] ❌ N critical-path regression(s) — covering group ran green but no passing
      @pytest.mark.critical_path test attested them (skipped or unmarked?): <ids>
```

The partition is `any(g in green for g in s.path.covered_by)` — green covering group present means the marker is the missing half.

One correction to the premise: not quite every path is `proof: marker` — `workflow-execution-fan-out` is `proof: group`, though it also has an empty `covered_by`, so it never reaches this branch. 17 of 18 are markers, so the shape holds.

Also named the failure mode in the PR description, and note the `mcp-platform-auth` case you cite is now doubly guarded: `MCP_PLATFORM_SERVER_ENABLED = True` in `test_base` means the `skipTest` in #2231 does not fire on either tree once the cloud side lands (see the reply on `settings/test.py`).

backend/backend/settings/test.py
chandrasekharan-zipstack: Right that it is not on cloud `main` — it is on a branch, which I should have linked here rather than leaving you to grep. Companion PR: **Zipstack/unstract-cloud#1684** (branch `test/mock-auth0-login`, commits `b5e239a1` + `f6fe6fa9`).

`test_cloud.py` there is now exactly the second line of the shape:

```python
from backend.settings.cloud import *  # noqa: F401, F403
from backend.settings.test_base import *  # noqa: F401, F403
```

— and the ordering constraint you name is already satisfied: `test_base` is imported second, so it overrides, while `MIDDLEWARE` stays whatever `cloud.py` derived from `PRODUCTION_MIDDLEWARE`. That commit also drops the exclusion set and the duplicated `DEBUG`/`PASSWORD_HASHERS`/cookie deltas, which is the de-duplication this half was for. Verified on a merged tree: 1681 passed, 32 skipped, with the MCP platform-auth suite running (6 tests + 6 subtests) rather than skipping.

One thing worth naming rather than glossing: #1684 is a mock-Auth0 PR, so "land in lockstep" is not as clean as it should be — the settings change is stacked on unrelated work in both repos (#2232 is the rig-report PR). That was your call to restack; flagging it because the ordering hazard you describe is real and the two halves are now gated on two unrelated reviews. If you would rather, I can split both settings commits into their own paired PRs so they land together and these two land independently.

Until then this half is, as you say, an OSS-only refactor — correct but inert on cloud.

backend/backend/settings/test_base.py
chandrasekharan-zipstack: Good catch, and the delta is gone entirely in 3c9c31fe1 rather than scoped.

Confirmed the bypass at `custom_auth_middleware.py:37`, and it is worse than the comment implied in one respect: that branch sits *after* the `/internal/` prefix skip, so it never serves internal-API requests at all — those are handled by `InternalAPIAuthMiddleware`. The setting's only effect on the OSS suite was the skeleton key on ordinary paths. The stated rationale was speculative and the side effect was the whole of it.

Rather than `override_settings` on the tests that drive `/internal/`, I checked whether any test needs it: nothing in `backend/` reads or sends the key outside `pipeline_v2/manager.py` (production code). Ran the OSS suite both ways — **856 passed, 29 skipped** with and without. So it was buying nothing and handing out a global credential. Deleted.

On the cookie flags: correct that they are a no-op against `base.py:109-113`, which reads them from env defaulting False. Keeping them, because `cloud.py:17-18` sets both to `True` unconditionally — they are the delta that makes the cloud suite work over plain HTTP, which is the case this module exists to serve. Not a straight move, as you say.

tests/rig/reporting.py
chandrasekharan-zipstack: Agreed — took the wording fix in 3c9c31fe1:

```
(covered by {covers}; no result reported in this build)
```

That is what junit presence actually supports. The three causes you list are genuinely indistinguishable at this point in the code, and claiming the first one reads as reassurance in exactly the artifact-loss case where it should not.

Did not take the second half (deriving scope from the tiers the workflow selected). It is the better fix and it would close the artifact-loss hole rather than just describing it honestly, but it means threading the workflow's tier selection into `report combine` as input, which is a bigger change than the rest of this PR and wants its own review. Worth a follow-up issue — happy to file one, or leave it to you if you would rather scope it.

tests/rig/reporting.py
chandrasekharan-zipstack: Taken in 3c9c31fe1, near enough verbatim:

```python
unexercised: list[CriticalPathStatus] = []
gaps: list[CriticalPathStatus] = []
for s in critical_statuses:
    if s.state != "gap":
        continue
    target = unexercised if not s.in_scope and s.path.covered_by else gaps
    target.append(s)
```

Folded the `state == "gap"` filter into the same pass, which drops the intermediate `all_gaps` list too. You are right that the old form was only correct by accident — structural equality on a `frozen=True` dataclass, with uniqueness resting on `path.id` rather than anything enforced.

Applied the same shape to the new regression partition in `cmd_report`, which I had written with the same `not in` pattern before reading this.

tests/rig/critical_paths.py
```diff
@@ -238,7 +242,8 @@ def merge_into_baseline(statuses: list[CriticalPathStatus], destination: Path) -
                 "Delete the cache entry and re-run."
             ) from exc
     fresh = {s.path.id for s in statuses if s.state == "covered"}
-    payload = {"covered_paths": sorted(existing | fresh)}
+    known = {s.path.id for s in statuses}
+    payload = {"covered_paths": sorted((existing | fresh) & known)}
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Valid — fixed in 68482924b. Both failure modes reproduce exactly as described: `[]` gives `AttributeError: 'list' object has no attribute 'get'` out of `merge_into_baseline`, and `{"covered_paths": 1}` sails through both functions and only breaks later.

One shared checker now backs both callers:

```python
def _read_baseline(source: Path) -> dict[str, Any]:
    parsed = json.loads(source.read_text())
    if not isinstance(parsed, dict):
        raise ValueError(f"expected a JSON object, got {type(parsed).__name__}")
    covered = parsed.get("covered_paths") or []
    if not isinstance(covered, list) or not all(isinstance(p, str) for p in covered):
        raise ValueError("'covered_paths' must be a list of strings")
    return {**parsed, "covered_paths": covered}
```

It raises `OSError`/`ValueError` (`JSONDecodeError` being one) and each caller translates to `BaselineCorruptError` with its own message, so the existing corrupt-cache handling in `cmd_report` covers the new cases without a second except arm.

Added `test_baseline_of_the_wrong_shape_is_corrupt`, parametrised over `[]`, `{"covered_paths": 1}` and `{"covered_paths": ["ok", 2]}` — your suggested cases plus a bad element type — asserting both entry points. Confirmed all three fail without the checker.

One correction to the severity framing: an unhandled `AttributeError` would still have *failed* the report job rather than letting it pass — the gate was not bypassed. What it lost was the summary and PR comment, and a traceback in place of the "delete the cache entry and re-run" message. Worth fixing on those grounds, which is why it is in.

praveen-formido (approved):

PR #2234, opened 2026-08-07
UN-2651 [FIX] Show execution logs to group and org-shared users

kirtimanmishrazipstack:
## What

- Teammates who get a deployment or pipeline shared with them through a group, or shared with the whole organisation, can now see its runs on the Logs page. That page used to come up empty for them.
- A run's logs, its per-file results and its execution list can now only be opened by people the deployment or pipeline was actually shared with.

## Why

- A shared deployment whose Logs page is empty reads as broken, and it let down exactly the teammates sharing was meant to help. Sharing with one person already worked; sharing through a group or with the whole org did not.
- Anyone in the organisation could read a run's logs if they had its link, even when nothing had been shared with them.

## How

- Executions resolve visibility through each resource's own `for_user`, which covers owner, co-owner, direct share, group share and `shared_to_org`.
- Every `execution/<id>/...` route now calls one shared gate (`workflow_manager/execution/access.py`) — logs, log export and the per-file list all take the same id and are authorised the same way. An unknown id is denied exactly like an inaccessible one, so responses do not reveal which ids exist; the server logs the difference instead.
- Dropped `IsOwner` from the log viewset and from `/workflow/<id>/execution/` — it implements only `has_object_permission`, which DRF never calls on list/export. Both viewsets are read-only and gate on the queryset instead.

- Three read endpoints narrow: `execution/<id>/logs/`, `execution/<id>/files/` and `workflow/<id>/execution/`. A user who could previously fetch any of them by id now needs access to that execution. No UI path exposed those ids to users without access, so normal flows are unaffected. The other change only widens visibility.

## Database Migrations

- None

## Env Config

- None

## Relevant Docs

-

## Related Issues or PRs

-

## Dependencies Versions

- None

### Automated

`backend/workflow_manager/execution/tests/test_shared_execution_access.py` covers the sharing matrix (owner, co-owner, direct viewer, group, org-wide, cross-org), the two bypass roles, and both log endpoints plus the per-file endpoint at the HTTP layer. Each gate was checked by mutation — removing it fails a named test.

### Manual

Two accounts in one organisation: an **owner/admin** and a **second user** who is a member of a group.

**Part 1 — the UI (visibility)**

| # | Who | Action | Result |
|---|---|---|---|
| 1 | Owner | Opens the Logs page for a workflow run | Logs listed |
| 2 | Owner | Exports the run's logs | CSV and JSON both download |
| 3 | Second user | Workflow is group-shared; opens the Logs page | Sees the same logs — **this is the fix** |
| 4 | Owner / admin | Unshares the workflow from the group (removing the user from the group does the same) | Workflow disappears from the second user's UI, and its logs with it |

**Part 2 — the API (access control)**

The UI never links to a resource you can't see, so the gate itself can only be exercised by calling the endpoint directly.

| # | Who | Action | Result |
|---|---|---|---|
| 5 | Second user | `GET /execution/<execution_id>/logs/` with the workflow unshared | **403** — `You do not have access to this execution.` |
| 6 | Second user | `GET /execution/<execution_id>/files/` with the workflow unshared | **403**, same message |

Worth knowing: access is granted through the workflow **or** the API deployment **or** the pipeline. Unsharing only the workflow still returns 200, correctly — every path has to be revoked before the 403 appears.

This is added in screenshot below

## Checklist

### DIFF
--- a/backend/workflow_manager/execution/tests/test_shared_execution_access.py
+++ b/backend/workflow_manager/execution/tests/test_shared_execution_access.py
@@ -6,18 +6,22 @@ import secrets
 import uuid
 from unittest.mock import patch
 
+from account_v2.models import Organization
 from api_v2.models import APIDeployment
 from django.contrib.contenttypes.models import ContentType
 from django.utils import timezone
+from permissions.roles import ResourceRole
+from pipeline_v2.models import Pipeline
 from rest_framework.exceptions import PermissionDenied
-from rest_framework.test import APIRequestFactory
+from rest_framework.test import APIRequestFactory, force_authenticate
 from tenant_account_v2.models import ResourceGroupShare
-from tenant_account_v2.tests import GroupSharingTestBase
+from tenant_account_v2.tests import GroupSharingTestBase, _add_viewers
 
 from workflow_manager.workflow_v2.enums import ExecutionStatus
 from workflow_manager.workflow_v2.execution_log_view import WorkflowExecutionLogViewSet
 from workflow_manager.workflow_v2.models.execution import WorkflowExecution
 from workflow_manager.workflow_v2.models.execution_log import ExecutionLog
+from workflow_manager.workflow_v2.models.workflow import Workflow
 
 _ADMIN_PREDICATE = (
     "tenant_account_v2.organization_member_service."
@@ -27,7 +31,7 @@ _ADMIN_PREDICATE = (
 
 class SharedExecutionAccessTests(GroupSharingTestBase):
     """``self.member`` belongs to ``self.group``; ``self.outsider`` is an org
-    member with no share of any kind. Neither owns anything here.
+    member with no share of any kind. ``self.owner`` owns every fixture here.
     """
 
     def setUp(self) -> None:
@@ -39,29 +43,48 @@ class SharedExecutionAccessTests(GroupSharingTestBase):
 
     def _api_deployment(self, *, shared_to_org: bool = False) -> APIDeployment:
         # api_name must be short — it defaults to a UUID longer than the column.
-        return APIDeployment.objects.create(
+        deployment = APIDeployment.objects.create(
             api_name=f"api-{secrets.token_hex(4)}",
             workflow=self.workflow,
             organization=self.org,
             created_by=self.owner,
             shared_to_org=shared_to_org,
         )
+        # Creator access flows through an OWNER row, not ``created_by`` (UN-2202);
+        # mirrors what ``APIDeploymentViewSet.perform_create`` does.
+        deployment.memberships.create(user=self.owner, role=ResourceRole.OWNER)
+        return deployment
 
-    def _execution(self, deployment: APIDeployment) -> WorkflowExecution:
+    def _pipeline(self, *, shared_to_org: bool = False) -> Pipeline:
+        pipeline = Pipeline.objects.create(
+            pipeline_name=f"pipe-{secrets.token_hex(4)}",
+            workflow=self.workflow,
+            organization=self.org,
+            created_by=self.owner,
+            shared_to_org=shared_to_org,
+        )
+        pipeline.memberships.create(user=self.owner, role=ResourceRole.OWNER)
+        return pipeline
+
+    def _execution(self, resource=None) -> WorkflowExecution:
+        """``resource=None`` builds a workflow-level execution (``pipeline_id``
+        NULL), which is the only shape that exercises the workflow branch of
+        ``for_user``.
+        """
         return WorkflowExecution.objects.create(
             workflow=self.workflow,
-            pipeline_id=deployment.id,
+            pipeline_id=resource.id if resource else None,
             execution_mode=WorkflowExecution.Mode.INSTANT,
             execution_method=WorkflowExecution.Method.DIRECT,
             execution_type=WorkflowExecution.Type.COMPLETE,
             status=ExecutionStatus.COMPLETED,
         )
 
-    def _share_with_group(self, deployment: APIDeployment) -> None:
+    def _share_with_group(self, resource) -> None:
         ResourceGroupShare.objects.create(
             group=self.group,
-            content_type=ContentType.objects.get_for_model(APIDeployment),
-            object_id=str(deployment.id),
+            content_type=ContentType.objects.get_for_model(type(resource)),
+            object_id=str(resource.pk),
             organization=self.org,
         )
 
@@ -79,6 +102,33 @@ class SharedExecutionAccessTests(GroupSharingTestBase):
         view.kwargs = {"pk": str(execution_id)}
         return view.get_queryset()
 
+    def _call(self, action: str, user, execution_id, **query):
+        """Drive the real endpoint, so ``dispatch`` translates the gate to HTTP."""
+        view = WorkflowExecutionLogViewSet.as_view({"get": action})
+        request = APIRequestFactory().get("/", query)
+        force_authenticate(request, user=user)
+        return view(request, pk=str(execution_id))
+
+    # --- visibility: who sees which executions ---------------------------------
+
+    def test_unshared_deployment_is_visible_only_to_its_owner(self) -> None:
+        execution = self._execution(self._api_deployment())
+        # The owner assertion is what makes the denial below a real control:
+        # without the OWNER membership row the deployment would be visible to
+        # nobody, and the denial would pass with the sharing filter deleted.
+        self.assertTrue(self._visible_to(self.owner, execution))
+        self.assertFalse(self._visible_to(self.outsider, execution))
+
+    def test_direct_viewer_sees_the_deployment_executions(self) -> None:
+        deployment = self._api_deployment()
+        _add_viewers(deployment, self.outsider)
+        self.assertTrue(self._visible_to(self.outsider, self._execution(deployment)))
+
+    def test_co_owner_sees_the_deployment_executions(self) -> None:
+        deployment = self._api_deployment()
+        deployment.memberships.create(user=self.member, role=ResourceRole.OWNER)
+        self.assertTrue(self._visible_to(self.member, self._execution(deployment)))
+
     def test_org_wide_share_exposes_the_deployment_executions(self) -> None:
         execution = self._execution(self._api_deployment(shared_to_org=True))
         self.assertTrue(self._visible_to(self.outsider, execution))
@@ -92,20 +142,61 @@ class SharedExecutionAccessTests(GroupSharingTestBase):
         # Same org, not in the group — still nothing.
         self.assertFalse(self._visible_to(self.outsider, execution))
 
-    def test_unshared_deployment_stays_invisible(self) -> None:
-        execution = self._execution(self._api_deployment())
+    def test_workflow_level_execution_follows_the_workflow_share(self) -> None:
+        execution = self._execution()  # pipeline_id NULL
+        self.assertTrue(self._visible_to(self.owner, execution))
+        self.assertFalse(self._visible_to(self.member, execution))
+
+        self._share_with_group(self.workflow)
+        self.assertTrue(self._visible_to(self.member, execution))
         self.assertFalse(self._visible_to(self.outsider, execution))
+
+    def test_pipeline_execution_follows_the_pipeline_share(self) -> None:
+        private = self._execution(self._pipeline())
+        self.assertTrue(self._visible_to(self.owner, private))
+        self.assertFalse(self._visible_to(self.outsider, private))
+
+        shared = self._execution(self._pipeline(shared_to_org=True))
+        self.assertTrue(self._visible_to(self.outsider, shared))
+
+    def test_org_wide_share_does_not_cross_organizations(self) -> None:
+        """``shared_to_org`` means *this* org — the tenant boundary for
+        ``/execution/`` is the manager, since the view drops the org filter
+        backend.
+        """
+        other_org = Organization.objects.create(
+            name="org-b", display_name="Org B", organization_id="org-b"
+        )
+        other_workflow = Workflow.objects.create(
+            workflow_name="wf-b", organization=other_org, created_by=self.owner
+        )
+        foreign = APIDeployment.objects.create(
+            api_name=f"api-{secrets.token_hex(4)}",
+            workflow=other_workflow,
+            organization=other_org,
+            created_by=self.owner,
+            shared_to_org=True,
+        )
+        execution = WorkflowExecution.objects.create(
+            workflow=other_workflow,
+            pipeline_id=foreign.id,
+            execution_mode=WorkflowExecution.Mode.INSTANT,
+            execution_method=WorkflowExecution.Method.DIRECT,
+            execution_type=WorkflowExecution.Type.COMPLETE,
+            status=ExecutionStatus.COMPLETED,
+        )
+
+        # UserContext still points at org A throughout.
+        self.assertFalse(self._visible_to(self.outsider, execution))
+        with self.assertRaises(PermissionDenied):
+            self._log_queryset(self.outsider, execution.id)
+
+    # --- the log endpoints ------------------------------------------------------
 
     def test_logs_denied_when_the_execution_is_not_accessible(self) -> None:
         execution = self._execution(self._api_deployment())
         with self.assertRaises(PermissionDenied):
             self._log_queryset(self.outsider, execution.id)
-
-    def test_logs_denied_when_the_execution_is_unknown(self) -> None:
-        # Same denial as an inaccessible execution — the response must not
-        # reveal which execution ids exist.
-        with self.assertRaises(PermissionDenied):
-            self._log_queryset(self.outsider, uuid.uuid4())
 
     def test_logs_readable_once_the_deployment_is_shared(self) -> None:
         execution = self._execution(self._api_deployment(shared_to_org=True))
@@ -113,3 +204,41 @@ class SharedExecutionAccessTests(GroupSharingTestBase):
             wf_execution=execution, data={"log": "hello"}, event_time=timezone.now()
         )
         self.assertIn(log, list(self._log_queryset(self.outsider, execution.id)))
+
+    def test_both_endpoints_return_403_for_an_inaccessible_execution(self) -> None:
+        execution = self._execution(self._api_deployment())
+        for action in ("list", "export"):
+            with self.subTest(action=action):
+                response = self._call(action, self.outsider, execution.id)
+                self.assertEqual(response.status_code, 403)
+
+    def test_both_endpoints_serve_a_shared_execution(self) -> None:
+        execution = self._execution(self._api_deployment(shared_to_org=True))
+        ExecutionLog.objects.create(
+            wf_execution=execution, data={"log": "hello"}, event_time=timezone.now()
+        )
+
+        listed = self._call("list", self.outsider, execution.id)
+        self.assertEqual(listed.status_code, 200)
+        listed.render()
+        self.assertIn(b"hello", listed.content)
+
+        exported = self._call(
+            "export", self.outsider, execution.id, file_format="csv"
+        )
+        self.assertEqual(exported.status_code, 200)
+        self.assertIn(b"hello", exported.content)
+
+    def test_unknown_execution_is_indistinguishable_from_an_inaccessible_one(
+        self,
+    ) -> None:
+        """An enumerator observes the response, not the exception — so the
+        status and body must match, not merely the exception type.
+        """
+        inaccessible = self._execution(self._api_deployment())
+        denied = self._call("list", self.outsider, inaccessible.id)
+        unknown = self._call("list", self.outsider, uuid.uuid4())
+
+        self.assertEqual(denied.status_code, 403)
+        self.assertEqual(unknown.status_code, denied.status_code)
+        self.assertEqual(unknown.data, denied.data)
--- a/backend/workflow_manager/execution/tests/test_shared_execution_access.py
+++ b/backend/workflow_manager/execution/tests/test_shared_execution_access.py
@@ -1,0 +1,108 @@
+"""Executions and their logs follow the sharing paths of the resource they belong
+to — memberships, group shares and ``shared_to_org`` (UN-2651).
+"""
+
+import secrets
+from unittest.mock import patch
+
+from api_v2.models import APIDeployment
+from django.contrib.contenttypes.models import ContentType
+from django.utils import timezone
+from rest_framework.exceptions import PermissionDenied
+from rest_framework.test import APIRequestFactory
+from tenant_account_v2.models import ResourceGroupShare
+from tenant_account_v2.tests import GroupSharingTestBase
+
+from workflow_manager.workflow_v2.enums import ExecutionStatus
+from workflow_manager.workflow_v2.execution_log_view import WorkflowExecutionLogViewSet
+from workflow_manager.workflow_v2.models.execution import WorkflowExecution
+from workflow_manager.workflow_v2.models.execution_log import ExecutionLog
+
+_ADMIN_PREDICATE = (
+    "tenant_account_v2.organization_member_service."
+    "OrganizationMemberService.is_user_organization_admin"
+)
+
+
+class SharedExecutionAccessTests(GroupSharingTestBase):
+    """``self.member`` belongs to ``self.group``; ``self.outsider`` is an org
+    member with no share of any kind. Neither owns anything here.
+    """
+
+    def setUp(self) -> None:
+        super().setUp()
+        # Pin admin resolution: "admins see everything" would mask the paths here.
+        patcher = patch(_ADMIN_PREDICATE, return_value=False)
+        patcher.start()
+        self.addCleanup(patcher.stop)
+
+    def _api_deployment(self, *, shared_to_org: bool = False) -> APIDeployment:
+        # api_name must be short — it defaults to a UUID longer than the column.
+        return APIDeployment.objects.create(
+            api_name=f"api-{secrets.token_hex(4)}",
+            workflow=self.workflow,
+            organization=self.org,
+            created_by=self.owner,
+            shared_to_org=shared_to_org,
+        )
+
+    def _execution(self, deployment: APIDeployment) -> WorkflowExecution:
+        return WorkflowExecution.objects.create(
+            workflow=self.workflow,
+            pipeline_id=deployment.id,
+            execution_mode=WorkflowExecution.Mode.INSTANT,
+            execution_method=WorkflowExecution.Method.DIRECT,
+            execution_type=WorkflowExecution.Type.COMPLETE,
+            status=ExecutionStatus.COMPLETED,
+        )
+
+    def _share_with_group(self, deployment: APIDeployment) -> None:
+        ResourceGroupShare.objects.create(
+            group=self.group,
+            content_type=ContentType.objects.get_for_model(APIDeployment),
+            object_id=str(deployment.id),
+            organization=self.org,
+        )
+
+    def _visible_to(self, user, execution: WorkflowExecution) -> bool:
+        return (
+            WorkflowExecution.objects.for_user(user).filter(pk=execution.pk).exists()
+        )
+
+    def _log_queryset(self, user, execution_id):
+        """Run the log viewset's queryset build for ``user`` — the access gate."""
+        request = APIRequestFactory().get("/")
+        request.user = user
+        view = WorkflowExecutionLogViewSet()
+        view.request = request
+        view.kwargs = {"pk": str(execution_id)}
+        return view.get_queryset()
+
+    def test_org_wide_share_exposes_the_deployment_executions(self) -> None:
+        execution = self._execution(self._api_deployment(shared_to_org=True))
+        self.assertTrue(self._visible_to(self.outsider, execution))
+
+    def test_group_share_exposes_the_deployment_executions(self) -> None:
+        deployment = self._api_deployment()
+        self._share_with_group(deployment)
+        execution = self._execution(deployment)
+
+        self.assertTrue(self._visible_to(self.member, execution))
+        # Same org, not in the group — still nothing.
+        self.assertFalse(self._visible_to(self.outsider, execution))
+
+    def test_unshared_deployment_stays_invisible(self) -> None:
+        execution = self._execution(self._api_deployment())
+        self.assertFalse(self._visible_to(self.outsider, execution))
+
+    def test_logs_denied_when_the_execution_is_not_accessible(self) -> None:
+        execution = self._execution(self._api_deployment())
+        with self.assertRaises(PermissionDenied):
+            self._log_queryset(self.outsider, execution.id)
+
+    def test_logs_readable_once_the_deployment_is_shared(self) -> None:
+        execution = self._execution(self._api_deployment(shared_to_org=True))
+        log = ExecutionLog.objects.create(
+            wf_execution=execution, data={"log": "hello"}, event_time=timezone.now()
+        )
+        self.assertIn(log, list(self._log_queryset(self.outsider, execution.id)))
--- a/backend/workflow_manager/execution/tests/test_shared_execution_access.py
+++ b/backend/workflow_manager/execution/tests/test_shared_execution_access.py
@@ -3,6 +3,7 @@ to — memberships, group shares and ``shared_to_org`` (UN-2651).
 """
 
 import secrets
+import uuid
 from unittest.mock import patch
 
 from api_v2.models import APIDeployment
@@ -99,6 +100,12 @@ class SharedExecutionAccessTests(GroupSharingTestBase):
         execution = self._execution(self._api_deployment())
         with self.assertRaises(PermissionDenied):
             self._log_queryset(self.outsider, execution.id)
+
+    def test_logs_denied_when_the_execution_is_unknown(self) -> None:
+        # Same denial as an inaccessible execution — the response must not
+        # reveal which execution ids exist.
+        with self.assertRaises(PermissionDenied):
+            self._log_queryset(self.outsider, uuid.uuid4())
 
     def test_logs_readable_once_the_deployment_is_shared(self) -> None:
         execution = self._execution(self._api_deployment(shared_to_org=True))
--- a/backend/workflow_manager/execution/tests/test_shared_execution_access.py
+++ b/backend/workflow_manager/execution/tests/test_shared_execution_access.py
@@ -34,8 +34,7 @@ _ADMIN_PREDICATE = (
 
 class SharedExecutionAccessTests(GroupSharingTestBase):
     """``self.member`` belongs to ``self.group``; ``self.outsider`` is an org
-    member with no share of any kind. ``self.owner`` owns every fixture in org A
-    (the cross-org test builds its own, deliberately unowned).
+    member with no share of any kind. ``self.owner`` owns the org A fixtures.
     """
 
     def setUp(self) -> None:
@@ -54,8 +53,7 @@ class SharedExecutionAccessTests(GroupSharingTestBase):
             created_by=self.owner,
             shared_to_org=shared_to_org,
         )
-        # Creator access flows through an OWNER row, not ``created_by`` (UN-2202);
-        # mirrors what ``APIDeploymentViewSet.create`` does after ``perform_create``.
+        # Creator access flows through an OWNER row, not ``created_by`` (UN-2202).
         deployment.memberships.create(user=self.owner, role=ResourceRole.OWNER)
         return deployment
 
@@ -98,7 +96,7 @@ class SharedExecutionAccessTests(GroupSharingTestBase):
         )
 
     def _log_queryset(self, user, execution_id):
-        """Run the log viewset's queryset build for ``user`` — the access gate."""
+        """Run the log viewset's queryset build for ``user``."""
         request = APIRequestFactory().get("/")
         request.user = user
         view = WorkflowExecutionLogViewSet()
@@ -117,9 +115,7 @@ class SharedExecutionAccessTests(GroupSharingTestBase):
 
     def test_unshared_deployment_is_visible_only_to_its_owner(self) -> None:
         execution = self._execution(self._api_deployment())
-        # The owner assertion is what makes the denial below a real control:
-        # without the OWNER membership row the deployment would be visible to
-        # nobody, and the denial would pass with the sharing filter deleted.
+        # Positive control: without it the denial passes even with no filter.
         self.assertTrue(self._visible_to(self.owner, execution))
         self.assertFalse(self._visible_to(self.outsider, execution))
 
@@ -164,11 +160,7 @@ class SharedExecutionAccessTests(GroupSharingTestBase):
         self.assertTrue(self._visible_to(self.outsider, shared))
 
     def test_workflow_share_does_not_expose_unshared_deployment_runs(self) -> None:
-        """Pins the ``pipeline_id__isnull=True`` conjunct on the workflow branch.
-
-        Workflow access must not leak into runs of a deployment that was
-        deliberately not shared — the "every path has to be revoked" behaviour.
-        """
+        """Workflow access must not leak into runs of an unshared deployment."""
         deployment = self._api_deployment()  # shared with nobody
         execution = self._execution(deployment)
         self._share_with_group(self.workflow)
@@ -177,18 +169,15 @@ class SharedExecutionAccessTests(GroupSharingTestBase):
         self.assertFalse(self._visible_to(self.member, execution))
 
     def test_deployment_share_does_not_expose_workflow_level_runs(self) -> None:
-        """The other half: a deployment share says nothing about the workflow's
-        own runs, which is why the two branches are not symmetric.
+        """The other direction: a deployment share says nothing about the
+        workflow's own runs.
         """
         deployment = self._api_deployment(shared_to_org=True)
         self.assertTrue(self._visible_to(self.outsider, self._execution(deployment)))
         self.assertFalse(self._visible_to(self.outsider, self._execution()))
 
     def test_org_wide_share_does_not_cross_organizations(self) -> None:
-        """``shared_to_org`` means *this* org — the tenant boundary for
-        ``/execution/`` is the manager, since the view drops the org filter
-        backend.
-        """
+        """``shared_to_org`` means *this* org."""
         other_org = Organization.objects.create(
             name="org-b", display_name="Org B", organization_id="org-b"
         )
@@ -211,8 +200,7 @@ class SharedExecutionAccessTests(GroupSharingTestBase):
             status=ExecutionStatus.COMPLETED,
         )
 
-        # Control: the same shape inside org A *is* visible, so the denial below
-        # is about the org boundary and not about the fixture being malformed.
+        # Control: the same shape inside org A is visible.
         local = self._execution(self._api_deployment(shared_to_org=True))
         self.assertTrue(self._visible_to(self.outsider, local))
 
@@ -282,8 +270,8 @@ class SharedExecutionAccessTests(GroupSharingTestBase):
         return view(request, pk=str(execution_id))
 
     def test_file_executions_follow_the_same_gate_as_the_logs(self) -> None:
-        """``<id>/files/`` carries file names, per-file errors and the latest log
-        line for the same id — 403 on logs and 200 here would defeat the point.
+        """``<id>/files/`` carries log text for the same id, so it must not
+        answer 200 where the logs answer 403.
         """
         execution = self._execution(self._api_deployment())
         self.assertEqual(self._call_files(self.outsider, execution.id).status_code, 403)
@@ -292,7 +280,7 @@ class SharedExecutionAccessTests(GroupSharingTestBase):
         self.assertEqual(self._call_files(self.outsider, shared.id).status_code, 200)
 
     def test_workflow_execution_list_is_scoped_to_accessible_workflows(self) -> None:
-        """``/workflow/<id>/execution/`` had the same dead ``IsOwner`` gate."""
+        """The execution list is scoped the same way as the logs."""
         execution = self._execution()  # workflow-level, owner-only
 
         view = WorkflowExecutionViewSet.as_view({"get": "list"})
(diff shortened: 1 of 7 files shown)


backend/workflow_manager/execution/tests/test_shared_execution_access.py
```diff
@@ -0,0 +1,108 @@
...
+        self.assertFalse(self._visible_to(self.outsider, execution))
+
+    def test_unshared_deployment_stays_invisible(self) -> None:
+        execution = self._execution(self._api_deployment())
+        self.assertFalse(self._visible_to(self.outsider, execution))
+
+    def test_logs_denied_when_the_execution_is_not_accessible(self) -> None:
+        execution = self._execution(self._api_deployment())
+        with self.assertRaises(PermissionDenied):
+            self._log_queryset(self.outsider, execution.id)
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @coderabbitai Valid — the indistinguishable-denial behaviour was documented in `execution_log_view.py:43-44` but untested. Added `test_logs_denied_when_the_execution_is_unknown` in `test_shared_execution_access.py`; 6/6 pass locally.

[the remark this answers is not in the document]
kirtimanmishrazipstack: @coderabbitai Fixed in ef3c01d43 — `test_logs_denied_when_the_execution_is_unknown` asserts `PermissionDenied` for a random UUID, matching the denial for an inaccessible execution so responses cannot be used to probe which ids exist. Suite is 6/6 green.

chandrasekharan-zipstack (commented):
## Standardized PR Review — INITIAL

**Verdict: `REQUEST CHANGES`**

**Critical: 0 · High: 4 · Medium: 5 · Low: 3 · Lenses run: 16/16**

The change itself is correct and is a net security improvement — it fixes a real intra-org log-read hole and a real visibility bug, and nothing in the diff is a regression. `REQUEST CHANGES` is for three things: the PR's stated security claim is falsified by a sibling endpoint on the same URL prefix, the identical anti-pattern the PR diagnoses is left in place one file away, and the negative control in the new test suite asserts nothing.

Reviewed at head `ef3c01d`. Posted as `COMMENT`, so it carries no merge-gate weight. The 3 Low findings are in a separate follow-up comment.

---

### Unanchored findings

Both of these are **High**, and both are about files this PR does not touch — so there is no diff line to attach them to. They are the two most consequential findings in this review.

#### [High] [Lens 1 — Spec & intent, Lens 4 — Security] — `/execution/<id>/files/` still returns log text for any execution id

**Would anchor to:** `backend/workflow_manager/file_execution/views.py:14-36`

The PR body states: *"A run's logs can now only be opened by people the deployment or pipeline was actually shared with."* That is not true after this diff.

`FileCentricExecutionViewSet` is mounted on the **same** `execution/` prefix as the two endpoints being gated — `backend/backend/urls_v2.py:65` and `:66` — and is fetched by the same UI page. It carries `permission_classes = [IsAuthenticated]` and keys its queryset on the URL id alone:

```python
execution_id = self.kwargs.get("pk")                      # views.py:23
return FileExecution.objects.filter(
    workflow_execution_id=execution_id                    # views.py:34
).annotate(latest_log_data=Subquery(latest_log_subquery))
```

No `for_user` anywhere. Any authenticated org member holding an execution id gets file names, `file_path`, sizes, statuses, `execution_error`, and — via `get_status_msg` (`file_execution/serializers.py:36,40`) — the latest `ExecutionLog.data["log"]` string for that run. That is the same data class this PR is protecting, leaking through the sibling endpoint at the same URL depth.

Note the resulting inconsistency: the detail route `/execution/<id>/` *is* gated (`workflow_manager/execution/views/execution.py:28-30`), so an unauthorized caller now gets **403 on detail and 200 on files**.

**Suggested fix:** the same one-line gate in `FileCentricExecutionViewSet.get_queryset`. Better still, factor the check into a small mixin shared with `WorkflowExecutionLogViewSet`, so the next `execution/<pk>/<thing>/` route inherits it instead of re-deriving it.

Pre-existing rather than introduced here — but it defeats the PR's own stated goal, so it belongs in this PR or an explicitly linked follow-up. Confidence: High.

#### [High] [Lens 4 — Security, Lens 2 — Precedent] — `GET /workflow/<workflow_id>/execution/` has the identical dead-`IsOwner` defect, left in place

**Would anchor to:** `backend/workflow_manager/workflow_v2/execution_view.py:13-26`

```python
permission_classes = [IsOwner]                                    # :15
...
queryset = WorkflowExecution.objects.filter(workflow_id=workflow_id)   # :22-24
```

No `for_user`. `IsOwner` implements only `has_object_permission` (`permissions/permission.py:113`), which DRF never invokes on `list` — verbatim the reasoning this PR gives for the log viewset. Any authenticated org member can enumerate every execution of any workflow id in the org (ids, statuses, timings, `execution_error`), including workflows never shared with them.

The route is live: `workflow_v2/urls/workflow.py:79-83` → `workflow_manager/urls.py:20` → `backend/urls.py:34`. `DEFAULT_PERMISSION_CLASSES` is `[]` (`settings/base.py:643`), so `[IsOwner]` is the entire gate. `OrganizationFilterBackend` still applies here (this viewset does not override `filter_backends`), so it is org-bounded but not user-bounded.

Secondary: the class also omits `IsAuthenticated` — not anonymously exploitable only because auth middleware covers it, which leaves it one settings change from being open.

**Suggested fix:** `WorkflowExecution.objects.for_user(self.request.user).filter(workflow_id=workflow_id)`, plus `IsAuthenticated` in `permission_classes`.

Separately and out of scope: `retrieve` on this viewset filters `workflow_id=pk` while `pk` is an execution id, so `execution/<uuid:pk>/` can never resolve — likely a dead route worth a look.

Pre-existing; flagged because leaving it makes this cleanup partial and preserves the exact anti-pattern the diff is retiring. Confidence: High.

---

### Lens checklist (16/16)

| # | Lens | Result |
|---|---|---|
| 1 | Spec & intent | See unanchored finding 1 |
| 2 | Architectural fit & precedent | See `execution_log_view.py:32`, unanchored finding 2 — the delegation to each resource's own `for_user` is the right consolidation; the objection is to what was left behind |
| 3 | Correctness & edge cases | See Low findings (follow-up comment) |
| 4 | Security | See both unanchored findings, `execution_log_view.py:32` |
| 5 | Data integrity & migrations | N/A — no migration, schema, or persisted-field change |
| 6 | Concurrency | N/A — no locks, threads, async ordering or retries in the diff |
| 7 | API & contract compatibility | Clean — the logs endpoint narrows 200→403 for unauthorized callers and the PR body declares it; no wire-format or field change |
| 8 | Reliability & resilience | N/A — no external calls, timeouts, retries or unbounded buffers in the diff |
| 9 | Performance & cost | See `models/execution.py:69-75` |
| 10 | Observability | Clean — the denial is logged via `ExceptionLoggingMiddleware.format_exc_and_log` (`middleware/exception.py:63-66`), so the new gate is not a blind spot |
| 11 | Operational safety | N/A — no IaC, Helm, CI or feature-flag change; rollback is a revert |
| 12 | LLM/agent | N/A — no model, prompt, tool or eval code touched |
| 13 | Testing | See `test_shared_execution_access.py:42-48`, `:50-58`, `:73-80`, `models/execution.py:66-68`, plus Lows |
| 14 | Dependencies & build | N/A — no dependency, lockfile or build change |
| 15 | Code quality | Clean |
| 16 | Doc & comment accuracy | See `models/execution.py:37-38` — the two comments this diff *adds* are both accurate; the defect is a pre-existing line inside the edited block |

---

### Notes

- **Nothing was weakened to green CI.** No existing test file is touched; no assertion removed or relaxed. The `IsOwner` removal is a genuine correctness fix, not test-silencing.
- **I did not run the test suite.** The 6/6 claim is unverified by me; none of the findings above depend on a test failing.
- **CodeRabbit's one finding** (unknown-id denial test) was addressed by the author in `ef3c01d`. Nothing to promote.
- **No cloud-side override** of these viewsets or managers exists — checked `unstract-cloud`.

### Open questions

1. `/execution/<id>/files/` — fix here, or land as a linked follow-up? Either is fine, but the PR body's security claim should be softened if it is deferred.
2. `WorkflowExecutionViewSet` — same question. The commit message diagnoses the pattern precisely; was the sibling viewset checked and consciously deferred?
3. Was the missing `OWNER` membership row in `_api_deployment` deliberate, or an oversight? It changes what four of the six tests actually prove.

backend/workflow_manager/execution/tests/test_shared_execution_access.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def _api_deployment(self, *, shared_to_org: bool = False) -> APIDeployment:
+        # api_name must be short — it defaults to a UUID longer than the column.
+        return APIDeployment.objects.create(
+            api_name=f"api-{secrets.token_hex(4)}",
+            workflow=self.workflow,
+            organization=self.org,
+            created_by=self.owner,
+            shared_to_org=shared_to_org,
+        )
```
chandrasekharan-zipstack: **[High] [Lens 13 — Testing] — The fixture builds an owner-less deployment, so the negative test asserts nothing**

Under UN-2202 `created_by` is audit-only; creator access flows through an `OWNER` membership row. The base class does exactly that for `self.workflow` (`tenant_account_v2/tests.py:98-99`), but `_api_deployment` creates no membership row — and no `post_save` signal supplies one (`tenant_account_v2/signals.py` has only `post_delete` receivers; no `ResourceMembership.objects.create` exists outside tests and migrations).

So every deployment built here is visible to nobody, including `self.owner`. Two consequences:

1. `test_unshared_deployment_stays_invisible` (:95-97) would still pass if `Q(pk__in=member_ids)` were deleted outright from `api_v2/models.py:52`. It reads as a negative control but constrains nothing.
2. The likeliest regression from this refactor — *the owner can no longer see their own executions* — has no test. The PR body claims coverage of "owner, co-owner, direct share, group share and `shared_to_org`"; only the last two are actually asserted.

**Suggested fix:** add the `OWNER` membership row in `_api_deployment`, then add `test_owner_sees_own_executions`, a direct-viewer test via the base module's currently-unused `_add_viewers` (`tenant_account_v2/tests.py:61-67`), and a co-owner test. Re-check that `test_unshared_deployment_stays_invisible` still passes once the deployment is genuinely owned — that version is the one that tests something.

Confidence: High.

backend/workflow_manager/execution/tests/test_shared_execution_access.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def _execution(self, deployment: APIDeployment) -> WorkflowExecution:
+        return WorkflowExecution.objects.create(
+            workflow=self.workflow,
+            pipeline_id=deployment.id,
+            execution_mode=WorkflowExecution.Mode.INSTANT,
+            execution_method=WorkflowExecution.Method.DIRECT,
+            execution_type=WorkflowExecution.Type.COMPLETE,
+            status=ExecutionStatus.COMPLETED,
+        )
```
chandrasekharan-zipstack: **[Medium] [Lens 13 — Testing] — Two of the three ORed branches in the rewritten `for_user` are never satisfied**

`_execution()` hardcodes `pipeline_id=deployment.id`, and all six tests build their execution through it. Against `models/execution.py:83`:

```python
final_filter = (workflow_filter & Q(pipeline_id__isnull=True)) | deployment_filter
```

- `workflow_filter & Q(pipeline_id__isnull=True)` can never be true here, so the swap to `Workflow.objects.for_user` at `models/execution.py:69` is **completely unexercised** — despite workflow-level executions being a primary flow, and despite being exactly the case in row 3 of the PR's manual test table.
- `pipeline_filter` (`models/execution.py:75`) is never satisfied — ETL/TASK pipeline executions have no coverage in either direction.

A regression hiding workflow-level executions from their owner, or leaking pipeline executions org-wide, ships green.

**Suggested fix:** parameterise `_execution(deployment=None)` so `pipeline_id` can be `None`; add (a) a group-shared `self.workflow` case asserting the member sees the workflow-level execution and the outsider does not, and (b) a `Pipeline`-backed twin of the existing grant/deny pair.

Confidence: High.

backend/workflow_manager/execution/tests/test_shared_execution_access.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def _log_queryset(self, user, execution_id):
+        """Run the log viewset's queryset build for ``user`` — the access gate."""
+        request = APIRequestFactory().get("/")
+        request.user = user
+        view = WorkflowExecutionLogViewSet()
+        view.request = request
+        view.kwargs = {"pk": str(execution_id)}
+        return view.get_queryset()
```
chandrasekharan-zipstack, line 105: **[High] [Lens 13 — Testing] — No test reaches HTTP; the 403 and the entire `export` action are unverified**

`_log_queryset` hand-constructs the viewset and calls `get_queryset()` directly, bypassing `as_view()`, `initialize_request`, `initial()`, `check_permissions()`, `dispatch()` and the DRF exception handler. Consequences:

- The PR's headline claim — `GET /execution/<id>/logs/` returns **403** — has no automated assertion anywhere. The test proves `PermissionDenied` is *raised*, not that the endpoint *returns 403*; that translation is entirely a `dispatch()`-level behaviour the test never enters.
- The `permission_classes` change from `[IsAuthenticated, IsOwner]` to `[IsAuthenticated]` is unpinned.
- **`export` has zero coverage.** It inherits the gate only incidentally, via `self.filter_queryset(self.get_queryset())` at `execution_log_view.py:73`, and nothing pins that coupling. A later streaming fast path, a `.values()` rewrite, an async hand-off, or hoisting the row-cap above the queryset build silently reopens bulk CSV/JSON exfiltration of any execution's logs by id — with CI green. This is the higher-impact of the two endpoints and the one with no test.

A future `try/except` around `get_queryset()` (a common pagination-hardening change) would downgrade the gate to an empty 200, and every test here still passes.

**Suggested fix:** route through `as_view({"get": "list"})` and `as_view({"get": "export"})` with `force_authenticate`, asserting 403 for the denied user and 200 for the shared user. The base module already establishes this idiom at `tenant_account_v2/tests.py:344-348`. Keep the existing unit-level tests as fast checks — they are fine, just not sufficient on their own for an authz change.

Confidence: High.

backend/workflow_manager/workflow_v2/execution_log_view.py
```diff
@@ -28,7 +29,7 @@
 
 class WorkflowExecutionLogViewSet(viewsets.ModelViewSet):
     versioning_class = URLPathVersioning
-    permission_classes = [IsAuthenticated, IsOwner]
+    permission_classes = [IsAuthenticated]
```
chandrasekharan-zipstack, line 33: **[Medium] [Lens 2 — Architectural fit, Lens 4 — Security] — Dropping `IsOwner` leaves a `ModelViewSet` whose write handlers carry only `IsAuthenticated`**

Removing `IsOwner` is correct — it implements only `has_object_permission`, which DRF never invokes on `list`/`export`, exactly as the commit message says. But the class is still a `ModelViewSet` (`:30`), so it retains `create` / `update` / `partial_update` / `destroy`.

Nothing is exploitable today because only `list` and `export` are routed (`workflow_manager/execution/urls.py:15-16`, `workflow_v2/urls/workflow.py:23-24`). The issue is the pattern: `create` never calls `get_queryset()`, so the new gate does not cover it, and the object-level class that used to sit in `permission_classes` is gone. "Drop the permission class, gate the queryset instead" is only safe on a read-only viewset, and nothing in the code says so — the next engineer who copies this onto a writable viewset inherits an unguarded write path.

**Suggested fix:** change the base to `viewsets.ReadOnlyModelViewSet`. One word, removes the latent write surface, and makes the queryset-gating pattern self-evidently sufficient. `ExecutionLog` is written only by workers and its fields are `editable=False`, so nothing legitimate is lost.

Confidence: High.

backend/workflow_manager/workflow_v2/models/execution.py
```diff
@@ -35,8 +34,8 @@ def for_user(self, user) -> QuerySet:
         """Filter user's workflow executions with proper access control.
 
         Returns executions where the user has access to:
-        - The workflow (created by user OR shared with user) AND/OR
-        - The pipeline/API deployment (created by user OR shared with user)
+        - The workflow AND/OR
+        - The pipeline/API deployment
```
chandrasekharan-zipstack, line 38: **[Medium] [Lens 16 — Doc accuracy] — Docstring in this block claims org scoping comes from the view; it now comes only from this manager**

Line 46 of this same docstring reads:

> Service accounts see all executions (org-scoped by view).

That is false, and this diff makes it more dangerous:

- The service-account branch org-scopes **itself** at `:54-58`; the view does not do it.
- The only caller, `ExecutionViewSet`, sets `filter_backends = [DjangoFilterBackend, DeterministicOrderingFilter]` (`workflow_manager/execution/views/execution.py:37`), which *replaces* `DEFAULT_FILTER_BACKENDS` and therefore drops `OrganizationFilterBackend` — whose own docstring states "Viewsets MUST NOT override filter_backends" (`utils/filters/organization_filter.py:31-33`).
- This diff relocates non-admin tenant isolation into the three delegated managers (see the new comment at `:66-68`), making `for_user` the **sole** tenant boundary for `/execution/`.

A maintainer trusting line 46 could remove the in-manager org filter and open a cross-tenant leak.

**Suggested fix:** rewrite line 46 to something like "Service accounts and org admins see all executions in the current organization; org scoping is enforced here, not by the view." Optionally, append to rather than replace `filter_backends` on `ExecutionViewSet` to restore the defence-in-depth.

For the record, the two comments this diff *adds* (`:66-68` here and `execution_log_view.py:42-44`) are both accurate — I checked each claim against the code.

Confidence: High.

backend/workflow_manager/workflow_v2/models/execution.py
```diff
@@ -64,23 +63,16 @@ def for_user(self, user) -> QuerySet:
...
-        # Filter for workflow access (owner or direct viewer via membership).
-        # ``created_by`` is audit-only (UN-2202); VIEWER rows replaced shared_users.
-        # ``object_id`` is varchar, so resolve the ids via the cast helper rather
-        # than a ``memberships`` JOIN (Postgres refuses ``uuid = varchar``).
-        workflow_filter = Q(
-            workflow_id__in=resources_visible_via_memberships(Workflow, user)
-        )
+        # Defer to each resource's own ``for_user`` so execution visibility matches
+        # the resource list: memberships, group shares and ``shared_to_org``
+        # (UN-2651). Those managers are org-scoped, so no explicit org arg.
```
chandrasekharan-zipstack: **[Medium] [Lens 13 — Testing] — No cross-organization test, on the change that makes `shared_to_org` load-bearing for execution visibility**

This comment's claim — "Those managers are org-scoped, so no explicit org arg" — is correct today (`DefaultOrganizationManagerMixin.get_queryset` filters `organization=UserContext.get_organization()`, `utils/models/organization_mixin.py:27-30`). But nothing in the suite pins it.

`WorkflowExecutionManager` extends `BaseModelManager` (`:30`) and is **not** org-scoped. After this change, the only thing stopping `Q(shared_to_org=True)` inside `APIDeployment.objects.for_user` (`api_v2/models.py:53`) from matching another tenant's deployments is that mixin reading a thread-local. The suite runs with exactly one `Organization`, set once in `GroupSharingTestBase.setUp`. Dropping the mixin from any of the three sub-managers would pass every test in the new file.

This PR is precisely what makes `shared_to_org` load-bearing for execution visibility, so it is the right PR to pin it.

**Suggested fix:** one test — create a second `Organization` with a `shared_to_org=True` `APIDeployment` and an execution, assert `self.outsider` (org A) neither sees it via `for_user` nor can read its logs. The two-org fixture pattern already exists at `tenant_account_v2/tests.py:207-216`.

Confidence: Medium.

backend/workflow_manager/workflow_v2/models/execution.py
```diff
@@ -64,23 +63,16 @@ def for_user(self, user) -> QuerySet:
...
-        api_filter = Q(
-            pipeline_id__in=resources_visible_via_memberships(APIDeployment, user)
-        )
+        api_filter = Q(pipeline_id__in=APIDeployment.objects.for_user(user).values("pk"))
 
         # Filter for Pipelines the user can access
-        pipeline_filter = Q(
-            pipeline_id__in=resources_visible_via_memberships(Pipeline, user)
-        )
+        pipeline_filter = Q(pipeline_id__in=Pipeline.objects.for_user(user).values("pk"))
```
chandrasekharan-zipstack, line 71: **[Medium] [Lens 9 — Performance] — `for_user` now costs three extra uncached admin lookups per call, on a polled path**

`for_user` resolves `is_user_organization_admin(user)` at `:60`, then delegates to `Workflow.objects.for_user`, `APIDeployment.objects.for_user` and `Pipeline.objects.for_user` — **each of which re-resolves the same predicate** as its second statement.

`OrganizationMemberService.is_user_organization_admin` is uncached (`tenant_account_v2/organization_member_service.py:31-45`): one `OrganizationMember.objects.get()` round-trip plus an `AuthenticationController()` construction per call. Net **+3 queries per `for_user` call** for every non-admin user.

Where that lands:
- `/execution/` (paginated list).
- `/execution/<id>/logs/` — which the Logs modal polls on a refresh interval, and which this PR newly routes through `for_user`.
- `ExecutionViewSet` calls `for_user` **twice** on retrieve (`workflow_manager/execution/views/execution.py:30` and `:44`), so that path pays it six times.

Separately, each subquery went from a single indexed lookup on `resource_membership` to a `SELECT DISTINCT` over the resource table wrapping 2–3 nested `IN`s. The subquery-not-materialised property that UN-2202 (#1797) deliberately tuned for this exact call site is preserved, so plan caching still holds — the added cost is depth, not an N+1.

**Suggested fix:** memoize the admin predicate. `permissions/permission.py:57-67` already has exactly this cache, but keyed on `request`, which the managers never see — simplest is a per-user memo inside `OrganizationMemberService.is_user_organization_admin`.

Confidence: High on the mechanism, Medium on impact — an `EXPLAIN` on a realistic org plus the modal's poll interval would settle whether this is Medium or Low.

chandrasekharan-zipstack:
### Standardized review — the 3 Low findings

Split out from [the review above](https://github.com/Zipstack/unstract/pull/2234#pullrequestreview-4907303994) so they don't compete with the High/Medium ones. None of these block anything.

---

**[Low] [Lens 3 — Correctness] — `for_user` now returns empty rather than membership-scoped results outside request context**

`backend/workflow_manager/workflow_v2/models/execution.py:66-78`

The removed helper documented an explicit `organization` argument for exactly this case:

> Falls back to `UserContext` so request paths need no change; pass `organization` explicitly on worker/management paths where `UserContext` is empty.
> — `tenant_account_v2/sharing_helpers.py:208-213`

The replacements go through `DefaultOrganizationManagerMixin.get_queryset`, which does `filter(organization=UserContext.get_organization())` (`utils/models/organization_mixin.py:27-30`). With no org context that becomes `organization_id IS NULL`, so a non-admin `for_user` call from Celery or a management command silently returns nothing.

That is **fail-closed**, so it is not a security issue, and I confirmed every current caller is a request-context viewset (`workflow_manager/execution/views/execution.py:30,44` and `execution_log_view.py:46`). But the escape hatch the old helper deliberately provided is gone, and the new comment at `:68` — "Those managers are org-scoped, so no explicit org arg" — reads as unconditional when it really means "correct inside a request".

*Suggested fix:* narrow the comment to something like "org-scoped via `UserContext`; request paths only", so a future worker-side caller is warned rather than surprised.

Confidence: Medium — would rise to High if a worker or management path is later shown to call `WorkflowExecution.objects.for_user`.

---

**[Low] [Lens 13 — Testing] — `test_logs_denied_when_the_execution_is_unknown` cannot fail independently**

`backend/workflow_manager/execution/tests/test_shared_execution_access.py:104-108`

The property being claimed is that an unknown id and an inaccessible id are indistinguishable *to an attacker probing for valid ids*. The test asserts only that `PermissionDenied` is raised — from the single `raise` at `execution_log_view.py:50` that both cases reach identically. It is true by construction and can only fail in lockstep with its sibling at `:99-102`. It never compares status code, body, or headers, which are what an enumerator actually observes.

There is also a real distinction at the HTTP layer that this test structurally cannot see: `execution/urls.py:22` uses `<uuid:pk>`, so a *malformed* id 404s from URL resolution while an unknown-but-valid UUID 403s.

*Suggested fix:* drive both cases through `as_view({"get": "list"})` and assert the two responses have equal `status_code` and equal `.data`. That is the actual property, and it can then fail on its own.

Confidence: High.

---

**[Low] [Lens 13 — Testing] — Randomised fixture value makes a failure non-reproducible from CI output**

`backend/workflow_manager/execution/tests/test_shared_execution_access.py:43`

```python
api_name=f"api-{secrets.token_hex(4)}"
```

Only uniqueness is required, and per-test rollback makes collisions a non-issue — so this is not a flake source. But nothing seeds or logs the value, so a failure whose cause depended on the name could not be reproduced from the CI output.

*Suggested fix:* derive the name from the test method name or a per-instance counter. Equivalent uniqueness, reproducible.

Confidence: High.

backend/workflow_manager/execution/tests/test_shared_execution_access.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Valid, and the sharper half is the second consequence — `_api_deployment` created no OWNER row, so the deployment was visible to nobody and the negative test would have passed with `Q(pk__in=member_ids)` deleted outright.

Fixed in 8e2649ca3: the fixture now creates the OWNER row that `APIDeploymentViewSet.perform_create` creates, `test_unshared_deployment_is_visible_only_to_its_owner` asserts both directions in one place, and direct-viewer (via the base module's `_add_viewers`) and co-owner tests are added. Checked it is no longer vacuous by deleting `Q(pk__in=member_ids)` from `api_v2/models.py:54` — 4 tests fail now, 0 did before.

backend/workflow_manager/execution/tests/test_shared_execution_access.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Valid — every execution was built with `pipeline_id` set, so neither `workflow_filter & Q(pipeline_id__isnull=True)` nor `pipeline_filter` could ever match.

Fixed in 8e2649ca3: `_execution(resource=None)` now builds a workflow-level execution, and `test_workflow_level_execution_follows_the_workflow_share` plus `test_pipeline_execution_follows_the_pipeline_share` cover both previously-dead branches in the grant and deny direction.

backend/workflow_manager/execution/tests/test_shared_execution_access.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Valid — the PR headline claimed a 403 that nothing asserted, and `export` had no coverage at all.

Fixed in 8e2649ca3: `_call()` drives `as_view({"get": action})` with `force_authenticate`, and `test_both_endpoints_return_403_for_an_inaccessible_execution` / `test_both_endpoints_serve_a_shared_execution` cover list and export in both directions. Your `try/except` scenario is now caught — replacing the `raise PermissionDenied` with `pass` fails 4 tests, including both export cases. Kept the unit-level checks as you suggested.

backend/workflow_manager/workflow_v2/execution_log_view.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Valid. Fixed in 8e2649ca3 — now `viewsets.ReadOnlyModelViewSet`, with a comment at `execution_log_view.py:31-33` recording why queryset-gating is sufficient here so the pattern is not copied onto a writable viewset. `export` is unaffected: it is a plain method reached through `as_view({"get": "export"})`, not a router action, and the new export tests confirm it still serves 200/403.

backend/workflow_manager/workflow_v2/models/execution.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Valid — the branch org-scopes itself at `:54-58`, and `ExecutionViewSet` replaces `filter_backends`, so the view contributes nothing.

Fixed in 8e2649ca3: the docstring now states the scoping is enforced in the manager and names the dropped `OrganizationFilterBackend`. Left the optional half alone — restoring the backend on `ExecutionViewSet` changes filtering on the main list endpoint and is wider than this PR; the new cross-org test pins the boundary directly instead. Thanks for checking the two added comments.

backend/workflow_manager/workflow_v2/models/execution.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Valid — this PR is what makes `shared_to_org` load-bearing for execution visibility, so it should pin it.

Fixed in 8e2649ca3: `test_org_wide_share_does_not_cross_organizations` builds a second org with a `shared_to_org=True` deployment and its execution, then asserts `self.outsider` neither sees it via `for_user` nor can read its logs. Dropping `DefaultOrganizationManagerMixin` from any of the three sub-managers now fails it.

backend/workflow_manager/workflow_v2/models/execution.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Valid, and confirmed against `origin/main`: the old body resolved the predicate once and then used `resources_visible_via_memberships` directly, so the delegation this PR adds is what turns one lookup into four.

Fixed in 8e2649ca3 by memoizing inside `OrganizationMemberService.is_user_organization_admin`, keyed on the `User` instance — Django rebuilds that per request, so the memo has the same lifetime as the `request`-keyed cache at `permissions/permission.py:57-67` that the managers cannot reach. The failure paths stay uncached deliberately, so a transient DB error cannot pin a user to non-admin for the rest of the request.

kirtimanmishrazipstack:
@chandrasekharan-zipstack On the three Lows — two fixed in 8e2649ca3, one I think is not worth acting on.

**L1 (comment reads unconditional)** — agreed, fixed. `models/execution.py:66-70` now reads "org-scope themselves via `UserContext`, so this is correct on request paths only — a worker or management command with no org context gets an empty queryset (fail-closed)."

**L2 (unknown-id test cannot fail independently)** — agreed, and your framing is the right one: the property is about what an enumerator observes. Replaced it with `test_unknown_execution_is_indistinguishable_from_an_inaccessible_one`, which drives both ids through `as_view({"get": "list"})` and asserts equal `status_code` **and** equal `.data`. It fails on its own now.

**L3 (randomised `api_name`)** — not taking this one. `API_NAME_MAX_LENGTH` is 30 and the method names here run past 50, so deriving the name from `_testMethodName` overflows the column. More fundamentally, no assertion reads `api_name`, so a failure cannot depend on its value — and `secrets.token_hex` is already the module’s idiom for throwaway uniqueness (`tenant_account_v2/tests.py:44`). Happy to revisit if you have a case where the name reaches an assertion.

kirtimanmishrazipstack:
## Self-review — high + medium findings

Ran a multi-pass review over this branch (correctness, tests, error handling, comment accuracy). Every claim below was verified against the code. Lows are listed at the bottom without detail.

---

### HIGH

**H1 — The fix is incomplete: `/execution/<id>/files/` takes the same id and is still ungated within the org**

`backend/workflow_manager/file_execution/views.py:14-35`

```python
class FileCentricExecutionViewSet(viewsets.ReadOnlyModelViewSet):
    permission_classes = [IsAuthenticated]
    def get_queryset(self):
        return FileExecution.objects.filter(workflow_execution_id=execution_id)...
```

Same `<uuid:pk>` — a `WorkflowExecution` id — with no `for_user` check. `OrganizationFilterBackend` still applies (this viewset does not override `filter_backends`), so cross-org is blocked, but nothing scopes it *within* the org.

The payload is not metadata-only. `FileCentricExecutionSerializer` uses `exclude = ["file_hash"]`, so the response carries `file_name`, `file_path`, `execution_error`, `provider_file_uuid`, `fs_metadata` — plus `status_msg`, which `get_status_msg` resolves from the latest non-DEBUG/WARN `ExecutionLog.data["log"]`. That is log text.

Both endpoints back the same screen: `DetailedLogs.jsx:177` fetches `/execution/${id}/files/` and `LogModal.jsx:47` fetches `/execution/${executionId}/logs/`, for the same id. After this PR a de-shared user gets 403 on one and filenames + per-file error text + a log line on the other.

Pre-existing, but same threat model, same id space, same page — and this PR's own rationale ("unknown ids are denied like inaccessible ones so the response does not reveal which ids exist") applies verbatim. Suggest extracting the gate into a shared `assert_execution_accessible(user, execution_id)` helper and calling it from both `get_queryset`s so the two cannot drift again.

---

**H2 — `execution.py:69-73`: the "empty queryset (fail-closed)" claim is false**

`DefaultOrganizationManagerMixin.get_queryset` (`backend/utils/models/organization_mixin.py:27-30`) is:

```python
organization = UserContext.get_organization()
return super().get_queryset().filter(organization=organization)
```

With no org context that is `filter(organization=None)`, which Django compiles to `organization_id IS NULL` — orphan rows, not zero rows. `organization` is `null=True` on the mixin, and `DefaultOrganizationMixin.save()` assigns it from `UserContext`, so rows created off the request path are *born* with NULL org — exactly what this branch would return.

The genuine fail-closed mechanism is `OrganizationFilterBackend` (`utils/filters/organization_filter.py:52-60`, explicit `.none()` plus a warning) — which `ExecutionViewSet` drops by overriding `filter_backends`. Either reword the comment to what the code does, or make the mixin `.none()` on a missing org.

---

**H3 — `execution_log_view.py:31-33`: the read-only rationale is inverted, and the `editable=False` half is wrong**

> Read-only on purpose: the access gate lives in `get_queryset`, which write handlers never call. `ExecutionLog` rows are written by workers and their fields are `editable=False`, so there is nothing to expose.

- DRF's `update` / `partial_update` / `destroy` all call `get_object()` → `filter_queryset(self.get_queryset())`. They are precisely the handlers that *do* run the gate. `create` is the only one that skips it.
- Only `id`, `execution_id`, `wf_execution`, `file_execution` carry `editable=False`. `data` (JSONField) and `event_time` (DateTimeField) do not, and the serializer is `fields = "__all__"` (`workflow_v2/serializers.py:154-157`), so DRF would expose both as writable on a PATCH.
- Rows are written by the backend Celery task `execution_log_utils.consume_log_history`, not by the `workers/` processes.

The decision to go read-only is right. The stated reason is the inverse of the mechanism, and would justify someone later adding a write handler on the belief it bypasses the gate. Worth also noting that the reason no write verb is reachable *today* is that both URLconfs bind GET only — not anything in this class.

---

**H4 — Test gap: the `pipeline_id__isnull=True` conjunct is unpinned**

`execution.py:88`:

```python
final_filter = (workflow_filter & Q(pipeline_id__isnull=True)) | deployment_filter
```

Deleting `& Q(pipeline_id__isnull=True)` breaks **no test**. Every fixture deployment hangs off `self.workflow` (test file lines 48, 61), and `self.workflow` is reachable only by `self.owner` — so the `assertFalse` in `test_group_share_exposes_the_deployment_executions` and `test_pipeline_execution_follows_the_pipeline_share` holds with or without the conjunct. The mutation survives.

That conjunct is what stops workflow access from leaking into runs of deployments deliberately *not* shared — a privilege widening on the endpoint this PR is hardening. It is also the behaviour the PR description calls out as counter-intuitive ("every path has to be revoked"), so it is the line most likely to be "simplified" later.

```python
def test_workflow_share_does_not_expose_unshared_deployment_runs(self) -> None:
    deployment = self._api_deployment()            # shared with nobody
    execution = self._execution(deployment)
    self._share_with_group(self.workflow)          # member now sees the workflow
    self.assertIn(self.workflow, Workflow.objects.for_user(self.member))  # control
    self.assertFalse(self._visible_to(self.member, execution))

def test_deployment_share_does_not_expose_workflow_level_runs(self) -> None:
    deployment = self._api_deployment(shared_to_org=True)
    self.assertTrue(self._visible_to(self.outsider, self._execution(deployment)))  # control
    self.assertFalse(self._visible_to(self.outsider, self._execution()))
```

---

**H5 — The admin memo is not keyed by organization, but the predicate it caches is org-dependent**

`tenant_account_v2/organization_member_service.py:10-14, 41-44`

`OrganizationMember.objects` is `OrganizationMemberModelManager(DefaultOrganizationManagerMixin, ...)` (`tenant_account_v2/models.py:12`), so `OrganizationMember.objects.get(user=user.id)` resolves against `UserContext`. The answer is a function of `(user, current org)`; the memo stores only the bool.

The justifying comment — *"Django rebuilds that instance per request, so the memo lives exactly one request"* — asserts an invariant nothing enforces. `authentication_controller.set_user_organization:178` flips `UserContext` mid-request while holding the same `request.user`, and the predicate is a public static called with non-request users (`prompt_studio_helper.py:241` passes `profile_manager.created_by`; `tool_instance_helper.py:525` takes an arbitrary user).

I traced the mid-request switch: that view is a bare `@api_view` with no permission class that would seed the memo beforehand, so **there is no live bug today**. But a stale `True` in `for_user` means "see every resource in the org" — the widest possible answer — and the guard is currently a comment rather than code. One-line fix:

```python
org_id = UserContext.get_organization_identifier()
memo = getattr(user, _ADMIN_MEMO_ATTR, None)
if memo is not None and memo[0] == org_id:
    return memo[1]
...
setattr(user, _ADMIN_MEMO_ATTR, (org_id, is_admin))
```

Separately worth asking whether the memo belongs in this PR. Measured saving is exactly 3 single-row indexed lookups per `for_user` call, non-admins only (admins short-circuit before the three delegates). That is 22 lines and a new staleness mode, in a module with 20+ call sites, riding along in an access-control fix. Hoisting the predicate to a local in `for_user` and passing it down saves the same 3 queries with no lifetime question at all.

---

### MEDIUM

**M1 — `organization_member_service.py:62-65`: `except AttributeError: pass` guards an unreachable case and silently disables the memo**

`AnonymousUser` cannot reach that line (line 38 returns first — `is_authenticated` is `False`), and it is not immutable anyway (plain `__dict__`, `setattr` succeeds). No user type in either repo raises here. If a future cloud proxy ever wraps `User` with a custom `__setattr__`, this `pass` silently restores the 4x query amplification the hunk exists to remove, with zero diagnostic signal. Drop the try/except, or log a warning instead of passing. It is also the only bare `except: pass` in the diff.

**M2 — The org-admin and service-account branches have zero test coverage on a cross-tenant boundary**

`execution.py:57-61` and `:63-67` — 4 lines each, **0 hits** in the integration coverage report for this branch's own CI run. The `setUp` patch (test file lines 40-42) pins `is_user_organization_admin` to `False` for the whole class, so neither branch ever executes.

These branches are the *only* tenant boundary for admins on `/execution/`, per this PR's own docstring. `GroupSharingTestBase` already creates `self.admin` with role `"admin"` and it is unused in this file. A sibling class without the patch closes both. Note the patch buys nothing for the current assertions either — every fixture user has role `"user"`, so real admin resolution already returns `False`.

**M3 — `execution.py:61, 67`: `return self.all()` is a cross-org fallback under a docstring that says it cannot happen**

Pre-existing and unchanged by this PR. But this PR's new docstring promotes this manager to *the* tenant boundary ("That org scoping is enforced here, not by the view"), and these two lines return every execution in every organization when `UserContext.get_organization()` is `None` — which `UserContext` returns on three silent paths (no StateStore key, `Organization.DoesNotExist`, `ProgrammingError`), none of them logged.

I confirmed all three non-test callers of `WorkflowExecution.objects.for_user` are request-path, so `self.none()` would break nothing. Either make it `self.none()` (2 lines, arguably in scope since the PR is claiming the guarantee) or soften the docstring so the claim does not outrun the code.

**M4 — `execution_log_view.py:57-59`: the backward-compat `Q(execution_id=...)` branch is dead in request context**

`ExecutionLog.objects` is an `OrgAwareManager` (`models/execution_log.py:42`) that BFS-resolves the org path to `wf_execution__workflow__organization`. That is an INNER JOIN, so any legacy row with `wf_execution IS NULL` — the only rows the second `Q` can match — is dropped before the OR is evaluated.

Pre-existing, but this PR makes it more confusing: the gate now explicitly confirms "you have access to this execution", and the body then returns `{"count": 0, "results": []}` with no error and no log line. Since this method's contract is being rewritten anyway, worth answering the question — delete the branch and the TODO, or make it reachable via an unfiltered manager plus an explicit org check.

**M5 — The denial is not logged with the user id**

`execution_log_view.py:53`. The global handler (`middleware.exception.drf_logging_exc_handler`) does log method, full URI and request id at ERROR, so this is not silent. What is missing is *who* — the field needed to answer "is someone enumerating execution ids?". Since this PR deliberately makes unknown and inaccessible ids indistinguishable to the client (correct), the server is now the only place that can tell them apart, and that distinction is worth recording:

```python
logger.warning(
    "Execution log access denied: user=%s execution=%s org=%s exists=%s",
    getattr(self.request.user, "id", None), execution_id,
    UserContext.get_organization_identifier(),
    WorkflowExecution.objects.filter(pk=execution_id).exists(),
)
```

**M6 — The PR description's test table is inaccurate**

It claims "6 / 6 pass (4 of them fail on `main`)". The file has **13** tests, and 8 fail on `main`. Two listed names do not exist in the file (`test_unshared_deployment_stays_invisible`, `test_logs_denied_when_the_execution_is_unknown`), and seven real tests are missing from the table. Reviewers calibrate on that table. A table mirroring a test file also rots on every edit — one sentence naming the file is more durable.

---

### Lows, not detailed here

`.values("pk")` redundant on `pk__in=<queryset>`; `.distinct()` vestigial (all terms are `IN (subquery)`, no outer join); `_log_queryset` redundant now that `_call` drives the real endpoint, and two of its three uses are strict subsets of the `_call` tests; `_share_with_group` hand-writes the ContentType row instead of using the production `set_resource_share_groups`; JSON export (the default `file_format`) untested; group-share and direct-viewer tested for `APIDeployment` but not `Pipeline`; export 400/413 paths uncovered; cross-org test has no positive control; `UserContext` never torn down in `GroupSharingTestBase`; comment cites `APIDeploymentViewSet.perform_create`, which does not exist (the OWNER row is created in `create()`); class docstring says `self.owner` "owns every fixture here", not true of the cross-org fixtures; six intermediate `Q` locals in `for_user` collapse to one expression.

---

### Holding up well

The gate itself is real and pinned — delete the `raise PermissionDenied` block and four tests fail. Dropping `IsOwner` genuinely widens nothing: it implements only `has_object_permission`, and all four routes bind `list`/`export`, neither of which calls `get_object()`. `test_unknown_execution_is_indistinguishable_from_an_inaccessible_one` asserting on the *response* (status and body) rather than the exception type is the right call — the usual `assertRaises` version ships an enumeration oracle. And `test_unshared_deployment_is_visible_only_to_its_owner` asserting its own positive control is the discipline H4 and the cross-org test are missing.

kirtimanmishrazipstack:
Addressed in `92a06516`. All five Highs and five of six Mediums fixed; M4 got a comment rather than a deletion. Also fixed @chandrasekharan-zipstack's second unanchored High from [the earlier review](https://github.com/Zipstack/unstract/pull/2234#pullrequestreview-3758843200), which this self-review did not repeat.

**H1 — `<id>/files/` ungated.** New `workflow_manager/execution/access.py` holds `assert_execution_accessible(user, execution_id)`; `WorkflowExecutionLogViewSet.get_queryset` and `FileCentricExecutionViewSet.get_queryset` both call it. One gate, so the two routes on the same id cannot drift apart again.

**H2 — the fail-closed claim.** Confirmed: `DefaultOrganizationManagerMixin` compiles `filter(organization=None)` to `organization_id IS NULL`, which matches orphan rows rather than nothing. Comment now states that, and points the reader at the request path. I did not change the mixin — that is a repo-wide behavioural change well outside this PR.

**H3 — the read-only rationale.** Both halves confirmed. `update`/`partial_update`/`destroy` do run the gate via `get_object()`; `create` is the single handler that does not. Only `id`, `execution_id`, `wf_execution` and `file_execution` are `editable=False` — `data` and `event_time` are writable under `fields = "__all__"`. Writer is the backend `consume_log_history` Celery task, not `workers/`. Comment rewritten to the actual mechanism.

**H4 — the `pipeline_id__isnull=True` conjunct.** Both suggested tests added verbatim in intent. Verified by mutation: deleting the conjunct now fails `test_workflow_share_does_not_expose_unshared_deployment_runs`.

**H5 — memo staleness.** Memo is now `(organization_identifier, is_admin)` and the comment no longer asserts an instance-lifetime invariant. Kept the memo rather than hoisting the predicate into `for_user`: the three delegates would each need a new parameter in their public signature, which is a wider change than the one it replaces.

**M1** — `except AttributeError: pass` deleted; plain `setattr`.

**M2** — new `ExecutionBypassRoleTests`: org admin sees unshared runs, service account sees them without any membership, both stay inside the current org, and no-org-in-context returns nothing. One correction to the finding, though: leaving the predicate to resolve for real does **not** work. `AuthenticationController` resolves `auth_service` from the active auth plugin, and the auth0 plugin's `UserRole.ADMIN` is `"unstract_admin"` where OSS's is `"admin"` — so `is_admin_by_role("admin")` returns `False` on any checkout carrying the plugin and `True` in OSS CI. A test depending on that is environment-dependent by construction. Patched deterministically instead (only `self.admin`), same idiom as `ShareAuthorizationServiceTests`, which pins what `for_user` does once the predicate is `True` — the role string belongs to the plugin, not to this PR.

**M3** — extracted `_org_scoped()`; `self.none()` plus a warning when `UserContext` has no org. Confirmed all three non-test callers are request-path.

**M4 — not deleting the branch.** The reachability analysis is right — the `OrgAwareManager` join through `wf_execution` drops legacy-only rows before the OR is evaluated — but "delete it" turns on whether rows with `wf_execution IS NULL` still exist in production, which I cannot verify from here, and the branch still functions in the unscoped contexts (Celery, shell) where the org filter is absent. Replaced the stale TODO with the actual reason it matches nothing in request context.

**M5** — the denial now logs `user`, `execution`, `org` and `exists=`. It sits in the shared helper, so both routes get it, and it runs only on the denial path.

**M6** — PR body updated. Dropped the table for one sentence naming the file, as suggested.

Four mutation checks run against the new tests, all caught, files restored: gate removed from `files/` → `test_file_executions_follow_the_same_gate_as_the_logs`; `for_user` removed from `WorkflowExecutionViewSet` → `test_workflow_execution_list_is_scoped_to_accessible_workflows`; `self.none()` → `self.all()` → `test_no_organization_in_context_returns_nothing`; conjunct removed → `test_workflow_share_does_not_expose_unshared_deployment_runs`.

Lows not acted on beyond the three I corrected in passing (the `perform_create` citation, the class docstring's ownership claim, and the missing positive control on the cross-org test).

chandrasekharan-zipstack (approved):
@kirtimanmishrazipstack consider making the code comments concise and generic. These can go stale quickly and also lead to context rot when the agent scans codebases

kirtimanmishrazipstack:
@chandrasekharan-zipstack Fair — done in `c2e28a98` (7 files, 42 insertions / 83 deletions, comments only).

What I cut: line-by-line retellings of the mechanism, `file.py:NN` citations, and "this used to be X" history. Those are the parts that rot on the next refactor and that a whole-file read has to wade through.

What I kept: the reason a thing is the way it is, where it isn't inferable from the code — why the log viewset is read-only, why the compat `Q` term still exists, why the admin memo is keyed by org, and why the bypass tests patch the predicate instead of resolving it. One or two lines each.

PR #2237, opened 2026-08-12
UN-4009 [MISC] Generate and commit the API deployment OpenAPI spec in-repo

chandrasekharan-zipstack:
## What

The OpenAPI description of the API deployment `execute` / `status` endpoints is now generated and committed in this repository, and a test fails when it drifts from the code. `drf-yasg`, which generated nothing anyone read, is removed.

## Why

The published Python client and the SDK generated for it are built from that description. It was produced by a script living outside this repository, so a route, serializer or response change here could silently invalidate it — the breakage would surface later, in a client repo, against a spec nobody in the PR could see.

## How

- **`drf-spectacular`** added as a backend dependency; `DEFAULT_SCHEMA_CLASS` and `SPECTACULAR_SETTINGS` set in `settings/base.py`. Both are read only during schema generation — no request-time behaviour changes. The schema class is `drf_standardized_errors.openapi.AutoSchema`, matching the exception handler these views actually run behind.
- **`api_v2/openapi_schema.py`** holds the schema annotation. The introspected schema is wrong in ways that matter to a generated client: a bare `FileField` maps to `format: uri` (right for output, wrong for a multipart upload), `result: null` while an execution is pending crashes a generated deserialiser without `allow_null`, and `operation_id` / `tags` decide the command names and module paths clients expose. It is a module of its own rather than part of `serializers.py`, because none of these serializers parses a request or builds a response.
- **`api_v2/deployment_spec_urls.py`** — a urlconf mirroring the real mount. Generating against the included sub-urlconf drops the prefix and produces paths the server does not serve. `SPEC_URLCONFS` is a tuple: widening the spec to another endpoint is one entry plus its `@extend_schema`.
- **`manage.py generate_docstudio_spec`** writes `specs/docstudio-oss.json` with sorted keys, so the committed file is a usable drift signal. `--check` fails instead of writing. Generation refuses to publish on any generator diagnostic, on a spec that is not valid OpenAPI, or on paths carrying a non-default `API_DEPLOYMENT_PATH_PREFIX`.
- **`api_v2/tests/test_docstudio_spec.py`** regenerates and compares, and — because drift alone passes on a spec that is uniformly wrong — anchors the upload encoding, the nullable result, the documented file-result fields and the error body to the code that produces them. It runs in the existing `unit-backend` group — no database, no new CI job. Its failure message names the repos that regenerate from the spec, since propagating the change downstream is the part CI cannot do.
- **`drf-yasg` and the `docs` app removed.** The redoc UI they served was built with `public=False` and no endpoint carried a `@swagger_auto_schema`, so an anonymous caller saw an empty schema. Nothing generated or consumed it. `backend/README.md` now points at the committed spec and the regeneration command instead of the deleted `/api/v1/doc/` route.

## Regenerating

```
uv run python manage.py generate_docstudio_spec
```

from `backend/`, then commit. The test says so in its failure message, along with the client and CLI repos that regenerate from the result.

The `/doc/` route is gone with `drf-yasg`, and drf-spectacular serves no UI in its place — the spec is a committed file, not an endpoint. Nothing else changes at request time: the settings added are read only during schema generation, and no route or response is modified. The only serializer change is `help_text` on the two `hitl_*` fields, which is published documentation and does not affect validation. The annotation is metadata on the view, not behaviour.

## Database Migrations

None.

## Env Config

None. `API_DEPLOYMENT_PATH_PREFIX` is read as before; generation now refuses to write a spec when it is overridden away from `deployment`, so a developer's local override cannot reach the committed artifact.

## Relevant Docs

- [`backend/README.md`](backend/README.md) — API Docs section, replaced with the committed spec's location and the regeneration command.

## Related Issues or PRs

- UN-4009 (sub-task of UN-4008)
- Zipstack/unstract-python-client#27 — the client generated from this spec

## Dependencies Versions

| Package | Before | After |
| --- | --- | --- |
| `drf-yasg` | `1.21.15` | removed |
| `drf-spectacular` | — | `0.30.0` (pinned exactly: the generated artifact is committed, so a minor bump would show as spec drift rather than as a dependency change) |

`drf-standardized-errors` is unchanged at `>=0.12.6`; it was already a dependency and is now also the source of the schema class. Transitively, `drf-yasg` required `inflection`, `packaging`, `pytz`, `pyyaml` and `uritemplate`; `drf-spectacular` requires `inflection`, `jsonschema`, `pyyaml` and `uritemplate`. All of them except `jsonschema` were already in the lock through other packages.

## Notes on Testing

`676 passed, 1 skipped` (plus 340 subtests) in the backend unit tier this lands in, including the 18 spec tests. Verified by hand that the two new generation gates fire: a typo in a hand-written schema fragment fails OpenAPI validation, and `API_DEPLOYMENT_PATH_PREFIX=private` fails the prefix check.

The generated spec is byte-identical to what the external script produced, apart from the root `tags` array — clients had nowhere to read group descriptions from, which is the one deliberate addition — and the review fixes in the latest commit (error body shape, nullability, the statuses each operation can actually return).

## Screenshots

n/a — no user-facing surface.

## Checklist

https://claude.ai/code/session_01KjRpEocCvnxGUSkFivgnhk


backend/api_v2/openapi_schema.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+class StatusResponse(serializers.Serializer):
+    status = serializers.CharField()
+    message = FileResult(many=True, required=False, allow_null=True)
+
+
+class ErrorResponse(serializers.Serializer):
+    status = serializers.CharField(required=False)
+    message = serializers.JSONField(required=False, allow_null=True)
```
ritwik-g: **[High] [Lens 7, 3, 2] — Every error response in the spec has a shape the server never sends**

`ErrorResponse` publishes `{status, message}`, but 400/401/403/404/409/429 are all raised as `APIException` and routed through `drf_standardized_errors` (`backend/middleware/exception.py:48`), which emits `{type, errors:[{code, detail, attr}]}` — a **disjoint** shape. A generated SDK models both documented fields, finds them permanently absent on every failure, and has nothing to read a message from.

`{status, message}` fits only the hand-built 406/422/500 bodies (`api_deployment_views.py:196-228`).

**Verified** — the pinned handler run against this repo's own exceptions:
```
401 {'type':'client_error','errors':[{'code':'error','detail':'Unauthorized','attr':None}]}
400 {'type':'validation_error','errors':[{'code':'invalid','detail':'You must provide at least one file or presigned URL.','attr':None}]}
```
Corroborated in-repo by `backend/middleware/test_exception.py:41-43` and `frontend/src/hooks/useExceptionHandler.jsx:35`.

**Fix:** the repo already ships `drf-standardized-errors==0.15.0` with `drf_standardized_errors/openapi.py` for exactly this — use its `AutoSchema` as `DEFAULT_SCHEMA_CLASS`; or minimally redefine `ErrorResponse` to the handler's shape and keep `{status, message}` for 406/422/500 only.

ritwik-g (commented):
## Standardized review — verdict: `REQUEST CHANGES`

**Critical: 0 · High: 2 · Medium: 8 · Low: 7 · Lenses run: 17/17**

Reviewed against a fixed 17-lens rubric (`unstract:standard-review`, plugin v0.30.1) at `51c7cd45`, diffed from merge-base `6b916eac` — 15 files, matching GitHub's count.

The engineering here is careful: the guard against spectacular's guessing, the urlconf that *selects* the real mount rather than restating it, the failure message naming the downstream repos. Two things to say up front:

- The claim that CI fails on drift is **real** — I suspected it wasn't and checked. `tests/groups.yaml:85-94` collects `backend/**` by glob, and this PR's own CI run shows `unit-backend` at 1009 passed.
- A dedicated security pass (the plugin's vendored prompt, separate from the lenses) found **no findings at confidence ≥ 8** — an empty result *within* its scope, not by exclusion. `/doc/` was never in `WHITELISTED_PATHS`, the committed spec has no `servers` block and no secrets, and removing the redoc UI net-reduces attack surface.

Both High findings were confirmed by **running the real serializers against the real DTOs** in a scratch venv, not by reading. Both reach a generated SDK, and this spec is the artifact `Zipstack/unstract-python-client#27` is generated from — so fixing them here will move that PR's generated code.

### Unanchored findings

- **[Medium] [Lens 17] `backend/README.md:178-179` now points at a route this PR deletes.** "access the API documentation that's auto generated at the backend endpoint `/api/v1/doc/`" — that route goes with `drf-yasg` and the `docs` app (`public_urls.py:31`, `public_urls_v2.py:28`). The section now describes an endpoint that 404s, on the page a new backend developer reads first. The only surviving `doc/` route is `admin/doc/` (Django admindocs), a different mount behind `ADMIN_ENABLED`. Suggest replacing the section with the committed spec and the regeneration command. *(Verified directly — one automated sweep reported no surviving references and was wrong.)*
- **PR title** — this repo states no title convention in writing (no `CLAUDE.md`, no `CONTRIBUTING` title section, template links out only), so there is nothing to judge against. `MISC` is the right type regardless: nothing here changes what the product does for a user.
- **PR body drops template sections**, including *Dependencies Versions* — on the one PR that swaps `drf-yasg` for a pinned `drf-spectacular==0.30.0`, that's the section this change most needed.
- **One description claim is unfalsifiable from this repo**: "byte-identical to what the external script produced, apart from the root `tags` array." The external script isn't here; `unstract-python-client#27` is where that gets checked.

### Open questions

1. Is a **cloud** spec generated from this same command? That decides whether the `hitl_*` finding wants a flat `exclude_fields` or a conditional.
2. Was the `{status, message}` `ErrorResponse` shape a deliberate simplification, or written before `drf-standardized-errors` became the handler?

### Assumptions

- `drf-standardized-errors` is the active handler for these endpoints (verified at `backend/middleware/exception.py:48`). If a deployment overrides `EXCEPTION_HANDLER`, the first finding's severity changes.
- `specs/docstudio-oss.json` is the OSS artifact only, per its filename and `DEFAULT_OUT`. If it also feeds cloud, the `hitl_*` finding becomes High.

### Lens checklist (17/17)

`1` see findings · `2` see findings · `3` see findings · `4` **Clean** — dedicated security pass, plus: all published paths match served routes, spec has no `servers`/secrets, `/doc/` was never whitelisted · `5` N/A — no migrations or persisted state · `6` N/A — no concurrency primitives · `7` see findings; consumer half checked separately — endpoint, status codes and wire format are untouched, so `unstract-cloud/load_test` and `unstract-docs` are unaffected; the spec's consumers are the unmerged `#27` and the CLI · `8` N/A — generation is offline, no external calls added · `9` **Clean** · `10` **Clean** · `11` **Clean** — `backend/docs/` fully removed, zero residual `drf_yasg` references, no docker/chart references, no request-time behaviour change · `12` **Clean** — the only agent-surface change is `@extend_schema(exclude=True)` on `MCPServerView`, documentation-only · `13` see findings · `14` **Clean** — pinned with a stated reason, canonical package, lockfile consistent, transitive set shrinks · `15` see findings · `16` see findings · `17` see Unanchored

*Posted as `COMMENT`, not `REQUEST_CHANGES` — the merge decision is yours, not the review's.*

backend/api_v2/openapi_schema.py
```diff
@@ -0,0 +1,161 @@
...
+    metadata = serializers.JSONField(required=False)
+    metrics = serializers.JSONField(required=False)
+    error = serializers.CharField(required=False, allow_null=True)
+
+
+class ExecutionMessage(APIExecutionResponseSerializer):
+    """The execution's identity and, once it has finished, its per-file
+    results.
+    """
+
```
ritwik-g, line 62: **[High] [Lens 7, 3] — `error` and `status_api` are published required non-nullable, but every successful 200 sends `error: null`**

`ExecutionResponse.__post_init__` (`workflow_manager/workflow_v2/dto.py:72`) forces falsy `error` to `None`, and DRF emits `None` verbatim for a bare `CharField`. So the **happy path violates the published schema**. `status_api` is `null` on both error returns in `deployment_helper.py:300-307` and `:361-367` — the 422 and 500 bodies.

**Verified** — the real serializer replayed over the real dataclass:
```
{'execution_status':'PENDING','status_api':'/deployment/...','error':None,'result':None}
```
A pydantic/Go/Java client raises on the success response.

The comment just above at `:52-55` diagnoses this exact hazard for `result` and fixes it one field over — `error` is null *more* often than `result` is.

**Fix:** restate both as `CharField(required=False, allow_null=True)`, regenerate.

backend/api_v2/openapi_schema.py
```diff
@@ -0,0 +1,161 @@
...
+    """The documents to run, and the options that shape the result.
+
+    Supply `files`, `presigned_urls`, or both.
+    """
+
+
+class FileResult(serializers.Serializer):
+    file = serializers.CharField()
+    file_execution_id = serializers.CharField(required=False)
+    status = serializers.CharField(required=False)
```
ritwik-g, line 46: **[Medium] [Lens 7] — `FileResult` declares a `metrics` the server never emits and omits `extracted_text` which it does**

Metrics live one level down at `item["result"]["metrics"]` — `remove_result_metrics` pops it off the *inner* dict (`dto.py:163-169`) — so no writer produces a top-level `item["metrics"]` and `file_result.metrics` is always empty.

Separately, `promote_extracted_text` sets `item["extracted_text"]` (`dto.py:143-154`) whenever `include_extracted_text=true`, and that field is absent here — so a typed client drops the payload the caller explicitly opted in for.

**Verified** — traced both API-result writers (`endpoint_v2/destination.py:536-556`, `endpoint_v2/dto.py:137-146`) and grepped every `metrics`/`extracted_text` write across `backend/` and `workers/`.

**Fix:** drop `metrics`; add `extracted_text = CharField(required=False, allow_null=True)`.

backend/api_v2/openapi_schema.py
```diff
@@ -0,0 +1,161 @@
...
+# from here, so this is part of the public API surface.
+DEPLOYMENT_EXECUTION_SCHEMA = extend_schema_view(
+    post=extend_schema(
+        operation_id="execute",
+        tags=["deployment"],
+        auth=DEPLOYMENT_AUTH,
+        parameters=DEPLOYMENT_PATH_PARAMETERS,
+        request={"multipart/form-data": ExecuteRequest},
+        responses={
+            200: ExecuteResponse,
```
ritwik-g, line 205: **[Medium] [Lens 7] — `execute` can return 413/502/504 and arbitrary upstream statuses the spec does not declare**

`fetch_presigned_file` raises 413 on oversize (`deployment_helper.py:597-601`), 504 on timeout (`:665-668`), 502 on connection error (`:669-674`), and — most consequentially — **the upstream response's own status verbatim** (`:659-662`, `status_code = e.response.status_code`).

So an expired S3 signature surfaces as a 403 whose spec description reads "No API key was supplied.", and an S3 404 as "No such active deployment." Undeclared statuses reach a generated client as an unmodelled response.

**Fix:** declare 413/502/504, and reword the 403/404 descriptions so they don't assert a cause the endpoint can't guarantee. Better still, normalise upstream statuses to a single 502 in `fetch_presigned_file` so the published contract is closed.

backend/api_v2/openapi_schema.py
```diff
@@ -0,0 +1,161 @@
...
+    f"rejected, and the two together may not exceed "
+    f"{ExecutionRequestSerializer.MAX_FILES_ALLOWED} documents.\n\n"
+    "With the default `timeout` of -1 the call returns as soon as the "
+    "execution is queued; read the outcome from the status endpoint."
+)
+
+STATUS_DESCRIPTION = (
+    "Read the result of a previously started execution.\n\n"
+    "This read is one-shot: the first call that observes a completed execution "
+    "acknowledges it and the stored result is discarded, so every later call "
```
ritwik-g, line 186: **[Medium] [Lens 7, 16] — `STATUS_DESCRIPTION` tells callers to poll without saying the pending poll answers 422**

"Poll while the execution is pending" reads as a 200-returning loop. `api_deployment_views.py:230-246` starts `response_status` at `HTTP_422_UNPROCESSABLE_ENTITY` and raises it to 200 only on `CeleryTaskState.COMPLETED`.

Generated SDKs raise on 4xx by default, so the documented polling loop throws on **every iteration** until completion. The shape is declared (`422: StatusResponse`) so nothing is strictly wrong — the caller is simply not told the normal path is a non-2xx.

**Fix:** one sentence — a still-running execution answers 422 with the current `status`; only a completed one answers 200.

backend/api_v2/openapi_schema.py
```diff
@@ -0,0 +1,161 @@
...
+    ExecutionQuerySerializer,
+    ExecutionRequestSerializer,
+)
+
+
+# Declares no field of its own, so a change to the real serializer moves the
+# spec. It exists to carry a caller-facing description and a stable name.
+@extend_schema_serializer(component_name="ExecuteRequest")
+class ExecuteRequest(ExecutionRequestSerializer):
+    """The documents to run, and the options that shape the result.
```
ritwik-g, line 31: **[Medium] [Lens 1, 3, 7] — The OSS spec advertises two request fields that always 400 in OSS**

`ExecuteRequest` inherits the whole of `ExecutionRequestSerializer`, so the committed `docstudio-oss.json` publishes `hitl_queue_name` and `hitl_packet_id`. In an OSS install both are unconditionally rejected — `validate_hitl_queue_name` catches `ModuleNotFoundError` on `pluggable_apps.manual_review_v2` and raises `ValidationError` (`serializers.py:272-305`).

An OSS SDK user gets a 400 with an enterprise sales message from a parameter the SDK told them exists.

**Fix:** `@extend_schema_serializer(exclude_fields=("hitl_queue_name","hitl_packet_id"))`, or gate on `apps.is_installed(...)`.

**Open question:** is a *cloud* spec generated from this same command? That decides flat-exclude vs conditional.

backend/api_v2/management/commands/generate_docstudio_spec.py
```diff
@@ -0,0 +1,104 @@
...
+
+    Shared with the drift test: two copies of this could disagree, and then
+    the gate rejects exactly the file the command it names produces.
+    """
+    GENERATOR_STATS.reset()
+    schema = SchemaGenerator(urlconf=URLCONF).get_schema(request=None, public=True)
+    if GENERATOR_STATS:
+        # spectacular downgrades "unable to guess serializer" to a warning and
+        # writes a plausible, wrong operation. Nothing downstream can tell that
+        # apart from an annotation that is simply thin.
```
ritwik-g: **[Medium] [Lens 3] — The generation gate rejects guessed shapes but never checks the spec is legal OpenAPI**

`GENERATOR_STATS` only fires on spectacular's *own* inferences. Hand-written fragments pass through verbatim — `PATH_SEGMENT` (`openapi_schema.py:74`) and `APPEND_COMPONENTS.securitySchemes` (`settings/base.py:665-673`). A typo in either emits no diagnostic, the drift test compares equal, CI is green, and the invalid spec ships to the two repos `DOWNSTREAM` names — breaking as a generator crash with nothing here pointing at the cause.

**Verified** — injecting `"patern"` and `"htttp"` typos: both validate INVALID against the bundled `openapi_3_0_schema.json`, while the spec as committed validates clean. So the check would have caught them and nothing else does.

**Fix:** call `drf_spectacular.validation.validate_schema(schema)` in `render_spec()` after the `GENERATOR_STATS` block, re-raising as `SpecGenerationFailed`.

backend/api_v2/tests/test_docstudio_spec.py
```diff
@@ -0,0 +1,132 @@
...
+        (path, method, operation)
+        for path, path_item in spec["paths"].items()
+        for method, operation in path_item.items()
+        if method in _METHODS
+    ]
+
+
+def test_committed_spec_matches_the_code() -> None:
+    assert DEFAULT_OUT.exists(), f"{DEFAULT_OUT} is missing"
+    assert DEFAULT_OUT.read_text() == render_spec(), (
```
ritwik-g, line 71: **[Medium] [Lens 13] — The suite pins no semantic invariant of the spec, so a wrong spec passes**

This drift test compares the committed artifact against a spec rendered *from the same source*, so it catches a **stale** file but never a **wrong** one — any incorrect edit to `openapi_schema.py` passes the moment the author regenerates.

Two consequences: (a) `FileResult`, `StatusResponse` and `ErrorResponse` are new in this PR and matched against nothing, which is how the two High findings above got through; (b) the two client-breaking defects this PR was *written to fix* — `format: binary` and `result` nullability — have no invariant test, so dropping either decorator and regenerating is silently absorbed.

The one test that does anchor documentation to real output (`:124-132`) covers only the pre-existing `APIExecutionResponseSerializer`.

**Fix:** extend that test to `FileResult` (anchor on `dto.py:43-52`) and `StatusResponse`; assert the `files` item schema is `{"type":"string","format":"binary"}` and that `result` is nullable.

backend/api_v2/tests/test_docstudio_spec.py
```diff
@@ -0,0 +1,132 @@
...
+
+    assert reads
+    for status_op in reads:
+        assert "one-shot" in status_op["description"]
+        assert status_op["responses"]["406"]["description"].strip()
+
+
+def test_the_documented_response_fields_are_ones_the_code_produces() -> None:
+    """The view returns the execution DTO as a dict rather than through this
+    serializer, so a renamed DTO field would otherwise reach clients as a field
```
ritwik-g: **[Medium] [Lens 16] — This docstring says the execute response bypasses `APIExecutionResponseSerializer`; it is produced by it**

The docstring claims the view returns the DTO as a dict "rather than through this serializer". In fact `deployment_helper.py:367` is `return APIExecutionResponseSerializer(result).data`, wrapped at `api_deployment_views.py:171`. It is live request-path code building the 200/422 body.

A maintainer trusting this — reinforced by `openapi_schema.py`'s "none of them builds a response", which is true of *that* file — can prune its fields believing only the spec moves, and change the wire format.

**Fix:** state the real coupling. Confirming the serializer *is* on the response path makes this test's value clearer, not weaker.

backend/api_v2/openapi_schema.py
```diff
@@ -0,0 +1,161 @@
...
+    post=extend_schema(
+        operation_id="execute",
+        tags=["deployment"],
+        auth=DEPLOYMENT_AUTH,
+        parameters=DEPLOYMENT_PATH_PARAMETERS,
+        request={"multipart/form-data": ExecuteRequest},
+        responses={
+            200: ExecuteResponse,
+            409: OpenApiResponse(
+                ErrorResponse, description="The deployment has no active API key."
```
ritwik-g: **[Low] [Lens 7] — Two declared statuses this endpoint cannot return**

409 on `execute` and 429 on `status` are dead branches in every generated client. `NoActiveAPIKeyError` is raised only at `api_deployment_views.py:399` (`download_postman_collection`) and `pipeline_v2/views.py:253` — never from `DeploymentExecution.post`. `RateLimitExceeded` is raised only at `api_deployment_views.py:118`, inside `post` — never on the `get` path.

**Fix:** drop 409 from `post`; split 429 out of `DEPLOYMENT_ERRORS` so it applies to `post` only.

backend/api_v2/openapi_schema.py
```diff
@@ -0,0 +1,161 @@
...
+class ExecuteRequest(ExecutionRequestSerializer):
+    """The documents to run, and the options that shape the result.
+
+    Supply `files`, `presigned_urls`, or both.
+    """
+
+
+class FileResult(serializers.Serializer):
+    file = serializers.CharField()
+    file_execution_id = serializers.CharField(required=False)
```
ritwik-g: **[Low] [Lens 7] — `file_execution_id` is declared non-nullable but the DTO defaults it to `None`**

`FileExecutionResult.to_json` always emits the key and the field defaults to `None` (`endpoint_v2/dto.py:125,141`), sourced from `file_processing_result.file_execution_id` (`workers/shared/utils/api_result_cache.py:156`).

**UNVERIFIED** — I did not find a producer that leaves it unset, so this is a question rather than an asserted defect: is there a path where a cached file result carries no `file_execution_id`? Please confirm before changing anything on the strength of it.

backend/api_v2/tests/test_docstudio_spec.py
```diff
@@ -0,0 +1,132 @@
...
+    assert DEFAULT_OUT.read_text() == render_spec(), (
+        f"{DEFAULT_OUT} is out of date. Run `{REGENERATE}` from `backend/` and "
+        f"commit the result.\n\n{DOWNSTREAM}"
+    )
+
+
+def test_generation_reports_no_diagnostics() -> None:
+    """A warned-about operation is published with guessed request and response
+    shapes, and the drift comparison certifies the guess.
+    """
```
ritwik-g, line 80: **[Low] [Lens 3, 13] — `test_generation_reports_no_diagnostics` asserts on state unreachable in the failing case**

`render_spec()` raises whenever `bool(GENERATOR_STATS)`, and `GeneratorStats.__bool__` is `bool(self._warn_cache or self._error_cache)` (`drf_spectacular/drainage.py:42-43`) — the exact condition these two asserts test. So if it returns, both caches are empty by construction and the assertions can never observe a failure.

Its coverage is identical to `test_committed_spec_matches_the_code`, which also calls `render_spec()`. It reads as an independent diagnostics gate and is not one — while reaching into two private attributes to do it.

**Fix:** either drop it, or make it test the guard for real — seed `GENERATOR_STATS._warn_cache` and assert `render_spec()` raises `SpecGenerationFailed`, which is the behaviour nothing currently covers.

backend/api_v2/tests/test_docstudio_spec.py
```diff
@@ -0,0 +1,132 @@
...
+    spec = _committed()
+    scheme = spec["components"]["securitySchemes"]["deploymentKey"]
+
+    assert (scheme["type"], scheme["scheme"]) == ("http", "bearer")
+    for path, method, operation in _operations(spec):
+        assert operation["security"] == [{"deploymentKey": []}], f"{method} {path}"
+
+
+def test_clients_can_branch_on_every_failure_they_will_see() -> None:
```
ritwik-g, line 130: **[Low] [Lens 13] — Two operation-level tests pass vacuously on an empty spec**

Both this and `test_operations_require_the_deployment_key` (`:97-98`) loop over `_operations(_committed())` with no non-emptiness guard, so a spec whose `paths` collapsed to `{}` satisfies them silently.

Today the file as a whole still catches that via `:87` and `:118` — but incidentally rather than by design. `test_the_one_shot_read_is_documented_where_a_client_will_see_it` already guards itself with `assert reads` at `:118`; these two didn't get the same treatment.

**Fix:** `assert operations` after materialising the list, or move the guard into `_operations()`.

backend/api_v2/management/commands/generate_docstudio_spec.py
```diff
@@ -0,0 +1,104 @@
...
+
+The spec is the contract the published clients and their generated SDKs are
+built from, so it is committed and CI fails on drift: change a route, a
+serializer or the schema annotation, and regenerate in the same PR.
+
+    uv run python manage.py generate_docstudio_spec           # from backend/
+    uv run python manage.py generate_docstudio_spec --check   # no write, drift is an error
+
+The generated paths carry ``API_DEPLOYMENT_PATH_PREFIX``, so regenerate in an
+environment that does not override it — the committed artifact describes the
```
ritwik-g: **[Low] [Lens 13] — The drift gate's outcome depends on ambient `API_DEPLOYMENT_PATH_PREFIX`, enforced only by this docstring**

Generated paths carry `settings.API_DEPLOYMENT_PATH_PREFIX` (`backend/backend/base_urls.py:20`), read from the environment at import with a default of `deployment`. A developer whose environment overrides it sees the drift test fail, regenerates as instructed, and commits a spec whose paths carry a private prefix — the test then passes on the wrong artifact and the downstream SDK repos are generated from it.

Low because nothing in-repo sets the variable (no `.env` sample, no compose file, and the rig's `backend_test_env` doesn't pin it) and a prefix change is visible in the `specs/` diff.

**Fix:** assert the rendered path prefix equals the default, or override the setting for the duration of `render_spec()`.

backend/api_v2/management/commands/generate_docstudio_spec.py
```diff
@@ -0,0 +1,104 @@
...
+        out: Path = options["out"]
+        if options["check"]:
+            current = out.read_text() if out.exists() else ""
+            if current != rendered:
+                raise CommandError(
+                    f"{out} is out of date. Run `{REGENERATE}` from `backend/` "
+                    f"and commit the result.\n\n{DOWNSTREAM}"
+                )
+            self.stdout.write(f"{out} is up to date")
+            return
```
ritwik-g, line 118: **[Low] [Lens 13] — The `--check` branch and the write branch are exercised by nothing**

`--check` is invoked by no CI job (`ci-test.yaml` runs only the tox tiers), no pre-commit hook, and no test. The actual drift gate in CI is `test_committed_spec_matches_the_code`, which reimplements the same comparison. So the branch the module docstring advertises as *the* drift command is untested, and the write branch's summary arithmetic at `:94-103` is likewise unexercised.

**Fix:** exercise `handle()` via `call_command` (up-to-date, drifted, and write-to-tmp via `--out`), or drop `--check` and point the docstring at the pytest gate that actually runs.

(For the record, the drift gate itself *is* real — `tests/groups.yaml:85-94` collects `backend/**` by glob, and CI shows `unit-backend` at 1009 passed. Only this branch is dead.)

backend/api_v2/management/commands/generate_docstudio_spec.py
```diff
@@ -0,0 +1,104 @@
...
+def render_spec() -> str:
+    """The committed artifact, byte for byte.
+
+    Shared with the drift test: two copies of this could disagree, and then
+    the gate rejects exactly the file the command it names produces.
+    """
+    GENERATOR_STATS.reset()
+    schema = SchemaGenerator(urlconf=URLCONF).get_schema(request=None, public=True)
+    if GENERATOR_STATS:
+        # spectacular downgrades "unable to guess serializer" to a warning and
```
ritwik-g: **[Low] [Lens 16] — This comment misstates spectacular's diagnostic severity and what it emits**

The comment says spectacular "downgrades 'unable to guess serializer' to a warning and writes a plausible, wrong operation" (restated at `test_docstudio_spec.py:55-57`). In the pinned 0.30.0, that message is emitted through `error()` so it lands in `_error_cache`, not `_warn_cache`; and serializer resolution returns `None`, so the operation is published with **no** request body and a `"No response body"` response — not a guessed shape.

The guard is correct because it checks both caches. But a maintainer debugging a future failure looks in the wrong bucket and expects a fabricated schema that is never there.

Evidence: `drf_spectacular/openapi.py:1269-1273` and `:1498-1499`.

Also here: `docstudio` (`:23`, `:25`) appears nowhere else in the repo, while the spec's own `info.title` is "Unstract API" — a maintainer has no path from either name to the other. Consider `generate_api_deployment_spec` / `specs/api-deployment-oss.json`, unless it's an established product name.

backend/mcp_server/views.py
```diff
@@ -23,6 +24,10 @@
 logger = logging.getLogger(__name__)
 
 
+# MCP speaks JSON-RPC over one POST, so it has no REST surface worth
+# describing; leaving it in would publish guessed request and response shapes
```
ritwik-g: **[Low] [Lens 16] — This comment states a consequence the PR's own guard makes impossible**

The justification is that "leaving it in would publish guessed request and response shapes to every client generated from the spec." `MCPServerView` has no `serializer_class`/`get_serializer`, so without the exclusion spectacular emits `unable to guess serializer` into `_error_cache` — and `render_spec()`, added in this same PR, raises `SpecGenerationFailed` on any non-empty `GENERATOR_STATS`.

Nothing would be published; the command and the drift test would fail loudly. The stated risk is precisely the case the new guard exists to prevent, so the comment misrepresents the safety net a maintainer is standing on.

**Fix:** "…leaving it in would fail generation, since spectacular cannot guess a serializer for a JSON-RPC APIView and `render_spec` treats any diagnostic as fatal."

backend/api_v2/openapi_schema.py
chandrasekharan-zipstack: Fixed. `DEFAULT_SCHEMA_CLASS` is now `drf_standardized_errors.openapi.AutoSchema`, and `ErrorResponse` is restated as the handler's real body: `{type: ErrorType, errors: [{code, detail, attr}]}`, with a new `ErrorDetail` component. `{status, message}` survives only on the branches the views build by hand — `AcknowledgedResponse` for the status 406, `ExecuteResponse`/`StatusResponse` for the 422 and 500 those two views return themselves.

One deviation worth flagging: I did not reuse the package's `ErrorResponse401Serializer`/`403`/`404`/`429` classes. They enumerate `code` as `not_authenticated`/`permission_denied`/`not_found`/`throttled`, but every deployment exception here inherits `APIException.default_code`, so the handler emits `code: "error"` — exactly the output you pinned. Publishing those enums would have replaced one disjoint field with a strict enum the server violates, so `code` is a free-form string with a comment saying why.

Note the auto-injection in that `AutoSchema` is inert for us either way: `_should_add_error_response` bails when `view.get_exception_handler() is not standardized_errors_handler`, and ours is the `drf_logging_exc_handler` wrapper. The explicit declarations are what carry the shape; the schema class is the correct base to sit on.

A new test replays `drf_logging_exc_handler` over a real `APIException` and a real `ValidationError` and asserts the emitted keys equal the spec's `ErrorResponse`/`ErrorDetail` required sets and that `type` is in the published enum.

backend/api_v2/openapi_schema.py
chandrasekharan-zipstack: Fixed. `ExecutionMessage` now restates `status_api` and `error` as `CharField(required=False, allow_null=True)` alongside `result`, with one comment covering all three: the real serializer declares them required and non-nullable while `ExecutionResponse.__post_init__` coerces every falsy value to `None`. Regenerated — both are `nullable: true` and out of `required` in the committed spec.

backend/api_v2/openapi_schema.py
chandrasekharan-zipstack: Fixed. `metrics` is dropped and `extracted_text` added as a nullable `CharField` with a description saying it is sent only when the request set `include_extracted_text`. I also marked `result` and `metadata` nullable while there — both are `| None` on `FileExecutionResult` and `to_json` emits them verbatim.

A new test now anchors this: the documented `FileResult` property set must be a subset of `FileExecutionResult`'s dataclass fields plus an explicit `_PROMOTED_FILE_RESULT_FIELDS = {"extracted_text"}`, which is the one key the workflow copies up rather than the DTO carrying it. Adding a phantom field like `metrics` back now fails.

backend/api_v2/openapi_schema.py
chandrasekharan-zipstack: Fixed on the spec side. 413, 502 and 504 are declared on `execute` in a new `EXECUTE_ERRORS` block, and the 403/404 descriptions no longer assert a cause — "The request was refused as unauthorized." and "No active deployment, or a referenced document, was found."

I left `fetch_presigned_file`'s upstream-status passthrough alone: normalising it to 502 changes what live callers get back, which is a behaviour change and out of scope for a spec PR. Worth its own issue.

A new test asserts `{413, 429, 502, 504}` are declared on `execute` and on nothing else, so the split cannot silently drift back.

backend/api_v2/openapi_schema.py
chandrasekharan-zipstack: Fixed. `STATUS_DESCRIPTION` gains: "A still-running execution answers 422 carrying its current `status`, so a polling loop should treat 422 as the normal reply and stop on 200. Clients that raise on any non-2xx need to allow for that." The 422 response description also now distinguishes still-running from finished-with-an-error. `test_the_one_shot_read_is_documented_where_a_client_will_see_it` asserts `"422"` appears in the description.

backend/api_v2/openapi_schema.py
chandrasekharan-zipstack: Answering the open question first: **no**, there is no separate cloud spec. `generate_docstudio_spec` produces one artifact and that artifact is what the published PyPI client is generated from, so it serves cloud users too. That settles it against excluding the fields — the endpoint surface is identical for a cloud deployment, and an SDK that omits `hitl_queue_name`/`hitl_packet_id` would make the enterprise feature unreachable from the official client.

So they stay, but the spec now says what they need. Both fields carry `help_text` on `ExecutionRequestSerializer` (published as the client-facing description): "Requires the enterprise manual-review capability; an installation without it rejects the request with 400." That turns the 400 you describe from a surprise into a documented precondition.

Leaving this unresolved for you, since keeping the fields is a call rather than a fix.

backend/api_v2/management/commands/generate_docstudio_spec.py
chandrasekharan-zipstack: Fixed. `render_spec()` now calls `drf_spectacular.validation.validate_schema(schema)` after the `GENERATOR_STATS` block and re-raises as `SpecGenerationFailed`. Verified against your repro: injecting `"patern"` into `PATH_SEGMENT` now fails generation with the jsonschema error pointing at the offending fragment, and the spec as committed validates clean.

Same block also grew a check that every generated path sits under `/deployment/` — see the `API_DEPLOYMENT_PATH_PREFIX` thread.

backend/api_v2/tests/test_docstudio_spec.py
chandrasekharan-zipstack: Fixed. The file now carries invariants that fail on a wrong spec, not just a stale one:

- `test_documents_are_uploaded_as_binary_not_as_urls` — asserts the `files` item schema is exactly `{"type": "string", "format": "binary"}`.
- `test_the_result_a_pending_execution_omits_is_documented_nullable` — asserts `ExecutionMessage.result` and `StatusResponse.message` are `nullable: true`.
- `test_the_documented_file_result_fields_are_ones_the_code_produces` — `FileResult`'s properties against `FileExecutionResult`'s dataclass fields plus the one explicitly promoted key.
- `test_the_status_read_documents_the_two_keys_it_returns` — pins `StatusResponse` to `{status, message}` and its item `$ref` to `FileResult`.
- `test_the_documented_error_body_is_the_one_the_handler_sends` — replays the real exception handler and matches its output against the spec's error components.

Dropping either decorator this PR was written for now fails regardless of regeneration.

backend/api_v2/tests/test_docstudio_spec.py
chandrasekharan-zipstack: Fixed — you are right, `deployment_helper.py:367` returns `APIExecutionResponseSerializer(result).data`. The docstring now reads: "`APIExecutionResponseSerializer` builds the live execute response, so a field documented here that the DTO no longer carries reaches clients as a field the server never sends." The module docstring in `openapi_schema.py` is unchanged and still accurate — it is scoped to the serializers in that file.

backend/api_v2/openapi_schema.py
chandrasekharan-zipstack: Fixed. 409 is gone — grepped `NoActiveAPIKeyError` and it is only raised at `api_deployment_views.py:399` and `pipeline_v2/views.py:253`, neither on this path. 429 moved out of `DEPLOYMENT_ERRORS` into the execute-only `EXECUTE_ERRORS`, since `RateLimitExceeded` is raised only inside `post`. `test_only_the_execution_endpoint_documents_the_statuses_only_it_returns` keeps the split honest, and the shared-errors test drops 429 accordingly.

backend/api_v2/openapi_schema.py
chandrasekharan-zipstack: Made it nullable, and the reason is in the DTO rather than in a producer hunt: `FileExecutionResult.file_execution_id` defaults to `None` and `to_json` emits the key unconditionally, so `null` is reachable by construction whatever the callers do today. `DestinationConfig.to_json` (`endpoint_v2/dto.py:100`) makes the same allowance explicitly — `str(...) if self.file_execution_id else None`.

I did not find a live producer that leaves it unset either, so this is a declaration matching the type rather than a bug fix. Nullable is the safe direction: a client that never sees `null` is unharmed, one that does would crash. `metadata` and `result` got the same treatment for the same reason.

backend/api_v2/tests/test_docstudio_spec.py
chandrasekharan-zipstack: Fixed, and thanks — seeding `_warn_cache` before the call does not work either, because `render_spec()` opens with `GENERATOR_STATS.reset()`. The test now monkeypatches `SchemaGenerator.get_schema` with a stub that emits a real `drf_spectacular.drainage.warn(...)` mid-generation and asserts `render_spec()` raises `SpecGenerationFailed` with that message. It reaches into no private attributes and fails if the guard is removed.

backend/api_v2/tests/test_docstudio_spec.py
chandrasekharan-zipstack: Fixed at the source: `_operations()` now ends with `assert operations, "the spec documents no operation at all"`, so every caller — including the two you named and the three added since — inherits the guard instead of each needing its own.

backend/api_v2/management/commands/generate_docstudio_spec.py
chandrasekharan-zipstack: Fixed, by asserting the output rather than overriding the setting — `override_settings` would not have worked, since `base_urls.urlpatterns` bakes the prefix in at import and the resolver is already built by the time `render_spec()` runs. So after the diagnostics gate, generation rejects any path not under `/deployment/` (`PUBLISHED_PATH_PREFIX`) with a message telling the developer to unset the variable. Verified: `API_DEPLOYMENT_PATH_PREFIX=private manage.py generate_docstudio_spec --check` now fails instead of writing a private-prefix spec. The module docstring says this rather than merely asking.

backend/api_v2/management/commands/generate_docstudio_spec.py
chandrasekharan-zipstack: Fixed. Three tests drive `handle()` through `call_command`: `--check` against the committed spec (passes), `--check --out <tmp>` against a drifted file (raises `CommandError`, matched on "out of date"), and `--out <tmp>/nested/spec.json` for the write branch, asserting the written bytes equal the committed file — which also exercises the `mkdir(parents=True)` and the summary arithmetic. Thanks for confirming the drift gate itself is real; that saved a detour.

backend/api_v2/management/commands/generate_docstudio_spec.py
chandrasekharan-zipstack: Comment fixed: it now says an unresolved operation is "published empty rather than dropped", and that both caches are drained because the severity a given diagnostic carries is spectacular's choice rather than something to rely on. No claim about a fabricated shape, and no claim about which cache it lands in. The test docstring says the same.

On the naming: keeping `docstudio`. It is the established internal product name for this surface, and renaming the command and the artifact would break the downstream repos already generating from `specs/docstudio-oss.json`. The `info.title` staying "Unstract API" is deliberate — that string is what SDK users see, and it should name the product, not our internal surface. Leaving unresolved since that half is a judgement call rather than a fix.

backend/mcp_server/views.py
chandrasekharan-zipstack: Fixed, reworded almost as you suggested: "…leaving it in would fail spec generation, which treats any generator diagnostic as fatal and cannot resolve a serializer for this view." The comment now points at the guard rather than at a consequence the guard prevents.

chandrasekharan-zipstack:
Thanks — this was a genuinely useful review; the two High findings were real and both would have shipped into the generated client. Everything is addressed in `eddd4b7`. Inline replies cover the anchored threads; the unanchored ones and your open questions:

**`backend/README.md:178-179`** — replaced. The *API Docs* section now points at `specs/docstudio-oss.json` as the committed contract and gives the regeneration and `--check` commands, instead of describing a route that 404s.

**PR body** — restored to the template, including **Dependencies Versions**: `drf-yasg 1.21.15` removed, `drf-spectacular==0.30.0` added (pinned exactly, because the generated artifact is committed and a minor bump would surface as spec drift rather than as a dependency change). Also filled in *Env Config*, *Relevant Docs*, *Screenshots* and the break-risk section.

**Open question 1 — is a cloud spec generated from this command?** No. This is the single artifact, and the PyPI client generated from it serves cloud users, so the `hitl_*` parameters stay and instead say what they require. Details on that thread.

**Open question 2 — was `{status, message}` deliberate?** Neither, really: it was written from the two hand-built bodies in `api_deployment_views.py` and then generalised to every failure without checking what the handler emits. Now `{status, message}` is declared only on the branches that actually build it by hand, and everything else carries the handler's shape.

**Assumption you flagged** — `drf-standardized-errors` as the active handler is now load-bearing rather than incidental: `DEFAULT_SCHEMA_CLASS` names its `AutoSchema`, and a test replays `drf_logging_exc_handler` over real exceptions and asserts the emitted keys match the published `ErrorResponse`/`ErrorDetail`. A deployment that overrode `EXCEPTION_HANDLER` would fail that test rather than quietly publish a wrong shape.

**Also added while in here**, both prompted by your findings rather than requested: generation now rejects a spec whose paths carry a non-default `API_DEPLOYMENT_PATH_PREFIX`, and the error-type enum is named `ErrorType` rather than spectacular's default `TypeEnum`, since that name becomes a class in the generated SDK.

One thing deliberately **not** done: `fetch_presigned_file` still passes the upstream response's status through verbatim. Normalising it to a single 502 changes what live callers receive, which does not belong in a spec PR — the spec now declares 413/502/504 and stops asserting a single cause for 403/404. Worth a follow-up issue.

Backend unit tier: `676 passed, 1 skipped` (340 subtests), 18 of them spec tests. Both new generation gates verified by hand — a `"patern"` typo in `PATH_SEGMENT` now fails OpenAPI validation, and `API_DEPLOYMENT_PATH_PREFIX=private` fails the prefix check.

ritwik-g (approved):
## FOLLOWUP review — verdict: `APPROVE`

**Prior findings: 17 · RESOLVED 16 · WAIVED 1 (with a documented rationale) · NOT RESOLVED 0**
**New findings: none. Scope change: NO.**

Verified against `879ec5ab` from the current merge-base `e452f778`. The 12 commits since my review are 11 from a `main` merge plus one fix commit, `eddd4b74`. **Status here is what the code shows, not what the replies claim.**

### The two blockers, verified against live behaviour rather than against the tests

**`ErrorResponse`** is now `{type: ErrorType, errors: [ErrorDetail]}`, both required, bound to 400/401/403/404 on both operations and 413/429/502/504 on `execute`. Rather than trust the new test, I replayed the real `drf_logging_exc_handler` over six real exceptions — `APIException`, `ValidationError`, `NotAuthenticated`, `PermissionDenied`, `NotFound`, `Throttled` — and the emitted keys match the published component exactly, with all three emitted `type` values inside the published enum. Leaving `code` free-form was the right call: it varies (`error`, `invalid`, `not_found`, `not_authenticated`, `permission_denied`, `throttled`), so an enum would have been violated.

I also checked the risk this fix could have introduced — that the hand-built bodies got the new shape by mistake. They didn't: 406 → `AcknowledgedResponse{status, message}`, `status` 422/500 → `StatusResponse{status, message}`, `execute` 422/500 → `ExecuteResponse{message}` with no `status` key, each matching its `Response(...)` call site.

**`error` / `status_api`** are `nullable: true` in the committed artifact and both removed from `required` (which is now just `execution_id`, `execution_status`). Confirmed in the regenerated spec, not only in the annotation source.

### The new invariants are real guards

Five were each broken and observed to fail: removing `@extend_schema_field(OpenApiTypes.BINARY)`, dropping `allow_null` from `result`, re-adding the phantom `metrics`, and moving 429 back onto `status` all now fail after regeneration. `test_the_documented_error_body_is_the_one_the_handler_sends` genuinely replays the live handler rather than restating the implementation. Your claim that "dropping either decorator this PR was written for now fails regardless of regeneration" holds as tested.

The `patern` typo repro from my finding now produces `SpecGenerationFailed: … is not valid under any of the given schemas`, and `API_DEPLOYMENT_PATH_PREFIX=private` is rejected by output assertion — a better lever than the `override_settings` I suggested, since `base_urls` bakes the prefix in at import.

### The one waiver, and two notes

The HITL fields stay in the OSS artifact because one spec serves both OSS and cloud — excluding them would make the enterprise feature unreachable from the official client. That answers the open question I raised and is a sound trade; the 400 is now a documented precondition via `help_text` published as the client-facing description, rather than a surprise.

Two things carried forward, neither blocking and neither introduced here: `deployment_helper.py:654-658` still passes an upstream document-store status verbatim, so a store answering e.g. 410 still surfaces undeclared (body shape is fine — it's an `APIException`); and `middleware/exception.py:38-47` has an `SdkError` branch emitting a third shape, which I could not demonstrate is reachable on these two operations.

Suite green: 18 in the spec tests, 260 across `api_v2 middleware mcp_server`, both drift tests passing so the artifact regenerates byte-identically. CI green including e2e. Nice work on `backend/README.md` — that closes the stale `/api/v1/doc/` pointer cleanly.

PR #2238, opened 2026-08-12
Treat MiniMax M2 thinking as always on

octo-patch:
Reason: Align MiniMax-M2.7 requests with its always-on thinking contract.

- Stop injecting an adaptive thinking payload for always-on M2-family models.
- Accept an explicit enabled setting without forwarding it, while rejecting disabled or configurable thinking payloads.
- Constrain the adapter schema and cover the always-on behavior with focused tests.

Checks:

- `uv run --project unstract/sdk1 --locked ruff check --ignore I001 unstract/sdk1/src/unstract/sdk1/adapters/base1.py unstract/sdk1/tests/test_branded_openai_adapters.py`
- `uv run --project unstract/sdk1 --locked ruff format --check unstract/sdk1/src/unstract/sdk1/adapters/base1.py unstract/sdk1/tests/test_branded_openai_adapters.py`
- `uv run --project unstract/sdk1 --locked pytest -q unstract/sdk1/tests/test_branded_openai_adapters.py -k 'minimax_m2 or minimax_schema_covers'`
- `git diff --check`

### DIFF
--- a/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
@@ -584,10 +584,9 @@ def _normalize_minimax_thinking(
     if thinking is None:
         return
     if is_m2_model:
-        raise ValueError(
-            f"{model_id} uses always-on thinking and does not accept "
-            "thinking configuration."
-        )
+        # M2.x always thinks; the provider accepts this parameter but ignores it.
+        adapter_metadata.pop("thinking")
+        return
     if not isinstance(thinking, dict) or thinking.get("type") not in {
         "adaptive",
         "disabled",
@@ -634,8 +633,6 @@ class MiniMaxLLMParameters(BaseChatCompletionParameters):
         _normalize_minimax_thinking(adapter_metadata, model_id)
 
         validated = MiniMaxLLMParameters(**adapter_metadata).model_dump()
-        if _is_minimax_m2_model(model_id):
-            validated.pop("thinking", None)
         validated["cost_model"] = f"{_MINIMAX_PROVIDER_PREFIX}{model_id}"
         if context_window := _minimax_context_window(model_id):
             validated["context_window"] = context_window
--- a/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
@@ -568,14 +568,14 @@ def _minimax_context_window(model_id: str) -> int | None:
 def _normalize_minimax_thinking(
     adapter_metadata: dict[str, "Any"], model_id: str
 ) -> None:
+    is_m2_model = _is_minimax_m2_model(model_id)
     if "enable_thinking" in adapter_metadata:
         enable_thinking = adapter_metadata.pop("enable_thinking")
         if not isinstance(enable_thinking, bool):
             raise ValueError("enable_thinking must be a boolean.")
-        if _is_minimax_m2_model(model_id):
-            if not enable_thinking:
-                raise ValueError(f"{model_id} does not support disabling thinking.")
-        else:
+        if is_m2_model and not enable_thinking:
+            raise ValueError(f"{model_id} does not support disabling thinking.")
+        if not is_m2_model:
             adapter_metadata["thinking"] = {
                 "type": "adaptive" if enable_thinking else "disabled"
             }
@@ -583,7 +583,7 @@ def _normalize_minimax_thinking(
     thinking = adapter_metadata.get("thinking")
     if thinking is None:
         return
-    if _is_minimax_m2_model(model_id):
+    if is_m2_model:
         raise ValueError(
             f"{model_id} uses always-on thinking and does not accept "
             "thinking configuration."
@@ -634,6 +634,8 @@ class MiniMaxLLMParameters(BaseChatCompletionParameters):
         _normalize_minimax_thinking(adapter_metadata, model_id)
 
         validated = MiniMaxLLMParameters(**adapter_metadata).model_dump()
+        if _is_minimax_m2_model(model_id):
+            validated.pop("thinking", None)
         validated["cost_model"] = f"{_MINIMAX_PROVIDER_PREFIX}{model_id}"
         if context_window := _minimax_context_window(model_id):
             validated["context_window"] = context_window
--- a/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
+++ b/unstract/sdk1/src/unstract/sdk1/adapters/base1.py
@@ -565,6 +565,36 @@ def _minimax_context_window(model_id: str) -> int | None:
     return None
 
 
+def _normalize_minimax_thinking(
+    adapter_metadata: dict[str, "Any"], model_id: str
+) -> None:
+    if "enable_thinking" in adapter_metadata:
+        enable_thinking = adapter_metadata.pop("enable_thinking")
+        if not isinstance(enable_thinking, bool):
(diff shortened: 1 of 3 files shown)


octo-patch:
Thanks for the review. Commit a235538 now omits the thinking parameter for MiniMax M2 requests and aligns schema matching with runtime handling. I ran 64 adapter tests plus Ruff format and lint checks.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -565,6 +565,36 @@ def _minimax_context_window(model_id: str) -> int | None:
...
+            }
+
+    thinking = adapter_metadata.get("thinking")
+    if thinking is None:
+        return
+    if is_m2_model:
+        raise ValueError(
+            f"{model_id} uses always-on thinking and does not accept "
+            "thinking configuration."
+        )
```
chandrasekharan-zipstack: Suggest stripping rather than raising here.

The MiniMax docs say the opposite of this message: *"For M2.x models, thinking cannot be disabled; `thinking: {"type": "disabled"}` is accepted but thinking remains on."* The API takes the parameter, it just ignores it.

Practical cost: this turns a currently-working call into a hard failure at `LLM.__init__` (`llm.py:250`), which runs on every completion and on Test Connection — not only at save time. And the form schema has no `thinking` property, so it can only ever fire for SDK/API callers, who are the least served by a message that contradicts the provider.

The `enable_thinking: false` raise above is the one worth keeping — that's a user explicitly asking for something impossible.

```suggestion
    if is_m2_model:
        # M2.x always thinks; the provider accepts this parameter but ignores it.
        adapter_metadata.pop("thinking")
        return
```

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -601,28 +631,11 @@ def validate(adapter_metadata: dict[str, "Any"]) -> dict[str, "Any"]:
...
-                "disabled",
-            }:
-                raise ValueError("thinking.type must be adaptive or disabled.")
-            if _is_minimax_m2_model(model_id) and thinking["type"] == "disabled":
-                raise ValueError(f"{model_id} does not support disabling thinking.")
+        _normalize_minimax_thinking(adapter_metadata, model_id)
 
         validated = MiniMaxLLMParameters(**adapter_metadata).model_dump()
+        if _is_minimax_m2_model(model_id):
+            validated.pop("thinking", None)
```
chandrasekharan-zipstack: This only ever deletes a `None`.

By the time it runs, `thinking` is guaranteed absent from `adapter_metadata` for M2 — the helper raises if it was passed, and the `enable_thinking` branch skips M2 — so Pydantic emits the declared field as `None`:

```
MiniMaxLLMParameters(model='minimax/MiniMax-M2.7', api_key='k').model_dump()
  → thinking key present: True | value: None
```

That's the same `None` the M3 path keeps and has shipped with since #2166. Popping it only for M2 leaves two shapes for "no thinking" (M2: key absent, M3: key `None`), which is an easy thing to trip over later.

Suggest deleting both lines. Re-validation stays safe — the helper short-circuits on `thinking is None`, confirmed with `validate({'model': 'MiniMax-M2.7', 'api_key': 'k', 'thinking': None})`.

The assertion at line 192 then becomes `assert validated["thinking"] is None`.

chandrasekharan-zipstack (commented):
Review angle, so the inline comments make sense together.

The underlying fact checks out: per [MiniMax's docs](https://platform.minimax.io/docs/api-reference/text-openai-api), M2.x thinking is always on. But the same docs also say `thinking: {"type": "disabled"}` is **accepted and ignored**, not rejected. So the previously-injected `{"type": "adaptive"}` was a no-op on the wire, and this is a hygiene change rather than a bug fix. Worth being explicit about that since there's no ticket behind the PR.

Happy to take the Python refactor. Extracting `_normalize_minimax_thinking()` is a genuine readability win over the inline block it replaces, and no longer sending a parameter the model ignores is correct.

Asking to drop three things that add failure surface without buying behaviour: the new raise on an explicit `thinking` payload, the `validated.pop`, and the `allOf` conditional in the schema. Details inline. That should land around +25/-20 instead of +98/-24.

Verified locally: the suite passes on this branch (64 tests), and `MiniMaxLLMParameters(model='minimax/MiniMax-M2.7', api_key='k').model_dump()` returns `thinking: None` — which is what the pop is deleting.

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/minimax.json
```diff
@@ -73,7 +73,28 @@
...
+      },
+      "then": {
+        "properties": {
+          "enable_thinking": {
+            "const": true
+          }
+        }
+      }
+    }
+  ]
```
chandrasekharan-zipstack: Suggest dropping this block.

It encodes "what is an M2 model" a third time — `_is_minimax_m2_model` (`base1.py:557`) and the prefix-stripping in `validate_model` are the other two — in a different language, with hand-expanded case-insensitivity because JSON Schema has no `/i` flag. Nothing fails if the two drift, so the next model family silently loses the hint while tests stay green.

One thing worth knowing before merging either way: I ran this schema through `@rjsf/utils@5` + `validator-ajv8@5`, which is what `RjsfFormLayout.jsx` uses. The condition does resolve correctly — M2 gets `const: true` and the checkbox defaults to checked — but unchecking yields two errors, and the second is raw ajv output:

```
[".enable_thinking must be equal to constant", "must match \"then\" schema"]
```

`transformErrors` (`RjsfFormLayout.jsx:177`) has no `if` case, so `must match "then" schema` reaches the user. No other adapter schema in the repo uses `allOf`/`if`, so this would become the pattern others copy.

The reworded description just above already conveys always-on, and the backend still rejects `enable_thinking: false`. If we want form-time feedback later, `ui:widget: hidden` for M2 reads better than a checkbox that accepts only one value — but it needs the same duplication solved, so it belongs in its own change.

unstract/sdk1/tests/test_branded_openai_adapters.py
```diff
@@ -321,6 +340,27 @@ def test_minimax_schema_covers_models_thinking_and_regions() -> None:
         "standard",
         "priority",
     ]
+    assert schema["allOf"][0]["then"]["properties"]["enable_thinking"] == {"const": True}
```
chandrasekharan-zipstack: `allOf[0]` couples this to schema layout — add a second conditional and it silently asserts the wrong branch. The `Draft202012Validator` loop just below already covers the same rule through behaviour, which is the version worth keeping.

Also `from jsonschema import Draft202012Validator` at line 329 should sit at the top of the file per our import convention.

Both go away if the `allOf` block is dropped.

octo-patch:
Thanks for the review. I dropped all three as requested and pushed commit 4ad8ebd.

- `_normalize_minimax_thinking` now strips an explicit `thinking` payload for M2.x instead of raising, so a call that works today keeps working.
- Removed the M2-only `validated.pop`, so `thinking` stays `None` when unset for the whole family rather than leaving two shapes for ""no thinking"".
- Dropped the `allOf` block from the schema, along with the `allOf[0]` assertion and the validator loop it existed for; that also removes the local import you flagged. The reworded description carries the always-on note.

The `enable_thinking: false` raise is kept. Agreed on the framing too: this is hygiene rather than a bug fix, since the parameter it stops sending was accepted and ignored on the wire.

Checks: `pytest tests/test_branded_openai_adapters.py` (64 passed) and Ruff v0.3.4 check + format, both clean.

PR #2240, opened 2026-08-13
UN-3991 [FIX] Show model names on Prompt Studio tiles for shared users

kirtimanmishrazipstack:
## What

- Prompt Studio output tiles now show the model name and icon for anyone who can open the project, not just its owner.
- The backend sends those display details along with each LLM profile, so the UI no longer has to look them up on its own.

## Why

- A user the project was shared with — directly or through a group — saw a blank model name and no icon on the tiles.
- The UI built those labels from the list of adapters that user personally has access to, and a shared project's adapters are not in it.

## How

- The profile serializer resolves the four adapter FKs at render time and returns `conf` (per-adapter model label) and `icon` — display data only, no credentials.
- `get_queryset` adds `select_related` on those four FKs so the extra reads don't become N+1.
- `PromptCardItems.jsx` renders the payload as-is; the frontend adapter lookup is gone.

- No. The payload only gains `conf`/`icon`; the adapter fields keep returning the adapter name as before. `AdapterInstance.objects.for_user` is untouched — this changes what is displayed, not what a user can reach or use.

## Database Migrations

- None

## Env Config

- None

## Relevant Docs

-

## Related Issues or PRs

- Zipstack/unstract-cloud#1722 — agentic Prompt Studio, same ticket

## Dependencies Versions

- None

## Notes on Testing

- `backend/prompt_studio/prompt_profile_manager_v2/tests/test_profile_display_info.py` — covers adapter FK resolution with no adapter access.
- Manual: share a Prompt Studio project that uses a private LLM (once directly, once via a group), open it as the invited user — the tiles show the model name.

## Screenshots
- 

- 

## Checklist

### DIFF
--- a/backend/adapter_processor_v2/adapter_processor.py
+++ b/backend/adapter_processor_v2/adapter_processor.py
@@ -20,7 +20,6 @@ from unstract.sdk1.adapters.adapterkit import Adapterkit
 from unstract.sdk1.adapters.base import Adapter
 from unstract.sdk1.adapters.x2text.constants import X2TextConstants
 from unstract.sdk1.constants import AdapterTypes
-from unstract.sdk1.constants import Common as common
 from unstract.sdk1.embedding import EmbeddingCompat
 from unstract.sdk1.exceptions import SdkError
 from unstract.sdk1.llm import LLM
@@ -113,29 +112,35 @@ class AdapterProcessor:
         return updated_adapters[0].get(key_value)
 
     @staticmethod
-    def get_display_info(adapter: AdapterInstance) -> tuple[str, str]:
-        """Icon and model label for an adapter, as (icon, model).
+    def get_icon(adapter: AdapterInstance) -> str:
+        """Registry icon for an adapter, or the warning icon if unresolvable."""
+        if not adapter.is_available:
+            return UNAVAILABLE_ADAPTER_ICON
+        try:
+            adapter_class = Adapterkit().get_adapter_class_by_adapter_id(
+                adapter.adapter_id
+            )
+        except Exception as e:
+            logger.warning(
+                "Adapter %s is not in the SDK registry: %s", adapter.adapter_id, e
+            )
+            return UNAVAILABLE_ADAPTER_ICON
+        return adapter_class.get_icon() or UNAVAILABLE_ADAPTER_ICON
 
-        Display data only, no credentials. Never raises - falls back to the
-        warning icon and the adapter id's provider prefix.
+    @staticmethod
+    def get_model_label(adapter: AdapterInstance) -> str:
+        """Model name from the adapter metadata.
+
+        Adapters without a model use the adapter id's provider prefix.
         """
-        icon = UNAVAILABLE_ADAPTER_ICON
-        if adapter.is_available:
-            try:
-                icon = (
-                    AdapterProcessor.get_adapter_data_with_key(
-                        adapter.adapter_id, common.ICON
-                    )
-                    or UNAVAILABLE_ADAPTER_ICON
-                )
-            except Exception as e:
-                logger.warning(f"No icon for adapter {adapter.adapter_id}: {e}")
         try:
             model = adapter.metadata.get("model")
         except Exception as e:
-            logger.warning(f"No metadata for adapter {adapter.adapter_id}: {e}")
+            logger.error(
+                "Could not read metadata for adapter %s: %s", adapter.adapter_id, e
+            )
             model = None
-        return icon, model or adapter.adapter_id.split("|")[0]
+        return model or adapter.adapter_id.split("|")[0]
 
     @staticmethod
     def test_adapter(adapter_id: str, adapter_metadata: dict[str, Any]) -> bool:
--- a/backend/adapter_processor_v2/adapter_processor.py
+++ b/backend/adapter_processor_v2/adapter_processor.py
@@ -122,8 +122,11 @@ class AdapterProcessor:
         icon = UNAVAILABLE_ADAPTER_ICON
         if adapter.is_available:
             try:
-                icon = AdapterProcessor.get_adapter_data_with_key(
-                    adapter.adapter_id, common.ICON
+                icon = (
+                    AdapterProcessor.get_adapter_data_with_key(
+                        adapter.adapter_id, common.ICON
+                    )
+                    or UNAVAILABLE_ADAPTER_ICON
                 )
             except Exception as e:
                 logger.warning(f"No icon for adapter {adapter.adapter_id}: {e}")
--- a/backend/adapter_processor_v2/adapter_processor.py
+++ b/backend/adapter_processor_v2/adapter_processor.py
@@ -20,6 +20,7 @@ from unstract.sdk1.adapters.adapterkit import Adapterkit
 from unstract.sdk1.adapters.base import Adapter
 from unstract.sdk1.adapters.x2text.constants import X2TextConstants
 from unstract.sdk1.constants import AdapterTypes
+from unstract.sdk1.constants import Common as common
 from unstract.sdk1.embedding import EmbeddingCompat
 from unstract.sdk1.exceptions import SdkError
 from unstract.sdk1.llm import LLM
@@ -27,6 +28,8 @@ from unstract.sdk1.llm import LLM
 from .models import AdapterInstance, UserDefaultAdapter
 
 logger = logging.getLogger(__name__)
+
+UNAVAILABLE_ADAPTER_ICON = "⚠️"
 
 try:
     from plugins.subscription.time_trials.subscription_adapter import add_unstract_key
@@ -108,6 +111,28 @@ class AdapterProcessor:
             )
             raise InValidAdapterId()
         return updated_adapters[0].get(key_value)
+
+    @staticmethod
+    def get_display_info(adapter: AdapterInstance) -> tuple[str, str]:
+        """Icon and model label for an adapter, as (icon, model).
+
+        Display data only, no credentials. Never raises - falls back to the
+        warning icon and the adapter id's provider prefix.
+        """
+        icon = UNAVAILABLE_ADAPTER_ICON
+        if adapter.is_available:
+            try:
+                icon = AdapterProcessor.get_adapter_data_with_key(
+                    adapter.adapter_id, common.ICON
+                )
+            except Exception as e:
+                logger.warning(f"No icon for adapter {adapter.adapter_id}: {e}")
+        try:
+            model = adapter.metadata.get("model")
+        except Exception as e:
+            logger.warning(f"No metadata for adapter {adapter.adapter_id}: {e}")
+            model = None
+        return icon, model or adapter.adapter_id.split("|")[0]
 
     @staticmethod
     def test_adapter(adapter_id: str, adapter_metadata: dict[str, Any]) -> bool:
@@ -213,27 +238,6 @@ class AdapterProcessor:
                 raise e
             else:
                 raise InternalServiceError()
-
-    @staticmethod
-    def get_adapter_instance_by_id(adapter_instance_id: str) -> Adapter:
-        """Get the adapter instance by its ID.
-
-        Parameters:
-        - adapter_instance_id (str): The ID of the adapter instance.
-
-        Returns:
-        - Adapter: The adapter instance with the specified ID.
-
-        Raises:
-        - Exception: If there is an error while fetching the adapter instance.
-        """
-        try:
-            adapter = AdapterInstance.objects.get(id=adapter_instance_id)
-        except Exception as e:
-            logger.error(f"Unable to fetch adapter: {e}")
-        if not adapter:
-            logger.error("Unable to fetch adapter")
-        return adapter.adapter_name
 
     @staticmethod
     def get_adapters_by_type(
--- a/backend/adapter_processor_v2/adapter_processor.py
+++ b/backend/adapter_processor_v2/adapter_processor.py
@@ -28,8 +28,6 @@ from .models import AdapterInstance, UserDefaultAdapter
 
 logger = logging.getLogger(__name__)
 
-UNAVAILABLE_ADAPTER_ICON = "⚠️"
-
 try:
     from plugins.subscription.time_trials.subscription_adapter import add_unstract_key
 except ImportError:
@@ -115,7 +113,7 @@ class AdapterProcessor:
     def get_icon(adapter: AdapterInstance) -> str:
         """Registry icon for an adapter, or the warning icon if unresolvable."""
         if not adapter.is_available:
-            return UNAVAILABLE_ADAPTER_ICON
+            return AdapterKeys.UNAVAILABLE_ICON
         try:
             adapter_class = Adapterkit().get_adapter_class_by_adapter_id(
                 adapter.adapter_id
@@ -124,8 +122,8 @@ class AdapterProcessor:
             logger.warning(
                 "Adapter %s is not in the SDK registry: %s", adapter.adapter_id, e
             )
-            return UNAVAILABLE_ADAPTER_ICON
-        return adapter_class.get_icon()
+            return AdapterKeys.UNAVAILABLE_ICON
+        return adapter_class.get_icon() or AdapterKeys.UNAVAILABLE_ICON
 
     @staticmethod
     def get_model_label(adapter: AdapterInstance) -> str:
--- a/backend/adapter_processor_v2/adapter_processor.py
+++ b/backend/adapter_processor_v2/adapter_processor.py
@@ -125,7 +125,7 @@ class AdapterProcessor:
                 "Adapter %s is not in the SDK registry: %s", adapter.adapter_id, e
             )
             return UNAVAILABLE_ADAPTER_ICON
-        return adapter_class.get_icon() or UNAVAILABLE_ADAPTER_ICON
+        return adapter_class.get_icon()
 
     @staticmethod
     def get_model_label(adapter: AdapterInstance) -> str:
--- a/backend/prompt_studio/prompt_profile_manager_v2/serializers.py
+++ b/backend/prompt_studio/prompt_profile_manager_v2/serializers.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Any
 
 from adapter_processor_v2.adapter_processor import AdapterProcessor
 
@@ -9,7 +10,7 @@ from .models import ProfileManager
 
 logger = logging.getLogger(__name__)
 
-# Adapter FK -> label shown on the Prompt Studio output tiles.
+# Adapter FK -> "conf" response key read by the UI.
 ADAPTER_LABELS = (
     (ProfileManagerKeys.LLM, "LLM"),
     (ProfileManagerKeys.EMBEDDING_MODEL, "Embedding Model"),
@@ -22,7 +23,8 @@ class ProfileManagerSerializer(AuditSerializer):
     class Meta:
         model = ProfileManager
         fields = "__all__"
-        # Uniqueness is enforced by the view; the auto-validator 400s on re-save.
+        # Drop DRF's auto unique-together validator so a duplicate create
+        # surfaces the view's DuplicateData instead of DRF's generic message.
         validators = []
 
     def to_representation(self, instance):  # type: ignore
@@ -30,17 +32,16 @@ class ProfileManagerSerializer(AuditSerializer):
 
         Not filtered by adapter access - display data only, no credentials.
         """
-        rep: dict[str, str] = super().to_representation(instance)
+        rep: dict[str, Any] = super().to_representation(instance)
         conf: dict[str, str] = {}
         for field, label in ADAPTER_LABELS:
-            adapter = getattr(instance, field, None)
+            adapter = getattr(instance, field)
             if not adapter:
                 continue
-            icon, model = AdapterProcessor.get_display_info(adapter)
             rep[field] = adapter.adapter_name
-            conf[label] = model
+            conf[label] = AdapterProcessor.get_model_label(adapter)
             if field == ProfileManagerKeys.LLM:
-                rep["icon"] = icon
+                rep["icon"] = AdapterProcessor.get_icon(adapter)
         if conf:
             conf["Profile Name"] = instance.profile_name
         rep["conf"] = conf
--- a/backend/prompt_studio/prompt_profile_manager_v2/serializers.py
+++ b/backend/prompt_studio/prompt_profile_manager_v2/serializers.py
@@ -2,6 +2,8 @@ import logging
 from typing import Any
 
 from adapter_processor_v2.adapter_processor import AdapterProcessor
+from adapter_processor_v2.models import AdapterInstance
+from rest_framework.serializers import ValidationError
 
 from backend.serializers import AuditSerializer
 from prompt_studio.prompt_profile_manager_v2.constants import ProfileManagerKeys
@@ -25,6 +27,24 @@ class ProfileManagerSerializer(AuditSerializer):
         fields = "__all__"
         # Dropped so a duplicate create surfaces the view's DuplicateData.
         validators = []
+
+    def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
+        """Reject a change to an adapter the requester cannot access.
+
+        An unchanged value passes, so a co-owner can still save a profile
+        that points at an adapter shared only with the owner.
+        """
+        request = self.context.get("request")
+        if not request:
+            return attrs
+        accessible = AdapterInstance.objects.for_user(request.user)
+        for field, _ in ADAPTER_LABELS:
+            adapter = attrs.get(field)
+            if not adapter or adapter == getattr(self.instance, field, None):
+                continue
+            if not accessible.filter(id=adapter.id).exists():
+                raise ValidationError({field: "No access to the selected adapter."})
+        return attrs
 
     def to_representation(self, instance):  # type: ignore
         """Resolve the adapter FKs to the name, model and icon the UI renders.
--- a/backend/prompt_studio/prompt_profile_manager_v2/serializers.py
+++ b/backend/prompt_studio/prompt_profile_manager_v2/serializers.py
@@ -23,8 +23,7 @@ class ProfileManagerSerializer(AuditSerializer):
     class Meta:
         model = ProfileManager
         fields = "__all__"
-        # Drop DRF's auto unique-together validator so a duplicate create
-        # surfaces the view's DuplicateData instead of DRF's generic message.
+        # Dropped so a duplicate create surfaces the view's DuplicateData.
         validators = []
 
     def to_representation(self, instance):  # type: ignore
@@ -38,6 +37,8 @@ class ProfileManagerSerializer(AuditSerializer):
             adapter = getattr(instance, field)
             if not adapter:
                 continue
+            # Keep the id for adapters the viewer cannot access.
+            rep[f"{field}_id"] = str(rep[field])
             rep[field] = adapter.adapter_name
             conf[label] = AdapterProcessor.get_model_label(adapter)
             if field == ProfileManagerKeys.LLM:
--- a/backend/prompt_studio/prompt_profile_manager_v2/serializers.py
+++ b/backend/prompt_studio/prompt_profile_manager_v2/serializers.py
@@ -9,33 +9,39 @@ from .models import ProfileManager
 
 logger = logging.getLogger(__name__)
 
+# Adapter FK -> label shown on the Prompt Studio output tiles.
+ADAPTER_LABELS = (
+    (ProfileManagerKeys.LLM, "LLM"),
+    (ProfileManagerKeys.EMBEDDING_MODEL, "Embedding Model"),
+    (ProfileManagerKeys.VECTOR_STORE, "Vector Store"),
+    (ProfileManagerKeys.X2TEXT, "Text Extractor"),
+)
+
 
 class ProfileManagerSerializer(AuditSerializer):
     class Meta:
         model = ProfileManager
         fields = "__all__"
-        # View owns uniqueness (IntegrityError->DuplicateData on create); drop
-        # the DRF auto-validator that 400s on re-save / PUT before the view runs.
+        # Uniqueness is enforced by the view; the auto-validator 400s on re-save.
         validators = []
 
     def to_representation(self, instance):  # type: ignore
+        """Resolve the adapter FKs to the name, model and icon the UI renders.
+
+        Not filtered by adapter access - display data only, no credentials.
+        """
         rep: dict[str, str] = super().to_representation(instance)
-        llm = rep[ProfileManagerKeys.LLM]
-        embedding = rep[ProfileManagerKeys.EMBEDDING_MODEL]
-        vector_db = rep[ProfileManagerKeys.VECTOR_STORE]
-        x2text = rep[ProfileManagerKeys.X2TEXT]
-        if llm:
-            rep[ProfileManagerKeys.LLM] = AdapterProcessor.get_adapter_instance_by_id(llm)
-        if embedding:
-            rep[ProfileManagerKeys.EMBEDDING_MODEL] = (
-                AdapterProcessor.get_adapter_instance_by_id(embedding)
-            )
-        if vector_db:
-            rep[ProfileManagerKeys.VECTOR_STORE] = (
-                AdapterProcessor.get_adapter_instance_by_id(vector_db)
-            )
-        if x2text:
-            rep[ProfileManagerKeys.X2TEXT] = AdapterProcessor.get_adapter_instance_by_id(
-                x2text
-            )
+        conf: dict[str, str] = {}
+        for field, label in ADAPTER_LABELS:
+            adapter = getattr(instance, field, None)
+            if not adapter:
+                continue
+            icon, model = AdapterProcessor.get_display_info(adapter)
+            rep[field] = adapter.adapter_name
+            conf[label] = model
+            if field == ProfileManagerKeys.LLM:
+                rep["icon"] = icon
+        if conf:
+            conf["Profile Name"] = instance.profile_name
+        rep["conf"] = conf
         return rep
--- a/backend/prompt_studio/prompt_profile_manager_v2/tests/test_profile_display_info.py
+++ b/backend/prompt_studio/prompt_profile_manager_v2/tests/test_profile_display_info.py
@@ -1,0 +1,93 @@
+"""Profile serializer resolves adapter FKs to display data without an access check.
+
+The DRF base is patched out so the assertions cover only that resolution.
+"""
+
+from __future__ import annotations
+
+import unittest
+from types import SimpleNamespace
+from unittest.mock import patch
+
+from backend.serializers import AuditSerializer
+
+from prompt_studio.prompt_profile_manager_v2.serializers import ProfileManagerSerializer
+
+
+def _adapter(name: str, model: str) -> SimpleNamespace:
+    return SimpleNamespace(adapter_name=name, model=model)
+
+
+def _represent(instance: SimpleNamespace, base_rep: dict) -> dict:
+    with (
+        patch.object(AuditSerializer, "to_representation", return_value=base_rep),
+        patch(
+            "prompt_studio.prompt_profile_manager_v2.serializers."
+            "AdapterProcessor.get_display_info",
+            side_effect=lambda adapter: ("openai.png", adapter.model),
+        ),
+    ):
+        return ProfileManagerSerializer().to_representation(instance)
+
+
+class ProfileDisplayInfoTests(unittest.TestCase):
+    def test_display_info_resolved_without_adapter_access(self) -> None:
+        instance = SimpleNamespace(
+            profile_name="Prod",
+            llm=_adapter("Shared GPT", "gpt-4o"),
+            embedding_model=_adapter("Shared Embed", "text-embedding-3-small"),
+            vector_store=_adapter("Shared Qdrant", "qdrant"),
+            x2text=_adapter("Shared LLMW", "llmwhisperer"),
+        )
+        base_rep = {
+            field: "some-uuid"
+            for field in ("llm", "embedding_model", "vector_store", "x2text")
+        }
+
+        rep = _represent(instance, base_rep)
+
+        self.assertEqual(
+            rep["conf"],
+            {
+                "LLM": "gpt-4o",
+                "Embedding Model": "text-embedding-3-small",
+                "Vector Store": "qdrant",
+                "Text Extractor": "llmwhisperer",
+                "Profile Name": "Prod",
+            },
+        )
+        # Only the LLM contributes the tile icon.
+        self.assertEqual(rep["icon"], "openai.png")
+        # FK ids are replaced by the adapter names.
+        self.assertEqual(rep["llm"], "Shared GPT")
+
+    def test_unset_adapters_are_skipped(self) -> None:
+        instance = SimpleNamespace(
+            profile_name="Half configured",
+            llm=_adapter("Shared GPT", "gpt-4o"),
+            embedding_model=None,
+            vector_store=None,
+            x2text=None,
+        )
+
+        rep = _represent(instance, {"llm": "some-uuid", "embedding_model": None})
+
+        self.assertEqual(
+            rep["conf"], {"LLM": "gpt-4o", "Profile Name": "Half configured"}
+        )
+        self.assertIsNone(rep["embedding_model"])
+
+    def test_profile_with_no_adapters_has_empty_conf(self) -> None:
+        instance = SimpleNamespace(
+            profile_name="Empty",
+            llm=None,
+            embedding_model=None,
+            vector_store=None,
+            x2text=None,
+        )
+
+        rep = _represent(instance, {})
+
+        # No "Profile Name" either - the tile has nothing to show.
+        self.assertEqual(rep["conf"], {})
+        self.assertNotIn("icon", rep)
--- a/backend/prompt_studio/prompt_profile_manager_v2/tests/test_profile_display_info.py
+++ b/backend/prompt_studio/prompt_profile_manager_v2/tests/test_profile_display_info.py
@@ -23,8 +23,13 @@ def _represent(instance: SimpleNamespace, base_rep: dict) -> dict:
         patch.object(AuditSerializer, "to_representation", return_value=base_rep),
         patch(
             "prompt_studio.prompt_profile_manager_v2.serializers."
-            "AdapterProcessor.get_display_info",
-            side_effect=lambda adapter: ("openai.png", adapter.model),
+            "AdapterProcessor.get_model_label",
+            side_effect=lambda adapter: adapter.model,
+        ),
+        patch(
+            "prompt_studio.prompt_profile_manager_v2.serializers."
+            "AdapterProcessor.get_icon",
+            return_value="/icons/adapter-icons/OpenAI.png",
         ),
     ):
         return ProfileManagerSerializer().to_representation(instance)
@@ -57,7 +62,7 @@ class ProfileDisplayInfoTests(unittest.TestCase):
             },
         )
         # Only the LLM contributes the tile icon.
-        self.assertEqual(rep["icon"], "openai.png")
+        self.assertEqual(rep["icon"], "/icons/adapter-icons/OpenAI.png")
         # FK ids are replaced by the adapter names.
         self.assertEqual(rep["llm"], "Shared GPT")
 
--- a/backend/prompt_studio/prompt_profile_manager_v2/views.py
+++ b/backend/prompt_studio/prompt_profile_manager_v2/views.py
@@ -36,7 +36,10 @@ class ProfileManagerView(viewsets.ModelViewSet):
         return [IsOwnerOrSharedUserOrSharedToOrg()]
 
     def get_queryset(self) -> QuerySet | None:
-        queryset = ProfileManager.objects.for_user(self.request.user)
+        # Serializer reads all four adapters per profile for the display info
+        queryset = ProfileManager.objects.for_user(self.request.user).select_related(
+            "llm", "embedding_model", "vector_store", "x2text"
+        )
         filter_args = FilterHelper.build_filter_args(
             self.request,
             ProfileManagerKeys.CREATED_BY,
--- a/backend/prompt_studio/prompt_profile_manager_v2/views.py
+++ b/backend/prompt_studio/prompt_profile_manager_v2/views.py
@@ -36,7 +36,7 @@ class ProfileManagerView(viewsets.ModelViewSet):
         return [IsOwnerOrSharedUserOrSharedToOrg()]
 
     def get_queryset(self) -> QuerySet | None:
-        # Serializer reads all four adapters per profile for the display info
+        # Serializer reads all four adapters for the display info
         queryset = ProfileManager.objects.for_user(self.request.user).select_related(
             "llm", "embedding_model", "vector_store", "x2text"
         )
--- a/frontend/src/components/custom-tools/prompt-card/PromptCardItems.jsx
+++ b/frontend/src/components/custom-tools/prompt-card/PromptCardItems.jsx
@@ -147,8 +147,7 @@ function PromptCardItems({
     if (isSimplePromptStudio) {
       return;
     }
-    // conf/icon come off the profile payload; the viewer may not have
-    // access to the project's adapters.
+    // conf and icon come from the profile payload, not the viewer's adapters.
     setLlmProfileDetails(
       (llmProfiles || [])
         .map((profile) => ({
--- a/frontend/src/components/custom-tools/prompt-card/PromptCardItems.jsx
+++ b/frontend/src/components/custom-tools/prompt-card/PromptCardItems.jsx
@@ -83,7 +83,6 @@ function PromptCardItems({
     indexDocs,
     isSimplePromptStudio,
     isPublicSource,
-    adapters,
     selectedHighlight,
     details,
     singlePassExtractMode,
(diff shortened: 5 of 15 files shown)


backend/adapter_processor_v2/adapter_processor.py
```diff
@@ -109,6 +112,28 @@ def get_adapter_data_with_key(adapter_id: str, key_value: str) -> Any:
...
+
+        Display data only, no credentials. Never raises - falls back to the
+        warning icon and the adapter id's provider prefix.
+        """
+        icon = UNAVAILABLE_ADAPTER_ICON
+        if adapter.is_available:
+            try:
+                icon = AdapterProcessor.get_adapter_data_with_key(
+                    adapter.adapter_id, common.ICON
+                )
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Valid. `get_adapter_data_with_key` returns the registry entry's value, so an adapter whose entry has no `icon` key yields `None` without raising — the `except` never fires and `None` replaced the fallback. Guarded with `or UNAVAILABLE_ADAPTER_ICON` in b30385de8.

backend/adapter_processor_v2/adapter_processor.py
```diff
@@ -109,6 +112,31 @@ def get_adapter_data_with_key(adapter_id: str, key_value: str) -> Any:
...
+        """
+        icon = UNAVAILABLE_ADAPTER_ICON
+        if adapter.is_available:
+            try:
+                icon = (
+                    AdapterProcessor.get_adapter_data_with_key(
+                        adapter.adapter_id, common.ICON
+                    )
+                    or UNAVAILABLE_ADAPTER_ICON
+                )
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Valid on the rendering — both output components pass `profile.icon` straight to antd `Image` as `src`, so an emoji loads as a URL and shows a broken image. Worth noting it is not new here: `AdapterListSerializer` already returns `⚠️` for unavailable adapters and the pre-change frontend fed that same value into the same `Image`.

Fixed in ca031c15b, following the emoji-vs-URL split the codebase already uses in `ResourceTable.jsx` — added `isImageUrl()` to `GetStaticData.js` and render the fallback as text when the icon is not a URL.

vishnuszipstack (commented):
## Automated review — PR Review Toolkit (6 agents)

Ran code review, silent-failure, type-design, test-coverage, comment-accuracy and simplification passes over the 9 changed files, then verified each finding against the actual code before posting. Skipped anything already covered in the existing `adapter_processor.py:127/:130` threads (the `or UNAVAILABLE_ADAPTER_ICON` guard and the emoji-as-`` fix) — both look correctly resolved.

**The core approach is right.** Moving model/icon resolution server-side is the correct fix for a shared-project viewer: no client-side join can produce a name for an adapter the viewer can't list. `select_related` is the right instinct, the frontend cleanup is complete (`adapters` came from the store, not props — no stale PropTypes or dead prop-drilling left behind), and deleting `get_adapter_instance_by_id` is safe (zero references repo-wide, and the `Adapter` import it used is still needed at line 192).

Three findings are worth blocking on:

**1. The `select_related` is on a queryset that never lists.** `ProfileManagerView` has no `list` route — `prompt_profile_manager_v2/urls.py` wires only `<uuid:pk>/`. The endpoint that actually loads `llmProfiles` is `PromptStudioCoreView.list_profiles` (`prompt_studio_core_v2/views.py:376`), which has no `select_related` and uses the same serializer with `many=True`. The N+1 is unmitigated on the one path the feature exists for.

**2. Each icon lookup rescans the whole SDK registry from disk, and 3 of 4 results are discarded.** `get_adapter_data_with_key` → `Adapterkit().get_adapters_list()`, which is not memoised and calls `get_json_schema()` — an `open().read()` — for all 37 registered adapters. `get_display_info` resolves the icon for every adapter but the serializer keeps only the LLM's. A 4-profile project is ~590 file opens and 16 `INFO` log lines per profile-list request, ~75% of it wasted. Splitting `get_display_info` into `get_icon()` / `get_model_label()` fixes this *and* the positional-tuple hazard in one move.

**3. The fix is partial.** `getLLMModelNamesForProfiles` (`GetStaticData.js:499`) is the same client-side join just deleted from `PromptCardItems.jsx`, still live at `CombinedOutput.jsx:151` and `OutputForDocModal.jsx:186`. Both fetch the *viewer's* `/adapter/?adapter_type=LLM` list, so the Combined Output view and the per-document output modal still show a blank model name for shared users. Fine to defer, but the PR description currently reads as though the bug is closed.

Beyond those: `InvalidEncryptionKey` — a documented platform-wide outage — is swallowed into a `logger.warning` on the one path a shared viewer can see, and the new test mocks out `get_display_info`, so both of the bug fixes in commits `b30385de` and `ca031c15` have zero coverage. Details inline.

Two notes on things I checked that are **not** problems: the `timeAgo` export move is Biome's `organizeImports` re-sorting the block, not noise — reverting it would fail `lint:changed`. And `conf["Profile Name"]` being gated on `conf` being non-empty faithfully reproduces the old frontend behaviour (the assignment sat inside `if (adapter)` on main).

One pre-existing quirk worth a comment rather than a change: `serializers.py:40` replaces the FK UUID with `adapter_name`, so a GET response can't be PUT back. That's load-bearing — `AddLlmProfile.jsx:129` matches the edit form by `item?.label === llmProfileDetails?.llm` — and predates this PR. Worth a line so nobody "fixes" it later.

backend/prompt_studio/prompt_profile_manager_v2/views.py
```diff
@@ -36,7 +36,10 @@ def get_permissions(self) -> list[Any]:
         return [IsOwnerOrSharedUserOrSharedToOrg()]
 
     def get_queryset(self) -> QuerySet | None:
-        queryset = ProfileManager.objects.for_user(self.request.user)
+        # Serializer reads all four adapters per profile for the display info
+        queryset = ProfileManager.objects.for_user(self.request.user).select_related(
```
vishnuszipstack, line 40: **P1 — the `select_related` landed on a queryset that never serializes more than one profile.**

`ProfileManagerView` has no `list` route. `prompt_profile_manager_v2/urls.py` wires only `profile-manager/<uuid:pk>/` → `retrieve/update/partial_update/destroy`, so this queryset only ever yields a single profile and the join saves at most 3 queries.

The endpoint that actually populates `llmProfiles` — the one this feature exists for — is `GET prompt-studio/prompt-studio-profile/<pk>/` (`frontend/src/components/helpers/custom-tools/CustomToolsHelper.js:102` → `prompt_studio_core_v2/urls.py:96`), which lands on `PromptStudioCoreView.list_profiles` at `backend/prompt_studio/prompt_studio_core_v2/views.py:376`:

```python
profile_manager_instances = ProfileManager.objects.filter(
    prompt_studio_tool=prompt_tool
)
serialized_instances = ProfileManagerSerializer(
    profile_manager_instances, many=True
).data
```

No `select_related`, same serializer, `many=True`. Since `to_representation` now dereferences all four FKs per row, that path does 4 lazy FK queries per profile — the exact N+1 this change was meant to prevent, on the hot path, unmitigated.

Suggested fix — add the same join in `list_profiles`:

```python
profile_manager_instances = ProfileManager.objects.filter(
    prompt_studio_tool=prompt_tool
).select_related("llm", "embedding_model", "vector_store", "x2text")
```

Worth keeping the join here too (it still helps `retrieve`), but the comment above it implies it protects a list read that this view doesn't serve.

backend/prompt_studio/prompt_profile_manager_v2/serializers.py
```diff
@@ -9,33 +9,39 @@
...
-        if x2text:
-            rep[ProfileManagerKeys.X2TEXT] = AdapterProcessor.get_adapter_instance_by_id(
-                x2text
-            )
+        conf: dict[str, str] = {}
+        for field, label in ADAPTER_LABELS:
+            adapter = getattr(instance, field, None)
+            if not adapter:
+                continue
+            icon, model = AdapterProcessor.get_display_info(adapter)
```
vishnuszipstack: **P1 — the icon is resolved for all four adapters and kept for one, and each resolution rescans the entire SDK registry from disk.**

`rep["icon"]` is only set in the `llm` branch below, but `get_display_info` resolves the icon unconditionally. The discarded work is not cheap:

`get_display_info` → `get_adapter_data_with_key` → `__fetch_adapters_by_key_value` (`adapter_processor.py:196-199`) → `Adapterkit().get_adapters_list()`.

`Adapterkit` is a `@singleton`, but `get_adapters_list()` is **not** memoised (`unstract/sdk1/src/unstract/sdk1/adapters/adapterkit.py:64-86`) — it loops every registered adapter and calls `m.get_json_schema()`, which is a file read:

```python
with open(schema_path) as f:
    return f.read()
```

There are 37 adapter schema files in the registry. So each icon lookup is a full registry scan with ~37 `open()` calls, plus an `INFO` log line (`adapter_processor.py:196`). For a 4-profile project that is 16 lookups ≈ 590 file opens and 16 INFO lines per profile-list request — and 12 of those 16 lookups have their result thrown away.

Two independent fixes, both small:

1. Only resolve the icon where it's used:
```python
for field, label in ADAPTER_LABELS:
    adapter = getattr(instance, field)
    conf[label] = AdapterProcessor.get_model_label(adapter)
    rep[field] = adapter.adapter_name
if instance.llm_id:
    rep["icon"] = AdapterProcessor.get_icon(instance.llm)
```

2. Make the icon lookup O(1) — `Adapterkit` already exposes a direct dict route that avoids `get_adapters_list()` entirely:
```python
icon = Adapterkit().get_adapter_class_by_adapter_id(adapter.adapter_id).get_icon()
```
Same underlying `get_icon()` value, no registry scan, no file I/O. It raises `RuntimeError` instead of `InValidAdapterId` for an unknown id, which the existing `except Exception` already covers.

frontend/src/components/custom-tools/prompt-card/PromptCardItems.jsx
```diff
@@ -166,18 +139,18 @@ function PromptCardItems({
...
-
-    // Update llmProfiles with additional fields
-    const updatedProfiles = llmProfiles?.map((profile) => {
-      return { ...getModelOrAdapterId(profile, adapterData), ...profile };
-    });
+    // conf/icon come off the profile payload; the viewer may not have
+    // access to the project's adapters.
     setLlmProfileDetails(
-      updatedProfiles
+      (llmProfiles || [])
```
vishnuszipstack, line 152: **P1 — the fix is partial; two other output surfaces still resolve model names from the viewer's own adapter list.**

Removing the client-side lookup here is right, but the identical pattern survives in two places, both routed through `getLLMModelNamesForProfiles` (`frontend/src/helpers/GetStaticData.js:499-514`), which maps `profile.llm` → `adapter.model` using the *caller's* adapter list:

- `frontend/src/components/custom-tools/combined-output/CombinedOutput.jsx:145-151`
- `frontend/src/components/custom-tools/output-for-doc-modal/OutputForDocModal.jsx:178-186`

Both fetch `/api/v1/unstract/{org}/adapter/?adapter_type=LLM`, which returns only the adapters the *viewer* owns — empty for a shared-project user. So the Combined Output view and the per-document output modal still show a blank model name for exactly the users this PR targets.

Now that the backend ships `conf.LLM` on every profile, both call sites can read `profile?.conf?.LLM` directly, which also deletes two adapter API round-trips and lets `getLLMModelNamesForProfiles` go away.

If that's intentionally out of scope, worth saying so in the PR description — the "Can this PR break any existing features" section currently reads as though the fix is complete.

backend/adapter_processor_v2/adapter_processor.py
```diff
@@ -109,6 +112,31 @@ def get_adapter_data_with_key(adapter_id: str, key_value: str) -> Any:
...
+            except Exception as e:
+                logger.warning(f"No icon for adapter {adapter.adapter_id}: {e}")
+        try:
+            model = adapter.metadata.get("model")
+        except Exception as e:
```
vishnuszipstack, line 136: **P2 — this `except Exception` swallows `InvalidEncryptionKey`, which the codebase treats as a call-the-admin outage.**

`adapter.metadata` is not a plain attribute — it's a decrypting property (`adapter_processor_v2/models.py:190-201`) that converts a Fernet `InvalidToken` into `InvalidEncryptionKey`. That exception is a 403 `APIException` whose detail reads *"Platform encryption key for storing adapter credentials has changed! All adapters are inaccessible. Please inform the platform admin immediately."* (`backend/utils/exceptions.py:6-12`).

Caught here, it becomes a `logger.warning` and a tile that renders a plausible-looking `"openai"`. Note the asymmetry: `AdapterListSerializer` does the same `instance.metadata.get("model")` **unguarded** at `adapter_processor_v2/serializers.py:206`, so the adapters page surfaces the actionable 403 while the profile payload silently degrades — and for a shared-project viewer the profile payload is the *only* path, so they get no signal at all.

The same handler also absorbs several unrelated failures, all reachable, none of which "No metadata for adapter" describes:
- `TypeError: cannot convert 'NoneType' object to bytes` — `adapter_metadata_b` is `BinaryField(null=True)` (`models.py:89`) and is only populated when metadata is truthy (`serializers.py:73-80`).
- `ValueError` from `Fernet(...)` on a malformed `ENCRYPTION_KEY`.
- `json.JSONDecodeError` on corrupt plaintext.

Suggested fix — let the outage propagate, narrow the rest, and log at `error`:

```python
try:
    model = adapter.metadata.get("model")
except InvalidEncryptionKey:
    raise  # platform-wide credential outage; must not degrade to a log line
except (TypeError, ValueError, AttributeError) as e:
    logger.error(
        "Unreadable metadata for adapter %s (%s): %s",
        adapter.id, adapter.adapter_id, e,
    )
    model = None
```

If a 403 on the profile list is considered too disruptive for a display path, that's a defensible call — but then it should be a deliberate one, with the log at `error` and a message that names the real condition rather than "No metadata".

backend/prompt_studio/prompt_profile_manager_v2/serializers.py
```diff
@@ -9,33 +9,39 @@
...
-            rep[ProfileManagerKeys.VECTOR_STORE] = (
-                AdapterProcessor.get_adapter_instance_by_id(vector_db)
-            )
-        if x2text:
-            rep[ProfileManagerKeys.X2TEXT] = AdapterProcessor.get_adapter_instance_by_id(
-                x2text
-            )
+        conf: dict[str, str] = {}
+        for field, label in ADAPTER_LABELS:
+            adapter = getattr(instance, field, None)
```
vishnuszipstack: **P2 — `getattr(instance, field, None)` silently swallows a broken FK, and the `continue` branch is otherwise unreachable.**

All four adapter FKs are `null=False, blank=False, on_delete=models.PROTECT` (`prompt_profile_manager_v2/models.py:61-91`). A `None` is therefore impossible under the schema — so `if not adapter: continue` can only fire in the one case that must *not* be silent.

Django's forward FK descriptor raises `RelatedObjectDoesNotExist`, which is constructed as a subclass of **both** `model.DoesNotExist` and `AttributeError` (`django/db/models/fields/related_descriptors.py`). Because it inherits `AttributeError`, the three-argument `getattr` catches it and returns the default. A dangling FK — raw SQL delete, partial restore, cross-schema drift — therefore produces no exception, no log, and no user-visible error: just a row quietly missing from the tooltip, and a blank icon and title if it's the `llm` FK.

Suggested fix — be explicit about the impossible case:

```python
from django.core.exceptions import ObjectDoesNotExist

for field, label in ADAPTER_LABELS:
    try:
        adapter = getattr(instance, field)
    except ObjectDoesNotExist:
        logger.error(
            "Profile %s references a missing %s adapter (%s)",
            instance.profile_id, field, getattr(instance, f"{field}_id", None),
        )
        continue
```

`ObjectDoesNotExist` catches the real condition without the `AttributeError` blanket, so a genuine typo in `ADAPTER_LABELS` still fails loudly.

Related: `test_unset_adapters_are_skipped` and `test_profile_with_no_adapters_has_empty_conf` both assert on FK states the schema forbids, so neither covers this.

backend/prompt_studio/prompt_profile_manager_v2/tests/test_profile_display_info.py
```diff
@@ -0,0 +1,93 @@
...
+    return SimpleNamespace(adapter_name=name, model=model)
+
+
+def _represent(instance: SimpleNamespace, base_rep: dict) -> dict:
+    with (
+        patch.object(AuditSerializer, "to_representation", return_value=base_rep),
+        patch(
+            "prompt_studio.prompt_profile_manager_v2.serializers."
+            "AdapterProcessor.get_display_info",
+            side_effect=lambda adapter: ("openai.png", adapter.model),
```
vishnuszipstack: **P2 — the test mocks out the method it's named after, so none of the new logic is covered.**

`_represent` patches `AdapterProcessor.get_display_info` wholesale, so the assertions only exercise the four-line loop in `to_representation`. Every genuinely new and failure-prone branch in `get_display_info` (`adapter_processor.py:115-138`) has zero coverage:

| Branch | Silent failure if it regresses |
|---|---|
| `is_available` gate | deprecated adapter raises `InValidAdapterId` into the list endpoint |
| `or UNAVAILABLE_ADAPTER_ICON` | **this is commit b30385de's entire fix** → `icon: None` |
| `except` around `get_adapter_data_with_key` | exception escapes → 500 on profile list |
| `except` around `adapter.metadata` | see the encryption-key comment above |
| `model or adapter_id.split("|")[0]` | tile shows `"openai"` instead of `"gpt-4o"` |

The mock also diverges from the real contract: the fake adapter exposes `adapter.model`, but production reads `adapter.metadata.get("model")` — a decrypting property. Mocking at the `get_display_info` seam is a defensible boundary for a *serializer* test, but it means no test anywhere touches the real attribute path.

`get_display_info` needs no Django — a `SimpleNamespace` plus one `patch` of `get_adapter_data_with_key` covers it. Suggested `backend/adapter_processor_v2/tests/test_adapter_display_info.py`:

```python
@pytest.mark.parametrize("is_available,registry_icon,metadata,expected", [
    (True,  "/icons/adapter-icons/OpenAI.png", {"model": "gpt-4o"}, ("/icons/adapter-icons/OpenAI.png", "gpt-4o")),
    (True,  None,                              {"model": "gpt-4o"}, ("⚠️", "gpt-4o")),        # b30385de
    (False, "/icons/adapter-icons/OpenAI.png", {"model": "gpt-4o"}, ("⚠️", "gpt-4o")),        # deprecated
    (True,  "/icons/adapter-icons/Qdrant.png", {},                  ("/icons/adapter-icons/Qdrant.png", "qdrant")),
])
```
plus `assert_not_called()` on `get_adapter_data_with_key` when `is_available` is `False` — that assertion is the only thing that pins the gate.

One more worth adding, given this PR's whole premise is exposing this payload to users without adapter access: seed the fake metadata with `{"model": "gpt-4o", "api_key": "sk-SECRET"}` and assert `"sk-SECRET" not in json.dumps(rep)`. One line, and it permanently pins the security claim in the docstring. There's precedent at `backend/mcp_server/tests/test_no_credential_leak.py`.

Two smaller notes:
- `patch(...)` here has no `autospec=True`. A rename is caught, but a signature change (`get_display_info(adapter, *, want_icon=True)`) would sail through green while production callers break.
- The icon fixture is `"openai.png"`, which `isImageUrl` classifies as **not** a URL (no leading `/`). Real registry icons are `/icons/adapter-icons/*.png`. Harmless here since the value is never rendered, but it's a misleading shape to copy.

backend/prompt_studio/prompt_profile_manager_v2/serializers.py
```diff
@@ -9,33 +9,39 @@
...
+)
+
 
 class ProfileManagerSerializer(AuditSerializer):
     class Meta:
         model = ProfileManager
         fields = "__all__"
-        # View owns uniqueness (IntegrityError->DuplicateData on create); drop
-        # the DRF auto-validator that 400s on re-save / PUT before the view runs.
+        # Uniqueness is enforced by the view; the auto-validator 400s on re-save.
```
vishnuszipstack: **P2 — this rewrite dropped the load-bearing half of the comment, and the shorter version is now inaccurate.**

Before: `View owns uniqueness (IntegrityError->DuplicateData on create); drop the DRF auto-validator that 400s on re-save / PUT before the view runs.`
After: `Uniqueness is enforced by the view; the auto-validator 400s on re-save.`

Three things were lost, and one inaccuracy was introduced:

1. **"on create" scoping is gone, and the replacement over-claims.** Only the create paths catch `IntegrityError` — `views.py:58-61` and `prompt_studio_core_v2/views.py:971-984`. `ProfileManagerView` inherits DRF's default `update`/`partial_update` with no `IntegrityError` handling, so on a PUT/PATCH name collision nothing in "the view" enforces uniqueness — it surfaces as an uncaught DB error. "Uniqueness is enforced by the view" is simply not true for updates.
2. **"before the view runs" is gone** — and that clause *is* the justification. Without it, "the auto-validator 400s on re-save" reads as a description of desirable behaviour, leaving the next reader to wonder why `validators = []` is there at all.
3. The DRF-version dependency is now unrecoverable: this only matters on DRF 3.15+, which auto-generates a `UniqueTogetherValidator` from `Meta.constraints` (`models.py:153-158`; DRF pinned at 3.17.1 in `backend/pyproject.toml:22`).

`validators = []` is the kind of line a future maintainer deletes as dead config. The original comment was what stopped that. Suggest restoring it — this is a case where shorter is strictly worse.

backend/adapter_processor_v2/adapter_processor.py
```diff
@@ -109,6 +112,31 @@ def get_adapter_data_with_key(adapter_id: str, key_value: str) -> Any:
             raise InValidAdapterId()
         return updated_adapters[0].get(key_value)
 
+    @staticmethod
+    def get_display_info(adapter: AdapterInstance) -> tuple[str, str]:
+        """Icon and model label for an adapter, as (icon, model).
+
+        Display data only, no credentials. Never raises - falls back to the
```
vishnuszipstack: **P3 — three claims in this docstring aren't quite what the code does.** (Separate from the icon-fallback thread above — this is about the contract text itself.)

**"Never raises"** — true today, but only by accident. `adapter.is_available` (line 123) and the whole return expression (line 138) sit outside both `try` blocks. The final `adapter.adapter_id.split("|")[0]` can't `IndexError` only because `adapter_id` is a non-null `CharField(default="")` (`models.py:75-79`) — nothing enforces that. This matters because the caller (`serializers.py:39`) runs inside `to_representation` with `many=True` and no guard, so any future raise here becomes a 500 on the whole profile list. Either weaken the claim to "best-effort" or move the return inside the guarded region so it's structurally true.

**"Display data only, no credentials"** — accurate about the *return value*, but it reads as "this function doesn't touch credentials", which is the opposite of what line 134 does: `adapter.metadata` Fernet-decrypts the full blob, API keys included, and only `model` escapes. Suggest making the mechanism explicit, e.g. *"Reads (and therefore decrypts) `adapter.metadata` to pull the model name; only the model string is returned — never credentials."*

**"falls back to … the adapter id's provider prefix"** — frames the normal path as an error path. The prefix is returned whenever metadata has no `model` key, which is the *expected* case for vector DBs and text extractors (Qdrant metadata is url/api_key; there is no model). A reader seeing `qdrant` on a tile will hunt for a bug that isn't there. Suggest: *"Returns the metadata `model` where the adapter has one (LLM/embedding); vector DB and x2text adapters have none, so the adapter id's provider prefix is used."*

Related, on the log message at line 132: `"No icon for adapter {id}"` fires when `get_adapter_data_with_key` raises, which means `InValidAdapterId` — *adapter not in the SDK registry*, not *adapter has no icon*. The genuine no-icon case is handled by the `or` on line 129 and logs nothing. Since `is_available` was already checked, reaching that `except` means `is_available=True` while the SDK no longer ships the adapter — i.e. the flag is stale (it's only written by migration `0003` and the `manage_deprecated_adapters` command, so it goes stale on every SDK upgrade). That's a real drift signal currently described as a missing icon. `get_adapter_data_with_key` also already logs the same event at line 108, so this line is a duplicate as well as a misnomer.

frontend/src/components/custom-tools/prompt-card/PromptOutput.jsx
```diff
@@ -355,13 +356,19 @@ function PromptOutput({
...
+                      {isImageUrl(profile?.icon) ? (
+                        <Image
+                          src={profile?.icon}
+                          width={15}
+                          height={15}
+                          preview={false}
+                          className="prompt-card-llm-icon"
+                        />
+                      ) : (
+                        <span className="prompt-card-llm-icon">
```
vishnuszipstack: **P3 — when `icon` is absent this renders a zero-width empty span with no placeholder and nothing in the console.**

`isImageUrl(undefined)` is `false`, so an undefined icon takes the `<span>` branch and renders nothing at all. Combined with `{profile?.conf?.LLM}` two lines down also being empty, the tile renders completely blank rather than indicating anything is wrong.

The backend always sets `rep["icon"]` when the `llm` FK resolves, so this isn't reachable today — but it becomes reachable via the dangling-FK path flagged on `serializers.py:36`, and via any stale cached profile payload predating this shape.

One-character fix, and it reuses the fallback the backend already picked:

```jsx
<span className="prompt-card-llm-icon">{profile?.icon || "⚠️"}</span>
```

frontend/src/components/custom-tools/prompt-card/PromptOutputsModal.jsx
```diff
@@ -68,13 +69,19 @@ function PromptOutputsModal({
...
-                          height={15}
-                          preview={false}
-                          className="prompt-card-llm-icon"
-                        />
+                        {isImageUrl(profile?.icon) ? (
+                          <Image
+                            src={profile?.icon}
+                            width={15}
+                            height={15}
+                            preview={false}
```
vishnuszipstack: **P3 — this 13-line block is byte-identical to `PromptOutput.jsx:359-371` apart from indentation.**

Worth extracting while it's only two copies — the `width={15} height={15} preview={false}` triple has to stay in sync with `.prompt-card-llm-icon` in `PromptCard.css:150`, which is exactly the kind of thing that drifts.

`frontend/src/components/widgets/` isn't the right home (everything there is generic — `space-wrapper`, `spinner-loader`, `empty-state`); this is coupled to a prompt-card CSS class. But `prompt-card/` already holds eight single-purpose leaf components (`CopyPromptOutputBtn.jsx`, `ExpandCardBtn.jsx`, `PromptRunCost.jsx`, …), so a new file there is conventional.

```jsx
// prompt-card/ProfileIcon.jsx
function ProfileIcon({ icon }) {
  if (!isImageUrl(icon)) {
    return <span className="prompt-card-llm-icon">{icon || "⚠️"}</span>;
  }
  return (
    <Image src={icon} width={15} height={15} preview={false}
           className="prompt-card-llm-icon" />
  );
}
```

Both call sites collapse to `<ProfileIcon icon={profile?.icon} />`, both files drop their now-unused `Image` and `isImageUrl` imports, and the fallback above only has to be fixed once.

frontend/src/helpers/GetStaticData.js
```diff
@@ -476,6 +476,12 @@ const isNonNegativeNumber = (value) => {
   return typeof value === "number" && !isNaN(value) && value >= 0;
 };
 
+// Icons are image URLs, except the emoji fallbacks used for unavailable
+// adapters. Detect the URL rather than the emoji - compound (ZWJ) emoji break
+// length heuristics.
+const isImageUrl = (value) =>
```
vishnuszipstack, line 481: **P3 — `isImageUrl` is the whole of commit ca031c15 and has no test, though the repo is set up for one.**

The frontend has vitest 3.2.6 + happy-dom wired (`frontend/vitest.config.mjs`, `"test": "vitest"`), and there are already colocated helper tests in this exact directory — `pagination.test.js` and `requestId.test.js`. So this is a cheap, convention-following gap.

The implementation is correct for the real data: every `get_icon()` in the SDK registry returns a root-relative `/icons/adapter-icons/*.png`, which the `^\/` branch matches, and `⚠️` correctly falls through. But that correctness rests entirely on an implicit cross-service contract with the SDK, which is precisely what a regression test is for.

```js
it.each([
  ["/icons/adapter-icons/OpenAI.png", true],
  ["https://cdn.example.com/x.png", true],
  ["data:image/png;base64,iVBOR", true],
  ["⚠️", false],        // must never regress into <img src>
  ["OpenAI.png", false],  // pins that a bare filename is not a URL
  [undefined, false], [null, false], [123, false],
])("isImageUrl(%p) === %p", (v, expected) => expect(isImageUrl(v)).toBe(expected));
```

The `"⚠️"` row is the valuable one — it prevents the emoji regressing back into an ``, which is the exact bug this commit fixed.

Caveat worth flagging separately: no workflow currently runs vitest (`ci-frontend-lint.yaml` only runs Biome, and there's no frontend group in `tests/groups.yaml`), so a test added here wouldn't be CI-gated today. That's a pre-existing infra gap, not a reason to skip the test.

Minor, on the comment above: "Icons are image URLs" — in production they're all root-relative *paths*; the `https?://` and `data:image/` branches match nothing today. And the emoji fallback fires not only for unavailable adapters but also when the SDK lookup fails or returns empty (`adapter_processor.py:125-132`), so "unavailable **or unresolvable** adapters" would be more accurate.

backend/prompt_studio/prompt_profile_manager_v2/serializers.py
```diff
@@ -9,33 +9,39 @@
...
+            if not adapter:
+                continue
+            icon, model = AdapterProcessor.get_display_info(adapter)
+            rep[field] = adapter.adapter_name
+            conf[label] = model
+            if field == ProfileManagerKeys.LLM:
+                rep["icon"] = icon
+        if conf:
+            conf["Profile Name"] = instance.profile_name
+        rep["conf"] = conf
```
vishnuszipstack, line 68: **P3 — `conf` is now a load-bearing wire contract with no schema and a client that can't tolerate its absence.**

Two things worth pinning down:

**1. The frontend dereferences `conf` unguarded.** `PromptOutput.jsx:132-140` does `Object.entries(adapterConf)?.map(...)`, called with `profile?.conf` at line 381. `Object.entries(undefined)` throws — the `?.` after the call is useless. Before this PR the client guaranteed `conf` existed because `getModelOrAdapterId` always returned `{ conf: {} }` locally; now its provenance is entirely remote. The backend does always set it, so this isn't broken today, but the guard was removed at the same moment the value became remote, which is the wrong direction. `Object.entries(adapterConf || {})` costs nothing.

**2. Display labels are being used as payload keys.** `"LLM"`, `"Embedding Model"`, `"Vector Store"`, `"Text Extractor"` are human-readable strings, and `"LLM"` is indexed *literally* by the client at `PromptOutput.jsx:374,376` and `PromptOutputsModal.jsx:86` (`profile?.conf?.LLM`). The other three only reach a generic `Object.entries` tooltip, so they really are display text — but nothing distinguishes them. The comment on `ADAPTER_LABELS` at line 12 ("label shown on the Prompt Studio output tiles") actively invites someone to rename `"LLM"` → `"Model"` as a copy change, which would silently blank the tile title.

Also mixing `"Profile Name"` — a profile attribute, not an adapter — into the same flat namespace means the tooltip and the adapter set share one dict, so a future adapter type called "Profile Name" would collide.

Minimum: tighten the comment to say these are response keys and that `"LLM"` is read by key. Better, if you're willing: key `conf` by the FK name (`llm`, `embedding_model`, …) and let the frontend own the display strings, which is where they belong.

Small accuracy note: `rep: dict[str, str]` on line 33 is no longer true — `conf` is a nested `dict`, and `super().to_representation()` returns plenty of non-`str` values (UUIDs, ints, bools). `dict[str, Any]` is honest.

backend/adapter_processor_v2/adapter_processor.py
```diff
@@ -109,6 +112,31 @@ def get_adapter_data_with_key(adapter_id: str, key_value: str) -> Any:
             raise InValidAdapterId()
         return updated_adapters[0].get(key_value)
 
+    @staticmethod
+    def get_display_info(adapter: AdapterInstance) -> tuple[str, str]:
```
vishnuszipstack: **P3 (type design) — `tuple[str, str]` is the weakest available encoding for `(icon, model)`, and splitting the function removes the hazard for free.**

Both elements are `str`, so `model, icon = AdapterProcessor.get_display_info(adapter)` type-checks and silently ships an emoji as the model name. Only the docstring says which slot is which.

Normally I'd weigh that against a `NamedTuple` and conclude it isn't worth it — this repo has 34 bare `-> tuple[...]` returns against exactly one production `NamedTuple`, and `adapter_processor_v2/` has zero dataclasses, so a 2-tuple is locally idiomatic and there's only one call site. But here a split is *smaller* than a `NamedTuple` and also fixes the wasted-work problem flagged on `serializers.py:39`:

```python
@staticmethod
def get_model_label(adapter: AdapterInstance) -> str: ...  # metadata["model"] or provider prefix

@staticmethod
def get_icon(adapter: AdapterInstance) -> str: ...         # registry icon or UNAVAILABLE_ADAPTER_ICON
```

Two named single-value returns can't be swapped, need no new type, and let the caller resolve the icon only for the LLM — cutting registry walks from 4 per profile to 1.

Two smaller accuracy notes on the annotation itself:
- `str` isn't actually enforced on either side. `get_adapter_data_with_key` returns `Any` (line 87) and `.metadata` returns `Any` (`models.py:191`), so both values are unchecked. It holds in practice (`model` is `"type": "string"` in all 37 adapter schemas), but mypy can't help — `[tool.mypy] strict = true` is set in `pyproject.toml`, yet the mypy pre-commit hook is commented out and no workflow invokes it.
- The "always non-empty" invariant has one reachable hole: `adapter_id` is `CharField(default="")` (`models.py:75-79`), so `"".split("|")[0]` is `""`. An adapter with no `model` in metadata and an empty `adapter_id` returns `("⚠️", "")` — satisfies `str`, renders as a blank tile, logs nothing.

backend/prompt_studio/prompt_profile_manager_v2/views.py
kirtimanmishrazipstack: @vishnuszipstack Right about the routing - no list route, and list_profiles is the endpoint that actually feeds llmProfiles. Added the same join there in 2ffdc9e06.

Two corrections though. It isn't P1: MAX_PROFILE_COUNT is 4 and it's enforced server-side (create_profile_manager counts first), so the worst case is 16 indexed PK lookups, not an open N+1. And it isn't a regression - the old code called get_adapter_instance_by_id per FK, which was the same 4 queries per profile on the same path. Also the detail join saves 4, not 3.

Keeping the join on the detail view: update/partial_update resolve through get_queryset too and re-serialize in their responses.

backend/prompt_studio/prompt_profile_manager_v2/serializers.py
kirtimanmishrazipstack: @vishnuszipstack Confirmed and fixed in 2ffdc9e06. get_adapters_list() isn't memoised and calls get_json_schema() per adapter, which is an open() and read() - so the whole registry gets walked and every schema file re-read on each icon lookup, four times per profile with three of them discarded.

Split get_display_info into get_icon and get_model_label so the icon resolves once, and switched it to Adapterkit().get_adapter_class_by_adapter_id(...).get_icon() - same value, dict lookup, no file I/O and no INFO line per call. That API was already in use in this file for test_adapter.

frontend/src/components/custom-tools/prompt-card/PromptCardItems.jsx
kirtimanmishrazipstack: @vishnuszipstack Correct, and this was the one worth catching. Both call sites now read profile?.conf?.LLM directly, so getLLMModelNamesForProfiles is gone along with the two /adapter/ round-trips. 2ffdc9e06.

backend/adapter_processor_v2/adapter_processor.py
kirtimanmishrazipstack: @vishnuszipstack The observation is right - that except does swallow InvalidEncryptionKey - but I don't think the re-raise is safe here.

ProfileManagerSerializer also backs the unauthenticated public-share path (public_shares_v2 prompt_share_controller). Re-raising turns a degraded tile label into a 403 reading "Please inform the platform admin immediately" served to an anonymous viewer, on the one display path this PR exists to keep working for non-owners.

The outage is already surfaced where it's actionable: prompt_studio_helper reads default_profile.x2text.metadata unguarded, so the first index or run raises, and both adapter serializers read metadata unguarded too.

What I did take: the log is now error rather than warning, names the real condition, and uses lazy % formatting. Left the catch broad - narrowing to (TypeError, ValueError, AttributeError) would turn any unlisted decrypt failure into a 500 on the whole profile list. 2ffdc9e06.

backend/prompt_studio/prompt_profile_manager_v2/serializers.py
kirtimanmishrazipstack: @vishnuszipstack Right - RelatedObjectDoesNotExist subclasses AttributeError, so the three-arg getattr was eating it. Dropped the default in 2ffdc9e06, so a dangling FK now raises instead of vanishing.

Left the falsy guard rather than adding an explicit ObjectDoesNotExist catch. With null=False and PROTECT on all four there's nothing meaningful left to log, and the smaller change already gets the loud failure you were after.

backend/prompt_studio/prompt_profile_manager_v2/tests/test_profile_display_info.py
kirtimanmishrazipstack: @vishnuszipstack Your read is accurate: the patch sits at the get_display_info seam so the branches inside it have no coverage, and the fake adapter exposing .model while production reads .metadata.get("model") is a real divergence.

I've only updated the patch targets for the split into get_icon/get_model_label. Not adding new test files in this PR - flagging rather than doing. The credential-leak assertion is a good idea and I'd like it in when we do.

backend/prompt_studio/prompt_profile_manager_v2/serializers.py
kirtimanmishrazipstack: @vishnuszipstack Agreed the shortened version lost the justification, but I read DRF 3.17.1 and the "400s on re-save" clause is wrong in both wordings. UniqueTogetherValidator.__call__ excludes the current instance, then builds checked_values only from fields whose value differs from the instance - so an unchanged re-save gives an empty list and never raises. enforce_required_fields also returns early when instance is not None, so a PUT doesn't trip it either.

The line's real effect is on create: without validators = [], a genuine duplicate returns DRF's generic "must make a unique set" instead of the view's DuplicateData. Rewrote the comment to say that. 2ffdc9e06.

backend/adapter_processor_v2/adapter_processor.py
kirtimanmishrazipstack: @vishnuszipstack Largely moot now - get_display_info is split into get_icon and get_model_label, so that docstring is gone and each half states what it actually does. The "falls back" framing is fixed: get_model_label says adapters without a model use the provider prefix.

On the log, agreed it was a misnomer. It now says the adapter is not in the SDK registry, and since the icon path no longer goes through get_adapter_data_with_key it's no longer a duplicate either.

One correction: "".split("|") returns [''], so that [0] can't IndexError for any string - the non-null default isn't what was holding it up. 2ffdc9e06.

frontend/src/components/custom-tools/prompt-card/PromptOutput.jsx
kirtimanmishrazipstack: @vishnuszipstack The mechanical part is right, but neither route you cite is reachable. llm is null=False with PROTECT, so a dangling FK needs raw SQL; and there's no cached profile payload - llmProfiles only ever comes from responses that went through this serializer, the public-share one included. get_icon is also seeded with the warning icon and can't return empty.

Added the fallback anyway, since the extraction in the next thread gave it a single home. 2ffdc9e06.

frontend/src/components/custom-tools/prompt-card/PromptOutputsModal.jsx
kirtimanmishrazipstack: @vishnuszipstack Done - ProfileIcon.jsx in prompt-card/. Both call sites collapse to one line and both drop their Image and isImageUrl imports, and the fallback now lives in one place. 2ffdc9e06.

frontend/src/helpers/GetStaticData.js
kirtimanmishrazipstack: @vishnuszipstack Tooling check is right, and so is the CI gap. Not adding the vitest file in this PR, same as the serializer test thread - happy to follow up.

On the first minor note I'd push back: /icons/adapter-icons/OpenAI.png is a relative URL, and the ^\/ branch exists precisely to match it, so "URLs" is accurate. The second note I took - the comment now says the emoji is the fallback when an icon can't be resolved, rather than only for unavailable adapters. 2ffdc9e06.

backend/prompt_studio/prompt_profile_manager_v2/serializers.py
kirtimanmishrazipstack: @vishnuszipstack Both taken. Object.entries now guards with || {}, and the annotation is dict[str, Any]. Also reworded the ADAPTER_LABELS comment so it reads as a response-key map rather than display copy, which was the thing inviting the rename.

Left the keys as they are - re-keying to the FK names would just move the display strings into the frontend for no user-visible gain. 2ffdc9e06.

backend/adapter_processor_v2/adapter_processor.py
kirtimanmishrazipstack: @vishnuszipstack Took the split - get_icon and get_model_label, two named single-value returns, nothing to mis-destructure, and the icon resolves only for the LLM. Same change that removes the wasted registry walks. 2ffdc9e06.

kirtimanmishrazipstack (commented):
Superseded by the round 2 review on this PR (head `4ee045b0`).

backend/adapter_processor_v2/adapter_processor.py
```diff
@@ -109,6 +111,37 @@ def get_adapter_data_with_key(adapter_id: str, key_value: str) -> Any:
...
+        try:
+            adapter_class = Adapterkit().get_adapter_class_by_adapter_id(
+                adapter.adapter_id
+            )
+        except Exception as e:
+            logger.warning(
+                "Adapter %s is not in the SDK registry: %s", adapter.adapter_id, e
+            )
+            return UNAVAILABLE_ADAPTER_ICON
+        return adapter_class.get_icon()
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps False positive — that branch can't execute. Not applying the suggestion.

The premise ("an available SDK adapter returns a falsy icon value") holds for the *previous* implementation but not the current one. Worth separating the two, because your earlier comment on this same line was correct:

**Old:** `get_adapter_data_with_key(adapter_id, common.ICON)` bottomed out in `updated_adapters[0].get(key_value)` — a dict `.get()`, which returns `None` on a missing key without raising. `or UNAVAILABLE_ADAPTER_ICON` was genuinely load-bearing there, exactly as you flagged.

**Now:** `get_icon` calls `Adapterkit().get_adapter_class_by_adapter_id(...)` and then the class's own `get_icon()`. Two different guarantees:

1. `get_adapter_class_by_adapter_id` never returns a falsy class — it's `if adapter_id in self._adapters: return ... else: raise RuntimeError(...)`, and the raise is already caught above.
2. `Adapter.get_icon()` is `-> str`, and every registered adapter overrides it with a literal `/icons/adapter-icons/*.png`.

Enumerated the live registry rather than reasoning about it:

```
registered: 36 | falsy get_icon(): []
base Adapter.get_icon() -> ''
base registered? False
```

The only implementations returning `""` are the abstract bases (`base.py:36`, `x2text_adapter.py`, `vectordb_adapter.py`, `ocr_adapter.py`). Those are never registered — `_adapters` is built from the concrete `adapters` dicts in `llm1/`, `embedding1/`, `vectordb/`, `x2text/`, `ocr/`, so `Adapter` itself isn't in it (third line above).

So `or UNAVAILABLE_ADAPTER_ICON` was dead code, which is why it came out. Adding it back would restore an unreachable branch, and it wouldn't make the tile safer — the tile's actual protection is `ProfileIcon`'s `isImageUrl` branch, which handles any non-URL value including `""`.

`UNAVAILABLE_ADAPTER_ICON` still covers the two reachable cases (`is_available=False`, and an id missing from the registry), both of which return it explicitly above.

Adding a guard for a fourth adapter type that ships without an icon is a real concern, but that's a registry invariant — it belongs in the SDK where `get_icon()` is defined, not as a `or` on every call site.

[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Retracting my previous reply — you're right, applying the fix.

My "unreachable" argument only held within this repo. `unstract-sdk` releases on its own cycle, so "all 36 registered adapters override `get_icon()` with a real path" describes the pinned version, not a contract this code can rely on. A new adapter landing there without an icon override inherits `Adapter.get_icon() -> ""` from the base and silently blanks the tile — with no failure anywhere near the change that caused it. A cross-repo invariant asserted by enumeration isn't an invariant.

Applied, with the constant moved to `AdapterKeys.UNAVAILABLE_ICON` in `adapter_processor_v2/constants.py` alongside the other adapter strings:

```python
return adapter_class.get_icon() or AdapterKeys.UNAVAILABLE_ICON
```

The two reachable cases above it (`is_available=False`, id missing from the registry) now use the same constant.

kirtimanmishrazipstack (commented):
## Self-review round 2 — UN-3991 (head `4ee045b0`)

Re-reviewed after the four commits that followed my last pass (`9a9a2b88`..`4ee045b0`). Those are all improvements — `UNAVAILABLE_ICON` moved into `AdapterKeys`, the `AdapterListSerializer` icon path collapsed onto one helper, `Object.entries(adapterConf || {})` guarded, and the `*_id` + disabled-`<Option>` work fixes a real bug where a co-owner editing a profile would silently drop an adapter they can't see.

**Verified clean:** `npm run build` exit 0 (31.14s); `ruff` via pre-commit on all 7 changed backend files — passed; `biome ci` on the 8 changed frontend files — 3 warnings, all pre-existing (`catch {}` ×2, `useBlockStatements` ×1); no dangling callers of the three deleted helpers.

Four things I checked rather than assumed:

- **`get_icon` is value-identical to the `get_adapter_data_with_key(id, "icon")` it replaces**, and strictly cheaper. SDK `get_adapters_list()` builds `"icon": m.get_icon()` (`adapterkit.py:71`), so the value matches; the old path rebuilt the whole registry list including a `get_json_schema()` file read per adapter, where the new one is a dict lookup on a `@singleton`. This resolves the earlier P1 on registry-scan cost.
- **Every SDK `get_icon()` returns a leading-slash absolute path** (27 implementations, plus `""` on the base), so `isImageUrl`'s `^(https?://|/|data:image/)` cannot misclassify a real icon as an emoji.
- **`adapterOptions` can't crash on `[...items]`** — all four `*Items` are `useState([])`, never undefined.
- **Public share still renders model names** — `public_shares_v2/.../prompt_share_controller.py:131` uses this same serializer, so `conf`/`icon` arrive. Dropping `publicAdapterApi` from the two consumers is safe.

**One Medium, no High.** Lows were reviewed and are tracked outside this thread.

---

### [Medium] The new `*_id` fields expose ids of adapters the viewer can't access, and nothing rejects those ids on write

`prompt_studio/prompt_profile_manager_v2/serializers.py` — `rep[f"{field}_id"] = str(rep[field])`

The field itself is justified: `AddLlmProfile.jsx` needs the id to preselect, and matching by adapter *name* was the bug this PR exists to fix. Recording what else it enables.

Before this change all four FKs were masked to `adapter.adapter_name` in the response. Now a user who is merely *shared* a project — `ProfileManager.objects.for_user` admits them via `Q(prompt_studio_tool__in=accessible_tools)` (`models.py:38`) — receives `llm_id`, `vector_store_id`, `embedding_model_id` and `x2text_id` for adapters that `AdapterInstance.objects.for_user` would not return to them.

The chain from there, each link read in source:

1. They POST those ids into a profile on a project **they own** (so `IsParentToolOwner` passes). `ProfileManagerSerializer` has `fields = "__all__"`, `validators = []`, no `validate_*`, and DRF builds the relation field against `AdapterInstance._default_manager` — `AdapterInstanceModelManager`, which is org-scoped but **not** user-scoped (`adapter_processor_v2/models.py:39-59`). Accepted.
2. Running it inside Prompt Studio is correctly **blocked** — `PromptStudioHelper.validate_profile_manager_owner_access` checks the profile *creator's* access, and the creator here is the requester. 10 call sites, well covered.
3. `export_tool` with `force_export: true` — a plain request-body boolean (`prompt_studio_registry_v2/serializers.py:47`) — skips the "prompts were not run" gate (`prompt_studio_registry_helper.py:396` and `:488`), and the registry helper never calls that guard.
4. `prompt_studio_registry_helper.py:406-409` then writes `str(prompt.profile_manager.llm.id)` and the other three into the tool spec. Deployed as an API deployment or ETL pipeline, prompt-service resolves the adapter by id and decrypts the owner's credentials.

The only other export gate, `invalidated_prompts` (`:389`), fires on *empty prompt text*.

Net effect: an org peer with read-only sharing can end up spending another user's LLM credentials.

**What is and isn't new.** Steps 1, 3 and 4 all pre-date this PR. What this PR adds is the missing precondition — knowing the UUID. Counter-argument worth weighing: a UUID was never an access control, so the hole was arguably always open and this only makes it reachable without guessing.

For reference, `prompt_studio_core_v2/serializers.py:133` (`validate_summarize_llm_adapter`) already does an `AdapterInstance.objects.for_user(request.user)` check on an adapter FK; `ProfileManagerSerializer` is the outlier among serializers that accept one.

**Caveat:** I read this path, I did not execute it end to end. Worth reproducing before acting on it, and I'd welcome being told the export/deploy step is gated somewhere I didn't look.

kirtimanmishrazipstack:
Closing out the Medium from [self-review round 2](https://github.com/Zipstack/unstract/pull/2240#pullrequestreview-4957481273) — fixed in c0cd6812c.

`ProfileManagerSerializer` now has a `validate()` that rejects a *change* to an adapter the requester can't access, checked against `AdapterInstance.objects.for_user`. Step 1 of the chain in that review no longer holds, so the harvested id has nowhere to land and steps 3–4 are unreachable by this route.

**Why it validates the diff rather than scoping the queryset.** The obvious fix — rebind the four FK querysets to `for_user` in `__init__`, the way `AgenticProjectSerializer` does — would have broken this PR's own feature. `AddLlmProfile.handleSubmit` sends `data: formDetails`, the whole form, on `PUT`. A co-owner editing a profile that references an adapter shared only with the owner would resend that id untouched and get a 400 on every save. The agentic serializer gets away with queryset scoping only because its frontend PATCHes `changed`. So the check skips a value equal to the one already on `self.instance` and only gates an actual change.

Loops `ADAPTER_LABELS`, so all four FKs are covered by one method. Guarded on `self.context.get("request")` — the cloud public-share controller instantiates this serializer with no context, and that path is read-only.

Verified `AuditSerializer` defines only `create`/`update`, so no `validate()` was shadowed, and `for_user` short-circuits for org admins and service accounts, so neither loses adapters they could pick before. pre-commit clean; frontend build green.

Leaving the `force_export` observation alone: `frame_export_json` still never calls `validate_profile_manager_owner_access`, but with the write scoped it isn't reachable this way. Wiring the guard in needs `request_user` threaded through three signatures, which is a bigger change than this PR should carry.

vishnuszipstack (approved):

PR #338, opened 2024-05-21
Request id in logs and docker memory optimizations

ritwik-g:
## What

- Added request id in logs using `django-log-request-id` module
- Commented out celery flower to reduce unwanted resource usage. End user may enable it locally for debugging
- Reduced thread count off backend to 2
- Added autoscale comment to celery execution consumer to avoid spawning of threads unnecessarily. Without this the celery allocates 8 threads from start increasing the resource usage.
- Totally I have observed around 3-4 GBs reduced after the changes in local.

## Why

- The request id helps to track a transaction during debugging
- Memory optimizations helps to reduce load when running the items locally

## How

-

## Can this PR break any existing features. If yes, please list possible items. If no, please explain why. (PS: Admins do not merge the PR without this section filled)

- No. This PR doesn't contain any code changes other than the additional logging change.

## Screenshots
- Request ID
Screenshot from 2024-05-21 15-50-48
- Docker Stats after optimisations. 
Screenshot from 2024-05-21 15-51-23

# Related docs

- https://github.com/dabapps/django-log-request-id

## Checklist

I have read and understood the [Contribution Guidelines]().

PR #2241, opened 2026-08-13
UN-3996 [FIX] Stop WebSockets exhausting the Gunicorn thread pool

kirtimanmishrazipstack:
## What

- `backend/entrypoint.sh`: Gunicorn workers and threads are now env-driven, with the thread default raised from `2` to `512`.
- `backend/entrypoint.sh`: `--worker-class gthread` is stated explicitly.
- `backend/sample.env`: ships both variables set to their defaults, so the knob is visible in the file every deployment copies.

## Why

- The Socket.IO WSGI app runs with `async_mode="threading"`, so every open WebSocket holds one gthread pool thread for that connection's lifetime rather than for a single request.
- With `--threads 2`, two logged-in browser tabs landing on the same worker left it with zero threads for HTTP, and it kept accepting connections it could never read — so requests stalled silently instead of failing over.
- Reported as Workflows / Prompt Studio / Agentic Prompt Studio loading very slowly, and far worse with two concurrent logins.

## How

- Verified chain: `frontend/src/index.jsx:64` mounts one `SocketProvider` per tab → `backend/backend/wsgi.py:42` wraps Django in the Socket.IO app (`async_mode="threading"`, `utils/log_events.py:29-31`) → `engineio/socket.py:222` runs the socket loop inline on a pool thread sized by `--threads` (`gunicorn/workers/gthread.py:98`).
- `512` matches the value already running in the cloud chart and is free at idle, since `--threads` is a `ThreadPoolExecutor` ceiling and CPython only starts a thread on `submit()` when none is idle.
- `--worker-class gthread` was already in effect — `config.py:109-110` coerces `sync` to `gthread` whenever `threads > 1` — so stating it is a no-op that just removes the ambiguity.
- `worker-connections` is deliberately left alone. `1000` is already Gunicorn's default (`config.py:732`), so passing it explicitly would change nothing, and making it settable only adds a way to configure it below `GUNICORN_THREADS`, which drives `max_keepalived` negative (`gthread.py:73`) and force-closes every response (`gthread.py:330`).

No — the change is confined to Gunicorn CLI arguments, both variables fall back to the values already in use, and no application code is touched. The only behavioural change is that a worker can now serve more than two concurrent connections.

## Database Migrations

- None.

## Env Config

- `GUNICORN_WORKERS` (default `2`, unchanged from before).
- `GUNICORN_THREADS` (default `512`, was a hardcoded `2`) — must exceed concurrent browser tabs per worker.
- Both are set to these values in `backend/sample.env` and reach the container through the existing `env_file: ../backend/.env` on the backend service. The entrypoint keeps a `:-` fallback per variable, so an `.env` that predates this PR starts with the same values.

## Relevant Docs

- None.

## Related Issues or PRs

- Jira: [UN-3996](https://zipstack.atlassian.net/browse/UN-3996).
- Zipstack/unstract-cloud#1724 — closed. The cloud chart already runs `--threads 512`, so no cloud change is needed.
- #338 lowered this value from `4` to `2` as part of a local-memory cleanup. The line it changed carried a `# NOTE updated socket threads` comment, added in the same commit that introduced the Socket.IO threading mode.

## Dependencies Versions

- None.

## Notes on Testing

- `bash -n` passes; argument array expanded and inspected with no env set and with both variables overridden.
- Measured locally with `backend/.venv` (Python 3.12.9) — `--threads` is a ceiling, not an allocation, so a high default is free until the connections exist:

  | scenario | threads alive | RSS delta |
  |---|---|---|
  | ceiling 512, at construction | 0 | 0 |
  | 50 sequential tasks | 1 | — |
  | 20 concurrent blocking tasks | 20 | — |
  | all 512 forced live | 512 | +11.8 MB |

  CPU with 512 threads parked: 0.005% of one core over 3s. `VmSize` grows ~12.5 GB at 512 live threads (glibc reserves an 8 MB stack per thread from `RLIMIT_STACK`), but that is reserved address space rather than RSS.
- Repro for reviewers: with `--threads 2`, open two logged-in tabs and loop `GET /api/v1/health`; one worker stops responding while `ss -tanp | grep :8000` shows ESTABLISHED sockets with non-zero `Recv-Q` and an empty LISTEN backlog.

## Screenshots

N/A — no user-facing surface.

## Checklist

[UN-3996]: https://zipstack.atlassian.net/browse/UN-3996?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/entrypoint.sh
+++ b/backend/entrypoint.sh
@@ -30,17 +30,16 @@ if [ "$migrate" = true ]; then
 fi
 
 # Configure Gunicorn based on --dev flag.
-# Threads must exceed the number of concurrently connected browser tabs: the
-# Socket.IO WSGI app runs with async_mode="threading", so each open WebSocket
-# holds one gthread pool thread for that connection's lifetime, not just for a
-# request. The pool spawns threads lazily, so a high ceiling is free at idle.
+# Threads must exceed concurrent browser tabs — each open Socket.IO WebSocket
+# holds one thread for its lifetime. The pool spawns threads lazily, so a high
+# ceiling is free at idle.
 gunicorn_args=(
     --bind 0.0.0.0:8000
     --workers "${GUNICORN_WORKERS:-2}"
     --threads "${GUNICORN_THREADS:-512}"
     --worker-class gthread
     --worker-connections "${GUNICORN_WORKER_CONNECTIONS:-1000}"
-    --log-level "${DEFAULT_LOG_LEVEL:-INFO}"
+    --log-level debug
     --timeout 600
     --access-logfile -
 )
--- a/backend/entrypoint.sh
+++ b/backend/entrypoint.sh
@@ -29,12 +29,18 @@ if [ "$migrate" = true ]; then
     .venv/bin/python manage.py migrate
 fi
 
-# Configure Gunicorn based on --dev flag
+# Configure Gunicorn based on --dev flag.
+# Threads must exceed the number of concurrently connected browser tabs: the
+# Socket.IO WSGI app runs with async_mode="threading", so each open WebSocket
+# holds one gthread pool thread for that connection's lifetime, not just for a
+# request. The pool spawns threads lazily, so a high ceiling is free at idle.
 gunicorn_args=(
     --bind 0.0.0.0:8000
-    --workers 2
-    --threads 2
-    --log-level debug
+    --workers "${GUNICORN_WORKERS:-2}"
+    --threads "${GUNICORN_THREADS:-512}"
+    --worker-class gthread
+    --worker-connections "${GUNICORN_WORKER_CONNECTIONS:-1000}"
+    --log-level "${DEFAULT_LOG_LEVEL:-INFO}"
     --timeout 600
     --access-logfile -
 )
--- a/backend/entrypoint.sh
+++ b/backend/entrypoint.sh
@@ -38,7 +38,6 @@ gunicorn_args=(
     --workers "${GUNICORN_WORKERS:-2}"
     --threads "${GUNICORN_THREADS:-512}"
     --worker-class gthread
-    --worker-connections "${GUNICORN_WORKER_CONNECTIONS:-1000}"
     --log-level debug
     --timeout 600
     --access-logfile -
--- a/backend/sample.env
+++ b/backend/sample.env
@@ -4,12 +4,11 @@ DJANGO_SETTINGS_MODULE='backend.settings.dev'
 SESSION_COOKIE_SECURE=False
 CSRF_COOKIE_SECURE=False
 
-# Default log level. Also drives Gunicorn's --log-level.
+# Default log level
 DEFAULT_LOG_LEVEL="INFO"
 
-# Gunicorn tuning. GUNICORN_THREADS must exceed the number of concurrently
-# connected browser tabs — each open Socket.IO WebSocket holds one thread for
-# its whole lifetime.
+# Gunicorn tuning. GUNICORN_THREADS must exceed concurrent browser tabs —
+# each open Socket.IO WebSocket holds one thread for its lifetime.
 # GUNICORN_WORKERS=2
 # GUNICORN_THREADS=512
 # GUNICORN_WORKER_CONNECTIONS=1000
--- a/backend/sample.env
+++ b/backend/sample.env
@@ -9,9 +9,9 @@ DEFAULT_LOG_LEVEL="INFO"
 
 # Gunicorn tuning. GUNICORN_THREADS must exceed concurrent browser tabs —
 # each open Socket.IO WebSocket holds one thread for its lifetime.
-# GUNICORN_WORKERS=2
-# GUNICORN_THREADS=512
-# GUNICORN_WORKER_CONNECTIONS=1000
+GUNICORN_WORKERS=2
+GUNICORN_THREADS=512
+GUNICORN_WORKER_CONNECTIONS=1000
 
 # Common
 PATH_PREFIX="api/v1"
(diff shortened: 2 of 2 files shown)


backend/entrypoint.sh
```diff
@@ -29,12 +29,18 @@ if [ "$migrate" = true ]; then
...
+# request. The pool spawns threads lazily, so a high ceiling is free at idle.
 gunicorn_args=(
     --bind 0.0.0.0:8000
-    --workers 2
-    --threads 2
-    --log-level debug
+    --workers "${GUNICORN_WORKERS:-2}"
+    --threads "${GUNICORN_THREADS:-512}"
+    --worker-class gthread
+    --worker-connections "${GUNICORN_WORKER_CONNECTIONS:-1000}"
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps The gap is real, but the suggested value is not safe: `worker_connections == threads` makes `max_keepalived = worker_connections - threads = 0` (`gthread.py:73`), which trips gunicorn's own "No keepalived connections can be handled" warning (`gthread.py:86-89`) and force-closes every response at `gthread.py:330`, disabling HTTP keep-alive entirely — cloud deliberately runs `--keep-alive 620`. The `1000 - 512 = 488` excess is precisely that keep-alive budget.

Residual exposure above 512 sockets per worker is the thread-per-connection ceiling itself, tracked as follow-up (gevent worker class, or hosting Socket.IO in its own process); this PR narrows the accept/thread gap from 998 to 488 rather than widening it.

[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps please review

kirtimanmishrazipstack:
@greptile-apps please re-review.

The one open finding (`backend/entrypoint.sh:40-42`) has been answered inline rather than patched: `worker-connections == threads` would set `max_keepalived = 0` (`gthread.py:73`), tripping gunicorn's own "No keepalived connections can be handled" warning (`gthread.py:86-89`) and force-closing every response (`gthread.py:330`), which disables HTTP keep-alive. The `1000 - 512 = 488` excess is that keep-alive budget, so the current values are intentional.

backend/entrypoint.sh
```diff
@@ -29,12 +29,18 @@ if [ "$migrate" = true ]; then
...
+# Socket.IO WSGI app runs with async_mode="threading", so each open WebSocket
+# holds one gthread pool thread for that connection's lifetime, not just for a
+# request. The pool spawns threads lazily, so a high ceiling is free at idle.
 gunicorn_args=(
     --bind 0.0.0.0:8000
-    --workers 2
-    --threads 2
-    --log-level debug
+    --workers "${GUNICORN_WORKERS:-2}"
+    --threads "${GUNICORN_THREADS:-512}"
```
ritwik-g, line 39: @kirtimanmishrazipstack a default 512 threads might eatup your local machine memory and cpu. I believe that is why it was left 2 deliberately in the beginning. @muhammad-ali-e would be better person to answer. 

What I would suggest is a bit more conservative default value something like 8 or 16.

Also if any other similar services need such a configurability please do include that as well.

backend/entrypoint.sh
```diff
@@ -29,12 +29,18 @@ if [ "$migrate" = true ]; then
...
+    --worker-class gthread
+    --worker-connections "${GUNICORN_WORKER_CONNECTIONS:-1000}"
+    --log-level "${DEFAULT_LOG_LEVEL:-INFO}"
```
ritwik-g: @kirtimanmishrazipstack I will suggest using a different env for this log level. There might be places when we need gunicorn at debug log level but not the application

backend/entrypoint.sh
kirtimanmishrazipstack: @ritwik-g `--threads` is a ceiling, not an allocation. Gunicorn passes it straight to `ThreadPoolExecutor(max_workers=...)` (`gthread.py:98`) and CPython only starts a thread on `submit()` when none is idle, so live threads track peak concurrency rather than the ceiling.

Measured locally with `backend/.venv` (3.12.9):

| scenario | threads alive | RSS delta |
|---|---|---|
| ceiling 512, at construction | 0 | 0 |
| 50 sequential tasks | 1 | — |
| 20 concurrent blocking tasks | 20 | — |
| all 512 forced live | 512 | +11.8 MB |

CPU with 512 threads parked: 0.005% of one core over 3s.

`VmSize` does grow ~12.5 GB at 512 live threads — glibc reserves an 8 MB stack per thread from `RLIMIT_STACK` — but that is address space, not RSS. It shows up in `docker stats` / `top` VSZ, which is probably where the concern comes from.

On the history: the initial commit `26ebb17d4` set `--threads 4` with `# NOTE updated socket threads` on the line directly above it, and the socketio threading mode shipped in that same commit — so 4 was chosen for the WebSockets. #338 took it to 2 as part of the memory cleanup without touching that note.

On 8 or 16: this entrypoint is also what self-hosted deployments run, not just local dev, so ~16 open tabs would put them back in the same hang. Since the ceiling costs nothing until used I would rather match the 512 cloud already runs — but if you still want it lower after these numbers, I will take it.

On other services: backend is the only gunicorn app running socketio in either repo, so nothing else has this bug.

backend/entrypoint.sh
kirtimanmishrazipstack: @ritwik-g agreed, this one was my mistake. `DEFAULT_LOG_LEVEL` drives Django's console handler and root logger, while gunicorn's loggers are `propagate=False` with their level from `cfg.loglevel` — two independent knobs, and I collapsed them.

Reverted — the line is back to `--log-level debug`, byte-identical to main, so it now shows in the diff only as context. This PR is the thread fix alone.

ritwik-g (approved):

PR #2245, opened 2026-08-17
UN-2238 [FIX] Enforce the frontend CSP (out of report-only mode)

Deepak-Kesavan:
## What

Takes the frontend CSP out of report-only mode, which is what UN-2238 asks for. Two commits:

1. **Widen the report-only policy** to the third-party origins the shipped frontend actually loads, found by sweeping a live deployment, and drop the bare `wss:` wildcard from `connect-src`. Adds `.claude/skills/csp-check` — scripts + procedure for re-checking the policy against a build or a running deployment (the "way to check compatibility of new features with the CSP policy" Hari asked for in the ticket).
2. **Flip the header** from `Content-Security-Policy-Report-Only` to `Content-Security-Policy`. The policy value is byte-identical across that commit — the only change is that an unlisted origin is now blocked rather than logged.

## Why

UN-2238. CSP shipped in report-only mode with no report collector, so violations only ever reached each user's browser console. The ticket's exit condition is "once there are no more violations, we can take CSP out of report mode" — commit 1 closes the violations, commit 2 does the flip.

### Commit 1 — what the sweep added

Two methods: browsing while recording `securitypolicyviolation` (landing, dashboard, prompt studio incl. PDF viewer, API/ETL/task, logs, users, LLM settings, platform settings, profile, manual review), plus per-directive probes for hosts the bundle references whose code paths were not exercised. Each host below traces to a chunk in the deployed bundle.

| Directive | Added | Loaded by |
|---|---|---|
| `style-src` | `cdn.jsdelivr.net`, `app.productfruits.com` | Monaco `editor.main.css`; ProductFruits CSS (observed in normal use) |
| `font-src` | `cdn.jsdelivr.net` | Monaco `codicon.ttf` |
| `img-src` | `cdn.jsdelivr.net`, `app.productfruits.com`, `cdn.productfruits.com`, `www.googletagmanager.com`, `www.google-analytics.com`, `www.google.com/recaptcha/`, `www.gstatic.com/recaptcha/`, `q.stripe.com` | emoji-datasource; ProductFruits media; GTM/GA pixels; reCAPTCHA assets; Stripe beacons |
| `connect-src` | `unpkg.com`, `api.productfruits.com`, `www.googletagmanager.com`, `region1.google-analytics.com`, `analytics.google.com`, `www.google.com/recaptcha/`, `m.stripe.network` | LLMW pdf worker; ProductFruits API; GA4 endpoints; reCAPTCHA api2; Stripe.js |
| `frame-src` | `www.googletagmanager.com`, `m.stripe.network` | react-gtm `ns.html`; Stripe.js |
| `media-src` | new directive: `'self' blob: cdn.productfruits.com` | ProductFruits video (previously fell back to `default-src 'self'`, no `blob:`) |

### Commit 2 — rechecking before the flip

That sweep predates the shadcn / React 19 frontend migration (359 files), so the policy was rechecked against both bundles:

- **OSS build** (`bun run build`, 111 chunks) — every external host it references is in the policy.
- **Live us-central deployment** (363 chunks, which includes the cloud-plugin code the OSS build does not have) — same.

The one host either scan turned up that no directive allows is `react.dev`, which appears only in React 19's error-message links and is never fetched. It went into `scan_origins.py`'s `IGNORED` set, not into the policy.

The skill's docs were also refreshed: they described a report-only policy and an `npm run build` / `frontend/dist` layout that is now `bun run build` / `frontend/build`.

Yes — this is the commit where a policy gap stops being a console message and starts being a broken feature. What was checked, and what is left:

- **Everything the bundle references is covered** by the two scans above, and the browse-through found exactly one real-use violation (ProductFruits `animations.css`), which commit 1 allows.
- **Login and form posts are unaffected.** Login is `window.location.href = ${origin}/api/v1/login` — a navigation, which `form-action` does not govern. App forms are React `onSubmit` handlers with no native submit.
- **Same-origin API and sockets are unaffected.** All API calls use relative `/api/v1/...` paths and socket.io connects to `window.location.origin`; `'self'` covers both, including same-origin ws/wss under CSP3.
- **`frame-ancestors 'self'`** was already enforced in practice by the existing `X-Frame-Options: SAMEORIGIN`.
- **Operator-facing change worth knowing:** `VITE_CUSTOM_LOGO_URL` and `VITE_FAVICON_PATH` (via `generate-runtime-config.sh`) let a deployment point the logo/favicon at an arbitrary URL. Under `img-src`, an *external* logo host now has to be added to the policy or the logo will not render. Same-origin values are unaffected.
- **Not covered by a static scan:** a host that is fetched only at runtime from a string the scanner cannot see, and a host allowed on the wrong directive (a font pulled from a script-src-only host still violates). The August browse-through covered this for the flows listed above; it has not been repeated against a deployment carrying the shadcn migration. **That browse-through on a dev/QA deployment of this branch is the remaining pre-merge check** — `.claude/skills/csp-check` check 3 is the procedure.

Rollback is a one-line revert of commit 2.

## Database Migrations

- None.

## Env Config

- None added. See the `VITE_CUSTOM_LOGO_URL` / `VITE_FAVICON_PATH` note above for an existing var whose behaviour now has a CSP constraint.

## Relevant Docs

- `.claude/skills/csp-check/SKILL.md`

## Related Issues or PRs

- Follows #1834 and #1875.
- UN-2238

## Dependencies Versions

- None.

## Notes on Testing

- `scan_origins.py --dist frontend/build` against a local production build of this branch: clean, 111 chunks.
- `scan_origins.py --url https://us-central.unstract.com`: clean, 363 chunks.
- `extract_policy.py` parses the flipped header and reports 13 directives, and the policy string is byte-identical to the pre-flip one (verified by diffing the two `add_header` values).
- From the previous round: `docker run` of `nginx:alpine` with this `nginx.conf` — config accepted, header emitted verbatim; 48 per-directive browser probes against it — every listed origin loads with no report, and all 4 controls to a disallowed host were reported. The conf's directive structure is unchanged since, only the header name differs. Docker was not available in this round, so that check was not re-run.


frontend/nginx.conf
```diff
@@ -64,8 +64,13 @@ http {
...
...m; img-src 'self' data: blob: https://eu.i.posthog.com https://eu-assets.i.posthog.com https://cdn.jsdelivr.net https://app.productfruits.com https://cdn.productfruits.com https://www.googletagmanager.com https://www.google-analytics.com https://www.google.com/recaptcha/ https://www.gstatic.com/recaptcha/ https://q.stripe.com; font-src 'self' data: https://cdn.jsdelivr.net; connect-src 'self' blob: https://cdn.jsdelivr.net https://unpkg.com https://eu.i.posthog.com https://eu-assets.i.posthog.com https://www.googletagmanager.com https://www.google-analytics.com https://region1.google-analytics.com https://analytics.google.com https://www.google.com/recaptcha/ https://api.stripe.com https://m.stripe.network https://app.productfruits.com https://api.productfruits.com; frame-src 'self' https://www.google.com/recaptcha/ https://recaptcha.google.com https://www.googletagmanager.com https://js.stripe.com https://hooks.stripe.com https://m.stripe.network; media-src 'self' blob: https://cdn.productfruits.com; worker-src 'self' blob: https://unpkg.com https://cdn.jsdelivr.net; object-src 'none'; base-uri 'self'; form-action 'self' https://checkout.stripe.com; frame-ancestors 'self'" always;
```
praveen-formido, line 73: **Enforced `img-src` blocks operator-configured branding URLs.** `generate-runtime-config.sh` exposes `VITE_CUSTOM_LOGO_URL` / `VITE_FAVICON_PATH` (and the `REACT_APP_*` fallbacks) to `config.logoUrl` / `config.favicon`. The on-prem Helm values in unstract-cloud say `REACT_APP_CUSTOM_LOGO_URL: "" # Replace with logo URL`. If an operator sets it to an absolute URL on their own host, for example `https://cdn.acme.com/logo.svg`, the `` in `TopNavBar.jsx` and the `<link rel=icon>` set in `index.jsx` are both governed by `img-src`, which only allows `'self'` and the listed vendor hosts. Once the header enforces, the logo is blocked, `onError` falls back to the default logo, and the favicon silently reverts. The bundle scan cannot see this because the value is supplied at runtime. Either document that these must be same-origin paths, or derive an `img-src` addition from the same env vars when the config is generated.

praveen-formido, line 73: **GA4 endpoints are pinned to specific hosts, which breaks once the policy enforces.** `connect-src` lists `region1.google-analytics.com` and `analytics.google.com`. gtag also sends to `region1.analytics.google.com`, and when Google Signals or ads features are on it sends to `*.g.doubleclick.net` and `www.google.com/g/collect`, `/ccm/collect`. Only `/recaptcha/` is allowed on `www.google.com`. Google's published GA4 CSP guidance uses `https://*.google-analytics.com https://*.analytics.google.com https://*.googletagmanager.com` for this reason. The GTM container is also edited outside the repo, so a tag someone adds later in the GTM UI (Signals, Ads conversion, a Custom HTML tag) is dropped with no report collector to show it. Report-only mode hid this; enforcing turns it into silently lost analytics. Consider the documented wildcards for the Google analytics hosts.

.claude/skills/csp-check/SKILL.md
```diff
@@ -0,0 +1,127 @@
...
+
+Three checks, cheapest first. Run 1 on every change that touches a frontend dependency;
+run 2 and 3 before widening the policy or shipping a new third-party integration.
+
+## 1. Static: does the bundle reference a host the policy never allows?
+
+```bash
+cd .claude/skills/csp-check/scripts
+python3 extract_policy.py                                  # what the policy says today
+python3 scan_origins.py --url https://us-central.unstract.com   # or --dist frontend/build
```
praveen-formido, line 33: **`--dist frontend/build` after `cd .claude/skills/csp-check/scripts` gives a false pass.** Run from this directory, the relative path resolves to `scripts/frontend/build`, which does not exist. `read_dist` calls `Path.rglob` on it, and `rglob` on a missing directory returns `[]` without raising. The script then prints `Scanned 0 bundle files` followed by `Every external host referenced by the bundle appears in the policy.` and exits 0. That is exactly the gate this skill exists to provide, and it passes vacuously. Suggested fix: use `--dist ../../../../frontend/build` in the doc, and in `scan_origins.py` exit non-zero when `args.dist` is not a directory or when `files` is empty.

.claude/skills/csp-check/scripts/scan_origins.py
```diff
@@ -0,0 +1,136 @@
...
+    "us-assets.i.posthog.com",
+    "docs.unstract.com",
+    "join-slack.unstract.com",
+    "billing.stripe.com",
+    "checkout.stripe.com",
+    "fonts.google.com",
+}
+
+
+def read_dist(dist: Path) -> dict[str, str]:
```
praveen-formido, line 65: **Silent empty scan (the code side of the SKILL.md issue).** `read_dist` does not check that `dist` exists or contains anything. `read_deployment` has the same problem: every chunk fetch failure is only printed to stderr (for example an auth wall, a redirect to login, or a CDN 403). In both cases `found` ends up empty, `uncovered` is empty, and `main` prints "Every external host ... appears in the policy" with exit code 0. Fail when `len(files) == 0`, and treat fetch errors as a non-zero exit, or at least a warning that changes the exit code. Separately, only `/assets/*.js|css` is scanned. Hosts that reach the page through `index.html` or `/config/runtime-config.js` (the custom logo and favicon URLs, for example) are never checked.

.claude/skills/csp-check/scripts/scan_origins.py
```diff
@@ -0,0 +1,136 @@
...
+        queue.extend("/assets/" + m[2:] for m in RELATIVE_ASSET_RE.findall(body))
+    return files
+
+
+def allowed_hosts(directives: dict[str, list[str]]) -> set[str]:
+    hosts = set()
+    for sources in directives.values():
+        for source in sources:
+            if source.startswith("https://"):
+                hosts.add(source[len("https://") :].split("/")[0])
```
praveen-formido, line 99: **`allowed_hosts` drops the path from path-scoped sources, so the scanner over-reports coverage.** `https://www.google.com/recaptcha/` becomes `www.google.com`, and every bundle reference to `https://www.google.com/...` is then marked `allowed`. The same applies to `www.gstatic.com`. The browser only permits `/recaptcha/...` on those hosts. So a new dependency that fetches `https://www.google.com/g/collect`, or a Google Maps or fonts path on `www.gstatic.com`, passes the static check and is then blocked by the enforcing policy. Either keep the path and match URL prefixes (URL_RE would need to capture the path), or mark path-scoped hosts as "needs probe" rather than `allowed`.

praveen-formido (commented):
Review findings (correctness pass).

praveen-formido (approved):

PR #2248, opened 2026-08-19
UN-4020 [FIX] Route AWS Bedrock Mantle models (GPT-5.6 Terra) via bedrock_mantle

praveen-formido:
## What

AWS serves the OpenAI GPT-5.x, Google Gemma 4 and xAI Grok families from **Bedrock Mantle** (`bedrock-mantle.<region>.api.aws`) — an OpenAI-compatible endpoint separate from the classic `bedrock-runtime` Converse/Invoke surface. Our adapter prefixed every model with `bedrock/`, so those models were sent to the wrong endpoint with the wrong body schema and never worked. Reported by an on-prem customer trying to use `openai.gpt-5.6-terra`.

This routes them as `bedrock_mantle/<model>` instead. **No UI change** — users keep entering the bare model ID (`openai.gpt-5.6-terra`); the adapter picks the provider itself, exactly as it already does for `bedrock/`.

## How the route is decided

Membership is an **exact lookup** in LiteLLM's own model registry, not a family/prefix match, because the two id namespaces are disjoint but look alike:

| User enters | Routed as |
|---|---|
| `openai.gpt-5.6-terra` | `bedrock_mantle/openai.gpt-5.6-terra` |
| `openai.gpt-oss-120b` | `bedrock_mantle/openai.gpt-oss-120b` |
| `openai.gpt-oss-120b-1:0` | `bedrock/openai.gpt-oss-120b-1:0` |
| `anthropic.claude-…` | `bedrock/anthropic.claude-…` |
| `us.openai.gpt-5.6-terra` | `bedrock/us.openai.gpt-5.6-terra` |

An explicit prefix is honoured verbatim — that keeps already-stored configs working and makes `validate()` a fixed point, which matters because `LLM.complete()` re-validates kwargs on every call.

## litellm 1.90.3 → 1.96.2

Required, but not for the reason you'd guess: 1.90.3 **already knew** the `bedrock_mantle` provider — it just ignored explicit `aws_access_key_id`/`aws_secret_access_key` on that route, so per-adapter Access Keys auth silently failed. Explicit keys are honoured from 1.95.1.

## Also in this PR

- **Guardrails / Application Inference Profile ARNs are stripped for Mantle models, with a warning.** Both are Converse-only features (AWS documents guardrails as "Converse API only" for these models). LiteLLM discards them silently, which would leave an operator believing a guardrail is enforced when it is not.
- **"Enable Extended Thinking" now maps to `reasoning_effort` for non-Anthropic families** (Mantle GPT-5.x, gpt-oss, Nova 2). The switch emitted Anthropic's `thinking` block for every model, which LiteLLM drops for everything except Claude — so the control had never done anything outside Claude.
- **Kept the cohere embed timeout patch**, re-pointed at 1.96.2. Upstream fixed only the async path; the synchronous `embedding()` that Bedrock Cohere embeddings actually use still builds an untimed `HTTPHandler`.
- **Deduplicated the Anthropic model gate** — the prompt-caching feature's inline check now reuses the same helper as the reasoning config, so the two cannot drift.
- **`boto3` added to sdk1's `test` group** — LiteLLM signs Bedrock requests with SigV4 via botocore, which the test venv needs to exercise signing.

## Testing

`563 passed` on litellm 1.96.2. 32 new tests: unit tests for routing, the Mantle strip, and the per-family reasoning shape; plus `test_bedrock_mantle_routing.py`, which feeds the adapter's output through `litellm.completion()` with the HTTP transport patched and asserts the real wire request — URL, SigV4/Bearer header and body. No AWS credentials, network or spend involved.

Reverting the routing makes **14 of the new tests fail**, so they genuinely pin the behaviour.

Verified end to end in Docker (all 24 containers), in both the backend and a worker:

```
litellm 1.96.2
openai.gpt-5.6-terra  -> bedrock_mantle/openai.gpt-5.6-terra
```

Cost resolves for `bedrock_mantle/openai.gpt-5.6-terra` = `(4.4, 19.8)` per 1M, so usage rows won't silently record $0.

## Known limitations (documented, not fixed here)

- `us.`/`global.` cross-Region inference profile IDs belong to `bedrock-runtime` and still don't work; noted in the model field's help text.
- On litellm ≤ 1.96.2 the Mantle **chat-completions** path (models without `mode: responses`, e.g. `openai.gpt-oss-120b`) still ignores explicit access keys — an upstream bug. GPT-5.x is unaffected as it uses the Responses path; bearer token and IAM role work everywhere.


unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -1081,6 +1090,128 @@ def _clean_str(value: str | None) -> str | None:
...
+    override the routing by typing the fully-qualified model id.
+    """
+    if model.startswith(_BEDROCK_MANTLE_PREFIX):
+        return True
+    if model.startswith(_BEDROCK_PREFIX):
+        return False
+    # Imported lazily: base1 is imported during adapter discovery, and litellm
+    # is a heavy import we do not want on that path unless it is needed.
+    import litellm
+
```
johnyrahul, line 1121: **[High] [Lens 3 · 8 · 10] — Endpoint routing is decided by a JSON fetched from the internet at import**

`litellm.model_cost` is populated at litellm *import* time by a GET to `raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json`, falling back to the bundled copy on failure. `LITELLM_LOCAL_MODEL_COST_MAP` is set **nowhere** in this repo outside the new test file, so I measured `source=remote, is_env_forced=False` in this branch's venv.

That makes *which AWS host, wire format and auth surface a customer's request reaches* a function of a third-party file on litellm's `main`, resolved per process. Consequences:
- If the loaded map lacks the key, `validate()` silently emits `bedrock/<model>` — the exact bug this PR fixes — **and** `_strip_unsupported_mantle_params` at L1311 is skipped with it, so a Guardrail is retained on a route that cannot honour it.
- `audit.py:104` swallows the matching cost miss to `logger.debug` and records **$0**.
- Two workers importing litellm at different moments can route the same saved adapter differently.

**Not currently a flake** — I ran the suite 5x clean, and it still passes with egress blocked, because the bundled 1.96.2 map carries all 13 `bedrock_mantle/` entries. The risk is latent: the first Mantle model AWS ships after this pin will exist in one map and not the other.

Reached independently by four review agents. Fix is a judgement call for you — pin `LITELLM_LOCAL_MODEL_COST_MAP=True` in the backend/worker env so routing follows the pinned wheel, or keep an explicit in-repo id set with the registry as fallback. Either makes routing a property of the deploy rather than of the network.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -1081,6 +1090,128 @@ def _clean_str(value: str | None) -> str | None:
...
+    as "Converse API only" for these models). LiteLLM silently discards them,
+    which would leave an operator believing a guardrail is enforced when it is
+    not -- so drop them here and say so loudly.
+    """
+    unsupported = [key for key in ("guardrailConfig", "model_id") if validated.get(key)]
+    if not unsupported:
+        return
+    for key in unsupported:
+        validated.pop(key, None)
+    logger.warning(
```
johnyrahul, line 1221: **[High] [Lens 4 · 8 · 10] — A configured Guardrail is silently disabled, and the warning goes to an audience that isn't the operator**

`logger.warning` here lands in backend/worker **stdout** — in Unstract Cloud that is Zipstack's log, not the tenant's. Meanwhile the adapter form still shows the guardrail ID populated, and nothing in the UI, the test-connection result, or the execution log stream says the control is inert. The docstring above names this exact hazard ("would leave an operator believing a guardrail is enforced when it is not") and then mitigates it to a log the operator cannot see.

Verified: fires once at LLM construction, guardrail absent from the payload thereafter.

Worth separating the two keys, since they are not equivalent — `model_id` is an Application Inference Profile ARN used only for cost attribution, so warn-and-drop is proportionate. `guardrailConfig` is a moderation control the operator explicitly enabled. `validate()` already has a path to the user: `llm.py:313` converts `ValueError` into `SdkError("Invalid LLM adapter metadata: ...")`, which surfaces in the UI. Was hard-failing considered and rejected? That is a product call, but degrading a moderation control on a stdout warning seems the wrong default.

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/bedrock.json
```diff
@@ -66,12 +66,12 @@
...
-      "description": "Enhance reasoning for complex tasks with step-by-step transparency. Available only for Claude 3.7 Sonnet."
+      "description": "Enhance reasoning for complex tasks with step-by-step transparency. For Claude models this enables extended thinking with the token budget below. For GPT-5.x (Bedrock Mantle), GPT-OSS and Nova 2 models it enables the model's own reasoning at medium effort; the budget below does not apply. Models without reasoning support ignore this setting."
     },
     "guardrail_identifier": {
       "type": "string",
       "title": "Bedrock Guardrail Identifier",
-      "description": "Optional. Apply an [AWS Bedrock Guardrail](https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails.html) to moderate prompts and completions. Provide the guardrail ID (e.g. `ff6ujrregl1q`). Leave blank to disable."
+      "description": "Optional. Apply an [AWS Bedrock Guardrail](https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails.html) to moderate prompts and completions. Provide the guardrail ID (e.g. `ff6ujrregl1q`). Leave blank to disable. Not supported for Bedrock Mantle models (GPT-5.x, Gemma 4, Grok) — ignored with a warning."
```
johnyrahul: **[High] [Lens 16] — This exclusion list omits GPT-OSS, which *is* served by Mantle**

"(GPT-5.x, Gemma 4, Grok)" leaves out `openai.gpt-oss-*` — **4 of the 13** Mantle models in the pinned 1.96.2 registry (`gpt-oss-120b`, `-20b`, `-safeguard-120b`, `-safeguard-20b`).

An operator configuring `openai.gpt-oss-120b` with a Bedrock Guardrail reads this, sees their family is not listed, and ships believing content moderation is enforced. It is not — `_strip_unsupported_mantle_params` drops it (see my comment on `base1.py`). This is the compliance-relevant direction of the error.

Note line 69 of this same file *does* list GPT-OSS, so the file contradicts itself.

Suggest either "(GPT-5.x, GPT-OSS, Gemma 4, Grok)" or dropping the enumeration entirely — the set is derived from a registry, so naming the mechanism ages better than naming the members.

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/bedrock.json
```diff
@@ -29,12 +29,12 @@
...
       "default": "amazon.titan-text-express-v1",
-      "description": "Model name. Refer to [Bedrock's documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html) for the list of available models."
+      "description": "Model name. Refer to [Bedrock's documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html) for the list of available models. Models served by the Bedrock Mantle endpoint (OpenAI GPT-5.x, Google Gemma 4, xAI Grok) are routed there automatically — enter the bare model ID, e.g. `openai.gpt-5.6-terra`, not a `us.`/`global.` cross-Region inference profile ID."
     },
     "model_id": {
       "type": "string",
       "title": "Application Inference Profile ARN",
-      "description": "Optional ARN for Application Inference Profile for cost tracking. Example: arn:aws:bedrock:us-east-1:000000000000:application-inference-profile/xxxx"
+      "description": "Optional ARN for Application Inference Profile for cost tracking. Example: arn:aws:bedrock:us-east-1:000000000000:application-inference-profile/xxxx. Not supported for Bedrock Mantle models (GPT-5.x, Gemma 4, Grok) — ignored with a warning."
```
johnyrahul: **[High] [Lens 16] — Same GPT-OSS omission as the guardrail description**

"(GPT-5.x, Gemma 4, Grok)" omits `openai.gpt-oss-*`, 4 of 13 Mantle models. Lower stakes than the guardrail case (this one only costs cost-attribution rather than moderation), but it should move with it.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -1081,6 +1090,128 @@ def _clean_str(value: str | None) -> str | None:
...
+
+    if not enable_thinking:
+        return False
+
+    # Both shapes require an unconstrained temperature.
+    metadata["temperature"] = 1
+
+    if not _is_bedrock_anthropic_model(metadata):
+        # LiteLLM routes `reasoning_effort` per family (Mantle ->
+        # reasoning.effort, gpt-oss -> additionalModelRequestFields, Nova 2 ->
```
johnyrahul, line 1173: **[Medium] [Lens 11 · 8] — Existing non-Claude adapters silently gain reasoning, and token spend, on deploy**

Per the PR description, the switch "never did anything outside Claude" — so any saved Bedrock adapter with `enable_thinking: true` on a non-Claude model has been a no-op, and some users will have left it on *precisely because* it did nothing.

After this deploy those same configs start emitting `reasoning_effort: "medium"`, which LiteLLM maps to a real wire field for Mantle GPT-5.x, gpt-oss on Converse, and Nova 2. Output content, latency and **token spend** all change on the next run, with no migration, no feature flag, no log line, and no rollback short of reverting the release.

The behaviour is the intended fix; the gap is that it activates invisibly. Suggest logging once per LLM construction when reasoning is enabled for a non-Anthropic Bedrock family, and calling the activation out in the release notes. Do you have a count of stored adapters with `enable_thinking: true` on non-Claude models? That number decides whether this is Medium or High.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -1184,9 +1293,23 @@ def validate(adapter_metadata: dict[str, "Any"]) -> dict[str, "Any"]:
         if not validated.get("guardrailConfig"):
             validated.pop("guardrailConfig", None)
 
-        # Add thinking config to final result if enabled
-        if enable_thinking and "thinking" in result_metadata:
-            validated["thinking"] = result_metadata["thinking"]
+        # Add the reasoning config to the final result if enabled. Exactly one
```
johnyrahul: **[Medium] [Lens 3 · 16] — "Exactly one of the two shapes is emitted" — the code emits both**

The non-Anthropic branch clears the other shape (`metadata.pop("thinking", None)`), but the Anthropic branch never pops `reasoning_effort`, so L1305 re-adds it alongside `thinking`. Reproduced:

```
validate({... 'model': 'anthropic.claude-3-7-sonnet-...', 'enable_thinking': True,
          'budget_tokens': 2048, 'reasoning_effort': 'high'})
  -> thinking         = {'type': 'enabled', 'budget_tokens': 2048}
     reasoning_effort = 'high'          # both present
```

Also sticky in the other direction: with the toggle **off**, a stray `reasoning_effort` latches reasoning back on and forces `temperature=1.0`.

Benign on the wire today only because `litellm.drop_params = True` discards the stray key on the Converse route — i.e. it rests on litellm's undocumented precedence. Three agents flagged it independently.

One-line fix: `metadata.pop("reasoning_effort", None)` in the Anthropic branch, mirroring the pop above. That makes the comment true.

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/bedrock.json
```diff
@@ -66,12 +66,12 @@
       "type": "boolean",
       "title": "Enable Extended Thinking",
       "default": false,
-      "description": "Enhance reasoning for complex tasks with step-by-step transparency. Available only for Claude 3.7 Sonnet."
+      "description": "Enhance reasoning for complex tasks with step-by-step transparency. For Claude models this enables extended thinking with the token budget below. For GPT-5.x (Bedrock Mantle), GPT-OSS and Nova 2 models it enables the model's own reasoning at medium effort; the budget below does not apply. Models without reasoning support ignore this setting."
```
johnyrahul: **[Medium] [Lens 16] — "Models without reasoning support ignore this setting" is false**

`_apply_bedrock_reasoning_config` sets `metadata["temperature"] = 1` *before* any family branch, so the toggle affects every model. Measured:

```
amazon.titan-text-express-v1  : off temp=0.1 -> on temp=1.0
meta.llama3-70b-instruct-v1:0 : off temp=0.1 -> on temp=1.0
```

An operator on Titan or Llama reads "ignore this setting", flips it to see what it does, and silently gets a different sampling temperature on every completion.

The same claim appears code-side in the `_apply_bedrock_reasoning_config` docstring ("which is why the switch never did anything outside Claude") — also false, and for the same reason: outside Claude it *did* pin temperature to 1, both before and after this PR.

Either scope the `temperature = 1` write to families that actually reason, or say that the toggle pins temperature to 1 for every model.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -1098,6 +1229,10 @@ class AWSBedrockLLMParameters(BaseChatCompletionParameters):
     # Declared so it survives Pydantic re-validation of kwargs.
     # Matches LiteLLM's Bedrock kwarg name, hence the mixed case.
     guardrailConfig: dict | None = None  # noqa: N815
+    # Reasoning knob for the non-Anthropic families (Mantle GPT-5.x, gpt-oss,
```
johnyrahul: **[Medium] [Lens 15 · 16] — This field is dead, and the comment states a mechanism that cannot operate**

`reasoning_effort` is stripped from `validation_metadata` **unconditionally** below, so Pydantic never receives it and the "otherwise Pydantic would drop it as an unknown field" branch is unreachable. Survival across re-validation comes entirely from the explicit re-attachment after `model_dump()`.

The comment ~60 lines below states the real — and opposite — consequence: *"`reasoning_effort` is a declared field, so Pydantic always re-emits it as None -- drop that placeholder first"*. The declaration is a liability the code then has to clean up, not a survival mechanism.

Empirically: deleting these four lines and running the suite gives **563 passed**, identical to baseline. Contrast `OpenAILLMParameters.validate`, which excludes the key *conditionally*, making its identical declaration genuinely load-bearing — the comment looks copied from there without re-checking.

Also **[Medium] [Lens 7]**: `str | None` is unconstrained. `'ULTRA'` validates then vanishes on the wire (silently no-op — the exact failure class this PR exists to fix), and `'none'` — semantically *disable* — latches reasoning **on** and forces `temperature=1`. Every sibling schema constrains this to `enum: ["low","medium","high"]` (`openai.json`, `mistral.json`, `azure.json`, `openrouter.json`, `nvidia_build.json`, `custom_openai.json`). `Literal["low","medium","high"] | None` would fail loudly at construction instead.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -1148,15 +1257,14 @@ def validate(adapter_metadata: dict[str, "Any"]) -> dict[str, "Any"]:
         # don't advertise the flag for other Bedrock families (Titan, Llama,
         # etc.). The LLM layer enforces the same model gate; this just keeps the
         # validated metadata honest.
-        # Check both ``model`` and ``model_id`` so callers routing through a
-        # Bedrock Application Inference Profile (opaque ARN in ``model``, Claude
-        # id in ``model_id``) still qualify.
-        bedrock_model_ids = " ".join(
-            str(result_metadata.get(field, "")) for field in _MODEL_ID_FIELDS
-        ).lower()
+        # The Anthropic check reuses the same helper as the reasoning config, so
```
johnyrahul, line 1272: **[Medium] [Lens 16] — "the two model gates cannot drift apart" — there is a third copy**

The preceding sentence says "The LLM layer enforces the same model gate", so "the two model gates" reads as {this one, the LLM layer's}. That reading is false: `llm.py:393` carries its own hardcoded `_BEDROCK_CACHE_MODEL_MARKERS = ("anthropic", "claude")` and its own inline `("model", "model_id")` tuple at L405-412. It does not call `_is_bedrock_anthropic_model` and does not use `_MODEL_ID_FIELDS`.

The dedup this PR performed was between the two call sites *inside* `validate()` only. So someone who later extends the helper — say, to recognise a Claude id inside an AIP ARN — will believe the LLM layer follows. It will not, and prompt caching will silently stay off for the newly-recognised configs.

Either narrow the claim to this function, or have `llm.py` call the helper and delete its marker tuple — which would make the original claim true.

johnyrahul (commented):
## Standardized review — 17-lens rubric

**Verdict: REQUEST CHANGES** — Critical: 0 · High: 2 · Medium: 12 · Low: 8 · Lenses run: 17/17

Solid PR — the exact-registry lookup over prefix matching is the right call given `openai.gpt-oss-120b` vs `openai.gpt-oss-120b-1:0`, and the `bedrock/` escape hatch keeps stored configs working. **These are review comments, not requested changes I have made** — nothing has been pushed. Please take or leave each on its merits; several may be deliberate choices I can't see from the diff, and I've flagged where I think the call is genuinely yours.

**Claims I verified rather than assumed:** `563 passed` on 1.96.2 (exact match) · reverting the routing breaks **15** tests, not 14 (the extra is `test_mantle_cost_is_resolvable`) — they genuinely pin the behaviour · all 9 lockfiles consistent, nothing beyond litellm 1.90.3→1.96.2, aiohttp 3.14.1→3.14.3, +`pydantic-settings` 2.15.0 · `boto3~=1.34.131` matches the existing `aws` extra exactly · the vendored cohere patch genuinely applies under 1.96.2 (`_SKIP_PATCH=False`, `handler.embedding` is `_patched_embedding`) and all six private litellm imports still resolve · `validate()` is a true fixed point.

---

### Unanchored findings

Both sit on lines this diff does not touch, so GitHub can't thread them.

**[Medium] [Lens 1 · 7] — `bedrock.json:184-197` still makes `budget_tokens` required for the families this PR says it doesn't apply to**

The `allOf` block makes `budget_tokens` (minimum 1024) **required** whenever `enable_thinking` is true, for every Bedrock model, with no family condition. This PR extends that switch to GPT-5.x / GPT-OSS / Nova 2 and its own description at `:69` now says "the budget below does not apply" for exactly those — while `:190-191` still describes the field as "Claude's internal reasoning". So two descriptions in the same file disagree, and an operator enabling reasoning on `openai.gpt-5.6-terra` is blocked by form validation until they type a number the adapter then discards.

The honest limit here is that JSON-Schema can't branch on a free-text model id. Minimally: drop `budget_tokens` from `required` (keep `default`) and retitle it Claude-only. Flagging rather than fixing because changing the form contract is your call.

**[Medium] [Lens 17] — The repo's own `adapter-ops` skill restates the Bedrock rules this PR changed, in six places**

These matter more than ordinary stale docs: `adapter-ops` is the documented path for "adding/removing models to existing adapters", so an engineer or agent following it will reproduce the defect this PR fixes.

| Site | Now says | Problem |
|---|---|---|
| `.claude/skills/adapter-ops/SKILL.md:47` | `\| Bedrock \| bedrock/ \| bedrock/anthropic.claude-v2 \|` | Prefix rule is no longer unconditional |
| `references/provider_capabilities.md:111` | `\| AWS Bedrock \| bedrock/ \| bedrock/ \|` | Same |
| `references/provider_capabilities.md:12` | `\| AWS Bedrock \| ❌ \| ✅ Claude 3.7+ \| …` | Reasoning column `❌`; the file's own Feature Definitions (L20-21) define Reasoning as `reasoning_effort`, which Bedrock now has. Compare the VertexAI row |
| `references/provider_capabilities.md:64-66` | `model, max_tokens, enable_thinking, budget_tokens` | Missing `reasoning_effort`; every sibling that has it lists it |
| `references/adapter_patterns.md:269-350` | "Thinking Configuration Pattern (**Anthropic, VertexAI, Bedrock**)" | The code block is, line for line, the pre-change `AWSBedrockLLMParameters.validate` this PR deleted — a *generative* site handing the next engineer the old shape |
| `SKILL.md:205-207` | "Add conditional `thinking` config in `validate()`" | Same rule, same staleness |

Not a finding: `provider_capabilities.md:32` is the *embedding* prefix table, and Mantle is LLM-only, so `bedrock/` there is still correct.

---

### Two claims I want to retract before you read the comments

I'd rather flag these than let them stand:

1. **There is no CI flake.** Some intermediate analysis suggested the routing suite failed 15 tests spontaneously "under network contention". It does not: I measured 5/5 clean runs, and it still passes with egress blocked, because the bundled map carries all 13 Mantle entries — so no registry state can produce `bedrock/openai.gpt-5.6-terra`. That failure only appears when the routing code is reverted. The registry risk in my first comment is **latent**, not active.
2. **The `os.environ` write at `test_bedrock_mantle_routing.py:254` does not leak.** `monkeypatch.setenv` already recorded the pre-test value, so teardown restores it. It's merely dead.

### Lens checklist

| # | Lens | Result |
|---|---|---|
| 1 | Spec & intent | See unanchored #1 |
| 2 | Architectural fit | See comments (registry as routing oracle) |
| 3 | Correctness & edge cases | See comments |
| 4 | Security | See comment on guardrail strip |
| 5 | Data integrity & migrations | N/A — no migrations, schema, or persisted-state change |
| 6 | Concurrency | Clean — no new shared mutable state |
| 7 | API & contract compatibility | See comments |
| 8 | Reliability & resilience | See comments |
| 9 | Performance & cost | Clean |
| 10 | Observability | See comments |
| 11 | Operational safety | See comment on silent reasoning activation |
| 12 | LLM/agent | See comments; cost resolves for the new prefix, `audit.py` prefix-strip correct |
| 13 | Testing | See comments |
| 14 | Dependencies & build | Clean — verified live under 1.96.2 |
| 15 | Code quality | See comments |
| 16 | Doc & comment accuracy | See comments |
| 17 | Cross-document consistency | See unanchored #2 |

### The two decisions I think are actually yours

1. **The routing oracle.** Pin `LITELLM_LOCAL_MODEL_COST_MAP=True` so routing follows the pinned wheel, or keep an in-repo id set with the registry as fallback? Both have real costs and it's a deployment-shape decision, not a code cleanup.
2. **Guardrail on a Mantle model — warn or hard-fail?** `llm.py:313` already turns `ValueError` into a UI-visible `SdkError`. Degrading a moderation control on a stdout warning seems the wrong default to me, but it's a product call.

Also worth confirming: were the non-Bedrock providers (Vertex, Azure, Mistral, Ollama, x2text) exercised against 1.96.2 anywhere? The bump crosses six minors and touches every adapter, justified by a Bedrock fix.

Happy to be wrong on any of these — please push back where I've misread the intent.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -1081,6 +1090,128 @@ def _clean_str(value: str | None) -> str | None:
...
+    """Drop Bedrock-only params that the Mantle endpoint cannot honour.
+
+    Guardrails and Application Inference Profiles are features of the
+    ``bedrock-runtime`` Converse/Invoke APIs; the Mantle endpoint speaks the
+    OpenAI wire format and has no field for either (AWS documents guardrails
+    as "Converse API only" for these models). LiteLLM silently discards them,
+    which would leave an operator believing a guardrail is enforced when it is
+    not -- so drop them here and say so loudly.
+    """
+    unsupported = [key for key in ("guardrailConfig", "model_id") if validated.get(key)]
```
johnyrahul: **[Low] [Lens 15] — The key list lives in two places and drifts into a `KeyError` inside a log call**

This tuple and the label dict below must stay in lockstep. Adding a third unsupported key here without adding its label raises `KeyError` from inside the `logger.warning` generator — at runtime, on the warning path, which is the path least likely to be exercised.

A module-level `_MANTLE_UNSUPPORTED: dict[str, str]` mapping key to label, iterated in both places, removes the tuple's independent existence.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -1081,6 +1090,128 @@ def _clean_str(value: str | None) -> str | None:
...
+def _is_bedrock_mantle_model(model: str) -> bool:
+    """Whether ``model`` is served by the AWS Bedrock Mantle endpoint.
+
+    Mantle (``bedrock-mantle.<region>.api.aws``) is a separate,
+    OpenAI-compatible endpoint from the classic ``bedrock-runtime``
+    Converse/Invoke surface, and LiteLLM exposes it as its own provider
+    (``bedrock_mantle/``). The two model-id namespaces are disjoint but look
+    confusingly similar -- ``openai.gpt-oss-120b-1:0`` is a Converse model
+    while ``openai.gpt-oss-120b`` is a Mantle one -- so membership is decided
+    by an **exact lookup** in LiteLLM's own registry, never by a family or
```
johnyrahul: **[Low] [Lens 16] — "never by a family or prefix match" is contradicted by the two prefix checks that open this function**

L1110-1113 *are* prefix matches, and they short-circuit before the registry lookup. The next paragraph does reconcile this, but the emphatic sentence is read first and reads as absolute. "…so membership **for an unprefixed id** is decided by an exact lookup…" would settle it.

Also worth stating the dependency this docstring currently omits: the enumeration of non-members (Claude, Titan, Nova, ARNs, `us.`/`global.`) reads as exhaustive, but the real set also includes *any genuine Mantle model missing from the registry this process loaded* — see my comment on the registry fetch.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -66,6 +66,15 @@
 # tracking — when callers route through an AIP, the standard model id often
 # only appears here, not in `model`.
 _MODEL_ID_FIELDS: tuple[str, ...] = ("model", "model_id")
+# LiteLLM provider prefixes for the two distinct AWS Bedrock endpoints:
+# `bedrock/` is the classic bedrock-runtime Converse/Invoke surface, while
+# `bedrock_mantle/` is the OpenAI-compatible bedrock-mantle endpoint that
+# serves the OpenAI GPT-5.x, Google Gemma 4 and xAI Grok families.
```
johnyrahul: **[Low] [Lens 16] — Family list omits GPT-OSS**

"OpenAI GPT-5.x, Google Gemma 4 and xAI Grok" omits the gpt-oss family, which is 4 of the 13 `bedrock_mantle/` registry entries — and `_is_bedrock_mantle_model`'s own docstring below uses `openai.gpt-oss-120b` as its worked example of a Mantle model. Same omission at `validate_model`'s docstring and `bedrock.json:32`.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
```diff
@@ -1081,6 +1090,128 @@ def _clean_str(value: str | None) -> str | None:
...
+    while every other reasoning-capable family (Mantle GPT-5.x, gpt-oss on
+    Converse, Nova 2) takes ``reasoning_effort``, which LiteLLM then maps to
+    that family's own wire field. Emitting the Anthropic shape for a
+    non-Anthropic model is a silent no-op -- LiteLLM drops it -- which is why
+    the switch never did anything outside Claude.
+
+    Returns whether reasoning ended up enabled, so the caller can decide which
+    keys to carry onto the validated payload.
+    """
+    enable_thinking = bool(metadata.get("enable_thinking", False))
```
johnyrahul, line 1155: **[Low] [Lens 3] — Explicit `enable_thinking=False` cannot switch reasoning off once `reasoning_effort` is present**

The latch promotes `enable_thinking` to true whenever `thinking` or `reasoning_effort` is present, and it runs *after* the explicit read — so an explicit `False` loses to the inferred `True`.

Flagging for the record rather than as a defect to fix here: the UI path is unaffected (stored metadata is the *form* metadata and the validated output is never written back, so the form's off-switch works correctly), and this latch is the established house pattern, identical to L339 (OpenAI), L751 and L1545. Worth changing repo-wide or not at all — doing it only here would create exactly the drift the surrounding comments are trying to prevent.

unstract/sdk1/src/unstract/sdk1/patches/litellm_cohere_timeout.py
```diff
@@ -29,10 +38,14 @@
...
 # warning so engineers know to verify compatibility.
-# Verified against litellm 1.90.3: async_embedding() and embedding() at
-# litellm/llms/cohere/embed/handler.py still do not forward `timeout`
-# to client.post() (the bug); the copied bodies below still match upstream.
-_PATCHED_LITELLM_VERSION = "1.90.3"
+# Verified against litellm 1.96.2: embedding() at
+# litellm/llms/cohere/embed/handler.py still does not forward `timeout`
+# to client.post() and still builds an untimed HTTPHandler (the bug);
+# async_embedding() is covered upstream via the client `params` but is
+# kept here so both paths stay consistent. The copied bodies below were
```
johnyrahul, line 45: **[Medium] [Lens 8 · 10] — On a version mismatch this degrades to an *unbounded* timeout, guarded only by an import-time warning**

`_SKIP_PATCH` skips the entire `else:` block on any version other than the exact pin, leaving Bedrock Cohere embeddings on upstream's untimed path — `HTTPHandler(concurrent_limit=1)` and `client.post()` with no timeout. The consequence is not a degraded result but a worker thread blocked forever on a socket, holding its Celery slot. The only guard is one `logger.warning` at module import during worker bootstrap, where it competes with the rest of the startup log.

The mechanism predates this PR, but moving the pin 1.90.3 → 1.96.2 re-arms it, which is why it is worth confirming now rather than after the next bump.

Worth surfacing the skipped state in the request path — a module flag `unstract.sdk1.embedding` re-warns from on the first Bedrock Cohere embedding call — or better, a default timeout floor that removes the dependency on the gate entirely.

**Verified accurate, for the record:** every claim in this file's version note holds against the installed 1.96.2. `async_embedding()` does now build its client with `params={"timeout": timeout}`; `embedding()` still does `HTTPHandler(concurrent_limit=1)` + untimed `client.post(...)`. The patch is genuinely still required on the sync path.

unstract/sdk1/src/unstract/sdk1/patches/litellm_cohere_timeout.py
```diff
@@ -73,7 +86,7 @@
 
     _DEFAULT_TIMEOUT = httpx.Timeout(None)
 
-    # Copied from litellm 1.83.10 cohere/embed/handler.py async_embedding().
+    # Copied from litellm 1.96.2 cohere/embed/handler.py async_embedding().
```
johnyrahul, line 89: **[Low] [Lens 16] — "ONLY CHANGE" understates the deviations**

The copied bodies also change `encoding: Any` → `encoding: object` and drop upstream's no-op `model = model`. Both are behaviourally inert, but "ONLY CHANGE" is the sentence the next engineer relies on when re-diffing against a new litellm — an inaccurate baseline makes that re-diff noisier than advertised. "ONLY BEHAVIOURAL CHANGE", plus a note of the two cosmetic deviations, would hold up.

unstract/sdk1/tests/test_bedrock_mantle_routing.py
```diff
@@ -0,0 +1,264 @@
...
+    "AWS_REGION_NAME",
+    "AWS_DEFAULT_REGION",
+)
+
+
+@pytest.fixture(autouse=True)
+def _isolated_litellm_env(monkeypatch: pytest.MonkeyPatch) -> Iterator[None]:
+    """Pin LiteLLM to its bundled cost map and strip ambient AWS config.
+
+    ``LITELLM_LOCAL_MODEL_COST_MAP`` keeps the test offline and deterministic:
```
johnyrahul: **[Medium] [Lens 13 · 16] — This fixture is inert; the docstring claims a hermeticity property the test does not have**

`litellm.model_cost` is built **once, at litellm import**, and `LITELLM_LOCAL_MODEL_COST_MAP` is read only there. A function-scoped `monkeypatch.setenv` runs long after litellm is in `sys.modules` — `tests/patches/test_litellm_cohere_timeout.py:7` (and `test_mock_response.py:12`) import it at collection time, before any fixture body executes.

Measured, from inside a test with the fixture applied:

```
full suite : env='True'  but  source='remote'  is_env_forced=False  n_models=3054
this file  :              source='local'   is_env_forced=True   n_models=2980
```

Two different registries for the same file, selected by which tests you run. The module docstring's "no AWS credentials, network access or spend are involved" (L6-7) is false in the same way.

`test_bedrock_adapter.py` has no pinning of its own at all, yet carries the same registry-dependent assertions.

Fix: `os.environ.setdefault("LITELLM_LOCAL_MODEL_COST_MAP", "True")` at module scope in `tests/conftest.py` — verified that conftest does not itself import litellm — or via pytest ini env. Then make the fixture self-verifying with `assert ..._cost_map_source_info.is_env_forced`, so it fails loudly if the pin ever stops taking effect rather than quietly reverting to the network.

Worth deciding explicitly: should these tests pin the *bundled* registry (reproducible, matches the `litellm==1.96.2` pin) or track upstream `main` (early warning for new AWS models, at the cost of CI nobody broke)? The fix differs.

unstract/sdk1/tests/test_bedrock_mantle_routing.py
```diff
@@ -0,0 +1,264 @@
...
+            payload = _responses_payload()
+        elif "/converse" in url:
+            payload = _converse_payload()
+        else:
+            payload = _chat_completion_payload()
+        return httpx.Response(200, json=payload, request=httpx.Request("POST", url))
+
+    completion_kwargs = AWSBedrockLLMParameters.validate(metadata)
+    # Mirrors unstract.sdk1.llm, which sets this globally so provider-specific
+    # rejections (e.g. gpt-5.x refusing temperature != 1) degrade to a drop.
```
johnyrahul: **[Medium] [Lens 13 · 3] — Setting `drop_params` here hides that production depends on it**

The adapter emits `temperature=0.1` (the `BaseChatCompletionParameters` default) for a non-reasoning Mantle call, and GPT-5.x rejects any temperature but 1. With `drop_params=False` the call raises **before any HTTP request is made**:

```
litellm.UnsupportedParamsError: gpt-5 models don't support temperature=0.1.
Only temperature=1 is supported.
```

So the entire Mantle path is load-bearing on `llm.py:100`. Because this test sets the flag itself rather than importing `unstract.sdk1.llm`, a refactor that removes or relocates that line breaks every GPT-5.x Bedrock completion in production while the suite stays green.

Suggest `import unstract.sdk1.llm  # noqa: F401` and asserting `litellm.drop_params is True` instead of setting it — and separately, a unit assertion that a Mantle model never carries `temperature != 1` out of `validate()`.

unstract/sdk1/tests/test_bedrock_mantle_routing.py
```diff
@@ -0,0 +1,264 @@
...
+    assert "messages" in req["body"]
+
+
+def test_mantle_cost_is_resolvable() -> None:
+    """A model string LiteLLM cannot price records usage at $0 silently.
+
+    ``audit.py`` looks cost up from the validated model string and swallows any
+    exception, so an unpriced prefix is revenue-affecting but invisible.
+    """
+    os.environ["LITELLM_LOCAL_MODEL_COST_MAP"] = "True"
```
johnyrahul: **[Low] [Lens 13] — Dead line**

This duplicates the autouse fixture and, like the fixture, has no effect on `litellm.model_cost` because litellm is already imported.

For the record, I checked whether it leaks and it does **not** — `monkeypatch.setenv` at L62 recorded the pre-test value, so teardown restores it. Safe to just delete once the fixture is fixed properly.

unstract/sdk1/tests/test_bedrock_adapter.py
```diff
@@ -648,3 +655,220 @@ def test_embedding_legacy_drops_stray_bearer_token() -> None:
...
+        "openai.gpt-oss-120b",
+        "google.gemma-4-31b",
+        "xai.grok-4.3",
+    ],
+)
+def test_llm_mantle_models_route_to_bedrock_mantle(model: str) -> None:
+    assert _validate_llm(model=model)["model"] == f"bedrock_mantle/{model}"
+
+
+@pytest.mark.parametrize(
```
johnyrahul, line 695: **[Medium] [Lens 13 · 3] — The registry-miss case that actually bites is untested and unlogged**

Every id in this parametrisation is one litellm already knows, so it covers "correctly excluded" but never "genuine Mantle model the registry has not catalogued". That is the case that bites: when AWS ships a Mantle model before litellm catalogues it, `_is_bedrock_mantle_model` returns `False` **silently** — no log, no warning — and the model routes to `bedrock/` and fails with an opaque AWS error.

Suggest a test that monkeypatches `litellm.model_cost` to drop `bedrock_mantle/openai.gpt-5.6-terra` and asserts the documented fallback plus a breadcrumb naming the model. Emitting that breadcrumb is the smaller half and turns an opaque 400 into a one-line diagnosis.

Related, on `us.`/`global.` (L704-707): the test asserts only the *string*. Pushed through this PR's own wire harness, `us.openai.gpt-5.6-terra` produces:

```
POST https://bedrock-runtime.us-east-1.amazonaws.com/model/us.openai.gpt-5.6-terra/invoke
body: {"model": "us.openai.gpt-5.6-terra", "messages": [...], "max_tokens": 64}
```

— the raw **Invoke** API carrying an OpenAI-shaped body, which nothing accepts. That is the id the AWS console surfaces, so it is the likeliest operator input, and the only guard is help text. A detection-only warning (strip a leading `us.`/`eu.`/`apac.`/`global.`; if the remainder *is* in the Mantle registry, warn) costs nothing and changes no routing.

unstract/sdk1/tests/test_bedrock_adapter.py
```diff
@@ -648,3 +655,220 @@ def test_embedding_legacy_drops_stray_bearer_token() -> None:
...
+    """
+    first = _validate_llm(model="openai.gpt-5.6-terra")
+    second = AWSBedrockLLMParameters.validate(dict(first))
+    assert first["model"] == second["model"] == "bedrock_mantle/openai.gpt-5.6-terra"
+
+
+# ── LLM: params the Mantle endpoint cannot honour ────────────────────────────
+
+
+def test_llm_mantle_strips_guardrail_and_aip_arn(
```
johnyrahul, line 764: **[Low] [Lens 13] — Warning-fires-once is load-bearing but unpinned**

I measured that the warning fires exactly once per LLM instance (1 after the first `validate()`, still 1 after re-validation), because the first pass already removed the keys. That is correct and it matters — `LLM.complete()` re-validates on every call (`llm.py:530, 655, 725, 806`), so a reorder that moved the strip after the Pydantic dump would emit this warning on *every completion*.

Cheap to pin: extend this test with a re-validation pass asserting `caplog` gains no second record.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
praveen-formido: Bedrock_mantle doesn't have support for guardrails yet. I will add a code comment on this and include it once it is available.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
praveen-formido: Fixed in ee8f2f5d2 — verified, and you were right that it emitted both.

Added `metadata.pop("reasoning_effort", None)` to the Anthropic branch, mirroring the pop in the other one. Before/after on your exact repro:

```
before: thinking={type: enabled, budget_tokens: 2048}  reasoning_effort=high   # both
after : thinking={type: enabled, budget_tokens: 2048}  reasoning_effort absent
```

Agreed on the reasoning too — resting on `litellm.drop_params` discarding the stray key is not a guarantee worth depending on, so I noted that in the comment rather than just making the code match the old claim.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
praveen-formido: Fixed in ee8f2f5d2 — you were right that it is dead, so I removed the declaration rather than repairing the comment.

Confirmed the mechanism cannot operate: `reasoning_effort` is stripped from `validation_metadata` unconditionally, so Pydantic never receives it, and survival across re-validation comes entirely from the re-attach after `model_dump()`. Suite is green without the four lines, and round-tripping still works (`validate(validate(x))` keeps `reasoning_effort=medium`). The comment now describes the re-attach, which is the real mechanism.

On the `str | None` half: removing the declaration makes it moot — there is no longer a field to constrain, and `reasoning_effort` has no UI field, so the only values in play are the `"medium"` we set or one a caller passes programmatically. If we ever expose it in the schema I agree it should land as `enum: ["low","medium","high"]` like the siblings.

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/bedrock.json
praveen-formido: Fixed in ee8f2f5d2 — measured it and got your numbers exactly.

```
amazon.titan-text-express-v1  : off temp=0.1 -> on temp=1.0
meta.llama3-70b-instruct-v1:0 : off temp=0.1 -> on temp=1.0
```

I chose to correct the claim rather than scope the write, because the unconditional `temperature = 1` predates this PR — scoping it would change behaviour for existing non-Claude adapters that have the toggle on, which is exactly the silent-activation problem you raise in the other comment. So the schema now says the toggle pins temperature to 1 for **every** Bedrock model, and the docstring says the switch produced no *reasoning* outside Claude but was never wholly inert.

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/bedrock.json
praveen-formido: Fixed in ee8f2f5d2 — this was the one I most wanted to get right, for the reason you give: it fails in the compliance-relevant direction.

Took your second suggestion and named the mechanism rather than the members, since the set is registry-derived: "**Not supported for models served by Bedrock Mantle** (including GPT-5.x, GPT-OSS, Gemma 4 and Grok) — the Mantle endpoint has no guardrail field, so the setting is ignored with a warning and no moderation is applied." The enumeration is now explicitly non-exhaustive and the consequence is stated outright.

unstract/sdk1/src/unstract/sdk1/adapters/llm1/static/bedrock.json
praveen-formido: Fixed in ee8f2f5d2, same wording treatment as the guardrail description — mechanism first, families as a non-exhaustive parenthetical.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
praveen-formido: Fixed in ee8f2f5d2. Added GPT-OSS and a pointer that the list is not the source of truth (`_is_bedrock_mantle_model` is), so the comment does not drift again as AWS adds models. Same fix applied to `validate_model`'s docstring and `bedrock.json:32`.

unstract/sdk1/tests/test_bedrock_mantle_routing.py
praveen-formido: Fixed in ee8f2f5d2 — reproduced your measurement exactly before changing anything:

```
remote (no env)          : 3054 models
bundled (env at import)  : 2980 models
env set AFTER import     : 3054 models   <- what the fixture was doing
```

Moved the pin to module scope in `tests/conftest.py` (confirmed conftest does not itself import litellm) and made the fixture self-verifying with an assert, so it fails loudly rather than silently reverting to the network. Verified the pin now takes: importing conftest then litellm gives 2980.

On your explicit question — I pinned the **bundled** registry. Reproducibility and consistency with the `litellm==1.96.2` pin matter more here than early warning, and a CI failure caused by AWS shipping a model is a signal nobody can act on from a red build. I used `setdefault` rather than a hard assignment so `LITELLM_LOCAL_MODEL_COST_MAP=False` still opts into the live registry for anyone who wants that check deliberately.

Also fixed the module docstring, which now says why the claim holds rather than just asserting it.

unstract/sdk1/tests/test_bedrock_mantle_routing.py
praveen-formido: Fixed in ee8f2f5d2 — good catch, and the failure mode you describe is the one that matters: suite green, production broken.

The harness now imports `unstract.sdk1.llm` for its import-time side effect and **asserts** `litellm.drop_params is True` instead of setting it, with the reasoning in the assert message. Confirmed your repro first — with `drop_params=False` the call raises `UnsupportedParamsError` before any HTTP request, so the dependency is real.

On the separate suggestion of a unit assertion that a Mantle model never carries `temperature != 1` out of `validate()`: I did not add it, because it would encode something that is currently false — a non-reasoning Mantle call legitimately emits `temperature=0.1` and relies on `drop_params`. Asserting otherwise would mean changing `validate()` to force `temperature=1` for Mantle models, which is a behaviour change I would rather not fold into a review pass. Happy to take it as a follow-up if you think the adapter should own that rather than delegating to litellm.

unstract/sdk1/tests/test_bedrock_mantle_routing.py
praveen-formido: Fixed in ee8f2f5d2 — deleted. Thanks for checking the leak question and reporting back; agreed it was dead rather than harmful.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
praveen-formido: Fixed in ee8f2f5d2. Added a module-level `_MANTLE_UNSUPPORTED_PARAMS: dict[str, str]` mapping key to label, iterated in both the detection and the message, so the tuple no longer has independent existence and a key cannot be added without its label. Low severity as you say, but a `KeyError` raised from inside a `logger.warning` on the rarely-exercised path is a nasty way to find out.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
praveen-formido: Fixed in ee8f2f5d2 — fair, the emphatic sentence is what gets read. Now reads "membership **for an unprefixed id** is decided by an exact lookup", with the prefix short-circuit stated in the same breath.

Also took your second point: the non-member enumeration now says explicitly that the set is not closed and includes any genuine Mantle model missing from the registry the process loaded.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
praveen-formido: Fixed in ee8f2f5d2 — you are right, and I checked: `llm.py` carries `_BEDROCK_CACHE_MODEL_MARKERS` plus its own inline `("model", "model_id")` tuple, and calls neither the helper nor `_MODEL_ID_FIELDS`.

I narrowed the claim rather than unifying the third copy. The comment now says the two gates *inside this function* cannot drift, and names `llm.py`'s markers as a separate gate that extending the helper will not move — which is precisely the trap you flagged. I kept the unification out of scope because it touches the prompt-caching path, which is not what this PR is about; happy to do it as a follow-up if you would rather have the stronger claim be true.

unstract/sdk1/tests/test_bedrock_adapter.py
praveen-formido: Took the first half in ee8f2f5d2, declined the second.

**Added** `test_llm_unknown_mantle_model_falls_back_to_standard_bedrock`, which monkeypatches `bedrock_mantle/openai.gpt-5.6-terra` out of `litellm.model_cost` and asserts the `bedrock/` fallback. You are right that this is the case that actually bites, and it was untested.

**Did not add** the breadcrumb log or the `us.`/`global.` detection warning. Both are new runtime behaviour rather than a fix to what this PR changed, and the breadcrumb in particular needs care to avoid firing on every legitimate `bedrock/` model — Claude, Titan and Nova all take the same code path as an uncatalogued Mantle model, so "not in the Mantle registry" is not on its own a signal worth logging. The `us.`/`global.` heuristic you describe (strip the prefix, check the remainder) does not have that problem and I think it is the right shape, but it is a new diagnostic rather than a correction, so I would rather it land on its own where it can be reviewed as such. Filed mentally as a follow-up — say the word if you want it in this PR instead.

Thanks for pushing the CRIS id through the wire harness; the raw Invoke call with an OpenAI-shaped body is a more concrete statement of the limitation than the one in the PR description.

unstract/sdk1/tests/test_bedrock_adapter.py
praveen-formido: Added in ee8f2f5d2. `test_llm_mantle_strip_warns_once_not_per_completion` re-validates the already-validated payload and asserts `caplog` gains no second record. Cheap as you say, and it pins something genuinely load-bearing — a reorder that put the strip after the Pydantic dump would warn on every completion, which is the kind of thing that gets noticed only as log volume.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
praveen-formido: Valid, and I have not added code for it — here is the reasoning.

The activation is real: any stored adapter with `enable_thinking: true` on a non-Claude model starts emitting `reasoning_effort: "medium"` after this deploy, with the latency and token-spend consequences you describe. I agree the gap is that it happens invisibly.

I did not add the per-construction log because it is observability for a one-off migration rather than a correction to what this PR changed, and a line that fires on every LLM construction forever is a high price for a signal that matters for one release. What I have done instead is make the schema honest about what the toggle now does, so the next operator to look at it sees the reasoning behaviour and the temperature effect stated plainly.

I do not have a count of affected stored adapters — that needs a production query I would rather run than guess at, and it is the number that decides whether this needs a migration or just release notes. I will raise it with the team and get the activation into the release notes either way. If the count is non-trivial I would rather gate it than log it.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
praveen-formido: Agreed on all counts, and not changing it here — thanks for pre-empting the question by checking the UI path.

Confirmed your reading: the latch runs after the explicit read, so `enable_thinking=False` loses to an inferred `True`. It is identical at L339 (OpenAI), L751 and L1545, and changing it only for Bedrock would create exactly the drift the surrounding comments exist to prevent. Repo-wide or not at all is the right call, and it is not this PR.

unstract/sdk1/src/unstract/sdk1/patches/litellm_cohere_timeout.py
praveen-formido: Fixed in ee8f2f5d2. Verified both deviations against the installed 1.96.2 (`encoding: Any` -> `object`, upstream's no-op `model = model` dropped) and reworded to "ONLY BEHAVIOURAL CHANGE", with the two cosmetic deviations spelled out so the next re-diff has an accurate baseline. That is the sentence someone will trust on the next bump, so it should be exact.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
praveen-formido: Confirmed the mechanism, declined the fix in this PR — I think the trade-off cuts the other way, and I would rather it be an explicit decision than a side effect of a Bedrock fix.

Reproduced your measurement: 3054 models from the network, 2980 bundled, and setting the env after import changes nothing.

The reason I did not pin `LITELLM_LOCAL_MODEL_COST_MAP=True` in the backend/worker env is that the same map is the pricing oracle, not just the routing oracle. Pinning it makes routing deterministic but freezes pricing to the wheel, so any model AWS or any other provider ships after this pin silently records **$0** in `audit.py` until we bump litellm — trading a latent routing miss for a guaranteed, ongoing revenue miss across every provider. That is a worse default, and it is a platform-wide change reached through a Bedrock PR.

The in-repo id set has the same freshness problem plus a second source of truth to maintain.

What I did instead: the fallback is now tested (`test_llm_unknown_mantle_model_falls_back_to_standard_bedrock`) and the docstring states that the non-member set is not closed and includes any Mantle model missing from the loaded registry. So the behaviour is pinned and documented rather than incidental.

Worth its own issue where the pricing side gets weighed properly — I do not think a Bedrock routing PR is the right place to decide how the platform sources its cost map. Flagging your point that two workers can route the same adapter differently: correct, and the strongest argument for eventually pinning; it needs the pricing question answered alongside.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
praveen-formido: Not changing the default here — this is the product call you identify, and I do not think a review pass is the right place to make it unilaterally.

Everything you describe is accurate: the warning goes to Zipstack's log rather than the tenant's, the form still shows the guardrail populated, and `llm.py:313` does give `validate()` a UI-visible path via `ValueError` -> `SdkError`.

The argument against hard-failing today is blast radius in the wrong direction: an adapter with a guardrail set on a Mantle model currently *runs* (unmoderated), and hard-failing turns that into a broken adapter at the next execution, with no migration and no warning to the operator beforehand. That may well still be the right trade for a moderation control — but it breaks running configurations, so it wants a deliberate decision and a release note, not a quiet flip.

What I did instead is close the information gap that made this dangerous: the guardrail field description now states outright that Mantle models ignore it and no moderation is applied, and it names GPT-OSS, which the previous wording omitted — that omission was the part that could actually mislead someone into shipping unmoderated.

I also agree with your instinct to separate the two keys; `model_id` is cost attribution and warn-and-drop is proportionate there, whereas `guardrailConfig` is not. If the team wants the hard fail I will take it as a follow-up, scoped to `guardrailConfig` only.

unstract/sdk1/src/unstract/sdk1/patches/litellm_cohere_timeout.py
praveen-formido: Valid, and out of scope for this PR — but worth recording that you confirmed the mechanism.

The unbounded-timeout degradation on version mismatch predates this change; the bump re-arms it rather than introducing it, and altering the patch's failure mode (a timeout floor, or re-warning from the request path) changes behaviour for every Bedrock Cohere embedding user. That deserves its own change where it can be reviewed on its own terms rather than riding along with a routing fix.

I did take the documentation half in ee8f2f5d2 — the version note and the "ONLY CHANGE" claims are now accurate against 1.96.2, so whoever picks this up next has a correct baseline. Thanks for independently verifying the sync path is still broken upstream; that was the load-bearing check for keeping the patch at all.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
praveen-formido: Thanks — that matches what I found: the Mantle endpoint has no guardrail field at all, and AWS documents guardrails as Converse-API-only for these models, so there is nothing for LiteLLM to forward even when it wants to.

I have made the schema say so explicitly rather than leaving it implicit (`guardrail_identifier` now states that Mantle models ignore it and no moderation is applied, and names GPT-OSS, which the earlier wording missed). When AWS does ship guardrail support on Mantle, the place to revisit is `_strip_unsupported_mantle_params` and its `_MANTLE_UNSUPPORTED_PARAMS` mapping in `base1.py` — dropping the `guardrailConfig` entry there is the whole change.

praveen-formido:
Thanks — this was a genuinely useful review. I verified every claim before acting on it and found them all accurate; replies are on each thread. All changes are in a single commit (`ee8f2f5d2`) to keep Greptile from re-triggering.

## The two unanchored findings — both taken

**`budget_tokens` required for families it doesn't apply to.** Fixed. You're right that this is a defect this PR introduced: extending the switch to GPT-5.x/GPT-OSS/Nova 2 while the `allOf` still made `budget_tokens` required meant those operators were blocked on a number the adapter then discards, and the two descriptions in the file contradicted each other. Dropped it from `required` (kept `default`) and retitled it "Thinking Budget Tokens (Claude only)".

**`adapter-ops` skill restating the rules this PR changed.** Fixed in all the sites you listed. I agree these matter more than ordinary stale docs — it's the documented path for "adding models to existing adapters", so it would have handed the next engineer the exact shape this PR removes. `adapter_patterns.md` was the worst of them: the code block was the pre-change `validate()` line for line, under a heading that named Bedrock. It's now scoped to Anthropic/VertexAI with an explicit warning that the reasoning shape is provider-specific and that emitting the wrong one is a silent no-op, plus the mutual-exclusion rule the real implementation had to learn. Prefix tables now carry `bedrock_mantle/`, and the reasoning column and param list are corrected. And thanks for checking `provider_capabilities.md:32` — agreed, the embedding table's `bedrock/` is still correct since Mantle is LLM-only, so I left it.

## Your closing question: non-Bedrock providers against 1.96.2

Fair challenge — a six-minor bump justified by a Bedrock fix. The 565-test suite is the coverage, and it exercises the adapter layer for OpenAI, Azure, VertexAI/Gemini, Mistral, Anthropic, Ollama, the branded OpenAI-compatible adapters, and the embedding adapters, all green on 1.96.2. What that does **not** cover is live provider traffic — the tests validate parameter shaping and mocked wire behaviour, not real completions against each vendor.

I also verified the whole platform boots on the bump: built all images and brought up 24 containers, with the routing confirmed live inside both the backend and a worker.

So: adapter-layer regression risk is covered, live cross-provider behaviour is not, and I'd flag that as the thing worth a smoke pass during RC rather than something I can close out here.

## Declined, with reasoning on each thread

- **Pinning the cost map** — the same map is the pricing oracle. Pinning trades a latent routing miss for guaranteed $0 cost rows on any post-pin model, across every provider. Deserves its own issue where the pricing side gets weighed.
- **Hard-failing on a Mantle guardrail** — breaks currently-running adapters with no migration; needs a deliberate call and a release note. I closed the information gap in the schema instead, including the GPT-OSS omission that was the genuinely misleading part.
- **`enable_thinking=False` latch** — house pattern; repo-wide or not at all, as you said.
- **Cohere version-mismatch timeout** — pre-existing mechanism, separate change. Took the documentation half.
- **Silent reasoning activation** — no code, but I'll get the production count and the release note; if the count is non-trivial I'd rather gate it than log it.
- **`us.`/`global.` detection warning** — right shape, but new diagnostic behaviour rather than a correction. Happy to add it here if you'd prefer.

Two corrections to my own PR description while I'm here: it's **15** tests that fail on revert, not 14, and the hermeticity claim I made for the routing tests was false until this commit — your measurement of the inert fixture was the most valuable finding in the review, since it meant those tests were silently reading the network registry.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
praveen-formido: Got the number you asked for. US prod, read replica:

```
Bedrock LLM adapters                    31   (12 orgs, 1 flagged in-use)
  wired to a Prompt Studio profile      22
  ever produced a usage row             27
  produced a usage row in last 90d       9
```

The decisive cut is by *model actually invoked* — `usage.model_name` records it, so the non-Claude half is answerable even though `enable_thinking` is not (it lives in the Fernet-encrypted `adapter_metadata_b`, so no SQL predicate reaches it).

Every Bedrock model invoked in the last 180 days, by call volume:

```
us.anthropic.claude-sonnet-4-6                          32796
arn:...inference-profile/us.anthropic.claude-sonnet-4-6  5763
us.anthropic.claude-sonnet-4-5-20250929-v1:0             3800
arn:...us.anthropic.claude-sonnet-4-20250514-v1:0        3627
us.anthropic.claude-3-7-sonnet-20250219-v1:0             2023
arn:...us.anthropic.claude-3-7-sonnet-20250219-v1:0      1834
us.anthropic.claude-opus-4-6-v1                          1645
openai.gpt-oss-120b-1:0                                  1550   <- the only non-Claude
global.anthropic.claude-sonnet-5                         1095
anthropic.claude-3-haiku-20240307-v1:0                    270
us.anthropic.claude-sonnet-4-20250514-v1:0                258
us.anthropic.claude-sonnet-5                               13
global.anthropic.claude-opus-4-7                            4
```

**One non-Claude adapter, last used 2026-04-27.** In the last 90 days: `non_claude_calls = 0`, `non_claude_adapters = 0`, against 32,819 Bedrock calls total.

So the affected set on US prod is a single dormant adapter, and only if it also has `enable_thinking` on — which I cannot confirm without decrypting, but it is an upper bound of one either way. That answers your Medium-or-High question: **Medium**, and I do not think it warrants a gate or a permanent per-construction log line. Release note it is.

Two caveats I would not want to overstate this with:
- **EU is unmeasured** — the session cookie is per-Superset-instance and mine is US-only. Worth a repeat before the release; I would be surprised if the shape differed, but I have not checked it.
- Note the one non-Claude id is `openai.gpt-oss-120b-1:0`, the **Converse** variant, so this PR keeps it on `bedrock/`. Its exposure is purely the reasoning mapping (`additionalModelRequestFields.reasoning_effort`), not the routing change.

unstract/sdk1/src/unstract/sdk1/adapters/base1.py
johnyrahul: Thanks — and agreed that adding guardrail support once Mantle exposes a field is the right end state. Not blocking on this, but I want to restate the concern precisely, because I think the reply answers a different question than the one I was asking.

"Mantle doesn't support guardrails yet" is the *cause*. My concern was the *signal*: what an operator who has already configured one experiences today. Concretely — they set a Guardrail on a Bedrock adapter while on Claude, later switch the model to `openai.gpt-5.6-terra`, and from that point every prompt and completion runs unmoderated. The form still shows the guardrail ID populated, and the only runtime trace is `logger.warning` to backend/worker stdout, which in Cloud is our log rather than the tenant's.

Two things have genuinely improved that in `ee8f2f5d2`, and they're most of what I was after:

- The description now states the consequence outright — "the setting is ignored with a warning and **no moderation is applied**" — rather than the softer "ignored with a warning".
- Naming the mechanism instead of enumerating families closes the GPT-OSS hole, which was the part that failed in the compliance-relevant direction.

So the *config-time* signal is now honest. What's still absent is any signal on the path where the change actually happens — the model switch. Options, roughly in order of cost:

1. Surface it in the adapter test-connection response, so it appears at the moment of the edit.
2. Return it in the execution log stream alongside other per-run warnings, so it reaches the tenant.
3. `raise ValueError` — `llm.py:313` already converts that into a UI-visible `SdkError`. Strictest, and I accept it may be too strict if anyone legitimately runs a Mantle model with a stale guardrail field they don't care about.

Happy for this to be a follow-up rather than something this PR carries — the doc change does the heavy lifting, and a moderation-signal decision deserves the same "own change, own review" treatment you've applied elsewhere here. Flagging mainly so the resolved thread doesn't imply the runtime-signal question was settled, since I don't think it has been.

For the record, everything else on this PR verified clean on my side: the both-shapes fix holds on the original repro and on a Mantle→Claude family switch, the conftest pin genuinely takes in a full-suite run (`source=local, env_forced=True, 2980 models` — it was `remote/False/3054`), and the suite is green at 565.

johnyrahul (approved):
## Approving — verified against `ee8f2f5d2`

Re-reviewed the follow-up rather than taking the "fixed in" claims at face value. **15 of the findings are fixed and independently confirmed; 5 were declined or deferred with reasoning I accept.**

**Verified behaviourally, not by reading the diff:**

- **Both-shapes bug** — `BOTH PRESENT = False` on the original repro *and* on a Mantle→Claude family switch. Idempotence still holds and `reasoning_effort` still round-trips.
- **Test hermeticity** — the one that mattered: the pin now takes in a **full-suite** run (`source=local, is_env_forced=True, 2980 models`, previously `remote/False/3054`). Self-verifying assert means it fails loudly if that ever regresses.
- **Dead field** — genuinely removed, not repaired in comment only.
- **Lens 17** — all six `adapter-ops` sites corrected, including dropping Bedrock from the thinking-pattern heading and adding the "check before copying" callout. `provider_capabilities.md:32` correctly left alone, since the embedding table is right as it stands.
- **`budget_tokens`** — `required` dropped and retitled, picked up from the summary body without a thread prompting it.

Suite green at **565 passed** (+2 new tests, both pinning things that were genuinely unpinned).

**On the registry finding — your counter-argument is better than my recommendation, and I'm withdrawing it.** I framed `litellm.model_cost` as a routing oracle; you're right that it is simultaneously the pricing oracle, and that pinning it would freeze `audit.py` pricing and silently record $0 for any model shipped after the pin, across every provider. Trading a latent routing miss for a guaranteed ongoing revenue miss is the worse default, and it isn't a decision a Bedrock routing PR should make. Pinning it in tests while leaving production live is the right split.

One option neither of us raised, for whenever that issue gets written up: the two oracles could be decoupled — resolve the *routing* membership check against the bundled map (cached once in a module-level frozenset) while leaving `litellm.model_cost` on the live fetch for pricing. That gets deterministic, version-pinned routing without freezing cost data. Worth weighing there rather than here.

You were also right to reject my breadcrumb suggestion — "not in the Mantle registry" isn't a signal on its own, since Claude, Titan and Nova take that same path.

### Open, by agreement — approving does not close these

1. **Guardrail runtime signal** — the config-time text is now honest, but a model switch that silently disables an already-configured guardrail still produces no tenant-visible signal. Detail in the thread. Note that thread is marked resolved — that's your call and I've left it that way — so this list is the durable record of the open question, not the thread state.
2. **Silent reasoning activation** — stored adapters with `enable_thinking: true` on non-Claude models gain reasoning, latency and token spend on deploy. Pending the production count and release notes; gate rather than log if the count is non-trivial.
3. **Cohere patch** unbounded-timeout degradation on version mismatch — pre-existing, re-armed by the pin move.
4. **`us.`/`global.` detection warning** — agreed shape, deferred to its own change.

Approving on the strength of the code in this PR. Nice work on the follow-up — the replies engaged with the substance rather than just the asks, and pushing back on two of them was the right call.

Deepak-Kesavan:
@pk-zipstack Please handle the sonar issue.

PR #2250, opened 2026-08-19
UN-4017 [FIX] Recover single-pass output from malformed LLM JSON, with always-on parse diagnostics

Deepak-Kesavan:
## What

- Let the JSON repair helper recover a dict-shaped result from a malformed LLM response, via an **opt-in** contract — the legacy behaviour is preserved verbatim for every existing caller.
- Add always-on, PII-free parse diagnostics, plus an optional raw-response capture gated to non-production.
- Carry the **whisper hash** in that trace, so a bad parse can be correlated with the LLMWhisperer extraction behind it without anyone having to dig on the LLMWhisperer side.
- Declare `json-repair`, which the workers imported but nothing declared.

## Why

- Single-pass extraction passes the parsed LLM response straight through as the outputs map. When the model wraps its JSON in prose, a reasoning block, or a stray fence marker, the repair returns a list and the backend 500s at `output_manager_helper.py:179`. Confirmed in production.
- The repair heuristics encode ~2 years of observed edge cases, so they must not change. The requirement was to extend, never to alter.
- The response shapes causing this are open-ended and grow with each model, so the design has to fail safe on shapes nobody has seen rather than enumerate them.
- Diagnosing these needed information the logs did not carry. A highlight or confidence mismatch is reported against a response that parsed *fine*, so failure-only capture would miss exactly those cases.
- The trace said what the parse did, but not *which extraction it was reading*. Tying the two together meant producing the raw payload — which production never logs — so in practice it meant a per-incident request to the LLMWhisperer side. The whisper hash is already in hand at the call site and closes that gap here, on our side only.
- CI proved `json_repair` was installed nowhere: every contract test returned a raw string. The image builds the same way, so wherever it is absent every JSON prompt silently degrades to a raw string and stores `{}` with no error.

## How

- `_repair_legacy` holds the existing logic verbatim, with a comment recording *why* the `"[" + json_str` wrap exists (fragmentation recovery) so nobody "cleans it up".
- With a `JsonContract`, the helper offers competing candidates — legacy parse, unwrapped parse, five salvage strategies — and picks the one carrying the most expected prompt keys. Scoring rather than first-match matters: "first dict" returns a schema example quoted in the model's prose instead of the answer.
- Responses carrying `// 0x..` metadata are never reshaped. Those comments are mapped onto the parsed structure **positionally**, so dropping or merging an element silently attaches line numbers to the wrong fields — worse than the 500. Detection requires the hex payload, because matching a bare `//` would treat ordinary document values (`"N/A // pending"`, a path) as annotated and silently switch the fix off.
- Adding a newly observed shape is one `@salvager` plus a corpus entry; `_repair_legacy` stays closed for modification.
- Diagnostics log structure only — shapes, key counts, which parse step produced what — at WARNING on failure and DEBUG on success, carrying `request_id` and `execution_id`. Only contract-declared keys are echoed, because repair promotes document text into key position on malformed input.
- Raw capture sits behind `DEBUG_LOG_RAW_LLM_RESPONSE` and a non-prod `DEPLOYMENT_ENV`, defaulting to `production` so unset and unrecognised values fail closed.
- `whisper_hash` is an optional argument on the helper and flows into the always-on line, the raw log sink and the webhook payload. It is the one identifier exempt from the structure-only rule, and deliberately so: it is an opaque extraction id, not derived from document text. Absent (a non-LLMWhisperer extractor) it logs `-` rather than an empty field that reads like a hash. `handle_json` reads it from the metadata dict `_run_completion` has already stamped; the single-pass caller in the cloud plugin can pass it the same way.

- No behavioural change for any existing caller, and this is the claim the PR is built to defend. Passing no contract makes no salvager reachable, so the returned value cannot differ. Nothing in this repo passes a contract today.
- Evidence: a frozen copy of the pre-change implementation lives in the tests, and a differential fuzz compares them across **2102 inputs including bytes — 0 differences**. Any future edit to `_repair_legacy` fails that pin.
- Two risks found in review and fixed, both listed here because they are the kind that would otherwise bite silently: the diagnostics narrowed the accepted input (`bytes` used to parse, then raised `TypeError`), and the annotation guard was broad enough to disable the fix for documents containing `//` in a value.
- Diagnostics add log records on every repair. They emit no document values, are wrapped so they cannot raise, and cost ~0.8ms on a 55KB payload.
- `whisper_hash` is diagnostics-only and never reaches the parse. Pinned over the whole corpus on both the contract and no-contract paths: the value returned with a hash equals the value returned without one.
- Declaring `json-repair` changes what ships: builds that were silently running without it will now have it. That is the intended fix, but it is a real behaviour change for any such environment.

## Database Migrations

- None.

## Env Config

- `DEBUG_LOG_RAW_LLM_RESPONSE` (default `false`), `DEBUG_LOG_RAW_LLM_SINK` (`log` | `webhook`), `DEBUG_RAW_LLM_WEBHOOK_URL`, `DEBUG_LOG_RAW_LLM_MAX_CHARS` (`0` = no clipping), `DEPLOYMENT_ENV` (default `production`). All documented in `workers/sample.env`; chart wiring in Zipstack/unstract-cloud#1738.

## Relevant Docs

- None.

## Related Issues or PRs

- UN-4017
- Chart wiring: Zipstack/unstract-cloud#1738
- The minimal ship-now guards, kept separate: #2251 (**merged**) and Zipstack/unstract-cloud#1739

## Dependencies Versions

- Declares `json-repair>=0.42` in `workers/pyproject.toml`; the lock resolves 0.63.4. A lower bound rather than a pin, by decision — the salvagers are verified against 0.42.0, 0.57.1 and 0.63.x.
- The root `uv.lock` change is now **11 added lines**, not the ~2100-line rewrite this branch previously carried. That rewrite was a `revision = 3` → `revision = 1` downgrade emitted by the lock automation, which is pinned to `uv 0.6.14`; it was also the branch's only merge conflict with main. Regenerating with a current uv keeps `revision = 3` and adds only the `json-repair` entry, and `uv 0.6.14 lock --check` accepts the result — so `uv-lock.sh`, which gates on exactly that check, stays a no-op instead of re-introducing the churn. Pinning that workflow's uv is still worth doing separately.

## Notes on Testing

- 251 tests: legacy characterisation, model explanations, code blocks, token-limit truncation at each awkward boundary, annotated responses, the debug sinks, the raw-capture gate, and the whisper-hash correlation — including the wiring test that `handle_json` actually forwards the hash, which the helper's own tests cannot catch.
- Differential fuzz on the no-contract path: 2102 inputs, 0 differences.
- **Absolute-value** pins against dependency drift. The characterisation pin cannot catch it — `_original_impl` calls the same installed `json_repair`, so a version bump moves both sides together. One of those pins documents current behaviour rather than endorsing it: on the no-contract path a fenced response is shredded into `["json\n{", 'invoice_number": "INV-001', {...}]`.
- No activated call site in this repo, deliberately: the single-pass caller lives in the cloud plugin. This PR is the mechanism, the regression safety net, and the diagnostics.

## Checklist

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 1
 requires-python = "==3.12.*"
 
 [[package]]
@@ -14,9 +14,9 @@ dependencies = [
     { name = "azure-storage-blob" },
     { name = "fsspec" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/b4/1e/6d5146676044247af566fa5843b335b1a647e6446070cec9c8b61c31b369/adlfs-2024.7.0.tar.gz", hash = "sha256:106995b91f0eb5e775bcd5957d180d9a14faef3271a063b1f65c66fd5ab05ddf", size = 48588, upload-time = "2024-07-22T12:10:33.849Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/b4/1e/6d5146676044247af566fa5843b335b1a647e6446070cec9c8b61c31b369/adlfs-2024.7.0.tar.gz", hash = "sha256:106995b91f0eb5e775bcd5957d180d9a14faef3271a063b1f65c66fd5ab05ddf", size = 48588 }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/6f/51/a71c457bd0bc8af3e522b6999ff300852c7c446e384fd9904b0794f875df/adlfs-2024.7.0-py3-none-any.whl", hash = "sha256:2005c8e124fda3948f2a6abb2dbebb2c936d2d821acaca6afd61932edfa9bc07", size = 41349, upload-time = "2024-07-22T12:10:32.226Z" },
+    { url = "https://files.pythonhosted.org/packages/6f/51/a71c457bd0bc8af3e522b6999ff300852c7c446e384fd9904b0794f875df/adlfs-2024.7.0-py3-none-any.whl", hash = "sha256:2005c8e124fda3948f2a6abb2dbebb2c936d2d821acaca6afd61932edfa9bc07", size = 41349 },
 ]
 
 [[package]]
@@ -29,9 +29,9 @@ dependencies = [
     { name = "botocore" },
     { name = "wrapt" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/ea/17/2f6305cc52976dea8156b56badc3602f162f86693a6cc8badc20d2c5cfe6/aiobotocore-2.13.3.tar.gz", hash = "sha256:ac5620f93cc3e7c2aef7c67ba2bb74035ff8d49ee2325821daed13b3dd82a473", size = 106736, upload-time = "2024-08-22T20:42:13.086Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/ea/17/2f6305cc52976dea8156b56badc3602f162f86693a6cc8badc20d2c5cfe6/aiobotocore-2.13.3.tar.gz", hash = "sha256:ac5620f93cc3e7c2aef7c67ba2bb74035ff8d49ee2325821daed13b3dd82a473", size = 106736 }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/81/98/1c1a1ba5d74f87cbabbf207d36a98551b082b8b187dc35f7808b43f4cf1a/aiobotocore-2.13.3-py3-none-any.whl", hash = "sha256:1272f765fd9414e1a68f8add71978367db94e17e36c3bf629cf1153eb5141fb9", size = 77194, upload-time = "2024-08-22T20:42:09.784Z" },
+    { url = "https://files.pythonhosted.org/packages/81/98/1c1a1ba5d74f87cbabbf207d36a98551b082b8b187dc35f7808b43f4cf1a/aiobotocore-2.13.3-py3-none-any.whl", hash = "sha256:1272f765fd9414e1a68f8add71978367db94e17e36c3bf629cf1153eb5141fb9", size = 77194 },
 ]
 
 [package.optional-dependencies]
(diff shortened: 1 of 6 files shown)


Deepak-Kesavan:
## Review pass — findings addressed

### Correction to an earlier claim in this PR

I wrote that "the characterisation pin compares against whatever is installed, so it cannot paper over a behaviour change." That is backwards. `_original_impl` calls the **same installed** `json_repair`, so a version bump moves both sides identically and the pin stays green. The pin protects against edits to `_repair_legacy`, not against dependency drift.

Since `json-repair` is declared with a lower bound (`>=0.42`) rather than pinned, drift is real — `_reparse_without_wrap` exists precisely because newer releases can mis-tokenise an intact response. Added **absolute-value** tests (Part 6) instead: they assert the exact output customers receive, so a routine `uv lock` refresh that changes extraction fails loudly rather than shipping.

One of those pins documents current behaviour rather than endorsing it: on the no-contract path a fenced response is **shredded** into `["json\n{", 'invoice_number": "INV-001', {...}]`. That is today's production behaviour, and the reason the contract path offers an unwrapped candidate.

### Fixed

| Severity | Finding |
|---|---|
| **Critical** | `_skeleton` echoed dict keys into an always-on WARNING trace. Keys are *not* schema-only on malformed responses — a prompt returning a map keyed by extracted values (party→role, line-item→amount) put PII in production logs with no flag set. Only contract-declared keys are echoed now; everything else becomes its length. |
| **Major** | `repair_json_with_best_structure(b'...')` raised `TypeError` where it previously returned a dict — `json.loads` accepts bytes. A regex scan in the diagnostics narrowed the accepted input, which is exactly the class of change this PR promises not to make. |
| **Major** | The alignment check reported `MISMATCH` on **every** well-formed highlight+word-confidence response — it stripped `%%%` for the comment count but took leaves from a parse of the unstripped text. A 400-field payload reported `comments=400 leaves=401`. A false positive on the precise class the check exists for. |
| **Major** | The reshape guard matched a bare `//`, so any document value containing one (`"N/A // pending"`, a path) was treated as annotated, the salvage was skipped, and the AttributeError this PR fixes still occurred — silently. Now requires the hex payload `// 0x`, and a *paired* word-confidence marker. |
| Minor | `%%%` was hardcoded while the consumer reads `WORD_CONFIDENCE_MARKER`; an operator override broke the guard and the check with nothing to notice. |
| Minor | The webhook posted synchronously on the calling thread — up to its timeout per prompt against a slow collector, on every repair, contradicting "cannot slow an extraction". Now a daemon thread. |
| Minor | `_log_cleansing_chain` was unguarded; any future addition inside it could fail an extraction. Wrapped. |
| Minor | Test hygiene: a real DNS call to `.invalid`, a thread-local `LogContext` leaking into later tests, and an assertion that accepted either outcome. |

`DEBUG_LOG_RAW_LLM_MAX_CHARS=1` also emitted the whole payload (`tail = 1//2 = 0`, and `text[-0:]` is the entire string).

**226 tests pass. Differential fuzz: 2102 inputs including bytes, 0 differences on the no-contract path.**

### Deliberately not fixed

The consumer-side `//`-in-a-value offset (`re.findall(r"//(.*)")` over raw text) is left as a known issue by agreement — it predates this PR. The new alignment check reports it rather than fixing it.

### No activated call site in this repo — by design

Worth stating plainly for reviewers: **nothing in this repo passes a contract**, so the reported 500 is not fixed by this PR alone. The only tracked caller is `answer_prompt.py:438` (single-prompt, where a list is valid and no contract is wanted). The single-pass path that produces the failure lives in the cloud plugin and uses a separate copy of this module. That wiring is a follow-up; this PR is the mechanism, the regression safety net, and the diagnostics.

PR #2251, opened 2026-08-19
UN-4017 [FIX] Reject non-mapping outputs at the prompt-output internal API

Deepak-Kesavan:
## What

- Reject a non-mapping `outputs` **or** `metadata` at the `prompt_output` internal API with a 400, instead of letting them raise `AttributeError` inside `OutputManagerHelper` and surface as a 500.
- Add the same guard to the in-backend `_handle_response` path.

## Why

- Production throws this on Prompt Studio single-pass runs (`unstract-prod` / `unstract-production`, caller `unstract-worker-ide-callback-v2`):

  ```
  internal_views.py:84            prompt_output
    → output_manager_helper.py:179  handle_prompt_output_update
        output = outputs.get(prompt.prompt_key)
  AttributeError: 'list' object has no attribute 'get'
  ```
- Reachable from ordinary traffic: single-pass extraction passes the LLM's parsed JSON straight through as the outputs map, and the repair returns a list whenever the model adds commentary or answers in prose. Hence intermittent and seemingly model-dependent.
- `metadata` has the identical exposure — indexed five times at `output_manager_helper.py:135-139`, unconditionally and **before** the `if not prompts` early exit — so valid `outputs` with `"metadata": []` produced the same 500.

## How

- One check covering both fields, after `_parse_json_body` and the `prompt_ids`/`document_id` validation, before any ORM work.
- The message names the actual type rather than calling it malformed — a JSON array is valid JSON.
- The single-pass prompt advice is emitted **only** for `outputs`. `metadata` is executor-assembled, so telling a user to rephrase a prompt would be a dead end for a defect that is ours.
- `_handle_response` gets the same guard, raising the local `AnswerFetchError` with 422.

- No. Every shape now rejected previously produced a 500 — a list, str, int or None has no `.get` — so nothing that used to succeed now fails.
- A `dict` passes through unchanged; absent `outputs`/`metadata` still default to `{}` and are accepted. Both covered.
- `dict` is correct here, not `Mapping`: both values come from `json.loads` with no `object_pairs_hook`, and `OrderedDict`/`defaultdict` are `dict` subclasses anyway. Plain Django view, so no DRF `QueryDict`/`ReturnDict` involved.
- Callers see 400 instead of 500. The 400 also stops a pointless retry storm — the current failure is retried 4x with backoff, all doomed, since the payload is identical each time.

## Database Migrations

- None.

## Env Config

- None.

## Relevant Docs

- None.

## Related Issues or PRs

- UN-4017
- Source-side guard, the other half of this fix: Zipstack/unstract-cloud#1739

## Dependencies Versions

- None. This PR touches no dependency or lockfile.

## Notes on Testing

- 11 cases: array rejected with a reason, non-mapping `metadata` rejected, other non-mappings rejected, the helper never reached for invalid input, dict outputs still reach it, absent `outputs` still accepted, the prompt advice absent for `metadata`, and four for the in-backend path including the negative case that the advice is omitted for single-prompt.
- Verified to **fail** with the guards reverted rather than merely to pass.

## Notes for the reviewer

- **The in-backend guard hardens a path with no live callers today.** `_handle_response` is reached only from `_execute_single_prompt` / `_execute_prompts_in_single_pass`, and both only from `PromptStudioHelper.prompt_responder`, which has **zero callers** in either repo — the live route is `views.single_pass_extraction` → `dispatch_with_callback` → executor → `ide_prompt_complete` → this internal API. Kept as defence-in-depth; deleting the chain instead is a reasonable alternative.
- **`prompt_ids` is deliberately unchecked.** A non-list makes `filter()` raise `ValidationError`, which the existing `except` turns into a 500 — but the producer always sends a list and the `if not prompt_ids` check covers the realistic cases.
- **Known, out of scope:** OSS declares `json_repair` nowhere, so `answer_prompt.py:438-442` persists `{}` for any answer that is not strictly valid JSON. Declaring it here forced a ~2100-line root `uv.lock` downgrade from the uv-lock automation (pinned to `uv 0.6.14`, older than whatever wrote `revision = 3`) that could not be reverted while the declaration was present. Tracked for a follow-up, along with pinning that workflow's `uv`.

## Checklist


backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -1875,10 +1875,33 @@ def _handle_response(
...
 
+        outputs = response["output"]
+        # Same guard as the internal API (UN-4017). This is the in-backend
+        # execution path — it dispatches the identical single_pass_extraction
+        # executor, so it can receive the identical bad shape. Without this,
+        # handle_prompt_output_update does outputs.get(prompt.prompt_key) on a
+        # list and raises AttributeError, which surfaces as a bare 500.
+        if not isinstance(outputs, dict):
+            detail = (
+                "LLM returned a JSON array where a single JSON object keyed by "
```
praveen-formido: **The message asserts "JSON array" for every non-dict, contradicting the type it prints — and `metadata` is left unguarded on this path.**

1. `detail` hardcodes "LLM returned a JSON array" but interpolates the real type. For `response["output"] = None` the user reads *"LLM returned a JSON array where a single JSON object keyed by field name is expected (got NoneType)"* — self-contradictory, and it sends them looking for an array that was never there. The cloud half of this ticket makes exactly this argument and pins it with a test (`test_message_names_the_actual_type_not_a_guessed_one`); this half does the opposite. Suggest dropping the shape claim, as in the cloud message.

2. Only `outputs` is guarded. Three lines below, `metadata=response["metadata"]` is still passed unchecked, and `handle_prompt_output_update` indexes it five times at `output_manager_helper.py:135-139` — unconditionally and *before* the `if not prompts` early exit. That is the same exposure the PR description identifies as "missed on the first pass" and fixes in `internal_views.py`; leaving it here means the in-backend path can still 500 on a non-mapping `metadata` even after this change.

```suggestion
                "LLM response could not be used as a single JSON object keyed by "
```

workers/pyproject.toml
```diff
@@ -65,7 +65,13 @@ test = [
...
     "pytest-timeout>=2.3.1",
     "factory-boy>=3.3.0",
-    "responses>=0.23.0"
+    "responses>=0.23.0",
+    # json_repair_helper imports this; the image gets it transitively from
+    # cloud plugin deps, so the OSS test env is the only place it is missing.
+    # Without it the helper silently takes `except ImportError: return
+    # json_str`, and any test asserting real repair behaviour passes while
+    # pinning the fallback (UN-4017).
+    "json-repair>=0.25.0",
```
praveen-formido: **`json-repair` belongs in the runtime dependencies, not only in `test`.**

The comment says "the OSS test env is the only place it is missing" — that isn't quite right. `worker-unified.Dockerfile:67` builds with `uv sync --group deploy --locked`, which does **not** install the `test` group, and `json-repair` is absent from `workers/uv.lock`'s runtime resolution before this PR. So the OSS worker image ships without it, and `json_repair_helper.repair_json_with_best_structure` takes `except ImportError: return json_str` at runtime, not just under pytest.

That has a live consequence in OSS today, independent of single-pass: `executor/executors/answer_prompt.py:438-442`

```python
parsed_data = repair_json_with_best_structure(answer)
if not isinstance(parsed_data, (dict, list)):
    logger.error("Error parsing response to JSON")
    structured_output[prompt_key] = {}
    return
```

Any `json`/`table`/`record` prompt whose answer isn't *strictly* valid JSON (trailing prose, an unterminated string, a stray fence) silently persists `{}` instead of the repaired object — the exact class of failure this ticket is about, just failing quietly instead of loudly. Cloud only escapes it because `line_item_extractor` / `smart_table_extractor` / `challenge` pull `json-repair` in transitively.

Since the PR is already touching this file for this dependency, promoting it to `[project].dependencies` fixes the runtime gap and gives the test group the package for free.

uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 1
 requires-python = "==3.12.*"
```
praveen-formido, line 3: **Unintended lockfile downgrade — this file was regenerated with an older `uv`.**

`revision = 3` -> `revision = 1`, and all ~1060 `upload-time = "..."` fields are stripped from every `sdist`/`wheel` entry. That accounts for the entire +1071/-1070 here; no dependency, version or hash actually changes, and nothing in this PR touches a root or `backend/` dependency (the `json-repair` addition is in `workers/pyproject.toml` and correctly lands in `workers/uv.lock`).

Consequences: it silently rolls the lock format back for everyone, `docker/dockerfiles/backend.Dockerfile:53,67` run `uv sync --locked` against it, and a ~2100-line no-op diff will conflict with any other in-flight lock change.

Please `git checkout main -- uv.lock` and re-run with the pinned `uv` version if a root re-lock is actually needed.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
Deepak-Kesavan: Both fixed in `ac6bf565f` + follow-up.

1. **Message.** Dropped the shape claim — it now reads *"LLM response could not be used as the {field} map — a single JSON object keyed by field name is expected (got {type})"*. You're right that it contradicted itself for `NoneType`, and right that the cloud half argued the opposite; the two now agree.

2. **`metadata`.** Guarded on this path too. Both fields go through one loop, so `response["metadata"]` is checked before the call rather than three lines after it.

Test updated to assert `"outputs map"` rather than `"JSON array"`, and the single-pass advice is still emitted only when `is_single_pass` — the negative case is pinned.

Worth flagging since you may want it removed instead: this whole path looks unreachable. `_handle_response` is only reached from `_execute_single_prompt` / `_execute_prompts_in_single_pass`, and both only from `PromptStudioHelper.prompt_responder`, which has no callers in either repo. I kept the guard as defence-in-depth and noted it in the description, but deleting the chain is a defensible alternative.

harini-venkataraman (approved):

workers/pyproject.toml
Deepak-Kesavan: Your analysis is right and I've verified it — but I've removed the declaration from this PR rather than promoting it, for a reason that only became clear after two attempts.

**Your finding stands.** `worker-unified.Dockerfile` builds with `uv sync --group deploy --locked`, the `test` group isn't installed, and `answer_prompt.py:438` therefore persists `{}` for any `json`/`table`/`record` prompt whose answer isn't strictly valid JSON. That's a live OSS bug independent of single-pass.

**Why not here.** `.github/workflows/uv-lock-automation.yaml` triggers on any `**/pyproject.toml` change and installs `uv 0.6.14`, which is older than whatever wrote the root `uv.lock` at `revision = 3`. It regenerates every lock and pushes. So touching a pyproject in this PR makes the root-lock downgrade you flagged in the next comment **unavoidable** — I reverted it twice and the automation re-added it both times, on `synchronize`. The two comments are in direct tension: I can satisfy one or the other, not both, in this PR.

**Where it went.** #2250 already declares `json-repair` and is the PR about this helper, so the dependency question belongs there — including your point that it should be `[project].dependencies` rather than a test group. I'll carry it over.

This PR is now three files and no lock churn.

Separately, that automation downgrades the root lockfile on **every** PR touching a pyproject, not just this one. Probably worth its own fix — pinning the workflow's `uv` to the version that wrote `revision = 3`.

uv.lock
Deepak-Kesavan: Gone — the PR is now three files, no `uv.lock` diff at all.

Your diagnosis was exactly right: `revision = 3` -> `1` with every `upload-time` stripped, no dependency or hash actually changing.

`git checkout main -- uv.lock` wasn't enough on its own, though. `.github/workflows/uv-lock-automation.yaml` runs on `synchronize` for any `**/pyproject.toml` change and installs `uv 0.6.14` — older than whatever wrote `revision = 3` — so it regenerated and re-pushed the downgrade after each of my two reverts. The only way to make it stay reverted was to stop touching a pyproject in this PR, which is why the `json-repair` declaration moved to #2250 (see the thread above).

I also merged `main` in to clear the resulting conflict, so the branch is current.

The automation will do this to **any** PR that touches a pyproject, so it's worth fixing at the source — pinning that workflow's `uv` to the version that produced `revision = 3`. Happy to open that separately if useful.

Deepak-Kesavan:
## Review round addressed — ready for another look

@pk-zipstack all three points are handled; replies are in-thread.

| Your finding | Outcome |
|---|---|
| Message asserts "JSON array" while printing the real type | Shape claim dropped — names the type, matching the cloud half |
| `metadata` unguarded on the in-backend path | Both fields checked in one loop before the call |
| Root `uv.lock` downgraded | **Gone** — PR is now 3 files, no lock diff |
| `json-repair` belongs in runtime deps | Agreed, but **moved to #2250** — see below |

**Why the dependency moved rather than getting promoted.** Your two comments were in direct tension. `.github/workflows/uv-lock-automation.yaml` fires on any `**/pyproject.toml` change and installs `uv 0.6.14` — older than whatever wrote the root lock at `revision = 3` — so it regenerates and re-pushes the downgrade on every `synchronize`. I reverted twice; it came back twice. Touching a pyproject here makes the lock churn unavoidable, so satisfying the revert meant not touching one. #2250 already declares `json-repair` and is the PR about this helper.

Your underlying finding is untouched by that and I've carried it over: the image ships without `json_repair`, so `answer_prompt.py:438` silently persists `{}` for any `json`/`table`/`record` prompt whose answer isn't strictly valid JSON.

**Also:** merged `main` (branch was `DIRTY`, now `MERGEABLE`); all checks green; 10 tests, verified to fail with each guard reverted.

**Worth its own fix:** that automation downgrades the root lockfile on *every* PR touching a pyproject. Pinning the workflow's `uv` would fix it repo-wide.

backend/prompt_studio/prompt_studio_core_v2/prompt_studio_helper.py
```diff
@@ -1875,14 +1875,48 @@ def _handle_response(
...
+                continue
+            # Name the type rather than asserting a shape: this fires for
+            # NoneType, int and bool too, and "LLM returned a JSON array
+            # (got NoneType)" contradicts itself.
+            detail = (
+                f"LLM response could not be used as the {field_name} map — a "
+                "single JSON object keyed by field name is expected (got "
+                f"{type(value).__name__})."
+            )
+            if is_single_pass:
```
praveen-formido: **Generalising the check into a loop made the prompt advice fire for `metadata` too, where it is wrong — same class of misdirection this rewrite just removed from the shape claim.**

`metadata` is not LLM output. It is assembled by the executor (`ExecutionResult.data[PSKeys.METADATA]` — `run_id`, `file_name`, `context`, plus highlight/confidence blocks), or by `PayloadModifier.execute_agentic_table`, which hand-builds `{"context": None, "challenge_data": None, ...}`. A non-dict there is an executor or serialization defect, never a consequence of how a prompt was worded.

So for `field_name == "metadata"` the user gets:

> LLM response could not be used as the metadata map … **Rephrase that prompt** to describe the value of its own field, or run these prompts with single-pass extraction turned off.

Both halves are false — the "LLM response" framing and the remediation. Rewording their prompts cannot change `metadata`, so this sends them down a dead end for a bug that is ours. The neutral wording in `internal_views.py` (`metadata must be a JSON object, got …`) gets this right; this path regressed relative to it when the two fields were merged into one loop.

Gating the advice on the field keeps the loop:

```suggestion
            if is_single_pass and field_name == "outputs":
```

Deepak-Kesavan: You're right, and it's a regression I introduced while fixing your earlier comment — the same misdirection I'd just removed from the shape claim, reintroduced one field over.

Fixed: the advice is gated on `field_name == "outputs"`.

Your reasoning is the part I should have applied myself. `metadata` is executor-assembled — `run_id`, `file_name`, `context`, plus the highlight/confidence blocks — so a non-dict there is our defect and no rewording of a prompt can affect it.

Added a test asserting the advice is **absent** for metadata. Worth noting why it slipped: I had a test pinning that the advice is omitted for single-prompt, so I'd covered the `is_single_pass` dimension — but not the field dimension, which only came into existence when I merged the two checks into a loop. Generalising the check without re-examining whether the message still fit both fields is exactly the gap.

Deepak-Kesavan:
## Correction: disregard earlier references to the other parser PR

That PR is stale and is not being merged. **This PR and Zipstack/unstract-cloud#1739 are self-contained — neither depends on anything else.** Description updated; please ignore the earlier thread where I said the `json-repair` question had moved there.

Where that leaves your finding, @pk-zipstack: it stands and is **not fixed here**. OSS declares `json_repair` nowhere, so `answer_prompt.py:438-442` persists `{}` for any answer that is not strictly valid JSON — a live gap independent of single-pass.

I tried declaring it here twice and reverted both times. GitHub matches the uv-lock automation's `paths` filter against the **whole PR diff**, not the individual push, so any PR containing a pyproject change re-fires it on every sync — and that workflow installs `uv 0.6.14`, older than whatever wrote the root `uv.lock` at `revision = 3`. It rewrote ~2100 lines with no dependency, version or hash changing. The revert could not stick while the declaration was present, so keeping this PR clean meant dropping it.

Tracked for a follow-up, together with pinning that workflow's `uv` — which would fix the churn for every PR that touches a pyproject, not just this one.

This PR is now three files, no dependency or lockfile changes.

praveen-formido (approved):

PR #2255, opened 2026-08-25
UN-3973 [DEV] Cut dashboard cron DB time by deriving monthly metrics from the daily tier

kirtimanmishrazipstack:
## What

Monthly dashboard totals are now added up from the daily figures we already store, instead of
re-reading the raw processing history. Each run looks at the last **2 days** instead of up to two
months, and a **once-a-day pass** looks back 7 days to repair anything the cron missed while down.

## Why

Every 15 minutes, for each of 38 organisations, the cron re-read up to two months of processing
history — just so it could total the month. Monthly figures do not need that freshness, and the
daily totals are already stored.

Worth about **460 s of database time per 6 h**. It will not move the two most expensive queries;
those are #2264's.

## How

- Constants: source window **2 days**, reconciliation **7 days**, active-org lookback **7 days**.
  The reconciliation pass is the same task at a wider bound — a second schedule row, no new code path.
- `_aggregate_single_metric` / `_aggregate_llm_combined` drop `monthly_start` / `monthly_agg`; the
  DAY query binds `daily_start` and the redundant Python guard goes.
- `_bulk_upsert_monthly` → `_rollup_monthly_from_daily()` — one ORM aggregate over
  `event_metrics_daily` for all orgs, so the monthly tier stops scaling with tenant count.
  `metric_type` is aggregated (`Min`), not grouped, or a mid-month type change breaks the upsert.
- The active-org prefilter is decoupled from `daily_start` and pinned at 7 days.

Monthly totals now inherit any holes in the stored daily figures. The once-a-day pass repairs the
last 7 days; older holes need the backfill command below, which is why it runs once at deploy.

Deleting old source data no longer corrects totals beyond that same window.

Nothing already counted changes, and a slow-finishing file cannot land in the wrong day.

## Database Migrations

`dashboard_metrics/0005_add_reconciliation_task.py` — adds the once-a-day 04:00 UTC schedule row.
No table changes. Safe to re-run and removable on rollback.

**Run once at deploy, before the first monthly rollup:** `manage.py backfill_metrics --days 60
--skip-hourly --skip-monthly`. Monthly is built from the daily figures now, so those have to be
complete for this month and last month first.

## Migration Order

The three UN-3883 PRs stack on the same integration branch. **Merge in this order.**

| Order | PR | Migration | Depends on | This PR? |
| --- | --- | --- | --- | --- |
| 1 | #2255 · UN-3973 | `dashboard_metrics/0005_add_reconciliation_task` | `0004_pg_periodic_tasks` (UN-3445, already on `main`) | **Yes — this PR** |
| 2 | #2264 · UN-3972 | `file_execution/0007_wfe_status_created_idx` | `file_execution/0006_…` | No — merge after this PR |
| 3 | #2265 · UN-3974 | `dashboard_metrics/0006_split_aggregation_schedule` | `0005_add_reconciliation_task` (#2255) | No — merge after this PR |

#2264's migration is in a different app and has no interaction with the other two. The one
hard dependency is **#2265 on #2255** — merged out of order, #2265 fails at graph build:

```
NodeNotFoundError: Migration dashboard_metrics.0006_split_aggregation_schedule
dependencies reference nonexistent parent node ('dashboard_metrics', '0005_add_reconciliation_task')
```

Its tests are unaffected — the backend suite runs with `--no-migrations`.

Verified on a throwaway Postgres: `0004` → `0005` → `0006` applies from an empty database,
reverses, and re-applies, with `makemigrations --check` clean at each step.
## Env Config

None.

## Relevant Docs

UN-3883 analysis §6.2, §6.3, §7 step 1, §8.

## Related Issues or PRs

Parent UN-3883. #2264 (UN-3972), #2265 (UN-3974) — see **Migration Order** above.

## Dependencies Versions

None.

## Notes on Testing

38 tests. Detail in Jira comment 45429.

| # | Acceptance criterion | Verdict |
| --- | --- | --- |
| 1 | Monthly totals derived from `event_metrics_daily`, not source tables | **Met** |
| 2 | Per-run daily source window is 2 days | **Met** |
| 3 | A once-daily 7-day reconciliation pass exists and is scheduled | **Met** |
| 4 | Daily and monthly figures match pre-change values, incl. across a month boundary | **Met** |
| 5 | Tests cover the month-boundary case and the reconciliation pass | **Met** |

## Screenshots

n/a — backend only.

## Checklist


backend/dashboard_metrics/tasks.py
```diff
@@ -165,32 +182,47 @@ def _bulk_upsert_daily(aggregations: dict) -> int:
...
+            tag=row["tag"],
+            metric_type=row["mtype"],
+            metric_value=row["value"],
+            metric_count=row["count"],
         )
+        for row in rows
+    ]
 
     if not objects:
         return 0
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Valid — fixed in 3ea08b721.

Worth flagging that this predates the PR: the old `_bulk_upsert_monthly` was equally upsert-only, so an obsolete monthly key survived a rebuild there too. But this PR is what makes monthly a pure derivation of the daily tier, so making that derivation faithful belongs here.

`_rollup_monthly_from_daily` now deletes rows in the rebuilt window that the rollup no longer produces, in the same transaction as the upsert. The `if not objects: return 0` guard is deliberately kept ahead of it — an empty daily tier means the source of truth is gone, not that every month is genuinely zero, so a wiped daily tier cannot cascade into deleting monthly history.

Two tests added: one that a month whose daily rows are deleted loses its monthly row, and one that the cleanup does not reach outside the rebuilt window.

backend/dashboard_metrics/tasks.py
```diff
@@ -165,32 +182,47 @@ def _bulk_upsert_daily(aggregations: dict) -> int:
...
-                metric_count=agg["count"],
-            )
+    rows = (
+        EventMetricsDaily._base_manager.filter(date__gte=month_start)
+        .annotate(month=TruncMonth("date"))
+        .values("organization_id", "month", "metric_name", "project", "tag")
+        .annotate(
+            value=Sum("metric_value"),
+            count=Sum("metric_count"),
+            mtype=Min("metric_type"),
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Confirmed as real — this is the one genuine regression in the PR. Before it, monthly was re-derived from source tables over a 62-day window every run, so a partial source deletion self-corrected; now monthly sums the daily tier, which keeps the stale bucket. And deletion is a routine path, not hypothetical: `Workflow` cascades to `WorkflowExecution`, which cascades to `WorkflowFileExecution`.

I'm still not fixing it in the cron, for three reasons:

1. A metric whose query raises is caught per-metric and contributes no keys — indistinguishable from "every row for it was deleted". Deleting in-window buckets on that basis turns a transient query failure into permanent data loss, which is strictly worse than staleness.
2. It would only correct the last 2 days, or 7 on the reconciliation pass. Delete a workflow with three months of executions and the dashboard stays wrong for everything older, so the scenario in the finding isn't actually closed by the fix.
3. It adds a scan and a delete to every run of a task whose entire purpose in this PR is cutting DB time.

The right fix is invalidation at the point of deletion — clearing the affected daily and monthly rows when the workflow goes away — which is a separate change with its own correctness argument, not a wider scan in the cron. The narrowed self-correction window is now called out in the PR description so it isn't silently inherited.

backend/dashboard_metrics/tasks.py
```diff
@@ -165,42 +183,102 @@ def _bulk_upsert_daily(aggregations: dict) -> int:
...
+            metric_value=row["value"],
+            metric_count=row["count"],
         )
+        for row in rows
+    ]
 
+    # An empty daily tier means the source of truth is gone, not that every
+    # month is genuinely zero, so leave the existing rows alone.
     if not objects:
         return 0
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Deliberate, and I'd like to keep it — note this contradicts your own review of the same function 19 seconds earlier, which called out the guard's placement as correct precisely because "a wiped daily tier cannot cascade into deleting monthly history."

The empty case can only arise one way. A monthly row for a window is created *only* from daily rows in that window, so an org that genuinely had no activity has neither daily nor monthly rows — there is nothing to orphan. Monthly rows existing while every daily row in the window is gone therefore always means the daily rows were deleted after the fact. That is exactly the state where the daily tier is not a trustworthy authority, and deleting monthly on its say-so destroys the last good copy of the data.

Worth adding that the guard only fires on a *total* wipe of the window across all organizations. A single surviving daily row anywhere makes `objects` non-empty and re-enables orphan cleanup for every org, including ones whose own daily rows are gone. So the "indefinitely stale" window is narrower than the finding implies.

On the two round-trips in `_delete_orphan_monthly`: considered and declined. The orphan test is a five-column composite key, which Django cannot express as a correlated `NOT IN (SELECT ...)` without dropping to raw SQL — and this repo's architecture guidelines rule that out for anything avoidable. At a two-month window the candidate set is small enough that the extra round-trip is not worth the raw statement.

Also taking your point on the future invalidation change: a partial deletion (one execution removed without the workflow) overstates the daily bucket without touching the workflow row, so a `post_delete` on `Workflow` alone would not be sufficient. That will need to hang off the execution cascade too.

kirtimanmishrazipstack:
@greptile-apps please review

backend/dashboard_metrics/tasks.py
```diff
@@ -165,42 +183,102 @@ def _bulk_upsert_daily(aggregations: dict) -> int:
...
+            row["metric_name"],
+            row["project"],
+            row["tag"],
+        )
+        not in fresh_keys
+    ]
+    if not stale_pks:
+        return 0
+
+    deleted, _ = EventMetricsMonthly._base_manager.filter(pk__in=stale_pks).delete()
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps False positive. The mechanism is real but the conclusion isn't — the deletion is the designed invariant, and it isn't a regression.

I reproduced your scenario directly, so the first half holds: a monthly key inside the rebuild window with no daily rows behind it does get deleted (`SKIP_DAILY_NO_BACKING -> None`). Three things stop that from being a defect.

**1. `--skip-daily` monthly rows were already transient before this PR.** On `main`, `_run_aggregation` recomputed monthly from the source tables over the same `monthly_start` window — current + previous month — and upserted over whatever `backfill_metrics` had written. A monthly-only backfill of the last two months was overwritten within 15 minutes then too. This PR changes *what* overwrites it (daily tier instead of source tables), not *whether* it survives.

**2. In the normal case the row is recomputed, not deleted.** `backfill_metrics` derives `monthly_agg` from the same DAY-granularity results it derives `daily_agg` from (`_ingest_daily_results`, `backfill_metrics.py:303-338`), so the monthly total it writes is by construction the sum of the daily rows for that key. If those daily rows are in the tier — which is the only reason to reach for `--skip-daily` — the rollup lands on the same number:

```
SKIP_DAILY_WITH_BACKING -> 999 → 42   (recomputed from daily, no delete)
```

Deletion fires only when monthly asserts a total the daily tier cannot support. Post-PR that is stale by definition.

**3. "Backfilled history disappears" overstates the blast radius.** `_delete_orphan_monthly` filters `month__gte=month_start`, so only the current and previous month are in scope. Anything `backfill_metrics --days 30/90` wrote before that is untouched:

```
OLD_MONTH -> 777.0   (2026-05, survives)
```

Daily retention is 365 days (`DASHBOARD_DAILY_METRICS_RETENTION_DAYS`) against a 2-month monthly window, so cleanup can't strip the daily rows backing a month inside the rebuild window either.

What is genuinely true is narrower than the finding: post-PR the way to repair the last two months of monthly is to backfill **daily** and let the rollup derive monthly — `--skip-daily` is the wrong tool for that window now. That's a help-text clarification on the flag, not a code defect. Happy to fold it in if you want it here.

kirtimanmishrazipstack:
@greptile-apps please review again

kirtimanmishrazipstack:
@greptile-apps On the summary verdict — the mechanism is right, "not safe to merge" isn't. This is the trade the ticket asks for, and it was in the PR description before the review.

**It's documented and accepted.** Under "Can this PR break any existing features", item 1:

> Monthly now inherits gaps in `event_metrics_daily`. […] the reconciliation pass heals 7 days back. Downtime longer than 7 days needs the `backfill_metrics` management command. This is the trade the ticket accepts by design.

Monthly stopped self-healing over 62 days deliberately — that 62-day re-derivation from source tables *is* the DB load UN-3973 exists to remove. Widening the rollup's source window to make orphan cleanup safe against arbitrarily old gaps would revert the PR.

**You did surface one thing the description understated, and I've fixed it.** "Inherits gaps" reads as undercounting; the deletion case is worse than that:

- *some* daily rows for a key survive the gap → monthly recomputed **low**
- *every* daily row for that key missing → key never enters `fresh_keys` → monthly row **deleted outright**

Confirmed against the test DB (`SKIP_DAILY_NO_BACKING -> None`). Both are bounded to the current and previous month by `month__gte=month_start`; a month outside the window is untouched (`OLD_MONTH -> 777.0`). Item 1 now states the deletion case explicitly rather than folding it into "gaps".

**The one-time exposure is real, and it's a deploy step rather than a code change.** Gaps already sitting in the daily tier would propagate into monthly on the first rollup after this ships. Added a `## Deploy Steps` section:

```
manage.py backfill_metrics --days 60 --skip-hourly --skip-monthly
```

That rebuilds daily from source across the rollup window; the scheduled rollup then derives monthly from a complete tier. `--skip-monthly` is deliberate — repair daily and let the rollup derive monthly, rather than writing monthly from source here. Needed once; the 04:00 reconciliation pass carries it forward after that.

Ongoing gaps beyond 7 days remain the accepted trade, with `backfill_metrics` as the documented remedy.

backend/dashboard_metrics/tasks.py
```diff
@@ -165,42 +183,102 @@ def _bulk_upsert_daily(aggregations: dict) -> int:
...
+            month__gte=month_start
+        ).values("pk", "organization_id", "month", "metric_name", "project", "tag")
+        if (
+            row["organization_id"],
+            row["month"],
+            row["metric_name"],
+            row["project"],
+            row["tag"],
+        )
+        not in fresh_keys
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps This is the summary verdict filed inline, and it was raised at 14:15 — before the PR description update (14:20) and my reply (14:19) landed. Answered in both; short version here.

Valid as a mechanism, and documented rather than coded around:

- **Description, item 1** now splits the two shapes instead of calling both "gaps": where *some* daily rows behind a key survive, monthly is recomputed **low**; where *every* one is missing, the key never enters `fresh_keys` and the row is **deleted**. Both bounded to the current and previous month by `month__gte=month_start`, so months outside the rebuild window are untouched.
- **New `## Deploy Steps` section** closes exactly the exposure you name — the gaps sitting in the daily tier *today*, before the derivation invariant takes effect:

```
manage.py backfill_metrics --days 60 --skip-hourly --skip-monthly
```

That rebuilds the daily tier across the rollup window; the scheduled rollup then derives monthly from a complete tier. Once only, after the migration and before the first rollup.

What stays as-is: ongoing gaps beyond 7 days remain the accepted trade of UN-3973, with `backfill_metrics` as the documented remedy. Making orphan cleanup safe against arbitrarily old gaps means widening the rollup's source window past 7 days — that is the query cost this PR exists to remove, so it would revert the change.

Separately, noted your withdrawal on the `--skip-daily` thread. Agreed on the help-text follow-up; say the word and I'll fold it into this PR, otherwise it rides with UN-3974.

kirtimanmishrazipstack:
@greptile-apps can you review and generate new confidence score?

kirtimanmishrazipstack:
@greptile-apps migration order is revised. Review again.

athul-rs (commented):
@kirtimanmishrazipstack — standardized pre-merge review (16-lens rubric, `unstract` plugin v0.18.1), mode INITIAL, at `5804a129`.

## Verdict: `BLOCK`

**Critical: 1 · High: 6 · Medium: 7 · Low: 7 · Lenses run: 16/16**

The Critical is the orphan-delete in `_rollup_monthly_from_daily`: its only guard is built from a query that is unscoped across all organisations, so it cannot fire in production, and any gap in `event_metrics_daily` either deletes a correct monthly row or overwrites it with an undercount — permanently, on the only unbounded-retention tier.

## Lens checklist

| # | Lens | Result |
|---|---|---|
| 2 | Architectural fit | Clean |
| 3 | Correctness & edge cases | See findings |
| 4 | Security | See findings (unvalidated boundary input) |
| 5 | Data integrity & migrations | See findings |
| 6 | Concurrency | See findings |
| 7 | API & contract compatibility | See findings (Low) |
| 8 | Reliability & resilience | See findings |
| 9 | Performance & cost | See findings (Low) |
| 10 | Observability | See findings |
| 12 | LLM/agent | N/A — no prompts, model calls, tools or evals |
| 14 | Dependencies & build | N/A — no manifest, lockfile, Dockerfile or CI workflow |
| 15 | Code quality | See findings (Low) |

Migration graph verified independently: `0005` → `0004` is correct, `0004` is the current leaf, and no other open PR claims that number.

### Unanchored findings

* **[Medium] [Lens 16] `backend/dashboard_metrics/README.md:301-302`, `:385-390`** still document "Daily: Last 7 days" and a monthly source read that no longer exists. The code block shows `daily_start`/`monthly_start` derivations this PR deleted.
* **[Medium] [Lens 16] `backend/dashboard_metrics/management/commands/backfill_metrics.py:1-11`, `:97-101`** still document the command as populating `EventMetricsMonthly` from source tables. Any monthly row it writes for the current or previous month is now recomputed — or deleted — by the next run. An operator following its own docs sees the repair silently revert.
* **[Low] [Lens 7]** `period.monthly.start` is now an ISO **date** (`2026-09-01`) rather than a datetime, while `hourly` and `daily` still emit full timestamps — the payload is internally inconsistent. No in-repo consumer found.
* **[Low] [Lens 13]** `test_default_window_bounds_the_daily_query` and `test_reconciliation_window_widens_the_daily_query` re-read `timezone.now()` *after* the run and compare truncated values — they flake across midnight UTC, and both would pass with the body of `_run_aggregation` deleted.
* **[Low] [Lens 13]** `test_mixed_metric_type_within_a_month_yields_one_row` never asserts *which* `metric_type` survives, so `Min` → `Max` stays green.
* **[Low] [Lens 15]** `_aggregate_org` mutates a caller-owned `stats` dict, against the repo's CRITICAL immutability rule; `_collect_org_metrics` directly above gets this right by returning a tuple.
* **[Low] [Lens 15]** The migration's literal `7` vs `DASHBOARD_RECONCILE_WINDOW_DAYS` asserted in `test_tasks.py:513` is a divergence trap — changing the constant fails a test whose only apparent fix is editing an applied migration, which never re-runs.
* **[Low]** PR body says "38 tests". `test_tasks.py` goes from 9 to 26 on this branch, i.e. **17 new / 26 total**; the figure only reconciles by counting parametrized cases in `test_pg_periodic_task_declarations.py`, which this PR does not touch.

### Cross-PR

* `#2255` and `#2265` both rewrite `tasks.py` from the same base and **conflict** — `git merge-tree` reports a content conflict. #2265's tier gating is written against `monthly_agg`/`_bulk_upsert_monthly`, which this PR deletes, so whoever merges second re-derives it rather than resolving it. Each PR shows `MERGEABLE` because each merges into the base alone.
* After both land, the reconciliation row carries no `tier`, so it resolves to `AggregationTier.ALL` and redoes the 24 h hourly tier once a day. It probably wants `{"tier": "daily_monthly", "source_window_days": 7}`.

Reviewed with the PR Review Toolkit agents (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer); lenses 5, 6, 8, 11, 12, 14 assessed directly. Verdict is advisory — posted as a comment, not a merge gate.

backend/dashboard_metrics/tasks.py
```diff
@@ -165,42 +179,83 @@ def _bulk_upsert_daily(aggregations: dict) -> int:
...
+    }
+
+    with transaction.atomic():
+        EventMetricsMonthly._base_manager.bulk_create(
+            objects,
+            update_conflicts=True,
+            unique_fields=["organization", "month", "metric_name", "project", "tag"],
+            update_fields=["metric_type", "metric_value", "metric_count"],
+        )
+        _delete_orphan_monthly(month_start, fresh_keys)
```
athul-rs: **[Critical] [Lens 1, 3, 5] — the orphan-delete destroys correct monthly rows, and its only guard cannot fire in production**

`_delete_orphan_monthly` deletes **every** `event_metrics_monthly` row with `month >= month_start` whose 5-tuple is absent from this run's aggregate. Two shapes, both silent:

* **whole-key gap** (org/metric with no daily rows for last month) → a correct row is deleted and cannot be rebuilt, because the daily tier is exactly what was missing;
* **partial gap** (daily present for 20 of 31 days — the realistic post-downtime shape) → the key stays in `fresh_keys`, so the correct total is **overwritten with an undercount**, permanently.

`event_metrics_monthly` is the only unbounded-retention tier (`README.md:302`, `Retention: Forever`), and this runs 96x/day.

Pre-change, monthly was re-read from source each run, so a hole self-healed on the next tick. It no longer does.

The agreed design in UN-3973 comments 44768/45016 was **upsert-only**; the delete is not in that spec and is what converts a recoverable undercount into unrecoverable loss.

**Suggested fix:** drop the deletion — a stale row is strictly less harmful than a deleted one. If it must stay: scope per-month, never touch the closed previous month, log the count, and bound it.

backend/dashboard_metrics/tasks.py
```diff
@@ -165,42 +179,83 @@ def _bulk_upsert_daily(aggregations: dict) -> int:
...
+            project=row["project"],
+            tag=row["tag"],
+            metric_type=row["mtype"],
+            metric_value=row["value"],
+            metric_count=row["count"],
+        )
+        for row in rows
+    ]
 
+    # An empty daily tier means the source is gone, not that every month is
```
athul-rs: **[Critical — same finding as the delete call site] — this guard is global, so it never fires in production**

```python
# An empty daily tier means the source is gone, not that every month is
# zero — leave existing rows alone.
if not objects:
    return 0
```

`objects` is built from `EventMetricsDaily._base_manager.filter(date__gte=month_start)` at line 217 — **unscoped across all organisations**. With 38 orgs it is never empty, so this fires only against a wholly empty table, i.e. in tests. The comment's promise to "leave existing rows alone" therefore never applies in production while `_delete_orphan_monthly` runs unconditionally underneath it.

The rationale ("the source is gone, not that every month is zero") is sound *per month* — the guard just is not.

backend/dashboard_metrics/migrations/0005_add_reconciliation_task.py
```diff
@@ -0,0 +1,61 @@
...
+        defaults={"timezone": "UTC"},
+    )
+
+    periodic_task_model.objects.update_or_create(
+        name=RECONCILE_TASK_NAME,
+        defaults={
+            "task": "dashboard_metrics.aggregate_from_sources",
+            "crontab": schedule_4am,
+            "queue": "dashboard_metric_events",
+            "kwargs": '{"source_window_days": 7}',
```
athul-rs: **[High] [Lens 5, 11] — this kwarg cannot reach the task on the PG scheduler path**

`kwargs='{"source_window_days": 7}'` on task path `dashboard_metrics.aggregate_from_sources`. Tracing it:

1. `mirror_pg_periodic_tasks` mirrors **every** non-pipeline Beat periodic — `_EXCLUDED_TASK_PATHS` covers only the two pipeline paths and `celery.backend_cleanup`, so this row is mirrored.
2. `entrypoint.sh:71` runs `converge_pg_scheduler --periodics` on every backend start, which adopts mirrored rows.
3. `pg_scheduler.py:348` dispatches `kwargs=dict(row.task_kwargs or {})`.
4. `workers/scheduler/dashboard_metrics_tasks.py:108` is `dashboard_metrics_aggregate()` — **no parameters** → `TypeError` every tick. `autoretry_for=(DatabaseError, OperationalError)` does not cover `TypeError`.

Widening the worker signature alone is not enough: `internal_views.py:97` forwards nothing, so the window would silently fall back to the 2-day default.

Net: once `PG_SCHEDULER_ADOPT_PERIODICS` is on, the reconciliation pass does not work — and the 2-day window's entire safety argument rests on it.

**Also:** `tests/test_pg_periodic_task_declarations.py:20` pins `_BEAT_MIGRATION` to `0002`, so the Beat/PG drift guard never looks at this migration and stays green with a fourth schedule on one side only.

backend/dashboard_metrics/migrations/0005_add_reconciliation_task.py
```diff
@@ -0,0 +1,61 @@
...
+
+def create_reconciliation_task(apps, schema_editor):
+    """Create the once-daily reconciliation periodic task."""
+    crontab_model = apps.get_model("django_celery_beat", "CrontabSchedule")
+    periodic_task_model = apps.get_model("django_celery_beat", "PeriodicTask")
+
+    # 4:00 AM UTC — clear of the 2:00 and 3:00 cleanup tasks
+    schedule_4am, _ = crontab_model.objects.get_or_create(
+        minute="0",
+        hour="4",
```
athul-rs, line 47: **[High] [Lens 6, 8] — the reconciliation pass silently no-ops for the day whenever it loses the lock**

This row reuses the same task name, so it contends for the single `AGGREGATION_LOCK_KEY`. The 15-minute row is an `IntervalSchedule(every=15, period="minutes")` (`0002_setup_periodic_tasks.py:20-21`), which **drifts** against this fixed 04:00 crontab.

On collision the run returns `{"success": True, "skipped": True, "reason": "lock_held"}` and is never retried. At roughly 138 s of work per 15-minute slot that is a materially recurring loss of the only repair mechanism for the narrowed window — and it is reported as success.

**Suggested fix:** give the reconciliation its own lock key, or retry on `lock_held`.

backend/dashboard_metrics/tasks.py
```diff
@@ -374,82 +406,175 @@ def _aggregate_llm_combined(
...
+    stats: dict[str, Any],
+    hourly_start: datetime,
+    daily_start: datetime,
+    monthly_start: date,
+    end_date: datetime,
+    skipped_reason: str | None = None,
+) -> dict[str, Any]:
+    """Shape the task's return value from the accumulated stats."""
+    result = {
+        "success": True,
```
athul-rs: **[High] [Lens 8, 10] — a run in which every metric for every org failed still returns `success: True`**

Trace the total-failure path: every metric raises → caught per-metric at 485-487 → aggregates empty → `orgs_processed` **still increments** at 525 → this literal returns `{"success": True, "organizations_processed": 38, "hourly": {"upserted": 0}, ..., "errors": 342}` → HTTP 200.

`workers/scheduler/dashboard_metrics_tasks.py:101` (`_log_if_skipped`) checks `result.get("skipped")`, but this path sets **`skipped_reason`** — a different key — so the guard stays silent. Dashboards freeze while the scheduler reports 96 healthy runs a day, which is precisely what that guard's docstring says it exists to prevent.

**Suggested fix:** `"success": stats["errors"] == 0` (or an explicit `degraded` flag); WARNING on the summary log when `errors > 0`; have the worker guard also read `errors` and `skipped_reason`.

backend/dashboard_metrics/tasks.py
```diff
@@ -460,109 +585,40 @@ def _run_aggregation() -> dict[str, Any]:
...
+            _aggregate_org(org, hourly_start, daily_start, end_date, stats)
         except Exception:
-            logger.exception("Error processing org %s", org_id)
+            logger.exception("Error processing org %s", org.id)
             stats["errors"] += 1
 
+    try:
+        stats["monthly"]["upserted"] = _rollup_monthly_from_daily(monthly_start)
+    except Exception:
+        logger.exception("Error rolling up monthly metrics from %s", monthly_start)
```
athul-rs, line 660: **[High] [Lens 8, 10] — a failed monthly rollup is indistinguishable from "nothing to do", and the blast radius is now all-orgs**

The assignment is inside the `try`, so on exception `stats["monthly"]["upserted"]` keeps its initial `0` from line 582 — the identical value produced by the legitimate empty-rollup path (243) and the `no_active_orgs` early return (595). Three states collapse into one, alongside `success: True`.

Pre-change a monthly failure was per-org inside the loop. It is now a single call covering every org, so the whole tier can stop updating at once.

**Also [High, Lens 5/11]:** the mandated `backfill_metrics --days 60 --skip-hourly --skip-monthly` is enforced only by a line in the PR description. Nothing here checks daily-tier coverage for `month_start`, and the 15-minute row fires unconditionally — so the destructive path above lands within 15 minutes of deploy, while a 60-day backfill is still running. There is no rollback: reverting `0005` does not restore deleted rows.

backend/dashboard_metrics/tasks.py
```diff
@@ -374,82 +406,175 @@ def _aggregate_llm_combined(
...
+        stats["daily"]["upserted"] += _bulk_upsert_daily(daily_agg)
+
+    stats["orgs_processed"] += 1
+
+
+def _active_org_ids(end_date: datetime) -> set:
+    """Organizations with execution activity in the prefilter lookback."""
+    return set(
+        WorkflowExecution.objects.filter(
+            created_at__gte=end_date - timedelta(days=DASHBOARD_ACTIVE_ORG_LOOKBACK_DAYS),
```
athul-rs: **[Medium] [Lens 3, 16] — the reconciliation pass has a blind spot at the far edge of its own window**

This cutoff is **untruncated** (`end_date - timedelta(days=7)`), while `daily_start` at line 574 is `_truncate_to_day(...)`. For the reconciliation run (`source_window_days = 7 = DASHBOARD_ACTIVE_ORG_LOOKBACK_DAYS`) the daily window therefore starts up to 24 h — 4 h for the scheduled 04:00 run — *before* the prefilter cutoff.

An org whose only `WorkflowExecution` activity falls in that sliver is inside the window the pass is meant to repair but is excluded from the org list, so it is never repaired.

This is a regression: on `main` the prefilter used `created_at__gte=daily_start`, the same truncated value, so the two could not diverge.

**Suggested fix:** `min(daily_start, end_date - timedelta(days=DASHBOARD_ACTIVE_ORG_LOOKBACK_DAYS))`.

backend/dashboard_metrics/tasks.py
```diff
@@ -33,6 +36,17 @@
...
+# Daily-tier source lookback, sized against the worst observed
+# created_at -> terminal-status lag.
+DASHBOARD_SOURCE_WINDOW_DAYS = 2
+
+# Wider lookback for the once-daily reconciliation pass.
+DASHBOARD_RECONCILE_WINDOW_DAYS = 7
+
+# Wider than the source window: metrics keyed on another column
+# (e.g. approved_at) can land for an org whose executions are older.
+DASHBOARD_ACTIVE_ORG_LOOKBACK_DAYS = 7
```
athul-rs, line 56: **[Medium] [Lens 16] — this comment asserts an invariant the same PR breaks**

"Wider than the source window" holds for the 2-day default but is **false for the 7-day reconciliation pass this PR schedules**, where the two constants are equal and the prefilter is additionally untruncated while `daily_start` is truncated. See the finding on `_active_org_ids` above.

Either make the code match the comment, or say the guarantee holds only for the default window.

backend/dashboard_metrics/tasks.py
```diff
@@ -260,22 +315,18 @@ def _acquire_aggregation_lock() -> bool:
...
 )
-def aggregate_metrics_from_sources() -> dict[str, Any]:
-    """Aggregate metrics from source tables into hourly, daily, and monthly tables.
-
-    This task runs periodically (every 15 minutes) to query metrics from
-    source tables (Usage, PageUsage, WorkflowExecution, etc.) and aggregate
-    them into EventMetricsHourly, EventMetricsDaily, and EventMetricsMonthly
-    tables for fast dashboard queries at different granularities.
+def aggregate_metrics_from_sources(
+    source_window_days: int = DASHBOARD_SOURCE_WINDOW_DAYS,
```
athul-rs, line 309: **[Medium] [Lens 3, 4] — `source_window_days` is unvalidated at a trust boundary**

The value arrives as JSON from a `PeriodicTask.kwargs` row, which is editable in the django-celery-beat admin. Nothing validates it:

* negative → `daily_start` in the future → every daily query matches nothing, no daily rows written, while the monthly rollup **and its delete** still run;
* `0` → the run never refreshes yesterday;
* `365` → restores the multi-month per-org scan this PR exists to eliminate, past `soft_time_limit=600`.

`internal_views.py` already has `_int_arg` for exactly this shape on `retention_days`. Reuse it.

backend/dashboard_metrics/tasks.py
```diff
@@ -165,42 +179,83 @@ def _bulk_upsert_daily(aggregations: dict) -> int:
     return len(objects)
 
 
-def _bulk_upsert_monthly(aggregations: dict) -> int:
-    """Bulk upsert monthly aggregations using INSERT ... ON CONFLICT.
+def _delete_orphan_monthly(month_start: date, fresh_keys: set[tuple]) -> int:
```
athul-rs: **[Medium] [Lens 16, 15] — the function name and docstring conceal that this deletes rows**

`_rollup_monthly_from_daily`'s docstring says only "Sum the daily tier from month_start into monthly, for all orgs at once", and its return feeds `stats["monthly"]["upserted"]`. Nothing at the call site signals that a `DELETE` against `event_metrics_monthly` is issued, so a maintainer could reasonably reuse it to refresh a single month without realising it prunes.

`_delete_orphan_monthly`'s own docstring says it drops rows "that the rollup no longer produces" / whose daily rows are "left" — implying formerly-derived, now-orphaned rows. The code is unconditional: it deletes rows written by the pre-change code, by `backfill_metrics`, or by a manual correction, for keys the daily tier never carried.

**Suggested fix:** rename to `_rebuild_monthly_from_daily`, document the deletion, return both counts, and surface `deleted` in `stats` — it is currently computed at line 205 and discarded at 257, which is what makes the Critical above silent.

backend/dashboard_metrics/tests/test_tasks.py
```diff
@@ -198,3 +215,318 @@ def test_cleanup_no_records_to_delete(self):
...
+        _run_aggregation()
+        _run_aggregation(source_window_days=DASHBOARD_RECONCILE_WINDOW_DAYS)
+
+        # The run worked — it just cannot reach that far back.
+        assert EventMetricsDaily._base_manager.filter(date=recent_day).exists()
+        assert not EventMetricsDaily._base_manager.filter(date=old_day).exists()
+
+
+class TestReconciliationSchedule(TestCase):
+    """Migration 0004 schedules the once-daily reconciliation pass.
```
athul-rs: **[High] [Lens 13, 1] — AC 4 is marked Met and is not tested; and the PR's headline behaviour has no end-to-end assertion**

Two gaps in this suite:

1. **`EventMetricsMonthly` is never read after a `_run_aggregation` call.** Delete the rollup at `tasks.py:617` and all 26 tests still pass while monthly silently freezes.
2. **No test pins the new monthly numbers against the pre-change ones.** Every monthly assertion feeds hand-written daily rows into `_rollup_monthly_from_daily` and checks the sum equals what the test just wrote — self-consistency, not equivalence. AC 4 ("figures match pre-change values, incl. across a month boundary") is claimed **Met** in the PR body.

Also on this line: the docstring says *"Migration 0004 schedules the once-daily reconciliation pass"* — it is `0005`; `0004_pg_periodic_tasks` is the dependency, and given the PG drift issue above it is exactly the file a reader should not be sent to.

**Suggested fix for (2):** one golden test seeding source rows, computing the oracle the old way (`get_documents_processed` at DAY granularity, bucketed by `_truncate_to_month`), asserting the written monthly value equals it.

backend/dashboard_metrics/tasks.py
kirtimanmishrazipstack: **[High] [Lens 10] — The task's only destructive write has no observability at all**

`_delete_orphan_monthly` computes and returns a deletion count. This call site discards it. The count reaches no log line, no `stats` key, and nothing in the task result — so a run that upserts 12 monthly rows and deletes 8,000 logs identically to one that deletes none.

Given monthly retention is Forever and no scheduled path recomputes it from source any more, the first symptom of monthly data loss is a user noticing a wrong chart. There is nothing an operator could alert on or grep for afterwards.

**Fix:** return both counts, record `stats["monthly"]["deleted"]`, and log at WARNING when non-zero — it should be rare in steady state, which makes it a good alerting signal rather than noise.

backend/dashboard_metrics/migrations/0005_add_reconciliation_task.py
kirtimanmishrazipstack: **[High] [Lens 7] — The reconciliation pass cannot run on the PG transport; its kwarg raises TypeError there**

This row is created with `task: "dashboard_metrics.aggregate_from_sources"` and `kwargs: '{"source_window_days": 7}'`. Two different functions are registered under that one task name:

- Django/Beat: `aggregate_metrics_from_sources(source_window_days=...)` — accepts it
- Worker/PG: `dashboard_metrics_aggregate()` — accepts nothing (`workers/scheduler/dashboard_metrics_tasks.py:108`)

`mirror_pg_periodic_tasks` copies Beat `kwargs` verbatim into `PgPeriodicTask.task_kwargs`, and the PG scheduler dispatches them as `kwargs`. So after adoption this row dispatches `source_window_days=7` into a zero-arg function; at `MAX_ATTEMPTS=1` the message is dropped rather than retried. `AggregateMetricsAPIView.post` has no body forwarding either, though the two `cleanup_*` views next to it plumb `retention_days` through `_int_arg`.

Note #2265 adds a `tier` parameter to the worker task but still not this one, so the gap survives the merged end state.

Latent rather than live — `pgScheduleMirror.enabled: false` and Beat stays on by default — but it lands exactly when PG adoption happens.

**Fix:** add `source_window_days: int | None = None` to `dashboard_metrics_aggregate` and plumb it through `AggregateMetricsAPIView` with the existing `_int_arg`, mirroring the cleanup views.

backend/dashboard_metrics/tasks.py
```diff
@@ -165,42 +179,83 @@ def _bulk_upsert_daily(aggregations: dict) -> int:
...
+            row["tag"],
+        )
+        not in fresh_keys
+    ]
+    if not stale_pks:
+        return 0
 
-    Args:
-        aggregations: Dict keyed by (org_id, month_str, metric_name, project, tag)
+    deleted, _ = EventMetricsMonthly._base_manager.filter(pk__in=stale_pks).delete()
```
kirtimanmishrazipstack: **[High] [Lens 5] — The orphan sweep is unscoped, so an incomplete daily tier deletes monthly rows rather than leaving them alone**

The monthly-derived-from-daily trade is deliberate and documented. This is narrower: the sweep deletes every row with `month >= month_start` whose key the rollup did not produce, and a key is absent for two indistinguishable reasons — the metric genuinely has no data, or the daily tier has no rows for it.

The `if not objects: return 0` guard covers only a *totally* empty daily tier. A partially populated one — one org missing, one month missing — passes it and proceeds to delete. Reachable today via `backfill_metrics --skip-daily`, a `cleanup_daily_metrics` run with short retention, or a daily tier younger than two months.

**Fix:** scope the delete to keys the rollup demonstrably covered — collect the `(organization_id, month)` pairs present in `objects` and only delete stale rows whose pair is in that set. A metric that genuinely stopped emitting is still dropped; a month/org the rollup produced nothing for is left alone, which is the stance the sibling guard already takes for the empty case.

backend/dashboard_metrics/tasks.py
```diff
@@ -374,82 +406,175 @@ def _aggregate_llm_combined(
...
+    if daily_agg:
+        stats["daily"]["upserted"] += _bulk_upsert_daily(daily_agg)
+
+    stats["orgs_processed"] += 1
+
+
+def _active_org_ids(end_date: datetime) -> set:
+    """Organizations with execution activity in the prefilter lookback."""
+    return set(
+        WorkflowExecution.objects.filter(
```
kirtimanmishrazipstack: **[Medium] [Lens 2] — The prefilter ignores `source_window_days`, so the reconcile run can skip the orgs it exists to repair**

`_active_org_ids` hardcodes `DASHBOARD_ACTIVE_ORG_LOOKBACK_DAYS` and never sees `source_window_days`. On the 7-day reconcile run, `daily_start` is truncated to midnight (reaching back up to 7d23h) while this cutoff is exactly `now - 7d` — so the prefilter is *narrower* than the query window.

Before this PR the two were aligned by construction: the cutoff **was** `daily_start`. An operator repairing a longer outage with `source_window_days=30` now gets a 30-day query that never runs for any org without a `WorkflowExecution` in the last 7 days, and the run reports `errors: 0`.

The comment above justifying 7 as "wider than the source window" holds only at the default of 2.

**Fix:** `min(daily_start, end_date - timedelta(days=DASHBOARD_ACTIVE_ORG_LOOKBACK_DAYS))`, so the prefilter is never narrower than the window while keeping the `approved_at` floor.

backend/dashboard_metrics/tasks.py
```diff
@@ -33,6 +36,17 @@
 DASHBOARD_HOURLY_METRICS_RETENTION_DAYS = 30
 DASHBOARD_DAILY_METRICS_RETENTION_DAYS = 365
 
+# Daily-tier source lookback, sized against the worst observed
+# created_at -> terminal-status lag.
+DASHBOARD_SOURCE_WINDOW_DAYS = 2
+
+# Wider lookback for the once-daily reconciliation pass.
+DASHBOARD_RECONCILE_WINDOW_DAYS = 7
```
kirtimanmishrazipstack, line 49: **[Medium] [Lens 15] — `DASHBOARD_RECONCILE_WINDOW_DAYS` drives nothing in production**

Repo-wide, this constant is read by its own definition, one docstring, and six test sites. Nothing in production reads it — the scheduled window is the literal `'{"source_window_days": 7}'` in migration 0005. Editing it changes nothing.

The docstring below compounds it by telling a maintainer the reconciliation pass "reruns this task at `DASHBOARD_RECONCILE_WINDOW_DAYS`". A maintainer widening it after a long outage gets no change in behaviour, and the suite stays green because the test asserting migration-kwargs == constant passes only because both happen to be 7.

**Fix:** pick one home. A migration must not import live app code, so either drop the constant and let the docstring cite the migration, or keep it and say in a comment that it mirrors the migration's literal rather than driving it.

backend/dashboard_metrics/tasks.py
```diff
@@ -460,109 +585,40 @@ def _run_aggregation() -> dict[str, Any]:
...
-
+            _aggregate_org(org, hourly_start, daily_start, end_date, stats)
         except Exception:
-            logger.exception("Error processing org %s", org_id)
+            logger.exception("Error processing org %s", org.id)
             stats["errors"] += 1
 
+    try:
+        stats["monthly"]["upserted"] = _rollup_monthly_from_daily(monthly_start)
+    except Exception:
```
kirtimanmishrazipstack, line 657: **[Medium] [Lens 3] — This `except Exception` swallows exactly the two errors the task is configured to retry**

The task decorator sets `autoretry_for=(DatabaseError, OperationalError)`. This catch sits inside the task and swallows both, so a lock timeout on `event_metrics_monthly`, a statement timeout on the `TruncMonth`/`Sum` aggregate, or an `IntegrityError` against the conflict target is logged once and the task returns `success: True` — the retry never fires.

Because the rollup is now a *single* statement covering every org rather than the old per-org upsert, one swallowed exception loses the entire monthly refresh for that run, and a persistent one leaves monthly permanently stale behind 96 successful-looking runs a day. `monthly.upserted == 0` on this path is also byte-identical to the legitimate no-op return.

**Fix:** let `DatabaseError`/`OperationalError` propagate so the configured retry fires. If it must stay non-fatal, mark the result distinctly (`"failed": True`) and log at ERROR.

backend/dashboard_metrics/tasks.py
```diff
@@ -165,42 +179,83 @@ def _bulk_upsert_daily(aggregations: dict) -> int:
...
-        update_conflicts=True,
-        unique_fields=["organization", "month", "metric_name", "project", "tag"],
-        update_fields=["metric_type", "metric_value", "metric_count"],
-    )
+    fresh_keys = {
+        (o.organization_id, o.month, o.metric_name, o.project, o.tag) for o in objects
+    }
+
+    with transaction.atomic():
+        EventMetricsMonthly._base_manager.bulk_create(
```
kirtimanmishrazipstack: **[Medium] [Lens 9] — First unbounded queryset in this module, on a 15-minute cadence, in a PR about cutting DB load**

Every other bulk write here is per-org and naturally bounded. This one is all-orgs-at-once: `orgs x metrics x 2 months` rows materialised into a Python list, a second full fetch of the monthly window materialised for the orphan sweep, and one `INSERT ... ON CONFLICT` with no `batch_size` — Django 4.2's base `bulk_batch_size` returns `len(objs)` and the PostgreSQL backend does not override it, so it really is one statement. Statement size, worker memory and the transaction's lock-hold on `event_metrics_monthly` all scale linearly with tenant count. `ARCHITECTURE_PRINCIPLES.md` §6 states "No unbounded querysets."

**Fix:** pass an explicit `batch_size`; iterate the aggregate queryset rather than list-comprehending it whole; express the orphan sweep as a SQL-side `.exclude()` delete instead of a Python anti-join. Moving the sweep to the nightly reconcile run would take it off the hot path entirely.

backend/dashboard_metrics/tasks.py
```diff
@@ -165,42 +179,83 @@ def _bulk_upsert_daily(aggregations: dict) -> int:
...
 
-    Args:
-        aggregations: Dict keyed by (org_id, month_str, metric_name, project, tag)
+    deleted, _ = EventMetricsMonthly._base_manager.filter(pk__in=stale_pks).delete()
+    return deleted
 
-    Returns:
-        Number of rows upserted
+
+def _rollup_monthly_from_daily(month_start: date) -> int:
```
kirtimanmishrazipstack: **[Medium] [Lens 2] — `backfill_metrics` is now a second, contradictory writer of this table**

The module now has two producers of the monthly tier with incompatible ownership models: this rollup derives it from daily and asserts exclusive ownership of the last two months (deleting anything it did not produce), while `backfill_metrics` still computes monthly from source.

Within the rollup window the command's monthly output is overwritten or deleted within 15 minutes, so `--skip-monthly` is a no-op there and `--skip-daily` is actively harmful. That matters because `backfill_metrics` is the documented recovery tool for exactly the >7-day gaps this change creates, and the README still says the two writers coexist safely because "upserts are idempotent" — which no longer holds now that one writer deletes the other's rows.

**Fix:** have the command call `_rollup_monthly_from_daily` for its monthly step so there is one derivation. At minimum, document in its help text and the README that monthly for the last two months is owned exclusively by the aggregation task.

backend/dashboard_metrics/tests/test_tasks.py
```diff
@@ -198,3 +215,318 @@ def test_cleanup_no_records_to_delete(self):
...
+        self._daily(date(2024, 3, 5), value=10, metric_type=MetricType.HISTOGRAM)
+        self._daily(date(2024, 3, 6), value=5, metric_type=MetricType.COUNTER)
+
+        assert _rollup_monthly_from_daily(date(2024, 3, 1)) == 1
+
+        rows = self._monthly_rows()
+        assert len(rows) == 1
+        assert rows[0].metric_value == 15
+
+    def test_no_daily_rows_upserts_nothing(self):
```
kirtimanmishrazipstack: **[Medium] [Lens 13] — This test cannot fail; the mutation it exists to catch passes it**

The production guard it covers (`if not objects: return 0`) exists solely to stop an empty daily tier being read as "every month is zero". Delete that guard and control falls through to `_delete_orphan_monthly(month_start, set())`, which deletes **every** monthly row with `month >= month_start`, for every org.

This test seeds no daily rows *and no monthly rows*, so both the guarded and unguarded implementations leave an empty table and the assertion passes either way. It converts "unverified" into "verified" for the one path whose failure loses the whole monthly tier.

**Fix:** seed a monthly row inside the window, delete all daily rows, call the rollup, and assert the pre-existing monthly row **survives**. That fails the moment the early return is removed.

backend/dashboard_metrics/tests/test_tasks.py
```diff
@@ -198,3 +215,318 @@ def test_cleanup_no_records_to_delete(self):
...
+        result = self._run_with_active_org(
+            source_window_days=DASHBOARD_RECONCILE_WINDOW_DAYS
+        )
+
+        expected = _truncate_to_day(
+            timezone.now() - timedelta(days=DASHBOARD_RECONCILE_WINDOW_DAYS)
+        )
+        assert result["period"]["daily"]["start"] == expected.isoformat()
+
+    def test_task_passes_the_window_through(self):
```
kirtimanmishrazipstack, line 862: **[Medium] [Lens 13] — This test takes and then deletes a real shared Redis lock**

It calls `aggregate_metrics_from_sources()` for real, which acquires the lock and then `cache.delete(AGGREGATION_LOCK_KEY)` in the task's `finally`. The default cache is `django_redis.cache.RedisCache`; neither `test_base.py` nor `conftest.py` overrides `CACHES` and `test.env` sets no `REDIS_*` — so this is the developer's real Redis, the same instance the local stack's 15-minute aggregation uses.

Two consequences: the test fails if a genuine run holds the lock when it executes, and — worse — its `finally` **releases a lock a live aggregation is holding**, which is precisely the overlap the lock exists to prevent.

**Fix:** this test only cares about kwarg forwarding — patch `_acquire_aggregation_lock` alongside the existing `_run_aggregation` patch, or `@override_settings` a LocMemCache.

kirtimanmishrazipstack (commented):
## Standardized review — FOLLOWUP

Full 19-lens pass at head `5804a1292`, diff scoped to `eebaa03f0..HEAD` (3 files, file count reconciled against GitHub). Mediums and Highs only — Lows withheld by request.

**3 High, 9 Medium.**

The monthly-derived-from-daily trade and the orphan deletion in principle are **already settled** and documented in the PR body; I have not re-raised them. Everything here is either new or a narrower refinement of what was accepted.

The one that changes a previous conclusion is the PG-transport finding on migration 0005. I earlier closed the missing PG twin as "not a bug" on the grounds that `converge_pg_scheduler --periodics` mirrors the row. It does — but I never checked what happens when the mirrored row is *dispatched*, and that is where it breaks.

### Unanchored findings

- **[Medium] [Lens 13] `backend/dashboard_metrics/tests/test_pg_periodic_task_declarations.py:20-31`** (not in this diff) — the Beat/PG drift guard hardcodes `_BEAT_MIGRATION = "...0002_setup_periodic_tasks"` and a fixed three-name `_EXPECTED_CRON`. It exists precisely to catch "a schedule added to Beat but not PG", and this PR adds a fourth Beat schedule in a different module, so it keeps comparing three against three and stays green while the invariant it names is violated. It reads as green cover for the PG finding above. Fix: discover every Beat-declaring migration in the app rather than naming one module.
- **[Medium] [Lens 17] `backend/dashboard_metrics/README.md:49, 112, 301, 377-388`** (not in this diff) — documents the daily query window as 7 days (now 2) and describes monthly as coming from source queries grouped by month, which is what this PR removed. The task table is also short by one row. A reader debugging a missing monthly figure follows it to the source tables and never reaches the daily tier that actually feeds it.

### Coverage

Lenses 5, 6, 8, 11, 12, 14, 17, 18, 19 assessed by me directly; 1, 2, 3, 9, 10, 15 by the correctness pass; 13 by the test pass; 16 by the comment pass. Lens 4 clean — boundary sweep traced 5 log sinks, only enum values and internal org ids reach them. Lens 12 and 14 N/A (no LLM/agent surface, no dependency changes).

backend/dashboard_metrics/tests/test_tasks.py
```diff
@@ -198,3 +215,318 @@ def test_cleanup_no_records_to_delete(self):
...
+        assert "event_metrics_daily" in sql
+        for source_table in (
+            "workflow_file_execution",
+            "workflow_execution",
+            "page_usage",
+        ):
+            assert source_table not in sql, f"monthly rollup read {source_table}"
+
+
+class TestSourceWindow(TestCase):
```
kirtimanmishrazipstack, line 826: **[Medium] [Lens 13] — The PR's actual behaviour change has no test through the real entry point**

`EventMetricsMonthly` appears in this file only inside `TestMonthlyRollup._monthly_rows`. This class calls `_run_aggregation()` six times and never looks at a monthly row, at `result["monthly"]`, or at `result["period"]["monthly"]["start"]`.

Two things are therefore uncovered: the new `monthly_start` arithmetic that replaced the old explicit January branch, and the destructive orphan delete as it actually runs — against a monthly table that already holds rows from previous runs. Every orphan-delete test calls the helper directly with a hand-chosen `month_start`; none goes through the code that computes it. If `monthly_start` regressed to the first of the *current* month, `TestMonthlyRollup` still passes in full and the previous month's rows are silently dropped from the rebuild.

**Fix:** one test that seeds daily rows in both months plus a monthly row *before* the window, runs `_run_aggregation()`, and asserts the computed `monthly_start` and that the out-of-window row survives.

backend/dashboard_metrics/tests/test_tasks.py
```diff
@@ -198,3 +215,318 @@ def test_cleanup_no_records_to_delete(self):
...
         assert result["success"] is True
         assert result["deleted"] == 0
+
+
+class TestMonthlyRollup(TestCase):
+    """Tests for deriving monthly metrics from the daily tier."""
+
+    def setUp(self):
+        """Set up test fixtures."""
+        self.org = Organization.objects.create(
```
kirtimanmishrazipstack, line 238: **[Medium] [Lens 13] — Every rollup test uses one org, so the org component of the delete key is unpinned**

Both `_rollup_monthly_from_daily` and `_delete_orphan_monthly` go through `_base_manager`, deliberately bypassing the org-scoped default manager, and the delete scans every org's monthly rows for `month >= month_start`.

Drop `organization_id` from either key tuple and every test here still passes, because with one org the tuples still match. With two orgs it misbehaves: org A's monthly row for a month where only org B has daily rows matches a "fresh" key and survives as a stale total. Org-scoping is a stated non-negotiable in this repo and this is a destructive cross-org query.

**Fix:** one test seeding two orgs in the same month, asserting per-org totals and that deleting org A's daily rows removes only org A's monthly row.

backend/dashboard_metrics/tasks.py
kirtimanmishrazipstack: Took the "at minimum" option rather than unifying the derivation, deliberately.

Documented in three places: the command's module docstring, the `--skip-daily` / `--skip-monthly` help text, and the README — all saying the current and previous month are owned exclusively by the aggregation task. `--skip-daily` without `--skip-monthly` now also prints a warning at run time, since that is the actively harmful combination rather than the merely useless one.

Not calling `_rollup_monthly_from_daily` for the command's monthly step: the two have different shapes. The command backfills an arbitrary historical window and supports `--org-id`, while the rollup helper is all-orgs and anchored to the rollup window — so wiring them together would silently widen a single-org backfill to every org, and reach months the caller did not ask for. One derivation is the right end state, but it wants the command's window semantics reworked, not a call swapped in here.

backend/dashboard_metrics/tasks.py
kirtimanmishrazipstack: Fixed the two parts that were real; the third does not apply.

**`batch_size`** — `bulk_create` now passes an explicit `MONTHLY_ROLLUP_BATCH_SIZE`, so the statement no longer scales with tenant count.

**The sweep no longer reads the whole monthly window.** It is scoped to the `(organization, month)` partitions the rollup actually produced — which is the same fix as the Lens 5 finding above, and it bounds the fetch to the rows the rollup already materialised instead of every org's.

**Not iterating the aggregate queryset** — `bulk_create` takes a list, so `objects` has to be materialised whatever happens. Iterating would move the materialisation, not remove it.

kirtimanmishrazipstack:
## Review findings addressed — 3 High, 9 Medium, 2 unanchored

All fixed except two answered inline (`backfill_metrics` derivation, iterating the aggregate queryset).

### High

- **Deletion count discarded** — `_rollup_monthly_from_daily` now returns `(upserted, deleted)`. The count lands in `stats["monthly"]["deleted"]`, in the task result, and in a WARNING log when non-zero.
- **Reconcile row unrunnable on the PG transport** — `dashboard_metrics_aggregate` takes `source_window_days`, `AggregateMetricsAPIView` forwards it, and `0005` now declares the **PG twin** alongside the Beat row instead of leaving the mirror to invent one.
- **Orphan sweep unscoped** — scoped to the `(organization, month)` partitions the rollup produced. A partition it produced nothing for is left alone: absent daily rows mean an incomplete tier, not a metric that went to zero.

### Medium

- **Prefilter narrower than the window** — `_active_org_ids(end_date, window_start)` takes `min(window_start, now - LOOKBACK)`, so a widened `source_window_days` can no longer skip the orgs it exists to repair.
- **`DASHBOARD_RECONCILE_WINDOW_DAYS` drove nothing** — kept, with a comment saying it mirrors `0005`'s literal rather than driving it.
- **`except Exception` swallowed the retry set** — `DatabaseError` / `OperationalError` propagate so `autoretry_for` fires; everything else keeps the non-fatal posture.
- **Unbounded bulk_create** — explicit `batch_size`; the sweep no longer reads the whole monthly window.
- **`backfill_metrics` as a second writer** — documented in the module docstring, both `--skip-*` help strings and the README, plus a run-time warning on `--skip-daily` without `--skip-monthly`.
- **Four test findings** — the empty-tier test now seeds a monthly row (it could not fail before); the kwarg-forwarding test no longer takes and releases the shared Redis lock; a new `TestMonthlyThroughTheTask` runs the real `_run_aggregation` and pins the computed `monthly_start` and the sweep against a populated table; two new multi-org tests pin the organization half of the delete key.
- **Beat/PG drift guard was stale by construction** — rewritten. It discovers **every** migration in the app, replays their `RunPython` forwards in order against fake models, and compares the captured rows. It no longer names a module, so the next schedule is in scope automatically. It also now derives the Beat cadence from the schedule row rather than a hardcoded table, binds every declared kwarg set to its task signature, and asserts every post-install Beat write bumps `PeriodicTasks.last_update`.
- **README** — daily window (7 → 2), monthly now derived from daily, the reconciliation row, and the ownership overlap with `backfill_metrics`.

### One change nobody asked for

The reconcile row moved from **04:00 to 04:40 UTC**. `*/15` fires at :00, so at 04:00 the reconciliation (all tiers, 7-day window) started alongside the 15-minute run, and their lock keys do not block each other — the same collision raised against #2265's new row. Off the `*/15` grid now.

backend/dashboard_metrics/tasks.py
kirtimanmishrazipstack: Fixed in b974eaea5.

`_rollup_monthly_from_daily` now returns `(upserted, deleted)`. The count lands in `stats["monthly"]["deleted"]`, in the task result, in the completion log line, and in a dedicated WARNING when non-zero — rare in steady state, so it alerts rather than adds noise.

`TestMonthlyThroughTheTask.test_the_run_reports_what_it_deleted` pins it through `_run_aggregation`; dropping the `stats` assignment fails it.

backend/dashboard_metrics/migrations/0005_add_reconciliation_task.py
kirtimanmishrazipstack: Fixed in b974eaea5, and one step further than suggested.

`dashboard_metrics_aggregate` takes `source_window_days: int | None = None` and builds the body the way the two `cleanup_*` tasks already do; `AggregateMetricsAPIView` forwards it through `_int_arg`, omitting it when absent so the task's own default applies rather than one invented at the boundary.

Beyond that: **`0005` now declares the PG twin itself.** The row existed on Beat only and relied on `mirror_pg_periodic_tasks` copying it — same shape as `0002`/`0004`, which declare both sides explicitly. The generalised drift guard would have failed on the Beat-only row, which is the right answer.

Covered on both sides of the Django boundary: `test_every_declared_kwarg_set_binds_to_the_task_signature` binds every migration-declared kwarg set to the Django signature, and `test_every_scheduled_kwarg_set_binds_to_the_proxy` does the same for the worker proxy, which cannot import Django.

backend/dashboard_metrics/tasks.py
kirtimanmishrazipstack: Fixed in b974eaea5, taking the suggested scoping.

`_delete_orphan_monthly(objects)` derives the `(organization, month)` partitions the rollup produced and only sweeps inside those. A metric that genuinely stopped emitting is still dropped; a partition the rollup produced nothing for is left alone.

Two tests, and both were checked by mutation:
- `test_a_month_with_no_daily_rows_of_its_own_is_left_alone` — the partially-populated case that passed the `if not objects` guard
- `test_the_sweep_is_scoped_per_organization` / `test_a_stale_row_is_dropped_for_one_organization_only` — two orgs in one month

Worth recording: the scoping makes the `if not objects` guard **redundant**. I mutated the guard away and the suite stayed green, because an empty `objects` yields no covered partitions and the loop never runs. It is kept as an early return to skip a pointless transaction, and its comment now says so rather than claiming to be the protection. Removing the guard *and* unscoping the sweep — the real pre-fix shape — fails five tests.

backend/dashboard_metrics/tasks.py
kirtimanmishrazipstack: Fixed in b974eaea5 with exactly the suggested `min`.

`_active_org_ids(end_date, window_start)` uses `min(window_start, end_date - DASHBOARD_ACTIVE_ORG_LOOKBACK_DAYS)`, so the prefilter is never narrower than the window while keeping the `approved_at` floor at the default.

`TestActiveOrgPrefilter` seeds an org whose only execution is 10 days old and asserts it is excluded at the default window and included at 30 days. Reverting to the hardcoded lookback fails the second.

backend/dashboard_metrics/tasks.py
kirtimanmishrazipstack: Fixed in b974eaea5 — kept, with the relationship stated.

A migration must not import live app code, so the literal in `0005` has to stay. The constant now carries a comment saying it mirrors that literal rather than driving the schedule, and the task docstring no longer implies editing it moves anything. One home for the value, and the direction of the dependency is written down where a maintainer widening it will read it.

backend/dashboard_metrics/tasks.py
kirtimanmishrazipstack: Fixed in b974eaea5.

`DatabaseError` / `OperationalError` re-raise so the configured `autoretry_for` fires; everything else keeps the non-fatal `stats["errors"]` posture, which is still right for a per-org data fault.

Two tests: `test_a_database_error_propagates_instead_of_reporting_success` and `test_an_unexpected_error_is_still_counted_rather_than_fatal`. Deleting the re-raise fails the first.

backend/dashboard_metrics/tests/test_tasks.py
kirtimanmishrazipstack: Fixed in b974eaea5, exactly as suggested.

The test now seeds a monthly row inside the window before calling the rollup with an empty daily tier, and asserts it survives with its value intact. Renamed to `test_an_empty_daily_tier_leaves_existing_monthly_rows_alone`, and the docstring says why the seeding is what makes the failure reachable at all.

One correction to the finding, from mutation: removing the early return alone no longer fails it — the scoped sweep from the Lens 5 fix already prevents the wipe, so the guard is now belt-and-braces. Removing the guard **and** unscoping the sweep does fail it, along with four others. I have left the guard in place (it skips a pointless transaction) and rewritten its comment so it no longer claims to be the protection.

backend/dashboard_metrics/tests/test_tasks.py
kirtimanmishrazipstack: Fixed in b974eaea5.

The test now patches `_acquire_aggregation_lock` and `cache` alongside the existing `_run_aggregation` patch. It only ever cared about kwarg forwarding, so nothing is lost — and it can no longer release a lock a live local aggregation is holding, which was the worse of the two failure modes.

backend/dashboard_metrics/tests/test_tasks.py
kirtimanmishrazipstack: Fixed in b974eaea5 — new `TestMonthlyThroughTheTask`.

`test_the_window_covers_the_previous_month_and_spares_what_precedes_it` seeds daily rows in both months plus a monthly row *before* the window, runs the real `_run_aggregation()`, and asserts the computed `period.monthly.start` is the first of the previous month and that the out-of-window row survives. A regression to the first of the current month fails it.

`test_the_run_reports_what_it_deleted` runs the sweep against a monthly table that already holds rows from a previous run, which was the other half of the gap.

backend/dashboard_metrics/tests/test_tasks.py
kirtimanmishrazipstack: Fixed in b974eaea5 — two tests, two different failures.

`test_the_sweep_is_scoped_per_organization` seeds two orgs in the same month and deletes one org's daily rows, asserting the other org's monthly row is untouched. `test_a_stale_row_is_dropped_for_one_organization_only` gives both orgs two metrics and removes one metric from one org, asserting only that row goes.

Dropping `organization_id` from the key tuple fails the second; the unscoped sweep fails the first.

backend/dashboard_metrics/tasks.py
kirtimanmishrazipstack: @athul-rs Valid, and decisive on the spec point — **the delete is gone** (2398a4332).

I checked comments 44768/45016 rather than arguing from the code. The SQL you accepted is a pure `INSERT ... ON CONFLICT DO UPDATE`; there is no DELETE in it. The sweep was scope I added beyond the agreed design, so removing it is conformance, not a concession.

Your reasoning holds independently of that. A stale row is recoverable — `backfill_metrics` rewrites it. A deleted one is not, because the daily rows that would rebuild it are exactly what was missing. On the partial-gap case: the total now tracks whatever the daily tier holds, so repairing daily repairs monthly on the next run, which is what makes upsert-only recoverable rather than merely less destructive. `test_a_partially_repopulated_month_is_overwritten_not_accumulated` pins that round trip.

`_rollup_monthly_from_daily` is back to returning one count and its docstring says upsert-only and why. The README and `backfill_metrics`' help text no longer describe an ownership model that no longer exists.

backend/dashboard_metrics/tasks.py
kirtimanmishrazipstack: @athul-rs Correct, and it survived my own review — I had scoped the sweep per `(organization, month)` partition, which is your "sound *per month*" reading, and mutation-tested that. What I did not check is what you did: that with 38 orgs `objects` is never empty, so the guard as written only ever fired in tests.

Moot now that the delete is gone (2398a4332). The guard remains as an early return purely to skip a pointless transaction, and its comment says that rather than claiming to protect anything.

Worth recording for the fix pass: mutating the guard away left the suite green even *with* the scoped sweep, because scoping subsumed it. Reading the test would not have told me that.

backend/dashboard_metrics/migrations/0005_add_reconciliation_task.py
kirtimanmishrazipstack: @athul-rs Fixed in 2398a4332, and the trace is right at every step.

`dashboard_metrics_aggregate` takes `source_window_days`, the endpoint forwards it through `_int_arg`, and `0005` now declares the **PG twin itself** rather than relying on the mirror to synthesise one — matching 0002/0004, which declare both sides explicitly.

On your "also": the drift guard no longer names a migration. It discovers every migration in the app, replays their `RunPython` forwards in order against fake models, and compares the captured Beat and PG rows in their final state. That change is what forced the PG twin — with the Beat-only row it failed, correctly. It also derives the Beat cadence from the schedule row rather than a hardcoded table, and binds every declared kwarg set to its task signature.

backend/dashboard_metrics/migrations/0005_add_reconciliation_task.py
kirtimanmishrazipstack: @athul-rs Fixed in 2398a4332, taking your first option.

`_aggregation_lock_key(source_window_days)` gives each schedule its own key, so the drifting 15-minute interval can no longer take the key the 04:40 crontab needs. Distinct windows are distinct jobs; both are idempotent upserts, so the once-a-day overlap costs duplicated work at worst — which is a much better trade than losing the repair pass roughly one day in seven and reporting it as success.

The skip also moved from INFO to WARNING and now names the window, so a run that does not happen is greppable.

Two related things while in here: the reconciliation row moved to **04:40**, because `*/15` fires at minute 0 and 04:00 put the 7-day all-tiers run alongside the 15-minute one; and on #2265 the same reasoning generalises — keys are per granularity written and namespaced by window, so `ALL` genuinely excludes a concurrent `hourly` run while the reconciliation still cannot be starved.

backend/dashboard_metrics/tasks.py
kirtimanmishrazipstack: @athul-rs Fixed in 2398a4332. All three parts.

`success` is no longer a literal — it is `stats["errors"] == 0`. The completion log rises to WARNING when the error count is non-zero. And the worker-side guard now reads all three did-nothing shapes rather than only `skipped`: `skipped`/`reason`, `skipped_reason`, and `errors`. You are right that it saw none of the latter two, which is exactly the case its own docstring says it exists for.

backend/dashboard_metrics/tasks.py
kirtimanmishrazipstack: @athul-rs Fixed in 2398a4332.

The rollup sets `stats["monthly"]["failed"] = True` on the non-retryable path, so a failure is no longer byte-identical to the legitimate empty rollup or to `no_active_orgs` — and with `success` now derived from the error count the run reports itself failed as well. `DatabaseError`/`OperationalError` propagate so the configured `autoretry_for` actually fires; they were being swallowed by the same `except Exception`.

On the blast radius: the single all-orgs call stays, but its worst case is now bounded differently — with the delete removed, a failed rollup leaves the previous monthly totals in place rather than a partially rebuilt tier.

On the "also": agreed that a line in a PR description is not enforcement, and I have not tried to enforce it in code — a migration cannot know whether a 60-day backfill has finished. What changed is that the consequence is no longer destructive: if the daily tier is short when the first rollup runs, monthly is written under-counted and **repairs itself** on the next run once the backfill lands. Rolling back 0005 restoring nothing is no longer true either, because there is nothing deleted to restore.

backend/dashboard_metrics/tasks.py
kirtimanmishrazipstack: @athul-rs Fixed in 2398a4332 with exactly the `min` you wrote.

`_active_org_ids(end_date, window_start)` takes `min(window_start, end_date - DASHBOARD_ACTIVE_ORG_LOOKBACK_DAYS)`, so the prefilter can never be narrower than the window it is filtering for. `TestActiveOrgPrefilter` seeds an org whose only execution is 10 days old and asserts it is excluded at the default window and included at 30 days; reverting to the hardcoded lookback fails the second.

Your framing of it as a regression rather than a latent bug is the useful part — on `main` the two were the same value by construction, so there was nothing to keep in step.

backend/dashboard_metrics/tasks.py
kirtimanmishrazipstack: @athul-rs Fixed in 2398a4332 — I made the code match the comment rather than weakening the comment.

With the `min` in place the prefilter is never narrower than the window, so the constant is now a **floor** rather than a claim about relative width, and the comment says that: at the 7-day reconciliation window the two are equal rather than this one being wider.

backend/dashboard_metrics/tasks.py
kirtimanmishrazipstack: @athul-rs Fixed in 2398a4332.

`_validate_source_window` coerces and bounds it, called from both the task and `_run_aggregation`. Rejects non-integers, anything below 1, and anything above `MAX_SOURCE_WINDOW_DAYS = 90` — your three cases, and the upper bound is set where it stops being a repair pass and starts being the multi-month scan this ticket removes.

The endpoint already reuses `_int_arg` for the same field, so the Beat leg and the PG leg now validate equivalently. `TestSourceWindowValidation` covers all four.

backend/dashboard_metrics/tasks.py
kirtimanmishrazipstack: @athul-rs Resolved by deletion rather than by renaming — the function no longer deletes anything (2398a4332), so there is nothing for the name to conceal.

Its docstring now states the upsert-only stance and why: a stale total is recoverable, a deleted one is not. The `_delete_orphan_monthly` helper is gone entirely.

Your point about the docstring implying "formerly-derived, now-orphaned" while the code was unconditional was the sharpest part of this — it would have deleted rows written by `backfill_metrics` or by a manual correction, and neither is an orphan in any sense the reader would expect.

backend/dashboard_metrics/tests/test_tasks.py
kirtimanmishrazipstack: @athul-rs Both gaps fixed in 2398a4332, and the docstring typo with them.

**(1)** `TestMonthlyThroughTheTask` reads `EventMetricsMonthly` after a real `_run_aggregation()`, asserts the computed `monthly_start` is the first of the *previous* month, and runs the rollup against a table that already holds rows from an earlier run. Deleting the rollup call now fails it.

**(2)** `TestMonthlyMatchesTheOldDerivation` is the golden test, built the way you specified: seed source rows, let the real aggregation populate the daily tier from them, then compute the oracle independently — `get_documents_processed` at DAY granularity bucketed by `_truncate_to_month` — and compare. It carries an anti-vacuity guard that corrupts one day of the daily tier and asserts the comparison then fails, so an oracle that always matches cannot pass silently.

One thing your review did not raise but the same class of bug: the fixture offsets derive from the month boundary rather than fixed day counts. "40 days ago" lands outside the monthly window entirely for part of the calendar, and the cross-boundary coverage disappears without failing.

Docstring corrected to 0005.

PR #2256, opened 2026-08-29
UN-3016 [FIX] Record why an execution failed; stop dispatching skipped files

hari-kuriakose:
## UN-3016 — an execution failed with no explanation

A customer (Moody's) execution ended in ERROR with nothing telling the user why. Root-caused from code and then **corroborated against the actual prod row**:

| Field | Value |
| --- | --- |
| status | `ERROR` |
| error_message | *(blank)* |
| total_files | 1 |
| file | `...Application.xlsm` (`application/vnd.ms-excel.sheet.macroenabled.12`) |
| file_hash | `temp-hash-9d260345dd444238a58eb42ab562bcd7` |

The prod logs for this run are long gone (30-day retention vs a 283-day-old incident), so the DB row is the evidence.

### What was happening

The `temp-hash-` prefix is written in exactly one place — the backend MIME-skip branch — which pins the mechanism:

1. The backend correctly rejected the `.xlsm` and took the **skip** branch, which deliberately never writes the file's bytes. But it returned the file with `is_executed=True` and a fake hash.
2. **Nothing downstream filters on `is_executed`**, so the file was counted in `total_files` and dispatched to a worker anyway.
3. The worker failed on a file that was never staged. Both halves of the stored error are bare paths — what `FileNotFoundError(path)` stringifies to.
4. `failed_files == total_files` (1 == 1) escalated the execution to ERROR.
5. `_determine_execution_status_unified` returned only a status and counts — **no reason string** — so every call site passed `error_message=None`. `WorkflowExecution.update_execution()` guards on `if error:`, so `error_message` was never written.

Verified present at `0da9da5a8`, the commit that was HEAD the day the ticket was filed — pre-existing, not a regression.

### Changes

1. **Executions now record why they failed.** `_determine_execution_status_unified` returns a reason built from the per-file errors already aggregated in `aggregated_results`; all call sites pass it through, including notifications. Capped to the `CharField(256)` that otherwise truncates *silently*.
2. **A skipped file is no longer dispatched.** Unsupported files are excluded from the staged set rather than returned with a fake hash. If *every* file is skipped, the request fails with a 400 naming the types instead of dispatching an empty execution that would finish as a vacuous success.
3. **Workers' `AllowedFileTypes` was missing XLSM** while the backend had it. The two MIME lists are now identical, so a file the API accepts cannot be rejected again inside the worker.

### PGMQ

All three apply to the PGMQ stack. It is not a separate callback path — it shares `_process_batch_callback_core` and `process_batch_callback_api` with `is_pg` as a kwarg flag, so both patched call sites cover it. Staging is backend-side and transport-agnostic, and the PG stack has no MIME list of its own. Its own error paths (`[pg-poison-drop]`, `[pg-barrier-abort]`, `[reaper-recovery]`) already wrote descriptive reasons and never had this defect.

### Tests

`workers/tests/test_un3016_execution_error.py` — 9 passing: the real incident error shape, the never-blank guarantee, the 256-char column fit, and a regression guard that no caller reintroduces `error_message=None`.

### Reviewer notes

- **The full worker suite was not run locally** — it needs the repo's `unstract.core` package, absent from this checkout. The 3→4 tuple change on `_determine_execution_status_unified` is the thing to watch; the two PG tests touching it mock with `assert_not_called`, so they should be unaffected, but **CI is the real check**.
- **Ruff was not run** (not installed here). Files compile.
- **Known, deliberately not fixed here:** the per-file error text is still a bare path pair built by interpolating two `FileNotFoundError` paths. This PR makes the execution *record* it rather than a blank — an improvement, not a cure. `_handle_null_execution_result` also still publishes nothing to the UI; proven not to be this incident's mechanism, so left alone. Both warrant a follow-up.
- XLSM was absent from the backend enum at incident time and has since been added, so this customer's file would be accepted today. The other two defects remained for any other unsupported type, which is why this is not just an enum entry.


hari-kuriakose:
### Note: partial file skips are silent to the API caller — deliberately not fixed here

Flagged during review and **waived**, not overlooked. Recording the reasoning so it is stated rather than rediscovered later.

**The behaviour.** The guard added in this PR is at `backend/workflow_manager/endpoint_v2/source.py:1294`:

```python
if skipped_files and not file_hashes:
    raise UnsupportedMimeTypeError(
        "No files could be processed. Unsupported file type(s): " + ", ".join(skipped_files)
    )
```

The `and not file_hashes` means this fires only when **every** uploaded file is rejected. On a partial skip — say 10 files uploaded, 2 unsupported — the execution proceeds over the surviving 8 and returns `200`. Nothing in the response names the 2 that were dropped; the reason exists only in the server log written just above (`workflow_log.log_error`). A caller has to count results to notice anything is missing.

**Why it is not fixed in this PR.**

1. Surfacing skipped files in the response is a **response-contract change**. That is out of scope for this fix and wants its own review, since anything parsing the execution response could be affected.
2. It is **pre-existing, and this PR strictly improves it**. Previously the unsupported file was returned with `is_executed=True` and a temp hash; nothing downstream filtered on `is_executed`, so the worker ran it anyway, failed on the missing file, and the whole execution died with an opaque `Execution: <path>; Destination: <path>` error. So the prior behaviour was *one bad file kills the entire run, confusingly*. This PR makes it *one bad file is skipped and the rest succeed*.

**The asymmetry worth knowing about.** Total failure is now loud and explicit; partial failure is quiet. That is a deliberate trade, not an oversight — but it does mean a user can get a successful run over fewer files than they uploaded without the response saying so.

**Suggested follow-up:** report skipped files to the caller, e.g. as a `skipped_files` field on the execution response, so a partial skip is as visible as a total one.

<sub>Reviewed with unstract:lite-remediation; finding F3, waived.</sub>

workers/callback/tasks.py
```diff
@@ -252,12 +252,53 @@ def _get_performance_stats() -> dict:
...
+    errors: dict[str, Any] = aggregated_results.get("errors") or {}
+    # `errors` is keyed by file name, so every entry is distinct already; the
+    # cap is what keeps a large batch from bloating the column.
+    entries = [
+        f"{file_name}: {str(error).strip()}"
+        for file_name, error in errors.items()
+        if error and str(error).strip()
+    ]
+    if not entries:
+        return f"All {total_files} file(s) failed."
```
chandrasekharan-zipstack, line 281: ### [High] [Lens 1, 3, 10, 13] — The error summary is a tautology in production: `aggregated_results["errors"]` is always empty

`_summarize_file_errors` reads only `aggregated_results["errors"]`. That map admits an entry only when `file_result.get("status") == "error"` (lowercase). **No producer ever emits that string:**

- **API path** — `workers/file_processing/tasks.py:1844` wraps every result in `FileExecutionResult`, whose `__post_init__` (`unstract/core/src/unstract/core/worker_models.py:301-305`) *forces* `status` to `ApiDeploymentResultStatus.SUCCESS`/`FAILED` = `"Success"` / `"Failed"` (`worker_models.py:108-112`).
- **ETL/TASK path** — `workers/file_processing/tasks.py:1073-1081` constructs `BatchExecutionResult` with **no `file_results=` argument at all**.

So `errors == {}` always, and `:281` always returns the fallback. **The Moody's incident row would now read `"All 1 file(s) failed."`** — a restatement of `status=ERROR` + `failed_files=1`. The user still cannot see *why*. The blank was replaced by a near-blank.

A repo-wide grep finds no `file_results` producer emitting `"error"`. The PR's own fixture hand-writes `{"status": "error", ...}` at `workers/tests/test_un3016_execution_error.py:85-93`, so `test_status_function_returns_a_reason` (`:97`) is green against a shape production cannot produce.

**Fix:** key on the presence of an error, not on a status string — `time_utils.py:180` → `if isinstance(file_result, dict) and file_result.get("error"):`, keying the name off `file_name` **or** `file`. Separately `file_processing/tasks.py:1073` must pass `file_results=` or the ETL path keeps summarising nothing. Then rebuild the fixture from a real `FileExecutionResult(...).to_dict()`.

Confidence: High — verified independently against PR-head source.

workers/callback/tasks.py
```diff
@@ -1485,7 +1535,7 @@ def _process_batch_callback_core(
                     final_status=execution_status,
                     aggregated_results=aggregated_results,
                     organization_id=context.organization_id,
-                    error_message=None,
+                    error_message=status_error,
```
chandrasekharan-zipstack, line 1538: ### [High] [Lens 3, 6, 10] — The tautology now overwrites specific error messages that previously survived

Both backend writers — `backend/workflow_manager/workflow_v2/models/execution.py:467` and `:496` — apply `if error: self.error_message = error[:EXECUTION_ERROR_LENGTH]` **unconditionally**, and `backend/workflow_manager/internal_views.py:570-573` documents that "ERROR is deliberately NOT protected."

Before this PR the callback passed `error_message=None`, so `if error:` was false and any earlier **specific** message survived. Now a late callback overwrites it.

Concrete sequence: the orchestrator writes a real cause, or `update_execution_err` records the staging cause; the batch callback then fires and replaces it with `"All 2 file(s) failed."`. Combined with H1 this is a **net regression against this PR's own goal** — a row that had a real diagnostic ends up with a tautology.

Same applies at `:1799`.

**Fix:** don't write `error_message` when the row already carries one (`if error and not self.error_message:`), or have the callback pass its summary only when the fetched execution's `error_message` is blank — `_determine_execution_status_unified` already fetches the execution at `:344-359`, so the current value is one field away. If overwrite is intended, say so in code.

Confidence: High on mechanism; Medium on frequency.

workers/shared/enums/file_types.py
```diff
@@ -37,6 +37,9 @@ class AllowedFileTypes(Enum):
     DOC = "application/msword"
     XLSX = "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
     XLS = "application/vnd.ms-excel"
+    # Kept in step with the backend's AllowedFileTypes (UN-3016): the two lists
+    # must agree or a file the API accepts is rejected again inside the worker.
+    XLSM = "application/vnd.ms-excel.sheet.macroenabled.12"
```
chandrasekharan-zipstack, line 42: ### [High] [Lens 1, 3, 13, 16] — This XLSM enum addition is dead code — defect #3 is not fixed

Two independent reasons the new member can never fire:

**1. Wrong path.** The only `AllowedFileTypes.is_allowed` call in `workers/` is `shared/workflow/execution/service.py:1226`, inside `_copy_filesystem_file` (`:1160`). The API path is `_copy_api_file` (`:1104-1157`) and performs **no MIME validation at all**. So the comment's claim — "a file the API accepts is rejected again inside the worker" — describes a rejection that cannot occur.

**2. Wrong MIME source.** On the filesystem path the type comes from `magic.from_buffer(chunk, mime=True)` (`service.py:1223`) — content sniffing, not the client header the backend trusts (`source.py:1233`). **libmagic's database has no `macroenabled` entry.** Verified inside the running worker container: all three of `/usr/share/misc/magic.mgc`, `/usr/lib/file/magic.mgc`, `/usr/share/file/magic.mgc` report `macroenabled=0` while `spreadsheetml=1`. A macro-enabled workbook sniffs as `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` — already allowed before this PR.

Also: zero tests reference `AllowedFileTypes` behaviourally anywhere in `workers/tests`, `backend/**/tests`, or `tests/`.

**Fix:** drop the change, or justify it with a real `.xlsm` fixture driven through `magic.from_buffer`. Correct the comment and the PR description's defect #3. If the lists genuinely must stay in lockstep, add a value-set equality test (it passes today — 22 members each, zero divergence).

Confidence: High — verified in the deployed worker image, not a synthetic fixture.

workers/callback/tasks.py
```diff
@@ -1540,7 +1590,7 @@ def _process_batch_callback_core(
                         workflow_id=context.workflow_id,
                         pipeline_name=context.pipeline_name,
                         pipeline_type=context.pipeline_type,
-                        error_message=None,
+                        error_message=status_error,
```
chandrasekharan-zipstack, line 1593: ### [High] [Lens 13 — Testing] — The entire caller-side half of the fix is unpinned: deleting all four `error_message=status_error` sites keeps CI green

The user-visible half of UN-3016 is "the execution row and the webhook payload carry a reason." Nothing tests that threading. Both callees default the parameter (`tasks.py:417`, `:635`), so a refactor dropping the kwarg silently restores the exact defect — ERROR status, blank `error_message`, webhook with no reason — with CI green.

**Mutation executed.** Deleting lines 1538, 1593, 1799 and 1860 → the new file gives **10 passed**; six callback-adjacent suites give **92 passed**.

The AST guard at `test_un3016_execution_error.py:153-191` only rejects a *literal* `error_message=None` keyword; an omitted keyword contributes no entry to `call.keywords`. Its docstring enumerates the evasions it misses — positional-None, indirection, `**kwargs` splat — but not "delete the keyword," which is the one that actually happens.

**Fix:** one behavioural test per entry point driving `_process_batch_callback_core` / `process_batch_callback_api` with an all-failed batch, asserting `api_client.update_workflow_execution_status.call_args.kwargs["error_message"]` and the notification call both carry the summary. That kills all four mutations and makes the AST guard redundant.

Confidence: High (mutation executed).

backend/workflow_manager/workflow_v2/views.py
```diff
@@ -258,17 +258,21 @@ def execute(
...
+            # partial stage: its guard is exactly this staging condition.
+            if has_uploads:
+                hashes_of_files = SourceConnector.add_input_file_to_api_storage(
+                    pipeline_id=pipeline_guid,
+                    workflow_id=workflow_id,
+                    execution_id=execution_id,
+                    file_objs=file_objs,
+                    use_file_history=False,
+                )
             workflow = self.get_workflow_by_id(workflow_id=workflow_id)
```
chandrasekharan-zipstack, line 276: ### [High] [Lens 13 — Testing] — The staging move is unpinned against dropping the staged hashes

Both tests in `test_un3016_execute_staging_cleanup.py` (`:57`, `:77`) raise before `execute_workflow` is reached, so nothing asserts the staged hashes still flow through.

Losing this assignment dispatches an execution with `hash_values_of_files={}` — the run reports success having processed zero uploaded files. A silent correctness failure on the primary upload flow.

**Mutation executed.** Changing `:269` to drop only the assignment → **2 passed**. No other test exercises `WorkflowViewSet.execute()`.

**Fix:** a third test asserting `execute_workflow.call_args.kwargs["hash_values_of_files"]` is the sentinel returned by staging, and that `use_file_history` is `False`.

Confidence: High (mutation executed).

chandrasekharan-zipstack (changes requested):
## Standardized review (FOLLOWUP) — High-severity findings only

**Verdict: `BLOCK`** — Critical 0 · High 5 · Medium 13 · Low 5 · Lenses 16/16.

This review posts **only the 5 High findings**. The 13 Medium and 5 Low are held back to keep the thread actionable; happy to post them on request.

**Why BLOCK.** The PR's three claimed fixes are (1) record a reason, (2) stop dispatching skipped files, (3) align the MIME lists. **#2 works. #1 produces a content-free string in production and actively destroys better error messages that already existed. #3 is dead code.** Two of the three are certified by tests that pass against shapes production never produces.

---

### Prior round reconciled

The `lite-remediation` loop that produced F1–F4 never posted its findings to this thread, so the ledger below is reconstructed from commit `9e1adc4ca8`'s body and your waiver comment. Previous-review boundary: `9e1adc4ca8`; surface added after it: `af64650ca9`, `1ceb723a5a`, `e15e3e2aad`, `2c95b6a7d8`.

| # | Sev | Title | Status |
|---|---|---|---|
| F1 | High | `UnsupportedMimeTypeError` escaped the try owning `delete_api_storage_dir` | **RESOLVED** |
| F2 | High | Whole-file substring assertion in the caller guard test | **PARTIALLY RESOLVED** |
| F3 | Med | Partial skips invisible to the API caller | **WAIVED** (yours, in writing — UN-4055; not re-raised) |
| F4 | Med | `_EXECUTION_ERROR_MAX_LENGTH` did not name its authority | **PARTIALLY RESOLVED — new defect introduced** |

- **F1 → RESOLVED.** `views.py:268-276` stages inside the try; `has_uploads` at `:261` is literally the same boolean as the cleanup guard at `:297`. Pinned by `test_un3016_execute_staging_cleanup.py:57`. But the move is unpinned on the success side — see **H5**.
- **F2 → PARTIALLY RESOLVED.** Now an AST check scoped to the two functions (`test_un3016_execution_error.py:179-191`). It rejects a literal `error_message=None` keyword but **not the mutation that actually happens** — omitting the keyword, which both callees default to `None`. See **H4**.
- **F4 → PARTIALLY RESOLVED, new defect.** The comment now names `EXECUTION_ERROR_LENGTH` (`tasks.py:257-260`) but cites a convention `undispatched_sweep.py` does not implement, and omits the real second hardcoded copy at `internal_views.py:899`/`:904`. Detail is in the held-back Mediums.
- The two false comments deleted in `9e1adc4ca8` — confirmed gone, not resurrected.

### Cleared, contra the earlier gate

- **OpenAPI needs no update.** The 422 outcome is already documented with `ExecuteResponse` carrying `execution_status`/`error` (`backend/api_v2/openapi_schema.py:206-208`), and 400 at `:143`.
- **The MIME lists are genuinely identical** — programmatic member-for-member diff at PR head: 22 members each, zero divergence. Your claim is accurate (though see H3 — the new entry is unreachable).
- **The 3→4 tuple breaks no consumer** — `_determine_execution_status_unified` is module-local; repo-wide grep plus `unstract-cloud` and `unstract-cloud-platform` find no external unpack.
- **No rate-limit slot leak** on the newly-hot error path. Checked specifically because this PR makes that branch routine; `zrem` is idempotent and the correct release already fires via `update_execution`.
- **Rolling-deploy safe** — `error_message` is pre-existing and optional on both the worker client and the serializer.

**Bots:** SonarCloud passed its gate but reports **0.0% coverage on new code**, corroborating H4/H5. Greptile rated this "5/5 — safe to merge, no concrete changed-code defect remaining"; H1 and H3 contradict that directly.

---

### Open questions

1. **H1 is the crux — was the `errors` map ever observed populated?** Do you have a real execution where `aggregated_results["errors"]` was non-empty? Static reading and `FileExecutionResult.__post_init__` say it cannot be. If so, what does this PR change about the Moody's row beyond blank → `"All 1 file(s) failed."`?
2. **H3 — do you have a real `.xlsm` that libmagic reports as `application/vnd.ms-excel.sheet.macroenabled.12`?** The deployed worker's magic DB has no `macroenabled` entry at all. If not, what does the enum line do?
3. **H2 — is overwriting an existing `error_message` on an already-ERROR row intended?** If yes it should be explicit in code; if not it is a regression this PR introduces.
4. **Should the public API-deployment endpoint return 400 or 422 for an all-unsupported upload?** The description says 400; the code delivers 422 (`deployment_helper.py:286`'s bare `except Exception` swallows the `APIException`, and `api_deployment_views.py:168` re-surfaces it). Both are documented, so this is a product call — but the two callers should not disagree, and the description is currently wrong. Detail in the held-back Mediums.

### Process note

The working tree at `/home/chandru/zipstuff/unstract` was switched off this PR's branch mid-review by a concurrent process, so early reads hit pre-PR code. That was detected, an isolated worktree was built at PR head, and **every citation below was re-verified**. All line numbers are PR-head numbers.

PR #2257, opened 2026-08-29
Sprint 4 [FIX] Backend fixes: UN-2900, UN-2902, UN-2953, UN-3038, UN-3133, UN-3176, UN-3333

hari-kuriakose:
## Sprint 4 — backend fixes

Six tickets investigated and fixed across the backend and SDK. Grouped onto one branch because they were worked as a single sprint-4 investigation pass; each commit is self-contained and can be reviewed independently.

| Commit | Ticket | Change |
| --- | --- | --- |
| `be9d7747` | UN-2953, UN-3038, UN-3333 | Adapter id validation, page billing, OCR threshold |
| `da198d0e` | UN-2902 | Honour the chunk limit in Keyword Table retrieval |
| `aa0df6f8` | UN-3176 | Apply the 15-significant-figure limit for values below 1 |
| `aa4f086a` | UN-3133 | Surface context-window overflow as its own error |
| `f45d4e0f` | UN-2900 | Flag variables that cannot resolve under single-pass extraction |
| `c830880f` | UN-2900 | Return the unresolvable-variable warning on prompt save too |
| `1f8a4a7c` | UN-3176 | Stop reporting BigQuery value errors as missing columns |

### Notes for reviewers

**UN-3133** — litellm's `ContextWindowExceededError` subclasses `BadRequestError` and therefore `openai.APIError`, so `parse_litellm_err` caught it in the generic branch and wrapped it as a plain "Error from &lt;provider&gt;" plus raw 400 text. It now maps ahead of the generic wrap, so the actionable message (reduce chunk size / limit pages / larger-context model) survives.

**UN-3176** — `bigquery.py` mapped *every* `google.api_core.exceptions.BadRequest` to `ColumnMissingException`, so a value-level failure told users to "make sure all the columns exist". Adds a `_is_value_error` discriminator; unrecognised shapes still fall through to `ColumnMissingException`, so this only narrows messages that were already wrong.

**UN-2900** — the ticket title points at the one thing that works: under single pass, `custom_data` resolves correctly while *every other* variable type silently fails. This surfaces the unresolvable ones as an authoring-time warning rather than a save-time block, since existing projects may already carry the combination and blocking would break them retroactively.

### Caveats

- Covers authoring only — an already-exported tool running single pass via API deployment still fails silently until re-saved.
- The OSS frontend does not yet render the `single_pass_unresolvable_variables` field; that lives on the C branch, which is inert without this one.


unstract/sdk1/src/unstract/sdk1/adapters/x2text/llm_whisperer_v2/src/static/json_schema.json
```diff
@@ -93,6 +97,14 @@
...
       "description": "States whether to reproduce horizontal lines in the document. Note: This parameter is not applicable if `mode` chosen is `native_text` and will not work if `mark_vertical_lines` is set to `false`."
     },
+    "word_confidence_threshold": {
+      "type": "number",
+      "title": "Word confidence threshold",
+      "default": 0.3,
+      "minimum": 0.0,
+      "maximum": 1.0,
+      "description": "Minimum OCR confidence a word must reach to be included in the extracted text. Lower this when words are dropped because the scan is faint or noisy. Note: This parameter is not applicable if `mode` chosen is `native_text`."
+    },
```
chandrasekharan-zipstack, line 107: ### [High] [Lens 1, 3, 7] — `word_confidence_threshold` is a dead UI control, and its default is 6× the service's

This setting renders in the adapter form, persists into `adapter_metadata`, and is **never sent to `/whisper`**. A user with a faint scan lowers it, saves, re-runs — output is byte-identical. UN-3333 is not delivered.

**Evidence:** a repo-wide grep for `word_confidence_threshold` returns exactly **one** hit — this schema line. `get_whisperer_params` (`.../src/helper.py:189-246`) builds an explicit allowlist with no `**config` passthrough; `WhispererConfig` / `WhispererDefaults` (`.../src/constants.py:60-127`) have no such key.

**Compounding it:** the service default is `0.05` (`unstract-llm-whisperer/backend/app/llm_whisperer_v2.py:144`, `providers/ocr/unstract_ocr_base.py:49`, `sample.env:65`). Once someone *does* wire this up, `default: 0.3` silently applies a **6× stricter** word floor to every existing V2 adapter, dropping low-confidence words with no migration.

Note this same file already carries a comment about `line_splitter_strategy` having had this exact defect (`helper.py:179-181`) — this is the second instance.

**Fix:** wire `WORD_CONFIDENCE_THRESHOLD` through `WhispererConfig` + `WhispererDefaults` + `get_whisperer_params` (gated off `native_text` as the description claims), set the default to `0.05`, and add a case to the params test file this PR already edits. Or drop the schema entry rather than ship a dead control.

Confidence: High.

unstract/sdk1/src/unstract/sdk1/exceptions.py
```diff
@@ -156,6 +174,26 @@ def parse_litellm_err(e: Exception, provider_name: str | None = None) -> SdkErro
...
+            status_code=status_code,
+        )
+        # Return early: the generic tail below would overwrite the actionable
+        # guidance with "Error from <provider>." and leave only the raw 400.
+        # The provider text is kept underneath it for support/debugging.
+        err.message = (
+            f"{ContextWindowExceededError.DEFAULT_MESSAGE}"
+            f"\n```\n{cleaned_message}\n```"
+        )
+        return err
```
chandrasekharan-zipstack, line 195: ### [High] [Lens 1, 3] — The UN-3133 fix cannot fire on the LLM completion path it describes

A prompt run whose document + prompt overflow the context window still shows `Error from LLM adapter '<x>': <raw 400>` — the exact symptom the ticket describes. This new branch only reaches embedding and indexing calls.

**Evidence:** `parse_litellm_err` has three call sites repo-wide — `embedding.py:119,149,165,179,195` and `workers/executor/executors/index.py:182`. `llm.py:18` imports only `LLMError, SdkError, strip_litellm_prefix`; the completion handler at `llm.py:593-611` does `except Exception as e: ... raise LLMError(message=f"Error from LLM adapter '{...}': {strip_litellm_prefix(str(e))}")`, never consulting `parse_litellm_err`.

**Fix:** call `parse_litellm_err` from `llm.py`'s generic handler before the `LLMError` fallback, or replicate the `isinstance` check there.

Confidence: High on reachability. See open question #1 in the review body — if UN-3133 is actually about indexing/embedding, this resolves and only the message wording is off.

unstract/connectors/src/unstract/connectors/databases/bigquery/bigquery.py
```diff
@@ -106,13 +107,14 @@ def _sanitize_for_bigquery(data: Any) -> Any:
...
-            safe_decimals = max(0, 15 - magnitude)
-            return float(f"{data:.{safe_decimals}f}")
+            # Limit total significant figures to 15 for IEEE 754 compatibility.
+            # BigQuery PARSE_JSON requires values that round-trip cleanly.
+            #
+            # UN-3176: `g` asks for significant figures directly. The previous
+            # form asked for a DECIMAL-place count derived from the magnitude,
+            # which is not the same quantity and drifted from 15 sig figs at
+            # both ends of the range.
+            return float(f"{data:.15g}")
```
chandrasekharan-zipstack, line 117: ### [High] [Lens 3, 5] — `%.15g` destroys precision on ≥1e15 floats that the old code preserved exactly

A 16–17-significant-digit float that is *exactly* representable as an IEEE-754 double is now silently rewritten before landing in the customer's BigQuery table. No error, no log — the warehouse just holds a wrong number. The stated contract is "values that round-trip cleanly"; these values already did.

**Both implementations run side by side:**

```
1760509016282637.0    old=1760509016282637.0    new=1760509016282640.0   old_exact=True  new_exact=False
1234567890123456.0    old=1234567890123456.0    new=1234567890123460.0   old_exact=True  new_exact=False
9.99999999999999e-05  old=0.0001                new=9.99999999999999e-05 old_exact=False new_exact=True
```

Only the third line is the bug UN-3176 describes ("values below 1"). The first two are a **new regression** — and `test_large_magnitude_is_limited_to_15_significant_figures` (`test_bigquery_db.py:206-211`) locks it in as desired behaviour.

**Fix:** skip the cap when the value already round-trips — e.g. return `data` unchanged when `data.is_integer() and abs(data) < 2**53`, or when `float(f"{data:.17g}") == data` and the 15g form differs. Update the test accordingly.

Confidence: High on the numeric behaviour (independently reproduced). Medium on frequency — the sanitizer is only reached for floats nested in dict/list payloads (`:270,287,410`). If tool metadata carries epoch-microsecond floats or large numeric IDs, this is Critical.

unstract/sdk1/src/unstract/sdk1/x2txt.py
```diff
@@ -118,6 +118,55 @@ def process(
...
+                        continue
+                selected.update(range(max(start, 1), min(end, total_pages) + 1))
+            else:
+                try:
+                    page = int(part)
+                except ValueError:
+                    continue
+                if 1 <= page <= total_pages:
+                    selected.add(page)
+        return len(selected)
```
chandrasekharan-zipstack, line 155: ### [High] [Lens 3, 16] — A partially-parseable `pages_to_extract` under-bills, contradicting the docstring's explicit guarantee

The docstring at `:160-161` states usage is *"never under-reported by a malformed value."* That holds only when **every** component fails. One bad component is dropped by `continue` while previously-accumulated pages survive.

**Confirmed against a 10-page document:**

```
"3-x,5"       -> billed 1  of 10
"1-3,junk,7"  -> billed 4  of 10
"2-4,1x"      -> billed 3  of 10
```

Silent revenue loss — and a maintainer trusting the docstring will not add validation.

`:145-146`'s `except ValueError: continue` abandons the range component but leaves `selected` populated; `:168`'s `return selected or page_count` only rescues the all-zero case.

**Fix:** track a `parse_failed` flag set by any of the four `continue` paths and return `page_count` when set; then correct the docstring.

Confidence: High. See open question #2 in the review body — if `push_page_usage_data` feeds a charged meter, this is arguably Critical.

backend/tool_instance_v2/tool_instance_helper.py
```diff
@@ -486,33 +486,57 @@ def validate_adapter_permissions(
...
-                adapter_id = tool_meta[llm.adapter_id]
+                adapter_id = tool_meta.get(llm.adapter_id)
             elif llm.is_enabled:
-                adapter_id = tool_meta[AdapterPropertyKey.DEFAULT_LLM_ADAPTER_ID]
+                adapter_id = tool_meta.get(AdapterPropertyKey.DEFAULT_LLM_ADAPTER_ID)
 
-            adapter_ids.add(adapter_id)
+            # UN-2953: a tool instance may carry "" for an adapter id.
+            # Adding it made the schema validator compare "" against the
+            # UUID enum and raise, once per validation pass.
```
chandrasekharan-zipstack, line 497: ### [High] [Lens 16, 15] — This comment (repeated ×4, verbatim) names a mechanism that does not exist on this path, and hides the bugs actually fixed

Also at `:509-511`, `:523-525`, `:535-537`.

The comment says adding `""` *"made the schema validator compare `""` against the UUID enum and raise."* `adapter_ids` is a function-local set whose **only** consumer is `validate_adapter_access` (`:553`) → `AdapterInstance.objects.filter(id__in=adapter_ids)`. No schema validator, no enum on that path. The only validator in this file (`:451`) validates `tool_meta` — a different input — and runs *after* this function returns. A maintainer will go read the JSON schema and find nothing.

The real failure was a Django UUID-coercion error (`AdapterInstance.id` is a `UUIDField`, `backend/adapter_processor_v2/models.py:63`), or the old `KeyError`.

**And the comment never mentions the two other bugs this diff silently fixes:** the old code left `adapter_id` unassigned when `is_enabled` was `False`, so `adapter_ids.add(adapter_id)` re-added the previous iteration's id (leaking across all four loops) or raised `UnboundLocalError` on the first. **The new `adapter_id = None` initialiser is the load-bearing change and is undocumented.**

**Fix:** one accurate comment above the loop group — `# Skip unset adapter ids: "" is not a valid UUID and breaks the id__in filter.` Hoist the four near-identical loops into a helper.

Confidence: High. Independently reached by three agents and a manual trace.

backend/tool_instance_v2/tool_instance_helper.py
```diff
@@ -486,33 +486,57 @@ def validate_adapter_permissions(
...
+                adapter_id = tool_meta.get(AdapterPropertyKey.DEFAULT_X2TEXT_ADAPTER_ID)
 
-            adapter_ids.add(adapter_id)
+            # UN-2953: a tool instance may carry "" for an adapter id.
+            # Adding it made the schema validator compare "" against the
+            # UUID enum and raise, once per validation pass.
+            if adapter_id:
+                adapter_ids.add(adapter_id)
 
         ToolInstanceHelper.validate_adapter_access(user=user, adapter_ids=adapter_ids)
```
chandrasekharan-zipstack, line 541: ### [High] [Lens 13 — Testing] — This permission-check rewrite ships with zero tests

Every `tool_meta[...]` became `tool_meta.get(...)` plus `if adapter_id:`, changing raise-semantics to skip-semantics on the sole input to an authorization check (`validate_adapter_access`, `:553`). Nothing exercises it.

`grep -rn "validate_adapter_permissions" --include=*.py .` returns only the definition (`:482`) and its one caller (`:439`) — no test file. The app's only suite, `backend/tool_instance_v2/tests/test_challenge_llm_seed_overlay.py`, touches neither adapters nor permissions.

**Fix — three cases against a stubbed `ToolProcessor.get_tool_by_uid`:**
1. `""` for an enabled adapter → no exception, `""` excluded from the set;
2. a real inaccessible adapter id → `PermissionDenied` still raised;
3. key **missing** entirely → pin the intended contract explicitly (this is the case the `.get()` widening newly swallows).

Confidence: High.

chandrasekharan-zipstack (changes requested):
## Standardized review — High-severity findings only

**Verdict: `BLOCK`** — Critical 0 · High 6 · Medium 11 · Low 8 · Lenses 16/16.

This review posts **only the 6 High findings**. The 11 Medium and 8 Low are held back to keep the thread actionable; happy to post them on request.

**Why BLOCK:** three of the seven tickets do not deliver their stated fix (UN-3333, UN-3133, and UN-2953's documented mechanism); the BigQuery change introduces a silent data-corruption regression outside its ticket's scope **that the PR's own test enshrines as intended**; and the billing change silently under-reports on a class of input its docstring promises to protect against. Each is individually merge-blocking; together they mean the branch does not do what the description says.

**Tests:** both new suites were executed at PR head and pass — `unstract/connectors` 12 passed (full unit suite 71), `unstract/sdk1` 26 passed (full unit suite 579). Both are wired into CI via `tests/groups.yaml`. Mutation-checked: deleting the `errors`-payload loop in `_is_value_error` fails exactly its dedicated test; deleting the `_get_billable_page_count` call in `push_usage_details` fails exactly its dedicated test. All four numeric assertions independently verified.

---

### On the authorization change — cleared, no privilege escalation

Since the permission rewrite is the riskiest-looking part of the diff, I traced it end to end rather than inferring:

1. `adapter_ids` feeds exactly one consumer — `AdapterInstance.objects.filter(id__in=adapter_ids)` (`tool_instance_helper.py:544`). `AdapterInstance.id` is a `UUIDField` primary key (`backend/adapter_processor_v2/models.py:63`), so `""` and `None` can never match a row. They contributed **zero** rows to the permission loop; dropping them removes no check. Pre-change, `""` did not fail closed *by permission* — it failed closed by **crashing** the ORM on UUID coercion.
2. Every truthy id still flows through unchanged, so a real adapter UUID the user cannot access still raises `PermissionDenied` exactly as before.
3. The only mechanism by which an absent id could later resolve to a *real* adapter is `DefaultsGeneratingValidator` (`unstract/sdk1/src/unstract/sdk1/tool/validator.py:26-33`), which `setdefault`s schema defaults into `tool_meta` at `:451` — **after** the permission check. Prompt Studio specs do carry a real adapter UUID as a default (`prompt_studio_registry_helper.py:87`). But `set_defaults` yields into `validate_properties` in the same pass, and that enum is built from `AdapterProcessor.get_adapters_by_type(..., user=user)` → `AdapterInstance.objects.for_user(user)`. An injected default the user cannot access fails enum validation. Fail-closed.
4. `setdefault` does not overwrite a present-but-empty `""`, so that value is never upgraded to a real id.

What the change *does* do is convert two fail-closed crashes into silent passes — a robustness/observability regression, rated Medium (held back with the rest).

---

### Open questions — the two that would move severities

1. **Is UN-3133 about the indexing/embedding path or the LLM completion path?** `parse_litellm_err` is only reachable from `embedding.py` and `index.py`. If the ticket's repro is a prompt run, H2 stands and the fix belongs in `llm.py`. If it is an indexing failure, H2 resolves and only the message wording ("the document and prompt together") needs adjusting.
2. **Does `Audit.push_page_usage_data` feed a charged meter or only telemetry?** If it is billing, H4 is revenue-affecting and arguably Critical.
3. **What was UN-2953's actual reported symptom?** If it was the schema validator rejecting `""` against the UUID enum, that was already fixed by #2209 (`66a3363fb`), whose comment in `prompt_studio_registry_helper.py:67-72` narrates exactly it — in which case this ticket may already be closed, and this diff is really three undocumented bug fixes.
4. **Was the `%.15g` change intended to affect values ≥ 1e15?** The ticket title says "for values below 1".
5. **Is `word_confidence_threshold` meant to ship inert here**, with the wiring on another branch (as the description notes for `single_pass_unresolvable_variables`)? The description does not say so for UN-3333.
6. **Was the keyword-table retrieval change evaluated against extraction quality?** It silently changes answers for every existing user of that strategy (details in the held-back Mediums).

### Assumptions that, if wrong, change severities

- Floats ≥ 1e15 reach `_sanitize_for_bigquery` in real tool-output metadata. If they never do, H3 drops to Low.
- The LLMWhisperer service defaults read from `unstract-llm-whisperer` at its current HEAD match what is deployed.
- Also noted: the PR description says "**Six** tickets" while the title and its own table list **seven**; and the three highest-risk items (UN-3038 billing, UN-2953 permissions, UN-3333 OCR default) get no "Notes for reviewers" entry while the three lower-risk ones do.

PR #2259, opened 2026-08-29
UN-3315 [FIX] Honour shared_to_org for Prompt Studio prompt edits

hari-kuriakose:
## UN-3315 — prompts stayed read-only in org-shared Prompt Studio projects

"Share with everyone" sets `shared_to_org` on the parent `CustomTool`. `IsOwnerOrSharedUserOrSharedToOrg` already honoured that flag, so the project was **visible** to the org — but `PromptAcesssToUser` did not check it, so editing a prompt inside that project stayed blocked for everyone except the owner.

One file, +9/−2: adds the `shared_to_org` check alongside the existing owner / direct-viewer / group-share / org-admin paths.

### Reviewer note — this widens access

Please sanity-check the intent: after this change, **any org member can edit prompts in a project shared with the whole org**. That is what "share with everyone" reads as, and it matches the visibility rule already in place, but it is a genuine broadening rather than a pure bug fix — worth a second opinion before merge.

The check reads `tool.shared_to_org` directly, matching the sibling gate at `permissions/permission.py:248`. (An earlier revision used `getattr(tool, "shared_to_org", False)` and this section claimed it "degrades safely if the attribute is absent" — that was inaccurate: `shared_to_org` is a non-null model field, so the default could never apply.)

**Update:** deletion is no longer included in the widened access. `get_permissions` on `ToolStudioPromptView` now routes `destroy` to `IsPromptParentToolOwner`, so an org-shared member can view and edit prompts but not delete them, matching the parent `CustomTool.destroy` being `IsOwner`-gated. Reads and edits are unchanged by that split.

**Known gap, not closed here:** the bulk `sync_prompts` route on `PromptStudioCoreView` still lets an org-shared member remove every prompt in the project (pre-existing, UN-3318). Tracked separately.


hari-kuriakose:
### Follow-up raised: UN-4054 — `tool_id` remains client-writable

Tracking the residual half of the reparenting Critical found by the cold read on this PR: **https://zipstack.atlassian.net/browse/UN-4054** (Medium).

**What this PR closes.** `636cedfa4` gates a reparent on the parent it *leaves*: in `ToolStudioPromptView.update`, if `tool_id` is present and differs from the stored value (`null` counts as a move), it requires `IsPromptParentToolOwner` against the **existing** parent. Moving a prompt out of a project is a deletion from that project's side, so it now needs what `destroy` needs, and it refuses visibly with a 403.

**What stays open.** `tool_id` is still a writable serializer field, so any other writer of that FK is ungated. The complete fix is one line — `read_only_fields = ["tool_id"]` — and it was **deliberately declined**, not missed: DRF *ignores* a read-only field rather than rejecting it, so a client legitimately sending `tool_id` on PATCH would get a `200` with no effect. A silent no-op was judged worse than a visible refusal, and it would change a published API contract.

**Note for whoever picks UN-4054 up:** a *blanket* `read_only_fields` would silently orphan every newly created prompt — `create_prompt` shares this serializer and takes `tool_id` from the request body, with nothing injecting it server-side. Any fix must be conditional on the update path. That analysis is preserved with the declined patch and summarised in the ticket.

<sub>Follow-up from the standardized review of this PR.</sub>

backend/prompt_studio/prompt_studio_v2/tests/test_prompt_permission_guards.py
```diff
@@ -0,0 +1,538 @@
...
+            patch(
+                "rest_framework.viewsets.ModelViewSet.update", return_value="updated"
+            ) as parent_update,
+        ):
+            gate.return_value.has_object_permission.return_value = owner_allows
+            try:
+                result = view.update(request)
+            except PermissionDenied:
+                return "denied", parent_update
+        return result, parent_update
```
chandrasekharan-zipstack, line 480: ### [High] [Lens 13 — Testing] — These reparent tests are tautological on the one property `update()` exists to enforce (mutant survived)

The production docstring (`views.py:69-72`) makes the security-critical claim that the gate must run against the **existing** parent. Because this test patches the permission class itself (`patch(f"{self.MODULE}.IsPromptParentToolOwner") as gate`, L470) and hard-codes its return (`gate.return_value.has_object_permission.return_value = owner_allows`, L475), the object passed to `has_object_permission` is never asserted.

A refactor gating the *destination* instead — reopening the reparent-then-delete bypass this PR closes — passes all five tests.

**Mutation-verified:** replacing `has_object_permission(request, self, instance)` with `has_object_permission(request, self, object())` at `views.py:88-90` → **27 passed**.

**Fix:** patch the collaborators, not the class, as `TestPromptDeletionIsOwnerOnly` already does at L122. Or minimally `assert gate.return_value.has_object_permission.call_args.args[2] is instance`.

Confidence: High.

backend/prompt_studio/prompt_studio_v2/views.py
```diff
@@ -32,33 +37,124 @@ class ToolStudioPromptView(viewsets.ModelViewSet):
...
             Response: The HTTP response indicating the status of the reorder operation.
         """
+        prompt_id = request.data.get(ToolStudioPromptKeys.PROMPT_ID)
+        if not prompt_id:
+            raise ValidationError({ToolStudioPromptKeys.PROMPT_ID: "This is required."})
+        # ToolStudioPrompt carries no organization of its own -- it is a plain
+        # BaseModel -- so scope through the parent tool, whose for_user()
+        # queryset is org-bound. A cross-org or invisible id 404s here rather
+        # than reaching the permission check. select_related because the
+        # permission class immediately dereferences the tool_id FK, and the
```
chandrasekharan-zipstack, line 149: ### [High] [Lens 13 — Testing] — `reorder_prompts` cross-org scoping has zero coverage (mutant survived)

The comment here claims "A cross-org or invisible id 404s here rather than reaching the permission check." All three reorder tests (`test_prompt_permission_guards.py:319, 344, 379`) patch `get_object_or_404` with a fixed return and never inspect its first argument, so the scoping queryset is never asserted.

**Mutation-verified:** replacing the scoped queryset with `ToolStudioPrompt.objects.all()` → **27 passed**.

**Fix:** capture the lookup mock; assert `custom_tool.objects.for_user.assert_called_once_with(SHARED_MEMBER)` and assert on `lookup.call_args.args[0]`, mirroring `TestMakeProfileDefaultIsScopedToItsTool` at L241/248.

Confidence: High.

backend/prompt_studio/prompt_studio_v2/tests/test_prompt_permission_guards.py
```diff
@@ -0,0 +1,538 @@
...
+
+Mechanism 3 is why these tests import the real modules rather than extracting
+method bodies with ``tests_common.source_extraction``, as the sibling
+``test_registry_tool_delete_guards.py`` does. That technique's own docstring
+records the blind spot: bodies are ``exec``-ed out of context, so *unreachable*
+code is indistinguishable from wired code -- and "the hook is never reached" is
+precisely the bug mechanism 3 fixes.
+
+No database is touched. ``get_permissions()`` is pure, and the collaborators
+that would hit the ORM are patched.
```
chandrasekharan-zipstack, line 32: ### [High] [Lens 13, 4] — No DB-backed test that a different-org member cannot reach a `shared_to_org=True` tool

This is the highest-consequence property of the whole change, and "No database is touched" is why it cannot be tested here.

`permission.py:62-63` returns `True` on `tool.shared_to_org` **before any other check and with no org predicate**. The only thing keeping a foreign-org user out is `CustomToolModelManager.get_queryset` inheriting `DefaultOrganizationManagerMixin`'s `UserContext.get_organization()` filter (`backend/utils/models/organization_mixin.py:27-30`). A `SimpleNamespace` tool (`_tool()`, L65-66, no `organization` attribute) and a patched `CustomTool` (L320/347/380/400) cannot exercise that.

**If the org filter ever regresses, every test in this file still passes while `shared_to_org=True` becomes a cross-tenant read+edit grant.** `for_user` also returns `self.all()` for org admins and service accounts (`prompt_studio_core_v2/models.py:33-37`) — a second uncovered path.

**Fix — precedented and cheap.** `backend/workflow_manager/execution/tests/test_shared_execution_access.py:179` already has `test_org_wide_share_does_not_cross_organizations` on `GroupSharingTestBase` (`backend/tenant_account_v2/tests.py:69`). Add a sibling with an Org B tool `shared_to_org=True` plus a prompt; assert an Org A user gets an empty `get_queryset()` and a 404 from the reorder lookup. `conftest.py:27-45` auto-marks it `integration` — no manifest edit needed.

Confidence: High.

backend/prompt_studio/permission.py
```diff
@@ -12,18 +12,55 @@
...
-    (``ResourceGroupShare`` on the parent tool), or are an org admin
-    (org-wide admin override, UN-3479).
+    (``ResourceGroupShare`` on the parent tool), reach it because the parent
+    tool is shared with the whole org (``shared_to_org``, UN-3315), or are an
+    org admin (org-wide admin override, UN-3479).
+
+    Deliberately broader than the workflow rule stated in
+    ``permissions.permission.is_workflow_mutator`` ("shared access grants read
+    only, never mutate"): a Prompt Studio share confers *edit* rights on the
+    project's prompts, matching ``CustomToolViewSet``, which already routes
```
chandrasekharan-zipstack, line 26: ### [High] [Lens 16 — Doc accuracy] — Three comments cite `CustomToolViewSet`, a class that does not exist

Also at `permission.py:84` and `prompt_studio_v2/views.py:44`.

All three carry the entire justification for this PR's central decision — "edits honour sharing, deletes do not, because the parent tool viewset does the same." A maintainer auditing that premise greps `CustomToolViewSet`, finds only these three comments referencing each other, and cannot verify it. One who concludes the class was deleted may decide the parity argument no longer holds and "re-align" the gate.

`grep -rn "CustomToolViewSet" backend --include="*.py"` returns exactly those three comment lines. The real class is `PromptStudioCoreView` (`prompt_studio_core_v2/views.py:131`) — the class whose `get_permissions` this PR edits at `:157-170`.

**Fix:** replace all three with `PromptStudioCoreView`. The substantive claims are correct once the name is fixed.

Confidence: High.

backend/prompt_studio/permission.py
```diff
@@ -33,6 +70,35 @@ def has_object_permission(self, request: Request, view: APIView, obj: Any) -> bo
...
+    cases is how these gates drift apart.
+
+    Exists because the parent ``CustomTool``'s own ``destroy`` is owner-only
+    (``IsOwner`` in ``CustomToolViewSet.get_permissions``). Without this,
+    UN-3315's org-wide share would let any org member delete every prompt
+    inside a project they cannot themselves delete.
+
+    ``tool_id`` is nullable (``SET_NULL``), so an orphaned prompt whose parent
+    tool was deleted falls back to the org-admin check -- it has no owner to
+    inherit from.
```
chandrasekharan-zipstack, line 90: ### [High] [Lens 16 — Doc accuracy] — This docstring claims parity with `IsParentToolOwner` and gives a false reason for its orphan behaviour

The docstring says it "Mirrors `permissions.permission.IsParentToolOwner` … **but** reads the parent through … `tool_id` rather than `prompt_studio_tool`", framing the FK name as the sole divergence. It is not — the two differ in **authorization outcome** for orphan rows:

- `IsParentToolOwner` (`backend/permissions/permission.py:163-164`) does `owner_resource = obj.prompt_studio_tool or obj` and documents the self-fallback as deliberate legacy behaviour.
- This class does `if tool is not None and _is_resource_owner(...)`, then drops to org-admin.

It then justifies itself with "it has no owner to inherit from", which is false: `ToolStudioPrompt` has `created_by` (`prompt_studio_v2/models.py:121`).

**Why it matters concretely:** commit `7d4f2813d` explicitly contemplates extracting a base class over these clones, and H8 independently recommends exactly that. A maintainer doing that consolidation on the strength of "mirrors" silently flips orphan-prompt authorization.

**Fix:** state the divergence instead of denying it; drop "it has no owner to inherit from".

Confidence: High.

backend/prompt_studio/permission.py
```diff
@@ -33,6 +70,35 @@ def has_object_permission(self, request: Request, view: APIView, obj: Any) -> bo
...
+    """
+
+    def has_object_permission(self, request: Request, view: APIView, obj: Any) -> bool:
+        if getattr(request.user, "is_service_account", False):
+            return True
+        tool = obj.tool_id
+        if tool is not None and _is_resource_owner(request.user, tool):
+            return True
+        return OrganizationMemberService.is_user_organization_admin(request.user)
```
chandrasekharan-zipstack, line 99: ### [High] [Lens 2 — Architectural fit] — Fifth near-clone parent-owner permission class, justified by an anti-drift argument the existing four already falsify

The docstring argues the clone exists so a shared class won't "drift apart". But the four pre-existing clones **have already drifted**, on three separate axes — so a future security fix lands in only some of them:

- **null-parent:** `IsParentToolOwner` / `IsRegistryToolOwner` fall back to the object's own ownership (`obj.prompt_studio_tool or obj`, `obj.custom_tool or obj`); `IsPromptParentToolOwner` denies (org-admin only); `IsParentWorkflowOwner` does not guard null at all.
- **admin check:** `IsParentWorkflowOwner` / `IsParentToolOwner` / `IsParentDeploymentOwner` use the request-caching `_is_organization_admin(request)`; both classes in `prompt_studio/permission.py` — including this new one — call `OrganizationMemberService.is_user_organization_admin(request.user)` directly, **bypassing the cache**.
- **service account:** the shared classes use `_is_service_account(request)`; this one inlines `getattr(request.user, "is_service_account", False)`.

The count, for the record: `permissions/permission.py:138` (`IsParentWorkflowOwner`), `:151` (`IsParentToolOwner`), `:171` (`IsParentDeploymentOwner`), `prompt_studio/permission.py:102` (`IsRegistryToolOwner`), and this one. Plus `IsOwner:104`, `IsOwnerOrSharedUser:225`, `IsOwnerOrSharedUserOrSharedToOrg:239` sharing the same body shape. The only stated blocker to sharing is an attribute name.

**Fix:** parameterise — `class IsParentResourceOwner(BasePermission): parent_attr: str; fallback_to_self: bool = True`, then a 3-line subclass. Smaller surface than the fifth copy. If the clone stays, at minimum switch `:94` and `:99` to `_is_service_account(request)` / `_is_organization_admin(request)`.

Confidence: High.

backend/prompt_studio/prompt_studio_v2/views.py
```diff
@@ -32,33 +37,124 @@ class ToolStudioPromptView(viewsets.ModelViewSet):
...
+    def get_permissions(self) -> list[Any]:
+        # Reads and edits honour project sharing (UN-3315), enforced on two
+        # levels: get_queryset scopes every action to the user's reachable
+        # tools, and these classes gate the object. Deleting a prompt requires
+        # ownership of the parent tool, matching CustomToolViewSet, whose own
+        # `destroy` is IsOwner-gated. The bulk `sync_prompts` route there is
+        # IsOwner-gated too. API-key gap: see PromptAcesssToUser.
+        if self.action == "destroy":
+            return [IsPromptParentToolOwner()]
+        return [PromptAcesssToUser()]
```
chandrasekharan-zipstack, line 49: ### [High] [Lens 2 — Architectural fit] — This `get_permissions` routing fails open, and prompt authorization is now spread across four mechanisms

The mapping is `destroy` → owner-only, **everything else** → the share-permissive `PromptAcesssToUser`. Any action added later — a bulk-delete, an "unlink" collection POST — is silently authorized with no code change and no test failure.

Compounding it: neither class defines `has_permission`, so `BasePermission.has_permission` returns `True` and any `detail=False` action gets **no** check unless the author remembers an explicit `check_object_permissions`. Not hypothetical — that is exactly the `reorder_prompts` hole this PR fixes, and exactly H1. There is no framework backstop either: `DEFAULT_PERMISSION_CLASSES` is `[]` (`backend/backend/settings/base.py:640`, commented `# TODO: Update once auth is figured`).

The mapping is exhaustive over the currently routed set (`urls.py:6-25`), so there is **no live bug** — the finding is the fail-open *direction*. `PromptStudioCoreView.get_permissions` (`prompt_studio_core_v2/views.py:157-170`) has the same shape and consequence: `import_project` (`detail=False`) resolves to `IsOwnerOrSharedUserOrSharedToOrg`, never consulted.

Authorization for one resource now lives in four places: this dispatch, `get_queryset()` scoping (`views.py:110-112`), an inline class instantiation in `update()` (`views.py:88`), and an explicit `check_object_permissions` in `reorder_prompts` (`views.py:157`).

**Fix:** invert the default — an explicit per-action dict where an unmapped action raises rather than passes; or default to the restrictive class. Separately, give both classes an explicit `has_permission` returning `request.user.is_authenticated`.

Confidence: High.

chandrasekharan-zipstack (changes requested):
## Standardized review — High-severity findings only

**Verdict: `REQUEST CHANGES`** — Critical 0 · High 9 · Medium 14 · Low 8 · Lenses 16/16.

This review posts **only the 9 High findings**. The 14 Medium and 8 Low are held back to keep the thread actionable; happy to post them on request.

**Test execution (verified, not assumed):** the suite runs — **27 passed in 0.68s** under the CI env (`tests/groups.yaml:104-113`, `unit-backend`), and CI does collect it. Bare `pytest` from `backend/` fails collection because `backend/test.env` is absent — pre-existing ergonomics, not this PR.

---

### On your reviewer note — the widening is right, and the description undersells it

You asked for a second opinion on whether org-wide prompt edit matches intent. It does, and the case is stronger than the body makes it: **prompts were the anomaly, not the fix.**

1. `shared_to_org` **already** conferred write on the parent object — `PromptStudioCoreView.get_permissions` (`prompt_studio_core_v2/views.py:157-170`) routes `update`/`partial_update` through `IsOwnerOrSharedUserOrSharedToOrg`. An org-shared member could already edit the project itself. Prompts being read-only inside a project you can rename was the inconsistency.
2. `create_prompt` is **completely ungated today** (see H1 below). "Org members can write prompts" is not new; this PR makes the coherent subset of it work.
3. The delete carve-out is the right line, correctly drawn: `destroy` → `IsPromptParentToolOwner`, `sync_prompts` → `IsOwner`.

So this is not a genuine broadening of the trust model — it closes a gap between two halves of a share that were already inconsistent. Saying it that way in the description removes the reviewer anxiety you were anticipating.

**Blast radius.** Every org-shared project in every tenant, immediately on deploy; no flag, rollback is a revert. But the *incremental* radius is narrower than it looks. What changes: org-shared members gain `update`/`partial_update`/`retrieve`/`list`/`reorder` on existing prompt rows. What they do **not** gain: `destroy` — and they *lose* `sync_prompts`, a tightening (see H6).

---

### H1 — [High] [Lens 2, 4] `create_prompt` has the exact collection-POST hole this PR fixes for `reorder_prompts`, left wide open

*Posted here rather than inline: the code is outside this PR's diff.*

- **Location:** `backend/prompt_studio/prompt_studio_core_v2/views.py:955-968` (route `urls.py:100-104`; gate `views.py:157-170`)
- **Failure mode:** `@action(detail=True)` but never calls `self.get_object()`. `IsOwnerOrSharedUserOrSharedToOrg` defines only `has_object_permission`, so `BasePermission.has_permission` returns `True` and nothing is checked. The `<uuid:pk>` is bound and never used; the parent comes entirely from `request.data["tool_id"]`, and `ToolStudioPromptSerializer` is `fields = "__all__"`. Any authenticated org member can POST with their own tool's UUID in the path and any project's UUID in the body, injecting a prompt into a project they have no share on. It then runs against the owner's LLM profile and documents.
- `create_profile_manager` (`:971-987`) is the same pattern with a partial mitigation — it calls `self.get_object()` **only when the body omits `prompt_studio_tool`**, so a body that supplies it bypasses the check too.
- **Evidence:** this PR's own comment one function over diagnoses this mechanism — `prompt_studio_v2/views.py:128-134`. No `perform_create` override on `PromptStudioCoreView`. Cross-org is blocked (DRF resolves the FK via the org-filtered `CustomTool._default_manager`); same-org is not.
- **Fix:** `tool = self.get_object()` at the top of `create_prompt`, then force `serializer.validated_data["tool_id"] = tool` before `perform_create`. Make `create_profile_manager`'s `get_object()` unconditional.
- **Confidence:** High. **Pre-existing, not introduced here** — which is why High and not Critical: merging does not cause it. But it sits in a file this PR edits, it is the same bug class the PR closes elsewhere, and it defeats the write-side model the PR's docstrings assert. The PR carefully gates reparenting a prompt *out* of a project while moving one *in*, to a project you cannot see, is free. If intra-org project privacy is a security boundary — and this PR presupposes it is — this deserves its own ticket at Critical priority.

### H6 — [High] [Lens 1, 16] The PR description describes a different change and inverts a security-relevant claim

*Posted here rather than inline: the finding is about the PR body.*

- **Failure mode:** the body says "**One file, +9/−2**"; the diff is 5 files, +736/−13. More seriously it states *"Known gap, not closed here: the bulk `sync_prompts` route … still lets an org-shared member remove every prompt in the project (pre-existing, UN-3318). Tracked separately."* **The code closes it** — `"sync_prompts"` is now in the `IsOwner()` list (`prompt_studio_core_v2/views.py:166`), pinned by `test_sync_prompts_resolves_the_owner_permission`.
- `sync_prompts` has no OSS or cloud frontend consumer, so its callers are API/SDK clients holding user tokens: **a shared-project caller that works today gets a 403 after merge**, with nothing in the description a release manager could turn into a changelog entry.
- Three further behaviour changes are absent from the description entirely: the `get_queryset` rewrite (previously-visible prompts now 404), the `tool_id` reparent 403, and the `make_profile_default` 404/400 changes. A reviewer working from the stated scope reviews none of them.
- **Fix:** rewrite the body to match the diff; delete the UN-3318 paragraph; call out the `sync_prompts` 403 as a breaking change for token-authenticated API clients.
- **Confidence:** High.

---

### Open questions

1. Do you agree the widening story should be reframed per the section above? It answers your own reviewer note.
2. Is `create_prompt` (H1) in scope here, or a follow-up ticket? If follow-up, it needs a ticket **before** merge — it is a bigger hole than the one being closed and it undercuts the reparent guard you built.
3. Was the `sync_prompts` → `IsOwner` change intended to close UN-3318? The code does; the description says it does not. Should UN-3318 be closed?
4. Is a 403 acceptable for token-authenticated API clients calling `sync_prompts` on a shared project? Needs a changelog line.
5. **Possible test flakiness:** on 2 of ~25 runs, `test_profile_from_another_tool_is_not_promotable` and the two `TestPromptListIsScopedToReachableTools` assertions failed with "mock called more times than asserted"; not reproducible after clearing `__pycache__` and ~20 further runs. No `pytest-randomly` installed, so ordering is deterministic. Low confidence it is real — worth a few dozen CI runs under the `unit-backend` xdist group.

### Assumption this review rests on

**Intra-org project privacy is a security boundary.** The PR presupposes it — that is what the delete carve-out protects. H1 is rated on that basis. If the team's position is that any org member may touch any project, H1 drops to Low and the `update()` reparent guard is largely unnecessary.

PR #2264, opened 2026-08-31
UN-3972 [DEV] Cut dashboard cron DB time by indexing workflow_file_execution on (status, created_at)

kirtimanmishrazipstack:
## What

One database index on processed files, covering status and date together.

The second index the ticket also asks for is **not** here — dropped in comment 45015.

## Why

The dashboard cron costs ~3,580 s of database time per 6 h on prod. Two queries are **83% of
that**, and both look up processed files by date and status.

Nothing organises the data that way, so the database reads all 1,278,885 execution rows on every
call. This index lets it start from the files in the date range instead.

## How

`SeparateDatabaseAndState`: `RunSQL` builds the index `CONCURRENTLY IF NOT EXISTS`, a state-only
`AddIndex` keeps Django's model state in step, and `atomic = False` because `CONCURRENTLY` cannot
run in a transaction. A `RAISE EXCEPTION` guard fails loudly on a leftover INVALID index, which
`IF NOT EXISTS` would otherwise keep while Django recorded the migration as applied.

> Do not regenerate with `makemigrations` — it emits a plain `AddIndex`, holding a `SHARE` lock
> for the whole build on a 3.4 GB table.

No. Nothing but an index is added — no behaviour changes — and it is built without blocking
writes. The cost is slightly slower writes on that table, against 1,679 ms of query time saved
per 6 h.

## Database Migrations

`file_execution/0007_wfe_status_created_idx.py` — builds the index without locking writes, and
reverses cleanly. Optionally build it by hand first; the deploy then skips it:

```sql
CREATE INDEX CONCURRENTLY IF NOT EXISTS wfe_status_created_idx
  ON workflow_file_execution (status, created_at);
```

## Migration Order

The three UN-3883 PRs stack on the same integration branch. **Merge in this order.**

| Order | PR | Migration | Depends on | This PR? |
| --- | --- | --- | --- | --- |
| 1 | #2255 · UN-3973 | `dashboard_metrics/0005_add_reconciliation_task` | `0004_pg_periodic_tasks` (UN-3445, already on `main`) | No — **merge before** this PR |
| 2 | #2264 · UN-3972 | `file_execution/0007_wfe_status_created_idx` | `file_execution/0006_…` | **Yes — this PR** |
| 3 | #2265 · UN-3974 | `dashboard_metrics/0006_split_aggregation_schedule` | `0005_add_reconciliation_task` (#2255) | No — merge after this PR |

#2264's migration is in a different app and has no interaction with the other two. The one
hard dependency is **#2265 on #2255** — merged out of order, #2265 fails at graph build:

```
NodeNotFoundError: Migration dashboard_metrics.0006_split_aggregation_schedule
dependencies reference nonexistent parent node ('dashboard_metrics', '0005_add_reconciliation_task')
```

Its tests are unaffected — the backend suite runs with `--no-migrations`.

Verified on a throwaway Postgres: `0004` → `0005` → `0006` applies from an empty database,
reverses, and re-applies, with `makemigrations --check` clean at each step.
## Env Config

None.

## Relevant Docs

UN-3883 analysis §6.4.

## Related Issues or PRs

Parent UN-3883. #2255 (UN-3973) — AC-3 depends on it. #2265 (UN-3974). See **Migration Order** above.

## Dependencies Versions

None.

## Notes on Testing

Nine scenarios and five unit tests. Detail in Jira comment 45381.

| # | Acceptance criterion | Verdict |
| --- | --- | --- |
| 1 | Index present, `indisvalid = t` | **Met** |
| 2 | Non-atomic + `CONCURRENTLY`, no write-blocking lock | **Met** |
| 3 | `get_documents_processed` free of a seq scan on `workflow_file_execution` | **Confirm on prod** |
| 4 | `get_failed_pages` free of a seq scan on `workflow_execution` | **Met** |
| 5 | `get_recent_activity` under 1 s | **Flagged** — belongs to the dropped index |

After deploy: confirm the index is valid, then re-pull the 6-hour Query Insights window against
the 870 / 809 ms baseline.

## Screenshots

n/a — schema-only change.

## Checklist


backend/dashboard_metrics/internal_views.py
```diff
@@ -0,0 +1,115 @@
...
+def _int_arg(request: Request, key: str, default: int) -> int:
+    """Read an optional positive integer from the request body."""
+    raw = request.data.get(key, default) if isinstance(request.data, dict) else default
+    try:
+        value = int(raw)
+    except (TypeError, ValueError) as exc:
+        raise ValueError(f"{key} must be an integer, got {raw!r}") from exc
+    if value < 1:
+        raise ValueError(f"{key} must be >= 1, got {value}")
+    return value
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps This file is not in this PR. The diff was computed against a stale base: the branch `UN-3883-Optimize-DB-cron-queries-causing-high-DB-load` was sitting one commit behind `main`, so the already-merged UN-3445 PostgreSQL queue commit (#2254) was being counted as part of this PR. The base is now `main` and the diff is the 2 files it should always have been — the `(status, created_at)` index and its migration. Nothing here touches the metrics internal API or the PG scheduler.

workers/queue_backend/pg_queue/pg_scheduler.py
```diff
@@ -230,3 +235,153 @@ def dispatch_due_schedules(conn: PgConnection) -> int:
...
+            base = cur.fetchone()[0]
+            cur.execute(
+                f"""
+                SELECT {", ".join(f'"{f}"' for f in _DuePeriodicTask._fields)}
+                FROM {qualified("pg_periodic_task")}
+                WHERE pg_owned AND enabled
+                  AND (next_run_at IS NULL OR next_run_at <= %s)
+                """,
+                (base,),
+            )
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps This file is not in this PR. The diff was computed against a stale base: the branch `UN-3883-Optimize-DB-cron-queries-causing-high-DB-load` was sitting one commit behind `main`, so the already-merged UN-3445 PostgreSQL queue commit (#2254) was being counted as part of this PR. The base is now `main` and the diff is the 2 files it should always have been — the `(status, created_at)` index and its migration. Nothing here touches the metrics internal API or the PG scheduler.

kirtimanmishrazipstack:
@greptile-apps please re-review.

The Confidence Score block above is stale — it still cites `backend/dashboard_metrics/internal_views.py` and `workers/queue_backend/pg_queue/pg_scheduler.py`, neither of which is in this PR. Both came from a stale-base diff that pulled the already-merged UN-3445 queue commit (#2254) in; the base is now `main` and the diff is 2 files. You retracted both findings in the threads and they are resolved.

For a human reader: this PR is `backend/workflow_manager/file_execution/models.py` (one `models.Index` line) and its hand-written concurrent migration. Nothing else.

kirtimanmishrazipstack:
@greptile-apps migration order is revised. Review again.

backend/workflow_manager/file_execution/tests/test_wfe_status_created_idx.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class MigrationShapeTests(SimpleTestCase):
+    """The properties that keep the build off the write path."""
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        super().setUpClass()
+        cls.migration = importlib.import_module(_MIGRATION).Migration
+        cls.operation = cls.migration.operations[0]
```
athul-rs, line 39: **[High] [Lens 13] — every assertion reads `operations[0]`, so a second write-blocking operation is invisible**

`setUpClass` binds `cls.operation = cls.migration.operations[0]` and no test asserts `len(migration.operations)`. Anything appended after the `SeparateDatabaseAndState` is never inspected. A plain `migrations.AddIndex` in that position **is** executed against the database, takes the `SHARE` lock for the whole build on the 3.4 GB table, and stalls file processing — exactly what this file exists to prevent.

Mutation-tested against this branch: appending

```python
migrations.AddIndex(
    model_name="workflowfileexecution",
    index=models.Index(fields=["status", "created_at"], name="wfe_dup_idx"),
),
```

to the real migration gives `5 passed, 2 warnings in 0.02s`.

**Suggested fix:** `self.assertEqual(len(self.migration.operations), 1)`, or iterate all operations and assert none is a bare `AddIndex`.

athul-rs (commented):
@kirtimanmishrazipstack — standardized pre-merge review (16-lens rubric, `unstract` plugin v0.18.1), mode INITIAL, at `261cd664`.

## Verdict: `REQUEST CHANGES`

**Critical: 0 · High: 2 · Medium: 8 · Low: 4 · Lenses run: 16/16**

**The shipped DDL is correct.** Everything below is about the guard that protects it, the numbers in the prose, and one design question. Both High findings were mutation-tested against this branch.

## Lens checklist

| # | Lens | Result |
|---|---|---|
| 2 | Architectural fit | See findings |
| 3 | Correctness & edge cases | Clean |
| 4 | Security | Clean — f-string interpolation of `INDEX_NAME` verified a non-issue (module-level literal, no external input; sqlparse 0.5.5 confirmed to keep the dollar-quoted block as one statement) |
| 5 | Data integrity & migrations | See findings |
| 6 | Concurrency | N/A — no shared state or locking in the diff |
| 7 | API & contract compatibility | N/A — no wire format, serializer or payload touched |
| 8 | Reliability & resilience | Clean |
| 9 | Performance & cost | See findings |
| 10 | Observability | N/A — no new code path to instrument |
| 12 | LLM/agent | N/A — no prompts, model calls, tools or evals |
| 14 | Dependencies & build | N/A — no manifest, lockfile, Dockerfile or CI workflow |
| 15 | Code quality | Clean |

### Verified sound — recorded so it is not re-litigated

* Column order `(status, created_at)` is right — equality-then-range against `services.py:107-118` and `:423-437`.
* The guard **does** fire in the case it was written for: with a leftover INVALID index the `CREATE ... IF NOT EXISTS` matches the existing relation and no-ops, then the guard raises; `atomic = False` means the raise leaves the migration unrecorded. Operation order is correct.
* Dependency on `0006_...` is genuinely the leaf on `main`, on the base branch and on every fetched ref; no competing branch claims `0007` (#2255 and #2265 add migrations to different apps).
* `(created_at DESC)` is **fully** dropped with no partial reintroduction. Because `status` leads, the planner cannot use this index for `get_recent_activity`, so **AC 5 remains unmet by design**, exactly as comment 45015 intended — the PR claims nothing otherwise.
* `Meta.indexes` and `state_operations` agree; rolling deploy is safe and reversible in both directions.

### Unanchored findings

* **[Medium] [Lens 16] PR description, "Can this PR break any existing features"** — *"against 1,679 ms of query time saved per 6 h"* is ~1,800x too small and argues **against** the change. 1,679 ms is `870 + 809`, the two queries' **per-call means**, relabelled as a 6-hour total. The real figure is ≈2,980 s per 6 h (870 ms x 1,776 calls + 809 ms x 1,776 calls), consistent with the same PR body's "83% of 3,580 s" two paragraphs earlier. A reviewer weighing slower writes on a million-insert table against 1.7 seconds saved should reject the index.
* **[Medium] [Lens 16] PR description, Notes on Testing row 3** — AC-3 is stated against a seq scan on `workflow_file_execution` while row 4 and the migration docstring say `workflow_execution`. The table contradicts itself between adjacent rows, and per UN-4045's plan `workflow_file_execution` is reached by index scan in both queries. Whoever runs the prod check will look for a scan that was never there.
* **[Medium] [Lens 13, 5]** Nothing in any lane parses or executes this migration's SQL — `--no-migrations` is repo-wide `addopts` (`pyproject.toml:115`), so a PL/pgSQL syntax error in the `DO $$ ... $$` block is undetectable here and surfaces first as a failed deploy-time `migrate` on the 3.4 GB table. AC 1 (`indisvalid = t`) rests entirely on a manual throwaway-Postgres run. `integration-backend` already provisions Postgres — one `@pytest.mark.integration` test would cover AC 1, the reversibility half of AC 2, and DO-block parseability. (AC 3/4 genuinely cannot be asserted in CI: on a tiny test table the planner seq-scans regardless.)
* **[Medium] [Lens 13, 2]** `workflow_v2/migrations/0023` is the *identical* construction — `atomic = False`, `SeparateDatabaseAndState`, a byte-identical validity guard — on a multi-million-row table, and has **no** shape guard today; its only test asserts on `Meta.indexes`. So this protects one of the two files that need it, and sets "hand-write an 83-line guard per concurrent-index migration" as the pattern. A parametrized guard walking the migration graph for `CONCURRENTLY` would cover 0023 and every future one for free.
* **[Low] [Lens 5]** The guard's `pg_class` lookup is not schema-qualified while `DB_SCHEMA` defaults to `unstract`, not `public` (`settings/base.py:155`) — false-alarm direction only.
* **[Low] [Lens 13]** `TABLE`/`INDEX_FIELDS` is a tautology: Django *field* names supply both the expectation and the SQL column check, so it cannot detect a field-name != column-name divergence. The repo already contains that trap (`workflow_v2/models/execution.py:156-163`, `db_column="workflow_id"`). Derive from `_meta.db_table` / `_meta.get_field(f).column` instead.
* **[Low] [Lens 13]** The migration is located by hardcoded dotted path, so a renumber during the stacked merge errors all five tests as an import failure that reads as "stale test" rather than "the guard may have been regenerated".
* **[Low] [Lens 13]** `test_model_meta_matches_the_migration` never reads the migration; it holds only transitively via `test_add_index_updates_state_only`.

### What the guard does earn

Recorded so none of the above reads as dismissal — three properties hold up under mutation: regenerating 0007 with `makemigrations` fails **four of five** tests; swapping `AddIndex` into `database_operations` fails the `RunSQL` check; and swapping create/guard order fails two tests, which is load-bearing and easy to lose in a refactor.

### Merge order

`#2264` first — it is the only one of the three with no code overlap. Then #2255, then #2265 rebased on it. AC-3 should be measured on prod only after **all three** land: at today's 52-day window `COMPLETED` is 20.4% of the table, so this index is inert for `get_documents_processed` until #2255's 2-day window ships.

Reviewed with the PR Review Toolkit agents (code-reviewer, pr-test-analyzer, comment-analyzer); lenses 5, 6, 8, 11, 12, 14 assessed directly. Verdict is advisory — posted as a comment, not a merge gate.

backend/workflow_manager/file_execution/tests/test_wfe_status_created_idx.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def test_invalid_index_guard_is_present(self) -> None:
+        """An interrupted concurrent build leaves an INVALID index.
+
+        ``IF NOT EXISTS`` would keep it and let Django record the migration as applied —
+        green, but the index costs on every write and is never read. The guard turns that
+        into a loud failure.
+        """
+        guard = self.operation.database_operations[1].sql
+        self.assertIn("indisvalid", guard)
```
athul-rs: **[High] [Lens 13, 16] — the INVALID-index guard is checked for three substrings; its semantics are entirely unasserted**

`assertIn("indisvalid", guard)`, `assertIn("RAISE EXCEPTION", guard)` and the `DROP INDEX ...` check are satisfied by any text containing those characters — including a SQL comment. Neither the **polarity** (`AND NOT i.indisvalid`, which is the whole point) nor the `c.relname = '<name>'` predicate that scopes the guard to this index is checked.

Mutation-tested, both green:
* replacing the entire guard with `f"-- indisvalid RAISE EXCEPTION DROP INDEX CONCURRENTLY IF EXISTS {INDEX_NAME}\nSELECT 1;"` → `5 passed`
* flipping `AND NOT i.indisvalid` → `AND i.indisvalid` → all four shape assertions still pass

Flip it one way and the migration raises on every healthy deploy; flip it the other and it never fires, so Django records the migration applied while the index is physically unusable — the exact silent failure the docstring above claims to turn into a loud one.

Note the third `assertIn` only matches the text inside the `RAISE EXCEPTION` *message string*, not the `WHERE` clause, so it carries no information about the predicate.

**Suggested fix:** `assertIn("NOT i.indisvalid", guard)` and `assertIn(f"c.relname = '{INDEX_NAME}'", guard)`.

backend/workflow_manager/file_execution/tests/test_wfe_status_created_idx.py
```diff
@@ -0,0 +1,83 @@
...
+    def test_migration_is_non_atomic(self) -> None:
+        """CREATE/DROP INDEX CONCURRENTLY is rejected inside a transaction block."""
+        self.assertIs(self.migration.atomic, False)
+
+    def test_index_is_built_and_dropped_concurrently(self) -> None:
+        """Both directions must stay off the write-blocking lock path."""
+        create = self.operation.database_operations[0]
+        self.assertIn("CREATE INDEX CONCURRENTLY IF NOT EXISTS", create.sql)
+        self.assertIn(INDEX_NAME, create.sql)
+        self.assertIn(f"{TABLE} ({', '.join(INDEX_FIELDS)})", create.sql)
```
athul-rs: **[Medium] [Lens 13] — this assertion couples to incidental SQL formatting**

`f"{TABLE} ({', '.join(INDEX_FIELDS)})"` demands the exact byte sequence `workflow_file_execution (status, created_at)`.

The **column-order** coupling is genuinely the point — `(created_at, status)` would not serve the cron's equality-plus-range filter, and this correctly catches that, and an appended `DESC`. The whitespace and casing coupling is not: mutation-tested, `ON workflow_file_execution(status, created_at);` (one space removed) fails, as does a wrapped line and uppercased identifiers. That is a false-failure mode with no matching true-failure mode, and the natural response to a red test on semantically identical SQL is to loosen the assertion.

**Also [Medium]:** `reverse_sql` is never checked for the index name. Setting it to `DROP INDEX CONCURRENTLY IF EXISTS wfe_typo_idx;` gives `5 passed` — a typo there makes rollback a silent no-op through `IF EXISTS`, so `migrate file_execution 0006` unapplies the migration while the index stays on the table, and Django's recorded state diverges from the physical schema.

**Suggested fix:** `re.search(rf"ON\s+{TABLE}\s*\(\s*status\s*,\s*created_at\s*\)", create.sql, re.I)`, and add `assertIn(INDEX_NAME, create.reverse_sql)`.

backend/workflow_manager/file_execution/migrations/0007_wfe_status_created_idx.py
```diff
@@ -0,0 +1,66 @@
+"""Add a (status, created_at) index to workflow_file_execution.
+
+The dashboard metrics cron filters this table on status and a created_at window.
+Every existing index leads with workflow_execution_id, so the planner cannot drive
+from here and scans workflow_execution in full instead. Measurements in UN-3883.
+
+Built CONCURRENTLY (atomic = False): a plain AddIndex holds a SHARE lock for the
+whole build and would block writes to a large, write-heavy table. Prefer building
```
athul-rs: **[Medium] [Lens 5, 11] — "prefer building it out of band" without saying what to build, and the nearest fallback runbook is harmful**

Commit `6d21dc52b` removed the `CREATE INDEX CONCURRENTLY ...` block this docstring refers to, so an operator following this instruction reaches for the nearest runbook — and UN-3972's description step 3 still carries the **two-index** version, including `wfe_created_at_desc_idx`.

That index was struck in comment 45015 as *worse than nothing*: with it present the planner gains an ordered-index-scan path costed against a near-uniform ~80 K estimate, so it looks nearly free and will likely be chosen, then walks up to 3.3 M entries with a random heap fetch each for a small tenant — slower than today's sequential scan plus top-N sort.

**Suggested fix:** restore the exact single-index `CREATE` statement to the PR body's Database Migrations section, and correct UN-3972's description so the stale snippet is not the nearest source.

backend/workflow_manager/file_execution/migrations/0007_wfe_status_created_idx.py
```diff
@@ -0,0 +1,66 @@
...
+            "0006_workflowfileexecution_wf_file_hash_path_status_idx_and_more",
+        ),
+    ]
+
+    operations = [
+        migrations.SeparateDatabaseAndState(
+            database_operations=[
+                migrations.RunSQL(
+                    sql=(
+                        f"CREATE INDEX CONCURRENTLY IF NOT EXISTS {INDEX_NAME} "
```
athul-rs, line 81: **[Medium] [Lens 5] — neither `IF NOT EXISTS` nor the guard checks the index's *shape***

Both match on **name only**. If an index called `wfe_status_created_idx` already exists with a different column list or order, the `CREATE` no-ops, the guard checks only `indisvalid` and passes, and Django records 0007 as applied — while `Meta.indexes` asserts `(status, created_at)` and the database holds something else. The shape test cannot catch this; it inspects the Python module, never the database.

**Suggested fix:** extend the same `DO` block with a second `IF` comparing `pg_get_indexdef(c.oid)` against the expected definition (~4 lines, no new pattern).

---

**[Medium] [Lens 9, 2] — full index where the measurement and the repo's own precedent argue for partial**

By comment 45015's own numbers, the only query this serves today is `get_failed_pages` (`ERROR` = 0.40% of rows); `COMPLETED` is 97.6%, so the 52-day `(COMPLETED, created_at)` range is 20.4% of the table and `get_documents_processed` keeps its scan.

More importantly `status` is this index's **leading key and is mutated on every row**: `models.py:140` (`self.status = status.value`, from `update_status`) rewrites it at each PENDING → EXECUTING → COMPLETED transition, relocating the entry each time. The real cost is roughly three index tuples plus two dead ones per row lifecycle, not the "one more index entry per inserted row" the ticket budgeted — on a table taking millions of inserts, with sustained autovacuum pressure in the in-flight regions.

The precedent this PR follows, `workflow_v2/migrations/0023`, chose **partial** for exactly this reason and documented the trade (~2 MB vs ~20 MB).

*Counter-argument that may be decisive:* the full shape is correct once #2255 narrows the window, and the merge table requires #2255 first. If that's the reasoning, one docstring line should say so.

backend/workflow_manager/file_execution/migrations/0007_wfe_status_created_idx.py
```diff
@@ -0,0 +1,66 @@
+"""Add a (status, created_at) index to workflow_file_execution.
+
+The dashboard metrics cron filters this table on status and a created_at window.
+Every existing index leads with workflow_execution_id, so the planner cannot drive
```
athul-rs: **[Medium] [Lens 16] — this docstring cites UN-3883 for a claim UN-3883 contradicts, and its absolute is false**

Two issues in one sentence.

1. *"scans `workflow_execution` in full instead. Measurements in UN-3883."* — the prose is **correct**, but UN-3883's 2026-08-10 analysis is the source of the *`workflow_file_execution`* seq-scan claim (quoted verbatim in UN-3972's description). The `workflow_execution` finding is UN-4045's, dated 2026-08-31, which explicitly overturns it. A maintainer chasing the citation lands on contradicting text with no way to tell which is current. Suggest: *"(execution plan in UN-4045; cost measurements in UN-3883)"*.

2. *"Every existing index leads with `workflow_execution_id`"* — false. The table carries 8 indexes and the primary-key index leads with `id` (`models.py:69`). The conclusion holds; the absolute doesn't, and it is the kind of claim a reader trusts rather than re-verifies. Suggest: *"No existing index leads with `status` or `created_at` — every secondary index is prefixed by `workflow_execution_id`."*

backend/workflow_manager/file_execution/models.py
```diff
@@ -198,6 +198,12 @@ class Meta:
                 ],
                 name="wf_provider_uuid_path_stat_idx",
             ),
+            # Serves the dashboard metrics cron's status + date-window filter; every
```
athul-rs: **[Medium] [Lens 16, 1] — this comment states a benefit the ticket's own measurement says is not realised for the larger query**

"Serves the dashboard metrics cron's status + date-window filter" reads as: this index fixes both cron queries. Per comment 45015 it fixes one. The next engineer reads this, concludes UN-3883's hot path is handled, and either doesn't ship the window narrowing the index depends on, or debugs a Query Insights window that has barely moved with no comment explaining why.

Two smaller points on the same line: *"every index above"* is scoped by **position in a list**, so appending or reordering an index anywhere above silently falsifies it with nothing to catch the drift — `test_model_meta_matches_the_migration` asserts nothing about the other entries. And `workflow_execution` here means the FK *column*, while the identical token in migration 0007's docstring means the *table*.

**Suggested fix:** *"every other index on this table is prefixed by the `workflow_execution` FK column, so none can serve a `status` + `created_at` filter. Effective for `get_failed_pages` today; the `COMPLETED` path stays inert until UN-3973 narrows the window. See migration 0007."*

backend/workflow_manager/file_execution/migrations/0007_wfe_status_created_idx.py
```diff
@@ -0,0 +1,66 @@
...
+        (
+            "file_execution",
+            "0006_workflowfileexecution_wf_file_hash_path_status_idx_and_more",
+        ),
+    ]
+
+    operations = [
+        migrations.SeparateDatabaseAndState(
+            database_operations=[
+                migrations.RunSQL(
```
kirtimanmishrazipstack, line 79: **[Medium] [Lens 5] — `IF NOT EXISTS` matches on name only, and the guard checks validity but not shape**

The module docstring above tells the operator to *prefer* building this index by hand out of band, so the intended production path is a human typing `CREATE INDEX CONCURRENTLY wfe_status_created_idx ...`. If that hand-built index has the columns transposed, a `WHERE` condition, or a different method, this statement matches on **name** and no-ops. The `_ASSERT_INDEX_VALID` guard that follows checks `indisvalid` and nothing about the definition, so it passes. Django then records 0007 as applied and `AddIndex(fields=["status", "created_at"])` into model state.

The result is a permanent, invisible divergence: `Meta.indexes` and `django_migrations` both assert `(status, created_at)`; the database holds something else. `makemigrations --check` compares model state to migration state, so it never notices, and the new test module reads the Python source rather than the database. A later reverse then drops whatever the operator built under that name.

Reproduced on a throwaway PostgreSQL 15 seeded to 800k rows over 255 days:

```
CREATE INDEX wfe_status_created_idx ON workflow_file_execution (created_at, status);  -- operator
CREATE INDEX CONCURRENTLY IF NOT EXISTS wfe_status_created_idx ON ... (status, created_at);
NOTICE:  relation "wfe_status_created_idx" already exists, skipping
DO                                   -- the validity guard passes
indexdef: ... USING btree (created_at, status)   -- what the DB actually has
```

**Fix:** extend the `DO` block to assert the definition, not just validity — compare `pg_get_indexdef(i.indexrelid)` against the expected `btree (status, created_at)` and `RAISE EXCEPTION` on mismatch, reusing the same "drop it and re-run" message. While in there, qualify the lookup by namespace (`JOIN pg_namespace n ON n.oid = c.relnamespace` against `current_schema()`) — `pg_class.relname` is unqualified today and this deployment puts application tables in the `unstract` schema, not `public`.

backend/workflow_manager/file_execution/tests/test_wfe_status_created_idx.py
```diff
@@ -0,0 +1,83 @@
...
+    def setUpClass(cls) -> None:
+        super().setUpClass()
+        cls.migration = importlib.import_module(_MIGRATION).Migration
+        cls.operation = cls.migration.operations[0]
+
+    def test_migration_is_non_atomic(self) -> None:
+        """CREATE/DROP INDEX CONCURRENTLY is rejected inside a transaction block."""
+        self.assertIs(self.migration.atomic, False)
+
+    def test_index_is_built_and_dropped_concurrently(self) -> None:
```
kirtimanmishrazipstack, line 57: **[Medium] [Lens 13] — Nothing pins that the migration is reversible; dropping the guard's `noop` breaks rollback with every test green**

This test reads `reverse_sql` on `database_operations[0]` only. The second operation — the INVALID-index guard — carries `reverse_sql=migrations.RunSQL.noop`, and nothing asserts it. Delete that kwarg and `RunSQL.reverse_sql` is `None`, `reversible` is `False`, and `SeparateDatabaseAndState.database_backwards` raises `NotImplementedError("You cannot reverse this operation")` before the pinned `DROP INDEX CONCURRENTLY` ever runs.

So `migrate file_execution 0006` — the rollback path for a concurrent build on a large write-heavy table — is dead, and all five tests still pass. The docstring says "Both directions must stay off the write-blocking lock path," which a reader takes as reverse being covered; it pins the reverse *SQL text* of one of two operations, not that the migration can be reversed at all.

**Fix:** one line — `self.assertTrue(all(op.reversible for op in self.operation.database_operations))`.

kirtimanmishrazipstack (commented):
## Standardized review — FOLLOWUP

Full 19-lens pass at head `261cd6640`, diff scoped to `eebaa03f0..HEAD` (4 files, file count reconciled against GitHub). Mediums and Highs only — Lows withheld by request.

**0 High, 2 Medium.** The cleanest of the three stacked PRs.

### Verified clean, reported explicitly rather than by silence

- `makemigrations --check --dry-run` → *No changes detected*; name, field list and order agree between `models.py` and the `AddIndex`.
- `atomic = False` on the Migration and all three operations under Django 4.2.30, so nothing re-wraps them in a transaction. Executed forward, forward-again, reverse and reverse-again against real PostgreSQL — all clean; `BEGIN; CREATE INDEX CONCURRENTLY` errors as expected, confirming `atomic = False` is load-bearing.
- Field order is right: equality column (`status`) before range column (`created_at`) for both consuming queries in `dashboard_metrics/services.py`. Measured 106 ms / 9336 buffers → 47 ms / 1695 buffers, and the planner does pick it.
- Tests land in the fast tier — `SimpleTestCase` is not auto-marked `integration` by `conftest.py` — and are non-vacuous: flipping `atomic`, renaming the module, or transposing `Meta.indexes` each fails a distinct test.
- Migration graph: `file_execution` has exactly one leaf, and one leaf per app across all three stacked PRs merged together.

### Coverage

Lenses 5, 6, 8, 11, 12, 14, 17, 18, 19 assessed by me directly; 1, 2, 3, 9, 10, 15 by the correctness pass; 13 by the test pass. Lens 4 N/A (no auth, input, or secret surface). Lens 7 N/A (no contract change). Lens 16 clean.

kirtimanmishrazipstack:
## Review findings addressed — 2 Medium

Both fixed.

- **`IF NOT EXISTS` matched on name only** — the guard now asserts the *definition*, not just validity. It reads `pg_get_indexdef` and rejects anything that is not `USING btree (status, created_at)` with the same "drop it and re-run" message, and the lookup is qualified by `current_schema()` (`pg_class.relname` was unqualified, and app tables live in `unstract`, not `public`). A hand-built index with the columns transposed now fails the migration instead of leaving `Meta.indexes` and the database permanently disagreeing.
- **Reversibility unpinned** — `test_every_database_operation_is_reversible` asserts `op.reversible` across both `database_operations`. Dropping `reverse_sql=migrations.RunSQL.noop` from the guard used to leave every test green while `migrate file_execution 0006` raised `NotImplementedError` before the `DROP INDEX CONCURRENTLY` ever ran.

The guard test was extended alongside: it now asserts `pg_get_indexdef`, the expected `USING btree (...)` fragment and the `n.nspname = current_schema()` predicate, so removing any of the three fails.

backend/workflow_manager/file_execution/migrations/0007_wfe_status_created_idx.py
kirtimanmishrazipstack: Fixed in 1628a1681, taking both halves.

The guard now selects `pg_get_indexdef(i.indexrelid)` and `i.indisvalid` in one lookup and raises on three conditions: the index missing, invalid, or not ending in `USING btree (status, created_at)`. The mismatch message includes the actual definition and the same "drop it and re-run" instruction.

The lookup is qualified — `JOIN pg_namespace n ON n.oid = c.relnamespace` against `current_schema()` — so it reads the index `CREATE INDEX` would have created, not a same-named one in another schema.

The test module was extended alongside, and each part is mutation-checked: transposing the expected fragment, or removing the `n.nspname = current_schema()` predicate, each fails a distinct assertion.

backend/workflow_manager/file_execution/tests/test_wfe_status_created_idx.py
kirtimanmishrazipstack: Fixed in 1628a1681 — the one line, as suggested.

`test_every_database_operation_is_reversible` asserts `op.reversible` across both `database_operations`. Verified by mutation: dropping `reverse_sql=migrations.RunSQL.noop` from the guard fails it, and failed nothing before.

backend/workflow_manager/file_execution/tests/test_wfe_status_created_idx.py
kirtimanmishrazipstack: @athul-rs Fixed in b9f340601. Your mutation is caught now.

Two assertions rather than one, because they fail for different reasons: `test_the_migration_has_exactly_one_operation` pins the count, and `test_no_operation_builds_an_index_against_the_database` asserts no operation is a bare `AddIndex` — so the guard survives the count legitimately changing.

Re-ran your exact mutation (appending the `wfe_dup_idx` AddIndex): 2 failed, 6 passed. It was 5 passed before.

backend/workflow_manager/file_execution/tests/test_wfe_status_created_idx.py
kirtimanmishrazipstack: @athul-rs Fixed in b9f340601 — both mutations caught.

The guard now asserts polarity and scope, not presence: `IF NOT idx_valid THEN`, `i.indisvalid`, `c.relname = '<name>'` and `n.nspname = current_schema()`. Flipping the polarity fails it (1 failed, 7 passed); replacing the whole guard with a comment containing those tokens fails it too, because a comment satisfies none of the structural fragments.

The guard itself grew while I was in there — it reads `pg_get_indexdef` and rejects anything that is not `USING btree (status, created_at)`, so a hand-built index with transposed columns no longer no-ops through `IF NOT EXISTS` while Django records `(status, created_at)` into model state. The namespace qualification matters here for the reason you would expect: app tables live in `unstract`, not `public`.

backend/workflow_manager/file_execution/tests/test_wfe_status_created_idx.py
kirtimanmishrazipstack: @athul-rs Both fixed in b9f340601, and I agree with the distinction you drew.

The column-order coupling is the point and is kept — as a regex: `ON\s+{TABLE}\s*\(\s*status\s*,\s*created_at\s*\)`. Whitespace and case are not, and your framing of it — a false-failure mode with no matching true-failure mode, whose natural response is to loosen the assertion — is why I took the regex rather than leaving it. Verified: removing the space now passes; transposing the columns still fails.

`assertIn(INDEX_NAME, create.reverse_sql)` added. Your typo mutation gave 5 passed before; it fails now.

backend/workflow_manager/file_execution/migrations/0007_wfe_status_created_idx.py
kirtimanmishrazipstack: @athul-rs Fixed in b9f340601 for the half that lives in this repo.

The migration docstring now carries the exact statement to run, verbatim and alone:

```sql
CREATE INDEX CONCURRENTLY IF NOT EXISTS wfe_status_created_idx
    ON workflow_file_execution (status, created_at);
```

and warns explicitly off the two-index variant, naming `wfe_created_at_desc_idx` as struck. So an operator following "prefer building it out of band" no longer has to reach for a runbook at all.

The Jira half is yours or mine to do outside the code — I have not edited UN-3972's description. Say the word and I will, or I can leave it to you since you struck the snippet in 45015.

backend/workflow_manager/file_execution/migrations/0007_wfe_status_created_idx.py
kirtimanmishrazipstack: @athul-rs Shape check fixed in b9f340601 — the `DO` block now compares `pg_get_indexdef` against the expected `USING btree (status, created_at)` and raises with the actual definition in the message, plus the namespace qualification.

On the partial-vs-full sub-finding: taking your own counter-argument, and I have written it into the docstring rather than leaving it implicit. Full is right *because* of the merge order — `get_failed_pages` benefits at today's window with `ERROR` at 0.40%, and `get_documents_processed` only once UN-3973 narrows to 2 days, since at 52 days the `COMPLETED` slice is 20.4% and the planner scans regardless. A partial index on `ERROR` would serve the first and never the second.

The `status`-mutation cost you raise is real and I have not tried to argue it away — three index tuples plus two dead ones per row lifecycle is the honest number, not one entry per insert. If the write amplification turns out to dominate in production, the partial variant is a follow-up with a measurement behind it rather than a guess.

backend/workflow_manager/file_execution/migrations/0007_wfe_status_created_idx.py
kirtimanmishrazipstack: @athul-rs Both fixed in b9f340601.

Citation now reads "Execution plan in UN-4045 (2026-08-31, which supersedes the earlier `workflow_file_execution` reading); cost measurements in UN-3883" — so a maintainer chasing it lands on the current one and knows the older text is superseded rather than contradictory.

The absolute is gone: "No existing index leads with `status` or `created_at` — every secondary index is prefixed by the `workflow_execution` FK column." You are right that the PK leads with `id`, and right that it is the kind of claim a reader trusts instead of re-verifying.

backend/workflow_manager/file_execution/models.py
kirtimanmishrazipstack: @athul-rs Fixed in b9f340601, close to your wording.

The comment now reads: every *other* index on this table is prefixed by the `workflow_execution` FK column, so none can serve a `status` + `created_at` filter; effective for `get_failed_pages` today; the `COMPLETED` path stays inert until UN-3973 narrows the window.

That covers all three points — the overstated benefit, the positional "every index above" (now "every other index on this table", which no reordering can falsify), and the `workflow_execution` column/table ambiguity, since the sentence now says "FK column" explicitly.

PR #2273, opened 2026-09-02
UN-2868 [FIX] Make settings read-only and block deletion on resources shared with a user

kirtimanmishrazipstack:
## What

Shared access now means different things depending on the resource:

- **Prompt Studio and Agentic Prompt Studio** are shared for **collaboration** — prompts, Settings and LLM profiles are all editable by a shared user.
- **Everything else** — workflows, ETL/Task pipelines, API deployments, adapters, connectors, Look-Ups — is shared for **use**: open, run, pass access on, but do not reconfigure.
- **Renaming and deleting stay with the owner** everywhere. Sharing onward stays open; only the owner revokes access or shares org-wide.
- **Locked controls are greyed out with a reason on hover**, not hidden.
- Two unrelated fixes: a shared user saw a project's internal ID instead of its name, and the workflow rename dialog opened empty.

Three holes found in review and closed here: **anyone in the org could attach a tool to a workflow they did not own** (which also switched that workflow on), the endpoint that clears execution markers answered a plain link click, and the workflow builder's connector-type dropdown was the one control on that card still offered to shared users.

## Why

Reported by a customer — a shared user could repoint another team's workflow at a different output folder. And where the server did refuse a write, the screen said nothing, so people filled in a form and lost the work.

The first attempt applied read-and-run-only everywhere. That was wrong for Prompt Studio: prompts, preamble, grammar and the LLM profile all go together on every run, so granting one and withholding the rest describes a job nobody does. Hence the split by what the resource is *for*.

## How

| Resource | A shared user can | Owner / co-owner only |
|---|---|---|
| **Prompt Studio** | Add, edit, reorder, delete prompts. Change every Settings tab. Create, edit, delete LLM profiles and set the default. Toggle SinglePass. Run and view output. | Rename, delete |
| **Agentic Prompt Studio** | Change project settings (LLM, agent LLM, lightweight LLM, text-extractor). Work on documents and prompts. Run. | Rename, delete |
| **Lookup Studio** | View and use it inside a shared Prompt Studio project. Now reachable via **group** sharing, which was missing. | Rename, delete |
| **Workflows** | Open, run, view runs. | Connector config, DB/tool/HITL settings, rename, delete |
| **ETL / Task pipelines** | Open, view runs and logs. | Enable toggle, cron, connector config, rename, delete |
| **API deployments** | Open, view and copy existing keys. | Enable toggle, add a key, a key's active toggle, rename, delete |
| **Adapters** | Use it in an LLM profile or workflow. | Edit, delete |
| **Connectors** | Use it in a workflow. | Edit, delete |

Notes on the mechanism:

- `canEditResource` (frontend) and `_can_access_tool` (backend) are the single definitions of the rule, so screens cannot drift apart.
- DRF never runs object permissions on `create`, so collection-level actions need their own guard — `WorkflowOwnerMutationMixin`, `ParentToolAccess`, and the tool-instance check added here. Without it those routes were open to the whole organisation.
- Renaming is checked per field, not per action: a project's name and its settings share one PATCH.
- `clear_file_marker` moves from GET to POST. It has no frontend caller.
- `IsParentToolOwner` is deleted — zero callers after the switch to `ParentToolAccess`.

- **Shared users lose configuration access outside Prompt Studio** — connector and tool settings, HITL, status toggles, key creation. These were writes to someone else's resource; this is the ticket's product decision.
- **Owners see one change** — the connector window's Save now also saves HITL rules and the separate "Save Rules" button is gone (cloud PR). Nothing writes until Save.
- **Prompt Studio is unchanged for shared users** — only rename and delete moved.
- Reading, running, sharing onward, service accounts and API keys are all untouched.

## Database Migrations

- None.

## Env Config

- None.

## Relevant Docs

- None.

## Related Issues or PRs

- **Cloud PR #1751 — must merge after this one.** It imports `WorkflowOwnerMutationMixin`, `canEditResource` and `ReadOnlyNotice` from here.

## Dependencies Versions

- None.

## Notes on Testing

| Surface | Shared user | Owner / admin |
|---|---|---|
| Prompt Studio — add, edit, reorder, delete a prompt | available | unchanged |
| Prompt Studio — Settings modal, all tabs | available | unchanged |
| Prompt Studio — create, edit, delete an LLM profile; set default | available | unchanged |
| Prompt Studio — rename | blocked (403) | saves |
| Prompt Studio — delete | greyed out | unchanged |
| Prompt Studio — profile create by a non-shared org member | blocked (403) | — |
| Workflow connector config — read / change | 200 / blocked (403) | 200 / 200 |
| Workflow connector config, not shared at all | blocked (404) | — |
| Workflow tool settings | read-only notice, no Save | unchanged |
| Rename from a resource list / builder header | greyed out | opens pre-filled, saves |
| Adapter, connector settings | greyed out | unchanged |
| ETL pipeline / API deployment status toggle | greyed out | unchanged |
| Manage Keys — New Key, per-key active toggle | greyed out | unchanged |
| Manage Keys — view and copy a key | available | unchanged |
| Delete / Share, on any resource | greyed out / available | unchanged |
| PS project name shown in a shared workflow | name, not ID | unchanged |
| Add a tool to a workflow | blocked (403) | 200 |

### Automated tests

New: `backend/permissions/tests/test_shared_user_gates.py`. `IsParentToolOwnerTests` is retargeted at the live gate as `ParentToolAccessTests`, with the viewer case flipped from denied to allowed.

| Test | Asserts |
|---|---|
| `test_shared_viewer_cannot_change_connector_config` | 403, config unchanged |
| `test_shared_viewer_cannot_delete_the_endpoint` | 403, row survives |
| `test_shared_viewer_can_still_read_it` | 200 — refusing a write must not hide the resource |
| `test_owner_and_co_owner_can_change_it` | 200 for both |
| `test_a_user_with_no_access_gets_404_not_403` | 404, so a refusal does not confirm the resource exists |
| `test_shared_viewer_cannot_add_a_tool` | 403 on tool-instance create |
| `test_a_user_with_no_access_gets_404` | 404 on tool-instance create |
| `test_shared_user_cannot_rename_the_project` | 403, name unchanged |
| `test_shared_user_can_change_a_settings_field` | 200 — same endpoint as the rename |
| `test_owner_can_rename` | 200 |
| `test_resending_the_same_name_is_not_a_rename` | 200 — echoing the current name is not a rename |
| `test_parent_tool_viewer_allowed_outsider_denied` | the collaboration rule on profiles |
| `test_create_resolves_the_parent_from_the_payload` | owner and viewer allowed, outsider refused |

Run: `tox -e groups -- integration-backend`. All are `django.test.TestCase`, so the backend conftest auto-marks them `integration` and they join that group with no manifest edit — verified: 24 collected under `-m integration` across both repos, 24 passed.

Every gate above was broken in turn to confirm the matching test fails. That caught one test of my own that passed for the wrong reason; it was deleted rather than kept.

## Checklist

https://claude.ai/code/session_015aPCgGhE6Ma2NEhB8LQP1c


backend/permissions/permission.py
```diff
@@ -148,6 +149,34 @@ def has_object_permission(self, request: Request, view: APIView, obj: Any) -> bo
...
+    def get_permissions(self) -> list[Any]:
+        if self.action in ("update", "partial_update", "destroy"):
+            return [IsParentWorkflowOwner()]
+        return list(super().get_permissions())
+
+    def perform_create(self, serializer: Any) -> None:
+        workflow = serializer.validated_data.get("workflow")
+        if workflow and not is_workflow_mutator(self.request, workflow):
+            raise PermissionDenied(self.mutation_denied_message)
+        serializer.save()
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Valid — fixed in 06198ccb7.

`perform_create` now fails closed:

```python
workflow = serializer.validated_data.get("workflow")
if not workflow or not is_workflow_mutator(self.request, workflow):
    raise PermissionDenied(self.mutation_denied_message)
```

Every viewset using the mixin has a required `workflow` field (`RuleEngineSerializer`, `SettingsSerializer` via `fields = "__all__"` on a non-null FK, and `WorkflowEndpointViewSet`, which routes no POST), so denying a payload without one cannot break a live path while removing the unchecked-mutation hole.

kirtimanmishrazipstack:
@greptile-apps Both findings fixed in 06198ccb7, plus three more found while tracing the second one.

**Partial save / false close** — valid, and the sharper of the two. `handleValidateAndSubmit` caught its own error and returned, so a failed endpoint write fell straight through to the HITL write and `handleSaveAndClose` dismissed the modal. Both writes now report success and `handleSave` bails on the first failure:

```js
if (!(await handleValidateAndSubmit(formDataConfig))) {
  return false;
}
```

**Create guard fails open** — fixed, replied inline.

Found alongside these:

- The HITL write now only runs when its form is actually dirty. Chaining it unconditionally meant every connector save also wrote a rule row.
- A view-only user could be asked whether to save unsaved changes on close, which they cannot have made. `hasUnsavedChanges()` returns false when the workflow is not editable.
- `ReadOnlyNotice` used `defaultProps`, which React 19 ignores on function components, so the default message never applied. Now a default parameter.

The companion cloud PR ([#1751](https://github.com/Zipstack/unstract-cloud/pull/1751)) also fixes the dirty flag that made this reachable: the HITL form was dirty from the moment it opened, because the baseline was either `null` or the server's raw `rule_json` compared against a builder-normalised round-trip. Sync destination and TTL are folded into that flag too, since they lost their own Save button.

Re-verified the permission matrix against real shared-workflow data after the change — owner and admin-not-owner write, shared viewer 403, non-shared 404, reads unaffected on both database and API destinations.

kirtimanmishrazipstack:
@greptile-apps review

Both findings from the previous pass are addressed in 06198ccb7, which landed after that review was generated.

frontend/src/components/agency/configure-connector-modal/ConfigureConnectorModal.jsx
```diff
@@ -318,34 +329,57 @@ function ConfigureConnectorModal({
...
+    // Save now writes everything the modal shows. Only when they actually
+    // changed -- otherwise every connector save would write a rule too.
+    if (ruleEngineHasChanges && ruleEngineRef.current?.save) {
+      setIsSavingEndpoint(true);
+      try {
+        // Keep the modal open on failure so the edit is not lost.
+        if (!(await ruleEngineRef.current.save())) {
+          return false;
+        }
+      } finally {
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Half of this is real and now fixed; the other half I am deliberately not doing.

**Fixed** in 8a84d68e7: the misleading message. `handleValidateAndSubmit` toasted "Configuration saved successfully" before the rule write ran, so a rule failure left a green toast beside a red one. The connector success message is now suppressed when a rule write follows, and `RuleEngine.save()` — which already toasts its own success and failure — reports the outcome for both.

**Not doing** the rollback. The endpoint write succeeded, so marking that form clean is correct: it matches what is on the server. The unsaved-changes flow discards uncommitted edits; it has never reverted committed ones, and making it do so would mean a compensating write to a second service with its own failure mode.

The state after a partial failure is recoverable, not stuck: `ruleEngineHasChanges` stays true, so Save stays enabled and closing still warns; the rule edit is preserved in the open modal; and pressing Save again skips the endpoint (no diff now) and retries only the rule.

backend/prompt_studio/permission.py
```diff
@@ -26,10 +27,12 @@ def has_object_permission(self, request: Request, view: APIView, obj: Any) -> bo
...
         if _is_resource_owner(request.user, tool):
             return True
-        if _is_resource_viewer(request.user, tool):
-            return True
-        if has_group_access(request.user, tool):
+        if request.method in permissions.SAFE_METHODS and (
+            _is_resource_viewer(request.user, tool)
+            or has_group_access(request.user, tool)
+        ):
             return True
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Valid at the time, and fixed in 14df8ab01.

The product call on the ticket reversed this: a Prompt Studio project is shared for collaboration, so prompts stay editable. `_can_access_tool` no longer filters by HTTP method, and shared users can add, edit, reorder and delete prompts again.

backend/permissions/permission.py
```diff
@@ -159,6 +190,27 @@ class IsParentToolOwner(permissions.BasePermission):
...
+        if _is_service_account(request) or _is_organization_admin(request):
+            return True
+        # Imported here: the models pull in this module at import time.
+        from prompt_studio.prompt_profile_manager_v2.constants import ProfileManagerKeys
+        from prompt_studio.prompt_studio_core_v2.models import CustomTool
+
+        tool = CustomTool.objects.filter(
+            tool_id=request.data.get(ProfileManagerKeys.PROMPT_STUDIO_TOOL)
+        ).first()
+        return bool(tool and _is_resource_owner(request.user, tool))
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Valid at the time, and fixed in 14df8ab01.

Profile creation follows project access now, not ownership — `ProfileManagerView` uses the new `ParentToolAccess` instead of `IsParentToolOwner`. The frontend gating on Add New LLM Profile is gone too, along with the edit, delete and default-profile gates.

The `has_permission` on create stays, but checks access rather than ownership: `create` is collection-level, so without it the route was open to the whole organisation rather than to collaborators.

frontend/src/components/pipelines-or-deployments/pipelines/PipelineCardConfig.jsx
```diff
@@ -324,12 +325,19 @@ function createPipelineCardConfig({
...
+                      ? "Disable pipeline"
+                      : "Enable pipeline"
+                    : "Only the owner can change this"
+                }
               >
                 <Switch
                   size="small"
                   checked={pipeline.active}
                   data-testid={`pipeline-toggle-${pipeline.id}`}
+                  disabled={!canEditResource(pipeline, sessionDetails)}
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Not fixing — the switch is intended to be owner-only.

The premise that "status management is meant to remain available separately from settings edits" is not the rule we settled on. Enabling or disabling a pipeline is a change to someone else's resource, so it sits with the owner alongside rename, delete and cron.

The backend already agreed: `pipeline_v2/views.py` puts `partial_update` behind `IsOwner`, and `api_deployment_views.py` does the same. A shared user pressing the switch got an optimistic flip followed by a revert and an error toast — on ETL pipelines it surfaced as "Api deployment not found". Disabling the control removes a dead end rather than removing a capability.

backend/prompt_studio/permission.py
```diff
@@ -11,6 +11,24 @@
...
+    the project's prompts and settings, the same as its owner. Only the
+    project's name, its existence and who else it is shared with stay with
+    the owner, and those are gated on the project viewset.
+    """
+    if _is_resource_owner(user, tool):
+        return True
+    if _is_resource_viewer(user, tool):
+        return True
+    if has_group_access(user, tool):
+        return True
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Not a defect — this is the intended behaviour as of 14df8ab01, and the reason the earlier commit was reverted.

The finding assumes shared users are "read-only recipients". That was the original proposal, and the product call on UN-2868 rejected it for Prompt Studio: a project shared read-only is not usable by a team, so a Prompt Studio share is a collaboration share. Shared users are meant to change prompts, settings and profiles. Adapters, connectors, workflows and deployments keep the read-only treatment you are describing.

Worth noting you flagged the opposite four hours earlier on the previous commit — "Shared prompt edits blocked" and "Shared profile creation blocked", on these same lines. Both directions cannot be defects; the intent moved, and this is the side it moved to.

What does stay owner-only, on every resource including this one: renaming, deleting, and deciding who else gets access. Renaming a project is enforced per-field in `CustomToolSerializer.validate_tool_name`, because the name and the settings share one PATCH endpoint.

Two checks in this diff are genuine tightenings rather than relaxations, and they stay:
- `reorder_prompts` is routed without a pk, so DRF ran no object check at all — any organisation member could reorder any project's prompts. It now resolves the prompt and calls `check_object_permissions`.
- `ParentToolAccess.has_permission` gates profile `create`, which is collection-level and was previously reachable by the whole organisation.

kirtimanmishrazipstack:
@greptile-apps please review again.

kirtimanmishrazipstack:
@greptile-apps Requesting a re-review. One clarification first, on the part of this PR most likely to read as an over-block.

**Blocking the workflow destination settings for shared users is intentional, and it deliberately includes HITL.**

The destination connector modal carries three things behind a single Save: the connector configuration (DB table, credentials), the HITL review rules and percentage, and the sync destination plus TTL. All three configure the same workflow endpoint — HITL is not a separate review surface bolted onto a settings screen, it is part of how that workflow writes its output.

So the whole modal is owner-only on a shared workflow, and the three viewsets are gated identically:

- `WorkflowEndpointViewSet` (OSS) uses `WorkflowOwnerMutationMixin`
- `RuleEngineViewSet` and `SettingsViewSet` (cloud `manual_review_v2/views.py:2831,2980`) use the same mixin

Gating the connector half and leaving the HITL half open would have been the inconsistency.

**What a shared user keeps:** opening the modal and reading every value, and running the workflow. **What they lose:** repointing someone else's pipeline at a different table, and changing what fraction of its output goes to human review.

This is intentionally different from Prompt Studio in the same PR. A Prompt Studio project is shared *for collaboration*, so prompts, settings and LLM profiles stay editable. A workflow is shared *for use*. The full per-resource split is the table in the How section of the description. On every resource, renaming, deleting and deciding who else gets access stay with the owner.

Both directions of this were flagged earlier in the review and then withdrawn once the intent was clear, so flagging it again either way is expected to be a false positive.

backend/workflow_manager/endpoint_v2/views.py
```diff
@@ -11,8 +12,19 @@
...
 
-class WorkflowEndpointViewSet(viewsets.ModelViewSet):
+class WorkflowEndpointViewSet(WorkflowOwnerMutationMixin, viewsets.ModelViewSet):
+    """Workflow source / destination endpoints.
+
+    Config here selects the connector and its settings -- the destination
+    folder for filesystem, the table for database. Shared users may read
+    it; only owners and org admins may change it.
+    """
+
```
chandrasekharan-zipstack, line 22: **[Critical] [Lens 4 — Security]** — `workflow_endpoint_list` has no permission class and never calls `get_object()`; any org member reads any workflow's connector credentials.

This new docstring states "Shared users may read it", which reads as a class-wide invariant. It holds for `update`/`partial_update`/`destroy` and not for the list action.

`workflow_endpoint_list` (`views.py:88-101`, routed at `workflow_manager/urls.py:14`) calls `WorkflowEndpointUtils.get_endpoints_for_workflow(pk)` directly, bypassing both `get_queryset()` — the only access filter on this viewset — and object permissions. `WorkflowHelper.get_workflow_by_id` is a bare `Workflow.objects.get(pk=id)` with no scoping, and `DEFAULT_PERMISSION_CLASSES` is `[]` (`backend/backend/settings/base.py:640`), so `get_permissions` returning `list(super().get_permissions())` for this action yields nothing.

The response serializes through `ConnectorInstanceSerializer`, whose `to_representation` pops only `CONNECTOR_AUTH` and blanks `connector_metadata` only for `UnstractCloudStorage` (`connector_v2/serializers.py:162-176`). S3/GCS/DB `connector_metadata` is returned decrypted to a user with no relationship to the workflow.

**Failure mode:** `GET /api/v1/unstract/<org>/workflow/<any-uuid>/endpoint/` — horizontal privilege escalation inside an org. No error is raised or logged, because the authorization that should have refused is simply never evaluated.

*Pre-existing, but this PR is what adds the docstring asserting the opposite.*

**Fix:**
```python
workflow = get_object_or_404(Workflow.objects.for_user(request.user), pk=pk)
endpoints = WorkflowEndpointUtils.get_endpoints_for_workflow(workflow.id)
```

Confidence: High (verified against the PR tree).

backend/permissions/permission.py
```diff
@@ -148,6 +149,36 @@ def has_object_permission(self, request: Request, view: APIView, obj: Any) -> bo
...
+    """
+
+    mutation_denied_message = (
+        "Only the workflow owner or an organization admin can change this."
+    )
+
+    def get_permissions(self) -> list[Any]:
+        if self.action in ("update", "partial_update", "destroy"):
+            return [IsParentWorkflowOwner()]
+        return list(super().get_permissions())
```
chandrasekharan-zipstack, line 171: **[High] [Lens 4 — Security]** — The gate validates the *current* parent, never the *target* parent, and `workflow` is writable.

`IsParentWorkflowOwner.has_object_permission` resolves ownership from `obj.workflow` — the parent the row has now. `WorkflowEndpointSerializer` is `fields = "__all__"` with only `workflow_name` marked read-only, so `workflow` itself is a writable FK.

**Failure mode:** the owner of workflow A sends `PATCH /workflow/endpoint/<A-endpoint>/` with `{"workflow": "<B-id>"}`. The object check passes (they do own A), then the save reassigns the row to B, owned by a different user in the same org. B's owner now executes against A's connector instance, bucket, or table.

That is the customer-reported bug — a shared user repointing another team's workflow at a different output folder — still reachable after the fix that targets it. The same shape applies to the cloud companion: `RuleEngineSerializer.read_only_fields` is `["id", "created_by", "modified_by"]` and `SettingsSerializer` is `fields = "__all__"`, so both HITL viewsets inherit it.

**Fix:** mark `workflow` read-only on `WorkflowEndpointSerializer` (and on the two cloud HITL serializers), or validate in the mixin that `validated_data.get("workflow")` equals `instance.workflow` on update.

Two notes in this class's favour, verified while checking: `perform_create` correctly fails **closed** on a missing `workflow`, and the MRO order is right at every current call site. The base-order fragility (a bare mixin overriding `get_permissions`/`perform_create`, which DRF also defines) is a trap for the next viewset that copies it, not a live bug — worth an `__init_subclass__` assertion given this is now a cross-repo contract.

Confidence: High on the field config (verified); the reassignment PATCH was not executed against a live DB.

frontend/src/components/navigations/tool-nav-bar/ToolNavBar.jsx
```diff
@@ -80,14 +82,23 @@ function ToolNavBar({
...
+                    <Button
+                      type="text"
+                      size="small"
+                      icon={<Pencil />}
+                      className="tool-nav-bar__edit-icon"
+                      onClick={onEditTitle}
+                      disabled={editTitleDisabled}
+                      aria-label="Edit title"
+                    />
+                  </Tooltip>
```
chandrasekharan-zipstack, line 101: **[High] [Lens 1 — Spec & intent]** — Every "hover for the reason" tooltip added on a `disabled` control never renders.

The PR's stated UX is "locked controls greyed out with a reason on hover, not hidden." The shim `Tooltip` renders `<TooltipTrigger asChild>` (Radix) and depends on pointer/focus events reaching the child. The shim `Button` carries `disabled:pointer-events-none` — asserted by this repo's own test at `frontend/src/components/ui/cascade-and-affordances.test.jsx:194-199` — and a native `disabled` button or switch swallows `mouseenter`. The tooltip content never mounts.

Affected, all added by this PR:
- `ToolNavBar.jsx:85-101` (this hunk)
- `CardFieldComponents.jsx:73-81` and `:113-121`
- `ManageKeys.jsx:264-272` and `:309-317`
- `ApiDeploymentCardConfig.jsx:99-113`
- `PipelineCardConfig.jsx:327-341`

**User impact:** 6 of the 8 new lock affordances communicate nothing — a dead grey control with no explanation, which is the "the screen said nothing" outcome this PR exists to eliminate. `ResourceTable.jsx:302-315` works only because it uses `aria-disabled` on a raw `<button>` rather than `disabled`, so the same feature behaves two different ways.

The codebase already documents this exact trap and its fix at `frontend/src/components/logging/log-modal/LogModal.jsx:262-267`:
> *"span wrapper — antd disabled Buttons swallow pointer events so a direct Tooltip child never receives mouseenter and the disabled-state tooltip never shows."*

**Fix:** wrap each disabled child in a `<span>` inside the `Tooltip`, or fix it once in the shim — when `title` is set and the child has `disabled`, render the trigger wrapped.

Confidence: High (mechanism confirmed by the repo's own test and precedent comment; UI not run).

backend/prompt_studio/prompt_profile_manager_v2/views.py
```diff
@@ -29,10 +27,10 @@ class ProfileManagerView(viewsets.ModelViewSet):
     serializer_class = ProfileManagerSerializer
 
     def get_permissions(self) -> list[Any]:
-        # Mutations require ownership of the parent tool (creator + co-owners);
-        # reads honor sharing.
+        # A profile is part of the project's design, so anyone the project is
+        # shared with manages it (UN-2868); reads honor sharing.
```
chandrasekharan-zipstack, line 31: **[High] [Lens 13 — Testing / Lens 4 — Security]** — This swap leaves a test asserting the inverse of production, still passing.

Switching to `ParentToolAccess` widens ProfileManager `create`/`destroy`/`partial_update`/`update` from owner-only to any VIEWER-role member (`prompt_studio/permission.py:24`, via `_can_access_tool`). That is the intended collaborative rule for Prompt Studio — no objection to the behaviour.

The problem is what it leaves behind. After this diff `IsParentToolOwner` has **zero** production callers (`grep -rn IsParentToolOwner backend --include='*.py'` returns only its definition, one test, and two docstrings), but `backend/permissions/tests/test_owner_management.py:326-369` still constructs `IsParentToolOwner()` directly and asserts `assertFalse(self._perm(self.viewer, child))` at `:359-362`.

**Failure mode:** CI reports green on ProfileManager mutation authz while the shipped endpoint returns 200 for exactly that user. The live gate, `ParentToolAccess`, is untested in both directions — so any future re-tightening or further loosening is invisible to the suite. A dead authz class that still has a passing test suite is also a trap for the next person who imports it by name; `prompt_studio/permission.py:86` still points readers at it as the pattern to mirror.

**Fix:** retarget `IsParentToolOwnerTests` at `ParentToolAccess` (flip viewer to `assertTrue`, keep outsider `assertFalse`, add a service-account case) and delete `IsParentToolOwner`.

Confidence: High.

backend/prompt_studio/prompt_studio_core_v2/serializers.py
```diff
@@ -124,7 +132,19 @@ class Meta:
...
+        # (UN-2868).
+        request = self.context.get("request")
+        if not self.instance or not request or value == self.instance.tool_name:
+            return value
+        user = request.user
+        if getattr(user, "is_service_account", False):
+            return value
+        if not self.instance.is_owner(user) and not is_org_admin(user):
+            raise ValidationError("Only the owner can rename this project.")
+        return value
```
chandrasekharan-zipstack, line 147: **[High] [Lens 7 — API contract / Lens 4 — Security]** — An authorization decision expressed as field validation: wrong status, wrong shape, and bypassable.

`raise ValidationError("Only the owner can rename this project.")` renders as `400 {"tool_name": ["Only the owner can rename this project."]}`. The sibling decision for the same resource — delete — goes through `IsOwner` (`views.py:160`) and returns `403`. Same actor, same resource, same predicate, two status codes and two response shapes.

Three consequences:
1. **Clients cannot distinguish** "that name is invalid" from "you are not allowed." Any form-error handler paints an authz refusal under the name input as though retyping would fix it.
2. **The rule is invisible to permission introspection** — `view.get_permissions()` reports rename as ungated, so the browsable API, schema generation, and any permission-walking test all read it as open.
3. **The guard is conditional on payload shape.** DRF runs a field validator only when the field is present and only through this serializer. Any sibling/import serializer that writes `tool_name`, a future `.save(tool_name=…)` helper, or a refactor folding this into `validate_name_field` silently drops the check.

Also worth deciding deliberately: this serializer is `fields = "__all__"`, and the gate covers exactly one field. `description`, `icon` and `author` remain writable by any shared user, which the comment at `views.py:155-156` does not acknowledge.

**Fix:** `raise rest_framework.exceptions.PermissionDenied(...)` — same placement, correct status, and it reads as authorization to anyone scanning the file. The cloud companion has the identical issue at `agentic_studio_v1/serializers.py:188`.

Confidence: High.

frontend/src/helpers/resourceAccess.js
```diff
@@ -0,0 +1,21 @@
...
+ * `is_owner` is set by every shareable resource's serializer.
+ */
+function canEditResource(resource, sessionDetails) {
+  // Payload not in yet. The backend still refuses the write, so assume
+  // editable rather than flash a read-only view at the resource's own owner
+  // while the request is in flight.
+  if (!resource || resource.is_owner === undefined) {
+    return true;
+  }
+  return Boolean(resource.is_owner || sessionDetails?.isAdmin);
```
chandrasekharan-zipstack, line 19: **[High] [Lens 3 — Correctness]** — `canEditResource` fails **open**, and cannot distinguish "still loading" from "this serializer does not emit the field".

`is_owner === undefined → true`. The comment justifies this as the in-flight-request window, but the predicate has no way to tell that apart from a payload that simply lacks the key.

Two concrete exposures:
- `APIDeploymentSerializer` — the non-list class returned for retrieve/create/update (`api_v2/api_deployment_views.py:305-308`) — does not carry `is_owner`. Any caller that seeds a store or table row from a non-list response unlocks every Edit/Delete/toggle on it.
- Store defaults are `details: {}` (`store/workflow-store.js:12`, `store/custom-tool-store.js:17`), which also takes the fail-open branch.

Worse, the two ends of this contract fail in **opposite** directions. Every producer emits `is_owner: false` when `self.context.get("request")` is missing (`prompt_studio_core_v2/serializers.py:123`, `workflow_v2/serializers.py:81`), so a serializer constructed without context silently demotes the resource's *owner* to read-only — and there are already two such call sites in tree (`workflow_v2/views.py:347`, `endpoint_v2/views.py:101`). One optional boolean is carrying three states (absent / false-because-no-context / false-because-genuinely-not-owner) with two collapsed and the third inverted at the boundary.

**Failure mode:** a serializer regression that drops `is_owner` produces no error, no warning, and no visible symptom — the UI just stops enforcing, silently restoring the pre-PR behaviour this change exists to prevent.

**Fix:** take an explicit `loading` argument rather than overloading `undefined`, fail closed on a fetched resource with no `is_owner`, and warn once so the gap is visible. On the producer side, either omit the key entirely when there is no request (so absent uniformly means unknown) or assert — a serializer emitting an authorization flag with no requester is always a bug.

Confidence: High that the branch exists; Medium that a live call site reaches it today.

frontend/src/components/widgets/resource-table/ResourceTable.jsx
```diff
@@ -288,21 +289,29 @@ function ResourceTable({
   const renderActions = (item) => {
     const deprecated = item?.is_deprecated;
     const disabledTitle = deprecated ? "This adapter is deprecated" : "";
+    // Sharing grants read only: no edit, no delete. Both controls stay on
+    // screen but disabled, so it is obvious they exist and why they are not
+    // available. Sharing onward stays open to shared users.
+    const canEdit = canEditResource(item, sessionDetails);
+    const locked = !canEdit;
+    const lockedTitle = "Only the owner can change this";
```
chandrasekharan-zipstack, line 300: **[High] [Lens 3 — Correctness]** — Frictionless adapters become undeletable, and the tooltip gives a reason that is false for them.

The new gate is `canEditResource(item, sessionDetails)` — i.e. `is_owner || isAdmin`. But the backend deliberately lets **any** org member delete a frictionless adapter: `AdapterInstanceViewSet.get_permissions` routes `destroy` to `IsFrictionLessAdapterDelete`, which returns `True` unconditionally when `obj.is_friction_less` (`adapter_processor_v2/views.py:171-172`, `permissions/permission.py:304-312`).

Those adapters are visible to every member (`adapter_processor_v2/models.py:54-59`, `Q(is_friction_less=True)`) and their owner is masked as "Unstract" in the payload (`adapter_processor_v2/serializers.py:194-196`), so `is_owner` is `False` for essentially everyone. `AdapterListSerializer` does not expose `is_friction_less` at all (`serializers.py:165-176`), so the frontend cannot special-case these rows even if it wanted to.

**User impact:** the only UI route to remove platform-provisioned onboarding adapters disappears for every non-admin, and the tooltip asserts an ownership rule that does not govern the row.

**Fix:** add `is_friction_less` to `AdapterListSerializer.Meta.fields` and gate delete on `canEdit || item?.is_friction_less` for adapters. Keep Edit disabled — `IsFrictionLessAdapter` correctly refuses update/retrieve on those rows for everyone.

Confidence: High.

backend/prompt_studio/permission.py
```diff
@@ -11,6 +11,24 @@
 from tenant_account_v2.organization_member_service import OrganizationMemberService
 
 
+def _can_access_tool(user: Any, tool: Any) -> bool:
+    """Whether ``user`` may work on ``tool``.
+
+    Prompt Studio is shared for collaboration (UN-2868): a shared user edits
+    the project's prompts and settings, the same as its owner. Only the
+    project's name, its existence and who else it is shared with stay with
+    the owner, and those are gated on the project viewset.
```
chandrasekharan-zipstack: **[High] [Lens 16 — Doc accuracy]** — This docstring states an owner-only rule the code does not enforce, on the sharing surface.

The comment asserts that "who it is shared with" stays with the owner. It does not. The `share` action is not in the `IsOwner()` list in `prompt_studio_core_v2/views.py:159` (which gates only `destroy`, `add_co_owner`, `remove_co_owner`), so it falls through to `IsOwnerOrSharedUserOrSharedToOrg`, and `ShareAuthorizationService` explicitly permits non-owners to add — *"direct shared users and group members can add (with scope limits) but not remove"* (`tenant_account_v2/sharing_helpers.py:381-389`).

This PR ships **both** statements of the rule. The frontend comments added in the same diff say the opposite and are correct: `CardFieldComponents.jsx:82-83` ("Sharing stays open to shared users: they may pass access on to a group they belong to, or to a user in the same organisation") and `ResourceTable.jsx:291`.

A maintainer auditing the share surface from this docstring will believe an owner-only gate exists where none does — the over-grant direction.

**Fix:** reword to match `ShareAuthorizationService` — only the project's *existence* (destroy) and co-ownership are owner-only; shared users may add viewers/groups but not remove them.

Two related accuracy issues in this file, same lens:
- `:19-20` says the rename is "gated on the project viewset". It is not — this PR deliberately put it in the serializer (`prompt_studio_core_v2/serializers.py:136-138` says so explicitly). Someone verifying "is rename protected?" by reading the viewset finds nothing.
- `:62` says `# Imported here: the models pull in this module at import time.` No model imports `prompt_studio.permission` — only three view modules and one test do. The stated reason for the non-obvious deferred import is false, so the next engineer either hoists it on a bad premise or preserves a workaround nobody can justify.

Confidence: High.

backend/prompt_studio/prompt_studio_core_v2/views.py
```diff
@@ -152,6 +152,10 @@ def get_serializer_class(self):
         return CustomToolSerializer
 
     def get_permissions(self) -> list[Any]:
+        # Settings are collaborative (UN-2868); only the project's existence
+        # and who it is shared with stay with the owner. Renaming is blocked
+        # per-field in the serializer, since it shares an endpoint with
+        # every settings write.
```
chandrasekharan-zipstack, line 158: **[High] [Lens 4 — Security]** — Org-wide-shared projects get a split rule this comment does not describe, and the frontend cannot represent it.

`_can_access_tool` (`prompt_studio/permission.py:14-29`) admits owner / direct viewer / group member / org admin — but **not** `shared_to_org`. `IsOwnerOrSharedUserOrSharedToOrg`, used here for `update`/`partial_update` and for profile reads at `prompt_profile_manager_v2/views.py:34`, **does** admit it.

**Failure mode:** a user whose only access is an org-wide share can PATCH the project's settings and read its profiles, but gets 403 on every prompt edit (`PromptAcesssToUser`, `permission.py:41-44`) and every profile write. Half the collaborative surface works and half refuses, for the same user on the same project.

The frontend cannot express that state: `usePromptStudioCanEdit` collapses to a single `is_owner` boolean, so this user is shown one consistent affordance that is wrong for half the actions.

The class docstring at `prompt_studio/permission.py:17-18` states the opposite of the code — "a shared user edits the project's prompts and settings, the same as its owner."

**Fix:** decide whether `shared_to_org` is inside or outside the collaborative set and apply it to both sides. Then pin it: (direct viewer, group, org-wide, unrelated member) × (settings PATCH, prompt PATCH, profile POST) is nine cheap API assertions that would have caught this without hand testing.

Confidence: High.

chandrasekharan-zipstack (commented):
## Standardized PR Review — UN-2868 (OSS half)

**Verdict: BLOCK** · Mode: INITIAL · rubric: `unstract` plugin v0.18.1 (16 lenses)

**Summary** — Critical: 2 · High: 11 · Medium: 12 · Low: 5 · Lenses run: 16/16

Posting **Critical and High only**, as requested. Medium/Low findings are held back and can be posted on request.

This PR is a real improvement on the status quo. It is blocked because the new gate ships alongside a docstring asserting an invariant the code does not hold — which is the specific failure mode that makes a permission change dangerous, since the next reader stops re-checking.

Reviewed **as if this PR merges first**; the cloud companion (Zipstack/unstract-cloud#1751) imports `WorkflowOwnerMutationMixin` and `canEditResource` from here. Cross-repo claims were verified against both PR trees, not against `main`.

---

### Unanchored findings

These sit outside this PR's diff, so GitHub has no line to attach them to. They are included because the PR's stated contract ("backend enforces; frontend gates") depends on them.

**[Critical] [Lens 4] — `tool_instance` create is ungated; the new Agency lock is frontend-only.**
`backend/tool_instance_v2/views.py:113-118` gates only `update`/`partial_update`/`destroy`. `ToolInstanceSerializer.create` (`backend/tool_instance_v2/serializers.py:257-260`) resolves `Workflow.objects.get(pk=workflow_id)` with no user scoping, and the same path flips `workflow.is_active = True` on that foreign workflow. Any authenticated org member can POST a tool onto a workflow they do not own. This PR disables the corresponding UI control at `Agency.jsx:1166`/`:1176`, so the lock is presentation-only here.
*Pre-existing, not introduced by this PR.* Fix: inherit `WorkflowOwnerMutationMixin` on `ToolInstanceViewSet` (deleting its duplicate `get_permissions`) and scope the workflow lookup through `for_user`.

**[High] [Lens 4] — File-history deletion is still open to read-only shared users.**
`backend/workflow_manager/workflow_v2/file_history_views.py:26` uses `IsWorkflowOwnerOrShared`, which admits `_is_resource_viewer` and `shared_to_org`. Both `destroy` (`:85`) and `clear` (`:99`) are destructive. Separately, `clear_file_marker` is **not** in the `IsOwner()` action list (`workflow_manager/workflow_v2/views.py:91-98`), so it falls through to `IsOwnerOrSharedUserOrSharedToOrg` — and it is exposed as a **GET** (`views.py:355`), making it reachable by prefetch or a pasted URL with no CSRF in the way. The PR title says "block deletion on resources shared with a user"; these are the most destructive per-workflow deletion paths and they are not blocked.
Fix: gate both on `is_workflow_mutator` (already exported at `permissions/permission.py:121`), and change `clear_file_marker` to POST.

**[High] [Lens 1] — The workflow builder's connector-type dropdown was left ungated and now fails.**
`frontend/src/components/agency/ds-settings-card/DsSettingsCard.jsx:238-251` is `disabled={!allowChangeEndpoint}` only — never `canEdit`. It fires `handleEndpointUpdate` straight at the endpoint this PR newly gates, so a shared user picks a connector type and gets a bare "Failed to update" toast. The `ConfigureConnectorModal` behind it was gated; its launcher was not. `useWorkflowCanEdit` is imported in four files but not this one.

**[High] [Lens 13] — Zero test files across 24 changed files, on a diff that is almost entirely authorization logic.**
`ParentToolAccess` widens ProfileManager writes to shared viewers, `WorkflowOwnerMutationMixin` adds a new collection-level create gate, and rename enforcement moved from a permission class into serializer field validation. None of it is exercised. `backend/CLAUDE.md` states tests are required after implementation, and the harness already exists at `backend/permissions/tests/test_owner_management.py`. Highest-value cases: shared viewer denied on endpoint update/destroy; shared viewer *allowed* on ProfileManager create/update/destroy; shared viewer denied on a `tool_name` PATCH but allowed on any other settings field.

---

*Generated by the standardized 16-lens review. Verified findings only — 11 agent claims were dropped as false after checking both PR trees.*

backend/workflow_manager/endpoint_v2/views.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Fixed in 44a06b968, with a follow-up in 28cfde682 — and one half deliberately left open for your call.

`workflow_endpoint_list` now resolves through `Workflow.objects.for_user()` before serialising, so an unrelated org member gets a 404 instead of the workflow's connector credentials.

Chasing it turned up two more paths to the same credentials:

- **A second route to them.** `WorkflowEndpointSerializer.connector_instance_id` is a writable `PrimaryKeyRelatedField` whose queryset was `ConnectorInstance.objects.all()` — org-scoped, not user-scoped, unlike every sibling. Any member could create their own workflow (becoming its owner, so `IsParentWorkflowOwner` passes), PATCH their endpoint onto a colleague's connector, and read the decrypted metadata out of the PATCH response. Scoped with `for_user` in 28cfde682.
- **`WorkflowEndpointAPIView` logged `dict(request.headers)` at INFO** — including the internal service bearer key that authenticates it, on the view that returns decrypted connector metadata for any workflow in any organization. No redaction filter exists anywhere. Dropped the headers entry and lowered the line to debug.

**Still open — your decision.** Scoping bounds *who reaches the workflow*, not *what the response contains*. A user the workflow is shared with (directly, by group, or org-wide) still receives the owner's decrypted S3/GCS/Postgres credentials, because the nested `ConnectorInstanceSerializer` decrypts everything except `UnstractCloudStorage`. Redacting it changes what the read-only connector view shows shared users, so I didn't decide that alone. I did narrow the two comments that implied the exposure was closed.

backend/permissions/permission.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Half of this held and half didn't — and the half that held is now closed at eight sites, not one.

**The OSS half is disconfirmed.** `WorkflowEndpointSerializer.workflow` is not writable: `WorkflowEndpoint.workflow` is `editable=False` (`models.py:34`), which DRF honours by forcing `read_only=True`. I checked by building the serializer rather than reading `Meta` — the only writable fields are `configuration`, `connection_type` and `connector_instance_id`. So the reparenting PATCH you describe can't land there. You'd flagged your own uncertainty on exactly this ("the reassignment PATCH was not executed against a live DB"); the field-config half is the part that was off.

**The cloud half was live and is fixed** (1d9c4f6a). Both `RuleEngineSerializer.workflow` and `SettingsSerializer.workflow` built `read_only=False`. One correction worth recording: only `SettingsSerializer` was actually exploitable — `RuleEngineSerializer` has a custom `update()` that assigns only `percentage`/`rule_string`/`rule_json`/`rule_logic` and never touches `workflow`. The guard there is defence-in-depth, not a closed hole, and I've said so in the commit so nobody removes it later believing it was load-bearing.

**The class is wider than either of us listed.** Sweeping by *gate* rather than by field found eight sites: RuleEngine, HITLSettings, ToolInstance, ProfileManager, APIKey (`api` and `pipeline`), ToolStudioPrompt, and WorkflowEndpoint — the last safe only by that `editable=False`. All closed across 1d9c4f6a / 28cfde682 / ac3151056 / c67c55c8d.

Two of those are worth your attention:

- **`APIKeySerializer.api`/`.pipeline` were writable on PATCH.** `create` spends ~60 lines refusing a path-vs-body mismatch; `update` had nothing, so a live key could be retargeted onto a deployment the caller doesn't own.
- **The eighth site was an alias of a field I'd already guarded in the same file.** `ToolInstance.workflow`'s attname *is* `workflow_id`, and the serializer declares a writable `workflow_id` alias. `validate_workflow` fires only on the key `workflow`; a body of `{"workflow_id": ...}` skipped it entirely and DRF's `setattr(instance, "workflow_id", uuid)` wrote the FK column directly.

That last one is the argument for your `__init_subclass__` suggestion, or better: hoisting the check into `WorkflowOwnerMutationMixin` so it covers attname aliases. Enumerating this class by field name provably cannot close it. I did **not** do that refactor — it touches eight viewsets across both repos and I didn't want it unreviewed at the end of an automated run. Recommended in the commit message.

frontend/src/components/navigations/tool-nav-bar/ToolNavBar.jsx
kirtimanmishrazipstack: @chandrasekharan-zipstack Fixed in d592324e4, then corrected in 81f2ceaae — your diagnosis was right and my first fix was incomplete.

Fixed once in the shim rather than at six call sites: `Tooltip` now wraps a `disabled` child so pointer events land on the wrapper.

The correction matters. A span alone does **not** restore hover — the browser suppresses mouse events on a disabled control and doesn't bubble them, so the span only becomes the hit target once the child stops hit-testing. `ui/button.tsx` sets `disabled:pointer-events-none` and worked; `switch`, `checkbox`, `input`, `label`, `radio-group`, `select` and `textarea` set only `cursor-not-allowed` and did **not**. That includes the live case this PR is about — the owner-only `Switch` in `ManageKeys.jsx:264`. So my first pass silently helped only the Buttons. The wrapper now neutralises the child itself, and carries `tabIndex` so the tooltip is reachable by keyboard at all, which a disabled child never is.

One correction to the finding: `ResourceTable`'s own action buttons use `aria-disabled`, not `disabled`, and `aria-disabled` elements receive pointer events normally — so those two tooltips already worked. The beneficiaries are the `Button`/`Switch` call sites.

backend/prompt_studio/prompt_profile_manager_v2/views.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Not fixed — blocked by a project rule, and I'd like your explicit go.

The finding is correct: after this diff `IsParentToolOwner` has zero production callers, while `test_owner_management.py:326-369` still constructs it directly and asserts `assertFalse(self._perm(self.viewer, child))`. CI is green on ProfileManager mutation authz while the shipped endpoint returns 200 for exactly that user, and the live gate `ParentToolAccess` is untested in both directions.

It can't be half-done: deleting the dead class breaks the test that still imports it, so the fix necessarily edits a test. `CLAUDE.md` says *"Never write, modify, or suggest tests unless the user explicitly uses the word 'test.'"* — so I stopped rather than work around it.

Say the word and it's one pass: retarget `IsParentToolOwnerTests` at `ParentToolAccess` (viewer flips to `assertTrue`, outsider stays `assertFalse`, add a service-account case) and delete the class.

Same blocker applies to test coverage for the four behaviour changes in this batch — the refusal paths (404 on a non-member endpoint list, the reparent refusals, 403-not-400 on rename, the delete-button state) have no assertions anywhere.

backend/prompt_studio/prompt_studio_core_v2/serializers.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Fixed in 44a06b968, both repos.

`PermissionDenied` instead of `ValidationError`, so the rename refusal returns 403 with the same shape as its sibling `destroy`. Verified against the installed DRF 3.17.1 and drf-standardized-errors 0.15.0 rather than from memory: `to_internal_value` catches only `ValidationError`/`DjangoValidationError`, so `PermissionDenied` propagates out of `is_valid()` and renders 403 — and `middleware/exception.py`'s only rewrite is gated on 404, so it passes through unmodified.

Your point 1 was the concrete one: `AddCustomToolFormModal.jsx:132` does `getBackendErrorDetail("tool_name", backendErrors)`, so the authz refusal was being painted under the name input as though retyping would fix it. That's gone. Checked the blast radius too — the axios interceptor logs out on 401 only, `useExceptionHandler` has a `client_error` branch that surfaces `errors[0].detail`, and the rename submit passes no `setBackendErrors`, so nothing depended on the old `attr: "tool_name"` binding.

Cloud twin at `agentic_studio_v1/serializers.py` fixed the same way in 1d9c4f6a.

Your closing note — that the gate covers one field while `description`, `icon` and `author` stay writable by any shared user — is untouched and still stands as a deliberate decision to make.

frontend/src/helpers/resourceAccess.js
kirtimanmishrazipstack: @chandrasekharan-zipstack Partly fixed in 28cfde682. The concrete exposure is closed; the contract itself is not, and that's deliberate.

**Closed:** `APIDeploymentSerializer` — the class returned for retrieve/create/update — now carries `is_owner`, so a consumer seeding a store from a non-list response no longer takes the fail-open branch.

**Not changed:** `is_owner === undefined → true` itself. Your fix (an explicit `loading` argument, fail closed on a fetched resource, warn once) is the right shape, but it inverts the default for eight call sites and both store defaults — `details: {}` in `workflow-store.js` and `custom-tool-store.js` both currently take the fail-open branch. Flipping that at the end of an automated run, with no test coverage on any of those call sites, is how an owner ends up locked out of their own resource. It wants doing deliberately with the tests to back it.

Your observation that the two ends fail in *opposite* directions is the part I'd most want preserved in whatever we do: producers emit `is_owner: false` when `self.context.get("request")` is missing, so a serializer built without context demotes the resource's own owner — and there are live no-context call sites. One optional boolean carrying three states, two collapsed and the third inverted at the boundary.

frontend/src/components/widgets/resource-table/ResourceTable.jsx
kirtimanmishrazipstack: @chandrasekharan-zipstack Fixed in 44a06b968, exactly as you specified.

`is_friction_less` added to `AdapterListSerializer.Meta.fields`, and the table's delete gate is now `locked && !item?.is_friction_less`. Edit stays locked — `IsFrictionLessAdapter` correctly refuses update/retrieve on those rows for everyone.

Checked the leak question before adding the field: `AdapterListSerializer` carries no `adapter_metadata`, and its `to_representation` already masks `created_by_email`/`owner_emails` to "Unstract" for frictionless rows, so the flag was already inferable from the response.

Also traced the four cases so the carve-out can't accidentally unlock anything else: `is_owner` absent → unchanged; `is_owner:false` on a non-adapter row (`is_friction_less` undefined) → still locked; ordinary adapter → still locked; frictionless → unlocked. Workflows, ListOfTools and ConnectorsPage rows are unaffected.

One thing I'd flag back: a later review pass argued this puts an adapter-specific field inside a table that also backs Workflows, Prompt Studio and Tool Settings, and that a `canDelete` predicate passed by the adapters page would keep the knowledge where it belongs. I left it as you wrote it — worth a view if you disagree.

backend/prompt_studio/permission.py
kirtimanmishrazipstack: @chandrasekharan-zipstack All three fixed in 0c9fb44e9.

- The sharing claim is reworded to match `ShareAuthorizationService`: renaming, deleting and *removing* access stay with the owner; sharing onward does not.
- The rename line now points at `CustomToolSerializer.validate_tool_name` rather than claiming the viewset gates it.
- The deferred-import comment is **deleted rather than reworded**. I first replaced it with "importing these at module scope creates a cycle", then tested that claim before committing — there is no cycle. `prompt_studio_core_v2.models` imports cleanly without `prompt_studio.permission`, and the reverse order works too. Rather than ship a second unverified reason in place of the first, I removed the justification and left the local import alone.

Worth knowing: a later review pass argued that comment was load-bearing and should be restored. I kept it deleted, because the assertion it made is measurably false and I'd rather have no reason than a wrong one. If you know the real reason the import is deferred, that's worth a line.

backend/prompt_studio/prompt_studio_core_v2/views.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Fixed in d592324e4 — decided in, not out.

`_can_access_tool` now admits `shared_to_org`, so both sides of the collaborative surface agree. An org-wide-shared user gets the same answer from `_can_access_tool` and `IsOwnerOrSharedUserOrSharedToOrg` instead of 200 on settings and 403 on every prompt edit.

Reasoning for that direction rather than the other: the PR's stated rule is that Prompt Studio is shared *for collaboration*, and `IsOwnerOrSharedUserOrSharedToOrg` — already used for settings PATCH and profile reads on the same project — admits it. Narrowing that side instead would have taken settings access away from a group who have it today.

Verified both directions: an org-shared user with no membership, no group and no admin role is now admitted; an unrelated user on a non-shared tool is still refused.

The class docstring that stated the opposite is corrected in 0c9fb44e9. Your nine-assertion suggestion — (direct viewer, group, org-wide, unrelated) × (settings PATCH, prompt PATCH, profile POST) — is exactly the right pin and is blocked only by the `CLAUDE.md` test rule; see my reply on the `IsParentToolOwner` thread.

kirtimanmishrazipstack:
### Note on the red checks — none are from this PR

**`test (integration)`, `e2e`, `report`** fail before a single test runs. The rig cannot
pull the MinIO image while bringing infrastructure up:

```
MinioContainer("minio/minio:latest").start()
docker.errors.ImageNotFound: pull access denied for minio/minio,
repository does not exist or may require 'docker login'
```

It dies in `tests/rig/runtime.py` → `runtime.up()`, so no test from this branch is
reached. `report` is red only because it aggregates those two.

Not specific to this branch: the same failure is currently hitting **six other
branches** (UN-3494, UN-3853, UN-4078, UN-1031, fix/single-pass-profile-export), and
`main`'s last green run was 2026-09-10. It needs a registry/runner fix, not a code change.

**`test (unit)` passes**, which is the tier that actually exercises this PR.

Run locally against the full backend, both tiers: **928 unit passed**, **471 integration
passed** (28 skips are the pre-existing credential-gated external-DB connector tests).

backend/workflow_manager/endpoint_v2/views.py
chandrasekharan-zipstack: **FOLLOWUP → RESOLVED.** `endpoint_v2/views.py:103-108` gates `workflow_endpoint_list` on `Workflow.objects.for_user` (404 outside it); `endpoint_v2/serializers.py:47-60` blanks `connector_metadata` for non-mutators, which closes the half I had left open.

Two residuals, raised as new findings in this round's review rather than here: the list action itself has no test (reverting `:105` to the unscoped query leaves the suite green), and the redaction is keyed on the workflow rather than the connector.

backend/permissions/permission.py
chandrasekharan-zipstack: **FOLLOWUP → RESOLVED.** Reparent guards verified at `tool_instance_v2/serializers.py:55-66` (both `workflow` and the `workflow_id` alias), `api_v2/serializers.py:181-191`, `prompt_studio_v2/serializers.py:27-31`, `prompt_profile_manager_v2/serializers.py:31-35`; `WorkflowEndpoint.workflow` is `editable=False` so the endpoint serializer needs none.

None of the four guards has a test — dropping both tool-instance guards leaves 59 green. Folded into the new coverage finding. The structural point (invariant lives in hand copies, not in the gate) is noted as Medium in the summary; your deferral stands.

frontend/src/components/navigations/tool-nav-bar/ToolNavBar.jsx
chandrasekharan-zipstack: **FOLLOWUP → RESOLVED.** `ui/shims/antd-overlays.tsx:577-584` wraps the disabled child in a `[&>*]:pointer-events-none` span used as the Radix trigger, so hover lands on the wrapper. c349a635 removed only `tabIndex`; hover survives, keyboard loss is stated at `:571-576`. No RTL case for the disabled path — noted Low.

backend/prompt_studio/prompt_profile_manager_v2/views.py
chandrasekharan-zipstack: **FOLLOWUP → RESOLVED.** `IsParentToolOwner` is gone from both trees; `test_owner_management.py:363-368` now asserts the viewer is allowed, `:378-394` builds a real `DRFRequest` for the create path. Mutating `ParentToolAccess.has_permission` to `True` fails both. Viewset *routing* (`prompt_profile_manager_v2/views.py:32-33`) is still only pinned at class level — in the coverage finding.

backend/prompt_studio/prompt_studio_core_v2/serializers.py
chandrasekharan-zipstack: **FOLLOWUP → RESOLVED.** `prompt_studio_core_v2/serializers.py:146` raises `PermissionDenied`; `partial_update` does not wrap `is_valid`, so it renders 403. Pinned by `test_shared_user_gates.py:218-242` (mutation-verified).

One consumer regressed with the status change: `AddCustomToolFormModal.jsx:83-85` discards `handleException`'s return, and a 403 (unlike the old field-bound 400) is returned rather than set inline — the modal now fails silently. Raised as Medium in the summary.

frontend/src/helpers/resourceAccess.js
chandrasekharan-zipstack: **FOLLOWUP → WAIVED (reviewer).** Accepting the reasoning. Verified every serializer feeding a `canEditResource` consumer now emits `is_owner` (`workflow_v2/serializers.py:81`, `connector_v2/serializers.py:177`, `adapter_processor_v2/serializers.py:204`, `api_v2/serializers.py:50,530`, `pipeline_v2/serializers/crud.py:33`, `prompt_studio_core_v2/serializers.py:50,105`; cloud `agentic_studio_v1/serializers.py:91`, `lookups/serializers/definition.py:30,124`), and backend gates hold, so the residual on a future non-emitting endpoint is a 403 toast, not a breach. The docstring at `resourceAccess.js:4-7` should say the open default also covers "serializer does not emit the field", not only "still loading".

frontend/src/components/widgets/resource-table/ResourceTable.jsx
chandrasekharan-zipstack: **FOLLOWUP → RESOLVED.** `adapter_processor_v2/serializers.py:176` exposes `is_friction_less`; `ResourceTable.jsx:300,344,358` unlock Delete only for those rows.

backend/prompt_studio/permission.py
chandrasekharan-zipstack: **FOLLOWUP → RESOLVED.** `prompt_studio/permission.py:29-33` reworded, rename now points at the serializer, false import comment removed at `:76-77`.

backend/prompt_studio/prompt_studio_core_v2/views.py
chandrasekharan-zipstack: **FOLLOWUP → RESOLVED (code).** `prompt_studio/permission.py:39-40` admits `shared_to_org`; `workflow_v2/permissions.py:49-55` mirrors `for_user` including groups.

Two residuals: dropping the `shared_to_org` branch leaves 59 green (the (viewer, group, org-wide, unrelated) × (settings, prompt, profile) matrix we agreed on was not added — in the coverage finding); and `prompt_studio_core_v2/views.py:155-158` still says "who it is shared with stay with the owner", while `share` falls through to the collaborative gate and sharees can add. Medium in the summary.

backend/api_v2/serializers.py
```diff
@@ -56,6 +59,10 @@ class Meta:
             "shared_to_org": {"read_only": True},
         }
```
chandrasekharan-zipstack, line 62: **[Critical] [Lens 4 — Security] — API deployments and pipelines accept any org workflow as `workflow`; any org member can deploy and execute a colleague's private workflow.**

- Location: `backend/api_v2/serializers.py:43-60` (`fields = "__all__"`, `workflow` writable; `validate_workflow` at `:94-156` checks endpoint configuration only); `backend/api_v2/api_deployment_views.py:352-366` (`create` → `is_valid` → `perform_create`, caller becomes OWNER and receives an API key); `backend/pipeline_v2/serializers/crud.py:28-48` and `backend/pipeline_v2/views.py:148-170` (same shape).
- Failure mode: `POST /api/deployment/ {"workflow": "<colleague's unshared workflow id>", ...}` from any org member. DRF binds `workflow` through `Workflow._default_manager` — `WorkflowModelManager` is org-scoped (`workflow_v2/models/workflow.py:26`), and `for_user` is never applied here (`grep -n for_user backend/api_v2/serializers.py backend/pipeline_v2/serializers/crud.py` → none). The caller executes a workflow they cannot even list, with its connectors and adapters, at the owner's cost. Same via `PATCH` (retarget an existing deployment/pipeline) and `POST /pipeline/`. This is the exact "execute another team's workflow" grant the sharing model withholds; it survived the sweep because the sweep keyed on gates that authorise against the *stored* parent, and these two bindings authorise against the binding's own owner.
- Pre-existing, not introduced here — reported on the same basis as the tool_instance create finding last round. The `APIKeySerializer` fix at `:181-191` sits forty lines below the unscoped field.
- Suggested fix: scope per request the way `WorkflowEndpointSerializer.get_fields` (`endpoint_v2/serializers.py:35-45`) already does — `fields["workflow"].queryset = Workflow.objects.for_user(request.user)` (`.none()` without a request) in both serializers, plus a reparent refusal on update. Decide whether "shared for use" should admit deploying at all or only `is_workflow_mutator`; `for_user` is the floor. Two tests: outsider POST → 400 invalid pk; viewer POST → per decision.
- Confidence: High on the field/manager configuration (read); the request was not executed.

backend/workflow_manager/endpoint_v2/views.py
```diff
@@ -85,6 +100,10 @@ def workflow_endpoint_list(self, request: Request, pk: str) -> Response:
             Response: The HTTP response containing the serialized list of
                 endpoints.
         """
-        endpoints = WorkflowEndpointUtils.get_endpoints_for_workflow(pk)
-        serializer = WorkflowEndpointSerializer(endpoints, many=True)
+        # Scoped to workflows the requester can reach; the serializer redacts
+        # connector_metadata for anyone who may only read.
+        if not Workflow.objects.for_user(request.user).filter(pk=pk).exists():
```
chandrasekharan-zipstack, line 105: **[High] [Lens 13 — Testing / Lens 4 — Security] — Three of last round's security fixes are unpinned; reverting any of them leaves the suite green.**

- Location: `endpoint_v2/views.py:103-107` (this list scoping — last round's Critical); `workflow_v2/file_history_views.py:33-36` (destroy/clear owner gate); `workflow_v2/views.py:358` + `urls/workflow.py:25` (`clear_file_marker` GET→POST). Also unpinned: all four reparent guards (`tool_instance_v2/serializers.py:55-66` including the `workflow_id` alias you identified as the bypass, `prompt_profile_manager_v2/serializers.py:31-34`, `prompt_studio_v2/serializers.py:27-30`, `api_v2/serializers.py:181-190`); the `shared_to_org` admission (`prompt_studio/permission.py:39-40`); ProfileManager viewset routing (`prompt_profile_manager_v2/views.py:32-33`, tested only at class level).
- Failure mode: mutation runs against `test_shared_user_gates.py` + `test_owner_management.py` — revert this action to the unscoped query → 59 passed; drop the file-history gate → 59 passed; flip `clear_file_marker` back to GET → 59 passed; remove both tool-instance guards → 59 passed; drop `shared_to_org` → 59 passed; route ProfileManager mutations to `IsOwner` → no endpoint-level failure. The bespoke `for_user().filter(pk=pk).exists()` here has nothing structural keeping it in place, and the file's own docstring (`:8-10`) says it exists to catch exactly this class of gate.
- Suggested fix: in `SharedWorkflowEndpointTests`: outsider → `workflow_endpoint_list` → 404; viewer → list → every `connector_metadata == {}`. File-history: viewer DELETE → 403, viewer POST `clear` → 403, row still exists; `clear_file_marker` GET → 405, viewer POST → 403. One reparent PATCH per site (`{"workflow_id": <other owned workflow>}` → 400, FK unchanged). Org-shared-only user → `create_prompt` 201 and profile create 201; viewer → `ProfileManagerView.partial_update` → 200. The fixtures already exist in these classes.
- Confidence: High (mutations executed).

chandrasekharan-zipstack (commented):
## Standardized PR Review — FOLLOWUP (UN-2868, OSS half) @ 8d966aff

**Verdict** — `BLOCK` on one pre-existing Critical (below). If it is scoped out to a tracked ticket with a written waiver here, the verdict becomes `REQUEST CHANGES` on the coverage High.

**Since** `d6dbc0ed` → `8d966aff`. **Scope change:** YES — new `permissions/tests/test_shared_user_gates.py`, rewritten `test_owner_management.py`, gate classes reshaped; re-scanned backend gates/serializers, agency/pipeline/custom-tools frontend, tests. Tests executed locally on an isolated DB (59 pass) and mutation-tested per gate.

**Prior findings** — 9 inline: 8 `RESOLVED`, #6 (`canEditResource`) `WAIVED` by reviewer. Status replies are in each thread. Unanchored: tool_instance create `RESOLVED` (`tool_instance_v2/views.py:139-148`, pinned); file-history writes `RESOLVED` (`file_history_views.py:33-36`, `workflow_v2/views.py:358`); DsSettingsCard `RESOLVED` (`DsSettingsCard.jsx:24,234-246`); zero tests `PARTIALLY RESOLVED` (22 tests, 9 of 14 gates pinned — see High below).

**New:** Critical: 1 · High: 1 · Medium: 5 · Low: 4. Critical + High are inline below; Medium/Low are in the full report and can be posted on request.

### Unanchored findings
_none — both new Critical/High anchored inline._

**Lens checklist** — 1 See M · 2 See M · 3 See M · 4 See C1, H1, M · 5 Clean · 6 Clean · 7 Clean · 8 Clean · 9 Clean · 10 Clean · 11 Clean · 12 N/A — no prompt/model changes · 13 See H1 · 14 Clean · 15 See M · 16 See M.

**Open** — external callers of `GET …/clear-file-marker/` (python client, docs) now 405; the new tests are DB-bound (`integration` tier) and have not run in CI at this head; H3 (lookup-definition dormant grants) waived by maintainer on the cloud PR.

backend/api_v2/serializers.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Valid — fixed in ba4f559b7, both serializers.

Confirmed as described: `workflow` writable under `fields = "__all__"`, bound through the merely org-scoped default manager, and `validate_workflow` checking endpoint configuration only. The pipeline serializer had no workflow validation at all.

Both now scope the field in `get_fields` — the `WorkflowEndpointSerializer` pattern you pointed at — plus a reparent refusal matching the guards already in this branch.

**On the decision you left open: `is_workflow_mutator`, not the `for_user` floor.** A deployment is a persistent execution surface its creator owns and can share onward, so standing one up reads as an owner act rather than a use of a shared workflow. Co-ownership is the route for anyone else. It is one line in `mutable_workflows_for` if the product call goes the other way.

One thing worth recording: `resources_visible_via_memberships` grew an optional `role` so the helper reuses its varchar/UUID cast. My first attempt filtered `memberships__role` directly and died on `operator does not exist: uuid = character varying` — `object_id` is varchar, and `_object_id_subquery` is the only thing that casts it.

Tests: outsider and shared viewer refused on both deploy and schedule, owner unaffected. They assert the `does_not_exist` **code**, not just the `workflow` attr — asserting the attr alone passed with the gate removed, because this serializer also rejects an unconfigured workflow on the same attr. Your mutation discipline caught that one on me.

backend/workflow_manager/endpoint_v2/views.py
kirtimanmishrazipstack: @chandrasekharan-zipstack Valid — pinned in ba4f559b7. Your mutation runs were right on every count.

Now covered in `test_shared_user_gates.py`:

- **endpoint list scoping** — outsider → 404, and a shared viewer's list rows carry `connector_metadata == {}`
- **both tool-instance reparent guards** — one PATCH each for `workflow` and the `workflow_id` alias → 400, FK unchanged
- **file history** — viewer DELETE → 403 and viewer POST `clear` → 403 with the row still present; owner clear → 200
- **`clear_file_marker`** — GET → 405, viewer POST → 403
- **the new deployment scoping** — outsider and viewer refused on deploy and schedule

Each gate was reverted in turn and the matching test failed. Two things that surfaced only because of that:

1. A test of mine passed with the guard removed — `test_refusal_does_not_activate_the_workflow` was really failing on a bogus tool id. Deleted rather than kept.
2. `CreateEndpointGrantsCreatorOwnershipTests._build_workflow_fixture` built its parent workflow with `created_by` and no OWNER row, so the new scoping refused it. That fixture contradicted the class's own docstring — `created_by` is audit-only — so it now grants the row the real create path grants. Flagging it explicitly since it is an existing test I changed.

Still unpinned, and I am not claiming otherwise: the `shared_to_org` admission in `prompt_studio/permission.py`, ProfileManager viewset routing beyond the class-level check, and the profile/prompt serializer reparent guards. Happy to take those next if you want them in this PR rather than a follow-up.

frontend/src/components/agency/configure-connector-modal/ConfigureConnectorModal.jsx
```diff
@@ -314,38 +328,66 @@ function ConfigureConnectorModal({
...
         setIsSavingEndpoint(false);
       }
     }
+    // Nothing to write.
+    return true;
   };
 
+  // The read-only styling stops the mouse but not the keyboard, so cut the
+  // form's own submit path too rather than let Enter fire a doomed request.
+  const submitIfEditable = canEdit ? handleValidateAndSubmit : undefined;
```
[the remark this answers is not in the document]
kirtimanmishrazipstack: @greptile-apps Valid — fixed in f1c4a3a68.

You were right that `submitIfEditable` pointed at the endpoint-only path. The
Save button went through `handleSave`, which writes the endpoint and then the
rules; the form's own submit did not, so Enter with both halves dirty wrote the
endpoint, showed "Configuration saved successfully", and dropped the rules.

Both paths now go through `handleSave`. It takes RJSF's validated data when the
form supplies it and falls back to component state for the button, so neither
route loses the fresher value.

One thing worth recording, because the obvious fix introduces a worse bug than
the one it closes: giving `handleSave` a parameter makes the existing
`onClick={handleSave}` hand it the click event. That event would then be
compared against the saved configuration, register as a change, and be written
as the configuration. The Save button's handler is wrapped as
`onClick={() => handleSave()}` for exactly that reason.

Verified by building the integrated tree (this PR plus the cloud plugins from
Zipstack/unstract-cloud#1751, which is where the HITL rule engine lives, since
it is not present in OSS alone): `bun run build` succeeds, and the backend
suites stay green at 928 unit / 494 integration.

chandrasekharan-zipstack (approved):

PR #2274, opened 2026-09-02
UN-3853 [FIX] Attribute platform-key-created resources to the API key's creator

kirtimanmishrazipstack:
## What

-   Anything created with a **platform API key** — Prompt Studio project, workflow, ETL pipeline, API deployment, connector or adapter — now belongs to **the person who created that key**, instead of belonging to nobody.
-   Resources created before this change are repaired automatically on deploy.
-   If that person has since left or been deleted, the "Owned By" column reads **Platform key** instead of showing an email address for someone who does not exist.

## Why

-   When a script talks to Unstract with a platform API key, Unstract signs in as an invisible robot account that belongs to the key. Everything the script created was recorded as owned by that robot.
-   Unstract deliberately hides robots anywhere a person's name belongs. So these resources ended up with **no human owner at all**: they did not appear in anyone's list, the "Owned By" column had nothing real to show, and only an org admin could open, share or delete them.
-   Reported by Chandrasekharan M — a project showing `UN / Unknown` in Owned By.
-   Today the damage is limited, because only org admins can create platform keys and admins can reach everything anyway. But the moment that admin is demoted or leaves, the resource has nobody attached to it.

## How

-   New helper `owner_user_for(user)` in `backend/platform_api/services.py`. If the acting user is a platform key's service account it returns the key's `created_by`; otherwise it returns the user unchanged, before any query.
-   This is not a new rule. `delete_api_user_for_key` already hands a key's resources to `created_by` when the key is deleted. This applies the same successor at **creation** time instead of only at deletion.
-   Exposed as `HasMembersMixin.grant_owner()`, so a create site is a single call and cannot forget the rule. Seven sites in this repo route through it.
-   A creator who has since left the organization is **not** granted ownership. Handing them a fresh OWNER row would survive a re-invite, which is exactly what `cleanup_user_org_access` purges rows to prevent.
-   `created_by` is untouched everywhere. It is an audit field, and "the robot did it" is the truthful answer.
-   The "Owned By" label rule lives in one place (`owner-display.js`) so the table and card views cannot drift apart.

-   **No.** For a normal signed-in session the helper returns the user unchanged before touching the database, so every create done through the UI behaves exactly as before.
-   The only behaviour that changes is the platform-key path, which today produces an ownerless resource.
-   The robot account keeps full access to everything in its org — permission checks short-circuit on it before ownership is ever consulted.
-   Signup and onboarding are unaffected: they always run as a real person, so the helper returns early.

## Database Migrations

-   **Yes, one.** `tenant_account_v2/0006_backfill_platform_key_ownership` re-points existing robot-owned rows to the key's creator, so resources made before this change get a real owner too.
-   It is safe to run twice, and it skips any creator who has left the organization.

## Env Config

-   None.

## Relevant Docs

-   None.

## Dependencies Versions

-   None.

## Notes on Testing

Two people in the same organization, both org admins:

-   **user-1** — `kirtiman.mishra+sept4_1@zipstack.com`
-   **org admin** — `kirtiman.mishra+sept4@zipstack.com`

Each created their own platform API key, then used their own key to create resources:

-   user-1's key created `un-3853-user-wf`
-   the org admin's key created `un-3853-api-key-1`

**What each of them sees in the Owned By column:**

| Resource | Created using | user-1 sees | org admin sees |
|---|---|---|---|
| `un-3853-user-wf` | user-1's key | **Me** | kirtiman.mishra+sept4_1 |
| `un-3853-api-key-1` | org admin's key | kirtiman.mishra+sept4 | **Me** |
| `un3853-live-owner-probe` | org admin's key, two owners | kirtiman.mishra+sept4 **+1** | **Me +1** |

The two views never disagree about *who* owns a row. They differ only in how "you" is written: each person reads **Me** on the rows they own, and the other person's email address on the rest. `+1` means the resource has a second owner.

Before this change every one of these rows read **Platform key**, and neither person could open, share or delete them — only an org admin could.

The same was checked by creating each resource type through a platform API key: workflow, ETL pipeline, API deployment, Prompt Studio project, connector, adapter, and the agentic project on the cloud side. Every one is owned by the key's creator.

**Automated coverage:** 16 backend tests and 6 frontend tests. The backend ones drive the real URL and login path with a key created inside the test, so the robot-account swap is genuinely exercised. Each was checked by reverting the fix and confirming the test fails.

## Screenshots

**What user-1 sees** — their own workflow reads "Me", the org admin's reads the admin's address:

**What the org admin sees** — the mirror image:

## Checklist

https://claude.ai/code/session_01BAfubrF8kR2rKewMSjXagg

### DIFF
--- a/backend/platform_api/services.py
+++ b/backend/platform_api/services.py
@@ -15,6 +15,10 @@ if TYPE_CHECKING:
 
     from platform_api.models import PlatformApiKey
 
+# Reserved domain for service-account addresses. The frontend matches on it to
+# label an ownerless resource "Platform key" instead of naming a machine.
+SERVICE_ACCOUNT_EMAIL_DOMAIN = "platform.internal"
+
 # Business app labels whose models may carry created_by / membership rows.
 # Restricts transfer_ownership to avoid scanning Django built-in and third-party models.
 _BUSINESS_APP_LABELS = {
@@ -45,7 +49,7 @@ def create_api_user_for_key(
         name_slug = _slugify_for_email(platform_api_key.name)
         user = User(
             username=f"svc-{name_slug}-{uid[:8]}",
-            email=f"{name_slug}-{uid[:8]}@platform.internal",
+            email=f"{name_slug}-{uid[:8]}@{SERVICE_ACCOUNT_EMAIL_DOMAIN}",
             user_id=uid,
             is_service_account=True,
         )
@@ -61,6 +65,37 @@ def create_api_user_for_key(
         platform_api_key.api_user = user
         platform_api_key.save(update_fields=["api_user"])
     return user
+
+
+def owner_user_for(user: User) -> User:
+    """Resolve the human who should own a resource created by ``user``.
+
+    A platform key authenticates as a service account, and service accounts are
+    filtered out of every owner surface (``HasMembersMixin``), so a resource
+    granted to one has no human owner: it is invisible to its creator and only
+    an org admin can manage it. Attribute it to the key's creator instead — the
+    same successor :func:`delete_api_user_for_key` already hands ownership to.
+
+    Returns ``user`` unchanged for a normal session, and for the residual case
+    where the key's creator has since been deleted (``created_by`` is
+    ``SET_NULL``) — such a resource stays deliberately ownerless and the UI
+    labels it "Platform key".
+
+    Org membership of the creator is deliberately not re-checked: a key can
+    outlive its creator's membership, and granting to an ex-member matches what
+    :func:`delete_api_user_for_key` already does. The row is inert until they
+    rejoin, which beats leaving the resource with no owner at all.
+    """
+    if not getattr(user, "is_service_account", False):
+        return user
+
+    # Imported here so the module keeps its models import behind TYPE_CHECKING.
+    from platform_api.models import PlatformApiKey
+
+    key = (
+        PlatformApiKey.objects.filter(api_user=user).select_related("created_by").first()
+    )
+    return key.created_by if key and key.created_by else user
 
 
 def _get_user_fk_fields(model: type) -> list[str]:
--- a/backend/README.md
+++ b/backend/README.md
@@ -175,7 +175,8 @@ psql -d unstract_db -U unstract_dev
 
 ## API Docs
 
-The OpenAPI spec for the API deployment endpoints is committed at
+The OpenAPI spec for the publicly served API -- the API deployment endpoints,
+and the organisation-less `whoami` -- is committed at
 [`specs/docstudio-oss.json`](../specs/docstudio-oss.json) and is the contract the published
 clients and their generated SDKs are built from. It is not served at runtime — regenerate it in
 the same PR as any route, serializer or schema-annotation change:
--- a/backend/account_v2/custom_auth_middleware.py
+++ b/backend/account_v2/custom_auth_middleware.py
@@ -1,3 +1,4 @@
+import hashlib
 import logging
 import uuid
 
@@ -14,6 +15,40 @@ from backend.constants import RequestHeader
 from backend.internal_api_constants import INTERNAL_API_PREFIX
 
 logger = logging.getLogger(__name__)
+
+
+def _client_ip(request: HttpRequest) -> str:
+    """Peer address for a rejection log line.
+
+    Deliberately `REMOTE_ADDR` only. `X-Forwarded-For` is caller-supplied and
+    nothing in this project validates a forwarding chain, so trusting it would
+    let the party being logged choose what gets recorded about them -- which
+    defeats the one purpose these lines have. It would also put an unvalidated
+    header value into the log stream. `internal_api_auth.py` and
+    `internal_base_urls.py`, the only other places that log a caller address,
+    read the same field.
+
+    Behind a proxy this records the proxy. Recovering the real client needs a
+    trusted-proxy setting this project does not have; adding one is the correct
+    fix, and guessing at the hop count here is not.
+    """
+    return request.META.get("REMOTE_ADDR", "unknown")
+
+
+def _log_key_rejection(request: HttpRequest, token: str, reason: str) -> None:
+    """Record a rejected Bearer credential without recording the credential.
+
+    The fingerprint is a truncated SHA-256 of the presented token: enough to
+    correlate repeated attempts against one candidate, useless for replay.
+    """
+    fingerprint = hashlib.sha256(token.encode("utf-8", "replace")).hexdigest()[:12]
+    logger.warning(
+        "Platform key rejected (%s), fingerprint %s, path %s, from %s",
+        reason,
+        fingerprint,
+        request.path,
+        _client_ip(request),
+    )
 
 
 class CustomAuthMiddleware:
@@ -56,6 +91,12 @@ class CustomAuthMiddleware:
 
         if is_authenticated:
             organization_id = UserSessionUtils.get_organization_id(request=request)
+            # `request.organization_id` is None on two kinds of path: one with no
+            # organisation segment at all, and one matched by
+            # ORGANIZATION_MIDDLEWARE_WHITELISTED_PATHS, which sets it to None
+            # deliberately. Either way this guard does not run, so a route added
+            # to that whitelist loses this check as well as the Bearer branch's
+            # -- see the note beside the setting.
             if request.organization_id and not organization_id:
                 return JsonResponse({"message": "Organization access denied"}, status=403)
             StateStore.set(Common.LOG_EVENTS_ID, request.session.session_key)
@@ -81,6 +122,12 @@ class CustomAuthMiddleware:
         try:
             key_uuid = uuid.UUID(token_str)
         except (ValueError, AttributeError):
+            # Logged like the branch below: these two are the ones an attacker
+            # traverses, and until now neither left a trace, so "was this key
+            # ever presented?" had no data to answer from. The token is never
+            # logged -- only a truncated SHA-256, which is enough to correlate
+            # repeated attempts against one candidate without storing it.
+            _log_key_rejection(request, token_str, "malformed key")
             return JsonResponse({"message": "Invalid API key format"}, status=401)
 
         try:
@@ -88,12 +135,22 @@ class CustomAuthMiddleware:
                 "created_by", "api_user", "organization"
             ).get(key=key_uuid, is_active=True)
         except PlatformApiKey.DoesNotExist:
+            _log_key_rejection(request, token_str, "unknown or inactive key")
             return JsonResponse({"message": "Invalid or inactive API key"}, status=401)
 
         # Validate the key belongs to the org in the URL
         if request.organization_id and str(key.organization.organization_id) != str(
             request.organization_id
         ):
+            # The highest-signal security event in this file: a valid key aimed
+            # at an organisation it does not belong to.
+            logger.warning(
+                "Platform key %s (org %s) refused for org %s from %s",
+                key.id,
+                key.organization.organization_id,
+                request.organization_id,
+                _client_ip(request),
+            )
             return JsonResponse(
                 {"message": "API key does not belong to this organization"},
                 status=403,
--- a/backend/adapter_processor_v2/views.py
+++ b/backend/adapter_processor_v2/views.py
@@ -16,6 +16,7 @@ from permissions.permission import (
 )
 from permissions.resource_share_views import ResourceShareManagementMixin
 from permissions.roles import ResourceRole
+from platform_api.services import owner_user_for
 from plugins import get_plugin
 from rest_framework import status
 from rest_framework.decorators import action
@@ -272,7 +273,8 @@ class AdapterInstanceViewSet(
             # ``created_by`` is audit-only; the creator's access flows through
             # an OWNER membership row (UN-2202 co-owners).
             instance.memberships.get_or_create(
-                user_id=request.user.id, defaults={"role": ResourceRole.OWNER}
+                user_id=owner_user_for(request.user).id,
+                defaults={"role": ResourceRole.OWNER},
             )
             organization_member = OrganizationMemberService.get_user_by_id(
                 request.user.id
--- a/backend/api_v2/api_deployment_views.py
+++ b/backend/api_v2/api_deployment_views.py
@@ -9,6 +9,7 @@ from permissions.membership_views import OwnerManagementMixin
 from permissions.permission import IsOwner, IsOwnerOrSharedUserOrSharedToOrg
 from permissions.resource_share_views import ResourceShareManagementMixin
 from permissions.roles import ResourceRole
+from platform_api.services import owner_user_for
 from plugins import get_plugin
 from prompt_studio.prompt_studio_registry_v2.models import PromptStudioRegistry
 from rest_framework import serializers, status, views, viewsets
@@ -326,7 +327,8 @@ class APIDeploymentViewSet(
         # ``created_by`` is audit-only; the creator's access flows through an
         # OWNER membership row (UN-2202 co-owners).
         serializer.instance.memberships.get_or_create(
-            user_id=request.user.id, defaults={"role": ResourceRole.OWNER}
+            user_id=owner_user_for(request.user).id,
+            defaults={"role": ResourceRole.OWNER},
         )
         api_key = DeploymentHelper.create_api_key(serializer=serializer, request=request)
         response_serializer = DeploymentResponseSerializer(
--- a/backend/api_v2/api_deployment_views.py
+++ b/backend/api_v2/api_deployment_views.py
@@ -3,12 +3,14 @@ import logging
 import uuid
 from typing import Any
 
-from django.db.models import F, OuterRef, QuerySet, Subquery
+from django.db.models import Count, F, IntegerField, OuterRef, QuerySet, Subquery
+from django.db.models.functions import Coalesce
 from django.http import HttpResponse
 from permissions.membership_views import OwnerManagementMixin
 from permissions.permission import IsOwner, IsOwnerOrSharedUserOrSharedToOrg
 from permissions.resource_share_views import ResourceShareManagementMixin
 from permissions.roles import ResourceRole
+from platform_api.openapi_schema import PlatformKeyAutoSchema
 from platform_api.services import owner_user_for
 from plugins import get_plugin
 from prompt_studio.prompt_studio_registry_v2.models import PromptStudioRegistry
@@ -34,7 +36,10 @@ from api_v2.exceptions import (
     contains_tool_not_found_error,
 )
 from api_v2.models import APIDeployment
-from api_v2.openapi_schema import DEPLOYMENT_EXECUTION_SCHEMA
+from api_v2.openapi_schema import (
+    API_DEPLOYMENT_LIST_SCHEMA,
+    DEPLOYMENT_EXECUTION_SCHEMA,
+)
 from api_v2.rate_limiter import APIDeploymentRateLimiter
 from api_v2.serializers import (
     APIDeploymentListSerializer,
@@ -247,10 +252,17 @@ class DeploymentExecution(views.APIView):
         )
 
 
+@API_DEPLOYMENT_LIST_SCHEMA
 class APIDeploymentViewSet(
     OwnerManagementMixin, ResourceShareManagementMixin, viewsets.ModelViewSet
 ):
     pagination_class = CustomPagination
+
+    # For schema generation only; get_queryset replaces it on every request.
+    queryset = APIDeployment.objects.none()
+
+    # Error examples have to match what the auth middleware sends.
+    schema = PlatformKeyAutoSchema()
     notification_resource_name_field = "display_name"
 
     def get_notification_resource_type(self, resource: Any) -> str | None:
@@ -276,19 +288,42 @@ class APIDeploymentViewSet(
             .order_by("-created_at")
             .values("created_at")[:1]
         )
+        run_count_subquery = (
+            WorkflowExecution.objects.filter(pipeline_id=OuterRef("id"))
+            .values("pipeline_id")
+            .annotate(total=Count("id"))
+            .values("total")
+        )
 
         # Avoid per-row queries for owner/co-owner + creator fields in list views
         queryset = (
             APIDeployment.objects.for_user(self.request.user)
             .select_related("created_by")
             .prefetch_related("memberships__user")
-            .annotate(last_run_time_annotated=Subquery(last_run_subquery))
-            .order_by(F("last_run_time_annotated").desc(nulls_last=True))
+            .annotate(
+                last_run_time_annotated=Subquery(last_run_subquery),
+                run_count_annotated=Coalesce(
+                    Subquery(run_count_subquery, output_field=IntegerField()), 0
+                ),
+            )
+            # `pk` last because the primary ordering ties on every deployment
+            # that has never run, and a paging client would then see a row
+            # twice or not at all.
+            .order_by(F("last_run_time_annotated").desc(nulls_last=True), "pk")
         )
 
-        # Filter by workflow ID if provided
+        # TODO: replace the hand-read params and their OpenApiParameter
+        # restatements with a FilterSet so the spec cannot drift from the code
         workflow_filter = self.request.query_params.get("workflow", None)
         if workflow_filter:
+            try:
+                uuid.UUID(workflow_filter)
+            except ValueError:
+                # Django raises on evaluation, past the handler that turns a bad
+                # request into a 400.
+                raise serializers.ValidationError(
+                    {"workflow": "Must be a valid UUID."}
+                ) from None
             queryset = queryset.filter(workflow_id=workflow_filter)
 
         # Search by display name
--- a/backend/api_v2/deployment_spec_urls.py
+++ b/backend/api_v2/deployment_spec_urls.py
@@ -9,11 +9,14 @@ Widening the spec to another endpoint means annotating its view with
 ``@extend_schema`` and adding its urlconf here.
 """
 
+from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
+from django.urls import path
 
+from api_v2.api_deployment_views import APIDeploymentViewSet
 from backend import base_urls
 
-SPEC_URLCONFS = ("api_v2.execution_urls",)
+SPEC_URLCONFS = ("api_v2.execution_urls", "platform_api.whoami_urls")
 
 urlpatterns = [
     entry
@@ -27,3 +30,14 @@ if missing:
         f"{', '.join(sorted(missing))} is not mounted in backend.base_urls; the "
         "spec would be generated for routes the server does not serve."
     )
+
+# Restated rather than selected: this route is served from a urlconf carrying
+# routes that are not published, and only its GET is. Both halves of the
+# restatement are held to the served route by `test_docstudio_spec`.
+urlpatterns += [
+    path(
+        f"{settings.TENANT_SUBFOLDER_PREFIX}/api/deployment/",
+        APIDeploymentViewSet.as_view({"get": "list"}),
+        name="api_deployment",
+    ),
+]
--- a/backend/api_v2/management/commands/generate_docstudio_spec.py
+++ b/backend/api_v2/management/commands/generate_docstudio_spec.py
@@ -7,16 +7,18 @@ serializer or the schema annotation, and regenerate in the same PR.
     uv run python manage.py generate_docstudio_spec           # from backend/
     uv run python manage.py generate_docstudio_spec --check   # no write, drift is an error
 
-The generated paths carry ``API_DEPLOYMENT_PATH_PREFIX``, so generation refuses
-to produce a spec mounted anywhere but the public default: the committed
-artifact describes the deployment as it is served publicly, not as one
-installation chooses to mount it.
+The generated paths carry ``API_DEPLOYMENT_PATH_PREFIX`` or ``PATH_PREFIX``
+depending on the mount, so generation refuses to produce a spec mounted anywhere
+but the public defaults: the committed artifact describes the API as it is
+served publicly, not as one installation chooses to mount it.
 """
 
 import json
+import re
 from pathlib import Path
 from typing import Any
 
+from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
 from drf_spectacular.drainage import GENERATOR_STATS
 from drf_spectacular.generators import SchemaGenerator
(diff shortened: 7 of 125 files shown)


kirtimanmishrazipstack:
### Merge this before the cloud PR

Zipstack/unstract-cloud#1753 imports `owner_user_for` from `platform_api.services`, and cloud CI pins the OSS tree to `refs/heads/main` with no override — so its `test (integration)` and `e2e` stay red until this merges. Once this lands, a re-run there goes green on its own.

kirtimanmishrazipstack:
### The red CI here is an infra issue, unrelated to this change

`test (integration)` and `e2e` both fail while pulling the MinIO images from Docker Hub, before any test runs:

```
pull access denied for minio/minio, repository does not exist or may require 'docker login'
No such image: minio/minio:latest
```

`report` is red only because those two are. Nothing in this branch is involved — the tiers that do exercise it pass.

Deepak-Kesavan (commented):
## 🤖 Automated review

4 inline comment(s) · 1 finding(s) not on changed lines (below).

### Findings off the diff

- `backend/api_v2/api_deployment_views.py:406`
**[minor] `by_prompt_studio_tool` serializes with `APIDeploymentListSerializer` on a queryset that has neither `select_related("created_by")` nor `prefetch_related("memberships__user")`, so the new `owner_emails` SerializerMethodField adds one membership query plus one user query per membership row, per deployment.**

GET /api/v1/unstract/<org>/api/deployment/by_prompt_studio_tool/?tool_id=<id> with N matching deployments. `get_owner_emails` (backend/api_v2/serializers.py:547-551) calls `obj.owner_emails()`, which iterates `self.memberships.all()` and dereferences `m.user.email` (backend/permissions/models.py:60-70). With no prefetch cache, `memberships.all()` issues a fresh SELECT per row and each `m.user` issues another, so the endpoint gains 1 + (#owners) queries per deployment on top of the pre-existing per-row cost. The two list viewsets are prefetched (`api_deployment_views.py:300`, `pipeline_v2/views.py:86`); this caller was missed.

**Suggested fix:** Add the same eager loading at backend/api_v2/api_deployment_views.py:406: `APIDeployment.objects.for_user(request.user).select_related("created_by").prefetch_related("memberships__user").filter(workflow_id__in=workflow_ids)`.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:165a1da9d29f</sub>

backend/permissions/models.py
```diff
@@ -76,3 +76,12 @@ def is_owner(self, user: Any) -> bool:
             m.user_id == user.id and m.role == ResourceRole.OWNER
             for m in self.memberships.all()  # type: ignore[attr-defined]
         )
+
+    def grant_owner(self, user: Any) -> None:
+        """Grant OWNER on create, resolving a platform key to its creator."""
+        # Lazy: platform_api.services imports models at import time.
+        from platform_api.services import owner_user_for
+
+        self.memberships.get_or_create(  # type: ignore[attr-defined]
```
Deepak-Kesavan, line 85: **[critical] `grant_owner` moves the OWNER row off the service account, but two authorization paths still authorize by OWNER membership and have **no** `is_service_account` bypass — so every workflow provisioned through a platform key starts failing at execution with `PermissionDenied`, and migration 0006 applies the same change to already-deployed installs.**

Provision entirely through a platform API key — the flow this PR exists for: POST an AdapterInstance, then a Workflow, then a tool instance on it.

`grant_owner` now writes the OWNER `ResourceMembership` row to the key's human creator instead of the service account S. `created_by` is untouched, so `Workflow.created_by` is still S (`AuditSerializer.create`, `backend/backend/serializers.py:12`).

At run time `WorkflowHelper.run_workflow` -> `validate_tool_instances_meta` sets `user = tool.workflow.created_by` (= S, `workflow_helper.py:289`, reached from `:321`) and calls `validate_adapter_access` (`tool_instance_v2/tool_instance_helper.py:520-551`). That gate evaluates:
- `is_admin` -> False (`OrganizationMemberService.is_user_organization_admin` early-returns False for service accounts, `organization_member_service.py:40-41`; `create_api_user_for_key` gives the account `UserRole.USER`)
- `adapter_instance.shared_to_org` -> False (default, `adapter_processor_v2/models.py:112-115`)
- `_is_resource_owner(S, adapter)` -> **False after this PR** (it consults only `memberships`, `permissions/permission.py:79-84`; pre-PR the row named S and this returned True)
- `_is_resource_viewer` / `has_group_access` -> False

=> `raise PermissionDenied` on **every** execution. Both transports hit it: the in-backend Celery path and the workers path via `tool_instance_v2/internal_views.py:352-366`.

The same gap hits `IsWorkflowOwnerOrShared` (`workflow_manager/workflow_v2/permissions.py:40-45`) — no service-account short-circuit — so the key gets 403 on the file-history endpoints for workflows it created.

**Reproduced live against Postgres.** With the post-PR grant, `validate_adapter_access(user=service_account)` raises `PermissionDenied`; the control writing main's service-account OWNER row passes the identical gate. Over real HTTP with platform-key bearer auth: `GET .../workflow/<id>/file-histories/` returns **403 permission_denied** post-PR and **200** with main's row.

Note `PromptStudioRegistry.objects.list_tools` *does* have the `is_service_account` bypass (`prompt_studio_registry_v2/models.py:32-35`) — which is likely why this gap went unnoticed. And migration 0006 filters `role=OWNER, user_id__in=<service accounts>` with **no content-type restriction**, so already-deployed, currently-working platform-key pipelines start failing the moment this deploys.

This is not covered by the object-permission short-circuit in `permission.py` — neither call site is a DRF `has_object_permission` on a service-account request.

**Suggested fix:** Either add the same `if getattr(user, "is_service_account", False): return` short-circuit that every other surface has to `ToolInstanceHelper.validate_adapter_access` and `IsWorkflowOwnerOrShared.has_permission`; **or** have `grant_owner` write *both* rows (service account + resolved creator). The owner surfaces already filter service accounts out, so the extra row stays invisible in the UI while nothing that resolves ownership through the membership table regresses.

Worth a test that executes a key-provisioned workflow end to end — the new test files assert the create-time row but never an execution.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:60d76e591c9b</sub>

backend/platform_api/services.py
```diff
@@ -173,11 +237,16 @@ def transfer_ownership(from_user: User, to_user: User | None) -> None:
...
+    short-circuits on ``None`` and the rows are dropped with the account, which
+    is what ``cleanup_user_org_access`` would have done to them anyway.
+    """
     api_user = platform_api_key.api_user
     if not api_user:
         return
 
     with transaction.atomic():
-        transfer_ownership(from_user=api_user, to_user=platform_api_key.created_by)
+        transfer_ownership(from_user=api_user, to_user=live_key_creator(platform_api_key))
```
Deepak-Kesavan: **[major] Swapping `platform_api_key.created_by` for `live_key_creator(...)` makes `transfer_ownership` short-circuit when the creator has left the org, while `api_user.delete()` still runs unconditionally — CASCADE-deleting the service account's OWNER rows and SET_NULLing `created_by`/`modified_by` on every resource that key created. Main preserved both.**

Alice creates platform key K and provisions resources with it. Alice later leaves the org — `OrganizationMemberService.delete_user` removes only the `OrganizationMember` row (`organization_member_service.py:133-149`), the global `User` survives, and nothing deactivates K (`tenant_account_v2/signals.py:19` touches GroupMembership/ResourceMembership only). So `live_key_creator` returns `None` while main's `created_by` was non-None.

Someone then deletes K. `pre_delete` (`platform_api/signals.py:12-16`) calls `delete_api_user_for_key`:
- `transfer_ownership` returns immediately at `:227-228` (`if not to_user: return`), skipping `_transfer_model_ownership` (`:157-180`) — the only code that moves `created_by`/`modified_by`
- `api_user.delete()` at `:252` still runs. `ResourceMembership.user` is `on_delete=CASCADE` (`tenant_account_v2/models.py:150-154`) -> OWNER rows destroyed. `created_by` is `SET_NULL` on all six business models -> audit pointer nulled.

**Reproduced live against Postgres** (org + Alice + key + SA, SA-held OWNER row on a Workflow, delete Alice's membership, delete the key; only `services.py` swapped between runs):
```
PR HEAD    : owners ['<sa>@platform.internal'] -> []                    created_by = None
origin/main: owners ['<sa>@platform.internal'] -> ['alice@example.com']  created_by = Alice
```

**Concrete downstream break:** `CustomTool.delete()` dereferences `self.created_by.user_id` (`prompt_studio_core_v2/models.py:204`), reached via `views.py:276` -> `perform_destroy`. With `created_by` NULL that is an AttributeError/500 — the Prompt Studio project can never be deleted.

**This is the dominant case, not a corner.** `owner_user_for` still falls back to the service account whenever `live_key_creator` is None, and the 0006 backfill skips the same keys — so virtually every surviving service-account OWNER row belongs to a key whose creator has departed, which is exactly where the new transfer short-circuits. Reachable with no pre-existing data: create through the key after the creator leaves, then delete the key.

**The new docstring's justification is contradicted by the path it cites.** `services.py:242-244` says this is "what `cleanup_user_org_access` would have done to them anyway". `_purge_replaceable_owner_rows` (`tenant_account_v2/signals.py:79-84, 129-135`) deliberately **keeps** a sole owner's row "so the resource is never left ownerless (spec §7.4)" — it purges only rows where another live owner remains. Verified by repro: running `cleanup_user_org_access` on the SA keeps the row.

In fairness: an org admin can still recover ownership (`IsOwner` admits `_is_organization_admin`, `permissions/permission.py:104-119`, and `AddOwnerSerializer` has no "must already have an owner" guard), so this is recoverable — but the NULLed `created_by` is not, and neither is the broken `CustomTool.delete()`.

**Suggested fix:** Keep the membership half if you want (closing the ex-member rejoin backdoor is a defensible trade-off), but don't let `api_user.delete()` run unconditionally when no successor was named — nulling the audit FKs has no security justification and main preserved them. Options: name a fallback successor (an org admin) so §7.4 "never ownerless" still holds; or move `created_by`/`modified_by` independently of the OWNER row so audit survives even when ownership does not.

Either way `services.py:242-244` needs rewording — it asserts an equivalence with `cleanup_user_org_access` that `signals.py:33-36` explicitly denies.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:9150a6d16a47</sub>

frontend/src/components/deployments/create-api-deployment-modal/CreateApiDeploymentModal.jsx
```diff
@@ -109,8 +110,9 @@ const CreateApiDeploymentModal = ({
             onDeploymentCreated();
           }
         } else {
-          // Add new deployment to list
-          setTableData((prev) => [res?.data, ...prev]);
+          // Refetch: the create response is a summary without the owner
+          // fields the list renders.
+          refreshList?.();
```
Deepak-Kesavan, line 115: **[minor] Replacing the optimistic prepend with `refreshList?.()` refetches the *current* page, but the deployment list is ordered `last_run_time DESC NULLS LAST, pk` — so a never-run deployment sorts behind everything that has ever run and the just-created row can be nowhere on the page.**

`refreshList?.()` (`:115`) replaces main's `setTableData((prev) => [res?.data, ...prev])`. `ApiDeployment.jsx:364` wires it to `handleListRefresh`, which replays `requestedRef.current` verbatim (`usePaginatedList.js:190-193`) — it does not go through `requestList` and never resets to page 1.

The server orders `.order_by(F("last_run_time_annotated").desc(nulls_last=True), "pk")` (`api_v2/api_deployment_views.py:310-311`, unchanged from base). `last_run_time_annotated` is NULL for a never-run deployment, the viewset declares no `filter_backends` and reads no `ordering` param, and the client sends only page/page_size/search (`api-deployments-service.js:18-31`). So a brand-new deployment sorts behind every deployment that has ever run, and the user cannot sort out of it.

`APIDeployment.id` is a random uuid4 (`api_v2/models.py:61`), so the `"pk"` tiebreak puts the new row at a **random position within the never-run block** — not deterministically at the tail. Concretely: once ~pageSize (10) deployments have ever run, the new row is off page 1 with certainty. It is also invisible whenever a search term is active, since `handleListRefresh` replays the search the new name may not match.

The premise for the change is right — the create response is `DeploymentResponseSerializer` (`serializers.py:594-601`: only is_active/id/api_key/api_endpoint/display_name/description/api_name), so the prepended row genuinely lacked `owner_emails`/`is_owner`/`last_run_time`. Dropping the prepend is correct; refetching the *current* page is what loses the row.

And yes, "refetch the current page" is this codebase's stated convention (`ToolSettings.jsx:211-213`, `ListOfTools.jsx:246-249`) — but those lists default to `-modified_at, pk` (`prompt_studio_core_v2/models.py:225`, `connector_v2/models.py:172`), so their new row lands at the top of page 1. The deployment list's NULLS-LAST ordering is the one case where the convention does not hold.

**Suggested fix:** Reset to page 1 and clear the search before refetching (or have `handleListRefresh` accept a `{resetPage: true}` option), so the refetch lands on a page where the new row can actually appear. Alternatively re-fetch the single created deployment through the list serializer and prepend that.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:652eff25a7c7</sub>

backend/tenant_account_v2/migrations/0006_backfill_platform_key_ownership.py
```diff
@@ -0,0 +1,73 @@
...
+    for row in rows.iterator():
+        new_user_id = successor[row.user_id]
+        clash = resource_membership_model.objects.filter(
+            user_id=new_user_id,
+            content_type_id=row.content_type_id,
+            object_id=row.object_id,
+        ).first()
+        if clash is None:
+            row.user_id = new_user_id
+            row.save(update_fields=["user"])
```
Deepak-Kesavan: **[minor] `row.save(update_fields=["user"])` and `clash.save(update_fields=["role"])` run on historical models, which do not carry `BaseModel.save`'s `modified_at` injection, so the rewritten ownership rows keep a stale `modified_at`.**

Every row the backfill re-points or promotes has its owner silently changed while `modified_at` still shows the original creation timestamp. An operator triaging 'when did this resource change hands?' after the migration sees no evidence that anything happened, and any downstream consumer that syncs on `modified_at` (the repo maintains `BaseModelQuerySet.update/bulk_update/bulk_create` overrides specifically to prevent this drift) will skip the changed rows. The codebase's own docstring names this exact gap: 'migration-time models returned by apps.get_model() all bypass these overrides.' CONFIRMED by code read.

**Suggested fix:** Pass the field list explicitly in the migration: set `row.modified_at = timezone.now()` and use `update_fields=["user", "modified_at"]` (likewise `["role", "modified_at"]`), since the historical model will not add it for you.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:97d1d40aaadb</sub>

backend/permissions/models.py
kirtimanmishrazipstack: @Deepak-Kesavan Confirmed and fixed in a57f99b3b.

Verified the whole chain rather than taking it on trust: `validate_tool_instances_meta` sets `user = tool.workflow.created_by` (the service account, since `created_by` is deliberately untouched), `validate_adapter_access` has no service-account bypass, `is_user_organization_admin` early-returns False for one, and `_is_resource_owner` reads only `memberships` — which after this PR names the human. So every gate term evaluates False. Same gap in `IsWorkflowOwnerOrShared.has_permission`.

Took your first option — the short-circuit — rather than writing both rows, because a second OWNER row would have to be kept in sync by every future create site and the migration, whereas the bypass is the pattern every other surface already uses (`IsOwner`, `for_user`, `list_tools`).

Added two regression tests. Worth noting the first one I wrote was vacuous: `AdapterInstance.objects` is org-scoped and `UserContext` was unset, so the gate looped over an empty queryset and passed with the fix reverted. Setting the org identifier makes it fail correctly without the bypass.

backend/platform_api/services.py
kirtimanmishrazipstack: @Deepak-Kesavan Confirmed and fixed in a57f99b3b.

The mechanism is exactly as described — `transfer_ownership` early-returns on a falsy `to_user` while `api_user.delete()` runs unconditionally, and `ResourceMembership.user` is CASCADE with `created_by` SET_NULL. `CustomTool.delete()` dereferencing `self.created_by.user_id` confirms the downstream break.

Your point about the docstring is right and was the more useful half: `_purge_replaceable_owner_rows` explicitly keeps a sole owner's row 'so the resource is never left ownerless (spec §7.4)', so the claimed equivalence was false. That wording is gone.

Split the two halves rather than naming a fallback successor. `transfer_ownership` now takes `membership_to`: audit FKs follow `created_by` even after they leave, since nulling them has no security value, while the OWNER row still only goes to a live member. Keeps the rejoin fix and restores main's audit behaviour. Regression test added.

frontend/src/components/deployments/create-api-deployment-modal/CreateApiDeploymentModal.jsx
kirtimanmishrazipstack: @Deepak-Kesavan Confirmed and fixed in a57f99b3b.

Verified both halves: the queryset orders `last_run_time_annotated DESC NULLS LAST, pk`, and `handleListRefresh` replays `requestedRef.current` verbatim with no page reset. A never-run deployment therefore sorts last and can be off-page.

Added `handleCreatedRefresh`, which re-requests page 1 while preserving the current search and sort, and wired the create modal to it. The plain refresh stays as-is for callers that should hold their page.

backend/tenant_account_v2/migrations/0006_backfill_platform_key_ownership.py
kirtimanmishrazipstack: @Deepak-Kesavan Confirmed and fixed in a57f99b3b.

I initially thought this might be a false positive, since `modified_at` is `DateTimeField(auto_now=True)` and the historical model in `0005` carries that flag — `auto_now` is field-level and does fire on a historical model's `save()`. But it doesn't rescue the case: `update_fields=["user"]` excludes `modified_at` from the UPDATE, and it's `BaseModel.save` — via `_with_modified_at` — that injects it into a partial field list. Historical models don't inherit that. So your conclusion holds for exactly the reason you gave.

Both save sites now set `modified_at = timezone.now()` and pass it in `update_fields`.

kirtimanmishrazipstack:
@Deepak-Kesavan Thanks — all five findings reproduced against the code, all five fixed in a57f99b3b (OSS) and 402f4ea1 (cloud). Replies are on each thread; this one covers the off-diff finding.

**`by_prompt_studio_tool` N+1 — confirmed, fixed.** The two list viewsets do carry `select_related("created_by").prefetch_related("memberships__user")`; this caller had neither while serializing with the same serializer, so the new `owner_emails` field fanned out per deployment. Added the same eager loading.

Two notes on the process, since they affected the outcome:

**One of my regression tests was initially vacuous.** The test for the critical adapter-access finding passed with the fix reverted, because `AdapterInstance.objects` is org-scoped and `UserContext` was unset — the gate looped over an empty queryset. Caught it by mutation-testing each new test rather than trusting green. Every test added here was verified by reverting its own fix.

**The cloud storage finding was invisible locally.** That test passed on my machine because my dev `.env` supplies storage config the CI rig doesn't set — so it was green here and would have errored in CI, with the mutation check passing for the wrong reason. Exactly as you called it.

Full suite after the fixes: 382 passed, spec check clean, hooks clean on both repos.

https://claude.ai/code/session_01BAfubrF8kR2rKewMSjXagg

frontend/src/components/deployments/api-deployment/ApiDeployment.jsx
```diff
@@ -83,10 +83,20 @@ function ApiDeployment() {
...
     fetchRef,
+    sort,
+    requestList,
     handlePaginationChange,
     handleSearch,
   } = usePaginatedList();
 
+  // A new deployment has never run, and the list sorts never-run rows last, so
+  // refreshing the current page can leave it out of view. Go back to page 1.
+  const handleCreatedRefresh = useCallback(
```
Deepak-Kesavan: **[major] **My earlier suggested fix was wrong and this implements it faithfully.** Never-run deployments sort *last*, not first — so jumping to page 1 targets the wrong end of the list. The new deployment is still off-screen, and the user is now also thrown off the page they were on.**

First, the correction: my previous comment on this line suggested "reset to page 1 ... so the refetch lands on a page where the new row can actually appear." That was wrong. `api_v2/api_deployment_views.py:310` orders `F("last_run_time_annotated").desc(nulls_last=True), "pk"`, and `nulls_last=True` exists precisely to push never-run rows to the **END** (Postgres defaults DESC to NULLS FIRST). A brand-new deployment has `last_run_time IS NULL`, so it lands on the **last** page. Page 1 is the wrong target. Sorry for the bad steer.

Concretely: org has 25 deployments that have all run at least once, `pageSize` 10. User is on page 3 and creates a new deployment. `handleCreatedRefresh` fires `requestList(1, 10, ...)`. The new row is at ordered index 25 -> page 3 of 3. Page 1 shows the 10 most-recently-run deployments; the new one is not among them. So the fix leaves it invisible **and** yanks the user from page 3 to page 1 — where the old `handleListRefresh` would at least have kept them on page 3, which is where the row actually is.

The comment at `:92-93` states the correct premise ("the list sorts never-run rows last") and then draws the inverted conclusion ("Go back to page 1").

No escape hatch exists: `ApiDeployment.jsx:170` declares `getApiDeploymentList(page, pageSize, search)` and `api-deployments-service.js:18-32` sends only `{page, page_size, search}` — the `sort.sortBy`/`sort.order` threaded into `requestList` are dropped before the wire, and `DeterministicOrderingFilter` never sees an `ordering` param. The `:310` ordering applies on every request.

Caveat: when the post-create total is <= pageSize there is only one page and the fix works, and the refetch does still fix the owner-column half of the original problem. The surfacing goal fails whenever total > pageSize.

**Suggested fix:** No fixed page is reachable-correct under this ordering — the new row's position inside the never-run block is a random uuid4 tiebreak. Two options that actually work:

1. **Add `created_at DESC` ahead of `pk`** in the `order_by` (or order by `-created_at` outright) so newly created rows are deterministically first; *then* page 1 is the right target and the current frontend change becomes correct. This also removes the random-uuid tiebreak, which is a paging hazard in its own right.

2. **Keep the optimistic prepend** but re-fetch that one deployment through the list serializer so it carries `owner_emails`/`is_owner`/`co_owners_count`, rather than prepending the thin `DeploymentResponseSerializer` payload.

If neither is in scope, reverting to `handleListRefresh` is strictly better than jumping to page 1 — it at least keeps the user on the page where the row can be.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:fac4f72e2b2c</sub>

Deepak-Kesavan (commented):
## 🤖 Automated review (re-review)

1 inline comment(s) · 1 finding(s) not on changed lines (below).

### Findings off the diff

- `backend/permissions/permission.py:285`
**[major] The fix patched the two authz gates that were reported, but `IsFrictionLessAdapterDelete` is a **third** ownership gate with no `is_service_account` bypass — so a full_access platform key can no longer delete an adapter it created through that same key.**

1. Admin mints a `full_access` platform key.
2. `POST /api/v1/unstract/<org>/adapter/` with `Authorization: Bearer <key>` creates an LLM adapter — the exact flow this PR's own `test_platform_key_resource_ownership.test_adapter` asserts. `grant_owner(request.user)` resolves the service account to the human creator, so the only OWNER row names the human.
3. The same key issues `DELETE /api/v1/unstract/<org>/adapter/<id>/`. `AdapterInstance.objects.for_user(SA)` returns `.all()`, so the object resolves; then `IsFrictionLessAdapterDelete.has_object_permission` runs (`adapter_processor_v2/views.py:170-171` routes `destroy` to it):
   - `obj.is_friction_less` -> False
   - `_is_resource_owner(SA, adapter)` -> **False** (post-PR the row is the human's; pre-PR it was the SA's and this returned True)
   - `_is_organization_admin(SA)` -> False by design (`organization_member_service.py:39` short-circuits service accounts)
   => **403**, where main returned 204.

`permissions/permission.py:274-287` is the only `_is_resource_owner` call site in the tree without a service-account short-circuit. Every sibling in the same file has one (`:114, :131, :163, :194, :229, :243, :268`), including `IsFrictionLessAdapter` immediately above it, and `prompt_studio/permission.py:24, :51` have one too. The asymmetry is right next to the fix.

It ships green: the new `ServiceAccountStaysAuthorizedTest` covers exactly the two sites that were reported, and `test_platform_key_resource_ownership.test_adapter` asserts the create but never the subsequent delete.

This also still falsifies the PR body's claim that "the robot account keeps full access to everything in its org".

**Suggested fix:** Add the bypass the same way every sibling in the file does, after the frictionless carve-out:

```python
def has_object_permission(self, request, view, obj):
    if obj.is_friction_less:
        return True
    if _is_service_account(request):
        return True
    if _is_resource_owner(request.user, obj):
        return True
    return _is_organization_admin(request)
```

Worth adding a `DELETE .../adapter/<id>/` case to `ServiceAccountStaysAuthorizedTest` — the two cases it has would not have caught this one.

<sub>🤖 Unstract PR review kit (Claude Code) · review-pr-bot:d82dbd8bdf70</sub>

frontend/src/components/deployments/create-api-deployment-modal/CreateApiDeploymentModal.jsx
Deepak-Kesavan: Correction on this one — **my original suggestion here was wrong**, and the fix in `a57f99b3b` implements it faithfully.

I suggested resetting to page 1. But `api_deployment_views.py:310` orders `F("last_run_time_annotated").desc(nulls_last=True), "pk"` — `nulls_last=True` pushes never-run rows to the **end**, so a brand-new deployment lands on the **last** page, not the first. Jumping to page 1 targets the wrong end: the new row is still off-screen, and the user is now also moved off the page it is actually on.

Details and two options that do work are in the new review comment on `ApiDeployment.jsx:94`. Sorry for the bad steer.

<sub>🤖 Unstract PR review kit (Claude Code) · automated pass</sub>

frontend/src/components/deployments/api-deployment/ApiDeployment.jsx
kirtimanmishrazipstack: @Deepak-Kesavan You're right, and thanks for catching your own steer — reverted in 040f20546.

I should have caught this myself. `nulls_last=True` exists precisely to push never-run rows to the end, and the comment I wrote said exactly that ("the list sorts never-run rows last") before concluding "Go back to page 1". That contradiction was sitting in my own diff and I implemented the suggestion without checking its premise against the ordering.

Reverted to `handleListRefresh`. As you say it is strictly better: it keeps the user on the page the row is actually on, and still fixes the half of the original finding that matters — the created row no longer renders without owner fields, because the list is refetched rather than prepended from the thin create response.

Left your two structural options alone deliberately. Ordering by `-created_at` changes how the deployment list is sorted for everyone, and re-fetching the single row through the list serializer adds a request to a path that already has one; both are worth doing but neither belongs in an ownership PR. The surfacing gap predates this change — a brand-new deployment landed on the last page before it too.

kirtimanmishrazipstack:
@Deepak-Kesavan Both re-review findings fixed in 040f20546.

**`IsFrictionLessAdapterDelete` — confirmed, fixed.** Verified the asymmetry you pointed at: seven `_is_service_account(request)` bypasses in that file (lines 114, 131, 163, 194, 229, 243, 268), including `IsFrictionLessAdapter` immediately above, and the Delete variant had none while `destroy` routes to it. Added the bypass after the frictionless carve-out, as suggested.

**I swept the rest rather than wait for a fourth round.** All 13 `_is_resource_owner` call sites across both repos:

| Site | Covered by |
|---|---|
| `permission.py` 116, 133, 166, 212, 232, 246, 270 | existing bypasses at 114, 131, 163, 194, 229, 243, 268 |
| `permission.py:287` | this commit |
| `tool_instance_helper.py:546`, `workflow_v2/permissions.py:43` | a57f99b3b |
| `prompt_studio/permission.py` 27, 54 | bypasses at 24, 51 |
| `prompt_studio_helper.py:102` | unreachable — its only caller short-circuits on `is_service_account` before calling it |

So the PR body's claim about the service account keeping org-wide access is now true rather than aspirational.

**On the page-1 revert** — reply is on that thread. Short version: you were right to correct it, and the contradiction was visible in my own comment, which said never-run rows sort last and then jumped to page 1 anyway. Reverted to `handleListRefresh`.

Regression test added for the delete gate, verified by reverting the bypass.

One caveat on verification: my local full-suite runs are unreliable right now — the Docker environment was wiped and the test-database lifecycle is thrashing, with the same failure profile on a clean baseline as with my changes. Per-app runs are clean and the new test is mutation-verified, but I am treating CI as the authority for the full suite rather than claiming a green local run.

https://claude.ai/code/session_01BAfubrF8kR2rKewMSjXagg

frontend/src/components/deployments/create-api-deployment-modal/CreateApiDeploymentModal.jsx
kirtimanmishrazipstack: @Deepak-Kesavan Acknowledged — reverted in 040f20546, and no apology needed; I should have caught it.

The contradiction was sitting in my own diff: the comment I wrote said "the list sorts never-run rows last" and then concluded "Go back to page 1". I implemented the suggestion without checking its premise against `nulls_last=True`, which is exactly the failure mode of taking a review suggestion on trust.

`refreshList` is back to `handleListRefresh`. The original half of this finding still stands and is still fixed — the created row no longer renders without owner fields, because the list is refetched rather than prepended from the thin create response. Only the page jump is gone.

Full detail on the thread you opened at `ApiDeployment.jsx:94`.

Deepak-Kesavan:
## 🤖 Automated re-review (pass 3) — clean

Re-reviewed `040f20546` + the `main` merge at `6a61ae6d2`. **No findings.** Both pass-2 items verified fixed, and the sweep claim in your commit message checks out.

**Verified:**
- `IsFrictionLessAdapterDelete` now carries `_is_service_account` in the same position as every sibling (`permission.py:285-286`). `AdapterInstance.is_friction_less` is `default=False`, so `test_adapter_delete_admits_a_service_account` reaches the bypass branch rather than short-circuiting on the frictionless carve-out — the test is not vacuous.
- The sweep is accurate: all 13 `_is_resource_owner(` call sites in OSS carry a bypass, except `prompt_studio_helper.py:102`, whose only production caller returns early at `:237` on `is_service_account`. The enterprise repo has zero such call sites on `main`, so "both repos" is covered.
- `ApiDeployment.jsx` is a clean revert — byte-identical to its `b2d830ed4` state, `useCallback` dropped from the import, no residue. Agreed that making a never-run deployment visible needs an ordering change and that is a separate product call.
- The `main` merge brought nothing into `backend/` beyond your fix; `tenant_account_v2/migrations/` still has exactly one `0006_*`, so no migration-number collision, and the six `*_absorb_shared_users` dependencies all still resolve.

**One candidate raised and refuted, recording it so it does not come back:** creating an adapter through a platform key then deleting it returns 500 `CannotDeleteDefaultAdapter`, because `create` writes the new adapter into the callers `UserDefaultAdapter`. Reproduced — but the control reproduces it identically for a plain human org admin. `UserDefaultAdapter` (`views.py:283`) and the destroy guard (`:355-378`) have no caller-type branching, so this is the long-standing default-adapter rule with its own remedy (`POST /adapter/default_triad/`), not a gap in this fix. Both lines are untouched by this PR.

Nothing blocking from this pass. Worth remembering `Zipstack/unstract-cloud#1753` merges after this one.

<sub>🤖 Unstract PR review kit (Claude Code) · automated pass · findings are starting points, not verdicts</sub>

Deepak-Kesavan (approved):

PR #2269, opened 2026-09-01
UN-4016 [FEAT] Resolve the organisation from a platform API key via a whoami endpoint

praveen-formido:
## What

`GET /api/v1/unstract/whoami/` — an organisation-less endpoint that authenticates with a platform API key and answers `{organization_id, organization_name, permission, key_name}`, read off the key row itself. It reaches the generated OpenAPI spec, so the published client and the CLI pick it up.

Builds on the spec pipeline from #2237 (UN-4009), which merged on 31 Aug — this was written against that branch and has been rebased onto `main` since, with the spec regenerated and re-verified against the merged result.

## Why

Before a caller can use any organisation-scoped endpoint they need `org_id`, and it has no documented source — it is the first path segment of every web-app URL (`useMainAppRoutes.js`, `<Route path=":orgName">`) and nothing says so. A platform key already *carries* its organisation by construction: `PlatformApiKey` inherits `DefaultOrganizationMixin`, whose FK is stamped at mint time from the minting admin's active org, and `key` is globally unique, so a bearer token maps to exactly one row and therefore one organisation. Nothing exposed that association over HTTP.

Every comparable CLI that predates OAuth (`doctl`, `twilio`, `sentry-cli`) issues one opaque token that carries its own scope and is validated the moment it is supplied. This is the endpoint that lets ours do the same.

## How

- **`platform_api/whoami_views.py`** — a field read. `permission_classes = []` because `CustomAuthMiddleware` has already resolved the token to a key row, bound the service account to `request.user` and enforced the tier against the method; a permission class would re-ask a question already answered. `authentication_classes` is deliberately *not* set — the project configures no `DEFAULT_AUTHENTICATION_CLASSES` that resolves a user, so DRF's default returns `None` and the middleware's `request.user` survives into the view. No extra query: `custom_auth_middleware.py` already `select_related`s `organization`.
- **`platform_api/whoami_urls.py` + a mount in `base_urls.py`** — its own module because the spec's urlconf selector can only pick out a mount declared with a dotted module path (the tenant and public mounts pass a *list*, so `urlconf_name.__name__` is `None` for them). Mounted ahead of the tenant urlconf so resolution is deterministic rather than relying on fall-through; organisation-scoped paths are rewritten before routing and none of those rewrites can produce `whoami/`.
[...]

PR #2278, opened 2026-09-08
UN-4016 [FEAT] Publish the API deployment listing in the OpenAPI spec

chandrasekharan-zipstack:
Stacked on #2269 — review that first; this diff is only what sits on top of it.

## What

A platform API key already reaches the organisation-scoped API deployment listing, and gets the whole organisation back. Nothing published it, so no generated client could call it. This publishes the endpoint that is already there.

No endpoint is added, no response shape changes, no route moves.

## Why the spec needed more than an annotation

`OrganizationMiddleware` rewrites `/api/v1/unstract/<org>/…` to `/api/v1/unstract/…` before anything is routed. A spec generated from the URLconf therefore describes a path no caller ever sends. Generation now restores the segment, driven by `ORGANIZATION_MIDDLEWARE_WHITELISTED_PATHS` — the same setting that decides which routes really are served without it — so the two cannot disagree. `whoami` stays organisation-less because that setting says so.

That part is one-time: every organisation-scoped endpoint published from here on gets the segment for free.

## What a CLI caller can now do

`whoami` resolves `organization_id` once; this lists what the organisation has deployed, each entry carrying `api_name` and `api_endpoint` (what an execution call needs) alongside `display_name` and `description` (what the deployment is for). The deployment's own API key is not in the listing, so a platform key cannot be widened into the ability to execute a deployment.

The detail (`GET …/deployment/{id}/`) operation is deliberately not published: the listing already carries every field that answers "what is this deployment for", and the detail serializer is `fields = "__all__"` — a wider public surface for no added answer.

## Notes for review

- The listing's mount is *restated* in `deployment_spec_urls` rather than selected out of the served urlconf, because it is served from one carrying a dozen routes this spec does not publish. Both halves of that restatement — the path and the method — are held to the served route by tests.
- The same path also serves POST to create a deployment. That is not published.
- `queryset = APIDeployment.objects.none()` on the viewset names the model for schema generation; `get_queryset` overrides it on every request and cannot run without an authenticated user.
- The autouse fixture in `test_docstudio_spec` is not cosmetic: generating the spec now reaches the organisation-scoped managers through `DjangoFilterBackend`, and an organisation left in the thread-local by an earlier test sends them to a database those tests do not open.

## Tests

`backend/api_v2/tests/test_deployment_listing.py` — request-level, real URLconf and middleware chain: a key lists its own organisation, cannot list another (403), an unknown or absent key is refused (401), an empty organisation lists nothing, and the listing carries no key material. It also covers deterministic paging, a malformed `workflow` filter (400), filtering by workflow, and the run-summary annotations.

`backend/api_v2/tests/test_docstudio_spec.py` — the documented path matches `reverse("tenant:api_deployment")`, the published method matches the served view's, and the operation declares the organisation it takes.

## Review round

`3167f6433` addresses the review: the listing now ends its ordering with `pk` so paging cannot skip or duplicate a row, a malformed `workflow` filter is a 400 rather than a 500 (and 400 is published), `run_count` is annotated alongside `last_run_time` so a row costs one query rather than three, and `api_name`/`display_name`/`description`/`is_active` are pinned `read_only` so they land in the component's `required` list.

`created_by_email` is kept — it is the same view org members already have in the web app — and the published description now says the entries carry it.

Full backend suite: same 76 failures as the base branch (all local Redis-connection failures), no new ones.

https://claude.ai/code/session_014f9oEEYspPH4fmPULTnLkJ

### DIFF
--- a/backend/api_v2/api_deployment_views.py
+++ b/backend/api_v2/api_deployment_views.py
@@ -9,6 +9,7 @@ from permissions.membership_views import OwnerManagementMixin
 from permissions.permission import IsOwner, IsOwnerOrSharedUserOrSharedToOrg
 from permissions.resource_share_views import ResourceShareManagementMixin
 from permissions.roles import ResourceRole
+from platform_api.openapi_schema import PlatformKeyAutoSchema
 from plugins import get_plugin
 from prompt_studio.prompt_studio_registry_v2.models import PromptStudioRegistry
 from rest_framework import serializers, status, views, viewsets
@@ -33,7 +34,10 @@ from api_v2.exceptions import (
     contains_tool_not_found_error,
 )
 from api_v2.models import APIDeployment
-from api_v2.openapi_schema import DEPLOYMENT_EXECUTION_SCHEMA
+from api_v2.openapi_schema import (
+    API_DEPLOYMENT_LIST_SCHEMA,
+    DEPLOYMENT_EXECUTION_SCHEMA,
+)
 from api_v2.rate_limiter import APIDeploymentRateLimiter
 from api_v2.serializers import (
     APIDeploymentListSerializer,
@@ -246,10 +250,19 @@ class DeploymentExecution(views.APIView):
         )
 
 
+@API_DEPLOYMENT_LIST_SCHEMA
 class APIDeploymentViewSet(
     OwnerManagementMixin, ResourceShareManagementMixin, viewsets.ModelViewSet
 ):
     pagination_class = CustomPagination
+
+    # Names the model for schema generation only: `get_queryset` overrides this
+    # for every request, and it cannot run without an authenticated user.
+    queryset = APIDeployment.objects.none()
+
+    # Narrows the standardized-errors example bodies to the statuses the
+    # authentication middleware answers itself. See PlatformKeyAutoSchema.
+    schema = PlatformKeyAutoSchema()
     notification_resource_name_field = "display_name"
 
     def get_notification_resource_type(self, resource: Any) -> str | None:
--- a/backend/api_v2/api_deployment_views.py
+++ b/backend/api_v2/api_deployment_views.py
@@ -256,12 +256,10 @@ class APIDeploymentViewSet(
 ):
     pagination_class = CustomPagination
 
-    # Names the model for schema generation only: `get_queryset` overrides this
-    # for every request, and it cannot run without an authenticated user.
+    # For schema generation only; get_queryset replaces it on every request.
     queryset = APIDeployment.objects.none()
 
-    # Narrows the standardized-errors example bodies to the statuses the
-    # authentication middleware answers itself. See PlatformKeyAutoSchema.
+    # Error examples have to match what the auth middleware sends.
     schema = PlatformKeyAutoSchema()
     notification_resource_name_field = "display_name"
 
--- a/backend/api_v2/api_deployment_views.py
+++ b/backend/api_v2/api_deployment_views.py
@@ -311,7 +311,8 @@ class APIDeploymentViewSet(
             .order_by(F("last_run_time_annotated").desc(nulls_last=True), "pk")
         )
 
-        # Filter by workflow ID if provided
+        # TODO: replace the hand-read params and their OpenApiParameter
+        # restatements with a FilterSet so the spec cannot drift from the code
         workflow_filter = self.request.query_params.get("workflow", None)
         if workflow_filter:
             try:
--- a/backend/api_v2/api_deployment_views.py
+++ b/backend/api_v2/api_deployment_views.py
@@ -3,7 +3,8 @@ import logging
 import uuid
 from typing import Any
 
-from django.db.models import F, OuterRef, QuerySet, Subquery
+from django.db.models import Count, F, IntegerField, OuterRef, QuerySet, Subquery
+from django.db.models.functions import Coalesce
 from django.http import HttpResponse
 from permissions.membership_views import OwnerManagementMixin
 from permissions.permission import IsOwner, IsOwnerOrSharedUserOrSharedToOrg
@@ -286,19 +287,41 @@ class APIDeploymentViewSet(
             .order_by("-created_at")
             .values("created_at")[:1]
         )
+        run_count_subquery = (
+            WorkflowExecution.objects.filter(pipeline_id=OuterRef("id"))
+            .values("pipeline_id")
+            .annotate(total=Count("id"))
+            .values("total")
+        )
 
         # Avoid per-row queries for owner/co-owner + creator fields in list views
         queryset = (
             APIDeployment.objects.for_user(self.request.user)
             .select_related("created_by")
             .prefetch_related("memberships__user")
-            .annotate(last_run_time_annotated=Subquery(last_run_subquery))
-            .order_by(F("last_run_time_annotated").desc(nulls_last=True))
+            .annotate(
+                last_run_time_annotated=Subquery(last_run_subquery),
+                run_count_annotated=Coalesce(
+                    Subquery(run_count_subquery, output_field=IntegerField()), 0
+                ),
+            )
+            # `pk` last because the primary ordering ties on every deployment
+            # that has never run, and a paging client would then see a row
+            # twice or not at all.
+            .order_by(F("last_run_time_annotated").desc(nulls_last=True), "pk")
         )
 
         # Filter by workflow ID if provided
         workflow_filter = self.request.query_params.get("workflow", None)
         if workflow_filter:
+            try:
+                uuid.UUID(workflow_filter)
+            except ValueError:
+                # Django raises on evaluation, past the handler that turns a bad
+                # request into a 400.
+                raise serializers.ValidationError(
+                    {"workflow": "Must be a valid UUID."}
+                ) from None
             queryset = queryset.filter(workflow_id=workflow_filter)
 
         # Search by display name
--- a/backend/api_v2/openapi_schema.py
+++ b/backend/api_v2/openapi_schema.py
@@ -18,6 +18,7 @@ from drf_spectacular.utils import (
 from platform_api.openapi_schema import PlatformKeyError
 from rest_framework import serializers
 
+from api_v2.models import API_NAME_MAX_LENGTH, DESCRIPTION_MAX_LENGTH
 from api_v2.serializers import (
     APIDeploymentListSerializer,
     APIExecutionResponseSerializer,
@@ -260,10 +261,11 @@ class APIDeploymentSummary(APIDeploymentListSerializer):
     # The model defaults these, so DRF reports them optional --- true of a
     # request body, wrong for a response the server always fills. Left alone,
     # a generated client types them nullable and every caller writes a check
-    # for a key that is always there.
-    api_name = serializers.CharField(read_only=True)
-    display_name = serializers.CharField(read_only=True)
-    description = serializers.CharField(read_only=True)
+    # for a key that is always there. The lengths are restated because DRF drops
+    # them from a read-only field.
+    api_name = serializers.CharField(read_only=True, max_length=API_NAME_MAX_LENGTH)
+    display_name = serializers.CharField(read_only=True, max_length=API_NAME_MAX_LENGTH)
+    description = serializers.CharField(read_only=True, max_length=DESCRIPTION_MAX_LENGTH)
     is_active = serializers.BooleanField(read_only=True)
 
 
--- a/backend/api_v2/openapi_schema.py
+++ b/backend/api_v2/openapi_schema.py
@@ -176,7 +176,9 @@ EXECUTE_ERRORS = {
 EXECUTE_SUMMARY = "Execute an API deployment against documents"
 
 EXECUTE_DESCRIPTION = (
-    "Runs an API deployment, authenticated by the deployment's own key.\n\n"
+    "Runs an API deployment. Takes a deployment key — either the "
+    "deployment's own key, or a global API deployment key that has access "
+    "to it.\n\n"
     "Supply the documents either as `files` (multipart upload) or as "
     "`presigned_urls` (HTTPS S3 URLs), or both — a request carrying neither is "
     f"rejected, and the two together may not exceed "
@@ -188,8 +190,8 @@ EXECUTE_DESCRIPTION = (
 STATUS_SUMMARY = "Read the result of an execution"
 
 STATUS_DESCRIPTION = (
-    "Reads a previously started execution, authenticated by the deployment's "
-    "own key. The read is one-shot: the first call that observes a completed "
+    "Reads a previously started execution, taking the same deployment key "
+    "that ran it. The read is one-shot: the first call that observes a completed "
     "execution acknowledges it and the stored result is discarded, so every "
     "later call for that execution answers 406. Keep the payload of the call "
     "that returns it — it cannot be fetched again.\n\n"
--- a/backend/api_v2/openapi_schema.py
+++ b/backend/api_v2/openapi_schema.py
@@ -257,6 +257,15 @@ class APIDeploymentSummary(APIDeploymentListSerializer):
     `display_name` and `description` say what the deployment is for.
     """
 
+    # The model defaults these, so DRF reports them optional --- true of a
+    # request body, wrong for a response the server always fills. Left alone,
+    # a generated client types them nullable and every caller writes a check
+    # for a key that is always there.
+    api_name = serializers.CharField(read_only=True)
+    display_name = serializers.CharField(read_only=True)
+    description = serializers.CharField(read_only=True)
+    is_active = serializers.BooleanField(read_only=True)
+
 
 # This operation takes a platform key rather than a deployment key, so its
 # credential failures come from the auth middleware in `PlatformKeyError` shape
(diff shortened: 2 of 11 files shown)


backend/api_v2/openapi_schema.py
```diff
@@ -242,3 +244,86 @@ class ErrorResponse(serializers.Serializer):
...
+LIST_API_DEPLOYMENTS_DESCRIPTION = (
+    "List the API deployments of an organisation.\n\n"
+    "Each entry carries the `api_name` and `api_endpoint` an execution call "
+    "needs, alongside the `display_name` and `description` that say what the "
+    "deployment is for. A key holder can therefore discover what is deployed "
+    "without being told, having resolved `org_id` once from `whoami`.\n\n"
+    "The deployment's own API key is not part of this listing, so a platform "
+    "key cannot be widened into the ability to execute a deployment by "
+    "reading it. Executing still needs the deployment key.\n\n"
+    "Results are ordered by most recent run first, and paginated: pass `page` "
```
praveen-formido: **Paginated listing has no deterministic ordering — a CLI paging through it can skip or duplicate deployments.**

`get_queryset` orders by `F("last_run_time_annotated").desc(nulls_last=True)` and nothing else. Every deployment that has never run annotates to `NULL`, so they all tie. `DeterministicOrderingFilter` (written for exactly this) only appends the `pk` tie-breaker when `get_ordering` returns something, and with no `?ordering` and no `ordering` attribute on the viewset it returns `None`, so `filter_queryset` leaves the queryset untouched.

Scenario: an org with 60 never-run deployments, `page_size=50`. Page 1 and page 2 are two separate `LIMIT/OFFSET` queries over a tie-heavy ordering; Postgres is free to return the tied rows in a different order for the second query, so a client that concatenates the pages can list a deployment twice and miss another entirely. This is invisible in the frontend (one page, one query) but is precisely the pattern this description now instructs generated clients to use.

Fix: set `ordering = ["-last_run_time_annotated", "pk"]` on the viewset, or append `"pk"` in `get_queryset`'s `order_by`.

backend/api_v2/api_deployment_views.py
```diff
@@ -246,10 +250,17 @@ def get(
         )
 
 
+@API_DEPLOYMENT_LIST_SCHEMA
```
praveen-formido, line 254: **The published operation inherits an `ordering` query parameter that 500s on most of the values the spec itself advertises.**

`DEFAULT_FILTER_BACKENDS` includes `DeterministicOrderingFilter`, so drf-spectacular emits `ordering` into the committed spec (see `specs/docstudio-oss.json`, the new path's parameter list) — it is now part of the public contract. The viewset sets no `ordering_fields`, so DRF falls back to `get_default_valid_fields`, which accepts any serializer field whose `source` is not a model *property*. `APIDeploymentListSerializer` contributes `run_count`, `last_run_time`, `last_5_run_statuses`, `created_by_email`, `is_owner` and `co_owners_count` — all `SerializerMethodField`s with no matching column.

Scenario: a generated client calls `?ordering=run_count`. DRF passes it through as valid, `queryset.order_by("run_count", "pk")` raises `django.core.exceptions.FieldError`, and the caller gets a 500 for a parameter the spec told it it could send.

Fix: declare an explicit allowlist on the viewset, e.g. `ordering_fields = ["display_name", "api_name", "created_at", "last_run_time_annotated"]`.

backend/api_v2/openapi_schema.py
```diff
@@ -242,3 +244,86 @@ class ErrorResponse(serializers.Serializer):
...
+    `api_name` and `api_endpoint` are what an execution call needs;
+    `display_name` and `description` say what the deployment is for.
+    """
+
+
+# This operation takes a platform key rather than a deployment key, so its
+# credential failures come from the auth middleware in `PlatformKeyError` shape
+# rather than from the project exception handler.
+API_DEPLOYMENT_LIST_QUERY_PARAMETERS = [
+    OpenApiParameter(
```
praveen-formido, line 283: **`workflow` is documented as `format: uuid`, but a malformed value returns 500 rather than 400.**

`get_queryset` does `queryset.filter(workflow_id=workflow_filter)` with the raw query-string value. `Workflow.id` is a `UUIDField`, so `?workflow=not-a-uuid` raises `django.core.exceptions.ValidationError` (`'not-a-uuid' is not a valid UUID`) during queryset evaluation. `drf_standardized_errors` handles DRF exceptions, not Django's core `ValidationError`, so it escapes to a 500.

Before this PR that path was only reachable from the frontend, which always sends a UUID it just read back. Publishing the parameter makes it reachable by any generated client, and hand-written callers get an opaque 500 for a client-side mistake. Worth validating the value (`uuid.UUID(...)` → `ValidationError` from DRF) before filtering, since the operation now declares 400 nowhere and 500 as "could not be served".

backend/api_v2/openapi_schema.py
```diff
@@ -242,3 +244,86 @@ class ErrorResponse(serializers.Serializer):
...
     ),
 )
+
+
+# Declares no field of its own, so a change to the real serializer moves the
+# spec. It carries a caller-facing description and a component name the
+# pagination envelope can be wrapped round without stuttering.
+@extend_schema_serializer(component_name="APIDeploymentSummary")
+class APIDeploymentSummary(APIDeploymentListSerializer):
+    """One API deployment, as it appears in an organisation's listing.
```
praveen-formido, line 262: **The listing publishes `created_by_email` — the creator's email address — to every platform API key holder.**

The PR description and `LIST_API_DEPLOYMENTS_DESCRIPTION` both reason carefully about key material ("the deployment's own API key is not part of this listing"), and `test_the_listing_carries_no_key_material` only asserts that no field name contains `"key"`. But `APIDeploymentSummary` inherits `APIDeploymentListSerializer.Meta.fields`, which includes `created_by_email` (plus `created_by`, `is_owner`, `co_owners_count`).

Scenario: a read-only platform key minted for a CI job now enumerates the email address of every user who created a deployment in the org. That is a wider disclosure than the operation's own description claims, and once it is in the committed spec it is a contract that is awkward to narrow later. If the emails are not part of what the CLI needs, override `Meta.fields` on `APIDeploymentSummary` to the four fields the description actually names; if they are intended, the description and the security test should say so.

backend/api_v2/openapi_schema.py
```diff
@@ -242,3 +244,86 @@ class ErrorResponse(serializers.Serializer):
...
+
+API_DEPLOYMENT_LIST_SCHEMA = extend_schema_view(
+    list=extend_schema(
+        operation_id="list_deployments",
+        tags=["deployment"],
+        auth=[{"platformKey": []}],
+        parameters=API_DEPLOYMENT_LIST_QUERY_PARAMETERS,
+        responses={
+            200: OpenApiResponse(
+                APIDeploymentSummary(many=True),
```
praveen-formido, line 324: **The published 200 costs 3+ queries per row, and `page_size` is documented with a 1000-row ceiling.**

`APIDeploymentListSerializer` resolves `run_count`, `last_run_time` and `last_5_run_statuses` with a separate `WorkflowExecution` query per deployment; `get_queryset` already annotates `last_run_time_annotated` but the serializer ignores it and re-queries. `CustomPagination.max_page_size` is `Pagination.MAX_PAGE_SIZE = 1000`.

Scenario: a generated client (or a retry loop) calls `?page_size=1000` against an org with 1000 deployments and issues ~3000 sequential queries in one request. Until now this endpoint was only driven by the frontend's own page size; publishing `page_size` in the contract makes the worst case something an external caller chooses. Either have the serializer read `last_run_time_annotated` and annotate `run_count`, or lower `max_page_size` for this operation.

praveen-formido (commented):
Careful single-pass review of the diff on top of #2269. Six findings, all about what publishing this listing commits the project to — the spec text and the served behaviour disagree in a few places, and two documented query parameters can be driven into a 500.

backend/api_v2/openapi_schema.py
```diff
@@ -242,3 +244,86 @@ class ErrorResponse(serializers.Serializer):
...
+
+
+# Declares no field of its own, so a change to the real serializer moves the
+# spec. It carries a caller-facing description and a component name the
+# pagination envelope can be wrapped round without stuttering.
+@extend_schema_serializer(component_name="APIDeploymentSummary")
+class APIDeploymentSummary(APIDeploymentListSerializer):
+    """One API deployment, as it appears in an organisation's listing.
+
+    `api_name` and `api_endpoint` are what an execution call needs;
```
praveen-formido, line 264: **`api_name` is optional in the published component, though the description calls it what an execution call needs.**

In `specs/docstudio-oss.json`, `APIDeploymentSummary.required` is `[api_endpoint, co_owners_count, created_by, created_by_email, id, is_owner, last_5_run_statuses, last_run_time, run_count, workflow, workflow_name]` — `api_name`, `display_name`, `description` and `is_active` are absent, because the model gives them defaults and DRF therefore marks them `required=False` for *input*. Without `COMPONENT_SPLIT_REQUEST` the same component is used for the response.

Scenario: the generated Python client types `api_name: str | None` and `display_name: str | None`, so every caller that follows the description's advice ("each entry carries the `api_name` … an execution call needs") has to write a `None` check or an assert for a field the server always sends. Since this serializer is response-only here, the cleanest fix is to pin them on `APIDeploymentSummary`, e.g. `api_name = serializers.CharField(read_only=True)` and the same for `display_name`/`description`/`is_active`.

backend/api_v2/openapi_schema.py
chandrasekharan-zipstack: Confirmed and fixed in 3167f6433. `DeterministicOrderingFilter.get_ordering` returns early on a falsy ordering, and with no `?ordering` and no `ordering` attribute it never appends the tie-breaker, so nothing reached the queryset.

Fixed in `get_queryset` rather than via an `ordering` attribute, so the tie-breaker holds whether or not a caller sends `?ordering` (the filter appends its own `pk` in that case, and a duplicate term is harmless). `test_paging_reaches_every_deployment_exactly_once` pages three never-run deployments one per page and asserts each is seen once.

backend/api_v2/openapi_schema.py
chandrasekharan-zipstack: Confirmed and fixed in 3167f6433. Reproduced: `?workflow=not-a-uuid` was a 500 (`ValueError: badly formed hexadecimal UUID string` out of `UUIDField.to_python`, escaping the handler as you said).

`get_queryset` now parses the value and raises a DRF `ValidationError`, and the operation declares 400 with the `ErrorResponse` shape. That made an existing invariant in `test_docstudio_spec` out of date — it asserted only the deployment operations can document 400 — so it now splits on whether an operation takes a body or a parameter at all, which is the real distinction. `test_a_malformed_workflow_filter_is_a_bad_request` covers the endpoint.

backend/api_v2/openapi_schema.py
chandrasekharan-zipstack: Confirmed and fixed in 3167f6433, taking the first of your two options.

`get_queryset` now annotates `run_count` alongside the `last_run_time` annotation it already had, and the serializer reads both when present, falling back to a query for `by_prompt_studio_tool`, which serializes a plain queryset. Three queries per row become one.

Not fixed: `last_5_run_statuses` still costs a query per row — it needs a per-row `LIMIT 5`, which is a lateral join rather than an annotation, and that felt like more than this PR should carry. `max_page_size` is left at 1000 because `CustomPagination` is shared; worst case is now ~1000 queries rather than ~3000. Say the word if you want either of those in here.

backend/api_v2/openapi_schema.py
chandrasekharan-zipstack: Confirmed and fixed in 3167f6433. `APIDeploymentSummary` now restates `api_name`, `display_name`, `description` and `is_active` as `read_only=True`, which puts all four in the component's `required` list, and `test_the_listed_fields_a_client_reads_are_not_optional` pins that.

Restated on the spec-only subclass rather than on `APIDeploymentListSerializer`, so nothing at request time changes.

backend/api_v2/api_deployment_views.py
chandrasekharan-zipstack: Not reproducible — no change made here. I probed this before acting.

`OrderingFilter.get_valid_fields` builds its list from the serializer, but it skips any field whose `source` is `*` or which has no model field behind it, so the `SerializerMethodField`s never enter it. On this view it returns:

```
['id', 'workflow', 'workflow__workflow_name', 'display_name', 'description', 'is_active', 'api_endpoint', 'api_name', 'created_by']
```

`?ordering=run_count` returns 200, not 500 — the term is silently dropped by `remove_invalid_fields`, exactly as `?ordering=nonsense` would be.

The spec does not enumerate `ordering` values, so there is nothing published that contradicts the behaviour. Happy to be shown a value that does 500 if you have one.

backend/api_v2/openapi_schema.py
chandrasekharan-zipstack: Kept, per a product call on this PR: the listing is the same view org members already have in the web app, and a platform key is an org-scoped credential, not a lower-trust one — it already reaches every deployment in the org and can execute them.

What changed in 3167f6433 is the honesty of the published description: `LIST_API_DEPLOYMENTS_DESCRIPTION` now states that entries carry ownership, recent-run metadata and the creator's email, so nobody generating a client is surprised by what comes back. Dropping the field is a response-shape change to an endpoint the web app already consumes, so it belongs in its own PR if we decide the exposure is wrong.

backend/api_v2/serializers.py
```diff
@@ -544,12 +544,18 @@ def get_is_owner(self, obj) -> bool:
...
+            return annotated
         return WorkflowExecution.objects.filter(pipeline_id=instance.id).count()
 
     def get_last_run_time(self, instance) -> str | None:
-        """Get the timestamp of the most recent execution."""
+        annotated = getattr(instance, "last_run_time_annotated", None)
+        if annotated is not None:
+            return annotated.isoformat()
         last_execution = (
             WorkflowExecution.objects.filter(pipeline_id=instance.id)
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Fixed in 7ace3b779 — right catch. `Coalesce` keeps `run_count_annotated` at 0, but `last_run_time_annotated` is genuinely `NULL` for a never-run deployment, so `is not None` sent exactly the rows the annotation was added for back to the database.

Both getters now use an `_UNANNOTATED` sentinel, so presence decides, not value. `test_a_never_run_deployment_costs_no_extra_summary_query` compares queries touching `workflow_execution` for a one-row page against a three-row page: the delta is 2, one per added row for `last_5_run_statuses`, which is not annotated.

backend/api_v2/openapi_schema.py
```diff
@@ -242,3 +244,103 @@ class ErrorResponse(serializers.Serializer):
...
+    `display_name` and `description` say what the deployment is for.
+    """
+
+    # The model defaults these, so DRF reports them optional --- true of a
+    # request body, wrong for a response the server always fills. Left alone,
+    # a generated client types them nullable and every caller writes a check
+    # for a key that is always there.
+    api_name = serializers.CharField(read_only=True)
+    display_name = serializers.CharField(read_only=True)
+    description = serializers.CharField(read_only=True)
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Fixed in 7ace3b779. DRF drops `max_length` from a read-only field (`include_extra_kwargs` pops it), so `extra_kwargs` would have lost the limits too — the lengths have to be restated on the declared field.

Now `max_length=API_NAME_MAX_LENGTH` on `api_name`/`display_name` and `max_length=DESCRIPTION_MAX_LENGTH` on `description`, imported from `api_v2.models` so they cannot drift from the columns. The spec regains `maxLength: 30`/`30`/`255`.

backend/api_v2/openapi_schema.py
```diff
@@ -180,14 +185,16 @@ class ErrorResponse(serializers.Serializer):
...
 STATUS_DESCRIPTION = (
-    "Read the result of a previously started execution.\n\n"
-    "This read is one-shot: the first call that observes a completed execution "
-    "acknowledges it and the stored result is discarded, so every later call "
-    "for that execution answers 406. Poll while the execution is pending, and "
-    "keep the payload of the call that returns it — it cannot be fetched again."
-    "\n\nA still-running execution answers 422 carrying its current `status`, "
-    "so a polling loop should treat 422 as the normal reply and stop on 200. "
+    "Reads a previously started execution, authenticated by the deployment's "
+    "own key. The read is one-shot: the first call that observes a completed "
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Confirmed and fixed in f5171ff0e — you are right, and the narrowing was mine: neither description mentioned a credential before this round.

`DeploymentHelper` (backend/api_v2/deployment_helper.py:104-112) calls `KeyHelper.validate_api_key`, and on `UnauthorizedKey` falls back to `validate_global_api_deployment_key`. Both `execute` and `status` go through that same `@DeploymentHelper.validate_api_key` decorator, so both accept either key.

`execute` now reads "Takes a deployment key — either the deployment's own key, or a global API deployment key that has access to it", and `status` takes "the same deployment key that ran it". The `deploymentKey` security scheme said "The API deployment's own key" too, which predates this PR and was the same error, so it is widened alongside.

Also pinned: `"global API deployment key"` is now in the behaviour table in `test_docstudio_spec.py`, so the claim cannot be edited away silently.

backend/api_v2/openapi_schema.py
```diff
@@ -242,3 +252,98 @@ class ErrorResponse(serializers.Serializer):
...
+    ),
+]
+
+LIST_API_DEPLOYMENTS_SUMMARY = "List an organisation's API deployments"
+
+LIST_API_DEPLOYMENTS_DESCRIPTION = (
+    "Lists what an organisation has deployed, authenticated by a platform API "
+    "key that belongs to it.\n\n"
+    "The deployment's own API key is not part of this listing; executing a "
+    "deployment still needs that key.\n\n"
```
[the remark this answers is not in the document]
chandrasekharan-zipstack: Correct, and fixed in ae400d8dd — the fallback in `deployment_helper.py:104` does authorize a global API deployment key, so "that key" was wrong.

Took the finding but not the wording: re-listing which keys qualify would put a third copy alongside the `deploymentKey` security scheme and `EXECUTE_DESCRIPTION`, and this commit exists to remove exactly that kind of duplication. Dropping the possessive is enough — "a deployment key ... still needs one" no longer excludes the global key, and the enumeration stays in the one place that owns it.

praveen-formido (approved):

PR #2276, opened 2026-09-02
UN-3883 [FEAT] Cut dashboard metrics cron DB load: narrower source windows, monthly from daily, two new indexes

kirtimanmishrazipstack:
## What

The dashboard metrics cron does far less database work. Four changes, already reviewed and merged individually as #2255, #2264 and #2265:

- **Monthly totals are added up from the daily totals** instead of being recalculated from the raw tables every time.
- **Each run looks back 2 days instead of ~2 months.** A separate pass at 04:40 UTC looks back 7 days once a day, so a short outage still repairs itself.
- **The expensive half of the work runs hourly instead of every 15 minutes.** Hourly figures keep their 15-minute cadence; daily and monthly move to hourly, at :20.
- **Two new database indexes** on the two tables the cron reads most.

## Why

On production the cron was burning roughly **55 minutes of database time every 6 hours**. It re-read between 32 and 62 days of raw data, for all 38 organisations, 96 times a day — to produce monthly figures that only change once a month.

Dashboards look the same and hourly numbers are as fresh as before. Daily and monthly numbers can now be up to an hour behind.

**The tradeoff worth knowing about is not the lag, it's the window.** Monthly is now the sum of the daily rows, so a record that only becomes final more than 7 days after it was created never lands in a daily row, and therefore never lands in the monthly total. That is an under-count, not a delay. We measured it: across 405,951 rows, **zero** took longer than 7 days. It stays safe as long as that holds, so the rollup now reports any monthly total it lowers, and `README.md` documents the same limit.

## How

- `_rollup_monthly_from_daily` replaces the per-org monthly source queries with one `INSERT ... ON CONFLICT DO UPDATE` over `event_metrics_daily` for every org. Upsert-only, per the design agreed on UN-3973: a monthly row the daily tier no longer produces is left in place, because a stale total is recoverable with `backfill_metrics` and a deleted one is not.
- `aggregate_metrics_from_sources(tier, source_window_days)` is called by three schedule rows. Both kwargs travel on both transports — Beat calls the Django task directly, the PG scheduler goes worker proxy → internal endpoint → the same function.
[...]

PR #2280, opened 2026-09-10
UN-4046 [DOCS] Document the PG queue transport in the dashboard-metrics README

jaseemjaskp:
## What

Documentation only. `backend/dashboard_metrics/README.md` described a Celery-only architecture and told readers to start the metrics cron with `celery -A backend beat` and `celery -A backend worker -Q dashboard_metric_events`. Neither process exists in the default compose stack any more.

- Adds a **Transport: how the cron actually fires** chapter.
- Renames `## Celery Tasks` to `## Scheduled Tasks` — the task bodies are shared, only the transport differs.
- Corrects Quick Commands, Setup, Verify, Troubleshooting and File Structure to match the current fleet.
- Keeps the Celery path documented throughout, labelled as the rollback target.

## Why

UN-3796 added the PG path; UN-4046 made it the default, deleted `celery-beat` and `worker-metrics` from `docker/docker-compose.yaml`, and added `worker-pg-metrics`. The README was never updated for either.

The concrete cost: someone onboarding from this file starts a worker the fleet no longer runs, sees no error and no queue depth, and has nothing in the docs pointing at `pg_owned` — which is the flag that actually decides whether anything fires. The README's setup step 5 checked `django_celery_beat.PeriodicTask.enabled` only, which on a current deploy reads `False` on healthy rows.

## How

The new chapter documents:

- **Dual-written schedule rows.** Every `dashboard_metrics.*` schedule is a pair — `django_celery_beat.PeriodicTask` and `pg_queue.PgPeriodicTask` — written from one spec in `migrations/0006_split_aggregation_schedule.py`. `pg_owned` selects the firer.
- **The full PG path**: leader-elected orchestrator tick (`workers/queue_backend/pg_queue/pg_scheduler.py::dispatch_due_periodic_tasks`, enqueue and `next_run_at` advance in one transaction) → `pg_queue_message` on `dashboard_metric_events` → `worker-pg-metrics` → the HTTP proxy in `workers/scheduler/dashboard_metrics_tasks.py` → `internal_views.py` → `tasks.py` verbatim.
- **Why the proxy hop exists** — the workers image has no Django, so the ORM-heavy aggregation cannot run there.
- **Why `worker-pg-metrics` is its own service** — the consumer's health heartbeat freezes while a task runs, so `HEALTH_STALE` bounds one task's wall clock; a minutes-long aggregation would trip the scheduler's 240s bound.
- **The two env gates and their defaults** — `PG_SCHEDULER_ENABLED` (defaults `true`) and `PG_SCHEDULER_ADOPT_PERIODICS` (must be `true` for the periodics), converged by `converge_pg_scheduler` on every backend start.
- **Failure modes specific to this transport** — fire-and-forget with `MAX_ATTEMPTS=1`, `autoretry_for` being inert on the internal-HTTP path, and the silent no-firer state.

Setup step 5 now prints `pg_owned` and `cron_string` from `PgPeriodicTask` alongside the Beat rows.

No. One markdown file changed, no code, no schema, no config, no schedule rows. Nothing imports or renders this README.

## Database Migrations

None.

## Env Config

None. The PR documents two existing variables (`PG_SCHEDULER_ENABLED`, `PG_SCHEDULER_ADOPT_PERIODICS`); it does not introduce or change either.

## Deploy Steps

None.

## Relevant Docs

This PR is the doc change. Source of truth verified against `origin/main` at `335b0a3a`:

- `backend/scheduler/ownership.py` — `PG_SCHEDULER_ENABLED` default
- `backend/entrypoint.sh` — convergence on start, `--periodics` derivation
- `backend/pg_queue/models.py` — `PgPeriodicTask` fields
- `backend/pg_queue/management/commands/converge_pg_scheduler.py` — ownership, both directions
- `workers/queue_backend/pg_queue/pg_scheduler.py` — the dispatch tick
- `workers/scheduler/dashboard_metrics_tasks.py` — the proxy
- `workers/run-worker.sh` — the `pg-metrics` role
- `docker/docker-compose.yaml` — `worker-pg-metrics`, `PG_SCHEDULER_ADOPT_PERIODICS=true`

## Related Issues or PRs

- Context: UN-3796 (added the PG path), UN-4046 (made it the default). #2276 updated this README's windows and schedules but not its transport.

## Dependencies Versions

No dependency changes.

## Notes on Testing

Docs only, so no tests. What was checked:

- Every claim read from the pulled `origin/main` source listed above, not from memory.
- `markdownlint-cli@0.42.0 --disable MD013` (the repo's pre-commit config) reports the **same 63 violations as the base commit** — all pre-existing, zero added. Four were introduced mid-edit (3× MD033 from a `<details>` block, 1× MD040) and removed.
- Internal anchor links resolve; the one `MD051` is the pre-existing `[Frontend Components](#frontend-components)` TOC entry, deliberately left alone as unrelated to this change.

## Screenshots

Not applicable.

## Checklist

### DIFF
--- a/backend/dashboard_metrics/README.md
+++ b/backend/dashboard_metrics/README.md
@@ -176,7 +176,7 @@ The dashboard reads from **pre-aggregated tables** (`event_metrics_hourly`, `eve
 - If the aggregation task fails, the dashboard shows stale data rather than crashing — up to 15 minutes old for hourly figures, up to an hour for daily and monthly.
 - A daily 04:40 UTC reconciliation pass reruns the same task over a 7-day source window, so a **daily- or monthly-tier** gap shorter than that repairs itself without a manual backfill. The hourly tier re-queries the last 24h on every run regardless of tier or window, so an hourly gap shorter than 24h repairs itself on the next tick; only one older than 24h needs `backfill_metrics`.
 - The 7-day window is also the ceiling on lag, not just on downtime. `documents_processed` and `failed_pages` filter on a terminal status but window and bucket on `created_at`, so a row whose status turns terminal more than 7 days after it was created is counted in no daily row — and therefore in no monthly total either, since monthly is the sum of daily. Before the monthly tier was derived from daily this was caught by the wider monthly source window.
-- Retry behaviour depends on the transport: on Celery the tasks carry `max_retries=3` with exponential backoff; on the PG transport (the default) there is **no** retry — `MAX_ATTEMPTS=1`, and the next scheduled tick supersedes a failed run. See [Transport](#transport-how-the-cron-actually-fires).
+- Retry behaviour depends on the transport: on Celery the tasks carry `max_retries=3` with exponential backoff; on the PG transport (the default) retries are governed by the consumer's `MAX_ATTEMPTS`, which compose sets to `1` — a failed run is dropped and the next scheduled tick supersedes it. See [Transport](#transport-how-the-cron-actually-fires).
 - Cleanup tasks (hourly: 30-day retention, daily: 365-day retention) prevent unbounded table growth.
 
 ---
@@ -491,10 +491,20 @@ process — the difference is what schedules it and what carries it.
 
 ### Schedule rows are dual-written
 
-Every `dashboard_metrics.*` schedule exists as a **pair** of rows, written from a single
-spec by the same migration so the two cannot drift: a `django_celery_beat.PeriodicTask`
-and a `pg_queue.PgPeriodicTask`. See `migrations/0006_split_aggregation_schedule.py`
-(`split_schedules`). A row's `pg_owned` flag says which scheduler fires it.
+Every `dashboard_metrics.*` schedule exists as a **pair** of rows: a
+`django_celery_beat.PeriodicTask` and a `pg_queue.PgPeriodicTask`. A row's `pg_owned`
+flag says which scheduler fires it.
+
+How the pair is kept in step differs by row, which matters when you edit one:
+
+| Rows | Declared in | Drift risk |
+|------|-------------|------------|
+| The two aggregation rows, and the reconciliation row | One spec in a single migration — `0005_add_reconciliation_task.py`, `0006_split_aggregation_schedule.py` (`split_schedules`) | None by construction |
+| The three original rows (`aggregate_from_sources`, both cleanups) | **Two separate migrations** — `0002_setup_periodic_tasks.py` (Beat) and its PG twin `0004_pg_periodic_tasks.py` | Real — edit one and forget the other and they diverge |
+
+The second row is guarded only by a test, `tests/test_pg_periodic_task_declarations.py`,
+which exists for exactly that failure mode. If you change one of those three schedules,
+change both migrations.
 
 ### PG transport (default)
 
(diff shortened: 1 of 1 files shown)


jaseemjaskp:
## Self-review (`/pr-review`) — findings and resolutions

Two review agents ran against this draft. Since it is a docs PR, the only thing that matters is whether the text is **true** and will **stay** true, so both passes fact-checked every technical claim against the live source rather than reading for style. Every finding below was independently re-verified before being acted on.

**Three findings were factually wrong claims I had introduced.** All fixed in `0fa100ff` and `a32e7a14`.

### Critical

| # | Finding | Resolution |
|---|---------|------------|
| 1 | **The process that fires the cron was never listed as something to start.** Setup listed only the metrics consumer and stated "there is no separate Beat-equivalent process to start". True of compose, false for the host-runner flow the same paragraph recommends: the tick runs in the reaper (`reaper.py:1310` → `dispatch_due_periodic_tasks`). Following it literally reproduces the silent no-firer state this PR exists to document. | `0fa100ff` — both processes listed in Quick Commands and Setup step 4, with the reason. |
| 2 | **The dual-write claim was wrong for 3 of the 5 rows.** I wrote that every schedule is written "from a single spec by the same migration so the two cannot drift", citing `0006`. That holds only for the aggregation rows and the reconciliation row. The three original rows come from **two** migrations — `0002` (Beat) and its PG twin `0004`. `0006`'s own docstring says so: *"cannot drift the way 0002 and 0004 can"*. | `a32e7a14` — replaced with a table splitting the two cases, pointing at the drift-guard test `tests/test_pg_periodic_task_declarations.py`. |
| 3 | **The rollback runbook could cause the outage it warns about.** Releasing the metrics rows needs `--periodics` in *both* directions, so `PG_SCHEDULER_ADOPT_PERIODICS` must stay `true` while `PG_SCHEDULER_ENABLED` goes `false`. Unsetting both — the intuitive "roll it all back" — leaves the rows `pg_owned` with Beat disabled: no firer. | `a32e7a14` — callout rewritten to lead with the flag interaction. |

### Important

- **`entrypoint.sh` does not converge on "every backend start"** — the block is gated on `--migrate` (`entrypoint.sh:27`). Compose passes it; a deployment running migrations as a separate job does not, which is exactly where the "ownership is a values change" promise breaks. Now stated precisely, with the caveat called out.
- **"the PG scheduler" was ambiguous** against the `worker-pg-scheduler` service named earlier in the same chapter — which is the pipeline-task consumer and fires no periodics. Someone debugging the silent state would have checked the wrong service and found it healthy. Now named `worker-pg-reaper` explicitly, with the distinction spelled out.
- **A pre-existing `max_retries=3` line contradicted the new failure-modes section.** Qualified by transport.

### Corrections to my own new text

- `MAX_ATTEMPTS=1` is one service's env in compose, **not** a property of the transport — the consumer default is `5`. Now attributed, so nobody assumes it holds in k8s.
- "logs at ERROR rather than raising" conflated two cases: a non-200 **does** raise; only a 200 reporting per-org `errors` is logged instead.
- The `autoretry_for` bullet had the right conclusion via the wrong mechanism — the decorator *is* in the call path; `retry()` re-raises when a task is called directly rather than dispatched by a worker.
- The endpoint in the diagram is relative to `INTERNAL_API_BASE_URL` (which ends `/internal`), so it can't be pasted into curl as-is. Noted.
- `--periodics` covers **every** mirrored non-pipeline periodic, not only `dashboard_metrics.*`. Hedged so it stays true when another is mirrored.
- Stopped quoting exact liveness numbers as fact — they are `${VAR:-default}` overrides and k8s takes its values from a chart in another repo.

### Deliberately not changed

- The remaining Celery invocations are all inside explicitly labelled rollback/legacy blocks — intentional, per the stated scope of keeping Celery documented as the rollback target.
- Reviewer flagged the File Structure tree gaining migration entries as mild scope creep. Left as-is: `0004` and `0006` are the transport's own migrations and the tree previously stopped at `0002`, which was itself misleading.
- `run-worker.sh`'s usage text omits the valid `pg-metrics` role (`run-worker.sh:227-235`). Real, but a script bug and out of scope for a docs PR — worth a follow-up.

### Verification

- Every claim traced to a source `file:line` on `origin/main` @ `335b0a3a`.
- All internal anchors resolve.
- `markdownlint-cli@0.42.0 --disable MD013` (the repo's own pre-commit config): **62 violations vs the base commit's 63** — none added, and the one removed is the `MD051` from the broken `#frontend-components` TOC anchor this PR repairs.

Not marking ready — leaving that to the author's call.

kirtimanmishrazipstack (approved):

